Issue #985, opened 2017-10-11
[python] Parallel training support for Python.


geoHeil:
Does the python package support parallel learning? If not are there plans to  offer support here in the future?

guolinke:
not yet. 
However, the c_api will expose the parallel related parameters.

@wxchan do you have time to enable this ?

TomAugspurger:
FWIW, I'm interested in this as well. I'd like to write a small library enabling [dask](http://dask.pydata.org/en/latest/) arrays to be used with LightGBM, similar to https://github.com/dask/dask-xgboost for xgboost.

Let me know if there's anything I could do to help with testing.

guolinke:
@TomAugspurger 
Thanks for your interesting. It will be very awesome if LightGBM could be in dask as well. 

network api is ready: https://github.com/Microsoft/LightGBM/blob/master/include/LightGBM/c_api.h#L738-L755 
@wxchan could you help to wrap the network api in python package ? 

BTW,  following is the pipeline of parallel training: 
1. each work get its own partitioned data
2. collect the ip and ports of all works
3. Initialize the Network (based on ip/ports, num_worker, ...)
4. use local data to construct the dataset (all workers should do this)
5. construct Booster ( should set parameter ```tree_learner``` )
6. training normally
7. Finalize the Network

It is almost the same as the training of single machine.

julioasotodv:
+1 for this feature. Dask + LightGBM would be the definitive solution for ML on large tabular data

TomAugspurger:
I see that the PR adding the python bindings has been merged, thanks.

I'll give the dask wrapper a shot next week then, and I'll post here if I hit any issues.

TomAugspurger:
I hope you don't mind if I use this thread as a support channel as I implement this. I'll repay with a PR updating the docs :)

Currently I have https://gist.github.com/a794b3a153548966baf5b62e0806b0b4. It seems like the new network APIs are being called correctly, and the output is:


```
scheduler tcp://127.0.0.1:60482
[LightGBM] [Info] Total Bins 720
[LightGBM] [Info] Number of data: 100, number of used features: 20
[LightGBM] [Warning] Set TCP_NODELAY failed.
[LightGBM] [Info] Trying to bind port 12400...
[LightGBM] [Info] Binding port 12400 succeeded
[LightGBM] [Warning] Set TCP_NODELAY failed.
[LightGBM] [Info] Listening...
[...]

PR #1019, opened 2017-10-25
[python] add network config api 

wxchan:
#985


guolinke:
@wxchan I think we need a simple test for this. 
It seems travis can listen the inbound port: https://github.com/travis-ci/travis-ci/issues/1304

PR #908, opened 2017-09-11
[optional] support loading and saving models with protobuf

wxchan:
related to #372 

- [x] solve conflicts
- [x] test windows support: mingw can work, vs not tested yet

PR #1035, opened 2017-11-02
[python] support protobuf

wxchan:


src/c_api.cpp
```diff
@@ -841,6 +852,31 @@ int LGBM_BoosterLoadModelFromString(
   API_END();
 }
 
+int LGBM_BoosterCreateForPythonProto(
```
wxchan: @guolinke I use two functions: one to create booster, one to pass pandas category from c++ to python. Is there a better idea?

guolinke:
@wxchan it seems this PR will break the VS compile as well. 
Actually, I would like the cross-platform solution, otherwise the maintain of code is much more complex.

guolinke:
BTW, it seems protobuf isn't have a good compression ratio: https://medium.com/@nitinpaliwal87/compression-and-serialization-techniques-benchmarking-fd1f34c1098b

guolinke:
I suggest to move to msgpack: https://github.com/msgpack/msgpack-c 
It uses binaries and cross platform

chivee:
@wxchan , what aspects should we benefit from is using protobuf? does some widely used  machine learning can leverage the protobuf version of LGB? it will be great that the serialization can be used across platform as well as languages. Protobuf cannot be supported by VS, which will break the developing line inside MS, I would like to agree with @guolinke to find an alternative for the serialization.

more options:
https://github.com/Microsoft/LightGBM/pull/908#issuecomment-335122669

wxchan:
@chivee @guolinke I agree too. You can revert existing protobuf codes.

StrikerRUS:
@guolinke Will the protobuf be reverted or will it be left till the alternative appears? I'm asking due to the need of docs update.

StrikerRUS:
ping @guolinke

chivee:
@StrikerRUS , we could left it till the alternative solution appears, marked (deprecated). @guolinke , we can give a schedule on the retirement.

guolinke:
@wxchan it seems the msgpack-c is very similar with protobuf. 
Can you replace current protobuf implementation with it ?

Issue #902, opened 2017-09-10
opencl version was incorrectly detected.


cjliux:
I followed the installation guide in wiki to build the gpu version of lightGBM. However, the version detected when I entered `sudo cmake -USE_MPI=ON -USE-GPU=1 ..` is 2.0, while the version actually installed is 1.0. 
Do any one know the reason?

huanzhang12:
The OpenCL 2.0 library is probably a generic ICD (installable client driver) library and headers provided by the OS, which is automatically detected by CMake. At runtime, the ICD library dispatches actual OpenCL function calls to the vendor specific OpenCL libraries provided by AMD/Nvidia/Intel. 

In LightGBM, only OpenCL 1.2 is required, and OpenCL 2.0 should be backwards compatible. In fact, all NVIDIA GPUs support 1.2 only; most AMD GPUs after 2014 theoretically support 2.0, but getting it working on Linux can be tricky. If LightGBM compiles and works fine for you, you don't need to worry about it.

cjliux:
Actually, OpenCL2.0 isn't present on my machine, and cmake raised the error because it cannot find the correct version identifier in the `libopencl.so` file since it had recognised it to be of version 2.0 rather than 1.2.

huanzhang12:
@cjliux Can you provide more information, including your GPU vendor and CMake log?

cjliux:
In [2]: import lightgbm
---------------------------------------------------------------------------
OSError                                   Traceback (most recent call last)
<ipython-input-2-bd597743fb90> in <module>()
----> 1 import lightgbm

/opt/LightGBM/python-package/lightgbm/__init__.py in <module>()
      6 from __future__ import absolute_import
      7 
----> 8 from .basic import Booster, Dataset
      9 from .callback import (early_stopping, print_evaluation, record_evaluation,
     10                        reset_parameter)

/opt/LightGBM/python-package/lightgbm/basic.py in <module>()
     30 
     31 
---> 32 _LIB = _load_lib()
     33 
     34 

/opt/LightGBM/python-package/lightgbm/basic.py in _load_lib()
     25     if len(lib_path) == 0:
     26         return None
---> 27     lib = ctypes.cdll.LoadLibrary(lib_path[0])
     28     lib.LGBM_GetLastError.restype = ctypes.c_char_p
     29     return lib

/usr/lib/python2.7/ctypes/__init__.pyc in LoadLibrary(self, name)
    438 
[...]

PR #1037, opened 2017-11-04
[docs] Update GPU-Tutorial.rst

### LINKED ISSUE #1028: [Python] GPU not working python package
Please search your question on previous issues, [stackoverflow](https://stackoverflow.com/questions/tagged/lightgbm) or other search engines before you open a new one.

For bugs and unexpected issues, please provide following information, so that we could reproduce on our system.

## Environment info
CPU: i7, Nvidia 1060
C++/Python/R version: Python 2.7

## Error Message:
LightGBMError: bin size 5858 cannot run on GPU
## Reproducible examples
lgb.train({'device':'gpu'}, ds)
## Steps to reproduce

Hi, I am sorry if this is already been answered but I did not find the answer for this. I was able to install the gpu version of lightgbm and I ran this with sucess: ./lightgbm config=lightgbm_gpu.conf data=higgs.train valid=higgs.test objective=binary metric=auc

However, with all the defaults, when I try to use the python package, with gpu support, this error occurs. Any idea of what might be causing this? Thank you.

chivee:
fix #1028. #902

### DIFF
--- a/docs/GPU-Tutorial.rst
+++ b/docs/GPU-Tutorial.rst
@@ -61,6 +61,8 @@ Now we are ready to checkout LightGBM and compile it with GPU support:
     cd LightGBM
     mkdir build ; cd build
     cmake -DUSE_GPU=1 .. 
+    # if you have installed the Nvidia OpenGL, please using following instead
+    # sudo cmake -DUSE_GPU=1 -DOpenCL_LIBRARY=/usr/local/cuda/lib64/libOpenCL.so -OpenCL_INCLUDE_DIR=/usr/local/cuda/include/ ..
     make -j$(nproc)
     cd ..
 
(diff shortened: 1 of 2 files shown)


docs/GPU-Tutorial.rst
```diff
@@ -61,6 +61,8 @@ Now we are ready to checkout LightGBM and compile it with GPU support:
     cd LightGBM
     mkdir build ; cd build
     cmake -DUSE_GPU=1 .. 
+    # if you have installed the Nvidia OpenGL, please using following instead
```
StrikerRUS: NVIDIA

StrikerRUS:
Maybe we need the same note directly into the Installation Guide?
https://github.com/Microsoft/LightGBM/blob/master/docs/Installation-Guide.rst#build-gpu-version

Issue #1036, opened 2017-11-02
Quantile regression in LightGBM


mandeldm:
Currently LightGBM does not appear to support 'quantile' loss metric. (This is a feature that can be found in sklearn's Gradient Boosting Regression http://scikit-learn.org/stable/auto_examples/ensemble/plot_gradient_boosting_quantile.html - helpful if one wishes to construct margins of error / probability distribution for each prediction.)

Are there plans to include quantile regression in this product?

Thank you.

mandeldm:
Adding a possible starting point:
https://www.bigdatarepublic.nl/regression-prediction-intervals-with-xgboost/ wherein Benoit Descamps provides an implementation of `quantile_loss` for xgboost (his code and results are near the bottom of the page).

His version of the loss function includes additional parameters which he calibrates using grid search. I don't feel qualified to say whether this implementation is completely correct or optimal.

ClimbsRocks:
This would be awesome, and for me, illustrates one of the biggest gaps between using ML academically, and using it in production environments. 

Every time we ship a new ML predictor, one of the first questions people ask is "Can you tell me which of these predictions you're confident in, and which ones have more uncertainty, so we can take different actions accordingly?". Sklearn's implementation of this is the best answer I've yet found to that question. But it's _so_ slow. 

If we could have that same functionality, but at lightgbm speeds and accuracies, we'd be pretty darn happy :) At that point, we'd probably make prediction intervals the default behavior in auto.ml, because this functionality is so frequently useful.

guolinke:
I think the first problem is to define a metric for quantile loss. Any suggestions?

mandeldm:
I would suggest the same metric (sometime called the "check" function, probably for the checkmark shape of its graph) that `QuantileLossFunction` uses in https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/ensemble/gradient_boosting.py

My intuition for it is that the [mean absolute error is the loss function for the median](https://math.stackexchange.com/a/113279) = 50th percentile, and "tilting" the MAE graph to the right (for lower percentiles) or to the left (for higher) gives us the general quantile loss function. I also found [this brief lecture video](https://www.youtube.com/watch?v=sgR55l054DQ) and [this explanation](https://stats.stackexchange.com/a/252043) helpful.

[...]

PR #1043, opened 2017-11-07
quantile objective function &  metric

guolinke:
#1036


wxchan:
@guolinke our features become more and more, maybe we keep some basic features to make a simple version and control it by macro?

guolinke:
@wxchan 
I think maintain macros is much complicate. As far as these feature don't impact the efficiency, I think we can put them together.

PR #1045, opened 2017-11-09
Add func to handle sparse testing data

WWLoveBasketball:
temp version, updating

### DIFF
--- a/include/LightGBM/tree.h
+++ b/include/LightGBM/tree.h
@@ -515,11 +515,7 @@ inline int Tree::GetLeafByMap(const std::unordered_map<int, double>& feature_val
     }
   } else {
     while (node >= 0) {
-      if(feature_values.count(split_feature_[node]) > 0) {
-        node = NumericalDecision(feature_values.at(split_feature_[node]), node);
-      } else {
-        node = threshold_[node] >= 0 ? left_child_[node] : right_child_[node];
-      }
+      node = NumericalDecision(feature_values.count(split_feature_[node]) > 0 ? feature_values.at(split_feature_[node]) : 0.0f, node);
     }
   }
   return ~node;
--- a/include/LightGBM/tree.h
+++ b/include/LightGBM/tree.h
@@ -511,12 +511,12 @@ inline int Tree::GetLeafByMap(const std::unordered_map<int, double>& feature_val
   int node = 0;
   if (num_cat_ > 0) {
     while (node >= 0) {
-      node = Decision(feature_values.count(split_feature_[node]) > 0 ? feature_values[split_feature_[node]] : 0.0f, node);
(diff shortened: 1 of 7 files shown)


include/LightGBM/tree.h
```diff
@@ -484,6 +507,25 @@ inline int Tree::GetLeaf(const double* feature_values) const {
...
+  if (num_cat_ > 0) {
+    while (node >= 0) {
+      node = Decision(feature_values.count(split_feature_[node]) > 0 ? feature_values.at(split_feature_[node]) : 0.0f, node);
+    }
+  } else {
+    while (node >= 0) {
+      if(feature_values.count(split_feature_[node]) > 0) {
+        node = NumericalDecision(feature_values.at(split_feature_[node]), node);
+      } else {
+        node = threshold_[node] >= 0 ? left_child_[node] : right_child_[node];
```
guolinke: I think it is better to use ```node = NumericalDecision(feature_values.count(split_feature_[node]) > 0 ? feature_values.at(split_feature_[node]) : 0.0f, node);``` .

src/application/predictor.hpp
```diff
@@ -234,6 +265,7 @@ class Predictor {
   int num_pred_one_row_;
   int num_threads_;
   std::vector<std::vector<double>> predict_buf_;
+  std::vector<std::unordered_map<int, double>> predict_buf_map_;
```
guolinke: Not allocate for ```predict_buf_map_``` ?

src/application/predictor.hpp
```diff
@@ -1,6 +1,9 @@
 #ifndef LIGHTGBM_PREDICTOR_HPP_
 #define LIGHTGBM_PREDICTOR_HPP_
 
+#define MAX_FEATURE 100
```
guolinke: max feature could be larger. e.g. 10k

src/application/predictor.hpp
WWLoveBasketball: I have written the func 'CopyToPredictMap' to assign the map,
and I think we do not need to allocate the map.

guolinke:
@WWLoveBasketball  you should fix this test: https://travis-ci.org/Microsoft/LightGBM/jobs/300027913

src/io/tree.cpp
```diff
@@ -411,6 +442,40 @@ std::string Tree::NodeToIfElse(int index, bool is_predict_leaf_index) const {
   return str_buf.str();
 }
 
+std::string Tree::NodeToIfElseByMap(int index, bool is_predict_leaf_index) const {
+	std::stringstream str_buf;
+	str_buf << std::setprecision(std::numeric_limits<double>::digits10 + 2);
+	if (index >= 0) {
+		// non-leaf
```
guolinke: please use 2 spaces for indent, not the tab.

guolinke:
@wxchan  can you help to review the if-else part ?

PR #1049, opened 2017-11-11
trivial refine CheckElementsIntervalClosed & ObtainMinMaxSum

wxchan:
save some compare times

### DIFF
--- a/include/LightGBM/utils/common.h
+++ b/include/LightGBM/utils/common.h
@@ -641,7 +641,7 @@ inline void ObtainMinMaxSum(const T1 *w, int nw, T1 *mi, T1 *ma, T2 *su) {
       minw = std::min(minw, w[i]);
       maxw = std::max(maxw, w[i - 1]);
     }
-    sumw = w[i - 1] + w[i];
+    sumw += w[i - 1] + w[i];
   }
   if (mi != nullptr) {
     *mi = minw;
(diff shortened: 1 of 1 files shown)


include/LightGBM/utils/common.h
```diff
@@ -582,11 +582,29 @@ static void ParallelSort(_RanIt _First, _RanIt _Last, _Pr _Pred) {
 // Check that all y[] are in interval [ymin, ymax] (end points included); throws error if not
 template <typename T>
 inline void CheckElementsIntervalClosed(const T *y, T ymin, T ymax, int ny, const char *callername) {
-  for (int i = 0; i < ny; ++i) {
-    if (y[i] < ymin || y[i] > ymax) {
-      std::ostringstream os;
-      os << "[%s]: does not tolerate element [#%i = " << y[i] << "] outside [" << ymin << ", " << ymax << "]";
-      Log::Fatal(os.str().c_str(), callername, i);
+  auto fatal_msg = [&](int i) {
```
guolinke: it is better to only pass used variable names, instead of capturing all.

include/LightGBM/utils/common.h
```diff
@@ -595,17 +613,45 @@ inline void CheckElementsIntervalClosed(const T *y, T ymin, T ymax, int ny, cons
...
+  }
+  for (; i < nw; i += 2) {
+    if (w[i - 1] < w[i]) {
+      minw = std::min(minw, w[i - 1]);
+      maxw = std::max(maxw, w[i]);
+    } else {
+      minw = std::min(minw, w[i]);
+      maxw = std::max(maxw, w[i - 1]);
+    }
+    sumw = w[i - 1] + w[i];
```
guolinke: sumw += w[i - 1] + w[i]; ?

PR #1055, opened 2017-11-14
fix sklearn objective function

guolinke:
for the bug report in #1036

### DIFF
--- a/python-package/lightgbm/sklearn.py
+++ b/python-package/lightgbm/sklearn.py
@@ -241,7 +241,6 @@ class LGBMModel(_LGBMModelBase):
             raise LightGBMError('Scikit-learn is required for this module')
 
         self.boosting_type = boosting_type
-        self.objective = objective
         self.num_leaves = num_leaves
         self.max_depth = max_depth
         self.learning_rate = learning_rate
@@ -264,7 +263,7 @@ class LGBMModel(_LGBMModelBase):
         self._best_score = None
         self._best_iteration = None
         self._other_params = {}
-        self._objective = None
+        self._objective = objective
         self._n_features = None
         self._classes = None
         self._n_classes = None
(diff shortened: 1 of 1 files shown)


python-package/lightgbm/sklearn.py
```diff
@@ -370,7 +370,7 @@ def fit(self, X, y,
         For multi-class task, the y_pred is group by class_id first, then group by row_id.
         If you want to get i-th row y_pred in j-th class, the access way is y_pred[j * num_data + i].
         """
-        if not hasattr(self, '_objective'):
+        if self.objective is not None:
```
guolinke: @wxchan why we need the objective and _objective ? they are confusing.

guolinke: the ```_objetcive``` is set to ```None``` by default, so this the ```_objective``` will never by update by the ```objective```.

wxchan: I am not sure. It's added by @StrikerRUS .

StrikerRUS: `objective` was for users to see what they've passed; `_objective` was for internal usage and `objective_` (public `_objective`) was for users to see what actual objective is used in the model.

It seems that it was really bugged implementation.

Is it OK that in case of multiclass we overwrite any objective?
https://github.com/Microsoft/LightGBM/blob/bd5e5e3ebb8c93da0ef13d76eb54618fd2d95426/python-package/lightgbm/sklearn.py#L634-L636

guolinke: @StrikerRUS I think the multi-class objective can only accept ```multiclass``` and ```multiclassova```. Maybe it is better for you to fix this. I can close this PR.

StrikerRUS: @guolinke As you wish. Or I can create PR to `sklearn-obj` branch.

BTW, I can't find any mentions of `multiclassova` in the docs...

guolinke: @StrikerRUS sorry, forget to add it. ```multiclassova``` is using k binary classification to do multi-class classification, each binary classification is used to classifier one class.

StrikerRUS: @guolinke thank you, I'll create a PR soon.

PR #1059, opened 2017-11-15
[python] [docs] fixed objective in sklearn wrapper; added missed objectives & metrics to docs

StrikerRUS:
Refer to #1055.

### DIFF
--- a/python-package/lightgbm/sklearn.py
+++ b/python-package/lightgbm/sklearn.py
@@ -285,6 +285,8 @@ class LGBMModel(_LGBMModelBase):
     def set_params(self, **params):
         for key, value in params.items():
             setattr(self, key, value)
+            if hasattr(self, '_' + key):
+                setattr(self, '_' + key, value)
             self._other_params[key] = value
         return self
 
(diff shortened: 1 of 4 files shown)


python-package/lightgbm/sklearn.py
```diff
@@ -285,6 +285,8 @@ def get_params(self, deep=True):
     def set_params(self, **params):
         for key, value in params.items():
             setattr(self, key, value)
+            if hasattr(self, '_' + key):
```
chivee: if an params was named `other_params`, could that disable line 290?

chivee:
@StrikerRUS , very detailed work, it will be great help for the users.

StrikerRUS:
@guolinke I've added note about `ignore_column`

python-package/lightgbm/sklearn.py
StrikerRUS: @chivee Yeah, you're right!
Any parameter with underscore (`_Booster`, `_classes`, especially `_n_features`) will break normal logic in some way. I think that we're assuming that LightGBM user is rather smart person and wont do this 😃 .

StrikerRUS:
@wxchan Please take a look at Python part of the code.

PR #1063, opened 2017-11-16
Update common.h

qrqpjxq:
add feature value handle 'null' and 'NULL'

### DIFF
--- a/include/LightGBM/utils/common.h
+++ b/include/LightGBM/utils/common.h
@@ -248,7 +248,8 @@ inline static const char* Atof(const char* p, double* out) {
     if (cnt > 0) {
       std::string tmp_str(p, cnt);
       std::transform(tmp_str.begin(), tmp_str.end(), tmp_str.begin(), Common::tolower);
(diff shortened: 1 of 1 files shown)


include/LightGBM/utils/common.h
```diff
@@ -248,7 +248,8 @@ inline static const char* Atof(const char* p, double* out) {
     if (cnt > 0) {
       std::string tmp_str(p, cnt);
       std::transform(tmp_str.begin(), tmp_str.end(), tmp_str.begin(), Common::tolower);
-      if (tmp_str == std::string("na") || tmp_str == std::string("nan")) {
+      if (tmp_str == std::string("na") || tmp_str == std::string("nan") ||
+          tmp_str == std::string("null") || tmp_str == std::string("NULL")) {
```
guolinke: line 250 already convert to lower case. So not need the  "NULL" .

qrqpjxq: ok. i change it

Issue #954, opened 2017-10-01
inconsistent parameter names: "n_estimators"


zkurtz:
The [python API](http://lightgbm.readthedocs.io/en/latest/python/lightgbm.html#lightgbm-package) has an "n_estimators" argument.  It took me a while to figure out that this is the same as "num_boost_round", since this parameter is not even mentioned on [the main parameters page](https://github.com/Microsoft/LightGBM/blob/master/docs/Parameters.md).

It seems like it would be simplest to replace `n_estimators` with `num_boost_round` everywhere. Short of that, update the documentation?

Laurae2:
@zkurtz This is the fault of the scikit-learn API, not the Python API (look up for parameters of `sklearn.ensemble.GradientBoostingClassifier` or similar).

zkurtz:
Oh, OK.  I guess that means that there's no point in trying to replace the `n_estimators` parameter. But I propose replacing line 163 [here](https://github.com/Microsoft/LightGBM/blob/0350a9a6ff3a72e9c24eb28038bd10d4f4408c47/python-package/lightgbm/sklearn.py) with something like "Number of boosted trees to fit (i.e. num_boost_round in lightgbm.train)" to be clear about the relationship between the lightgbm parameters and the scikit-learn parameters.

Laurae2:
The parameter is the same as [scikit-learn](http://scikit-learn.org/stable/modules/generated/sklearn.ensemble.GradientBoostingClassifier.html).

If we were to add the equivalent to the true API, we would have to add them to each parameter. However, `n_estimators` ("Number of boosted trees to fit.") is very easy to understand as it is equal to the number of boosting iterations (which is the same as `num_boost_round`), unless the user does not know what is boosting (and in that case, the user should just avoid using boosting due to its overfitting nature).

zkurtz:
I'm well aware of what `num_boost_round` means, but was not previously familiar with the sklearn API and `n_estimators` seemed ambiguous to me.  For one thing, if sounds like it could refer to a collection of boosted trees, treating the output of a "single" lightgbm instance (with, say, `num_boost_round = 100`) as one estimator. 

If your lightgbm sklearn documentation is really just a copy of the sklearn documentation with no direct translation between the lightgbm parameters and the sklearn parameters, would it be sufficient then to just refer the reader to the sklearn documentation rather than reproducing it?

mathcass:
[...]

PR #1079, opened 2017-11-22
[CLI][python-package][docs] Add n_estimators as num_iteration alias

mathcass:
Scikit-Learn's ensemble methods use the term `n_estimators` for the number of
iterations of training models. To make it more accessible for newcomers who are
familiar with Scikit-Learn, it would help if the Parameters page mentioned
`n_estimators` and what parameter that maps to within LightGBM's API.

Addresses discussion brought up in #954 

---

I felt like this might help smooth adoption for any people coming from Scikit-Learn's API.


Laurae2:
@mathcass Did you add the parameter in C++? I can't find it in your commit.

As `n_estimators` seems to be a scikit-learn exclusive parameter, it should not be in docs/Parameters.rst .

Laurae2:
ping @wxchan or @StrikerRUS

StrikerRUS:
I suppose the best solution is to edit the first Note in `Parameters.rst`:
>... methods instead and ``n_estimators`` in sklearn wrapper

mathcass:
@StrikerRUS, Thanks!

Here's how changing the wording on the Scikit-Learn API would look https://github.com/mathcass/LightGBM/commit/8ad98ea568780902045f74d775fddd820c8e19c2
I think in some sense, that may make more sense, because users of other APIs don't need to know a thing about `n_estimators` and thus don't need another term to get confused about. @Laurae2's original question in the PR made me realize that `Paramters.rst` might not be the **best** place for this.

Alternatively, here's how your recent suggestion would look, https://github.com/mathcass/LightGBM/commit/07482be795426179a916b57c3c2c244a8b7714e8
so that it sits alongside the notes about Python & R-specific parameters. 

I think personally, I favor the note being specifically in the sklearn implementation because the sklearn API sits atop the core API and not vice versa.

StrikerRUS:
@mathcass What do you think about merging two your variants into one? I think it's good point to write as more details as we can if it leads to better users' understanding.

But please notice that your second variant needs paraphrase because there is no `train` and `cv` methods in sklearn wrapper.

Laurae2:
I think it is better to add the parameter both in **C++ and the documentation** so there are no confusion for quick readers who don't read the documentation carefully.

StrikerRUS:
@mathcass 
OK, update please default value here to 100 for consistency:
https://github.com/Microsoft/LightGBM/blob/8a5ec366aa5b0ec6301cb6ca8e85a70cc7ac4cf1/python-package/lightgbm/sklearn.py#L136
https://github.com/Microsoft/LightGBM/blob/8a5ec366aa5b0ec6301cb6ca8e85a70cc7ac4cf1/python-package/lightgbm/sklearn.py#L157
https://github.com/Microsoft/LightGBM/blob/8a5ec366aa5b0ec6301cb6ca8e85a70cc7ac4cf1/python-package/lightgbm/engine.py#L325
https://github.com/Microsoft/LightGBM/blob/8a5ec366aa5b0ec6301cb6ca8e85a70cc7ac4cf1/python-package/lightgbm/engine.py#L310



Add `n_estimators` here:
https://github.com/Microsoft/LightGBM/blob/8a5ec366aa5b0ec6301cb6ca8e85a70cc7ac4cf1/python-package/lightgbm/engine.py#L96
https://github.com/Microsoft/LightGBM/blob/8a5ec366aa5b0ec6301cb6ca8e85a70cc7ac4cf1/python-package/lightgbm/engine.py#L386
https://github.com/Microsoft/LightGBM/blob/8a5ec366aa5b0ec6301cb6ca8e85a70cc7ac4cf1/include/LightGBM/config.h#L404

StrikerRUS:
@mathcass 
Do you have time to take this job?

mathcass:
@StrikerRUS, sorry, had a busy week. 

I can give this a shot in the coming week. Unless there's some urgency to get it done sooner.

StrikerRUS:
@mathcass Yeah, sure! It's not urgent at all.

mathcass:
Thank you for the pointers on making this change @StrikerRUS! I've pushed them up to this PR. I can confirm that building the app works and running the CLI tool on each of the examples works. Changing a tree-based parameter in any of the `train.conf` files within the examples seems to work. 

```
cd examples/binary_classification
sed -i 's/num_trees = 100/n_estimators = 47/' train.conf
../../lightgbm config=train.conf

[LightGBM] [Info] Finished loading parameters
[LightGBM] [Info] Loading weights...

...

[LightGBM] [Info] Iteration:47, valid_1 binary_logloss : 0.508427
[LightGBM] [Info] 0.328891 seconds elapsed, finished iteration 47
[LightGBM] [Info] Finished training
```

StrikerRUS:
@mathcass Many thanks for your contribution!

Issue #994, opened 2017-10-15
Categorical features use causes crash returning from train function


tomcwalker:
Hi,

I am using the Python version of LGBM on windows 7 ultimate.

I will try to get a reproduction for this, but essentially, I find that if I use the categorical_feature override to the train function (specifying the columns as a list of strings), then the model trains fine, and  prints out:
Early stopping, best iteration is:
[97]	valid_0's l1: 0.047323
But then dies with:
  File "C:/Projects/Zillow/ZillowHandStacker/SearchTreeParams.py", line 468, in search_lgbm_zillow
    model = lgbm.train(train_set=d_train, params=lgbm_params, categorical_feature = cat_cols_for_lgbm, num_boost_round=2000, valid_sets=d_cv, early_stopping_rounds=50)
  File "C:\Users\Tom\Anaconda3\lib\site-packages\lightgbm\engine.py", line 223, in train
    booster._load_model_from_string(booster._save_model_to_string())
  File "C:\Users\Tom\Anaconda3\lib\site-packages\lightgbm\basic.py", line 1608, in _save_model_to_string
    ptr_string_buffer))
OSError: exception: access violation writing 0x000000000AC71000

Lowering the number of booster rounds helps somewhat, but it's still slow. I have also found that using categorical features makes the CV function blow up on memory.

What's odd with the above is it seems to be in the creation of the return data, as the actual fitting has already been done.

I have some fairly large categorical features - not sure exactly but a couple will be about 20-30k or so.

When I don't use this parameter all is fine, and in fact training seems better.

Perhaps some guidelines on when to use this feature/what it does under the hood might be useful if this is expected behaviour (though I guess access violation is never expected :) ) If these categories with very many distinct values are being one-hotted under the hood I can see why that would blow up memory.

guolinke:
I think the reason is the string model is too large. 
@wxchan can you confirm this ?

guolinke:
BTW, current master branch has a bug of categorical feature in python-package.
When use validation data, the validation result may is fake (it should be better) . 
refer to this PR: https://github.com/Microsoft/LightGBM/pull/993

guolinke:
@wxchan updates ?

guolinke:
@tomcwalker you can try the latest code.

PR #1080, opened 2017-11-24
fix save_model_to_string for large model

guolinke:


wxchan:
Is this the reason of #994 ?

guolinke:
@wxchan yeah

PR #1083, opened 2017-11-25
Speed up saving and loading model

guolinke:
1. reduce many overheads in saving and loading
2. multi-threading save and load
3. still use text format. Binary format will be added after this PR.

Both saving and loading are about 10x faster now.


StrikerRUS:
@guolinke **Protobuf Support** is still in `docs/Installation-Guide.rst`.

guolinke:
@StrikerRUS I just realize use max_bin in lgb.Dataset argument isn't a good idea.
This will break the sklearn users, they will not be able to set max_bin any more, since sklearn interfaces don't have lgb.dataset .

StrikerRUS:
@guolinke we've scheduled removing it in 2.0.12 version.

guolinke:
@StrikerRUS will we remove it from argument ?

StrikerRUS:
@guolinke Yeah, leave only in params.
https://github.com/Microsoft/LightGBM/blob/750a11b08cba9adcd05e6bcb92d30495783572da/python-package/lightgbm/basic.py#L645-L648

guolinke:
@StrikerRUS 
I am confused by the warning of travis outputs. 
Why the max_bin warning are always output without set or not set it ?

StrikerRUS:
@guolinke Your last commit successfuly removed the warning about `max_bin` from `test_basic.py`.

guolinke:
@StrikerRUS our next version is v2.0.12, so I will remove the max_bin argument directly. 
BTW, it seems v2.0.11 will always print many warnings.

StrikerRUS:
>BTW, it seems v2.0.11 will always print many warnings

Only in case of `max_bin` is not `None`. So users will be ready to pass `max_bin` through `params`.

guolinke:
@wxchan @StrikerRUS 
It seems there are some warnings about categorical features, when use validation data. I think they also can be removed.

> Only in case of max_bin is not None. So users will be ready to pass max_bin through params.

If user uses sklearn interface, the warning will always output. When using sklearn interface, max_bin are set both in param dict and argument.

StrikerRUS:
@guolinke 
> When using sklearn interface, max_bin are set both in param dict and argument.

Yeah. I think it was OK, because users now should pass `max_bin` via `params` even in sklearn.

wxchan:
@guolinke I think there is some bug about warning of categorical features, you can remove it.

StrikerRUS:
@guolinke please remove `max_bin` argument from sklearn's `LGBMModel` too.

guolinke:
@wxchan I am not very sure about these warning. I think it is better for you to fix it.

PR #1009, opened 2017-10-22
[docs] Update Installation-Guide.rst for macOS

thierrygosselin:
*  "brew-less" detailed instructions

PR #977, opened 2017-10-09
fix osx test

wxchan:

PR #1088, opened 2017-11-29
[ci] [docs] compilation on macOS

### LINKED ISSUE #782: Update macOS install instructions
This regards https://github.com/Microsoft/LightGBM/wiki/Installation-Guide

I cannot edit the file, so creating an issue.

Just want to note that the normal `brew install gcc` worked for me. This removes about two hours from the installation process for macOS as currently documented in this file.

StrikerRUS:
Fixed #782.
Refer to #977, #1009, https://github.com/dmlc/xgboost/pull/2784.


guolinke:
@StrikerRUS do we need to update the installation docs for OSX  ?

StrikerRUS:
@guolinke What do you mean?

I've already removed obsolete option `--without-multilib` and excess `brew install gcc` after `brew install open-mpi` (gcc is a dependence of open-mpi and are installed automatically) in this PR.

guolinke:
@StrikerRUS for example: 
Does ```# export CXX=g++-7 CC=gcc-7  # for macOS users only``` work now ?

StrikerRUS:
@guolinke Yeah, it works correctly:
https://github.com/Microsoft/LightGBM/blob/d51f2b4fb78d533bbb00c13d1d051b9b9d6efb62/.travis/test.sh#L17-L20

It could be simplified to `CXX=g++ CC=gcc` because they are aliases now according to [this](http://formulae.brew.sh/repos/Homebrew/homebrew-core/formula/gcc). Or if user has g++-5 this command should be replaced with `CXX=g++@5 CC=gcc@5` (I'm not sure, better ask macOS users). However, all our guides is correct but maybe not comprehensive.

fulldecent:
Looks great, thank you.

PR #1095, opened 2017-12-02
[python] add try/except in __del__

### LINKED ISSUE #1093: [Python] AttributeError: 'Booster' object has no attribute 'network'
## Environment info
Operating System: Windows Server 2012 R2 Standard
CPU: Intel Xeon CPU X5650 @ 2.67GHZ
C++/Python/R version: Python 3.6, pip installed LightGBM 2.0.11

## Error Message:
```
Exception ignored in: <bound method Booster.__del__ of <lightgbm.basic.Booster object at 0x0000004A00609BE0>>
Traceback <most recent call last>:
  File "C:\Continuum\Anaconda3\lib\site-package\lightgbm\basic.py", line 1351, in __del__
    if self.network:
AttributeError: 'Booster' object has no attribute 'network'
```
The code seems running through but throws this exception at the very end. There is no traceback to the code so not sure where it gets wrong. There is no exception when running the code on other Windows PC (Windows 7).

The code is only using LightGBM for prediction with a model read from a Pickle file:

```Python
def predict(model, data):
    pred_prob = pd.Series(model.predict_proba(data)[:, 1], name='pred_prob')
    pred = pd.Series(model.predict(data), name='pred')
    LOGGER.info("Prediction Success")
    LOGGER.info("{:d} True Out Of {:d} Total ({:.2f}%)".format(
        int(pred.sum()), len(pred), pred.sum()/len(pred)*100))
    return pred_prob, pred
```

And later use ```Lime``` to plot an Explainer:
```Python
def ......
            cat_features = [x_cols.index(i) for i in cat_cols]
            predict_fn = lambda x: model.predict_proba(x).astype(float)
            explainer = lime.lime_tabular.LimeTabularExplainer(
                np.array(data),
                feature_names=x_cols,
                class_names=['False', 'True'],
                categorical_features=cat_features,
                categorical_names=cat_cols)
            neg_samples = np.random.choice(np.reshape(np.asarray(np.where(
                pred_prob < 0.2)), -1), 5, False)
            pos_samples = np.random.choice(np.reshape(np.asarray(np.where(
                pred_prob > 0.8)), -1), 5, False)
```

Please help to figure out where causes this exception.

Thank you!

wxchan:
fix #1093 

I tested some cases, should work fine.

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -346,8 +346,11 @@ class _InnerPredictor(object):
         self.pred_parameter = param_dict_to_str(pred_parameter)
 
     def __del__(self):
-        if self.__is_manage_handle:
-            _safe_call(_LIB.LGBM_BoosterFree(self.handle))
+        try:
+            if self.__is_manage_handle:
+                _safe_call(_LIB.LGBM_BoosterFree(self.handle))
+        except:
+            pass
 
     def __getstate__(self):
         this = self.__dict__.copy()
@@ -609,7 +612,10 @@ class Dataset(object):
         self.params_back_up = None
 
     def __del__(self):
-        self._free_handle()
+        try:
+            self._free_handle()
+        except:
+            pass
 
     def _free_handle(self):
         if self.handle is not None:
@@ -1339,10 +1345,16 @@ class Booster(object):
             raise TypeError('Need at least one training dataset or model file to create booster instance')
 
     def __del__(self):
-        if self.network:
-            self.free_network()
-        if self.handle is not None:
-            _safe_call(_LIB.LGBM_BoosterFree(self.handle))
+        try:
+            if self.network:
+                self.free_network()
+        except:
+            pass
+        try:
+            if self.handle is not None:
+                _safe_call(_LIB.LGBM_BoosterFree(self.handle))
+        except:
+            pass
 
     def __copy__(self):
         return self.__deepcopy__(None)
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -349,7 +349,7 @@ class _InnerPredictor(object):
         try:
             if self.__is_manage_handle:
                 _safe_call(_LIB.LGBM_BoosterFree(self.handle))
-        except:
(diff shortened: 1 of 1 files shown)


wxchan:
@StrikerRUS I am not sure either.

StrikerRUS:
Maybe @guolinke can help to assess the level of danger?

guolinke:
@wxchan @StrikerRUS I think it is okay. 
However, I think it is better to print the exception message

wxchan:
@guolinke I think all error msg will be something like:
```
Exception ignored in: <bound method Booster.__del__ of <lightgbm.basic.Booster object at >>
AttributeError: 'NoneType' object has no attribute 'network'
```
which are confusing and meaningless.

guolinke:
@wxchan can we except that exception, and print others ?

ClimbsRocks:
For what it's worth, printing exceptions makes logs really difficult to parse. If it's an ignored exception, and you really want the user to know, I think that `warnings` are more appropriate than printing a bunch of lines that look an awful lot like exceptions when grepping through logs.

wxchan:
seems gpu tests fail

```
Saving to: ‘AMD-SDK.tar.bz2’

    [ <=>                                   ] 44,134      --.-K/s   in 0.03s   

2017-12-09 08:26:41 (1.57 MB/s) - ‘AMD-SDK.tar.bz2’ saved [44134]

bzip2: (stdin) is not a bzip2 file.
tar: Child returned status 2
tar: Error is not recoverable: exiting now
sh: 0: Can't open AMD-APP-SDK*.sh
chmod: cannot access ‘/home/travis/AMDAPPSDK/bin/x86_64/clinfo’: No such file or directory
.travis/test.sh: line 11: /home/travis/AMDAPPSDK/bin/x86_64/clinfo: No such file or directory
```

wxchan:
@huanzhang12 the gpu tests on master is somehow broken, can you help fix it?

guolinke:
@huanzhang12 you can create another PR to fix GPU test issue. Thanks !

huanzhang12:
@guolinke fixed in PR #1113

PR #1099, opened 2017-12-04
fix bug in feature fraction

guolinke:


Laurae2:
@guolinke I think AppVeyor stalled (still waiting for 4 hours already).

PR #1114, opened 2017-12-13
[python] added class_weight parameter in sklearn wrapper

### LINKED ISSUE #1107: set class_weight for LightGBM in sklearn
Is there a way to set class weight for multi-class classification task when using LightGBM with sklearn? From reading this past issue https://github.com/Microsoft/LightGBM/issues/434, it seems like a weight file can be used to set weight per each data. I'm using LightGBM as one stage of a complex sklearn pipeline, the input data has been transformed multiple times before fed into LightGBM, thus, it is not straightforward to make LightGBM consume a weight file. Is there any way to specify class_weight as a parameter to LightGBM in a similar fashion as sklearn's logistic regression class_weight parameter?

StrikerRUS:
Fixed #1107.

Should we somehow take into account `scale_pos_weight` and `is_unbalance` parameters?

### DIFF
--- a/python-package/lightgbm/compat.py
+++ b/python-package/lightgbm/compat.py
@@ -63,6 +63,7 @@ try:
     from sklearn.base import BaseEstimator
     from sklearn.base import RegressorMixin, ClassifierMixin
     from sklearn.preprocessing import LabelEncoder
+    from sklearn.utils.class_weight import compute_sample_weight
     from sklearn.utils.multiclass import check_classification_targets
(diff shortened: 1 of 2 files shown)


guolinke:
it looks good to me. ping @wxchan  for review.

guolinke:
BTW, I think the mulit-class should only use the ```class_wight```. 
And `scale_pos_weight` and `is_unbalance` is only used in binary classification. 
We can address this in document.

guolinke:
@StrikerRUS please merge if this is ready.

PR #1141, opened 2017-12-25
[R-package] Singlethread make to lower RAM usage significantly

### LINKED ISSUE #1125: Issue with "make -j" in R installation
Hi,

I ran into some odd behavior installing LightGBM from rstudio, using a command like:
```
library(devtools)
options(devtools.install.args = "--no-multiarch") 
install_github("Microsoft/LightGBM", subdir = "R-package")
```

My machine would become nonresponsive for many minutes, apparently crashing. I cloned the LightGBM repo and ran
```
R CMD INSTALL --build . --no-multiarch
```
With apparently the same affect. What fixed the issue for me was removing the "-j" from the make command, e.g. by editing install.libs.R. Evidently a few people have run into a similar problem, e.g.:
https://github.com/Microsoft/LightGBM/issues/605
https://github.com/Microsoft/LightGBM/issues/373

Would it make sense to consider removing the -j by default from the R installer script? Or, perhaps adding an argument to -j such as -j4? Alternatively, is there some option one can pass to the installer to disable parallel make?

Thank you for the consideration.

## Environment info
Operating System: Ubuntu 16.04 x86_64
CPU: VMWare VM with 1 processor clocked at 3.3Ghz from guest

C++/Python/R version:
g++ (Ubuntu 6.2.0-5ubuntu12) 6.2.0 20161005
Python 2.7.12+
R version 3.4.3 (2017-11-30) -- "Kite-Eating Tree"

## Error Message:
No error message appeared, my virtual machine became unresponsive.

## Reproducible examples
The command I ran was:
```
library(devtools)
options(devtools.install.args = "--no-multiarch") 
install_github("Microsoft/LightGBM", subdir = "R-package")
```
But I think the behavior depends on the specific machine, so this might be hard to reproduce.

Laurae2:
Fixes #1125.

To merge when tested (currently being tested) - will drop a msg in this PR when it's tested.

### DIFF
--- a/R-package/src/install.libs.R
+++ b/R-package/src/install.libs.R
@@ -41,7 +41,7 @@ if (!use_precompile) {
   
   # Prepare installation steps
   cmake_cmd <- "cmake "
-  build_cmd <- "make _lightgbm -j1"
+  build_cmd <- "make _lightgbm"
   lib_folder <- file.path(R_PACKAGE_SOURCE, "src", fsep = "/")
   
   if (use_gpu) {
@@ -52,7 +52,7 @@ if (!use_precompile) {
   if (WINDOWS) {
     if (use_mingw) {
       cmake_cmd <- paste0(cmake_cmd, " -G \"MinGW Makefiles\" ")
-      build_cmd <- "mingw32-make.exe _lightgbm -j1"
+      build_cmd <- "mingw32-make.exe _lightgbm"
       system(paste0(cmake_cmd, " ..")) # Must build twice for Windows due sh.exe in Rtools
     } else {
       try_vs <- 0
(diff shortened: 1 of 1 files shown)


Laurae2:
@wsnavely @guolinke Can you test the MinGW compilation and report if it uses only 1 thread? I have VS 2017 installed on all my computers unfortunately.

Use this code to install from my fork:

```r
devtools::install_github("Laurae2/LightGBM@fix-1125", subdir = "R-package")
```

guolinke:
`-j1` may is too small ?  it seems `-j4` is widely-used.

Laurae2:
@guolinke -j4 apparently crashes my AMD E-350 laptop (embedded 64-bit device dual core 1 GHz, 3 GB RAM). LightGBM uses too much RAM (the VS compiler uses only 1 core)

guolinke:
okay, however, you can you ```make``` directly, without `-j1`.

PR #1145, opened 2017-12-26
Fixed a typo in GPU installation command

sbtx:


StrikerRUS:
@sbtx Thank you for your PR! But this fix is already included into #1034.

Issue #979, opened 2017-10-09
Maybe a bug about Hessian matrix of regression_l1 objective function


WXFMAV:
## Maybe a bug  about Hessian matrix of regression_l1 objective function

There may be a bug in Hessian matrix of regression_l1 objective function.

It was strange and unreasonable in function **ApproximateHessianWithGaussian** that the gaussian variance is set as 
**const double c = std::max((std::fabs(y) + std::fabs(t)) * eta, 1.0e-10);** 

The proper code maybe as 
**const double c = std::max((std::fabs(y-t)) * eta, 1.0e-10);** 

Though I am not very sure if this is a bug, my program fails in weighted sampling as is reported in issue https://github.com/Microsoft/LightGBM/issues/939

**reference:**
The Code in File: https://github.com/Microsoft/LightGBM/blob/master/src/objective/regression_objective.hpp
between line 98 and line 123
```
  void GetGradients(const double* score, score_t* gradients,
                    score_t* hessians) const override {
    if (weights_ == nullptr) {
      #pragma omp parallel for schedule(static)
      for (data_size_t i = 0; i < num_data_; ++i) {
        const double diff = score[i] - label_[i];
        if (diff >= 0.0f) {
          gradients[i] = 1.0f;
        } else {
          gradients[i] = -1.0f;
        }
        hessians[i] = static_cast<score_t>(Common::ApproximateHessianWithGaussian(score[i], label_[i], gradients[i], eta_));
      }
    } else {
      #pragma omp parallel for schedule(static)
      for (data_size_t i = 0; i < num_data_; ++i) {
        const double diff = score[i] - label_[i];
        if (diff >= 0.0f) {
          gradients[i] = weights_[i];
        } else {
          gradients[i] = -weights_[i];
        }
        hessians[i] = static_cast<score_t>(Common::ApproximateHessianWithGaussian(score[i], label_[i], gradients[i], eta_, weights_[i]));
      }
    }
  }

```

And the code in File:
https://github.com/Microsoft/LightGBM/blob/87fa8b542e0027e03027197ba07aab2812d6755d/include/LightGBM/utils/common.h
between line 486 and line 495
```
inline static double ApproximateHessianWithGaussian(const double y, const double t, const double g,
                                                    const double eta, const double w=1.0f) {
  const double diff = y - t;
  const double pi = 4.0 * std::atan(1.0);
  const double x = std::fabs(diff);
[...]

PR #1150, opened 2017-12-28
add SMAPE objective function

### LINKED ISSUE #624: can add a mape objective function in lightgbm?
Such as the title.
reference to：https://en.wikipedia.org/wiki/Mean_absolute_percentage_error

could show the mape objective function code segment in python now.

lakshayg:
closes #624

### DIFF
--- a/src/metric/regression_metric.hpp
+++ b/src/metric/regression_metric.hpp
@@ -171,6 +171,19 @@ public:
   }
 };
 
+/*! \brief SMAPE loss for regression task */
+class SMAPEMetric: public RegressionMetric<SMAPEMetric> {
+public:
+  explicit SMAPEMetric(const MetricConfig& config) :RegressionMetric<SMAPEMetric>(config) {}
+
+  inline static double LossOnPoint(label_t label, double score, const MetricConfig&) {
+    return std::fabs(score - label) / (std::fabs(label) + std::fabs(score));
+  }
+  inline static const char* Name() {
+    return "smape";
+  }
+};
+
 /*! \brief Huber loss for regression task */
 class HuberLossMetric: public RegressionMetric<HuberLossMetric> {
 public:
--- a/src/objective/regression_objective.hpp
+++ b/src/objective/regression_objective.hpp
@@ -162,6 +162,73 @@ private:
 };
 
 /*!
+* \brief SMAPE regression loss
+*/
+class RegressionSMAPEloss: public RegressionL2loss {
+public:
+  explicit RegressionSMAPEloss(const ObjectiveConfig& config): RegressionL2loss(config) { }
+
+  explicit RegressionSMAPEloss(const std::vector<std::string>& strs): RegressionL2loss(strs) { }
+
+  ~RegressionSMAPEloss() {}
+
+  void GetGradients(const double* score, score_t* gradients,
+                    score_t* hessians) const override {
+    if (weights_ == nullptr) {
+      #pragma omp parallel for schedule(static)
+      for (data_size_t i = 0; i < num_data_; ++i) {
+        const double diff = score[i] - label_[i];
+        const double sumabs = std::fabs(score[i]) + std::fabs(label_[i]);
+        if (diff == 0) {
+          gradients[i] = 0.0f;
+          hessians[i] = 2.0f / sumabs;
+        } else {
+          const double absdiff = std::fabs(diff);
+          const double sumabs2 = sumabs * sumabs;
+          const double sumabs3 = sumabs2 * sumabs;
+          const double t1 = (2.0f / sumabs3 - (2.0f * (score[i] == 0)) / (label_[i] * label_[i]));
+          const double t2 = 2.0f  * sgn(label_[i]) * sgn(diff) / sumabs2;
+          gradients[i] = (sgn(diff) * sumabs - sgn(score[i]) * absdiff) / sumabs2;
+          hessians[i] = absdiff * t1 - t2;
+        }
+      }
+    } else {
+      #pragma omp parallel for schedule(static)
+      for (data_size_t i = 0; i < num_data_; ++i) {
+        const double diff = score[i] - label_[i];
+        const double sumabs = std::fabs(score[i]) + std::fabs(label_[i]);
+        if (diff == 0) {
+          gradients[i] = 0.0f;
+          hessians[i] = 2.0f * weights_[i] / sumabs;
+        } else {
+          const double absdiff = std::fabs(diff);
+          const double sumabs2 = sumabs * sumabs;
+          const double sumabs3 = sumabs2 * sumabs;
+          const double t1 = (2.0f / sumabs3 - (2.0f * (score[i] == 0)) / (label_[i] * label_[i]));
+          const double t2 = 2.0f  * sgn(label_[i]) * sgn(diff) / sumabs2;
+          gradients[i] = ((sgn(diff) * sumabs - sgn(score[i]) * absdiff)) * weights_[i] / sumabs2;
+          hessians[i] = (absdiff * t1 - t2) * weights_[i];
+        }
+      }
+    }
+  }
+
+  const char* GetName() const override {
+    return "smape";
+  }
+
+  bool IsConstantHessian() const override {
+    return false;
+  }
+
+private:
+  template <typename T>
+  int sgn(T x) const {
+    return (x > T(0)) - (x < T(0));
+  }
+};
+
+/*!
 * \brief Huber regression loss
 */
 class RegressionHuberLoss: public RegressionL2loss {
--- a/src/metric/metric.cpp
+++ b/src/metric/metric.cpp
@@ -15,6 +15,8 @@ Metric* Metric::CreateMetric(const std::string& type, const MetricConfig& config
     return new RMSEMetric(config);
   } else if (type == std::string("l1") || type == std::string("mean_absolute_error") || type == std::string("mae")) {
     return new L1Metric(config);
+  } else if (type == std::string("symmetric_mean_absolute_percentage_error") || type == std::string("smape")) {
+    return new SMAPEMetric(config);
   } else if (type == std::string("quantile")) {
     return new QuantileMetric(config);
   } else if (type == std::string("huber")) {
--- a/src/objective/objective_function.cpp
+++ b/src/objective/objective_function.cpp
@@ -13,6 +13,8 @@ ObjectiveFunction* ObjectiveFunction::CreateObjectiveFunction(const std::string&
     return new RegressionL2loss(config);
   } else if (type == std::string("regression_l1") || type == std::string("mean_absolute_error")  || type == std::string("mae")) {
     return new RegressionL1loss(config);
+  } else if (type == std::string("symmetric_mean_absolute_percentage_error") || type == std::string("smape")) {
+    return new RegressionSMAPEloss(config);
   } else if (type == std::string("quantile")) {
     return new RegressionQuantileloss(config);
   } else if (type == std::string("quantile_l2")) {
@@ -46,6 +48,8 @@ ObjectiveFunction* ObjectiveFunction::CreateObjectiveFunction(const std::string&
     return new RegressionL2loss(strs);
   } else if (type == std::string("regression_l1")) {
     return new RegressionL1loss(strs);
+  } else if (type == std::string("smape")) {
+    return new RegressionSMAPEloss(strs);
   } else if (type == std::string("quantile")) {
     return new RegressionQuantileloss(strs);
   } else if (type == std::string("quantile_l2")) {


src/metric/regression_metric.hpp
```diff
@@ -171,6 +171,19 @@ class L1Metric: public RegressionMetric<L1Metric> {
   }
 };
 
+/*! \brief MAPE loss for regression task */
+class MAPEMetric: public RegressionMetric<MAPEMetric> {
+public:
+  explicit MAPEMetric(const MetricConfig& config) :RegressionMetric<MAPEMetric>(config) {}
+
+  inline static double LossOnPoint(label_t label, double score, const MetricConfig&) {
+    return 100 * std::fabs(score - label) / label;
```
guolinke: label could be zero.

src/objective/regression_objective.hpp
```diff
@@ -161,6 +161,55 @@ class RegressionL1loss: public RegressionL2loss {
...
+    if (weights_ == nullptr) {
+      #pragma omp parallel for schedule(static)
+      for (data_size_t i = 0; i < num_data_; ++i) {
+        const double diff = score[i] - label_[i];
+        if (diff >= 0.0f) {
+          gradients[i] = 1.0f;
+        } else {
+          gradients[i] = -1.0f;
+        }
+        hessians[i] = static_cast<score_t>(0.0);
```
guolinke: cannot use zero hessians, since the output is grad/hess.

src/metric/regression_metric.hpp
lakshayg: What do you suggest in such a case?

guolinke: maybe, when label=0 and score=0, return 0, 
otherwise return score directly.

guolinke: @Laurae2 any suggestions  ?

lakshayg: Do you mean

```
if (label == 0 && score == 0) => return 0;
if (label == 0 && score != 0) => return score;
otherwise return (score - label) / label;
```

src/objective/regression_objective.hpp
lakshayg: What should be done in this case?

guolinke: you can use some approximate loss function. e.g. 1. https://en.wikipedia.org/wiki/Symmetric_mean_absolute_percentage_error . 
2. use L2 form: loss = (y-label)^2/label .
BTW, your gradient and hess actually is the L1, not the MAPE.

lakshayg: > BTW, your gradient and hess actually is the L1, not the MAPE.
Yes, I just realized that I messed up. Will update shortly

guolinke:
I think it is an interesting problem for the loss with hessian = 0.
for example, the L1, quantile, mape and huber. 
Are there some working solutions ? 
ping @Laurae2 @henry0312 .

lakshayg:
To prevent dealing with label == 0 or hessian == 0, I have implemented [SMAPE][1] instead of [MAPE][2].

[1]: https://en.wikipedia.org/wiki/Symmetric_mean_absolute_percentage_error
[2]: https://en.wikipedia.org/wiki/Mean_absolute_percentage_error

lakshayg:
The CI error is due to SSLError. Can one of the devs have a look, it seems unrelated to this change.

src/objective/regression_objective.hpp
```diff
@@ -161,6 +161,73 @@ class RegressionL1loss: public RegressionL2loss {
...
+        if (diff == 0) {
+          gradients[i] = 0.0f;
+          hessians[i] = 2.0f * weights_[i] / sumabs;
+        } else {
+          const double absdiff = std::fabs(diff);
+          const double sumabs2 = sumabs * sumabs;
+          const double sumabs3 = sumabs2 * sumabs;
+          const double t1 = (2.0f / sumabs3 - (2.0f * (score[i] == 0)) / (label_[i] * label_[i]));
+          const double t2 = 2.0f  * sgn(label_[i]) * sgn(diff) / sumabs2;
+          gradients[i] = ((sgn(diff) * sumabs - sgn(score[i]) * absdiff)) * weights_[i] / sumabs2;
```
guolinke: you can use `std::copysign` .

lakshayg:
```
sgn(a) * b != std::copysign(b, a) // when a = 0
```

Should I still replace sgn with copysign? Perhaps there is a better place to keep copysign so that it can be used in other places as well. A number of places in code can make use of this function rather than having branches.

guolinke: copysign is more efficient, it will avoid the if...else check.

lakshayg: I meant that the function I wrote relies on the fact that `sgn(0) = 0`. This is not the case with copysign. I am also unsure about how copysign will avoid if-else. Can you please elaborate?

guolinke: you can see the compiled Assemble code, copysign will use the cpu instructions to get sign directly

guolinke: i think copysign is safe for 0 . you can test it

lakshayg:
```
  std::cout << std::copysign(1, 0) << std::endl;
```

The above statement print 1 and not 0

lakshayg: As far as branching in `sgn` is concerned, the version I wrote does not have any branches https://godbolt.org/g/AUnpfi

Laurae2:
@guolinke @lakshayg Gradient and Hessian of SMAPE:

```r
grad <- function(x, y) {
  (2 * y * (x - y)) / ((x + y) * (x + y) * abs(x - y))
}

hess <- function(x, y) {
  (-4 * y * (x - y)) / ((x + y) * (x + y) * (x + y) * abs(x - y))
}
```

Tested and working in [Kaggle Web Traffic Competition](https://www.kaggle.com/c/web-traffic-time-series-forecasting/discussion), identical results to the symbolic R math function conversion.

Below is the symbolic math function conversion from R:

```r
grad <- function(x, y) {
  .e2 <- abs(x) + y
  .e3 <- x - y
  (sign(.e3) - abs(.e3) * sign(x)/.e2)/.e2
}

hess <- function(x, y) {
  .e2 <- abs(x) + y
  .e3 <- sign(x)
  .e4 <- x - y
  -(2 * ((sign(.e4) - abs(.e4) * .e3/.e2) * .e3/.e2^2))
}
```

To guarantee a converging Gradient/Hessian of MAPE (because of hessian = 0), it requires to approximate a function which is both similar to the MAPE (linearly) and a quadratic version of itself (such as how Fair loss, etc. are done for L1/MAE loss). Otherwise, there is no guarantee MAPE can converge (it will explode in most of cases). SMAPE usually is not an alternative as MAPE has the very strong intrinsic feature of underpredicting.

When label = 0, loss should be considered 0 otherwise it explodes to infinity. When loss is 0, the associated gradients and hessians should be also 0.

lakshayg:
@Laurae2 Thanks a lot for the detailed comment. Here is what I think should be the way to go then:

1) Update the SMAPE grad and hessian with what you suggested
2) This PR should not  close the issue since you said that SMAPE is not an alternative to MAPE (could you suggest an alternative?)
3) Close the issue once an alternative to MAPE is implemented (implementing MAPE itself is probably not a good idea since you say that it mostly explodes)

What are your views?

src/objective/regression_objective.hpp
guolinke: okay, great

lakshayg:
@Laurae2 Could you clarify what `x` and `y` mean? Is `x = prediction` and `y = label`?

Laurae2:
@lakshayg It must be `x = prediction` and `y = label`.

Test:

```r
library(xgboost)

data(agaricus.train, package='xgboost')
data(agaricus.test, package='xgboost')

dtrain <- xgb.DMatrix(agaricus.train$data, label = agaricus.train$label + 1)
dtest <- xgb.DMatrix(agaricus.test$data, label = agaricus.test$label + 1)
watchlist <- list(train = dtrain, eval = dtest)

param <- list(max_depth = 2, eta = 1, silent = 1, nthread = 2, 
              objective = "binary:logistic", eval_metric = "auc")
bst <- xgb.train(param, dtrain, nrounds = 2, watchlist)

grad <- function(x, y) {
  (2 * y * (x - y)) / ((x + y) * (x + y) * abs(x - y))
}

hess <- function(x, y) {
  (-4 * y * (x - y)) / ((x + y) * (x + y) * (x + y) * abs(x - y))
}

smape_obj <- function(preds, dtrain) {
  labels <- getinfo(dtrain, "label")
  grad <- grad(labels, preds)
  hess <- hess(labels, preds)
  return(list(grad = grad, hess = hess))
}

smape_loss <- function(preds, dtrain) {
  labels <- getinfo(dtrain, "label")
  err <- mean((abs(preds - labels)) / ((abs(preds) + abs(labels)) / 2))
  return(list(metric = "smape", value = err))
}

# These functions could be used by passing them either:
#  as 'objective' and 'eval_metric' parameters in the params list:
param <- list(max_depth = 2, eta = 0.1, nthread = 1, 
              objective = smape_obj, eval_metric = smape_loss)
bst <- xgb.train(param, dtrain, nrounds = 10, watchlist)
```

Yields convergence:

```r
[1]	train-smape:0.786086	eval-smape:0.786127 
[2]	train-smape:0.654798	eval-smape:0.655078 
[3]	train-smape:0.529639	eval-smape:0.530132 
[4]	train-smape:0.410331	eval-smape:0.411013 
[5]	train-smape:0.296609	eval-smape:0.297458 
[6]	train-smape:0.216386	eval-smape:0.216352 
[7]	train-smape:0.237770	eval-smape:0.237861 
[8]	train-smape:0.254103	eval-smape:0.254304 
[9]	train-smape:0.266287	eval-smape:0.266585 
[10]	train-smape:0.298564	eval-smape:0.298592 
```

Inverting `preds` and `labels` in `smape_obj` won't converge:

```r
[1]	train-smape:0.923793	eval-smape:0.923567 
[2]	train-smape:0.923793	eval-smape:0.923567 
[3]	train-smape:0.923793	eval-smape:0.923567 
[4]	train-smape:0.923793	eval-smape:0.923567 
[5]	train-smape:0.923793	eval-smape:0.923567 
[6]	train-smape:0.923793	eval-smape:0.923567 
[7]	train-smape:0.923793	eval-smape:0.923567 
[8]	train-smape:0.923793	eval-smape:0.923567 
[9]	train-smape:0.923793	eval-smape:0.923567 
[10]	train-smape:0.923793	eval-smape:0.923567
```

(due to the labels, using SMAPE is way too much inappropriate but this is just for the example - using a simple quadratic loss yields 0.01 SMAPE nearly instantly)

lakshayg:
@Laurae2 Can you explain how you came up with this derivative? I tried using wolfram alpha but could not reduce its output to the gradient that you specified. Are there any assumptions that I might be missing?

http://www.wolframalpha.com/input/?i=differentiate+2*+%7Cx-y%7C+%2F+(%7Cx%7C+%2B+%7Cy%7C)

Laurae2:
@lakshayg use this: https://www.derivative-calculator.net/

Input this formula: `(abs(x - y)) / ((abs(x) + abs(y)) / 2)`

Request the second derivative:

image

Interesting plot explaining why it's very hard for SMAPE with hessian (change it to 2 and you will see why):

image

guolinke:
I think we can continue the discussion here: https://github.com/Microsoft/LightGBM/issues/979

src/metric/regression_metric.hpp
```diff
@@ -171,6 +171,19 @@ class L1Metric: public RegressionMetric<L1Metric> {
   }
 };
 
+/*! \brief SMAPE loss for regression task */
+class SMAPEMetric: public RegressionMetric<SMAPEMetric> {
+public:
+  explicit SMAPEMetric(const MetricConfig& config) :RegressionMetric<SMAPEMetric>(config) {}
+
+  inline static double LossOnPoint(label_t label, double score, const MetricConfig&) {
+    return std::fabs(score - label) / (std::fabs(label) + std::fabs(score));
```
guolinke: I think ```std::fabs(label) + std::fabs(score)``` also could be zero.

guolinke:
@Laurae2 Is grad/hess of current implementation okay ?

lakshayg:
I think std::fabs(label) + std::fabs(score) also could be zero.

Yes you are correct. I am working on implementing this again and will update the PR sometime this weekend.

src/metric/regression_metric.hpp
lakshayg: Yes you are correct. I am working on implementing this again and will update the PR sometime this weekend.

guolinke: great! thanks very much

Laurae2:
@lakshayg Which exact formula are you using for grad/hess?

I'd like to test them as custom losses to test in xgboost/LightGBM to check.

@guolinke It seems to be OK when looking at them, didn't check in depth though.

guolinke:
@lakshayg please update the document accordingly.

Laurae2:
@lakshayg any news?

@guolinke I think SMAPE should use the gradient only (since you managed to make hessian less LightGBM). With hessian, it is unstable.

lakshayg:
Currently busy with work, I guess I will be able to work this/next weekend. Sorry for the delay.

StrikerRUS:
@lakshayg Are you still interested in this?

lakshayg:
Sorry for the delay in responding. Yes, I'm still interested. Will give this another try this weekend.

On 31 May 2018 12:22 pm, Nikita Titov <notifications@github.com> wrote:

@lakshayg<https://github.com/lakshayg> Are you still interested in this?

lakshayg:
Sorry, I have not been able to find time to continue working on this. I will close the PR for now and will send in a different PR if/when I am able to get back to it.

PR #1152, opened 2017-12-28
[python] avoid to set all weight to 1.

guolinke:


guolinke:
@StrikerRUS using all weights with 1 is not efficient.

python-package/lightgbm/sklearn.py
```diff
@@ -452,11 +453,12 @@ def get_meta_data(collection, i):
                         else:
                             raise TypeError('eval_sample_weight, eval_class_weight, eval_init_score, and eval_group should be dict or list')
                     valid_weight = get_meta_data(eval_sample_weight, i)
-                    valid_class_sample_weight = _LGBMComputeSampleWeight(get_meta_data(eval_class_weight, i), valid_data[1])
-                    if valid_weight is None or len(valid_weight) == 0:
-                        valid_weight = valid_class_sample_weight
-                    else:
-                        valid_weight = np.multiply(valid_weight, valid_class_sample_weight)
+                    if self.class_weight is not None:
```
StrikerRUS: @guolinke I suppose you mean here:
`if get_meta_data(eval_class_weight, i) is not None: `

PR #1161, opened 2017-12-31
[ci] boost_root and opencl_include_dir as pip options

StrikerRUS:


StrikerRUS:
@guolinke Please take a look to Appveyor error - it's rather strange and hasn't been caused by this PR.

guolinke:
@StrikerRUS It seems that is bug of Appveyor. I just rebuild it.

PR #1162, opened 2018-01-01
[docs] Typo

kant:


StrikerRUS:
@kant Next time please make all similar changes within one PR.

PR #1164, opened 2018-01-01
[docs] Minor fix

kant:


StrikerRUS:
@kant You can fix table by deleting one space after the closing bracket
```
converted from float values)   | 
                             ^
```

PR #1169, opened 2018-01-01
added output_freq to aliases

StrikerRUS:
Added ``output_freq`` to aliases in docs and removed ``metric_freq`` from params set because it's an alias.
Refer to fe8e2162ff12ef4b0c51bf901a14fedfe1caf75a.


include/LightGBM/config.h
```diff
@@ -468,7 +468,7 @@ struct ParameterAlias {
       "pred_early_stop_margin", "use_missing", "sigmoid",
       "fair_c", "poission_max_delta_step", "scale_pos_weight",
       "boost_from_average", "max_position", "label_gain",
-      "metric", "metric_freq", "time_out",
+      "metric", "time_out",
```
guolinke: I think ```output_freq``` should be here

Issue #1185, opened 2018-01-08
Can we use GPU in python package? 


PayneJoe:
I've already install GPU versioned Lightgbm successfully, unfortunately not got a way to use it yet. Providing any examples about this is much appreciated!

StrikerRUS:
You need to set an additional parameter `device` : `gpu` (along with your other options like` learning_rate`, `num_leaves`, etc) to use GPU in Python. It will use the first GPU installed on the system by default (`gpu_platform_id=0 `and `gpu_device_id=0`). https://lightgbm.readthedocs.io/en/latest/Parameters.html#gpu-parameters

You may read the following (examples in R, but everything should be clear):
https://lightgbm.readthedocs.io/en/latest/GPU-Targets.html

StrikerRUS:
BTW, @huanzhang12 is it OK that GPU-Tutorial says
>It will use the first GPU installed on the system by default (gpu_platform_id=0 and gpu_device_id=0).

But actually default values are -1 (like it's said in Parameters):
>gpu_platform_id, default=-1, type=int
...
default value is -1, means the system-wide default platform

>gpu_device_id, default=-1, type=int
...
default value is -1, means the default device in the selected platform

?

huanzhang12:
@StrikerRUS Yes I think you are right. The language in documentation is inaccurate. I will fix it.

My understanding is that a id = 0 means the first platform/GPU. When setting the id to -1, it will use the default platform/GPU. The default/GPU does not have to be the first platform/GPU (with id = 0).

PayneJoe:
Thanks @StrikerRUS ! Cannot imagine a super parameter can get GPU running. I'll go and confirm it later on.

PR #1187, opened 2018-01-08
[docs] Update GPU-Tutorial.rst

huanzhang12:
Better description for GPU and platform IDs in documentation, as suggested by @StrikerRUS in #1185


StrikerRUS:
@huanzhang12 Thanks!

I think this's valuable information and it could be duplicated [here ](https://github.com/Microsoft/LightGBM/blob/master/docs/GPU-Windows.rst#debugging-lightgbm-crashes-in-cli) and somewhere [here](https://github.com/Microsoft/LightGBM/blob/master/docs/GPU-Targets.rst).

huanzhang12:
@StrikerRUS I also updated the other two files as you suggested. Will merge after checks are passed.

PR #1210, opened 2018-01-22
try to fix problem with multi-dimensional sliced object.

### LINKED ISSUE #1205: Incorrect training when label array comes from slicing multi-dimensional numpy array
See below for script containing example illustrating the issue. It appears that when the array of labels passed to a Dataset comes from slicing a multi-dimensional array lightgbm does not use the correct values in the array but access values from the presliced array, my guess would be that it is not correctly taking into account the strides parameter of the array.

Let me know if anything is needed.

## Environment info
Operating System: Ubuntu
CPU: Intel
Python version: 2.7.6
Lightgbm version: 2.0.5

## Reproducible examples
See below for script that should be able to reproduce issue

```
import lightgbm
import numpy

def train_and_get_predictions(features, labels):
    dataset = lightgbm.Dataset(features, label=labels)

    lgb_params = {
        'application': 'binary',
        'verbose': 0,
    }

    lgbm_model = lightgbm.train(
        params=lgb_params,
        train_set=dataset,
        num_boost_round=40,
    )

    predictions = lgbm_model.predict(features)
    return numpy.mean(predictions)

num_samples = 100
# Just give random features so model should converge at sample mean.
features = numpy.random.rand(num_samples, 5)

# Labels should be 25% ones and 75% zeros                                                                                                                              
# so dummy predictor should just give base score of 0.25.                                                                                                              
positive_samples = int(num_samples*0.25)
labels = numpy.append(
    numpy.ones(positive_samples, dtype=numpy.float32),
    numpy.zeros(num_samples - positive_samples, dtype=numpy.float32),
)
print("Prediction mean when using samples: {}".format(train_and_get_predictions(features, labels)))
print("Labels strides: {}".format(labels.strides))

# If we column stack the labels and then slice the predictions change.                                                                                                 
stacked_labels = numpy.column_stack((labels, numpy.ones(num_samples, dtype=numpy.float32)))
sliced_labels = stacked_labels[:, 0]
assert numpy.all(sliced_labels == labels)
print("Prediction mean when using sliced samples: {}".format(train_and_get_predictions(features, sliced_labels)))
print("Sliced labels strides: {}".format(sliced_labels.strides))

# If we make a copy of the sliced array everything works again.                                                                                                        
copied_sliced_labels = numpy.copy(sliced_labels)
assert train_and_get_predictions(features, copied_sliced_labels) == train_and_get_predictions(features, labels)
print("Prediction mean when using copied sliced samples: {}".format(train_and_get_predictions(features, copied_sliced_labels)))
print("Copied sliced labels strides: {}".format(copied_sliced_labels.strides))

# We can make an array that behaves like the sliced array suggesting the other columns from the                                                                        
# column_stack are being accessed.                                                                                                                                     
sliced_array_looks_like = numpy.append(numpy.tile([1, 1], num_samples/4), numpy.tile([0, 1], num_samples/4))
assert train_and_get_predictions(features, sliced_array_looks_like) == train_and_get_predictions(features, sliced_labels)
```

Output of this script when I run it locally
```
Prediction mean when using samples: 0.254352023845
Labels strides: (4,)
Prediction mean when using sliced samples: 0.745772696814
Sliced labels strides: (8,)
Prediction mean when using copied sliced samples: 0.254352023845
Copied sliced labels strides: (4,)
```

guolinke:
to fix #1205

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -181,9 +181,10 @@ FIELD_TYPE_MAPPER = {"label": C_API_DTYPE_FLOAT32,
 
 def convert_from_sliced_object(data):
     """fix the memory of multi-dimensional sliced object"""
-    if data.base is not None and isinstance(data.base, np.ndarray):
+    if data.base is not None and isinstance(data, np.ndarray) and isinstance(data.base, np.ndarray):
         base_shape = data.base.shape
         if base_shape != data.shape and len(base_shape) > 1 and base_shape[1] > 1:
+            warnings.warn("Use subset(sliced data) of np.ndarray is not recommended due to it will double the peak memory cost in LightGBM.")
             return np.copy(data)
     return data
 
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -182,8 +182,7 @@ FIELD_TYPE_MAPPER = {"label": C_API_DTYPE_FLOAT32,
 def convert_from_sliced_object(data):
     """fix the memory of multi-dimensional sliced object"""
     if data.base is not None and isinstance(data, np.ndarray) and isinstance(data.base, np.ndarray):
-        base_shape = data.base.shape
-        if base_shape != data.shape and len(base_shape) > 1 and base_shape[1] > 1:
+        if not data.flags.c_contiguous:
             warnings.warn("Use subset(sliced data) of np.ndarray is not recommended due to it will double the peak memory cost in LightGBM.")
             return np.copy(data)
     return data
@@ -195,6 +194,7 @@ def c_float_array(data):
         data = np.array(data, copy=False)
     if is_numpy_1d_array(data):
         data = convert_from_sliced_object(data)
+        assert data.flags.c_contiguous
         if data.dtype == np.float32:
             ptr_data = data.ctypes.data_as(ctypes.POINTER(ctypes.c_float))
             type_data = C_API_DTYPE_FLOAT32
@@ -215,6 +215,7 @@ def c_int_array(data):
         data = np.array(data, copy=False)
     if is_numpy_1d_array(data):
         data = convert_from_sliced_object(data)
+        assert data.flags.c_contiguous
         if data.dtype == np.int32:
             ptr_data = data.ctypes.data_as(ctypes.POINTER(ctypes.c_int32))
             type_data = C_API_DTYPE_INT32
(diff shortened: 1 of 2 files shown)


aldanor:
While this would work - it would involve copying all of the data (and then copying the data again after it's binned?..). 

Is there a way to avoid this maybe? E.g. respecting the buffer strides when creating the Dataset object.

guolinke:
@aldanor The copy is already a temporary buffer. The copy will be released after Dataset constructed.

aldanor:
@guolinke Right; it will be released, however it will peak at over 2x memory usage, which can be avoided, it's just a little bit more work. E.g., if your RAM is 64GB and the dataset is 40GB, you won't be able to make that temporary copy at all.

guolinke:
@aldanor Yeah, it is possible to reduce the peak memory, but it is not so straightforward:
1. sample some rows from sliced data and use them to construct Dataset by this function: https://github.com/Microsoft/LightGBM/blob/master/include/LightGBM/c_api.h#L70-L77
2. Push rows streamly to the dataset: https://github.com/Microsoft/LightGBM/blob/master/include/LightGBM/c_api.h#L121-L131 

However, I am very busy recently. So I don't have time for that implementation. 
Welcome to contribute if you are interested.

guolinke:
@StrikerRUS

python-package/lightgbm/basic.py
```diff
@@ -472,7 +485,7 @@ def __pred_for_np2d(self, mat, num_iteration, predict_type):
         else:
             """change non-float data to float data, need to copy"""
             data = np.array(mat.reshape(mat.size), dtype=np.float32)
-        ptr_data, type_ptr_data = c_float_array(data)
+        ptr_data, type_ptr_data, new_data = c_float_array(data)
```
StrikerRUS: @guolinke If we don't need `new_data` maybe it's better to name it like non-used variable:
```
ptr_data, type_ptr_data, _ = c_float_array(data) 
```
?

guolinke: yeah, `new_data` is used to avoid the temporary copy is freed.

python-package/lightgbm/basic.py
```diff
@@ -502,8 +515,8 @@ def __pred_for_csr(self, csr, num_iteration, predict_type):
         preds = np.zeros(n_preds, dtype=np.float64)
         out_num_preds = ctypes.c_int64(0)
 
-        ptr_indptr, type_ptr_indptr = c_int_array(csr.indptr)
-        ptr_data, type_ptr_data = c_float_array(csr.data)
+        ptr_indptr, type_ptr_indptr, new_indptr = c_int_array(csr.indptr)
```
StrikerRUS: The same remark about naming of `new_indptr`.

guolinke:
@StrikerRUS Yeah, I think this is enough for most cases.

StrikerRUS:
@guolinke Agree with you.

PR #1223, opened 2018-01-29
Adding SWIG Java wrappers to LightGBM

imatiach-msft:
Adding SWIG Java wrappers to LightGBM

If user installs SWIG and enables it in CMake file, they can now generate a jar which will contain the Java wrappers for the c_api.h file.


CMakeLists.txt
```diff
@@ -125,7 +145,17 @@ file(GLOB SOURCES
...
+add_library(_lightgbm SHARED src/c_api.cpp src/lightgbm_R.cpp  ${SOURCES})
+
+if(USE_SWIG)
+  set_property(SOURCE swig/lightgbmlib.i PROPERTY CPLUSPLUS ON)
+  swig_add_module(_lightgbm_swig java swig/lightgbmlib.i)
+  swig_link_libraries(_lightgbm_swig _lightgbm)
+  add_custom_command(TARGET _lightgbm_swig POST_BUILD
+    COMMAND "${Java_JAVAC_EXECUTABLE}" -d com/microsoft/ml/lightgbm java/*.java
+    COMMAND cp "${PROJECT_SOURCE_DIR}/*.so" com/microsoft/ml/lightgbm/linux/x86_64
+    COMMAND "${Java_JAR_EXECUTABLE}" -cfM lightgbmlib.jar -C com .)
```
imatiach-msft: is there any variable I can use to set the version here -- eg "lightgbmlib<VERSION>.jar"

CMakeLists.txt
```diff
@@ -136,6 +168,11 @@ if(USE_MPI)
   TARGET_LINK_LIBRARIES(_lightgbm ${MPI_CXX_LIBRARIES})
 endif(USE_MPI)
 
+if(USE_OPENMP)
```
guolinke: it seems we didn't need this before. why link to OpenMP_CXX_FLAGS ?

imatiach-msft: I added this when I was getting a seg fault near the code that was doing a parallel loop with OpenMP, but later on I found it was a different issue with my Java wrapper code.  It looks like it is not needed - let me verify and I will remove.

imatiach-msft: verified that it is not needed

PR #1229, opened 2018-02-04
Permit use of custom objective function for multiclass problems

jtilly:
Hi everyone,

This pull request permits the use of a custom objective function for multiclass problems. The only real obstacle that prevented me from doing this before is [this line](https://github.com/Microsoft/LightGBM/blob/master/src/io/config.cpp#L197), because custom objective functions will always have `objective_type_multiclass = false`. I changed that line and added a demo in R for how to implement a simple softmax problem with custom objective function and custom eval metric.

I'm not sure if there are any fundamental problems with my approach or if people just don't use custom objective functions for multiclass problems. If it is the former, feel free to disregard this pull request.

Best,
Jan

### DIFF
--- a/R-package/demo/multiclass_custom_objective.R
+++ b/R-package/demo/multiclass_custom_objective.R
@@ -1,0 +1,85 @@
+require(lightgbm)
+
+# We load the default iris dataset shipped with R
+data(iris)
+
+# We must convert factors to numeric
+# They must be starting from number 0 to use multiclass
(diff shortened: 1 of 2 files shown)


StrikerRUS:
ping @guolinke for clarifying aliases issue.

guolinke:
@StrikerRUS 
yeah, maybe we should constraint the name of "custom" objective function. 
I think using "none", "null" and "custom" is enough.

Issue #1230, opened 2018-02-05
Custom objective and evaluation functions


nimamox:
Hi folks,
The problem is that when I set `fobj` to my customized objective function, the prediction error I receive after training differs from what is reported by `feval` at the last iteration. The errors are identical, as expected, if I remove `fobj`.

```python
def my_logistic_obj(y_hat, dtrain):
    y = dtrain.get_label()
    p = y_hat 
    grad = 4 * p * y + p - 5 * y
    hess = (4 * y + 1) * (p * (1.0 - p))
    return grad, hess

def my_err_rate(y_hat, dtrain):
    y = dtrain.get_label()
    y_hat = np.clip(y_hat, 10e-7, 1-10e-7)
    loss_fn = y*np.log(y_hat)
    loss_fp = (1.0 - y)*np.log(1.0 - y_hat)
    return 'myloss', np.sum(-(5*loss_fn+loss_fp))/len(y), False

def calc_loss(y, yp): #same as my_err_rate
    yp = np.clip(yp, 10e-7, 1.0-10e-7)
    loss_fn = y*np.log(yp)
    loss_fp = (1.0-y)*np.log(1.0-yp)
    return np.sum(-(5*loss_fn+loss_fp))/y.shape[0]

params = {
    'task': 'train',
    'objective': 'regression',
    'boosting': 'gbdt',
    'metric': 'auc',
    'train_metric': '+',
    'num_leaves': 260,
    'learning_rate': 0.0245,
    'feature_fraction': 0.9,
    'bagging_fraction': 0.8,
    'bagging_freq': 5,
    'max_depth': 15,
    'max_bin': 512
}

categoricals = None
lgb_train = lgb.Dataset(X_trn, y_trn, feature_name=features, categorical_feature=categoricals, free_raw_data=True) 
lgb_eval = lgb.Dataset(X_val, y_val, feature_name=features, categorical_feature=categoricals, reference=lgb_train, free_raw_data=True)

evals_result = {}
gbm = lgb.train(params,
                lgb_train,
                num_boost_round = 5,
                valid_sets=[lgb_eval, lgb_train],
#                 early_stopping_rounds = 2,
#                 fobj = my_logistic_obj, # <~~ First without fobj
                feval = my_err_rate,
                evals_result=evals_result,
                verbose_eval=1
               )
```

This gives me,

```
...
[...]

PR #1231, opened 2018-02-06
Auto set `objective=None` when using fobj

guolinke:
refer to #1230


guolinke:
@StrikerRUS the check-docs fails. can you help to check it ?

StrikerRUS:
@guolinke Strange error. It seems that the site is OK:
image

I suppose the error is connected to the specific Python-package installed on Travis.

guolinke:
ping @Laurae2 to check R's code

PR #1243, opened 2018-02-15
Experimental support for HDFS

ebernhardson:
* Read and write datsets from hdfs.
* Only enabled when cmake is run with -DUSE_HDFS:BOOL=TRUE
* Introduces VirtualFile(Reader|Writer) to asbtract VFS differences


ebernhardson:
I've forwarded the CLA to legal, should be resolved soon.

include/LightGBM/bin.h
```diff
@@ -98,7 +98,7 @@ class BinMapper {
   * \brief Save binary data to file
   * \param file File want to write
   */
-  void SaveBinaryToFile(FILE* file) const;
+  void SaveBinaryToFile(std::shared_ptr<VirtualFileWriter> file) const;
```
guolinke: why use `share_ptr` ? we often use `const` raw pointers. what are the advantages of share_ptr ?

include/LightGBM/utils/file_io.h
```diff
@@ -0,0 +1,281 @@
...
+      }
+    }
+  }
+
+  virtual bool Init(void) {
+    if (file_ == NULL) {
+      if (fs_ == NULL) {
+        auto namenode = std::vector<char>(100);
+        if (1 != sscanf(filename_.c_str(), "hdfs://%99[^/]/", namenode.data())) {
+          Log::Warning("Could not parse namenode from uri [%s]", filename_.c_str());
```
ebernhardson: Log::Fatal, or at least return false

include/LightGBM/utils/file_io.h
```diff
@@ -0,0 +1,281 @@
...
+public:
+  LocalFileWriter(const char* filename) : VirtualFileWriter(filename), file_(NULL) {}
+  virtual ~LocalFileWriter() {
+    if (file_ != NULL) {
+      fclose(file_);
+    }
+  }
+
+  static inline bool Exists(const char* filename) {
+    FILE* file_ = fopen(filename, "rb");
```
ebernhardson: missing ifdef's for MSC

include/LightGBM/utils/file_io.h
```diff
@@ -0,0 +1,281 @@
...
+      return false;
+    } else {
+      bool exists = 0 == hdfsExists(fs, filename);
+      hdfsDisconnect(fs);
+      return exists;
+    }
+  }
+
+  virtual bool Init() {
+    if (file_ == NULL) {
```
ebernhardson: this init code is pretty much duplicated for Read/Write. Can it be shared?

include/LightGBM/bin.h
ebernhardson: It's only a personal preference, these objects are very low volume so the overhead should be minimal, and the shared_ptr made thinking about the lifetime much easier. If you prefer though i can re-work this so the base instance gets allocated on the stack and we pass const pointers. Stack allocation should keep the relatively simple lifetime.

guolinke: okay, got it. The pointer is actual not 'const' inside these functions.

include/LightGBM/utils/file_io.h
```diff
@@ -0,0 +1,33 @@
...
+
+namespace LightGBM{
+
+struct VirtualFileWriter {
+  VirtualFileWriter(const char* filename) : filename_(filename) {}
+  virtual ~VirtualFileWriter() {};
+
+  virtual bool Init()=0;
+  virtual size_t Write(const void* data, size_t bytes) const=0;
+  static std::shared_ptr<VirtualFileWriter> Make(const char* filename);
```
guolinke: unique_ptr ?

src/io/parser.cpp
```diff
@@ -70,12 +70,19 @@ enum DataType {
...
-  std::ifstream tmp_file;
-  tmp_file.open(filename);
-  if (!tmp_file.is_open()) {
-    Log::Fatal("Data file %s doesn't exist'", filename);
+  auto reader = VirtualFileReader::Make(filename);
+  if (!reader->Init()) {
+    Log::Fatal("Data file %s doesn't exist'", reader->filename_.c_str());
   }
   std::string line1, line2;
+  size_t buffer_size = 32*1024;
```
guolinke: I think this may is not enough for the long lines.

guolinke:
@ebernhardson sorry, we are in the Chinese New Year holiday, so the review is slow.
Did you test this ? 

@StrikerRUS it seems the check-the-docs test failed again.

StrikerRUS:
@guolinke Yeah, I see...

Sites with Yahoo dataset and MPICH paper are down. SWIG site returns this warning: `Project web is currently offline pending the final migration of its data to our new datacenter.`

I think that Yahoo and SWIG will be online in a few days and we should just wait. Not sure about site with paper, maybe it should be replaced with another site.

StrikerRUS:
@guolinke All sites are alive now, I've re-run Travis.

src/io/parser.cpp
ebernhardson: I've updated this so it should work for any length line possible, and added a test on the python package to assert it's working.

ebernhardson:
@guolinke  I've tested this against a hadoop cluster I have available, running CDH 5.5, on input files up to 100GB. Works for reading in text or binary datasets and writing out binary datasets. I haven't explicitly tested the other code paths like predict but see no reason it should vary as long as the local file tests also pass.

I've also added a second patch to this PR, but i could split it into a second if preferred. The second patch allows providing a comma separated list of filenames and have them treated a single file concatenated together. This is particularly important for hdfs as it is very typical (and highly performant when writing) to write out a single large file as lots of smaller files of perhaps 256M or 1GB each. Some python tests were added to show this working.

guolinke:
@ebernhardson thanks very much. 
I think use another PR for the mulit-file support is better.

guolinke:
@ebernhardson 
Do you have any speedtest for the IO time (dataset loading) after this PR ? 
you can use higgs(https://github.com/guolinke/boosting_tree_benchmarks/tree/master/data) dataset to have a quick test.

ebernhardson:
Ran each of these for 10 iterations against the full higgs dataset (xeon e5-2640@2.4ghz). Numbers are all close I would judge the patch as having no performance impact. HDFS speed is also not so bad at 80% of line speed (and the nic is shared).

| | min | mean | max
| master | 11.1s | 11.7s | 14.7s
| hdfs patch, USE_HDFS=0, local file | 10.9s | 11.4s | 13.9s
| hdfs patch, USE_HDFS=1, local file | 11.2s | 11.9s | 14.8s
| hdfs patch, hdfs file | 1m15s | 1m18s | 1m22s

guolinke:
@ebernhardson thanks very much. It is a very great PR.

src/io/file_io.cpp
```diff
@@ -0,0 +1,178 @@
...
+#include <hdfs.h>
+#endif
+
+namespace LightGBM{
+
+struct LocalFile : VirtualFileReader, VirtualFileWriter {
+  LocalFile(const std::string& filename, const std::string& mode) : filename_(filename), mode_(mode) {}
+  virtual ~LocalFile() {
+    if (file_ != NULL) {
+      fclose(file_);
```
guolinke: set file to NULL after close ? I am not sure will `fclose` change FILE to NULL or not.

PR #1250, opened 2018-02-27
minor change for #1243

chivee:
align coding style & adding file to .filter, for #1243


guolinke:
@chivee wrong line ending ?

PR #1276, opened 2018-03-20
[R-package] Fix version number in README

jameslamb:
Right now the build instructions for R reference 2.0.4, but currently the R package is at 2.1.0. Updating it so instructions can be copy-paste.


jameslamb:
fyi @Laurae2

Laurae2:
Ping @guolinke, can't accept changes from my phone

PR #1295, opened 2018-04-01
[docs] fix broken url

guolinke:


chivee:
a reasonable solution is to extend timeout time by adding `--timeout=NUMBER`

guolinke:
@StrikerRUS we can check its log regularly, and manually fix them.

StrikerRUS:
@chivee Do you think that 60sec for timeout in not enough?

PR #1310, opened 2018-04-11
added forced_splits as a parameter

jerryjliu:
The `forced_splits` parameter allows the user to specify a .json file that defines splits to "force" at the top of every decision tree during training. 

The format of the `.json` file looks something like this: 
```
{
   "feature": 4,
   "threshold": 1.5,
   "left": {
   ...
   },
   "right": {
   ...
   }
}
```
where "left" and "right" are present if there are additional subsplits to define. The json dictionary can be nested arbitrarily. 

Having this field is useful for a few reasons: 

1. **Debugging and Analysis**: Allowing users to force splits in decision tree learning will give them additional insight into how informative their features are. For instance, maybe the user has a hypothesis that the model behaves very differently when the feature < X compared to when the feature > X. They can force the split (feature, X) to see which values X lead to the lowest loss. Note that this doesn't imply that forcing the split is supposed to be better than black-box learning; here, forcing splits isn't meant to improve the performance of the model so much as helping the user understand their features. 

2. **Improved loss for subgroups**: When a model is trained on data, it optimizes a global objective function such as mean squared error or avg log loss across all the data; as a result, what ends up happening is that certain subgroups become neglected in optimizing for this global average. For instance, if your data consists of 1 million US users and 10k Spanish users, the resulting loss on the Spanish users can be quite high because they're such a small proportion of the overall data. Forcing splits can help explicitly define the subgroups you care about at the very top of every decision tree so that subsequent splits will help optimize directly within that subgroup. This, combined with setting appropriate weights on the data, can help ensure that the loss isn't significantly worse on one subgroup than another. It will reduce the need to train separate models on each subgroup. 

A few additional notes: 
- for sake of consistency with the existing code, the forced "threshold" will snap to that of the closest threshold bin. 
- If forcing the split causes a negative gain at any point, then forcing is aborted and regular best-first learning will resume. 

Thanks so much and let me know if you have any questions! This work was done as part of efforts to improve machine learning models at [Quora](https://github.com/quora).


include/LightGBM/tree_learner.h
```diff
@@ -4,9 +4,12 @@
 
 #include <LightGBM/meta.h>
 #include <LightGBM/config.h>
+#include <LightGBM/json.hpp>
 
 #include <vector>
 
+using json = nlohmann::json;
```
guolinke: can we use `json_t` for the better reading ?

docs/Parameters.rst
```diff
@@ -504,6 +504,15 @@ IO Parameters
...
    -  separate by ``,`` for multi-validation data
 
+-  ``forced_split``, default=\ ``""``, type=string
+
+   -  path to a ``.json`` file that specifies splits to force at the top of every decision tree before best-first learning commences.
+
+   - ``.json`` file can be arbitrarily nested, and each split contains ``feature``, ``threshold`` fields, as well as ``left`` and ``right`` 
+     fields representing subsplits
+
+   -  see ``examples/binary_classification/forced_splits.json`` as an example.
```
guolinke: should this be applied to all trees, or just first k trees ?

guolinke: can we simplify the input without json format?

include/LightGBM/config.h
```diff
@@ -105,6 +105,10 @@ struct IOConfig: public ConfigBase {
   std::string output_result = "LightGBM_predict_result.txt";
   std::string convert_model = "gbdt_prediction.cpp";
   std::string input_model = "";
+
+  /* filename of forced splits */
+  std::string forcedsplits_filename = "";
```
guolinke: can we put this into boosting_config?

docs/Parameters.rst
guolinke: can we support for the categorical feature too ?

src/treelearner/feature_histogram.hpp
```diff
@@ -101,6 +101,79 @@ class FeatureHistogram {
     output->gain -= min_gain_shift;
   }
 
+  void GatherInfoForThreshold(double sum_gradient, double sum_hessian,
```
guolinke: why need this function ?

src/treelearner/serial_tree_learner.cpp
```diff
@@ -172,18 +173,29 @@ Tree* SerialTreeLearner::Train(const score_t* gradients, const score_t *hessians
   int cur_depth = 1;
   // only root leaf can be splitted on first time
   int right_leaf = -1;
-  for (int split = 0; split < tree_config_->num_leaves - 1; ++split) {
+
+  int init_splits = 0;
+  bool aborted = false;
+  if (forced_split_json != nullptr) {
+    init_splits = ForceSplits(tree.get(), forced_split_json, &left_leaf,
```
guolinke: it seems the force_split is only used in the root node, is that correct?

guolinke: I find there is loop inside `ForceSplits`, however, the num_leaves is only reduced by 1 here.

jerryjliu:
Hey @guolinke, thanks so much for all your comments! I'll push out an updated version soon with support for categorical splits. Regarding the input format, I was curious if you had ideas for alternatives besides json? The reason I picked json was because it's standardized and can support this notion of splits without too much preprocessing / postprocessing.

guolinke:
@jerryjliu i think you can use json for now.

jerryjliu:
@guolinke just pushed an updated version that addressed most of your comments and added support for categorical splits. I could add an extra param to only force splits for the first k trees but I kept it as is (force for every tree) for simplicity. 

Also the Travis CI build check is failing because the GCC version (4.8.4) isn't supported by the [JSON library](https://github.com/nlohmann/json) I'm using - I can swap to a different one (https://github.com/open-source-parsers/jsoncpp) if desired.

guolinke:
Thanks @jerryjliu very much. 
I think it is almost ready now. 
BTW, do you have some performance(speed) tests? I think this PR should have very little impact on the training without force splits.
Though from the code, the performance impact should be very slight. But it is better if there are some numbers.

jerryjliu:
Thanks @guolinke! I did a few speed comparisons in the binary classification example with and without the force_splits parameter and found no statistically significant difference in speed. The median time for both was around 2.15-2.2 seconds, with +0.05 sec variance. In fact the force_split parameter might have sped up the time a little (more early stopping maybe). 

I also tested an earlier diff without my PR and found no difference.

src/treelearner/serial_tree_learner.cpp
```diff
@@ -172,18 +173,29 @@ Tree* SerialTreeLearner::Train(const score_t* gradients, const score_t *hessians
...
+                              &right_leaf, &cur_depth, &aborted);
+  }
+
+  for (int split = init_splits; split < tree_config_->num_leaves - 1; ++split) {
     #ifdef TIMETAG
     start_time = std::chrono::steady_clock::now();
     #endif
     // some initial works before finding best split
-    if (BeforeFindBestSplit(tree.get(), left_leaf, right_leaf)) {
+    if (!aborted && BeforeFindBestSplit(tree.get(), left_leaf, right_leaf)) {
```
guolinke: what is `aborted` used for ?

src/treelearner/serial_tree_learner.cpp
```diff
@@ -528,6 +540,159 @@ void SerialTreeLearner::FindBestSplitsFromHistograms(const std::vector<int8_t>&
   #endif
 }
 
+int32_t SerialTreeLearner::ForceSplits(Tree* tree, Json forced_split_json, int* left_leaf,
```
guolinke: I think the argument of these `Json` could be `const Json &`, since you didn't change their contents, and reference can save the copy time.

src/treelearner/serial_tree_learner.cpp
jerryjliu: Aborted is set if force splits aborts - if the gain from the forced split is worse than without the split. At this point, BeforeFindBestSplit and FindBestSplits have already been called in ForceSplits, so they can't be called again in the first iteration of the main training loop. This is because the functions aren't idempotent - calling them twice leads to wrong behavior. Passing in an aborted flag was the best way to detect this without modifying the existing code.

guolinke: got it, thanks very much

guolinke: maybe rename `aborted` to `aborted_last_force_split` ?

examples/binary_classification/forced_splits.json
```diff
@@ -0,0 +1,12 @@
...
+    "threshold": 1.30,
+    "left": {
+        "feature": 26,
+        "threshold": 0.85
+    },
+    "right": {
+        "feature": 26,
+        "threshold": 0.85
+    }
+}
```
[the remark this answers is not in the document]
jerryjliu: Thanks for the reminder - my code previously worked for certain unbalanced trees, but in doing some more stress tests I noticed a few subtle bugs which I fixed in the latest diff (including adding splits with nan gains). Should be fixed now!

guolinke:
@jerryjliu can you fix these tests ?

Issue #14, opened 2016-10-18
[Feature] Support for monotonic constraints?


alexvorobiev:
Are you planning support for monotonic constraints? See e.g. here https://github.com/dmlc/xgboost/issues/1514

chivee:
I'm pasting the snippets for the monotonic constraints here

`IF (split is a continuous variable and monotonic)`

`THEN take average of left and right child nodes if current split is used`

`   IF monotonic increasing THEN CHECK left average <= right average`

`   IF monotonic decreasing THEN CHECK left average >= right average`

@alexvorobiev , do you  have referable papers for this features?

alexvorobiev:
@chivee I only have the reference to the R GBM package https://cran.r-project.org/package=gbm

chivee:
@alexvorobiev , thanks for your sharing. I'm trying to get the idea behind this method.

AbdealiJK:
Note that the given pseudo code only ensures the split to be in the correct order and not the whole model as a later split could lead the model to be non monotonic

aldanor:
Any thoughts on this?

mayer79:
From practical perspective (outside kaggle-world!), this feature would be extremely helpful in many applications where reasonable model behavior is relevant.

aldanor:
@guolinke Would you be able to advise how to approach this and whether it's feasible? I.e., where should it belong, would it be sufficient to implement it just somewhere in `feature_histogram.hpp`? I guess `FeatureMetainfo` could just contain the `-1/0/1` constraint then.

Here's the meat of the implementation in XGBoost, for reference: https://github.com/dmlc/xgboost/blob/master/src/tree/param.h#L422 -- all of it pretty much contained in `CalcSplitGain()`, plus `CalcWeight()`. Where would stuff like this go in LightGBM?

guolinke:
@aldanor 
I don't know the details about the monotonic constraints. 
What is the idea? And why it is needed? 

following may is useful:

The split gain calculation: https://github.com/Microsoft/LightGBM/blob/master/src/treelearner/feature_histogram.hpp#L291-L297 

The leaf-output calculation:
https://github.com/Microsoft/LightGBM/blob/master/src/treelearner/feature_histogram.hpp#L305-L308

aldanor:
> @aldanor
> I don't know the details about the monotonic constraints.
> What is the idea? And why it is needed?

[...]

PR #1314, opened 2018-04-15
Support monotone constraint

guolinke:
for #14 

ping @alexvorobiev @AbdealiJK @aldanor @redditur for the tests.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,7 +97,7 @@ if(USE_HDFS)
 endif(USE_HDFS)
 
 if(UNIX OR MINGW OR CYGWIN)
-    SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -pthread -O3 -Wextra -Wall -Wno-ignored-attributes -Wno-unknown-pragmas")
+    SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -pthread -O3 -Wextra -Wall -Wno-ignored-attributes -Wno-unknown-pragmas -Wno-return-type")
 endif()
 
 if(WIN32 AND MINGW)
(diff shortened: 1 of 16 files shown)


mayer79:
On https://github.com/mayer79/partialPlot there is a simple partial dependence plot function in R that also works for lightGBM (see regression examples). With monotone constraints, the curves would be expected to be monotone.

mayer79:
@guolinke tested on R-Windows-VC-CPU with the following promising results for diamonds properties on log-price

### Without constraints
no_constraints

### With `monotone_constraints = '0,1,1,0,1,-1'`

constrained

### Problems/Comments

- `lgb.unloader(wipe = TRUE)` is required now whenever the parameters are to be changed

- During prediction, I get the message "[LightGBM] [Warning] Unknown parameter: NA"

- I tested one example with a constrained categorical variable. It seems to be (correctly) ignored.

guolinke:
@mayer79 thanks very much, it seems the monotone constraint is working. 
@Laurae2 any ideas about these R's errors ?

mayer79:
@guolinke: this will be an inherently useful parameter in real-world applications where users can play around with input parameters. Big thanks for making it run.

redditur:
@guolinke: thanks for doing this. I've had a look through the code and the logic seems correct to me. I've also ran some tests and everything seems to work as expected.

Given this is a non-trivial addition to LightGBM, it would probably make sense to bump the version; to 2.1.1 perhaps?

UserQuestions:
Is this in a state that would be easy for me to build and use?  I have an application that might be a good test.

junegit:
Is the algorithm for Light GBM  same as of xgboost to enforce monotonicity? Is it global or local ?
Also can one of the expert helps me in understanding line [L412-417](https://github.com/dmlc/xgboost/blob/master/src/tree/param.h#L412).  from xgboost.  Why "w" is bounded- upper and lower. How this helps to maintain Monotonicity. Line 457 - Why "mid" is used? Is any whitepaper on the monotonic constraints in Light GM available?

PR #1323, opened 2018-04-18
[R] fix warning at prediction

guolinke:


guolinke:
@mayer79  can you test this for the warning in prediction ?

guolinke:
thanks @mayer79

Issue #1344, opened 2018-04-26
[R-package] [Fatal] Check failed: num_data > 0


vspinu:
## Environment info
 setting  value                                             
 version  R Under development (unstable) (2018-02-17 r74265)
 os       Ubuntu 16.04.4 LTS                                
 system   x86_64, linux-gnu                                 
 ui       X11                                               
 language                                                   
 collate  en_GB.UTF-8                                       
 tz       Europe/Amsterdam                                  
 date     2018-04-27                                        

 lightgbm    * 2.1.0       2018-04-26 local                            

## Error Message:

```
[LightGBM] [Fatal] Check failed: num_data > 0 at /home/vspinu/VC/LightGBM/R-package/src/src/io/dataset.cpp, line 27 .

Error in dtrain$construct() : 
  lgb.Dataset.construct: cannot create Dataset handle
```

## Reproducible examples

```R
mat <- structure(c(1, 1, 0, 0, 0, 2, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 2, 3),
                 .Dim = c(10L, 2L),
                 .Dimnames = list(NULL, c("holid_nat", "holid_state")))
label <- 1:10
dtrain <- lgb.Dataset(mat, label = label)
dtrain$construct()
```

Could be related to #1341 and #1280.

guolinke:
you can use 3.4.x R for now.

guolinke:
fixed in the latest commit

vspinu:
This doesn't seem to be fixed.  I am still seeing it with the most recent master:

```R
─ Session info ───────────────────────────────────────────────────────────────
 setting  value
 version  R Under development (unstable) (2018-12-01 r75733)
 os       Ubuntu 18.04.1 LTS
 system   x86_64, linux-gnu
 ui       X11
 language
 collate  en_GB.UTF-8
 ctype    en_GB.UTF-8
 tz       Europe/Amsterdam
 date     2019-01-11

─ Packages ───────────────────────────────────────────────────────────────────
 package     * version    date       lib source
 assertthat    0.2.0      2017-04-11 [1] CRAN (R 3.5.0)
 backports     1.1.2      2017-12-13 [1] CRAN (R 3.5.0)
 callr         3.0.0.9002 2018-12-03 [1] Github (r-lib/callr@1cbce9d)
 cli           1.0.1      2018-09-25 [1] CRAN (R 3.6.0)
[...]

Issue #1341, opened 2018-04-26
R-Package in new R version 3.5.0


weltherrschaf:
Is the R-Package of LGB supported under the latest R release? Can't get it to work yet, is anybody experiencing similar issues?

## Environment info
Operating System: Windows 10
C++/Python/R version: R: 3.5.0

## Error Message:
Just updating R and using the old LGB install I get:

"Error: package or namespace load failed for ‘lightgbm’:
 package ‘lightgbm’ was installed by an R version with different internals; it needs to be reinstalled for use with this R version"

After reinstalling LGB using the visual studio approach without errors, I'm getting this error when trying to run a minimal example:

"Loading required package: Matrix
Error in data$construct() : 
  lgb.Dataset.construct: cannot create Dataset handle"

## Reproducible examples

library(lightgbm)
data(agaricus.train, package = "lightgbm")
train <- agaricus.train
dtrain <- lgb.Dataset(train$data, label = train$label)
params <- list(objective = "regression", metric = "l2")
model <- lgb.cv(params,
                dtrain,
                10,
                nfold = 5,
                min_data = 1,
                learning_rate = 1,
                early_stopping_rounds = 10)

## Steps to reproduce

1. Start with a Running LGB install on R 3.4.3
2. Update R to either R 3.5.0 or R 3.5.0patched
3. Build new LGB

I tried several clean installs, getting the same error though

guolinke:
@Laurae2

Laurae2:
@guolinke I am getting the following in `construct`:

```r
R version 3.5.0 (2018-04-23) -- "Joy in Playing"
Copyright (C) 2018 The R Foundation for Statistical Computing
Platform: x86_64-w64-mingw32/x64 (64-bit)

R is free software and comes with ABSOLUTELY NO WARRANTY.
You are welcome to redistribute it under certain conditions.
Type 'license()' or 'licence()' for distribution details.

R is a collaborative project with many contributors.
Type 'contributors()' for more information and
'citation()' on how to cite R or R packages in publications.

Type 'demo()' for some demos, 'help()' for on-line help, or
'help.start()' for an HTML browser interface to help.
Type 'q()' to quit R.

(...)

Browse[2]> lgb.call("LGBM_DatasetCreateFromCSC_R", 
+          ret = handle, private$raw_data@p, private$raw_data@i, 
+          private$raw_data@x, length(private$raw_data@p), 
+          length(private$raw_data@x), nrow(private$raw_data), 
[...]

PR #1346, opened 2018-04-27
fix R build for version 3.5 and above

guolinke:
refer to #1341 #1344


guolinke:
The table of matrix is run on R 3.5 vesion? as a result, these fails are expected?
Is it expected that the l2 results are different?

Laurae2:
@guolinke The results are expected, I inverted R 3.4.4 and R 3.5 initially though (corrected now).

The L2 results are different due to the CV fold randomness. Using the same seed for all runs, the results become identical.

PR #1347, opened 2018-04-29
Bump LightGBM version to 2.1.1, fix R integer input crash

Laurae2:
Bumping the R version and changing the last update date (so we can see the difference between before/after R 3.5).


guolinke:
I think the new version is 2.1.1, and we should update cli/python version as well

Laurae2:
@guolinke Switched R/Py/CLI to 2.1.1.

Is there anything else to change for the Python version?

guolinke:
there are some bug fixed and new features, e.g. force splits

Issue #1317, opened 2018-04-16
[python] Adding learning_rates to args of sklearn API


mlisovyi:
The training API in the python package has a very handy feature of adjusting learning rate with iterations. However, this is not on the list of args of `LGBMModel.fit()`. Is there any particular reason not to do so? Otherwise, this feature would be very welcome.

From the `sklearn.py ` code structure it seems to be as easy as adding an arg to the `fit()` method and passing it over to the `train()` call.

Thanks in advance, 
Misha

guolinke:
I am not very familiar with the sklearn interface. However, I think learning_rates isn't the standard features of sklearn. I don't know adding this is right or not, since it may break the compatibility with sklearn.
@StrikerRUS any ideas about this ?

mlisovyi:
@guolinke you are right, `learning_rates` is not a standard feature of the sklearn interface. However, it seems that there are many non-standard args of the LGBMModel.fit() method. Since I'm a new user I might have overlooked a basic conceptual reason from the LightGBM philosophy.

StrikerRUS:
@guolinke I have the same thoughts as you. Sklearn interface itself is very conservative one. On the one hand, I don't think that overloading sklearn interface by all possible features is good idea, because it will confuse new users, who expect only standard (for sklearn estimatior) `fit(), .predict()` behavior with minimum params. On the other hand, LightGBM's sklearn wrapper has **already** additional functionality, so maybe it's OK to add another one.

mlisovyi:
Any decision on this issue? 

I personally find the `learning_rates` very useful. But as a user I also strongly prefer to use the sklearn API, as it allows to make use of many nice sklearn functionalities. Thus, it would be wonderful to have such advanced (and useful) feature available in this interface. This arg can have the default disabled value and a beginner user doesn't have to be exposed to it

stevenknguyen:
@mlisovyi Can't you do this by passing an argument as a `reset_parameter` callback in the sklearn api?

mlisovyi:
@stevenknguyen  This sounds interesting. Since I'm a new user, I have not mastered LightGBM callbacks yet. Do you have an example of what you have in mind?

guolinke:
@mlisovyi refer to https://lightgbm.readthedocs.io/en/latest/Python-API.html#lightgbm.reset_parameter
you can use `callbacks= [callback.reset_parameter(learning_rate=learning_rates)]`

stevenknguyen:
[...]

PR #1354, opened 2018-05-03
[python][docs] add info on adaptive learning rate in the sklearn API

mlisovyi:
A small add-on to the docs to clarify how users of the sklearn API can implement shrinkage of the learning rate via a callback
The PR follows the discussion in #1317

### DIFF
--- a/docs/Python-Intro.rst
+++ b/docs/Python-Intro.rst
@@ -196,8 +196,7 @@ The model will train until the validation score stops improving.
 Validation error needs to improve at least every ``early_stopping_rounds`` to continue training.
 
 If early stopping occurs, the model will have an additional field: ``bst.best_iteration``.
-Note that ``train()`` will return a model from the last iteration, not the best one.
-And you can set ``num_iteration=bst.best_iteration`` when saving model.
+Note that ``train()`` will return a model from the best iteration.
 
 This works with both metrics to minimize (L2, log loss, etc.) and to maximize (NDCG, AUC).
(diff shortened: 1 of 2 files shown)


StrikerRUS:
@mlisovyi Thanks for the PR!

I suppose, this description is too long and isn't related to the describing parameter directly. Common users don't want to know such detailed example. I think, something like "You can use ``callbacks`` parameter of ``fit`` method to schedule learning rate by ``reset_parameter`` callback" will be enough. Common user will read this and go to callbacks description to learn more, while experienced user will just recollect described feature in his memory.

mlisovyi:
@StrikerRUS OK, that makes sense. I'll reimplement it later today following your proposal

mlisovyi:
I've shortened the learning_rate description but kept a note that this would ignore the actual argument value. Let me know if you do not like it.

I've also made another small update to the docs following #1355

StrikerRUS:
@mlisovyi Thanks! LGTM!
But what about to add a phrase ".. to shrink/adapt learning rate **while training** ..."?

mlisovyi:
This indeed makes the statement more clear, but i changed it to "*in* training" (following my poor English taste :smile:  )

StrikerRUS:
Sure! 😃 

Now please fix whitespaces and I'll be able to merge.
image

mlisovyi:
fixed, apologies, I overlooked that

StrikerRUS:

StrikerRUS:
@guolinke I cannot merge due to broken GPU tests. If you have no comments, please merge.

PR #1358, opened 2018-05-05
[WIP][ci][python] updated pep8 to pycodestyle

StrikerRUS:
pep8 is now pycodestyle.
Refer to https://github.com/PyCQA/pycodestyle/issues/466 .

This PR removes the following warning:
image

New version finds some warnings, which should be fixed.


StrikerRUS:
Added W503 into ignore list according to https://github.com/PyCQA/pycodestyle/issues/498#issuecomment-331454473 .

Also I've excluded `.nuget` folder from checks because the line `      <file src="lib\**" target="lib"/>` yields "W605 invalid escape sequence '\*'". Is there any way to escape * without breaking the functionality?

Issue #1336, opened 2018-04-23
Why not have binary classification initialize to the base rate by default?


zkurtz:
Currently, a logistic regression on unbalanced data evidently initializes the probability(TRUE) = 0.5 regardless of the base rate. For example, with a base rate of ~0.1 and a single boosting round with a low learning rate, the resulting scores are all very close to 0.5:

```
library(lightgbm)

# Simulate some data
set.seed(0)
n_samples = 10000
x1 = rnorm(n_samples)
x2 = rnorm(n_samples)
X = cbind(x1, x2)
binary_labels = rbinom(n_samples, size = 1, prob = 0.1)
train = lightgbm::lgb.Dataset(data = X, label = binary_labels)

# Fit a booster
bst = lightgbm::lgb.train(
    params = list(
        objective = "binary",
        learning_rate = 0.01
    ),
    data = train,
    nrounds = 1,
    verbose = 0)

# Fitted rate of label == 1:
fitted_values = predict(bst, X)
mean(fitted_values) # = 0.49598
# Base rate of label == 1:
mean(binary_labels) # = 0.0982
```

This is particularly problematic in small datasets where overfitting is an immediate concern. It means that the booster must be run a sufficient number of iterations at a sufficiently high learning rate _just to learn the base rate_ even if fewer iterations at lower learning rate are sufficient to arrive at the "overfitting frontier". Is there any reason not to be initializing to the base rate instead of 0.5?

To be slightly more regularized, one could initialize to the [Bayesian + uniform prior](https://www2.stat.duke.edu/courses/Fall10/sta290/Lectures/Binomial/binomial.pdf) estimator of the base rate, biasing towards 0.5 slightly: (y + 1)/(n + 2) where y is the number of positive cases and n is the number of cases.

guolinke:
@zkurtz we have `BoostFromAverage` (https://github.com/Microsoft/LightGBM/blob/master/src/boosting/gbdt.cpp#L353-L375). I also test it for binary classification cases. However, most of them don't have improvement (with enough iterations and early stopping), although it can speed up the convergence slightly.

zkurtz:
Yes I think this would be an improvement for binary classification. I'll do some testing via `init_score`, which looks like it could achieve the desired behavior. If that works, I'll try to figure out how to modify the source code and do a PR -- this would be my first!

olofer:
You can achieve what you want already (unless I'm mistaken) by using the boost-from-average option with the xentropy objective.

[...]

PR #1361, opened 2018-05-08
Enable boost_from_average as default for binary objective

zkurtz:
This appears to fix issue #1336, leading to greater default boosting efficiency particularly for small unbalanced binary regression problems.

The change is simple; I copied verbatim the BoostFromScore method in the xentropy objective:

https://github.com/Microsoft/LightGBM/blob/5392c9eaf616b10028170346db898a598739a1a7/src/objective/xentropy_objective.hpp#L107-L128

To the future: There is much in common between `xentropy` and `binary` objectives; it's not clear to me yet whether one should subsume the other.  I.e., if `xentropy` ends up doing essentially the same thing as `binary`, maybe `xentropy` should become simply an alias for `binary`.

### DIFF
--- a/src/objective/binary_objective.hpp
+++ b/src/objective/binary_objective.hpp
@@ -114,6 +114,29 @@ public:
       }
     }
   }
+  
+  // implement custom average to boost from (if enabled among options)
+  double BoostFromScore() const override {
+    double suml = 0.0f;
+    double sumw = 0.0f;
+    if (weights_ != nullptr) {
+      #pragma omp parallel for schedule(static) reduction(+:suml,sumw)
+      for (data_size_t i = 0; i < num_data_; ++i) {
+        suml += label_[i] * weights_[i];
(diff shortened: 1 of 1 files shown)


olofer:
@zkurtz xentropy allows continuous target labels in [0,1] whereas binary assumes targets that are distinct 0 or 1; "probability regression" versus classification if you will, sometimes that distinction can be useful; binary can also be slightly optimized/simplified for the two-value case.

zkurtz:
Great feedback, I'll close this for now while I take a closer look.

guolinke:
@zkurtz can you discard these duplicated changes ?

zkurtz:
@guolinke Does this look better? I'm not sure how the duplication occurred, but doing `git push origin +HEAD^:master` just one time seemed to fix it as per https://stackoverflow.com/a/8225166/2232265

PR #1362, opened 2018-05-09
[python] decode error description

StrikerRUS:
This PR makes `LightGBMError` description string type.

@guolinke Could the following logs contain new line character to make them easy-read?

For instance:
image

https://github.com/Microsoft/LightGBM/blob/7d3206e0a43cb7e65f846338a055c298dac55b90/src/network/linkers_socket.cpp#L113-L114

https://github.com/Microsoft/LightGBM/blob/e005cdb049271ecdd4e631043517ac18f6c9a867/src/io/dataset_loader.cpp#L54-L56

https://github.com/Microsoft/LightGBM/blob/e005cdb049271ecdd4e631043517ac18f6c9a867/src/io/dataset_loader.cpp#L49-L50

https://github.com/Microsoft/LightGBM/blob/e005cdb049271ecdd4e631043517ac18f6c9a867/src/io/dataset_loader.cpp#L86-L88

https://github.com/Microsoft/LightGBM/blob/e005cdb049271ecdd4e631043517ac18f6c9a867/src/io/dataset_loader.cpp#L106-L108

https://github.com/Microsoft/LightGBM/blob/e005cdb049271ecdd4e631043517ac18f6c9a867/src/io/dataset_loader.cpp#L126-L128

https://github.com/Microsoft/LightGBM/blob/e005cdb049271ecdd4e631043517ac18f6c9a867/src/io/dataset_loader.cpp#L150-L152

https://github.com/Microsoft/LightGBM/blob/e005cdb049271ecdd4e631043517ac18f6c9a867/src/io/dataset_loader.cpp#L164-L165

https://github.com/Microsoft/LightGBM/blob/aa78a6b930c398b1cbeac80e4f8350d8c84f3285/src/metric/dcg_calculator.cpp#L133-L134

https://github.com/Microsoft/LightGBM/blob/e005cdb049271ecdd4e631043517ac18f6c9a867/src/io/dataset.cpp#L229-L232

https://github.com/Microsoft/LightGBM/blob/0af44ac819fd86b39c3a80a3ba56ab08499d3b1b/src/c_api.cpp#L45-L46


guolinke:
add new_line will bring a new problem: there are different new lines in linux, windows, mac os...

StrikerRUS:
Oh, I see...

But there are already `\n` chars in Log class:

https://github.com/Microsoft/LightGBM/blob/92840c5cda1cca915a4a76a07c840252c479de12/include/LightGBM/utils/log.h#L81
https://github.com/Microsoft/LightGBM/blob/92840c5cda1cca915a4a76a07c840252c479de12/include/LightGBM/utils/log.h#L93


I'm not cpp experienced user at all, but StackOverflow says `\n` is enough for cross-platform new line.
https://stackoverflow.com/a/34915271

Also, I have both Linux and macOS VMs and can test output.

guolinke:
@StrikerRUS oh, sorry, i forget that.
then sure, we can use `\n` .

guolinke:
@StrikerRUS these is a confilct

StrikerRUS:
@guolinke 
>these is a conflict

Fixed.

Issue #1371, opened 2018-05-16
feval metric score ignored for early stopping with Python API


ClimbsRocks:
When using `early_stopping_rounds`, the `feval` function is ignored, and early stopping is determined by the default metric (binary logloss in this case). I included a reproducible example below.

I don't see any way of turning off the default metric (I tried manually setting it to `'metric': None` in the params dict, but it still exhibited this behavior). 

I created a trivial "custom metric" that's just an incrementing counter. I would expect the model to train indefinitely, but instead we see it trains only as long as binary logloss is improving. 

```
[210]	training's binary_logloss: 0.00115081	training's feval_func: -210
[211] 	training's binary_logloss: 0.00113343	training's feval_func: -211
[212]	training's binary_logloss: 0.00112853	training's feval_func: -212
[213]	training's binary_logloss: 0.00104986	training's feval_func: -213
[214]	training's binary_logloss: 0.000965577	training's feval_func: -214
[215]	training's binary_logloss: 0.000960396	training's feval_func: -215
[216]	training's binary_logloss: 0.000971713	training's feval_func: -216
[217]	training's binary_logloss: 0.000941262	training's feval_func: -217
[218]	training's binary_logloss: 0.000849178	training's feval_func: -218
[219]	training's binary_logloss: 0.000841552	training's feval_func: -219
[220]	training's binary_logloss: 0.000803302	training's feval_func: -220
[221]	training's binary_logloss: 0.000685291	training's feval_func: -221
[222]	training's binary_logloss: 0.000689522	training's feval_func: -222
[223]	training's binary_logloss: 0.000736344	training's feval_func: -223
[224]	training's binary_logloss: 0.000816562	training's feval_func: -224
[225]	training's binary_logloss: 0.000829289	training's feval_func: -225
[226]	training's binary_logloss: 0.000799258	training's feval_func: -226
[227]	training's binary_logloss: 0.000837895	training's feval_func: -227
[228]	training's binary_logloss: 0.000844479	training's feval_func: -228
[229]	training's binary_logloss: 0.000862614	training's feval_func: -229
[230]	training's binary_logloss: 0.000855565	training's feval_func: -230
[231]	training's binary_logloss: 0.000778359	training's feval_func: -231
Early stopping, best iteration is:
[221]	training's binary_logloss: 0.000685291	training's feval_func: -221
```

Is feval supposed to only be used for logging, and not `early_stopping_rounds`? If so, should I look into custom callbacks to determine early_stopping? 

[...]

PR #1373, opened 2018-05-17
Clarify docs for feval

ClimbsRocks:
Clearing up the confusion from: https://github.com/Microsoft/LightGBM/issues/1371

### DIFF
--- a/python-package/lightgbm/engine.py
+++ b/python-package/lightgbm/engine.py
@@ -40,7 +40,9 @@ def train(params, train_set, num_boost_round=100,
         Customized objective function.
     feval : callable or None, optional (default=None)
         Customized evaluation function.
+        Should accept two parameters: preds, train_data.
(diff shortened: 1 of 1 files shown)


python-package/lightgbm/engine.py
```diff
@@ -40,7 +40,9 @@ def train(params, train_set, num_boost_round=100,
         Customized objective function.
     feval : callable or None, optional (default=None)
```
StrikerRUS: @ClimbsRocks Please add that it could be a string.
`feval : callable, string or None, optional (default=None)`

ClimbsRocks:
@StrikerRUS just added. Sorry for the extra commit fixing the missed space- seems like a good candidate for a squash and merge, if that's not already the default for PRs like this.

StrikerRUS:
@ClimbsRocks Thank you for the PR!

PR #1376, opened 2018-05-18
[docs] documented crash in case categorical values is bigger max int32

### LINKED ISSUE #1359: Categorical features with large int cause segmentation fault
## Environment info
Operating System: Ubuntu server 16.04 64bit
CPU: Intel(R) Xeon(R) CPU E5-2640 v4 @ 2.40GHz * 2
C++/Python/R version: Python 3.5.2

## Error Message:
Python output:
> /home/zhang/.local/lib/python3.5/site-packages/lightgbm/basic.py:1038: UserWarning: categorical_feature in Dataset is overridden. New categorical_feature is ['item_id', 'user_id']
>   warnings.warn('categorical_feature in Dataset is overridden. New categorical_feature is {}'.format(sorted(list(categorical_feature))))
> [LightGBM] [Warning] Met negative value in categorical features, will convert it to NaN
> [LightGBM] [Warning] Met negative value in categorical features, will convert it to NaN
> [LightGBM] [Warning] Met negative value in categorical features, will convert it to NaN
> [1]    69368 segmentation fault  python3 train.py
> 

GDB output:
> [Thread debugging using libthread_db enabled]
> Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
> Core was generated by `python3 train.py'.
> Program terminated with signal SIGSEGV, Segmentation fault.
> #0  LightGBM::BinMapper::FindBin (this=<optimized out>, values=<optimized out>, num_sample_values=<optimized out>, total_sample_cnt=3, max_bin=255, min_data_in_bin=3, min_split_data=20, 
>     bin_type=LightGBM::CategoricalBin, use_missing=true, zero_as_missing=false) at /home/zhang/lightgbm/LightGBM/src/io/bin.cpp:322
> 322	      if (distinct_values_int[0] == 0) {
> [Current thread is 1 (Thread 0x7f22a3957700 (LWP 44144))]

## Reproducible examples
```Python
import lightgbm as lgb
import pandas as pd

data = {'user_id':[4505772604969228686, 2692638157208937547, 5247924392014515924],
       'item_id': [3412720377098676069, 3412720377098676069, 3412720377098676069]}
df = pd.DataFrame(data=data)

lgb_train = lgb.Dataset(df, label=[0, 1, 1])
params = {
    'objective': 'binary',
    'metric': 'binary_logloss'
}

gbm = lgb.train(params, lgb_train, categorical_feature=['user_id', 'item_id'])
```
## Steps to reproduce
1. Run example above

## Possible reason
Seems like it's caused by Python int to C++ int conversion error. Large Python int become negative in C++ side. If all values within a DataFrame column are too large, which is common in ID features, these values will be treated as missing values. Then vector `distinct_values_int` will be empty and `distinct_values_int[0]` will cause access violation.

Use `sklearn.preprocessing,LabelEncoder` can solve this problem. But I think this should be fixed or at least throw Python error message instead of segmentation fault since it will cause Python notebook kernel death.

StrikerRUS:
Closed #1359.

This PR **doesn't** fix the case when large categorical value comes from file (data is a string with path to the file). @guolinke is it possible to check on Python side? Also I don't know whether we need to check data for prediction task. I mean, if there are no such large values in train dataset, what is the chance to meet them in test dataset?

In additional, I'm in doubt about the necessity of these time-consuming checks...


guolinke:
@StrikerRUS current checking is time-consuming ? as it will iterate all int values ?

StrikerRUS:
@guolinke In my opinion, yes. On my ordinal home computer the results are:
```
import numpy as np
import scipy.sparse as sp

MAX_INT32 = (1 << 31) - 1

data = np.random.rand(100000, 100)*100
data[4][2] = 1 << 31

%%timeit
(data[:, [1,2,4]] > MAX_INT32).any()
```

```
100 loops, best of 3: 2.83 ms per loop
```

```
data = sp.random(100000, 100, density=0.4, format='csr')*100
data[4, 2] = 1 << 31

%%timeit
(data[:, [1,2,4]] > MAX_INT32).nnz
```

```
100 loops, best of 3: 16.8 ms per loop
```

Laurae2:
@StrikerRUS In which case are we able to have a >int32 categorical? That's a huge cardinality (it should start from 0 anyway).

I could think about its usage when targeting users for ads but having 2+billion distinct users is a lot..

StrikerRUS:
@Laurae2 The case was described here #1359. It was describing treating **ids** (which are random values >max_int32 and don't start from 0) as categories.

Laurae2:
@StrikerRUS This seems more a preprocessing issue than a LightGBM issue.

In R to avoid such issue we use a rule generator we can apply to new datasets: https://github.com/Microsoft/LightGBM/blob/master/R-package/R/lgb.prepare_rules.R

It transforms the categorical features to numeric features starting from 0 (0 is NA, 1..cardinality are categorical values) so they can be used afterwards as categorical features in LightGBM and other libraries. This issue exists for many other machine learning libraries.

StrikerRUS:
@Laurae2 The same things are done only for Pandas in Python-package at present. My opinion is that it's not LightGBM issue too, but at least we should document this.

guolinke:
@StrikerRUS @Laurae2 
I agreed. it is no reason to have such a great additional cost.

StrikerRUS:
@guolinke I've just found a way to speed it up. But it's practically unnoticeable.

So, what's the conclusion? Reject all checks and enhance the documentation?

StrikerRUS:
@guolinke OK. Then I'll commit a little bit more efficient check and ask you to measure time cost on any real dataset with significant size (unfortunatelly, I can't do it right now because my SSD is completely full).

StrikerRUS:
Done!
Please, compare time consuming of this branch and master on any real rather big dataset.

@guolinke 
I think that enhancing documents is needed anyway, even if we'll perform checks, because we don't preprocess data loaded directly from file and any prediction data. I'll write notes soon.

StrikerRUS:
No matter what will bring the results of time cost, notes about max values in categorical features have been added in docs and docstrings in last commit.

python-package/lightgbm/basic.py
```diff
@@ -772,14 +779,16 @@ def __init_from_np2d(self, mat, params_str, ref_dataset):
...
 
-    def __init_from_csr(self, csr, params_str, ref_dataset):
+    def __init_from_csr(self, csr, params_str, ref_dataset, categorical_indices=None):
         """
         Initialize data from a CSR matrix.
         """
         if len(csr.indices) != len(csr.data):
             raise ValueError('Length mismatch: {} vs {}'.format(len(csr.indices), len(csr.data)))
-        self.handle = ctypes.c_void_p()
+        if categorical_indices is not None and len(categorical_indices) != 0 and csr[:, list(categorical_indices)].max() > MAX_INT32:
```
guolinke: will this faster than csr.data.max() ?

guolinke:
@StrikerRUS 
Just test it.

case 1, without col indices:

```python
import numpy as np
import scipy.sparse as sp
import time

MAX_INT32 = (1 << 31) - 1
loop = 200
data = np.random.rand(1000000, 100)*100

start = time.time()
for _ in range(loop):
    t = data.max()

end = time.time()
print((end - start)/loop)


data = sp.rand(100000,10000,0.01).tocsr()
start = time.time()
for _ in range(loop):
    t = data.max()

end = time.time()
print((end - start)/loop)


data = sp.rand(100000,10000,0.01).tocsc()
start = time.time()
for _ in range(loop):
    t = data.max()

end = time.time()
print((end - start)/loop)
```
all of them are fast, time cost is 
```
0.0329028058052063
0.003496755361557007
0.00352044939994812
```

case 2, with col indices:
```python
import numpy as np
import scipy.sparse as sp
import time

MAX_INT32 = (1 << 31) - 1
loop = 20
data = np.random.rand(1000000, 100)*100
catcols = list(range(0,10)) + list(range(50,60)) + list(range(80,100))

start = time.time()
for _ in range(loop):
    t = data[:,catcols].max()

end = time.time()
print((end - start)/loop)


data = sp.rand(100000,10000,0.01).tocsr()
start = time.time()
for _ in range(loop):
    t = data[:,catcols].max()

end = time.time()
print((end - start)/loop)


data = sp.rand(100000,10000,0.01).tocsc()
start = time.time()
for _ in range(loop):
    t = data[:,catcols].max()

end = time.time()
print((end - start)/loop)
```
it is about 10x slower for dense and csr:
```
0.3913775205612183
0.022434639930725097
0.002259492874145508
```

StrikerRUS:
@guolinke Many thanks!

So what? Remove all checks and leave only notes in docs, right?

guolinke:
yeah, sure

StrikerRUS:
Done!

PR #1380, opened 2018-05-19
[docs] added a note about g++ versions for macOS users

### LINKED ISSUE #1369: LightGBM and gcc 8 in MacOS: `Library not loaded: /usr/local/opt/gcc/lib/gcc/7/libgomp.1.dylib`
For Mac Users
```
Version 10.13.4
Python 3.6.5
```

The 2nd of May `gcc-8` was released, which is a major released. 

Therefore, if, since then, you run 
```
brew update
brew upgrade
```

gcc-8 version will be installed: 

```
➜  ~ gcc-8 --version
gcc-8 (Homebrew GCC 8.1.0) 8.1.0
Copyright (C) 2018 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

## ISSUE
If the lightGBM version you have was built with `gcc-7`, when you try to load it now you might get: 

```
OSError: dlopen(/usr/local/lib/python3.6/site-packages/lightgbm/lib_lightgbm.so, 6): Library not loaded: /usr/local/opt/gcc/lib/gcc/7/libgomp.1.dylib
  Referenced from: /usr/local/lib/python3.6/site-packages/lightgbm/lib_lightgbm.so
  Reason: image not found
```

## SOLUTION
I came around this error by installing it from source doing: 

```
pip uninstall lightgbm
git clone --recursive https://github.com/Microsoft/LightGBM ; cd LightGBM
export CXX=g++-8 CC=gcc-8
mkdir build ; cd build
cmake ..
make -j4
```

and then 

```
pip install --no-binary :all: lightgbm
```

```
In [1]: import lightgbm

In [2]: lightgbm.__version__
Out[2]: '2.1.1'
```

By the way, similar issue will happen to `xgboost`.

If someone thinks this is more for stackoverflow I am happy to remove it from here immediately

Cheers

### LINKED ISSUE #1387: Mac OS install instructions need to be updated
The gcc has been updated to version 8.1 since 17-05-18. The install documentation needs to be updated. Instead of 
`git clone --recursive https://github.com/Microsoft/LightGBM ; cd LightGBM
export CXX=g++-7 CC=gcc-7
mkdir build ; cd build
cmake ..
make -j4`

It should now be
`git clone --recursive https://github.com/Microsoft/LightGBM ; cd LightGBM
export CXX=g++-8 CC=gcc-8
mkdir build ; cd build
cmake ..
make -j4`

This needs to be changed in the install docs here https://lightgbm.readthedocs.io/en/latest/Installation-Guide.html

StrikerRUS:
Closed #1369.
Closed #1387.

Next release will be built with g++-8.


guolinke:
@StrikerRUS 
It seems these documents are still gcc7:
https://github.com/Microsoft/LightGBM/tree/master/R-package#install
https://github.com/Microsoft/LightGBM/blob/master/docs/Installation-Guide.rst#macos

should we fix them too ?

StrikerRUS:
@guolinke To be honest, I don't think we should alter instructions about **compiling** of LightGBM. I mean, it still works perfectly even if compiled with gcc-4 (like at our Travis with Linux). The problem was when someone wanted to use old precompiled binaries with new compiler. The main idea is to compile and use with the same compiler, no matter new or old one it is.

What do you think? How to formulate it clearer?

StrikerRUS:
We need to ask someone with mac about it. Or test at Travis. I can only say that `export CXX=g++` doesn't work when we run before it `brew install gcc` (I already checked it). It seems that it expects concrete version to link and I may be mistaken but I read about that g++ is symlinked to AppleClang.

StrikerRUS:
@guolinke What do you think about just adding these words (or something similar)?
```
# export CXX=g++-7 CC=gcc-7  # for macOS (replace 7 with version of gcc installed on your machine)
```

PR #1382, opened 2018-05-19
[docs][examples] Clarify relationship between `xentropy` and `binary`

zkurtz:
These two objectives are closely related. I included (1) a note in the docs to be explicit that `xentropy` is necessary when labels are not binary and (2) an example that makes the relationship clear for users coming at this with "logistic regression" in mind. 

Non-binary labels are a trivial extension of binary ones under log-loss; any lack of generalization is in the code (i.e. enforcing `int` instead of `float`) rather than mathematical, and so it would be understandable if a user mistakenly tried applying `binary` with non-binary labels. If fact I was doing this in R and have yet to identify why it gave reasonable results and did not crash (while `binary` with non-binary labels does crash in python).

### DIFF
--- a/docs/Parameters.rst
+++ b/docs/Parameters.rst
@@ -69,7 +69,7 @@ Core Parameters
 
       -  ``tweedie``, Tweedie regression with log-link. It might be useful, e.g., for modeling total loss in insurance, or for any target that might be `tweedie-distributed`_
 
-   -  ``binary``, binary `log loss`_ classification (or logistic regression). Requires labels in {0,1}; see `xentropy`_ for general probability labels in [0, 1]
+   -  ``binary``, binary `log loss`_ classification (or logistic regression). Requires labels in {0, 1}; see ``xentropy`` for general probability labels in [0, 1]
 
    -  multi-class classification application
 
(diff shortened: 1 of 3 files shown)


docs/Parameters.rst
```diff
@@ -69,7 +69,7 @@ Core Parameters
 
       -  ``tweedie``, Tweedie regression with log-link. It might be useful, e.g., for modeling total loss in insurance, or for any target that might be `tweedie-distributed`_
 
-   -  ``binary``, binary `log loss`_ classification application
+   -  ``binary``, binary `log loss`_ classification (or logistic regression). Requires labels in {0,1}; see `xentropy`_ for general probability labels in [0, 1]
```
StrikerRUS: Please add a space into a set symbol and remove nonexistent reference to xentropy:
```
... labels in {0, 1}; see ``xentropy`` for ...
```

StrikerRUS:
@zkurtz Can you please also append a short description of what your example is about to [examples README](https://github.com/Microsoft/LightGBM/blob/master/examples/python-guide/README.md)?

StrikerRUS:
Thank you, @zkurtz !

PR #1383, opened 2018-05-19
[ci] generate fresh config_auto.cpp at CI side

StrikerRUS:
I think it's good to always have fresh `config_auto.cpp` at CI side and in consequence in all artifacts.

But we cannot completely remove pre-generated `config_auto.cpp` from the repo, because we cannot force users, who compiles from sources, to run python code on their machines, right?

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -34,6 +34,8 @@ install:
   - activate test-env
 
 build_script:
+  - cd helper && python parameter_generator.py
+  - cd ..
   - mkdir build && cd build
   - cmake -DCMAKE_GENERATOR_PLATFORM=x64 .. && cmake --build . --target ALL_BUILD --config Release
   - cd ..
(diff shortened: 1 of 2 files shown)


guolinke:
@StrikerRUS 
I think the `config_auto.cpp` is generated by developers offline, not the users. 
So it is not needed in CI side, what do you think  ?

StrikerRUS:
I suppose, the best way is to generate it at CI side and push generated file back to the Git repo. This will guarantee always fresh `config_auto.cpp` (even if someone of developers forget to update it after altering `config.hpp`).

But then Travis will create a new separate commit with generated `config_auto.cpp` after each normal commit/PR. So we need to make Travis detect modified files and only in case `config.h` was modified ask it to push back generated `config_auto.cpp`. However, it seems to be rather complicated... At this moment it's really easier ask to regenerate file manually.

guolinke:
thanks, maybe we can refine it in future.

Issue #1374, opened 2018-05-17
Should document the case train_set is specified in valid_sets of lightgbm.train


fujii:
In the document,

http://lightgbm.readthedocs.io/en/latest/Python-API.html#training-api

> If there’s more than one, will check all of them. 

This is not correct. Looking into the source code, train_set is handled differently.

https://github.com/Microsoft/LightGBM/blob/master/python-package/lightgbm/engine.py#L136

guolinke:
@fujii 
Sorry, I didn't got your point. what is the problem ? 

This handling is for efficiency: it reuses the train-set if it is listed in valid_sets.

fujii:
Thank you for your feedback, @guolinke .
The document of `early_stopping_rounds` says it will check *all* of `valid_names`. But, it doesn't check the dataset specified in `train_set`.

guolinke:
okay, got it.
Actually, using early stopping over training set is not reasonable, since the loss over training set will always be reduced.
But you are right, we should point it out in document. And you are welcomed to create a PR for the changes.

fujii:
Thanks. I will try.

StrikerRUS:
Fixed in #1393.

PR #1393, opened 2018-05-23
[docs][python][R] early_stopping_rounds doesn't check all of eval_set

fujii:
The document of `early_stopping_rounds` says it will check all of
eval_set. But, this is not true. It doesn't check the dataset
specified as the training data.

This change appends an extra phrase "except the training data" to all
of the sentences "If there's more than one, will check all of them" in
documents.


fujii:
Fix for https://github.com/Microsoft/LightGBM/issues/1374

Issue #1397, opened 2018-05-24
BUG in Dataset::ConstructHistograms


IbanezJS:
@guolinke 
commit 850f03913a5f182188abf4e291c0ec71db72b7b6  "remove some duplicate codes"
Did bug:
Old Code: `if (!is_groud_used) { continue; }`

After refactoring that line disappered. So now used_group always equals to all groups.

guolinke:
@IbanezJS oh, you are right. Thanks for pointing it out.

PR #1400, opened 2018-05-25
fix a bug in ConstructHistograms

guolinke:


guolinke:
fix #1397, thanks @IbanezJS .
This indeed is a bug, and will hurt the efficiency of using `feature_fraction`

PR #1381, opened 2018-05-19
Refine config object

guolinke:
Previous config objective is complex for developing and documenting. 

With this PR, the changes of parameters can be simply written in `config.h`. 
Then we can use `helper\parameter_generator.py` to generate related codes and documents. 

I just finish the code generating part.
@StrikerRUS can you help for the doc part ?

PR #1409, opened 2018-05-31
[docs] generate parameters description from config file. Stage 1

StrikerRUS:
This is the continuation of #1381.

@guolinke It seems that `helper\parameter_generator.py` cannot handle links in params' descriptions (`desc` and `descl2` fields) due to this:

https://github.com/Microsoft/LightGBM/blob/28832d9e55e016d02c157e541293b39b18dc44af/helper/parameter_generator.py#L27

Can we hotfix it? Maybe three forward slashes?


guolinke:
@StrikerRUS 
refer to my solution here: https://github.com/Microsoft/LightGBM/blob/master/helper/parameter_generator.py#L27-L29 

you can join [1:]

StrikerRUS:
@guolinke 
Sorry, didn't get it.

Did you mean this:
```
tokens = '//'.join(line.split("//")[1:]).split("=")
```

guolinke:
@StrikerRUS yeah, Or just
```line.strip()[2:].stripe()``` ,
which can remove the first two `//` .

**Comment by StrikerRUS** [comment]
@guolinke OK, thank you!

File: include/LightGBM/config.h [hunk 1]
```diff
@@ -76,126 +78,125 @@ struct Config {
...
 
   // [doc-only]
-  // alias=random_seed
-  // desc=Use this seed to generate seeds for others, e.g. data_random_seed.
-  // desc=Will be override if set other seeds as well
-  // default=none
+  // alias = random_seed
+  // desc = this seed is used to generate other seeds, e.g. ``data_random_seed``, ``feature_fraction_seed``
+  // desc = will be overridden, if you set other seeds
   int seed = 0;
```
StrikerRUS: Why default=none?

guolinke: I think it is a typo.

StrikerRUS: I had the same thoughts, so deleted it.

What is your opinion about it https://github.com/Microsoft/LightGBM/pull/1409#pullrequestreview-125173261 ?

StrikerRUS: Rolled back in #1872.

Issue #1412, opened 2018-06-02
"Unknown parameter: device" warning when using sklearn wrapper with gpu support


chencjGene:
params 
{"n_estimators": 50, "max_depth": 5,
             "objective": "multiclass",
             "verbose": -1, "n_jobs": -1, 
             "learning_rate": 0.1,
              "device": "gpu", 
              "num_class": 6
}

StrikerRUS:
@chencjGene what is the version of LightGBM?

You can get it by
```
import lightgbm as lgb

lgb.__version__
```

chencjGene:
My lightgbm version is 2.1.1. This warning does not show up in 2.1.0 version

StrikerRUS:
Does it work despite warning?

guolinke:
@StrikerRUS I think we forget the alias of `device` in the new version, it is `device_type` now.

chencjGene:
@StrikerRUS It seems work. @guolinke Thanks for informing.

StrikerRUS:
@guolinke Yeah! The alias was lost, but `device` is still using here:
https://github.com/Microsoft/LightGBM/blob/a39c848e6456d473d2043dff3f5159945a36b567/src/io/config.cpp#L123

StrikerRUS:
Seems that auto-closing mechanism of GitHub is broken...

PR #1413, opened 2018-06-02
[docs] generate parameters description from config file. Stage 2

StrikerRUS:
This is continuation of #1409.

### DIFF
--- a/.travis/test.sh
+++ b/.travis/test.sh
@@ -86,7 +86,7 @@ fi
 if [[ ${TASK} == "gpu" ]]; then 
     conda install --yes -c conda-forge boost=1.63.0
     if [[ ${METHOD} == "pip" ]]; then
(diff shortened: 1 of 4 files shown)


StrikerRUS:
@guolinke It seems that with new config file this check was lost
https://github.com/Microsoft/LightGBM/blob/497e60ed144007d051bcdfc8bdb36d4f2f5ff681/src/io/config.cpp#L307

guolinke:
@StrikerRUS 
it is moved to https://github.com/Microsoft/LightGBM/blob/master/src/io/config.cpp#L174-L176

StrikerRUS:
@guolinke I'm not sure this is fair replacement of previous check. I mean, at present there is no check if both `valid_data_filenames` and `valid` are specified by user.

guolinke:
@StrikerRUS got it, we can add this checking at that place (config.cpp) too.

StrikerRUS:
@guolinke Please take a look to new commits:
added missed `CHECK` into `config,cpp` and aliases in **Core Parameters** section.
Fixed #1412.

PR #1415, opened 2018-06-03
[docs] generate parameters description from config file. Stage 3

StrikerRUS:
This is the continuation of #1413.


StrikerRUS:
Added alias which is used in config files from `examples` folder.

Issue #1342, opened 2018-04-26
AMD OpenCL SDK download link is down


guolinke:
@huanzhang12 the download link: https://developer.amd.com/amd-accelerated-parallel-processing-app-sdk/ is 404 now. That causes the build failed on check-the-docs and gpu tests.

huanzhang12:
Seems to be a problem with AMD. We need to find another mirror for AMD APP SDK.

StrikerRUS:
They say it's "temporary server problem":
https://community.amd.com/thread/227948
and provide link to the Dropbox:
https://community.amd.com/message/2859384#2859384

Also some mirror is provided here:
https://github.com/fireice-uk/xmr-stak/issues/1511#issuecomment-385120692

StrikerRUS:
Another possible solution is to move GPU tests to the Appveyor. Some links and paths could be fond [here](https://github.com/boostorg/compute/blob/master/.appveyor.yml).

StrikerRUS:
No chance we'll see SDK on the official website:

>Hi Andrey,
As I've been informed, the concerned team has decided to drop the support for APP SDK and hence, any related package on developer.amd.com might not be accessible anymore. Instead, they are recommending to use this lite SDK package available here: https://github.com/GPUOpen-LibrariesAndSDKs/OCL-SDK/releases
If I get more information about this, I'll share with you.
Regards,

https://community.amd.com/thread/228114#2863567

guolinke:
maybe the better solution is deploying the gpu test on Appveyor.
@StrikerRUS can you help for that? Thanks very much.

StrikerRUS:
@guolinke I'll give it a try.

The one thing I'm afraid is that Appveyor is **extremely** slow without concurrent jobs and GPU test will run for ages...

Does anyone familiar with Jenkins? There are machines with NVIDIA cards in it and we'll be able to run tests with CUDA. Guys from XGBoost do this: https://github.com/dmlc/xgboost/blob/master/Jenkinsfile

StrikerRUS:
@huanzhang12 Despite that GPU targeting table from docs says CUDA fails on any CPU, it seems that guys from OpenMM managed to compile and run tests with CUDA at Travis side:
https://github.com/pandegroup/openmm/blob/a885a6ae6bc14390f9540d383fbad8a58a087998/.travis.yml#L153
https://travis-ci.org/pandegroup/openmm/jobs/385865954

[...]

PR #1428, opened 2018-06-06
[ci] bring GPU tests back to Travis

StrikerRUS:
I've uploaded AMD APP SDK to the one of the our existing GitHub releases. It's important to note that it's turned out that downloading from GitHub releases **without authorization** is possible only in case of **tagged release**.

That's why APP SDK has been attached to old `v2.0.12` (last tagged) release. Also I think all next LightGBM releases should be tagged.

Just to note: I was thinking about the caching mechanism but then found this description: https://docs.travis-ci.com/user/caching/#Things-not-to-cache . And this is exactly our case:
>Large files that are quick to install but slow to download do not benefit from caching


StrikerRUS:
Closed #1342.

StrikerRUS:
@guolinke Don't you know why autoclosing issues doesn't work?..

guolinke:
@StrikerRUS 
Sorry, I don't know...

PR #1429, opened 2018-06-06
[ci] simplified and speeded up Travis

StrikerRUS:
All things we need are already installed on Travis's trusty images.


guolinke:
@StrikerRUS it seems there are many failed builds in travis due to the branch clone...

StrikerRUS:
@guolinke I see only `check-docs` fails for the second day, but I don't know how to fix it. Maybe problem will disappear in a few days without our help or we can exclude Laurae's table site from the Linkchecker by parameter.

guolinke:
@Laurae2 can you help to fix the url error of your site ?

StrikerRUS:
I think the best hotfix now, since it's can be reached by ordinal browsers, is just not to check this site. I've created a PR: #1430.

Laurae2:
@guolinke Seems I can't fix it, probably related to Tableau/Amazon certificates.

guolinke:
@StrikerRUS it seems the gpu test is down after merged this PR.

PR #1430, opened 2018-06-08
[ci] ignore tableau site

StrikerRUS:
Aghhh!..

Now libsvm site is down.
I remember it was already happened. Need to wait they repair.


StrikerRUS:
libsvm site is alive now. Can be merged.

PR #1433, opened 2018-06-09
Revert "[ci] simplified and speeded up Travis"

guolinke:
Reverts Microsoft/LightGBM#1429


guolinke:
@StrikerRUS you can add these changes back.

PR #1435, opened 2018-06-09
updated Boost Compute submodule

StrikerRUS:
@huanzhang12 Is it safe to use the latest `compute` submodule in LightGBM? The current one is one year old and 100 commits behind.


huanzhang12:
@StrikerRUS Thank you for updating the boost compute module! 

In fact boost compute should already be included officially in libboost>=1.61 according to the release note: https://www.boost.org/users/history/version_1_61_0.html, but the version on git usually contains latest bug fixes. Also with this submodule included we can compile on libboost>=1.56.

I think it is safe to merge this.

guolinke:
@StrikerRUS you can fix the conflict then merge.

StrikerRUS:
@huanzhang12 Thank you for the inside info!

PR #1443, opened 2018-06-11
[ci] update AMD APP SDK version

StrikerRUS:
I've uploaded a new version of AMD APP SDK, as pointed by @huanzhang12 in https://github.com/Microsoft/LightGBM/issues/1342#issuecomment-396119206 .
But something seems to be wrong: https://travis-ci.org/Microsoft/LightGBM/jobs/390738317#L938

Can you guys help me please? Or maybe it's not worth it to spend time on making it works?..

### DIFF
--- a/.travis/setup.sh
+++ b/.travis/setup.sh
@@ -26,8 +26,9 @@ conda update -q conda
 
 if [[ $TASK == "gpu" ]] && [[ $TRAVIS_OS_NAME == "linux" ]]; then
     wget https://github.com/Microsoft/LightGBM/releases/download/v2.0.12/AMD-APP-SDKInstaller-v3.0.130.136-GA-linux64.tar.bz2
-    tar -xjf AMD-APP-SDK*.tar.bz2 || exit -1
+    tar -xjf AMD-APP-SDK*.tar.bz2
     mkdir -p $OPENCL_VENDOR_PATH
-    sh AMD-APP-SDK*.sh --tar -xf -C $AMDAPPSDK || exit -1
-    echo libamdocl64.so > $OPENCL_VENDOR_PATH/amdocl64.icd || exit -1
+    sh AMD-APP-SDK*.sh --tar -xf -C $AMDAPPSDK
+    mv $AMDAPPSDK/lib/x86_64/sdk/* $AMDAPPSDK/lib/x86_64/
+    echo libamdocl64.so > $OPENCL_VENDOR_PATH/amdocl64.icd
 fi
--- a/.travis/setup.sh
+++ b/.travis/setup.sh
@@ -28,6 +28,6 @@ if [[ $TASK == "gpu" ]] && [[ $TRAVIS_OS_NAME == "linux" ]]; then
     wget https://github.com/Microsoft/LightGBM/releases/download/v2.0.12/AMD-APP-SDKInstaller-v3.0.130.136-GA-linux64.tar.bz2
     tar -xjf AMD-APP-SDK*.tar.bz2 || exit -1
     mkdir -p $OPENCL_VENDOR_PATH
-    sh AMD-APP-SDK*.sh --tar -xf -C $AMDAPPSDK
-    echo libamdocl64.so > $OPENCL_VENDOR_PATH/amdocl64.icd
+    sh AMD-APP-SDK*.sh --tar -xf -C $AMDAPPSDK || exit -1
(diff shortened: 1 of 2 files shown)


.travis/setup.sh
```diff
@@ -25,9 +25,9 @@ conda config --set always_yes yes --set changeps1 no
...
 if [[ $TASK == "gpu" ]] && [[ $TRAVIS_OS_NAME == "linux" ]]; then
-    wget https://github.com/Microsoft/LightGBM/releases/download/v2.0.12/AMD-APP-SDKInstaller-v3.0.130.135-GA-linux64.tar.bz2
-    tar -xjf AMD-APP-SDK*.tar.bz2
+    wget https://github.com/Microsoft/LightGBM/releases/download/v2.0.12/AMD-APP-SDKInstaller-v3.0.130.136-GA-linux64.tar.bz2
+    tar -xjf AMD-APP-SDK*.tar.bz2 || exit -1
     mkdir -p $OPENCL_VENDOR_PATH
-    sh AMD-APP-SDK*.sh --tar -xf -C $AMDAPPSDK
-    echo libamdocl64.so > $OPENCL_VENDOR_PATH/amdocl64.icd
+    sh AMD-APP-SDK*.sh --tar -xf -C $AMDAPPSDK || exit -1
+    echo libamdocl64.so > $OPENCL_VENDOR_PATH/amdocl64.icd || exit -1
```
huanzhang12: It seems OPENCL_VENDOR_PATH is not defined

huanzhang12:
`setup.sh` seems to miss many necessary steps for GPU setup. You can try to fix it according to an earlier revision of `test.sh`: https://github.com/Microsoft/LightGBM/blob/7f7bc44775d61367ce827ec4215e2a9af7b50747/.travis/test.sh

.travis/setup.sh
huanzhang12: Never mind, it seems they are moved to `.travis.yml`

StrikerRUS:
Yeah, sorry, I haven't ping you when proposed these changes.
GPU installation is in `setup.sh` now (that is logically more correct), and all exports are in `.travis.yml` now because they are not saved between different `*.sh`s:

https://github.com/Microsoft/LightGBM/blob/d04648c9ab54bdcb5453543f0486672b0379f98e/.travis/setup.sh#L27-L33
https://github.com/Microsoft/LightGBM/blob/d04648c9ab54bdcb5453543f0486672b0379f98e/.travis.yml#L47-L49

huanzhang12:
@StrikerRUS I see. Thanks for reorganizing these travis files, they look much better then before :)
I am investigating this issue now.

StrikerRUS:
@huanzhang12 Thanks!

Maybe the new package is simply broken?..
Since they have dropped the official support, they can do anyting, even providing broken packages on Dropbox 😄 .

huanzhang12:
It seems an additional step is need for this version of AMD APP SDK:
```
mv $AMDAPPSDK/lib/x86_64/sdk/* $AMDAPPSDK/lib/x86_64/
```
You can add it after line 31 in `setup.sh` and see if it works.

StrikerRUS:
Wow, perfect!

Thank you very much for your time!

huanzhang12:
@StrikerRUS The OpenCL tests seem to work now. Thanks for working on fixing it!

During the code review I found another thing which might also need a fix (although it is harmless for now)

https://github.com/Microsoft/LightGBM/blob/361548717922b9cd6f099ad0582a84ddfd86ff30/.travis/test.sh#L72

This `sed` changes the default device from cpu to gpu in `config.hpp`, so that all following tests will run on GPU by default. However, if `config.hpp` is changed and the string `std::string device_type = "cpu"` is not found, sed will not give any warning and all GPU tests become CPU tests. This is dangerous since there is no warning that GPU code is not acutally tested.

@StrikerRUS Can you also add a `grep` after this `sed` to make sure the `sed` command indeed replaced the string? If not, just exit with `-1` and we can investigate the test failure manually.

```
grep -q 'std::string device_type = "gpu"' $TRAVIS_BUILD_DIR/include/LightGBM/config.h || exit -1
```

StrikerRUS:
@huanzhang12 Sure! Good catch!

>However, if config.hpp is changed and the string std::string device_type = "cpu" is not found, sed will not give any warning and all GPU tests become CPU tests

The described case has been already happened after #1381, but GPU tests were completely disabled due to miss of AMD APP SDK and that didn't make any sense.

PR #1451, opened 2018-06-14
Update Advanced-Topics.rst

xmpy:
I guess that in LightSVM format unshown should be treated as missing, zero should be still a zero.


guolinke:
I think the original description is right. If zero_as_missing=false, the NaN needs to be explicitly set.

PR #1465, opened 2018-06-21
[python] dont format string values with precision

JasonTam:
I was running into this:
`ValueError: Unknown format code 'f' for object of type 'str'`
for string values when setting precision in `create_tree_digraph`

### DIFF
--- a/python-package/lightgbm/plotting.py
+++ b/python-package/lightgbm/plotting.py
@@ -269,7 +269,7 @@ def _to_graphviz(tree_info, show_info, feature_names, precision=None,
 
     def float2str(value, precision=None):
         return "{0:.{1}f}".format(value, precision) \
-                if (precision is not None) and (not isinstance(value, str)) else str(value)
+            if (precision is not None) and (not isinstance(value, str)) else str(value)
 
     def add(root, parent=None, decision=None):
         """recursively add node or edge"""
--- a/python-package/lightgbm/plotting.py
+++ b/python-package/lightgbm/plotting.py
@@ -268,7 +268,8 @@ def _to_graphviz(tree_info, show_info, feature_names, precision=None,
         raise ImportError('You must install graphviz to plot tree.')
 
     def float2str(value, precision=None):
-        return "{0:.{1}f}".format(value, precision) if precision is not None else str(value)
+        return "{0:.{1}f}".format(value, precision) \
+                if (precision is not None) and (not isinstance(value, str)) else str(value)
 
     def add(root, parent=None, decision=None):
         """recursively add node or edge"""
--- a/python-package/lightgbm/plotting.py
+++ b/python-package/lightgbm/plotting.py
@@ -269,7 +269,7 @@ def _to_graphviz(tree_info, show_info, feature_names, precision=None,
 
     def float2str(value, precision=None):
         return "{0:.{1}f}".format(value, precision) \
-            if (precision is not None) and not isinstance(value, string_type) else str(value)
+            if precision is not None and not isinstance(value, string_type) else str(value)
 
     def add(root, parent=None, decision=None):
         """recursively add node or edge"""
--- a/python-package/lightgbm/plotting.py
+++ b/python-package/lightgbm/plotting.py
@@ -10,7 +10,7 @@ from io import BytesIO
 import numpy as np
 
 from .basic import Booster
-from .compat import MATPLOTLIB_INSTALLED, GRAPHVIZ_INSTALLED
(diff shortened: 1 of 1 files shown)


StrikerRUS:
@JasonTam Hi! Thanks for creating PR.

Can you please provide the example in which you got described error?

JasonTam:
Sure, here is a minimal example:
```
import pandas as pd
import numpy as np
import lightgbm as lgb
np.random.seed(0)

x = pd.concat([
    pd.Series(np.random.choice(['a', 'b'], size=(100)), dtype='category'),
    pd.Series(np.random.randint(0, 2, size=100)),
], axis=1)
y = np.random.randint(0, 2, size=100)
ds = lgb.Dataset(x, y)

params = {'min_child_samples': 2}
gbm = lgb.train(params, ds)
lgb.create_tree_digraph(gbm, precision=2)
```

Dunno when I'll have time to look at the test

StrikerRUS:
Thanks for the example!

The error is caused by the scientific notation, e.g. `1.0000000180025095e-35`. However, we are sill able to convert this value to float with specified precision instead of silent ignoring `precision` argument.

I think that `float2str` could be rewritten as follows:
```
    def float2str(value, precision=None):
        if precision is None:
            return str(value)
        else:
            try:
                return "{0:.{1}f}".format(float(value), precision)
            except TypeError:
                warnings.warn('Some notification that precison cannot be set to some values')
                return str(value)
```

Now it works correctly with scientific notation and notifies user that `precision` cannot be set in case of strange values.

JasonTam:
I thought the error is caused by categorical values? The split threshold would be on a set of categories ie: `"0||1||2"` which is a string. In my fix, I ignore precision if the value is of type string. Not sure if scientific notation is related to this.

StrikerRUS:
LightGBM cannot handle strings at all.

JasonTam:
Im not talking about the categorical values or codes of the dataset itself. I'm talking about the threshold that will be shown in the plot. It's a string if the split is based on set membership.

StrikerRUS:
Threshold cannot be string as well.

JasonTam:
Idk man, when I use categorical data, `value` becomes a string when plotting. You can try my example with a debugger and take a look at `value` and see what I'm talking about

StrikerRUS:
Just run your example completely without `float2str` function:

tree

StrikerRUS:
Please provide the example and the resulted graph with strings you're talking about.

JasonTam:
Not sure exactly what you're asking for. 
With my example on master branch, I can't create the graph -- it's going to error out with the following:
```python
ValueError: Unknown format code 'f' for object of type 'str'
> /Users/jtam/tools/LightGBM/python-package/lightgbm/plotting.py(271)float2str()
    269 
    270     def float2str(value, precision=None):
--> 271         return "{0:.{1}f}".format(value, precision) if precision is not None else str(value)
    272 
    273     def add(root, parent=None, decision=None):

ipdb> value
'1'
```

If I apply my PR, I will get the following:
image

The `1` in the top node is actually a string (note how it's been skipped in the precision formatting , where all other values actually have precision=2).

Could it perhaps be my python or pandas version? -- not sure
(Python 3.6.0  pandas 0.22.0)

StrikerRUS:
Oh, I see. I thought you're having the threshold value like this one `"0||1||2"`, but you're talking about the one element from this set `"1"`.

JasonTam:
I actually do have thresholds like `"0|1|2"` , but for a private dataset. If I have time, I'll try to write a small example with it too.
screen shot 2018-06-21 at 5 14 57 pm

python-package/lightgbm/plotting.py
```diff
@@ -268,7 +268,8 @@ def _to_graphviz(tree_info, show_info, feature_names, precision=None,
         raise ImportError('You must install graphviz to plot tree.')
 
     def float2str(value, precision=None):
-        return "{0:.{1}f}".format(value, precision) if precision is not None else str(value)
+        return "{0:.{1}f}".format(value, precision) \
+            if (precision is not None) and (not isinstance(value, str)) else str(value)
```
StrikerRUS: Please remove the brackets and use `string_type` from `compat.py` module instead of `str` in `isinstance()`.

StrikerRUS:
Please remove these brackets too and I'll merge.
```
(precision is not None)
```

JasonTam:
I think the travis error is unrelated?

StrikerRUS:
Thank you very much for your contribution and patience!

>I think the travis error is unrelated?

Yeah, I've restart the test.

Issue #1466, opened 2018-06-21
Choice of `feature_importance_` in sklearn API


mlisovyi:
I would like to ask to add a new argument to `lightgbm.LGBMModel` to be able to choose the feature importance type that is returned by `lightgbm.LGBMModel.feature_importance_`. 

Right now it is always the default of `lighgbm.Booster.feature_importance()`, that defaults to `split`, but that has undesired behavior in particular for categorical features treated internally in lightgbm, see [this kaggle kernel](https://www.kaggle.com/mlisovyi/beware-of-categorical-features-in-lgbm?scriptVersionId=4070588) for an illustration. In contrast, `gain` behaves better in such case.

Of course, one can use `lightgbm.LGBMModel.booster_.feature_importance(importance_type='gain')`, but this is not sufficient, if one wants to use feature selection algorithms in sklearn (those use `model.feature_importances_` to sorf features, see [RFE](http://scikit-learn.org/stable/modules/generated/sklearn.feature_selection.RFE.html#sklearn.feature_selection.RFE), as an example).

P.S. I'm aware of SHAP, which is discussed in #1350. But it is slow at the moment and one needs to write a derived wrapper class to implement `model.feature_importances_` as mentioned above.

StrikerRUS:
Good idea about adding the `importance_type` argument!

Does new argument should be added to `__init__()`?

StrikerRUS:
Please clarify this moment about `model.importance_type`.

>feature selection algorithms in sklearn (those use model.importance_type to sorf features, see RFE, as an example).

Scikit-learn estimators don't have this attribute. Related issue: https://github.com/scikit-learn/scikit-learn/issues/7926.

mlisovyi:
My proposal was to have an attribute of `lgbm.LGBMModel.importance_type`, that would be set in constructor. That attribute would be passed into the function [here](https://github.com/Microsoft/LightGBM/blob/c803b5cbca7b53ad83b473ffa1e1ed0436015e5c/python-package/lightgbm/sklearn.py#L613). This way the user would be able to configure what is to be stored in `model.feature_importances_`.

Apologies, there was a typo of `model.importance_type` in place of `model.feature_importances_` in the original issue. I will edit it

StrikerRUS:
Thanks for the comment. Now everything is clear.

Would you mind working on it?

mlisovyi:
sure, i will do it tomorrow.

PR #1470, opened 2018-06-22
[python] Configure choice of `feature_importance_` in sklearn API 

mlisovyi:
This is a PR following the discussion in #1466. Docs were updated according to the changes. Tagging @StrikerRUS 

A small additional change was to include vim temporary files into gitignore


python-package/lightgbm/sklearn.py
```diff
@@ -193,6 +193,8 @@ def __init__(self, boosting_type="gbdt", num_leaves=31, max_depth=-1,
             Number of parallel threads.
         silent : bool, optional (default=True)
             Whether to print messages while running boosting.
+        importance_type : str, optional (default='split')
+            The type of feature importance to be filled into ``feature_importances_``.
```
StrikerRUS: Please add the description of possible values here.
https://github.com/Microsoft/LightGBM/blob/707ee62ce2a21a7feff55ba78a648fd4c79c65f4/python-package/lightgbm/basic.py#L1875-L1876

python-package/lightgbm/sklearn.py
```diff
@@ -606,11 +610,13 @@ def feature_importances_(self):
         Note
         ----
         Feature importance in sklearn interface used to normalize to 1,
-        it's deprecated after 2.0.4 and same as Booster.feature_importance() now.
+        it's deprecated after 2.0.4 and is the same as Booster.feature_importance() now.
+        ``importance_type`` attribute is passed to the function to configure the
```
StrikerRUS: Please add some explanation from where `importance_type` appears.

``importance_type`` attribute **specified in the constructor** is passed

python-package/lightgbm/sklearn.py
mlisovyi: Added the two options and a reference to the Booster method (to avoid duplication of information). Let me know, if this is not sufficient

StrikerRUS: Good catch about the duplication of information! However, I suppose that users are not so patient to search another docstring (btw, in different module) to get the parameter description.
So, I think it's better just to copy-paste.
BTW, scikit-learn provides even large examples in docstrings to help people understand what's going on right now and right here 😃

StrikerRUS:
pylint is still red:
image

mlisovyi:
The docs were updated and the training space (my archenemy! :rofl: ), that was upsetting travis, was removed.

One idea came to my mind: is the `specified in the constructor` clarification  maybe misleading? Because one can also set any attribute via `model.set_params()` interface. And in fact a user might find it very handy to do that- he can do analysis of the different importance values fitting a model once and then extracting `model.feature_importances_` for difference importance choices.

For the test suggestion. Could you elaborate on what would be the final goal of it? do we want to involve RCE or maybe just a dump of top-N features with the two importance choices? I'm not sure if that will be easy, as this needs particular dataset (i'm not sure what is the required dataset qualities, but i assume a categorical feature with a large number of classes is a must).

StrikerRUS:
Speaking about the place for `importance_type`, I agree that setting this in constructor is rather misleading. Unfortunately, we cannot add this parameter directly in `feature_importances_`. I believe that `set_params` should works fine. I mean,
```
model = LGBMModel()
model.fit()
model.set_params(importance_type='gain')
model.feature_importances_
```
computes importance according to `gain`. Am I wrong?..


To speak about the test, actually any test is appreciated 😃 . It should check that different `importance_type` leads to different results of `feature_importances_`. Involving any sklearn feature selection tool will give us confidence that it works in probably most common cases in which users passes models to tools with sklearn feature selection interface.

mlisovyi:
You are right that the setter will work, I'm just in doubt if the comment in the documentation is not mis-leading, as a user might imagine that one needs to instantiate and train a new model to compare different choices (since we mention that the config is set **in the constructor**). Maybe it is self-evident and I overreact.

I think, we can have a basic check if the two arrays of feature importances is identical or not. This should be straightforward that implement and it would be triggered by a slightest change between two modes, thus we do not need to hand-craft a dataset

StrikerRUS:
Yeah, got it!

Maybe it's really better to remove my clarification about the constructor... I thought that it will prevent users from doing such things:
```
model.feature_importances_(importance_type='gain')
```

OK, this test is fine. Could you please write it?

mlisovyi:
sure, will do in the next couple of days

StrikerRUS:
ping @mlisovyi 
This is rather useful enhancement, let's finish the PR!

mlisovyi:
sorry, I was overwhelmed with other duties in the past few weeks and also will be in the upcoming week. I can start working on it in 1 week (16-20 July)

StrikerRUS:
Sure, no problem!
It was just to make sure you haven't abandoned or forgotten about this.

mlisovyi:
hi @StrikerRUS, it seems i have some time to work on the test this week. Could you mention if there is a script/tool to execute all defined (or a subset) of tests? Or how does one typically run those that are already in place?

StrikerRUS:
Hi @mlisovyi ! Great!

We run tests by `pytest` tool. You can install it locally by simple `pip install pytest`. And then run a file or folder with tests:
https://github.com/Microsoft/LightGBM/blob/9187c75fbb85c93304183b16247bece2efd0d72b/.travis/test.sh#L56

You can pass only one file you are needed in (more precise: `test_sklearn.py`).

mlisovyi:
@StrikerRUS I've added a draft version of the test and also dropped the mention of constructor setting of the importance parameter, as we have discussed some time ago. Let me know what do you think. Thanks in advance

StrikerRUS:
@mlisovyi Many thanks for implementing this enhancement! I hope it'll be useful thing.

The last remaining thing is pylint. Please fix it.
```
./tests/python_package_test/test_sklearn.py:179:29: E221 multiple spaces before operator
./tests/python_package_test/test_sklearn.py:179:57: E241 multiple spaces after ','
./tests/python_package_test/test_sklearn.py:181:1: W293 blank line contains whitespace
```

Unfortunately, at present something weird is happening with Travis macOS builds (https://github.com/travis-ci/travis-ci/issues/9865) and due to this I'm not able to merge this PR before all checks get green.

mlisovyi:
Travis failed, but it seems to be unrelated to the new additions.

mlisovyi:
Who could rerun the travis test or manually merge the PR?

Issue #1264, opened 2018-03-14
Proposal: Support creating dataset from multiple input arrays


slon:
Hi,

I would like to discuss the problem I have experienced and propose the solution.

I'm using LightGBM for a dataset about 100G in size (in float32 numpy format). Dataset is generated in memory just before training and passed to lightgbm in a form of one huge numpy matrix. What way I'm able to avoid writing 100G to disk, which would be a lot slower.

Dataset is generated using 24 threads. Each thread writing its part of a dataset into private numpy array.
LightGBM expects single numpy array as input, so I have to concatenate 24 parts using `numpy.concatenate`. That is sequential operation running on a single core.

I would like to eliminate this bottleneck and pass dataset a list of numpy arrays.

# Proposal

1. Extend C api with additional function.

```c
int LGBM_DatasetCreateFromMats(int32_t nmats,
                               const void** data,
                               int data_type,
                               int32_t* nrow,
                               int32_t* ncol,
                               int is_row_major,
                               const char* parameters,
                               const DatasetHandle reference,
                               DatasetHandle* out);
```

It works just like [`LGBM_DatasetCreateFromMat`](https://github.com/Microsoft/LightGBM/blob/master/include/LightGBM/c_api.h#L198), but accepts list of dense matrices.

2. Extend python `lightgbm.Dataset` class to take advantage that new API.

guolinke:
@slon Actually, c api side already support this:
1. use sampled row to construct a Dataset, by calling LGBM_DatasetCreateFromSampledColumn
2. use LGBM_DatasetPushRows to adding new rows to the Dataset

All we need is the implementation in python-package side.

contribution is very welcomed : )

slon:
@guolinke Would you clarify some moments for me, please?

> 1. use sampled row to construct a Dataset, by calling LGBM_DatasetCreateFromSampledColumn

That step is calculating histogram buckets? So I need to sample rows for this step evenly from the whole dataset?

> All we need is the implementation in python-package side.

What the interface should look like? How about just adding another type switch [here](https://github.com/Microsoft/LightGBM/blob/master/python-package/lightgbm/basic.py#L713)?

guolinke:
@slon yeah, first step is used to construct histogram buckets. You need to sample evenly from full dataset.

It uses CSR format, to save the memory. https://github.com/Microsoft/LightGBM/blob/master/python-package/lightgbm/basic.py#L709-L710

slon:
[...]

PR #1474, opened 2018-06-24
Support creating Dataset from list of matrices

slon:
Previous discussion in https://github.com/Microsoft/LightGBM/issues/1264

I was able to achieve significant speedup and memory savings after applying this patch.

In the earlier discussion @guolinke suggested, that I should use LGBM_DatasetCreateFromSampledColumn API and perform sampling in python. I tried that approach https://github.com/slon/LightGBM/tree/create_from_hist , but it resulted in code duplication between python/c++ and suboptimal performance.

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -712,6 +712,8 @@ class Dataset(object):
             self.__init_from_csc(data, params_str, ref_dataset)
         elif isinstance(data, np.ndarray):
             self.__init_from_np2d(data, params_str, ref_dataset)
+        elif isinstance(data, list) and len(data) > 0 and all(isinstance(x, np.ndarray) for x in data):
+            self.__init_from_list_np2d(data, params_str, ref_dataset)
         else:
             try:
                 csr = scipy.sparse.csr_matrix(data)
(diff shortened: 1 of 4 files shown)


guolinke:
@slon How this solution can save memory ? 
It seems you need to put all matrices into memory at the same time.

python-package/lightgbm/basic.py
```diff
@@ -712,6 +712,8 @@ def _lazy_init(self, data, label=None, reference=None,
             self.__init_from_csc(data, params_str, ref_dataset)
         elif isinstance(data, np.ndarray):
             self.__init_from_np2d(data, params_str, ref_dataset)
+        elif isinstance(data, list) and len(data) > 0 and all(isinstance(x, np.ndarray) for x in data):
```
StrikerRUS: Speaking about Python side, and except the remark about the memory saving, maybe not only `list` of numpy arrays, but any indexable and iterable object?
```
isinstance(data, list)
```

src/c_api.cpp
```diff
@@ -483,22 +504,39 @@ int LGBM_DatasetCreateFromMat(const void* data,
     omp_set_num_threads(config.num_threads);
   }
   std::unique_ptr<Dataset> ret;
-  auto get_row_fun = RowFunctionFromDenseMatric(data, nrow, ncol, data_type, is_row_major);
+  int32_t total_nrow = 0;
+  for (int j = 0; j < nmat; j++) {
```
StrikerRUS: It seems that LightGBM uses `++j` (prefix) style, but it's better to ask @guolinke.

slon:
@guolinke yes, but without this patch I need 2x memory to concatenate them into a one big array, before creating dataset.

guolinke:
@slon it is not possible to concat one-by-one ? if one matrix is concated, you can delete and free it.

slon:
@guolinke sure, but that would not be very fast.

src/c_api.cpp
```diff
@@ -483,22 +504,39 @@ int LGBM_DatasetCreateFromMat(const void* data,
...
       auto idx = sample_indices[i];
-      auto row = get_row_fun(static_cast<int>(idx));
-      for (size_t j = 0; j < row.size(); ++j) {
-        if (std::fabs(row[j]) > kZeroThreshold || std::isnan(row[j])) {
-          sample_values[j].emplace_back(row[j]);
-          sample_idx[j].emplace_back(static_cast<int>(i));
+      while ((idx - offset) >= nrow[j]) {
+        offset += nrow[j];
+        j++;
+      }
```
guolinke: it seem this is inefficient. you can do it via math calculation, instead of loop.

src/c_api.cpp
guolinke: yeah, ++j style

src/c_api.cpp
```diff
@@ -507,22 +545,27 @@ int LGBM_DatasetCreateFromMat(const void* data,
...
   }
-  OMP_INIT_EX();
-  #pragma omp parallel for schedule(static)
-  for (int i = 0; i < nrow; ++i) {
-    OMP_LOOP_EX_BEGIN();
-    const int tid = omp_get_thread_num();
-    auto one_row = get_row_fun(i);
-    ret->PushOneRow(tid, i, one_row);
-    OMP_LOOP_EX_END();
+  int start_row = 0;
```
chivee: int32_t

python-package/lightgbm/basic.py
slon: We might use [collections.abc.Sequence](https://docs.python.org/3/library/collections.abc.html#collections.abc.Sequence), but it is supported only in python >3.3.

Maybe `instance(data, (list, tuple))` is going to be good enough? Or just leave it as is.

slon:
Fixed code style.

Rebased branch in order to fix GPU tests on travis.

python-package/lightgbm/basic.py
StrikerRUS: To be honest, I'm worrying most of all about numpy array. So, `list, tuple,  np.ndarray` is enough, I suppose.

slon: Is it possible to create `ndarray` of `ndarray`-s?

StrikerRUS: Seems to be that not. You're right. I thought that it's possible.
Then `list` and `tuple` will be enough.

PR #1479, opened 2018-06-29
Revert "add simple definition to vsts-ci"

guolinke:
Reverts Microsoft/LightGBM#1478


StrikerRUS:
Should the empty `.vsts-ci.yml` file be also removed?

Issue #1453, opened 2018-06-15
user defined eval metric function has different data interface


zhangfiona:
Parameter datatypes are different when user defined eval_metric function is used in train function of training api and fit function of scikit-learn api. And the sample code is as follows:

```
# self-defined eval metric
# f(preds: array, train_data: Dataset) -> name: string, value: array, is_higher_better: bool
# binary error
def eval_metric(preds, train_data):
    labels = train_data.get_label()
    return 'error', np.mean(labels != (preds > 0.5)), False

# Self-defined eval_metric is used in fit function of scikit-learn api, which causes an error
clf = lgb.LGBMClassifier()
clf.fit(X, y, eval_set=[[X,y]], eval_metric=eval_metric, early_stopping_rounds=5)

# Self-defined eval_metric is used in train function of training api, which executes sucessfully
data = lgb.Dataset(data=X, label=y)
lgb.train(params=params, train_set=data, feval=eval_metric, early_stopping_rounds=5, valid_sets=[data])
 ```
Self-defined eval_metric function is _not universal_ in these two train methods. The problem exists in **_eval_function_wrapper(func) in sklearn.py**, and this decorator changes the parameter datatypes. 
The code in sklean.py is here: [https://github.com/Microsoft/LightGBM/blob/master/python-package/lightgbm/sklearn.py#L113](https://github.com/Microsoft/LightGBM/blob/master/python-package/lightgbm/sklearn.py#L113)

guolinke:
This design is from xgboost, to make sure the smooth immigration.

tobycheese:
Why does your eval_metric take the parameters in the order (preds, train_data)? The docs state the opposite order:

> Custom eval function expects a callable with following functions: func(y_true, y_pred)

_eval_function_wrapper(func) seems to swap the arguments, I think your way is correct and the docs are wrong. This is confusing...

zhangfiona:
[...]

PR #1503, opened 2018-07-11
[python][examples] Added example of custom feval with sklearn wrapper

StrikerRUS:
Addressed to #1453.


StrikerRUS:
Examples could be run by the same command on both Windows and Linux

StrikerRUS:
`conf` files had been written years ago and contains outdated information in many cases (e.g. possible values of parameter). I've added a note about it.

PR #1325, opened 2018-04-18
Metric doc update

mlisovyi:
This is an update of the python package docs as well as of parameter description attempting to improve the info on metric usage in evaluation. 
The PR follows the discussion in #1318

Issue #1318, opened 2018-04-16
[python] Reported evaluation metric and early stop


mlisovyi:
The issue that I face is: when one runs with the early stopping enabled, one aims to be able to **stop specifically on the eval_metric metric**. However, python API of LightGBM checks __all__ metrics that are monitored. This is different from the XGBoost choice, where they check the last item from the eval list, but this is also a justifiable choice. However, this choice should not prevent us from reaching the aforementioned goal.

There is some undocumented feature of the code, related to the list of evaluation metrics that are considered. Namely, if one uses any predefined metric as the objective (with the corresponding name string), then the objective metric is prepended to the list of evaluation metrics. Which is unexpected on the user side and not desired in terms of the goal defined before.  If one uses a custom objective function, then there no such extra metric prepended. 

Is there a chance to either have the early-stopping criterion to be configurable (with an extra argument in the fit function) or to drop the objective that appears in the evaluation result and that is not documented anywhere?

### Environment
I observe this feature with lightgbm v2.1.0.

### How to reproduce
Here is a   basic toy example:
```python
import numpy as np
import lightgbm as lgb

# generate random data
X = np.random.randn(10000,5)
y = np.random.randn(10000)

# define MSE evaluation and objective functions (for the sake of example)
def evaluate_mse(truth, prediction):
    return('xxx', np.mean(np.square(truth - prediction)), False)

def obj_mse(truth, prediction):
    grad = 2 * (prediction - truth)
    hess = 2 * np.ones(shape=truth.shape)
    return(grad, hess)

clf = lgb.LGBMRegressor()
# Train with the default 'mse' objective
clf.set_params(**{'objective': 'mse'})
clf.fit(X, y, 
        eval_set=[(X,y)],
        eval_names=['validation_0'],
        eval_metric=evaluate_mse,
        verbose=50)
# Train with the custom implementation of the MSE
clf.set_params(**{'objective': obj_mse})
clf.fit(X, y, 
        eval_set=[(X,y)],
        eval_names=['validation_0'],
        eval_metric=evaluate_mse,
        verbose=50)
```
In the example above there is no early stopping, but you can already see the mismatch in the monitored metrics

guolinke:
you can try the latest master branch with following code:
```
import numpy as np
import lightgbm as lgb

# generate random data
X = np.random.randn(10000,5)
y = np.random.randn(10000)

# define MSE evaluation and objective functions (for the sake of example)
def evaluate_mse(truth, prediction):
[...]

PR #1505, opened 2018-07-12
fix custom metric for multiclass

guolinke:


src/io/config.cpp
```diff
@@ -215,12 +215,14 @@ void Config::CheckParamConflict() {
     }
   }
   for (std::string metric_type : metric) {
+    bool metric_custom = metric_type == std::string("none") || metric_type == std::string("null") || metric_type == std::string("custom");
```
StrikerRUS: I think this should be consistent with `aliases` here:
https://github.com/Microsoft/LightGBM/blob/1e53445e85fe7d9597a274e4ddc1b1af190d0a64/include/LightGBM/config.h#L662

StrikerRUS:
@guolinke I've spent some time for fun with metrics and it seems to be that this clarification
https://github.com/Microsoft/LightGBM/blob/1e53445e85fe7d9597a274e4ddc1b1af190d0a64/include/LightGBM/config.h#L660
(the part `**in addition** to what is provided in the training arguments`)
discussed here #1318 and done here #1325 is true only for case of **custom metric**. I mean, only custom metric is **appended** to the default one, while built-in metric **overrides** default one. Or I'm missing something...

```
import numpy as np
import lightgbm as lgb

X = np.random.randn(10000,5)
y = np.random.randn(10000)

def evaluate_mse(truth, prediction):
    return('xxx', np.mean(np.square(truth - prediction)), False)

clf = lgb.LGBMModel(objective='mse')
clf.fit(X, y, 
        eval_set=[(X,y)],
        eval_metric='rmse')  # built-in ovverrides


clf = lgb.LGBMModel(objective='mse')
clf.fit(X, y, 
        eval_set=[(X,y)],
        eval_metric=evaluate_mse)  # custom is appended
```

```
[1]	valid_0's rmse: 1.00529
[2]	valid_0's rmse: 1.00349
[3]	valid_0's rmse: 1.00182
[4]	valid_0's rmse: 1.00004
[5]	valid_0's rmse: 0.998544
[6]	valid_0's rmse: 0.997113
[7]	valid_0's rmse: 0.995659
[8]	valid_0's rmse: 0.994117
[9]	valid_0's rmse: 0.992778
[10]	valid_0's rmse: 0.991077
```

```
[1]	valid_0's l2: 1.01061	valid_0's xxx: 1.01061
[2]	valid_0's l2: 1.00699	valid_0's xxx: 1.00699
[3]	valid_0's l2: 1.00365	valid_0's xxx: 1.00365
[4]	valid_0's l2: 1.00007	valid_0's xxx: 1.00007
[5]	valid_0's l2: 0.99709	valid_0's xxx: 0.99709
[6]	valid_0's l2: 0.994235	valid_0's xxx: 0.994235
[7]	valid_0's l2: 0.991337	valid_0's xxx: 0.991337
[8]	valid_0's l2: 0.988269	valid_0's xxx: 0.988269
[9]	valid_0's l2: 0.985609	valid_0's xxx: 0.985609
[10]	valid_0's l2: 0.982234	valid_0's xxx: 0.982234
```

StrikerRUS:
@guolinke Sorry for writing about this here, but I think it's better to solve all problems with metrics at once (in this PR).

It seems that this PR #1373 (followed the discussion started here #1371 ) and the part of this one #1488 should be reverted, because `feval='None'` crashes:

```
import numpy as np
import lightgbm as lgb

X = np.random.randn(10000,5)
y = np.random.randn(10000)
train_set = lgb.Dataset(X, y)

gbm = lgb.cv({'objective': 'rmse'}, stratified=False, train_set=train_set, feval='None', verbose_eval=True)
```

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-34-907ea19253a6> in <module>()
      1 
      2 
----> 3 gbm = lgb.cv({'objective': 'rmse'}, stratified=False, train_set=train_set, feval='None', verbose_eval=True)

C:\Program Files\Anaconda3\lib\site-packages\lightgbm\engine.py in cv(params, train_set, num_boost_round, folds, nfold, stratified, shuffle, metrics, fobj, feval, init_model, feature_name, categorical_feature, early_stopping_rounds, fpreproc, verbose_eval, show_stdv, seed, callbacks)
    450                                     evaluation_result_list=None))
    451         cvfolds.update(fobj=fobj)
--> 452         res = _agg_cv_result(cvfolds.eval_valid(feval))
    453         for _, key, mean, _, std in res:
    454             results[key + '-mean'].append(mean)

C:\Program Files\Anaconda3\lib\site-packages\lightgbm\engine.py in handlerFunction(*args, **kwargs)
    247             ret = []
    248             for booster in self.boosters:
--> 249                 ret.append(getattr(booster, name)(*args, **kwargs))
    250             return ret
    251         return handlerFunction

C:\Program Files\Anaconda3\lib\site-packages\lightgbm\basic.py in eval_valid(self, feval)
   1651             List with evaluation results.
   1652         """
-> 1653         return [item for i in range_(1, self.__num_dataset)
   1654                 for item in self.__inner_eval(self.name_valid_sets[i - 1], i, feval)]
   1655 

C:\Program Files\Anaconda3\lib\site-packages\lightgbm\basic.py in <listcomp>(.0)
   1652         """
   1653         return [item for i in range_(1, self.__num_dataset)
-> 1654                 for item in self.__inner_eval(self.name_valid_sets[i - 1], i, feval)]
   1655 
   1656     def save_model(self, filename, num_iteration=-1):

C:\Program Files\Anaconda3\lib\site-packages\lightgbm\basic.py in __inner_eval(self, data_name, data_idx, feval)
   1924             else:
   1925                 cur_data = self.valid_sets[data_idx - 1]
-> 1926             feval_ret = feval(self.__inner_predict(data_idx), cur_data)
   1927             if isinstance(feval_ret, list):
   1928                 for eval_name, val, is_higher_better in feval_ret:

TypeError: 'str' object is not callable
```

guolinke:
okay, I am busy with other things, will fix this when have time.

guolinke:
@StrikerRUS I think `feval` should be a function.

guolinke:
@StrikerRUS 
I think current behavior is normal. Since the `feval` could be passed dynamic, and `metric` is initialized before training. Thus, user should set `metric=None` to disable the built-in metric.

StrikerRUS:
> Thus, user should set metric='None' to disable the built-in metric.

Yeah! But not `feval='None'`. So docs for `feval` should be corrected.

And could you confirm this https://github.com/Microsoft/LightGBM/pull/1505#issuecomment-404472494 ?

guolinke:
@StrikerRUS yeah, the docs should be corrected. 
The logic is, when `metric` is not set, it will use the name of objective to create a metric as default one.

StrikerRUS:
> The logic is, when metric is not set, it will use the name of objective to create a metric as default one.

Got it!

And it seems that `eval_metric` when built-in (string) overwrites that default one, but when custom (callable) is appended to that default one. Right?

guolinke:
yeah, since the cpp side didn't know there is a custom metric.

StrikerRUS:
@guolinke Do you think that docs fixes should be done in a separate PR?

guolinke:
@StrikerRUS yeah, since it is 'feval', not the build-in metrics.

StrikerRUS:
> yeah, since it is 'feval', not the build-in metrics.

Got it!

src/io/config.cpp
```diff
@@ -215,12 +216,15 @@ void Config::CheckParamConflict() {
     }
   }
   for (std::string metric_type : metric) {
+    bool metric_custom = metric_type == std::string("none") || metric_type == std::string("null") 
+                         || metric_type == std::string("custom") || metric_type == std::string("na");
     bool metric_type_multiclass = (CheckMultiClassObjective(metric_type)
```
StrikerRUS: Maybe better `bool metric_type_multiclass` -> `bool metric_type_multiclass_or_custom`?

src/io/config.cpp
```diff
@@ -215,12 +216,15 @@ void Config::CheckParamConflict() {
...
+    bool metric_custom = metric_type == std::string("none") || metric_type == std::string("null") 
+                         || metric_type == std::string("custom") || metric_type == std::string("na");
     bool metric_type_multiclass = (CheckMultiClassObjective(metric_type) 
-                                    || metric_type == std::string("multi_logloss")
-                                    || metric_type == std::string("multi_error"));
+                                   || metric_type == std::string("multi_logloss")
+                                   || metric_type == std::string("multi_error")
+                                   || metric_custom);
     if ((objective_type_multiclass && !metric_type_multiclass)
       || (!objective_type_multiclass && metric_type_multiclass)) {
```
StrikerRUS: Indent.

I'm doubt about the following situation:
`(!objective_type_multiclass && metric_type_multiclass)` : let say `objective=l2 && metric_type="None"`. Seems that it will trigger fatal error.

guolinke: yeah, good point.

guolinke: fixed

src/io/config.cpp
```diff
@@ -215,12 +216,15 @@ void Config::CheckParamConflict() {
     }
   }
   for (std::string metric_type : metric) {
-    bool metric_type_multiclass = (CheckMultiClassObjective(metric_type) 
-                                    || metric_type == std::string("multi_logloss")
-                                    || metric_type == std::string("multi_error"));
+    bool metric_custom_or_none = metric_type == std::string("none") || metric_type == std::string("null") 
+                         || metric_type == std::string("custom") || metric_type == std::string("na");
```
StrikerRUS: Indent.

src/io/config.cpp
```diff
@@ -215,12 +216,15 @@ void Config::CheckParamConflict() {
...
+    bool metric_type_multiclass = (CheckMultiClassObjective(metric_type)
+                                   || metric_type == std::string("multi_logloss")
+                                   || metric_type == std::string("multi_error")
+                                   || (metric_custom_or_none && num_class_check > 1));
     if ((objective_type_multiclass && !metric_type_multiclass)
       || (!objective_type_multiclass && metric_type_multiclass)) {
```
StrikerRUS: Indent.

guolinke: Thanks

PR #1511, opened 2018-07-12
[docs] fixed typo in parameter list Parameters.rst

Quetzalcohuatl:
fixed typo from contributs to contributes


StrikerRUS:
@guolinke We need some policy for such PRs. Users correct typos in `rst` file but the typo should be fixed in `config.h` file. However, we cannot force contributors do it, because they may have not Python installed on their machines.

What do you think?

guolinke:
@StrikerRUS I think we should let contributors do that, and python environment is common and easy to install.

Quetzalcohuatl:
@guolinke I don't have Python on my machine and mostly use cloud services to do my work. Even then, I would have no idea where the config.h file would be located. I was just reading and noticed this typo. I understand a .h file is supposed to be a header file for C and C++. I am not familiar with the .rst extension and the fact that the header file could contain the definition of the parameters to me is something I have not encountered before (and if it is common practice, it's because my knowledge of Python is not strong enough).

In short, it was very easy for me to click on the Pencil and edit this typo. It would take time for me to learn where specific files are located and then propose changes on those .h files, however. And if I did locate them, I would edit them using a text editor, not within an IDE, because I do not have Python installed. Cheers, though, for both of you on working on this amazing project.

StrikerRUS:
@guolinke I agree with @Quetzalcohuatl . Often users even don't clone the repo and correct typos just online at GitHub document page.

I see two variants:
- force contributors to fix source file with something like: "Hey man! Thanks for your contribution, we really appreciate it, but you should correct the typo in <this file> and then run <this Python script>." (Bad idea because of reasons listed above.)
- close PR with something like this: "Hey man! Thanks for your contribution, we really appreciate it, but your changes should be done in source file." and open new PR with proposed changes. (Bad idea because of that contribution will not be counted by GitHub and it's duplicated work).

guolinke:
okay, another option is editing both 2 files  (config.h, parameter.rst).

StrikerRUS:
@guolinke So, ask to move proposed changes to `.h` file and run script; if it's hard, then ask to manually edit `.h` file too; if there is no response or any other situation, then close PR and fix in new PR. Right?

guolinke:
yeah, and we can regularly run the script to check their consistency.

chivee:
@StrikerRUS , I think it can be fixed in the same PR rather than in new PR.

StrikerRUS:
@chivee 
Clone contributor's repo to own machine, switch to branch with changes, make changes on source file, run script and commit, right?

guolinke:
@StrikerRUS can we add a test for the consistency in CI ? 
Like generate the doc via config.h, and compare it with the old one ?

StrikerRUS:
@guolinke OK. I'll try this

chivee:
@StrikerRUS , Yes, but I think it easier to just 
`git checkout https://github.com/Microsoft/LightGBM/pull/1511`
then `push`

StrikerRUS:
@chivee Thanks! Good to know!

PR #1512, opened 2018-07-12
[docs] updated status badges

StrikerRUS:
Fixed VSTS status badge and added link for the latest build at master branch.

Also reordered badges because of design reasons: VSTS and Appveyor badges have icons and same color pallet and now are shown close to each other.
Result: https://github.com/Microsoft/LightGBM/blob/badges/README.md

Not sure about `definitionId=7` in VSTS URL. I just took this from here:
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,9 @@
 LightGBM, Light Gradient Boosting Machine
 =========================================
 
-[![Build Status](https://lightgbm.visualstudio.com/LightGBM-Build/_apis/build/status/LightGBM-Build-CI&branchName=master)]()
-[![Build Status](https://travis-ci.org/Microsoft/LightGBM.svg?branch=master)](https://travis-ci.org/Microsoft/LightGBM)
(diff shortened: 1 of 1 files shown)


StrikerRUS:
>And what is the decision about Travis and Appveyor after this PR will be merged? Will they be removed or left along with VSTS?

>I think we can keep Travis, since it is very fast.

@guolinke BTW, I think we can not remove Appveyor checks at all, because they are very stable. instead we can find the number of jobs which will finish approximately in the same time as VSTS or Travis ones. For instance, it could be MinGW builds.

guolinke:
@StrikerRUS I think vsts also can build mingw version. But I don't have time to enable it...

StrikerRUS:
@guolinke Yeah, but you said that we are limited by 10 parallel jobs at VSTS, so I think we can benefit from balancing the builds between Appveyor and VSTS. At least, Appveyor can be "11th" parallel build.

I'll do some experiments with the building time and create a PR with new tests allocation between CI services.

PR #1513, opened 2018-07-12
[ci][python] improved paths in setup.py and small CI refactoring

StrikerRUS:
Made paths in `setup.py` more readable and platform independent.

Less `cd` commands at Appveyor and full paths at Travis.

Unfortunately, we still cannot perform installation of Python-package from other directory due to relative paths in `Manifest.in`.


python-package/setup.py
```diff
@@ -32,25 +31,31 @@ def find_lib():
...
-        distutils.file_util.copy_file("../windows/LightGBM.sln", "./compile/windows/LightGBM.sln")
-        distutils.file_util.copy_file("../windows/LightGBM.vcxproj", "./compile/windows/LightGBM.vcxproj")
+        if not os.path.exists(os.path.join(CURRENT_DIR, "compile", "windows")):
+            os.makedirs(os.path.join(CURRENT_DIR, "compile", "windows"))
+        distutils.file_util.copy_file(os.path.join(CURRENT_DIR, os.path.pardir, "windows", "LightGBM.sln"),
+                                      os.path.join(CURRENT_DIR, "compile", "windows", "LightGBM.sln"))
+        distutils.file_util.copy_file(os.path.join(CURRENT_DIR, os.path.pardir, "windows", "LightGBM.vcxproj"),
+                                      os.path.join(CURRENT_DIR, "compile", "windows", "LightGBM.vcxproj"))
+        distutils.file_util.copy_file(os.path.join(CURRENT_DIR, os.path.pardir, "windows", "LightGBM.vcxproj.filters"),
+                                      os.path.join(CURRENT_DIR, "compile", "windows", "LightGBM.vcxproj.filters"))
```
StrikerRUS: @guolinke Do we need to copy this file too?

Refer to https://github.com/Microsoft/LightGBM/commit/7689a4d0647b07cc3357e28e3eb85ae24d7e919a

guolinke: I think `filters` is not needed for build

StrikerRUS:
@guolinke Removed copying of `filters` file and refined paths in nugget creation script.

StrikerRUS:
@guolinke Because you are saying that we don't need `LightGBM.vcxproj.filters`, I've removed it from `MANIFEST.IN` file.

PR #1515, opened 2018-07-12
[docs] updated docs of num_iterations parameter

StrikerRUS:
The current behavior is opposite to one described in docs. If there is an alias in param dict it will overwrite the value of the `num_boost_round` argument of `train`/`cv` method.

https://github.com/Microsoft/LightGBM/blob/5d7eb230a12b8b37e3a40faf4bd769e1563e7f17/python-package/lightgbm/engine.py#L101-L105

```
import numpy as np
import lightgbm as lgb

X = np.random.randn(10000,5)
y = np.random.randn(10000)
train_set = lgb.Dataset(X, y)

gbm = lgb.train({'objective': 'mse', 'num_iterations': 5}, train_set=train_set, valid_sets=[train_set], verbose_eval=True)
```

```
[1]	training's l2: 1.00718
[2]	training's l2: 1.00345
[3]	training's l2: 0.999723
[4]	training's l2: 0.996173
[5]	training's l2: 0.99284

C:\Program Files\Anaconda3\lib\site-packages\lightgbm\engine.py:102: UserWarning: Found `num_iterations` in params. Will use it instead of argument
  warnings.warn("Found `{}` in params. Will use it instead of argument".format(alias))
```

ping @Laurae2 for R-package.


chivee:
Should we wait for R package for completing?

StrikerRUS:
@chivee I think we should wait @Laurae2's answer. I can fix R's description right here in this PR, or maybe he'd like to create a separate new one.

StrikerRUS:
It seems to be that R has the similar logic:
https://github.com/Microsoft/LightGBM/blob/c6cdea75a3f7732a13c21c760c1bf221441a5c79/R-package/R/lgb.cv.R#L142-L148
https://github.com/Microsoft/LightGBM/blob/c6cdea75a3f7732a13c21c760c1bf221441a5c79/R-package/R/lgb.train.R#L119-L125

StrikerRUS:
re-ping @Laurae2

Laurae2:
@StrikerRUS Number of iterations in R is a special parameter, for the R-package the default parameter nrounds is overwritten if another similar parameter is passed. This allows to avoid parsing in/out lists for optimization (the new doc description of this pull request is correct for the behavior of the R-package).

docs/Parameters.rst
```diff
@@ -127,7 +127,7 @@ Core Parameters
 
    -  number of boosting iterations
 
-   -  **Note**: for Python/R-package, **this parameter is ignored**, use ``num_boost_round`` (Python) or ``nrounds`` (R) input arguments of ``train`` and ``cv`` methods instead
+   -  **Note**: for R-package, **this parameter is ignored**, use ``nrounds`` input arguments of ``train`` and ``cv`` methods instead
```
Laurae2: It will overwrite the default wrapper parameter when provided

StrikerRUS: So I can remove the whole line, not only for Python, right?

PR #1547, opened 2018-07-28
fix all negative values in cat features

### LINKED ISSUE #1545: Categorical Features with All Negative Values Cause Segmentation Fault
I found that LightGBM would cause the Segmentation (core dump) when meeting a Categorical feature with all negative values.

## Environment info
Operating System: Windows10
CPU: Intel Xeon E5-2609
LightGBM version: 2.1.2

## Error Message:
```
$ Python‣2.7.15 :: Anaconda, Inc.  
▶ python issue.py                                                                                                                               
/home/admin0/anaconda2/lib/python2.7/site-packages/lightgbm/basic.py:1040: UserWarning: Using categorical_feature in Dataset.                   
  warnings.warn('Using categorical_feature in Dataset.')                                                                                        
[LightGBM] [Warning] Met negative value in categorical features, will convert it to NaN                                                         
[1]    13221 segmentation fault (core dumped)  python issue.py                                                                                  
(base)
```

## Reproducible examples
```
import pandas as pd
import lightgbm as lgb

df = pd.DataFrame({'x':[-2,-2,-2], 'y':[0,0,1]})

train_data = lgb.Dataset(df[['x']], label=df['y'], categorical_feature=['x'])
params = {'learning_rate': 0.15,'boosting_type': 'gbdt','objective': 'binary'}
bst = lgb.train(params, train_data, 10)
```
## Steps to reproduce
execute the above example code on python2.7

I suspect the error handling of categorical doesn't deal with this case (all negative values), and would take a look at the source code later.

Thanks.

guolinke:
fix #1545

### DIFF
--- a/src/io/bin.cpp
+++ b/src/io/bin.cpp
@@ -316,55 +316,58 @@ namespace LightGBM {
           }
         }
       }
-      // sort by counts
-      Common::SortForPair<int, int>(counts_int, distinct_values_int, 0, true);
-      // avoid first bin is zero
-      if (distinct_values_int[0] == 0) {
-        if (counts_int.size() == 1) {
-          counts_int.push_back(0);
-          distinct_values_int.push_back(distinct_values_int[0] + 1);
-        }
-        std::swap(counts_int[0], counts_int[1]);
-        std::swap(distinct_values_int[0], distinct_values_int[1]);
-      }
-      // will ignore the categorical of small counts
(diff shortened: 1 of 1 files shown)


StrikerRUS:
@guolinke I suppose **Question 9** should be paraphrased after this fix applied: https://github.com/Microsoft/LightGBM/blob/master/docs/FAQ.rst#lightgbm

PR #1548, opened 2018-07-28
[python] open files with utf8 encoding during installation

StrikerRUS:
This PR prevents such errors like
https://github.com/RGF-team/rgf/issues/65
https://github.com/henry0312/pytest-codestyle/pull/34

### DIFF
--- a/python-package/setup.py
+++ b/python-package/setup.py
@@ -4,6 +4,7 @@
 from __future__ import absolute_import
 
 import distutils
+import io
 import logging
 import os
 import shutil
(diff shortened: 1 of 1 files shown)


python-package/setup.py
```diff
@@ -4,6 +4,7 @@
 from __future__ import absolute_import
 
 import distutils
+import io
```
henry0312: Do you use `io` for Python 2?

StrikerRUS: You are absolutely right! :-)

PR #1522, opened 2018-07-13
[ci] added tests with Clang/AppleClang compilation at Travis

StrikerRUS:
Compiling CMake from sources is extremely sloooow, but at least it works :-) .

Refer to #1501.

PR #1553, opened 2018-07-31
[docs] Add installation instructions using MacPorts

Mottl:


StrikerRUS:
**Blocked** due to the **conflict** with #1522.

The installation guide is rather large, so I'm not sure that changes `brew` -> `port` desire the own section in the guide. They are unrelated to the LightGBM installation process :-/
I suppose that a note like "replace `brew` command with `port` is you are using MacPorts" will be enough.

chivee:
@StrikerRUS, maybe we can using < details > ?

StrikerRUS:
@chivee I like your idea! Also, we can add `Contents` section for better navigation. I can take this, after merging #1522.

However, I still think that for `brew` -> `port` a note will be enough.

StrikerRUS:
Closed as stale.

PR #1565, opened 2018-08-03
add start_iteration in model saving

guolinke:


python-package/lightgbm/basic.py
```diff
@@ -1710,13 +1710,15 @@ def eval_valid(self, feval=None):
         return [item for i in range_(1, self.__num_dataset)
                 for item in self.__inner_eval(self.name_valid_sets[i - 1], i, feval)]
 
-    def save_model(self, filename, num_iteration=-1):
+    def save_model(self, filename, start_iteration=0, num_iteration=-1):
```
StrikerRUS: I understand that the current order of arguments is logically correct, but it'll break backward compatibility. Maybe it's better to swap the order of `start_iteration` and `num_iteration`?

python-package/lightgbm/basic.py
```diff
@@ -1748,7 +1756,10 @@ def _load_model_from_string(self, model_str, verbose=True):
             print('Finished loading model, total used %d iterations' % (int(out_num_iterations.value)))
         self.__num_class = out_num_class.value
 
-    def _save_model_to_string(self, num_iteration=-1):
+    def dump_model_to_string(self, start_iteration=0, num_iteration=-1):
+        return self._save_model_to_string(start_iteration, num_iteration)
```
StrikerRUS: Maybe it's better just to rename old private function instead of creating new proxy one?

python-package/lightgbm/basic.py
guolinke: good point.

guolinke:
@StrikerRUS to review

python-package/lightgbm/basic.py
```diff
@@ -1721,15 +1721,23 @@ def save_model(self, filename, num_iteration=None):
...
             ctypes.c_int(num_iteration),
             c_str(filename)))
         _save_pandas_categorical(filename, self.pandas_categorical)
 
+    def shuffle_models(self):
+        """Shuffle models
+        """
+        _safe_call(_LIB.LGBM_BoosterShuffleModels(self.handle))
+
     def _load_model_from_string(self, model_str, verbose=True):
```
StrikerRUS: Don't you want to make `_load_model_from_string` publicly available, since `model_to_string` has become public?

python-package/lightgbm/basic.py
```diff
@@ -1721,17 +1721,38 @@ def save_model(self, filename, num_iteration=None):
             Index of the iteration that should be saved.
             If None, if the best iteration exists, it is saved; otherwise, all iterations are saved.
             If <= 0, all iterations are saved.
+        start_iteration: int, optional (default=0)
+            Start index of the iteration that should to saved.
```
StrikerRUS: "... should **be** saved."

python-package/lightgbm/basic.py
```diff
@@ -1721,17 +1721,38 @@ def save_model(self, filename, num_iteration=None):
...
+        """Load Booster from a string.
+
+        Parameters
+        ----------
+        model_str: string
+            Model will be loaded from this string.
+        verbose: Bool, optional (default=True)
+            Set to False to disable log when loading model.
+
+        Returns
```
StrikerRUS: Seems that this method doesn't return anything.

python-package/lightgbm/basic.py
```diff
@@ -1785,6 +1823,8 @@ def dump_model(self, num_iteration=None):
             Index of the iteration that should be dumped.
             If None, if the best iteration exists, it is dumped; otherwise, all iterations are dumped.
             If <= 0, all iterations are dumped.
+        start_iteration: int, optional (default=0)
+            Start index of the iteration that should be dumped
```
StrikerRUS: Dot at the end

python-package/lightgbm/basic.py
```diff
@@ -1721,17 +1721,38 @@ def save_model(self, filename, num_iteration=None):
...
+        _safe_call(_LIB.LGBM_BoosterShuffleModels(self.handle))
+
+    def model_from_string(self, model_str, verbose=True):
+        """Load Booster from a string.
+
+        Parameters
+        ----------
+        model_str: string
+            Model will be loaded from this string.
+        verbose: Bool, optional (default=True)
```
StrikerRUS: `Bool` -> `bool`

guolinke:
@StrikerRUS How about returning `self` in most functions, to support the chained calls ?

StrikerRUS:
@guolinke I think it's good idea!

guolinke:
@StrikerRUS We can do it in another PR.

StrikerRUS:
Yep!

Issue #1566, opened 2018-08-03
Support for .Net4.6 in LightGBM nuget.


codemzs:
Hi @guolinke,

We need a structure like below to also make .net 4.6 work. 

* lightgbm
  * 2.2.2.2
     * net46
       * runtimes
          * lib_lightgbm.dll
           * lightgbm.exe
     * netstandard2.0
        * runtimes
          * linux-x86
            * native
              * lib_lightgbm.so
             * osx-x64
                * native
                  * lib_lightgbm.dylib
             * win-x64
                * native
                  * lib_lightgbm.dll
                  * lightgbm.exe

Let me know once you have the nuget ready and I can verify it on ML.NET side before you push it. 
@https://github.com/dotnet/machinelearning/issues/494
CC: @eerhardt @shauheen @galoshri @tomfinley @sharpwood

sharpwood:
How is progress, can this be released with the ML 0.4 release? @codemzs

eerhardt:
For anyone looking for a work around until this issue gets fixed, see https://stackoverflow.com/a/51728825/3680432.

@guolinke - a possible fix for this issue is to follow the same pattern we use in the `Microsoft.ML` package, which is to add a `.props` file to the package that adds the native assembly when `packages.config` is being used.  (Note we also have a `.targets` file that checks the project is targeting `x64`.)

Feel free to copy this code into your NuGet package:

https://github.com/dotnet/machinelearning/blob/94a3d90cba7787973d1ee300e0a88a30653b2d9a/pkg/Microsoft.ML/build/netstandard2.0/Microsoft.ML.props
https://github.com/dotnet/machinelearning/blob/94a3d90cba7787973d1ee300e0a88a30653b2d9a/pkg/Microsoft.ML/build/netstandard2.0/Microsoft.ML.targets

Put those 2 files in a `build` folder in the nupkg and edit any references to Microsoft.ML -> LightGBM.  Note the name of the files need to match the name of your nuget package.

codemzs:
@sharpwood Unfortunately this won't go in for 0.4 release.

eerhardt:
@guolinke - MSBuild files go into the `build` folder.

Check out this document, which has a table showing what each folder in a NuGet package is used for:

https://docs.microsoft.com/en-us/nuget/create-packages/creating-a-package#from-a-convention-based-working-directory

[...]

PR #1569, opened 2018-08-06
fix nuget

guolinke:
@codemzs


.nuget/create_nuget.py
```diff
@@ -36,7 +40,7 @@
         <dependencies> </dependencies>
     </metadata>
         <files>
-        <file src="runtimes\**" target="runtimes"/>
+        <file src="lib\**" target="lib"/>
```
guolinke: @codemzs I am not sure about this. Is this correct?

.nuget/create_nuget.py
```diff
@@ -7,18 +7,22 @@
...
+    windows_folder_path = os.path.join(current_dir, "lib", "netstandard2.0", "runtimes", "win-x64", "native")
     if not os.path.exists(windows_folder_path):
         os.makedirs(windows_folder_path)
+    net46_folder_path = os.path.join(current_dir, "lib", "net46")
+    if not os.path.exists(net46_folder_path):
+        os.makedirs(net46_folder_path)
     copy_file(os.path.join(source, "lib_lightgbm.so"), os.path.join(linux_folder_path, "lib_lightgbm.so"))
     copy_file(os.path.join(source, "lib_lightgbm.dylib"), os.path.join(osx_folder_path, "lib_lightgbm.dylib"))
     copy_file(os.path.join(source, "lib_lightgbm.dll"), os.path.join(windows_folder_path, "lib_lightgbm.dll"))
+    copy_file(os.path.join(source, "lib_lightgbm.dll"), os.path.join(net46_folder_path, "lib_lightgbm.dll"))
```
codemzs: What about the exe?

.nuget/create_nuget.py
codemzs: I'm not sure either. Can you please send me this nuget so that I can try locally? 

---
In reply to: [207795534](https://github.com/Microsoft/LightGBM/pull/1569#discussion_r207795534) [](ancestors = 207795534)

.nuget/create_nuget.py
guolinke: @codemzs the exe is removed in the latest nuget package, do we still need it ?

.nuget/create_nuget.py
guolinke: [Nuget.zip](https://github.com/Microsoft/LightGBM/files/2264737/Nuget.zip)

guolinke: you also can download it from vsts bulid: https://lightgbm.visualstudio.com/LightGBM-Build/LightGBM-Build%20Team/_build

eerhardt:
@guolinke - check out my comment [here](https://github.com/Microsoft/LightGBM/issues/1566#issuecomment-411071907).  I believe that is a better strategy to fix this problem.

The way you have the folders structured in this PR isn't recognized by NuGet. You can't/shouldn't put native binaries under the `lib` folder.  Native binaries should go under the `runtimes` folder.  Check out https://docs.microsoft.com/en-us/nuget/create-packages/creating-a-package#from-a-convention-based-working-directory for more information.

guolinke:
Thanks @eerhardt.
So what contents should be put in the lib/net46 ?
BTW, How to set the “target“  field?

guolinke:
@eerhardt the “files” tab in nuspec file.

guolinke:
@eerhardt @codemzs 
I just update it according to @eerhardt 's solution. 
Can you help to check that ?

.nuget/create_nuget.py
```diff
@@ -36,9 +40,44 @@
...
         <file src="runtimes\**" target="runtimes"/>
         </files>
     </package>
     ''' % version
+    prop_str = '''
+    <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+    <ItemGroup Condition="Exists('packages.config') OR
+                            Exists('$(MSBuildProjectName).packages.config') OR
+                            Exists('packages.$(MSBuildProjectName).config')">
+        <Content Include="$(MSBuildThisFileDirectory)\..\runtimes\win-x64\native\*.dll"
```
eerhardt: This doesn't work because `\r` and `\n` are getting treated as carriage return and newline characters.

I would check the `.props` and `.targets` files into the repo, and treat them like any normal file.

.nuget/create_nuget.py
```diff
@@ -36,9 +40,44 @@
...
+                Condition="'$(PlatformTarget)' == 'x64'">
+        <Content Include="$(MSBuildThisFileDirectory)\..\runtimes\win-x64\native\*.exe"
```
eerhardt: This is not valid MSBuild syntax (or even XML for that matter).

```xml
        <Content Include="$(MSBuildThisFileDirectory)\runtimes\win-x64\native\*.dll"
                Condition="'$(PlatformTarget)' == 'x64'">
        <Content Include="$(MSBuildThisFileDirectory)\runtimes\win-x64\native\*.exe"
                Condition="'$(PlatformTarget)' == 'x64'">
            <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
            <Visible>false</Visible>
        </Content>
```

You have 2 open tags for `Content`, but only 1 closing tag.  Also `Content` items can't be nested.

eerhardt: I'm not sure why `.exe` needs to be copied to the output directory.  But if you have user scenarios for it, then I guess it should be.  Is this really needed?

guolinke:
@codemzs can you test the new package ? 

https://lightgbm.visualstudio.com/c88b2f0f-1413-4908-8f10-5f0de4d87924/_apis/build/builds/542/artifacts?artifactName=Nuget&api-version=5.0-preview.3&%24format=zip

.nuget/create_nuget.py
```diff
@@ -36,9 +40,47 @@
...
+        <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+        <Visible>false</Visible>
+        </Content>
+    </ItemGroup>
+    </Project>
+    '''
+    target_str = '''
+    <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+    <PropertyGroup>
+        <EnableMLUnsupportedPlatformTargetCheck Condition="'$(EnableMLUnsupportedPlatformTargetCheck)' == ''">true</EnableMLUnsupportedPlatformTargetCheck>
```
eerhardt: I think we should use a different MSBuild property name here.  `EnableMLUnsupportedPlatformTargetCheck` should be used for ML.NET (since it is "Enable ML Unsupported...")

.nuget/create_nuget.py
```diff
@@ -36,9 +40,47 @@
...
+    </PropertyGroup>
+    <Target Name="_CheckForUnsupportedPlatformTarget"
```
eerhardt: Similar to above, since MSBuild doesn't have the concept of namespaces, we should use a different Target name here than what is used in ML.NET. If someone uses both packages, the targets will conflict and only 1 will be executed.

Maybe something like `_CheckForLightGBMUnsupportedPlatformTarget`

.nuget/create_nuget.py
```diff
@@ -66,11 +66,11 @@
...
     <PropertyGroup>
-        <EnableMLUnsupportedPlatformTargetCheck Condition="'$(EnableMLUnsupportedPlatformTargetCheck)' == ''">true</EnableMLUnsupportedPlatformTargetCheck>
+        <EnableLightGBMUnsupportedPlatformTargetCheck Condition="'$(EnableLightGBMUnsupportedPlatformTargetCheck)' == ''">true</EnableLightGBMUnsupportedPlatformTargetCheck>
     </PropertyGroup>
-    <Target Name="_CheckForUnsupportedPlatformTarget"
-            Condition="'$(EnableMLUnsupportedPlatformTargetCheck)' == 'true'"
-            AfterTargets="_CheckForInvalidConfigurationAndPlatform">
+    <Target Name="_LightGBMCheckForUnsupportedPlatformTarget"
+            Condition="'$(EnableLightGBMUnsupportedPlatformTargetCheck)' == 'true'"
+            AfterTargets="_LightGBMCheckForInvalidConfigurationAndPlatform">
```
eerhardt: This one actually needs to stay the same:  `AfterTargets="_CheckForInvalidConfigurationAndPlatform"`.

`AfterTargets` is an MSBuild extensiblity mechanism to say "run this target after you run this other target".  See https://docs.microsoft.com/en-us/visualstudio/msbuild/target-element-msbuild#attributes-and-elements and https://msdn.microsoft.com/en-us/library/ee216359.aspx.

 `_CheckForInvalidConfigurationAndPlatform` is a well-known target that is defined in the common MSBuild files.

guolinke:
@codemzs can you try it again ? 
[Nuget (3).zip](https://github.com/Microsoft/LightGBM/files/2308780/Nuget.3.zip)

Issue #1494, opened 2018-07-07
Inconsistent meaning of num_iteration=-1: "all models" vs. "best model"


jdanbrown:
- In C, R, and in python `_InnerPredictor`, `num_iteration=-1` means "use all models".
- In python `Booster`, `num_iteration=-1` means "use the best model". Is this "best model" behavior incorrect? It's confusing, at least.

"All models" behavior:
- C
  - `SaveModelToFile` / `SaveModelToString`: [gbdt.h](https://github.com/Microsoft/LightGBM/blob/v2.1.2/src/boosting/gbdt.h#L244), [gbdt_model_text.cpp](https://github.com/Microsoft/LightGBM/blob/master/src/boosting/gbdt_model_text.cpp#L261-L264)
  - `DumpModel`: [gbdt.h](https://github.com/Microsoft/LightGBM/blob/v2.1.2/src/boosting/gbdt.h#L214), [gbdt_model_text](https://github.com/Microsoft/LightGBM/blob/master/src/boosting/gbdt_model_text.cpp#L31-L34)
  - `InitPredict`: [gbdt.h](https://github.com/Microsoft/LightGBM/blob/v2.1.2/src/boosting/gbdt.h#L299-L302)
  - `NumPredictOneRow`: [gbdt.h](https://github.com/Microsoft/LightGBM/blob/v2.1.2/src/boosting/gbdt.h#L181-L186)
- R
  - [lightgbm_R.h](https://github.com/Microsoft/LightGBM/blob/v2.1.2/include/LightGBM/lightgbm_R.h)
  - [lgb.Predictor.R](https://github.com/Microsoft/LightGBM/blob/v2.1.2/R-package/R/lgb.Predictor.R)
- Python
  - `_InnerPredictor.predict`: [basic.py](https://github.com/Microsoft/LightGBM/blob/v2.1.2/python-package/lightgbm/basic.py#L376)

"Best model" behavior:
- Python
  - `Booster.save_model`: [basic.py](https://github.com/Microsoft/LightGBM/blob/v2.1.2/python-package/lightgbm/basic.py#L1667-L1668)
  - `Booster.save_model_to_string`: [basic.py](https://github.com/Microsoft/LightGBM/blob/v2.1.2/python-package/lightgbm/basic.py#L1696-L1697)
  - `Booster.dump_model`: [basic.py](https://github.com/Microsoft/LightGBM/blob/v2.1.2/python-package/lightgbm/basic.py#L1735-L1736)
[...]

PR #1575, opened 2018-08-08
[python] use None for best_iter and -1 for all iters

StrikerRUS:
Addressed #1494.

Now all Python methods/functions use `None` for **best iterations** (default behavior) and `-1` for **all iterations**.


python-package/lightgbm/basic.py
```diff
@@ -1748,9 +1749,9 @@ def _load_model_from_string(self, model_str, verbose=True):
             print('Finished loading model, total used %d iterations' % (int(out_num_iterations.value)))
         self.__num_class = out_num_class.value
 
-    def _save_model_to_string(self, num_iteration=-1):
+    def _save_model_to_string(self, num_iteration=None):
         """[Private] Save model to string"""
-        if num_iteration <= 0:
+        if num_iteration is None:
             num_iteration = self.best_iteration
```
guolinke: just to confirm: self.best_iteration is default -1 ?

henry0312:
I think we should check this PR carefully.

StrikerRUS:
@henry0312 Yeah! Please do it.

python-package/lightgbm/sklearn.py
```diff
@@ -530,6 +531,8 @@ def predict(self, X, raw_score=False, num_iteration=-1,
                              "match the input. Model n_features_ is %s and "
                              "input n_features is %s "
                              % (self._n_features, n_features))
+        if num_iteration is None:
+            num_iteration = self.best_iteration_ if self.best_iteration_ is not None else -1
```
henry0312: Is this just `num_iteration = self.best_iteration_`  because `self.best_iteration` is defult -1?
https://github.com/Microsoft/LightGBM/pull/1575/files#r209197852

StrikerRUS: Good catch!

It seems that we even can remove this line completely, because of
https://github.com/Microsoft/LightGBM/blob/c78f26ed9e20575ab770d0dfa2800d9438c0f4c6/python-package/lightgbm/sklearn.py#L275
https://github.com/Microsoft/LightGBM/blob/c78f26ed9e20575ab770d0dfa2800d9438c0f4c6/python-package/lightgbm/sklearn.py#L486
https://github.com/Microsoft/LightGBM/blob/c78f26ed9e20575ab770d0dfa2800d9438c0f4c6/python-package/lightgbm/sklearn.py#L581-L585

henry0312: Ah, you're right!
we don't need this change.

PR #1584, opened 2018-08-13
[python][ci] start to support Python 3.7

StrikerRUS:
GPU task fails with Python 3.7. I suppose it's due to the miss of the `boost` library for py37 at conda.

Drop Python 3.4 task as misleading because there are no packages for it at conda and conda silently replaces `3.4 -> 3.7` ([log](https://travis-ci.org/Microsoft/LightGBM/jobs/415410861)) in this case (Travis scenario). Appveyor has pinned Miniconda versions and in this case conda doesn't replace missed version with the newest one, but simply crashes:
```
UnsatisfiableError: The following specifications were found to be in conflict:
  - python 3.4*
  - python-graphviz -> python 3.5*
Use "conda info <package>" to see the dependencies for each package.
Command exited with code 1
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,25 +12,25 @@ os:
 env:
   global:  # default values
     - COMPILER=gcc
-    - PYTHON_VERSION=3.6
+    - PYTHON_VERSION=3.7
   matrix:
     - TASK=regular
     - TASK=regular COMPILER=clang
-    - TASK=mpi PYTHON_VERSION=2.7
+    - TASK=mpi
     - TASK=pylint
     - TASK=check-docs
     - TASK=if-else
-    - TASK=sdist PYTHON_VERSION=3.4
-    - TASK=bdist PYTHON_VERSION=3.5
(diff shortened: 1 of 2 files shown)


henry0312:
Hum, why not test with both Python 3.6 and 3.7?

StrikerRUS:
@henry0312 Sorry, didn't get it. Python 3.6 is tested here:
```
TASK=gpu METHOD=pip PYTHON_VERSION=3.6
```

henry0312:
@StrikerRUS Isn't a test with cpu and Python 3.6 needed?

StrikerRUS:
What do you think about such change?
```
TASK=regular PYTHON_VERSION=3.6
```

Python 3.7 will be evaluated in `bdist` tsk

henry0312:
This depends on which versions we have more priority over.
As for me, I prefer `TASK=regular PYTHON_VERSION=3.6` until release of Python 3.7.1.

StrikerRUS:
Is it OK now?

PR #1586, opened 2018-08-14
fix include

### LINKED ISSUE #1585: c_api.h uses undefined THREAD_LOCAL macro
## Environment info

Operating System: x86_64-linux-gnu

CPU: Intel(R) Core(TM) i7-7700 CPU @ 3.60GHz

C++/Python/R version: gcc version 5.4.0 20160609 (Ubuntu 5.4.0-6ubuntu1~16.04.9)

## Error message

/usr/local/include/LightGBM/c_api.h: In function ‘char* LastErrorMsg()’:
/usr/local/include/LightGBM/c_api.h:765:38: error: ‘THREAD_LOCAL’ does not name a type
 static char* LastErrorMsg() { static THREAD_LOCAL char err_msg[512] = "Everything is fine"; return err_msg; }
                                      ^
/usr/local/include/LightGBM/c_api.h:765:100: error: ‘err_msg’ was not declared in this scope
 static char* LastErrorMsg() { static THREAD_LOCAL char err_msg[512] = "Everything is fine"; return err_msg; }

## Reproducible examples

In any source file, including <LightGBM/c_api.h>
results in compile error about undefined THREAD_LOCAL

The only workaround is to include <LightGBM/utils/log.h> before including <LightGBM/c_api.h>. Moving the definition of THREAD_LOCAL to c_api.h (or export.h) would fix the issue.

guolinke:
fix #1585

### DIFF
--- a/include/LightGBM/c_api.h
+++ b/include/LightGBM/c_api.h
@@ -1,14 +1,8 @@
 #ifndef LIGHTGBM_C_API_H_
 #define LIGHTGBM_C_API_H_
 
-#include <LightGBM/meta.h>
-#include <LightGBM/utils/log.h>
-
 #include <cstdint>
-#include <exception>
-#include <stdexcept>
 #include <cstring>
-#include <string>
 
 /*!
 * To avoid type conversion on large data, most of our expose interface support both for float_32 and float_64.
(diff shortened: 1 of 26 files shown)


include/LightGBM/c_api.h
```diff
@@ -2,6 +2,7 @@
 #define LIGHTGBM_C_API_H_
 
 #include <LightGBM/meta.h>
```
alokpr: Is this file supposed to expose a c-api, which can be used by a C program?
If so, including header files with C++ code will not compile. I believe both - meta.h and log.h - have C++ code?
Other C++ stdlib headers (cstdint, exception, etc) included in this file would also not compile.

guolinke: We don't think it is used for c program.

alokpr: Forgive my ignorance. Why is it called "C" API then, when it cannot be used from a "C" program?

guolinke: that is for python package abd other language to call. Anyway, I just reduce the dependence of head files.

PR #1589, opened 2018-08-15
various improvements around metric param and early_stopping_rounds param description

StrikerRUS:
- consistent and more clear (in part of "all of them except the training data") description of `early_stopping_rounds` parameter;
- added alias ``lambdarank`` for ``ndcg`` metric to fix the expected behavior "(empty string or not specified) means that metric corresponding to specified ``objective`` will be used", which was broken for `lambdarank` objective;
- added `"na"` alias to custom objective for consistency with description and code [here](https://github.com/Microsoft/LightGBM/blob/ad7738ac513f96954eff646bc52e9db604da66ae/src/io/config.cpp#L205-L206);
- it turned out that ReadTheDocs didn't show all members in **Scikit-learn API** section, `:inherited-members:` option fixes this;
- added default metric to the evaluation list in case of *string* (built-in) `feval` parameter in sklearn wrapper to match the description: "In either case, the ``metric`` from the model parameters will be evaluated and used as well." (refer to https://github.com/Microsoft/LightGBM/pull/1505#issuecomment-404472494)


StrikerRUS:
ping @guolinke for metric part;
ping @henry0312 for python and docstrings part;
ping @Laurae2 for R part

python-package/lightgbm/sklearn.py
```diff
@@ -417,7 +419,24 @@ def fit(self, X, y,
...
+                elif isinstance(self, LGBMRanker):
+                    original_metric = "ndcg"
+            # overwrite default metric by explicitly set metric
+            for metric_alias in ['metric', 'metrics', 'metric_types']:
+                if metric_alias in params:
+                    original_metric = params.pop(metric_alias)
+            # concatenate metric from params (or default if not provided in params) and eval_metric
+            original_metric = [original_metric] if isinstance(original_metric, (string_type, type(None))) else original_metric
+            eval_metric = [eval_metric] if isinstance(eval_metric, (string_type, type(None))) else eval_metric
+            params['metric'] = set(original_metric + eval_metric)
```
henry0312: I have a questions.

1. Can we pass just `eval_metric`?

I guess `original_metric` is always used, and I can't run `fit` under a case when `eval_metric` is different from `original_metric`.
(ex. want to use "l2" for train loss and only "l1" for eval metric)

StrikerRUS: @henry0312 
>Can we pass just eval_metric?

I suppose it's confusing. Users used to pass `metric` in params and happy with this.

This PR was done with the aim to make cases with callable and string `feval` consistent. At present behavior is different for callable and string `feval` (please see https://github.com/Microsoft/LightGBM/pull/1505#issuecomment-404472494). And it's impossible to omit `metric` in case of callable `feval` because it will break the possibility to disable default metric at cpp side by `metric="None"`.

henry0312: I sometime set `eval_metric` different from train loss (`original_metric` in the PR).
This PR won't allow me to do 😢
For now, I don't understand why we need to consider `original_metric` as default for `eval_metric`.

henry0312: Is it possible to set `original_metric` when `eval_set is not None` and `eval_metric is None`?

StrikerRUS: >I sometime set eval_metric different from train loss (original_metric in the PR).
This PR won't allow me to do 😢

I think you are able to do this with this PR. You can set `eval_metric` to anything you want and it will be passed without any changes.

Maybe I'm missing something in your use case. Please provide an example with `metric` and `eval_metric` you want to use and have doubts about its viable.

StrikerRUS: >Is it possible to set original_metric when eval_set is not None and eval_metric is None?

Sorry, I didn't get it. `eval_set` wasn't touched in this PR (only excess phrase was removed from the docstring).

There is no `original_metric` param for users. There are only `metric` (and its aliases) and `eval_metric` and they are independently treated here (OK, I lied, at the end `set` is used to leave only unique values in their union).

You can set any allowed value from Parameters.rst in `metric` and this value will overwrite default metric (internal `original_metric` variable). At the same time, `eval_metric` can be `None` (please pay attention NOT `"None"`), which means that only `metric` will be used. Or you can set `eval_metric` to some certain value and this value will be **appended** to the value(s) in `metric`.

StrikerRUS:
@guolinke While we're waiting @henry0312 with his use-case, are you OK with this PR?

guolinke:
LGTM

python-package/lightgbm/sklearn.py
henry0312: I'm sorry for late response, because I've been busy with my work.

And I am very sorry I found that I misunderstood the logic of `eval_metric`.
This PR will also allow me to run lgm in the same way with the current way.

StrikerRUS: No problem, @henry0312 !
Just to clarify, the only change for users is that default metric will be evaluated in case of **string** `eval_metric`, like it is done right now for the case of **callable** `eval_metric` (for the consistency purposes). Also, users are able to disable the evaluation of the default metric by setting `metric="None"`.

PR #1594, opened 2018-08-19
[docs] added possibility for collapsing sections

StrikerRUS:
Refer to https://github.com/Microsoft/LightGBM/pull/1553#issuecomment-410292318 .

Due to the fact that the installation guide is getting bigger and bigger
- added **Contents** at the top of the file;
- added possibility to collapse sections with building non-regular versions (collapsed by default).

Unfortunately, I can't find the way to implement collapsing technique with less lines of code.
If I'm not mistaken, approximately year ago @henry0312 you were against JS code, so I want to know your opinion. Maybe just adding **Contents** is enough, I don't know...

It looks like:
collapsed:
image

uncollapsed:
image

If someone forgot how to build documentation locally: https://github.com/Microsoft/LightGBM/blob/collapse/docs/README.rst#build

I'm not sure that `^` and `v` is the best choice, I'm open for proposals from this list https://fontawesome.com/v4.7.0/icons/ .


chivee:
Nice and neat!

Issue #1557, opened 2018-07-31
Build error of Apple Clang 8.0 and below


StrikerRUS:
Warnings: https://github.com/Microsoft/LightGBM/pull/1522#issuecomment-404970003

They should be fixed:

>@guolinke 

>I think all these warning should be fixed, at least set some ignore tags in cmake.

Errors: https://github.com/Microsoft/LightGBM/pull/1522#issuecomment-404973928 https://github.com/Microsoft/LightGBM/pull/1522#issuecomment-404983245

ping @guolinke @chivee @henry0312

chivee:
according to [this](https://stackoverflow.com/a/29929949) , `thread_local` was supported since AppleClang 8. I think we might set the minimum version to that.

as for the `std::move` error, I'll create an patch for that.
[ref](http://www.cplusplus.com/forum/general/186411/)

StrikerRUS:
@chivee Thank you!

Unfortunately, AppleClang 8.0 fails with segfault: https://github.com/Microsoft/LightGBM/pull/1522#issuecomment-404983245

chivee:
@StrikerRUS Then we could change the title and leave it for people searching for answers.

StrikerRUS:
@chivee "... and below" is already [answered](https://github.com/Microsoft/LightGBM/issues/1557#issuecomment-409423016) by you previously :-)
>according to this , thread_local was supported since AppleClang 8. I think we might set the minimum version to that.

However, warnings https://github.com/Microsoft/LightGBM/pull/1522#issuecomment-404970003 are still not fixed.

To be honest, I'd like close this issue after fixing warnings. 8.1 is good version to start support and users are protected from 8.0 segfaults by these lines
https://github.com/Microsoft/LightGBM/blob/c78f26ed9e20575ab770d0dfa2800d9438c0f4c6/CMakeLists.txt#L28-L30

chivee:
@StrikerRUS  Yes, I mean to change to the title so that the user knows well why we don't offer the support of version below 8.1

StrikerRUS:
@chivee Of course, as you wish!

brigoodwinan:
Apologies if this is a potential noob question, but it appears that LightGBM just isn't going to support compilers lower than clang 8.1, correct? I'm running on a Mac Pro 2009, which cannot (officially) update to clang 8.1, but is stuck at 8.0... Is there a work around?

`
build $ cmake ..
CMake Error at CMakeLists.txt:30 (message):
[...]

PR #1596, opened 2018-08-20
remove unnecessary  std::move & unused capture

chivee:
according to the [discussion](https://stackoverflow.com/questions/4986673/c11-rvalues-and-move-semantics-confusion-return-statement/4986802#4986802) here.

and the C++11 §12.8/32

```
When the criteria for elision of a copy/move operation are met, but not for an exception-declaration, and the
object to be copied is designated by an lvalue, or when the expression in a return statement is a (possibly
parenthesized) id-expression that names an object with automatic storage duration declared in the body or
parameter-declaration-clause of the innermost enclosing function or lambda-expression, overload resolution
to select the constructor for the copy is first performed as if the object were designated by an rvalue. If
the first overload resolution fails or was not performed, or if the type of the first parameter of the selected
constructor is not an rvalue reference to the object’s type (possibly cv-qualified), overload resolution is
performed again, considering the object as an lvalue. [ Note: This two-stage overload resolution must be
performed regardless of whether copy elision will occur. It determines the constructor to be called if elision
is not performed, and the selected constructor must be accessible even if the call is elided. — end note ]
```

C++11 §12.8/31
```
When certain criteria are met, an implementation is allowed to omit the copy/move construction of a class
object, even if the constructor selected for the copy/move operation and/or the destructor for the object
have side effects. In such cases, the implementation treats the source and target of the omitted copy/move
operation as simply two different ways of referring to the same object. If the first parameter of the selected
constructor is an rvalue reference to the object’s type, the destruction of that object occurs when the target
would have been destroyed; otherwise, the destruction occurs at the later of the times when the two objects
would have been destroyed without the optimization.122 This elision of copy/move operations, called copy
elision, is permitted in the following circumstances (which may be combined to eliminate multiple copies):
```
fixing #1557 

edit:
removing all the explicit captures and leave them to compilers.

### DIFF
--- a/include/LightGBM/utils/pipeline_reader.h
+++ b/include/LightGBM/utils/pipeline_reader.h
@@ -46,7 +46,7 @@ public:
     while (read_cnt > 0) {
       // start read thread
       std::thread read_worker = std::thread(
-        [&, buffer_size] {
+        [&] {
         last_read_cnt = reader->Read(buffer_read.data(), buffer_size);
       }
       );
--- a/include/LightGBM/utils/pipeline_reader.h
+++ b/include/LightGBM/utils/pipeline_reader.h
@@ -46,7 +46,7 @@ public:
     while (read_cnt > 0) {
       // start read thread
       std::thread read_worker = std::thread(
-        [&reader, &buffer_read, buffer_size, &last_read_cnt] {
+        [&, buffer_size] {
         last_read_cnt = reader->Read(buffer_read.data(), buffer_size);
       }
       );
--- a/include/LightGBM/utils/text_reader.h
+++ b/include/LightGBM/utils/text_reader.h
@@ -136,7 +136,7 @@ public:
   */
   INDEX_T ReadAllLines() {
     return ReadAllAndProcess(
-      [this](INDEX_T, const char* buffer, size_t size) {
+      [=](INDEX_T, const char* buffer, size_t size) {
       lines_.emplace_back(buffer, size);
     });
   }
@@ -221,7 +221,7 @@ public:
 
   INDEX_T CountLine() {
     return ReadAllAndProcess(
-      [this](INDEX_T, const char*, size_t) {
+      [=](INDEX_T, const char*, size_t) {
     });
   }
 
--- a/include/LightGBM/utils/text_reader.h
+++ b/include/LightGBM/utils/text_reader.h
@@ -136,7 +136,7 @@ public:
   */
   INDEX_T ReadAllLines() {
     return ReadAllAndProcess(
-      [](INDEX_T, const char* buffer, size_t size) {
+      [this](INDEX_T, const char* buffer, size_t size) {
       lines_.emplace_back(buffer, size);
     });
   }
@@ -221,7 +221,7 @@ public:
 
   INDEX_T CountLine() {
     return ReadAllAndProcess(
-      [](INDEX_T, const char*, size_t) {
+      [this](INDEX_T, const char*, size_t) {
     });
   }
 
--- a/include/LightGBM/utils/text_reader.h
+++ b/include/LightGBM/utils/text_reader.h
@@ -84,7 +84,7 @@ public:
     last_line_ = "";
     INDEX_T total_cnt = 0;
     PipelineReader::Read(filename_, skip_bytes_,
-      [this, &total_cnt, &process_fun]
+      [&]
     (const char* buffer_process, size_t read_cnt) {
       size_t cnt = 0;
       size_t i = 0;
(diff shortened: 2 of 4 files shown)


StrikerRUS:
Speaking about red CI, csie.ntu.edu has very unstable web-server which crashes very often and causes red `check-docs` test. We cannot do anything, and should just wait they repair they server and re-run the test. And there is phantom ` OSError: exception: access violation writing` error at Appveyor again (I have no permissions to re-run Appveyor).

chivee:
@StrikerRUS  can we write some exception for thses sites? 

The second part will be pushed later.

StrikerRUS:
> can we write some exception for thses sites?

Yep, we can include them in the ignore list here: https://github.com/Microsoft/LightGBM/blob/1b53fa0763bc6e71a021be002ee9985ba079d5bb/docs/.linkcheckerrc#L7-L9

But what if they'll become unreachable permanently at some point in the future? We'll never discover it.

PS. It takes csie.ntu.edu support team approximately a day (if it's not a weekend) to repair the web-server.

chivee:
```
Scanning dependencies of target _lightgbm
[  3%] Building CXX object CMakeFiles/_lightgbm.dir/src/c_api.cpp.o
[  6%] Building CXX object CMakeFiles/_lightgbm.dir/src/lightgbm_R.cpp.o
[  9%] Building CXX object CMakeFiles/_lightgbm.dir/src/application/application.cpp.o
[ 12%] Building CXX object CMakeFiles/_lightgbm.dir/src/boosting/boosting.cpp.o
[ 15%] Building CXX object CMakeFiles/_lightgbm.dir/src/boosting/gbdt.cpp.o
[ 18%] Building CXX object CMakeFiles/_lightgbm.dir/src/boosting/gbdt_model_text.cpp.o
[ 21%] Building CXX object CMakeFiles/_lightgbm.dir/src/boosting/gbdt_prediction.cpp.o
[ 25%] Building CXX object CMakeFiles/_lightgbm.dir/src/boosting/prediction_early_stop.cpp.o
[ 28%] Building CXX object CMakeFiles/_lightgbm.dir/src/io/bin.cpp.o
[ 31%] Building CXX object CMakeFiles/_lightgbm.dir/src/io/config.cpp.o
[ 34%] Building CXX object CMakeFiles/_lightgbm.dir/src/io/config_auto.cpp.o
[ 37%] Building CXX object CMakeFiles/_lightgbm.dir/src/io/dataset.cpp.o
[ 40%] Building CXX object CMakeFiles/_lightgbm.dir/src/io/dataset_loader.cpp.o
[ 43%] Building CXX object CMakeFiles/_lightgbm.dir/src/io/file_io.cpp.o
[ 46%] Building CXX object CMakeFiles/_lightgbm.dir/src/io/json11.cpp.o
[ 50%] Building CXX object CMakeFiles/_lightgbm.dir/src/io/metadata.cpp.o
[ 53%] Building CXX object CMakeFiles/_lightgbm.dir/src/io/parser.cpp.o
[ 56%] Building CXX object CMakeFiles/_lightgbm.dir/src/io/tree.cpp.o
[ 59%] Building CXX object CMakeFiles/_lightgbm.dir/src/metric/dcg_calculator.cpp.o
[ 62%] Building CXX object CMakeFiles/_lightgbm.dir/src/metric/metric.cpp.o
[ 65%] Building CXX object CMakeFiles/_lightgbm.dir/src/network/linker_topo.cpp.o
[ 68%] Building CXX object CMakeFiles/_lightgbm.dir/src/network/linkers_mpi.cpp.o
[ 71%] Building CXX object CMakeFiles/_lightgbm.dir/src/network/linkers_socket.cpp.o
[ 75%] Building CXX object CMakeFiles/_lightgbm.dir/src/network/network.cpp.o
[ 78%] Building CXX object CMakeFiles/_lightgbm.dir/src/objective/objective_function.cpp.o
[ 81%] Building CXX object CMakeFiles/_lightgbm.dir/src/treelearner/data_parallel_tree_learner.cpp.o
[ 84%] Building CXX object CMakeFiles/_lightgbm.dir/src/treelearner/feature_parallel_tree_learner.cpp.o
[ 87%] Building CXX object CMakeFiles/_lightgbm.dir/src/treelearner/gpu_tree_learner.cpp.o
[ 90%] Building CXX object CMakeFiles/_lightgbm.dir/src/treelearner/serial_tree_learner.cpp.o
[ 93%] Building CXX object CMakeFiles/_lightgbm.dir/src/treelearner/tree_learner.cpp.o
[ 96%] Building CXX object CMakeFiles/_lightgbm.dir/src/treelearner/voting_parallel_tree_learner.cpp.o
[100%] Linking CXX shared library ../lib_lightgbm.so
[100%] Built target _lightgbm
```
building success.

PR #1602, opened 2018-08-22
[python] return self in most functions in basic.py module and fix docstrings

StrikerRUS:
Refer to https://github.com/Microsoft/LightGBM/pull/1565#issuecomment-415022270 .

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -184,7 +184,7 @@ def convert_from_sliced_object(data):
     """fix the memory of multi-dimensional sliced object"""
     if data.base is not None and isinstance(data, np.ndarray) and isinstance(data.base, np.ndarray):
         if not data.flags.c_contiguous:
-            warnings.warn("Usage subset(sliced data) of np.ndarray is not recommended due to it will double the peak memory cost in LightGBM.")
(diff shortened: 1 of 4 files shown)


python-package/lightgbm/basic.py
```diff
@@ -1239,17 +1297,14 @@ def get_group(self):
...
         """
         if self.group is None:
             self.group = self.get_field('group')
             if self.group is not None:
                 # group data from LightGBM is boundaries data, need to convert to group size
-                new_group = []
-                for i in range_(len(self.group) - 1):
-                    new_group.append(self.group[i + 1] - self.group[i])
-                self.group = new_group
+                self.group = np.diff(self.group)
```
StrikerRUS: `np.diff`

python-package/lightgbm/basic.py
```diff
@@ -1990,8 +2104,7 @@ def feature_importance(self, importance_type='split', iteration=-1):
             importance_type_int = 1
         else:
             importance_type_int = -1
-        num_feature = self.num_feature()
-        result = np.array([0 for _ in range_(num_feature)], dtype=np.float64)
+        result = np.zeros(self.num_feature(), dtype=np.float64)
```
StrikerRUS: `np.zeros`

Issue #1534, opened 2018-07-20
predict() only uses one core


UserQuestions:
When I use predict() on a lightgbm.sklearn.LGBMRegressor, lightgbm appears only to use a single core.  My understanding was that this was fixed by [fix](https://github.com/Microsoft/LightGBM/commit/728e50a9cfcd33a128ab5df9d70f743b92ea82fc) in response to [original issue](https://github.com/Microsoft/LightGBM/issues/144). Am I doing something wrong? Or perhaps is the use of SKLearn preventing things from working correctly?  Any help or advice would be appreciated. Thank you!

## Environment info
Operating System: CentOS7
C++/Python/R version: Python 3.6.3
LightGBM Version: 2.1.1

guolinke:
@UserQuestions 
It depends how many data used in prediction. 
if there are only one sample, it is single core. otherwise it will use multi-threading.

UserQuestions:
@guolinke 
Thank you for your prompt response. I am not sure I understand. I thought .predict() only accepted a single dataset at once.  I am trying to predict a large dataset (potentially hundreds of millions of observations), and I give it all to .predict() at once.  I have a large number of cores but only see one core working when I use model.predict(dataset).

guolinke:
@UserQuestions In such large dataset, the prediction should be multi-threading. 
Is that okay if you using lightgbm.predict ? 
you can do it via https://lightgbm.readthedocs.io/en/latest/Python-API.html#lightgbm.Booster.predict ? 
The booster can be got via: https://lightgbm.readthedocs.io/en/latest/Python-API.html#lightgbm.LGBMModel.booster_

UserQuestions:
Thanks again for the prompt response @guolinke . I tried it using predfun.booster_.predict(data), however it still uses just one core. It's worth noting that type(predfun.booster_) is lightgbm.basic.Booster, not lightgbm.Booster.

guolinke:
how about set 'num_threads=8' in pred_parameter dict, when calling https://lightgbm.readthedocs.io/en/latest/Python-API.html#lightgbm.Booster.predict ?

UserQuestions:
@guolinke 
That works! Thanks for the help! It's interesting that a deprecated parameter is the solution (though I'm not running the newest version).

guolinke:
you can use 'kwargs' , just set num_threads=8 in the arguments of predict function.

mirekphd:
[...]

PR #1603, opened 2018-08-23
add support of refit-decay

guolinke:


tests/python_package_test/test_engine.py
```diff
@@ -633,3 +633,25 @@ def is_correctly_constrained(learner):
...
+        evals_result = {}
+        gbm = lgb.train(params, lgb_train,
+                        num_boost_round=20,
+                        valid_sets=lgb_eval,
+                        verbose_eval=False,
+                        evals_result=evals_result)
+        err_pred = log_loss(y_test, gbm.predict(X_test))
+        new_gbm = gbm.refit(X_test, y_test)
+        new_err_pred = log_loss(y_test, new_gbm.predict(X_test))
+        assert err_pred > new_err_pred
```
StrikerRUS: `self.assertGreater`

include/LightGBM/config.h
```diff
@@ -368,6 +368,13 @@ struct Config {
   // desc = see `this file <https://github.com/Microsoft/LightGBM/tree/master/examples/binary_classification/forced_splits.json>`__ as an example
   std::string forcedsplits_filename = "";
 
+  // type = double
+  // default = 0.9
+  // check = >=0.0
+  // check = <=1.0
+  // desc = Decay rate of "refit" task, will use ``leaf_output = refit_decay_rate * old_leaf_output + (1.0 - refit_decay_rate) * new_leaf_output`` to refit trees.
+  double refit_decay_rate = 0.9;
```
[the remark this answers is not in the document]
guolinke: refit is supported in python-package after this PR.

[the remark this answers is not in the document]
StrikerRUS: Sorry, but I'm a little bit confused. I interpret these parameters like the following: I can add any of them into `params` argument of `train` or `cv` function.
But `refit_decay_rate` parameter can not be used as stand alone param (or I'm mistaken), only as an **argument of `refit` method of Booster**.

BTW, the same about the `task` parameter (I can't use, let say, `params={'task': 'refit', 'refit_decay_rate': 0.5, 'verbose': -1}`).

Maybe for the `task` param it's better to paraphrase like

```
**Note**: can be used only in CLI version; for language-specific packages you can use the correspondent functions
```

And for `refit_decay_rate`: 

```
used only in ``refit`` task in CLI version or as argument in ``refit`` function in language-specific package
```

tests/python_package_test/test_engine.py
```diff
@@ -633,3 +633,24 @@ def is_correctly_constrained(learner):
...
+    def test_refit(self):
+        X, y = load_breast_cancer(True)
+        X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.1, random_state=42)
+        params = {
+            'objective': 'binary',
+            'metric': 'binary_logloss',
+            'verbose': -1
+        }
+        lgb_train = lgb.Dataset(X_train, y_train)
+        lgb_eval = lgb.Dataset(X_test, y_test, reference=lgb_train)
```
StrikerRUS: Do we need validation dataset in this test?

tests/python_package_test/test_engine.py
```diff
@@ -633,3 +633,25 @@ def is_correctly_constrained(learner):
...
+        X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.1, random_state=42)
+        X_train, X_valid, y_train, y_valid = train_test_split(X_train, y_train, test_size=0.3, random_state=24)
+        params = {
+            'objective': 'binary',
+            'metric': 'binary_logloss',
+            'verbose': -1,
+            'min_data': 10
+        }
+        lgb_train = lgb.Dataset(X_train, y_train)
+        lgb_eval = lgb.Dataset(X_valid, y_valid, reference=lgb_train)
```
StrikerRUS: I probably was unclear, sorry. I meant, `lgb_eval` are not used in this test (neither for early stopping, nor metric evaluation).

guolinke: oh, got you.

guolinke:
@StrikerRUS you are right, I will fix the doc.

tests/python_package_test/test_engine.py
```diff
@@ -633,3 +633,22 @@ def is_correctly_constrained(learner):
...
+        X, y = load_breast_cancer(True)
+        X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.1, random_state=42)
+        params = {
+            'objective': 'binary',
+            'metric': 'binary_logloss',
+            'verbose': -1,
+            'min_data': 10
+        }
+        lgb_train = lgb.Dataset(X_train, y_train)
+        evals_result = {}
```
StrikerRUS: `evals_result` is not needed in this test too.

StrikerRUS:
@guolinke I'm afraid that `refit` Python method can suffer from the same [issue](https://github.com/Microsoft/LightGBM/issues/1534) as `predict`, because there are no additional params for `refit`.

guolinke:
@StrikerRUS oh, you are right, we can add kwargs to it.

Issue #1523, opened 2018-07-14
Apple Clang support


StrikerRUS:
Since we've started to support Apple Clang in #1501, we should clarify installation instructions with the notes about the possibility of compiling certain versions of LightGBM with Apple Clang.

For ordinary version we have the requirement `CMake >=3.12` because of `FindOpenMP`. And what about
- MPI version;
- GPU version;
- HDFS version

?

ping @guolinke @chivee @levyfan @huanzhang12 @ebernhardson @henry0312

StrikerRUS:
@chivee Sorry, probably I was unclear, but I wanted to know about the general possibility of compiling these LightGBM versions with AppleClang and its required version for doing this.

StrikerRUS:
@guolinke 
>How about the binary(python wheel) ?
can we replace it with clang version ? and then users don't need to install gcc. However, it may still require to upgrade clang.

Maybe it's overkill, but what about providing several Python wheels with different versions? For instance, `lightgbm`, `lightgbm-gpu` (like `tensorflow` does), `lightgbm-gcc` (compiled with gcc on macOS and with MinGW on Windows).

StrikerRUS:
To speak about `MPI` version, compilation at Travis with Apple Clang 8.1 is successful:
https://travis-ci.org/Microsoft/LightGBM/jobs/407115085#L3681

StrikerRUS:
Hi @huanzhang12 !

Don't you know what `Invalid Work Group Size` error means?
https://travis-ci.org/Microsoft/LightGBM/builds/409535904

https://github.com/Microsoft/LightGBM/tree/gpu-apple (base branch is `macos`)

The error appears even after adding OpenCL headers from Khronos too. https://stackoverflow.com/a/23079478

This file cannot be placed into system directory due to SIP on new macOS, so I made the following:

https://github.com/Microsoft/LightGBM/blob/f1fb6438f74dd7f551acf778e63e8ae830364f76/.travis/test.sh#L84-L87
https://github.com/Microsoft/LightGBM/blob/f1fb6438f74dd7f551acf778e63e8ae830364f76/.travis/test.sh#L107

StrikerRUS:
Thank you very much, @huanzhang12 ! As usual, detailed answer :-)

Shame on me that I've missed #667.

To speak about the issue title, are there any special requirements for Apple Clang/CMake/Boost version?

[...]

PR #1606, opened 2018-08-24
support clang with openmp in vsts-ci test

guolinke:


guolinke:
BTW, if a user want to use the binaries built by us, with clang with openmp, what dependency is needed for the user? 
only `brew install libomp` ?

StrikerRUS:
@guolinke 
>only brew install libomp ?

I believe in this.

What do you think about this idea? https://github.com/Microsoft/LightGBM/issues/1523#issuecomment-405010860

If you want to change the library in wheel at PyPI, I think it'll be very painful for users. I mean, everyone already know that they need the latest **gcc** to run LightGBM. Changing the easiest (and most common, I believe) way to install LightGBM from `brew install gcc` to `brew install libomp` will produce a lot of issues here. For instance, remember the case when we  switched compiling the library for PyPI from `gcc-7` to `gcc-8`.

BTW, what ring LightGBM is? https://github.com/MicrosoftDocs/vsts-docs/issues/610#issuecomment-415412630 
Does VSTS PR check available for us?

guolinke:
can we add some log or warnings when user install wheel?

StrikerRUS:
@guolinke Seems to be that no...

https://github.com/pypa/pip/issues/2933#issuecomment-290407667

guolinke:
@StrikerRUS It seems install `libopm` is much easier than `gcc`. 
Moreover, users doesn't need to install the specific version of gcc.
So I prefer to change to clang in osx. 

Maybe we can have a detailed release note from now, and tell user what changes since last release.

StrikerRUS:
@guolinke 
>It seems install libopm is much easier than gcc.

That's for sure!

> So I prefer to change to clang in osx.

Me too. But I want find the way to make this change less painful for users.

I'm worrying about that a lot of users used to install gcc and then install LightGBM. After changing to Apple Clang all of them will start to post issues, because they even don't know where to read about recent changes, except digging into commits history. In addition, all other internet resources (StackOverflow, Kaggle, personal blogs, etc) will suggest them installation with gcc 😢

guolinke:
@StrikerRUS maybe one temporal solution is giving some warnings in python-package when import LightGBM, if detecting the host os is macOS ?

StrikerRUS:
@guolinke Yeah, good idea!

PR #1607, opened 2018-08-24
Small optimizations and fixes

dmitryikh:
Please consider my changes.
1. Exploit the fact that we know `std::vector` size before filling in `RowFunctionFromCSR`
2. I suppose here is typo in `ClearPredictBuffer`. In mine opinion we should use memset in the case when more than a half of elements are non-zeros.

### DIFF
--- a/src/application/predictor.hpp
+++ b/src/application/predictor.hpp
@@ -220,7 +220,7 @@ private:
   }
 
   void ClearPredictBuffer(double* pred_buf, size_t buf_size, const std::vector<std::pair<int, double>>& features) {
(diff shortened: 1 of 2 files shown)


src/application/predictor.hpp
```diff
@@ -220,7 +220,7 @@ class Predictor {
   }
 
   void ClearPredictBuffer(double* pred_buf, size_t buf_size, const std::vector<std::pair<int, double>>& features) {
-    if (features.size() < static_cast<size_t>(buf_size / 2)) {
+    if (features.size() > static_cast<size_t>(buf_size / 2)) {
```
guolinke: thanks, it indeed is bug.

PR #1612, opened 2018-08-25
add future warning for macOS users about switching to Apple Clang

StrikerRUS:
Refer to https://github.com/Microsoft/LightGBM/pull/1606#issuecomment-415932865 .
Is `2.1.4` version OK? 
My thoughts are that at least one version should be uploaded to PyPI with this warning but without switching to Apple Clang.

Connected to #1523.


guolinke:
it sounds good

PR #1615, opened 2018-08-25
[python] add possibility to predict dataset with nrow > max_int32

### LINKED ISSUE #1590: ValueError: negative dimensions are not allowed
## Environment info

Operating System: Debian 

CPU: x86_64

C++/Python/R version: python 3.4

## Error message

  File "/home/beda/venv/system3/lib/python3.4/site-packages/shap/explainers/tree.py", line 144, in shap_values
    phi = self.model.predict(X, num_iteration=tree_limit, pred_contrib=True)
  File "/home/beda/venv/system3/lib/python3.4/site-packages/lightgbm/basic.py", line 1802, in predict
    return predictor.predict(data, num_iteration, raw_score, pred_leaf, pred_contrib, data_has_header, is_reshape)
  File "/home/beda/venv/system3/lib/python3.4/site-packages/lightgbm/basic.py", line 434, in predict
    predict_type)
  File "/home/beda/venv/system3/lib/python3.4/site-packages/lightgbm/basic.py", line 518, in __pred_for_csr
    preds = np.zeros(n_preds, dtype=np.float64)
ValueError: negative dimensions are not allowed

## Steps to reproduce

1. Call LGBMClassifier.predict with pred_contrib=True
2. As data pass CSR matrix with shape  (50000, 48793) or bigger

Probably 32-bit integer overflow when computing number of predictions.

StrikerRUS:
Fixed #1590.

- [x] numpy array

- [x] scipy.sparce csr

- [x] scipy.sparse csc

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -16,7 +16,7 @@ import scipy.sparse
 from .compat import (DataFrame, LGBMDeprecationWarning, Series,
                      decode_string, integer_types,
                      json, json_default_with_numpy,
-                     numeric_types, range_, string_type)
+                     numeric_types, range_, zip_, string_type)
 from .libpath import find_lib_path
 
 
@@ -31,11 +31,6 @@ def _load_lib():
 
 
 _LIB = _load_lib()
-
-
-class LightGBMError(Exception):
-    """Error throwed by LightGBM"""
-    pass
 
 
 def _safe_call(ret):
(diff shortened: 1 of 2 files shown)


StrikerRUS:
The solution for numpy array is based on `array_split` function and impossibility to use it with int param due to this issue https://github.com/numpy/numpy/issues/11809.

The solution for csr matrix is based on direct slicing.

The solution for csc matrix is based on converting to csr matrix due to inefficient **double** converting while transposing when using direct slicing:
https://github.com/scipy/scipy/blob/14142ff70d84a6ce74044a27919c850e893648f7/scipy/sparse/csc.py#L164-L167
https://github.com/scipy/scipy/blob/14142ff70d84a6ce74044a27919c850e893648f7/scipy/sparse/csc.py#L111-L121

PR #1619, opened 2018-08-26
[R-package] update dependencies

jameslamb:
@Laurae2 one more small maintenance PR for the R package. In this PR I addressed the following:

* added a few missing `#' @importFrom` tags and `::` namespaces of functions
* alphabetized imports in `DESCRIPTION`. I know this seems trivial but I've found that as projects grow, it makes it really easy to visually scan and see what the package's dependencies are.

You will see a few other lines that look like I removed extra whitespace. I didn't do this intentionally...I was editing in RStudio and my settings automatically trim excess horizontal whitespace when I save files. If you want me to revert touching those lines, just let me know and I'd be happy to.


Laurae2:
@jameslamb For the whitespace, we can look at it later (not that critical).

PR #1628, opened 2018-08-30
[python] pass params to _InnerPredictor in train and cv and verbose fix

### LINKED ISSUE #1157: Suppress LightGBM Warning
The following line is replicated throughout the training for each iteration and it doesn't seem to be generating through Python's standard (warnings)[https://docs.python.org/3/library/warnings.html] module:

`[LightGBM] [Warning] No further splits with positive gain, best gain: -inf`

How can I suppress this warning? It is generated [here](https://github.com/Microsoft/LightGBM/blob/92f2a570efbd9af898cbb4bf42dec1a595c6e9d7/src/treelearner/serial_tree_learner.cpp#L191). Also, if possible, can you tell me the meaning of this warning?

StrikerRUS:
Fixed #1157.

@guolinke Is it critical that all `params` are passed? It seems that unknown (for `_InnerPredictor`) keys in `params` are just silently ignored and it's OK.


guolinke:
I think it is okay.

StrikerRUS:
Updated `verbosity` param description according to the following code:
https://github.com/Microsoft/LightGBM/blob/7b6f80f350f41d522535d5f62ef9ad7a0cbb0f85/src/io/config.cpp#L182-L190

Treat `silent` argument as **Fatal** log level (`verbose=-1`) to prevent showing warnings with `verbose=-1` in `params` and `silent=True` passed at the same time like in [this use case](https://github.com/Microsoft/LightGBM/issues/1157#issuecomment-401746956).

PR #1629, opened 2018-08-30
[python] added kwargs in refit method

StrikerRUS:
Refer to https://github.com/Microsoft/LightGBM/pull/1603#issuecomment-417114777.

Maybe it's better to predict here with **best** iteration, not **all**?
https://github.com/Microsoft/LightGBM/blob/abd737651f5c8f3c42fde693fb1a18c30bb68be3/python-package/lightgbm/basic.py#L2117

However, `Dataset` creation here is still left nonparameterized. Maybe at least add new `group` param for allowing to refit for ranking task?
https://github.com/Microsoft/LightGBM/blob/abd737651f5c8f3c42fde693fb1a18c30bb68be3/python-package/lightgbm/basic.py#L2120

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -2095,7 +2095,7 @@ class Booster(object):
             num_iteration = self.best_iteration
         return predictor.predict(data, num_iteration, raw_score, pred_leaf, pred_contrib, data_has_header, is_reshape)
 
-    def refit(self, data, label, decay_rate=0.9):
+    def refit(self, data, label, decay_rate=0.9, **kwargs):
         """Refit the existing Booster by new data.
 
         Parameters
@@ -2107,6 +2107,8 @@ class Booster(object):
             Label for refit.
         decay_rate : float, optional (default=0.9)
             Decay rate of refit, will use ``leaf_output = decay_rate * old_leaf_output + (1.0 - decay_rate) * new_leaf_output`` to refit trees.
+        **kwargs : other parameters for refit
+            These parameters will be passed to ``predict`` method.
 
         Returns
         -------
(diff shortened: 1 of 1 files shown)


guolinke:
@StrikerRUS sorry for miss the comments.
For the first question, I think it should use all models for refit, since the copied booster has all models. However, it can use `best_iter` for the prediction of the new booster. 

For the dataset, I think it doesn't need to complete clone as well. Actually, only the X and label are used in refit task, other fields like weight, group are not used.

StrikerRUS:
@guolinke Many thanks for the explanation and sorry for the false alarm!

StrikerRUS:
@guolinke Sorry, but I still can't get the full understanding about what's going on behind the `refit` method. So, I can't answer your question.

I can only add that it seems the same situation is about `best_score` and other attributes set in `Booster`'s `__init__`  (`network` or `__train_data_name`, for example).

guolinke:
@StrikerRUS refit doesn't use a valid data to check the best_score, so it is hard to say which iteration is the best. Maybe we can let it `-1` and give a doc to users about this setting.

StrikerRUS:
@guolinke Yeah, I **suppose** it's good solution.

But what about the case when users called, for example, `set_attr`, `set_train_data_name` or `set_network` and then `refit`? It seems that this info will be lost in the refitted booster, because it's actually a **new** booster.

StrikerRUS:
@guolinke Sorry, under "network" I meant `network` attribute which is switched to `True` in the `set_network` method:
https://github.com/Microsoft/LightGBM/blob/1d075f4dda618515490cb1e9bdbd4fa0a4603c8d/python-package/lightgbm/basic.py#L1595
and then reset back to `False` in the constructor:
https://github.com/Microsoft/LightGBM/blob/1d075f4dda618515490cb1e9bdbd4fa0a4603c8d/python-package/lightgbm/basic.py#L1433

guolinke:
@StrikerRUS oh, got you. Yeah, that should be copied too.

PR #1630, opened 2018-08-31
[python] use kwargs in tree plotting functions

StrikerRUS:
Move all **non-LightGBM-specific** arguments in tree plotting functions to `**kwargs`. It will free us from reviewing Digraph constructor arguments.

Users who are using positional arguments will see the warning.
Users who are already using keyword arguments will see no difference.

### DIFF
--- a/python-package/lightgbm/plotting.py
+++ b/python-package/lightgbm/plotting.py
@@ -348,14 +348,14 @@ def create_tree_digraph(booster, tree_index=0, show_info=None, precision=None,
                        'old_node_attr', 'old_edge_attr', 'old_body']:
         param = locals().get(param_name)
         if param is not None:
-            warnings.warn('{0} parameter is deprecated and will be removed in 2.2 version.\n'
-                          'Please use **kwargs to pass this parameter.'.format(param_name),
+            warnings.warn('{0} parameter is deprecated and will be removed in 2.3 version.\n'
+                          'Please use **kwargs to pass {1} parameter.'.format(param_name, param_name[4:]),
                           LGBMDeprecationWarning)
             if param_name[4:] not in kwargs:
                 kwargs[param_name[4:]] = param
     if locals().get('strict'):
-        warnings.warn('old_strict parameter is deprecated and will be removed in 2.2 version.\n'
-                      'Please use **kwargs to pass this parameter.',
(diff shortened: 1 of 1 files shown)


python-package/lightgbm/plotting.py
```diff
@@ -373,6 +343,23 @@ def create_tree_digraph(booster, tree_index=0, show_info=None, precision=None,
     elif not isinstance(booster, Booster):
         raise TypeError('booster must be Booster or LGBMModel.')
 
+    for param_name in ['old_name', 'old_comment', 'old_filename', 'old_directory',
+                       'old_format', 'old_engine', 'old_encoding', 'old_graph_attr',
+                       'old_node_attr', 'old_edge_attr', 'old_body']:
+        param = locals().get(param_name)
+        if param is not None:
+            warnings.warn('{0} parameter is deprecated and will be removed in 2.2 version.\n'
```
chivee: Maybe relax a litter bitter? for like 2.3?

python-package/lightgbm/plotting.py
```diff
@@ -304,24 +301,22 @@ def add(root, parent=None, decision=None):
...
     add(tree_info['tree_structure'])
 
     return graph
 
 
 def create_tree_digraph(booster, tree_index=0, show_info=None, precision=None,
-                        name=None, comment=None, filename=None, directory=None,
-                        format=None, engine=None, encoding=None, graph_attr=None,
-                        node_attr=None, edge_attr=None, body=None, strict=False):
+                        old_name=None, old_comment=None, old_filename=None, old_directory=None,
```
henry0312: why do we need to change args name?
(eg. `name` -> `old_name`)

StrikerRUS: Because it's impossible to save old behavior with positional arguments and new one with `**kwards` due to duplicate argument name. For example, `name` as positional argument and `name` as `kwarg`.

StrikerRUS: Also, in such case we cannot check was it kwarg and should we show the warning.

henry0312: Ah, I see.
You leave these arguments for users who use positional args 👍

StrikerRUS: Yeah! Not sure, how popular these functions are and whether this backward compatibility will be useful for someone or not. But why not, if we can? 💪

PR #1501, opened 2018-07-10
support AppleClang with openmp

levyfan:
support #1492 
Tested with cmake 3.12.0-rc3 and Apple LLVM version 9.1.0 (clang-902.0.39.2) and libomp 5.0.1.

### Howto
* install libomp by homebrew `brew install libomp`
* download and install cmake >= 3.12.0
* mkdir build ; cd build
* cmake ..
* make -j

PR #1634, opened 2018-09-01
[docs] updated comment about gcc in build R script

StrikerRUS:
Refer to #1501, #1522.

For Apple Clang users this note is unnecessary.


jameslamb:
@StrikerRUS wait is clang supported for this project? Should we close #1523 ?

StrikerRUS:
@jameslamb Yeah! Technically, we've started to support Apple Clang since #1501.

To speak about #1523, I want to close it after this thing will happen #1612 (Apple Clang will be default compiler for macOS).

PR #1635, opened 2018-09-02
[R-package] fix examples in lgb.plot.importance

jameslamb:
In #1626 , I accidentally left the examples on `lgb.plot.importance` with normal `#` instead of Roxygen `#'` comment indicators. This means that the corresponding documentation file was malformed and `R CMD CHECK` started complaining.

Please consider this PR to fix that example documentation.

### DIFF
--- a/R-package/R/lgb.plot.importance.R
+++ b/R-package/R/lgb.plot.importance.R
@@ -1,39 +1,39 @@
 #' Plot feature importance as a bar graph
-#' 
+#'
 #' Plot previously calculated feature importance: Gain, Cover and Frequency, as a bar graph.
-#' 
+#'
 #' @param tree_imp a \code{data.table} returned by \code{\link{lgb.importance}}.
 #' @param top_n maximal number of top features to include into the plot.
 #' @param measure the name of importance measure to plot, can be "Gain", "Cover" or "Frequency".
 #' @param left_margin (base R barplot) allows to adjust the left margin size to fit feature names.
 #' @param cex (base R barplot) passed as \code{cex.names} parameter to \code{barplot}.
-#' 
+#'
 #' @details
 #' The graph represents each feature as a horizontal bar of length proportional to the defined importance of a feature.
 #' Features are shown ranked in a decreasing importance order.
-#' 
+#'
 #' @return
 #' The \code{lgb.plot.importance} function creates a \code{barplot}
 #' and silently returns a processed data.table with \code{top_n} features sorted by defined importance.
-#' 
+#'
 #' @examples
-# data(agaricus.train, package = "lightgbm")
-# train <- agaricus.train
-# dtrain <- lgb.Dataset(train$data, label = train$label)
-# 
-# params <- list(
-#     objective = "binary"
-#     , learning_rate = 0.01
-#     , num_leaves = 63
-#     , max_depth = -1
-#     , min_data_in_leaf = 1
-#     , min_sum_hessian_in_leaf = 1
-# )
-# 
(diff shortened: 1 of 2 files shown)


R-package/R/lgb.plot.importance.R
```diff
@@ -1,68 +1,68 @@
 #' Plot feature importance as a bar graph
-#'
```
jameslamb: as I've mentioned before, I have my `RStudio` (the IDE I use to edit R code) setup to automatically trim horizontal whitespace whenever I save a `.R` file. If these lines are annoying and you want me to undo them please do let me know and I'd be happy to.

jameslamb:
I clicked "rebuild" on Travis. I would be SHOCKED if the initial errors this PR's build got are related to these changes. Some Python GPU-based test was failing, which should have nothing to do with R examples. I'm guessing it's just one of those transient Travis things.

StrikerRUS:
@jameslamb The reason is here:
```
CondaHTTPError: HTTP 500 INTERNAL SERVER ERROR for url <https://conda.anaconda.org/conda-forge/linux-64/repodata.json>
Elapsed: 00:00.298675
CF-RAY: 45415cdbfda47143-ORD
A remote server error occurred when trying to retrieve this URL.
A 500-type error (e.g. 500, 501, 502, 503, etc.) indicates the server failed to
fulfill a valid request.  The problem may be spurious, and will resolve itself if you
try your request again.  If the problem persists, consider notifying the maintainer
of the remote server.
```

Conda-forge is unavailable, so the `boost` package cannot be downloaded, which leads to test failure.

Also, I should notify you that you should be ready for **frequent** `docs-test` job failures. The thing is that in this test we check all links in our docs by the LinkChecker software. Very often they become unavailable, broken; sites goes to down for maintenance; or ssl errors may occur (e.g., #1583, 1490). Don't be afraid! It's normal case. For example, Kaggle and libsvm sites "like" to yield 5xx error 😄 . Just re-run the problematic test one or two times (ideally, after one hour) or ping me for the future investigation. If this's the real problem with a link, I'll quickly create corresponding PR.

R-package/R/lgb.plot.importance.R
StrikerRUS: Maybe you can create a separate PR for this purpose for other files which haven't been edited by you?

I mean, just open every file in the package and send a PR? 😃 

Don't you mind, @Laurae2 ?

StrikerRUS:
@jameslamb No problem! Feel free to ask me any questions about tests.

R-package/R/lgb.plot.importance.R
Laurae2: @jameslamb Either you do it or I'll do it in a new PR.

I have also this setting trailing whitespace removal setting on my productive 📦 (but not on development 📦 ), so I always see how funny it gets in git to see all those whitespaces removal when moving from development 📦 to production 📦

Laurae2:
@jameslamb Tell me when to merge.

R-package/R/lgb.plot.importance.R
jameslamb: ha ok! I'll do it in a new PR. I think that type of thing really belongs in its own change.

jameslamb:
@Laurae2 I think we are good to merge! Now that I'm a collaborator, I can merge it. Big moment!

R-package/R/lgb.plot.importance.R
jameslamb: created #1642 so I don't forget

PR #1636, opened 2018-09-03
Docs & Warning on sparse categorical features

dmitryikh:
Hello,

I was trying to use text hash (random value from 0 to `Int32.MaxValue`) as a categorical feature in LightGBM and got into the trouble of memory consumption. Later I realize how lightGBM make decision rule for categorical variables. Using bitset. It seems that for every decision rule we need a bitset with length max(value). That is for `Int32.MaxValue` it leads to ~250 MiB for one decision!

Please, consider my small edits to the documentation and code to help other users to avoid such mistake. 

Thanks!

### DIFF
--- a/docs/Parameters.rst
+++ b/docs/Parameters.rst
@@ -573,6 +573,8 @@ IO Parameters
 
    -  **Note**: all values should be less than ``Int32.MaxValue`` (2147483647)
 
+   -  **Note**: using large values could be memory consuming. Tree decision rule works best when categorical features are presented by consecutive integers started from zero
+
    -  **Note**: all negative values will be treated as **missing values**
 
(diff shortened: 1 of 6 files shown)


dmitryikh:
Could you please look what's wrong with docs check?

StrikerRUS:
Hi @dmitryikh !

Thank you for the proposed docs enhancement.

Unfortunately, modifying parameters description is not so trivial. You should make your changes into the `config.h` [file](https://github.com/Microsoft/LightGBM/blob/master/include/LightGBM/config.h) and the run `python ./helper/parameter_generator.py` [script](https://github.com/Microsoft/LightGBM/blob/master/helper/parameter_generator.py). The python script will update the docs. That is how it works.

Also, I think that your valuable information should be reflected in other places where the information about categorical features appears. You can find these places from #1567.

StrikerRUS:
@guolinke What do you think about this PR? If it's valuable note, then I can finish this PR today, so it will get in 2.1.3 release.

dmitryikh:
Sorry for delay. I will have access to my laptop only tommorow evening to finish my work. But It is OK if you finish it by you selves because of deadline.

StrikerRUS:
@dmitryikh No problem! Glad that you haven't abandoned the PR! It's OK that you'll finish it in next couple of days.

PR #1637, opened 2018-09-03
fix RF's bug with MAPE & support RF with multi-class

guolinke:
fix #1579 
fix #881

PR #1649, opened 2018-09-08
new release

guolinke:
- [x] complete #1637


StrikerRUS:
@guolinke Pease take a look at https://github.com/Microsoft/LightGBM/pull/1629#issue-212222788.
I think it also should be resolved before new release.

henry0312:
And we should check https://github.com/Microsoft/LightGBM/pull/1630.
Is v2.1.3 really right?
Is it sure that we follow [Semantic versioning](https://www.python.org/dev/peps/pep-0440/#semantic-versioning)?

guolinke:
@henry0312 do you mean change the format of version?

henry0312:
@guolinke yeah, I think we will release **v2.2.0** as next version if we follow [Semantic Versioning](https://semver.org/).

guolinke:
@henry0312 why you think we need to jump a major version now?
@StrikerRUS what is your opinion?

StrikerRUS:
To be honest, I agree with @henry0312 .

But I don't know what to do with the following code (list may be not completed)... Should we review these warnings?
https://github.com/Microsoft/LightGBM/blob/4a1e5fa59a4d45a6beab2e8039125bdfa7d12027/python-package/lightgbm/__init__.py#L39-L45
https://github.com/Microsoft/LightGBM/blob/4a1e5fa59a4d45a6beab2e8039125bdfa7d12027/python-package/lightgbm/basic.py#L2097-L2098
https://github.com/Microsoft/LightGBM/blob/4a1e5fa59a4d45a6beab2e8039125bdfa7d12027/python-package/lightgbm/sklearn.py#L578-L579
https://github.com/Microsoft/LightGBM/blob/4a1e5fa59a4d45a6beab2e8039125bdfa7d12027/python-package/lightgbm/plotting.py#L351-L352
https://github.com/Microsoft/LightGBM/blob/4a1e5fa59a4d45a6beab2e8039125bdfa7d12027/python-package/lightgbm/plotting.py#L427-L428
https://github.com/Microsoft/LightGBM/blob/4a1e5fa59a4d45a6beab2e8039125bdfa7d12027/src/objective/binary_objective.hpp#L54

guolinke:
@StrikerRUS I think we should fix them accordingly.

StrikerRUS:
@guolinke Do you have thoughts how exactly?

guolinke:
@StrikerRUS my most concern is the one with apple clang. 

For the rest, we can according to the new version number to deal with them.

StrikerRUS:
@guolinke 
>my most concern is the one with apple clang.

Change the version in the warning to 2.2.1 or 2.3?

guolinke:
@StrikerRUS I think 2.2.1 is okay.

StrikerRUS:
@guolinke Me too.
I think for `pred_parameter` and `apply method` we can change nothing and remove after 2.2 release like it's said in the warning. For `plotting kwargs` and `binary objective` we can change from 2.3. to 2.4, I suppose.

henry0312:
that seems good to me.
https://github.com/Microsoft/LightGBM/pull/1649#issuecomment-420952334

guolinke:
@StrikerRUS can you help to address these warnings in another PR ?

StrikerRUS:
@guolinke Sure!

guolinke:
yeah, it had tested it.

VERSION.txt
```diff
@@ -1 +1 @@
-2.1.2.4
+2.2
```
henry0312: I think this should be **2.2.0**

guolinke:
@StrikerRUS I think the docker are checked in the "tag", not release.

StrikerRUS:
@guolinke Ah, seems you're right.

PR #1652, opened 2018-09-08
Test for build forks & remove auto-deploy due to security 

guolinke:


guolinke:
@StrikerRUS now the VSTS-CI can build from forks!
I think we can move some jobs from appveyor.

StrikerRUS:
@guolinke Cool!

But still no **rebuild** button even for owners (only close/reopen)?

Also, I remember there was a problem about status updates: https://github.com/Microsoft/vsts-agent/issues/1699#issue-340451340

I see that you want to move artifacts builds to VSTS, right? But does it support deployment to GitHub?

Yeah, I remember that I've promised to balance the load between CI services. I will try to find the optimal split so that all three CIs will finish their jobs nearly the same time.

guolinke:
@StrikerRUS It seems the only solution is re-open the PR now...
Since our release is not so frequently now, I think I can upload it to github manually. 
Another solution is pushing them to pypi directly in vsts-ci.

StrikerRUS:
@guolinke Yeah, you're right about infrequent release. To be honest, I think that manual uploading to PyPI is even preferable (since you will upload to GitHub anyway).

StrikerRUS:
OK, then lets release 2.1.3 and I'll start to work on rebalancing CI jobs.

PR #1653, opened 2018-09-09
refine CI 

guolinke:


StrikerRUS:
@guolinke Is everything OK (except the known issues like manual re-opening PRs) with VSTS builds?

guolinke:
@StrikerRUS yeah, the PR from forks should work now.

PR #1646, opened 2018-09-07
fix lgb.cv with ndcg

guolinke:

PR #1654, opened 2018-09-09
added regression test for subsetting with group and cv with group-based metrics

StrikerRUS:
Added regression test for #1646.


StrikerRUS:
Seems that VSTS check is not required?..

image

guolinke:
@StrikerRUS I will enable it

Issue #1475, opened 2018-06-25
OSError: exception: access violation writing


leftluoyi:
Windows 10 + Python 3.6.3 x64 + CUDA 8.0

Have successfully installed **LightGBM** and can run the **binary_classification** example from command prompt.

If set the classifier as **XGBClassifier**, everything is okay. When change to **LGBMClassifier**, the following exceptions is thrown.

```
 File "D:\Anaconda3\lib\site-packages\lightgbm\sklearn.py", line 695, in fit
    callbacks=callbacks)
  File "D:\Anaconda3\lib\site-packages\lightgbm\sklearn.py", line 474, in fit
    callbacks=callbacks)
  File "D:\Anaconda3\lib\site-packages\lightgbm\engine.py", line 183, in train
    booster = Booster(params=params, train_set=train_set)
  File "D:\Anaconda3\lib\site-packages\lightgbm\basic.py", line 1307, in __init__
    train_set.construct().handle,
  File "D:\Anaconda3\lib\site-packages\lightgbm\basic.py", line 860, in construct
    categorical_feature=self.categorical_feature, params=self.params)
  File "D:\Anaconda3\lib\site-packages\lightgbm\basic.py", line 714, in _lazy_init
    self.__init_from_np2d(data, params_str, ref_dataset)
  File "D:\Anaconda3\lib\site-packages\lightgbm\basic.py", line 776, in __init_from_np2d
    ctypes.byref(self.handle)))
OSError: exception: access violation writing 0xFFFFFFFF98090000
```
Any idea?

======================== UPDATE ==========================

Now I can run [this](https://github.com/Microsoft/LightGBM/blob/master/examples/python-guide/simple_example.py) example with CPU version and GPU version with lightgbm.exe, but the GPU version with python throws this exception.

guolinke:
are you on the latest version ?

guolinke:
BTW, can you provide a re-produce code with randomly generated data ?

leftluoyi:
I tried with [this](https://github.com/Microsoft/LightGBM/blob/master/examples/python-guide/simple_example.py) example. The problem still exists.

guolinke:
can you try to uninstall and re-install ? you can install version 2.1.2 from pypi directly.

leftluoyi:
@guolinke I will try to reinstall. BTW, here is what I got.
[...]

PR #1660, opened 2018-09-10
[ci] rebalanced Windows tests

StrikerRUS:
Now we have the following setting for Windows:
 - VSTS: `regular` (with examples), `sdist`, `bdist` under **VS 2017** (~10 min)
 - Appveyor: `regular` (with examples) under **VS 2015** and **MinGW** (~10-15 min, was ~40 min)

Also, MinGW version has been updated from 6.3.0 to 7.3.0.

Known bug: conda at Windows cannot handle Python 3.7 properly yet (`Importing the multiarray numpy extension module failed.  Most likely you are trying to import a failed build of numpy. Original error was: DLL load failed: The specified module could not be found.` at VSTS and `Command executed with exception:   from collections import Sequence` at Appveyor).

While working on this PR, I found weird bug. The same symptoms were described in #1475. When I tried to run **all** examples by `pytest %APPVEYOR_BUILD_FOLDER%\tests` instead of `pytest %APPVEYOR_BUILD_FOLDER%\tests\python_package_test`, two C_API tests were passed succesfully, but all other tests failed with `OSError: exception: access violation writing 0xFFFFFFFF93400000`. At present this bug is avoided by running only Python tests (without C_API). I suppose, it's somehow connected to that the same `lib_lightgbm.dll` is loaded twice from different places (from installation directory for C_API tests and from python-package directory for Python tests).

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -44,6 +44,7 @@ test_script:
   - cd %APPVEYOR_BUILD_FOLDER%\examples\python-guide
   - ps: >-
       @("import matplotlib", "matplotlib.use('Agg')") + (Get-Content "plot_example.py") | Set-Content "plot_example.py"  # prevent interactive window mode
+      (Get-Content "plot_example.py").replace('graph.render(view=True)', 'graph.render(view=False)') | Set-Content "plot_example.py"
   - ps: >-
       foreach ($file in @(Get-ChildItem *.py)) {
         python $file
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,22 +1,22 @@
 ﻿version: 2.1.2.{build}
 
+image: Visual Studio 2015
+platform: x64
 configuration:  # a trick to construct a build matrix
   - 3.5
   - 3.6
 
 environment:
   matrix:
-    - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
-      COMPILER: MSVC
-    - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
-      COMPILER: MINGW
+    - COMPILER: MSVC
+    - COMPILER: MINGW
 
 clone_depth: 50
 
 install:
   - git submodule update --init --recursive  # get `compute` folder
   - set PATH=%PATH:C:\Program Files\Git\usr\bin;=%  # delete sh.exe from PATH (mingw32-make fix)
-  - set PATH=C:\mingw-w64\x86_64-6.3.0-posix-seh-rt_v5-rev1\mingw64\bin;%PATH%
+  - set PATH=C:\mingw-w64\x86_64-7.3.0-posix-seh-rt_v5-rev0\mingw64\bin;%PATH%
   - set PYTHON_VERSION=%CONFIGURATION%
   - ps: >-
       switch ($env:PYTHON_VERSION) {
@@ -34,16 +34,13 @@ install:
   - activate test-env
 
 build_script:
-  - mkdir %APPVEYOR_BUILD_FOLDER%\build && cd %APPVEYOR_BUILD_FOLDER%\build
-  - cmake -DCMAKE_GENERATOR_PLATFORM=x64 .. && cmake --build . --target ALL_BUILD --config Release
+  - cd %APPVEYOR_BUILD_FOLDER%\python-package
+  - IF "%COMPILER%"=="MINGW" (
+    python setup.py install --mingw)
+    ELSE (
+    python setup.py install)
 
 test_script:
-  - pytest %APPVEYOR_BUILD_FOLDER%\tests\c_api_test\test_.py
-  - cd %APPVEYOR_BUILD_FOLDER%\python-package && python setup.py sdist --formats gztar
-  - IF "%COMPILER%"=="MINGW" (
-    pip install %APPVEYOR_BUILD_FOLDER%\python-package\dist\lightgbm-%LGB_VER%.tar.gz --install-option=--mingw -v)
-    ELSE (
-    pip install %APPVEYOR_BUILD_FOLDER%\python-package\dist\lightgbm-%LGB_VER%.tar.gz -v)
   - pytest %APPVEYOR_BUILD_FOLDER%\tests\python_package_test
   - cd %APPVEYOR_BUILD_FOLDER%\examples\python-guide
   - ps: >-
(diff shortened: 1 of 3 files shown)


.appveyor.yml
```diff
@@ -54,13 +50,7 @@ test_script:
         if ($LastExitCode -ne 0) { $host.SetShouldExit($LastExitCode) }
       }  # run all examples
   - IF "%COMPILER%"=="MINGW" appveyor exit  # skip all further steps
-  - cd %APPVEYOR_BUILD_FOLDER%\python-package && python setup.py bdist_wheel --plat-name=win-amd64 --universal
 
 artifacts:
-- path: Release/lib_lightgbm.dll
+- path: python-package\compile\windows\x64\DLL\lib_lightgbm.dll
   name: Library
```
StrikerRUS: May I leave this artifact for my personal purpose? Due to some reason I have installed only VS 2013 on my machine, but sometimes need to test the latest library file from `master`.

guolinke: sure

StrikerRUS: Thanks!

guolinke:
it seems azure pipeline is release and free for open source projects. maybe we can move on it?

StrikerRUS:
Wow, how many CIs ! 😃 

My opinion is that we need to balance between our current CIs (VSTS, Appveyor, Travis), because we have enough experience in them, and after that give a try to Azure.

BTW, which OS are supported at Axure?

guolinke:
@StrikerRUS I think Azure Pipeline is actually the vsts-ci, but it is free for open-source project.

StrikerRUS:
Sorry, am I right, that you want to drop VSTS  and move to Azure?

guolinke:
@StrikerRUS Yeah, I think they share the same config. But Azure Pipeline may is able to add community users to manage, while vsts-ci can't.

StrikerRUS:
@guolinke Got it!

Seems Azure Pipelines is very young! 😃 They have been announced yesterday: http://disq.us/p/1vm7rh5

>I think they share the same config. 

Does it mean that I can continue rebalancing (at present I'm working on Linux, practically finished)? And then we'll need just to perform some minor management work, like register an account, rename the config file, etc?

StrikerRUS:
Yesterday I noticed that the badge has been changed, now the name of service.

image

guolinke:
@StrikerRUS yeah, the free version still has 10 concurrent jobs.

guolinke:
@StrikerRUS just move to it, seems everything is fine.

StrikerRUS:
@guolinke I've noticed it: 404 error happened just right when I wanted to check my rebalancing for Linux  status 😄 

GPU failed. Seems it doesn't perform `--recursive` clone.

guolinke:
@StrikerRUS now you can admin the LightGBM-CI on Azure Pipeline 😄

StrikerRUS:
@guolinke Thank you very much! Azure Pipeline seems to be more convenient than VSTS: there are no useless things on the screen, only important tabs and settings, and finally cancel and rebuild buttons!

StrikerRUS:
@guolinke Hmm... I can rebuild PR from you, but cannot rebuild master.

image


image

StrikerRUS:
@guolinke Also, it seems to be they use outdated gnutls library instead of openssl. Therefor, phantom error while pulling submodule happens.

```
fatal: unable to access 'https://github.com/Microsoft/LightGBM/': gnutls_handshake() failed: Error in the pull function.
```

DavidStaheli:
Hi @StrikerRUS and @guolinke.  I'm a PM from the Azure Pipelines team.  Thanks a lot for giving it a try.  Please let me know if anything is missing that you need.  We want to help however we can.

DavidStaheli:
We'll look into these and get each of them addressed.  Thanks a lot, @StrikerRUS.

StrikerRUS:
@DavidStaheli Many thanks!

I've recollected another inconvenient case which consists in the fact that you need to run `chmod` before executing shell script (at both Ubuntu and macOS).
https://github.com/Microsoft/LightGBM/blob/6214657aed2c9c682e1937748d39d5bf78698164/.vsts-ci.yml#L94-L100

DavidStaheli:
Thanks @StrikerRUS!  Looking into the chmod issue, I've seen reports of people needing to add the execute permission before committing & pushing the file, and then the git fetch will automatically re-apply the execute permission.  [Here's an example](https://stackoverflow.com/questions/40978921/how-to-add-chmod-permissions-to-file-in-git/40979016).  Do you know if that's already been done for your .sh files?  If so, it sounds like something _we_ need to address.

StrikerRUS:
@DavidStaheli Thank you very much! The problem was solved using the solution from your link. It's rather strange that, for example, Travis doesn't suffer from the same problem. Anyway, thank you for the solution and sorry for disturbing you.

brcrista:
@StrikerRUS I see you are running with V0 of the task.  V1 of the task doesn't require you to create an environment to use Conda.  So as a workaround, try upgrading to V1 of the task and make sure "Create a custom environment" isn't checked.

StrikerRUS:
Thanks @brcrista for the reply! But could you please elaborate your workaround? Simple change `CondaEnvironment@0` to `CondaEnvironment@1` here
https://github.com/Microsoft/LightGBM/blob/64002b2c0de9d2308a5bb90e8e91fca184fcb744/.vsts-ci.yml#L82-L87
broke entire conda setup.

brcrista:
Look at the YAML snippet at https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/package/conda-environment?view=vsts

I'm saying `createCustomEnvironment: false` and omit `environmentName`.  Let me know if that works

StrikerRUS:
@brcrista I've already seen that snippet. Unfortunately, we need named environment in our test scripts, so I'd better leave my current hotfix till you'll release a real fix for the permission error.

brcrista:
Can you try running on Hosted Linux Preview instead of Hosted Ubuntu 16.04?  I believe that should work.

StrikerRUS:
@brcrista I'm facing this issue on macOS, not Linux:
https://github.com/Microsoft/LightGBM/blob/64002b2c0de9d2308a5bb90e8e91fca184fcb744/.vsts-ci.yml#L62-L87

brcrista:
sorry, never mind then.

PR #1663, opened 2018-09-12
[ci] rebalanced Linux tests

StrikerRUS:
Now we have the following setting for Linux:

- VSTS: `regular` (with examples), `sdist`, `bdist`, `if-else`, `mpi-source`, `gpu-source` with **gcc** (6 jobs, were 8)

- Travis: `regular` (with examples), `sdist`, `bdist`, `pylint`, `check-docs`, `mpi-source`, `mpi-pip`, `gpu-source`, `gpu-pip` with **Clang** (9 jobs, were 11)

Travis can easy run 15 concurrent jobs, so there is no need in reduction of job numbers. In contrast, at Azure Pipelines we should try to stick to 10 tasks.

Use the same name `BUILD_SOURCESDIRECTORY` across build scripts: https://github.com/MicrosoftDocs/vsts-docs/issues/665

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,16 +11,14 @@ os:
 
 env:
   global:  # default values
-    - COMPILER=gcc
+    - COMPILER=clang
     - PYTHON_VERSION=3.7
   matrix:
     - TASK=regular PYTHON_VERSION=3.6
-    - TASK=regular COMPILER=clang
-    - TASK=pylint
-    - TASK=check-docs
-    - TASK=if-else
     - TASK=sdist PYTHON_VERSION=2.7
     - TASK=bdist
+    - TASK=pylint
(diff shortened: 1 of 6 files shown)


guolinke:
@StrikerRUS actually, I think we may need to add a tmp oxs test with gcc in vsts side (for 2.1.3 release), since we don't have the artifacts from travis.

StrikerRUS:
@guolinke Sure! I'm planning to rebalance macOS part tomorrow. Will switch the `COMPILER` variable to **gcc** for 2.1.3.

StrikerRUS:
Oh, forgot to mention that I believe we should run `check-docs` **only** at Travis due to it fails rather frequently and everyone should be able easy re-run it. Seems it's possible only at Travis.

PR #1665, opened 2018-09-12
[ci] rebalanced macOS tests

StrikerRUS:
Now we have the following setting for macOS:

- VSTS: `regular` (with examples), `sdist`, `bdist`, with **gcc** (3 jobs, nothing changed)

- Travis: `regular` (with examples), `sdist`, `bdist`, `mpi-source`, `mpi-pip` with **Clang** (5 jobs, nothing changed)

The main change is that now compilers can be switched via `COMPILER` variable at both VSTS and Travis.

### DIFF
--- a/.vsts-ci.yml
+++ b/.vsts-ci.yml
@@ -59,31 +59,39 @@ phases:
 ###########################################
 - phase: MacOS
 ###########################################
+  variables:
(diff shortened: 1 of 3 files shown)


guolinke:
@StrikerRUS great!

PR #1672, opened 2018-09-16
Enforce R-package dependencies in build_r.R before compilation begins

zkurtz:
I found myself forgetting to install a couple of R-package dependencies repeatedly on successive LightGBM installations. This wasted some time because compilation runs for quite a while before getting to the part where these packages are required.

### DIFF
--- a/build_r.R
+++ b/build_r.R
@@ -5,6 +5,22 @@
 # Sys.setenv("CXX" = "/usr/local/bin/g++-8")
 # Sys.setenv("CC" = "/usr/local/bin/gcc-8")
 
+# Identify R-packages required for build_r.R to complete
+pkgs = list(
+    installed = rownames(installed.packages()),
+    required = c("testthat", "roxygen2", "devtools")
+)
+pkgs$missing = setdiff(pkgs$required, pkgs$installed)
+if(length(pkgs$missing) > 0){
+    pkgs_string = paste0("    ", pkgs$missing, "\n")
+    message = paste0(
+        "The following R-packages are required for ",
+        "the LightGBM R-package install: \n",
+        paste0(pkgs_string, collapse = ''),
+        "Please try `Rscript build_r.R` again after installing them.")
+    stop(message)
+}
+
 # R returns FALSE (not a non-zero exit code) if a file copy operation
 # breaks. Let's fix that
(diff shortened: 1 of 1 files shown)


build_r.R
```diff
@@ -5,6 +5,22 @@
 # Sys.setenv("CXX" = "/usr/local/bin/g++-8")
 # Sys.setenv("CC" = "/usr/local/bin/gcc-8")
 
+# Identify R-packages required for build_r.R to complete
+pkgs = list(
+    installed = rownames(installed.packages()),
+    required = c("testthat", "roxygen2", "devtools")
+)
+pkgs$missing = setdiff(pkgs$required, pkgs$installed)
+if(length(pkgs$missing) > 0){
```
Laurae2: add spacing

build_r.R
```diff
@@ -5,6 +5,22 @@
...
+    pkgs_string = paste0("    ", pkgs$missing, "\n")
+    message = paste0(
+        "The following R-packages are required for ",
+        "the LightGBM R-package install: \n",
+        paste0(pkgs_string, collapse = ''),
```
Laurae2: "" instead of '' for consistency

build_r.R
```diff
@@ -5,6 +5,22 @@
 # Sys.setenv("CXX" = "/usr/local/bin/g++-8")
 # Sys.setenv("CC" = "/usr/local/bin/gcc-8")
 
+# Identify R-packages required for build_r.R to complete
```
jameslamb: Thanks @zkurtz , and apologies for the inconvenience.

But this is excessive complexity. Can you just add the following to the top of the script:

```
library(testthat)
library(roxygen2)
library(devtools)
```

Then anyone who doesn't have those will have this script immediately fail, and can just install them with `install.packages`

jameslamb: also please keep an eye on #629 ! We are trying to get CRAN-ready as fast as possible so hopefully you soon won't need to build from source.

jameslamb:
@zkurtz looks good to me! Thanks for the change

@Laurae2 I will merge when all the CI stuff builds

Laurae2:
@jameslamb To merge?

jameslamb:
@Laurae2 yep looks good to me! I'll merge

PR #1676, opened 2018-09-18
using a different version in master branch, compared with release

guolinke:


.travis.yml
```diff
@@ -42,6 +42,7 @@ before_install:
   - export BUILD_DIRECTORY="$TRAVIS_BUILD_DIR"
   - if [[ $TRAVIS_OS_NAME == "osx" ]]; then
         export OS_NAME="macos";
+        export COMPILER="gcc"
     else
         export OS_NAME="linux";
```
StrikerRUS: I think that we should remove "global" `COMPILER` (L14) and add here `export COMPILER="clang";`, since we use different compilers for different OS. At present it's seems to be a little bit confusing.

StrikerRUS:
@guolinke  The problem is in `;` at the end of a line

guolinke:
Thanks @StrikerRUS

PR #1684, opened 2018-09-20
auto year in copyright of readthedocs

guolinke:


docs/conf.py
```diff
@@ -21,6 +21,7 @@
 import sys
 import sphinx
 from sphinx.errors import VersionRequirementError
+import datetime
```
StrikerRUS: Please move this import upper to other similar imports

docs/conf.py
```diff
@@ -71,7 +72,7 @@
 
 # General information about the project.
 project = 'LightGBM'
-copyright = '2017, Microsoft Corporation'
+copyright = '%s, Microsoft Corporation' %(str(datetime.datetime.now().year))
```
StrikerRUS: We don't check PEP8 in docs , but I think it's better to avoid making errors in the new code.
Here is `E225 missing whitespace around operator` and outer brackets are excess.

PR #1694, opened 2018-09-23
[R-package] Fix build_r.R failing in Windows

### LINKED ISSUE #1693: [R-package] build_r.R fails in Windows
With Windows and MinGW, build_r.R fails with the following error:

```
$ Rscript build_r.R
Error: Can't find 'E:\sandbox\LightGBM\lightgbm_r\'.
Execution halted
```

Laurae2:
Fixes #1693

### DIFF
--- a/build_r.R
+++ b/build_r.R
@@ -38,7 +38,6 @@ result <- file.copy(from = "src/",
 
 result <- file.copy(from = "CMakeLists.txt",
                     to = file.path("lightgbm_r", "inst", "bin/"),
-                    recursive = TRUE,
                     overwrite = TRUE)
 .handle_result(result)
 
--- a/build_r.R
+++ b/build_r.R
@@ -31,8 +31,8 @@ result <- file.copy(from = "include/",
 .handle_result(result)
 
 result <- file.copy(from = "src/",
-                    to = file.path("lightgbm_r", "src/")
-                    , recursive = TRUE,
+                    to = file.path("lightgbm_r", "src/"),
+                    recursive = TRUE,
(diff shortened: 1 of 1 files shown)


build_r.R
```diff
@@ -1,4 +1,4 @@
-# for macOS users who have decided to use gcc
+# Fr macOS users who have decided to use gcc
```
StrikerRUS: **For**

build_r.R
```diff
@@ -7,76 +7,63 @@
...
 
-result <- file.copy(
-    from = "CMakeLists.txt"
-    , to = file.path("lightgbm_r", "inst", "bin/")
-    , recursive = TRUE
-    , overwrite = TRUE
-)
+result <- file.copy(from = "CMakeLists.txt",
+                    to = file.path("lightgbm_r", "inst", "bin/"),
+                    overwrite = TRUE)
```
jameslamb: is this what was causing the problem? (looks like you removed the `recursive = TRUE`)

A lot of these are formatting changes, so trying to find whichever thing was causing the issue

Laurae2: I had two errors:

* on `devtools::document(pkg = "lightgbm_r/")` (because of the trailing slash)
* on the file with recursive (it causes a warning, which causes an error when R is setup to throw an error on warnings)

Laurae2:
@jameslamb Do you have the rights to delete the branch fix-#1693 ? It seems I cannot delete it anymore (getting GitHub 404 error).

jameslamb:
Thanks @StrikerRUS !

PR #1699, opened 2018-09-25
[python] copy old Booster's attributes to new one in refit method

### LINKED ISSUE #1657: [python] copy info from old Booster to new one in refit method
Refer to the discussion in #1629.

StrikerRUS:
Fixed #1657.

I'm confused about the need to copy
https://github.com/Microsoft/LightGBM/blob/a760eae4ff06fa8d522ed53e7df8c31d9e8298f0/python-package/lightgbm/basic.py#L1438

Also, I'm not sure whether `__attr` should be **copied**.

But I suppose that it's logically correct to reset the info connected to the previous data (e.g. `__train_data_name`, `best_iteration`, `valid_sets`, etc.) after calling the `refit` method.

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -2131,6 +2131,8 @@ class Booster(object):
             ptr_data,
             ctypes.c_int(nrow),
             ctypes.c_int(ncol)))
+        new_booster.network = self.network
+        new_booster.__attr = self.__attr
         return new_booster
 
     def get_leaf_output(self, tree_id, leaf_id):
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -1736,7 +1736,7 @@ class Booster(object):
         is_finished = ctypes.c_int(0)
         if fobj is None:
             if self.__set_objective_to_none:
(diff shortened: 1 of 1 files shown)


StrikerRUS:
@guolinke What is your opinion on `__attr` and should `__set_objective_to_none` be copied?

guolinke:
I think the `__attr` can be copied. For the `__set_objective_to_none`, as refit cannot support the custom objective, maybe we should check it?

StrikerRUS:
Sorry, I'm not sure that I fully understood the part about `__set_objective_to_none`.

guolinke:
@StrikerRUS it this ready to merge ?
BTW, it seems appveyor is down..

StrikerRUS:
@guolinke
Please, clarify what did you mean under 
>  maybe we should check it [__set_objective_to_none]?

 I suppose, Appveyor fails due to the new release (0.20) of scikit-learn which brought new (Deprecation and Future) warnings. I will take a look at this today.

guolinke:
@StrikerRUS 
__set_objective_to_none means there exists a custom objective function.
In this case, we cannot use refit.
So when calling refit in a booster with __set_objective_to_none, we should throw a error.

PR #1710, opened 2018-09-28
add indices in shuffle model.

guolinke:


python-package/lightgbm/basic.py
```diff
@@ -1945,15 +1945,26 @@ def save_model(self, filename, num_iteration=None, start_iteration=0):
...
+    def shuffle_models(self, start_iteration=0, end_iteration=-1):
         """Shuffle models.
 
+        Parameters
+        ----------
+        start_iteration : int, optional (default=0)
+            Index of the iteration that will start to shuffle.
+        end_iteration : int, optional (default=-1)
+            The last iteration that will be shuffled.
+            If <= 0, means the last iterations.
```
StrikerRUS: `… last iteration.`
Without **s**.

src/boosting/gbdt.h
```diff
@@ -70,16 +70,21 @@ class GBDT : public GBDTBase {
...
   }
 
-  void ShuffleModels() override {
+  void ShuffleModels(int start_iter, int end_iter) override {
     int total_iter = static_cast<int>(models_.size()) / num_tree_per_iteration_;
+    start_iter = std::max(0, start_iter);
+    if (end_iter <= 0) {
+      end_iter = total_iter;
+    }
+    end_iter = std::min(total_iter, start_iter);
```
StrikerRUS: Maybe
```
end_iter = std::min(total_iter, end_iter); 
```
?

PR #1716, opened 2018-09-30
test changing pool

ganik:


guolinke:
@ganik as the glibc problem is fixed, I think this PR could be closed.

PR #1719, opened 2018-10-01
fix indices type in csr and csc

### LINKED ISSUE #1689: Silent failure when sparse training matrix has > 2^31 non-zero elements
When training data is in the form of a scipy.sparse.csr_matrix and the number of non-zero values exceeds 2^31 - 1 (haven't checked precisely but max_int32 is the only thing that makes sense here), model predictions suddenly become awful. I strongly suspect there's a 32 bit integer used as an index into the non-zero elements of the sparse matrix, which ends up mixing up data from different columns incoherently.  

## Environment info
Operating System: Ubuntu 14.04.4
CPU: 4x Intel Xeon 24 core processors
Python version: 2.7.6
LightGBM built from source on commit 44ed4270eef68f0dd2ba093d4e5aa34460ddeecc

## Reproducible examples
Note the predictions improve as we add more training data until the number of non-zero entries is too large, at which point the predictions are very close to 0.5 and the model is providing essentially no predictive power:
results_1000
results_10000
results_100000
results_5300000
results_5400000

## Script to produce examples:
```
import matplotlib
matplotlib.use('Agg')

import numpy as np
import lightgbm as lgb
import scipy.sparse
from scipy.special import logit, expit

np.random.seed(0)

def true_proba(X):
    return X[:, 0]

def simulate_y(X):
    probs = true_proba(X)
    return (np.random.random(len(X)) < probs)

def simulate_X(nrows, n_sparse_cols, non_zero_frac):
    X = np.zeros((nrows, n_sparse_cols + 1))
    X[:, 0] = np.random.random(nrows)
    for i in range(n_sparse_cols):
        X[:, i+1] = (np.random.random(nrows) < non_zero_frac)
    return X

def get_data(nrows, n_sparse_cols, non_zero_frac):
    X = simulate_X(nrows, n_sparse_cols, non_zero_frac)
    y = simulate_y(X)
    X = scipy.sparse.csr_matrix(X)
    return X, y

def run_study(figfile, params, nrows, n_sparse_cols, non_zero_frac=0.2, n_test=1000):
    X, y = get_data(nrows, n_sparse_cols, non_zero_frac)
    print "With %s rows, got %s non-zero entries"%(nrows, X.nnz)
    data = lgb.Dataset(X, label=y)
    model = lgb.train(params, data, num_boost_round=10)
    X_test = simulate_X(n_test, n_sparse_cols, non_zero_frac)
    preds = model.predict(X_test)
    plt.plot(true_proba(X_test), preds, 'o')
    plt.title("nnz in train = %s"%(X.nnz))
    plt.xlabel("True probability")
    plt.ylabel("Model prediction")
    plt.savefig(figfile)
    plt.clf()

params = {
    'learning_rate': 1.0,
    'objective': 'binary',
    'max_depth': 2,
}

for nrows in [1000, 10000, 100000, 5400000, 5300000]:
    run_study('results_%s.png'%(nrows), params, nrows, 2000)
```

I'll be happy to provide further environment information if needed and can re-run modified versions of the script if more output or model information is needed.

guolinke:
fix #1689

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -548,7 +548,7 @@ class _InnerPredictor(object):
             ptr_indptr, type_ptr_indptr, __ = c_int_array(csr.indptr)
             ptr_data, type_ptr_data, _ = c_float_array(csr.data)
 
-            assert csr.shape[1] <= np.iinfo(np.int32).max
+            assert csr.shape[1] <= MAX_INT32
             csr.indices = csr.indices.astype(np.int32, copy=False)
 
             _safe_call(_LIB.LGBM_BoosterPredictForCSR(
@@ -599,7 +599,7 @@ class _InnerPredictor(object):
         ptr_indptr, type_ptr_indptr, __ = c_int_array(csc.indptr)
         ptr_data, type_ptr_data, _ = c_float_array(csc.data)
 
-        assert csc.shape[0] <= np.iinfo(np.int32).max
+        assert csc.shape[0] <= MAX_INT32
         csc.indices = csc.indices.astype(np.int32, copy=False)
 
         _safe_call(_LIB.LGBM_BoosterPredictForCSC(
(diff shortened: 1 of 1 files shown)


guolinke:
Thanks @bauks ! Can you try this fix ?

bauks:
> Thanks @bauks ! Can you try this fix ?

@guolinke @StrikerRUS This fix works, thanks for resolving it quickly!

PR #1722, opened 2018-10-01
[python] do not modify users params

### LINKED ISSUE #1706: The Python Training API should not alter "params" argument in place
It appears that the training API's ` train()` method will attempt to modify the `params` argument in place with `pop()`, new key assignment, etc. when trying to use the categorical columns from a Pandas dataframe. This is likely to cause unexpected behaviour when the `params` values are reused. For instance, I have the following params dict:
```
reg_params = {
        "objective": "regression",
        "metric": "rmse",
        "max_depth": 10,
        "min_child_samples": 20,
        "lambda_l1": 1,
        "lambda_l2": 1,
        "num_leaves": 150,
        "num_iterations": 20000,
        "learning_rate": 0.01,
        "subsample": 0.8,
        "colsample_bytree": 0.8,
        "verbosity": -1
    }
```
And I used the dict in a KFold setup in which I used the same parameters in each fold. After one fold, the dict becomes:
`{'objective': 'regression', 'metric': 'rmse', 'max_depth': 10, 'min_child_samples': 20, 'lambda_l1': 1, 'lambda_l2': 1, 'num_leaves': 150, 'learning_rate': 0.01, 'subsample': 0.8, 'colsample_bytree': 0.8, 'verbosity': -1, 'categorical_column': [0, 2, 3, 5, 6, 7, 8, 9, 14, 17, 18, 20, 21, 22, 27, 28, 29]}`
Notice that the `num_iterations` value is lost despite not altering the dict manually anywhere.

The expected behaviour should be that the API reads the params dict and makes a run-specific copy of it instead of altering its value in place. Something like
```
def train(og_params, *):
    params = og_params.copy()
```
should do the trick.

StrikerRUS:
Fixed #1706.

### DIFF
--- a/tests/python_package_test/test_engine.py
+++ b/tests/python_package_test/test_engine.py
@@ -438,7 +438,8 @@ class TestEngine(unittest.TestCase):
         lgb_train = lgb.Dataset(X_train, y_train)
         # shuffle = False, override metric in params
         params_with_metric = {'metric': 'l2', 'verbose': -1}
-        cv_res = lgb.cv(params_with_metric, lgb_train, num_boost_round=10, nfold=3, stratified=False, shuffle=False,
+        cv_res = lgb.cv(params_with_metric, lgb_train, num_boost_round=10,
+                        nfold=3, stratified=False, shuffle=False,
                         metrics='l1', verbose_eval=False)
         self.assertIn('l1-mean', cv_res)
         self.assertNotIn('l2-mean', cv_res)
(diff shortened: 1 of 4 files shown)


tests/python_package_test/test_engine.py
```diff
@@ -464,21 +465,21 @@ def test_cv(self):
...
-        cv_res_lambda = lgb.cv(params_lambdarank, lgb_train, num_boost_round=10, nfold=3,
-                               verbose_eval=False)
-        self.assertEqual(len(cv_res_lambda), 2)
-        self.assertFalse(np.isnan(cv_res_lambda['ndcg@3-mean']).any())
         # ... with l2 metric
         cv_res_lambda = lgb.cv(params_lambdarank, lgb_train, num_boost_round=10, nfold=3,
                                metrics='l2', verbose_eval=False)
         self.assertEqual(len(cv_res_lambda), 2)
         self.assertFalse(np.isnan(cv_res_lambda['l2-mean']).any())
+        # ... with NDCG (default) metric
```
StrikerRUS: Switching `# ... with l2 metric` and `# ... with NDCG (default) metric` is needed to ensure that `params_lambdarank` is not modified during cv.

noklam:
Was just looking into the same issue, found parameters is changed when I check my log.

noklam:
Thanks,  I am aware of it,  the fix is just making a copy of params,  nice and clean fix.

StrikerRUS:
@noklam Ah, OK. Thank you!

PR #1727, opened 2018-10-03
v2.2.1 release

guolinke:
Since 2.2.0 requires higher glibc, and therefore cannot be run in some systems. 
I think we release the new binaries ASAP.

- [x] #1726 
- [x] #1728

### DIFF
--- a/R-package/DESCRIPTION
+++ b/R-package/DESCRIPTION
@@ -2,7 +2,7 @@ Package: lightgbm
 Type: Package
 Title: Light Gradient Boosting Machine
 Version: 2.2.1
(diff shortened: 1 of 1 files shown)


StrikerRUS:
@guolinke What about macOS? Is everything OK? If so, then we are switching to AppleClang as it was scheduled (warning should be updated, at least its type), right?

guolinke:
I think most macos users used the more recent packages. We can downgrade it in the next release, if necessary.

guolinke:
@StrikerRUS you can approve this if you think it is ready.

PR #1728, opened 2018-10-03
[python] updated warning for macOS users

StrikerRUS:


python-package/lightgbm/__init__.py
```diff
@@ -35,11 +35,12 @@
            'print_evaluation', 'record_evaluation', 'reset_parameter', 'early_stopping',
            'plot_importance', 'plot_metric', 'plot_tree', 'create_tree_digraph']
 
+# REMOVEME: remove warning after 2.3.0 version release
```
StrikerRUS: Is `2.3.0` version OK?

guolinke: yeah

PR #1735, opened 2018-10-07
average predictions for constant features

### LINKED ISSUE #1697: mean prediction for data with all constant features
Getting this even if follow the advice.  Certainly this particular case has a constant column, but LightGBM should predict the average model at least as a last resort, not result in a fatal error in this simplest case.

[lgb_min_data_issue.zip](https://github.com/Microsoft/LightGBM/files/2412973/lgb_min_data_issue.zip)

guolinke:
to fix #1697

### DIFF
--- a/src/io/dataset.cpp
+++ b/src/io/dataset.cpp
@@ -226,7 +226,7 @@ void Dataset::Construct(
     }
   }
   if (used_features.empty()) {
-    Log::Warning("There are not meaningful features, as all feature values are constant.");
+    Log::Warning("There are no meaningful features, as all feature values are constant.");
   }
   auto features_in_group = NoGroup(used_features);
 
--- a/src/io/dataset.cpp
+++ b/src/io/dataset.cpp
@@ -227,15 +227,10 @@ void Dataset::Construct(
   }
   if (used_features.empty()) {
     Log::Warning("There are not meaningful features, as all feature values are constant.");
-    num_features_ = 0;
-    num_groups_ = 0;
-    group_bin_boundaries_.clear();
-    group_bin_boundaries_.push_back(0);
-    return;
   }
   auto features_in_group = NoGroup(used_features);
 
-  if (io_config.enable_bundle) {
+  if (io_config.enable_bundle && !used_features.empty()) {
     features_in_group = FastFeatureBundling(bin_mappers,
                                             sample_non_zero_indices, num_per_col, total_sample_cnt,
                                             used_features, io_config.max_conflict_rate,
--- a/src/io/dataset.cpp
+++ b/src/io/dataset.cpp
@@ -226,10 +226,12 @@ void Dataset::Construct(
     }
   }
   if (used_features.empty()) {
-    Log::Fatal("Cannot construct Dataset since there are no useful features.\n"
-               "It should be at least two unique rows.\n"
-               "If the num_row (num_data) is small, you can set min_data=1 and min_data_in_bin=1 to fix this.\n"
-               "Otherwise, please make sure you are using the right dataset");
+    Log::Warning("There are not meaningful features, as all feature values are constant.");
+    num_features_ = 0;
+    num_groups_ = 0;
+    group_bin_boundaries_.clear();
+    group_bin_boundaries_.push_back(0);
+    return;
   }
   auto features_in_group = NoGroup(used_features);
 
@@ -323,14 +325,16 @@ void Dataset::Construct(
 
 void Dataset::FinishLoad() {
   if (is_finish_load_) { return; }
-  OMP_INIT_EX();
-  #pragma omp parallel for schedule(guided)
-  for (int i = 0; i < num_groups_; ++i) {
-    OMP_LOOP_EX_BEGIN();
-    feature_groups_[i]->bin_data_->FinishLoad();
-    OMP_LOOP_EX_END();
+  if (num_groups_ > 0) {
+    OMP_INIT_EX();
+#pragma omp parallel for schedule(guided)
+    for (int i = 0; i < num_groups_; ++i) {
+      OMP_LOOP_EX_BEGIN();
+      feature_groups_[i]->bin_data_->FinishLoad();
+      OMP_LOOP_EX_END();
+    }
+    OMP_THROW_EX();
   }
-  OMP_THROW_EX();
   is_finish_load_ = true;
 }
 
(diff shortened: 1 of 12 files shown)


src/io/dataset.cpp
```diff
@@ -226,14 +226,11 @@ void Dataset::Construct(
     }
   }
   if (used_features.empty()) {
-    Log::Fatal("Cannot construct Dataset since there are no useful features.\n"
-               "It should be at least two unique rows.\n"
-               "If the num_row (num_data) is small, you can set min_data=1 and min_data_in_bin=1 to fix this.\n"
-               "Otherwise, please make sure you are using the right dataset");
+    Log::Warning("There are not meaningful features, as all feature values are constant.");
```
StrikerRUS: `not` -> `no`

tests/python_package_test/test_engine.py
```diff
@@ -739,3 +739,59 @@ def test_mape_dart(self):
         pred = gbm.predict(X)
         pred_mean = pred.mean()
         self.assertGreater(pred_mean, 18)
+
+    def test_constant_features_regression(self):
+        x = [1, 1, 1, 1]
+        y = [0, 10, 0, 10]
+
+        X_train = np.array(x).reshape(len(x), 1)
```
StrikerRUS: I think it could be `np.ones((4, 1))`.

tests/python_package_test/test_engine.py
```diff
@@ -739,3 +739,59 @@ def test_mape_dart(self):
...
+        y_train = np.array(y)
+        lgb_train = lgb.Dataset(X_train, y_train)
+        lgb_eval = lgb.Dataset(X_train, y_train)
```
StrikerRUS: Is `lgb_eval` needed here?

tests/python_package_test/test_engine.py
```diff
@@ -739,3 +739,59 @@ def test_mape_dart(self):
...
+        }
+        gbm = lgb.train(params, lgb_train,
+                        num_boost_round=2,
+                        valid_sets=lgb_eval,
+                        verbose_eval=True)
+        pred = gbm.predict(X_train)
+        self.assertAlmostEqual(pred[0], 5, places=5)
+        self.assertAlmostEqual(pred[1], 5, places=5)
+        self.assertAlmostEqual(pred[0], 5, places=5)
+        self.assertAlmostEqual(pred[1], 5, places=5)
```
StrikerRUS: It can be one-line check:
```
np.testing.assert_almost_equal(pred, np.mean(y_train))
```

tests/python_package_test/test_engine.py
```diff
@@ -739,3 +739,59 @@ def test_mape_dart(self):
...
+
+    def test_constant_features_binary(self):
+        x = [1, 1, 1, 1]
+        y = [0, 1, 0, 1]
+
+        X_train = np.array(x).reshape(len(x), 1)
```
StrikerRUS: I think it could be `np.ones((4, 1))`.

tests/python_package_test/test_engine.py
```diff
@@ -739,3 +739,59 @@ def test_mape_dart(self):
...
+        y_train = np.array(y)
+        lgb_train = lgb.Dataset(X_train, y_train)
+        lgb_eval = lgb.Dataset(X_train, y_train)
```
StrikerRUS: Is `lgb_eval` needed here?

tests/python_package_test/test_engine.py
```diff
@@ -739,3 +739,59 @@ def test_mape_dart(self):
...
+        }
+        gbm = lgb.train(params, lgb_train,
+                        num_boost_round=2,
+                        valid_sets=lgb_eval,
+                        verbose_eval=True)
+        pred = gbm.predict(X_train)
+        self.assertAlmostEqual(pred[0], 0.5, places=5)
+        self.assertAlmostEqual(pred[1], 0.5, places=5)
+        self.assertAlmostEqual(pred[0], 0.5, places=5)
+        self.assertAlmostEqual(pred[1], 0.5, places=5)
```
StrikerRUS: It can be one-line check:
```
np.testing.assert_almost_equal(pred, np.mean(y_train))
```

tests/python_package_test/test_engine.py
```diff
@@ -24,6 +24,28 @@ def multi_logloss(y_true, y_pred):
     return np.mean([-math.log(y_pred[i][y]) for i, y in enumerate(y_true)])
 
 
+def test_constant_features(y_true, expected_pred, more_params):
```
guolinke: @StrikerRUS  how to stop the test on this function?

guolinke:
@pseudotensor added

tests/python_package_test/test_engine.py
StrikerRUS: @guolinke Sorry, what do you mean under "to stop the test"? A breakpoint? Or skip all further functions?

If latter, you can simply rename all unneeded functions so that they don't have `test_` prefix in their names.

guolinke: Thanks, I fixed that by default parameters.

tests/python_package_test/test_engine.py
```diff
@@ -739,3 +739,57 @@ def test_mape_dart(self):
...
+                'min_data_in_bin': 1,
+                'boost_from_average': True
+            }
+            params.update(more_params)
+            lgb_train = lgb.Dataset(X_train, y_train, params=params)
+            gbm = lgb.train(params, lgb_train,
+                            num_boost_round=2)
+            pred = gbm.predict(X_train)
+            for i in range(pred.shape[0]):
+                np.testing.assert_almost_equal(pred[i], expected_pred)
```
StrikerRUS: It's possible to omit the loop and write just
```
 np.testing.assert_almost_equal(pred, expected_pred) 
```

tests/python_package_test/test_engine.py
StrikerRUS: Oh, I see. However, such approach adds dummy record into `pytest`'s overall number of tests.

tests/python_package_test/test_engine.py
guolinke: I tried this, but it cannot work for multi-class, which compares 2d with 1d.

StrikerRUS: Oops, sorry. Then it's possible to write the following:
```
self.assertTrue(np.allclose(pred, expected_pred))
```

StrikerRUS:
@guolinke I think that **FAQ Question 9** should be updated.

pseudotensor:
@StrikerRUS Anything else?  Or is this good to merge?  Would love to get this.

PR #1737, opened 2018-10-09
[docs] Fixed some typos in Python API Docs

zaffnet:
Just fixed some typos in Python API docs. 

Contributed as a part of Hacktoberfest 2018

### DIFF
--- a/python-package/lightgbm/sklearn.py
+++ b/python-package/lightgbm/sklearn.py
@@ -316,7 +316,7 @@ class LGBMModel(_LGBMModelBase):
         group : array-like or None, optional (default=None)
             Group data of training data.
         eval_set : list or None, optional (default=None)
-            A list of (X, y) tuple pairs to use as a validation sets.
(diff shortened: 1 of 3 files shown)


python-package/lightgbm/sklearn.py
```diff
@@ -316,7 +316,7 @@ def fit(self, X, y,
         group : array-like or None, optional (default=None)
             Group data of training data.
         eval_set : list or None, optional (default=None)
-            A list of (X, y) tuple pairs to use as a validation sets.
+            A list of (X, y) tuple pairs to use as a validation set.
```
StrikerRUS: LightGBM supports multiple validation sets, so please leave `sets`.

zaffnet: Fixed it.

StrikerRUS: Thanks again! Will merge right after passing tests.

PR #1738, opened 2018-10-09
[docs] fixed some typos and grammatical errors

pushkyn:


docs/Python-Intro.rst
```diff
@@ -112,11 +112,11 @@ or
 
 And you can use ``Dataset.set_init_score()`` to set initial score, and ``Dataset.set_group()`` to set group/query data for ranking tasks.
 
-**Memory efficent usage:**
+**Memory efficient usage:**
 
 The ``Dataset`` object in LightGBM is very memory-efficient, due to it only need to save discrete bins.
 However, Numpy/Array/Pandas object is memory cost.
-If you concern about your memory consumption, you can save memory according to following:
+If your concern about your memory consumption, you can save memory according to the following:
```
StrikerRUS: I think that `concern` is a verb here, so `you` is OK. Or you can change `concern` to `worry`.

pushkyn: You are right, I updated it

pushkyn:
> Hi @pushkyn !
> 
> Thanks for you contribution!
> 
> Unfortunately, modifying parameters description (`Parameters.rst`) is not so trivial. You should make your changes into the `config.h` [file](https://github.com/Microsoft/LightGBM/blob/master/include/LightGBM/config.h) and then run `python ./helper/parameter_generator.py` [script](https://github.com/Microsoft/LightGBM/blob/master/helper/parameter_generator.py). The python script will update the docs. That is how it works.

I'm sorry, I was not attentive enough when I read the docs for contribute. Now I updated `config.h` file. Thanks.

StrikerRUS:
@pushkyn No problem! Thanks a lot!

PR #1688, opened 2018-09-20
[tests] fixed codestyle, removed unused code and added several new checks

StrikerRUS:

PR #1739, opened 2018-10-10
fix ranking tasks consistency

guolinke:
fix problems in #1688

### DIFF
--- a/src/boosting/gbdt_model_text.cpp
+++ b/src/boosting/gbdt_model_text.cpp
@@ -306,7 +306,7 @@ std::string GBDT::SaveModelToString(int start_iteration, int num_iteration) cons
   // sort the importance
   std::stable_sort(pairs.begin(), pairs.end(),
                    [](const std::pair<size_t, std::string>& lhs,
-                   const std::pair<size_t, std::string>& rhs) {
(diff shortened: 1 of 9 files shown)


src/boosting/gbdt_model_text.cpp
```diff
@@ -304,7 +304,7 @@ std::string GBDT::SaveModelToString(int start_iteration, int num_iteration) cons
     }
   }
   // sort the importance
-  std::sort(pairs.begin(), pairs.end(),
+  std::stable_sort(pairs.begin(), pairs.end(),
             [](const std::pair<size_t, std::string>& lhs,
```
StrikerRUS: Indent

src/io/dataset.cpp
```diff
@@ -162,7 +162,7 @@ std::vector<std::vector<int>> FastFeatureBundling(std::vector<std::unique_ptr<Bi
     sorted_idx.emplace_back(i);
   }
   // sort by non zero cnt, bigger first
-  std::sort(sorted_idx.begin(), sorted_idx.end(),
+  std::stable_sort(sorted_idx.begin(), sorted_idx.end(),
             [&feature_non_zero_cnt](int a, int b) {
```
StrikerRUS: Indent

PR #1754, opened 2018-10-12
[ci] downgrade Xcode version on Azure

StrikerRUS:
Refer to https://github.com/Microsoft/LightGBM/pull/1740#issuecomment-429195138.

Use the lowest possible version of Xcode on Azure.

To be honest, I'm not sure whether it's good decision or not... macOS users tend to use the latest version of the OS. So maybe it's better to focus on the latest AppleClang version?

### DIFF
--- a/.ci/setup.sh
+++ b/.ci/setup.sh
@@ -4,6 +4,9 @@ if [[ $OS_NAME == "macos" ]]; then
     if  [[ $COMPILER == "clang" ]]; then
         brew install libomp
         brew reinstall cmake  # CMake >=3.12 is needed to find OpenMP at macOS
+        if [[ $AZURE == "true" ]]; then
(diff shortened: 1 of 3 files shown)


StrikerRUS:
@guolinke What's your opinion on this?

guolinke:
@StrikerRUS I think as long as the higher version xcode can use the binaries, and the runtime are the same, it is okay.

StrikerRUS:
Do you have a quick access to macOS with installed, let say, Xcode 10 to check the binaries from this branch?

guolinke:
@StrikerRUS 
can we test this in CI ? 
That is, building it with low version xcode, but run it with high version.

StrikerRUS:
@guolinke OK, I'll try.

PR #1758, opened 2018-10-15
[docs] corrected misleading note about best_iteration

StrikerRUS:
I think model (`Booster`) has `best_iteration` field even without early stopping occurrence.

https://github.com/Microsoft/LightGBM/blob/09919669f4d889d76929e13890c9fa78cee9178a/python-package/lightgbm/basic.py#L1472
https://github.com/Microsoft/LightGBM/blob/09919669f4d889d76929e13890c9fa78cee9178a/python-package/lightgbm/engine.py#L201
https://github.com/Microsoft/LightGBM/blob/09919669f4d889d76929e13890c9fa78cee9178a/python-package/lightgbm/engine.py#L230


guolinke:
yeah.
But we just remove them? or add clearer descriptions ?

StrikerRUS:
@guolinke You're right! I've updated the description instead of removing.

docs/Python-Intro.rst
```diff
@@ -195,7 +195,7 @@ Early stopping requires at least one set in ``valid_sets``. If there is more tha
 The model will train until the validation score stops improving.
 Validation score needs to improve at least every ``early_stopping_rounds`` to continue training.
 
-If early stopping occurs, the model will have an additional field: ``bst.best_iteration``.
+If early stopping occurs, the iteration index will be saved in the ``bst.best_iteration`` field.
```
guolinke: It seems, as long as set the early_stopping_rounds, the best_iteration will be set whatever early stopping occurs or not.

PR #1760, opened 2018-10-15
[DO NOT MRG] Xcode test

StrikerRUS:
@guolinke Seems that library compiled under Xcode 8.3.1 runs OK with Xcode 10: see 
https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=662&view=logs
MacOS regular.

What do you think? Should we test more versions?


guolinke:
@StrikerRUS I think it is okay, let us use Xcode 8.3.1.

PR #1766, opened 2018-10-20
[python] save pandas_categorical to model string and JSON

StrikerRUS:
Suppress #1764.

Despite the fact that the issue, described in #1764, was caused by the usage of the undocumented feature, I think that saving `pandas_categorical` field to the model not only in `save_model`, but in `model_to_string` and `dump_model`, is good idea.


rostyboost:
Thanks for the quick turn-around

PR #1793, opened 2018-10-28
Update network.cpp

SreejuZzz:
Just made some lines more structural


guolinke:
sorry, why change the cpp style.

PR #1799, opened 2018-10-29
[docs] updated MS MPI download link

StrikerRUS:
MS MPI v10 has been released and download link has been changed.

Also, filename has become lowcase.


StrikerRUS:
Previous link was dead (404) for some time which caused Travis `docs` test fail: https://github.com/Microsoft/LightGBM/pull/1798#issuecomment-433874453, but now it seems to be alive again.

PR #1806, opened 2018-10-30
[ci] minor cleanup in CI installation stage

StrikerRUS:


guolinke:
@StrikerRUS I think we should check the effort introduced by clang dependency (gcc 4.9 etc) first, before merge this PR.

PR #1780, opened 2018-10-24
[python] add option to disable openmp in python package build

guolinke:

PR #1811, opened 2018-10-31
[docs] documented version without OpenMP support

StrikerRUS:
Refer to #1780.

### DIFF
--- a/docs/Installation-Guide.rst
+++ b/docs/Installation-Guide.rst
@@ -15,6 +15,8 @@ For the build of Python-package and R-package, please refer to `Python-package`_
 
 -  `Docker <#docker>`__
 
+- `Threadless Version (not Recommended) <#build-threadless-version-not-recommended>`__
+
 -  `MPI Version <#build-mpi-version>`__
 
 -  `GPU Version <#build-gpu-version>`__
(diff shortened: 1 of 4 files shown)


docs/Installation-Guide.rst
```diff
@@ -177,6 +179,151 @@ Docker
...
+- **Visual Studio**;
+
+- **CMake** and **VS Build Tools**;
+
+- **CMake** and **MinGW**.
+
+Visual Studio (or VS Build Tools)
+*********************************
+
+With GUI
```
StrikerRUS: Unfortunately, [I can't check](https://github.com/Microsoft/LightGBM/pull/1660#discussion_r216440029) whether it's enough just to uncheck OpenMP property. Can anyone test this, please?

StrikerRUS:
Should we relax CMake verion for OpenMP-less LightGBM?

docs/Installation-Guide.rst
guolinke: @StrikerRUS I just check it. It can work if merged https://github.com/Microsoft/LightGBM/tree/nomp-err

guolinke: create a new PR #1813 to fix that first

guolinke:
@StrikerRUS I think we can just let it be, as only mac os needs higher version to support openmp, and most macos users prefer newer versions.

docs/Installation-Guide.rst
StrikerRUS: @guolinke Great! Thanks!

StrikerRUS:
@guolinke Agree.

StrikerRUS:
Should OpenMP-less version be checked at CI side? Or, since it's not recommended version, we may not waste CI time?

Laurae2:
@StrikerRUS If `nthread=1` then OpenMP is ignored when running a LightGBM model.

StrikerRUS:
@Laurae2 Sorry, I was unclear. I mean **compilation** of the OpenMP-less version for the protection from cases like one was fixed in #1813.

PR #1813, opened 2018-11-01
fix bug with disable openmp

guolinke:


guolinke:
refer to https://github.com/Microsoft/LightGBM/pull/1780#issuecomment-433641541

include/LightGBM/utils/pipeline_reader.h
```diff
@@ -9,6 +9,7 @@
 #include <thread>
 #include <memory>
 #include <algorithm>
+#include <vector>
```
chivee: what does this for?

PR #1817, opened 2018-11-02
Fix index out-of-range exception generated by BaggingHelper on small datasets.

mjmckp:
Prior to this change, the line "score_t threshold = tmp_gradients[top_k - 1];" would generate an exception, since tmp_gradients would be empty when the cnt input value to the function is zero.


guolinke:
@mjmckp Thanks very much ! 
I think this function https://github.com/Microsoft/LightGBM/blob/master/src/boosting/gbdt.cpp#L192-L206 should be updated as well.

PR #1825, opened 2018-11-06
[ci] update Xcode on Travis

StrikerRUS:
Continuation of #1822.

Previous update for fix works, but it is unstable:
https://github.com/Microsoft/LightGBM/commit/b0a938aee175232bc1f20f0041346d07ff049827#commitcomment-31186324

This PR completely eliminate the need of `softwareupdate` command by the usage of latest `Xcode` version.

BTW, I think it's good point to build artifacts on Azure Pipelines with the lowest possible `Xcode` and sometimes to test building process on Travis with latest `Xcode` (instead of default `9.4`).

### DIFF
--- a/.ci/setup.sh
+++ b/.ci/setup.sh
@@ -9,11 +9,9 @@ if [[ $OS_NAME == "macos" ]]; then
         fi
     else
         if [[ $TRAVIS == "true" ]]; then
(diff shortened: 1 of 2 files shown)


chivee:
Yes, I think docker is more convenient for the test.

PR #1832, opened 2018-11-09
Removed `cd build` for Linux (caused CMake error)

nskeip:
Under Linux, you should not to change directory to build when building `cd build`. Otherwise, CMake error will occur:

```
CMake Error: The source directory "[...]/LightGBM/build" does not appear to contain CMakeLists.txt.
Specify --help for usage, or press the help button on the CMake GUI.
```


guolinke:
There is a parameter .. (double dots) in cmake command.

nskeip:
> There is a parameter .. (double dots) in cmake command.

Yes, you are correct. Thank you. I think I'll close the request.

PR #1849, opened 2018-11-16
[ci][docs] updated temp fix for warning written to stderr, fixed conda permissions and updated bench link

StrikerRUS:
Refer to https://github.com/scikit-learn/scikit-learn/issues/12226.

This temp fix should be removed after scikit-learn `0.20.1` release.

@guolinke Could you please fix dockerfile?

```
NotWritableError: The current user does not have write permissions to a required path.
path: /usr/local/miniconda/pkgs/cache/3e39a7aa.json
uid: 1001
gid: 1001

If you feel that permissions on this path are set incorrectly, you can manually
change them by executing

$ sudo chown 1001:1001 /usr/local/miniconda/pkgs/cache/3e39a7aa.json

In general, it's not advisable to use 'sudo conda'.
```

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -31,7 +31,7 @@ install:
   - ps: $env:LGB_VER = (Get-Content VERSION.txt).trim()
   - conda config --set always_yes yes --set changeps1 no
   - conda update -q conda
-  - conda create -q -n test-env python=%PYTHON_VERSION% joblib matplotlib nose numpy pandas pytest python-graphviz scikit-learn scipy
+  - conda create -q -n test-env python=%PYTHON_VERSION% cloudpickle joblib matplotlib nose numpy pandas pytest python-graphviz scikit-learn scipy
   - activate test-env
 
 build_script:
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -31,7 +31,7 @@ install:
   - ps: $env:LGB_VER = (Get-Content VERSION.txt).trim()
   - conda config --set always_yes yes --set changeps1 no
   - conda update -q conda
-  - conda create -q -n test-env python=%PYTHON_VERSION% cloudpickle joblib matplotlib nose numpy pandas pytest python-graphviz scikit-learn scipy
+  - conda create -q -n test-env python=%PYTHON_VERSION% cloudpickle=0.5.6 joblib matplotlib nose numpy pandas pytest python-graphviz scikit-learn scipy
   - activate test-env
 
 build_script:
--- a/.ci/setup.sh
+++ b/.ci/setup.sh
@@ -56,6 +56,6 @@ fi
 
 if [[ $TRAVIS == "true" ]]; then
     sh conda.sh -b -p $HOME_DIRECTORY/miniconda
-    conda config --set always_yes yes --set changeps1 no
-    conda update -q conda
 fi
+conda config --set always_yes yes --set changeps1 no
+conda update -q conda
--- a/.vsts-ci.yml
+++ b/.vsts-ci.yml
@@ -4,7 +4,7 @@ variables:
 resources:
   containers:
   - container: ubuntu1404
-    image: lightgbm/vsts-agent:ubuntu-14.04
+    image: lightgbm/vsts-agent:ubuntu-14.04-dev
 jobs:
 ###########################################
 - job: Linux
@@ -32,10 +32,8 @@ jobs:
         METHOD: source
         PYTHON_VERSION: 2.7
   steps:
-  - task: CondaEnvironment@1
-    inputs:
-      updateConda: true
   - script: |
+      echo "##vso[task.setvariable variable=PATH]$CONDA/bin:$PATH"
       echo "##vso[task.setvariable variable=HOME_DIRECTORY]$AGENT_HOMEDIRECTORY"
       echo "##vso[task.setvariable variable=BUILD_DIRECTORY]$BUILD_SOURCESDIRECTORY"
       echo "##vso[task.setvariable variable=OS_NAME]linux"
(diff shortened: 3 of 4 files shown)


guolinke:
@StrikerRUS sure, it is building the new docker.
BTW, I am in the vacation. Therefore, I cannot reply to the issues in time. 
If there is something urgent, you can email ( guolin.ke@microsoft.com ) it to me. 
Thanks!

StrikerRUS:
@guolinke Thanks a lot!
OK, got it. Have a nice vacation!

StrikerRUS:
Hmmm, didn't help...

guolinke:
@StrikerRUS should I add this to docker file ?
```
sudo chown 1001:1001 /usr/local/miniconda/pkgs/cache/3e39a7aa.json
```

StrikerRUS:
@guolinke This won't work because of different name of temp files from run to run.

I've tried different fixes: `chmod, chown, unset SUDO_UID`. Only `sudo chmod -R 777 /usr/local/miniconda` works but gives +4 min to build time.

It turned out that this is very widespread problem:
https://github.com/conda/conda/issues/7267
https://github.com/conda/conda/issues/6953
https://github.com/conda/conda/issues/6576

StrikerRUS:
I suppose that the problem in that `CondaEnvironment@1` task transforms into the following commands
```
Prepending PATH environment variable with directory: /usr/local/miniconda/bin
[command]/usr/bin/sudo /usr/local/miniconda/bin/conda update --name base conda --yes
```
executed with `sudo`.

StrikerRUS:
@brcrista can you please help? 
A few days ago this task
https://github.com/Microsoft/LightGBM/blob/a694712b7fb86cd532eea2c1781b58d4ba58436a/.vsts-ci.yml#L35-L37
started to use `sudo`, which produces a well-known conda issue

> It turned out that this is very widespread problem:
conda/conda#7267
conda/conda#6953
conda/conda#6576

with [`NotWriteableError`](https://github.com/Microsoft/LightGBM/pull/1849#issue-231684672) during the next conda usage.

guolinke:
@StrikerRUS maybe we can update conda by ourself?
I remember there an env-var called "CONDA", which save the conda path.

StrikerRUS:
@guolinke Yeah! Good point! Let's try this.

StrikerRUS:
@guolinke Unfortunately, in case of manual conda update another error occurs:
```
 ERROR conda.core.link:_execute(502): An error occurred while uninstalling package 'defaults::cryptography-2.3.1-py37hc365091_0'.
PermissionError(13, 'Permission denied')
Attempting to roll back.
Rolling back transaction: ...working... done
PermissionError(13, 'Permission denied')
```

Log: https://lightgbm-ci.visualstudio.com/lightgbm-ci/_build/results?buildId=967&view=logs.

Links: https://github.com/conda/conda/issues/7404, https://github.com/conda/conda/issues/7824, https://stackoverflow.com/questions/49181799/conda-update-conda-permission-error.

Popular solutions are
- `chmod/chown` which leads to +5 min to build time,
- using `sudo` which leads to the previous error,
- installation under the same user.

StrikerRUS:
@guolinke I don't know, but can we solve this issue by changing conda installation folder?
https://stackoverflow.com/a/27364203

guolinke:
@StrikerRUS 
I trigger a new docker build in branch `docker`, and change the conda location.
you can use tagname ubuntu-14.04-dev to try it.
You also can update the `docker` branch to trigger the new docker build.

BTW, you can give me your dockerhub account, and I can add you to LightGBM docker org. 

Thanks

StrikerRUS:
@guolinke Thank you very much!

conda update and installation finally work fine with `ubuntu-14.04-dev`. Please merge `docker` branch into `master` and then I'll remove `-dev` suffix from the docker image here.

Let me register docker account and come back to you.

guolinke:
@StrikerRUS great! I will create a PR to merge it.

brcrista:
Thanks @StrikerRUS , I wasn't aware of this problem.  I haven't seen it before.  Do you have logs you can share?  (running with system.debug=true would be most helpful)

I do see the line that says
`In general, it's not advisable to use 'sudo conda'.`
and found this article: https://conda.io/docs/user-guide/troubleshooting.html#permission-denied-errors-after-using-sudo-conda-command

Right now we have to sudo based on how Conda is installed -- see Microsoft/azure-pipelines-image-generation#259

StrikerRUS:
@brcrista Thanks a lot for coming here!

The problem is that we use Docker in which Miniconda is installed in `/usr/local/miniconda`:
https://github.com/Microsoft/LightGBM/blob/a694712b7fb86cd532eea2c1781b58d4ba58436a/.ci/dockers/ubuntu-14.04/Dockerfile#L44-L52
Then `CondaEnvironment@1` task runs conda with `sudo` and after that any attempts to run conda (to be more precise we run `conda create`)
https://github.com/Microsoft/LightGBM/blob/a694712b7fb86cd532eea2c1781b58d4ba58436a/.ci/test.sh#L11
fail with the error mentioned above.

Here are logs with this error: https://lightgbm-ci.visualstudio.com/lightgbm-ci/_build/results?buildId=945&view=logs

FYI, we've managed to solve this issue by installing miniconda in `/opt/conda`
https://github.com/Microsoft/LightGBM/blob/cb9043d4ee1ecaf4c43c54930cb3fec6f8c3cfcf/.ci/dockers/ubuntu-14.04/Dockerfile#L44-L54
which allows to avoid using `sudo` and updating conda by ourselves (without `CondaEnvironment@1` task):
https://github.com/Microsoft/LightGBM/blob/2571384e4674a23da580f80bc78cbd4e15b7e8e0/.ci/setup.sh#L60-L61

StrikerRUS:
@guolinke Here is my docker hub link:
https://hub.docker.com/u/strikerrus/

Thanks!

brcrista:
I see -- I thought you were on a hosted agent.

For your setup, I would recommend just scripting calls to conda instead of using the task.  This usually works better for YAML definitions (the task was made for the old designer workflows).

One difference from using the task: if you're running scripts in multiple steps that operate in the context of the Conda environment, you will need to activate the environment at the beginning of every steps.

StrikerRUS:
@brcrista Got it! Thank you!

PR #1862, opened 2018-11-21
When loading a binary file, take feature penalty from config if given there.

remcob-gr:
This enables users to override the feature penalty when loading a dataset from a binary file.
For instance, it enables disabling/enabling features via:
```Python
dataset = lgbm.Dataset('data.bin',params={'feature_contri': [0,1,1,0]})
....
```
The above is especially useful when the data set is large, so the cost of recreating the data set from the source data is large.

### DIFF
--- a/src/io/dataset_loader.cpp
+++ b/src/io/dataset_loader.cpp
@@ -381,10 +381,19 @@ Dataset* DatasetLoader::LoadFromBinFile(const char* data_filename, const char* b
     dataset->monotone_types_.clear();
   }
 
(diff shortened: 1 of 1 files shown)


StrikerRUS:
Hi @remcob-gr !
Thanks for your contribution!

Sorry for the inconvenience but you should sign the CLA and rebase the branch against the latest `master` to pass all checks.

remcob-gr:
Closing this PR in favour of https://github.com/Microsoft/LightGBM/pull/1881 , which I've rebased on top of master.

PR #1870, opened 2018-11-24
[python] added get_data() method to Dataset class

### LINKED ISSUE #1690: Accessing input training data from `Dataset` object?
## Environment info

CPU: Yes (not GPU if that's what this means?)

C++/Python/R version: Python

In [advanced example](https://github.com/Microsoft/LightGBM/blob/master/examples/python-guide/advanced_example.py) we got the labels from the training data but I want to get the features themselves. How would I do so? 

I need to implement a custom loss function and I need acces to the `X's` for a custom regression task. Specifically, I want to do a regression, and I'd like to include a regularizer. The regularizer would look like `np.inner((label - X[-1], (predicted - X[-1])`) having the same sign

### LINKED ISSUE #1763: Dataset has no data when passed to fpreproc in Python cv function
## Environment info

CPU/GPU model: 

C++/Python/R version: 
3.7.0 (default, Oct  9 2018, 10:31:47) 
[GCC 7.3.0]

```
lgb.__version__, pd.__version__, np.__version__
```
`('2.2.1', '0.23.4', '1.15.2')`

## Error message

`dtrain.data` should be a pandas DataFrame and should not be `None`. When I get the below error, I can see that the dtrain.data _is_ being set to `None`:

`AttributeError: 'NoneType' object has no attribute 'assign'`

## Reproducible examples

```
import numpy as np
import pandas as pd
import lightgbm as lgb


data = pd.DataFrame({'x1': np.random.rand(1000),
                     'x2': 5 + np.random.rand(1000),
                     'target': np.random.randint(0, 2, 1000)})

data_train = lgb.Dataset(data.drop(['target'], axis=1), data.target, free_raw_data=False)


def fpreproc_rebalance(dtrain, dtest, params,):
    data = dtrain.data.assign(target=data.label)
    neg = data.loc[lambda x: x.target == 0]

    data = (
        pd.concat([data] + [neg] * negative_upsample_rate)
    )

    fdtrain = lgb.Dataset(data.drop('target', axis=1), data.target, free_raw_data=False)
    return fdtrain, dtest, params


params = {
    'objective': 'xentlambda',
    'feature_fraction': 0.8,
    'learning_rate': 0.4,
}

results = lgb.cv(
    params,
    data_train,
    nfold=5,
    fpreproc=fpreproc_rebalance,
)
```

## Steps to reproduce

1. Run above code.
2. Weep into keyboard ;)

StrikerRUS:
Fixed #1690.
Fixed #1763.


python-package/lightgbm/basic.py
```diff
@@ -1375,6 +1378,22 @@ def get_init_score(self):
             self.init_score = self.get_field('init_score')
         return self.init_score
 
+    def get_data(self):
+        """Get the raw data of the Dataset.
+
+        Returns
+        -------
+        data : string, numpy array, pandas DataFrame, scipy.sparse, list of numpy arrays or None
```
StrikerRUS: What about `string` and `list of numpy arrays`? Hot to "slice" them?..

StrikerRUS:
Simplified example from #1763

```
import numpy as np
import pandas as pd
import lightgbm as lgb

full_data = pd.DataFrame({'x1': np.random.rand(100),
                          'x2': 5 + np.random.rand(100),
                          'target': np.random.randint(0, 2, 100)})

data_train = lgb.Dataset(full_data.drop(['target'], axis=1), full_data.target, free_raw_data=False)

def fpreproc_rebalance(dtrain, dtest, params,):
    train_data = dtrain.get_data()
    test_data = dtest.get_data()
    train_data.loc[train_data['x1'] > 0.5, ['x1']] = 0.5
    test_data.loc[test_data['x1'] > 0.5, ['x1']] = 0.5
    fdtrain = lgb.Dataset(train_data, dtrain.label, free_raw_data=False)
    fdtest = lgb.Dataset(test_data, dtest.label, free_raw_data=False)
    return fdtrain, fdtest, params

results = lgb.cv({}, data_train, fpreproc=fpreproc_rebalance)
```

fails with

```
LightGBMError: Cannot add validation data, since it has different bin mappers with training data
```

python-package/lightgbm/basic.py
guolinke: I think we cannot slide them, maybe we should throw a warning or error here.

python-package/lightgbm/basic.py
```diff
@@ -1040,8 +1042,9 @@ def subset(self, used_indices, params=None):
         """
         if params is None:
             params = self.params
-        ret = Dataset(None, reference=self, feature_name=self.feature_name,
-                      categorical_feature=self.categorical_feature, params=params)
+        ret = Dataset(self.data, reference=self, feature_name=self.feature_name,
+                      categorical_feature=self.categorical_feature, params=params,
+                      free_raw_data=self.free_raw_data)
```
guolinke: maybe a better logic still pass `None` for `data` here?
Then in `construct`, use reference.data to `get_data` ?
Then, the logic of construct are the same as before.

StrikerRUS:
Now [this snippet](https://github.com/Microsoft/LightGBM/pull/1870#issuecomment-441334368) should be modified with `construct()` like the following:
```
...
def fpreproc_rebalance(dtrain, dtest, params,):
    train_data = dtrain.construct().get_data()
    test_data = dtest.construct().get_data()
    ...
```

python-package/lightgbm/basic.py
```diff
@@ -974,6 +975,8 @@ def construct(self):
                         ctypes.c_int(used_indices.shape[0]),
                         c_str(params_str),
                         ctypes.byref(self.handle)))
+                    self.data = self.reference.data
+                    self.get_data()
```
guolinke: maybe we don't need the `get_data` here ?

guolinke: okay, I see. As only subset will call the slicing, I think the overhead cost is acceptable. 
Change the definition of `data` may cause many dependency problems.

StrikerRUS: Yeah, it's expected that only subsetting branch of `construct` calls `get_data()`. Also, `get_data()` itself checks `used_indices`:
https://github.com/Microsoft/LightGBM/blob/1ff6c6d6bca05de2c13490264ba3b6bf7f993d4a/python-package/lightgbm/basic.py#L1392

StrikerRUS:
@guolinke please help with [this](https://github.com/Microsoft/LightGBM/pull/1870#issuecomment-441334368).

StrikerRUS:
@guolinke Unfortunately, it still fails (`data` branch with dll file from latest `master`):
```
import numpy as np
import pandas as pd
import lightgbm as lgb

full_data = pd.DataFrame({'x1': np.random.rand(100),
                          'x2': 5 + np.random.rand(100),
                          'target': np.random.randint(0, 2, 100)})

data_train = lgb.Dataset(full_data.drop(['target'], axis=1), full_data.target, free_raw_data=False)

def fpreproc_rebalance(dtrain, dtest, params,):
    train_data = dtrain.construct().get_data()
    test_data = dtest.construct().get_data()
    train_data.loc[train_data['x1'] > 0.5, ['x1']] = 0.5
    test_data.loc[test_data['x1'] > 0.5, ['x1']] = 0.5
    fdtrain = lgb.Dataset(train_data, dtrain.label, free_raw_data=False)
    fdtest = lgb.Dataset(test_data, dtest.label, free_raw_data=False)
    return fdtrain, fdtest, params

results = lgb.cv({}, data_train, fpreproc=fpreproc_rebalance)
```

```
LightGBMError: Cannot add validation data, since it has different bin mappers with training data
```

guolinke:
@StrikerRUS you need to set the reference for valid data:
```
    fdtrain = lgb.Dataset(train_data, dtrain.label, free_raw_data=False)
    fdtest = lgb.Dataset(test_data, dtest.label, free_raw_data=False, reference=fdtrain )
```

StrikerRUS:
@guolinke Test has been added. Please check.

Issue #768, opened 2017-08-02
[GPU] further improving GPU performance


guolinke:
Refer to a benchmark here: https://blogs.technet.microsoft.com/machinelearning/2017/07/25/lessons-learned-benchmarking-fast-machine-learning-algorithms/  and https://github.com/Microsoft/LightGBM/issues/620#issuecomment-309955704 .

It seems the LightGBM GPU still can be further improved. Current GPU implementation has a overhead, which is the additional memory copy cost between GPU and CPU. 
As a result, when #data is small, using GPU may slow than CPU. 

@huanzhang12 any updates for this ?

guolinke:
@huanzhang12 another question: Can moving the GPU code to CUDA improve the speed ?

huanzhang12:
Yes, LightGBM GPU can still be improved in many ways. Currently the GPU implementation only uses like 30%-50% of full GPU potential.

The major reason the GPU is slow for small data is that, we need to transfer the histograms from GPU to CPU to find the best split after the feature histograms are built. This is not ideal. The overhead of data transfer is significant on dataset with a lot of features or small data; also, it requires the CPU to do too much job, and the CPU can become a bottleneck. For better efficiency, we should find the best split on GPU, preferably in GPU local memory. But it needs some work because we need to re-implement the histogram pool and some other functions (like historgram fixup, split finding for numerical/categorical features) on GPU, which are non-trivial.

After these are implemented, I expect about 2X speedup on large datasets, and significant speedup on smaller datasets (since there are no data transfer overhead). GPU training could become a standard for GBDTs, like what we are doing for deep learning.

Also, we need to work on enabling multi-GPU training. This seems not very hard, as it can be viewed as a special case of distributed learning.

I have rough ideas on how to implement these things, and I really want to continue working on improving the GPU algorithm. But unfortunately right now I am quite busy with my internship, and can only work on this project during my limited spare time :'( I don't think I can finish it any time soon.

Let me know if you have any better ideas on this issue.

RAMitchell:
Multi-gpu and some of these other features can be harder in OpenCL due to library availability. We have nccl for p2p multi-gpu allreduce which helps a lot. I actually started off in OpenCL and switched to Cuda due to frustrations with boost compute.

huanzhang12:
[...]

PR #1873, opened 2018-11-24
[docs] Added the links to the libraries used.

KOLANICH:


StrikerRUS:
@KOLANICH Can you please sign the CLA?

Also, please see this: https://github.com/Microsoft/LightGBM/blob/master/docs/GPU-Targets.rst#gpu-targets-table (a note for the table) and fix markdown to make Travis happy.

KOLANICH:
>On Windows, OpenCL is included in AMD graphics driver.

Yes, but

1 In my case OpenCL.dll from driver lacked some exported functions, so there were problems with linking and working. After I have replaced OpenCL.dll in %SYSTEM32% with the one from APP SDK, everything started working.
2 For compilation https://github.com/KhronosGroup/OpenCL-Headers can be used, though I still prefer APP SDK, because it is shipped with a compiler and examples.

> AMD APP SDK is deprecated.

OpenCL itself is dying. It's inconvenient to use (manual setup, though there are frameworks doing this for us), it's deprecated by Apple, it's not supported by most of computation networks frameworks, new implementations of CUDA for non-NVIDIA GPUs target Vulkan rather than OpenCL.

StrikerRUS:
@KOLANICH Thanks a lot for sharing your valuable experience and information! There is a separate open [issue](#768) for GPU-version improvements.

StrikerRUS:
ping @huanzhang12

KOLANICH:
>It is likely to be inside the mingw installation folder (I don't have Windows machine at this time so I cannot verify).

They can be unpacked anywhere, but one will have to provide the paths to them into the corresponding CMake variables. Everything is as usual.

huanzhang12:
@KOLANICH sounds good. Could you just add a sentence on that? Just tell the users to set the CMAKE path variables according to the unpacked directory.

KOLANICH:
This is usually applied not only to Boost, but to all third-party libs having a CMake module to be discovered.

KOLANICH:
>is there any problem with it?

* The page doesn't work without JS enabled. Knowing M$ I don't want to enable JS on it.
* Why should I? MIT license is permissive enough, so no objection to distribute my contribution under it. 
* I strongly dislike dealing with any legal documents and prefer not to sign any.

So

>Can you please sign the CLA?

I'm sorry, but no. I don't care about copyright as much as needed to keep me out of any kegal trouble, and wish it to be abolished, so feel free to close the PR and add the links and everything else you feel needed yourself from your account and your name, I don't object, act as if my contribution was under [CC0](https://creativecommons.org/publicdomain/zero/1.0/). If my contribution is needed, it would be added anyway without any CLAs. If it is not needed, it is a bad idea to merge it.

StrikerRUS:
@KOLANICH Got it! Thanks for the explanation.
Unsigned CLA doesn't allow maintainers to merge a PR:
image

ping @guolinke @chivee for the situation around the CLA.

StrikerRUS:
ping @guolinke

guolinke:
Thanks so much and understand.
However, we cannot merge the PR without CLA, as I cannot represent Microsoft to make this decision... 😢

Laurae2:
@guolinke We can recreate the PR on our side as mentioned by @KOLANICH.

StrikerRUS:
Doesn't work, still waiting the CLA. I've reopened this PR from my account: #1962.

PR #1874, opened 2018-11-25
[R] add reset_data in lgb.cv

### LINKED ISSUE #1854: Please add reset_data to lgb.cv
The lgb.train has the argument `reset_data = T` to release the data from the model as in https://github.com/Microsoft/LightGBM/issues/879. In the case of multiple k-fold cv the RAM is eaten so fast, so it would much better if lgb.cv also has the argument `reset_data`. 

Thank you very much.

guolinke:
fix #1854

### DIFF
--- a/R-package/R/lgb.cv.R
+++ b/R-package/R/lgb.cv.R
@@ -41,7 +41,7 @@ CVBooster <- R6::R6Class(
 #'        type str represents feature names
 #' @param callbacks list of callback functions
 #'        List of callback functions that are applied at each iteration.
-#' @param free_boosters Boolean, setting it to TRUE (the default value) will free the Booster objects in cv-folds
+#' @param reset_data Boolean, setting it to TRUE (not the default value) will transform the booster model into a predictor model which frees up memory and the original datasets
 #' @param ... other parameters, see Parameters.rst for more information. A few key parameters:
 #'            \itemize{
 #'                \item{boosting}{Boosting type. \code{"gbdt"} or \code{"dart"}}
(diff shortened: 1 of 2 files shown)


R-package/R/lgb.cv.R
```diff
@@ -303,7 +305,9 @@ lgb.cv <- function(params = list(),
     if (env$met_early_stop) break
 
   }
-
+  if (free_boosters) {
+    cv_booster$boosters = NULL
```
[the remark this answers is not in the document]
jameslamb: @guolinke would anyone ever NOT want to free them? I'd rather just make this new behavior and leave the function signature the same. Adding flags to the code makes it harder to maintain, in my experience.

guolinke:
@jameslamb @Laurae2 
please review this PR. Thanks !

R-package/R/lgb.train.R
```diff
@@ -12,6 +12,8 @@
 #' @param categorical_feature list of str or int
 #'        type int represents index,
 #'        type str represents feature names
+#' @param callbacks list of callback functions
+#'        List of callback functions that are applied at each iteration.
```
jameslamb: how is this change related?

R-package/R/lgb.cv.R
```diff
@@ -303,7 +305,19 @@ lgb.cv <- function(params = list(),
     if (env$met_early_stop) break
 
   }
-
+  if (reset_data) {
+    lapply(cv_booster$boosters, function(fd) {
+      # Store temporarily model data elsewhere
```
jameslamb: Makes sense to me

R-package/R/lgb.train.R
guolinke: I find the doc of this parameter is misning, compared with lgb.cv.

PR #1876, opened 2018-11-25
support to override some parameters in Dataset

### LINKED ISSUE #1853: monotonic constraints not working?
I have rerun my [checks on monotonic constraints](https://github.com/Microsoft/LightGBM/pull/1314#issuecomment-382113794) and could not reproduce them under the current lightgbm release 2.2.3 as well as under 2.1.x. The constraints are not satisfied.

## Environment info

Operating System: Windows

CPU/GPU model: CPU

C++/Python/R version: R 3.5.0

## Error message

None.

## Steps to reproduce

```
library(tidyverse)
library(caret)
library(lightgbm)
library(xgboost)

# Generate function with one input x
set.seed(20)
x <- seq(-0.2, 1, by = 0.001)
y <- x^2 + rnorm(length(x), sd = 0.1)

dat <- data.frame(x, y)

base_plot <- ggplot(dat, aes(x = x, y = y)) + 
  geom_point()
base_plot

# train/test split
ind <- caret::createDataPartition(y, p = 0.8, list = FALSE) %>% c
train <- dat[ind, ] %>% data.matrix
test <- dat[-ind, ] %>% data.matrix

# LGB
lgb_data <- lgb.Dataset(train[, "x", drop = FALSE], label = train[, "y"])

# Some reasonable parameters (would need to be optimized)
params <- list(learning_rate = 0.03,
               num_leaves = 31,
               min_data_in_leaf = 10,
               bagging_fraction = 0.7,
               bagging_freq = 4,
               nthread = 4,
               max_bin = 255,
               metric = "rmse",
               objective = "regression")

# Use cross-validation to figure out an appropriate number of iterations
fit_cv <- lgb.cv(params = params, 
                 data = lgb_data, 
                 nfold = 5, 
                 nrounds = 1000, 
                 early_stopping_rounds = 20,
                 verbose = 0)
fit_cv$best_iter

fit <- lgb.train(params = params, 
                 data = lgb_data, 
                 nrounds = fit_cv$best_iter,
                 verbose = 0)

# Visualize predictions on test data
test_pred <- data.frame(test, pred = predict(fit, test)) 
  
base_plot +
  geom_line(data = test_pred, aes(x = x, y = pred), 
            col = "blue", lwd = 2)

# Now with monotonicity constraints
fit_monotone <- lgb.train(params = c(params, monotone_constraints = '1'), 
                          data = lgb_data, 
                          nrounds = fit_cv$best_iter,
                          verbose = 1)

# Visualize predictions on test data
test_pred$pred_monotone <- predict(fit_monotone, test) 

base_plot +
  geom_line(data = test_pred, aes(x = x, y = pred_monotone), 
            col = "blue", lwd = 2)
```
image

guolinke:
fix #1853

### DIFF
--- a/R-package/R/lgb.Dataset.R
+++ b/R-package/R/lgb.Dataset.R
@@ -492,7 +492,7 @@ Dataset <- R6::R6Class(
     update_params = function(params) {
 
       # Parameter updating
-      if (lgb.is.null.handle(private$handle)) {
+      if (!lgb.is.null.handle(private$handle)) {
         lgb.call("LGBM_DatasetUpdateParam_R", ret = NULL, private$handle, lgb.params2str(params))
         return(invisible(self))
       }
--- a/R-package/R/lgb.Dataset.R
+++ b/R-package/R/lgb.Dataset.R
@@ -493,7 +493,7 @@ Dataset <- R6::R6Class(
 
       # Parameter updating
       if (lgb.is.null.handle(private$handle)) {
-        lgb.call("LGBM_DatasetUpdateParam_R", ret = NULL, private$handle, lgb.c_str(params))
+        lgb.call("LGBM_DatasetUpdateParam_R", ret = NULL, private$handle, lgb.c_str(lgb.params2str(params)))
         return(invisible(self))
       }
       private$params <- modifyList(private$params, params)
--- a/R-package/R/lgb.Dataset.R
+++ b/R-package/R/lgb.Dataset.R
@@ -492,6 +492,10 @@ Dataset <- R6::R6Class(
     update_params = function(params) {
 
       # Parameter updating
+      if (lgb.is.null.handle(private$handle)) {
+        lgb.call("LGBM_DatasetUpdateParam_R", ret = NULL, private$handle, lgb.c_str(params))
+        return(invisible(self))
+      }
       private$params <- modifyList(private$params, params)
       return(invisible(self))
 
--- a/R-package/R/lgb.Dataset.R
+++ b/R-package/R/lgb.Dataset.R
@@ -493,7 +493,7 @@ Dataset <- R6::R6Class(
 
       # Parameter updating
       if (lgb.is.null.handle(private$handle)) {
-        lgb.call("LGBM_DatasetUpdateParam_R", ret = NULL, private$handle, lgb.c_str(lgb.params2str(params)))
+        lgb.call("LGBM_DatasetUpdateParam_R", ret = NULL, private$handle, lgb.params2str(params))
         return(invisible(self))
       }
       private$params <- modifyList(private$params, params)
--- a/include/LightGBM/c_api.h
+++ b/include/LightGBM/c_api.h
@@ -309,6 +309,14 @@ LIGHTGBM_C_EXPORT int LGBM_DatasetGetField(DatasetHandle handle,
                                            const void** out_ptr,
                                            int* out_type);
 
+
+/*!
+* \brief Update parameters for a Dataset
+* \param handle a instance of data matrix
+* \param parameters parameters
+*/
+LIGHTGBM_C_EXPORT int LGBM_DatasetUpdateParam(DatasetHandle handle, const char* parameters);
+
 /*!
 * \brief get number of data.
 * \param handle the handle to the dataset
--- a/include/LightGBM/dataset.h
+++ b/include/LightGBM/dataset.h
@@ -575,6 +575,8 @@ public:
     return bufs;
   }
 
+  void ResetConfig(const char* parameters);
+
   /*! \brief Get Number of data */
   inline data_size_t num_data() const { return num_data_; }
 
@@ -615,6 +617,11 @@ private:
   std::vector<int8_t> monotone_types_;
   std::vector<double> feature_penalty_;
   bool is_finish_load_;
+  int max_bin_;
+  int bin_construct_sample_cnt_;
+  int min_data_in_bin_;
+  bool use_missing_;
+  bool zero_as_missing_;
 };
 
 }  // namespace LightGBM
--- a/include/LightGBM/lightgbm_R.h
+++ b/include/LightGBM/lightgbm_R.h
@@ -172,6 +172,16 @@ LIGHTGBM_C_EXPORT LGBM_SE LGBM_DatasetGetField_R(LGBM_SE handle,
   LGBM_SE call_state);
 
 /*!
+* \brief Update parameters for a Dataset
+* \param handle a instance of data matrix
+* \param parameters parameters
+* \return 0 when succeed, -1 when failure happens
+*/
+LIGHTGBM_C_EXPORT LGBM_SE LGBM_DatasetUpdateParam_R(LGBM_SE handle,
+  LGBM_SE params,
+  LGBM_SE call_state);
+
+/*!
 * \brief get number of data.
 * \param handle the handle to the dataset
 * \param out The address to hold number of data
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -1066,26 +1066,8 @@ class Dataset(object):
         return self
 
     def _update_params(self, params):
-        if self.handle is not None:
-            key_lists = [
-                "min_data", "min_data_in_leaf", "min_data_per_leaf", "min_child_samples",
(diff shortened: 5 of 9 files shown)


guolinke:
ping @Laurae2 
@Laurae2 if you are not available recently, we can merge the modification in python first.

mayer79:
@guolinke: Great. Below a proposal for R (modifying the "update_params" function in R-Package/R/lgb.Dataset.R [as I am not familiar how to commit to a pull request].

The current implementation for Python (and R) has two minor issues. I think we can live with them for the moment:

1. If we e.g. use "min_data = 10" instead of "min_data_in_leaf = 10", the warning will be raised unnecessarily.
2. If we e.g. start with "min_data = 1", then set "min_data = 10" we correctly get a warning. Then, if we reset the parameters back to the original "min_data = 1", we will get again a warning. 

    ```
    # Update parameters
    update_params = function(params) {
      
      # Check if new parameter list is different from old list wrt problematic parameters.
      if (!lgb.is.null.handle(private$handle)) {
        key_list <- c("min_data", "min_data_in_leaf", "min_data_per_leaf", "min_child_samples",
                      "bin_construct_sample_cnt", "subsample_for_bin",
                      "max_bin", 
                      "min_data_in_bin",
                      "enable_bundle", "is_enable_bundle", "bundle",
                      "max_conflict_rate",
                      "is_enable_sparse", "sparse", "is_sparse", "enable_sparse",
                      "sparse_threshold",
                      "use_missing",
                      "zero_as_missing",
                      "categorical_feature", "cat_feature", "categorical_column", "cat_column",
                      "monotone_constraints", "monotone_constraint", "mc")
        
        for (key in names(params)) {
          if (key %in% key_list && (!(key %in% names(private$params)) || private$params[[key]] != params[[key]])) {
            warning("Cannot update parameter ", sQuote(key), 
                    " after construct Dataset. Please re-declare a new Dataset object with new parameters.")
          }
        }
      } 
      
      # Overwrite parameters, even if inconsistent with Dataset construction.
      if (lgb.is.null.handle(private$handle)) {
        private$params <- params
      } else {
        private$params <- modifyList(private$params, params)
      } 
      
      return(invisible(self))
      
    },
```

**Comment by guolinke** [comment]
@mayer79 
the second issue will not happen, as the parameter will be reserved after using a new booster:
https://github.com/Microsoft/LightGBM/pull/1876/files#diff-732a5a5220860efcac575e9e956bbaeaL1076

**Comment by mayer79** [comment]
@guolinke: In Python, this is indeed implemented. In R, there is no reverse update in train.R yet.

**Comment by guolinke** [comment]
todo: the same functionality in R.

**Comment by Laurae2** [comment]
@guolinke If I understand correctly, we check the list of parameters before training so if another parameter in that list shows up, then we throw a warning? (would need to check if dataset is constructed beforehand or not)

By design in R should we enforce dataset parameters in `lgb.Dataset` only or should we allow them the first time in `lgb.train`?

**Comment by guolinke** [comment]
@Laurae2 sorry for late response. 
I think the warning condition is: user try to override existing parameters, after the handle of dataset is constructed. 
As the handle of dataset is lazily constructed, it is not enough to just enforce this before in lgb.Dataset or lgb.train.

**Comment by guolinke** [comment]
@Laurae2 
Actually, I think this should be done inside cpp. as there are alias and default values (which is not show in python/R).
I will move these logics into cpp side when I have time.

**Comment by guolinke** [comment]
@mayer79 @Laurae2 
Please check the new implementation. 
Now, the mc and feature_contrib could be updated

**Comment by guolinke** [comment]
ping @mayer79 @Laurae2

**Comment by mayer79** [comment]
I tested the commit
```
commit a3595d00a0ae2e0c7e56b24135d589a2301c4e61 (HEAD, origin/re-construct-dataset)
Author: Guolin Ke <guolin.ke@outlook.com>
Date:   Fri Dec 28 15:48:09 2018 +0800

    Update basic.py
```

in R with the following findings:

The first time when I run lightgbm in a session, I get a warning:
```
Warning message:
In charToRaw(as.character(x)) :
  argument should be a character vector of length 1
all but the first element will be ignored
```
This happens when `update_params(params)` is called for the second time. 

```
debug: data$update_params(params)
Browse[2]> 
[LightGBM] [Warning] Unknown parameter 0.03
debug: data$.__enclos_env__$private$set_predictor(predictor)
```

I have never seen this warning with older versions of the code, so it is probably due to the changes.


Then, if I change a problematic parameter (min_data_in_leaf from 10 to 20) and run lgb.train again, there is no warning at all. So it is not yet fully compatible with the R wrapper.


This is the reproducible code used:
```
library(tidyverse)
library(caret)
library(lightgbm)

# Generate function with one input x
set.seed(20)
x <- seq(-0.2, 1, by = 0.001)
y <- x^2 + rnorm(length(x), sd = 0.1)

dat <- data.frame(x, y)

# train/test split
ind <- caret::createDataPartition(y, p = 0.8, list = FALSE) %>% c
train <- dat[ind, ] %>% data.matrix
test <- dat[-ind, ] %>% data.matrix

# LGB
lgb_data <- lgb.Dataset(train[, "x", drop = FALSE], label = train[, "y"])

# Some reasonable parameters (would need to be optimized)
params <- list(learning_rate = 0.03,
               num_leaves = 31,
               min_data_in_leaf = 10,
               bagging_fraction = 0.7,
               bagging_freq = 4,
               nthread = 4,
               max_bin = 255,
               metric = "rmse",
               objective = "regression")

# Fit for first time gives warning
fit <- lgb.train(params = params, 
                 data = lgb_data, 
                 nrounds = 2)

# Now change tricky param
params$min_data_in_leaf <- 20

lgb.train(params = params, 
          data = lgb_data, 
          nrounds = 100)
```

**Comment by guolinke** [comment]
Thanks @mayer79, there is a bug in R. just fix it.

**Comment by mayer79** [comment]
I am afraid that there is still a small problem with that call:

```
data$update_params(params)
[LightGBM] [Warning] Unknown parameter 6c
Warning message:
In charToRaw(as.character(x)) :
  argument should be a character vector of length 1
all but the first element will be ignored
Browse[2]> lgb.params2str(params)
  [1] 6c 65 61 72 6e 69 6e 67 5f 72 61 74 65 3d 30 2e 30 33 20 6e 75 6d 5f 6c 65 61 76 65 73 3d 33 31 20 6d 69 6e 5f 64
 [39] 61 74 61 5f 69 6e 5f 6c 65 61 66 3d 31 30 20 62 61 67 67 69 6e 67 5f 66 72 61 63 74 69 6f 6e 3d 30 2e 37 20 62 61
 [77] 67 67 69 6e 67 5f 66 72 65 71 3d 34 20 6e 74 68 72 65 61 64 3d 34 20 6d 61 78 5f 62 69 6e 3d 32 35 35 20 6d 65 74
[115] 72 69 63 3d 72 6d 73 65 20 6f 62 6a 65 63 74 69 76 65 3d 72 65 67 72 65 73 73 69 6f 6e 20 76 65 72 62 6f 73 65 3d
[153] 31 00
```

guolinke:
@mayer79 yeah, you are right! thanks for point it out

mayer79:
Oops, now fatal error & crash ;)

guolinke:
Any logs?

mayer79:
I am afraid no - RGui crashes without log. So far I was unable to collect a dump file in R Windows.

guolinke:
@mayer79 
sorry, I find another bug.
Very sorry, I am very busy and cannot verify these codes carefully...

mayer79:
I did not spot that one - crap! I will test after work. No need to apologize - your work is incredibly useful.

mayer79:
Now it works: After changing "max_bin", the following message pops up

`[LightGBM] [Warning] Cannot change max_bin after constructed Dataset handle.`

If I change the parameter back to its original value, the warning vanishes.

So the behaviour is as desired, thanks!

Is the parameter "min_data" aka "min_data_in_leaf", "min_data_per_leaf", "min_child_samples" not in the list of problematic factors? I remember that the python prototype mentioned it but now it can be updated without warning. Would be fantastic as this is one of the most important parameters to tune.

guolinke:
@mayer79 
The `min_data` is used to filter some features that cannot split. For example, when min_data=20 and one feature with 10 `0`s and 30 `1`s, this feature cannot be splitted. This is for the efficiency purpose.
Therefore, Updating 'min_data` in dataset will not affect the result, as these filter features will not be chosen by trees as well.

PR #1879, opened 2018-11-25
[docs] fixing max_depth param description

KOLANICH:
https://github.com/Microsoft/LightGBM/blob/59f10453dc4a3841339702524124063fa5db6870/src/treelearner/serial_tree_learner.cpp#L351

BTW, should it be unsigned?

### DIFF
--- a/include/LightGBM/config.h
+++ b/include/LightGBM/config.h
@@ -208,7 +208,7 @@ public:
   #pragma region Learning Control Parameters
 
   // desc = limit the max depth for tree model. This is used to deal with over-fitting when ``#data`` is small. Tree still grows leaf-wise
-  // desc = ``< 0`` means no limit
+  // desc = ``<= 0`` means no limit
   int max_depth = -1;
 
   // alias = min_data_per_leaf, min_data, min_child_samples


KOLANICH:
BTW, why not to do it on CI automatically?

StrikerRUS:
Because after generating anything on CI side it should be pushed back to GitHub.

KOLANICH:
>Because after generating anything on CI side it should be pushed back to GitHub.

1. What's the problem with it? Azure Pipelines support secret variables.
2. If the docs are generated from the source, why not to remove them completely from the repo?

StrikerRUS:
> 1. What's the problem with it? Azure Pipelines support secret variables.

It's still insecure. It produces additional mess in the repo. Imagine: every commit is duplicated by the CI.

>2.  If the docs are generated from the source, 

Only params description is generated from the source.

StrikerRUS:
@KOLANICH Remembering your position about the CLA https://github.com/microsoft/LightGBM/pull/1873#issuecomment-454150389, we are taking over your PRs, if you don't mind.
> so feel free to close the PR and add the links and everything else you feel needed yourself from your account and your name, I don't object, …

Can you please clarify your comment?
https://github.com/microsoft/LightGBM/pull/1879#issue-233389039
>https://github.com/Microsoft/LightGBM/blob/59f10453dc4a3841339702524124063fa5db6870/src/treelearner/serial_tree_learner.cpp#L351
BTW, should it be unsigned?

KOLANICH:
I just wonder why a signed int is used for a parameter which semantics is to be a number >= 0. Shouldn't an unsigned int be used for that?

StrikerRUS:
Ah, got it, thanks!
But users can pass a negative value, e.g. `-1` to indicate "no limit".

KOLANICH:
>But users can pass a negative value, e.g. -1 to indicate "no limit".

But 0 is also "no limit".

StrikerRUS:
You are right!
It's possible save all semantics with unsigned value. However, this change will be not backward compatible.

BTW, at present, all params are signed values, even `num_boost_round` and other which should be strictly greater zero.

StrikerRUS:
Proposed changes were made in #2155.

Issue #1893, opened 2018-12-06
Not able to train with `dart` and `early_stopping_rounds`


RickoClausen:
## Environment info

Operating System: MacOS Mojave and Ubuntu

CPU/GPU model: CPU

C++/Python/R version: Python

## Error message

When using the `dart` booting type the model is not trained when applying `early_stopping_rounds`.
The rmse after training is not the same as it was at the stopping point in the training.

When I use `gbdt` the model trains fine, and I am able to reproduce the rmse from the in-training.

## Reproducible examples
```python
import lightgbm
import numpy as np

np.random.seed(1234)

params = {
    "early_stopping_rounds": 100,
    "metric": "root_mean_squared_error",
    "objective": "regression",
    "num_boost_round": 1000,
    "boosting_type": "dart",
}

size = (245688, 470)
x = np.random.exponential(scale=10, size=size)
y = 2 * x[:, 0] + np.random.exponential(scale=2, size=(size[0],))

x_val = np.random.exponential(scale=10, size=(int(size[0] / 13), size[1]))
y_val = 2 * x_val[:, 0] + np.random.exponential(scale=2, size=(int(size[0] / 13),))

model = lightgbm.LGBMModel(**params)

model.fit(x, y, eval_set=[(x, y), (x_val, y_val)], verbose=50)
train_pred = model.predict(x)
rmse = np.sqrt(np.mean((y - train_pred) ** 2))
print(f"Train rmse: {rmse}")
```

Output:
```bash
UserWarning: Starting from version 2.2.1, the library file in distribution wheels for macOS is built by the Apple Clang (Xcode_8.3.1) compiler.
This means that in case of installing LightGBM from PyPI via the ``pip install lightgbm`` command, you don't need to install the gcc compiler anymore.
Instead of that, you need to install the OpenMP library, which is required for running LightGBM on the system with the Apple Clang compiler.
You can install the OpenMP library by the following command: ``brew install libomp``.
  "You can install the OpenMP library by the following command: ``brew install libomp``.", UserWarning)
UserWarning: Found `num_boost_round` in params. Will use it instead ofargument
  warnings.warn("Found `{}` in params. Will use it instead of argument".format(alias))
UserWarning: Found `early_stopping_rounds` in params. Will use it instead of argument
  warnings.warn("Found `{}` in params. Will use it instead of argument".format(alias))
Training until validation scores don't improve for 100 rounds.
[...]

PR #1895, opened 2018-12-06
[python] disabled early stopping in dart mode

StrikerRUS:
Refer to https://github.com/Microsoft/LightGBM/issues/1893#issuecomment-444803315.

Maybe error is better, because users tend to completely ignore warnings?

### DIFF
--- a/python-package/lightgbm/callback.py
+++ b/python-package/lightgbm/callback.py
@@ -178,8 +178,16 @@ def early_stopping(stopping_rounds, verbose=True):
     best_iter = []
     best_score_list = []
     cmp_op = []
+    enabled = [True]
 
     def _init(env):
+        enabled[0] = not any((boost_alias in env.params
+                              and env.params[boost_alias] == 'dart') for boost_alias in ('boosting',
+                                                                                         'boosting_type',
+                                                                                         'boost'))
+        if not enabled[0]:
+            warnings.warn('Early stopping is not available in dart mode')
+            return
         if not env.evaluation_result_list:
             raise ValueError('For early stopping, '
                              'at least one dataset and eval metric is required for evaluation')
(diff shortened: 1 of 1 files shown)


guolinke:
@StrikerRUS I think we can disable early_stoping when using dart, even when early_stopping is set.

StrikerRUS:
@guolinke At cpp side?

guolinke:
@StrikerRUS the change of cpp side is in another PR.
The logic of early stopping is independent in cpp/python/R, so we need to disable it in python side as well.

StrikerRUS:
@guolinke Ah, I see!

StrikerRUS:
@guolinke I've updated PR. Unfortunately, I'm not familiar with `callback` module yet, therefore this solution might be ugly. Please check.

python-package/lightgbm/callback.py
```diff
@@ -198,8 +199,17 @@ def _init(env):
                 cmp_op.append(lt)
 
     def _callback(env):
+        enabled = not any((boost_alias in env.params
+                           and env.params[boost_alias] == 'dart') for boost_alias in ('boosting',
+                                                                                      'boosting_type',
+                                                                                      'boost'))
```
guolinke: maybe this can be moved into _init(env) ?

StrikerRUS: @guolinke Sorry, I was in a hurry and mislooked [this aspect](https://stackoverflow.com/questions/5218895/python-nested-functions-variable-scoping), which made me think that it's impossible to set `enabled` variable in `_init` function. I've reworked this.

Teeeto:
Sometimes early stopping is needed, disabling it may render dart mode quite useless. I'd suggest to save/override the best model (or at least a mask for dropped trees) in memory at each iteration and return the proper model when the early stopping is triggered.

PR #1906, opened 2018-12-15
Fix bugs in RF

### LINKED ISSUE #1843: Issue with trained model output for objectives with raw score transformations
For some objective types, LightGBM transforms the raw score (sum of tree outputs) to produce the model output.  For example, for Poisson, Gamma, and Tweedie regressions, the raw score is passed through the `exp` function.  However, when the boosting type is `rf` (random forest), these transforms don't appear to be applied, instead the output of the model is just the average of the tree outputs.

Similarly, when `reg_sqrt` is true, the raw score is squared to produce the model output.  However, when either the boosting type is `rf` OR the objective type is Poisson, Gamma, Tweedie, or Huber, this output transform does not appear to be applied.

For binary classification, again if the boosting type is `rf`, it appears that the model outputs the average of the tree outputs, but does not apply the `1 / Math.Exp(-Sigmoid * RawScore)` transform to produce a probability.

Is this to be expected?  Shouldn't the choice of `reg_sqrt` and whether or not boosting type `rf` is used be independent of the model output meaning?

PS: Just a guess, but looking at the generated C++ code for the tree:
```
void GBDT::Predict(const double* features, double *output, const PredictionEarlyStopInstance* early_stop) const {
	PredictRaw(features, output, early_stop);
	if (average_output_) {
		for (int k = 0; k < num_tree_per_iteration_; ++k) {
			output[k] /= num_iteration_for_pred_;
		}
	}
	else if (objective_function_ != nullptr) {
		objective_function_->ConvertOutput(output, output);
	}
}
```
It seems like the `else if` should just be an `if`, as otherwise if averaging is done (as is the case for random forest), the `ConvertOutput` methods are never called?

guolinke:
fix #1843

### DIFF
--- a/include/LightGBM/objective_function.h
+++ b/include/LightGBM/objective_function.h
@@ -42,6 +42,13 @@ public:
                                  const data_size_t*,
                                  data_size_t) const { return ori_output; }
 
+  virtual double RenewTreeOutput(double ori_output, double,
+                                  const data_size_t*,
+                                  const data_size_t*,
+                                  data_size_t) const {
+                                  return ori_output;
+  }
+
   virtual double BoostFromScore(int /*class_id*/) const { return 0.0; }
 
   virtual bool ClassNeedTrain(int /*class_id*/) const { return true; }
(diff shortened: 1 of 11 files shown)


mjmckp:
I've tested this with https://github.com/rca22/LightGBM.Net and it seems to have resolved the issue, thanks.

guolinke:
@mjmckp Thank you!

PR #451, opened 2017-04-24
python-package: fix creating eval_set in LGBMClassifier

henry0312:
This PR fixes creating eval_set in LGBMClassifier.
The current master (0eb9921991d21f7a61f5404494bad810811b1209) treats traing set as valid set even when valid set is traing set.  

### Sample script

```python
import numpy
numpy.random.seed(42)
import lightgbm

train_x = numpy.random.rand(100, 10)
train_y = numpy.random.randint(2, size=100)
valid_x = numpy.random.rand(100, 10)
valid_y = numpy.random.randint(2, size=100)

gbm = lightgbm.LGBMClassifier()
gbm.fit(train_x, train_y, eval_set=[(train_x, train_y), (valid_x, valid_y)])
```

### Before

```
❯ python test.py
[1]	valid_0's binary_logloss: 0.680061	valid_1's binary_logloss: 0.693353
[2]	valid_0's binary_logloss: 0.6688	valid_1's binary_logloss: 0.698276
[3]	valid_0's binary_logloss: 0.659384	valid_1's binary_logloss: 0.701111
[4]	valid_0's binary_logloss: 0.651245	valid_1's binary_logloss: 0.706889
[5]	valid_0's binary_logloss: 0.643804	valid_1's binary_logloss: 0.707129
[6]	valid_0's binary_logloss: 0.637556	valid_1's binary_logloss: 0.713327
[7]	valid_0's binary_logloss: 0.630633	valid_1's binary_logloss: 0.711053
[8]	valid_0's binary_logloss: 0.626307	valid_1's binary_logloss: 0.711603
[9]	valid_0's binary_logloss: 0.621181	valid_1's binary_logloss: 0.712893
[10]	valid_0's binary_logloss: 0.616207	valid_1's binary_logloss: 0.712305
```

### After

```
❯ python test.py
[1]	training's binary_logloss: 0.680061	valid_1's binary_logloss: 0.693353
[2]	training's binary_logloss: 0.6688	valid_1's binary_logloss: 0.698276
[3]	training's binary_logloss: 0.659384	valid_1's binary_logloss: 0.701111
[4]	training's binary_logloss: 0.651245	valid_1's binary_logloss: 0.706889
[5]	training's binary_logloss: 0.643804	valid_1's binary_logloss: 0.707129
[6]	training's binary_logloss: 0.637556	valid_1's binary_logloss: 0.713327
[7]	training's binary_logloss: 0.630633	valid_1's binary_logloss: 0.711053
[...]

PR #1916, opened 2018-12-19
[python] fix creating train_set in fit

henry0312:
https://github.com/Microsoft/LightGBM/blob/cc99f0d36ae929eb02b22a072823ab7c6d3155ab/python-package/lightgbm/sklearn.py#L519
may False even if valid_data[0] is X and valid_data[1] is y actually, because `check_X_y` might return copy of X and y.
https://scikit-learn.org/0.20/modules/generated/sklearn.utils.check_X_y.html

cf. https://github.com/Microsoft/LightGBM/pull/451


henry0312:
I want to merge this PR as possible as we can, because it is needed in #1912.
Thanks.

tests/python_package_test/test_sklearn.py
```diff
@@ -279,3 +279,13 @@ def test_predict(self):
         self.assertRaises(AssertionError,
                           np.testing.assert_allclose,
                           res_engine, res_sklearn_params)
+
+    def test_evaluate_train_set(self):
+        X, y = load_boston(True)
+        X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.1, random_state=42)
+        gbm = lgb.LGBMRegressor(n_estimators=10, silent=True)
+        gbm.fit(X_train, y_train, eval_set=[(X_train, y_train), (X_test, y_test)], verbose=False)
+        self.assertTrue('training' in gbm.evals_result_)
```
StrikerRUS: There is `assertIn` check: https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertIn
It is already used in our tests:
https://github.com/Microsoft/LightGBM/blob/cba824474897c8d7e71a3df261faaefa091a40c5/tests/python_package_test/test_engine.py#L383

henry0312:
Ah, yes. `assertIn` is better. I'll fix.

PR #1919, opened 2018-12-20
Fix build on macOS Mojave

henry0312:
Fixed #1898

Thanks to https://github.com/cryfs/cryfs/commit/08ed96680876c960d5aab0aabebb25206f0bda06

PR #1923, opened 2018-12-22
Fix build on macOS Mojave

### LINKED ISSUE #1898: LightGBM installation error_MacOSX10.14_Cmake
gcc -v
Configured with: --prefix=/Library/Developer/CommandLineTools/usr --with-gxx-include-dir=/Library/Developer/CommandLineTools/SDKs/MacOSX10.14.sdk/usr/include/c++/4.2.1
Apple LLVM version 10.0.0 (clang-1000.10.44.4)
Target: x86_64-apple-darwin18.2.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin

**Issue:**
I followed the installation guide and on 'cmake ..' step got following error:
############

-- The C compiler identification is Clang 4.0.1
-- The CXX compiler identification is Clang 4.0.1
-- Check for working C compiler: /anaconda3/bin/x86_64-apple-darwin13.4.0-clang
-- Check for working C compiler: /anaconda3/bin/x86_64-apple-darwin13.4.0-clang -- broken
CMake Error at /anaconda3/share/cmake-3.12/Modules/CMakeTestCCompiler.cmake:52 (message):
  The C compiler

    "/anaconda3/bin/x86_64-apple-darwin13.4.0-clang"

  is not able to compile a simple test program.

  It fails with the following output:

    Change Dir: /Users/adityajuvali/LightGBM/build/CMakeFiles/CMakeTmp
    
    Run Build Command:"/anaconda3/bin/make" "cmTC_4d700/fast"
    /anaconda3/bin/make -f CMakeFiles/cmTC_4d700.dir/build.make CMakeFiles/cmTC_4d700.dir/build
    make[1]: Entering directory '/Users/adityajuvali/LightGBM/build/CMakeFiles/CMakeTmp'
    Building C object CMakeFiles/cmTC_4d700.dir/testCCompiler.c.o
    /anaconda3/bin/x86_64-apple-darwin13.4.0-clang   -march=core2 -mtune=haswell -mssse3 -ftree-vectorize -fPIC -fPIE -fstack-protector-strong -O2 -pipe  -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX10.14.sdk   -o CMakeFiles/cmTC_4d700.dir/testCCompiler.c.o   -c /Users/adityajuvali/LightGBM/build/CMakeFiles/CMakeTmp/testCCompiler.c
    Linking C executable cmTC_4d700
    /anaconda3/bin/cmake -E cmake_link_script CMakeFiles/cmTC_4d700.dir/link.txt --verbose=1
    /anaconda3/bin/x86_64-apple-darwin13.4.0-clang -march=core2 -mtune=haswell -mssse3 -ftree-vectorize -fPIC -fPIE -fstack-protector-strong -O2 -pipe  -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX10.14.sdk -Wl,-search_paths_first -Wl,-headerpad_max_install_names -Wl,-pie -Wl,-headerpad_max_install_names -Wl,-dead_strip_dylibs  CMakeFiles/cmTC_4d700.dir/testCCompiler.c.o  -o cmTC_4d700 
    ld: warning: ignoring file /Library/Developer/CommandLineTools/SDKs/MacOSX10.14.sdk/usr/lib/libSystem.tbd, file was built for unsupported file format ( 0x2D 0x2D 0x2D 0x20 0x21 0x74 0x61 0x70 0x69 0x2D 0x74 0x62 0x64 0x2D 0x76 0x33 ) which is not the architecture being linked (x86_64): /Library/Developer/CommandLineTools/SDKs/MacOSX10.14.sdk/usr/lib/libSystem.tbd
    ld: dynamic main executables must link with libSystem.dylib for architecture x86_64
    clang-4.0: error: linker command failed with exit code 1 (use -v to see invocation)
    make[1]: *** [CMakeFiles/cmTC_4d700.dir/build.make:87: cmTC_4d700] Error 1
    make[1]: Leaving directory '/Users/adityajuvali/LightGBM/build/CMakeFiles/CMakeTmp'
    make: *** [Makefile:121: cmTC_4d700/fast] Error 2
    

  

  CMake will not be able to correctly generate this project.
  CMakeLists.txt:7 (PROJECT)

-- Configuring incomplete, errors occurred!

#######

Please help. I am not able install even other models due to this same error.

henry0312:
Fixed #1898

- https://iscinumpy.gitlab.io/post/omp-on-high-sierra/
- https://cliutils.gitlab.io/modern-cmake/chapters/packages/OpenMP.html
- https://github.com/Homebrew/homebrew-core/pull/20589

### DIFF
--- a/docs/Installation-Guide.rst
+++ b/docs/Installation-Guide.rst
@@ -146,7 +146,13 @@ Only **Apple Clang** version 8.1 or higher is supported.
 
      git clone --recursive https://github.com/Microsoft/LightGBM ; cd LightGBM
      mkdir build ; cd build
-     cmake ..
+     cmake \
+       -DOpenMP_C_FLAGS="-Xpreprocessor -fopenmp -I/usr/local/opt/libomp/include" \
+       -DOpenMP_C_LIB_NAMES="omp" \
+       -DOpenMP_CXX_FLAGS="-Xpreprocessor -fopenmp -I/usr/local/opt/libomp/include" \
+       -DOpenMP_CXX_LIB_NAMES="omp" \
+       -DOpenMP_omp_LIBRARY=/usr/local/opt/libomp/lib/libomp.dylib \
+       ..
      make -j4
 
 gcc
--- a/docs/Installation-Guide.rst
+++ b/docs/Installation-Guide.rst
@@ -149,11 +149,11 @@ Only **Apple Clang** version 8.1 or higher is supported.
 
      # For Mojave (10.14)
      cmake \
-       -DOpenMP_C_FLAGS="-Xpreprocessor -fopenmp -I/usr/local/opt/libomp/include" \
+       -DOpenMP_C_FLAGS="-Xpreprocessor -fopenmp -I$(brew --prefix libomp)/include" \
        -DOpenMP_C_LIB_NAMES="omp" \
-       -DOpenMP_CXX_FLAGS="-Xpreprocessor -fopenmp -I/usr/local/opt/libomp/include" \
+       -DOpenMP_CXX_FLAGS="-Xpreprocessor -fopenmp -I$(brew --prefix libomp)/include" \
        -DOpenMP_CXX_LIB_NAMES="omp" \
-       -DOpenMP_omp_LIBRARY=/usr/local/opt/libomp/lib/libomp.dylib \
+       -DOpenMP_omp_LIBRARY=$(brew --prefix libomp)/lib/libomp.dylib \
        ..
 
      # For High Sierra or earlier (<= 10.13)
--- a/docs/Installation-Guide.rst
+++ b/docs/Installation-Guide.rst
@@ -147,7 +147,7 @@ Only **Apple Clang** version 8.1 or higher is supported.
      git clone --recursive https://github.com/Microsoft/LightGBM ; cd LightGBM
      mkdir build ; cd build
 
-     # For Mojave
+     # For Mojave (10.14)
      cmake \
        -DOpenMP_C_FLAGS="-Xpreprocessor -fopenmp -I/usr/local/opt/libomp/include" \
        -DOpenMP_C_LIB_NAMES="omp" \
@@ -156,7 +156,7 @@ Only **Apple Clang** version 8.1 or higher is supported.
        -DOpenMP_omp_LIBRARY=/usr/local/opt/libomp/lib/libomp.dylib \
        ..
 
-     # For High Sierra or earlier
+     # For High Sierra or earlier (<= 10.13)
      cmake ..
 
      make -j4
--- a/docs/Installation-Guide.rst
+++ b/docs/Installation-Guide.rst
@@ -146,6 +146,8 @@ Only **Apple Clang** version 8.1 or higher is supported.
 
      git clone --recursive https://github.com/Microsoft/LightGBM ; cd LightGBM
      mkdir build ; cd build
+
+     # For Mojave
      cmake \
        -DOpenMP_C_FLAGS="-Xpreprocessor -fopenmp -I/usr/local/opt/libomp/include" \
        -DOpenMP_C_LIB_NAMES="omp" \
@@ -153,6 +155,10 @@ Only **Apple Clang** version 8.1 or higher is supported.
        -DOpenMP_CXX_LIB_NAMES="omp" \
        -DOpenMP_omp_LIBRARY=/usr/local/opt/libomp/lib/libomp.dylib \
        ..
+
+     # For High Sierra or earlier
+     cmake ..
+
      make -j4
 
 gcc
--- a/docs/Installation-Guide.rst
+++ b/docs/Installation-Guide.rst
@@ -448,7 +448,20 @@ Only **Apple Clang** version 8.1 or higher is supported.
 
      git clone --recursive https://github.com/Microsoft/LightGBM ; cd LightGBM
      mkdir build ; cd build
+
+     # For Mojave (10.14)
+     cmake \
+       -DUSE_MPI=ON \
+       -DOpenMP_C_FLAGS="-Xpreprocessor -fopenmp -I$(brew --prefix libomp)/include" \
+       -DOpenMP_C_LIB_NAMES="omp" \
+       -DOpenMP_CXX_FLAGS="-Xpreprocessor -fopenmp -I$(brew --prefix libomp)/include" \
+       -DOpenMP_CXX_LIB_NAMES="omp" \
+       -DOpenMP_omp_LIBRARY=$(brew --prefix libomp)/lib/libomp.dylib \
+       ..
+
+     # For High Sierra or earlier (<= 10.13)
      cmake -DUSE_MPI=ON ..
+
      make -j4
 
 gcc
--- a/python-package/setup.py
+++ b/python-package/setup.py
@@ -119,14 +119,18 @@ def compile_cpp(use_mingw=False, use_gpu=False, use_mpi=False, nomp=False,
     if nomp:
         cmake_cmd.append("-DUSE_OPENMP=OFF")
     if system() == 'Darwin' and not nomp:
-        # https://github.com/Homebrew/homebrew-core/pull/20589
-        cmake_cmd.extend([
-            '-DOpenMP_C_FLAGS="-Xpreprocessor -fopenmp -I/usr/local/opt/libomp/include"',
-            '-DOpenMP_C_LIB_NAMES="omp"',
-            '-DOpenMP_CXX_FLAGS="-Xpreprocessor -fopenmp -I/usr/local/opt/libomp/include"',
-            '-DOpenMP_CXX_LIB_NAMES="omp"',
-            '-DOpenMP_omp_LIBRARY=/usr/local/opt/libomp/lib/libomp.dylib',
-        ])
+        cc = os.environ.get('CC')
+        cxx = os.environ.get('CXX')
+        if not (cc and cc.startswith('gcc') and cxx and cxx.startswith('g++')):
+            # Apple Clang
+            # https://github.com/Homebrew/homebrew-core/pull/20589
+            cmake_cmd.extend([
+                '-DOpenMP_C_FLAGS="-Xpreprocessor -fopenmp -I/usr/local/opt/libomp/include"',
+                '-DOpenMP_C_LIB_NAMES="omp"',
+                '-DOpenMP_CXX_FLAGS="-Xpreprocessor -fopenmp -I/usr/local/opt/libomp/include"',
+                '-DOpenMP_CXX_LIB_NAMES="omp"',
+                '-DOpenMP_"omp"_LIBRARY=/usr/local/opt/libomp/lib/libomp.dylib'
+            ])
     if use_hdfs:
         cmake_cmd.append("-DUSE_HDFS=ON")
 
--- a/python-package/setup.py
+++ b/python-package/setup.py
@@ -125,11 +125,11 @@ def compile_cpp(use_mingw=False, use_gpu=False, use_mpi=False, nomp=False,
             # Apple Clang
             # https://github.com/Homebrew/homebrew-core/pull/20589
             cmake_cmd.extend([
-                '-DOpenMP_C_FLAGS="-Xpreprocessor -fopenmp -I/usr/local/opt/libomp/include"',
-                '-DOpenMP_C_LIB_NAMES="omp"',
-                '-DOpenMP_CXX_FLAGS="-Xpreprocessor -fopenmp -I/usr/local/opt/libomp/include"',
-                '-DOpenMP_CXX_LIB_NAMES="omp"',
-                '-DOpenMP_"omp"_LIBRARY=/usr/local/opt/libomp/lib/libomp.dylib'
+                '-DOpenMP_C_FLAGS=-Xpreprocessor -fopenmp -I/usr/local/opt/libomp/include',
+                '-DOpenMP_C_LIB_NAMES=omp',
+                '-DOpenMP_CXX_FLAGS=-Xpreprocessor -fopenmp -I/usr/local/opt/libomp/include',
+                '-DOpenMP_CXX_LIB_NAMES=omp',
+                '-DOpenMP_omp_LIBRARY=/usr/local/opt/libomp/lib/libomp.dylib'
             ])
     if use_hdfs:
         cmake_cmd.append("-DUSE_HDFS=ON")
--- a/python-package/setup.py
+++ b/python-package/setup.py
@@ -118,8 +118,18 @@ def compile_cpp(use_mingw=False, use_gpu=False, use_mpi=False, nomp=False,
         cmake_cmd.append("-DUSE_MPI=ON")
     if nomp:
         cmake_cmd.append("-DUSE_OPENMP=OFF")
+    if system() == 'Darwin' and not nomp:
+        # https://github.com/Homebrew/homebrew-core/pull/20589
+        cmake_cmd.extend([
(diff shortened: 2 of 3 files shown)


docs/Installation-Guide.rst
```diff
@@ -146,7 +146,13 @@ Only **Apple Clang** version 8.1 or higher is supported.
 
      git clone --recursive https://github.com/Microsoft/LightGBM ; cd LightGBM
      mkdir build ; cd build
-     cmake ..
+     cmake \
+       -DOpenMP_C_FLAGS="-Xpreprocessor -fopenmp -I/usr/local/opt/libomp/include" \
+       -DOpenMP_C_LIB_NAMES="omp" \
+       -DOpenMP_CXX_FLAGS="-Xpreprocessor -fopenmp -I/usr/local/opt/libomp/include" \
```
henry0312: https://github.com/Homebrew/homebrew-core/blob/master/Formula/libomp.rb#L36

henry0312:
This is much better than https://github.com/Microsoft/LightGBM/pull/1919.

StrikerRUS:
@henry0312 
> This is much better than #1919.

Totally agree! :-)

Maybe it should be mentioned that this fearful command is needed only for Mojave? Other users can still use simple `cmake ..`, right?

henry0312:
> Maybe it should be mentioned that this fearful command is needed only for Mojave?

Yes, the cmake options is needed only for Mojave, although they works fine for High Sierra too.

> Other users can still use simple `cmake ..`, right?

yes.

StrikerRUS:
> Yes, the cmake options is needed only for Mojave, although they works fine for High Sierra too.

OK, I see. Please document it.

Also, I think that we need to add second attempt for compilation with new CMake options for Apple here:
https://github.com/Microsoft/LightGBM/blob/2323cb3befa0515dd266d4a1393a37c833c11b83/python-package/setup.py#L164-L168

Something like:
```
if Apple and status != 0:
    second attempt
```

python-package/setup.py
```diff
@@ -118,8 +118,22 @@ def compile_cpp(use_mingw=False, use_gpu=False, use_mpi=False, nomp=False,
...
+        cxx = os.environ.get('CXX')
+        if not (cc and cc.startswith('gcc') and cxx and cxx.startswith('g++')):
+            # Apple Clang
+            # https://github.com/Homebrew/homebrew-core/pull/20589
+            cmake_cmd.extend([
+                '-DOpenMP_C_FLAGS="-Xpreprocessor -fopenmp -I/usr/local/opt/libomp/include"',
+                '-DOpenMP_C_LIB_NAMES="omp"',
+                '-DOpenMP_CXX_FLAGS="-Xpreprocessor -fopenmp -I/usr/local/opt/libomp/include"',
+                '-DOpenMP_CXX_LIB_NAMES="omp"',
+                '-DOpenMP_"omp"_LIBRARY=/usr/local/opt/libomp/lib/libomp.dylib'
```
henry0312: I don't really know why this is `-DOpenMP_"omp"_LIBRARY`, not `-DOpenMP_omp_LIBRARY` 🤔
However, `python setup.py install` works only when ``-DOpenMP_"omp"_LIBRARY``.
If `-DOpenMP_omp_LIBRARY`, the bellow errors will occur:

```
-- The C compiler identification is AppleClang 10.0.0.10001145
-- The CXX compiler identification is AppleClang 10.0.0.10001145
-- Check for working C compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc
-- Check for working C compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++
-- Check for working CXX compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
CMake Error at /usr/local/Cellar/cmake/3.13.2/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find OpenMP_C (missing: OpenMP_"omp"_LIBRARY)
  /usr/local/Cellar/cmake/3.13.2/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:378 (_FPHSA_FAILURE_MESSAGE)
  /usr/local/Cellar/cmake/3.13.2/share/cmake/Modules/FindOpenMP.cmake:473 (find_package_handle_standard_args)
  CMakeLists.txt:78 (find_package)


-- Configuring incomplete, errors occurred!
See also "/Users/henry/.ghq/github.com/henry0312/LightGBM/python-package/build_cpp/CMakeFiles/CMakeOutput.log".
See also "/Users/henry/.ghq/github.com/henry0312/LightGBM/python-package/build_cpp/CMakeFiles/CMakeError.log".
```

Would anyone tell me what happens in `setup.py`, please.

henry0312:
Okay, updated.

docs/Installation-Guide.rst
```diff
@@ -146,7 +146,19 @@ Only **Apple Clang** version 8.1 or higher is supported.
 
      git clone --recursive https://github.com/Microsoft/LightGBM ; cd LightGBM
      mkdir build ; cd build
+
+     # For Mojave
```
StrikerRUS:
```suggestion
      # For Mojave (10.14)
```

StrikerRUS:
```suggestion
      # For Mojave (10.14)
```

StrikerRUS:
```
# For Mojave (10.14)
```

docs/Installation-Guide.rst
```diff
@@ -146,7 +146,19 @@ Only **Apple Clang** version 8.1 or higher is supported.
...
+     cmake \
+       -DOpenMP_C_FLAGS="-Xpreprocessor -fopenmp -I/usr/local/opt/libomp/include" \
+       -DOpenMP_C_LIB_NAMES="omp" \
+       -DOpenMP_CXX_FLAGS="-Xpreprocessor -fopenmp -I/usr/local/opt/libomp/include" \
+       -DOpenMP_CXX_LIB_NAMES="omp" \
+       -DOpenMP_omp_LIBRARY=/usr/local/opt/libomp/lib/libomp.dylib \
+       ..
+
+     # For High Sierra or earlier
```
StrikerRUS:
```
… or earlier (<= 10.13)
```

docs/Installation-Guide.rst
```diff
@@ -146,7 +146,19 @@ Only **Apple Clang** version 8.1 or higher is supported.
 
      git clone --recursive https://github.com/Microsoft/LightGBM ; cd LightGBM
      mkdir build ; cd build
+
+     # For Mojave
+     cmake \
+       -DOpenMP_C_FLAGS="-Xpreprocessor -fopenmp -I/usr/local/opt/libomp/include" \
```
StrikerRUS: I think that 
```
-I"$(brew --prefix libomp)/include"
```
is better, because it shows the general case, from where this path comes from and how to customize it.

We provide all instructions for `brew`, so `brew` here is OK too.

docs/Installation-Guide.rst
```diff
@@ -146,7 +146,19 @@ Only **Apple Clang** version 8.1 or higher is supported.
...
+       -DOpenMP_C_LIB_NAMES="omp" \
+       -DOpenMP_CXX_FLAGS="-Xpreprocessor -fopenmp -I/usr/local/opt/libomp/include" \
```
StrikerRUS: Same as above about general case for path.

docs/Installation-Guide.rst
```diff
@@ -146,7 +146,19 @@ Only **Apple Clang** version 8.1 or higher is supported.
...
+       -DOpenMP_CXX_LIB_NAMES="omp" \
+       -DOpenMP_omp_LIBRARY=/usr/local/opt/libomp/lib/libomp.dylib \
```
StrikerRUS: Same as above about general case for path.

python-package/setup.py
```diff
@@ -118,8 +118,22 @@ def compile_cpp(use_mingw=False, use_gpu=False, use_mpi=False, nomp=False,
         cmake_cmd.append("-DUSE_MPI=ON")
     if nomp:
         cmake_cmd.append("-DUSE_OPENMP=OFF")
+    if system() == 'Darwin' and not nomp:
+        cc = os.environ.get('CC')
+        cxx = os.environ.get('CXX')
+        if not (cc and cc.startswith('gcc') and cxx and cxx.startswith('g++')):
+            # Apple Clang
+            # https://github.com/Homebrew/homebrew-core/pull/20589
+            cmake_cmd.extend([
```
StrikerRUS: We shouldn't use hardcode paths for the first build attempt. Let's use hardcoded paths only after first attempt fail:
```
...
 else:  # Linux, Darwin (macOS), etc.
        logger.info("Starting to compile with CMake.")
        if system() == 'Darwin' and not nomp and not (cc and cc.startswith('gcc') and cxx and cxx.startswith('g++')):
            status = silent_call(cmake_cmd)
            if status != 0:
                logger.warning("Compilation failed.")
                logger.info("Starting to compile with OpenMP path guesses.")
                clear_path(os.path.join(CURRENT_DIR, "build_cpp"))
                <your modification here>
        else:
            silent_call(cmake_cmd, raise_error=True, error_msg='Please install CMake and all required dependencies first')
...
```

python-package/setup.py
StrikerRUS: Please try to remove **all** quotes from the command.

python-package/setup.py
henry0312: > We shouldn't use hardcoded paths for the first build attempt.

Why is this?
I think this trial is a really waste.

python-package/setup.py
henry0312: I'm sorry, I don't understand what you mean 😢
Which quotes should I remove from the command?

python-package/setup.py
StrikerRUS: 1. This workaround was tested only on High Sierra and Mojave. We don't know how it works on earlier versions. However, the current method was tested by the time and there were no any complaints.
2. There could be some other problems with this workaround which we don't know.
3. User can install OpenMP in another folder.

python-package/setup.py
StrikerRUS: I suppose **all**. For example:
```
'-DOpenMP_CXX_FLAGS=-Xpreprocessor -fopenmp -I/usr/local/opt/libomp/include',
'-DOpenMP_CXX_LIB_NAMES=omp',
```

StrikerRUS: It's only my assumption, based on the fact that we do not use `shell=True`. Sorry, that I don't check my hypotheses by myself, but I don't have access to Mojave.

henry0312:
```python
            cmake_cmd.extend([
                '-DOpenMP_C_FLAGS=-Xpreprocessor -fopenmp -I/usr/local/opt/libomp/include',
                '-DOpenMP_C_LIB_NAMES=omp',
                '-DOpenMP_CXX_FLAGS=-Xpreprocessor -fopenmp -I/usr/local/opt/libomp/include',
                '-DOpenMP_CXX_LIB_NAMES=omp',
                '-DOpenMP_omp_LIBRARY=/usr/local/opt/libomp/lib/libomp.dylib'
            ])
```

works!
Would you gime me some documetns that tell me why?

python-package/setup.py
henry0312: To be honest, it is correct for macOS users to set cmake options for AppleClang with OpenMP.
In ohter words, the current way (`cmake ..`) is wrong (than this change) because `libomp` requires the options.

Laurae2: @henry0312 @StrikerRUS What about something like this instead in README.md? (and let the user change the paths?)

```sh
DOpenMP_C_LIB_NAMES="omp" DOpenMP_CXX_FLAGS="-Xpreprocessor -fopenmp -I/usr/local/opt/libomp/include" DOpenMP_CXX_LIB_NAMES="omp" DOpenMP_omp_LIBRARY=/usr/local/opt/libomp/lib/libomp.dylib bash -c 'cmake ..'
```

python-package/setup.py
StrikerRUS: Sorry, it was my experience from VS command building (you can see it above in `setup.py`) more than year ago. I may be wrong, but my intuition is in that quotes are needed only for interaction through a shell to escape whitespaces. However, we are using `subprocess` without `shell=True` (which is insecure, BTW, as Python docs says), so all quotes are goes directly as part of an argument.

I'll try to find some official information about it for you

python-package/setup.py
StrikerRUS: @henry0312 
> In ohter words, the current way (cmake ..) is wrong (than this change) because libomp requires the options.

Maybe... But if everything works (OK, work**ed** before 10.14), why should we ask users do more things then needed? 

@henry0312 @Laurae2 Hmm, what do you think about new setup args `openmp-include-dir` and `openmp-library` like we already have for [GPU version](https://github.com/Microsoft/LightGBM/tree/master/python-package#build-gpu-version)?

Then we can do the following:
```
1. Try to compile with default "cmake .." for earlier macOS versions and future CMake changes.
2. If new args are provided, compile with them.
3. If not 2. (not provided or failed), then try to compile with path guesses for the maximum backward compatibility.
```

henry0312: > why should we ask users do more things then needed?

No, no.
This change will never ask users to do more anything.
For users, nothing will be changed.
For us, this change will correct building options.  That's all.

docs/Installation-Guide.rst
henry0312: Unfortunately, subshell doesn't seem to work in subprocess 😢
https://gist.github.com/henry0312/81664a9710056f0d59136813ebff2651 (L.15, 16 and 54).

StrikerRUS: Sorry, maybe I'm missing something, but my comment was about manual compilation of CLI version. Did you mean Python's subprocess?

henry0312: I'm sorry I misunderstood 🙇
I'll fix.

henry0312:
I updated docs and fixed setup.py again.

docs/Installation-Guide.rst
StrikerRUS: No problem! Thank you very much for your brilliant work! 😃

python-package/setup.py
StrikerRUS: @henry0312 I've tried to google, but unfortunately without any luck. Sorry, it was more that year ago and I can't recollect in memory what I found that made me omit quotes. 😢 
Here's only partially related answer on stackoverflow: https://stackoverflow.com/a/14632378

docs/Installation-Guide.rst
```diff
@@ -146,7 +146,19 @@ Only **Apple Clang** version 8.1 or higher is supported.
 
      git clone --recursive https://github.com/Microsoft/LightGBM ; cd LightGBM
      mkdir build ; cd build
+
+     # For Mojave (10.14)
```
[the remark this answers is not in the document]
henry0312: done

StrikerRUS:
@henry0312 The feedback has been received. Can we merge now?

Issue #1846, opened 2018-11-15
[R] Custom objective for multi-class 


albedan:
I'm trying to use a custom objective and metric for a multiclass problem, following @Laurae2 example here: https://github.com/Microsoft/LightGBM/blob/master/R-package/demo/early_stopping.R

Using R 3.5.1 and lightgbm 2.2.2.
So far I managed to implement the eval metric, but objective seems tricky.

I think a trivial example should be useful: maybe just implementing the default multi class logloss with a custom objective in R (in order to be able to compare the results with the default objective).

BruceZhaoR:
@albedan https://github.com/Microsoft/LightGBM/issues/1682  maybe helpful :grin:

albedan:
> 
> 
> @albedan #1682 maybe helpful grin

Thanks @BruceZhaoR, but that one is about the evaluation metric (and I have no issues with that).
I'm referring to the objective function, i.e. the one that defines the optimization through gradient and hessian

BruceZhaoR:
multi class logloss with a custom objective :
https://github.com/Microsoft/LightGBM/blob/425503d7e254130eaea4a5ff287f324f5a5cc1ea/R-package/demo/multiclass_custom_objective.R#L37-L55

It's that what you want?

albedan:
@BruceZhaoR ,
Actually I think there could be a couple of issues:
1. The example has balanced classes. Let's change it at the beginning of the script:
```
train <- as.matrix(iris[c(1:30, 51:95, 101:140), ])
test <- as.matrix(iris[c(31:50, 96:100, 141:150), ])
```
The builtin model starts with a "boost from average" while the custom one does not (and setting the flag does not seem to work).
2. Running the example as it is, the final "identical" function returns False, as the two predictions are different. Actually, reshaping and applying manually a softmax function gets to identical predictions (but I expected the raw ones to be identical, too).

vnijs:
I'm seeing the same as @albedan. See output below.

@albedan Could you perhaps share/post the code you used to get identical probability predictions? Thanks

```
> # compare predictions
> identical(preds_builtin, preds_custom)
[1] FALSE
> head(data.frame(preds_builtin, preds_custom))
  preds_builtin preds_custom
1      1.540058     2.638670
2     -2.912387    -1.813775
3     -2.912387    -1.813775
4      1.540058     2.638670
5     -2.912387    -1.813775
6     -2.912387    -1.813775
```

vnijs:
[...]

PR #1940, opened 2019-01-11
[R] Fix multiclass demo

maximilianeber:
This PR solves two issues with the R demo for custom multiclass objectives as mentioned in #1846:

- Currently, the built-in method and the custom method do not return the same predictions in the sample script (the last statement evaluates as `FALSE`)
- The two methods do return the same probabilities but only if the classes are balanced

Both issues are due to the fact that the built-in method boosts from an `init_score` equal to the class mean by default. By contrast, the custom method boosts from zero. In the case of balanced classes, we are off only by a constant (hence we get the same probabilities); the case of imbalanced classes is more problematic.

Setting `boost_from_average=FALSE` fixes this. Alternatively, we could set `init_score` to zero explicitly.

### DIFF
--- a/R-package/demo/multiclass_custom_objective.R
+++ b/R-package/demo/multiclass_custom_objective.R
@@ -8,14 +8,20 @@ data(iris)
 # For instance: 0, 1, 2, 3, 4, 5...
 iris$Species <- as.numeric(as.factor(iris$Species)) - 1
 
-# We cut the data set into 80% train and 20% validation
+# Create imbalanced training data (20, 30, 40 examples for classes 0, 1, 2)
+train <- as.matrix(iris[c(1:20, 51:80, 101:140), ])
 # The 10 last samples of each class are for validation
-
-train <- as.matrix(iris[c(1:40, 51:90, 101:140), ])
 test <- as.matrix(iris[c(41:50, 91:100, 141:150), ])
+
 dtrain <- lgb.Dataset(data = train[, 1:4], label = train[, 5])
 dtest <- lgb.Dataset.create.valid(dtrain, data = test[, 1:4], label = test[, 5])
-valids <- list(test = dtest)
+
+# Built-in method uses class mean as initial score by default
+# Set to zero for equivalence with custom objective function
+setinfo(dtrain, "init_score", matrix(0, nrow = nrow(dtrain), ncol = 3))
+setinfo(dtest,  "init_score", matrix(0, nrow = nrow(dtest), ncol = 3))
+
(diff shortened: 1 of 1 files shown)


Laurae2:
@maximilianeber Can you create the demo in another file (like `multiclass_unbalanced`), and add a comment on the original file to check the other file you will create about boost_from_average / unbalanced data?

maximilianeber:
@Laurae2 I am happy to do that. However, I am not sure whether a second file adds much since the old example has to be fixed anyway (a minimal version would be setting `boost_from_average=FALSE`). My other edits are there only to clarify the discussion in #1846 (so we can close it). 

Please let me know what you prefer.

Laurae2:
Yes you are correct for #1846, actually most multiclass problems are not perfectly balanced like the case presented in the demo. We will keep the original file with your modifications.

R-package/demo/multiclass_custom_objective.R
```diff
@@ -78,8 +79,9 @@ model_custom <- lgb.train(list(),
...
 
-preds_custom <- predict(model_custom, test[, 1:4], rawscore = TRUE)
+preds_custom <- predict(model_custom, test[, 1:4], rawscore = TRUE, reshape = TRUE)
+probs_custom <- exp(preds_custom) / rowSums(exp(preds_custom))
 
 # compare predictions
-identical(preds_builtin, preds_custom)
-
+stopifnot(identical(probs_builtin, probs_custom))
+stopifnot(identical(preds_builtin, preds_custom))
```
Laurae2: Add trailing line (line at the end)

R-package/demo/multiclass_custom_objective.R
```diff
@@ -8,18 +8,19 @@ data(iris)
...
-
-train <- as.matrix(iris[c(1:40, 51:90, 101:140), ])
 test <- as.matrix(iris[c(41:50, 91:100, 141:150), ])
+
 dtrain <- lgb.Dataset(data = train[, 1:4], label = train[, 5])
 dtest <- lgb.Dataset.create.valid(dtrain, data = test[, 1:4], label = test[, 5])
-valids <- list(test = dtest)
+valids <- list(train = dtrain, test = dtest)
 
 # Method 1 of training with built-in multiclass objective
```
Laurae2: add comment about `boost_from_average = FALSE` : you can point to https://github.com/Microsoft/LightGBM/issues/1846

maximilianeber: Done

PR #1959, opened 2019-01-19
fixed minor typos in documentation

jameslamb:
Had a few minutes to kill today, so I ran `aspell check` over the `docs/` folder here. Please consider this PR to fix a few minor typos.


src/treelearner/gpu_tree_learner.cpp
```diff
@@ -538,7 +538,7 @@ void GPUTreeLearner::AllocateGPUMemory() {
   }
   // data transfer time
   std::chrono::duration<double, std::milli> end_time = std::chrono::steady_clock::now() - start_time;
-  Log::Info("%d dense feature groups (%.2f MB) transfered to GPU in %f secs. %d sparse feature groups", 
+  Log::Info("%d dense feature groups (%.2f MB) transferred to GPU in %f secs. %d sparse feature groups",
```
jameslamb: @Laurae2 good call. This is the only one I found (with `git grep transfered`)

jameslamb:
The only thing that failed on my Travis CI build is the `check-docs` task, looks like it's because one of the links is down:

```
Real URL   https://www.kaggle.com/c/allstate-claims-severity/discussion/24520
Check time 2.229 seconds
Size       89B
Result     Error: 502 Bad Gateway
```

I can get there in the browser, so guessing it was a temporary outage at Kaggle. Going to manually trigger a rebuild on the task

.gitignore
```diff
@@ -390,3 +390,6 @@ R-package/src/src/
 lightgbm_r/*
 lightgbm*.tar.gz
 lightgbm.Rcheck/
+
+# Files generate by aspell
```
StrikerRUS: "generate" -> "generate**d**"?

jameslamb: ha yes good catch, will fix right now

jameslamb: fixed

PR #1960, opened 2019-01-20
fix R's overflow

### LINKED ISSUE #1884: [R] int overflow in LGBM_BoosterCalcNumPredict_R
I'm skipping the reproducible example, but I can add one on request.

If I call `predict.lgb.Booster()` on a sufficiently large test dataset (e.g. 120 cols, 3.5M rows) in a multiclass setting with 700 classes (`n_col * n_class > .Machine$integer.max`), I get an error

```
Error in numeric(npred) : invalid 'length' argument

11: numeric(npred)
10: predictor$predict(data, num_iteration, rawscore, predleaf, predcontrib,
        header, reshape)
9: object$predict(data, num_iteration, rawscore, predleaf, predcontrib,
       header, reshape, ...)
8: predict.lgb.Booster(mod, test_dat, reshape = TRUE)
```

This comes from `LGBM_BoosterCalcNumPredict_R` returning something that looks like an overflowed value (`-1792749607` in my case), which in turn is an invalid `length` arg to `numeric()`.

A fix similar to e55c815821f0abbb23d52a391a622a2c96e6ba14, in response to #1861, might resolve this issue as well.

The same problem *might* also show up in
* `LGBM_BoosterGetNumPredict_R`
* `LGBM_BoosterGetPredict_R `

guolinke:
fix #1884
ping @Laurae2 , any other places we can use int64_t? 
I remember the length of string, and the number of elements in CSC are still limited by int32.

### DIFF
--- a/include/LightGBM/R_object_helper.h
+++ b/include/LightGBM/R_object_helper.h
@@ -124,9 +124,13 @@ typedef union { VECTOR_SER s; double align; } SEXPREC_ALIGN;
 
 #define R_INT_PTR(x)  ((int *) DATAPTR(x))
 
+#define R_INT64_PTR(x)  ((int64_t *) DATAPTR(x))
+
 #define R_REAL_PTR(x)     ((double *) DATAPTR(x))
 
 #define R_AS_INT(x) (*((int *) DATAPTR(x)))
+
+#define R_AS_INT64(x) (*((int64_t *) DATAPTR(x)))
 
 #define R_IS_NULL(x) ((*(LGBM_SE)(x)).sxpinfo.type == 0)
 
--- a/src/lightgbm_R.cpp
+++ b/src/lightgbm_R.cpp
@@ -466,7 +466,7 @@ LGBM_SE LGBM_BoosterGetNumPredict_R(LGBM_SE handle,
   R_API_BEGIN();
   int64_t len;
   CHECK_CALL(LGBM_BoosterGetNumPredict(R_GET_PTR(handle), R_AS_INT(data_idx), &len));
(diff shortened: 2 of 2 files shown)


guolinke:
@Laurae2 As these data are all pointers, is that safe to use int64_t for them?

Laurae2:
@guolinke Not sure whether it is safe for all of them.

R automatically (in most cases) changes the type of integer to numeric when it overflows. As we are modifying directly the backend, we might want to check for 64-bit length data.

Some R specifics:

* nrow and ncol should remain int32 for maximum compatibility
* nrow * ncol can be int64 for maximum compatibility, except for dgCMatrix (CSC)

I have code ready in C++ to handle big sparse matrices (64-bit length, 64-bit rows, 64-bit columns for extremely large data), however I noticed it costs a lot of memory. Do you think for dgCMatrix (CSC) it is enough to support only 64-bit length and 32-bit rows/columns?

guolinke:
okay, I see. So I think maybe it is better to keep these types (for rows, cols) `int32`, to avoid the unsafe pointer. 
If there are future demand for int64, we can figure out how to support them. 

In short, I think this PR is ready to merge for now. What do you think ?

guolinke:
ping @Laurae2

Laurae2:
@guolinke It should be good for merging.

Laurae2:
@guolinke Ideally, we can let @nbenn test whether it solves #1884 before merging.

guolinke:
@Laurae2 Since there is no response now, I will merge this first. If there are unsolved problems, we can fix them in another PRs.

PR #1964, opened 2019-01-24
[tests][python] added test for huge string model

### LINKED ISSUE #1598: Error training model: Negative size passed to PyBytes_FromStringAndSize

AWS EC2 Instance : r5.24xlarge(vcpu: 96 , Memory: 768 GB)

Train Shape: 89M rows, 224 features

## Error message
  File "Model_run.py", line 336, in <module>
    lgbmodel = lgb.train(params, train_lgb)
 File "/usr/local/lib/python3.5/dist-packages/lightgbm/engine.py", line 228, in train
    booster._load_model_from_string(booster._save_model_to_string(), False)
  File "/usr/local/lib/python3.5/dist-packages/lightgbm/basic.py", line 1719, in _save_model_to_string
    return string_buffer.value.decode()
SystemError: Negative size passed to PyBytes_FromStringAndSize

## Parameters
params = {
		'objective': 'mape',
		'metric': 'mape',
		'boosting': 'gbdt',
		'learning_rate': 0.012,
		'verbose': 0,
		'num_leaves': 150000,
		'bagging_freq': 0,
		'min_data_in_leaf': 300,
		'max_bin': 255,
		'max_depth': 28,
		'num_rounds': 200,
		'min_gain_to_split': 0.0,
		'save_binary': True
	}

Python3: Model is generated successfully with nrounds=100.

PS: The same parameters (with nrounds=200) successfully yield a model object in R.

StrikerRUS:
Closed #1598.

### DIFF
--- a/tests/python_package_test/test_engine.py
+++ b/tests/python_package_test/test_engine.py
@@ -3,6 +3,7 @@
 import copy
 import math
 import os
+import psutil
 import unittest
 
 import lightgbm as lgb
@@ -828,3 +829,20 @@ class TestEngine(unittest.TestCase):
         results = lgb.cv(params, dataset, num_boost_round=10, fpreproc=preprocess_data)
         self.assertIn('multi_logloss-mean', results)
         self.assertEqual(len(results['multi_logloss-mean']), 10)
+
+    @unittest.skipIf(psutil.virtual_memory().free / 1024 / 1024 / 1024 < 3, 'not enough RAM')
+    def test_model_size(self):
+        X, y = load_boston(True)
+        data = lgb.Dataset(X, y)
+        bst = lgb.train({'verbose': -1}, data, num_boost_round=2)
+        y_pred = bst.predict(X)
+        model_str = bst.model_to_string()
+        one_tree = model_str[model_str.find('Tree=1'):model_str.find('end of trees')].replace('Tree=1',
+                                                                                              'Tree={}')
+        begin, sep, end = model_str.rpartition('end of trees')
+        multiplier = int(2**31 / len(one_tree)) + 1
+        new_model_str = begin + (one_tree * multiplier).format(*range(2, multiplier + 2)) + sep + end
+        self.assertGreater(len(new_model_str), 2**31)
+        bst.model_from_string(new_model_str, verbose=False)
+        y_pred_new = bst.predict(X)
+        np.testing.assert_allclose(y_pred, y_pred_new)
--- a/tests/python_package_test/test_engine.py
+++ b/tests/python_package_test/test_engine.py
@@ -1222,13 +1222,17 @@ class TestEngine(unittest.TestCase):
         one_tree = one_tree.replace('Tree=1', 'Tree={}')
         multiplier = 100
         total_trees = multiplier + 2
-        new_model_str = (model_str[:model_str.find('tree_sizes')]
-                         + '\n' * (2**31 - one_tree_size * total_trees)
-                         + model_str[model_str.find('Tree=0'):model_str.find('end of trees')]
-                         + (one_tree * multiplier).format(*range(2, total_trees))
-                         + model_str[model_str.find('end of trees'):])
-        self.assertGreater(len(new_model_str), 2**31)
-        bst.model_from_string(new_model_str, verbose=False)
-        self.assertEqual(bst.num_trees(), total_trees)
-        y_pred_new = bst.predict(X, num_iteration=2)
-        np.testing.assert_allclose(y_pred, y_pred_new)
+        try:
+            new_model_str = (model_str[:model_str.find('tree_sizes')]
+                             + '\n\n'
+                             + model_str[model_str.find('Tree=0'):model_str.find('end of trees')]
+                             + (one_tree * multiplier).format(*range(2, total_trees))
+                             + model_str[model_str.find('end of trees'):]
+                             + ' ' * (2**31 - one_tree_size * total_trees))
+            self.assertGreater(len(new_model_str), 2**31)
+            bst.model_from_string(new_model_str, verbose=False)
+            self.assertEqual(bst.num_trees(), total_trees)
+            y_pred_new = bst.predict(X, num_iteration=2)
+            np.testing.assert_allclose(y_pred, y_pred_new)
+        except MemoryError:
+            self.skipTest('not enough RAM')
--- a/tests/python_package_test/test_engine.py
+++ b/tests/python_package_test/test_engine.py
@@ -840,14 +840,10 @@ class TestEngine(unittest.TestCase):
         one_tree = model_str[model_str.find('Tree=1'):model_str.find('end of trees')]
         one_tree_size = len(one_tree)
         one_tree = one_tree.replace('Tree=1', 'Tree={}')
-        multiplier = int(2**31 / (one_tree_size + len(str(one_tree_size)) + 1))
+        multiplier = 100
         total_trees = multiplier + 2
-        tree_sizes = model_str[model_str.find('tree_sizes'):model_str.find('Tree=0')].strip()
         new_model_str = (model_str[:model_str.find('tree_sizes')]
-                         + tree_sizes
-                         + ' '
-                         + ' '.join(str(one_tree_size + len(str(i)) - 1) for i in range(2, total_trees))
-                         + '\n\n'
+                         + '\n' * (2**31 - one_tree_size * total_trees)
                          + model_str[model_str.find('Tree=0'):model_str.find('end of trees')]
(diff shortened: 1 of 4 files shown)


tests/python_package_test/test_engine.py
```diff
@@ -828,3 +829,20 @@ def preprocess_data(dtrain, dtest, params):
         results = lgb.cv(params, dataset, num_boost_round=10, fpreproc=preprocess_data)
         self.assertIn('multi_logloss-mean', results)
         self.assertEqual(len(results['multi_logloss-mean']), 10)
+
+    @unittest.skipIf(psutil.virtual_memory().free / 1024 / 1024 / 1024 < 3, 'not enough RAM')
```
StrikerRUS: To be honest, I don't know whether it's cross-platform or not.
On my Windows with Python 3.6 it shows approximately 2 GB.


Or maybe it's better to wrap it in
```
try:
   …
except MemoryError:
   pass
```

guolinke: It seems some tests throw memory error..

StrikerRUS: Yeah!..
I'll add `try/except`, but I don't know should I bump free memory requirement too (and to what value) or not. What do you think?

guolinke: @StrikerRUS I think we can keep them both, can use a memory size larger than 4GB and smaller than 8GB, such as 7GB?

StrikerRUS: @guolinke OK. I agree with this. One more question: did you mean 7GB of **free** RAM or overall?

StrikerRUS: https://github.com/Microsoft/LightGBM/blob/cbae0c1b1681dedd6a519759c13964f4d3ee7e35/tests/python_package_test/test_engine.py#L833
Unfortunately, it caused even more MemoryErrors. Will try with 4 GB of free volume.

StrikerRUS: There is no difference in results for 4, 5, 6 GB. Only requirement of 7 GB of free memory makes this test to be skipped which results in green CI checks.
https://github.com/Microsoft/LightGBM/blob/e72ee0335cb5c05462ce3dddbe517d1eb4447bc8/tests/python_package_test/test_engine.py#L833
Seems that this `skipIf` is not useful actually...

StrikerRUS:
It turned out that `tree_sizes=` field should be updated too.

StrikerRUS:
```
len(new_model_str) - 2**31
>>> 7105873
```
I think it's too much and it's caused by that calculation of trees number is performed for `Tree=1`.

tests/python_package_test/test_engine.py
```diff
@@ -828,3 +829,30 @@ def preprocess_data(dtrain, dtest, params):
...
+    def test_model_size(self):
+        X, y = load_boston(True)
+        data = lgb.Dataset(X, y)
+        bst = lgb.train({'verbose': -1}, data, num_boost_round=2)
+        y_pred = bst.predict(X)
+        model_str = bst.model_to_string()
+        one_tree = model_str[model_str.find('Tree=1'):model_str.find('end of trees')]
+        one_tree_size = len(one_tree)
+        one_tree = one_tree.replace('Tree=1', 'Tree={}')
+        multiplier = int(2**31 / (one_tree_size + len(str(one_tree_size)) + 1))
```
StrikerRUS: @guolinke Now test fails not with `MemoryError`, but with `ci/test.sh: line 120:  4299 Killed  pytest $BUILD_DIRECTORY/tests`.
Should I add a magic number here to take into account the growing size of tree(due to Tree=1,...100,...,10000) and decrease the difference between actual model size and desired? Or is any other easy way to calculate the needed number of trees?

guolinke: actually, the tree_size field could be deleted directly. This will slow down the model loading, but I think it is okay for test. 
BTW, I think the simplest solution is padding many `\r\n`, as it will increase the length of model string and will be ignored by model parsing.

StrikerRUS: Oh, `tree_sizes` can be simply removed, it's great!

Now, some fixed little number of trees (100) is appended to the model and then `\n`-padding is used to reach 2**31 total model size. Unfortunately, it causes weird memory consumption behaviour during the model loading process, please see the screenshot below:

image

Compare with simply huge number of trees without `\n`-padding behaviour (`multiplier=1370000` instead of 100, and the total model size is even bigger compared with `\n`-padding):

image

guolinke: That is weird, it should be in this loop...
https://github.com/Microsoft/LightGBM/blob/b37065dbd5510079aac341a34895f4854c782fcb/src/boosting/gbdt_model_text.cpp#L442-L459

StrikerRUS: I tried to append '\n's at the model's end - got the same weird behavior.

guolinke: will this cause memory error in local ?

StrikerRUS: Nope! On my home machine it consumed all 32GB in last peak and then released all excess memory; after that model was loaded successfully.

Actually, `\n` padding at the end is inappropriate action because we have the following code, which checks only the last and one before last lines for pandas section:
https://github.com/Microsoft/LightGBM/blob/b96849918d802f750657b3dc3d17324e070993a2/python-package/lightgbm/basic.py#L318-L323

Also, I tried to move `cur_line` under the `if` statement, but it didn't help.
https://github.com/Microsoft/LightGBM/blob/b96849918d802f750657b3dc3d17324e070993a2/src/boosting/gbdt_model_text.cpp#L445-L446

StrikerRUS: I tried to use spaces instead of new lines. The situation became much better.

```diff
index 049c682..b06ce16 100644
--- a/tests/python_package_test/test_engine.py
+++ b/tests/python_package_test/test_engine.py
@@ -843,7 +843,7 @@ class TestEngine(unittest.TestCase):
         multiplier = 100
         total_trees = multiplier + 2
         new_model_str = (model_str[:model_str.find('tree_sizes')]
-                         + '\n' * (2**31 - one_tree_size * total_trees)
+                         + '\n\n\n' + ' ' * (2**31 - one_tree_size * total_trees)
                          + model_str[model_str.find('Tree=0'):model_str.find('end of trees')]
                          + (one_tree * multiplier).format(*range(2, total_trees))
                          + model_str[model_str.find('end of trees'):])
```

image

But surprisingly it loaded 101 trees instead of 102...

After that I tried to move whitespaces to model's end: everything became even much better and the correct number of trees was loaded.

```diff
index 049c682..31743b8 100644
--- a/tests/python_package_test/test_engine.py
+++ b/tests/python_package_test/test_engine.py
@@ -843,10 +843,11 @@ class TestEngine(unittest.TestCase):
         multiplier = 100
         total_trees = multiplier + 2
         new_model_str = (model_str[:model_str.find('tree_sizes')]
-                         + '\n' * (2**31 - one_tree_size * total_trees)
+                         + '\n\n'
                          + model_str[model_str.find('Tree=0'):model_str.find('end of trees')]
                          + (one_tree * multiplier).format(*range(2, total_trees))
-                         + model_str[model_str.find('end of trees'):])
+                         + model_str[model_str.find('end of trees'):]
+                         + ' ' * (2**31 - one_tree_size * total_trees))
         self.assertGreater(len(new_model_str), 2**31)
         bst.model_from_string(new_model_str, verbose=False)
         self.assertEqual(bst.num_trees(), total_trees)
```

image

guolinke: great! can this pass the test now?

StrikerRUS: Haven't tried yet.

BTW, seems that this line is the root cause of weird memory consumption.
https://github.com/Microsoft/LightGBM/blob/b96849918d802f750657b3dc3d17324e070993a2/python-package/lightgbm/basic.py#L318


image
image

Don't you know why spaces in the middle of the model cause wrong number of trees are being loaded?
> But surprisingly it loaded 101 trees instead of 102...

guolinke: because it will break here:
https://github.com/Microsoft/LightGBM/blob/b37065dbd5510079aac341a34895f4854c782fcb/src/boosting/gbdt_model_text.cpp#L455
maybe we should remove this break in cpp side.

StrikerRUS: Oh, got it! Thanks!

> maybe we should remove this break in cpp side.

Not sure that it's actually needed. Unexpected whitespaces in a model file are not something that can happen during save/load use cases, and manual editing is not something that we're supporting. That question was just my curiosity.

PR #1973, opened 2019-01-27
fix nan in eval results

### LINKED ISSUE #1971: TypeError: 'NoneType' object is not iterable
[fail.zip](https://github.com/Microsoft/LightGBM/files/2799905/fail.zip)

Code was compiled for GPU.  Uses hash: d169ac8b7049d3e8345e70288f19eaa7dfb8fe38

Issue has been there for many months, but we noticed it happening for non-trivial data as well so now posting issue with example.

guolinke:
fix #1971

### DIFF
--- a/python-package/lightgbm/callback.py
+++ b/python-package/lightgbm/callback.py
@@ -213,7 +213,7 @@ def early_stopping(stopping_rounds, verbose=True):
             return
         for i in range_(len(env.evaluation_result_list)):
             score = env.evaluation_result_list[i][2]
-            if cmp_op[i](score, best_score[i]):
+            if best_score_list[i] is None or cmp_op[i](score, best_score[i]):
                 best_score[i] = score
                 best_iter[i] = env.iteration
                 best_score_list[i] = env.evaluation_result_list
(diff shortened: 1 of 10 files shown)


StrikerRUS:
maybe add a regression test?

StrikerRUS:
I've added `assert`s for certainty. Because I remember we had a test in which `cv` run successfully but there was rubbish in result's content (it was cv for lambdarank, if I\m not mistaken).

python-package/lightgbm/callback.py
```diff
@@ -213,7 +213,7 @@ def _callback(env):
             return
         for i in range_(len(env.evaluation_result_list)):
             score = env.evaluation_result_list[i][2]
-            if cmp_op[i](score, best_score[i]):
+            if best_score_list[i] is None or cmp_op[i](score, best_score[i]):
```
StrikerRUS: Is this additional condition for the first iter?
https://github.com/Microsoft/LightGBM/blob/fe77eac5e7c95b46085bc78a527809493c9046d6/python-package/lightgbm/callback.py#L201

guolinke: this is to forcely push the score of the first iter.

PR #1975, opened 2019-01-27
[python] added OpenMP options for python-package installation

StrikerRUS:
Refer to #1898, #1919, #1923.

This PR adds two new installation options for Apple Clang users: `--openmp-include-dir` and `--openmp-library`.

#1919 fixed compilation with Apple Clang on Mojave for only one particular case: Homebrew default paths. This PR allows to pass custom paths via these options.

Also, this PR tries to use default MacPorts paths in case of installation failure with Homebrew ones.

Unfortunately, I don't have access to Mojave, so I'm gently asking @henry0312 to check this. And please do not merge until this.

### DIFF
--- a/python-package/setup.py
+++ b/python-package/setup.py
@@ -90,7 +90,8 @@ def silent_call(cmd, raise_error=False, error_msg=''):
 def compile_cpp(use_mingw=False, use_gpu=False, use_mpi=False, nomp=False,
                 use_hdfs=False, boost_root=None, boost_dir=None,
                 boost_include_dir=None, boost_librarydir=None,
-                opencl_include_dir=None, opencl_library=None):
+                opencl_include_dir=None, opencl_library=None,
+                openmp_include_dir=None, openmp_library=None):
 
     if os.path.exists(os.path.join(CURRENT_DIR, "build_cpp")):
         shutil.rmtree(os.path.join(CURRENT_DIR, "build_cpp"))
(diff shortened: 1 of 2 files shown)


henry0312:
Oh, what do you ask me to check? build and install?

StrikerRUS:
@henry0312 Yep! Please, try 3 scenarios on Mojave:
1. simple `python setup.py install` - Homebrew path guesses should be used;
2. `python setup.py install --openmp-include-dir=/usr/local/opt/libomp/include/ --openmp-library=/usr/local/opt/libomp/lib/libomp.dylib` - specified paths should be used;
3. `python setup.py install --openmp-include-dir=/any/bad/path/ --openmp-library=/another/any/bad/path/libomp.dylib` - Homebrew fallback should be used.

StrikerRUS:
BTW, is there any other popular package managers on macOS whose paths we can hardcode?

henry0312:
Ok, I'll try the three cases.

> BTW, are there any other popular package managers on macOS whose paths we can hardcode?

MacPorts uses `/opt/local` as the default prefix.

python-package/setup.py
```diff
@@ -177,9 +165,41 @@ def compile_cpp(use_mingw=False, use_gpu=False, use_mpi=False, nomp=False,
...
+                clear_path(os.path.join(CURRENT_DIR, "build_cpp"))
+                status = silent_call(cmake_cmd + get_cmake_opts('/usr/local/opt/libomp/include',
+                                                                '/usr/local/opt/libomp/lib/libomp.dylib'))
+                status += silent_call(["make", "_lightgbm", "-j4"])
+            if status != 0:
+                logger.warning("Compilation failed.")
+                logger.info("Starting to compile with MacPorts OpenMP paths guesses.")
+                clear_path(os.path.join(CURRENT_DIR, "build_cpp"))
+                silent_call(cmake_cmd + get_cmake_opts('/opt/local/include/libomp',
+                                                       '/opt/local/lib/libomp/libomp.dylib'),
```
StrikerRUS: @henry0312 
 >  MacPorts uses `/opt/local` as the default prefix.

Here it is.

henry0312:
Sorry, I don't have much time to check the scenarios at home, only where I can use Mojave.
Please wait until this weekend.

StrikerRUS:
@henry0312 Sure! Since we have hotfix for Homebrew, this's not urgent.

chivee:
@StrikerRUS , I have tested it on Mojave. and it works well. However, we can wait @henry0312 for robustness.

StrikerRUS:
Thanks a lot @chivee ! Sounds good! And I agree about extra check.

StrikerRUS:
Now we can test Mojave at CI side: https://blog.travis-ci.com/2019-02-12-xcode-10-2-beta-2-is-now-available.

I've set `osx_image` to `xcode10.2` and made `export COMPILER="clang"`. Result: https://travis-ci.org/Microsoft/LightGBM. `sdist` and `bdist` are fine; `regular` expectedly fails without @henry0312's fix.

PR #1979, opened 2019-01-30
[python] refined pandas categorical field loading

StrikerRUS:
This PR was inspired by the following: https://github.com/Microsoft/LightGBM/pull/1964#pullrequestreview-196488992.

The current solution is quite wasteful, because it loads entire file or duplicates the model in a list entries while we need only one line actually. In consequence, this function becomes a bottleneck during the loading big models.


chivee:
codes look good to me. @StrikerRUS , do you have any test result of time usage for loading ?

PR #2024, opened 2019-02-20
[docs] added notes about params usage when data is provided via path and removed unused param

### LINKED ISSUE #2023: weight parameter not working
Good day,

I am interesting to use the weight param to set my weight for my objective function. I believe this weight give me the option to set individual weight to each row and this will in turn leads to higher weight == higher contribution in the terms of loss. 

Based on the documentation: 
image

I passed this as a dictionary params: (just a simple example). My weight column is the first column. 

```python 
params = {
        'nthread' : 6,
        'learning_rate': 0.01,
        'num_leaves':80,
        'max_depth': 7,
        'weight': 0}

```

and later used it in train. 

After training, when i check the feature importance and find that the column weight is used as a feature. Does this mean that the weight parameter is not taken into account? It will be good if you can help 

 Edit: is this weight_column suppose to be in .dataset and not in param dictionary?

StrikerRUS:
Closed #2023.

Usage example:

```
train = lgb.Dataset(r'../regression/regression.train')
lgb.train({'objective': 'regression', 'weight_column': 5, 'group_column': 7}, train)
```

What note should be written for [`save_binary`](https://lightgbm.readthedocs.io/en/latest/Parameters.html#save_binary) and [`enable_load_from_binary_file`](https://lightgbm.readthedocs.io/en/latest/Parameters.html#enable_load_from_binary_file)?

### DIFF
--- a/docs/Parameters.rst
+++ b/docs/Parameters.rst
@@ -513,11 +513,7 @@ IO Parameters
 
    -  if ``true``, LightGBM will save the dataset (including validation data) to a binary file. This speed ups the data loading for the next time
 
--  ``enable_load_from_binary_file`` :raw-html:`<a id="enable_load_from_binary_file" title="Permalink to this parameter" href="#enable_load_from_binary_file">&#x1F517;&#xFE0E;</a>`, default = ``true``, type = bool, aliases: ``load_from_binary_file``, ``binary_load``, ``load_binary``
-
-   -  set this to ``true`` to enable autoloading from previous saved binary datasets
-
-   -  set this to ``false`` to ignore binary datasets
+   -  **Note**: can be used only in CLI version; for language-specific packages you can use the correspondent function
 
 -  ``header`` :raw-html:`<a id="header" title="Permalink to this parameter" href="#header">&#x1F517;&#xFE0E;</a>`, default = ``false``, type = bool, aliases: ``has_header``
 
--- a/docs/Parameters.rst
+++ b/docs/Parameters.rst
@@ -451,7 +451,7 @@ IO Parameters
 
    -  if ``""``, will use ``train_data_file`` + ``.init`` (if exists)
 
-   -  **Note**: can be used only in CLI version
+   -  **Note**: works only in case of loading data directly from file
 
 -  ``valid_data_initscores`` :raw-html:`<a id="valid_data_initscores" title="Permalink to this parameter" href="#valid_data_initscores">&#x1F517;&#xFE0E;</a>`, default = ``""``, type = string, aliases: ``valid_data_init_scores``, ``valid_init_score_file``, ``valid_init_score``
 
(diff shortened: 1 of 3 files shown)


StrikerRUS:
Usage example:

```
train = lgb.Dataset(r'../regression/regression.train')
lgb.train({'objective': 'regression', 'weight_column': 5, 'group_column': 7}, train)
```

StrikerRUS:
Hmmm, according to the log
```
[LightGBM] [Info] Using column number 5 as weight
[LightGBM] [Info] Using column number 7 as group/query id
---->  [LightGBM] [Info] Loading initial scores...  <----
[LightGBM] [Info] Loading query weights...
[LightGBM] [Info] Total Bins 5632
[LightGBM] [Info] Number of data: 7000, number of used features: 26
```
it seems to me that the **Note** for [`initscore_filename`](https://lightgbm.readthedocs.io/en/latest/Parameters.html#initscore_filename) and [`valid_data_initscores`](https://lightgbm.readthedocs.io/en/latest/Parameters.html#valid_data_initscores) should be fixed from `Note: can be used only in CLI version` to `Note: works only in case of loading data directly from file`. Am I right?

guolinke:
Yeah you are right

StrikerRUS:
@guolinke Thanks for the clarification!

And what about `*binary*` params?
> What note should be written for [`save_binary`](https://lightgbm.readthedocs.io/en/latest/Parameters.html#save_binary) and [`enable_load_from_binary_file`](https://lightgbm.readthedocs.io/en/latest/Parameters.html#enable_load_from_binary_file)?

guolinke:
I think they are `Note: works only in case of loading data directly from file`

StrikerRUS:
Shouldn't the note for `save_binary` be `Note: can be used only in CLI version; for language-specific packages you can use the correspondent function`?

https://github.com/Microsoft/LightGBM/blob/90127b526be8b13bdf840effacde9e61fa765a6f/src/application/application.cpp#L113-L116
https://github.com/Microsoft/LightGBM/blob/462612b42c6fe7bb1be1ccd8b72d77021b9f2d38/src/c_api.cpp#L808-L814
https://github.com/Microsoft/LightGBM/blob/b37065dbd5510079aac341a34895f4854c782fcb/R-package/R/lgb.Dataset.R#L567-L577

And `enable_load_from_binary_file` seems to be unused at all: https://github.com/Microsoft/LightGBM/search?q=enable_load_from_binary_file&unscoped_q=enable_load_from_binary_file

What am I missing?..

guolinke:
sorry, you are right about save_binary.
as for enable_load_from_binary_file, I think it is a deprecated parameter.

StrikerRUS:
Thank you for the confirmation!

> as for enable_load_from_binary_file, I think it is a deprecated parameter.

Should it be removed then?

guolinke:
@StrikerRUS 
Yeah, they can be removed.

StrikerRUS:
Done!

PR #2025, opened 2019-02-25
add min_acceptable_gain for better early stopping

robwec:
I run Optuna to optimize hyperparameters and often notice that it spends a long time "optimizing" for small loss gains, e.g. +/-0.0001. Therefore, it would make sense to stop early when the loss hasn't improved by at least some amount for N rounds. However, when I try to create a custom early stop callback imitating lightgbm's own early stop callback, it doesn't seem to play well with the exception. Therefore, editing the package source made the most sense. This is a useful change. I can run grid search faster.

### DIFF
--- a/python-package/callback.py
+++ b/python-package/callback.py
@@ -1,0 +1,232 @@
+# coding: utf-8
+# pylint: disable = invalid-name, W0105, C0301
+"""Callbacks library."""
+from __future__ import absolute_import
+
+import collections
+import warnings
+from operator import gt, lt
+
+from .compat import range_
+
+
+class EarlyStopException(Exception):
+    """Exception of early stopping."""
+
+    def __init__(self, best_iteration, best_score):
+        """Create early stopping exception.
+
+        Parameters
+        ----------
+        best_iteration : int
+            The best iteration stopped.
+        best_score : float
+            The score of the best iteration.
+        """
+        super(EarlyStopException, self).__init__()
+        self.best_iteration = best_iteration
+        self.best_score = best_score
+
+
+# Callback environment used by callbacks
+CallbackEnv = collections.namedtuple(
+    "LightGBMCallbackEnv",
+    ["model",
+     "params",
+     "iteration",
+     "begin_iteration",
+     "end_iteration",
+     "evaluation_result_list"])
+
+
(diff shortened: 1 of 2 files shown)


Laurae2:
@guolinke should this parameter be put in C++ side instead of only Python?

robwec:
Oh, also I notice I sort of didn't correct the logic for choosing the best iteration. For an early stop based on non-significantly-decreasing rate you would want to check the last few iterations and choose the one with the best result, instead of using the one from early_stopping_rounds iterations ago (which is what the normal early_stop call does), because with this kind of early stopping there will generally be a more recent iteration with the best metric, even though it will only be the best by a small margin.

Didn't make a big difference in my use case, but it does look like the kind of thing someone might complain about later.

guolinke:
ping @StrikerRUS

Laurae2:
@robwec Seeing the 744 line code modifications, this can be more easily done using a custom metric (make a closure or a function returning a function which returns the custom metric).

big possible small hack: inside the custom metric, store the best iteration value and keep a log of the best value.

If:

* Metric is worse than best: return decreasing value
* Metric is identical to best: return identical value (or slightly lower than the best intentionally)
* Metric is better than best: check against a threshold, if threshold is lower than the improvement, then return the unmodified metric, else return slightly worse metric than the best

This way in less than 30 lines of code you should have your automated early stopping. You can store the original metric values elsewhere on the fly.

---

Note: there is already an engine.py and callback.py. You should modify the original files instead of re-adding these files elsewhere in the Python package.

guolinke:
I am confused about the file location, should them be in `python-package/lightgbm` ?

StrikerRUS:
@guolinke Yeah!
@robwec Please move files back to `LightGBM/python-package/lightgbm/` folder and make your changes there.

robwec:
Ok, I will be closing this pull request now. I am able to use this on my own. (The reason I had to edit the source was when I tried a custom callback with Optuna it failed on the EarlyStoppingException. No idea why that happened, but I don't need help with that anymore.)

After staring at walls of numbers within 0.001 of each other on grid search / Optuna I thought of this type of callback.

I hope the idea is useful :)

PR #2032, opened 2019-02-27
Added additional APIs to better support JNI on Spark

eisber:
Performance improvement for prediction is roughly 3-4x on our datasets.
Training data loading was also improved by avoiding memory copy. Performance measurements are still in progress.


src/c_api.cpp
```diff
@@ -22,6 +22,7 @@
 #include <stdexcept>
 #include <mutex>
 #include <functional>
+// #include <chrono>
```
imatiach-msft: remove (?)

src/c_api.cpp
```diff
@@ -1175,6 +1285,35 @@ int LGBM_BoosterPredictForCSR(BoosterHandle handle,
...
 
+int LGBM_BoosterPredictForCSRSingleRow(BoosterHandle handle,
+                              const void* indptr,
+                              int indptr_type,
+                              const int32_t* indices,
+                              const void* data,
+                              int data_type,
+                              int64_t nindptr,
+                              int64_t nelem,
+                              int64_t,
```
imatiach-msft: what is the variable here?

imatiach-msft: (int64_t,)

imatiach-msft: looks like this is num col, shouldn't it be used here (or removed in signature otherwise)

swig/lightgbmlib.i
```diff
@@ -14,18 +14,152 @@
...
+                                     int64_t* out_len,
+                                     double* out_result) {
+    // Alternatives
+    // - GetIntArrayElements: performs copy
+    // - GetDirectBufferAddress: fails on wrapped array
+    // Some words of warning for GetPrimitiveArrayCritical
+    // https://stackoverflow.com/questions/23258357/whats-the-trade-off-between-using-getprimitivearraycritical-and-getprimitivety
+    
+    jboolean isCopy;
+    int* indices0 = (int*)jenv->GetPrimitiveArrayCritical(indices, &isCopy);
```
imatiach-msft: this is very cool 👍

swig/lightgbmlib.i
```diff
@@ -14,18 +14,152 @@
...
+  };
+  
+  int LGBM_DatasetCreateFromCSRDirect(JNIEnv *jenv,
+                                     jobjectArray arrayOfSparseVector,
+                                     int num_rows,
+                                     int64_t num_col,
+                                     const char* parameters,
+                                     const DatasetHandle reference,
+                                     DatasetHandle* out) {
+   jclass sparseVectorClass = jenv->FindClass("org/apache/spark/ml/linalg/SparseVector");
```
imatiach-msft: seems a bit strange to have spark datastructures leak into lightgbm.  Might be good to have something like Spark in the name of the method to make it clear that this isn't even just for Java but an optimization for spark in particular.

src/c_api.cpp
eisber: good point. I copied this from the function above. you want to remove it there too?

swig/lightgbmlib.i
eisber: fair point

src/c_api.cpp
```diff
@@ -197,6 +197,41 @@ class Booster {
...
+        is_predict_leaf = true;
+      } else if (predict_type == C_API_PREDICT_RAW_SCORE) {
+        is_raw_score = true;
+      } else if (predict_type == C_API_PREDICT_CONTRIB) {
+        predict_contrib = true;
+      } else {
+        is_raw_score = false;
+      }
+
+        // TODO: config could be optimized away... (maybe using lambda callback?)
```
imatiach-msft: nit: align comment with code

src/c_api.cpp
imatiach-msft: no, if it's copied I guess we should keep it then

src/c_api.cpp
```diff
@@ -1175,6 +1285,35 @@ int LGBM_BoosterPredictForCSR(BoosterHandle handle,
...
+                              int64_t nindptr,
+                              int64_t nelem,
+                              int64_t,
+                              int predict_type,
+                              int num_iteration,
+                              const char* parameter,
+                              int64_t* out_len,
+                              double* out_result) {
+  API_BEGIN();
+  auto param = Config::Str2Map(parameter);
```
imatiach-msft: some of this code looks very similar to below, would be nice to refactor to a single common function

eisber: I agree it looks very much the same, but then again it's mostly boilerplate code that's replicated in other functions too (e.g. LGBM_BoosterPredictForFile). Maybe packaging up the Config creation and translation to OMP would be better. Different PR?

imatiach-msft: ok, ignore my comment then, this is fine.

guolinke:
@StrikerRUS I think we can use these apis to improve the prediction in python package as well

PR #2041, opened 2019-03-08
[python] added get_split_value_histogram method

### LINKED ISSUE #2013: get_split_value_histogram XGBoost equivalent in LightGBM
I want histogram/values of all the splits happening in a particular feature. Just like get_split_value_histogram in XGBoost. Is there any way to get it currently? 

I am aware of plot_tree and create_tree_digraph methods. But those are for visualizations

StrikerRUS:
Closed #2013.

Calculate feature split histogram in XGBoost-style (https://github.com/dmlc/xgboost/blob/master/python-package/xgboost/core.py#L1694-L1737).

**UPD:** … and in standard (numpy) style.

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -2428,8 +2428,8 @@ class Booster(object):
         else:
             return result
 
-    def get_split_value_histogram(self, feature, bins=None):
-        """Get split value histogram of the specified feature.
+    def get_split_value_histogram(self, feature, bins=None, xgboost_style=False):
+        """Get split value histogram for the specified feature.
 
         Parameters
         ----------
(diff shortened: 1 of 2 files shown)


StrikerRUS:
Is docstring OK? I tried to make it informative and simple at the same time, but I'm still not satisfied 😕

guolinke:
@StrikerRUS it looks good to me

StrikerRUS:
OK, thanks! Can we merge then or second review is needed?

guolinke:
@StrikerRUS I think it is ready for merge

StrikerRUS:
@guolinke I've found that it crashes with **categorical feature**. What do you think we can do with it? Simply raise error for cat. feature or is there any smarter solution?

guolinke:
@StrikerRUS 
can we use a string type directly for the x-axis in histogram for categorical features?

StrikerRUS:
`numpy.histogram()` crashes with string type. Also, how to sort cat. string values, let say `['0', '3', '2', '0||3']`?

guolinke:
@StrikerRUS yeah, the combination is a problem. Let us skip the categorical features for now.

StrikerRUS:
@guolinke OK. Created #2045 for this.

PR #2064, opened 2019-03-22
addressed cpplint error about C-style cast

StrikerRUS:
Refer to #1990 and http://google.github.io/styleguide/cppguide.html#Casting.

Please review **attentively**, because I'm not a cpp coder.

- Is `static_cast<label_t*>(nullptr)` OK?

- What can be done with `R_object_helper.h`? E.g.

    https://github.com/Microsoft/LightGBM/blob/7a89d0054bcf0db9f96af3c53a078b6aa3152efe/include/LightGBM/R_object_helper.h#L121-L125

- How to rewrite `(void*)`? E.g.
   https://github.com/Microsoft/LightGBM/blob/7a89d0054bcf0db9f96af3c53a078b6aa3152efe/src/io/dataset.cpp#L808

### DIFF
--- a/include/LightGBM/tree.h
+++ b/include/LightGBM/tree.h
@@ -259,7 +259,7 @@ class Tree {
       }
       int_fval = 0;
     }
-    int cat_idx = int(threshold_[node]);
+    int cat_idx = static_cast<int>(threshold_[node]);
     if (Common::FindInBitset(cat_threshold_.data() + cat_boundaries_[cat_idx],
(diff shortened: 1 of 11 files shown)


guolinke:
thanks @StrikerRUS 
there are not good solutions for these casts. I think we can live with them for now.

StrikerRUS:
@guolinke Thank you, got it! I'll try to find the to ignore them in `cpplint`.

PR #2066, opened 2019-03-22
reworked includes in source files

StrikerRUS:
- Added missed includes (fixed `build/include_what_you_use` error).
- Reordered includes (fixed `build/include_order` error).

According to the [guide](http://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes) the order should be:
```
dir2/foo2.h.
A blank line
C system files.
C++ system files.
A blank line
Other libraries' .h files.
Your project's .h files.
```

But due to [the bug](https://github.com/cpplint/cpplint/issues/43) only the following order doesn't produce errors:
```
dir2/foo2.h.
A blank line
Your project's .h files in include folder.
A blank line
C system files.
C++ system files.
A blank line
Other libraries' .h files.
Your project's .h files in src folder.
```

IMHO, any order is OK in case it's the same across the whole project.

Refer to #1990, #1585, #1586 and http://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes.


src/treelearner/gpu_tree_learner.cpp
```diff
@@ -1,15 +1,14 @@
...
-#include <LightGBM/network.h>
 #include <LightGBM/bin.h>
+#include <LightGBM/network.h>
+#include <LightGBM/utils/array_args.h>
 
 #include <algorithm>
-#include <vector>
 
+#include "../io/dense_bin.hpp"
+#include "../io/dense_nbits_bin.hpp"
```
StrikerRUS: I don't know how to bring this into consistency with the guide:
> All of a project's header files should be listed as descendants of the project's source directory without use of UNIX directory shortcuts . (the current directory) or .. (the parent directory).

guolinke: I think it is okay to include the hpp file by this way.
ping @huanzhang12

StrikerRUS: @guolinke Can this be merged without these fixes because this PR will produce many conflicts (too many files touched)? And later we'll create another PR for `..` in case @huanzhang12 will have good ideas.

guolinke: Sure

PR #2078, opened 2019-04-03
[docs] Python wrapper doesn't support params in form of list of pairs

StrikerRUS:
`params` is only supported in form of dict.
Also, fixed some typos.

PR #2079, opened 2019-04-05
[ci] update CI stuff

StrikerRUS:
- Use our Ubuntu 14.04 Docker for GPU test on Azure Pipelines.
- Update CMake and Clang in Docker to new stable versions.
- Install `libomp-dev` inside Docker instead of via the setup script for Clang.
- Update MinGW-w64 on Appveyor.
- Update Boost library.
- Update Boost submodule.

Also, this PR resolves the issue about GPU test described in https://github.com/Microsoft/LightGBM/pull/2078#issuecomment-479605231

### DIFF
--- a/.vsts-ci.yml
+++ b/.vsts-ci.yml
@@ -10,7 +10,7 @@ jobs:
 - job: Linux
 ###########################################
   variables:
-    COMPILER: gcc
+    COMPILER: clang
   pool:
     vmImage: 'ubuntu-16.04'
   container: ubuntu1404
@@ -63,7 +63,7 @@ jobs:
 - job: MacOS
 ###########################################
   variables:
-    COMPILER: clang
+    COMPILER: gcc
   pool:
     vmImage: 'macOS-10.13'
   strategy:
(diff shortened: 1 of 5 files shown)


StrikerRUS:
Inverted compilers (Clang instead gcc and gcc instead Clang) work fine too:
3ecf8965ec66c7a62cf3f7d97e902c4f261e8c0d

image

.vsts-ci.yml
```diff
@@ -4,7 +4,7 @@ variables:
 resources:
   containers:
   - container: ubuntu1404
-    image: lightgbm/vsts-agent:ubuntu-14.04
+    image: lightgbm/vsts-agent:ubuntu-14.04-dev
```
StrikerRUS: This is to test and show that everything works.
I'll remove `dev` postfix before merging.

guolinke:
@StrikerRUS this is to fix current fails of CI?

StrikerRUS:
@guolinke Yeah

PR #1956, opened 2019-01-18
[python][R][docs] added possibility to install with Visual Studio 2019 Preview

StrikerRUS:
Append new toolset and generator name instead of prepend due to that this is **Preview** release.

CMake `3.14` will support VS 2019 Preview. Also, they dropped the possibility to specify architecture inside generator name (`Win64`)
https://gitlab.kitware.com/cmake/cmake/merge_requests/2789

Tested on Appveyor `Visual Studio 2019 Preview` image with [this fix](https://github.com/appveyor/ci/issues/2776#issuecomment-448016098) and [nightly CMake](https://cmake.org/files/dev/cmake-3.13.20190116-gf8d457-win64-x64.msi) through [RDP](https://www.appveyor.com/docs/how-to/rdp-to-build-worker/).

### Manual installation: ✔️

<details>
 <summary><b>Logs</b> (clickable)</summary>

```
$C:\LightGBM>cmake --version
cmake version 3.13.20190116-gf8d457

CMake suite maintained and supported by Kitware (kitware.com/cmake).

$C:\LightGBM>set path=C:\Program Files (x86)\Microsoft Visual Studio\2019\Preview\MSBuild\Current\Bin;%path%

$C:\LightGBM>msbuild --version
Microsoft (R) Build Engine version 16.0.218-preview+g5c625fa747 for .NET Framework
Copyright (C) Microsoft Corporation. All rights reserved.

MSBUILD : error MSB1001: Unknown switch.
Switch: --version

For switch syntax, type "MSBuild /help"

$C:\LightGBM>cd build

$C:\LightGBM\build>cmake -DCMAKE_GENERATOR_PLATFORM=x64 -G "Visual Studio 16 2019 Win64" ..
CMake Error: Could not create named generator Visual Studio 16 2019 Win64

Generators
  Visual Studio 16 2019        = Generates Visual Studio 2019 project files.
                                 Use -A option to specify architecture.
  Visual Studio 15 2017 [arch] = Generates Visual Studio 2017 project files.
                                 Optional [arch] can be "Win64" or "ARM".
  Visual Studio 14 2015 [arch] = Generates Visual Studio 2015 project files.
                                 Optional [arch] can be "Win64" or "ARM".
  Visual Studio 12 2013 [arch] = Generates Visual Studio 2013 project files.
                                 Optional [arch] can be "Win64" or "ARM".
  Visual Studio 11 2012 [arch] = Generates Visual Studio 2012 project files.
[...]

PR #2083, opened 2019-04-07
[python][R][docs][ci] better compatibility with Visual Studio 2019

StrikerRUS:
This PR is related to #1956.

[VS 2019](https://visualstudio.microsoft.com/downloads/
) has been released recently and the latest [CMake 3.14.1](https://cmake.org/cmake/help/v3.14/generator/Visual%20Studio%2016%202019.html) supports it. Therefore we can reorder generators (and Platform Toolsets) to try to build LightGBM with VS 2019 firstly.

Due to that [CMake dropped the possibility to specify architecture inside generator name](https://gitlab.kitware.com/cmake/cmake/merge_requests/2789) (`Win64`), we can set it via [`-A` option](https://cmake.org/cmake/help/v3.14/generator/Visual%20Studio%2016%202019.html#platform-selection), which was presented in CMake since 3.1 and we require >=3.8 for Windows (it's shorter and better than setting `CMAKE_GENERATOR_PLATFORM` directly and allows to handle all generators in the same way).

I'm doubt about only one thing: whether we need to specify [`-T host=` option](https://cmake.org/cmake/help/v3.14/variable/CMAKE_GENERATOR_TOOLSET.html#variable:CMAKE_GENERATOR_TOOLSET) or not. Its default behavior differs in VS 2019 compared to all other versions:
2019:
> [By default this generator uses the 64-bit variant on x64 hosts and the 32-bit variant otherwise.](https://cmake.org/cmake/help/v3.14/generator/Visual%20Studio%2016%202019.html#toolset-selection)

2017:
> [By default this generator uses the 32-bit variant even on a 64-bit host.](https://cmake.org/cmake/help/v3.14/generator/Visual%20Studio%2015%202017.html#toolset-selection)

This option controls the following:
```
-- Check for working CXX compiler: C:/Program Files (x86)/Microsoft Visual Studio/2017/Enterprise/VC/Tools/MSVC/14.16.27023/bin/Hostx86/x64/cl.exe -- works
                                                                                                                               |------|
                                                                                                                                  ^
                                                                                                                                  |
                                                                                                                                  v
                                                                                                                               |------|
-- Check for working CXX compiler: C:/Program Files (x86)/Microsoft Visual Studio/2017/Enterprise/VC/Tools/MSVC/14.16.27023/bin/Hostx64/x64/cl.exe -- works
```
Refer to https://gitlab.kitware.com/cmake/cmake/issues/17451#note_341418

Remaining things:
- ~add a link to MSVC Redistributable 2019 [here](https://github.com/Microsoft/LightGBM/tree/master/R-package#prerequisites); it's still unavailable from [MS support site](https://support.microsoft.com/en-us/help/2977003/the-latest-supported-visual-c-downloads)~ Done!;
- ~add a bullet for matching Boost binaries with VS 2019 [here](https://github.com/Microsoft/LightGBM/blob/master/docs/Installation-Guide.rst#windows-3); latest 1.70.0b1 [doesn't contain](https://sourceforge.net/projects/boost/files/boost-binaries/1.70.0_b1/) 14.2 version.~ **UPD:** ... But 1.70 does: https://sourceforge.net/projects/boost/files/boost-binaries/1.70.0/.


StrikerRUS:
Here is a couple of links and quotes for anyone who sees `-T host=` option for the first time, just like me:
- https://gitlab.kitware.com/cmake/cmake/issues/15622,
> * If I open `VS2015 x64 Native Tools Command Prompt` then `where cl.exe` produces `C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\bin\amd64\cl.exe`.
> * If I open `VS2015 x86 x64 Cross Tools Command Prompt` then `where cl.exe` produces `C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\bin\x86_amd64\cl.exe`.

- https://gitlab.kitware.com/cmake/cmake/merge_requests/155
> Milestone 
3.8.0 

means that we may safely use this option.

- https://docs.microsoft.com/en-us/cpp/build/how-to-enable-a-64-bit-visual-cpp-toolset-on-the-command-line?view=vs-2019
> The 32-bit and 64-bit tools generate identical code, but the 64-bit tools support more memory for precompiled header symbols and the Whole Program Optimization (/GL and /LTCG) options.

henry0312:
I don't know much about Windows.

StrikerRUS:
@Laurae2 Link to MSVC Redistributable 2019 has been added.

StrikerRUS:
@guolinke @Laurae2 Thank you for review!

What're your opinions about `-T host=`? Do we need to suggest this flag for users or/and use it at CI tests?

guolinke:
@StrikerRUS 
It was the `hostx86/x64/cl.exe` before this PR:
image

It seems there are no differences in the generated codes, from this https://docs.microsoft.com/en-us/cpp/build/how-to-enable-a-64-bit-visual-cpp-toolset-on-the-command-line?view=vs-2019. 
So I think the hostx86 is enough for now.

StrikerRUS:
@guolinke Yeah, you're absolutely right! It's all about compiler performance (speed, ability to compile huge projects, etc.). The code itself should be identical.

I agree, let's leave 32bit compilers for now.

StrikerRUS:
OK, I think it's ready to merge.

Issue #1105, opened 2017-12-06
Suggestion: Add training_data metrics in cv return 


CliveHugo:
## Environment info
Operating System: WIN10
CPU: 
C++/Python/R version: python3.6

a suggestion : 
Add training_data performance(metrics) in cv return for drawing validation curve (just as Xgboost.cv's return) since skleran.validation_curve is very inefficient (run n time , n =number of iteration). Don't know if anyone else has this need though.

------------------------------------------------------------------
Example from return of Xgboost.cv (after` pandas.from_dict):

Out[23]: 
    test-auc-mean  test-auc-std  train-auc-mean  train-auc-std
0        0.847553      0.011909        0.849978       0.000601
1        0.852323      0.012572        0.856189       0.000791
2        0.856945      0.011464        0.863983       0.000770
3        0.857423      0.011400        0.869057       0.000768
4        0.857955      0.010694        0.872355       0.000579
5        0.858113      0.011444        0.874590       0.000896
6        0.857701      0.011529        0.876404       0.000870
`

jcrosskey:
This would be a nice feature to have, it'll help monitor whether overfit occurred judging from cv.

wjmj:
It will be a very useful feature, I hope it will be implemented as soon as possible.

StrikerRUS:
Closed via #2089.

PR #2089, opened 2019-04-09
[python] add flag of displaying train loss for lgb.cv()

matsuken92:
Adding flag on lgb.cv() for displaying train loss during training iteration.


StrikerRUS:
Closed #1105.
Glad to see how very old issues are closing! 😃

.idea/LightGBM.iml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
henry0312: Please don't include `.idea`.
The directory isn't needed for LightGBM, but for IntelliJ IDEA.

matsuken92: Oh... these files was included mistakenly. So deleted. Could you re-check? Thank you!

python-package/lightgbm/engine.py
```diff
@@ -408,6 +415,8 @@ def cv(params, train_set, num_boost_round=100,
     callbacks : list of callables or None, optional (default=None)
         List of callback functions that are applied at each iteration.
         See Callbacks in Python API for more information.
+    disp_train_loss : bool, (default=False)
```
StrikerRUS: I think that `show_train_loss` will be better. There are already `show_stdv` param, for instance.
`optional` word is missed after the comma.

matsuken92: Thank you for your advice. Revised the argument name and the comment of show_train_loss.

tests/python_package_test/test_engine.py
```diff
@@ -464,6 +464,17 @@ def test_cv(self):
...
+
+        # enable display training loss
+        cv_res = lgb.cv(params_with_metric, lgb_train, num_boost_round=10,
+                        nfold=3, stratified=False, shuffle=False,
+                        metrics='l1', verbose_eval=False, show_train_loss=True)
+        self.assertIn('train l1-mean', cv_res)
+        self.assertIn('valid l1-mean', cv_res)
+        self.assertNotIn('train l2-mean', cv_res)
+        self.assertNotIn('valid l2-mean', cv_res)
+        self.assertEqual(len(cv_res['valid l1-mean']), 10)
```
StrikerRUS: @matsuken92 Please add `train` length check here too.

StrikerRUS:
I began to doubt in new parameter name. Seems that I was wrong when suggested to rename `disp_*` -> `show_*`. Both these words describe some appearance. There are even the following remark in `show_stdv` description:
> Results are not affected by this parameter, and always contain std.

In contrast, new parameter affects performance because requires additional metric computation for train dataset.
I think we need a better word. What about calc(ulate), comp(ute), eval(uate)? I personally prefer eval.

BTW, `loss` should be replaced with `metric` for the clarity too.

python-package/lightgbm/engine.py
```diff
@@ -408,6 +415,8 @@ def cv(params, train_set, num_boost_round=100,
     callbacks : list of callables or None, optional (default=None)
         List of callback functions that are applied at each iteration.
         See Callbacks in Python API for more information.
+    disp_train_loss : bool, optional　(default=False)
```
henry0312: disp_train_loss --> show_train_loss

henry0312: I agree to https://github.com/Microsoft/LightGBM/pull/2089#issuecomment-481918964.
I think`eval_train_loss` is better than `show_train_loss`, because the prameter will actually evaluate (calculate) train loss, not show (display) it.

StrikerRUS: @henry0312 I think it should be `metric`, not `loss`. WDYT?

tests/python_package_test/test_engine.py
matsuken92: Added one more test item for `train` case.

python-package/lightgbm/engine.py
matsuken92: Changed the aug name show_train_loss -> eval_train_metric. I also think 'metric' is match in this case.

python-package/lightgbm/engine.py
```diff
@@ -408,6 +415,9 @@ def cv(params, train_set, num_boost_round=100,
     callbacks : list of callables or None, optional (default=None)
         List of callback functions that are applied at each iteration.
         See Callbacks in Python API for more information.
+    eval_train_metric : bool, optional　(default=False)
```
StrikerRUS: Replace the char (non-breaking space?) after `optional` with normal space.

python-package/lightgbm/engine.py
```diff
@@ -408,6 +415,9 @@ def cv(params, train_set, num_boost_round=100,
...
+        Whether to display the train metric in progress. The score of the metric is calculated
```
StrikerRUS: Start the sentence `The score of ...` from new line.

python-package/lightgbm/engine.py
matsuken92: Oh, it is mistake. .Fixed to Half-width space.

python-package/lightgbm/engine.py
matsuken92: Add line break just before `The score of ...`.

henry0312:
I'm sorry that I didn't have much time because of moving.
I'll check today.

PR #2098, opened 2019-04-12
[python] make possibility to create Booster from string official

### LINKED ISSUE #2097: Cannot instantiate model from a string or buffer
I would like to load a LightGBM model (for prediction only) from a string or buffer rather than a file on disk.

It seems that there is a method called `model_from_string` [documentation link][1] but ... it produces an error, which seemingly defeats the purpose of the method as I understand it.

```
import boto3
import lightgbm as lgb
import io

model_path = 'some/path/here'
s3_bucket = boto3.resource('s3').Bucket('some-bucket')

obj = s3_bucket.Object(model_path)
buf = io.BytesIO()
try:
    obj.download_fileobj(buf)
except Exception as e:
    raise e
else:
    model = lgb.Booster().model_from_string(buf.read().decode("UTF-8"))
```
which produces the following error....

```
TypeError: Need at least one training dataset or model file to create booster instance
```

Alternatively, I thought that I might be able to use the regular loading method 
```
lgb.Booster(model_file=buf.read().decode("UTF-8"))
``` 
... but this also doesn't work. 

```
FileNotFoundError: [Errno 2] No such file or directory: ''
```

Now, I realize that I can create a workaround by writing the buffer to disk, and then reading it. However, this feels very redundant and inefficient.

Thus, my question is, how can instantiate a model to use for prediction without pointing to a an actual file on disk?

If there is no such functionality, can this please be implemented / fixed? 

  [1]: https://lightgbm.readthedocs.io/en/latest/Python-API.html#lightgbm.Booster.model_from_string

StrikerRUS:
Fixed #2097.

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -397,7 +397,7 @@ class _InnerPredictor(object):
             self.num_total_iteration = out_num_iterations.value
             self.pandas_categorical = None
         else:
-            raise TypeError('Need Model file or Booster handle to create a predictor')
+            raise TypeError('Need model_file or booster_handle to create a predictor')
 
         pred_parameter = {} if pred_parameter is None else pred_parameter
         self.pred_parameter = param_dict_to_str(pred_parameter)
@@ -1578,7 +1578,7 @@ class Dataset(object):
 class Booster(object):
     """Booster in LightGBM."""
 
-    def __init__(self, params=None, train_set=None, model_file=None, silent=False):
+    def __init__(self, params=None, train_set=None, model_file=None, model_str=None, silent=False):
         """Initialize the Booster.
 
         Parameters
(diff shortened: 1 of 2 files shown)


python-package/lightgbm/basic.py
```diff
@@ -1578,7 +1578,7 @@ def dump_text(self, filename):
 class Booster(object):
     """Booster in LightGBM."""
 
-    def __init__(self, params=None, train_set=None, model_file=None, silent=False):
+    def __init__(self, params=None, train_set=None, model_file=None, model_str=None, silent=False):
```
StrikerRUS: Yeah, I know that for the backward compatibility new param should be the last one. But it's too illogical to split `train_set=None, model_file=None, model_str=None` params (especially `model_file=None, model_str=None`), in which at least one should be specified by a user.

PR #2099, opened 2019-04-12
minor fixes in CMakeLists

StrikerRUS:


CMakeLists.txt
```diff
@@ -82,15 +82,15 @@ if(USE_OPENMP)
     SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}")
 else()
     # Ignore unknown #pragma warning
-    if( (CMAKE_CXX_COMPILER_ID MATCHES "[cC][lL][aA][nN][gG]")
-      OR (CMAKE_CXX_COMPILER_ID MATCHES "[gG][nN][uU]"))
+    if((CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+      OR (CMAKE_CXX_COMPILER_ID STREQUAL "GNU"))
```
StrikerRUS: CMake uses "GNU" and "Clang" starting from `2.6.0` (maybe earlier) version inside itself:
https://github.com/Kitware/CMake/blob/a00f1beb530c0c9fff8deab150a07b4c3839285b/Modules/CMakeDetermineCCompiler.cmake#L130-L139
https://github.com/Kitware/CMake/blob/a00f1beb530c0c9fff8deab150a07b4c3839285b/Modules/CMakeDetermineCXXCompiler.cmake#L140-L149
And this variable is well-documented starting from version `2.8.10`:
https://cmake.org/cmake/help/v2.8.10/cmake.html#variable:CMAKE_LANG_COMPILER_ID
The oldest CMake version supported by LightGBM is `2.8`.

The only case for using these regexes can be user-defined compilers, but I think it's not the case we are going to support.
https://stackoverflow.com/a/40653688

PR #2102, opened 2019-04-12
fixed cpplint errors about spaces and newlines

StrikerRUS:
Refer to #1990.

### DIFF
--- a/helpers/parameter_generator.py
+++ b/helpers/parameter_generator.py
@@ -306,7 +306,7 @@ def gen_parameter_code(config_hpp, config_out_cpp):
     # tails
     str_to_write += "  return str_buf.str();\n"
     str_to_write += "}\n\n"
-    str_to_write += "}\n"
+    str_to_write += "}  // namespace LightGBM\n"
     with open(config_out_cpp, "w") as config_out_cpp_file:
         config_out_cpp_file.write(str_to_write)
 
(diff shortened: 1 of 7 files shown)


helpers/parameter_generator.py
```diff
@@ -306,7 +306,7 @@ def gen_parameter_code(config_hpp, config_out_cpp):
     # tails
     str_to_write += "  return str_buf.str();\n"
     str_to_write += "}\n\n"
-    str_to_write += "}\n"
+    str_to_write += "}  // namespace LightGBM\n"
```
StrikerRUS: namespace end

include/LightGBM/utils/common.h
```diff
@@ -810,20 +810,20 @@ inline static void ObtainMinMaxSum(const T1 *w, int nw, T1 *mi, T1 *ma, T2 *su)
   }
 }
 
-inline static std::vector<uint32_t> EmptyBitset(int n){
+inline static std::vector<uint32_t> EmptyBitset(int n) {
   int size = n / 32;
-  if(n % 32 != 0) size++;
+  if (n % 32 != 0) ++size;
```
StrikerRUS: prefix increment

src/c_api.cpp
```diff
@@ -216,9 +216,9 @@ class Booster {
         is_raw_score = false;
       }
 
-      // TODO: config could be optimized away... (maybe using lambda callback?)
+      // TODO(eisber): config could be optimized away... (maybe using lambda callback?)
```
StrikerRUS: nickname of TODO creator

PR #2104, opened 2019-04-13
[docs] Updating Linux Pre-req & Installation Instructions

viveksnh:
- Added some additional pre requisites for a successful installation on ubuntu
- Added clarity to the installation code after git clone regarding how to use cmake

### DIFF
--- a/R-package/README.md
+++ b/R-package/README.md
@@ -26,6 +26,13 @@ For users who wants to install online with GPU or want to choose a specific comp
 
 You can perform installation either with **Apple Clang** or **gcc**. In case you prefer **Apple Clang**, you should install **OpenMP** (details for installation can be found in [Installation Guide](https://github.com/Microsoft/LightGBM/blob/master/docs/Installation-Guide.rst#apple-clang)) first and **CMake** version 3.12 or higher is required. In case you prefer **gcc**, you need to install it (details for installation can be found in [Installation Guide](https://github.com/Microsoft/LightGBM/blob/master/docs/Installation-Guide.rst#gcc)) and set some environment variables to tell R to use `gcc` and `g++`. If you install these from Homebrew, your versions of `g++` and `gcc` are most likely in `/usr/local/bin`, as shown below.
 
+#### Linux OS Preparation
+1. Install CMake `sudo apt-get install cmake`
+2. Install curl `sudo apt-get install libcurl4-gnutls-dev` 
+3. Install openssl `sudo apt-get install openssl-dev`
+4. Install xml2 `sudo apt-get install libxml2-dev`
+5. Install R devtools `sudo -i R` --> `install.packages("devtools", dependencies = TRUE)`
+
 ```
 # replace 8 with version of gcc installed on your machine
 export CXX=/usr/local/bin/g++-8 CC=/usr/local/bin/gcc-8
@@ -37,7 +44,10 @@ Build and install R-package with the following commands:
 
 ```sh
 git clone --recursive https://github.com/Microsoft/LightGBM
-cd LightGBM
+cd LightGBM; 
+mkdir build ; cd build
+cmake ..
+make -j4
 Rscript build_r.R
 ```
 


R-package/README.md
```diff
@@ -26,6 +26,13 @@ For users who wants to install online with GPU or want to choose a specific comp
 
 You can perform installation either with **Apple Clang** or **gcc**. In case you prefer **Apple Clang**, you should install **OpenMP** (details for installation can be found in [Installation Guide](https://github.com/Microsoft/LightGBM/blob/master/docs/Installation-Guide.rst#apple-clang)) first and **CMake** version 3.12 or higher is required. In case you prefer **gcc**, you need to install it (details for installation can be found in [Installation Guide](https://github.com/Microsoft/LightGBM/blob/master/docs/Installation-Guide.rst#gcc)) and set some environment variables to tell R to use `gcc` and `g++`. If you install these from Homebrew, your versions of `g++` and `gcc` are most likely in `/usr/local/bin`, as shown below.
 
+#### Linux OS Preparation
+1. Install CMake `sudo apt-get install cmake`
+2. Install curl `sudo apt-get install libcurl4-gnutls-dev`
```
StrikerRUS: Just out of curiosity, why `gnutls` and not `openssl`?

R-package/README.md
```diff
@@ -26,6 +26,13 @@ For users who wants to install online with GPU or want to choose a specific comp
 
 You can perform installation either with **Apple Clang** or **gcc**. In case you prefer **Apple Clang**, you should install **OpenMP** (details for installation can be found in [Installation Guide](https://github.com/Microsoft/LightGBM/blob/master/docs/Installation-Guide.rst#apple-clang)) first and **CMake** version 3.12 or higher is required. In case you prefer **gcc**, you need to install it (details for installation can be found in [Installation Guide](https://github.com/Microsoft/LightGBM/blob/master/docs/Installation-Guide.rst#gcc)) and set some environment variables to tell R to use `gcc` and `g++`. If you install these from Homebrew, your versions of `g++` and `gcc` are most likely in `/usr/local/bin`, as shown below.
 
+#### Linux OS Preparation
```
Laurae2: The steps are not for Linux OS but are for Ubuntu in your case. It will not work on other Linux OSes.

R-package/README.md
```diff
@@ -26,6 +26,13 @@ For users who wants to install online with GPU or want to choose a specific comp
...
+1. Install CMake `sudo apt-get install cmake`
```
Laurae2: It is Ubuntu specific, it will not work on other OSes (SUSE, Red Hat, etc.).

R-package/README.md
Laurae2: It is Ubuntu specific, it will not work on other OSes (SUSE, Red Hat, etc.).

R-package/README.md
```diff
@@ -26,6 +26,13 @@ For users who wants to install online with GPU or want to choose a specific comp
...
+2. Install curl `sudo apt-get install libcurl4-gnutls-dev` 
+3. Install openssl `sudo apt-get install openssl-dev`
```
Laurae2: It is Ubuntu specific, it will not work on other OSes (SUSE, Red Hat, etc.).

R-package/README.md
```diff
@@ -26,6 +26,13 @@ For users who wants to install online with GPU or want to choose a specific comp
...
+4. Install xml2 `sudo apt-get install libxml2-dev`
```
Laurae2: It is Ubuntu specific, it will not work on other OSes (SUSE, Red Hat, etc.).

R-package/README.md
```diff
@@ -26,6 +26,13 @@ For users who wants to install online with GPU or want to choose a specific comp
...
+5. Install R devtools `sudo -i R` --> `install.packages("devtools", dependencies = TRUE)`
```
Laurae2: R users are assumed to have devtools already installed. The steps you mentioned are better to be put directly in the devtools readme: https://cran.r-project.org/web/packages/devtools/readme/README.html

Also, `sudo` is not always needed. It will depend on the user use-case and OS/R setup.

R-package/README.md
```diff
@@ -37,7 +44,10 @@ Build and install R-package with the following commands:
 
 ```sh
 git clone --recursive https://github.com/Microsoft/LightGBM
-cd LightGBM
+cd LightGBM; 
+mkdir build ; cd build
```
Laurae2: not needed for R

R-package/README.md
```diff
@@ -37,7 +44,10 @@ Build and install R-package with the following commands:
...
+cmake ..
```
Laurae2: not needed for R

R-package/README.md
```diff
@@ -37,7 +44,10 @@ Build and install R-package with the following commands:
...
+make -j4
```
Laurae2: not needed for R

Laurae2:
Weird, GitHub duplicated comments as the lines added.

R-package/README.md
Laurae2: In addition, libcurl4 should be the one which was used to compile R. Otherwise, non-static compilation of libraries may break (especially for users of RStudio Server).

R-package/README.md
jameslamb: Agree with @Laurae2 , `apt` package manager isn't going to exist on all OSes

Laurae2:
Closing this in favor of #2123.

PR #2110, opened 2019-04-16
regression_metric.QuantileMetric loss bug

amniskin:
The quantile loss function had a bug where instead of returning (1-alpha) * diff it was returning (alpha - 1) * diff. See https://github.com/scikit-learn/scikit-learn/blob/7b136e9/sklearn/ensemble/gradient_boosting.py#L709 for an outside reference.

I also changed the comment on the line above (not really a bug fix, but it scratches my OCD itch)

### DIFF
--- a/src/metric/regression_metric.hpp
+++ b/src/metric/regression_metric.hpp
@@ -148,7 +148,7 @@ class L2Metric: public RegressionMetric<L2Metric> {
   }
 };
 
-/*! \brief L2 loss for regression task */
+/*! \brief Quantile loss for regression task */
 class QuantileMetric : public RegressionMetric<QuantileMetric> {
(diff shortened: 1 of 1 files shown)


amniskin:
That was dumb. My bad. diff is signed.

src/metric/regression_metric.hpp
```diff
@@ -148,7 +148,7 @@ class L2Metric: public RegressionMetric<L2Metric> {
   }
 };
 
-/*! \brief L2 loss for regression task */
+/*! \brief Quantile loss for regression task */
```
StrikerRUS: Hi @amniskin !

Can you please create a PR for this typo? Thanks!

Issue #2107, opened 2019-04-16
How can I get the right probability for unbalanced binary classification problem?


jiayouanan:
The parameter **scale_pos_weight** is used to adjust the imbalance between positive and negative samples. But I don't know how to judge the test data sets by predicting probabilities? Judging by 0.5?

XGB can also handle unbalanced data sets, [which](https://xgboost.readthedocs.io/en/latest/tutorials/param_tuning.html) is described as:

> **Handle Imbalanced Dataset**
For common cases such as ads clickthrough log, the dataset is extremely imbalanced. This can affect the training of XGBoost model, and there are two ways to improve it.
If you care only about the overall performance metric (AUC) of your prediction
Balance the positive and negative weights via scale_pos_weight
Use AUC for evaluation
**If you care about predicting the right probability**
In such a case, you cannot re-balance the dataset
Set parameter max_delta_step to a finite number (say 1) to help convergence

I am concerned that the meaning of LGBM's **scale_pos_weight** is the same as XGB?
That is to say,  for an unbalanced data set in LGBM, if I only care about overall performance metric (AUC) of the test set, I use **scale_pos_weight** is appropriate. But if I want to really know the right probability of prediction, then should I **not use** scale_pos_weight?

Thank you for your great help.

guolinke:
@jiayouanan it is the same as XGB. 
@StrikerRUS I think we should add these descriptions in our documents as well.
`is_unbalance`, `class_weight` are also affected.

StrikerRUS:
@guolinke You mean, add a note about that these params will spoil probability?

guolinke:
@StrikerRUS yeah, exactly.

StrikerRUS:
Hmm, autoclosing GitHub feature didn't work again.
Closed in #2113.

PR #2113, opened 2019-04-17
[docs] added note about the spoiled probabilities

StrikerRUS:
What about `sample_weight`/`weight_column` params?


guolinke:
The weights over samples indeed will affect the probabilities as well. However, it depends on the user's purpose. If the samples indeed have different weights, then the probabilities are right; if the weights are for imbalance purpose, users should aware of this.

StrikerRUS:
@guolinke Thank you for the clarification! Then I think it's ready for review.

Closed #2107.

PR #2115, opened 2019-04-18
[python] ignore pandas ordered categorical columns by default 

### LINKED ISSUE #2108: Are SKLEARN API estimators can distinguish PANDAS nominal type feature and PANDAS ordinal type feature?
Are SKLEARN API estimators can recognize difference between
'ordinal feature (applying pandas.api.types.CategoricalDtype(data, ordered=True))'
and
'nominal feature (applying pandas.api.types.CategoricalDtype(data, ordered=False))' ?

thank you for your help.

StrikerRUS:
Closed #2108.

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -258,7 +258,8 @@ def _data_from_pandas(data, feature_name, categorical_feature, pandas_categorica
             raise ValueError('Input data must be 2 dimensional and non empty.')
         if feature_name == 'auto' or feature_name is None:
             data = data.rename(columns=str)
-        cat_cols = data.select_dtypes(include=['category']).columns
+        cat_cols = list(data.select_dtypes(include=['category']).columns)
+        cat_cols_not_ordered = [col for col in cat_cols if not data[col].cat.ordered]
         if pandas_categorical is None:  # train dataset
             pandas_categorical = [list(data[col].cat.categories) for col in cat_cols]
         else:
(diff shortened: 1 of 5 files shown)


python-package/lightgbm/basic.py
```diff
@@ -267,26 +268,25 @@ def _data_from_pandas(data, feature_name, categorical_feature, pandas_categorica
...
             data[cat_cols] = data[cat_cols].apply(lambda x: x.cat.codes).replace({-1: np.nan})
         if categorical_feature is not None:
             if feature_name is None:
                 feature_name = list(data.columns)
             if categorical_feature == 'auto':
-                categorical_feature = list(cat_cols)
+                categorical_feature = cat_cols_not_ordered
             else:
-                categorical_feature = list(categorical_feature) + list(cat_cols)
+                categorical_feature = list(categorical_feature) + cat_cols_not_ordered
```
StrikerRUS: I wonder whether this is a good decision to add default categorical columns even in case user explicitly specify categorical features. Firstly, this feature is not documented. Secondly, it seems that this approach makes some discomfort for users: the only way to disable default cat features is to convert them in source dataframe. I think it's more comfortable just not include them in list of cat features. The intuition behind this is that user want **either** use default cat features in cat columns, **or** specify cat features in param.

guolinke: good point, this indeed is not comfortable.

StrikerRUS: @guolinke Cool! I'll address this in next PR, OK?

guolinke: yeah, sure

PR #2124, opened 2019-04-21
[WIP] added SWIG wrappers to build steps

imatiach-msft:
adding SWIG wrapper to build steps based on:
https://github.com/Microsoft/LightGBM/issues/1945
helpful stack overflow for Java install:
https://stackoverflow.com/questions/31196567/installing-java-in-docker-image
note this is currently untested

PR #2125, opened 2019-04-22
[ci] added job for swig compilation on Linux with artifacts

### LINKED ISSUE #1945: lightGBM 2.2.200 still has glibc issue /lib64/libstdc++.so.6: version : GLIBCXX_3.4.20' not found
Thank you @xiaoxiaofengzi for the clarification!
Unfortunately this is a known problem caused by that MMLSpark uses old and broken (in terms of dependencies) version of LightGBM.

You can track this [PR](https://github.com/Azure/mmlspark/pull/391) and these issues [1](https://github.com/Azure/mmlspark/issues/335), [2](https://github.com/Azure/mmlspark/issues/390) to know when the problem will be solved on the Spark-package side. I think that you can copy-paste your log there to hurry them gently.
From the our side starting from `2.2.2` version we automatically guarantee `GLIBC <= 2.14` and `GLIBCXX <= 3.4.19`:

https://github.com/Microsoft/LightGBM/blob/a694712b7fb86cd532eea2c1781b58d4ba58436a/helpers/check_dynamic_dependencies.py#L18-L23
https://github.com/Microsoft/LightGBM/blob/a694712b7fb86cd532eea2c1781b58d4ba58436a/helpers/check_dynamic_dependencies.py#L25-L31

Thanks for your patience and sorry for the inconvenience!

_Originally posted by @StrikerRUS in https://github.com/Microsoft/LightGBM/issues/1858#issuecomment-440610638_

StrikerRUS:
Closed #1945, closed #2124.

Compilation of SWIG wrapper with Clang was tested too: 3ee90d2bab2484aa896be67725ddf683f6edb2fa
image

### DIFF
--- a/.ci/dockers/ubuntu-14.04/Dockerfile
+++ b/.ci/dockers/ubuntu-14.04/Dockerfile
@@ -41,6 +41,26 @@ RUN curl -sL https://cmake.org/files/v3.14/cmake-3.14.1-Linux-x86_64.sh -o cmake
  && ./cmake.sh --prefix=/usr/local --exclude-subdir \
  && rm cmake.sh
 
+# Install Java
+RUN add-apt-repository ppa:openjdk-r/ppa -y \
+ && apt-get update \
+ && apt-get install -y --no-install-recommends \
+    openjdk-8-jdk \
+ && rm -rf /var/lib/apt/lists/*
+
+ENV JAVA_HOME=/usr/lib/jvm/java-8-openjdk-amd64/
+
+# Install SWIG
+RUN curl -sL https://downloads.sourceforge.net/project/swig/swig/swig-3.0.12/swig-3.0.12.tar.gz -o swig.tar.gz \
+ && tar -xzf swig.tar.gz \
+ && cd swig-3.0.12 \
+ && ./configure --prefix=/usr/local --without-pcre \
+ && make \
+ && make install \
+ && cd .. \
+ && rm swig.tar.gz \
+ && rm -rf swig-3.0.12
+
 # Install Miniconda
 RUN curl -sL https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -o miniconda.sh \
  && chmod +x miniconda.sh \
--- a/.ci/test.sh
+++ b/.ci/test.sh
@@ -74,9 +74,9 @@ fi
 
 if [[ $TASK == "if-else" ]]; then
     conda install -q -y -n $CONDA_ENV numpy
-    mkdir $BUILD_DIRECTORY/build && cd $BUILD_DIRECTORY/build && cmake "${CMAKE_OPTS[@]}" .. && make -j4 || exit -1
+    mkdir $BUILD_DIRECTORY/build && cd $BUILD_DIRECTORY/build && cmake "${CMAKE_OPTS[@]}" .. && make lightgbm -j4 || exit -1
     cd $BUILD_DIRECTORY/tests/cpp_test && ../../lightgbm config=train.conf convert_model_language=cpp convert_model=../../src/boosting/gbdt_prediction.cpp && ../../lightgbm config=predict.conf output_result=origin.pred || exit -1
-    cd $BUILD_DIRECTORY/build && make -j4 || exit -1
+    cd $BUILD_DIRECTORY/build && make lightgbm -j4 || exit -1
     cd $BUILD_DIRECTORY/tests/cpp_test && ../../lightgbm config=predict.conf output_result=ifelse.pred && python test.py || exit -1
     exit 0
 fi
@@ -137,7 +137,7 @@ else
     cmake "${CMAKE_OPTS[@]}" ..
 fi
 
-make -j4 || exit -1
+make _lightgbm -j4 || exit -1
 
 cd $BUILD_DIRECTORY/python-package && python setup.py install --precompile --user || exit -1
 pytest $BUILD_DIRECTORY/tests || exit -1
--- a/.ci/test.sh
+++ b/.ci/test.sh
@@ -18,6 +18,19 @@ else
     CMAKE_OPTS=()
 fi
 
+if [[ $AZURE == "true" ]] && [[ $OS_NAME == "linux" ]] && [[ $TASK == "swig" ]]; then
+    mkdir $BUILD_DIRECTORY/build && cd $BUILD_DIRECTORY/build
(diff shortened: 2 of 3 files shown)


.ci/dockers/ubuntu-14.04/Dockerfile
```diff
@@ -41,6 +41,26 @@ RUN curl -sL https://cmake.org/files/v3.14/cmake-3.14.1-Linux-x86_64.sh -o cmake
  && ./cmake.sh --prefix=/usr/local --exclude-subdir \
  && rm cmake.sh
 
+# Install Java
+RUN add-apt-repository ppa:openjdk-r/ppa -y \
```
StrikerRUS: OpenJDK-8 is available from default ppa only since Ubuntu 14.10, but we use 14.04.
https://askubuntu.com/a/944260

.ci/dockers/ubuntu-14.04/Dockerfile
```diff
@@ -41,6 +41,26 @@ RUN curl -sL https://cmake.org/files/v3.14/cmake-3.14.1-Linux-x86_64.sh -o cmake
...
+ && apt-get update \
+ && apt-get install -y --no-install-recommends \
+    openjdk-8-jdk \
+ && rm -rf /var/lib/apt/lists/*
+
+ENV JAVA_HOME=/usr/lib/jvm/java-8-openjdk-amd64/
+
+# Install SWIG
+RUN curl -sL https://downloads.sourceforge.net/project/swig/swig/swig-3.0.12/swig-3.0.12.tar.gz -o swig.tar.gz \
```
StrikerRUS: `apt-get` installs SWIG 2.0 on Ubuntu 14.04.

.ci/dockers/ubuntu-14.04/Dockerfile
```diff
@@ -41,6 +41,26 @@ RUN curl -sL https://cmake.org/files/v3.14/cmake-3.14.1-Linux-x86_64.sh -o cmake
...
+ && tar -xzf swig.tar.gz \
+ && cd swig-3.0.12 \
+ && ./configure --prefix=/usr/local --without-pcre \
```
StrikerRUS: @imatiach-msft Is it OK to build SWIG without regex? Do you need this functionality in SWIG?

.ci/test.sh
```diff
@@ -18,6 +18,19 @@ else
     CMAKE_OPTS=()
 fi
 
+if [[ $AZURE == "true" ]] && [[ $OS_NAME == "linux" ]] && [[ $TASK == "swig" ]]; then
+    mkdir $BUILD_DIRECTORY/build && cd $BUILD_DIRECTORY/build
+    cmake -DUSE_SWIG=ON "${CMAKE_OPTS[@]}" ..
+    make -j4 || exit -1
+    if [[ $COMPILER == "gcc" ]]; then
+        objdump -T $BUILD_DIRECTORY/lib_lightgbm.so > $BUILD_DIRECTORY/objdump.log || exit -1
+        objdump -T $BUILD_DIRECTORY/lib_lightgbm_swig.so >> $BUILD_DIRECTORY/objdump.log || exit -1
```
StrikerRUS: @imatiach-msft Are there any other dependencies which we may want to check, except those that are already mentioned [here](https://github.com/Microsoft/LightGBM/blob/master/helpers/check_dynamic_dependencies.py)?

.vsts-ci.yml
```diff
@@ -4,7 +4,7 @@ variables:
 resources:
   containers:
   - container: ubuntu1404
-    image: lightgbm/vsts-agent:ubuntu-14.04
+    image: lightgbm/vsts-agent:ubuntu-14.04-dev
```
StrikerRUS: I'll remove `dev` postfix before merging.

StrikerRUS:
@imatiach-msft Please review

.ci/dockers/ubuntu-14.04/Dockerfile
imatiach-msft: I noticed that too, thanks for fixing

.ci/dockers/ubuntu-14.04/Dockerfile
imatiach-msft: Actually not sure.  Let's start with this for now and validate the binaries work.

.ci/test.sh
imatiach-msft: nope, those are the only two.  you have a script to validate glibc and gomp version?  That's very cool.

.vsts-ci.yml
```diff
@@ -25,6 +25,8 @@ jobs:
       bdist:
         TASK: bdist
         PYTHON_VERSION: 3.6
+      swig:
+        TASK: swig
```
imatiach-msft: ah, so this is how you turn that 'task' flag on, I see

.vsts-ci.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     vmImage: 'ubuntu-16.04'
   container: ubuntu1404
   strategy:
-    maxParallel: 6
+    maxParallel: 7
```
imatiach-msft: what does maxParallel do?

StrikerRUS: https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=schema#maximum-parallelism

StrikerRUS:
@imatiach-msft Now `jar` file will be generated for each commit, but not PR, unfortunately, due to this restriction: https://github.com/Microsoft/LightGBM/blob/d115769c2a2ddffadc76c7b84739a47937114c77/.vsts-ci.yml#L57

You can download it from Azure Pipelines page. Navigate here: https://lightgbm-ci.visualstudio.com/lightgbm-ci/_build?definitionId=1, select commit, and click to `Artifacts` in the upper right corner. You need `PackageAssets` artifact.

image


For the latest artifacts from `master` branch it's easier to use download badge here:
https://lightgbm.readthedocs.io/en/latest/Installation-Guide.html

image

StrikerRUS:
@guolinke Please merge it. Since I cannot trigger main docker update from any branch other than `master`, CI checks will always fail.

StrikerRUS:
@guolinke Also please re-run CI for merged commit when docker will be ready. I suppose CI is faster than docker and will use outdated docker image for the first time, which will result in failing CI and artifacts' unavailability.

PR #2145, opened 2019-05-04
[ci] ignore failing Yahoo URL

StrikerRUS:
OK, seems Yahoo are not in a hurry fixing their servers. But we have to merge too many PRs.

Since I haven't found any mirrors for this dataset, let's stub corresponding URL for a while. Then we'll remove this URL from ignoring list when Yahoo will wake up.

@guolinke Are you OK with this solution?


guolinke:
@StrikerRUS yeah

PR #2152, opened 2019-05-07
[docs] updated Microsoft GitHub URL

guolinke:


guolinke:
The `Microsoft` in repo URL changed to 'microsoft', and cause this problem.

StrikerRUS:
😮 Just out of curiosity, don't you know why?

I think we need to adjust other links too. I can help a little bit later.

StrikerRUS:
Huh, seems that `Microsoft` had something like a "gold status" at Travis: all jobs started at the same time immediately after a commit. After this change jobs started to spawn with some delay after a commit and their number are limited with 10 now like for other projects.

guolinke:
@StrikerRUS I guess the paid Travis plan of Microsoft ended, for Microsoft encourages Azure DevOps now...

StrikerRUS:
@guolinke It's a pity... Now we have no unlimited jobs neither on Travis, nor on Azure 😢

StrikerRUS:
BTW, please check Slack when you have time.

StrikerRUS:
Lucky us! Seems that Travis instant unlimited jobs are back.

PR #2162, opened 2019-05-09
[ci] removed temp hotfix for Yahoo URL

StrikerRUS:
Yahoo server is alive finally!

Refer to #2145.


StrikerRUS:
@guolinke Please check Slack as well (two points about SWIG and GitHub Release).

Issue #2026, opened 2019-02-25
Installation error on R by `Could NOT find OpenMP_C`


JiaxiangBU:
I search answers from Stack Overflow (https://stackoverflow.com/questions/tagged/lightgbm) but unfortunately get the solution. Here are related solutions.

* [cannot install lightgbm in R with devtools on macOS - Stack Overflow](https://stackoverflow.com/questions/50704874/cannot-install-lightgbm-in-r-with-devtools-on-macos?r=SearchResults)
* [cannot install lightgbm in R with devtools on macOS - Stack Overflow](https://stackoverflow.com/questions/50704874/cannot-install-lightgbm-in-r-with-devtools-on-macos/51005728?r=SearchResults#51005728)

## Environment info

Operating System: macOS Mojave (10.14.1)

CPU/GPU model: CPU

C++/Python/R version: R

LightGBM version or commit hash:

## Error message

```
$ Rscript build_r.R
   checking for file ‘/Users/vija/Downloads/work/LightGBM/lightgbm_r/DESCRIPTION✔  checking for file ‘/Users/vija/Downloads/work/LightGBM/lightgbm_r/DESCRIPTION’
─  preparing ‘lightgbm’:
✔  checking DESCRIPTION meta-information ...
─  cleaning src
─  checking for LF line-endings in source and make files and shell scripts
─  checking for empty or unneeded directories
   WARNING: directory ‘lightgbm/src/compute’ is empty
─  looking to see if a ‘data/datalist’ file should be added
─  building ‘lightgbm_2.2.4.tar.gz’
   
[1] "/Users/vija/Downloads/work/LightGBM/lightgbm_2.2.4.tar.gz"
* installing to library ‘/Library/Frameworks/R.framework/Versions/3.5/Resources/library’
* installing *source* package ‘lightgbm’ ...
** libs
clang -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/usr/local/include   -fPIC  -Wall -g -O2  -c R_init.c -o R_init.o
clang -dynamiclib -Wl,-headerpad_max_install_names -undefined dynamic_lookup -single_module -multiply_defined suppress -L/Library/Frameworks/R.framework/Resources/lib -L/usr/local/lib -o lightgbm.so R_init.o -F/Library/Frameworks/R.framework/.. -framework R -Wl,-framework -Wl,CoreFoundation
[...]

PR #2164, opened 2019-05-09
mojave install r openmp fix

orenov:
This PR is related to https://github.com/microsoft/LightGBM/issues/2026

### DIFF
--- a/R-package/src/install.libs.R
+++ b/R-package/src/install.libs.R
@@ -44,6 +44,15 @@ if (!use_precompile) {
     cmake_cmd <- paste0(cmake_cmd, " -DUSE_R35=ON ")
   }
 
+  APPLE <- Sys.info()['sysname'] == 'Darwin'
+  if (APPLE) {
+    cmake_cmd <- paste(cmake_cmd, ' -DOpenMP_C_FLAGS="-Xpreprocessor -fopenmp -I$(brew --prefix libomp)/include" ')
+    cmake_cmd <- paste(cmake_cmd, ' -DOpenMP_C_LIB_NAMES="omp" ')
+    cmake_cmd <- paste(cmake_cmd, ' -DOpenMP_CXX_FLAGS="-Xpreprocessor -fopenmp -I$(brew --prefix libomp)/include" ')
+    cmake_cmd <- paste(cmake_cmd, ' -DOpenMP_CXX_LIB_NAMES="omp" ')
+    cmake_cmd <- paste(cmake_cmd, ' -DOpenMP_omp_LIBRARY="$(brew --prefix libomp)/lib/libomp.dylib" ')
+  }
+
   # Check if Windows installation (for gcc vs Visual Studio)
   if (WINDOWS) {
     if (use_mingw) {
--- a/R-package/src/install.libs.R
+++ b/R-package/src/install.libs.R
@@ -44,8 +44,7 @@ if (!use_precompile) {
     cmake_cmd <- paste0(cmake_cmd, " -DUSE_R35=ON ")
   }
 
-  APPLE <- Sys.info()['sysname'] == 'Darwin'
-  if (APPLE) {
+  if (Sys.info()['sysname'] == 'Darwin' & !(grepl('^gcc', Sys.getenv('CC', '')) & grepl('^g++'), Sys.getenv('CXX', ''))) {
     cmake_cmd <- paste(cmake_cmd, ' -DOpenMP_C_FLAGS="-Xpreprocessor -fopenmp -I$(brew --prefix libomp)/include" ')
     cmake_cmd <- paste(cmake_cmd, ' -DOpenMP_C_LIB_NAMES="omp" ')
     cmake_cmd <- paste(cmake_cmd, ' -DOpenMP_CXX_FLAGS="-Xpreprocessor -fopenmp -I$(brew --prefix libomp)/include" ')
--- a/R-package/src/install.libs.R
+++ b/R-package/src/install.libs.R
@@ -45,7 +45,7 @@ if (!use_precompile) {
   }
 
   # Could NOT find OpenMP_C on Mojave workaround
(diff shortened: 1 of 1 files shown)


Laurae2:
ping @jameslamb

StrikerRUS:
Maybe also support MacPorts for the consistency with Python installation script?
https://github.com/microsoft/LightGBM/blob/abbbbd7545e6a0549c17a182b7a59d7468cfb58e/python-package/setup.py#L192-L195

Won't it break the installation with gcc?
For Python we perform the following check:
https://github.com/microsoft/LightGBM/blob/abbbbd7545e6a0549c17a182b7a59d7468cfb58e/python-package/setup.py#L168-L170

orenov:
> Maybe also support MacPorts for the consistency with Python installation script?
> 
> https://github.com/microsoft/LightGBM/blob/abbbbd7545e6a0549c17a182b7a59d7468cfb58e/python-package/setup.py#L192-L195

I see what you mean. There are some restructuring of install Rscript required, so I propose to do it as a separate pr.
> 
> Won't it break the installation with gcc?
> For Python we perform the following check:
> 
> https://github.com/microsoft/LightGBM/blob/abbbbd7545e6a0549c17a182b7a59d7468cfb58e/python-package/setup.py#L168-L170

This one I've changed. Can't make it exactly as python `if` statement, because it seems like I cannot build LightGBM R-package without OpenMP, right?

StrikerRUS:
> I see what you mean. There are some restructuring of install Rscript required, so I propose to do it as a separate pr. (if it doesn't make any problems ofc)

Agree!
> This one I've changed. Can't make it exactly as python if statement, because it seems like I cannot build LightGBM R-package without OpenMP, right?

Thanks!

@Laurae2 @jameslamb Is `USE_OPENMP` option exposed to R-package?
https://github.com/microsoft/LightGBM/blob/abbbbd7545e6a0549c17a182b7a59d7468cfb58e/CMakeLists.txt#L10

R-package/src/install.libs.R
```diff
@@ -44,6 +44,14 @@ if (!use_precompile) {
     cmake_cmd <- paste0(cmake_cmd, " -DUSE_R35=ON ")
   }
 
+  if (Sys.info()['sysname'] == 'Darwin' & !(grepl('^gcc', Sys.getenv('CC', '')) & grepl('^g++'), Sys.getenv('CXX', ''))) {
```
jameslamb: could you please add a comment above this block explaining its intent?

Also, could you please use `&&` so we get short-circuiting? If you've never seen that before in R, try this:

```
> FALSE & stop("error")
Error: error

> FALSE && stop("error")
[1] FALSE
```

jameslamb:
@StrikerRUS `USE_OPENMP` is not currently exposed in the R package but I don't see any reason it couldn't be (similar to [how we expose `USE_R35` flag](https://github.com/microsoft/LightGBM/blob/master/R-package/src/install.libs.R#L44)).

@orenov is exposing this flag to `install.libs.R` the "restructuring of install Rscript" you are referring to?

Parts of `lightgbm` specific to the R API do use OpenMP directives (e.g. [this line](https://github.com/microsoft/LightGBM/blob/master/src/lightgbm_R.cpp#L120)), but we could still expose the flag and those directives would just be ignored at compile time, right? Sorry, I am not very familiar with OpenMP.

StrikerRUS:
@jameslamb 
>Parts of lightgbm specific to the R API do use OpenMP directives ...

I think there should be no difference compared with other cpp code...
https://github.com/microsoft/LightGBM/blob/abbbbd7545e6a0549c17a182b7a59d7468cfb58e/include/LightGBM/utils/openmp_wrapper.h#L61-L67

R-package/src/install.libs.R
orenov: Thanks for a good example.

orenov:
> @StrikerRUS `USE_OPENMP` is not currently exposed in the R package but I don't see any reason it couldn't be (similar to [how we expose `USE_R35` flag](https://github.com/microsoft/LightGBM/blob/master/R-package/src/install.libs.R#L44)).
> 
I can't figure out how a user can pass `USE_OPENMP=OFF` to R install script. If you can give me a hint about how it's done, I can easily add it. The only thing I see is just to add next lines of code which are similar to `use_gpu`.

```
use_openmp <- TRUE
...
if (!use_openmp) {
    cmake_cmd <- paste0(cmake_cmd, " -DUSE_OPENMP=OFF ")
}
....
if (Sys.info()['sysname'] == 'Darwin' && use_openmp ... )
```
The problem for me, again,  I see this confusing as I'm not aware how user during installation can pass to install scipt `use_openmp=FALSE`. Any ideas?

> @orenov is exposing this flag to `install.libs.R` the "restructuring of install Rscript" you are referring to?
> 
Not actually. So the change which was proposed by @StrikerRUS required running cmake command a few times in the worst case. Different cases are guesses of usual OpenMP location on Mac OS. It needs to keep track of cmake status command and if it's not successful rerun cmake with different parameters.

Laurae2:
@orenov You can pass and get those arguments using `commandArgs(trailingOnly=TRUE)` in R.

In the long term we might want to change to this instead of using hardcoded stuff in the installation script.

@StrikerRUS not yet. The user can modify itself the installation script when required to disable it. I think this is what we do currently for GPU.

StrikerRUS:
I think that then MacPorts support and switching to `commandArgs` are content for another PR(s).

jameslamb:
Agree with the comments from @Laurae2 and @StrikerRUS . We right now do not have the ability to pass through installation flags using `R CMD INSTALL`, and instead user has to manually change `install.libs.R`.

We'd be grateful if you open an issue to track the (separate) work of changing the install script.

StrikerRUS:
@Laurae2 @jameslamb 
Can this PR be merged?

jameslamb:
I am comfortable with, but don't want to merge without @Laurae2 's approval (since I'm not as familiar with OpenMP)

attamica:
This is a life saver. I have tried everything but they all failed until I came across this change this PR suggests...Finally got the package installed successfully.

StrikerRUS:
ping @Laurae2

PR #2170, opened 2019-05-13
[ci] added swig compilation on macOS and Windows with artifacts

StrikerRUS:
- Compile SWIG wrapper under the `sdist` job to speed up overall CI run (`sdist` was chosen because we run less tests in it);
- add SWIG compilation on macOS and Windows;
- enhance `condition` field to fix warnings about empty artifacts folder: 
  image

Azul Zulu OpenJDK is used to compile SWIG wrapper because it's:
- pre-installed on Azure Hosted Agents,
- free to use,
- updated regularly.

Blog post: https://azure.microsoft.com/en-us/blog/microsoft-and-azul-systems-bring-free-java-lts-support-to-azure/

`JAVA_HOME_8_X64` variable was taken from Azure image generation script:
- https://github.com/microsoft/azure-pipelines-image-generation/blob/34c24f3acd807b203a6e2c9eddde337631f75535/images/win/scripts/Installers/Install-JavaTools.ps1#L52;
- https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/tool/java-tool-installer?view=azure-devops.


StrikerRUS:
ping @imatiach-msft

StrikerRUS:
@imatiach-msft I haven't included merging script into pipeline because there is no possibly to remove already published artifacts from previous jobs. I mean, if I downloaded JARs from Linux, Windows and macOS and then merged them into one final JAR and publish it, all OS-specific JARs are still in `PackageAssets` artifact. This enlarges overall artifact size significantly.

PR #2171, opened 2019-05-13
[ci] create GitHub release automatically

StrikerRUS:
Tags do not trigger Azure Pipelines by default, that's why `trigger` field in YAML file is needed: https://docs.microsoft.com/en-us/azure/devops/pipelines/build/triggers?view=azure-devops&tabs=yaml#tags.

I tested it and here are results:

```
D:\Users\nekit\Downloads\LightGBM>git tag v2.2.4
D:\Users\nekit\Downloads\LightGBM>git push origin v2.2.4
Total 0 (delta 0), reused 0 (delta 0)
To github.com:microsoft/LightGBM.git
 * [new tag]         v2.2.4 -> v2.2.4
```

image

Clicking on `Edit` 
image
 and then on `Publish release`
image
results in successful release which is available for users:

image

@guolinke If you feel that we do not need `Draft` stage for any adjustments, then it's possible to publish ordinary release directly from Azure: `isDraft: false`.


guolinke:
@StrikerRUS I think it is good to use `isDraft`, so that we can avoid the release by accident push.

StrikerRUS:
@guolinke Yeah, I think so too.

Docker triggering seems to be broken. I sent you some screens in Slack.

StrikerRUS:
@guolinke Can we merge this?

PR #2159, opened 2019-05-08
[R-package] Fix best_iter and best_score

Laurae2:
This should fix #2158 and #2029.

"best" rule:
* Take the first iteration where the best score is attained, even if the further iterations have identical scores (do not take the last)
* When there is no scoring done, the best score becomes NA and the best iteration is -1 (currently)

Later, we should enforce the metric used for early stopping should be only the first one, or at worst give the user the ability to choose the metric (best is the first).

note: @jameslamb spaces from RStudio to fix

Example (change to `metric = "auc"` and `max_depth = 3` to test maximization):

```r
library(lightgbm)
data(agaricus.train, package = "lightgbm")
train <- agaricus.train
dtrain <- lgb.Dataset(train$data, label = train$label)
data(agaricus.test, package = "lightgbm")
test <- agaricus.test
dtest <- lgb.Dataset.create.valid(dtrain, test$data, label = test$label)
params <- list(objective = "regression", metric = "l2")
valids <- list(test = dtest)
model <- lgb.train(params,
                   dtrain,
                   100,
                   min_data = 1,
                   learning_rate = 0.5)
model$best_score
model$best_iter
# > model$best_score
# [1] NA
# > model$best_iter
# [1] -1

model <- lgb.train(params,
                   dtrain,
                   100,
                   valids,
                   min_data = 1,
                   learning_rate = 0.5)
model$best_score
model$best_iter
# > model$best_score
# [1] 7.497024e-62
# > model$best_iter
# [1] 100

model <- lgb.train(params,
                   dtrain,
                   100,
                   valids,
                   min_data = 1,
                   learning_rate = 0.5,
                   early_stopping_rounds = 10)
model$best_score
model$best_iter
# > model$best_score
# [1] 7.497024e-62
# > model$best_iter
# [1] 100

model <- lgb.cv(params,
             dtrain,
             10,
             nfold = 5,
             min_data = 1,
             learning_rate = 1)
model$best_score
model$best_iter
# > model$best_score
# [1] 0.0003072197
# > model$best_iter
# [1] 1

model <- lgb.cv(params,
             dtrain,
             10,
             nfold = 5,
             min_data = 1,
             learning_rate = 1,
             early_stopping_rounds = 5)
model$best_score
model$best_iter
# > model$best_score
# [1] 5.369631e-18
# > model$best_iter
# [1] 3
```

PR #2172, opened 2019-05-14
first metric only in earlystopping for cli

guolinke:
refer https://github.com/microsoft/LightGBM/pull/2159#issuecomment-491576482

### DIFF
--- a/include/LightGBM/config.h
+++ b/include/LightGBM/config.h
@@ -261,7 +261,6 @@ struct Config {
   int early_stopping_round = 0;
 
   // desc = set this to ``true``, if you want to use only the first metric for early stopping
-  // desc = **Note**: can be used only in CLI version
   bool first_metric_only = false;
 
   // alias = max_tree_output, max_leaf_output
--- a/include/LightGBM/config.h
+++ b/include/LightGBM/config.h
@@ -260,7 +260,8 @@ struct Config {
   // desc = ``<= 0`` means disable
   int early_stopping_round = 0;
 
-  // desc = will only use first metric for early stopping if set to ``True``
+  // desc = set this to ``true``, if you want to use only the first metric for early stopping
+  // desc = **Note**: can be used only in CLI version
   bool first_metric_only = false;
 
   // alias = max_tree_output, max_leaf_output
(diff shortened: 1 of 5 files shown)


StrikerRUS:
I added the note about that this param is for CLI only and slightly paraphrased description for the consistency with others.

include/LightGBM/config.h
```diff
@@ -260,7 +260,8 @@ struct Config {
   // desc = ``<= 0`` means disable
   int early_stopping_round = 0;
 
-  // desc = will only use first metric for early stopping if set to ``True``
+  // desc = set this to ``true``, if you want to use only the first metric for early stopping
+  // desc = **Note**: can be used only in CLI version
```
guolinke: Although this parameter is actually used in cli version, there is another parameter with the same name in python/R package. Therefore, I think the note is not needed, as well as the early_stopping_round.

StrikerRUS: It's a little bit another situation (in Python, at least).
`early_stopping_round` is a parameter for `train`/`cv`, so everything is OK, but `first_metric_only` is a parameter for **callback**.
https://github.com/microsoft/LightGBM/blob/2a3691703c1f5aa5109881455cfc909f80c7f50f/python-package/lightgbm/callback.py#L153

guolinke: okay, I think the first_metric_only should be in train/cv as well, like the early_stopping_round:
https://github.com/microsoft/LightGBM/blob/2a3691703c1f5aa5109881455cfc909f80c7f50f/python-package/lightgbm/engine.py#L482-L483

StrikerRUS: Do you think that we should try to retrieve `first_metric_only` from `params` and pass to callback?
https://github.com/microsoft/LightGBM/blob/2a3691703c1f5aa5109881455cfc909f80c7f50f/python-package/lightgbm/engine.py#L183-L184

guolinke: yes, just like the early stopping rounds:
https://github.com/microsoft/LightGBM/blob/2a3691703c1f5aa5109881455cfc909f80c7f50f/python-package/lightgbm/engine.py#L122-L126

StrikerRUS: Got it! Then I'll create a PR for this and remove my note.

PR #2174, opened 2019-05-14
[ci] migrate to new Dockerfile

StrikerRUS:


.vsts-ci.yml
```diff
@@ -4,7 +4,7 @@ variables:
 resources:
   containers:
   - container: ubuntu1404
-    image: lightgbm/vsts-agent:ubuntu-14.04
+    image: lightgbm/vsts-agent:latest
```
guolinke: sorry,  I think the tag name should be changed to ubuntu-14.04.
I will fix it in docker hub

StrikerRUS: Got it!

StrikerRUS: @guolinke Done! Can we remove unused tags on Docker Hub?

StrikerRUS: I mean `latest` and `test`.

guolinke: sure

Issue #1139, opened 2017-12-25
Metric Parameters: top-k error rate for multi-class classification?


mingwugmail:
Is it possible to add this feature?

I assume the current

multi_error, error rate for mulit-class classification

is top-1 error rate.

Laurae2:
It is feasible in R/Python wrappers, you can also contribute to add it in C++.

StrikerRUS:
Added in #2178.

PR #2178, opened 2019-05-15
Top k multi error

btrotta:
Implements the feature requested in issue #1139 . Added new parameter top_k_threshold. When the metric is set to `multi_error`, parameter `top_k_threshold` can be used to obtain the top-k multi-error. By default this parameter is set to 1, which gives the usual multi-error.


StrikerRUS:
@btrotta Thank you very much for your contribution!

Seems that your new tests fail on Python 2.7. Can you please take a look?

Also please move them to the existent `test_engine.py` file. Maybe in the future we should divide our tests in more elegant way, but for now there is no need to create a new separate file for 2 small tests.

src/metric/multiclass_metric.hpp
```diff
@@ -129,20 +129,22 @@ class MulticlassMetric: public Metric {
...
 
-  inline static double LossOnPoint(label_t label, std::vector<double>& score) {
+  inline static double LossOnPoint(label_t label, std::vector<double>& score, const Config& config) {
     size_t k = static_cast<size_t>(label);
+	int num_larger = 0;
     for (size_t i = 0; i < score.size(); ++i) {
-      if (i != k && score[i] >= score[k]) {
-        return 1.0f;
+		if (score[i] > score[k]) num_larger++;
+		if (num_larger >= config.top_k_threshold) return 1.0f;
```
StrikerRUS: Can you please replace tabs with spaces here?

tests/python_package_test/test_multiclass_metrics.py
```diff
@@ -0,0 +1,47 @@
...
+        params = {'objective': 'multiclass', 'num_classes': 10, 'metric': 'multi_error', 'top_k_threshold': 1,
+                  'num_leaves': 4, 'seed': 0}
+        results = {}
+        est = lgb.train(params, lgb_data, valid_sets=[lgb_data], valid_names=['train'], evals_result=results)
+        predict_1 = est.predict(X)
+        # check that default gives same result as k = 1
+        self.assertTrue(np.max(np.abs(predict_1 - predict_default)) < 1e-5)
+        # check against independent calculation
+        err = top_k_error(y, predict_default, 1)
+        self.assertTrue(abs(results['train']['multi_error'][-1] - err) < 1e-5)
```
StrikerRUS: I think you can use `numpy.testing` module to simply code significantly. 
https://docs.scipy.org/doc/numpy/reference/routines.testing.html

StrikerRUS:
We already have `top_k`/`topk` parameter for voting parallel: https://lightgbm.readthedocs.io/en/latest/Parameters.html#top_k.
I suppose that for this parameter we should take less confusing name. What about `multi_error_top_k_threshold`?
@guolinke

guolinke:
yes, I think it is better to change the parameter name.

btrotta:
@StrikerRUS Good point! I've fixed this in the latest commit. The metric is now defined so that the top-k error on a sample is 0 if there are at least `num_classes - k` predictions strictly less than the prediction on the true class.

StrikerRUS:
@btrotta Thank you for hotfixing that!
I think we can enhance the metric name to help users orient in logs.
Just like `map` or `ndcg`:
https://github.com/microsoft/LightGBM/blob/0a4a7a86f5a1d3146c36c7d8c082154a193d4893/src/metric/map_metric.hpp#L38-L40
The only thing I'm doubt about is that we cannot change `multi_error` -> `multi_error@1` due to backward compatibility with users' existing codebase which relies on the previous name.

btrotta:
@StrikerRUS Done

btrotta:
Formatting issues fixed in latest commit.

include/LightGBM/config.h
```diff
@@ -744,6 +744,14 @@ struct Config {
   // desc = `NDCG <https://en.wikipedia.org/wiki/Discounted_cumulative_gain#Normalized_DCG>`__ and `MAP <https://makarandtapaswi.wordpress.com/2012/07/02/intuition-behind-average-precision-and-map/>`__ evaluation positions, separated by ``,``
   std::vector<int> eval_at;
 
+  // check = >0
+  // desc = used only with ``multi-error`` metric
```
StrikerRUS: `multi-error` -> `multi_error`
Refer to 
https://github.com/microsoft/LightGBM/blob/a66c34407dcb3f94f6d29b0a5098d7ee0e1dd59b/include/LightGBM/config.h#L726

include/LightGBM/config.h
```diff
@@ -744,6 +744,14 @@ struct Config {
...
+  // descl2 = threshold for top-k multi-error metric
+  // descl2 = the error on each sample is ``0`` if the true class is among the top ``multi_error_top_k`` predictions, and ``1`` otherwise
+  // descl2 = more precisely, the error on a sample is ``0`` if there are at least ``num_classes - multi_error_top_k`` predictions strictly less than the prediction on the true class
```
StrikerRUS: I'm sorry, I had to be more clear! `descl2` is only for this one clarification line, which starts from `more precisely ...` and was in brackets. All other lines should remain marked as `desc`.

btrotta:
@StrikerRUS Thanks so much for your careful review! Those 2 issues with docs are fixed now.

StrikerRUS:
@guolinke Can you please give a second review to this PR?

guolinke:
it looks good to me

StrikerRUS:
@guolinke Should this PR be in 2.2.4 release?

guolinke:
@StrikerRUS yeah, it can be.

StrikerRUS:
@guolinke Then I'm merging 😃

PR #2181, opened 2019-05-17
[docs] remove duplicated param in Python-Intro.rst

leasunhy:
`num_round` is redundant here because it will be overrideen by `num_trees` in the `param` dictionary.


leasunhy:
@StrikerRUS Updated.

leasunhy:
@StrikerRUS Of course! I've updated again.

PR #2191, opened 2019-05-25
[python] fixed picklability of sklearn models with custom obj and updated docstings for custom obj

### LINKED ISSUE #2189: AttributeError: Can\'t pickle local object \'_objective_function_wrapper.<locals>.inner
LightGBM models are normally picklable, but when using a custom objective (and I assume eval_function etc.) they are not picklable.

```
import lightgbm as lgb
import _pickle as pickle
import numpy as np

def custom_asymmetric_objective(y_true, y_pred):
    residual = (y_true - y_pred).astype("float")
    grad = np.where(residual < 0, -2 * 10.0 * residual, -2 * residual)
    hess = np.where(residual < 0, 2 * 10.0, 2.0)
    return grad, hess

params = {}
params['objective'] = custom_asymmetric_objective
model = lgb.LGBMRegressor(**params)
X = np.array([[1,2,3,4], [2,3,4,5]]).transpose()
y = np.array([1,2,3,3])
model.fit(X, y)
preds = model.predict(X)
pickle.dump(model, open("model.pkl", "wb"))
print("Done")
```

Leads to:

```
  File "lgbm.py", line 22, in <module>
    pickle.dump(model, open("model.pkl", "wb"))
AttributeError: Can't pickle local object '_objective_function_wrapper.<locals>.inner'
jon@pseudotensor:~$ 

```

The problem is the def inside def way of defining the function:
https://github.com/microsoft/LightGBM/blob/master/python-package/lightgbm/sklearn.py#L18-L78

The solution is to either:
1) remove such def inside def uses and just directly make the function.
2) If using a decorator, then you need to use the standard way that allows them to be pickleable via `functools.wraps`

http://gael-varoquaux.info/programming/decoration-in-python-done-right-decorating-and-pickling.html

StrikerRUS:
- Fixed #2189.
- Updated docstings to specify signatures of fobj and feval custom functions for every occurence.

@pseudotensor Please confirm the fix.

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -59,7 +59,7 @@ def is_numeric(obj):
 
 
 def is_numpy_1d_array(data):
-    """Check whether data is a 1-D numpy array."""
+    """Check whether data is a numpy 1-D array."""
     return isinstance(data, np.ndarray) and len(data.shape) == 1
 
 
@@ -69,7 +69,7 @@ def is_1d_list(data):
 
 
 def list_to_1d_numpy(data, dtype=np.float32, name='list'):
-    """Convert data to 1-D numpy array."""
+    """Convert data to numpy 1-D array."""
     if is_numpy_1d_array(data):
         if data.dtype == dtype:
             return data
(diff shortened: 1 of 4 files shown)


pseudotensor:
can we merge?

pseudotensor:
Thanks!

PR #2076, opened 2019-04-03
[ci] for autogenerating docs

nabokovas:
Comments were reworked there for easier autogenerating documentation.
No changes for the functions.
No description for LGBM_NetworkInitWithFunctions.

PR #2195, opened 2019-05-26
[docs] fixed and enhanced format of C API

StrikerRUS:
Final result: https://lightgbm.readthedocs.io/en/c_api/C-API.html.

### DIFF
--- a/include/LightGBM/c_api.h
+++ b/include/LightGBM/c_api.h
@@ -1,6 +1,14 @@
 /*!
- * Copyright (c) 2016 Microsoft Corporation. All rights reserved.
- * Licensed under the MIT License. See LICENSE file in the project root for license information.
+ * \file c_api.h
+ * \copyright Copyright (c) 2016 Microsoft Corporation. All rights reserved.
+ *            Licensed under the MIT License. See LICENSE file in the project root for license information.
+ * \note
+ * To avoid type conversion on large data, the most of our exposed interface supports both float32 and float64,
+ * except the following:
+ *   1. gradient and Hessian;
+ *   2. current score for training and validation data.
+ *   .
+ * The reason is that they are called frequently, and the type conversion on them may be time-cost.
  */
 #ifndef LIGHTGBM_C_API_H_
 #define LIGHTGBM_C_API_H_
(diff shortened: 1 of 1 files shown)


include/LightGBM/c_api.h
```diff
@@ -10,55 +18,45 @@
...
+ */
 LIGHTGBM_C_EXPORT int LGBM_DatasetCreateFromFile(const char* filename,
                                                  const char* parameters,
                                                  const DatasetHandle reference,
                                                  DatasetHandle* out);
 
 /*!
- * \fn LGBM_DatasetCreateFromSampledColumn
  * \brief Create an empty dataset by sampling data.
+ * \param sample_data Sampled data, grouped by the column
```
StrikerRUS: This lost param was the reason why #2076 ruined all C API docs.
It's quite strange, but `\fn` directive makes Doxygen produce xml without descriptions instead of generating the error.
Also, Doxygen's docs say that we don't need to use `\fn` directives when comments are placed directly before function:
> Warning
Do not use this command if it is not absolutely needed, since it will lead to duplication of information and thus to errors.
http://www.doxygen.nl/manual/commands.html#cmdfn

So I removed them all.

include/LightGBM/c_api.h
```diff
@@ -10,55 +18,45 @@
...
-*/
+ */
 LIGHTGBM_C_EXPORT int LGBM_DatasetCreateFromFile(const char* filename,
                                                  const char* parameters,
                                                  const DatasetHandle reference,
                                                  DatasetHandle* out);
 
 /*!
- * \fn LGBM_DatasetCreateFromSampledColumn
  * \brief Create an empty dataset by sampling data.
```
StrikerRUS: Is it really empty?

guolinke: it will allocate the spaces, so no really empty. But the data are push afterward.

StrikerRUS: Thanks, got it! Maybe then change it to `Allocate the space for dataset by sampling data.`?

guolinke: sure, BTW, it will also bucket the feature bins (construct bin_mapper) according to sample data.

StrikerRUS: Updated. And one next function description too.

PR #2214, opened 2019-06-04
balanced bagging

guolinke:
This may work for imbalanced binary-class data.


StrikerRUS:
@guolinke I added new line in docs which explicitly states that this param is used only in `binary` task

candalfigomoro:
@guolinke @StrikerRUS Does this also affect the random forest mode?

guolinke:
Yeah, they share the same bagging logic.

PR #2224, opened 2019-06-12
add "download" badge

guolinke:


README.md
```diff
@@ -7,6 +7,7 @@ LightGBM, Light Gradient Boosting Machine
 [![Documentation Status](https://readthedocs.org/projects/lightgbm/badge/?version=latest)](https://lightgbm.readthedocs.io/)
 [![GitHub Issues](https://img.shields.io/github/issues/Microsoft/LightGBM.svg)](https://github.com/microsoft/LightGBM/issues)
 [![License](https://img.shields.io/badge/license-MIT-blue.svg)](https://github.com/microsoft/LightGBM/blob/master/LICENSE)
+[![Downloads](https://pepy.tech/badge/lightgbm)](https://pepy.tech/project/lightgbm)
```
StrikerRUS: I think it's a little bit confusing to have `Downloads` badge in general README because it's not clear what these "downloads" mean: GitHub releases, Python-package, R-package, or something else.

guolinke: good point. maybe we can try to add number of clones and download of release in main page.

StrikerRUS: > maybe we can try to add number of clones and download of release in main page.

Just replaced issue count badge with number of releases downloads. I think issue badge was useless, because one can see number of issues upper at the same page.

StrikerRUS: Number of clones: https://github.com/badges/shields/issues/880.

guolinke: it seems the number of release download is very low. I guess most users use pip to get, or clone then build by themself.

StrikerRUS: Agree! 33k during more than 2 years...

Maybe JARs will increase downloads count...

guolinke: I think it is better to remove the download counts of release.

PR #2269, opened 2019-07-18
add weight in tree model output

### LINKED ISSUE #2037: Saving weight values of a node
Hi, right now for lightgbm trees following information is stored
1.Node index
2.Leaf count
3.Splitting variable
4.Threshold
5.Internal value
6.Leaf value
Is it possible to store the sum of weights too? It would be a great help in debugging or understanding the prediction.

guolinke:
fix #2037

### DIFF
--- a/src/boosting/gbdt_model_text.cpp
+++ b/src/boosting/gbdt_model_text.cpp
@@ -14,7 +14,7 @@
 
 namespace LightGBM {
 
-const std::string kModelVersion = "v2";
+const std::string kModelVersion = "v3";
 
 std::string GBDT::DumpModel(int start_iteration, int num_iteration) const {
   std::stringstream str_buf;
--- a/examples/python-guide/notebooks/interactive_plot_example.ipynb
+++ b/examples/python-guide/notebooks/interactive_plot_example.ipynb
@@ -355,7 +355,9 @@
     "                          'split_gain',\n",
     "                          'internal_value',\n",
     "                          'internal_count',\n",
-    "                          'leaf_count'],\n",
+    "                          'internal_weight',\n",
+    "                          'leaf_count',\n",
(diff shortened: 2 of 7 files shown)


src/boosting/gbdt_model_text.cpp
```diff
@@ -14,7 +14,7 @@
 
 namespace LightGBM {
 
-const std::string kModelVersion = "v2";
+const std::string kModelVersion = "v3";
```
StrikerRUS: Is new model format backward compatible with current v2?

StrikerRUS:
@guolinke I can update plotting part if you don't mind.
https://github.com/microsoft/LightGBM/blob/611cf5d414a8273dfdd261651e2f127322a3db89/python-package/lightgbm/plotting.py#L392-L396

src/boosting/gbdt_model_text.cpp
guolinke: Yeah, it can load the v2 model file

guolinke:
@StrikerRUS sure, thanks very much!

StrikerRUS:
@guolinke Just updated plotting module in Python in the latest commit.

StrikerRUS:
ping @jameslamb and @Laurae2 for R-package updates

https://github.com/microsoft/LightGBM/blob/master/R-package/R/lgb.model.dt.tree.R
https://github.com/microsoft/LightGBM/blob/master/R-package/R/lgb.interprete.R

guolinke:
@StrikerRUS could this PR merge first? then a new PR for R package could be proposed.

StrikerRUS:
@guolinke Yeah, sure!

jameslamb:
Created #2281 to track the R work, agree with the decision that was made to merge this and come back to R.

Issue #2268, opened 2019-07-17
Change FAQs in docs so we can link to individual questions


jameslamb:
While helping out with #2267 tonight, I wanted to link to a very specific question [in our FAQs](https://lightgbm.readthedocs.io/en/latest/FAQ.html). However, because they're set up as a bulleted list, they can't be linked to.

Changing the questions to header-3s or header-4s will create hyperlinks! To do this in `sphinx`, one could go into [the FAQ rst file](https://github.com/microsoft/LightGBM/blob/master/docs/FAQ.rst) and change stuff like this:

```
**Question 4**: I am using Windows. Should I use Visual Studio or MinGW for compiling LightGBM?
```

into something like

```
I am using Windows. Should I use Visual Studio or MinGW for compiling LightGBM?
'''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''
```

I would also recommend removing the "Question 1", "Question 2", etc. numbering, as I don't think it's providing any value.

StrikerRUS:
> I would also recommend removing the "Question 1", "Question 2", etc. numbering, as I don't think it's providing any value.

It's better to leave numbering for "backward compatibility", because in previous issues we have many refs to FAQ with words like "See Question 4 and 8", or "Refer to FAQ Question 9".

hayesall:
#2272 should probably to be merged before addressing this issue (at least to avoid merge conflicts and different formatting).

I made local changes to address this issue, and can open a pull request soon:

<img width="1267" alt="Screen Shot 2019-07-27 at 1 07 45 PM" src="https://user-images.githubusercontent.com/11916674/61997464-c2594700-b06f-11e9-8bf7-7fd1495ba233.png">

jameslamb:
> > I would also recommend removing the "Question 1", "Question 2", etc. numbering, as I don't think it's providing any value.
> 
> It's better to leave numbering for "backward compatibility", because in previous issues we have many refs to FAQ with words like "See Question 4 and 8", or "Refer to FAQ Question 9".

Does "previous issues" mean things that we control in this repo? If so, having hyperlinks means we could change those to point to the actual FAQs. If it's only things in this repo that we control, I would be willing to do the work to change them.

StrikerRUS:
[...]

PR #2272, opened 2019-07-19
[docs] added FAQ on early stopping (fixes #2270)

### LINKED ISSUE #2270: API design choice: why not enable early stopping by default?
In scikit-learn https://github.com/scikit-learn/scikit-learn/issues/14303 we're considering enabling early stopping by default.

We're curious about why  you chose not to enable it by default in LightGBM, considering that early stopping is almost always useful in practice?

Thanks!

jameslamb:
Added the discussion from #2270 to our documentation. Thanks again for the question @NicolasHug

### DIFF
--- a/docs/FAQ.rst
+++ b/docs/FAQ.rst
@@ -171,9 +171,9 @@ LightGBM
 
 -  **Solution 12**: Early stopping involves choosing a validation set, a special type of holdout which is used to evaluate the current state of the model after each iteration to see if training can stop.
 
-    In ``LightGBM``, `we have decided to require that users specify this set directly <./Python-Intro.html#early-stopping>`_. Many options exist for splitting training data into training, test, and validation sets.
+   In ``LightGBM``, `we have decided to require that users specify this set directly <./Parameters.html#valid>`_. Many options exist for splitting training data into training, test, and validation sets.
 
-    The appropriate splitting strategy depends on the task and domain of the data, information that a modeler has but which ``LightGBM`` as a general-purpose tool does not.
+   The appropriate splitting strategy depends on the task and domain of the data, information that a modeler has but which ``LightGBM`` as a general-purpose tool does not.
 
 --------------
 
--- a/docs/FAQ.rst
+++ b/docs/FAQ.rst
@@ -167,6 +167,16 @@ LightGBM
 
 --------------
 
+-  **Question 12**: Why is early stopping not enabled by default in LightGBM?
+
+-  **Solution 12**: Early stopping involves choosing a validation set, a special type of holdout which is used to evaluate the current state of the model after each iteration to see if training can stop.
+
+    In ``LightGBM``, `we have decided to require that users specify this set directly <./Python-Intro.html#early-stopping>`_. Many options exist for splitting training data into training, test, and validation sets.
+
+    The appropriate splitting strategy depends on the task and domain of the data, information that a modeler has but which ``LightGBM`` as a general-purpose tool does not.
+
+--------------
+
 R-package
 ~~~~~~~~~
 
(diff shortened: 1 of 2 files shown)


docs/FAQ.rst
```diff
@@ -167,6 +167,16 @@ LightGBM
 
 --------------
 
+-  **Question 12**: Why is early stopping not enabled by default in LightGBM?
+
+-  **Solution 12**: Early stopping involves choosing a validation set, a special type of holdout which is used to evaluate the current state of the model after each iteration to see if training can stop.
+
+    In ``LightGBM``, `we have decided to require that users specify this set directly <./Python-Intro.html#early-stopping>`_. Many options exist for splitting training data into training, test, and validation sets.
```
StrikerRUS: Since this FAQ entry is in general section, not in `Python` one, link should point to something language package independent, I think. Maybe use this link https://github.com/microsoft/LightGBM/blob/master/docs/Parameters.rst#valid?

StrikerRUS:
Bad indents cause the second and the third paragraphs look like quoting.

image

docs/FAQ.rst
jameslamb: yep good point, I'll revise. Thank you!

GitHub won't let me comment on your note about the indentation so I'll tell you here...I'll fix that too and actually run `make html` and do a visual inspection.

StrikerRUS:
ping @jameslamb 

This PR is blocking #2268.

jameslamb:
@StrikerRUS thanks for the ping, let me push an update right now!

docs/FAQ.rst
jameslamb: image

^ think I've fixed that bad indentation now. Added this link as well.

I also touched one line in `README.rst`, saw an unnecessary line break.

I also rebased to `master` since there have been some changes since I first opened this PR.

docs/FAQ.rst
```diff
@@ -167,6 +167,16 @@ LightGBM
 
 --------------
 
+-  **Question 12**: Why is early stopping not enabled by default in LightGBM?
+
+-  **Solution 12**: Early stopping involves choosing a validation set, a special type of holdout which is used to evaluate the current state of the model after each iteration to see if training can stop.
+
+   In ``LightGBM``, `we have decided to require that users specify this set directly <./Parameters.html#valid>`_. Many options exist for splitting training data into training, test, and validation sets.
```
StrikerRUS:
```suggestion
   In ``LightGBM``, `we have decided to require that users specify this set directly <./Parameters.rst#valid>`_. Many options exist for splitting training data into training, test, and validation sets.
```

StrikerRUS: @jameslamb Can you please approve the suggestion?

jameslamb: @StrikerRUS Sorry for the delay. I disagree with it. It needs to be `.html` so that the link will work correctly.

See, for example, how all the links on the site are to specific `.html` files: https://lightgbm.readthedocs.io/en/latest/Python-Intro.html

`sphinx` will always take `foo.rst` and create a `foo.html`, so we can rely on the `.html` being there after running `sphinx-build`.

StrikerRUS: @jameslamb There was a long story behind this. In short, we came to the agreement about that we need correct links on both GitHub and RTD. `.rst` extension allows to navigate correctly on GitHub, then we have a JS script which replaces `.rst` with `.html` after Sphinx job:
https://github.com/microsoft/LightGBM/blob/274f340ae8232c707428d1747a857f4fc2511146/docs/_static/js/script.js#L2-L3

Your link is working only for RTD. Take a look at any other docs for examples.

jameslamb: Oh wow ok! Sorry, I had not noticed that before and wasn't aware we were doing that before. Ok awesome!

Thanks for the clarification.

jameslamb: change committed

StrikerRUS: @jameslamb No problem! I tried to describe all details :-)

PR #2209, opened 2019-06-01
[python] Bug fix for first_metric_only on earlystopping.

matsuken92:
1. Order of metric list is not fixed even if it is defined by list. So explicitly indicating which metric is used on early stopping.
1. Due to introducing eval_train_metric feature, if the feature is enabled, then the first metric became train score of the metrics, so it does not appropriate for early stopping. So until the specified metrics of validation is coming, skipping to check early stopping in the loop.

(This PR is latest version of #2127. Once this PR is activated #2127 should be closed.)

PR #2273, opened 2019-07-20
fix metric alias

guolinke:
refer to https://github.com/microsoft/LightGBM/pull/2209#issuecomment-513459716

### DIFF
--- a/include/LightGBM/objective_function.h
+++ b/include/LightGBM/objective_function.h
@@ -89,7 +89,6 @@ class ObjectiveFunction {
   */
   LIGHTGBM_EXPORT static ObjectiveFunction* CreateObjectiveFunction(const std::string& str);
 
-  static std::string ParseObjective(const std::string& type);
 };
 
 }  // namespace LightGBM
--- a/include/LightGBM/objective_function.h
+++ b/include/LightGBM/objective_function.h
@@ -88,6 +88,8 @@ class ObjectiveFunction {
   * \brief Load objective function from string object
   */
   LIGHTGBM_EXPORT static ObjectiveFunction* CreateObjectiveFunction(const std::string& str);
+
+  static std::string ParseObjective(const std::string& type);
 };
 
 }  // namespace LightGBM
--- a/include/LightGBM/objective_function.h
+++ b/include/LightGBM/objective_function.h
@@ -88,7 +88,6 @@ class ObjectiveFunction {
   * \brief Load objective function from string object
   */
   LIGHTGBM_EXPORT static ObjectiveFunction* CreateObjectiveFunction(const std::string& str);
-
 };
 
 }  // namespace LightGBM
--- a/src/io/config.cpp
+++ b/src/io/config.cpp
@@ -4,6 +4,7 @@
  */
 #include <LightGBM/config.h>
 #include <LightGBM/metric.h>
+#include <LightGBM/objective_function.h>
 
 #include <LightGBM/utils/common.h>
 #include <LightGBM/utils/log.h>
@@ -68,7 +69,7 @@ void GetObjectiveType(const std::unordered_map<std::string, std::string>& params
   std::string value;
   if (Config::GetString(params, "objective", &value)) {
     std::transform(value.begin(), value.end(), value.begin(), Common::tolower);
-    *objective = value;
+    *objective = ObjectiveFunction::ParseObjective(value);
   }
 }
 
--- a/src/io/config.cpp
+++ b/src/io/config.cpp
@@ -3,8 +3,6 @@
  * Licensed under the MIT License. See LICENSE file in the project root for license information.
  */
 #include <LightGBM/config.h>
-#include <LightGBM/metric.h>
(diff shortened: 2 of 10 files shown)


StrikerRUS:
@guolinke Is it possible to share `std::string GetMetricType(const std::string& type)` function for objective too? It seems to be that the difference is in `l2`/`l2_root` and `multiclass`/`multiclassova` only.

StrikerRUS:
@guolinke I updated docs accordingly.

guolinke:
Sorry, didn't get it. Could explain with more details?

StrikerRUS:
@guolinke I mean, we have practically the same code for objective function here
https://github.com/microsoft/LightGBM/blob/master/src/objective/objective_function.cpp as for metric function which you've updated in this PR. So I wonder can we use `std::string GetMetricType(const std::string& type)` for replacing aliases with main names for both metrics and objectives.

StrikerRUS:
@guolinke In other words, replace 
https://github.com/microsoft/LightGBM/blob/7189743d224a405bb48050834460d107fb7c1c0b/src/io/config.cpp#L70
with
```
Metric::ParseMetrics(value, objective);
```

This will allow to work only with objectives names in further code, without considering aliases.

guolinke:
@StrikerRUS Thanks, good point! will update it accordingly.

StrikerRUS:
@guolinke Don't you want to merge `GetMetricType` and `ParseObjective` functions into one function (e.g. `ParseFunctionAliases`)? And that function can take one additional enum argument `function_type` with two possible values `metric` and `objective` to help with `l2`/`l2_root` and `multiclass`/`multiclassova` cases, which are different in two current functions.

guolinke:
@StrikerRUS I tried it, and think it is more easy-to-read to use two functions.
Actually, the duplicated aliases are no many, just two or three.
 but introducing the function_type, with caring about the duplicate or not, will increase the complexity of this function.

StrikerRUS:
@guolinke OK, got it!

include/LightGBM/objective_function.h
```diff
@@ -88,6 +88,7 @@ class ObjectiveFunction {
   * \brief Load objective function from string object
   */
   LIGHTGBM_EXPORT static ObjectiveFunction* CreateObjectiveFunction(const std::string& str);
+
```
StrikerRUS: Excess diff.

src/io/config.cpp
```diff
@@ -3,6 +3,8 @@
  * Licensed under the MIT License. See LICENSE file in the project root for license information.
  */
 #include <LightGBM/config.h>
+#include <LightGBM/metric.h>
+#include <LightGBM/objective_function.h>
```
StrikerRUS: I guess that these includes are not needed anymore.

src/io/config.cpp
```diff
@@ -63,41 +65,88 @@ void GetBoostingType(const std::unordered_map<std::string, std::string>& params,
   }
 }
 
+std::string ParseObjectiveAlias(const std::string& type) {
+  if (type == std::string("regression") || type == std::string("regression_l2")
+    || type == std::string("mean_squared_error") || type == std::string("mse")
+    || type == std::string("l2_root") || type == std::string("root_mean_squared_error") || type == std::string("rmse")) {
+    return "regression";
```
StrikerRUS: Maybe add `l2` alias here (and in docs)?

src/io/config.cpp
```diff
@@ -63,41 +65,88 @@ void GetBoostingType(const std::unordered_map<std::string, std::string>& params,
...
+  } else if (type == std::string("regression_l1") || type == std::string("mean_absolute_error") || type == std::string("mae")) {
+    return "regression_l1";
```
StrikerRUS: Maybe add `l1` alias here (and in docs)?

src/io/config.cpp
```diff
@@ -63,41 +65,88 @@ void GetBoostingType(const std::unordered_map<std::string, std::string>& params,
...
+    return "cross_entropy_lambda";
+  } else if (type == std::string("mean_absolute_percentage_error") || type == std::string("mape")) {
+    return "mape";
+  } else if (type == std::string("none") || type == std::string("null") || type == std::string("custom") || type == std::string("na")) {
+    return "custom";
+  }
+  return type;
+}
+
+std::string ParseMetricAlias(const std::string& type) {
```
StrikerRUS: Add aliases for `custom` too?

StrikerRUS:
I updated docs according to new objective names.

StrikerRUS:
I think that the following code can be simplified by removing checks for aliases, since we have already done it earlier:

https://github.com/microsoft/LightGBM/blob/7189743d224a405bb48050834460d107fb7c1c0b/src/io/config.cpp#L198-L234

StrikerRUS:
I added new aliases for regression objectives in docs.

StrikerRUS:
@guolinke Can we merge?

guolinke:
Yes

PR #2279, opened 2019-07-23
Conda install would be stuck at this stage

findalexli:
Adding --yes at the end of each conea command. 

Otherwise, the docker install process would be stuck with the following prompt that you cannot answer to 
Proceed ([y]/n)?

### DIFF
--- a/docker/gpu/dockerfile.gpu
+++ b/docker/gpu/dockerfile.gpu
@@ -74,8 +74,8 @@ RUN echo 'export PATH=/opt/conda/bin:$PATH' > /etc/profile.d/conda.sh && \
     /bin/bash ~/miniconda.sh -b -p /opt/conda && \
     rm ~/miniconda.sh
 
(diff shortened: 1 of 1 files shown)


findalexli:
In fact, the process can keep on running after roughly 2-5 minutes of inactivity from the user end. But I think passing --yes would be better.

StrikerRUS:
@findalexli Thanks for spotting this issue! I think it should be better to set `always_yes` in conda's config to avoid doing this for each command.
https://github.com/microsoft/LightGBM/blob/86a957830b7caf2cedb7aaa23e824df6b38f7c7f/.ci/setup.sh#L59

StrikerRUS:
@findalexli Sorry, but why did you close the PR?

findalexli:
@StrikerRUS Hello. Let me reopen this. I did it by accident.

jameslamb:
Thanks for reopening! I agree, setting conda config one time like @StrikerRUS has suggested would be preferred. Could you please make that change?

PR #2286, opened 2019-07-25
[docs] 🎨 Sphinx Autosummary for generating Python-API documentation

hayesall:
### Executive Summary

This replaces `autoclass/autofunction` with [sphinx autosummary](https://www.sphinx-doc.org/en/master/usage/extensions/autosummary.html) to create and paginate the Python API.

### Summary of Changes

- Add `docs/.gitignore` to not track autosummary stubs
- Add `sphinx.ext.autosummary` in `docs/conf.py`
  - Add 'members' and 'inherited-members' as default parameters
  - Add 'autosummary = True' for setting output with `:toctree:`
- Add `.. autosummary::` tags to replace `.. autoclass::` and `.. autofunction::`

### Pull Request Motivation 📝

Previously the `Python-API.rst` dumped all of the Python API onto a single page when `sphinx-build` generated html pages.

This replaces the Python-API documentation with an index listing all modules, and paginates all functions and classes onto separate pages.

In summary, this pull request:

- Requires less scrolling on the [Python-API](https://lightgbm.readthedocs.io/en/latest/Python-API.html) page
- Makes it easier to find specific pieces of the full Python documentation

### Alternatives

This could also be accomplished by creating individual pages by hand and putting `autoclass` and `autofunction` on each. Autosummary simply makes this process more scalable.

Any feedback would be appreciated. **Thank you developers!**

--- 

## Screenshots

**Before (left) and After (right) of the `Python-API.html` landing page**

before_after

**Training API**: `lightgbm.cv`. The `lightgbm.cv` function has its own page.

Screen Shot 2019-07-25 at 1 49 09 PM

### DIFF
--- a/docs/.gitignore
+++ b/docs/.gitignore
@@ -1,4 +1,0 @@
-# docs-specific .gitignore
-
-# `sphinx.ext.autosummary` should output to `pythonapi/` using :toctree:
-pythonapi/
--- a/docs/.gitignore
+++ b/docs/.gitignore
@@ -1,0 +1,4 @@
+# docs-specific .gitignore
+
+# `sphinx.ext.autosummary` should output to `pythonapi/` using :toctree:
+pythonapi/
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -74,7 +74,7 @@ extensions = [
     'sphinx.ext.napoleon',
 ]
 
-autodoc_default_flags = ['members', 'inherited-members']
+autodoc_default_flags = ['members', 'inherited-members', 'show-inheritance']
 
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ['_templates']
@@ -123,7 +123,7 @@ pygments_style = 'default'
 todo_include_todos = False
 
 # Both the class' and the __init__ method's docstring are concatenated and inserted.
-autoclass_content = 'both'
+autoclass_content = 'class'
 
 # -- Configuration for C API docs generation ------------------------------
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -74,6 +74,7 @@ extensions = [
     'sphinx.ext.napoleon',
 ]
 
+autodoc_default_flags = ['members', 'inherited-members', 'show-inheritance']
 autodoc_default_options = {
     "members": True,
     "inherited-members": True,
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -74,7 +74,11 @@ extensions = [
     'sphinx.ext.napoleon',
 ]
 
-autodoc_default_flags = ['members', 'inherited-members', 'show-inheritance']
+autodoc_default_options = {
+    "members": True,
+    "inherited-members": True,
+    "show-inheritance": True,
+}
 
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ['_templates']
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -68,13 +68,19 @@ if needs_sphinx > sphinx.__version__:
 # ones.
 extensions = [
     'sphinx.ext.autodoc',
+    'sphinx.ext.autosummary',
     'sphinx.ext.todo',
     'sphinx.ext.viewcode',
     'sphinx.ext.napoleon',
 ]
 
+autodoc_default_flags = ['members', 'inherited-members']
+
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ['_templates']
+
+# Generate autosummary pages. Output should be set with: `:toctree: pythonapi/`
+autosummary_generate = True
 
 # The suffix(es) of source filenames.
 # You can specify multiple suffix as a list of string:
--- a/.gitignore
+++ b/.gitignore
@@ -339,6 +339,7 @@ instance/
 
 # Sphinx documentation
 docs/_build/
+docs/pythonapi/
 
 # Doxygen documentation
 docs/doxyoutput/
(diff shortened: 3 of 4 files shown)


docs/.gitignore
```diff
@@ -0,0 +1,4 @@
+# docs-specific .gitignore
+
+# `sphinx.ext.autosummary` should output to `pythonapi/` using :toctree:
+pythonapi/
```
StrikerRUS: Please move it to the general `.gitignore`. We keep rules for all folders there: https://github.com/microsoft/LightGBM/blob/5d3a3ea47eb8693c924ffb891387e93534f92896/.gitignore#L340

hayesall: Dropped the `docs/.gitignore` in e7cf3be. Added it to `.gitignore` instead, use should be fairly clear from context.

hayesall:
✅ Fixed the `'show-inheritance` by adding it to the `autodoc_default_flags` variable:

```python
autodoc_default_flags = ['members', 'inherited-members', 'show-inheritance']
```

---

✅ Changing [`autoclass_content`](http://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autoclass_content) to `'class'` removes the duplicate `__init__` entries.

---

❓ I don't have a solution for third-party links, but here is what I have currently:

- Existing links a specific portion (e.g. `Python-API.html#lightgbm.LGBMModel`) will still direct users to the index (in this case: `Python-API.html`) rather than a 404 error page.
- [ReadTheDocs has a special case for handling redirect links](https://docs.readthedocs.io/en/stable/user-defined-redirects.html), but these are handled through their website rather than configuration files.
- Some [plugins support JavaScript redirects](https://github.com/sphinx-contrib/redirects), but would introduce another dependency.

docs/conf.py
```diff
@@ -68,14 +68,20 @@ def run(self):
 # ones.
 extensions = [
     'sphinx.ext.autodoc',
+    'sphinx.ext.autosummary',
     'sphinx.ext.todo',
     'sphinx.ext.viewcode',
     'sphinx.ext.napoleon',
 ]
 
+autodoc_default_flags = ['members', 'inherited-members', 'show-inheritance']
```
StrikerRUS: @hayesall Thank you very much for quick fixes! Seems that this config value is deprecated. I think we can duplicate these setting in new `autodoc_default_options` config value to support all versions of Sphinx.
> Deprecated since version 1.8: Integrated into autodoc_default_options.
    https://www.sphinx-doc.org/en/2.0/usage/extensions/autodoc.html#confval-autodoc_default_flags

StrikerRUS:
@hayesall 
> Existing links to a specific portion (e.g. Python-API.html#lightgbm.LGBMModel) will still direct users to the index (in this case: Python-API.html) rather than a 404 error page.

Indeed! Since index page is rather small now, it should be not so hard for users to seek for the specified API member manually.

docs/conf.py
hayesall: @StrikerRUS Really glad you caught this. It doesn't look like sphinx raises a warning yet, but I know a few other places that should make this change as well.

From [autodoc_default_options](https://www.sphinx-doc.org/en/2.0/usage/extensions/autodoc.html#confval-autodoc_default_options) documentation:

> Setting None or True to the value is equivalent to giving only the option name to the directives.

`True` seemed more intuitive than `None`, so I replaced it this with:

```python
autodoc_default_options = {
    "members": True,
    "inherited-members": True,
    "show-inheritance": True,
}
```

The `sphinx-build` result appears to be the same as before.

StrikerRUS: @hayesall Sorry, I probably was unclear, but we should keep **both** (deprecated `autodoc_default_flags` and new `autodoc_default_options`) config values to support all versions of Sphinx.

hayesall: I did misinterpret that. Both are included now 👍

StrikerRUS:
@hayesall I'm sorry for the late question, but according to the logs it seems that `autosummary` scans each `.rst` file to generate a code. Is it possible to help it and specify that we need to scan only `Python-API.rst`? I guess it will save some time.

```
D:\Users\nekit\Downloads\LightGBM\docs>make html
Running Sphinx v1.8.5
making output directory...
[autosummary] generating autosummary for: Advanced-Topics.rst, C-API.rst, Development-Guide.rst, Experiments.rst, FAQ.rst, Features.rst, GPU-Performance.rst, GPU-Targets.rst, GPU-Tutorial.rst, GPU-Windows.rst, Installation-Guide.rst, Parallel-Learning-Guide.rst, Parameters-Tuning.rst, Parameters.rst, Python-API.rst, Python-Intro.rst, Quick-Start.rst, README.rst, gcc-Tips.rst, index.rst

...
```

hayesall:
@StrikerRUS Sphinx's [`generate.py#L193-L228`](https://github.com/sphinx-doc/sphinx/blob/a498960de9039b0d0c8d24f75f32fa4acd5b75e1/sphinx/ext/autosummary/generate.py#L193-L228) module looks like it should support only scanning a specific list of files:

The [`autosummary_generate`](https://www.sphinx-doc.org/en/master/usage/extensions/autosummary.html#confval-autosummary_generate) variable looks to be where this behavior is accessed, but I would need to experiment a bit to be sure.

I can look later today. If it's an easy fix I'll try to patch it in with #2293. If it's a bit more involved I'll open an issue.

StrikerRUS:
@hayesall Thanks a lot for your answer with starting points! It's absolutely not urgent and critical.

Please **do not** include any fix in #2293, because that PR is about different things. Just post updates here if you find anything valuable or make a new PR if you'd like (and have enough free time 😄 ).

PR #2291, opened 2019-07-27
change the priority of init_score and init_model.

### LINKED ISSUE #2290: Continued train with the same dataset
There is a bug for continued train with the same dataset.
Refer to example, 
https://github.com/microsoft/LightGBM/blob/5d3a3ea47eb8693c924ffb891387e93534f92896/examples/python-guide/advanced_example.py#L101-L134 ,

This will raise a warning: `warnings.warn("The prediction of init_model will be overridden by init_score.")"`, and the second continued train actually failed.

A workaround is adding `lgb_train = lgb.Dataset(X_train, y_train, weight=W_train, free_raw_data=False)`, before the second continued train.

guolinke:
to fix #2290

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -848,13 +848,12 @@ class Dataset(object):
             self.set_weight(weight)
         if group is not None:
             self.set_group(group)
-        # load init score
-        if init_score is not None:
-            self.set_init_score(init_score)
-            if predictor is not None:
-                warnings.warn("The prediction of init_model will be overridden by init_score.")
-        elif isinstance(predictor, _InnerPredictor):
+        if isinstance(predictor, _InnerPredictor):
+            if self._predictor is None and init_score is not None:
+                warnings.warn("The init_score will be overridden by the prediction of init_model.")
             self._set_init_score_by_predictor(predictor, data)
+        elif init_score is not None:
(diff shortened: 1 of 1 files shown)


python-package/lightgbm/basic.py
```diff
@@ -848,13 +848,12 @@ def _lazy_init(self, data, label=None, reference=None,
...
         if group is not None:
             self.set_group(group)
-        # load init score
-        if init_score is not None:
-            self.set_init_score(init_score)
-            if predictor is not None:
-                warnings.warn("The prediction of init_model will be overridden by init_score.")
-        elif isinstance(predictor, _InnerPredictor):
+        if isinstance(predictor, _InnerPredictor):
+            if self._predictor is None and init_score is not None:
```
StrikerRUS: Can you please clarify `if self._predictor is None` check?

guolinke: It means the init_score is set by user, no init_model

StrikerRUS: So, the case where everything `is not None` will silently replace old predictor and the overwrite init_score, right?

guolinke: yeah, replace the init_model will be silent.

StrikerRUS: Got it, thanks!

PR #2293, opened 2019-07-29
[docs] 📝 FAQ overhaul for linking to individual questions

### LINKED ISSUE #2268: Change FAQs in docs so we can link to individual questions

hayesall:
## Executive Summary

Fixes #2268. Individual questions can be linked to.

## Overview

- Reformat "Contents" to use the `.. contents::` directive
- Reword "Critical" into "Critical Issues"
- Reformat "Critical" section to define "critical issues" first
- Reformat FAQ sections to follow a new format (below)
- Reformat FAQ sections so individual questions have links

All sections now follow a new format (below).
A "frequently asked question" may also include a possible
cause and a solution (if the two are not obvious from
the context):

```rst
FAQ Section Title
=================

.. contents::
  :local:
  :backlinks: none

1. Question 1
-------------

**Possible Cause**: This is likely due to...

**Solution**: Fix with...
```

**Some questions were renamed** to try and make their contents more explicit. (e.g.: **Python Package: Question 1**). Many questions could begin with "I see an error message like this." Having more-explicit names could be helpful if the FAQ index gets longer, or if a person reads the index/greps for error messages.

```diff
Question 1: 

- I see error messages like this when install from GitHub using python setup.py install
+ Error: setup script specifies an absolute path when installing from GitHub using python setup.py install
```

## Alternatives / Discussion Points

All feedback is appreciated! The following are some specific points that might warrant further discussion:

- Long questions are autogenerated with long links (e.g. `FAQ.html#why-is-early-stopping-not-enabled-by-default-in-lightgbm`). An alternative might be `FAQ.html#general-12`, but this might require setting more things manually.
- Each question is a subsection. This causes the side bar to have a large number of entries in the General section (see the preview image). This has been a [known issue](https://github.com/readthedocs/sphinx_rtd_theme/issues/277) with `sphinx_rtd_theme` for some time. [An extension](https://sphinx-navtree.readthedocs.io/en/latest/) exists for limiting the depth, but would introduce a dependency.
- The current formatting makes it a little hard to tell where one section ends and another begins (e.g. R to Python, see image below).

---

### Preview Images

Screen Shot 2019-07-29 at 6 38 37 PM

**Long side bar**:

Screen Shot 2019-07-29 at 6 39 11 PM

**R/Python border**

Screen Shot 2019-07-29 at 7 01 15 PM

### DIFF
--- a/docs/FAQ.rst
+++ b/docs/FAQ.rst
@@ -1,24 +1,23 @@
 LightGBM FAQ
-============
+############
 
-**Contents**
+.. contents:: LightGBM Frequently Asked Questions
+    :depth: 1
+    :local:
+    :backlinks: none
 
--  `Critical <#critical>`__
+------
 
--  `LightGBM <#lightgbm>`__
+Critical Issues
+===============
 
--  `R-package <#r-package>`__
+A **critical issue** could be a *crash*, *prediction error*, *nonsense output*, or something else requiring
+immediate attention.
 
--  `Python-package <#python-package>`__
+Please post such an issue in the `Microsoft/LightGBM repository <https://github.com/microsoft/LightGBM/issues>`__.
 
---------------
-
-Critical
-~~~~~~~~
-
-Please post an issue in `Microsoft/LightGBM repository <https://github.com/microsoft/LightGBM/issues>`__ for any
-LightGBM issues you encounter. For critical issues (crash, prediction error, nonsense outputs...), you may also ping a
-member of the core team according to the relevant area of expertise by mentioning them with the arobase (@) symbol:
+You may also ping a member of the core team according to the relevant area of expertise by mentioning them with the
+arabase (@) symbol:
 
 -  `@guolinke <https://github.com/guolinke>`__ **Guolin Ke** (C++ code / R-package / Python-package)
 -  `@chivee <https://github.com/chivee>`__ **Qiwei Ye** (C++ code / Python-package)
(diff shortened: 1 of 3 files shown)


jameslamb:
thanks @hayesall ! I'll look more closely tomorrow

docs/FAQ.rst
```diff
@@ -47,212 +46,247 @@ When submitting issues, please keep in mind that this is largely a volunteer eff
...
---------------
+This error should be solved in latest version.
+If you still meet this error, try to remove ``lightgbm.egg-info`` folder in your Python-package and reinstall,
+or check `this thread on stackoverflow <http://stackoverflow.com/questions/18085571/pip-install-error-setup-script-specifies-an-absolute-path>`__.
 
--  **Question 2**: I see error messages like
+---
 
-   ::
+2. Error messages: ``Cannot ... before construct dataset``
```
StrikerRUS: Dot at the end is needed.

hayesall:
> In addition, can you please elaborate what exactly is required to make FAQ links short? Maybe one example?
> > Long questions are autogenerated with long links (e.g. FAQ.html#why-is-early-stopping-not-enabled-by-default-in-lightgbm). An alternative might be FAQ.html#general-12, but this might require setting more things manually.

This can be accomplished through [`:ref:` tags](https://www.sphinx-doc.org/en/1.5/markup/inline.html).

For example:

```rst
**Contents**

- :ref:`general-1`
- :ref:`general-2`

.. _general-1:

1. Where do I find more details about LightGBM parameters?
----------------------------------------------------------

(answer)

.. _general-2:

2. Another Question
-------------------

(answer)
```

All occurrences of `general-1` are replaced with the title ("1. Where do I..."), and the html hyperlink is: `FAQ.html#general-1`

hayesall:
Still failing a [`check-docs` test](https://travis-ci.org/microsoft/LightGBM/jobs/565483577), looking into it.

docs/FAQ.rst
hayesall: Should be fixed now 👍

StrikerRUS:
> Still failing a check-docs test, looking into it.

It's `linkchecker` issue. Never mind! I'll re-run the tests till it pass.

StrikerRUS:
> This can be accomplished through :ref: tags. ...


Ah, I see. Thanks!
It requires manual edition of `contents` section too each time new entry is added, right?..

docs/Installation-Guide.rst
```diff
@@ -94,7 +94,8 @@ The ``.exe`` and ``.dll`` files will be in ``LightGBM/`` folder.
 
 **Note**: You may need to run the ``cmake -G "MinGW Makefiles" ..`` one more time if you encounter the ``sh.exe was found in your PATH`` error.
 
-It is recommended to use **Visual Studio** for its better multithreading efficiency in **Windows** for many-core systems (see `FAQ <./FAQ.rst#lightgbm>`__ Question 4 and Question 8).
+It is recommended to use **Visual Studio** for its better multithreading efficiency in **Windows** for many-core systems
+(see `FAQ <./FAQ.rst#general-lightgbm-questions>`__ `Question 4 <./FAQ.rst#i-am-using-windows-should-i-use-visual-studio-or-mingw-for-compiling-lightgbm>`__ and `Question 8 <./FAQ.rst#cpu-usage-is-low-like-10-in-windows-when-using-lightgbm-on-very-large-datasets-with-many-core-systems>`__).
```
StrikerRUS: @hayesall Please remove the link from word `FAQ`. It's very nonintuitive to have two hyperlinks followed one by one, because you cannot visually distinguish where the first link is ended and the second one starts. So, one user can click on `FAQ` and will be force to scroll down for `Question 4` while another user click on `Question 4` and get the answer instantly.


Also, please update the link in Python wrapper's README in the same manner.

> Also, it'll cool if you upgrade one another remaining link to FAQ so that it'll point to individual entry.
    https://github.com/microsoft/LightGBM/blob/7da11ffeacf79be81839431b1cbc7c293ba7d9bd/python-package/README.rst#build-with-mingw-w64-on-windows
    image
   https://github.com/microsoft/LightGBM/pull/2293#pullrequestreview-268301671

hayesall:
> It requires manual edition of `contents` section too each time new entry is added, right?..

Yep.

> It's linkchecker issue.

Sphinx's `make linkcheck` is also raising an error on my local machine (below). Maybe GitHub tweaked something on their backend and now the links aren't behaving the same?

```console
Warning, treated as error:
/Users/hayesall/LightGBM/docs/FAQ.rst.rst:289:broken link:
https://github.com/microsoft/LightGBM/blob/master/python-package/README.rst#build-from-sources
(Anchor 'build-from-sources' not found)
make: *** [linkcheck] Error 2
```

StrikerRUS:
> Sphinx's make linkcheck is also raising an error on my local machine (below). Maybe GitHub tweaked something on their backend and now the links aren't behaving the same?

We are using old `LinkChecker 8.6              Copyright (C) 2000-2013 Bastian Kleineidam` linkchecker on the CI. Maybe this is the reason? Because on CI the error is about YouTube:
```
URL        `https://www.youtube.com/watch?v=iqzXhp5TxUY'
Name       `here'
Parent URL file:///home/travis/build/microsoft/LightGBM/docs/_build/html/Parallel-Learning-Guide.html, line 262, col 119
Real URL   https://www.youtube.com/watch?v=iqzXhp5TxUY
Check time 0.531 seconds
D/L time   0.568 seconds
Size       1KB
Info       Redirected to
           `https://www.google.com/sorry/index?continue=https://www.youtube.com/watch%3Fv%3DiqzXhp5TxUY&q=EgRoxoM6GKiegeoFIhkA8aeDS518C8SE-j-34lK4u8ejzutjLzF1MgFy'.
Warning    [http-empty-content] No Content
Result     Valid: 204 No Content
```

I remember we pin old version exactly due to anchors false positive errors.

docs/Installation-Guide.rst
hayesall: This should be resolved with the latest commit. I ran `grep -r "FAQ" *` and didn't find anywhere else that needed patched.

PR #2296, opened 2019-07-31
[docs] remove aliases from options in params and remove separators in FAQ

StrikerRUS:
- Remove aliases from boosting options in parameters for the consistency with other parameters.
- Remove question separators from FAQ as they consume a lot of space on the page and for brief solutions it may seem that there is no solution at all (your attention is focused on a big dash instead of an answer).
   image

### DIFF
--- a/docs/FAQ.rst
+++ b/docs/FAQ.rst
@@ -11,13 +11,11 @@ LightGBM FAQ
 Critical Issues
 ===============
 
-A **critical issue** could be a *crash*, *prediction error*, *nonsense output*, or something else requiring
-immediate attention.
+A **critical issue** could be a *crash*, *prediction error*, *nonsense output*, or something else requiring immediate attention.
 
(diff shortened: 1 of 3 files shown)


docs/FAQ.rst
```diff
@@ -182,7 +160,7 @@ An alternative, if multithreading is really necessary inside the forked sessions
 Intel toolchain. Intel compilers are unaffected by this bug.
 
 For C/C++ users, any OpenMP feature cannot be used before the fork happens. If an OpenMP feature is used before the
-fork happens (ex: using OpenMP for forking), OpenMP will hang inside the forked sessions. Use new processes instead
+fork happens (example: using OpenMP for forking), OpenMP will hang inside the forked sessions. Use new processes instead
```
jameslamb: this is a good change, I think shortened words like this can be a bit unfriendly to users for whom English is not their first language

PR #2297, opened 2019-07-31
[docs] limit number of files to scan for autosummary generation and reorganize conf.py

StrikerRUS:
Refer to https://github.com/microsoft/LightGBM/pull/2286#issuecomment-516373439.

Before:
```
D:\Users\nekit\Downloads\LightGBM\docs>make html
Running Sphinx v1.8.5
making output directory...
[autosummary] generating autosummary for: Advanced-Topics.rst, C-API.rst, Development-Guide.rst, Experiments.rst, FAQ.rst, Features.rst, GPU-Performance.rst, GPU-Targets.rst, GPU-Tutorial.rst, GPU-Windows.rst, Installation-Guide.rst, Parallel-Learning-Guide.rst, Parameters-Tuning.rst, Parameters.rst, Python-API.rst, Python-Intro.rst, Quick-Start.rst, README.rst, gcc-Tips.rst, index.rst

...
```

Now:
```
D:\Users\nekit\Downloads\LightGBM\docs>make html
Running Sphinx v1.8.5
making output directory...
[autosummary] generating autosummary for: Python-API.rst

...
```

@hayesall Can you please give your review?

**UPD:** Also, group options in `conf.py` logically.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -84,12 +84,11 @@ autodoc_default_options = {
 # Generate autosummary pages. Output should be set with: `:toctree: pythonapi/`
 autosummary_generate = ['Python-API.rst']
 
-# Add any paths that contain templates here, relative to this directory.
-templates_path = ['_templates']
+# Only the class' docstring is inserted.
+autoclass_content = 'class'
 
-# The suffix(es) of source filenames.
-# You can specify multiple suffix as a list of string:
-# source_suffix = ['.rst', '.md']
+# If true, `todo` and `todoList` produce output, else they produce nothing.
+todo_include_todos = False
 
 # The master toctree document.
 master_doc = 'index'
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -81,12 +81,12 @@ def run(self):
     "show-inheritance": True,
 }
 
+# Generate autosummary pages. Output should be set with: `:toctree: pythonapi/`
+autosummary_generate = ['Python-API.rst']
+
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ['_templates']
```
hayesall: The only change I would suggest is moving these lines back below the `templates_path` variable to keep the diffs smaller.

StrikerRUS: @hayesall Thanks for your review! I moved `autosummary_generate` variable with the aim to keep logic grouping: `autodoc_*` variables are closer to `autosummary_*` and `templates_path` is closer to `source_suffix`, so the first group is about `auto_` extensions and the second one is about source file paths.

IDK, maybe you're right and smaller diff is more important than grouping. Let's wait for one more opinion from another reviewer...

hayesall: @StrikerRUS Logical ordering is more important small diffs, I like this now that I see it this way 👍 

While we're on the subject: it doesn't appear that templates are used in this repository, so we could drop the `templates_path` variable.

StrikerRUS: @hayesall Good suggestion! Thanks! I also dropped commented option and reorganized two other options connected to extensions in the latest commit 98ebbd833410f8e3d9cd03522af23c60068c0a78.

PR #2299, opened 2019-07-31
Bug fix: small values of max_bin cause program to crash

btrotta:
Fixes a minor bug where setting `max_bin < 4` can cause lightgbm to crash. Below is an example to reproduce the bug. I think the problem is that in `bin.cpp`, the method `FindMaxBinWithZeroAsOneBin` sometimes attempts to create more bins than allowed by `max_bin`.

```
import numpy as np
import lightgbm as lgb

np.random.seed(0)
y = np.random.choice([0, 1], 100)
X = np.zeros((100, 1))
X[:30, 0] = -1
X[30:60, 0] = 1
X[60:, 0] = 2
X[0] = np.nan
params = {'objective': 'binary', 'seed': 0, 'min_data_in_leaf': 1, 'max_bin': 2}  # causes crash in training
#params = {'objective': 'binary', 'seed': 0, 'min_data_in_leaf': 1, 'max_bin': 3}  # works normally
lgb_x = lgb.Dataset(X, label=y)
est = lgb.train(params, lgb_x)

# another example, max_bin is 3 and X contains nan
X[0, 0] = np.nan
params = {'objective': 'binary', 'seed': 0, 'min_data_in_leaf': 1, 'max_bin': 3}  # causes crash
#params = {'objective': 'binary', 'seed': 0, 'min_data_in_leaf': 1, 'max_bin': 4}  # works
lgb_x = lgb.Dataset(X, label=y)
est = lgb.train(params, lgb_x)
```

### DIFF
--- a/tests/python_package_test/test_engine.py
+++ b/tests/python_package_test/test_engine.py
@@ -921,7 +921,6 @@ class TestEngine(unittest.TestCase):
         est = lgb.train(params, lgb_x, num_boost_round=5)
         np.random.seed()  # reset seed
 
-
     def test_refit(self):
         X, y = load_breast_cancer(True)
         X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.1, random_state=42)
--- a/tests/python_package_test/test_engine.py
+++ b/tests/python_package_test/test_engine.py
@@ -901,6 +901,25 @@ class TestEngine(unittest.TestCase):
         est = lgb.train(params, lgb_data, num_boost_round=1)
         self.assertEqual(len(np.unique(est.predict(X))), 3)
 
+    def test_small_max_bin(self):
+        np.random.seed(0)
+        y = np.random.choice([0, 1], 100)
+        x = np.zeros((100, 1))
+        x[:30, 0] = -1
+        x[30:60, 0] = 1
+        x[60:, 0] = 2
+        params = {'objective': 'binary',
+                  'seed': 0,
+                  'min_data_in_leaf': 1,
+                  'verbose': -1,
+                  'max_bin': 2}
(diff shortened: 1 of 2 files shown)


tests/python_package_test/test_engine.py
```diff
@@ -901,6 +901,25 @@ def test_max_bin_by_feature(self):
         est = lgb.train(params, lgb_data, num_boost_round=1)
         self.assertEqual(len(np.unique(est.predict(X))), 3)
 
+    def test_small_max_bin(self):
+        np.random.seed(0)
```
StrikerRUS: Please reset the seed before exit from this test.
https://github.com/microsoft/LightGBM/blob/04a56010ec94124da9e8c37006278a5a4b34b842/tests/python_package_test/test_engine.py#L652

btrotta:
I have changed an existing test to account for the new binning behavior. The method `FindMaxBinWithZeroAsOneBin` now will always create a zero bin if `max_bin >=3`. (I assume this is the intended behavior, based on the method name; let me know if not.) This means it needs to have bin upper bounds `-kZeroThreshold, kZeroThreshold, inf` even if there are no negative values. This reduces the number of distinct predicted values in the test I modified.

Here is a description of the current behavior of `FindMaxBinWithZeroAsOneBin` when`max_bin <= 3`:
* If `max_bin = 1`, returns `[inf]`
* If `max_bin = 2` and there are no negative data values, returns `[kZeroThreshold, inf]`
* if `max_bin = 2` and there is at least one negative data value, returns `[-kZeroThreshold, inf]`
* if `max_bin = 3`, returns `[-kZeroThreshold, kZeroThreshold, inf]`

Issue #2277, opened 2019-07-22
Support for infinite values with Nans


NicolasHug:
I'm trying to understand how the support for missing values is implemented and I came across this weird edge case, I'm not sure if it's a bug or not:

```py
import numpy as np
from lightgbm import LGBMRegressor

# 1 feature, 5 samples
X = np.array([0, 1, np.inf, np.nan, np.nan]).reshape(-1, 1)
# Easy split: all nans go to right child, the rest go to left child
y = np.array([0, 0, 0, 1, 1])

gbdt = LGBMRegressor(n_estimators=1, min_child_samples=1)
print(gbdt.fit(X, y).predict(X))
```

I'm getting
`[0.36 0.36 0.46 0.46 0.46]`, (+inf sample goes to right node)
while I would have expected
`[0.36 0.36 0.36 0.46 0.46]` (+inf sample goes to left node)

----

As far as I understand the code: the threshold for the last non-missing bin is hardcoded to be `+inf`, and this is the bin where the split happens. Since `+inf`  thresholds are capped to `1e300`, the `+inf` sample is mapped to the right child because `1e300 < +inf`

Changing the `+inf` sample to another value (e.g. 5): the threhsold for the last non-missing bin is still `1e300`, but this time `5 <= 1e300` so it will be mapped to the left node, as expected

guolinke:
I think the root cause is this PR:
https://github.com/microsoft/LightGBM/pull/617

I think a better solution is to let json support these inf/nan, not to change our model format.
Also refer to #2266

PR #2303, opened 2019-08-01
fix nan in tree model

### LINKED ISSUE #2266: JSONDecodeError: Expecting value when using lgb.plot_tree
## Issue
Hi, issue I am experiencing is very similar to issue in #1797. I can't use lgb.plot_tree as it always crashes on parsing JSON file. I have about 90 features, from which when I print out LightGBMRegressor feature_importance, about 50 are used in prediction making.

## Environment info

Operating System: Ubuntu 18.04

LightGBM version: 2.2.3 installed via pip

## Error message
This is the message I receive in logs:

```
....
  File "/usr/lib/python3/dist-packages/simplejson/decoder.py", line 370, in decode
    obj, end = self.raw_decode(s)
  File "/usr/lib/python3/dist-packages/simplejson/decoder.py", line 400, in raw_decode
    return self.scan_once(s, idx=_w(s, idx).end())
simplejson.errors.JSONDecodeError: Expecting value: line 26 column 14 (char 2057)
```

Somehow after fitting LightGBM generates invalid JSON file:

errorScreen

Why split_gain value is nan as it is invalid JSON value?
```
gbm = lgb.LGBMRegressor()
gbm_fitted = gbm.fit(X_train, y_train,
                         eval_set=[(X_val, y_val)],
                         eval_metric='rmse',
                         early_stopping_rounds=20, feature_name=column_names)
... plotting features, importances...
lgb.plot_tree(gbm_fitted) # <- this is where I receive error
```
Interesting part is that XGBoost plot_tree with same data after training is able to plot decision tree perfectly. 

What could I do to make LGBM plot this tree?

guolinke:
to fix #2266

also refer to #2277 
I think a better solution is let json support inf and nan, not to chang model content.

### DIFF
--- a/include/LightGBM/tree.h
+++ b/include/LightGBM/tree.h
@@ -422,7 +422,7 @@ inline void Tree::Split(int leaf, int feature, int real_feature,
   split_feature_inner_[new_node_idx] = feature;
   split_feature_[new_node_idx] = real_feature;
 
-  split_gain_[new_node_idx] = Common::AvoidInf(gain);
+  split_gain_[new_node_idx] = gain;
   // add two new leaves
   left_child_[new_node_idx] = ~leaf;
   right_child_[new_node_idx] = ~num_leaves_;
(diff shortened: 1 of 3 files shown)


guolinke:
The new solution, only converting inf/nan when dump json.

PR #2305, opened 2019-08-01
Less restrictive monotone constraints

CharlesAuguste:
+@aldanor @redditur

Currently, LightGBM is too strict in the way it enforces monotonic constraints. This PR introduces two methods; one just as computationally efficient as the current implementation, and another which is up to twice as slow, which enforces monotonic constraints in a looser way thereby improving performance of the resulting model.

Moreover, we introduce a novel way of penalizing monotonic splits which aims to reduce the loss in performance brought about by greedily splitting at the top of a decision tree with disregard to the constraints that this split may impose lower down the tree.

In the attached report we show that these methods can produce models with significantly improved performance and, therefore, recommend replacing the current approach LightGBM takes to enforcing monotonic constraints. It contains some details about the methods we implemented, and why we think they should replace the current one, including some results we generated on a public dataset to compare the methods. Please let us know if you find something unclear in the report or if you would like more information about anything.

**Broken files**
About the implementation, the monotone constraints should be fully working in the tree learner serial_tree_learner.cpp and serial_tree_learner.h. However, we barely modified the rest of the tree learners. Therefore, the following files will compile, but are broken:

- feature_parallel_tree_learner.cpp
- voting_parallel_tree_learner.cpp
- parallel_tree_learner.h

And the following files won’t compile and broken too:

- gpu_tree_learner.h
- gpu_tree_learner.cpp

If you agree to go forward with the pull request, we think that given the work we have already done, it shouldn’t be too hard to adapt everything to the other tree learners. However, as we are not familiar with them, and would represent a lot of work for us alone so we hoped you would help for the implementation.
We think there are 3 options:

- If you do not wish to work on monotone constraints, but still want to merge our work, then we can just make sure that the other tree learners will work with the current LightGBM constraining method, as they currently do. This can easily be done by copy-pasting a few functions in feature_histogram.hpp, and adding back what we removed from the files leaf_splits.hpp and split_info.hpp, but that would not be very clean. If you wish to go with this option, please advise what you think the best way to go with this is;
- If you are ready to work with our code, but don't want to spend too much time on it, then we can try to make the tree learners work for with the fast method only;
- If you are ready to work with our code and are ready to spend some time on it, when we can try to make both our slow and fast methods work for all tree learners.

**Interactions between parameters**
Finally, here is the summary of how the new parameters interact with the existing program:

- When there are no constraints, the program should behave the same with or without the code we added;
- For “monotone_precise_mode” to be set to True or for "monotone_penalty" to be non-0, actual constraints have to exist;
- When there are constraints and “monotone_precise_mode” is set to False, then the way of handling missing values and categorical features is the same as before;
- When there are constraints and “monotone_precise_mode” is set to True, then the missing value handle has to be deactivated (we believe that it would be possible to make it working, but we did not work on it because it is of no interest to us). Also, even if “monotone_precise_mode” is set to True, the categorical features won’t be split using the smart method (we believe that it should be possible to modify FindBestThresholdCategorical in a very similar way to what we did with FindBestThresholdNumerical in feature_histogram.hpp, but once again this is of no interest to us).

[PR-monotone-constraints-report.pdf](https://github.com/microsoft/LightGBM/files/3457826/PR-monotone-constraints-report.pdf)


CharlesAuguste:
Has anyone started to / is anyone planning to look into this PR? 

It took us some time to come up with and code our methods in LightGBM, and we do think this can be a major improvement (to those using monotone constraints at least) for the library. So we really are serious about this PR.

This might be a bit heavier than the usual PR, so please let me know if you want me to clarify anything, or if you have ideas on how to improve the code. Thanks!

aldanor:
@StrikerRUS @guolinke Wondering what are your thoughts about this?

We think this is a novel method of handling monotone constraints, more efficient than in any of the existing GBM libraries including lightgbm/xgboost/catboost. We chose LightGBM because that’s what we currently use most frequently.

The presented implementation is backed by months of extensive research, testing and benchmarking, both on public and private data.

If we decide to go forward with this, here are the steps:

- Someone needs to get familiar with core concepts and ideas presented in the pdf
- Someone needs to read through the implementation of these ideas for serial tree learner
- For non-serial learners, there’s two options: leave the old method and is, or implement the new method(s) for all of them. This will be mostly copy/paste but will need to be done preferably by someone very familiar with those parts of code, like gpu learner.

We’d be glad to discuss any of the above or provide any help with familiarising reviewers with this PR.

StrikerRUS:
@aldanor I think this is great contribution! I'm so sorry, but this is beyond the area of my expertise. The appropriate pings will be @guolinke and @chivee. And of course, everyone interested in this is welcome to the discussion!

guolinke:
Sorry, i am on vacation. Will check this when have time.

guolinke:
@CharlesAuguste @aldanor 
We are very welcome for this PR! Please go head, and ping me directly if you need any helps.

include/LightGBM/tree.h
```diff
@@ -402,12 +427,22 @@ class Tree {
   std::vector<int> leaf_depth_;
   double shrinkage_;
   int max_depth_;
+  // add parent node information
+  std::vector<int> node_parent_;
+  // Keeps track of the monotone splits above the leaf
+  std::vector<bool> leaf_is_in_monotone_subtree_;
```
guolinke: do we really need to store these states into tree model? Or other structures only for MC? 
Since some users don't use MC, this will change the model format and cause the compatibility problem.

guolinke:
And I think we can have the fast method in this PR. And have another PR for the slow method, after this PR is merged.

include/LightGBM/tree.h
guolinke: another suggestion: it seems there are lots of changes in FeatureHistogram. This will hurt the readability of current codes. Could we decouple MC with other parts? Like an independent class, which could be called in FeatureHistogram.

redditur: > This will hurt the readability of current codes. Could we decouple MC with other parts?

We were very careful to implement this in a way that limited the impact on readability and, in my opinion, I think that we did a good job in this regard, given the complexity of what's being implemented. In fact, in my opinion, the current implementation of monotonic constraints isn't particularly readable. If anything, we've made it more readable, although, obviously, more complex.

If you have any specific examples where you feel readability has been impacted let us know and we can discuss it on an example-by-example basis.

> Could we decouple MC with other parts?

We could, but I'm not convinced this would make things substantially more readable, and it would be some effort on our part to make such changes. It could even hurt readability due to increased indirection.

guolinke: Yeah, actually I mean the maintainability. All implementation of MC is in one place, not distributed in the project. And i think good maintainability is also more readable.

CharlesAuguste: I will not be able to work on this during this week and next week, sorry about that. But I will make the necessary changes to improve the code after that.

mayer79:
@guolinke: is it possible to keep also your current mc implementation? It e.g. ensures monotonic partial dependence profiles, which is sweet.

CharlesAuguste:
Sorry for the delay, I should be able to work on this again in the next few days. I can indeed split this PR. However, there are some code and structures that we created because of the slow method, but that are now used by every method. For example, we have a vector of doubles for determining constraints when splitting instead of just a double, because, for the slow method, constraints can depend on where the split happens. For readability, every method uses this vector of constraints, with the vector being only one element for the fast method for example. Can I leave these structures for the PR of the fast method? Or do I have to rework the code to make it consistent for the fast method only?

include/LightGBM/tree.h
CharlesAuguste: @guolinke about the new states in the Tree class which would cause compatibility issues. First, the states are absolutely necessary to the functioning of monotone constraints as we implemented them, there is no way we can do without. And I think that these are sensible states that should belong to the tree class (I could even see these reused for other purposes). 

An alternative I can think about would be to create an alternative class which would look like tree, but with these states only, and this class would be used whenever the Tree class is currently used. But that would create some redundancies, and would probably not be good for readability. 

Since these fields currently won't be used if monotone constraints are not used, is it not possible to solve the compatibility problem instead? I am not sure when these issues would arise, but maybe there is a way to specify somewhere that if format don't match, then these fields should remain void, or be filled with dummy values. What do you think?

guolinke: @CharlesAuguste I see. could we have these states in tree class, but not save and load them into model file? If this is possible, it is okay as well.

guolinke: just double-check the code, it seems the new states are not in model file. So I think it is okay.

guolinke:
@CharlesAuguste 
You can use the convenient way for you. 
Also refer to #2341, 
I plan to separate monotone constraints, forced splits and cost efficient splits from the tree_learner. 
I think your PR could help for the `monotone constraints` .

guolinke:
@CharlesAuguste 
you can follow the style of pr #2407 , which refactors the cost effective gradient boosting.

CharlesAuguste:
@guolinke I was able to do some refactoring. Can you let me know if this is going in the right direction? I am having trouble writing clean and understandable code, but I am ready to follow your advice to improve the code of the PR. Moreover, I removed some functions related only to the Slow method, to make this PR simpler, as you requested. 

There still is a piece of code that I don't know how to refactor; the functions "GoUpToFindLeavesToUpdate" and "GoDownToFindLeavesToUpdate". The functions are used exclusively for monotone constraints, so I think you want me to move them to another file. However, when the code runs with monotone constraints, the stack of function calls goes like  "Split -> GoUpToFindLeavesToUpdate -> GoDownToFindLeavesToUpdate -> UpdateBestSplitsFromHistograms -> ComputeBestSplitForFeature". The functions Split and ComputeBestSplitForFeature (at the start and end of the stack) absolutely belong to serial_tree_learner.cpp, and should stay in the file. My functions are called only in the presence of monotone constraints, and I do not know what the best way to move them somewhere else is, as they use parameter and methods from the SerialTreeLearner class. Can you please let me know how you would like me to do that? Please do let me know if something is unclear, and what else I can do to refactor the code and make it better. Thanks!

src/treelearner/serial_tree_learner.cpp
```diff
@@ -23,8 +23,11 @@ std::chrono::duration<double, std::milli> hist_time;
 std::chrono::duration<double, std::milli> find_split_time;
 std::chrono::duration<double, std::milli> split_time;
 std::chrono::duration<double, std::milli> ordered_bin_time;
+std::chrono::duration<double, std::milli> refit_leaves_time;
 #endif  // TIMETAG
 
+double EPS = 1e-12;
```
guolinke: these is a `kEpsilon` you can use directly.

guolinke:
Thanks so much! It is much clearer now. Please ping me if you met any problems. 
I quick through the code of GoUpToFindLeavesToUpdate  and GoDownToFindLeavesToUpdate. It seems there are kinds of static methods, and related to MC. Maybe move them to the MC class as the static methods?

CharlesAuguste:
I indeed think it is a good idea to consider them as static methods of MC, but as I mentioned, they also need access to some non-static methods of SerialTreeLearner where they currently are. So should I add SerialTreeLearner as an argument of  GoUpToFindLeavesToUpdate and GoDownToFindLeavesToUpdate for them to be able to call the SerialTreeLearner methods, or is that not good?

guolinke:
Do you mean the `UpdateBestSplitsFromHistograms` ? I only find this method is called in these two functions.
Is that possible to move `UpdateBestSplitsFromHistograms` out of this function?

guolinke:
BTW, we don't want to involve additional overhead when the user doesn't enable MC.
So, like we do in #2407, when the function is disabled, the nullptr is used to avoid the unnecessary runs.

CharlesAuguste:
I meant GoUpToFindLeavesToUpdate calls GoDownToFindLeavesToUpdate which calls UpdateBestSplitsFromHistograms which calls ComputeBestSplitForFeature. I created all 4 methods. The first 3 are only about monotone constraints and do not belong to this file. The 4th one (ComputeBestSplitForFeature) is just basically some code that was duplicated in the version of master LightGBM I pulled, so I encapsulated the duplicated code into that method, so it does belong in the serial_tree_learner.cpp file, and is used even when there are no monotone constraints.

Currently the first 3 methods can easily call the 4th one because they all belong to the same class. However, if I move the first 3 somewhere else and make them static, then they still need to call the 4th one, and I am not sure what the best way to do that is. I think I can pass SerialTreeLearner as an argument of these methods while making them static, but I am not sure if that is a good practice.

guolinke:
@CharlesAuguste 
I don't think it is a good practice too. It is better to avoid passing a class and edit its contents by its non-constant method.

Could you move the `ComputeBestSplitForFeature` to the MC, with some arguments from tree_learner?

guolinke:
@CharlesAuguste it seems you can change the `ComputeBestSplitForFeature` to the const function, even the static function. In that cases, I think it is safe to call it.

CharlesAuguste:
@guolinke I was able to work a bit more on the PR. Hopefully you will find the code clearer now. I am not sure where exactly you want me to use the nullptr to disable things when monotone constraints are not used. Can you point where you want me to make the changes in the code? Let me know what you think and how I can improve this PR. Thanks

CharlesAuguste:
I brought the branch back up to date with the current LightGBM.

CharlesAuguste:
@guolinke What do you think of the latest changes? What else should I add to the code to improve the pull request? Thanks!

guolinke:
@CharlesAuguste Thanks! I am quite busy recently. I will take a look at this week.

include/LightGBM/dataset.h
```diff
@@ -586,6 +586,15 @@ class Dataset {
     return bufs;
   }
 
+  // This function retrieves the number of bins for a specific feature
+  int NumBin(int feature_idx) const {
```
guolinke: duplicated with `FeatureNumBin`

src/treelearner/cost_effective_gradient_boosting.hpp
```diff
@@ -20,6 +20,8 @@ namespace LightGBM {
 
 class CostEfficientGradientBoosting {
  public:
+  std::vector<SplitInfo> splits_per_leaf_;
```
guolinke: why change this to public?

src/treelearner/feature_histogram.hpp
```diff
@@ -80,44 +85,61 @@ class FeatureHistogram {
...
     output->gain *= meta_->penalty;
   }
 
-  void FindBestThresholdNumerical(double sum_gradient, double sum_hessian, data_size_t num_data, double min_constraint, double max_constraint,
-                                  SplitInfo* output) {
+  void FindBestThresholdNumerical(double sum_gradient, double sum_hessian,
+                                  data_size_t num_data, SplitInfo *output,
+                                  SplittingConstraints &constraints) {
     is_splittable_ = false;
+    could_be_splittable_ = false;
```
guolinke: what differences between `is_splittable_` and `could_be_splittable_ ` ?

src/treelearner/feature_histogram.hpp
```diff
@@ -80,44 +85,61 @@ class FeatureHistogram {
...
-    output->min_constraint = min_constraint;
-    output->max_constraint = max_constraint;
   }
 
-  void FindBestThresholdCategorical(double sum_gradient, double sum_hessian, data_size_t num_data,
-                                    double min_constraint, double max_constraint,
-                                    SplitInfo* output) {
+  void FindBestThresholdCategorical(
+      double sum_gradient, double sum_hessian, data_size_t num_data,
+      SplitInfo *output, SplittingConstraints& constraints) {
```
guolinke: I think categorcial features don't need the constraints, for there are not numerical ordering relations.

src/treelearner/feature_histogram.hpp
```diff
@@ -489,6 +529,22 @@ class FeatureHistogram {
     return ret;
   }
 
+ private:
+  static double GetSplitGains(double sum_left_gradients, double sum_left_hessians,
+                              double sum_right_gradients, double sum_right_hessians,
+                              double l1, double l2, double max_delta_step,
+                              double min_constraint_right, double max_constraint_right,
+                              double min_constraint_left, double max_constraint_left, int8_t monotone_constraint) {
```
guolinke: I think it is clearer to pass `constraints` directly, not its many members.

src/treelearner/feature_histogram.hpp
```diff
@@ -80,44 +85,61 @@ class FeatureHistogram {
     }
   }
 
-  void FindBestThreshold(double sum_gradient, double sum_hessian, data_size_t num_data, double min_constraint, double max_constraint,
-                         SplitInfo* output) {
+  void FindBestThreshold(double sum_gradient, double sum_hessian,
+                         data_size_t num_data, SplitInfo *output,
+                         SplittingConstraints &constraints) {
```
guolinke: use pointer type for `constraints`, so that we could just run it only when it is not `nullptr`.

src/treelearner/feature_histogram.hpp
```diff
@@ -582,6 +670,8 @@ class FeatureHistogram {
         t = -1;
       }
 
+      constraints.InitializeIndices(dir);
```
guolinke: `constraints` is not the reay-only object?

guolinke: could we pass `dir` to  `CurrentMinConstraintRight` (and other similar function), to make `constraints` read-only?

src/treelearner/feature_histogram.hpp
```diff
@@ -505,12 +561,23 @@ class FeatureHistogram {
...
 
     double best_sum_left_gradient = NAN;
     double best_sum_left_hessian = NAN;
+
+    // when the monotone precise mode is enabled, then the left and the right children may not
+    // have the same min and max constraints because constraints can depend on the thresholds
+    double best_min_constraint_left = NAN;
+    double best_max_constraint_left = NAN;
+    double best_min_constraint_right = NAN;
+    double best_max_constraint_right = NAN;
```
guolinke: Use a structure or class to save these states?

src/treelearner/serial_tree_learner.h
```diff
@@ -180,6 +189,13 @@ class SerialTreeLearner: public TreeLearner {
   std::vector<int> ordered_bin_indices_;
   bool is_constant_hessian_;
   std::unique_ptr<CostEfficientGradientBoosting> cegb_;
+
+  std::vector<std::vector<double> > dummy_min_constraints;
+  std::vector<std::vector<double> > min_constraints;
+  std::vector<std::vector<double> > dummy_max_constraints;
+  std::vector<std::vector<double> > max_constraints;
```
guolinke: where use these ?

src/treelearner/serial_tree_learner.h
```diff
@@ -180,6 +189,13 @@ class SerialTreeLearner: public TreeLearner {
...
+
+  CurrentConstraints current_constraints;
```
guolinke: `current_constraints` may need to sync among machines in distributed learning, refer to `min_constraint` and `max_constraint` in SplitInfo.

guolinke:
@CharlesAuguste Thanks very much. I think the current framework is much better, we can base on that for the improvement.

BTW, for the nullpty problem, you can check https://github.com/microsoft/LightGBM/pull/2407/files#diff-110aae7c40f3150eb5d8264c1cb0663eR136-R138

CharlesAuguste:
@guolinke Thank you for your review, I will start making changes this week.

src/treelearner/cost_effective_gradient_boosting.hpp
CharlesAuguste: I need to access it in my function UpdateBestSplitsFromHistograms in monotone_constraints.cpp. Should I write a getter instead to make it cleaner?

src/treelearner/feature_histogram.hpp
CharlesAuguste: In the fast method (which the only thing we are looking at right now because I removed the more complicated Slow method for the moment), could_be_splittable is not used. This can be confirmed by looking at the return value of the is_splittable() function.

In the Slow method, if we decide to add it and push it, could_be_splittable is used because sometimes a split in the tree can unconstrain a leaf. This way, a leaf that was not splittable before can become splittable. Because of this, in the Slow method, we do not want to exclude for good some leaves that are not splittable right now, because they may become splittable later when the tree keeps getting built. 

Is that clear? Thanks

src/treelearner/feature_histogram.hpp
CharlesAuguste: There are 3 things I want to highlight regarding this function. 

First from a code point of view, in the same file, the function find_best_threshold_fun_ is a bind for both FindBestThresholdCategorical and FindBestThresholdNumerical so these 2 functions need to have the same template. So I would have to change a significant amount of code to make the 2 not have the same template.

Second, in theory, Say we have a dataset with 2 features, 1 being categorical and 1 numerical, and we apply a monotonic constraint on the numerical one. Then even when when we split the data with respect to the categorical feature, then the best constraints for the resulting leaves may vary depending on which categories you split the categorical feature on. For example if you look at example 11 of the pdf report I posted in my initial comment, then you can imagine that the unconstrained feature is actually a categorical feature, and you see that the constraints applied to the leaves also depend on the categorical splits. Currently this is not implemented in the code, and I don't plan on implementing it in this PR, because it has a pretty marginal use. But I or someone else may, one day, want to implement that and the current way the code works is perfect for that.

Finally, there are some research papers that want to apply monotone constraints to categorical data (e.g. credit ratings, yes-no questions...). So maybe one day someone will want to make a PR implementing monotone constraints for categorical features.

For these reasons, I don't agree with removing the constraints argument from the categorical features splitting function you mentioned. But if you still think that's best in light of the above then I will do it.

src/treelearner/serial_tree_learner.h
CharlesAuguste: My bad, these were used before refactoring, and I forgot to remove them.

src/treelearner/cost_effective_gradient_boosting.hpp
guolinke: yeah, setter and getter would be better.

src/treelearner/feature_histogram.hpp
guolinke: yeah, you can keep these function arguments. Actually, we will have some upgrade for the categorical features (refer to #2463), and the old categorical split will be abandoned.

src/treelearner/feature_histogram.hpp
guolinke: okay, I see. Could you move `could_be_splittable_`  into "constraints" object? It will relief the confuse in `is_splittable_ ` and `could_be_splittable_`.

src/treelearner/serial_tree_learner.h
CharlesAuguste: Could be more specific about distributed learning? Our method already supports having several threads, because leaves are treated separately, and each leaf gets associated to 1 array (we use "int tid" to identify what thread we are in).

guolinke: I mean the multi-machine, refer to the states you delete in SplitInfo. The SplitInfo will be sync in different nodes.

CharlesAuguste: I am sorry but I don't understand. My understanding is that the object current_constraints contains splitting_constraints_vector which is an array which contains as many elements as there are threads. Each element of the array is used in one specific thread, and does not need to communicate with the others because the computations for 1 leaf is made by at most 1 thread. 

Can you be more specific about the part of SplitInfo that is relevant? Is it relevant with all tree learners, or only with some? I have only been using serial_tree_learner, so I don't really know a lot about the others. Thanks

guolinke: Refer to the DataParallelTreeLearner or other parallel learners, these learn will find the best split in different feature sets, over different machines. And these states in different machines cannot be accessed directly. We need to sync them explicitly. And SplitInfo will be synced in these parallel learners. So the min_constraints, and other states you delete in SplitInfo, is not synced anymore..

CharlesAuguste: Okay, so if I am following correctly, I need to make a functions like SyncUpGlobalBestSplit, CopyFrom, CopyTo (from splitInfo), but for my variables constraints_per_leaf_ which is an array of LeafConstraints is that right? This variable contains the constraints for all leaves in the tree, and I think it is the only one that needs to be shared. Do you agree?

src/treelearner/feature_histogram.hpp
CharlesAuguste: Up on this in case you didn't see it @guolinke.

src/treelearner/serial_tree_learner.h
guolinke: yes, the easiest way is to save the constraints into SplitInfo, then you don't need to care these. The splitInfo is per leaf too, so you can only save the information of current leaf to it, and get it back to your `constraints_per_leaf_`.

src/treelearner/monotone_constraints.hpp
```diff
@@ -0,0 +1,612 @@
...
+namespace LightGBM {
+
+struct CostEfficientGradientBoosting;
+struct CurrentConstraints;
+class HistogramPool;
+struct LeafConstraints;
+
+struct LearnerState {
+  const Config *config_;
+  std::unique_ptr<DataPartition> &data_partition_;
```
guolinke: these reference are dangerous, could you make them const?

guolinke: And I think you can use `const T* ` directly.

src/treelearner/feature_histogram.hpp
```diff
@@ -240,16 +276,31 @@ class FeatureHistogram {
...
+      double current_min_constraint_left = -std::numeric_limits<double>::max();
+      double current_max_constraint_left = std::numeric_limits<double>::max();
+      double current_min_constraint_right = -std::numeric_limits<double>::max();
+      double current_max_constraint_right = std::numeric_limits<double>::max();
+      if (constraints != nullptr) {
+        current_min_constraint_left = constraints->CurrentMinConstraintLeft();
+        current_max_constraint_left = constraints->CurrentMaxConstraintLeft();
+        current_min_constraint_right = constraints->CurrentMinConstraintRight();
+        current_max_constraint_right = constraints->CurrentMaxConstraintRight();
+      }
```
guolinke: could we avoid these expansions?  maybe have two structures for two leaves, e.g. 
```constraints->left->min```, ```constraints->left->max``` ,```constraints->right->min```, ```constraints->right->max```.
Then pass `constraints->left` and `constraints->right` to CalculateSplittedLeafOutput, respectively.

File: src/treelearner/serial_tree_learner.h (hunk 12 above)
Comment (CharlesAuguste) [reply src/treelearner/serial_tree_learner.h]: I tried to look a bit at that part of the code but I am very unfamiliar with it. Could you explain to me how the multi-machine works overall? For example what tasks are split among machines, why we need to sync up SplitInfo, but not other variables... Also, it seems like only 2 SplitInfo are being sync'ed across machines (smaller_best_split and larger_best_split). However, in our constraining methods, a split on a leaf can generate an update in the constraints of many other leaves, so I think I still need to sync all constraints. Overall this task seems pretty difficult for me, do you think you could provide help for the actual coding? Thanks!

Comment (guolinke) [reply src/treelearner/serial_tree_learner.h]: I see. Maybe we can ignore this part first. Let us focus on other parts first.

Comment (CharlesAuguste) [reply src/treelearner/serial_tree_learner.h]: Okay, I think I pushed changes to answer about all your other comments, let me know what else I can improve!

**Comment by aldanor** [comment]
@guolinke Is there any way we can progress with this? It's going to require continuous rebasing just to keep up with the current master branch. 

// I'm not sure we can do much more than what is currently done here, with distributed learning we'll require external help since we're not well familiar with how it works - or alternatively the new improved method would have to be disabled there completely in favor for the old one.

**Comment by guolinke** [comment]
@aldanor sorry, I am being very busy recently. I think a temporal solution is to disable current changes in distributed mode. That is, if mc is set in parallel tree learner, you can throw the error directly. 
I will help you for the distributed code after the serial version is ready.

File: include/LightGBM/tree.h [hunk 15]
```diff
@@ -80,9 +80,14 @@ class Tree {
...
-                       const uint32_t* threshold, int num_threshold, double left_value, double right_value,
-                       int left_cnt, int right_cnt, double left_weight, double right_weight, float gain, MissingType missing_type);
+
+  int SplitCategorical(int leaf, int feature, int real_feature,
+                       const uint32_t *threshold_bin, int num_threshold_bin,
+                       const uint32_t *threshold, int num_threshold,
+                       double left_value, double right_value, int left_cnt,
+                       int right_cnt, double left_weight, double right_weight,
+                       float gain, MissingType missing_type,
+                       bool feature_is_monotone);
```
guolinke: is `feature_is_monotone` needed?

src/io/config_auto.cpp
```diff
@@ -476,6 +494,10 @@ void Config::GetMembersFromString(const std::unordered_map<std::string, std::str
 
   GetBool(params, "use_missing", &use_missing);
 
+  GetBool(params, "monotone_precise_mode", &monotone_precise_mode);
+  CHECK(!monotone_precise_mode || !use_missing);
+  CHECK(!monotone_precise_mode || constraints_exist);
+
```
guolinke: I think this file is auto generated, you cannot edit it by hand.
You could have some post-processing in `config.cpp`.

src/treelearner/feature_histogram.hpp
```diff
@@ -135,6 +158,10 @@ class FeatureHistogram {
     int best_threshold = -1;
     int best_dir = 1;
 
+    if (constraints != nullptr) {
+      constraints->InitializeIndices(1);
+    }
```
guolinke: I don't think we need constraints in this function, you can pass it, but don't need to call it.

src/treelearner/feature_histogram.hpp
```diff
@@ -240,16 +276,29 @@ class FeatureHistogram {
...
-                                                        min_constraint, max_constraint);
+      SplittingConstraint *left;
+      SplittingConstraint *right;
+      if (constraints != nullptr) {
+        left = &(constraints->left);
+        right = &(constraints-> right);
+      }
+      else {
+        left =  new SplittingConstraint();
+        right =  new SplittingConstraint();
```
guolinke: maybe initialize `left` and `right` to nullptr? and handle the nullptr in `CalculateSplittedLeafOutput` function?

src/treelearner/feature_histogram.hpp
```diff
@@ -457,38 +506,49 @@ class FeatureHistogram {
...
+                              const SplittingConstraints *constraints, int8_t monotone_constraint) {
+    const SplittingConstraint *left;
+    const SplittingConstraint *right;
+    if (constraints != nullptr) {
+      left = &(constraints->left);
+      right = &(constraints->right);
+    }
+    else {
+      left =  new SplittingConstraint();
+      right =  new SplittingConstraint();
```
guolinke: maybe initialize left and right to nullptr? and handle the nullptr in CalculateSplittedLeafOutput function?

src/treelearner/serial_tree_learner.cpp
```diff
@@ -521,10 +537,15 @@ void SerialTreeLearner::ConstructHistograms(const std::vector<int8_t>& is_featur
...
-void SerialTreeLearner::FindBestSplitsFromHistograms(const std::vector<int8_t>& is_feature_used, bool use_subtract) {
+void SerialTreeLearner::FindBestSplitsFromHistograms(
+    const std::vector<int8_t> &is_feature_used, bool use_subtract,
+    const Tree *tree) {
   #ifdef TIMETAG
   auto start_time = std::chrono::steady_clock::now();
   #endif
+    LearnerState learner_state(config_, data_partition_.get(), train_data_,
+                               constraints_per_leaf_, tree, current_constraints,
+                               cegb_);
```
guolinke: can this only be called one-time? not everytime in `FindBestSplitsFromHistograms`.

guolinke: BTW, could we avoid the extra calls in serial tree learner, when mc is not sed?

src/treelearner/serial_tree_learner.cpp
```diff
@@ -817,41 +841,61 @@ void SerialTreeLearner::Split(Tree* tree, int best_leaf, int* left_leaf, int* ri
...
+
+  // when the monotone precise mode is disabled it is very easy to compute the constraints of
+  // the children of a leaf, but when it is enabled, one needs to go through the tree to do so,
+  // and it is done directly before computing best splits
+  if (!config_->monotone_precise_mode) {
+    LeafConstraints::SetChildrenConstraintsFastMethod(
+        constraints_per_leaf_, right_leaf, left_leaf,
+        best_split_info.monotone_type, best_split_info.right_output,
+        best_split_info.left_output, is_numerical_split);
+  }
```
guolinke: will this be called when mc is not set?

src/treelearner/serial_tree_learner.cpp
```diff
@@ -817,41 +841,61 @@ void SerialTreeLearner::Split(Tree* tree, int best_leaf, int* left_leaf, int* ri
...
+
+  // if there is a monotone split above, we need to make sure the new
+  // values don't clash with existing constraints in the subtree,
+  // and if they do, the existing splits need to be updated
+  if (tree->leaf_is_in_monotone_subtree(*right_leaf) && !config_->monotone_constraints.empty()) {
```
guolinke: may be check `config_->monotone_constraints.empty()` first

src/treelearner/serial_tree_learner.cpp
```diff
@@ -892,4 +936,56 @@ void SerialTreeLearner::RenewTreeOutput(Tree* tree, const ObjectiveFunction* obj
...
+    double penalty = LeafConstraints::ComputeMonotoneSplitGainPenalty(
+        depth, learner_state.config_->monotone_penalty);
+    new_split.gain *= penalty;
+  }
+
+  if (new_split > bests[tid]) {
+    bests[tid] = new_split;
+  }
+}
+
```
guolinke: I think we should avoid the extra calls in serial_tree_learner and feature_histogram as possible.

CharlesAuguste:
@guolinke I will work on that over the next few days! Thanks

include/LightGBM/tree.h
CharlesAuguste: Right now it is not needed as we don't support categorical features. But that makes it more consistent with the standard split, and in the future maybe we will make categorical variables supported for monotonic constraints. Should I remove it?

src/io/config_auto.cpp
CharlesAuguste: I am not sure I understand, I locally tested if these checks worked and they did. What do you mean by the file being auto generated? I am not very familiar with how the library is built I am sorry.

src/treelearner/serial_tree_learner.cpp
CharlesAuguste: Yes, this is still the same logic as before the PR was made. When there are no monotone constraints, then the max constraint for a leaf is +inf and the min constraints is -inf.

src/treelearner/serial_tree_learner.cpp
CharlesAuguste: > can this only be called one-time? not everytime in `FindBestSplitsFromHistograms`.

@guolinke I don't really understand what you mean here. If I am not mistaken the Learner state class is created once an only once every time we look for splits, so in a tree this will be called as many times as there are splits. This also a very fast call, because it is just setting references. So I don't understand why we would want to call it less than that. If we do, I as least need to call it once per tree, but then we would have "duplicated pointers" (pointers both in SerialTreeLearner and in LearnerState pointing to the same objects like config, data_partition_ or tree...) all the time, which I am not sure is a good idea. In light of this can you please clarify how you would like the code to be modified? Thanks  
@aldanor can you help me with that also if you have a good idea of how to improve the code?

CharlesAuguste: > BTW, could we avoid the extra calls in serial tree learner, when mc is not sed?

Given how we transformed the code now, I don't think we can avoid this call, otherwise we would have to duplicate the functions that use learner_state. Do you have a specific idea in mind of how you would like this to work?

src/treelearner/serial_tree_learner.cpp
CharlesAuguste: I am sorry, but I don't understand what functions in particular you are referring to. I think the code, right now, when there are no constraints, is working in a very similar way to they way it used to work. But I am also happy to make the changes you request if you can be more specific about this.

CharlesAuguste:
@guolinke @shiyu1994 I did the changes I could on @guolinke 's request, but I need more guidance on the one I commented. Please let me know how to improve the PR. Thanks

StrikerRUS:
Just out of curiosity: can this PR be split into sequence of smaller PRs? It's very hard to review PRs with a such large diff size. If this PR is splitable, splitting it will increase the speed and quality of reviews and in consequence speed up merging the proposed changes. As a good example I can refer to https://github.com/dmlc/xgboost/issues/5104.

CharlesAuguste:
You are right, I indeed think it would make sense to split this PR into smaller PRs. This will be require a significant effort but we do think this PR is a valuable addition to the library (and hope you agree with us on that!), and I understand that such a big PR is very hard to review for you so I am ready to spend more time on it. Here is how I could proceed: 

- 1st PR with just refactoring the c++ code. In this PR, I would change the structure around monotone constraints to have them as close as possible to the structures this PR. In this PR, the behavior of the library would not be changed at all, just some refactoring would be done (for example adding the files monotone_constraint.cpp and monotone_constraints.hpp);
- 2nd PR with just the algorithm for the Fast method added. In this PR I would implement the simplest algorithm, and try not to over-complicate it with stuff belonging to the Slow method. I will try to make as little changes as possible to feature_histogram.hpp in this pull request.
- 3rd PR with the monotone penalty. This PR should be pretty short, just a few simple functions, and a new parameter for the library users.
- 4th PR with the Slow method implemented on top of the existing code.

Do you agree with the plan above. Do you see any way to improve it? Let me know what you think. I will start working on that in the next few days. Thanks

jameslamb:
@CharlesAuguste thank you again for all your hard work on this! Now that we have #3264, can this pull request be closed?

StrikerRUS:
@jameslamb 

> Now that we have #3264, can this pull request be closed?

According to the following @CharlesAuguste's [words](https://github.com/microsoft/LightGBM/pull/3264#issue-458554579), 

> 2 things in the original PR that are missing in here:
> - The code was more optimized for time in the original PR, and the method is a bit slower here. Though the optimizations made the code significantly more complicated, and I think the time saved was not much. I can implement the optimization in another PR later, as I think this PR is complicated enough as is.
> - In the original PR, once a tree is build, there was a final leaves refitting, that I have not implemented here yet. I can either implement it here or in another PR later. This can slightly improve performance.

there maybe one or two more PRs related to the monotone constraints.

jameslamb:
@StrikerRUS I understand there may be more PRs in the future, but that doesn't mean this one needs to stay open, right? If we're never going to merge this specific PR, it should just be closed.

StrikerRUS:
@jameslamb 
Yeah, maybe you are right! I was thinking about this PR as about "TODO issue" because all changes from those `PR1, PR2, ...` are implemented here originally.

CharlesAuguste:
@StrikerRUS @jameslamb I confirm this PR will not be merged. So I agree it can be closed. Thanks

jameslamb:
> @StrikerRUS @jameslamb I confirm this PR will not be merged. So I agree it can be closed. Thanks

thanks very much!

Issue #2163, opened 2019-05-09
Doc should mention that is_unbalance is relevant for multiclassova


rauhs:
It seems that the parameter `is_unbalance` is also relevant for MultiClass OVA objective. Isn't that true? Should the doc also mention this?

nwnlp:
@rauhs I think 'is_unbalance' is not necessary for OVA classifiers. for example, we have k classes , when using OVA, all classifiers may suffer from unbalanced problem. no class has  an advantage or disadvantage. so for the class which has few samples, it is relatively fair.

guolinke:
yeah, is_unbalance will be used in the binary classifiers of `multiclassova`.

guolinke:
closed via https://github.com/microsoft/LightGBM/commit/c17a5f5b8402908da1b1dfdc74a6e40530a6ffe0

StrikerRUS:
@guolinke Doesn't `pos_bagging_fraction`, `neg_bagging_fraction` and `boost_from_average` affect MultiClassOVA training indirectly as well?

guolinke:
@StrikerRUS yeah, you are right! would you help to refine them as well?

PR #2306, opened 2019-08-02
[docs] fixed params description

StrikerRUS:
Refer to https://github.com/microsoft/LightGBM/issues/2163#issuecomment-517393323.

@guolinke To be honest, I'm not sure that `pos_bagging_fraction` and `neg_bagging_fraction` are used in MultiClassOVA. It seems to me that according to the code below, `num_pos_data = 0` and `balance_bagging_cond = false` because MultiClassOVA objective doesn't override `NumPositiveData()`.
https://github.com/microsoft/LightGBM/blob/55656f27e13e35a634eb01a11493b10ca1c65c1c/src/boosting/gbdt.cpp#L728-L732

Please tell how it is actually working.


StrikerRUS:
@guolinke Was I right about `*_bagging_fraction` params?

guolinke:
I think they are only used for binary classification.

PR #2308, opened 2019-08-04
add Dockerfile for R

nuest:
Hi! I've added a Dockerfile for using LightGBM in R.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -84,7 +84,7 @@ docker build -t lightgbm-r -f dockerfile-r .
 
 After the build is finished you have two options to run the container:
 
-1. Start [RStudio](), an interactive development environment, so that you can develop your analysis using LightGBM or simply try out the R package.
+1. Start [RStudio](https://www.rstudio.com/products/rstudio/), an interactive development environment, so that you can develop your analysis using LightGBM or simply try out the R package. You can open RStudio in your web browser.
 2. Start a regular R session.
 
 In both cases you can simply call
--- a/docker/README.md
+++ b/docker/README.md
@@ -70,3 +70,26 @@ After build finished, run the container:
 ```
 docker run --rm -it lightgbm
 ```
+
+## Running the R-package Сontainer
+
+Build the container based on the [`verse` Rocker image](https://www.rocker-project.org/images/), for R users:
+
+```
+mkdir lightgbm-docker
+cd lightgbm-docker
+wget https://raw.githubusercontent.com/Microsoft/LightGBM/master/docker/dockerfile-r
+docker build -t lightgbm-r -f dockerfile-python .
+```
+
+After build finished, run the container:
+
+```
+docker run --rm -it -e PASSWORD=lightgbm -p 8787:8787 lightgbm-r
+```
+
+Open the browser at http://localhost:8787 and load the package:
+
+```
+library("lightgbm")
+```
--- a/docker/README.md
+++ b/docker/README.md
@@ -79,17 +79,35 @@ Build the container based on the [`verse` Rocker image](https://www.rocker-proje
 mkdir lightgbm-docker
 cd lightgbm-docker
 wget https://raw.githubusercontent.com/Microsoft/LightGBM/master/docker/dockerfile-r
-docker build -t lightgbm-r -f dockerfile-python .
+docker build -t lightgbm-r -f dockerfile-r .
 ```
 
-After build finished, run the container:
+After the build is finished you have two options to run the container:
+
+1. Start [RStudio](), an interactive development environment, so that you can develop your analysis using LightGBM or simply try out the R package.
+2. Start a regular R session.
+
+In both cases you can simply call
+
+```
+library("lightgbm")
+```
+
+to load the installed LightGBM R package.
+
+**RStudio**
 
 ```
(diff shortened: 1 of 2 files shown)


docker/README.md
```diff
@@ -70,3 +70,26 @@ After build finished, run the container:
 ```
 docker run --rm -it lightgbm
 ```
+
+## Running the R-package Сontainer
+
+Build the container based on the [`verse` Rocker image](https://www.rocker-project.org/images/), for R users:
```
jameslamb: Thanks! Why was it necessary to use the `verse` image instead of the lighter rocker images like `r-ver`? Similarly, could you explain the reasoning behind using an image with RStudio in it vs. one which can just run `LightGBM` in a vanilla R process?

I'm not opposed to the choices you've made, just want to understand the reasoning. I think if we leave these as-is, we might want to call this an "R training" image to note that we have some things (like RStudio) that are only necessary when doing interactive development.

docker/README.md
```diff
@@ -70,3 +70,26 @@ After build finished, run the container:
...
+
+```
+mkdir lightgbm-docker
+cd lightgbm-docker
+wget https://raw.githubusercontent.com/Microsoft/LightGBM/master/docker/dockerfile-r
+docker build -t lightgbm-r -f dockerfile-python .
```
jameslamb: you accidentally left `dockerfile-python` here, I think you meant `dockerfile-r`

StrikerRUS: `dockerfile-python`?

docker/dockerfile-r
```diff
@@ -0,0 +1,10 @@
+FROM rocker/verse:latest
+
+WORKDIR /lgbm
+
+# Python
```
StrikerRUS: `Python`?

docker/README.md
```diff
@@ -70,3 +70,26 @@ After build finished, run the container:
...
+```
+
+After build finished, run the container:
+
+```
+docker run --rm -it -e PASSWORD=lightgbm -p 8787:8787 lightgbm-r
+```
+
+Open the browser at http://localhost:8787 and load the package:
```
jameslamb: Could you please be really explicit that what this is doing is running **RStudio** inside the container? If we decide to go that path, it's important for people reading the documentation to understand that that's what is happening.

docker/README.md
nuest: I see the Dockerfiles in this directory as a help for users who want to write their own `Dockerfile` with LightGBM to try it out. As there are no automated builds from the `Dockerfile`s (or at least they're not documented in `docker/README.md`), I expect users to copy an paste the instructions into their own stacks or just use them for trying things out. You can start a vanilla R process with the same container at the cost of image size. I added that option to the docs.

An alternative would be to provide two (or more) images, which differ in the base image. Then the question is: How many versions of `r-ver` ? Would you regularly update the different version tags and update the configurations? Is it worth the effort to provide a `lightgbmr/r-ver:x.x.x` for many different versions?

This only makes sense for me if LightGBM would publish "official" automated builds, and I wouldn't expect you to do that. That's why I went with the "training" option as you put it.

docker/README.md
nuest: I updated the README. See also my [comment above](https://github.com/microsoft/LightGBM/pull/2308#discussion_r310969731).

docker/dockerfile-r
nuest: Sorry, I first created a Dockerfile with _all_ languages (see https://github.com/nuest/rocker-lightgbm), fixed now.

docker/README.md
nuest: Apologies, should have re-read after a good night sleep. Fixed!

docker/README.md
```diff
@@ -70,3 +70,44 @@ After build finished, run the container:
...
+```
+mkdir lightgbm-docker
+cd lightgbm-docker
+wget https://raw.githubusercontent.com/Microsoft/LightGBM/master/docker/dockerfile-r
+docker build -t lightgbm-r -f dockerfile-r .
+```
+
+After the build is finished you have two options to run the container:
+
+1. Start [RStudio](), an interactive development environment, so that you can develop your analysis using LightGBM or simply try out the R package.
```
StrikerRUS: Did you accidently forget a link here `[RStudio]()` or was it just to emphasize the word? If latter, please use another way to do that.

nuest:
@jameslamb I'm interested to hear what you think about the changes and replies since your review.

docker/README.md
jameslamb: Makes sense to me, thank you for the explanation. I really like the clarification you added (e.g. saying "...so that you can develop your analysis using LightGBM or simply try out the R package")!

docker/dockerfile-r
```diff
@@ -0,0 +1,10 @@
+FROM rocker/verse:latest
+
+WORKDIR /lgbm
+
+# R
+RUN apt-get update && \
+    apt-get install -y build-essential && \
+    git clone --recursive --branch stable https://github.com/Microsoft/LightGBM && \
+    cd LightGBM && \
+    Rscript build_r.R
```
jameslamb: very clean, thank you

PR #2316, opened 2019-08-07
updated docker folder

StrikerRUS:
Closed #2279.

Also, updated docker related links, Python 3.5 -> 3.6, using `stable` tag for all images and some minor formatting changes.

### DIFF
--- a/docker/dockerfile-python
+++ b/docker/dockerfile-python
@@ -17,7 +17,7 @@ RUN apt-get update && \
     conda config --set always_yes yes --set changeps1 no && \
     # lightgbm
     conda install -q -y numpy scipy scikit-learn pandas && \
-    git clone --recursive --branch stable https://github.com/Microsoft/LightGBM && \
+    git clone --recursive --branch stable --depth 1 https://github.com/Microsoft/LightGBM && \
     cd LightGBM/python-package && python setup.py install && \
 
 # clean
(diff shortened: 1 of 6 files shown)


docker/dockerfile-python
```diff
@@ -1,21 +1,28 @@
...
     # miniconda
     wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh && \
     /bin/bash Miniconda3-latest-Linux-x86_64.sh -f -b -p /opt/conda && \
     export PATH="/opt/conda/bin:$PATH" && \
+    conda config --set always_yes yes --set changeps1 no && \
     # lightgbm
-    conda install -y numpy scipy scikit-learn pandas && \
-    git clone --recursive https://github.com/Microsoft/LightGBM && \
+    conda install -q -y numpy scipy scikit-learn pandas && \
+    git clone --recursive --branch stable https://github.com/Microsoft/LightGBM && \
```
jameslamb: could you add `--depth 1`? For the use cases these Dockerfiles are intended for, it's probably unnecessary to have the entire history.

This would slightly reduce the size of the resulting image.

docker/dockerfile-python
```diff
@@ -1,21 +1,28 @@
...
     cd LightGBM/python-package && python setup.py install && \
 
 # clean
     apt-get autoremove -y && apt-get clean && \
-    conda clean -i -l -t -y && \
+    conda clean -a -y && \
```
jameslamb: agree!

StrikerRUS:
@jameslamb Thanks for your review! I added ` --depth 1` option to git command in all docker files in the latest commit.

PR #2322, opened 2019-08-12
In the rank objective, lambdas and hessians need to factor sigmoid_ into the computation.

sbruch:
Additionally, the sigmoid function has an arbitrary factor of 2 in the exponent; it is not just non-standard but the gradients are not computed correctly anyway.

### DIFF
--- a/src/objective/rank_objective.hpp
+++ b/src/objective/rank_objective.hpp
@@ -147,10 +147,10 @@ class LambdarankNDCG: public ObjectiveFunction {
         }
         // calculate lambda for this pair
         double p_lambda = GetSigmoid(delta_score);
-        double p_hessian = p_lambda * (2.0f - p_lambda);
+        double p_hessian = p_lambda * (1.0f - p_lambda);
         // update
-        p_lambda *= -delta_pair_NDCG;
-        p_hessian *= 2 * delta_pair_NDCG;
+        p_lambda *= -sigmoid_ * delta_pair_NDCG;
+        p_hessian *= sigmoid_ * sigmoid_ * delta_pair_NDCG;
         high_sum_lambda += p_lambda;
         high_sum_hessian += p_hessian;
         lambdas[low] -= static_cast<score_t>(p_lambda);
@@ -193,7 +193,7 @@ class LambdarankNDCG: public ObjectiveFunction {
     // cache
     for (size_t i = 0; i < _sigmoid_bins; ++i) {
       const double score = i / sigmoid_table_idx_factor_ + min_sigmoid_input_;
-      sigmoid_table_[i] = 2.0f / (1.0f + std::exp(2.0f * score * sigmoid_));
+      sigmoid_table_[i] = 1.0f / (1.0f + std::exp(score * sigmoid_));
     }
   }
 
--- a/src/objective/rank_objective.hpp
+++ b/src/objective/rank_objective.hpp
@@ -104,13 +104,6 @@ class LambdarankNDCG: public ObjectiveFunction {
     }
     std::stable_sort(sorted_idx.begin(), sorted_idx.end(),
                      [score](data_size_t a, data_size_t b) { return score[a] > score[b]; });
-    // get best and worst score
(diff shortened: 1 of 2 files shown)


sbruch:
I am also proposing to remove what looks like a heuristic that normalizes the gradients by the difference in document scores. This is not part of the LambdaMART algorithm and at any rate does not help with convergence rate or generalization. An alternative to this change is to make such heuristics optional.

I have also trained a model on MSLR Web30k (Fold 1) and Yahoo! Learning-to-Rank Challenge (Set 1) datasets and what I have obtained is no worse or better than a model trained from the head:

NDCG@5 on Web30k before the change is 49.20 (+/-0.07) and after the change it is 49.05 (+/-0.09)
NDCG@5 on Yahoo LTRC before the change is 74.16 (+/-0.14) and after the change it is 74.22 (+/-0.08)

where sigmoid_ is set to two (post change).

* Note that in my experiments I have removed the queries that have no relevant documents to avoid NDCG inflation -- LightGBM computes an NDCG of 1 for such queries.

sbruch:
Update: I wanted to share an updated NDCG@5 after much fine-tuning on validation sets:

Web30K Fold 1: 49.39 (+/-0.08) 
Yahoo! Set 1: 74.97 (+/-0.09) 

As noted in an earlier comment, queries with no relevant documents were discarded from the test set. Here are the hyperparameters I used if you were to reproduce the results:

For Web30k (Yahoo): learning rate is 0.05 (0.05), num_leaves is 400 (400), min_data_in_leaf is 50 (100), and min_sum_hessian_in_leafis set to 200 (10), and sigmoid_ is set to 2.

src/objective/rank_objective.hpp
```diff
@@ -141,16 +134,12 @@ class LambdarankNDCG: public ObjectiveFunction {
...
-        }
         // calculate lambda for this pair
         double p_lambda = GetSigmoid(delta_score);
-        double p_hessian = p_lambda * (2.0f - p_lambda);
+        double p_hessian = p_lambda * (1.0f - p_lambda);
         // update
-        p_lambda *= -delta_pair_NDCG;
-        p_hessian *= 2 * delta_pair_NDCG;
+        p_lambda *= -sigmoid_ * delta_pair_NDCG;
+        p_hessian *= sigmoid_ * sigmoid_ * delta_pair_NDCG;
```
guolinke: I am thinking about eliminating sigmoid_ here. That is, no sigmoid_ in p_lambda, and only sigmoid_ in p_hessian. What is your opinion?

sbruch: So the LambdaMART (implicit) cost function has the following form:

log(1 + exp(-sigmoid_ * (s_i - s_j)))

The first derivative of the above has the sigmoid_ term -- so mathematically p_lambda in the code must have the sigmoid_ term in its computation. The second derivative will have sigmoid_^2 as proposed in the code change.

For more details, please see Section 7 of this paper: http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.180.634&rep=rep1&type=pdf

guolinke:
@sbruch 
I just remember why the normalization is added. 
In our internal cases, without the normalization, some rare queries will be ignored due to the small grad/hess.
Add the normalization will enforce the equality of different queries.
So I think we should add them back.

guolinke:
@sbrush would you mind to create another PR for the normalization?
We can add a bool parameter for that function.
And I think it is better to open  by default, for the consistent behavious as before.

sbruch:
Thanks for accepting and merging the pull request. I'd certainly be happy
to send another PR and add the normalization piece back conditioned on a
boolean; but a couple of questions: (1) Would it make sense to add a flag
to the config file? (e.g., lambdarank_normalize_gradients_by_score_diff)
and (2) Could it be false by default in the public library and set to true
in internal configuration?

guolinke:
@sbruch refer to #2331, besides, I add the normalization for lambdas/hessians. However, both of them are not tested.

Issue #1829, opened 2018-11-07
Even when forced_splits is set, the threshold is chosen from the bin_upper_bounds.


AnchorBlues:
In the lightgbm, each numerical variable is replaced with discrete bins, and the threshold for split is chosen from the upper bounds of bin(variable name `bin_upper_bounds`).
Even when `forced_splits` is set, the lightgbm decides the threshold like that.
As a result, an unexpected result occurs.

## Environment info

Operating System:
Ubuntu 14.04.5 LTS, Trusty Tahr

CPU/GPU model: 
cpu
Intel(R) Core(TM) i7-6800K CPU @ 3.40GHz

C++/Python/R version:

Python 3.6.6 (default, Oct  9 2018, 12:34:16)
[GCC 7.3.0] :: Anaconda, Inc. on linux

lightgbm version : newest version (merged the commit until the commit  id ca4b6664c4823a7eaaa0(new version for master branch (#1824))

## source

```python
data = load_breast_cancer()
x = data.data
y = data.target

np.random.seed(0)
# create a new feature that takes a value 0 or 1.
dummy = np.random.randint(0, 2, size=len(x)) 

x = np.c_[dummy, x]

# create json file that forces the tree to split by the dummy feature. threshold is 0.5.
s = """
{
    "feature": 0,
    "threshold": 0.5,
    "left": {
    },
    "right": {
    }
}
"""

with open("forced_splits-0.json", mode='w') as f:
    f.write(s)

# model training
model = lgb.LGBMClassifier(random_state=42, forced_splits="forced_splits-0.json", num_leaves=3)
model.fit(x, y)

# dump model
json_obj = model.booster_.dump_model(1)
trees = json_obj['tree_info']

# the first tree
tree = trees[0]

tree['tree_structure']
```

## result
The json output of the first decision tree is as follows:

```
{'split_index': 0,
 'split_feature': 0,
 'split_gain': 5.158199787139893,
 'threshold': 1e+300,
 'decision_type': '<=',
 'default_left': True,
 'missing_type': 'None',
 'internal_value': 0,
 'internal_count': 569,
 'left_child': {'split_index': 1,
  'split_feature': 24,
  'split_gain': 57102.30078125,
  'threshold': 700.6500000000001,
  'decision_type': '<=',
  'default_left': True,
  'missing_type': 'None',
  'internal_value': 0.204333,
  'internal_count': 569,
  'left_child': {'leaf_index': 0,
[...]

PR #2325, opened 2019-08-13
Predefined bin thresholds

btrotta:
Implement the request in #1829 (ability to specify binning thresholds). The thresholds can be specified in a json file using the parameter `forcedbins_filename`.


src/io/dataset.cpp
```diff
@@ -8,12 +8,17 @@
 #include <LightGBM/utils/array_args.h>
 #include <LightGBM/utils/openmp_wrapper.h>
 #include <LightGBM/utils/threading.h>
+#include <LightGBM/json11.hpp>
```
StrikerRUS: Please keep alphabetical order, and move this after `feature_group.h`

btrotta:
@StrikerRUS thanks, fixed! One of the checks is still failing because the docs contain a reference to the URL https://github.com/microsoft/LightGBM/tree/master/examples/regression/forced_bins.json which does not exist yet. Do I need to do anything about that?

include/LightGBM/config.h
```diff
@@ -402,6 +402,11 @@ struct Config {
   // desc = see `this file <https://github.com/microsoft/LightGBM/tree/master/examples/binary_classification/forced_splits.json>`__ as an example
   std::string forcedsplits_filename = "";
 
+  // desc = path to a ``.json`` file that specifies bin upper bounds for some or all features
+  // desc = ``.json`` file should contain an array of objects, each containing the word ``feature`` (integer feature index) and ``bin_upper_bounds`` (array of thresholds for binning)
```
StrikerRUS: According to the code and example file, it searches for `bin_upper_bound` (without `s` at the end) field, so either docs or code should be updated.

src/io/dataset_loader.cpp
```diff
@@ -2,8 +2,8 @@
  * Copyright (c) 2016 Microsoft Corporation. All rights reserved.
  * Licensed under the MIT License. See LICENSE file in the project root for license information.
  */
-#include <LightGBM/dataset_loader.h>
 
+#include <LightGBM/dataset_loader.h>
```
StrikerRUS: Sorry, I probably was not clear enough, it should be
```
*/
#include <LightGBM/dataset_loader.h>

#include <LightGBM/network.h>
```

In other words, it should not produce any diff.

guolinke:
Thanks very much! I am on vacation, so don't have much time to check this carefully.
Some quick questions regards to the changes in the `bin.cpp`:
1. Are the bin boundaries the same as before, when without force bin upper bounds?
2. Why need to update bin finding code? it seems you can just load bin upper bounds from the file.

src/io/dataset_loader.cpp
btrotta: Ah ok, I get it now. Sorry, I didn't read your previous comment carefully enough!

btrotta:
@guolinke 
1. If the user doesn't specify forced bins, then the binning behavior should be the same as my previous PR #2299. That last PR corrects a smaller bug in the binning algorithm and therefore changes the binning behavior slightly from the original. I have written a more detailed description of the new behavior in the comments for PR #2299.
2. The idea is to allow the user to specify some or all of the bin thresholds. If the number of thresholds specified in the json file is less than `max_bin`, then the binning algorithm will find the remaining ones.

guolinke:
#2299 is merged. Maybe need to rebase to the master?

StrikerRUS:
@btrotta Can you please rebase the branch against the latest `master`?

src/io/bin.cpp
```diff
@@ -150,8 +150,10 @@ namespace LightGBM {
   }
 
   std::vector<double> FindBinWithZeroAsOneBin(const double* distinct_values, const int* counts,
-    int num_distinct_values, int max_bin, size_t total_sample_cnt, int min_data_in_bin) {
+    int num_distinct_values, int max_bin, size_t total_sample_cnt, int min_data_in_bin, std::vector<double> forced_upper_bounds) {
```
guolinke: could we add an independent function,`FindBinWithPredefinedBin`, to include these changes? I am afraid of these changes may introduce some bugs.

btrotta: @guolinke Yes, I can make that change.

src/io/dataset.cpp
```diff
@@ -14,6 +15,10 @@
 #include <cstdio>
 #include <sstream>
 #include <unordered_map>
+#include <fstream>
```
StrikerRUS: Please keep the alphabetical order.

src/io/bin.cpp
guolinke: ping @btrotta

btrotta: @guolinke I think that would result in sub-optimal choice of bins. Currently my code takes into account the forced bins when choosing the remaining ones, so that we can find evenly sized bins (i.e. it will not choose thresholds too close to the forced thresholds). But if we use `FindBinWithZeroAsOneBin` to create the non-forced thresholds, the thresholds it chooses may be very close to the forced thresholds in which case the bins will not be evenly sized.

guolinke:
Thanks @btrotta .
One slight improvement: As categorical features don't need the pre-defined bin, maybe we could throw a warning when user set predefined bin for categorical features?

src/io/bin.cpp
```diff
@@ -207,8 +306,19 @@ namespace LightGBM {
     return bin_upper_bound;
   }
 
+  std::vector<double> FindBinWithZeroAsOneBin(const double* distinct_values, const int* counts, int num_distinct_values,
+    int max_bin, size_t total_sample_cnt, int min_data_in_bin, std::vector<double> forced_upper_bounds) {
```
guolinke: use `const std::vector<double>& forced_upper_bounds` anywhere.

include/LightGBM/bin.h
```diff
@@ -146,9 +146,10 @@ class BinMapper {
   * \param bin_type Type of this bin
   * \param use_missing True to enable missing value handle
   * \param zero_as_missing True to use zero as missing value
+  * \param forced_upper_bounds Vector of split points that must be used (if this has size less than max_bin, remaining splits are found by the algorithm)
   */
   void FindBin(double* values, int num_values, size_t total_sample_cnt, int max_bin, int min_data_in_bin, int min_split_data, BinType bin_type,
-               bool use_missing, bool zero_as_missing);
+               bool use_missing, bool zero_as_missing, std::vector<double> forced_upper_bounds);
```
guolinke: please use const T& for container object anywhere.

include/LightGBM/dataset.h
```diff
@@ -596,6 +596,9 @@ class Dataset {
 
   void addFeaturesFrom(Dataset* other);
 
+  static std::vector<std::vector<double>> GetForcedBins(std::string forced_bins_path, int num_total_features, 
+                                                        std::unordered_set<int> categorical_features);
```
guolinke: please use const T& for container object anywhere.

src/io/bin.cpp
```diff
@@ -149,8 +149,107 @@ namespace LightGBM {
     return bin_upper_bound;
   }
 
-  std::vector<double> FindBinWithZeroAsOneBin(const double* distinct_values, const int* counts,
-    int num_distinct_values, int max_bin, size_t total_sample_cnt, int min_data_in_bin) {
+  std::vector<double> FindBinWithPredefinedBin(const double* distinct_values, const int* counts,
+    int num_distinct_values, int max_bin, size_t total_sample_cnt, int min_data_in_bin, std::vector<double>& forced_upper_bounds) {
```
guolinke: please use const T& for container object anywhere.

src/io/bin.cpp
```diff
@@ -207,8 +306,19 @@ namespace LightGBM {
     return bin_upper_bound;
   }
 
+  std::vector<double> FindBinWithZeroAsOneBin(const double* distinct_values, const int* counts, int num_distinct_values,
+    int max_bin, size_t total_sample_cnt, int min_data_in_bin, std::vector<double>& forced_upper_bounds) {
```
guolinke: please use const T& for container object anywhere.

src/io/dataset.cpp
```diff
@@ -1027,10 +1070,48 @@ void Dataset::addFeaturesFrom(Dataset* other) {
...
-
+  PushClearIfEmpty(max_bin_by_feature_, num_total_features_, other->max_bin_by_feature_, other->num_total_features_, -1);
   num_features_ += other->num_features_;
   num_total_features_ += other->num_total_features_;
   num_groups_ += other->num_groups_;
 }
 
+
+std::vector<std::vector<double>> Dataset::GetForcedBins(std::string forced_bins_path, int num_total_features, 
+                                                        std::unordered_set<int> categorical_features) {
```
guolinke: please use const T& for container object anywhere.

guolinke:
Just to confirm: the categorical feature index is read from the bin mapper, but it seems the bin mapper is not constructed at that time?

src/io/dataset.cpp
```diff
@@ -324,6 +329,14 @@ void Dataset::Construct(
...
     max_bin_by_feature_.assign(io_config.max_bin_by_feature.begin(), io_config.max_bin_by_feature.end());
   }
+  // get categorical features from the bin types so that we can read the forced bin bounds
+  std::unordered_set<int> categorical_features;
+  for (int i = 0; i < num_total_features_; ++i){
+    if ((bin_mappers[i] != nullptr) && (bin_mappers[i]->bin_type() == BinType::CategoricalBin)){
+      categorical_features.insert(i);
+    }
+  }
+  forced_bin_bounds_ = Dataset::GetForcedBins(io_config.forcedbins_filename, num_total_features_, categorical_features);
```
btrotta: @guolinke Do you mean this part? My understanding is that in this method `bin_mappers` is passed in as an argument and the bin mappers are constructed already. It seems that there are 2 uses of this method (`DatasetLoader::ConstructBinMappersFromTextData` and `DatasetLoader::CostrcutFromSampleData`). In both cases it looks like the bins are found before calling `Dataset::Construct` and passed in to the `Construct` method. Please let me know if I'm misunderstanding.

guolinke: okay, you are right!

src/io/dataset.cpp
```diff
@@ -356,6 +369,15 @@ void Dataset::ResetConfig(const char* parameters) {
   if (param.count("sparse_threshold") && io_config.sparse_threshold != sparse_threshold_) {
     Log::Warning("Cannot change sparse_threshold after constructed Dataset handle.");
   }
+  if (param.count("forcedbins_filename")) {
+    /* Since the dataset is already constructed we don't know which bins are categorical.
+    Therefore read forced bins assuming no categorical features, and warn if not the same as original. */
+    std::vector<std::vector<double>> config_bounds = Dataset::GetForcedBins(io_config.forcedbins_filename, 
+                                                                            num_total_features_, std::unordered_set<int>());
```
guolinke: categorical feature is needed here?

btrotta: In this case the dataset is already constructed, and I think there is no way to know which features are categorical (because this information is only stored in the `DatasetLoader` object, not in `Dataset`). Therefore I think the best we can do is get the forced bins assuming no categorical features, and warn the user if these are different from the existing bins.

guolinke: actually, you can call this to get the bin mapper: https://github.com/microsoft/LightGBM/blob/f1a1486929f8e981c44cba45132f5b4929453f8a/include/LightGBM/dataset.h#L476-L480 . And I think config_bounds will always no equal to forced_bin_bounds_ when categorical feature is set.

btrotta: Thanks, I changed it to get the categorical features from FeatureBinMapper. But you're right: when the json file contains forced bins for categorical features it always gives the warning, even if forced bins are unchanged.

guolinke:
@btrotta could you resolve the conflict? 
BTW, as the json file is missing in the master branch, the test failed.
Maybe you can create a new PR to submit that josn file first?

src/io/dataset.cpp
```diff
@@ -1027,10 +1077,49 @@ void Dataset::addFeaturesFrom(Dataset* other) {
...
   PushClearIfEmpty(&feature_penalty_, num_total_features_, other->feature_penalty_, other->num_total_features_, 1.0);
+  PushClearIfEmpty(&max_bin_by_feature_, num_total_features_, other->max_bin_by_feature_, other->num_total_features_, -1);
 
   num_features_ += other->num_features_;
   num_total_features_ += other->num_total_features_;
   num_groups_ += other->num_groups_;
 }
 
+
+std::vector<std::vector<double>> Dataset::GetForcedBins(std::string forced_bins_path, int num_total_features,
```
guolinke: Do you think that moving the `GetForcedBins` to the dataset_loader is better? It will be much easier to access the `categorical_features`.

guolinke: And as the pre-defined bin cannot be updated, we could don't check it resetconfig.

guolinke:
@btrotta I am a little bit confused now. Could the `forced_bin` is processed in only in dataset_loader? 
It seems the dataset class only needs to save/load the forced_bin, not other purposes.
you can save it to the dataset object in dataset_loader methods directly, similar to set_feature_name: https://github.com/microsoft/LightGBM/blob/master/src/io/dataset_loader.cpp#L696

btrotta:
@guolinke Ok, I think I see what you mean now, I will try that.

src/io/dataset.cpp
```diff
@@ -326,14 +326,7 @@ void Dataset::Construct(
...
   }
-  // get categorical features from the bin types so that we can read the forced bin bounds
-  std::unordered_set<int> categorical_features;
-  for (int i = 0; i < num_total_features_; ++i){
-    if ((bin_mappers->at(i) != nullptr) && (bin_mappers->at(i)->bin_type() == BinType::CategoricalBin)){
-      categorical_features.insert(i);
-    }
-  }
-  forced_bin_bounds_ = DatasetLoader::GetForcedBins(io_config.forcedbins_filename, num_total_features_, categorical_features);
+  forced_bin_bounds_ = forced_bins;
```
guolinke: `CopyFeatureMapperFrom`, `CreateValid` also need to copy `forced_bin_bounds_`.

include/LightGBM/dataset.h
```diff
@@ -290,6 +290,7 @@ class Dataset {
 
   void Construct(
     std::vector<std::unique_ptr<BinMapper>>* bin_mappers,
+    std::vector<std::vector<double>>& forced_bins,
```
guolinke: const T&.

StrikerRUS:
@btrotta Thank you very much!

btrotta:
Thanks for your help @guolinke  and @StrikerRUS

PR #2331, opened 2019-08-18
normalize the lambdas in lambdamart objective

guolinke:
- [x] benchmark

benchmark run by https://github.com/guolinke/boosting_tree_benchmarks

ndcg at Yahoo LTR:

| Metric      | original | norm |
|----|  ----| ---- | 
| ndcg@1|0.72757|0.732981|
| ndcg@3|0.731267|0.735689|
| ndcg@5|0.750774|0.75352|
| ndcg@10|0.791388|0.793498|

ndcg at Yahoo LTR:

| Metric      | original | norm |
|----|  ----| ---- | 
| ndcg@1|0.512893|0.519751|
| ndcg@3|0.50118|0.499677|
| ndcg@5|0.505995|0.5058|
| ndcg@10|0.524724|0.524123|

### DIFF
--- a/include/LightGBM/config.h
+++ b/include/LightGBM/config.h
@@ -721,7 +721,7 @@ struct Config {
   // desc = used only in ``lambdarank`` application
   // desc = set this to ``true`` to normalize the lambdas for different queries, and improve the performance of unbalanced data
   // desc = set this to ``false`` to enforce the original lambdamart algorithm
-  bool lambdamart_norm = false;
+  bool lambdamart_norm = true;
 
   // type = multi-double
   // default = 0,1,3,7,15,31,63,...,2^30-1
--- a/include/LightGBM/config.h
+++ b/include/LightGBM/config.h
@@ -719,7 +719,7 @@ struct Config {
   int max_position = 20;
 
   // desc = used only in ``lambdarank`` application
-  // desc = set this to ``true`` to normalize the lambdas for different queries, and improve the performance of unbalanced data
+  // desc = set this to ``true`` to normalize the lambdas for different queries, and improve the performance for unbalanced data
   // desc = set this to ``false`` to enforce the original lambdamart algorithm
   bool lambdamart_norm = true;
 
--- a/include/LightGBM/config.h
+++ b/include/LightGBM/config.h
@@ -718,6 +718,10 @@ struct Config {
   // desc = optimizes `NDCG <https://en.wikipedia.org/wiki/Discounted_cumulative_gain#Normalized_DCG>`__ at this position
   int max_position = 20;
 
+  // desc = set this to ``true`` to normalize the lambdas for different queries, and improve the performance of unbalanced data
+  // desc = set this to ``false`` to enforce the original lambdamart algorithm
+  bool lambdamart_norm = true;
+
   // type = multi-double
   // default = 0,1,3,7,15,31,63,...,2^30-1
   // desc = used only in ``lambdarank`` application
--- a/include/LightGBM/config.h
+++ b/include/LightGBM/config.h
@@ -720,7 +720,7 @@ struct Config {
 
   // desc = set this to ``true`` to normalize the lambdas for different queries, and improve the performance of unbalanced data
   // desc = set this to ``false`` to enforce the original lambdamart algorithm
-  bool lambdamart_norm = true;
+  bool lambdamart_norm = false;
 
   // type = multi-double
   // default = 0,1,3,7,15,31,63,...,2^30-1
--- a/include/LightGBM/config.h
+++ b/include/LightGBM/config.h
@@ -718,6 +718,7 @@ struct Config {
   // desc = optimizes `NDCG <https://en.wikipedia.org/wiki/Discounted_cumulative_gain#Normalized_DCG>`__ at this position
   int max_position = 20;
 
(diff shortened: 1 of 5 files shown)


include/LightGBM/config.h
```diff
@@ -718,6 +718,10 @@ struct Config {
   // desc = optimizes `NDCG <https://en.wikipedia.org/wiki/Discounted_cumulative_gain#Normalized_DCG>`__ at this position
   int max_position = 20;
 
+  // desc = set this to ``true`` to normalize the lambdas for different queries, and improve the performance of unbalanced data
+  // desc = set this to ``false`` to enforce the original lambdamart algorithm
+  bool lambdamart_norm = false;
```
guolinke: will change to true if enable it by default is better. Will benchmark that in the next few days.

include/LightGBM/config.h
```diff
@@ -718,6 +718,10 @@ struct Config {
   // desc = optimizes `NDCG <https://en.wikipedia.org/wiki/Discounted_cumulative_gain#Normalized_DCG>`__ at this position
   int max_position = 20;
 
+  // desc = set this to ``true`` to normalize the lambdas for different queries, and improve the performance of unbalanced data
```
StrikerRUS: The first line should be `desc = used only in ``lambdarank`` application`

src/objective/rank_objective.hpp
```diff
@@ -104,6 +105,14 @@ class LambdarankNDCG: public ObjectiveFunction {
     }
     std::stable_sort(sorted_idx.begin(), sorted_idx.end(),
                      [score](data_size_t a, data_size_t b) { return score[a] > score[b]; });
+    // get best and worst score	
+    const double best_score = score[sorted_idx[0]];
+    data_size_t worst_idx = cnt - 1;
+    if (worst_idx > 0 && score[sorted_idx[worst_idx]] == kMinScore) {
+      worst_idx -= 1;
+    }
+    const double wrost_score = score[sorted_idx[worst_idx]];
```
StrikerRUS: `worst_score`: swap `o` and `r`

guolinke:
I think the results with norm is better, so I will enable it by default.

src/objective/rank_objective.hpp
```diff
@@ -104,6 +105,14 @@ class LambdarankNDCG: public ObjectiveFunction {
     }
     std::stable_sort(sorted_idx.begin(), sorted_idx.end(),
                      [score](data_size_t a, data_size_t b) { return score[a] > score[b]; });
+    // get best and worst score	
+    const double best_score = score[sorted_idx[0]];
+    data_size_t wrost_idx = cnt - 1;
```
StrikerRUS: Mis-correction in b84c14e2ccbf5d40c793e73625c1ebd1c232a36a: now two variables have `wrost_` prefix instead of `worst`.

sbruch:
Is there a technical report or a paper you could share with the learning-to-rank research community that explains what this normalization does and how it was derived/it mathematically makes sense?

guolinke:
@sbruch 
It is just a simple trick for tree-based model, not for the general learning-to-rank applications.
The idea is similar to the `reg_sqrt` in regression tasks.
In short, the tree model is sensitive to the label ranges, for its outputs are the average of label values.
For the skew distribution, the leaf outputs will be drifted to the large outputs. 
And add normalization can somehow solve this problem.

guolinke:
@StrikerRUS is this okay to merge?

include/LightGBM/config.h
```diff
@@ -718,6 +718,11 @@ struct Config {
   // desc = optimizes `NDCG <https://en.wikipedia.org/wiki/Discounted_cumulative_gain#Normalized_DCG>`__ at this position
   int max_position = 20;
 
+  // desc = used only in ``lambdarank`` application
+  // desc = set this to ``true`` to normalize the lambdas for different queries, and improve the performance of unbalanced data
```
StrikerRUS: `performance of` -> `performance for`

tests/python_package_test/test_sklearn.py
```diff
@@ -79,11 +79,11 @@ def test_lambdarank(self):
...
-                eval_group=[q_test], eval_at=[1, 3], early_stopping_rounds=5, verbose=False,
-                callbacks=[lgb.reset_parameter(learning_rate=lambda x: 0.95 ** x * 0.1)])
-        self.assertLessEqual(gbm.best_iteration_, 12)
-        self.assertGreater(gbm.best_score_['valid_0']['ndcg@1'], 0.6173)
-        self.assertGreater(gbm.best_score_['valid_0']['ndcg@3'], 0.6479)
+                eval_group=[q_test], eval_at=[1, 3], early_stopping_rounds=10, verbose=False,
+                callbacks=[lgb.reset_parameter(learning_rate=lambda x: max(0.01, 0.1 - 0.01 * x))])
+        self.assertLessEqual(gbm.best_iteration_, 25)
+        self.assertGreater(gbm.best_score_['valid_0']['ndcg@1'], 0.60)
+        self.assertGreater(gbm.best_score_['valid_0']['ndcg@3'], 0.60)
```
StrikerRUS: Can we make these asserts more strict/precise? I remember this test helped to spot the issue of inconsistent results on different platforms due to using `sort` instead of `stable_sort`?

PR #2342, opened 2019-08-20
fix the bug in bin with small values

guolinke:


guolinke:
ping @btrotta for https://github.com/microsoft/LightGBM/pull/2299
#2299 will bring a check fail for all negative feature values, and produce additional bins for all positive feature values, and is different from previous bing results.

btrotta:
@guolinke Sorry for the mistake. I'll update my current PR #2325 to match this behavior.

PR #2339, opened 2019-08-19
[R-package] fixed overly-strict compiler check in install.libs.R (fixes #2259)

jameslamb:
As of #2164 , the R package picked up a fix in `install.libs.R` that handles difficulty finding OpenMP stuff when building on Mac OS Mojave.

Per discussion on #2259 , we found that that fix only works if environment variable `CC` is set to something that literally starts with `gcc` (and the same with `CXX` and `g++`). That could cause build failures if you pass a fully-qualified path like `CXX=/usr/local/bin/g++-8`. Thanks to @brandenkmurray for the help identifying that as the problem!

In this PR, I relax that check so that both `g++-8`, `g++`, and `/some/path/to/g++-8` will all work. I did not take an opinion on the versions of `g++` that are known to work with `lightgbm` (as suggested in [this comment](https://github.com/microsoft/LightGBM/issues/2259#issuecomment-522330822))  as I think that would increase our maintenance burden by duplicating the number of places that is specified.

I tested that this works (lucky for us I'm running Mojave 10.14.5 on my laptop) and can confirm it's fixed. Code to test that it worked:

```bash
export CC=gcc-8 CXX=g++-8
Rscript build_r.R
```

```bash
export CXX=/usr/local/bin/g++-8 CC=/usr/local/bin/gcc-8
Rscript build_r.R
```

PR #2345, opened 2019-08-21
[python] correctly handle full path to compiler in CC and CXX env vars

StrikerRUS:
For instance, `CC` can be `/usr/bin/gcc`. Refer to #2339.

@jameslamb Seems that the regex from #2339 for R-package will produce false positive result for something like `C:/Users/gccharlee/pgcc-8` (any **intermediate folder name** which starts with `gcc`).

### DIFF
--- a/python-package/setup.py
+++ b/python-package/setup.py
@@ -168,8 +168,8 @@ def compile_cpp(use_mingw=False, use_gpu=False, use_mpi=False,
     else:  # Linux, Darwin (macOS), etc.
         logger.info("Starting to compile with CMake.")
         # Apple Clang with OpenMP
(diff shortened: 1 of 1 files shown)


jameslamb:
Ah! Forgot a `$`. Will make a PR momentarily, thanks for catching it.

StrikerRUS:
@jameslamb  Thanks for your review!
> This does mean R and Python will have slightly different behavior (since you are using startswith() but I think it covers most cases and is definitely an improvement over the previous state.

Yeah, I decided to leave `startswith` because of its' better readability and simplicity in comparison with `regex`. And the most important - it doesn't require additional imports.

BTW, it's not critical since we run this compiler check only for macOS, but the regex from #2339 doesn't work for Windows (e.g. `gcc.exe`).

PR #904, opened 2017-09-11
[python] bring pandas support to the sklearn wrapper back

StrikerRUS:
My fault is that in recent [PR](https://github.com/Microsoft/LightGBM/pull/870 ) with sklearn compatibility I forgot about pandas support. This PR is expected to bring pandas support back to the sklearn wrapper.
Fix #901 .

Since scikit-learn has poor pandas support, I think that the best thing to do is just pass `pandas.DataFrame` to Booster as is.

PR #2352, opened 2019-08-23
[python] error on nan in y for DataFrames

bartlesy:
I noticed that passing in a Pandas Series with a NaN will result in a model that only predicts 0s rather than an error or warning because CheckXY is only being run when X is a numpy array. The `sklearn.utils.validation.check_array` function being called is compatible with a DataFrame/Series as input, so it seemed reasonable to run the same check when X is a DataFrame as when X is an array.

Previous behavior:
```python
import numpy as np
import lightgbm as lgb
import pandas as pd

X = pd.DataFrame(np.array([0, np.nan, 0, 1, 1]).reshape(-1, 1))
y = pd.Series(np.array([0, np.nan, 0, 1, 1]))

gbdt = lgb.LGBMRegressor(n_estimators=1, min_child_samples=1)
print(gbdt.fit(X, y).predict(X))
```
>>> [0.0 0.0 0.0 0.0 0.0]

Now this throws a `ValueError`

### DIFF
--- a/python-package/lightgbm/sklearn.py
+++ b/python-package/lightgbm/sklearn.py
@@ -525,7 +525,7 @@ class LGBMModel(_LGBMModelBase):
             eval_metric = [eval_metric] if isinstance(eval_metric, (string_type, type(None))) else eval_metric
             params['metric'] = set(original_metric + eval_metric)
 
-        if not isinstance(X, (DataFrame, DataTable)):
+        if not isinstance(X, DataTable):
             _X, _y = _LGBMCheckXY(X, y, accept_sparse=True, force_all_finite=False, ensure_min_samples=2)
             _LGBMCheckConsistentLength(_X, _y, sample_weight)
         else:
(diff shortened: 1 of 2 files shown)


bartlesy:
> Hi @bartlesy!
> 
> LightGBM supports NANs and pandas.DataFrame with categorical types. However, `sklearn.utils.validation.check_array()` is not pandas-friendly. Refer to #904 for more info.

Hey @StrikerRUS - Is it possible that was changed since? It seems to work fine now:
```python
from sklearn.utils.validation import check_array
import pandas as pd
import numpy as np

X = pd.DataFrame(np.array([0, np.nan, 0, 1, 1]).reshape(-1, 1))
y = pd.Series(np.array([0, np.nan, 0, 1, 1]))

try:
    print(type(y))
    print(check_array(y.dropna(), ensure_2d=False))
    check_array(y, ensure_2d=False)
except Exception as e:
    print(e)
    
try:
    print(type(X))
    print(check_array(X.dropna()))
    check_array(X)
except Exception as e:
    print(e)
    
print(pd.__version__)
```
><class 'pandas.core.series.Series'>
[0. 0. 1. 1.]
Input contains NaN, infinity or a value too large for dtype('float64').
<class 'pandas.core.frame.DataFrame'>
[[0.]
 [0.]
 [1.]
 [1.]]
Input contains NaN, infinity or a value too large for dtype('float64').
0.24.2

StrikerRUS:
@bartlesy
> Is it possible that was changed since?

Maybe something has been changed... However, as you can see, tests with categorical features fail. So, it's still incompatible check.

bartlesy:
Ah you're right, I see

PR #2354, opened 2019-08-24
change network retry delay strategy

### LINKED ISSUE #2348: Add Smarter Backoffs for MPI ring connection
Hey, I am running many LightGBM training runs as part of LightGBM on Spark's CI process and i notice that it does alot of waiting because of bad network and this dramatically slows the tests:

https://github.com/microsoft/LightGBM/blob/0a4a7a86f5a1d3146c36c7d8c082154a193d4893/src/network/linkers_socket.cpp#L190

Could you potentially make it first wait for half a second, then wait for 10 seconds? Or make this configurable. Thanks for the help!

guolinke:
to fix #2348

### DIFF
--- a/src/network/linkers_socket.cpp
+++ b/src/network/linkers_socket.cpp
@@ -187,20 +187,22 @@ void Linkers::Construct() {
   listener_->Listen(incoming_cnt);
   std::thread listen_thread(&Linkers::ListenThread, this, incoming_cnt);
   const int connect_fail_retry_cnt = 20;
+  const int connect_fail_retry_first_delay_interval = 200; // 0.2 s
+  const float connect_fail_retry_delay_factor = 1.3f;
   // start connect
   for (auto it = need_connect.begin(); it != need_connect.end(); ++it) {
     int out_rank = it->first;
     // let smaller rank connect to larger rank
     if (out_rank > rank_) {
       TcpSocket cur_socket;
-      int connect_fail_delay_time = 200;  // 0.2s
+      int connect_fail_delay_time = connect_fail_retry_first_delay_interval;
       for (int i = 0; i < connect_fail_retry_cnt; ++i) {
         if (cur_socket.Connect(client_ips_[out_rank].c_str(), client_ports_[out_rank])) {
           break;
         } else {
           Log::Warning("Connecting to rank %d failed, waiting for %d milliseconds", out_rank, connect_fail_delay_time);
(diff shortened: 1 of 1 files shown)


src/network/linkers_socket.cpp
```diff
@@ -187,19 +187,20 @@ void Linkers::Construct() {
...
     if (out_rank > rank_) {
       TcpSocket cur_socket;
+      int connect_fail_delay_time = 200;  // 0.2s
       for (int i = 0; i < connect_fail_retry_cnt; ++i) {
         if (cur_socket.Connect(client_ips_[out_rank].c_str(), client_ports_[out_rank])) {
           break;
         } else {
           Log::Warning("Connecting to rank %d failed, waiting for %d milliseconds", out_rank, connect_fail_delay_time);
           std::this_thread::sleep_for(std::chrono::milliseconds(connect_fail_delay_time));
+          connect_fail_delay_time = static_cast<int>(connect_fail_delay_time * 1.5);
```
jameslamb: @guolinke now that we're using this backoff strategy, I think `connect_fail_retry_cnt` should be reduced. Under the current state (on `master`) it would take 200s to determine no connectivity. Now it will take roughly 1329s to determine no connectivity (summing `0.2 * 1.5^n` for `n` in `0:19`). That means if we merge this PR, `LightGBM` will now takes 6x as long to diagnose a connectivity problem.

I like this new backoff strategy but think we should consider reducing `connect_fail_retry_cnt` to, say, 16 (which would mean it takes 262s to determine no connectivity). 

Would you consider that change? Or was the intent of this PR to ALSO wait longer to declare connection lost?

Also, per #2348 do we eventually want to make this retry logic something user-configurable?

StrikerRUS:
Can't we simply expose `connect_fail_retry_cnt` and `connect_fail_delay_time` to [config](https://lightgbm.readthedocs.io/en/latest/Parameters.html#network-parameters) and let users decide how long they want to wait according to their network configuration?

guolinke:
@StrikerRUS I think the exponential backoff strategy is better for the network connecting retry.
For the user-configurable, actually, I want to expose them for the first time. But then I think it is not a good idea, for we don't want to user to tune them. 

And exponential backoff strategy should work for most cases:
for stable network env, it should connect successfully in a short time.
for unstable network env, it also provides enough time buffers to retry.

StrikerRUS:
@guolinke Ah, got it! Thanks a lot for your detailed explanation!

PR #2356, opened 2019-08-25
[java][mmlspark] Fix cached predictor causing bad values for predicted probabilities

imatiach-msft:
In mmlspark, the predictied probabilities column was appearing the same as raw predicted values when both columns were computed from the model.  This is due to a caching bug in the native code logic that was recently added.  I added an array of predictors, one for each type, to fix the caching bug - and also added some additional checks in case the cached predictor needs to be reinitialized (not used by mmlspark directly).

### DIFF
--- a/src/c_api.cpp
+++ b/src/c_api.cpp
@@ -46,6 +46,8 @@ catch(std::string& ex) { return LGBM_APIHandleException(ex); } \
 catch(...) { return LGBM_APIHandleException("unknown exception"); } \
 return 0;
 
+const int PREDICTOR_TYPES = 4;
+
 class Booster {
  public:
   explicit Booster(const char* filename) {
@@ -205,8 +207,15 @@ class Booster {
                const Config& config,
                double* out_result, int64_t* out_len) {
     std::lock_guard<std::mutex> lock(mutex_);
-
-    if (single_row_predictor_.get() == nullptr) {
+    if (single_row_predictor_[predict_type].get() == nullptr ||
+	single_row_predictor_pred_early_stop[predict_type] != config.pred_early_stop ||
+	single_row_predictor_pred_early_stop_freq[predict_type] != config.pred_early_stop_freq ||
+	single_row_predictor_pred_early_stop_margin[predict_type] != config.pred_early_stop_margin ||
+	single_row_predictor_iter_[predict_type] != num_iteration) {
+      single_row_predictor_iter_[predict_type] = num_iteration;
+      single_row_predictor_pred_early_stop[predict_type] = config.pred_early_stop;
+      single_row_predictor_pred_early_stop_freq[predict_type] = config.pred_early_stop_freq;
+      single_row_predictor_pred_early_stop_margin[predict_type] = config.pred_early_stop_margin;
       bool is_predict_leaf = false;
       bool is_raw_score = false;
       bool predict_contrib = false;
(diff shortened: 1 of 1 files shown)


imatiach-msft:
adding @eisber

src/c_api.cpp
```diff
@@ -364,9 +373,13 @@ class Booster {
  private:
   const Dataset* train_data_;
   std::unique_ptr<Boosting> boosting_;
-  std::unique_ptr<Predictor> single_row_predictor_;
-  PredictFunction single_row_predict_function_;
-  int64_t single_row_num_pred_in_one_row_;
+  std::unique_ptr<Predictor> single_row_predictor_[PREDICTOR_TYPES];
+  bool single_row_predictor_pred_early_stop[PREDICTOR_TYPES];
+  int single_row_predictor_pred_early_stop_freq[PREDICTOR_TYPES];
```
imatiach-msft: just to confirm, will these arrays be freed correctly with the class on the heap once class is deleted?

src/c_api.cpp
```diff
@@ -364,9 +373,13 @@ class Booster {
...
+  double single_row_predictor_pred_early_stop_margin[PREDICTOR_TYPES];
+  int single_row_predictor_iter_[PREDICTOR_TYPES];
```
imatiach-msft: the config has a lot of info, but it seems I only care about these values (early_stop, early_stop_freq, early_stop_margin, iter), so I only cache them in an array

guolinke:
I remember the single row prediction is added by @eisber 
ping @eisber to review.

src/c_api.cpp
```diff
@@ -205,8 +207,15 @@ class Booster {
...
-    if (single_row_predictor_.get() == nullptr) {
+    if (single_row_predictor_[predict_type].get() == nullptr ||
+	single_row_predictor_pred_early_stop[predict_type] != config.pred_early_stop ||
+	single_row_predictor_pred_early_stop_freq[predict_type] != config.pred_early_stop_freq ||
+	single_row_predictor_pred_early_stop_margin[predict_type] != config.pred_early_stop_margin ||
+	single_row_predictor_iter_[predict_type] != num_iteration) {
+      single_row_predictor_iter_[predict_type] = num_iteration;
+      single_row_predictor_pred_early_stop[predict_type] = config.pred_early_stop;
+      single_row_predictor_pred_early_stop_freq[predict_type] = config.pred_early_stop_freq;
+      single_row_predictor_pred_early_stop_margin[predict_type] = config.pred_early_stop_margin;
```
guolinke: besides of configs, I think the model update (e.g.  more trees) should be considered.

src/c_api.cpp
```diff
@@ -364,9 +373,13 @@ class Booster {
...
-  std::unique_ptr<Predictor> single_row_predictor_;
-  PredictFunction single_row_predict_function_;
-  int64_t single_row_num_pred_in_one_row_;
+  std::unique_ptr<Predictor> single_row_predictor_[PREDICTOR_TYPES];
+  bool single_row_predictor_pred_early_stop[PREDICTOR_TYPES];
+  int single_row_predictor_pred_early_stop_freq[PREDICTOR_TYPES];
+  double single_row_predictor_pred_early_stop_margin[PREDICTOR_TYPES];
+  int single_row_predictor_iter_[PREDICTOR_TYPES];
+  PredictFunction single_row_predict_function_[PREDICTOR_TYPES];
+  int64_t single_row_num_pred_in_one_row_[PREDICTOR_TYPES];
```
guolinke: maybe have a new class for the single row prediction? due to there are much more member variables of it.
It will be easier to maintain.

src/c_api.cpp
guolinke: I think num_iteration could be always -1 or 0, which means using all models.
So you should better check the real number of trees in the model.

src/c_api.cpp
guolinke: I think in this file is okay.

guolinke:
@imatiach-msft Thanks! much cleaner now.

src/c_api.cpp
```diff
@@ -205,33 +254,17 @@ class Booster {
...
-
-      // TODO(eisber): config could be optimized away... (maybe using lambda callback?)
-      single_row_predictor_.reset(new Predictor(boosting_.get(), num_iteration, is_raw_score, is_predict_leaf, predict_contrib,
-                                                config.pred_early_stop, config.pred_early_stop_freq, config.pred_early_stop_margin));
-      single_row_num_pred_in_one_row_ = boosting_->NumPredictOneRow(num_iteration, is_predict_leaf, predict_contrib);
-      single_row_predict_function_ = single_row_predictor_->GetPredictFunction();
+    if (single_row_predictor_[predict_type].get() == nullptr ||
+	!single_row_predictor_[predict_type]->IsPredictorEqual(config, num_iteration, *boosting_.get())) {
+      single_row_predictor_[predict_type].reset(new SingleRowPredictor(predict_type, *boosting_.get(),
+								       config, num_iteration));
```
StrikerRUS: Please use whitespaces everywhere, not tabs.

src/c_api.cpp
eisber: those are fixed size arrays and they'll be deallocated directly.

I'd suggest to use create a 

# new class containing (unique_ptr, bool, int) 
# implement operator== (const Config&) and operator=(const Config&)
# would make the assignment and comparison a bit more readable.

src/c_api.cpp
eisber: +1 on the approach

PR #2364, opened 2019-08-29
[java][mmlspark] fix lightgbm swig wrapper on windows

imatiach-msft:
part of fix for https://github.com/Azure/mmlspark/issues/483
specifically, we should be using int64_t directly instead of converting from long values as we are currently doing in some parts of the code


StrikerRUS:
Close-reopen to re-trigger CI.

Issue #1143, opened 2017-12-25
[R-package] Add pkgdown documentation support


Laurae2:
I will be adding [pkgdown](https://github.com/r-lib/pkgdown) support to the documentation of LightGBM.

You can check the branch here: https://github.com/Laurae2/LightGBM/tree/pkgdown

To add support for pkgdown, it will require the following tasks:

- [x] Find hosting for the pkgdown documentation (I'll host it, max number of simultaneous users basically infinite thanks to static site hosted on CDNs)
- [x] Add badges to R README.md
- [x] Change a bit the README.md documentation
- [x] Add Authors@R field in DESCRIPTION (@guolinke (aut+cre) + @Laurae2 (ctb) + @yanyachen (ctb) )
- [x] Fix Author field in DESCRIPTION
- [x] Move `Matrix` and `methods` to Depends instead of Imports (fixes many instances of example crashes)
- [x] Fix all instances of crashes in function examples
- [x] Update date of DESCRIPTION
- [x] Regenerate the whole documentation from scratch
- [x] Run successfully `pkgdown::build_site()` on the R-package folder without vignettes
- [x] Convert demos to vignettes
- [x] Find workarounds for all function examples which crashes and cannot have a direct or immediate workaround
- [x] Fix all instances of crashes in vignettes
- [x] Find workarounds for all vignettes which crashes and cannot have a direct or immediate workaround
- [x] Run successfully `pkgdown::build_site()` on the R-package folder with vignettes

Extras:

- [ ] Add Travis support for running pkgdown so it can perform all examples / vignettes test in one straight command: `pkgdown::build_site(run_dont_run = TRUE)` (no idea how to do it, if someone can contribute to do it for us that would be great...)

I'm sharing my setup for running `pkgdown` quickly:

```r
# Load useful libraries for development
library(devtools) # install.packages("devtools")
library(roxygen2) # devtools::install_github("klutometis/roxygen")
library(pkgdown) # devtools::install_github("hadley/pkgdown")

# Set the working directory to where I am
setwd("E:/GitHub/LightGBM/R-package")

# Install package
install()

# Generate documentation
document()

# Check for errors
# devtools::check(document = FALSE) # fails/crashes on LightGBM

# Build static website
pkgdown::build_site(run_dont_run = TRUE) # currently fails/crashes

# Remember to cleanup R-package/src folder after testing (keep only install.libs.R, Makevars, Makevars.win)
# Remember to delete R-package/docs after testing
```

[...]

PR #2370, opened 2019-09-01
[docs] fix version type in the R-package tooltip

StrikerRUS:
Fix one issue from https://github.com/microsoft/LightGBM/issues/1143#issuecomment-526936690.

As we generate docs for the each commit to the `master` branch, version cannot be "released".

Before: 

image

After:

image

Refer to https://pkgdown.r-lib.org/reference/build_site.html#development-mode.

### DIFF
--- a/R-package/_pkgdown.yml
+++ b/R-package/_pkgdown.yml
@@ -6,6 +6,9 @@ site:
   root: ''
   title: LightGBM, Light Gradient Boosting Machine
 
+development:
+  mode: unreleased
+
 authors:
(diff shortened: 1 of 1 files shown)


R-package/_pkgdown.yml
```diff
@@ -6,6 +6,9 @@ site:
   root: ''
   title: LightGBM, Light Gradient Boosting Machine
 
+development:
+  mode: unreleased
```
jameslamb: oh wow I had never seen this before

PR #2380, opened 2019-09-04
[Py] In _data_from_pandas, avoid copying df.values if it's already float

dnouri:
astype() here creates a copy of the entire dataset; we'll avoid that
if it's not necessary.

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -296,7 +296,9 @@ def _data_from_pandas(data, feature_name, categorical_feature, pandas_categorica
             raise ValueError("DataFrame.dtypes for data must be int, float or bool.\n"
                              "Did not expect the data types in the following fields: "
                              + ', '.join(data.columns[bad_indices]))
-        data = data.values.astype('float')
+        data = data.values
+        if data.dtype != 'float':
+            data = data.astype('float')
     else:
         if feature_name == 'auto':
             feature_name = None


python-package/lightgbm/basic.py
```diff
@@ -296,7 +296,9 @@ def _data_from_pandas(data, feature_name, categorical_feature, pandas_categorica
             raise ValueError("DataFrame.dtypes for data must be int, float or bool.\n"
                              "Did not expect the data types in the following fields: "
                              + ', '.join(data.columns[bad_indices]))
-        data = data.values.astype('float')
+        data = data.values
+        if data.dtype != 'float':
```
jameslamb: Thanks for this! I apologize in advance if you got to this PR because you experienced an out-of-memory error.

I think it's a good change, but I don't think this syntax is quite right. `pandas` DataFrames don't have a `.dtype`, they have an array of dtypes (one per column)

```python
import pandas as pd
df = pd.DataFrame({'a': ['1', '2', '3'], 'b': [4, 5, 6]})
df.dtype
```
will return

```
  File "<stdin>", line 1, in <module>
  File "/Users/jlamb/miniconda3/lib/python3.6/site-packages/pandas/core/generic.py", line 4376, in __getattr__
    return object.__getattribute__(self, name)
AttributeError: 'DataFrame' object has no attribute 'dtype'
```

where `.dtypes` will give you

```
df.dtypes
a    object
b     int64
dtype: object
```

I think you want something like this

```python
import numpy as np
all_float = all([col_type == np.dtype('float') for col_type in data.dtypes])
if not all_float:
    data = data.astype('float')
```

@StrikerRUS I'm surprised that this syntax error (`data.dtype`) wasn't caught by our CI. Have I made a mistake in my understanding of what `data` is?

dnouri: @jameslamb [`data.values` returns a numpy array](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.values.html), thus `data.values.dtype` is fine.

jameslamb: 😆 I read straight past the `.values`, haha. Sorry! Ok then I am good with this change! Will leave it for @StrikerRUS to review just in case, since I am not the main owner of the Python side for this project, but this seems good to me.

Also TIL that `'float' == dtype('float')` works! That's interesting.

```
import pandas as pd
df = pd.DataFrame({'a': [1.0, 2.0, 3.0], 'b': [4.0, 5.0, 6.0]})
print(df.values.dtype)
print(df.values.dtype == 'float')
```

StrikerRUS:
Ah, isn't `numpy` quite smart to not cast float to float?.. Let me check please.

dnouri:
> Ah, isn't `numpy` quite smart to not cast float to float?.. Let me check please.

The `numpy` docs are explicit about the fact that [the object returned by `astype()` is (always) a copy](https://docs.scipy.org/doc/numpy/reference/generated/numpy.ndarray.astype.html).

StrikerRUS:
> The numpy docs are explicit about the fact that the object returned by astype() is (always) a copy.

Not sure.

```
import numpy as np

arr = np.array([[1.0, 2.0, 3.0], [4.0, 5.0, 6.0]])
arr is arr.astype('float', copy=False)  # True
id(arr) == id(arr.astype('float', copy=False))  # True
```

However, `id(arr) == id(arr.astype('float32', copy=False))` is `False`.

dnouri:
> > The numpy docs are explicit about the fact that the object returned by astype() is (always) a copy.
> 
> Not sure.

Fair enough.  If you pass `copy=False` it won't copy, sometimes.  :-D

I'm fine with having #2383 close this.  I was a little afraid to make further-reaching changes. I noticed that your change affects the dtype that this function returns (before it was always float64, with your PR it's sometimes float32, sometimes float64).

In any case, it'll be cool to cut down peak memory usage by approximately a factor of 2.

StrikerRUS:
@dnouri
> I noticed that your change affects the dtype that this function returns (before it was always float64, with your PR it's sometimes float32, sometimes float64).

Yeah, LightGBM can work with both float32 and float64 values, so we can allow float32 at Python side.

I'm sorry, I haven't notify you about #2383 yesterday - needed to rush. But you have already found it by yourself 😃 . I decided to not bother you and ask to rewrite other places. If you don't mind #2383 will supersede this PR.

dnouri:
> I'm sorry, I haven't notify you about #2383 yesterday - needed to rush. But you have already found it by yourself  . I decided to not bother you and ask to rewrite other places. If you don't mind #2383 will supersede this PR.

Sure, no worries, and keep up the good work!

PR #2383, opened 2019-09-05
[python] avoid data copy where possible

StrikerRUS:
Closed #2380.


python-package/lightgbm/basic.py
```diff
@@ -296,7 +296,9 @@ def _data_from_pandas(data, feature_name, categorical_feature, pandas_categorica
             raise ValueError("DataFrame.dtypes for data must be int, float or bool.\n"
                              "Did not expect the data types in the following fields: "
                              + ', '.join(data.columns[bad_indices]))
-        data = data.values.astype('float')
+        data = data.values
+        if data.dtype != np.float32 and data.dtype != np.float64:
```
StrikerRUS: Why not simple `'float'`:
```
import numpy as np

arr = np.array([[1.0, 2.0, 3.0], [4.0, 5.0, 6.0]], dtype=np.float32)
arr.dtype == 'float'  # False
```

StrikerRUS:
@guolinke
I'd like to use precise type names to avoid possible problems. Is it `int32` or `int64`?
https://github.com/microsoft/LightGBM/blob/29525ffefba94ea994e28d0db1d8029a6041e3ae/python-package/lightgbm/basic.py#L2510-L2511

guolinke:
@StrikerRUS I think it is int32, refer to:
https://github.com/microsoft/LightGBM/blob/317b1bfbbfdd8f8a4f5af16b970526bb9486f588/src/c_api.cpp#L1539

guolinke:
@StrikerRUS 
seems related to this PR:
https://github.com/microsoft/LightGBM/pull/2384

maybe a better solution is to save the numpy type inside lgb.Dataset?

python-package/lightgbm/basic.py
```diff
@@ -487,7 +491,7 @@ def predict(self, data, num_iteration=-1,
             preds, nrow = self.__pred_for_np2d(data, num_iteration, predict_type)
         elif isinstance(data, list):
             try:
-                data = np.array(data)
+                data = np.array(data, copy=False)
```
dnouri: This is right after a check for `isinstance(data, list)`. The `copy=False` doesn't seem to make sense here.

StrikerRUS: @dnouri Thanks! Fixed.

StrikerRUS:
@guolinke
> @StrikerRUS I think it is int32, refer to:

Thanks!

> maybe a better solution is to save the numpy type inside lgb.Dataset?

Sorry, didn't get it.

python-package/lightgbm/basic.py
```diff
@@ -80,10 +80,7 @@ def list_to_1d_numpy(data, dtype=np.float32, name='list'):
     elif isinstance(data, Series):
         if _get_bad_pandas_dtypes([data.dtypes]):
             raise ValueError('Series.dtypes must be int, float or bool')
-        if hasattr(data.values, 'values'):  # SparseArray
```
StrikerRUS: Fix `FutureWarning`:
```
FutureWarning: The SparseArray.values attribute is deprecated and will be removed in a future version. You can use `np.asarray(...)` or the `.to_dense()` method instead.
```

python-package/lightgbm/basic.py
```diff
@@ -80,10 +80,7 @@ def list_to_1d_numpy(data, dtype=np.float32, name='list'):
...
-            return data.values.values.astype(dtype)
-        else:
-            return data.values.astype(dtype)
+        return np.array(data, dtype=dtype, copy=False)  # SparseArray should be supported as well
```
StrikerRUS: ### Dense:

```
import numpy as np
import pandas as pd

y = pd.Series([0., 1., 2., 3.])
id(y.values) == id(np.array(y, copy=False))  # True
```

### Sparse:

```
import numpy as np
import pandas as pd

y = pd.Series(pd.SparseArray([0., 1., 2., 3.]))
id(y.values.values) == id(np.array(y, copy=False))  # True
```

python-package/lightgbm/basic.py
```diff
@@ -311,7 +313,9 @@ def _label_from_pandas(label):
             raise ValueError('DataFrame for label cannot have multiple columns')
         if _get_bad_pandas_dtypes(label.dtypes):
             raise ValueError('DataFrame.dtypes for label must be int, float or bool')
-        label = label.values.astype('float').flatten()
+        label = np.ravel(label.values)
```
StrikerRUS: `ravel` vs `flatten`: https://stackoverflow.com/a/28930580

python-package/lightgbm/basic.py
```diff
@@ -80,10 +80,10 @@ def list_to_1d_numpy(data, dtype=np.float32, name='list'):
     elif isinstance(data, Series):
         if _get_bad_pandas_dtypes([data.dtypes]):
             raise ValueError('Series.dtypes must be int, float or bool')
-        if hasattr(data.values, 'values'):  # SparseArray
-            return data.values.values.astype(dtype)
+        if data.dtype == np.float32 or data.dtype == np.float64:
+            return np.array(data, dtype=data.dtype, copy=False)
         else:
-            return data.values.astype(dtype)
+            return np.array(data, dtype=dtype, copy=False)  # SparseArray should be supported as well
```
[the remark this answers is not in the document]
StrikerRUS: Tests had already been there when I joined the project. So, I can't say much about it... I guess, the rule about tests inside the package is crucial for mono-language repos.

python-package/lightgbm/basic.py
```diff
@@ -80,10 +80,10 @@ def list_to_1d_numpy(data, dtype=np.float32, name='list'):
     elif isinstance(data, Series):
         if _get_bad_pandas_dtypes([data.dtypes]):
             raise ValueError('Series.dtypes must be int, float or bool')
-        if hasattr(data.values, 'values'):  # SparseArray
-            return data.values.values.astype(dtype)
+        if data.dtype == np.float32 or data.dtype == np.float64:
+            return np.array(data, dtype=data.dtype, copy=False)
```
StrikerRUS: @guolinke According to this note
https://github.com/microsoft/LightGBM/blob/15566421bcc96d46bac353dee17b981a12827604/include/LightGBM/c_api.h#L310-L315
it seems that we cannot allow these memory savings, right?

guolinke: sorry for missing this comment.
yes, the `field` in dataset don't allow arbitrary data types.

guolinke: I think I could have a parameter, maybe called `new_type` with default value `None`. And when it is not none, convert to the new type if it is not that type.

StrikerRUS: @guolinke Seems we always need to force the type, except training/test data, but it's not going through this function. So, I can simply remove that `if` statement.

StrikerRUS: ping @guolinke

PR #2389, opened 2019-09-07
[ci] add if-else job to Travis

StrikerRUS:
To be honest, I have no idea when it was lost from there.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ env:
     - TASK=regular PYTHON_VERSION=3.6
     - TASK=sdist PYTHON_VERSION=2.7
     - TASK=bdist
+    - TASK=if-else
     - TASK=pylint
     - TASK=check-docs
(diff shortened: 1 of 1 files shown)


StrikerRUS:
@jameslamb
> Anyway, looks good to me! The only thing failing on Travis is the check-docs task, link checker is getting a 502 from https://www.mcs.anl.gov/~thakur/papers/ijhpca-coll.pdf

Thanks for your review! Yeah, this site is down again. 😠 I think till Monday, based on the previous experience.

> what does if-else task do?

I converts a model, refer to
https://lightgbm.readthedocs.io/en/latest/Parameters.html#convert_model
https://github.com/microsoft/LightGBM/blob/b6d4ad8361379a902091e3a584491c39fa89ef2d/.ci/test.sh#L60-L67

StrikerRUS:
Close-reopen for CI.

Issue #2302, opened 2019-08-01
feature requests


guolinke:
This issue is to maintain all features request on one page.

Note to contributors: If you want to work for a requested feature, re-open the linked issue. Everyone is welcome to work on any of the issues below.

Note to maintainers: All feature requests should be consolidated to this page. When there are new feature request issues, close them and create the new entries, with the link to the issues, in this page. 

- [ ] More tests (#261 )
- [ ] Matlab support (#743)
- [ ] String as categorical input directly (#789)
- [ ] AWS S3 support(#1039 )
- [ ] CoreML support(#1074)
- [ ] Parquet file support (#1286)
- [ ] Object importantce (#1460)
- [ ] Include init_score in predict function (#1978)
- [ ] Multiple file as input (#2031)
- [ ] Extracting decision path(#2187)

new algorithm:
- [ ] Regularized Greedy Forest (#315)
- [ ] Accelerated Gradient Boosting (#1257)
- [ ] Piece-wise linear tree(#1315)
- [ ] Multi layer gradient boosting tree(#1423)
- [ ] Adaptive neural tree(#1542) 
- [ ] Probabilistic Random Forest(#1946)
- [ ] Sparrow(#2001)

objective and metric:
- [ ] Multi-output regression (#524)
- [ ] Earth Mover Distance (#1256)
- [ ] Cox Proportional Hazard Regression (#1837)
- [ ] Ranking metric for regression objective (#1911)
- [ ] Density estimation (#2056)

python package:
- [ ] Keep cv predicted values (#283)
- [ ] Python input checking (#812)
- [ ] Log redirect in python (#1493)

R package:
- [ ] Yarn support (#790)
- [ ] Plotting in R-package(#1222)
- [ ] H2O datatable support (#2003)
- [ ] [R-package] Allow data to live in memory mapped file (#2148)

GPU:
- [ ] Multiple GPU support (#620)
- [ ] GPU performance improvement (#768)
- [ ] GPU binarines release (#2263)

MotoRZR:
For everyone's information, the sparrow algorithm has been implemented in CatBoost

MotoRZR:
There’s a reference to minimum variance sampling here:

https://catboost.ai/docs/concepts/algorithm-main-stages_bootstrap-options.html

Although I think it just speeds up training rather than providing out of core training.

momijiame:
[...]

PR #2400, opened 2019-09-11
Fix Integer Overflow #2357 

### LINKED ISSUE #2357: lgb.cv is constantly crashing in R
Unfortunately, my R Session is always aborted after a few seconds when running lgb.cv. I call lgb.cv several times since it's part of a scoring function I use to do Bayesian Optimization and hence find the optima hyper parameters. However, the error does seem to come from lightgbm. 

In RStudio, I get the following error:

[LightGBM] [Fatal] Check failed: cache_size >= 2 at /private/var/folders/6q/.../T/RtmpGrFClE/R.INSTALL10ba5306a71e/lightgbm/src/src/treelearner/feature_histogram.hpp, line 676.

If you have any idea what is this supposed to mean and how I could solve the issue, I would highly appreciate your help.

chris-smith-zocdoc:
Fixes #2357

Please let me know if this is what you meant in your comment @guolinke 

Where should I add tests for this?


chris-smith-zocdoc:
> Please use two-space indent, not 4.

Fixed. We could add an https://editorconfig.org/ file to automatically configure editor settings for people who use it. [Example](https://github.com/exercism/cpp/blob/master/.editorconfig)

StrikerRUS:
How about simple explicit prohibition of large `max_depth` values in config as per https://github.com/microsoft/LightGBM/issues/2357#issuecomment-530402138?

`  // check = <=16`
https://github.com/microsoft/LightGBM/blob/b310fb417e4078cc30b153a5ef8f60c06d9ca9cf/include/LightGBM/config.h#L214-L216

StrikerRUS:
> We could add an https://editorconfig.org/ file to automatically configure editor settings for people who use it. Example

Thanks a lot! Cool feature! I'm adding this to our **feature requests** (#2302) near the cpp lint test.

chris-smith-zocdoc:
@guolinke @StrikerRUS Would you like anything else added to this pr? Should I add tests or update docs?

guolinke:
ping @StrikerRUS for the test and doc.

StrikerRUS:
@chris-smith-zocdoc @guolinke I think that new fatal error message is quite clear and no clarifications in docs are needed.

chris-smith-zocdoc:
Thanks @StrikerRUS I think this is ready to merge then.

src/io/config.cpp
```diff
@@ -307,12 +307,21 @@ void Config::CheckParamConflict() {
...
   if (max_depth > 0) {
-    int full_num_leaves = static_cast<int>(std::pow(2, max_depth));
+    double full_num_leaves = std::pow(2, max_depth);
     if (full_num_leaves > num_leaves
         && num_leaves == kDefaultNumLeaves) {
       Log::Warning("Accuracy may be bad since you didn't set num_leaves and 2^max_depth > num_leaves");
     }
-    num_leaves = std::min(num_leaves, 2 << max_depth);
+    
+    const int MAX_LEAVES = 2 << 16;
```
StrikerRUS: Sorry, why do we use `16` here?

chris-smith-zocdoc: Only because @guolinke suggested it https://github.com/microsoft/LightGBM/issues/2357#issuecomment-530402138

StrikerRUS: Ah, OK!
Why do we perform this check only in case of `if (max_depth > 0)`?

chris-smith-zocdoc: No reason I guess, I put it there while I was adding the guard for the overflow. But now that you mention it, I don't see a reason why it couldn't be outside this block

StrikerRUS: I think this check should be performed independently of the `max_depth` value. In this case, the check and `kMaxNumLeaves` can be moved to `config.h` and simplify the code.

chris-smith-zocdoc: I moved the constant to config.h and the check outside the max_depth check.

I'm not sure where I'd place the check inside `config.h` If you want, feel free to add commits to this pr directly

src/io/config.cpp
```diff
@@ -306,13 +306,21 @@ void Config::CheckParamConflict() {
...
   if (max_depth > 0) {
-    int full_num_leaves = static_cast<int>(std::pow(2, max_depth));
+    double full_num_leaves = std::pow(2, max_depth);
     if (full_num_leaves > num_leaves
         && num_leaves == kDefaultNumLeaves) {
       Log::Warning("Accuracy may be bad since you didn't set num_leaves and 2^max_depth > num_leaves");
     }
-    num_leaves = std::min(num_leaves, 2 << max_depth);
+
+    if (full_num_leaves < kMaxNumLeaves && full_num_leaves < num_leaves) {
```
guolinke: as `num_leaves <= kMaxNumLeaves`, do we need to check `full_num_leaves < kMaxNumLeaves` here ?

chris-smith-zocdoc: Yes to prevent the overflow during the double to int cast

src/io/config.cpp
StrikerRUS: > I'm not sure where I'd place the check inside config.h If you want, feel free to add commits to this pr directly

Unfortunately, I have no permission to commit in your branch.
https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork

I would like to replace `if (num_leaves > kMaxNumLeaves)` in `config.cpp` with `// check = <=131072` in `config.h` so that the check will be created automatically and users will be aware of the max value early during reading the docs.

chris-smith-zocdoc: updated

src/io/config.cpp
chris-smith-zocdoc: Alternative I can make this `num_leaves  <= INT_MAX` if that makes this more clear

StrikerRUS: @chris-smith-zocdoc  I'm sorry, but I think that `full_num_leaves < kMaxNumLeaves` is excess too, because the second part `full_num_leaves < num_leaves` is more strict due to that we have a guarantee `num_leaves <= kMaxNumLeaves` (from `config.h`). Can you please elaborate a little bit more why it's needed?

chris-smith-zocdoc: Ah yes, you are correct. I'll make the change

StrikerRUS:
@guolinke Can we merge this?

Issue #2360, opened 2019-08-27
Missing JNI Exception Checks


chris-smith-zocdoc:
I'm attempting to track down a memory corruption bug that is occurring during training of LightGBM on spark (using https://github.com/Azure/mmlspark) 

But when I enable the jvm options `-Xcheck:jni -verbose:jni` I am seeing an excessive number of warnings suggesting that errors aren't being checked in the jni code. 

I haven't written jni before, but only call to `CallObjectMethod` is [here](https://github.com/microsoft/LightGBM/blob/master/swig/lightgbmlib.i#L157-L158)

A quick google search indicates that we should be using `env->ExceptionCheck()` or `env-> ExceptionOccurred()` after calling methods that can throw exceptions https://www.developer.com/java/data/exception-handling-in-jni.html

## Environment info

Databricks 5.5 ML Runtime

Operating System:
```
4.4.0-1087-aws

Ubuntu 16.04.6 LTS
```

CPU/GPU model:
```
AWS r4.xlarge

Intel(R) Xeon(R) CPU E5-2686 v4 @ 2.30GHz
```

LightGBM version or commit hash:
```
com.microsoft.ml.lightgbm lightgbmlib-2.2.350.jar
```

JVM Info
```
openjdk version "1.8.0_212"
OpenJDK Runtime Environment (build 1.8.0_212-8u212-b03-0ubuntu1.16.04.1-b03)
OpenJDK 64-Bit Server VM (build 25.212-b03, mixed mode)
```

## Error message

```
WARNING in native method: JNI call made without checking exceptions when required to from CallObjectMethodV
	at com.microsoft.ml.lightgbm.lightgbmlibJNI.LGBM_DatasetCreateFromCSRSpark(Native Method)
	at com.microsoft.ml.lightgbm.lightgbmlib.LGBM_DatasetCreateFromCSRSpark(lightgbmlib.java:286)
	at com.microsoft.ml.spark.lightgbm.LightGBMUtils$.generateSparseDataset(LightGBMUtils.scala:246)
	at com.microsoft.ml.spark.lightgbm.TrainUtils$.generateDataset(TrainUtils.scala:79)
	at com.microsoft.ml.spark.lightgbm.TrainUtils$.translate(TrainUtils.scala:258)
	at com.microsoft.ml.spark.lightgbm.TrainUtils$.trainLightGBM(TrainUtils.scala:411)
[...]

PR #2405, opened 2019-09-13
[Java] add missing JNI exception checks to fix warnings

imatiach-msft:
fixes the warnings as described in https://github.com/microsoft/LightGBM/issues/2360
by @chris-smith-zocdoc
FYI @eisber

### DIFF
--- a/swig/lightgbmlib.i
+++ b/swig/lightgbmlib.i
@@ -133,6 +133,15 @@
           int size;
   };
 
+  void check_exception(JNIEnv *jenv) {
+    jboolean flag = jenv->ExceptionCheck();
+    if (flag) {
+      jthrowable jerror = jenv->ExceptionOccurred();
+      jenv->ExceptionClear();
+      jenv->Throw(jerror);
+    }
+  }
+
   int LGBM_DatasetCreateFromCSRSpark(JNIEnv *jenv,
                                      jobjectArray arrayOfSparseVector,
                                      int num_rows,
@@ -156,7 +165,9 @@
 
       // get the size, indices and values
       auto indices = (jintArray)jenv->CallObjectMethod(objSparseVec, sparseVectorIndices);
+      check_exception(jenv);
       auto values = (jdoubleArray)jenv->CallObjectMethod(objSparseVec, sparseVectorValues);
+      check_exception(jenv);
       int size = jenv->GetArrayLength(indices);
 
       // Note: when testing on larger data (e.g. 288k rows per partition and 36mio rows total)
--- a/swig/lightgbmlib.i
+++ b/swig/lightgbmlib.i
@@ -157,11 +157,11 @@
       // get the size, indices and values
       auto indices = (jintArray)jenv->CallObjectMethod(objSparseVec, sparseVectorIndices);
       if (jenv->ExceptionCheck()) {
-	return -1;
+        return -1;
       }
       auto values = (jdoubleArray)jenv->CallObjectMethod(objSparseVec, sparseVectorValues);
       if (jenv->ExceptionCheck()) {
-	return -1;
+        return -1;
       }
       int size = jenv->GetArrayLength(indices);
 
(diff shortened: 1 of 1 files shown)


swig/lightgbmlib.i
```diff
@@ -133,6 +133,15 @@
           int size;
   };
 
+  void check_exception(JNIEnv *jenv) {
+    jboolean flag = jenv->ExceptionCheck();
+    if (flag) {
+      jthrowable jerror = jenv->ExceptionOccurred();
+      jenv->ExceptionClear();
+      jenv->Throw(jerror);
```
eisber: I'm not sure this does what you want. jenv->Throw() will not throw, but rather mark the Java stack to throw a Java exception. if you call another Java method from the JNI side it will crash (at least in my local tests)

chris-smith-zocdoc: I think we should have `LGBM_DatasetCreateFromCSRSpark` exit early and return `-1` in addition to having the java side throw.

imatiach-msft: done, I also validated by adding:

```
jenv->GetObjectArrayElement(arrayOfSparseVector, -1);
```

prior to check_exception, and the program seemed to fail gracefully with:

```
Exception in thread "main" java.lang.ArrayIndexOutOfBoundsException: -1
	at com.microsoft.ml.lightgbm.lightgbmlibJNI.LGBM_DatasetCreateFromCSRSpark(Native Method)
	at com.microsoft.ml.lightgbm.lightgbmlib.LGBM_DatasetCreateFromCSRSpark(lightgbmlib.java:290)
	at program.generateSparseDataset(program.java:22)
	at program.main(program.java:54)
```

imatiach-msft: and of course, I also validated that the warnings are still fixed

swig/lightgbmlib.i
```diff
@@ -133,6 +133,17 @@
           int size;
   };
 
+  int check_exception(JNIEnv *jenv) {
+    jboolean flag = jenv->ExceptionCheck();
+    if (flag) {
+      jthrowable jerror = jenv->ExceptionOccurred();
```
eisber: do you really need to retrieve the exception and set it again? when I tested if you don't clear, there is no need to Throw again.

imatiach-msft: good point, since I'm re-throwing the same error I guess I might as well just return without clearing

https://www.developer.com/java/data/exception-handling-in-jni.html

imatiach-msft: I'm seeing this error if I don't rethrow, which looks good

Exception in thread "main" java.lang.ArrayIndexOutOfBoundsException: -1
[Dynamic-linking native method java.lang.Throwable.getStackTraceDepth ... JNI]
[Dynamic-linking native method java.lang.Throwable.getStackTraceElement ... JNI]
	at com.microsoft.ml.lightgbm.lightgbmlibJNI.LGBM_DatasetCreateFromCSRSpark(Native Method)
	at com.microsoft.ml.lightgbm.lightgbmlib.LGBM_DatasetCreateFromCSRSpark(lightgbmlib.java:286)
	at program.generateSparseDataset(program.java:24)
	at program.main(program.java:59)

swig/lightgbmlib.i
```diff
@@ -156,7 +156,13 @@
 
       // get the size, indices and values
       auto indices = (jintArray)jenv->CallObjectMethod(objSparseVec, sparseVectorIndices);
+      if (jenv->ExceptionCheck()) {
+	return -1;
```
StrikerRUS: @imatiach-msft  Please don't use tabs
https://github.com/microsoft/LightGBM/blob/023749239ebf390b5d56e08f227a013cbd972eb0/.editorconfig#L7-L8

imatiach-msft: done!  @StrikerRUS sorry about that!  I really should figure out how to configure my emacs to not use tabs when I have the time...

eisber:
looks good to me

StrikerRUS:
@guolinke Can we merge this?

PR #2414, opened 2019-09-16
[docs] added installation guide with Homebrew 

### LINKED ISSUE #2408: Homebrew Formula for LightGBM
Upstream issue: https://github.com/Homebrew/homebrew-core/pull/43244.

@ankane Thanks a lot for your efforts!

Please ping us if any help is needed from our side.

Want to hear opinion of macOS users @jameslamb @Laurae2 @henry0312.

StrikerRUS:
Closed #2408.

### DIFF
--- a/docs/Installation-Guide.rst
+++ b/docs/Installation-Guide.rst
@@ -121,9 +121,9 @@ Only **Apple Clang** version 8.1 or higher is supported.
 Install Using ``Homebrew``
 **************************
 
-   .. code::
+.. code::
 
-     brew install lightgbm
+  brew install lightgbm
 
 Install from GitHub
 *******************
--- a/docs/Installation-Guide.rst
+++ b/docs/Installation-Guide.rst
@@ -118,6 +118,16 @@ Apple Clang
 
 Only **Apple Clang** version 8.1 or higher is supported.
 
+Install Using ``Homebrew``
+**************************
+
+   .. code::
+
+     brew install lightgbm
+
+Install from GitHub
+*******************
+
 1. Install `CMake`_ (3.12 or higher):
 
    .. code::
--- a/docs/Installation-Guide.rst
+++ b/docs/Installation-Guide.rst
@@ -111,7 +111,7 @@ Also, you may want to read `gcc Tips <./gcc-Tips.rst>`__.
 macOS
 ~~~~~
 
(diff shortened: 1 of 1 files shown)


ankane:
Thank you both.

@StrikerRUS I feel like it probably shouldn't be nested under Clang, since it's already pre-compiled. I'd probably do something like:

---

On macOS LightGBM can be installed with Homebrew or built using CMake and Apple Clang or gcc.

### Homebrew

...

### Apple Clang

...

### gcc

...

StrikerRUS:
@ankane 
> I feel like it probably shouldn't be nested under Clang

Oh, it was my headache for about an hour when I prepared this PR! Yes, with Homebrew users mostly install pre-compiled bottles. But can these bottles be used with gcc?

ankane:
@StrikerRUS As far as I know, the bottles don't require any compiler. Overall, I'm thinking it be good to separate out the install and build from source instructions. Currently, it looks like the Homebrew instructions are part of the build from source docs (imho).

StrikerRUS:
@ankane 
> As far as I know, the bottles don't require any compiler.

According to [the docs](https://docs.brew.sh/Bottles#format), bottle can be treated as compiled binary file. According to [this part of formula](https://github.com/Homebrew/homebrew-core/blob/921945f412dd8b30af5b8494ab5e6b5a51d7635c/Formula/lightgbm.rb#L15-L33), bottles are built against AppleClang with libomp. I'm not sure that library linked to libomp can run successfully on system with gomp installed. As a side note it's worth to say that there is a lot of headache with supporting projects with OpenMP, e.g.,
https://github.com/microsoft/LightGBM/blob/023749239ebf390b5d56e08f227a013cbd972eb0/.ci/test.sh#L71-L74

So, I think we should inform users about that downloaded bottles will run under AppleClang. For someone it may be important.

> Overall, I'm thinking it'd be good to separate out the install and build from source instructions. Currently, it looks like the Homebrew instructions are part of the build from source docs (imho).

Agree! Seems that using your descriptive sentence with the current structure will help. I'll push changes in a little while.

StrikerRUS:
@ankane It should be more clear now that Homebrew allows installation in contrast to building from sources.

ankane:
Thanks @StrikerRUS. I still think it'll be easier for users to understand if it's not under the Apple Clang section (most Homebrew users shouldn't need to know anything about how it was compiled), but understand if you don't agree.

PR #2407, opened 2019-09-14
code refactoring: cost effective gradient boosting

guolinke:
refer to #2341

PR #2416, opened 2019-09-16
[ci] added cpplint check

StrikerRUS:
Refer to #1990, https://github.com/microsoft/LightGBM/pull/2407#issuecomment-531627164.

### DIFF
--- a/.ci/test.sh
+++ b/.ci/test.sh
@@ -53,7 +53,7 @@ fi
 if [[ $TASK == "lint" ]]; then
     conda install -q -y -n $CONDA_ENV pycodestyle pydocstyle
     pip install --user cpplint
-    cpplint --filter=-build/header_guard,-whitespace/line_length --recursive ./src ./include || exit 0  # TODO: change to -1 after fixing #1990
+    cpplint --filter=-build/header_guard,-whitespace/line_length --recursive ./src ./include || exit 0
     pycodestyle --ignore=E501,W503 --exclude=./compute,./.nuget . || exit -1
     pydocstyle --convention=numpy --add-ignore=D105 --match-dir="^(?!^compute|test|example).*" --match="(?!^test_|setup).*\.py" . || exit -1
     exit 0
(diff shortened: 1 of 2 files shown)


.ci/test.sh
```diff
@@ -50,8 +50,10 @@ if [[ $TRAVIS == "true" ]] && [[ $TASK == "check-docs" ]]; then
     exit 0
 fi
 
-if [[ $TASK == "pylint" ]]; then
+if [[ $TASK == "lint" ]]; then
     conda install -q -y -n $CONDA_ENV pycodestyle pydocstyle
+    pip install --user cpplint
+    cpplint --filter=-build/header_guard,-whitespace/line_length --recursive ./src ./include || exit 0  # TODO: change to -1 after fixing #1990
```
jameslamb: @StrikerRUS I'm a fan of never putting `TODO`s in code, but instead keeping such details in the issues. In this case, you could just add that comment to #1990 if you haven't already.

Also I'm a fan of having commands with more than one argument indented like below, so their git blames are cleaner and so that it's more obvious what they are doing:

```bash
cpplint \
    --filter=-build/header_guard,-whitespace/line_length \
    --recursive \
    ./src \
    ./include \
|| exit 0
```

Up to you on both of these!

StrikerRUS: @jameslamb Thanks for your review!
Agree with your first comment. Will remove `TODO` from the code.

Also, agree with your second comment, but I'm a fan of having consistent state, even if it's not perfect 😃 . I mean, we have a lot of such commands with plenty of args in this (and neighboring) file and this line will be "rara avis" here. So, I think such style improvements should be done at once in a separate PR instead of iterative one-line corrections.

StrikerRUS:
I just changed the order of checks in the latest commit 984d15b6e7bfecc58e2f342123879420c572e049 with the aim to prevent fast exit due to temporary `|| exit 0` and allow pylint take effect.

PR #2420, opened 2019-09-17
fix the objective init issues in distributed mode

### LINKED ISSUE #2419: LightGBM distributed training can get stuck in some edge-case scenarios with labels missing on some partitions
LightGBM distributed training can get stuck in some edge-case scenarios, specifically:

1.) In binary classification scenario where some workers have a dataset with only 1 label instead of 2 labels, the workers with 1 label exit early.  The workers with 2 labels then get stuck waiting on the workers with 1 label.

2.) In multiclass classification scenario, if a worker has a dataset with non-consecutive integer labels, eg 0,1,2,4,5 with "3" missing, for some reason training gets stuck and times out (note unlike the binary classification scenario the worker does not exit successfully).

## Environment info

Operating System: windows, linux, macosx

CPU/GPU model: CPU

C++/Python/R version: Distributed MMLSpark version only

LightGBM version or commit hash: latest master

## Error message

LightGBM times out during training

## Reproducible examples
I have a test case that can reproduce this, but currently I have been able to fix the issue by tracking labels and throwing in case some label is missing or if there are fewer than 2 labels for classification on any of the partitions.

https://github.com/Azure/mmlspark/blob/master/src/test/scala/com/microsoft/ml/spark/lightgbm/split1/VerifyLightGBMClassifier.scala#L337

## Steps to reproduce

1. Create a pre-partitioned dataset where one of the partitions only has a single label but the others have 2 labels
2. Train lightgbm on that distributed dataset, note one of the workers finishes early with Warning message, while others are stuck waiting for it to finish

guolinke:
to fix #2419

### DIFF
--- a/include/LightGBM/network.h
+++ b/include/LightGBM/network.h
@@ -188,7 +188,6 @@ class Network {
     });
     return global;
   }
-
   template<class T>
   static T GlobalSyncUpByMax(T& local) {
     T global = local;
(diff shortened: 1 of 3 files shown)


include/LightGBM/network.h
```diff
@@ -214,25 +213,30 @@ class Network {
...
-        src += type_size;
-        dst += type_size;
-        used_size += type_size;
-      }
-    });
-    return static_cast<T>(global / num_machines_);
+      sizeof(local), sizeof(local),
+      reinterpret_cast<char*>(&global),
+      [](const char* src, char* dst, int type_size, comm_size_t len) {
+        comm_size_t used_size = 0;
```
StrikerRUS: Is it result of the recently added `.editorconfig` file?

guolinke: I edit the code by the vs2019, so I am not sure.

PR #2426, opened 2019-09-20
fix many cpp lint errors

guolinke:
refer to #1990

### DIFF
--- a/include/LightGBM/R_object_helper.h
+++ b/include/LightGBM/R_object_helper.h
@@ -123,26 +123,26 @@ typedef struct VECTOR_SER {
 
 typedef union { VECTOR_SER s; double align; } SEXPREC_ALIGN;
 
-#define DATAPTR(x)  (((SEXPREC_ALIGN *) (x)) + 1)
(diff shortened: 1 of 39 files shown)


guolinke:
BTW, this will have some conflicts with other PRs.
@StrikerRUS I think we should merge this ASAP.

guolinke:
@StrikerRUS yeah, but it seems the reinterpret_cast is the only way, and it actually equals to c style cast.

guolinke:
@StrikerRUS BTW, as `json11` is not maintained by us, we can disable the lint check over it.

StrikerRUS:
@guolinke Yes, but why?.. As it's not a "submodule" (as `compute`) it falls into situation when it's maintained by no one. I think it's not right. Seems that the license allows to modify it. I remember there were some dangerous warnings from cpplint for `json11`.

guolinke:
Okay, I think we won't modify it, so we can keep it as it is. However, if these warnings are easy to fix, we could fix it directly.

PR #2427, opened 2019-09-20
[ci] added build/include_subdir check to filter for cpplint

StrikerRUS:
Refer to https://github.com/microsoft/LightGBM/issues/1990#issuecomment-482695889, https://github.com/microsoft/LightGBM/issues/1990#issuecomment-482776312.


StrikerRUS:
Force-pushed with zero diff to re-trigger CI, because close-reopen doesn't work for branches.

Issue #2335, opened 2019-08-19
R unit tests are not run in CI


jameslamb:
Are the R package's tests run anywhere in our CI? I looked through Travis, Azure, and Appveyor and as far as I can tell they aren't.

I ask because I just submitted #2334 which is passing all checks, but running the unit tests on the R package fails locally for me.

After noticing that, I saw that even on `master` the R unit tests locally for me.

**Steps to reproduce**

Clone `lightgbm` repo, then from the root of the repo:

```bash
# build the package
Rscript build.R

# run the tests
cd R-package/tests
Rscript testthat.R
```

I get one test failure when doing this:

```
── 1. Error: training continuation works (@test_basic.R#71)  ───────────────────
(list) object cannot be coerced to type 'double'
1: lgb.get.eval.result(bst2, "train", "binary_logloss", 10) at testthat/test_basic.R:71
```

**Environment Information**

<details><summary>Mac OS</summary>

![image](https://user-images.githubusercontent.com/7608904/63238894-281f9580-c20e-11e9-948d-16d03ae382a8.png)

</details>

<details><summary>R session</summary>

<p>

```
R version 3.5.0 (2018-04-23)
Platform: x86_64-apple-darwin15.6.0 (64-bit)
Running under: macOS  10.14.5

Matrix products: default
BLAS: /System/Library/Frameworks/Accelerate.framework/Versions/A/Frameworks/vecLib.framework/Versions/A/libBLAS.dylib
LAPACK: /Library/Frameworks/R.framework/Versions/3.5/Resources/lib/libRlapack.dylib

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

attached base packages:
[1] stats     graphics  grDevices utils     datasets 
[6] methods   base     

other attached packages:
[1] lightgbm_2.2.4 R6_2.3.0.9000  testthat_2.2.1

loaded via a namespace (and not attached):
 [1] Rcpp_1.0.0.1      rex_1.1.2         rstudioapi_0.7   
 [4] magrittr_1.5      covr_3.2.1        usethis_1.5.1    
 [7] devtools_2.1.0    pkgload_1.0.2     rlang_0.3.1      
[10] tools_3.5.0       pkgbuild_1.0.4    data.table_1.12.0
[13] sessioninfo_1.1.1 cli_1.1.0         withr_2.1.2.9000 
[...]

PR #1497, opened 2018-07-07
add vsts ci

guolinke:

PR #2443, opened 2019-09-25
[R-package] Disabled early stopping when using 'dart' boosting strategy

jameslamb:
See discussion in #1893 for background. In this PR, I propose disabling early stopping when the user chooses `'dart'` boosting. The same change was made in the Python package in #1895 .

This PR also adds the additional early stopping alias that was added in #2431 .

While working on this piece of code, I refactored it to make it a bit more obvious what is happening. I found that the same code is copied in `lgb.cv()` and `lgb.train()`. I think this duplication could lead to inconsistencies in the future, but I considered reducing that duplication out of scope.

### DIFF
--- a/R-package/R/aliases.R
+++ b/R-package/R/aliases.R
@@ -12,6 +12,12 @@
             , "boost"
             , "boosting_type"
         )
+        , "early_stopping_round" = c(
+            "early_stopping_round"
+            , "early_stopping_rounds"
+            , "early_stopping"
+            , "n_iter_no_change"
+        )
         , "metric" = c(
             "metric"
             , "metrics"
@@ -20,6 +26,17 @@
         , "num_class" = c(
             "num_class"
             , "num_classes"
+        )
+        , "num_iterations" = c(
+            "num_iterations"
+            , "num_iteration"
+            , "n_iter"
+            , "num_tree"
+            , "num_trees"
+            , "num_round"
+            , "num_rounds"
+            , "num_boost_round"
+            , "n_estimators"
         )
     ))
 }
--- a/R-package/R/aliases.R
+++ b/R-package/R/aliases.R
@@ -1,0 +1,15 @@
+# Central location for paramter aliases.
+# See https://lightgbm.readthedocs.io/en/latest/Parameters.html#core-parameters
+
+# [description] List of respected parameter aliases. Wrapped in a function to take advantage of
+#               lazy evaluation (so it doesn't matter what order R sources files during installation).
+# [return] A named list, where each key is a main LightGBM parameter and  each value is a character
+#          vector of corresponding aliases.
+.PARAMETER_ALIASES <- function(){
+    return(list(
+        "boosting" = c(
+            "boosting_type"
+            , "boost"
+        )
+    ))
+}
--- a/R-package/R/aliases.R
+++ b/R-package/R/aliases.R
@@ -8,8 +8,18 @@
 .PARAMETER_ALIASES <- function(){
     return(list(
         "boosting" = c(
-            "boosting_type"
+            "boosting"
             , "boost"
+            , "boosting_type"
+        )
+        , "metric" = c(
+            "metric"
+            , "metrics"
+            , "metric_types"
+        )
+        , "num_class" = c(
+            "num_class"
+            , "num_classes"
         )
     ))
 }
--- a/R-package/R/callback.R
+++ b/R-package/R/callback.R
@@ -38,9 +38,8 @@ cb.reset.parameters <- function(new_params) {
     # Some parameters are not allowed to be changed,
     # since changing them would simply wreck some chaos
     not_allowed <- c(
-      "num_class"
-      , "metric"
-      , "boosting"
+      .PARAMETER_ALIASES()[["num_class"]]
+      , .PARAMETER_ALIASES()[["metric"]]
       , .PARAMETER_ALIASES()[["boosting"]]
     )
     if (any(pnames %in% not_allowed)) {
--- a/R-package/R/callback.R
+++ b/R-package/R/callback.R
@@ -37,7 +37,12 @@ cb.reset.parameters <- function(new_params) {
 
     # Some parameters are not allowed to be changed,
     # since changing them would simply wreck some chaos
-    not_allowed <- c("num_class", "metric", "boosting_type")
+    not_allowed <- c(
+      "num_class"
+      , "metric"
+      , "boosting"
+      , .PARAMETER_ALIASES()[["boosting"]]
+    )
     if (any(pnames %in% not_allowed)) {
       stop(
         "Parameters "
--- a/R-package/R/lgb.cv.R
+++ b/R-package/R/lgb.cv.R
@@ -136,17 +136,7 @@ lgb.cv <- function(params = list(),
     begin_iteration <- predictor$current_iter() + 1
   }
   # Check for number of rounds passed as parameter - in case there are multiple ones, take only the first one
-  n_trees <- c(
-    "num_iterations"
-    , "num_iteration"
-    , "n_iter"
-    , "num_tree"
-    , "num_trees"
-    , "num_round"
-    , "num_rounds"
-    , "num_boost_round"
-    , "n_estimators"
-  )
+  n_rounds <- .PARAMETER_ALIASES()[["num_iterations"]]
   if (any(names(params) %in% n_trees)) {
     end_iteration <- begin_iteration + params[[which(names(params) %in% n_trees)[1]]] - 1
   } else {
@@ -227,7 +217,7 @@ lgb.cv <- function(params = list(),
 
   # If early stopping was passed as a parameter in params(), prefer that to keyword argument
   # early_stopping_rounds by overwriting the value in 'early_stopping_rounds'
-  early_stop <- c("early_stopping_round", "early_stopping_rounds", "early_stopping", "n_iter_no_change")
+  early_stop <- .PARAMETER_ALIASES()[["early_stopping_round"]]
   early_stop_param_indx <- names(params) %in% early_stop
   if (any(early_stop_param_indx)) {
     first_early_stop_param <- which(early_stop_param_indx)[[1]]
@@ -238,7 +228,7 @@ lgb.cv <- function(params = list(),
   # Did user pass parameters that indicate they want to use early stopping?
   using_early_stopping_via_args <- !is.null(early_stopping_rounds)
 
-  boosting_param_names <- c("boosting", .PARAMETER_ALIASES()[["boosting"]])
+  boosting_param_names <- .PARAMETER_ALIASES()[["boosting"]]
   using_dart <- any(
     sapply(
       X = boosting_param_names
--- a/R-package/R/lgb.cv.R
+++ b/R-package/R/lgb.cv.R
@@ -225,30 +225,30 @@ lgb.cv <- function(params = list(),
     callbacks <- add.cb(callbacks, cb.record.evaluation())
   }
 
-  # Check for early stopping passed as parameter when adding early stopping callback
+  # If early stopping was passed as a parameter in params(), prefer that to keyword argument
+  # early_stopping_rounds by overwriting the value in 'early_stopping_rounds'
   early_stop <- c("early_stopping_round", "early_stopping_rounds", "early_stopping", "n_iter_no_change")
-  if (any(names(params) %in% early_stop)) {
-    if (params[[which(names(params) %in% early_stop)[1]]] > 0) {
-      callbacks <- add.cb(
-        callbacks
-        , cb.early.stop(
-          params[[which(names(params) %in% early_stop)[1]]]
-          , verbose = verbose
-        )
+  early_stop_param_indx <- names(params) %in% early_stop
+  if (any(early_stop_param_indx)) {
+    first_early_stop_param <- which(early_stop_param_indx)[[1]]
+    first_early_stop_param_name <- names(params)[[first_early_stop_param]]
+    early_stopping_rounds <- params[[first_early_stop_param_name]]
+  }
+
+  using_early_stopping <- !is.null(early_stopping_rounds)
+  if (using_early_stopping && identical(params$boosting, "dart")){
+    warning("Early stopping is not available in 'dart' mode")
+    use_early_stopping <- FALSE
+  }
+
+  if (using_early_stopping){
+    callbacks <- add.cb(
+      callbacks
+      , cb.early.stop(
+        stopping_rounds = early_stopping_rounds
+        , verbose = verbose
       )
-    }
-  } else {
-    if (!is.null(early_stopping_rounds)) {
-      if (early_stopping_rounds > 0) {
-        callbacks <- add.cb(
-          callbacks
-          , cb.early.stop(
-            early_stopping_rounds
-            , verbose = verbose
-          )
-        )
-      }
-    }
+    )
   }
 
   # Categorize callbacks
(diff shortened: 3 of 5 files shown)


StrikerRUS:
@jameslamb Please see [this my comment](https://github.com/microsoft/LightGBM/issues/1893#issuecomment-476201045).

jameslamb:
> @jameslamb Please see [this my comment](https://github.com/microsoft/LightGBM/issues/1893#issuecomment-476201045).

@StrikerRUS apologies, I actually meant to address that comment in PR description. What do you think should be done? Since users are just passing in functions to `callbacks`, I don't know how to tell whether those functions are for early stopping or some other purpose.

Are you just suggesting that I invoke this warning WHENEVER someone uses `dart` boosting, so that people who passed early stopping custom callbacks will know?

StrikerRUS:
@jameslamb In my comment from the original issue I meant that this fix covers the case when user passes `boosting = 'dart'` and `early_stopping_rounds = x`. However, it seems that the fix doesn't cover the case when user passes `boosting = 'dart'` and `callbacks = list(cb.early.stop(x))`. Am I wrong? The Python-package fix covers the both cases. I think it's possible just to mimic the Python code fix or
> Since users are just passing in functions to callbacks, I don't know how to tell whether those functions are for early stopping or some other purpose.

https://github.com/microsoft/LightGBM/blob/1f88721a7982a804731b1dc108ee40625405639c/R-package/R/callback.R#L386

maybe somehow use this information?

Laurae2:
@jameslamb @StrikerRUS From what I read, is it `any(unlist(lapply(callback, function(x) {attr(x, "name")})) == "cb.early.stop")` which should be also added as a condition?

StrikerRUS:
@jameslamb @Laurae2 Maybe simply move the check into the callback function as it was done for Python-package? https://github.com/microsoft/LightGBM/pull/1895/files

jameslamb:
> @jameslamb @StrikerRUS From what I read, is it `any(unlist(lapply(callbacks, function(x) {attr(x, "name")})) == "cb.early.stop")` which should be also added as a condition?

to your @StrikerRUS 's comment...yes I could add that and am happy to, but `cb.early.stop()` is not exported from the package (https://github.com/microsoft/LightGBM/blob/master/R-package/NAMESPACE). So we can add that check (or something similar) to 100% prevent "any early stopping callbacks that are part of the LightGBM package", but I don't think there is anything we can do about the case where some user writes their own totally-custom thing and passing it to `callbacks`. I think that's fine, just want to clarify the difference.

As far as pushing the logic down into `callback()` like #1895 , I personally could not figure out how to do that in the R package. The structure of `callback.R` is quite a bit different from `callback.py`, as far as I can tell. Let me take another look at `cb.early.stop()` and see if I can mimic the behavior.

StrikerRUS:
@jameslamb 
> ... but cb.early.stop() is not exported from the package 

Ah, I didn't know that!
So, I added known issue 6 in https://github.com/microsoft/LightGBM/issues/1143#issuecomment-526936690 by a mistake, right?

> but I don't think there is anything we can do about the case where some user writes their own totally-custom thing and passing it to callbacks.

I don't think that someone would write a custom callback for early stopping, since we already have early stopping callback. Even if one will, they will use our implementation as a reference and will see the prohibition of using early stopping with the `dart` mode.

jameslamb:
> @jameslamb
> 
> > ... but cb.early.stop() is not exported from the package
> 
> Ah, I didn't know that!
> So, I added known issue 6 in [#1143 (comment)](https://github.com/microsoft/LightGBM/issues/1143#issuecomment-526936690) by a mistake, right?
> 
> > but I don't think there is anything we can do about the case where some user writes their own totally-custom thing and passing it to callbacks.
> 
> I don't think that someone would write a custom callback for early stopping, since we already have early stopping callback. Even if one will, they will use our implementation as a reference and will see the prohibition of using early stopping with the `dart` mode.

I don't think that known issue was a mistake, maybe you just meant something different than what I thought. We do have a keyword argument `callbacks` in `lgb.train()` and `lgb.cv()`, so I assumed your comment was about more thoroughly documenting the expect form of functions passed to that argument. Right now the document just says "A list of callback functions to evaluate at each iteration"

So our approach right now is to say "you can enable early stopping by passing `early_stopping_rounds` and providing a validation set to `valids`", not "add function `cb.early_stop()` to `callbacks` argument". `callbacks` is a generic mechanism the R package uses for "any code you want to run after each iteration". By the way this is the closest I've looked at this section of the code in a while so @Laurae2 may correct me, but I think that is how it works.

I think for now it would be sufficient to add the check using `attr()` to see if `cb.early.stop` has added to the list in `callbacks`.

StrikerRUS:
@jameslamb Nope, I thought right about documenting functions in `callback.R`, because I was sure that users are encouraged take them and pass to `callbacks` argument.

jameslamb:
> @jameslamb Nope, I thought right about documenting functions in `callback.R`, because I was sure that users are encouraged take them and pass to `callbacks` argument.

Ok then yes, I misinterpreted what you meant. Documenting would also mean exporting. This is one of those things where R and Python are really different...in R you have to explicitly declare things as exported to make them usable. I'll update the language in #1143 to say "export callback functions" if that's alright.

jameslamb:
Ok @Laurae2 @StrikerRUS whenever you have time, could you take another look? I've added protection against `cb.early.stop()` being passed directly via `callbacks` parameter.

I still left this in `lgb.train()` and `lgb.cv()` because I couldn't figure out how to do it all the way down in `callback()` like in the Python package.

StrikerRUS:
@jameslamb 
> I'll update the language in #1143 to say "export callback functions" if that's alright.

I think that then this is another issue, even more feature request. But anyway, it's not related to pkgdown issue. I'll remove it from that list. Can you please create a separate issue for exporting callback functions?

R-package/R/lgb.cv.R
```diff
@@ -211,18 +211,42 @@ lgb.cv <- function(params = list(),
...
+    first_early_stop_param <- which(early_stop_param_indx)[[1]]
+    first_early_stop_param_name <- names(params)[[first_early_stop_param]]
+    early_stopping_rounds <- params[[first_early_stop_param_name]]
+  }
+
+  # Did user pass parameters that indicate they want to use early stopping?
+  using_early_stopping_via_args <- !is.null(early_stopping_rounds)
+
+  # Cannot use early stopping with 'dart' boosting
+  if (identical(params$boosting, "dart")){
```
StrikerRUS: Does it cover cases when `dart` is passed via [boosting aliases](https://lightgbm.readthedocs.io/en/latest/Parameters.html#boosting)?

jameslamb: Seems we don't actually use boosting aliases in the R package at all currently. We will pass through anything you pass into `...` in `lgb.train()` or `lgb.cv()` but we have no code I see in the R package that explictly needs to access that value. I'll add them to this PR since it seems like it would be in scope.

jameslamb: Ok I just pushed https://github.com/microsoft/LightGBM/pull/2443/commits/cab2bcdd011e85c4dda008e69acc2ac390b63674 which adds the boosting aliases. I also added a unit test in that commit which will check that the warning is generated for all the aliases.

I think this approach of centralizing the alias map is preferable to having the aliases hard-coded as string literals in multiple places. For now I only added the boosting-related ones as that's within the scope of this PR. Let me know what you think about the approach! If you and @Laurae2 are ok with it, we could do that for all the aliases (in a separate PR later)

jameslamb:
> @jameslamb
> 
> > I'll update the language in #1143 to say "export callback functions" if that's alright.
> 
> I think that then this is another issue, even more feature request. But anyway, it's not related to pkgdown issue. I'll remove it from that list. Can you please create a separate issue for exporting callback functions?

Ok makes sense. I've created #2479 and attached it to #2302

R-package/R/lgb.cv.R
StrikerRUS: @jameslamb 
> we could do that for all the aliases (in a separate PR later)

I don't think we need a map for aliases of **all parameters**, since they are handled at cpp side. Only aliases of params which are used at R side should be there, I guess.

R-package/R/callback.R
```diff
@@ -37,7 +37,12 @@ cb.reset.parameters <- function(new_params) {
 
     # Some parameters are not allowed to be changed,
     # since changing them would simply wreck some chaos
-    not_allowed <- c("num_class", "metric", "boosting_type")
+    not_allowed <- c(
+      "num_class"
+      , "metric"
+      , "boosting"
+      , .PARAMETER_ALIASES()[["boosting"]]
```
StrikerRUS: I think `boosting` itself can be in returned values, so that only `.PARAMETER_ALIASES()[["boosting"]]` will be needed instead of current `"boosting", .PARAMETER_ALIASES()[["boosting"]]`.

StrikerRUS:
@jameslamb I'm sorry, but could you please explain what is wrong with the following try to mimic Python solution on [pseudo?]R? https://www.diffchecker.com/LylbJ8EG

R-package/R/callback.R
jameslamb: you are right. If aliases are all in the cpp library there shouldn't a reason to care (in language wrappers) about what the "main" one is and what the "alias" ones are. I'll fix that.

jameslamb:
> @jameslamb I'm sorry, but could you please explain what is wrong with the following try to mimic Python solution on [pseudo?]R? https://www.diffchecker.com/LylbJ8EG

I have never seen diffchecker.com before, super cool!

In my opinion, this code proposed in your link is going to be difficult to debug and reason about.

It's relying on a flag called `enabled` being present one level up the call stack (since it is not passed into `init()` and `params` being available an unknown number of levels up the call stack (since it is not passed in to `init()` or to `cb.early.stop()`.

So in my opinion while my proposal in this PR feels more tedious, it is also more _obvious_ and easier to debug / change.

I made that choice partially because I just feel it is a better design for the R package, but also partially because I am not so familiar with all of the R code package code that I could know the full implications of doing something like you proposed.

I will defer to @Laurae2 (who understand the internals of the R package more thoroughly) though. If @Laurae2 prefers something like the way this was handled in the Python package and pseudocode in your in link here, I'll happily change the approach from this PR and push the logic down into `cb.early.stop()`.

R-package/R/lgb.cv.R
jameslamb: Agree, sorry, I also mean "all aliases used in the R side"

StrikerRUS:
@jameslamb Thanks a lot for your explanation!

jameslamb:
I just rebased this to current `master`. ping @Laurae2 for a review

jameslamb:
I just rebased this to latest `master` to catch up with changes over the last two weeks. It is ready for review.

Laurae2:
@jameslamb As you pointed out there's a lot of things in common in `lgb.train` and `lgb.cv` which are still separated, you did the right things at the moment to duplicate the code.

Ideally in the future we should be able to merge all the duplicate code in one, but it requires rewriting the R package.

R-package/R/aliases.R
```diff
@@ -0,0 +1,15 @@
+# Central location for paramter aliases.
```
StrikerRUS:
```suggestion
# Central location for parameter aliases.
```

R-package/R/aliases.R
```diff
@@ -0,0 +1,15 @@
...
+# See https://lightgbm.readthedocs.io/en/latest/Parameters.html#core-parameters
+
+# [description] List of respected parameter aliases. Wrapped in a function to take advantage of
+#               lazy evaluation (so it doesn't matter what order R sources files during installation).
+# [return] A named list, where each key is a main LightGBM parameter and  each value is a character
```
StrikerRUS:
```suggestion
# [return] A named list, where each key is a main LightGBM parameter and each value is a character
```

R-package/R/callback.R
```diff
@@ -37,7 +37,12 @@ cb.reset.parameters <- function(new_params) {
 
     # Some parameters are not allowed to be changed,
     # since changing them would simply wreck some chaos
-    not_allowed <- c("num_class", "metric", "boosting_type")
+    not_allowed <- c(
+      "num_class"
+      , "metric"
```
StrikerRUS: I think these params have their aliases, just like `boosting`.

R-package/R/callback.R
```diff
@@ -37,7 +37,12 @@ cb.reset.parameters <- function(new_params) {
...
+      , "boosting"
```
StrikerRUS: Refer to https://github.com/microsoft/LightGBM/pull/2443#discussion_r329503844.

R-package/R/callback.R
jameslamb: just changed this so now all valid names ("main" or otherwise) are returned

R-package/R/callback.R
jameslamb: added! good catch

R-package/R/callback.R
jameslamb: changed, thank you! Now it is

```r
not_allowed <- c(
      .PARAMETER_ALIASES()[["num_class"]]
      , .PARAMETER_ALIASES()[["metric"]]
      , .PARAMETER_ALIASES()[["boosting"]]
    )
```

jameslamb:
I've rebased to `master` and implemented requested changes to handling of aliases.

R-package/R/callback.R
StrikerRUS: I think it's much better now! Thanks!

R-package/R/lgb.cv.R
```diff
@@ -134,7 +134,7 @@ lgb.cv <- function(params = list(),
     begin_iteration <- predictor$current_iter() + 1
   }
   # Check for number of rounds passed as parameter - in case there are multiple ones, take only the first one
-  n_trees <- c("num_iterations", "num_iteration", "n_iter", "num_tree", "num_trees", "num_round", "num_rounds", "num_boost_round", "n_estimators")
+  n_rounds <- .PARAMETER_ALIASES()[["num_iterations"]]
   if (any(names(params) %in% n_trees)) {
     end_iteration <- begin_iteration + params[[which(names(params) %in% n_trees)[1]]] - 1
```
StrikerRUS: I guess it should be either `n_trees`, or `n_rounds`.

jameslamb: I am bad at writing code haha

Thank you, good catch. As soon as I do #2335 CI will catch this for us.

jameslamb: ok this is fixed

jameslamb:
Ok this has been rebased to `master` and I've fixed the `n_trees` vs. `n_rounds` thing

jameslamb:
> LGTM, except two actually unresolved comments about grammar from [#2443 (review)](https://github.com/microsoft/LightGBM/pull/2443#pullrequestreview-306246905).

ah! ok fixing

jameslamb:
Thank you for the reviews everyone! I am going to merge this once CI is done.

jameslamb:
was getting this on `check-docs` task on Travis:

> 429: too many requests
> http://www.ozone3d.net/gpu_caps_viewer/

Restarting that task after waiting a few minutes seemed to solve it, but something to keep an eye on I guess.

StrikerRUS:
@jameslamb 
> Restarting that task after waiting a few minutes seemed to solve it, but something to keep an eye on I guess.

Yeah, I see this error very often. And `docs` job is needed to be restarted quite often, that is why we keep it only at Travis where each collaborator has enough right to re-run it (https://github.com/microsoft/LightGBM/pull/1497#discussion_r201301028).

PR #2453, opened 2019-09-26
fix bug in reset config for dataset

### LINKED ISSUE #2452: reset feature_contrib not working
I used feature_contrib to select features( set to 0 to exclude the feature and set to 1 to include feature). That works well if just set feature_contrib before first booster update, but not work if I reset it within callback function. The proof of not working is that the excluded features are still got splitted according to `env.model.features_importance(env.iteration)-env.model.features_importance(env.iteration-1).`

my callback functions is as below:

```
def  reset_feature_contrib(step_size):

    def _callback(env):
        current_fb=env.model.params.get('feature_contri')
        f_imp=env.model.feature_importance(importance_type='split',iteration=env.model.num_trees())
        last_imp=env.model.feature_importance(importance_type='split',iteration=env.iteration)
        masklist=pd.Series(f_imp).sort_values(ascending=False).head(10).index.values

        if env.iteration%step_size!=0 or env.iteration==0:
            print(f'masked index is {masklist},last split num is {(f_imp-last_imp)[masklist]}')
            return
        else:
            new_fb=np.array(current_fb)
            new_fb[masklist]=0
            print(f'[set for next iter]masked index is {masklist}')
            env.model.reset_parameter({'feature_contri':new_fb.tolist()})
            return

    return _callback
```

guolinke:
to fix #2452

### DIFF
--- a/src/treelearner/feature_histogram.hpp
+++ b/src/treelearner/feature_histogram.hpp
@@ -748,6 +748,7 @@ class HistogramPool {
       OMP_LOOP_EX_END();
     }
     OMP_THROW_EX();
+    train_data_ = train_data;
   }
 
   void ResetConfig(const Config* config) {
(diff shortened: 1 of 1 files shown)


aprilffff:
> to fix #2452

should add `const Dataset* train_data` in function declaration?

aprilffff:
still not work

guolinke:
@aprilffff 
before `nv.model.reset_parameter({'feature_contri':new_fb.tolist()})`, you should try to reset parameter to train-set as well.  https://github.com/microsoft/LightGBM/blob/00d1e6938a3f8566526d470057f8d26ec1289208/python-package/lightgbm/basic.py#L1167-L1175

aprilffff:
do I need _reverse_update_params after _update_params?

aprilffff:
Dont worry, your advice is right. It's completely fixed now. Thx.

Issue #812, opened 2017-08-09
[python package]: suggestion: lgb.Booster.predict() should check that the input X data makes sense


j-mark-hou:
In particular, I'm thinking about these things:

1. if the input is an np.array, check that the columns is the same as the number of features the lgb.Booster object uses.  if not, throw a warning.
1. if the input is a pd.Dataframe object, should check that the feature_names of the lgb.Booster object is a superset of the columns of the pd.Dataframe
2. if feature names in the booster object are repeated, or if column names in the pd.Dataframe are repeated, fall back to 0.

if these things sound reasonable I'd be happy to add these checks to the lgb.Booster.predict() function prior to calling the lgb._InnerPredictor.predict()

guolinke:
@j-mark-hou  sure, very happy to see this feature.

guolinke:
@j-mark-hou any updates ?

j-mark-hou:
I think maybe this should be rolled into a more systematic rewrite of the pandas api.  There's some current things about the implementation that I don't quite understand, and unfortunately I don't currently have the time to dig deeper into this.  Sorry, I'll let you know if I revisit this at some point in the future.

arsenyinfo:
Currently, it leads to some kind of inconsistency:

```
In [1]: from lightgbm import LGBMClassifier, Dataset, train
   ...: import numpy as np
   ...:
   ...: x_data, y_data = np.random.rand(1000, 100), np.random.rand(1000) > .5
   ...: x_bad = np.random.rand(1000, 101)
   ...:
   ...:
   ...: def sklearn_style():
   ...:     clf = LGBMClassifier()
   ...:     clf.fit(x_data, y_data)
   ...:     return clf.predict(x_bad)
   ...:
   ...:
   ...: def xgboost_style():
   ...:     dataset = Dataset(x_data, y_data)
   ...:     params = {'application': 'binary'}
   ...:     booster = train(params, dataset)
   ...:     return booster.predict(x_data)
   ...:

In [2]: sklearn_style().shape
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-2-281be6aeee0d> in <module>()
----> 1 sklearn_style().shape

<ipython-input-1-adc7de9c0c86> in sklearn_style()
      9     clf = LGBMClassifier()
     10     clf.fit(x_data, y_data)
---> 11     return clf.predict(x_bad)
     12
     13

[...]

Issue #2457, opened 2019-09-26
Ranking model giving different results when model.txt file loaded in python.


nalin-mathur:
Hi, 
I have trained a ranking model using the executable lightgbm (obtained after building directory via cmake and then running make -j4) file along with the config file LightGBM/examples/lambdarank/train.conf using my own data. The model is saved in the LightGBM_model.txt file and then the command './lightgbm config='predict.conf' is used to make predictions using the saved model. 
The predictions are stored in LightGBM_predict_result.txt within the same directory. 

I wanted to load the model saved in LightGBM_model.txt in a python script so that the model can be further deployed into production. 

I used the following code to load model into python and make predictions:
```
from sklearn.datasets import load_svmlight_file
import lightgbm as lab
x_test, _ = load_svmlight_file('test_data.txt')
model = lgb.Booster(model_file='LightGBM_model.txt')
y_predictions = model.predict(x_test)
```
where test_data.txt is the test data in libsvm format. 

The results obtained using the two methods (loading model in python using lgb.Booster method and using lightgbm executable file along with predict.conf) are different. Why is this happening and how can this be rectified? 

I also tried changing the x_test from a scipy.sparse.csr.csr_matrix to a numpy.ndarray and it had no effect as expected.

guolinke:
you can try to set `num_iterations` in predict explicitly. python-package will automatically use the best iteration, while cli version doesn't.

nalin-mathur:
The number of iterations was set to -1 which automatically takes the boosted model to the last iteration. There were no changes in the results after I explicitly changed the `num_iterations` in predict.

guolinke:
just saw your use `load_svmlight_file`, which is the root cause.
Refer to https://github.com/microsoft/LightGBM/issues/1908 and https://github.com/microsoft/LightGBM/issues/1776.

LightGBM use the zero-based libsvm file (when pass libsvm file to LightGBM), while `load_svmlight_file` is one-based. 
So if your model is train by the one-based format, you should use the one-based data for the prediction, vice versa.

guolinke:
@StrikerRUS there are several issues regard to this. Should I add it to the FAQ?
BTW, #2464 could raise the error in this case.

StrikerRUS:
@guolinke 
> there are several issues regard to this. Should I add it to the FAQ?

Yeah, I think it's definitely worth to mention that in the FAQ!

[...]

PR #2464, opened 2019-09-27
check the shape for mat, csr and csc in prediction

guolinke:
partially improve #812


StrikerRUS:
Close-reopen for CI.

src/c_api.cpp
```diff
@@ -249,17 +249,19 @@ class Booster {
     boosting_->RollbackOneIter();
   }
 
-  void PredictSingleRow(int num_iteration, int predict_type,
+  void PredictSingleRow(int num_iteration, int predict_type, int ncol,
                std::function<std::vector<std::pair<int, double>>(int row_idx)> get_row_fun,
                const Config& config,
                double* out_result, int64_t* out_len) {
+    if (ncol != boosting_->MaxFeatureIdx() + 1) {
+      Log::Fatal("The number of feature in data (%d) is not same as in training (%d).", ncol, boosting_->MaxFeatureIdx() + 1);
```
StrikerRUS: Maybe better
```
"The number of features in data (%d) is not the same as it was in training data (%d)."
```
?

src/c_api.cpp
```diff
@@ -268,10 +270,13 @@ class Booster {
   }
 
 
-  void Predict(int num_iteration, int predict_type, int nrow,
+  void Predict(int num_iteration, int predict_type, int nrow, int ncol,
                std::function<std::vector<std::pair<int, double>>(int row_idx)> get_row_fun,
                const Config& config,
                double* out_result, int64_t* out_len) {
+    if (ncol != boosting_->MaxFeatureIdx() + 1) {
+      Log::Fatal("The number of feature in data (%d) is not same as in training (%d).", ncol, boosting_->MaxFeatureIdx() + 1);
```
StrikerRUS: The same as above.

src/c_api.cpp
```diff
@@ -1315,7 +1320,7 @@ int LGBM_BoosterPredictForCSR(BoosterHandle handle,
   Booster* ref_booster = reinterpret_cast<Booster*>(handle);
   auto get_row_fun = RowFunctionFromCSR(indptr, indptr_type, indices, data, data_type, nindptr, nelem);
   int nrow = static_cast<int>(nindptr - 1);
-  ref_booster->Predict(num_iteration, predict_type, nrow, get_row_fun,
+  ref_booster->Predict(num_iteration, predict_type, nrow, static_cast<int32_t>(num_col), get_row_fun,
```
StrikerRUS: Is it safe? Maybe simply change the original type, as it was not used before?

src/c_api.cpp
```diff
@@ -1343,7 +1348,7 @@ int LGBM_BoosterPredictForCSRSingleRow(BoosterHandle handle,
   }
   Booster* ref_booster = reinterpret_cast<Booster*>(handle);
   auto get_row_fun = RowFunctionFromCSR(indptr, indptr_type, indices, data, data_type, nindptr, nelem);
-  ref_booster->PredictSingleRow(num_iteration, predict_type, get_row_fun, config, out_result, out_len);
+  ref_booster->PredictSingleRow(num_iteration, predict_type, static_cast<int32_t>(num_col), get_row_fun, config, out_result, out_len);
```
StrikerRUS: The same as above.

src/c_api.cpp
```diff
@@ -268,10 +270,13 @@ class Booster {
   }
 
 
-  void Predict(int num_iteration, int predict_type, int nrow,
+  void Predict(int num_iteration, int predict_type, int nrow, int ncol,
                std::function<std::vector<std::pair<int, double>>(int row_idx)> get_row_fun,
                const Config& config,
                double* out_result, int64_t* out_len) {
+    if (ncol != boosting_->MaxFeatureIdx() + 1) {
```
StrikerRUS: Doesn't it break the part about `0`?
https://github.com/microsoft/LightGBM/blob/1f88721a7982a804731b1dc108ee40625405639c/include/LightGBM/c_api.h#L686

StrikerRUS:
Please add a simple test from the original issue:
```
import numpy as np
import lightgbm as lgb

x_data = np.random.rand(100, 10)
x_bad_data = np.random.rand(100, 11)
y_data =  np.random.rand(100) > .5
self.assertNotEqual(x_data.shape[-1], x_bad_data.shape[-1])
train_dataset = lgb.Dataset(x_data, y_data)
bst = lgb.train({'objective': 'binary'}, train_dataset)
with np.testing.assert_raises_regex(lgb.basic.LightGBMError,
                                    'The number of features in data*'):
    bst.predict(x_bad_data)
```

src/c_api.cpp
guolinke: just realize we actually didn't guss the num_col when it is zero. 
I think it is better to pass non-zero values.

guolinke:
There are more changes than I expected. So I think we should test all cases, including the the mat, libsvm file and CSR format. 
@StrikerRUS could you help for the test cases?

StrikerRUS:
@guolinke 
> could you help for the test cases?

Sure!

What do you think about changing the original type to avoid casting? https://github.com/microsoft/LightGBM/pull/2464#discussion_r329257560

Does this PR already include [that check](https://github.com/microsoft/LightGBM/issues/2457#issuecomment-535911772)?

src/c_api.cpp
guolinke: change the original will cause lots of API changes. 
I think we actually don't support too many columns.
I will add a check for the num_col in CSR format, to ensure its value is smaller than int32_max.

guolinke:
> Does this PR already include that check?
Yeah, as the zero-based and one-based libsvm format have the different number of columns.

StrikerRUS:
@guolinke I added tests in the latest commit.

As we do not support a case when data for prediction is a list of arrays, I tried to check it for validation. And it seems that validation shape is not covered by this PR, right?


```
bad_valid_data = train_data.create_valid(bad_X_test, label=y_test)
bst.add_valid(bad_valid_data, "valid_bad")
bst.eval_valid()
```

guolinke:
yeah, validation is not converted for now.

guolinke:
due the need of pr #2485, I will merge this

PR #2489, opened 2019-10-04
[python] handle params aliases centralized

StrikerRUS:
Inspired by https://github.com/microsoft/LightGBM/pull/2443#discussion_r329402346.

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -170,6 +170,57 @@ class LightGBMError(Exception):
     pass
 
 
+class _ConfigAliases(object):
+    aliases = {"boosting": {"boosting",
+                            "boosting_type",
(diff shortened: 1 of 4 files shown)


python-package/lightgbm/basic.py
```diff
@@ -196,6 +196,48 @@ class LightGBMError(Exception):
                      "feature_penalty": C_API_DTYPE_FLOAT64,
                      "monotone_constraints": C_API_DTYPE_INT8}
 
+CONFIG_ALIASES = {"boosting": {"boosting",
```
chivee: Can we please turn this into a class? class params

StrikerRUS: Sorry, not sure that I understood you correctly. Please take a look.

PR #2495, opened 2019-10-06
[ci] reduced number of commits cloned to CI environments

jameslamb:
Minor change. We could slightly speed up CI by only pulling the most recent commit when cloning this project's source code during CI builds.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -10,7 +10,7 @@ environment:
     - COMPILER: MSVC
     - COMPILER: MINGW
 
-clone_depth: 1
+clone_depth: 5
 
 install:
   - git submodule update --init --recursive  # get `compute` folder
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -10,7 +10,7 @@ environment:
     - COMPILER: MSVC
     - COMPILER: MINGW
 
-clone_depth: 50
+clone_depth: 1
 
 install:
   - git submodule update --init --recursive  # get `compute` folder
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: cpp
 
 git:
   submodules: true
+  depth: 1
 
 os:
   - linux
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: cpp
 
 git:
   submodules: true
-  depth: 1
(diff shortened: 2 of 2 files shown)


jameslamb:
> Are we OK with the following limitation?
> 
> > Please note that if you use a depth of 1 and have a queue of jobs, Travis CI won’t build commits that are in the queue when you push a new commit.
> 
> LGTM!

Yes I think this is ok. I think in the case where we push a change, then 30 seconds later push another, it's actually _desirable_ to have Travis disregard the earlier one and only test the head commit (since there are some limits on total number of concurrent jobs).

StrikerRUS:
OK! But then we need to be more accurate with merging PRs into `master`. I personally don't want to see plenty of red crosses in commits' history in case of merging a lot of PRs at once.

jameslamb:
> OK! But then we need to be more accurate with merging PRs into `master`. I personally don't want to see plenty of red crosses in commits' history in case of merging a lot of PRs at once.

I don't want to introduce any manual steps or "you just have to know" conventions. I wasn't aware that you looked at build statuses in the commit _history_.

What if we set it to, say, 10? That would still cut the amount of commits we're pulling (currently 50 on Appveyor) while still providing good protection in the case of a few PRs merged in quick succession.

I definitely do not want to make anyone's experience worse in exchange for cutting a few seconds off our builds.

StrikerRUS:
> I wasn't aware that you looked at build statuses in the commit history.

Yep, I do... 🙂 It helps me to spot some breaking changes in third party software, which require any actions from us.

Moreover, we **have** to keep Azure Pipelines build statuses of `master` branch always-green because our nightly builds are produced by that CI service and in case of any failure artifacts publishing jobs are canceled.

> Also you can download artifacts of the latest successful build in master branch
https://github.com/microsoft/LightGBM/blob/master/docs/Installation-Guide.rst#installation-guide


> What if we set it to, say, 10? 

Maybe 5? 😃  Seems that simultaneous merging of 5+ PRs is something "rare".

jameslamb:
Thanks @StrikerRUS . I just updated these to pull 5 commits (I learned some new things about out Azure pipelines from [this comment](https://github.com/microsoft/LightGBM/pull/2495#issuecomment-538974625), much appreciated!)

PR #2496, opened 2019-10-06
fixed miscellaneous typos in documentation

jameslamb:
This is what happens when I'm on a long flight with no internet 😂 

This PR fixes a variety of minor typos and grammar issues in the code base. Please let me know if there are any other documentation files I need to regenerate as a result of changing some comments on the C++ side.

### DIFF
--- a/include/LightGBM/meta.h
+++ b/include/LightGBM/meta.h
@@ -17,10 +17,10 @@ namespace LightGBM {
 /*! \brief Type of data size, it is better to use signed type*/
 typedef int32_t data_size_t;
 
-// Enable following marco to use double for score_t
+// Enable following macro to use double for score_t
 // #define SCORE_T_USE_DOUBLE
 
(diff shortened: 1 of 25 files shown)


include/LightGBM/meta.h
```diff
@@ -17,10 +17,10 @@ namespace LightGBM {
 /*! \brief Type of data size, it is better to use signed type*/
 typedef int32_t data_size_t;
 
-// Enable following marco to use double for score_t
+// Enable following macro to use double for score_t
 // #define SCORE_T_USE_DOUBLE
 
-// Enable following marco to use double for label_t
+// Enable following Macao to use double for label_t
```
StrikerRUS: 🤣 

Is it a spoiler with your flight destination?

jameslamb: haha no! Just me clearly mistyping my selection in `aspell`. Thanks for catching 😆

Issue #2468, opened 2019-09-29
[R-package] training continuation unit test is broken


jameslamb:
One unit test in the R package is currently broken. Steps to reproduce on Mac

```r
export CXX=/usr/local/bin/g++-8 CC=/usr/local/bin/gcc-8
Rscript build_r.R
cd R-package/tests
Rscript testthat.R
```

This results in the following error at the ends of the logs

```
[LightGBM] [Info] Saving data to binary file /var/folders/xq/wktq4zdx4jd3qdpk34d28m940000gn/T//RtmpiY1DzV/lgb.Dataset_15553eef826
══ testthat results  ═══════════════════════════════════════════════════════════
[ OK: 36 | SKIPPED: 0 | WARNINGS: 0 | FAILED: 1 ]
1. Error: training continuation works (@test_basic.R#71)
```

I'm uncertain when the issue was first introduced (we do not run R tests in CI yet, #2335), but this behavior was observed at least as of https://github.com/microsoft/LightGBM/commit/cc7a1e27398d86e9befdf6eb08141359882f3d49.

jameslamb:
Looked into this some more tonight. I still haven't gotten to the root cause.

This was failing with basically an out-of-index error. When you do training continuation, it seems the second

```r
# check this out:
bst1$record_evals[[data_name]][[eval_name]]$eval
bst2$record_evals[[data_name]][[eval_name]]$eval
```

things you'll notice:

* best iteration from `bst1` is iteration 3
* if you compare the errors, it looks like `bst2` starts from iteration 3 (notice the 4th iteration in `bst1` and first in `bst2` have identical error)
* `bst2$current_iter()` is 8, also consistent with starting from the best iteration
* I think the issue is here in `lgb.train()`:

    ```r
    if (is.character(init_model)) {
        predictor <- Predictor$new(init_model)
    ```

The `Predictor` created from the initial model should be the same as the booster state after 5 iterations, I think, but it is not.

So basically you can get into a situation where the Booster thinks it has seen 8 iterations, but only 5 iterations of results are stored.

* `length(bst$record_evals[["train"]][["binary_logloss"]]$eval)` is 10
* `length(bst1$record_evals[["train"]][["binary_logloss"]]$eval)` is 5
* `length(bst2$record_evals[["train"]][["binary_logloss"]]$eval)` is 5

PR #2498, opened 2019-10-07
[R-package] Added unit tests

jameslamb:
On a flight this weekend, I spent some time learning the internals of `covr` and trying to figure out why `covr::package_coverage()` doesn't work on the `LightGBM` R package.

While I still haven't figured that out, I did figure out how to measure the unit test coverage of the R package with a modified version of `covr::package_coverage()`. See https://github.com/jameslamb/lightgbm-r-coverage.

Currently on `master`, the R package's test coverage is `56.37%`. The tests in this PR takes it to `65.48%`.

In this PR, I also propose formally skipping the broken test referenced in #2468 with `testthat::skip()`. I'm making some progress on that issue but in the interim we might as well have our CI work and not have to "just remember" that that one test has been broken for some time. This skip is necessary to compute test coverage.


build_r.R
```diff
@@ -75,7 +75,7 @@ version <- gsub(
 )
 tarball <- file.path(getwd(), sprintf("lightgbm_%s.tar.gz", version))
 
-cmd <- sprintf("R CMD INSTALL %s --no-multiarch", tarball)
+cmd <- sprintf("R CMD INSTALL %s --no-multiarch --install-tests --with-keep.source --with-keep.parse.data", tarball)
```
StrikerRUS: @jameslamb My R knowledge is quite poor, so I removed a review request from me. Sorry.

I'm just worrying about `--install-tests`. What is it? Will it increase the installation time significantly, which may cause `RTD` builds to fail due to time limits?

jameslamb: good point @StrikerRUS . I actually think I can refactor this to make that unnecessary. Those flags do increase installation time (although it should be very small), but actually we don't need them for this at all. They're passed at build time by `covr::package_coverage()` but since I'm using a super-LightGBM-specific mutation of `covr::package_coverage()`, I can just manually handle copying the test files.

jameslamb: done!

jameslamb:
ping @Laurae2 for a review

build_r.R
StrikerRUS: Thanks!

jameslamb:
I just rebased this to latest `master` to catch up with changes over the last two weeks. It is ready for review.

R-package/tests/testthat/test_basic.R
```diff
@@ -51,6 +51,7 @@ test_that("use of multiple eval metrics works", {
 
 
 test_that("training continuation works", {
+  testthat::skip("This test is currently broken. See #2468 for details.")
```
Laurae2: Just to make it a bit more obvious (for any user), change to `This test is currently broken. See issue #2468 for details.`

But if it's only for us then there's no need to change.

jameslamb: oh sure! I can add that, good idea

PR #2528, opened 2019-10-27
[docs] update MS-MPI link

StrikerRUS:
10.0 -> 10.1.1


jameslamb:
The new link worked for me in my browser, but seeing this in Travis

image

I tried re-running that stage, let's see if it was transient or if link-checker is actually getting a 404

PR #2530, opened 2019-10-27
[R-package][ci] added CI stage  for R package (fixes #2335, fixes #2569)

jameslamb:
I'm still committed to getting R CI running in GitHub Actions (#2353 ) but I think there is also an immediate benefit to having the R package built + unit tests run _at all_, and tacking that on to Travis is the easiest path.

In this PR, I propose adding an `r-pkg` stage to our Travis configuration to at least test the R package on Ubuntu and macOS. That will catch issues like #2524 , where a change on the C++ side breaks the R package. We can handle tests on Windows, tests with different versions of R, running on GitHub Actions, etc. in later PRs.

Opening as a draft PR since I may need to build a few times to get feedback from Travis. I'll change to regular PR and add reviewers when I think it's ready and worthy of your time, but comments are welcome any time!

PR #2533, opened 2019-10-29
[R-package][ci] Added more R linters (fixes #2477)

### LINKED ISSUE #2477: [R-package] Add additional linters to CI
Linting for the R code in this project was introduced in #2437 . As of that PR, the following linters from the `lintr` project were not available from the distribution of `lintr` on CRAN.

```r
github_only_linters <- list(
    "equals_na" = lintr::equals_na_linter
    , "function_left" = lintr::function_left_parentheses_linter
    , "implicit_integers" = lintr::implicit_integer_linter
    , "paren_brace_linter" = lintr::paren_brace_linter
    , "semiconlon" = lintr::semicolon_terminator_linter
    , "seq" = lintr::seq_linter
    , "todo_comments" = lintr::todo_comment_linter
    , "true_false" = lintr::T_and_F_symbol_linter
)
```

To close this issue, add those linters to `.ci/lint_r_code.R` once `lintr` does a new release (> 1.0.3) to CRAN.

jameslamb:
`lintr` released 2.0.0 to CRAN (and `r-lintr` to `conda-forge`).

In this PR, I propose adding additional linters on the R code in the project, to enforce particular style. See #2477 for background.

Start with the diff to `.ci/lint_r_code.R` to see the new linters being added.

One change you will see a LOT of is something like replacing `i + 1` with `i +1L` to satisfy the `implicit_integer_linter`. This one caught a lot of cases where we were relying on R sometimes treating `1` as double precision and sometimes as integer. This is basically always fine in R, but it can mean you're using more memory than necessary. So think of the addition of the `L` as a tiny tiny memory savor and, more importantly, documentation for people reading the code.


.ci/test.sh
```diff
@@ -54,7 +54,7 @@ if [[ $TASK == "lint" ]]; then
     conda install -q -y -n $CONDA_ENV \
         pycodestyle \
         pydocstyle \
-        r-lintr
+        r-lintr==2.0.0
```
StrikerRUS: @jameslamb 

Maybe `>=2.0.0` is better to free us from bumping the version after each new release?

> lintr released 2.0.0 to CRAN (and r-lintr to conda-forge).

```
Preparing transaction: ...working... done
Verifying transaction: ...working... done
Executing transaction: ...working... done
Collecting package metadata (current_repodata.json): ...working... done
Solving environment: ...working... failed with initial frozen solve. Retrying with flexible solve.
Collecting package metadata (repodata.json): ...working... done
Solving environment: ...working... failed with initial frozen solve. Retrying with flexible solve.
PackagesNotFoundError: The following packages are not available from current channels:
  - r-lintr==2.0.0
Current channels:
  - https://repo.anaconda.com/pkgs/main/linux-64
  - https://repo.anaconda.com/pkgs/main/noarch
  - https://repo.anaconda.com/pkgs/r/linux-64
  - https://repo.anaconda.com/pkgs/r/noarch
To search for alternate channels that may provide the conda package you're
looking for, navigate to
    https://anaconda.org
and use the search bar at the top of the page.
```

The you should either wait for release in default (`r`) channel, or install it from `conda-forge` channel:
```
conda install -q -y -n $CONDA_ENV -c conda-forge r-lintr
```

jameslamb: @StrikerRUS ah good catch on the channels, my mistake

I think `==2.0.0` is actually better for this case. Since this is just a linter, I think it is desirable to have it keep performing the same and then at some point in the future (6 months or 1 year) update if they have a new release. Otherwise we risk some PR unrelated to the R code having its CI break because a new `lintr` release is found. So in other words...I think pegging to an exact version in this case is _less_ maintenance burden, not more.

StrikerRUS: > Otherwise we risk some PR unrelated to the R code having its CI break because a new lintr release is found.

Can you please be more precise? In case of introducing of new linters which LightGBM code doesn't follow, then yes. But we have the exact list of checking linters, so I guess it's not possible. New versions will (potentially) bring bug fixes and speedups, which is good to have ASAP after release.

jameslamb: I was able to fix this (I think!) by re-installing the `stringi` package. This was suggested [deep in the arch linux forum archive](https://bbs.archlinux.org/viewtopic.php?id=235861)

jameslamb:
Just rebased to most recent `master`. I think this is ready for review

.ci/test.sh
```diff
@@ -51,10 +51,13 @@ if [[ $TRAVIS == "true" ]] && [[ $TASK == "check-docs" ]]; then
 fi
 
 if [[ $TASK == "lint" ]]; then
+    sudo apt-get update
```
StrikerRUS: Why do we need this?

.ci/test.sh
```diff
@@ -51,10 +51,13 @@ if [[ $TRAVIS == "true" ]] && [[ $TASK == "check-docs" ]]; then
...
     conda install -q -y -n $CONDA_ENV \
         pycodestyle \
-        pydocstyle \
-        r-lintr
+        pydocstyle
+    conda install -q -y -n $CONDA_ENV -c conda-forge \
+        r-lintr==2.0.0
```
StrikerRUS: Please comment https://github.com/microsoft/LightGBM/pull/2533/files#r340198019.

.ci/test.sh
```diff
@@ -51,10 +51,13 @@ if [[ $TRAVIS == "true" ]] && [[ $TASK == "check-docs" ]]; then
...
+    Rscript -e "install.packages('stringi', repos = 'http://cran.rstudio.com')"
```
StrikerRUS:
```suggestion
    Rscript -e "install.packages('stringi', repos = 'https://cran.rstudio.com')"
```

StrikerRUS: It takes soo much time! Is it the only available fix? Maybe it will be faster to update `lintr` from CRAN?

.ci/test.sh
jameslamb: Sorry, resolved this too fast. I didn't realize this was also the `==` vs . `>=` thread.

I am ok with using `>=` so that we get bugfixes and performance enhancements ASAP. I'll make that change.

.ci/test.sh
StrikerRUS: Oh no! I think we should stick to conda's environment, since it's significantly faster than ordinary one and suits to our purpose (just linting) very well.
Sorry for my ambiguous comment above! I just meant to try `Rscript -e "install.packages('lintr', repos = 'http://cran.rstudio.com')"` to see whether it's faster than reinstalling `stringi` or not.

.ci/test.sh
jameslamb: sorry, resolved that thread too early! I just commented there. I think installing from CRAN is going to be a better option (testing that now), so it will effectively be `>=2.0.0` anyway.

.ci/test.sh
jameslamb: ah I see, ok. Yes I think I misinterpreted what you were saying. I'll change it back and try.

jameslamb: Looks like linting task took [almost 7 minutes to run](https://travis-ci.org/microsoft/LightGBM/jobs/607877757?utm_medium=notification&utm_source=github_status) still with this approach (using alternative R setup)

image

.ci/test.sh
```diff
@@ -53,16 +53,38 @@ fi
...
+    # # https://cran.r-project.org/bin/linux/ubuntu/#installation
+    # sudo apt-get update
+    # sudo add-apt-repository \
+    #     "deb https://cloud.r-project.org/bin/linux/ubuntu bionic-cran35/"
+    # sudo apt-get update
+    # sudo apt-get install \
+    #     -y \
+    #         r-base \
+    #         r-base-dev \
+    #         || exit -1
```
jameslamb: I am testing `lintr` install from CRAN. If that works, will remove the commented code here

.ci/test.sh
jameslamb: creating an R environment from `conda` and then `install.packages('lintr')` from CRAN, the `lint` task ran for 8m38s and then failed with

```
ERROR: dependency ‘curl’ is not available for package ‘httr’
```

I could add an `install.packages('curl')` but that would make it run even longer than 8 minutes.

I decided to summarize all the attempts here (and add a few new ideas):

1. ([-failure-](https://travis-ci.org/microsoft/LightGBM/builds/605269042?utm_medium=notification&utm_source=github_status)) R env from conda, `lintr` from conda-forge
2. ([success](https://travis-ci.org/microsoft/LightGBM/jobs/606612630?utm_medium=notification&utm_source=github_status)) R env from conda, `lintr` from conda-forge, `install.packages('stringi')`: 6m9s
3. ([success](https://travis-ci.org/microsoft/LightGBM/builds/607877752?utm_medium=notification&utm_source=github_status)) R env from CRAN like #2530, `install.packages('lintr')`: 6m48s
4. ([-failure-](https://travis-ci.org/microsoft/LightGBM/builds/607939450?utm_medium=notification&utm_source=github_status)) R env from conda, `install.packages('lintr')`: 8+ minutes
5. ([-failure-](https://travis-ci.org/microsoft/LightGBM/jobs/607962730?utm_medium=notification&utm_source=github_status)) R env from conda, `lintr` from conda-forge, re-installing `stringi` from default conda channels: failed
6. ([success](https://travis-ci.org/microsoft/LightGBM/jobs/607963803?utm_medium=notification&utm_source=github_status)) R env from conda, installing `stringi` from conda first, then `lintr` from conda-forge: 1m57s

So I think we should go with installing `stringi` explicitly, THEN `lintr`! This works and runs in under 2 minutes 🎉🎉 🎉

.ci/test.sh
jameslamb: it worked!

.ci/test.sh
jameslamb: see https://github.com/microsoft/LightGBM/pull/2533#discussion_r342883220, I have this working now and am using `>=`

.ci/test.sh
jameslamb: ok now I think this can actually be resolved. I'm using . `>=`

.ci/test.sh
```diff
@@ -53,16 +53,24 @@ fi
...
     conda install -q -y -n $CONDA_ENV \
         pycodestyle \
-        pydocstyle \
-        r-lintr
+        pydocstyle
     pip install --user cpplint
     echo "Linting Python code"
     pycodestyle --ignore=E501,W503 --exclude=./compute,./.nuget . || exit -1
     pydocstyle --convention=numpy --add-ignore=D105 --match-dir="^(?!^compute|test|example).*" --match="(?!^test_|setup).*\.py" . || exit -1
+
```
StrikerRUS: Please remove this empty line for the consistency with the rest code in this file.

.ci/test.sh
```diff
@@ -53,16 +53,24 @@ fi
...
+    # stringi and r-lintr
```
StrikerRUS: I don't think that this comment adds any value. Instead, we should have a comment about **why** `stringi` is needed here and that we can drop it in the future after some upstream fixes.

.ci/test.sh
```diff
@@ -53,16 +53,24 @@ fi
...
     echo "Linting R code"
```
StrikerRUS: Please move this line right before `Rscript ...` call where the linting is actually happening.

.ci/test.sh
```diff
@@ -53,16 +53,24 @@ fi
...
+    conda install -y -n $CONDA_ENV \
+        r-stringi
+    conda install -y -n $CONDA_ENV \
+        -c conda-forge \
+            r-lintr>=2.0
     Rscript ${BUILD_DIRECTORY}/.ci/lint_r_code.R ${BUILD_DIRECTORY} || exit -1
+
```
StrikerRUS: Ditto for empty line.

.ci/test.sh
```diff
@@ -53,16 +53,24 @@ fi
...
     echo "Linting C++ code"
     cpplint --filter=-build/c++11,-build/include_subdir,-build/header_guard,-whitespace/line_length --recursive ./src ./include || exit 0
+
```
StrikerRUS: Ditto for empty line.

.ci/test.sh
```diff
@@ -53,16 +53,24 @@ fi
 if [[ $TASK == "lint" ]]; then
```
StrikerRUS: Have you tested it at Azure Pipelines? If it doesn't work there, then we should have a similar check to what we have for `check-docs` job.

.ci/test.sh
jameslamb: can do!

.ci/test.sh
jameslamb: ah ok sure, can do

.ci/test.sh
jameslamb: ah you're right. I haven't tested it on Azure. Since it's just linting, I don't think we need to run it more than once

I can update the condition

.ci/test.sh
jameslamb: removed

.ci/test.sh
jameslamb: removed

.ci/test.sh
jameslamb: moved

.ci/test.sh
jameslamb: updated!

.ci/test.sh
jameslamb: Yep totally agree! Just changed it in the most recent commit

jameslamb:
ping @Laurae2 for review on changes to the R code

.ci/test.sh
```diff
@@ -50,15 +50,21 @@ if [[ $TRAVIS == "true" ]] && [[ $TASK == "check-docs" ]]; then
...
-        pydocstyle \
-        r-lintr
+        pydocstyle
     pip install --user cpplint
     echo "Linting Python code"
     pycodestyle --ignore=E501,W503 --exclude=./compute,./.nuget . || exit -1
     pydocstyle --convention=numpy --add-ignore=D105 --match-dir="^(?!^compute|test|example).*" --match="(?!^test_|setup).*\.py" . || exit -1
+    # stringi needs to be installed separate from r-lintr
+    # to avoid issues like 'unable to load shared object stringi.so'
+    conda install -y -n $CONDA_ENV \
```
StrikerRUS:
```suggestion
    conda install -q -y -n $CONDA_ENV \
```

.ci/test.sh
```diff
@@ -50,15 +50,21 @@ if [[ $TRAVIS == "true" ]] && [[ $TASK == "check-docs" ]]; then
...
+        r-stringi
+    conda install -y -n $CONDA_ENV \
```
StrikerRUS:
```suggestion
    conda install -q -y -n $CONDA_ENV \
```

R-package/R/lgb.Dataset.R
```diff
@@ -924,7 +924,7 @@ dimnames.lgb.Dataset <- function(x) {
 #' train <- agaricus.train
 #' dtrain <- lgb.Dataset(train$data, label = train$label)
 #'
-#' dsub <- lightgbm::slice(dtrain, 1:42)
+#' dsub <- lightgbm::slice(dtrain, 1L:42L)
```
Laurae2: `seq_len(42L)` for safety

R-package/R/lgb.Dataset.R
```diff
@@ -1059,7 +1059,7 @@ setinfo.lgb.Dataset <- function(dataset, name, info, ...) {
 #' dtrain <- lgb.Dataset(train$data, label = train$label)
 #' lgb.Dataset.save(dtrain, "lgb.Dataset.data")
 #' dtrain <- lgb.Dataset("lgb.Dataset.data")
-#' lgb.Dataset.set.categorical(dtrain, 1:2)
+#' lgb.Dataset.set.categorical(dtrain, 1L:2L)
```
Laurae2: `seq_len(2L)` for safety

R-package/R/lgb.interprete.R
```diff
@@ -29,14 +29,14 @@
...
+#'     , num_leaves = 63L
+#'     , max_depth = -1L
+#'     , min_data_in_leaf = 1L
+#'     , min_sum_hessian_in_leaf = 1.0
 #' )
-#' model <- lgb.train(params, dtrain, 10)
+#' model <- lgb.train(params, dtrain, 10L)
 #'
-#' tree_interpretation <- lgb.interprete(model, test$data, 1:5)
+#' tree_interpretation <- lgb.interprete(model, test$data, 1L:5L)
```
Laurae2: `seq_len(5L)` for safety

R-package/demo/leaf_stability.R
```diff
@@ -112,28 +112,29 @@ new_data2 <- data.frame(
...
      )
 )
-new_data2$Z <- -1 * (agaricus.test$label * log(new_data2$Y) + (1 - agaricus.test$label) * log(1 - new_data2$Y))
+new_data2$Z <- -1.0 * (agaricus.test$label * log(new_data2$Y) + (1L - agaricus.test$label) * log(1L - new_data2$Y))
 new_data2$binned <- .bincode(
     x = new_data2$X
     , breaks = quantile(
         x = new_data2$X
-        , probs = (1:9) / 10
+        , probs = (1L:9L) / 10.0
```
Laurae2: `seq_len(9L)` for safety

R-package/demo/leaf_stability.R
```diff
@@ -180,20 +181,20 @@ new_data3 <- data.frame(
...
     )
 )
-new_data3$Z <- -1 * (agaricus.test$label * log(new_data3$Y) + (1 - agaricus.test$label) * log(1 - new_data3$Y))
+new_data3$Z <- -1.0 * (agaricus.test$label * log(new_data3$Y) + (1L - agaricus.test$label) * log(1L - new_data3$Y))
 new_data3$binned <- .bincode(
     x = new_data3$X
     , breaks = quantile(
         x = new_data3$X
-        , probs = (1:9) / 10
+        , probs = (1L:9L) / 10.0
```
Laurae2: `seq_len(9L)` for safety

R-package/R/lgb.interprete.R
Laurae2: (ignore, misread the function)

.ci/test.sh
```diff
@@ -50,15 +50,21 @@ if [[ $TRAVIS == "true" ]] && [[ $TASK == "check-docs" ]]; then
...
-        r-lintr
+        pydocstyle
     pip install --user cpplint
     echo "Linting Python code"
     pycodestyle --ignore=E501,W503 --exclude=./compute,./.nuget . || exit -1
     pydocstyle --convention=numpy --add-ignore=D105 --match-dir="^(?!^compute|test|example).*" --match="(?!^test_|setup).*\.py" . || exit -1
+    # stringi needs to be installed separate from r-lintr
+    # to avoid issues like 'unable to load shared object stringi.so'
+    conda install -q -y -n $CONDA_ENV \
+        r-stringi
```
StrikerRUS: I think it could be 
```
conda install -q -y -n $CONDA_ENV \
    pycodestyle \
    pydocstyle \
    r-stringi  # stringi needs to be installed separate from r-lintr to avoid issues like 'unable to load shared object stringi.so'
conda install -q -y -n $CONDA_ENV \
    -c conda-forge \
        r-lintr>=2.0
```
to avoid multiple `conda install ...` commands and keep them together in one place for a better readability.

R-package/demo/leaf_stability.R
jameslamb: could you explain how `seq_len(9L)` is safer than `1L:9L`?

jameslamb: I'm happy to change these in this PR for consistency with the other code where we initialize vectors like this, but I don't think there is a difference in safeness since `1L` and `9L` are both integer literals.

.ci/test.sh
jameslamb: sure I'm ok with that

jameslamb: done

R-package/demo/leaf_stability.R
jameslamb: changed these

jameslamb:
just rebased this to most recent `master`

R-package/demo/leaf_stability.R
Laurae2: @jameslamb It's just in case if those are modified in the future, the probability vector would throw an error if its value is negative (quantile already throws another kind of error message with negative probs).

We also usually don't expect quantile (as users) to take the 0 percentile (minimum), it might happen though (and that would have to be explicit).

PR #2539, opened 2019-11-04
renamed variables from bias to offset

SiNZeRo:
as discussed with @guolinke


thvasilo:
Hello @SiNZeRo , I was recently looking into the histogram code and trying to understand why the name bias was used in this context. I was mostly interested because it seems to affect the measured size of histograms in memory.

By this switch to offset, which I guess was made to better describe the purpose of the variable, is because it indicates whether the histogram uses a "default" bin or not? Not sure how the default bin is used, but I'm going to guess it has to do something with zero-valued histograms?

Any chance you could explain a bit the function/purpose of the `offset` (nee `bias_`) variable?

guolinke:
hi @thvasilo ,
The offset is used to optimize sparse features. 
For a sparse feature, you can only pass the non-zero(or not default bin) feature values, to construct the histogram, then restore the default-bin by the subtraction of the sum of all grad/hess and the sum of remaining bins.

Issue #2344, opened 2019-08-20
Add AUC mu metric for multiclass training


MotoRZR:
There was a paper written on a new AUC metric for multiclass problems.  

http://proceedings.mlr.press/v97/kleiman19a/kleiman19a.pdf

The authors have provided the code in the repository below so this should be straightforward to incorporate as an inbuilt metric.

https://github.com/kleimanr/auc_mu/blob/master/auc_mu.py

StrikerRUS:
For the reference: catboost has already implemented AUC Mu: https://github.com/catboost/catboost/commit/298ca12c7c7c8bda54ef1291b3fb1ea0888e6539.

StrikerRUS:
Closed in favor of being in #2302. We decided to keep all feature requests in one place.

Welcome to contribute this feature! Please re-open this issue (or post a comment if you are not a topic starter) if you are actively working on implementing this feature.

btrotta:
PR is open for this.

StrikerRUS:
Closed via #2567.

PR #2567, opened 2019-11-14
auc-mu

btrotta:
Add auc-mu multi-class metric as requested in #2344


include/LightGBM/config.h
```diff
@@ -806,6 +807,12 @@ struct Config {
   // desc = when ``multi_error_top_k=1`` this is equivalent to the usual multi-error metric
   int multi_error_top_k = 1;
 
+  // desc = file containing matrix of weights for calculating auc-mu multi-class metric
```
StrikerRUS:
```suggestion
  // desc = file containing matrix of weights for calculating AUC-mu multi-class metric
```

src/io/config.cpp
```diff
@@ -153,6 +155,63 @@ void GetTreeLearnerType(const std::unordered_map<std::string, std::string>& para
...
+        for (int j = 0; j < num_class; ++j) {
+          if ((oneline_ind >= oneline_weights_nonzero.size()) || (oneline_weights_nonzero[oneline_ind].first > j)) {
+            oneline_weights.emplace_back(0);
+          } else {
+            oneline_weights.emplace_back(oneline_weights_nonzero[oneline_ind].second);
+            oneline_ind++;
+          }
+        }
+        if (static_cast<int>(oneline_weights.size()) != num_class) {
+          Log::Fatal("Auc-mu weights matrix must have %d columns but found row with %d entries.",
```
StrikerRUS:
```suggestion
          Log::Fatal("AUC-mu weights matrix must have %d columns but found row with %d entries.",
```

src/io/config.cpp
```diff
@@ -153,6 +155,63 @@ void GetTreeLearnerType(const std::unordered_map<std::string, std::string>& para
...
+                     num_class, oneline_weights.size());
+        }
+        std::vector<double> curr_line_weights;
+        for (size_t j = 0; j < num_class; ++j) {
+          curr_line_weights.push_back(oneline_weights[j]);
+          if (oneline_weights[j] < 0) {
+            Log::Fatal("Auc-mu weights matrix must contain only non-negative values. Found negative value at position [%d, %d]",
```
StrikerRUS:
```suggestion
            Log::Fatal("AUC-mu weights matrix must contain only non-negative values. Found negative value at position [%d, %d]",
```

src/io/config.cpp
```diff
@@ -153,6 +155,63 @@ void GetTreeLearnerType(const std::unordered_map<std::string, std::string>& para
...
+                       i, j);
+          }
+          if ((i == j) && (oneline_weights[j] > kEpsilon)) {
+            Log::Warning("Diagonal of auc-mu weights matrix must be zero. Overriding entry at position [%d, %d] with zero.",
```
StrikerRUS:
```suggestion
            Log::Warning("Diagonal of AUC-mu weights matrix must be zero. Overriding entry at position [%d, %d] with zero.",
```

src/io/config.cpp
```diff
@@ -153,6 +155,63 @@ void GetTreeLearnerType(const std::unordered_map<std::string, std::string>& para
...
+                         i, j);
+            curr_line_weights.push_back(0);
+          }
+        }
+      auc_mu_weights.push_back(curr_line_weights);
+      }
+    };
+  reader.ReadAllAndProcessParallel(process_fun);
+  if (auc_mu_weights.size() != num_class) {
+    Log::Fatal("Auc-mu matrix must have %d rows but found %d", num_class, auc_mu_weights.size());
```
StrikerRUS:
```suggestion
    Log::Fatal("AUC-mu matrix must have %d rows but found %d", num_class, auc_mu_weights.size());
```

tests/python_package_test/test_engine.py
```diff
@@ -428,6 +428,72 @@ def test_multi_class_error(self):
...
+        results_auc_mu = {}
+        lgb.train(params, lgb_X, num_boost_round=10, valid_sets=[lgb_X], evals_result=results_auc_mu)
+        lgb_X = lgb.Dataset(X, label=(y == 0).astype(int))
+        params = {'objective': 'binary',
+                  'metric': 'auc',
+                  'verbose': -1,
+                  'seed': 0}
+        results_auc = {}
+        lgb.train(params, lgb_X, num_boost_round=10, valid_sets=[lgb_X], evals_result=results_auc)
+        self.assertAlmostEqual(results_auc_mu['training']['auc_mu'][-1], results_auc['training']['auc'][-1])
```
StrikerRUS: Can we check the whole result array `results_auc_mu['training']['auc_mu']` instead of checking the only last element with the help of `np.testing.assert_allclose()`?

btrotta:
@StrikerRUS  thanks for your review, I have fixed those issues now.

examples/multiclass_classification/loss.weights
```diff
@@ -0,0 +1,5 @@
+0.00000	0.67822	0.66155	0.25288	0.24816
```
StrikerRUS: Is everything OK with this file? Seems that the first and second columns are merged.

btrotta: Sorry, I'm not sure what you mean about the columns being merged? When I look at the file in my editor it shows the whitespace characters are all tabs.

StrikerRUS: Ah, seems that it's a GitHub issue then - the first tab (which separates the first and second columns) is not visible at all:

image

src/io/config.cpp
```diff
@@ -153,6 +155,63 @@ void GetTreeLearnerType(const std::unordered_map<std::string, std::string>& para
   }
 }
 
+void Config::GetAucMuWeights(const std::unordered_map<std::string, std::string>& params) {
+  std::string filename;
+  if (Config::GetString(params, "auc_mu_weights_file", &filename)) {
+    // read weights from file
+    TextReader<size_t> reader(filename.c_str(), false);
+    Parser* parser = Parser::CreateParser(filename.c_str(), false, num_class, -1);
```
guolinke: will the `auc_mu_weight_file` be large? if not, maybe just use `ReadAllLines` and process them one-by-one...
BTW, are there any default values for the class weights when file is not provided? like ones array or random array?

src/io/config.cpp
```diff
@@ -153,6 +155,63 @@ void GetTreeLearnerType(const std::unordered_map<std::string, std::string>& para
...
+            curr_line_weights.push_back(0);
+          }
+        }
+      auc_mu_weights.push_back(curr_line_weights);
+      }
+    };
+  reader.ReadAllAndProcessParallel(process_fun);
+  if (auc_mu_weights.size() != static_cast<size_t>(num_class)) {
+    Log::Fatal("AUC-mu matrix must have %d rows but found %d", num_class, auc_mu_weights.size());
+  }
```
StrikerRUS: Bad indent?

src/metric/multiclass_metric.hpp
```diff
@@ -178,5 +178,137 @@ class MultiSoftmaxLoglossMetric: public MulticlassMetric<MultiSoftmaxLoglossMetr
...
+    ans = 2 * ans / (num_class_ * (num_class_ - 1));
+    return std::vector<double>(1, ans);
+  }
+
+private:
+  /*! \brief Number of data*/
+  data_size_t num_data_;
+  /*! \brief Pointer to label*/
+  const label_t* label_;
+  /*! \brief Name of test set*/
```
StrikerRUS: Name of this metric?

tests/python_package_test/test_engine.py
```diff
@@ -428,6 +428,72 @@ def test_multi_class_error(self):
...
+        y_new[y != 0] = 1
+        lgb_X = lgb.Dataset(X, label=y_new)
+        params = {'objective': 'multiclass',
+                  'metric': 'auc_mu',
+                  'verbose': -1,
+                  'num_classes': 2,
+                  'seed': 0}
+        results_auc_mu = {}
+        lgb.train(params, lgb_X, num_boost_round=10, valid_sets=[lgb_X], evals_result=results_auc_mu)
+        lgb_X = lgb.Dataset(X, label=(y == 0).astype(int))
```
StrikerRUS: I think `lgb_X` from L436 can be re-used for AUC calculation at L450.

tests/python_package_test/test_engine.py
```diff
@@ -428,6 +428,72 @@ def test_multi_class_error(self):
...
+                  'metric': 'auc_mu',
+                  'verbose': -1,
+                  'num_classes': 2,
+                  'min_data_in_leaf': 20,
+                  'seed': 0}
+        results_auc_mu = {}
+        lgb.train(params, lgb_X, num_boost_round=10, valid_sets=[lgb_X], evals_result=results_auc_mu)
+        self.assertAlmostEqual(results_auc_mu['training']['auc_mu'][-1], 0.5)
+        # should give 1 when accuracy = 1
+        X, y = load_digits(10, True)
```
StrikerRUS: Seems that `X` and `y` remained untouched before this line, so there is no need to load data once again.

tests/python_package_test/test_engine.py
```diff
@@ -428,6 +428,72 @@ def test_multi_class_error(self):
...
+        results = {}
+        lgb.train(params, lgb_X, num_boost_round=100, valid_sets=[lgb_X], evals_result=results)
+        self.assertAlmostEqual(results['training']['auc_mu'][-1], 1)
+        # test loading weights
+        Xy = np.loadtxt(os.path.join(os.path.dirname(os.path.realpath(__file__)),
+                                     '../../examples/multiclass_classification/multiclass.train'))
+        y = Xy[:, 0]
+        X = Xy[:, 1:]
+        lgb_X = lgb.Dataset(X, label=y)
+        weights_file = os.path.join(os.path.join(os.path.dirname(os.path.realpath(__file__)),
```
StrikerRUS: One call of `os.path.join()` should be enough, I believe.

src/io/config.cpp
btrotta: Thanks, changed as you suggested. (The weights file is small, the number of lines is equal to number of classes.) If the file is not provided, the default is equal weights for all classes (i.e. matrix is all ones with zeros on diagonal).

R-package/R/lgb.Booster.R
```diff
@@ -563,7 +563,7 @@ Booster <- R6::R6Class(
           # Parse and store privately names
           names <- strsplit(names, "\t")[[1]]
           private$eval_names <- names
-          private$higher_better_inner_eval <- grepl("^ndcg|^map|^auc$", names)
+          private$higher_better_inner_eval <- grepl("^ndcg|^map|^auc", names)
```
jameslamb: looks ok to me

include/LightGBM/config.h
```diff
@@ -806,6 +807,12 @@ struct Config {
   // desc = when ``multi_error_top_k=1`` this is equivalent to the usual multi-error metric
   int multi_error_top_k = 1;
 
+  // desc = file containing matrix of weights for calculating AUC-mu multi-class metric
+  // desc = matrix must be square, with number of rows and columns equal to number of classes
+  // desc = if not specified, will use equal weights for all classes
+  // desc = see `this file <https://github.com/microsoft/LightGBM/tree/master/examples/multiclass_classification/loss.weights>`__ as an example
+  std::string auc_mu_weights_file = "";
```
guolinke: python/R package should pass a file as well?
Could we have another hyper-parameter, maybe called `auc_mu_weights`, whose size is n*n. And you can use `auc_mu_weights[i * c + j]` to get the `auc_mu_weights[i][j]`. We have many other hyper-parameters which are array format too, like `label_gain`.
Then, it will be easier to use in python/R package. And maybe `auc_mu_weights_file` could be removed too.

btrotta: Thanks, I made this change and removed the parameter `auc_mu_weights_file`.

include/LightGBM/config.h
```diff
@@ -806,6 +807,14 @@ struct Config {
   // desc = when ``multi_error_top_k=1`` this is equivalent to the usual multi-error metric
   int multi_error_top_k = 1;
 
+  // type = multi-double
+  // default = None
+  // desc = list representing flattened matrix (in row-major order) giving loss weights for classification errors
+  // desc = list should have ``n * n`` elements, where ``n`` is the number of classes
+  // desc = the matrix co-ordinate ``[i, j]`` should correspond to the ``i * n + j``-th element of the list
+  // desc = if not specified, will use equal weights for all classes
+  std::vector<double> auc_mu_weights;
```
StrikerRUS: After removing example file with weights, I expect some issues in the future with questions about how to use `auc_mu_weights`. I'm not saying that the description is not clear, it's perfect! But still, it'll be nice to have an example we can refer as an answer to those questions.

Can we append `auc_mu` here 
https://github.com/microsoft/LightGBM/blob/6129208006c91f9ac94891690aca6f0c49d9b08e/examples/multiclass_classification/train.conf#L24
and add `auc_mu_weights` example in that config with original matrix view in comments like
```
#  0  1  2  3  4
#  5  0  6  7  8
#  9 10  0 11 12
# 13 14 15  0 16
# 17 18 19 20  0
auc_mu_weights = 0,1,2,3,4,5,0,6,7,8,9,10,0,11,12,13,14,15,0,16,17,18,19,20,0
```
?

PR #791, opened 2017-08-06
[R-package] CV Improvements

Laurae2:
- [x] Handle cross-validation split by groups
- [x] Handle cross-validation info splits correctly (`init_score`, `weight`)
- [x] Refine random splitting by adjusting automatically fold size during split generation (instead of at the beginning)
- [ ] in another PR if feasible: add a slice wrapper to keep info splits correctly

Fixes:

- [R-package] Attempt to add group to CV using RLE [WIP]: https://github.com/Microsoft/LightGBM/pull/397
- [R] Lambdarank cross validation: https://github.com/Microsoft/LightGBM/issues/390
- Attributes aren't visible inside custom evaluation function in CV https://github.com/Microsoft/LightGBM/issues/581

Install: 

```r
devtools::install_github("Laurae2/LightGBM/R-package@patch-1")
```

Allows CV split by groups, average gain is 1 (perfect):

```r
library(lightgbm)
data(agaricus.train, package = "lightgbm")
train <- agaricus.train
dtrain <- lgb.Dataset(train$data, label = train$label, group = c(rep(500, 12), 513), weight = inverse.rle(list(lengths = c(rep(500, 12), 513), values = 1:13)), init_score = inverse.rle(list(lengths = c(rep(500, 12), 513), values = (1:13) * 2)))
params <- list(objective = "lambdarank",
               metric = "ndcg",
               ndcg_at = "1,2,3",
               metric_freq = 1,
               max_position = 3,
               label_gain = "0,1,3,7")
model <- lgb.cv(params,
                dtrain,
                10,
                nfold = 4,
                min_data = 1,
                learning_rate = 1,
                early_stopping_rounds = 10)
```

Allows to keep additional info (weight, init_score) when doing CV:

```r
lgb.unloader(wipe = TRUE)
library(lightgbm)
data(agaricus.train, package = "lightgbm")
train <- agaricus.train
dtrain <- lgb.Dataset(train$data, label = train$label, weight = inverse.rle(list(lengths = c(rep(500, 12), 513), values = 1:13)), init_score = inverse.rle(list(lengths = c(rep(500, 12), 513), values = (1:13) * 2)))
params <- list(objective = "regression", metric = "l2")

# PERFORMANCE VERY POOR, THIS IS NORMAL
# Look at the init_score values, so poor default settings
model <- lgb.cv(params,
                dtrain,
                1000,
                nfold = 4,
                min_data = 1,
                learning_rate = 0.01,
                early_stopping_rounds = 10)
```

[...]

PR #397, opened 2017-04-10
[R-package] Attempt to add group to CV using RLE [WIP]

Laurae2:
Attempt to fix issue #390.

Test on the following:

```r
# lgb.unloader(wipe = TRUE) # Instant environment wiping
library(lightgbm)
data(agaricus.train, package = "lightgbm")
train <- agaricus.train
train$label[train$data[, 126] == 1] <- 3
train$label <- train$label + 1
digits <- c(1000, 1999, 2001, 1513)
dtrain <- lgb.Dataset(train$data, label = train$label, group = digits, free_raw_data = FALSE)
params <- list(objective = "lambdarank",
               metric = "ncdg",
               ncdg_at = "1,2,3",
               metric_freq = 1,
               max_position = 3,
               label_gain = "0,1,3,7")
model <- lgb.cv(params,
                dtrain,
                10,
                nfold = 4,
                min_data = 1,
                learning_rate = 1,
                early_stopping_rounds = 10,
                folds = list(1:1000,
                             1001:2999,
                             3000:5000,
                             5001:6513))
```

PR #2573, opened 2019-11-16
[R-package] fixed sorting issues in lgb.cv() when using a model with observation weights (fixes #2572)

### LINKED ISSUE #2572: [R-package] sorting of observation weights and initial scores is broken in lgb.cv()
In #2503 , it was reported that using certain sampling strategies could cause incorrect results if the indices for subsetting weren't sorted. This was "fixed" in the R package by #2524 , however that PR introduced a bug noted in https://github.com/microsoft/LightGBM/pull/2524#issuecomment-545719165.

Essentially, `slice.lgb.Dataset()` takes in a vector of indices (row numbers) and returns a new `Dataset` with that subset. On its way to doing that, it sorts those indices. If a model had observation weights or initial scores, they don't get sorted the same way, meaning they'll refer to the wrong observations.

To fix this, need to ensure that tuples of indices, weights, and initial scores are all kept together when re-sorting indices.

jameslamb:
See #2572 for details.

### DIFF
--- a/R-package/R/lgb.cv.R
+++ b/R-package/R/lgb.cv.R
@@ -193,12 +193,12 @@ lgb.cv <- function(params = list()
 
     # Create folds
     folds <- generate.cv.folds(
-      nfold
-      , nrow(data)
-      , stratified
-      , getinfo(data, "label")
-      , getinfo(data, "group")
-      , params
+      nfold = nfold
+      , nrows = nrow(data)
+      , stratified = stratified
+      , label = getinfo(data, "label")
+      , group = getinfo(data, "group")
+      , params = params
     )
 
   }
@@ -271,12 +271,19 @@ lgb.cv <- function(params = list()
   bst_folds <- lapply(
     X = seq_along(folds)
     , FUN = function(k) {
-      if (is.list(folds[[k]])) {
+
+      # For learning-to-rank, each fold is a named list with two elements:
+      #   * `fold` = an integer vector of row indices
+      #   * `group` = an integer vector describing which groups are in the fold
+      # For classification or regression tasks, it will just be an integer
+      # vector of row indices
+      folds_have_group <- "group" %in% names(folds[[k]])
+      if (folds_have_group) {
         test_indices <- folds[[k]]$fold
-        group <- folds[[k]]$group
+        test_groups <- folds[[k]]$group
+        train_groups <- getinfo(data, "group")[-test_groups]
       } else {
         test_indices <- folds[[k]]
-        group <- NULL
       }
       train_indices <- seq_len(nrow(data))[-test_indices]
 
@@ -302,9 +309,9 @@ lgb.cv <- function(params = list()
       setinfo(dtrain, "weight", indexDT$weight)
       setinfo(dtrain, "init_score", indexDT$init_score)
 
-      if (!is.null(group)) {
-        setinfo(dtest, "group", group)
-        setinfo(dtrain, "group", group)
+      if (folds_have_group) {
+          setinfo(dtest, "group", test_groups)
+          setinfo(dtrain, "group", train_groups)
       }
 
       booster <- Booster$new(params, dtrain)
--- a/R-package/R/lgb.cv.R
+++ b/R-package/R/lgb.cv.R
@@ -280,8 +280,9 @@ lgb.cv <- function(params = list()
       folds_have_group <- "group" %in% names(folds[[k]])
       if (folds_have_group) {
         test_indices <- folds[[k]]$fold
-        test_groups <- folds[[k]]$group
-        train_groups <- getinfo(data, "group")[-test_groups]
+        test_group_indices <- folds[[k]]$group
+        test_groups <- getinfo(data, "group")[test_group_indices]
+        train_groups <- getinfo(data, "group")[-test_group_indices]
       } else {
         test_indices <- folds[[k]]
       }
@@ -310,8 +311,8 @@ lgb.cv <- function(params = list()
       setinfo(dtrain, "init_score", indexDT$init_score)
 
       if (folds_have_group) {
-          setinfo(dtest, "group", test_groups)
-          setinfo(dtrain, "group", train_groups)
+        setinfo(dtest, "group", test_groups)
+        setinfo(dtrain, "group", train_groups)
       }
 
       booster <- Booster$new(params, dtrain)
--- a/R-package/R/lgb.cv.R
+++ b/R-package/R/lgb.cv.R
@@ -70,6 +70,7 @@ CVBooster <- R6::R6Class(
 #'   , learning_rate = 1.0
 #'   , early_stopping_rounds = 5L
 #' )
+#' @importFrom data.table data.table setorderv
 #' @export
 lgb.cv <- function(params = list()
                    , data
@@ -95,8 +96,7 @@ lgb.cv <- function(params = list()
                    ) {
 
   # Setup temporary variables
-  addiction_params <- list(...)
-  params <- append(params, addiction_params)
+  params <- append(params, list(...))
   params$verbose <- verbose
   params <- lgb.check.obj(params, obj)
   params <- lgb.check.eval(params, eval)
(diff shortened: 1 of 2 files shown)


StrikerRUS:
Please rebase to the latest `master` to spot some linting issues.

R-package/R/lgb.cv.R
```diff
@@ -267,35 +267,56 @@ lgb.cv <- function(params = list()
...
+        group <- folds[[k]]$group
+      } else {
+        test_indices <- folds[[k]]
+        group <- NULL
+      }
+      train_indices <- seq_len(nrow(data))[-test_indices]
+
+      # set up test set
+      indexDT <- data.table::data.table(
+        indices = train_indices
```
StrikerRUS: `train_indices` to set up "test set"?

R-package/R/lgb.cv.R
```diff
@@ -267,35 +267,56 @@ lgb.cv <- function(params = list()
...
+        , init_score = getinfo(data, "init_score")[train_indices]
+      )
+      data.table::setorderv(indexDT, "indices", order = 1L)
+      dtrain <- slice(data, indexDT$indices)
+      setinfo(dtrain, "weight", indexDT$weight)
+      setinfo(dtrain, "init_score", indexDT$init_score)
+
+      if (!is.null(group)) {
+        setinfo(dtest, "group", group)
+        setinfo(dtrain, "group", group)
```
StrikerRUS: Is `group` fields shared across train and test sets?

jameslamb:
> Sorry if my comments are not so smart, but since review was requested from me, let me show which things look confusing from point of view of first-time code reader.

thanks for the review! I really appreciate it. They are good comments 😀

R-package/R/lgb.cv.R
jameslamb: it was in the original code. Let me double-check what that is doing.

R-package/R/lgb.cv.R
jameslamb: ah! there is a lot of repetition in this code right now between train and test sets, so this was bound to happen. Let me think if there is a cleaner way to do this that reduces the duplication...I think I have an idea how

R-package/R/lgb.cv.R
jameslamb: ok I see now that I misunderstood how `group` is working. I haven't worked with the learning-to-rank side of `LightGBM` at all, and I'm struggling to understand [this documentation](https://github.com/microsoft/LightGBM/blob/master/docs/Parameters.rst#query-data).

It seems that `group` going to a vector like `c(10,  15, 20)` which means "the first 10 records are from one query, the next 15 are samples from another query, etc.".

If that's true, I don't understand what `getinfo(data, "group")[-folds[[k]]$group]` could be doing. We don't have any R learning-to-rank examples I could find, so I need some help from @guolinke or someone else.

guolinke:
Very sorry for the late response, I will take a look today

R-package/R/lgb.cv.R
guolinke: please refer to the python implementation: https://github.com/microsoft/LightGBM/blob/master/python-package/lightgbm/engine.py#L308-L324

guolinke: I think there is a similar logics in R: https://github.com/microsoft/LightGBM/blob/master/R-package/R/lgb.cv.R#L409-L438

jameslamb: Sorry, I have been staring at this code for a while and am still confused what it's doing. I am going to try to figure out a learning-to-rank example in R (since we don't have any in the documentation) and use that to step through the code. Going to try to figure out how to do this from the Python tests using lambdarank: https://github.com/microsoft/LightGBM/blob/master/tests/python_package_test/test_engine.py#L635

guolinke: I remember this is written by @Laurae2

jameslamb: I think this is working! Thanks to @Laurae2 for pointing me to the examples of learning-to-rank code for the R package in #791 and #397. Based on those, I've added unit tests on `lgb.train()` and `lgb.cv()` that at least cover the LTR-specific branches of that code.

Reviewers, could you take another look at this PR when you have a chance?

jameslamb: oh also I did rebase to most recent `master`, to get the recent minor changes to the R package included (#2652 and #2654) in the tests

R-package/tests/testthat/test_learning_to_rank.R
```diff
@@ -0,0 +1,94 @@
...
+    )
+    nfold <- 4L
+    model <- lgb.cv(
+        params = params
+        , data = dtrain
+        , nrounds = 10L
+        , nfold = nfold
+        , min_data = 1L
+        , learning_rate = 1.0
+        , early_stopping_rounds = 10L
```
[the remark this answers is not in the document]
jameslamb: 😬 bad copy-pasting, thank you

R-package/tests/testthat/test_learning_to_rank.R
```diff
@@ -0,0 +1,94 @@
+context("Learning to rank")
```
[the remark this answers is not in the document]
jameslamb: totally agree. Since this is currently a user-facing bug (i.e. the functionality is already incorrect on `master`, #2572 ), I wanted to get something that I think is a fix up as fast as possible. I'm happy to keep expanding these tests and pushing on this PR until @Laurae2 has a chance to review the changes to `lgb.cv()`, but once we get that approval I don't think we should delay merging this for the sake of more testing.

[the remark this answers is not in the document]
StrikerRUS: Sure! I think adding a check for adequate values of NDCG will be enough for now.

R-package/tests/testthat/test_learning_to_rank.R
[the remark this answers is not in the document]
jameslamb: I decided to drop early stopping altogether in these tests

R-package/tests/testthat/test_learning_to_rank.R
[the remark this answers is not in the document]
jameslamb: Thanks for the  feedback @StrikerRUS , very good idea. I've added more to these tests in https://github.com/microsoft/LightGBM/pull/2573/commits/f14458bd3678d8b8883f8a8c38a750c769b8f223

I'm testing the  following things for  learning-to-rank with  `lambdarank`.  I think this should be sufficient.

`lgb.train()`

* returns an `lgb.Booster`  object, with the following properties
    - `max_feature_idx` matching dimensions of the training data
    -  one set  of `eval_results` per metric (3 in these tests, since I had  `ndcg_at = "1,2,3"`
     - all evaluation results have reasonable values (between 0.0 and 1.0 since they're nDCG)

`lgb.cv()`

* returns an `lgb.CVBooster` object, with the following properties
    - `boosters`, containing a list of length `nfold`  of  `lgb.Booster`
    - `best_iter`  with a reasonable value  (>1, <=  `nrounds`)
    - `best_score`  is based on the first metric (`ncdg@1`, in this case)
    -  one set  of `eval_results` per metric (3 in these tests, since I had  `ndcg_at = "1,2,3"`. For which:
         - `eval` and `eval_err`  are arrays with length `nrounds`  (since  I did not use  early stopping)
          - all `eval`  values  have reasonable values (between 0.0 and 1.0 since they're nDCG)


@guolinke or  @Laurae2 could I get one more review of the changes in this PR from one of you?

jameslamb:
I see there are some lingering linting issues. Will fix shortly!

R-package/tests/testthat/test_learning_to_rank.R
```diff
@@ -0,0 +1,118 @@
...
+
+test_that("learning-to-rank with lgb.cv() works as expected", {
+    data(agaricus.train, package = "lightgbm")
+    # just keep a few features,to generate an model with imperfect fit
+    train <- agaricus.train
+    train_data <- train$data[, 1:20]
+    dtrain <- lgb.Dataset(
+        train_data
+        , label = train$label
+        , group = c(rep(4L, 1500L), 513L)
```
Laurae2: Should there be 1501 groups? (1500 groups of length 4 and 1 group of length 513?)

R-package/tests/testthat/test_learning_to_rank.R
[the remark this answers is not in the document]
StrikerRUS: @jameslamb 

> all evaluation results have reasonable values (between 0.0 and 1.0 since they're nDCG)

I think that tests can be more strict to the final metric value in that aspect to ensure that training actually happens. Refer to Python tests as an example:
https://github.com/microsoft/LightGBM/blob/301402c83bc732f4ec41135cc5e5f6f5984348d8/tests/python_package_test/test_sklearn.py#L117-L118

R-package/tests/testthat/test_learning_to_rank.R
jameslamb: That was my intention, but it's just an artifact of copying other code from the examples you shared. I "one big group" of 513 is probably unnecessary...I'm going to drop this so we keep only the first 6000 observations and split them into 1500 groups of 4, that's cleaner.

R-package/tests/testthat/test_learning_to_rank.R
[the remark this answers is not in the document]
jameslamb: oh sure I can go down to that level. I avoided it because I thought it would make the tests too fragile (even with setting the seed), but if it's been working on the Python side then I'm ok with it. Let me try that.

jameslamb:
> You can merge and lint after on a separate PR.

Thanks for the review! I'm going to fix the linting issues right now. I think it's better that we don't get in the habit of skipping linting and coming back to it later.

R-package/tests/testthat/test_learning_to_rank.R
jameslamb: fixed in https://github.com/microsoft/LightGBM/pull/2573/commits/ead17bcc6ddb06fa506071e76ce0bba213101f8c

R-package/tests/testthat/test_learning_to_rank.R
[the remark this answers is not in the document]
jameslamb: thanks @StrikerRUS , I've made these tests much more specific in https://github.com/microsoft/LightGBM/pull/2573/commits/ead17bcc6ddb06fa506071e76ce0bba213101f8c. Could you take a look?

jameslamb:
Thanks for the reviews @Laurae2 and @StrikerRUS . I think we are really close! I've addressed linting and other issues in the tests in this commit: https://github.com/microsoft/LightGBM/pull/2573/commits/ead17bcc6ddb06fa506071e76ce0bba213101f8c

R-package/tests/testthat/test_learning_to_rank.R
jameslamb: fixed in  https://github.com/microsoft/LightGBM/pull/2573/commits/ee037c9a2f32415c7bd7e63a4d1c093687d110e8 (had to rebase to `master` and force push with other R PRs that were just merged)

R-package/tests/testthat/test_learning_to_rank.R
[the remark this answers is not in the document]
jameslamb: fixed in  https://github.com/microsoft/LightGBM/pull/2573/commits/ee037c9a2f32415c7bd7e63a4d1c093687d110e8 (had to rebase to `master` and force push with other R PRs that were just merged)

jameslamb:
The failures in CI on the previous commit were caused by [this interesting behavior](https://github.com/mailund/pmatch/issues/65#issuecomment-454688322), where basically I was using `&&` when I should have been using  `&`,  and the code range fine on my machine and failed with  `r-devel`  releases.

`&&` should only be  used with inputs that are length-1, and I was using it (incorrectly) to compare to vectors. Just pushed a fix and hopefully that will work.

R-package/tests/testthat/test_learning_to_rank.R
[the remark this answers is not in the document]
StrikerRUS: @jameslamb Thanks you! Tests are more reliable now!

jameslamb:
Thanks for the reviews @Laurae2 and @StrikerRUS . I'm happy to get this one merged and fix this issue.

PR #2540, opened 2019-11-05
Fix build error when HDFS is enabled.

shishaochen:
Without this revert, error messages come like:
```
/data00/home/shishaochen/github-repos/microsoft/LightGBM/src/io/file_io.cpp: In static member function ‘static hdfs_internal* LightGBM::HDFSFile::GetHDFSFileSystem(const string&)’:
/data00/home/shishaochen/github-repos/microsoft/LightGBM/src/io/file_io.cpp:120:43: error: request for member ‘length’ in ‘LightGBM::kHdfsProto’, which is of non-class type ‘const char*’
     size_t end = uri.find("/", kHdfsProto.length());
                                           ^
/data00/home/shishaochen/github-repos/microsoft/LightGBM/src/io/file_io.cpp:125:50: error: request for member ‘length’ in ‘LightGBM::kHdfsProto’, which is of non-class type ‘const char*’
     std::string hostport = uri.substr(kHdfsProto.length(), end - kHdfsProto.length());
                                                  ^
/data00/home/shishaochen/github-repos/microsoft/LightGBM/src/io/file_io.cpp:125:77: error: request for member ‘length’ in ‘LightGBM::kHdfsProto’, which is of non-class type ‘const char*’
     std::string hostport = uri.substr(kHdfsProto.length(), end - kHdfsProto.length());
                                                                             ^
At global scope:
cc1plus: warning: unrecognized command line option "-Wno-ignored-attributes"
CMakeFiles/_lightgbm.dir/build.make:231: recipe for target 'CMakeFiles/_lightgbm.dir/src/io/file_io.cpp.o' failed
```
This build error is brought from #2426.

PR #2586, opened 2019-11-22
fix compile warning for HDFS version

StrikerRUS:
Refer to https://github.com/microsoft/LightGBM/pull/2540#issuecomment-557579831.

### DIFF
--- a/src/io/file_io.cpp
+++ b/src/io/file_io.cpp
@@ -55,9 +55,10 @@ struct LocalFile : VirtualFileReader, VirtualFileWriter {
 };
 
 const char* kHdfsProto = "hdfs://";
-const size_t kHdfsProtoLength = strlen(kHdfsProto);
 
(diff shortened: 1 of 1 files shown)


src/io/file_io.cpp
```diff
@@ -55,9 +55,10 @@ struct LocalFile : VirtualFileReader, VirtualFileWriter {
 };
 
 const char* kHdfsProto = "hdfs://";
-const size_t kHdfsProtoLength = strlen(kHdfsProto);
 
 #ifdef USE_HDFS
+const size_t kHdfsProtoLength = strlen(kHdfsProto);
```
guolinke: maybe ·const size_t kHdfsProtoLength = static_cast<size_t>(strlen(kHdfsProto));` ?

StrikerRUS: Done, thanks!

PR #2587, opened 2019-11-22
Add more debug logging to show data load progress.

ashok-ponnuswami-msft:
1) Add file load progress after every 10 GBs. 2) Print number of data points in the validation set(s) as well. Update the text for the log line with number train set data points to be consistent with it. 3) Fix printf format string in a label range error message.

### DIFF
--- a/include/LightGBM/utils/text_reader.h
+++ b/include/LightGBM/utils/text_reader.h
@@ -28,8 +28,8 @@ class TextReader {
   * \param filename Filename of data
   * \param is_skip_first_line True if need to skip header
   */
-  TextReader(const char* filename, bool is_skip_first_line):
-    filename_(filename), is_skip_first_line_(is_skip_first_line) {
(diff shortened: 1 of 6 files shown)


include/LightGBM/utils/text_reader.h
```diff
@@ -119,6 +120,15 @@ class TextReader {
       if (last_i != read_cnt) {
         last_line_.append(buffer_process + last_i, read_cnt - last_i);
       }
+
+      size_t prev_bytes_read = bytes_read;
+      bytes_read += read_cnt;
+      if (prev_bytes_read / read_progress_interval_bytes_ < bytes_read / read_progress_interval_bytes_)
```
guolinke: I am a little bit confused about this `if`. 
maybe `prev_bytes_read` should be in this `if` block?

ashok-ponnuswami-msft: This is just checking if the integer value after dividing by read_progress_interval_bytes_ changed before and after reading the current block of data. So if the previous bytes read was 9.9GB, and now it is 10.1GB, we will print a log line.

PR #2588, opened 2019-11-22
LightGBM would hang on MPI run if only some nodes have a fatal error.…

ashok-ponnuswami-msft:
… The issue is that the destructor of Application calls MPI_Finalize(), which will cause the program to hand and prevent from exiting. We could call MPI_Abort(), but it appears to have the same affect as abort(). So we just use abort() in MPI mode if there is an unhandled exception.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -9,7 +9,38 @@
 int main(int argc, char** argv) {
   try {
     LightGBM::Application app(argc, argv);
-    app.Run();
+
+    std::string exception_str;
+    try
+    {
+      app.Run();
+    }
+    catch (const std::exception& ex) {
+      exception_str = ex.what();
+    }
+    catch (const std::string& ex) {
+      exception_str = ex;
+    }
+    catch (...) {
+      exception_str = "Unknown Exceptions";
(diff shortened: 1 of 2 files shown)


src/main.cpp
```diff
@@ -9,7 +9,38 @@
...
+      // Instead we may want to call MPI_Abort(). However that right now doesn't do much more call abort(). So we just do it here.
+      if (app.IsParallel())
+      {
+        std::cerr << "Aborting because of exception in MPI mode:" << std::endl;
+        std::cerr << exception_str << std::endl;
+        std::cerr.flush();
+        abort();
+      }
+#endif
+      throw exception_str;
```
guolinke: Thanks so much! Indeed the destructor will cause the problem. 
However, I think the two-level `try ... catch` may is not a good fix. 
I think a possible solution to move the `MPI_Finalize` out of destructor, and call `MPI_Abort` if any exception met, otherwise call the `MPI_Finalize`.

ashok-ponnuswami-msft:
Will create a new pull request with requested changes.

Issue #2208, opened 2019-06-01
incorrect logged parameters in model file


guolinke:
There is a config2string after training model:
https://github.com/microsoft/LightGBM/blob/b1e5a843ab57c64b01c8896ef3cda6fdf907d417/src/io/config_auto.cpp#L549-L654

However, as some parameters are not visible in cpp side when using python/R package, such as `num_iteration` in python package. Therefore, some logged parameters are not correct.

guolinke:
to fix this, I think we should pass all parameters to cpp code, even though some of them are not used in cpp side.

Xiad32:
So, what other configurations are needed here beside num_iteration? Would they differ based on the class where they are used; RF,GOSS or DART?

guolinke:
@Xiad32 these configurations just log the error values in the model file, it doesn't affect the normal usage.

StrikerRUS:
Python-package has been fixed in #2589.

ping @jameslamb and @Laurae2 for R-package.

jameslamb:
@StrikerRUS hey thanks for the ping, this one was not on my radar. I'll pick it up

StrikerRUS:
@jameslamb Great! Thank you.

PR #2589, opened 2019-11-24
[python] save all param values into model file

StrikerRUS:
Fix for Python part of #2208.


python-package/lightgbm/basic.py
```diff
@@ -2407,6 +2407,13 @@ def predict(self, data, num_iteration=None,
         result : numpy array
             Prediction result.
         """
+        params_to_update = copy.deepcopy(self.params)
+        params_to_update.update(dict(kwargs,
+                                     predict_raw_score=raw_score,
+                                     predict_leaf_index=pred_leaf,
+                                     predict_contrib=pred_contrib,
+                                     num_iteration_predict=num_iteration))
+        self.reset_parameter(params_to_update)
```
StrikerRUS: @guolinke This seems to be quite computationally expensive part and it doesn't work in case params contain any "core" parameters, e.g.
```
[LightGBM] [Fatal] Cannot change metric during training
[LightGBM] [Fatal] Cannot change num_class during training
[LightGBM] [Fatal] Cannot change boosting during training
```

Maybe we can put predict parameters into config directly at cpp side during prediction itself?

guolinke: Yeah, use reset parameter is not a good idea. BTW, do we need to save predict parameters?

StrikerRUS: Do you mean remove predict params from model file completely? I think it makes sense, because they are not needed to restore a model and seems to be more "logging" stuff (or at least something different than params used to train model). I think we can introduce naming schema like `predict_*` for them and filter later to not write params in model file which start with `predict_`.

guolinke: yeah, it makes senses.

StrikerRUS: OK, then I'll remove my attempts to record prediction time params from here. Can you please help with cpp part to not store them at all (here or in a separate PR)?

guolinke: I think it is generated automatically: https://github.com/microsoft/LightGBM/blob/master/src/io/config_auto.cpp#L660-L663. 
Maybe we can add a tag, e.g. `[no-save]`, to these parameters, and skip their `to_string`.

StrikerRUS: Cool idea!

StrikerRUS: Hmmm, seems that we already have `[doc-only]` directive:
https://github.com/microsoft/LightGBM/blob/dc65e0ac6ee8faa5790263a8dc2d9cc2a8dd7c4b/helpers/parameter_generator.py#L301-L302

BTW, why are for example `boosting` and `objective` already `doc-only`?

guolinke: As they are not automatically generated. We manually write their code.

StrikerRUS: @guolinke 
> Maybe we can add a tag, e.g. [no-save], to these parameters, and skip them in to_string.

I've added `[no-save]` tag in the latest commit and applied it to `predict` and `convert_model` tasks' params.
Maybe we can apply it to "output" training values as well? I mean, params like
https://github.com/microsoft/LightGBM/blob/102893a11c1dd712bb70d87b3b129ddf42daf06f/include/LightGBM/config.h#L488-L491
https://github.com/microsoft/LightGBM/blob/102893a11c1dd712bb70d87b3b129ddf42daf06f/include/LightGBM/config.h#L564-L567
https://github.com/microsoft/LightGBM/blob/102893a11c1dd712bb70d87b3b129ddf42daf06f/include/LightGBM/config.h#L784-L787

StrikerRUS:
@guolinke Just out of curiosity, what is the reason to keep text and JSON representations different?

StrikerRUS:
@jameslamb @Laurae2 Would you mind fixing R-package right here or prefer create a separate PR later?

StrikerRUS:
@guolinke One more problem is that it seems params for `Dataset` are not saved at all:
```
import numpy as np
import lightgbm as lgb

X = np.random.random((100, 2))
y = np.random.random(100)
lgb_data = lgb.Dataset(X, y, categorical_feature=[0, 1], params={"max_bin": 100})
bst = lgb.train({}, lgb_data, num_boost_round=5)
bst.save_model('model.txt')
```

```
...
[max_bin: 255]
...
[categorical_feature: ]
...
```

guolinke:
@StrikerRUS yes, it is a problem. Currently, we only copy the params in lgb.train/... to the Dataset, but the copy to Booster from Dataset.

guolinke:
@StrikerRUS 

> @guolinke Just out of curiosity, what is the reason to keep text and JSON representations different?

which part you refer to?

StrikerRUS:
@guolinke 
> yes, it is a problem. Currently, we only copy the params in lgb.train/... to the Dataset, but the copy to Booster from Dataset.

Do you have any thoughts how to fix it?..

> which part you refer to?

I mean, some fields that presented in text format are not included into JSON format and vice versa. For example, *parameters*, *feature importances*.


```
import json

import numpy as np
import lightgbm as lgb

X = np.random.random((100, 2))
y = np.random.random(100)
lgb_data = lgb.Dataset(X, y)
bst = lgb.train({}, lgb_data, num_boost_round=3)
bst.save_model('save_model.txt')
with open('dump_model.json', 'w') as json_dump:
    json.dump(bst.dump_model(), json_dump, indent=2)
```


image

image

guolinke:
maybe in Booster Construct:
https://github.com/microsoft/LightGBM/blob/bbc45fed573b246556055e4d4dc335b92442d603/python-package/lightgbm/basic.py#L1713-L1716
we can construct dataset first, and then copy its params to booster, then construct booster.

For the field mismatch, I think it is caused by multiple contributors, we can unify them.

StrikerRUS:
@guolinke 
> we can construct dataset first, and then copy its params to booster, then construct booster.

OK, I see. Let's work on that within #2208 after merging this PR and #2594.

> For the field mismatch, I think it is caused by multiple contributors, we can unify them.

OK, I think it'll be good refactoring to have and it will help a lot of third-party libraries that work with LightGBM model's dumps. I'll create a separate issue for this.

python-package/lightgbm/basic.py
StrikerRUS: @guolinke WDYT about the list of candidate params for the tag?

guolinke:
@StrikerRUS actually, I think the parameter write back should be in #2594, otherwise, the reset config checking for dataset may fail..

StrikerRUS:
Blocked by #2594.

StrikerRUS:
@guolinke 
As #2594 has been merged, I think we can get back to this PR. I'm copying my old comments from the thread above to make it easier to follow the discussion.


> Maybe we can add a tag, e.g. [no-save], to these parameters, and skip them in to_string.

> I've added `[no-save]` tag in the latest commit and applied it to `predict` and `convert_model` tasks' params.
Maybe we can apply it to "output" training values as well? I mean, params like
https://github.com/microsoft/LightGBM/blob/102893a11c1dd712bb70d87b3b129ddf42daf06f/include/LightGBM/config.h#L488-L491
https://github.com/microsoft/LightGBM/blob/102893a11c1dd712bb70d87b3b129ddf42daf06f/include/LightGBM/config.h#L564-L567
https://github.com/microsoft/LightGBM/blob/102893a11c1dd712bb70d87b3b129ddf42daf06f/include/LightGBM/config.h#L784-L787

> WDYT about the list of candidate params for the tag?

guolinke:
Thanks @StrikerRUS , I agree with you, some of the parameters don't need to save.
```
output_model
verbosity
metric_freq
save_binary 
```

StrikerRUS:
@guolinke I added some more params in the latest commit, please check. Maybe we need to ignore more params, e.g. `data`, `valid`?

guolinke:
@StrikerRUS I think `data` and `valid` need to save, for the cli users.

PR #2591, opened 2019-11-25
[docs] provide link to m2cgen from convert_model_language param description

StrikerRUS:
`m2cgen` can be used to transpile sklearn wrapper classes (JSON model representation actually) into native code for different programming languages. It is already in the list of our "friendly" third-party projects: https://github.com/microsoft/LightGBM#external-unofficial-repositories.


StrikerRUS:
@jameslamb Thanks for the review!

BTW, maybe you as R expert with great experience know some use cases when `m2cgen` is a good choice exactly for R?
https://github.com/BayesWitnesses/m2cgen/pull/139#partial-pull-merging

Issue #2578, opened 2019-11-18
Output tree structure to dataframe


pford221:
Hi -

Similar to how XGBoost provides a method for its Booster class to output the tree structures learned to a Pandas.DataFrame, would the maintainers have interest in having something similar for LightGBM?

I have a prototype working with Python API that I could clean up and submit a PR?

Here's the XGBoost Python implementation:
https://github.com/dmlc/xgboost/blob/master/python-package/xgboost/core.py#L1712

StrikerRUS:
Hi @pford221 !

Sure, go for it!

As I can observe, users are often interested in everything connected to model exploration.

StrikerRUS:
Closed via #2592.

Issue #2320, opened 2019-08-11
Example for calculating param values from model dump


Mahdi-Hosseinali:
The documentation has a nice guideline about [dealing with over fitting](https://lightgbm.readthedocs.io/en/latest/Parameters-Tuning.html), however, it is not clear how to set the mentioned parameters. For example, user does not have feedback about `num_leaves`, `min_data_in_leaf`, `min_gain_to_split`, or `max_depth` of a booster that is using default parameters but overfits. Although all four parameters are easy to calculate from the booster object, it would be nice to have them as inbuilt functions rather than copying the functions in every problem.

I've written the python code and would be happy to contribute them if I know how to do it properly. I understand that it might be a better to implement them in the core so all APIs have access to them. Or it might be too much of a hassle to guide me on how to implement them rather than translating the core by a conversant contributor, so here's the code. It outputs all four parameters as lists for each tree, so the user can plot them and decide what would be a better value.

```python
def depth(d):
    if isinstance(d, dict):
        return 1 + (max(map(depth, d.values())) if d else 0)
    return 0

def min_gain_to_split(tree):
    if 'split_index' in tree:
        lst = [tree['split_gain'],
               min_gain_to_split(tree['left_child']),
               min_gain_to_split(tree['right_child']) ]
        return min([i for i in lst if i>0])
    return -1

def min_data_in_leaf(tree):
    if 'leaf_count' in tree:
        return tree['leaf_count']
    return min(
        min_data_in_leaf(tree['left_child']),
        min_data_in_leaf(tree['right_child']) )

def get_tree_inf(booster):
    trees = booster.dump_model()['tree_info']
    leaves, depths, min_data, min_gains = [], [], [], []
    for tree in trees:
        leaves.append(tree['num_leaves'])
        tree_struct = tree['tree_structure']
        depths.append(depth(tree_struct))
        min_data.append(min_data_in_leaf(tree_struct))
        min_gains.append(min_gain_to_split(tree_struct))
    return [leaves, depths, min_data, min_gains]
```

StrikerRUS:
@Mahdi-Hosseinali Thanks for sharing this useful snippet! To be honest, I'm not sure this functionality should be a part of API. Maybe [`examples`](https://github.com/microsoft/LightGBM/tree/master/examples/python-guide)is better place for it?

Mahdi-Hosseinali:
that sounds good, which example do you think it would fit better?

StrikerRUS:
[...]

PR #2592, opened 2019-11-26
[python] Output model to a pandas DataFrame

pford221:
Added `trees_to_dataframe` method to Booster class in python API. Based on issue [#2578](https://github.com/microsoft/LightGBM/issues/2578).

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -1859,7 +1859,7 @@ class Booster(object):
 
         Returns
         -------
-        result: DataFrame
+        result: pandas DataFrame
             Returns a pandas DataFrame of the parsed model.
         """
         if not PANDAS_INSTALLED:
@@ -1878,7 +1878,8 @@ class Booster(object):
                 tree_num = str(tree_index) + '-' if tree_index is not None else ''
                 is_split = _is_split_node(tree)
                 node_type = 'S' if is_split else 'L'
-                node_num = str(tree['split_index' if is_split else 'leaf_index'])
+                # if a single node tree it won't have `leaf_index` so return 0
+                node_num = str(tree.get('split_index' if is_split else 'leaf_index', 0))
                 return tree_num + node_type + node_num
 
             def _get_split_feature(tree, feature_names):
@@ -1890,6 +1891,10 @@ class Booster(object):
                 else:
                     feature_name = None
                 return feature_name
+
+            def _is_single_node_tree(tree):
+                tree_keys = [str(key) for key in tree.keys()]
+                return tree_keys == ['leaf_value']
 
             # Create the node record, and populate universal data members
             node = OrderedDict()
@@ -1923,8 +1928,9 @@ class Booster(object):
                 node['count'] = tree['internal_count']
             else:
                 node['value'] = tree['leaf_value']
-                node['weight'] = tree['leaf_weight']
-                node['count'] = tree['leaf_count']
+                if not _is_single_node_tree(tree):
+                    node['weight'] = tree['leaf_weight']
+                    node['count'] = tree['leaf_count']
 
             return node
 
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -1860,10 +1860,10 @@ class Booster(object):
         Returns
         -------
         result: DataFrame
-            Returns a pandas DataFrame of the parsed model
+            Returns a pandas DataFrame of the parsed model.
         """
         if not PANDAS_INSTALLED:
-            raise LightGBMError('This method cannot be run without Pandas installed')
+            raise LightGBMError('This method cannot be run without pandas installed')
 
         if self.num_trees() == 0:
             raise LightGBMError('There are no trees in this Booster and thus nothing to parse')
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -1858,35 +1858,34 @@ class Booster(object):
 
         Returns
         -------
-        Dataframe or list
-            Returns a pandas Dataframe if pandas is installed. Otherwise returns a list.
+        DataFrame or list
+            Returns a pandas Dataframe if pandas is installed, and a list otherwise.
         """
         if not PANDAS_INSTALLED:
-            print('Pandas is not installed and so the result will be returned as a list')
+            warnings.warn('Pandas is not installed and so the result will be returned as a list')
 
         if self.num_trees() == 0:
-            print('There are no trees in this Booster and thus nothing to return')
-            return
+            raise LightGBMError('There are no trees in this Booster and thus nothing to parse')
 
-        def _is_split_node(t):
-            return 'split_index' in t.keys()
+        def _is_split_node(tree):
+            return 'split_index' in tree.keys()
 
-        def create_node_record(t, node_depth=1, tree_index=None,
+        def create_node_record(tree, node_depth=1, tree_index=None,
                                feature_names=None, parent_node=None):
 
-            def _get_node_index(t, tree_index):
+            def _get_node_index(tree, tree_index):
                 tree_num = str(tree_index) + '-' if tree_index is not None else ''
-                node_type = 'S' if _is_split_node(t) else 'L'
-                node_num = str(
-                    t['split_index' if _is_split_node(t) else 'leaf_index'])
+                is_split = _is_split_node(tree)
+                node_type = 'S' if is_split else 'L'
+                node_num = str(tree['split_index' if is_split else 'leaf_index'])
                 return tree_num + node_type + node_num
 
-            def _get_split_feature(t, feature_names):
-                if _is_split_node(t):
+            def _get_split_feature(tree, feature_names):
+                if _is_split_node(tree):
                     if feature_names is not None:
-                        feature_name = feature_names[t['split_feature']]
+                        feature_name = feature_names[tree['split_feature']]
                     else:
-                        feature_name = t['split_feature']
+                        feature_name = tree['split_feature']
                 else:
                     feature_name = None
                 return feature_name
@@ -1894,39 +1893,44 @@ class Booster(object):
             # Create the node record, and populate universal data members
             node = {'tree_index': tree_index,
                     'node_depth': node_depth,
-                    'node_index': _get_node_index(t, tree_index),
+                    'node_index': _get_node_index(tree, tree_index),
                     'left_child': None,
                     'right_child': None,
                     'parent_index': parent_node,
-                    'split_feature': _get_split_feature(t, feature_names),
+                    'split_feature': _get_split_feature(tree, feature_names),
                     'split_gain': None,
                     'threshold': None,
                     'decision_type': None,
                     'missing_direction': None,
                     'missing_type': None,
-                    'leaf_weight': None,
-                    'internal_count': None}
+                    'internal_value': None,
+                    'leaf_value': None,
+                    'weight': None,
+                    'count': None}
 
             # Update values to reflect node type (leaf or split)
-            if _is_split_node(t):
-                node['left_child'] = _get_node_index(t['left_child'], tree_index)
-                node['right_child'] = _get_node_index(t['right_child'], tree_index)
-                node['split_gain'] = t['split_gain']
-                node['threshold'] = t['threshold']
-                node['decision_type'] = t['decision_type']
-                node['missing_direction'] = 'left' if t['default_left'] else 'right'
-                node['missing_type'] = t['missing_type']
-                node['internal_count'] = t['internal_count']
+            if _is_split_node(tree):
+                node['left_child'] = _get_node_index(tree['left_child'], tree_index)
+                node['right_child'] = _get_node_index(tree['right_child'], tree_index)
+                node['split_gain'] = tree['split_gain']
+                node['threshold'] = tree['threshold']
+                node['decision_type'] = tree['decision_type']
+                node['missing_direction'] = 'left' if tree['default_left'] else 'right'
+                node['missing_type'] = tree['missing_type']
+                node['internal_value'] = tree['internal_value']
+                node['weight'] = tree['internal_weight']
+                node['count'] = tree['internal_count']
             else:
-                node['leaf_weight'] = t['leaf_value']
-                node['internal_count'] = t['leaf_count']
+                node['leaf_value'] = tree['leaf_value']
+                node['weight'] = tree['leaf_weight']
+                node['count'] = tree['leaf_count']
 
             return node
 
-        def tree_dict_to_node_list(t, node_depth=1, tree_index=None,
+        def tree_dict_to_node_list(tree, node_depth=1, tree_index=None,
                                    feature_names=None, parent_node=None):
 
-            node = create_node_record(t,
+            node = create_node_record(tree,
                                       node_depth=node_depth,
                                       tree_index=tree_index,
                                       feature_names=feature_names,
@@ -1934,33 +1938,33 @@ class Booster(object):
 
             res = [node]
 
-            if _is_split_node(t):
+            if _is_split_node(tree):
                 # traverse the next level of the tree
                 children = ['left_child', 'right_child']
                 for child in children:
                     subtree_list = tree_dict_to_node_list(
-                        t[child],
+                        tree[child],
                         node_depth=node_depth + 1,
                         tree_index=tree_index,
                         feature_names=feature_names,
                         parent_node=node['node_index'])
-                    # In tree format, "subtree_list"is a list of node records (dicts), and we
-                    # add node to the list.
+                    # In tree format, "subtree_list" is a list of node records (dicts),
+                    # and we add node to the list.
                     res.extend(subtree_list)
             return res
 
-        m_dict = self.dump_model()
-        feature_names = m_dict['feature_names']
-        m_list = []
-        for t in m_dict['tree_info']:
-            m_list.extend(tree_dict_to_node_list(t['tree_structure'],
-                                                 tree_index=t['tree_index'],
-                                                 feature_names=feature_names))
+        model_dict = self.dump_model()
+        feature_names = model_dict['feature_names']
+        model_list = []
+        for tree in model_dict['tree_info']:
+            model_list.extend(tree_dict_to_node_list(tree['tree_structure'],
+                                                     tree_index=tree['tree_index'],
+                                                     feature_names=feature_names))
 
         if PANDAS_INSTALLED:
-            return DataFrame(m_list, columns=m_list[0].keys())
+            return DataFrame(model_list, columns=model_list[0].keys())
         else:
-            return m_list
+            return model_list
 
     def set_train_data_name(self, name):
         """Set the name to the training Dataset.
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -1859,7 +1859,7 @@ class Booster(object):
 
         Returns
         -------
-        result: pandas DataFrame
+        result : pandas DataFrame
             Returns a pandas DataFrame of the parsed model.
         """
         if not PANDAS_INSTALLED:
@@ -1893,8 +1893,7 @@ class Booster(object):
                 return feature_name
 
             def _is_single_node_tree(tree):
-                tree_keys = [str(key) for key in tree.keys()]
-                return tree_keys == ['leaf_value']
+                return tree.keys() == {'leaf_value'}
 
             # Create the node record, and populate universal data members
             node = OrderedDict()
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -1853,6 +1853,115 @@ class Booster(object):
         self.network = False
         return self
 
+    def trees_to_dataframe(self):
+        """Parse the fitted model and return in an easy-to-read pandas DataFrame.
+
+        Returns
+        -------
+        Dataframe or list
+            Returns a pandas Dataframe if pandas is installed. Otherwise returns a list.
+        """
+        if not PANDAS_INSTALLED:
+            print('Pandas is not installed and so the result will be returned as a list')
+
+        if self.num_trees() == 0:
+            print('There are no trees in this Booster and thus nothing to return')
+            return
+
+        def _is_split_node(t):
+            return 'split_index' in t.keys()
+
+        def create_node_record(t, node_depth=1, tree_index=None,
+                               feature_names=None, parent_node=None):
+
+            def _get_node_index(t, tree_index):
+                tree_num = str(tree_index) + '-' if tree_index is not None else ''
+                node_type = 'S' if _is_split_node(t) else 'L'
+                node_num = str(
+                    t['split_index' if _is_split_node(t) else 'leaf_index'])
+                return tree_num + node_type + node_num
+
+            def _get_split_feature(t, feature_names):
+                if _is_split_node(t):
+                    if feature_names is not None:
+                        feature_name = feature_names[t['split_feature']]
+                    else:
+                        feature_name = t['split_feature']
+                else:
+                    feature_name = None
+                return feature_name
+
+            # Create the node record, and populate universal data members
+            node = {'tree_index': tree_index,
+                    'node_depth': node_depth,
+                    'node_index': _get_node_index(t, tree_index),
+                    'left_child': None,
+                    'right_child': None,
+                    'parent_index': parent_node,
+                    'split_feature': _get_split_feature(t, feature_names),
+                    'split_gain': None,
+                    'threshold': None,
+                    'decision_type': None,
+                    'missing_direction': None,
+                    'missing_type': None,
+                    'leaf_weight': None,
+                    'internal_count': None}
+
+            # Update values to reflect node type (leaf or split)
+            if _is_split_node(t):
+                node['left_child'] = _get_node_index(t['left_child'], tree_index)
+                node['right_child'] = _get_node_index(t['right_child'], tree_index)
+                node['split_gain'] = t['split_gain']
+                node['threshold'] = t['threshold']
+                node['decision_type'] = t['decision_type']
+                node['missing_direction'] = 'left' if t['default_left'] else 'right'
+                node['missing_type'] = t['missing_type']
+                node['internal_count'] = t['internal_count']
+            else:
+                node['leaf_weight'] = t['leaf_value']
+                node['internal_count'] = t['leaf_count']
+
+            return node
+
+        def tree_dict_to_node_list(t, node_depth=1, tree_index=None,
+                                   feature_names=None, parent_node=None):
+
+            node = create_node_record(t,
+                                      node_depth=node_depth,
+                                      tree_index=tree_index,
+                                      feature_names=feature_names,
+                                      parent_node=parent_node)
+
+            res = [node]
+
+            if _is_split_node(t):
+                # traverse the next level of the tree
+                children = ['left_child', 'right_child']
+                for child in children:
+                    subtree_list = tree_dict_to_node_list(
+                        t[child],
+                        node_depth=node_depth + 1,
+                        tree_index=tree_index,
+                        feature_names=feature_names,
+                        parent_node=node['node_index'])
+                    # In tree format, "subtree_list"is a list of node records (dicts), and we
+                    # add node to the list.
+                    res.extend(subtree_list)
+            return res
+
+        m_dict = self.dump_model()
+        feature_names = m_dict['feature_names']
+        m_list = []
+        for t in m_dict['tree_info']:
+            m_list.extend(tree_dict_to_node_list(t['tree_structure'],
+                                                 tree_index=t['tree_index'],
+                                                 feature_names=feature_names))
+
+        if PANDAS_INSTALLED:
+            return DataFrame(m_list, columns=m_list[0].keys())
+        else:
+            return m_list
+
     def set_train_data_name(self, name):
         """Set the name to the training Dataset.
 
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -7,6 +7,7 @@ import ctypes
 import os
 import warnings
 from tempfile import NamedTemporaryFile
+from collections import OrderedDict
 
 import numpy as np
 import scipy.sparse
@@ -1858,11 +1859,11 @@ class Booster(object):
 
         Returns
         -------
-        DataFrame or list
-            Returns a pandas Dataframe if pandas is installed, and a list otherwise.
+        result: DataFrame
+            Returns a pandas DataFrame of the parsed model
         """
         if not PANDAS_INSTALLED:
-            warnings.warn('Pandas is not installed and so the result will be returned as a list')
+            raise LightGBMError('This method cannot be run without Pandas installed')
 
         if self.num_trees() == 0:
             raise LightGBMError('There are no trees in this Booster and thus nothing to parse')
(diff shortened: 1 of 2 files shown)


pford221:
This is embarrassing. I left a python debugger `set_trace()` call in the unit test. Would you like me to resubmit the PR with it removed?

StrikerRUS:
You can simply add new commits in the `df_trees` branch and they will appear here.
And it seems that your new unit test fails itself. Please fix it along with PEP8 errors (you can find them here https://travis-ci.org/microsoft/LightGBM/jobs/616989890#L502). Thanks!

pford221:
Hi,

I have formatted according to pep8 standard and the unit test I added is working.

StrikerRUS:
Tests fail due to bad indents in the file. One more reminder: you are **NOT** restricted to 80 chars line length. I guess it was the reason you decided to change indents.
https://github.com/microsoft/LightGBM/tree/master/python-package#development-guide

pford221:
Hi. I appreciate your patience! I'm not too familiar with PEP 8 so I relied on the PEP 8 auto formatting python library and maybe it doesn't exactly conform to LightGBM's standards or maybe something happened in the copy and paste from the PEP 8 formatted .py file to basic.py. 

I went through all the travis-ci logs and corrected the formatting issues that I found. I then committed and squashed all the commits into one for easier review.

Thanks!

StrikerRUS:
No problem! Thank you very much for this PR! I'll review it shortly.

python-package/lightgbm/basic.py
```diff
@@ -1853,6 +1853,115 @@ def free_network(self):
         self.network = False
         return self
 
+    def trees_to_dataframe(self):
+        """Parse the fitted model and return in an easy-to-read pandas DataFrame.
+
+        Returns
+        -------
+        Dataframe or list
```
```suggestion
        result : DataFrame or list
```

python-package/lightgbm/basic.py
```diff
@@ -1853,6 +1853,115 @@ def free_network(self):
...
+            Returns a pandas Dataframe if pandas is installed. Otherwise returns a list.
```
```suggestion
            Returns a pandas DataFrame if pandas is installed, and a list otherwise.
```

python-package/lightgbm/basic.py
```diff
@@ -1853,6 +1853,115 @@ def free_network(self):
...
+        """
+        if not PANDAS_INSTALLED:
+            print('Pandas is not installed and so the result will be returned as a list')
```
```suggestion
            warnings.warn('Pandas is not installed and so the result will be returned as a list')
```

python-package/lightgbm/basic.py
```diff
@@ -1853,6 +1853,115 @@ def free_network(self):
...
+
+        if self.num_trees() == 0:
+            print('There are no trees in this Booster and thus nothing to return')
```
```suggestion
            raise LightGBMError('There are no trees in this Booster and thus nothing to parse')
```

python-package/lightgbm/basic.py
```diff
@@ -1853,6 +1853,115 @@ def free_network(self):
...
+            return
+
+        def _is_split_node(t):
```
StrikerRUS: Please do not use one-letter variable names (`t`).

python-package/lightgbm/basic.py
```diff
@@ -1853,6 +1853,115 @@ def free_network(self):
...
+            return 'split_index' in t.keys()
+
+        def create_node_record(t, node_depth=1, tree_index=None,
+                               feature_names=None, parent_node=None):
+
+            def _get_node_index(t, tree_index):
+                tree_num = str(tree_index) + '-' if tree_index is not None else ''
+                node_type = 'S' if _is_split_node(t) else 'L'
```
StrikerRUS: Cache result from `_is_split_node(t)` in a variable.

python-package/lightgbm/basic.py
```diff
@@ -1853,6 +1853,115 @@ def free_network(self):
...
+                    if feature_names is not None:
+                        feature_name = feature_names[t['split_feature']]
+                    else:
+                        feature_name = t['split_feature']
+                else:
+                    feature_name = None
+                return feature_name
+
+            # Create the node record, and populate universal data members
+            node = {'tree_index': tree_index,
```
StrikerRUS: I don't see `internal_value`, `internal_weight`, `leaf_value` and `leaf_weight` fields in this dict.

python-package/lightgbm/basic.py
```diff
@@ -1853,6 +1853,115 @@ def free_network(self):
...
+                node['left_child'] = _get_node_index(t['left_child'], tree_index)
+                node['right_child'] = _get_node_index(t['right_child'], tree_index)
+                node['split_gain'] = t['split_gain']
+                node['threshold'] = t['threshold']
+                node['decision_type'] = t['decision_type']
+                node['missing_direction'] = 'left' if t['default_left'] else 'right'
+                node['missing_type'] = t['missing_type']
+                node['internal_count'] = t['internal_count']
+            else:
+                node['leaf_weight'] = t['leaf_value']
```
StrikerRUS: Mistake here?

python-package/lightgbm/basic.py
```diff
@@ -1853,6 +1853,115 @@ def free_network(self):
...
+                node['internal_count'] = t['leaf_count']
```
StrikerRUS: If you use this key for both `internal_count` from non-terminal nodes and `leaf_count` from leaves, then it's better to name it simply `count`.

python-package/lightgbm/basic.py
```diff
@@ -1853,6 +1853,115 @@ def free_network(self):
...
+            return 'split_index' in t.keys()
+
+        def create_node_record(t, node_depth=1, tree_index=None,
+                               feature_names=None, parent_node=None):
+
+            def _get_node_index(t, tree_index):
+                tree_num = str(tree_index) + '-' if tree_index is not None else ''
+                node_type = 'S' if _is_split_node(t) else 'L'
+                node_num = str(
+                    t['split_index' if _is_split_node(t) else 'leaf_index'])
```
StrikerRUS: We do not have 80-chars line limit, so there is no need to split this quite short line.

python-package/lightgbm/basic.py
```diff
@@ -1853,6 +1853,115 @@ def free_network(self):
...
+                # traverse the next level of the tree
+                children = ['left_child', 'right_child']
+                for child in children:
+                    subtree_list = tree_dict_to_node_list(
+                        t[child],
+                        node_depth=node_depth + 1,
+                        tree_index=tree_index,
+                        feature_names=feature_names,
+                        parent_node=node['node_index'])
+                    # In tree format, "subtree_list"is a list of node records (dicts), and we
```
```suggestion
                    # In tree format, "subtree_list" is a list of node records (dicts), and we
```

python-package/lightgbm/basic.py
```diff
@@ -1853,6 +1853,115 @@ def free_network(self):
...
+                    # add node to the list.
+                    res.extend(subtree_list)
+            return res
+
+        m_dict = self.dump_model()
```
StrikerRUS: `model_dict` for better readability. Or what did you mean under the `m` letter?

python-package/lightgbm/basic.py
```diff
@@ -1853,6 +1853,115 @@ def free_network(self):
...
+        feature_names = m_dict['feature_names']
+        m_list = []
+        for t in m_dict['tree_info']:
+            m_list.extend(tree_dict_to_node_list(t['tree_structure'],
+                                                 tree_index=t['tree_index'],
+                                                 feature_names=feature_names))
+
+        if PANDAS_INSTALLED:
+            return DataFrame(m_list, columns=m_list[0].keys())
```
StrikerRUS: `m_list[0].keys()` is not expected to be sorted and it can cause problems in different Python versions/implementations:
> Changed in version 0.25.0: If data is a list of dicts, column order follows insertion-order for Python 3.6 and later.
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.html

I guess you can simply omit `columns` argument and pandas will name columns properly by default.

tests/python_package_test/test_basic.py
```diff
@@ -328,3 +328,47 @@ def check_asserts(data):
...
+                    imptcs.append(impcts_dict[col])
+                except KeyError:
+                    imptcs.append(0.)
+            return np.array(imptcs)
+
+        X, y = load_breast_cancer(True)
+        data = lgb.Dataset(X, label=y)
+        params = {
+            "objective": "binary",
+            "metric": "auc",
```
StrikerRUS: Leave default metric because it isn't something that we test here.

tests/python_package_test/test_basic.py
```diff
@@ -328,3 +328,47 @@ def check_asserts(data):
...
+            "min_data": 10,
+            "num_leaves": 15,
+            "verbose": -1,
+            "num_threads": 1,
```
StrikerRUS: Why is it needed here?

tests/python_package_test/test_basic.py
```diff
@@ -328,3 +328,47 @@ def check_asserts(data):
...
+                      .groupby('split_feature')['split_gain']
+                      .sum()
+                      .to_dict())
+
+        tree_split = _imptcs_to_numpy(X, split_dict)
+        tree_gains = _imptcs_to_numpy(X, gains_dict)
+        mod_split = bst.feature_importance('split')
+        mod_gains = bst.feature_importance('gain')
+        np.testing.assert_equal(tree_split, mod_split)
+        np.testing.assert_almost_equal(tree_gains, mod_gains)
```
StrikerRUS:
```suggestion
        np.testing.assert_allclose(tree_gains, mod_gains)
```

https://docs.scipy.org/doc/numpy/reference/generated/numpy.testing.assert_allclose.html#numpy.testing.assert_allclose

pford221:
Thank you for the thorough review, Striker. I will make these edits in the next day or two!

python-package/lightgbm/basic.py
pford221: Updated.

python-package/lightgbm/basic.py
pford221: Update.

python-package/lightgbm/basic.py
pford221: Updated.

python-package/lightgbm/basic.py
pford221: Changed `t` to `tree`.

python-package/lightgbm/basic.py
pford221: Done.

python-package/lightgbm/basic.py
pford221: I don't understand this suggestion. Not passing in the columns from `m_list[0].keys()` will cause the columns to be sorted alphabetically in Pandas versions prior to 0.25.0. Therefore, to ensure consistency in all versions of Pandas, we explicitly pass the order of the columns.

tests/python_package_test/test_basic.py
pford221: Removed.

tests/python_package_test/test_basic.py
pford221: It's not. I thought these hyperparameters were standard across unit tests.

tests/python_package_test/test_basic.py
pford221: Updated.

python-package/lightgbm/basic.py
pford221: Done.

python-package/lightgbm/basic.py
pford221: I've added all of these. Note `internal_weight` and `leaf_weight` both map to `weight` in the output which I assume is the sum of the hessians.

python-package/lightgbm/basic.py
pford221: Not a mistake just confusing. I'm used to calling the predictions for a particular leaf (i.e. terminal node) the leaf weight. I've changed this to be consistent with LightGBM terminology.

python-package/lightgbm/basic.py
pford221: Yes, I've done this. As mentioned above, I've also mapped both `leaf_weight` and `internal_weight` to `weight`.

python-package/lightgbm/basic.py
pford221: Changed.

python-package/lightgbm/basic.py
pford221: Thanks for the typo fix.

pford221:
Hello. I can't debug why the Travis CI build failed. It appears to be related to Too Many Requests network error. Please let me know if it's something in this PR that's causing the issue.

Thanks!

StrikerRUS:
Thank you for your updates! I'll review them soon.

Speaking about Travis, I'm sorry for making you confused with failed test. That test checks all URLs in our docs and is quite unstable due to obvious reasons. I've simply re-run it and now everything seems to be OK.

tests/python_package_test/test_basic.py
StrikerRUS: Sorry for the confusion! My comment was addressed to `"num_threads": 1`. Anyway, other params are not necessary as well, and everything seems good now.

python-package/lightgbm/basic.py
StrikerRUS: Ah, I see! But now there is no particular order at all, as `m_list[0].keys()` is a set-like collection. Maybe it should be `sorted(m_list[0].keys())` to ensure that columns are always sorted alphabetically for any Python implementation and pandas version?

python-package/lightgbm/basic.py
```diff
@@ -1853,6 +1853,119 @@ def free_network(self):
         self.network = False
         return self
 
+    def trees_to_dataframe(self):
+        """Parse the fitted model and return in an easy-to-read pandas DataFrame.
+
+        Returns
+        -------
+        DataFrame or list
```
```suggestion
        result : DataFrame or list
```

python-package/lightgbm/basic.py
```diff
@@ -1853,6 +1853,119 @@ def free_network(self):
...
+            Returns a pandas Dataframe if pandas is installed, and a list otherwise.
```
```suggestion
            Returns a pandas DataFrame if pandas is installed, and a list otherwise.
```

python-package/lightgbm/basic.py
```diff
@@ -1853,6 +1853,119 @@ def free_network(self):
...
+                node['split_gain'] = tree['split_gain']
+                node['threshold'] = tree['threshold']
+                node['decision_type'] = tree['decision_type']
+                node['missing_direction'] = 'left' if tree['default_left'] else 'right'
+                node['missing_type'] = tree['missing_type']
+                node['internal_value'] = tree['internal_value']
+                node['weight'] = tree['internal_weight']
+                node['count'] = tree['internal_count']
+            else:
+                node['leaf_value'] = tree['leaf_value']
```
StrikerRUS: Maybe merge `leaf_value` and `internal_value` into simple `value` as it was done for `count` and `weight`?

python-package/lightgbm/basic.py
```diff
@@ -1853,6 +1853,119 @@ def free_network(self):
...
+        model_list = []
+        for tree in model_dict['tree_info']:
+            model_list.extend(tree_dict_to_node_list(tree['tree_structure'],
+                                                     tree_index=tree['tree_index'],
+                                                     feature_names=feature_names))
+
+        if PANDAS_INSTALLED:
+            return DataFrame(model_list, columns=model_list[0].keys())
+        else:
+            return model_list
```
StrikerRUS: Just a question: what are the cases when this kind of output is useful? To be a source for any other dataframe-like object or what for? In other words, in what terms it's better than raw output from `dump_model()`? I'm asking because it contradicts with the method name a little bit.

python-package/lightgbm/basic.py
pford221: Hi. I can do this, but I have a couple concerns. The first is that I don't know what `internal_value` represents. I originally excluded it from the output of this method because I didn't think/know if it would be useful. Second, users will intuitively be able to interpret `leaf_value` column as the terminal node predictions for each tree, so they might be confused to find a single `value` column for every node and have trouble interpreting its meaning.

python-package/lightgbm/basic.py
pford221: This is very convenient way to look at things like maximum and minimum split gains which is useful for selecting hyperparamters or ranges of hyperparameters such as `min_gain_to_split`. You would have to write a lot of custom code to that from `dump_model()` output, but with pandas it's very easy. Also, it's a convenient way to see how much trees are being pruned, how imbalanced they are, etc. Finally, it's useful for figuring out which variables tend to have "interaction" effects which can tell you something about your data. All of those things are much more difficult when the data is in nested key:value structure like from `dump.model()`.

python-package/lightgbm/basic.py
pford221: My intention is not to have the columns be alphabetically sorted even thought that would be consistent across versions of python using `sorted()` as you suggest. I think the ordering of the columns is intuitive to how one might think of a tree's hierarchy: `tree_index`, `node_index`, ...,leaf `count`. 

A coupe options: 
1.) For users with a pre-3.6 version of Python, their column ordering will be non-deterministic. I'll defer to you as a library maintainer to judge whether that is too unappealing. 
2.) Instead of using a `dict` , we use a `collections.OrderedDict` to preserve the order of the key,value pairs. Not sure if this will have any performance or other downstream effects, but it should make sure the ordering is consistent. 
3.) I hard-code the order of the columns with a non-dynamic list. I don't love this option because if dictionary keys in the `model_list` ever change, then we'll have to remember to update this static list of column names/dictionary keys.

python-package/lightgbm/basic.py
pford221: I will update this once we agree to what other changes are needed. :)

python-package/lightgbm/basic.py
pford221: I will update this once we agree to what other changes are needed. :)

python-package/lightgbm/basic.py
StrikerRUS: Yes, I love your way of ordering too. But we should be as much consistent as it possible. So, I personally see only №2 as a good workaround. Can we try it? I don't think that there will be any significant overhead in comparison to simple dict. If pandas is compatible with OrderedDict, then everything should be OK, I hope.

python-package/lightgbm/basic.py
StrikerRUS: Thank you very much for providing a rationale of this PR, but I was referring to a case when there is no pandas installed: `return model_list`. Shouldn't we simply raise an error when pandas was not detected?

> ... but with pandas it's very easy.

python-package/lightgbm/basic.py
StrikerRUS: https://github.com/microsoft/LightGBM/blob/d7f8aa53b8be5212887e97e474c90bc5d8beac19/include/LightGBM/tree.h#L395-L396
https://github.com/microsoft/LightGBM/blob/d7f8aa53b8be5212887e97e474c90bc5d8beac19/include/LightGBM/tree.h#L432-L434
I suppose that it is `leaf_value` from previous boosting stages. So I guess it should be merged, because it represents the same thing and now confuses even more with significant number of NaN values in the corresponding columns. With `S`/`L` encoding in `node_index` field I think that users will easily find actual leaf output values.

python-package/lightgbm/basic.py
pford221: OK. I went with option #2. I'm import `OrderedDict()` at the top of basic.py. I hope that's OK.

python-package/lightgbm/basic.py
pford221: I have combined `leaf_value` and `internal_value` to be `value`. However, it seems like `internal_value` is on a different scale than `leaf_value`. Here's a screenshot of the output from the breast cancer dataset for a single tree model with `max_depth = 2`.

image

python-package/lightgbm/basic.py
pford221: Ah, I see. I agree, I don't think a list format would be useful at all. In the latest commit I'm not raising a LightGBM exception if pandas is not installed.

python-package/lightgbm/basic.py
```diff
@@ -1853,6 +1854,116 @@ def free_network(self):
         self.network = False
         return self
 
+    def trees_to_dataframe(self):
+        """Parse the fitted model and return in an easy-to-read pandas DataFrame.
+
+        Returns
+        -------
+        result: DataFrame
+            Returns a pandas DataFrame of the parsed model
```
```suggestion
            Returns a pandas DataFrame of the parsed model.
```
nit.

python-package/lightgbm/basic.py
```diff
@@ -1853,6 +1854,116 @@ def free_network(self):
...
+        """
+        if not PANDAS_INSTALLED:
+            raise LightGBMError('This method cannot be run without Pandas installed')
```
```suggestion
            raise LightGBMError('This method cannot be run without pandas installed')
```
nit.

tests/python_package_test/test_basic.py
```diff
@@ -328,3 +328,35 @@ def check_asserts(data):
...
+
+        def _imptcs_to_numpy(X, impcts_dict):
+            cols = ['Column_' + str(i) for i in range(X.shape[1])]
+            imptcs = []
+            for col in cols:
+                try:
+                    imptcs.append(impcts_dict[col])
+                except KeyError:
+                    imptcs.append(0.)
+            return np.array(imptcs)
```
StrikerRUS: nit.
Please add a blank line after `return`.

tests/python_package_test/test_basic.py
```diff
@@ -328,3 +328,35 @@ def check_asserts(data):
...
+                      .groupby('split_feature')['split_gain']
+                      .sum()
+                      .to_dict())
+
+        tree_split = _imptcs_to_numpy(X, split_dict)
+        tree_gains = _imptcs_to_numpy(X, gains_dict)
+        mod_split = bst.feature_importance('split')
+        mod_gains = bst.feature_importance('gain')
+        np.testing.assert_equal(tree_split, mod_split)
+        np.testing.assert_allclose(tree_gains, mod_gains)
```
StrikerRUS: Can we have some more tests? For instance, we can check that there are exactly 10 trees in the DataFrame.

pford221: I added two more tests. One to make sure the node count in the top-level node of each of the trees is the same length as the data and the other to ensure we have 10 trees (which makes me a bit worried as I've seen whole trees get pruned but with `min_gain_to_split` at 0 in this test, we should be fine).

StrikerRUS: Very awesome! Many thanks!

tests/python_package_test/test_basic.py
```diff
@@ -328,3 +328,42 @@ def check_asserts(data):
...
+        tree_gains = _imptcs_to_numpy(X, gains_dict)
+        mod_split = bst.feature_importance('split')
+        mod_gains = bst.feature_importance('gain')
+        num_trees_from_df = tree_df.tree_index.nunique()
+        obs_counts_from_df = tree_df.loc[tree_df.node_depth == 1, 'count'].values
+
+        np.testing.assert_equal(tree_split, mod_split)
+        np.testing.assert_allclose(tree_gains, mod_gains)
+        np.testing.assert_equal(num_trees_from_df, num_trees)
+        np.testing.assert_equal(obs_counts_from_df, len(y))
```
StrikerRUS: nit.
It's better to use `self.assertEqual()` for less overhead.

StrikerRUS:
For auto-closing mechanism: fixed #2578. Also I think it can be treated as a fix for #2320, because with this PR it would be possible to obtain actual values as following: `actual_min_gain_to_split = model.trees_to_dataframe()['split_gain'].min()`

tests/python_package_test/test_basic.py
```diff
@@ -328,3 +328,42 @@ def check_asserts(data):
...
+        gains_dict = (tree_df
+                      .groupby('split_feature')['split_gain']
+                      .sum()
+                      .to_dict())
+
+        tree_split = _imptcs_to_numpy(X, split_dict)
+        tree_gains = _imptcs_to_numpy(X, gains_dict)
+        mod_split = bst.feature_importance('split')
+        mod_gains = bst.feature_importance('gain')
+        num_trees_from_df = tree_df.tree_index.nunique()
```
jameslamb: could you change this to `tree_df['tree_index']` and replace all other `df.{column_name}` type references in a similar way?

This is safer and more future-proof, since `pandas` will not prevent you from creating a column whose name conflicts with a public attribute of `pd.DataFrame` but will prefer the attribute to the column when there is a conflict.

example:

```python
import pandas as pd
df = pd.DataFrame({'index': [1,2,3]})
df.index

# >>> RangeIndex(start=0, stop=3, step=1)
```

pford221: Sure. I just made this change and pushed.

StrikerRUS:
@jameslamb Were your comments fully addressed?

jameslamb:
> @jameslamb Were your comments fully addressed?

ah, totally missed the followup comment and commit in my GitHub emails!  Yes they were, thank you for the fix @pford221

tests/python_package_test/test_basic.py
jameslamb: Looks great, thank you! I'll resolve this thread

StrikerRUS: @jameslamb Please update your review status as well. Your requested changes block merging:
> Merging is blocked
Merging can be performed automatically once the requested changes are addressed.

python-package/lightgbm/basic.py
```diff
@@ -1853,6 +1854,116 @@ def free_network(self):
         self.network = False
         return self
 
+    def trees_to_dataframe(self):
+        """Parse the fitted model and return in an easy-to-read pandas DataFrame.
+
+        Returns
+        -------
+        result: DataFrame
```
```suggestion
        result : pandas DataFrame
```

python-package/lightgbm/basic.py
```diff
@@ -1853,6 +1854,116 @@ def free_network(self):
         self.network = False
         return self
 
+    def trees_to_dataframe(self):
```
StrikerRUS: @pford221 I've just found a bug, please fix the following edge case:
```
import numpy as np
import lightgbm as lgb

X = np.ones((10, 2))
y = np.random.rand(10)
lgb_data = lgb.Dataset(X, y)

bst = lgb.train({}, lgb_data, num_boost_round=3)
bst.trees_to_dataframe()
```

```
KeyError                                  Traceback (most recent call last)
<ipython-input-5-86ad7722a0da> in <module>
----> 1 bst.trees_to_dataframe()

D:\Miniconda3\lib\site-packages\lightgbm\basic.py in trees_to_dataframe(self)
   1961             model_list.extend(tree_dict_to_node_list(tree['tree_structure'],
   1962                                                      tree_index=tree['tree_index'],
-> 1963                                                      feature_names=feature_names))
   1964 
   1965         return DataFrame(model_list, columns=model_list[0].keys())

D:\Miniconda3\lib\site-packages\lightgbm\basic.py in tree_dict_to_node_list(tree, node_depth, tree_index, feature_names, parent_node)
   1936                                       tree_index=tree_index,
   1937                                       feature_names=feature_names,
-> 1938                                       parent_node=parent_node)
   1939 
   1940             res = [node]

D:\Miniconda3\lib\site-packages\lightgbm\basic.py in create_node_record(tree, node_depth, tree_index, feature_names, parent_node)
   1896             node['tree_index'] = tree_index
   1897             node['node_depth'] = node_depth
-> 1898             node['node_index'] = _get_node_index(tree, tree_index)
   1899             node['left_child'] = None
   1900             node['right_child'] = None

D:\Miniconda3\lib\site-packages\lightgbm\basic.py in _get_node_index(tree, tree_index)
   1879                 is_split = _is_split_node(tree)
   1880                 node_type = 'S' if is_split else 'L'
-> 1881                 node_num = str(tree['split_index' if is_split else 'leaf_index'])
   1882                 return tree_num + node_type + node_num
   1883 

KeyError: 'leaf_index'
```

tests/python_package_test/test_basic.py
```diff
@@ -328,3 +328,42 @@ def check_asserts(data):
...
+    def test_trees_to_dataframe(self):
+
+        def _imptcs_to_numpy(X, impcts_dict):
+            cols = ['Column_' + str(i) for i in range(X.shape[1])]
+            imptcs = []
+            for col in cols:
+                try:
+                    imptcs.append(impcts_dict[col])
+                except KeyError:
+                    imptcs.append(0.)
```
StrikerRUS: I think it can be replaced with one-line solution `imptcs.append(impcts_dict.get(col, 0.))`

python-package/lightgbm/basic.py
StrikerRUS: I'm sorry, this comment had to go to the previous my review but GitHub lagged and it hadn't been appended there.

python-package/lightgbm/basic.py
pford221: Done in most recent (6th) commit.

python-package/lightgbm/basic.py
pford221: OK. This has been handled through a `try...expect` clause. Also, since a single node tree doesn't have a `count` or `weight` keys, I have to deal with this as well. I wrote a robust function to check if each leaf node comes from a single node tree.

tests/python_package_test/test_basic.py
pford221: Changed in the final (6th) commit.

python-package/lightgbm/basic.py
```diff
@@ -1853,6 +1854,122 @@ def free_network(self):
...
+                        feature_name = feature_names[tree['split_feature']]
+                    else:
+                        feature_name = tree['split_feature']
+                else:
+                    feature_name = None
+                return feature_name
+
+            def _is_single_node_tree(tree):
+                tree_keys = [str(key) for key in tree.keys()]
+                return tree_keys == ['leaf_value']
```
StrikerRUS: I guess comparing sets directly will be more efficient and requires less code:
```
tree.keys() == {'leaf_value'}
```
And `keys` are already strings, so there is no need to cast them.

tests/python_package_test/test_basic.py
```diff
@@ -328,3 +328,39 @@ def check_asserts(data):
         lgb_data.set_weight(sequence)
         lgb_data.set_init_score(sequence)
         check_asserts(lgb_data)
+
+    def test_trees_to_dataframe(self):
+
+        def _imptcs_to_numpy(X, impcts_dict):
+            cols = ['Column_' + str(i) for i in range(X.shape[1])]
+            imptcs = []
+            for col in cols:
```
StrikerRUS: Now after removing `try`/`except` logic, it can be rewritten with list comprehension for better efficiency and readability:
```
imptcs = [impcts_dict.get(col, 0.) for col in cols]
```

python-package/lightgbm/basic.py
```diff
@@ -1853,6 +1854,122 @@ def free_network(self):
         self.network = False
         return self
 
+    def trees_to_dataframe(self):
+        """Parse the fitted model and return in an easy-to-read pandas DataFrame.
+
+        Returns
+        -------
+        result: pandas DataFrame
```
StrikerRUS: Please fix formatting issue [I mentioned in the previous review](https://github.com/microsoft/LightGBM/pull/2592#discussion_r364430743):
```suggestion
        result : pandas DataFrame
```

tests/python_package_test/test_basic.py
pford221: Got it.

python-package/lightgbm/basic.py
pford221: Thanks for the suggestion. I didn't know that the keys of a dictionary as a collection was a set!

Issue #1493, opened 2018-07-05
Capture LGBM model's .fit messages and redirect them to python logger


MTDzi:
I'm using the `logging` Python module in my project and would like to be able to capture the messages produced by the `lightgbm.sklearn.LGBMClassifier.fit` method.

So far, I've re-directed the `STDOUT` to the logging file:
```python
logging.basicConfig(filename=log_filename, level=logging.INFO)
channel = logging.StreamHandler(sys.stdout)
sys.stdout = open(log_filename, 'a', 1)
channel.setLevel(logging.INFO)
root = logging.getLogger()
root.setLevel(logging.INFO)
root.addHandler(channel)
```

but in doing so, I'm also re-directing output from the debugger -- not good.

How can I do it better?

StrikerRUS:
At present practically all messages (except a few warnings) come directly from the C++ code to the stdout/stderr. So, there is no way to manage them by Python's `setLevel()` functionality. As a workaround you can write a function that will filter messages you don't want to write into a log and re-redirect them back to the console. The code of C++ logger may help you: https://github.com/Microsoft/LightGBM/blob/master/include/LightGBM/utils/log.h

StrikerRUS:
@guolinke @chivee 
Maybe we can have something similar?
https://github.com/dmlc/xgboost/pull/3438

MTDzi:
Hey,

Thanks for the reply!

As a work-around (because I was interested in the messages related to early stopping), I just took the `LGBMClassifier.evals_result_` and passed that as a message to the logger.

But, for anyone else interested in the matter, **I highly discourage from using my previous work-around, the one with the** 
```
...
sys.stdout = open(log_filename, 'a', 1)
...
```
**hack.**

StrikerRUS:
@MTDzi Thank you for sharing your experience!

StrikerRUS:
@guolinke @chivee 
Is it possible to implement https://github.com/Microsoft/LightGBM/issues/1493#issuecomment-402839772 in LightGBM?
Such thing will allow, for example, to log right into Jupyter Notebook cell output.

StrikerRUS:
ping @guolinke @chivee

guolinke:
@StrikerRUS 
Only need to redirect the output in training, without all other logs?

StrikerRUS:
@guolinke I mean, create a callback at cpp side, which can be used at Python/R/whatever_language side:
[...]

PR #2594, opened 2019-11-26
[python] [R-package] refine the parameters for Dataset

### LINKED ISSUE #2590: reset_param feature_contrib not fixed
this bug is mentioned and fixed below,but found not really fixed. pls check again
[#2452](https://github.com/microsoft/LightGBM/issues/2452) 
[#2453](https://github.com/microsoft/LightGBM/issues/2453)

guolinke:
to fix #2590

- [x] remove the pre-filter by `min_data`


guolinke:
@StrikerRUS could you help to add the tests, to ensure the warnings are shown expected?

StrikerRUS:
@guolinke Sure, I'll try!

src/c_api.cpp
```diff
@@ -185,8 +185,44 @@ class Booster {
     if (param.count("metric")) {
       Log::Fatal("Cannot change metric during training");
     }
-
+    auto old_config = config_;
     config_.Set(param);
+
+    if (param.count("max_bin") && config_.max_bin != old_config.max_bin) {
```
StrikerRUS: @guolinke Can't `LGBM_DatasetUpdateParamWarning` be called here to avoid code duplication?

StrikerRUS:
@guolinke 
> could you help to add the tests, to ensure the warnings are shown as expected?

Without implementing #1493, it was not so straightforward as wanted, but seems that with the help of [link1](https://root-forum.cern.ch/t/how-to-redirect-the-output-of-function-printf-in-pyroot/9915) and [link2](https://eli.thegreenplace.net/2015/redirecting-all-kinds-of-stdout-in-python/) test works fine. Please check my batch of commits.

src/c_api.cpp
```diff
@@ -173,6 +173,46 @@ class Booster {
     }
   }
 
+  static void CheckDatasetResetConfig(const Config& old_config, const std::unordered_map<std::string, std::string>& new_param) {
```
StrikerRUS: Maybe add the following params for the completeness? 
```
pre_partition
enable_bundle
max_conflict_rate
is_enable_sparse
```

src/c_api.cpp
```diff
@@ -173,6 +173,46 @@ class Booster {
...
+    Config new_config;
+    new_config.Set(new_param);
+    if (new_param.count("max_bin") && new_config.max_bin != old_config.max_bin) {
+      Log::Warning("Cannot change max_bin after constructed Dataset handle.");
```
StrikerRUS: Not sure, but maybe `Fatal` is better here?.. I think it will help to avoid any confusion about why changing params didn't take effect. I mean, it seems to me that the most users either set `verbose=-1`, or use Jupyter Notebook where LightGBM's log is printed in Jupyter console, and in consequence simply cannot see any warning.

StrikerRUS:
Should we mark our next release as `2.4.0` due to the removal of `get_monotone_constraints()` and `get_feature_penalty()`?

guolinke:
Thanks @StrikerRUS .
In Dataset.update_params, when raw_data is not freed, I think we could still update it. could you help to test the behavior too?

python-package/lightgbm/basic.py
```diff
@@ -1691,6 +1662,9 @@ def __init__(self, params=None, train_set=None, model_file=None, model_str=None,
             if not isinstance(train_set, Dataset):
                 raise TypeError('Training data should be Dataset instance, met {}'
                                 .format(type(train_set).__name__))
+            train_set.construct()
```
StrikerRUS: Shouldn't network initialization go before dataset construction?

StrikerRUS:
@guolinke 
> In Dataset.update_params, when raw_data is not freed, I think we could still update it. could you help to test the behavior too?

Sure, I'll update the test!
Can you please take a look at 4 other failing tests? Are failures expected there?

guolinke:
@StrikerRUS I think the error is fixed. 
@jameslamb could you help to review the R's code? I almost forget how to write R's code...

R-package/R/lgb.Dataset.R
```diff
@@ -533,22 +533,47 @@ Dataset <- R6::R6Class(
...
-          , ret = NULL
-          , private$handle
+      if (lgb.is.null.handle(private$handle)) {
+        private$params <- modifyList(private$params, params)
+      } else {
+        call_state <- .Call(
+          "LGBM_DatasetUpdateParamWarning_R"
+          , lgb.params2str(private$params)
           , lgb.params2str(params)
+          , call_state
```
jameslamb: where is this `call_state` coming from? It's not being passed into `update_params()`

R-package/R/lgb.Dataset.R
```diff
@@ -533,22 +533,47 @@ Dataset <- R6::R6Class(
...
-      private$params <- modifyList(private$params, params)
       return(invisible(self))
 
     },
 
+    get_params = function() {
+      keys <- c("has_header", "header", "max_bin", "max_bin_by_feature", "bin_construct_sample_cnt", "subsample_for_bin", "min_data_in_bin", "use_missing",
+        "zero_as_missing", "sparse_threshold", "categorical_feature", "cat_feature", "categorical_column", "cat_column", 
+        "feature_pre_filter", "pre_partition", "is_pre_partition", "enable_bundle", "is_enable_bundle", "bundle", "max_conflict_rate ",
+        "is_enable_sparse", "is_sparse", "enable_sparse", "sparse", "forcedbins_filename")
```
jameslamb: could you make this an alphabetized list with one key per line? I think with lists that are this large, it's easy to make mistakes (both now and in future PRs). So like this

```
keys <- c(
    "bin_construct_sample_cnt"
    ...
    , "zero_as_missing"
)
```

R-package/R/lgb.Dataset.R
```diff
@@ -533,22 +533,47 @@ Dataset <- R6::R6Class(
...
+      ret <- list()
+      for (key in names(private$params)) {
```
jameslamb: could you name this iteration variable `param_key` or something? `key()` is a function exported from `data.table` (one of our dependencies) so we should avoid that name to prevent any weird things happening

R-package/R/utils.R
```diff
@@ -19,6 +19,36 @@ lgb.encode.char <- function(arr, len) {
 
 }
 
+lgb.last_error <- function() {
```
jameslamb: I like this change! I'm surprised we didn't have a function like this before, actually.

R-package/R/lgb.Dataset.R
```diff
@@ -533,22 +533,47 @@ Dataset <- R6::R6Class(
...
+          self$finalize()
+        }
       }
-      private$params <- modifyList(private$params, params)
       return(invisible(self))
 
     },
 
+    get_params = function() {
+      keys <- c("has_header", "header", "max_bin", "max_bin_by_feature", "bin_construct_sample_cnt", "subsample_for_bin", "min_data_in_bin", "use_missing",
```
jameslamb: what is the purpose of using this list instead of just returning `private$params` directly?

tests/python_package_test/test_basic.py
```diff
@@ -328,3 +281,77 @@ def check_asserts(data):
         lgb_data.set_weight(sequence)
         lgb_data.set_init_score(sequence)
         check_asserts(lgb_data)
+
+    def test_dataset_update_params_warning(self):
```
jameslamb: could you either implement this same test on the R side or create an issue (to add to #2302 ) documenting the need for such a test on the R side?

R-package/R/lgb.Dataset.R
guolinke: There are some failed tests in python side with returning private$params only. 

this is to write the params from Dataset to Booster. As parameters may is changes in Booster side, we only return the params of Dataset actually used, otherwise it may write some old parameters back.

python-package/lightgbm/basic.py
```diff
@@ -761,22 +750,40 @@ def __init__(self, data, label=None, reference=None,
...
     def __del__(self):
         try:
             self._free_handle()
         except AttributeError:
             pass
 
+    def _filter_params(self, params):
+        if params is None:
+            return params
+        keys = ['has_header', 'header', 'max_bin', 'max_bin_by_feature', 'bin_construct_sample_cnt', 'subsample_for_bin', 'min_data_in_bin', 'use_missing',\
```
StrikerRUS: Is this [IO params](https://lightgbm.readthedocs.io/en/latest/Parameters.html#io-parameters) + `forcedbins_filename` + `min_data_in_leaf`? Then a lot of params are missed, e.g. `histogram_pool_size`, `two_round`, etc.

python-package/lightgbm/basic.py
```diff
@@ -1691,6 +1683,9 @@ def __init__(self, params=None, train_set=None, model_file=None, model_str=None,
             if not isinstance(train_set, Dataset):
                 raise TypeError('Training data should be Dataset instance, met {}'
                                 .format(type(train_set).__name__))
+            train_set.construct()
+            # copy the parameters from train_set
+            params.update(train_set.get_params())
```
StrikerRUS: Can't 
```
updated_params = train_set.params
updated_params.update(params)
```
(make Booster's params more important than Dataset's ones) help to get rid of `get_params()` method?

include/LightGBM/c_api.h
```diff
@@ -341,12 +340,13 @@ LIGHTGBM_C_EXPORT int LGBM_DatasetGetField(DatasetHandle handle,
...
- * \param handle Handle of dataset
- * \param parameters Parameters
+ * \brief Raise warnings for attempts to update dataset parameters.
+ * \param old_parameters Current dataset parameters
+ * \param new_parameters New dataset parameters
+ * \return 0 when succeed, -1 when failure happens
  */
-LIGHTGBM_C_EXPORT int LGBM_DatasetUpdateParam(DatasetHandle handle,
-                                              const char* parameters);
+LIGHTGBM_C_EXPORT int LGBM_DatasetUpdateParamWarning(const char* old_parameters,
```
StrikerRUS: Update name (not warning anymore, but fatal)?

python-package/lightgbm/basic.py
guolinke: Indeed, this can save many codes

guolinke:
@StrikerRUS 
it seems this cannot work, the dataset will store some old parameters.
```
self = <test_engine.TestEngine testMethod=test_cv>

    def test_cv(self):
        X_train, y_train = load_boston(True)
        params = {'verbose': -1}
        lgb_train = lgb.Dataset(X_train, y_train)
        # shuffle = False, override metric in params
        params_with_metric = {'metric': 'l2', 'verbose': -1}
        cv_res = lgb.cv(params_with_metric, lgb_train, num_boost_round=10,
                        nfold=3, stratified=False, shuffle=False,
                        metrics='l1', verbose_eval=False)
        self.assertIn('l1-mean', cv_res)
        self.assertNotIn('l2-mean', cv_res)
        self.assertEqual(len(cv_res['l1-mean']), 10)
        # shuffle = True, callbacks
        cv_res = lgb.cv(params, lgb_train, num_boost_round=10, nfold=3, stratified=False, shuffle=True,
                        metrics='l1', verbose_eval=False,
                        callbacks=[lgb.reset_parameter(learning_rate=lambda i: 0.1 - 0.001 * i)])
        self.assertIn('l1-mean', cv_res)
        self.assertEqual(len(cv_res['l1-mean']), 10)
        # enable display training loss
        cv_res = lgb.cv(params_with_metric, lgb_train, num_boost_round=10,
                        nfold=3, stratified=False, shuffle=False,
                        metrics='l1', verbose_eval=False, eval_train_metric=True)
        self.assertIn('train l1-mean', cv_res)
        self.assertIn('valid l1-mean', cv_res)
        self.assertNotIn('train l2-mean', cv_res)
        self.assertNotIn('valid l2-mean', cv_res)
        self.assertEqual(len(cv_res['train l1-mean']), 10)
        self.assertEqual(len(cv_res['valid l1-mean']), 10)
        # self defined folds
        tss = TimeSeriesSplit(3)
        folds = tss.split(X_train)
        cv_res_gen = lgb.cv(params_with_metric, lgb_train, num_boost_round=10, folds=folds,
                            verbose_eval=False)
        cv_res_obj = lgb.cv(params_with_metric, lgb_train, num_boost_round=10, folds=tss,
                            verbose_eval=False)
        np.testing.assert_allclose(cv_res_gen['l2-mean'], cv_res_obj['l2-mean'])
        # lambdarank
        X_train, y_train = load_svmlight_file(os.path.join(os.path.dirname(os.path.realpath(__file__)),
                                                           '../../examples/lambdarank/rank.train'))
        q_train = np.loadtxt(os.path.join(os.path.dirname(os.path.realpath(__file__)),
                                          '../../examples/lambdarank/rank.train.query'))
        params_lambdarank = {'objective': 'lambdarank', 'verbose': -1, 'eval_at': 3}
        lgb_train = lgb.Dataset(X_train, y_train, group=q_train)
        # ... with l2 metric
        cv_res_lambda = lgb.cv(params_lambdarank, lgb_train, num_boost_round=10, nfold=3,
                               metrics='l2', verbose_eval=False)
        self.assertEqual(len(cv_res_lambda), 2)
        self.assertFalse(np.isnan(cv_res_lambda['l2-mean']).any())
        # ... with NDCG (default) metric
        cv_res_lambda = lgb.cv(params_lambdarank, lgb_train, num_boost_round=10, nfold=3,
                               verbose_eval=False)
        self.assertEqual(len(cv_res_lambda), 2)
>       self.assertFalse(np.isnan(cv_res_lambda['ndcg@3-mean']).any())
E       KeyError: 'ndcg@3-mean'
```

python-package/lightgbm/basic.py
guolinke: I think histogram_pool_size is not related to the dataset loading

guolinke:
@jameslamb @StrikerRUS it seems the check the docs test will stop when met one error. Is that possible to throw all errors then stop?

python-package/lightgbm/basic.py
StrikerRUS: @guolinke 
OK, I see.
What about `forcedbins_filename`? It seems to me that it can be moved under **IO** section.

Also, I'm worrying about `seed` (`data_seed`) and `nthreads` parameters. Will they work correctly after this PR?

python-package/lightgbm/basic.py
StrikerRUS: @guolinke  ping

StrikerRUS:
@guolinke 
> it seems this cannot work, the dataset will store some old parameters.

Was cv test the only one which failed? I remember we had exactly the same error when our `train` and `cv` functions were not pure, they used to modify mutable `params` dict in the past. Maybe we can investigate that case? Because I personally don't think that I like we need to maintain so big params list at language wrapper sides. I would really want to move as much logic to cpp side as possible and keep wrappers as thin as they can be.

> it seems the check the docs test will stop when met one error. Is that possible to throw all errors then stop?

Do you mean `docs` or `docstring` test? Because `docs` test shows all errors and warnings before it fails.


Have I added all test cases that you wanted to cover? @jameslamb Can you please help with R test?

guolinke:
I remember there are many errors.  
For example, the dataset could be used in the first lgb.train with metric parameter. Then using it in the second lgb.train witiout metric, the metric will be override while we expect the default metric here.

StrikerRUS:
@guolinke In the latest commits I
- refactored Python code to use alias managing;
- updated test;
- fixed two important typos in code.

python-package/lightgbm/basic.py
StrikerRUS: ... as well as about `initscore_filename` and `valid_data_initscores` params.

StrikerRUS: > I think histogram_pool_size is not related to the dataset loading

Can we have a subsection under `IO` params named `Dataset` (or somehow better) params? Yes, we have no `Booster`/`Dataset` abstraction at CLI level, but with this PR distinguishing becomes very important. Without such info, users have to understand where are Dataset params only via fatal errors they get.


Or maybe we should add a **Note** to each Dataset parameter about that it cannot be changed after dataset construction.

WDYT?

python-package/lightgbm/basic.py
guolinke: sorry, will fix it

StrikerRUS:
@guolinke Just out of curiosity, how hard will it be to create a C API function which returns all param names with their aliases? For example, it can be a string format like `param1:param1_alias1,param1_alias2;param2:param2_alias1;param3;param4:...`
It will allow to deduplicate aliases handling code in language wrappers and create param aliases handling mechanism by simple string parsing got from C API.

guolinke:
@StrikerRUS that will be cool, it could be done by the config code automatically generating. 
As for the objective and metric, we may need to add them manually.

StrikerRUS:
@guolinke 
> that will be cool, it could be done by the config code automatically generating.
As for the objective and metric, we may need to add them manually.

Don't you mind I add this to our feature requests? Or do you want to implement it within this PR?

python-package/lightgbm/basic.py
StrikerRUS: @guolinke Can you please take a look at this thread? I think that is all things I'm worrying about within this PR.

guolinke:
@StrikerRUS I think it is better to have another PR for that.

StrikerRUS:
@guolinke 
>  I think it is better to have another PR for that.

OK, done: #2633!

jameslamb:
> Have I added all test cases that you wanted to cover? @jameslamb Can you please help with R test?

@guolinke could you take a look at #2530 ? @StrikerRUS has given me a really thorough review and the two of us have talked about it a lot. It is building correctly and we are in agreement, but need one more set of eyes from either you or @Laurae2 

If we get that merged, you could rebase this branch to `master` and that would run all the R tests.

I can also push some R tests here if you'd like.

jameslamb:
Now that #2530 has been merged, if you rebase this to `master` you will get all of the R tests to see if anything needs to be fixed.

StrikerRUS:
@guolinke Do you want to add anything else in this PR?

guolinke:
Do we need tests for R in this PR?

StrikerRUS:
@guolinke I hope that after resolving remaining questions (see https://github.com/microsoft/LightGBM/pull/2594#discussion_r355011770), @jameslamb will help utilize aliases mechanism and write test for R.

jameslamb:
> @guolinke I hope that after resolving remaining questions (see [#2594 (comment)](https://github.com/microsoft/LightGBM/pull/2594#discussion_r355011770)), @jameslamb will help utilize aliases mechanism and write test for R.

yep! Would be happy to add the tests once that thread is resolved

python-package/lightgbm/basic.py
guolinke: sorry, I didn't see this thread ...
I agree with the subsection for dataset object, it will be much clearer.

guolinke: However, how about changing it in another PR? as it is related to documentation.

StrikerRUS: @guolinke Thank you for your answers!

> However, how about changing it in another PR? as it is related to documentation.

Agree! I will help with docs in a separate PR.

Can you please answer to one remained my question?

> Also, I'm worrying about `seed` (`data_seed`) and `nthreads` parameters. Will they work correctly after this PR?

> ... as well as about `initscore_filename` and `valid_data_initscores` params.

guolinke: I think `nthreads` is safe for it doesn't affect the result of dataset. 
For `seed`, I think only consider `data_seed` is enough.
For `init_scores`, I think there are not loaded by python/R package, as we set it to `""` in the code:
https://github.com/microsoft/LightGBM/blob/509c2e50c25eded99fc0997afe25ebee1b33285d/src/c_api.cpp#L481

StrikerRUS: > I think nthreads is safe for it doesn't affect the result of dataset.

Hmm, but it is used in many places during creating of dataset, e.g.
https://github.com/microsoft/LightGBM/blob/509c2e50c25eded99fc0997afe25ebee1b33285d/src/io/sparse_bin.hpp#L82
https://github.com/microsoft/LightGBM/blob/dbb804f2350f280f3c927f792f7e55ac9c021707/src/io/dataset.cpp#L662
I suppose I'm missing something...

> For seed, I think only consider data_seed is enough.

`seed` will cause changing of `data_seed` (except the case when `rand.NextShort()` will choose the same value as it was) and in consequence trigger the error, right?

> For init_scores, I think there are not loaded by python/R package, as we set it to "" in the code:

Oh, then our docs are incorrect:
https://github.com/microsoft/LightGBM/blob/509c2e50c25eded99fc0997afe25ebee1b33285d/include/LightGBM/config.h#L534-L535
https://github.com/microsoft/LightGBM/blob/509c2e50c25eded99fc0997afe25ebee1b33285d/include/LightGBM/config.h#L542-L543

guolinke: for `num_threads`, there are some temporary buffer, to speed-up the data loading by multi-threading. with different num_threads, the constructed dataset should be identical. As for `hist_buf_`, it is used when tree learning, and it is a temporary buffer too.

For seed, yes, changing `seed` will also changing `data_seed`, and the error will be raised too.

StrikerRUS: Thank you very much for the detailed explanation!

Is it possible to update code for `init_scores` to match described behavior in docs?

guolinke: unfortunately, It is not trivial. As we don't know the filename passed in c_api is training or k-th valid data. It is hard to retrieve the corresponding init score filename. The easier way in python/R is read the init_score_file by user, and set `init_score` field in dataset.

guolinke: another solution is to force the init_score_filename to `concat(data_filename, ".init"), and remove all initscore filenames parameters.

guolinke:
@StrikerRUS I think we can move the discussion to the main thread. it is hard to notice the new conversion in the above sub-thread.

StrikerRUS:
@guolinke  
> unfortunately, It is not trivial. As we don't know the filename passed in c_api is training or k-th valid data. It is hard to retrieve the corresponding init score filename. The easier way in python/R is read the init_score_file by user, and set init_score field in dataset.

> another solution is to force the init_score_filename to concat(data_filename, ".init"), and remove all initscore filename parameters.

I think removing filename params is good idea! It will decrease overall number of LightGBM params which is good change for newcomers who struggling with learning a huge number of settings. Also, it will fix inconsistency in behavior when for language packages default filename is respected but it cannot be changed by those params. As we've already removed some sparse-related params in #2699 and minor (or major) version bump is required, it is OK in terms of backward compatibility.

guolinke:
yeah, maybe do it in a new PR?

StrikerRUS:
@guolinke 
> yeah, maybe do it in a new PR?

Sure, as you wish!

guolinke:
ping @jameslamb for R tests.
@StrikerRUS any other issues need to be addressed?

guolinke:
ping @jameslamb for R's part. Maybe R test could be in a new PR?

jameslamb:
> ping @jameslamb for R's part. Maybe R test could be in a new PR?

just  added #2767 . Let me know if you think anything is missing!

PR #2600, opened 2019-11-29
LightGBM would hang on MPI run if only some nodes have a fatal error.

ashok-ponnuswami-msft:
The issue is that the destructor of Application calls MPI_Finalize(), which will cause the program to hand and prevent from exiting. As a solution, we remove MPI_Finalize() from the destructor and let main() decide whether to call MPI_Finalize() or MPI_Abort() based on the reason it is ending.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -3,26 +3,41 @@
  * Licensed under the MIT License. See LICENSE file in the project root for license information.
  */
 #include <LightGBM/application.h>
+#include "network/linkers.h"
 
 #include <iostream>
 
+
+
 int main(int argc, char** argv) {
+  bool success = false;
   try {
     LightGBM::Application app(argc, argv);
     app.Run();
+
+#ifdef USE_MPI
+    LightGBM::Linkers::MpiFinalizeIfIsParallel();
+#endif
+
+    success = true;
   }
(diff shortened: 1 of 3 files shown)


src/main.cpp
```diff
@@ -3,26 +3,39 @@
...
 #include <iostream>
 
 int main(int argc, char** argv) {
+  bool success = false;
   try {
     LightGBM::Application app(argc, argv);
     app.Run();
+
+#ifdef USE_MPI
+    LightGBM::Linkers::MpiFinalizeIfIsParallel();
```
guolinke: just to confirm, will this be blocked when some nodes are failed?

ashok-ponnuswami-msft: You mean what happens when some nodes call MPI_Finalize(), but some nodes call MPI_Abort()? It appears that in that case, the nodes that MPI_Finalize() don't fail and can complete their execution, but the nodes that call MPI_Abort() will exit with non-zero exit code. So none of them hang. In another scenario, if during the phase when the nodes are still communicating with each other (like when loading data or training is still in progress), some nodes want to keep communicating while some others call MPI_Abort(), all processes exit with non-zero error code. Previously all processes used to hang in this scenario because the failing nodes were calling MPI_Finalize().

PR #2606, opened 2019-12-02
[python][R-package] removed duplicated code from language wrappers

StrikerRUS:
Removed checks from language wrappers that are handled at cpp side.

https://github.com/microsoft/LightGBM/blob/bbc45fed573b246556055e4d4dc335b92442d603/src/c_api.cpp#L176-L187


StrikerRUS:
@guolinke According to this commit, seems that `metric` cannot be reset anymore, right?
https://github.com/microsoft/LightGBM/commit/714c673257b4f2aefabd8602eace13b968d7ed3f
So, this code is outdated too.

StrikerRUS:
@guolinke Can you please confirm this https://github.com/microsoft/LightGBM/pull/2606#issuecomment-560501048?

guolinke:
@StrikerRUS it seems you are right. but I don't remember why `__need_reload_eval_info` is there...

StrikerRUS:
@guolinke 
> it seems you are right. but I don't remember why __need_reload_eval_info is there...

Maybe in some point in the past it was possible to reset `metric` param? `__need_reload_eval_info` is used only once here:
https://github.com/microsoft/LightGBM/blob/f3afe98bedd469b04758024c93c9ad77a43dc195/python-package/lightgbm/basic.py#L2691-L2694

Anyway, as at present resetting `metric` leads to fatal error, keeping that code seems to be unreasonable.

StrikerRUS:
Thanks @jameslamb !

I'm about to merge this PR, but want to make sure that @guolinke hasn't accidentally remembered why `self.__need_reload_eval_info = True` was here.

guolinke:
@StrikerRUS I think you can remove the `__need_reload_eval_info `

Issue #1818, opened 2018-11-02
MinGW fails with access violation writing error


StrikerRUS:
Just moving my [comment](https://github.com/Microsoft/LightGBM/pull/1660#issue-214413278) from old PR to a separate issue for the discussion and easier searching for users.

While working on that PR, I found weird bug under MinGW. The same symptoms were described in #1475. When I tried to run **all** examples by `pytest %APPVEYOR_BUILD_FOLDER%\tests` instead of `pytest %APPVEYOR_BUILD_FOLDER%\tests\python_package_test`, two C_API tests were passed succesfully, but all other tests failed with `OSError: exception: access violation writing 0xFFFFFFFF93400000`. At present this bug is avoided by running only Python tests (without C_API). I suppose, it's somehow connected to that the same `lib_lightgbm.dll` is loaded twice from different places (from installation directory for C_API tests and from python-package directory for Python tests). 

## Environment info

Operating System: Windows

Compiler: any MinGW-w64

## Error message

```
E       OSError: exception: access violation writing 0xFFFFFFFF93160000
```

Full log: https://ci.appveyor.com/project/guolinke/lightgbm/builds/20010549/job/ewn044xw9dja8g9y

## Steps to reproduce

1. Modify `.appveyor.yml` file:
```diff
test_script:
- - pytest %APPVEYOR_BUILD_FOLDER%\tests\python_package_test
+ - pytest %APPVEYOR_BUILD_FOLDER%\tests
```

guolinke:
@StrikerRUS I am not sure about the root cause, maybe some build flags for mingw are wrong ?

StrikerRUS:
@guolinke To be honest, I'm not familiar with this, I remember only I added `static` linking to eliminate [confusion](https://github.com/dmlc/xgboost/issues/977) with `conda`'s libraries:

https://github.com/Microsoft/LightGBM/blob/a0efb07bb8ef9d8a80f246a933c623e253a8a0a3/CMakeLists.txt#L119-L121

StrikerRUS:
All tests are passed when `-O3` flag is removed.
```diff
diff --git a/.appveyor.yml b/.appveyor.yml
index 6ea6da0..6f8daf2 100644
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -42,7 +42,7 @@ build_script:
     python setup.py install)

 test_script:
-  - pytest %APPVEYOR_BUILD_FOLDER%\tests\python_package_test
[...]

PR #2620, opened 2019-12-07
Implementation of XE_NDCG_MART for the ranking task

sbruch:
This Pull Request implements the recently proposed ranking loss (paper at https://arxiv.org/abs/1911.09798), a variant of the cross-entropy loss that is a convex bound on (shifted and scaled) NDCG. Models trained with this loss perform as well as or better than LambdaMART and, more importantly, exhibit higher robustness to noise in the training data.


sbruch:
I don't believe the one test case that failed is because of a change in this PR (https://travis-ci.org/microsoft/LightGBM/jobs/621876769?utm_medium=notification&utm_source=github_status)

StrikerRUS:
@sbruch 
> I don't believe the one test case that failed is because of a change in this PR

Thank you very much for your PR and sorry about that! It was failing due to server temporary problems. I've re-run it - now everything is OK.

sbruch:
@StrikerRUS - Thank you!

sbruch:
Hi - I was just wondering if I could get initial feedback on whether you accept PRs (that propose a new algorithm) like this one at all?

StrikerRUS:
@sbruch We are so sorry for the delay! But we are lacking of resources and busy with some bug fixes (bug fixes have higher priority than enhancements). We'll definitely review your PR! And of course we are welcome such kinds of PRs (you may take a look at our [feature requests hub](https://github.com/microsoft/LightGBM/issues/2302) or recently merged [PR with AUC-mu implementation](https://github.com/microsoft/LightGBM/pull/2567) if you have any doubts).

sbruch:
@StrikerRUS Oh no worries at all. Understood. I just wanted to make sure that this sort of PR would be welcome. Please take your time with the review.

include/LightGBM/config.h
```diff
@@ -127,6 +127,7 @@ struct Config {
   // descl2 = label should be ``int`` type in lambdarank tasks, and larger number represents the higher relevance (e.g. 0:bad, 1:fair, 2:good, 3:perfect)
   // descl2 = `label_gain <#objective-parameters>`__ can be used to set the gain (weight) of ``int`` label
   // descl2 = all values in ``label`` must be smaller than number of elements in ``label_gain``
+  // desc = ``rank_xendcg``, `XE_NDCG_MART <https://arxiv.org/abs/1911.09798>`__ ranking objective function.
```
StrikerRUS: Can we have some aliases, e.g. `xe_ndcg_mart`, `cross_entropy_lambdamart`, `xe_lambdamart`, `xe_ndcg`, etc. (you know better which aliases may help users to find and remember this objective)?

src/objective/rank_xendcg_objective.hpp
```diff
@@ -0,0 +1,162 @@
...
+  data_size_t num_queries_;
+  /*! \brief Pointer of label */
+  const label_t* label_;
+  /*! \brief Query boundries */
+  const data_size_t* query_boundaries_;
+
+  // A pseudo-random number generator.
+  mutable std::mt19937 generator_;
+  // The Uniform distribution to sample from.
+  mutable std::uniform_real_distribution<double> distribution_;
```
StrikerRUS: Is it possible to enhance already existing `Random` class and then use it here?
https://github.com/microsoft/LightGBM/blob/d93984345445e09cea083032e85c17163edcb789/include/LightGBM/utils/random.h#L15-L18

src/objective/rank_xendcg_objective.hpp
```diff
@@ -0,0 +1,162 @@
...
+    const data_size_t start = query_boundaries_[query_id];
+    const data_size_t cnt =
+      query_boundaries_[query_id + 1] - query_boundaries_[query_id];
+    // add pointers with offset
+    const label_t* label = label_ + start;
+    score += start;
+    lambdas += start;
+    hessians += start;
+
+    // Turn scores into a probability distribution using Softmax.
```
StrikerRUS: I remember LightGBM already has `Softmax` implementation:
https://github.com/microsoft/LightGBM/blob/d93984345445e09cea083032e85c17163edcb789/include/LightGBM/utils/common.h#L598-L631

src/objective/rank_xendcg_objective.hpp
```diff
@@ -0,0 +1,162 @@
...
+    // Finally, prepare lambdas and hessians.
+    for (data_size_t i = 0; i < cnt; ++i) {
+      lambdas[i] = static_cast<score_t>(
+          L1s[i] + rho[i]*L2s[i] + rho[i]*L3s[i]);
+      hessians[i] = static_cast<score_t>(rho[i] * (1.0 - rho[i]));
+    }
+  }
+
+  double phi(const label_t l, double g) const {
+    return std::pow(2, l) - g;
```
StrikerRUS: https://github.com/microsoft/LightGBM/blob/d93984345445e09cea083032e85c17163edcb789/include/LightGBM/utils/common.h#L166
https://github.com/microsoft/LightGBM/blob/a8dc03250e235a00baf3ecefe4155023624dd239/include/LightGBM/utils/common.h#L587

include/LightGBM/config.h
sbruch: That makes sense. I've added a few aliases.

src/objective/rank_xendcg_objective.hpp
sbruch: Updated the code to reuse the Random class instead.

src/objective/rank_xendcg_objective.hpp
sbruch: Thanks! I've updated the code to use Common::Softmax().

src/objective/rank_xendcg_objective.hpp
sbruch: Using Common::Pow() now.

sbruch:
> I'm not a cpp review, but I have noticed some things related to general design:

Thanks for the feedback! I have addressed the issues you raised in a new commit.

sbruch:
@StrikerRUS I'm not sure I understand the continuous-integration failure above. It does not seem to be related to this PR. What do you think?

include/LightGBM/config.h
```diff
@@ -984,6 +986,9 @@ inline std::string ParseObjectiveAlias(const std::string& type) {
     return "cross_entropy_lambda";
   } else if (type == std::string("mean_absolute_percentage_error") || type == std::string("mape")) {
     return "mape";
+  } else if (type == std::string("xendcg") || type == std::string("xe_ndcg")
```
StrikerRUS:
```suggestion
  } else if (type == std::string("rank_xendcg") || type == std::string("xendcg") || type == std::string("xe_ndcg")
```

src/objective/rank_xendcg_objective.hpp
```diff
@@ -0,0 +1,149 @@
...
+#include <limits>
+#include <string>
+#include <algorithm>
+#include <cmath>
+#include <cstdio>
+#include <cstring>
+#include <math.h>
+#include <random>
+#include <vector>
+#include <random>
```
StrikerRUS: I think includes can be refined after the latest refactoring. At least, there is a duplicated `<random>` entry.

include/LightGBM/config.h
sbruch: Done.

src/objective/rank_xendcg_objective.hpp
sbruch: Thanks for catching this! I've cleaned up the includes.

sbruch:
> Oh, please do not worry! That failure is not related to this PR. It's a random failure we observe sometimes with MinGW exclusively and have no idea what causes it. Possibly, it's related to #1818.

Thanks! That's good to know.

tests/python_package_test/test_sklearn.py
```diff
@@ -117,6 +117,24 @@ def test_lambdarank(self):
...
+    def test_xendcg(self):
+        X_train, y_train = load_svmlight_file(os.path.join(os.path.dirname(os.path.realpath(__file__)),
+                                                           '../../examples/xendcg/rank.train'))
+        X_test, y_test = load_svmlight_file(os.path.join(os.path.dirname(os.path.realpath(__file__)),
+                                                         '../../examples/xendcg/rank.test'))
+        q_train = np.loadtxt(os.path.join(os.path.dirname(os.path.realpath(__file__)),
+                                          '../../examples/xendcg/rank.train.query'))
+        q_test = np.loadtxt(os.path.join(os.path.dirname(os.path.realpath(__file__)),
+                                         '../../examples/xendcg/rank.test.query'))
+        gbm = lgb.LGBMRanker(n_estimators=50, objective='rank_xendcg', random_state=31, n_jobs=1)
```
StrikerRUS: I think we should get the same results but with better speed without limiting `n_jobs=1`.

src/objective/rank_xendcg_objective.hpp
```diff
@@ -0,0 +1,141 @@
+#ifndef LIGHTGBM_OBJECTIVE_RANK_XENDCG_OBJECTIVE_HPP_
```
StrikerRUS: Can you please update https://github.com/microsoft/LightGBM/tree/master/windows and include the new file there?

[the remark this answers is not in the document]
sbruch: Done.

examples/xendcg/README.md
```diff
@@ -0,0 +1,33 @@
+XE_NDCG Ranking Example
+==================
```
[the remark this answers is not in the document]
sbruch: Done.

examples/xendcg/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+Here is an example for LightGBM to train a ranking model with the XE_NDCG loss [arxiv.org/abs/1911.09798].
```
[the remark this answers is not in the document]
sbruch: Done.

examples/xendcg/predict.conf
```diff
@@ -0,0 +1,6 @@
+
```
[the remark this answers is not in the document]
sbruch: Sorry about that. Removed.

include/LightGBM/config.h
```diff
@@ -127,6 +127,8 @@ struct Config {
   // descl2 = label should be ``int`` type in lambdarank tasks, and larger number represents the higher relevance (e.g. 0:bad, 1:fair, 2:good, 3:perfect)
   // descl2 = `label_gain <#objective-parameters>`__ can be used to set the gain (weight) of ``int`` label
   // descl2 = all values in ``label`` must be smaller than number of elements in ``label_gain``
+  // desc = ``rank_xendcg``, `XE_NDCG_MART <https://arxiv.org/abs/1911.09798>`__ ranking objective function.
+  // descl2 = aliases: ``xendcg``, ``xe_ndcg``, ``xe_ndcg_mart``, ``xendcg_mart``.
```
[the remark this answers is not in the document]
sbruch: Done.

tests/python_package_test/test_sklearn.py
```diff
@@ -117,6 +117,24 @@ def test_lambdarank(self):
         self.assertGreater(gbm.best_score_['valid_0']['ndcg@1'], 0.6333)
         self.assertGreater(gbm.best_score_['valid_0']['ndcg@3'], 0.6048)
 
+    def test_xendcg(self):
+        X_train, y_train = load_svmlight_file(os.path.join(os.path.dirname(os.path.realpath(__file__)),
```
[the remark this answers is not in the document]
sbruch: Done.

tests/python_package_test/test_sklearn.py
sbruch: If I increased the number of jobs, the calls to generate the next random number may be invoked in a different order, leading to a different value of NDCG.

src/objective/rank_xendcg_objective.hpp
sbruch: Done.

sbruch:
@StrikerRUS Thanks for the review! Please let me know if you have any further concerns/questions.

tests/python_package_test/test_sklearn.py
StrikerRUS: @sbruch 
Ah, I see! The following line causes irreproducible results for number of threads different from 1 under the OMP loop, right?
```
gammas[i] = rand_->NextFloat();
```

We do receive a lot of issues related to the reproducibility and do our best in this direction. In case users can obtain reproducible results with **XE_NDCG_MART** objective function only if they set `num_threads` to 1, I think it's definitely worth to mention this fact in LightGBM's docs. For example, in `config.h`:
```
// descl2 = to obtain reproducible results, you should disable parallelism by setting ``num_threads`` to 1
```

And/or in FAQ, like it is done for **bagging**: https://lightgbm.readthedocs.io/en/latest/FAQ.html#bagging-is-not-reproducible-when-changing-the-number-of-threads.

cc @guolinke

sbruch: Correct. Sorry I should have explained better why num_threads>1 fails the test case.

That makes sense. I've added a note to that effect in config.h as requested.

sbruch:
@StrikerRUS I was wondering if there are other changes I need to make and/or approvals I need to obtain to merge this PR?

src/objective/rank_xendcg_objective.hpp
```diff
@@ -0,0 +1,145 @@
...
+    std::vector<double> L1s(cnt);
+    for (data_size_t i = 0; i < cnt; ++i) {
+      L1s[i] = -phi(label[i], gammas[i])/sum_labels + rho[i];
+    }
+    // Second-order terms.
+    std::vector<double> L2s(cnt);
+    for (data_size_t i = 0; i < cnt; ++i) {
+      for (data_size_t j = 0; j < cnt; ++j) {
+        if (i == j) continue;
+        L2s[i] += L1s[j] / (1 - rho[j]);
```
StrikerRUS: Can't division by zero happen here (and in some other places)?

sbruch: A division by zero could only happen if \rho is 1. That could only happen if there is only a single document in the list (i.e., cnt is 1) -- that is because \rho is a probability distribution generated by Softmax. However, if cnt is 1, line 96 would cause the loop to exit.

So neither line 97 nor line 105 below should result in a division by zero. Does that make sense?

StrikerRUS: > Does that make sense?

Absolutely! Thank you!

src/objective/rank_xendcg_objective.hpp
```diff
@@ -0,0 +1,145 @@
...
+* \brief Implementation of the learning-to-rank objective function, XE_NDCG [arxiv.org/abs/1911.09798].
+*/
+class RankXENDCG: public ObjectiveFunction {
+ public:
+  explicit RankXENDCG(const Config& config) {
+    if (config.seed != 0) {
+      rand_ = new Random(config.seed);
+    } else {
+      rand_ = new Random();
+    }
```
StrikerRUS: @guolinke Does LightGBM require special treating for `seed == 0`? I thought seed is always required for reproducibility.

sbruch: If 'seed' is not set the expected behavior should be that a seed is selected at random, right? Here ideally I'd like to check if a seed has been set, though the only way to do that I believe is comparing with 0. Is there a different, preferred way of checking if seed is set/not-set?

StrikerRUS: @sbruch 
> If 'seed' is not set the expected behavior should be that a seed is selected at random, right?

I don't think so.

When `seed` is set, it's used to generate other seeds:
https://github.com/microsoft/LightGBM/blob/f6b8ecf648f77f48d95a15e9ba5220e52f5d584c/src/io/config.cpp#L187-L195

Otherwise, default values of concrete seeds are used. For instance, 
https://github.com/microsoft/LightGBM/blob/f6b8ecf648f77f48d95a15e9ba5220e52f5d584c/include/LightGBM/config.h#L270
https://github.com/microsoft/LightGBM/blob/f6b8ecf648f77f48d95a15e9ba5220e52f5d584c/include/LightGBM/config.h#L290
https://github.com/microsoft/LightGBM/blob/f6b8ecf648f77f48d95a15e9ba5220e52f5d584c/include/LightGBM/config.h#L349

I guess, we need a new param, let say, `objective_seed` to be consistent with the existing codebase. I can't find any references where `seed` is used directly.

cc @guolinke

sbruch: I'm fine doing whatever you and @guolinke decide is appropriate. As far as I understand, either:

1) Reduce the highlighted code to: rand_ = new Random(config.seed); or,
2) Introduce an objective_seed

are what you have suggested so far.

As an aside, note that "new Random()" does not factor config.seed in the initialization at all and sets the seed to a randomly generated number. That's what gave me the impression that if seed is unset the behavior is expected to be non-deterministic.

StrikerRUS: I personally vote for `#2` "Introduce an objective_seed". Let's wait for other opinions.

sbruch: Hi, I was wondering if you all have made a decision?

guolinke: I think an `objective_seed` is good.

guolinke:
@sbruch did you compared the time cost of the ranking objective?
Refer to #2701

src/objective/rank_xendcg_objective.hpp
sbruch: Done.

sbruch:
> @sbruch did you compared the time cost of the ranking objective?
> Refer to #2701

I do not have precise measurement of the training time, but in repeated experiments it appears on par with or faster than lambdarank.

include/LightGBM/config.h
```diff
@@ -746,6 +748,10 @@ struct Config {
   // desc = separate by ``,``
   std::vector<double> label_gain;
 
+  // desc = random seed for objectives
+  // desc = used only in the ``rank_xendcg`` objective
+  int objective_seed = 1;
```
StrikerRUS: I believe it should be `5`: `data_random_seed=1`, `feature_fraction_seed=2`, `bagging_seed=3`, `drop_seed=4`, ...

sbruch: Ah, right. Sorry about that. Fixed now.

guolinke:
LGTM

PR #2635, opened 2019-12-15
fixed some cpplint issues

StrikerRUS:
Refer to #1990.

97 -> 88

Address some recently introduced cpplint errors.

### DIFF
--- a/include/LightGBM/utils/text_reader.h
+++ b/include/LightGBM/utils/text_reader.h
@@ -17,6 +17,8 @@
 
 namespace LightGBM {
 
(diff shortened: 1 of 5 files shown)


include/LightGBM/utils/text_reader.h
```diff
@@ -17,6 +17,8 @@
 
 namespace LightGBM {
 
+const size_t kGbs = size_t(1024) * 1024 * 1024;
```
StrikerRUS: Define `const` only once.

PR #2649, opened 2019-12-26
warning for init_score in save_binary

### LINKED ISSUE #2639: Dataset.save_binary doesn't save init_score
Example to reproduce the bug below. I would expect the rmse after 1 tree for loading the Dataset from the saved binary file to be equal to 0.0108203 not 0.88188. This behavior indicates that `Dataset.save_binary` isn't including the `init_score` values in the file. 
```
import lightgbm as lgb
import numpy as np
data = np.random.rand(100, 10)
label = np.sum(data, axis=1)
init_score = label + np.random.rand(100) * 0.02
dat = lgb.Dataset(data, label=label, init_score=init_score)
dat_without_init_score = lgb.Dataset(data, label=label)

p = dict()
p['num_iterations'] = 1
p['metric'] = 'rmse'
print('rmse after 1 tree for the original Dataset')
lgb.train(p, dat, valid_sets=[dat], verbose_eval=1)


print('rmse after 1 tree for the original Dataset without init_score')
lgb.train(p, dat_without_init_score, valid_sets=[dat_without_init_score], verbose_eval=1)

filename = 'abc.lgb'
dat.save_binary(filename)
dat2 = lgb.Dataset(filename)

print('rmse after 1 tree for loading from binary')
lgb.train(p, dat2, valid_sets=[dat2], verbose_eval=1)
```

Results:
```
rmse after 1 tree for the original Dataset
[1]	training's rmse: 0.0108203
rmse after 1 tree for the original Dataset without init_score
[1]	training's rmse: 0.88188
rmse after 1 tree for loading from binary
[1]	training's rmse: 0.88188
```

guolinke:
fix #2639

### DIFF
--- a/R-package/R/lgb.Dataset.R
+++ b/R-package/R/lgb.Dataset.R
@@ -1109,6 +1109,9 @@ lgb.Dataset.set.reference <- function(dataset, reference) {
 
 #' Save \code{lgb.Dataset} to a binary file
 #'
+#' Please note that \code{init_score} is not saved in binary file.
+#' If you need it, please set it again after loading Dataset.
+#'
 #' @param dataset object of class \code{lgb.Dataset}
 #' @param fname object filename of output file
 #'
--- a/R-package/man/lgb.Dataset.save.Rd
+++ b/R-package/man/lgb.Dataset.save.Rd
@@ -15,7 +15,8 @@ lgb.Dataset.save(dataset, fname)
 passed dataset
 }
 \description{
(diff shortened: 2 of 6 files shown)


StrikerRUS:
I can confirm that the warning is shown as expected:



However, I believe that not many users read logs and the most of them are used to set `verbose=-1` and do not see any warnings at all. Therefore, I think it's worth to duplicate this warning in the docs.

StrikerRUS:
@guolinke Can we merge?

PR #2653, opened 2019-12-28
[R-package] fixed typo for variable valid_contain_train

jameslamb:
I've been working my way through `lgb.train()`, trying to refresh my understanding of it and learn more about some of the code that I haven't seen before.

I noticed that there is a misspelled variable...`vaild_contain_train` instead of `valid_contain_train`. This isn't causing any issues since it's completely internal, but it does mean that it's hard to use `grep`  or some other tool to search for the corresponding Python code. I do that often to see how the Python package handled a particular problem.

image

In this PR, I propose fixing those typos.

### DIFF
--- a/R-package/R/lgb.train.R
+++ b/R-package/R/lgb.train.R
@@ -157,7 +157,7 @@ lgb.train <- function(params = list(),
 
   # Construct datasets, if needed
   data$construct()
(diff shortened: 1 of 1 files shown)


StrikerRUS:
Close-reopen to re-trigger CI.

Issue #2604, opened 2019-12-02
Unify out results of LGBM_BoosterDumpModel and LGBM_BoosterSaveModel


StrikerRUS:
At present these two functions produce different output content. For example, text representation contains parameters and feature importance, while JSON format - not. These functions should produce the same content but in different formats. It will help many third-party libraries that work with LightGBM model's dumps.

```
import json

import numpy as np
import lightgbm as lgb

X = np.random.random((100, 2))
y = np.random.random(100)
lgb_data = lgb.Dataset(X, y)
bst = lgb.train({}, lgb_data, num_boost_round=3)
bst.save_model('save_model.txt')
with open('dump_model.json', 'w') as json_dump:
    json.dump(bst.dump_model(), json_dump, indent=2)
```

![image](https://user-images.githubusercontent.com/25141164/69920740-8f4f1600-149c-11ea-8c48-3ef8828d7475.png)

![image](https://user-images.githubusercontent.com/25141164/69920742-9413ca00-149c-11ea-8074-d93abbf8cf1e.png)

Unify:
- [x] feature importance (#2656)
- [x] feature infos (#2660)
- [ ] parameters
- [ ] trees' structure

StrikerRUS:
Closed in favor of being in #2302. We decided to keep all feature requests in one place.

Welcome to contribute this feature! Please re-open this issue (or post a comment if you are not a topic starter) if you are actively working on implementing this feature.

jameslamb:
This was locked accidentally. I just unlocked it. We'd still welcome contributions related to this feature!

PR #2656, opened 2019-12-29
added feature importance to JSON dump 

StrikerRUS:
Part of #2604.

Implementation is copied from `SaveModelToString` without sorting.

### DIFF
--- a/src/boosting/gbdt_model_text.cpp
+++ b/src/boosting/gbdt_model_text.cpp
@@ -60,7 +60,6 @@ std::string GBDT::DumpModel(int start_iteration, int num_iteration) const {
   }
   str_buf << "]," << '\n';
 
-
   std::vector<double> feature_importances = FeatureImportance(num_iteration, 0);
   // store the importance first
   std::vector<std::pair<size_t, std::string>> pairs;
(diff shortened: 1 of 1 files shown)


StrikerRUS:
@guolinke WDYT about allowing to specify feature importance type in dump / save function? Seems that it will break the existing compatibility for C API.

guolinke:
@StrikerRUS could you explain it with more details?

StrikerRUS:
@guolinke I mean, at present, feature importance is always dumped with `importance_type=0`:
https://github.com/microsoft/LightGBM/blob/edb9149c13fe5a901898db525c22b441f0db6e49/src/boosting/gbdt_model_text.cpp#L63
https://github.com/microsoft/LightGBM/blob/edb9149c13fe5a901898db525c22b441f0db6e49/src/boosting/gbdt_model_text.cpp#L333

Will be it useful to allow to specify importance type during saving / dumping model?

guolinke:
As we have standalone FeatureImportance API
https://github.com/microsoft/LightGBM/blob/516bd37a28a0dfa7c2bc4a24587dbbcb8a697eb2/src/c_api.cpp#L1563-L1574

I think adding `importance_type` to dump/save is not very necessary.

StrikerRUS:
Agree! Especially, given that it requires C API breakage.

StrikerRUS:
> WDYT about allowing to specify feature importance type in dump / save function? Seems that it will break the existing compatibility for C API.

Implemented in #3220.

PR #2662, opened 2020-01-03
[python] [R-package] Use the same address when updated label/weight/query

guolinke:


guolinke:
better fix (todo): use metadata.label/weight in objective and metric.

R-package/R/lgb.Booster.R
```diff
@@ -207,6 +208,12 @@ Booster <- R6::R6Class(
     # Perform boosting update iteration
     update = function(train_set = NULL, fobj = NULL) {
 
+      if (is.null(train_set)) {
+        if (private$train_set$.__enclos_env__$private$version != private$train_set_version) {
```
jameslamb: Instead of `x != y`, could you use `!identical(x, y)` here? The issue is that `having a `NULLL` on either side of a logical like this in R will cause an error.

image

R-package/R/lgb.Booster.R
```diff
@@ -55,6 +55,7 @@ Booster <- R6::R6Class(
 
           # Create private booster information
           private$train_set <- train_set
+          private$train_set_version <- train_set$.__enclos_env__$private$version
```
jameslamb: Maybe a naive question, but what does it mean to have a new "version" of the training set? Is it mainly updating weights as part of boosting? I think I am missing some context, so I don't understand what problem this patch solves.

R-package/R/lgb.Booster.R
guolinke: given there are two integers, do we still need to use `identical`?

R-package/R/lgb.Booster.R
guolinke: currently, Dataset is allowed to update label/weight/... (field data) during training, but the booster cannot capture these changes. And ideally, we should update related classes, such as the cached label/weight/.. in Objective/Metric. So I add a version number to indicate the changes of these fields.

R-package/R/lgb.Booster.R
jameslamb: I wasn't sure if they were guaranteed to always be non-null. You can use `!=`  if they will always be integers.

R-package/R/lgb.Booster.R
jameslamb: Ah I see ok, makes sense. Thanks!

R-package/R/lgb.Booster.R
guolinke: yeah, if there is a NULL, it means there are bugs in the code.

PR #2663, opened 2020-01-04
[R-package] moved parameter validations up earlier in function calls

jameslamb:
While working through the R package, I noticed several places where we have some validation of inputs to function calls below other running code. For example, right now if you pass something like a `data.frame` to `data` in `lgb.cv()` without including `labels`, you won't find out that you have a malformed training dataset  until after a lot of other steps have run, including initializing a `Predictor`.

In this PR,  I propose a few changes throughout  the package which move input validation up toward the beginning of functions, so that users are alerted sooner when something is wrong. I've also added unit tests on these checks  to be sure they're behaving correctly.

### DIFF
--- a/R-package/R/callback.R
+++ b/R-package/R/callback.R
@@ -29,11 +29,10 @@ cb.reset.parameters <- function(new_params) {
   # Run some checks in the beginning
   init <- function(env) {
 
-    # Store boosting rounds
-    nrounds <<- env$end_iteration - env$begin_iteration + 1L
-
     # Check for model environment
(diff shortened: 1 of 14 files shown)


jameslamb:
closed and re-opened to trigger CI. Appveyor failed with what I think is a transient error.

jameslamb:
@Laurae2 or @guolinke could you take a look at this when you have some time? Thank you.

jameslamb:
Rebased to `master`  to fix conflicts since we recently merged a few other R PRs (thanks @StrikerRUS !)

I'll merge this one after CI runs

PR #2669, opened 2020-01-06
option to disable the shape checking in prediction.

### LINKED ISSUE #2668: 2.3.1 does not handle model.predict("file.svm") well if the file has a smaller cardinality of features than what the model was trained with
I appear to have found a regression with 2.3.1 on how lightgbm  handles predictions from a libsvm file. After train, we have a holdout-lite set that we use for analysis. This holdout-lite set does not contain labels, we just use it to plot the distribution of outputs in different scenarios. These files may not have all the features the model was trained with. In 2.3.0, this was fine - the prediction happened anyway and the missing features were treated as mising. In 2.3.1 though, it breaks with a fatal exception error. See below:

## Environment info

Operating System: Linux

CPU/GPU model: i9900k

C++/Python/R version: Python / v2.3.1

## Error message

LightGBMError: The number of features in data (8013) is not the same as it was in training data (8355).

## Reproducible examples

https://gist.github.com/andrewortman/27997cb467ab77b4311f26efd26ecfae

## Steps to reproduce

1.  Use gist above as a template. Create a sample_train.svm file with say 5 features, but create a sample_predict.svm file with only 3 features
2. When using v2.3.0, prediction succeeds
3. When using 2.3.1, prediction fails with fatal exception that number of columns do not match

I would expect the prediction will still occur, allowing unseen columns to be treated as missing in the tree.

guolinke:
to fix #2668

### DIFF
--- a/docs/Parameters.rst
+++ b/docs/Parameters.rst
@@ -729,11 +729,11 @@ IO Parameters
 
    -  used only in ``prediction`` task
 
-   -  Control whether or not LightGBM raises an error when you try to predict on data with a different number of features than the training data
+   -  control whether or not LightGBM raises an error when you try to predict on data with a different number of features than the training data
 
-   -  If ``false`` (the default), a fatal error will be raised if the number of features in the dataset you predict on differs from the number seen during training
+   -  if ``false`` (the default), a fatal error will be raised if the number of features in the dataset you predict on differs from the number seen during training
 
-   -  If ``true``, LightGBM will attempt to predict on whatever data you provide. This is dangerous because you might get incorrect predictions, but you could use it in situations where it is difficult or expensive to generate some features and you are very confident that they were never chosen for splits in the model.
+   -  if ``true``, LightGBM will attempt to predict on whatever data you provide. This is dangerous because you might get incorrect predictions, but you could use it in situations where it is difficult or expensive to generate some features and you are very confident that they were never chosen for splits in the model
 
    -  **Note**: be very careful setting this parameter to ``true``
 
--- a/docs/Parameters.rst
+++ b/docs/Parameters.rst
@@ -725,6 +725,12 @@ IO Parameters
 
    -  the threshold of margin in early-stopping prediction
 
+-  ``predict_disable_shape_check`` :raw-html:`<a id="predict_disable_shape_check" title="Permalink to this parameter" href="#predict_disable_shape_check">&#x1F517;&#xFE0E;</a>`, default = ``false``, type = bool
+
+   -  used only in ``prediction`` task
+
+   -  if ``true``, will disable the shape checking in prediction. This is dangerous, please aware what you are doing
+
 -  ``convert_model_language`` :raw-html:`<a id="convert_model_language" title="Permalink to this parameter" href="#convert_model_language">&#x1F517;&#xFE0E;</a>`, default = ``""``, type = string
 
    -  used only in ``convert_model`` task
--- a/docs/Parameters.rst
+++ b/docs/Parameters.rst
@@ -729,7 +729,11 @@ IO Parameters
 
    -  used only in ``prediction`` task
 
-   -  if ``true``, will disable the shape checking in prediction. This is dangerous, please aware what you are doing
+   -  Control whether or not LightGBM raises an error when you try to predict on data with a different number of features than the training data.
+
+   -  If ``false`` (the default), a fatal error will be raised if the number of features in the dataset you predict on differs from the number seen during training.
+
+   -  If ``true``, LightGBM will attempt to predict on whatever data you provide. This is dangerous because you might get incorrect predictions, but you could use it in situations where it is difficult or expensive to generate some features and you are very confident that they were never chosen for splits in the model. Be very careful setting this parameter to ``true``.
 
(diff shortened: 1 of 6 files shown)


docs/Parameters.rst
```diff
@@ -725,6 +725,12 @@ IO Parameters
 
    -  the threshold of margin in early-stopping prediction
 
+-  ``predict_disable_shape_check`` :raw-html:`<a id="predict_disable_shape_check" title="Permalink to this parameter" href="#predict_disable_shape_check">&#x1F517;&#xFE0E;</a>`, default = ``false``, type = bool
+
+   -  used only in ``prediction`` task
+
+   -  if ``true``, will disable the shape checking in prediction. This is dangerous, please aware what you are doing
```
jameslamb: I think it would be valuable to elaborate on which "shape checking" this refers to. I didn't know until I read the text of the error message in the diff.

Would you consider something like this?

> Control whether or not `LightGBM` raises an error when you try to predict on data with a different number of features than the training data. If `false` (the default), a fatal error will be raised if the number of features  in the dataset you predict on differs from the number seen during training. If `true`, `LightGBM` will attempt to predict on whatever data you provide. This is dangerous because you might get incorrect predictions, but you could use it in situations where it is difficult or expensive to generate some features and you are very confident that they were never chosen for splits in the model. Be very careful setting this parameter to `true`.

jameslamb: Also, I am sort of guessing what the context is for this PR, but if you could give a more specific example for when it would be good to set this to `true` I think it would help users.

guolinke: Thanks @jameslamb for detailed documentation!

src/application/predictor.hpp
```diff
@@ -141,8 +141,9 @@ class Predictor {
     if (parser == nullptr) {
       Log::Fatal("Could not recognize the data format of data file %s", data_filename);
     }
-    if (!header && parser->NumFeatures() != boosting_->MaxFeatureIdx() + 1) {
-      Log::Fatal("The number of features in data (%d) is not the same as it was in training data (%d).", parser->NumFeatures(), boosting_->MaxFeatureIdx() + 1);
+    if (!header && !disable_shape_check && parser->NumFeatures() != boosting_->MaxFeatureIdx() + 1) {
+      Log::Fatal("The number of features in data (%d) is not the same as it was in training data (%d). " \
```
StrikerRUS:
```suggestion
      Log::Fatal("The number of features in data (%d) is not the same as it was in training data (%d).\n" \
```

src/application/predictor.hpp
```diff
@@ -141,8 +141,9 @@ class Predictor {
...
+                 "You can set ``predict_disable_shape_check=true`` to discard this error, but please aware what you are doing.", parser->NumFeatures(), boosting_->MaxFeatureIdx() + 1);
```
StrikerRUS:
```suggestion
                 "You can set ``predict_disable_shape_check=true`` to discard this error, but please be aware what you are doing.", parser->NumFeatures(), boosting_->MaxFeatureIdx() + 1);
```

src/c_api.cpp
```diff
@@ -253,8 +253,9 @@ class Booster {
                std::function<std::vector<std::pair<int, double>>(int row_idx)> get_row_fun,
                const Config& config,
                double* out_result, int64_t* out_len) {
-    if (ncol != boosting_->MaxFeatureIdx() + 1) {
-      Log::Fatal("The number of features in data (%d) is not the same as it was in training data (%d).", ncol, boosting_->MaxFeatureIdx() + 1);
+    if (!config.predict_disable_shape_check && ncol != boosting_->MaxFeatureIdx() + 1) {
+      Log::Fatal("The number of features in data (%d) is not the same as it was in training data (%d). "\
```
```suggestion
      Log::Fatal("The number of features in data (%d) is not the same as it was in training data (%d).\n"\
```

src/c_api.cpp
```diff
@@ -253,8 +253,9 @@ class Booster {
...
+                 "You can set ``predict_disable_shape_check=true`` to discard this error, but please aware what you are doing.", ncol, boosting_->MaxFeatureIdx() + 1);
```
```suggestion
                 "You can set ``predict_disable_shape_check=true`` to discard this error, but please be aware what you are doing.", ncol, boosting_->MaxFeatureIdx() + 1);
```

src/c_api.cpp
```diff
@@ -274,8 +275,9 @@ class Booster {
                std::function<std::vector<std::pair<int, double>>(int row_idx)> get_row_fun,
                const Config& config,
                double* out_result, int64_t* out_len) {
-    if (ncol != boosting_->MaxFeatureIdx() + 1) {
-      Log::Fatal("The number of features in data (%d) is not the same as it was in training data (%d).", ncol, boosting_->MaxFeatureIdx() + 1);
+    if (!config.predict_disable_shape_check && ncol != boosting_->MaxFeatureIdx() + 1) {
+      Log::Fatal("The number of features in data (%d) is not the same as it was in training data (%d)." \
```
```suggestion
      Log::Fatal("The number of features in data (%d) is not the same as it was in training data (%d).\n" \
```

src/c_api.cpp
```diff
@@ -274,8 +275,9 @@ class Booster {
...
+                 "You can set ``predict_disable_shape_check=true`` to discard this error, but please aware what you are doing." , ncol, boosting_->MaxFeatureIdx() + 1);
```
```suggestion
                 "You can set ``predict_disable_shape_check=true`` to discard this error, but please be aware what you are doing.", ncol, boosting_->MaxFeatureIdx() + 1);
```

Issue #2583, opened 2019-11-21
Extremely randomized trees


joegaotao:
The variety of mdels plays an important role in model ensemble. I tried some parameters such as "colsample_bytree, colsample_bynode" to make model more stable and different, but trees still grow by some criterion, resulting in the similar models. However, I tried the combination "extratree+lgb", and randomized tree could be used as the certain feature embeding tool  that improves model variety. So I suggest adding extremely randomized tree as base learner.

StrikerRUS:
Closed in favor of being in #2302. We decided to keep all feature requests in one place.

Welcome to contribute this feature! Please re-open this issue (or post a comment if you are not a topic starter) if you are actively working on implementing this feature.

StrikerRUS:
Opening as we have active PR for this feature: #2671.

StrikerRUS:
Closed via #2671.

PR #2671, opened 2020-01-06
Extremely randomized trees

btrotta:
Option to use extremely randomized trees as base learner, as requested in #2583.


btrotta:
Here is a short script to test the performance on the data from Porto Seguro competition on Kaggle. Overfitting is reduced slightly using extra-trees. However it's a little slower, I'm not sure why this is.
```
import pandas as pd
import numpy as np
import lightgbm as lgb
import time as time
import os

df = pd.read_csv(os.path.join('porto', 'train.csv'))
train_cols = [c for c in df.columns if c not in ['id', 'target'] and ('cat' not in c)]
cat_cols = [c for c in train_cols if '_cat' in c]
train_ind = np.random.choice(df.index, len(df) // 2, replace=False)
train_bool = df.index.isin(train_ind)
lgb_train = lgb.Dataset(df.loc[train_bool, train_cols], label=df.loc[train_bool, 'target'])
lgb_test = lgb.Dataset(df.loc[~train_bool, train_cols], label=df.loc[~train_bool, 'target'])

# base
params = {'objective': 'binary', 'seed': 0, 'num_leaves': 32, 'learning_rate': 0.01, 'metric': 'binary_logloss'}
t = time.time()
res = {}
est = lgb.train(params, lgb_train, valid_sets=valid_sets, valid_names=valid_names, num_boost_round=1000,
                categorical_feature=cat_cols, evals_result=res)
print('time ', time.time() - t)
plt.figure()
plt.plot(res['train']['binary_logloss'])
plt.plot(res['valid']['binary_logloss'])

# extra-trees
params['extra_trees'] = True
t = time.time()
res2 = {}
est = lgb.train(params, lgb_train, valid_sets=valid_sets, valid_names=valid_names, num_boost_round=1000,
                categorical_feature=cat_cols, evals_result=res2)
print('time ', time.time() - t)
plt.plot(res2['train']['binary_logloss'])
plt.plot(res2['valid']['binary_logloss'])
plt.legend(['train', 'test', 'train_extra', 'test_extra'])
```
extra_example

include/LightGBM/config.h
```diff
@@ -289,6 +289,11 @@ struct Config {
   // desc = random seed for ``feature_fraction``
   int feature_fraction_seed = 2;
 
+  // desc = use extremely randomized trees
+  // desc = if set to ``true``, when evaluating node splits LightGBM will check only one randomly-chosen threshold for each feature
+  // desc = can be used to deal with over-fitting
```
StrikerRUS: Please add it here as well: https://github.com/microsoft/LightGBM/blob/master/docs/Parameters-Tuning.rst#deal-with-over-fitting.

src/treelearner/feature_histogram.hpp
```diff
@@ -94,16 +95,20 @@ class FeatureHistogram {
     double gain_shift = GetLeafSplitGain(sum_gradient, sum_hessian,
                                          meta_->config->lambda_l1, meta_->config->lambda_l2, meta_->config->max_delta_step);
     double min_gain_shift = gain_shift + meta_->config->min_gain_to_split;
+    int rand_threshold = 0;
+    if (meta_->num_bin - 2 > 0){
+      rand_threshold = rand_.NextInt(0, std::max(meta_->num_bin - 2, 0));
+    }
```
StrikerRUS: Seems that `std::max(meta_->num_bin - 2, 0)` is excess here because `meta_->num_bin - 2 > 0`.

guolinke:
@btrotta for random trees, I think it is not need to construct histograms, which is the most time-consuming part in lightgbm. you can simply generate the random trees, and then predict over these trees, and then boosting, and next tree, ...

btrotta:
@guolinke Thanks for the advice! I will try implementing it that way.

btrotta:
@guolinke I made an attempt at implementing your suggestion but it actually made the performance worse, so I think I must be doing something wrong. Code is on a new branch (https://github.com/btrotta/LightGBM/tree/extra2) if you want to take a look.

Here's a summary of how I tried to implement it. When finding each new node, we need to try a random split of each feature on `smaller_leaf_splits` and `larger_leaf_splits`. For each of these, I first order the gradients and hessians (as is also done in the current code before constructing histograms). Then I do a parallel for-loop over the features. For each feature:
 1. Choose a random threshold to split.
 2. Iterate over the feature's bins to calculate the left and right sum of gradient and hessian.
 3. Calculate split gain.

Then choose the feature having the best random split and split the node on that feature.

In theory, it seems to me that this should be at least as fast as the normal GBDT algorithm, since instead of constructing the full histogram (where we need to aggregate gradients and hessians for many bins and save the result in memory), we only have to sum up gradients and hessians for one split.

But in fact it takes around twice as long (using the example script above), and I don't understand why. I'd be grateful if you have any insights.

Note: I didn't re-implement `ordered_sparse_bin`, since it seems like that would add more complexity, and I'm not sure if the efficiency gains are worth it in this case. But I don't think this is the reason for the bad performance: I tried my test script with all zeros replaced by -1 (so no sparse features), and performance is similar.

guolinke:
@btrotta sorry for the late response. 
I think for the extremely randomized tree, we don't need to gather gradient/hessian during tree growing.
I think the procedure may is:
1. generate a random tree, maybe generate one random int for selected feature, and one random float, for the threshold  ("threshold_in_bin = std::round(random_float*max_bin) ").
2. similar to `refit`, update the leaf output, according to grad and hess:
https://github.com/microsoft/LightGBM/blob/ef0b2d8287c21a1960b4e212f58e828f0ef264c4/src/treelearner/serial_tree_learner.cpp#L239-L264
BTW, the `refit_decay_rate` is zero in this case.

In this procedure, you don't need to call `SerialTreeLearner::Train`, maybe just a function similar to `SerialTreeLearner::FitByExistingTree`, and is called in GBDT class.

guolinke:
also refer to 
https://github.com/microsoft/LightGBM/blob/516bd37a28a0dfa7c2bc4a24587dbbcb8a697eb2/src/boosting/gbdt.cpp#L298-L321

you may need to get leaf preds before refit.

btrotta:
@guolinke 
>     1. generate a random tree, maybe generate one random int for selected feature, and one random float, for the threshold  ("threshold_in_bin = std::round(random_float*max_bin) "). And then, call `Tree::Split` to grow trees.

If I understand correctly, you're suggesting that for each new node we just randomly choose 1 feature and 1 threshold, and then split. I think this is not the usual definition of extremely randomized trees. For example see the sklearn docs (https://scikit-learn.org/stable/modules/ensemble.html#forest)
which say the following: 
> As in random forests, a random subset of candidate features is used, but instead of looking for the most discriminative thresholds, thresholds are drawn at random for each candidate feature and the best of these randomly-generated thresholds is picked as the splitting rule.

So it chooses 1 random threshold for each feature, but it evaluates many features then splits on the best one. 

I think if we only choose 1 random feature, the algorithm may not fit the data well. Indeed, in the original paper on extremely randomized trees (https://link.springer.com/content/pdf/10.1007/s10994-006-6226-1.pdf), they experiment with varying the size `K` of the subset of candidate features considered for each split. The results for `K=1` are very poor (see Figure 2). (Of course the situation is slightly different since their model is a random forest, not a boosting model like LightGBM. But still, I think the same idea probably applies.)

guolinke:
@btrotta I see. 
So the construcHistogram is also needed, therefore, the total time cost will be almost the same as before. And we can just use the code changes in this PR.

btrotta:
Updated benchmark script:

```
import pandas as pd
import numpy as np
import lightgbm as lgb
import time as time
import os
from sklearn.ensemble import ExtraTreesClassifier
from sklearn.metrics import log_loss

df = pd.read_csv(os.path.join('porto', 'train.csv'))
train_cols = [c for c in df.columns if c not in ['id', 'target'] and ('cat' not in c)]
cat_cols = [c for c in train_cols if '_cat' in c]
train_ind = np.random.choice(df.index, len(df) // 2, replace=False)
train_bool = df.index.isin(train_ind)
lgb_train = lgb.Dataset(df.loc[train_bool, train_cols], label=df.loc[train_bool, 'target'])
lgb_test = lgb.Dataset(df.loc[~train_bool, train_cols], label=df.loc[~train_bool, 'target'])

# base
params = {'objective': 'binary', 'seed': 0, 'num_leaves': 32, 'learning_rate': 0.01, 'metric': 'binary_logloss'}
t = time.time()
res = {}
est = lgb.train(params, lgb_train, valid_sets=valid_sets, valid_names=valid_names, num_boost_round=1000,
                categorical_feature=cat_cols, evals_result=res)

with open('log.txt', 'a') as f:
    f.write('lightgbm without extra_trees, time {} \n'.format(time.time() - t))
plt.figure()
plt.plot(res['train']['binary_logloss'])
plt.plot(res['valid']['binary_logloss'])

# extra-trees
params['extra_trees'] = True
t = time.time()
res2 = {}
est = lgb.train(params, lgb_train, valid_sets=valid_sets, valid_names=valid_names, num_boost_round=1000,
                categorical_feature=cat_cols, evals_result=res2)
with open('log.txt', 'a') as f:
    f.write('lightgbm with extra_trees, time {} \n'.format(time.time() - t))
plt.plot(res2['train']['binary_logloss'])
plt.plot(res2['valid']['binary_logloss'])

# extra-trees random forest
params['extra_trees'] = True
t = time.time()
res3 = {}
params['boosting'] = 'rf'
params['bagging_fraction'] = 0.8
params['bagging_freq'] = 1
est = lgb.train(params, lgb_train, valid_sets=valid_sets, valid_names=valid_names, num_boost_round=1000,
                categorical_feature=cat_cols, evals_result=res3)
with open('log.txt', 'a') as f:
    f.write('lightgbm random forest with extra_trees, time {} \n'.format(time.time() - t))
plt.plot(res3['train']['binary_logloss'])
plt.plot(res3['valid']['binary_logloss'])

# extra-trees random forest from sklearn
t = time.time()
est = ExtraTreesClassifier(n_estimators=1000, max_depth=5, max_features=1.0, max_samples=0.8, random_state=0)
est.fit(df.loc[train_bool, train_cols], df.loc[train_bool, 'target'])
prediction = pd.Series(est.predict_proba(df[train_cols])[:, 1], index=df.index)
train_loss = log_loss(df.loc[train_bool, 'target'], prediction.loc[train_bool])
test_loss = log_loss(df.loc[~train_bool, 'target'], prediction.loc[~train_bool])
with open('log.txt', 'a') as f:
    f.write('sklearn with extra_trees, time {} \n'.format(time.time() - t))
plt.scatter([1000], [train_loss])
plt.scatter([1000], [test_loss])

plt.legend(['train', 'test', 'train_extra', 'test_extra', 'train_extra_rf_lgb', 'test_extra_rf_lgb',
            'train_extra_rf_sklearn', 'test_extra_rf_sklearn'])
```

Output:

```
lightgbm without extra_trees, time 22.96505045890808 
lightgbm with extra_trees, time 22.61457848548889 
lightgbm random forest with extra_trees, time 27.370033502578735 
sklearn with extra_trees, time 1282.728404045105 
```
extra2

src/treelearner/feature_histogram.hpp
```diff
@@ -557,23 +569,25 @@ class FeatureHistogram {
         if (sum_left_hessian < meta_->config->min_sum_hessian_in_leaf) break;
 
         double sum_left_gradient = sum_gradient - sum_right_gradient;
-        // current split gain
-        double current_gain = GetSplitGains(sum_left_gradient, sum_left_hessian, sum_right_gradient, sum_right_hessian,
-          meta_->config->lambda_l1, meta_->config->lambda_l2, meta_->config->max_delta_step,
-          min_constraint, max_constraint, meta_->monotone_type);
-        // gain with split is worse than without split
-        if (current_gain <= min_gain_shift) continue;
+        if (!meta_->config->extra_trees || t - 1 + offset == rand_threshold) {
```
guolinke: maybe change `FindBestThresholdSequence` to the template method: `template<bool is_rand> FindBestThresholdSequence` ?

guolinke: with the template, the function will be expanded at compiling, and don't affect the run-time performance.

src/treelearner/feature_histogram.hpp
```diff
@@ -64,6 +64,7 @@ class FeatureHistogram {
       find_best_threshold_fun_ = std::bind(&FeatureHistogram::FindBestThresholdCategorical, this, std::placeholders::_1
         , std::placeholders::_2, std::placeholders::_3, std::placeholders::_4, std::placeholders::_5, std::placeholders::_6);
     }
+    rand_ = Random(meta_->config->seed);
```
StrikerRUS: @btrotta  Please check the following thread: https://github.com/microsoft/LightGBM/pull/2620#discussion_r365257424. I believe `seed` shouldn't be used for anything except generating other seeds.

guolinke:
@StrikerRUS is this ready to merge?

StrikerRUS:
@guolinke 
>  is this ready to merge?

Yes, I think so. But we should wait for CI fixes.

StrikerRUS:
@btrotta 
As usual, thank you very much for the high quality contribution! We really appreciate your efforts.

btrotta:
@StrikerRUS @guolinke Thanks for your reviews!

PR #2674, opened 2020-01-08
removed OpenMP ugly fix for Mojave

StrikerRUS:
@henry0312 I think 3.16.2 is quite stable and we can require 3.16 for macOS with AppleClang as we discussed earlier.


StrikerRUS:
PR to Homebrew has been proposed: https://github.com/Homebrew/homebrew-core/pull/48822.

StrikerRUS:
Hmmm, everything works fine on our CI services, but fails during building Homebrew formula (even for macOS 10.13, which should be unaffected).

@henry0312 Would you mind take a look: https://jenkins.brew.sh/job/Homebrew%20Core%20Pull%20Requests/54922 ?

henry0312:
Sorry, I don't know about Homebrew.

StrikerRUS:
@henry0312 OK, no problem! I think it's OK to leave the formula as is because it doesn't affect users' experience.

PR #2677, opened 2020-01-11
prefetch hint (1.3x speed-up)

guolinke:
set prefetch hint to increase cache hit chance.

benchmark:

| Data      | Master|  This PR | Improvement |  
|----|  ----| ----- | ----|
| Higgs|252.526271  |192.186991 s | 1.31x |
| Yahoo LTR|158.277699 s |129.526579 s | 1.22x |
| MS LTR|357.12529 s|251.736296 s | 1.42x |

### DIFF
--- a/include/LightGBM/bin.h
+++ b/include/LightGBM/bin.h
@@ -343,17 +343,18 @@ class Bin {
   *        ordered_gradients and ordered_hessians are preprocessed, and they are re-ordered by data_indices.
   *        Ordered_gradients[i] is aligned with data_indices[i]'s gradients (same for ordered_hessians).
   * \param data_indices Used data indices in current leaf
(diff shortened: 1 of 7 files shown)


Laurae2:
Here is a scalability (by number of threads) benchmark result on Bosch using non-sparse columns:

image

Hardware:

* 2x Dual Xeon 6154 (2x 18 cores / 36 threads)
* 768GB RAM

As we can notice, there is a substantial improvement below 36 threads (1 NUMA node), but using 2 NUMA nodes hit the performance.

Most number of threads were run 7 times but the tasks were extremely short (few seconds), hence the large variation.

guolinke:
@Laurae2 I guess that `prefetch_size` may need to tune according to `num_threads`.

PR #2687, opened 2020-01-13
[R-package] added test that Dataset$setinfo() converts 'group' input to integer

jameslamb:
Current, `Dataset$setinfo()` has the behavior "if you pass in something for  'group', convert it to integer before storing it". That behavior is not currently tested. This PR adds a test for it.


jameslamb:
rebased to `master` since some other R PRs have been merged

PR #2690, opened 2020-01-13
[python] fix trees_to_dataframe and enhance test

StrikerRUS:
- transfer `test_trees_to_dataframe` test from `test_basic` to `test_engine` with `skipIf` guard as we use `lgb.train()` there;
- add test for the edge case described in https://github.com/microsoft/LightGBM/pull/2592#discussion_r364435337;
- it turned out that dictionary keys are `list` in Python 2, so it's needed to convert them into a `set`

### DIFF
--- a/tests/python_package_test/test_engine.py
+++ b/tests/python_package_test/test_engine.py
@@ -1810,3 +1810,54 @@ class TestEngine(unittest.TestCase):
         predicted = est.predict(new_x)
         self.assertNotAlmostEqual(predicted[0], predicted[1])
         self.assertAlmostEqual(predicted[1], predicted[2])
+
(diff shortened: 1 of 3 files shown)


tests/python_package_test/test_engine.py
```diff
@@ -1810,3 +1810,54 @@ def test_binning_same_sign(self):
         predicted = est.predict(new_x)
         self.assertNotAlmostEqual(predicted[0], predicted[1])
         self.assertAlmostEqual(predicted[1], predicted[2])
+
+    @unittest.skipIf(not lgb.compat.PANDAS_INSTALLED, 'pandas is not installed')
```
jameslamb: oh this is great!

PR #2691, opened 2020-01-13
[ci] indicate support only for the last 3 macOS versions in wheel name

StrikerRUS:
According to this article, it seems that Apple officially supports last 3 versions.
https://www.howtogeek.com/350901/which-releases-of-macos-are-supported-with-security-updates/

I think we can do the same and avoid super huge filename which can cause errors in some file systems.


StrikerRUS:
Thanks @jameslamb ! I added a note about it.

PR #216, opened 2017-01-15
[WIP]optimize for the sparse feature

guolinke:
Remove OrderedSparseBin.
Combine many sparse bin to a pool and construct them simultaneously. 

- [ ] improve multi-threading performance

PR #2699, opened 2020-01-20
Support both row-wise and col-wise multi-threading

guolinke:
To continue #216 

Before this PR, LightGBM only supports col-wise multi-threading, which may is inefficient for sparse data.
The row-wise multi-threading is efficient for sparse data, but the overhead (more histograms and the merge cost) will increase with num_threads.

Therefore, as both of them are not perfect, we implement them both and automatically choose the faster one in run-time.

Two new parameters are added, ``force_col_wise`` and ``force_row_wise``.

Some other changes in this PR:
1.  remove `cnt` in Histogram, to reduce histogram merge cost
2.  Better `Timer` to profile the run time cost

Benchmark:

| Data      | v2.3.1|  Master | This PR | Speed-up to v2.3.1 |
|----|  ----| ----| ----- | ----- | 
| Higgs | 251.237950 s| 192.107919 s | 132.967057 s | 1.88x |
| Yahoo LTR | 163.160723 s |130.946914 s | 108.800116 s | 1.50x | 
| MS LTR| 356.769460 s|249.381964 s |186.277296 s | 1.91x |
| Expo | 122.580457 s | 115.084357 s | 68.163928  s | 1.79x |
| Allstate | 294.604991 s |335.108286 s |149.482392 s | 1.97x |

Run by 16 threads, on Azure ND24s VM

Todo:
- [x] fix tests
- [x] fix gpu code

updated:
A new PR for lint fix.


guolinke:
@Laurae2 for more benchmarks

src/treelearner/ocl/histogram16.cl
```diff
@@ -437,6 +404,8 @@ R""()
             // thread 8, 9, 10, 11, 12, 13, 14, 15 now process feature 0, 1, 2, 3, 4, 5, 6, 7's gradients for example 8, 9, 10, 11, 12, 13, 14, 15
             #if CONST_HESSIAN == 0
             atomic_local_add_f(gh_hist + addr2, stat2);
+            #else
+            atom_inc((uint*)(gh_hist + addr2));
```
guolinke: @huanzhang12 is this added okay? 
or simply `atomic_local_add_f(gh_hist + addr2, 1.0f)` ?

src/treelearner/ocl/histogram16.cl
```diff
@@ -642,22 +558,14 @@ R""()
...
-    }
     
     // now thread 0 - 7  holds feature 0 - 7's gradient for bin 0 and counter bin 0
     // now thread 8 - 15 holds feature 0 - 7's hessian  for bin 0 and counter bin 1
     // now thread 16- 23 holds feature 0 - 7's gradient for bin 1 and counter bin 2
     // now thread 24- 31 holds feature 0 - 7's hessian  for bin 1 and counter bin 3
     // etc,
 
+    // FIXME: correct way to fix hessians
 #if CONST_HESSIAN == 1
```
guolinke: @huanzhang12 could you help to fix the hessian fixing here?

guolinke:
@huanzhang12 I made some changes on `histogram16.cl`, could you help to review it?

src/treelearner/ocl/histogram16.cl
huanzhang12: Can you elaborate a little bit on what you want to do here? Do we need to add anything new to the histogram?

src/treelearner/ocl/histogram16.cl
huanzhang12: If the content at address (gh_hist + addr2) is an integer, we can use atom_inc(). If it is a floating point number, we must use atomic_local_add_f(gh_hist + addr2, 1.0f). I am not sure if you want to store a float or int here?

atomic_local_add_f() can be hundreds times slower than atom_inc(), since there is no native floating point atomics support on GPUs. So using atom_inc() if possible (but we must make sure the content is an integer).

src/treelearner/ocl/histogram16.cl
guolinke: Thanks! As the count is removed, the naive solution is to add gh_hist as before and remove the cnt_hist. However, for const_hessian cases, we can use `+1` in h_hist firstly and then multiply all h_hist with hessians[0], If `+1` is faster than `+hessians[i]`. Therefore, here we need to multiply all h_hist with hessians[0].

src/treelearner/ocl/histogram16.cl
guolinke: could we store it as int firstly and use inc(), then multiply it with hessian[0] and change it to float in-place?

guolinke:
@jameslamb could you help for the R's tests?

StrikerRUS:
Reminder for myself: run tests against swapped compilers before merging.

jameslamb:
> @jameslamb could you help for the R's tests?

sure!  Want me to push to this PR or make a separate one?

guolinke:
@jameslamb you can directly push to this PR.
ping @huanzhang12 for the GPU part.

guolinke:
@StrikerRUS I used the auto-formatting tool for google cpp style. 
The only problem is, some cpp doc string (the long lines) may break.

StrikerRUS:
@guolinke 
> I used the auto-formatting tool for google cpp style.
The only problem is, some cpp doc string (the long lines) may break.

Ah, I see. Now 102 files are affected in this PR which makes it extremely huge and unreviewable! I think it's better to revert the commit with automatic formatting and address cpplint issue in a separate PR (I can do it).

docs/Parameters.rst
```diff
@@ -190,6 +190,38 @@ Core Parameters
 Learning Control Parameters
 ---------------------------
 
+-  ``force_col_wise`` :raw-html:`<a id="force_col_wise" title="Permalink to this parameter" href="#force_col_wise">&#x1F517;&#xFE0E;</a>`, default = ``false``, type = bool
+
+   -  set ``force_col_wise=true`` will force LightGBM to use col-wise histogram build
+
+   -  Recommend ``force_col_wise=true`` when:
+
+      -  the number of cloumns is large, or the total number of bin is large
```
jameslamb:
```suggestion
      -  the number of columns is large, or the total number of bin is large
```

include/LightGBM/bin.h
```diff
@@ -469,6 +443,47 @@ class Bin {
   virtual Bin* Clone() = 0;
 };
 
+
+
```
jameslamb:
```suggestion
```

include/LightGBM/config.h
```diff
@@ -214,6 +214,24 @@ struct Config {
 
   #pragma region Learning Control Parameters
 
+  // desc = set ``force_col_wise=true`` will force LightGBM to use col-wise histogram build
+  // desc = Recommend ``force_col_wise=true`` when:
+  // descl2 = the number of cloumns is large, or the total number of bin is large
```
jameslamb:
```suggestion
  // descl2 = the number of columns is large, or the total number of bin is large
```

jameslamb:
> @jameslamb you can directly push to this PR.
> ping @huanzhang12 for the GPU part.

@guolinke I ended up deciding to make a separate branch and PR into this one, since  I saw you're still actively pushing commits to this branch. See #2719 , you can merge it whenever you want. Let me know if there's anything else you think should be tested.

Also, sorry if I missed it...is it expected that the learning-to-rank model produced changes  from this PR? I had to change my expected scores for lambdarank and saw you had to change them in Python too.

guolinke:
@jameslamb as the "count" in the histogram is removed, the performance on very small data, like the test data, may be affected.

jameslamb:
> @jameslamb as the "count" in the histogram is removed, the performance on very small data, like the test data, may be affected.

oh got it, ok thanks!

guolinke:
@jameslamb 

> [[1]]
> /home/travis/build/microsoft/LightGBM/R-package/tests/testthat/test_basic.R:291:46: style: There should be a space between right parenthesis and an opening curly brace.
>   for (bst in list(bst_row_wise, bst_colwise)){
>                                              ^~
> The command "bash .ci/test.sh" exited with 255.

R-package/tests/testthat/test_basic.R
```diff
@@ -252,3 +252,46 @@ test_that("lgb.train() throws an informative error if 'valids' contains lgb.Data
...
+    , nrounds = nrounds
+  )
+
+  expected_error <- 0.003070782
+  expect_equal(bst_colwise$eval_train()[[1L]][["value"]], expected_error)
+  expect_equal(bst_row_wise$eval_train()[[1L]][["value"]], expected_error)
+
+  # check some basic details of the boosters just to be sure force_col_wise
+  # and force_row_wise are not causing any weird side effects
+  for (bst in list(bst_row_wise, bst_colwise)){
```
guolinke:
```suggestion
  for (bst in list(bst_row_wise, bst_colwise)) {
```

guolinke:
@huanzhang12 I think the GPU code is almost fixed, you can review it when you have time : ).
@StrikerRUS I think this PR is almost ready, we can have a new PR for cpp lint after this.

src/treelearner/ocl/histogram16.cl
```diff
@@ -642,7 +727,7 @@ R""()
     // The is done by using an global atomic counter.
     // On AMD GPUs ideally this should be done in GDS,
     // but currently there is no easy way to access it via OpenCL.
-    __local uint * counter_val = (__local uint *)(gh_hist + 2 * DWORD_FEATURES * NUM_BINS * NUM_BANKS);
+    __local uint * counter_val = (__local uint *)(gh_hist + 2 * DWORD_FEATURES * NUM_BINS * NUM_BANKS);;
```
guolinke:
```suggestion
    __local uint * counter_val = (__local uint *)(gh_hist + 2 * DWORD_FEATURES * NUM_BINS * NUM_BANKS);
```

src/io/multi_val_dense_bin.hpp
```diff
@@ -0,0 +1,168 @@
...
+ */
+#ifndef LIGHTGBM_IO_MULTI_VAL_DENSE_BIN_HPP_
+#define LIGHTGBM_IO_MULTI_VAL_DENSE_BIN_HPP_
+
+
+#include <LightGBM/bin.h>
+
+#include <cstdint>
+#include <cstring>
+#include <omp.h>
```
StrikerRUS: @guolinke Shouldn't this include (here and in all other files) be handle via [`openmp_wrapper`](https://github.com/microsoft/LightGBM/blob/master/include/LightGBM/utils/openmp_wrapper.h) to not break [threadless version building](https://lightgbm.readthedocs.io/en/latest/Installation-Guide.html#build-threadless-version-not-recommended)?

guolinke:
@StrikerRUS do you know why `test_sklearn_integration` randomly failed?

guolinke:
also refer to https://github.com/scikit-learn/scikit-learn/issues/14106

StrikerRUS:
@guolinke 
> do you know why test_sklearn_integration randomly failed?

Nope, I have never seen that this test failing before. Is AVX2 critical for this PR?

guolinke:
@StrikerRUS 
I benchmarked for that, interestingly, disable avx2 is slightly faster.

guolinke:
@StrikerRUS I am gonging to merge this PR, do we need to swap back the compilers in CI?

StrikerRUS:
@guolinke 
> I benchmarked for that, interestingly, disable avx2 is slightly faster.

Hmm, really interesting...

> I am gonging to merge this PR, do we need to swap back the compilers in CI?

Yes. As it's already merged, I'll do this in a separate PR.

Issue #2628, opened 2019-12-11
[RFC] compatibility with scikit-learn


StrikerRUS:
New scikit-learn version (0.22) breaks compatibility with our current estimators from sklearn wrapper (not presented in `default` conda channel yet, that's why we have green `master`). There are two failing tests there:

```
============================= test session starts ==============================
platform linux -- Python 3.6.9, pytest-5.3.1, py-1.8.0, pluggy-0.13.rootdir: /home/travis/build/microsoft/LightGBM
collected 91 items                                                             
../tests/c_api_test/test_.py ..                                          [  2%]
../tests/python_package_test/test_basic.py .............                 [ 16%]
../tests/python_package_test/test_consistency.py ....                    [ 20%]
../tests/python_package_test/test_engine.py ............................ [ 51%]
................                                                         [ 69%]
../tests/python_package_test/test_plotting.py .....                      [ 74%]
../tests/python_package_test/test_sklearn.py ..........F...........F     [100%]
=================================== FAILURES ===================================
_________________________ TestSklearn.test_grid_search _________________________
self = <test_sklearn.TestSklearn testMethod=test_grid_search>
    def test_grid_search(self):
        X, y = load_iris(True)
        y = np.array(list(map(str, y)))  # utilize label encoder at it's max power
        X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.1, random_state=42)
        params = {'subsample': 0.8,
                  'subsample_freq': 1}
        grid_params = {'boosting_type': ['rf', 'gbdt'],
                       'n_estimators': [4, 6],
                       'reg_alpha': [0.01, 0.005]}
        fit_params = {'verbose': False,
                      'eval_set': [(X_test, y_test)],
                      'eval_metric': constant_metric,
                      'early_stopping_rounds': 2}
        grid = GridSearchCV(lgb.LGBMClassifier(**params), grid_params, cv=2)
>       grid.fit(X, y, **fit_params)
../tests/python_package_test/test_sklearn.py:164: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../../../miniconda/envs/test-env/lib/python3.6/site-packages/sklearn/model_selection/_search.py:652: in fit
    fit_params_values = indexable(*fit_params.values())
[...]

PR #2713, opened 2020-01-27
[python][ci] start to support Python 3.8

StrikerRUS:
Currently is blocked by the following upstream conda packaging issue: https://github.com/ContinuumIO/anaconda-issues/issues/11514.

Not changing default `PYTHON_VERSION` at CI services as 3.8 is still having a lot of problems in terms of third-party packages availability.


StrikerRUS:
https://github.com/ContinuumIO/anaconda-issues/issues/11514 was fixed yesterday.
But now I realized that this PR is actually blocked by one more issue: #2628. That issue was the reason we're supporting only `"scikit-learn<=0.21.3"` for now and support for Python 3.8 in scikit-learn was introduced only in [`0.22` version](https://github.com/scikit-learn/scikit-learn/releases). So, we should wait for https://github.com/scikit-learn/scikit-learn/issues/16241.

jameslamb:
> Looks good to me!

 ^ to be clear, I'm ok with the changes whenever the build works. Just mean I have no other things to add beyond that.

StrikerRUS:
@henry0312 I remember you were against setting new Python as default version before `x.1` release. Now that Python `3.8.2` is released, I'm bumping default versions at CI side.

Issue #2720, opened 2020-01-31
Improving compatibility with Scikit-learn


daviddiazvico:
## Summary

Move parameter validation and "_ parameter" setting from `__init__` to `fit` in the sklearn wrapper.

## Motivation

Attain full compatibility with sklearn and skopt automated hyper-parameter search. 

## Description

According to the [Scikit-learn development guide](https://scikit-learn.org/stable/developers/develop.html#parameters-and-init), the `__init__` method should not do any parameter validation or set any "trailing _" parameter.

In order to have an estimator compatible with the automated hyper-parameter search objects (GridSearchCV, RandomizedSearchCV, skopt.BayesSearchCV, etc.), that should be handled by the `fit` method.

In particular, the instantiation of `_Booster` and other parameters should be moved from `__init__` to the beginning of `fit` in the sklearn wrapper.

This is related to (a sub problem of) #2628 

Related PR: #2721 

Thanks!

## References

* [Scikit-learn development guide](https://scikit-learn.org/stable/developers/develop.html#parameters-and-init)

StrikerRUS:
Refer to https://github.com/microsoft/LightGBM/pull/2721#issuecomment-580902791.

PR #2721, opened 2020-01-31
Improved sklearn compatibility

daviddiazvico:
Related issue: #2720

### DIFF
--- a/python-package/lightgbm/sklearn.py
+++ b/python-package/lightgbm/sklearn.py
@@ -317,19 +317,8 @@ class LGBMModel(_LGBMModelBase):
         self.n_jobs = n_jobs
         self.silent = silent
         self.importance_type = importance_type
-        self._Booster = None
-        self._evals_result = None
-        self._best_score = None
-        self._best_iteration = None
-        self._other_params = {}
-        self._objective = objective
         self.class_weight = class_weight
-        self._class_weight = None
-        self._class_map = None
-        self._n_features = None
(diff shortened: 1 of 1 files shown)


StrikerRUS:
@daviddiazvico First of all, thank you very much for your efforts!

We are aware of the issue you've mentioned, please refer to #2628. However, scikit-learn official development guide doesn't say that it is prohibited to set any params with **leading** `_` in `__init__`, only the ones with **trailing** `_` are prohibited.

There is an upstream issue in which scikit-learn team are discussing right now how inconsistency between public API tests and development documentation should be fixed: https://github.com/scikit-learn/scikit-learn/issues/16241. Actually, an expected from third-party estimators behavior is described in [this comment](https://github.com/scikit-learn/scikit-learn/issues/16241#issuecomment-578991507).

Our sklearn wrapper is quite well tested with time and we are not aware of any confirmed issues with it. So, I don't think we want to refactor it right now. Instead, we'll wait for a fix of https://github.com/scikit-learn/scikit-learn/issues/16241, which I suppose will be in removing confusing public API check.

In addition, I've already prepared for any other decision on https://github.com/scikit-learn/scikit-learn/issues/16241 and prepared a branch a while ago which removes private attributes from `__init__` and looks a little bit more complete comparing to your PR: https://github.com/microsoft/LightGBM/compare/sklearn. So, I'm going to close this PR. Thanks again!

PR #2688, opened 2020-01-13
[R-package][docs] made roxygen2 tags explicit and cleaned up documentation

jameslamb:
We use `roxygen2` to document the R package. You can read more about this project [here](http://r-pkgs.had.co.nz/man.html), but basically you use specially-formatted comments in source code files and they get translated into R's special flavor of LaTeX (`.Rd`).

Right now, `lightgbm`'s roxygen2 stuff is relying on the  _convention_ that the first of these comments maps to the `title` element of a documentation page and the the first of these comments after that (broken by a blank line) becomes the `description` element. This can be error-prone, so in the interest of being explicit in this PR I propose that we _explicitly_ use the `@title` and `@description` tags.

Other changes in this PR:

* added previously-undeclared import of `utils`
* removed redundant uses of `@rdname` and replaced them with `@name`
* added docs on the expected form of `group` when using  `Dataset$setinfo("group")`
* clarified documentation on the expected form of `categorical_feature`

This PR also fixes LaTeX formatting issues that were smashing parameter names into their descriptions like this:

**Before**

lgb train-params-before

**After**

lgb train-params-after

PR #2725, opened 2020-01-31
[docs] fix RTD builds due to conda's problems

StrikerRUS:
image

Also, refer to https://github.com/microsoft/LightGBM/pull/2688#issuecomment-574174971.

Versions were taken from an accidental successful build: https://readthedocs.org/api/v2/build/10331271.txt.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -232,8 +232,7 @@ def generate_r_docs(app):
         r-matrix=1.2_14=r351h96ca727_0 \
         r-testthat=2.0.0=r351h29659fb_0 \
         cmake=3.14.0=h52cb24c_0 \
-        ca-certificates=2019.11.27=0 \
-        glib=2.56.2=hd408876_0
+        ca-certificates=2019.11.27=0
     /home/docs/.conda/bin/conda install -q -y -n r_env -c conda-forge \
         r-pkgdown=1.3.0=r35h6115d3f_1001 \
         r-roxygen2=6.1.1=r35h0357c0b_1001
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -233,7 +233,7 @@ def generate_r_docs(app):
         r-testthat=2.0.0=r351h29659fb_0 \
         cmake=3.14.0=h52cb24c_0 \
         ca-certificates=2019.11.27=0 \
-        glib=2.56.2=hd408876_0 \
+        glib=2.56.2=hd408876_0
     /home/docs/.conda/bin/conda install -q -y -n r_env -c conda-forge \
         r-pkgdown=1.3.0=r35h6115d3f_1001 \
         r-roxygen2=6.1.1=r35h0357c0b_1001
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -231,9 +231,10 @@ def generate_r_docs(app):
         r-jsonlite=1.5=r351h96ca727_0 \
         r-matrix=1.2_14=r351h96ca727_0 \
         r-testthat=2.0.0=r351h29659fb_0 \
-        cmake=3.14.0=h52cb24c_0
+        cmake=3.14.0=h52cb24c_0 \
+        ca-certificates=2019.11.27=0
     /home/docs/.conda/bin/conda install -q -y -n r_env -c conda-forge \
-        r-pkgdown=1.3.0=r351h6115d3f_1000 \
+        r-pkgdown=1.3.0=r35h6115d3f_1001 \
```
jameslamb: @StrikerRUS what if we just removed build numbers and did  `r-pkgdown=1.3.0`? Won't we have to fix this again if the RTD build machines change?

StrikerRUS: It may fail. In some cases even builds are important and with one number everything is OK, while with another one build crashes.

Sorry, I didn't get it: how RTD build machines are related to conda packages build numbers?

jameslamb: That string has some extra details in it that some packages use for builds that used, say, specific compilers for creation.

From [conda docs](https://docs.conda.io/projects/conda-build/en/latest/resources/package-spec.html):

> Differentiates builds of packages with otherwise identical names and versions, such as:

I've found in the past that my conda env files are not portable across operating systems until I remove those.

This is why `conda  env export` ships with the argument `--no-builds`, so you can build a conda environment file on one machine and feel confident it will work on others

image

StrikerRUS: Usually build increment indicates bug fixes:
https://github.com/conda-forge/r-pkgdown-feedstock#updating-r-pkgdown-feedstock
> In order to produce a uniquely identifiable distribution:
If the version of a package is not being increased, please add or increase the build/number.
If the version of a package is being increased, please remember to return the build/number back to 0.

jameslamb: I see, ok guess I misunderstood. Thanks

StrikerRUS: From the link you have provided above:
> 

build_number | integer | A non-negative integer representing the build number of the package. Unlike the build string, the build_number is inspected by conda. Conda uses it to sort packages that have otherwise identical names and versions to determine the latest one. This is important because new builds that contain bug fixes for the way a package is built may be added to a repository.
-- | -- | --

StrikerRUS: Despite that it seems we talked about different "build numbers" (you about `build`, me - `build_number`) and I have no idea how to specify `build_number` without `build` string, I guess you were confused by the world "inspects".

> Nothing in conda actually inspects the build string.

Maybe conda doesn't inspect build strings, it compares them foe sure. Otherwise, the following example will not work:
> The build string constraint `"numpy=1.11.2=*nomkl*"` matches the NumPy 1.11.2 packages without MKL but not the normal MKL NumPy 1.11.2 packages.
https://docs.conda.io/projects/conda-build/en/latest/resources/package-spec.html#examples

PR #2729, opened 2020-02-02
add is_sparse back

guolinke:


guolinke:
merge due to the need in #2594

Issue #2736, opened 2020-02-03
Upper and lower bound of the value a tree can give


JoanFM:
## Summary
Have an upper and lower bound of the value a tree can give

## Motivation
It would be able if the output need too be rescaled or given an offset for instanc

## Description
Have an upper and lower bound of the value a tree can give, given by the sum of all the max or min leaf values of each subtree

## Proposal
Proposal PR with potential solution: https://github.com/microsoft/LightGBM/pull/2737

PR #2737, opened 2020-02-03
Add capability to get possible max and min values for a model

JoanFM:
Proposal to provide feature described in: https://github.com/microsoft/LightGBM/issues/2736


src/io/tree.cpp
```diff
@@ -204,6 +204,14 @@ void Tree::AddPredictionToScore(const Dataset* data,
   }
 }
 
+void Tree::AddMaxValueToScore(double& score) const {
```
guolinke: why not just return the max/min values, and add outside?

JoanFM: I can do it, it just felt a little more coherent with methods in the same class that are accumulating score values, I know that in those cases they pass pointers because they are arrays. Just let me know which solution you like better.

JoanFM:
Apparently It fails to compile at linking time, but I do not experience this error locally

src/io/tree.cpp
guolinke: I think just returning the max/min score is better, as it is more standalone and will be possible used by others.

guolinke:
You could move the implementation from gbdt_prediction.cpp to gbdt.cpp .

src/io/tree.cpp
JoanFM: Done, thank you

JoanFM:
Changed according to comments

include/LightGBM/c_api.h
```diff
@@ -988,6 +988,24 @@ LIGHTGBM_C_EXPORT int LGBM_BoosterFeatureImportance(BoosterHandle handle,
                                                     int importance_type,
                                                     double* out_results);
 
+/*!
+ * \brief Get model upper bound value.
+ * \param handle Handle of booster
+ * \param[out] out_results Result pointer to max value
```
StrikerRUS: For the consistency with `LGBM_BoosterGetLowerBoundValue` function.
```suggestion
 * \param[out] out_results Result pointing to max value
```

include/LightGBM/tree.h
```diff
@@ -113,6 +113,18 @@ class Tree {
                             const data_size_t* used_data_indices,
                             data_size_t num_data, double* score) const;
 
+  /*!
+  * \brief Get upper bound leaf value of this tree model
+  * \param score Will add value to score
```
StrikerRUS: Copy-past issue? I don't see any params for this method.

include/LightGBM/tree.h
```diff
@@ -113,6 +113,18 @@ class Tree {
...
+  */
+  double GetUpperBoundValue() const;
+
+  /*!
+  * \brief Get lower bound leaf value of this tree model
+  * \param score Will add valu to score
```
StrikerRUS: Copy-past issue?

src/boosting/gbdt.cpp
```diff
@@ -657,6 +657,24 @@ void GBDT::GetPredictAt(int data_idx, double* out_result, int64_t* out_len) {
   }
 }
 
+double GBDT::GetUpperBoundValue() const {
+  double max_value = 0.0;
+  int total_tree = num_iteration_for_pred_ * num_tree_per_iteration_;
```
StrikerRUS: Possible overflow for int in case of large number of classes?

src/boosting/gbdt.cpp
```diff
@@ -657,6 +657,24 @@ void GBDT::GetPredictAt(int data_idx, double* out_result, int64_t* out_len) {
...
+  for (int i = 0; i < total_tree; ++i) {
+    max_value += models_[i]->GetUpperBoundValue();
+  }
+  return max_value;
+}
+
+double GBDT::GetLowerBoundValue() const {
+  double min_value = 0.0;
+  int total_tree = num_iteration_for_pred_ * num_tree_per_iteration_;
```
StrikerRUS: Possible overflow for int in case of large number of classes?

include/LightGBM/c_api.h
```diff
@@ -988,6 +988,24 @@ LIGHTGBM_C_EXPORT int LGBM_BoosterFeatureImportance(BoosterHandle handle,
...
+ * \return 0 when succeed, -1 when failure happens
+ */
+LIGHTGBM_C_EXPORT int LGBM_BoosterGetUpperBoundValue(BoosterHandle handle,
```
StrikerRUS: Can you please add corresponding methods for Python and R wrappers somewhere around here
https://github.com/microsoft/LightGBM/blob/dbb804f2350f280f3c927f792f7e55ac9c021707/python-package/lightgbm/basic.py#L2163-L2203
and here
https://github.com/microsoft/LightGBM/blob/dbb804f2350f280f3c927f792f7e55ac9c021707/R-package/R/lgb.Booster.R#L312-L322
https://github.com/microsoft/LightGBM/blob/dbb804f2350f280f3c927f792f7e55ac9c021707/src/lightgbm_R.cpp#L415-L423
respectively?

Also, a simple test will be really appreciated 🙂 .
https://github.com/microsoft/LightGBM/blob/dbb804f2350f280f3c927f792f7e55ac9c021707/tests/python_package_test/test_basic.py#L39-L41

StrikerRUS:
@JoanFM New asserts in basic test cause failures for all OSes:
```
>       self.assertAlmostEqual(bst.upper_bound(), 3.2911086148583832)
E       AssertionError: 3.3182142872462883 != 3.2911086148583832 within 7 places
```
I guess it's due to you have not the latest commit installed locally. Also, I think 4-6 digits after a dot will be enough to allow future contributions with insignificant changes will not require value adjustments.

JoanFM:
Thank you for your comments @StrikerRUS , I have done the changes, I hope it helps

tests/python_package_test/test_basic.py
```diff
@@ -39,6 +39,8 @@ def test(self):
         self.assertEqual(bst.current_iteration(), 20)
         self.assertEqual(bst.num_trees(), 20)
         self.assertEqual(bst.num_model_per_iteration(), 1)
+        self.assertAlmostEqual(bst.upper_bound(), 3.32, places=2)
+        self.assertAlmostEqual(bst.lower_bound(), -3.13, places=2)
```
StrikerRUS: `places=2` seems to be very poor comparison. Do you have any thoughts why is it fail with more strict checks?

JoanFM: I agree it is really poor, I do not know what is the reason, may be something about the interaction and transformation with Python? I based my basic.py implementation on get_leaf_output which is not tested, and I wonder if it might also have some dependency on platforms? I am not sure how python transforms c types. May you have a clue about this?

JoanFM: Or could it be that the trained model has slight variations? maybe it would be best to only test from a precomputed model that it gets the value it is supposed to get with less interactions

JoanFM:
Hello @StrikerRUS, Thank you for your advice, I changed the way I was testing and I could spot an error, I was obtaining max or min values based on the leaf_value_ vector, but it is initialized to the max_leaves which can be larger than the actual number of leaves that a tree might grow.

I hope it solves the issues

src/io/tree.cpp
```diff
@@ -207,11 +207,25 @@ void Tree::AddPredictionToScore(const Dataset* data,
 #undef PredictionFun
 
 double Tree::GetUpperBoundValue() const {
-    return *std::max_element(leaf_value_.begin(), leaf_value_.end());
+  double upper_bound = leaf_value_[0];
+  size_t size = std::min(static_cast<size_t>(num_leaves_), leaf_value_.size());
```
guolinke: you can just use the `num_leaves_`.

tests/python_package_test/test_basic.py
```diff
@@ -36,9 +74,13 @@ def test(self):
             if i % 10 == 0:
                 print(bst.eval_train(), bst.eval_valid())
 
+        (lower_bound, upper_bound) = find_upper_and_lower_bounds(bst)
+
         self.assertEqual(bst.current_iteration(), 20)
         self.assertEqual(bst.num_trees(), 20)
         self.assertEqual(bst.num_model_per_iteration(), 1)
+        self.assertAlmostEqual(bst.upper_bound(), upper_bound, places=5)
+        self.assertAlmostEqual(bst.lower_bound(), lower_bound, places=5)
```
StrikerRUS: Is it possible to hardcode expected values now?

src/io/tree.cpp
JoanFM: Done!, thanks

tests/python_package_test/test_basic.py
JoanFM: I added it along now! Thanks

JoanFM: I tried to add it also, but for lower bound I still see some precision loss even with 4 digits, with 3 digits it works. But since the test with the parsed json tree works, I think it is not anything comming from this feature implementation

StrikerRUS: Hmmm, I'm getting the following numbers:

```
Upper
3.3182142872462883 Linux
3.3182142872462883 macOS
3.3182142872462883 Windows

Lower
-2.9040190126976606 Linux
-2.9040190126976606 macOS
-2.9040190126976606 Windows
```

Please remove all `places=` arguments from asserts and we'll see with what exact numbers it fails.

JoanFM: Done, let's see what happens

JoanFM: For the lower bound the error appears in the 3rd decimal for the CI tests, but I am  not sure what platform or conditions it fails under

StrikerRUS: The failing tests are GPU ones. Please add `"gpu_use_dp": True` here. Everything should be OK then with any precision.
https://github.com/microsoft/LightGBM/blob/1e5049a1a6bdde4e37701a237b025bfe060e4119/tests/python_package_test/test_basic.py#L22-L30

JoanFM: Done, thank you very much for your help @StrikerRUS

StrikerRUS: Thanks a lot for all your hard debugging work! I think we can now remove `find_upper_and_lower_bounds` helper and leave only comparison with hardcoded values in the test.

JoanFM: Done! Thanks

src/c_api.cpp
```diff
@@ -357,6 +357,16 @@ class Booster {
     return boosting_->FeatureImportance(num_iteration, importance_type);
   }
 
+  double UpperBoundValue() const  {
```
```suggestion
  double UpperBoundValue() const {
```

src/io/tree.cpp
```diff
@@ -206,6 +206,26 @@ void Tree::AddPredictionToScore(const Dataset* data,
 
 #undef PredictionFun
 
+double Tree::GetUpperBoundValue() const {
+  double upper_bound = leaf_value_[0];
+  for (auto i = 1; i < num_leaves_; i++) {
```
StrikerRUS:
```suggestion
  for (int i = 1; i < num_leaves_; ++i) {
```

src/io/tree.cpp
```diff
@@ -206,6 +206,26 @@ void Tree::AddPredictionToScore(const Dataset* data,
...
+    if (leaf_value_[i] > upper_bound) {
+      upper_bound = leaf_value_[i];
+    }
+  }
+  return upper_bound;
+}
+
+double Tree::GetLowerBoundValue() const {
+  double lower_bound = leaf_value_[0];
+  for (auto i = 1; i < num_leaves_; i++) {
```
StrikerRUS:
```suggestion
  for (int i = 1; i < num_leaves_; ++i) {
```

tests/python_package_test/test_basic.py
```diff
@@ -9,6 +9,8 @@
 from scipy import sparse
 from sklearn.datasets import load_breast_cancer, dump_svmlight_file, load_svmlight_file
 from sklearn.model_selection import train_test_split
+import json
+import operator
```
StrikerRUS: Remove this.

tests/python_package_test/test_basic.py
```diff
@@ -26,7 +28,8 @@ def test(self):
             "num_leaves": 15,
             "verbose": -1,
             "num_threads": 1,
-            "max_bin": 255
+            "max_bin": 255,
+            "gpu_use_dp": "True"
```
StrikerRUS:
```suggestion
            "gpu_use_dp": True
```

tests/python_package_test/test_basic.py
JoanFM: Done!

src/c_api.cpp
```diff
@@ -357,6 +357,16 @@ class Booster {
     return boosting_->FeatureImportance(num_iteration, importance_type);
   }
 
+  double UpperBoundValue() const {
+    std::lock_guard<std::mutex> lock(mutex_);
```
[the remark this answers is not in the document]
JoanFM: to guarantee that this value is read while no one is changing any inner value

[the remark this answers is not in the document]
guolinke: okay, that is why we need reader/writer locks.

R-package/R/lgb.Booster.R
```diff
@@ -321,6 +321,30 @@ Booster <- R6::R6Class(
...
+
+    },
+
+    # Get lower bound
+    lower_bound_ = function() {
+
+      lower_bound <- 0L
+      lgb.call(
+        "LGBM_BoosterGetLowerBoundValue_R"
+        , ret = upper_bound
```
jameslamb: I was too slow getting to review here, apologies! @guolinke I'm going to make a followup PR  with some  changes:

* using trailing  `_` is very uncommon in R, these methods should be called  `upper_bound()`  and `lower_bound()`
* there is a bug on this line...return of the call to `LGBM_BoosterGetLowerBoundValue_R` should go into `lower_bound`, not `upper_bound`
* there should be unit tests added whenever we introduce new methods in the public API of the package

PR #2743, opened 2020-02-05
Config static initialization

jcipar:
We were running into a static destructor order bug caused by these two static data structures. This PR changes them to be static functions that return static member variables, which avoids the problem.

The references below are general info about the Static Initialization Order Fiasco. In our case, it was actually the *destructors* that were causing problems.

https://www.quora.com/What-is-static-initialization-order-fiasco-in-C

http://www.cs.technion.ac.il/users/yechiel/c++-faq/static-init-order-on-first-use-members.html


jcipar:
https://github.com/microsoft/LightGBM/issues/2745#issue-561262755

PR #2748, opened 2020-02-07
fix subset bug

guolinke:
to fix #2744

PR #2752, opened 2020-02-08
[ci] [R-package] Changed installation strategy to use binaries for data.table

jameslamb:
Sometimes the R builds on macOS fail, complaining about OpenMP issues in compiling `data.table`. See https://github.com/microsoft/LightGBM/pull/2748#issuecomment-583676147.

Per Matt Dowle's recommendation in https://github.com/Rdatatable/data.table/issues/2406, this PR changes our CI strategy to use the precompiled binaries for `data.table` available from CRAN. Those tend to only be a week or two between the source distribution, which I think is a reasonable price to pay in exchange for better stability in `LightGBM`'s CI.

For information on the change, see this detail from `?install.packages` in R:

image

### DIFF
--- a/.ci/test_r_package.sh
+++ b/.ci/test_r_package.sh
@@ -70,7 +70,7 @@ Rscript -e "install.packages(c('jsonlite', 'Matrix', 'R6', 'testthat'))" || exit
 if [[ $OS_NAME == "macos" ]]; then
     Rscript -e "install.packages('data.table', type = 'mac.binary')" || exit -1
 else
-    Rscript -e "install.packages('data.table', type = 'both')" || exit -1
+    Rscript -e "install.packages('data.table', type = 'source')" || exit -1
 fi
 
 cd ${BUILD_DIRECTORY}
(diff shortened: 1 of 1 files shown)


jameslamb:
I see this in the logs:

> Warning message:
package ‘data.table’ is not available (as a binary package for R version 3.6.1)

But when I run `install.packages('data.table', type='source')` on my Mac (running Mojave, and R 3.6.x), it works and I get this in the logs:

> trying URL 'https://cran.rstudio.com/bin/macosx/el-capitan/contrib/3.6/data.table_1.12.8.tgz'
Content type 'application/x-gzip' length 2133679 bytes (2.0 MB)
==================================================
downloaded 2.0 MB

I can look into it, but I'm about to start two days of travel. I know this is blocking other PRs so would appreciate if either @Laurae2 can look into it or (worst case) we comment-out the Mac builds and come back and restore them before the next time we do a release.

StrikerRUS:
I pushed about 15 commits in different branches and didn't see that error anymore. Is it possible that it was a temporary issue with CRAN infrastructure?

jameslamb:
> I pushed about 15 commits in different branches and didn't see that error anymore. Is it possible that it was a temporary issue with CRAN infrastructure?

Yes maybe it was just something transient? Seems to be working on other branches so let's close this PR.

PR #2754, opened 2020-02-08
Fix add features 

### LINKED ISSUE #2603: After executing add_features_from() many dataset fields remain unchanged
Dataset's `add_features_from()` method makes a lot of fields outdated and it's very confusing.
For example:

```
import numpy as np
import lightgbm as lgb

X1 = np.random.random((100, 2))
y1 = np.random.random(100)
lgb_data1 = lgb.Dataset(X1, y1, free_raw_data=False).construct()

X2 = np.random.random((100, 1))
y2 = np.random.random(100)
lgb_data2 = lgb.Dataset(X2, y2, free_raw_data=False).construct()

lgb_data2.add_features_from(lgb_data1)

lgb_data2.get_data().shape
>>> (100, 1)  # should be (100, 3)

lgb_data2._dump_text('dump.txt')
>>> feature_names: Column_0, Column_0, Column_1,  # duplicated column names
```

Gently ping @remcob-gr as author of #2006.

### LINKED ISSUE #2744: subset function cause corruption using 2.3.2
I have updated to  version 2.3.2. During dataset construction, espectially when I use the function `dataset.subset`, It caused corruption occasionally.  Corruptions not happened every time, but mostly on small datasets(less than 50k samples).

also，`dataset.add_features_from` caused corruptions at the same time.

Any idea on that？

guolinke:
fix #2744 and fix #2603  

@StrikerRUS could you help for the test cases?

- [x] test cases
- [x] merge data sources at Python side


python-package/lightgbm/basic.py
```diff
@@ -1640,6 +1640,11 @@ def add_features_from(self, other):
         if self.handle is None or other.handle is None:
             raise ValueError('Both source and target Datasets must be constructed before adding features')
         _safe_call(_LIB.LGBM_DatasetAddFeaturesFrom(self.handle, other.handle))
+        if other.data is None:
+            self.data = None
+        if self.data is not None:
+            # FIXME: concat two dataset
+            self.data = [self.data, other.data]
```
guolinke: @StrikerRUS here may need to concat two data by col.

StrikerRUS: @guolinke Should we consider all possible combinations of supported data sources?
https://github.com/microsoft/LightGBM/blob/1c1a276519439bdb10a91855bc8af8801cd819d5/python-package/lightgbm/basic.py#L1532-L1541

StrikerRUS:
@guolinke 
> could you help for the test cases?

Sure, but #2594 removes `test_add_features_*` tests completely. I think we should merge it firstly.

python-package/lightgbm/basic.py
guolinke: maybe we can convert the type of `other.data` to the type of `self.data` ?

StrikerRUS: OK, I'll try to do something.

StrikerRUS:
@guolinke I'm so sorry for the huge delay! Just found time to get back to this PR.

I started to add tests and accidently discovered that we do not have wrapper for `LGBM_DatasetGetFeatureNames` in Python (only for `LGBM_BoosterGetFeatureNames`). Will add it and continue with tests then.

Meanwhile can you please fix merge conflicts in cpp code? Also, I wonder are categorical features handled properly?

guolinke:
@StrikerRUS 
I think categorical features are only specially handled in `FeatureHistogram`, in most other places, it is the same as numerical features.

StrikerRUS:
@guolinke OK, so here we need set them reversely in Python wrapper, right?

jameslamb:
I'm going to remove myself from the list of reviewers. Please `@` me if you need me to come back or if this has implications for the R package.

guolinke:
@StrikerRUS 
sorry, I didn't get it. what is `reversely'?

StrikerRUS:
@guolinke For example, apply `df[cat_cols].astype('category')` for newly created pandas DataFrame if one source was `numpy` and another one was `pandas` with `auto` categorical features.

guolinke:
@StrikerRUS oh, I see. I think we should handle them in python/R-package correctly.

StrikerRUS:
@guolinke Seems that feature names resolving doesn't work in case of incremental adding of features. Please check the example below:
```
import numpy as np
import lightgbm as lgb

n_row = 100
n_col = 5
X = np.random.random((n_row, n_col))
names = ['col_%d' % i for i in range(n_col)]

d1 = lgb.Dataset(X, feature_name=names, free_raw_data=False).construct()
for _ in range(3):
    d2 = lgb.Dataset(X, feature_name=names, free_raw_data=False).construct()
    d1.add_features_from(d2)
    print(d1.get_feature_name())
```

```
['col_0', 'col_1', 'col_2', 'col_3', 'col_4', 'D2_col_0', 'D2_col_1', 'D2_col_2', 'D2_col_3', 'D2_col_4']
['col_0', 'col_1', 'col_2', 'col_3', 'col_4', 'D2_col_0', 'D2_col_1', 'D2_col_2', 'D2_col_3', 'D2_col_4', 'D2_col_0', 'D2_col_1', 'D2_col_2', 'D2_col_3', 'D2_col_4']
['col_0', 'col_1', 'col_2', 'col_3', 'col_4', 'D2_col_0', 'D2_col_1', 'D2_col_2', 'D2_col_3', 'D2_col_4', 'D2_col_0', 'D2_col_1', 'D2_col_2', 'D2_col_3', 'D2_col_4', 'D2_col_0', 'D2_col_1', 'D2_col_2', 'D2_col_3', 'D2_col_4']
```

guolinke:
@StrikerRUS 
sorry for missing that. I will fix it when I have time.

guolinke:
ping @StrikerRUS

src/io/dataset.cpp
```diff
@@ -1293,10 +1262,11 @@ void Dataset::ConstructHistogramsInner(
   int multi_val_groud_id = -1;
   used_dense_group.reserve(num_groups_);
   for (int group = 0; group < num_groups_; ++group) {
+    const int f_start = group_feature_start_[group];
```
guolinke, line 1265:
```suggestion
    const int f_start = group_feature_start_[group];
```

guolinke:
@StrikerRUS do you think this is ready?

StrikerRUS:
> do you think this is ready?

No, I don't think so. Python package still lacks tests and logic for updating feature names and `category` fields of Dataset class (https://github.com/microsoft/LightGBM/pull/2754#issuecomment-623126116) (maybe something more). I'm sorry, I really don't have a lot of time right now to propose a thoughtful code update. PR reviews looks more urgent and takes much time, unfortunately.
I believe that this PR should not be blocking us from 3.0 release as there are no breaking changes here, only bug fixes that we can release in a next minor release.

guolinke:
@StrikerRUS no problem, we can skip this for now.

StrikerRUS:
Also fix #3221.

StrikerRUS:
@guolinke 

Speaking about categorical features, it looks like it is not trivial to set them back at the Python side after calling `AddFeaturesFrom()`. In `other` dataset they can be provided by names, by indices, or even have `auto` value with pandas dataframe. The same cases are valid for `self` dataset. I don't know how to merge all possible combinations properly. Moreover, after cpp side we can have changed feature names (in case of duplicated names). I think the best option for now is just to warn users about this.

TBH, I don't want to hold this PR more...

guolinke:
@StrikerRUS Thanks very much. 
I think this is a rare use case, so no problem for the incomplete function. 
As long as the numerical features could be merged, I think it is okay.

guolinke:
@StrikerRUS  I cannot approve my own PR, you can approve and merge it if you think it is ready

StrikerRUS (approved):
@guolinke Thank you! I think it's ready.

Issue #1858, opened 2018-11-20
The higher version dependencies when build lightGBM


xiaoxiaofengzi:
` /lib64/libstdc++.so.6: version GLIBCXX_3.4.20 not found`
When I build LightGBM project, the lib_lightgbm.so needs the higher version dependencies. What can I do when I don't have the root authority of Linux machine? Thank you very much!

StrikerRUS:
Hi @xiaoxiaofengzi !

Can you please provide more information about your environment from GitHub issue template?

Also, it's not clear what do you mean by
> When I build LightGBM project ...

Are you trying to build LightGBM library from sources or your research project with LightGBM?

xiaoxiaofengzi:
Hi @StrikerRUS !
What I do is that I train a LightGBM classification model by using the scala version of mmlspark. And the version of mmlspark is `com.microsoft.ml.spark:mmlspark_2.11:0.14.dev9+1.g5783ce91`. However, when I solved the problem about the the version of GLIBC, a new problem about the version of GLIBCXX appeared.
The wrong details are as follows:

`18/11/19 15:11:26 WARN scheduler.TaskSetManager: Lost task 1.0 in stage 3.0 (TID 30, hadoop-188-10, executor 1): java.lang.UnsatisfiedLinkError: /data/deploy/opt/data/hadoop/tmp/hadoop-deploy/nm-local-dir/usercache/deploy/appcache/application_1542027375692_2593/container_e23_1542027375692_2593_01_000002/tmp/mml-natives6225024011850668628/lib_lightgbm.so: /lib64/libstdc++.so.6: version 'GLIBCXX_3.4.20' not found (required by /data/deploy/opt/data/hadoop/tmp/hadoop-deploy/nm-local-dir/usercache/deploy/appcache/application_1542027375692_2593/container_e23_1542027375692_2593_01_000002/tmp/mml-natives6225024011850668628/lib_lightgbm.so)
	at java.lang.ClassLoader$NativeLibrary.load(Native Method)
	at java.lang.ClassLoader.loadLibrary0(ClassLoader.java:1941)
	at java.lang.ClassLoader.loadLibrary(ClassLoader.java:1824)
	at java.lang.Runtime.load0(Runtime.java:809)
	at java.lang.System.load(System.java:1086)
	at com.microsoft.ml.spark.NativeLoader.loadLibraryByName(NativeLoader.java:59)
[...]

PR #2760, opened 2020-02-14
feat: Change locking strategy of Booster, allow for share and unique locks

JoanFM:
**Changes introduced**
I propose to have a different locking strategy that would allow to do singlePredictions in a parallel way.  
I propose having shared lock for **const** methods and **unique locks** for non const that are potentially writing values.
I added **const** to the methods that do Prediction since they should not alter the Booster and should be able to access the mutex in shared mode.

I hope this can help.


CMakeLists.txt
```diff
@@ -150,7 +150,7 @@ if(${MM_MALLOC})
 endif()
 
 if(UNIX OR MINGW OR CYGWIN)
-    SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -pthread -O3 -Wextra -Wall -Wno-ignored-attributes -Wno-unknown-pragmas -Wno-return-type")
+    SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++17 -pthread -O3 -Wextra -Wall -Wno-ignored-attributes -Wno-unknown-pragmas -Wno-return-type")
```
StrikerRUS: To support wide range of OSes we target gcc 4.8.2. Unfortunately, C++-17 is supported only starting from gcc 5.
Refer to 
https://github.com/microsoft/LightGBM/blob/704e08300ead3e70625d5431c845141f2e564123/CMakeLists.txt#L20-L38

JoanFM: So I assume this change is currently impossible, any plan to migrate?

guolinke: I think there are two options:
1) implement shared_mutex by our own. But this seems untrivial. 
2) define a preprocessor, called `DUSE_CPP17`, like `DUSE_MPI`, in cmake, and add `-std=c++17` if DUSE_CPP17 is set in cmake. And when `DUSE_CPP17` is used, use shared_mutex in cpp code.

JoanFM: Hello @guolinke thanks for the proposal , I did some changes with this proposed implementation

python-package/setup.py
```diff
@@ -116,6 +116,8 @@ def compile_cpp(use_mingw=False, use_gpu=False, use_mpi=False,
             cmake_cmd.append("-DOpenCL_LIBRARY={0}".format(opencl_library))
     if use_mpi:
         cmake_cmd.append("-DUSE_MPI=ON")
+    if use_cpp17:
+        cmake_cmd.append("_DUSE_CPP17=ON")
```
StrikerRUS:
```suggestion
        cmake_cmd.append("-DUSE_CPP17=ON")
```

StrikerRUS:
@JoanFM Please provide a rationale why is it useful to do singlePredictions in a parallel way and where it can be used to decide whether introduced maintenance burden worth it or not.

JoanFM:
> @JoanFM Please provide a rationale why is it useful to do singlePredictions in a parallel way and where it can be used to decide whether introduced maintenance burden worth it or not.

I think it is useful to have singlePrediction in a parallel way for instance in the case where a LightGBM model may be exposed in a service (maybe in a production environment) and might receive multiple queries at the same time. Allowing to have a single instance serve predictions in parallel can save the burden of needing to instance several boosters. For instance, in MMLSpark, some partitions can have more than one executor and benefit from this feature.

I know it may be a little bit of a problem to mantain or to add such a feature, so I cannot assess this impact. I just hope you find it useful.

Thank you very much

guolinke:
refer to https://stackoverflow.com/questions/14306797/c11-equivalent-to-boost-shared-mutex
Is that possible to implement a `shared_mutex` ?

JoanFM:
> refer to https://stackoverflow.com/questions/14306797/c11-equivalent-to-boost-shared-mutex
> and https://stackoverflow.com/questions/12033188/how-would-you-implement-your-own-reader-writer-lock-in-c11
> 
> Is that possible to implement a `shared_mutex` ?

Hi @guolinke , it is possible but I think it is really more complicated and error prone, it would involve I think atomic variables, waiting and signaling, I think that since I guess no one till now asked for this, and eventually at some point LightGBM may migrate c++17 it is easier to mantain the current implementation, and if someone wants this performance boosting for now, force them to have c++17 available.

Having said so, I know is not my call to make and if you prefer to keep a unique implementation for it and prefer a custom-made one I will try to propose one in this PR. 

Let me know what your opinions are.

guolinke:
@JoanFM Thanks, I agree with you.

StrikerRUS:
@JoanFM 
> and eventually at some point LightGBM may migrate c++17 

Due to the strong demand of support older systems, I don't think that it will be happen in the near future. For example, please check the following issues:
- #1945
- #1858
- https://github.com/Azure/mmlspark/issues/335
- https://github.com/Azure/mmlspark/issues/390
- https://github.com/Azure/mmlspark/issues/514

Introducing new CMake option for c++-17 means that we need to almost double all our CI jobs and install new compilers to ensure that it works fine.

As the proposed changes seem to be quite straightforward, I think it will be enough to create a guide in which describe the way one can compile LightGBM with allowed singlePredictions in a parallel way.

JoanFM:
> @JoanFM
> 
> > and eventually at some point LightGBM may migrate c++17
> 
> Due to the strong demand of support older systems, I don't think that it will be happen in the near future. For example, please check the following issues:
> 
> * #1945
> * #1858
> * [Azure/mmlspark#335](https://github.com/Azure/mmlspark/issues/335)
> * [Azure/mmlspark#390](https://github.com/Azure/mmlspark/issues/390)
> * [Azure/mmlspark#514](https://github.com/Azure/mmlspark/issues/514)
> 
> Introducing new CMake option for c++-17 means that we need to almost double all our CI jobs and install new compilers to ensure that it works fine.
> 
> As the proposed changes seem to be quite straightforward, I think it will be enough to create a guide with code examples in which describe the way one can compile LightGBM with allowed singlePredictions in a parallel way.

Hello @StrikerRUS , I guess it works, do you mean this proposal is good if we add some documentation on instructions to run with cpp17 that would activate the parallel prediction?

StrikerRUS:
@JoanFM Unfortunately, cpp is outside of the area of my expertise, thus I cannot tell you whether this exact code change proposal is good enough or not. It will do other reviewers.

However, I can judge the general conception. And I believe that proposed changes are quite niche and straightforward to introduce new build option and then maintain both c++-11 and c++-17 versions. Instead, I suggest do not change existing codebase at all, but write a doc with description and code examples which will show users how they can activate the parallel prediction in LightGBM if they need.

Please let me know what you think about this idea.

guolinke:
If moving to c++-17 will not happen in the near future, to reduce the maintenance cost, I think it is better to implement a shared mutex by ourown. 
I searched this in GitHub, and it seems not complicated, like https://github.com/Emanem/shared_mutex/blob/master/src/shared_mutex.h

JoanFM:
Hi @StrikerRUS, I am not sure I got it straight, do you propose to cancel this PR and just do some kind of documentation on how one would code a solution based on shared locking strategy?, or keep this one and add some documentation on how to activate the c++17 option?

Hi @guolinke, it may or not be an option, but I think it is really non trivial to come up with an implementation or take the idea from some other repo without knowing how much tested it is, There is also the option to use boost shared_mutex implementation that is what std lib is based upon, but it requires c++14. Would that change be more feasible?

StrikerRUS:
@JoanFM 
> I am not sure I got it straight, do you propose to cancel this PR and just do some kind of documentation on how one would code a solution based on shared locking strategy?

Yeah, exactly!

But @guolinke's proposal I like more!

CMakeLists.txt
```diff
@@ -150,7 +157,12 @@ if(${MM_MALLOC})
 endif()
 
 if(UNIX OR MINGW OR CYGWIN)
-    SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -pthread -O3 -Wextra -Wall -Wno-ignored-attributes -Wno-unknown-pragmas -Wno-return-type")
+    if(USE_BOOST_THREAD)
+        SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++1y")
```
guolinke: https://stackoverflow.com/questions/28825113/difference-between-gcc-compile-options-std-c1y-and-std-c14

CMakeLists.txt
```diff
@@ -119,6 +120,12 @@ if(USE_HDFS)
     SET(HDFS_CXX_LIBRARIES ${HDFS_LIB} ${JAVA_JVM_LIBRARY})
 endif(USE_HDFS)
 
+if(USE_BOOST_THREAD)
+    message(STATUS "Using boost threading library")
+    ADD_DEFINITIONS(-DUSE_BOOST_THREAD)
+    find_package(Boost COMPONENTS thread REQUIRED)
```
guolinke: If `Boost` is required, I prefer `c++-17`.  Except for GPU version, we always try to reduce the external dependencies as possible, for better cross-platform compatibility, and avoid the merge issues or possible bugs from upstream.

JoanFM: I just wanted to test if it worked. I will revert.

I just think that implementing it ourselves or trusting someone's implementation that is not standard can open a box of problems bigger than the ones that mantaining c++17 or extra dependencies would mean

JoanFM: Discovering a bug in the custom implementation would be much more difficult to deal with in my opinion I mean

guolinke:
@JoanFM @StrikerRUS 
what is the states of this PR? we use c++-17 or implement a new one?
Or is there any trade-off solution?

JoanFM:
> @JoanFM @StrikerRUS
> what is the states of this PR? we use c++-17 or implement a new one?
> Or is there any trade-off solution?

Let @StrikerRUS take the decision. Thanks!

StrikerRUS:
@guolinke 
> what is the states of this PR? we use c++-17 or implement a new one?
Or is there any trade-off solution?

I was hoping that someone suggested new idea or expressed their opinion. Since no one has done this, I still prefer one of the folloving options
- > I am not sure I got it straight, do you propose to cancel this PR and just do some kind of documentation on how one would code a solution based on shared locking strategy?

- > to reduce the maintenance cost, I think it is better to implement a shared mutex by ourown.

I clearly understand that implementing own (or borrowing) shared mutex is quite complicated and full of possible problems solution. But I suppose that it's the right one for a long-term perspective.

Also, I strongly believe that switching to `c++-17` right now is impossible as a lot of production environments use quite outdated (or very basic) OSes and cannot easily upgrade them.

JoanFM:
> @guolinke
> 
> > what is the states of this PR? we use c++-17 or implement a new one?
> > Or is there any trade-off solution?
> 
> I was hoping that someone suggested new idea or expressed their opinion. Since no one has done this, I still prefer one of the folloving options
> 
> * > I am not sure I got it straight, do you propose to cancel this PR and just do some kind of documentation on how one would code a solution based on shared locking strategy?
> * > to reduce the maintenance cost, I think it is better to implement a shared mutex by ourown.
> 
> I clearly understand that implementing own (or borrowing) shared mutex is quite complicated and full of possible problems solution. But I suppose that it's the right one for a long-term perspective.
> 
> Also, I strongly believe that switching to `c++-17` right now is impossible as a lot of production environments use quite outdated (or very basic) OSes and cannot easily upgrade them.

So maybe it is better to start a JIRA ticker or any other tool to track this development?

JoanFM:
Would it be okey if I try to use https://github.com/yohhoy/yamc for C++11 shared_mutex?

guolinke:
@JoanFM it seems yamc is under MIT license, so it is okay to be used in our project. 
there are 2 options:
1. use yamc as submodule, and include its code when building lightgbm.
2. like json11, we can copy its implement for shared_mutex (with original license) in LightGBM/utils.

JoanFM:
Hello @guolinke,

Thank you very much for your comment. I added them in LightGBM/utils, but I am afraid lint is failing for them. Do you have a util that adapts a source file to your lint standards?

Thanks

guolinke:
I mainly use the visual studio to re-format to google cpp style. 
But there are some issues, especially like reference and pointer, needs to fix manually. 
you can add me to your fork repo, I can help you for the automatically re-format part.

JoanFM:
> I mainly use the visual studio to re-format to google cpp style.
> But there are some issues, especially like reference and pointer, needs to fix manually.
> you can add me to your fork repo, I can help you for the automatically re-format part.

Hello @guolinke thank you for your response, I would really appreciate your support on that. I added you in my fork.

guolinke:
@JoanFM now you can fix the remaining issues.

StrikerRUS:
Happy to see this PR is moving forward! 🎉 

Just one comment. Will it make sense to move new files into a `yamc` folder `LightGBM/include/LightGBM/utils/yamc`?

Also ping @imatiach-msft as a potential benefiter and because I guess this PR breaks `MMLSpark` due to new function `SetSingleRowPredictor`.

JoanFM:
> Happy to see this PR is moving forward! 
> 
> Just one comment. Will it make sense to move new files into a `yamc` folder `LightGBM/include/LightGBM/utils/yamc`?
> 
> Also ping @imatiach-msft as a potential benefiter and because I guess this PR can break `MMLSpark` due to new function `SetSingleRowPredictor`.

I think it would make sense, yes.

JoanFM:
> Happy to see this PR is moving forward! 🎉
> 
> Just one comment. Will it make sense to move new files into a `yamc` folder `LightGBM/include/LightGBM/utils/yamc`?
> 
> Also ping @imatiach-msft as a potential benefiter and because I guess this PR can break `MMLSpark` due to new function `SetSingleRowPredictor`.

Done!

src/c_api.cpp
```diff
@@ -46,6 +49,12 @@ catch(std::string& ex) { return LGBM_APIHandleException(ex); } \
 catch(...) { return LGBM_APIHandleException("unknown exception"); } \
 return 0;
 
+#define UNIQUE_LOCK(mtx) \
+yamc::scoped_lock<yamc::alternate::shared_mutex> lock(mtx);
```
guolinke, line 53: do we need to use yamc for unique lock?

JoanFM: Makes sense! Done! Thanks!

windows/LightGBM.vcxproj.filters
```diff
@@ -210,6 +210,18 @@
...
     </ClInclude>
+    <ClInclude Include="..\include\LightGBM\utils\yamc\alternate_shared_mutex.hpp">
+      <Filter>include\LightGBM\utils</Filter>
+    </ClInclude>
+    <ClInclude Include="..\include\LightGBM\utils\yamc\yamc_rwlock_sched.hpp">
+      <Filter>include\LightGBM\utils</Filter>
+    </ClInclude>
+    <ClInclude Include="..\include\LightGBM\utils\yamc\yamc_scoped_lock.hpp">
+      <Filter>include\LightGBM\utils</Filter>
+    </ClInclude>
```
guolinke, line 221:
```suggestion
```

windows/LightGBM.vcxproj
```diff
@@ -255,6 +256,9 @@
     <ClInclude Include="..\include\LightGBM\utils\random.h" />
     <ClInclude Include="..\include\LightGBM\utils\text_reader.h" />
     <ClInclude Include="..\include\LightGBM\utils\threading.h" />
+    <ClInclude Include="..\include\LightGBM\utils\yamc\yamc_rwlock_sched.hpp" />
+    <ClInclude Include="..\include\LightGBM\utils\yamc\yamc_scoped_lock.hpp" />
```
guolinke, line 260:
```suggestion
```

JoanFM:
Are the failures real? or may they come from the github outage? Is there any way to just relaunch the jobs?

JoanFM:
> Are the failures real? or may they come from the github outage? Is there any way to just relaunch the jobs

Thanks @guolinke , I am not sure what are the issues with these r-packages.

StrikerRUS:
Closing-reopening doesn't help in this case.

@jameslamb Please take a look:
```
  Warning: package ‘R6’ was built under R version 4.0.1
```

windows/LightGBM.vcxproj.filters
```diff
@@ -210,6 +210,15 @@
     <ClInclude Include="..\src\treelearner\col_sampler.hpp">
       <Filter>src\treelearner</Filter>
     </ClInclude>
+    <ClInclude Include="..\include\LightGBM\utils\yamc\alternate_shared_mutex.hpp">
+      <Filter>include\LightGBM\utils</Filter>
```
StrikerRUS, line 214: Not sure.
```suggestion
      <Filter>include\LightGBM\utils\yamc</Filter>
```

windows/LightGBM.vcxproj.filters
```diff
@@ -210,6 +210,15 @@
...
+    </ClInclude>
+    <ClInclude Include="..\include\LightGBM\utils\yamc\yamc_rwlock_sched.hpp">
+      <Filter>include\LightGBM\utils</Filter>
```
StrikerRUS, line 217: Not sure.
```suggestion
      <Filter>include\LightGBM\utils\yamc</Filter>
```

windows/LightGBM.vcxproj.filters
```diff
@@ -210,6 +210,15 @@
...
+    </ClInclude>
+    <ClInclude Include="..\include\LightGBM\utils\yamc\yamc_shared_lock.hpp">
+      <Filter>include\LightGBM\utils</Filter>
```
StrikerRUS, line 220: Not sure.
```suggestion
      <Filter>include\LightGBM\utils\yamc</Filter>
```

StrikerRUS:
> Closing-reopening doesn't help in this case.

Should be fixed by #3225.

StrikerRUS:
> Should be fixed by #3225.

Merged!
@JoanFM Sorry for the inconvenience, but could you please rebase to the latest `master`?

JoanFM:
> > Should be fixed by #3225.
> 
> Merged!
> @JoanFM Sorry for the inconvenience, but could you please rebase to the latest `master`?

Done!

include/LightGBM/utils/yamc/alternate_shared_mutex.hpp
```diff
@@ -0,0 +1,212 @@
+/*
+ * alternate_shared_mutex.hpp
+ *
+ * MIT License
+ *
+ * Copyright (c) 2017 yohhoy
```
imatiach-msft, line 6: What is this copyright?  Did this file come from a different repository?
" * Copyright (c) 2017 yohhoy"

JoanFM: Yes, it comes from https://github.com/yohhoy/yamc

imatiach-msft: looks like it comes from here:
https://github.com/yohhoy/yamc
Hmm, is there a way to get the built artifacts from that library instead of copying some of the files over?  Ideally we wouldn't want to create duplicate code (or a "fork" in some sense) - it would be better to update the mutex library code in one place (eg for bug fixes, new features, etc).

Oh, what about git submodules?  Maybe that could be used here?
Example link:
https://git-scm.com/book/en/v2/Git-Tools-Submodules

JoanFM: Since it is a header-only library, I think they do not have any built artifacts.

imatiach-msft:
@StrikerRUS @JoanFM @guolinke the API changes look great to me.  I'm a bit concerned about having duplicate yamc code though, since it can drift from the original code in the repo.  Maybe we could either use the artifacts from that repo or use git submodules?  What do you think?  How different is the yamc code in this PR from the original repo?

include/LightGBM/utils/yamc/alternate_shared_mutex.hpp
JoanFM: the ideal soloution I was proposing in this PR was having the possibility to use c++17, but too much trouble would be caused. And coming with a custom implementation of shared mutex and locks is really error-prone.

JoanFM:
> @StrikerRUS @JoanFM @guolinke the API changes look great to me. I'm a bit concerned about having duplicate yamc code though, since it can drift from the original code in the repo. Maybe we could either use the artifacts from that repo or use git submodules? What do you think? How different is the yamc code in this PR from the original repo?

There are some changes caused by the lint style.Non-const references are forbidden and should be changed by pointers. 

Using git submodules can be a great solution but I do not know how they work.

guolinke:
@JoanFM @imatiach-msft 
I think if yamc is stable and not updated frequently, current solution is okay. 
From the commit logs, it seems our used file is not updated for two years. 
Using submodule is also good. It all depends on you.

JoanFM:
> @JoanFM @imatiach-msft 
> I think if yamc is stable and not updated frequently, current solution is okay. 
> From the commit logs, it seems our used file is not updated for two years. 
> Using submodule is also good. It all depends on you.

Could we go on with this PR, and in another one I try to apply the submodule solution?

guolinke:
@JoanFM is that possible to test the functionality of shared_lock? as it is from the third part.

JoanFM:
> @JoanFM is that possible to test the functionality of shared_lock? as it is from the third part.

might be possible to import tests from the original library

guolinke:
yeah, they test for the c++ directly. It is a good start for us to enable taste over c++ codes.

JoanFM:
> yeah, they test for the c++ directly. It is a good start for us to enable tests over c++ codes.

yes, definitely. Would it be possible for someone in the team to put a first test and set it into the CMake and ci so that I can continue  purting these tests? I am not familiar with the process you follow and would make it easier for me.

StrikerRUS:
@guolinke 
> is that possible to test the functionality of shared_lock?

Isn't it tested indirectly by our existing Python and R tests where the changed underlying cpp functions are used?
Actually we need tests for **new** functionality "allow to do singlePredictions in a parallel way", right? I believe it could be done in a separate PR to not slowdown merging this one because cpp tests are not something that we can ship fast.

JoanFM:
> @guolinke
> 
> > is that possible to test the functionality of shared_lock?
> 
> Isn't it tested indirectly by our existing Python and R tests where the changed underlying cpp functions are used?
> Actually we need tests for **new** functionality "allow to do singlePredictions in a parallel way", right? I believe it could be done in a separate PR to not slowdown merging this one because cpp tests are not something that we can ship fast.

I agree, I think adding git submodules and cpp tests deserve a separate PR actually

imatiach-msft:
@JoanFM agreed, I think it's ok to not use git submodules for now, especially if they are not updating their code frequently.  However, long term I always try to reduce code duplication/forking since it increases maintenance costs.  However, it sounds like this is not something we want to do right now.

JoanFM:
Hello @StrikerRUS @guolinke, after solving a couple of conflicts I have issues again with the r-package.

guolinke:
ping @jameslamb for R's CI.

StrikerRUS:
Actually, all tests are failing, not just R ones.

I believe the problem is in conflicts after #2992 was merged. For example, see compilation errors:
```
[ 96%] Building CXX object CMakeFiles/_lightgbm.dir/src/c_api.cpp.o
/home/travis/build/microsoft/LightGBM/src/c_api.cpp:1982:86: error: too many
      arguments to function call, expected 6, have 7
  ...get_row_fun, fastConfig->config, out_result, out_len);
                                                  ^~~~~~~
/home/travis/build/microsoft/LightGBM/src/c_api.cpp:381:3: note: 
      'PredictSingleRow' declared here
  void PredictSingleRow(int predict_type, int ncol,
  ^
/home/travis/build/microsoft/LightGBM/src/c_api.cpp:2116:53: error: too many
      arguments to function call, expected 6, have 7
                                        out_result, out_len);
                                                    ^~~~~~~
/home/travis/build/microsoft/LightGBM/src/c_api.cpp:381:3: note: 
      'PredictSingleRow' declared here
  void PredictSingleRow(int predict_type, int ncol,
  ^
2 errors generated.
CMakeFiles/_lightgbm.dir/build.make:439: recipe for target 'CMakeFiles/_lightgbm.dir/src/c_api.cpp.o' failed
```

JoanFM:
> Actually, all tests are failing, not just R ones.
> 
> I believe the problem is in conflicts after #2992 was merged. For example, see compilation errors:
> 
> ```
> [ 96%] Building CXX object CMakeFiles/_lightgbm.dir/src/c_api.cpp.o
> /home/travis/build/microsoft/LightGBM/src/c_api.cpp:1982:86: error: too many
>       arguments to function call, expected 6, have 7
>   ...get_row_fun, fastConfig->config, out_result, out_len);
>                                                   ^~~~~~~
> /home/travis/build/microsoft/LightGBM/src/c_api.cpp:381:3: note: 
>       'PredictSingleRow' declared here
>   void PredictSingleRow(int predict_type, int ncol,
>   ^
> /home/travis/build/microsoft/LightGBM/src/c_api.cpp:2116:53: error: too many
>       arguments to function call, expected 6, have 7
>                                         out_result, out_len);
>                                                     ^~~~~~~
> /home/travis/build/microsoft/LightGBM/src/c_api.cpp:381:3: note: 
>       'PredictSingleRow' declared here
>   void PredictSingleRow(int predict_type, int ncol,
>   ^
> 2 errors generated.
> CMakeFiles/_lightgbm.dir/build.make:439: recipe for target 'CMakeFiles/_lightgbm.dir/src/c_api.cpp.o' failed
> ```
Thanks @StrikerRUS ,

Sorry, my bad, I did the conflict resolution on the browser and did not double check. I solved the issues now, but now there are some unused parameters, but I guess removing them would break a lot of the interfaces.

JoanFM:
Hi @StrikerRUS @guolinke  @imatiach-msft, is there any issue blocking the merge of this PR? Thanks

guolinke:
Thanks @JoanFM so much, looks good to me. ping @StrikerRUS for the final review.

StrikerRUS:
Everything looks good to me either! Actually, I gave my approval some days ago: https://github.com/microsoft/LightGBM/pull/2760#pullrequestreview-447229824. Sorry for the confusion!

I just wonder, should new functions in C API from the recently merged #2992 be enhanced with new locks?

JoanFM:
Hey @StrikerRUS, I think he did not add any new method inside the Booster object, am I right? If it is so, that PR should directly benefit from the new locking strategy. Buy yes it is important to keep up with these locks when new methods arise, and if possible ensure const-correctness for new methods.

StrikerRUS:
@JoanFM Yeah, seems you are right!

Thanks a lot for your contribution and a great patience!

jameslamb:
> yamc seems to be a pretty bad implementation with regards to the way it manages shared locks

@Ten0 are you interested in submitting a pull request changing the behavior you're talking about? Or writing up in more detail a proposal for some change that someone else could implement?

We'd welcome your help in improving LightGBM if you see some opportunity for improvement.

Ten0:
> Are you interested in submitting a pull request changing the behavior you're talking about?

Hello! Thank you for your interest!
Indeed I might be! I'm currently working towards a benchmark for #6024 (which should be feature-complete otherwise), and I'll make sure to specifically benchmark the impact of the sub-optimal locking. If it is large, I may work on this next. However if it turns out that running the decision tree is expensive enough that the sub-optimal locking becomes negligible and allowing parallelism from #6024 is enough, I probably won't prioritize it.

Ten0:
I've finished the benchmarks and as it turns out, running our decision tree is expensive enough that the sub-optimal locking becomes negligible and allowing parallelism from https://github.com/microsoft/LightGBM/pull/6024 is totally enough for us ATM, so I won't dedicate time to this.
image

PR #2767, opened 2020-02-17
[R-package] centralize Dataset parameter aliases and add tests on parameter updating

jameslamb:
@StrikerRUS @guolinke please see this PR for my proposed R changes for #2594


jameslamb:
The failures on Azure don't look related to this PR

image

PR #2776, opened 2020-02-19
remove init-score parameter 

### LINKED ISSUE #2773: remove init_score params
Refer to https://github.com/microsoft/LightGBM/pull/2594#discussion_r376681859.
> Is it possible to update code for `init_scores` to match described behavior in docs?

> another solution is to force the init_score_filename to `concat(data_filename, ".init")`, and remove all initscore filename parameters.

guolinke:
to fix #2773

### DIFF
--- a/docs/Parameters.rst
+++ b/docs/Parameters.rst
@@ -591,24 +591,6 @@ IO Parameters
 
    -  **Note**: can be used only in CLI version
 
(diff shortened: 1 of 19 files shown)


StrikerRUS:
@guolinke I removed mentions of `init_score` params from example configs and docs in the latest commit.

PR #2779, opened 2020-02-19
[docs] Added subsections into param list

### LINKED ISSUE #2772: make the breakage in docs into Dataset and Booster params
Refer to https://github.com/microsoft/LightGBM/pull/2594#discussion_r373819109.
> Can we have a subsection under `IO` params named `Dataset` (or somehow better) params? Yes, we have no `Booster`/`Dataset` abstraction at CLI level, but with this PR distinguishing becomes very important. Without such info, users have to understand where are Dataset params only via fatal errors they get.
Or maybe we should add a **Note** to each Dataset parameter about that it cannot be changed after dataset construction.

> I agree with the subsection for dataset object, it will be much clearer.

StrikerRUS:
Fixed #2772.

Only reordering was made.

@guolinke Please check the content of new subsections.


include/LightGBM/config.h
```diff
@@ -478,22 +474,62 @@ struct Config {
...
+
   // alias = verbose
   // desc = controls the level of LightGBM's verbosity
   // desc = ``< 0``: Fatal, ``= 0``: Error (Warning), ``= 1``: Info, ``> 1``: Debug
   int verbosity = 1;
 
+  // alias = hist_pool_size
+  // desc = max cache size in MB for historical histogram
+  // desc = ``< 0`` means no limit
+  double histogram_pool_size = -1.0;
```
StrikerRUS: Maybe move these params (or at least only `histogram_pool_size`) into `Learning Control` section?

guolinke:
I think we should merge #2776 first

include/LightGBM/config.h
guolinke: agree move `histogram_pool_size` to Learning Control

include/LightGBM/config.h
```diff
@@ -478,22 +474,62 @@ struct Config {
...
+  // desc = if ``true``, LightGBM will save the dataset (including validation data) to a binary file. This speed ups the data loading for the next time
+  // desc = **Note**: ``init_score`` is not saved in binary file
+  // desc = **Note**: can be used only in CLI version; for language-specific packages you can use the correspondent function
+  bool save_binary = false;
+
+  // alias = save_period
+  // desc = frequency of saving model file snapshot
+  // desc = set this to positive value to enable this function. For example, the model file will be snapshotted at each iteration if ``snapshot_freq=1``
+  // desc = **Note**: can be used only in CLI version
+  int snapshot_freq = -1;
```
guolinke: to Learning Control

include/LightGBM/config.h
```diff
@@ -478,22 +474,62 @@ struct Config {
...
+  // alias = model_output, model_out
+  // desc = filename of output model in training
+  // desc = **Note**: can be used only in CLI version
+  std::string output_model = "LightGBM_model.txt";
+
+  // alias = is_save_binary, is_save_binary_file
+  // desc = if ``true``, LightGBM will save the dataset (including validation data) to a binary file. This speed ups the data loading for the next time
+  // desc = **Note**: ``init_score`` is not saved in binary file
+  // desc = **Note**: can be used only in CLI version; for language-specific packages you can use the correspondent function
+  bool save_binary = false;
```
guolinke: I think this is better in "Dataset" section.

include/LightGBM/config.h
```diff
@@ -478,22 +474,62 @@ struct Config {
...
+  // desc = path(s) of file(s) with validation initial scores
+  // desc = if ``""``, will use ``valid_data_file`` + ``.init`` (if exists)
+  // desc = separate by ``,`` for multi-validation data
+  // desc = **Note**: works only in case of loading data directly from file
+  std::vector<std::string> valid_data_initscores;
+
   // alias = verbose
   // desc = controls the level of LightGBM's verbosity
   // desc = ``< 0``: Fatal, ``= 0``: Error (Warning), ``= 1``: Info, ``> 1``: Debug
   int verbosity = 1;
```
guolinke: to learning control

include/LightGBM/config.h
guolinke: except `save_binary`, I think others could be moved to learning control.

StrikerRUS:
@guolinke Please give it another round of review.

docs/Parameters.rst
```diff
@@ -190,6 +190,36 @@ Core Parameters
 Learning Control Parameters
 ---------------------------
 
+-  ``input_model`` :raw-html:`<a id="input_model" title="Permalink to this parameter" href="#input_model">&#x1F517;&#xFE0E;</a>`, default = ``""``, type = string, aliases: ``model_input``, ``model_in``
```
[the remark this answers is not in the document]
StrikerRUS: @guolinke I moved CLI-only params to the bottom of section: https://github.com/microsoft/LightGBM/blob/docs_gen/docs/Parameters.rst. Please let me know if you think more reordering is needed.

PR #2785, opened 2020-02-20
[R-package] Fixed R implementation of upper_bound() and lower_bound() for lgb.Booster

jameslamb:
@JoanFM thanks for contributing #2737 ! Unfortunately, there were some issues in the R side. This PR attempts to address them:

* removes trailing  `_` from method names
* fixes bug where `lower_bound()` method was storing its return value in variable `upper_bound`
* adds unit tests to ensure the implementation is working
* added new functions to `lightgbm_R.h` (without that, `LGBM_BoosterGetUpperBoundValue_R` and  `LGBM_BoosterGetLowerBoundValue_R` are not callable from R)

I still need some help though @guolinke @JoanFM @StrikerRUS .... I am not getting the answers  I'd expect. For example, I would expect lower bound to be 0 and upper bound to be 1  for binary classification, but running this:

```r
train <- agaricus.train
test <- agaricus.test
nrounds <- 10L
  bst <- lightgbm(
    data = train$data
    , label = train$label
    , num_leaves = 5L
    , nrounds = nrounds
    , objective = "binary"
    , metric = "binary_error"
  )
```

I'm confused by the results:

> > bst$lower_bound()
[1] -1950774382
> bst$upper_bound()
[1] 1196082214

I'll look back through #2737 , but maybe there is just something fundamental that I've misunderstood?

### DIFF
--- a/R-package/tests/testthat/test_basic.R
+++ b/R-package/tests/testthat/test_basic.R
@@ -71,7 +71,7 @@ test_that("use of multiple eval metrics works", {
 })
 
 test_that("lgb.Booster.upper_bound() and lgb.Booster.lower_bound() work as expected for binary classification", {
-  set.seed(708)
+  set.seed(708L)
   nrounds <- 10L
   bst <- lightgbm(
     data = train$data
@@ -87,7 +87,7 @@ test_that("lgb.Booster.upper_bound() and lgb.Booster.lower_bound() work as expec
 })
 
 test_that("lgb.Booster.upper_bound() and lgb.Booster.lower_bound() work as expected for regression", {
-  set.seed(708)
+  set.seed(708L)
   nrounds <- 10L
   bst <- lightgbm(
     data = train$data
--- a/R-package/tests/testthat/test_basic.R
+++ b/R-package/tests/testthat/test_basic.R
@@ -7,6 +7,8 @@ test <- agaricus.test
 
 windows_flag <- grepl("Windows", Sys.info()[["sysname"]])
 
+TOLERANCE <- 1e-6
+
 test_that("train and predict binary classification", {
   nrounds <- 10L
   bst <- lightgbm(
@@ -28,7 +30,7 @@ test_that("train and predict binary classification", {
   expect_equal(length(pred1), 6513L)
   err_pred1 <- sum((pred1 > 0.5) != train$label) / length(train$label)
   err_log <- record_results[1L]
-  expect_lt(abs(err_pred1 - err_log), 10e-6)
+  expect_lt(abs(err_pred1 - err_log), TOLERANCE)
 })
 
 
@@ -81,7 +83,6 @@ test_that("lgb.Booster.upper_bound() and lgb.Booster.lower_bound() work as expec
     , objective = "binary"
     , metric = "binary_error"
   )
-  TOLERANCE <- 1e-06
   expect_true(abs(bst$lower_bound() - -1.590853) < TOLERANCE)
   expect_true(abs(bst$upper_bound() - 1.871015) <  TOLERANCE)
 })
@@ -97,7 +98,6 @@ test_that("lgb.Booster.upper_bound() and lgb.Booster.lower_bound() work as expec
     , objective = "regression"
     , metric = "l2"
   )
-  TOLERANCE <- 1e-06
   expect_true(abs(bst$lower_bound() - 0.1513859) < TOLERANCE)
   expect_true(abs(bst$upper_bound() - 0.9080349) < TOLERANCE)
 })
--- a/R-package/tests/testthat/test_basic.R
+++ b/R-package/tests/testthat/test_basic.R
@@ -70,6 +70,33 @@ test_that("use of multiple eval metrics works", {
   expect_false(is.null(bst$record_evals))
 })
 
+test_that("lgb.Booster.upper_bound() and lgb.Booster.lower_bound() work as expected for binary classification", {
+  nrounds <- 10L
+  bst <- lightgbm(
+    data = train$data
+    , label = train$label
+    , num_leaves = 5L
+    , nrounds = nrounds
+    , objective = "binary"
+    , metric = "binary_error"
+  )
+  expect_true(bst$lower_bound() == 0)
+  expect_true(bst$upper_bound() == 1)
+})
+
+test_that("lgb.Booster.upper_bound() and lgb.Booster.lower_bound() work as expected for regression", {
+  bst <- lightgbm(
+    data = train$data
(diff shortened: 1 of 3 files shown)


jameslamb:
> I'm confused by the results:

> bst$lower_bound()
[1] -1950774382
bst$upper_bound()
[1] 1196082214

> I'll look back through #2737 , but maybe there is just something fundamental that I've misunderstood?

Ok I realized one thing that was causing surprising  results...the return type needs to be double, not int. Fixed in https://github.com/microsoft/LightGBM/pull/2785/commits/a0dc638c1db8aa72bac165dc5f60491289e93023

Now that same example returns values like this:

> bst$lower_bound()
[1] -1.590853
> bst$upper_bound()
[1] 1.871015

Is there something fundamental I'm missing? If the target in the training data is bounded between 0 and 1, how is it possible for something tree-based to predict a value lower than 0 or great than 1?

When I re-predict on all of the training data, I get results I'd expect

> preds <- bst$predict(data = train$data)

> summary(preds)

>   Min. 1st Qu.  Median    Mean 3rd Qu.    Max. 
 0.2046  0.2048  0.2137  0.4821  0.7934  0.7934

JoanFM:
Hello @jameslamb,  there might be an error, but there is not direct relationship between the minimmum you get in your data and the bound values, the bound values may be totally unreachable. 

These bounds are taken by looking at the min and max values of every tree's leaves and adding them up. But they are most likely not reachable values. They serve as a lower or upper bound that one can be 100% sure that the model will not surpass without taking a look at any kind of data.

I hope it helps

jameslamb:
> Hello @jameslamb, there might be an error, but there is not direct relationship between the minimmum you get in your data and the bound values, the bound values may be totally unreachable.
> 
> These bounds are taken by looking at the min and max values of every tree's leaves and adding them up. But they are most likely not reachable values. They serve as a lower or upper bound that one can be 100% sure that the model will not surpass without taking a look at any kind of data.
> 
> I hope it helps

Thanks, Joan. Maybe there's something I've misunderstood, I still don't get how a tree-based  model that has only ever seen data between 0 and 1 could ever predict a value outside of that range even theoretically, since the leaf values are taken by voting or averaging (depending on task). Right?

JoanFM:
> > Hello @jameslamb, there might be an error, but there is not direct relationship between the minimmum you get in your data and the bound values, the bound values may be totally unreachable.
> > These bounds are taken by looking at the min and max values of every tree's leaves and adding them up. But they are most likely not reachable values. They serve as a lower or upper bound that one can be 100% sure that the model will not surpass without taking a look at any kind of data.
> > I hope it helps
> 
> Thanks, Joan. Maybe there's something I've misunderstood, I still don't get how a tree-based model that has only ever seen data between 0 and 1 could ever predict a value outside of that range even theoretically, since the leaf values are taken by voting or averaging (depending on task). Right?

hey James, it does not need a theoretical value. It is just equivalent to parse the tree, and take all the time the min or max of leaf_values, and add them up. Imagine:
tree 0
leafs -1, 1

tree 1
leafs 0.5, -0.5.

The upper and lower bounds would be -1.5 and 1.5. But maybe these values are not reachable, they are just a conservative bound.

jameslamb:
> > > Hello @jameslamb, there might be an error, but there is not direct relationship between the minimmum you get in your data and the bound values, the bound values may be totally unreachable.
> > > These bounds are taken by looking at the min and max values of every tree's leaves and adding them up. But they are most likely not reachable values. They serve as a lower or upper bound that one can be 100% sure that the model will not surpass without taking a look at any kind of data.
> > > I hope it helps
> > 
> > 
> > Thanks, Joan. Maybe there's something I've misunderstood, I still don't get how a tree-based model that has only ever seen data between 0 and 1 could ever predict a value outside of that range even theoretically, since the leaf values are taken by voting or averaging (depending on task). Right?
> 
> hey James, it does not need a theoretical value. It is just equivalent to parse the tree, and take all the time the min or max of leaf_values, and add them up. Imagine:
> tree 0
> leafs -1, 1
> 
> tree 1
> leafs 0.5, -0.5.
> 
> The upper and lower bounds would be -1.5 and 1.5. But maybe these values are not reachable, they are just a conservative bound.

Thanks @JoanFM . I was thinking about this the wrong way.  Fixed the tests in https://github.com/microsoft/LightGBM/pull/2785/commits/ec9a941c415763bfaa65b553e0ac6d9999681654

I think we are good!

JoanFM:
> > > > Hello @jameslamb, there might be an error, but there is not direct relationship between the minimmum you get in your data and the bound values, the bound values may be totally unreachable.
> > > > These bounds are taken by looking at the min and max values of every tree's leaves and adding them up. But they are most likely not reachable values. They serve as a lower or upper bound that one can be 100% sure that the model will not surpass without taking a look at any kind of data.
> > > > I hope it helps
> > > 
> > > 
> > > Thanks, Joan. Maybe there's something I've misunderstood, I still don't get how a tree-based model that has only ever seen data between 0 and 1 could ever predict a value outside of that range even theoretically, since the leaf values are taken by voting or averaging (depending on task). Right?
> > 
> > 
> > hey James, it does not need a theoretical value. It is just equivalent to parse the tree, and take all the time the min or max of leaf_values, and add them up. Imagine:
> > tree 0
> > leafs -1, 1
> > tree 1
> > leafs 0.5, -0.5.
> > The upper and lower bounds would be -1.5 and 1.5. But maybe these values are not reachable, they are just a conservative bound.
> 
> Thanks @JoanFM . I was thinking about this the wrong way. Fixed the tests in [ec9a941](https://github.com/microsoft/LightGBM/commit/ec9a941c415763bfaa65b553e0ac6d9999681654)
> 
> I think we are good!

you are welcome, thank you and sorry for the screw up with the R-package in the previous PR

jameslamb:
> you are welcome, thank you and sorry for the screw up with the R-package in the previous PR

No problem! My fault for not giving you a review sooner. Thanks again for the contributions!

R-package/tests/testthat/test_basic.R
```diff
@@ -70,6 +70,38 @@ test_that("use of multiple eval metrics works", {
...
+  nrounds <- 10L
+  bst <- lightgbm(
+    data = train$data
+    , label = train$label
+    , num_leaves = 5L
+    , nrounds = nrounds
+    , objective = "binary"
+    , metric = "binary_error"
+  )
+  TOLERANCE <- 1e-06
```
StrikerRUS: Maybe make `TOLERANCE` global const?

jameslamb: sure, makes sense

Issue #2798, opened 2020-02-22
Wrong implementation of SingleRowPredictor::IsPredictorEqual in c_api


AlbertoEAF:
Hello,

I was looking at the code and found this definition in the c_api.cpp:

```
  bool IsPredictorEqual(const Config& config, int iter, Boosting* boosting) {
    return early_stop_ != config.pred_early_stop ||
      early_stop_freq_ != config.pred_early_stop_freq ||
      early_stop_margin_ != config.pred_early_stop_margin ||
      iter_ != iter ||
      num_total_model_ != boosting->NumberOfTotalModel();
  }
```

Shouldn't it be `return early_stop_ == config.pred_early_ctop && ...` ?

This is then used in PredictSingleRow(...) to reset the predictor in case the predictor is not the same:

```
    if (single_row_predictor_[predict_type].get() == nullptr ||
        !single_row_predictor_[predict_type]->IsPredictorEqual(config, num_iteration, boosting_.get())) {
      single_row_predictor_[predict_type].reset(new SingleRowPredictor(predict_type, boosting_.get(),
                                                                       config, num_iteration));
    }
```

which will result in creating a new SingleRowPredictor every time, and worse, not creating a new one if conditions change.

This worsens the performance for repeated calls, and results in not creating a new predictor in case parameters change (bug).

If you want I can gladly do a pull request to fix it.

guolinke:
fixed in #2799

PR #2799, opened 2020-02-22
Fix SingleRowPredictor::IsPredictorEqual comparison (invert)

AlbertoEAF:
This fixes the issue https://github.com/microsoft/LightGBM/issues/2798.

As stated in that issue, this should fix the bug in case the parameters change, and improve performance when they don't.

Please give me feedback if I'm doing anything in a less than desirable way.

### DIFF
--- a/src/c_api.cpp
+++ b/src/c_api.cpp
@@ -77,13 +77,15 @@ class SingleRowPredictor {
     predict_function = predictor_->GetPredictFunction();
     num_total_model_ = boosting->NumberOfTotalModel();
   }
+
   ~SingleRowPredictor() {}
+
   bool IsPredictorEqual(const Config& config, int iter, Boosting* boosting) {
-    return early_stop_ != config.pred_early_stop ||
-      early_stop_freq_ != config.pred_early_stop_freq ||
-      early_stop_margin_ != config.pred_early_stop_margin ||
-      iter_ != iter ||
-      num_total_model_ != boosting->NumberOfTotalModel();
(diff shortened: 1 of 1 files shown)


guolinke:
ping @imatiach-msft to confirm

imatiach-msft:
@AlbertoEAF great catch, this should improve performance during scoring.  The main bugfix that introduced this bug was changing the correctness from:
(single_row_predictor_.get() == nullptr)
to:
single_row_predictor_[predict_type].get() == nullptr
so that when calling raw predict and predict for probabilities the function would return the correct values, but it looks like now we are not re-using the cached predictor which is horrible.  Sorry I introduced this performance issue after fixing the correctness issue before.

"worse, not creating a new one if conditions change"
I don't think this can happen from someone using mmlspark which is calling those APIs (since the model is immutable after it is trained) but this is definitely something that could help anyone who is using the native API directly, although I'm not aware of anyone other than mmlspark using LGBM_BoosterPredictForMatSingleRow and LGBM_BoosterPredictForCSRSingleRow.  Regardless this is a great fix, and thank you for sending it out!

imatiach-msft:
also adding @eisber for context who added the original optimizations, it looks like his optimizations of improving scoring by 3-4X are missing in latest mmlspark

AlbertoEAF:
Thank you @matiach-smft :)

Regarding no one else using the native API, I'm working on a solution that uses it for high throughput low latency scoring. If I can help a bit this project great, I'd like to learn and help more :)

PR #2800, opened 2020-02-23
Make timer thread-safe

guolinke:


guolinke:
It seems windows will stuck, will debug on when have time.

PR #2803, opened 2020-02-23
[R-package] [ci] added more linters

jameslamb:
This PR adds a few more R linters:

* `absolute_path_linter`: prevent paths like `~/repos/`
* `nonportable_path_linter`: prevent paths like `C:\Users\` which only work on certain operating systems
* `undesirable_function_linter`: prevent the use of some functions which could cause issues, specifically:
    - `dyn.load()` / `dyn.unload()`: direct calls to use symbols in `.dll` / `.so` files. My expectation is that someone using this in a PR either doesn't understand how to write R packages that call C++ code or is introducing something that will be very difficult to maintain.
    - `help()`: call up help documentation. This should only be used interactively, not in code checked into this repository.
    - `install.packages()`: install new libraries, e.g. from CRAN. This should only be used interactively, not in code checked into this repository.
    - `ifelse()`: This is a convenient but dangerous function, because its output type depends on whether or not the input is all-TRUE, all-FALSE, or a mix.
        * image
    - `rbind()`: stack two data frames on top of each other (similar to `pd.concat()` in `pandas`). `data.table::rbindlist()` is faster and safer in the case where the data frames have different columns
    - `is.list()`: check if an object is a list. I found that in all uses in `lightgbm`, we are checking if something is exactly a list, not "list-like". `is.list()` will be `TRUE` on a `data.table` object, so using `identical(class(x), "list")` is safer.
    - `require()`: load a package if it hasn't been loaded yet and raise a warning if it is not available. `library()` is preferred, since hitting a loud error right away that says "this package does not exist" makes debugging easier than getting some of error like "function 'something' not found" further down in the code
* `undesirable_operator_linter`: prevent the use of certain operators in R code, specifically:
    - `%>%`, `%.%`, `%..%`: variations of "pipe" operators, commonly used in some R packages. The `lightgbm` package doesn't currently use this programming model, and preventing it with this linter is about preventing a mix of different programming styles in the codebase.
    - `?`: call up help documentation. This should only be used interactively, not in code checked into this repository.

This PR also re-ordered the list of linters in `.ci/lint_r_code.R` to alphabetize them for readability.

### DIFF
--- a/.ci/lint_r_code.R
+++ b/.ci/lint_r_code.R
@@ -29,31 +29,32 @@ interactive_text <- paste0(
 )
 
 LINTERS_TO_USE <- list(
-    "absolute_path" = lintr::absolute_path_linter
-    , "assignment" = lintr::assignment_linter
-    , "closed_curly" = lintr::closed_curly_linter
-    , "commas" = lintr::commas_linter
-    , "equals_na" = lintr::equals_na_linter
-    , "function_left" = lintr::function_left_parentheses_linter
-    , "implicit_integers" = lintr::implicit_integer_linter
-    , "infix_spaces" = lintr::infix_spaces_linter
-    , "long_lines" = lintr::line_length_linter(length = 120L)
-    , "tabs" = lintr::no_tab_linter
-    , "non_portable_path" = lintr::nonportable_path_linter
-    , "open_curly" = lintr::open_curly_linter
-    , "paren_brace_linter" = lintr::paren_brace_linter
-    , "semicolon" = lintr::semicolon_terminator_linter
-    , "seq" = lintr::seq_linter
-    , "single_quotes" = lintr::single_quotes_linter
-    , "spaces_inside" = lintr::spaces_inside_linter
-    , "spaces_left_parens" = lintr::spaces_left_parentheses_linter
-    , "todo_comments" = lintr::todo_comment_linter(c("todo", "fixme", "to-do"))
-    , "trailing_blank" = lintr::trailing_blank_lines_linter
-    , "trailing_white" = lintr::trailing_whitespace_linter
-    , "true_false" = lintr::T_and_F_symbol_linter
+    "absolute_path"          = lintr::absolute_path_linter
+    , "assignment"           = lintr::assignment_linter
+    , "closed_curly"         = lintr::closed_curly_linter
+    , "commas"               = lintr::commas_linter
+    , "equals_na"            = lintr::equals_na_linter
+    , "function_left"        = lintr::function_left_parentheses_linter
+    , "implicit_integers"    = lintr::implicit_integer_linter
+    , "infix_spaces"         = lintr::infix_spaces_linter
+    , "long_lines"           = lintr::line_length_linter(length = 120L)
+    , "no_tabs"              = lintr::no_tab_linter
+    , "non_portable_path"    = lintr::nonportable_path_linter
+    , "open_curly"           = lintr::open_curly_linter
+    , "paren_brace_linter"   = lintr::paren_brace_linter
+    , "semicolon"            = lintr::semicolon_terminator_linter
+    , "seq"                  = lintr::seq_linter
+    , "single_quotes"        = lintr::single_quotes_linter
+    , "spaces_inside"        = lintr::spaces_inside_linter
+    , "spaces_left_parens"   = lintr::spaces_left_parentheses_linter
+    , "todo_comments"        = lintr::todo_comment_linter(c("todo", "fixme", "to-do"))
+    , "trailing_blank"       = lintr::trailing_blank_lines_linter
+    , "trailing_white"       = lintr::trailing_whitespace_linter
+    , "true_false"           = lintr::T_and_F_symbol_linter
     , "undesirable_function" = lintr::undesirable_function_linter(
         fun = c(
-            "dyn.load" = "Directly loading/unloading .dll/.so files in package code should not be necessary."
+            "cbind" = "cbind() is an unsafe way to build up a data frame. merge() or direct column assignment is preferred"
+            , "dyn.load" = "Directly loading/unloading .dll/.so files in package code should not be necessary."
             , "dyn.unload" = "Directly loading/unloading .dll/.so files in package code should not be necessary."
             , "help" = interactive_text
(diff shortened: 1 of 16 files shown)


R-package/tests/testthat/test_parameters.R
```diff
@@ -48,7 +48,7 @@ context("parameter aliases")
 
 test_that(".PARAMETER_ALIASES() returns a named list of character vectors, where names are unique", {
   param_aliases <- .PARAMETER_ALIASES()
-  expect_true(is.list(param_aliases))
+  expect_true(identical(class(param_aliases), "list"))
```
StrikerRUS:
```suggestion
  expect_identical(class(param_aliases), "list")
```
For the consistency with other checks.

jameslamb: ok yeah good catch

jameslamb:
@Laurae2 can you take a look when you have a chance? I'm making the R code more opinionated here, would like to be sure you agree with it before merging.

jameslamb:
gently ping @Laurae2 for a review

.ci/lint_r_code.R
```diff
@@ -16,17 +16,30 @@ FILES_TO_LINT <- list.files(
...
     , "closed_curly" = lintr::closed_curly_linter
+    , "commas" = lintr::commas_linter
     , "equals_na" = lintr::equals_na_linter
     , "function_left" = lintr::function_left_parentheses_linter
-    , "commas" = lintr::commas_linter
-    , "concatenation" = lintr::unneeded_concatenation_linter
     , "implicit_integers" = lintr::implicit_integer_linter
     , "infix_spaces" = lintr::infix_spaces_linter
     , "long_lines" = lintr::line_length_linter(length = 120L)
     , "tabs" = lintr::no_tab_linter
```
Laurae2: Follow alphabet order

Laurae2:
@jameslamb I looked thoroughly and I agree with the changes.

Do you think we should also add `cbind()` to `undesirable_function_linter`?

.ci/lint_r_code.R
jameslamb: ah yeah! Good catch

jameslamb:
> @jameslamb I looked thoroughly and I agree with the changes.
> 
> Do you think we should also add `cbind()` to `undesirable_function_linter`?

Yep I agree! I'll do that

jameslamb:
> > @jameslamb I looked thoroughly and I agree with the changes.
> > Do you think we should also add `cbind()` to `undesirable_function_linter`?
> 
> Yep I agree! I'll do that

@Laurae2 I had to change one of the demos where we were using `cbind()`. Could you take a look and let me know if it's ok? https://github.com/microsoft/LightGBM/pull/2803/commits/b60e94b2545b8b817919783d7099b58840512559

.ci/lint_r_code.R
```diff
@@ -16,28 +16,73 @@ FILES_TO_LINT <- list.files(
     , include.dirs = FALSE
 )
 
+# text to use for pip operators from packages like 'magrittr'
```
jameslamb:
```suggestion
# text to use for pipe operators from packages like 'magrittr'
```

jameslamb:
I just rebased to `master` to catch up with recent changes.

.ci/lint_r_code.R
```diff
@@ -16,28 +16,73 @@ FILES_TO_LINT <- list.files(
...
+    , "spaces_left_parens"   = lintr::spaces_left_parentheses_linter
+    , "todo_comments"        = lintr::todo_comment_linter(c("todo", "fixme", "to-do"))
+    , "trailing_blank"       = lintr::trailing_blank_lines_linter
+    , "trailing_white"       = lintr::trailing_whitespace_linter
+    , "true_false"           = lintr::T_and_F_symbol_linter
+    , "undesirable_function" = lintr::undesirable_function_linter(
+        fun = c(
+            "cbind" = paste0(
+                "cbind is an unsafe way to build up a data frame. merge() or direct "
+                , "column assignment is preferred"
```
StrikerRUS:
```suggestion
                , "column assignment is preferred."
```

.ci/lint_r_code.R
```diff
@@ -16,28 +16,73 @@ FILES_TO_LINT <- list.files(
...
+            , "ifelse" = "The use of ifelse() is dangerous because it will silently allow mixing types."
+            , "install.packages" = interactive_text
+            , "is.list" = paste0(
+                "This project uses data.table, and is.list(x) is TRUE for a data.table. "
+                , "identical(class(x), 'list') is a safer way to check that something is an R list object."
+            )
+            , "rbind" = "data.table::rbindlist() is faster and safer than rbind(), and is preferred in this project."
+            , "require" = paste0(
+                "library() is preferred to require() because it will raise an error immediately "
+                , "if a package is missing"
```
StrikerRUS:
```suggestion
                , "if a package is missing."
```

jameslamb:
> nit.
> Dots for the consistency with other descriptions.

no problem, thanks!

by the way I just pulled this and rebased it to current `master`. None of the changes on `master` fail these new linting rules, so we can merge this as-is (without force pushing the rebased version of this branch I have locally) once it builds

StrikerRUS:
@jameslamb Did you see this warning:
```
* checking R code for possible problems ... NOTE
multiple.tree.plot.interpretation: no visible binding for global
  variable ‘bar_color’
  bar_color
```

I guess it's from this PR.

jameslamb:
> @jameslamb Did you see this warning:
> 
> ```
> * checking R code for possible problems ... NOTE
> multiple.tree.plot.interpretation: no visible binding for global
>   variable ‘bar_color’
> Undefined global functions or variables:
>   bar_color
> ```
> 
> I guess it's from this PR.

@StrikerRUS I did see that a bit after we merged, working on #2901  I was trying to avoid any more R PRs since we have so many open right now, but I think this one would be easy to fix. This comes from the fact that `data.table` uses non-standard evaluation. so ` tree_interpretation[, bar_color]` (where `tree_interpretation` is a `data.table`) will work correctly, but static analysis of the code makes it look like a variable that doesn't exist is being referenced.

I'll submit a PR right now

PR #2804, opened 2020-02-23
decouple bagging with num_threads

guolinke:
The bagging results should be identical whatever num_threads is.


guolinke:
@StrikerRUS @Laurae2 
I remember we have some documents address this, could you help to remove the related descriptions?

StrikerRUS:
@guolinke 
> I remember we have some documents address this, could you help to remove the related descriptions?

Done!

Issue #1597, opened 2018-08-20
For early stopping, at least one dataset and eval metric is required for evaluation


Drxan:
## Environment info

Operating System:ubuntu16.04 amd64

CPU: 12  Intel(R) Core(TM) i7-6800K CPU @ 3.40GHz

C++/Python/R version:Python 3.6.5 :: Anaconda, Inc.

## Error message
Traceback (most recent call last):
  File "/home/yw/study/competition/datagrand/hehe.py", line 223, in <module>
    feat_imp=modelfit(lgbc,x,y,cat_vars,cvparams,feature_names,useTrainCV=True,cv_folds=5,early_stopping_rounds=30, metric='f1_macpro')
  File "/home/yw/study/competition/datagrand/hehe.py", line 113, in modelfit
    verbose_eval=20, show_stdv=True, seed=9)
  File "/home/yw/anaconda3/lib/python3.6/site-packages/lightgbm/engine.py", line 459, in cv
    evaluation_result_list=res))
  File "/home/yw/anaconda3/lib/python3.6/site-packages/lightgbm/callback.py", line 197, in callback
    init(env)
  File "/home/yw/anaconda3/lib/python3.6/site-packages/lightgbm/callback.py", line 178, in init
    raise ValueError('For early stopping, at least one dataset and eval metric is required for evaluation')
ValueError: For early stopping, at least one dataset and eval metric is required for evaluation

## Reproducible examples

    from sklearn.datasets import make_classification
    x,y=make_classification(n_samples=1000,n_redundant=0, n_features=20, n_informative=20, n_classes=19,random_state=9)
    x = pd.DataFrame(x,columns=['feat_'+str(i) for i in range(x.shape[1])])

    def modelfit(alg, X, y, cat_features, cvparams, feature_names, folds=None, useTrainCV=True, cv_folds=5,
             early_stopping_rounds=30, metric=metric):
    # feature_names=list(X.columns)
        if useTrainCV == True:
            dtrain = lgb.Dataset(X.values, label=y, reference=None,
                             weight=None, group=None, silent=False, feature_name=feature_names,
                             categorical_feature=cat_features, params=None, free_raw_data=True)
            dtrain.construct()
            cvresult = lgb.cv(cvparams, dtrain, num_boost_round=5000, folds=folds, nfold=cv_folds, stratified=True,
                          metrics=metric, feature_name=feature_names, categorical_feature=cat_features,
                          early_stopping_rounds=early_stopping_rounds,
[...]

PR #2815, opened 2020-02-24
Add lock bot

guolinke:


guolinke:
BTW, there seems are some interesting bot for linting:
like https://github.com/marketplace/actions/autopep8
Maybe we can use github actions to auto format codes and documentation generation. 
ping @StrikerRUS and @jameslamb

StrikerRUS:
Why do we need locking old issues? Even after closing users may leave useful comments for others.

Speaking about pep8 bot, we already have configured CI linting job which tells contributors what should be fixed before PR will be approved. 
https://github.com/microsoft/LightGBM/blob/224b8b98e38f41e8ef3c7287e6097d4afe0b816a/.ci/test.sh#L43-L60

guolinke:
@StrikerRUS 
Actually, for the old issues, I think replying to them is easy to be ignored. As they are closed, I think they are solved at that time point. If the issues happen again, it may cause by different reasons. 

For some important issues, like feature requests, bug, we can exclude them to be auto locked.
What do you think?

As for the bots, I want the "auto" methods, like auto format the codes and generate the documents, without we run some scripts or fix them one-by-one manually.

StrikerRUS:
@guolinke I see your point! And I absolutely agree with you that in some cases locking old issues is required to prevent misleading comments. Refer for example to #1369. However, some other issues can benefit from new comments, e.g. #1597.

According to the [present set of created labels](https://github.com/microsoft/LightGBM/labels), I think we should exclude `blocking`, `bug`, `effectiveness`, `efficiency`, `feature-request`, `help-wanted`, `in progress`.


> As for the bots, I want the "auto" methods, like auto format the codes and generate the documents, without we run some scripts or fix them one-by-one manually.

I remember we had some discussion about docs auto-generation and inconvenience for contributors some time ago: #1511.
For auto-linting bots, where they will create PRs: into our repository after merging a PR with linting issues, or into contributor's fork when they submit a PR with linting issues?

guolinke:
Refer to autopep8:

> This action is designed to be used in conjunction with Create Pull Request. This will automatically create a pull request to merge fixes that autopep8 makes to python code in your repository.

It seems will create a new PR for the formatting. 

Personally, I think auto-format the code in the original PR is better, without an additional PR, but i don't know could it be supported or not.

StrikerRUS:
@guolinke 
> Personally, I think auto-format the code in the original PR is better (new PR to contributors' repo is okay), but i don't know could it be supported or not.

In case of new PR, we will consciously merge "bad" code, and for some time the repo will be in undesirable state. Also, I'm not sure that auto-fixes are better than manual ones. Sometimes they may be over-strict or strange, and we won't be able to review them, only propose fixes for fixes in one more PR.

jameslamb:
> @guolinke
> 
> > Personally, I think auto-format the code in the original PR is better (new PR to contributors' repo is okay), but i don't know could it be supported or not.
> 
> In case of new PR, we will consciously merge "bad" code, and for some time the repo will be in undesirable state. Also, I'm not sure that auto-fixes are better than manual ones. Sometimes they may be over-strict or strange, and we won't be able to review them, only propose fixes for fixes in one more PR.

^ I agree with this perspective. Every merge to master should leave the repo in a better state  that it was as of the previous commit. And I agree that `black` might become yet another thing to manage,  and add a new barrier to entry for contributors.

I am ok with having a bot lock old issues though! We as admins can always manually re-open issues that are locked if necessary, and GitHub's linking via comments means we won't lose the flow  of conversation someone can easily open a new issue and say "I am experiencing  something that looks similar to issue #whatever".

So to summarize,  here is where I stand:

* `black` or other auto-formatting bot: 👎 
* `lock-bot`: 👍

guolinke:
@StrikerRUS @jameslamb thanks, got it. Let us only enable lock bot for now.

PR #2816, opened 2020-02-24
replace std::runtime_error with Log::Fatal

StrikerRUS:


guolinke:
@StrikerRUS It seems this PR produces some warnings when complied in VS, debug mode.
```
1>D:\src\LightGBM\src\c_api.cpp(1806): warning C4715: 'RowFunctionFromDenseMatric': not all control paths return a value
1>D:\src\LightGBM\src\c_api.cpp(1910): warning C4715: 'RowFunctionFromCSR': not all control paths return a value
1>D:\src\LightGBM\src\c_api.cpp(1975): warning C4715: 'IterateFunctionFromCSC': not all control paths return a value
```

StrikerRUS:
@guolinke 
> debug mode.

Ah, that's why I haven't spot this issue with Azure Pipelines!

Don't you know, will dummy `return` similarly to it in objective function code fix the problem?
https://github.com/microsoft/LightGBM/blob/e676af236625fcfc548ca60d0b89f360c736a744/src/objective/objective_function.cpp#L51-L52

guolinke:
@StrikerRUS yeah, the explicitly return can eliminate the warnings.

PR #2463, opened 2019-09-27
Add CTR conversion for categorical features (with bundling support)

shiyu1994:
This version keeps the original categorical split decisions in trees. We will change to pure numerical decisions in trees soon by recording the mappings from categorical values to CTR values in model. The modification so far only handles the ordinary train and prediction process of LightGBM, other functions like ForceSplits requires further check.

PR #2824, opened 2020-02-26
avoid most_freq_bin to be 0 in categorical features

### LINKED ISSUE #2742: LightGBMError: Check failed: best_split_info.left_count > 0 for ranking task
hi, 

when would lightgbm 2.3.2 be released? as the documentation online (https://lightgbm.readthedocs.io/) is 2.3.2 with some new features but the latest python package is lightgbm 2.3.1. Thanks.

guolinke:
to fix #2742

### DIFF
--- a/src/io/bin.cpp
+++ b/src/io/bin.cpp
@@ -510,20 +510,24 @@ namespace LightGBM {
 
     if (!is_trivial_) {
       default_bin_ = ValueToBin(0);
+      most_freq_bin_ =
+          static_cast<uint32_t>(ArrayArgs<int>::ArgMax(cnt_in_bin));
       if (bin_type_ == BinType::CategoricalBin) {
-        CHECK(default_bin_ > 0);
+        if (most_freq_bin_ == 0) {
+          CHECK(num_bin_ > 1);
+          // FIXME: how to enable `most_freq_bin_ = 0` for categorical features
+          most_freq_bin_ = 1;
+        }
       }
-    }
-    if (!is_trivial_) {
-      most_freq_bin_ = static_cast<uint32_t>(ArrayArgs<int>::ArgMax(cnt_in_bin));
-      // calculate sparse rate
-      sparse_rate_ = static_cast<double>(cnt_in_bin[default_bin_]) / total_sample_cnt;
-      const double max_sparse_rate = static_cast<double>(cnt_in_bin[most_freq_bin_]) / total_sample_cnt;
-      if (most_freq_bin_ != default_bin_ && max_sparse_rate > 0.7f) {
-        sparse_rate_ = max_sparse_rate;
-      } else {
+      const double max_sparse_rate =
+          static_cast<double>(cnt_in_bin[most_freq_bin_]) / total_sample_cnt;
+      // When most_freq_bin_ != default_bin_, there are some additional data loading costs.
+      // so use most_freq_bin_  = default_bin_ when there is not so sparse
+      if (most_freq_bin_ != default_bin_ && max_sparse_rate < 0.7f) {
         most_freq_bin_ = default_bin_;
       }
+      sparse_rate_ =
+          static_cast<double>(cnt_in_bin[most_freq_bin_]) / total_sample_cnt;
     } else {
       sparse_rate_ = 1.0f;
     }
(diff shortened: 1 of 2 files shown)


src/io/bin.cpp
```diff
@@ -510,20 +510,24 @@ namespace LightGBM {
...
-      sparse_rate_ = static_cast<double>(cnt_in_bin[default_bin_]) / total_sample_cnt;
-      const double max_sparse_rate = static_cast<double>(cnt_in_bin[most_freq_bin_]) / total_sample_cnt;
-      if (most_freq_bin_ != default_bin_ && max_sparse_rate > 0.7f) {
-        sparse_rate_ = max_sparse_rate;
-      } else {
+      const double max_sparse_rate =
+          static_cast<double>(cnt_in_bin[most_freq_bin_]) / total_sample_cnt;
+      // When most_freq_bin_ != default_bin_, there are some additional data loading costs.
+      // so use most_freq_bin_  = default_bin_ when there is not so sparse
+      if (most_freq_bin_ != default_bin_ && max_sparse_rate < 0.7f) {
```
guolinke:
```suggestion
      if (most_freq_bin_ != default_bin_ && max_sparse_rate <= 0.7f) {
```

tests/python_package_test/test_engine.py
```diff
@@ -335,6 +335,43 @@ def test_categorical_handle_na(self):
...
+        x = [1, 1, 1, 1, 1, 1, 2, 2]
+        y = [1, 1, 1, 1, 1, 1, 0, 0]
+
+        X_train = np.array(x).reshape(len(x), 1)
+        y_train = np.array(y)
+        lgb_train = lgb.Dataset(X_train, y_train)
+        lgb_eval = lgb.Dataset(X_train, y_train)
+
+        params = {
+            'objective': 'binary',
```
guolinke:
```suggestion
            'objective': 'regression',
```

src/io/bin.cpp
```diff
@@ -510,20 +510,24 @@ namespace LightGBM {
...
-      sparse_rate_ = static_cast<double>(cnt_in_bin[default_bin_]) / total_sample_cnt;
-      const double max_sparse_rate = static_cast<double>(cnt_in_bin[most_freq_bin_]) / total_sample_cnt;
-      if (most_freq_bin_ != default_bin_ && max_sparse_rate > 0.7f) {
-        sparse_rate_ = max_sparse_rate;
-      } else {
+      const double max_sparse_rate =
+          static_cast<double>(cnt_in_bin[most_freq_bin_]) / total_sample_cnt;
+      // When most_freq_bin_ != default_bin_, there are some additional data loading costs.
+      // so use most_freq_bin_  = default_bin_ when there is not so sparse
+      if (most_freq_bin_ != default_bin_ && max_sparse_rate <= 0.7f) {
```
StrikerRUS: `kSparseThreshold`?
https://github.com/microsoft/LightGBM/blob/e676af236625fcfc548ca60d0b89f360c736a744/include/LightGBM/bin.h#L36

guolinke:
```suggestion
      if (most_freq_bin_ != default_bin_ && max_sparse_rate < kSparseThreshold) {
```

guolinke: change to `<` to consistent with other sparse conditions.

src/io/bin.cpp
```diff
@@ -510,20 +510,24 @@ namespace LightGBM {
 
     if (!is_trivial_) {
       default_bin_ = ValueToBin(0);
+      most_freq_bin_ =
+          static_cast<uint32_t>(ArrayArgs<int>::ArgMax(cnt_in_bin));
       if (bin_type_ == BinType::CategoricalBin) {
-        CHECK(default_bin_ > 0);
+        if (most_freq_bin_ == 0) {
+          CHECK(num_bin_ > 1);
+          // FIXME: how to enable `most_freq_bin_ = 0` for categorical features
```
StrikerRUS: @guolinke Create issue for FIXME to not lose it?

jameslamb: ^ I agree, we should be using issues instead of `TODO`, `FIXME`, etc.

guolinke: I think this a not trivial, so a `fixme` is here. 
BTW, PR #2463 will rewrite categorical feature solutions, and these codes are not needed.

PR #2769, opened 2020-02-18
some code refactoring

guolinke:

PR #2829, opened 2020-02-26
fix index and remove unused includes

StrikerRUS:
Refer to https://github.com/microsoft/LightGBM/pull/2769#pullrequestreview-363118368.

### DIFF
--- a/src/treelearner/feature_histogram.hpp
+++ b/src/treelearner/feature_histogram.hpp
@@ -595,7 +595,7 @@ class FeatureHistogram {
 
         double sum_left_gradient = sum_gradient - sum_right_gradient;
         if (IS_RAND) {
-          if (t + offset != rand_threshold) {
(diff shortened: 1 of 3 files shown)


src/treelearner/feature_histogram.hpp
```diff
@@ -595,7 +595,7 @@ class FeatureHistogram {
 
         double sum_left_gradient = sum_gradient - sum_right_gradient;
         if (IS_RAND) {
-          if (t + offset != rand_threshold) {
+          if (t - 1 + offset != rand_threshold) {
```
jameslamb: is there a unit test we can write that would have caught this and would prevent it from being re-introduced? Or is this too deep in the code to target with a test?

guolinke: I think we should enable c++ test, but I don't have effort for it..

PR #2801, opened 2020-02-23
Code refactoring for ranking objective & Faster ndcg_xendcg

guolinke:
1. optimize the O(n^2) loop in `ndcg_xendcg` to O(n).
2. Fix non-thread-safe random generator in `ndcg_xendcg`, which should not share a `Random` object in multi-threading. Furthermore, the random results should be identical now regards to different num_threads.

Also some code refactorings.

Benchmark:

Boosting cost:
| Data      | Master|  This PR |
|----|  ----| ----| 
| Yahoo LTR | 9.36 s |1.45 s |
| MS LTR|72.02 s | 2.72 s|

PR #2830, opened 2020-02-26
fixed cpplint issues and updated docs

StrikerRUS:
Refer to #1990 and #2801.

### DIFF
--- a/docs/Advanced-Topics.rst
+++ b/docs/Advanced-Topics.rst
@@ -39,7 +39,7 @@ LambdaRank
 
 -  Use ``label_gain`` to set the gain(weight) of ``int`` label.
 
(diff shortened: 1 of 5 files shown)


StrikerRUS:
@jameslamb 

cpplint argued
```
src/objective/rank_objective.hpp:72:  "virtual" is redundant since function is already declared as "override"  [readability/inheritance] [4]
```

jameslamb:
> @jameslamb
> 
> cpplint argued
> 
> ```
> src/objective/rank_objective.hpp:72:  "virtual" is redundant since function is already declared as "override"  [readability/inheritance] [4]
> ```

oh ok! I did not know about `override`

Issue #1909, opened 2018-12-17
[R} CRAN is unhappy about calls to stderr in source code


jameslamb:
This issue is a blocker for #629 (the path to CRAN distribution of `LightGBM`.

## Overview

CRAN does not permit compiled code in packages to terminate R or directly write to stdout/stderr. See this NOTE that arises:

```
* checking compiled code ... NOTE
Note: information on .o files for x64 is not available
File 'E:/sandbox/LightGBM/lightgbm.Rcheck/lightgbm/libs/x64/lib_lightgbm.dll':
  Found '_exit', possibly from '_exit' (C)
  Found 'abort', possibly from 'abort' (C), 'runtime' (Fortran)
  Found 'exit', possibly from 'exit' (C), 'stop' (Fortran)
  Found 'printf', possibly from 'printf' (C)

Compiled code should not call entry points which might terminate R nor
write to stdout/stderr instead of to the console, nor use Fortran I/O
nor system RNGs. The detected symbols are linked into the code but
might come from libraries and not actually be called.
It is good practice to register native routines and to disable symbol
search.

See 'Writing portable packages' in the 'Writing R Extensions' manual.
```

This is a tough issue to solve because the package uses code shared between R and Python.  Solving it for LightGBM probably involves doing something similar to what the XGBoost folks had to do ([see here](https://github.com/dmlc/xgboost/commit/1495a43cea05071e9451d008addffa598e4e7341)).

## How to run R CMD CHECK

To run R CMD CHECK on the package, clone the repo and install from source [per these directions](https://github.com/Microsoft/LightGBM/blob/master/R-package/README.md), then build the package and run R CMD CHECK.

On Mac OS, this looks like running the following from your local clone of `LightGBM`:

```
export CXX=/usr/local/bin/g++-8
export CC=/usr/local/bin/gcc-8

# build the package
Rscript build_r.R

# Check
LGB_VERSION=$(cat R-package/DESCRIPTION | grep ^Version | tr -d "Version: ")
R CMD CHECK lightgbm_${LGB_VERSION}.tar.gz --as-cran | tee check.log | cat
```

## How to close this issue

Contributions must pass our existing CI tests to be accepted. A contribution that fixes this issue will additionally result in R CMD CHECK not returning the note above on Mac OS.

guolinke:
@jameslamb 
I remember LightGBM R's package is a hacking solution, which actually doesn't call any R's functions. 
We choose this due to the license issues.
[...]

Issue #1440, opened 2018-06-10
[R-package] Ability to control the printed messages?


Laurae2:
Currently, LightGBM is not using the R printf message, leading to bypassing any sink (see #513).

An alternative, without losing performance on metric computation, is to return to R the metric values required and have a callback to print the messages.

Currently, we can log the printed messages by using the metric callback. But that's not the best scenario unfortunately (performance wise, especially on large datasets).

Taking 100M observations, training a model and computing AUC on my machine for 10 iterations:

* Using LightGBM metric: 628 seconds
* Using optimized R metric: 1159 seconds (optimized R AUC metric has around/less than 2M/s observation throughput, see https://github.com/Laurae2/R_benchmarking#area-under-the-curve-roc-benchmarks)

It would be beneficial for logging the performance of the model throughout the training to a file. Such case would be very useful, especially in RStudio Server: if a user is disconnected (for any reason) from the R session, he/she would not be able to get back on RStudio Server until LightGBM finishes training (logging performance would be a great addition).

guolinke:
@Laurae2 
Maybe "redirect" can solve your problem ? 
https://stackoverflow.com/questions/6674327/redirect-all-output-to-file

guolinke:
@Laurae2 any response ?

Laurae2:
Redirecting this way is not possible when the session creation is not controlled by the user (same issue with any server spawning R sessions).

guolinke:
@Laurae2 got it. However, this require to rewrite the R package, more specifically, the dynamic library used by R. 
Maybe the better solution is using some third-party high-performance metric evaluation packages.

StrikerRUS:
Closed in favor of being in #2302. We decided to keep all feature requests in one place.

Welcome to contribute this feature! Please re-open this issue (or post a comment if you are not a topic starter) if you are actively working on implementing this feature.

jameslamb:
Reopening this one since we have an active PR (#2901) for it

PR #2837, opened 2020-02-28
[R-package] Introduce R-specific control of lib_lightgbm

jameslamb:
I've been working on #1440 / #1909 (as part of  #629 ), and am still struggling with it.  But I think that a  small part of what I've done can be introduced right now, and that it actually might be easier to discuss it as its own thing.

I'm currently working on this NOTE from `R CMD CHECK`:

```
Note: information on .o files is not available
File ‘/Users/jlamb/repos/lgb-logging-thing/LightGBM/lightgbm.Rcheck/lightgbm/libs/lib_lightgbm.so’:
  Found ‘___assert_rtn’, possibly from ‘assert’ (C)
  Found ‘___stderrp’, possibly from ‘stderr’ (C)
  Found ‘___stdoutp’, possibly from ‘stdout’ (C)
  Found ‘_printf’, possibly from ‘printf’ (C)
  Found ‘_putchar’, possibly from ‘putchar’ (C)
  Found ‘_vprintf’, possibly from ‘vprintf’ (C)

```

Similar to the way [XGBoost solved this](https://github.com/dmlc/xgboost/commit/1495a43cea05071e9451d008addffa598e4e7341), I'd like to introduce a new define that tells the compiler "hey `lib_lightgbm` is being compiled for use with the R package".

I first tried to do this to introduce `Rprintf()`, the R equivalent to `printf()`, in `log.h`. I've been running into some complications there that will take some more time to address, so for this PR I'd like to focus on one isolated case not related to logging.

>   Found ‘___assert_rtn’, possibly from ‘assert’ (C)

This comes from `src/io/json11.cpp`. This PR uses `-DLGB_R_BUILD` to say "exclude `cassert` if you're building the R package"...and I can confirm it works! After running the code below:

```
export CXX=/usr/local/bin/g++-8
export CC=/usr/local/bin/gcc-8
Rscript build_r.R
R CMD CHECK lightgbm_2.3.2.tar.gz --as-cran --no-manual
```

I get the following output (notice that `assert` is no longer mentioned!):

```
Note: information on .o files is not available
File ‘/Users/jlamb/repos/lgb-logging-thing/LightGBM/lightgbm.Rcheck/lightgbm/libs/lib_lightgbm.so’:
  Found ‘___stderrp’, possibly from ‘stderr’ (C)
  Found ‘___stdoutp’, possibly from ‘stdout’ (C)
  Found ‘_printf’, possibly from ‘printf’ (C)
  Found ‘_putchar’, possibly from ‘putchar’ (C)
  Found ‘_vprintf’, possibly from ‘vprintf’ (C)
```

That command also runs all the R tests, so this should pass CI and doesn't seem to be affecting the R package's functionality.

If this PR is approved and merged, then the next thing I'll work on is `log.h`, where for example I'll try to make something like this work:

```
#include <R.h>

static void Write(LogLevel level, const char* level_str, const char *format, va_list val) {
    if (level <= GetLevel()) {  // omit the message with low level
      // write to STDOUT
      #ifndef LGB_R_BUILD
        printf("[LightGBM] [%s] ", level_str);
        vprintf(format, val);
        printf("\n");
      #else
        Rprintf("[LightGBM] [%s] ", level_str);
        Rvprintf(format, val);
        Rprintf("\n");
      #endif
      fflush(stdout);
    }
}
```

Thanks for your time!


guolinke:
thanks @jameslamb !
BTW, should we add something for pre-compiled dll?

StrikerRUS:
Just want to say that I'm really excited to see we have some progress on #629! Thank you James!

StrikerRUS:
Can't we achieve the same goal via CMake options to not play around with `export`? I'm asking because afraid that `export` inside the `system` call introduce side effects for user's environment, doesn't it? Or it only affects one particular R session?

https://github.com/microsoft/LightGBM/blob/53137e2541691fa8bcae2917eb75b7ddf0ec98c8/R-package/src/install.libs.R#L43-L45

jameslamb:
> Can't we achieve the same goal via CMake options to not play around with `export`? I'm asking because afraid that `export` inside the `system` call introduce side effects for user's environment, doesn't it? Or it only affects one particular R session?
> 
> https://github.com/microsoft/LightGBM/blob/53137e2541691fa8bcae2917eb75b7ddf0ec98c8/R-package/src/install.libs.R#L43-L45

`system()`  spins off an isolated shell session that doesn't impact the calling environment:

image

BUT you're right, passing the argument to `cmake`  directly would be cleaner, and would be portable to Windows. Let me try that!

jameslamb:
> thanks @jameslamb !
> BTW, should we add something for pre-compiled dll?
> I mean, the build flag, and related documentation for the pre-compiled dll for R package.

Can you point me to the code where we build the dll / so files that we include with each release? I'd be happy to update that code / documentation  to show how to build a version of the library specifically for R (with this flag).

But I don't think we need to do anything in `install.libs.R`...by the time you reach that script, you have to already have the pre-compiled dll in hand. Also I'll add a note on #629 to this effect, but I think our first priority should be getting R _source installs_ working in a CRAN-compliant way, and then work back to pre-compiled.

StrikerRUS:
@guolinke 
> BTW, should we add something for pre-compiled dll?
I mean, the build flag, and related documentation for the pre-compiled dll for R package.


I think we will publish a separate artifacts in our `Releases` which will be sent to CRAN. But I guess it should be done after closing #629. Or am I misunderstand something and you are speaking about different things?

guolinke:
I mean, sometimes user will build the dll first, then install python /R packages. After this PR, I think python/R will not share the same dll. So we need the build options to build R dll, maybe including msvc project file.

jameslamb:
> I mean, sometimes user will build the dll first, then install python /R packages. After this PR, I think python/R will not share the same dll. So we need the build options to build R dll, maybe including msvc project file.

yep, makes sense! I agree, from this point we should have a different `lib_lightgbm.so` / `lib_lightgbm.dll` based on whether or not you are using the R package.

Another option we could give is an R package `.tar.gz` that just *has*  that library in it.

@StrikerRUS to answer your question...yes CRAN allows for precompiled packages, but you cannot publish them yourself. That is done for security reasons, so that CRAN tightly controls what software people can get into your environment via CRAN.

We have to have a working *source* distribution (one that gets compiled when users download and install it) first, which means yes we have to close #629 first.

From the [CRAN policies](https://cran.r-project.org/web/packages/policies.html#Binary-packages):

> Binary packages are not accepted from maintainers: CRAN will only host binary packages prepared by those responsible for the binary areas. Their packages are made automatically by batch jobs and can take a day or two to appear on the CRAN master site (maybe longer to reach CRAN mirrors).

Laurae2:
@jameslamb According to my [lgbdl](https://github.com/Laurae2/lgbdl/blob/master/R/lgb.dl.R#L93) package, the precompiled library should be at the root of the LightGBM folder.

The file is then checked by [install.libs.R](https://github.com/microsoft/LightGBM/blob/master/R-package/src/install.libs.R#L114-L137) if [use_precompile](https://github.com/microsoft/LightGBM/blob/master/R-package/src/install.libs.R#L2) is `TRUE`.

jameslamb:
> @jameslamb According to my [lgbdl](https://github.com/Laurae2/lgbdl/blob/master/R/lgb.dl.R#L93) package, the precompiled library should be at the root of the LightGBM folder.
> 
> The file is then checked by [install.libs.R](https://github.com/microsoft/LightGBM/blob/master/R-package/src/install.libs.R#L114-L137) if [use_precompile](https://github.com/microsoft/LightGBM/blob/master/R-package/src/install.libs.R#L2) is `TRUE`.

perfect, thanks

jameslamb:
I'm going to rebase to `master` and rebuild just to be sure, then I'll merge this. Thanks for the reviews!

PR #2842, opened 2020-02-29
update lockbot

StrikerRUS:
Seems that I misunderstood the way this bot locks issues [here](https://github.com/microsoft/LightGBM/pull/2815#issuecomment-590464219).
As it locks **only closed** issues, I think that removed in this PR labels can be allowed to be locked as we keep them **open** when they are actual.

### DIFF
--- a/.github/lock.yml
+++ b/.github/lock.yml
@@ -11,7 +11,7 @@ skipCreatedBefore: false
 exemptLabels: ['effectiveness', 'efficiency', 'feature request', 'help wanted']
 
 # Label to add before locking, such as `outdated`. Set to `false` to disable
-lockLabel: true
+lockLabel: false
 
 # Comment to post before locking. Set to `false` to disable
 lockComment: >
--- a/.github/lock.yml
+++ b/.github/lock.yml
@@ -14,10 +14,7 @@ exemptLabels: ['effectiveness', 'efficiency', 'feature request', 'help wanted']
 lockLabel: false
 
 # Comment to post before locking. Set to `false` to disable
-lockComment: >
-  This thread has been automatically locked since there has not been
-  any recent activity after it was closed. Please open a new issue for
-  related bugs.
+lockComment: false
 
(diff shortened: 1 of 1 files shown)


guolinke:
maybe our config is wrong. In my expectation, it should close and lock the old issues..

.github/lock.yml
```diff
@@ -8,7 +8,7 @@ daysUntilLock: 60
 skipCreatedBefore: false
 
 # Issues and pull requests with these labels will be ignored. Set to `[]` to disable
-exemptLabels: ['blocking', 'bug', 'effectiveness', 'efficiency', 'feature request', 'help wanted', 'in progress']
+exemptLabels: ['effectiveness', 'efficiency', 'feature request', 'help wanted']
 
 # Label to add before locking, such as `outdated`. Set to `false` to disable
 lockLabel: true
```
guolinke: maybe this should set to `false`?

StrikerRUS:
@guolinke 
> In my expectation, it should close and lock the old issues..

According to bot's description, it locks only closed issues.
> GitHub App that locks closed issues and pull requests after a period of inactivity

.github/lock.yml
StrikerRUS: As we do not have `outdated` label or any similar to this, I think you are right.

guolinke:
@StrikerRUS yeah, but from its recent logs, it seems there are only comments, not locks, like this one:
image

StrikerRUS:
@guolinke 
> it seems there are only comments, not locks, like this one:

I think you can comment because you are the owner of this repo. I guess ordinary users cannot comment after that.

Or maybe the bot lacks of rights to perform a loc?..

StrikerRUS:
It looks different in XGBoost repo. We do not have the following line 

image

guolinke:
The permission is good:
image

maybe related to `lockLabel`.

StrikerRUS:
Perhaps. Let's merge and see 🙂

PR #2845, opened 2020-02-29
[R-package] removed unused imports from R-specific lib_lightgbm components

jameslamb:
In this PR, I propose removing some unused includes from `R_object_helper.h`, `lightgbm_R.cpp`, and `R_object_helper.h`.


jameslamb:
> LGTM as it passes CI checks and doesn't introduce new cpplint errors about "include what you use".

@StrikerRUS oh yeah! I should have added in the description that I ran `cpplint` looking for exactly those warnings, sorry.

PR #2846, opened 2020-02-29
[R-package] [docs] Added documentation on lgb.dl() and data.table version (fixes #2715)

### LINKED ISSUE #2715: lgb.cv data.table error - R package
Hi,

I have been running LightGBM 2.3.1 in R (version 3.5) for the past few months in a linux environment with no issues. My colleague has installed the package in the past few days, however when running the example code provided on the git page, they experience an error (whereas it works on my install):

## Error message

`Error in data.table::data.table(indices = test_indices, weight = getinfo(data,  : 
  column or argument 2 is NULL`

## Reproducible Code
```library(lightgbm)
  data(agaricus.train, package='lightgbm')
  train <- agaricus.train
  dtrain <- lgb.Dataset(train$data, label=train$label)
  params <- list(objective="regression", metric="l2")
  model <- lgb.cv(params, dtrain, 10, nfold=5, min_data=1, learning_rate=1, 
  early_stopping_rounds=10)
```
Appreciate any help with this!

jameslamb:
Per conversationi n #2715 , this PR adds documentation on the need to upgrade `data.table`  if you are using `lgb.dl()` to install the R package.

### DIFF
--- a/docs/FAQ.rst
+++ b/docs/FAQ.rst
@@ -201,11 +201,15 @@ Run ``lgb.unloader(wipe = TRUE)`` in the R console, and recreate the LightGBM da
 Due to the pointers, choosing to not wipe variables will not fix the error.
 This is a known issue: `Microsoft/LightGBM#698 <https://github.com/microsoft/LightGBM/issues/698>`__.
 
-2. I used ``setinfo``, tried to print my ``lgb.Dataset``, and now the R console froze!
---------------------------------------------------------------------------------------
+2. I used ``setinfo()``, tried to print my ``lgb.Dataset``, and now the R console froze!
+----------------------------------------------------------------------------------------
 
 Avoid printing the ``lgb.Dataset`` after using ``setinfo``.
 This is a known bug: `Microsoft/LightGBM#539 <https://github.com/microsoft/LightGBM/issues/539>`__.
+
+3. `error in data.table::data.table()...argument 2 is NULL`
+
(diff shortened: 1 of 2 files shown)


StrikerRUS:
Seems that I was wrong when suggested to write that warning in R README. I think [R section in FAQ](https://github.com/microsoft/LightGBM/blob/master/docs/FAQ.rst#r-package) is a better place for it. WDYT?

jameslamb:
> Seems that I was wrong when suggested to write that warning in R README. I think [R section in FAQ](https://github.com/microsoft/LightGBM/blob/master/docs/FAQ.rst#r-package) is a better place for it. WDYT?

yep I agree. I forget sometimes that we have that. I'll move this there, and I'll add a note in the README pointing to the FAQ (since we don't currently have that).

jameslamb:
Moved everything over to the FAQs page

docs/FAQ.rst
```diff
@@ -201,12 +201,16 @@ Run ``lgb.unloader(wipe = TRUE)`` in the R console, and recreate the LightGBM da
...
 
-2. I used ``setinfo``, tried to print my ``lgb.Dataset``, and now the R console froze!
---------------------------------------------------------------------------------------
+2. I used ``setinfo()``, tried to print my ``lgb.Dataset``, and now the R console froze!
+----------------------------------------------------------------------------------------
 
 Avoid printing the ``lgb.Dataset`` after using ``setinfo``.
 This is a known bug: `Microsoft/LightGBM#539 <https://github.com/microsoft/LightGBM/issues/539>`__.
 
+3. `error in data.table::data.table()...argument 2 is NULL`
```
StrikerRUS: Underscore to fix formatting.

image

jameslamb: ah! I knew I'd miss one

jameslamb: image

ok looks better now

Issue #2814, opened 2020-02-24
Java SWIG - how to extract featureNames?


AlbertoEAF:
Hello,

I'm having issues calling the SWIG wrappers in Java to extract the model feature names.
I get a segmentation fault when calling LGBM_BoosterGetFeatureNames.

In the Java-exposed lightgbmJNI.LGBM_BoosterGetFeatureNames method, the parameters are:

`LGBM_BoosterGetFeatureNames(long boosterHandle, long outNumFeaturesPtr, String[] strings)`
(https://lightgbm.readthedocs.io/en/latest/C-API.html#c.LGBM_BoosterGetFeatureNames)

The issue is that LGBM_BoosterGetFeatureNames requires an already allocated array of strings on the C side, which I cannot do if I pass a java String[] object. This then results in a memory segfault.

I've seen this comment related to implementing typemaps in SWIG for trickier cases https://github.com/microsoft/LightGBM/issues/909#issuecomment-368148784.

I also saw some notices in the swig interface file:
`/* Note: there is a bug in the SWIG generated string arrays when creating
   a new array with strings where the strings are prematurely deallocated
*/
%array_functions(char *, stringArray)`

Am I using the API incorrectly or currently this function cannot be used?

Thank you! :D

AlbertoEAF:
I really need to use that function, so I made an extension to the lightgbmlib.i SWIG interface as an MVP just to make sure I can extract the feature names in my project until I get a reply from you on how to do it. 

**Disclaimer** 
_I didn't care at all for building clean code, just implementing an MVP that works correctly so I can use LGBM_BoosterGetFeatureNames reliably. This is not production quality code, I could clean it up quite a bit if you find it useful.
A properly working String[] typemap would also be much better, but I have little to no experience with SWIG yet to do one._


_To avoid touching any existing code at all, I added a single line at the end of lightgbmlib.i `%include "char_array_helpers.i"` that adds a series of `custom_...` functions with to the exported lightgbmlib and a variant of LGBM_BoosterGetFeatureNames that works for me without crashes. You can find the code in this branch: https://github.com/AlbertoEAF/LightGBM/tree/feat/swig-strings_

In that interface file I create a class FeatureNames + typedef FeatureNamesHandle that manages an array of fixed length strings.

I then build a wrapper to the `LGBM_BoosterGetFeatureNames` that creates a new FeatureNames instance from which the user can extract the feature names:

 This is how you use it:

`

[...]

PR #2850, opened 2020-03-01
Fix SWIG methods that return char**

AlbertoEAF:
The main reason to add this pull request is outlined in https://github.com/microsoft/LightGBM/issues/2814.

After discussing with @imatiach-msft and the devs at SWIG (https://github.com/swig/swig/issues/1740), 
the devised solution aiming at simplicity and correctness was to add a StringArray to wrap any char** parameters so memory can be managed.
This relies on the already existing `various.i` extensions to return a String[] through a single call.

@imatiach-msft also requested that LGBM_BoosterGetEvalNamesSWIG be refactored to use the same type of API. This meant changing that method signature, and I did so by following the LGBM_* signature of returning the error code as an int 0/-1.

I'm very interested in getting feedback from the team, specially @imatiach-msft with whom I've been discussing this topic. If possible I'd like to know your thoughts on these questions first: https://github.com/microsoft/LightGBM/issues/2814.

I also believe this should be improved upon before merging. I left some "@reviewer" notes throughout the code. If you can read those and discuss with me I'd appreciate it, as I believe those points should be addressed to get the code style more in line with the rest of the code in the repo, and to streamline the code as well.
Also, can we find a way to add the docstrings in the interface files to the documentation?

You can now get hold of String[] by just calling `swigStringArrayHandle.data()`.Working example:

```
    public String[] getBoosterFeatureNames() {

        final int MAX_FEATURE_NAME_SIZE = 512; // Will segfault if a string has a name larger than this.

        long swigStringArrayHandleOutPtr = lightgbmlibJNI.voidpp_handle();
        long swigStringArrayHandle;

        try {
            int retcodeLGBM = lightgbmlibJNI.LGBM_BoosterGetFeatureNamesSWIG(
                    swig.swigBoosterHandle,
                    MAX_FEATURE_NAME_SIZE,
                    swigStringArrayHandleOutPtr // Returns a newly allocated FeatureNames handle
            );
            if (retcodeLGBM == -1) {
                logger.error("Couldn't read feature names");
                throw new LightGBMException();
            }
            swigStringArrayHandle = lightgbmlibJNI.voidpp_value(swigStringArrayHandleOutPtr); // Get the handle
        }
        finally {
            lightgbmlibJNI.delete_voidpp(swigStringArrayHandleOutPtr);
        }

        String[] featureNames = lightgbmlibJNI.StringArrayHandle_get_strings(swigStringArrayHandle);

        for(int i = 0; i < featureNames.length; ++i) {
            logger.debug("LightGBM model feature " + i + ": " + featureNames[i]);
        }
        logger.trace("Deallocating feature names.");
        lightgbmlibJNI.StringArrayHandle_free(swigStringArrayHandle);

        return featureNames;
    }
```

 Thank you!

By the way, here is the changelog:

+ Add StringArray class to manage and manipulate arrays of fixed-length strings.

  This class is now used to wrap any char** parameters, manage memory and
  manipulate the strings.

+ Fix SWIG LGBM_BoosterGetFeatureNames call that would result in a segfault.

  Added wrapper LGBM_BoosterGetFeatureNamesSWIG (the method it wraps didn't work)

+ For consistency, LGBM_BoosterGetEvalNames was wrapped as well.
 
  [breaking api change]
  Re-factored wrapper LGBM_BoosterGetEvalNames to follow LGBM_*
  convention of 0/-1 in case of success/failure.

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -2925,7 +2925,7 @@ class Booster(object):
                     raise ValueError("Length of eval names doesn't equal with num_evals")
                 if reserved_string_buffer_size < required_string_buffer_size.value:
                     raise BufferError(
-                        "Allocated eval buffer size ({}) was inferior to the needed size ({})."
+                        "Allocated eval name buffer size ({}) was inferior to the needed size ({})."
                         .format(reserved_string_buffer_size, required_string_buffer_size.value)
                     )
                 self.__name_inner_eval = \
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -2711,14 +2711,24 @@ class Booster(object):
         num_feature = self.num_feature()
         # Get name of features
         tmp_out_len = ctypes.c_int(0)
-        string_buffers = [ctypes.create_string_buffer(255) for i in range_(num_feature)]
+        reserved_string_buffer_size = 255
+        required_string_buffer_size = ctypes.c_size_t(0)
+        string_buffers = [ctypes.create_string_buffer(reserved_string_buffer_size) for i in range_(num_feature)]
         ptr_string_buffers = (ctypes.c_char_p * num_feature)(*map(ctypes.addressof, string_buffers))
         _safe_call(_LIB.LGBM_BoosterGetFeatureNames(
             self.handle,
+            num_feature,
             ctypes.byref(tmp_out_len),
+            reserved_string_buffer_size,
+            ctypes.byref(required_string_buffer_size),
             ptr_string_buffers))
         if num_feature != tmp_out_len.value:
             raise ValueError("Length of feature names doesn't equal with num_feature")
+        if reserved_string_buffer_size < required_string_buffer_size.value:
+            raise BufferError(
+                "Allocated feature name buffer size ({}) was inferior to the needed size ({})."
+                .format(reserved_string_buffer_size, required_string_buffer_size.value)
+            )
         return [string_buffers[i].value.decode() for i in range_(num_feature)]
 
     def feature_importance(self, importance_type='split', iteration=None):
@@ -2898,14 +2908,26 @@ class Booster(object):
             if self.__num_inner_eval > 0:
                 # Get name of evals
                 tmp_out_len = ctypes.c_int(0)
-                string_buffers = [ctypes.create_string_buffer(255) for i in range_(self.__num_inner_eval)]
+                reserved_string_buffer_size = 255
+                required_string_buffer_size = ctypes.c_size_t(0)
+                string_buffers = [
+                    ctypes.create_string_buffer(reserved_string_buffer_size) for i in range_(self.__num_inner_eval)
+                ]
                 ptr_string_buffers = (ctypes.c_char_p * self.__num_inner_eval)(*map(ctypes.addressof, string_buffers))
                 _safe_call(_LIB.LGBM_BoosterGetEvalNames(
                     self.handle,
+                    self.__num_inner_eval,
                     ctypes.byref(tmp_out_len),
+                    reserved_string_buffer_size,
+                    ctypes.byref(required_string_buffer_size),
                     ptr_string_buffers))
                 if self.__num_inner_eval != tmp_out_len.value:
                     raise ValueError("Length of eval names doesn't equal with num_evals")
+                if reserved_string_buffer_size < required_string_buffer_size.value:
+                    raise BufferError(
+                        "Allocated eval buffer size ({}) was inferior to the needed size ({})."
+                        .format(reserved_string_buffer_size, required_string_buffer_size.value)
+                    )
                 self.__name_inner_eval = \
                     [string_buffers[i].value.decode() for i in range_(self.__num_inner_eval)]
                 self.__higher_better_inner_eval = \
--- a/src/c_api.cpp
+++ b/src/c_api.cpp
@@ -521,6 +521,24 @@ class Booster {
     return idx;
   }
 
+  size_t GetLargestEvalNameSize() const {
+    size_t max_len = 0;
+    for (const auto& metric : train_metric_) {
+      for (const auto& name : metric->GetName()) {
+        max_len = std::max(name.size(), max_len);
+      }
+    }
+    return max_len;
+  }
+
+  size_t GetLargestFeatureNameSize() const {
+    size_t max_len = 0;
+    for (const auto& name : boosting_->FeatureNames()) {
+      max_len = std::max(name.size(), max_len);
+    }
+    return max_len;
+  }
+
   const Boosting* GetBoosting() const { return boosting_.get(); }
 
  private:
@@ -1367,6 +1385,20 @@ int LGBM_BoosterGetFeatureNames(BoosterHandle handle, int* out_len, char** out_s
   API_BEGIN();
   Booster* ref_booster = reinterpret_cast<Booster*>(handle);
   *out_len = ref_booster->GetFeatureNames(out_strs);
+  API_END();
+}
+
+int LGBM_BoosterGetLargestEvalNameSize(BoosterHandle handle, size_t* out_len) {
+  API_BEGIN();
+  Booster* ref_booster = reinterpret_cast<Booster*>(handle);
+  *out_len = ref_booster->GetLargestEvalNameSize();
+  API_END();
+}
+
+int LGBM_BoosterGetLargestFeatureNameSize(BoosterHandle handle, size_t* out_len) {
+  API_BEGIN();
+  Booster* ref_booster = reinterpret_cast<Booster*>(handle);
+  *out_len = ref_booster->GetLargestFeatureNameSize();
   API_END();
 }
 
--- a/src/c_api.cpp
+++ b/src/c_api.cpp
@@ -501,42 +501,34 @@ class Booster {
     return ret;
   }
 
-  int GetEvalNames(char** out_strs) const {
+  int GetEvalNames(char** out_strs, const int len, const size_t buffer_len, size_t *out_buffer_len) const {
+    *out_buffer_len = 0;
     int idx = 0;
     for (const auto& metric : train_metric_) {
       for (const auto& name : metric->GetName()) {
-        std::memcpy(out_strs[idx], name.c_str(), name.size() + 1);
+        if (idx < len) {
+          std::memcpy(out_strs[idx], name.c_str(), std::min(name.size() + 1, buffer_len));
+          out_strs[idx][buffer_len-1] = '\0';
+        }
+        *out_buffer_len = std::max(name.size() + 1, *out_buffer_len);
         ++idx;
       }
     }
     return idx;
   }
 
-  int GetFeatureNames(char** out_strs) const {
+  int GetFeatureNames(char** out_strs, const int len, const size_t buffer_len, size_t *out_buffer_len) const {
+    *out_buffer_len = 0;
     int idx = 0;
     for (const auto& name : boosting_->FeatureNames()) {
-      std::memcpy(out_strs[idx], name.c_str(), name.size() + 1);
+      if (idx < len) {
+        std::memcpy(out_strs[idx], name.c_str(), std::min(name.size() + 1, buffer_len));
+        out_strs[idx][buffer_len-1] = '\0';
+      }
+      *out_buffer_len = std::max(name.size() + 1, *out_buffer_len);
       ++idx;
     }
     return idx;
-  }
-
-  size_t GetLargestEvalNameSize() const {
-    size_t max_len = 0;
-    for (const auto& metric : train_metric_) {
-      for (const auto& name : metric->GetName()) {
-        max_len = std::max(name.size(), max_len);
-      }
-    }
-    return max_len;
-  }
-
-  size_t GetLargestFeatureNameSize() const {
-    size_t max_len = 0;
-    for (const auto& name : boosting_->FeatureNames()) {
-      max_len = std::max(name.size(), max_len);
-    }
-    return max_len;
   }
 
   const Boosting* GetBoosting() const { return boosting_.get(); }
@@ -1374,31 +1366,31 @@ int LGBM_BoosterGetEvalCounts(BoosterHandle handle, int* out_len) {
   API_END();
 }
 
-int LGBM_BoosterGetEvalNames(BoosterHandle handle, int* out_len, char** out_strs) {
+int LGBM_BoosterGetEvalNames(
+  BoosterHandle handle,
+  const int len,
+  int* out_len,
+  const size_t buffer_len,
+  size_t* out_buffer_len,
+  char** out_strs)
+{
   API_BEGIN();
   Booster* ref_booster = reinterpret_cast<Booster*>(handle);
-  *out_len = ref_booster->GetEvalNames(out_strs);
+  *out_len = ref_booster->GetEvalNames(out_strs, len, buffer_len, out_buffer_len);
   API_END();
 }
 
-int LGBM_BoosterGetFeatureNames(BoosterHandle handle, int* out_len, char** out_strs) {
+int LGBM_BoosterGetFeatureNames(
+  BoosterHandle handle,
+  const int len,
+  int* out_len,
+  const size_t buffer_len,
+  size_t* out_buffer_len,
+  char** out_strs)
+{
   API_BEGIN();
   Booster* ref_booster = reinterpret_cast<Booster*>(handle);
-  *out_len = ref_booster->GetFeatureNames(out_strs);
-  API_END();
-}
-
-int LGBM_BoosterGetLargestEvalNameSize(BoosterHandle handle, size_t* out_len) {
-  API_BEGIN();
-  Booster* ref_booster = reinterpret_cast<Booster*>(handle);
-  *out_len = ref_booster->GetLargestEvalNameSize();
-  API_END();
-}
-
-int LGBM_BoosterGetLargestFeatureNameSize(BoosterHandle handle, size_t* out_len) {
-  API_BEGIN();
-  Booster* ref_booster = reinterpret_cast<Booster*>(handle);
-  *out_len = ref_booster->GetLargestFeatureNameSize();
+  *out_len = ref_booster->GetFeatureNames(out_strs, len, buffer_len, out_buffer_len);
   API_END();
 }
 
--- a/src/c_api.cpp
+++ b/src/c_api.cpp
@@ -508,7 +508,7 @@ class Booster {
       for (const auto& name : metric->GetName()) {
         if (idx < len) {
           std::memcpy(out_strs[idx], name.c_str(), std::min(name.size() + 1, buffer_len));
-          out_strs[idx][buffer_len-1] = '\0';
+          out_strs[idx][buffer_len - 1] = '\0';
         }
         *out_buffer_len = std::max(name.size() + 1, *out_buffer_len);
         ++idx;
@@ -523,7 +523,7 @@ class Booster {
     for (const auto& name : boosting_->FeatureNames()) {
       if (idx < len) {
         std::memcpy(out_strs[idx], name.c_str(), std::min(name.size() + 1, buffer_len));
-        out_strs[idx][buffer_len-1] = '\0';
+        out_strs[idx][buffer_len - 1] = '\0';
       }
       *out_buffer_len = std::max(name.size() + 1, *out_buffer_len);
       ++idx;
@@ -1366,28 +1366,24 @@ int LGBM_BoosterGetEvalCounts(BoosterHandle handle, int* out_len) {
   API_END();
 }
 
-int LGBM_BoosterGetEvalNames(
-  BoosterHandle handle,
-  const int len,
-  int* out_len,
-  const size_t buffer_len,
-  size_t* out_buffer_len,
-  char** out_strs)
-{
+int LGBM_BoosterGetEvalNames(BoosterHandle handle,
+                             const int len,
+                             int* out_len,
+                             const size_t buffer_len,
+                             size_t* out_buffer_len,
+                             char** out_strs) {
   API_BEGIN();
   Booster* ref_booster = reinterpret_cast<Booster*>(handle);
   *out_len = ref_booster->GetEvalNames(out_strs, len, buffer_len, out_buffer_len);
   API_END();
 }
 
-int LGBM_BoosterGetFeatureNames(
-  BoosterHandle handle,
-  const int len,
-  int* out_len,
-  const size_t buffer_len,
-  size_t* out_buffer_len,
-  char** out_strs)
-{
+int LGBM_BoosterGetFeatureNames(BoosterHandle handle,
+                                const int len,
+                                int* out_len,
+                                const size_t buffer_len,
+                                size_t* out_buffer_len,
+                                char** out_strs) {
   API_BEGIN();
   Booster* ref_booster = reinterpret_cast<Booster*>(handle);
   *out_len = ref_booster->GetFeatureNames(out_strs, len, buffer_len, out_buffer_len);
--- a/src/lightgbm_R.cpp
+++ b/src/lightgbm_R.cpp
@@ -465,8 +465,7 @@ LGBM_SE LGBM_BoosterGetEvalNames_R(LGBM_SE handle,
       R_GET_PTR(handle),
       len, &out_len,
       reserved_string_size, &required_string_size,
-      ptr_names.data()
-  ));
+      ptr_names.data()));
   CHECK_EQ(out_len, len);
   CHECK_GE(reserved_string_size, required_string_size);
   auto merge_names = Common::Join<char*>(ptr_names, "\t");
--- a/src/lightgbm_R.cpp
+++ b/src/lightgbm_R.cpp
@@ -449,15 +449,23 @@ LGBM_SE LGBM_BoosterGetEvalNames_R(LGBM_SE handle,
   R_API_BEGIN();
   int len;
   CHECK_CALL(LGBM_BoosterGetEvalCounts(R_GET_PTR(handle), &len));
+
+  const size_t reserved_string_size = 128;
   std::vector<std::vector<char>> names(len);
   std::vector<char*> ptr_names(len);
   for (int i = 0; i < len; ++i) {
-    names[i].resize(128);
+    names[i].resize(reserved_string_size);
     ptr_names[i] = names[i].data();
   }
+
   int out_len;
-  CHECK_CALL(LGBM_BoosterGetEvalNames(R_GET_PTR(handle), &out_len, ptr_names.data()));
+  size_t required_string_size;
+  CHECK_CALL(LGBM_BoosterGetEvalNames(R_GET_PTR(handle),
+                                      len, &out_len,
+                                      reserved_string_size, &required_string_size,
+                                      ptr_names.data()));
   CHECK_EQ(out_len, len);
+  CHECK_GE(reserved_string_size, required_string_size);
   auto merge_names = Common::Join<char*>(ptr_names, "\t");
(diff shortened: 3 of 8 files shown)


StrikerRUS:
@AlbertoEAF Sorry for the inconvenience, but can you please rebase to the latest `master` to fix CI checks and sign the CLA?

AlbertoEAF:
Hello @StrikerRUS,

I'd very much like to, but I'm waiting for approval from my company to sign it and
I don't know yet when they'll decide, but when they do I'll let you know.

AlbertoEAF:
@StrikerRUS done :)

Regarding the code, the code of `StringArray.hpp` is duplicated inside the interface file `char_array_helpers.i` (and it should be placed outside only). I'd like to get it in a separate file, maybe in src+include or as a separate .i file which I couldn't manage to do directly as I would have to modify the CMake procedure to add more .cpp files I guess?

Also please don't merge directly, and give me feedback first on the comments that say "@reviewer".

Also, char_array_helpers.i should also likely be renamed to `StringArray_api_extensions.i` or something.

AlbertoEAF:
Hello, I changed the API to be simpler to use and removed the choice of string allocation size from the user, to be impossible to result in an irrecoverable segmentation fault in case the user decides to allocate too small strings when asking for the feature names.

I changed the 2 wrappers to be closer to what @imatiach-msft had initially for `LGBM_BoosterGetEvalNames`, by using the return value to return the array of strings, but wrapped in a StringArrayHandle, which can then correctly manipulated and/or deallocated on the java side by using `StringArrayHandle_free()`.

The new usage is much simpler now than it was on the first submission (nullptr is returned in case of any error). Example:
```
    public String[] getBoosterFeatureNames() {

        long swigStringArrayHandle = lightgbmlibJNI.LGBM_BoosterGetFeatureNamesSWIG2(swig.swigBoosterHandle);
        if (swigStringArrayHandle == 0) {
            logger.error("Couldn't read feature names");
            throw new LightGBMException();
        }

        String[] featureNames = lightgbmlibJNI.StringArrayHandle_get_strings(swigStringArrayHandle);
        logger.debug("LightGBM model features: {}.", Arrays.toString(featureNames));

        logger.trace("Deallocating feature names.");
        lightgbmlibJNI.StringArrayHandle_free(swigStringArrayHandle);

        return featureNames;
    }
```

Could someone tell me how to add the `swig/StringArray.hpp` file and remove that whole block which is for now copy-pasted in `swig/StringArray_API_extensions.i`?

Also, can someone review? All inputs are appreciated :)

Thanks!

swig/StringArray_API_extensions.i
```diff
@@ -0,0 +1,281 @@
...
+     *
+     * @param num_elements Number of strings to store in the array.
+     * @param string_size The size of each string in the array.
+     */
+    void _allocate_strings(int num_elements, int string_size)
+    {
+        for (int i = 0; i < num_elements; ++i)
+        {
+            // Leave space for \0 terminator:
+            _array[i] = new (std::nothrow) char[string_size + 1];
```
imatiach-msft: hmm this seems wasteful, why make all of the strings the same size?

imatiach-msft:
"Could someone tell me how to include the swig/StringArray.hpp file and remove its copy-pasted code from swig/StringArray_API_extensions.i?"

I would think that in swig/lightgbmlib.i you would be able to include StringArray.hpp using just:

```
 #include "./StringArray.hpp" 
```

similar to how the export.h file is included in swig/lightgbmlib.i:

```
#include "../include/LightGBM/export.h"
```

please let me know if I misunderstood something!

swig/StringArray_API_extensions.i
```diff
@@ -0,0 +1,281 @@
...
+
+    #define API_OK_OR_VALUE(api_return, return_value) if (api_return == -1) return return_value
+    #define API_OK_OR_NULL(api_return) API_OK_OR_VALUE(api_return, nullptr)
+
+    typedef void* StringArrayHandle;
+
+
+
+
+// TODO: @reviewer - I'd like to see the StringArray class moved to another .i file or to the src + include folders.
```
imatiach-msft: either is possible, I think you can declare it in a .i file or keep it as hpp and reference it in the lightgbmlib.i file

swig/StringArray_API_extensions.i
```diff
@@ -0,0 +1,281 @@
...
+     * @return StringArrayHandle with the eval names (or nullptr in case of error)
+     */
+    StringArrayHandle LGBM_BoosterGetEvalNamesSWIG(BoosterHandle handle,
+                                                   int eval_counts)
+    {
+        std::unique_ptr<StringArray> strings(nullptr);
+
+        try {
+            // TODO: @reviewer, 128 was the chosen size, any particular reason for this constraint?
+            strings.reset(new StringArray(eval_counts, 128));
```
imatiach-msft: great question, I'm not sure what I was thinking back then but I'm guessing I based it off of this code in the R wrapper:

https://github.com/microsoft/LightGBM/blob/master/src/lightgbm_R.cpp#L453

not sure if this is correct

Interestingly in python layer it looks like it is 255?

https://github.com/microsoft/LightGBM/blob/master/python-package/lightgbm/basic.py#L2901

maybe I am misreading the code.

swig/StringArray_API_extensions.i
```diff
@@ -0,0 +1,281 @@
...
+        } catch (std::bad_alloc &e) {
+            LGBM_SetLastError("Failure to allocate memory."); // TODO: @reviewer - why is this not setting the message?
```
imatiach-msft: sorry, can you elaborate - do you mean why the previous code prior to this PR was not doing this or do you mean that this is currently not working?  If you mean the previous implementation that's a good point, we probably should have done this.

swig/StringArray_API_extensions.i
```diff
@@ -0,0 +1,281 @@
...
+     *
+     * Allocates a new StringArray. You must free it yourself if it succeeds.
+     * @see StringArrayHandle_free().
+     * In case of failure such resource is freed and nullptr is returned.
+     * Check for that case with null (lightgbmlib) or 0 (lightgbmlibJNI).
+     *
+     * @param handle Booster handle
+     * @return StringArrayHandle with the feature names (or nullptr in case of error)
+     */
+    StringArrayHandle LGBM_BoosterGetFeatureNamesSWIG2(BoosterHandle handle)
```
imatiach-msft: let's just keep it as LGBM_BoosterGetFeatureNamesSWIG?

swig/StringArray_API_extensions.i
imatiach-msft: nevermind, I take that back, it seems like other lightgbm code does similar things, I guess in these cases the wasted space is very small and makes the code less complex which is fine

swig/StringArray_API_extensions.i
```diff
@@ -0,0 +1,281 @@
...
+        int num_features;
+        API_OK_OR_NULL(LGBM_BoosterGetNumFeature(handle, &num_features));
+
+        int max_feature_name_size;
+        API_OK_OR_NULL(LGBM_BoosterGetLargestFeatureNameSize(handle, &max_feature_name_size));
+
+        // 2) Allocate an array of strings:
+        try {
+            // TODO: @reviewer should we also figure out the size of the biggest string and remove parameter max_feature_name_size?
+            strings.reset(new StringArray(num_features, max_feature_name_size));
```
imatiach-msft: this seems fine

swig/StringArray_API_extensions.i
```diff
@@ -0,0 +1,281 @@
...
+        } catch (std::bad_alloc &e) {
+            LGBM_SetLastError("Failure to allocate memory."); // TODO: @reviewer - why is this not setting the message?
```
imatiach-msft: same question as above

swig/lightgbmlib.i
```diff
@@ -249,9 +237,11 @@
 %array_functions(float, floatArray)
 %array_functions(int, intArray)
 %array_functions(long, longArray)
+
 /* Note: there is a bug in the SWIG generated string arrays when creating
    a new array with strings where the strings are prematurely deallocated
 */
+// TODO: I think we can get rid of this line and the above comments @reviewer?
```
imatiach-msft: yes, I think you can remove this now

swig/StringArray_API_extensions.i
AlbertoEAF: Yes, we'd need to rewrite the C API interface to add methods that allocate only the needed space, or even make it return a newly allocated char**.

swig/StringArray_API_extensions.i
AlbertoEAF: Hi :)

I mean that it is not setting the message. If you add a `throw std::bad_alloc()` in the try block you'll see that the output message from `LGBM_GetLastError` is the default, that everything is ok.
Is this because we're in a separate compilation unit and the message array is static on the original code? Either way, we should find a way to make it work.

swig/StringArray_API_extensions.i
AlbertoEAF: You're perfectly right! Great catch! I'm sorry, was doing some experimentation and forgot to put that back.

AlbertoEAF:
Ok, here's my current TODO based on @imatiach-msft's comments, I'll change the request to WIP until I fix those:

- Include StringArray.hpp
- Remove the char[128] assumption from `LGBM_BoosterGetEvalNames`, and compute what's the needed size so no crashes can occur (like it was done for `LGBM_BoosterGetFeatureNames`.
- Figure out why the error message is not being set (`LGBM_SetLastError`).

AlbertoEAF:
Thanks @imatiach-msft !

thanks for the pointer to the include, I didn't use the %{%} around the include before which was the trick ;) I also refactored the StringArray into StringArray.hpp + StringArray.i, which is then used in StringArray_API_extensions.i and contains the desired LightGBM function wrappers.

I've addressed everything except the `Lgbm_SetLastError`. 
This might be due to thread_local storage? Do you know anyone else who might be able to answer/fix this?

Otherwise, I've removed the assumptions of 128 characters and changed the `LGBM_BoosterGetEvalNamesSWIG` API, which alike `LGBM_BoosterGetFeatureNamesSWIG`, now only requires the BoosterHandle and figures out how much memory to allocate. This means now it's not possible to pass a too small of a string, resulting in writing out of bounds, which is undefined behaviour and could cause an irrecoverable crash.

What do you think?

swig/StringArray_API_extensions.i
AlbertoEAF: I've removed the assumption to avoid hitting out-of-bounds copies on the lower levels and undefined behaviour which could cause irrecoverable crashes.

imatiach-msft:
@AlbertoEAF so you are calling LGBM_SetLastError(<string>), and when calling LGBM_GetLastError() you are getting the default error message?  Where are you calling LGBM_GetLastError() from in your test case?  Is it on the same thread?  Supposedly it should work - it has worked for me in the past.

The underlying code is very simple, this is where set last error is defined, which just does a string copy:

https://github.com/microsoft/LightGBM/blob/master/include/LightGBM/c_api.h#L1059

and this is get last error:

https://github.com/microsoft/LightGBM/blob/master/src/c_api.cpp#L584

Which just gets the string in LastErrorMsg:

https://github.com/microsoft/LightGBM/blob/master/include/LightGBM/c_api.h#L1051

It really should be very simple, and if you are seeing issues adding some prints there should clear everything up.

imatiach-msft:
"What do you think?"

This looks great to me.

AlbertoEAF:
> @AlbertoEAF so you are calling LGBM_SetLastError(), and when calling LGBM_GetLastError() you are getting the default error message? Where are you calling LGBM_GetLastError() from in your test case? Is it on the same thread? Supposedly it should work - it has worked for me in the past.

@imatiach-msft yes, I was just using one thread in Java and calling LGBM_GetLastError() after getting the error. I also think it's it's not incorrectly implemented, so even if it doesn't work in my case I guess we're done right?

I'm going to rebase to master and squash all commits into one.

AlbertoEAF:
Hello @guolinke and @chivee :)

I've been iterating this with @imatiach-msft's feedback and we're both happy with the end result.

Can one of you review?
All the commits have been squashed, and rebased against the master.

src/c_api.cpp
```diff
@@ -521,6 +521,24 @@ class Booster {
     return idx;
   }
 
+  size_t GetLargestEvalNameSize() const {
+    size_t max_len = 0;
+    for (const auto& metric : train_metric_) {
```
guolinke: why `train_metric_`?

guolinke:
I think a better fix for this is like we do in 
https://github.com/microsoft/LightGBM/blob/643dd7b9064eac598f05e023393e6ec3f81f27fe/src/c_api.cpp#L1655-L1669

the api will pass the `buffer_len` and the `out_len`. If `buffer_len` is enough, it will copy into it, otherwise it returns the `out_len`. and then the caller could know this call is success or not. if not, it can re-allocate the buffer and call again.

src/c_api.cpp
AlbertoEAF: I'm just wrapping LGBM_BoosterGetEvalNames, so I looked into the c_api.cpp and it calls `GetEvalNames` (though it can copy out of bounds), so I just copied the code from `GetEvalNames`:
image

and changed the loop code to figure out the largest name size for the copy.

Is `GetEvalNames` wrong?

AlbertoEAF:
Hello @guolinke I also like your suggestion for it avoids adding `GetLargestFeatureNameSize` and `GetLargestEvalNameSize`.

However, for that I'll need to **change the existing C API** (which I was avoiding), namely:
- [internal C implementation only] 
`GetEvalNames(char** out_strs)` becomes 
`GetEvalNames(char** out_strs, int buffer_len, int *out_buffer_len)`
and
`GetFeatureNames(char** out_strs)` becomes
`GetFeatureNames(char** out_strs, int buffer_len, int *out_buffer_len)`

- [exposed C API] 
`LGBM_BoosterGetEvalNames(BoosterHandle handle, int* out_len, char** out_strs)` becomes 
`LGBM_BoosterGetEvalNames(BoosterHandle handle, int buffer_len, int* out_buffer_len, int* out_len, char** out_strs)`
and
`LGBM_BoosterGetFeatureNames(BoosterHandle handle, int* out_len, char** out_strs)` becomes 
`LGBM_BoosterGetFeatureNames(BoosterHandle handle, int buffer_len, int* out_buffer_len, int* out_len, char** out_strs)`

Is that ok?

_**Note**: This change **will affect the Python API and R API** as well and **will require changes** in there._

guolinke:
It think it is okay.

AlbertoEAF:
Hello @guolinke, @imatiach-msft and all to whom this may concern,

**Changelog**

1. The first commit strived to avoid impacting any other API's and bring all needed functionality for char** to Java's SWIG wrappers and defines their final form (discussed with @imatiach-msft). 
2. To follow @guolinke's request, I had to change the internals of the C API (and had to change accordingly all impacted wrappers in Java, R and Python) so correctness is mantained across the whole codebase.

**C API changes (safety)**
The changed C API functions no longer result in undefined behaviour if the received storage space is not enough both in number of strings and their size. The changed functions now:
- receive both the number of strings allocated and their allocated size,
- output the actual number of required strings and string sizes required.

With the new C API implementation, if you provide a lower number of allocated strings or too small strings for the true sizes there is no problem, as it will then copy only as many strings as there are allocated, and copy each string up to its available size only and properly null-terminates it.

**New API usage**
This also means that one can provide estimates for storage space necessary and know after the call if it was enough or not, and must be repeated with more space.

Since this functions are not in hot-code regions, in the Java SWIG wrappers I chose simply to pass a null pointer and 0-sizes for both number of strings and their size, just to allocate memory accordingly and then repeat the call.
Note that with the new API one doesn't need any other methods, by two consecutive calls it is possible to compute allocation space needed and then retrieve the data. Example from the SWIG wrappers:
```
int num_features;
size_t max_feature_name_size;
std::unique_ptr<StringArray> strings(nullptr);

// Retrieve required allocation space:
API_OK_OR_NULL(LGBM_BoosterGetFeatureNames(handle,
                                            0, &num_features,
                                            0, &max_feature_name_size,
                                            nullptr));

strings.reset(new StringArray(num_features, max_feature_name_size));

API_OK_OR_NULL(LGBM_BoosterGetFeatureNames(handle,
                                            num_features, &num_features,
                                            max_feature_name_size, &max_feature_name_size,
                                            strings->data()));

return strings.release();
```

**R and Python wrappers changes**

- To minimize code changes, although the new SWIG wrappers support allocating the correct strings size, I didn't add support for resizing in the R and Python wrappers for they already assumed a fixed size of 128 and 255 characters respectively. 
- Added assert for the string size (which couldn't be computed before). 
- Instead of the assert, a warning could be issued instead, for with the new C API implementation, even if you don't allocate enough size in the strings, the strings will have the first N-1 characters initialized and the last character set with the null terminator instead of crashing.

If you want, a warning instead of an error could be triggered instead, or proper sizing of the content like in the SWIG wrappers without assumptions.

src/lightgbm_R.cpp
```diff
@@ -449,15 +449,23 @@ LGBM_SE LGBM_BoosterGetEvalNames_R(LGBM_SE handle,
...
     ptr_names[i] = names[i].data();
   }
+
   int out_len;
-  CHECK_CALL(LGBM_BoosterGetEvalNames(R_GET_PTR(handle), &out_len, ptr_names.data()));
+  size_t required_string_size;
+  CHECK_CALL(LGBM_BoosterGetEvalNames(R_GET_PTR(handle),
+                                      len, &out_len,
+                                      reserved_string_size, &required_string_size,
+                                      ptr_names.data()));
```
jameslamb: Looks ok to me. Could you change the indentation to make it clearer?

```
CHECK_CALL(
    LGBM_BoosterGetEvalNames(
        R_GET_PTR(handle),
        len,
        &out_len,
        reserved_string_size,
        &required_string_size,
        ptr_names.data()
    )
);
```

AlbertoEAF: Sure :)

AlbertoEAF: Thanks @jameslamb :)

AlbertoEAF:
Hello @guolinke and @StrikerRUS  :), 

Do you know if anyone else needs to approve? I'm not sure how the procedure goes now.
Should all the other 4 code owners review as well? Or is it good?

Thanks!

src/c_api.cpp
```diff
@@ -501,21 +501,31 @@ class Booster {
...
-  int GetEvalNames(char** out_strs) const {
+  int GetEvalNames(char** out_strs, const int len, const size_t buffer_len, size_t *out_buffer_len) const {
+    *out_buffer_len = 0;
     int idx = 0;
     for (const auto& metric : train_metric_) {
       for (const auto& name : metric->GetName()) {
-        std::memcpy(out_strs[idx], name.c_str(), name.size() + 1);
+        if (idx < len) {
+          std::memcpy(out_strs[idx], name.c_str(), std::min(name.size() + 1, buffer_len));
+          out_strs[idx][buffer_len-1] = '\0';
```
```suggestion
          out_strs[idx][buffer_len - 1] = '\0';
```

src/c_api.cpp
```diff
@@ -501,21 +501,31 @@ class Booster {
...
 
-  int GetFeatureNames(char** out_strs) const {
+  int GetFeatureNames(char** out_strs, const int len, const size_t buffer_len, size_t *out_buffer_len) const {
+    *out_buffer_len = 0;
     int idx = 0;
     for (const auto& name : boosting_->FeatureNames()) {
-      std::memcpy(out_strs[idx], name.c_str(), name.size() + 1);
+      if (idx < len) {
+        std::memcpy(out_strs[idx], name.c_str(), std::min(name.size() + 1, buffer_len));
+        out_strs[idx][buffer_len-1] = '\0';
```
```suggestion
        out_strs[idx][buffer_len - 1] = '\0';
```

swig/StringArray.hpp
```diff
@@ -0,0 +1,146 @@
+#ifndef __STRING_ARRAY_H__
```
StrikerRUS: We don't check `swig` folder content for style issues, so I'll not leave any comments here, except the need of copyright info.
```suggestion
/*!
 * Copyright (c) 2020 Microsoft Corporation. All rights reserved.
 * Licensed under the MIT License. See LICENSE file in the project root for license information.
 */
#ifndef __STRING_ARRAY_H__
```

swig/StringArray.i
```diff
@@ -0,0 +1,91 @@
+/**
```
StrikerRUS:
```suggestion
/*!
 * Copyright (c) 2020 Microsoft Corporation. All rights reserved.
 * Licensed under the MIT License. See LICENSE file in the project root for license information.
 */
/**
```

swig/StringArray_API_extensions.i
```diff
@@ -0,0 +1,103 @@
+/**
```
StrikerRUS:
```suggestion
/*!
 * Copyright (c) 2020 Microsoft Corporation. All rights reserved.
 * Licensed under the MIT License. See LICENSE file in the project root for license information.
 */
/**
```

src/lightgbm_R.cpp
```diff
@@ -449,15 +449,27 @@ LGBM_SE LGBM_BoosterGetEvalNames_R(LGBM_SE handle,
...
-  CHECK_CALL(LGBM_BoosterGetEvalNames(R_GET_PTR(handle), &out_len, ptr_names.data()));
+  size_t required_string_size;
+  CHECK_CALL(
+    LGBM_BoosterGetEvalNames(
+      R_GET_PTR(handle),
+      len, &out_len,
+      reserved_string_size, &required_string_size,
+      ptr_names.data()
+    )
+  );
```
StrikerRUS:
```suggestion
    ));
```

src/c_api.cpp
```diff
@@ -1356,17 +1366,31 @@ int LGBM_BoosterGetEvalCounts(BoosterHandle handle, int* out_len) {
...
 
-int LGBM_BoosterGetEvalNames(BoosterHandle handle, int* out_len, char** out_strs) {
+int LGBM_BoosterGetEvalNames(
+  BoosterHandle handle,
+  const int len,
+  int* out_len,
+  const size_t buffer_len,
+  size_t* out_buffer_len,
+  char** out_strs)
+{
```
```suggestion
  char** out_strs) {
```

src/c_api.cpp
```diff
@@ -1356,17 +1366,31 @@ int LGBM_BoosterGetEvalCounts(BoosterHandle handle, int* out_len) {
...
 
-int LGBM_BoosterGetFeatureNames(BoosterHandle handle, int* out_len, char** out_strs) {
+int LGBM_BoosterGetFeatureNames(
+  BoosterHandle handle,
+  const int len,
+  int* out_len,
+  const size_t buffer_len,
+  size_t* out_buffer_len,
+  char** out_strs)
+{
```
```suggestion
  char** out_strs) {
```

AlbertoEAF:
Ok @StrikerRUS :)

Anything left? Thanks!

python-package/lightgbm/basic.py
```diff
@@ -2898,14 +2908,26 @@ def __get_eval_info(self):
...
+                    self.__num_inner_eval,
                     ctypes.byref(tmp_out_len),
+                    reserved_string_buffer_size,
+                    ctypes.byref(required_string_buffer_size),
                     ptr_string_buffers))
                 if self.__num_inner_eval != tmp_out_len.value:
                     raise ValueError("Length of eval names doesn't equal with num_evals")
+                if reserved_string_buffer_size < required_string_buffer_size.value:
+                    raise BufferError(
+                        "Allocated eval buffer size ({}) was inferior to the needed size ({})."
```
```suggestion
                        "Allocated eval name buffer size ({}) was inferior to the needed size ({})."
```

src/lightgbm_R.cpp
```diff
@@ -449,15 +449,26 @@ LGBM_SE LGBM_BoosterGetEvalNames_R(LGBM_SE handle,
...
   int out_len;
-  CHECK_CALL(LGBM_BoosterGetEvalNames(R_GET_PTR(handle), &out_len, ptr_names.data()));
+  size_t required_string_size;
+  CHECK_CALL(
+    LGBM_BoosterGetEvalNames(
+      R_GET_PTR(handle),
+      len, &out_len,
+      reserved_string_size, &required_string_size,
+      ptr_names.data()
+  ));
```
StrikerRUS:
```suggestion
      ptr_names.data()));
```

imatiach-msft:
@AlbertoEAF I tried to update to latest code today but I am getting a segfault when calling the new LGBM_BoosterGetEvalNamesSWIG API, any idea what might be going wrong?  Will take a deeper look over this week. 

#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00007f51b93942fd, pid=13026, tid=0x00007f51c0f61700
#
# JRE version: OpenJDK Runtime Environment (8.0_144-b01) (build 1.8.0_144-b01)
# Java VM: OpenJDK 64-Bit Server VM (25.144-b01 mixed mode linux-amd64 compressed oops)
# Problematic frame:
# C  [lib_lightgbm_swig.so+0x82fd]  LGBM_BoosterGetEvalNamesSWIG(void*)+0x2d

This is the new scala code I am using, but it doesn't seem to get past just the LGBM_BoosterGetEvalNamesSWIG call:

```
  def getEvalNames(boosterPtr: Option[SWIGTYPE_p_void]): Array[String] = {
    // Need to keep track of best scores for each metric, see callback.py in lightgbm for reference
    // For debugging, can get metric names
    val stringArrayHandle = lightgbmlib.LGBM_BoosterGetEvalNamesSWIG(boosterPtr.get)
    // LightGBMUtils.validateArray(stringArrayHandle, "Booster Get Eval Names")
    val evalNames = lightgbmlib.StringArrayHandle_get_strings(stringArrayHandle)
    lightgbmlib.StringArrayHandle_free(stringArrayHandle)
    evalNames
  }
```

AlbertoEAF:
Hello @imatiach-msft , just to be sure, what you did in https://github.com/microsoft/LightGBM/pull/2958 fixes it right?

It makes sense to me that it would.

imatiach-msft:
@AlbertoEAF yes, thanks!

src/c_api.cpp
```diff
@@ -501,21 +501,31 @@ class Booster {
...
-  int GetEvalNames(char** out_strs) const {
+  int GetEvalNames(char** out_strs, const int len, const size_t buffer_len, size_t *out_buffer_len) const {
+    *out_buffer_len = 0;
     int idx = 0;
     for (const auto& metric : train_metric_) {
       for (const auto& name : metric->GetName()) {
-        std::memcpy(out_strs[idx], name.c_str(), name.size() + 1);
+        if (idx < len) {
+          std::memcpy(out_strs[idx], name.c_str(), std::min(name.size() + 1, buffer_len));
+          out_strs[idx][buffer_len - 1] = '\0';
```
guolinke, line 511: here sometimes will have memory problems, maybe it is better for 
``` 
          auto cur_len = std::min(name.size(), buffer_len - 1);
          std::memcpy(out_strs[idx], name.c_str(), cur_len);
          out_strs[idx][cur_len] = '\0';
```
btw, is `idx < len` needed?

ping @AlbertoEAF @StrikerRUS

StrikerRUS: Hmm, maybe it is also a cause of #3398?..

AlbertoEAF: I agree with the change, it does the same but it's easier to read @guolinke and sets the null byte only once when the string has the limit size.

Yes, the `idx < len` check is to ensure if you have allocated a char** that only has space for `len` string pointers, you don't write outside it, even if internally we have a bigger array that would require more space. We stop the copy before writing outside allocated memory.

AlbertoEAF: P.S.: In other words, segfaults/memory write violations can only occur if the function receives wrong `len` and `buffer_len` sizes from the caller. I.e., the user/caller actually pre-allocated smaller/less "strings" in `char **out_strs` than that the values he passed in len/buffer_len respectively.

Issue #2615, opened 2019-12-05
[mmlspark] unable to allocate more than int max array


chris-smith-zocdoc:
The `lightgbmlib.new_doubleArray` function in the jvm lib limits its size to integers. We should be able to allocate more memory on 64bit systems.

Is there any reason not to allow this? Does the rest of LightGBM handle large data arrays correctly?

Current signature 

```java
public static SWIGTYPE_p_double new_doubleArray(int var0)
```

Proposal: Change int to long
```java
public static SWIGTYPE_p_double new_doubleArray(long var0)
```

I'm new to SWIG so I'm not sure if their macros allow for this, but we can just write our own small functions (copies of their generated code with the types changed)

```cpp
  static double *new_doubleArray_2(int64_t nelements) {
    return new double[nelements]();
  }

  static double doubleArray_getitem_2(double *ary, int64_t index) {
      return ary[index];
  }
  static void doubleArray_setitem_2(double *ary, int64_t index, double value) {
      ary[index] = value;
  }
```

## Reproducible examples

```java
long numRows = 17000000;
int numCols = 130;
SWIGTYPE_p_double data = lightgbmlib.new_doubleArray(numCols * numRows);
```

Related mmlspark issue https://github.com/Azure/mmlspark/issues/753

StrikerRUS:
@chris-smith-zocdoc 
> Does the rest of LightGBM handle large data arrays correctly?

I remember that **prediction** cannot handle data with more than MAX_INT rows. Support can be added in a language wrapper (#1615).

> due to the efficiency concern, LightGBM only accept the int32 nrow.
https://github.com/microsoft/LightGBM/issues/1590#issuecomment-415066108

@imatiach-msft @guolinke

chris-smith-zocdoc:
@StrikerRUS int32 for nrows is ok in my case. The use case is millions of rows and hundreds of columns, the result of nrows * ncols exceeds int32.

In my example, 17 million rows * 130 columns gives 2,210,000,000 which overflows.

In my actual dataset I have both more rows and more columns

guolinke:
in lightgbm side, both nrow and ncol could be int32, and the total size could be int32*int32. refer to https://github.com/microsoft/LightGBM/blob/master/src/c_api.cpp#L599-L607

ping @imatiach-msft

imatiach-msft:
[...]

PR #2859, opened 2020-03-03
[mmlspark] allow allocating more than int max array

imatiach-msft:
resolves issue https://github.com/microsoft/LightGBM/issues/2615 by introducing long versions of SWIG-generated array
ping @chris-smith-zocdoc @guolinke 
validated jar in mmlspark

### DIFF
--- a/swig/lightgbmlib.i
+++ b/swig/lightgbmlib.i
@@ -233,11 +233,6 @@
 %pointer_cast(int32_t *, void *, int32_t_to_voidp_ptr)
 %pointer_cast(int64_t *, void *, int64_t_to_voidp_ptr)
 
-%array_functions(double, doubleArray)
-%array_functions(float, floatArray)
-%array_functions(int, intArray)
-%array_functions(long, longArray)
-
 /* Custom pointer manipulation template */
 %define %pointer_manipulation(TYPE, NAME)
 %{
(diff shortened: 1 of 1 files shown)


StrikerRUS:
@imatiach-msft Sorry for the inconvenience, but can you please rebase to the latest master to fix CI checks?

imatiach-msft:
@StrikerRUS done!

StrikerRUS:
Hey @imatiach-msft,

Can you please resolve conflicts that appeared after merging #2850?

imatiach-msft:
@StrikerRUS done!  Resolved conflicts, updated PR.

imatiach-msft:
ping @guolinke @StrikerRUS thanks!

StrikerRUS:
@imatiach-msft 
The diff looks nice to me, but unfortunately I have no idea what is going on here, sorry! 😃 
`static TYPE *new_##NAME`
`%{}`
...

imatiach-msft:
@StrikerRUS this is just a slightly modified version of the SWIG file here which we were already using in the previous %array_functions call:
https://github.com/swig/swig/blob/master/Lib/carrays.i
note how the values with "int" type are now just replaced to have "int64_t" type.
Hope this helps clear up any confusion!

AlbertoEAF:
Hello @imatiach-msft,

If we're going to change that, then why not use an unsigned tpe like size_t?

AlbertoEAF:
And another point @imatiach-msft ,

SWIG's [carrays.i](https://github.com/swig/swig/blob/master/Lib/carrays.i), is NOT memory-allocation safe, which means your whole application can crash just because the OS kills your underlying C process.
Before talking with you, I posted an issue at SWIG's repo to see if they wanted to fix it but they never answered back: https://github.com/swig/swig/issues/1755.

So my suggestion is since we're changing the Java API anyways, just fix it ourselves. We could throw an exception or even easier, use new (nothrow). Anyone who wishes to handle that case just has to test in Java for the null pointer, all other users who are not concerned with such issues (non-production) can keep working like before.

What do you say?

swig/lightgbmlib.i
```diff
@@ -278,6 +273,36 @@ TYPE *NAME##_handle();
 
 %enddef
 
+%define %long_array_functions(TYPE,NAME)
+%{
+  static TYPE *new_##NAME(int64_t nelements) { %}
+  %{  return new TYPE[nelements](); %}
```
AlbertoEAF:
```suggestion
  %{  return new (std::nothrow) TYPE[nelements](); %}
```

AlbertoEAF: And you also need to `#include <new>`, or it won't compile.

imatiach-msft:
@AlbertoEAF "then can we use an unsigned type like size_t"
I think long is way more than enough in terms of size for the use-case.  I don't see size_t used as much in the lightgbm code anyway (also I'm not sure how I would do this from the scala code whereas I've already tested this with int64_t).

imatiach-msft:
@AlbertoEAF  "So my suggestion is since we're changing the Java API anyways, just fix it ourselves"

Agree that it should be fixed, but can this be part of a separate PR?  This issue is much more pressing and I really want to fix it so I can publish a new lightgbm jar for customers with very large data.  The additional change would require more testing on my side.

swig/lightgbmlib.i
AlbertoEAF: Addressing in another PR as requested by @imatiach-msft .

StrikerRUS:
gently ping @guolinke for the review

PR #2867, opened 2020-03-04
speed up `FindBestThresholdFromHistogram` 

guolinke:
and some code refactorings.


guolinke:
Todo: put l1/l2 into template

PR #2869, opened 2020-03-04
[ci] fixed cpplint warnings about braces

jameslamb:
This PR eliminates 5 of the 49 remaining `cpplint` issues, noted in #1990 --> https://github.com/microsoft/LightGBM/issues/1990#issuecomment-594687740

```
src/io/file_io.cpp:173:  If an else has a brace on one side, it should have it on both  [readability/braces] [5]
src/io/file_io.cpp:186:  If an else has a brace on one side, it should have it on both  [readability/braces] [5]
src/io/file_io.cpp:197:  If an else has a brace on one side, it should have it on both  [readability/braces] [5]
src/io/file_io.cpp:197:  If/else bodies with multiple statements require braces  [readability/braces] [4]
Done processing src/io/file_io.cpp
...
include/LightGBM/tree.h:154:  Line ends in whitespace.  Consider deleting these extra spaces.  [whitespace/end_of_line] [4]
```

4 of the 5 relate to handling of HDFS files in `src/io/file_io.cpp`. Instead of using an `if-else` construct, I think we can get around the linting issue by just relying on the fact that the affected functions already use multiple return statements.

Looks like the relevant lines were introduced in #2769 , @guolinke

### DIFF
--- a/include/LightGBM/tree.h
+++ b/include/LightGBM/tree.h
@@ -134,7 +134,6 @@ class Tree {
   inline int PredictLeafIndex(const double* feature_values) const;
   inline int PredictLeafIndexByMap(const std::unordered_map<int, double>& feature_values) const;
 
-
   inline void PredictContrib(const double* feature_values, int num_features, double* output);
 
   /*! \brief Get Number of leaves*/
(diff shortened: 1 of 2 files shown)


include/LightGBM/tree.h
```diff
@@ -151,7 +151,7 @@ class Tree {
   /*! \brief Get the number of data points that fall at or below this node*/
   inline int data_count(int node) const { return node >= 0 ? internal_count_[node] : leaf_count_[~node]; }
 
-  
+
```
StrikerRUS: I think this line can be removed as we usually use 1 blank line between methods.

jameslamb: even better!

PR #2887, opened 2020-03-08
[docs] update latest benchmark results

guolinke:


docs/Experiments.rst
```diff
@@ -6,6 +6,12 @@ Comparison Experiment
 
 For the detailed experiment scripts and output logs, please refer to this `repo`_.
 
+History
+^^^^^^^
+08 Mar, 2020: update according to the latest master barnch. (``xgboost_extra`` is not updated for it is too slow.)
```
guolinke:
```suggestion
08 Mar, 2020: update according to the latest master branch. (``xgboost_extra`` is not updated for it is too slow.)
```

Laurae2:
@guolinke Which commits of xgboost / LightGBM did you use for the benchmarks?

StrikerRUS:
@guolinke Very impressive!
What about `Parallel Experiment`? I think it should be updated as well or be marked as from `first version`.

guolinke:
@StrikerRUS as we didn't improve parallel algorithm, we can mark them to `first version`

PR #2888, opened 2020-03-08
fixed cpplint errors and disable warning only for VS

StrikerRUS:
Refer to #1990 and #2883.

### DIFF
--- a/include/LightGBM/meta.h
+++ b/include/LightGBM/meta.h
@@ -76,8 +76,10 @@ const int kAlignedSize = 32;
 #define SIZE_ALIGNED(t) ((t) + kAlignedSize - 1) / kAlignedSize * kAlignedSize
 
 // Refer to https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4127?view=vs-2019
-#pragma warning(disable : 4127)
+#ifdef _MSC_VER
+  #pragma warning(disable : 4127)
+#endif
 
 }  // namespace LightGBM
 
-#endif   // LightGBM_META_H_
+#endif  // LightGBM_META_H_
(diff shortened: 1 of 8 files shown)


include/LightGBM/meta.h
```diff
@@ -76,8 +76,10 @@ const int kAlignedSize = 32;
 #define SIZE_ALIGNED(t) ((t) + kAlignedSize - 1) / kAlignedSize * kAlignedSize
 
 // Refer to https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4127?view=vs-2019
-#pragma warning(disable : 4127)
+#ifdef _MSC_VER
+  #pragma warning(disable : 4127)
```
guolinke: Maybe some other warning disable could use the msc ver?

StrikerRUS: @guolinke Sorry, didn't get it. Can you please elaborate?

guolinke: I means, there are some other `#pragma warning(disable : xxxx)`, which are not inside `#ifdef _MSC_VER`.

StrikerRUS: Ah, got it! I thought that the reference to VS in the corresponding comment means that it is actually needed only  for VS. Like in
https://github.com/microsoft/LightGBM/blob/6b6709b9b3ae302d4a085aadb327540792a502f0/include/LightGBM/utils/openmp_wrapper.h#L68-L70

And I was aware only about that `pragma warning` in `openmp_wrapper` code. Should I search for others and wrap them as well?

guolinke: yeah, I remember there are some in c_api.

StrikerRUS: Done!

Issue #2890, opened 2020-03-08
[bug] SWIG Booster read/write dependent on machine locale


AlbertoEAF:
**Bug**

Reading a LightGBM model through the SWIG API changes the read model parameters depending on the machine locale, i.e., the same model is not read consistently across environments. This results in broken scores - and in some cases constant ones. 

Here is a screenshot of the same model read in two machines with different locales:
![image](https://user-images.githubusercontent.com/3976753/76172057-bb5bf080-6189-11ea-9b01-5723fee826d6.png)

And an example of training with PT locale through the SWIG/Java API (left) and in python (right):
![image](https://user-images.githubusercontent.com/3976753/76218980-010dcd00-620d-11ea-8552-e487f42d3b6f.png)

**Context**
In some cases it's easy to do trickery with the machine locale to fix it, but in others it has repercussions on the running processes, or controlling the environment is not even possible.
This brings hurdles to deploying a model consistently across varied environments.

Whilst using the machine locale settings to read/write **data** files stored with different representations might be ok or even desired, regarding the **model** file integrity, our model format should **not**  depend on the localization settings.

Notice this kind of error doesn't occur with the Python bindings, so with the SWIG Java interface we should strive for the same consistent behaviour (not affected by locale).

**Related issues**

The following issues that seemed to be JEP only, are actually SWIG/Java-related:

https://github.com/microsoft/LightGBM/issues/2500
https://github.com/ninia/jep/issues/205

PR #2891, opened 2020-03-08
Fix Booster read/write locale dependency

AlbertoEAF:
This request fixes the bug mentioned on the issue https://github.com/microsoft/LightGBM/issues/2890.

In C++, the process is by default is initialized to use the minimal "C" locale, however, in Java, when using the SWIG wrappers, this default is broken and the current machine locale is used instead.

Since during model read/write we rely on many C/C++ functions, some of which are 100% dependant on the locale, such as `std::stod()`, overriding the locale to a sane default is required.
Instead of setting the locale globally across the full process, as we might want to still read/write datasets according to the locale, a class to manage the locale context `LocaleContext` was created.
We **only** activate/deactivate that context (to force the "C" locale) **during model reads/writes**, to ensure our model format is independent of any locale changes in the machine or calling processes.

Motivation for this:

- In Python, the locale is always "C" by default, still as seen in the comment https://github.com/ninia/jep/issues/205#issuecomment-544078377, it is possible to break the model by changing locales.
- In Java, by using the SWIG interface, there isn't even a safety net like in Python, of using the "C" locale by default.

To avoid such issues, the fix should be done on the core library, so no bindings can be affected by locale issues, whatever happens, and this way the model format is always consistent.


AlbertoEAF:
Hello @guolinke and @chivee, 

can you tell me if this looks good to you and if so, to approve/revise?

guolinke:
@StrikerRUS any style issues?

include/LightGBM/utils/LocaleContext.h
```diff
@@ -0,0 +1,38 @@
+#ifndef __LOCALE_CONTEXT_H__
```
StrikerRUS: Please add copyright message.
```
/*!
 * Copyright (c) 2020 Microsoft Corporation. All rights reserved.
 */
```

include/LightGBM/utils/LocaleContext.h
```diff
@@ -0,0 +1,38 @@
...
+#include <locale>
+
+/**
+ * Class to override the program global locale during this object lifetime.
+ * After the object is destroyed, the locale is returned to its original state.
+ *
+ * @warn This is not thread-safe.
+*/
+class LocaleContext
+{
```
StrikerRUS:
```
{ should almost always be at the end of the previous line  [whitespace/braces] [4]
```

include/LightGBM/utils/LocaleContext.h
```diff
@@ -0,0 +1,38 @@
...
+public:
```
StrikerRUS:
```
public: should be indented +1 space inside class LocaleContext  [whitespace/indent] [3]
```

include/LightGBM/utils/LocaleContext.h
```diff
@@ -0,0 +1,38 @@
...
+
```
StrikerRUS:
```
Do not leave a blank line after "public:"  [whitespace/blank_line] [3]
```

include/LightGBM/utils/LocaleContext.h
```diff
@@ -0,0 +1,38 @@
...
+  /**
+   * Override the current program global locale during this object lifetime.
+   *
+   * @param target_locale override the locale to this locale setting.
+   * @warn This is not thread-safe.
+   * @note This doesn't override cout, cerr, etc.
+   */
+  LocaleContext(const char* target_locale = "C") {
```
StrikerRUS:
```
Constructors callable with one argument should be marked explicit.  [runtime/explicit] [5]
```

include/LightGBM/utils/LocaleContext.h
```diff
@@ -0,0 +1,38 @@
...
+    std::locale::global(std::locale(target_locale));
+  }
+
+  /**
+   * Restores the old global locale.
+   */
+  ~LocaleContext() {
+    std::locale::global(_saved_global_locale);
+  }
+private:
```
StrikerRUS:
```
private: should be indented +1 space inside class LocaleContext  [whitespace/indent] [3]
```

include/LightGBM/utils/LocaleContext.h
```diff
@@ -0,0 +1,38 @@
...
+  std::locale _saved_global_locale; //!< Stores global locale at initialization.
+};
+
+#endif // __LOCALE_CONTEXT_H__
```
StrikerRUS:
```
At least two spaces is best between code and comments  [whitespace/comments] [2]
```

include/LightGBM/utils/LocaleContext.h
StrikerRUS: rename file to `locale_context.h`

StrikerRUS:
```suggestion
#ifndef LIGHTGBM_LOCALE_CONTEXT_H_
```

include/LightGBM/utils/LocaleContext.h
```diff
@@ -0,0 +1,38 @@
+#ifndef __LOCALE_CONTEXT_H__
+#define __LOCALE_CONTEXT_H__
```
StrikerRUS:
```suggestion
#define LIGHTGBM_LOCALE_CONTEXT_H_
```

include/LightGBM/utils/LocaleContext.h
StrikerRUS:
```suggestion
#endif  // LIGHTGBM_LOCALE_CONTEXT_H_
```

src/c_api.cpp
```diff
@@ -17,6 +17,7 @@
 #include <LightGBM/utils/openmp_wrapper.h>
 #include <LightGBM/utils/random.h>
 #include <LightGBM/utils/threading.h>
+#include <LightGBM/utils/LocaleContext.h>
```
StrikerRUS: Follow alphabetical order.

include/LightGBM/utils/LocaleContext.h
```diff
@@ -0,0 +1,38 @@
...
+
+#include <clocale>
+#include <locale>
+
+/**
+ * Class to override the program global locale during this object lifetime.
+ * After the object is destroyed, the locale is returned to its original state.
+ *
+ * @warn This is not thread-safe.
+*/
```
StrikerRUS:
```suggestion
 */
```

include/LightGBM/utils/LocaleContext.h
```diff
@@ -0,0 +1,38 @@
+#ifndef __LOCALE_CONTEXT_H__
+#define __LOCALE_CONTEXT_H__
+
+#include <clocale>
+#include <locale>
+
+/**
```
StrikerRUS:
```suggestion
/*!
```

include/LightGBM/utils/LocaleContext.h
```diff
@@ -0,0 +1,38 @@
...
+ * Class to override the program global locale during this object lifetime.
+ * After the object is destroyed, the locale is returned to its original state.
+ *
+ * @warn This is not thread-safe.
+*/
+class LocaleContext
+{
+public:
+
+  /**
```
StrikerRUS:
```suggestion
  /*!
```

include/LightGBM/utils/LocaleContext.h
```diff
@@ -0,0 +1,38 @@
...
+   *
+   * @param target_locale override the locale to this locale setting.
+   * @warn This is not thread-safe.
+   * @note This doesn't override cout, cerr, etc.
+   */
+  LocaleContext(const char* target_locale = "C") {
+    std::locale::global(std::locale(target_locale));
+  }
+
+  /**
```
StrikerRUS:
```suggestion
  /*!
```

include/LightGBM/utils/LocaleContext.h
```diff
@@ -0,0 +1,38 @@
...
+   * Restores the old global locale.
+   */
+  ~LocaleContext() {
+    std::locale::global(_saved_global_locale);
+  }
+private:
+  std::locale _saved_global_locale; //!< Stores global locale at initialization.
```
StrikerRUS:
```
At least two spaces is best between code and comments  [whitespace/comments] [2]
```
```suggestion
  std::locale _saved_global_locale;  // Stores global locale at initialization.
```

AlbertoEAF: This was to enrich the doxygen-style comments for the member data.

I'm applying the suggestion anyway.

AlbertoEAF:
Changed everything you requested @StrikerRUS :)

Anything else?

AlbertoEAF:
Thanks for merging @StrikerRUS :)

PR #2901, opened 2020-03-12
[R-package] Use Rprintf for logging in the R package (fixes #1440, fixes #1909)

### LINKED ISSUE #1440: [R-package] Ability to control the printed messages?

### LINKED ISSUE #1909: [R} CRAN is unhappy about calls to stderr in source code

jameslamb:
This PR proposes a change to get the R package closer to CRAN's standards (#629). See #1440 and #1909 for background. The basic issue, as documented in [Writing R Extensions](https://cran.r-project.org/doc/manuals/r-release/R-exts.html), is this:

> Compiled code is checked for symbols corresponding to functions which might terminate R or write to stdout/stderr instead of the console. 

To get LightGBM into compliance with this check, I propose using `Rprintf()`  and  `Rvprintf()` from R's headers. This is [the approach CRAN recommends](https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Printing)

> The most useful function for printing from a C routine compiled into R is Rprintf. This is used in exactly the same way as printf, but is guaranteed to write to R’s output (which might be a GUI console rather than a file, and can be re-directed by sink). It is wise to write complete lines (including the "\n") before returning to R. It is defined in R_ext/Print.h.

**Notes for Reviewers**

1.  `cmake/modules/FindLibR.cmake` was included to find R and its headers on any system. It borrows heavily from [`xgboost`'s implementation](https://github.com/dmlc/xgboost/blob/master/cmake/modules/FindLibR.cmake).
2. I've tested this approach on [my personal Travis](https://travis-ci.org/github/jameslamb/LightGBM/builds/661561425?utm_medium=notification&utm_source=github_status) and it seems to be working. One task (`mpi`) is failing but I think it is unrelated.
3. I ran `R CMD check lightgbm_2.3.2.tar.gz --as-cran` and  can confirm that this NOTE is gone!

```shell
File ‘/home/travis/build/microsoft/LightGBM/lightgbm.Rcheck/lightgbm/libs/lib_lightgbm.so’:
  Found ‘printf’, possibly from ‘printf’ (C)
  Found ‘putchar’, possibly from ‘putchar’ (C)
  Found ‘stderr’, possibly from ‘stderr’ (C)
  Found ‘stdout’, possibly from ‘stdout’ (C)

```


R-package/src/cmake/modules/FindLibR.cmake
```diff
@@ -0,0 +1,209 @@
...
+    find_program(
+      LIBR_EXECUTABLE
+      NO_DEFAULT_PATH
+      HINTS "${CMAKE_CURRENT_BINARY_DIR}" "/usr/bin" "/usr/lib/"
+      NAMES R R.exe
+    )
+
+    if(LIBR_EXECUTABLE MATCHES ".*lightgbm\\.Rcheck.*")
+      message(FATAL_ERROR "Hit this block. '${LIBR_EXECUTABLE}'")
+    endif()
```
jameslamb: This section is something I added that you won't find in the `xgboost`  script. It solves an issue that only matters at testing time...basically inside of  `R CMD check`, CRAN actually runs a different command called `RD CMD  check` which uses a special version of `R` / `R.exe`.  We faced this [in pkgnet](https://github.com/uptake/pkgnet/pull/108/files#r230631169) as well.

From [Writing R Extensions](https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Writing-portable-packages)

> Do not invoke R by plain R, Rscript or (on Windows) Rterm in your examples, tests, vignettes, makefiles or other scripts. As pointed out in several places earlier in this manual, use something like

>    "$(R_HOME)/bin/Rscript"
>    "$(R_HOME)/bin$(R_ARCH_BIN)/Rterm"

> with appropriate quotes (as, although not recommended, R_HOME can contain spaces).

The hints that I gave here will only work on Unix-alike operating systems. When I do #2335 it will get changed. But since this only matters for testing (in the context of `R CMD check`), not for normal use, I think it's ok to merge.

Without this block, the builds will fail with this error:

> 'R' should not be used without a path -- see par. 1.6 of the manual

include/LightGBM/R_object_helper.h
```diff
@@ -39,9 +39,9 @@
 
   // 64bit pointer
   #if INTPTR_MAX == INT64_MAX
-    typedef int64_t R_xlen_t;
+    typedef int64_t xlen_t;
```
jameslamb: I had to rename this type to avoid conflicts with  `R_xlen_t` defined in R's headers.

include/LightGBM/utils/log.h
```diff
@@ -14,6 +14,13 @@
 #include <iostream>
 #include <stdexcept>
 
+#ifdef LGB_R_BUILD
+  #define R_NO_REMAP
+  #define R_USE_C99_IN_CXX
+  #include <R_ext/Error.h>
+  #include <R_ext/Print.h>
+#endif
+
```
jameslamb: for those who are new  to it:

* `R_NO_REMAP`:  Do not change symbols like `Rf_error()` to `error()`
* `R_USE_C99_IN_CXX`: Make  sure that `Rvprintf()` is defined. ([reference](https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Printing))

R-package/src/cmake/modules/FindLibR.cmake
```diff
@@ -0,0 +1,209 @@
+# CMake module used to fine the location of R's
```
jameslamb:
```suggestion
# CMake module used to find the location of R's
```

jameslamb:
Travis is struggling today, seeing a lot of these

image

R-package/src/cmake/modules/FindLibR.cmake
```diff
@@ -0,0 +1,209 @@
...
+    # 'R' should not be used without a path -- see par. 1.6 of the manual
+    find_program(
+      LIBR_EXECUTABLE
+      NO_DEFAULT_PATH
+      HINTS "${CMAKE_CURRENT_BINARY_DIR}" "/usr/bin" "/usr/lib/"
+      NAMES R R.exe
+    )
+
+    if(LIBR_EXECUTABLE MATCHES ".*lightgbm\\.Rcheck.*")
+      message(FATAL_ERROR "Hit this block. '${LIBR_EXECUTABLE}'")
```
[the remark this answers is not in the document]
jameslamb: Yep definitely. I had this in during debugging, meant to come back and improve it. I can do that.

R-package/src/cmake/modules/FindLibR.cmake
```diff
@@ -0,0 +1,209 @@
...
+        OUTPUT_VARIABLE LIBR_HOME
+      )
+    endif()
+
+    # if R executable not available, query R_HOME path from registry
+    if(NOT LIBR_HOME)
+
+      # Windows users might want to change this to their R version:
+      if(NOT R_VERSION)
+        set(R_VERSION "3.4.1")
```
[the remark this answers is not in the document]
jameslamb: oh nice find!  Ok let me try with that.

R-package/src/cmake/modules/FindLibR.cmake
```diff
@@ -0,0 +1,209 @@
...
+      endif()
+
+      get_filename_component(LIBR_HOME
+        "[HKEY_LOCAL_MACHINE\\SOFTWARE\\R-core\\R\\${R_VERSION};InstallPath]"
```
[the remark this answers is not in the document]
jameslamb: Sure I can try!

To be clear, the Windows parts of this were directly copied from `xgboost`'s version of this script. I didn't write them from scratch. So for some of these suggestions, if they work we may want to contribute them back to `xgboost` too!

R-package/src/cmake/modules/FindLibR.cmake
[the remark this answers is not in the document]
jameslamb: ok!  Let me try that

R-package/src/cmake/modules/FindLibR.cmake
```diff
@@ -0,0 +1,209 @@
...
+#  LIBR_HOME
+#  LIBR_EXECUTABLE
+#  LIBR_INCLUDE_DIRS
+#  LIBR_LIB_DIR
+#  LIBR_CORE_LIBRARY
+# and a cmake function to create R.lib for MSVC
+
+if(NOT R_ARCH)
+  if("${CMAKE_SIZEOF_VOID_P}" STREQUAL "4")
+    set(R_ARCH "i386")
```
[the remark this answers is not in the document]
jameslamb: Oh interesting, I thought we were when I saw https://github.com/microsoft/LightGBM/blob/master/include/LightGBM/R_object_helper.h#L41 but forgot about that doc.

If we're not supporting 32-bit windows for R I'll remove this part

R-package/src/cmake/modules/FindLibR.cmake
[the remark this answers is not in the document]
jameslamb: Actually....since we're already using `install.libs.R`, R had to have already been found for this code to even be invoked. I'm going to try just using R's built-in `R.Version()` and passing it in as a cmake variable, e.g. `cmake .. -DCMAKE_R_VERSION=3.6.1`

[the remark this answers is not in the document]
jameslamb: ^  I tried this (just passing it in from R) in https://github.com/microsoft/LightGBM/pull/2901/commits/9a337d4f4d4004346d7d8d64ced4a0b719fc07ae and it seemed to work locally. Let's see if CI agrees with me.

I think that is a better solution than consulting the registry, since the things you linked mention that creating those registry entries is optional. I think the average R on Windows user will probably just have done an install with all the default settings.

R-package/src/cmake/modules/FindLibR.cmake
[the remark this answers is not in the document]
jameslamb: added in https://github.com/microsoft/LightGBM/pull/2901/commits/9a337d4f4d4004346d7d8d64ced4a0b719fc07ae, good find.

jameslamb:
I just tried building this branch on my Windows laptop (Windows 10, R3.6.1, Visual Studio 16 2019) with `Rscript build_r.R` and can confirm this worked!

The first time I ran I did hit https://github.com/jameslamb/LightGBM/blob/feat/r-header/R-package/src/cmake/modules/FindLibR.cmake#L41, but that was easy to fix by following the error message and adding the Rtools minGW `bin/` directory to the path.

R-package/src/cmake/modules/FindLibR.cmake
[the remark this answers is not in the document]
jameslamb: I see this now: https://github.com/microsoft/LightGBM/blob/master/R-package/src/install.libs.R#L7

Guess we  are not currently supporting 32-bit operating systems for R.

@StrikerRUS what do you think about this --> https://github.com/microsoft/LightGBM/pull/2901/commits/4c154005c89e0d6953af0d6eddcf679d571b4642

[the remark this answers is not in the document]
StrikerRUS: I guess it should be

```
if(NOT R_ARCH)
  if("${CMAKE_SIZEOF_VOID_P}" STREQUAL "4")
    set(R_ARCH "i386")
  else()
    set(R_ARCH "x64")
  endif()
endif()

if(NOT ("${R_ARCH}" STREQUAL "x64"))
  message(FATAL_ERROR "LightGBM's R package currently only supports 64-bit operating systems")
endif()
```
to cover the case when R_ARCH is set earlier.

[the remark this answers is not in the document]
jameslamb: ah ok you're right, that's cleaner.

R-package/src/cmake/modules/FindLibR.cmake
```diff
@@ -0,0 +1,224 @@
...
+#  LIBR_LIB_DIR
+#  LIBR_CORE_LIBRARY
+# and a cmake function to create R.lib for MSVC
+
+if(NOT R_ARCH)
+  if("${CMAKE_SIZEOF_VOID_P}" STREQUAL "4")
+    message(FATAL_ERROR "LightGBM's R package currently only supports 64-bit operating systems")
+  endif()
+  set(R_ARCH "x64")
+endif()
```
jameslamb:
```suggestion
if(NOT R_ARCH)
  if("${CMAKE_SIZEOF_VOID_P}" STREQUAL "4")
    set(R_ARCH "i386")
  else()
    set(R_ARCH "x64")
  endif()
endif()

if(NOT ("${R_ARCH}" STREQUAL "x64"))
  message(FATAL_ERROR "LightGBM's R package currently only supports 64-bit operating systems")
endif()
```

jameslamb: @StrikerRUS check it out! In case you haven't seen, GitHub now supports multi-line suggestions 😀

StrikerRUS: Yeah, cool feature!

R-package/src/cmake/modules/FindLibR.cmake
```diff
@@ -0,0 +1,229 @@
...
+    get_filename_component(_LIBR_LIBRARIES "${LIBR_LIBRARIES}" REALPATH)
+    get_filename_component(_LIBR_LIBRARIES_DIR "${_LIBR_LIBRARIES}" DIRECTORY)
+    set(LIBR_EXECUTABLE "${_LIBR_LIBRARIES_DIR}/../bin/R")
+    execute_process(
+      COMMAND ${LIBR_EXECUTABLE} "--slave" "--vanilla" "-e" "cat(R.home())"
+      OUTPUT_VARIABLE LIBR_HOME
+    )
+    set(LIBR_HOME ${LIBR_HOME} CACHE PATH "R home directory")
+    set(LIBR_INCLUDE_DIRS "${LIBR_HOME}/include" CACHE PATH "R include directory")
+    set(LIBR_LIB_DIR "${LIBR_HOME}/lib" CACHE PATH "R lib directory")
```
StrikerRUS: Why can't we do the same for macOS?
https://github.com/jameslamb/LightGBM/blob/2858c1c4a0e1c91f564c0222b815e42f2dd1c0ec/R-package/src/cmake/modules/FindLibR.cmake#L125-L135

R-package/src/cmake/modules/FindLibR.cmake
```diff
@@ -0,0 +1,229 @@
...
+    endif()
+
+    # set exe location based on R_ARCH
+    if(NOT LIBR_EXECUTABLE)
+      set(LIBR_EXECUTABLE "${LIBR_HOME}/bin/${R_ARCH}/R.exe")
+    endif()
+
+    # set other R paths based on home path
+    set(LIBR_INCLUDE_DIRS "${LIBR_HOME}/include")
+    set(LIBR_LIB_DIR "${LIBR_HOME}/bin/${R_ARCH}")
```
StrikerRUS: Why can't we do the same for Windows?
https://github.com/jameslamb/LightGBM/blob/2858c1c4a0e1c91f564c0222b815e42f2dd1c0ec/R-package/src/cmake/modules/FindLibR.cmake#L125-L135

R-package/src/cmake/modules/FindLibR.cmake
```diff
@@ -0,0 +1,229 @@
...
+
+    message(STATUS "LIBR_HOME [${LIBR_HOME}]")
+    message(STATUS "LIBR_EXECUTABLE [${LIBR_EXECUTABLE}]")
+    message(STATUS "LIBR_INCLUDE_DIRS [${LIBR_INCLUDE_DIRS}]")
+    message(STATUS "LIBR_LIB_DIR [${LIBR_LIB_DIR}]")
+    message(STATUS "LIBR_CORE_LIBRARY [${LIBR_CORE_LIBRARY}]")
```
StrikerRUS: We output this info in the main CMakeLists. I don't think we should do it twice. Moreover, we haven't set `LIBR_CORE_LIBRARY` yet.

R-package/src/cmake/modules/FindLibR.cmake
```diff
@@ -0,0 +1,229 @@
...
+  endif()
+
+endif()
+
+# look for the core R library
+find_library(LIBR_CORE_LIBRARY NAMES R
+  HINTS "${CMAKE_CURRENT_BINARY_DIR}" "${LIBR_LIB_DIR}" "${LIBR_HOME}/bin" "${LIBR_LIBRARIES}")
+
+if(LIBR_CORE_LIBRARY-NOTFOUND)
+  message(STATUS "Could not find R core shared library.")
```
StrikerRUS: Is this library required?

R-package/src/cmake/modules/FindLibR.cmake
```diff
@@ -0,0 +1,229 @@
...
+# detection for OSX
+if(APPLE)
+
+  find_library(LIBR_LIBRARIES R)
+
+  if(LIBR_LIBRARIES MATCHES ".*\\.framework")
+    set(LIBR_HOME "${LIBR_LIBRARIES}/Resources" CACHE PATH "R home directory")
+    set(LIBR_INCLUDE_DIRS "${LIBR_HOME}/include" CACHE PATH "R include directory")
+    set(LIBR_EXECUTABLE "${LIBR_HOME}/R" CACHE PATH "R executable")
+    set(LIBR_LIB_DIR "${LIBR_HOME}/lib" CACHE PATH "R lib directory")
```
StrikerRUS: To deduplicate code.
https://github.com/jameslamb/LightGBM/blob/2858c1c4a0e1c91f564c0222b815e42f2dd1c0ec/R-package/src/cmake/modules/FindLibR.cmake#L210-L214
```suggestion
    set(LIBR_HOME "${LIBR_LIBRARIES}/Resources")
    set(LIBR_INCLUDE_DIRS "${LIBR_HOME}/include")
    set(LIBR_EXECUTABLE "${LIBR_HOME}/R")
    set(LIBR_LIB_DIR "${LIBR_HOME}/lib")
```

R-package/src/cmake/modules/FindLibR.cmake
StrikerRUS: To deduplicate code.
https://github.com/jameslamb/LightGBM/blob/2858c1c4a0e1c91f564c0222b815e42f2dd1c0ec/R-package/src/cmake/modules/FindLibR.cmake#L210-L214
```suggestion
    set(LIBR_INCLUDE_DIRS "${LIBR_HOME}/include")
    set(LIBR_LIB_DIR "${LIBR_HOME}/lib")
```

CMakeLists.txt
```diff
@@ -72,6 +72,13 @@ if(USE_R35)
 endif(USE_R35)
 
 if(BUILD_FOR_R)
+    list(APPEND CMAKE_MODULE_PATH "${lightgbm_SOURCE_DIR}/cmake/modules")
```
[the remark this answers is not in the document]
jameslamb: I saw this [all over xgboost's cmake scripts](https://github.com/dmlc/xgboost/search?q=xgboost_source_dir&unscoped_q=xgboost_source_dir). I thought it might be that this is defined for `<project>_SOURCE_DIR` where `<project>` is a name you gave to `PROJECT()`, like we do at https://github.com/microsoft/LightGBM/blob/master/CMakeLists.txt#L7

The documentation on `SOURCE_DIR` is pretty minimal https://cmake.org/cmake/help/v3.16/prop_dir/SOURCE_DIR.html 😅

CMakeLists.txt
```diff
@@ -72,6 +72,13 @@ if(USE_R35)
...
+    set(CMAKE_R_VERSION ${CMAKE_R_VERSION})
```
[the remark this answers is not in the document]
jameslamb: I think this line is redundant, you're right. I left it in from debugging. Let me try removing it.

R-package/src/cmake/modules/FindLibR.cmake
jameslamb: oh yeah I like that! thank you

R-package/src/cmake/modules/FindLibR.cmake
jameslamb: hmmmm let me try without it, it might not be

R-package/src/cmake/modules/FindLibR.cmake
jameslamb: I'm not sure about this one @StrikerRUS . I think we still need to set  `LIBR_HOME`, since it is used further down.

Did you mean to just remove the `CACHE PATH` stuff since that's done all at the bottom? https://github.com/jameslamb/LightGBM/blob/feat/r-header/R-package/src/cmake/modules/FindLibR.cmake#L210

R-package/src/cmake/modules/FindLibR.cmake
jameslamb: Good idea! I think main CMakeLists is probably the right place. Because that way we're only printing after everything has been found, and the printing will also tell us if variables are being set here in the way that the main CmakeLists expects

R-package/src/cmake/modules/FindLibR.cmake
jameslamb: do you mean "for non-Windows"? This block is in a part where  we are working with Windows.

I have seen that on Window's, R's source, shared library, and executables are always bundled together like this. On different Linux systems, R's installation follows the conventions of that OS and doesn't necessarily keep everything together.

For example on Debian (the [OS used by the Dockerfile we provide](https://github.com/jameslamb/LightGBM/blob/2858c1c4a0e1c91f564c0222b815e42f2dd1c0ec/docker/dockerfile-r#L2) I've seen this

```
-- LIBR_HOME [/usr/local/lib/R]
-- LIBR_EXECUTABLE [/usr/local/bin/R]
-- LIBR_INCLUDE_DIRS [/usr/local/lib/R/include]
-- LIBR_LIB_DIR [/usr/local/lib/R/lib]
```

R-package/src/cmake/modules/FindLibR.cmake
jameslamb: wait yes, I think it is needed by the linker.

https://github.com/jameslamb/LightGBM/blob/feat/r-header/CMakeLists.txt#L311

CMakeLists.txt
```diff
@@ -72,6 +72,13 @@ if(USE_R35)
...
+    message(STATUS "passed-in R version: " ${CMAKE_R_VERSION})
```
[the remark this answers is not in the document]
jameslamb: ah yeah this is leftover from debugging. I wanted to check that `install.libs.R` was passing `CMAKE_R_VERSION` to `CmakeLists` correctly, and then that guard you mentioned checks that `FindLibR.cmake`  was getting the right information from `CMakeLists`.

I think now I can eliminate the one here and just leave the one in `FindLibR.cmake`, to be sure that variable made it all the way through

CMakeLists.txt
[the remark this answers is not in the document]
jameslamb: found the docs! https://cmake.org/cmake/help/v3.0/variable/PROJECT-NAME_SOURCE_DIR.html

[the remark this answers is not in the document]
jameslamb: there are [a bunch of these](https://cmake.org/cmake/help/v3.0/manual/cmake-variables.7.html)

image

R-package/src/cmake/modules/FindLibR.cmake
```diff
@@ -0,0 +1,229 @@
...
+
+  if(NOT EXISTS "${LIBR_LIB_DIR}")
+    message(FATAL_ERROR "LIBR_LIB_DIR was not set!")
+  endif()
+
+  find_program(GENDEF_EXE gendef)
+  find_program(DLLTOOL_EXE dlltool)
+
+  if(NOT GENDEF_EXE OR NOT DLLTOOL_EXE)
+    message(FATAL_ERROR "\nEither gendef.exe or dlltool.exe not found!\
```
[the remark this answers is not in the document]
jameslamb: oh you're right! good call

R-package/src/cmake/modules/FindLibR.cmake
```diff
@@ -0,0 +1,229 @@
...
+      \nDo you have Rtools installed with its MinGW's bin/ in PATH?")
+  endif()
+
+  # extract symbols from R.dll into R.def and R.lib import library
+  execute_process(COMMAND gendef
```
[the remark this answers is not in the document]
jameslamb: Oh interesting. This is something I took straight from `xgboost`, but you're right I think it would be better to use the cache variable. From [the cmake docs](https://cmake.org/cmake/help/v3.12/command/find_program.html):

> `find_program (<VAR> name ...)`

> A cache entry named by <VAR> is created to store the result of this command. If the program is found the result is stored in the variable and the search will not be repeated unless the variable is cleared. If nothing is found, the result will be <VAR>-NOTFOUND, and the search will be attempted again the next time find_program is invoked with the same variable.

I think the `execute_process()` calls then must be ignoring the cache and searching for this again! But also I don't think your selection is correct, it might need to be `${GENDEF_EXE}`, kind of like [in this example](https://gitlab.kitware.com/cmake/community/-/wikis/FAQ#how-can-i-apply-resources-on-mac-os-x-automatically)

jameslamb:
I just rebased to `master` since there have been a lot of changes on there since this PR was opened. Made sure to pull the suggestions committed in the browser before doing that, so they didn't get reverted by the force push.

include/LightGBM/utils/log.h
```diff
@@ -107,19 +112,33 @@ class Log {
     vsprintf(str_buf, format, val);
 #endif
     va_end(val);
-    fprintf(stderr, "[LightGBM] [Fatal] %s\n", str_buf);
-    fflush(stderr);
-    throw std::runtime_error(std::string(str_buf));
+
+    // R code should write back to R's error stream,
+    // otherwise to stderr
+    #ifndef LGB_R_BUILD
```
Laurae2: We should agree on preprocessor directive indentations on a later PR. ping @guolinke @StrikerRUS @jameslamb for later.

include/LightGBM/utils/log.h
Laurae2: Note: `R_USE_C99_IN_CXX` is defined in `R_ext/Print.h` in R 4.0. Will this break when using R 4.0?

jameslamb:
> See R-devel (R 4.0): https://cran.r-project.org/doc/manuals/r-devel/NEWS.html in C-level facilities:
> 
> > Header ‘R_ext/Print.h’ defines R_USE_C99_IN_CXX and hence exposes Rvprintf and REvprintf if used with a C++11 (or later) compiler.
> 
> Do we need a workaround for it?

Still investigating, but for the sake of documenting the discussion...`divest` (as one example), [uses this same pattern](https://github.com/jonclayden/divest/blob/5d8509ab887155b871946dc4047752481a0d2dcf/src/dcm2niix/print.h#L12-L13) and recently passed CRAN nightly checks against `r-devel` ([which is 4.0.0 now](https://cran.r-project.org/bin/windows/base/rdevel.html)) https://cran.r-project.org/web/checks/check_results_divest.html

jameslamb:
Thanks for the review @Laurae2 ! I won't merge until @StrikerRUS and I agree on some of the lingering comments on `FindLibR.cmake`

CMakeLists.txt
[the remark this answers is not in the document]
StrikerRUS: I think we should stick to the currently using [`PROJECT_SOURCE_DIR`](https://cmake.org/cmake/help/v3.0/variable/PROJECT_SOURCE_DIR.html), as we have only one project.

CMakeLists.txt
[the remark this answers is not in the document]
StrikerRUS: Good idea!

R-package/src/cmake/modules/FindLibR.cmake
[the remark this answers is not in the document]
StrikerRUS: Oh, sure! Forgot about enclosing `${...}`.

R-package/src/cmake/modules/FindLibR.cmake
StrikerRUS: I mean, use code from the link https://github.com/jameslamb/LightGBM/blob/2858c1c4a0e1c91f564c0222b815e42f2dd1c0ec/R-package/src/cmake/modules/FindLibR.cmake#L125-L135
```
    # ask R for the include dir
    execute_process(
      COMMAND ${LIBR_EXECUTABLE} "--slave" "--no-save" "-e" "cat(R.home('include'))"
      OUTPUT_VARIABLE LIBR_INCLUDE_DIRS
    )

    # ask R for the lib dir
    execute_process(
      COMMAND ${LIBR_EXECUTABLE} "--slave" "--no-save" "-e" "cat(R.home('lib'))"
      OUTPUT_VARIABLE LIBR_LIB_DIR
    )
```
as a try to find dirs on Windows and macOS as well. At present this code is used only for Unix.

In other words, why don't we use `R.home('include')` and `R.home('lib')` on Windows and macOS?

R-package/src/cmake/modules/FindLibR.cmake
StrikerRUS: Agree!

R-package/src/cmake/modules/FindLibR.cmake
StrikerRUS: Maybe move logging part to CMakeLists as per https://github.com/microsoft/LightGBM/pull/2901#discussion_r393696132?

As this library is required, building process will fail without it anyway. Seems that this message doesn't make a lot of sense.

R-package/src/cmake/modules/FindLibR.cmake
StrikerRUS: > Did you mean to just remove the CACHE PATH stuff since that's done all at the bottom?

Yeah, exactly! For the sake of deduplication.
But without `CACHE PATH` that line becomes
```
set(LIBR_HOME ${LIBR_HOME})
```
I don't think it makes any sense, just like https://github.com/microsoft/LightGBM/pull/2901#discussion_r393291279

R-package/src/cmake/modules/FindLibR.cmake
jameslamb: ohhhhhh ok. Yeah that's a good idea! Maybe I can move things around to reduce duplication if we do that. Push all the complexity back to R and out of my slapped-together `.cmake` script 😀

R-package/src/cmake/modules/FindLibR.cmake
jameslamb: ah ok true! Yes I can do that.

R-package/src/cmake/modules/FindLibR.cmake
jameslamb: ha oh true, good eye. That was unnecessary to include. Suggestion accepted!

CMakeLists.txt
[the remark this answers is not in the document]
jameslamb: that is very reasonable! I'll change it

R-package/src/cmake/modules/FindLibR.cmake
```diff
@@ -0,0 +1,229 @@
...
+    "-" "${LIBR_LIB_DIR}/R.dll"
+    OUTPUT_FILE "${CMAKE_CURRENT_BINARY_DIR}/R.def")
+  execute_process(COMMAND dlltool
```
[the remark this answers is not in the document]
jameslamb: I'm going to ignore this suggestion and do `${DLLTOOL_EXE}` in the next commit

R-package/src/cmake/modules/FindLibR.cmake
jameslamb: Actually I just decided to get rid of this message, I think you're right that it's not providing any value

R-package/src/cmake/modules/FindLibR.cmake
jameslamb: just updated in https://github.com/microsoft/LightGBM/commit/9adc7a94fafeacf29e0bd88ffa320299406bbd65

I also rebased to `master` since a lot of R code was touched in #2803

R-package/src/cmake/modules/FindLibR.cmake
jameslamb: removed in https://github.com/microsoft/LightGBM/commit/9adc7a94fafeacf29e0bd88ffa320299406bbd65

R-package/src/cmake/modules/FindLibR.cmake
jameslamb: @StrikerRUS ok I think I was able to take your suggestions and simplify this a lot! https://github.com/microsoft/LightGBM/pull/2901/commits/20f845255513a862b61ddd236b4bb6633a3ba697

Now each OS-specific block just tries to find `LIBR_EXECUTABLE`, and then re-uses that to get the other variables. This worked locally on my Mac (with both `Rscript build_r.R` and `R CMD CHECK`). Going to check on my Windows 10 laptop right now

jameslamb: Ok I tested with Windows 10, compiling with Visual Studio 16 2019. I made a few small changes (https://github.com/microsoft/LightGBM/pull/2901/commits/4375584f43badfc8b83efe6f2c752700195c7c5e), but it works!

jameslamb:
@StrikerRUS I saw you resolved the comment threads on `FindLibR.cmake`.  Am I ok to merge this?

Thanks for the review! btw I made PR into `xgboost`  with one of the improvements you suggested: https://github.com/dmlc/xgboost/pull/5427

StrikerRUS:
@jameslamb Please give me some time to make a second round of review.

jameslamb:
> @jameslamb Please give me some time to make a second round of review.

sure no hurry at all! Glad that I asked

R-package/src/cmake/modules/FindLibR.cmake
```diff
@@ -0,0 +1,199 @@
...
+endfunction(create_rlib_for_msvc)
+
+# R version information is used to search for R's libraries in
+# the registry on Windows. Since this code is orchestrated by
+# an R script (src/install.libs.R), that script uses R's built-ins to
+# find the version of R and pass it through as a CMake variable
+if(CMAKE_R_VERSION)
+  message("R version passed into FindLibR.cmake: ${CMAKE_R_VERSION}")
+else()
+  message(FATAL_ERROR "Expected CMAKE_R_VERSION to be passed in but none was provided. Check src/install.libs.R")
```
StrikerRUS: I think we can require ``R_VERSION`` only on Windows for the sake of general usage of FindLibrR.
```suggestion
elseif(WIN32)
  message(FATAL_ERROR "Expected CMAKE_R_VERSION to be passed in on Windows but none was provided. Check src/install.libs.R")
```

R-package/src/cmake/modules/FindLibR.cmake
```diff
@@ -0,0 +1,199 @@
...
+    set(LIBR_HOME "${LIBR_LIBRARIES}/Resources")
+    set(LIBR_EXECUTABLE "${LIBR_HOME}/R")
+  else()
+    get_filename_component(_LIBR_LIBRARIES "${LIBR_LIBRARIES}" REALPATH)
+    get_filename_component(_LIBR_LIBRARIES_DIR "${_LIBR_LIBRARIES}" DIRECTORY)
+    set(LIBR_EXECUTABLE "${_LIBR_LIBRARIES_DIR}/../bin/R")
+  endif()
+
+# Unix
+elseif(UNIX)
```
StrikerRUS: I believe we can omit this comment 😃

R-package/src/cmake/modules/FindLibR.cmake
```diff
@@ -0,0 +1,199 @@
...
+    if(LIBR_EXECUTABLE MATCHES ".*lightgbm\\.Rcheck.*")
+      message(FATAL_ERROR "If you are seeing this error, it means you are running R CMD check and R is using '${LIBR_EXECUTABLE}'.\
+        \nEdit src/cmake/modulesFindLibR.cmake and add your R path to HINTS near this error")
+    endif()
+  endif()
+
+# Windows
+else()
+
+  # if R executable not available, query R_HOME path from registry
```
StrikerRUS: I don't see a check for `LIBR_EXECUTABLE`. Maybe we should add it here as it is stated in the comment?

R-package/src/cmake/modules/FindLibR.cmake
```diff
@@ -0,0 +1,199 @@
...
+        LIBR_HOME
+        "[HKEY_CURRENT_USER\\SOFTWARE\\R-core\\R\\${CMAKE_R_VERSION};InstallPath]"
+        ABSOLUTE
+      )
+    endif()
+
+    if(NOT LIBR_HOME)
+      message(FATAL_ERROR "\nUnable to locate R executable.\
+        \nEither add its location to PATH or provide it through the LIBR_EXECUTABLE CMake variable")
+    endif()
```
StrikerRUS: I think these `if`s can be outside of the first `if(NOT LIBR_HOME)`.

R-package/src/cmake/modules/FindLibR.cmake
```diff
@@ -0,0 +1,199 @@
...
+    if(NOT LIBR_HOME)
+      get_filename_component(
+        LIBR_HOME
+        "[HKEY_CURRENT_USER\\SOFTWARE\\R-core\\R\\${CMAKE_R_VERSION};InstallPath]"
+        ABSOLUTE
+      )
+    endif()
+
+    if(NOT LIBR_HOME)
+      message(FATAL_ERROR "\nUnable to locate R executable.\
```
StrikerRUS:
```suggestion
      message(FATAL_ERROR "Unable to locate R executable.\
```

R-package/src/cmake/modules/FindLibR.cmake
```diff
@@ -0,0 +1,199 @@
...
+  find_library(LIBR_LIBRARIES R)
+
+  if(LIBR_LIBRARIES MATCHES ".*\\.framework")
+    set(LIBR_HOME "${LIBR_LIBRARIES}/Resources")
+    set(LIBR_EXECUTABLE "${LIBR_HOME}/R")
+  else()
+    get_filename_component(_LIBR_LIBRARIES "${LIBR_LIBRARIES}" REALPATH)
+    get_filename_component(_LIBR_LIBRARIES_DIR "${_LIBR_LIBRARIES}" DIRECTORY)
+    set(LIBR_EXECUTABLE "${_LIBR_LIBRARIES_DIR}/../bin/R")
+  endif()
```
StrikerRUS: Should we wrap this block into `if(NOT LIBR_EXECUTABLE)` as for `UNIX` to allow passing `LIBR_EXECUTABLE` directly?

R-package/src/cmake/modules/FindLibR.cmake
```diff
@@ -0,0 +1,199 @@
...
+
+  endif()
+
+endif()
+
+if(NOT LIBR_EXECUTABLE)
+  message(FATAL_ERROR "Unable to locate R executable.\
+    \nEither add its location to PATH or provide it through the LIBR_EXECUTABLE CMake variable"
+  )
+endif()
```
StrikerRUS:
```suggestion
    \nEither add its location to PATH or provide it through the LIBR_EXECUTABLE CMake variable")
endif()
```

R-package/src/cmake/modules/FindLibR.cmake
```diff
@@ -0,0 +1,199 @@
...
+  OUTPUT_VARIABLE LIBR_HOME
+)
+
+# ask R for the include dir
+execute_process(
+  COMMAND ${LIBR_EXECUTABLE} "--slave" "--no-save" "-e" "cat(normalizePath(R.home('include'), winslash='/'))"
+  OUTPUT_VARIABLE LIBR_INCLUDE_DIRS
+)
+
+# C:/PROGRA~1/R/R-36~1.1/include
```
StrikerRUS: ??

jameslamb:
> @jameslamb Finally got a minute to drop a second round of review. Sorry for the delay!

no problem! I'll look at these right now.

R-package/src/cmake/modules/FindLibR.cmake
jameslamb: good idea!

R-package/src/cmake/modules/FindLibR.cmake
```diff
@@ -0,0 +1,199 @@
...
+
+    # CRAN may run RD CMD CHECK instead of R CMD CHECK,
+    # which can lead to this infamous error:
+    # 'R' should not be used without a path -- see par. 1.6 of the manual
+    find_program(
+      LIBR_EXECUTABLE
+      NO_DEFAULT_PATH
+      HINTS "${CMAKE_CURRENT_BINARY_DIR}" "/usr/bin" "/usr/lib/" "/usr/local/bin/"
+      NAMES R R.exe
+    )
```
StrikerRUS: Hmm, I remember I was typing this comment but it was not included in ,the review... OK, retyping it again.

I think we can do it before any actions for finding R with the help of  `LIBR_HOME` or `LIBR_LIBRARIES`. Also, I don't think that `NAME` `R.exe` will be useful for Unix 😃

jameslamb: 😬 ha R.exe, bad copy-pasting

Can you clarify what you mean by "before any actions for finding R with the help of LIBR_HOME or LIBR_LIBRARIES"? Do you mean this should be run before `if(APPLE)` (so outside of all the OS-specific checks)?

R-package/src/cmake/modules/FindLibR.cmake
jameslamb: we can actually wrap all of this in `if(NOT LIBR_EXECUTABLE)` for that purpose, now that I think about it. Since the point of these OS-specific blocks is to just find R and then we use R to find everything else.

CMakeLists.txt
```diff
@@ -72,6 +72,14 @@ if(USE_R35)
 endif(USE_R35)
 
 if(BUILD_FOR_R)
+    list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake/modules")
+    find_package(LibR REQUIRED)
+    message(STATUS "LIBR_HOME [${LIBR_HOME}]")
```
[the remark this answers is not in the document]
jameslamb: no it should always be a single string. I can remove them

R-package/src/cmake/modules/FindLibR.cmake
jameslamb: ha you're right

R-package/src/cmake/modules/FindLibR.cmake
jameslamb: yes I think so, and I think it should be this: https://github.com/microsoft/LightGBM/pull/2901#discussion_r396154303

CMakeLists.txt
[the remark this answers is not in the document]
jameslamb: removed in https://github.com/microsoft/LightGBM/pull/2901/commits/2403a6133c283df64db71e8fadb09a130c53f16f

R-package/src/cmake/modules/FindLibR.cmake
jameslamb: removed in https://github.com/microsoft/LightGBM/pull/2901/commits/2403a6133c283df64db71e8fadb09a130c53f16f

R-package/src/cmake/modules/FindLibR.cmake
jameslamb: I implemented https://github.com/microsoft/LightGBM/pull/2901#discussion_r396154303 in https://github.com/microsoft/LightGBM/pull/2901/commits/2403a6133c283df64db71e8fadb09a130c53f16f

R-package/src/cmake/modules/FindLibR.cmake
jameslamb: implemented in https://github.com/microsoft/LightGBM/pull/2901/commits/2403a6133c283df64db71e8fadb09a130c53f16f

R-package/src/cmake/modules/FindLibR.cmake
jameslamb: implemented in https://github.com/microsoft/LightGBM/pull/2901/commits/2403a6133c283df64db71e8fadb09a130c53f16f

R-package/src/cmake/modules/FindLibR.cmake
jameslamb: ha! removed in https://github.com/microsoft/LightGBM/pull/2901/commits/2403a6133c283df64db71e8fadb09a130c53f16f

R-package/src/cmake/modules/FindLibR.cmake
StrikerRUS: > Do you mean this should be run before if(APPLE) (so outside of all the OS-specific checks)?

Yeah, exactly! I think as a first thing we can simply try to find R via `find_command()` for all OSes before any OS-specific hints. Like
```
if(NOT LIBR_EXECUTABLE)
  find_program( 
        LIBR_EXECUTABLE
        NAMES R R.exe
      )
endif()

if(NOT LIBR_EXECUTABLE)
  if(APPLE)
    ...
endif()
```

jameslamb: @StrikerRUS so the reason I _don't_ have that currently is related to  https://github.com/microsoft/LightGBM/pull/2901#discussion_r391733668

When I just had

```
if(NOT LIBR_EXECUTABLE)
  find_program( 
        LIBR_EXECUTABLE
        NAMES R R.exe
      )
endif()
```

During `R  CMD check` it was finding the weird not-exactly-R R executable that is used by CRAN during checks, which results in the error

> 'R' should not be used without a path -- see par. 1.6 of the manual

I didn't see any option [in find_program()](https://cmake.org/cmake/help/v3.12/command/find_program.html) to exclude certain paths (like those that contain `.Rcheck`) so this thing with hints was my workaround until we can come up with something better. 

I guess I could probably do this and then add code that says "if you found the .Rcheck/ one, unset `LIBR_EXECUTABLE` and continue", but I don't see a clean way to say "use  `find_program()` and ignore the R executable that is used by `R CMD check`".

StrikerRUS: Can `NO_PACKAGE_ROOT_PATH` help?

Also, I think that it can be done in the following way. Or am I missing something?..
```
if(NOT LIBR_EXECUTABLE)
  find_program( 
        LIBR_EXECUTABLE
        NAMES R R.exe
      )
  # workaround for R CMD check
  if(LIBR_EXECUTABLE MATCHES ".*lightgbm\\.Rcheck.*")
    unset(LIBR_EXECUTABLE)
  endif()
endif()

if(NOT LIBR_EXECUTABLE)
  if(APPLE)
    ...
endif()
```
https://cmake.org/cmake/help/v3.0/command/unset.html

jameslamb: Yep! this is exactly what I meant (I think I was not clear) by:

> I guess I could probably do this and then add code that says "if you found the .Rcheck/ one, unset LIBR_EXECUTABLE and continue", but I don't see a clean way to say "use find_program() and ignore the R executable that is used by R CMD check".

So basically we won't get "the second thing `find_program()` would have found" if the first one is the .Rcheck one. We'll have to fail back to the OS-specific expectations about directory structure. Honestly that's not too bad. I think I'll do that.

as for `NO_PACKAGE_ROOT`...I don't think it will matter for us since we won't have specified `R_ROOT`.

jameslamb: wait now I think I understand `NO_PACKAGE_ROOT`!  Ok let me try...

jameslamb: Ok I ended up not needing  `NO_PACKAGE_ROOT`. @StrikerRUS whenever you have a chance (thanks by the way for the thorough review!) take a look at my attempt here:

https://github.com/microsoft/LightGBM/pull/2901/commits/5b60e8e7ced86e68cab9b374431ebe846c19d48d

Moving that block up, I had to account for one more special case besides R  CMD CHECK. On Mac, the built-in R GUI is bundled as [R.app](https://cran.r-project.org/bin/macosx/RMacOSX-FAQ.html#R_002eapp) and the executable is (regrettably), also called R. Something like `/Applications/R.app/Contents/MacOS/R`. So that needs to be removed too.

I don't _think_ we'll have a similar issue on Windows ([where you have `Rterm.exe` and `Rgui.exe`](https://cran.r-project.org/bin/windows/base/rw-FAQ.html#How-do-I-run-it_003f)).

StrikerRUS: Thank you for the investigation!

> Ok I ended up not needing `NO_PACKAGE_ROOT`

It's cool! Because `NO_PACKAGE_ROOT_PATH` was introduced only in CMake 3.12.

jameslamb: oh nice, ok! I also decided to add https://github.com/microsoft/LightGBM/pull/2901/commits/c9dd50d8bbb4d26b672ccdde7f57b5ea9afa8466, which ensures that all the uses of  `LIBR_EXECUTABLE` ignore environment variables and any `.Rprofile` files.

Just proposed the same to `xgboost`: https://github.com/dmlc/xgboost/pull/5438

jameslamb:
R + Mac CI on Travis died with this error:

image

Looks like a timeout for Homebrew. I have been seeing that slightly more the last few weeks. I imagine all the people now working remote is putting a lot of pressure on ISPs.

StrikerRUS:
@jameslamb

> Looks like a timeout for Homebrew. I have been seeing that slightly more the last few weeks. I imagine all the people now working remote is putting a lot of pressure on ISPs.

Yeah, it seems to be so...

I cannot remember how many times I re-run Travis macOS builds last days. Also, Travis had some problems with macOS recently: https://www.traviscistatus.com/

R-package/src/cmake/modules/FindLibR.cmake
```diff
@@ -0,0 +1,210 @@
...
+    LIBR_EXECUTABLE
+    NAMES R R.exe
+  )
+
+  # CRAN may run RD CMD CHECK instead of R CMD CHECK,
+  # which can lead to this infamous error:
+  # 'R' should not be used without a path -- see par. 1.6 of the manual
+  if(LIBR_EXECUTABLE MATCHES ".*\\.Rcheck.*")
+    unset(LIBR_EXECUTABLE PARENT_SCOPE)
+    unset(LIBR_EXECUTABLE CACHE)
```
StrikerRUS: Why do we unset in `PARENT_SCOPE` and `CACHE` variable, but don't unset ordinary variable? I think we need to unset only variable that was set by `find_program(LIBR_EXECUTABLE NAMES R R.exe)` two lines above.

R-package/src/cmake/modules/FindLibR.cmake
```diff
@@ -0,0 +1,210 @@
...
+  endif()
+
+  # ignore the R bundled with R.app on Mac, since that is GUI-only
+  if(LIBR_EXECUTABLE MATCHES ".+R\\.app.*")
+    unset(LIBR_EXECUTABLE PARENT_SCOPE)
+    unset(LIBR_EXECUTABLE CACHE)
```
StrikerRUS: The same about unsetting ordinary variable.

R-package/src/cmake/modules/FindLibR.cmake
```diff
@@ -0,0 +1,210 @@
...
+  # CRAN may run RD CMD CHECK instead of R CMD CHECK,
+  # which can lead to this infamous error:
+  # 'R' should not be used without a path -- see par. 1.6 of the manual
+  if(LIBR_EXECUTABLE MATCHES ".*\\.Rcheck.*")
+    unset(LIBR_EXECUTABLE PARENT_SCOPE)
+    unset(LIBR_EXECUTABLE CACHE)
+  endif()
+
+  # ignore the R bundled with R.app on Mac, since that is GUI-only
+  if(LIBR_EXECUTABLE MATCHES ".+R\\.app.*")
```
StrikerRUS: Shouldn't it end with `$` `".+R\\.app$"` as we search for `R.app` executable?

R-package/src/cmake/modules/FindLibR.cmake
```diff
@@ -0,0 +1,210 @@
...
+    endif()
+
+  elseif(UNIX)
+
+    # attempt to find R executable
+    if(NOT LIBR_EXECUTABLE)
+      find_program(
+        LIBR_EXECUTABLE
+        NO_DEFAULT_PATH
+        HINTS "${CMAKE_CURRENT_BINARY_DIR}" "/usr/bin" "/usr/lib/" "/usr/local/bin/"
```
StrikerRUS: Can you please explain why `"${CMAKE_CURRENT_BINARY_DIR}"` is here?

CMakeLists.txt
```diff
@@ -72,6 +72,14 @@ if(USE_R35)
 endif(USE_R35)
 
 if(BUILD_FOR_R)
+    list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake/modules")
+    find_package(LibR REQUIRED)
+    message(STATUS "LIBR_HOME: ${LIBR_HOME}")
+    message(STATUS "LIBR_EXECUTABLE: ${LIBR_EXECUTABLE}")
+    message(STATUS "LIBR_INCLUDE_DIRS: ${LIBR_INCLUDE_DIRS}")
+    message(STATUS "LIBR_LIB_DIR: ${LIBR_LIB_DIR}")
+    message(STATUS "LIBR_CORE_LIBRARY: ${LIBR_CORE_LIBRARY}")
```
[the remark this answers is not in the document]
jameslamb: I think it's important to have `LIBR_EXECUTABLE` too. I have been using that a lot in debugging. Since we use that to find everything else, it's important to know which one was found.

I'd be ok with dropping  `LIBR_HOME` and `LIBR_LIB_DIR` though.

R-package/src/cmake/modules/FindLibR.cmake
jameslamb: When I originally had just `unset(LIBR_EXECUTABLE)`, it didn't work. I faced https://github.com/microsoft/LightGBM/pull/2901#discussion_r396594997 even after `unset(LIBR_EXECUTABLE)`. Once I changed to this it was working correctly. I found this behavior surprising, but it was what worked.

R-package/src/cmake/modules/FindLibR.cmake
jameslamb: No we're not searching for an executable called `R.app`. This is found because we're searching for an executable called `R` that ends up in a path like `/Applications/R.app/Contents/MacOS/R`. It's regrettable that they called the GUI-only executable `R` (the same as the headless one).

So this check is "if you found something in any path with `R.app` in it

R-package/src/cmake/modules/FindLibR.cmake
jameslamb: I guess this could matter if you have fully self-contained apps where you are installing R and your application's code into the same location (or even if you are bundling the R executable alongside your code). That seems pretty non-standard but I guess it is possible?

Honestly I'm just speculating. I copied the use of this straight from `xgboost`'s setup. I left it because I thought "well it's working for them, maybe I shouldn't touch it". I am definitely ok dropping it if you'd like.

CMakeLists.txt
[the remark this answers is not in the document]
StrikerRUS: Sure, agree with you!

R-package/src/cmake/modules/FindLibR.cmake
StrikerRUS: Hmm, okey... And both of them were needed to make it work, right?

R-package/src/cmake/modules/FindLibR.cmake
StrikerRUS: Ah, I see! Thanks for the clarification!

R-package/src/cmake/modules/FindLibR.cmake
jameslamb: I tried these one at a time and I thought both were needed, but I just tried again and I think only `unset(LIBR_EXECUTABLE CACHE)` is needed! I'll remove the `PARENT_SCOPE` one.

R-package/src/cmake/modules/FindLibR.cmake
StrikerRUS: OK, let's leave it as is. I was just curios about use-cases and  I thought you added this path intentionally.

jameslamb: I think my "self-contained apps" thing is uncommon but it's possible. I have worked with systems before where on the only directory you have write access to is the one that your code runs in and its subdirectories.

CMakeLists.txt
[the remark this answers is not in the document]
jameslamb: removed in https://github.com/microsoft/LightGBM/pull/2901/commits/20aeb82830f3c474fc4f2a55e4afedb06b5016b5

R-package/src/cmake/modules/FindLibR.cmake
jameslamb: removed in https://github.com/microsoft/LightGBM/pull/2901/commits/ab870d8895739a6535408f13f4ad1df946dc10a9 (forgot to include in the previous commit)

R-package/src/cmake/modules/FindLibR.cmake
jameslamb: removed in https://github.com/microsoft/LightGBM/pull/2901/commits/ab870d8895739a6535408f13f4ad1df946dc10a9 per https://github.com/microsoft/LightGBM/pull/2901#discussion_r396797894

R-package/src/cmake/modules/FindLibR.cmake
StrikerRUS: I think you can contribute to RStudio as well where XGBoost borrowed a lot from.

https://github.com/rstudio/rstudio/blob/master/cmake/modules/FindLibR.cmake

https://github.com/dmlc/xgboost/blob/1de36cdf1e9ff1c93ec648de91ecce288d5430a2/cmake/modules/FindLibR.cmake#L2

Issue #730, opened 2017-07-26
Add RandomState support for the sklearn wrapper


ekerazha:
Some libs that expect a sklearn-compatible estimator, pass a random_state parameter.

The random_state parameter is similar to the seed parameter of LightGBM, but it can also receive a RandomState instance.

    random_state : int, RandomState instance or None; default=None
        If int, random_state is the seed used by the random number generator;
        If RandomState instance, random_state is the random number generator;
        If None, the random number generator is the RandomState instance used
        by `np.random`.

At the moment, if I pass a random_state parameter with a RandomState instance, I get a `Unknown type of parameter` exception from **line 134** in **basic.py**

Would it be possible to add support for a sklearn-compatible random_state parameter for the sklearn wrapper?

guolinke:
@wxchan

wxchan:
@ekerazha all our current random numbers are generated in c++ code, so do you want to use RandomState to generate random number, or only the random seed matters (the seed will pass to c++ and still generate random number in c++)?

ekerazha:
@wxchan I think it should be ok to just take the seed, however it is not possible to get the original seed from a RandomState instance (see: https://stackoverflow.com/questions/32172054/how-can-i-retrieve-the-current-seed-of-numpys-random-number-generator ). Maybe we can still get a value (although it's not the original seed) calling get_state() (see: https://docs.scipy.org/doc/numpy/reference/generated/numpy.random.get_state.html ) and pass this value to the c++ code.

So the approach could be this:
1) If random_state is a int value, pass it as seed to the c++ code (just like for the seed parameter).
2) If random_state is a RandomState instance, call get_state() and derive from the returned value (how?) a seed for the c++ code. Ideally we would directly use RandomState to generate random numbers, but I don't know if this is feasible.

wxchan:
@ekerazha The simplest way is to generate a random number as the seed, but I think it won't be what you want. (because it makes no differences as passing an integer.) But actually for current codes, even if you pass an integer, it is still used to generate 4 different random number for different purposes (bagging, sub-feature etc).

ekerazha:
@wxchan I think I'd go for the two steps I listed above, so:

- If random_state is a int value, just pass that value as seed
- If random_state is a RandomState instance, call get_state() and derive a int value from that (crc32 on some value returned by get_state()?) to use as seed

What do you think about this?

wxchan:
[...]

PR #2904, opened 2020-03-13
[python] handle RandomState object in Scikit-learn Api

lpfann:
I am using LightGBM with its scikit-learn API in a wrapper model called [Boruta](https://github.com/scikit-learn-contrib/boruta_py).
Boruta is using multiple cloned LGBM models in sequence.
In every iteration, it is cloning and setting parameters, including the random_state used in random number generation.

LGBM is not able to handle a RandomState object from Numpys random module which is the standard in scikit-learn models.
Because of that, many wrappers just pass down the RandomState in their inner models.
This was already discussed in #730.
LGBM is using an integer to seed its number generators and expects an integer, which makes it unusable in this context.

In this PR I added a random_state property with a setter method to handle RandomState objects.
If an RS object is passed, It derives an integer based on its state.
The advantage of the property is, that it handles both use cases:
1. normal initialisation
2. cloning and setting parameters via set_params

This change is quite minimal in my opinion and still allows the passing of integers as before.

### DIFF
--- a/tests/python_package_test/test_sklearn.py
+++ b/tests/python_package_test/test_sklearn.py
@@ -230,17 +230,19 @@ class TestSklearn(unittest.TestCase):
     def test_random_state_object(self):
         X, y = load_iris(True)
         X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.1, random_state=42)
-        state = np.random.RandomState(123)
-        clf = lgb.LGBMClassifier(n_estimators=10, subsample=0.5, subsample_freq=1, random_state=state)
-        self.assertIs(clf.random_state, state)
-        clf.fit(X_train, y_train)
-        y_pred1 = clf.predict(X_test, raw_score=True)
-        clf.fit(X_train, y_train)
-        y_pred2 = clf.predict(X_test, raw_score=True)
-        self.assertIs(clf.random_state, state)
-        self.assertRaises(AssertionError,
-                          np.testing.assert_allclose,
-                          y_pred1, y_pred2)
+        state1 = np.random.RandomState(123)
+        state2 = np.random.RandomState(123)
+        clf1 = lgb.LGBMClassifier(n_estimators=10, subsample=0.5, subsample_freq=1, random_state=state1)
+        clf2 = lgb.LGBMClassifier(n_estimators=10, subsample=0.5, subsample_freq=1, random_state=state2)
+        self.assertIs(clf1.random_state, state1)
+        clf1.fit(X_train, y_train)
+        clf2.fit(X_train, y_train)
+        y_pred1 = clf1.predict(X_test, raw_score=True)
+        y_pred2 = clf2.predict(X_test, raw_score=True)
+        self.assertEqual(clf1.best_iteration_, clf2.best_iteration_)
+        self.assertEqual(clf1.best_score_, clf2.best_score_)
+        np.testing.assert_array_equal(y_pred1, y_pred2)
+        np.testing.assert_array_equal(clf1.feature_importances_, clf2.feature_importances_)
 
     def test_feature_importances_single_leaf(self):
         data = load_iris()
--- a/tests/python_package_test/test_sklearn.py
+++ b/tests/python_package_test/test_sklearn.py
@@ -239,10 +239,12 @@ class TestSklearn(unittest.TestCase):
         clf2.fit(X_train, y_train)
         y_pred1 = clf1.predict(X_test, raw_score=True)
         y_pred2 = clf2.predict(X_test, raw_score=True)
-        self.assertEqual(clf1.best_iteration_, clf2.best_iteration_)
         self.assertEqual(clf1.best_score_, clf2.best_score_)
         np.testing.assert_array_equal(y_pred1, y_pred2)
         np.testing.assert_array_equal(clf1.feature_importances_, clf2.feature_importances_)
+        df1 = clf1.booster_.model_to_string(num_iteration=0)
+        df2 = clf2.booster_.model_to_string(num_iteration=0)
+        self.assertMultiLineEqual(df1,df2)
 
     def test_feature_importances_single_leaf(self):
         data = load_iris()
--- a/tests/python_package_test/test_sklearn.py
+++ b/tests/python_package_test/test_sklearn.py
@@ -244,7 +244,7 @@ class TestSklearn(unittest.TestCase):
         np.testing.assert_array_equal(clf1.feature_importances_, clf2.feature_importances_)
         df1 = clf1.booster_.model_to_string(num_iteration=0)
         df2 = clf2.booster_.model_to_string(num_iteration=0)
-        self.assertMultiLineEqual(df1,df2)
+        self.assertMultiLineEqual(df1, df2)
 
     def test_feature_importances_single_leaf(self):
         data = load_iris()
--- a/tests/python_package_test/test_sklearn.py
+++ b/tests/python_package_test/test_sklearn.py
@@ -228,11 +228,19 @@ class TestSklearn(unittest.TestCase):
         np.testing.assert_allclose(pred_origin, pred_pickle)
 
     def test_random_state_object(self):
-        data = load_iris()
+        X, y = load_iris(True)
+        X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.1, random_state=42)
         state = np.random.RandomState(123)
-        clf = lgb.LGBMClassifier(random_state=state)
-        self.assertIsInstance(clf.random_state, np.random.RandomState)
-        clf.fit(data.data, data.target)
+        clf = lgb.LGBMClassifier(n_estimators=10, subsample=0.5, subsample_freq=1, random_state=state)
+        self.assertIs(clf.random_state, state)
+        clf.fit(X_train, y_train)
+        y_pred1 = clf.predict(X_test, raw_score=True)
+        clf.fit(X_train, y_train)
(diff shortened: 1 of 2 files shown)


lpfann:
The tests should all pass. 
The only error is caused by an unrelated URL check which is apparently rate limited by Google (youtube video in docs)?!

StrikerRUS:
@lpfann Thanks for your PR! I think this is useful feature.

However, I believe that the following recently added to XGBoost implementation is a little bit clearer: https://github.com/dmlc/xgboost/pull/5049.

It requires less code, doesn't create additional properties and the most important follows scikit-learn dev guide. You can replace `RandomState` object with int value somewhere here:
https://github.com/microsoft/LightGBM/blob/0d878b17fe696e6656606f2072053cc7327300c5/python-package/lightgbm/sklearn.py#L498-L518

> The only error is caused by an unrelated URL check which is apparently rate limited by Google (youtube video in docs)?!

Please never mind! This test is flaky.

lpfann:
> 
> 
> @lpfann Thanks for your PR! I think this is useful feature.
> 
Thanks for your feedback!

> However, I believe that the following recently added to XGBoost implementation is a little bit clearer: [dmlc/xgboost#5049](https://github.com/dmlc/xgboost/pull/5049).
> 
> It requires less code, doesn't create additional properties and the most important follows scikit-learn dev guide. You can replace `RandomState` object with int value somewhere here:
> 
> https://github.com/microsoft/LightGBM/blob/0d878b17fe696e6656606f2072053cc7327300c5/python-package/lightgbm/sklearn.py#L498-L518
> 
I adapted it now according to your hint.
Its simpler and also preserves the original RandomState object.
It only derives the integer when actually fitting and passing the parameters down to the low level library.

tests/python_package_test/test_sklearn.py
```diff
@@ -227,6 +227,13 @@ def test_joblib(self):
         pred_pickle = gbm_pickle.predict(X_test)
         np.testing.assert_allclose(pred_origin, pred_pickle)
 
+    def test_random_state_object(self):
+        data = load_iris()
+        state = np.random.RandomState(123)
+        clf = lgb.LGBMClassifier(random_state=state)
+        self.assertIsInstance(clf.random_state, np.random.RandomState)
+        clf.fit(data.data, data.target)
```
StrikerRUS:
```suggestion
        X, y = load_iris(True)
        X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.1, random_state=42)
        state = np.random.RandomState(123)
        clf = lgb.LGBMClassifier(n_estimators=10, subsample=0.5, subsample_freq=1, random_state=state)
        self.assertIs(clf.random_state, state)
        clf.fit(X_train, y_train)
        y_pred1 = clf.predict(X_test, raw_score=True)
        clf.fit(X_train, y_train)
        y_pred2 = clf.predict(X_test, raw_score=True)
        self.assertIs(clf.random_state, state)
        self.assertRaises(AssertionError,
                          np.testing.assert_allclose,
                          y_pred1, y_pred2)
```

StrikerRUS:
I'd like to get one more review from @henry0312, @guolinke or @jameslamb before merging

tests/python_package_test/test_sklearn.py
```diff
@@ -227,6 +227,21 @@ def test_joblib(self):
...
+        clf = lgb.LGBMClassifier(n_estimators=10, subsample=0.5, subsample_freq=1, random_state=state)
+        self.assertIs(clf.random_state, state)
+        clf.fit(X_train, y_train)
+        y_pred1 = clf.predict(X_test, raw_score=True)
+        clf.fit(X_train, y_train)
+        y_pred2 = clf.predict(X_test, raw_score=True)
+        self.assertIs(clf.random_state, state)
+        self.assertRaises(AssertionError,
+                          np.testing.assert_allclose,
+                          y_pred1, y_pred2)
```
jameslamb: two models that cleanly separate the test cases in `iris` could still be different, so I don't think this is a sufficient test that we're controlling randomness.

could you add a test that more thoroughly tests that we're getting the same exact model, like checking the feature importances? (https://github.com/microsoft/LightGBM/pull/2904/files#diff-1613c46333fcf60f940b13627ac833c7R218)

lpfann: I added new checks for equality of feature importance and booster state.
For the booster state, I now compare the output of the model_to_string() method.
Does that account for everything?

StrikerRUS: @jameslamb Generally, I agree with you! But after merging #2804, two models with identical parameters should produce exactly the same results on the same machine even with multithreaded bagging.
> two models that cleanly separate the test cases in `iris` could still be different,

To prevent ideal class predictions, I set
-  small boosting rounds (to save CI time as well 😄 ),
-  `raw_score=True`, so we are not predicting classes.

StrikerRUS:
I think we should combine two test cases: first that that same `RandomState` results in the same models, second that re-fitting model results in different predictions as new seed is sampled at each `fit` stage.

tests/python_package_test/test_sklearn.py
jameslamb: ohhhhhh I didn't notice `raw_score=True`. I thought scikit-learn interfaces for classification were always

* `predict()` = class labels
* `predict_proba()` = predict probability for each possible label

Then ok! I like the changes you pushed @lpfann. I don't have any requested  changes to the implementation, it looks solid esp. since it is backwards-compatible.

So from my perspective, this PR is good to go whenever you resolve https://github.com/microsoft/LightGBM/pull/2904#issuecomment-600795398 with @StrikerRUS 

Thanks again for the contribution!

lpfann:
> 
> 
> I think we should combine two test cases: first that the same `RandomState` results in the same models, second that re-fitting model results in different predictions as new seed is sampled at each `fit` stage.

I now added such a test in https://github.com/microsoft/LightGBM/pull/2904/commits/26eb6e6ef628983df1490d902243d8def39b9f38

tests/python_package_test/test_sklearn.py
```diff
@@ -227,6 +227,37 @@ def test_joblib(self):
...
 
+    def test_random_state_object(self):
+        X, y = load_iris(True)
+        X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.1, random_state=42)
+        state1 = np.random.RandomState(123)
+        state2 = np.random.RandomState(123)
+        clf1 = lgb.LGBMClassifier(n_estimators=10, subsample=0.5, subsample_freq=1, random_state=state1)
+        clf2 = lgb.LGBMClassifier(n_estimators=10, subsample=0.5, subsample_freq=1, random_state=state2)
+        # Test if random_state is properly stored
+        self.assertIs(clf1.random_state, state1)
```
StrikerRUS:
```suggestion
        self.assertIs(clf1.random_state, state1)
        self.assertIs(clf2.random_state, state2)
```

tests/python_package_test/test_sklearn.py
```diff
@@ -227,6 +227,37 @@ def test_joblib(self):
...
+        # Test if two random states produce identical models
+        clf1.fit(X_train, y_train)
+        clf2.fit(X_train, y_train)
+        y_pred1 = clf1.predict(X_test, raw_score=True)
+        y_pred2 = clf2.predict(X_test, raw_score=True)
+        self.assertEqual(clf1.best_score_, clf2.best_score_)
+        np.testing.assert_array_equal(y_pred1, y_pred2)
```
StrikerRUS: `best_score_` is empty here.
```suggestion
        np.testing.assert_allclose(y_pred1, y_pred2)
```

tests/python_package_test/test_sklearn.py
```diff
@@ -227,6 +227,37 @@ def test_joblib(self):
...
+        np.testing.assert_array_equal(clf1.feature_importances_, clf2.feature_importances_)
+        df1 = clf1.booster_.model_to_string(num_iteration=0)
+        df2 = clf2.booster_.model_to_string(num_iteration=0)
+        self.assertMultiLineEqual(df1, df2)
+        # Test if subsequent fits sample from random_state object and produce different models
+        clf1.fit(X_train, y_train)
+        y_pred1_refit = clf1.predict(X_test, raw_score=True)
+        df3 = clf1.booster_.model_to_string(num_iteration=0)
+        self.assertRaises(AssertionError,
+                          np.testing.assert_array_equal,
```
StrikerRUS:
```suggestion
                          np.testing.assert_allclose,
```

tests/python_package_test/test_sklearn.py
```diff
@@ -227,6 +227,37 @@ def test_joblib(self):
...
+        self.assertEqual(clf1.best_score_, clf2.best_score_)
+        np.testing.assert_array_equal(y_pred1, y_pred2)
+        np.testing.assert_array_equal(clf1.feature_importances_, clf2.feature_importances_)
+        df1 = clf1.booster_.model_to_string(num_iteration=0)
+        df2 = clf2.booster_.model_to_string(num_iteration=0)
+        self.assertMultiLineEqual(df1, df2)
+        # Test if subsequent fits sample from random_state object and produce different models
+        clf1.fit(X_train, y_train)
+        y_pred1_refit = clf1.predict(X_test, raw_score=True)
+        df3 = clf1.booster_.model_to_string(num_iteration=0)
```
StrikerRUS:
```suggestion
        df3 = clf1.booster_.model_to_string(num_iteration=0)
        self.assertIs(clf1.random_state, state1)
        self.assertIs(clf2.random_state, state2)
```

PR #2907, opened 2020-03-14
[python] fix the bug when use different params with reference

guolinke:


python-package/lightgbm/basic.py
```diff
@@ -1097,6 +1097,9 @@ def construct(self):
         """
         if self.handle is None:
             if self.reference is not None:
+                if self.get_params() != self.reference.get_params():
```
StrikerRUS: Maybe store reference params in variable to not call `get_params()` twice?
```
reference_params = self.reference.get_params()
```

python-package/lightgbm/basic.py
```diff
@@ -1097,6 +1097,9 @@ def construct(self):
...
+                    warnings.warn('Override the parameters from Reference Dataset.')
```
```suggestion
                    warnings.warn('Overriding the parameters from Reference Dataset.')
```

python-package/lightgbm/basic.py
```diff
@@ -1222,6 +1225,8 @@ def save_binary(self, filename):
         return self
 
     def _update_params(self, params):
+        if not params:
+            return
```
```suggestion
            return self
```

python-package/lightgbm/basic.py
```diff
@@ -1097,6 +1097,9 @@ def construct(self):
         """
         if self.handle is None:
             if self.reference is not None:
+                if self.get_params() != self.reference.get_params():
+                    warnings.warn('Override the parameters from Reference Dataset.')
+                    self._update_params(self.reference.get_params())
```
StrikerRUS: Do we need a test for these changes?

tests/python_package_test/test_engine.py
```diff
@@ -1958,6 +1958,16 @@ def test_dataset_update_params(self):
...
 
+    def test_dataset_params_with_refence(self):
+        default_params = {"max_bin": 100}
+        X = np.random.random((100, 2))
+        y = np.random.random(100)
+        X_val = np.random.random((100, 2))
+        y_val = np.random.random(100)
+        lgb_train = lgb.Dataset(X, y, params=default_params, free_raw_data=False).construct()
+        lgb_val = lgb.Dataset(X_val, y_val, reference=lgb_train, free_raw_data=False).construct()
+        model = lgb.train(default_params, lgb_train, valid_sets = [lgb_val])
```
guolinke:
```suggestion
        model = lgb.train(default_params, lgb_train, valid_sets=[lgb_val])
```

PR #2908, opened 2020-03-14
[R-package] fixed some minor issues with testing on Windows

jameslamb:
I started working on #2335 tonight, specifically adding CI for Windows.

In this PR, I'd like to propose a few small fixes to start improving support on Windows.

1. Added architecture-specific source directories to `.gitignore`. These come from multi-architecture R package builds on Windows, and shouldn't be checked into source control
2. Minor readability cleanup in Windows section of `install.libs.R`, added some logs and clarified existing logs
3. Fixed one unit test that fails on Windows because of different default quoting between Windows and other operating system

I tested these changes locally on Windows 10, using Visual Studio 16 2019, with `Rscript build_r.R`

### DIFF
--- a/R-package/src/install.libs.R
+++ b/R-package/src/install.libs.R
@@ -4,7 +4,7 @@ use_gpu <- FALSE
 use_mingw <- FALSE
 
 if (.Machine$sizeof.pointer != 8L) {
-  stop("Only support 64-bit R, please check your the version of your R and Rtools.")
+  stop("LightGBM only supports 64-bit R, please check the version of R and Rtools.")
 }
 
 R_int_UUID <- .Internal(internalsID())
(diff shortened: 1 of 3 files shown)


R-package/src/install.libs.R
```diff
@@ -52,22 +57,28 @@ if (!use_precompile) {
       build_cmd <- "mingw32-make.exe _lightgbm"
       system(paste0(cmake_cmd, " ..")) # Must build twice for Windows due sh.exe in Rtools
     } else {
-      try_vs <- 0L
```
StrikerRUS: Why this initialization was removed?

jameslamb: This isn't necessary, I think maybe it is a misunderstanding about how `system()` works. You don't need to allocate an int and then fill it. `system()` will just return it.

StrikerRUS: OK, so in R variable defined in nested scope remains visible in outer scope later, right?

jameslamb: That's right. Python is the same way

```python
for i in range(5):
    x = i
```

If you think that isn't obvious though I can restore this! But I'd change it to `1L` so that that default assumption is "VS didn't work" instead of "assume it worked"

StrikerRUS: I see, thanks for the explanation!

I'd better leave it up to other reviewers.

jameslamb: no prob! @Laurae2 or @guolinke what do you think?

jameslamb: gently ping @guolinke

guolinke: I think it could be remvoed.

jameslamb: great, thank you

PR #2909, opened 2020-03-14
[R-package] added command to remove build/ dir in install.libs.R

jameslamb:
Another piece of adding Windows CI for the R package (#2335) and, as a result, #629

This PR fixes this Windows-specific NOTE that shows up in `R CMD check --as-cran`:

```
* checking line endings in C/C++/Fortran sources/headers ... NOTE
Found the following sources/headers with CR or CRLF line endings:
    src/build/CMakeFiles/3.16.5/CompilerIdC/CMakeCCompilerId.c
src/build/CMakeFiles/3.16.5/CompilerIdCXX/CMakeCXXCompilerId.cpp
src/build/CMakeFiles/FindOpenMP/OpenMPCheckVersion.c
src/build/CMakeFiles/FindOpenMP/OpenMPCheckVersion.cpp
src/build/CMakeFiles/FindOpenMP/OpenMPTryFlag.c
src/build/CMakeFiles/FindOpenMP/OpenMPTryFlag.cpp
Some Unix compilers require LF line endings.
* checking line endings in Makefiles ... OK
```

R does have a method for cleaning up intermediate files, with a `cleanup` script ([reference](https://colinfay.me/writing-r-extensions/creating-r-packages.html#configure-and-cleanup)). I tried that first, but it didn't work. I think those scripts are used in `R CMD build` and not run immediately after installation, which is why adding them with content like `rm -r src/build` did not work.

### DIFF
--- a/R-package/src/install.libs.R
+++ b/R-package/src/install.libs.R
@@ -142,8 +142,18 @@ if (!use_precompile) {
 dest <- file.path(R_PACKAGE_DIR, paste0("libs", R_ARCH), fsep = "/")
 dir.create(dest, recursive = TRUE, showWarnings = FALSE)
 if (file.exists(src)) {
-  cat("Found library file: ", src, " to move to ", dest, sep = "")
+  print(paste0("Found library file: ", src, " to move to ", dest))
   file.copy(src, dest, overwrite = TRUE)
 } else {
   stop(paste0("Cannot find lib_lightgbm", SHLIB_EXT))
+}
+
+# clean up the "build" directory
+if (dir.exists(build_dir)){
+  print("Removing 'build/' directory")
+  unlink(
+    x = file.path(R_PACKAGE_SOURCE, "src", "build")
+    , recursive = TRUE
+    , force = TRUE
+  )
 }
--- a/R-package/src/install.libs.R
+++ b/R-package/src/install.libs.R
@@ -149,7 +149,7 @@ if (file.exists(src)) {
 }
 
 # clean up the "build" directory
-if (dir.exists(build_dir)){
+if (dir.exists(build_dir)) {
   print("Removing 'build/' directory")
   unlink(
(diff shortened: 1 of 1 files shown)


R-package/src/install.libs.R
```diff
@@ -142,8 +142,18 @@ if (!use_precompile) {
...
 if (file.exists(src)) {
-  cat("Found library file: ", src, " to move to ", dest, sep = "")
+  print(paste0("Found library file: ", src, " to move to ", dest))
   file.copy(src, dest, overwrite = TRUE)
 } else {
   stop(paste0("Cannot find lib_lightgbm", SHLIB_EXT))
 }
+
+# clean up the "build" directory
+if (dir.exists(build_dir)) {
```
StrikerRUS: `build_dir` variable is not defined when `use_precompile == TRUE`, isn't it?

jameslamb: 😬 yep you're right!  Let me fix that

jameslamb: Moved it up in https://github.com/microsoft/LightGBM/pull/2909/commits/56635a7cf10b53f543db3b96cd1831dd159ff6b8. This also exposed that that `setwd(source_dir)` was unnecessary.

thanks!

R-package/src/install.libs.R
```diff
@@ -142,8 +145,18 @@ if (!use_precompile) {
...
+  print("Removing 'build/' directory")
+  unlink(
+    x = file.path(R_PACKAGE_SOURCE, "src", "build")
```
StrikerRUS: I think it will be clearer: 

`x = file.path(R_PACKAGE_SOURCE, "src", "build")` -> `x = file.path(build_dir)`

(Suggestions are disabled after merging a PR.)

jameslamb: blegh sorry, yeah you're right. That would eliminate a source of duplication. Let me add that right now.

PR #2924, opened 2020-03-18
[ci] fix VS project files

StrikerRUS:
Refer to #2891.


StrikerRUS:
@jameslamb 
> Is there any way we could catch this in CI?

I don't know any 🙁 

> or is it not worth it

I guess, yes.

PR #2926, opened 2020-03-20
Apply black

daikikatsuragawa:
Apply code style: black

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,7 @@ LightGBM, Light Gradient Boosting Machine
 [![PyPI Version](https://img.shields.io/pypi/v/lightgbm.svg?logo=pypi&logoColor=white)](https://pypi.org/project/lightgbm)
 [![Join Gitter at https://gitter.im/Microsoft/LightGBM](https://badges.gitter.im/Microsoft/LightGBM.svg)](https://gitter.im/Microsoft/LightGBM?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
 [![Slack](https://lightgbm-slack-autojoin.herokuapp.com/badge.svg)](https://lightgbm-slack-autojoin.herokuapp.com)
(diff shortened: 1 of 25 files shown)


StrikerRUS:
@daikikatsuragawa Thanks for your PR!

We already have linting tool at our CI tests and our codebase successfully passes it: 
https://github.com/microsoft/LightGBM/blob/91185c3aa403bdad8711fb32de6be33a0ad8d04f/.ci/test.sh#L52-L54

And we are against too strict out-dated rule of 80-chars max line length.

I personally find the current codebase even more readable than it will be after this PR. Moreover, it definitely doesn't worth to produce so huge diff (`+5,532 −3,293`), break the traceability of changes and waste reviewers' time only for `'` -> `"` replacements.

However, I'm not closing this PR to hear at least one more opinion.

daikikatsuragawa:
So that's it.
Sorry, I'm convinced at all, so it's okay to close!

jameslamb:
> @daikikatsuragawa Thanks for your PR!
> 
> We already have linting tool at our CI tests and our codebase successfully passes it:
> https://github.com/microsoft/LightGBM/blob/91185c3aa403bdad8711fb32de6be33a0ad8d04f/.ci/test.sh#L52-L54
> 
> BTW, this PR doesn't pass it.
> 
> And we are against too strict out-dated rule of 80-chars max line length.
> 
> I personally find the current codebase even more readable than it will be after this PR. Moreover, it definitely doesn't worth to produce so huge diff (`+5,532 −3,293`), break the traceability of changes and waste reviewers' time only for `'` -> `"` replacements.
> 
> However, I'm not closing this PR to hear at least one more opinion.

I agree with @StrikerRUS . Thank you for taking the time to contribute and we hope that you will continue to contribute (see our [issues](https://github.com/microsoft/LightGBM/issues) and [feature requests](https://github.com/microsoft/LightGBM/issues/2302)), but we already have linting set up for the Python side of this project and we chose it carefully.

PR #2928, opened 2020-03-20
[R-package] fixed NOTE about unrecognized global variable

jameslamb:
This PR fixes a new `R CMD CHECK` note introduced in #2803 . See https://github.com/microsoft/LightGBM/pull/2803#issuecomment-601840250 for background.

The note is:

```
* checking R code for possible problems ... NOTE
multiple.tree.plot.interpretation: no visible binding for global
  variable ‘bar_color’
  bar_color
```

### DIFF
--- a/R-package/R/lightgbm.R
+++ b/R-package/R/lightgbm.R
@@ -182,6 +182,7 @@ globalVariables(c(
     , ".N"
     , ".SD"
     , "abs_contribution"
+    , "bar_color"
     , "Contribution"
     , "Cover"
     , "Feature"


jameslamb:
> @jameslamb Thanks!
> 
> I can say nothing about the approach itself, but can confirm that the issue is eliminated:
> 
> ```
> * checking R code for possible problems ... OK
> ```

thanks! The approach is the same we've taken for other references in `data.table`.

I use this approach in every R package that has non-standard evaluation like those that use `data.table` or `ggplot2`. It's something that prominent R developers [recommend](https://stackoverflow.com/a/12429344/3986677).

It's funny, if you look in the documentation for `?globalVariables` in R (which ships with R by default), it literally says:

> For globalVariables, the names supplied are of functions or other objects that should be regarded as defined globally when the check tool is applied to this package

So it's basically like using  `// NOLINT` to suppress false positives from a linter.

Given all this, I think it's ok to merge without another review, assuming CI succeeds.

Issue #987, opened 2017-10-12
Standard outputs infos even when disabled [+ Usage question]


elDan101:
Sorry, I didn't realise I was on a pre-release version (2.0.6.). After updating to 2.0.7. the issue is resolved. I leave the text below anyway.

## Usage questions:
1) 
What does "Finished loading 120 models" mean? I get this as an output after incremental training. 
Are these single trees (=models) in the Boosting tree ensemble? It is increasing every time I incrementally train (not staying the same number of decrease).

2) 
When training incrementally I get many warnings `[WARNING] No further splits with positive gain, best gain: -inf` Is there a "best practice" how to tackle this, to avoid these warnings?

Thanks a lot, I will leave the issue open for the usage question, but close it then.  

## Obsolete issue:
Hi,

my target function is a time series. I split my dataset in multiple training/testing phases.

At each training phase I want to continue training from previous models (incremental learning).

Inside a training phase I do cross validation using [TimeSeriesSplit](http://scikit-learn.org/stable/modules/generated/sklearn.model_selection.TimeSeriesSplit.html) for cross validation, where I test multiple parameter settings.

At the moment I am having troubles with the console outputs. Here is a sample:

Phase 1 refers to the first training phase (there is no existing model yet).
```

******************************************************
INFO: fitting LightGBM in phase 1

--------------------------------------------

INFO: Start parameter tuning

INFO: Handling level0
INFO: **improved cv_error** with params={'verbose': 0, 'learning_rate': 0.01, 'num_leaves': 10, 'max_bin': 255} || current best result: 8207.489200311997
INFO: **improved cv_error** with params={'verbose': 0, 'learning_rate': 0.01, 'num_leaves': 20, 'max_bin': 255} || current best result: 8090.180625524847
INFO: tested for params={'verbose': 0, 'learning_rate': 0.01, 'num_leaves': 30, 'max_bin': 255} || result: 8252.27902379209
INFO: tested for params={'verbose': 0, 'learning_rate': 0.02, 'num_leaves': 10, 'max_bin': 255} || result: 8195.467582308242
INFO: tested for params={'verbose': 0, 'learning_rate': 0.02, 'num_leaves': 20, 'max_bin': 255} || result: 8136.15878262733
[...]
```

But when I enter phase 2, I get the following:

```
******************************************************
INFO: fitting LightGBM in phase 2
--------------------------------------------
INFO: Start parameter tuning
INFO: Handling level0
[...]

PR #2936, opened 2020-03-23
[R-package] [ci] Add Windows CI for R package (fixes #2335)

### LINKED ISSUE #2335: R unit tests are not run in CI

jameslamb:
This PR adds CI support for the R package on Windows. It also bumps up the version of R we test against (on all operating systems) from 3.6.1 to 3.6.3.

As of this PR, 4 R CMD check NOTES remain on Windows. All of them can be addressed in later PRs and none of them were introduced by changes in this PR.

<details><summary>R CMD CHECK NOTES (click for details)</summary>

**note-1**

```
* checking package dependencies ... NOTE
Packages suggested but not available for checking:
  'ggplot2', 'knitr', 'rmarkdown'
```

**note-2**

```
  installed size is  6.9Mb
    libs   6.3Mb
```

**note-3**

```
* checking compiled code ... NOTE
File 'lightgbm/libs/x64/lib_lightgbm.dll':
  Found no calls to: 'R_registerRoutines', 'R_useDynamicSymbols'

It is good practice to register native routines and to disable symbol
search.
```

**note-4**

```
* checking compiled code ... NOTE
Note: information on .o files for x64 is not available
File 'C:/projects/lightgbm/lightgbm.Rcheck/lightgbm/libs/x64/lib_lightgbm.dll':
  Found 'abort', possibly from 'abort' (C), 'runtime' (Fortran)
```

* `note-1`: not relevant for us, won't actually show up on CRAN and we shouldn't incur the extra time to install those packages. CRAN will have all Suggests dependencies available.
* `note-2`: only showing up on Windows. Probably some build files not caught by .Rbuildignore. Can be addressed in a follow-up PR.
* `note-3`: addressed by microsoft#2911
* `note-4`: Can be addressed in a follow-up PR.

</details>

Currently, `.appveyor.yml` on `master`  only has code for the Python package. It has some code there (like setting up the `test-env`  conda environment) that is identical to code in `.vsts-ci.yml`. In this PR, I decided to consolidate that code in `.ci/test_windows.ps1`. That made it easier to do things like `if $env:TASK -eq "r-package"`, and is similar to how we handle Mac and Linux in  `.ci/test.sh`.

The R portion of `.ci/test_windows.ps1` borrows heavily from @StrikerRUS 's  setup for the `RGF` package  (https://github.com/RGF-team/rgf/blob/master/R-package/.R.appveyor.ps1). Thank you for that!

This PR has so many commits in it because I spent a lot of time getting this working in the Travis, Appveyor, and Azure DevOps setups for my fork.

<details><summary>successful builds from jameslamb/LightGBM fork</summary>

[Azure DevOps](https://dev.azure.com/JayLamb20/jameslamb-lightgbm/_build/results?buildId=98&view=results)

image

[AppVeyor](https://ci.appveyor.com/project/jameslamb/lightgbm/builds/31641466)

image

[Travis](https://travis-ci.org/github/jameslamb/LightGBM/builds/665668080)

image

</details>

This PR supports the progress on #629 and would give us more confidence in things like #2901 and #987


jameslamb:
Azure is failing with this error, which I did see one other time during my testing:

image

This is a server for TeX packages. The one time I saw this while I was preparing this PR, it was fixed  by just waiting a minute or two and rebuilding, so I'm going to close and reopen to trigger CI.

Everything [on Travis](https://travis-ci.org/github/microsoft/LightGBM/builds/665681847?utm_source=github_status&utm_medium=notification) is passing except our old friend, the `check-docs` task.

jameslamb:
close-reopen to trigger CI

StrikerRUS:
@jameslamb Awesome job, thanks!

I see Azure is failing again with the error you've mentioned. As we know that it occurs quite often, maybe we should retry download? For example: https://forums.hak5.org/topic/36911-powershell-retry-download-until-complete/?tab=comments#comment-268464

.ci/test_windows.ps1
```diff
@@ -50,3 +74,78 @@ if ($env:TASK -eq "regular") {
   conda install -q -y -n $env:CONDA_ENV ipywidgets notebook
   jupyter nbconvert --ExecutePreprocessor.timeout=180 --to notebook --execute --inplace *.ipynb ; Check-Output $?  # run all notebooks
 }
+
+# test R package
+if ($env:TASK -eq "r-package"){
```
StrikerRUS: I think we should have a separate script file as we do for non-Windows:
https://github.com/microsoft/LightGBM/blob/3a335e9d4f63da10094d690a92cbefd0f4e6faf6/.ci/test.sh#L71-L74

jameslamb: oh ok sure! I can do that

jameslamb:
Ok I addressed and in https://github.com/microsoft/LightGBM/commit/e54954ad077389f019ea19add4829f23fb4aa7a0.

Once I got past those, I also had to add some changes for MSVC. I realized we don't really need `LIBR_LIB_DIR` in `FindLibR.cmake`...it was just an intermediary used to find `R.dll`. So I tried changing some things there to get MSVC  working.

On my personal Azure DevOps, the R build took 32 minutes to complete (but did complete successfully) and the Appveyor build timed out after an hour. I'm investigating more on my fork.

I think the issue here is that this PR was working for just "compile lib_lightgbm and build the R  package that calls it". But now that as of #2901 we're linking to R's shared library, there are additional things I need to fix.

jameslamb:
I'm currently working in my fork on getting the builds to work again now that we're linking to `R.dll`. I  am seeing this tough-to-understand behavior where, when using MSVC, the build just hangs for **an hour** and dies: https://ci.appveyor.com/project/jameslamb/lightgbm/builds/31695220/job/ia9pdstbxx0r3anj

I'm not seeing that behavior locally on my laptop running Windows 10, using MSVC bundled  with Visual Studio 16 2019. Will let you all know when this is ready for review, it was probably premature to turn it from Draft to Ready :/

jameslamb:
Ok here's where we stand:

* Appveyor build with `MinGW` seems to be working, and ran in 14 minute for the R package which seems ok.
* On AzureDevops, I'm still seeing an issue with downloading MiKTeX. The retry-logic is working but now it's hitting a different error where some step in `miktexsetup.exe` is trying to hit a server that's timing out. Did you ever face that in `RGF` @StrikerRUS ?

StrikerRUS:
@jameslamb 
> Did you ever face that in `RGF`?

Never. I remember only one issue related to `miktexsetup`:
> Update Miktex installation: https://miktex.org/announcement/miktex-portable-sfx-discontinued.
    https://github.com/RGF-team/rgf/pull/295

Maybe Azure blocks some IPs like Travis blocks apts outside of their whitelist?..
https://github.com/travis-ci/apt-package-safelist

jameslamb:
> @jameslamb
> 
> > Did you ever face that in `RGF`?
> 
> Never. I remember only one issue related to `miktexsetup`:
> 
> > Update Miktex installation: https://miktex.org/announcement/miktex-portable-sfx-discontinued.
> > [RGF-team/rgf#295](https://github.com/RGF-team/rgf/pull/295)
> 
> Maybe Azure blocks some IPs like Travis blocks apts outside of their whitelist?..
> https://github.com/travis-ci/apt-package-safelist

yeah I was thinking it might be something Azure-specific because I've only ever seen it there. That's a good idea!

I have an idea too, going to try it on my fork 😀

StrikerRUS:
@jameslamb 
If nothing helped, we can re-host installation package at our releases page like we do for AMD-APP-SDK
https://github.com/microsoft/LightGBM/blob/c870c57a14b61329bf493b3b1a9a2893a62405cf/.ci/setup.sh#L31

build_r.R
```diff
@@ -5,6 +5,9 @@
 # Sys.setenv("CXX" = "/usr/local/bin/g++-8")
```
jameslamb: this is not strictly related to this PR and is being added in #2957 . It'll disappear from the diff when that gets merged

jameslamb:
> @jameslamb
> If nothing helped, we can re-host installation package at our releases page like we do for AMD-APP-SDK
> https://github.com/microsoft/LightGBM/blob/c870c57a14b61329bf493b3b1a9a2893a62405cf/.ci/setup.sh#L31

It seems like MiKTeX installation is working ok now.

jameslamb:
@StrikerRUS I think it would make this PR easier to review if I make the Python changes a separate PR (moving things out of `.appveyor.yml` and  `.travis.yml` and into `.ci/test_windows.ps1`). And there is some independent benefit, reducing duplication in the definition of the `test-env` conda environment.

What do you think?

jameslamb:
As of https://github.com/microsoft/LightGBM/pull/2936/commits/b60826a5acaf175d8d1206e1d8fd057975af9b74, the R builds on [Azure DevOps](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=5689&view=logs&j=b51fa6ee-4364-59e6-4527-dc4aaafd2db2&t=85bde6ad-0953-589b-e1e8-633bb2d9d530) and [AppVeyor](https://ci.appveyor.com/project/jameslamb/lightgbm/builds/31839325/job/4bdc2g9hceu1ue35) are working!

I moved the `r-package` tasks to the beginning of `.appveyor.yml` and `.vsts-ci.yml`  because they take so long. Azure DevOps was taking 30+ minutes when I first opened this PR. I got it down to about 19 minutes by running a reduced version of `R CMD check`, but I know that's still kind of long. AppVeyor is taking about 13 minutes.

AppVeyor is being used to cover compilation with `MinGW`, Azure DevOps is covering `MSVC`.

At this point, with both of them working, I'd like to stop and wait for some review comments. @StrikerRUS @Laurae2 @guolinke looking forward to hearing what you think!

build_r.R
jameslamb: now that #2957 has been merged, I'll rebase to `master` before the next commit I push here

StrikerRUS:
@jameslamb 
> I think it would make this PR easier to review if I make the Python changes a separate PR 

Absolutely agree! Please do it.

jameslamb:
> @jameslamb
> 
> > I think it would make this PR easier to review if I make the Python changes a separate PR
> 
> Absolutely agree! Please do it.

thanks ! created #2962

jameslamb:
Just rebased to `master` to pick up changes from #2911 and #2957

jameslamb:
> Just rebased to `master` to pick up changes from #2911 and #2957

this seems to be working! Azure DevOps is failing just because of Azure capacity limitations

image

jameslamb:
Merged in the diff from #2962 and it seems that it's working and made the diff here smaller. I also just opened #2965 , which I think can cut the build time for the `r-package` task from 20 minutes to 9 minutes 😀

jameslamb:
You can see in this thread (https://github.com/microsoft/LightGBM/pull/2965#discussion_r402655226), there is another `gcc.exe` laying around on Azure DevOps and we could accidentally pick that up. That's a problem because the Azure DevOps build for R here is supposed to be testing that building the R package with Visual Studio works.

So in https://github.com/microsoft/LightGBM/pull/2936/commits/d100027e2c8a781f8b07bc433d72b2fb09b7d869 I added code to echo the logs from `R CMD check`

jameslamb:
Ok I got the Azure DevOps task down to about 15 minutes ([see this build from my fork](https://dev.azure.com/JayLamb20/jameslamb-lightgbm/_build/results?buildId=283&view=logs&j=b51fa6ee-4364-59e6-4527-dc4aaafd2db2&t=85bde6ad-0953-589b-e1e8-633bb2d9d530)). Since I'm proposing that we save time by not testing any of the R documentation on Windows on Azure DevOps (running `R CMD check --no-examples --no-manual --ignore-vignette`), we don't need to install MiKTeX or `pandoc`.

see this commit: https://github.com/microsoft/LightGBM/pull/2936/commits/912a2fd4a33cf97fde2a05472787f4c6ebdd50dc

I hope that will also make the CI a bit more reliable in case 
 the issue mentioned  in https://github.com/microsoft/LightGBM/pull/2936#issuecomment-604112267 comes back.

I think it's ok to skip all those documentation tests on Azure DevOps. We're already running all those tests on the docs on Mac with `gcc` and `clang`, Linux with `gcc` and  `clang`, and Windows with `MinGW` (on AppVeyor). The Azure DevOps builds for Windows will still test that the library is installable and that all unit tests run without error.

jameslamb:
I decided to just fix this NOTE directly in this PR. It doesn't have an obvious user-facing impact, so we don't have to break it into a separate PR.

```
* checking dependencies in R code ... NOTE
Namespace in Imports field not imported from: 'Matrix'
  All declared Imports should be used.
```

committ: https://github.com/microsoft/LightGBM/pull/2936/commits/d76b551f3090a4084abe48b869d07f8436f3b7fc

.appveyor.yml
```diff
@@ -1,4 +1,4 @@
-﻿version: 2.3.2.{build}
+version: 2.3.2.{build}
```
jameslamb, line 1: this is probably just a line-ending change. I'll try to revert it

jameslamb:
@StrikerRUS @guolinke @Laurae2 I'd like to move this PR forward so we can keep making progress on #629.

I think at this point, the Visual Studio stuff is the main sticking point (e.g. this PR is waiting on #2965 and #2963 ). Right now, it's hard to test changes on those to PRs because I need to merge them into some branch forked off of this one to get all the setup in `.ci/test_r_package_windows.ps1`.

I'd like to propose the following:

1. Remove Visual Studio changes from this PR. So it will add `.ci/test_r_package_windows.ps1` and only testing of the R package  with  `MinGW` on Appveyor
2. Consolidate all Visual Studio stuff on #2963 
    - `FindLibR.cmake` + `CMakeLists.txt`  changes from this PR
    - `.vsts-ci.yml` changes from this PR
    - all changes on #2965 
3. close #2965 

What do you think?

Laurae2:
@jameslamb You can put all on #2963 as needed (especially to have CI working).

jameslamb:
> @jameslamb You can put all on #2963 as needed (especially to have CI working).

awesome, thank you! I think it will help

.appveyor.yml
jameslamb: Found it! It was this https://www.freecodecamp.org/news/a-quick-tale-about-feff-the-invisible-character-cd25cd4630e7/

image

R-package/src/cmake/modules/FindLibR.cmake
```diff
@@ -174,17 +174,19 @@ execute_process(
...
+
 # look for the core R library
 find_library(
   LIBR_CORE_LIBRARY
-  NAMES R
-  HINTS "${CMAKE_CURRENT_BINARY_DIR}" "${LIBR_LIB_DIR}" "${LIBR_HOME}/bin" "${LIBR_LIBRARIES}"
+  NAMES R R.dll
+  HINTS "${CMAKE_CURRENT_BINARY_DIR}" "${LIBR_HOME}/lib" "${LIBR_HOME}/bin/${R_ARCH}" "${LIBR_HOME}/bin" "${LIBR_LIBRARIES}"
+  PATHS "${LIBR_HOME}/bin/${R_ARCH}/R.dll"
 )
```
jameslamb, line 188: @StrikerRUS I'm starting a thread here for [this discussion](https://github.com/microsoft/LightGBM/pull/2963#discussion_r401164197) on how to find `R.dll`.

jameslamb: to summarize for reviewers:

`find_library()` doesn't consider `.dll` by default ([link 1](http://cmake.3232098.n2.nabble.com/find-library-doesn-t-find-dll-on-windows-td7597643.html), [link 2](https://stackoverflow.com/questions/14243524/cmake-find-library-matching-behavior))

`find_library()` is similar to `find_file()`, but it knows about common library prefixes (e.g. `lib`) and suffixes  (e.g. `.so`).

jameslamb: This seems to be working!  In https://ci.appveyor.com/project/guolinke/lightgbm/builds/32185340/job/5tcrah61l6hlv0ab I see

```
Warning message:
In normalizePath(path.expand(path), winslash, mustWork) :
  path[1]="C:/RLibrary/R/lib": The system cannot find the file specified
-- Found LibR: C:/RLibrary/R  
-- LIBR_EXECUTABLE: C:/RLibrary/R/bin/x64/R.exe
-- LIBR_INCLUDE_DIRS: C:/RLibrary/R/include
-- LIBR_CORE_LIBRARY: C:/RLibrary/R/bin/x64/R.dll
```

Don't worry about that "cannot find the file specified" up above those messages, I'm fixing it in #2963

jameslamb:
> > Just rebased to `master` to pick up changes from #2911 and #2957
> 
> this seems to be working! Azure DevOps is failing just because of Azure capacity limitations
> 
> image
> 
> Looks like we are down to 3 NOTEs running `R CMD check --no-multiarch --as-cran` with `MinGW`:
> 
> Appveyor logs (click me)
> And 2 NOTEs running `R CMD check --no-multiarch --no-manual --ignore-vignettes` with `MSVC`:
> 
> Azure DevOps logs (click me)
> I'll add a check on the number of notes as part of this PR, similar to #2942

added this check in https://github.com/microsoft/LightGBM/pull/2936/commits/1b2363a64145a2931f0f02947de751c1f973578c

jameslamb:
> I'd like to propose the following:
> 
> 1. Remove Visual Studio changes from this PR. So it will add 
> 2. Consolidate all Visual Studio stuff on #2963
> 3. close #2965
> 

Alright this PR seems to be working! You can see [in the AppVeyor logs](https://ci.appveyor.com/project/guolinke/lightgbm/builds/32186193/job/xs92r2lkrgxfu1px) that we are getting MinGW picked up for the compiler and R CMD check is passing.

I'd like to kindly request a new review from @StrikerRUS and one R maintainer (either @Laurae2 or @guolinke  ).

I did something _similar_ to the proposal quoted above tonight. I moved all the MSVC-specific stuff out of this PR, including testing on Azure DevOps. If we merge this PR today, we will go from 0 CI coverage of the R package to fully running `R CMD check` on a Windows system using MinGW.

I think we should try to merge the 3 open Windows PRs in the following order

1. #2936 (this PR)
    - _provides CI guarantees that the project can be built on Windows_
2. #2963
    - _fixes builds for VIsual Studio. No CI proving that, but we have evidence from at least one use (https://github.com/microsoft/LightGBM/issues/2995#issuecomment-613196639) that building with MSVC is currently broken and that that PR fixes it. I can manually run on my Windows laptop if we need further evidence but we shouldn'tt delay it waiting to get CI figured out (since we already have no CI coverage of that today and #2936 will prevent a regression for people using MinGW)_
3. #2965 
    - _add CI to cover MSVC, and speed up MSVC builds with some changes to `install.libs.R`_

R-package/src/cmake/modules/FindLibR.cmake
```diff
@@ -174,17 +174,19 @@ execute_process(
...
+set(LIBR_LIB_DIR ${LIBR_LIB_DIR} CACHE PATH "R shared libraries directory")
+
 # look for the core R library
 find_library(
   LIBR_CORE_LIBRARY
-  NAMES R
-  HINTS "${CMAKE_CURRENT_BINARY_DIR}" "${LIBR_LIB_DIR}" "${LIBR_HOME}/bin" "${LIBR_LIBRARIES}"
+  NAMES R R.dll
+  HINTS "${CMAKE_CURRENT_BINARY_DIR}" "${LIBR_HOME}/lib" "${LIBR_HOME}/bin/${R_ARCH}" "${LIBR_HOME}/bin" "${LIBR_LIBRARIES}"
+  PATHS "${LIBR_HOME}/bin/${R_ARCH}/R.dll"
```
StrikerRUS, line 187: Hmm, is `PATHS` here really useful?.. According to the API,
> Additional search locations can be specified after the PATHS argument. 
   https://cmake.org/cmake/help/v3.0/command/find_library.html

and WIKI
> PATHS is used to provide additional search paths for CMake to look into and it should not generally be defined for things other than pkg-config (CMake has its built-in defaults and more can be added as required by various config variables). If you are not using it, leave out the entire section.
https://gitlab.kitware.com/cmake/community/-/wikis/doc/tutorials/How-To-Find-Libraries#finding-files

In addition, I thought that `PATHS` should contain folders, not exact paths.

jameslamb: Ah yeah you are probably right. I originally used it because of the line in the docs that says "PATHS should be used for hard-coded guesses". I'll remove it, good catch!

.ci/test_r_package_windows.ps1
```diff
@@ -0,0 +1,107 @@
...
+$relevant_line = $note_str -match '.*Status: (\d+) NOTE.*'
+$NUM_CHECK_NOTES = $matches[1]
+$ALLOWED_CHECK_NOTES=3
+if ( [int]$NUM_CHECK_NOTES -gt $ALLOWED_CHECK_NOTES ){
+    Write-Output "Found ${NUM_CHECK_NOTES} NOTEs from R CMD check. Only ${ALLOWED_CHECK_NOTES} are allowed"
+    Check-Output $False
+}
+
+Write-Output "No issues were found checking the R package"
+Exit 0
```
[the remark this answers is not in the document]
jameslamb: ah ok let me try!

.ci/test_r_package_windows.ps1
```diff
@@ -0,0 +1,107 @@
...
+if (Get-Content "$LOG_FILE_NAME" | Select-String -Pattern "WARNING" -Quiet) {
+    echo "WARNINGS have been found by R CMD check!"
+    Check-Output $False
+}
+
+$note_str = Get-Content "${LOG_FILE_NAME}" | Select-String -Pattern ' NOTE' | Out-String ; Check-Output $?
+$relevant_line = $note_str -match '.*Status: (\d+) NOTE.*'
+$NUM_CHECK_NOTES = $matches[1]
+$ALLOWED_CHECK_NOTES=3
+if ( [int]$NUM_CHECK_NOTES -gt $ALLOWED_CHECK_NOTES ){
```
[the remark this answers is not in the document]
jameslamb: ha oh ok. I thought that it was like shell scripts where you had to use `[[ ` with a space to use operators like `-gt`.

btw, I was able to write this without pulling out my Windows laptop. I found a free powershell repl: https://tio.run/#powershell

.ci/test_r_package_windows.ps1
```diff
@@ -0,0 +1,107 @@
...
+
+    Write-Output "Installing Rtools"
+    Start-Process -FilePath Rtools.exe -NoNewWindow -Wait -ArgumentList "/VERYSILENT /DIR=$env:R_LIB_PATH/Rtools" ; Check-Output $?
+    Write-Output "Done installing Rtools"
+
+}
+
+# MiKTeX and pandoc can be skipped on Azure builds, since we don't
+# build the package documentation there
+if ($env:AZURE -ne "true") {
```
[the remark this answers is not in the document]
jameslamb: I actually literally meant `AZURE` here. I am not sure yet if the issue is because of Azure or `MSVC` but I really expect that it is  Azure.

I can't find the old comment, so here is the summary...when I was testing earlier commits here, the Azure builds for the R package were  taking 20-25  minutes, even after #2957 . So I decided to change the `R CMD check` commands for AZURE builds to not check any docs:

```
if ($env:AZURE -eq "true") {
  Write-Output "Running R CMD check without checking documentation"
  R.exe CMD check --no-multiarch --no-examples --no-manual --ignore-vignettes ${PKG_FILE_NAME} ; Check-Output $?
```

Once I did that, I was also able to save time by cutting out MiKTeX and pandoc installation. If you are running with `--no-examples --no-manual --ignore-vignettes`, you never need to create the package documentation.

I am ok with this because the docs are already checked in Linux / Mac builds and on the AppVeyor build.

Another possibility that the jobs took SO LONG to run might be based on how Azure does virtualization vs how AppVeyor does it.  I have seen that we use `/MP` [for MSVC builds](https://github.com/microsoft/LightGBM/blob/master/CMakeLists.txt#L196).

From the MSVC docs for `/MP` ([docs](https://docs.microsoft.com/en-us/cpp/build/reference/mp-build-with-multiple-processes?view=vs-2019)):

> /MP[processMax]
> If you omit the processMax argument, the compiler retrieves the number of effective processors on your computer from the operating system, and creates a process for each processor.

If you'd like, I can move any `if ($env:AZURE...` stuff out of this and into #2965 if you'd prefer. I originally didn't do that because I thought it would be complicated to remember to add it back in the right places.

.ci/test_r_package_windows.ps1
```diff
@@ -0,0 +1,107 @@
...
+    [string]$destfile
+  )
+  do {
+    Write-Output "Downloading '${url}'"
+    sleep 5;
+    (New-Object System.Net.WebClient).DownloadFile($url, $destfile)
+  } while(!$?);
+}
+
+$env:R_LIB_PATH = "C:/RLibrary"
```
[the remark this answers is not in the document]
jameslamb: oh ok, good to know! I think it's a good practice in general to try to work below the CI-provided build directory, thanks for the nudge.

.appveyor.yml
```diff
@@ -7,6 +7,9 @@ configuration:  # a trick to construct a build matrix with multiple Python versi
 
 environment:
   matrix:
+    - COMPILER: MINGW
+      TASK: r-package
+      R_WINDOWS_VERSION: 3.6.3
```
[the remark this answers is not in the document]
jameslamb: oh right! I forgot we left it in config files for `test_r_package.sh` because of the weird version numbers for Linux. Ok let me do that, good idea.

.ci/test_r_package_windows.ps1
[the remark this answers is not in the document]
StrikerRUS: OK, I see. Thanks for the explanation! But what about MinGW on Azure? Does it take huge amount of time too?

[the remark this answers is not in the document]
jameslamb: I never tried MinGW on Azure. Let me add a build and see.

.ci/test_r_package_windows.ps1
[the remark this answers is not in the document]
jameslamb: yep that worked!

.ci/test_r_package_windows.ps1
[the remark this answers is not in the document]
jameslamb: seems we still have issues with CTAN mirror connectivity from Azure

image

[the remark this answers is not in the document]
StrikerRUS: The problem is that the URL Azure trying to access is `http://ftp.yzu.edu.tw/CTAN/systems/win32/miktex/tm/packages/miktex-dvisvgm-bin-2.9.tar.lzma%22`. Note `%22` at the end.

And `https://miktex.org/download/win/miktexsetup-x64.zip'` - quote sign at the end.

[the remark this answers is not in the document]
StrikerRUS: Hmm, maybe we can try to pin CRAN mirror?

> --remote-package-repository=url
Download from the specified URL. Use --list-repositories to download an up-to-date list of possible repositories.
https://docs.miktex.org/manual/miktexsetup.html

https://github.com/MiKTeX/miktex/issues/495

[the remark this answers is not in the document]
jameslamb: gaaaah I was looking for this! Thank you.

I chose one from University of Illinois because it seems to have good uptime recently ([CTAN mirror status](http://dante.ctan.org/mirmon/)), although that site is pretty limited so 🤷 

```
$env:CTAN_MIRROR = "https://ctan.math.illinois.edu/systems/win32/miktex/tm/packages/"
```

.ci/test_r_package_windows.ps1
```diff
@@ -0,0 +1,107 @@
...
+[Void][System.IO.Directory]::CreateDirectory($env:R_LIB_PATH)
+
+if ($env:COMPILER -eq "MINGW") {
+  Write-Output "Telling R to use MinGW"
+  $install_libs = "$env:BUILD_SOURCESDIRECTORY\R-package\src\install.libs.R"
+  ((Get-Content -path $install_libs -Raw) -replace 'use_mingw <- FALSE','use_mingw <- TRUE') | Set-Content -Path $install_libs
+}
+
+# set up R if it doesn't exist yet
+if (!(Get-Command R.exe -errorAction SilentlyContinue)) {
```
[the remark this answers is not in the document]
jameslamb: yep great idea!

.ci/test_r_package_windows.ps1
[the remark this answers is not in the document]
StrikerRUS: Wow, it is working now! Thanks a lot for the fix!

StrikerRUS:
Haha, hello  `chocolatey`!

```
-- The C compiler identification is GNU 8.1.0
-- The CXX compiler identification is GNU 8.1.0
-- Check for working C compiler: C:/ProgramData/chocolatey/bin/gcc.exe
-- Check for working C compiler: C:/ProgramData/chocolatey/bin/gcc.exe - works
```

.ci/test_r_package_windows.ps1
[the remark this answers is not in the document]
jameslamb: > Haha, hello `chocolatey`!
> 
> ```
> -- The C compiler identification is GNU 8.1.0
> -- The CXX compiler identification is GNU 8.1.0
> -- Check for working C compiler: C:/ProgramData/chocolatey/bin/gcc.exe
> -- Check for working C compiler: C:/ProgramData/chocolatey/bin/gcc.exe - works
> ```

haha NOOOOOOOOOOO

I probably need to put the RTools-bundled MinGW on `PATH`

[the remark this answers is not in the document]
jameslamb: haha yes I kind of agree. I think pinning to a CTAN mirror will still be useful, it's one less thing that can change in our builds.

jameslamb:
> Haha, hello `chocolatey`!
> 
> ```
> -- The C compiler identification is GNU 8.1.0
> -- The CXX compiler identification is GNU 8.1.0
> -- Check for working C compiler: C:/ProgramData/chocolatey/bin/gcc.exe
> -- Check for working C compiler: C:/ProgramData/chocolatey/bin/gcc.exe - works
> ```

haha NOOOOOOOOOOO

I probably need to put the RTools-bundled MinGW on PATH

R-package/src/cmake/modules/FindLibR.cmake
jameslamb: Struggling to get this working on Azure. I wrote up this comment bit-by-bit as I tried things, which is how it got so long.

**the problem**:

> We have `"${LIBR_HOME}/bin/${R_ARCH}"` in the set of paths to search, and `LIBR_CORE_LIBRARY` isn't being found...but when I set it directly it works.

When I say "directly", I mean going back to what I originally had

```
set (LIBR_CORE_LIBRARY ${LIBR_HOME}/bin/${R_ARCH}/R.dll CACHE PATH "R core shared library")
```

Let me show you:

* Just adding MinGW to PATH (plus other extra logging that I'll remove):
    - commit: https://github.com/jameslamb/LightGBM/pull/19/commits/05c59f48505c2c1231d3d5e1d2b9059a55602c02
    - AppVeyor: [works](https://ci.appveyor.com/project/jameslamb/lightgbm/builds/32280450/job/vk10elspubecgufx)
    - Azure: [fails](https://dev.azure.com/JayLamb20/jameslamb-lightgbm/_build/results?buildId=389&view=logs&j=b51fa6ee-4364-59e6-4527-dc4aaafd2db2&t=85bde6ad-0953-589b-e1e8-633bb2d9d530)
        - `Could NOT find LibR (missing: LIBR_CORE_LIBRARY)`
* Hardcoding path guess:
    -  commit: https://github.com/jameslamb/LightGBM/pull/19/commits/fdd57243537a125da0e0c176128f6e1beee9a043
    - AppVeyor: [works](https://ci.appveyor.com/project/jameslamb/lightgbm/builds/32280547)
        - you can see log message `LIBR_CORE_LIBRARYwas found by find_library()`
    - Azure: [works](https://dev.azure.com/JayLamb20/jameslamb-lightgbm/_build/results?buildId=391&view=logs&j=b51fa6ee-4364-59e6-4527-dc4aaafd2db2&t=85bde6ad-0953-589b-e1e8-633bb2d9d530)
        - you can see log message `LIBR_CORE_LIBRARY still not found, trying hard-coding`

From the `Get-ChildItem` calls and added cmake logs, I can see in the logs that these are the relevant paths

* AppVeyor
    - R.dll: `C:\projects\lightgbm\RLibrary\R\bin\x64\R.dll`
    - `LIBR_HOME`: `C:/projects/lightgbm/RLibrary/R`
* Azure
    - R.dll: `D:\a\1\s\RLibrary\R\bin\x64\R.dll`
    - `LIBR_HOME`: `D:/a/1/s/RLibrary/R`

So then I thought "ok, the paths are the same...maybe the `CMake` versions are different and `find_library()` behaves differently...they are! and it does!

**CMake versions**

* Appveyor: 3.16
    -  I can see `C:/Program Files (x86)/CMake/share/cmake-3.16` in the logs
* Azure: 3.17
    - following the [Microsoft docs on the 'vs2017-win2016' image we use](https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops) to [the list of installed software](https://github.com/actions/virtual-environments/blob/master/images/win/Windows2016-Readme.md#cmake)

We did hear in https://github.com/microsoft/LightGBM/issues/2995#issuecomment-613167751 that our setup wasn't working with `CMake` 3.17.

I found an important-looking line in https://cmake.org/cmake/help/v3.17/release/3.17.html#id17

> When using MinGW tools, the find_library() command no longer finds .dll files by default. Instead, it expects .dll.a import libraries to be available.

I'm confused though, it seems like passing `R.dll`  to `NAMES` should work ([CMake docs](https://cmake.org/cmake/help/latest/command/find_library.html)):

> Each library name given to the NAMES option is first considered as a library file name and then considered with platform-specific prefixes (e.g. lib) and suffixes (e.g. .so). Therefore one may specify library file names such as libfoo.a directly. 

I'm going to try something like this:

```
# look for the core R library
find_library(
  LIBR_CORE_LIBRARY
  NAMES R R.dll
  HINTS "${CMAKE_CURRENT_BINARY_DIR}" "${LIBR_HOME}/lib" "${LIBR_HOME}/bin/${R_ARCH}" "${LIBR_HOME}/bin" "${LIBR_LIBRARIES}"
)

if (WIN32 AND NOT LIBR_CORE_LIBRARY)
    find_file(
        LIBR_CORE_LIBRARY
        NAME R.dll
        HINTS "${CMAKE_CURRENT_BINARY_DIR}" "${LIBR_HOME}/lib" "${LIBR_HOME}/bin/${R_ARCH}" "${LIBR_HOME}/bin" "${LIBR_LIBRARIES}"
    )
endif()
```

I'd rather do that than replace the origin `find_library()` call completely, because I have seen it called different things on some Linux system (like `libR.so` or `R.so`) and I don't want our CMake code to have to know about all the variables CMake uses to store different library prefixes and suffixes.

jameslamb: I tried that strategy and it seems like it worked!

* AppVeyor: [works](https://ci.appveyor.com/project/jameslamb/lightgbm/builds/32280807/job/0mw3esec8kaigbcm)
* Azure: [works](https://dev.azure.com/JayLamb20/jameslamb-lightgbm/_build/results?buildId=392&view=logs&j=b51fa6ee-4364-59e6-4527-dc4aaafd2db2&t=85bde6ad-0953-589b-e1e8-633bb2d9d530)
    - _and you can see `Rtools/mingw_64/bin/g++.exe` is being chosen as the compiler_

.ci/test_r_package_windows.ps1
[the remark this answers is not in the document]
jameslamb: changed back to "don't run R docs on Azure" here: https://github.com/microsoft/LightGBM/commit/d843795087e363ffe7bbc53c502fdbe8d36df9fc

R-package/src/cmake/modules/FindLibR.cmake
StrikerRUS: Hmm, so the root cause was in different CMake versions, right? Thanks a lot for your investigation!

> put MinGW compilers on PATH so they're found instead of the ones from chocolatey

I think that more reliable solution will be to use `$env:CXX = path_to_right_compiler`, like we do in the following snippet 
https://github.com/microsoft/LightGBM/blob/0d3e2046533acdda62f15936fb8636f7c8bc27d6/.ci/test.sh#L3-L9
and it is suggested in R installation script
https://github.com/microsoft/LightGBM/blob/0d3e2046533acdda62f15936fb8636f7c8bc27d6/build_r.R#L1-L6

> changed back from "don't run R docs on MinGW builds" to "don't run R docs on Azure"

Sorry, didn't get it. Why did you change it back? I think we can test docs in case of MinGW on Azure.

.ci/test_r_package_windows.ps1
```diff
@@ -0,0 +1,108 @@
...
+  )
+  do {
+    Write-Output "Downloading ${url}"
+    sleep 5;
+    (New-Object System.Net.WebClient).DownloadFile($url, $destfile)
+  } while(!$?);
+}
+
+$env:R_WINDOWS_VERSION = "3.6.3"
+$env:R_LIB_PATH = "$env:BUILD_SOURCESDIRECTORY/RLibrary" -replace '[\\]', '/'
```
StrikerRUS, line 16: Hmm, R tools cannot work with Win paths?! Interesting...

.vsts-ci.yml
```diff
@@ -119,6 +119,9 @@ jobs:
   strategy:
     maxParallel: 3
     matrix:
+      rpkg:
```
StrikerRUS, line 122: For the consistency with Linux and macOS builds.
```suggestion
      r_package:
```

R-package/src/cmake/modules/FindLibR.cmake
```diff
@@ -174,17 +174,28 @@ execute_process(
...
+set(LIBR_INCLUDE_DIRS ${LIBR_INCLUDE_DIRS} CACHE PATH "R include directory")
+set(LIBR_LIB_DIR ${LIBR_LIB_DIR} CACHE PATH "R shared libraries directory")
+
 # look for the core R library
 find_library(
   LIBR_CORE_LIBRARY
-  NAMES R
-  HINTS "${CMAKE_CURRENT_BINARY_DIR}" "${LIBR_LIB_DIR}" "${LIBR_HOME}/bin" "${LIBR_LIBRARIES}"
+  NAMES R R.dll
+  HINTS "${CMAKE_CURRENT_BINARY_DIR}" "${LIBR_HOME}/lib" "${LIBR_HOME}/bin/${R_ARCH}" "${LIBR_HOME}/bin" "${LIBR_LIBRARIES}"
```
StrikerRUS, line 186: Please save `HINTS` into a variable to not list them twice in `find_library` and `find_file`.

R-package/src/cmake/modules/FindLibR.cmake
```diff
@@ -174,17 +174,28 @@ execute_process(
...
 )
 
-set(LIBR_HOME ${LIBR_HOME} CACHE PATH "R home directory")
-set(LIBR_EXECUTABLE ${LIBR_EXECUTABLE} CACHE PATH "R executable")
-set(LIBR_INCLUDE_DIRS ${LIBR_INCLUDE_DIRS} CACHE PATH "R include directory")
-set(LIBR_LIB_DIR ${LIBR_LIB_DIR} CACHE PATH "R shared libraries directory")
+# starting from CMake 3.17, find_library() will not find .dll files by default
+# https://cmake.org/cmake/help/v3.17/release/3.17.html#id17
```
StrikerRUS, line 190:
```suggestion
# https://cmake.org/cmake/help/v3.17/release/3.17.html#other-changes
```

.ci/test_r_package_windows.ps1
jameslamb: I was really surprised, but the builds started breaking after I changed from `C:/Rlibrary` to `BUILD_SOURCESDIRECTORY`, and the only difference was that `BUILD_SOUORCEDIRECTORY` included backslashes.

R-package/src/cmake/modules/FindLibR.cmake
jameslamb: oh ok sure

R-package/src/cmake/modules/FindLibR.cmake
jameslamb: > why did you change it back

Isn't that what you were supporting by https://github.com/microsoft/LightGBM/pull/2936#discussion_r410226345? Maybe I misunderstood.

> I think that more reliable solution will be to use $env:CXX = path_to_right_compiler, like we do in the following snippet

Sure I can do that! But since both chocolatey and MinGW are `g++.exe`, unlike in the example you shared I'll have to do the full path. Like `$env:CXX = "$env:R_LIBS_PATH/Rtools/mingw_64/bin/g++.exe"`.

Let me try that.

StrikerRUS: > Isn't that what you were supporting by #2936 (comment)? Maybe I misunderstood.

That comment was about what I thought **before** CTAN pinning allowed us to move forward. Sorry for confusing you.

The current plan may be the following.
1. In this PR enable full (`--as-cran`) `MINGW` builds at both Appveyor and Azure.
2. In next PR after speeding up VS builds change `MINGW` to `VS` at Azure to cover more use cases.
3. If building time will still take so long, change `--as-cran` to config without docs building in case of `VS`.

Please let me know WDYT about that.

StrikerRUS: > Sure I can do that! But since both chocolatey and MinGW are `g++.exe`, unlike in the example you shared I'll have to do the full path. Like `$env:CXX = "$env:R_LIBS_PATH/Rtools/mingw_64/bin/g++.exe"`.

Yeah, for sure! Just like in 
https://github.com/microsoft/LightGBM/blob/0d3e2046533acdda62f15936fb8636f7c8bc27d6/build_r.R#L4

jameslamb: I agree with this plan --> https://github.com/microsoft/LightGBM/pull/2936#discussion_r410927991

jameslamb: Ok these changes  worked: https://github.com/jameslamb/LightGBM/pull/19/commits/c145b7dac650e95ebe0491ee6c6d32768fd22dc8

* Azure [works, 13 minutes](https://dev.azure.com/JayLamb20/jameslamb-lightgbm/_build/results?buildId=393&view=logs&j=9f61088a-36ee-57e4-0947-03329fd6ee90&t=2528b315-ee14-54c7-8330-fcfcaf0cbb8e)
* AppVeyor [works, 11 minutes](https://ci.appveyor.com/project/jameslamb/lightgbm/builds/32288119/job/f4uswqd26m6v7uvw)

Added back to this PR in: https://github.com/microsoft/LightGBM/pull/2936/commits/6ca2be8f3c66f16607f655344e51030e4e56c8ec

jameslamb: One note...in https://github.com/microsoft/LightGBM/pull/2936/commits/6ca2be8f3c66f16607f655344e51030e4e56c8ec, I removed `Rtools` mingw from `PATH`, so we could be sure that setting `CXX` explicitly was working.  I think that in #2965 we'll have to add it back, because that directory is also where `gendef.exe` is based.

I'll add a link to this comment on this PR.

R-package/src/cmake/modules/FindLibR.cmake
jameslamb: ah I forgot to do this in recent commits. Added in https://github.com/microsoft/LightGBM/pull/2936/commits/cdf1df8696846b86853d1858996e0211805d0eed

jameslamb:
@StrikerRUS thanks for the great review so far!

@Laurae2 @guolinke could one of you take a look at this? @StrikerRUS and I will keep talking through miscellaneous things in CI scripts, but I don't want to merge this until an R reviewer also approves.

.ci/test_r_package_windows.ps1
```diff
@@ -0,0 +1,110 @@
...
+Write-Output "R_LIB_PATH: $env:R_LIB_PATH"
```
StrikerRUS, line 17: I guess this line was accidentally left after debugging, wasn't it?

.ci/test_r_package_windows.ps1
```diff
@@ -0,0 +1,110 @@
...
+Write-Output "Done installing R"
+
+Write-Output "Installing Rtools"
+Start-Process -FilePath Rtools.exe -NoNewWindow -Wait -ArgumentList "/VERYSILENT /DIR=$env:R_LIB_PATH/Rtools" ; Check-Output $?
+Write-Output "Done installing Rtools"
+
+# MiKTeX and pandoc can be skipped on non-MINGW builds, since we don't
+# build the package documentation for those
+if ($env:COMPILER -eq "MINGW") {
+
```
StrikerRUS, line 54: Please remove this empty line.

.ci/test_r_package_windows.ps1
```diff
@@ -0,0 +1,110 @@
...
+$packages = "c('data.table', 'jsonlite', 'Matrix', 'R6', 'testthat'), dependencies = c('Imports', 'Depends', 'LinkingTo')"
+Rscript --vanilla -e "options(install.packages.check.source = 'no'); install.packages($packages, repos = '$env:CRAN_MIRROR', type = 'binary', lib = '$env:R_LIB_PATH')" ; Check-Output $?
+
+Write-Output "Building R package"
+Rscript build_r.R --skip-install ; Check-Output $?
+
+$PKG_FILE_NAME = Get-Item *.tar.gz
+$LOG_FILE_NAME = "lightgbm.Rcheck/00check.log"
+
+$env:_R_CHECK_FORCE_SUGGESTS_=0
```
StrikerRUS, line 81:
```suggestion
$env:_R_CHECK_FORCE_SUGGESTS_ = 0
```

.ci/test_r_package_windows.ps1
```diff
@@ -0,0 +1,110 @@
...
+Write-Output "Looking for issues with R CMD check results"
+if (Get-Content "$LOG_FILE_NAME" | Select-String -Pattern "WARNING" -Quiet) {
+    echo "WARNINGS have been found by R CMD check!"
+    Check-Output $False
+}
+
+$note_str = Get-Content "${LOG_FILE_NAME}" | Select-String -Pattern ' NOTE' | Out-String ; Check-Output $?
+$relevant_line = $note_str -match '.*Status: (\d+) NOTE.*'
+$NUM_CHECK_NOTES = $matches[1]
+$ALLOWED_CHECK_NOTES=3
```
StrikerRUS, line 104:
```suggestion
$ALLOWED_CHECK_NOTES = 3
```

.ci/test_r_package_windows.ps1
```diff
@@ -0,0 +1,110 @@
...
+Write-Output "R CMD check build logs:"
+Get-Content -Path $env:BUILD_SOURCESDIRECTORY\lightgbm.Rcheck\00install.out
+
+Check-Output $check_succeeded
+
+Write-Output "Looking for issues with R CMD check results"
+if (Get-Content "$LOG_FILE_NAME" | Select-String -Pattern "WARNING" -Quiet) {
+    echo "WARNINGS have been found by R CMD check!"
+    Check-Output $False
+}
```
StrikerRUS, line 99: Hmmm... Is something wrong with R here?
```
WARNING: directory 'lightgbm/src/compute/test' is empty
* looking to see if a 'data/datalist' file should be added
* building 'lightgbm_2.3.2.tar.gz'
```
but then
```
Status: 3 NOTEs
```

.vsts-ci.yml
```diff
@@ -119,6 +119,9 @@ jobs:
   strategy:
     maxParallel: 3
```
StrikerRUS, line 120:
```suggestion
    maxParallel: 4
```

R-package/src/cmake/modules/FindLibR.cmake
```diff
@@ -174,17 +174,31 @@ execute_process(
   OUTPUT_VARIABLE LIBR_LIB_DIR
 )
 
+set(LIBR_HOME ${LIBR_HOME} CACHE PATH "R home directory")
+set(LIBR_EXECUTABLE ${LIBR_EXECUTABLE} CACHE PATH "R executable")
+set(LIBR_INCLUDE_DIRS ${LIBR_INCLUDE_DIRS} CACHE PATH "R include directory")
+set(LIBR_LIB_DIR ${LIBR_LIB_DIR} CACHE PATH "R shared libraries directory")
+
+# where is R.so / R.dll / libR.so likely to be found?
+set(LIBR_PATH_HINTS "${CMAKE_CURRENT_BINARY_DIR}" "${LIBR_HOME}/lib" "${LIBR_HOME}/bin/${R_ARCH}" "${LIBR_HOME}/bin" "${LIBR_LIBRARIES}")
```
StrikerRUS, line 183:
```suggestion
set(LIBR_PATH_HINTS "${CMAKE_CURRENT_BINARY_DIR}" "${LIBR_LIB_DIR}" "${LIBR_HOME}/bin/${R_ARCH}" "${LIBR_HOME}/bin" "${LIBR_LIBRARIES}")
```
Or am I missing something?

.ci/test_r_package_windows.ps1
jameslamb: oh sure, can be removed

.ci/test_r_package_windows.ps1
jameslamb: that isn't an `R CMD check` warning, it's a warning from `R CMD build`. We can ignore it.

jameslamb:
> @jameslamb Can we fix this warning?
> 
> ```
> Warning message:
> In normalizePath(path.expand(path), winslash, mustWork) :
>   path[1]="C:/projects/lightgbm/RLibrary/R/lib": The system cannot find the file specified
> ```
> 
> Please check one more round of review below.
> 
> @guolinke Printing out build logs with MinGW allowed to spot many compilation warnings. Can you please check them?

On the point of compilation warnings...if we determine that they were not introduced by this PR (I really think they were not), let's please open a separate issue and  not delay this PR any further.

The `normalizePath()` warning is fixed in #2963. The issue is that we use `R.home('lib')` but that function just constructs a directory. It doesn't guarantee that that directory exists. That code is used to generate `LIBR_LIB_DIR`, which is only necessary for MSVC builds. I completely removed it in #2963. It can safely be ignored in this PR and will be addressed in the followup MSVC PRs.

R-package/src/cmake/modules/FindLibR.cmake
jameslamb: that's fine. This and LIBR_LIB_DIR are going away in #2963 anyway.

.ci/test_r_package_windows.ps1
jameslamb: removed  in https://github.com/microsoft/LightGBM/pull/2936/commits/747378eb41b74d5124b905f9cca5f76727d3a5ae

.ci/test_r_package_windows.ps1
jameslamb: removed in https://github.com/microsoft/LightGBM/pull/2936/commits/747378eb41b74d5124b905f9cca5f76727d3a5ae

.vsts-ci.yml
jameslamb: ok added https://github.com/microsoft/LightGBM/pull/2936/commits/c0b87d6bb22b0d863bb99e1024371510133f8e44

.ci/test_r_package_windows.ps1
StrikerRUS: Hmm, interesting! Thanks for the explanation!

StrikerRUS:
@jameslamb 
> On the point of compilation warnings...if we determine that they were not introduced by this PR (I really think they were not), let's please open a separate issue and not delay this PR any further.

I wanted to init the discussion here where we can see these logs live. Of course, they shouldn't prevent merging of this PR in any way.

> The normalizePath() warning is fixed in #2963. The issue is that we use R.home('lib') but that function just constructs a directory.

Ah, OK!

.ci/test_r_package_windows.ps1
```diff
@@ -0,0 +1,108 @@
...
+    Add-Type -AssemblyName System.IO.Compression.FileSystem
+    [System.IO.Compression.ZipFile]::ExtractToDirectory("miktexsetup-x64.zip", "miktex")
+    Write-Output "Setting up MiKTeX"
+    .\miktex\miktexsetup.exe --remote-package-repository="$env:CTAN_MIRROR" --local-package-repository=./miktex/download --package-set=essential --quiet download ; Check-Output $?
+    Write-Output "Installing MiKTeX"
+    .\miktex\download\miktexsetup.exe --remote-package-repository="$env:CTAN_MIRROR" --portable="$env:R_LIB_PATH/miktex" --quiet install ; Check-Output $?
+    Write-Output "Done installing MiKTeX"
+
+    initexmf --set-config-value [MPM]AutoInstall=1
+    conda install -y --no-deps pandoc
```
StrikerRUS, line 64:
```suggestion
    conda install -q -y --no-deps pandoc
```

.ci/test_r_package_windows.ps1
```diff
@@ -0,0 +1,108 @@
...
+Write-Output "Building R package"
+Rscript build_r.R --skip-install ; Check-Output $?
+
+$PKG_FILE_NAME = Get-Item *.tar.gz
+$LOG_FILE_NAME = "lightgbm.Rcheck/00check.log"
+
+$env:_R_CHECK_FORCE_SUGGESTS_ = 0
+if ($env:COMPILER -ne "MINGW") {
+  Write-Output "Running R CMD check without checking documentation"
+  R.exe CMD check --no-multiarch --no-examples --no-manual --ignore-vignettes ${PKG_FILE_NAME} ; $check_succeeded=$?
```
StrikerRUS, line 82:
```suggestion
  R.exe CMD check --no-multiarch --no-examples --no-manual --ignore-vignettes ${PKG_FILE_NAME} ; $check_succeeded = $?
```

.ci/test_r_package_windows.ps1
```diff
@@ -0,0 +1,108 @@
...
+} else {
+  Write-Output "Running R CMD check as CRAN"
+  R.exe CMD check --no-multiarch --as-cran ${PKG_FILE_NAME} ; $check_succeeded=$?
```
StrikerRUS, line 85:
```suggestion
  R.exe CMD check --no-multiarch --as-cran ${PKG_FILE_NAME} ; $check_succeeded = $?
```

R-package/src/cmake/modules/FindLibR.cmake
```diff
@@ -174,17 +174,31 @@ execute_process(
...
+  HINTS ${LIBR_PATH_HINTS}
 )
 
-set(LIBR_HOME ${LIBR_HOME} CACHE PATH "R home directory")
-set(LIBR_EXECUTABLE ${LIBR_EXECUTABLE} CACHE PATH "R executable")
-set(LIBR_INCLUDE_DIRS ${LIBR_INCLUDE_DIRS} CACHE PATH "R include directory")
-set(LIBR_LIB_DIR ${LIBR_LIB_DIR} CACHE PATH "R shared libraries directory")
+# starting from CMake 3.17, find_library() will not find .dll files by default
+# https://cmake.org/cmake/help/v3.17/release/3.17.html#other-changes
+if (WIN32 AND NOT LIBR_CORE_LIBRARY)
```
StrikerRUS, line 194: Sorry, I forgot: do we need this for MinGW too? Or we can add `AND MSVC`? 
```suggestion
if (WIN32 AND MSVC AND NOT LIBR_CORE_LIBRARY)
```

.ci/test_r_package_windows.ps1
```diff
@@ -0,0 +1,108 @@
...
+}
+
+Write-Output "R CMD check build logs:"
+Get-Content -Path $env:BUILD_SOURCESDIRECTORY\lightgbm.Rcheck\00install.out
```
StrikerRUS, line 89: Will it make sense to do the same for non-Windows builds in `test_r_package.sh`?

R-package/src/cmake/modules/FindLibR.cmake
jameslamb: We need it for both. `LIBR_CORE_LIBRARY` gets passed to the linker when building the `lib_lightgbm.dll` --> https://github.com/microsoft/LightGBM/blob/master/CMakeLists.txt#L318-L320

.ci/test_r_package_windows.ps1
jameslamb: ah yes definitely!  I'll do that in a separate PR right now

jameslamb: #3009

jameslamb:
Just merged in `master` to get all the new tests from #3002, #3004, #3005, #3006, #3007, and #3009.

@Laurae2 @guolinke once this builds, it is just waiting on a  review  from one of you whenever you have time. Thank you!

R-package/src/install.libs.R
```diff
@@ -67,6 +67,7 @@ if (!use_precompile) {
   # Check if Windows installation (for gcc vs Visual Studio)
   if (WINDOWS) {
     if (use_mingw) {
+      print("Trying to build with MinGW")
```
Laurae2, line 70: Change in another PR `print` for `message`.

.ci/test_r_package_windows.ps1
```diff
@@ -0,0 +1,108 @@
...
+    [string]$destfile
+  )
+  do {
+    Write-Output "Downloading ${url}"
+    sleep 5;
+    (New-Object System.Net.WebClient).DownloadFile($url, $destfile)
+  } while(!$?);
+}
+
+$env:R_WINDOWS_VERSION = "3.6.3"
```
Laurae2, line 15: Do you think we should test with R 4.0 ?

R-package/src/cmake/modules/FindLibR.cmake
```diff
@@ -174,17 +174,31 @@ execute_process(
   OUTPUT_VARIABLE LIBR_LIB_DIR
 )
 
+set(LIBR_HOME ${LIBR_HOME} CACHE PATH "R home directory")
+set(LIBR_EXECUTABLE ${LIBR_EXECUTABLE} CACHE PATH "R executable")
+set(LIBR_INCLUDE_DIRS ${LIBR_INCLUDE_DIRS} CACHE PATH "R include directory")
+set(LIBR_LIB_DIR ${LIBR_LIB_DIR} CACHE PATH "R shared libraries directory")
```
Laurae2, line 180: In another PR, maybe we should change the name and use `LIBR_LIB_DIRS` instead of `LIBR_LIB_DIR` (cmake Find result variables should be plural for include_dirs / lib_dirs, usually they are placeholders for include_dir / lib_dir if they are singular directories).

.ci/test_r_package_windows.ps1
jameslamb: yes I think we should have a separate issue for testing on R 4.0.0 across all platforms, now that it's officially released. Good call!

I just opened #3024 to capture that work (and added it to  #2302 )

R-package/src/cmake/modules/FindLibR.cmake
jameslamb: In this case it is a single directory. This isn't "all directories passed to the linker in `-L` flags", it is "what is the one directory where `R.dll` / `R.so`  is likely to live".

Anyway, I'm removing `LIBR_LIB_DIR` completely in #2963 . It's only being used in MSVC installs, and only indirectly as a way to find `LIBR_CORE_LIBRARY`.

R-package/src/install.libs.R
jameslamb: sure! Let's talk about it in a separate issue / PR. We have several other `print()` statements in `install.libs.R` and no uses of `message()` so I'm curious what makes `message()` preferable here.

jameslamb:
Thanks for the helpful reviews! I'm excited to merge this one 😀

StrikerRUS:
@guolinke Now as we run R-package tests at Appveyor we'll face corresponding job failures due to CRAN/CTAN network availability issues quite often (for example, refer to the [latest master failure](https://ci.appveyor.com/project/guolinke/lightgbm/builds/32879719/job/qpoqp18pa7mqd9xp)). I wonder, is it possible to give maintainers privileges to re-run failed jobs as we have for Travis/Azure?

It seems that it is technically possible now: https://www.appveyor.com/docs/team-setup/.
But it may require to manually transfer the current account (via e-mail request).

jameslamb:
^ related to https://github.com/microsoft/LightGBM/pull/2936#issuecomment-628828742, without that we can find ourselves in this situation where one long-running AppVeyor build is blocking all others from starting:

image

Notice that the build at the bottom has been running for 45 minutes. It is stuck trying to download something from a CTAN mirror. Ideally, I'd like to be able to kill it in the AppVeyor UI. But since I don't have permissions right now all other builds have to wait.

StrikerRUS:
Gently ping @guolinke for https://github.com/microsoft/LightGBM/pull/2936#issuecomment-628828742

guolinke:
@StrikerRUS @jameslamb 
I have sent invitation to you.

jameslamb:
@guolinke thanks! Although now that we have #3168 , as soon as that is merged AppVeyor should no longer be a problem 😂 

That PR and all others are blocked until you can help us address  https://github.com/microsoft/LightGBM/pull/3000#issuecomment-646380113

StrikerRUS:
@guolinke Thanks a lot! Now I can re-run failed jobs.

PR #2938, opened 2020-03-24
replace built-in assert with own CHECK macro

StrikerRUS:
It removes the difference between R-specific library and ordinary one.

Refer to #2837.

### DIFF
--- a/src/io/json11.cpp
+++ b/src/io/json11.cpp
@@ -20,13 +20,12 @@
  */
 #include <LightGBM/utils/json11.h>
 
-#include <limits>
-#ifndef LGB_R_BUILD
-  #include <cassert>
-#endif
+#include <LightGBM/utils/log.h>
+
 #include <cmath>
 #include <cstdio>
 #include <cstdlib>
+#include <limits>
 
 namespace json11 {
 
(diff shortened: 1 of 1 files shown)


StrikerRUS:
@jameslamb 
> Why is `LightGBM::Log` required for this?

Because Json11 uses its own namespace 

https://github.com/microsoft/LightGBM/blob/0341906c45666ccaeca92ee5fbee17fa22282470/include/LightGBM/utils/json11.h#L59

and `Log` is defined in `LightGBM` namespace

https://github.com/microsoft/LightGBM/blob/0341906c45666ccaeca92ee5fbee17fa22282470/include/LightGBM/utils/log.h#L24

Refer to https://stackoverflow.com/a/11791202.

jameslamb:
> @jameslamb
> 
> > Why is `LightGBM::Log` required for this?
> 
> Because Json11 uses its own namespace
> 
> https://github.com/microsoft/LightGBM/blob/0341906c45666ccaeca92ee5fbee17fa22282470/include/LightGBM/utils/json11.h#L59
> 
> and `Log` is defined in `LightGBM` namespace
> 
> https://github.com/microsoft/LightGBM/blob/0341906c45666ccaeca92ee5fbee17fa22282470/include/LightGBM/utils/log.h#L24
> 
> Refer to https://stackoverflow.com/a/11791202.

oooooooh wild, ok! thanks for the explanation

jameslamb:
@chivee or @guolinke could you take a look and let us know if you are ok with this?

Thanks!

PR #2942, opened 2020-03-25
[R-package] [ci] added check on number of R CMD CHECK notes

jameslamb:
Currently in CI, we run `R CMD check` on the R package and fail builds if ERRORs or WARNINGs  are encountered. As we get closer to CRAN (#629), I'd like to start limiting the NOTEs being generated. Currently Linux and macOS builds have 3 NOTEs. One is being addressed in #2911 , the other two are not problematic.

Following @StrikerRUS 's suggestion in https://github.com/microsoft/LightGBM/pull/2911#issuecomment-602979163, this PR proposing putting an upper limit on the number of NOTEs.

This PR has it set to 3, and mainly introduces _how we'll enforce this_. I think we should merge this, then change it from 3 to 2 on #2911 . I think it's better for this to be a separate PR so that other changes in the repo don't introduce new notes (like I accidentally did in #2803 and then had to fix in #2928 ).

### DIFF
--- a/.ci/test_r_package.sh
+++ b/.ci/test_r_package.sh
@@ -91,4 +91,15 @@ if grep -q -R "WARNING" "$LOG_FILE_NAME"; then
     exit -1
 fi
 
+ALLOWED_CHECK_NOTES=3
+NUM_CHECK_NOTES=$(
+    cat ${LOG_FILE_NAME} \
+        | grep -e '^Status: .* NOTE.*' \
+        | sed 's/[^0-9]*//g'
+)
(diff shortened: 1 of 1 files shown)


jameslamb:
> LGTM, thanks!
> 
> Can you please push a commit with `ALLOWED_CHECK_NOTES=2` and then revert it to make sure that corresponding CI builds fail as expected?

oh sure! I did that on my fork already actually

Travis:  https://travis-ci.org/github/jameslamb/LightGBM
Commit: https://github.com/jameslamb/LightGBM/commit/05d86944896f3b0d17698f309771e1dbd4dfd1a3

You can see this error in the logs:

image

StrikerRUS:
@jameslamb 
Great, many thanks!

Issue #2944, opened 2020-03-25
[R-package] [ci] Cover 100% of R-to-C++ calls with unit tests


jameslamb:
`LightGBM`'s R package calls its C++ library using R's `.Call()` interface. That C++ is one codebase that is reused as a standalone C++ library, with Java, with Spark (via the Java interface), with Python, and from a command-line interface.

Because that library is shared, it's regularly updated by developers who have no knowledge of the `lightgbm` R package. The R package's unit tests are the best defense against such updates having unintended consequences and breaking the R package.

The following calls to the C++ side from within the R package are not covered by this project's unit tests (this will be updated as tests are added):

* [x] `LGBM_BoosterFree_R` (#2945)
* [x] `LGBM_BoosterCreateFromModelfile_R` (#2945)
* [x] `LGBM_BoosterGetCurrentIteration_R` (#2945)
* [x] `LGBM_BoosterMerge_R` (#3007)
* [x] `LGBM_BoosterLoadModelFromString_R` (#2945)
* [x] `LGBM_BoosterResetTrainingData_R` (#3020)
* [x] `LGBM_BoosterRollbackOneIter_R` (#3006)
* [x] `LGBM_BoosterSaveModelToString_R` (#2945)
* [x] `LGBM_GetLastError_R` (#3005)

I figured that out by following [the coverage instructions in the R-package README](https://github.com/microsoft/LightGBM/tree/master/R-package#testing).

Covering those calls is important because it would improve our confidence in releases and allow non-R developers to work on the C++ library without worrying that their changes are negatively impacting the R package.

This issue will likely not be closed by a single PR. Any PR that addresses some of those functions above would be a welcome addition.

jameslamb:
Closing this issue in favor of being in #2302 . If you'd like to address this issue, leave a comment here and it will be re-opened.

jameslamb:
I'm going to open this since we have an open PR for it and I'm working on the other PRs in the next  few days

jameslamb:
All functions have been covered now that we've merged #3020, so this can be closed

PR #2945, opened 2020-03-25
[R-package] Added unit tests on creating Booster from file or string

jameslamb:
This PR adds unit tests on the code in the R package that handles saving a Booster to file or string and creating a new Booster from a file or string. This brings the unit test coverage on the package from 78.20% to 79.71%.

Importantly, it covers calls from R to C++ functions `LGBM_BoosterSaveModelToString_R`, `LGBM_BoosterCreateFromModelfile_R`, and `LGBM_BoosterLoadModelFromString_R`  which were previously not covered (see #2944 for why this is valuable).

This PR does not change the behavior of `lgb.load()` at all, but it does reorganize the internals to cut out some redundant checks and make the logic a little easier to follow.

This PR also removes an unnecessary step from the docs on generating test coverage in `R-package/README.md`. Removing the `build/` directory is no longer necessary, as of #2909 .

### DIFF
--- a/R-package/R/lgb.Booster.R
+++ b/R-package/R/lgb.Booster.R
@@ -798,18 +798,18 @@ lgb.load <- function(filename = NULL, model_str = NULL) {
   filename_provided <- !is.null(filename)
   model_str_provided <- !is.null(model_str)
 
-  if (filename_provided){
-    if (!is.character(filename)){
+  if (filename_provided) {
+    if (!is.character(filename)) {
       stop("lgb.load: filename should be character")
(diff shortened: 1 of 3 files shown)


jameslamb:
Rebased to `master`, to get the changes  from #2954 , so the R test issues should be resolved. That will also fix the weird state of the Travis build, which was a result of [the recent Travis outage](https://www.traviscistatus.com/).

jameslamb:
gently ping @Laurae2 @guolinke for a review

PR #2949, opened 2020-03-27
[python] Re-enable scikit-learn 0.22+ support

### LINKED ISSUE #2628: [RFC] compatibility with scikit-learn

StrikerRUS:
Skip failing `check_no_attributes_set_in_init` test, refer to  https://github.com/microsoft/LightGBM/issues/2628#issuecomment-604519043.

Minor refactoring to pass new `check_sample_weights_not_an_array` test.

Reverted #2637. Fixed #2628. Closed #2946.

### DIFF
--- a/.ci/test.sh
+++ b/.ci/test.sh
@@ -73,7 +73,7 @@ if [[ $TASK == "r-package" ]]; then
     exit 0
 fi
 
-conda install -q -y -n $CONDA_ENV joblib matplotlib numpy pandas psutil pytest python-graphviz "scikit-learn<=0.21.3" scipy
+conda install -q -y -n $CONDA_ENV joblib matplotlib numpy pandas psutil pytest python-graphviz scikit-learn scipy
 
 if [[ $OS_NAME == "macos" ]] && [[ $COMPILER == "clang" ]]; then
     # fix "OMP: Error #15: Initializing libiomp5.dylib, but found libomp.dylib already initialized." (OpenMP library conflict due to conda's MKL)
--- a/.ci/test.sh
+++ b/.ci/test.sh
@@ -15,7 +15,7 @@ cd $BUILD_DIRECTORY
 
 if [[ $TRAVIS == "true" ]] && [[ $TASK == "check-docs" ]]; then
     cd $BUILD_DIRECTORY/docs
-    conda install -q -y -n $CONDA_ENV -c conda-forge doxygen
+    conda install -q -y -n $CONDA_ENV -c conda-forge doxygen libiconv
     pip install --user -r requirements.txt rstcheck
     # check reStructuredText formatting
     cd $BUILD_DIRECTORY/python-package
--- a/.ci/test.sh
+++ b/.ci/test.sh
@@ -15,7 +15,7 @@ cd $BUILD_DIRECTORY
 
 if [[ $TRAVIS == "true" ]] && [[ $TASK == "check-docs" ]]; then
     cd $BUILD_DIRECTORY/docs
-    conda install -q -y -n $CONDA_ENV -c conda-forge doxygen libiconv
+    conda install -q -y -n $CONDA_ENV -c conda-forge doxygen
     pip install --user -r requirements.txt rstcheck
     # check reStructuredText formatting
     cd $BUILD_DIRECTORY/python-package
--- a/python-package/setup.py
+++ b/python-package/setup.py
@@ -276,7 +276,7 @@ if __name__ == "__main__":
           install_requires=[
               'numpy',
               'scipy',
-              'scikit-learn<=0.21.3'
+              'scikit-learn'
           ],
           maintainer='Guolin Ke',
           maintainer_email='guolin.ke@microsoft.com',
--- a/python-package/setup.py
+++ b/python-package/setup.py
@@ -276,7 +276,7 @@ if __name__ == "__main__":
           install_requires=[
               'numpy',
               'scipy',
-              'scikit-learn'
+              'scikit-learn!=0.22.0'
           ],
           maintainer='Guolin Ke',
           maintainer_email='guolin.ke@microsoft.com',
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -30,7 +30,7 @@ install:
   - activate
   - conda config --set always_yes yes --set changeps1 no
   - conda update -q -y conda
-  - conda create -q -y -n test-env python=%PYTHON_VERSION% joblib matplotlib numpy pandas psutil pytest python-graphviz "scikit-learn<=0.21.3" scipy
+  - conda create -q -y -n test-env python=%PYTHON_VERSION% joblib matplotlib numpy pandas psutil pytest python-graphviz scikit-learn scipy
   - activate test-env
 
 build_script:
--- a/.vsts-ci.yml
+++ b/.vsts-ci.yml
@@ -132,7 +132,7 @@ jobs:
   - powershell: Write-Host "##vso[task.prependpath]$env:CONDA\Scripts"
     displayName: Enable conda
   - script: |
-      cmd /c 'activate & conda config --set always_yes yes --set changeps1 no & conda update -q -y conda & conda create -q -y -n %CONDA_ENV% python=%PYTHON_VERSION% joblib matplotlib numpy pandas psutil pytest python-graphviz "scikit-learn<=0.21.3" scipy'
+      cmd /c "activate & conda config --set always_yes yes --set changeps1 no & conda update -q -y conda & conda create -q -y -n %CONDA_ENV% python=%PYTHON_VERSION% joblib matplotlib numpy pandas psutil pytest python-graphviz scikit-learn scipy"
       cmd /c "activate %CONDA_ENV% & powershell -ExecutionPolicy Bypass -File %BUILD_SOURCESDIRECTORY%/.ci/test_windows.ps1"
     displayName: Test
   - task: PublishBuildArtifacts@1
--- a/docker/dockerfile-python
+++ b/docker/dockerfile-python
@@ -18,7 +18,7 @@ RUN apt-get update && \
     export PATH="$CONDA_DIR/bin:$PATH" && \
     conda config --set always_yes yes --set changeps1 no && \
     # lightgbm
-    conda install -q -y numpy scipy "scikit-learn<=0.21.3" pandas && \
+    conda install -q -y numpy scipy scikit-learn pandas && \
     git clone --recursive --branch stable --depth 1 https://github.com/Microsoft/LightGBM && \
     cd LightGBM/python-package && python setup.py install && \
     # clean
--- a/docker/gpu/dockerfile.gpu
+++ b/docker/gpu/dockerfile.gpu
@@ -75,8 +75,8 @@ RUN echo "export PATH=$CONDA_DIR/bin:"'$PATH' > /etc/profile.d/conda.sh && \
     rm ~/miniconda.sh
 
 RUN conda config --set always_yes yes --set changeps1 no && \
-    conda create -y -q -n py2 python=2.7 mkl numpy scipy "scikit-learn<=0.21.3" jupyter notebook ipython pandas matplotlib && \
-    conda create -y -q -n py3 python=3.6 mkl numpy scipy "scikit-learn<=0.21.3" jupyter notebook ipython pandas matplotlib
+    conda create -y -q -n py2 python=2.7 mkl numpy scipy scikit-learn jupyter notebook ipython pandas matplotlib && \
+    conda create -y -q -n py3 python=3.6 mkl numpy scipy scikit-learn jupyter notebook ipython pandas matplotlib
 
 #################################################################################################################
 #           LightGBM
--- a/docs/GPU-Tutorial.rst
+++ b/docs/GPU-Tutorial.rst
@@ -1,4 +1,4 @@
-﻿LightGBM GPU Tutorial
+LightGBM GPU Tutorial
 =====================
 
 The purpose of this document is to give you a quick step-by-step tutorial on GPU training.
@@ -78,7 +78,7 @@ If you want to use the Python interface of LightGBM, you can install it now (alo
 ::
 
     sudo apt-get -y install python-pip
-    sudo -H pip install setuptools numpy scipy "scikit-learn<=0.21.3" -U
+    sudo -H pip install setuptools numpy scipy scikit-learn -U
     cd python-package/
     sudo python setup.py install --precompile
     cd ..
--- a/docs/Python-API.rst
+++ b/docs/Python-API.rst
@@ -24,10 +24,6 @@ Training API
 Scikit-learn API
 ----------------
 
-.. warning::
-
-    The last supported version of scikit-learn is ``0.21.3``. Our estimators are incompatible with newer versions.
-
 .. autosummary::
     :toctree: pythonapi/
 
--- a/docs/Python-Intro.rst
+++ b/docs/Python-Intro.rst
@@ -15,11 +15,11 @@ Install
 -------
 
 Install Python-package dependencies,
-``setuptools``, ``wheel``, ``numpy`` and ``scipy`` are required, ``scikit-learn<=0.21.3`` is required for sklearn interface and recommended:
+``setuptools``, ``wheel``, ``numpy`` and ``scipy`` are required, ``scikit-learn`` is required for sklearn interface and recommended:
 
 ::
 
-    pip install setuptools wheel numpy scipy "scikit-learn<=0.21.3" -U
+    pip install setuptools wheel numpy scipy scikit-learn -U
 
 Refer to `Python-package`_ folder for the installation guide.
 
--- a/examples/python-guide/README.md
+++ b/examples/python-guide/README.md
@@ -8,7 +8,7 @@ You should install LightGBM [Python-package](https://github.com/microsoft/LightG
 You also need scikit-learn, pandas, matplotlib (only for plot example), and scipy (only for logistic regression example) to run the examples, but they are not required for the package itself. You can install them with pip:
 
 ```
-pip install "scikit-learn<=0.21.3" pandas matplotlib scipy -U
+pip install scikit-learn pandas matplotlib scipy -U
 ```
 
 Now you can run examples in this folder, for example:
--- a/python-package/lightgbm/compat.py
+++ b/python-package/lightgbm/compat.py
@@ -125,7 +125,6 @@ try:
         from sklearn.cross_validation import StratifiedKFold, GroupKFold
         from sklearn.utils.validation import NotFittedError
     SKLEARN_INSTALLED = True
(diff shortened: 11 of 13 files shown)


.ci/test.sh
```diff
@@ -15,7 +15,7 @@ cd $BUILD_DIRECTORY
 
 if [[ $TRAVIS == "true" ]] && [[ $TASK == "check-docs" ]]; then
     cd $BUILD_DIRECTORY/docs
-    conda install -q -y -n $CONDA_ENV -c conda-forge doxygen
+    conda install -q -y -n $CONDA_ENV -c conda-forge doxygen libiconv
```
StrikerRUS: To fix weird R lint error:
```
Linting R code
/home/travis/miniconda/envs/test-env/lib/R/bin/exec/R: error while loading shared libraries: libiconv.so.2: cannot open shared object file: No such file or directory
```

Hope this is temporary conda packaging error.

StrikerRUS:
@jameslamb Can you please help with failing R test on macOS?
```
Installing packages into ‘/Users/travis/Rlib’
(as ‘lib’ is unspecified)
Warning: unable to access index for repository https://cran.rstudio.com/bin/macosx/el-capitan/contrib/3.6:
  cannot open URL 'https://cran.rstudio.com/bin/macosx/el-capitan/contrib/3.6/PACKAGES'
   packages ‘data.table’, ‘jsonlite’, ‘Matrix’, ‘R6’, ‘testthat’ are available as source packages but not as binaries
Warning message:
packages ‘data.table’, ‘jsonlite’, ‘Matrix’, ‘R6’, ‘testthat’ are not available (as a binary package for R version 3.6.1) 
```

.ci/test.sh
StrikerRUS: Never mind. They fixed it. I'm reverting this temp fix.

StrikerRUS:
@jameslamb I tried to change CRAN repo 
`echo 'options(repos = "https://cran.rstudio.com")' > ${HOME}/.Rprofile` -> `echo 'options(repos = "https://cloud.r-project.org")' > ${HOME}/.Rprofile`

The same error:
```
Installing packages into ‘/Users/travis/Rlib’
(as ‘lib’ is unspecified)
Warning: unable to access index for repository https://cloud.r-project.org/bin/macosx/el-capitan/contrib/3.6:
  cannot open URL 'https://cloud.r-project.org/bin/macosx/el-capitan/contrib/3.6/PACKAGES'
   packages ‘data.table’, ‘jsonlite’, ‘Matrix’, ‘R6’, ‘testthat’ are available as source packages but not as binaries
Warning message:
packages ‘data.table’, ‘jsonlite’, ‘Matrix’, ‘R6’, ‘testthat’ are not available (as a binary package for R version 3.6.1) 
```

jameslamb:
> @jameslamb I tried to change CRAN repo
> `echo 'options(repos = "https://cran.rstudio.com")' > ${HOME}/.Rprofile` -> `echo 'options(repos = "https://cloud.r-project.org")' > ${HOME}/.Rprofile`
> 
> The same error:
> 
> ```
> Installing packages into ‘/Users/travis/Rlib’
> (as ‘lib’ is unspecified)
> Warning: unable to access index for repository https://cloud.r-project.org/bin/macosx/el-capitan/contrib/3.6:
>   cannot open URL 'https://cloud.r-project.org/bin/macosx/el-capitan/contrib/3.6/PACKAGES'
>    packages ‘data.table’, ‘jsonlite’, ‘Matrix’, ‘R6’, ‘testthat’ are available as source packages but not as binaries
> Warning message:
> packages ‘data.table’, ‘jsonlite’, ‘Matrix’, ‘R6’, ‘testthat’ are not available (as a binary package for R version 3.6.1) 
> ```

looking right now

jameslamb:
If you rebase to `master`, to get the changes  from #2954 , the R test issues should be resolved.

StrikerRUS:
@henry0312 Please take a look when have time

henry0312:
Sure, I'm going to check tomorrow

python-package/setup.py
```diff
@@ -276,7 +276,7 @@ def run(self):
           install_requires=[
               'numpy',
               'scipy',
-              'scikit-learn<=0.21.3'
+              'scikit-learn!=0.22.0'
```
henry0312, line 279: is this `'scikit-learn~=0.22,!=0.22.0'`?

StrikerRUS: Sorry, I can't find `~` in the `setuptools` docs. Can you please clarify what do you mean?
> A version specifier is one of the operators <, >, <=, >=, == or !=, followed by a version identifier.
    https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-dependencies

henry0312: Version specifiers
https://www.python.org/dev/peps/pep-0440/#version-specifiers

StrikerRUS: OK. I guess that this specifier is not compatible with `setuptools` as it is not listed there, is it?

Also, why do we need it? We are compatible with all (reasonable) versions except `0.22.0`.

StrikerRUS: ping @henry0312

henry0312: > OK. I guess that this specifier is not compatible with setuptools as it is not listed there, is it?

I don't know the specific version but `setuptools` used in sklearn, numpy, scipy and pandas supports version specifier as least.

> Also, why do we need it? We are compatible with all (reasonable) versions except 0.22.0.

Not enough.
As you know, many python packages follow [Semantic Versioning](https://semver.org/), including scikit-learn.

If we specify `scikit-learn!=0.22.0` and v1.0.0 of scikit-learn is released, LightGBM will allow users to use v1.0.0, which can lead to break backward compatibility.
Therefore, if we follow [Semantic Versioning](https://semver.org/), we should specify libraries' version by Version specifiers.

StrikerRUS: > If we specify scikit-learn!=0.22.0 and v1.0.0 of scikit-learn is released, LightGBM will allow users to use v1.0.0,

I don't see any problems with it. Moreover, I believe that this should be an expected behavior.
Before now we have never blocked **future** versions of libraries, assuming that by default we are compatible with them. If it is not true, someone comes and posts an issue and we are investigating.
In case of prohibiting future releases it is highly likely that we'll forget to update those requirements or will do it with huge delay making user experience inconvenient.

henry0312: > Before now we have never blocked future versions of libraries, assuming that by default we are compatible with them. 

When that is our policy, as you say, there is no problem and I'll approve this PR, though I don't think the policy is common in software engineering.
In other words, a new version which can breaks backward compatibility in terms of Semantic Versioning is released, we should check its compatibility and will add suuport of it in series.

henry0312:
By the way, why is there the same (similar?) PR, https://github.com/microsoft/LightGBM/pull/2946?

python-package/setup.py
henry0312: So I'll approve this when our support policy is clear.

henry0312: In fact, if we specified Sphinx version, https://github.com/microsoft/LightGBM/commit/ce08242567e1e37e1f947038a78ac671c98ed18e#diff-e0c7b285f2f2d8028247b8cfd4dc76fcR1 were not needed, though that are just tests.

I think unless we follow the basic rules of software engineering, the quality of library won't increase so much even If we add more and more tests.

So I'll approve this when our support policy is clear.

ekerazha: @henry0312 Actually it's pretty common in software engineering to allow for future versions of dependencies in **libraries** (while it's common to be more strict in end-user products). If a dependency follows Semantic Versioning (SemVer), you could put a restriction on major versions (e.g. scikit-learn<1.0), but I don't think sklearn is following SemVer. Unless you're ready to release a new version of LightGBM every time a new minor version or patch for a dependency is released, of course.

StrikerRUS: @henry0312 
> though I don't think the policy is common in software engineering.

It's kind of debatable statements. And I don't think that this PR is a good place to discuss it. I think we should create a separate issue where we should at least get opinions of other maintainers before changing the current policy.

> we should check its compatibility and will add suuport of it in series.

In ideal situation, yes! You are absolutely right! But given that we are lacking human resources to do it in time, I suppose that it will simply abandoned in the near future. Which will result in very old dependencies of LightGBM.

StrikerRUS: > Unless you're ready to release a new version of LightGBM every time a new minor version or patch for a dependency is released, of course.

Absolutely agree! It is definitely not our case when only 2-3 maintainers work actively at the same time.

StrikerRUS:
@henry0312 
> By the way, why is there the same (similar?) PR, #2946?

That PR only reverts version constraint. This one fixes some new tests in addition. Refer to https://github.com/microsoft/LightGBM/pull/2946#issuecomment-604550363.

python-package/setup.py
henry0312: @ekerazha @StrikerRUS I don't mean all of the future versions wont' be allowed, but only the future versions that can breaks backward compatibility (in other words, a release changes major version), which version specifiers (i.e. `scikit-learn~=0.22,!=0.22.0` or `scikit-learn>=0.22, == 0.*`) realizes.
And is LightGBM an almost end-user products? I think so.

> I don't think sklearn is following SemVer.

oh, really...? Actually, I'm not sure.

henry0312:
> @henry0312
> 
> > By the way, why is there the same (similar?) PR, #2946?
> 
> That PR only reverts version constraint. This one fixes some new tests in addition. Refer to [#2946 (comment)](https://github.com/microsoft/LightGBM/pull/2946#issuecomment-604550363).

Thank you for your explanation. I see 😄

python-package/setup.py
ekerazha: @henry0312 
> And is LightGBM an almost end-user products? I think so.
LightGBM is a library, a travel website which uses LightGBM to recommend travels would be the end-user product.

> oh, really...? Actually, I'm not sure.
scikit-learn had breaking API changes in minor versions, so it's not following SemVer.

StrikerRUS: @henry0312 I guess `0.22` is a **minor** release, but it broke  backward compatibility.

Again, please let's move all our discussion to a separate issue where everyone will be able to comment. It's very hard to find it here and I want to merge this PR ASAP to enable the latest scikit-learn support.

guolinke:
@StrikerRUS @henry0312 is this ready to merge?

henry0312:
@guolinke @StrikerRUS yes it is.

Issue #2790, opened 2020-02-21
[Discussion] accuracy improvements


guolinke:
This is to call the accuracy improvement for the tree-based algorithms, include but not limited to:
- Better regularization methods that improve generalization abilities.
- Better base tree learner, like piece-wise linear trees, to improve the effectiveness of base learner
- Better objective functions for special tasks/metrics

If you have any ideas, you can discuss with us here, and open the corresponding issues/pull requests if needed.

MaxHalford:
There's an idea I've encoutered in the litterature that seems promising, it's called **path smoothing**.

Trees overfit because they are are too deep and thus their leaves have too few observations in them. When a leaf has too few observations, then the distribution of the target in said leaf is biased and unreliable. The common way to solve is to restrict the maximal depth of the tree, or to apply pruning.

Another less common way is to do path smoothing: calculate a weighted average with respect to each node from root to the leaf for a given observation. Each node thus contributes to the weighted average. There are different weighting schemes that are possible, they just have to capture the fact that the contribution should increase with the number of observations and the depth. What's nice is that this way, even if a leaf has few observations, it's parents will correct the target distribution by smoothing it out. You can think of it as a Bayesian or hierarchical prior.

btrotta:
@MaxHalford sounds interesting, do you have a reference for this?

MaxHalford:
Here are some papers which I have managed to find:

- [Obtaining calibrated probability estimates from decision trees and naive Bayesian classifiers](https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=2&cad=rja&uact=8&ved=2ahUKEwjWkNOBzZnoAhWHxoUKHaahAI4QFjABegQIBhAB&url=http%3A%2F%2Fcseweb.ucsd.edu%2F~elkan%2Fcalibrated.pdf&usg=AOvVaw0geu_-DpVldsDa3Lg8s8B8): in subsection 2.3 they introduce *curtailment*, which involves using the first leaf along the path which contains at least `k` observations. It's therefore very similar to pruning and isn't worth exploring. It's of use to me because I'm researching decision trees for online learning, and this simple heuristic helps a ton.
- [Learning Classification Trees](https://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/19920016875.pdf) discusses path smoothing (they call it something else) but the details weren't very clear to me.
[...]

PR #2950, opened 2020-03-27
Path smoothing

btrotta:
This implements the path smoothing idea proposed by @MaxHalford in #2790.

I implemented a slightly simpler version which does not use leaf depth. For each node (except the root), the output is calculated as `((n / s) * original_output + parent_output) / (n / s + 1)` where `n` is number of data in leaf, `s` is a regularisation parameter (larger `s` means more smoothing), `original_output` is the unsmoothed output, and `parent_output` is the output of the parent node (which has itself been smoothed previously).

The reasoning behind this is that it's similar to the Bayesian calculation of the posterior mean given a prior expectation (see, e.g. Section 2 here: http://www.ams.sunysb.edu/~zhu/ams570/Bayesian_Normal.pdf). The posterior estimate of the true mean (in our case, the leaf output) is the weighted average (weighted by sample size and the sample and prior variance) of the sample and prior mean (the parent output). (This is more of an analogy than a real model, since the leaf output isn't actually the mean of a sample.) The parameter `s` represents the expected ratio of child node variance to parent node variance; small `s` means we expect the ratio to be small, i.e. the data is not noisy and the model can explain a lot of the variance, so we don't need much regularisation.

I haven't tested this much on real datasets. Based on the couple that I tried, it definitely improves regularisation, but doesn't seem to be dramatically different to existing regularisation methods. Given that it requires non-trivial code changes, it may not be worth the maintenance effort. Feel free to close this PR if so.


btrotta:
Below is a basic script to test on the Boston housing data and the results. The script tries several different regularisation parameters (individually, not in combination) and selects the best value for each, then plots the results.

```
import pandas as pd
import numpy as np
import lightgbm as lgb
import os
from sklearn import datasets

X, y = datasets.load_boston(True)
df = pd.DataFrame(np.concatenate([X, y[:, np.newaxis]], axis=1))
train_cols = df.columns[1:]
np.random.seed(0)
train_ind = np.random.choice(df.index, len(df) // 2, replace=False)
train_bool = df.index.isin(train_ind)
lgb_train = lgb.Dataset(df.loc[train_bool, train_cols], label=df.loc[train_bool, 0])
lgb_test = lgb.Dataset(df.loc[~train_bool, train_cols], label=df.loc[~train_bool, 0])

# base
params = {'objective': 'regression_l2', 'seed': 0, 'num_leaves': 32, 'learning_rate': 0.01, 'metric': 'rmse',
          'lambda_l1': 0, 'lambda_l2': 0, 'min_data_in_leaf': 2, 'path_smooth': 0, 'subsample': 0.8,
          'feature_pre_filter': False}

# find best regularisation parameter
reg_param_names = ['min_data_in_leaf', 'path_smooth', 'lambda_l1', 'lambda_l2']
best_value = {p: None for p in reg_param_names}
best_res = {p: None for p in reg_param_names}
param_range = {p: [0, 1, 2, 4, 8, 16] for p in reg_param_names}
param_range['min_data_in_leaf'] = [2, 4, 8, 16]
for reg_param in reg_param_names:
    best_test_loss = np.inf
    for param_value in param_range[reg_param]:
        params[reg_param] = param_value
        res = {}
        est = lgb.train(params, lgb_train, valid_sets=valid_sets, valid_names=valid_names,
                        num_boost_round=100, evals_result=res)
        test_loss = res['valid']['rmse'][-1]
        if test_loss < best_test_loss:
            best_test_loss = test_loss
            best_value[reg_param] = param_value
            best_res[reg_param] = res
    params[reg_param] = 0
    if reg_param == 'min_data_in_leaf':
        params[reg_param] = 2


plt.figure()
color_list = ['b', 'y', 'r', 'c', 'm']
legend_list = []
for i, reg_param in enumerate(reg_param_names):
    res = best_res[reg_param]
    plt.plot(res['train']['rmse'], color=color_list[i])
    plt.plot(res['valid']['rmse'], color=color_list[i], linestyle=':')
    legend_list.append('{} = {} train'.format(reg_param, best_value[reg_param]))
    legend_list.append('{} = {} test'.format(reg_param, best_value[reg_param]))

# no regularisation
res = {}
params = {'objective': 'regression_l2', 'seed': 0, 'num_leaves': 32, 'learning_rate': 0.01, 'metric': 'rmse',
          'lambda_l1': 0, 'lambda_l2': 0, 'min_data_in_leaf': 2, 'path_smooth': 0, 'subsample': 0.8,
          'feature_pre_filter': False}
est = lgb.train(params, lgb_train, valid_sets=valid_sets, valid_names=valid_names,
                num_boost_round=100, evals_result=res)
i += 1
plt.plot(res['train']['rmse'], color=color_list[i])
plt.plot(res['valid']['rmse'], color=color_list[i], linestyle=':')
legend_list.append('no reg train')
legend_list.append('no reg test')
plt.legend(legend_list)
```
Figure_1

btrotta:
I think I fixed the problem with the gpu version but CI is still failing, because of a problem with the R package. Not sure if this is due to my code changes or not...

jameslamb:
> I think I fixed the problem with the gpu version but CI is still failing, because of a problem with the R package. Not sure if this is due to my code changes or not...

@btrotta If you rebase to `master,` to get the changes from #2954 , the R test issues and intermittent linting issues should be resolved.

Sorry for the inconvenience!

StrikerRUS:
@jameslamb We need your help here:
```
  installed size is  5.1Mb
    libs   4.5Mb
```

jameslamb:
> @jameslamb We need your help here:
> 
> ```
> * checking installed package size ... NOTE
>   installed size is  5.1Mb
>   sub-directories of 1Mb or more:
>     libs   4.5Mb
> ```

ah! I checked the diff and this note is not a problem for now. This is mainly there to keep the hosting costs for CRAN down, so that people don't (for example) check large files with datasets into their packages. But the check is interesting because it's checked on the _installed_ package, so in our case `libs/` doesn't even exist in our source but is created to contain `lib_lightgbm.so` / `lib_lightgbm.dll` at install time.

I want to investigate this more, but it's not a problem caused by this PR. The NOTE is triggered by an installed package over 5.0 MB and I think we were juuuust under that on Mac and linux before. This is one of those weird NOTEs that CRAN will sometimes ignore if you explain it to them.

For now @btrotta , just change this line to 4 allowed notes: https://github.com/microsoft/LightGBM/blob/master/.ci/test_r_package.sh#L94.

This doesn't need to block this PR and I can address it with a more long-term answer in a separate PR. For what it's worth, I see it on #2936 s well.

btrotta:
@jameslamb thanks, that fixed it! It's now failing because of a broken link in the docs (not related to this PR).

jameslamb:
> @jameslamb thanks, that fixed it! It's now failing because of a broken link in the docs (not related to this PR).

Ha sorry, we have a few annoying things happening in our CI this week. The `check-docs` tasks failing is being fixed in #2956. Thanks for being patient with us.

jameslamb:
> > @jameslamb thanks, that fixed it! It's now failing because of a broken link in the docs (not related to this PR).
> 
> Ha sorry, we have a few annoying things happening in our CI this week. The `check-docs` tasks failing is being fixed in #2956. Thanks for being patient with us.

@btrotta as soon as I wrote that, I went to review that PR and saw that it could be merged! If you rebase to `master` again to get those changes, your PR should pass.

include/LightGBM/config.h
```diff
@@ -488,6 +488,15 @@ struct Config {
...
   std::vector<double> cegb_penalty_feature_coupled;
 
+  // check = >= 0.0
+  // desc = controls smoothing applied to tree nodes
+  // desc = helps prevent overfitting on leaves with few samples
+  // desc = if set to zero, no smoothing is applied
+  // desc = if `path_smooth > 0` then `min_data_in_leaf` must be at least `2`.
+  // desc = larger values give stronger regularisation
+  // descl2 = the weight of each node is `(n / path_smooth) * w + w_p / (n / path_smooth + 1)`, where `n` is the number of samples in the node, `w` is the calculated node weight, and `w_p` is the weight of the parent node
+  double path_smooth = 0;
```
guolinke, line 498: more details about `w` and `w_p`?

src/treelearner/feature_histogram.hpp
```diff
@@ -670,35 +771,42 @@ class FeatureHistogram {
...
-      return ret;
     } else {
-      double ret = -sum_gradients / (sum_hessians + l2);
+      ret = -sum_gradients / (sum_hessians + l2);
       if (USE_MAX_OUTPUT) {
         if (max_delta_step > 0 && std::fabs(ret) > max_delta_step) {
-          return Common::Sign(ret) * max_delta_step;
+          ret = Common::Sign(ret) * max_delta_step;
         }
       }
```
guolinke, line 794: it seems this could be moved outside too.

src/treelearner/feature_histogram.hpp
```diff
@@ -85,22 +85,29 @@ class FeatureHistogram {
...
                         SplitInfo* output, int* rand_threshold) {
     is_splittable_ = false;
     output->monotone_type = meta_->monotone_type;
-    double gain_shift = GetLeafGain<USE_L1, USE_MAX_OUTPUT>(
-        sum_gradient, sum_hessian, meta_->config->lambda_l1,
-        meta_->config->lambda_l2, meta_->config->max_delta_step);
+    double gain_shift;
+    if (USE_SMOOTHING) {
+      gain_shift = GetLeafGainGivenOutput<USE_L1>(
+          sum_gradient, sum_hessian, meta_->config->lambda_l1, meta_->config->lambda_l2, parent_output);
```
guolinke, line 105: @btrotta could you double confirm here? if this is correct, could you add some detailed comments to make it clear?

src/treelearner/feature_histogram.hpp
```diff
@@ -670,35 +771,42 @@ class FeatureHistogram {
...
       if (USE_MAX_OUTPUT) {
         if (max_delta_step > 0 && std::fabs(ret) > max_delta_step) {
-          return Common::Sign(ret) * max_delta_step;
+          ret = Common::Sign(ret) * max_delta_step;
         }
       }
-      return ret;
     }
+    if (USE_SMOOTHING) {
+      ret = ret * (num_data / smoothing) / (num_data / smoothing + 1) \
```
guolinke, line 797: maybe only apply this when `smoothing > 0` ?

src/treelearner/feature_histogram.hpp
```diff
@@ -609,29 +674,65 @@ class FeatureHistogram {
...
-    double current_gain =
-        GetLeafGain<true, true>(sum_right_gradient, sum_right_hessian,
-                                meta_->config->lambda_l1, l2,
-                                meta_->config->max_delta_step) +
-        GetLeafGain<true, true>(sum_left_gradient, sum_left_hessian,
-                                meta_->config->lambda_l1, l2,
-                                meta_->config->max_delta_step);
+    double current_gain;
+    bool use_smoothing = meta_->config->path_smooth > kEpsilon;
+    if (use_smoothing) {
```
guolinke, line 679: for these `GatherInfoxxx` functions, maybe you can use `GetLeafGain<true, true, true>`, and check `path_smooth > kEpsilon' inside?

guolinke:
Thanks @btrotta !
I feel like that these template expansions in FeatureHistogram introduced by me are hard for development and review. 
I will have another refactoring for them when have time.

src/treelearner/feature_histogram.hpp
btrotta: You're right, this was not quite correct. Have made a slight change, let me know if it's still unclear.

include/LightGBM/config.h
btrotta: Added some more explanation, hope it's clearer now.

src/treelearner/feature_histogram.hpp
btrotta: I have done the check for `smoothing > kEpsilon` outside this function (e.g. in `FuncForNumricalL2` and `FuncForCategoricalL1`) , and I only pass the template parameter `USE_SMOOTHING=true` if `smoothing > kEpsilon`. Alternatively we could move the check inside `CalculateSplittedLeafOutput`, and then we wouldn't need the template parameter `USE_SMOOTHING`. This would be simpler but we would lose the speedup that the template provides.

src/treelearner/feature_histogram.hpp
btrotta: See comment above.

src/treelearner/feature_histogram.hpp
```diff
@@ -609,29 +670,65 @@ class FeatureHistogram {
...
```
stanpcf, line 675: there are lot of `if(use_smoothing)` in code.
```
if(use_smoothing){
    func<true, true, true>();
}else {
    func<true, true, false>();
}
```
it's better use `func<true, true, use_smoothing>();` instead

btrotta: This gives a compiler error because the value of `use_smoothing` isn't known at compile time.

src/treelearner/feature_histogram.hpp
```diff
@@ -85,22 +85,25 @@ class FeatureHistogram {
...
                         SplitInfo* output, int* rand_threshold) {
     is_splittable_ = false;
     output->monotone_type = meta_->monotone_type;
-    double gain_shift = GetLeafGain<USE_L1, USE_MAX_OUTPUT>(
-        sum_gradient, sum_hessian, meta_->config->lambda_l1,
-        meta_->config->lambda_l2, meta_->config->max_delta_step);
+    double output_without_split = CalculateSplittedLeafOutput<USE_L1, USE_MAX_OUTPUT, USE_SMOOTHING>(
+        sum_gradient, sum_hessian, meta_->config->lambda_l1, meta_->config->lambda_l2,
+        meta_->config->max_delta_step, meta_->config->path_smooth, num_data, parent_output);
+    double gain_shift = GetLeafGainGivenOutput<USE_L1>(
```
guolinke, line 105: may be a new function that `CalculateSplittedLeafOutput` then `GetLeafGainGivenOutput` ? it seems some other places also use this.

src/treelearner/feature_histogram.hpp
guolinke: okay, maybe we can make `GatherInfoxxx` to template function too.  For example, 'template<bool USE_SMOOTHING> GatherInfoForThresholdNumericalInner'. And check the use_smoothing in GatherInfoForThresholdNumerical. By this way, we can reduce these if ... else.

src/treelearner/feature_histogram.hpp
btrotta: Actually I realised we can just use the existing function `GetLeafGain` here.

src/treelearner/feature_histogram.hpp
btrotta: Changed as suggested.

src/treelearner/feature_histogram.hpp
stanpcf: `USE_SMOOTHING` may be useless for this function. the soothing output can be simplified as below。
weight for current output is `w = n/(n+s)` and `ret = w * ret + (1-w) * parent_out`. `n` must be greater zero. `s=0` means no path smoothing.

StrikerRUS:
@jameslamb please take a look
```
Linting R code
Error: 'absolute_path_linter' is not an exported object from 'namespace:lintr'
Execution halted
```

StrikerRUS:
@guolinke Can you please check your DM in Slack?

StrikerRUS:
@jameslamb Created #2986 to not pollute this PR.

guolinke:
@StrikerRUS Thanks!
@btrotta Thanks very much for your many contributions, would you like to join LightGBM project as a collaborator?

btrotta:
@guolinke Yes that would be great! Thanks!

guolinke:
it this ready to merge?

btrotta:
@guolinke The CI is still failing, I'm not sure how to fix it.

jameslamb:
> @guolinke The CI is still failing, I'm not sure how to fix it.

@btrotta can you please merge in `master` to this PR? We have had several PRs to fix CI issues in the last few weeks.

btrotta:
@jameslamb Thanks, that has fixed the problem from before. I also fixed some linting errors. But now I'm getting a new failure which wasn't happening previously:
```
.ci/test.sh: line 119: pytest: command not found
The command "bash .ci/test.sh" exited with 255.
```

StrikerRUS:
> But now I'm getting a new failure which wasn't happening previously:

It was a random network conection issue during conda installation:
```
('Connection broken: OSError("(104, \'ECONNRESET\')")', OSError("(104, 'ECONNRESET')"))
```
I re-run that job and now everything is green.

.ci/test_r_package.sh
```diff
@@ -98,7 +98,7 @@ if grep -q -R "WARNING" "$LOG_FILE_NAME"; then
     exit -1
 fi
 
-ALLOWED_CHECK_NOTES=2
+ALLOWED_CHECK_NOTES=4
```
StrikerRUS, line 101: @jameslamb Should be 3, right?

docs/Parameters-Tuning.rst
```diff
@@ -82,3 +82,5 @@ Deal with Over-fitting
 -  Try ``extra_trees``
 
 .. _Optuna: https://medium.com/optuna/lightgbm-tuner-new-optuna-integration-for-hyperparameter-optimization-8b7095e99258
+
+-  Try increasing ``path_smoothing``
```
StrikerRUS, line 86: Please move before `.. _Optuna`.

include/LightGBM/config.h
```diff
@@ -495,6 +495,16 @@ struct Config {
   // desc = applied once per forest
   std::vector<double> cegb_penalty_feature_coupled;
 
+  // check = >= 0.0
+  // desc = controls smoothing applied to tree nodes
+  // desc = helps prevent overfitting on leaves with few samples
+  // desc = if set to zero, no smoothing is applied
+  // desc = if `path_smooth > 0` then `min_data_in_leaf` must be at least `2`.
```
StrikerRUS, line 502:
```suggestion
  // desc = if ``path_smooth > 0`` then ``min_data_in_leaf`` must be at least ``2``.
```

include/LightGBM/config.h
```diff
@@ -495,6 +495,16 @@ struct Config {
...
+  // desc = larger values give stronger regularisation
+  // descl2 = the weight of each node is `(n / path_smooth) * w + w_p / (n / path_smooth + 1)`, where `n` is the number of samples in the node, `w` is the optimal node weight to minimise the loss (approximately `-sum_gradients / sum_hessians`), and `w_p` is the weight of the parent node
```
StrikerRUS, line 504:
```suggestion
  // descl2 = the weight of each node is ``(n / path_smooth) * w + w_p / (n / path_smooth + 1)``, where ``n`` is the number of samples in the node, ``w`` is the optimal node weight to minimise the loss (approximately ``-sum_gradients / sum_hessians``), and ``w_p`` is the weight of the parent node
```

include/LightGBM/config.h
```diff
@@ -495,6 +495,16 @@ struct Config {
...
+  // descl2 = (note that the parent output `w_p` itself has smoothing applied, unless it is the root node, so that the smoothing effect accumulates with the tree depth)
```
StrikerRUS, line 505:
```suggestion
  // descl2 = note that the parent output ``w_p`` itself has smoothing applied, unless it is the root node, so that the smoothing effect accumulates with the tree depth
```

src/treelearner/feature_histogram.hpp
```diff
@@ -85,22 +85,24 @@ class FeatureHistogram {
...
                         SplitInfo* output, int* rand_threshold) {
     is_splittable_ = false;
     output->monotone_type = meta_->monotone_type;
-    double gain_shift = GetLeafGain<USE_L1, USE_MAX_OUTPUT>(
-        sum_gradient, sum_hessian, meta_->config->lambda_l1,
-        meta_->config->lambda_l2, meta_->config->max_delta_step);
+
+    double gain_shift = GetLeafGain<USE_L1, USE_MAX_OUTPUT, USE_SMOOTHING>(
+        sum_gradient, sum_hessian, meta_->config->lambda_l1, meta_->config->lambda_l2,
+      meta_->config->max_delta_step, meta_->config->path_smooth, num_data, parent_output);
```
StrikerRUS, line 105:
```suggestion
        meta_->config->max_delta_step, meta_->config->path_smooth, num_data, parent_output);
```

.ci/test_r_package.sh
jameslamb: yes  @btrotta can you please change to 3?

A lot has changed in our CI since this PR was first opened, sorry 😬

jameslamb:
```suggestion
ALLOWED_CHECK_NOTES=3
```

StrikerRUS: @jameslamb Is a similar increment needed for Windows?
https://github.com/microsoft/LightGBM/blob/2c18a0f3ed732df77d4304ed1d2b46a345d661c8/.ci/test_r_package_windows.ps1#L102

btrotta: No worries, fixed now.

btrotta:
@StrikerRUS thanks for your comments, I've fixed those issues now.

docs/Parameters-Tuning.rst
```diff
@@ -81,4 +81,6 @@ Deal with Over-fitting
 
 -  Try ``extra_trees``
 
+-  Try increasing ``path_smoothing``
```
StrikerRUS, line 84: We should either add `path_smoothing` into param's aliases or fix name here to ``path_smooth`. Otherwise users will not be able to find this term anywhere in our docs.

docs/Parameters.rst
```diff
@@ -522,6 +522,22 @@ Learning Control Parameters
...
 
+-  ``path_smooth`` :raw-html:`<a id="path_smooth" title="Permalink to this parameter" href="#path_smooth">&#x1F517;&#xFE0E;</a>`, default = ``0``, type = double, constraints: ``path_smooth >=  0.0``
+
+   -  controls smoothing applied to tree nodes
+
+   -  helps prevent overfitting on leaves with few samples
+
+   -  if set to zero, no smoothing is applied
+
+   -  if ``path_smooth > 0`` then ``min_data_in_leaf`` must be at least ``2``.
```
StrikerRUS, line 533:
```suggestion
   -  if ``path_smooth > 0`` then ``min_data_in_leaf`` must be at least ``2``
```

include/LightGBM/config.h
```diff
@@ -495,6 +495,16 @@ struct Config {
   // desc = applied once per forest
   std::vector<double> cegb_penalty_feature_coupled;
 
+  // check = >= 0.0
+  // desc = controls smoothing applied to tree nodes
+  // desc = helps prevent overfitting on leaves with few samples
+  // desc = if set to zero, no smoothing is applied
+  // desc = if ``path_smooth > 0`` then ``min_data_in_leaf`` must be at least ``2``.
```
StrikerRUS, line 502:
```suggestion
  // desc = if ``path_smooth > 0`` then ``min_data_in_leaf`` must be at least ``2``
```

.ci/test_r_package.sh
StrikerRUS: ping @jameslamb for Windows updates

jameslamb: @StrikerRUS nothing is required on this PR since [the Windows R tests are passing](https://ci.appveyor.com/project/guolinke/lightgbm/builds/32470697/job/lkhm0f63w8titj1w).  We have it set to 3  and are generating exactly  3:

image

StrikerRUS: @jameslamb OK. I just was afraid that we can hit https://github.com/microsoft/LightGBM/pull/2950#issuecomment-605696533 on Windows if let say the current size is 4.99 during some runs it can be calculated as 5, like in #2988.

jameslamb: If we do we can bump it back up and it wouldn't bother me. I just don't think this PR has to be the place where we deal with it. If CI is passing for the R package right now, nothing should be changed.

PR #2956, opened 2020-03-29
[ci][docs] fixed broken link

StrikerRUS:
Old site with free pdf is down now and seems will not be recovered in the future.

image


jameslamb:
I'm able to reproduce the 403 showing up in the task from my laptop

image

The only option I've found that returns a 200 is the Google Scholar result (https://scholar.google.com/scholar?q=%22On+Grouping+for+Maximum+Homogeneity%22) :/

StrikerRUS:
Hmm, 403, really?!

OK...

jameslamb:
Thanks @StrikerRUS ! That link returns a `302`, but I guess linkchecker must follow redirects

image

Issue #2441, opened 2019-09-24
[R-package] use commandArgs instead of hardcoded stuff in the installation script


StrikerRUS:
Today, installing the R package with CMake like this:

```shell
Rscript build_r.R
```

Might require you to manually change some flags in `R-package/install.libs.R`.

https://github.com/microsoft/LightGBM/blob/e77658205d3b1bc5c19e4ba3821f4222f5be8c05/R-package/src/install.libs.R#L1-L4

Refer to https://github.com/microsoft/LightGBM/pull/2164#issuecomment-491576282.

To close this issue, expose those flags as command-line flags so users can do something like this:

```shell
Rscript build_r.R --use-gpu
```

StrikerRUS:
Closed in favor of being in #2302. We decided to keep all feature requests in one place.

Welcome to contribute this feature! Please re-open this issue (or post a comment if you are not a topic starter) if you are actively working on implementing this feature.

jameslamb:
if you're reading this and interested in contributing this feature, please comment here. Feel free to `@` me with questions

PR #2957, opened 2020-03-30
[R-package] [ci] Add option to skip installation in build_r.R

jameslamb:
Right now, `Rscript build_r.R` will build a source distribution of the R package and install it. In this PR, I propose an addition to the script to tell `build_r.R` to just build the package tarball but not install it.

This should allow us to cut a few minutes out of the `r-package` tasks in CI!

Right now we are doing this:

```
Rscript build_r.R
R CMD check lightgbm_*.tar.gz
```

That means we're building `lib_lightgbm` twice, without getting any extra testing benefit. In fact, I think that the scenario of checking a source tarball on a system where `lightgbm` hasn't been installed already is an even better test of the experience for our users!

Note that this PR won't change the existing behavior for anyone using  `build_r.R`. The default will still be to both build and install.

### DIFF
--- a/.ci/test_r_package.sh
+++ b/.ci/test_r_package.sh
@@ -72,7 +72,7 @@ fi
 Rscript --vanilla -e "install.packages(${packages}, repos = '${CRAN_MIRROR}', lib = '${R_LIB_PATH}')" || exit -1
 
 cd ${BUILD_DIRECTORY}
-Rscript build_r.R || exit -1
+Rscript build_r.R --skip-install || exit -1
 
 PKG_TARBALL="lightgbm_${LGB_VER}.tar.gz"
 LOG_FILE_NAME="lightgbm.Rcheck/00check.log"
--- a/.ci/test_r_package.sh
+++ b/.ci/test_r_package.sh
@@ -69,7 +69,7 @@ packages="c('data.table', 'jsonlite', 'Matrix', 'R6', 'testthat')"
 if [[ $OS_NAME == "macos" ]]; then
     packages+=", type = 'binary'"
 fi
-Rscript --vanilla -e "install.packages(${packages}, repos = '${CRAN_MIRROR}', lib = '${R_LIB_PATH}')" || exit -1
+Rscript --vanilla -e "install.packages(${packages}, repos = '${CRAN_MIRROR}', lib = '${R_LIB_PATH}', dependencies = c('Depends', 'Imports', 'LinkingTo'))" || exit -1
 
 cd ${BUILD_DIRECTORY}
 Rscript build_r.R --skip-install || exit -1
--- a/build_r.R
+++ b/build_r.R
@@ -5,6 +5,9 @@
 # Sys.setenv("CXX" = "/usr/local/bin/g++-8")
 # Sys.setenv("CC" = "/usr/local/bin/gcc-8")
 
+args <- commandArgs(trailingOnly=TRUE)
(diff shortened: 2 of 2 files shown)


StrikerRUS:
Linking #2441 for example of `commandArgs` usage.

build_r.R
```diff
@@ -86,4 +89,8 @@ version <- gsub(
 tarball <- file.path(getwd(), sprintf("lightgbm_%s.tar.gz", version))
 
 cmd <- sprintf("R CMD INSTALL %s --no-multiarch --with-keep.source", tarball)
-.run_shell_command(cmd)
+if (INSTALL_AFTER_BUILD) {
+  .run_shell_command(cmd)
+} else  {
```
StrikerRUS:
```suggestion
} else {
```

jameslamb:
Seeing this on the [linux CI task in Azure DevOps](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=5678&view=logs&j=a0acac97-df52-5651-dae8-9fb37a4728fc&t=d2c32b51-981b-5bcd-fe31-029dd8d8998c) and I am very confused

image

I don't understand how this could be failing on Azure DevOps but not Travis 🤔 

My best guess is that some very new version of one of our dependencies previously relied on `Matrix` and just factored it out, and maybe we got a slightly newer version on Linux because we get source packages there and binary packages on Mac.

Either way, we probably should have been installing  `Matrix` all along since it is listed in `Imports`, so I'll just add it.

jameslamb:
This PR is ready to merge, but I think we're facing a bug Travis faced a few days ago and thought they'd fixed.

Successful build: https://travis-ci.org/github/microsoft/LightGBM/builds/669022302

Details on the issue from https://www.traviscistatus.com/:

image

jameslamb:
close-reopen to retrigger CI to try to get past transient issues

jameslamb:
This seems to be working! Going to merge it so we can start getting the benefits in our CI 😀

.ci/test_r_package.sh
```diff
@@ -69,10 +69,10 @@ packages="c('data.table', 'jsonlite', 'Matrix', 'R6', 'testthat')"
 if [[ $OS_NAME == "macos" ]]; then
     packages+=", type = 'binary'"
 fi
-Rscript --vanilla -e "install.packages(${packages}, repos = '${CRAN_MIRROR}', lib = '${R_LIB_PATH}')" || exit -1
+Rscript --vanilla -e "install.packages(${packages}, repos = '${CRAN_MIRROR}', lib = '${R_LIB_PATH}', dependencies = c('Depends', 'Imports', 'LinkingTo'))" || exit -1
```
StrikerRUS: @jameslamb 
Just curious: what is the default set of options?

jameslamb: This set is the default set. From `?install.packages`

> The default, NA, means c("Depends", "Imports", "LinkingTo").

Those are the minimum ones needed to load a package. In case you're not familiar with them:

* `Depends`: Has to be present to install your package. Will be loaded whenever your package is loaded.
* `Imports`: Has to be present to install your package. Will only be loaded when your package calls code from them.
* `LinkingTo`: Has to be present to install your package. These are usually header-only files that your code links to, e.g. the [BH](https://cran.r-project.org/web/packages/BH/index.html) package for Boost.

This commit just makes that explicit.

StrikerRUS: Got it, thanks a lot for the detailed explanation!

PR #2958, opened 2020-03-30
[SWIG][mmlspark] fix seg fault in LGBM_BoosterGetEvalNamesSWIG

imatiach-msft:
fix seg fault in LGBM_BoosterGetEvalNamesSWIG


imatiach-msft:
tagging @AlbertoEAF - this issue was causing a seg fault when I tried to update to the new LGBM_BoosterGetEvalNamesSWIG API
also adding @StrikerRUS @guolinke for review

imatiach-msft:
@StrikerRUS @guolinke can the fix be merged - I can't update to latest right now because of the regression.  Thank you!

Issue #2472, opened 2019-09-29
[R-package] exclude training data from being checked for early stopping


StrikerRUS:
Check all datasets in validation list for early stopping excluding the training data to match the Python-package behavior:
> Early stopping requires at least one set in valid_sets. If there is more than one, it will use all of them except the training data:
https://github.com/microsoft/LightGBM/blob/master/docs/Python-Intro.rst#early-stopping

Refer to #2371.
Python code:
https://github.com/microsoft/LightGBM/blob/5c0daca0a1ae601aa2bdcaa0d2fc16bcf3be150a/python-package/lightgbm/callback.py#L238-L241

C++ code:
https://github.com/microsoft/LightGBM/blob/5c0daca0a1ae601aa2bdcaa0d2fc16bcf3be150a/src/io/config.cpp#L237-L248

StrikerRUS:
Closed in favor of being in #2302. We decided to keep all feature requests in one place.

Welcome to contribute this feature! Please re-open this issue (or post a comment if you are not a topic starter) if you are actively working on implementing this feature.

jameslamb:
this was fixed in #2961

Issue #2525, opened 2019-10-23
[R-package] Add unit tests for best metric iteration/value


Laurae2:
After PR https://github.com/microsoft/LightGBM/pull/2471 gets merged, unit tests should be added to ensure correctness of the iteration and value of the best metric during training of a LightGBM model.

A test matrix could be the following:

* Train + 1 metric
* Train + n metric
* n test + 1 metric
* n test + n metric
* Train + n test + 1 metric
* Train + n test + n metric

StrikerRUS:
@Laurae2 Are you going to work on this issue? If not, we should add `help wanted` label and close it according to the feature requests policy.

StrikerRUS:
Closed in favor of being in #2302. We decided to keep all feature requests in one place.

Welcome to contribute this feature! Please re-open this issue (or post a comment if you are not a topic starter) if you are actively working on implementing this feature.

jameslamb:
This has been solved by #2961

PR #2961, opened 2020-03-30
[R-package] fixed best_iter and best_score when training data is passed (fixes #2295, #2525)

### LINKED ISSUE #2295: Error in best score and best iter, wrong value returned

jameslamb:
I started to pick up #2471 but found that I needed to make enough changes that I thought it made sense to just put them in a new PR and close that one. I also think this makes the review process cleaner, so I'm not technically updating a PR that belongs to @Laurae2 while also acting as a reviewer on my own code.

I tried to address #2295 in this PR. Basically, when early stopping is not used, the best score and best iteration should be determined using the first metric and first validation set given, **excluding the training data**. #2295 suggested that right now, if you pass the training data to `valids` it's possible that `best_iter` and  `best_score` will be based on the training data.

This PR proposes the following improvements:

* exclude training data from consideration for `best_iter` and `best_score` in `lgb.train()`
* replace positional references like `booster$record_evals[[2L]][[1L]][[1L]]` with the actual names used in those nested lists
* centralize the naming scheme for `record_evals` in package constants to reduce risk of silent errors caused by hard-coding 
* add tests on `best_score` and `best_iter` behavior for classification and regression, for metrics to maximize and those to minimize, and for cases where the training data is passed to `valids` with a name other than `"train"` (to be sure code detecting it with `identical()` works)

This PR does not attempt to fix https://github.com/microsoft/LightGBM/issues/2472, as noted in https://github.com/microsoft/LightGBM/pull/2471#issuecomment-536300877. That will be fixed in a later PR.

### DIFF
--- a/R-package/R/callback.R
+++ b/R-package/R/callback.R
@@ -1,3 +1,11 @@
+# constants that control naming in lists
+.EVAL_KEY <- function() {
+  return("eval")
+}
+.EVAL_ERR_KEY <- function() {
+  return("eval_err")
+}
+
 #' @importFrom R6 R6Class
 CB_ENV <- R6::R6Class(
   "lgb.cb_env",
@@ -216,8 +224,8 @@ cb.record.evaluation <- function() {
 
         # Create dummy lists
         env$model$record_evals[[data_name]][[name]] <- list()
-        env$model$record_evals[[data_name]][[name]]$eval <- list()
-        env$model$record_evals[[data_name]][[name]]$eval_err <- list()
+        env$model$record_evals[[data_name]][[name]][[.EVAL_KEY()]] <- list()
+        env$model$record_evals[[data_name]][[name]][[.EVAL_ERR_KEY()]] <- list()
 
       }
 
@@ -238,12 +246,12 @@ cb.record.evaluation <- function() {
       name <- eval_res$name
 
       # Store evaluation data
-      env$model$record_evals[[data_name]][[name]]$eval <- c(
-        env$model$record_evals[[data_name]][[name]]$eval
+      env$model$record_evals[[data_name]][[name]][[.EVAL_KEY()]] <- c(
+        env$model$record_evals[[data_name]][[name]][[.EVAL_KEY()]]
         , eval_res$value
       )
-      env$model$record_evals[[data_name]][[name]]$eval_err <- c(
-        env$model$record_evals[[data_name]][[name]]$eval_err
+      env$model$record_evals[[data_name]][[name]][[.EVAL_ERR_KEY()]] <- c(
+        env$model$record_evals[[data_name]][[name]][[.EVAL_ERR_KEY()]]
         , eval_err
       )
 
--- a/R-package/R/lgb.Booster.R
+++ b/R-package/R/lgb.Booster.R
@@ -5,7 +5,7 @@ Booster <- R6::R6Class(
   public = list(
 
     best_iter = -1L,
-    best_score = NA,
+    best_score = NA_real_,
     record_evals = list(),
 
     # Finalize will free up the handles
@@ -989,11 +989,11 @@ lgb.get.eval.result <- function(booster, data_name, eval_name, iters = NULL, is_
   }
 
   # Create result
-  result <- booster$record_evals[[data_name]][[eval_name]]$eval
+  result <- booster$record_evals[[data_name]][[eval_name]][[.EVAL_KEY()]]
 
   # Check if error is requested
   if (is_err) {
-    result <- booster$record_evals[[data_name]][[eval_name]]$eval_err
+    result <- booster$record_evals[[data_name]][[eval_name]][[.EVAL_ERR_KEY()]]
   }
 
   # Check if iteration is non existant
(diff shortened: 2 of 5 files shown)


jameslamb:
Gah our CI is  🔥🔥🔥 this week! 

From [Azure DevOps](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=5679&view=logs&j=a0acac97-df52-5651-dae8-9fb37a4728fc&t=d2c32b51-981b-5bcd-fe31-029dd8d8998c)

image

image

jameslamb:
close re-open to re-trigger CI and check if the failures are transient

jameslamb:
After reading #2525, I think this PR will fix that too

jameslamb:
I've rebased this to update it with latest `master`. This PR is ready for review.

jameslamb:
Thanks for the review @guolinke ! I'm glad to get this one merged, think it will improve the user experience

PR #2962, opened 2020-03-31
[ci] reduce duplication in Python Windows CI

jameslamb:
Per discussion from https://github.com/microsoft/LightGBM/pull/2936#issuecomment-606720235. Right now some CI logic is duplicated in `.vsts-ci.yml` and `.appveyor.yml`. Similar to how we've centralized that logic for mac and linux builds in `.ci/` scripts, in this PR I propose cutting out that duplication and centralizing the logic in `.ci/tetst_windows.ps1`.

This should make it a little easier to make changes (such as changing our pegged version of scikit-learn whenever #2949 is merged) and should make #2936 easier to review.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -9,10 +9,8 @@ environment:
   matrix:
     - COMPILER: MSVC
       TASK: python
-      CONDA_ENV: test-env
     - COMPILER: MINGW
       TASK: python
-      CONDA_ENV: test-env
 
 clone_depth: 5
 
@@ -21,6 +19,7 @@ install:
   - set PATH=%PATH:C:\Program Files\Git\usr\bin;=%  # delete sh.exe from PATH (mingw32-make fix)
   - set PATH=C:\mingw-w64\x86_64-8.1.0-posix-seh-rt_v6-rev0\mingw64\bin;%PATH%
   - set PYTHON_VERSION=%CONFIGURATION%
+  - set CONDA_ENV="test-env"
   - ps: >-
       switch ($env:PYTHON_VERSION) {
           "2.7" {$env:MINICONDA = "C:\Miniconda-x64"}
@@ -35,5 +34,4 @@ install:
 build: false
 
 test_script:
-  - conda init powershell
   - powershell.exe -ExecutionPolicy Bypass -File .\.ci\test_windows.ps1
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -36,4 +36,4 @@ build: false
 
 test_script:
   - conda init powershell
-  - poowershell.exe -ExecutionPolicy Bypass -File .\.ci\test_windows.ps1
+  - powershell.exe -ExecutionPolicy Bypass -File .\.ci\test_windows.ps1
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -35,4 +35,4 @@ build: false
 
 test_script:
   - conda init powershell
-  - powershell.exe -ExecutionPolicy Bypass -File $env:APPVEYOR_BUILD_FOLDER\.ci\test_windows.ps1
+  - powershell.exe -ExecutionPolicy Bypass -File %APPVEYOR_BUILD_FOLDER%\.ci\test_windows.ps1
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -8,7 +8,11 @@ configuration:  # a trick to construct a build matrix with multiple Python versi
 environment:
   matrix:
     - COMPILER: MSVC
+      TASK: appveyor-python
+      CONDA_ENV: test-env
     - COMPILER: MINGW
+      TASK: appveyor-python
+      CONDA_ENV: test-env
 
 clone_depth: 5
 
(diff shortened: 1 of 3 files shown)


.vsts-ci.yml
```diff
@@ -129,11 +129,13 @@ jobs:
         TASK: bdist
         PYTHON_VERSION: 3.5
   steps:
-  - powershell: Write-Host "##vso[task.prependpath]$env:CONDA\Scripts"
-    displayName: Enable conda
+  - powershell: |
+      Write-Host "##vso[task.prependpath]$env:CONDA\Scripts"
+      Write-Host "##vso[task.setvariable variable=AZURE]true"
```
StrikerRUS: I think we should have the same for Appveyor. I mean, `$env:APPVEYOR = "true"`. And break the condition `$env:TASK -eq "appveyor-python"` into `$env:APPVEYOR -eq "true"` and `$env:TASK -eq "python"`.

jameslamb: good idea,  ok!

jameslamb: let me know if this is what you're thinking: https://github.com/microsoft/LightGBM/pull/2962/commits/79c48abcf9ad0ae2c843b033e99587562ea90687

I thought it was better to set `APPVEYOR` inside the script since that is a variable already provided on AppVeyor builds. Setting in the scripts guarantees that we're overwriting it. It also means we don't have to copy it `n_tasks` times in the `matrix:` block

jameslamb: oops, forgot the task name: https://github.com/microsoft/LightGBM/pull/2962/commits/a331e2b1a13da7a9b82ae2d37b2677499d2bd8af

.appveyor.yml
```diff
@@ -8,7 +8,11 @@ configuration:  # a trick to construct a build matrix with multiple Python versi
 environment:
   matrix:
     - COMPILER: MSVC
+      TASK: python
+      CONDA_ENV: test-env
```
StrikerRUS, line 12: I don't think we need `CONDA_ENV` in build matrix. Set it like `PYTHON_VERSION`
https://github.com/microsoft/LightGBM/blob/dbc3d05c9c6f89d296a29b73a85a847c818892ad/.appveyor.yml#L19

.ci/test_windows.ps1
```diff
@@ -6,6 +6,20 @@ function Check-Output {
...
+  $env:APPVEYOR = "true"
+  $env:BUILD_SOURCESDIRECTORY = $env:APPVEYOR_BUILD_FOLDER
+}
+
+# setup for Python
+conda init powershell
+conda activate
+conda config --set always_yes yes --set changeps1 no
+conda update -q -y conda
+conda create -q -y -n $env:CONDA_ENV python=$env:PYTHON_VERSION joblib matplotlib numpy pandas psutil pytest python-graphviz "scikit-learn<=0.21.3" scipy wheel ; Check-Output $?
```
StrikerRUS, line 20: Why do we need `wheel` here explicitly?

.ci/test_windows.ps1
```diff
@@ -33,15 +47,27 @@ elseif ($env:TASK -eq "bdist") {
   python setup.py bdist_wheel --plat-name=win-amd64 --universal ; Check-Output $?
   cd dist; pip install @(Get-ChildItem *.whl) ; Check-Output $?
   cp @(Get-ChildItem *.whl) $env:BUILD_ARTIFACTSTAGINGDIRECTORY
+} elseif (($env:APPVEYOR -eq "true") -and ($env:TASK -eq "python")) {
+  cd $env:BUILD_SOURCESDIRECTORY\python-package
+  if ($env:COMPILER -eq "MINGW") {
+    python setup.py install --mingw | Check-Output $?
```
StrikerRUS, line 53:
```suggestion
    python setup.py install --mingw ; Check-Output $?
```

.ci/test_windows.ps1
```diff
@@ -33,15 +47,27 @@ elseif ($env:TASK -eq "bdist") {
...
+  } else {
+    python setup.py install | Check-Output $?
```
StrikerRUS, line 55:
```suggestion
    python setup.py install ; Check-Output $?
```

.vsts-ci.yml
```diff
@@ -129,11 +129,13 @@ jobs:
...
+    displayName: 'Set Variables'
   - script: |
-      cmd /c 'activate & conda config --set always_yes yes --set changeps1 no & conda update -q -y conda & conda create -q -y -n %CONDA_ENV% python=%PYTHON_VERSION% joblib matplotlib numpy pandas psutil pytest python-graphviz "scikit-learn<=0.21.3" scipy'
-      cmd /c "activate %CONDA_ENV% & powershell -ExecutionPolicy Bypass -File %BUILD_SOURCESDIRECTORY%/.ci/test_windows.ps1"
+      cmd /c "conda init powershell"
```
StrikerRUS, line 137: I think we do `init` in `test_windows.ps1#16` and this line is redundant.

.appveyor.yml
jameslamb: ah good idea

.ci/test_windows.ps1
jameslamb: I was getting a failure with `bdist` because `bdist_wheel` wasn't recognized. But I know that task has been working for a while so maybe it was something else I was doing wrong. I'll try removing it

.vsts-ci.yml
jameslamb: I intentionally did it twice, actually.

When you run this command it says:

```
For changes to take effect, close and re-open your current shell. 
```

And I was getting errors like "`pytest` is not a recognized cmdlet..." that told me that `conda activate`  wasn't working.

I can push a commit testing without it though! Maybe those errors were caused by something else I fixed while testing stuff.

jameslamb: As of https://github.com/microsoft/LightGBM/pull/2962/commits/64ef6ac6aeac47d443b9079547dd58d6091c8b5c I see the issue:

image

Once I put back in the duplicate `conda init powershell`, it works again (https://github.com/microsoft/LightGBM/pull/2962/commits/109e152cd5dea022e0b769f9fc900a5d7f3434c3)

StrikerRUS: Ah, I see! Then I think it's worth commenting this line to prevent its removal in the future.

StrikerRUS: Never mind! I see you added this line into both yaml configurations, so it should be quite obvious.

.appveyor.yml
```diff
@@ -17,41 +19,20 @@ install:
...
-      foreach ($file in @(Get-ChildItem *.py)) {
-        @("import sys, warnings", "warnings.showwarning = lambda message, category, filename, lineno, file=None, line=None: sys.stdout.write(warnings.formatwarning(message, category, filename, lineno, line))") + (Get-Content $file) | Set-Content $file
-        python $file
-        if (!$?) { $host.SetShouldExit(-1) }
-      }  # run all examples
-  - cd %APPVEYOR_BUILD_FOLDER%\examples\python-guide\notebooks
-  - conda install -q -y -n test-env ipywidgets notebook
-  - jupyter nbconvert --ExecutePreprocessor.timeout=180 --to notebook --execute --inplace *.ipynb  # run all notebooks
+  - conda init powershell
+  - powershell.exe -ExecutionPolicy Bypass -File .\.ci\test_windows.ps1
```
StrikerRUS, line 38: Please use absolute path here. And if possible for `$env:LGB_VER` too.

jameslamb: Done https://github.com/microsoft/LightGBM/pull/2962/commits/70525bd2708289d78de748ca7f669f735becc519

I checked  `.vsts-ci.yml` too, looks like everything there is already absolute paths

PR #2963, opened 2020-03-31
[R-package] CMake fixes to support MSVC

jameslamb:
As of #2901 , when you build the R package the R core library now needs to be linked to `lib_lightgbm`. As a result, I think that building the R package is broken for MSVC right now. We have code to create [the .lib file for linking](https://docs.microsoft.com/en-us/cpp/build/reference/link-input-files?redirectedfrom=MSDN&view=vs-2019) but aren't currently linking it.

If you try to build the R package from the version on `master` with MSVC you will get a linker error.

In this PR, I propose adding that step to `CMakeLists.txt`. I've also proposed some simplifications to `FindLibR.cmake`. It seems that the variable `LIBR_LIB_DIR` isn't actually necessary. It's found using the R code `R.home('lib')`...there is nothing special about this code. I learned that `R.home(some_directory)` returns a string with the value of `some_directory` tacked onto the end of `R.home()`, regardless of whether or not that exists!

image

Given that, we might as well just use `"${LIBR_HOME}/lib"` directly in the search for `LIBR_CORE_LIBRARY` and not go through that extra step with R.

The code from this PR is working on #2936 , and you can confirm that it works with MSVC [in the Azure builds for that PR](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=5689&view=logs&j=b51fa6ee-4364-59e6-4527-dc4aaafd2db2&t=85bde6ad-0953-589b-e1e8-633bb2d9d530) and with MinGW [in the AppVeyor builds for that PR](https://ci.appveyor.com/project/guolinke/lightgbm/builds/31839326/job/xx2uv1bxn0ccar72). You can confirm that these changes don't break linux and mac by looking at [the Travis builds for that PR](https://travis-ci.org/github/microsoft/LightGBM/builds/669074761?utm_source=github_status&utm_medium=notification).

Splitting this out into its own PR to fix the MSVC problem quicker and to make #2936 easier to review.

PR #2965, opened 2020-04-02
[R-package] speed up installation with Visual Studio

jameslamb:
This is another PR that contributes to #2936 , but which I think has standalone value.

Right now on Windows, to figure out which version of Visual Studio is available we try to build `lib_lightgbm` by generating and executing the corresponding `CMake` Makefile for the 3 most recent versions of VS. This is pretty expensive, and means that you have to build `lib_lightgbm` .

I went into this thinking that this is why r-package CI builds on Azure Devops with Visual Studio [take 20 minutes](https://dev.azure.com/JayLamb20/jameslamb-lightgbm/_build/results?buildId=248&view=logs&j=87d556e4-1858-51ea-bb1c-8e61473c09c8)...and it seems like it! I tested merging these changes into #2936 [on my fork](https://github.com/jameslamb/LightGBM/pull/13), and it takes the build time down from 20 minutes to [about 9 minutes](https://dev.azure.com/JayLamb20/jameslamb-lightgbm/_build/results?buildId=260&view=logs&j=b51fa6ee-4364-59e6-4527-dc4aaafd2db2&t=85bde6ad-0953-589b-e1e8-633bb2d9d530).

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -10,6 +10,9 @@ environment:
     - COMPILER: MINGW
       TASK: r-package
     - COMPILER: MSVC
+      TASK: r-package
+      APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
+    - COMPILER: MSVC
       TASK: python
     - COMPILER: MINGW
       TASK: python
--- a/.ci/test_r_package_windows.ps1
+++ b/.ci/test_r_package_windows.ps1
@@ -14,6 +14,7 @@ function Download-File-With-Retries {
 
 $env:R_WINDOWS_VERSION = "3.6.3"
 $env:R_LIB_PATH = "$env:BUILD_SOURCESDIRECTORY/RLibrary" -replace '[\\]', '/'
+$env:R_LIBS = "$env:R_LIB_PATH"
 $env:PATH = "$env:R_LIB_PATH/Rtools/bin;" + "$env:R_LIB_PATH/R/bin/x64;" + "$env:R_LIB_PATH/miktex/texmfs/install/miktex/bin/x64;" + $env:PATH
 $env:CRAN_MIRROR = "https://cloud.r-project.org/"
 $env:CTAN_MIRROR = "https://ctan.math.illinois.edu/systems/win32/miktex/tm/packages/"
@@ -64,45 +65,65 @@ if ($env:COMPILER -eq "MINGW") {
     conda install -q -y --no-deps pandoc
 }
 
-Add-Content .Renviron "R_LIBS=$env:R_LIB_PATH"
-
 Write-Output "Installing dependencies"
-$packages = "c('data.table', 'jsonlite', 'Matrix', 'R6', 'testthat'), dependencies = c('Imports', 'Depends', 'LinkingTo')"
+$packages = "c('data.table', 'jsonlite', 'Matrix', 'processx', 'R6', 'testthat'), dependencies = c('Imports', 'Depends', 'LinkingTo')"
 Rscript --vanilla -e "options(install.packages.check.source = 'no'); install.packages($packages, repos = '$env:CRAN_MIRROR', type = 'binary', lib = '$env:R_LIB_PATH')" ; Check-Output $?
 
 Write-Output "Building R package"
-Rscript build_r.R --skip-install ; Check-Output $?
 
-$PKG_FILE_NAME = Get-Item *.tar.gz
-$LOG_FILE_NAME = "lightgbm.Rcheck/00check.log"
+# R CMD check is not used for MSVC builds
+if ($env:COMPILER -ne "MSVC") {
+  Rscript build_r.R --skip-install ; Check-Output $?
 
-$env:_R_CHECK_FORCE_SUGGESTS_ = 0
-if ($env:COMPILER -ne "MINGW") {
-  Write-Output "Running R CMD check without checking documentation"
-  R.exe CMD check --no-multiarch --no-examples --no-manual --ignore-vignettes ${PKG_FILE_NAME} ; $check_succeeded = $?
-} else {
+  $PKG_FILE_NAME = Get-Item *.tar.gz
+  $LOG_FILE_NAME = "lightgbm.Rcheck/00check.log"
+
+  $env:_R_CHECK_FORCE_SUGGESTS_ = 0
   Write-Output "Running R CMD check as CRAN"
   R.exe CMD check --no-multiarch --as-cran ${PKG_FILE_NAME} ; $check_succeeded = $?
+
+  Write-Output "R CMD check build logs:"
+  $INSTALL_LOG_FILE_NAME = "$env:BUILD_SOURCESDIRECTORY\lightgbm.Rcheck\00install.out"
+  Get-Content -Path "$INSTALL_LOG_FILE_NAME"
+
+  Check-Output $check_succeeded
+
+  Write-Output "Looking for issues with R CMD check results"
+  if (Get-Content "$LOG_FILE_NAME" | Select-String -Pattern "WARNING" -Quiet) {
+      echo "WARNINGS have been found by R CMD check!"
+      Check-Output $False
+  }
+
+  $note_str = Get-Content "${LOG_FILE_NAME}" | Select-String -Pattern ' NOTE' | Out-String ; Check-Output $?
+  $relevant_line = $note_str -match '.*Status: (\d+) NOTE.*'
+  $NUM_CHECK_NOTES = $matches[1]
+  $ALLOWED_CHECK_NOTES = 3
+  if ([int]$NUM_CHECK_NOTES -gt $ALLOWED_CHECK_NOTES) {
+      Write-Output "Found ${NUM_CHECK_NOTES} NOTEs from R CMD check. Only ${ALLOWED_CHECK_NOTES} are allowed"
+      Check-Output $False
+  }
+} else {
+  $INSTALL_LOG_FILE_NAME = "$env:BUILD_SOURCESDIRECTORY\00install_out.txt"
+  Rscript build_r.R *> $INSTALL_LOG_FILE_NAME ; $install_succeeded = $?
+  Write-Output "----- build and install logs -----"
+  Get-Content -Path "$INSTALL_LOG_FILE_NAME"
+  Write-Output "----- end of build and install logs -----"
+  Check-Output $install_succeeded
 }
 
-Write-Output "R CMD check build logs:"
-Get-Content -Path $env:BUILD_SOURCESDIRECTORY\lightgbm.Rcheck\00install.out
-
-Check-Output $check_succeeded
-
-Write-Output "Looking for issues with R CMD check results"
-if (Get-Content "$LOG_FILE_NAME" | Select-String -Pattern "WARNING" -Quiet) {
-    echo "WARNINGS have been found by R CMD check!"
-    Check-Output $False
+# Checking that we actually got the expected compiler. The R package has some logic
+# to fail back to MinGW if MSVC fails, but for CI builds we need to check that the correct
+# compiler was used.
+$checks = Select-String -Path "${INSTALL_LOG_FILE_NAME}" -Pattern "Check for working CXX compiler.*$env:COMPILER"
+if ($checks.Matches.length -eq 0) {
+  Write-Output "The wrong compiler was used. Check the build logs."
+  Check-Output $False
 }
 
-$note_str = Get-Content "${LOG_FILE_NAME}" | Select-String -Pattern ' NOTE' | Out-String ; Check-Output $?
-$relevant_line = $note_str -match '.*Status: (\d+) NOTE.*'
-$NUM_CHECK_NOTES = $matches[1]
-$ALLOWED_CHECK_NOTES = 3
-if ([int]$NUM_CHECK_NOTES -gt $ALLOWED_CHECK_NOTES) {
-    Write-Output "Found ${NUM_CHECK_NOTES} NOTEs from R CMD check. Only ${ALLOWED_CHECK_NOTES} are allowed"
-    Check-Output $False
+if ($env:COMPILER -eq "MSVC") {
+  Write-Output "Running tests with testthat.R"
+  cd R-package/tests
+  Rscript testthat.R ; Check-Output $?
 }
 
 Write-Output "No issues were found checking the R package"
--- a/.vsts-ci.yml
+++ b/.vsts-ci.yml
@@ -120,7 +120,7 @@ jobs:
     matrix:
       r_package:
         TASK: r-package
-        COMPILER: MINGW
+        COMPILER: MSVC
       regular:
         TASK: regular
         PYTHON_VERSION: 3.6
--- a/R-package/src/install.libs.R
+++ b/R-package/src/install.libs.R
@@ -164,6 +164,7 @@ if (!use_precompile) {
     }
   } else {
       .run_shell_command("cmake", c(cmake_args, ".."))
+      makefiles_already_generated <- TRUE
   }
 
   # generate build files
--- a/R-package/src/install.libs.R
+++ b/R-package/src/install.libs.R
@@ -12,7 +12,77 @@ R_ver <- as.double(R.Version()$major) + as.double(R.Version()$minor) / 10.0
 
 if (!(R_int_UUID == "0310d4b8-ccb1-4bb8-ba94-d36a55f60262"
     || R_int_UUID == "2fdf6c18-697a-4ba7-b8ef-11c0d92f1327")) {
-  print("Warning: unmatched R_INTERNALS_UUID, may cannot run normally.")
+  warning("Warning: unmatched R_INTERNALS_UUID, may not run normally.")
+}
+
+# system() will not raise an R exception if the process called
+# fails. Wrapping it here to get that behavior.
+#
+# system() introduces a lot of overhead, at least on Windows,
+# so trying processx if it is available
+.run_shell_command <- function(cmd, args, strict = TRUE) {
+    on_windows <- .Platform$OS.type == "windows"
+    has_processx <- suppressMessages({
+      suppressWarnings({
+        require("processx")  # nolint
+      })
+    })
+    if (has_processx && on_windows) {
+      result <- processx::run(
+        command = cmd
+        , args = args
+        , windows_verbatim_args = TRUE
+        , error_on_status = FALSE
+        , echo = TRUE
+      )
+      exit_code <- result$status
+    } else {
+      if (on_windows) {
+        message(paste0(
+          "Using system() to run shell commands. Installing "
+          , "'processx' with install.packages('processx') might "
+          , "make this faster."
+        ))
+      }
+      cmd <- paste0(cmd, " ", paste0(args, collapse = " "))
+      exit_code <- system(cmd)
+    }
+
+    if (exit_code != 0L && isTRUE(strict)) {
+        stop(paste0("Command failed with exit code: ", exit_code))
+    }
+    return(invisible(exit_code))
+}
+
+# try to generate Visual Studio build files
+.generate_vs_makefiles <- function(cmake_args) {
+  vs_versions <- c(
+    "Visual Studio 16 2019"
+    , "Visual Studio 15 2017"
+    , "Visual Studio 14 2015"
+  )
+  working_vs_version <- NULL
+  for (vs_version in vs_versions) {
+    message(sprintf("Trying '%s'", vs_version))
+    # if the build directory is not empty, clean it
+    if (file.exists("CMakeCache.txt")) {
+      file.remove("CMakeCache.txt")
+    }
+    vs_cmake_args <- c(
+      cmake_args
+      , "-G"
+      , shQuote(vs_version)
+      , "-A"
+      , "x64"
+    )
+    exit_code <- .run_shell_command("cmake", c(vs_cmake_args, ".."), strict = FALSE)
+    if (exit_code == 0L) {
+      message(sprintf("Successfully created build files for '%s'", vs_version))
+      return(invisible(TRUE))
+    }
+
+  }
+  return(invisible(FALSE))
 }
 
 # Move in CMakeLists.txt
@@ -41,17 +111,18 @@ if (!use_precompile) {
   setwd(build_dir)
 
   # Prepare installation steps
-  cmake_cmd <- "cmake "
-  build_cmd <- "make _lightgbm"
+  cmake_args <- NULL
+  build_cmd <- "make"
+  build_args <- "_lightgbm"
   lib_folder <- file.path(source_dir, fsep = "/")
 
   if (use_gpu) {
-    cmake_cmd <- paste0(cmake_cmd, " -DUSE_GPU=ON ")
+    cmake_args <- c(cmake_args, "-DUSE_GPU=ON")
   }
   if (R_ver >= 3.5) {
-    cmake_cmd <- paste0(cmake_cmd, " -DUSE_R35=ON ")
+    cmake_args <- c(cmake_args, "-DUSE_R35=ON")
   }
-  cmake_cmd <- paste0(cmake_cmd, " -DBUILD_FOR_R=ON ")
+  cmake_args <- c(cmake_args, "-DBUILD_FOR_R=ON")
 
   # Pass in R version, used to help find R executable for linking
   R_version_string <- paste(
@@ -59,53 +130,46 @@ if (!use_precompile) {
     , R.Version()[["minor"]]
     , sep = "."
   )
-  cmake_cmd <- sprintf(
-    paste0(cmake_cmd, " -DCMAKE_R_VERSION='%s' ")
-    , R_version_string
-  )
+  r_version_arg <- sprintf("-DCMAKE_R_VERSION='%s'", R_version_string)
+  cmake_args <- c(cmake_args, r_version_arg)
+
+  # the checks below might already run `cmake -G`. If they do, set this flag
+  # to TRUE to avoid re-running it later
+  makefiles_already_generated <- FALSE
 
   # Check if Windows installation (for gcc vs Visual Studio)
   if (WINDOWS) {
     if (use_mingw) {
-      print("Trying to build with MinGW")
-      cmake_cmd <- paste0(cmake_cmd, " -G \"MinGW Makefiles\" ")
-      build_cmd <- "mingw32-make.exe _lightgbm"
-      system(paste0(cmake_cmd, " ..")) # Must build twice for Windows due sh.exe in Rtools
+      message("Trying to build with MinGW")
+      # Must build twice for Windows due sh.exe in Rtools
+      cmake_args <- c(cmake_args, "-G", shQuote("MinGW Makefiles"))
+      .run_shell_command("cmake", c(cmake_args, ".."), strict = FALSE)
+      build_cmd <- "mingw32-make.exe"
+      build_args <- "_lightgbm"
     } else {
-      local_vs_def <- ""
-      vs_versions <- c(
-        "Visual Studio 16 2019"
-        , "Visual Studio 15 2017"
-        , "Visual Studio 14 2015"
-      )
-      for (vs in vs_versions) {
-        print(paste0("Trying to build with: '", vs, "'"))
-        vs_def <- paste0(" -G \"", vs, "\" -A x64")
-        tmp_cmake_cmd <- paste0(cmake_cmd, vs_def)
-        try_vs <- system(paste0(tmp_cmake_cmd, " .."))
-        if (try_vs == 0L) {
-          local_vs_def <- vs_def
-          print(paste0("Building with '", vs, "' succeeded"))
-          break
-        } else {
-          unlink("./*", recursive = TRUE) # Clean up build directory
-        }
-      }
-      if (try_vs == 1L) {
-        print("Building with Visual Studio failed. Attempted with MinGW")
-        cmake_cmd <- paste0(cmake_cmd, " -G \"MinGW Makefiles\" ")
-        system(paste0(cmake_cmd, " ..")) # Must build twice for Windows due sh.exe in Rtools
-        build_cmd <- "mingw32-make.exe _lightgbm"
+      visual_studio_succeeded <- .generate_vs_makefiles(cmake_args)
+      if (!isTRUE(visual_studio_succeeded)) {
+        warning("Building with Visual Studio failed. Attempting with MinGW")
+        # Must build twice for Windows due sh.exe in Rtools
+        cmake_args <- c(cmake_args, "-G", shQuote("MinGW Makefiles"))
+        .run_shell_command("cmake", c(cmake_args, ".."), strict = FALSE)
+        build_cmd <- "mingw32-make.exe"
+        build_args <- "_lightgbm"
       } else {
-        cmake_cmd <- paste0(cmake_cmd, local_vs_def)
-        build_cmd <- "cmake --build . --target _lightgbm --config Release"
+        build_cmd <- "cmake"
+        build_args <- c("--build", ".", "--target", "_lightgbm", "--config", "Release")
         lib_folder <- file.path(source_dir, "Release", fsep = "/")
+        makefiles_already_generated <- TRUE
       }
     }
+  } else {
+      .run_shell_command("cmake", c(cmake_args, ".."))
   }
 
-  # Install
-  system(paste0(cmake_cmd, " .."))
+  # generate build files
+  if (!makefiles_already_generated) {
+    .run_shell_command("cmake", c(cmake_args, ".."))
+  }
 
   # R CMD check complains about the .NOTPARALLEL directive created in the cmake
   # Makefile. We don't need it here anyway since targets are built serially, so trying
@@ -130,7 +194,8 @@ if (!use_precompile) {
     )
   }
 
-  system(build_cmd)
+  # build the library
+  .run_shell_command(build_cmd, build_args)
   src <- file.path(lib_folder, paste0("lib_lightgbm", SHLIB_EXT), fsep = "/")
 
 } else {
@@ -169,7 +234,7 @@ if (!use_precompile) {
 dest <- file.path(R_PACKAGE_DIR, paste0("libs", R_ARCH), fsep = "/")
 dir.create(dest, recursive = TRUE, showWarnings = FALSE)
 if (file.exists(src)) {
-  print(paste0("Found library file: ", src, " to move to ", dest))
+  message(paste0("Found library file: ", src, " to move to ", dest))
   file.copy(src, dest, overwrite = TRUE)
 
   symbols_file <- file.path(source_dir, "symbols.rds")
@@ -183,7 +248,7 @@ if (file.exists(src)) {
 
 # clean up the "build" directory
 if (dir.exists(build_dir)) {
-  print("Removing 'build/' directory")
+  message("Removing 'build/' directory")
   unlink(
     x = build_dir
     , recursive = TRUE
--- a/build_r.R
+++ b/build_r.R
@@ -17,12 +17,42 @@ INSTALL_AFTER_BUILD <- !("--skip-install" %in% args)
 }
 
 # system() will not raise an R exception if the process called
-# fails. Wrapping it here to get that behavior
-.run_shell_command <- function(cmd, ...) {
-    exit_code <- system(cmd, ...)
-    if (exit_code != 0L) {
+# fails. Wrapping it here to get that behavior.
+#
+# system() introduces a lot of overhead, at least on Windows,
+# so trying processx if it is available
+.run_shell_command <- function(cmd, args, strict = TRUE) {
+    on_windows <- .Platform$OS.type == "windows"
+    has_processx <- suppressMessages({
+      suppressWarnings({
+        require("processx")  # nolint
+      })
+    })
+    if (has_processx && on_windows) {
+      result <- processx::run(
+        command = cmd
+        , args = args
+        , windows_verbatim_args = TRUE
+        , error_on_status = FALSE
+        , echo = TRUE
+      )
+      exit_code <- result$status
+    } else {
+      if (on_windows) {
+        message(paste0(
+          "Using system() to run shell commands. Installing "
+          , "'processx' with install.packages('processx') might "
+          , "make this faster."
+        ))
+      }
+      cmd <- paste0(cmd, " ", paste0(args, collapse = " "))
+      exit_code <- system(cmd)
+    }
+
+    if (exit_code != 0L && isTRUE(strict)) {
         stop(paste0("Command failed with exit code: ", exit_code))
     }
+    return(invisible(exit_code))
 }
 
 # Make a new temporary folder to work in
@@ -73,8 +103,7 @@ result <- file.copy(
 # NOTE: --keep-empty-dirs is necessary to keep the deep paths expected
 #       by CMake while also meeting the CRAN req to create object files
 #       on demand
-cmd <- "R CMD build lightgbm_r --keep-empty-dirs"
-.run_shell_command(cmd)
+.run_shell_command("R", c("CMD", "build", "lightgbm_r", "--keep-empty-dirs"))
 
 # Install the package
 version <- gsub(
@@ -88,9 +117,11 @@ version <- gsub(
 )
 tarball <- file.path(getwd(), sprintf("lightgbm_%s.tar.gz", version))
 
-cmd <- sprintf("R CMD INSTALL %s --no-multiarch --with-keep.source", tarball)
+install_cmd <- "R"
+install_args <- c("CMD", "INSTALL", "--no-multiarch", "--with-keep.source", tarball)
 if (INSTALL_AFTER_BUILD) {
-  .run_shell_command(cmd)
+  .run_shell_command(install_cmd, install_args)
 } else {
+  cmd <- paste0(install_cmd, " ", paste0(install_args, collapse = " "))
   print(sprintf("Skipping installation. Install the package with command '%s'", cmd))
 }
--- a/R-package/DESCRIPTION
+++ b/R-package/DESCRIPTION
@@ -27,6 +27,7 @@ Biarch: false
 Suggests:
     ggplot2 (>= 1.0.1),
     knitr,
+    processx,
     rmarkdown,
     testthat
 Depends:


R-package/src/install.libs.R
```diff
@@ -15,6 +15,39 @@ if (!(R_int_UUID == "0310d4b8-ccb1-4bb8-ba94-d36a55f60262"
...
+    build_dir <- tempdir()
+    setwd(build_dir)
+    writeLines(
+      text = "PROJECT(testing)"
+      , con = file.path(build_dir, "CMakeLists.txt")
+    )
+    cmake_cmd <- paste0(
+      "cmake -Wno-dev -G "
+      , shQuote(vs_version)
+      , " ."
```
StrikerRUS, line 40: Wasn't 64-bit qualificator `-A x64` lost?

R-package/src/install.libs.R
```diff
@@ -15,6 +15,39 @@ if (!(R_int_UUID == "0310d4b8-ccb1-4bb8-ba94-d36a55f60262"
   print("Warning: unmatched R_INTERNALS_UUID, may cannot run normally.")
 }
 
+.find_vs_version <- function() {
+  current_working_dir <- getwd()
+  on.exit({
+    setwd(current_working_dir)
+  })
```
StrikerRUS, line 22: Should we also remove `build_dir`?

jameslamb: Each R session creates a temporary directory that is attached to the session and destroyed whenever the session exits. `tempdir()` creates a directory under that one, so it also automatically gets destroyed.

R-package/src/install.libs.R
jameslamb: whoops yeah you're right! But we don't need it on this line

jameslamb: added: https://github.com/microsoft/LightGBM/pull/2965/commits/5e125b92f7f1de59d28052ea920793246a4169ad

testing on my fork: https://github.com/jameslamb/LightGBM/pull/13

jameslamb: seems that it succeeded!

AppVeyor: https://ci.appveyor.com/project/jameslamb/lightgbm/builds/31913976
Azure DevOps: https://dev.azure.com/JayLamb20/jameslamb-lightgbm/_build/results?buildId=266&view=logs&jobId=b51fa6ee-4364-59e6-4527-dc4aaafd2db2&j=b51fa6ee-4364-59e6-4527-dc4aaafd2db2&t=85bde6ad-0953-589b-e1e8-633bb2d9d530

R-package/src/install.libs.R
jameslamb: not bad for a programming language invented by and for statisticians, huh 😀

R-package/src/install.libs.R
StrikerRUS: Unfortunately, those logs cannot help to check whether the build was successful with VS or due to fallback to MinGW. Or am I just searching badly? 🙁

jameslamb: oh right! Because the install is happening inside `R CMD check`'s step that is like `checking if 'lightgbm' can be installed...`  and those logs are suppressed.

Let me do a build where I remove the `--skip-install` on `build_r.R` so you can see it.

jameslamb: ok try these:

AppVeyor (MinGW): https://ci.appveyor.com/project/jameslamb/lightgbm/builds/31919068/job/66631ptg2ublemex

image

Azure DevOps (MSVC from Visual Studio): https://dev.azure.com/JayLamb20/jameslamb-lightgbm/_build/results?buildId=267&view=logs&j=b51fa6ee-4364-59e6-4527-dc4aaafd2db2

image

^ @StrikerRUS I'm glad you asked! It looks like it is finding the  `gcc.exe` bundled  with chocolatey 😬 . I know that when I was testing before, Visual Studio was found. I will have to test some more :/

StrikerRUS: Also I can say that according to the logs we should remove temp build directory between tries.

jameslamb: Yep definitely agree. I'm testing on my fork.

You can see on this commit that MSVC from Visual studio is being used ([build link](https://dev.azure.com/JayLamb20/jameslamb-lightgbm/_build/results?buildId=269&view=logs&j=b51fa6ee-4364-59e6-4527-dc4aaafd2db2))

image

But I still haven't gotten a successful build. Will `@` you when I do.

Really really glad you asked! I never considered that `gcc` might just be laying around from something else 😬

jameslamb: I think I should print the contents of `lightgbm.Rcheck/00install.out` so we can see this in our logs going forward. That's a more permanent fix than removing `--skip-install`

jameslamb: ok this did finally succeed but it took *49 minutes*. 

https://dev.azure.com/JayLamb20/jameslamb-lightgbm/_build/results?buildId=273&view=logs&j=b51fa6ee-4364-59e6-4527-dc4aaafd2db2&t=85bde6ad-0953-589b-e1e8-633bb2d9d530

I think I might know what's happening, trying something else.

StrikerRUS: >  I never considered that gcc might just be laying around from something else 😬

I was surprised too!

 >I think I should print the contents of `lightgbm.Rcheck/00install.out` so we can see this in our logs going forward. That's a more permanent fix than removing `--skip-install`

If this is the only way to force displaying logging `print`s, then yes, I agree with you.

> ok this did finally succeed but it took 49 minutes.

Oh, my! 😬

jameslamb: Ok I think this is working! But....it takes 26 minutes :/

https://dev.azure.com/JayLamb20/jameslamb-lightgbm/_build/results?buildId=275&view=logs&jobId=b51fa6ee-4364-59e6-4527-dc4aaafd2db2&j=b51fa6ee-4364-59e6-4527-dc4aaafd2db2&t=85bde6ad-0953-589b-e1e8-633bb2d9d530

I think the 20 minutes I referenced in https://github.com/microsoft/LightGBM/pull/2965#issue-397341495 was suffering from silently getting the `gcc.exe` from chocolatey, so it wasn't real :/

I know this is super long. I tried to mitigate the effect on total runtime with two changes in #2936:

* moving `r-package` task first in `.vsts-ci`
* only running `R CMD check --no-manual --ignore-vignettes` to cut out time spent testing documentation (which is already tested on AppVeyor)

I might be able to cut out more time, but either way I don't think that should impact this PR. I think that build link is evidence  that what I'm doing here in this PR does successfully build the R package with MSVC

jameslamb:
@Laurae2 @StrikerRUS could you have another look?

R-package/src/install.libs.R
```diff
@@ -67,36 +106,20 @@ if (!use_precompile) {
...
+      local_vs_def <- .find_vs_version()
+      if (is.null(local_vs_def)) {
+        print("Building with Visual Studio failed. Attempting with MinGW")
         cmake_cmd <- paste0(cmake_cmd, " -G \"MinGW Makefiles\" ")
         system(paste0(cmake_cmd, " ..")) # Must build twice for Windows due sh.exe in Rtools
         build_cmd <- "mingw32-make.exe _lightgbm"
       } else {
-        cmake_cmd <- paste0(cmake_cmd, local_vs_def)
+        print(paste0("Building with ", shQuote(local_vs_def)))
+        cmake_cmd <- paste0(cmake_cmd, " -G ", shQuote(local_vs_def), " -A x64")
```
StrikerRUS, line 122: Why do you use manual escaping during compiling a fake project and `shQuote` here?

R-package/src/install.libs.R
```diff
@@ -67,36 +106,20 @@ if (!use_precompile) {
...
-          break
-        } else {
-          unlink("./*", recursive = TRUE) # Clean up build directory
-        }
-      }
-      if (try_vs == 1L) {
-        print("Building with Visual Studio failed. Attempted with MinGW")
+      local_vs_def <- .find_vs_version()
+      if (is.null(local_vs_def)) {
+        print("Building with Visual Studio failed. Attempting with MinGW")
```
StrikerRUS, line 116: Is there something more "powerful" than `print` which will be able to show this critically important message even in `R CMD check`?

jameslamb: Not that I'm aware of! `R CMD check` is only a concern for us developers, not for our users, so I think just `cat`-ing out the `00install.out` file generated by `R CMD CHECK` is ok.

`print()` is enough for people who run `Rscript build_r.R` per our installation docs to see this message.

R-package/src/install.libs.R
jameslamb: ah just an oversight, they probably should both be `shQuote`. I'll change it

R-package/src/install.libs.R
StrikerRUS: >  so I think just `cat`-ing out the `00install.out` file generated by `R CMD CHECK` is ok.

Yeah, I just wanted to check logs from time to time at our CIs to ensure that we are really using VS.

StrikerRUS: Hmm, just an idea. Maybe it makes sense to disable auto-fallback to MinGW at our CI by env. variable or by simply removing lines from the build script like in building for docs?

https://github.com/microsoft/LightGBM/blob/8c335352c5dc54fb4c223e16a2b6df9448ca9da7/build_r.R#L72
https://github.com/microsoft/LightGBM/blob/8c335352c5dc54fb4c223e16a2b6df9448ca9da7/docs/conf.py#L242

.vsts-ci.yml
```diff
@@ -119,6 +119,9 @@ jobs:
   strategy:
     maxParallel: 3
     matrix:
+      rpkg:
+        TASK: r-package
+        R_WINDOWS_VERSION: 3.6.3
```
jameslamb, line 124: I moved this over from #2936 . It'll work once that PR is merged, and that PR should definitely be merged before we do this one.

jameslamb:
note to self: per https://github.com/microsoft/LightGBM/pull/2936#discussion_r410953032, I think that in CI we need  to add `"$env:R_LIB_PATH/Rtools/mingw_64/bin;` to `$env:PATH`, so that `gendef` and `dlltool`  are found.

If it isn't in our installation instructions already, we should make it clear there that `Rtools/mingw_*/` needs to be on your path for Visual Studio installation to work.

jameslamb:
Now that #2936, I'll get this updated tonight and add CI support for Visual Studio builds. I'll request new reviews after that.

R-package/src/install.libs.R
jameslamb: @StrikerRUS Ok now that #2936 and #2963 have been merged, we can come back to this!

I think it's a good idea to have something more reliable that us manually looking at the logs to be sure our CI is using the right compiler.

I have a counter-proposal...instead of introducing more complexity into `install.libs.R` (code that our users run), I think we should put that complexity `test_r_package_windows.ps1` (code that only matters to us in CI).

What do you think about this addition I just made in https://github.com/microsoft/LightGBM/pull/2965/commits/096edeaf919986d84fe4e4195c8e29c08103bb87?

```powershell
# Checking that we actually got the expected compiler. The R package has some logic
# to fail back to MinGW if MSVC fails, but for CI builds we need to check that the correct
# compiler ws used.
if ($env:COMPILER -ne "MINGW") {
  $checks = Select-String -Path $INSTALL_LOG_FILE_NAME -Pattern 'Check for working CXX compiler.*MSVC'
} else {
  $checks = Select-String -Path $INSTALL_LOG_FILE_NAME -Pattern 'Check for working CXX compiler.*mingw'
}

if ($checks.Matches.length -ne 0) {
  Write-Output "The wrong compiler was used. Check the build logs."
  Check-Output $False
}
```

R-package/src/install.libs.R
jameslamb: fixed in https://github.com/microsoft/LightGBM/pull/2965/commits/096edeaf919986d84fe4e4195c8e29c08103bb87

R-package/src/install.libs.R
jameslamb: It seems that this isn't working though...I see that [this build which found g++ from chocolatey](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=5993&view=logs&j=9f61088a-36ee-57e4-0947-03329fd6ee90&t=2528b315-ee14-54c7-8330-fcfcaf0cbb8e) still passed

StrikerRUS: I like your idea! Have you managed to fix compiler checks?

StrikerRUS:
@jameslamb Please remove all speedups (`if`s in CI scripts) for MSVC and run full `--as-cran` test according to our plan (https://github.com/microsoft/LightGBM/pull/2936#discussion_r410927991), so that we will be able to check how much time it takes to run after all other fixes were merged.

jameslamb:
> @jameslamb Please remove all speedups (`if`s in CI scripts) for MSVC and run full `--as-cran` test according to our plan ([#2936 (comment)](https://github.com/microsoft/LightGBM/pull/2936#discussion_r410927991)), so that we will be able to check how much time it takes to run after all other fixes were merged.

@StrikerRUS I will do that, but I'm pretty confident we're going to decide  not to use `--as-cran`. Even with all the `ifs`, it's taking 26 minutes right now :/

https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=5994&view=logs&j=9f61088a-36ee-57e4-0947-03329fd6ee90&t=2528b315-ee14-54c7-8330-fcfcaf0cbb8e

Toggling the timestamps, I can see 22 minutes of that is just the install:

image

I really think it might be that `/MP` flag to `MSVC` isn't being used. I'm going to experiment a big on my fork to not take up our CI capacity. Will report back.

R-package/src/install.libs.R
jameslamb: Ok yep they are fixed! Here's a commit on my fork where I intentionally made R fail back to MinGW. You can see that installation works correctly but that `chocolatey`'s `g++` gets used, and that the build fails.

image

build: https://dev.azure.com/JayLamb20/jameslamb-lightgbm/_build/results?buildId=404&view=logs&jobId=9f61088a-36ee-57e4-0947-03329fd6ee90&j=9f61088a-36ee-57e4-0947-03329fd6ee90&t=2528b315-ee14-54c7-8330-fcfcaf0cbb8e

commit: https://github.com/jameslamb/LightGBM/pull/21/commits/cf6333c7a300d388c13f9fa5091ff01de328cc49

jameslamb:
I have been trying on my fork and cannot get the build times for R + MSVC under 24 minutes. They run very fast (`Rscript build_r.R` runs in 1-2 minutes) on my laptop, where I have Visual Studio 16 2019 installed but I never actually use it as an IDE, so it still has all of the default settings.

I have narrowed down the slowness with a lot of print debugging...it takes 10 minutes between when this line appears in the log

>  Build files have been written to: C:/Users/VssAdministrator/AppData/Local/Temp/RtmpWAMlj1/R.INSTALL32c6d499da/lightgbm/src/build

and when compilation starts. Even for the test project that has no source files and whose `CMakeLists.txt` is just one `PROJECT()` call.

That means `system(paste0(cmake_cmd, ..))` is taking 10 minutes to complete on Azure DevOps. That command is running the Visual Studio generator, e.g. `cmake -G"Visual Studio 15 2017" ...`.

I've at least narrowed it down to that, which rules out build-time flags like `/MP` as a root cause.

Theories I tried which were not correct:

* Adding an explicit timeout in `system()` calls from R
* Setting `CMAKE_BUILD_TYPE` to `"Release"` ([link 1](https://cmake.org/pipermail/cmake/2019-March/069195.html) | [link 2](https://cmake.org/cmake/help/latest/variable/CMAKE_CONFIGURATION_TYPES.html#variable:CMAKE_CONFIGURATION_TYPES))

StrikerRUS:
@jameslamb 
>  That means `system(paste0(cmake_cmd, ..))` is taking 10 minutes to complete on Azure DevOps. That command is running the Visual Studio generator, e.g. `cmake -G"Visual Studio 15 2017" ...`.
I've at least narrowed it down to that, which rules out build-time flags like `/MP` as a root cause.

What about `--as-cran` option? It adds ~4min to 24min build time, right?

StrikerRUS:
Hmm, I've noticed that current R builds with MinGW on Windows have interesting variance:
 - 18min https://lightgbm-ci.visualstudio.com/lightgbm-ci/_build/results?buildId=6004&view=results
- 13min https://lightgbm-ci.visualstudio.com/lightgbm-ci/_build/results?buildId=6003&view=results
- 22min https://lightgbm-ci.visualstudio.com/lightgbm-ci/_build/results?buildId=5995&view=results
- 11min https://lightgbm-ci.visualstudio.com/lightgbm-ci/_build/results?buildId=5989&view=results
- 12min https://lightgbm-ci.visualstudio.com/lightgbm-ci/_build/results?buildId=5988&view=results

jameslamb:
> Hmm, I've noticed that current R builds with MinGW on Windows have interesting variance:
> 
> * 18min https://lightgbm-ci.visualstudio.com/lightgbm-ci/_build/results?buildId=6004&view=results
> * 13min https://lightgbm-ci.visualstudio.com/lightgbm-ci/_build/results?buildId=6003&view=results
> * 22min https://lightgbm-ci.visualstudio.com/lightgbm-ci/_build/results?buildId=5995&view=results
> * 11min https://lightgbm-ci.visualstudio.com/lightgbm-ci/_build/results?buildId=5989&view=results
> * 12min https://lightgbm-ci.visualstudio.com/lightgbm-ci/_build/results?buildId=5988&view=results

Taking a quick look at those, if you click `Toggle Timestamps` on the log, it seems like a lot of the variability is about installing `MiKTeX` and friends upstream of the actual build of the R package.

`6004` (7 minutes):

image

`6003` (3 minutes)

image

This is kind of a manual and painful process though, I'm sure if we looked across all builds we'd find other sources of variability. Maybe I'll try writing a script tonight that does that, so I know where to focus my effort.

StrikerRUS:
Nice investigation! But I think that the main goal now is to try to minimize actual building time of R package.

Can you please clarify that?
https://github.com/microsoft/LightGBM/pull/2965#issuecomment-624055496

jameslamb:
> Nice investigation! But I think that the main goal now is to try to minimize actual building time of R package.
> 
> Can you please clarify that?
> [#2965 (comment)](https://github.com/microsoft/LightGBM/pull/2965#issuecomment-624055496)

right sorry, let me clarify:

1. The R package takes 1-2 minutes to build on my laptop with Visual Studio
2. The R package takes around 20 minutes to build in CI
3. I see in the logs that it takes 10 minutes for each call to `system(cmake_cmd)` to complete in the Azure CI environment, and that is unrelated to the LightGBM project specifically. Since it takes 10 minutes to build a fake project where the `CMakeLists.txt` file is just one `PROJECT()` call.
4. Removing `--as-cran` and ignoring all the documentation-related things in CI could save about 4 minutes.

Things I tried to fix this, which did not work:

1. Explicitly setting build type to `Release` so the other three types of Visual Studio builds would be ignored.
2. Putting a `timeout` on `system()`

Things I'm going to try next:

1. Running Visual Studio builds on AppVeyor instead
2. Finding a way to detect Visual Studio in `install.libs.R` that does not involve running `cmake -G`

jameslamb:
There is a lot of information I'm reading through in this answer: https://stackoverflow.com/a/37327661

jameslamb:
The situation looks even worse on AppVeyor. I cancelled this build after 34 minutes 😱 

https://ci.appveyor.com/project/jameslamb/lightgbm/builds/32675576

I ran again with `Rscript build_r.R` (so we could directly see the logs instead of them being hidden in R CMD CHECK):

https://ci.appveyor.com/project/jameslamb/lightgbm/builds/32675702

You can see from the timestamps on the logs there that it takes about 2 and a half minutes to get to the point where `install.libs.R` is run....and then 25 minutes later the build hasn't progressed at all.

[log (1).txt](https://github.com/microsoft/LightGBM/files/4584578/log.1.txt)

jameslamb:
Ok I might be making progress! To investigate my theory that this is about disk I/O (since `cmake -G"Visual Studio..."` writes a lot of files), I ran the following Powershell command:

```powershell
Get-CimInstance Win32_PerfRawData_PerfDisk_LogicalDisk
```

| environment|  Disk | DiskWriteBytesPersec  |  DiskWritesPersec |
|---|---|---|---|
|AppVeyor   |  C: | 208896 |  11  |
| Azure |  C: |  565897728 |  8787 |
|  Azure |   D:|  43876352 |   1225|
| my laptop| C: | 88364674560 | 3343288|
| my laptop| D: |364544 | 85 |

NOTE: There is no `D:` available on AppVeyor.

AppVeyor log: https://ci.appveyor.com/project/jameslamb/lightgbm/builds/32676579?fullLog=true

Azure DevOps log: https://dev.azure.com/JayLamb20/jameslamb-lightgbm/_build/results?buildId=421&view=logs&j=9f61088a-36ee-57e4-0947-03329fd6ee90&t=2528b315-ee14-54c7-8330-fcfcaf0cbb8e

This is exciting, and I think it's good evidence of what's happening!

* `D:` on the VM we're getting from Azure DevOps is around 200x slower than the `C:` disk on my laptop. That explains the difference in performance between my laptop and Azure DevOps. `D:` is where we install R and where the project build directory is by default on Azure DevOps.
* `C:` on the VM we're getting for AppVeyor is around 210x slower than `D:` on Azure DevOps, which explains why the builds get stuck at `cmake -G` there and never even finish

I'm going to see if moving to `C:\` on Azure DevOps is possible and if it's faster.

**but wait...why hasn't this been a problem for Python builds?**

I don't think we ever test using the Visual Studio generator (`cmake -G"Visual Studio..."`) for the Python package! Our builds use the pre-made `MSBuild` solution file in https://github.com/microsoft/LightGBM/tree/master/windows

https://github.com/microsoft/LightGBM/blob/18c706dc840c72341a2a85e05dfb5fad6366902a/python-package/setup.py#L136-L148

jameslamb:
# Short Summary

Disk throughput does the difference between install times on CI and my laptop, but moving from `D:\` to `C:\` on Azure DevOps builds doesn't help our build times.

## How I reached this conclusion

Based on https://github.com/microsoft/LightGBM/pull/2965#issuecomment-624421092, I tried adding some logic that says "if you are on Azure, work in `C:\` not `D:\`". I expected a small speedup from this since any writes to disk should be 10x faster.

I tested [on my fork](https://github.com/microsoft/LightGBM/pull/2965#issuecomment-624421092). I tried 5 builds with copying to `C:\` and 5 without. The times for the R package build are reported below.

All builds below are already using the configuration of `R CMD check` that skips all documentation.

## Test Results

**with** ([build link](https://dev.azure.com/JayLamb20/jameslamb-lightgbm/_build/results?buildId=431&view=results))

*median: 26m22s*

image

**without**

*median: 25m53s*

image

## Conclusion

1. Yes, disk throughput is a lot lower in our CI environments than on my laptop (it is not a very good laptop, by the way), and that is causing `cmake -G"Visual Studio..."` to take around 10 minutes to run.
2. That command is run twice, once to figure out which version (if any) of Visual Studio is installed and once to actually build LightGBM. Because [finding installed versions of Visual Studio has changed from version to version](https://github.com/microsoft/vswhere), I don't think it would be easy to eliminate the first `cmake -G` call used to determine if Visual Studio is working.
3. Moving all the files to `C:\` at the beginning of an Azure builds doesn't improve the build times. That operation has a cost, and might even be slightly _slowing down_ builds.

## Next Step

It _might_ be possible to cut one of the `cmake -G` calls, but it will increase the complexity of `install.libs.R`, which is already somewhat complex. If `.find_vs_version()` ran `cmake -G` on LightGBM itself (not a test project), we could skip running it again before building.

This is the only other idea I have, so I'll try that tomorrow. If it doesn't work or we decide it introduces too much complexity, I think 25-minute builds will be something we have to live with to prevent things like #2995 and #2997 from re-emerging.

jameslamb:
# Got it down to 16 minutes!

I know this PR has a lot of long comments, I apologize. Wanted to document the investigation as I went, since it reveals a lot about how we're building the R package and the project in general.

I now have the MSVC builds down to 16 minutes!

## What Changed?

I tried this idea from https://github.com/microsoft/LightGBM/pull/2965#issuecomment-624461306

> It might be possible to cut one of the cmake -G calls, but it will increase the complexity of install.libs.R, which is already somewhat complex. If .find_vs_version() ran cmake -G on LightGBM itself (not a test project), we could skip running it again before building.

I did [5 builds](https://dev.azure.com/JayLamb20/jameslamb-lightgbm/_build/results?buildId=433&view=results)

*median: 15m49s*
image

StrikerRUS:
@jameslamb Your efforts are overwhelming! Many, many thanks!
I believe that I/O operations was the root cause because at CI drivers are actually network disks and they are suffering like all other remote things during pandemic (or maybe CI are limiting network bandwidth manually now).

> Got it down to 16 minutes!

This is amazing result! I think we can now drop all our complexity for not checking docs because 4 mins look not so significant thing but allow us to be sure that **everything** is working fine.

StrikerRUS:
> but wait...why hasn't this been a problem for Python builds?
I don't think we ever test using the Visual Studio generator (cmake -G"Visual Studio...") for the Python package! Our builds use the pre-made MSBuild solution file in 

This is true only for Appveyor builds.

https://github.com/microsoft/LightGBM/blob/18c706dc840c72341a2a85e05dfb5fad6366902a/.ci/test_windows.ps1#L55-L62

```
...
running install
INFO:LightGBM:Starting to compile the library.
INFO:LightGBM:Starting to compile with MSBuild from existing solution file.
running build
running build_py
creating build
...
```

At Azure we use CMake directly:

https://github.com/microsoft/LightGBM/blob/18c706dc840c72341a2a85e05dfb5fad6366902a/.ci/test_windows.ps1#L28-L32

Moreover, in `bdist` job where it is theoretically possible to use MSBuild tool and solution file, Azure fails back to CMake because there is no MSBuild in PATH variable:
```
...
running install
INFO:LightGBM:Starting to compile the library.
INFO:LightGBM:Starting to compile with MSBuild from existing solution file.
WARNING:LightGBM:Compilation with MSBuild from existing solution file failed.
INFO:LightGBM:Starting to compile with Visual Studio 16 2019.
INFO:LightGBM:Starting to compile with Visual Studio 15 2017.
running install_lib
...
```

In addition, in `sdist` job we are compiling LightGBM with CMake twice! First time from sdist Python artifact itself
https://github.com/microsoft/LightGBM/blob/18c706dc840c72341a2a85e05dfb5fad6366902a/.ci/test_windows.ps1#L38-L39
and second for SWIG artifact
https://github.com/microsoft/LightGBM/blob/18c706dc840c72341a2a85e05dfb5fad6366902a/.ci/test_windows.ps1#L47
So, this is very weird that Python builds didn't encounter I/O problems...

StrikerRUS:
>So, this is very weird that Python builds didn't encounter I/O problems...

Is it possible that `system()` call itself is very costly in R scripts?..

jameslamb:
> > So, this is very weird that Python builds didn't encounter I/O problems...
> 
> Is it possible that `system()` call itself is very costly in R scripts?..

It is possible, but I don't know. The other thing we do differently in R MSVC builds is creating the `R.lib` and then linking to it. That could make it more expensive than the otherwise-equivalent Python build.

StrikerRUS:
> The other thing we do differently in R MSVC builds is creating the R.lib and then linking to it. That could make it more expensive than the otherwise-equivalent Python build.

Yeah, of course! But you've said that building of an empty project takes a lot of time.

jameslamb:
> > The other thing we do differently in R MSVC builds is creating the R.lib and then linking to it. That could make it more expensive than the otherwise-equivalent Python build.
> 
> Yeah, of course! But you've said that building of an empty project takes a lot of time.

ha oh right! Excellent point.

jameslamb:
> > > The other thing we do differently in R MSVC builds is creating the R.lib and then linking to it. That could make it more expensive than the otherwise-equivalent Python build.
> > 
> > 
> > Yeah, of course! But you've said that building of an empty project takes a lot of time.
> 
> ha oh right! Excellent point. Tonight I'll try on my fork to compare the time of these two:
> 
> * directly running `cmake -G"Visual Studio ..."` on a test project
> * running the same command via `Rscript -e "system('cmake ...')"`
> 
> Then we'll know if `system()` is introducing a lot of overhead

WOW it does seem like `system()` is introducing a lot of overhead :face_with_thermometer: 

I tried some builds with a `CMakeLists.txt` that just has this content:

```cmake
project(nonsense)
```

Running `cmake -G"Visual Studio 15 2017" -A x64 ..` directly, it takes about 1 minute to build the project Makefiles. For various versions of calling this from R, it takes 10 minutes.

Right now, I don't have any ideas how to avoid calling `cmake` from R code. I think the convenience of having an `Rscript build_r.R` is worth it, and I'd rather do that than telling people "ok first you have to run this `cmake` command from a terminal, then run this other R script and be sure you are in the right working directory".

Things I tried from R:

* `system(cmake_cmd)`
* `system2(cmake_cmd, stdout = "", stderr = "")`
* `system2(cmake_cmd, stdout = FALSE, stderr = FALSE)`
* `system2(cmake_cmd)`
* `shell(cmake_cmd)`

**without `system()`** 

* [build link](https://dev.azure.com/JayLamb20/jameslamb-lightgbm/_build/results?buildId=439&view=results)
* median time: 1m7s

image

**with `system(cmake_cmd)`*

* [build link](https://dev.azure.com/JayLamb20/jameslamb-lightgbm/_build/results?buildId=441&view=results)
* median time: roughly 10 minutes
    - *(these builds include an extra 2.5 minutes to install R)*

image

**with `system2(cmake_cmd, stdout = "", stderr = "")`**

* [build link](https://dev.azure.com/JayLamb20/jameslamb-lightgbm/_build/results?buildId=444&view=results)
* median time: roughly 10 minutes

image

**with `system2(cmake_cmd, stdout = FALSE, stderr = FALSE)`**

* [build link](https://dev.azure.com/JayLamb20/jameslamb-lightgbm/_build/results?buildId=445&view=results)
* median time: roughly 10 minutes

image

**with `system2(cmake_cmd)`, no arguments to `stdout` and `stderr`**

* [build link](https://dev.azure.com/JayLamb20/jameslamb-lightgbm/_build/results?buildId=446&view=results):
* median time: roughly 10 minutes

image

**with `shell(cmake_cmd)`**

* [build link](https://dev.azure.com/JayLamb20/jameslamb-lightgbm/_build/results?buildId=449&view=results)
* median time: roughly 10 minutes

image

StrikerRUS:
> WOW it does seem like system() is introducing a lot of overhead 🤒

Huh! interesting! More than x10 slowdown 😢 

If you asked me how had I come to this assumption, I remembered your words about that `system` calls work in isolated session: https://github.com/microsoft/LightGBM/pull/2837#issuecomment-592538973.

However, I strongly believe that there is some workaround for our problem! Is there something similar to Python's [`subprocess`](https://docs.python.org/3/library/subprocess.html) module in R?

StrikerRUS:
According to the one of the logs, it seems that 10 minutes is passed between the last log message from CMake and `print(exitCode)` command. wat??

So, it finishes writing project files and then hangs for 10 minutes, right?..

image

jameslamb:
> So, it finishes writing project files and then hangs for 10 minutes, right?..
> 

technically yes. I think there  are other things `cmake` does after that message  "Build files have been written", but they aren't reflected in the logs even with  `--verbose`.

That is exactly the pattern I've seen...those two lines in the log are how I narrowed it  down to  the  `cmake -G` call.

>  I strongly believe that there is some workaround for our problem! Is there something similar to Python's subprocess module in R?

Not that I'm aware of!  Everything I've searched so far points back to `system()`, `system2()`, and `shell()`. I'll look a bit more.

jameslamb:
I just tried putting the command in a `.bat` file and then calling the script from R, just to see if adding one more level of separation would help (I wasn't hopeful).

Can confirm that didn't help either:

image

StrikerRUS:
> I'll look a bit more.

Can you please transfer your forked branch here in our repo please? I'd like to get my hands dirty 🙂 .

jameslamb:
> > I'll look a bit more.
> 
> Can you please transfer your forked branch here in our repo please? I'd like to get my hands dirty 🙂 .

Sure, just pushed LightGBM branch `testing-vs-install`. It's in an ugly state since I'm commenting out a lot of things and throwing errors to focus only on this experiment.

Next things I'm trying:

* `sys`: https://github.com/jeroen/sys
* `processx`: https://github.com/r-lib/processx

I also created a test project there in a directory. See `testing-dir/`  from the root of the repo.

image

jameslamb:
Currently with `processx`, I can see that `cmake` is able to tell that Visual Studio 15 2017 is installed, but can't actually generate makefiles.

image

But the same works with `system()`, `system2()`, etc.

I hate software

jameslamb:
`processx` looks really promising!!! 

You can see here that it's already made it into compilation after only 3 minutes 😀 

image

jameslamb:
Ok I got pretty far with `processx` on my fork tonight (thanks to @alistaire47 for the tip!!). I've pushed my recent changes to this branch.

Tomorrow I need to investigate more...I think I am getting different performance from `Rscript build_r.R` and from the installation inside of `R CMD check`. Since @Laurae2 and I have already talked about Visual Studio builds not being supported in CRAN builds in the future (it is not part of CRAN toolchain), if I find tomorrow that that is true then we might consider skipping `R CMD check` entirely and doing this for MSVC builds:

```r
Rscript build_r.R
cd R-package/tests
Rscript testthat.R
```

...but let's see what I find tomorrow.

I've removed the `awaiting review` label and added `in progress` back while we figure this out.

StrikerRUS:
> Ok I got pretty far with `processx` on my fork tonight ...

This is wonderful! Impressive 3 mins! 🎉 

I tried to play with your test setup in `testing-vs-install` branch, but unfortunately made no progress. The same result as yours ~12mins.

I think `processx` with a fallback to slow `system` calls is very good solution. We get fast CI builds and users don't see any difference comparing with their previous experience for LightGBM installation. Only one thought: does it make sense to try `processx` at Linux and macOS? Maybe we should limit our workaround only for Windows?

jameslamb:
> I tried to play with your test setup in `testing-vs-install` branch, but unfortunately made no progress. The same result as yours ~12mins.

My next suspicion is that `R CMD check` is introducing some additional overhead. My 3-minute example was a direct `Rscript build_r.R`. I'll test that tonight.

> 
> I think `processx` with a fallback to slow `system` calls is very good solution. ... does it make sense to try `processx` at Linux and macOS? Maybe we should limit our workaround only for Windows?

I think we should limit to Windows for now, just in the spirit of limiting how many things we can possibly break with one PR. If people complain about build time for the R package in the future, then we'll know it's something we can test.

jameslamb:
Ok @StrikerRUS I just pushed a commit I think we should discuss. On my fork, I was able to get the MSVC R package builds down **under 5 minutes**!

image

two changes made this possible:

1. Use  `processx` everywhere (in `R-package/src/install.libs.R` and `build_r.R`)
2. Don't run `R CMD check` for MSVC builds.

`R CMD check`  is still introducing some overheard we can't code around. I found that I could get the R + `MSVC` builds with full `R CMD check` down to about 14 minutes (with `1)` above) but no quicker than that.

As of https://github.com/microsoft/LightGBM/pull/2965/commits/f4f8fd152c5c6d73a127a8ec39e7fd2258b79f2d my proposal is to run this more limited build for MSVC R builds. It does these two things:

```shell
Rscript build_r.R
cd R-package/tests/testthat.R
```

That's a very reasonable approximation of what our users face...installing with `Rscript build_r.R` and then running the code similar to the way it is run in the tests. That will catch a lot of user-facing issues in the library. I think it is ok to forgo the other `R CMD check` things because:

1. We are running full `R CMD check` on Mac, Linux, and Windows (`MinGW` builds) already, so almost all problems will be caught
2. We are never going to support installation from CRAN with Visual Studio ([since the CRAN toolchain for Windows only uses MinGW](https://github.com/r-windows/docs/blob/master/faq.md#what-is-rtools)), so if the MSVC-specific version of the library violates some CRAN rule it doesn't matter.

One other thing to consider....if we use this limited very-quick build, we could add a second MSVC R build to CI with the [`windows-2019` image from Azure DevOps](https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops). That image uses Visual Studio 16 2019, so then we'd be able to cover the previous two major releases of Visual Studio!

.ci/test_r_package_windows.ps1
```diff
@@ -64,45 +67,67 @@ if ($env:COMPILER -eq "MINGW") {
...
+  Rscript testthat.R ; Check-Output $?
+}
 
-Write-Output "Looking for issues with R CMD check results"
-if (Get-Content "$LOG_FILE_NAME" | Select-String -Pattern "WARNING" -Quiet) {
-    echo "WARNINGS have been found by R CMD check!"
-    Check-Output $False
+# Checking that we actually got the expected compiler. The R package has some logic
+# to fail back to MinGW if MSVC fails, but for CI builds we need to check that the correct
+# compiler ws used.
```
jameslamb, line 121:
```suggestion
# compiler was used.
```

StrikerRUS:
@jameslamb 
> On my fork, I was able to get the MSVC R package builds down under 5 minutes!

Unbelievable! Great job! 👍 

> As of f4f8fd1 my proposal is to run this more limited build for MSVC R builds.

Hmm, OK. As MSVC won't be used at CRAN we shouldn't care about any failed checks from it. Sounds reasonable.

In addition, more covered use cases (installation by `R CMD check` and `Rscript build_r.R`) are always better! It's like we build Python package via `python setup.py install` at Appveyor, `pip install *.whl`, `python setup.py install --precompile` and so on at Azure.

> One other thing to consider....if we use this limited very-quick build, we could add a second MSVC R build to CI with the windows-2019 image from Azure DevOps. 

OK, agree to add one more CI job with another VS version. But I think it should be Appveyor as Azure is already loaded with 10+ jobs. And it will allow to test not only another VS but another environment. WDYT?

jameslamb:
> OK, agree to add one more CI job with another VS version. But I think it should be Appveyor as Azure is already loaded with 10+ jobs. And it will allow to test not only another VS but another environment. WDYT?

Yep I like that! Trying [on AppVeyor right now](https://ci.appveyor.com/project/jameslamb/lightgbm/builds/32754238), let's see how it goes

image

.appveyor.yml
```diff
@@ -1,6 +1,8 @@
-﻿version: 2.3.2.{build}
+version: 2.3.2.{build}
```
jameslamb, line 1: ugh this line-ending change is back. I will try to revert it.

jameslamb:
> > OK, agree to add one more CI job with another VS version. But I think it should be Appveyor as Azure is already loaded with 10+ jobs. And it will allow to test not only another VS but another environment. WDYT?
> 
> Yep I like that! Trying [on AppVeyor right now](https://ci.appveyor.com/project/jameslamb/lightgbm/builds/32754238), let's see how it goes
> 
> image

ok took a bit of work, but it's working!

image

So as of this PR, based on the conversation that ended in https://github.com/microsoft/LightGBM/pull/2965#issuecomment-626078869, we'll have this coverage for the R package:

| Operating system | compiler | install | test | CI provider |
|--------------|------------|--------------|--------|----|
| Windows      | MinGW       | `R CMD CHECK`  | `R CMD CHECK --as-cran` | AppVeyor |
| Windows      | MSVC (VS 2019) | `build_r.R`           | `testthat.R`| AppVeyor |
| Windows       | MSVC (VS 2017) | `R CMD CHECK`  | `R CMD CHECK --as-cran` | Azure |
| MacOS          | `clang`            | `R CMD CHECK` | `R CMD CHECK --as-cran` | Azure|
| MacOS          | `gcc`            | `R CMD CHECK` | `R CMD CHECK --as-cran` | Travis|
| Linux             | `clang`          | `R CMD CHECK` | `R CMD CHECK --as-cran` | Travis |
| Linux             | `gcc`          | `R CMD CHECK` | `R CMD CHECK --as-cran` | Azure |

I think that will put the R package in a pretty good place!

.appveyor.yml
jameslamb: fixed it in https://github.com/microsoft/LightGBM/pull/2965/commits/c53cbcad8beeafc6e0228579c048da2f7c668a08

image

R-package/src/install.libs.R
```diff
@@ -41,71 +111,65 @@ if (!use_precompile) {
...
-  cmake_cmd <- "cmake "
-  build_cmd <- "make _lightgbm"
+  cmake_args <- NULL
+  build_cmd <- "make"
+  build_args <- "_lightgbm"
   lib_folder <- file.path(source_dir, fsep = "/")
 
   if (use_gpu) {
-    cmake_cmd <- paste0(cmake_cmd, " -DUSE_GPU=ON ")
+    cmake_args <- c(cmake_args, "-DUSE_GPU=ON")
```
jameslamb, line 120: switching to a separation of `command` and `args` here because `processx::run()` has that interface. It's easier to keep them separate and jam them back into one string for `system()` then to keep building up one string and break it back apart for `processx::run()`, in my opinion.

R-package/src/install.libs.R
```diff
@@ -41,71 +111,65 @@ if (!use_precompile) {
...
   # Check if Windows installation (for gcc vs Visual Studio)
   if (WINDOWS) {
     if (use_mingw) {
       print("Trying to build with MinGW")
-      cmake_cmd <- paste0(cmake_cmd, " -G \"MinGW Makefiles\" ")
-      build_cmd <- "mingw32-make.exe _lightgbm"
-      system(paste0(cmake_cmd, " ..")) # Must build twice for Windows due sh.exe in Rtools
+      # Must build twice for Windows due sh.exe in Rtools
+      cmake_args <- c(cmake_args, "-G", shQuote("MinGW Makefiles"))
+      .run_shell_command("cmake", c(cmake_args, ".."), strict = FALSE)
```
jameslamb, line 146: I know the `strict = FALSE` looks a little strange here. I was surprised to learn that right now (on `master`, not specific to this PR), `cmake -G "MinGW Makefiles"` is working but returning a non-zero exit code.

You can see some errors in [the most recent build on `master`](https://ci.appveyor.com/project/guolinke/lightgbm/builds/32728411/job/jpvhc1pkgghirltt)

image

I did see that it might be possible to fix this by making sure mingw is on `PATH` (https://stackoverflow.com/a/32922266/3986677), but when I tried a few versions of that it didn't work.

We can safely ignore this for the sake of this PR. It's not a user-facing issue, since the non-0 exit code is handled internally in this R script and since installation succeeds.

I would like to address it eventually so that it doesn't lead to bug reports or waste users' time when they're debugging. I've created #3060

jameslamb:
Alright I think this is ready for another review! @StrikerRUS @Laurae2 @guolinke I look forward to your comments, thank you!

R-package/src/install.libs.R
StrikerRUS: @jameslamb Maybe it is because we have 2 instances of MinGW in PATH? One comes from RTools and another one is ordinary MinGW-w64 we use to test Python builds:
https://github.com/microsoft/LightGBM/blob/1b4cddfbb233d8483680cc35ab4385887d4e342f/.appveyor.yml#L22

.appveyor.yml
```diff
@@ -1,6 +1,8 @@
 ﻿version: 2.3.2.{build}
 
-image: Visual Studio 2015
+image:
+  - Visual Studio 2015
+  - Visual Studio 2019
```
StrikerRUS, line 5: I think `APPVEYOR_BUILD_WORKER_IMAGE` can be used to simplify config and get rid of `matrix exclude`.

https://www.appveyor.com/docs/build-environment/#using-multiple-images-for-the-same-build

R-package/src/install.libs.R
StrikerRUS: Anyway, I think this line should be executed only for Python builds to prevent weird bugs in the future.

StrikerRUS:
> One other thing to consider....if we use this limited very-quick build, we could add a second MSVC R build to CI with the windows-2019 image from Azure DevOps.

Also, it is one step towards #3024 as we can set another R version there.

.appveyor.yml
jameslamb: WOW I did not know that! Ok let me try, that would be a lot easier to read and manage than using `matrix exclude`

R-package/src/install.libs.R
jameslamb: Ah interesting! I could try moving that into only Python builds.

jameslamb: I just tried this today and I don't think it worked. But I think it's not necessary to fix for t his PR...I'll post my findings on #3060

.ci/test_r_package_windows.ps1
```diff
@@ -14,6 +14,9 @@ function Download-File-With-Retries {
 
 $env:R_WINDOWS_VERSION = "3.6.3"
 $env:R_LIB_PATH = "$env:BUILD_SOURCESDIRECTORY/RLibrary" -replace '[\\]', '/'
+$env:R_LIBS="$env:R_LIB_PATH"
+$env:R_LIBS_SITE="$env:R_LIB_PATH/R/library"
+$env:R_LIBS_USER="$env:R_LIB_PATH/R/library"
```
StrikerRUS, line 19: Are these variables used somewhere? I can't find them down in the script.

```suggestion
$env:R_LIBS = "$env:R_LIB_PATH"
$env:R_LIBS_SITE = "$env:R_LIB_PATH/R/library"
$env:R_LIBS_USER = "$env:R_LIB_PATH/R/library"
```

.ci/test_r_package_windows.ps1
```diff
@@ -64,45 +67,67 @@ if ($env:COMPILER -eq "MINGW") {
...
-if (Get-Content "$LOG_FILE_NAME" | Select-String -Pattern "WARNING" -Quiet) {
-    echo "WARNINGS have been found by R CMD check!"
-    Check-Output $False
+# Checking that we actually got the expected compiler. The R package has some logic
+# to fail back to MinGW if MSVC fails, but for CI builds we need to check that the correct
+# compiler was used.
+if ($env:COMPILER -eq "MSVC") {
+  $checks = Select-String -Path "${INSTALL_LOG_FILE_NAME}" -Pattern 'Check for working CXX compiler.*MSVC'
+} else {
+  $checks = Select-String -Path "${INSTALL_LOG_FILE_NAME}" -Pattern 'Check for working CXX compiler.*mingw'
```
StrikerRUS, line 125: I think it should be checked earlier right after printing these logs (before running `Rscript testthat.R`) to save CI time.

CMakeLists.txt
```diff
@@ -193,7 +193,7 @@ if(MSVC)
         CMAKE_CXX_FLAGS_RELEASE
         CMAKE_CXX_FLAGS_RELWITHDEBINFO
     )
-    SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /W4 /MP")
+    SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /W4 /MP /incremental")
```
StrikerRUS, line 196: ping @guolinke  here

R-package/src/install.libs.R
```diff
@@ -12,7 +12,77 @@ R_ver <- as.double(R.Version()$major) + as.double(R.Version()$minor) / 10.0
...
+    "Visual Studio 16 2019"
+    , "Visual Studio 15 2017"
+    , "Visual Studio 14 2015"
+  )
+  working_vs_version <- NULL
+  for (vs_version in vs_versions) {
+    print(sprintf("Trying '%s'", vs_version))
+    # if the build directory is not empty, clean it
+    if (file.exists("CMakeCache.txt")) {
+      file.remove("CMakeCache.txt")
```
StrikerRUS, line 69: Is it enough to remove only one cache file to reset CMake state?

build_r.R
```diff
@@ -74,7 +104,7 @@ result <- file.copy(
 #       by CMake while also meeting the CRAN req to create object files
 #       on demand
 cmd <- "R CMD build lightgbm_r --keep-empty-dirs"
```
StrikerRUS, line 106: `cmd` is not used anymore?

R-package/src/install.libs.R
jameslamb: yes, I believe so. In the Azure build we use Visual Studio 15 2017 but the first things searched by this code is Visual Studio 16 2019, so I think it's working correctly.

.ci/test_r_package_windows.ps1
jameslamb: ok sure, I can do that

build_r.R
jameslamb: ah you're right we can remove it now

jameslamb: removed in https://github.com/microsoft/LightGBM/pull/2965/commits/828023f8c04002e00bfe66e43069bb383e7e0fea

.ci/test_r_package_windows.ps1
jameslamb: changed in https://github.com/microsoft/LightGBM/pull/2965/commits/828023f8c04002e00bfe66e43069bb383e7e0fea.

I didn't want to copy this code into both branches of the `if`, so my solution for your request was to move `testthat.R` further down the script and guard it with `if ($env:COMPILER -eq "MSVC")...`

build_r.R
jameslamb: Ah whoops, we do actually use it! To print the command when you run with `--skip-install`. I'll put it back.

image

jameslamb: or more accurately...we had two `cmd` in that part of the code and needed to delete one...but I deleted the wrong one! fixed in https://github.com/microsoft/LightGBM/pull/2965/commits/f9e1a5203966c9f8d24df38ebfd2eb054248c699

build_r.R
```diff
@@ -17,12 +17,42 @@ INSTALL_AFTER_BUILD <- !("--skip-install" %in% args)
...
 # system() will not raise an R exception if the process called
-# fails. Wrapping it here to get that behavior
-.run_shell_command <- function(cmd, ...) {
-    exit_code <- system(cmd, ...)
-    if (exit_code != 0L) {
+# fails. Wrapping it here to get that behavior.
+#
+# system() introduces a lot of overhead, at least on Windows,
+# so trying processx if it is available
+.run_shell_command <- function(cmd, args, strict = TRUE) {
```
StrikerRUS, line 24: Is it possible to import this function from `R-package/src/install.libs.R` somehow? I assume this script `build_r.R` will be never run without content from this repo or this use case is weird  at least.

build_r.R
```diff
@@ -88,9 +117,9 @@ version <- gsub(
 )
 tarball <- file.path(getwd(), sprintf("lightgbm_%s.tar.gz", version))
 
-cmd <- sprintf("R CMD INSTALL %s --no-multiarch --with-keep.source", tarball)
 if (INSTALL_AFTER_BUILD) {
-  .run_shell_command(cmd)
+  .run_shell_command("R", c("CMD", "INSTALL", tarball, "--no-multiarch", "--with-keep.source"))
 } else {
+  cmd <- sprintf("R CMD INSTALL %s --no-multiarch --with-keep.source", tarball)
```
StrikerRUS, line 123: I believe we should save `c("CMD", "INSTALL", tarball, "--no-multiarch", "--with-keep.source")` in a variable and reuse it in these two commands.

build_r.R
jameslamb: There is not an R equivalent of importing one function from a module the way you can in Python.

R has the concept of running ALL of a single file with  `source()`, and I did consider that. But I think we should avoid it. I would rather have this code duplicated in two places than introduce more indirection and complexity.

We have CI to catch issues, so I think `.run_shell_command()`  is a case where copying code is better than having shared code.

R-package/src/install.libs.R
```diff
@@ -41,71 +111,65 @@ if (!use_precompile) {
...
     }
+  } else {
+      .run_shell_command("cmake", c(cmake_args, ".."))
   }
 
-  # Install
-  system(paste0(cmake_cmd, " .."))
+  # generate build files
+  if (!makefiles_already_generated) {
+    .run_shell_command("cmake", c(cmake_args, ".."), strict = FALSE)
```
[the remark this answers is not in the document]
jameslamb: I think you're right. @Laurae2 was explaining that to me, forgot to come back and change this one

.ci/test_r_package_windows.ps1
```diff
@@ -64,45 +65,70 @@ if ($env:COMPILER -eq "MINGW") {
...
 
-Check-Output $check_succeeded
+# Checking that we actually got the expected compiler. The R package has some logic
+# to fail back to MinGW if MSVC fails, but for CI builds we need to check that the correct
+# compiler was used.
+if ($env:COMPILER -eq "MSVC") {
+  $checks = Select-String -Path "${INSTALL_LOG_FILE_NAME}" -Pattern 'Check for working CXX compiler.*MSVC'
+} else {
+  $checks = Select-String -Path "${INSTALL_LOG_FILE_NAME}" -Pattern 'Check for working CXX compiler.*mingw'
+}
```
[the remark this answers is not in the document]
jameslamb: this will not work. `COMPILER` will be "MINGW" (all caps) for `MINGW` builds and this check will fail.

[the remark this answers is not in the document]
StrikerRUS: By default `Select-String` is case insensitive. While you are not passing `-CaseSensitive` option it should work. Refer to https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/select-string?view=powershell-7.

build_r.R
StrikerRUS: > There is not an R equivalent of importing one function from a module the way you can in Python.

Okay 🙁

build_r.R
jameslamb: makes sense! I can change that

.ci/test_r_package_windows.ps1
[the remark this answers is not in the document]
jameslamb: ohhhhhhhhhhhh I did not know that. That makes it different from every other command line pattern-matching tool I've ever used 😂 

ok then yes I like this

CMakeLists.txt
jameslamb: Since it affects more than just the R package and only provides a small time savings, I think it should be removed and put in a separate PR for discussion.

build_r.R
jameslamb: updated in https://github.com/microsoft/LightGBM/pull/2965/commits/313827e90be65feafafac26023f75e5308e4e1c7

R-package/src/install.libs.R
```diff
@@ -12,7 +12,77 @@ R_ver <- as.double(R.Version()$major) + as.double(R.Version()$minor) / 10.0
...
+# try to generate Visual Studio build files
+.generate_vs_makefiles <- function(cmake_args) {
+  vs_versions <- c(
+    "Visual Studio 16 2019"
+    , "Visual Studio 15 2017"
+    , "Visual Studio 14 2015"
+  )
+  working_vs_version <- NULL
+  for (vs_version in vs_versions) {
+    print(sprintf("Trying '%s'", vs_version))
```
Laurae2, line 66: `message` instead of `print`

R-package/src/install.libs.R
```diff
@@ -12,7 +12,77 @@ R_ver <- as.double(R.Version()$major) + as.double(R.Version()$minor) / 10.0
...
+    vs_cmake_args <- c(
+      cmake_args
+      , "-G"
+      , shQuote(vs_version)
+      , "-A"
+      , "x64"
+    )
+    exit_code <- .run_shell_command("cmake", c(vs_cmake_args, ".."), strict = FALSE)
+    if (exit_code == 0L) {
+      print(sprintf("Successfully created build files for '%s'", vs_version))
```
Laurae2, line 80: `message` instead of `print`

jameslamb: Thanks! I'll make the `print()` --> `message()` changes in this PR.

jameslamb: I replaced all the `print()` with `message()` (and one `warning()`) here: https://github.com/microsoft/LightGBM/pull/2965/commits/a1d2596e8f83fde5dbf832ac660095ae60bff283

R-package/src/install.libs.R
jameslamb: updated in https://github.com/microsoft/LightGBM/pull/2965/commits/a1d2596e8f83fde5dbf832ac660095ae60bff283

R-package/src/install.libs.R
```diff
@@ -41,71 +111,65 @@ if (!use_precompile) {
...
-        }
-      }
-      if (try_vs == 1L) {
-        print("Building with Visual Studio failed. Attempted with MinGW")
-        cmake_cmd <- paste0(cmake_cmd, " -G \"MinGW Makefiles\" ")
-        system(paste0(cmake_cmd, " ..")) # Must build twice for Windows due sh.exe in Rtools
-        build_cmd <- "mingw32-make.exe _lightgbm"
+      visual_studio_succeeded <- .generate_vs_makefiles(cmake_args)
+      if (!isTRUE(visual_studio_succeeded)) {
+        message("Building with Visual Studio failed. Attempting with MinGW")
```
StrikerRUS, line 152: Maybe `warning` here as users didn't ask to build with MinGW explicitly and expect good CPU utilization by building via MSVC?

jameslamb: I like that
```suggestion
        warning("Building with Visual Studio failed. Attempting with MinGW")
```

jameslamb:
> Something went wrong during merging branches.

I'm very confused about what could have happened. Trying to figure it out. I don't understand why changes like #3078 are showing up in the diff, since they already exist on `master`

jameslamb:
Ok I think I fixed the conflicts! I still have no idea how we got into this state...the danger of long-lived feature branches, I guess :grimacing: 

I am also very unsure how this PR only has one commit now. I only `git reset` back about 6 commits.

Either way, I think this PR is back to the state it was as of all previous reviews, including the most recent suggestion (https://github.com/microsoft/LightGBM/pull/2965#discussion_r424628101)

R-package/src/install.libs.R
```diff
@@ -41,71 +111,65 @@ if (!use_precompile) {
...
-        cmake_cmd <- paste0(cmake_cmd, local_vs_def)
-        build_cmd <- "cmake --build . --target _lightgbm --config Release"
+        build_cmd <- "cmake"
+        build_args <- c("--build", ".", "--target", "_lightgbm", "--config", "Release")
         lib_folder <- file.path(source_dir, "Release", fsep = "/")
+        makefiles_already_generated <- TRUE
       }
     }
+  } else {
+      .run_shell_command("cmake", c(cmake_args, ".."))
```
StrikerRUS, line 166: Should we avoid second build (`makefiles_already_generated <- TRUE`) for all non-Windows builds?

jameslamb: :scream: GOOD POINT! I completely missed that the side effect here is we're now running `cmake -G` twice for non-Windows.
```suggestion
      .run_shell_command("cmake", c(cmake_args, ".."))
      makefiles_already_generated <- TRUE
```

jameslamb:
Thanks for the reviews everyone! I'll merge this when it builds.

jameslamb:
oh no :scream: 

image

link: https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=6128&view=results

jameslamb:
Looks like Azure capacity problems to me:

image

jameslamb:
Ok well now it's ✅ but I can see the `linux sdist` task never ran. Going to push an empty commit, just to be cautious

jameslamb:
`check-docs` task is failing on one of our links 😬 

image

this link is definitely valid though: 

https://github.com/microsoft/LightGBM/blob/master/python-package/README.rst#build-from-sources

image

StrikerRUS:
Seems that GitHub did something with their DOM / js (`user-content-` prefix)...

image

jameslamb:
> Seems that GitHub did something with their DOM / js (`user-content-` prefix)...
> 
> image

:face_with_thermometer: ok yeah, I agree. Very weird that it works from a `curl` request like my example above but fails linkchecker.

Either way, I think we need to update it to `user-content`. I'll push a PR right now.

StrikerRUS:
> Very weird that it works from a curl request like my example above but fails linkchecker.

I hope that with [`10.0` release of linkchecker](https://github.com/linkchecker/linkchecker/issues/359) we will suffer a little bit less 😉 .

PR #2971, opened 2020-04-04
fixed cpplint errors

StrikerRUS:
Refer to #1990.

### DIFF
--- a/src/lightgbm_R.cpp
+++ b/src/lightgbm_R.cpp
@@ -9,14 +9,14 @@
 #include <LightGBM/utils/openmp_wrapper.h>
 #include <LightGBM/utils/text_reader.h>
 
+#include <R_ext/Rdynload.h>
+
 #include <string>
 #include <cstdio>
 #include <cstring>
 #include <memory>
 #include <utility>
 #include <vector>
-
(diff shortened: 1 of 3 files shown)


src/lightgbm_R.cpp
```diff
@@ -9,15 +9,15 @@
 #include <LightGBM/utils/openmp_wrapper.h>
 #include <LightGBM/utils/text_reader.h>
 
+#include <R_ext/Rdynload.h>
+
```
jameslamb, line 13: ah ok, thanks! I should have looked at `cpplint` manually in the logs when I did #2911

StrikerRUS:
@jameslamb Thanks!

>  I Do see failures on Azure DevOps that we'll have to investigate 

Random network error during loading SWIG.

```
Invoke-WebRequest : Oops, looks like something went wrong.
```

src/lightgbm_R.cpp
StrikerRUS: I hope soon we will not have to do this.

jameslamb: +1 I agree

PR #2975, opened 2020-04-06
fix cpp lint in json11

guolinke:
json11 issues from #1990

### DIFF
--- a/include/LightGBM/utils/json11.h
+++ b/include/LightGBM/utils/json11.h
@@ -91,7 +91,7 @@ class Json final {
 
   // Implicit constructor: anything with a to_json() function.
   template <class T, class = decltype(&T::to_json)>
-  Json(const T &t) : Json(t.to_json()) {}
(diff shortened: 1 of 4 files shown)


StrikerRUS:
Awesome PR! Only 4 issues remains! I'm so happy. 🎉 

Seems the most diff produced here is due to code redactor settings, right?

guolinke:
yes, as the indent is 4 previously, I reformat the style by copy-and-paste.

jameslamb:
> Awesome PR! Only 4 issues remain! I'm so happy. 🎉
> 
> Seems the most diff produced here is due to code redactor settings, right?

@StrikerRUS after this PR is done, I think we should put `// NOLINT` on the remaining 4 issues, capture the work to remove those in an issue, and permanently enable `cpplint` as a proper CI check to prevent any new issues being introduced.

StrikerRUS:
@jameslamb I like your suggestion!

PR #2980, opened 2020-04-07
Revert "Fix Booster read/write locale dependency (#2891)"

### LINKED ISSUE #2979: [CRITICAL BUG][Python] cannot wrire() UTF-8 strings by UnicodeEncodeError
## Environment info

Operating System: Any

CPU/GPU model: Any

Python version: Any

LightGBM version or commit hash: ce082425

## Error message

```
UnicodeEncodeError: 'ascii' codec can't encode characters in position 0-2: ordinal not in range(128)
```

## Reproducible examples

```python
import lightgbm
import numpy

X = numpy.random.random(size=(100, 3))
y = numpy.random.random(size=100)

dataset = lightgbm.Dataset(X, y)
params = {}
bst = lightgbm.train(params, dataset)

text = '日本語'  # This means Japanese :)
open('tmp.txt', 'w').write(text)
```

The below code works fine.

```python
text = '日本語'
open('tmp.txt', 'w').write(text)
```

## Steps to reproduce

1. Just run the above code

----

6b68967d9ee6ab68e353d698cbb32c992b22ee81 (#2891) breaks user locale settings.
LightGBM is just a library, and it **has never** to change user locale, which is set in users' own envrionment😡

henry0312:
This reverts commit 6b68967d9ee6ab68e353d698cbb32c992b22ee81 (#2891) and fixes #2979.


henry0312:
Please read #2979.
I hope you will understand that #2891 is a wrong way to fix the SWIG problem and it breaks a lot of user enviroment.
Also, that will lead to many issues to be reported if next version is released.
For now, because the commit is only in master,  users don't notice the locale issue.

guolinke:
ping @AlbertoEAF

henry0312:
ping @guolinke

AlbertoEAF:
As we discussed I agree with reverting my commit until we understand what is going on with Python's locale interaction: https://github.com/microsoft/LightGBM/issues/2979#issuecomment-611189639

PR #2985, opened 2020-04-09
more cpp lints

guolinke:
refer to #1990


include/LightGBM/utils/common.h
```diff
@@ -313,6 +313,7 @@ inline static unsigned CountDecimalDigit32(uint32_t n) {
     1000000000
   };
 #ifdef _MSC_VER
+  // NOLINT
```
guolinke, line 316: `_BitScanReverse` only accepts `unsigned long`.

StrikerRUS:
I guess `NOLINT` should be on the same line.
> False positives can be ignored by putting // NOLINT at the end of the line or // NOLINTNEXTLINE in the previous line.
  https://google.github.io/styleguide/cppguide.html#cpplint

include/LightGBM/utils/common.h
guolinke, line 316:
```suggestion
  // NOLINTNEXTLINE
```

src/network/socket_wrapper.hpp
```diff
@@ -218,6 +218,7 @@ class TcpSocket {
         continue;
       }
       if (ifa->ifa_addr->sa_family == AF_INET) {
+        // NOLINT
```
guolinke, line 221:
```suggestion
        // NOLINTNEXTLINE
```

StrikerRUS:
@jameslamb 
We have a new `NOTE` here on macOS
```
* checking examples ... NOTE
Examples with CPU or elapsed time > 5s
     user system elapsed
dim 5.426  0.264    5.79
```

Can you please take a look?

jameslamb:
> @jameslamb
> We have a new `NOTE` here on macOS
> 
> ```
> * checking examples ... NOTE
> Examples with CPU or elapsed time > 5s
>      user system elapsed
> dim 5.426  0.264    5.79
> ```
> 
> Can you please take a look?

😬 wow looks like we have been just barely getting under some of these checks.

I'll take a look. Opened #2988 , should be quick to fix

jameslamb:
> > @jameslamb
> > We have a new `NOTE` here on macOS
> > ```
> > * checking examples ... NOTE
> > Examples with CPU or elapsed time > 5s
> >      user system elapsed
> > dim 5.426  0.264    5.79
> > ```
> > 
> > 
> > Can you please take a look?
> 
> 😬 wow looks like we have been just barely getting under some of these checks.
> 
> I'll take a look. Opened #2988 , should be quick to fix

created #2989 to fix it

PR #2977, opened 2020-04-06
[ci][docs] downdrage sphinx and fix LaTeX error in R-package tests

StrikerRUS:
Fixed #2974.

While we are waiting new breathe release which will be compatible with sphinx 3, let's downgrade sphinx version to fix our CI.
Refer to https://github.com/michaeljones/breathe/pull/491.

PR #2990, opened 2020-04-10
[ci] fixed r-lintr version constraint

### LINKED ISSUE #2986: [ci] R linter is broken
We have the following error in `master`:
```
Linting R code
Error: 'absolute_path_linter' is not an exported object from 'namespace:lintr'
Execution halted
```

https://travis-ci.org/github/microsoft/LightGBM/jobs/672985225

StrikerRUS:
Fixed #2986.

### DIFF
--- a/.ci/test_r_package.sh
+++ b/.ci/test_r_package.sh
@@ -35,6 +35,10 @@ fi
 # Installing R precompiled for Mac OS 10.11 or higher
 if [[ $OS_NAME == "macos" ]]; then
 
+    # temp fix for basictex
+    if [[ $AZURE == "true" ]]; then
+        brew update
+    fi
     brew install qpdf
     brew cask install basictex
     export PATH="/Library/TeX/texbin:$PATH"
--- a/.ci/test.sh
+++ b/.ci/test.sh
@@ -48,7 +48,7 @@ if [[ $TRAVIS == "true" ]] && [[ $TASK == "lint" ]]; then
     conda install -q -y -n $CONDA_ENV \
         -c conda-forge \
             libxml2 \
-            r-lintr>=2.0
+            "r-lintr>=2.0"
     pip install --user cpplint
     echo "Linting Python code"
     pycodestyle --ignore=E501,W503 --exclude=./compute,./.nuget . || exit -1


StrikerRUS:
> This is WILD how did you know to try this?

It was the first thing that came to my mind when I saw your words about 

 > I see the first conda install reflected in the logs
  but not the second one with libxml2 and r-lintr

jameslamb:
> > This is WILD how did you know to try this?
> 
> It was the first thing that came to my mind when I saw your words about
> 
> > I see the first conda install reflected in the logs
> > but not the second one with libxml2 and r-lintr

amazing

StrikerRUS:
> Also I see this issue on Mac R builds now 🤒

```
```

jameslamb:
> > Also I see this issue on Mac R builds now 🤒
> 
> ```
> curl: (22) The requested URL returned error: 404 Not Found
> ```

It has to be a transient thing, right? Because the Mac builds on Travis use the exact same code and worked fine

StrikerRUS:
> It has to be a transient thing, right?

I hope so... Or it's something wrong with certificates like in #2977.
Let's wait a little bit a see 🙂

StrikerRUS:
I guess here is the root cause:
https://github.com/Homebrew/homebrew-cask/pull/80370

> May take time for all mirrors around the world to update.

StrikerRUS:
> Let's wait a little bit a see

OK, I think it's time to ask what's going on: https://github.com/Homebrew/homebrew-cask/pull/80370#issuecomment-612246863.

StrikerRUS:
Hmm, seems that 24h is not enough to update the URL. Guess we should wait till Monday.

StrikerRUS:
Still no updates for some other mirros (they are changed automatically when I re-run CI job).

.ci/test_r_package.sh
```diff
@@ -35,6 +35,10 @@ fi
 # Installing R precompiled for Mac OS 10.11 or higher
 if [[ $OS_NAME == "macos" ]]; then
 
+    # temp fix for basictex
+    if [[ $AZURE == "true" ]]; then
+        brew update
+    fi
```
jameslamb, line 41: Why is this change only for AZURE? just to save build time?

ugh the Azure Mac builds are such a pain. I had to do another "if Azure" thing on a PR I'm working on: 

image

StrikerRUS: > Why is this change only for AZURE? just to save build time?

Yeah! Seems that Travis does it automatically during `setup.sh`.

jameslamb: I saw those exact warnings actually. Thanks for the reminder!

Still just experimenting with things 😬

jameslamb:
hey glad that it worked! Thank you!!

Issue #2935, opened 2020-03-22
[performance] Single row predictions speedup.


AlbertoEAF:
Hello,

**Context**

I know you're looking for ways to improve performance throughout the codebase.

I'm working on a production scenario with LightGBM for real-time prediction systems where throughtput and latency are both important.

Due to those and other constraints, since I receive individual events which must be scored as quickly as possible, I'm using the method `LGBM_BoosterPredictForMatSingleRow` for it states in the documentation that it partly reuses internal structures to speed up computation.

**Change proposal**

I looked at the code to see if there were any easy wins we could pull off and saw that for every single prediction we create a new `Config` object from scratch. This object has roughly 200 members, and also must parse the configuration string into the different properties.

I split that `LGBM_BoosterPredictForMatSingleRow` call into a "configuration/init" call that creates the config and a "scoring" call that uses that config. With some small twraks I got in a very basic case almost 2x the throughtput. This requires adding 2 functions to the C API (without touching existing code at all):
- `LGBM_BoosterPredictForMatSingleRowFastInit` (creates the config before scoring lots of events)
- `LGBM_BoosterPredictForMatSingleRowFast` (score using the pre-build config - as we're not changing parameters)

Check here the call graph for the current (non-patched code) with a binary classifier test with 1 thread, 7 features and 100 trees (extremely simple model):
![image](https://user-images.githubusercontent.com/3976753/77259216-e9582f00-6c77-11ea-820b-810f59162e07.png)
**Roughly 1/3 of the time (27.40% + 6.40%) is spent recreating the same Config over and over.**
Notice only the left branch is doing "meaningful" work when you mantain the config properties - i.e. score lots of events with the same configuration.

In that implementation I was speaking of that cost no longer exists.

You can find a prototype code in here: https://github.com/AlbertoEAF/LightGBM/blob/ft-af-PULSEDEV-30690-optimize-single-row-predict-fast/src/c_api.cpp.

Can we add those 2 functions to the C API?
This is just the prototype and I could work a bit more on it to further improve it.

imatiach-msft:
@AlbertoEAF that looks very promising! I could definitely reuse that in mmlspark's version as well to improve performance.

StrikerRUS:
Also ping @eisber as the original author and this method is only used in SWIG wrapper which in turn is used by MMLSpark.

guolinke:
@AlbertoEAF good catch! Contribution is very welcome.

StrikerRUS:
[...]

Issue #261, opened 2017-01-25
Need much more tests


henry0312:
It's necessary to write much more tests to keep the quality of LightGBM.
I can contribute to python-package.

TODO:

- [ ] Tests for C++
- [x] Tests for Python
- [ ] Tests for R

chivee:
That will be great! Please help us on the test, and leveraging travis for a better release strategies. we are planing a better versioning for the release, so that the package users can stick to the stable version

wxchan:
Adding some tests on multi-class and lambdarank tasks will be great.

henry0312:
I start with python-package from today :)

Laurae2:
@henry0312 tests are completed for Python I think. For R we need to setup Travis for it.

henry0312:
@Laurae2 no, the current tests doesn't cover all of the python-package.

@wxchan I'd like to use [nose](http://nose.readthedocs.io/en/latest/index.html). What do you think about nose?

henry0312:
I'll go with [nose](http://nose.readthedocs.io/en/latest/index.html)

StrikerRUS:
@b1shal711 It will be great!

StrikerRUS:
Closed in favor of being in #2302. We decided to keep all feature requests in one place.

Welcome to contribute this feature! Please re-open this issue (or post a comment if you are not a topic starter) if you are actively working on implementing this feature.

StrikerRUS:
@jameslamb Thanks to all your hard work, I believe we can tick `Tests for R` now in the original comment. WDYT?

jameslamb:
I think so, yes! I think the R tests are in a much better place than they were when this issue was opened

PR #2992, opened 2020-04-11
Feat/optimize single prediction

AlbertoEAF:
This follows in sequence of these findings: https://github.com/microsoft/LightGBM/issues/2935.

A new struct is added to the C API, along with some methods to leverage it for faster single-row predictions.
There are no impacts to the current API as all changes are new independent additions:
- The method of interest: `LGBM_BoosterPredictForMatSingleRowFast()`
- Auxiliary methods:
  - `LGBM_BoosterPredictForMatSingleRowFastInit()` - instantiates the `FastConfig`
  - `LGBM_FastConfigFree()` - releases the `FastConfig`

This method splits up set-up from scoring. As it reduces memory allocations it reduces not only the throughtput but latency as well.

It however requires 2 more calls from who is using it, one for setup of `FastConfig`, and another for its release.

To score one simply passes the `FastConfig` and not the array data again, as this also "caches" the `get_row_fun`. Basically to score new instances, one just replaces the data in the single-row features array and makes a new FastPredict call.

I also have a commit for profiling this, which I didn't merge as it would add more files to the repo, but shows how one can score:
https://github.com/AlbertoEAF/LightGBM/commit/d965ba1e74568fac60fc8ea1c03feded836d8ab1

Don't know however, if we shouldn't start adding C tests as well. The ones at that branch can be explored for that purpose.

Initially there was just a static `FastConfig` instantiated inside `c_api.cpp` which the user of the C APi didn't even know about,  but decided to change it to be public, so that many can be managed. I think this is ready to start the review process :)


jameslamb:
Hi @AlbertoEAF sorry for the inconvenience, but can you please merge to most recent `master`? We just merged some fixes for the continuous integration failures we've had recently. It should make your build green.

AlbertoEAF:
Hello @jameslamb, no problem, but it seems there are still some issues, namely linting in codes I didn't touch. I'm assuming it's best not to touch those as someone else who added that is likely to fix them in the master. I'm not sure when that will be fixed to rebase now though.

StrikerRUS:
@AlbertoEAF Hi!

Those linting errors are introduced in this PR.

```
src/c_api.cpp:1676:  "private:" should be preceded by a blank line  [whitespace/blank_line] [3]
src/c_api.cpp:1699:  At least two spaces is best between code and comments  [whitespace/comments] [2]
src/c_api.cpp:1701:  Closing ) should be moved to the previous line  [whitespace/parens] [2]
Done processing src/c_api.cpp
```

Please fix them.

Also there are some problems (forgotten argument?) with  Doxygen documentation for new functions that should be resolved before starting a review process for this PR.

```
/home/travis/build/microsoft/LightGBM/include/LightGBM/c_api.h:915: error: argument 'fast_config_handle' of command @param is not found in the argument list of LGBM_BoosterPredictForMatSingleRowFast(FastConfigHandle fastConfig_handle, int predict_type, int num_iteration, int64_t *out_len, double *out_result) (warning treated as error, aborting now)
```

AlbertoEAF:
Thanks for the pointer @StrikerRUS! , I misread the linter as if the changes were on another file, my bad.
---

@imatiach-msft: I want to know if reusing the input features array is a limitation to your use case.

One could choose to pass at every prediction the pointer to the new input features address, but by reusing the same data address (which implies you must overwrite it prior to a new prediction), brings a tiny extra speed up still, and helps latencies by avoiding more allocations (lambda capture followed by get_row_fun's std::function creation).

AlbertoEAF:
@imatiach-msft ping :D

imatiach-msft:
@AlbertoEAF yes, this is very nice.  I don't think re-using the same native data array is a limitation, but I think there is some misunderstanding, as mmlspark uses the methods LGBM_BoosterPredictForCSRSingle for sparse prediction which @eisber added:
https://github.com/Azure/mmlspark/blob/master/src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMBooster.scala#L247
and LGBM_BoosterPredictForMatSingle for dense prediction:
https://github.com/Azure/mmlspark/blob/master/src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMBooster.scala#L266
I believe these take the native representation from java array directly, so there shouldn't be any memory churn, unless I am misunderstanding something, eg see:
https://github.com/microsoft/LightGBM/blob/master/swig/lightgbmlib.i#L119
```
    int* indices0 = (int*)jenv->GetPrimitiveArrayCritical(indices, &isCopy);
    double* values0 = (double*)jenv->GetPrimitiveArrayCritical(values, &isCopy);
```

imatiach-msft:
also it looks like the current code only addresses the dense case, but we also use CSR predict function for sparse case

imatiach-msft:
"helps latencies by avoiding more allocations (lambda capture followed by get_row_fun's std::function creation)."

sorry could you explain this a bit more - What is the lambda capture that you are referring to specifically?  Would this be an issue for mmlspark or is this only for native-only callers who are not using mmlspark?  For mmlspark, there shouldn't be more memory allocated for the input data, just memory passed from java data structures directly that were previously already allocated.

The get_row_fun's std::function creation seems like it could be separate from re-using the same memory for the input data, eg that part could be optimized even if a new array was passed in every time?  Maybe I am missing something there.

AlbertoEAF:
Hmm interesting @imatiach-msft,

In the Java implementation I'm working on, a C array is initialized at program startup (pointer `swigResources.swigInstancePtr`), and for every new runtime instance, I get new features from an external source in a custom format (Instance), and then copy those values into the C array. I was using:

```java
private void initSWIGInstanceArray(final Instance instance) {

    int skipTargetOffset = 0; // set to 1 after passing the target (if it exists)
    for (int i = 0; i < this.schemaNumFields; ++i) {
        // If the label is not present, targetIndex=-1, thus "if" wont't trigger:
        if (i == this.schemaTargetIndex) {
            skipTargetOffset = -1;
        } else {
            lightgbmlibJNI.doubleArray_setitem(
                    this.swigResources.swigInstancePtr,
                    i + skipTargetOffset,
                    instance.getValue(i)
            );
        }
    }
}
```
and then use that pointer for the prediction call. Thus, my address is always the same.

_To use that function I'd need first to copy to a Java double array and then use that array data (which is then passed to C by `GetPrimitiveArrayCritical`. Maybe the overhead of copying the data to a Java double array and then using that function relying on `GetPrimitiveArrayCritical` is smaller when there are many features (given the amount of JNI calls I have to do is proportional to #features)? If it requires a copy I'm not sure however regarding the behaviour (2 copies total)._


With the `getCritical` version there are no guarantees that we won't get a copy (thus a new address), so I guess the address of the feature array data cannot be set only once and reused, right @imatiach-msft ?

When I spoke of the lambda capture & std::function I meant both steps on the C++ side. Clearly the biggest overhead there comes from creating the new std::function to accomodate the new features address.


> The get_row_fun's std::function creation seems like it could be separate from re-using the same memory for the input data, eg that part could be optimized even if a new array was passed in every time?

How do you suggest doing that @imatiach-msft ?
I'm not seeing how to replace this:
https://github.com/microsoft/LightGBM/blob/7076cb8a3ac3a7b32dcf37be5593dddf27bf7f16/src/c_api.cpp#L1640

Right now a std::function is generated and `data` is not one of its parameters, we'd have to change all the code that relies on it so we could pass an extra pointer `data` at the instant of getting the values? And thus also have to keep track of that pointer `data` when we need to use that function?

AlbertoEAF:
Gentle ping to @imatiach-msft :)

+ @guolinke @StrikerRUS as maybe you guys also have input regarding eliminating the generation of a new std::function `RowPairFunctionFromDenseMatric` on every prediction with new `data` addresses :)

AlbertoEAF:
@imatiach-msft I decided to remove the optimization of reusing the function.

We lose 5% of the 30% speed improvement but now it is usable for my case (generic) and MMLSpark's (with the GetCritical calls).

How do you want to deal with the Java wrappers?

Have the current version in lightgbmlib.i as it requires no separate steps for setup and teardown AND the new one? Or replace that by the scoring call of the new one? (Setup and Teardown can use directly the C API, no need for wrapping there).

Regarding the CSC optimizations I think we can leave that for later? You already use that only for specific cases, the general one is the DenseRow right? 
(Adding it should bring no more complexity, I believe the same FastConfig can be used, it's just rinse and repeat).

imatiach-msft:
@AlbertoEAF you have a lot of questions :)
"We lose 5% of the 30% speed improvement but now it is usable for my case (generic) and MMLSpark's (with the GetCritical calls)."
is there a way we could enable both cases?  5% improvement would still be nice to have.

"To use that function I'd need first to copy to a Java double array and then use that array data"
No, I think the way you are doing it right now is much better, don't copy to a separate Java array just for this.

Do you actually have to predict one row at a time from your program?  You could probably improve performance much more by batching, if that was possible in your scenario, putting multiple instances in a single native array and then calling predict once?

"GetPrimitiveArrayCritical version there are no guarantees that we won't get a copy"
That seems like another good reason for why you should do it the way you are doing now and not create a separate Java array just for this, it doesn't really make sense I think if you are trying to really optimize performance?  That method for Spark is really optimized just for Spark.

"Right now a std::function is generated and data is not one of its parameters, we'd have to change all the code that relies on it so we could pass an extra pointer data at the instant of getting the values? "
At first I thought yes, then with your hesitation I thought no, but then looking at the code more it seems like this might work, you can pass in data either to the function, or you could pass in a double pointer to data and modify which row it points to between calls, which may make the code more confusing but you wouldn't need to refactor as much:
```
Option 1:
std::function<std::vector<double>(int row_idx, const void* data)>
RowFunctionFromDenseMatric(int num_row, int num_col, int data_type, int is_row_major) {
Option 2:
std::function<std::vector<double>(int row_idx)>
RowFunctionFromDenseMatric(void** data_ptr, int num_row, int num_col, int data_type, int is_row_major) {
```
Agree that it might not be worth it if it involves a lot of refactoring.  If that actually is hurting performance, it may make sense to refactor it.  I would check with @guolinke first though.

Actually, looking through the code more, I see some optimizations that could be made.  We are taking in an array, copying it to a vector from index->value, and then in predict function we are again copying it to a buffer by accessing indexes in dense case - but it seems like we could just bypass this and copy the original data array:
https://github.com/microsoft/LightGBM/blob/0aa7bfee65eb62a99bf22137ad02d7cbf12390c2/src/application/predictor.hpp#L107

See LGBM_BoosterPredictForMatSingleRow:
https://github.com/microsoft/LightGBM/blob/7076cb8a3ac3a7b32dcf37be5593dddf27bf7f16/src/c_api.cpp#L1640
See the code that calls PredictSingleRow and predict_function:
https://github.com/microsoft/LightGBM/blob/7076cb8a3ac3a7b32dcf37be5593dddf27bf7f16/src/c_api.cpp#L380
See code which creates the Predict function:
https://github.com/microsoft/LightGBM/blob/master/src/c_api.cpp#L418
See predict_function which is call to GetPredictFunction:
https://github.com/microsoft/LightGBM/blob/7076cb8a3ac3a7b32dcf37be5593dddf27bf7f16/src/c_api.cpp#L77

This optimization might take too much time to implement though and I'm not sure how much it's worth it to make those complex changes in the code.  Conceptually though, there is room for some optimization because the intermediate vector representation isn't really useful in the dense case and it just increases memory churn and time to copy over data, even though in the end it's not even the data structure that is used for prediction.

"Regarding the CSC optimizations I think we can leave that for later? You already use that only for specific cases, the general one is the DenseRow right?"
Sparse scenario is super important, especially with the work I am doing right now with (Microsoft's AzureML) AutoML and their model interpretability feature, where if their data has strings it is almost always featurized into a sparse representation, which is why I'm trying to add a sparse TreeShap implementation here: https://github.com/microsoft/LightGBM/pull/3000.  That's not related to MMLSpark's LightGBM though, where sparse structures aren't used as often, although sparse scenario is very important in many use-cases - I don't know the exact ratio, I would guess maybe 70 dense/30 sparse percent split?

AlbertoEAF:
>  especially with the work I am doing right now with (Microsoft's AzureML) AutoML and their model interpretability feature,

Cool stuff!

I hope to have time this week to get back to you on all that you said :p

AlbertoEAF:
Regarding your last comment @imatiach-msft, I agree that optimizing that would be nice, but I'd rather avoid it as passing the original data pointer around further complicates the calls. @guolinke what are your thoughts on that?

> Do you actually have to predict one row at a time from your program?
Yes I do :) It's meant for real-time use, as little latency as possible, so I cannot wait for more events.

> Sparse scenario is super important
I'll try to add the corresponding sparse methods as well to SWIG then.


The sparse predictors like `LGBM_BoosterPredictForCSR` take num_col as int64_t, but `LGBM_BoosterPredictForMatSingleRow` takes int32_t.

As lots of code depends on that I didn't change int64_t to int32_t, so I added an IntUnion to FastConfig, so any type can be used, which meant not changing any existing functions in the C API, although in LGBM_BoosterPredictForCSRSingleRow, there was already a cast of num_col from int64_t to int32_t, further reinforcing the point above.

------


I think the code is done, can you review? @imatiach-msft , @guolinke, @StrikerRUS 


I tested by introducing 2 profiling C++ "tests" with which you can compare the time of the `LGBM_BoosterPredictForMatSingleRowFast` and `LGBM_BoosterPredictForMatSingleRow` function calls and scores. 

Each .cpp scores the same instance a bunch of times and prints the last call's score. You can see how they are the same, and the Fast variant is with the Higgs model (28 features) 19-20% faster even when the config has no parameters in it.

I added a profiling folder (probably should move it to tests?) - It builds a binary that scores the same event lots of times using the fast and non fast variants. It prints the last score (all will be the same). I get the same score with the Fast variant and with the regular SingleRow method:

image

I didn't upload the model as it's still 14MB (though uploading through git lfs would be ok - it wouldn't add to the git repo weight)

You can repeat the benchmarks yourselves by following the instructions on the last commit message.

Tell me what you think :)

imatiach-msft:
@AlbertoEAF  "Regarding your last comment @imatiach-msft, I agree that optimizing that would be nice, but I'd rather avoid it as passing the original data pointer around further complicates the calls"

sure, definitely not saying you need to do that right now, was just brainstorming ways to optimize the code more

imatiach-msft:
"I also added the helpers in SWIG for the new dense and sparse variants which rely on the Critical JNI call as well for maximum speed @imatiach-msft ;)"

that's awesome!  thanks a lot!

AlbertoEAF:
@StrikerRUS any clue on what is making this fail the doxygen build? Did I make something wrong?

StrikerRUS:
@AlbertoEAF Very strange error! Maybe wrong `return` here?
```
\return LIGHTGBM_C_EXPORT LGBM_FastConfigFree
```
I guess as always it should be 
```
\return 0 when it succeeds, -1 when failure happens
```

AlbertoEAF:
Any clue why Sphinx is failing?

StrikerRUS:
> Any clue why Sphinx is failing?

Not sure that this is the reason (maybe it is another bug) but here is the inconsistency:
```
\param ncol Number of columns
```
```
const int64_t num_col,
```
in `LGBM_BoosterPredictForCSRSingleRowFastInit`

AlbertoEAF:
> > Any clue why Sphinx is failing?
> 
> Not sure that this is the reason (maybe it is another bug) but here is the inconsistency:
> 
> ```
> \param ncol Number of columns
> ```
> 
> ```
> const int64_t num_col,
> ```
> 
> in `LGBM_BoosterPredictForCSRSingleRowFastInit`

Can I ask how you find that out @StrikerRUS ? I didn't see that in the CI logs. You did? Or you have to check manually the source code? Asking to avoid nagging you next time :p Thanks :)

guolinke:
What is the state of this PR ?

StrikerRUS:
@AlbertoEAF 

> Can I ask how you find that out @StrikerRUS ?

Sorry, I overlooked your question! Yes, I 

> Or you have to check manually the source code?

started to read the diff word by word very attentively.

AlbertoEAF:
> What is the state of this PR ?

Working, profiled and tested, but it's stuck on doxygen's documentation parsing @guolinke. 

@StrikerRUS has been helping a lot tracking what's wrong, but I'm still stuck. I will try following his suggestion this weekend but right now don't know where's the documentation issue still.

Is there another way to ease this doxygen error-tracking process? I'm open to more tips. For instance, running it manually will give me a more descriptive error message?

StrikerRUS:
@AlbertoEAF 
>  I will try following his suggestion this weekend but right now don't know where's the documentation issue still.

By trial and error I've learned that the problem is in `LGBM_BoosterPredictForMatSingleRowFast` and `LGBM_BoosterPredictForCSRSingleRowFast` functions. The current state of my investigations: removing them from file allows to run `make html` docs successfully.

StrikerRUS:
@AlbertoEAF Got it! The problem was in `#threads` 😄 , more precisely in the `#` sign. I guess it is a special sign in doxygen's language.

```diff
diff --git a/include/LightGBM/c_api.h b/include/LightGBM/c_api.h
index 23937c93..d7948a94 100644
--- a/include/LightGBM/c_api.h
+++ b/include/LightGBM/c_api.h
@@ -813,9 +813,9 @@ LIGHTGBM_C_EXPORT int LGBM_BoosterPredictForCSRSingleRowFastInit(BoosterHandle h
  * initializing the config only once, instead of once per call.
  *
  * \note
- *   Setting up #threads is only done once at `LGBM_BoosterPredictForCSRSingleRowFastInit`
+ *   Setting up threads is only done once at `LGBM_BoosterPredictForCSRSingleRowFastInit`
  *   instead of at each prediction.
- *   If you use a different #threads in other calls, you need to start the setup process over,
+ *   If you use a different threads in other calls, you need to start the setup process over,
  *   or that number of threads will be used for this calls as well.
  *
  * \note
@@ -993,9 +993,9 @@ LIGHTGBM_C_EXPORT int LGBM_BoosterPredictForMatSingleRowFastInit(BoosterHandle h
  * initializing the config only once, instead of once per call.
  *
  * \note
- *   Setting up #threads is only done once at `LGBM_BoosterPredictForMatSingleRowFastInit`
+ *   Setting up threads is only done once at `LGBM_BoosterPredictForMatSingleRowFastInit`
  *   instead of at each prediction.
- *   If you use a different #threads in other calls, you need to start the setup process over,
+ *   If you use a different threads in other calls, you need to start the setup process over,
  *   or that number of threads will be used for this calls as well.
  *
  * \param fastConfig_handle FastConfig object handle returned by `LGBM_BoosterPredictForMatSingleRowFastInit`
```

CMakeLists.txt
```diff
@@ -234,6 +235,12 @@ file(GLOB SOURCES
     src/treelearner/*.cpp
 )
 
+if(BUILD_PROFILING_TESTS)
+    # For profiling builds with valgrind/callgrind use -DDEBUG=ON
+    add_executable(lightgbm_profile_single_row_predict profiling/profile_single_row_predict.cpp ${SOURCES} src/c_api.cpp)
+    add_executable(lightgbm_profile_single_row_predict_fast profiling/profile_single_row_predict_fast.cpp ${SOURCES} src/c_api.cpp)
+endif(BUILD_PROFILING_TESTS)
```
StrikerRUS, line 242: I'm not sure that we should add profiling part into the repo... @guolinke WDYT?

AlbertoEAF: If that is the decision should I also remove the new profiling folder and its contents?

StrikerRUS: > If that is the decision

Let's wait for some other opinions.

> should I also remove the new profiling folder and its contents?

I think yes. Right now `profiling` folder looks some kind of unnatural here. I believe we should transform it into a test during the work on #261 or integrate it into other benchmarks (https://github.com/guolinke/boosting_tree_benchmarks).

StrikerRUS:
Test failures look like a merging conflict due to #3000. @AlbertoEAF Can you please take a look?

AlbertoEAF:
> Test failures look like a merging conflict due to #3000. @AlbertoEAF Can you please take a look?

Hi @StrikerRUS sure but these builds were all passing, and I only removed 4 "#" in the docstrings to make doxygen build. How could this change have broken all that?

AlbertoEAF:
> Test failures look like a merging conflict due to #3000. @AlbertoEAF Can you please take a look?

Sure @StrikerRUS but I don't understand how it can be from a merge conflict,  in my computer the `cmake .. -DUSE_SWIG=ON && make -j4` in this branch works with no errors.

I don't understand what's going wrong, looking at the error logs of https://travis-ci.org/github/microsoft/LightGBM/jobs/704885372 for instance, I get this line numbers, but my code doesn't have such lines at all:

```
/home/travis/build/microsoft/LightGBM/src/c_api.cpp:1978:22: error: no matching
      function for call to 'RowFunctionFromCSR'
  auto get_row_fun = RowFunctionFromCSR(indptr, indptr_type, indices, da...
                     ^~~~~~~~~~~~~~~~~~
/home/travis/build/microsoft/LightGBM/src/c_api.cpp:827:1: note: candidate
      template ignored: couldn't infer template argument 'T'
RowFunctionFromCSR(const void* indptr, int indptr_type, const int32_t* indices,```

For instance, this is my line 1978:
```  }
  Log::Fatal("Unknown data type in RowFunctionFromDenseMatric");
  return nullptr;
} // line 1978
```

and line 827:
```ret->PushOneRow(tid, start_row + i, one_row);```

Can you explain why this is happening? I don't understand how it can come from a merge conflict.

**Comment by StrikerRUS** [comment]
@AlbertoEAF 
> Sure @StrikerRUS but I don't understand how it can be from a merge conflict,

Ah, I though that you had rebased your branch. But now I can see that this is not true:

image


Your branch is quite old and  have no changes from already merged #3000. That is why everything is OK locally on your machine. However, all CI tests are run against our most recent `master` where codebase is already updated. You should pull changes from our `master` to fix conflicts (actually, these are not merge conflicts, but function signature missmatch due to recent changes). Here is the guide how to do it:
https://stackoverflow.com/a/7244456 (replace the link with `https://github.com/microsoft/LightGBM.git`).

**Comment by AlbertoEAF** [comment]
Ah got it, I didn't know the tests were done against the master. I'll rebase + fix conflicts in the weekend StrikerRUS.

**Comment by AlbertoEAF** [comment]
Rebased and ready for final review @StrikerRUS @guolinke.

Thanks for all the help @StrikerRUS ;)

File: CMakeLists.txt (hunk 1 above)
Comment (StrikerRUS) [reply CMakeLists.txt:243]: gently ping @guolinke for your opinion

Comment (guolinke) [reply CMakeLists.txt:243]: IMO, it is better to remove this in the official repo.

Comment (guolinke) [reply CMakeLists.txt:243]: But I can keep a branch for the profiling

File: src/c_api.cpp [hunk 2]
```diff
@@ -1886,6 +1936,51 @@ int LGBM_BoosterPredictForCSRSingleRow(BoosterHandle handle,
...
+                                               const int data_type,
+                                               const int64_t num_col,
+                                               const char* parameter,
+                                               FastConfigHandle *out_fastConfig) {
+  API_BEGIN();
+  if (num_col <= 0) {
+    Log::Fatal("The number of columns should be greater than zero.");
+  } else if (num_col >= INT32_MAX) {
+    Log::Fatal("The number of columns should be smaller than INT32_MAX.");
+  }
```
guolinke, line 1949: I am a little confused about `ncol.int64`. It seems both CSR and Mat use the 32-bit ncol, why int64 is needed?

CMakeLists.txt
StrikerRUS: @guolinke I believe that a separate repo (just like for benchmarks) will be more intuitive.

@AlbertoEAF OK, seems that we got a decision. Please exclude profiling from this PR, but also save the corresponding code for some time somewhere in your fork.

src/c_api.cpp
AlbertoEAF: Great catch @guolinke , in my questions before:

> The sparse predictors like `LGBM_BoosterPredictForCSR` take num_col as int64_t, but `LGBM_BoosterPredictForMatSingleRow` takes int32_t.

I hadn't noticed that pre-condition in the function receiving int64_t and thought both types had to be supported.

Just removed that and the profiling/ folder.

CMakeLists.txt
AlbertoEAF: Thanks for the input :)

I just removed it and kept it in a branch in my fork.

AlbertoEAF:
Anything left for the merge @guolinke ?

include/LightGBM/c_api.h
```diff
@@ -844,6 +853,73 @@ LIGHTGBM_C_EXPORT int LGBM_BoosterPredictForCSRSingleRow(BoosterHandle handle,
...
+/*!
+ * \brief Initialize and return a `FastConfigHandle` for use with `LGBM_BoosterPredictForCSRSingleRowFast`.
+ *
+ * Release the `FastConfig` by passing its handle to `LGBM_FastConfigFree` when no longer needed.
+ *
+ * \param handle Booster handle
+ * \param data_type Type of ``data`` pointer, can be ``C_API_DTYPE_FLOAT32`` or ``C_API_DTYPE_FLOAT64``
+ * \param num_col Number of columns
+ * \param parameter Other parameters for prediction, e.g. early stopping for prediction
+ * \param[out] out_fastConfig FastConfig object with which you can call `LGBM_BoosterPredictForMatSingleRowFast`
```
StrikerRUS, line 865:
```suggestion
 * \param[out] out_fastConfig FastConfig object with which you can call `LGBM_BoosterPredictForCSRSingleRowFast`
```

include/LightGBM/c_api.h
```diff
@@ -844,6 +853,73 @@ LIGHTGBM_C_EXPORT int LGBM_BoosterPredictForCSRSingleRow(BoosterHandle handle,
...
+ * Score single rows after setup with `LGBM_BoosterPredictForCSRSingleRowFastInit`.
+ *
+ * By removing the setup steps from this call extra optimizations can be made like
+ * initializing the config only once, instead of once per call.
+ *
+ * \note
+ *   Setting up the number of threads is only done once at `LGBM_BoosterPredictForCSRSingleRowFastInit`
+ *   instead of at each prediction.
+ *   If you use a different number of threads in other calls, you need to start the setup process over,
+ *   or that number of threads will be used for this calls as well.
```
StrikerRUS, line 886:
```suggestion
 *   or that number of threads will be used for these calls as well.
```

include/LightGBM/c_api.h
```diff
@@ -957,6 +1033,57 @@ LIGHTGBM_C_EXPORT int LGBM_BoosterPredictForMatSingleRow(BoosterHandle handle,
...
+ * Score a single row after setup with `LGBM_BoosterPredictForMatSingleRowFastInit`.
+ *
+ * By removing the setup steps from this call extra optimizations can be made like
+ * initializing the config only once, instead of once per call.
+ *
+ * \note
+ *   Setting up the number of threads is only done once at `LGBM_BoosterPredictForMatSingleRowFastInit`
+ *   instead of at each prediction.
+ *   If you use a different number of threads in other calls, you need to start the setup process over,
+ *   or that number of threads will be used for this calls as well.
```
StrikerRUS, line 1066:
```suggestion
 *   or that number of threads will be used for these calls as well.
```

CMakeLists.txt
StrikerRUS: Ah, I've just come the thought that you can simply post the content of profiling files in comment for this PR or original issue. 🙂

AlbertoEAF:
Changed @StrikerRUS :)

Anything left for merging?

PR #2989, opened 2020-04-10
[R-package] [docs] Simplified examples to cut example run time (fixes #2988)

jameslamb:
See #2988 . The R package is very close to hitting this `R CMD CHECK` NOTE:

```
* checking examples ... NOTE
Examples with CPU or elapsed time > 5s
     user system elapsed
dim 5.426  0.264    5.79
```

In this PR, I cut the runtime for our examples by:

* removing all `library(lightgbm)` (these are not necessary and it is conventional to omit them)
* cutting `nrounds` from  `10L` to smaller numbers in all examples with training
* wrapping examples for code that relies on the filesystem in `\donttest`
* wrapping plotting code in `\donttest`

These tests run in about 2.9 seconds on my laptop. To get the timings for yourself

```shell
pushd R-package
    Rscript -e "roxygen2::roxygenize(load = 'installed')"
popd

Rscript build_r.R --skip-install

R CMD CHECK lightgbm*.tar.gz \
    --no-tests \
    --no-manual \
    --timings

Rscript -e "
    timing_df <- read.delim('lightgbm.Rcheck/lightgbm-Ex.timings');
    print(timing_df);
    print(paste0('total time (s) ', sum(timing_df[['name']])))
    "
```

Results:

```text
                             name  user system elapsed
dim                         0.705 0.038  0.746      NA
dimnames.lgb.Dataset        0.049 0.006  0.046      NA
getinfo                     0.061 0.004  0.057      NA
lgb.Dataset                 0.050 0.008  0.050      NA
lgb.Dataset.construct       0.044 0.002  0.037      NA
lgb.Dataset.create.valid    0.035 0.003  0.037      NA
lgb.Dataset.save            0.042 0.003  0.036      NA
lgb.Dataset.set.categorical 0.043 0.002  0.036      NA
lgb.Dataset.set.reference   0.034 0.003  0.038      NA
lgb.cv                      0.287 0.078  0.203      NA
lgb.dump                    0.001 0.000  0.000      NA
lgb.get.eval.result         0.118 0.017  0.063      NA
lgb.importance              0.167 0.018  0.137      NA
lgb.interprete              0.409 0.024  0.216      NA
lgb.load                    0.000 0.000  0.000      NA
lgb.model.dt.tree           0.251 0.040  0.145      NA
lgb.plot.importance         0.146 0.021  0.108      NA
lgb.plot.interpretation     0.000 0.000  0.000      NA
lgb.prepare                 0.017 0.001  0.009      NA
[...]

PR #2999, opened 2020-04-16
[R-package] fix R examples and lgb.plot.interpretation()

jameslamb:
See comments from @StrikerRUS 

* https://github.com/microsoft/LightGBM/pull/2989#issuecomment-614374151
* https://github.com/microsoft/LightGBM/pull/2989#issuecomment-614375963

A few of the R examples are broken right now. These weren't caught by tests because one is only a warning, one is a plot that is generated successfully but is incorrect, and one is an issue specific to our readthedocs builds.

I think the issue for `lgb.cv()` is #2715. In that issue we found that `data.table 1.11.4` and R 3.6.0 lead to a data.table error ending in `column or argument 2 is NULL`. I think this can be fixed by upgrading to the `data.table`  1.12.1 in `conf.py`.

To fix the `lgb.set.categorical()` example, I changed file names across several examples to be sure that no two examples write to the same file name.

image

This issue with `lgb.plot.interpretation()` is because I made a change and forgot to use `abs()`.

image

new plot:

image

I was not able to reproduce the `data.table` error in  https://github.com/microsoft/LightGBM/pull/2989#issuecomment-614374151. It

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -406,6 +406,15 @@ lightgbm_r/*
 lightgbm*.tar.gz
 lightgbm.Rcheck/
 
+# Files created by R examples and tests
+**/lgb-Dataset.data
+**/lgb-model.rds
+**/lgb.Dataset.data
+**/model.rds
+**/model.txt
+**/lgb-model.txt
+
+
 # Files from interactive R sessions
 .Rproj.user
 **/.Rhistory
--- a/.gitignore
+++ b/.gitignore
@@ -414,7 +414,6 @@ lightgbm.Rcheck/
 **/model.txt
 **/lgb-model.txt
 
-
 # Files from interactive R sessions
 .Rproj.user
 **/.Rhistory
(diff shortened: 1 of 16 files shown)


StrikerRUS:
Thanks @jameslamb for taking a look!

> I think the issue for lgb.cv() is #2715.

But we do not use `lgb.dl()` at RTD and R version is 3.5 there. Probably we should update our FAQ entry...

If you work from LightGBM's branch, not from your fork, then I'll be able to enable RTD builds for that branch and make sure that all issues are gone. Will it make sense?

jameslamb:
> Thanks @jameslamb for taking a look!
> 
> > I think the issue for lgb.cv() is #2715.
> 
> But we do not use `lgb.dl()` at RTD and R version is 3.5 there. Probably we should update our FAQ entry...
> 
> If you work from LightGBM's branch, not from your fork, then I'll be able to enable RTD builds for that branch and make sure that all issues are gone. Will it make sense?

Sure! I can recreate the PR from a LightGBM branch.

jameslamb:
closing in favor of #3002

PR #3000, opened 2020-04-16
adding sparse support to TreeSHAP in lightgbm

imatiach-msft:
Adding sparse support to TreeSHAP algorithm in lightgbm.  The feature importances for a sparse matrix should be returned as a sparse matrix as well.  This should improve both performance and memory usage for very sparse datasets.

Unlike other predict APIs, it's not as easy to figure out prior to prediction what the size will be of the sparse matrix result, so we allocate the data on the native-side and expose an additional API to deallocate the sparse matrix arrays.

### DIFF
--- a/include/LightGBM/c_api.h
+++ b/include/LightGBM/c_api.h
@@ -33,6 +33,9 @@ typedef void* BoosterHandle;  /*!< \brief Handle of booster. */
 #define C_API_PREDICT_LEAF_INDEX (2)  /*!< \brief Predict leaf index. */
 #define C_API_PREDICT_CONTRIB    (3)  /*!< \brief Predict feature contributions (SHAP values). */
 
+#define C_API_MATRIX_TYPE_CSR (0)  /*!< \brief CSR sparse matrix type. */
+#define C_API_MATRIX_TYPE_CSC (1)  /*!< \brief CSC sparse matrix type. */
+
 /*!
  * \brief Get string message of the last error.
  * \return Error information
@@ -743,31 +746,32 @@ LIGHTGBM_C_EXPORT int LGBM_BoosterPredictForCSR(BoosterHandle handle,
                                                 double* out_result);
 
 /*!
- * \brief Make sparse prediction for a new dataset in CSR format.  Currently only used for feature contributions.
+ * \brief Make sparse prediction for a new dataset in CSR or CSC format.  Currently only used for feature contributions.
  * \note
  * The outputs are pre-allocated, as they can vary for each invocation, but the shape should be the same:
  *   - for feature contributions, the shape of sparse matrix will be ``num_class * num_data * (num_feature + 1)``.
  * The output indptr_type for the sparse matrix will be the same as the given input indptr_type.
  * \param handle Handle of booster
- * \param indptr Pointer to row headers
+ * \param indptr Pointer to row headers for CSR or col headers for CSC
  * \param indptr_type Type of ``indptr``, can be ``C_API_DTYPE_INT32`` or ``C_API_DTYPE_INT64``
- * \param indices Pointer to column indices
+ * \param indices Pointer to column indices for CSR or row indices for CSC
  * \param data Pointer to the data space
  * \param data_type Type of ``data`` pointer, can be ``C_API_DTYPE_FLOAT32`` or ``C_API_DTYPE_FLOAT64``
  * \param nindptr Number of rows in the matrix + 1
  * \param nelem Number of nonzero elements in the matrix
- * \param num_col Number of columns
+ * \param num_col_or_row Number of columns for CSR or number of rows for CSC
  * \param predict_type What should be predicted, only feature contributions supported currently
  *   - ``C_API_PREDICT_CONTRIB``: feature contributions (SHAP values)
  * \param num_iteration Number of iterations for prediction, <= 0 means no limit
  * \param parameter Other parameters for prediction, e.g. early stopping for prediction
+ * \param matrix_type The type of matrix that is input and output, 0 for CSR and 1 for CSC
  * \param[out] out_len Length of output indices and data
- * \param[out] out_indptr Pointer to output row headers
- * \param[out] out_indices Pointer to sparse indices
+ * \param[out] out_indptr Pointer to output row headers for CSR or col headers for CSC
+ * \param[out] out_indices Pointer to sparse column indices for CSR or row indices for CSC
  * \param[out] out_data Pointer to sparse data space
  * \return 0 when succeed, -1 when failure happens
  */
-LIGHTGBM_C_EXPORT int LGBM_BoosterPredictSparseForCSR(BoosterHandle handle,
+LIGHTGBM_C_EXPORT int LGBM_BoosterPredictSparseOutput(BoosterHandle handle,
                                                       const void* indptr,
                                                       int indptr_type,
                                                       const int32_t* indices,
@@ -775,10 +779,11 @@ LIGHTGBM_C_EXPORT int LGBM_BoosterPredictSparseForCSR(BoosterHandle handle,
                                                       int data_type,
                                                       int64_t nindptr,
                                                       int64_t nelem,
-                                                      int64_t num_col,
+                                                      int64_t num_col_or_row,
                                                       int predict_type,
                                                       int num_iteration,
                                                       const char* parameter,
+                                                      int matrix_type,
                                                       int64_t* out_len,
                                                       void** out_indptr,
                                                       int32_t** out_indices,
@@ -879,48 +884,6 @@ LIGHTGBM_C_EXPORT int LGBM_BoosterPredictForCSC(BoosterHandle handle,
                                                 const char* parameter,
                                                 int64_t* out_len,
                                                 double* out_result);
-
-/*!
- * \brief Make sparse prediction for a new dataset in CSC format.  Currently only used for feature contributions.
- * \note
- * The outputs are pre-allocated, as they can vary for each invocation, but the shape should be the same:
- *   - for feature contributions, the shape of sparse matrix will be ``num_class * num_data * (num_feature + 1)``.
- * The output indptr_type for the sparse matrix will be the same as the given input indptr_type.
- * \param handle Handle of booster
- * \param col_ptr Pointer to column headers
- * \param col_ptr_type Type of ``col_ptr``, can be ``C_API_DTYPE_INT32`` or ``C_API_DTYPE_INT64``
- * \param indices Pointer to row indices
- * \param data Pointer to the data space
- * \param data_type Type of ``data`` pointer, can be ``C_API_DTYPE_FLOAT32`` or ``C_API_DTYPE_FLOAT64``
- * \param ncol_ptr Number of columns in the matrix + 1
- * \param nelem Number of nonzero elements in the matrix
- * \param num_row Number of rows
- * \param predict_type What should be predicted
- *   - ``C_API_PREDICT_CONTRIB``: feature contributions (SHAP values)
- * \param num_iteration Number of iteration for prediction, <= 0 means no limit
- * \param parameter Other parameters for prediction, e.g. early stopping for prediction
- * \param[out] out_len Length of output indices and data
- * \param[out] out_col_ptr Pointer to output column headers
- * \param[out] out_indices Pointer to sparse indices
- * \param[out] out_data Pointer to sparse data space
- * \return 0 when succeed, -1 when failure happens
- */
-LIGHTGBM_C_EXPORT int LGBM_BoosterPredictSparseForCSC(BoosterHandle handle,
-                                                      const void* col_ptr,
-                                                      int col_ptr_type,
-                                                      const int32_t* indices,
-                                                      const void* data,
-                                                      int data_type,
-                                                      int64_t ncol_ptr,
-                                                      int64_t nelem,
-                                                      int64_t num_row,
-                                                      int predict_type,
-                                                      int num_iteration,
-                                                      const char* parameter,
-                                                      int64_t* out_len,
-                                                      void** out_col_ptr,
-                                                      int32_t** out_indices,
-                                                      void** out_data);
 
 /*!
  * \brief Make prediction for a new dataset.
--- a/include/LightGBM/c_api.h
+++ b/include/LightGBM/c_api.h
@@ -746,13 +746,14 @@ LIGHTGBM_C_EXPORT int LGBM_BoosterPredictForCSR(BoosterHandle handle,
                                                 double* out_result);
 
 /*!
- * \brief Make sparse prediction for a new dataset in CSR or CSC format.  Currently only used for feature contributions.
+ * \brief Make sparse prediction for a new dataset in CSR or CSC format. Currently only used for feature contributions.
  * \note
  * The outputs are pre-allocated, as they can vary for each invocation, but the shape should be the same:
  *   - for feature contributions, the shape of sparse matrix will be ``num_class * num_data * (num_feature + 1)``.
  * The output indptr_type for the sparse matrix will be the same as the given input indptr_type.
+ * Call ``LGBM_BoosterFreePredictSparse`` to deallocate resources.
  * \param handle Handle of booster
- * \param indptr Pointer to row headers for CSR or col headers for CSC
+ * \param indptr Pointer to row headers for CSR or column headers for CSC
  * \param indptr_type Type of ``indptr``, can be ``C_API_DTYPE_INT32`` or ``C_API_DTYPE_INT64``
  * \param indices Pointer to column indices for CSR or row indices for CSC
  * \param data Pointer to the data space
@@ -764,9 +765,9 @@ LIGHTGBM_C_EXPORT int LGBM_BoosterPredictForCSR(BoosterHandle handle,
  *   - ``C_API_PREDICT_CONTRIB``: feature contributions (SHAP values)
  * \param num_iteration Number of iterations for prediction, <= 0 means no limit
  * \param parameter Other parameters for prediction, e.g. early stopping for prediction
- * \param matrix_type The type of matrix that is input and output, 0 for CSR and 1 for CSC
+ * \param matrix_type Type of matrix input and output, can be ``C_API_MATRIX_TYPE_CSR`` or ``C_API_MATRIX_TYPE_CSC``
  * \param[out] out_len Length of output indices and data
- * \param[out] out_indptr Pointer to output row headers for CSR or col headers for CSC
+ * \param[out] out_indptr Pointer to output row headers for CSR or column headers for CSC
  * \param[out] out_indices Pointer to sparse column indices for CSR or row indices for CSC
  * \param[out] out_data Pointer to sparse data space
  * \return 0 when succeed, -1 when failure happens
@@ -790,8 +791,8 @@ LIGHTGBM_C_EXPORT int LGBM_BoosterPredictSparseOutput(BoosterHandle handle,
                                                       void** out_data);
 
 /*!
- * \brief Method corresponding to LGBM_BoosterPredictSparseForCSR to free the allocated data.
- * \param indptr Pointer to output row headers or col headers to be deallocated
+ * \brief Method corresponding to ``LGBM_BoosterPredictSparseOutput`` to free the allocated data.
+ * \param indptr Pointer to output row headers or column headers to be deallocated
  * \param indices Pointer to sparse indices to be deallocated
  * \param data Pointer to sparse data space to be deallocated
  * \param indptr_type Type of ``indptr``, can be ``C_API_DTYPE_INT32`` or ``C_API_DTYPE_INT64``
--- a/include/LightGBM/c_api.h
+++ b/include/LightGBM/c_api.h
@@ -743,6 +743,59 @@ LIGHTGBM_C_EXPORT int LGBM_BoosterPredictForCSR(BoosterHandle handle,
                                                 double* out_result);
 
 /*!
+ * \brief Make sparse prediction for a new dataset in CSR format.  Currently only used for feature contributions.
+ * \note
+ * The outputs are pre-allocated, as they can vary for each invocation, but the shape should be the same:
+ *   - for feature contributions, the shape of sparse matrix will be ``num_class * num_data * (num_feature + 1)``.
+ * The output indptr_type for the sparse matrix will be the same as the given input indptr_type.
+ * \param handle Handle of booster
+ * \param indptr Pointer to row headers
+ * \param indptr_type Type of ``indptr``, can be ``C_API_DTYPE_INT32`` or ``C_API_DTYPE_INT64``
+ * \param indices Pointer to column indices
+ * \param data Pointer to the data space
+ * \param data_type Type of ``data`` pointer, can be ``C_API_DTYPE_FLOAT32`` or ``C_API_DTYPE_FLOAT64``
+ * \param nindptr Number of rows in the matrix + 1
+ * \param nelem Number of nonzero elements in the matrix
+ * \param num_col Number of columns
+ * \param predict_type What should be predicted, only feature contributions supported currently
+ *   - ``C_API_PREDICT_CONTRIB``: feature contributions (SHAP values)
+ * \param num_iteration Number of iterations for prediction, <= 0 means no limit
+ * \param parameter Other parameters for prediction, e.g. early stopping for prediction
+ * \param[out] out_len Length of output indices and data
+ * \param[out] out_indptr Pointer to output row headers
+ * \param[out] out_indices Pointer to sparse indices
+ * \param[out] out_data Pointer to sparse data space
+ * \return 0 when succeed, -1 when failure happens
+ */
+LIGHTGBM_C_EXPORT int LGBM_BoosterPredictSparseForCSR(BoosterHandle handle,
+                                                      const void* indptr,
+                                                      int indptr_type,
+                                                      const int32_t* indices,
+                                                      const void* data,
+                                                      int data_type,
+                                                      int64_t nindptr,
+                                                      int64_t nelem,
+                                                      int64_t num_col,
+                                                      int predict_type,
+                                                      int num_iteration,
+                                                      const char* parameter,
+                                                      int64_t* out_len,
+                                                      void** out_indptr,
+                                                      int32_t** out_indices,
+                                                      void** out_data);
+
+/*!
+ * \brief Method corresponding to LGBM_BoosterPredictSparseForCSR to free the allocated data.
+ * \param indptr Pointer to output row headers or col headers to be deallocated
+ * \param indices Pointer to sparse indices to be deallocated
+ * \param data Pointer to sparse data space to be deallocated
+ * \param indptr_type Type of ``indptr``, can be ``C_API_DTYPE_INT32`` or ``C_API_DTYPE_INT64``
+ * \param data_type Type of ``data`` pointer, can be ``C_API_DTYPE_FLOAT32`` or ``C_API_DTYPE_FLOAT64``
+ * \return 0 when succeed, -1 when failure happens
+ */
+LIGHTGBM_C_EXPORT int LGBM_BoosterFreePredictSparse(void* indptr, int32_t* indices, void* data, int indptr_type, int data_type);
+
+/*!
  * \brief Make prediction for a new dataset in CSR format. This method re-uses the internal predictor structure
  *        from previous calls and is optimized for single row invocation.
  * \note
@@ -826,6 +879,48 @@ LIGHTGBM_C_EXPORT int LGBM_BoosterPredictForCSC(BoosterHandle handle,
                                                 const char* parameter,
                                                 int64_t* out_len,
                                                 double* out_result);
+
+/*!
+ * \brief Make sparse prediction for a new dataset in CSC format.  Currently only used for feature contributions.
+ * \note
+ * The outputs are pre-allocated, as they can vary for each invocation, but the shape should be the same:
+ *   - for feature contributions, the shape of sparse matrix will be ``num_class * num_data * (num_feature + 1)``.
+ * The output indptr_type for the sparse matrix will be the same as the given input indptr_type.
+ * \param handle Handle of booster
+ * \param col_ptr Pointer to column headers
+ * \param col_ptr_type Type of ``col_ptr``, can be ``C_API_DTYPE_INT32`` or ``C_API_DTYPE_INT64``
+ * \param indices Pointer to row indices
+ * \param data Pointer to the data space
+ * \param data_type Type of ``data`` pointer, can be ``C_API_DTYPE_FLOAT32`` or ``C_API_DTYPE_FLOAT64``
+ * \param ncol_ptr Number of columns in the matrix + 1
+ * \param nelem Number of nonzero elements in the matrix
+ * \param num_row Number of rows
+ * \param predict_type What should be predicted
+ *   - ``C_API_PREDICT_CONTRIB``: feature contributions (SHAP values)
+ * \param num_iteration Number of iteration for prediction, <= 0 means no limit
+ * \param parameter Other parameters for prediction, e.g. early stopping for prediction
+ * \param[out] out_len Length of output indices and data
+ * \param[out] out_col_ptr Pointer to output column headers
+ * \param[out] out_indices Pointer to sparse indices
+ * \param[out] out_data Pointer to sparse data space
+ * \return 0 when succeed, -1 when failure happens
+ */
+LIGHTGBM_C_EXPORT int LGBM_BoosterPredictSparseForCSC(BoosterHandle handle,
+                                                      const void* col_ptr,
+                                                      int col_ptr_type,
+                                                      const int32_t* indices,
+                                                      const void* data,
+                                                      int data_type,
+                                                      int64_t ncol_ptr,
+                                                      int64_t nelem,
+                                                      int64_t num_row,
+                                                      int predict_type,
+                                                      int num_iteration,
+                                                      const char* parameter,
+                                                      int64_t* out_len,
+                                                      void** out_col_ptr,
+                                                      int32_t** out_indices,
+                                                      void** out_data);
 
 /*!
  * \brief Make prediction for a new dataset.
--- a/include/LightGBM/meta.h
+++ b/include/LightGBM/meta.h
@@ -5,9 +5,9 @@
 #ifndef LIGHTGBM_META_H_
 #define LIGHTGBM_META_H_
 
-#include <limits>
 #include <cstdint>
 #include <functional>
+#include <limits>
 #include <memory>
 #include <unordered_map>
 #include <utility>
--- a/include/LightGBM/meta.h
+++ b/include/LightGBM/meta.h
@@ -11,6 +11,7 @@
 #include <memory>
 #include <utility>
 #include <vector>
+#include <unordered_map>
 
 #if (defined(_MSC_VER) && (defined(_M_IX86) || defined(_M_AMD64))) || defined(__INTEL_COMPILER) || MM_PREFETCH
   #include <xmmintrin.h>
@@ -57,6 +58,9 @@ typedef int32_t comm_size_t;
 
 using PredictFunction =
 std::function<void(const std::vector<std::pair<int, double>>&, double* output)>;
+
+using PredictSparseFunction =
+std::function<void(const std::vector<std::pair<int, double>>&, std::vector<std::unordered_map<int, double>>* output)>;
 
 typedef void(*ReduceFunction)(const char* input, char* output, int type_size, comm_size_t array_size);
 
--- a/include/LightGBM/meta.h
+++ b/include/LightGBM/meta.h
@@ -9,9 +9,9 @@
 #include <cstdint>
 #include <functional>
 #include <memory>
+#include <unordered_map>
 #include <utility>
 #include <vector>
-#include <unordered_map>
 
 #if (defined(_MSC_VER) && (defined(_M_IX86) || defined(_M_AMD64))) || defined(__INTEL_COMPILER) || MM_PREFETCH
   #include <xmmintrin.h>
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -538,7 +538,8 @@ class _InnerPredictor(object):
         Returns
         -------
         result : numpy array, scipy.sparse or list of scipy.sparse
-            Prediction result, can be sparse for feature contributions (when ``pred_contrib=True``).
+            Prediction result.
+            Can be sparse or a list of sparse objects (each element represents predictions for one class) for feature contributions (when ``pred_contrib=True``).
         """
         if isinstance(data, Dataset):
             raise TypeError("Cannot use Dataset instance for prediction, please use raw data instead")
@@ -2817,7 +2818,8 @@ class Booster(object):
         Returns
         -------
         result : numpy array, scipy.sparse or list of scipy.sparse
-            Prediction result, can be sparse for feature contributions (when ``pred_contrib=True``).
+            Prediction result.
+            Can be sparse or a list of sparse objects (each element represents predictions for one class) for feature contributions (when ``pred_contrib=True``).
         """
         predictor = self._to_predictor(copy.deepcopy(kwargs))
         if num_iteration is None:
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -116,7 +116,15 @@ def cint32_array_to_numpy(cptr, length):
     if isinstance(cptr, ctypes.POINTER(ctypes.c_int32)):
         return np.fromiter(cptr, dtype=np.int32, count=length)
     else:
-        raise RuntimeError('Expected int pointer')
+        raise RuntimeError('Expected int32 pointer')
+
+
+def cint64_array_to_numpy(cptr, length):
+    """Convert a ctypes int pointer array to a numpy array."""
+    if isinstance(cptr, ctypes.POINTER(ctypes.c_int64)):
+        return np.fromiter(cptr, dtype=np.int64, count=length)
+    else:
+        raise RuntimeError('Expected int64 pointer')
 
 
 def c_str(string):
@@ -656,30 +664,24 @@ class _InnerPredictor(object):
         else:
             return inner_predict(mat, num_iteration, predict_type)
 
-    def __convert_ctypes_to_numpy(self, ctypes_array, ctypes_array_len, array_type):
-        """Convert the ctypes array to a numpy array, note memory will still need to be managed."""
-        array_size = np.dtype(array_type).itemsize * ctypes_array_len
-        if sys.version_info.major >= 3:
-            bfm = ctypes.pythonapi.PyMemoryView_FromMemory
-            bfm.restype = ctypes.py_object
-            bfm.argtypes = (ctypes.c_void_p, ctypes.c_int, ctypes.c_int)
-            PyBUF_READ = 0x100
-            py_buffer = bfm(ctypes_array, array_size, PyBUF_READ)
-        else:
-            bfm = ctypes.pythonapi.PyBuffer_FromMemory
-            bfm.restype = ctypes.py_object
-            bfm.argtypes = (ctypes.c_void_p, ctypes.c_int)
-            py_buffer = bfm(ctypes_array, array_size)
-        return np.frombuffer(py_buffer, array_type)
-
     def __create_sparse_native(self, cs, out_shape, out_ptr_indptr, out_ptr_indices, out_ptr_data,
-                               np_indptr_type, np_data_type, indptr_type, data_type, is_csr=True):
+                               indptr_type, data_type, is_csr=True):
         # create numpy array from output arrays
         data_indices_len = out_shape[0]
         indptr_len = out_shape[1]
-        out_indptr = self.__convert_ctypes_to_numpy(out_ptr_indptr, indptr_len, np_indptr_type)
-        out_indices = self.__convert_ctypes_to_numpy(out_ptr_indices, data_indices_len, np.int32)
-        out_data = self.__convert_ctypes_to_numpy(out_ptr_data, data_indices_len, np_data_type)
+        if indptr_type == C_API_DTYPE_INT32:
+            out_indptr = cint32_array_to_numpy(out_ptr_indptr, indptr_len)
+        elif indptr_type == C_API_DTYPE_INT64:
+            out_indptr = cint64_array_to_numpy(out_ptr_indptr, indptr_len)
+        else:
+            raise TypeError("Expected int32 or int64 type for indptr")
+        if data_type == C_API_DTYPE_FLOAT32:
+            out_data = cfloat32_array_to_numpy(out_ptr_data, data_indices_len)
+        elif data_type == C_API_DTYPE_FLOAT64:
+            out_data = cfloat64_array_to_numpy(out_ptr_data, data_indices_len)
+        else:
+            raise TypeError("Expected float32 or float64 type for data")
+        out_indices = cint32_array_to_numpy(out_ptr_indices, data_indices_len)
         # break up indptr based on number of rows (note more than one matrix in multiclass case)
         per_class_indptr_shape = cs.indptr.shape[0]
         # for CSC there is extra column added
@@ -698,9 +700,9 @@ class _InnerPredictor(object):
             cs_shape = [cs.shape[0], cs.shape[1] + 1]
             # note: make sure we copy data as it will be deallocated next
             if is_csr:
-                cs_output_matrices.append(scipy.sparse.csr_matrix((cs_data, cs_indices, cs_indptr), cs_shape, copy=True))
+                cs_output_matrices.append(scipy.sparse.csr_matrix((cs_data, cs_indices, cs_indptr), cs_shape))
             else:
-                cs_output_matrices.append(scipy.sparse.csc_matrix((cs_data, cs_indices, cs_indptr), cs_shape, copy=True))
+                cs_output_matrices.append(scipy.sparse.csc_matrix((cs_data, cs_indices, cs_indptr), cs_shape))
             # free the temporary native indptr, indices, and data
         _safe_call(_LIB.LGBM_BoosterFreePredictSparse(out_ptr_indptr, out_ptr_indices, out_ptr_data,
                                                       ctypes.c_int(indptr_type), ctypes.c_int(data_type)))
@@ -750,17 +752,13 @@ class _InnerPredictor(object):
             csr_indices = csr.indices.astype(np.int32, copy=False)
             matrix_type = C_API_MATRIX_TYPE_CSR
             if type_ptr_indptr == C_API_DTYPE_INT32:
-                np_indptr_type = np.int32
                 out_ptr_indptr = ctypes.POINTER(ctypes.c_int32)()
             else:
-                np_indptr_type = np.int64
                 out_ptr_indptr = ctypes.POINTER(ctypes.c_int64)()
             out_ptr_indices = ctypes.POINTER(ctypes.c_int32)()
             if type_ptr_data == C_API_DTYPE_FLOAT32:
-                np_data_type = np.float32
                 out_ptr_data = ctypes.POINTER(ctypes.c_float)()
             else:
-                np_data_type = np.float64
                 out_ptr_data = ctypes.POINTER(ctypes.c_double)()
             out_shape = np.zeros(2, dtype=np.int64)
             _safe_call(_LIB.LGBM_BoosterPredictSparseOutput(
@@ -782,8 +780,7 @@ class _InnerPredictor(object):
                 ctypes.byref(out_ptr_indices),
                 ctypes.byref(out_ptr_data)))
             matrices = self.__create_sparse_native(csr, out_shape, out_ptr_indptr, out_ptr_indices, out_ptr_data,
-                                                   np_indptr_type, np_data_type, type_ptr_indptr,
-                                                   type_ptr_data, is_csr=True)
+                                                   type_ptr_indptr, type_ptr_data, is_csr=True)
             nrow = len(csr.indptr) - 1
             return matrices, nrow
 
@@ -812,17 +809,13 @@ class _InnerPredictor(object):
             csc_indices = csc.indices.astype(np.int32, copy=False)
             matrix_type = C_API_MATRIX_TYPE_CSC
             if type_ptr_indptr == C_API_DTYPE_INT32:
-                np_indptr_type = np.int32
                 out_ptr_indptr = ctypes.POINTER(ctypes.c_int32)()
             else:
-                np_indptr_type = np.int64
                 out_ptr_indptr = ctypes.POINTER(ctypes.c_int64)()
             out_ptr_indices = ctypes.POINTER(ctypes.c_int32)()
             if type_ptr_data == C_API_DTYPE_FLOAT32:
-                np_data_type = np.float32
                 out_ptr_data = ctypes.POINTER(ctypes.c_float)()
             else:
-                np_data_type = np.float64
                 out_ptr_data = ctypes.POINTER(ctypes.c_double)()
             out_shape = np.zeros(2, dtype=np.int64)
             _safe_call(_LIB.LGBM_BoosterPredictSparseOutput(
@@ -844,8 +837,7 @@ class _InnerPredictor(object):
                 ctypes.byref(out_ptr_indices),
                 ctypes.byref(out_ptr_data)))
             matrices = self.__create_sparse_native(csc, out_shape, out_ptr_indptr, out_ptr_indices, out_ptr_data,
-                                                   np_indptr_type, np_data_type, type_ptr_indptr,
-                                                   type_ptr_data, is_csr=False)
+                                                   type_ptr_indptr, type_ptr_data, is_csr=False)
             nrow = csc.shape[0]
             return matrices, nrow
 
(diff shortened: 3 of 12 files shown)


include/LightGBM/c_api.h
```diff
@@ -726,6 +726,45 @@ LIGHTGBM_C_EXPORT int LGBM_BoosterPredictForCSR(BoosterHandle handle,
...
+						      const int32_t* indices,
+						      const void* data,
+						      int data_type,
+						      int64_t nindptr,
+						      int64_t nelem,
+						      int64_t num_col,
+						      int predict_type,
+						      int num_iteration,
+						      const char* parameter,
+						      int64_t* out_len,
```
imatiach-msft, line 764: wasn't sure about the interface here, does it seem reasonable to have a separate method that returns the indices and data when returning a sparse matrix?

include/LightGBM/c_api.h
```diff
@@ -726,6 +726,45 @@ LIGHTGBM_C_EXPORT int LGBM_BoosterPredictForCSR(BoosterHandle handle,
                                                 int64_t* out_len,
                                                 double* out_result);
 
+/*!
+ * \brief Make sparse prediction for a new dataset in CSR format.  Currently only used for feature contributions.
+ * \note
+ * The outputs are pre-allocated, as they can vary for each invocation, but the shape should be the same:
```
imatiach-msft, line 732: unlike other methods, the user can't pre-allocate because they can't anticipate what the size of the indices and data will be for the sparse result.  In that case, is it reasonable for us to do the sparse matrix allocation within the method?

src/application/predictor.hpp
```diff
@@ -87,13 +87,19 @@ class Predictor {
...
+	// get feature importances
+	boosting_->PredictContrib(predict_buf_[tid].data(), output);
+	ClearPredictBuffer(predict_buf_[tid].data(), predict_buf_[tid].size(),
+			   features);
+      };
+      predict_sparse_fun_ = [=](const std::vector<std::pair<int, double>>& features,
+			        std::vector<std::unordered_map<int, double>>& output) {
+	auto buf = CopyToPredictMap(features);
+	// get sparse feature importances
+	boosting_->PredictContribByMap(buf, output);
```
guolinke, line 100: fix the indent in these lines?

guolinke:
@imatiach-msft ping me when this is ready to review.

src/application/predictor.hpp
imatiach-msft: done!

imatiach-msft:
@guolinke thank you, pinging as this PR is ready for review, also tagging @slundberg for review

src/boosting/gbdt.cpp
```diff
@@ -581,17 +580,20 @@ void GBDT::PredictContrib(const double* features, double* output, const Predicti
     for (int k = 0; k < num_tree_per_iteration_; ++k) {
       models_[i * num_tree_per_iteration_ + k]->PredictContrib(features, num_features, output + k*(num_features + 1));
     }
-    // check early stopping
-    ++early_stop_round_counter;
```
imatiach-msft, line 585: @slundberg checking if this is ok to remove, I don't think early stopping is ever used/applies here

slundberg: if you mean that early stopping is not exposed in the external API, then I would think we should expose it rather than remove support. If I train a model with 10k trees and then decide to use just the first 3k for prediction then I think we want the same ability to explain the model that only uses the first 3k trees.

slundberg:
Other than my early stopping comment above it all looks good to me. Great work!

src/boosting/gbdt.cpp
slundberg: @imatiach-msft clarified that this is for the number of samples not ntrees so I agree it can go.

include/LightGBM/c_api.h
guolinke: it is okay, but aware to avoid memory leak.

python-package/lightgbm/basic.py
```diff
@@ -652,6 +705,7 @@ def inner_predict(csr, num_iteration, predict_type, preds=None):
             ptr_data, type_ptr_data, _ = c_float_array(csr.data)
 
             assert csr.shape[1] <= MAX_INT32
+            # Note: not sure why this was csr.indices, we shouldn't be modifying the input matrix
```
guolinke, line 708: yeah, maybe a temp var should be used

python-package/lightgbm/basic.py
```diff
@@ -637,6 +638,58 @@ def inner_predict(mat, num_iteration, predict_type, preds=None):
...
+            bfm.restype = ctypes.py_object
+            bfm.argtypes = (ctypes.c_void_p, ctypes.c_int, ctypes.c_int)
+            PyBUF_READ = 0x100
+            py_buffer = bfm(ctypes_array, array_size, PyBUF_READ)
+        else:
+            bfm = ctypes.pythonapi.PyBuffer_FromMemory
+            bfm.restype = ctypes.py_object
+            bfm.argtypes = (ctypes.c_void_p, ctypes.c_int)
+            py_buffer = bfm(ctypes_array, array_size)
+        return np.frombuffer(py_buffer, array_type)
```
guolinke, line 655: difference of this function to https://github.com/microsoft/LightGBM/blob/f155379c1fe4e53d5d1c7fd195987a8d3aec4331/python-package/lightgbm/basic.py#L88-L109 ?

src/c_api.cpp
```diff
@@ -422,6 +433,214 @@ class Booster {
...
+      Log::Fatal("Unknown indptr type in PredictSparseCSR");
+      return;
+    }
+    // aggregated per row feature contribution results
+    std::vector<std::vector<std::unordered_map<int, double>>> agg(nrow);
+    int elements_size = 0;
+    PredictSparse(num_iteration, predict_type, nrow, ncol, get_row_fun, config, &elements_size, &agg,
+                  out_indices, out_data, data_type, &is_data_float32, num_matrices);
+    // copy vector results to output for each row
+    // TODO(imatiach-msft): should this be parallelized, or is there no perf benefit?
```
guolinke, line 504: you can parallel the copy of `out_data` and `out_indptr'.

include/LightGBM/c_api.h
guolinke: BTW, these new apis are for the sparse output? Maybe we can use a clearer name to indicate this.

include/LightGBM/c_api.h
```diff
@@ -812,6 +865,48 @@ LIGHTGBM_C_EXPORT int LGBM_BoosterPredictForCSC(BoosterHandle handle,
...
+                                                      int64_t ncol_ptr,
+                                                      int64_t nelem,
+                                                      int64_t num_row,
+                                                      int predict_type,
+                                                      int num_iteration,
+                                                      const char* parameter,
+                                                      int64_t* out_len,
+                                                      void** out_col_ptr,
+                                                      int32_t** out_indices,
+                                                      void** out_data);
```
guolinke, line 908: maybe merge the csr and csc into one interface? and use a new argument, like `matrix_type` to distinguish csr or csc?

include/LightGBM/c_api.h
imatiach-msft: discussed in comment below, seems reasonable, resolving comment

include/LightGBM/c_api.h
imatiach-msft: sure, maybe just:
```
LGBM_BoosterPredictSparseOutput
```
since we are going to unify the CSR and CSC APIs

include/LightGBM/c_api.h
imatiach-msft: sure, will do.  I was trying to go along with the APIs:
LGBM_BoosterPredictForCSC
LGBM_BoosterPredictForCSR
which are two separate methods.  If we are going to combine these into one API it might make sense to combine those two methods as well.

python-package/lightgbm/basic.py
imatiach-msft: done!

include/LightGBM/c_api.h
imatiach-msft: done!

python-package/lightgbm/basic.py
imatiach-msft: great question, I wasn't sure myself at first what the difference is between np.frombuffer and np.fromiter, but based on this:
https://stackoverflow.com/questions/58091879/whats-the-difference-between-frombuffer-and-fromiter-in-numpywhy-and-when-to-us
and docs here:
https://numpy.org/doc/stable/reference/generated/numpy.frombuffer.html
https://numpy.org/doc/stable/reference/generated/numpy.fromiter.html

it looks like frombuffer uses the existing ctypes array whereas fromiter allocates a new array, so I think it's better to use frombuffer in order to reduce memory usage and data copying.

```
 numpy.frombuffer(buffer, dtype=float, count=-1, offset=0)
    Interpret a buffer as a 1-dimensional array.
```
```
 numpy.fromiter(iterable, dtype, count=-1)
    Create a new 1-dimensional array from an iterable object.
```

src/c_api.cpp
imatiach-msft: will do, working on this now, running into some indexing errors in the parallel code

imatiach-msft: done, parallelized the CSR matrix code

imatiach-msft:
@guolinke might you be able to take another look at this PR when you have time?  I've updated it based on your comments.  Thank you!

python-package/lightgbm/basic.py
guolinke: great ! also ping @StrikerRUS , for the performance, should we avoid the copy as possible?

guolinke: @imatiach-msft however, as these memories are in c++, you need a copy after that. like line 696 and 698.

imatiach-msft: @guolinke good point, I am doing a copy regardless afterward, since the array was allocated from native c++ and numpy can't track it.  Do you think it would be better to try to use the existing methods and do the copy then, to avoid the copies on lines 696 and 698, or should I leave the code as-is?

StrikerRUS: @guolinke 
> also ping @StrikerRUS , for the performance, should we avoid the copy as possible?

yeah, absolutely! I remember we had some reports about that conversion from `pandas` used to double the memory usage and we kind of fixed it.

guolinke: @imatiach-msft if there are no performance gain, we think we should keep the code simple, and reuse the existing methods as possible.

imatiach-msft: done! changed the code to use the existing methods that utilize fromiter instead of frombuffer, re-ran the test and it still seems to be working

imatiach-msft:
close-reopen due to Travis URL check timeout:

```
URL        `https://webscope.sandbox.yahoo.com/catalog.php?datatype=c'

Name       `link4'

Parent URL file:///home/travis/build/microsoft/LightGBM/docs/_build/html/GPU-Performance.html, line 265, col 8

Real URL   https://webscope.sandbox.yahoo.com/catalog.php?datatype=c

Check time 66.212 seconds

Result     Error: ReadTimeout: HTTPSConnectionPool(host='webscope.sandbox.yahoo.com', port=443): Read timed out. (read timeout=60)
```

imatiach-msft:
seems this Docs build is ruthless, another HTTP error, closing and reopening the PR again to retrigger CI:
```
URL        `https://www.microsoft.com/en-us/download/details.aspx?id=100593'

Name       `MS MPI'

Parent URL file:///home/travis/build/microsoft/LightGBM/docs/_build/html/Installation-Guide.html, line 569, col 28

Real URL   https://www.microsoft.com/en-us/download/details.aspx?id=100593

Check time 61.938 seconds

Result     Error: ReadTimeout: HTTPSConnectionPool(host='www.microsoft.com', port=443): Read timed out. (read timeout=60)
```

imatiach-msft:
@guolinke might you be able to take another look at this PR when you have time? I've updated it based on your comments. Thank you!

guolinke:
Thanks @imatiach-msft , looks good to me

python-package/lightgbm/basic.py
```diff
@@ -646,6 +659,52 @@ def inner_predict(mat, num_iteration, predict_type, preds=None):
...
+            cs_data = out_data[offset + cs_indptr[0]:offset + matrix_indptr_len]
+            offset += matrix_indptr_len
+            # same shape as input csr or csc matrix except extra column for expected value
+            cs_shape = [cs.shape[0], cs.shape[1] + 1]
+            # note: make sure we copy data as it will be deallocated next
+            if is_csr:
+                cs_output_matrices.append(scipy.sparse.csr_matrix((cs_data, cs_indices, cs_indptr), cs_shape))
+            else:
+                cs_output_matrices.append(scipy.sparse.csc_matrix((cs_data, cs_indices, cs_indptr), cs_shape))
+            # free the temporary native indptr, indices, and data
```
StrikerRUS, line 701:
```suggestion
        # free the temporary native indptr, indices, and data
```

python-package/lightgbm/basic.py
```diff
@@ -682,6 +741,46 @@ def inner_predict(csr, num_iteration, predict_type, preds=None):
...
+                ctypes.byref(out_ptr_indptr),
+                ctypes.byref(out_ptr_indices),
+                ctypes.byref(out_ptr_data)))
+            matrices = self.__create_sparse_native(csr, out_shape, out_ptr_indptr, out_ptr_indices, out_ptr_data,
+                                                   type_ptr_indptr, type_ptr_data, is_csr=True)
+            nrow = len(csr.indptr) - 1
+            return matrices, nrow
+
+        if predict_type == C_API_PREDICT_CONTRIB:
+            return inner_predict_sparse(csr, num_iteration, predict_type)
```
StrikerRUS, line 783: Is possibility to predict feature contributions for matrices with `nrow > MAX_INT32` preserved?

src/c_api.cpp
```diff
@@ -422,6 +433,237 @@ class Booster {
...
+      OMP_LOOP_EX_BEGIN();
+      auto one_row = get_row_fun(i);
+      agg[i] = std::vector<std::unordered_map<int, double>>(num_matrices);
+      pred_sparse_fun(one_row, &agg[i]);
+      OMP_LOOP_EX_END();
+    }
+    OMP_THROW_EX();
+    // calculate the nonzero data and indices size
+    int elements_size = 0;
+    for (int i = 0; i < static_cast<int>(agg.size()); i++) {
```
StrikerRUS, line 457:
```suggestion
    for (int i = 0; i < static_cast<int>(agg.size()); ++i) {
```

src/c_api.cpp
```diff
@@ -422,6 +433,237 @@ class Booster {
...
+      auto row_vector = agg[i];
+      for (int j = 0; j < static_cast<int>(row_vector.size()); j++) {
```
StrikerRUS, line 459:
```suggestion
      for (int j = 0; j < static_cast<int>(row_vector.size()); ++j) {
```

src/c_api.cpp
```diff
@@ -422,6 +433,237 @@ class Booster {
...
+    }
+    // aggregated per row feature contribution results
+    std::vector<std::vector<std::unordered_map<int, double>>> agg(nrow);
+    int elements_size = 0;
+    PredictSparse(num_iteration, predict_type, nrow, ncol, get_row_fun, config, &elements_size, &agg,
+                  out_indices, out_data, data_type, &is_data_float32, num_matrices);
+    std::vector<int> row_sizes(num_matrices * nrow);
+    std::vector<int> row_matrix_offsets(num_matrices * nrow);
+    int row_vector_cnt = 0;
+    for (int m = 0; m < num_matrices; m++) {
```
StrikerRUS, line 506:
```suggestion
    for (int m = 0; m < num_matrices; ++m) {
```

src/c_api.cpp
```diff
@@ -422,6 +433,237 @@ class Booster {
...
+      for (int i = 0; i < static_cast<int>(agg.size()); i++) {
```
StrikerRUS, line 507:
```suggestion
      for (int i = 0; i < static_cast<int>(agg.size()); ++i) {
```

src/c_api.cpp
```diff
@@ -422,6 +433,237 @@ class Booster {
...
+          row_matrix_offsets[row_vector_cnt] = 0;
+        } else {
+          row_matrix_offsets[row_vector_cnt] = row_sizes[row_vector_cnt - 1] + row_matrix_offsets[row_vector_cnt - 1];
+        }
+        row_vector_cnt++;
+      }
+    }
+    // copy vector results to output for each row
+    int indptr_index = 0;
+    for (int m = 0; m < num_matrices; m++) {
```
StrikerRUS, line 522:
```suggestion
    for (int m = 0; m < num_matrices; ++m) {
```

src/c_api.cpp
```diff
@@ -422,6 +433,237 @@ class Booster {
...
+      if (is_indptr_int32) {
+        (reinterpret_cast<int32_t*>(*out_indptr))[indptr_index] = 0;
+      } else {
+        (reinterpret_cast<int64_t*>(*out_indptr))[indptr_index] = 0;
+      }
+      indptr_index++;
+      int matrix_start_index = m * agg.size();
+      OMP_INIT_EX();
+      #pragma omp parallel for schedule(static)
+      for (int i = 0; i < static_cast<int>(agg.size()); i++) {
```
StrikerRUS, line 532:
```suggestion
      for (int i = 0; i < static_cast<int>(agg.size()); ++i) {
```

src/c_api.cpp
```diff
@@ -422,6 +433,237 @@ class Booster {
...
+    }
+    // aggregated per row feature contribution results
+    std::vector<std::vector<std::unordered_map<int, double>>> agg(nrow);
+    int elements_size = 0;
+    PredictSparse(num_iteration, predict_type, nrow, ncol, get_row_fun, config, &elements_size, &agg,
+                  out_indices, out_data, data_type, &is_data_float32, num_matrices);
+    // calculate number of elements per column to construct
+    // the CSC matrix with random access
+    std::vector<std::vector<int>> column_sizes(num_matrices);
+    for (int m = 0; m < num_matrices; m++) {
```
StrikerRUS, line 594:
```suggestion
    for (int m = 0; m < num_matrices; ++m) {
```

src/c_api.cpp
```diff
@@ -422,6 +433,237 @@ class Booster {
...
+      column_sizes[m] = std::vector<int>(num_output_cols, 0);
+      for (int i = 0; i < static_cast<int>(agg.size()); i++) {
```
StrikerRUS, line 596:
```suggestion
      for (int i = 0; i < static_cast<int>(agg.size()); ++i) {
```

src/c_api.cpp
```diff
@@ -422,6 +433,237 @@ class Booster {
...
+      }
+    }
+    // keep track of column counts
+    std::vector<std::vector<int>> column_counts(num_matrices);
+    // keep track of beginning index for each column
+    std::vector<std::vector<int>> column_start_indices(num_matrices);
+    // keep track of beginning index for each matrix
+    std::vector<int> matrix_start_indices(num_matrices, 0);
+    int col_ptr_index = 0;
+    for (int m = 0; m < num_matrices; m++) {
```
StrikerRUS, line 610:
```suggestion
    for (int m = 0; m < num_matrices; ++m) {
```

src/c_api.cpp
```diff
@@ -422,6 +433,237 @@ class Booster {
...
+      int col_ptr_value = 0;
+      column_start_indices[m] = std::vector<int>(num_output_cols, 0);
+      column_counts[m] = std::vector<int>(num_output_cols, 0);
+      if (is_col_ptr_int32) {
+        (reinterpret_cast<int32_t*>(*out_col_ptr))[col_ptr_index] = col_ptr_value;
+      } else {
+        (reinterpret_cast<int64_t*>(*out_col_ptr))[col_ptr_index] = col_ptr_value;
+      }
+      col_ptr_index++;
+      for (int i = 1; i < static_cast<int>(column_sizes[m].size()); i++) {
```
StrikerRUS, line 620:
```suggestion
      for (int i = 1; i < static_cast<int>(column_sizes[m].size()); ++i) {
```

src/c_api.cpp
```diff
@@ -422,6 +433,237 @@ class Booster {
...
+      } else {
+        (reinterpret_cast<int64_t*>(*out_col_ptr))[col_ptr_index] = last_column_start_index + last_column_size;
+      }
+      if (m != 0) {
+        matrix_start_indices[m] = matrix_start_indices[m - 1] +
+          last_column_start_index +
+          last_column_size;
+      }
+    }
+    for (int m = 0; m < num_matrices; m++) {
```
StrikerRUS, line 643:
```suggestion
    for (int m = 0; m < num_matrices; ++m) {
```

src/c_api.cpp
```diff
@@ -422,6 +433,237 @@ class Booster {
...
+      for (int i = 0; i < static_cast<int>(agg.size()); i++) {
```
StrikerRUS, line 644:
```suggestion
      for (int i = 0; i < static_cast<int>(agg.size()); ++i) {
```

src/io/tree.cpp
```diff
@@ -727,6 +727,56 @@ void Tree::TreeSHAP(const double *feature_values, double *phi,
...
 }
 
+// recursive sparse computation of SHAP values for a decision tree
+void Tree::TreeSHAPByMap(const std::unordered_map<int, double>& feature_values, std::unordered_map<int, double>* phi,
+                         int node, int unique_depth,
+                         PathElement *parent_unique_path, double parent_zero_fraction,
+                         double parent_one_fraction, int parent_feature_index) const {
+  // extend the unique path
+  PathElement* unique_path = parent_unique_path + unique_depth;
+  if (unique_depth > 0) std::copy(parent_unique_path, parent_unique_path + unique_depth, unique_path);
```
StrikerRUS, line 737:
```suggestion
  if (unique_depth > 0) {
    std::copy(parent_unique_path, parent_unique_path + unique_depth, unique_path);
  }
```

src/io/tree.cpp
```diff
@@ -727,6 +727,56 @@ void Tree::TreeSHAP(const double *feature_values, double *phi,
...
+
+  // leaf node
+  if (node < 0) {
+    for (int i = 1; i <= unique_depth; ++i) {
+      const double w = UnwoundPathSum(unique_path, unique_depth, i);
+      const PathElement &el = unique_path[i];
+      (*phi)[el.feature_index] += w*(el.one_fraction - el.zero_fraction)*leaf_value_[~node];
+    }
+
+    // internal node
```
StrikerRUS, line 749:
```suggestion
  // internal node
```

tests/python_package_test/test_engine.py
```diff
@@ -14,6 +14,7 @@
                               load_iris, load_svmlight_file)
 from sklearn.metrics import log_loss, mean_absolute_error, mean_squared_error, roc_auc_score
 from sklearn.model_selection import train_test_split, TimeSeriesSplit, GroupKFold
+from sklearn.datasets import make_multilabel_classification
```
StrikerRUS, line 17: There is already `from sklearn.datasets import ...` at L13.

tests/python_package_test/test_engine.py
```diff
@@ -941,6 +942,40 @@ def test_contribs(self):
...
+        self.assertLess(np.linalg.norm(gbm.predict(X_test, raw_score=True)
+                                       - np.sum(contribs_dense, axis=1)), 1e-4)
+        # validate using CSC matrix
+        X_train_csc = X_train.tocsc()
+        X_test_csc = X_test.tocsc()
+        lgb_train = lgb.Dataset(X_train_csc, y_train)
+        gbm = lgb.train(params, lgb_train, num_boost_round=20)
+        contribs_csc = gbm.predict(X_test_csc, pred_contrib=True)
+        # validate the values are the same
+        np.testing.assert_allclose(contribs_csr.toarray(), contribs_csc.toarray())
```
StrikerRUS, line 977: To save CI time.
```suggestion
        np.testing.assert_allclose(contribs_dense, contribs_csc.toarray())
```

include/LightGBM/c_api.h
```diff
@@ -734,6 +737,61 @@ LIGHTGBM_C_EXPORT int LGBM_BoosterPredictForCSR(BoosterHandle handle,
                                                 int64_t* out_len,
                                                 double* out_result);
 
+/*!
+ * \brief Make sparse prediction for a new dataset in CSR or CSC format.  Currently only used for feature contributions.
```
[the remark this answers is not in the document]
imatiach-msft: done!

include/LightGBM/c_api.h
```diff
@@ -734,6 +737,61 @@ LIGHTGBM_C_EXPORT int LGBM_BoosterPredictForCSR(BoosterHandle handle,
...
+ * \note
+ * The outputs are pre-allocated, as they can vary for each invocation, but the shape should be the same:
+ *   - for feature contributions, the shape of sparse matrix will be ``num_class * num_data * (num_feature + 1)``.
+ * The output indptr_type for the sparse matrix will be the same as the given input indptr_type.
+ * \param handle Handle of booster
+ * \param indptr Pointer to row headers for CSR or col headers for CSC
```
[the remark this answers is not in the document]
imatiach-msft: done!

include/LightGBM/c_api.h
```diff
@@ -734,6 +737,61 @@ LIGHTGBM_C_EXPORT int LGBM_BoosterPredictForCSR(BoosterHandle handle,
...
+ * \param data Pointer to the data space
+ * \param data_type Type of ``data`` pointer, can be ``C_API_DTYPE_FLOAT32`` or ``C_API_DTYPE_FLOAT64``
+ * \param nindptr Number of rows in the matrix + 1
+ * \param nelem Number of nonzero elements in the matrix
+ * \param num_col_or_row Number of columns for CSR or number of rows for CSC
+ * \param predict_type What should be predicted, only feature contributions supported currently
+ *   - ``C_API_PREDICT_CONTRIB``: feature contributions (SHAP values)
+ * \param num_iteration Number of iterations for prediction, <= 0 means no limit
+ * \param parameter Other parameters for prediction, e.g. early stopping for prediction
+ * \param matrix_type The type of matrix that is input and output, 0 for CSR and 1 for CSC
```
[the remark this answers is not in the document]
imatiach-msft: done!
changed to:
```
Type of matrix input and output, can be ``C_API_MATRIX_TYPE_CSR`` or ``C_API_MATRIX_TYPE_CSC``
```

include/LightGBM/c_api.h
```diff
@@ -734,6 +737,61 @@ LIGHTGBM_C_EXPORT int LGBM_BoosterPredictForCSR(BoosterHandle handle,
...
+ * \param[out] out_len Length of output indices and data
+ * \param[out] out_indptr Pointer to output row headers for CSR or col headers for CSC
```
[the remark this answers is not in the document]
imatiach-msft: done!

include/LightGBM/c_api.h
```diff
@@ -734,6 +737,61 @@ LIGHTGBM_C_EXPORT int LGBM_BoosterPredictForCSR(BoosterHandle handle,
...
+                                                      int num_iteration,
+                                                      const char* parameter,
+                                                      int matrix_type,
+                                                      int64_t* out_len,
+                                                      void** out_indptr,
+                                                      int32_t** out_indices,
+                                                      void** out_data);
+
+/*!
+ * \brief Method corresponding to LGBM_BoosterPredictSparseForCSR to free the allocated data.
```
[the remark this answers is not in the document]
imatiach-msft: done!

include/LightGBM/c_api.h
```diff
@@ -734,6 +737,61 @@ LIGHTGBM_C_EXPORT int LGBM_BoosterPredictForCSR(BoosterHandle handle,
...
+ * \param indptr Pointer to output row headers or col headers to be deallocated
```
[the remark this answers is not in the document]
imatiach-msft: done!

include/LightGBM/c_api.h
```diff
@@ -734,6 +737,61 @@ LIGHTGBM_C_EXPORT int LGBM_BoosterPredictForCSR(BoosterHandle handle,
                                                 int64_t* out_len,
                                                 double* out_result);
 
+/*!
+ * \brief Make sparse prediction for a new dataset in CSR or CSC format.  Currently only used for feature contributions.
+ * \note
+ * The outputs are pre-allocated, as they can vary for each invocation, but the shape should be the same:
+ *   - for feature contributions, the shape of sparse matrix will be ``num_class * num_data * (num_feature + 1)``.
+ * The output indptr_type for the sparse matrix will be the same as the given input indptr_type.
+ * \param handle Handle of booster
```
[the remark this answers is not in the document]
imatiach-msft: great idea, done!

include/LightGBM/meta.h
```diff
@@ -11,6 +11,7 @@
 #include <memory>
 #include <utility>
 #include <vector>
+#include <unordered_map>
```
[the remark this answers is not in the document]
imatiach-msft: should limits import be later then too?
#include <limits>
#include <cstdint>
I made the change just for my <unordered_map> import

python-package/lightgbm/basic.py
```diff
@@ -5,6 +5,7 @@
 import copy
 import ctypes
 import os
+import sys
```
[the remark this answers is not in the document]
imatiach-msft: removed, done!

python-package/lightgbm/basic.py
```diff
@@ -520,8 +533,8 @@ def predict(self, data, num_iteration=-1,
 
         Returns
         -------
-        result : numpy array
-            Prediction result.
+        result : numpy array or scipy.sparse or list[scipy.sparse]
```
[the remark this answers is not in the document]
imatiach-msft: done!

python-package/lightgbm/basic.py
```diff
@@ -520,8 +533,8 @@ def predict(self, data, num_iteration=-1,
...
+            Prediction result, can be sparse for feature contributions (when pred_contrib=True).
```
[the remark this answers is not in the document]
imatiach-msft: done!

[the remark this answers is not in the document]
imatiach-msft: done!

python-package/lightgbm/basic.py
imatiach-msft: done!

python-package/lightgbm/basic.py
imatiach-msft: great question, why can't those native methods be changed to handle >MAX_INT32 rows for dense case?  For sparse case, only problem I see right now is the cast here for the passed in number of rows which can overflow:
int nrow = static_cast<int>(nindptr - 1);
and the code that follows this in PredictSparseCSR and CSC that stores row indexes.
However, I was just following the dense API style here.  I can change this but I don't quite understand why dense case has to do this batching and can't be modified to handle the full data.
Interestingly, dense case has this:
# __get_num_preds() cannot work with nrow > MAX_INT32, so calculate overall number of predictions piecemeal
but I don't call that method for sparse case.
I'm also not sure if this is really needed for computing local feature contributions, whereas I can understand this requirement more for predicting probabilities/values.

src/c_api.cpp
imatiach-msft: done!

src/c_api.cpp
imatiach-msft: done!

src/c_api.cpp
imatiach-msft: done!

src/c_api.cpp
imatiach-msft: done!

src/c_api.cpp
imatiach-msft: done!

src/c_api.cpp
imatiach-msft: done!

src/c_api.cpp
imatiach-msft: done!

src/c_api.cpp
imatiach-msft: done!

src/c_api.cpp
imatiach-msft: done!

src/c_api.cpp
imatiach-msft: done!

src/c_api.cpp
imatiach-msft: done!

src/c_api.cpp
imatiach-msft: done!

src/io/tree.cpp
imatiach-msft: done!

src/io/tree.cpp
imatiach-msft: interesting, I was wondering how to enter the whitespace here, done!

tests/python_package_test/test_engine.py
imatiach-msft: moved there, done!

tests/python_package_test/test_engine.py
imatiach-msft: done!

imatiach-msft:
@StrikerRUS thank you for the great comments, I've updated the PR

include/LightGBM/meta.h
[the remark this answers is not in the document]
StrikerRUS: It would be great! But it is up to you, as it is unrelated to this PR.

python-package/lightgbm/basic.py
StrikerRUS: I remember, this bunch of code with batching [was introduced](https://github.com/microsoft/LightGBM/pull/1615) exactly to implement [a feature request](https://github.com/microsoft/LightGBM/issues/1590) for supporting large data for feature contribution predictions.

And here is the reason behind `int32`:
> due to the efficiency concern, LightGBM only accept the int32 nrow.
For the nrow exceed the 32-bit ranges, please consider split it to multi data for prediction, and then concat back.
  https://github.com/microsoft/LightGBM/issues/1590#issuecomment-415066108

imatiach-msft: that's crazy, that someone would want so many local feature importance values.  Why is using int32 instead of int64 nrow an efficiency concern?  How big is the perf impact?

imatiach-msft: I think concatenating sparse matrices together into one from multiple is a bit more expensive than for dense data - eg you have to re-calculate the indptr

imatiach-msft: maybe just for the sparse API I added we can accept >Int32 MAX values for nrow

imatiach-msft: I'll update the PR with this fix

StrikerRUS: > Why is using int32 instead of int64 nrow an efficiency concern? How big is the perf impact?

ping @guolinke

include/LightGBM/meta.h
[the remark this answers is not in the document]
imatiach-msft: done!  fixed this in a separate commit

python-package/lightgbm/basic.py
imatiach-msft: @StrikerRUS I've fixed the code, actually I didn't need to change the external API as it was already accepting int64_t.  I think the fix is actually more trivial than I thought, and we should be able to accept CSR and CSC sparse matrices with more than int32 max rows.

imatiach-msft:
@StrikerRUS would you be able to take another look at the PR?  I think I've resolved all comments now.

python-package/lightgbm/basic.py
StrikerRUS: Great! Thank you very much!

StrikerRUS:
@guolinke Please update your review due `int -> int64` changes in https://github.com/microsoft/LightGBM/pull/3000/commits/b54a0327af4c2acf1f008e56e8659d0f914068db

imatiach-msft:
@StrikerRUS egads, the ramifications of supporting >intmax are much wider than I initially surmised, but hopefully latest fix should address those issues.  I also fixed the build warnings you mentioned, I don't seem to see those warnings on my linux (ubuntu) build.

imatiach-msft:
@StrikerRUS I also added a test to check the >int32 max rows case in the latest commit which I validated works.  It does add 23 seconds to CI though, so I'm fine with either disabling or removing it if you think it's not worth having the test.

imatiach-msft:
@StrikerRUS interesting, seems the test I wrote works on linux but fails on windows, will need to look into it more

StrikerRUS:
@imatiach-msft Thanks a lot for all your hard work!

But I think we should wait for the @guolinke's opinion. Maybe he is against using `int64 nrow` and then all your efforts are pointless.

imatiach-msft:
@StrikerRUS I think, if I get it to work, it will reveal the full scope of the changes required, so it is useful.  So far it has been pretty limited to just my code (sparse tree shap local importance values), and doesn't touch dense case or other types of predictions.  I think, if it will still be the case with a green build, then it will be fine to support int64 in the native side and not have to do this kind of batching.

imatiach-msft:
@StrikerRUS great news, I was able to reproduce the issue on windows.  The structure of the matrix was fine.  The problem was that this python code was called on a sparse matrix:

https://github.com/microsoft/LightGBM/blob/master/python-package/lightgbm/basic.py#L577

However in this case size returns not the number of rows but instead stored elements.
The code then subsequently tries to call reshape:

https://github.com/microsoft/LightGBM/blob/master/python-package/lightgbm/basic.py#L579

But surprisingly, there is a bug in reshape!  It does not support >int32max:
https://github.com/scipy/scipy/blob/master/scipy/sparse/sputils.py#L287

And we hit a known issue:
https://github.com/scipy/scipy/issues/12301

This cascade of errors causes the error to appear.
I've updated the PR with the fix.

But, this still did not explain the mystery of how this ever worked in linux.  I went and debugged this as well - the code path was exactly the same, but to my confusion this somehow worked:

(Pdb) np.prod((3000000, 1001), dtype=int)
3003000000

Why np.prod overflows on windows but works on linux with dtype=int is a mystery I didn't really want to venture in more, as the real fix should be further upstream anyway for sparse matrices.

imatiach-msft:
close reopen to re-trigger CI due to unrelated R-package test failure (from build "GitHub Actions / r-package (ubuntu-latest, gcc) (pull_request)"):

```
W: An error occurred during the signature verification. The repository is not updated and the previous index files will be used. GPG error: https://storage.googleapis.com/bazel-apt stable InRelease: The following signatures were invalid: EXPKEYSIG 3D5919B448457EE0 Bazel Developer (Bazel APT repository key) <bazel-dev@googlegroups.com>
```

imatiach-msft:
close reopen for CI:
```
URL        `http://www.doxygen.nl/index.html'

Name       `Doxygen'

Parent URL file:///home/travis/build/microsoft/LightGBM/docs/_build/html/README.html, line 171, col 109

Real URL   http://www.doxygen.nl/index.html

Check time 2.350 seconds

Result     Error: ConnectionError: ('Connection aborted.', BadStatusLine('No status line received - the server has closed the connection',))
```

imatiach-msft:
ping @StrikerRUS for review, I have a green build, but I'm not sure if I should remove the extra test I added (or maybe I can disable it just for CI somehow?) since it does take some time and it does use quite a bit of memory due to the large number of empty rows generated

tests/python_package_test/test_engine.py
```diff
@@ -941,6 +941,54 @@ def test_contribs(self):
...
+        }
+        train_features = np.random.rand(100, 1000)
+        train_targets = [0] * 50 + [1] * 50
+        lgb_train = lgb.Dataset(train_features, train_targets)
+        gbm = lgb.train(params, lgb_train, num_boost_round=5)
+        test_features = csr_matrix((3000000, 1000))
+        for i in range(0, 3000000, 500000):
+            for j in range(0, 1000, 100):
+                test_features[i, j] = random.random()
+        gbm.predict(test_features, pred_contrib=True)
```
StrikerRUS, line 990: Let's check the shape of result here.
```
self.assertTupleEqual(y_pred.shape, ...)
```

python-package/lightgbm/basic.py
```diff
@@ -574,7 +586,8 @@ def predict(self, data, num_iteration=-1,
             preds, nrow = self.__pred_for_csr(csr, num_iteration, predict_type)
         if pred_leaf:
             preds = preds.astype(np.int32)
-        if is_reshape and preds.size != nrow:
+        is_sparse = scipy.sparse.issparse(preds) or isinstance(preds, list)
```
StrikerRUS, line 589: Should `predict_type` be checked here? If I'm not mistaken, for all other predictions other than `C_API_PREDICT_CONTRIB` we still need to reshape result.

python-package/lightgbm/sklearn.py
```diff
@@ -668,7 +668,7 @@ def predict(self, X, raw_score=False, num_iteration=None,
             The predicted values.
         X_leaves : array-like of shape = [n_samples, n_trees] or shape = [n_samples, n_trees * n_classes]
             If ``pred_leaf=True``, the predicted leaf of every tree for each sample.
-        X_SHAP_values : array-like of shape = [n_samples, n_features + 1] or shape = [n_samples, (n_features + 1) * n_classes]
+        X_SHAP_values : array-like of shape = [n_samples, n_features + 1] or shape = [n_samples, (n_features + 1) * n_classes] or list of such objects
```
StrikerRUS, line 671:
```suggestion
        X_SHAP_values : array-like of shape = [n_samples, n_features + 1] or shape = [n_samples, (n_features + 1) * n_classes] or list with n_classes length of such objects
```

python-package/lightgbm/sklearn.py
```diff
@@ -886,7 +886,7 @@ def predict_proba(self, X, raw_score=False, num_iteration=None,
             The predicted probability for each class for each sample.
         X_leaves : array-like of shape = [n_samples, n_trees * n_classes]
             If ``pred_leaf=True``, the predicted leaf of every tree for each sample.
-        X_SHAP_values : array-like of shape = [n_samples, (n_features + 1) * n_classes]
+        X_SHAP_values : array-like of shape = [n_samples, (n_features + 1) * n_classes] or list of such objects
```
StrikerRUS, line 889:
```suggestion
        X_SHAP_values : array-like of shape = [n_samples, (n_features + 1) * n_classes] or list with n_classes length of such objects
```

python-package/lightgbm/basic.py
```diff
@@ -520,8 +532,8 @@ def predict(self, data, num_iteration=-1,
 
         Returns
         -------
-        result : numpy array
-            Prediction result.
+        result : numpy array, scipy.sparse or list of scipy.sparse
+            Prediction result, can be sparse for feature contributions (when ``pred_contrib=True``).
```
StrikerRUS, line 536:
```suggestion
            Prediction result.
            Can be sparse or a list of sparse objects (each element represents predictions for one class) for feature contributions (when ``pred_contrib=True``).
```

python-package/lightgbm/basic.py
```diff
@@ -2640,8 +2779,8 @@ def predict(self, data, num_iteration=None,
...
```
StrikerRUS, line 2783:
```suggestion
            Prediction result.
            Can be sparse or a list of sparse objects (each element represents predictions for one class) for feature contributions (when ``pred_contrib=True``).
```

tests/python_package_test/test_engine.py
imatiach-msft: done!

python-package/lightgbm/basic.py
imatiach-msft: no, it is not necessary, since all other predictions will be dense, only C_API_PREDICT_CONTRIB for sparse input can output a sparse output

python-package/lightgbm/sklearn.py
imatiach-msft: done!

python-package/lightgbm/sklearn.py
imatiach-msft: done!

python-package/lightgbm/basic.py
imatiach-msft: done!

python-package/lightgbm/basic.py
imatiach-msft: done!

imatiach-msft:
thanks @StrikerRUS for great comments, I've updated the PR

python-package/lightgbm/basic.py
StrikerRUS: Yeah, that's right! But only for now. Anyway, we should keep in mind that future contributions may require updating this condition.

imatiach-msft:
@StrikerRUS @guolinke I think all comments have been resolved so far, would you be able to take another look at this PR?  Thank you!

tests/python_package_test/test_engine.py
```diff
@@ -941,6 +941,59 @@ def test_contribs(self):
...
+        X, y = make_multilabel_classification(n_samples=n_samples,
+                                              sparse=True,
+                                              n_features=n_features,
+                                              n_classes=1,
+                                              n_labels=2)
+        y = y.flatten()
+        X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.1, random_state=42)
+        params = {
+            'objective': 'binary',
+            'metric': 'binary_logloss',
```
StrikerRUS, line 957:
```suggestion
            'objective': 'binary',
```

tests/python_package_test/test_engine.py
```diff
@@ -941,6 +941,59 @@ def test_contribs(self):
...
+        # validate the values are the same
+        np.testing.assert_allclose(contribs_csr.toarray(), contribs_dense)
+        self.assertLess(np.linalg.norm(gbm.predict(X_test, raw_score=True)
+                                       - np.sum(contribs_dense, axis=1)), 1e-4)
+        # validate using CSC matrix
+        X_train_csc = X_train.tocsc()
+        X_test_csc = X_test.tocsc()
+        lgb_train = lgb.Dataset(X_train_csc, y_train)
+        gbm = lgb.train(params, lgb_train, num_boost_round=20)
+        contribs_csc = gbm.predict(X_test_csc, pred_contrib=True)
```
StrikerRUS, line 974: Does it really make any sense to retrain with csc data?
```suggestion
        X_test_csc = X_test.tocsc()
        contribs_csc = gbm.predict(X_test_csc, pred_contrib=True)
```

tests/python_package_test/test_engine.py
```diff
@@ -941,6 +941,59 @@ def test_contribs(self):
...
+        X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.1, random_state=42)
+        params = {
+            'objective': 'binary',
+            'metric': 'binary_logloss',
+            'verbose': -1,
+        }
+        lgb_train = lgb.Dataset(X_train, y_train)
+        gbm = lgb.train(params, lgb_train, num_boost_round=20)
+        contribs_csr = gbm.predict(X_test, pred_contrib=True)
+        # convert data to dense and get back same contribs
```
StrikerRUS, line 963:
```suggestion
        contribs_csr = gbm.predict(X_test, pred_contrib=True)
        self.assertTrue(isspmatrix_csr(contribs_csr))
        # convert data to dense and get back same contribs
```

tests/python_package_test/test_engine.py
```diff
@@ -941,6 +941,59 @@ def test_contribs(self):
...
+        np.testing.assert_allclose(contribs_csr.toarray(), contribs_dense)
+        self.assertLess(np.linalg.norm(gbm.predict(X_test, raw_score=True)
+                                       - np.sum(contribs_dense, axis=1)), 1e-4)
+        # validate using CSC matrix
+        X_train_csc = X_train.tocsc()
+        X_test_csc = X_test.tocsc()
+        lgb_train = lgb.Dataset(X_train_csc, y_train)
+        gbm = lgb.train(params, lgb_train, num_boost_round=20)
+        contribs_csc = gbm.predict(X_test_csc, pred_contrib=True)
+        # validate the values are the same
```
StrikerRUS, line 975:
```suggestion
        contribs_csc = gbm.predict(X_test_csc, pred_contrib=True)
        self.assertTrue(isspmatrix_csc(contribs_csc))
        # validate the values are the same
```

tests/python_package_test/test_engine.py
```diff
@@ -11,7 +11,7 @@
 import numpy as np
 from scipy.sparse import csr_matrix
```
StrikerRUS, line 12:
```suggestion
from scipy.sparse import csr_matrix, isspmatrix_csr, isspmatrix_csc
```

tests/python_package_test/test_engine.py
```diff
@@ -941,6 +941,59 @@ def test_contribs(self):
...
+        params = {
+            'objective': 'binary'
+        }
+        train_features = np.random.rand(100, 1000)
+        train_targets = [0] * 50 + [1] * 50
+        lgb_train = lgb.Dataset(train_features, train_targets)
+        gbm = lgb.train(params, lgb_train, num_boost_round=5)
+        csr_input_shape = (3000000, 1000)
+        test_features = csr_matrix(csr_input_shape)
+        for i in range(0, 3000000, 500000):
```
StrikerRUS, line 989:
```suggestion
        for i in range(0, csr_input_shape[0], csr_input_shape[0] // 6):
```

tests/python_package_test/test_engine.py
imatiach-msft: done!

tests/python_package_test/test_engine.py
imatiach-msft: done, removed!

tests/python_package_test/test_engine.py
imatiach-msft: done!

tests/python_package_test/test_engine.py
imatiach-msft: done!

tests/python_package_test/test_engine.py
imatiach-msft: done!

tests/python_package_test/test_engine.py
imatiach-msft: done!

imatiach-msft:
"And please eliminate these compiler warnings..."
Unfortunately, I found for CSC case there was an issue based on warnings, but it would be difficult to fix (indices and out_indices are int32 but they would have to be int64 for CSC matrix in num rows>int32 max case - furthermore, a bunch of code for converting the CSC to rows in get_row_fun would have to be fixed as well to support int64 indices), in fact the python code just converts to csr for non-sparse output case so I did the same for sparse-output case.  I kept the logic inside the PredictSparseCSC to potentially be able to handle int64 in the future in case the code in get_row_fun is fixed (which requires quite a bit of work, maybe using templates) and the interface is slightly changed to also pass in the type for indices (which should be simple to do).

imatiach-msft:
@StrikerRUS would you be able to take another look at the PR?  Thanks!

StrikerRUS:
>  I kept the logic inside the PredictSparseCSC to potentially be able to handle int64 in the future in case the code in get_row_fun is fixed

Am I right that this PR breaks SHAP predictions for CSC matrices with rows >INT32_MAX? If so, I don't think we should do this. What about to check number of rows at Python side and use old bunch of code used for dense cases if #rows of CSC matrix > INT32_MAX?

imatiach-msft:
" this PR breaks SHAP predictions for CSC matrices with rows >INT32_MAX"
no, because we call the CSR code in >INT32_MAX case now with the most recent commit:

https://github.com/microsoft/LightGBM/pull/3000/files#diff-732a5a5220860efcac575e9e956bbaeaR842

The CSR case works for >INT32_MAX, the CSC code path has a problem that I can fix in another PR (much more involved fix)

imatiach-msft:
"check number of rows at Python side and use old bunch of code used for dense cases if #rows of CSC matrix > INT32_MAX"

@StrikerRUS  nope not necessary, the CSR case works for >INT32_MAX rows, and it's even tested in the python code :)

tests/python_package_test/test_engine.py
```diff
@@ -941,6 +941,57 @@ def test_contribs(self):
...
+        train_features = np.random.rand(100, 1000)
+        train_targets = [0] * 50 + [1] * 50
+        lgb_train = lgb.Dataset(train_features, train_targets)
+        gbm = lgb.train(params, lgb_train, num_boost_round=5)
+        csr_input_shape = (3000000, 1000)
+        test_features = csr_matrix(csr_input_shape)
+        for i in range(0, csr_input_shape[0], csr_input_shape[0] // 6):
+            for j in range(0, 1000, 100):
+                test_features[i, j] = random.random()
+        y_pred = gbm.predict(test_features, pred_contrib=True)
```
StrikerRUS, line 990: > no, because we call the CSR code in >INT32_MAX case now with the most recent commit:

Ah, sorry, completely missed these latest changes!

Maybe then it worth to add a test for CSC here as well, like `gbm.predict(test_features.tocsc(), pred_contrib=True)`? WDYT?

tests/python_package_test/test_engine.py
```diff
@@ -941,6 +941,57 @@ def test_contribs(self):
...
+
+    @unittest.skipIf(psutil.virtual_memory().available / 1024 / 1024 / 1024 < 3, 'not enough RAM')
+    def test_int32_max_sparse_contribs(self):
+        params = {
+            'objective': 'binary'
+        }
+        train_features = np.random.rand(100, 1000)
+        train_targets = [0] * 50 + [1] * 50
+        lgb_train = lgb.Dataset(train_features, train_targets)
+        gbm = lgb.train(params, lgb_train, num_boost_round=5)
```
StrikerRUS, line 984: To save CI time.
```suggestion
        gbm = lgb.train(params, lgb_train, num_boost_round=2)
```

imatiach-msft: done!

tests/python_package_test/test_engine.py
imatiach-msft: done!

imatiach-msft:
@StrikerRUS would you be able to take another look?  Thank you!  Also, I noticed some r-package tests were expected but didn't complete, it looks like some github configuration issue,

jameslamb:
> @imatiach-msft Thank you very much for your awesome work and your patience! LGTM!
> 
> > Also, I noticed some r-package tests were expected but didn't complete
> 
> Sorry about that! We have some problems with GitHub Actions config and should fix them to remove outdated tests from list. Refer to [#3164 (comment)](https://github.com/microsoft/LightGBM/pull/3164#issuecomment-644177785).

@guolinke could you make  @StrikerRUS or me an administrator on the repo just for a few days, until the GitHub Actions work stabilizes?

All PRs are currently blocked, the set of jobs will change again when we merge #3168 and then they will change again when I move https://github.com/jameslamb/LightGBM/pull/25 into this repo to fix #2960

guolinke:
@jameslamb sorry, I just remove the old ci checks

jameslamb:
> @jameslamb sorry, I just remove the old ci checks

thsbks @guolinke ! I just merged #3168 ... could you update the set of required checks again? Now there should be 4 more Windows checks on GitHub Actions

StrikerRUS:
Not merging this until @guolinke's re-review due to recent changes around int64.

imatiach-msft:
@guolinke would you be able to re-review this PR?  Thank you!

imatiach-msft:
ping @guolinke

guolinke:
@imatiach-msft sorry, I will review this today.

jameslamb:
@imatiach-msft can you please merge in the latest changes from `master`? Now that #3168 has been merged, there are additional checks this PR needs to pass. That will resolve these pending tasks:

image

Sorry for the inconvenience.

imatiach-msft:
Thank you for the review @guolinke .  @jameslamb I've updated the PR to latest master as suggested.  Thanks!

imatiach-msft:
@jameslamb  for some reason I still see "GitHub Actions / r-package (windows-latest, MINGW, R 3.6) (pull_request) " builds fail

imatiach-msft:
Close-reopen for CI (?)

jameslamb:
> @jameslamb for some reason I still see "GitHub Actions / r-package (windows-latest, MINGW, R 3.6) (pull_request) " builds fail

I will look once they rebuild. Sorry, they are still new and GitHub Actions is also still a bit rough to work with.

jameslamb:
@imatiach-msft I see the two Windows R4.0 builds are now failing. From the logs, I don't think it's a result of this PR. Will try to reproduce tonight and get it resolved quickly.

StrikerRUS:
@jameslamb FYI, there are a lot of PRs failing with these GitHub Actions jobs (R 4). `master` is also failing. Network issues again?..

jameslamb:
> @jameslamb FYI, there are a lot of PRs failing with these GitHub Actions jobs (R 4). `master` is also failing. Network issues again?..

gah! They don't look like networking issues, but I'm not sure. Investigating in #3191

jameslamb:
Ok @imatiach-msft now that we've merged #3193 , I think if you merge that into this branch the R CI jobs will be working

imatiach-msft:
@jameslamb done, thanks!

imatiach-msft:
close-reopen for CI, getting ".ci/test.sh: line 150: pytest: command not found"

jameslamb:
> close-reopen for CI, getting ".ci/test.sh: line 150: pytest: command not found"

which job had that failure? Could you share a link?

imatiach-msft:
great to see this merged, thank you for the great reviews!

PR #3002, opened 2020-04-17
[R-package] fix R examples and lgb.plot.interpretation()

jameslamb:
**NOTE:* Moving #2999 over here (based on a LightGBM branch) so we can test readthedocs builds. See https://github.com/microsoft/LightGBM/pull/2999#issuecomment-614648492

--------

See comments from @StrikerRUS 

* https://github.com/microsoft/LightGBM/pull/2989#issuecomment-614374151
* https://github.com/microsoft/LightGBM/pull/2989#issuecomment-614375963

A few of the R examples are broken right now. These weren't caught by tests because one is only a warning, one is a plot that is generated successfully but is incorrect, and one is an issue specific to our readthedocs builds.

I think the issue for `lgb.cv()` is #2715. In that issue we found that `data.table 1.11.4` and R 3.6.0 lead to a data.table error ending in `column or argument 2 is NULL`. I think this can be fixed by upgrading to the `data.table`  1.12.1 in `conf.py`.

To fix the `lgb.set.categorical()` example, I changed file names across several examples to be sure that no two examples write to the same file name.

image

This issue with `lgb.plot.interpretation()` is because I made a change and forgot to use `abs()`.

image

new plot:

image

I was not able to reproduce the `data.table` error in  https://github.com/microsoft/LightGBM/pull/2989#issuecomment-614374151. It

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -406,6 +406,15 @@ lightgbm_r/*
 lightgbm*.tar.gz
 lightgbm.Rcheck/
 
+# Files created by R examples and tests
+**/lgb-Dataset.data
+**/lgb-model.rds
+**/lgb.Dataset.data
+**/model.rds
+**/model.txt
+**/lgb-model.txt
+
+
 # Files from interactive R sessions
 .Rproj.user
 **/.Rhistory
--- a/.gitignore
+++ b/.gitignore
@@ -414,7 +414,6 @@ lightgbm.Rcheck/
 **/model.txt
 **/lgb-model.txt
 
-
 # Files from interactive R sessions
 .Rproj.user
 **/.Rhistory
--- a/R-package/DESCRIPTION
+++ b/R-package/DESCRIPTION
@@ -41,4 +41,4 @@ Imports:
     utils
 SystemRequirements:
     C++11
-RoxygenNote: 7.0.2
+RoxygenNote: 7.1.0
--- a/R-package/R/lgb.Booster.R
+++ b/R-package/R/lgb.Booster.R
@@ -845,7 +845,7 @@ lgb.load <- function(filename = NULL, model_str = NULL) {
 #'   , learning_rate = 1.0
 #'   , early_stopping_rounds = 5L
 #' )
-#' lgb.save(model, "model.txt")
+#' lgb.save(model, "lgb-model.txt")
 #' }
 #' @export
 lgb.save <- function(booster, filename, num_iteration = NULL) {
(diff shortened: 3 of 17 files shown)


StrikerRUS:
Building this branch right now.

StrikerRUS:
It is failing due to wrong R version of `r-data.table`.

StrikerRUS:
I believe we should bump version to `1.12` here
https://github.com/microsoft/LightGBM/blob/995a5974c988049e302fa9c5a26f72731c1b2a16/R-package/DESCRIPTION#L36

and correct the FAQ entry, because not only `lgb.dl` is affected
https://github.com/microsoft/LightGBM/blob/master/docs/FAQ.rst#3-error-in-datatabledatatableargument-2-is-null

jameslamb:
> I believe we should bump version to `1.12` here
> https://github.com/microsoft/LightGBM/blob/995a5974c988049e302fa9c5a26f72731c1b2a16/R-package/DESCRIPTION#L36
> 
> and correct the FAQ entry, because not only `lgb.dl` is affected
> https://github.com/microsoft/LightGBM/blob/master/docs/FAQ.rst#3-error-in-datatabledatatableargument-2-is-null

I disagree about updating the version floor. It seems that this is not specific to `lgb.dl()`, but I showed  in https://github.com/microsoft/LightGBM/issues/2715#issuecomment-589484656  that the incompatibility is more complicated than just "LightGBM doesn't work with 1.11.4". The `1.12` series [is only about 15 months  old](https://github.com/Rdatatable/data.table/releases/tag/1.12.0), so I'm not prepared to force people to upgrade to it yet.

I do agree on updating the FAQ entry...I'll do that directly in this PR.

jameslamb:
@StrikerRUS are you ok with the update I made to the FAQ? I think it came after your last approval.

StrikerRUS:
@jameslamb I'm OK with that you are against bumping the required version. But the FAQ entry should state that in case of that error version upgrade is required anyway for someone who encoutrered it, I believe.

Sure, please do it right in this PR.

jameslamb:
> @jameslamb I'm OK with that you are against bumping the required version. But the FAQ entry should state that in case of that error version upgrade is required anyway for someone who encoutrered it, I believe.
> 
> Sure, please do it right in this PR.

ha I forgot to push the commit!! I pushed to `origin` (my fork) a  few days ago...forgot this PR is on a LightGBM branch. My ask "are you ok with the update" probably was confusing, sorry!

Here it is: https://github.com/microsoft/LightGBM/pull/3002/commits/0eb02430c3425407f5a43b7aa38f98bc07211f5f

jameslamb:
Since this one is a user-facing fix  (fixing `lgb.plot.interpretation()`)  and a non-controversial one, I feel ok merging it without an approval from an R maintainer.

Thanks for the help  @StrikerRUS !

Issue #1910, opened 2018-12-17
[R] R CMD CHECK warning about registering routines


jameslamb:
## Overview

Running `R CMD CHECK` on the R package on Mac OS yields the following NOTE:

```
File ‘lightgbm/libs/lib_lightgbm.so’:
  Found no calls to: ‘R_registerRoutines’, ‘R_useDynamicSymbols’

It is good practice to register native routines and to disable symbol
search.

See ‘Writing portable packages’ in the ‘Writing R Extensions’ manual.
```

I attempted to address this in #1499 by adding [init.c](https://github.com/Microsoft/LightGBM/pull/1499/commits/94cc895565dd4ad936cee76b91bd4fc1c3b73b9f), but the issue persists.

I have noticed when running R CMD CHECK that at the compilation stage the `init.c` step happens outside of compiling `lightgbm.so` with `cmake`.

```
clang -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/usr/local/include   -fPIC  -Wall -g -O2  -c R_init.c -o R_init.o
clang -dynamiclib -Wl,-headerpad_max_install_names -undefined dynamic_lookup -single_module -multiply_defined suppress -L/Library/Frameworks/R.framework/Resources/lib -L/usr/local/lib -o lightgbm.so R_init.o -F/Library/Frameworks/R.framework/.. -framework R -Wl,-framework -Wl,CoreFoundation
```

I _believe_ (though have not had time to validate) that this one file is getting compiled and linked to `lightgbm.so` created by R's standard process for building projects with compiled code, and then the legit installation with `cmake` is both ignoring it AND overwriting `lightgbm.so`.

## How to run R CMD CHECK

To run R CMD CHECK on the package, clone the repo and install from source [per these directions](https://github.com/Microsoft/LightGBM/blob/master/R-package/README.md), then build the package and run R CMD CHECK.

On Mac OS, this looks like running the following from your local clone of `LightGBM`:

```
export CXX=/usr/local/bin/g++-8
export CC=/usr/local/bin/gcc-8

# build the package
Rscript build_r.R

# Check
LGB_VERSION=$(cat R-package/DESCRIPTION | grep ^Version | tr -d "Version: ")
R CMD CHECK lightgbm_${LGB_VERSION}.tar.gz --as-cran | tee check.log | cat
```

## How to close this issue

[...]

PR #3009, opened 2020-04-21
[R-package] [ci] added echoing of R CMD check logs to CI

jameslamb:
No that we use `Rscript build_r.R --skip--install`, the only build of the R package is from `R CMD check`. `R CMD check` suppresses output from the build and writes it to a file. Those logs are useful for us to check for compiler warnings and debug issues.

See https://github.com/microsoft/LightGBM/pull/2936/files#r411607175

### DIFF
--- a/.ci/test_r_package.sh
+++ b/.ci/test_r_package.sh
@@ -86,9 +86,17 @@ export _R_CHECK_FORCE_SUGGESTS_=0
 
 # fails tests if either ERRORs or WARNINGs are thrown by
 # R CMD CHECK
+check_succeeded="yes"
 R CMD check ${PKG_TARBALL} \
     --as-cran \
-|| exit -1
+|| check_succeeded="no"
+
+echo "R CMD check build logs:"
+cat ${BUILD_DIRECTORY}/lightgbm.Rcheck/00install.out
+
+if [[ $check_succeeded == "no" ]]; then
+    exit -1
+fi
 
 if grep -q -R "WARNING" "$LOG_FILE_NAME"; then
     echo "WARNINGS have been found by R CMD check!"


jameslamb:
> @jameslamb Thanks!
> Seems that it is really usefull!
> There are a lot of warnings like this one produced by `gcc-9` compiler (macOS job):
> 
> ```
> /Users/travis/build/microsoft/LightGBM/lightgbm.Rcheck/00_pkg_src/lightgbm/src/src/lightgbm_R.cpp:664:52: warning: cast between incompatible function types from 'LGBM_SER* (*)(LGBM_SE, LGBM_SE, LGBM_SE)' {aka 'LGBM_SER* (*)(LGBM_SER*, LGBM_SER*, LGBM_SER*)'} to 'DL_FUNC' {aka 'void* (*)()'} [-Wcast-function-type]
>   664 |   {"LGBM_GetLastError_R"              , (DL_FUNC) &LGBM_GetLastError_R              , 3},
>       |                                                    ^~~~~~~~~~~~~~~~~~~
> ```

Thanks! Yeah I _believe_ these can safely be ignored right now, but I also think I know how to fix them eventually.

These are all coming from our calls to `R_RegisterRoutines` (https://github.com/microsoft/LightGBM/blob/master/src/lightgbm_R.cpp#L709). I believe I set that up correctly per [CRAN's "Writing R Extensions"](https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Registering-native-routines). Recall that you'll get an `R CMD check` warning if you don't call `R_RegisterRoutines` (#1910). My guess is that we're getting these warnings because we're using our custom R-to-C interface instead of `SEXP` from `R.h`.

It would be good to eventually just use the R headers instead of our own, but I don't think that it will be a blocker to getting to CRAN, so I haven't prioritized it yet. I'll create an issue and put it in #2302

jameslamb:
Added #3016 to capture the work to suppress that compiler warning from https://github.com/microsoft/LightGBM/pull/3009#pullrequestreview-397342607

PR #3019, opened 2020-04-24
[ci] minor CI scripts update

StrikerRUS:


.ci/test_r_package.sh
```diff
@@ -34,11 +34,6 @@ fi
 
 # Installing R precompiled for Mac OS 10.11 or higher
 if [[ $OS_NAME == "macos" ]]; then
-
-    # temp fix for basictex
-    if [[ $AZURE == "true" ]]; then
-        brew update
-    fi
```
jameslamb, line 41: oh nice!

StrikerRUS: Yeah, it should give some speedup! 🙂

PR #3031, opened 2020-04-28
Fix loss computation in rank cross-entropy objective

sbruch:
An earlier refactoring led to an error in the loss computation. This change is intended to fix that so the loss is correctly computed following the paper (https://arxiv.org/abs/1911.09798)

### DIFF
--- a/src/objective/rank_objective.hpp
+++ b/src/objective/rank_objective.hpp
@@ -323,7 +323,7 @@ class RankXENDCG : public RankingObjective {
     double sum_l1 = 0.0f;
     for (data_size_t i = 0; i < cnt; ++i) {
       l1s[i] = -l1s[i] / sum_labels + rho[i];
-      sum_l1 += l1s[i];
+      sum_l1 += l1s[i] / (1. - rho[i]);
     }
     if (cnt <= 1) {
       // when cnt <= 1, the l2 and l3 are zeros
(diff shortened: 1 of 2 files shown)


src/objective/rank_objective.hpp
```diff
@@ -336,13 +336,13 @@ class RankXENDCG : public RankingObjective {
       std::vector<double> l2s(cnt, 0.0);
       double sum_l2 = 0.0;
       for (data_size_t i = 0; i < cnt; ++i) {
-        l2s[i] = (sum_l1 - l1s[i]) / (1 - rho[i]);
```
stanpcf, line 339: Hi @sbruch , thanks for your great work for LTR. 
But I have some confusion when I the paper with the code. 
I don't know why use l2 and l3 to calculate the gradient. paper only use l1 to compute gradient? 
I'm looking forward to hearing from you. thank you.

sbruch: Hi. You can find the details of what becomes the gradient and the hessian in Section 4.4. In particular the very end of the section, in fact the very last equation, is relevant to this discussion: What ends up in this implementation as the "gradient" is the numerator in the Newton step (Hessian^{-1} * gradient}, and the "hessian" in this implementation is the denominator.

This was my way of piggybacking off the current framework to ultimately compute a Newton step.

By the way, regardless of the technical details, the code as of today does not match the logic in my original implementation in https://github.com/microsoft/LightGBM/pull/2620. This PR is meant to address that particular issue.

stanpcf: many thanks for your reply

sbruch:
Thank you all for the quick review!

PR #2843, opened 2020-02-29
add more bots 

guolinke:
Release Drafter: https://probot.github.io/apps/release-drafter/
Work In Progress: https://probot.github.io/apps/wip/
Reminders: https://probot.github.io/apps/reminders/
No Response: https://probot.github.io/apps/no-response/

PR #3035, opened 2020-04-29
[R-package] Remove lgbdl from docs until fix

Laurae2:
Since the method of installing LightGBM for R changed significantly, using [Laurae2/lgbdl](https://github.com/Laurae2/lgbdl) package is now broken.

This PR removes lgbdl temporarily from the documentation until a new working replacement package is made (to avoid confusion for end users who attempt using it and see it does not work).

The replacement package is likely to be named differently as there are currently scripts relying on the package pointing to old commits for compatibility.

@StrikerRUS `doc` is this the right label for this PR?

### DIFF
--- a/R-package/README.md
+++ b/R-package/README.md
@@ -16,7 +16,7 @@ Installation
 
 You need to install git and [CMake](https://cmake.org/) first.
 
-Note: 32-bit R/Rtools is not supported.
+Note: 32-bit (i386) R/Rtools is currently not supported.
 
 #### Windows Preparation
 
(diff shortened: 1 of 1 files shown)


StrikerRUS:
@Laurae2 

> `doc` is this the right label for this PR?

Yeah, I think so! Refer to https://github.com/microsoft/LightGBM/pull/2843#issuecomment-593025035.

R-package/README.md
```diff
@@ -16,7 +16,7 @@ Installation
 
 You need to install git and [CMake](https://cmake.org/) first.
 
-Note: 32-bit R/Rtools is not supported.
+Note: 32-bit (i386) R/Rtools is currently not supported.
```
jameslamb, line 19: good change!

PR #3048, opened 2020-05-05
Add public accessors to tree for decision_type and threshold

VerTiGoEtrex:
Two very short changes to enable 3rd-party libraries to parse LightGBM trees more safely.

- Expose public accessors to `threshold_` and `decision_type_`. Libraries that want to read LightGBM trees need access to these fields if they want to reuse `tree.h` for parsing `LightGBM_Model.txt` files rather than parsing it themselves.
- Convert the `#define`s to `constexpr`s. As of today, including `tree.h` pollutes C and C++ codebases with defines they may accidentially overwrite or forget to `#undef`. In general, for global constants, we should prefer namespaces.

### DIFF
--- a/include/LightGBM/tree.h
+++ b/include/LightGBM/tree.h
@@ -16,8 +16,8 @@
 
 namespace LightGBM {
 
-#define kCategoricalMask (1)
-#define kDefaultLeftMask (2)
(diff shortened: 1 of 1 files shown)


VerTiGoEtrex:
Need to make some additional changes here, it turns out.

PR #3055, opened 2020-05-07
typo fix

guanqun:

PR #3056, opened 2020-05-07
[ci] fix flaky test

StrikerRUS:
I noticed that Linux MPI builds at Azure Pipelines have started to fail (refer for example to #3055). Seems that dropping outdated Python 2.7 version from that build fixes the issue.

On Linux we have tests with Python 2.7 here, so this version is still covered by our CIs.
https://github.com/microsoft/LightGBM/blob/18c706dc840c72341a2a85e05dfb5fad6366902a/.travis.yml#L18

Refer to https://github.com/microsoft/LightGBM/pull/1964#issuecomment-458674944.

### DIFF
--- a/.vsts-ci.yml
+++ b/.vsts-ci.yml
@@ -37,7 +37,6 @@ jobs:
       mpi_source:
         TASK: mpi
         METHOD: source
-        PYTHON_VERSION: 2.7
       gpu_source:
         TASK: gpu
         METHOD: source


StrikerRUS:
Thanks! Yeah, our Python wrapper doesn't require any special features from latest versions and thanks to Python internal compatibility our old code works fine with the latest version.

Proportion breakdown says that we can safely drop Python 2 support. But in absolute values it is still several thousands of downloads per day. And those thousands are our users. I think we can try to support it for some months or before next minor release as it doesn't cost us anything.

https://pypistats.org/packages/lightgbm

Issue #2353, opened 2019-08-24
CI via GitHub actions


guolinke:
It seems we can use github actions for the CI tests:
https://help.github.com/en/articles/about-github-actions

> Workflows run in Linux, macOS, Windows, and containers on GitHub-hosted servers. You can create workflows using actions defined in your repository, open source actions in a public repository on GitHub, or a published Docker container image.

> You can execute up to 20 workflows concurrently per repository.
> You can execute up to 1000 API requests in an hour across all actions within a repository.
> Each job in a workflow can run for up to 6 hours of execution time.
> You can run up to 20 jobs concurrently per repository across all workflows.

From these descriptions, it seems better than other CIs we are using.

@StrikerRUS

hayesall:
Badges in the README can be updated too, here's an example:

<a href="https://github.com/actions/toolkit"><img alt="GitHub Actions status" src="https://github.com/actions/toolkit/workflows/Main%20workflow/badge.svg"></a>

The syntax should look like this (courtesy of Reddit user /u/peaceiris in [this thread](https://www.reddit.com/r/github/comments/csehoc/github_actions_official_status_badges/))

```markdown
https://github.com/{github_id}/{repository}/workflows/{workflow_name}/badge.svg
```

StrikerRUS:
> It seems we can use github actions for the CI tests:

I think we can start from tests for R-package. @jameslamb created #2335 for that recently. R tests will require new script files for stuff installation and test itself, so it's good way to not adjust our existing scripts, but develop new ones from the scratch and learn new CI platform.

jameslamb:
@StrikerRUS yes that's a good idea, I've been wanting to learn GitHub actions for CI.

I'll try that for #2335 (just assigned to myself).

hayesall:
I adapted the Travis matrix to a GitHub Actions matrix. It's nowhere close to being ready for a pull request, but there aren't many examples yet so I figured I'd share what I've found.

- [hayesall/LightGBM: .github/workflows/ccpp.yml](https://github.com/hayesall/LightGBM/blob/ciworkflow/.github/workflows/ccpp.yml)
- [hayesall/LightGBM: Actions Example build](https://github.com/hayesall/LightGBM/runs/203283214)

The [workflow documentation](https://help.github.com/en/categories/automating-your-workflow-with-github-actions) is a good place to start, but it's still being expanded.

StrikerRUS:
[...]

PR #3065, opened 2020-05-10
[R-package] Add support for R 4.0 (fixes #3064, fixes #3024)

### LINKED ISSUE #3024: [R-package] [ci] Add tests on R 4.0 compatibility
Currently, the R package is only tested against 3.6.x in continuous integration. This is controlled by variables  `R_MAC_VERSION`, `R_LINUX_VERSIONS`, and (soon, as of #2936 ) `R_WINDOW_VERSION` in CI config files:

* `.vsts-ci.yml` (Azure DevOps)
* `.travis.yml` (Travis CI)
* `.appveyor.yml` (AppVeyor)

To close this issue, add tests to those builds that use R 4.0. We will have to decide whether the extra testing time is worth it and ALL tests should be duplicated (e.g. one R4.0.x build for every current R 3.6.x build) or whether we should for now just have one R 4.0 build as a smoke test.

For now, do not change the container used in our Azure DevOps builds (https://github.com/guolinke/lightgbm-ci-docker).

### LINKED ISSUE #3064: [R-package] Installation with R4.0.0 on Windows is broken
Is it possible to install LightGBM with R 4.0.0? So far, I failed with different configs. This is my latest try. On my other Windows machine, I had no troubles today to build under R 3.6.3.

**With a hack, it is possible, see my response**

R 4.0.0 has been released with a new RTools version with different install paths etc.

## Config
- Operating System: Windows 10 - version 10.0.18362 Build 18362
- CPU: Intel(R) Core(TM) i7-7700 CPU @ 3.60GHz
- C++ compiler version: Visual Studio 16 2019, MSVC 19.25.28614.0
- CMake version: 3.17.2
- **R version: 4.0.0**
- **RTools40**

PATH
image

## Error
PS C:\Users\Michael\LightGBM> rscript build_r.R
* checking for file 'lightgbm_r/DESCRIPTION' ... OK
* preparing 'lightgbm':
* checking DESCRIPTION meta-information ... OK
* cleaning src
* checking for LF line-endings in source and make files and shell scripts
WARNING: directory 'lightgbm/src/compute/test' is empty
* looking to see if a 'data/datalist' file should be added
* building 'lightgbm_2.3.2.tar.gz'

* installing to library 'C:/Users/Michael/Documents/R/win-library/4.0'
* installing *source* package 'lightgbm' ...
** using staged installation
** libs
installing via 'install.libs.R' to C:/Users/Michael/Documents/R/win-library/4.0/00LOCK-lightgbm/00new/lightgbm
[1] "Trying to build with: 'Visual Studio 16 2019'"
-- Selecting Windows SDK version 10.0.15063.0 to target Windows 10.0.18362.
-- The C compiler identification is MSVC 19.25.28614.0
-- The CXX compiler identification is MSVC 19.25.28614.0
-- Check for working C compiler: C:/Program Files (x86)/Microsoft Visual Studio/2019/Community/VC/Tools/MSVC/14.25.28610/bin/Hostx64/x64/cl.exe
-- Check for working C compiler: C:/Program Files (x86)/Microsoft Visual Studio/2019/Community/VC/Tools/MSVC/14.25.28610/bin/Hostx64/x64/cl.exe - works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: C:/Program Files (x86)/Microsoft Visual Studio/2019/Community/VC/Tools/MSVC/14.25.28610/bin/Hostx64/x64/cl.exe
-- Check for working CXX compiler: C:/Program Files (x86)/Microsoft Visual Studio/2019/Community/VC/Tools/MSVC/14.25.28610/bin/Hostx64/x64/cl.exe - works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
R version passed into FindLibR.cmake: 4.0.0
Creating R.lib and R.def
CMake Error at cmake/modules/FindLibR.cmake:48 (message):
  Either gendef.exe or dlltool.exe not found!

  Do you have Rtools installed with its MinGW's bin/ in PATH?
  cmake/modules/FindLibR.cmake:204 (create_rlib_for_msvc)
  CMakeLists.txt:76 (find_package)

### Comments
dlltool.exe is here:
image

In that folder, `gendef.exe` is not available though. On my other machine with old R and old RTools, "gendef.exe" is present. Hmm.

jameslamb:
Experimental fix for https://github.com/microsoft/LightGBM/issues/3064. Opening this PR to show the way I'm thinking about adding support for R 4.0. I will remove the `[WIP]` and request reviewers when it's ready for review, but any comments before then are welcomed if you have time!

## R4.0 breaking changes that affect LightGBM

* `gendef.exe` removed from Rtools 😭 
* `mingw32-make.exe` is no longer bundled with `Rtools` 😭 
* `Rtools` paths removed underscores, so `mingw_64/` becomes `mingw64/` 😂 

## Short description of the fix

**gendef.exe fix**

Building the `LightGBM` R package on Windows with Visual Studio compilers requires `gendef.exe`. That was bundled in previous versions of `RTools` but isn't part of `Rtools` 4.0.

This PR replaces it with an R script that uses `objdump.exe`, software bundled in old and new distributions of `Rtools`.

**mingw32-make.exe fix**

Using `Rtools/usr/bin/make.exe` instead of `Rtools/mingw_64/bin/mingw32-make.exe` if using R version 4.0 or greater.

**mingw path change fix**

It  includes updates to docs and CI scripts for the `mingw_64/` --> `mingw64/` change.

## How you can test this

In a Windows environment with R 4.0, Rtools 4.0, CMake, and Visual Studio, run the following:

```shell
git clone https://github.com/jameslamb/LightGBM.git
cd LightGBM
git fetch origin fix/r-4.0
git checkout fix/r-4.0

# install LightGBM
Rscript build_r.R

# test that it worked
cd R-package/tests
Rscript testthat.R
```

## Long Description

Here's how building the R package with Visual Studio works today:

1. The LightGBM R package has to be linked to `R.dll` to use R-provided C/C++ functions such as `Rprintf` for printing.
2. Linking with Visual Studio compilers requires a `.lib` file (https://docs.microsoft.com/en-us/cpp/build/reference/link-input-files?redirectedfrom=MSDN&view=vs-2019)
3. To create a `.lib` file from `R.dll` on Windows, we create a definition file (`.def`) and then use that to make a library file (`.lib`).
    - `R.dll` ships with R
    - A file `R.def` is created from `R.dll`,  using `gendef.exe`
    - A file `R.lib` is created from `R.def` using `dlltool.exe`

The issue: `gendefe.exe` is not part of [RTools](https://cran.r-project.org/bin/windows/Rtools/) 4.0.

My solution in this PR is to use `objdump.exe` + an R script to generate `R.def`. `objdump` creates output intended for humans not machines...it uses indentation and special chracters to make the file visually appealing. So the R script is used to run `objdump`, but more importantly to read in the text it produces and do some simple cleaning to get it into a machine-friendly format that can be re-written as a `.def` file.

From the "How to create a def file from a dll" section of [this amazing MinGW documentation](https://www.cs.colorado.edu/~main/cs1300/doc/mingwfaq.html).

> If the previous options don't work, you can still try to create a def file using the output from the objdump program (from the mingw distribution).
Here's an example.
objdump -p file.dll > dll.fil

> Search for [Ordinal/Name Pointer] Table in dll.fil and use the list of functions following it to create your def file.

## References

A lot of the content for this PR was totally new to me. Here are some links I found very helpful.

* `.def` file: text file that contains the names of objections exported from a DLL
    - ["Exporting from a DLL using def files"](https://docs.microsoft.com/en-us/cpp/build/exporting-from-a-dll-using-def-files?view=vs-2019)
* `.lib` file: a library file, basically a specific format that describe DLL so a linker understands how to link to it
    - [StackOverflow answer explaining .lib vs .dll](https://stackoverflow.com/questions/913691/dll-and-lib-files-what-and-why/913744)
    - [".lib file as linker input"](https://docs.microsoft.com/en-us/cpp/build/reference/dot-lib-files-as-linker-input?view=vs-2019)
* R uses the MinGW toolchain for R
    - ["MinGW FAQ"](https://www.cs.colorado.edu/~main/cs1300/doc/mingwfaq.html) is an AMAZING resource on how to work with libraries using MinGW tools
* RTools
    - [official CRAN docs on RTools 4.0](https://cran.r-project.org/bin/windows/Rtools/)
    - [more detailed docs on what is new in RTools 4.0 (work in progress)](https://github.com/r-windows/docs)

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -14,10 +14,6 @@ branches:
 environment:
   matrix:
     - TASK: r-package
-      R_VERSION: 3.6
-      COMPILER: MINGW
-      TOOLCHAIN: MINGW
-    - TASK: r-package
       R_VERSION: 4.0
       COMPILER: MINGW
       TOOLCHAIN: MSYS
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -15,13 +15,13 @@ environment:
   matrix:
     - COMPILER: MINGW
       TASK: r-package
-      R_WINDOWS_VERSION: 3.6.3
+      R_VERSION: 3.6
     - COMPILER: MINGW
       TASK: r-package
-      R_WINDOWS_VERSION: 4.0.0
+      R_VERSION: 4.0
     - COMPILER: MSVC
       TASK: r-package
-      R_WINDOWS_VERSION: 4.0.0
+      R_VERSION: 4.0
       APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
     # - COMPILER: MSVC
     #   TASK: python
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -16,9 +16,11 @@ environment:
     - COMPILER: MINGW
       TASK: r-package
       R_VERSION: 3.6
+      TOOLCHAIN: MINGW
     - COMPILER: MINGW
       TASK: r-package
       R_VERSION: 4.0
+      TOOLCHAIN: MSYS
     - COMPILER: MSVC
       TASK: r-package
       R_VERSION: 4.0
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -13,20 +13,20 @@ branches:
 
 environment:
   matrix:
-    # - COMPILER: MINGW
-    #   TASK: r-package
-    #   R_WINDOWS_VERSION: 3.6.3
-    # - COMPILER: MINGW
-    #   TASK: r-package
-    #   R_WINDOWS_VERSION: 4.0.0
-    # - COMPILER: MSVC
-    #   TASK: r-package
-    #   R_WINDOWS_VERSION: 4.0.0
-    #   APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
-    - COMPILER: MSVC
-      TASK: python
     - COMPILER: MINGW
-      TASK: python
+      TASK: r-package
+      R_WINDOWS_VERSION: 3.6.3
+    - COMPILER: MINGW
+      TASK: r-package
+      R_WINDOWS_VERSION: 4.0.0
+    - COMPILER: MSVC
+      TASK: r-package
+      R_WINDOWS_VERSION: 4.0.0
+      APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
+    # - COMPILER: MSVC
+    #   TASK: python
+    # - COMPILER: MINGW
+    #   TASK: python
 
 clone_depth: 5
 
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -13,13 +13,14 @@ branches:
 
 environment:
   matrix:
-    - TASK: r-package
+    - COMPILER: MINGW
+      TASK: r-package
+      R_VERSION: 3.6
+      TOOLCHAIN: MINGW
+    - COMPILER: MSVC 
+      TASK: r-package
       R_VERSION: 4.0
-      COMPILER: MINGW
-      TOOLCHAIN: MSYS
-    - TASK: r-package
-      R_VERSION: 4.0
-      COMPILER: MSVC
+      TOOLCHAIN: MSVC
       APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
     - COMPILER: MSVC
       TASK: python
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -13,18 +13,18 @@ branches:
 
 environment:
   matrix:
-    - COMPILER: MINGW
-      TASK: r-package
-      R_VERSION: 3.6
-      TOOLCHAIN: MINGW
+    # - COMPILER: MINGW
+    #   TASK: r-package
+    #   R_VERSION: 3.6
+    #   TOOLCHAIN: MINGW
     - COMPILER: MINGW
       TASK: r-package
       R_VERSION: 4.0
       TOOLCHAIN: MSYS
-    - COMPILER: MSVC
-      TASK: r-package
-      R_VERSION: 4.0
-      APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
+    # - COMPILER: MSVC
+    #   TASK: r-package
+    #   R_VERSION: 4.0
+    #   APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
     # - COMPILER: MSVC
     #   TASK: python
     # - COMPILER: MINGW
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -13,22 +13,22 @@ branches:
 
 environment:
   matrix:
-    # - COMPILER: MINGW
-    #   TASK: r-package
-    #   R_VERSION: 3.6
-    #   TOOLCHAIN: MINGW
-    - COMPILER: MINGW
-      TASK: r-package
+    - TASK: r-package
+      R_VERSION: 3.6
+      COMPILER: MINGW
+      TOOLCHAIN: MINGW
+    - TASK: r-package
       R_VERSION: 4.0
+      COMPILER: MINGW
       TOOLCHAIN: MSYS
-    # - COMPILER: MSVC
-    #   TASK: r-package
-    #   R_VERSION: 4.0
-    #   APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
-    # - COMPILER: MSVC
-    #   TASK: python
-    # - COMPILER: MINGW
-    #   TASK: python
+    - TASK: r-package
+      R_VERSION: 4.0
+      COMPILER: MSVC
+      APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
+    - COMPILER: MSVC
+      TASK: python
+    - COMPILER: MINGW
+      TASK: python
 
 clone_depth: 5
 
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -15,6 +15,10 @@ environment:
   matrix:
     - COMPILER: MINGW
       TASK: r-package
+      R_WINDOWS_VERSION: 3.6.3
+    - COMPILER: MINGW
+      TASK: r-package
+      R_WINDOWS_VERSION: 4.0.0
     - COMPILER: MSVC
       TASK: r-package
       APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -13,15 +13,16 @@ branches:
 
 environment:
   matrix:
-    - COMPILER: MINGW
-      TASK: r-package
-      R_WINDOWS_VERSION: 3.6.3
-    - COMPILER: MINGW
-      TASK: r-package
-      R_WINDOWS_VERSION: 4.0.0
-    - COMPILER: MSVC
-      TASK: r-package
-      APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
+    # - COMPILER: MINGW
+    #   TASK: r-package
+    #   R_WINDOWS_VERSION: 3.6.3
+    # - COMPILER: MINGW
+    #   TASK: r-package
+    #   R_WINDOWS_VERSION: 4.0.0
+    # - COMPILER: MSVC
+    #   TASK: r-package
+    #   R_WINDOWS_VERSION: 4.0.0
+    #   APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
     - COMPILER: MSVC
       TASK: python
     - COMPILER: MINGW
--- a/.ci/test_r_package_windows.ps1
+++ b/.ci/test_r_package_windows.ps1
@@ -71,7 +71,7 @@ if ($env:TOOLCHAIN -eq "MINGW") {
 } elseif ($env:TOOLCHAIN -eq "MSYS") {
   Write-Output "Telling R to use MSYS"
   $install_libs = "$env:BUILD_SOURCESDIRECTORY/R-package/src/install.libs.R"
-  ((Get-Content -path $install_libs -Raw) -replace 'use_msys <- FALSE','use_msys <- TRUE') | Set-Content -Path $install_libs
+  ((Get-Content -path $install_libs -Raw) -replace 'use_msys2 <- FALSE','use_msys2 <- TRUE') | Set-Content -Path $install_libs
 }
 
 # download R and RTools
--- a/.ci/test_r_package_windows.ps1
+++ b/.ci/test_r_package_windows.ps1
@@ -19,12 +19,16 @@ $env:CRAN_MIRROR = "https://cloud.r-project.org/"
 $env:CTAN_MIRROR = "https://ctan.math.illinois.edu/systems/win32/miktex/tm/packages/"
 
 # some paths and file names are different on R4.0
-if ($env:R_WINDOWS_VERSION -eq "4.0.0") {
+$env:R_MAJOR_VERSION=$env:R_VERSION.split('.')[0]
+if ($env:R_MAJOR_VERSION -eq "3") {
+  $env:RTOOLS_MINGW_BIN = "$env:R_LIB_PATH/Rtools/mingw_64/bin"
+  $env:RTOOLS_EXE_FILE = "Rtools35.exe"
+} elseif ($env:R_MAJOR_VERSION -eq "4") {
   $env:RTOOLS_MINGW_BIN = "$env:R_LIB_PATH/Rtools/mingw64/bin"
   $env:RTOOLS_EXE_FILE = "rtools40-x86_64.exe"
 } else {
-  $env:RTOOLS_MINGW_BIN = "$env:R_LIB_PATH/Rtools/mingw_64/bin"
-  $env:RTOOLS_EXE_FILE = "Rtools35.exe"
+  Write-Output "Unrecognized R version: $env:R_VERSION"
+  Check-Output $false
 }
 
 if ($env:COMPILER -eq "MINGW") {
@@ -44,13 +48,8 @@ if ($env:COMPILER -eq "MINGW") {
 
 # download R and RTools
 Write-Output "Downloading R and Rtools"
-Download-File-With-Retries -url "https://cloud.r-project.org/bin/windows/base/old/$env:R_WINDOWS_VERSION/R-$env:R_WINDOWS_VERSION-win.exe" -destfile "R-win.exe"
-
-if ($env:R_WINDOWS_VERSION -eq "4.0.0") {
-  Download-File-With-Retries -url "https://cran.r-project.org/bin/windows/Rtools/$env:RTOOLS_EXE_FILE" -destfile "Rtools.exe"
-} else {
-  Download-File-With-Retries -url "https://cloud.r-project.org/bin/windows/Rtools/$env:RTOOLS_EXE_FILE" -destfile "Rtools.exe"
-}
+Download-File-With-Retries -url "https://cloud.r-project.org/bin/windows/base/old/$env:R_VERSION/R-$env:R_VERSION-win.exe" -destfile "R-win.exe"
+Download-File-With-Retries -url "https://cloud.r-project.org/bin/windows/Rtools/$env:RTOOLS_EXE_FILE" -destfile "Rtools.exe"
 
 # Install R
 Write-Output "Installing R"
--- a/.ci/test_r_package_windows.ps1
+++ b/.ci/test_r_package_windows.ps1
@@ -53,7 +53,7 @@ if ($env:TOOLCHAIN -eq "MINGW") {
 if ($env:TOOLCHAIN -eq "MSYS") {
   Write-Output "Telling R to use MSYS"
   $install_libs = "$env:BUILD_SOURCESDIRECTORY/R-package/src/install.libs.R"
-  ((Get-Content -path $install_libs -Raw) -replace 'use_msys <- FALSE','use_mingw <- TRUE') | Set-Content -Path $install_libs
+  ((Get-Content -path $install_libs -Raw) -replace 'use_msys <- FALSE','use_msys <- TRUE') | Set-Content -Path $install_libs
 }
 
 # download R and RTools
--- a/.ci/test_r_package_windows.ps1
+++ b/.ci/test_r_package_windows.ps1
@@ -92,9 +92,6 @@ Write-Output "Installing dependencies"
 $packages = "c('data.table', 'jsonlite', 'Matrix', 'processx', 'R6', 'testthat'), dependencies = c('Imports', 'Depends', 'LinkingTo')"
 Rscript --vanilla -e "options(install.packages.check.source = 'no'); install.packages($packages, repos = '$env:CRAN_MIRROR', type = 'binary', lib = '$env:R_LIB_PATH')" ; Check-Output $?
 
-Rscript build_r.R
-Check-Output $false
-
 Write-Output "Building R package"
 
 # R CMD check is not used for MSVC builds
@@ -144,6 +141,16 @@ $checks = Select-String -Path "${INSTALL_LOG_FILE_NAME}" -Pattern "Check for wor
 if ($checks.Matches.length -eq 0) {
   Write-Output "The wrong compiler was used. Check the build logs."
   Check-Output $False
+}
+
+# Checking that we got the right toolchain for MinGW. If using MinGW, both
+# MinGW and MSYS toolchains are supported
+if ($env:COMPILER -eq "MINGW") {
+  $checks = Select-String -Path "${INSTALL_LOG_FILE_NAME}" -Pattern "Trying to build with.*$env:TOOLCHAIN"
+  if ($checks.Matches.length -eq 0) {
+    Write-Output "The wrong toolchain was used. Check the build logs."
+    Check-Output $False
+  }
 }
 
 if ($env:COMPILER -eq "MSVC") {
--- a/.ci/test_r_package_windows.ps1
+++ b/.ci/test_r_package_windows.ps1
@@ -18,7 +18,10 @@ $env:PATH = "$env:R_LIB_PATH/Rtools/bin;" + "$env:R_LIB_PATH/Rtools/usr/bin;" + 
 $env:CRAN_MIRROR = "https://cloud.r-project.org/"
 $env:CTAN_MIRROR = "https://ctan.math.illinois.edu/systems/win32/miktex/tm/packages/"
 
-# some paths and file names are different on R4.0
+# Get details needed for installing R components
+#
+# NOTES:
+#    * some paths and file names are different on R4.0
 $env:R_MAJOR_VERSION=$env:R_VERSION.split('.')[0]
 if ($env:R_MAJOR_VERSION -eq "3") {
   $env:RTOOLS_MINGW_BIN = "$env:R_LIB_PATH/Rtools/mingw_64/bin"
--- a/.ci/test_r_package_windows.ps1
+++ b/.ci/test_r_package_windows.ps1
@@ -75,7 +75,8 @@ if ($env:TOOLCHAIN -eq "MINGW") {
 } elseif ($env:TOOLCHAIN -eq "MSVC") {
   # no customization for MSVC
 } else {
-  Write-Output "[ERROR] Unrecognized coompiler: $env:TOOLCHAIN"
+  Write-Output "[ERROR] Unrecognized compiler: $env:TOOLCHAIN"
+  Check-Output $false
 }
 
 # download R and RTools
--- a/.ci/test_r_package_windows.ps1
+++ b/.ci/test_r_package_windows.ps1
@@ -26,9 +26,11 @@ $env:R_MAJOR_VERSION=$env:R_VERSION.split('.')[0]
 if ($env:R_MAJOR_VERSION -eq "3") {
   $env:RTOOLS_MINGW_BIN = "$env:R_LIB_PATH/Rtools/mingw_64/bin"
   $env:RTOOLS_EXE_FILE = "Rtools35.exe"
+  $env:R_WINDOWS_VERSION = "3.6.3"
 } elseif ($env:R_MAJOR_VERSION -eq "4") {
   $env:RTOOLS_MINGW_BIN = "$env:R_LIB_PATH/Rtools/mingw64/bin"
   $env:RTOOLS_EXE_FILE = "rtools40-x86_64.exe"
+  $env:R_WINDOWS_VERSION = "4.0.0"
 } else {
   Write-Output "Unrecognized R version: $env:R_VERSION"
   Check-Output $false
@@ -51,7 +53,7 @@ if ($env:COMPILER -eq "MINGW") {
 
 # download R and RTools
 Write-Output "Downloading R and Rtools"
-Download-File-With-Retries -url "https://cloud.r-project.org/bin/windows/base/old/$env:R_VERSION/R-$env:R_VERSION-win.exe" -destfile "R-win.exe"
+Download-File-With-Retries -url "https://cloud.r-project.org/bin/windows/base/old/$env:R_WINDOWS_VERSION/R-$env:R_WINDOWS_VERSION-win.exe" -destfile "R-win.exe"
 Download-File-With-Retries -url "https://cloud.r-project.org/bin/windows/Rtools/$env:RTOOLS_EXE_FILE" -destfile "Rtools.exe"
 
 # Install R
--- a/.ci/test_r_package_windows.ps1
+++ b/.ci/test_r_package_windows.ps1
@@ -51,7 +51,7 @@ if ($env:R_MAJOR_VERSION -eq "3") {
   $env:RTOOLS_EXE_FILE = "rtools40-x86_64.exe"
   $env:R_WINDOWS_VERSION = "4.0.0"
 } else {
-  Write-Output "Unrecognized R version: $env:R_VERSION"
+  Write-Output "[ERROR] Unrecognized R version: $env:R_VERSION"
   Check-Output $false
 }
 
@@ -72,6 +72,10 @@ if ($env:TOOLCHAIN -eq "MINGW") {
   Write-Output "Telling R to use MSYS"
   $install_libs = "$env:BUILD_SOURCESDIRECTORY/R-package/src/install.libs.R"
   ((Get-Content -Path $install_libs -Raw) -Replace 'use_msys2 <- FALSE','use_msys2 <- TRUE') | Set-Content -Path $install_libs
+} elseif ($env:TOOLCHAIN -eq "MSVC") {
+  # no customization for MSVC
+} else {
+  Write-Output "[ERROR] Unrecognized coompiler: $env:TOOLCHAIN"
 }
 
 # download R and RTools
--- a/.ci/test_r_package_windows.ps1
+++ b/.ci/test_r_package_windows.ps1
@@ -14,7 +14,7 @@ function Download-File-With-Retries {
 
 $env:R_LIB_PATH = "$env:BUILD_SOURCESDIRECTORY/RLibrary" -replace '[\\]', '/'
 $env:R_LIBS = "$env:R_LIB_PATH"
-$env:PATH = "$env:R_LIB_PATH/Rtools/bin;" + "$env:R_LIB_PATH/R/bin/x64;" + "$env:R_LIB_PATH/miktex/texmfs/install/miktex/bin/x64;" + $env:PATH
+$env:PATH = "$env:R_LIB_PATH/Rtools/bin;" + "$env:R_LIB_PATH/Rtools/usr/bin;" + "$env:R_LIB_PATH/R/bin/x64;" + "$env:R_LIB_PATH/miktex/texmfs/install/miktex/bin/x64;" + $env:PATH
 $env:CRAN_MIRROR = "https://cloud.r-project.org/"
 $env:CTAN_MIRROR = "https://ctan.math.illinois.edu/systems/win32/miktex/tm/packages/"
 
--- a/.ci/test_r_package_windows.ps1
+++ b/.ci/test_r_package_windows.ps1
@@ -36,7 +36,7 @@ if ($env:R_MAJOR_VERSION -eq "3") {
   Check-Output $false
 }
 
-if ($env:COMPILER -ne "MSVC") {
+if ($env:COMPILER -ne "MINGW") {
   $env:CXX = "$env:RTOOLS_MINGW_BIN/g++.exe"
   $env:CC = "$env:RTOOLS_MINGW_BIN/gcc.exe"
 }
@@ -49,8 +49,7 @@ if ($env:TOOLCHAIN -eq "MINGW") {
   Write-Output "Telling R to use MinGW"
   $install_libs = "$env:BUILD_SOURCESDIRECTORY/R-package/src/install.libs.R"
   ((Get-Content -path $install_libs -Raw) -replace 'use_mingw <- FALSE','use_mingw <- TRUE') | Set-Content -Path $install_libs
-}
-if ($env:TOOLCHAIN -eq "MSYS") {
+} elseif ($env:TOOLCHAIN -eq "MSYS") {
   Write-Output "Telling R to use MSYS"
   $install_libs = "$env:BUILD_SOURCESDIRECTORY/R-package/src/install.libs.R"
   ((Get-Content -path $install_libs -Raw) -replace 'use_msys <- FALSE','use_msys <- TRUE') | Set-Content -Path $install_libs
@@ -70,7 +69,7 @@ Write-Output "Installing Rtools"
 Start-Process -FilePath Rtools.exe -NoNewWindow -Wait -ArgumentList "/VERYSILENT /DIR=$env:R_LIB_PATH/Rtools" ; Check-Output $?
 Write-Output "Done installing Rtools"
 
-# MiKTeX and pandoc can be skipped on non-MINGW builds, since we don't
+# MiKTeX and pandoc can be skipped on MSVC builds, since we don't
 # build the package documentation for those
 if ($env:COMPILER -ne "MSVC") {
     Write-Output "Downloading MiKTeX"
@@ -119,7 +118,7 @@ if ($env:COMPILER -ne "MSVC") {
   $note_str = Get-Content "${LOG_FILE_NAME}" | Select-String -Pattern ' NOTE' | Out-String ; Check-Output $?
   $relevant_line = $note_str -match '.*Status: (\d+) NOTE.*'
   $NUM_CHECK_NOTES = $matches[1]
-  $ALLOWED_CHECK_NOTES = 3
+  $ALLOWED_CHECK_NOTES = 4
   if ([int]$NUM_CHECK_NOTES -gt $ALLOWED_CHECK_NOTES) {
       Write-Output "Found ${NUM_CHECK_NOTES} NOTEs from R CMD check. Only ${ALLOWED_CHECK_NOTES} are allowed"
       Check-Output $False
--- a/.ci/test_r_package_windows.ps1
+++ b/.ci/test_r_package_windows.ps1
@@ -41,7 +41,7 @@ $env:CTAN_PACKAGE_ARCHIVE = "$env:CTAN_MIRROR/tm/packages/"
 #
 # NOTES:
 #    * some paths and file names are different on R4.0
-$env:R_MAJOR_VERSION=$env:R_VERSION.split('.')[0]
+$env:R_MAJOR_VERSION = $env:R_VERSION.split('.')[0]
 if ($env:R_MAJOR_VERSION -eq "3") {
   $env:RTOOLS_MINGW_BIN = "$env:R_LIB_PATH/Rtools/mingw_64/bin"
   $env:RTOOLS_EXE_FILE = "Rtools35.exe"
@@ -67,11 +67,11 @@ tzutil /s "GMT Standard Time"
 if ($env:TOOLCHAIN -eq "MINGW") {
   Write-Output "Telling R to use MinGW"
   $install_libs = "$env:BUILD_SOURCESDIRECTORY/R-package/src/install.libs.R"
-  ((Get-Content -path $install_libs -Raw) -replace 'use_mingw <- FALSE','use_mingw <- TRUE') | Set-Content -Path $install_libs
+  ((Get-Content -Path $install_libs -Raw) -Replace 'use_mingw <- FALSE','use_mingw <- TRUE') | Set-Content -Path $install_libs
 } elseif ($env:TOOLCHAIN -eq "MSYS") {
   Write-Output "Telling R to use MSYS"
   $install_libs = "$env:BUILD_SOURCESDIRECTORY/R-package/src/install.libs.R"
-  ((Get-Content -path $install_libs -Raw) -replace 'use_msys2 <- FALSE','use_msys2 <- TRUE') | Set-Content -Path $install_libs
+  ((Get-Content -Path $install_libs -Raw) -Replace 'use_msys2 <- FALSE','use_msys2 <- TRUE') | Set-Content -Path $install_libs
 }
 
 # download R and RTools
--- a/.ci/test_r_package_windows.ps1
+++ b/.ci/test_r_package_windows.ps1
@@ -67,7 +67,8 @@ Write-Output "Done installing Rtools"
 
 # MiKTeX and pandoc can be skipped on non-MINGW builds, since we don't
 # build the package documentation for those
-if ($env:COMPILER -eq "MINGW") {
+#if ($env:COMPILER -eq "MINGW") {
+if ($false) {
     Write-Output "Downloading MiKTeX"
     Download-File-With-Retries -url "https://miktex.org/download/win/miktexsetup-x64.zip" -destfile "miktexsetup-x64.zip"
     Add-Type -AssemblyName System.IO.Compression.FileSystem
@@ -85,6 +86,9 @@ if ($env:COMPILER -eq "MINGW") {
 Write-Output "Installing dependencies"
 $packages = "c('data.table', 'jsonlite', 'Matrix', 'processx', 'R6', 'testthat'), dependencies = c('Imports', 'Depends', 'LinkingTo')"
 Rscript --vanilla -e "options(install.packages.check.source = 'no'); install.packages($packages, repos = '$env:CRAN_MIRROR', type = 'binary', lib = '$env:R_LIB_PATH')" ; Check-Output $?
+
+Rscript build_r.R
+Check-Output $false
 
 Write-Output "Building R package"
 
--- a/.ci/test_r_package_windows.ps1
+++ b/.ci/test_r_package_windows.ps1
@@ -36,7 +36,7 @@ if ($env:R_MAJOR_VERSION -eq "3") {
   Check-Output $false
 }
 
-if ($env:COMPILER -eq "MINGW") {
+if ($env:COMPILER -ne "MSVC") {
   $env:CXX = "$env:RTOOLS_MINGW_BIN/g++.exe"
   $env:CC = "$env:RTOOLS_MINGW_BIN/gcc.exe"
 }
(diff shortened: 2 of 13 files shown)


StrikerRUS:
@jameslamb Great efforts! Thanks  a lot!

> - gendef.exe removed from Rtools 😭
> - mingw32-make.exe is no longer bundled with Rtools

Is there any available discussion or reason a least why they did it? Or it was done without public discussion? Maybe you know.

jameslamb:
> @jameslamb Great efforts! Thanks a lot!
> 
> > * gendef.exe removed from Rtools 😭
> > * mingw32-make.exe is no longer bundled with Rtools
> 
> Is there any available discussion or reason a least why they did it? Or it was done without public discussion? Maybe you know.

I don't know of any, but the fact that I don't know doesn't mean it was done without public discussion. I haven't had much luck using search engines and these changes are not mentioned [at the official Rtools40 page](), so I just posted a message to the `r-package-devel` mailing list. Will let you know what they say!

<details><summary>message sent to `r-package-devel`</summary>

> Hello,

> I am a maintainer on the LightGBM project, focused on that project's R package. The R package is not available on CRAN yet (we are working on it), so for now our users must build it from source.

> The package includes compilation of a C++ library, and we link to R.dll / R.so to use R-provided functions like Rprintf.

> With the release of R4.0 and Rtools40, we recently received reports from our users that they are unable to build our package on Windows systems with R 4.0 and Rtools 4.0 (https://github.com/microsoft/LightGBM/issues/3064).

> After some investigation, I've learned that the following changes in Rtools40 (relative to Rtools35) broke our installation process:
gendef.exe was removed
mingw32-make.exe was removed
paths like "mingw_64/bin" were changed to "mingw64/bin"
I do not expect that our off-CRAN installation process is supported. I understand that by maintaining our own process, we are taking on the burden of keeping up with new releases of R and Rtools.

> My question is this...is there public documentation about why the changes I mentioned above were made? I have not found any mention of them in the following places, and am not sure where else to look.
https://cran.r-project.org/bin/windows/Rtools/   
https://github.com/r-windows/docs/blob/master/faq.md#readme
https://jeroen.github.io/rstudio2019  
Thank you very much for your time and consideration,

> -James Lamb

</details>

I am going to keep working on this, so we can unblock users who've started upgrading to R 4.0.

jameslamb:
Got a response to the email I sent to the `r-package-devel` mailing list! The responder, Jeroen Ooms, posted https://github.com/microsoft/LightGBM/issues/3064#issuecomment-627171838 as well.

The full response is included below. My takeaways:

1. I think we should move forward with this PR (see my comment in https://github.com/microsoft/LightGBM/issues/3064#issuecomment-627731402)
2. I think we should follow Jeroen's advice and try to use `make.exe` for everything, instead of keeping `mingw32-make.exe` for pre-R4.0 and `make.exe` for R4.0 onwards

<details><summary>Full response</summary>

> Some utilities that were previously bundled with all rtools
installations can now be installed with the package manager.

> To install gendef use:

     pacman -S mingw-w64-{i686,x86_64}-tools

> To install mingw32-make.exe use:

      pacman -S mingw-w64-{i686,x86_64}-make

> To install both of the at once you can use:

      pacman -S mingw-w64-{i686,x86_64}-{tools,make}

> Note that it's often better to use 'make.exe' which is included with
all rtools installations instead of mingw32-make.exe.

> These changes were a result of switching to an msys2 based toolchain.
The default rtools40 installer only includes the things needed to
build CRAN packages or base-R. Extra stuff such as system libraries,
debuggers, cmake, etc, are all optionally available via the package
manager.

</details>

StrikerRUS:
> These changes were a result of switching to an msys2 based toolchain.

Makes sense!
As a side note, with R 4.0 we will get CI test with MSYS toolchain for compiling LightGBM for free in the addition to our current MinGW-w64 and MSVC tests.

> My takeaways:

Absolutely agree with you! (But I'm not an R maintainer!)

StrikerRUS:
@jameslamb Found the following note in CMake docs. Just want to let you know.
> Use this generator under a Windows command prompt with MinGW (Minimalist GNU for Windows) in the PATH and using mingw32-make as the build tool. The generated makefiles use cmd.exe as the shell to launch build rules. They are not compatible with MSYS or a unix shell.
    To build under the MSYS shell, use the MSYS Makefiles generator.
  https://cmake.org/cmake/help/v3.17/generator/MinGW%20Makefiles.html

jameslamb:
> @jameslamb Found the following note in CMake docs. Just want to let you know.
> 
> > Use this generator under a Windows command prompt with MinGW (Minimalist GNU for Windows) in the PATH and using mingw32-make as the build tool. The generated makefiles use cmd.exe as the shell to launch build rules. They are not compatible with MSYS or a unix shell.
> > To build under the MSYS shell, use the MSYS Makefiles generator.
> > https://cmake.org/cmake/help/v3.17/generator/MinGW%20Makefiles.html

🤒 yep thank you for sharing this. I've been experimenting with it tonight and found that to be true. That really complicates things. I think until the package is on CRAN, we need to keep up with the changes in the R toolchain in `install.libs.R`.

That means

* all R versions: Visual Studio if you have it
* R 3.x: MinGW Makefiles + `mingw32-make`
* R 4.x: MSYS Makefiles + `make` (since R 4.0 has moved to MSYS and those are the tools bundled in Rtools40)

StrikerRUS:
@jameslamb 
What logic you think should be applied if user manually specifies `use_mingw` with R 4.0?

jameslamb:
> @jameslamb
> What logic you think should be applied if user manually specifies `use_mingw` with R 4.0?

In my proposal, if you set `use_mingw`, the `MinGW` toolchain (MinGW compilers + `mingw32-make`) will be used, regardless of R version.

If a user does that with Rtools35, it will "just work" because those tools are bundled in Rtools35. If a user does that on R4.0, they will be responsible for making sure there is a `mingw32-make.exe` on PATH, maybe by downloading with `pacman` (https://github.com/microsoft/LightGBM/issues/3064#issuecomment-627171838) or with `mingw-get` from MinGW (http://www.mingw.org/wiki/getting_started) or some other means.

Does that make sense?

jameslamb:
`R CMD check` for MSYS toolchain is failing with this new NOTE I've never seen before:

> * checking for non-standard things in the check directory ... NOTE
Found the following files/directories:
  'data.bin' 'lgb-Dataset.data' 'lgb-model.rds' 'lgb-model.txt'
  'lgb.Dataset.data' 'model.rds' 'model.txt'

Those are all files created from the unit tests or examples. I think that to keep this PR's size from growing (so that it can be reviewed more quickly), we should just bump the allowed NOTEs to 4 here, then have a followup PR that removes it and sets allowed NOTEs back to 3.

jameslamb:
Ok @StrikerRUS @Laurae2 @guolinke this seems to be passing CI, so I think it's ready for review!

Some notes:

* I've labeled this `feature` because I think adding support for a new major version of R is a feature (new-new functionality) not a fix (repairing something that should work but isn't)
* This PR introduces 3 new CI jobs for R 4.0. These do add to our total build times, so I'm open to suggestions if you think this is too much.
    - Mac (Travis)
    - Linux (Travis)
    - Windows with MinGW compiler + MSYS makefiles (AppVeyor)
* The issues for Windows could have been solved by telling R 4.0 users "just download `gendef.exe` and `mingw32-make.exe` and put them on `PATH`", but I really wanted to preserve the experience of the package working just with `CMake` +  `Rtools` (https://github.com/microsoft/LightGBM/issues/3064#issuecomment-627731402)

Laurae2:
( Quickly answering about:

> Those are all files created from the unit tests or examples.

To follow CRAN policies during checks, files created by tested functions must be written in `tempdir()`.

Hence the following:

> then have a followup PR that removes it and sets allowed NOTEs back to 3.

is a good way to prevent this PR growing too large (we can fix it after). )

.appveyor.yml
```diff
@@ -13,10 +13,17 @@ branches:
...
+      R_VERSION: 3.6
+      COMPILER: MINGW
+      TOOLCHAIN: MINGW
+    - TASK: r-package
+      R_VERSION: 4.0
+      COMPILER: MINGW
+      TOOLCHAIN: MSYS
+    - TASK: r-package
+      R_VERSION: 4.0
+      COMPILER: MSVC
```
StrikerRUS, line 26: It's too much I think! Let's keep all possible combinations for #2353. For now it should be enough 3 R jobs: 2 here and 1 at Azure.

.ci/test_r_package_windows.ps1
```diff
@@ -12,32 +12,53 @@ function Download-File-With-Retries {
...
 $env:CTAN_MIRROR = "https://ctan.math.illinois.edu/systems/win32/miktex/tm/packages/"
 
-if ($env:COMPILER -eq "MINGW") {
-  $env:CXX = "$env:R_LIB_PATH/Rtools/mingw_64/bin/g++.exe"
-  $env:CC = "$env:R_LIB_PATH/Rtools/mingw_64/bin/gcc.exe"
+# Get details needed for installing R components
+#
+# NOTES:
+#    * some paths and file names are different on R4.0
+$env:R_MAJOR_VERSION=$env:R_VERSION.split('.')[0]
```
StrikerRUS, line 25:
```suggestion
$env:R_MAJOR_VERSION = $env:R_VERSION.split('.')[0]
```

.ci/test_r_package_windows.ps1
```diff
@@ -12,32 +12,53 @@ function Download-File-With-Retries {
...
+} elseif ($env:R_MAJOR_VERSION -eq "4") {
+  $env:RTOOLS_MINGW_BIN = "$env:R_LIB_PATH/Rtools/mingw64/bin"
+  $env:RTOOLS_EXE_FILE = "rtools40-x86_64.exe"
+  $env:R_WINDOWS_VERSION = "4.0.0"
+} else {
+  Write-Output "Unrecognized R version: $env:R_VERSION"
+  Check-Output $false
+}
+
+if ($env:COMPILER -ne "MINGW") {
```
StrikerRUS, line 39: I'm confused here. Maybe it should be `eq`?

.travis.yml
```diff
@@ -24,7 +24,8 @@ env:
     - TASK=mpi METHOD=pip PYTHON_VERSION=3.7
     - TASK=gpu METHOD=source PYTHON_VERSION=3.5
     - TASK=gpu METHOD=pip PYTHON_VERSION=3.6
-    - TASK=r-package
+    - TASK=r-package R_VERSION=3.6
+    - TASK=r-package R_VERSION=4.0
```
StrikerRUS, line 28: It's too much I think! Let's keep all possible combinations for #2353. We can test different R versions at Travis and Azure, e.g. 4 here and 3 at Azure or vise versa.

R-package/README.md
```diff
@@ -20,14 +20,52 @@ Note: 32-bit (i386) R/Rtools is currently not supported.
...
 
 **Warning for Windows users**: it is recommended to use *Visual Studio* for its better multi-threading efficiency in Windows for many core systems. For very simple systems (dual core computers or worse), MinGW64 is recommended for maximum performance. If you do not know what to choose, it is recommended to use [Visual Studio](https://visualstudio.microsoft.com/downloads/), the default compiler. **Do not try using MinGW in Windows on many core systems. It may result in 10x slower results than Visual Studio.**
 
+**Visual Studio (default)**
+
+By default, the package will be built with [Visual Studio Build Tools](https://visualstudio.microsoft.com/downloads/).
+
+**MinGW (R 3.x)**
+
+If you are using R 3.x and installation fails with Visual Studio, `LightGBM` will fall back to using [MinGW](http://www.mingw.org/). This should work with the tools already bundled in `Rtools` 3.5.
```
StrikerRUS, line 51: I believe the correct link is http://mingw-w64.org/doku.php, because LightGBM cannot be built with original MinGW due to miss of `posix` threads.

R-package/README.md
```diff
@@ -20,14 +20,52 @@ Note: 32-bit (i386) R/Rtools is currently not supported.
...
+
+If you want to force `LightGBM` to use MinGW (for any R version), open `R-package/src/install.libs.R` and change `use_mingw`:
+
+```r
+use_mingw <- TRUE
+```
+
+**MSYS (R 4.x)**
+
+If you are using R 4.x and installation fails with Visual Studio, `LightGBM` will fall back to using [MSYS](http://www.mingw.org/wiki/msys). This should work with the tools already bundled in `Rtools` 4.0.
```
StrikerRUS, line 61: Guess it is https://www.msys2.org/

UPD: Now I'm sure: 

image


https://rstudio.com/resources/rstudioconf-2019/a-preview-of-rtools-4-0/

R-package/src/cmake/modules/FindLibR.cmake
```diff
@@ -41,20 +41,19 @@ function(create_rlib_for_msvc)
...
   endif()
 
   set(LIBR_MSVC_CORE_LIBRARY "${CMAKE_CURRENT_BINARY_DIR}/R.lib" CACHE PATH "R.lib filepath")
 
-  # extract symbols from R.dll into R.def and R.lib import library
-  execute_process(COMMAND ${GENDEF_EXE}
-    "-" "${LIBR_CORE_LIBRARY}"
-    OUTPUT_FILE "${CMAKE_CURRENT_BINARY_DIR}/R.def"
+  execute_process(
+    COMMAND "${LIBR_HOME}/bin/Rscript"
```
StrikerRUS, line 54: I think we should search for `Rscript` similarly to `LIBR_EXECUTABLE`.

.appveyor.yml
jameslamb: Sorry, what do you mean by "keep all possible combinations for #2353"?

Do you mean "if you want more concurrent jobs, write them in GitHub Actions"? I'm happy to try that in a followup PR, but it will take some time.

These are the minimum relevant things to cover for the R package on Windows

* R version
    - 3.6.x
    - 4.0.x
* toolchain:
    - MINGW
    - MSYS
    - Visual Studio
        - 14 (2015)
        - 15 (2017)
        - 16 (2019)

That is 10 different combinations, conservatively. I know we can't test all of them, but I feel we should have at least 4 Windows R builds:

* Visual Studio, 4.0.x
* Visual Studio, 3.6.x
* non-Visual Studio (MinGW), 3.6.x
* non-Visual Studio (MSYS), 4.0.x

If you still disagree and think we should cap it at 3 for now, I'm I guess I'd remove the Visual Studio + R 3.6 one

* Visual Studio, 4.0.x
* ~Visual Studio, 3.6.x~
* non-Visual Studio (MinGW), 3.6.x
* non-Visual Studio (MSYS), 4.0.x

R-package/README.md
jameslamb: oh interesting, I see "msys2" at https://cran.r-project.org/bin/windows/Rtools/ as well. Ok, thank you

.ci/test_r_package_windows.ps1
jameslamb: yes you're right, that was a mistake

jameslamb: this is getting confusing because we now have two dimensions (`COMPILER` and `TOOLCHAIN`). Soon, if we follow the proposal I'll put up for CRAN (https://github.com/jameslamb/LightGBM/pull/25), we'll have a third dimension: `R_BUILD_TYPE` which is either `cmake` or `CRAN`.

I think it might make sense to refactor this file (in a later PR) into some functions like:

* `Install-R`
* `Install-Miktex-And-Pandoc`
* `Run-R-CMD-Check`

Then the top half of the file would define the logic, and the bottom half would look more like ` .travis.yml` build matrix, where you can tell which pieces run for each combination. Something like this pseudocode:

```powershell
Install-R-and-Rtools
if ($env:COMPILER -eq "MSVC") {
    Build-With-Cmake -SkipInstall $false
    Run-Tests-With-Testthat
} else {
   Install-Miktex-and-Pandoc 
   if ($env:TOOLCHAIN -eq "MINGW") {
        Build-With-Cmake -SkipInstall $true -Use "MINGW"
    } elseif ($env:TOOLCAHIN -eq "MSYS") {
        Build-With-Cmake -SkipInstall $true -Use "MSYS"
   }
   Run-R-CMD-Check
}
```

.appveyor.yml
jameslamb: To add more context, I understand what you're saying with https://github.com/microsoft/LightGBM/pull/3065#discussion_r428251729 that maybe there is too much duplication in Mac / Linux builds across Travis + Azure, but for Windows I think it's important to test all four of the cases I mentioned above, since the toolchain changed between major versions

.travis.yml
jameslamb: ok that's a good point. We're getting the same information from both places. I do like that when Azure is failing and Travis isn't (or vice versa), it gives us some information that an issue might be CI-provider-specific, but that's probably not worth the cost of the duplicated builds.

I'll test R4.0 on Travis and 3.6 on Azure (since Azure Linux builds involve updating that CI container). So I think it would actually be 2 in Travis (Mac + Linux, R4.0) and 2 on Azure (Mac + Linux, R3.6)

.appveyor.yml
StrikerRUS: Yeah, I mean that more covering test cases can be added for GitHub Actions to not increase the overall CI time dramatically. 

> - Visual Studio, 4.0.x
> - Visual Studio, 3.6.x

Can you please clarify what will be the difference in these test cases?

.ci/test_r_package_windows.ps1
StrikerRUS: > we'll have a third dimension: R_BUILD_TYPE which is either cmake or CRAN.

Oh my! 😄 

> I think it might make sense to refactor this file (in a later PR) into some functions like:

👍

.appveyor.yml
jameslamb: Got it! Ok yes I agree with that.

> Can you please clarify what will be the difference in these test cases?

* `Visual Studio, 4.0.x` = Visual Studio, R 4.0, and `Rtools40` (would have caught #3064 )
* `Visual Studio, 3.6.x` = Visual Studio, R 3.6.3, and `Rtools35` (would catch if this PR breaks backwards compatibility with older R versions)

R-package/src/cmake/modules/FindLibR.cmake
jameslamb: ok sure! good idea

.appveyor.yml
StrikerRUS: But after this PR is merged, all logic for Visual Studio will be the same (trial projects generation, `.def` file generation). Difference will be only in PATH content, but all issues with R-version-specific paths can be caught by MinGW/MSYS builds. Are my statements correct or I'm missing something?

jameslamb: Even though LightGBM code would look the same, these types of things would not be caught if we don't have a build with Visual Studio + R 3.6.x:

* a linker flag is added to `CMakeLists.txt` in an `if(MSVC)` block that works when linking to the `R.lib` for R 4.0 but not to the `R.lib` for R 3.6.3
* C++ code is added in a block guarded by `#ifdef _MSC_VER` which uses a function that R prohibits (like `abort`), or which conflicts with some other symbol being read in from `R.h` or the other R headers

Basically the existence of `if(MSVC)` in `CMakeLists.txt` and `#ifdef _MSCV_VER` in our C++ code opens the possibility of something being incompatible between a `LightGBM` compiled with MSVC and R 3.6.3.

We know as of this moment that there are no such issues because we have a CI build that tests that combination. If we're very worried about CI time, I am ok dropping the build that is "Visual Studio + R 3.6.3" because of that, but it does mean we're losing some coverage of the issues I mentioned above.

StrikerRUS: Ah, the second point about conflicts with different (due to major version difference) R headers makes a lot of sense! TBH, I didn't get the first one.

OK, keeping in mind that all R jobs will be held by GitHub Actions eventually (I hope), maybe at least we should balance these 4 jobs better for now? Appveyor is extremely slow, you know.

jameslamb: since I made so much progress on #3116 , I'm confident we'll have GitHub Actions in a few days. We could get that reviewed / merged first, then I could change this PR to only add GitHub Actions jobs!

20  concurrent tasks mean we could comfortably add all the different R combinations

R-package/src/cmake/modules/FindLibR.cmake
jameslamb: Take a look at the changes I just pushed to https://github.com/microsoft/LightGBM/commit/fafac03de0b44f19a600a9fb60776dd3765731ac

Instead of repeating the search everywhere, I'm relying on the fact that `R` and `Rscript` are always installed in the same location.

image

R-package/README.md
jameslamb: updated in https://github.com/microsoft/LightGBM/commit/fafac03de0b44f19a600a9fb60776dd3765731ac

R-package/README.md
jameslamb: updated in https://github.com/microsoft/LightGBM/commit/fafac03de0b44f19a600a9fb60776dd3765731ac

.travis.yml
jameslamb: removed the R 3.6 builds on Travis: https://github.com/microsoft/LightGBM/pull/3065/commits/983dd2d2b573b0909e4bd7c1ba8330eba59685eb

docs/GPU-Windows.rst
```diff
@@ -135,7 +135,7 @@ To check whether you need 32-bit or 64-bit MinGW for R, install LightGBM as usua
 If it says ``mingw_64`` then you need the 64-bit version (PATH with ``c:\Rtools\bin;c:\Rtools\mingw_64\bin``),
 otherwise you need the 32-bit version (``c:\Rtools\bin;c:\Rtools\mingw_32\bin``), the latter being a very rare and untested case.
 
---------------
+NOTE: As of `Rtools` 4.0, the `_` in these paths has been removed. If you are using `Rtools` 4.0 or later, the path will have `mingw64` instead of `mingw_64` and `mingw32` instead of `mingw_32`.
```
Laurae2, line 138: > If you are using `Rtools` 4.0 or later, the path will have `mingw64` instead of `mingw_64` and `mingw32` instead of `mingw_32`.

For consistency, change to:

> If you are using `Rtools` 4.0 or later, the path will have ``mingw64`` instead of ``mingw_64`` (PATH with ``C:\rtools40\mingw64\bin``), and ``mingw32`` instead of ``mingw_32`` (``C:\rtools40\mingw32\bin``). The 32-bit version remains an unsupported solution under Rtools 4.0.

.appveyor.yml
StrikerRUS: Amazing!

R-package/src/install.libs.R
```diff
@@ -196,6 +246,7 @@ if (!use_precompile) {
   }
 
   # build the library
+  message("building lib_lightgbm")
```
Laurae2, line 249: `message("building lib_lightgbm")`

For consistency, use upper case:

`message("Building lib_lightgbm")`

R-package/README.md
```diff
@@ -20,14 +20,52 @@ Note: 32-bit (i386) R/Rtools is currently not supported.
...
 #### Windows Preparation
 
-Installing [Rtools](https://cran.r-project.org/bin/windows/Rtools/) is mandatory, and only support the 64-bit version. It requires to add to PATH the Rtools MinGW64 folder, if it was not done automatically during installation.
+Installing a 64-bit version of [Rtools](https://cran.r-project.org/bin/windows/Rtools/) is mandatory.
 
-The default compiler is Visual Studio (or [VS Build Tools](https://visualstudio.microsoft.com/downloads/)) in Windows, with an automatic fallback to Rtools or any [MinGW64](https://sourceforge.net/projects/mingw-w64/files/Toolchains%20targetting%20Win64/Personal%20Builds/mingw-builds/) (x86_64-posix-seh) available (this means if you have only Rtools and CMake, it will compile fine).
+After installing `Rtools` and `CMake`, be sure the following paths are added to the environment variable `PATH`. These may have been automatically added when installing other software.
 
-To force the usage of Rtools / MinGW, you can set `use_mingw` to `TRUE` in `R-package/src/install.libs.R`.
+* `Rtools` MinGW bin:
```
Laurae2, line 27: Can simplify the Rtools example using two examples (to also cover the underscore case) such as:

`Rtools`:
- For `Rtools` 3.x, example: `C:\Rtools\mingw_64\bin`
- For `Rtools` 4.0, example: both `C:\rtools40\mingw64\bin` and `C:\rtools40\usr\bin`

R-package/src/install.libs.R
```diff
@@ -1,7 +1,14 @@
 # User options
 use_precompile <- FALSE
 use_gpu <- FALSE
+
+# Package will be built with Visual Studio unless you set one of these to TRUE
```
Laurae2, line 5: Change to the following (or similar, at least specifying Windows requirement):

```
# Under Windows, the package will be built using Visual Studio unless you set one of these to TRUE
```

R-package/README.md
```diff
@@ -20,14 +20,52 @@ Note: 32-bit (i386) R/Rtools is currently not supported.
...
 
 **Warning for Windows users**: it is recommended to use *Visual Studio* for its better multi-threading efficiency in Windows for many core systems. For very simple systems (dual core computers or worse), MinGW64 is recommended for maximum performance. If you do not know what to choose, it is recommended to use [Visual Studio](https://visualstudio.microsoft.com/downloads/), the default compiler. **Do not try using MinGW in Windows on many core systems. It may result in 10x slower results than Visual Studio.**
 
+**Visual Studio (default)**
+
+By default, the package will be built with [Visual Studio Build Tools](https://visualstudio.microsoft.com/downloads/).
+
+**MinGW (R 3.x)**
+
+If you are using R 3.x and installation fails with Visual Studio, `LightGBM` will fall back to using [MinGW](http://mingw-w64.org/doku.php). This should work with the tools already bundled in `Rtools` 3.5.
```
Laurae2, line 51: For simplification, can change `will fall back to using MinGW` to `will fall back to using Rtools bundled MinGW`.

docs/GPU-Windows.rst
jameslamb: oh good idea, ok

R-package/README.md
jameslamb: good suggestion, thanks!

jameslamb: fixed in https://github.com/microsoft/LightGBM/pull/3065/commits/d1d8022cc36b5bff616192474a938861fcb7afce

R-package/src/install.libs.R
jameslamb: fixed in https://github.com/microsoft/LightGBM/pull/3065/commits/d1d8022cc36b5bff616192474a938861fcb7afce

R-package/README.md
jameslamb: fixed in https://github.com/microsoft/LightGBM/pull/3065/commits/d1d8022cc36b5bff616192474a938861fcb7afce

R-package/src/install.libs.R
jameslamb: fixed in https://github.com/microsoft/LightGBM/pull/3065/commits/d1d8022cc36b5bff616192474a938861fcb7afce

docs/GPU-Windows.rst
jameslamb: made this change in https://github.com/microsoft/LightGBM/pull/3065/commits/d1d8022cc36b5bff616192474a938861fcb7afce

R-package/inst/make-r-def.R
```diff
@@ -0,0 +1,61 @@
...
+# use objdump to dump all the symbols
+OBJDUMP_FILE <- "objdump-out.txt"
+exit_code <- system2(
+    command = "objdump"
+    , args = c(
+        "-p"
+        , shQuote(IN_DLL_FILE)
+    )
+    , stdout = OBJDUMP_FILE
+)
```
jameslamb, line 28: ah! I need to change this to use `{processx}`. The MSYS job is taking 25+ minutes and I'm guessing that is the reason.

jameslamb: Just  pushed https://github.com/microsoft/LightGBM/pull/3065/commits/8f28725e9a2f5b09e2712407942f84f60d0ec4f4, let's see if it helps the time.

You'll see I added a `.pipe_shell_command_to_stdout()` that looks similar to the `.run_shell_command()` we have in other scripts (https://github.com/microsoft/LightGBM/blob/3caf359a79754825aac17ac77d557d0b7288c70e/R-package/src/install.libs.R#L23). That is because `processx::run()` can't be used to redirect command stdout (https://github.com/r-lib/processx/issues/254)

jameslamb:
I just realized that my comment in https://github.com/microsoft/LightGBM/pull/3065#discussion_r432881044 doesn't make sense...MSYS builds should not be using `make-r-def.R`, since it's specific to MSVC builds. Well, the `{processx}` thing is good anyway.

Looking in the logs for [the MSYS build that took 40 minutes](https://ci.appveyor.com/project/guolinke/lightgbm/builds/33216832/job/mvpqnddmgeyicyh3), I see a lot of this

image

Looking into it!

.travis.yml
jameslamb: I just rebased to master and now all R Mac and Travis jobs run on GitHub Actions. Right now I have the matrix set up to run 8 total tasks, all combinations of:

1. operating system: [Mac, Linux]
2. compiler: [gcc, clang]
3. R versions: [3.6, 4.0]

GitHub Actions gives you 20 concurrent builds so I think having 8 R builds is reasonable given the current pace of development on this project. If it becomes too much, it can definitely be cut down.

.appveyor.yml
jameslamb: I've moved all Linux and Mac jobs to GitHub Actions in #3116 , and took your advice to balance Windows jobs better for now (since adding Windows jobs to GitHub Actions was blocked by https://github.com/PowerShell/PowerShell/issues/12823, from https://github.community/t/powershell-steps-fail-nondeterministically/115496/9)

As of this PR, we'll have two R builds on AppVeyor and two on Azure

* AppVeyor, R 4.0, MSYS2
* AppVeyor, R 4.0, Visual Studio 2019
* Azure, R 3.6, MINGW
* Azure, R 3.6, Visual Studio 2017

jameslamb:
@guolinke @StrikerRUS ok I have this passing all CI, and I've added GitHub Actions jobs for R 4.0. Could you please give this a review when you have time?

.appveyor.yml
```diff
@@ -13,10 +13,13 @@ branches:
...
-      TASK: r-package
+    - TASK: r-package
+      R_VERSION: 4.0
+      COMPILER: MINGW
+      TOOLCHAIN: MSYS
+    - TASK: r-package
+      R_VERSION: 4.0
+      COMPILER: MSVC
       APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
     - COMPILER: MSVC
```
StrikerRUS, line 24: Please use the same order for variables inside jobs: either `TASK`-first or `COMPILER`-first for better readbility.

.vsts-ci.yml
```diff
@@ -109,11 +109,17 @@ jobs:
...
     vmImage: 'vs2017-win2016'
   strategy:
-    maxParallel: 4
+    maxParallel: 5
     matrix:
-      r_package:
+      r_package_msvc:
         TASK: r-package
         COMPILER: MSVC
+        R_VERSION: 3.6
```
StrikerRUS, line 117: I think it makes sense to mix R_VERSION: 3.6 for one job and 4.0 for another one. And the same thing for Appveyor.

R-package/README.md
```diff
@@ -20,14 +20,56 @@ Note: 32-bit (i386) R/Rtools is currently not supported.
...
+        - `C:\rtools40\usr\bin`
+* `CMake`
+    - example: `C:\Program Files\CMake\bin`
+* `R`
+    - example: `C\Program Files\R\R-3.6.1\bin`
+
+As of `Rtools` 4.0, some common paths changed and software was removed from `Rtools`. If you are using `R` 4.0 or later and the corresponding `Rtools`, you need to add one additional path to `PATH`.
+
+* `Rtools` usr bin:
+    - example: `C:\Rtools\usr\bin`
```
StrikerRUS, line 41: Why not just add this path to the case ```If you have `Rtools` 4.0, example:```?

R-package/README.md
```diff
@@ -20,14 +20,56 @@ Note: 32-bit (i386) R/Rtools is currently not supported.
...
+
+```r
+use_mingw <- TRUE
+```
+
+**MSYS2 (R 4.x)**
+
+If you are using R 4.x and installation fails with Visual Studio, `LightGBM` will fall back to using [MSYS](https://www.msys2.org/). This should work with the tools already bundled in `Rtools` 4.0.
+
+If you want to force `LightGBM` to use MSYS2 (for any R version), open `R-package/src/install.libs.R` and change `use_msys`:
```
StrikerRUS, line 67:
```suggestion
If you want to force `LightGBM` to use MSYS2 (for any R version), open `R-package/src/install.libs.R` and change `use_msys2`:
```

R-package/README.md
```diff
@@ -20,14 +20,56 @@ Note: 32-bit (i386) R/Rtools is currently not supported.
...
+* `Rtools`
+    - If you have `Rtools` 3.x, example:
+        - `C:\Rtools\mingw_64\bin`
+    - If you have `Rtools` 4.0, example:
+        - `C:\rtools40\mingw64\bin`
+        - `C:\rtools40\usr\bin`
+* `CMake`
+    - example: `C:\Program Files\CMake\bin`
+* `R`
+    - example: `C\Program Files\R\R-3.6.1\bin`
```
StrikerRUS, line 36:
```suggestion
    - example: `C:\Program Files\R\R-3.6.1\bin`
```

R-package/inst/make-r-def.R
```diff
@@ -0,0 +1,98 @@
...
+      })
+    })
+    if (has_processx) {
+        p <- processx::process$new(
+            command = command
+            , args = args
+            , stdout = out_file
+            , windows_verbatim_args = FALSE
+        )
+        invisible(p$wait())
```
StrikerRUS, line 37: Does it really return exit status?
https://processx.r-lib.org/#exit-statuses

Please unify all functions for checking exit statuses:
https://github.com/microsoft/LightGBM/blob/ce95d9ca8aab8fd007a2682d4cc80157b4e1b801/build_r.R#L26
https://github.com/microsoft/LightGBM/blob/3caf359a79754825aac17ac77d557d0b7288c70e/R-package/src/install.libs.R#L23

R-package/inst/make-r-def.R
```diff
@@ -0,0 +1,98 @@
...
+# use objdump to dump all the symbols
+OBJDUMP_FILE <- "objdump-out.txt"
+.pipe_shell_command_to_stdout(
+    command = "objdump"
+    , args = c("-p", IN_DLL_FILE)
+    , out_file = OBJDUMP_FILE
+)
+
+objdump_results <- readLines(OBJDUMP_FILE)
+result <- file.remove(OBJDUMP_FILE)
```
StrikerRUS, line 65: Seems that `result` is not used.

R-package/README.md
```diff
@@ -20,14 +20,56 @@ Note: 32-bit (i386) R/Rtools is currently not supported.
...
+
+If you want to force `LightGBM` to use MinGW (for any R version), open `R-package/src/install.libs.R` and change `use_mingw`:
+
+```r
+use_mingw <- TRUE
+```
+
+**MSYS2 (R 4.x)**
+
+If you are using R 4.x and installation fails with Visual Studio, `LightGBM` will fall back to using [MSYS](https://www.msys2.org/). This should work with the tools already bundled in `Rtools` 4.0.
```
StrikerRUS, line 65:
```suggestion
If you are using R 4.x and installation fails with Visual Studio, `LightGBM` will fall back to using [MSYS2](https://www.msys2.org/). This should work with the tools already bundled in `Rtools` 4.0.
```

R-package/inst/make-r-def.R
```diff
@@ -0,0 +1,98 @@
...
+
+print("------- objdump results --------------")
+print(objdump_results)
```
StrikerRUS, line 68: Remove this please! 100k+ lines of log is just unreadable. It even hangs Appveyor for my laptop  with `Rendering console...` 😄

jameslamb: AH!!!! this was a mistake, thank you

R-package/inst/make-r-def.R
jameslamb: this function isn't intended to return the exit code. It runs the command and throws an error if the command fails. `p$wait()` will raise an R error if the command fails.

`processx::proocess$new()` might do this:

image

or if it takes longer to run, `p$wait()` will do it (I couldn't easily reproduce that).

R-package/inst/make-r-def.R
jameslamb: this was just to suppress a log message (`TRUE` is printed when the file was removed). I'll just make it `invisible()` instead if it will reduce confusion

.ci/test_r_package_windows.ps1
```diff
@@ -115,7 +136,7 @@ if ($env:COMPILER -ne "MSVC") {
   $note_str = Get-Content "${LOG_FILE_NAME}" | Select-String -Pattern ' NOTE' | Out-String ; Check-Output $?
   $relevant_line = $note_str -match '.*Status: (\d+) NOTE.*'
   $NUM_CHECK_NOTES = $matches[1]
-  $ALLOWED_CHECK_NOTES = 3
+  $ALLOWED_CHECK_NOTES = 4
```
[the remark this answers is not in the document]
jameslamb: * https://github.com/microsoft/LightGBM/pull/3065#issuecomment-629834577
* https://github.com/microsoft/LightGBM/pull/3065#issuecomment-629835810

.vsts-ci.yml
jameslamb: ok

.appveyor.yml
jameslamb: ok

.ci/test_r_package_windows.ps1
```diff
@@ -29,34 +29,55 @@ function Download-Miktex-Setup {
...
-if ($env:COMPILER -eq "MINGW") {
+if ($env:TOOLCHAIN -eq "MINGW") {
   Write-Output "Telling R to use MinGW"
   $install_libs = "$env:BUILD_SOURCESDIRECTORY/R-package/src/install.libs.R"
   ((Get-Content -path $install_libs -Raw) -replace 'use_mingw <- FALSE','use_mingw <- TRUE') | Set-Content -Path $install_libs
+} elseif ($env:TOOLCHAIN -eq "MSYS") {
+  Write-Output "Telling R to use MSYS"
+  $install_libs = "$env:BUILD_SOURCESDIRECTORY/R-package/src/install.libs.R"
+  ((Get-Content -path $install_libs -Raw) -replace 'use_msys2 <- FALSE','use_msys2 <- TRUE') | Set-Content -Path $install_libs
 }
```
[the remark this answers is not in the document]
jameslamb: fixed in https://github.com/microsoft/LightGBM/pull/3065/commits/c65a91078d23001962a956abfc915ad7ebfba155

.appveyor.yml
jameslamb: fixed in https://github.com/microsoft/LightGBM/pull/3065/commits/c65a91078d23001962a956abfc915ad7ebfba155

.vsts-ci.yml
jameslamb: changed in https://github.com/microsoft/LightGBM/pull/3065/commits/c65a91078d23001962a956abfc915ad7ebfba155

R-package/inst/make-r-def.R
jameslamb: removed in https://github.com/microsoft/LightGBM/pull/3065/commits/c65a91078d23001962a956abfc915ad7ebfba155

jameslamb:
> Add R version to the name for fast navigation from a PR page:
> https://github.com/microsoft/LightGBM/blob/ce95d9ca8aab8fd007a2682d4cc80157b4e1b801/.github/workflows/main.yml#L13

fixed in https://github.com/microsoft/LightGBM/commit/c65a91078d23001962a956abfc915ad7ebfba155

R-package/inst/make-r-def.R
jameslamb: could you re-phrase? I don't understand what you're asking me to do with with `Please  unify all functions for checking exist statuses`.

I am still opposed to having one shared function used by all three of these scripts. I think it's important that the scripts are self-contained as much as possible.

R-package/inst/make-r-def.R
jameslamb: fixed in https://github.com/microsoft/LightGBM/pull/3065/commits/709ddf3a0b0e5f909b6998d146ba2e0863d65102

.ci/test_r_package_windows.ps1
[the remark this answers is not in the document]
StrikerRUS: Got it! Thanks for clarification.

R-package/inst/make-r-def.R
StrikerRUS: > It runs the command and throws an error if the command fails.

Got it!

> I am still opposed to having one shared function used by all three of these scripts.

Yeah, I remember! I mean, use the same code for all three functions. Or do you keep the difference between them intentionally?

R-package/README.md
```diff
@@ -20,14 +20,56 @@ Note: 32-bit (i386) R/Rtools is currently not supported.
...
+        - `C:\rtools40\usr\bin`
+* `CMake`
+    - example: `C:\Program Files\CMake\bin`
+* `R`
+    - example: `C:\Program Files\R\R-3.6.1\bin`
+
+As of `Rtools` 4.0, some common paths changed and software was removed from `Rtools`. If you are using `R` 4.0 or later and the corresponding `Rtools`, you need to add one additional path to `PATH`.
+
+* `Rtools` usr bin:
+    - example: `C:\Rtools\usr\bin`
```
StrikerRUS, line 41: Why not just add this path to the case ```If you have `Rtools` 4.0, example:```?

.ci/test_r_package_windows.ps1
```diff
@@ -29,34 +29,59 @@ function Download-Miktex-Setup {
...
-  ((Get-Content -path $install_libs -Raw) -replace 'use_mingw <- FALSE','use_mingw <- TRUE') | Set-Content -Path $install_libs
+  ((Get-Content -Path $install_libs -Raw) -Replace 'use_mingw <- FALSE','use_mingw <- TRUE') | Set-Content -Path $install_libs
+} elseif ($env:TOOLCHAIN -eq "MSYS") {
+  Write-Output "Telling R to use MSYS"
+  $install_libs = "$env:BUILD_SOURCESDIRECTORY/R-package/src/install.libs.R"
+  ((Get-Content -Path $install_libs -Raw) -Replace 'use_msys2 <- FALSE','use_msys2 <- TRUE') | Set-Content -Path $install_libs
+} elseif ($env:TOOLCHAIN -eq "MSVC") {
+  # no customization for MSVC
+} else {
+  Write-Output "[ERROR] Unrecognized coompiler: $env:TOOLCHAIN"
```
StrikerRUS, line 78:
```suggestion
  Write-Output "[ERROR] Unrecognized compiler: $env:TOOLCHAIN"
```

.ci/test_r_package_windows.ps1
```diff
@@ -29,34 +29,59 @@ function Download-Miktex-Setup {
...
 }
```
StrikerRUS, line 79:
```suggestion
  Check-Output $false
}
```

.vsts-ci.yml
```diff
@@ -109,11 +109,18 @@ jobs:
...
+        TOOLCHAIN: MSVC
+      r_package_mingw:
```
StrikerRUS, line 119:
```suggestion
      r_package_msys:
```

build_r.R
```diff
@@ -115,6 +115,13 @@ for (src_file in c("lightgbm_R.cpp", "lightgbm_R.h", "R_object_helper.h")) {
   .handle_result(result)
 }
 
+result <- file.copy(
+  from = file.path("R-package", "inst", "make-r-def.R")
+  , to = file.path("lightgbm_r", "inst", "bin/")
```
StrikerRUS, line 120: Not sure here but I remember

> This PR also proposes some simplifications to `build_r.R`. We were repeating the string literal `"lightgbm_r"` a lot, and I think it would be better to store it in one constant at the top of the script.
   https://github.com/microsoft/LightGBM/pull/3087#issue-418377725
```suggestion
  , to = file.path(TEMP_R_DIR, "inst", "bin/")
```

R-package/README.md
jameslamb: huh I thought I did that already! sorry, will do

R-package/inst/make-r-def.R
jameslamb: The functions in `install.libs.R` and `build_r.R` are identical because they do the same thing: run a shell command, raise en error on non-0 exit codes if `strict = TRUE`, and return the exit code.

This function has a different purpose because it is running a command and redirecting the `stdout` of that command to a file. I don't want to add `out_file` as an argument to the function in `install.libs.R` and `build_r.R` (where it isn't needed) or add the option `strict` here (where it will always be `TRUE`.

R-package/README.md
jameslamb: oh! now I remember. That section you're referring to is documenting how the R package will fall back to msys2 if you don't have Visual Studio.

Adding this `/usr/bin` path is important specifically for the case of Visual Studio (the default), because that path is where `objdump.exe` is.

R-package/inst/make-r-def.R
StrikerRUS: OK, I see! Thanks a lot for the explanation!

R-package/README.md
StrikerRUS: Sorry, I didn't get it. I'm referring the following section:

image


Fallbacks are documented much later in the README.

jameslamb: ooooooooh I misunderstood! Yes you're right, ok those two are definitely duplicates

jameslamb:
> LGTM! However, I'm not very happy that now we have 2 Windows R jobs each longer than 10m. Hope they will be moved to GitHub Actions ASAP.
> 

I hope we can change it too, but I really think it's necessary to have those jobs since R 4.0 is so new. Once we merge this I will try again with GitHub Actions and Windows.

> Seems that changing the template of GitHub Actions job name led to duplicated entries in check statuses.
> 
> image

I think that when @guolinke made the 4 GitHub Actions steps from #3119 required, the mechanism in GitHub is to store the job names. So now that we have new names, GitHub thinks 8 non-required things have run and 4 required things have not yet run.

I think the solution will be that once we merge this, the set of required checks needs to be updated. That will have to be done _again_ whenever I manage to move the Windows R jobs, and at least one more time when we introduce CRAN builds (which I am waiting to do until this PR is finalized and merged).

After those things, we should be changing CI jobs much less frequently and focusing on the actual contents of the library.

jameslamb:
Thanks as always for the thorough review @StrikerRUS !

@guolinke can you please review when you have time? I would like an approval from an R maintainer before this is merged.

R-package/README.md
jameslamb: almost missed this one, sorry!  Just fixed this in https://github.com/microsoft/LightGBM/pull/3065/commits/d0e995f443531669c7e5faa9db8eec94b5ff0445.

I also added one more fix in that commit noticed during testing...if using `system2()` instead of `{processx}`, you have to `shQuote(args)` in because on of the args is a file path to `R.dll`, which might have spaces. This isn't necessary with `{processx}` because it does that quoting for you on Windows.

jameslamb:
@guolinke apologies for bothering you...since you are the only admin (I think) on the repo, we'll need your help to merge this. The set of required GitHub Actions tasks needs to change since the names of the GitHub Actions tasks have changed: https://github.com/microsoft/LightGBM/pull/3065#issuecomment-642887012

I think you will need to use administrator privileges to merge this pull request, then go into the repo's settings and change the set of required tasks. Sorry for the inconvenience, it is one of the annoying things about GitHub Actions :/

guolinke:
@jameslamb no problem!

PR #3068, opened 2020-05-11
fix MSVC warning about control paths (fixes #3067)

### LINKED ISSUE #3067: compiler warning: 'not all control paths return a value'
Thanks to @gmobaz for pointing out that building the R package with Visual Studio compilers results in a few compilation warnings (https://github.com/microsoft/LightGBM/issues/3064#issuecomment-626364626). One of these is 'not all control paths return a value'.

We should eliminate this warning if possible, to save users time in debugging.

LightGBM component: R package

## Environment info

Operating System: Windows 10

CPU/GPU model: N/A

C++ compiler version: Visual Studio 16 2019, MSVC 14.24.28314

CMake version: 3.16.5

Java version: N/A

Python version: N/A

R version: 3.6.1

> C:\Users\James\AppData\Local\Temp\RtmpKw7PIG\R.INSTALL3afce51189c\lightgbm\src\src\boosting\prediction_early_stop.cpp(86): warning C4715: 'LightGBM::CreatePredictionEarlyStopInstance': not all control paths return a value [C:\Users\James\AppData\Local\Temp\RtmpKw7PIG\R.INSTALL3afce51189c\lightgbm\rc\build\_lightgbm.vcxproj]

image

## Reproducible example(s)

```shell
Rscript build_r.R
```

## Steps to reproduce

1. Make sure Visual Studio 16 2019 is installed
2. Make sure `Rtools35` is installed and its MinGW is on `PATH`
3. run `Rscript build_r.R` from the root of the repo

jameslamb:
See #3067 for details. When building for the R package, `Log::Fatal` calls `Rf_error`:

https://github.com/microsoft/LightGBM/blob/master/include/LightGBM/utils/log.h#L123

That function will result in a fatal error, but not on the C++ side...it tells the calling R process to throw n R error. I think that is why MSVC thinks it's possible for control to reach the end of `PredictionEarlyStopInstance`.

In this PR, I'd like to propose refactoring `PredictionEarlyStopInstance` to make it clearer to compilers that this function's control flow does always end.

We don't have CI for MSVC + R builds yet (it's in progress in #2965 ), but I tested this on my laptop tonight (Windows 10, Visual Studio 16 2019, R 3.6.1, Rtools 3.5). You can see in this log from running `Rscript build_r.R` that the warning is gone:

image

### DIFF
--- a/src/boosting/prediction_early_stop.cpp
+++ b/src/boosting/prediction_early_stop.cpp
@@ -74,18 +74,18 @@ PredictionEarlyStopInstance CreateBinary(const PredictionEarlyStopConfig& config
 
 PredictionEarlyStopInstance CreatePredictionEarlyStopInstance(const std::string& type,
                                                               const PredictionEarlyStopConfig& config) {
-  if (type != "none" && type != "multiclass" && type != "binary") {
-    Log::Fatal("Unknown early stopping type: %s", type.c_str());
-  }
-
   if (type == "none") {
     return CreateNone(config);
   } else if (type == "multiclass") {
     return CreateMulticlass(config);
+  } else if (type == "binary") {
+    return CreateBinary(config);
+  } else {
+    Log::Fatal("Unknown early stopping type: %s", type.c_str());
   }
 
-  // "binary"
-  return CreateBinary(config);
+  // Fix for compiler warnings about reaching end of control
+  return CreateNone(config);
 }
 
 }  // namespace LightGBM
--- a/src/boosting/prediction_early_stop.cpp
+++ b/src/boosting/prediction_early_stop.cpp
@@ -74,15 +74,19 @@ PredictionEarlyStopInstance CreateBinary(const PredictionEarlyStopConfig& config
 
 PredictionEarlyStopInstance CreatePredictionEarlyStopInstance(const std::string& type,
                                                               const PredictionEarlyStopConfig& config) {
+
+  if (type != "none" && type != "multiclass" && type != "binary") {
+    Log::Fatal("Unknown early stopping type: %s", type.c_str());
+  }
+
   if (type == "none") {
     return CreateNone(config);
   } else if (type == "multiclass") {
     return CreateMulticlass(config);
-  } else if (type == "binary") {
-    return CreateBinary(config);
-  } else {
-    Log::Fatal("Unknown early stopping type: %s", type.c_str());
   }
+
+  // "binary"
+  return CreateBinary(config);
 }
 
 }  // namespace LightGBM
--- a/src/boosting/prediction_early_stop.cpp
+++ b/src/boosting/prediction_early_stop.cpp
@@ -74,7 +74,6 @@ PredictionEarlyStopInstance CreateBinary(const PredictionEarlyStopConfig& config
 
 PredictionEarlyStopInstance CreatePredictionEarlyStopInstance(const std::string& type,
                                                               const PredictionEarlyStopConfig& config) {
-
   if (type != "none" && type != "multiclass" && type != "binary") {
     Log::Fatal("Unknown early stopping type: %s", type.c_str());
   }


StrikerRUS:
@jameslamb I think this warning can be fixed with less diff and without requiring code modification in multiple place for further updates (with the current fix one should update first `if` (`if (type != "none" && type != "multiclass" && type != "binary")`) and only then make actual modification with new `else if (type == "something_new")`). Please refer to for the example: #2866.

https://github.com/microsoft/LightGBM/blob/0a4a7a86f5a1d3146c36c7d8c082154a193d4893/include/LightGBM/prediction_early_stop.h#L15-L23

jameslamb:
> @jameslamb I think this warning can be fixed with less diff and without requiring code modification in multiple place for further updates (with the current fix one should update first `if` (`if (type != "none" && type != "multiclass" && type != "binary")`) and only then make actual modification with new `else if (type == "something_new")`). Please refer to for the example: #2866.
> 
> https://github.com/microsoft/LightGBM/blob/0a4a7a86f5a1d3146c36c7d8c082154a193d4893/include/LightGBM/prediction_early_stop.h#L15-L23

😱 this is great! I'll try it.

jameslamb:
Unfortunately the `nullptr` trick did not work. I used my fork of #2965 to test this on Visual Studio quickly 

* [build](https://dev.azure.com/JayLamb20/jameslamb-lightgbm/_build/results?buildId=507&view=logs&jobId=9f61088a-36ee-57e4-0947-03329fd6ee90&j=9f61088a-36ee-57e4-0947-03329fd6ee90&t=2528b315-ee14-54c7-8330-fcfcaf0cbb8e)
* [commit](https://github.com/jameslamb/LightGBM/pull/22/commits/b5dc61391872297799d33f49097597221e20b27e)

I got this new warning:

> 'return': cannot convert from 'nullptr' to 'LightGBM::PredictionEarlyStopInstance'

StrikerRUS:
> Unfortunately the nullptr trick did not work.

Ah, it was expected. Return type should match function signature. I imbedded code of `PredictionEarlyStopInstance` definition in my previous comment which should be returned from the function you are refactoring in this PR. I guess empty `PredictionEarlyStopInstance` instead of `nullptr` should do the trick.

jameslamb:
> > Unfortunately the nullptr trick did not work.
> 
> Ah, it was expected. Return type should match function signature. I imbedded code of `PredictionEarlyStopInstance` definition in my previous comment which should be returned from the function you are refactoring in this PR. I guess empty `PredictionEarlyStopInstance` instead of `nullptr` should do the trick.

ohhhhhhhh I misunderstood, ok I like that

jameslamb:
Ok just pushed a change to be sure a `PredictionEarlyStopInstance` is returned (even though that line is never actually reached at runtime).

Seems from my fork that this works and suppresses the warning: https://dev.azure.com/JayLamb20/jameslamb-lightgbm/_build/results?buildId=508&view=logs&j=9f61088a-36ee-57e4-0947-03329fd6ee90&t=9f61088a-36ee-57e4-0947-03329fd6ee90

^ don't worry about that build failing. I added this to `.ci/test_r_package_windows.ps1` just to get meaningful install logs and then not spend CI time on tests

```powershell
Rscript build_r.R
Check-Output $false
```

jameslamb:
> Thanks!
> But I don't think that this PR should be marked as `[R-package]`. R-package allowed only to spot the warning due to compiler version.

I think it is R-package specific. I think that the warning is only spotted because `Log::Fatal` ends in calling `Rf_error` for the R package.

This warning doesn't show up for non-R builds with MSVC, e.g. the `sdist` build we do for Python (https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=6111&view=logs&j=67ae2ff8-4b33-578b-0813-9c0440091667&t=130bfb73-8505-5e4f-f030-9ec1bace59ea)

image

StrikerRUS:
> This warning doesn't show up for non-R builds with MSVC, e.g. the sdist build we do for Python 

We use VS 2017 there, but your screenshot from the original message shows that you were using VS 2019. Also please note that the same warnings fixed in #2866 were spotted only in `Debug` mode. So, I believe, `Log::Fatal` as a terminate command indeed causes those warnings but only under some conditions. I think it can be easily checked by running VS 2019 in `Debug` mode.

jameslamb:
> > This warning doesn't show up for non-R builds with MSVC, e.g. the sdist build we do for Python
> 
> We use VS 2017 there, but your screenshot from the original message shows that you were using VS 2019. Also please note that the same warnings fixed in #2866 were spotted only in `Debug` mode. So, I believe, `Log::Fatal` as a terminate command indeed causes those warnings but only under some conditions. I think it can be easily checked by running VS 2019 in `Debug` mode.

ohhhhhhh ok I see! I'm fine to leave `[R-package]` off the description then

jameslamb:
Adding that I just got this same warning on my Mac with `clang++` tonight:

image

jameslamb:
Adding that now I see this warning with MinGW `g++` from `Rtools`

image

This has become an `R CMD check` warning for the CRAN  package, so now I think this PR is officially important for #629 , not just minor maintenance to remove a warning that might distract users.

jameslamb:
gently pinging C++ reviewers for a review. I think this one should be quick :eyes:

src/boosting/prediction_early_stop.cpp
```diff
@@ -83,6 +83,9 @@ PredictionEarlyStopInstance CreatePredictionEarlyStopInstance(const std::string&
   } else {
     Log::Fatal("Unknown early stopping type: %s", type.c_str());
   }
+
+  // Fix for compiler warnings about reaching end of control
+  return CreateNone(config);
```
guolinke, line 88: maybe line 77 to 78 could be removed too?

jameslamb: that would change the behavior of this function. This line I added should technically never be reached (since `Log::Fatal` will be called if an unrecognized `type` is passed).

If 77-78 was deleted, code that uses `type == "none"` would result in `Log::Fatal`

guolinke: thanks, got it.

Issue #3025, opened 2020-04-27
[R-package] MSVC compiler warnings


jameslamb:
Working on #2963 , I see two warnings generated when building the R package using MSVC.

> C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Microsoft\VC\v160\Microsoft.CppBuild.targets(467,5):
    warning MSB8029: The Intermediate directory or Output directory cannot reside under the Temporary directory as it could lead to issues with incremental build.

> config.cpp
C:\Users\James\AppData\Local\Temp\Rtmpy8QQoo\R.INSTALL20184a692267\lightgbm\src\src\boosting\prediction_early_stop.cpp(86):
    warning C4715: 'LightGBM::CreatePredictionEarlyStopInstance': not all control paths return a value

I don't know yet if these are specific to the R package, and I'm not aware of any reported issues related to them. But we should still try to fix these so that other more worrying warnings are not lost in the noise of the logs.

StrikerRUS:
Seems that all issues here were fixed and we can close it, can't we?

jameslamb:
yes, thank you!

Issue #3016, opened 2020-04-23
[R-package] Factor out custom R interface to lib_lightgbm


jameslamb:
## Summary

The R  package calls a C++ library (`lib_lightgbm.so`) with the  R function `.Call()`.  Doing this requires an interface to map R data types to C data types and back. Right now we do not use the [SEXP ("S expression") interface that ships with R's built-in headers](https://cran.r-project.org/doc/manuals/r-release/R-exts.html#System-and-foreign-language-interfaces).

So far our custom implementation has worked well, but there is a risk that future versions of R  will not work with it.

To close this issue, remove `R_object_helper.h`, including `LGBM_SER` and  `LGBM_SE`. Replace them with the corresponding symbols from `R.h`, `Rinternals.h`, and the other R built-in headers.

It is likely possible to do this in pieces....it does not have to be a single pull request.

## Motivation

This may not be strictly necessary to get to CRAN, but it will bring the R package into tighter compatibility with the R build tools, so should improve the chances we'll be able to support new  versions of R with no changes.

I _suspect_, though I don't know for sure, that this would also resolve these warnings we're getting from some compilers:

![image](https://user-images.githubusercontent.com/7608904/80054960-2a956600-84e6-11ea-9d51-18cb5698ce79.png)

See https://github.com/microsoft/LightGBM/pull/3009#issuecomment-617219336 and

jameslamb:
Closing this in favor of being in #2302 , the place where we keep all feature requests. Anyone is welcome to contribute this! If you want to work on it, leave a comment here.

david-cortes:
I don't get the context in full:
* Why does the GH version use `lib_lightgbm` while the CRAN version uses `lightgbm`? Why not make both `lightgbm`?
* What was the reason why it had replacements for R's headers in the first place? Is there something that will break if it were to switch to R's headers?
* Why even the `.Call` are wrapped into some `lgb.*` function?
* If it were to use R headers, should it use R's external pointers system in order to call the C++ destructors? Or was it intentional that those destructors were put into R6 finalizers?
* Is there some particular reason why e.g. the pointers are stored in objects of type `double`? Is the R package meant to be compilable in non-standard CPU architectures?
* What about things like the logger? Should it also switch to R's functions?
* Is the R library meant to be compiled in non-standard ways too? e.g. will you create R builds using msvc or something like that?

jameslamb:
[...]

PR #3071, opened 2020-05-11
v3.0.0rc1

guolinke:
@StrikerRUS @jameslamb @Laurae2 @henry0312  @btrotta 
please list PRs or issues that you think should be fixed in this new release.


StrikerRUS:
As `3.0` major release is a big step forward, I'd like to propose the following TODOs:

PRs:
- #3022
- #3000
- #2992
- #2912
- #2754
- #2463
- #2305 (last remained part)

New Features
- #1493
- #2937
- #2884
- #2604
- Add some features to R-package that are already in C API and Python package (#2281, #2472, #2369, #2983)
- CRAN release of course #629

Fixes:
- All known GPU bugs (#2793, #2648, #2550, #2126, #1717, #1003)
- #2392
- #2363
- #3016

guolinke:
@StrikerRUS 
I agree with you, except for the GPU bugs. 
most of them are hard to reproduce and to fix. also ping @huanzhang12

guolinke:
BTW, I am busy with the 'company work' recently. 
so the new features assigned to me may delay, maybe we can move them to the next release (`3.0.0rc2` or `3.0.0`).
I will have more time after June.

StrikerRUS:
@guolinke 
> I agree with you, except for the GPU bugs.
most of them are hard to reproduce and to fix. also ping @huanzhang12

OK, maybe at least it worth to revisit them? A lot of them haven't got any updates for months.

> BTW, I am busy with the 'company work' recently.
so the new features assigned to me may delay, maybe we can move them to the next release 

Sure! Are you speaking about #2884, right?
From new features section, I personally really would like to see #1493 in next release. I remember you said
> I think it is easy to implement. 
  https://github.com/microsoft/LightGBM/issues/1493#issuecomment-483910794

StrikerRUS:
Also, I'd like to propose a release checklist which we can copy-paste from old releases and it will help us to not forget any required action. Not sure that it will be actual for RC updates though.

- [ ] Update [VERSION](https://github.com/microsoft/LightGBM/blob/master/VERSION.txt) number.
- [ ] Update version in [Appveyor config file](https://github.com/microsoft/LightGBM/blob/master/.appveyor.yml).
- [ ] Update version and release date in [`DESCRIPTION` file of R-package](https://github.com/microsoft/LightGBM/blob/master/R-package/DESCRIPTION).
- [ ] Push commit with `v*` tag to trigger `GitHubRelease` action at Azure Pipelines.
- [ ] Convert automatic GitHub release from `Draft` to normal one.
- [ ] Update [`stable` tag at GitHub](https://github.com/microsoft/LightGBM/releases/tag/stable).
- [ ] Upload release to [PyPI](https://pypi.org/project/lightgbm/).
- [ ] Upload release to CRAN (soon).
- [ ] Upload release to [NuGet](https://www.nuget.org/packages/LightGBM/).
- [ ] Update version and hash in [Homebrew formula](https://github.com/Homebrew/homebrew-core/blob/master/Formula/lightgbm.rb).

guolinke:
refer to the version normalizing problem: https://github.com/pypa/setuptools/issues/308

henry0312:
I think it's time to drop suuport of Python 2.7, since it has reached the end of life.
cf. https://www.python.org/doc/sunset-python-2/
If users that use Python 2.7, they can use LightGBM v2.x

henry0312:
And if you want, I'll create a PR to drop the support.

StrikerRUS:
@henry0312 
> I think it's time to drop suuport of Python 2.7, since it has reached the end of life.

Please take a look at the following comment: https://github.com/microsoft/LightGBM/pull/3056#issuecomment-625525675.
I propose that we mark `3.0.0` release as the latest one which supports Python 2, and stop supporting it with our next release (`3.0.1` or `3.1.0` whatever will be the next after `3.0.0`). I'm looking forward to hear from you what do you think about it!

henry0312:
@StrikerRUS 
Returning to the previous discussion about Semantic Versioning (SemVer) in https://github.com/microsoft/LightGBM/pull/2949, increment of major version can only breaks backward compatibility.
When we drop the support of Python 2.x, we will break backward compatibility because we maintain some (dirty) codes (like https://github.com/microsoft/LightGBM/blob/b60294b81df95c6735285ff76242c448757925e4/python-package/lightgbm/compat.py#L13-L18) to support it.
Therefore, I think it is a just time to only support Python 3.x or later.
And I think as long as popular libraries (including numpy, scipy and pandas) support Python 2. the number of the users to use Python 2 won't decrease, that is not the future Python community hope.

henry0312:
So if we need support of Python 2.x (I don't need any more), I propose we will release  both v2.4.x that will be the last version to support it and v3.x that will support Python 3.5 or later.

jameslamb:
I agree with @StrikerRUS 's suggestions in https://github.com/microsoft/LightGBM/pull/3071#issuecomment-626732193.

#629 is my primary focus, and I do think we should try to finish it for 3.0.0. In my experience more than half of the issues we receive about the R package are related to installation challenges, so I want the 3.0.0 release to come with easier install for R users. That will mean:

* #3016 
* #629 
* #3065 
* #3024 
* #3025 
* #3068 

Since 3.0.0 is an opportunity to make breaking changes,

jameslamb:
I agree with @StrikerRUS 's suggestions in https://github.com/microsoft/LightGBM/pull/3071#issuecomment-626732193.

#629 is my primary focus, and I do think we should try to finish it for 3.0.0. In my experience more than half of the issues we receive about the R package are related to installation challenges, so I want the 3.0.0 release to come with easier install for R users.

Improving installation

* #629 
* #3016 
* #3024 
* #3025 
* #3065 
* #3068 

Improving documentation:

* #1944 

Simplifying the usage of the R package

* #3075

I just created #3075. I think we should take the opportunity of a new major release to remove `lgb.prepare()` and `lgb.prepare_rules()` and move people only to `lgb.prepare2()` and `lgb.prepare_rules2()`.

StrikerRUS:
@henry0312 
> When we drop the support of Python 2.x, we will break backward compatibility because we maintain some (dirty) codes

That's true.

> And I think as long as popular libraries (including numpy, scipy and pandas) support Python 2. the number of the users to use Python 2 won't decrease, that is not the future Python community hope.

All these libraries (our dependencies) are dropping or have already dropped Python 2 support.
pandas: 
> Officially Python 3.6.1 and above, 3.7, and 3.8.
    https://pandas.pydata.org/pandas-docs/stable/getting_started/install.html#python-version-support

numpy: https://numpy.org/neps/nep-0014-dropping-python2.7-proposal.html
scipy : https://www.scipy.org/scipylib/faq.html#do-numpy-and-scipytill-support-python-2-7

Also, from the plot I provided earlier, it can be observed that number of our Python 2 users constantly decreasing starting from January 2020:

image

But still it is several thousand downloads per day for now (~2k). For example daily downloads for macOS are never greater 400, but it is not the reason to not support macOS and stop building wheels for it.

> I propose we will release both v2.4.x that will be the last version to support it and v3.x that will support Python 3.5 or later.

What commits do you want to include in `2.4.x` branch?

btrotta:
I've opened a PR for #2884.

guolinke:
@StrikerRUS 
Do we have a checklist for the remaining issues/requests, for the new release?

StrikerRUS:
@guolinke 
> Do we have a checklist for the remaining issues/requests, for the new release?

I don't think so...

I can make one from https://github.com/microsoft/LightGBM/pull/3071#issuecomment-626732193 and https://github.com/microsoft/LightGBM/pull/3071#issuecomment-627403290.

guolinke:
Thanks @StrikerRUS .
It is being a while, I will push the new release recently.

StrikerRUS:
@guolinke I've just converted lists from https://github.com/microsoft/LightGBM/pull/3071#issuecomment-626732193 and https://github.com/microsoft/LightGBM/pull/3071#issuecomment-627403290 into task lists and updated their statuses. So now it's easy to check the progress.

Also, I'd like to continue the discussion about the removal of Python 2 support. Especially, after creating #3212.

jameslamb:
@guolinke I think we should take the current state of the repo and make it a 2.4.0 release (bumping the minor version because so many new features have been merged since 2.3.1), and then allow PRs with breaking changes that might be included in 3.0.0.

That would allow users to rely on package managers for a stable release while things get broken on the `master` branch on our way to 3.0.0. What do you think?

StrikerRUS:
@jameslamb We already have some breaking commits in `master`: 

image


How is it possible to revert them for `2.4.0` release? I mean, that not selecting them into `2.4.0` can break the whole codebase (e.g. #2699).

jameslamb:
> @jameslamb We already have some breaking commits in `master`:
> 
> image
> 
> How is it possible to revert them for `2.4.0` release? I mean, that not selecting them into `2.4.0` can break the whole codebase (e.g. #2699).

:grimacing: oh I didn't realize that. Ok then I guess we can't do the 2.4.0 idea and need to get to 3.0.0 as fast as possible.

Does it mean I'm free to open PRs with other breaking changes, like #3212 ?

StrikerRUS:
> Does it mean I'm free to open PRs with other breaking changes, like #3212 ?

I think so!

guolinke:
It seems the most remaining PRs/issues cannot be fixed in a short time. 
Can we have a pre-release for v3.0.0, which is stable but may missing some on-going features?
Considering the last version is released about 8 months ago.

ping @StrikerRUS @jameslamb @Laurae2 @henry0312 @btrotta for discussion.

AlbertoEAF:
That would be great! :)
There's a lot of added value in the last months that is not being put to great use :)

Maybe in the future releases don't have to take so long and can be more incremental? If there's enough added value then it might be worth to release.

jameslamb:
it's ok with me if we release a 3.0.0-rc1 now

StrikerRUS:
I don't mind making the release right now. But I'd like to hear something about Python 2 support (https://github.com/microsoft/LightGBM/pull/3071#issuecomment-627416688 and https://github.com/microsoft/LightGBM/pull/3071#issuecomment-655032529).

AlbertoEAF:
Yes, please check the new bug https://github.com/microsoft/LightGBM/issues/3241, which broke the new Fast methods, maybe we should fix that before freezing the new API and going to the 3.0 branch, as in the other hand allows us to further speed up the Fast methods if we change their API a bit more still.

guolinke:
@StrikerRUS 
I agree with 
> I propose that we mark 3.0.0 release as the latest one which supports Python 2, and stop supporting it with our next release (3.0.1 or 3.1.0 whatever will be the next after 3.0.0). I'm looking forward to hear from you what do you think about it!

I remember support python 2.x is easier than R's different version, except that python 2.x needs CI resources to test. 
So, we can drop the CI tests for python 2.x after the 3.0 release.

guolinke:
Hi @AlbertoEAF , any updates for the fix?

AlbertoEAF:
Hello,

It will be finished this weekend.

guolinke:
I plan to make a release before tomorrow (24 hours later).
I will check all existing PRs to ensure all bug fixes are included.

jameslamb:
@guolinke for the open R ones:

* #3269 should be merged before the release because it contains breaking changes
* #3270 and #2912 do not need to make it in

So I am ok with a 3.0.0 release any time as soon as #3269 is merged. As of now I can't think of anything else on the R side that will break, and it's fine for CRAN and other improvements to be in a later release.

StrikerRUS:
#3272 is a `breaking` PR.

guolinke:
@jameslamb do you know why some R's tests failed?

jameslamb:
> @jameslamb do you know why some R's tests failed?

I see this on only the mac builds

```
* checking PDF version of manual ... WARNING
* checking PDF version of manual without hyperrefs or index ... ERROR
Re-running with no redirection of stdout/stderr.
Hmm ... looks like a package
Error in texi2dvi(file = file, pdf = TRUE, clean = clean, quiet = quiet,  : 
  pdflatex is not available
Error in texi2dvi(file = file, pdf = TRUE, clean = clean, quiet = quiet,  : 
  pdflatex is not available
Error in running tools::texi2pdf()
```

And I see

```
==> Downloading from http://mirror.utexas.edu/ctan/systems/mac/mactex/mactex-basictex-20200407.pkg
                                                                                        
sudo: tlmgr: command not found
sudo: tlmgr: command not found
```

So I think this is one of those things where LaTeX mirrors are down :/. We Just have to keep rebuilding, it usually lasts a few hours

jameslamb:
Ok @guolinke the R builds are passing now (https://github.com/microsoft/LightGBM/pull/3269) but I see unrelated failures on Azure and Travis

StrikerRUS:
> We Just have to keep rebuilding,

Confirm, was able to get all green after 5-10 rebuilds.


> but I see unrelated failures on Azure and Travis

Refer to https://github.com/microsoft/LightGBM/pull/3263#issuecomment-669611612.

jameslamb:
ok I just merged #3269 ! So from my perspective, R side is ready for 3.0.0. I can't think of any other improvements that
 are breaking changes.

R-package/DESCRIPTION
```diff
@@ -1,8 +1,8 @@
 Package: lightgbm
 Type: Package
 Title: Light Gradient Boosting Machine
-Version: 2.3.2
-Date: 2019-11-26
+Version: 3.0.0rc1
```
guolinke, line 4:
```suggestion
Version: 3.0.0-rc.1
```

.appveyor.yml
```diff
@@ -1,4 +1,4 @@
-﻿version: 2.3.2.{build}
+﻿version: 3.0.0rc1.{build}
```
guolinke, line 1:
```suggestion
﻿version: 3.0.0-rc.1.{build}
```

VERSION.txt
```diff
@@ -1 +1 @@
-2.3.2
+3.0.0rc1
```
guolinke, line 1:
```suggestion
3.0.0-rc.1
```

R-package/DESCRIPTION
```diff
@@ -1,8 +1,8 @@
 Package: lightgbm
 Type: Package
 Title: Light Gradient Boosting Machine
-Version: 2.3.2
-Date: 2019-11-26
+Version: 3.0.0-rc.1
```
guolinke, line 4: It seems R doesn't support `rc` in version number. 
I am going to replace it with `3.0.0.1` in R.

.appveyor.yml
```diff
@@ -1,4 +1,4 @@
-﻿version: 2.3.2.{build}
+version: 3.0.0-rc.1.{build}
```
guolinke, line 1:
```suggestion
version: 3.0.0rc1.{build}
```

VERSION.txt
```diff
@@ -1 +1 @@
-2.3.2
+3.0.0-rc.1
```
guolinke, line 1:
```suggestion
3.0.0rc1
```

R-package/DESCRIPTION
guolinke, line 4:
```suggestion
Version: 3.0.0.1
```

.appveyor.yml
```diff
@@ -1,4 +1,4 @@
-﻿version: 2.3.2.{build}
+version: 3.0.0rc1.{build}
```
guolinke, line 1:
```suggestion
version: 3.0.0-1.{build}
```

VERSION.txt
guolinke, line 1:
```suggestion
3.0.0-1
```

VERSION.txt
```diff
@@ -1 +1 @@
-2.3.2
+3.0.0-1
```
guolinke, line 1:
```suggestion
3.0.0rc1
```

.appveyor.yml
```diff
@@ -1,4 +1,4 @@
-﻿version: 2.3.2.{build}
+version: 3.0.0-1.{build}
```
guolinke, line 1:
```suggestion
version: 3.0.0rc1.{build}
```

R-package/src/VERSION.txt
```diff
@@ -0,0 +1 @@
+3.0.0-1
```
guolinke, line 1: R now has its own version file.

jameslamb: What's the reason?

You'll also have to change these places that were using the main LightGBM `VERSION.txt`:

* https://github.com/microsoft/LightGBM/blob/53c64bac52dcbd43505bb3363b00048ecf0e3417/build-cran-package.sh#L24
* https://github.com/microsoft/LightGBM/blob/53c64bac52dcbd43505bb3363b00048ecf0e3417/.ci/test_r_package.sh#L137
* https://github.com/microsoft/LightGBM/blob/53c64bac52dcbd43505bb3363b00048ecf0e3417/R-package/configure.ac#L7

guolinke:
@jameslamb it seems now is complicated to update R's version. 
Some fails:
```
The downloaded source packages are in
	‘/tmp/RtmpNiwgDT/downloaded_packages’
'configure' in the R package has changed. Please recreate it and commit the changes.
Changed files:
 R-package/configure | 18 +++++++++---------
 1 file changed, 9 insertions(+), 9 deletions(-)
See R-package/README.md for details on how to recreate this script.
```
should we manually run some scripts when update version number?
I don't have the machine with R env right now, could you help me to fix these R's fails?

R-package/src/VERSION.txt
guolinke: R doesn't support `rc`` in version number...
I tried several solutions, and find `-1` can work.
However, using `-1` in python package will be normalized to `post1`, so I use the standalone version number for R.

guolinke:
due to the problems in R's version number, I think the release will delay for an additional day.

guolinke:
BTW, this cause the CI error.
```
* checking if this is a source package ... WARNING
Subdirectory ‘src’ contains:
  VERSION.txt
These are unlikely file names for src files.
```

.github/workflows/main.yml
```diff
@@ -137,7 +137,7 @@ jobs:
           export CONDA_ENV="test-env"
           export CONDA="$HOME/miniconda"
           export PATH="$CONDA/bin:${HOME}/.local/bin:$PATH"
-          export LGB_VER=$(head -n 1 VERSION.txt)
+          export LGB_VER=$(head -n 1 R-package/src/VERSION.txt)
```
guolinke, line 140:
```suggestion
```

.gitignore
```diff
@@ -408,7 +408,6 @@ R-package/src/lib_lightgbm.so.dSYM/
 R-package/src/src/
 R-package/src-x64
 R-package/src-i386
```
guolinke, line 410:
```suggestion
R-package/src-i386
R-package/**/VERSION.txt
```

guolinke:
@jameslamb I am trying to de-couple R with `version.txt`. 
For this:
https://github.com/microsoft/LightGBM/blob/53c64bac52dcbd43505bb3363b00048ecf0e3417/R-package/configure.ac#L7
, is that possible to read the version number from `Description`?

jameslamb:
> @jameslamb I am trying to de-couple R with `version.txt`.
> For this:
> https://github.com/microsoft/LightGBM/blob/53c64bac52dcbd43505bb3363b00048ecf0e3417/R-package/configure.ac#L7
> 
> 
> , is that possible to read the version number from `Description`?

I think reading from `DESCRIPTION` is too complicated. I don't want to do YAML processing inside that `configure` script.

The documentation on how to update `configure` is here: https://github.com/microsoft/LightGBM/blob/master/R-package/README.md#changing-the-cran-package

But I can update it and fix the other warnings you're seeing right now! Can I push to this branch?

R-package/configure
```diff
@@ -1,6 +1,6 @@
 #! /bin/sh
 # Guess values for system-dependent variables and create Makefiles.
-# Generated by GNU Autoconf 2.69 for lightgbm 2.3.2.
```
jameslamb, line 3: by the way,  this  file  is automatically generated and should  never be  updated manually:  https://github.com/microsoft/LightGBM/blob/master/R-package/README.md#preparing-a-cran-package-and-installing-it

jameslamb:
@guolinke I just pushed https://github.com/microsoft/LightGBM/pull/3071/commits/5d79e77ba1e2657070bc5d7320ea082f2920784c which should fix the other  issues with  the R package. That hard-coding is fine for right now to not delay the release. I will improve the automation in a follow-up PR.

StrikerRUS:
@guolinke Will RCs be a part of our regular development process? I'm asking because 3.0 looks like an edge case as 3.0 will never be published on CRAN (R-package is not ready yet as per #629). So I believe we can set `VERSION` to 3.0, then manually postprocess names of files for PyPI (add `rc` suffix) and bump the version for development (as we do after each release) to `3.0.1`.

guolinke:
@StrikerRUS 
Actually, I think the new solution for R version number is better.
In previous, it depends on both the number in `DESCRIPTION` and `version.txt`, and `version.txt` is mainly used to locate the generate package file in tests. So it is safe to decouple it.
And using `3.0.0-1` is more uniform with `3.0.0rc1`. Refer to https://semver.org/#spec-item-9 .

@jameslamb in this case, Maybe we can generate the version number to `DESCRIPTION`, instead of manually update it?
I mean,  auto-fill the version number (and date) to `DESCRIPTION`  in `recreate-configure.sh`.
No in this release, I can enhance this latter.

guolinke:
@StrikerRUS @jameslamb 
I will merge this in about 2 hours. Do we have any other problems?

StrikerRUS:
@guolinke 
> I will merge this in about 2 hours.

OK!

Are you planning to go through all items from https://github.com/microsoft/LightGBM/pull/3071#issuecomment-626924823 (except updating `stable` tag of course)? I think RC release to PyPI will be enough. Not sure about NuGet though.

jameslamb:
I'm ok with a release any time. Trying to fix #3280 for it right now.

For R, after this release, we could do this:

1.  add a placeholder in `DESCRIPTION` like `~~VERSION~~`
2. in `build_r.R`, and `build-cran-package.sh`, replace that with the version in `VERSION.txt` (and similar approach for Date)
3. add a similar placeholder to `configure.ac`, and have `recreate-configure.sh` and `build-cran-package.sh` replace it. That way, we never have to rely on having `configure.ac` read a file
4. All this code that reads `VERSION.txt` can replace any `rc` with `-`

If we did all of that, it could be totally automated and the only thing that has to be changed is `VERSION.txt`.

And I think we could create a `Makefile` with a `release` recipe that handles all these things.

guolinke:
@StrikerRUS very strange, it seems azure pipeline does not build for `tag` now.

guolinke:
Thanks all contributors so much! The new version is released!   🍺🍺🍺

The python package is uploaded. you can install it via ```pip install lightgbm==3.0.0rc1```.

For the NuGet package, I think we can release it in the `3.0.0` version.
@StrikerRUS how about the homebrew version? should we update it now or wait for `3.0.0`?

StrikerRUS:
> how about the homebrew version?

I think PyPI is enough for and homebrew can wait for the stable version.

PR #3072, opened 2020-05-11
[R-package] [docs] remove devtools dependency for docs builds (fixes #3036)

### LINKED ISSUE #3036: Bump version of pkgdown
## Summary

Build R documents with new version of `pkgdown`.

## Motivation

`1.5` version introduces `repo$url` config option. This option will allow to drop our ugly postprocessing JS workaround for incorrect source files links as our R package is located in a subdirectory:
https://github.com/microsoft/LightGBM/blob/master/R-package/pkgdown/extra.js

## References
https://github.com/microsoft/LightGBM/blob/2c18a0f3ed732df77d4304ed1d2b46a345d661c8/docs/conf.py#L238

> You can now control the links to source files (in reference pages and
articles) and issues and users (in the NEWS) with new repo$url config
option (https://github.com/r-lib/pkgdown/issues/1238). This makes it easier to use pkgdown with GitHub enterprise, packages in subdirectories, and other source hosts (like bitbucket).

https://github.com/r-lib/pkgdown/releases/tag/v1.5.0

jameslamb:
I'd like to propose removing the dependency on `devtools` in our readthedocs builds. We are only using it for `devtools::document()`, which is just a thin wrapper around `roxygen2::roxygenize()`.

`devtools` is a very expensive dependency (it depends recursively on successfully building 101 other R packages) compared to `roxygen2` (41 recursive dependencies), and we're already installing `roxygen2` anyway.

I'm also proposing bumping up our version of `roxygen2` to 7.1.0, the most recent version used in `DESCRIPTION` 

https://github.com/microsoft/LightGBM/blob/05d89a1a1f48a6096e3d2342a517e048d6765bdd/R-package/DESCRIPTION#L44

### How I got dependency counts

I figured out those recursive dependency numbers with the `pkgnet` package, a project I work on with some friends from a previous job. In an R environment, you can run `pkgnet::CreatePackageReport('devtools')` (for example), to see a summary of the dependencies.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -232,7 +232,6 @@ def generate_r_docs(app):
         r-testthat=2.0.0=r351h29659fb_0 \
         cmake=3.14.0=h52cb24c_0
     /home/docs/.conda/bin/conda install -q -y -n r_env -c conda-forge \
-        ca-certificates=2020.4.5.1=hecc5488_0 \
         r-data.table=1.12.8=r35hcdcec82_0 \
         r-pkgdown=1.5.1=r35h6115d3f_0 \
         r-roxygen2=7.1.0=r35h0357c0b_0
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -230,12 +230,12 @@ def generate_r_docs(app):
         r-jsonlite=1.5=r351h96ca727_0 \
         r-matrix=1.2_14=r351h96ca727_0 \
         r-testthat=2.0.0=r351h29659fb_0 \
-        cmake=3.14.0=h52cb24c_0 \
-        ca-certificates=2019.11.27=0
+        cmake=3.14.0=h52cb24c_0
     /home/docs/.conda/bin/conda install -q -y -n r_env -c conda-forge \
+        ca-certificates=2020.4.5.1=hecc5488_0 \
         r-data.table=1.12.8=r35hcdcec82_0 \
-        r-pkgdown=1.3.0=r35h6115d3f_1001 \
-        r-roxygen2=7.1.0=r36h0357c0b_1
+        r-pkgdown=1.5.1=r35h6115d3f_0 \
+        r-roxygen2=7.1.0=r35h0357c0b_0
     source /home/docs/.conda/bin/activate r_env
     export TAR=/bin/tar
     cd {0}
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -240,6 +240,7 @@ def generate_r_docs(app):
     export TAR=/bin/tar
     cd {0}
     sed -i'.bak' '/# Build the package (do not touch this line!)/q' build_r.R
+    export R_LIBS="/home/docs/.conda/envs/r_env/lib/R/library"
     Rscript build_r.R
     Rscript build_r_site.R
     """.format(os.path.join(CURR_PATH, os.path.pardir))
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -239,7 +239,6 @@ def generate_r_docs(app):
     source /home/docs/.conda/bin/activate r_env
     export TAR=/bin/tar
     cd {0}
-    #sed -i'.bak' '/# Build the package (do not touch this line!)/q' build_r.R
     export R_LIBS="/home/docs/.conda/envs/r_env/lib/R/library"
     Rscript build_r.R
     Rscript build_r_site.R
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -238,7 +238,7 @@ def generate_r_docs(app):
     source /home/docs/.conda/bin/activate r_env
     export TAR=/bin/tar
     cd {0}
-    export R_LIBS="/home/docs/.conda/envs/r_env/lib/R/library"
+    export R_LIBS="$CONDA_PREFIX/lib/R/library"
     Rscript build_r.R
     Rscript build_r_site.R
     """.format(os.path.join(CURR_PATH, os.path.pardir))
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -239,7 +239,7 @@ def generate_r_docs(app):
     source /home/docs/.conda/bin/activate r_env
     export TAR=/bin/tar
     cd {0}
-    sed -i'.bak' '/# Build the package (do not touch this line!)/q' build_r.R
+    #sed -i'.bak' '/# Build the package (do not touch this line!)/q' build_r.R
     export R_LIBS="/home/docs/.conda/envs/r_env/lib/R/library"
     Rscript build_r.R
     Rscript build_r_site.R
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -227,7 +227,6 @@ def generate_r_docs(app):
     echo 'options(repos = "https://cran.rstudio.com")' > $HOME/.Rprofile
     /home/docs/.conda/bin/conda create -q -y -n r_env \
         r-base=3.5.1=h1e0a451_2 \
-        r-devtools=1.13.6=r351h6115d3f_0 \
         r-jsonlite=1.5=r351h96ca727_0 \
         r-matrix=1.2_14=r351h96ca727_0 \
         r-testthat=2.0.0=r351h29659fb_0 \
@@ -236,7 +235,7 @@ def generate_r_docs(app):
     /home/docs/.conda/bin/conda install -q -y -n r_env -c conda-forge \
         r-data.table=1.12.8=r35hcdcec82_0 \
         r-pkgdown=1.3.0=r35h6115d3f_1001 \
-        r-roxygen2=6.1.1=r35h0357c0b_1001
+        r-roxygen2=7.1.0=r36h0357c0b_1
     source /home/docs/.conda/bin/activate r_env
     export TAR=/bin/tar
     cd {0}
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -12,7 +12,7 @@
 # main C++ code
 include/    @guolinke @chivee @btrotta
 src/    @guolinke @chivee @btrotta
-CmakeLists.txt    @guolinke @chivee @Laurae2 @jameslamb @wxchan @henry0312 @StrikerRUS @huanzhang12 @btrotta
+CMakeLists.txt    @guolinke @chivee @Laurae2 @jameslamb @wxchan @henry0312 @StrikerRUS @huanzhang12 @btrotta
 
 # R code
 include/LightGBM/lightgbm_R.h    @Laurae2 @jameslamb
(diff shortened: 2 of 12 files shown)


jameslamb:
@StrikerRUS I created this as a LightGBM branch, not on my fork. Could you enable it on readthedocs so we can test?

StrikerRUS:
@jameslamb 
> Could you enable it on readthedocs so we can test?

Sure, done!

docs/conf.py
```diff
@@ -236,7 +235,7 @@ def generate_r_docs(app):
     /home/docs/.conda/bin/conda install -q -y -n r_env -c conda-forge \
         r-data.table=1.12.8=r35hcdcec82_0 \
         r-pkgdown=1.3.0=r35h6115d3f_1001 \
-        r-roxygen2=6.1.1=r35h0357c0b_1001
+        r-roxygen2=7.1.0=r36h0357c0b_1
```
StrikerRUS, line 238: @jameslamb 

Build hangs due to `conda` cannot resolve this issue: `r-base=3.5` and `r-roxygen2=r36...`

> I'm also proposing bumping up our version of `roxygen2` to 7.1.0, the most recent version used in `DESCRIPTION` 

I remember it took me several days to find out the concrete versions of packages that do not conflict with each other and do not lead to the error about that DLL could not be loaded (one famous upstream `pkgdown` issue unfortuanately I cannot find references to). I believe that bumping roxygen2 version can be done in a separate PR or together with #3036 (most likely that we will have to bump all versions due to 1 change).

jameslamb: hmmmm actually I think there is so much work involved in doing this, it might be worth it for me to take on #3036 within this PR.

StrikerRUS: > hmmmm actually I think there is so much work involved in doing this, ...

Yeah, indeed. So maybe delay this for later time and concentrate on CRAN release? I mean, docs are run OK right now and this is not something user-related.

jameslamb: I don't want to pick back up the next big CRAN step (https://github.com/jameslamb/LightGBM/pull/15 ) while #2965 is awaiting review from R reviewers, since I really need the CI coverage and speedups there to move forward with confidence.

So until we get that merged, I'm spending some time on these other side things.

jameslamb:
@StrikerRUS hey it looks like this is working!

* build: https://readthedocs.org/projects/lightgbm/builds/11019657/
* docs: https://lightgbm.readthedocs.io/en/fix-remove-devtools/

Some notes:

* I had to replace `devtools::document()` with `roxygen2::roxygenize(load='installed')` (not just `roxygen2::roxygenize()`) because of  https://github.com/r-lib/roxygen2/issues/822#issuecomment-533183162
* given ^, it's important that `conda`'s `R` knows to look for `lightgbm` in the same place it was installed by `build_r.R`. That's why I added `export R_LIBS...`
* the first few builds were very tough to debug on https://readthedocs.org/projects/lightgbm/builds/ because the logs from `build_r.R` were  disabled by the `sed` call. Can we please remove it so the logs can be used for debugging? They aren't THAT long, and I think they're valuable

By "it's working", I mean updating `pkgdown` and `roxygen2` in the conda environment. I still need too add the use of `pkgdown` url option for #3036. Trying that next.

jameslamb:
The changes for #3036 do not seem to be working :/

* build: https://readthedocs.org/projects/lightgbm/builds/11019779/
* docs: https://lightgbm.readthedocs.io/en/fix-remove-devtools/R/reference/

The GitHub logo was updated correctly, but links like `Source: R/lightgbm.R` point to the wrong place.

For example, `Source: R/lightgbm.R` on https://lightgbm.readthedocs.io/en/fix-remove-devtools/R/reference/agaricus.train.html points to https://github.com/Microsoft/LightGBM/blob/master/R/lightgbm.R

I'm investigating

jameslamb:
Ok I think this is fully working now and closes #3036 . This is ready for review.

docs/conf.py
StrikerRUS: Ah, I see, you are full of power! Great! 🎉

StrikerRUS:
> the first few builds were very tough to debug on https://readthedocs.org/projects/lightgbm/builds/ because the logs from build_r.R were disabled by the sed call. Can we please remove it so the logs can be used for debugging? They aren't THAT long, and I think they're valuable

Are you speaking about this `sed`?

https://github.com/microsoft/LightGBM/blob/b60294b81df95c6735285ff76242c448757925e4/docs/conf.py#L243
This `sed` was used to prevent double building of LightGBM similarly to trick for CI with `R cmd check` and allows us to fit in RTD build time limits.


You can safely drop `ca-certificates` dependency now.

In `build_reference()` `document` argument is replaced by `devel` since `1.4.0` version.
https://github.com/r-lib/pkgdown/blob/c181e5c9ac09b28974a97f5e32deddc36690fcf2/R/build-reference.R#L139

Refer to https://github.com/r-lib/pkgdown/releases/tag/v1.4.0
> build_site(), build_reference() and build_home() gain a parameter
devel which controls whether you're in deployment or development mode.
It generalises and replaces (with deprecation) the existing document
argument.

jameslamb:
> > the first few builds were very tough to debug on https://readthedocs.org/projects/lightgbm/builds/ because the logs from build_r.R were disabled by the sed call. Can we please remove it so the logs can be used for debugging? They aren't THAT long, and I think they're valuable
> 
> Are you speaking about this `sed`?
> 
> https://github.com/microsoft/LightGBM/blob/b60294b81df95c6735285ff76242c448757925e4/docs/conf.py#L243
> 
> 
> This `sed` was used to prevent double building of LightGBM similarly to trick for CI with `R cmd check` and allows us to fit in RTD build time limits.
> You can safely drop `ca-certificates` dependency now.
> 
> In `build_reference()` `document` argument is replaced by `devel` since `1.4.0` version.
> https://github.com/r-lib/pkgdown/blob/c181e5c9ac09b28974a97f5e32deddc36690fcf2/R/build-reference.R#L139
> 
> Refer to https://github.com/r-lib/pkgdown/releases/tag/v1.4.0
> 
> > build_site(), build_reference() and build_home() gain a parameter
> > devel which controls whether you're in deployment or development mode.
> > It generalises and replaces (with deprecation) the existing document
> > argument.

Oh ok! Yes I misunderstood what that `sed` command was doing, thanks for the clarification. But I still think we should remove it. `roxygen2::roxygenize(load = "installed")` will say "don't try to rebuild the package, just assume it has already been built and installed".

I'll try removing `ca-certificates` and updating `build_reference()` call tonight.

jameslamb:
Seems that removing `ca-certificates` worked! 

* build: https://readthedocs.org/projects/lightgbm/builds/11027879/
* docs: https://lightgbm.readthedocs.io/en/fix-remove-devtools/R/reference/agaricus.train.html

jameslamb:
close-reopen to re-trigger CI. Azure got stuck trying to download MiKTeX :grimacing:

docs/conf.py
```diff
@@ -227,20 +227,18 @@ def generate_r_docs(app):
...
         r-data.table=1.12.8=r35hcdcec82_0 \
-        r-pkgdown=1.3.0=r35h6115d3f_1001 \
-        r-roxygen2=6.1.1=r35h0357c0b_1001
+        r-pkgdown=1.5.1=r35h6115d3f_0 \
+        r-roxygen2=7.1.0=r35h0357c0b_0
     source /home/docs/.conda/bin/activate r_env
     export TAR=/bin/tar
     cd {0}
-    sed -i'.bak' '/# Build the package (do not touch this line!)/q' build_r.R
+    export R_LIBS="/home/docs/.conda/envs/r_env/lib/R/library"
```
StrikerRUS, line 241: We may encounter errors if RTD maintainers decide to change default folder for storing conda environments.
```suggestion
    export R_LIBS="$CONDA_PREFIX/lib/R/library"
```

docs/conf.py
```diff
@@ -227,20 +227,18 @@ def generate_r_docs(app):
...
     /home/docs/.conda/bin/conda install -q -y -n r_env -c conda-forge \
         r-data.table=1.12.8=r35hcdcec82_0 \
-        r-pkgdown=1.3.0=r35h6115d3f_1001 \
-        r-roxygen2=6.1.1=r35h0357c0b_1001
+        r-pkgdown=1.5.1=r35h6115d3f_0 \
+        r-roxygen2=7.1.0=r35h0357c0b_0
     source /home/docs/.conda/bin/activate r_env
     export TAR=/bin/tar
     cd {0}
-    sed -i'.bak' '/# Build the package (do not touch this line!)/q' build_r.R
```
StrikerRUS, line 243: OK, indeed `roxygenize(load = "installed")` results in that LightGBM library is built only once! Great! Please remove this creepy comment in the source file then 😃 .
```
(do not touch this line!)
```

docs/conf.py
jameslamb: ah ok, we need to change it in all three place it's used then. I'll do that

docs/conf.py
jameslamb: oh right! yep I'll remove that

docs/conf.py
jameslamb: how did you know we could use `CONDA_PREFIX`? I don't see that documented in https://docs.readthedocs.io/en/latest/guides/conda.html, and I don't know what the corresponding environment variable should be to replace our two current uses of `/home/docs/.conda/bin/`...or do we not need to worry about those?

StrikerRUS: > ah ok, we need to change it in all three place it's used then. I'll do that

Nope! `$CONDA_PREFIX` is created only after `activate` script. It is pointing to the **current active** environment location.

> how did you know we could use `CONDA_PREFIX`? 

https://docs.conda.io/projects/conda-build/en/latest/user-guide/environment-variables.html#id6

> or do we not need to worry about those?

Yeah! WE should leave them as is because they are environment-independent.

jameslamb: ohhhhhhhhh I see, ok! Thanks for the explanation.

I did find that page when I searched fro `CONDA_PREFIX`, but disregarded it because I thought that `conda-build` was only used for creating package recipes and uploading the packages built from those recipes to conda channels. That is what the [conda-build homepage](https://docs.conda.io/projects/conda-build/en/latest/) says:

> Conda-build contains commands and tools to use conda to build your own packages. It also provides helpful tools to constrain or pin versions in recipes. Building a conda package requires installing conda-build and creating a conda recipe. You then use the conda build command to build the conda package from the conda recipe.

StrikerRUS: You are right! Even conda maintainers agree with that it is hard to find docs about `CONDA_PREFIX`:
https://github.com/conda/conda/issues/2764#issuecomment-339860431

jameslamb: WOW what a small world! The commenter right above that link, @mGalarnyk, and I used to work together in Chicago!

PR #3076, opened 2020-05-13
use incremental linking for MSVC builds

jameslamb:
Moving this proposed changed from https://github.com/microsoft/LightGBM/pull/2965#discussion_r423135643 .  While working on #2965 , I was searching around for ways to speed up builds with MSVC.

I came across a Microsoft blog, ["Recommendations to speed C++ builds in Visual Studio"](https://devblogs.microsoft.com/cppblog/recommendations-to-speed-c-builds-in-visual-studio/).

One of its recommendations is to use the linker option [`/incremental`](https://docs.microsoft.com/en-us/cpp/build/reference/incremental-link-incrementally?view=vs-2019)

> /incremental – Incremental link
Incremental link enables the linker to significantly speed up link times. With this feature turned on, linker can process just the diffs between two links to generate the image and thus speed up link times by 4-10x in most cases after the first build. In VS2015 this feature was enhanced to handle additional common scenarios that were previously not supported.

> Recommendation:
> DO USE linker switch /incremental 

In this PR, I propose adding that flag to our MSVC builds. Looking for feedback from other maintainers, since I am not very familiar with Visual Studio compilers.

## Questions for reviewers

* should we do this at all?
* if yes, should we add this to `LightGBM.vcxproj` per https://github.com/microsoft/LightGBM/pull/2965#discussion_r423135643?


StrikerRUS:
I'd better delegate this to C++ maintainers as they know much more about MSVC linking flags than me.

jameslamb:
> I'd better delegate this to C++ maintainers as they know much more about MSVC linking flags than me.

ah! We need to add @chivee and @btrotta to the `CODEOWNERS` For `CMakeLists.txt`. 

image

I'll do that right now.

guolinke:
refer to https://docs.microsoft.com/en-us/cpp/build/reference/incremental-link-incrementally?view=vs-2019 

> An incrementally linked program is functionally equivalent to a program that is non-incrementally linked. However, because it is prepared for subsequent incremental links, an incrementally linked executable, static library, or dynamic-link library file:
> 
> Is larger than a non-incrementally linked program because of padding of code and data. Padding enables the linker to increase the size of functions and data without recreating the file.
> 
> May contain jump thunks to handle relocation of functions to new addresses.
> 
>  Note
> 
> To ensure that your final release build does not contain padding or thunks, link your program non-incrementally.

Could we only use the `incremental` for the tests, but not for release binaries and cmake ?

jameslamb:
> refer to https://docs.microsoft.com/en-us/cpp/build/reference/incremental-link-incrementally?view=vs-2019
> 
> > An incrementally linked program is functionally equivalent to a program that is non-incrementally linked. However, because it is prepared for subsequent incremental links, an incrementally linked executable, static library, or dynamic-link library file:
> > Is larger than a non-incrementally linked program because of padding of code and data. Padding enables the linker to increase the size of functions and data without recreating the file.
> > May contain jump thunks to handle relocation of functions to new addresses.
> > Note
> > To ensure that your final release build does not contain padding or thunks, link your program non-incrementally.
> 
> Could we only use the `incremental` for the (slow) tests, but not for release binaries and cmake ?

Thanks! When I read that doc, I missed this:

> To ensure that your final release build does not contain padding or thunks, link your program non-incrementally..

I found that the test speedup is not very much (around a minute for R builds), so I don't think introducing any `if (TEST)` complexity is worth it.

Based on that, I think we should just close this PR.

StrikerRUS:
@jameslamb 
> I found that the test speedup is not very much (around a minute for R builds), so I don't think introducing any if (TEST) complexity is worth it.
Based on that, I think we should just close this PR.

I'm +1 🙂 .

BTW, is a warning `msb8029` from https://github.com/microsoft/LightGBM/issues/3067#issue-615509984 about the same "incremental building"? Or it is different things?

jameslamb:
> @jameslamb
> 
> > I found that the test speedup is not very much (around a minute for R builds), so I don't think introducing any if (TEST) complexity is worth it.
> > Based on that, I think we should just close this PR.
> 
> I'm +1  .
> 
> BTW, is a warning `msb8029` from [#3067 (comment)](https://github.com/microsoft/LightGBM/issues/3067#issue-615509984) about the same "incremental building"? Or it is different things?

I don't think know if it's about the same thing, but it is definitely not caused by use of the `/incremental` flag. I see it on [the most recent AppVeyor build on master](https://ci.appveyor.com/project/guolinke/lightgbm/builds/32890583/job/n9e9f9x0j3hmjtdb) and we don't use that flag there.

image

I'm going to close this PR and document that warning in an issue where we can add links and comments.

StrikerRUS:
> I see it on the most recent AppVeyor build on master and we don't use that flag there.

Ah, I see! However, I think warning is not saying that user **is using incremental building right now**. It just indicate potential issues **if you will use it**.

jameslamb:
> > I see it on the most recent AppVeyor build on master and we don't use that flag there.
> 
> Ah, I see! However, I think warning is not saying that user **is using incremental building right now**. It just indicate potential issues **if you will use it**.

I agree with you! I think maybe Visual Studio has a way to be clever about only re-compiling things whose source files (or their dependencies) have changed within a project? Some of the answers I saw from search engines talked about that.

Anyway, I've opened #3088 for an investigation

PR #3077, opened 2020-05-13
fix goss with constant hessian

### LINKED ISSUE #3053: Under the goss mode, the l2 loss is suddenly increased during training

LightGBM component: **Python package**

## Environment info

Operating System: **Windows**

CPU/GPU model: **CPU**

Other:

LightGBM version or commit hash:   **2.3.1**

## The log info and loss curve is below：
log

## The parameters：
`model = LGBMRegressor(max_depth=10, learning_rate=0.01, n_estimators=1000, num_leaves=250,min_child_samples=200, colsample_bytree=0.2, boosting_type="goss", top_rate=0.95, other_rate=0.01, random_state=0, n_jobs=52)`

## data info:
the dataset is about 1300000 samples and 3500 features

## my questions

1. how did this happen?
In my understanding, the prediction is the sum of the leaves values which times learning_rate. And the leaf's value is calculated by 
w
The loss function I used is mse, so H is 2, G is the residual.

2. how can I prevent it 

## Similar case in Kaggle
https://www.kaggle.com/c/home-credit-default-risk/discussion/60921

### LINKED ISSUE #3074: segmentation fault encountered when 'boosting = goss'

## Environment info

Operating System: ubuntu-20.04-desktop-amd64

CPU/GPU model: CPU

C++ compiler version: 9.3.0

CMake version: 3.16.3

Java version: N/A

Python version: 3.8.2

R version: N/A

Other: N/A

LightGBM version or commit hash: 2.3.2

Segmentation fault (core dumped)

## Reproducible example(s)

from sklearn.datasets import fetch_california_housing
import lightgbm as lgbm

X, y = fetch_california_housing(return_X_y=True)
param = { 'boosting': 'goss' }
dataset = lgbm.Dataset(X, y)
model = lgbm.train(param, dataset)

## Steps to reproduce
I updated lightgbm on my machine today. However, when I set boosting to goss, segmentation fault was encountered. Setting boosting to gbdt worked well.

guolinke:
fix #3074 and fix #3053

### DIFF
--- a/src/boosting/gbdt.cpp
+++ b/src/boosting/gbdt.cpp
@@ -70,11 +70,10 @@ void GBDT::Init(const Config* config, const Dataset* train_data, const Objective
   objective_function_ = objective_function;
   num_tree_per_iteration_ = num_class_;
   if (objective_function_ != nullptr) {
-    is_constant_hessian_ = objective_function_->IsConstantHessian();
     num_tree_per_iteration_ = objective_function_->NumModelPerIteration();
-  } else {
-    is_constant_hessian_ = false;
   }
+
+  is_constant_hessian_ = GetIsConstHessian(objective_function);
 
   tree_learner_ = std::unique_ptr<TreeLearner>(TreeLearner::CreateTreeLearner(config_->tree_learner, config_->device_type, config_.get()));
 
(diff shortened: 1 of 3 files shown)


StrikerRUS:
@guolinke Should we add a regression test for this?

guolinke:
@StrikerRUS yeah, I think we should have tests for goss.

PR #3081, opened 2020-05-13
[docs] document one more workaround for CMake and MinGW build and fix typos

StrikerRUS:
> The MinGW Makefiles generator no longer issues an error if sh.exe is present in the environment’s PATH.
  https://cmake.org/cmake/help/v3.17/release/3.17.html#other-changes

Refer to #3078.


jameslamb:
Looks good, thanks!

PR #3090, opened 2020-05-16
redirect log to python console

guolinke:
for #1493


python-package/lightgbm/basic.py
```diff
@@ -21,13 +21,21 @@
 from .libpath import find_lib_path
 
 
+def _log_callback(msg):
+    """Redirect logs from native library into Python console"""
+    print("{0:s}".format(decode_string(msg)))
```
guolinke, line 26:
```suggestion
    print("{0:s}".format(decode_string(msg)), end='')
```

include/LightGBM/utils/log.h
```diff
@@ -113,39 +118,57 @@ class Log {
...
-        Rvprintf(format, val);
-        Rprintf("\n");
-      #endif
+      } else {
+        const size_t buf_size = 512;
+        char buf[buf_size];
+        snprintf(buf, buf_size, "[LightGBM] [%s] ", level_str);
+        GetLogCallBack()(buf);
+        vsnprintf(buf, buf_size, format, val);
+        GetLogCallBack()(buf);
```
guolinke, line 150:
```suggestion
        GetLogCallBack()(buf);
        GetLogCallBack()("\n");
```

guolinke:
@jameslamb it seems this redir method could be used in R. should we unify them?

jameslamb:
> @jameslamb it seems this redir method could be used in R. should we unify them?

@guolinke I believe `Rprintf` and `Rf_error` that we use from the R headers already redirect logs to the R console: https://github.com/microsoft/LightGBM/blob/0341906c45666ccaeca92ee5fbee17fa22282470/include/LightGBM/utils/log.h#L137-L140

There is a lot of magic in those functions so I'd rather leave them alone than have us maintain code that needs to understand how redirecting output to R works.

python-package/lightgbm/basic.py
```diff
@@ -21,13 +21,21 @@
 from .libpath import find_lib_path
 
 
+def _log_callback(msg):
+    """Redirect logs from native library into Python console"""
+    print("{0:s}".format(decode_string(msg)), end='')
```
guolinke, line 26: it seems python 2 don't support `end=''`. @StrikerRUS @henry0312  any better solution? or we drop the support of python 2?

StrikerRUS: Just add `print_function` to the 3rd line

```
from __future__ import absolute_import, print_function
```

jameslamb:
I missed an opportunity in https://github.com/microsoft/LightGBM/pull/3090#issuecomment-629591294 to say "one of the few places where the R package is ahead of the Python package" 😛

jameslamb:
We recently merged some fixes to CI (#3092). Sorry for the inconvenience, but if you merge the most recent `master` into this branch it should fix the issues with the R package builds on Travis.

python-package/lightgbm/basic.py
```diff
@@ -21,13 +21,21 @@
 from .libpath import find_lib_path
 
 
+def _log_callback(msg):
+    """Redirect logs from native library into Python console"""
```
StrikerRUS, line 25: To make pytest docstring checks happy. It doesn't raise error right now because by default it doesn't check private members.
```suggestion
    """Redirect logs from native library into Python console."""
```

python-package/lightgbm/basic.py
```diff
@@ -21,13 +21,21 @@
...
+    print("{0:s}".format(decode_string(msg)), end='')
+
 def _load_lib():
     """Load LightGBM library."""
     lib_path = find_lib_path()
     if len(lib_path) == 0:
         return None
     lib = ctypes.cdll.LoadLibrary(lib_path[0])
     lib.LGBM_GetLastError.restype = ctypes.c_char_p
+    CALLBACK = ctypes.CFUNCTYPE(None, ctypes.c_char_p)
```
StrikerRUS, line 35: I think ALL_CAPITAL (`CALLBACK`) naming schema is only for module-level variables.

python-package/lightgbm/basic.py
```diff
@@ -21,13 +21,21 @@
...
+    lib.callback = CALLBACK(_log_callback)
+    if lib.LGBM_RegisterLogCallback(lib.callback) != 0:
+        raise LightGBMError(decode_string(_LIB.LGBM_GetLastError()))
```
StrikerRUS, line 38: Seems that we cannot use `_LIB` here because it leads to circular reference. `lib.LGBM_GetLastError()`?

guolinke, line 38:
```suggestion
        raise LightGBMError(decode_string(lib.LGBM_GetLastError()))
```

python-package/lightgbm/basic.py
```diff
@@ -21,13 +21,21 @@
...
+
 def _load_lib():
     """Load LightGBM library."""
     lib_path = find_lib_path()
     if len(lib_path) == 0:
         return None
     lib = ctypes.cdll.LoadLibrary(lib_path[0])
     lib.LGBM_GetLastError.restype = ctypes.c_char_p
+    CALLBACK = ctypes.CFUNCTYPE(None, ctypes.c_char_p)
+    lib.callback = CALLBACK(_log_callback)
```
guolinke, line 36:
```suggestion
    callback= ctypes.CFUNCTYPE(None, ctypes.c_char_p)
    lib.callback = callback(_log_callback)
```

python-package/lightgbm/basic.py
```diff
@@ -21,13 +21,21 @@
 from .libpath import find_lib_path
 
 
+def _log_callback(msg):
+    """Redirect logs from native library into Python console."""
+    print("{0:s}".format(decode_string(msg)), end='')
+
```
guolinke, line 27:
```suggestion


```

python-package/lightgbm/basic.py
```diff
@@ -21,13 +21,21 @@
...
 def _load_lib():
     """Load LightGBM library."""
     lib_path = find_lib_path()
     if len(lib_path) == 0:
         return None
     lib = ctypes.cdll.LoadLibrary(lib_path[0])
     lib.LGBM_GetLastError.restype = ctypes.c_char_p
+    callback= ctypes.CFUNCTYPE(None, ctypes.c_char_p)
```
guolinke, line 35:
```suggestion
    callback = ctypes.CFUNCTYPE(None, ctypes.c_char_p)
```

include/LightGBM/utils/log.h
```diff
@@ -113,39 +118,58 @@ class Log {
...
-        Rprintf("[LightGBM] [%s] ", level_str);
-        Rvprintf(format, val);
-        Rprintf("\n");
-      #endif
+      } else {
+        const size_t buf_size = 512;
+        char buf[buf_size];
+        snprintf(buf, buf_size, "[LightGBM] [%s] ", level_str);
+        GetLogCallBack()(buf);
+        vsnprintf(buf, buf_size, format, val);
```
guolinke, line 149:
```suggestion
        const size_t kBufSize = 512;
        char buf[kBufSize];
        snprintf(buf, kBufSize, "[LightGBM] [%s] ", level_str);
        GetLogCallBack()(buf);
        vsnprintf(buf, kBufSize, format, val);
```

StrikerRUS:
Just rebased to `master` with the aim to pass all CI tests and removed one word from param description.

PR #3092, opened 2020-05-16
[R-package] [ci] Fix CI timeouts for Mac builds on Travis (fixes #3091)

### LINKED ISSUE #3091: [R-package] [ci] MacOS Travis builds timing out
See the conversation starting at https://github.com/microsoft/LightGBM/pull/3086#issuecomment-629481888. Currently, MacOS `r-package` builds on Travis are failing with an error like this:

> * checking whether package ‘lightgbm’ can be installed ...

> No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

In this case, "no output" means "nothing written to the log".

It seems that our build times for the R package on Mac were already very close to 10 minutes but just under this. It's possible the issue is caused by one or some combination of these:

* maybe installation of the R package got slightly slower
* maybe something about the Travis Mac environment is slowing us down (for example, slower disk I/O)
* maybe changing `print()` (which prints to `stdout`) to `message()` (which prints to stderr) in #2965 is extending the time between logs to stdout

Some possible fixes for this, found by @StrikerRUS 

* https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received
* travis-ci/travis-ci#4190 (comment)

jameslamb:
See #3091 for details. This PR fixes the recent issues with R Mac builds timing out on Travis.

It's not clear why these builds, which used to take around 10 minutes total, now take closer to 19 minutes

I'm a little worried that recent R PRs introduced a Mac installation performance problem, but I've also found (across many projects, not just LightGBM), that the Mac environments on Travis are not as reliable as the Linux environments.

Either way, that investigation can come later...this PR should unblock us.

### DIFF
--- a/.ci/test_r_package.sh
+++ b/.ci/test_r_package.sh
@@ -93,7 +93,7 @@ check_succeeded="yes"
 #     * https://github.com/travis-ci/travis-ci/issues/4190#issuecomment-169987525
 #     * https://stackoverflow.com/a/29890106/3986677
 CHECK_PID=$!
-while kill -0 ${CHECK_PID}; do
+while kill -0 ${CHECK_PID} >/dev/null 2>&1; do
     echo -n -e " \b"
     sleep 5
 done
--- a/.ci/test_r_package.sh
+++ b/.ci/test_r_package.sh
@@ -64,11 +64,9 @@ conda install \
 
 # Manually install Depends and Imports libraries + 'testthat'
 # to avoid a CI-time dependency on devtools (for devtools::install_deps())
-packages="'data.table', 'jsonlite', 'Matrix', 'R6', 'testthat'"
+packages="c('data.table', 'jsonlite', 'Matrix', 'R6', 'testthat')"
 if [[ $OS_NAME == "macos" ]]; then
-    packages="c($packages, 'processx'), type = 'binary'"
-else
-    packages="c($packages)"
+    packages+=", type = 'binary'"
 fi
 Rscript --vanilla -e "install.packages(${packages}, repos = '${CRAN_MIRROR}', lib = '${R_LIB_PATH}', dependencies = c('Depends', 'Imports', 'LinkingTo'))" || exit -1
 
--- a/.ci/test_r_package.sh
+++ b/.ci/test_r_package.sh
@@ -64,9 +64,11 @@ conda install \
 
 # Manually install Depends and Imports libraries + 'testthat'
 # to avoid a CI-time dependency on devtools (for devtools::install_deps())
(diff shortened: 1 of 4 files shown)


jameslamb:
Tested on my fork, the Mac build takes almost 19 minutes but doesn't time out: https://travis-ci.com/github/jameslamb/LightGBM/jobs/335519105

.ci/test_r_package.sh
```diff
@@ -82,9 +82,21 @@ export _R_CHECK_FORCE_SUGGESTS_=0
...
+        --as-cran \
+    || check_succeeded="no"
+) &
+
+# R CMD check suppresses output, some CIs kill builds after
+# a few minutes with no output. This trick gives R CMD check more time
+#     * https://github.com/travis-ci/travis-ci/issues/4190#issuecomment-169987525
+#     * https://stackoverflow.com/a/29890106/3986677
+CHECK_PID=$!
+while kill -0 ${CHECK_PID}; do
```
StrikerRUS, line 96: Missed part from the original workaround to remove unneeded error from logs:

```
/Users/travis/build/microsoft/LightGBM/.ci/test_r_package.sh: line 96: kill: (12865) - No such process
```

```suggestion
while kill -0 ${CHECK_PID} >/dev/null 2>&1; do
```

.ci/test_r_package.sh
```diff
@@ -82,9 +82,21 @@ export _R_CHECK_FORCE_SUGGESTS_=0
...
+    echo -n -e " \b"
+    sleep 5
```
StrikerRUS, line 98:
```suggestion
    sleep 60
```

Aren't 5 seconds too frequent? With 10 minutes limit 1 minute signal seems to be quite enough.

.ci/test_r_package.sh
jameslamb: good call!

.ci/test_r_package.sh
jameslamb: The polling time is the maximum amount of time the build will spend doing nothing once  `R CMD CHECK` completes.

If `R CMD CHECK` finishes running one second after a call to `sleep 60`, that is `59` seconds where the build will just be sitting doing nothing. I think 1 minute of added time is meaningful on the scale of our build times, but 5 seconds is not.

StrikerRUS: Ah, got it! Agree with you.

jameslamb: thanks! I'll merge this as soon as it builds.

Really good find with https://github.com/travis-ci/travis-ci/issues/4190 by the way, I  never would have thought of that but it's a trick I'll be able to use in other projects in the future

Issue #3075, opened 2020-05-12
[R-package] Remove lgb.prepare() and lgb.prepare_rules()


jameslamb:
The R package currently has 4 functions for converting non-numeric inputs to numeric values for use with LightGBM training:

* `lgb.prepare()` ([docs](https://lightgbm.readthedocs.io/en/latest/R/reference/lgb.prepare.html))
* `lgb.prepare2()` ([docs](https://lightgbm.readthedocs.io/en/latest/R/reference/lgb.prepare2.html))

* `lgb.prepare_rules()` ([docs](https://lightgbm.readthedocs.io/en/latest/R/reference/lgb.prepare_rules.html))
* `lgb.prepare_rules2()` ([docs](https://lightgbm.readthedocs.io/en/latest/R/reference/lgb.prepare_rules2.html))

The `*2()` versions of these functions accomplish everything that their predecessors do, and in a better way. I think that `lgb.prepare()` and `lgb.prepare_rules()` should be deprecated in the next 2.x release, and that they should be removed completely in 3.0.0.

This will simplify the interface of the R package for users and reduce the maintenance burden for maintainers.

@Laurae2 @guolinke what do you think?

jameslamb:
Just put this in #2302 . Will close it once I hear from the maintainers tagged in the description about whether this is a good idea.

Laurae2:
@jameslamb as LightGBM does not crash anymore when using an R integer matrix, we can deprecate the non `*2()` versions.

jameslamb:
@Laurae2 ok thanks! I'll open a PR with a deprecation warning, and then when we do 3.0.0 (#3071 ) we can remove them

jameslamb:
Closing for now since we've agreed on direction. This issue is in #2302 with other features.

Issue #3094, opened 2020-05-16
[R-package] LightGBM results change between int/num


Laurae2:
## How you are using LightGBM?

LightGBM component: R-package

## Environment info

(...)

Other: applies to any R version / compiler combination

LightGBM version or commit hash: 0e3509c

## Error message and / or logs

Training a model with integer labels seem to provide wrong results and/or change LightGBM behavior.

Training matrix :
* row 1: variables: 0, 1 => label 1
* row 2: variables: 1, 0 => label 0

Expected prediction (should predict "1 0"):
* row 1: predict 1
* row 2: predict 0

Results:

| Matrix Type / Labels Type | Integer | Numeric |
| --- | :--- | :--- |
| Integer | **0 0 (KO)** | 0 1 (KO) |
| Numeric | **0 0 (KO)** | 0 1 (OK) |

Increased the number of iterations but no better results.

Changing labels to (1, 2) instead of (0, 1) leads to:

* OK case: predicting (1, 2) instead of (0, 1) (still correct)
* KO case: predicting (1, 1) instead of (0, 1) (still wrong)

Full logs:

```r
> # Data Int / Labels Int
> train_mat <- matrix(c(0L, 1L, 1L, 0L), nrow = 2, ncol = 2)
> train_labels <- c(0L, 1L)
> dtrain <- lgb.Dataset(train_mat, label = train_labels)
> model <- lgb.train(
+   params = list(objective = "regression", metric = "l2")
+   , data = dtrain
+   , nrounds = 1L
+   , min_data = 1L
+   , learning_rate = 1.0
+   , verbose = -1
+ )
> round(predict(model, train_mat), digits = 10)
[1] 0 0
> round(sum(abs(predict(model, train_mat) - train_labels)), digits = 10)
[1] 1
> 
> # Data Num / Labels Int
> train_mat <- matrix(c(0, 1, 1, 0), nrow = 2, ncol = 2)
> train_labels <- c(0L, 1L)
> dtrain <- lgb.Dataset(train_mat, label = train_labels)
> model <- lgb.train(
+   params = list(objective = "regression", metric = "l2")
+   , data = dtrain
+   , nrounds = 1L
+   , min_data = 1L
+   , learning_rate = 1.0
+   , verbose = -1
+ )
> round(predict(model, train_mat), digits = 10)
[1] 0 1
> round(sum(abs(predict(model, train_mat) - train_labels)), digits = 10)
[1] 0
> 
> # Data Int / Labels Num
[...]

PR #3095, opened 2020-05-16
[R-package] deprecate lgb.prepare() and lgb.prepare2()

jameslamb:
Part of #3075 . `lgb.prepare()` and `lgb.prepare2()` have a lot of code overlap and it could be confusing to know which one to use.

To reduce the public API of the R package and make it easier for uses  to  understand how too get from their raw data to a trained model, this PR deprecates `lgb.prepare()` and `lgb.prepare_rules()` in favor of`lgb.prepare2()` and  `lgb.prepare_rules2()` . They'll be removed completely in LightGBM  3.0.0 (#3071 )


jameslamb:
gently ping @Laurae2 for a review

R-package/R/lgb.prepare_rules.R
```diff
@@ -50,6 +53,12 @@
 #' @export
 lgb.prepare_rules <- function(data, rules = NULL) {
 
+  .Deprecated(
+    new = "lgb.prepare2"
+    , package = "lightgbm"
+    , msg = "lgb.prepare is deprecated and will be removed in LightGBM 3.0.0. lgb.prepare2() is preferred."
```
Laurae2, line 59: `msg` should read: `lgb.prepare_rules is deprecated and will be removed in LightGBM 3.0.0. lgb.prepare_rules2() is preferred.` (change accordingly the other field).

StrikerRUS: Maybe we should also rename `prepare2` into `prepare` with 3.0 release? I believe `2` without `1` makes no sense.

jameslamb: in my opinion it is not worth breaking existing code just for name consistency

guolinke: I think the name should not be changed. However, maybe we can set an alias for the name consistency, and remove the old one after the next release.

guolinke: @jameslamb does `prepare2` do the same thing as `prepare`? If yes, I agree with @StrikerRUS .

guolinke:
@jameslamb please check my comments inline.

R-package/R/lgb.prepare_rules.R
jameslamb: They don't do the exact same thing.

`lgb.prepare()` converts columns that are `character` or `factor` to `numeric`: https://lightgbm.readthedocs.io/en/latest/R/reference/lgb.prepare.html

`lgb.prepare2()` converts columns that are `character` or `factor` to `integer`: https://lightgbm.readthedocs.io/en/latest/R/reference/lgb.prepare2.html

Since LightGBM can accept integer input, I think we should only support the one that produces integers, since it is a more efficient representation.

Both functions achieve the same goal: take an R data frame that is not suitable as a training dataset and convert it to one that is.

This is why I think we should remove `lgb.prepare()`, but not change the names. Or, if you are worried about the `2`, we should pick a new and completely different name, like `lgb.convert()` or something. I believe that breaking changes in a new major version should be loud.

StrikerRUS: I'm +1 for deprecating both versions and the completely new name in 3.0.

jameslamb: when I opened this PR in May, the intention was to truly "deprecate" these, meaning "raise a warning saying they'll be removed in 3.0.0".

At this point, I think it's too late for that. I think we should just remove them and make that one of the breaking changes in the 3.0.0 release. 

So specifically:

1. Remove `lgb.prepare()` and `lgb.prepare_rules()`
2. Rename `lgb.prepare2()` to `lgb.convert()`
3. Rename `lgb.prepare_rules2()` to `lgb.convert_with_rules()`

@guolinke what do you think?

guolinke: @jameslamb looks good to me.

jameslamb: ok I will change this PR to do that then, thank you!

jameslamb: ok this has been changed! It's ready for another review

jameslamb:
I just merged `master`  into this to get the latest changes

jameslamb:
I've merged in the changes from #3188 . This is ready for review.

R-package/R/removed.R
```diff
@@ -0,0 +1,36 @@
+
```
StrikerRUS, line 1: Blank line.

R-package/demo/categorical_features_rules.R
```diff
@@ -32,7 +32,7 @@ bank_train <- bank[1L:4000L, ]
 bank_test <- bank[4001L:4521L, ]
 
 # We must now transform the data to fit in LightGBM
-# For this task, we use lgb.prepare
+# For this task, we use lgb.convert
```
StrikerRUS, line 35:
```suggestion
# For this task, we use lgb.convert_with_rules
```

R-package/tests/testthat/test_lgb.convert_with_rules.R
```diff
@@ -111,13 +111,13 @@ test_that("lgb.prepare_rules2() should work correctly for a dataset with numeric
         expect_identical(rules[["character_col"]], c("a" = 1L, "b" = 2L, "c" = 3L))
         expect_identical(rules[["factor_col"]], c("n" = 1L, "y" = 2L))
 
-        # today, lgb.prepare2() does  not convert numeric  columns
+        # today, lgb.convert() does  not convert numeric  columns
```
StrikerRUS, line 114:
```suggestion
        # today, lgb.convert() does not convert numeric columns
```

StrikerRUS, line 114: `convert_with_rules`?

R-package/tests/testthat/test_lgb.convert_with_rules.R
```diff
@@ -151,10 +151,10 @@ test_that("lgb.prepare_rules2() should work correctly for a dataset with missing
             expect_identical(converted_dataset[[col]], rep(NA_integer_, nrow(converted_dataset)))
         }
 
-        # today, lgb.prepare2() does not convert logical columns
+        # today, lgb.convert() does not convert logical columns
```
StrikerRUS, line 154: `convert_with_rules`?

R-package/tests/testthat/test_lgb.convert_with_rules.R
```diff
@@ -151,10 +151,10 @@ test_that("lgb.prepare_rules2() should work correctly for a dataset with missing
...
         expect_identical(class(converted_dataset[["na_col"]]), "logical")
 
-        # today, lgb.prepare2() does not convert numeric columns to integer
+        # today, lgb.convert() does not convert numeric columns to integer
```
StrikerRUS, line 157: `convert_with_rules`?

jameslamb:
```suggestion
        # today, lgb.convert_with_rules() does not convert numeric columns to integer
```

R-package/tests/testthat/test_lgb.convert_with_rules.R
jameslamb:
```suggestion
        # today, lgb.convert_with_rules() does not convert logical columns
```

R-package/tests/testthat/test_lgb.convert_with_rules.R
jameslamb:
```suggestion
        # today, lgb.convert_with_rules() does  not convert numeric  columns
```

R-package/R/removed.R
jameslamb:
```suggestion
```

R-package/tests/testthat/test_lgb.convert_with_rules.R
jameslamb: weirdly GitHub wouldn't let me commit this suggestion, not sure why. So I just pushed a commit: https://github.com/microsoft/LightGBM/pull/3095/commits/0ac6a59189faa327223ed5601ae5dc1fa29bf525

R-package/R/lgb.convert.R
```diff
@@ -1,11 +1,13 @@
...
+#' @name lgb.convert
 #' @title Data preparator for LightGBM datasets (integer)
 #' @description Attempts to prepare a clean dataset to prepare to put in a \code{lgb.Dataset}.
-#'              Factors and characters are converted to numeric (specifically: integer).
-#'              Please use \code{\link{lgb.prepare_rules2}} if you want to apply this transformation to
+#'              Factors and characters are converted to integer.
+#'              Please use \code{\link{lgb.convert_with_rules}} if you want to apply this transformation to
 #'              other datasets. This is useful if you have a specific need for integer dataset instead
 #'              of numeric dataset. Note that there are programs which do not support integer-only
 #'              input. Consider this as a half memory technique which is dangerous, especially for LightGBM.
```
Laurae2, line 8: Remove these two lines as it's obsolete since https://github.com/microsoft/LightGBM/issues/3094 is fixed:

```
Note that there are programs which do not support integer-only
input. Consider this as a half memory technique which is dangerous, especially for LightGBM.
```

R-package/tests/testthat/test_lgb.convert.R
```diff
@@ -75,20 +75,20 @@ test_that("lgb.prepare2() should work correctly for a dataset with numeric, fact
...
-        converted_dataset <- lgb.prepare2(input_data)
+        converted_dataset <- lgb.convert(input_data)
         expect_identical(class(input_data), class(converted_dataset))
         expect_identical(class(converted_dataset[["character_col"]]), "integer")
         expect_identical(class(converted_dataset[["factor_col"]]), "integer")
         expect_identical(converted_dataset[["character_col"]], c(1L, 2L, 3L))
         expect_identical(converted_dataset[["factor_col"]], c(1L, 1L, 2L))
 
-        # today, lgb.prepare2() does  not convert numeric  columns
+        # today, lgb.convert() does  not convert numeric  columns
```
Laurae2, line 85: one whitespace instead of two whitespaces

R-package/R/lgb.convert.R
jameslamb: oh yeah!! Thank you

R-package/tests/testthat/test_lgb.convert.R
jameslamb: these have been fixed by accepting previous suggestions in the browser

R-package/R/lgb.convert.R
jameslamb: updated! https://github.com/microsoft/LightGBM/pull/3095/commits/d023d94763582212546574526606e49cd902efa9

jameslamb:
@StrikerRUS have you also seen this recently on AppVeyor?

image

StrikerRUS:
> have you also seen this recently on AppVeyor?

Nope...

jameslamb:
> > have you also seen this recently on AppVeyor?
> 
> Nope...

ok, hopefully it was just something transient. I don't see any incidents on AppVeyor: https://status.appveyor.com/

PR #3096, opened 2020-05-16
[ci] avoid duplicate AppVeyor builds

jameslamb:
Currently, we run the same build logic twice on AppVeyor for every commit to `master` or a branch with a PR into `master`.

This slows development because on the free tier of AppVeyor we are limited to one build at a time.

In this PR, I propose that we only build once on PRs and on every commit to `master`. I think this should work per: https://help.appveyor.com/discussions/problems/14586-build-any-pull-requests-against-any-branches.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -5,6 +5,12 @@ platform: x64
 configuration:  # a trick to construct a build matrix with multiple Python versions
   - 3.7
 
+# only build pull requests and
+# commits to 'master'
+branches:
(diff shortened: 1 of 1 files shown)


jameslamb:
so far so good...only the `/pr` build for AppVeyor was triggered on this PR!

image

jameslamb:
Can confirm that this worked! 

I just merged and we got a `master` branch build generated on AppVeyor:

image

PR #3098, opened 2020-05-17
[R-package] [docs] replace build_r_site.R with pkgdown::build_site()

jameslamb:
Per this conversation (https://github.com/microsoft/LightGBM/issues/1143#issuecomment-629472613), this PR removes `build_r_site.R` and replaces it with a call to `pkgdown::build_site(install = FALSE)`.

Documentation tools like `roxygen2::roxygenize()`, `pkgdown::build_site()`, and `devtools::document()` used to try to rebuild R packages and this work well with ours. But now these tools support installing once and then running their functionality without re-building. So this flow works and only builds the package one time:

```shell
Rscript build_r.R

Rscript -e "roxygen2::royxgenize(load = 'installed')"

Rscript -e "pkdown::build_site()"
```

I tried building this on readthedocs and it looks ok!

* build: https://readthedocs.org/projects/lightgbm/builds/11055947/
* docs: https://lightgbm.readthedocs.io/en/fix-remove-devtools/R/index.html


docs/conf.py
```diff
@@ -240,8 +240,20 @@ def generate_r_docs(app):
     cd {0}
     export R_LIBS="$CONDA_PREFIX/lib/R/library"
     Rscript build_r.R
-    Rscript build_r_site.R
-    """.format(os.path.join(CURR_PATH, os.path.pardir))
+    cd {1}
+    Rscript -e "roxygen2::roxygenize(load = 'installed')"
+    Rscript -e "pkgdown::build_site( \
+            lazy = FALSE \
```
StrikerRUS, line 246: Should we explicitly pin `new_process` argument here?

build_r_site.R
```diff
@@ -1,20 +0,0 @@
-library(pkgdown)
-library(roxygen2)
-
-setwd("lightgbm_r")
-if (!dir.exists("docs")) {
-  dir.create("docs")
-}
```
StrikerRUS, line 7: Why can we drop this?

StrikerRUS:
I think `doc` PR label is for something that is changing any documentation for **users**. This PR leaves everything visible for users the same. So I believe `maintaince` label should be used for invisible at user side changes.

build_r_site.R
jameslamb: `pkgdown::build_site()`  creates a `doc/` directory automatically: https://github.com/r-lib/pkgdown/blob/aea6864792d681f64a9e1b8c52fe24668b2e4f55/R/build.r#L23-L24

docs/conf.py
jameslamb: yeah probably a good idea. I will add that.

jameslamb:
After pinning `new_process`  in https://github.com/microsoft/LightGBM/pull/3098/commits/c428cb09328a640246419b6077221a088fc85958, can confirm this is still working:

* rtd build: https://readthedocs.org/projects/lightgbm/builds/11059218/
* https://lightgbm.readthedocs.io/en/fix-remove-devtools/R/index.html

I'll merge this if CI build successfully

jameslamb:
@StrikerRUS now that this is merged, you can disable the `fix/remove-devtools` branch on readthedocs and I'll remove it from the repo here. Thanks for the help!

StrikerRUS:
> `pkgdown::build_site()`  creates a `docs` directory automatically:

Ah, makes perfect sense! Completely missed it.

> now that this is merged, you can disable the `fix/remove-devtools` branch on readthedocs and I'll remove it from the repo here.

OK. Done.

Thank you very much for this great simplification!

PR #3101, opened 2020-05-19
[R-package] fix issue where early stopping thinks higher MAPE is desirable (fixes #3099)

### LINKED ISSUE #3099: [R-package] MAPE metric not being minimized in R
Hi, 

I am running Lightgbm in R (R version 4.0.0, Lightgbm version 2.3.2), with metric = "mape". I see that when I run cross validation (lgb.cv), the best iteration is always the one with the maximum value of mape.

For example, I run cross validation with nrounds = 200. Last value:

[200]:  valid's mape:0.47669+0.033924

best.iter <- lgb.cv$best_iter

<lgb.CVBooster>
  Public:
    best_iter: 200
    best_score: 0.476690009429843
    boosters: list
    initialize: function (x) 
    record_evals: list
    reset_parameter: function (new_params) 

However, if I see other values, 

[151]:  valid's mape:0.436838+0.0311321

In this case, the best iteration should be 151 and not 200.

I do not see this behavior in python.

Am I doing something wrong?

Thank you.

jameslamb:
## Background

This PR fixes #3099. Several metrics are currently being misclassified for early stopping purposes.

Today, the R package uses a regular expression on metric name to decide if a metric should be maximized or minimized.

That logic says:

> if a metric name starts with `"ndcg"`, `"map"`, or `"auc"`, assume it should be maximized. Otherwise, assume it should be minimized

~The author of #3099 ran into this issue with `mape`, but I believe that several of the [support edmetrics](https://lightgbm.readthedocs.io/en/latest/Parameters.html#metric-parameters) are currently misclassified:~

* ~`lambdarank`~
* ~all `xendcg` aliases and flavors`~
* ~`mean_average_precision` (alias for `map`)~
* `auc_mu`

UPDATE: ^ this comment about all those metrics being misclassified is incorrect, see https://github.com/microsoft/LightGBM/pull/3101#issuecomment-631065628.

## Solution

My proposed solution is to remove the regular expression and hard-code in the list of supported metrics and whether they should be minimized or maximized.

I think this is more reliable than a regular expression and should prevent similar issues in the future, but it does introduce one more file that has to be touched when we have contributions like #2567 that introduce new metrics.

Reviewers, please look carefully at the list in `metrics.R`. I typed this in manually so there was a lot of opportunity for mistakes.

### DIFF
--- a/R-package/R/metrics.R
+++ b/R-package/R/metrics.R
@@ -26,7 +26,6 @@
         , "binary_error" = FALSE
         , "auc_mu" = TRUE
         , "multi_logloss" = FALSE
-        , "multiclass" = FALSE
         , "multi_error" = FALSE
         , "cross_entropy" = FALSE
         , "cross_entropy_lambda" = FALSE
--- a/R-package/R/metrics.R
+++ b/R-package/R/metrics.R
@@ -3,7 +3,7 @@
 #
 # [return] A named logical vector, where each key is a metric name and each value is a boolean.
 #          TRUE if higher values of the metric are desirable, FALSE if lower values are desirable.
-.METRICS_HIGHER_BETTER <- function(){
+.METRICS_HIGHER_BETTER <- function() {
     return(c(
         "l1" = FALSE
         , "l2" = FALSE
--- a/R-package/R/metrics.R
+++ b/R-package/R/metrics.R
@@ -1,0 +1,56 @@
+# [description] List of metrics known to LightGBM. The most up to date list can be found
+#               at https://lightgbm.readthedocs.io/en/latest/Parameters.html#metric-parameters
+#
+# [return] A named logical vector, where each key is a metric name and each value is a boolean.
+#          TRUE if higher values of the metric are desirable, FALSE if lower values are desirable.
+.METRICS_HIGHER_BETTER <- function(){
+    return(c(
+        "l1" = FALSE
+        , "l2" = FALSE
+        , "mae" = FALSE
+        , "mape" = FALSE
+        , "mean_absolute_percentage_error" = FALSE
+        , "mse" = FALSE
+        , "mean_squared_error" = FALSE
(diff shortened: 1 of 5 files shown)


StrikerRUS:
@jameslamb I don't think you should store all aliases. cpp code always returns a main name, not aliases.

jameslamb:
> @jameslamb I don't think you should store all aliases. cpp code always returns a main name, not aliases.

:grimacing: ok I can remove the aliases

jameslamb:
Removed the metric aliases in https://github.com/microsoft/LightGBM/pull/3101/commits/58a66df1eff5aa3f429c97aaebf44e2fa844f146

That commit also fixes an issue that was causing the tests to fail, and probably reason we were using a regular expression in the first place...when you use `ndcg_at` to specify the number of query results to use in evaluating `ndcg` for a ranking problem, the metric names that come back are `ndcg@1, ndcg@2,...`, so they won't match `"ndcg"` in the list of metric names.

R-package/R/metrics.R
```diff
@@ -0,0 +1,35 @@
...
+        , "gamma_deviance" = FALSE
+        , "tweedie" = FALSE
+        , "ndcg" = TRUE
+        , "map" = TRUE
+        , "auc" = TRUE
+        , "binary_logloss" = FALSE
+        , "binary_error" = FALSE
+        , "auc_mu" = TRUE
+        , "multi_logloss" = FALSE
+        , "multiclass" = FALSE
```
StrikerRUS, line 29: `multiclass` is alias for `multi_logloss`.

I can confirm that all metric names are listed here. But I don't think that I should review this PR as these changes are all about R-package.

jameslamb: oh good eye! I'll remove it

jameslamb: removed, thanks!

Issue #2981, opened 2020-04-07
[mmlspark] seeing memory corruption error on LGBM_BoosterFree with latest version


imatiach-msft:
when trying to upgrade lightgbm to latest version on master I am seeing a random memory corruption error in LGBM_BoosterFree inside SerialTreeLearner, pasting stack trace below:

[LightGBM] [Info] Finished linking network in 0.006790 seconds

*** Error in `java': double free or corruption (!prev): 0x00007fb0ac9342a0 ***

======= Backtrace: =========

/lib/x86_64-linux-gnu/libc.so.6(+0x777e5)[0x7fb0c91467e5]

/lib/x86_64-linux-gnu/libc.so.6(+0x8037a)[0x7fb0c914f37a]

/lib/x86_64-linux-gnu/libc.so.6(cfree+0x4c)[0x7fb0c915353c]

/tmp/mml-natives484132555220552210/lib_lightgbm.so(_ZN8LightGBM23DataParallelTreeLearnerINS_17SerialTreeLearnerEED2Ev+0x78)[0x7fb05c3915f8]

/tmp/mml-natives484132555220552210/lib_lightgbm.so(_ZN8LightGBM23DataParallelTreeLearnerINS_17SerialTreeLearnerEED0Ev+0x9)[0x7fb05c391639]

/tmp/mml-natives484132555220552210/lib_lightgbm.so(_ZN8LightGBM4GBDTD1Ev+0xa30)[0x7fb05c1c8c50]

/tmp/mml-natives484132555220552210/lib_lightgbm.so(_ZN8LightGBM4GBDTD0Ev+0x9)[0x7fb05c1c9219]

/tmp/mml-natives484132555220552210/lib_lightgbm.so(_ZN8LightGBM7BoosterD2Ev+0x5f9)[0x7fb05c4401c9]

/tmp/mml-natives484132555220552210/lib_lightgbm.so(LGBM_BoosterFree+0xe)[0x7fb05c42ec1e]

guolinke:
@imatiach-msft it that possible to locate which function/code block causes the memory corruption?

imatiach-msft:
I'm currently trying to locate the function/code block.  Removing the call to LGBM_BoosterFree, StringArrayHandle_free and LGBM_DatasetFree still seems to result in segfault or sigabort errors in other places, and they seem to vary from run to run.  Using a single partition seems to fix all issues, so I suspect it's something that only happens for distributed case.

[...]

PR #3110, opened 2020-05-22
memory corruption fix for distributed data parallel version before SyncUpGlobalBestSplit

imatiach-msft:
memory corruption was found using valgrind

see relevant issue for debugging info:
https://github.com/microsoft/LightGBM/issues/2981

### DIFF
--- a/src/treelearner/data_parallel_tree_learner.cpp
+++ b/src/treelearner/data_parallel_tree_learner.cpp
@@ -29,7 +29,14 @@ void DataParallelTreeLearner<TREELEARNER_T>::Init(const Dataset* train_data, boo
   // allocate buffer for communication
   size_t buffer_size = this->train_data_->NumTotalBin() * kHistEntrySize;
 
-  input_buffer_.resize(buffer_size);
+  auto max_cat_threshold = this->config_->max_cat_threshold;
+  // input_buffer_ needs to be able to hold smaller and larger best splits in SyncUpGlobalBestSplit
+  int splitInfoSize = SplitInfo::Size(max_cat_threshold) * 2;
+  if (buffer_size < static_cast<size_t>(splitInfoSize)) {
+    input_buffer_.resize(splitInfoSize);
+  } else {
+    input_buffer_.resize(buffer_size);
+  }
   output_buffer_.resize(buffer_size);
 
   is_feature_aggregated_.resize(this->num_features_);
@@ -231,11 +238,6 @@ void DataParallelTreeLearner<TREELEARNER_T>::FindBestSplitsFromHistograms(const 
   }
 
   // sync global best info
-  auto max_cat_threshold = this->config_->max_cat_threshold;
-  int splitInfoSize = SplitInfo::Size(max_cat_threshold);
-  if (input_buffer_.size() < splitInfoSize * 2) {
-    input_buffer_.resize(splitInfoSize * 2);
-  }
   SyncUpGlobalBestSplit(input_buffer_.data(), input_buffer_.data(), &smaller_best_split, &larger_best_split, this->config_->max_cat_threshold);
 
   // set best split
--- a/src/treelearner/data_parallel_tree_learner.cpp
+++ b/src/treelearner/data_parallel_tree_learner.cpp
@@ -26,17 +26,15 @@ void DataParallelTreeLearner<TREELEARNER_T>::Init(const Dataset* train_data, boo
   // Get local rank and global machine size
   rank_ = Network::rank();
   num_machines_ = Network::num_machines();
-  // allocate buffer for communication
-  size_t buffer_size = this->train_data_->NumTotalBin() * kHistEntrySize;
 
   auto max_cat_threshold = this->config_->max_cat_threshold;
-  // input_buffer_ needs to be able to hold smaller and larger best splits in SyncUpGlobalBestSplit
-  int splitInfoSize = SplitInfo::Size(max_cat_threshold) * 2;
-  if (buffer_size < static_cast<size_t>(splitInfoSize)) {
-    input_buffer_.resize(splitInfoSize);
-  } else {
-    input_buffer_.resize(buffer_size);
-  }
+  // need to be able to hold smaller and larger best splits in SyncUpGlobalBestSplit
+  size_t split_info_size = static_cast<size_t>(SplitInfo::Size(max_cat_threshold) * 2);
+
+  // allocate buffer for communication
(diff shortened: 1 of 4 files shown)


StrikerRUS:
Close-reopen to retrigger Appveyor.

src/treelearner/data_parallel_tree_learner.cpp
```diff
@@ -231,6 +231,11 @@ void DataParallelTreeLearner<TREELEARNER_T>::FindBestSplitsFromHistograms(const
   }
 
   // sync global best info
+  auto max_cat_threshold = this->config_->max_cat_threshold;
+  int splitInfoSize = SplitInfo::Size(max_cat_threshold);
+  if (input_buffer_.size() < splitInfoSize * 2) {
+    input_buffer_.resize(splitInfoSize * 2);
+  }
```
guolinke, line 238: maybe we can do this in Init function?

src/treelearner/split_info.hpp
```diff
@@ -49,7 +49,7 @@ struct SplitInfo {
   bool default_left = true;
   int8_t monotone_type = 0;
   inline static int Size(int max_cat_threshold) {
-    return 2 * sizeof(int) + sizeof(uint32_t) + sizeof(bool) + sizeof(double) * 9 + sizeof(data_size_t) * 2 + max_cat_threshold * sizeof(uint32_t) + sizeof(int8_t);
+    return 2 * sizeof(int) + sizeof(uint32_t) + sizeof(bool) + sizeof(double) * 7 + sizeof(data_size_t) * 2 + max_cat_threshold * sizeof(uint32_t) + sizeof(int8_t);
```
imatiach-msft, line 52: minor bugfix -- I noticed min_constraint/max_constraint were removed, but the size calculation was not updated:

https://github.com/microsoft/LightGBM/pull/1314/files

note this doesn't fix any bug, I just noticed this by chance and updated

src/treelearner/data_parallel_tree_learner.cpp
imatiach-msft: done!  is input_buffer_ used just as a temporary variable throughout the class?  I don't have a good grasp of the code so I'm not sure if I'm actually making the right fix here.

src/treelearner/split_info.hpp
imatiach-msft: it looks like this is the PR that removed them but didn't update the size calculation:
https://github.com/microsoft/LightGBM/commit/3670e476b1918e72e72ccf00bca69d398827e24f#diff-85924b6a07e712436ab5e7782288e2a7

src/treelearner/data_parallel_tree_learner.cpp
guolinke: yeah, I just want to minimize the `if ... else' in runtime. 
BTW, should we fix the corresponding code in feature/voting parallel?

imatiach-msft:
```
BTW, should we fix the corresponding code in feature/voting parallel?
```
good find, I wasn't aware that was an issue there too

imatiach-msft: done, updated voting parallel version.  It looks like feature parallel already had this logic, but I changed it to use Size method on SplitInfo instead of doing a custom calculation.  Also I modified the data parallel calculation to use std::max to be more in line with the style in the voting parallel class for calculating the input_buffer memory.

imatiach-msft:
Close-reopen to retrigger travis URL checker

imatiach-msft:
the PR previously failed on a Mac build that complained about different types passed to std::max, I added a fix for it (cast) that seems to have resolved the build issue

imatiach-msft:
@guolinke might you be able to take another look at this PR?  thank you!

src/treelearner/voting_parallel_tree_learner.cpp
```diff
@@ -37,6 +37,10 @@ void VotingParallelTreeLearner<TREELEARNER_T>::Init(const Dataset* train_data, b
   }
   // calculate buffer size
   size_t buffer_size = 2 * top_k_ * std::max(max_bin * kHistEntrySize, sizeof(LightSplitInfo) * num_machines_);
+  auto max_cat_threshold = this->config_->max_cat_threshold;
+  // need to be able to hold smaller and larger best splits in SyncUpGlobalBestSplit
+  size_t splitInfoSize = static_cast<size_t>(SplitInfo::Size(max_cat_threshold) * 2);
+  buffer_size = std::max(buffer_size, splitInfoSize);
```
StrikerRUS, line 43: This just cosmetic change will allow to find all occurences of this variable across the files in case of the need to change the value in the future.
```suggestion
  size_t split_info_size = static_cast<size_t>(SplitInfo::Size(max_cat_threshold) * 2);
  buffer_size = std::max(buffer_size, split_info_size);
```

imatiach-msft: @StrikerRUS great suggestion, done!

imatiach-msft:
close reopen for CI
```
URL        `https://en.wikipedia.org/wiki/Gamma_distribution#Occurrence_and_applications'

Name       `gamma-distributed'

Parent URL file:///home/travis/build/microsoft/LightGBM/docs/_build/html/Parameters.html, line 239, col 220

Real URL   https://en.wikipedia.org/wiki/Gamma_distribution#Occurrence_and_applications

Check time 3.469 seconds

D/L time   1.042 seconds

Size       345.98KB

Warning    [None] Anchor `Occurrence_and_applications' not found.

           Available anchors: `coll-download-as-rl', `footer',

           `footer-copyrightico', `footer-icons', `footer-info',

           `footer-info-copyright', `footer-info-lastmod',

           `footer-places', `footer-places-about',

           `footer-places-contact',

           `footer-places-cookiestatement',

           `footer-places-developers',

           `footer-places-disclaimer',

           `footer-places-mobileview', `footer-places-privacy',

           `footer-places-statslink', `footer-poweredbyico',

           `n-aboutsite', `n-contactpage', `n-portal',

           `n-recentchanges', `p-coll-print_export',

           `p-coll-print_export-label', `p-lang', `p-lang-label',

           `p-tb', `p-tb-label', `p-wikibase-otherprojects',

           `p-wikibase-otherprojects-label', `t-cite', `t-info',

           `t-permalink', `t-print', `t-recentchangeslinked',

           `t-specialpages', `t-upload', `t-whatlinkshere',

           `t-wikibase'.
```

PR #3113, opened 2020-05-24
[ci] peg MiKTeX mirror

jameslamb:
I learned tonight that the download link from the main MiKTeX site redirects to individual mirrors. I'm not sure what strategy it uses, but it seems to change pretty frequently. Here's an example from four requests executed within 15 seconds:

image

I think that pegging to a single mirror will reduce the frequency of errors like this downloading MiKTeX:

image

In theory that redirect mechanism they have is great because it means the download link should work whenever a particular mirror fails, but as we've seen it regularly routes us to a mirror that we can't successfully download from (for whatever reason). Since the `r-package` jobs already rely on the University of Illinois mirror being up (to install additional LaTeX packages), I'd like to propose that we should just go directly to that mirror for MiKTeX too.


jameslamb:
ugh wait actually it seems that that mirror only hosts one artifact at a very very specific version number (https://ctan.math.illinois.edu/systems/win32/miktex/setup/windows-x64/)

image

which means this code will break every time a new version is pushed.

Ok I will close this and look for other strategies, sorry!

PR #3119, opened 2020-05-26
[ci] use GitHub Actions for R CI jobs (fixes #2353)

jameslamb:
NOTE: moved from #3116 to use a branch on `LightGBM` instead of my fork.

-----------------

This PR proposes adding GitHub Actions CI and moving all R jobs to it.

GitHub Actions gives you 20 concurrent jobs for free and access to Mac, Linux, and Windows environments.

See #2353 and discussion in this thread: https://github.com/microsoft/LightGBM/pull/3065#discussion_r428246650

This PR should improve CI times (since AppVeyor runs tasks sequentially and now those tasks will be run concurrently), and should give enough extra CI capacity to test more installation paths for the R package(like R 3.6.x AND R 4.0.0x on #3065 )

## Summary of Changes

* move `lint` job from Travis to GitHub Actions
* remove R Linux and Mac CI jobs from Azure
* remove R jobs from AppVeyor
* remove R jobs from Travis
* add 6 R jobs to GitHub Actions:
    - Windows, MSVC
    - Windows, MINGW
    - Mac, gcc
    - Mac, clang
    - Linux, gcc
    - Linux, clang
* add `.ci/download-miktex.R`, to make `miktexsetup.zip` downloads more reliable

## Notes for reviewers

* I developed this on my fork, and you can look at the build history there: https://github.com/jameslamb/LightGBM/pull/28
* I'd like to keep the Azure Windows `r-package` job, since it gives us access to Visual Studio 2017. GitHub Actions only has Visual Studio 2019 ([reference]( https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners#supported-runners-and-hardware-resources))
* I added `.ci/download-miktex.R` because maybe 1 out of every 8 builds hit this issue: https://github.com/microsoft/LightGBM/pull/3113#issue-422362595
* the conda step comes from here: https://github.com/marketplace/actions/setup-miniconda
* I had an issue where any code that writes too `stderr` causes the build to fail, even if the command itself succeeds. I tracked that down to [this section of the docs](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsrun), which state:

> Fail-fast behavior when possible. For pwsh and powershell built-in shell, we will prepend $ErrorActionPreference = 'stop' to script contents.

## Other references

* https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_iddefaultsrun
[...]

PR #3124, opened 2020-05-28
[ci] make MiKTeX downloads more reliable

jameslamb:
Moved this over from #3119 once we decided not to introduce Windows tests in that PR.

---------------

I learned in #3113 that the download link from the main MiKTeX site redirects to individual mirrors. I'm not sure what strategy it used but it seems to change pretty frequently. That makes me think the strategy is some very simple random load balancing that doesn't account for the health of the target mirror.

Here's an example from four requests executed within 15 seconds:

image

I think that pegging to a single mirror will reduce the frequency of errors like this downloading MiKTeX:

image

In theory that redirect mechanism they have is great because it means the download link should work whenever a particular mirror fails, but as we've seen it regularly routes us to a mirror that we can't successfully download from (for whatever reason).

Since the `r-package` jobs already rely on the University of Illinois mirror being up (to install additional LaTeX packages), I'd like to propose that we should just go directly to that mirror for MiKTeX too.


jameslamb:
Ok this did fail [on AppVeyor](https://ci.appveyor.com/project/guolinke/lightgbm/builds/33162149/job/8y8ffr0f0cdyv9aa) with an error related to MiKTeX but NOT one that is caused by (or could be solved by) the changes in this PR.

image

This PR occurs after the MIKTeX installer has been downloaded, when its being set up.

.ci/test_r_package_windows.ps1
```diff
@@ -48,11 +48,15 @@ Write-Output "Installing Rtools"
...
 Write-Output "Done installing Rtools"
 
+Write-Output "Installing dependencies"
+$packages = "c('data.table', 'httr', 'jsonlite', 'Matrix', 'processx', 'R6', 'testthat'), dependencies = c('Imports', 'Depends', 'LinkingTo')"
+Rscript --vanilla -e "options(install.packages.check.source = 'no'); install.packages($packages, repos = '$env:CRAN_MIRROR', type = 'binary', lib = '$env:R_LIB_PATH')" ; Check-Output $?
+
 # MiKTeX and pandoc can be skipped on non-MINGW builds, since we don't
 # build the package documentation for those
 if ($env:COMPILER -eq "MINGW") {
     Write-Output "Downloading MiKTeX"
```
StrikerRUS, line 58: I think you can remove this line because in R script we have `print(sprintf("downloading %s", full_zip_url))`.

.ci/test_r_package_windows.ps1
```diff
@@ -48,11 +48,15 @@ Write-Output "Installing Rtools"
 Start-Process -FilePath Rtools.exe -NoNewWindow -Wait -ArgumentList "/VERYSILENT /DIR=$env:R_LIB_PATH/Rtools" ; Check-Output $?
 Write-Output "Done installing Rtools"
 
+Write-Output "Installing dependencies"
+$packages = "c('data.table', 'httr', 'jsonlite', 'Matrix', 'processx', 'R6', 'testthat'), dependencies = c('Imports', 'Depends', 'LinkingTo')"
```
StrikerRUS, line 52: We don't need `httr` package for non-MinGW builds. It just increases installation time and the number of installed packages that potentially can conflict with each other. Also, it is technically not a `dependency` of LightGBM (refer to the log message above).

jameslamb:
Since this is only used by R builds, what is the benefit of re-writing it into Powershell? Web scraping in Powershell does not sound fun to me

StrikerRUS:
> Since this is only used by R builds, what is the benefit of re-writing it into Powershell?

As this particular script is needed only for R builds, it is OK from the first glance. But in general we should avoid using any specific languages for administration stuff. It will allow collaborators without knowledge of one specific programming language to maintain those tools in the future.

Going this way we may end up using dozens of languages calling one script from another for repo maintenance purposes and one ought to know all of them to just fix failing CI job.

Some other bad examples of unjustified use of Python:
https://github.com/microsoft/LightGBM/tree/master/helpers
https://github.com/microsoft/LightGBM/tree/master/.nuget

.ci/test_r_package_windows.ps1
jameslamb: oh true! I'm going to move an `install.packages("httr")` into the script downloading MiKTeX. Something like

```r
if (!require("httr")){
    install.packages("httr")
}
```

That way, we won't have to introduce another `if ($env:COMPILER -ne "MINGW")` stuff in `test_r_package_windows.ps1`.

jameslamb:
> > Since this is only used by R builds, what is the benefit of re-writing it into Powershell?
> 
> As this particular script is needed only for R builds, it is OK from the first glance. But in general we should avoid using any specific languages for administration stuff. It will allow collaborators without knowledge of one specific programming language to maintain those tools in the future.
> 
> Going this way we may end up using dozens of languages calling one script from another for repo maintenance purposes and one ought to know all of them to just fix failing CI job.
> 
> Some other bad examples of unjustified use of Python:
> https://github.com/microsoft/LightGBM/tree/master/helpers
> https://github.com/microsoft/LightGBM/tree/master/.nuget

Makes sense, thank you for the explanation!

In this case I really do think R code makes sense since this is specific to the R package, and since the task (scraping the source of an archive page to find the name of a file) is one that I personally don't know how to do in Powershell. I know it's possible, but it would take me a while and I'm guessing that what I come up with would be error prone.

We already have other R code in R-specific CI administration (https://github.com/microsoft/LightGBM/blob/master/.ci/lint_r_code.R), and I know that any time `r-package` tasks fail in CI, I'll get `@`-ed.

.ci/test_r_package_windows.ps1
StrikerRUS: I like your idea!

StrikerRUS:
> We already have other R code in R-specific CI administration 

That script uses API of `lintr` package, so it is required to be written in R.

> and I know that any time `r-package` tasks fail in CI, I'll get `@`-ed.

Yeah, it is exactly what I'm saying about! And it is not good, because maintainers can leave projects for some reasdon and it is a natural process. It will result in a some kind of "bus factor" (e.g. similar problems: https://github.com/dmlc/xgboost/issues/4958).

Please treat all my thoughts above as a my general opinion to the whole repo management. I'm not against that this particular PR will add parsing a web page code in R.

jameslamb:
> I'm not against that this particular PR will add parsing a web page code in R.

Thanks for this @StrikerRUS . I decided to take learning how to do this in Powershell as a learning opportunity, and I'm really glad that I did! It was a lot easier than I expected. See the changes in https://github.com/microsoft/LightGBM/pull/3124/commits/23cf95a9f480c88a5839c2b69f9509c983ebd958

It was easier than I expected because I learned something amazing...[you can run Powershell on a Mac](https://docs.microsoft.com/en-us/powershell/scripting/install/installing-powershell-core-on-macos?view=powershell-7)! 

image

This made it really easy to test and experiment in my usual dev environment 😀

.ci/test_r_package_windows.ps1
```diff
@@ -48,27 +67,26 @@ Write-Output "Installing Rtools"
...
     Write-Output "Done installing MiKTeX"
 
     initexmf --set-config-value [MPM]AutoInstall=1
     conda install -q -y --no-deps pandoc
 }
 
-Write-Output "Installing dependencies"
-$packages = "c('data.table', 'jsonlite', 'Matrix', 'processx', 'R6', 'testthat'), dependencies = c('Imports', 'Depends', 'LinkingTo')"
-Rscript --vanilla -e "options(install.packages.check.source = 'no'); install.packages($packages, repos = '$env:CRAN_MIRROR', type = 'binary', lib = '$env:R_LIB_PATH')" ; Check-Output $?
-
```
jameslamb, line 71: I'll undo this change. This doesn't have to move now that we aren't installing `{htttr}`  for the MiKTeX download!

StrikerRUS:
@jameslamb 

Thanks a lot for rewriting this in PowerShell!

> It was easier than I expected because I learned something amazing...you can run Powershell on a Mac!

Yeah! PS Core is a great thing along with .NET Core 😃 ,

.ci/test_r_package_windows.ps1
```diff
@@ -12,12 +12,31 @@ function Download-File-With-Retries {
   } while(!$?);
 }
 
+# mirrors that host miktexsetup.zip do so only with explicitly-named
+# files like miktexsetup-2.4.5.zip, so hard-coding a link to an archive as a
+# way to peg to one mirror does not work
+#
+# this function will find the specific version of miktexsetup.zip at a given
+# given mirror and download it
```
StrikerRUS, line 20:
```suggestion
# mirror and download it
```

.ci/test_r_package_windows.ps1
```diff
@@ -48,27 +67,26 @@ Write-Output "Installing Rtools"
...
 Write-Output "Done installing Rtools"
 
+Write-Output "Installing dependencies"
+$packages = "c('data.table', 'jsonlite', 'Matrix', 'processx', 'R6', 'testthat'), dependencies = c('Imports', 'Depends', 'LinkingTo')"
+Rscript --vanilla -e "options(install.packages.check.source = 'no'); install.packages($packages, repos = '$env:CRAN_MIRROR', type = 'binary', lib = '$env:R_LIB_PATH')" ; Check-Output $?
+
 # MiKTeX and pandoc can be skipped on non-MINGW builds, since we don't
 # build the package documentation for those
 if ($env:COMPILER -eq "MINGW") {
-    Write-Output "Downloading MiKTeX"
```
StrikerRUS, line 54: I think we can bring back this log message after switching from R script to PS script 😃 .

jameslamb: yep exactly!

.ci/test_r_package_windows.ps1
jameslamb: 😆

.ci/test_r_package_windows.ps1
jameslamb: undid these changes in https://github.com/microsoft/LightGBM/pull/3124/commits/10e32faed00502d6656108019aa52715b27c6a83, so now all that is left is the MiKTeX stuff.

If it build successfully I'll merge  this!  Hopefully it will help reduce frequency failing R builds that affect other PRs

jameslamb:
I just rebased to `master` to get the latest changes, especially #3133 

Happy to see that GitHub Actions got triggered on this PR!

image

Since it isn't required yet (https://github.com/microsoft/LightGBM/pull/3119#issuecomment-637192398), I'll be sure to check them manually before merging.

Issue #2884, opened 2020-03-07
Interaction constraints


mayer79:
## Feature interaction

Monotonicity constraints is one way to make the black-box more intuitive and interpretable. For tree based models, using [Interaction constraints](https://xgboost.readthedocs.io/en/latest/tutorials/feature_interaction_constraint.html) is another highly interesting possibility: Passing a nested list like [[0, 1], [2]] specifies which features may be selected in the same tree branch, see explanations in https://github.com/dmlc/xgboost/issues/4075 in general and https://discuss.xgboost.ai/t/feature-interaction-constraint-overlap/506 for what happens if the sets overlap.

Quoting hcho3: "Let S the set of features used in all the ancestor nodes of the proposed split (plus itself). The proposed split is admitted whenever at least one set in the interaction constraint is a superset of S."

I think this would be a big step towards making boosted trees more interpretable and to e.g. being able to specify "additive" components as well as to reflect ethical constraints into a model.

guolinke:
It seems is a very interesting feature, contribution is very welcomed.
And I will implement it when I have time and if not PR is proposed.

mayer79:
That would be simply fantastic as my C++ skills are even worse than me speaking French. Still, you can test the implementation e.g. with Friedman's H statistic, a model agnostic approach to measure interaction strength. It is implemented in R packages iml and flashlight (mine):

```
library(xgboost)

n <- 1000
m <- 4

set.seed(1)
X <- matrix(rnorm(n * m), ncol = m, dimnames = list(NULL, paste0("X", 1:m)))
cols <- colnames(X)
y <- X[, 1] * X[, 2] * X[, 3] + X[, 4] + rnorm(n)

params <- list(
  learning_rate = 0.1,
  depth = 4, 
  objective = "reg:linear",
  #interaction_constraints = list(0, c(1, 2, 3)) # col index starts at 0
  interaction_constraints = list(0, 1, 2, 3) # col index starts at 0
)

train <- xgb.DMatrix(X, label = y)

fit <- xgb.train(params, train, nrounds = 50)


library(flashlight)
fl <- flashlight(model = fit, data = data.frame(y, X), y = "y", label = "xgb",
                 predict_fun = function(m, X) predict(m, data.matrix(X[, cols])))
plot(imp <- light_interaction(fl, v = cols, pairwise = FALSE))

```

The first parametrization gives

[...]

PR #3126, opened 2020-05-28
Interaction constraints

btrotta:
Implement interaction constraints feature requested in #2884.

### DIFF
--- a/R-package/R/lgb.train.R
+++ b/R-package/R/lgb.train.R
@@ -124,6 +124,60 @@ lgb.train <- function(params = list(),
     end_iteration <- begin_iteration + nrounds - 1L
   }
 
+  # Convert interaction constraints to feature numbers
+  if (!is.null(params[["interaction_constraints"]])) {
+
+      string_constraints = list()
+
+      # Get feature names
+      cnames <- NULL
+      if (!is.null(colnames)) {
+        cnames <- colnames
+      } else if (!is.null(data$get_colnames())) {
+        cnames <- data$get_colnames()
+      }
+
+      for (constraint in params[["interaction_constraints"]]) {
+      
+        # Check for character name
+        if (is.character(constraint)) {
+
+            constraint_indices <- as.list(match(constraint, cnames) - 1L)
+
+            # Provided indices, but some indices are not existing?
+            if (sum(is.na(constraint_indices)) > 0L) {
+              stop(
+                "lgb.train: supplied an unknown feature in interaction_constraints "
+                , sQuote(constraint[is.na(constraint_indices)])
+              )
+            }
+
+          } else {
+
+            # Check that constraint indices are at most number of features
+            if (max(constraint) > length(cnames)) {
+              stop(
+                "lgb.train: supplied a too large value in interaction_constraints: "
+                , max(constraint)
+                , " but only "
+                , length(cnames)
+                , " features"
+              )
+            }
+
+            # Store indices as [0, n-1] indexed instead of [1, n] indexed
+            constraint_indices <- as.list(constraint - 1L)
+
+          }
+
+          # Convert constraint to string
+          constraint_string <- paste0("[", paste0(constraint_indices, collapse=","),"]")
+          string_constraints <- append(string_constraints, constraint_string)
+      }
+
+      params[["interaction_constraints"]] = string_constraints
+  }
+
   # Update parameters with parsed parameters
   data$update_params(params)
 
--- a/R-package/R/lgb.train.R
+++ b/R-package/R/lgb.train.R
@@ -124,58 +124,8 @@ lgb.train <- function(params = list(),
     end_iteration <- begin_iteration + nrounds - 1L
   }
 
-  # Convert interaction constraints to feature numbers
   if (!is.null(params[["interaction_constraints"]])) {
-
-      string_constraints <- list()
-
-      # Get feature names
-      cnames <- NULL
-      if (!is.null(colnames)) {
-        cnames <- colnames
-      } else if (!is.null(data$get_colnames())) {
-        cnames <- data$get_colnames()
-      }
-
-      for (constraint in params[["interaction_constraints"]]) {
-
-        # Check for character name
-        if (is.character(constraint)) {
-
-            constraint_indices <- as.list(match(constraint, cnames) - 1L)
-
-            # Provided indices, but some indices are not existing?
-            if (sum(is.na(constraint_indices)) > 0L) {
-              stop(
-                "lgb.train: supplied an unknown feature in interaction_constraints "
-                , sQuote(constraint[is.na(constraint_indices)])
-              )
-            }
-
-          } else {
-
-            # Check that constraint indices are at most number of features
-            if (max(constraint) > length(cnames)) {
-              stop(
-                "lgb.train: supplied a too large value in interaction_constraints: "
-                , max(constraint)
-                , " but only "
-                , length(cnames)
-                , " features"
-              )
-            }
-
-            # Store indices as [0, n-1] indexed instead of [1, n] indexed
-            constraint_indices <- as.list(constraint - 1L)
-
-          }
-
-          # Convert constraint to string
-          constraint_string <- paste0("[", paste0(constraint_indices, collapse = ","), "]")
-          string_constraints <- append(string_constraints, constraint_string)
-      }
-
-      params[["interaction_constraints"]] <- string_constraints
+    stop("lgb.train: interaction_constraints is not implemented")
   }
 
   # Update parameters with parsed parameters
--- a/R-package/R/lgb.train.R
+++ b/R-package/R/lgb.train.R
@@ -127,7 +127,7 @@ lgb.train <- function(params = list(),
   # Convert interaction constraints to feature numbers
   if (!is.null(params[["interaction_constraints"]])) {
 
-      string_constraints = list()
+      string_constraints <- list()
 
       # Get feature names
       cnames <- NULL
@@ -138,7 +138,7 @@ lgb.train <- function(params = list(),
       }
 
       for (constraint in params[["interaction_constraints"]]) {
-      
+
         # Check for character name
         if (is.character(constraint)) {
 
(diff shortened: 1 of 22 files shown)


docs/Parameters.rst
```diff
@@ -538,6 +538,18 @@ Learning Control Parameters
...
 
+-  ``interaction_constraints`` :raw-html:`<a id="interaction_constraints" title="Permalink to this parameter" href="#interaction_constraints">&#x1F517;&#xFE0E;</a>`, default = ``""``, type = string
+
+   -  controls which features can appear in the same branch
+
+   -  for CLI, specify as lists separated by commas, e.g. ``[0,1,2],[2,3]``
+
+   -  for Python, specify as list of lists, e.g. ``[[0, 1, 2], [2, 3]]``
+
+   -  for R, specify as a string similar to the CLI format (ensuring there are no spaces), e.g. ``"[0,1,2],[2,3]"``
```
jameslamb, line 549: For R, I think that we should make the interface "a list of character vectors, where each vector is a list of feature names", and then inside the R package turn the feature names into indices.

I think passing a string of an array in a comma-delimited list puts too much burden on R users.  @Laurae2 what do you think?

python-package/lightgbm/basic.py
```diff
@@ -135,7 +135,12 @@ def param_dict_to_str(data):
     pairs = []
     for key, val in data.items():
         if isinstance(val, (list, tuple, set)) or is_numpy_1d_array(val):
-            pairs.append(str(key) + '=' + ','.join(map(str, val)))
+            def to_string(x):
+                if isinstance(x, list):
+                    return '[' + ','.join(map(str, x)) + ']'
```
jameslamb, line 140:
```suggestion
                    return "[{}]".format(','.join(map(str, x)))
```

I think this would be a little easier to read using string interpolation. I recommend  `.format()` since it's an option supported in Python 2 and 3. @StrikerRUS what do you think?

docs/Parameters.rst
btrotta: @jameslamb Good suggestion, thanks! I've tried to implement this, but I'm not that familiar with R so would be great if you could take a look when you get a chance.

jameslamb: I think it would be ok to say here that it is not supported in the R package yet, and then create an issue like `[R-package] add support for interaction_constraints` and add it to #2302 . Then I or another R maintainer can get to it without delaying this pull request.

btrotta: Done, see #3130

R-package/R/lgb.train.R
```diff
@@ -124,6 +124,60 @@ lgb.train <- function(params = list(),
     end_iteration <- begin_iteration + nrounds - 1L
   }
 
+  # Convert interaction constraints to feature numbers
+  if (!is.null(params[["interaction_constraints"]])) {
```
jameslamb, line 128: thanks for this! 

@btrotta I'll leave some comments here and you can decide, but I think it would make sense to do this in a separate pull request so it doesn't hold this one up.

* Unfortunately we have a lot of duplication between `lgb.train()` and  `lgb.cv()` today 
 (https://github.com/microsoft/LightGBM/blob/master/R-package/R/lgb.cv.R), so handling for this parameter needs to be added in both places. I think it would make sense to create a new internal function like  `.check_interaction_constraints()`, similar to how we have [`lgb.check.obj()`](https://github.com/microsoft/LightGBM/blob/5304885c0aa4f26e29ae2a4d451c0fb39d70f24b/R-package/R/utils.R#L186), that is then called by both functions. That would also make it easier to unit test the handling of this parameter.
* We should add some validation that `interaction_constraints` is a list of either numeric or character vectors. Like this

```r
interaction_constraints <- params[["interaction_constraints"]]
if (!is.null(interaction_constraints)){
    # validation
    if (!methods::is(interaction_constraints, "list")){
        stop("interaction_constraints must be a list")
    }
    if (!all(sapply(interaction_constraints, function(x){is.character(x) || is.numeric(x)}))){
        stop("every element in interaction_constraints must be a character vector or numeric vector)
    }

    # rest of the code
}
```

*  I think the uses of `as.list()` in this code should be replaced with `as.integer()`
* I'd like too have unit tests, for both `lgb.train()` and `lgb.cv()`. That will catch some issues and prevent us from breaking this functionality in the future.
    - validation that you get an informative error if you don't pass a `list` object
    - validation that you get an informative error if you pass a `list` of something other than character vectors and numeric vectors
    - validation that you get an informative error for indices that are greater than `max_features`
    - test training with only passing indices
    - test training with only passing feature names
    - test training with passing a mix of indices and feature names

btrotta: Ok sure, I have moved these changes to new PR #3136.

guolinke:
@btrotta is this ready to review?

btrotta:
@guolinke  Yes

src/treelearner/col_sampler.hpp
```diff
@@ -83,32 +88,92 @@ class ColSampler {
     }
   }
 
-  std::vector<int8_t> GetByNode() {
-    if (fraction_bynode_ >= 1.0f) {
-      return std::vector<int8_t>(train_data_->num_features(), 1);
+  std::vector<int8_t> GetByNode(const Tree* tree, int leaf) {
+    // get interaction constraints for current branch
+    std::unordered_set<int> allowed_features;
```
guolinke, line 93: is that possible to cache the used features of all branches, rather than re-generate them every time?

btrotta: I've changed the code so that we keep track of the features in each branch as we build the tree.

guolinke: Thanks so much! However, it seems the tracking is still enabled even when interaction constraint is not used?

btrotta: Thanks, I changed it now so it only keeps track if interaction constraint is enabled.

include/LightGBM/config.h
```diff
@@ -505,6 +505,13 @@ struct Config {
   // descl2 = note that the parent output ``w_p`` itself has smoothing applied, unless it is the root node, so that the smoothing effect accumulates with the tree depth
   double path_smooth = 0;
 
+  // desc = controls which features can appear in the same branch
+  // desc = for CLI, specify as lists separated by commas, e.g. ``[0,1,2],[2,3]``
+  // desc = for Python, specify as list of lists, e.g. ``[[0, 1, 2], [2, 3]]``
+  // desc = not yet supported for R
+  // desc = any two features can only appear in the same branch only if there exists a constraint containing both feaures
```
StrikerRUS, line 512: I think we should clarify, that this is disabled by default. Something like 
```
// desc = by default interaction constraints are disabled, to enable them you can specify
// descl2 = for CLI, lists separated by commas, ...
// descl2 = for Python-package, list of lists, ...
// descl2 = for R-package, **not yet supported**
```

```suggestion
  // desc = any two features can only appear in the same branch only if there exists a constraint containing both features
```

tests/python_package_test/test_engine.py
```diff
@@ -2185,3 +2185,34 @@ def _imptcs_to_numpy(X, impcts_dict):
...
             self.assertIsNone(tree_df.loc[0, col])
+
+    def test_interaction_constraints(self):
+        X, y = load_boston(True)
+        num_features = X.shape[1]
+        train_data = lgb.Dataset(X, label=y)
+        # check that constraint containing all features is equivalent to no constraint
+        params = {'verbose': -1,
+                  'seed': 0}
+        est = lgb.train(params, train_data)
```
StrikerRUS, line 2196: Can we reduce the number of rounds to save CI time? Like `num_boost_round=5` or `10`.

tests/python_package_test/test_engine.py
```diff
@@ -2185,3 +2185,34 @@ def _imptcs_to_numpy(X, impcts_dict):
...
+        pred1 = est.predict(X)
+        params = {'interaction_constraints': [list(range(num_features))],
+                  'verbose': -1,
+                  'seed': 0}
+        est = lgb.train(params, train_data)
```
StrikerRUS, line 2201: I think we can reduce the duplication of code and increase params readability by writing
```
params = {'verbose': -1,
          'seed': 0}
est = lgb.train(params, train_data)
...
est = lgb.train(dict(params,
                     interaction_constraints=[list(range(num_features))]),
                train_data)
...
est = lgb.train(dict(params,
                     interaction_constraints=[list(range(num_features // 2)),
                                              list(range(num_features // 2, num_features))]),
                train_data)
```

btrotta:
@StrikerRUS Thanks, have updated according to your suggestions.

mayer79:
@btrotta: Simply fantastic, thank you so much for digging into this and the great implementation. 

I made some tests in R and all make perfect sense:

```
library(lightgbm)
library(flashlight)

n <- 1000
m <- 4

set.seed(1)
X <- matrix(rnorm(n * m), ncol = m, dimnames = list(NULL, paste0("X", 1:m)))
cols <- colnames(X)
y <- X[, 1] * X[, 2] * X[, 3] + X[, 4] + rnorm(n)

train <- lgb.Dataset(X, label = y)
```

## Example 1: Additive model with depth > 1
```
params <- list(
  learning_rate = 0.1,
  num_leaves = 15, 
  objective = "regression",
  interaction_constraints = list(1, 2, 3, 4)
)

fit <- lgb.train(params, train, nrounds = 50)

fl <- flashlight(model = fit, data = data.frame(y, X), y = "y", label = "lgb",
                 predict_fun = function(m, X) predict(m, data.matrix(X[, cols])))
plot(imp <- light_interaction(fl, v = cols, pairwise = TRUE))

```
image

Okay -> pairwise H-statistic values are all 0. H measures the interaction strength in a model agnostic way based on bivariate partial dependence curves.

## Some additive, some pairwise
```
params <- list(
  learning_rate = 0.1,
  num_leaves = 15, 
  objective = "regression",
  interaction_constraints = list(c(1, 2), 3, 4)
)

fit <- lgb.train(params, train, nrounds = 50)

fl <- flashlight(model = fit, data = data.frame(y, X), y = "y", label = "lgb",
                 predict_fun = function(m, X) predict(m, data.matrix(X[, cols])))
plot(imp <- light_interaction(fl, v = cols, pairwise = TRUE))
```
image

Okay -> there is only one interacting variable pair.

## Overlapping sets
```

params <- list(
  learning_rate = 0.1,
  num_leaves = 15, 
  objective = "regression",
  interaction_constraints = list(c(1, 2), c(1, 3))
)

fit <- lgb.train(params, train, nrounds = 50)

fl <- flashlight(model = fit, data = data.frame(y, X), y = "y", label = "lgb",
                 predict_fun = function(m, X) predict(m, data.matrix(X[, cols])))
plot(imp <- light_interaction(fl, v = cols, pairwise = TRUE))
```
image

-> Okay as well! X2 and X3 are not interacting, as intended.

And note to myself: Unlisted variables are modelled without interaction.

btrotta:
@mayer79 Thanks for doing those additional tests, good to know it's working as expected!

PR #3129, opened 2020-05-30
[python][scikit-learn] add new attribute for used number of features

a-wozniakowski:
Fixes issue related to [#17353](https://github.com/scikit-learn/scikit-learn/issues/17353) in scikit-learn (see [SLEP010](https://scikit-learn-enhancement-proposals.readthedocs.io/en/latest/slep010/proposal.html) for further details).

### DIFF
--- a/python-package/lightgbm/sklearn.py
+++ b/python-package/lightgbm/sklearn.py
@@ -257,7 +257,7 @@ class LGBMModel(_LGBMModelBase):
         ----------
         n_features_ : int
             The number of features of fitted model.
-        n_features_in_: int
+        n_features_in_ : int
             The number of features of fitted model.
         classes_ : array of shape = [n_classes]
             The class label array (only for classification problem).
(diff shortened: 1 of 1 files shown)


a-wozniakowski:
@StrikerRUS thanks!

[```BaseEstimator```](https://github.com/scikit-learn/scikit-learn/blob/7cc815c21c60dabe8252b30ada77621b88e2136c/sklearn/base.py) has a private method ```_validate_data```, which checks the data then checks the number of features. This method is typically used in an estimator's ```fit``` method, so I modified the pull request accordingly.

python-package/lightgbm/sklearn.py
```diff
@@ -257,6 +257,8 @@ def __init__(self, boosting_type='gbdt', num_leaves=31, max_depth=-1,
         ----------
         n_features_ : int
             The number of features of fitted model.
+        n_features_in_: int
```
StrikerRUS, line 260:
```suggestion
        n_features_in_ : int
```

Issue #3130, opened 2020-05-31
[R-package] add support for interaction_constraints


btrotta:
Related to #3126. I have made an attempt at implementing the R interface to specify the interaction constraints, but needs review from the R maintainers.

StrikerRUS:
Closed via #3136.

PR #3136, opened 2020-06-01
[R-package] Interface for interaction constraints

btrotta:
Related to PR #3126 and issue #3130. Add the R interface for interaction constraints. I have made an attempt at implementing this (thanks for the guidance @jameslamb), but I'm not an R expert, so needs review from R maintainers. I have enabled "Allow edits by maintainers" so feel free to commit directly to the branch.


R-package/tests/testthat/test_basic.R
```diff
@@ -1021,3 +1021,73 @@ test_that("using lightgbm() without early stopping, best_iter and best_score com
...
+  params <- list(objective = "regression",
+                 interaction_constraints=list(list(1L, length(colnames(train$data)) + 1), list(3L)))
+    expect_error({
+      bst <- lightgbm(
+        data = dtrain
+        , params = params
+        , nrounds = 2
+      )
+    }, "every element in interaction_constraints must be a character vector or numeric vector")
+})
```
jameslamb, line 1060: I don't think this test is right. We shouldn't get the "must be a character or numeric vector" for this one, but an error like `supplied a too large value in interaction_constraints`

R-package/tests/testthat/test_basic.R
```diff
@@ -1021,3 +1021,73 @@ test_that("using lightgbm() without early stopping, best_iter and best_score com
...
+
+test_that("lgb.train() gives same result when interaction_constraints is specific as a list of character vectors, numeric vectors, or a combination", {
```
jameslamb, line 1062:
```suggestion
test_that("lgb.train() gives same result when interaction_constraints is specified as a list of character vectors, numeric vectors, or a combination", {
```

R-package/tests/testthat/test_basic.R
```diff
@@ -1021,3 +1021,73 @@ test_that("using lightgbm() without early stopping, best_iter and best_score com
...
+
+  params <- list(objective = "regression", list(c(1L, 2L), c(3L)))
+  bst <- lightgbm(
+    data = dtrain
+    , params = params
+    , nrounds = 2
+  )
+  pred1 <- bst$predict(test$data)
+
+  cnames = colnames(train$data)
```
jameslamb, line 1073:
```suggestion
  cnames <- colnames(train$data)
```

Could you please use `<-` as the assignment operator everywhere? I think you will get linter errors about it.

R-package/tests/testthat/test_basic.R
```diff
@@ -1021,3 +1021,73 @@ test_that("using lightgbm() without early stopping, best_iter and best_score com
...
+  params <- list(objective = "regression", list(c(cnames[[1]], cnames[[2]]), c(cnames[[3]])))
+  bst <- lightgbm(
+    data = dtrain
+    , params = params
+    , nrounds = 2
```
jameslamb, line 1078: you'll see linter errors about "implicit integers" or something like that. Can you change all of these to `1L`, `2L`, etc.?

R-package/tests/testthat/test_basic.R
```diff
@@ -1021,3 +1021,73 @@ test_that("using lightgbm() without early stopping, best_iter and best_score com
...
+        , nrounds = 2
+      )
+    }, "every element in interaction_constraints must be a character vector or numeric vector")
+})
+
+test_that("lgb.train() gives same result when interaction_constraints is specific as a list of character vectors, numeric vectors, or a combination", {
+  dtrain <- lgb.Dataset(train$data, label = train$label)
+
+  params <- list(objective = "regression", list(c(1L, 2L), c(3L)))
+  bst <- lightgbm(
```
jameslamb, line 1066: I think you need to set a seed at the top of this test (right above `dtrain <- ...`). Like `set.seed(708L)`

docs/Parameters.rst
```diff
@@ -538,6 +538,18 @@ Learning Control Parameters
...
 
+-  ``interaction_constraints`` :raw-html:`<a id="interaction_constraints" title="Permalink to this parameter" href="#interaction_constraints">&#x1F517;&#xFE0E;</a>`, default = ``""``, type = string
+
+   -  controls which features can appear in the same branch
+
+   -  for CLI, specify as lists separated by commas, e.g. ``[0,1,2],[2,3]``
+
+   -  for Python, specify as list of lists, e.g. ``[[0, 1, 2], [2, 3]]``
+
+   -  for R, specify as a list of character or numeric vectors, e.g. ``list(c("var1", "var2", "var3"), c("var3", "var4"))`` or ``list(c(1L, 2L, 3L), c(3L, 4L))``
```
jameslamb, line 549: Can you clarify here that the numeric vectors need to be 0-based? R vectors are 1-based so I could see that leading to confusion.

Something like this (but feel free to change the language)

> If numeric, should be 0-based. So `0` means "the first feature", `1` means "the second feature", etc.

R-package/tests/testthat/test_basic.R
btrotta: Fixed, thanks. It was actually passing before despite this mistake, but that's because I had another error in the test!

docs/Parameters.rst
btrotta: Actually it expects 1-based vectors then converts to zero-based, similar to set_categorical_feature (where I copied most of the code from). I have updated the docs to clarify that.

jameslamb: oh great, ok thanks!

jameslamb:
@btrotta are you planning to come back to this? Or do you want me to pick it up?

btrotta:
@jameslamb  I left this for a while while I was waiting for the main PR #3126 to be merged into master. This has been done now so I have merged the changes from there to here. I think I've addressed all your previous comments, let me know if there are any other changes you'd like me to make.

docs/Parameters.rst
```diff
@@ -548,7 +548,7 @@ Learning Control Parameters
 
       -  for Python-package, list of lists, e.g. ``[[0, 1, 2], [2, 3]]``
 
-      -  for R-package, **not yet supported**
+      -  for R-package, list of character or numeric vectors, e.g. ``list(c("var1", "var2", "var3"), c("var3", "var4"))`` or ``list(c(1L, 2L, 3L), c(3L, 4L))``. Numeric vectors should use 1-based indexing, where ``1L`` is the first feature, ``2L`` is the second feature, etc
```
jameslamb, line 551: this is perfect, thank you!

src/io/tree.cpp
```diff
@@ -41,6 +42,7 @@ Tree::Tree(int max_leaves, bool track_branch_features)
   leaf_value_[0] = 0.0f;
   leaf_weight_[0] = 0.0f;
   leaf_parent_[0] = -1;
+  internal_parent_[0] = -1;
```
jameslamb, line 45: how do the changes here and in `tree.h` relate to the R package?

btrotta: Sorry, this was a remnant of the old version I branched off. I have removed this code now.

R-package/tests/testthat/test_basic.R
btrotta: I think the line numbers have changed and this comment was originally referring to this test: "lgb.train() gives same result when interaction_constraints is specified as a list of character vectors, numeric vectors, or a combination" which does have the seed set now.

jameslamb: I think you're right. Thanks!

src/io/tree.cpp
jameslamb: got it, thanks!

jameslamb:
@btrotta in case you haven't seen it in this project, this issue in the Windows R jobs on GitHub Actions are not the fault of this pull request:

image

Testing R packages means you have to build their documentation, which includes some LaTeX dependencies. The servers that provide these dependencies are not very reliable. In my experience, this CTAN server we use rarely has periods of downtime longer than 5 minutes, but also rarely goes 24 hours without any downtime.

If you re-run GitHub Actions jobs after a few minutes, they should pass.

btrotta:
@jameslamb  I tried rerunning it a few times today, but it's still failing every time.

jameslamb:
> @jameslamb I tried rerunning it a few times today, but it's still failing every time.

ah! I've opened #3198 to investigate this, looking into it now.

jameslamb:
it looks like `MiKTeX`'s versioning scheme was updated (https://github.com/MiKTeX/miktex/issues/554), and `miktexsetup` was bumped from 2.x to 4.x: https://github.com/MiKTeX/miktex/commit/f04aefa4578cba09a570cf85a9e8b0e3bd639197

jameslamb:
Ok @btrotta I think #3199 will fix the R jobs, sorry!

jameslamb:
Ok now that #3199 has been fixed, HOPEFULLY merging it into this PR will fix the  tests 😅

Issue #3139, opened 2020-06-02
prediction failure with integer data


mayer79:
System: R 3.6.3, lgb 2.3.2, Windows 10

LightGBM predictions fail silently for data matrix of type integer. This e.g. happens if a data set consisting of factors only is converted via `data.matrix`. 

```
library(lightgbm)

# Create integer only data matrix
set.seed(1)
n <- 1000
p <- 3
X <- as.matrix(sample(1:10, n * p, replace = TRUE), ncol = p)
y <- rowSums(X) + rnorm(n, sd = 10) 
mean(y) # 6.060638

# Training works
dtrain <- lgb.Dataset(X, label = y)
fit <- lgb.train(params = list(objective = "regression"), data = dtrain, nrounds = 100)

# Prediction works but the values are completely wrong
mean(predict(fit, X)) # 1.927871 

# Let's turn to double
storage.mode(X) <- "numeric"
mean(predict(fit, X)) # 6.060638 Hurray!
```

In the R-function  `lgb.Dataset`, there is a block

  ```
    # Check for matrix format
      if (is.matrix(data)) {
        # Check whether matrix is the correct type first ("double")
        if (storage.mode(data) != "double") {
          storage.mode(data) <- "double"
        }
      }
```
This block is missing in `lgb.Predictor`
```

 # Check if data is a matrix
        if (is.matrix(data)) {
          preds <- lgb.call(
            "LGBM_BoosterPredictForMat_R"
            , ret = preds
            , private$handle
            , data
            , as.integer(nrow(data))
            , as.integer(ncol(data))
            , as.integer(rawscore)
            , as.integer(predleaf)
            , as.integer(predcontrib)
            , as.integer(num_iteration)
            , private$params
          )
```

So I think the bug can be fixed by adding to predict the following lines.
```
 #Check whether matrix is the correct type first ("double")
      if (storage.mode(data) != "double") {
        storage.mode(data) <- "double"
      }
```

PR #3140, opened 2020-06-02
[R-package] Turn matrix to storage mode "double"

mayer79:
Fixes https://github.com/microsoft/LightGBM/issues/3139

Just copy-pasted code snippet from lgb.Dataset initalizer.

As I am still not very familiar with the LightGBM data logic, @Laurae2 and @jameslamb would need to confirm that this makes any sense.

### DIFF
--- a/R-package/R/lgb.Predictor.R
+++ b/R-package/R/lgb.Predictor.R
@@ -142,7 +142,7 @@ Predictor <- R6::R6Class(
 
         # Check if data is a matrix
         if (is.matrix(data)) {
-        # Check whether matrix is the correct type first ("double")
+          # Check whether matrix is the correct type first ("double")
           if (storage.mode(data) != "double") {
             storage.mode(data) <- "double"
           }
--- a/R-package/R/lgb.Predictor.R
+++ b/R-package/R/lgb.Predictor.R
@@ -142,6 +142,10 @@ Predictor <- R6::R6Class(
 
         # Check if data is a matrix
         if (is.matrix(data)) {
+        # Check whether matrix is the correct type first ("double")
+          if (storage.mode(data) != "double") {
+            storage.mode(data) <- "double"
+          }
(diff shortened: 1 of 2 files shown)


Laurae2:
@jameslamb This issue will also solve https://github.com/microsoft/LightGBM/issues/3094

R-package/R/lgb.Predictor.R
```diff
@@ -142,6 +142,10 @@ Predictor <- R6::R6Class(
 
         # Check if data is a matrix
         if (is.matrix(data)) {
+        # Check whether matrix is the correct type first ("double")
+          if (storage.mode(data) != "double") {
```
jameslamb, line 146:
```suggestion
          # Check whether matrix is the correct type first ("double")
          if (storage.mode(data) != "double") {
```

align comment start with the code it describes

R-package/tests/testthat/test_Predictor.R
```diff
@@ -0,0 +1,17 @@
...
+test_that("predictions do not fail for integer input", {
+    X <- as.matrix(as.integer(iris[, "Species"]), ncol = 1L)
+    y <- iris[["Sepal.Length"]]
+    dtrain <- lgb.Dataset(X, label = y)
+    fit <- lgb.train(
+        data = dtrain
+        , objective = "regression"
+        , verbose = -1L
+    )
+    X_double <- X_integer <- X[c(1L, 51L, 101L), , drop = FALSE]
```
jameslamb, line 12: do you mind breaking this up into two lines please? We don't use double-`<-` anywhere else in the codebase, as I think it's a little difficult to read.

```r
X_double <- X[c(1L, 51L, 101L), , drop = FALSE]
X_integer <- X_double
```

jameslamb:
thanks for the contribution and for using `LightGBM`, @mayer79 !

PR #3143, opened 2020-06-03
Windows GPU wheels

### LINKED ISSUE #2263: Suggestion: Also create a lightgbm-gpu release
Hi,
what about the idear that you could also create a lightgbm-gpu package? Like tensorflow also does?
Can you do that?
Thanks
Philip

itamarst:
Fixes #2263.

This builds on top of #3144, so that'll need to get merged first. Once that is merged, it will be possible to create a single wheel on Windows that work with both CPU and GPU devices. This PR then changes the CI build process to actually use that functionality, so that the standard Windows wheels will work on GPU devices too.

Unlike my previous attempt, the same wheel should work on both GPU and CPU, which makes life easier for everyone, both maintainers (you) and users.

* [x] Test it works on my computer.
* [x] Update wheel code so it can support the new `--opencl-python-package` option.
* [x] Have CI use that option.
* [ ] Get CI tests passing with CPU.
* [x] Run CI tests both with CPU device and GPU device.
* [ ] Get CI tests passing with GPU.
* [ ] Review comments, if any.


mirekphd:
> 
>  it will be possible to create a single wheel on Windows that work with both CPU and GPU devices. 
I realize that Linux world is more diverse, but catboost python library devs manage to achieve this goal also for Linux. It saves users a lot of time... (At the cost of pinning to some Cuda build - ideally there should be versions built for Cuda 8, 9, 10, and 11).

jameslamb:
Whenever you have time, can you please merge in `master`?  Some changes have been made to our CI jobs to fix failures there. I apologize for the inconvenience.

PR #3159, opened 2020-06-11
[WIP][java][mmlspark] speed up inference by ~20% in java layer

imatiach-msft:
Speed up inference by returning java array from LGBM_BoosterPredictForMatSingle and LGBM_BoosterPredictForCSRSingle.  This reduces JNI calls when getting the values from the array and also removes one step to allocate the output array in the Java side, removing another JNI call.

Testing in mmlspark, time was reduced from 82159416923ns to 65852779813ns, ~20% improvement.

FYI @eisber

### DIFF
--- a/swig/lightgbmlib.i
+++ b/swig/lightgbmlib.i
@@ -74,41 +74,48 @@
     return dst;
   }
 
-  int LGBM_BoosterPredictForMatSingle(JNIEnv *jenv,
-                                      jdoubleArray data,
-                                      BoosterHandle handle,
-                                      int data_type,
-                                      int ncol,
-                                      int is_row_major,
-                                      int predict_type,
-                                      int num_iteration,
-                                      const char* parameter,
-                                      int64_t* out_len,
-                                      double* out_result) {
+  jdoubleArray LGBM_BoosterPredictForMatSingle(JNIEnv *jenv,
+                                               jdoubleArray data,
+                                               BoosterHandle handle,
+                                               int data_type,
+                                               int ncol,
+                                               int is_row_major,
+                                               int predict_type,
+                                               int num_iteration,
+                                               const char* parameter,
+                                               int64_t* out_len) {
     double* data0 = (double*)jenv->GetPrimitiveArrayCritical(data, 0);
+    // Note: we allocate the output array in the native side but return new java array
+    double* out_result = new double[*out_len];
 
     int ret = LGBM_BoosterPredictForMatSingleRow(handle, data0, data_type, ncol, is_row_major, predict_type,
                                                  num_iteration, parameter, out_len, out_result);
 
     jenv->ReleasePrimitiveArrayCritical(data, data0, JNI_ABORT);
+    
+    if (ret != 0) {
+      return nullptr;
+    }
 
-    return ret;
+    jdoubleArray new_array = jenv->NewDoubleArray(*out_len);
+    jenv->SetDoubleArrayRegion(new_array, 0, *out_len, out_result);
+    delete[] out_result;
+    return new_array;
   }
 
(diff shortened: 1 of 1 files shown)


swig/lightgbmlib.i
```diff
@@ -74,41 +74,48 @@
...
+    // Note: we allocate the output array in the native side but return new java array
+    double* out_result = new double[*out_len];
 
     int ret = LGBM_BoosterPredictForMatSingleRow(handle, data0, data_type, ncol, is_row_major, predict_type,
                                                  num_iteration, parameter, out_len, out_result);
 
     jenv->ReleasePrimitiveArrayCritical(data, data0, JNI_ABORT);
+    
+    if (ret != 0) {
+      return nullptr;
```
imatiach-msft, line 97: looks like I need to delete before returning here

delete[] out_result;

swig/lightgbmlib.i
```diff
@@ -74,41 +74,48 @@
...
+                                               int data_type,
+                                               int ncol,
+                                               int is_row_major,
+                                               int predict_type,
+                                               int num_iteration,
+                                               const char* parameter,
+                                               int64_t* out_len) {
     double* data0 = (double*)jenv->GetPrimitiveArrayCritical(data, 0);
+    // Note: we allocate the output array in the native side but return new java array
+    double* out_result = new double[*out_len];
```
eisber, line 89: can you use 

```
vector<double> data0;
data0.resize(*out_len);
```

this will get you RAII behavior for errors (exceptions,...)

swig/lightgbmlib.i
eisber: can you use 

```
vector<double> data0;
data0.resize(*out_len);
```

this will get you RAII behavior for errors (exceptions,...)

swig/lightgbmlib.i
```diff
@@ -127,7 +135,13 @@
     jenv->ReleasePrimitiveArrayCritical(values, values0, JNI_ABORT);
     jenv->ReleasePrimitiveArrayCritical(indices, indices0, JNI_ABORT);
 
-    return ret;
+    if (ret != 0) {
+      return nullptr;
+    }
+    jdoubleArray new_array = jenv->NewDoubleArray(*out_len);
```
eisber, line 141: does this allocate memory underneath? I'm confused by doc :)

imatiach-msft: yes, it should return a new double array

imatiach-msft: I am worried about allocating both a new array on native side and also on java side, maybe I should stick to reusing the same native array (and passing it in as arg?) but returning a new java array

swig/lightgbmlib.i
imatiach-msft: great idea

AlbertoEAF:
One question @imatiach-msft - why not reuse the array instead of allocating a new one each time?

I have another extremely fast  high-throughput low-latency implementation that used this functions and instead reuses the array *out_result for every call, meaning you only have to read the values, and no allocations are necessary. Isn't this new change going to make it slower than that solution?

imatiach-msft:
@AlbertoEAF great point, I am investigating that as well, note this is still WIP.  Sorry haven't had more time to look into this lately.  I did discover an issue in how I was calculating the speedup, so this isn't quite ready yet.

jameslamb:
Whenever you have time, can you please merge in `master`?  Some changes have been made to our CI jobs to fix failures there. I apologize for the inconvenience.

guolinke:
@imatiach-msft 
What is the status of this PR?
do you have a plan for it?

imatiach-msft:
@guolinke yes, I will work on it again, let me close it for now

PR #3171, opened 2020-06-16
Adding static library option

dpayne:
This adds a cmake option to build the library statically which is useful when linking into a static executable. The option is disabled by default.


StrikerRUS:
I guess this piece of code should be updated as well:
https://github.com/microsoft/LightGBM/blob/e83042f20633d7f74dda0d18624721447a610c8b/CMakeLists.txt#L216-L220

dpayne:
I believe the shared library suffix code should be fine as is since I don't think CMAKE_SHARED_LIBRARY_SUFFIX is used explicitly anywhere.

StrikerRUS:
@dpayne I mean, `APPLE_OUTPUT_DYLIB` should affect suffix for **static** library, shouldn't it?

dpayne:
Apple should only be using DYLIB when building the shared library. I know this isn't definitive but at least on my local mac book I get
```[100%] Linking CXX static library ../lib_lightgbm.a```
The suffix there is that static lib suffix.

jameslamb:
@dpayne I apologize for the inconvenience, but could you please merge the most recent changes from `master` into this branch?

We have had some recent changes to our continuous integration setup that you'll need for this PR to pass checks.

Thanks for using and contributing to `LightGBM`!

pianoboysai:
very cool, static library is necessary !

PR #3175, opened 2020-06-18
[ci] update Boost version

StrikerRUS:


jameslamb:
I think @huanzhang12 is a better review for this PR

PR #3176, opened 2020-06-19
[ci][docs] fix broken links

StrikerRUS:


jameslamb:
I see that a `libsvm` link was broken, but it seems to still be there:

image

restarting the `check-docs` task

jameslamb:
> restarting the `check-docs` task

it's working!!

PR #3179, opened 2020-06-22
[ci][docs] added GitHub Actions badge and simplified condition in config

StrikerRUS:


.github/workflows/main.yml
```diff
@@ -99,12 +99,12 @@ jobs:
...
           $GITHUB_WORKSPACE/.ci/test.sh
       - name: Use conda on Windows
-        if: matrix.os == 'windows-latest' || matrix.os == 'windows-2016' || matrix.os == 'windows-2019'
+        if: startsWith(matrix.os, 'windows')
         uses: goanpeca/setup-miniconda@v1
         with:
           auto-update-conda: false
       - name: Setup and run tests on Windows
-        if: matrix.os == 'windows-latest' || matrix.os == 'windows-2016' || matrix.os == 'windows-2019'
+        if: startsWith(matrix.os, 'windows')
```
jameslamb, line 107: oh I didn't know you could do `startsWith()`! awesome

StrikerRUS: There are a lot of functions, btw: https://help.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#functions

PR #3168, opened 2020-06-16
[ci] Move R Windows CI to GitHub Actions

jameslamb:
This PR addresses https://github.com/microsoft/LightGBM/pull/3065#pullrequestreview-429217760, moving all Windows R CI jobs to GitHub Actions.

##  Why this is important

This should reduce the total runtime of CI for this project. GitHub Actions allows 20 concurrent builds...AppVeyor (which we currently use) allows 1 😬 

While testing [on my fork](https://github.com/jameslamb/LightGBM/actions/runs/136691678), I found that thanks to this concurrency, all 12 jobs for R run in 12-13 minutes!!!

image

As of this PR, we will have four R Windows jobs all in GitHub Actions:

* R 4.0 + `MSYS2` toolchain
* R 4.0 + Visual Studio 2019
* R 3.6 + `MINGW` toolchain
* R 3.6 + Visual Studio 2017

You can see the full details of software versions for the GitHub Actions environments [here](https://github.com/actions/virtual-environments).

## What I Changed

We didn't move R jobs into GitHub Actions in #3119 because of the issues with Powershell jobs on GitHub Actions (https://github.community/t/powershell-steps-fail-nondeterministically/115496) / Powershell itself (https://github.com/PowerShell/PowerShell/issues/12823).

`R`, `Rscript`, and `initexmf` all produce harmless messages to `stderr`, which could trigger the issues linked above. I tried many approaches to fix this, including:

* redirecting to `$null`
* setting `$ErrorActionPreference = "SilentlyContinue"`
* using `-ErrorAction` flag
* piping to `Out-Null` with `|`
* piping to `Out-String` with `|`
* redirecting to a file

None of these approaches worked, but the one in this PR did: calling these commands with `R` and using R's built-in redirection, [`sink()`](https://www.rdocumentation.org/packages/base/versions/3.6.2/topics/sink).

PR #3181, opened 2020-06-23
[R-package] fix R CMD check NOTE about leftover files

jameslamb:
This fixes the following NOTE from `R CMD check` on Windows:

> checking for non-standard things in the check directory ... NOTE
Found the following files/directories:
'data.bin' 'lgb-Dataset.data' 'lgb-model.rds' 'lgb-model.txt'
  'lgb.Dataset.data' 'model.rds' 'model.txt'

It fixes this by only using `tempfile()` files everywhere that examples and tests write to the file system.

It also cleans up the logs in Windows R jobs. As of #3168 , I used `{processx}` to run some commands and redirect `stderr`, to avoid transient failures.

That produces hard-to-read logs like this:

image

This PR resolves that by assigning the result of `processx::run()` and setting `echo = TRUE` in those calls too stream out the logs.


jameslamb:
it looks like this did eliminate the NOTE! https://github.com/microsoft/LightGBM/pull/3181/checks?check_run_id=797820357

PR #3193, opened 2020-06-27
[R-package] [ci] update to R 4.0.2 for Windows CI jobs (fixes #3191)

### LINKED ISSUE #3191: [R-package] Windows R 4.0 jobs fail after "Building R Package"
**Error on Windows + MSVC + R 4.0 job**

See for example https://github.com/microsoft/LightGBM/pull/3000/checks?check_run_id=811619463

image

Some R CI jobs on GitHub Actions are failing at this step:

```
The downloaded binary packages are in
	C:\Users\runneradmin\AppData\Local\Temp\Rtmpi08q9C\downloaded_packages
Building R package
##[error]Process completed with exit code 1.
```

**Error on Windows + MINGW + R4.0 job**

See for example https://github.com/microsoft/LightGBM/pull/3000/checks?check_run_id=811619438

image

jameslamb:
This PR attempts to fix #3191 , and issue that is blocking other PRs. I'm not sure why, but updating to R 4.0.2 and changing the output redirection strategy from builds seems to be enough to fix the issues documented there.

The R 4.0.2 release notes are here, if you're curious: https://cran.r-project.org/doc/manuals/r-devel/NEWS.html

I tried this on my fork, and got 3 successful builds of R 4.0 + MSVC and 3 successful builds of R 4.0 + MINGW: https://github.com/jameslamb/LightGBM/pull/32/checks?check_run_id=813328837

### DIFF
--- a/.ci/test_r_package_windows.ps1
+++ b/.ci/test_r_package_windows.ps1
@@ -165,7 +165,7 @@ if ($env:COMPILER -ne "MSVC") {
 } else {
   $env:TMPDIR = $env:USERPROFILE  # to avoid warnings about incremental builds inside a temp directory
   $INSTALL_LOG_FILE_NAME = "$env:BUILD_SOURCESDIRECTORY\00install_out.txt"
-  Rscript build_r.R
+  #Rscript build_r.R
   Run-R-Code-Redirect-Stderr "source('build_r.R')" 1> $INSTALL_LOG_FILE_NAME ; $install_succeeded = $?
   Write-Output "----- build and install logs -----"
   Get-Content -Path "$INSTALL_LOG_FILE_NAME"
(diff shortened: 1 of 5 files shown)


.ci/test_r_package_windows.ps1
```diff
@@ -165,7 +165,7 @@ if ($env:COMPILER -ne "MSVC") {
 } else {
   $env:TMPDIR = $env:USERPROFILE  # to avoid warnings about incremental builds inside a temp directory
   $INSTALL_LOG_FILE_NAME = "$env:BUILD_SOURCESDIRECTORY\00install_out.txt"
-  Run-R-Code-Redirect-Stderr "source('build_r.R')" *> $INSTALL_LOG_FILE_NAME ; $install_succeeded = $?
+  Run-R-Code-Redirect-Stderr "source('build_r.R')" 1> $INSTALL_LOG_FILE_NAME ; $install_succeeded = $?
```
StrikerRUS, line 168: Oh, wait, does it mean that we lose all not success logs that are written not to stdout?

jameslamb: We already lost any of them by the changes from #3168 , I think. This step will still break if something is broken, but to get better logs  we'd have to re-run with this set to `Rscript build_r.R` instead of using `Run-R-Code-Redirect-Stderr`.

I know that's bad, but the main thing is that we know the build will fail if a chance breaks the R package. I hope that https://github.com/PowerShell/PowerShell/issues/12823 will get resolved and GitHub Actions will be updated soon, then maybe we can remove the workaround to suppress stderr

PR #3196, opened 2020-06-29
Store the true split gain (without subtracting min_split_gain) in tree models (fix Issue #3054)

### LINKED ISSUE #3054: Issue with min_gain_to_split and how to calculate weight as shown in create_tree_digraph
I am seeing the following, when min_gain_to_spilt is specified I can see some splits are created with gain < the value. I created similar issue some time back, could not reproduce the error that time in a dummy data.
```
import pandas as panda
import numpy as numpy
from sklearn.metrics import roc_auc_score,log_loss

np.random.seed(42)
x=pd.DataFrame({"y":[int(np.random.random()>.75) for i in range(100)],
               "x1":[np.round(np.random.random(),1) for i in range(100)],"x2":[np.round(np.random.random(),1) for i in range(100)]})
x["weight"]=1
import lightgbm as lgb

valid_features=["x1","x2"]

lgb_train_data = lgb.Dataset(x[valid_features], label=x.y.values, 
                                  weight=x.weight.values,
                                  free_raw_data=False)
parameters = {
    'num_leaves': 10,
"boosting_type": 'gbdt', 
"objective" : 'binary',
    "min_data":2,
     'min_child_weight': 0,
    "random_state":2,"boost_from_average":False,"min_gain_to_split":0.4
}
print("running ")
model2 = lgb.train(parameters,
lgb_train_data,
num_boost_round=20,valid_names=['train'])

lgb.create_tree_digraph(model2,show_info=['split_gain', 'internal_value', 'internal_count', 'internal_weight',
    'leaf_count', 'leaf_weight', 'data_percentage'],tree_index=7)

```

Screenshot 2020-05-06 at 6 08 17 PM

I need help in an another issue, how can I can calculate the "weight" as in create_tree_digraph? I understand it is min_child_weight or sum of weighted hessian. However not able to match the same.
I am testing for root node 1 of tree_index=0.

```
x["lpred"]=np.where(x.x1<.15,-0.171,-0.93)
x["pred"]=np.exp(x["lpred"])/(1+np.exp(x["lpred"]))
x["hess"]=2*x["pred"]*(1-x["pred"])*(x["weight"])
print(x[x.x1>.15].hess.sum())
lgb.create_tree_digraph(model2,show_info=['split_gain', 'internal_value', 'internal_count', 'internal_weight',
    'leaf_count', 'leaf_weight', 'data_percentage'],tree_index=0)
```

Screenshot 2020-05-06 at 6 21 06 PM

Any help is really appreciated .

shiyu1994:
This PR is to fix #3054 .
The split_gain_ in Tree is only used by feature importance, apart from being saved into models.  

> https://github.com/microsoft/LightGBM/blob/9f367d11ad52ae8f2f1cf7b3447735723281d820/src/boosting/gbdt_model_text.cpp#L587-L621

Storing the true split gain in trees will also change the feature importance value. Feature importance values without subtracting the min_split_gain should be more meaningful, which reflects the true loss reduction by splitting with the feature.

BTW, is the comparing of split gains with 0 necessary in the code above? I thought split gains in tree models should always be >= 0, even after subtracting min_split_gain.

### DIFF
--- a/src/treelearner/serial_tree_learner.cpp
+++ b/src/treelearner/serial_tree_learner.cpp
@@ -577,7 +577,8 @@ void SerialTreeLearner::SplitInner(Tree* tree, int best_leaf, int* left_leaf,
         static_cast<data_size_t>(best_split_info.right_count),
         static_cast<double>(best_split_info.left_sum_hessian),
         static_cast<double>(best_split_info.right_sum_hessian),
-        static_cast<float>(best_split_info.gain),
+        // store the true split gain in tree model
+        static_cast<float>(best_split_info.gain + config_->min_gain_to_split),
         train_data_->FeatureBinMapper(inner_feature_index)->missing_type(),
         best_split_info.default_left);
   } else {
@@ -613,7 +614,8 @@ void SerialTreeLearner::SplitInner(Tree* tree, int best_leaf, int* left_leaf,
         static_cast<data_size_t>(best_split_info.right_count),
         static_cast<double>(best_split_info.left_sum_hessian),
         static_cast<double>(best_split_info.right_sum_hessian),
-        static_cast<float>(best_split_info.gain),
+        // store the true split gain in tree model
+        static_cast<float>(best_split_info.gain + config_->min_gain_to_split),
         train_data_->FeatureBinMapper(inner_feature_index)->missing_type());
   }
 


guolinke:
@shiyu1994 can you merge the latest master to pass CI ?

PR #3197, opened 2020-07-01
Allow the minimal feature to 1 in column sampling

guolinke:


guolinke:
@jameslamb the R's tests are not stable?

jameslamb:
> @jameslamb the R's tests are not stable?

ok this is very bad. @btrotta has been seeing these issues on #3136 too

image

The R tests rely on downloading some LaTeX things from a CTAN mirror at University of Illinois. I pegged it to one mirror because we were getting such inconsistent behavior from the main CTAN download links.

Let me try changing mirrors, sorry for the inconvenience.

jameslamb:
I've opened #3198 to investigate this, looking into it now.

jameslamb:
ok @guolinke I think #3199 will fix it!

PR #3199, opened 2020-07-01
[R-package] [ci] Fix MiKTeX downloads (fixes #3198)

### LINKED ISSUE #3198: [R-package] [ci] R CI jobs failed with to download MiKTeX
See https://github.com/microsoft/LightGBM/pull/3197#issuecomment-652137494 and comments on #3136 . R jobs are failing with an error like 

> Exception calling "DownloadFile" with "2" argument(s): "The remote server returned an error: (404) Not Found."

I have been able to reproduce this on my laptop

image

It's weird because the mirror is definitely up:

image

jameslamb:
See #3198 for details. Basically, R Windows jobs in CI today have to scrape a TeX archive page to find the name of a `miktexsetup` distribution to download. The scraping logic assumes that only one `miktexsetup` will be found there. That has been true for a while, but `miktexsetup` just had a major version bump (from 2.10.x to 4.0.0).

This PR fixes our CI by looking for a specific version of `miktexsetup`.

### DIFF
--- a/.ci/test_r_package_windows.ps1
+++ b/.ci/test_r_package_windows.ps1
@@ -24,7 +24,7 @@ function Download-Miktex-Setup {
         [string]$destfile
     )
     $PageContent = Invoke-WebRequest -Uri $archive -Method Get
-    $SetupExeFile = $PageContent.Links.href | Select-String -Pattern 'miktexsetup.*'
+    $SetupExeFile = $PageContent.Links.href | Select-String -Pattern 'miktexsetup-4.*'
     $FileToDownload = "${archive}/${SetupExeFile}"
     Download-File-With-Retries $FileToDownload $destfile
 }
--- a/.ci/test_r_package_windows.ps1
+++ b/.ci/test_r_package_windows.ps1
@@ -24,7 +24,7 @@ function Download-Miktex-Setup {
         [string]$destfile
     )
     $PageContent = Invoke-WebRequest -Uri $archive -Method Get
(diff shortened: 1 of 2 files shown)


jameslamb:
Ok I tried getting version 4.0 but it wasn't working (maybe it's too new?)

image

So for now, we should peg to 2.10.x and then we'll have to update again when that gets removed :/

StrikerRUS:
Maybe the issue is here?
```

Downloading https://ctan.math.illinois.edu/systems/win32/miktex/setup/windows-x64//miktexsetup-2.9.7442-x64.zip
                                                                   --------------^--------------
```

jameslamb:
> Maybe the issue is here?
> 
> ```
> 
> Downloading https://ctan.math.illinois.edu/systems/win32/miktex/setup/windows-x64//miktexsetup-2.9.7442-x64.zip
>                                                                    --------------^--------------
> ```

I don't think so. That link worked in `pwsh` on my laptop. and if you look at the most recent logs, you'll see `miktexsetup` was successfully downloaded but it's failing when it is actually run

image

That 503 means that the package server was unavailable. 

From http://dante.ctan.org/mirmon/ it looks like that Illinois mirrored was refreshed 7 hours ago, so I just tried re-running the jobs. Maybe the breaking version changes have caused issues in mirrors updating. If re-running doesn't work, I'll try picking a different mirror.

jameslamb:
All of the checks passed after re-running. I think this is another issue where it just took a day for changes to propagate to all mirrors.

Issue #2105, opened 2019-04-15
Why is _CVBooster object hidden class?


matsuken92:
Some of Kaggler use out-of-fold prediction (oof) which is calculated with validation dataset for stacking, blending, evaluation etc. In order to  calculate oof with lgb.cv(), we need to return Boosters however it is not returned from the function. If I customize to return `_CVBooster` object, it can be calculated, but `_CVBooster` is hidden class, so I think it is not appropriate to return officially. 

Is it possible to remove prefix underscore from  `_CVBooster`? Then I can customize to return the object on lgb.cv().

I would appreciate it if you could advise to it.

### _CVBooster
https://github.com/Microsoft/LightGBM/blob/master/python-package/lightgbm/engine.py#L248-L267

### Example
https://www.kaggle.com/kenmatsu4/using-trained-booster-from-lightgbm-cv-w-callback

guolinke:
ping @StrikerRUS

matsuken92:
Thank you for your comment!  Minimum requirements are:
  * Returning list of `Boosters`
  * Returning `best_iteration` to identify which iteration is best for predict when using early stopping.

In addition to above, current `_CVBooster.__getattr__` function is useful for batch operation of boosters on each fold. The following link is the part of usage of the `_CVBooster.__getattr__` calculating prediction result for test data.

https://github.com/matsuken92/LightGBM_test/blob/master/src/cv_booster_test.py#L117

I will send my idea as PR, so I would appreciate it if you could review it then!

StrikerRUS:
@matsuken92 I see! Thanks for the clarification!

Of course, I'll be happy to review it.

Can you please also think about backward compatibility in terms of `cv` function return type? Maybe a new key in results dict or something else, idk...

StrikerRUS:
Closing in favor of being in #2302. We decided to keep all feature requests in one place.

Welcome to contribute this feature! Please re-open (or post a comment if you are not a topic starter) this issue if you are actively working on implementing this feature.

StrikerRUS:
Closed via #3204.

Issue #1445, opened 2018-06-12
Lightgbm cv feature importance python 


MotoRZR:
It would be useful for the cv function to retain mean/std of feature importances across the cv folds.  I don’t believe this is currently implemented.

guolinke:
you can do this by yourself simply, just get Booster from cv folds, and then calculate mean/std.

MotoRZR:
Hi, thanks for the quick response.  The LightGBM native CV function I believe just returns a dictionary of results rather than the individual booster parameters.  Or am I missing something.

guolinke:
@StrikerRUS can we return the `cvfolds` in lgb.cv as well ?

StrikerRUS:
@guolinke Did you mean `return dict(results), cvfolds` tuple instead of
https://github.com/Microsoft/LightGBM/blob/3f401477872813a81ca98e90b9b42d085b1013d2/python-package/lightgbm/engine.py#L469

It'll be breaking changes. Maybe push `cvfolds` into the dict to save the current method signature?

aerdem4:
@guolinke @StrikerRUS It would be nice to have an "ensemble of models" object. The benefits of having such a class is to return importances and predictions as an average. Currently, especially on Kaggle, people always do the cross-validation and ensembling at the same time. cv function can either return all generated models or out of sample predictions and average ensembled predictons for the test set if given. If you think that this feature is nice to have, I would like to help.

JoshuaC3:
@StrikerRUS could we not do something like,

```
if return_feature_importances:
    return dict(results), dict(agg_feaure_importances)
else:
    return dict(results)
```

with default as `None` or `False`. Default behaviour then doesn't break.

StrikerRUS:
@JoshuaC3 I suppose it's possible. Would you mind to create a PR?

StrikerRUS:
Closed in favor of being in #2302. We decided to keep all feature requests in one place.

Welcome to contribute this feature! Please re-open this issue (or post a comment if you are not a topic starter) if you are actively working on implementing this feature.

StrikerRUS:
Implemented in #3204.
Now it is possible to get feature importances from `CVBooster`.

Issue #283, opened 2017-02-05
Add option to keep cv predicted values


JackStat:
Forgive me if I missed something but I have review the code and documentation and didn't see a way to keep the cv probabilities.

guolinke:
@JackStat 
I think a simple solution is saving all cv models. Then get the predictions by these models.
Welcome to contribute for this, I think it is easy to implement.

JackStat:
Absolutely.  I looked through the code and thought that would be a good strategy as well but I could not find an object that had the holdout data.frame.

So it looks like this chunk creates the 3 boosters (assuming 3-fold cv) 

  ```
# construct booster
  bst_folds <- lapply(seq_along(folds), function(k) {
    dtest   <- slice(data, folds[[k]])
    dtrain  <- slice(data, unlist(folds[-k]))
    booster <- Booster$new(params, dtrain)
    booster$add_valid(dtest, "valid")
    list(booster = booster)
  })
```

Then you can run something with `lapply` to get predictions from each booster using `bst_folds[[1]]$booster$predict`  Now I just need to know where the cv data.frames are kept so I can apply the predictions to those.  I dug into the objects and couldn't see them.

Any help would be appreciated and I will open the pull req.
Thanks

**Comment by yanyachen** [comment]
@guolinke I looked through the code and I found that predicting form lgb.Dataset hasn't been supported yet. Could you support that when you got time? Otherwise we can not use all cv models to predict on each fold.

Below is a simple function that generating cv predictions from original dataset, @JackStat you can use that for your problem, though I think you had figured it out by yourself.
```
LGB_CV_Predict <- function(lgb_cv, data, num_iteration = NULL, folds) {
  if (is.null(num_iteration)) {
    num_iteration <- lgb_cv$best_iter
  }
  cv_pred_mat <- foreach::foreach(i = seq_along(lgb_cv$boosters), .combine = "rbind") %do% {
    lgb_tree <- lgb_cv$boosters[[i]][[1]]
    predict(lgb_tree, 
            data[folds[[i]],], 
            num_iteration = num_iteration, 
            rawscore = FALSE, predleaf = FALSE, header = FALSE, reshape = TRUE)
  }
  if (ncol(cv_pred_mat) == 1) {
    as.double(cv_pred_mat)[order(unlist(folds))]
  } else {
    cv_pred_mat[order(unlist(folds)), , drop = FALSE]
  }
}
```

guolinke:
@yanyachen 
[...]

PR #3204, opened 2020-07-03
[python] add return_cvbooster flag to cv func and publish _CVBooster (#283,#2105,#1445)

momijiame:
This PR allows users to get trained boosters from cv function directly.
This feature is useful for using ensemble techniques and OOF predictions.

Add the changes that proposed in #283 discussion.

- Add ``return_cvbooster`` flag to cv function
- Rename ``_CVBooster`` to make public


momijiame:
Hmm... At a glance, I think this change did not trigger breaking CI.

`/__w/1/s/.ci/test.sh: line 103: pytest: command not found`
https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=6577&view=logs&j=02a2c3ba-81f8-54e3-0767-5d5adbb0daa9&t=720ee3fa-96d4-5b47-dbf4-01607b74ade2&l=4772

StrikerRUS:
> Hmm... At a glance, I think this change did not trigger breaking CI.

Random network error. Fixed. Sorry for the inconvenience.

StrikerRUS:
@momijiame Please add new public class to the documentation: https://github.com/microsoft/LightGBM/blob/master/docs/Python-API.rst.

@matsuken92 Can you please help to review this PR?

momijiame:
@StrikerRUS Thank you quick response! I added the CVBooster to the documentation.

python-package/lightgbm/engine.py
```diff
@@ -276,19 +276,26 @@ def train(params, train_set, num_boost_round=100,
     return booster
 
 
-class _CVBooster(object):
-    """Auxiliary data struct to hold all boosters of CV."""
+class CVBooster(object):
+    """CVBooster in LightGBM.
+
+    Auxiliary data struct to hold and redirect all boosters of CV.
```
StrikerRUS, line 282: Add some words about how one can use this class. We should document that this class has the same methods (members?) as `Booster` class. All method calls are actually performed for underlying Boosters and then all returned results are returned in a list.

tests/python_package_test/test_engine.py
```diff
@@ -735,6 +735,47 @@ def test_cv(self):
...
+        params = {
+            'objective': 'binary',
+            'metric': 'binary_logloss',
+            'verbose': -1,
+        }
+        lgb_train = lgb.Dataset(X_train, y_train)
+        # with early stopping
+        cv_res = lgb.cv(params, lgb_train,
+                        num_boost_round=100,
+                        early_stopping_rounds=10,
```
StrikerRUS, line 750: Decrease values to save CI time.

tests/python_package_test/test_engine.py
```diff
@@ -735,6 +735,47 @@ def test_cv(self):
...
+                        verbose_eval=False,
+                        nfold=3,
+                        return_cvbooster=True)
+        self.assertIn('cvbooster', cv_res)
+        cvb = cv_res['cvbooster']
+        self.assertIsInstance(cvb, lgb.engine.CVBooster)
```
StrikerRUS, line 756:
```suggestion
        self.assertIsInstance(cvb, lgb.CVBooster)
```
(Will be available after adding this class into `__all__`.)

momijiame:
@StrikerRUS Thank you so much for the feedback! I reflected the comments.

python-package/lightgbm/engine.py
```diff
@@ -276,19 +276,28 @@ def train(params, train_set, num_boost_round=100,
...
```
StrikerRUS, line 282:
```suggestion
    Auxiliary data structure to hold and redirect all boosters of ``cv`` function.
```

python-package/lightgbm/engine.py
```diff
@@ -276,19 +276,28 @@ def train(params, train_set, num_boost_round=100,
...
+    This class has the same methods as Booster class.
+    All method calls are actually performed for underlying Boosters and then all returned results are returned in a list.
+    """
```
StrikerRUS, line 285: Document available class attributes.

```suggestion
    All method calls are actually performed for underlying Boosters and then all returned results are returned in a list.

    Attributes
    ----------
    boosters : list of Booster
        The list of underlying fitted models.
    best_iteration : int
        The best iteration of fitted model.
    """
```

tests/python_package_test/test_engine.py
```diff
@@ -735,6 +735,47 @@ def test_cv(self):
...
+        # with early stopping
+        cv_res = lgb.cv(params, lgb_train,
+                        num_boost_round=25,
+                        early_stopping_rounds=5,
+                        verbose_eval=False,
+                        nfold=3,
+                        return_cvbooster=True)
+        self.assertIn('cvbooster', cv_res)
+        cvb = cv_res['cvbooster']
+        self.assertIsInstance(cvb, lgb.CVBooster)
```
StrikerRUS, line 756: More checks.

```suggestion
        self.assertIsInstance(cvb, lgb.CVBooster)
        self.assertIsInstance(cvb.boosters, list)
        self.assertEqual(len(cvb.boosters), 3)
        self.assertTrue(all(isinstance(bst, lgb.Booster) for bst in cvb.boosters))
```

momijiame:
Thank you for the great suggestion. I committed.

StrikerRUS:
Gently ping @matsuken92 as you wanted to give your review for the general idea of this PR.

momijiame:
I noticed that this PR also solves #1445, so I updated the title.
Users can extract feature importances from fitted Boosters.

NOTE: It would be more useful to support `CVBooster` and/or `List[Booster]` for `plot_importance()`

mirekphd:
@momijiame : thank you for this much needed feature!
I seem to have found a rather important omission: did we ensure `best_iteration` is returned for each fold from `CVBooster` when using early stopping? 

After running `lgbm_cv_dict = lgbm.cv(params=params_dict, ..., return_cvbooster=True)`, with `params_dict` containing a custom `early_stopping_rounds` value,
`lgbm_cv_dict['cvbooster'].boosters[k].best_iteration` seems to be missing (equal to -1 for all folds), even though `lgbm_cv_dict['cvbooster'].best_iteration` is returned correctly (e.g. 123) (if would be -1 if early stopping rounds were not specified and early stopping were not used). 

For example:
```
Numbers of iterations (current and best) for re-fitted CV models :

Fold 0:
154
-1

Fold 1:
150
-1

Fold 2:
155
-1

Fold 3:
151
-1

Fold 4:
155
-1

CVBooster best iteration:
123
```

Without best iteration it is not possible to obtain predictions for each fold nor feature importances (which rely on the iterations number). In fact I suspect they may be incorrect now, because they are the same regardless of the value their `iteration` parameter is set to:

```
refit_lgbm_cv_ret_dict['cvbooster'].boosters[k].feature_importance()
# or
refit_lgbm_cv_ret_dict['cvbooster'].boosters[k].feature_importance(iteration=fold_best_iter_num)
# or
refit_lgbm_cv_ret_dict['cvbooster'].boosters[k].feature_importance(iteration=fold_cur_iter_num)

# where:
fold_best_iter_num=refit_lgbm_cv_ret_dict['cvbooster'].boosters[k].best_iteration
fold_cur_iter_num=refit_lgbm_cv_ret_dict['cvbooster'].boosters[k].current_iteration()
```

momijiame:
@mirekphd Thank you for your feedback.
I do not know if it is documented or not, but it is expected behavior. From before, cv() function updates Boosters while internally synchronizing all folds. early_stopping is triggered by the average metric of all folds. Therefore, best_iteration is equal in all folds. IMHO, this behavior makes cross validation conservative.

We could consider setting best_iteration of Boosters to all the same value, but that would break backwards compatibility. Even before return_cvbooster option was added, the technique of retrieving CVBooster (_CVBooster) by using Callback was exist.

Also, if we want to train Boosters on each folds, we can simply use train() function. The documentation could be improved on the above, but I do not think the behavior should be changed.

mirekphd:
> Therefore, best_iteration is equal in all folds.

Not in this case:). Please note that we are talking only about **early stopping** here, where `best_iteration` is unknown until each fold model completes. Only then can you compute average best iteration across all folds and return it as CVBooster's "best iteration". 

To see that each fold uses a different number of iterations its enough to see that we still have only one training pass per model (as opposed to two passes), and can be also seen from my example above, where the training stopped at iterations ranging from 150 to 155 rounds, as reported by `.current_iteration()` for each fold.

The information is already available somewhere, we just need to expose it to the user, to replace of those -1 values, just like in case of `.current_iteration()`. 

Unless of course ligthgbm makes two modeling passes and uses average iterations number from the first pass in the final pass (which would be rather inefficient IMO:)

guolinke:
@mirekphd 
I think in both XGBoost and LightGBM, the CV will use the average scores from all folds, and use this for the early stopping. Therefore, best_iteration is the same in all folds.
I think this is more stable since the average_score is computed over all data samples, not just over the current fold.

PR #3205, opened 2020-07-04
[ci] [R-package] catch builds that have not updated docs

jameslamb:
From @StrikerRUS 's suggestion in https://github.com/microsoft/LightGBM/pull/3188#discussion_r449203929, this pull request proposes adding a new R CI jobs called `check-r-docs`. This job regenerates the documentation for the R package and checks if there are any changes that have not been committed to the repo.

In R, the process for builds the docs in the `R-package/man/` folder is like this:

1. change files in `R-package/R/`
2. install the liibrary
3. run `roxygen2::roxygenize(load = 'installed')`

Because this process relies on installing the library, I think it makes sense to add into `./.ci/test_r_package.sh`. It requires installing one additional somewhat-expensive dependency (`roxygen2`) and the documentation is not dependent on operating system so I think it makes sense to do this in a single new TASK instead of running it in all of the R CI jobs.

### DIFF
--- a/.ci/test.sh
+++ b/.ci/test.sh
@@ -69,7 +69,7 @@ if [[ $TASK == "if-else" ]]; then
     exit 0
 fi
 
-if [[ $TASK == "r-package" ]]; then
+if [[ $TASK == "r-package" ]] || [[ $TASK == "check-r-docs" ]]; then
     bash ${BUILD_DIRECTORY}/.ci/test_r_package.sh || exit -1
     exit 0
 fi
--- a/.ci/test.sh
+++ b/.ci/test.sh
@@ -69,7 +69,7 @@ if [[ $TASK == "if-else" ]]; then
     exit 0
 fi
 
-if [[ $TASK == "r-package" ]] || [[ $TASK == "check-r-docs" ]]; then
+if [[ "${TASK:0:9}" == "r-package" ]]; then
     bash ${BUILD_DIRECTORY}/.ci/test_r_package.sh || exit -1
     exit 0
 fi
--- a/.ci/test_r_package.sh
+++ b/.ci/test_r_package.sh
@@ -4,7 +4,7 @@
 CRAN_MIRROR="https://cloud.r-project.org/"
 R_LIB_PATH=~/Rlib
 mkdir -p $R_LIB_PATH
-echo "R_LIBS=$R_LIB_PATH" > ${HOME}/.Renviron
+export R_LIBS=$R_LIB_PATH
 export PATH="$R_LIB_PATH/R/bin:$PATH"
 
 # Get details needed for installing R components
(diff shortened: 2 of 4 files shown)


R-package/R/lgb.importance.R
```diff
@@ -1,5 +1,5 @@
 #' @name lgb.importance
-#' @title Compute feature importance in a model
+#' @title Compute feature importance in a model (testing)
```
jameslamb, line 2: I added this as an example of the type of change that should trigger a failure. I'll remove it once we see if the new task works as expected.

jameslamb: it  looks like this worked as expected!

https://github.com/microsoft/LightGBM/pull/3205/checks?check_run_id=836077809

image

.ci/test_r_package.sh
```diff
@@ -4,7 +4,7 @@
 CRAN_MIRROR="https://cloud.r-project.org/"
 R_LIB_PATH=~/Rlib
 mkdir -p $R_LIB_PATH
-echo "R_LIBS=$R_LIB_PATH" > ${HOME}/.Renviron
+export R_LIBS=$R_LIB_PATH
```
jameslamb, line 7: I made a mistake while developing...code with `Rscript --vanilla -e` failed to load a library because `--vanilla` ignores the `.Renviron`

I think we should remove this and use normal environment variables instead, to make the code a little more explicit and easier to reason about.

.ci/test.sh
```diff
@@ -69,7 +69,7 @@ if [[ $TASK == "if-else" ]]; then
     exit 0
 fi
 
-if [[ $TASK == "r-package" ]]; then
+if [[ $TASK == "r-package" ]] || [[ $TASK == "check-r-docs" ]]; then
```
StrikerRUS, line 72: Just a suggestion to not overcomplicate this `if` statement in the future. Name task with something like `r-package-check-docs` and then just check that it starts with `r-package`.

.ci/test_r_package.sh
```diff
@@ -90,6 +90,24 @@ if [[ $OS_NAME == "macos" ]]; then
 fi
 Rscript --vanilla -e "install.packages(${packages}, repos = '${CRAN_MIRROR}', lib = '${R_LIB_PATH}', dependencies = c('Depends', 'Imports', 'LinkingTo'))" || exit -1
 
+if [[ $TASK == "check-r-docs" ]]; then
+    Rscript build_r.R
```
StrikerRUS, line 94: Maybe for fast-fail?
```suggestion
    Rscript build_r.R || exit -1
```

.ci/test.sh
jameslamb: oh good idea! Just pushed https://github.com/microsoft/LightGBM/pull/3205/commits/0b17315310b2cf2f97a19fb9c907dcd8eab4e86b to implement this

StrikerRUS:
@guolinke Please make new job required. Thank you!

PR #3208, opened 2020-07-04
[ci] bump gcc version for macOS

StrikerRUS:
Refer to https://github.com/Homebrew/homebrew-core/pull/54364#issuecomment-653448513.

Fix current MPI jobs failures.


.ci/setup.sh
```diff
@@ -9,6 +9,7 @@ if [[ $OS_NAME == "macos" ]]; then
     else  # gcc
         if [[ $TASK != "mpi" ]]; then
             brew install gcc
+            brew upgrade gcc
```
StrikerRUS, line 12: Can be reverted later after Travis updates.

PR #3211, opened 2020-07-07
[python] fix early_stopping_round = 0

### LINKED ISSUE #3203: Early stopping round = 0 doesn't disable early stop as described in the documentation
* Python package

LightGBM component: 2.3.1

## Environment info

Operating System: Ubuntu

CPU/GPU model:

C++ compiler version:

CMake version:

Java version:

Python version: 3.7

LightGBM version or commit hash: 2.3.1

In the Parameters documentation about [early_stopping_round](https://github.com/Microsoft/LightGBM/blob/master/docs/Parameters.rst#early_stopping_round), it states "<= 0 means disable". However, when using the python API and passing `early_stopping_round=0` in the `params` argument for train, the model actually stop at the first iteration that get worse instead of disabling the early stop.

In the source code for the python API, we can see that it only accepts `None` to disable early stop but not 0.
https://github.com/microsoft/LightGBM/blob/master/python-package/lightgbm/engine.py#L214-L215

## Reproducible example(s)

```python
import lightgbm as lgb
df = pd.DataFrame([[1, 1]]*100 + [[2, 2]]*100, columns=['x', 'y'])
lgb_train = lgb.Dataset(df[['x']], df['y'])
lgb_val = lgb.Dataset(df[['x']], -1*df['y'])
params = {'early_stopping_round': 0, 'metric': 'mse'}
lgb.train(params, lgb_train, valid_sets=[lgb_val])
```
You get
```
[1]	valid_0's l2: 9.3025
Training until validation scores don't improve for 0 rounds
Early stopping, best iteration is:
[1]	valid_0's l2: 9.3025
```

## Suggestion
Either fix the documentation or fix the python api to disable early stop when `early_stopping_round <= 0`.

guolinke:
to fix #3203
also, ping @jameslamb to check (and fix) the corresponding problem in R packages.

### DIFF
--- a/python-package/lightgbm/engine.py
+++ b/python-package/lightgbm/engine.py
@@ -211,7 +211,7 @@ def train(params, train_set, num_boost_round=100,
     elif isinstance(verbose_eval, integer_types):
         callbacks.add(callback.print_evaluation(verbose_eval))
 
-    if early_stopping_rounds is not None:
+    if early_stopping_rounds is not None and early_stopping_rounds > 0:
         callbacks.add(callback.early_stopping(early_stopping_rounds, first_metric_only, verbose=bool(verbose_eval)))
 
     if learning_rates is not None:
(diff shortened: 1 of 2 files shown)


StrikerRUS:
@guolinke Should that fix be applied here as well?
https://github.com/microsoft/LightGBM/blob/9f367d11ad52ae8f2f1cf7b3447735723281d820/python-package/lightgbm/sklearn.py#L605

Is the semantics for `<=0` like "you didn't provide early stopping"?

guolinke:
@StrikerRUS 
yeah, `<=0` is the same as `None`.

jameslamb:
> to fix #3203
> also, ping @jameslamb to check (and fix) the corresponding problem in R packages.

@guolinke do you want me to push directly to this PR? Or create a separate one

guolinke:
@jameslamb you can create the separate one.

Issue #3011, opened 2020-04-21
save_model() Feature Importance Section can use a different importance_type


little-eyes:
## Summary

Currently, the booster.save_model() function does not accept additional params to change the importance type when output the feature importance to the model file. The c++ layer is also hard coded to use 'split' as the importance_type as below.

https://github.com/microsoft/LightGBM/blob/9478e52f9fd751353d4e0e04899c2b090f1ff0fe/src/boosting/gbdt_model_text.cpp#L367

So it would be great to have an additional params to be able to change the behavior of feature importance, e.g. uses 'gain' as importance type.

## Motivation

We uses LightGBM in our production work, and we show some diagnostic plots, such as feature importance plot, to our customers. But when they inspect the model file, they found a totally different feature importance results than our plots because we use 'gain' as importance_type, while the model file uses 'split'. So it would be great to show consistency to our customers by using the same importance type.

## Description

Can we expose APIs like below:

```python
model.booster_.save_model(path, model.booster_.best_iteration, importance_type='gain')
```

StrikerRUS:
Refer to https://github.com/microsoft/LightGBM/pull/2656#issuecomment-570387134 and ping @guolinke .

StrikerRUS:
Closed via #3220.

PR #3220, opened 2020-07-12
feature importance type in saved model file

guolinke:


StrikerRUS:
@guolinke Don't you want to expose this parameter to C API? With 3.0 I guess it's good time to break some code (if any). Then wrappers will be able to provide this param directly in function like start/end iterations 

https://github.com/microsoft/LightGBM/blob/2792923d35e7f3d392474ceffe2185c661047b2e/python-package/lightgbm/basic.py#L2734-L2750

guolinke:
Thanks @StrikerRUS , I change all related functions.

R-package/R/lgb.Booster.R
```diff
@@ -424,7 +424,7 @@ Booster <- R6::R6Class(
     },
 
     # Save model
-    save_model = function(filename, num_iteration = NULL) {
+    save_model = function(filename, num_iteration = NULL, feature_importance_type = 0) {
```
jameslamb, line 427: the R linter we use requires explicitly differentiating between integers and doubles.
```suggestion
    save_model = function(filename, num_iteration = NULL, feature_importance_type = 0L) {
```

R-package/R/lgb.Booster.R
```diff
@@ -445,7 +446,7 @@ Booster <- R6::R6Class(
     },
 
     # Save model to string
-    save_model_to_string = function(num_iteration = NULL) {
+    save_model_to_string = function(num_iteration = NULL, feature_importance_type = 0) {
```
jameslamb, line 449:
```suggestion
    save_model_to_string = function(num_iteration = NULL, feature_importance_type = 0L) {
```

R-package/R/lgb.Booster.R
```diff
@@ -457,12 +458,13 @@ Booster <- R6::R6Class(
...
         , private$handle
         , as.integer(num_iteration)
+        , as.integer(feature_importance_type)
       ))
 
     },
 
     # Dump model in memory
-    dump_model = function(num_iteration = NULL) {
+    dump_model = function(num_iteration = NULL, feature_importance_type = 0) {
```
jameslamb, line 467:
```suggestion
    dump_model = function(num_iteration = NULL, feature_importance_type = 0L) {
```

src/c_api.cpp
```diff
@@ -689,21 +689,25 @@ class Booster {
...
     size_t len = std::strlen(model_str);
     boosting_->LoadModelFromString(model_str, len);
   }
 
-  std::string SaveModelToString(int start_iteration, int num_iteration) {
-    return boosting_->SaveModelToString(start_iteration, num_iteration);
+  std::string SaveModelToString(int start_iteration, int num_iteration,
+                                int feature_importance_type) {
+    return boosting_->SaveModelToString(start_iteration,
+                                        feature_importance_type, num_iteration);
```
guolinke, line 704:
```suggestion
                                        num_iteration, feature_importance_type);
```

guolinke:
ping @imatiach-msft for swig fix

imatiach-msft:
@guolinke this looks great!

StrikerRUS:
@guolinke In my commit a2755330ab4ddf72254bd0e02ccc0b9eccfc478c I
replaced `cout` type with `split` type according to the existing terminology:
https://github.com/microsoft/LightGBM/blob/61b3c30842810205bc95736e6e66baa790b098cd/include/LightGBM/c_api.h#L1067-L1076
added defines in C API for better user experience and changed in Python package param type from int to string for better usage and to unify with existing codebase:
https://github.com/microsoft/LightGBM/blob/61b3c30842810205bc95736e6e66baa790b098cd/python-package/lightgbm/sklearn.py#L244-L247
https://github.com/microsoft/LightGBM/blob/61b3c30842810205bc95736e6e66baa790b098cd/python-package/lightgbm/basic.py#L2956-L2959

include/LightGBM/config.h
```diff
@@ -532,6 +532,11 @@ struct Config {
   // desc = **Note**: can be used only in CLI version
   std::string output_model = "LightGBM_model.txt";
 
+  // desc = the feature importance type in the saved model file
+  // desc = ``0``: count-based feature importance (numbers of splits are counted); ``1``: gain-based feature importance (values of gain are counted)
+  // desc = **Note**: can be used only in CLI version
+  int saved_feature_importance_type = 0;
```
StrikerRUS, line 538: @guolinke I added `CLI-only` flag. Should it be marked as `[no-save]` as well?

StrikerRUS:
Fixed #3011.

include/LightGBM/config.h
guolinke: yeah, `CLI-only` is needed. As for `[no-save]`, I think it is better not marked, as user may want to know which feature importance type is saved in the model file.

guolinke:
Thank you @StrikerRUS  for the doc improvement, as always!

include/LightGBM/config.h
StrikerRUS: OK. Agree with you!

StrikerRUS:
@jameslamb Is something left for R-package?

StrikerRUS:
@jameslamb Thank you! I think it's better to have a separate PR for R tests.

Issue #1944, opened 2019-01-13
[R-package] Rewrite R demos, replace with vignettes


Laurae2:
The R package should have all the demos remade, short, concise, and focusing on specific topics, instead of currently being confusing to end-users who are trying to pick up the library (see some the recent issues, especially the multiclass ones).

I'll rewrite most of them later in the next months.

StrikerRUS:
Closed in favor of being in #2302. We decided to keep all feature requests in one place.

Welcome to contribute this feature! Please re-open this issue (or post a comment if you are not a topic starter) if you are actively working on implementing this feature.

jameslamb:
I want to add (for whoever comes back to this issue), that we should consider factoring `ggplot2` out of the demos when they are re-written. Right now the demos are the only reason `ggplot2` is still a `Suggests` dependency of the R package (see #2543 for more info) and I don't think it's giving us a lot of added benefit beyond what can be done in the base plotting options available in R.

jameslamb:
I think that when we come to this, we should change from the current `demo/` set up to `vignettes` in R markdown. That will provide a better experience for people reading the documentation from [the docs site](https://lightgbm.readthedocs.io/en/latest/R/index.html) because we can mix rich text (in markdown) with the code. Right now the demos are all code and comments, which you can only find if you know how to use the `demo()` function in the terminal.

jameslamb:
I just updated the title of this issue per this comment: https://github.com/microsoft/LightGBM/issues/1944#issuecomment-599140770

mayer79:
@jameslamb Are there still plans to replace demos by simple vignettes? If yes, I could give it a try.

jameslamb:
Yes definitely! I have been focused on the new Dask module in the Python package recently, so haven't been giving as much attention to the R side. I'd love if you submitted a proposal for this.

I don't think you should try to move all demos at once. I think it would be good to have a first pull request that just adds an introductory "Getting Started with LightGBM" vignette, to replace https://github.com/microsoft/LightGBM/blob/master/R-package/demo/basic_walkthrough.R.

Thank you so much for offering to help 😀

mayer79:
Sure thing!

jameslamb:
Adding some relevant links I've found while investigating #3946:

* https://blog.r-hub.io/2020/06/03/vignettes/
* http://www.markvanderloo.eu/yaRb/2019/01/11/add-a-static-pdf-vignette-to-an-r-package/
[...]

PR #3224, opened 2020-07-13
[R-package] factor out {ggplot2}

jameslamb:
Today the R package has a `Suggests` dependency on `{ggplot2}`.  This is only for one demo, `demo/leaf_stability.R`. This pull request proposes factoring it out and replacing it with `base` R code.

This will allow us to cut out this `R CMD check` NOTE:

> * checking package dependencies ... NOTE
Package suggested but not available for checking: ‘ggplot2’

More importantly, it's an expensive dependency that anyone installing the `Suggests` dependencies of `{lightgbm}`has to pay for.

Some of these plots look different from the comments around them, but this PR does not address that...it can be addressed in #1944 . This PR is limited to just factoring out `{ggplot2}`, to make the package as lightweight as possible when it goes to CRAN (work-in-progress, #629 ).

## Comparisons

I've included before and after screenshots for each plot below. The "Depth Density" plots look noticeably different, I think because of changes in how `geom_density()` works in `{ggplot2}`. `demo/leaf_stability.R` was written more than 3 years ago.

Thanks to @alistaire47 for helping me with `{ggplot2}` stuff 😀 

<details><summary>plot 1</summary>

**before**
image

**after**

image

</details>

<details><summary>plot 2</summary>

**before**

image

**after**

image

</details>

<details><summary>plot 3</summary>

**before**

image

**after**

image

</details>

<details><summary>plot 4</summary>

**before**

image

**after**

image

</details>

<details><summary>plot 5</summary>

**before**

image

**after**

image

</details>

<details><summary>plot 6</summary>

**before**

image

**after**

image

</details>

<details><summary>plot 7</summary>

**before**

image

**after**

image

</details>

<details><summary>plot 8</summary>

**before**

image

**after**

image

</details>

### DIFF
--- a/R-package/demo/leaf_stability.R
+++ b/R-package/demo/leaf_stability.R
@@ -4,7 +4,79 @@
 
 library(lightgbm)
 
-# Second, we load our data
+# define helper functions for creating plots
+
+# output of `RColorBrewer::brewer.pal(10, "RdYlGn")`, hardcooded here to avoid a dependency
+.diverging_palette <- c(
+  "#A50026", "#D73027", "#F46D43", "#FDAE61", "#FEE08B"
+  , "#D9EF8B", "#A6D96A", "#66BD63", "#1A9850", "#006837"
+)
+
+.prediction_depth_plot <- function(df) {
+  plot(
+    x = df$X
+    , y = df$Y
+    , type = "p"
+    , main = "Prediction Depth"
+    , xlab = "Leaf Bin"
+    , ylab = "Prediction Probability"
+    , pch = 19L
+    , col = .diverging_palette[df$binned + 1L]
+  )
+  legend(
+    "topright"
+    , title = "bin"
+    , legend = sort(unique(df$binned))
+    , pch = 19L
+    , col = .diverging_palette[sort(unique(df$binned + 1L))]
+    , cex = 0.7
+  )
+}
+
+.prediction_depth_spread_plot <- function(df) {
+  plot(
+    x = df$binned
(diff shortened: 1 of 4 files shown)


R-package/demo/leaf_stability.R
```diff
@@ -2,19 +2,89 @@
...
 # Do not do this in a real scenario.
 
-# First, we load our libraries
 library(lightgbm)
-library(ggplot2)
 
-# Second, we load our data
+# define helper functions for creating plots
+
+# output of `RColorBrewer::brewer.pal(10, "RdYlGn")`, hardcooded here to avoid a dependency
```
StrikerRUS, line 9:
```suggestion
# output of `RColorBrewer::brewer.pal(10, "RdYlGn")`, hardcoded here to avoid a dependency
```

StrikerRUS:
@jameslamb Can we now remove these lines?
https://github.com/microsoft/LightGBM/blob/50560573487e677037d392bfc0411e178ea80e6f/.ci/test_r_package.sh#L117-L118
https://github.com/microsoft/LightGBM/blob/50560573487e677037d392bfc0411e178ea80e6f/.ci/test_r_package_windows.ps1#L141

jameslamb:
oooooooooooooooo yes we can! I forgot about those. Yes I'll remove them right now. We can always add back if we introduce new `Suggests` dependencies (like whenever #1944 is picked up, for `rmarkdown`) but for now it would be better to simplify those files.

I'll open a PR right now, thanks for the suggestion!

PR #3231, opened 2020-07-16
[R-package] add test on importance types

jameslamb:
This PR proposes a test for the R package on the configuration of feature types introduced in #3220 .

### DIFF
--- a/R-package/tests/testthat/test_lgb.Booster.R
+++ b/R-package/tests/testthat/test_lgb.Booster.R
@@ -385,3 +385,66 @@ test_that("Booster$update() throws an informative error if you provide a non-Dat
         )
     }, regexp = "lgb.Booster.update: Only can use lgb.Dataset", fixed = TRUE)
 })
+
+context("save_model")
+
+test_that("Saving a model with different feature importance types works", {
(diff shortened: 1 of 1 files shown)


R-package/tests/testthat/test_lgb.Booster.R
```diff
@@ -385,3 +385,65 @@ test_that("Booster$update() throws an informative error if you provide a non-Dat
         )
     }, regexp = "lgb.Booster.update: Only can use lgb.Dataset", fixed = TRUE)
 })
+
+context("save_model")
+
+test_that("Saving a model with different feature importance types works", {
+    set.seed(708L)
+    data(agaricus.train, package = "lightgbm")
+    data(agaricus.test, package = "lightgbm")
```
StrikerRUS, line 394: Sorry for commenting in the closed PR, but is seems that test data is not used in this test.

jameslamb: blegh yeah I copied and pasted from other tests. Cut out the `valids` since it wasn't necessary, but I forgot to remove the test data. I'll just bundle that into #3240

PR #3235, opened 2020-07-17
[R-package] Add tests on lgb.check.eval, lgb.unloader, and lgb.encode.char

jameslamb:
This PR proposes adding a few more tests on R code that isn't currently covered by tests.

It also proposes fixing an inconsistency I found while writing the tests. Today `lgb.check.eval()` sometimes sets `metric` as a character vector, and sometimes sets it as a list. This PR fixes that and guarantees that it always uses a list. A `list` is preferred because it allows for mixing string metric  names and custom eval function objects.

### DIFF
--- a/R-package/R/utils.R
+++ b/R-package/R/utils.R
@@ -11,12 +11,10 @@ lgb.is.null.handle <- function(x) {
 }
 
 lgb.encode.char <- function(arr, len) {
-
   if (!is.raw(arr)) {
-    stop("lgb.encode.char: Can only encode from raw type") # Not an object of type raw
+    stop("lgb.encode.char: Can only encode from raw type")
   }
-  rawToChar(arr[seq_len(len)]) # Return the conversion of raw type to character type
-
+  return(rawToChar(arr[seq_len(len)]))
 }
 
 # [description] Raise an error. Before raising that error, check for any error message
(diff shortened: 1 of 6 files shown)


R-package/tests/testthat/test_utils.R
```diff
@@ -68,3 +77,50 @@ test_that("lgb.last_error() correctly returns errors from the C++ side", {
...
+})
+
+test_that("lgb.check.eval adds eval to metric in params", {
+    params <- lgb.check.eval(
+        params = list(metric = "auc")
+        , eval = "binary_error"
+    )
+    expect_named(params, "metric")
+    expect_identical(params[["metric"]], list("auc", "binary_error"))
+})
```
StrikerRUS, line 108: @jameslamb These two tests look identical. Is it expected or just copy-paste issue?

R-package/R/utils.R
```diff
@@ -311,17 +309,26 @@ lgb.check.obj <- function(params, obj) {
 
 }
 
+# [description]
+#     make sure that "metric" is populated on params,
+#     and add any eval values to itt
```
StrikerRUS, line 314: `to itt` -> `to it`

In case you will create a PR to remove duplicated test, you can fix this typo at the same time 😃

R-package/tests/testthat/test_utils.R
jameslamb: oh yes! Must be a copy-paste mistake. This is where I miss Python tests, where each test is a named function and a linter can warn you if you have two with the same name

PR #3240, opened 2020-07-20
[ci] remove unnecessary R_CHECK_FORCE_SUGGESTS in CI scripts

jameslamb:
See https://github.com/microsoft/LightGBM/pull/3224#issuecomment-661125688

#3224 removed the last `Suggests:` dependency for the R package that isn't installed in CI, so we can remove the environment variables set to change the `WARNING` about suggests dependencies into a `NOTE`.


jameslamb:
Also removed some unused test data, per https://github.com/microsoft/LightGBM/pull/3231/files#r457520122

PR #3246, opened 2020-07-22
fix bug in CEGB when reset training data or config

### LINKED ISSUE #3245: Bug for using CEGB with bagging
C++ library

LightGBM component: 

## Environment info

Operating System: macOS Mojave

CPU/GPU model: 

C++ compiler version: 

CMake version:

Java version:

Python version:

R version:

Other:

LightGBM version or commit hash: latest. last commit f5f27ca88c69d194ca03df88c5c2a58acce15e33 (7/16/2020)

## Reproducible example(s)
set `cegb_penalty_feature_coupled` to be a vector with positive doubles and `cegb_tradeoff` a bit bigger for the data (I set 4000) to see the effect easily
Then run a training job with iteration number to be say 200.
First with `bagging_fraction` set to be 1 (no bagging), then it will finish 200 iterations as expected
But if we set` bagging_fraction` < 1, say 0.8 . Then for my data, it stop training after 20 iterations with message: 
"Stopped training because there are no more leaves that meet the split requirements"

## Steps to reproduce
I think I found out the cause to this issue. It is a bug in the code. The cegb class is wrongly initialized again in each iteration (in serial_tree_learner.cpp L112) when use bagging. Then if we use bagging, the vector for storing the information of whether a feature has been used by previous tree is initialized for each iteration, thus at each iteration from root we are penalizing the information gain for every feature (which shouldn't be the case, we should only penalize a feature if it is not used by previous tree). That explains why we can not split at the root for only after 20 iterations when use bagging (I checked the model.txt to confirm that it is the iteration when the raw information gain at the root for every feature is less then the cegb penalty).

To fix, we just need to remove the `cegb_->Init();` at L112 in serial_tree_learner.cpp. cegb class should only be initialized once when we create the tree learner class. 

It's better if we can have the original authors of cegb paper to confirm.

guolinke:
to fix #3245

### DIFF
--- a/src/treelearner/serial_tree_learner.cpp
+++ b/src/treelearner/serial_tree_learner.cpp
@@ -141,7 +141,9 @@ void SerialTreeLearner::ResetConfig(const Config* config) {
   col_sampler_.SetConfig(config_);
   histogram_pool_.ResetConfig(train_data_, config_);
   if (CostEfficientGradientBoosting::IsEnable(config_)) {
-    cegb_.reset(new CostEfficientGradientBoosting(this));
+    if (cegb_ == nullptr){
+      cegb_.reset(new CostEfficientGradientBoosting(this));
+    }
     cegb_->Init();
   }
   constraints_.reset(LeafConstraintsBase::Create(config_, config_->num_leaves));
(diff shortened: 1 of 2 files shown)


src/treelearner/serial_tree_learner.cpp
```diff
@@ -141,7 +141,9 @@ void SerialTreeLearner::ResetConfig(const Config* config) {
   col_sampler_.SetConfig(config_);
   histogram_pool_.ResetConfig(train_data_, config_);
   if (CostEfficientGradientBoosting::IsEnable(config_)) {
-    cegb_.reset(new CostEfficientGradientBoosting(this));
+    if (cegb_ == nullptr){
```
guolinke, line 144:
```suggestion
    if (cegb_ == nullptr) {
```

PR #3256, opened 2020-07-28
update multi-class objective (softmax)

guolinke:
update the multi-class objective (softmax)

refer to 
image

from paper https://statweb.stanford.edu/~jhf/ftp/trebst.pdf


src/objective/multiclass_objective.hpp
```diff
@@ -25,6 +25,7 @@ class MulticlassSoftmax: public ObjectiveFunction {
  public:
   explicit MulticlassSoftmax(const Config& config) {
     num_class_ = config.num_class;
+    factor_ = static_cast<double>(num_class_ - 1) / num_class_;
```
guolinke, line 28:
```suggestion
    factor_ = static_cast<double>(num_class_) / (num_class_ - 1.0f);
```

src/objective/multiclass_objective.hpp
```diff
@@ -40,6 +41,7 @@ class MulticlassSoftmax: public ObjectiveFunction {
     if (num_class_ < 0) {
       Log::Fatal("Objective should contain num_class field");
     }
+    factor_ = static_cast<double>(num_class_ - 1) / num_class_;
```
guolinke, line 44:
```suggestion
    factor_ = static_cast<double>(num_class_) / (num_class_ - 1.0f);
```

src/objective/multiclass_objective.hpp
```diff
@@ -25,6 +25,7 @@ class MulticlassSoftmax: public ObjectiveFunction {
  public:
   explicit MulticlassSoftmax(const Config& config) {
     num_class_ = config.num_class;
+    factor_ = static_cast<double>(num_class_) / (num_class_ - 1.0f);
```
guolinke, line 28:
```suggestion
    //  This factor is to rescale the redundant form of K-classification, to the non-redundant form. 
    // In the traditional settings of K-classification, there is one redundant class, whose output is set to 0 (like the class 0 in binary classification). 
    // This is from the Friedman GBDT paper.
    factor_ = static_cast<double>(num_class_) / (num_class_ - 1.0f);
```

R-package/tests/testthat/test_basic.R
```diff
@@ -41,7 +41,7 @@ test_that("train and predict softmax", {
     data = as.matrix(iris[, -5L])
     , label = lb
     , num_leaves = 4L
-    , learning_rate = 0.1
+    , learning_rate = 0.05
```
jameslamb, line 44: how is this change related?

jameslamb: or, sorry, let me rephrase...why was it necessary to reduce the learning rate for this test?

guolinke: @jameslamb I just have a try. In previous, the output is multiplied by `1/2`. while currently, it is `3/4` (with num_class=4). So I reduce the learning rate a little bit.

guolinke: or can you help me to fix this test?

jameslamb: oooo I see, thanks! It looks like you were able to fix the test, sorry for my delayed reply

src/objective/multiclass_objective.hpp
```diff
@@ -25,6 +25,10 @@ class MulticlassSoftmax: public ObjectiveFunction {
  public:
   explicit MulticlassSoftmax(const Config& config) {
     num_class_ = config.num_class;
+    //  This factor is to rescale the redundant form of K-classification, to the non-redundant form.
```
jameslamb, line 28:
```suggestion
    // This factor is to rescale the redundant form of K-classification, to the non-redundant form. 
```

src/objective/multiclass_objective.hpp
```diff
@@ -25,6 +25,10 @@ class MulticlassSoftmax: public ObjectiveFunction {
  public:
   explicit MulticlassSoftmax(const Config& config) {
     num_class_ = config.num_class;
+    // This factor is to rescale the redundant form of K-classification, to the non-redundant form.
```
guolinke, line 28:
```suggestion
    // This factor is to rescale the redundant form of K-classification, to the non-redundant form.
```

src/objective/multiclass_objective.hpp
```diff
@@ -25,6 +25,10 @@ class MulticlassSoftmax: public ObjectiveFunction {
  public:
   explicit MulticlassSoftmax(const Config& config) {
     num_class_ = config.num_class;
+    // This factor is to rescale the redundant form of K-classification, to the non-redundant form. 
+    // In the traditional settings of K-classification, there is one redundant class, whose output is set to 0 (like the class 0 in binary classification).
```
guolinke, line 29:
```suggestion
    // In the traditional settings of K-classification, there is one redundant class, whose output is set to 0 (like the class 0 in binary classification).
```

lorentzenchr:
@guolinke I know this is an old one and, effectively, nothing really changes here except that the learning_rate is effected.

Do you know where this factor $\frac{K-1}{K}$ stems from? I never understood its derivation (I tried without success).

Even more irritating is that a numerical example contradicts this scaling factor:
```python
import numpy as np


def prob(F):
    c = F.shape[1]
    p = np.exp(F) / np.sum(np.exp(F), axis=1)[:, None]
    return p

def loss(y, F):
    p = prob(F)
    return -np.sum(y * np.log(p))


y = np.array([[0, 1], [0, 1], [1, 0]])
F = np.array([[0.2, 0.8], [0.3, 0.7], [0.4, 0.6]])
p = prob(F)
K = y.shape[1]

f_no_scaling = (y - p) /(p * (1-p))
f_with_scaling = (K - 1)/K * f_no_scaling  # the factor is 1/2

loss(y, F + f_no_scaling), loss(y, F + f_with_scaling)
```
`(0.06220676974920512, 0.353509963139608)`

So the one without scaling gives a smaller / better loss.

PR #3263, opened 2020-07-29
create buffer for gradients and hessians with goss and customized objective (fixes #3243)

### LINKED ISSUE #3243: [Python] Cannot run `goss` using custom objective function `fobj`

## Environment info

Operating System: Ubuntu 18.04.4 LTS

CPU/GPU model: Intel(R) Core(TM) i7-8650U CPU @ 1.90GHz (x86_64) / Intel Corporation UHD Graphics 620

C++ compiler version: gcc (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0

CMake version: 3.10.2

Java version: openjdk version "1.8.0_252"
OpenJDK Runtime Environment (build 1.8.0_252-8u252-b09-1~18.04-b09)
OpenJDK 64-Bit Server VM (build 25.252-b09, mixed mode)

Python version: 3.7.7

LightGBM version or commit hash: [f257636d](https://github.com/microsoft/LightGBM/tree/f257636d01dab724ed7a800cde20e68d5ebcc867) (latest commit).

```bash
[LightGBM] [Warning] Auto-choosing row-wise multi-threading, the overhead of testing was 0.000101 seconds.
You can set `force_row_wise=true` to remove the overhead.
And if memory is not enough, you can set `force_col_wise=true`.
[LightGBM] [Info] Total Bins 2550
[LightGBM] [Info] Number of data points in the train set: 10000, number of used features: 10
[LightGBM] [Info] Using GOSS
[LightGBM] [Info] Using GOSS
[LightGBM] [Info] Using GOSS
[1]    30327 segmentation fault (core dumped)  python -m goss_fobj
```

## Reproducible example(s)

```python
import numpy as np
import sklearn.datasets as sd
import lightgbm as lgb

# Auxiliar Functions
def sigmoid(x: np.ndarray) -> np.ndarray:
    """A numerically stable version of the logistic sigmoid function."""
    pos_mask = (x >= 0)
    neg_mask = (x < 0)
    z = np.zeros_like(x)
    z[pos_mask] = np.exp(-x[pos_mask])
    z[neg_mask] = np.exp(x[neg_mask])
    top = np.ones_like(x)
    top[neg_mask] = z[neg_mask]

    return top / (1 + z)

def x_entropy_udf(preds: np.ndarray, d: lgb.Dataset) -> tuple:
    labels = d.get_label()

    preds = sigmoid(preds)
    grads = preds - labels
    hesss = preds * (1.0 - preds)
    return grads, hesss


# Constants
SEED = 101
N_ROUNDS = 50
N_SAMPLES = 10 * 1000


X, y = sd.make_hastie_10_2(N_SAMPLES, random_state=SEED)
data = lgb.Dataset(X, y)
params = {
    'objective': None,
    'boosting': 'goss', 
    'seed': SEED,
}
lgb.train(params, data, num_boost_round=N_ROUNDS, fobj=x_entropy_udf)
```
## Steps to reproduce

1. Create conda environment with `python=3.7.7`
2. Activate environment and `python -m pip install setuptools wheel numpy scipy scikit-learn -U`
3. Download code from commit [f257636d](https://github.com/microsoft/LightGBM/tree/f257636d01dab724ed7a800cde20e68d5ebcc867), change directory to `LightGBM/python_package` and run `python -m setup install`
4. Move to home directory and run code above.

## Notes 

I'm able to run the code above for `N_ROUNDS=10`, however, raising this value leads to a segmentation fault. Neither the previous issues nor the posts in StackOverflow ([segmentation-fault-when-defining-a-custom-loss-in-lightgbm](https://stackoverflow.com/questions/60120979/segmentation-fault-when-defining-a-custom-loss-in-lightgbm)) have an answer to this error. 

As suggested in issue [#2156](https://github.com/microsoft/LightGBM/issues/2156),  I even tried running the [`advanced_example.py`](https://github.com/microsoft/LightGBM/blob/74ce2cfe9418b0def315c5e17f6244eefac64a60/examples/python-guide/advanced_example.py#L156-L162) but instead passed the parameters defined above. It still crashes.

shiyu1994:
This is to fix Issue #3243.
The gradients_ and hessians_ of GBDT object is of size 0 if customized objective function is used.
> https://github.com/microsoft/LightGBM/blob/1cf13dbaa433c414b2ec871ac4090bb853eceb1f/src/boosting/gbdt.cpp#L95-L99

However, they are used by the bagging function of GOSS, for example
> https://github.com/microsoft/LightGBM/blob/1cf13dbaa433c414b2ec871ac4090bb853eceb1f/src/boosting/goss.hpp#L103-L106

which causes the segment fault.

### DIFF
--- a/src/boosting/goss.hpp
+++ b/src/boosting/goss.hpp
@@ -61,7 +61,7 @@ class GOSS: public GBDT {
       CHECK(hessians != nullptr && objective_function_ == nullptr);
       size_t total_size = static_cast<size_t>(num_data_) * num_tree_per_iteration_;
       #pragma omp parallel for schedule(static)
-      for(size_t i = 0; i < total_size; ++i) {
+      for (size_t i = 0; i < total_size; ++i) {
         gradients_[i] = gradients[i];
         hessians_[i] = hessians[i];
       }
--- a/src/boosting/goss.hpp
+++ b/src/boosting/goss.hpp
@@ -36,6 +36,12 @@ class GOSS: public GBDT {
             const std::vector<const Metric*>& training_metrics) override {
     GBDT::Init(config, train_data, objective_function, training_metrics);
     ResetGoss();
+    if (objective_function_ == nullptr) {
+      // use customized objective function
+      size_t total_size = static_cast<size_t>(num_data_) * num_tree_per_iteration_;
+      gradients_.resize(total_size, 0.0f);
+      hessians_.resize(total_size, 0.0f);
+    }
   }
 
   void ResetTrainingData(const Dataset* train_data, const ObjectiveFunction* objective_function,
@@ -47,6 +53,23 @@ class GOSS: public GBDT {
   void ResetConfig(const Config* config) override {
     GBDT::ResetConfig(config);
     ResetGoss();
+  }
+
+  bool TrainOneIter(const score_t* gradients, const score_t* hessians) override {
+    if (gradients != nullptr) {
+      // use customized objective function
+      CHECK(hessians != nullptr && objective_function_ == nullptr);
+      size_t total_size = static_cast<size_t>(num_data_) * num_tree_per_iteration_;
+      #pragma omp parallel for schedule(static)
+      for(size_t i = 0; i < total_size; ++i) {
+        gradients_[i] = gradients[i];
+        hessians_[i] = hessians[i];
+      }
+      return GBDT::TrainOneIter(gradients_.data(), hessians_.data());
+    } else {
+      CHECK(hessians == nullptr);
+      return GBDT::TrainOneIter(nullptr, nullptr);
+    }
   }
 
   void ResetGoss() {


guolinke:
Hi @shiyu1994 , can we only fix this in goss.hpp. It seems the normal GBDT is not affected.

shiyu1994:
The code of normal GBDT is only modified to include two more parameters `gradients` and `hessians` for `Bagging` method and this won't affect its execution. These two parameters are only effective when GOSS is used. With GOSS we are supporting a bagging according to `gradients` and `hessians`. So I think including them in the interface of `Bagging` method should be meaningful, which indicates that our bagging process can use gradient information. 

The original `Bagging` and `BaggingHelper` do not provide parameters for `gradients` and `hessians`. With customized objective, these two arrays will be passed through the arguments of `TrainOneIter` from the outside of GBDT object. In that case, I think expanding the parameters to include 'gradients' and 'hessians' should be necessary. Otherwise GOSS code will have no access to the arrays passed from outside the GBDT object to `TrainOneIter` (inherited by GOSS without modification) which further calls the `Bagging` method. 

Is there a better way to do this or any other suggestion?

guolinke:
I was thinking to override the `TrainOneIter` in `GOSS`. 
in the `GOSS:TrainOneIter`, we can firstly copy the external grad/hess to grad_/hess_, then call `GBDT::TrainOneIter(grad_, hess_)`.

shiyu1994:
Ok. That seems more elegant. I'll modify this.

src/boosting/goss.hpp
```diff
@@ -49,6 +55,23 @@ class GOSS: public GBDT {
     ResetGoss();
   }
 
+  bool TrainOneIter(const score_t* gradients, const score_t* hessians) override {
+    if (gradients != nullptr) {
+      // use customized objective function
+      CHECK(hessians != nullptr && objective_function_ == nullptr);
+      size_t total_size = static_cast<size_t>(num_data_) * num_tree_per_iteration_;
+      #pragma omp parallel for schedule(static)
+      for(size_t i = 0; i < total_size; ++i) {
```
guolinke, line 64:
```suggestion
      for (size_t i = 0; i < total_size; ++i) {
```

StrikerRUS:
Compilation of current `master` fails for Windows with the following error:
```
d:\a\1\s\src\boosting\goss.hpp(62): error C3016: 'i': index variable in OpenMP 'for' statement must have signed integral type (compiling source file D:\a\1\s\src\boosting\boosting.cpp) [D:\a\1\s\build\lightgbm.vcxproj]
```

I believe it was introduced here.

BTW, it is quite strange that CI allowed to merge this PR.

guolinke:
@StrikerRUS sorry, it is may bad!
I fix this directly in 9b2637354a84e0da813f8e2b22608a87eee01e4c

StrikerRUS:
Ah, OK! Thanks a lot for the prompt fix!

Issue #3241, opened 2020-07-20
[bug] Introduced bug in *Fast() methods with the locking strategy PR


AlbertoEAF:
The https://github.com/microsoft/LightGBM/pull/2760 PR introduced 2 bugs in the fast methods introduced in https://github.com/microsoft/LightGBM/pull/2992/.

The non-fast variants were correctly changed, but the Fast variants were not updated in the same manner:

- https://github.com/microsoft/LightGBM/commit/1c35c3b9ede9adab8ccc5fd7b4b2b6af188a79f0#diff-59f374370af59cd4cbd753881fc246c5R1936-R1981
- https://github.com/microsoft/LightGBM/commit/1c35c3b9ede9adab8ccc5fd7b4b2b6af188a79f0#diff-59f374370af59cd4cbd753881fc246c5R2079-R2114

I only noticed that when building the latest version, since the parameter "num_iteration" is not used anymore after the PR https://github.com/microsoft/LightGBM/pull/2760.

I can add a PR to fix it. However, this change to a separate `SetSingleRowPredictor` presents now an opportunity for further optimization, moving these calls to the FastInit, or a new FastUpdate, which would change the API of the Fast methods again (and in turn impact with the 3.0.0 release - https://github.com/microsoft/LightGBM/pull/3071). Also, not reusing `get_row_fun` as originally in https://github.com/microsoft/LightGBM/pull/2992/ considerably reduced the performance gains. Maybe it would be interesting to reintroduce that possibility. If a new data pointer is passed to Fast() scoring, that generates a new function, otherwise it uses the one stored in the FastConfig. Or, we can cache the data pointer and test if the pointer changed.
All this leads me to ask to delay a bit the 3.0 release again, since there's this new bug as well and changing the API a bit further would allow us to make LightGBM even faster for the Fast calls.

What are your thoughts @guolinke, @StrikerRUS, @StrikerRUS, @imatiach-msft, et al.?

Thanks!

guolinke:
@AlbertoEAF I think delaying the release of  3.0.0-rc1 for about one-week (or slightly longer) is not a big deal. 
What is the ETA of this fix?

JoanFM:
Hey @AlbertoEAF 

Sorry for the inconvenience, did I screw doing some realignment? Let me know if I can be of help on fixing the issue.

AlbertoEAF:
[...]

PR #3268, opened 2020-08-01
Fast single row predict API v2

AlbertoEAF:
- Fix bug introduced in PR #2992 for Fast predict (closes https://github.com/microsoft/LightGBM/issues/3241)
- Uses the newer API to make *Fast predict calls faster 
  The latest single-thread benchmark with a model of 42 numerical features achieved a speedup 1.30x-1.85x vs the standard single row prediction API)
- Stricter API with const parameters


AlbertoEAF:
@imatiach-msft @StrikerRUS @guolinke please review, it is now even faster by the latest tests.

Thank you for waiting for the patch and good luck with the v3 release! ;)

StrikerRUS:
I'm confused by `const`/`non-const` params mismatch in C API functions.

guolinke:
For the non pointer arguments, I think we can unify them to the non constant types.

AlbertoEAF:
@StrikerRUS good catch! I had const mismatches even between the .c and .h in the function signatures which made things extremely weird.

@guolinke I put instead everything const just because many were lacking const when they could be. Now I think it's really easy to interpret from the user's viewpoint. Only the input booster - which will be modified - and output pointers, are non-const.

AlbertoEAF:
@guolinke do you think it's ready to merge?

Thank you

guolinke:
Yeah, looks good to me

Issue #2681, opened 2020-01-12
[R-package] lgb.prepare functions should warn on unconverted columns of unsupported types 


jameslamb:
## Summary

`lgb.prepare()`, `lgb.prepare_rules()`, `lgb.prepare2()`, and `lgb.prepare_rules2()` should tell users in a log message if any columns remain unconverted because they are of a type that the function does not support. The names of those columns should be logged.

## Motivation

The R package currently exports four functions that can be used to convert tabular datasets into model-ready form:

* `lgb.prepare()`: converts columns of type `"character"` and `"factor"` to `"numeric"`
* `lgb.prepare2()`: converts columns of type `"character"` and `"factor"` to `"integer"`
* `lgb.prepare_rules()`: similar to `lgb.prepare()`, but returns a set of "rules" describing how non-numeric values were mapped to numeric values. Also allows for user-provided rules, useful for cases where you want to be sure the encoding is the same on multiple datasets  (e.g. training, test, and validation datasets)
* `lgb.prepare_rules2()`: similar to `lgb.prepare()`, but returns a set of "rules" describing how non-numeric values were mapped to numeric values. Also allows for user-provided rules, useful for cases where you want to be sure the encoding is the same on multiple datasets  (e.g. training, test, and validation datasets)

These functions are intended to make it easier to create a model-ready dataset (all numeric or all integer). The user expectation is likely that after using calling one of these functions on a dataset, that dataset is ready to use in a model. If that dataset contains columns of other types (not integer or numeric), the user should be notified in a log message.

Column types that these  functions are unlikely to support:

* `POSIX*`
* `Date`
* `list`
* `data.frame`
* `data.table`

jameslamb:
Closed in favor of being in #2302. We decided to keep all feature requests in one place.

Welcome to contribute this feature! Please re-open this issue (or post a comment if you are not a topic starter) if you are actively working on implementing this feature.

jameslamb:
I just updated this description and title now that #3095 has been merged.

PR #3269, opened 2020-08-03
[R-package] refactor and improvements to lgb.convert() functions (fixes #2678, #2681)

### LINKED ISSUE #2678: [R-package] lgb.convert() functions should convert columns of type 'logical'
## Summary

`lgb.convert()` and `lgb.convert_with_rules()` should convert columns of type `"logical"` to type `"integer"`.

## Motivation

The R package currently exports two functions that can be used to convert tabular datasets into model-ready form:

* `lgb.convert()`: converts columns of type `"character"` and `"factor"` to `"integer"`
* `lgb.convert_with_rules()`: similar to `lgb.convert()`, but returns a set of "rules" describing how non-numeric values were mapped to integer values. Also allows for user-provided rules, useful for cases where you want to be sure the encoding is the same on multiple datasets  (e.g. training, test, and validation datasets)

These functions are intended to make it easier to create a model-ready dataset (all numeric or all integer). The fact that they do not convert columns of type `"logical"` means that they don't cover a very common case...converting `TRUE/FALSE` to `1/0`.

jameslamb:
This PR fixes #2678 and #2681 by adding the following improvements to `lgb.convert()` functions:

* `lgb.convert()`:
    - now fills in missing values in factor and character columns with 0
* `lgb.convert_with_rules()`:
    - `rules` no longer contains rules where the "name" in the rule is a missing value
* both:
    - now converts `logical` columns to `integer`, filling in missing values with -1
    - now warns if any columns remain which are not `numeric` or `integer`
    - moved code for getting column classes into shared internal function to reduce duplication
    - changed from `inherits(data, "data.table")` to `data.table::is.data.table()`
    - moved error handling up to the beginning of the functions
    - clarified documentation

## Question for reviewers

@Laurae2 @guolinke could we remove `lgb.convert()` and just offer `lgb.convert_with_rules()`? Using `lgb.convert_with_rules()` without passing your own rules (the default), has identical behavior to `lgb.convert()` and is better because it returns the `rules` on how columns were changed, which improves reproducibility. Since we're considering a `3.0.0` release in #3071 right now, I think this is the right time for such a breaking change.


guolinke:
Yeah, i think it is great

jameslamb:
> Yeah, i think it is great

ok thanks! I can do that

jameslamb:
I just removed `lgb.convert()` in https://github.com/microsoft/LightGBM/pull/3269/commits/db2792088f9bbd2272ba203360057cd6d5cdd551. I also removed the demo specific to that function from `demo/`.

jameslamb:
@guolinke I'll merge `master`  into this, fix the conflicts, and merge it once CI passes. Hopefully should be quick.

PR #3281, opened 2020-08-07
bump version for development

guolinke:


guolinke:
@jameslamb the version number in R is not updated yet, you can change it together with the new solution of R's versioning.

jameslamb:
> @jameslamb the version number in R is not updated yet, you can change it together with the new solution of R's versioning.

ok sounds good. Added #3283 to track the work

Issue #3283, opened 2020-08-07
[R-package] automate release process for R


jameslamb:
See https://github.com/microsoft/LightGBM/pull/3071#issuecomment-670243598 for background. To close this issue, implement the solution in that comment or some other approach that allows for building a release of the R package without manually updating anything other than `VERSION.txt`

jameslamb:
this should also include automating the creation of binaries: https://github.com/microsoft/LightGBM/pull/3188#issuecomment-667683956

jameslamb:
this should also include creating an artifact with the source distribution we'd upload to CRAN:  https://github.com/microsoft/LightGBM/pull/3285#issuecomment-670968342

jameslamb:
When I pick this up, I want to be sure to include in the release notes a shout-out to the maintainers of https://github.com/curso-r/rightgbm/issues/2 who have already done good work figuring out a process for some of this.

PR #3285, opened 2020-08-08
[R-package] Add docs on building and using precompiled binaries for the R package

### LINKED ISSUE #3187: [R-package] 32-bit library support
## Summary

As I've investigated #2960 as part of #629 , I've learned that packages on CRAN have to be able to build on 32-bit Windows.

**I believe that we need to add support for a 32-bit build LightGBM to get the R package to CRAN.**

### Evidence that this is required

I have not yet found any official statement that says "all R packages have to build on 32-bit Windows", but I see strong evidence that that is true.

You can find the current list of check environments at https://cran.r-project.org/web/checks/check_flavors.html. None of those seem to be 32-bit

image

but in ["Writing R Extensions"](https://cran.r-project.org/doc/manuals/r-release/R-exts.html) I see the following:

> On systems which support multiple sub-architectures (principally Windows), R CMD check will install and check a package which contains compiled code under all available sub-architectures.

And in the ["R  for Windows FAQs"](https://cran.r-project.org/bin/windows/base/rw-FAQ.html):

> Packages without compiled code nor a configure.win script will run on both 32- and 64-bit R.

> Packages with compiled code but no configure.win nor src/Makefile.win file will be built for both when running on a 64-bit version of Windows if both versions of R are installed.

> An empty configure.win is treated in the same way as if none existed. Also, there is a list of packages known to have an architecture-independent configure.win hardcoded into R CMD INSTALL, and for these packages, both architectures will be built under the above conditions. Other packages can be installed with configure.win run for just the first architecture by using option --force-biarch.

> Any package can be installed for first one architecture and then the other with option --merge-multiarch, but the package source must be a tarball (and as before, running on a 64-bit version of Windows with both versions of R installed).

> Finally, a package without a src/Makefile.win file and no or empty or architecture-independent configure.win file can be installed for both architectures from 32-bit Windows if the 64-bit components were selected when R was installed and option --compile-both is given. Obviously, only the 32-bit installation can be tested.

This is why you see pull requests this in other projects: https://github.com/dmlc/xgboost/pull/5777

I've tried looking in [the source for R CMD check](https://github.com/wch/r-source/blob/f27cbf1a52a31cd9b9676340394946a22041a4ae/src/library/tools/R/check.R) but it is very difficult to follow.

## Motivation

I believe that we need to add support for a 32-bit build LightGBM to get the R package to CRAN.

## Description

To check if the R package works with 32-bit R, try the following:

```shell
sh build-cran-package.sh
"C:\Program Files\R\R-4.0.2\bin\i386\R.exe" CMD check lightgbm_3.0.0-1.tar.gz --as-cran
```

## References

There was some discussion about this in #1129 , where it was decided that supporting a 32-bit is possible but that the maintenance burden would be too much. Given that it is needed to get LightGBM to CRAN, I think it should be reconsidered.

jameslamb:
This PR comes out of this discussion: https://github.com/microsoft/LightGBM/pull/3188#issuecomment-664083054.

We're really close to getting the R package to CRAN, but I'm not sure how long it will take to fix  #3187 . To make life easier for users until the package gets to CRAN, starting with https://github.com/microsoft/LightGBM/releases/tag/v3.0.0rc1 I think we should start providing precompiled binaries for releases.

When #3283 is done, these binaries can be produced automatically and attached to the releases page. For now, I think it will be useful to document how to use them (for users) and how to build them (for maintainers and adventurous users).


R-package/README.md
```diff
@@ -13,6 +13,36 @@ LightGBM R-package
...
+### Installing Precompiled Binaries
+
+Starting with `LightGBM` 3.0.0, precompiled binaries for the R package are created for each release. These packages do not require compilation, so they will be faster and easier to install than packages that are built from source. These packages are created with R 4.0 and are not guaranteed to work with other R versions.
+
+Binaries are available for Windows, Mac, and Linux systems. They are not guaranteed to work with all variants and versions of these operating systems. Please [open an issue](https://github.com/microsoft/LightGBM/issues) if you encoutnter any problems.
+
+To install a binary for the R package:
+
+1. Choose a release from [the "Releases" page](https://github.com/microsoft/LightGBM/releases)
+2. Choose a file based on your operating system. Right-click it and choose "copy link address"
```
StrikerRUS, line 25:
```suggestion
1. Choose a release from [the "Releases" page](https://github.com/microsoft/LightGBM/releases).
2. Choose a file based on your operating system. Right-click it and choose "copy link address".
```

R-package/README.md
```diff
@@ -221,6 +251,73 @@ At build time, `configure.win` will be run and used to create a file `Makevars.w
 1. Edit `configure.win` directly
 2. Edit `src/Makevars.win.in`
 
+### Build Precompiled Binaries of the CRAN Package
+
+This section is mainly for maintainers. As long as the R package is not available on CRAN (which will build precompiled binaries automatically) you may want to build precompiled versions of the R package manually, since these will be easier for users to install.
+
+For more details, see ["Writing R Extensions"](https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Building-binary-packages).
+
+Packages built like this will only work for the minor version of R used to build them. They may or may not work across different versions of operating systems
```
StrikerRUS, line 260:
```suggestion
Packages built like this will only work for the minor version of R used to build them. They may or may not work across different versions of operating systems.
```

R-package/README.md
```diff
@@ -221,6 +251,73 @@ At build time, `configure.win` will be run and used to create a file `Makevars.w
...
+
+**Mac**
+
+Binary produced: `lightgbm_${VERSION}-r40-macos.tgz`
```
StrikerRUS, line 264:
```suggestion
Binary produced: `lightgbm_${VERSION}-r40-macos.tgz`.
```

R-package/README.md
```diff
@@ -221,6 +251,73 @@ At build time, `configure.win` will be run and used to create a file `Makevars.w
...
+sh build-cran-package.sh
+R CMD INSTALL --build lightgbm_${LGB_VERSION}.tar.gz
+mv \
+    lightgbm_${LGB_VERSION}.tgz \
+    lightgbm_${LGB_VERSION}-r40-macos.tgz
+```
+
+**Linux**
+
+Binary produced: `lightgbm_${VERSION}-r40-linux.tar.gz`
```
StrikerRUS, line 277:
```suggestion
Binary produced: `lightgbm_${VERSION}-r40-linux.tar.gz`.
```

R-package/README.md
```diff
@@ -221,6 +251,73 @@ At build time, `configure.win` will be run and used to create a file `Makevars.w
...
+```shell
+R_VERSION=4.0.2
+
+docker run \
+    -v $(pwd):/opt/LightGBM \
+    -it rocker/verse:${R_VERSION} \
+        /bin/bash
+```
+
+From inside that container, the commands to create a precompiled binary are very similar
```
StrikerRUS, line 290:
```suggestion
From inside that container, the commands to create a precompiled binary are very similar.
```

R-package/README.md
```diff
@@ -221,6 +251,73 @@ At build time, `configure.win` will be run and used to create a file `Makevars.w
...
+
+Exit the container, and the binary package should still be there on the host system.
+
+```shell
+exit
+```
+
+**Windows**
+
+Binary produced: `lightgbm_${VERSION}.zip`
```
StrikerRUS, line 310:
```suggestion
Binary produced: `lightgbm_${VERSION}.zip`.
```

StrikerRUS, line 310: At release we have
```
lightgbm-${VERSION}.zip
        ^
```

image


Was it a typo?

StrikerRUS, line 310: At the Release page we have
```
lightgbm-${VERSION}.zip
        ^
```

image


Was it a typo?

StrikerRUS:
During reading this comment https://github.com/microsoft/LightGBM/issues/3187#issuecomment-670966615 I came to the thought that it might be useful to upload the source distribution to Releases page which will be then submitted to CRAN. WDYT @jameslamb ?

R-package/README.md
jameslamb: no I'll make them all `-` instead of `_` in the docs

R-package/README.md
```diff
@@ -13,6 +13,36 @@ LightGBM R-package
...
+To install a binary for the R package:
+
+1. Choose a release from [the "Releases" page](https://github.com/microsoft/LightGBM/releases).
+2. Choose a file based on your operating system. Right-click it and choose "copy link address".
+3. Copy that link into `PKG_URL` in the code below and run it.
+
+This sample code installs versioon 3.0.0-1 of the R package on Mac.
+
+```r
+PKG_URL <- "https://github.com/microsoft/LightGBM/releases/download/v3.0.0rc1/lightgbm_3.0.0-1-r40-macos.tgz"
```
StrikerRUS, line 31:
```suggestion
PKG_URL <- "https://github.com/microsoft/LightGBM/releases/download/v3.0.0rc1/lightgbm-3.0.0-1-r40-macos.tgz"
```

R-package/README.md
```diff
@@ -221,6 +251,73 @@ At build time, `configure.win` will be run and used to create a file `Makevars.w
...
+From inside that container, the commands to create a precompiled binary are very similar.
+
+```shell
+cd /opt/LightGBM
+LGB_VERSION="3.0.0-1"
+sh build-cran-package.sh
+R CMD INSTALL --build lightgbm_${LGB_VERSION}.tar.gz
+mv \
+    lightgbm_${LGB_VERSION}_R_*-linux-gnu.tar.gz \
+    lightgbm_${LGB_VERSION}-r40-linux.tgz
```
StrikerRUS, line 299:
```suggestion
    lightgbm-${LGB_VERSION}-r40-linux.tgz
```

R-package/README.md
```diff
@@ -221,6 +251,73 @@ At build time, `configure.win` will be run and used to create a file `Makevars.w
...
+
+Exit the container, and the binary package should still be there on the host system.
+
+```shell
+exit
+```
+
+**Windows**
+
+Binary produced: `lightgbm-${VERSION}.zip`.
```
StrikerRUS, line 310:
```suggestion
Binary produced: `lightgbm-${VERSION}-r40-windows.zip`.
```

R-package/README.md
```diff
@@ -221,6 +251,73 @@ At build time, `configure.win` will be run and used to create a file `Makevars.w
...
+
+```shell
+LGB_VERSION="3.0.0-1"
+sh build-cran-package.sh
+R CMD INSTALL --build lightgbm_${LGB_VERSION}.tar.gz
+mv \
+    lightgbm_${LGB_VERSION}.tgz \
+    lightgbm-${LGB_VERSION}-r40-windows.tgz
```
StrikerRUS, line 318:
```suggestion
    lightgbm-${LGB_VERSION}-r40-windows.zip
```

R-package/README.md
StrikerRUS: I listed some other inconsistent places in https://github.com/microsoft/LightGBM/pull/3285#pullrequestreview-463895208.

Also, I guess files should be renamed at Releases page:

image

jameslamb: blegh sorry! Updated

jameslamb:
> During reading this comment [#3187 (comment)](https://github.com/microsoft/LightGBM/issues/3187#issuecomment-670966615) I came to the thought that it might be useful to upload the source distribution to Releases page (automatically after resolving #3283 ) which will be then submitted to CRAN. WDYT @jameslamb ?

yes definitely!  I think that is a great idea and something we should be doing. I'll add it and a note in these docs.

And then we can add it automatically in future releases (#3283 )

jameslamb:
I just added a source distribution (built with `sh build-cran-package.sh`) to the 3.0.0rc1  release page!

I also added it to the list of artifacts that should be produced by #3283 

It can be installed like this from R.

```r
remotes::install_url(
    "https://github.com/microsoft/LightGBM/releases/download/v3.0.0rc1/lightgbm-3.0.0-1-cran.tar.gz"
)
```

jameslamb:
I just added notes on installing the CRAN package from the releases page, thanks for the nudge @StrikerRUS !

StrikerRUS:
@jameslamb 
> I also added it to the list of artifacts that should be produced by #3283

Thanks a lot!

I think after we put LightGBM on CRAN, we will have to produce only that artifact with source files and upload it to CRAN. All other artifacts with binaries should be uploaded reversely: from CRAN to Releases page (if needed) to have identical versions in both places and avoid errors like one just reported about glibc (https://github.com/microsoft/LightGBM/pull/3188#issuecomment-671088281). Please correct me if I'm wrong in my thoughts about further release pipeline.

jameslamb:
> @jameslamb
> 
> > I also added it to the list of artifacts that should be produced by #3283
> 
> Thanks a lot!
> 
> I think after we put LightGBM on CRAN, we will have to produce only that artifact with source files and upload it to CRAN. All other artifacts with binaries should be uploaded reversely: from CRAN to Releases page (if needed) to have identical versions in both places and avoid errors like one just reported about glibc ([#3188 (comment)](https://github.com/microsoft/LightGBM/pull/3188#issuecomment-671088281)). Please correct me if I'm wrong in my thoughts about further release pipeline.

Yes I agree with this. By the way, CRAN does not provide binaries for Linux (only Mac and Windows), so I think ours should be considered experimental.

StrikerRUS:
> By the way, CRAN does not provide binaries for Linux (only Mac and Windows), so I think ours should be considered experimental.

Hmm, then we should build them with our Docker we use for artifacts at Azure Pipelines to ensure maximum compatibility. Or even better with CentOS image.
https://github.com/pypa/manylinux/blob/master/docker/Dockerfile-x86_64

R-package/README.md
jameslamb: ok they are all consistent now

image

R-package/README.md
```diff
@@ -13,6 +13,55 @@ LightGBM R-package
...
 ------------
 
+### Installing the CRAN package
+
+As of this writing, `LightGBM`'s R package is not available on CRAN. However, start with `LightGBM` 3.0.0, you can install a released source distribution. This is the same type of package that you'd install from CRAN. It does not require `CMake`, Visual Studio, or anything else outside the CRAN toolchain.
+
+To install this package on any operating system:
+
+1. Choose a release from [the "Releases" page](https://github.com/microsoft/LightGBM/releases)
+2. Look for the artifact with a name like `lgb-{VERSION}-r-cran.tar.gz`. Right-click it and choose "copy link address".
```
StrikerRUS, line 23:
```suggestion
2. Look for the artifact with a name like `lightgbm-{VERSION}-r-cran.tar.gz`. Right-click it and choose "copy link address".
```

R-package/README.md
```diff
@@ -221,6 +270,73 @@ At build time, `configure.win` will be run and used to create a file `Makevars.w
...
+sh build-cran-package.sh
+R CMD INSTALL --build lightgbm_${LGB_VERSION}.tar.gz
+mv \
+    lightgbm_${LGB_VERSION}.tgz \
+    lightgbm-${LGB_VERSION}-r40-macos.tgz
+```
+
+**Linux**
+
+Binary produced: `lightgbm-${VERSION}-r40-linux.tar.gz`.
```
StrikerRUS, line 296:
```suggestion
Binary produced: `lightgbm-${VERSION}-r40-linux.tgz`.
```

jameslamb:
thanks for keeping me consistent @StrikerRUS , sorry for all the errors.

PR #3272, opened 2020-08-04
[Python] add start_iteration to python predict interface (#3058)

shiyu1994:
Suggested by Issue #3058. Add start_iteration for python prediction.

PR #3288, opened 2020-08-09
simplify start_iteration param for predict in Python and some code cleanup for start_iteration

StrikerRUS:
Refer to #3272.

cc @shiyu1994


python-package/lightgbm/basic.py
```diff
@@ -2823,14 +2823,14 @@ def predict(self, data, start_iteration=None, num_iteration=None,
         data : string, numpy array, pandas DataFrame, H2O DataTable's Frame or scipy.sparse
             Data source for prediction.
             If string, it represents the path to txt file.
-        start_iteration : int or None, optional (default=None)
+        start_iteration : int, optional (default=0)
```
StrikerRUS, line 2826: In `num_iteration` None is useful because it allows to set `best_iteration` implicitly. But for `start_iteration` I don't see any reasons to allow `None`. Actually, it was just an alias for `<=0`. I believe it doesn't worth to maintain `None`.

include/LightGBM/c_api.h
```diff
@@ -724,7 +724,7 @@ LIGHTGBM_C_EXPORT int LGBM_FastConfigFree(FastConfigHandle fastConfig);
  * \note
  * You should pre-allocate memory for ``out_result``:
  *   - for normal and raw score, its length is equal to ``num_class * num_data``;
- *   - for leaf index, its length is equal to ``num_class * num_data * num_iteration``;
+ *   - for leaf index, its length is equal to ``num_class * num_data * (start_iteration + num_iteration)``;
```
shiyu1994, line 727: I think here should be  `* num_iteration` instead of `* (start_iteration + num_iteration)`. Since in `PredictLeafIndex` in gbdt_prediction.cpp, we only output the leaf indices from `start_iteration`.

guolinke: is `num_iteration` the count, or the `end_iteration`?

shiyu1994: It's the count.

StrikerRUS: @shiyu1994 Thanks a lot! Indeed, those `start_iteration +` changes should be reverted. Initially, when I started to read your original PR I thought that `num_iteration` means the last used iteration (just like in `dump_model` or `save_model`). But after getting to cpp files I understood that it is number of iteration actually and forgot to revert docs changes I had already done.

StrikerRUS: Done in f55619d3b8a2bc8a0dff4eccaa9c957e07e1272d.

Issue #1315, opened 2018-04-16
Piecewise linear trees


sosmond:
I recently discovered [this paper](https://arxiv.org/pdf/1802.05640.pdf) which was submitted earlier this year showing impressive results of GBDTs utilizing piecewise linear trees. It's claimed that PL trees converge faster and are more accurate than the constant trees xgboost & lightgbm uses, and that optimizations exist to make GBDT-PL competitive speed-wise. 

The repo for the paper can be found [here](https://github.com/GBDT-PL/GBDT-PL). Is there any chance of adding this feature?

StrikerRUS:
Closed in favor of being in #2302. We decided to keep all feature requests in one place.

Welcome to contribute this feature! Please re-open this issue (or post a comment if you are not a topic starter) if you are actively working on implementing this feature.

kruda:
I'd like to work on this issue.

jameslamb:
Thanks @kruda ! I'll re-open and assign to you. Let us know if you have any questions.

StrikerRUS:
@jameslamb I believe we should re-open a feature request issue only after any evidence of active work on it. Otherwise, we may fall down into dozens of open stale issues like it was before #2302, e.g. here https://github.com/microsoft/LightGBM/issues/261#issuecomment-436995220.

@kruda You can submit a PR any time you want marking it `[WIP]` so that it will be easier to discuss implementation aspects and ask related questions.

jameslamb:
> @jameslamb I believe we should re-open a feature request issue only after any evidence of active work on it. Otherwise, we may fall down into dozens of open stale issues like it was before #2302, e.g. here [#261 (comment)](https://github.com/microsoft/LightGBM/issues/261#issuecomment-436995220).
> 
> @kruda You can submit a PR any time you want marking it `[WIP]` so that it will be easier to discuss implementation aspects and ask related questions.

ok fair enough!

stanpcf:
Hi @kruda How is your process going?

StrikerRUS:
Implemented in #3299.

PR #3299, opened 2020-08-12
Trees with linear models at leaves

btrotta:
Implements boosting for trees with linear models at the leaves (sometimes called M5 trees). This is a hybrid between traditional tree boosting and the model proposed in the paper Gradient Boosting with Piece-Wise Linear Regression Trees by Shi, Li, and Li (https://arxiv.org/pdf/1802.05640.pdf), which is mentioned in #1315. In this PR, the tree structure is created by finding the best split in the normal way, but then we calculate a linear model on each leaf. In contrast, in the paper, the splits are chosen by calculating the linear models for each potential split point, which is much more computationally intensive and would require more significant code changes in LightGBM. (The paper above actually mentions M5 trees, in Appendix D, but only tests existing slow implementations, which give poor results compared to their code. I think with the better implementation from this PR, M5 trees would come close to the performance of the fully-linear approach.)

The running time of the linear-leaf model is around 10-20% more than traditional tree boosting (depending on the dataset etc), but it converges faster, so overall it gives a small improvement in training time (and also can achieve slightly better accuracy). Memory use is higher since we need so store the full feature data.

Regularisation can be controlled with the parameter `linear_lambda`; this is important because it's more prone to overfitting than the traditional tree-boosting model. Also it's important to scale the data before training so that all features have similar mean and standard deviation.

The code uses parts of the Eigen library licensed under MPL2.

I have only implemented this for Python. I think to get it working for R it would require some changes to the data loading interface, but I'm not very familiar with R, so maybe someone else would like to take that on.

Here is a test script to measure performance on the SUSY physics data (https://archive.ics.uci.edu/ml/datasets/SUSY).
```
import pandas as pd
import numpy as np
import lightgbm as lgb
import time as time
from sklearn.metrics import log_loss, roc_auc_score

# read data
train = pd.read_csv('susy.csv', header=None)
np.random.seed(0)

# normalise
for i in range(len(train.columns) - 1):
    m, s = train.iloc[:, i + 1].agg(['mean', 'std'])
    train.iloc[:, i + 1] = (train.iloc[:, i + 1] - m) / s

train_bool = train.index <= train.index.max() - 500000  # last 500000 is the hold-out test set
valid_bool = train.index <= 500000  # validation set for early stopping

# linear model
t = time.time()
lgb_train = lgb.Dataset(train.loc[train_bool & ~valid_bool, list(range(1, 19))], label=train.loc[train_bool & ~valid_bool, 0])
lgb_test = lgb.Dataset(train.loc[valid_bool, list(range(1, 19))], label=train.loc[valid_bool, 0])
# linear model
params = {'objective': 'binary', 'seed': 0, 'num_leaves': 16, 'learning_rate': 0.05,
          'metric': 'binary_logloss', 'verbose': 2, 'linear_lambda': 0.01, 'linear_tree': True}
res = {}
t0 = time.time()
time_arr_linear = []
timer_callback = lambda env: time_arr_linear.append(time.time() - t0)
est_linear = lgb.train(params, lgb_train, num_boost_round=500,valid_sets=valid_sets, valid_names=valid_names,
                       categorical_feature=[], evals_result=res, callbacks=[timer_callback])
linear_time = time.time() - t
print("Linear model")
print("training time ", linear_time)

# normal model
params = {'objective': 'binary', 'seed': 0, 'num_leaves': 16, 'learning_rate': 0.05,
          'metric': 'binary_logloss', 'verbose': 2, 'linear_tree': False}
t = time.time()
res2 = {}
lgb_train = lgb.Dataset(train.loc[train_bool & ~valid_bool, list(range(1, 19))], label=train.loc[train_bool & ~valid_bool, 0])
lgb_test = lgb.Dataset(train.loc[valid_bool, list(range(1, 19))], label=train.loc[valid_bool, 0])
t0 = time.time()
time_arr_nonlinear = []
timer_callback = lambda env: time_arr_nonlinear.append(time.time() - t0)
est = lgb.train(params, lgb_train, valid_sets=valid_sets, valid_names=valid_names, num_boost_round=500,
                categorical_feature=[], evals_result=res2, callbacks=[timer_callback])
nonlinear_time = time.time() - t

# plot loss vs iterations
plt.figure()
plt.plot(res['train']['binary_logloss'])
plt.plot(res['valid']['binary_logloss'])
plt.plot(res2['train']['binary_logloss'])
plt.plot(res2['valid']['binary_logloss'])
plt.legend(['train_linear', 'test_linear', 'train', 'test'])

# plot loss vs time
plt.figure()
plt.plot(time_arr_linear, res['train']['binary_logloss'])
plt.plot(time_arr_linear, res['valid']['binary_logloss'])
plt.plot(time_arr_nonlinear, res2['train']['binary_logloss'])
plt.plot(time_arr_nonlinear, res2['valid']['binary_logloss'])
plt.legend(['train_linear', 'test_linear', 'train', 'test'])

# evaluate performance on hold-out set
t = time.time()
p_linear = est_linear.predict(train.loc[~train_bool, list(range(1, 19))], num_boost_round=est_linear.best_iteration)
linear_pred_time = time.time() - t
print("Linear model")
print("training time ", linear_time)
print("prediction time ", linear_pred_time)
print("log loss ", log_loss(train.loc[~train_bool, 0], p_linear))
print("auc ", roc_auc_score(train.loc[~train_bool, 0], p_linear))

t = time.time()
p = est.predict(train.loc[~train_bool, list(range(1, 19))], num_boost_round=est.best_iteration)
nonlinear_pred_time = time.time() - t
print("\n\nConstant model")
print("training time ", nonlinear_time)
print("prediction time ", nonlinear_pred_time)
print("log loss ", log_loss(train.loc[~train_bool, 0], p))
print("auc ", roc_auc_score(train.loc[~train_bool, 0], p))

plt.title('Accuracy vs time on SUSY dataset (500 iterations)', fontsize=12)
plt.xlabel('Time (seconds)', fontsize=12)
plt.ylabel('Log loss', fontsize=12)

plt.figure()
plt.scatter(p, p_linear)
plt.xlim([0, 0.2])
plt.ylim([0, 0.2])
```
Output:
```
Linear model
training time  101.74181818962097
prediction time  3.5460383892059326
log loss  0.4245472152015672
auc  0.8775945941602206

Constant model
training time  89.05728363990784
prediction time  2.811807155609131
log loss  0.4258133037378132
auc  0.8769452618339838
Out[3]: (0, 0.2)
```

Full training graph:
susy_test

Training graph zoomed in on y-axis to show different convergence:
susy_test_zoom


guolinke:
@btrotta 
Thank you so much! 
BTW, it seems there are lots of Eigen codes, maybe it is better to include it as submodule?

guolinke:
also cc @shiyu1994 for the GBDT-PL

src/treelearner/serial_tree_learner.cpp
```diff
@@ -775,4 +859,231 @@ void SerialTreeLearner::RecomputeBestSplitForLeaf(int leaf, SplitInfo* split) {
...
+  if (is_first_tree) {
+    for (int leaf_num = 0; leaf_num < num_leaves; ++leaf_num) {
+      tree->SetLeafConst(leaf_num, tree->LeafOutput(leaf_num));
+    }
+    return;
+  }
+
+  // calculate coefficients using the additive method described in https://arxiv.org/pdf/1802.05640.pdf
+  // the coefficients vector is given by
+  // - (X_T * H * X + lambda) ^ (-1) * (X_T * H * y + g_T X)
```
stanpcf, line 889: thank you for your great work. this line is Eq3 or Eq4 in the paper. I can't understand the `y`.

shiyu1994: It seems that the code below does not use the 'y' stuff. Since the linear models are fitted after the tree is grown, no half-additive is needed.

stanpcf: I only find leave linear model to fit. Is there calc split gain use Eq4. If not use Eq4 (but use lgb default) split gain, the tree split is not corresponded to the loss.

btrotta: Sorry, this comment was for an older version of the code. I've updated it now. It's referring to Eq 3 in the paper.

include/LightGBM/dataset.h
```diff
@@ -710,6 +762,11 @@ class Dataset {
   bool use_missing_;
   bool zero_as_missing_;
   std::vector<int> feature_need_push_zeros_;
+  std::vector<std::vector<double>> raw_data_;
```
guolinke, line 765: Instead of raw float data, is that possible to use the bucketed value (map from int to bucketed float value) for linear function?

btrotta: Yes, I had this idea as well and tried it out. The accuracy was not as good. Also I think that the result could be very sensitive to outliers in the first and last bins which will change the bin average a lot. Using float instead of double is a good idea, I will give that a try.

btrotta:
It seems R cmd check is finding some problems with the Eigen files:
```
File which contains pragma(s) suppressing important diagnostics
  ‘src/eigen/Eigen/src/Core/arch/SSE/Complex.h’
Files which contain pragma(s) suppressing diagnostics:
  ‘src/eigen/Eigen/src/Core/util/DisableStupidWarnings.h’
  ‘src/eigen/test/AnnoyingScalar.h’ ‘src/eigen/test/indexed_view.cpp’
  ‘src/eigen/test/packetmath_test_shared.h’
  ‘src/eigen/test/symbolic_index.cpp’
  ‘src/eigen/test/vectorization_logic.cpp
```
Since I haven't implemented the R interface anyway, I have used the preprocessor directive `LGB_R_BUILD` to avoid compiling the code which requires Eigen. We would need to revisit this if/when someone implements the R interface.

src/treelearner/serial_tree_learner.cpp
```diff
@@ -66,6 +72,48 @@ void SerialTreeLearner::Init(const Dataset* train_data, bool is_constant_hessian
     cegb_.reset(new CostEfficientGradientBoosting(this));
     cegb_->Init();
   }
+
+  if (config_->linear_tree) {
+    InitLinear(train_data, config_->num_leaves);
+  }
+}
+
+void SerialTreeLearner::InitLinear(const Dataset* train_data, const int max_leaves) {
```
jameslamb, line 81: @btrotta I'm very inexperienced with C++, could you help me understand how the R package tests are working? Shouldn't compilation here fail if `Eigen/Dense`  isn't included? And won't it not be included for the R package because of `ifndef LGB_R_BUILD`?

docs/Parameters.rst
```diff
@@ -117,6 +117,20 @@ Core Parameters
...
+
+   -  fit piecewise linear gradient boosting tree, only works with cpu and serial tree learner
+
+      -  tree splits are chosen in the usual way, but the model at each leaf is linear instead of constant
+
+      -  missing values must be encoded as ``np.nan`` (Python) or ``NA`` (cli), not ``0``
+
+      -  it is recommended to rescale data before training so that features have similar mean and standard deviation
+
+      -  not yet supported in R-package
```
jameslamb, line 130: what is blocking this from being implemented in R?

jameslamb: oh sorry, I see this in the PR description (missed it the first time):

> I have only implemented this for Python. I think to get it working for R it would require some changes to the data loading interface, but I'm not very familiar with R, so maybe someone else would like to take that on.

Could you please open an issue with tag `r-package`  describing the changes you think are needed in R to support this feature? Then we can implement it later without blocking this PR.

src/treelearner/serial_tree_learner.cpp
btrotta: The only code that uses `Eigen/Dense` is the method `CalculateLinear`. This method only gets called if `linear_tree=True`. If `LGB_R_BUILD` is defined, then the code at line 878 (https://github.com/microsoft/LightGBM/pull/3299/files/4410c9cd1313ffc7333ca077e488296baac6f0ea#diff-110aae7c40f3150eb5d8264c1cb0663eR878) means that the compiler will compile a different version of `CalculateLinear` that just throws an error if it gets called (which would happen if an R-package user tried to set `linear_tree=True`).

docs/Parameters.rst
btrotta: Done in #3319

guolinke:
@btrotta thanks very much! is this ready to review?

windows/LightGBM.vcxproj
```diff
@@ -244,7 +250,6 @@
     <ClInclude Include="..\include\LightGBM\prediction_early_stop.h" />
     <ClInclude Include="..\include\LightGBM\tree.h" />
     <ClInclude Include="..\include\LightGBM\tree_learner.h" />
-    <ClInclude Include="..\include\LightGBM\utils\yamc\alternate_shared_mutex.hpp" />
```
guolinke, line 247: I think these are false deletion.

btrotta:
@guolinke Yes, ready for review

windows/LightGBM.vcxproj
btrotta: Yes, sorry, not sure how that happened.

include/LightGBM/dataset.h
```diff
@@ -674,6 +689,43 @@ class Dataset {
 
   void AddFeaturesFrom(Dataset* other);
 
+  /*! \brief Get has_raw_ */
+  inline double has_raw() const { return has_raw_; }
```
guolinke, line 693: why this is double/

include/LightGBM/dataset.h
```diff
@@ -674,6 +689,43 @@ class Dataset {
...
+      raw_data_.resize(num_numeric_features_);
+    }
+    for (size_t i = 0; i < raw_data_.size(); ++i) {
+      raw_data_[i].resize(num_rows);
+    }
+    int curr_size = raw_data_.size();
+    for (int i = curr_size; i < num_features_; ++i) {
+      int feat_num = numeric_feature_map_[i];
+      if (feat_num > -1) {
+        raw_data_.push_back(std::vector<float>(num_rows, 0));
```
guolinke, line 719: so, categorical features is not supported? maybe we should document it.

include/LightGBM/tree.h
```diff
@@ -501,20 +580,65 @@ inline void Tree::Split(int leaf, int feature, int real_feature,
...
-    int leaf = GetLeaf(feature_values);
-    return LeafOutput(leaf);
+  if (is_linear_) {
+      int leaf = GetLeaf(feature_values);
+      double output = leaf_const_[leaf];
+      bool nan_found = false;
+      for (size_t i = 0; i < leaf_features_[leaf].size(); ++i) {
+        int feat_raw = leaf_features_[leaf][i];
+        double feat_val = feature_values[feat_raw];
+        if (std::isnan(feat_val)) {
```
guolinke, line 590: is that okay to convert all `nan` to zeros, to avoid the constant conversion? which solution is better?

python-package/lightgbm/basic.py
```diff
@@ -1979,11 +1979,13 @@ def __init__(self, params=None, train_set=None, model_file=None, model_str=None,
...
-            # construct booster object
-            train_set.construct()
             # copy the parameters from train_set
-            params.update(train_set.get_params())
+            if train_set.get_params():
+                params.update(train_set.get_params())
             params_str = param_dict_to_str(params)
+            train_set._update_params(params)
+            # construct booster object
+            train_set.construct()
```
guolinke, line 1988: why about change is needed? also cc @StrikerRUS

guolinke: is `linear_tree` parameter needed in both the dataset and booster? I remember we fix that problem in https://github.com/microsoft/LightGBM/pull/2594 , and you change this accordingly.

src/boosting/gbdt.cpp
```diff
@@ -269,6 +275,16 @@ void GBDT::RefitTree(const std::vector<std::vector<int>>& tree_leaf_prediction)
...
   int num_iterations = static_cast<int>(models_.size() / num_tree_per_iteration_);
   std::vector<int> leaf_pred(num_data_);
+  if (linear_tree_) {
+    int max_leaves = 0;
+    for (size_t i = 0; i < tree_leaf_prediction.size(); ++i) {
+      for (size_t j = 0; j < tree_leaf_prediction[i].size(); ++j) {
+        max_leaves = std::max(max_leaves, tree_leaf_prediction[i][j]);
+      }
+    }
+    max_leaves += 1;
```
guolinke, line 285: why `+1` here?

src/boosting/score_updater.hpp
```diff
@@ -85,7 +85,7 @@ class ScoreUpdater {
   * \param tree_learner
   * \param cur_tree_id Current tree for multiclass training
   */
-  inline void AddScore(const TreeLearner* tree_learner, const Tree* tree, int cur_tree_id) {
+  inline void AddScore(TreeLearner* tree_learner, const Tree* tree, int cur_tree_id) {
```
guolinke, line 88: tree learner will be changed in score updating?

src/treelearner/serial_tree_learner.h
```diff
@@ -215,6 +286,17 @@ class SerialTreeLearner: public TreeLearner {
   const Json* forced_split_json_;
   std::unique_ptr<TrainingShareStates> share_state_;
   std::unique_ptr<CostEfficientGradientBoosting> cegb_;
+  /*! \brief whether numerical features contain any nan values, used for linear model */
+  std::vector<int8_t> contains_nan_;
+  /*! whether any numerical feature contains a nan value, used for linear model */
+  bool any_nan_;
+  /*! \brief map dataset to leaves, used for linear model */
+  std::vector<int> leaf_map_;
```
guolinke, line 294: is this different with DataPartition?

src/treelearner/serial_tree_learner.h
```diff
@@ -215,6 +286,17 @@ class SerialTreeLearner: public TreeLearner {
...
+  /*! \brief temporary storage for calculating linear model */
+  std::vector<std::vector<float>> XTHX_;
+  std::vector<std::vector<float>> XTg_;
+  std::vector<std::vector<std::vector<float>>> XTHX_by_thread_;
+  std::vector<std::vector<std::vector<float>>> XTg_by_thread_;
```
guolinke, line 299: above temporary storage could be mutable, so that we can make tree_learner const.

src/treelearner/serial_tree_learner.cpp
```diff
@@ -775,4 +862,236 @@ void SerialTreeLearner::RecomputeBestSplitForLeaf(int leaf, SplitInfo* split) {
...
+    if (is_refit) {
+      double old_const = tree->LeafConst(leaf_num);
+      tree->SetLeafConst(leaf_num, decay_rate * old_const + (1.0 - decay_rate) * coeffs(num_feat) * shrinkage);
+    } else {
+      tree->SetLeafConst(leaf_num, coeffs(num_feat));
+    }
+    tree->SetNan(data_has_nan);
+  }
+}
+#endif  // LGB_R_BUILD
```
guolinke, line 1095: maybe we could have a separate file for linear tree-related calculation, also move these temporary storages into it.

include/LightGBM/dataset.h
btrotta: Mistake, fixed.

include/LightGBM/dataset.h
btrotta: Added some more documentation in `config.h`. Categorical features are still used in splits as normal, but the linear models at each leaf only use numerical features.

include/LightGBM/dataset.h
```diff
@@ -387,6 +387,12 @@ class Dataset {
         const int group = feature2group_[feature_idx];
         const int sub_feature = feature2subfeature_[feature_idx];
         feature_groups_[group]->PushData(tid, sub_feature, row_idx, feature_values[i]);
+        if (has_raw_) {
+          int num_feat = numeric_feature_map_[feature_idx];
+          if (num_feat > -1) {
```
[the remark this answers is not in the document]
btrotta: Fixed.

include/LightGBM/dataset.h
```diff
@@ -402,6 +408,12 @@ class Dataset {
         const int group = feature2group_[feature_idx];
         const int sub_feature = feature2subfeature_[feature_idx];
         feature_groups_[group]->PushData(tid, sub_feature, row_idx, inner_data.second);
+        if (has_raw_) {
+          int num_feat = numeric_feature_map_[feature_idx];
+          if (num_feat > -1) {
+            raw_data_[num_feat][row_idx] = inner_data.second;
```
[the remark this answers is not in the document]
btrotta: Fixed.

include/LightGBM/dataset.h
```diff
@@ -674,6 +689,43 @@ class Dataset {
 
   void AddFeaturesFrom(Dataset* other);
 
+  /*! \brief Get has_raw_ */
+  inline double has_raw() const { return has_raw_; }
+
+  /*! \brief Set has_raw_ */
+  inline void SetRaw(bool has_raw) { has_raw_ = has_raw; }
```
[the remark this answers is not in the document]
btrotta: Fixed.

include/LightGBM/tree.h
```diff
@@ -257,6 +279,51 @@ class Tree {
...
+  inline void SetLeafFeaturesInner(int leaf, std::vector<int> features) {
+    leaf_features_inner_[leaf] = features;
+  }
+
+  /*! \brief Set the linear model features on one leaf */
+  inline void SetLeafFeatures(int leaf, std::vector<int> features) {
+    leaf_features_[leaf] = features;
+  }
+
+  inline bool GetLinear() const { return is_linear_; }
```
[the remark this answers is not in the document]
btrotta: Fixed.

include/LightGBM/tree.h
```diff
@@ -257,6 +279,51 @@ class Tree {
...
+
+  inline void SetLinear(bool is_linear) {
```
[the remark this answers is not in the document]
btrotta: Fixed.

src/boosting/score_updater.hpp
btrotta: Sorry, I think this was from an older approach and I forgot to change it back. Fixed now.

btrotta:
Thanks @guolinke. I'm still working through your comments but have fixed some of the easier ones as noted above.

include/LightGBM/dataset.h
```diff
@@ -387,6 +387,12 @@ class Dataset {
         const int group = feature2group_[feature_idx];
         const int sub_feature = feature2subfeature_[feature_idx];
         feature_groups_[group]->PushData(tid, sub_feature, row_idx, feature_values[i]);
+        if (has_raw_) {
+          int num_feat = numeric_feature_map_[feature_idx];
+          if (num_feat > -1) {
+            raw_data_[num_feat][row_idx] = feature_values[i];
```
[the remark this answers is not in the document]
btrotta: I tried this, the runtime is longer (around 210 seconds for the example script above, vs around 110 for column-wise). Obviously it depends on the dataset, but I think column-wise can be quite efficient. If trees tend to use the same features repeatedly, then the relevant part of the feature arrays are already cached. Whereas with row-wise, if there are lots of unused features, when we access data for a feature, we will likely waste some cache space on unused features that are near it in the feature row.

src/boosting/gbdt.cpp
btrotta: The maximum of tree_leaf_prediction is 1 less than the number of leaves in the tree (since leaf predictions are labelled 0, 1, ..., num_leaves - 1).

include/LightGBM/tree.h
```diff
@@ -189,9 +189,21 @@ class Tree {
     for (int i = 0; i < num_leaves_ - 1; ++i) {
       leaf_value_[i] = MaybeRoundToZero(leaf_value_[i] * rate);
       internal_value_[i] = MaybeRoundToZero(internal_value_[i] * rate);
+      if (is_linear_) {
+        leaf_const_[i] = MaybeRoundToZero(leaf_const_[i] * rate);
```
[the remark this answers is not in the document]
btrotta: I don't think so. These values are not the same, and we need both because when we predict for a row containing nan feature values we can't use the linear model and so we predict `leaf_value_`.

src/treelearner/serial_tree_learner.h
btrotta: Done.

include/LightGBM/dataset.h
```diff
@@ -674,6 +689,43 @@ class Dataset {
...
   void AddFeaturesFrom(Dataset* other);
 
+  /*! \brief Get has_raw_ */
+  inline double has_raw() const { return has_raw_; }
+
+  /*! \brief Set has_raw_ */
+  inline void SetRaw(bool has_raw) { has_raw_ = has_raw; }
+
+  /*! \brief Get size of raw data */
+  inline data_size_t get_raw_size() const {
```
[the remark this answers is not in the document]
btrotta: You're right, these are equivalent. I've removed `get_raw_size` now.

include/LightGBM/tree.h
btrotta: I don't think we can change `nan` to zero, this would result in quite different linear models if there are lots of `nan` values.

include/LightGBM/tree.h
```diff
@@ -257,6 +279,51 @@ class Tree {
...
+    leaf_features_[leaf] = features;
+  }
+
+  inline bool GetLinear() const { return is_linear_; }
+
+  inline void SetLinear(bool is_linear) {
+    is_linear_ = is_linear;
+  }
+
+  inline bool has_nan() const { return data_has_nan_; }
```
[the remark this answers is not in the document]
btrotta: It was used in `SerialTreeLearner::AddPredictionToScore`: if there are `nan` values then we need to check each row for `nan` when making the prediction, which is slower. But looking at it again now, it seems like a hack to store this in the tree object, so I changed it to use the already-calculated `any_nan_` and `contains_nan_` properties of `SerialTreeLearner`.

src/treelearner/serial_tree_learner.h
btrotta: Yes, I think so. `DataPartition` stores the data indices sorted by leaf number. `leaf_map_` stores the leaf number of each data index, with the data indices in original order.

src/treelearner/serial_tree_learner.cpp
btrotta: Do you mean a separate class, or just split the code across 2 files?

include/LightGBM/boosting.h
```diff
@@ -312,6 +312,8 @@ class LIGHTGBM_EXPORT Boosting {
   * \return The boosting object
   */
   static Boosting* CreateBoosting(const std::string& type, const char* filename);
+
+  virtual bool IsLinear() const { return false; }
```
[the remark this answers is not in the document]
btrotta: It's used in the `refit` method of the python interface. When we construct the new dataset for refitting, we need to know if the model is linear, so that we can save the full feature data not just the binned data.

python-package/lightgbm/basic.py
btrotta: I've now updated `get_params` to include `linear_tree` so this change is not needed anymore.

include/LightGBM/tree.h
guolinke: I think that discard nan in linear function, is the same as adding zeros.
assuming x2 have lots of nans. 
`a1x1 + a3x3` is the same as `a1x1 + a2 0 + a3x3`

btrotta: If a feature contains nan, then I am not applying the linear function at all for that row, instead I predict leaf_value (same as standard LightGBM model). Also when I train the linear model, I exclude rows where any feature of the linear model is nan.

src/treelearner/serial_tree_learner.cpp
guolinke: yeah, a separate class, and also in another file.

guolinke:
@btrotta BTW, will this affect the training speed when disabling linear_tree?

btrotta:
> 
> 
> @btrotta BTW, will this affect the training speed when disabling linear_tree?

@guolinke No, we don't do any extra calculation unless linear_tree is True. I just checked this by training 5 times in the script above and averaging. Timing below:
Old version: 103.35433449745179 seconds
Code from this PR with linear_tree=False: 102.98684806823731

guolinke:
ping @shiyu1994 for another round of review.

guolinke:
@jameslamb @btrotta 
BTW, is this feature disabled in R ? can this pass the cran checks?

btrotta:
> BTW, is this feature disabled in R ? can this pass the cran checks?

Yes, in the R package if a user tries to use this feature they will get a fatal error. I'm not sure about the CRAN requirements but all the CI checks are passing currently.

include/LightGBM/dataset.h
```diff
@@ -402,6 +408,12 @@ class Dataset {
         const int group = feature2group_[feature_idx];
         const int sub_feature = feature2subfeature_[feature_idx];
         feature_groups_[group]->PushData(tid, sub_feature, row_idx, inner_data.second);
+        if (has_raw_) {
```
shiyu1994, line 411: Shall we consider store the raw feature value in the PushOneData method of Dataset, which is called in c_api.cpp for CSC format data?

include/LightGBM/dataset.h
```diff
@@ -682,6 +697,34 @@ class Dataset {
...
+
+  /*! \brief Resize raw_data_, use current number of features */
+  inline void ResizeRaw(int num_rows) {
+    if (static_cast<int>(raw_data_.size()) > num_numeric_features_) {
+      raw_data_.resize(num_numeric_features_);
+    }
+    for (size_t i = 0; i < raw_data_.size(); ++i) {
+      raw_data_[i].resize(num_rows);
+    }
+    int curr_size = raw_data_.size();
```
shiyu1994, line 714: Add static_cast

include/LightGBM/dataset.h
```diff
@@ -682,6 +697,34 @@ class Dataset {
...
+    for (int i = curr_size; i < num_features_; ++i) {
```
shiyu1994, line 715: Even when the number of features and number of numerical features are not changed, this will always push a new vector in to raw_data_, as long as there is a numerical feature with inner feature index >= cur_size.

src/boosting/gbdt.cpp
```diff
@@ -280,6 +286,16 @@ void GBDT::RefitTree(const std::vector<std::vector<int>>& tree_leaf_prediction)
   CHECK_EQ(static_cast<size_t>(models_.size()), tree_leaf_prediction[0].size());
   int num_iterations = static_cast<int>(models_.size() / num_tree_per_iteration_);
   std::vector<int> leaf_pred(num_data_);
+  if (linear_tree_) {
+    int max_leaves = 0;
+    for (size_t i = 0; i < tree_leaf_prediction.size(); ++i) {
```
shiyu1994, line 291: Need multi-threading to speedup.

src/treelearner/linear_tree_learner.cpp
```diff
@@ -0,0 +1,382 @@
...
+    }
+    leaf_features.push_back(numerical_features);
+    raw_data_ptr.push_back(data_ptr);
+    leaf_num_features.push_back(numerical_features.size());
+    if (static_cast<int>(numerical_features.size()) > max_num_features) {
+      max_num_features = numerical_features.size();
+    }
+  }
+  // clear the coefficient matrices
+  for (int tid = 0; tid < num_threads; ++tid) {
```
shiyu1994, line 230: Need multi-threading.

include/LightGBM/dataset.h
btrotta: I think I've implemented what you're suggesting, but let me know if I misunderstood.

ayun2001:
This linear model can support linear regression ？

guolinke:
I think we should merge this before 3.2.0, ping @shiyu1994 for another round for review.

src/treelearner/linear_tree_learner.cpp
```diff
@@ -0,0 +1,384 @@
...
+void LinearTreeLearner::Init(const Dataset* train_data, bool is_constant_hessian) {
+  SerialTreeLearner::Init(train_data, is_constant_hessian);
+  LinearTreeLearner::InitLinear(train_data, config_->num_leaves);
+}
+
+void LinearTreeLearner::InitLinear(const Dataset* train_data, const int max_leaves) {
+  leaf_map_ = std::vector<int>(train_data->num_data(), -1);
+  contains_nan_ = std::vector<int8_t>(train_data->num_features(), 0);
+  any_nan_ = false;
+  // identify features containing nans
```
shiyu1994, line 26:
```suggestion
  // identify features containing nans
  #pragma omp parallel for schedule(static)
```

src/treelearner/linear_tree_learner.cpp
```diff
@@ -0,0 +1,384 @@
...
+  for (int feat = 0; feat < train_data->num_features(); ++feat) {
+    auto bin_mapper = train_data_->FeatureBinMapper(feat);
+    if (bin_mapper->bin_type() == BinType::NumericalBin) {
+      const float* feat_ptr = train_data_->raw_index(feat);
+      for (int i = 0; i < train_data->num_data(); ++i) {
+        if (std::isnan(feat_ptr[i])) {
+          contains_nan_[feat] = 1;
+          any_nan_ = true;
```
shiyu1994, line 34:
```suggestion
```

src/treelearner/linear_tree_learner.cpp
```diff
@@ -0,0 +1,384 @@
...
+          break;
+        }
+      }
+    }
+  }
```
shiyu1994, line 39:
```suggestion
  }
  for (int feat = 0; feat < train_data->num_features(); ++feat) {
    if (contains_nan_[feat]) {
      any_nan_ = true;
      break;
    }
  }
```

src/treelearner/linear_tree_learner.cpp
```diff
@@ -0,0 +1,384 @@
...
+      }
+      OMP_LOOP_EX_END();
+    }
+  }
+  OMP_THROW_EX();
+  auto total_nonzero = std::vector<int>(tree->num_leaves());
+  // aggregate results from different threads
+  for (int tid = 0; tid < num_threads; ++tid) {
+    for (int leaf_num = 0; leaf_num < num_leaves; ++leaf_num) {
+      int num_feat = leaf_features[leaf_num].size();
```
shiyu1994, line 303:
```suggestion
  #pragma omp parallel for schedule(static)
  for (int leaf_num = 0; leaf_num < num_leaves; ++leaf_num) {
    int num_feat = leaf_features[leaf_num].size();
    for (int tid = 0; tid < num_threads; ++tid) {
```

shiyu1994:
This PR is ready for merge. @btrotta Could you resolve the conflicts? And there're two more suggestions in linear_tree_learner.cpp to use multi-threading. Thanks a lot.

include/LightGBM/dataset.h
```diff
@@ -682,6 +703,31 @@ class Dataset {
...
+
+  /*! \brief Set has_raw_ */
+  inline void SetHasRaw(bool has_raw) { has_raw_ = has_raw; }
+
+  /*! \brief Resize raw_data_ */
+  inline void ResizeRaw(int num_rows) {
+    if (static_cast<int>(raw_data_.size()) > num_numeric_features_) {
+      raw_data_.resize(num_numeric_features_);
+    }
+    for (size_t i = 0; i < static_cast<int>(raw_data_.size()); ++i) {
```
shiyu1994, line 717:
```suggestion
    for (size_t i = 0; i < raw_data_.size(); ++i) {
```

shiyu1994: To remove compilation warning.

include/LightGBM/tree_learner.h
```diff
@@ -36,6 +36,9 @@ class TreeLearner {
   */
   virtual void Init(const Dataset* train_data, bool is_constant_hessian) = 0;
 
+  /*! Initialise some temporary storage, only needed for the linear tree; needs to be a method of TreeLearner since we call it in GBDT::RefitTree */
+  virtual void InitLinear(const Dataset* train_data, const int max_leaves) {}
```
shiyu1994, line 40:
```suggestion
  virtual void InitLinear(const Dataset* /*train_data*/, const int /*max_leaves*/) {}
```
To remove compilation warning.

src/treelearner/serial_tree_learner.cpp
```diff
@@ -149,7 +149,7 @@ void SerialTreeLearner::ResetConfig(const Config* config) {
   constraints_.reset(LeafConstraintsBase::Create(config_, config_->num_leaves, train_data_->num_features()));
 }
 
-Tree* SerialTreeLearner::Train(const score_t* gradients, const score_t *hessians) {
+Tree* SerialTreeLearner::Train(const score_t* gradients, const score_t *hessians, bool is_first_tree) {
```
shiyu1994, line 152:
```suggestion
Tree* SerialTreeLearner::Train(const score_t* gradients, const score_t *hessians, bool /*is_first_tree*/) {
```
To remove compilation warning.

shiyu1994:
> This linear model can support linear regression ？

From the code, it supports regression, except regression with L1 loss.

btrotta:
@shiyu1994 Thanks for the detailed review! I've merged the changes from master and updated as per your comments.

include/LightGBM/config.h
```diff
@@ -148,6 +148,16 @@ struct Config {
...
   std::string boosting = "gbdt";
 
+  // desc = fit piecewise linear gradient boosting tree, only works with cpu and serial tree learner
+  // descl2 = tree splits are chosen in the usual way, but the model at each leaf is linear instead of constant
+  // descl2 = the linear model at each leaf includes all the numerical features in that leaf's branch
+  // descl2 = categorical features are used for splits as normal but are not used in the linear models
+  // descl2 = missing values must be encoded as ``np.nan`` (Python) or ``NA`` (cli), not ``0``
+  // descl2 = it is recommended to rescale data before training so that features have similar mean and standard deviation
+  // descl2 = not yet supported in R-package
+  // descl2 = ``regression_l1`` objective is not supported with linear tree boosting
```
guolinke, line 158: It seems the linear tree will increase the memory usage. I think we should note this in doc

include/LightGBM/dataset.h
```diff
@@ -337,6 +337,12 @@ class Dataset {
         const int group = feature2group_[feature_idx];
         const int sub_feature = feature2subfeature_[feature_idx];
         feature_groups_[group]->PushData(tid, sub_feature, row_idx, feature_values[i]);
+        if (has_raw_) {
+          int feat_ind = numeric_feature_map_[feature_idx];
+          if (feat_ind >= 0) {
+            raw_data_[feat_ind][row_idx] = feature_values[i];
```
guolinke, line 343: why not use row-first raw data? I guess linear model is faster with row-first data?

btrotta: Yes, see previous comment: https://github.com/microsoft/LightGBM/pull/3299#discussion_r482742127

btrotta: See previous comment: https://github.com/microsoft/LightGBM/pull/3299#discussion_r482742127

guolinke: great, i missed it.

guolinke:
@shiyu1994 you can approve this PR if there are no problems.

shiyu1994:
Tuned the PR on small subsets of our benchmark datasets. Following shows the results. Linear leaves do improve the results on some datasets. Since these datasets are small, we expect linear models to overfit the data easier. And the features are not normalized, this can also affect the accuracy of linear leaves. 

No bug is reported in all the tuned cases, whether use linear models or not. I'll merge this PR.

| leaf type | higgs | yahoo | msltr | dataexpo | allstate | year |
| ---- | ---- | ---- | ---- | ---- | ---- | ---- |
| - | AUC | NDCG@5 |NDCG@5 | AUC| AUC| RMSE|
| constant | 0.820457<br>±0.000496| **0.726132**<br>±0.002645 | 0.427234<br>±0.006895 | **0.717934**<br>±0.001222 | **0.537855**<br>±0.017714| 8.832463<br>±0.012105|
|linear | **0.822166**<br>±0.000633 | 0.722051<br>±0.001225 | **0.430325**<br>±0.004525 | 0.716811<br>±0.000925 | 0.535730<br>±0.015776 | **8.816403**<br>±0.010700|

shiyu1994 (approved):

.ci/test.sh
```diff
@@ -52,8 +52,8 @@ if [[ $TRAVIS == "true" ]] && [[ $TASK == "lint" ]]; then
             "r-lintr>=2.0"
     pip install --user cpplint
     echo "Linting Python code"
-    pycodestyle --ignore=E501,W503 --exclude=./compute,./.nuget,./external_libs . || exit -1
-    pydocstyle --convention=numpy --add-ignore=D105 --match-dir="^(?!^compute|external_libs|test|example).*" --match="(?!^test_|setup).*\.py" . || exit -1
+    pycodestyle --ignore=E501,W503 --exclude=./compute,./eigen,./.nuget,./external_libs . || exit -1
+    pydocstyle --convention=numpy --add-ignore=D105 --match-dir="^(?!^compute|^eigen|external_libs|test|example).*" --match="(?!^test_|setup).*\.py" . || exit -1
```
jameslamb, line 56: /gha run r-valgrind

jameslamb: https://github.com/microsoft/LightGBM/runs/1613120058?check_suite_focus=true

gianpDomiziani:
is linear_tree available with the 3.1.1 version of the python API?

ChipKerchner:
@StrikerRUS @guolinke IMO this pull introduced a bug or unexpected behavior in the class method SizesInByte.  It no long returns the data_.size() but the AlignedSize(data_.size()).  This is BAD if you are using it for allocation or copying from a desire location in the data.  In the past, SizesInByte was NEVER BIGGER than num_data_ but NOW in some case it is.  There should be some way of getting the original data_.size().

shiyu1994:
@ChipKerchner It seems that the AlignedSize(data_.size()) in SizesInByte is not introduced by this PR, but #3415

StrikerRUS:
@shiyu1994 @guolinke 
Related to the `AlignedSize` problem: https://github.com/microsoft/LightGBM/issues/3450#issuecomment-758583425.

JoshuaC3:
When will this go to pip please? Really interested to give this a try!

jameslamb:
The next formal release is being tracked in #3872 , you can subscribe to that for updates. We cannot give an exact date at this time, but you can see a list there of the work that still needs to be done.

StrikerRUS:
@JoshuaC3 
> When will this go to pip please? Really interested to give this a try!

For now please feel free to install from nightly wheels.

https://lightgbm.readthedocs.io/en/latest/Installation-Guide.html

image

JoshuaC3:
@StrikerRUS I only looked at the git pip install on the pip website itself. This broke because "Windows build issues". I will give the Wheels a spin!!

Also, @jameslamb, I recognised you name but wasn't sure where from. I've just realised!! [recent-developments-in-lightgbm](http://datascience.la/recent-developments-in-lightgbm/). Excellent video so thanks. Hopefully we get another one with 4.0 😉🤞 

Thank both/all 👏

spiralulam:
Hi @btrotta, thanks for your work! Is it possible to access the coefficients and offsets of the linear models at each leaf? I did not find this information in e.g. the `dump_model()` method.

shiyu1994:
Hi @spiralulam, thanks for using LightGBM. If you dump the tree model with `linear_tree` enabled, you should see an entry named `leaf_coeff` in the part of each tree. That's the coefficients of the linear leaves.

spiralulam:
> Hi @spiralulam, thanks for using LightGBM. If you dump the tree model with `linear_tree` enabled, you should see an entry named `leaf_coeff` in the part of each tree. That's the coefficients of the linear leaves.

Thanks for the answer. Does that mean I basically have to parse a string to obtain these information using `save_model()`? The `dump_model()` method which returns a dictionary does not contain `leaf_coeff` nor `leaf_const`.

shiyu1994:
Just checked the method that dumps a tree to JSON. Unfortunately, information of linear leaves are not handled. So currently parsing from the model text file seems to be the only solution. Sorry for the inconvenient.

This should be a very useful case when using linear trees, and I believe we should provide direct access to linear model coefficients through C++, Python and R API.

spiralulam:
That would be awesome, indeed.

cc22226:
Hi @btrotta, Thx for your work! Does the code allow using a subset of the features for the tree splits and another totally different to estimate the linear model at each leaf?

shiyu1994:
@cc22226 Thanks for using LightGBM. Currently the linear models at leaves will consider all the numerical features (or non-categorical features). And there's no parameter to control which features are used in the linear models and which are used in the splits. I think it would be nice sometimes to have these two parts of features being separated. And maybe we can leave that as a feature request.

cc22226:
Thanks for your response.
Cesar

PR #3306, opened 2020-08-14
Updated network retry delay strategy to scale

### LINKED ISSUE #3301: Make MPI ring connection retry count configurable
## Summary

Currently the maximum number of retries is [20](https://github.com/microsoft/LightGBM/blob/master/src/network/linkers_socket.cpp#L189), which might seem enough for normal cases, but when running with large number of machines, this isn't enough. If this can be made into a parameter that can be passed to the config, that'll be helpful

aakarshg:
This allows for network retries, to scale well with the
number of machines, and still retains the existing functionality
for cases with smaller num_machines ( 500 )

Fixes #3301

### DIFF
--- a/src/network/linkers_socket.cpp
+++ b/src/network/linkers_socket.cpp
@@ -8,9 +8,10 @@
 #include <LightGBM/utils/common.h>
 #include <LightGBM/utils/text_reader.h>
 
-#include <string>
+#include <algorithm>
 #include <chrono>
 #include <cstring>
+#include <string>
 #include <thread>
 #include <unordered_map>
 #include <unordered_set>
@@ -186,7 +187,9 @@ void Linkers::Construct() {
   listener_->SetTimeout(socket_timeout_);
   listener_->Listen(incoming_cnt);
   std::thread listen_thread(&Linkers::ListenThread, this, incoming_cnt);
-  const int connect_fail_retry_cnt = 20;
+  const int connect_fail_constant_factor = 20;
+  const int connect_fail_retries_scale_factor = static_cast<int>(num_machines_ / connect_fail_constant_factor);
+  const int connect_fail_retry_cnt = std::max(connect_fail_constant_factor, connect_fail_retries_scale_factor);
   const int connect_fail_retry_first_delay_interval = 200;  // 0.2 s
   const float connect_fail_retry_delay_factor = 1.3f;
   // start connect
--- a/include/LightGBM/c_api.h
+++ b/include/LightGBM/c_api.h
@@ -47,7 +47,7 @@ typedef void* FastConfigHandle; /*!< \brief Handle of FastConfig. */
 LIGHTGBM_C_EXPORT const char* LGBM_GetLastError();
 
 /*!
- * \brief Register a callback function for log redirecting. 
+ * \brief Register a callback function for log redirecting.
  * \param callback The callback function to register
  * \return 0 when succeed, -1 when failure happens
  */


src/network/linkers_socket.cpp
```diff
@@ -186,7 +187,9 @@ void Linkers::Construct() {
   listener_->SetTimeout(socket_timeout_);
   listener_->Listen(incoming_cnt);
   std::thread listen_thread(&Linkers::ListenThread, this, incoming_cnt);
-  const int connect_fail_retry_cnt = 20;
+  const int connect_fail_retries_factor_machine = 25;
+  const int connect_fail_retries_scale_factor = static_cast<int>(num_machines_ / connect_fail_retries_factor_machine);
+  const int connect_fail_retry_cnt = std::max(20, connect_fail_retries_scale_factor);
```
guolinke, line 192: so it is 20 when `num_machines_ = 500`, Is this too small?

src/network/linkers_socket.cpp
```diff
@@ -15,6 +15,7 @@
 #include <unordered_map>
 #include <unordered_set>
 #include <vector>
+#include <algorithm>
```
StrikerRUS, line 18: Please use alphabetical order for includes.

src/network/linkers_socket.cpp
aakarshg: it's 20 for `num_machines_` less than 500, correct. It's too small if we're spinning say > 1000 `num_machines_`, in which case time to spin them up and get them running itself takes a while ( especially when we launch them as containers say in kubernetes cluster )

src/network/linkers_socket.cpp
aakarshg: @StrikerRUS fixed it, also organized so that string library is also moved into the alphabetical order.

aakarshg:
@StrikerRUS and @guolinke can you PTAL at the PR again when you get a chance? Thanks :)

guolinke (approved):

StrikerRUS (commented):
I'm not a cpp reviewer, but the concept of this PR looks OK to me!
Just wonder, why do we need two different values `20` and `25`?

guolinke:
@StrikerRUS maybe @aakarshg can have better naming for these variates. 
@aakarshg to me, the `25` is strange too. The number of nodes in distributed learning usually is the power of 2.

aakarshg:
@StrikerRUS I put in 2 different values 20 and 25, for following reasons:
1. 20 is the current max number of retries, and i wanted to maintain the same upto the scale of 500 nodes so that it'll still fail early at small scale.
2. I'm dividing number of machines by 25, so that upto 500 nodes ... (500/25) the maximum value of (num_machines/25, 20 ) will be 25 only.

But if we are more than 525 machines then the number of retries will be 21 and then scale up accordingly as we increase num_machines. 

Hope that explains things. 

About the number of nodes being usually power of 2, that's not true tbh. There are cases ( can't quite go into detail ) where num_machines is more tightly died to number of files needed to train data on. I was looking at around 900 or so machines :)

guolinke:
@aakarshg 
LightGBM uses many collective communication algorithms, which work better with power of 2 machines.

aakarshg:
> I find both 400 and 500 machines in cluster very big number, IMHO. So I don't see any difference between 400 and 500. Is 500 something special threshold?

nothing really, its more of an opinion. 

> @aakarshg
> 
> > I'm dividing number of machines by 25
> 
> Why can't number of machines be divided by `20`? I'm asking because I suppose that 1 magic number in code is better that 2 ones.

agreed, updated the PR. PTAL again thanks :)

src/network/linkers_socket.cpp
```diff
@@ -186,7 +187,8 @@ void Linkers::Construct() {
   listener_->SetTimeout(socket_timeout_);
   listener_->Listen(incoming_cnt);
   std::thread listen_thread(&Linkers::ListenThread, this, incoming_cnt);
-  const int connect_fail_retry_cnt = 20;
+  const int connect_fail_retries_scale_factor = static_cast<int>(num_machines_ / 20);
+  const int connect_fail_retry_cnt = std::max(20, connect_fail_retries_scale_factor);
```
StrikerRUS, line 191: @aakarshg Thank you for addressing comments! I'm not sure but maybe it will be better to store `20` in a constant? If we will need to update this value, that constant will allow us to do it in only one place.

guolinke: ping @aakarshg

jameslamb: gently ping @aakarshg

aakarshg: yes that'll be absolutely okay, i'll update the pr :) and apologies for late replies.

StrikerRUS: Thank you!

PR #3305, opened 2020-08-14
fix zero bin in categorical split

guolinke:
to fix #3273

PR #3264, opened 2020-07-29
Pr4 advanced method monotone constraints

CharlesAuguste:
@aldanor @redditur

@guolinke @jameslamb @StrikerRUS 
Following PRs #2305, #2717, #2770, and #2939 here is the last PR of the series. #2305 was judged to be not merge-able because it was too big. Therefore, in my ultimate comment I said I would split it into smaller PRs easier to merge. #2717 was some refactoring, #2770 was the intermediate method, #2939 was the monotone penalization, and this PR is about the advanced method.

As a reminder, the basic method over-constrains leaves, and creates "unused space" between them. The intermediate method is smarter in updating the constraints and creates less "space" between the leaves, but it doesn't take into account that depending on thresholds, a leaf on the right and on the left of a monotone splits may not have to constrain each other. This advanced method takes it into account, and recomputes leaves constraints from the beginning when needed. This leads in a more expressive model, with a meaningfully better training loss. More details in the original report https://github.com/microsoft/LightGBM/files/3457826/PR-monotone-constraints-report.pdf. Feel free to ask for more details. 

About the code, most changes happen in the monotone_constraints.hpp file, where the new AdvancedLeafConstraints has been added. Tests have been updated to include this new constraining method. 

2 things in the original PR that are missing in here:
- The code was more optimized for time in the original PR, and the method is a bit slower here. Though the optimizations made the code significantly more complicated, and I think the time saved was not much. I can implement the optimization in another PR later, as I think this PR is complicated enough as is.
- In the original PR, once a tree is build, there was a final leaves refitting, that I have not implemented here yet. I can either implement it here or in another PR later. This can slightly improve performance.

I am looking forward to your review. Thanks.

PR #3308, opened 2020-08-15
[ci] fix flaky R CMD check test about library size

jameslamb:
We now have two PRs hitting CI failures on R builds for this issue:

```text
  installed size is  5.2Mb
    libs   4.5Mb
```

https://github.com/microsoft/LightGBM/pull/3305#issuecomment-674348634, https://github.com/microsoft/LightGBM/pull/3264#issuecomment-671509757

That note is very unpredictable and likely to be allowed by CRAN. This PR proposes allowing it in CI jobs.

### DIFF
--- a/.ci/test_r_package.sh
+++ b/.ci/test_r_package.sh
@@ -199,11 +199,7 @@ if grep -q -R "WARNING" "$LOG_FILE_NAME"; then
     exit -1
 fi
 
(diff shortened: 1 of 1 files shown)


jameslamb:
the `check-docs` failure looks legitimate

image

jameslamb:
it seems `research.microsoft.com` is unavailable 😬

PR #3311, opened 2020-08-15
[ci] Test 32-bit builds for R package

jameslamb:
Now that #3307 has ben merged and the R package is passing `R CMD check` for both `x64` and `i386` architectures, this PR proposes testing that in our CI. Since the 32-bit build isn't enable when  building with CMake (and I think we should not enable it), as of this PR `R CMD check` would be run without `--no-multiarch` only for CRAN builds.

This PR also proposes removing the `--no-multiarch` option from the documentation for installing with `{remotes}`. If we do that, I'd want to update the artifact `lightgbm-3.0.0-1-r-cran.tar.gz` on release `3.0.0rc`. @guolinke is that ok? Generally I think it's bad to update artifacts once they're published, but in this case since it's something very new and experimental, since its only a pre-release, and since the new artifact would't have any breaking changes, I think it's worth it to remove the need for that flag from the documentation.

### DIFF
--- a/.ci/test_r_package_windows.ps1
+++ b/.ci/test_r_package_windows.ps1
@@ -158,8 +158,14 @@ if ($env:COMPILER -ne "MSVC") {
     cd "C:\$R_CMD_CHECK_DIR\"
   }
 
-  Write-Output "Running R CMD check as CRAN"
-  Run-R-Code-Redirect-Stderr "result <- processx::run(command = 'R.exe', args = c('CMD', 'check', '--no-multiarch', '--as-cran', '--run-dontrun', '$PKG_FILE_NAME'), echo = TRUE, windows_verbatim_args = FALSE)" ; $check_succeeded = $?
+  Write-Output "Running R CMD check"
+  if ($env:R_BUILD_TYPE -eq "cran") {
+    # CRAN packages must pass without --no-multiarch (build on 64-bit and 32-bit)
+    $check_args = "c('CMD', 'check', '--as-cran', '--run-dontrun', '$PKG_FILE_NAME')"
+  } else {
+    $check_args = "c('CMD', 'check', '--no-multiarch', '--as-cran', '--run-dontrun', '$PKG_FILE_NAME')"
+  }
+  Run-R-Code-Redirect-Stderr "result <- processx::run(command = 'R.exe', args = $check_args, echo = TRUE, windows_verbatim_args = FALSE)" ; $check_succeeded = $?
 
   Write-Output "R CMD check build logs:"
   $INSTALL_LOG_FILE_NAME = "lightgbm.Rcheck\00install.out"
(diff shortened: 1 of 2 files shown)


guolinke:
@jameslamb I think it is okay

jameslamb:
ok I just updated `lightgbm_3.0.0-1-r-cran.tar.gz` on the `3.0.0rc` release, so this PR's docs changes will be accurate. `--no-multiarch` flag can be safely omitted

jameslamb:
I'm very worried about the Travis tests

image

I wonder if they've finally ended support for the "Legacy" Travis (.org) and we have to switch to .com?

I asked Travis about this a while ago when we faced an issue with stalled builds (https://github.com/microsoft/LightGBM/pull/2957#issuecomment-606350022)

At that time, they recommended moving to travis-ci.com from travis-ci.org (https://docs.travis-ci.com/user/migrate/open-source-repository-migration)

jameslamb:
I'm seeing this Travis issue on other projects I work on too. I've contacted Travis support.

StrikerRUS:
> I wonder if they've finally ended support for the "Legacy" Travis (.org) and we have to switch to .com?

Don't panic! 😃 Seems everything is OK. I rerun one short job to get updates from Travis.

StrikerRUS:
@jameslamb Should we also update docs about that 32-bit is unsopported (change to that it is supported but strongly not reccomended)? 

https://github.com/microsoft/LightGBM/tree/master/R-package#preparation
https://github.com/microsoft/LightGBM/blob/master/docs/GPU-Windows.rst#mingw-correct-compiler-selection

jameslamb:
> @jameslamb Should we also update docs about that 32-bit is unsopported (change to that it is supported but strongly not reccomended)?
> 
> https://github.com/microsoft/LightGBM/tree/master/R-package#preparation
> https://github.com/microsoft/LightGBM/blob/master/docs/GPU-Windows.rst#mingw-correct-compiler-selection

no I don't think so. It won't be supported for CMake installation (https://github.com/microsoft/LightGBM/issues/3187#issuecomment-673239785).

https://github.com/microsoft/LightGBM/tree/master/R-package#preparation already has a lot of stuff that you have to understand, so I'd like to avoid adding additional reading for users there if it's just of the form "here is something that is technically possible but we don't recommend it".

StrikerRUS:
@jameslamb OK, fair enough!

jameslamb:
> > I wonder if they've finally ended support for the "Legacy" Travis (.org) and we have to switch to .com?
> 
> Don't panic!  Seems everything is OK. I rerun one short job to get updates from Travis.

I was panicked haha. I'm glad it's working again!

Looking forward to being scared about this again three months from now :laughing:

StrikerRUS:
@jameslamb 

> I'm seeing this Travis issue on other projects I work on too. I've contacted Travis support.

> Looking forward to being scared about this again three months from now 😆

Have they answered anything?

jameslamb:
nothing yet :/

If they don't get back to me today, I'm going to tag them on Twitter :grimacing:

PR #3314, opened 2020-08-18
Added winning solution using LightGBM

julioasotodv:
Added winning solution using LightGBM for M5 forecasting competition


StrikerRUS:
Thanks a lot @julioasotodv !

It's unbelievable that just single LightGBM (not as a part of ensemble like in all other contests) was used to win the competition!

PR #3315, opened 2020-08-18
[ci][docs] explicitly recuire fresh versions of RTD stuff

StrikerRUS:
Temp fixes for current `master` docs builds... again.

For some reason RTD by default installs old versions of `sphinx_rtd_theme` and `readthedocs-sphinx-ext` which results in

```
...

WARNING: the readthedocs_ext.readthedocs extension does not declare if it is safe for parallel reading, assuming it isn't - please ask the extension author to check and make it explicit
WARNING: doing serial read

...

build finished with problems, 2 warnings.
```

image


StrikerRUS:
Thank you @jameslamb !

PR #3298, opened 2020-08-11
[R-package] DESCRIPTION changes to address CRAN feedback

jameslamb:
This addresses feedback from our first ever submission to CRAN!

https://github.com/microsoft/LightGBM/issues/629#issuecomment-671648923

* The note about "possible misspellings" can be fixed by putting `LightGBM` in quotes. I remember we had to do this when we first submitted [`uptasticsearch`](https://github.com/uptake/uptasticsearch/blob/master/r-pkg/DESCRIPTION#L17)
* the error from "not installed on i386" is from setting `Biarch: false` in `DESCRIPTION`. I thought that flag could be a way to publish a package that doesn't support 32-bit, but I don't think that works. Let's remove that field and we'll get a better error on CRAN 32-bit checks.

This also proposes adding a `cran-comments.md` to track feedback from CRAN. Since the CRAN upload process involves back-and-forth with humans (not just automated checks), a lot of projects keep this type of file to hold a record of the things CRAN has asked them to do in the past. We do this in [`uptasticsearch`](https://github.com/uptake/uptasticsearch/blob/master/cran-comments.md).

PR #3318, opened 2020-08-19
[R-package] remove blank lines and not-really-comments from .Rbuildignore

jameslamb:
I saw this really interesting discussion on the `r-pkg-devel` mailing list: https://stat.ethz.ch/pipermail/r-package-devel/2020q3/005894.html

Basically, the `.Rbuildignore` file does not support comments. Lines that start with `#` will be evaluated as patterns that could match a file.

It also does not skip blank lines, so having blank lines just increases the time it takes to build a package.

This PR should slightly reduce the build time for the R package, and should avoid the unlikely but not impossible case where one of the "commented" lines matches and ignores a file that should not be ignored.

Since I'm touching this file, I also took the opportunity to rearrange it so that it is in alphabetical order (ignoring non-alphanumeric characters).

### DIFF
--- a/R-package/.Rbuildignore
+++ b/R-package/.Rbuildignore
@@ -1,27 +1,19 @@
-^build_r.R$
-\.gitkeep$
-^docs/.*$
-^pkgdown$
-^cran-comments\.md$
-
-# Objects created by compilation
-^.*\.o
-^.*\.so
-^.*\.dll
-^.*\.out
-^.*\.bin
-
-# Code copied in at build time
-^src/CMakeLists.txt$
-^Makefile$
-^src/build/.*$
-^autom4te.cache/.*$
-
-# files only used during development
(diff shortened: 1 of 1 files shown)


StrikerRUS (approved):

StrikerRUS:
Maybe blank line after `cran-comments.md` in #3298 caused problems?

jameslamb:
> Maybe blank line after `cran-comments.md` in #3298 caused problems?

ooooo maybe! now that I know the exact R code that checks `.Rbuildignore`, I can probably come back in a follow-up PR and populate that file correctly. Might allow us to remove some `rm` statements from https://github.com/microsoft/LightGBM/blob/1804fd12ea240af438a8c2d6c67d6a0f9bdf4c5e/build-cran-package.sh#L33-L35

PR #3323, opened 2020-08-19
[ci] bump Xcode version at Travis

StrikerRUS:


StrikerRUS:
Seems that with new image it is possible to get rid of temp workaround for gcc installation introduced in #3208.

StrikerRUS:
Hmm, miniconda installation exceeds 10min no output Travis limit at
```
PREFIX=/Users/travis/miniconda
Unpacking payload ...
```
step after this update. Will try to fix it.

StrikerRUS:
> Hmm, miniconda installation exceeds 10min no output Travis limit

Seems that it is (was) network or another temporary issue. After several reruns all checks are green. It's quite ironic that MPI builds issue happened in the PR with changes under `if [[ $TASK != "mpi" ]]; then` condition 🙂 .

Issue #3293, opened 2020-08-10
[WIP] next release (3.0.0)


guolinke:
This is to continue #3071, we have several new items that are not finished.
I plan to release the 3.0.0 by the end of Sep. 

PRs:
- [ ] #2912
- [ ] #2754
- [ ] #2463
- [ ] #2305 (last remained part)

New Features
- [ ] #2937
- [ ] #2604
- [x] Add some features to R-package that are already in C API and Python package (#2281, #2472, #2369, #2983)
- [ ] CRAN release of course #629

Fixes (may is not included in 3.0.0):
- [ ] All known GPU bugs (#2793, #2648, #2550, #2126, #1717, #1003)
- [ ] MMLSpark issues(#2392, #2363)

guolinke:
cc @shiyu1994 for CTR features (and symmetric trees).

jameslamb:
@guolinke sounds good! I think that to get to CRAN, I really need the help of you and other C++ maintainers on https://github.com/microsoft/LightGBM/issues/3187. I'm trying to get better at C++ but it's beyond my ability to fix that issue by September.

I also think we should attempt a submission to CRAN right now (of 3.0.0-1). I think it's very unlikely that it will be accepted without https://github.com/microsoft/LightGBM/issues/3187, but through that submission we might learn of other things that need to be fixed.

See this comment: https://github.com/microsoft/LightGBM/issues/3187#issuecomment-670966615

guolinke:
@jameslamb no problem. sorry for missing  #3187, I will try to submit it to cran.
@shiyu1994 can you help with the 32-bit build problem?

guolinke:
@jameslamb 
The submit results:
![image](https://user-images.githubusercontent.com/16040950/89774573-be904980-db38-11ea-941c-74d6935e9396.png)

shiyu1994:
> @jameslamb no problem. sorry for missing #3187, I will try to submit it to cran.
> @shiyu1994 can you help with the 32-bit build problem?

Ok.

StrikerRUS:
@guolinke I believe RC candidates are for beta testing features that will be in the corresponding ordinary release. No new features are allowed after RC is released, only bug fixes reported during beta testing period. All features you've mentioned in your staring comment should be included in `3.1.0` version as they are non-breaking ones.

ping @henry0312 to confirm

jameslamb:
> @jameslamb
> The submit results:
> 
[...]

PR #3335, opened 2020-08-25
[R-package] [ci] Reduce manual effort for version updates

jameslamb:
This is the first step towards #3283 . It does not automate creating any artifacts (https://github.com/microsoft/LightGBM/issues/3283#issuecomment-670830603, https://github.com/microsoft/LightGBM/pull/3285#issuecomment-670968342), but it removes some of the manual steps needed to update the R package when the version of LightGBM is updated.

In https://github.com/microsoft/LightGBM/pull/3071#issuecomment-670243598 I proposed adding a `Makefile` with a `make release` recipe, but I decided not to do that in this PR because I thought it would be very thin and therefore not worth adding another file to the repo.

As of this PR, the only change that will need to be made in a release PR is to regenerate `R-package/configure`, which can be done with this command if you have Docker available (documented [here](https://github.com/microsoft/LightGBM/blob/master/R-package/README.md#changing-the-cran-package)):

```shell
docker run \
    -v $(pwd):/opt/LightGBM \
    -t ubuntu:18.04 \
    /bin/bash -c "cd /opt/LightGBM && ./R-package/recreate-configure.sh"
``` 

I tested this with values `3.0.0`, `5.2.1` and `5.2.1rc` in `VERSION.txt`, just to make sure it all works as expected.


build_r.R
```diff
@@ -155,6 +155,31 @@ result <- file.copy(
 )
 .handle_result(result)
 
+# R packages cannot have versions like 3.0.0rc1, but
+# 3.0.0-1 is acceptable
+LGB_VERSION <- readLines("VERSION.txt")[1]
```
guolinke, line 160:
```suggestion
LGB_VERSION <- readLines("VERSION.txt")[1L]
```

guolinke:
@jameslamb I think this will be included in 3.0.0 ?

jameslamb:
yes it should be!

Sorry, should have clarified...I built the package in https://github.com/microsoft/LightGBM/issues/3293#issuecomment-679431463 from this branch, so it is ok.

guolinke (approved):

StrikerRUS:
@jameslamb 

> In #3071 (comment) I proposed adding a Makefile with a make release recipe, but I decided not to do that in this PR because I thought it would be very thin and therefore not worth adding another file to the repo.

Absolutely agree!

R-package/recreate-configure.sh
```diff
@@ -4,6 +4,14 @@
 # this script should run on Ubuntu 18.04
 AUTOCONF_VERSION=$(cat R-package/AUTOCONF_UBUNTU_VERSION)
 
+# R packages cannot have versions like 3.0.0rc1, but
+# 3.0.0-1 is acceptable
+LGB_VERSION=$(cat VERSION.txt | sed "s/rc/-/g")
```
StrikerRUS, line 9: I'm confused a little bit. Where do we copy `VERSION.txt` from the root repo folder to `R-package`? Or how is it working?

StrikerRUS (approved):
Thanks, LGTM if it is working inside a Docker.

R-package/recreate-configure.sh
jameslamb: you don't need to! When you run `./R-package/recreate-configure.sh`, the effective working directory is the root of the repo. So no need to copy it from anywhere.

StrikerRUS: Aha, got it! Thanks for the explanation!

jameslamb:
> Thanks, LGTM if it is working inside a Docker.

thanks! I can confirm that it is. The diff in `configure` in this PR was generated with the `docker run` command in the README. Everything else in this PR will work outside of Docker, for example anywhere that people run `build-cran-package.sh` or `Rscript build_r.R`.

Now that this is automated, it will also fail builds where we update `VERSION.txt` but forget to update `R-package/configure`, which I think is a good thing!

Issue #3332, opened 2020-08-22
[R-package] logo in documentation


jameslamb:
See the following links:

* https://github.com/microsoft/LightGBM/pull/3327#issue-471319731
* https://pkgdown.r-lib.org/reference/build_favicons.html

To close this issue, generate favicons for the `pkgdown` site for the R package.

StrikerRUS:
Closed via #3336.

PR #3327, opened 2020-08-21
[docs] added logo to docs

StrikerRUS:
Added `svg` versions of logo, also uploaded different (`huge`-`large`-`medium`-`small`-`tiny`) versions of `png` images.

In addition `favicon` was created.

Everything is located in `docs` folder to not pollute root folder of the repo.

@jameslamb Seems that `pkgdown` has special function for favicon generation. Probably it worth to investigate. Also, feel free to add logo at R-package pages in this PR directly or in a separate one.

To test: https://lightgbm.readthedocs.io/en/logo/

image

image

PR #3336, opened 2020-08-27
[R-package] Add R logo (fixes #3331, #3332)

### LINKED ISSUE #3331: [R-package] hex logo
now that a logo has been introduced (#3327 ), we should add a hex logo for the R package! These are very very popular in the R world. For example, see RStudio's collection here: https://github.com/rstudio/hex-stickers

I tried creating some tonight in R using the `{hexSticker}` package. That is working really well! here's an example:

image

```r
hexSticker::sticker(
    "docs/logo/LightGBM_logo_black_text_medium.png"
    , package = ""
    , p_size = 9.5
    , p_y = 1.35
    , s_x = 1
    , s_y = 1.0
    , s_width = 0.9
    , filename = "logo/lightgbm-r-hex.png"
    , h_fill = "white"
    , h_color = "#1B9AD7"
    , h_size = 0.9
    , white_around_sticker = FALSE
    , dpi = 600
)
```

But here is my problem...in the R community, the name on the logo is always the name of the package. Our R package is called `lightgbm`, not `LightGBM`.

@StrikerRUS @guolinke what do you think about making a version of the logo that says `lightgbm` instead of `LightGBM`?

jameslamb:
This PR introduces a hex logo for the R package! (#3331) These logos are very popular in the R community (see, for example, https://github.com/rstudio/hex-stickers). Thanks to @guolinke for generating these logos.

It adds this file in `man/figures/logo.png`, the preferred place recognized by `{pkgdown}`.

Following @StrikerRUS 's suggestion in https://github.com/microsoft/LightGBM/pull/3327#issue-471319731, I also generated favicons for the site using `pkgdown::build_favicons()` (#3332).

`{pkgdown}` site looks good (if you squint you can see the favicon in the tab)

image

and this renders nicely in GitHub too!

image


jameslamb:
@StrikerRUS I created this as a LightGBM branch so we can test on readthedocs. Could you enable the build there?

docs/logo/r-package-hex-logo.svg
```diff
@@ -0,0 +1,31 @@
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="819" height="945" viewBox="0 0 819 945">
+  <metadata><?xpacket begin="﻿" id="W5M0MpCehiHzreSzNTczkc9d"?>
+<x:xmpmeta xmlns:x="adobe:ns:meta/" x:xmptk="Adobe XMP Core 5.6-c142 79.160924, 2017/07/13-01:06:39        ">
+   <rdf:RDF xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#">
+      <rdf:Description rdf:about=""/>
+   </rdf:RDF>
+</x:xmpmeta>
```
jameslamb, line 7: @guolinke I see now that this svg just has the PNG data base64-encoded. Is it possible to provide a "true" SVG like this?

https://github.com/microsoft/LightGBM/blob/master/docs/logo/LightGBM_logo_black_text.svg?short_path=a341128

I think that would be more friendly to edit in the future if we need to.

jameslamb: sorry, I have very little experience editing image files  😂

StrikerRUS:
@jameslamb 
> Could you enable the build there?

Sure! Just enabled.
https://readthedocs.org/projects/lightgbm/builds/

jameslamb:
```text
* checking for future file timestamps ... NOTE
unable to verify current time
```

huh, never saw that before. Looking into it. Since that problem doesn't seem isolated to a particular operating system, R version, or build type, I would be surprised to learn that it's related to the changes in #3335

docs/logo/r-package-hex-logo.svg
StrikerRUS: I can do this.

StrikerRUS:
I think we should add `LightGBM_logo_hex.svg` into https://github.com/microsoft/LightGBM/tree/master/docs/logo to be consistent with our FAQ-15: `You can find LightGBM's logo in different file formats and resolutions here.` if one want to add hex logo.
@jameslamb Is it possible to copy file from `../docs/logo` to the R-package target location at a build time or provide customized path with the aim to not duplicate files?

docs/logo/r-package-hex-logo.svg
jameslamb: thanks!

jameslamb:
> I think we should add `LightGBM_logo_hex.svg` into https://github.com/microsoft/LightGBM/tree/master/docs/logo to be consistent with our FAQ-15: `You can find LightGBM's logo in different file formats and resolutions here.` if one want to add hex logo.
> @jameslamb Is it possible to copy file from `../docs/logo` to the R-package target location at a build time or provide customized path with the aim to not duplicate files?

I did add the SVG there, but for the `{pkgdown}` stuff to work, I believe your logo file has to just be called `logo.png` and it has to be in `man/figures`. When I tried with other locations, it wasn't found when rendering the `{pkgdown}` site. I think it's ok for the PNG to be in the expected place in the R package (to make `{pkgdown}` work) and for the SVG to be in `docs/logo`.

jameslamb:
> ```
> * checking for future file timestamps ... NOTE
> unable to verify current time
> ```
> 
> huh, never saw that before. Looking into it. Since that problem doesn't seem isolated to a particular operating system, R version, or build type, I would be surprised to learn that it's related to the changes in #3335

This was just reported on the `r-pkg-devel` mailing list. Apparently `R CMD check --as-cran` has a dependency on two small "what time is it right now" services. One of which doesn't work with the way R is asking it for the time, and the other of which is currently down :upside_down_face: 

https://stat.ethz.ch/pipermail/r-package-devel/2020q3/005928.html

jameslamb:
I'm able to reproduce the time problem on my laptop too. If it is still a problem a few hours from now, I'll just bump the number of allowed checks so our CI can passs.

We shouldn't stop all development on LightGBM just because R doesn't know what time it is :joy:

docs/logo/r-package-hex-logo.svg
StrikerRUS: Please try this file. I "fixed" source file as well: made red triangle vector instead of bitmap (it was the reason of embeded image in SVG) and increased size of the page. Now online convertors should work OK with this file, I believe.

StrikerRUS: GitHub doesn't want to allow me to upload those files 😢 

May I create a PR in your repo with files?

image

StrikerRUS:
@jameslamb 
> I did add the SVG there, ...

OK, I see. Seems there is no room for logo path customization. But at least, according to this, logo can be in SVG:
https://github.com/r-lib/pkgdown/blob/3b4d5b3d724d923973cc28f1c975833c8dba0f29/R/build-logo.R#L15

docs/logo/r-package-hex-logo.svg
StrikerRUS: OK, it was not a temporary issue. I cannot upload files with the same error right now.

I forgot that this PR is from main repo's branch, so I just opened #3340.

jameslamb: oh I've seen that problem! I remember having an issue uploading a `.html` file as well. Thanks for the PR, I've merged it here.

jameslamb:
> OK, I see. Seems there is no room for logo path customization. But at least, according to this, logo can be in SVG:
> https://github.com/r-lib/pkgdown/blob/3b4d5b3d724d923973cc28f1c975833c8dba0f29/R/build-logo.R#L15

yep! It can be but it doesn't have to be. I've generated the favicons once from `R-package/man/figures/logo.png`. Are you saying I should re-do it with the SVG?

StrikerRUS:
@jameslamb 
> Are you saying I should re-do it with the SVG?

Ah, no, I think there is no difference from which source those tiny images were generated. I meant place svg in `man` `R-package/man/figures/logo.svg` instead of png to have good scaling.

jameslamb:
oh I see! Ok I can do that right now

StrikerRUS (approved):
Amazing, thank you!

Seems that pkgdown is happy too!

image

image

jameslamb:
WOO awesome!

PR #3338, opened 2020-08-27
[R-package] miscellaneous changes to comply with CRAN requirements

jameslamb:
This PR attempts to address the most recent request for changes from CRAN, https://github.com/microsoft/LightGBM/issues/3293#issuecomment-682001047, on our way to #629

See changes to `cran-comments.md` for details.

PR #3341, opened 2020-08-29
[ci] fix R CMD check issue 'unable to verify current time'

jameslamb:
This PR skips this `R CMD check` check that is breaking builds

```text
* checking for future file timestamps ... NOTE
unable to verify current time
```

You can see https://stat.ethz.ch/pipermail/r-package-devel/2020q3/005930.html for details. Essentially `R CMD check` relies on two small web services that provide a global clock, and this NOTE can show up when those services are unavailable. They are experiencing some issues this week 🙃 

I think this hack is better than just bumping `ALLOWED_CHECK_NOTES`, as described here: https://github.com/microsoft/LightGBM/pull/3338#discussion_r479346363

I've added this fix in #3336 and #3298 , but since it's blocking any other PRs, I thought it made sense to pull it out into its own PR we can merge quickly.

### DIFF
--- a/.ci/test_r_package.sh
+++ b/.ci/test_r_package.sh
@@ -7,6 +7,10 @@ mkdir -p $R_LIB_PATH
 export R_LIBS=$R_LIB_PATH
 export PATH="$R_LIB_PATH/R/bin:$PATH"
 
+# hack to get around this:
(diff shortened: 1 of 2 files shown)


StrikerRUS (approved):
Thanks!

PR #3344, opened 2020-09-01
bump version

guolinke:


guolinke:
@jameslamb could you update R's version accordingly? you can create a new PR if needed.

jameslamb:
We don't necessarily know that the next release will be 3.0.1 instead of 3.1.0, right?

What do you think about the convention where the development version is like 3.0.0.9999 (newer than 3.0.0, older than any future release we do with a 3-part number), and then we change to 3.0.1 or 3.1.0 based on what is new?

If any new features are added between now and the next release, then that release should be 3.1.0. If it is only maintenance tasks, 3.0.1

guolinke:
I like the proposal. With that, we can distinguish the dev and release versions.

.appveyor.yml
```diff
@@ -1,4 +1,4 @@
-version: 3.0.0.{build}
+version: 3.0.1.{build}
```
guolinke, line 1:
```suggestion
version: 3.0.0.99.{build}
```

VERSION.txt
```diff
@@ -1 +1 @@
-3.0.0
+3.0.1
```
guolinke, line 1:
```suggestion
3.0.0.99
```

guolinke:
After merge this, we can start to merge the blocked PRs about new features.

jameslamb:
I'll update the R version right now and make a PR into this one

jameslamb:
I can also upload the R artifacts to the 3.0.0 release manually tonight

jameslamb:
created #3345

StrikerRUS (approved):

Issue #3026, opened 2020-04-27
Add Precision Recall AUC as an metric for binary classification


kdlin:
For binary classification problems, Precision Recall AUC (as opposed to ROC AUC) is a good metric for unbalanced data. Currently, I have to use a custom eval function even though I use logloss as the ojective function. As a result, both logloss and PR-AUC are used to stop training. (I use early_stopping_rounds= 25 to stop the iterations.)

I would like to stop the iterations with just PR-AUC as the metric. Using custom eval function slows down the speed of LightGBM too. Additionally, XGBoost has PR-AUC as a metric. (They called it aucpr.) 

I propose that PR-AUC to be added as a built-in metric.

My workaround is as follows for the time being:

model=lgb.train(params, lgb_train, 
                num_boost_round=2000,
                valid_sets=[lgb_valid],
                feval=f_pr_auc,
                early_stopping_rounds=25,
                verbose_eval=50)

def pr_auc(y_true, probas_pred):
    p, r, _ = precision_recall_curve(y_true, probas_pred)
    return auc(r, p)

def f_pr_auc(probas_pred, y_true):
    probas_pred=sigmoid(probas_pred)
    labels=y_true.get_label()
    p, r, _ = precision_recall_curve(labels, probas_pred)
    score=auc(r,p) 
    return "pr_auc", score, True

btrotta:
Closed in favor of being in #2302. PR is welcome for this feature!

btrotta:
Reopened since there is an active PR.

StrikerRUS:
Closed via #3347.

PR #3347, opened 2020-09-02
Average precision score

btrotta:
Implement average precision score for binary classification. This is very similar to the precision-recall AUC requested in #3026, but it does not interpolate the points of the curve, since this is not recommended for PR curves (see https://scikit-learn.org/stable/modules/generated/sklearn.metrics.average_precision_score.html).

### DIFF
--- a/docs/Parameters.rst
+++ b/docs/Parameters.rst
@@ -1010,7 +1010,7 @@ Metric Parameters
 
       -  ``auc``, `AUC <https://en.wikipedia.org/wiki/Receiver_operating_characteristic#Area_under_the_curve>`__
 
-      -  ``ap``, `average precision score <https://scikit-learn.org/stable/modules/generated/sklearn.metrics.average_precision_score.html>`__
+      -  ``average_precision``, `average precision score <https://scikit-learn.org/stable/modules/generated/sklearn.metrics.average_precision_score.html>`__
 
       -  ``binary_logloss``, `log loss <https://en.wikipedia.org/wiki/Cross_entropy>`__, aliases: ``binary``
 
--- a/docs/Parameters.rst
+++ b/docs/Parameters.rst
@@ -1010,6 +1010,8 @@ Metric Parameters
 
       -  ``auc``, `AUC <https://en.wikipedia.org/wiki/Receiver_operating_characteristic#Area_under_the_curve>`__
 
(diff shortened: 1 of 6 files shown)


docs/Parameters.rst
```diff
@@ -1010,6 +1010,8 @@ Metric Parameters
 
       -  ``auc``, `AUC <https://en.wikipedia.org/wiki/Receiver_operating_characteristic#Area_under_the_curve>`__
 
+      -  ``ap``, `average precision score <https://scikit-learn.org/stable/modules/generated/sklearn.metrics.average_precision_score.html>`__
```
jameslamb, line 1013: could this be called `average_precision` instead of `ap`? I think the very short names can be confusing unless they're very common acronyms like AUC.

btrotta: Done :)

jameslamb: awesome, thanks! I think it will help

jameslamb (approved):
looks good to me from the R side! I think another reviewer should approve before this is merged though.

jameslamb:
@btrotta in case you aren't aware of the practice...I just added the `feature` label to this so it will be categorized correctly in the next release: https://github.com/microsoft/LightGBM/releases/tag/untagged-bc49d5d1a6b7637e4d22

this is done automatically based on tags by https://github.com/microsoft/LightGBM/blob/master/.github/release-drafter.yml

btrotta:
@jameslamb  Thanks! I didn't realize the tags were used for the release notes.

jameslamb:
no problem!

guolinke (approved):

src/metric/binary_metric.hpp
```diff
@@ -263,5 +263,126 @@ class AUCMetric: public Metric {
...
+  const std::vector<std::string>& GetName() const override {
+    return name_;
+  }
+
+  double factor_to_bigger_better() const override {
+    return 1.0f;
+  }
+
+  void Init(const Metadata& metadata, data_size_t num_data) override {
+    name_.emplace_back("ap");
```
StrikerRUS, line 287: @btrotta 
Was name left `ap` (instead of `average_precision`, https://github.com/microsoft/LightGBM/pull/3347#discussion_r482112568) intentionally?

btrotta: Oops, this was a mistake, fixed now.

PR #3351, opened 2020-09-04
[ci] add an allgood job for easy [required] builds

graingert:
see https://github.community/t/is-it-possible-to-require-all-github-actions-tasks-to-pass-without-enumerating-them/117957/4?u=graingert

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -160,9 +160,10 @@ jobs:
           $env:TASK = "${{ matrix.task }}"
           conda init powershell
           & "$env:GITHUB_WORKSPACE/.ci/test_windows.ps1"
-  allgood:
+  all-successful:
+    # https://github.community/t/is-it-possible-to-require-all-github-actions-tasks-to-pass-without-enumerating-them/117957/4?u=graingert
     runs-on: ubuntu-latest
     needs: [test]
     steps:
-    - name: echo allgood
-      run: echo "allgood"
+    - name: check that all tests succeeded
+      run: echo "🎉"
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -165,5 +165,5 @@ jobs:
     runs-on: ubuntu-latest
     needs: [test]
     steps:
-    - name: check that all tests succeeded
+    - name: note that all tests succeeded
       run: echo "🎉"
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -160,3 +160,9 @@ jobs:
           $env:TASK = "${{ matrix.task }}"
           conda init powershell
           & "$env:GITHUB_WORKSPACE/.ci/test_windows.ps1"
+  allgood:
+    runs-on: ubuntu-latest
+    needs: [test]
+    steps:
+    - name: echo allgood
+      run: echo "allgood"


.github/workflows/main.yml
```diff
@@ -160,3 +160,6 @@ jobs:
           $env:TASK = "${{ matrix.task }}"
           conda init powershell
           & "$env:GITHUB_WORKSPACE/.ci/test_windows.ps1"
+  allgood:
+    runs-on: ubuntu-latest
+    needs: [test]
```
graingert, line 165: looks like `steps` are required

```suggestion
  allgood:
    runs-on: ubuntu-latest
    needs: [test]
    steps:
    - name: echo allgood
      run: echo "allgood"
```

graingert:
@jameslamb you should be able to add "required" to  `GitHub Actions / allgood (pull_request) Successful in 1s `

.github/workflows/main.yml
```diff
@@ -160,3 +160,9 @@ jobs:
...
+    steps:
+    - name: echo allgood
+      run: echo "allgood"
```
StrikerRUS, line 168: Could you please make wording more intelligent please? 😄

StrikerRUS (commented):
@graingert Thanks a lot! Indeed, this looks like something we were looking for!
Now we need just to mark this new job as `required`, right?

image

.github/workflows/main.yml
graingert:
```suggestion
    - name: echo more intelligent please
      run: echo "more intelligent please"
```

jameslamb: sorry, we mean could you make it something like

```yaml
- name: simple-task
   echo: "this is a minimal task"
```

jameslamb:
@StrikerRUS yeah, I think this is right: https://github.com/microsoft/LightGBM/pull/3351#pullrequestreview-482579697

.github/workflows/main.yml
jameslamb: and instead of `allgood`, could you please name this job `all-github-actions-tests`, to make it clearer what it is there for?

and can you add a comment linking to our discussion? https://github.community/t/is-it-possible-to-require-all-github-actions-tasks-to-pass-without-enumerating-them/117957/4?u=graingert

So that in the future, when someone looks at this file, they understand why we have this extra job that seems to only run a small `echo`

graingert, line 168: I don't like `simple-task` as it's not very explanatory:

how about:

```suggestion
  all-successful:
    runs-on: ubuntu-latest
    needs: [test]
    steps:
    - name: "Proclaim that all jobs in the workflow are successful, for easy [requires]"
      run: echo "🎉"
```

jameslamb: echoing an emoji is ok with me! I think this test name is too long. Can you please just call is `check all tests succeeded`, and then add a comment with the discussion link like I recommended in https://github.com/microsoft/LightGBM/pull/3351#discussion_r483687906?

graingert: all-github-actions-tests isn't good because it's already in the "Github Actions" workflow namespace

check isn't quite accurate either, so I think commemorate is more correct:

```suggestion
  all-successful:  # https://github.community/t/is-it-possible-to-require-all-github-actions-tasks-to-pass-without-enumerating-them/117957/4?u=graingert
    runs-on: ubuntu-latest
    needs: [test]
    steps:
    - name: commemorate that all tests succeeded
      run: echo "🎉"
```

jameslamb: We have many maintainers and contributors on this project who are not native English speakers, so I have a strong preference for plain language, avoiding words like `proclaim` or `commemorate`.

Please call the step `check all tests succeeded` and please put the comment on its own line (not inline with `all-successful: `).

.github/workflows/main.yml
```diff
@@ -160,3 +160,10 @@ jobs:
           $env:TASK = "${{ matrix.task }}"
           conda init powershell
           & "$env:GITHUB_WORKSPACE/.ci/test_windows.ps1"
+  all-successful:
+    # https://github.community/t/is-it-possible-to-require-all-github-actions-tasks-to-pass-without-enumerating-them/117957/4?u=graingert
+    runs-on: ubuntu-latest
+    needs: [test]
+    steps:
+    - name: check that all tests succeeded
```
graingert, line 168:
```suggestion
    - name: note that all tests succeeded
```

StrikerRUS (approved):
@graingert Thanks a lot for the workaround!

And thanks for making wordings easier to understand, LGTM!

jameslamb (approved):
works for me, thanks!

@StrikerRUS @guolinke I'll switch the set of required checks after merging

jameslamb:
oh nevermind, it looks like I don't have access to change the required checks.

@guolinke sorry to bother you, could you change the set of required checks for GitHub Actions? Now we only need to require the `all-successful` task and all others can be disabled. After that, we'll be able to add and remove jobs without needing you too change the settings again.

guolinke:
@jameslamb no problem

graingert:
Have a look at https://github.com/probot/settings it allows setting
settings with PRs

jameslamb:
thanks @graingert . We change these settings rarely enough that I think it's better to have them done manually than to have another integration to maintain.

.github/workflows/main.yml
```diff
@@ -160,3 +160,10 @@ jobs:
           $env:TASK = "${{ matrix.task }}"
           conda init powershell
           & "$env:GITHUB_WORKSPACE/.ci/test_windows.ps1"
+  all-successful:
```
webknjaz, line 163: @graingert FYI this check may be problematic with failures and branch protection. Here's a more comprehensive solution: https://github.com/re-actors/alls-green.

graingert: Ah nice one!

danepowell: Could you be more specific? How might this cause problems with branch protection?

webknjaz: @danepowell it's explained at https://github.com/marketplace/actions/alls-green#why but TL;DR — when some job dependencies fail, this job may end up being skipped and branch protection treats skipped jobs as success.

webknjaz: I've also filed https://github.com/microsoft/LightGBM/issues/5501.

PR #3355, opened 2020-09-06
[ci] remove Azure checks from R CI

jameslamb:
Now that the R CI seems to be working ok on GitHub Actions (#3119 ) and the R components have been removed from the image used for Azure (https://github.com/guolinke/lightgbm-ci-docker/pull/11), I think we can simplify the CI code for R by removing `if $AZURE == "true"`branches from the code.

I also think it's possible to remove the dependency `texlive-fonts-extra` on Linux. Using this PR to test if that's true.

### DIFF
--- a/.ci/test_r_package.sh
+++ b/.ci/test_r_package.sh
@@ -12,32 +12,24 @@ export PATH="$R_LIB_PATH/R/bin:$PATH"
 export _R_CHECK_SYSTEM_CLOCK_=0
 
 # Get details needed for installing R components
-#
-# NOTES:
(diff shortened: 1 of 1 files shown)


jameslamb:
Ok all the linux jobs failed, I guess we can't remove `texlive-fonts-extra`. It was worth a shot!

image

image

jameslamb:
@guolinke sorry, can you please remove the check called `allgood`? Only `all-successful` is needed

StrikerRUS (approved):
Great!

PR #3356, opened 2020-09-06
[R-package] remove unnecessary comments

jameslamb:
This PR proposes removing some comments in the R package that I think are unnecessary.

There are a lot of things like this in the R code:

```r
    # return ret
    return(ret)
```

Comments like this don't add any understanding of the code, but they do make it slightly harder to read and possibly lead to false positives when using `git grep` or similar tools to search through the codebase for things.

### DIFF
--- a/R-package/R/callback.R
+++ b/R-package/R/callback.R
@@ -25,7 +25,6 @@ CB_ENV <- R6::R6Class(
 
 cb.reset.parameters <- function(new_params) {
 
-  # Check for parameter list
   if (!identical(class(new_params), "list")) {
     stop(sQuote("new_params"), " must be a list")
   }
@@ -151,7 +150,6 @@ merge.eval.string <- function(env) {
 
   }
 
-  # Return tabulated separated message
   paste0(msg, collapse = "\t")
 
 }
@@ -188,7 +186,6 @@ cb.print.evaluation <- function(period = 1L) {
   attr(callback, "call") <- match.call()
   attr(callback, "name") <- "cb.print.evaluation"
 
-  # Return callback
   callback
 
 }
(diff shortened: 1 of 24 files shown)


StrikerRUS:
@jameslamb 
Could you please fix some typos from #2912 in this PR as well?

I believe it is not "loss" but "metric" or "evaluation function".
https://github.com/microsoft/LightGBM/blob/d4325c5a4428b120d8611b6cba9b821f3a5f3b12/R-package/R/lgb.cv.R#L121


``structture``, ``higher_bettter``
https://github.com/microsoft/LightGBM/blob/d4325c5a4428b120d8611b6cba9b821f3a5f3b12/R-package/R/lgb.cv.R#L595


Same as above for `loss`
https://github.com/microsoft/LightGBM/blob/d4325c5a4428b120d8611b6cba9b821f3a5f3b12/R-package/R/lgb.train.R#L98


Maybe
```
assign(x = ACCUMULATOR_NAME, value = 0.0, envir = .GlobalEnv)
```
?
https://github.com/microsoft/LightGBM/blob/d4325c5a4428b120d8611b6cba9b821f3a5f3b12/R-package/tests/testthat/test_basic.R#L15-L16

jameslamb:
@StrikerRUS sure, I agree with all those! Just pushed them: https://github.com/microsoft/LightGBM/pull/3356/commits/727fe46e45e8af3bf8a3abad7ddbcfc807299bcd

StrikerRUS:
@jameslamb Thank you!

StrikerRUS (commented):
LGTM! I agree that those removed comments don't add any value. But I guess that actual approval should come from anyone who is reading/writing R code.

guolinke (approved):

PR #3359, opened 2020-09-07
[ci] host some artifacts needed by R Windows CI jobs

jameslamb:
Today, the Windows CI jobs sometimes fail downloading artifacts like Rtools and mixtexsetup.exe.

For example, see https://github.com/microsoft/LightGBM/pull/2912#issuecomment-686534527

Following the suggestion in https://github.com/microsoft/LightGBM/pull/2912#issuecomment-686556271, I've hosted the artifacts where we see these failures most often on a very old release (https://github.com/microsoft/LightGBM/releases/tag/v2.0.12). I also updated that release to explain that those artifacts are not a part of LightGBM. I think the "right" way to do this is to create a bucket in an object store like Azure Blob Store, but this is free and easier to maintain.

This PR proposes the necessary changes to download these files from GitHub instead of their current sources. I think that this PR will make our R CI jobs on Windows more reliable and reduce the number of times we have to manually re-run GitHub Actions.

## Notes for Reviewers

* I don't think we should do this for R itself yet. I think it's important that we can change the R version just in `.github/workflows/main.yml`. If we hosted R itself, upgrading to a new R version would also mean manually updating the artifacts on that releases page
* this PR also upgrades us from MiKTeX 2.9 to MiKTex 4.0 (the newest version)

### DIFF
--- a/.ci/test_r_package_windows.ps1
+++ b/.ci/test_r_package_windows.ps1
@@ -86,8 +86,6 @@ if ($env:R_BUILD_TYPE -eq "cmake") {
   }
 }
 
-https://github.com/microsoft/LightGBM/releases/download/v2.0.12/rtools35-x86_64.exe
-
 # download R and RTools
 Write-Output "Downloading R and Rtools"
 Download-File-With-Retries -url "https://cran.r-project.org/bin/windows/base/old/$env:R_WINDOWS_VERSION/R-$env:R_WINDOWS_VERSION-win.exe" -destfile "R-win.exe"
--- a/.ci/test_r_package_windows.ps1
+++ b/.ci/test_r_package_windows.ps1
@@ -12,23 +12,6 @@ function Download-File-With-Retries {
   } while(!$?);
 }
 
-# mirrors that host miktexsetup.zip do so only with explicitly-named
-# files like miktexsetup-2.4.5.zip, so hard-coding a link to an archive as a
-# way to peg to one mirror does not work
-#
-# this function will find the specific version of miktexsetup.zip at a given
-# mirror and download it
-function Download-Miktex-Setup {
-    param(
-        [string]$archive,
-        [string]$destfile
-    )
-    $PageContent = Invoke-WebRequest -Uri $archive -Method Get
-    $SetupExeFile = $PageContent.Links.href | Select-String -Pattern 'miktexsetup-2.*'
-    $FileToDownload = "${archive}/${SetupExeFile}"
-    Download-File-With-Retries $FileToDownload $destfile
-}
-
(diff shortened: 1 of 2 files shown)


StrikerRUS:
@jameslamb Could you please do the same for the SWIG files as well? It fails to download from sourceforge quite often and given that only two members have an access to re-run Azure builds and our nightly artifats are generated there, I think it's quite important to for us. Also, note that we use pinned version of SWIG (`3.0.12`), so changing downloading address will not change anything in the pipeline.
https://github.com/microsoft/LightGBM/blob/afc76d2cb8234f6876ed75d923a7916bfef9a1e5/.ci/test_windows.ps1#L42

jameslamb:
@StrikerRUS sure! just added it

* commit: https://github.com/microsoft/LightGBM/pull/3359/commits/71c5899d02bbef064d720131742b37b022f6e849
* updated release: https://github.com/microsoft/LightGBM/releases/tag/v2.0.12

StrikerRUS (approved):
Awesome! I strongly believe it will help to minimize number of CI failure,

> If we hosted R itself, upgrading to a new R version would also mean manually updating the artifacts on that releases page

What about Rtools 4? Is it updated regularly?

jameslamb:
> What about Rtools 4? Is it updated regularly?

Rtools 4 is updated, but I have no idea what frequency. It is just marked "not frozen" at https://cran.r-project.org/bin/windows/Rtools/history.html

image

For now, I've decided to host it in that release. We should not have to update it nearly as often as we update R versions.

StrikerRUS:
> Rtools 4 is updated, but I have no idea what frequency. 

Looks like not so often. It is still `4.0.0`.

image


https://cran.r-project.org/bin/windows/Rtools/VERSION.txt

jameslamb:
> > Rtools 4 is updated, but I have no idea what frequency.
> 
> Looks like not so often. It is still `4.0.0`.
> 
> image
> 
> https://cran.r-project.org/bin/windows/Rtools/VERSION.txt

you can't tell from that. "not frozen" means that they upload new artifacts and keep calling them 4.0 (overwriting the previous one) :grimacing:

StrikerRUS:
> "not frozen" means that they upload new artifacts and keep calling them 4.0 (overwriting the previous one) 😬

Oh, really?! Bad practice! I thought they at least increase patch part of version (`4.0.x`)...

jameslamb:
> > "not frozen" means that they upload new artifacts and keep calling them 4.0 (overwriting the previous one) grimacing
> 
> Oh, really?! Bad practice! I thought they at least increase patch part of version (`4.0.x`)...

it's terrible practice, I know :grimacing:

Issue #3354, opened 2020-09-05
[R-package] Error installing R package with local pre-complie option in `build_r.R`


esvhd:
Hi,

I'm having problem install lightgbm for R after I complied from source with git master branch. 

My locally complied version works for my python install. But struggling with R. It seems to not be able to find `lib_lightgbm.so`, which I clearly have in `~/git/LightGBM/lib_lightgbm.so`.

I followed the instruction [here](https://github.com/microsoft/LightGBM/tree/master/R-package#install). Specifically, I followed this instruction: 

"If you are using a precompiled dll/lib locally, you can move the dll/lib into LightGBM root folder, modify `LightGBM/R-package/src/install.libs.R`'s 2nd line (change `use_precompile <- FALSE` to `use_precompile <- TRUE`), and install R-package as usual."

**What exact files do I need to copy for Linux?** `~/git/LightGBM/lib_lightgbm.so`? And to where?

I attached the console output after I ran `Rscript build_r.R` below.

THANKS!

## How you are using LightGBM?

* R package
* Python package

## Environment info

Operating System: Ubnutu 20.04

CPU/GPU model: Nvidia GeForece 1080Ti

C++ compiler version: GCC 9.3

CMake version: 3.16.3

R version: 4.0.2

Other:

LightGBM version or commit hash: git master branch

## Error message and / or logs

```
>Rscript build_r.R

 checking for file ‘/home/abc/git/LightGBM/lightgbm_r/DESCRIPTION’ ... OK
* preparing ‘lightgbm’:
* checking DESCRIPTION meta-information ... OK
* cleaning src
* checking for LF line-endings in source and make files and shell scripts
* checking for empty or unneeded directories
WARNING: directory ‘lightgbm/src/compute/test’ is empty
* looking to see if a ‘data/datalist’ file should be added
* building ‘lightgbm_3.0.0.99.tar.gz’

* installing to library ‘/home/abc/R/x86_64-pc-linux-gnu-library/4.0’
* installing *source* package ‘lightgbm’ ...
** using staged installation
** libs
installing via 'install.libs.R' to /home/abc/R/x86_64-pc-linux-gnu-library/4.0/00LOCK-lightgbm/00new/lightgbm

Error in eval(ei, envir) : Cannot find lib_lightgbm.so

* removing ‘/home/abc/R/x86_64-pc-linux-gnu-library/4.0/lightgbm’
Error in .run_shell_command(install_cmd, install_args) : 
[...]

Issue #2714, opened 2020-01-28
[R-package] Installation with pre-compiled DLL fails: 'cannot create dataset handle'


abhmodi:
Hi, 

I tried other methods mentioned in earlier threads/issues about fixing such issues but none seem to work. 

Apart from the below mentioned reproducible example, I had tried with code mentioned on the test code mentioned here: https://github.com/microsoft/LightGBM/tree/master/R-package#install

Running that code doesn't give "[LightGBM] [Fatal] Check failed: num_data > 0 at D:\a\1\s\src\io\dataset.cpp, line 30 ." error but only about "cannot create dataset handle"

## Environment info

Operating System: Windows 10

CPU/GPU model:

C++/Python/R version: R-3.5.2

LightGBM version or commit hash: 2.3.2

## Error message

[LightGBM] [Fatal] Check failed: num_data > 0 at D:\a\1\s\src\io\dataset.cpp, line 30 .

Error in dataset$construct() : 
  lgb.Dataset.construct: cannot create Dataset handle

## Reproducible examples

lgb.unloader(wipe=TRUE)
require(lightgbm)
data(ChickWeight) # Built in data set
 
ChickWeight <- sapply(ChickWeight, function(x) as.numeric(as.character(x)))
 
 y <- ChickWeight[,"weight"]
 x <- ChickWeight[,!colnames(ChickWeight)=="weight"]
 
 dtrain <- lgb.Dataset(data = as.matrix(x),
                                                 label = as.numeric(y),
                                                free_raw_data=FALSE
                          )
lgb.Dataset.construct(dtrain)

## Other methods tried from earlier threads/issues

1. Re-installed LightGBM using RGUI
2. Tried with restarting R-sessions; same issues with another code

jameslamb:
Hello @abhmodi . I tried running this code from your example on the latest version of the `lightgbm` R package:

```r
library(lightgbm)
lgb.unloader(wipe=TRUE)
require(lightgbm)
data(ChickWeight) # Built in data set

ChickWeight <- sapply(ChickWeight, function(x) as.numeric(as.character(x)))

y <- ChickWeight[,"weight"]
x <- ChickWeight[,!colnames(ChickWeight)=="weight"]

dtrain <- lgb.Dataset(
    data = as.matrix(x)
    , label = as.numeric(y)
    , free_raw_data = FALSE
)
lgb.Dataset.construct(dtrain)
```

[...]

PR #3360, opened 2020-09-07
[R-package] Remove ability to install precompiled lib_lightgbm (fixes #3320)

### LINKED ISSUE #3320: R README installation instruction hierarchy
The README in `master/R-package` has a title hierarchy that is a bit confusing to me. Here I summarized the part of it that I find hard to follow:

```
Installation
------------
...

### Installing the CRAN package
...

### Installing Precompiled Binaries

...

### Preparation (<- this is a bit confusing to me. 
                      Should it be hierarchy lower and have a header 'Building and installing from source'  
                      or similar that also contains the '### Install'?)

...

#### Windows Preparation

...

#### Windows Toolchain Options

...

#### Mac OS Preparation

...

### Install

...


Examples
--------

...

```

I am excited about recent efforts to distribute this package via CRAN, so maybe this README is WIP anyways and the issue is redunant. Also, since it's quite a long README and contains some information for maintainers and not end users, maybe it's worth considering to move some of this content to vignettes. In any case, thanks for this package.

jameslamb:
This PR has two changes that I couldn't see a clean way to separate, so I've combined them:

* clarify how to install the R package (fixes #3320 )
* remove the ability to use a precompiled `lib_lightgbm.so` with the R package

I think that the use of a precompiled `lib_lightgbm.so` in the R package has been broken for a while: #1730, #2714 . This was a valuable option before we started linking to R when building the library (added in #2901 ) and you could just download `lib_lightgbm.so` from the releases page. Now, it is really difficult to manually build a library that works well with the R package.

I think we should eliminate documentation and code about using a pre-built `lib_lightgbm`, and focus on making customizations you can't get with the CRAN package easier via #2441 (like I recommended in https://github.com/microsoft/LightGBM/issues/3354#issuecomment-688008534)

**current README**

image

**proposal**

image


R-package/README.md
```diff
@@ -14,63 +19,56 @@
...
-install.packages(
-    pkgs = local_file
-    , type = "binary"
-    , repos = NULL
-)
+remotes::install_url(PKG_URL)
 ```
 
-### Preparation
+### Installing from Source with CMake <a name="install"></a>
```
jameslamb, line 62: I've aliased this to `#install` so that old links still work

lorenzwalthert: Wow that's cool. Did not even think this is possible.

StrikerRUS: Indeed, nice hack! But right now contents item isn't working. I believe, removing a space before `<a name ...` will fix it.

image

R-package/README.md
```diff
@@ -5,15 +5,44 @@
 ### Contents
 
 * [Installation](#installation)
+    - [Installing the CRAN package](#installing-the-cran-package)
```
StrikerRUS, line 8:
```suggestion
    - [Installing the CRAN Package](#installing-the-cran-package)
```

R-package/README.md
```diff
@@ -5,15 +5,44 @@
...
 * [Known Issues](#known-issues)
 
 Installation
 ------------
 
+For the easiest installation, go to ["Installing the CRAN package"](#installing-the-cran-package).
+
+If you experience any issues with that, try ["Installing from Source with CMake"](#install). This can produce a more efficient version of the library on Windows systems with Visual Studio.
+
+To build a GPU-enabled version of the package, follow the steps in ["Installing a GPU-enabled Build"](#installing-a-gpu-enabled-build)
```
StrikerRUS, line 26:
```suggestion
To build a GPU-enabled version of the package, follow the steps in ["Installing a GPU-enabled Build"](#installing-a-gpu-enabled-build).
```

lorenzwalthert:
Looks good. One minor thing: I think the distinction between CPU and GPU packages is very important and hence one could mention that the CRAN package and pre-compiled binaries don't support GPU training (If I got that right). One can figure that out from the header on how to build a GPU accelerated version, but explicit is better than implicit often.

R-package/README.md
```diff
@@ -5,15 +5,44 @@
...
+        objective = "regression"
+        , metric = "l2"
+    )
+    , data = dtrain
+)
+```
+
 ### Installing the CRAN package
 
 As of this writing, `LightGBM`'s R package is not available on CRAN. However, start with `LightGBM` 3.0.0, you can install a released source distribution. This is the same type of package that you'd install from CRAN. It does not require `CMake`, Visual Studio, or anything else outside the CRAN toolchain.
```
StrikerRUS, line 48:
```suggestion
As of this writing, `LightGBM`'s R package is not available on CRAN. However, start with `LightGBM` 3.0.0, you can install a released source distribution. This is the same type of package that you'd install from CRAN. It does not require `CMake`, `Visual Studio`, or anything else outside the CRAN toolchain.
```

StrikerRUS: Cannot comment on this line, but we need dot here:

image

R-package/README.md
```diff
@@ -30,47 +59,31 @@ PKG_URL <- "https://github.com/microsoft/LightGBM/releases/download/v3.0.0/light
...
+CXX11=g++-8
 ```
 
-### Preparation
+### Installing from Source with CMake <a name="install"></a>
 
 You need to install git and [CMake](https://cmake.org/) first.
 
-Note: 32-bit (i386) R/Rtools is currently not supported.
+Note: this method is only supported on 64-bit systems. If you need to run LightGBM on 32-bit Windows (i386), follow the instructions in ["Installing the CRAN Package"](#installing-the-cran-package).
```
StrikerRUS, line 81: Is it still true?..

R-package/README.md
```diff
@@ -5,15 +5,44 @@
...
 * [Testing](#testing)
-* [Preparing a CRAN Package and Installing It](#preparing-a-cran-package-and-installing-it)
+* [Preparing a CRAN Package](#preparing-a-cran-package)
 * [External Repositories](#external-unofficial-repositories)
 * [Known Issues](#known-issues)
 
 Installation
 ------------
 
+For the easiest installation, go to ["Installing the CRAN package"](#installing-the-cran-package).
```
StrikerRUS, line 22: I think we should say here that **the recomended way is CMake + VS** due to performance reasons.

StrikerRUS (commented):
Just some minor comments.

R-package/README.md
jameslamb: yes, we haven't added support for a 32-bit build with CMake and I don't think we should.

R-package/README.md
jameslamb: I disagree with that being "the recommended way". I don't think we should assume that the average user of the R package would prefer efficiency of the library over ease of installation. The CMake path is an order of magnitude more difficult than the CRAN one.

I did note the performance expectations in the next line

> If you experience any issues with that, try ["Installing from Source with CMake"](#install). This can produce a more efficient version of the library on Windows systems with Visual Studio.

R-package/README.md
jameslamb: the spacing for HTML tags is irrelevant, except for very special ones like `<pre>`. The issue here was that I got the anchor wrong in the table of contents: https://github.com/microsoft/LightGBM/pull/3360/commits/df5bf3f6b193b68a9e83dc16b5e31341f142a355

this is working now

image

jameslamb:
> Looks good overall. One minor thing: I think the distinction between CPU and GPU packages is very important and hence one could mention that the CRAN package and pre-compiled binaries don't support GPU training (If I got that right). One can figure that out from the header on how to build a GPU accelerated version, but explicit is better than implicit often.

thanks @lorenzwalthert . I disagree with this. As you mentioned in #3320, the README is already hard to follow and has a lot of information. I want to make this as simple as possible.

I think anyone arriving at this page looking to build for the GPU will be drawn to "Installing a GPU-enabled Build" in the list of options.

image

StrikerRUS (approved):

R-package/README.md
StrikerRUS: cc @guolinke and @Laurae2

jameslamb: I won't merge this until this thread is answered. Gently ping @guolinke

jameslamb:
I just merged `master` into this. This is ready for review and does not need to wait for #3338

R-package/README.md
guolinke: sorry for missing this. For cran and R, I think easy-to-install is preferred by users. So we can add a note about cmake+vs is more efficiency in windows.

guolinke (approved):

R-package/README.md
jameslamb: ok great, thanks! Then I think this is ok as of I've written it, with the note on the the choice of cmake + VS for a more efficient version

jameslamb:
I just merged `master` into this. Since we've agreed on https://github.com/microsoft/LightGBM/pull/3360#discussion_r484522114, I'll merge this one it builds.

PR #3254, opened 2020-07-27
[Python] Refactors scikit-learn API to allow a list of evaluation metrics

gramirezespinoza:
This PR refactors the code in LGBM's scikit-learn API to allow the user to pass a list of evaluation metrics to the parameter `eval_metric` in the `fit` method of `LGBMModel` and its subclasses.

The snippet bellow shows an example of the possible uses of this functionality:

```python
import ...

def custom_recall_sklearn(y_true, y_pred):
    return 'recall', recall_score(y_true, y_pred > 0.5), True


def custom_precision_sklearn(y_true, y_pred):
    return 'precision', precision_score(y_true, y_pred > 0.5), True

parameters = {
    "objective": "binary",
    "metric": "binary_logloss",
}

model = lightgbm.LGBMClassifier(**parameters)

model.fit(
    X=train,
    y=train_label,
    eval_set=(validation, validation_label),
    eval_metric=[custom_recall_sklearn, custom_precision_sklearn, "auc"])
```

Where `eval_metric` receives a list that can combine custom and built-in evaluation functions. The output looks like this:

```python
[1]	valid_0's auc: 0.951963	valid_0's binary_logloss: 0.620274	valid_0's recall: 0.963916	valid_0's precision: 0.932162
[2]	valid_0's auc: 0.972375	valid_0's binary_logloss: 0.558293	valid_0's recall: 0.971234	valid_0's precision: 0.93151
[3]	valid_0's auc: 0.975112	valid_0's binary_logloss: 0.507648	valid_0's recall: 0.967701	valid_0's precision: 0.935594
[4]	valid_0's auc: 0.975005	valid_0's binary_logloss: 0.463648	valid_0's recall: 0.96644	valid_0's precision: 0.938495
[5]	valid_0's auc: 0.975215	valid_0's binary_logloss: 0.426748	valid_0's recall: 0.966944	valid_0's precision: 0.939676
```

The user can monitor during training as many metrics as needed.

@StrikerRUS rightly mentioned in PR [3165](https://github.com/microsoft/LightGBM/pull/3165) and in issue [2182](https://github.com/microsoft/LightGBM/issues/2182) that it is possible to monitor multiple metrics already. This PR leverages that functionality to make it a bit more intuitive.

This PR is backwards compatible.

PR #3367, opened 2020-09-07
[python][examples] updated examples with multiple custom metrics

StrikerRUS:
After merging #3254 it is possible to simplify `feval` argument with multiple custom metrics significantly and remove proxy functions.

cc @gramirezespinoza


jameslamb (approved):
love it!

PR #3368, opened 2020-09-07
[R-package] Fix incorrect num_iterations, early_stopping_round in saved models (fixes #2208, #2468)

### LINKED ISSUE #2208: incorrect logged parameters in model file

jameslamb:
Today, the R wrapper for LightGBM has some keyword arguments which can be used to set key [parameters](https://lightgbm.readthedocs.io/en/latest/Parameters.html#core-parameters):

* `nrounds` --> `num_iterations`
* `early_stopping_rounds` --> `early_stopping_round`

If these arguments are used instead of entries in `params`, the saved model generated by C++ function `LGBM_BoosterSaveModelToString_R` will be incorrect (#2208).

This PR proposes fixes to `lgb.train()` and `lgb.cv()` to ensure that `params` is updated with the correct details so that models are saved correctly.

I was pleasantly surprised to learn that this fix also fixes training continuation in the R package based on a model file! (#2468)

## How this improves `{lightgbm}`

This PR ensures that models stored in a text file (e.g. `lightgbm.model`) are correct, so that `{lightgbm}` behaves as expected when that model is later loaded for scoring or continued training.


guolinke:
I sometimes also confused about `num_boost_round` and 'early_stopping_rounds` in lgb.train. (to consistent with xgboost).
maybe with `num_*` doesn't need the `s` ?

jameslamb:
> I sometimes also confused about `num_boost_round` and `early_stopping_rounds` in lgb.train. (to consistent with xgboost).
> maybe with `num_*` doesn't need the `s` ?

it is confusing but I changing those argument names now would be a breaking change, and I don't think we should do that. With the aliases, `early_stopping_round` and `early_stopping_rounds` are both supported (https://lightgbm.readthedocs.io/en/latest/Parameters.html#early_stopping_round) as are `num_iteration` and `num_iterations` (https://lightgbm.readthedocs.io/en/latest/Parameters.html#num_iterations)

jameslamb:
I just merged `master` into this. I think this is ready for review.

guolinke (approved):

PR #3370, opened 2020-09-08
[docs] remove unrelated entry from winning solutions list

StrikerRUS:
Correct link for the 2nd place solution is https://www.kaggle.com/c/abstraction-and-reasoning-challenge/discussion/154391 and I don't see any mentions of LightGBM there. Also, I checked 1st and 3rd solutions - no usage of LightGBM as well.


jameslamb (approved):
ah you're right, sorry I missed it

PR #3373, opened 2020-09-09
Avoid segment fault in ResetConfig for GBDT in prediction (fix #3317)

### LINKED ISSUE #3317: segfault in LGBM_BoosterResetParameter()
Hello, in 2.3.2 and 3.0.0 the following segfaults in LGBM_BoosterResetParameter():

int out_num_iterations = 0;
BoosterHandle myBooster = 0;
int ret = LGBM_BoosterLoadModelFromString( model_str, &out_num_iterations, &myBooster );
if ( (!ret) && myBooster ) {
  ret = LGBM_BoosterResetParameter( myBooster, "num_threads=1" );
}

The crash occurs in gpdt.cpp
void GBDT::ResetConfig(const Config* config)
if (config_->forcedsplits_filename != new_config->forcedbins_filename) {
config_ is non-null but bogus, hence the segfault.

No crash in 2.2.3.

For 2.3.2/3.0.0, if I don't call LGBM_BoosterResetParameter() the booster works fine for inferencing.

Thank you for the help.

shiyu1994:
This PR is to fix #3317. When a booster is created from model file, its config_ member will be undefined. However, the ResetConfig method of GBDT is not aware of this. As @glarb pointed out.

### DIFF
--- a/src/boosting/gbdt.cpp
+++ b/src/boosting/gbdt.cpp
@@ -20,6 +20,7 @@ namespace LightGBM {
 GBDT::GBDT()
     : iter_(0),
       train_data_(nullptr),
+      config_(nullptr),
       objective_function_(nullptr),
       early_stopping_round_(0),
       es_first_metric_only_(false),
@@ -718,7 +719,7 @@ void GBDT::ResetConfig(const Config* config) {
   if (train_data_ != nullptr) {
     ResetBaggingConfig(new_config.get(), false);
   }
-  if (config_->forcedsplits_filename != new_config->forcedbins_filename) {
+  if (config_.get() != nullptr && config_->forcedsplits_filename != new_config->forcedbins_filename) {
     // load forced_splits file
     if (!new_config->forcedsplits_filename.empty()) {
       std::ifstream forced_splits_file(


guolinke (approved):

src/boosting/gbdt.cpp
```diff
@@ -718,7 +719,7 @@ void GBDT::ResetConfig(const Config* config) {
   if (train_data_ != nullptr) {
     ResetBaggingConfig(new_config.get(), false);
   }
-  if (config_->forcedsplits_filename != new_config->forcedbins_filename) {
+  if (config_.get() != nullptr && config_->forcedsplits_filename != new_config->forcedbins_filename) {
```
btrotta, line 722: @guolinke I think there might be a typo here, should it be `new_config->forcedsplits_filename` (instead of `new_config->forcedbins_filename`)?

guolinke: @btrotta you are right, it is typo.

guolinke: @btrotta can you help to create a PR?

btrotta: @guolinke Done in #3392. 

@shiyu1994 Apologies, I just now realized you are the author of this PR, not guolinke. I should have mentioned you in my original comment.

glarb:
Hi @shiyu1994, I think it would have been better for LGBM_BoosterResetParameter() to just create the config if it doesn't exist.  That way folks can set parameters once instead of for each predict (like we could before).  Please reconsider this fix, thanks.

PR #3377, opened 2020-09-11
[python] fix dangerous default for eval_at in LGBMRanker

jameslamb:
I ran `pylint` over the Python package tonight, and found some tiny warnings that I think are worth addressing.

This PR fixes the following:

```text
python-package/lightgbm/sklearn.py:944:4: W0102: Dangerous default value [] as argument (dangerous-default-value)
```

Using literals of mutable data structures like `[]` or `{}`  is dangerous because Python initializes them and then every call gets a reference to the same object.

Like this

```python
def stuff(x=[]):
     x.append("a")
     return(x)

stuff()
stuff()
stuff()
```

image

## Question for Reviewers

@StrikerRUS what do you think about running `pylint` in CI to catch things like this? I have found that it catches things like this that `pycodestyle` doesn't catch, and which can affect the correctness of the code. It's also a very lightweight dependency that runs quickly, so it shouldn't make the `lint` task on Travis too much slower or fragile.

If you're open to this, I'll open a separate PR to add `pylint`  to the `lint` task on Travis. We could also be very aggressive about ignoring style items to focus only on things that impact the correctness of the code, like this: https://github.com/jameslamb/doppel-cli/blob/84eb4c1973d96f7bbc742ee0b5231e314740ce1d/.ci/lint-py.sh#L54-L56


StrikerRUS:
> I have found that it catches things like this that pycodestyle doesn't catch, and which can affect the correctness of the code.

Yeah, I've noticed that too. `pylint` is quite stricter than `pycodestyle`... I believe we will have to ignore a lot of style warnings or have to rewrite Python wrapper completely! 😄  But I'm OK to have a test PR with `pylint` and see what we will get.

jameslamb:
> I believe we will have to ignore a lot of style warnings or have to rewrite Python wrapper completely

yeah I agree! Honestly even if we ignore 85% of the warnings that are style-only, it can still be worth it for some of these things that affect the correctness of the code. I'll try a PR this weekend, let's see how bad it looks

jameslamb:
but to be clear it will be separate from this PR

python-package/lightgbm/sklearn.py
```diff
@@ -945,11 +945,12 @@ def fit(self, X, y,
             sample_weight=None, init_score=None, group=None,
             eval_set=None, eval_names=None, eval_sample_weight=None,
             eval_init_score=None, eval_group=None, eval_metric=None,
-            eval_at=[1, 2, 3, 4, 5], early_stopping_rounds=None, verbose=True,
+            eval_at=None, early_stopping_rounds=None, verbose=True,
             feature_name='auto', categorical_feature='auto',
             callbacks=None, init_model=None):
         """Docstring is inherited from the LGBMModel."""
         # check group data
+        eval_at = eval_at or [1, 2, 3, 4, 5]
```
StrikerRUS, line 953: scikit-learn doesn't like when we modify `fit` arguments: https://github.com/microsoft/LightGBM/pull/3254#discussion_r483997893.

I believe changing default value from `list` to `tuple` (`[1, 2, 3, 4, 5] -> (1, 2, 3, 4, 5)`) will be enough and will not break anything.

StrikerRUS (changes requested):

python-package/lightgbm/sklearn.py
jameslamb: in this case, the behavior is identical. Just the way we achieve it is different. I followed that comment and then the PR it linked to and I'm not sure that the change I'm proposing would actually break anything.

I'm nervous that changing data types will break something, because now you have the default as a tuple and users probably passing in values as a list

StrikerRUS: > I'm nervous that changing data types will break something, because now you have the default as a tuple and users probably passing in values as a list

That `pylint` warning you are addressing in this PR is about dangerous **default** values. There should be nothing wrong with that users pass lists intentionally. 

I don't really like you approach because with it default value is implicit: documentation says it is `1,2,3,4,5` but actually it is `None`. It may be misleading. 
Also, scikit-learn prefers everything to be explicit. It is not allowed to change `init` args or even have `kwargs`, for example. So I'm pretty sure it is better to not have implicit argument overwriting for avoiding any possible (future) issues.

jameslamb: ok. To be honest I still don't understand the scikit-learn thing but I believe you.

>  There should be nothing wrong with that users pass lists intentionally.

Using a tuple as the argument means that if we have any code anywhere down the stack that is expecting specifically a list and not a tuple (for example, something that does `eval_at.append()`), it will break. But let me try and see what the tests say.

jameslamb: sorry I forgot about this one! Just updated to a tuple. Hopefully the tests will still pass. I searched around with `git grep eval_at` and didn't find any list-specific code so I think it should work.

python-package/lightgbm/sklearn.py
```diff
@@ -984,6 +984,6 @@ def fit(self, X, y,
     _base_doc = fit.__doc__
     _before_early_stop, _early_stop, _after_early_stop = _base_doc.partition('early_stopping_rounds :')
     fit.__doc__ = (_before_early_stop
-                   + 'eval_at : list of int, optional (default=[1, 2, 3, 4, 5])\n'
+                   + 'eval_at : iterable of int, optional (default=[1, 2, 3, 4, 5])\n'
```
StrikerRUS, line 987: Or we can silently change default from list to tuple and continue asking users to provide only lists.
```suggestion
                   + 'eval_at : iterable of int, optional (default=(1, 2, 3, 4, 5))\n'
```

StrikerRUS (approved):
LGTM! Thank you!

Issue #3372, opened 2020-09-09
instability caused by floating point errors


rebidaldal:
## How you are using LightGBM?

LightGBM component:  Python package

## Environment info

Operating System:  Ubuntu 20.04.1 LTS

CPU/GPU model: Intel(R) Xeon(R) Platinum 8259CL CPU @ 2.50GHz

C++ compiler version: gcc version 9.3.0 (Ubuntu 9.3.0-10ubuntu2)

CMake version: 3.16.3

Java version: N/A

Python version:  3.8.5

R version: N/A

Other:

LightGBM version or commit hash: 3.0.0 (also happens in 3.0.0.99 afc76d2cb8234f6876ed75d923a7916bfef9a1e5)

## Error message and / or logs

Given the same data/parameters/seeds lightgbm sometimes produces different models/outputs. This makes experiment reproducibility problematic. I believe this is the result of some kind of floating point operation error propagation. Models initially start the same but when the pickled model files are inspected they start to differ after >250 iterations. I have noticed this error before in lightgbm 2.3.1 but worked around it by disabling bagging.  #2598 In this version I get the same error even without using bagging. In the notebook inside the linked repo I have tried running the models 5 times and got different results every time. Results also change when the notebook is restarted. 

## Reproducible example(s)

Data and code required to reproduce this bug:  [https://github.com/rebidaldal/lightgbmBugreport](https://github.com/rebidaldal/lightgbmBugreport)
```
X_train = pd.read_feather("X_train.feather")
y_train = pd.read_feather("y_train.feather")
X_test = pd.read_feather("X_test.feather")
```

```
param = {}
param['boosting_type']      = 'gbdt'
param['feature_fraction']   = 0.2
param['feature_fraction_bynode']   = 0.5
param['lambda_l2']          = 1
param['learning_rate']      = 0.02
param['max_bin']          = 31
param['max_delta_step']          = 63
param['max_depth']          = 20
param['metric']             = 'rmse'
param['min_data_in_bin']       = 8191
param['min_data_in_leaf']   = 8191
param['min_gain_to_split']       = 1
param['num_leaves']         = 100
param['objective']          = 'regression'
param['verbosity']          = -1
param['seed'] = 1
```

```
dTrain = lgb.Dataset(X_train, label=y_train) 
model = lgb.train( param, dTrain, 1000)
preds = model.predict(X_test)
preds
```

array([15.032409  , 15.06296509, 14.98345116, ..., 15.24531394,
[...]

PR #2717, opened 2020-01-29
Refactoring monotone constraints (linked to #2305)

CharlesAuguste:
@aldanor @redditur

@guolinke @chivee @jameslamb  @huanzhang12 @Laurae2 @StrikerRUS 
You were all reviewers of the PR #2305. This PR was judged to be not merge-able because it is too big. Therefore, in my ultimate comment (https://github.com/microsoft/LightGBM/pull/2305#issuecomment-574560469) I said I would split into smaller PR easier to merge. This is the first PR in relation with #2305.

The goal of this PR is just to refactor the current code around monotone constraints. **The behavior of the library should not be affected by this PR**. Refactoring the monotone constraints that way was asked by @guolinke in the previous PR (from the comment https://github.com/microsoft/LightGBM/pull/2305#issuecomment-528675809). We think that refactoring is very important because it will make it easier to develop more efficient constraining methods. 

The main issue with this PR is that I was not able to refactor the files data_parallel_tree_learner.cpp, gpu_tree_learner.cpp, voting_parallel_tree_learner because I am not familiar with them (already mentioned in https://github.com/microsoft/LightGBM/pull/2305#issue-303471370 and in the discussion below). I can spend more time on it, but I would really appreciate getting help on that specific part if that is possible for you. @guolinke Since this refactor is something you were looking forward to, are you willing to fix these files for us? If you are familiar with these tree learners, I hope that would not be too much work. Please let me know what you think. Thank you very much.

PR #3384, opened 2020-09-12
improve subfeature_bynode

guolinke:
this is useful to filter the histogram that cannot split.

### DIFF
--- a/src/treelearner/serial_tree_learner.cpp
+++ b/src/treelearner/serial_tree_learner.cpp
@@ -727,6 +727,8 @@ void SerialTreeLearner::ComputeBestSplitForFeature(
         leaf_splits->leaf_index(), config_->monotone_penalty);
     new_split.gain *= penalty;
   }
+  // it is needed to filter the features after above code. 
+  // Otherwise, the `is_splittable` in `FeatureHistogram` will be wrong, and cause some features being accidentally filtered in the later nodes.
   if (new_split > *best_split && is_feature_used) {
     *best_split = new_split;
   }
--- a/src/treelearner/serial_tree_learner.cpp
+++ b/src/treelearner/serial_tree_learner.cpp
@@ -702,9 +702,6 @@ void SerialTreeLearner::ComputeBestSplitForFeature(
     FeatureHistogram* histogram_array_, int feature_index, int real_fidx,
     bool is_feature_used, int num_data, const LeafSplits* leaf_splits,
     SplitInfo* best_split) {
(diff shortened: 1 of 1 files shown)


src/treelearner/serial_tree_learner.cpp
```diff
@@ -702,9 +702,6 @@ void SerialTreeLearner::ComputeBestSplitForFeature(
     FeatureHistogram* histogram_array_, int feature_index, int real_fidx,
     bool is_feature_used, int num_data, const LeafSplits* leaf_splits,
     SplitInfo* best_split) {
-  if (!is_feature_used) {
-    return;
-  }
   SplitInfo new_split;
   double parent_output;
   if (leaf_splits->leaf_index() == 0) {
```
guolinke, line 707: @btrotta it seems the parent output is calculated repeatedly here for root node. I think this cost could be reduced.

guolinke:
This seems can partial fix #3372 .
Just track the commit, #2717 is the root to break the feature_fraction_by_node.

src/treelearner/serial_tree_learner.cpp
```diff
@@ -730,7 +727,7 @@ void SerialTreeLearner::ComputeBestSplitForFeature(
         leaf_splits->leaf_index(), config_->monotone_penalty);
     new_split.gain *= penalty;
   }
-  if (new_split > *best_split) {
+  if (new_split > *best_split && is_feature_used) {
```
guolinke, line 730:
```suggestion
  // it is needed to filter the features after above code. 
  // Otherwise, the `is_splittable` in `FeatureHistogram` will be wrong, and cause some features being accidently filtered in the later nodes.
  if (new_split > *best_split && is_feature_used) {
```

src/treelearner/serial_tree_learner.cpp
```diff
@@ -730,7 +727,9 @@ void SerialTreeLearner::ComputeBestSplitForFeature(
         leaf_splits->leaf_index(), config_->monotone_penalty);
     new_split.gain *= penalty;
   }
-  if (new_split > *best_split) {
+  // it is needed to filter the features after above code.
```
guolinke, line 730:
```suggestion
  // it is needed to filter the features after the above code.
```

src/treelearner/serial_tree_learner.cpp
btrotta: Ok, I'll take a look

PR #3385, opened 2020-09-12
stable multi-threading sum reduction

guolinke:
omp reduction seems unstable when #threading is large.
This PR implements a customed multi-threading reduction, it may be slightly slower than omp version, but it will be stable and produce the consistent results.

This PR should be merge after #3384 .


src/treelearner/leaf_splits.hpp
```diff
@@ -88,16 +90,17 @@ class LeafSplits {
...
-      tmp_sum_gradients += gradients[idx];
-      tmp_sum_hessians += hessians[idx];
-    }
-    sum_gradients_ = tmp_sum_gradients;
-    sum_hessians_ = tmp_sum_hessians;
+    Threading::SumReduction<data_size_t, double, double>(
+        0, num_data_in_leaf_, 2048,
+        [=](int, data_size_t start, data_size_t end, double* s1, double* s2) {
+          for (data_size_t i = start; i < end; ++i) {
+            *s1 = *s2 = 0;
```
guolinke, line 97:
```suggestion
          *s1 = *s2 = 0;
          for (data_size_t i = start; i < end; ++i) {
```

include/LightGBM/utils/threading.h
```diff
@@ -18,13 +18,15 @@ namespace LightGBM {
 
 class Threading {
  public:
+
```
guolinke, line 21:
```suggestion
```

src/boosting/gbdt.cpp
```diff
@@ -229,6 +229,10 @@ void GBDT::Bagging(int iter) {
         bag_data_indices_.data());
     bag_data_cnt_ = left_cnt;
     Log::Debug("Re-bagging, using %d data to train", bag_data_cnt_);
+    // check bagging results, for debug. should be removed before merge.
+    for (data_size_t i = 1; i < bag_data_cnt_; ++i) {
+      CHECK_GT(bag_data_indices_[i], bag_data_indices_[i - 1]);
+    }
```
guolinke, line 235:
```suggestion
```

Issue #3390, opened 2020-09-13
[R-package] use keyword arguments for all internal function calls


jameslamb:
## Summary

Today in the R package, there are a lot of internal function calls which use only positional arguments. Change them to use keyword arguments for extra safety.

I've added this issue to provide a small, focused contribution opportunity for [Hacktoberfest 2020](https://hacktoberfest.digitalocean.com/) participants. If you are an experienced open source contributor, please leave this for beginners and consider contributing [on a different backlog item](https://github.com/microsoft/LightGBM/issues).

## Motivation

Using positional arguments in internal code is dangerous, because it can allow mistakes to slip through or can lead to confusing errors.

This is especially dangerous for cases where the values you'd pass to two parameters take on similar values. For example, in `lightgbm()`, `eval_freq` and `early_stopping_rounds` are both positive integers and can take on similar values. You might not known, unless you're very careful, that you accidentally specified them in the wrong order.

## How to Contribute

To help with this issue, propose a pull request which replaces uses of positional arguments in the R package with keyword arguments.

That means changing calls like this:

```r
# before
folds <- generate.cv.folds(
    nfold
    , nrow(data)
    , stratified
    , getinfo(data, "label")
    , getinfo(data, "group")
    , params
)

# after
folds <- generate.cv.folds(
    nfold = nfold
    , nrows = nrow(data)
    , stratified = stratified
    , label = getinfo(data, "label")
    , group = getinfo(data, "group")
    , params = params
)
```

1. Only change code in `R-package/R/`
1. Before submitting your PR, re-generate the documentation. Be sure you have `{roxygen2}` 7.1.1 installed. If you have any issues with this, just ask when you submit your PR and I can regenerate the docs for you
    ```shell
    sh build-cran-package.sh
    R CMD INSTALL --with-keep.source lightgbm_3.*.tar.gz
    cd R-package
    Rscript -e "roxygen2::roxygenize(load = 'installed')"
    ```
1. Do not modify any calls to `lgb.call()`
1. Do not worry about any function calls from the standard library (like `print()`,  `length()`, `is.character()`, etc.)
1. If adding these arguments makes lines too long, use comma-first style with one argument per line
    ```r
    result <- some_function(
        arg1 = TRUE
        , arg2 = 4
        , arg3 = "hello"
        , ...
    )
    ```

Refer to #3391 as an example. If you have any questions, tag me on this issue and I can help.

## Assignments

[...]

PR #3391, opened 2020-09-13
[R-package] use keyword arguments in internal function calls

jameslamb:
I've just opened #3390. I think it's a great thing for Hacktoberfest contributors to work on.

Today the R package has a lot of internal function calls that use positional arguments instead of keyword arguments. This can be dangerous and lead to silent mistakes if you mix up values parameters that can take similar input ranges, for example `nrounds` and `early_stopping_rounds`.

I'm opening this PR to fix the cases in `lightgbm.R` and to serve as an example for #3390 .


jameslamb:
I just rebased this to `master`. This is ready for review and does not need to wait for #3338

guolinke (approved):

PR #3393, opened 2020-09-13
[ci] Use Ubuntu Focal image at Travis

StrikerRUS:
Refer to https://blog.travis-ci.com/2020-08-10-focal-build-environment.
Our policy is to test LightGBM against the most recent versions of OSes at Travis CI (in the opposite to Azure where we try to use the oldest images for producing public artifacts).

### DIFF
--- a/.ci/test.sh
+++ b/.ci/test.sh
@@ -16,7 +16,7 @@ cd $BUILD_DIRECTORY
 if [[ $TRAVIS == "true" ]] && [[ $TASK == "check-docs" ]]; then
     cd $BUILD_DIRECTORY/docs
     conda install -q -y -n $CONDA_ENV -c conda-forge doxygen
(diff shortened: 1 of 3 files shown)


StrikerRUS:
Sorry for the mess in commits. I found that combination Ubuntu Focal + gcc + Python 3.5 results in failing GPU test (segfault).
But good news is that today Python 3.5 is reaching its EOL: https://devguide.python.org/#branchstatus 😄 
So I'll just prepare a separate PR in which we'll drop the support of it (will not run CI tests with it).
@guolinke @henry0312

StrikerRUS:
`WIP` because `linkchecker` is not available for Focal.

StrikerRUS:
Install `linkchecker` via `pip` from GitHub repo directly.

jameslamb (approved):
I like it!

PR #3395, opened 2020-09-13
[python] Drop Python 3.5 support

StrikerRUS:
Python 3.5 is reaching its End of Life today and no need to support it anymore.

### DIFF
--- a/.vsts-ci.yml
+++ b/.vsts-ci.yml
@@ -30,7 +30,6 @@ jobs:
         TASK: regular
       sdist:
         TASK: sdist
-        PYTHON_VERSION: 3.5
       bdist:
         TASK: bdist
         PYTHON_VERSION: 3.7
@@ -81,7 +80,6 @@ jobs:
         PYTHON_VERSION: 3.7
       sdist:
         TASK: sdist
-        PYTHON_VERSION: 3.5
       bdist:
         TASK: bdist
   steps:
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -29,7 +29,6 @@ install:
   - ps: >-
       switch ($env:PYTHON_VERSION) {
           "2.7" {$env:MINICONDA = "C:\Miniconda-x64"}
-          "3.5" {$env:MINICONDA = "C:\Miniconda35-x64"}
           "3.6" {$env:MINICONDA = "C:\Miniconda36-x64"}
           "3.7" {$env:MINICONDA = "C:\Miniconda37-x64"}
           default {$env:MINICONDA = "C:\Miniconda37-x64"}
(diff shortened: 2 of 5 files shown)


.vsts-ci.yml
```diff
@@ -81,7 +80,6 @@ jobs:
         PYTHON_VERSION: 3.7
       sdist:
         TASK: sdist
-        PYTHON_VERSION: 3.5
```
jameslamb, line 84: so what version of Python will we get in these jobs now? Will it just float?

jameslamb (commented):
Can you please let me know if I'm correct about this:

By "not supporting" Python <3.6.0, we just mean "we don't test against this or guarantee it will work", but we are intentionally not adding  `python_requires=">=3.6"`to `setup_user.py`  because people can still to install the library on those versions.

guolinke:
how about 2.7?

StrikerRUS:
@jameslamb 

> Can you please let me know if I'm correct about this:

Yeah, you're absolutely right! We just stop running tests against Python 3.5. But we do not adding `python_requires` field in `setup.py` to let users use any version they want on their own risk, while "supported" (tested) versions are listed here
https://github.com/microsoft/LightGBM/blob/164818b029b02253796c9b2fe6391c572608f345/python-package/setup.py#L302-L308

@guolinke 
> how about 2.7?

TBH, I don't have any strong opinion on this anymore. [PyPI Stats](https://pypistats.org/packages/lightgbm) shows that Python 2 is still used by 1.5k+ users each day. Do you have plans to add or refactor Python code which will require only 3 version?

.vsts-ci.yml
StrikerRUS: We have "default" values in each CI config which are applied if this field is empty for particular job:
https://github.com/microsoft/LightGBM/blob/164818b029b02253796c9b2fe6391c572608f345/.vsts-ci.yml#L10-L11
https://github.com/microsoft/LightGBM/blob/164818b029b02253796c9b2fe6391c572608f345/.travis.yml#L15-L17
https://github.com/microsoft/LightGBM/blob/164818b029b02253796c9b2fe6391c572608f345/.appveyor.yml#L5-L6

jameslamb (approved):
ok thanks for the explanation. I support this!

.vsts-ci.yml
jameslamb: ooooo I missed that, ok

guolinke:
@StrikerRUS okay, as python 2.7 only increases the CI cost, I think we can continue to support it.

StrikerRUS:
@guolinke 
> okay, as python 2.7 only increases the CI cost, I think we can continue to support it.

Agree. Also, I think we can drop Python 2.7 in two stages:
- stop running tests against it and remove from list of supported Python versions;
- actually drop the support by removing the following code 
   https://github.com/microsoft/LightGBM/blob/571cad7e5512ab6dafc410ac490a0e34f0a98113/python-package/lightgbm/compat.py#L12-L48

PR #3396, opened 2020-09-14
Improve performance of path smoothing

btrotta:
When applying path smoothing, we need to calculate the output of the parent node. Previously this was calculated repeatedly for each potential split feature, but it's more efficient to just do it once per node. See comment here: https://github.com/microsoft/LightGBM/pull/3384/files/6d03633cdf41f7fd68118cdee7f169cac8b88eb0#r487355787

I also fixed a small bug: I was using `leaf_split->leaf_index() == 0` to test whether we are splitting the root node, but this is incorrect, now using `tree->num_leaves() == 1`.


guolinke (approved):
much clearer now, thanks so much!

StrikerRUS:
@btrotta Could you please resolve conflicts?

PR #3397, opened 2020-09-14
[R-package] mark CMake option for building R package as private

StrikerRUS:
Refer to https://github.com/microsoft/LightGBM/pull/3144#discussion_r487533033.


jameslamb (approved):
ok with me!

PR #3403, opened 2020-09-22
[python] Windows wheels that support GPUs via OpenCL

itamarst:
This is a follow-up to #3144, to actually use the new infrastructure to build Python wheels that can work both with and without GPUs.


.ci/test_windows.ps1
```diff
@@ -48,8 +48,23 @@ elseif ($env:TASK -eq "sdist") {
...
+    Write-Output "OpenCL install failed, aborting"
+    exit 1
+  }
+  RefreshEnv
+  Write-Output "Current OpenCL drivers:"
+  Get-ItemProperty -Path Registry::HKEY_LOCAL_MACHINE\SOFTWARE\Khronos\OpenCL\Vendors
+
   cd $env:BUILD_SOURCESDIRECTORY/python-package
-  python setup.py bdist_wheel --plat-name=win-amd64 --universal ; Check-Output $?
+  python setup.py bdist_wheel --integrated-opencl --plat-name=win-amd64 --universal ; Check-Output $?
```
itamarst, line 67: Possibly want to delete the `--universal`?

itamarst:
Will investigate failures tomorrow.

itamarst:
@StrikerRUS do you think you could review this? Thank you!

StrikerRUS:
@itamarst Yeah, I'll review this PR. Thanks for the ping!

But the first thing to do is to decide whether we want to replace our current Windows wheel with the new one which has GPU support. Or we want to create a separate project on PyPI with let say `lightgbm-gpu` name.


> I guess after merging this PR we can simply replace Windows wheel with this new one instead of creating a separate PyPI package for GPU support, right? @tpboudreau @itamarst @huanzhang12 @guolinke @jameslamb
As I've already mentioned, library size increased from ~1.8Mb to ~2.5Mb. So I believe this is not a problem at all.
  https://github.com/microsoft/LightGBM/pull/3144#issuecomment-691572257

jameslamb:
I think it is ok to have a single wheel with support for both.

itamarst:
Having one wheel is nicer user experience, I think:

1. Users don't have to figure out a different way to install something, they install one thing and it Just Works.

2. Use case from my client that paid for this work: they set up identical Python environments across machines with different hardware requirements. Having to customize per-environment is extra work.

StrikerRUS:
I also agree with that one wheel is better than two. But I'd like to hear more opinions from maintainers for this quite important change. @huanzhang12 @guolinke @henry0312 

@itamarst BTW, haven't you done any speed/accuracy benchmarks for CPU-only/old-fashioned locally compiled GPU version vs new wheel?

itamarst:
I haven't done any benchmarks.

huanzhang12:
Keeping a single wheel is a great idea! I really appreciate the hard work by @itamarst @tpboudreau to unify the build process with and without OpenCL. I don't have a Windows machine to test this PR, but the diff looks good to me.

BTW shall we also do the same on Linux wheels?

tpboudreau:
> BTW shall we also do the same on Linux wheels?

The required upstream change to the OpenCL repo just merged.  So I can now proceed on getting the integrated OpenCL build option supported on Linux too.  I'll try to get a PR up soon.

huanzhang12:
> > BTW shall we also do the same on Linux wheels?
> 
> The required upstream change to the OpenCL repo just merged. So I can now proceed on getting the integrated OpenCL build option supported on Linux too. I'll try to get a PR up soon.

That is great!

.ci/test_windows.ps1
```diff
@@ -69,6 +84,11 @@ if (($env:TASK -eq "sdist") -or (($env:APPVEYOR -eq "true") -and ($env:TASK -eq
 }
 pytest $tests ; Check-Output $?
 
+if ($env:TASK -eq "bdist") {
+  # Make sure we can do both CPU and GPU; see tests/python_package_test/test_dual.py
+  $env:LIGHTGBM_TEST_DUAL_CPU_GPU = "1"
+  pytest $env:BUILD_SOURCESDIRECTORY\tests\python_package_test\test_dual.py ; Check-Output $?
+}
```
StrikerRUS, line 91: As we are going to produce one universal wheel file for both GPU and CPU on Windows and given growing popularity of GPU usage among our users, I believe we can make "gpu" mode main for this wheel (only at CI side, not for produced artifacts) and run all our tests for it. Please refer to the following line to how we do that for `TASK="gpu"` on Linux
https://github.com/microsoft/LightGBM/blob/0c1c36cdbdf771a03c9ae0873eb28585e6c3d624/.ci/test.sh#L122-L123

tests/python_package_test/test_dual.py
```diff
@@ -0,0 +1,46 @@
...
+
+    try:
+        param = {"num_leaves": 31, "objective": "binary", "device": "gpu"}
+        lgb.train(param, train_data, 10, valid_sets=[validation_data])
+    except LightGBMError as e:
+        if str(e) == "No OpenCL device found":
+            # This is fine, it means there's no OpenCL device available,
+            # and OpenCL device is only searched for if we successfully
+            # loaded OpenCL GPU backend.
+            pass
```
StrikerRUS, line 44: Could you elaborate more on this? Why do we need this?

itamarst: This is run when the wheel is compiled for both CPU and GPU. It is explicitly requesting GPU usage.

There are two possibilities:

1. An OpenCL driver is available.
2. An OpenCL driver is not available.

In the first case, no exception is raised, test passes: we asked for GPU, got GPU, everything is good.

In the second case, a LightGBMError is raised... but because it mention "No OpenCL device found", that tells us that it tried to find a "GPU" (really, a OpenCL implementation). We asked for OpenCL, and tried to find it, so test should pass too.

StrikerRUS: Yeah, thanks, I understand the general meaning of this case. I was wondering about our concrete CI services. Where is this workaround needed: at Appveyor or Azure Pipelines? Looks like in both cases `opencl_runtime_18.1_x64_setup.msi` will be installed and integrated GPU will be used, won't it?

itamarst: It's needed anytime that test is run and OpenCL is not installed. It's possible that is nowhere, at the moment?

StrikerRUS: Yeah, I think we can remove it because we have the same `if` guards (`if ($env:TASK -eq "bdist")`) for installing OpenCL and running dual tests. Also, `pytest`'s `skipif` with env variable protects ordinary users from running that test.

Moreover, we do not need the second `if` and explicit command to run one new test file because that test file will be included in `$tests` variable. 
https://github.com/microsoft/LightGBM/blob/332559c0f84c2a23ec01c07737814b0aeb0ce474/.ci/test_windows.ps1#L64-L70
The only thing we need is to set `$env:LIGHTGBM_TEST_DUAL_CPU_GPU = "1"` before running tests.

.ci/test_windows.ps1
tpboudreau: First, I want to apologize for taking so long to get back on this.  The delay is entirely my fault.

Is it possible that we can defer this suggestion of building the library with device defaulted to 'gpu' (for CI) until a later time?

I have altered the config.h source code before building the wheel as suggested and have run into several difficulties in testing.  Although I haven't completely uncovered the root causes of the all these failures, I think three things are intersecting: (1) using a CPU based opencl platform (many test assertions fail because results exceed the test tolerance); (2) exposing a few possible bugs in the Windows build for the library (a few tests fail with access violations on Windows even when using an nvidia opencl platform); and (3) installing an opencl driver in a running CI vm and updating the environment visible to the tests (for example, tests occasionally fail because the newly installed driver is not found).

It will take more time than I can spare right this moment to sort thru all these issues (number (2) above is probably the most important to the general user population).

Please let me know if you think we can allow this PR to merge while building and testing the traditional library into the wheel.  Thanks!

tests/python_package_test/test_dual.py
tpboudreau: I've made this change.  test_dual.py now runs only once, with the environment variable set to "1".

.ci/test_windows.ps1
StrikerRUS: @tpboudreau Thanks a lot for your response! No problem, take your time!

> Please let me know if you think we can allow this PR to merge while building and testing the traditional library into the wheel.

I think we can then merge only changes in `setup.py` with the aim for future continuation. All other changes look like a kind of confusing and unused code which brings no value right now. I mean, if we have `if str(e) == "No OpenCL device found": pass`, `test_dual.py` actually asserts that `test_cpu_works()` which only enlarges our CI time. But we cannot produce any artifacts without any tests. In other words, I don't think that adding a test with "skip always" flag makes any sense, but just confuses maintainers.

tests/python_package_test/test_dual.py
StrikerRUS: Thank you very much for all your work!

.ci/test_windows.ps1
tpboudreau: Thanks @StrikerRUS  for continuing to look over and help out with this PR.  I'm not sure I completely follow your comment about test_dual.py -- it is currently active in the Windows bdist test build.  But I've made a change that I think addresses your concerns and makes more sense than the current code.

The primary thing we're testing here is that the python package works, with the new integrated-opencl build, whether or not there is an OpenCL platform present at runtime.  So I've changed the test_windows.ps1 script to run test_dual.py twice (for Windows, bdist builds only): once before OpenCL is installed in the CI VM, and once after it is installed.  The environment variable LIGHTGBM_TEST_DUAL_CPU_GPU is first set to "1" to indicate that there is no OpenCL, and then reset to "2" to indicate that OpenCL is present.  The expected behavior is tested for in test_dual.py (see the new comments in that file for details).  For all other builds, LIGHTGBM_TEST_DUAL_CPU_GPU is not set at all and both tests in test_dual.py are skipped.

(For some reason, right now the final test is failing because the newly installed OpenCL is not being recognized/found -- this is a problem that has occurred sporadically before, and I expect is solvable, so I'm debugging that offline.  I just wanted you to get a look at this proposal to see if it makes more sense to you.)

StrikerRUS: > For some reason, right now the final test is failing because the newly installed OpenCL is not being recognized/found -- this is a problem that has occurred sporadically before, and I expect is solvable, so I'm debugging that offline.

After this problem is solved, I'm totally OK with previous code (before cbfec02659816f0cb321e1184994a9c2f37d2a77) but with removed `if str(e) == "No OpenCL device found": pass`. I think only such kind of test will make us confident that new library works in GPU mode. CPU mode for new library is covered by all other tests we have.

tpboudreau: Hi, @StrikerRUS.  I've reverted the test_dual.py tests as you requested and worked through the issues that caused the build to fail sporadically.

The OpenCL platform installer file used for testing is quite large, so I split it into several chunks to avoid timeout failures during download.  Right now, the installer files are being hosted on a server I'm running (gamma-rho.com) -- I'm willing to keep that server up for the time being, but we should find a more appropriate place for these resources (LightGBM's github?) as soon as possible.

I also inserted a pipeline cache into the build to avoid downloading the OpenCL installer program completely if possible.  On a bad run, downloading the installer can take more than 15 minutes.  The cache reduces it to a few seconds.  It should also make the build more reliable.

StrikerRUS: @tpboudreau Thanks a lot for the updates!

> but we should find a more appropriate place for these resources (LightGBM's github?) as soon as possible.

Uploading to LightGBM's GitHub Releases page right now. We already host Linux AMD OpenCL library there (I guess you've noticed that): https://github.com/microsoft/LightGBM/blob/b69364e9280888914ea716f6e310add457b88d74/.ci/setup.sh#L31 
From our experience this is quite stable place to store files: I personally can't remember any cases we encountered downloading issues with `AMD-APP-SDKInstaller` (187 Mb) from there. WDYT, maybe it will be better to merge installer's parts together back in one file? I believe it should speed up CI a little bit and simplify it. GitHub allows to host up to 2Gb per file.
> Each file included in a release must be under 2 GB. There is no limit on the total size of a release, nor bandwidth usage.
  https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/about-releases#storage-and-bandwidth-quotas

> On a bad run, downloading the installer can take more than 15 minutes.

Thanks for sharing this info! Maybe you've noticed installation time as well?


From you previous message,
> Although I haven't completely uncovered the root causes of the all these failures,

could you please estimate how far we are from changing default device type to `gpu` in Windows wheels?

> I also inserted a pipeline cache into the build to avoid downloading the OpenCL installer program completely if possible.

Do you cache only installer or full installation?
If former, I don't think that it will be useful after uploading the installer to GitHub servers. The whole `setup.py` script with downloading `AMD-APP-SDKInstaller` (187 Mb) takes about a 1 minute.

Could you please elaborate about this line?
> $module = "$env:ChocolateyInstall\helpers\chocolateyProfile.psm1"

StrikerRUS: > Uploading to LightGBM's GitHub Releases page right now.

Done!
https://github.com/microsoft/LightGBM/releases/tag/v2.0.12

Example link: https://github.com/microsoft/LightGBM/releases/download/v2.0.12/AMD-APP-SDKInstaller-v3.0.130.135-GA-windows-F-x64.exe.1

StrikerRUS: I remember you used OpenCL Runtime from Intel. Are there any problems with it that we should propagate to LightGBM users in our docs?

tpboudreau: Thanks for the fast feedback!

> Uploading to LightGBM's GitHub Releases page right now

Thank you for uploading these files to github so quickly; yes, that's the place I had in mind.  

I've tested out downloading the parts from the new Github location into my private repo's azure pipeline and it still takes several minutes.  I'm surprised the hear that downloading a bigger file elsewhere takes so little time.  

Would you be willing to upload the full, combined installer file to the same github download/v2.0.12 location so I can see how that performs?  You can grab the full file here: https://gamma-rho.com/full/AMD-APP-SDKInstaller-v3.0.130.135-GA-windows-F-x64.exe .

Since Github is a reliable file server, if it's pretty fast to download the full file, I'll remove the cache task and combine downloading the full file with the install step, simplifying the pipeline.  But I think it if it takes more than about 5 minutes, it's probably worth keeping the cache step (it's a pretty uncomplicated use of their caching and the work is already done).

How does that sound?  

(BTW, to answer your other question, I'm only caching the installer program -- the full installation affects registry settings and who knows what else.  The only way to avoid the install step I think would be to have the OpenCL platform included in the base image.)

> Could you please elaborate about this line?
> > $module = "$env:ChocolateyInstall\helpers\chocolateyProfile.psm1"

After installing OpenCL you need to refresh the environment variables in the shell in order for programs to find it; chocolatey (which is pre-installed in the vs2017-win2016 image) has a RefreshEnv command to do this, but it wasn't working properly, even though it said it was succeeding.  

It seems that the way PS is invoked in the pipeline, it runs without a startup profile script which would normally contain lines like those above.  Without them, the RefreshEnv command was aliased incorrectly so that it was updating the environment variables in the outer shell "cmd.exe", not the current powershell session that we're running the tests in.  Including these lines in the test script just resets the command alias in the PS session so that RefreshEnv updates the correct environment and the LightGBM library finds the OpenCL runtime.

Since others will probably have the same question you did, I'll add a short comment in the script.

> could you please estimate how far we are from changing default device type to gpu in Windows wheels?

There were issues related to getting the pipeline to run at all and other issues related to actual test results (mostly floating point tolerance issues, but a few crashes iirc).  I believe we have solved the former in this PR, I haven't had a chance to investigate the latter yet, so I'll need a bit more time give an estimate and work up a PR if necessary on that. 

> I remember you used OpenCL Runtime from Intel. Are there any problems with it that we should propagate to LightGBM users in our docs?

I've run against the python package using three different versions the Intel CPU OpenCL platform.  They all run successfully and give reasonable results, but for some reason they either sporadically hang at shutdown or terminate with the user script returning 'False'.  Either of these will cause the pytests to "fail".  I haven't had time track down the reason for this behavior (it might be failure in the runtime to terminate a thread at shutdown or it might be something odd about how LightGBM is using OpenCL, though that seems less likely).  So while there's definitely some issue, I'm not sure how to report this usefully in the documents.  (BTW, the same library build linked into the LightGBM cli tool works without any apparent issues using the latest Intel CPU OpenCL runtime -- the issue appears to be limited to the python wheel.)

But as part of getting to using device=gpu as the build/test default, I may be able to get a clearer picture of what's going on and of course I'll let you all know.

StrikerRUS: > Would you be willing to upload the full, combined installer file to the same github download/v2.0.12 location so I can see how that performs?

Sure, done!
https://github.com/microsoft/LightGBM/releases/download/v2.0.12/AMD-APP-SDKInstaller-v3.0.130.135-GA-windows-F-x64.exe


> But I think it if it takes more than about 5 minutes, it's probably worth keeping the cache step

OK, agree.

> Since others will probably have the same question you did, I'll add a short comment in the script.

Thank you very much for the explanation. Yeah, small inline comment will be useful for others.

And many thanks for all your other clarifications as well!

tpboudreau: Thanks for uploading the full file.  I ran the build 5 times offline in an Azure pipeline and the full file download consistently took about 3 minutes.  So I removed the caching step and as you can see the download in this pipeline takes more than 25 minutes.  I'm not sure how to explain the discrepancy between these two pipelines or the very fast Linux download you mentioned above.

I'll leave it up to you whether you want me to put the caching task back in and/or download the file in parts (that seemed to go faster, but not enough trials to be sure).  Or I could leave it the way it is now (simple but slow).  Just let me know what you want me to do, thanks!

StrikerRUS: OK, thanks for trying direct downloading.

> I'm not sure how to explain the discrepancy between these two pipelines or the very fast Linux download you mentioned above.

I believe it's due to differences in geo locations of servers with Linux and Windows machines.

Please leave the fastest variant for now. And we will think about further optimisations outside of this PR.

tpboudreau: OK, I'm downloading the full file and using the pipeline cache for subsequent runs.  Very easy to change later if necessary.  I think this might be ready for final review and approval.  Thanks.

StrikerRUS: OK, thank you! I'll provide my review in a few days!

.ci/install_opencl.ps1
```diff
@@ -0,0 +1,45 @@
...
+    Write-Output "Setting EXIT"
+    $host.SetShouldExit(-1)
+    Exit -1
+  }
+}
+
+# Install OpenCL platform from installer executable expected in workspace cache
+
+Write-Output "Running OpenCL installer"
+Invoke-Command -ScriptBlock {Start-Process "$cache\$installer" -ArgumentList '/S /V"/quiet /norestart /passive /log opencl.log"' -Wait}
```
StrikerRUS, line 30: Just a question. What does this syntax mean? `/V" ... opencl.log"`. I mean double quotes.

StrikerRUS (approved):
Awesome contribution! Thanks a lot guys! Hope that ability to set default device to `gpu` will come soon.

I'll update docs to notify users about this useful feature that will allow them to simplify GPU version installation later today and will ping you to review if you don't mind.

.ci/install_opencl.ps1
tpboudreau: The installer package has a program inside a program: the outer (extractor) program takes the /S and /V arguments.  The value of the /V argument is the list of arguments you want passed to the inner (installer) program.  You need to quote them to make sure they're all passed together, not interpreted as (unrecognized) arguments to the outer program.  Since the ArgumentList is already single-quoted, you have to use double-quotes for the interior quotes.

tpboudreau:
> I'll update docs to notify users about this useful feature that will allow them to simplify GPU version installation later today and will ping you to review if you don't mind.

No problem.  Thanks for all your contributions to this PR.

PR #3287, opened 2020-08-08
dockerfile for cli-only on GPU

rsdmse:
A new dockerfile for CLI-only on GPU that uses multistage build to reduce the final image size from 2.6 GB to 104 MB (base image `nvidia/opencl` instead of `nvidia/cuda`).

The container was benchmarked on our HPC platform (Rivanna cluster, University of Virginia) using [this tutorial](https://lightgbm.readthedocs.io/en/latest/GPU-Tutorial.html#dataset-preparation). It took 33 seconds for 50 iterations on a GPU, compared to 92 seconds without GPU.

See issue #3286 for details.

PR #3408, opened 2020-09-25
added new GPU docker files

### LINKED ISSUE #3286: Dockerfile.gpu improvement
## Summary

A Dockerfile for the CLI is prepared and tested on GPU. Improvements are made to reduce the final image size compared to the current Dockerfile.gpu.

## Motivation

The final image size can be drastically reduced from 2.6 GB (Dockerfile.gpu but without python) to 111 MB. (Here I just compare CLI-only images. Adding conda and the necessary python packages would increase the final image size in both cases by roughly 1 GB; i.e. 3.6 GB vs 1.1 GB.)

## Description

Problems with the current Dockerfile.gpu:

1. The cleanup commands in Dockerfile.gpu do not reduce the image size. They must be in the same RUN statement as the apt-get/conda installation commands.

1. The use of `cuda*-cudnn*-devel` as the production base image is unnecessary.

I have prepared a CLI-only Dockerfile [here](https://github.com/uvarc/rivanna-docker/blob/master/lightgbm/2.3.1/Dockerfile). LightGBM version 2.3.1 is built from source with CUDA 10.2 and cuDNN 8 in a multistage build. The CLI binary and necessary libraries are copied to the final stage.

Note that the `cuda10.2-cudnn8-devel` image is almost 2.5 GB by itself. If this were used as the production image, I estimate the image size to be around 2.6 GB. Using `cuda10.2-base` as the production image, I obtained an image size of only 111 MB.

The container was benchmarked on our HPC platform (Rivanna cluster, University of Virginia) using [this tutorial](https://lightgbm.readthedocs.io/en/latest/GPU-Tutorial.html#dataset-preparation). It took 33 seconds for 50 iterations on a GPU, compared to 92 seconds without GPU.

It is not clear to me whether CUDA math libraries/NCCL/cuDNN are needed. They were not needed for the above tutorial at runtime. But if yes, we could use `runtime` as the production image. Its estimated image size of 1.2 GB would still be less than half of the current image size of 2.6 GB.

## References

Please feel free to test our container: https://hub.docker.com/r/uvarc/lightgbm

The tag must be provided explicitly since we do not use the `latest` tag:
```
docker pull uvarc/lightgbm:2.3.1
```

StrikerRUS:
Closed #3286.
Replaces #3287 due to CLA problems.

@rsdmse Thank you very much for your contribution! Please help to review this PR.

### DIFF
--- a/docker/gpu/dockerfile-cli-only.gpu
+++ b/docker/gpu/dockerfile-cli-only.gpu
@@ -1,0 +1,69 @@
+# Copyright (c) 2020 The Rector and Visitors of the University of Virginia
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated
+# documentation files (the "Software"), to deal in the Software without restriction, including without
+# limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the
+# Software, and to permit persons to whom the Software is furnished to do so, subject to the following
+# conditions:
+#
+# The above copyright notice and this permission notice shall be included in all copies or substantial portions
+# of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED
+# TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF
+# CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+# DEALINGS IN THE SOFTWARE.
+
+FROM nvidia/opencl:devel AS build
+ARG DEBIAN_FRONTEND=noninteractive
+ARG OPENCL_LIBRARIES=/usr/lib/x86_64-linux-gnu
+ARG OPENCL_INCLUDE_DIR=/usr/include/CL
+
+# SYSTEM
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        build-essential \
+        git \
+        ca-certificates \
+        libglib2.0-0 \
+        libxext6 \
+        libsm6 \
+        libxrender1 \
+        cmake \
+        libboost-dev \
+        libboost-system-dev \
+        libboost-filesystem-dev \
+        gcc \
+        g++ && \
+    rm -rf /var/lib/apt/lists/*
+
+# LightGBM
+WORKDIR /opt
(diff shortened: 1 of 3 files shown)


docker/gpu/dockerfile-cli-only.gpu
```diff
@@ -0,0 +1,69 @@
...
+        libboost-system-dev \
+        libboost-filesystem-dev \
+        gcc \
+        g++ && \
+    rm -rf /var/lib/apt/lists/*
+
+COPY --from=build \
+    /opt/LightGBM/lightgbm \
+    /opt/LightGBM/lib_lightgbm.so \
+    /opt/LightGBM/
```
StrikerRUS, line 61: Shouldn't we apply this patch for this image?
https://github.com/microsoft/LightGBM/blob/df37bcee2f9345eb4333e9561f169ba5079a8863/docker/gpu/dockerfile.gpu#L59-L61
https://github.com/microsoft/LightGBM/pull/3408/files#diff-1dd8c717ecdb4eb66c585ba89bbc2649R69

rsdmse: The file should be included in the `nvidia/opencl:devel` base image already, so in Line 69 I just simply copied the file:
```
COPY --from=build /etc/OpenCL/vendors/nvidia.icd /etc/OpenCL/vendors/nvidia.icd
```

StrikerRUS: Got it! Thanks for the detailed response!

jameslamb (commented):
looks great to me! I won't leave a :heavy_check_mark: because I don't have an environment to test this, but overall I'm ok with it. Exciting addition!

docker/gpu/dockerfile-cli-only.gpu
```diff
@@ -0,0 +1,69 @@
...
+
+ENV PATH /opt/LightGBM:${PATH}
+ENV LANG C.UTF-8
+ENV LC_ALL C.UTF-8
+
+LABEL maintainer=rs7wz@virginia.edu
```
jameslamb, line 67: since this code is being brought into LightGBM, I think that this label should be removed. If users have issues with this Dockerfile, it'll be LightGBM maintainers fixing it, right?

rsdmse: Right, please feel free to edit the Dockerfile! (It was copied directly from our repo.)

StrikerRUS: I was confused  by this line just like you @jameslamb! But I googled a little bit, and it seems that in Docker world `maintainer` means `author`...
> The MAINTAINER instruction sets the Author field of the generated images. 
   https://docs.docker.com/engine/reference/builder/#maintainer-deprecated

jameslamb: oh ok, I see. I think that we can just remove the label completely. I just don't want someone with an issue to send an email there instead of opening an issue in this repo, especially since after this PR is merge we might make other changes as LightGBM evolves

StrikerRUS: OK, removed in a9d85c3e50bace8afd770d190ed62a83f60366f5 (also replaced tabs with whitespaces).

guolinke (approved):

PR #3411, opened 2020-09-27
[ci] removed unused environment variable CONDA_ENV

jameslamb:
I realized today that the environment variable `CONDA_ENV` is not used by any of the R CI jobs. So it can safely be removed from the GitHub Actions config for this jobs.


StrikerRUS (approved):
Thanks!

PR #3423, opened 2020-09-30
[python] Use ctypes for parameters of DLL functions for Dataset

StrikerRUS:
Continuation of #3419.


guolinke (approved):

StrikerRUS:
Unfortunately, this is **not** fixing issue #1818.

PR #3160, opened 2020-06-11
Add support for CUDA-based GPU build

ChipKerchner:
This is the initial CUDA work.  It should work similarly to the GPU/OCL work.

To compile use - 'USE_CUDA=1'.  Python unit tests should include 'device': 'cuda' where needed.

All unit tests pass for CPU, GPU/OCL and CUDA.  CPU & CUDA were tested on ppc64le and GPU/OCL was tested on x86_64

PR #3424, opened 2020-09-30
setup CUDA CI job

StrikerRUS:
Closed #3402.

PR #3428, opened 2020-10-02
[docs] document CUDA version support

### LINKED ISSUE #3416: Traning model on multi GPUs?
I prefer to use multi GPUs to speed up traning model, but there is only few information about this in the document. Does lightGBM works well on multi GPUs ?

StrikerRUS:
In follow up of #3160.

Closed #3416.

cc @ChipKerchner


docs/Installation-Guide.rst
```diff
@@ -558,6 +558,35 @@ Docker
...
 Refer to `GPU Docker folder <https://github.com/microsoft/LightGBM/tree/master/docker/gpu>`__.
 
+Build CUDA Version (Experimental)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The `original GPU version <#build-gpu-version>`__ of LightGBM is based on OpenCL. This version is a separate implementation and requires NVIDIA graphics card with compute capability 6.0 and higher.
+
+This version is very experimental and we suggest using it only when it is impossible to use OpenCL version, for instance, on IBM POWER microprocessors.
+
+**Note**: only Linux is supported, other operating systems are not supported yet.
```
jameslamb, line 568:
```suggestion
The `original GPU build <#build-gpu-version>`__ of LightGBM is based on OpenCL.

The CUDA-based build is a separate implementation and requires an NVIDIA graphics card with compute capability 6.0 and higher. It should be considered experimental, and we suggest using it only when it is impossible to use OpenCL version (for example, on IBM POWER microprocessors).

**Note**: only Linux is supported, other operating systems are not supported yet.
```

When I read this, I was really confused whether "this version" refers to OpenCL or CUDA. Could we clarify it?

docs/Installation-Guide.rst
```diff
@@ -558,6 +558,35 @@ Docker
...
+
+Linux
+^^^^^
+
+On Linux a CUDA version of LightGBM can be built using **CUDA**, **CMake** and **gcc** or **Clang**.
+
+The following dependencies should be installed before compilation:
+
+-  **CUDA** libraries.
```
jameslamb, line 577: is there some documentation from NVIDIA that we can link to here? I think "install CUDA libraries" is too vague.

Maybe something like [this](https://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html)?

docs/Installation-Guide.rst
```diff
@@ -558,6 +558,35 @@ Docker
...
+-  **CMake** 3.16 or later.
+
+To build LightGBM CUDA version, run the following commands:
+
+.. code::
+
+  git clone --recursive https://github.com/microsoft/LightGBM ; cd LightGBM
+  mkdir build ; cd build
+  cmake -DUSE_CUDA=1 ..
+  make -j4
```
jameslamb, line 588: I have a slight preference for putting each command on its own line, instead of grouping them with `;`. That makes it a little easier to understand at first glance, in my opinion.

```suggestion
  git clone --recursive https://github.com/microsoft/LightGBM
  cd LightGBM
  mkdir build
  cd build
  cmake -DUSE_CUDA=1 ..
  make -j4
```

What do you think? If you disagree, it's ok to leave it as-is. It's not worth holding up the PR.

docs/Parameters.rst
```diff
@@ -1124,7 +1124,9 @@ GPU Parameters
 
 -  ``gpu_use_dp`` :raw-html:`<a id="gpu_use_dp" title="Permalink to this parameter" href="#gpu_use_dp">&#x1F517;&#xFE0E;</a>`, default = ``false``, type = bool
 
-   -  set this to ``true`` to use double precision math on GPU (by default single precision is used in OpenCL implementation and double precision is used in CUDA implementation)
+   -  set this to ``true`` to use double precision math on GPU (by default single precision is used)
+
+   -  **Note**: can be used only in OpenCL implementation, in CUDA implementation only double precision is supported yet
```
jameslamb, line 1129:
```suggestion
   -  **Note**: can be used only in OpenCL implementation, in CUDA implementation only double precision is currently supported
```

include/LightGBM/config.h
```diff
@@ -967,7 +967,8 @@ struct Config {
   // desc = **Note**: refer to `GPU Targets <./GPU-Targets.rst#query-opencl-devices-in-your-system>`__ for more details
   int gpu_device_id = -1;
 
-  // desc = set this to ``true`` to use double precision math on GPU (by default single precision is used in OpenCL implementation and double precision is used in CUDA implementation)
+  // desc = set this to ``true`` to use double precision math on GPU (by default single precision is used)
+  // desc = **Note**: can be used only in OpenCL implementation, in CUDA implementation only double precision is supported yet
```
jameslamb, line 971:
```suggestion
  // desc = **Note**: can be used only in OpenCL implementation, in CUDA implementation only double precision is currently supported
```

python-package/README.rst
```diff
@@ -110,6 +110,17 @@ All available options:
...
 For more details see `FindBoost <https://cmake.org/cmake/help/latest/module/FindBoost.html>`__ and `FindOpenCL <https://cmake.org/cmake/help/latest/module/FindOpenCL.html>`__.
 
+Build CUDA Version
+~~~~~~~~~~~~~~~~~~
+
+.. code:: sh
+
+    pip install lightgbm --install-option=--cuda
+
+All remarks from `Build from Sources section <#build-from-sources>`__ are actual in this case and `CMake`_ (version 3.16 or higher) is strongly required.
```
jameslamb, line 120:
```suggestion
All requirements from `Build from Sources section <#build-from-sources>`__ apply for this installation option as well, and `CMake`_ (version 3.16 or higher) is strongly required.
```

python-package/README.rst
```diff
@@ -166,6 +177,8 @@ Run ``python setup.py install --mingw``, if you want to use **MinGW-w64** on **W
 
 Run ``python setup.py install --gpu`` to enable GPU support. All remarks from `Build GPU Version section <#build-gpu-version>`__ are actual in this case. To pass additional options to **CMake** use the following syntax: ``python setup.py install --gpu --opencl-include-dir=/usr/local/cuda/include/``, see `Build GPU Version section <#build-gpu-version>`__ for the complete list of them.
 
+Run ``python setup.py install --cuda`` to enable CUDA support. All remarks from `Build CUDA Version section <#build-cuda-version>`__ are actual in this case.
```
jameslamb, line 180:
```suggestion
Run ``python setup.py install --cuda`` to enable CUDA support. All requirements from `Build CUDA Version section <#build-cuda-version>`__ apply for this installation option as well.
```

jameslamb (changes requested):
thanks for doing this! I left a few suggestions on clarifying the language, but otherwise I agree with these changes.

docs/Installation-Guide.rst
StrikerRUS: I totally agree! But could it be done in a separate PR to change all occurrences of such commands across the doc?

python-package/README.rst
StrikerRUS: This is standard phrase across this doc. Again, if you do not mind I'll change it in a separate PR to keep the doc consistent.

python-package/README.rst
StrikerRUS: The same about a separate PR to change the wording of all paragraphs at once.

docs/Installation-Guide.rst
jameslamb: oh sure, no problem. I didn't know there were other places that do this. It can be another PR, and not urgent at all

python-package/README.rst
jameslamb: ah ok no problem. I was just taking the diff as-is...didn't realize some other these things were patterns reused elsewhere in the doc. Changing them in a different PR is ok and not urgent.

docs/Installation-Guide.rst
StrikerRUS: Fixed in 4a175a9625fe2542abb600bcd6647927157b4dbe.

docs/Installation-Guide.rst
StrikerRUS: Fixed in 4a175a9625fe2542abb600bcd6647927157b4dbe.

docs/Parameters.rst
StrikerRUS: Fixed in 4a175a9625fe2542abb600bcd6647927157b4dbe.

include/LightGBM/config.h
StrikerRUS: Fixed in 4a175a9625fe2542abb600bcd6647927157b4dbe.

docs/Installation-Guide.rst
StrikerRUS: Thanks! I'll prepare a new PR right after merging this one.

jameslamb (approved):
looks good, thank you for this!

StrikerRUS:
Depends on fixing CUDA error reported in #3450 and https://github.com/microsoft/LightGBM/pull/3424#issuecomment-702189313.

bordaw:
Do you know when the official LightGBM documentation will be updated to describe the CUDA support?

StrikerRUS:
@bordaw Right after fixing critical bugs of the initial CUDA implementation: https://github.com/microsoft/LightGBM/pull/3428#issuecomment-721307376. Unfortunately, it seems that it won't be very quickly: https://github.com/microsoft/LightGBM/pull/3160#issuecomment-745338726.

bordaw:
Thanks for the prompt response! Please let me know if I can help in the fix as I am one of the original authors of the CUDA implementation.

StrikerRUS:
@bordaw Wow, this will be just awesome! Thank you very much for wishing to help!

I think you can try to reproduce the issue locally first: https://github.com/microsoft/LightGBM/pull/3160#issuecomment-724301806. Please let me know if you need any additional details.

bordaw:
Thanks! Will download the latest code and take a look..

StrikerRUS:
@jameslamb @guolinke Please (re-)review when have time.

jameslamb (approved):
I have no changes to suggest, this looks good to me. Thanks for adding the explanation on how the CUDA build is different from OpenCL and why you might use it. Really helpful.

PR #3429, opened 2020-10-02
[ci] rename R-package GitHub Actions workflow

StrikerRUS:
Renaming is for the better visual distinguishing jobs at `Actions` tab.

image

### DIFF
--- a/.ci/test_r_package.sh
+++ b/.ci/test_r_package.sh
@@ -7,10 +7,6 @@ mkdir -p $R_LIB_PATH
 export R_LIBS=$R_LIB_PATH
 export PATH="$R_LIB_PATH/R/bin:$PATH"
 
-# hack to get around this:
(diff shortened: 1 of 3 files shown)


jameslamb (changes requested):
I am ok with the rename but I think we should keep the environment variable that suppresses that time check in.

It's very inexpensive (just setting an environment variable) and saves us maintenance time. As recently as last night, I was still seeing the "unable to verify current time" when running `R CMD check` locally, so I'm pretty sure the problem still exists.

StrikerRUS:
@jameslamb 

> As recently as last night, I was still seeing the "unable to verify current time" when running R CMD check locally, so I'm pretty sure the problem still exists.

Ah, OK! I just thought that that error is super rare one 😄 .

jameslamb (approved):
thanks!

PR #3430, opened 2020-10-03
Update callback.R with keyword arguments

philip-khor:
Addresses part of #3390. Also added `isTRUE()` for some of the logical checks, please advise if this is okay

I also had issues with generating the roxygen2 documentation: was not able to run the roxygenize step due to this error: 
```
Error in .f(.x[[i]], ...) : 
  R6 class (lgb.Booster) without source references. If you use the `installed` load method in `DESCRIPTION`, then try re-installing the package with option '--with-keep.source'. E.g. `install.packages(..., INSTALL_OPTS = "--with-keep.source")`.
```

The steps I did are as follows:
1. open LightGBM as R project in RStudio 
2. latest roxygen2 version on CRAN is 7.1.1 -> `install.packages("roxygen2")` works
3. regular 'sh' did not work: to workaround, in R console, run `system("./build-cran-package.sh")`. Did this on WSL2. 
4. build package with `R CMD INSTALL lightgbm_3.*.tar.gz` (note underscore)

### DIFF
--- a/R-package/R/callback.R
+++ b/R-package/R/callback.R
@@ -80,7 +80,7 @@ cb.reset.parameters <- function(new_params) {
 
     # Check if rounds is null
     if (is.null(nrounds)) {
-      init(env)
+      init(env = env)
     }
 
     # Store iteration
@@ -95,7 +95,7 @@ cb.reset.parameters <- function(new_params) {
     })
 
     if (!is.null(env$model)) {
-      env$model$reset_parameter(pars)
+      env$model$reset_parameter(params = pars)
     }
 
   }
@@ -141,12 +141,12 @@ merge.eval.string <- function(env) {
 
     # Store evaluation error
     eval_err <- NULL
-    if (is_eval_err) {
+    if (isTRUE(is_eval_err)) {
       eval_err <- env$eval_err_list[[j]]
     }
 
     # Set error message
-    msg <- c(msg, format.eval.string(env$eval_list[[j]], eval_err))
+    msg <- c(msg, format.eval.string(eval_res = env$eval_list[[j]], eval_err = eval_err))
 
   }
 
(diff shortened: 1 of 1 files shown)


jameslamb:
Thanks @philip-khor !

I just fixed the typo (`lightgbm_` instead of `lightgbm.`) in #3390 , sorry about that!

I remember seeing that `{roxygen2}` error a while ago...maybe it worked differently for me because I had previous installations of `{lightgbm}` around. I just updated #3390 to include the `--with-keep.source` flag. 

<hr>

Can you clarify what it means that `sh` "did not work"?

jameslamb (approved):
This looks great, thank you so much for contributing!

I'd like to leave the "keyword arguments for internal functions" fix on other files from #3390 for other Hacktoberfest contributes, but if you have the time and interest we'd welcome another pull request that fixes other places where you think we should be using `isTRUE()`.

philip-khor:
Okay that's weird, `sh build-cran-package.sh` works now. 

But I remember previously hitting errors along the lines of "missing options for set", even when I replaced it with `bash`.

philip-khor:
> 
> 
> This looks great, thank you so much for contributing!
> 
> I'd like to leave the "keyword arguments for internal functions" fix on other files from #3390 for other Hacktoberfest contributes, but if you have the time and interest we'd welcome another pull request that fixes other places where you think we should be using `isTRUE()`.

Got it, thanks!

jameslamb:
> Okay that's weird, `sh build-cran-package.sh` works now.
> 
> But I remember previously hitting errors along the lines of "missing options for set", even when I replaced it with `bash`.

hmmm ok, I'm not sure what could cause that. If you see it again, please open an issue and we'll take a look.

PR #3431, opened 2020-10-03
Add support for cuda version less then 10.0

shipengfei92:
For `cudaPointerAttributes`, the attribute name has been updated from `memoryType` to `type` since 10.0, add support for compiling version less then 10.0 .
https://docs.nvidia.com/cuda/archive/9.2/cuda-runtime-api/structcudaPointerAttributes.html#structcudaPointerAttributes
https://docs.nvidia.com/cuda/archive/10.0/cuda-runtime-api/structcudaPointerAttributes.html#structcudaPointerAttributes

### DIFF
--- a/include/LightGBM/cuda/vector_cudahost.h
+++ b/include/LightGBM/cuda/vector_cudahost.h
@@ -65,9 +65,15 @@ struct CHAllocator {
       if (LGBM_config_::current_device == lgbm_device_cuda) {
         cudaPointerAttributes attributes;
         cudaPointerGetAttributes(&attributes, p);
-        if ((attributes.type == cudaMemoryTypeHost) && (attributes.devicePointer != NULL)) {
-          cudaFreeHost(p);
-        }
+        #if CUDA_VERSION >= 10000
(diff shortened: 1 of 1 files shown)


guolinke (approved):

StrikerRUS:
@guolinke I think we should explicitly set minimal CUDA version we want to support. At least in our docs or better in CMake file. Any ideas what version it should be?

guolinke:
@StrikerRUS I am not sure, maybe we can recommend to use cuda>=10.0, but cannot guarantee when cuda <10.0 ?

StrikerRUS:
@guolinke 
Our friends at XGBoost explicitly require CUDA 10.0 at least:
https://github.com/dmlc/xgboost/blob/c8ec62103a36f1717d032b1ddff2bf9e0642508a/CMakeLists.txt#L126-L128

TBH, I think that new (and marked as experimental) feature is a good chance to start from the most recent dependency version to use all available efficiency opportunities and do not struggle with backward compatibility with earlier versions.

jameslamb:
^ I agree with @StrikerRUS

PR #3435, opened 2020-10-04
[R-package] [ci] Make Windows CI fail if R CMD check fails

jameslamb:
In https://github.com/microsoft/LightGBM/pull/3405#issuecomment-702207635, I noticed that if `R CMD check` raises an `ERROR` on Windows, LightGBM's CI doesn't fail. Maybe that is a result of the way we have to intercept `stderr` on GitHub Actions Windows environments (https://github.com/microsoft/LightGBM/blob/0fc4c5c490f86b67e2aa8a8f202f3331035790c0/.ci/test_r_package_windows.ps1#L15-L28).

This PR proposes a fix, to be sure that CI fails on changes that break the R package.

### DIFF
--- a/.ci/test_r_package_windows.ps1
+++ b/.ci/test_r_package_windows.ps1
@@ -185,6 +185,11 @@ if ($env:COMPILER -ne "MSVC") {
   Get-Content -Path "$INSTALL_LOG_FILE_NAME"
   Write-Output "----- end of build and install logs -----"
   Check-Output $install_succeeded
+  # some errors are not raised above, but can be found in the logs
+  if (Get-Content "$INSTALL_LOG_FILE_NAME" | Select-String -Pattern "ERROR" -Quiet) {
+      echo "ERRORs have been found installing lightgbm"
+      Check-Output $False
+  }
 }
 
 # Checking that we actually got the expected compiler. The R package has some logic
--- a/.ci/test_r_package_windows.ps1
+++ b/.ci/test_r_package_windows.ps1
@@ -25,7 +25,6 @@ function Run-R-Code-Redirect-Stderr {
   )
   $decorated_code = "out_file <- file(tempfile(), open = 'wt'); sink(out_file, type = 'message'); $rcode; sink()"
   Rscript --vanilla -e $decorated_code
-  Exit $?
 }
 
 # Get details needed for installing R components
@@ -154,9 +153,6 @@ if ($env:COMPILER -ne "MSVC") {
   }
   Run-R-Code-Redirect-Stderr "result <- processx::run(command = 'R.exe', args = $check_args, echo = TRUE, windows_verbatim_args = FALSE, error_on_status = TRUE)" ; $check_succeeded = $?
 
-  Write-Output "------------------"
-  Write-Output "exit code: $check_succeeded"
-  Write-Output "------------------"
   Write-Output "R CMD check build logs:"
   $INSTALL_LOG_FILE_NAME = "lightgbm.Rcheck\00install.out"
   Get-Content -Path "$INSTALL_LOG_FILE_NAME"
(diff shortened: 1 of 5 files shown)


R-package/R/aliases.R
```diff
@@ -6,6 +6,7 @@
 #               R sources files during installation).
 # [return] A named list, where each key is a parameter relevant to lgb.DataSet and each value is a character
 #          vector of corresponding aliases.
+,,,,
```
jameslamb, line 9: I'm starting by intentionally pushing a change that raises an ERROR in `R CMD check`, so we can see that the Windows tests pass.

jameslamb: even with this syntax error, all of the Windows tests passed!

https://github.com/microsoft/LightGBM/pull/3435/checks?check_run_id=1204672524

image

image

jameslamb: After adding https://github.com/microsoft/LightGBM/pull/3435/commits/f9165756bbae605f7710058dde878cd325da4acb, the MINGW tests fail as expected

https://github.com/microsoft/LightGBM/pull/3435/checks?check_run_id=1204732555

image

But the MSVC ones do not! Another issue with how errors are caught by `Run-R-Code-Redirect-Stderr`, maybe.

jameslamb: after adding https://github.com/microsoft/LightGBM/pull/3435/commits/e1916737de3d4d71e456da70a83b69af89c7e6d1, all Windows checks now fail as expected

https://github.com/microsoft/LightGBM/pull/3435/checks?check_run_id=1204756009

image

StrikerRUS (approved):
LGTM! Looks like we cannot rely on `Run-R-Code-Redirect-Stderr`...

.ci/test_r_package_windows.ps1
```diff
@@ -181,6 +185,11 @@ if ($env:COMPILER -ne "MSVC") {
   Get-Content -Path "$INSTALL_LOG_FILE_NAME"
   Write-Output "----- end of build and install logs -----"
   Check-Output $install_succeeded
+  # some errors are not raised above, but can be found in the logs
+  if (Get-Content "$INSTALL_LOG_FILE_NAME" | Select-String -Pattern "ERROR" -Quiet) {
+      echo "ERRORs have been found installing lightgbm"
```
StrikerRUS, line 190: Oops! Seems that during VS trials and **errors** (we use them to find any supported VS version) we trigger this condition:
```
Trying 'Visual Studio 16 2019'

-- Configuring incomplete, errors occurred!
```

StrikerRUS (changes requested):

.ci/test_r_package_windows.ps1
jameslamb: I always forget that `Select-String` is case-insensitive by default 🙃

jameslamb:
> LGTM! Looks like we cannot rely on Run-R-Code-Redirect-Stderr...

Oh wait, I see why! I might be able to fix this by fixing `Run-R-Code-Redirect-Stderr`.

My next few commits here will be to test...I'll comment when it's ready for another review

.ci/test_r_package_windows.ps1
jameslamb: > > LGTM! Looks like we cannot rely on Run-R-Code-Redirect-Stderr...
> 
> Oh wait, I see why! I might be able to fix this by fixing `Run-R-Code-Redirect-Stderr`.
> 
> My next few commits here will be to test...I'll comment when it's ready for another review

ok sadly my experiment failed. I thought I just needed to add a return or Exit in `Run-R-Code-Redirect-Stderr`  but that didn't work. All of the ways we have to catch errors makes it really complicated :/

jameslamb: fixed!

StrikerRUS (approved):
Thanks for the neat solution!

PR #3444, opened 2020-10-08
[R-package] remove deprecation warnings on lgb.prepare() functions

jameslamb:
In `LightGBM` 3.0.0, the `lgb.prepare*` family of functions were removed (#3075 , #3095). That release still technically left those functions in `{lightgbm}`, but had them throw an error encouraging the use of `lgb.convert()` and`lgb.convert_with_rules()`.

This PR proposes removing those functions completely. It's been more than a month since `LightGBM` 3.0.0, so I think the deprecation has been in the package long enough.

Removing these functions should slightly reduce the time it takes to build, install, and test the R package, and slightly reduce the size of the installed package.


jameslamb:
gently ping @Laurae2 or @guolinke for review. I think this would be a very quick review, whenever you have time.

guolinke (approved):

PR #3449, opened 2020-10-09
pin Ubuntu version in distroless docker file

StrikerRUS:
Fix for #3408.

I believe this is needed because when Ubuntu version will be updated software versions might be updated as well and these lines will fail due to specific versions of libraries
https://github.com/microsoft/LightGBM/blob/3f71d7546778e682e6036ebcda1af5b5b6f720f2/docker/gpu/dockerfile-cli-only-distroless.gpu#L52-L68

For example, in Ubuntu 20.04 default version of `libboost` is 1.71: https://packages.ubuntu.com/focal/libboost-dev.

@rsdmse  Please review.


rsdmse:
@StrikerRUS Yes, that's a good call. I totally agree.

StrikerRUS:
@rsdmse Thanks a lot for your prompt response!

jameslamb (approved):
I agree, this is a good change

PR #3460, opened 2020-10-16
[R-package] Suppresses unknown pragma warnings during CRAN build

### LINKED ISSUE #3433: [R-package] Suppress 'warning: unknown pragma' for CRAN package
## Description of the problem

As of the changes in #3160 , builds of the CRAN version of the R package with `gcc` now raise "unknown pragma" warnings.

CRAN does not allow the use of `-Wunknown-pragmas` to suppress such warnings, so we remove them with `sed` when building the R package.

The "bug" here is that LightGBM's CI did not catch this. That should also be fixed.

## How to fix this

Add `pragma: warning` to the set of checks at https://github.com/microsoft/LightGBM/blob/5b0ab05c3d69eb9c0015dcadccaf8ad637b587c0/build-cran-package.sh#L53-L65.

Add some code to the CI script for Mac/Linux that tests for the presence of this warning and fails if it finds it. Similar to  https://github.com/microsoft/LightGBM/blob/5b0ab05c3d69eb9c0015dcadccaf8ad637b587c0/.ci/test_r_package.sh#L206-L216.

## How fixing this improves `{lightgbm}`

This warning does not affect the correctness of an installation, but showing it in logs may waste users' time by leading them to believe that something is wrong with their installation.

## Environment info

LightGBM version or commit hash: latest `master` (https://github.com/microsoft/LightGBM/commit/5b0ab05c3d69eb9c0015dcadccaf8ad637b587c0)

image

```text
clang++ -mmacosx-version-min=10.13 -std=gnu++11 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG -I./include -DUSE_SOCKET -DLGB_R_BUILD  -I/usr/local/include  -Xclang -fopenmp -pthread -fPIC  -Wall -g -O2  -c treelearner/voting_parallel_tree_learner.cpp -o treelearner/voting_parallel_tree_learner.o
In file included from treelearner/voting_parallel_tree_learner.cpp:11:
In file included from treelearner/parallel_tree_learner.h:15:
treelearner/cuda_tree_learner.h:255:13: warning: unknown pragma ignored [-Wunknown-pragmas]
    #pragma warning(disable : 4702)
            ^
1 warning generated.
```

## Reproducible example(s)

On a Mac, using `gcc`:

```shell
sh build-cran-package.sh

R CMD INSTALL lightgbm_*.tar.gz
```

cctechwiz:
This PR addresses pragma warnings that gcc was outputting since CRAN builds cannot suppress them. See issue #3433 for more information and recreation steps.

Major changes:
1. `build-cran-package.sh` was modified to remove `#pragma warning`s from all `h`, `hpp`, and `cpp` files in the temporary build folder (`lightgbm_r`).
2. An additional check was added to the end of the CI script `test_r_package.sh` that ensures no unknown pragma warnings were omitted during the build or else it will exit.

Closes #3433

### DIFF
--- a/.ci/test_r_package.sh
+++ b/.ci/test_r_package.sh
@@ -226,7 +226,7 @@ if [[ $R_BUILD_TYPE == "cran" ]]; then
         | wc -l
     )
     if [[ $pragma_warning_present -ne 0 ]]; then
-        echo "Unknown pragma warning is present, pragmas should have been remove before build"
+        echo "Unknown pragma warning is present, pragmas should have been removed before build"
         exit -1
     fi
 fi
--- a/.ci/test_r_package.sh
+++ b/.ci/test_r_package.sh
@@ -219,7 +219,7 @@ fi
 
 # this check makes sure that no "warning: unknown pragma ignored" logs
 # reach the user leading them to believe that something went wrong
-if [[ $OS_NAME == "macos" || $OS_NAME == "linux" ]] && [[ $R_BUILD_TYPE == "cran" ]]; then
+if [[ $R_BUILD_TYPE == "cran" ]]; then
     pragma_warning_present=$(
         cat $BUILD_LOG_FILE \
         | grep -E "warning: unknown pragma ignored" \
--- a/.ci/test_r_package.sh
+++ b/.ci/test_r_package.sh
@@ -216,3 +216,17 @@ if [[ $OS_NAME == "macos" ]] && [[ $R_BUILD_TYPE == "cran" ]]; then
         exit -1
     fi
 fi
+
+# this check makes sure that no "warning: unknown pragma ignored" logs
+# reach the user leading them to believe that something went wrong
+if [[ $OS_NAME == "macos" || $OS_NAME == "linux" ]] && [[ $R_BUILD_TYPE == "cran" ]]; then
+    pragma_warning_present=$(
+        cat $BUILD_LOG_FILE \
+        | grep -E "warning: unknown pragma ignored" \
+        | wc -l
+    )
+    if [[ $pragma_warning_present -ne 0 ]]; then
+        echo "Unknown pragma warning is present, pragmas should have been remove before build"
+        exit -1
+    fi
+fi
--- a/build-cran-package.sh
+++ b/build-cran-package.sh
@@ -55,7 +55,7 @@ cd ${TEMP_R_DIR}
     # not allow you to use compiler flag '-Wno-unknown-pragmas' or
     # pragmas that suppress warnings.
     echo "Removing unknown pragmas in headers"
-    for file in $(find . -regex ".*\.\(h\|cpp\|hpp\)$"); do
+    for file in $(find . -regex '.*\.\(h\|cpp\|hpp\)$'); do
       sed \
         -i.bak \
         -e 's/^.*#pragma region.*$//' \
@@ -63,7 +63,7 @@ cd ${TEMP_R_DIR}
         -e 's/^.*#pragma warning.*$//' \
         "${file}"
     done
-    find . -regex ".*\.\(h\|cpp\|hpp\)\.bak$" -exec rm {} \;
+    find . -regex '.*\.\(h\|cpp\|hpp\)\.bak$' -exec rm {} \;
 
     # When building an R package with 'configure', it seems
     # you're guaranteed to get a shared library called
--- a/build-cran-package.sh
+++ b/build-cran-package.sh
@@ -50,19 +50,20 @@ cd ${TEMP_R_DIR}
     sed -i.bak -e "s/~~VERSION~~/${LGB_VERSION}/" DESCRIPTION
     sed -i.bak -e "s/~~DATE~~/${CURRENT_DATE}/" DESCRIPTION
 
-    # Remove 'region' and 'endregion' pragmas. This won't change
-    # the correctness of the code. CRAN does not allow you
-    # to use compiler flag '-Wno-unknown-pragmas' or
+    # Remove 'region', 'endregion', and 'warning' pragmas.
+    # This won't change the correctness of the code. CRAN does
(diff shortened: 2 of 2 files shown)


.ci/test_r_package.sh
```diff
@@ -216,3 +216,17 @@ if [[ $OS_NAME == "macos" ]] && [[ $R_BUILD_TYPE == "cran" ]]; then
         exit -1
     fi
 fi
+
+# this check makes sure that no "warning: unknown pragma ignored" logs
+# reach the user leading them to believe that something went wrong
+if [[ $OS_NAME == "macos" || $OS_NAME == "linux" ]] && [[ $R_BUILD_TYPE == "cran" ]]; then
```
jameslamb, line 222: I love this extra check in CI, thanks!

You can safely remove this first condition. This shell script will only be run on Linux and Mac builds.

We have a separate one for Windows: https://github.com/microsoft/LightGBM/blob/master/.ci/test_r_package_windows.ps1

```suggestion
if [[ $R_BUILD_TYPE == "cran" ]]; then
```

build-cran-package.sh
```diff
@@ -50,19 +50,20 @@ cd ${TEMP_R_DIR}
...
+    for file in $(find . -regex ".*\.\(h\|cpp\|hpp\)$"); do
       sed \
         -i.bak \
         -e 's/^.*#pragma region.*$//' \
         -e 's/^.*#pragma endregion.*$//' \
+        -e 's/^.*#pragma warning.*$//' \
         "${file}"
     done
-    rm src/include/LightGBM/*.h.bak
+    find . -regex ".*\.\(h\|cpp\|hpp\)\.bak$" -exec rm {} \;
```
jameslamb, line 66: oh nice solution! I didn't know about `find -regex`, I like this. Thanks for also including `.hpp`.

jameslamb (commented):
Thanks for taking the time to contribute to LightGBM, @cctechwiz !

I left one suggestion. Other than that, I'll wait to approve until CI builds and I can go check the logs just to double-check that the new code you added in `.ci/test_r_package.sh` is working.

build-cran-package.sh
```diff
@@ -50,19 +50,20 @@ cd ${TEMP_R_DIR}
...
+    # not allow you to use compiler flag '-Wno-unknown-pragmas' or
     # pragmas that suppress warnings.
     echo "Removing unknown pragmas in headers"
-    for file in src/include/LightGBM/*.h; do
+    for file in $(find . -regex ".*\.\(h\|cpp\|hpp\)$"); do
       sed \
         -i.bak \
         -e 's/^.*#pragma region.*$//' \
         -e 's/^.*#pragma endregion.*$//' \
+        -e 's/^.*#pragma warning.*$//' \
```
jameslamb, line 63: from https://github.com/microsoft/LightGBM/pull/3460/checks?check_run_id=1265073334, it looks like this did not work for all cases

image

but I'm not sure why because your code looks correct :thinking:

.ci/test_r_package.sh
```diff
@@ -216,3 +216,17 @@ if [[ $OS_NAME == "macos" ]] && [[ $R_BUILD_TYPE == "cran" ]]; then
...
+# this check makes sure that no "warning: unknown pragma ignored" logs
+# reach the user leading them to believe that something went wrong
+if [[ $R_BUILD_TYPE == "cran" ]]; then
+    pragma_warning_present=$(
+        cat $BUILD_LOG_FILE \
+        | grep -E "warning: unknown pragma ignored" \
+        | wc -l
+    )
+    if [[ $pragma_warning_present -ne 0 ]]; then
+        echo "Unknown pragma warning is present, pragmas should have been remove before build"
```
jameslamb, line 229:
```suggestion
        echo "Unknown pragma warning is present, pragmas should have been removed before build"
```

sorry, missed this typo in the first review

jameslamb (changes requested):

build-cran-package.sh
jameslamb: I tried `sh build-cran-package.sh` and then running `find . -regex ".*\.\(h\|cpp\|hpp\)$"` from inside `lightgbm_r`...I see `cuda_tree_learner.h` in the list of files. So I don't think it's a problem with the `find` command

But maybe there's a difference between my shell and the one running in this CI job. Could you try using single quotes for the argument to `regex`?

```shell
find . -regex '.*\.\(h\|cpp\|hpp\)$'
```

Maybe that trailing `$` sign is being interpreted as the start of interpretation instead of a literal for the regex? I know it feels like that ISN'T it, but it's the only idea I have right now

cctechwiz: I changed the quotes, but it looks like our issue is going to be more along the lines of what is posted here: https://stackoverflow.com/questions/5905493/why-isnt-this-regex-working-find-regex-m-h

I'll figure out a more portable way to do that find and make another commit soon. (gotta run to work now)

cctechwiz: Looks like `find ./lightgbm_r/ -name *.h -o -name *.hpp -o -name *.cpp` will work too, it's just ugly. If you can confirm it works on macOS I'll change the code to use that.

cctechwiz: This might work too:
```
find -E . -regex '.*\.(h|cpp|hpp)$'
```

jameslamb: oooooo this is one of those "some of the unix tools work different on Mac" things. yikes haha

Ok thanks for investigating. I might push some attempts later today, when I can try it out on my Mac.

cctechwiz: I had a few extra minutes this morning so just made that change and pushed it so we could just see if the checks passed. And it looks like they have.

jameslamb (approved):
I just looked through all the logs for the CRAN jobs, and can confirm this suppressed those warnings. Thanks so much for the contribution @cctechwiz :tada:

cctechwiz:
Glad I could help!

PR #3443, opened 2020-10-08
[ci] [R-package] Fix memory leaks found by valgrind

jameslamb:
In #3338 , `{lightgbm}` was rejected from CRAN for multiple failed submission attempts. As of that submission, the package was failing tests with a version of R instrumented to use `valgrind`.

https://cran-archive.r-project.org/web/checks/2020/2020-10-02_check_results_lightgbm.html

image

This PR attempts to add a CI job that replicates these checks.

## Notes

* This is a work in progress. I'll add reviewers when I think it's ready.
* Once I confirm that the job fails as it should, I'll try merging in the changes from https://github.com/microsoft/LightGBM/tree/fix-memory-leakage, which might fix the issues! Then we can develop on this branch until those issues are fixed
* If the job takes too long, I'll switch it to an optional job triggered by a comment (like #3402 )

PR #3468, opened 2020-10-17
[ci] ignore R CMD CHECK warnings on new R version

jameslamb:
In CI for the R package,  we check for `R CMD CHECK` warnings by looking for the string "warning" in its logs. That test is currently breaking builds (https://github.com/microsoft/LightGBM/runs/1268855023)  for a  warning that  can safely be ignored:

> Warning: package 'R6' was built under R version 4.0.3

This PR proposes ignoring those warnings. It also slightly simplifies the config for the UBSAN tests added in #3439  (based on https://github.com/microsoft/LightGBM/pull/3443#discussion_r506961056)

### DIFF
--- a/.ci/test_r_package.sh
+++ b/.ci/test_r_package.sh
@@ -18,8 +18,8 @@ if [[ "${R_MAJOR_VERSION}" == "3" ]]; then
     export R_LINUX_VERSION="3.6.3-1bionic"
     export R_APT_REPO="bionic-cran35/"
 elif [[ "${R_MAJOR_VERSION}" == "4" ]]; then
-    export R_MAC_VERSION=4.0.2
-    export R_LINUX_VERSION="4.0.2-1.1804.0"
+    export R_MAC_VERSION=4.0.3
+    export R_LINUX_VERSION="4.0.3-1.1804.0"
     export R_APT_REPO="bionic-cran40/"
 else
     echo "Unrecognized R version: ${R_VERSION}"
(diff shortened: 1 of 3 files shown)


StrikerRUS (approved):
Thank you!

StrikerRUS:
Seems that it is mote complicated:

```
* checking whether package 'lightgbm' can be installed ... WARNING
  Warning: package 'R6' was built under R version 4.0.3
```

jameslamb:
> Seems that it is more complicated:
> 
> ```
> * checking whether package 'lightgbm' can be installed ... WARNING
> Found the following significant warnings:
>   Warning: package 'R6' was built under R version 4.0.3
> ```

maybe I can turn off this specific warning with an  environment variable. That would be better than allowing up to 1 warning in CI. I'll investigate

StrikerRUS:
> That would be better than allowing up to 1 warning in CI.

Yes, definitely!

jameslamb:
it looks like we can not skip this check 😢 

I don't see any environment variables mentioned in the relevant code for `R CMD check`:

https://github.com/wch/r-source/blob/157225dfb95ea6eee615f2c34d677a0abf2015f8/src/library/base/R/library.R#L119-L150

I also tried to look for all of the environment variables that control `R CMD CHECK`. I'm not aware of a comprehensive list of these, but I found them by searching for `Sys.getenv()`:

https://github.com/wch/r-source/search?q=Sys.getenv

I'm going to ask the `r-package-devel` mailing list. But first, to unblock our CI, in this PR I'll upgrade us to R 4.0.3. It is often 2-3 months between version upgrades (https://cran.r-project.org/bin/windows/base/old/), so that will buy us some time.

jameslamb:
ok I've updated the versions to R 4.0.3

StrikerRUS (approved):

PR #3477, opened 2020-10-22
[ci] update SWIG installation on macOS

StrikerRUS:
Howbrew has recently removed the possibility to install old versions of packages from GitHub commits.
```
Error: Calling Installation of swig from a GitHub commit URL is disabled! Use 'brew extract swig' to stable tap on GitHub instead.
```
Now one should maintain their own GitHub repo for that.
More reading:
https://github.com/Homebrew/brew/issues/8791
https://discourse.brew.sh/t/has-brew-install-force-formula-raw-path-been-taken-out/8793

But fortunately I've found `SWIG@3` official formulae. I remember I hadn't got a luck to find it when setup SWIG building pipeline.

@imatiach-msft Just wonder, can we eventually migrate to SWIG 4? How about backward compatibility in this case?

### DIFF
--- a/.ci/setup.sh
+++ b/.ci/setup.sh
@@ -15,7 +15,7 @@ if [[ $OS_NAME == "macos" ]]; then
         brew install open-mpi
     fi
     if [[ $AZURE == "true" ]] && [[ $TASK == "sdist" ]]; then
-        brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/f3544543a3115023fc7ca962c21d14b443f419d0/Formula/swig.rb  # swig 3.0.12
+        brew install swig@3
     fi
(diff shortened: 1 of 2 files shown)


jameslamb (approved):
thanks for the fix!

imatiach-msft:
@StrikerRUS yes we should migrate to SWIG4 eventually.  I don't see any backwards compatibility issue, assuming no one is depending on the SWIG generated code directly, which has had breaking changes in the past already.

imatiach-msft (approved):
LGTM!

StrikerRUS:
@imatiach-msft Super! I believe that MMLSpark in the main and probably the only consumer of our `jar` artifacts. Other developers  build SWIG wrappers by their own.

PR #3478, opened 2020-10-23
remove std::move

StrikerRUS:
Fix warnings from Clang compilation: https://github.com/microsoft/LightGBM/pull/3443#discussion_r507728708.

### DIFF
--- a/src/boosting/gbdt_model_text.cpp
+++ b/src/boosting/gbdt_model_text.cpp
@@ -393,7 +393,7 @@ std::string GBDT::SaveModelToString(int start_iteration, int num_iteration, int 
     ss << loaded_parameter_ << "\n";
     ss << "end of parameters" << '\n';
   }
-  return std::move(ss.str());
+  return ss.str();
 }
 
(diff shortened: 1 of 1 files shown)


src/boosting/gbdt_model_text.cpp
```diff
@@ -393,7 +393,7 @@ std::string GBDT::SaveModelToString(int start_iteration, int num_iteration, int
     ss << loaded_parameter_ << "\n";
     ss << "end of parameters" << '\n';
   }
-  return std::move(ss.str());
+  return ss.str();
```
StrikerRUS, line 396: /gha run-valgrind

StrikerRUS, line 396: /gha run r-valgrind

StrikerRUS: Valgrind is happy: https://github.com/microsoft/LightGBM/actions/runs/323142426.

guolinke (approved):

src/boosting/gbdt_model_text.cpp
jameslamb: if valgrind is happy, I am happy 😀

PR #3486, opened 2020-10-26
[ci] [python] reduce unnecessary data loading in tests

jameslamb:
I spent some time this weekend learning [`snakeviz`](https://jiffyclub.github.io/snakeviz/), a Python profiler. I tried running it over the Python package tests here, and I think I found an opportunity to cut the test run time.

I realized we're spending time loading and reloading the same datasets.  This PR proposes reducing those calls by caching the results of `sklearn.datasets.load_*` calls. The datasets are really small so I don't think holding them all in memory will cause an issue.

**UPDATE:** Ok from some experiments, the speedup from this seems really small on my laptop. But might be larger in CI environments, where we know that disk I/O is a lot slower (https://github.com/microsoft/LightGBM/pull/2965#issuecomment-624421092).

## Results

These results were obtained on my laptop.  The blockquote results come from one run of each test. Then I ran them all again 10 times to get a better estimate of the mean time.

### `test_basic.py`

**before (mean = 1.33s)**

> 58216 function calls (57604 primitive calls) in 0.462 seconds
    ==== 10 passed, 3 warnings in 2.18s ====

* trials (seconds):  [1.37, 1.40, 1.34, 1.37, 1.31, 1.30, 1.31, 1.31, 1.31, 1.31]

**after (mean = 1.36s)**

> 53470 function calls (52858 primitive calls) in 0.396 seconds
    ==== 10 passed, 3 warnings in 1.36s ====

* trials (seconds):  [1.39, 1.39, 1.40, 1.36, 1.35, 1.41, 1.40, 1.29, 1.31, 1.35]

### `test_engine.py`

**before (mean = 12.78s)**

> 2749043 function calls (2651956 primitive calls) in 9.456 seconds
    ==== 63 passed, 2 skipped, 20 warnings in 13.69s ====

* trials (seconds): [14.16, 13.13, 12.85, 12.66, 12.34, 12.60, 13.17, 12.98, 12.40, 11.50]

**after (mean = 11.84s)**

> 2749043 function calls (2651956 primitive calls) in 9.456 seconds
    ==== 63 passed, 2 skipped, 20 warnings in 11.29s ====

* trials (seconds): [12.18, 10.71, 11.09, 10.88, 10.77, 12.64, 12.47, 14.90, 11.62, 11.14]

### `test_sklearn.py`

**before(mean = 9.68s)**

> 4292703 function calls (4172793 primitive calls) in 8.044 seconds
    ==== 34 passed, 15 warnings in 11.07s ====

* trials (seconds): [10.17, 9.52, 9.34, 9.58, 9.22, 9.34, 9.75, 10.21, 9.67, 9.99]

**after (mean = 9.50s)**

> 2945200 function calls (2843392 primitive calls) in 8.867 seconds
    ==== 34 passed, 15 warnings in 9.26s ====

* trials (seconds): [9.67, 8.64, 9.32, 8.71, 8.73, 9.29, 8.97, 12.73, 9.48, 9.44]

## How to reproduce these tests

```shell
# install lightgbm
pushd python-package
python setup.py install
popd

# install dependencies
pip install snakeviz pytest-profiling

# profile tests
pytest --profile tests/python_package_test/test_basic.py
pytest --profile tests/python_package_test/test_engine.py
pytest --profile tests/python_package_test/test_sklearn.py

# (optional) visualize profiling data
snakeviz prof/combined.prof
```

### DIFF
--- a/tests/python_package_test/test_basic.py
+++ b/tests/python_package_test/test_basic.py
@@ -3,6 +3,8 @@ import os
 import tempfile
 import unittest
 
+from functools import lru_cache
+
 import lightgbm as lgb
 import numpy as np
 
@@ -11,10 +13,15 @@ from sklearn.datasets import load_breast_cancer, dump_svmlight_file, load_svmlig
 from sklearn.model_selection import train_test_split
 
 
+@lru_cache
+def _load_breast_cancer(**kwargs):
+    return load_breast_cancer(**kwargs)
+
+
 class TestBasic(unittest.TestCase):
 
     def test(self):
-        X_train, X_test, y_train, y_test = train_test_split(*load_breast_cancer(return_X_y=True),
+        X_train, X_test, y_train, y_test = train_test_split(*_load_breast_cancer(return_X_y=True),
                                                             test_size=0.1, random_state=2)
         train_data = lgb.Dataset(X_train, label=y_train)
         valid_data = train_data.create_valid(X_test, label=y_test)
@@ -86,7 +93,7 @@ class TestBasic(unittest.TestCase):
         os.remove(tname)
 
     def test_chunked_dataset(self):
-        X_train, X_test, y_train, y_test = train_test_split(*load_breast_cancer(return_X_y=True), test_size=0.1, random_state=2)
+        X_train, X_test, y_train, y_test = train_test_split(*_load_breast_cancer(return_X_y=True), test_size=0.1, random_state=2)
 
         chunk_size = X_train.shape[0] // 10 + 1
         X_train = [X_train[i * chunk_size:(i + 1) * chunk_size, :] for i in range(X_train.shape[0] // chunk_size + 1)]
@@ -273,7 +280,7 @@ class TestBasic(unittest.TestCase):
             self.assertAlmostEqual(data.label[1], data.weight[1])
             self.assertListEqual(data.feature_name, data.get_feature_name())
 
-        X, y = load_breast_cancer(return_X_y=True)
+        X, y = _load_breast_cancer(return_X_y=True)
         sequence = np.ones(y.shape[0])
         sequence[0] = np.nan
         sequence[1] = np.inf
--- a/tests/python_package_test/test_basic.py
+++ b/tests/python_package_test/test_basic.py
@@ -14,6 +14,7 @@ try:
     from functools import lru_cache
 except ImportError:
     warnings.warn("Could not import functools.lru_cache", RuntimeWarning)
+
     def lru_cache(user_function, maxsize=None):
         @wraps(user_function)
         def wrapper(*args, **kwargs):
--- a/tests/python_package_test/test_basic.py
+++ b/tests/python_package_test/test_basic.py
@@ -10,19 +10,7 @@ from scipy import sparse
 from sklearn.datasets import load_breast_cancer, dump_svmlight_file, load_svmlight_file
 from sklearn.model_selection import train_test_split
 
-try:
-    from functools import lru_cache
-except ImportError:
-    warnings.warn("Could not import functools.lru_cache", RuntimeWarning)
-
-    def lru_cache(user_function, maxsize=None):
-        @wraps(user_function)
-        def wrapper(*args, **kwargs):
-            arg_key = tuple(args, [item for item in kwargs.items()])
-            if arg_key not in cache:
-                cache[arg_key] = user_function(*args)
-            return cache[arg_key]
-        return wrapper
+from .utils import lru_cache
 
 
 @lru_cache(maxsize=None)
--- a/tests/python_package_test/test_basic.py
+++ b/tests/python_package_test/test_basic.py
@@ -3,7 +3,7 @@ import os
 import tempfile
 import unittest
 
-from functools import lru_cache
+from functools import cache
 
 import lightgbm as lgb
 import numpy as np
@@ -13,7 +13,7 @@ from sklearn.datasets import load_breast_cancer, dump_svmlight_file, load_svmlig
 from sklearn.model_selection import train_test_split
 
 
-@lru_cache
+@cache
 def _load_breast_cancer(**kwargs):
     return load_breast_cancer(**kwargs)
 
--- a/tests/python_package_test/test_basic.py
+++ b/tests/python_package_test/test_basic.py
@@ -3,14 +3,25 @@ import os
 import tempfile
 import unittest
 
-from functools import lru_cache
-
 import lightgbm as lgb
 import numpy as np
 
 from scipy import sparse
 from sklearn.datasets import load_breast_cancer, dump_svmlight_file, load_svmlight_file
 from sklearn.model_selection import train_test_split
+
+try:
+    from functools import lru_cache
+except ImportError:
+    warnings.warn("Could not import functools.lru_cache", RuntimeWarning)
+    def lru_cache(user_function, maxsize=None):
+        @wraps(user_function)
+        def wrapper(*args, **kwargs):
+            arg_key = tuple(args, [item for item in kwargs.items()])
+            if arg_key not in cache:
+                cache[arg_key] = user_function(*args)
+            return cache[arg_key]
+        return wrapper
 
 
 @lru_cache(maxsize=None)
--- a/tests/python_package_test/test_basic.py
+++ b/tests/python_package_test/test_basic.py
@@ -7,21 +7,16 @@ import lightgbm as lgb
 import numpy as np
 
 from scipy import sparse
-from sklearn.datasets import load_breast_cancer, dump_svmlight_file, load_svmlight_file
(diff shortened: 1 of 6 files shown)


jameslamb:
interesting, I see this in some tests:

```text
../../../../miniconda/envs/test-env/lib/python3.6/functools.py:477: in lru_cache
    raise TypeError('Expected maxsize to be an integer or None')
E   TypeError: Expected maxsize to be an integer or None
```

Maybe there was not a default value in older versions of Python? Because today it has a default of 128: https://docs.python.org/3/library/functools.html#functools.lru_cache

Anyway, I'm going to try switching these to just `@cache`. Since they're just tests that we completely control, not user code, I think it's ok to have an unbounded cache. And that should be faster. From https://docs.python.org/3/library/functools.html#functools.cache:

> Returns the same as lru_cache(maxsize=None), creating a thin wrapper around a dictionary lookup for the function arguments. Because it never needs to evict old values, this is smaller and faster than lru_cache() with a size limit.

StrikerRUS:
@jameslamb I think we can have `maxsize` equals to `2` for functions with only one bool argument `return_X_y=True/False` and 32 for `load_digits`. `None` doesn't look as a good default value:
https://github.com/python/cpython/blob/920cb647ba23feab7987d0dac1bd63bfc2ffc4c0/Lib/functools.py#L549-L562

jameslamb:
@StrikerRUS can you explain why you think `None` is a bad default?

From that code snippet, it looks like it would be faster than setting a `maxsize` (since there is no code about evicting things from the cache).

I understand why it would be bad in user code, since the cache can grow without limit, but for these unit tests where we completely control the set of unique combinations and know it to be small, I think we should have a preference for the faster option.

StrikerRUS:
@jameslamb 
`maxsize=None` means no LRU feature. I thought you want to use it.

jameslamb:
> @jameslamb
> `maxsize=None` means no LRU feature. I thought you want to use it.

oh I see. No I really just cared more about the caching than using Least Recently Used (LRU), since we have so few variations of kwargs for each dataset.

I did forget though that `functools.lru_cache` isn't available in Python 2.7. Will push something to fix that :grimacing:

StrikerRUS:
@jameslamb Ah OK! I misunderstood you then.

tests/python_package_test/test_basic.py
```diff
@@ -10,11 +10,30 @@
...
+    warnings.warn("Could not import functools.lru_cache", RuntimeWarning)
+
+    def lru_cache(user_function, maxsize=None):
+        @wraps(user_function)
+        def wrapper(*args, **kwargs):
+            arg_key = tuple(args, [item for item in kwargs.items()])
+            if arg_key not in cache:
+                cache[arg_key] = user_function(*args)
+            return cache[arg_key]
+        return wrapper
```
StrikerRUS, line 25: @jameslamb I think you can move this code to `compat.py` and reuse it in other files from `python_package_test`.

jameslamb: I think we should avoid adding code to the library that is only used for tests.

* it can make `import lightgbm` slower for users
* it makes the library larger which can be a problem for cases like [AWS Lambda](https://docs.aws.amazon.com/lambda/latest/dg/configuration-layers.html), where there is a strict 250MB limit for the total size of all dependencies

I have an alternative suggestion...I just pushed https://github.com/microsoft/LightGBM/pull/3486/commits/25a9bc7736010eb4278583c576dc761599c0040d, which adds a `utils.py` that the tests can import from.

StrikerRUS: Yeah, for sure, it will be better!

tests/python_package_test/utils.py
```diff
@@ -0,0 +1,15 @@
+
+try:
```
StrikerRUS, line 2:
```suggestion
try:
```

tests/python_package_test/utils.py
```diff
@@ -0,0 +1,15 @@
...
+    from functools import lru_cache
+except ImportError:
+    import warnings
+    warnings.warn("Could not import functools.lru_cache", RuntimeWarning)
+
+    def lru_cache(user_function, maxsize=None):
+        @wraps(user_function)
```
StrikerRUS, line 9: You should import `wraps` I think: https://docs.python.org/2.7/library/functools.html#functools.wraps.

jameslamb: 😆 ok it's time for me to just make a python 2.7 conda env and actually test this before pushing commits

tests/python_package_test/utils.py
jameslamb: I didn't accept this change in the browser but I did make the change (removing the beginning blank line)

tests/python_package_test/utils.py
jameslamb: ok that worked! I didn't install `lightgbm` in that env, but I was able to test the data loading

```
conda create --name py27 python=2.7 scikit-learn
```

and it looks like all tests are passing now

tests/python_package_test/utils.py
```diff
@@ -0,0 +1,17 @@
+try:
```
StrikerRUS, line 1:
```suggestion
# coding: utf-8
try:
```

tests/python_package_test/utils.py
```diff
@@ -0,0 +1,17 @@
...
+except ImportError:
+    import warnings
+    warnings.warn("Could not import functools.lru_cache", RuntimeWarning)
+
+    def lru_cache(maxsize=None):
+        cache = {}
+
+        def _lru_wrapper(user_function):
+            def wrapper(*args, **kwargs):
+                arg_key = (args, tuple([item for item in kwargs.items()]))
```
StrikerRUS, line 12:
```suggestion
                arg_key = (args, tuple(kwargs.items()))
```

tests/python_package_test/utils.py
```diff
@@ -0,0 +1,17 @@
+try:
+    from functools import lru_cache
+except ImportError:
+    import warnings
+    warnings.warn("Could not import functools.lru_cache", RuntimeWarning)
+
+    def lru_cache(maxsize=None):
```
StrikerRUS, line 7: Please memoize this too
https://github.com/microsoft/LightGBM/blob/5cc9e671977f67126efe4196d6161ed5bd40aaca/tests/python_package_test/test_plotting.py#L19

tests/python_package_test/test_sklearn.py
```diff
@@ -74,10 +76,35 @@ def multi_logloss(y_true, y_pred):
...
+
+
+@lru_cache(maxsize=None)
+def _load_iris(**kwargs):
+    return load_iris(**kwargs)
+
+
+@lru_cache(maxsize=None)
+def _load_linnerud(**kwargs):
+    return load_linnerud(**kwargs)
```
StrikerRUS, line 101: Can we move all these proxies to `utils.py` with the aim to deduplicate code?

StrikerRUS (changes requested):
@jameslamb Please check my comments below.

tests/python_package_test/test_sklearn.py
jameslamb: oh sure, that's a good point!

tests/python_package_test/utils.py
jameslamb: what would be the purpose? That's the only call to `load_breast_cancer()` in that module. So I think the caching would add a tiny bit of overhead for no benefit.

StrikerRUS: How is it differs with other calls you've memoized? 5 calls of `load_breast_cancer()` in `test_plotting.py` is even more than 3 calls of the same function in `test_basic.py`, for example.

jameslamb: there is only 1 call in `test_plotting.py`

```shell
git grep load_breast_cancer tests/python_package_test/test_plotting.py
```

image

StrikerRUS: Yes, you are right, but the method in which this call is performed (`setUp`) is called before each test. So, actually we have 5 calls.

jameslamb: OOOOOOOOOOO haha ok. I haven't used `unittest.TestCase` in a while, I forgot which one was a  "run before every test" setup and which one was a "run exactly once, before any tests" one.

Ok yes I'll update this

jameslamb: added in https://github.com/microsoft/LightGBM/pull/3486/commits/dfb0fd35cdea555d8b8aac800fbb01528fbe0ca4

StrikerRUS: Thanks! Actually I think we should refactor this to "run exactly once, before any tests" (`setUpClass()`), but it is another issue, of course.

StrikerRUS (approved):
LGTM, thanks a lot!

PR #3501, opened 2020-10-29
bump minimal CMake version

StrikerRUS:
According to the following discussion: https://github.com/microsoft/LightGBM/pull/3405#discussion_r502223468.

CMake < 3 is very-very old (in terms of their frequent releases). With 3.0 it will be possible to check API docs easier at least (docs for < 3 are hiddedn deep at their site and not structured).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,7 +5,7 @@ elseif(USE_GPU OR APPLE)
 elseif(USE_CUDA)
   cmake_minimum_required(VERSION 3.16)
 else()
-  cmake_minimum_required(VERSION 2.8)
+  cmake_minimum_required(VERSION 3.0)
 endif()
 
 if(USE_CUDA)


jameslamb (commented):
I don't know enough  about the possible impact, so I don't think my review should count as a ❌ or ✅ 

But I do want to add  for other reviewers that, specifically, `cmake` 3.0.0 was released in December 2018: https://github.com/Kitware/CMake/releases/tag/v3.0.0

StrikerRUS:
@jameslamb 

> But I do want to add for other reviewers that, specifically, cmake 3.0.0 was released in December 2018: https://github.com/Kitware/CMake/releases/tag/v3.0.0

Oh, please don't be confused by that info! Actually, CMake 3.0.0 was released June 10, 2014: https://blog.kitware.com/cmake-3-0-0-available-for-download/. 

Seems the thing is that Kitware created their helper bot `kwrobot` in December 2018 and it pushed all old releases of CMake in one day from their main GitLab repo.
https://github.com/Kitware/CMake/releases?after=v3.11.0

image

image

guolinke (approved):

PR #3502, opened 2020-10-29
[DOCS] [Python] Update custom objective function docs

### LINKED ISSUE #3149: [Improvement] [Python] Improve docs relating to custom objective functions
The current docs regarding using custom objective functions with the scikit-learn API, as can be found [here](https://lightgbm.readthedocs.io/en/latest/pythonapi/lightgbm.LGBMClassifier.html), state:

> A custom objective function can be provided for the ``objective`` parameter.
> In this case, it should have the signature
> ``objective(y_true, y_pred) -> grad, hess`` or
> ``objective(y_true, y_pred, group) -> grad, hess``:
> 
>     y_true : array-like of shape = [n_samples]
>         The target values.
>     y_pred : array-like of shape = [n_samples] or shape = [n_samples * n_classes] (for multi-class task)
>         The predicted values.
>     group : array-like
>         Group/query data, used for ranking task.
>     grad : array-like of shape = [n_samples] or shape = [n_samples * n_classes] (for multi-class task)
>         The value of the first order derivative (gradient) for each sample point.
>     hess : array-like of shape = [n_samples] or shape = [n_samples * n_classes] (for multi-class task)
>         The value of the second order derivative (Hessian) for each sample point.
> 
> For binary task, the y_pred is margin.
> For multi-class task, the y_pred is group by class_id first, then group by row_id.
> If you want to get i-th row y_pred in j-th class, the access way is y_pred[j * num_data + i]
> and you should group grad and hess in this way as well.

As far as I can tell, there are two issues with this:
1. There is no notion of what the gradient is with respect to. I assume it is with respect to the outputs of the model, i.e. with respect to P(sample i belongs to class j) for classification or with respect to E[sample i] for regression.
2. We do not need to provide the hessian, but the diagonal of the hessian.

hsorsky:
Hopefully this serves to provide better documentation for using custom objective functions. 

AFAIK the logic stated here is correct, but please let me know if not - I was the one who opened this issue originally because I didn't understand the inputs, so I may still not.

Closes #3149. 

**TODO:**
- [ ] I'd also like to add an example of using a custom objective function (or even just an example of the regular objective function, but manually implemented) as part of this PR, but I don't know where is best to add it. Any suggestions?

### DIFF
--- a/python-package/lightgbm/sklearn.py
+++ b/python-package/lightgbm/sklearn.py
@@ -263,17 +263,17 @@ class LGBMModel(_LGBMModelBase):
 
             y_true : array-like of shape = [n_samples]
                 The target values.
-            y_pred : array-like of shape = ``[n_samples]`` or shape = ``[n_samples * n_classes]`` (for multi-class
-                task). The predicted values. Note: For classification tasks, ``y_pred`` is assumed to be in the
+            y_pred : array-like of shape = [n_samples] or shape = [n_samples * n_classes] (for multi-class tasks).
+                The predicted values. Note: For classification tasks, ``y_pred`` is assumed to be in the
                 pre-transformed space, i.e. pre-logistic-transform for binary classification or pre-softmax-transform
                 for multi-class classification.
             group : array-like
                 Group/query data, used for ranking task.
-            grad : array-like of shape = ``[n_samples]`` or shape = ``[n_samples * n_classes]`` (for multi-class tasks).
+            grad : array-like of shape = [n_samples] or shape = [n_samples * n_classes] (for multi-class tasks).
                 The value of the first order derivative (gradient) of the loss with respect to the elements of
                 ``y_pred`` for each sample point. E.g. For multi-class tasks, ``grad[0]`` is the derivative of the
                 loss function with respect to the pre-transformed output for sample 0, class 0.
-            hess : array-like of shape = ``[n_samples]`` or shape = ``[n_samples * n_classes]`` (for multi-class task).
+            hess : array-like of shape = [n_samples] or shape = [n_samples * n_classes] (for multi-class task).
                 The value of the second order derivative (Hessian) of the loss with respect to the elements of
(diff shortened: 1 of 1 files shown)


btrotta:
@hsorsky Thanks, this is a really useful improvement!

Re your todo, we already have an example here: https://github.com/microsoft/LightGBM/blob/e83042f20633d7f74dda0d18624721447a610c8b/examples/python-guide/advanced_example.py#L136 
But I think it would be good to add a link to this in the docs

python-package/lightgbm/sklearn.py
```diff
@@ -263,19 +263,25 @@ def __init__(self, boosting_type='gbdt', num_leaves=31, max_depth=-1,
...
+                loss function with respect to the pre-transformed output for sample 0, class 0.
+            hess : array-like of shape = [n_samples] or shape = [n_samples * n_classes] (for multi-class task).
+                The value of the second order derivative (Hessian) of the loss with respect to the elements of
+                ``y_pred`` for each sample point. E.g. ``hess[0]`` is the second derivative of the loss function
+                with respect to the pre-transformed output for sample 0, class 0.
+
+        For binary tasks, ``y_pred`` is margin.
+        For multi-class tasks, ``y_pred`` is grouped by class_id first, then by row_id.
+        I.e. if you want to get the prediction for the i-th sample in the j-th class, you would access it via
+        ``y_pred[j * num_data + i]``, and you should group ``grad`` and ``hess`` in this way as well.
```
jameslamb, line 284:
```suggestion
        ``y_pred[j * num_data + i]``, and you should group ``grad`` and ``hess`` in this way as well.
        
        For an example using a custom objective function, please see
        https://github.com/microsoft/LightGBM/blob/master/examples/python-guide/advanced_example.py
```

jameslamb (changes requested):
@hsorsky thanks for your contribution! Could you please update this branch to the latest `master`? I left one suggestion as well about where to put a link to the custom-objective example, based on @btrotta 's comment.

StrikerRUS:
Please feel free to reopen if still interested in this contribution.

Issue #3357, opened 2020-09-06
[docs] update list of winning solutions


jameslamb:
This project's documentation includes a list of winning solutions for machine learning competitions which used LightGBM.

https://github.com/microsoft/LightGBM/tree/master/examples#machine-learning-challenge-winning-solutions

With one exception (), this has not been updated since December 2019.

## How to close this issue

Look through "completed" competitions (from December 2019 to now) at https://www.kaggle.com/competitions.

![image](https://user-images.githubusercontent.com/7608904/92319025-af86a500-efd9-11ea-9183-c911085faef7.png)

For each competition, look through the message boards to find posts about the 1st, 2nd, or 3rd place solutions. If these used LightGBM (even as part of an ensemble), create a pull request which adds them to https://github.com/microsoft/LightGBM/blob/master/examples/README.md.

Juniper-23:
Hi, I can help with the above. I need a quick clarification, from my understanding the above issue requires me to check for the usage of LightGBM in the solution description/discussions of the first three winning entries?

guolinke:
@Juniper-23 
yeah, you can review the shared winning solutions in Kaggle/KDD, etc.

StrikerRUS:
Can we close this issue? I think the majority of solutions from Kaggle are covered right now.

jameslamb:
yeah I think so

PR #3503, opened 2020-10-29
[docs] added Data Science Bowl Comp Solution 2019

nabokovas:
It is the second place solution. https://github.com/microsoft/LightGBM/issues/3357


jameslamb (approved):
oh great, thanks! I think I missed this one because the description just says "LGB" and when I search before, I looked for "lightgbm" and "lgbm".

StrikerRUS (approved):
LGTM! Thanks!

guolinke (approved):

PR #3504, opened 2020-10-29
[tests][python] remove excess iterations

nabokovas:
There was deleted excess initialisation and iteration


StrikerRUS (approved):
Huh, nice catch! Thanks a lot!
I guess you were looking through GitHub history to find this inconspicuous mistake because I remember we merged this code very recently.

PR #3505, opened 2020-10-29
[ci] Update MacOS version at Travis

nabokovas:
There is a new version of The macOS Build Environment. So I propose changes for you to use a modern one. https://docs.travis-ci.com/user/reference/osx/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ os:
   - linux
   - osx
 dist: focal
-osx_image: xcode12
+osx_image: xcode12.2
 
 env:
   global:  # default values


jameslamb (changes requested):
Thanks for taking the time to contribute! But I'm -1 on this change.

We do try to test LightGBM against somewhat-reason versions of software, but we also want to be sure that the library works well on widely-used configurations.  XCode 12.2 is less than a month old (https://xcodereleases.com/), so I'd guess that many of our users who are using LightGBM on Mac haven't updated to it.

But I'll defer to @StrikerRUS 's opinion

StrikerRUS (approved):
@nabokovas Thank you very much for this update! You've just done my job 🙂 

@jameslamb Let me kindly disagree with you. According to our policy, we try to use the most recent environments at Travis (https://github.com/microsoft/LightGBM/pull/3393#issue-486186591) and update them quite often.

image


I believe it's better to spot a compilation problem at our CI before users come and report it.
Moreover, we are speaking about macOS where it seems users tend to use the last available software in contrast to Linux where one have to use 10-years old CentOS distribution.

FYI, yesterday I tried to use Big Sur, but unfortunately without any luck due to dependency unavailability: #3499.

jameslamb:
ah I see, ok then I am ok with this change.

jameslamb (approved):

PR #3514, opened 2020-11-01
[WIP][DO NOT MERGE] Test workflows

StrikerRUS:


.ci/get_workflow_status.py
```diff
@@ -0,0 +1,54 @@
+import json
```
StrikerRUS, line 1: /gha run

StrikerRUS: /gha run

PR #3484, opened 2020-10-26
3.1.0 release

guolinke:

Issue #2791, opened 2020-02-21
[Discussion] efficiency improvements


guolinke:
This is to call the efficiency improvement for LightGBM, include but not limited to:
- Tree learning algorithm acceleration
- I/O related, like dataset loading and model saving
- Inference speed improvement
- Better GPU acceleration
- Distributed learning improvement

If you have any ideas, you can discuss with us here, and open the corresponding issues/pull requests if needed.

StrikerRUS:
Linking #2934 and #2935.

MaxHalford:
I have a suggestion but I'm not sure it's in-scope. It concerns improving efficiency during the prediction speed of trees. This in turn could improve the overall training time because gradient boosting requires asking the weak models to predict gradients.

I'm not familiar with the tree implementation in LightGBM. The intuitive way of having two pointers that refer to the left and right childs doesn't seem to be the best format for evaluating a tree. Basically, it seems that there are two alternative ways:

1. Compile the tree to a sequence of `if`s and `else`s. AFAIK this is what [treelite](https://github.com/dmlc/treelite) allows you to do.
2. Convert a tree into a heap, which means the nodes are contiguous in memory. Each branch contains two integers that indicate where to go next in the heap. This is quite well explained in [this](https://www.wikiwand.com/en/Binary_heap#/Heap_implementation) Wikipedia article section. 

I thought about suggesting this after having read [this great blog post](http://tullo.ch/articles/decision-tree-evaluation/) by Andrew Tulloch. I don't know if this is something LightGBM already integrates. Also I'm not well-versed in C++ so I can't help on that side. However these solutions seem rather simple to implement.

julioasotodv:
Apart from the possibility to use CUDA instead of OpenCL for GPU trees (even though it would reduce compatibility), xgboost has made a lot of progress using Dask for distributed training (in fact, I believe that other experimental distributed training techniques they had are now deprecated).

I believe someone started a project fro smoother integration between LightGBM and Dask, but I am unable to find it right now...

As for GPU, I think that guys at Rapids (Nvidia) are looking forward to developing more CUDA implementations for different ML algorithms (just like they have done with xgboost, as they have rewritten the whole gpu tree learner in the library). Maybe contacting them could be a good idea.

jameslamb:
`dask-lightgbm` is here: https://github.com/dask/dask-lightgbm

It's not something we officially maintain or contribute to yet, but I agree it could be a useful thing for us to help with.

[...]

PR #3515, opened 2020-11-01
[python] [dask] add initial dask integration

SfinxCZ:
As agreed with @jameslamb in https://github.com/dask/community/issues/104 this pull request moves all functionality from https://github.com/dask/dask-lightgbm repo directly into main LightGBM repo. 

Note that this is WIP and most of the code should be changed to match LightGBM standards.

### DIFF
--- a/.ci/test.sh
+++ b/.ci/test.sh
@@ -70,7 +70,7 @@ if [[ $TASK == "if-else" ]]; then
     exit 0
 fi
 
-conda install -q -y -n $CONDA_ENV joblib matplotlib numpy pandas psutil pytest python-graphviz scikit-learn scipy dask distributed dask-ml
+conda install -q -y -n $CONDA_ENV dask dask-ml distributed joblib matplotlib numpy pandas psutil pytest python-graphviz scikit-learn scipy
 
 if [[ $OS_NAME == "macos" ]] && [[ $COMPILER == "clang" ]]; then
     # fix "OMP: Error #15: Initializing libiomp5.dylib, but found libomp.dylib already initialized." (OpenMP library conflict due to conda's MKL)
--- a/.ci/test.sh
+++ b/.ci/test.sh
@@ -70,7 +70,7 @@ if [[ $TASK == "if-else" ]]; then
     exit 0
 fi
 
-conda install -q -y -n $CONDA_ENV joblib matplotlib numpy pandas psutil pytest python-graphviz scikit-learn scipy
+conda install -q -y -n $CONDA_ENV joblib matplotlib numpy pandas psutil pytest python-graphviz scikit-learn scipy dask distributed dask-ml
 
 if [[ $OS_NAME == "macos" ]] && [[ $COMPILER == "clang" ]]; then
     # fix "OMP: Error #15: Initializing libiomp5.dylib, but found libomp.dylib already initialized." (OpenMP library conflict due to conda's MKL)
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -30,6 +30,10 @@ R-package/    @Laurae2 @jameslamb
 # Python code
 python-package/    @StrikerRUS @chivee @wxchan @henry0312
 
+# Dask integration
+python-package/lightgbm/dask_distributed.py    @jameslamb
+tests/python_package_test/test_dask_distributed.py    @jameslamb
+
 # helpers
 helpers/    @StrikerRUS @guolinke
 
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -31,8 +31,8 @@ R-package/    @Laurae2 @jameslamb
 python-package/    @StrikerRUS @chivee @wxchan @henry0312
 
 # Dask integration
-python-package/lightgbm/dask_distributed.py    @jameslamb
-tests/python_package_test/test_dask_distributed.py    @jameslamb
+python-package/lightgbm/dask.py    @jameslamb
+tests/python_package_test/test_dask.py    @jameslamb
 
 # helpers
 helpers/    @StrikerRUS @guolinke
--- a/python-package/lightgbm/dask.py
+++ b/python-package/lightgbm/dask.py
@@ -40,7 +40,7 @@ def _build_network_params(worker_addresses, local_worker_ip, local_listen_port, 
     worker_addresses : iterable of str - collection of worker addresses in `<protocol>://<host>:port` format
     local_worker_ip : str
     local_listen_port : int
-    listen_time_out : int
+    time_out : int
 
     Returns
     -------
@@ -206,6 +206,24 @@ def _predict(model, data, proba=False, dtype=np.float32, **kwargs):
 
 class _LGBMModel:
 
+    def _fit(self, model_factory, X, y=None, sample_weight=None, client=None, **kwargs):
+        """Docstring is inherited from the LGBMModel."""
+        if client is None:
+            client = default_client()
+
+        params = self.get_params(True)
+        model = _train(client, X, y, params, model_factory, sample_weight, **kwargs)
+
+        self.set_params(**model.get_params())
+        self._copy_extra_params(model, self)
+
+        return self
+
+    def _to_local(self, model_factory):
+        model = model_factory(**self.get_params())
+        self._copy_extra_params(self, model)
+        return model
+
     @staticmethod
     def _copy_extra_params(source, dest):
         params = source.get_params()
@@ -220,17 +238,7 @@ class LGBMClassifier(_LGBMModel, LocalLGBMClassifier):
 
     def fit(self, X, y=None, sample_weight=None, client=None, **kwargs):
         """Docstring is inherited from the LGBMModel."""
-        if client is None:
-            client = default_client()
-
-        model_factory = LocalLGBMClassifier
-        params = self.get_params(True)
-        model = _train(client, X, y, params, model_factory, sample_weight, **kwargs)
-
-        self.set_params(**model.get_params())
-        self._copy_extra_params(model, self)
-
-        return self
+        return self._fit(LocalLGBMClassifier, X, y, sample_weight, client, **kwargs)
     fit.__doc__ = LocalLGBMClassifier.fit.__doc__
 
     def predict(self, X, **kwargs):
@@ -250,9 +258,7 @@ class LGBMClassifier(_LGBMModel, LocalLGBMClassifier):
         -------
         model : lightgbm.LGBMClassifier
         """
-        model = LocalLGBMClassifier(**self.get_params())
-        self._copy_extra_params(self, model)
-        return model
+        return self._to_local(LocalLGBMClassifier)
 
 
 class LGBMRegressor(_LGBMModel, LocalLGBMRegressor):
@@ -260,17 +266,7 @@ class LGBMRegressor(_LGBMModel, LocalLGBMRegressor):
 
     def fit(self, X, y=None, sample_weight=None, client=None, **kwargs):
         """Docstring is inherited from the lightgbm.LGBMRegressor.fit."""
-        if client is None:
-            client = default_client()
-
-        model_factory = LocalLGBMRegressor
-        params = self.get_params(True)
-        model = _train(client, X, y, params, model_factory, sample_weight, **kwargs)
-
-        self.set_params(**model.get_params())
-        self._copy_extra_params(model, self)
-
-        return self
+        return self._fit(LocalLGBMRegressor, X, y, sample_weight, client, **kwargs)
     fit.__doc__ = LocalLGBMRegressor.fit.__doc__
 
     def predict(self, X, **kwargs):
@@ -285,6 +281,4 @@ class LGBMRegressor(_LGBMModel, LocalLGBMRegressor):
         -------
         model : lightgbm.LGBMRegressor
         """
-        model = LocalLGBMRegressor(**self.get_params())
-        self._copy_extra_params(self, model)
-        return model
+        return self._to_local(LocalLGBMRegressor)
--- a/python-package/lightgbm/dask.py
+++ b/python-package/lightgbm/dask.py
@@ -18,7 +18,6 @@ from toolz import assoc, first
 import lightgbm
 from .basic import _LIB, _safe_call
 from .sklearn import LGBMClassifier as LocalLGBMClassifier, LGBMRegressor as LocalLGBMRegressor
-from .compat import _LGBMModelBase
 
 try:
     import scipy.sparse as ss
@@ -184,7 +183,6 @@ def _predict(model, data, proba=False, dtype=np.float32, **kwargs):
 
     Parameters
     ----------
-    client: dask.Client - client
     model :
     data : dask array of shape = [n_samples, n_features]
         Input feature matrix.
@@ -206,34 +204,7 @@ def _predict(model, data, proba=False, dtype=np.float32, **kwargs):
         raise TypeError('Data must be either Dask array or dataframe. Got %s.' % str(type(data)))
 
 
-class _LGBMModel(_LGBMModelBase):
-
-    def __init__(self, model_factory, client=None) -> None:
-        if client is None:
-            client = default_client()
-        self.client = client
-        self.model_factory = model_factory
-
-    def fit(self, X, y=None, sample_weight=None, **kwargs):
-        """Docstring is inherited from the appropriate model_factory."""
-        params = self.get_params(True)
-        model = _train(self.client, X, y, params, self.model_factory, sample_weight, **kwargs)
-
-        self.set_params(**model.get_params())
-        self._copy_extra_params(model, self)
-
-        return self
-
-    def to_local(self):
-        """Create regular version of lightgbm.LGBMRegressor from the distributed version.
-
-        Returns
-        -------
-        model : lightgbm.LGBMRegressor
-        """
-        model = self.model_factory(**self.get_params())
-        self._copy_extra_params(self, model)
-        return model
+class _LGBMModel:
 
     @staticmethod
     def _copy_extra_params(source, dest):
@@ -247,11 +218,20 @@ class _LGBMModel(_LGBMModelBase):
 class LGBMClassifier(_LGBMModel, LocalLGBMClassifier):
     """Distributed version of lightgbm.LGBMClassifier."""
 
-    def __init__(self, client=None, **kwargs) -> None:
-        super().__init__(LocalLGBMClassifier, client)
-        super(_LGBMModel, self).__init__(**kwargs)
+    def fit(self, X, y=None, sample_weight=None, client=None, **kwargs):
+        """Docstring is inherited from the LGBMModel."""
+        if client is None:
+            client = default_client()
 
-    _LGBMModel.fit.__doc__ = LocalLGBMClassifier.fit.__doc__
+        model_factory = LocalLGBMClassifier
+        params = self.get_params(True)
+        model = _train(client, X, y, params, model_factory, sample_weight, **kwargs)
+
+        self.set_params(**model.get_params())
+        self._copy_extra_params(model, self)
+
+        return self
+    fit.__doc__ = LocalLGBMClassifier.fit.__doc__
 
     def predict(self, X, **kwargs):
         """Docstring is inherited from the lightgbm.LGBMClassifier.predict."""
@@ -263,17 +243,48 @@ class LGBMClassifier(_LGBMModel, LocalLGBMClassifier):
         return _predict(self.to_local(), X, proba=True, **kwargs)
     predict_proba.__doc__ = LocalLGBMClassifier.predict_proba.__doc__
 
+    def to_local(self):
+        """Create regular version of lightgbm.LGBMClassifier from the distributed version.
+
+        Returns
+        -------
+        model : lightgbm.LGBMClassifier
+        """
+        model = LocalLGBMClassifier(**self.get_params())
+        self._copy_extra_params(self, model)
+        return model
+
 
 class LGBMRegressor(_LGBMModel, LocalLGBMRegressor):
     """Docstring is inherited from the lightgbm.LGBMRegressor."""
 
-    def __init__(self, client=None, **kwargs) -> None:
-        super().__init__(LocalLGBMRegressor, client)
-        super(LocalLGBMRegressor, self).__init__(**kwargs)
+    def fit(self, X, y=None, sample_weight=None, client=None, **kwargs):
+        """Docstring is inherited from the lightgbm.LGBMRegressor.fit."""
+        if client is None:
+            client = default_client()
 
-    _LGBMModel.fit.__doc__ = LocalLGBMRegressor.fit.__doc__
+        model_factory = LocalLGBMRegressor
+        params = self.get_params(True)
+        model = _train(client, X, y, params, model_factory, sample_weight, **kwargs)
+
+        self.set_params(**model.get_params())
+        self._copy_extra_params(model, self)
+
+        return self
+    fit.__doc__ = LocalLGBMRegressor.fit.__doc__
 
     def predict(self, X, **kwargs):
         """Docstring is inherited from the lightgbm.LGBMRegressor.predict."""
         return _predict(self.to_local(), X, **kwargs)
     predict.__doc__ = LocalLGBMRegressor.predict.__doc__
+
+    def to_local(self):
+        """Create regular version of lightgbm.LGBMRegressor from the distributed version.
+
+        Returns
+        -------
+        model : lightgbm.LGBMRegressor
+        """
+        model = LocalLGBMRegressor(**self.get_params())
+        self._copy_extra_params(self, model)
+        return model
--- a/python-package/lightgbm/dask.py
+++ b/python-package/lightgbm/dask.py
@@ -1,0 +1,299 @@
+"""Distributed training with LightGBM and Dask.distributed.
+
+This module enables you to perform distributed training with LightGBM on Dask.Array and Dask.DataFrame collections.
+It is based on dask-xgboost package.
+"""
+import logging
+from collections import defaultdict
+
+try:
+    from urllib.parse import urlparse
+except ImportError:
+    from urlparse import urlparse
+
+import dask.array as da
+import dask.dataframe as dd
+import lightgbm
+import numpy as np
+import pandas as pd
+from dask import delayed
+from dask.distributed import wait, default_client, get_worker
+from lightgbm.basic import _safe_call, _LIB
+from toolz import first, assoc
+
+try:
+    import scipy.sparse as ss
+except ImportError:
+    ss = False
+
+logger = logging.getLogger(__name__)
+
+
+def _parse_host_port(address):
+    parsed = urlparse(address)
+    return parsed.hostname, parsed.port
+
+
+def build_network_params(worker_addresses, local_worker_ip, local_listen_port, time_out):
+    """Build network parameters suiltable for LightGBM C backend.
+
+    Parameters
+    ----------
+    worker_addresses : iterable of str - collection of worker addresses in `<protocol>://<host>:port` format
+    local_worker_ip : str
+    local_listen_port : int
+    listen_time_out : int
+
+    Returns
+    -------
+    params: dict
+    """
+    addr_port_map = {addr: (local_listen_port + i) for i, addr in enumerate(worker_addresses)}
+    params = {
+        'machines': ','.join('%s:%d' % (_parse_host_port(addr)[0], port) for addr, port in addr_port_map.items()),
+        'local_listen_port': addr_port_map[local_worker_ip],
+        'time_out': time_out,
+        'num_machines': len(addr_port_map)
+    }
+    return params
+
+
+def _concat(seq):
+    if isinstance(seq[0], np.ndarray):
+        return np.concatenate(seq, axis=0)
+    elif isinstance(seq[0], (pd.DataFrame, pd.Series)):
+        return pd.concat(seq, axis=0)
+    elif ss and isinstance(seq[0], ss.spmatrix):
+        return ss.vstack(seq, format='csr')
+    else:
+        raise TypeError('Data must be one of: numpy arrays, pandas dataframes, sparse matrices (from scipy). Got %s.' % str(type(seq[0])))
+
+
+def _train_part(params, model_factory, list_of_parts, worker_addresses, return_model, local_listen_port=12400,
+                time_out=120, **kwargs):
+    network_params = build_network_params(worker_addresses, get_worker().address, local_listen_port, time_out)
+    params.update(network_params)
+
+    # Concatenate many parts into one
+    parts = tuple(zip(*list_of_parts))
+    data = _concat(parts[0])
+    label = _concat(parts[1])
+    weight = _concat(parts[2]) if len(parts) == 3 else None
+
+    try:
+        model = model_factory(**params)
+        model.fit(data, label, sample_weight=weight, **kwargs)
+    finally:
+        _safe_call(_LIB.LGBM_NetworkFree())
+
+    return model if return_model else None
+
+
+def _split_to_parts(data, is_matrix):
+    parts = data.to_delayed()
+    if isinstance(parts, np.ndarray):
+        assert (parts.shape[1] == 1) if is_matrix else (parts.ndim == 1 or parts.shape[1] == 1)
+        parts = parts.flatten().tolist()
+    return parts
+
+
+def train(client, data, label, params, model_factory, weight=None, **kwargs):
+    """Inner train routine.
+
+    Parameters
+    ----------
+    client: dask.Client - client
+    X : dask array of shape = [n_samples, n_features]
+        Input feature matrix.
+    y : dask array of shape = [n_samples]
+        The target values (class labels in classification, real numbers in regression).
+    params : dict
+    model_factory : lightgbm.LGBMClassifier or lightgbm.LGBMRegressor class
+    sample_weight : array-like of shape = [n_samples] or None, optional (default=None)
+            Weights of training data.
+    """
+    # Split arrays/dataframes into parts. Arrange parts into tuples to enforce co-locality
+    data_parts = _split_to_parts(data, is_matrix=True)
+    label_parts = _split_to_parts(label, is_matrix=False)
+    if weight is None:
+        parts = list(map(delayed, zip(data_parts, label_parts)))
+    else:
+        weight_parts = _split_to_parts(weight, is_matrix=False)
+        parts = list(map(delayed, zip(data_parts, label_parts, weight_parts)))
+
+    # Start computation in the background
+    parts = client.compute(parts)
+    wait(parts)
+
+    for part in parts:
+        if part.status == 'error':
+            return part  # trigger error locally
+
+    # Find locations of all parts and map them to particular Dask workers
+    key_to_part_dict = dict([(part.key, part) for part in parts])
+    who_has = client.who_has(parts)
+    worker_map = defaultdict(list)
+    for key, workers in who_has.items():
+        worker_map[first(workers)].append(key_to_part_dict[key])
+
+    master_worker = first(worker_map)
+    worker_ncores = client.ncores()
+
+    if 'tree_learner' not in params or params['tree_learner'].lower() not in {'data', 'feature', 'voting'}:
+        logger.warning('Parameter tree_learner not set or set to incorrect value '
+                       '(%s), using "data" as default', params.get("tree_learner", None))
+        params['tree_learner'] = 'data'
+
+    # Tell each worker to train on the parts that it has locally
+    futures_classifiers = [client.submit(_train_part,
+                                         model_factory=model_factory,
+                                         params=assoc(params, 'num_threads', worker_ncores[worker]),
+                                         list_of_parts=list_of_parts,
+                                         worker_addresses=list(worker_map.keys()),
+                                         local_listen_port=params.get('local_listen_port', 12400),
+                                         time_out=params.get('time_out', 120),
+                                         return_model=(worker == master_worker),
+                                         **kwargs)
+                           for worker, list_of_parts in worker_map.items()]
+
+    results = client.gather(futures_classifiers)
+    results = [v for v in results if v]
+    return results[0]
+
+
+def _predict_part(part, model, proba, **kwargs):
+    data = part.values if isinstance(part, pd.DataFrame) else part
+
+    if data.shape[0] == 0:
+        result = np.array([])
+    elif proba:
+        result = model.predict_proba(data, **kwargs)
+    else:
+        result = model.predict(data, **kwargs)
+
+    if isinstance(part, pd.DataFrame):
+        if proba:
+            result = pd.DataFrame(result, index=part.index)
+        else:
+            result = pd.Series(result, index=part.index, name='predictions')
+
+    return result
+
+
+def predict(client, model, data, proba=False, dtype=np.float32, **kwargs):
+    """Inner predict routine.
+
+    Parameters
+    ----------
+    client: dask.Client - client
+    model :
+    data : dask array of shape = [n_samples, n_features]
+        Input feature matrix.
+    proba : bool
+        Should method return results of predict_proba (proba == True) or predict (proba == False)
+    dtype : np.dtype
+        Dtype of the output
+    kwargs : other parameters passed to predict or predict_proba method
+    """
+    if isinstance(data, dd._Frame):
+        return data.map_partitions(_predict_part, model=model, proba=proba, **kwargs).values
+    elif isinstance(data, da.Array):
+        if proba:
+            kwargs['chunks'] = (data.chunks[0], (model.n_classes_,))
+        else:
+            kwargs['drop_axis'] = 1
+        return data.map_blocks(_predict_part, model=model, proba=proba, dtype=dtype, **kwargs)
+    else:
+        raise TypeError('Data must be either Dask array or dataframe. Got %s.' % str(type(data)))
+
+
+class _LGBMModel:
+
+    @staticmethod
+    def _copy_extra_params(source, dest):
+        params = source.get_params()
+        attributes = source.__dict__
+        extra_param_names = set(attributes.keys()).difference(params.keys())
+        for name in extra_param_names:
+            setattr(dest, name, attributes[name])
+
+
+class LGBMClassifier(_LGBMModel, lightgbm.LGBMClassifier):
+    """Distributed version of lightgbm.LGBMClassifier."""
+
+    def fit(self, X, y=None, sample_weight=None, client=None, **kwargs):
+        """Docstring is inherited from the LGBMModel."""
+        if client is None:
+            client = default_client()
+
+        model_factory = lightgbm.LGBMClassifier
+        params = self.get_params(True)
+        model = train(client, X, y, params, model_factory, sample_weight, **kwargs)
+
+        self.set_params(**model.get_params())
+        self._copy_extra_params(model, self)
+
+        return self
+    fit.__doc__ = lightgbm.LGBMClassifier.fit.__doc__
+
+    def predict(self, X, client=None, **kwargs):
+        """Docstring is inherited from the lightgbm.LGBMClassifier.predict."""
+        if client is None:
+            client = default_client()
+        return predict(client, self.to_local(), X, dtype=self.classes_.dtype, **kwargs)
+    predict.__doc__ = lightgbm.LGBMClassifier.predict.__doc__
+
+    def predict_proba(self, X, client=None, **kwargs):
+        """Docstring is inherited from the lightgbm.LGBMClassifier.predict_proba."""
+        if client is None:
+            client = default_client()
+        return predict(client, self.to_local(), X, proba=True, **kwargs)
+    predict_proba.__doc__ = lightgbm.LGBMClassifier.predict_proba.__doc__
+
+    def to_local(self):
+        """Create regular version of lightgbm.LGBMClassifier from the distributed version.
+
+        Returns
+        -------
+        model : lightgbm.LGBMClassifier
+        """
+        model = lightgbm.LGBMClassifier(**self.get_params())
+        self._copy_extra_params(self, model)
+        return model
+
+
+class LGBMRegressor(_LGBMModel, lightgbm.LGBMRegressor):
+    """Docstring is inherited from the lightgbm.LGBMRegressor."""
+
+    def fit(self, X, y=None, sample_weight=None, client=None, **kwargs):
+        """Docstring is inherited from the lightgbm.LGBMRegressor.fit."""
+        if client is None:
+            client = default_client()
+
+        model_factory = lightgbm.LGBMRegressor
+        params = self.get_params(True)
+        model = train(client, X, y, params, model_factory, sample_weight, **kwargs)
+
+        self.set_params(**model.get_params())
+        self._copy_extra_params(model, self)
+
+        return self
+    fit.__doc__ = lightgbm.LGBMRegressor.fit.__doc__
+
+    def predict(self, X, client=None, **kwargs):
+        """Docstring is inherited from the lightgbm.LGBMRegressor.predict."""
+        if client is None:
+            client = default_client()
+        return predict(client, self.to_local(), X, **kwargs)
+    predict.__doc__ = lightgbm.LGBMRegressor.predict.__doc__
+
+    def to_local(self):
+        """Create regular version of lightgbm.LGBMRegressor from the distributed version.
+
+        Returns
+        -------
+        model : lightgbm.LGBMRegressor
+        """
+        model = lightgbm.LGBMRegressor(**self.get_params())
+        self._copy_extra_params(self, model)
+        return model
--- a/python-package/lightgbm/dask.py
+++ b/python-package/lightgbm/dask.py
@@ -15,7 +15,6 @@ from dask import delayed
 from dask.distributed import default_client, get_worker, wait
 from toolz import assoc, first
 
-import lightgbm
 from .basic import _LIB, _safe_call
 from .sklearn import LGBMClassifier as LocalLGBMClassifier, LGBMRegressor as LocalLGBMRegressor
 
--- a/python-package/lightgbm/dask.py
+++ b/python-package/lightgbm/dask.py
@@ -16,7 +16,7 @@ from dask import delayed
 from dask.distributed import default_client, get_worker, wait
 
 from .basic import _LIB, _safe_call
-from .sklearn import LGBMClassifier as LocalLGBMClassifier, LGBMRegressor as LocalLGBMRegressor
+from .sklearn import LGBMClassifier, LGBMRegressor
 
 import scipy.sparse as ss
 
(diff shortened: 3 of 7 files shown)


python-package/setup.py
```diff
@@ -292,6 +292,13 @@ def run(self):
               'scipy',
               'scikit-learn!=0.22.0'
           ],
+          extras_requires={
+              'dask': [
+                'dask>=0.16.0',
+                'distributed>=1.15.2'
+
+              ],
+          },
```
jameslamb, line 301: I agree with this, making `dask` an optional dependency that people install separately or with `pip install lightgbm[dask]`. I definitely don't want to force all existing `lightgbm` users to install `dask`.

I think this section should also include `pandas`  though...`pandas` is not strictly required for `lightgbm`: https://github.com/microsoft/LightGBM/blob/cf69591b68c68a78ea4064e6941b8270b8a00eef/python-package/lightgbm/compat.py#L61

jameslamb (changes requested):
Thanks  @SfinxCZ ! I left a small suggestion for `setup.py`.

Please also fix the the linting issues (https://travis-ci.org/github/microsoft/LightGBM/jobs/740698092). You can replicate them by running this from the root of the repo:

```shell
pycodestyle --ignore=E501,W503 --exclude=./compute,./.nuget .

pydocstyle --convention=numpy --add-ignore=D105 --match-dir="^(?!^compute|test|example).*" --match="(?!^test_|setup).*\.py" .
```

jameslamb:
@SfinxCZ I want to be sure to respect your time...would you like me to open a pull request into this branches that tries to fix the other CI jobs?

Once CI is working, then we can have reviews that discuss the API, documentation, etc.

python-package/setup.py
SfinxCZ: I am not sure if adding pandas as package dependency is necessary, since the dask integration does not need the pandas. 

The only place where it is required are tests but I was not able to find where are listed dev dependencies. Can you please point me to correct place where I can add the dev dependencies?

jameslamb: oh sure!

If it's a test-only dependency, then you're already covered. We install `pandas` into our testing conda environment:

* https://github.com/microsoft/LightGBM/blob/cf69591b68c68a78ea4064e6941b8270b8a00eef/.ci/test.sh#L73
* https://github.com/microsoft/LightGBM/blob/cf69591b68c68a78ea4064e6941b8270b8a00eef/.ci/test_windows.ps1#L25

jameslamb:
As you look at CI, you can safely ignore these:

* GitHub Actions: `R GitHub Actions / r-package (windows-latest, MINGW, R 3.6, cmake) (pull_request) `
* Travis: `check-docs`

We have open PRs trying to fix some issues with them right now.

jameslamb:
> As you look at CI, you can safely ignore these:
> 
> * GitHub Actions: `R GitHub Actions / r-package (windows-latest, MINGW, R 3.6, cmake) (pull_request) `
> * Travis: `check-docs`
> 
> We have open PRs trying to fix some issues with them right now.

ok, these PRs I referenced have been merged. So if you merge the latest `master` into this branch, it should help with CI.

jameslamb:
One thing I noticed that I want to bring to your attention @SfinxCZ ....I notice that all of your commits look anonymous (not connected to your GitHub account)

image

This probably means you are pushing from a place where you git config has an email that is not linked to your GitHub. I want to be sure that you get credit in the contributors list when we merge this, so I think you should overwrite these commits and be sure to configure git with an email tied to your GitHub account. I've documented how to do this here, linking in case it helps: https://github.com/uptake/pkgnet/pull/284#pullrequestreview-518230330

SfinxCZ:
I've fixed the CI scripts, the only problem is in https://travis-ci.org/github/microsoft/LightGBM/jobs/743647169 which I assume is some random fail, however I cannot rerun the pipeline. @jameslamb, can you please rerun the pipeline to check whether the fail was indeed some random fluctuation?

The code is ready for review. There are some restrictions that I needed to set. First, only linux is supported for now (can be extended in different PR, as suggested by @jameslamb). And second is that only python >= 3.6 is supported. The reason to drop support for python 2.7 is that dask does not support python 2.7 either since version 2.0. If you really need the support I can try to add it but I believe that supporting python 2.7 is not necessary these days.

StrikerRUS:
> If you really need the support I can try to add it but I believe that supporting python 2.7 is not necessary these days.

We are going to drop Python 2.7 support in a next few days with `3.1.0` release: https://github.com/microsoft/LightGBM/pull/3484#issuecomment-723484953. So, don't worry about it.

jameslamb:
> @jameslamb, can you please rerun the pipeline to check whether the fail was indeed some random fluctuation?

Yep, no problem! I just restarted it. That task (Travis, `TASK=check-docs`) scrapes all the links from our documentation and checks that they're still active. It fails sometimes  due to random network issues.

jameslamb (changes requested):
Thanks very much for all the efforts and for being open to this, @SfinxCZ !

I left a few suggestions. I tried to only focus my review on things like the public API of this new module. We can consider other changes to the internals of these functions / classes through normal maintenance in later PRs.

I'd like to make two additional requests, related to governance over this code after this PR:

1. Can you please add a line to `CODEOWNERS` (https://github.com/microsoft/LightGBM/blob/master/.github/CODEOWNERS#L32) that makes sure I'm added as a reviewer on any PRs that touch the Dask module? I committed to the other LightGBM maintainers that if we bring `dask-lightgbm` into this project, I'll take on the primary maintenance burden (https://github.com/microsoft/LightGBM/issues/2791#issuecomment-716695841)

2.  Can you please let me know how involved you want to be after this PR is merged? I want to be respectful of all the awesome work you and the other `dask-lightgbm`  maintainers have done, but I also want to give you the opportunity to eventually step away from it if you'd like. It sounds like @striajan is ok not being involved (https://github.com/dask/community/issues/104#issuecomment-720321177).

----------------

### Notes for other LightGBM maintainers

I think we should not merge this until 3.1.0 is released (#3484 ), and then try to target including it in an eventual 3.2.0 release. That way, we can make breaking changes if necessary, add documentation with examples, try to add support for Mac and Windows, etc.

python-package/lightgbm/dask_distributed.py
```diff
@@ -0,0 +1,301 @@
...
+    if isinstance(part, pd.DataFrame):
+        if proba:
+            result = pd.DataFrame(result, index=part.index)
+        else:
+            result = pd.Series(result, index=part.index, name='predictions')
+
+    return result
+
+
+def predict(client, model, data, proba=False, dtype=np.float32, **kwargs):
```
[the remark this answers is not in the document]
SfinxCZ: This is an internal method that do most of the heavy lifting for the distributed training (from the dask side) and the only difference between `predict` and `predict_proba` is which method from the lightgbm model will be called on the workers. 

So I renamed it to `_predict` to hide this method from the user, since it should not be used directly.

.github/CODEOWNERS
```diff
@@ -30,6 +30,10 @@ R-package/    @Laurae2 @jameslamb
 # Python code
 python-package/    @StrikerRUS @chivee @wxchan @henry0312
 
+# Dask integration
+python-package/lightgbm/dask_distributed.py    @jameslamb
+tests/python_package_test/test_dask_distributed.py    @jameslamb
```
jameslamb, line 35:
```suggestion
python-package/lightgbm/dask.py    @jameslamb
tests/python_package_test/test_dask.py    @jameslamb
```

python-package/lightgbm/dask.py
```diff
@@ -0,0 +1,284 @@
...
+
+import numpy as np
+import pandas as pd
+from dask import array as da
+from dask import dataframe as dd
+from dask import delayed
+from dask.distributed import default_client, get_worker, wait
+from toolz import assoc, first
+
+import lightgbm
```
jameslamb, line 18: can you please remove this `import lightgbm`? Like I mentioned in https://github.com/microsoft/LightGBM/pull/3515#discussion_r523698247, I think it could cause circular-import issues.

It also looks like it might be unnecessary and that you're getting everything you need from `.basic` and `.sklearn` already

jameslamb (changes requested):
Thanks! I left a few more suggestions based on the newest round of changes.

Could you also let me know your answer to this question from https://github.com/microsoft/LightGBM/pull/3515#pullrequestreview-530763704?

> How involved do you want to be after this PR is merged?

I think we'll have a lot of activity on this new interface in the next month or two as we prepare for LightGBM 3.2.0. Please let me know if you want to be added as a reviewer on any of those PRs.

python-package/lightgbm/dask.py
SfinxCZ: Done, I've forgot to remove it.

.github/CODEOWNERS
SfinxCZ: Done

python-package/lightgbm/dask.py
jameslamb: no problem, thanks!

SfinxCZ:
I've incorporated the review comments. Problem is that the code will not work with python 2.7. The problem is with the module `lightgbm.dask` which is in conflict with the main dask module. Other than that it should work.

jameslamb:
Thanks!

> The problem is with the module lightgbm.dask which is in conflict with the main dask module

I don't think the issues on CI are because of a conflict between `lightgbm` and `dask. I can explain each of the CI failures below.

### Windows Builds (Appveyor, Azure DevOps)

* https://ci.appveyor.com/project/guolinke/lightgbm/builds/36340316/job/o85ktoo4wmc9xtc5
* https://ci.appveyor.com/project/guolinke/lightgbm/builds/36340316/job/hrhjta96xhbdunm0
* https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=8000&view=logs&j=e0c3ed1e-0c79-508f-2647-33e12d6ed3d4&t=d764710f-c78e-54fd-dda4-11d6a80fefc7
* https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=8000&view=logs&j=67ae2ff8-4b33-578b-0813-9c0440091667
* https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=8000&view=logs&j=ea56812e-e7ae-55d0-6abc-4a217857fa9f

```text
___________ ERROR collecting tests/python_package_test/test_dask.py ___________
ImportError while importing test module 'C:\projects\lightgbm\tests\python_package_test\test_dask.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
C:\Miniconda37-x64\envs\test-env\lib\importlib\__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
..\tests\python_package_test\test_dask.py:4: in <module>
    import dask.array as da
E   ModuleNotFoundError: No module named 'dask'
```

These failures on Windows jobs are because of my request to skip installing `dask` dependencies on Windows for now: https://github.com/microsoft/LightGBM/pull/3515#discussion_r523702484. 

**Solution**: Any Dask imports in that test file should also be done conditionally based on whether or not the tests are run on Windows.

I think you could do something like this near the top of `test_dask.py`.

```python
import os
import sys
import pytest

if sys.platform.startswith("win"):
    pytest.skip("lightgbm.dask is not supported in Windows environments", allow_module_level=True)
```

### Travis (sdist, Python 2.7)

* https://travis-ci.org/github/microsoft/LightGBM/jobs/744006503
* https://travis-ci.org/github/microsoft/LightGBM/jobs/744006513

```text
ImportError while importing test module '/home/travis/build/microsoft/LightGBM/tests/python_package_test/test_dask.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
../tests/python_package_test/test_dask.py:16: in <module>
    import lightgbm.dask as dlgbm
../../../../.local/lib/python2.7/site-packages/lightgbm/dask.py:8: in <module>
E   ImportError: No module named parse
```

**Solution**: Like you said, this is a problem with Python 2.7 compatibility specifically and you don't need to take any action since we'll remove Python 2.7 CI jobs in a few days, now that 3.1.0 has officially been released (#3484).

### Travis, `check-docs`

* https://travis-ci.org/github/microsoft/LightGBM/jobs/744006507

**Solution**: I'll re-run this manually, nothing you need to do. (https://github.com/microsoft/LightGBM/pull/3515#issuecomment-727355368)

tests/python_package_test/test_dask.py
```diff
@@ -0,0 +1,210 @@
...
+from sklearn.datasets import make_blobs, make_regression
+
+import lightgbm
+import lightgbm.dask as dlgbm
+
+data_output = ['array', 'scipy_csr_matrix', 'dataframe']
+data_centers = [[[-4, -4], [4, 4]], [[-4, -4], [4, 4], [-4, 4]]]
+
+pytestmark = [
+    pytest.mark.skipif(sys.platform != "linux", reason="Only linux is currently supported"),
```
jameslamb, line 22: Is it true that only Linux is supported? Or is it "not Windows"?

I don't recall seeing any issues in the CI jobs on Mac. Could this check and skip message be changed?

```suggestion
    pytest.mark.skipif(sys.platform.startswith("win"), reason="lightgbm.dask is not currently supported on Windows"),
```

jameslamb (changes requested):

tests/python_package_test/test_dask.py
SfinxCZ: I had problems with osx builds where the build timed-out after 1hour. I am not entirely sure what caused the problems. So for now, I suggest to support only linux and solve the issue with osx in different PR.

jameslamb: oh ok no problem! I'm ok with that suggestion. I just didn't remember it.

jameslamb:
@SfinxCZ could you please merge `master` into this branch? We recently fixed the issue with the GitHub Actions jobs on Windows (#3573)

jameslamb:
@SfinxCZ I just changed the title of this PR. PR titles become changelog entries in our releases, so I wanted it to read more like a changelog item.

All CI is passing! I'm going to pull this and test right now with `dask-cloudprovider`. I'll offer another review after that, but I'd like to also get reviews from both @StrikerRUS and  @guolinke since this is a major addition.

StrikerRUS:
Yeah, thanks, I'll provide my review soon.

One fundamental question so far: will Dask integration be in a form of enhancement of the current Python-package or maybe it is better to make a new separate package (`lightgbm-dask`, for example) due to [size](https://github.com/microsoft/LightGBM/pull/3579#issue-522976037) and additional dependencies reasons?

jameslamb:
> Yeah, thanks, I'll provide my review soon.
> 
> One fundamental question so far: will Dask integration be in a form of enhancement of the current Python-package or maybe it is better to make a new separate package (`lightgbm-dask`, for example) due to [size](https://github.com/microsoft/LightGBM/pull/3579#issue-522976037) and additional dependencies reasons?

My vote is that we keep it inside `lightgbm` instead of creating another package. The size effect is minimal (it only adds a single 300ish-line Python file and a few lines in setup.py).

I think we could keep the additional dependencies optional and minimize their effect by adopting the approach `xgboost` has taken (which they borrowed from `tensorflow` 😂 ) to lazy-load Dask dependencies:

* https://github.com/dmlc/xgboost/blob/42d31d9dcb6f7c1cb7d0545e9ab3a305ecad0816/python-package/xgboost/compat.py#L139
* https://github.com/dmlc/xgboost/blob/42d31d9dcb6f7c1cb7d0545e9ab3a305ecad0816/python-package/xgboost/compat.py#L105-L111
* https://github.com/dmlc/xgboost/blob/42d31d9dcb6f7c1cb7d0545e9ab3a305ecad0816/python-package/xgboost/dask.py#L41-L45

I also think adding another package for users to think about might get confusing. Imagine reading a tutorial that says

> make sure to `pip install lightgbm-dask`, not `pip install dask-lightgbm`

@trivialfis @hcho3 could you share with us any other relevant information on why you decided to keep the XGBoost Dask integration inside of the  `xgboost` package, instead of just managing a separate package inside the XGBoost repo?

StrikerRUS:
> make sure to pip install `lightgbm-dask`, not pip install `dask-lightgbm`

It shouldn't be a problem. We can continue uploading new releases under the old name (I took `lightgbm-dask` out if nothing). Thanks that PyPI maintainer is right here and can add us as collaborators at PyPI site to allow us update that old project.

jameslamb:
ooooo I like the idea of publishing new releases under the name `dask-lightgbm` from this repo! That's a very interesting idea, and it's not what I thought you meant by "separate package".

Let's hear opinions from @SfinxCZ and our friends from XGBoost (@trivialfis @hcho3)

SfinxCZ:
Sorry for late response, I was kinda busy these few days.

Regarding the size of the package as pointed by @StrikerRUS (https://github.com/microsoft/LightGBM/pull/3515#issuecomment-731600166), adding dask support adds roughly 10kb which is acceptable for me. For now, the dependencies required in dask support are added as optional extras so it does not increase the size of the package or its dependencies. Using some elaborate loading scheme with lazy loading seems a bit overkill to me as it would be nightmare to debug or even test such functionality.

One thing I don't quite understand is how would you publish the `dask-lightgbm` package? I can give you access there is no problem with that, but it would be standard `lightgbm` with `dask` as dependencies? Or it would be something else (e.g. only `dask` the support module)? Would it have the same version as `lightgbm`?
The reason why I like the current approach with `dask` extras is that you can compile/install `lightgbm` with e.g. GPU support AND use `dask` extras to get distributed GPU training (if such thing actually works). Plus it is more pythonic way how to handle optional dependencies, rather than publishing completely different package.

Last note is my involvement in future development (https://github.com/microsoft/LightGBM/pull/3515#pullrequestreview-530763704). As you have seen, I don't have enough time to support this package so I would like if someone (probably @jameslamb ) would take over the `dask` integration and provide the much needed support. I surely can help you, but relying solely on me solving future issues would not be optimal.

jameslamb:
> I would like if someone (probably @jameslamb ) would take over the dask integration and provide the much needed support

Thanks! I am happy (and I think qualified) to take it over. Just wanted to be sure I was respectful of all the hard work you and your co-maintainers on `dask-lightgbm` did to get the integration this far.

> The reason why I like the current approach with dask extras is that you can compile/install lightgbm with e.g. GPU support AND use dask extras to get distributed GPU training (if such thing actually works).

This and the other points you raised are convincing for me. I vote that we keep this integration in `lightgbm` instead of releasing a separate package.

One goal of https://github.com/dask/community/issues/104 was to make "use LightGBM on Dask" as easy as possible, and I think keeping this integration in `lightgbm` accomplishes that with very little downside.

StrikerRUS:
@SfinxCZ 

> One thing I don't quite understand is how would you publish the dask-lightgbm package? I can give you access there is no problem with that, but it would be standard lightgbm with dask as dependencies? Or it would be something else (e.g. only dask the support module)?


I thought it will be a small package `dask-lightgbm` (https://github.com/dask/dask-lightgbm/blob/master/dask_lightgbm/core.py) with `dask` and `lightgbm` packages as requirements in `setup.py` file.

> Would it have the same version as lightgbm?


Yes, we use the same version for all LightGBM packages:
https://github.com/microsoft/LightGBM/blob/master/VERSION.txt

TomAugspurger:
My preference is for a lightgbm.dask submodule, mimicking how xgboost.dask does things.

jameslamb (approved):
Ok I just tested with `dask-cloudprovider`. Looks good to me!

I watched the Dask dashboard and it looked like the task graph I expected was created. Those green dots to the far right in the graph are `train_part` tasks, one per worker.

image

The fitted model looks correct as well.

image

image

From my perspective, this PR is approved. It's a major addition, though, so we shouldn't merge until @StrikerRUS and @guolinke also comment.

On the question of "should this be a separate package or a submodule of `lightgbm`", I vote for "submodule of `lightgbm`". It seems that @SfinxCZ (https://github.com/microsoft/LightGBM/pull/3515#issuecomment-735253636), @TomAugspurger (https://github.com/microsoft/LightGBM/pull/3515#issuecomment-735412125) and @hcho3 (:+1: 
 on https://github.com/microsoft/LightGBM/pull/3515#issuecomment-735412125) favor that as well.

Also because this is a new feature and we'll want some time to test and change things, it shouldn't be merged until after the 3.1.1 release is out (#3611 ).

<details><summary>how to test with dask-cloudprovider (click me)</summary>

1. Place a file `aws.env` in your working directory, with content like this:
    * ```text
       AWS_SECRET_ACCESS_KEY=#YOUR KEY#
       AWS_ACCESS_KEY_ID=#YOUR KEY ID#
       ```
2. Place a file `test-dask-lgb.sh` in your working directory. This script will build a wheel from this PR and publish it to a public-readable S3 bucket, so the location of that wheel can be used with `pip install`. Change the bucket name in this script to one that you control.
    * ```shell
      #!/bin/bash

      set -eo pipefail

      BUCKET_NAME=dask-lightgbm-test-2
      RANDOM_DIR=/tmp/lgb-test-$(python -c "import uuid; print(str(uuid.uuid4()))")

      echo ""
      echo "running tests from ${RANDOM_DIR}/LightGBM"
      echo ""

      mkdir -p ${RANDOM_DIR}

      pushd ${RANDOM_DIR}

        # clone over current state of this PR
        cd LightGBM
        git remote add dask-lgb https://github.com/SfinxCZ/LightGBM.git

        git fetch dask-lgb master
        git checkout dask-lgb/master
        git checkout -b dask-lgb-test
        git submodule init
        git submodule update

        echo ""
        echo "building Python wheel"
        echo ""
        pushd $(pwd)/python-package

            python setup.py bdist_wheel --plat-name=manylinux1_x86_64 --universal
            WHEEL_FILE=lightgbm-3.1.0.99-py2.py3-none-manylinux1_x86_64.whl

            echo ""
            echo "installing lightgbm"
            echo ""
            pip uninstall -y lightgbm
            pip install dist/${WHEEL_FILE}

            # smoke test the installation
            python -c "from lightgbm.dask import LGBMRegressor"

            echo ""
            echo "done installing"
            echo ""

            echo ""
            echo "copying package to S3"
            echo ""
            S3_KEY=s3://${BUCKET_NAME}/${WHEEL_FILE}
            aws s3 cp \
                --acl public-read-write \
                    dist/${WHEEL_FILE} \
                    ${S3_KEY}

            echo "S3 location: ${S3_KEY}"
          popd

      popd
      ```
3. Work from a container running the `daskdev/dask:latest` image. This command looks gross with all those `EXTRA_PIP_PACKAGES`, but that was my way to get around super-long installation times from the new `pip` resolver (https://github.com/pypa/pip/issues/9187), so it's not related to this PR's changes.
    * ```shell
      docker run \
           --env-file aws.env \
           -e EXTRA_APT_PACKAGES="build-essential cmake" \
           -e EXTRA_PIP_PACKAGES="--no-deps aiobotocore==1.1.2 aiohttp==3.3.1 aioitertools==0.5.1 async-timeout attrs awscli==1.18.188 botocore==1.17.44 colorama==0.4.3 dask-cloudprovider[aws]==0.5.1 docutils==0.15 joblib==0.17.0 jmespath==0.10.0 multidict==4.0 numpy==1.19.4 pandas==1.1.4 pyasn1==0.4.8 rsa==4.5.0 s3transfer==0.3.0 scikit-learn==0.23.2 scipy==1.5.4 wrapt==1.10.10 yarl" \
           -v $(pwd):/tmp/LightGBM \
           -w /tmp/LightGBM \
           -it \
           daskdev/dask:latest \
           /bin/bash

      ./test-dask-lgb.sh
      ```
4. Once that script is done running, start up a `python` session inside the container and run this code to test. You can navigate in your browser to the dashboard link that is printed, to watch the performance of the Dask cluster.

```python
import dask.array as da
import os

from dask_cloudprovider.aws import FargateCluster
from dask.distributed import Client, wait

from lightgbm.dask import LGBMRegressor

BUCKET_NAME = "dask-lightgbm-test-2"
PACKAGE_FILE = "lightgbm-3.1.0.99-py2.py3-none-manylinux1_x86_64.whl"
LIGHTGBM_TARBALL = f"https://{BUCKET_NAME}.s3-us-east-2.amazonaws.com/{PACKAGE_FILE}"

n_workers = 3

os.environ["AWS_DEFAULT_REGION"] = "us-east-2"

cluster = FargateCluster(
    image="daskdev/dask:latest",
    worker_cpu=512,
    worker_mem=4096,
    n_workers=n_workers,
    fargate_use_private_ip=False,
    scheduler_timeout="15 minutes",
    find_address_timeout=60 * 10,
    environment={
        "EXTRA_PIP_PACKAGES": (
            "--no-deps aiobotocore==1.1.2 aiohttp==3.3.1 aioitertools==0.5.1 "
            "async-timeout attrs awscli==1.18.188 botocore==1.17.44 colorama==0.4.3 "
            "dask-cloudprovider[aws]==0.5.1 docutils==0.15 joblib==0.17.0 "
            f"{LIGHTGBM_TARBALL} "
            "jmespath==0.10.0 multidict==4.0 numpy==1.19.4 pandas==1.1.4 pyasn1==0.4.8 "
            "rsa==4.5.0 s3transfer==0.3.0 scikit-learn==0.23.2 scipy==1.5.4 wrapt==1.10.10 "
            "yarl"
        ),
        "EXTRA_APT_PACKAGES": "build-essential cmake"
    }
)
client = Client(cluster)

num_rows = 1e6
num_features = 1e2
num_partitions = 10
rows_per_chunk = num_rows / num_partitions

data = da.random.random(
    (num_rows, num_features),
    (rows_per_chunk, num_features)
)

labels = da.random.random(
    (num_rows, 1),
    (rows_per_chunk, 1)
)


data = data.persist()
labels = labels.persist()
_ = wait(data)
_ = wait(labels)


dask_reg = LGBMRegressor(
    silent=False,
    max_depth=5,
    random_state=708,
    objective="regression_l2",
    learning_rate=0.1,
    tree_learner="data",
    n_estimators=10,
    min_child_samples=1,
    n_jobs=-1
)

dask_reg.fit(
    client=client,
    X=data,
    y=labels,
)
````

5. All of the resources in AWS will be torn down after 15 minutes of inactivity. You can also manually delete them from the ECR console in AWS.

</details>

**Comment by jameslamb** [comment]
@SfinxCZ , we've recently merged the PR that officially ends support for Python 2 in this project (#3581 )

Can you please merge `master` into this branch, and remove the `skipif` logic in tests that refers to Python version? I want to be sure that our CI jobs will break if we accidentally get Python 2 (like, for example, getting `/usr/bin/Python` instead of the one we install from conda).

Thanks very much!

**Comment by guolinke** [comment]
Thank you! @SfinxCZ @jameslamb . 
The PR looks good to me. BTW, can we add an example about how to use dask-lightgbm?

**Comment by jameslamb** [comment]
> BTW, can we add an example about how to use dask-lightgbm?

I think the tests and my test script in https://github.com/microsoft/LightGBM/pull/3515#pullrequestreview-543175003 are a good enough start.

My preference is to merge this PR, and then I'll create several issues and list out "these are the issues that we should try to complete before 3.2.0".

I think that's preferable so then @SfinxCZ can step away from this, and we can do it in smaller PRs that are easier to review.

Is that ok?

**Comment by guolinke** [comment]
@jameslamb sounds great! thank you

**Review by guolinke — APPROVED** [review]

**Comment by jameslamb** [comment]
@StrikerRUS could you review this again before it's merged? (keeping https://github.com/microsoft/LightGBM/pull/3515#issuecomment-742193027 in mind)

**Comment by StrikerRUS** [comment]
@jameslamb Yeah, I'll take a look this weekend.

File: python-package/lightgbm/dask.py [hunk 6]
```diff
@@ -0,0 +1,283 @@
...
+    key_to_part_dict = dict([(part.key, part) for part in parts])
+    who_has = client.who_has(parts)
+    worker_map = defaultdict(list)
+    for key, workers in who_has.items():
+        worker_map[first(workers)].append(key_to_part_dict[key])
+
+    master_worker = first(worker_map)
+    worker_ncores = client.ncores()
+
+    if 'tree_learner' not in params or params['tree_learner'].lower() not in {'data', 'feature', 'voting'}:
```
StrikerRUS, line 139: Support aliases in both param name (via `_ConfigAliases`)
https://github.com/microsoft/LightGBM/blob/b69364e9280888914ea716f6e310add457b88d74/python-package/lightgbm/basic.py#L200
and param values (`feature_parallel`, `data_parallel`, `voting_parallel`).
https://lightgbm.readthedocs.io/en/latest/Parameters.html#tree_learner

python-package/lightgbm/dask.py
```diff
@@ -0,0 +1,283 @@
...
+from collections import defaultdict
+from urllib.parse import urlparse
+
+import numpy as np
+import pandas as pd
+from dask import array as da
+from dask import dataframe as dd
+from dask import delayed
+from dask.distributed import default_client, get_worker, wait
+from toolz import assoc, first
```
StrikerRUS, line 16: Do we really need the whole new library as a dependency just for

```
def assoc(d, key, value, factory=dict):
    """ Return a new dict with new key value pair

    New dict has d[key] set to value. Does not modify the initial dictionary.

    >>> assoc({'x': 1}, 'x', 2)
    {'x': 2}
    >>> assoc({'x': 1}, 'y', 3)   # doctest: +SKIP
    {'x': 1, 'y': 3}
    """
    d2 = factory()
    d2.update(d)
    d2[key] = value
    return d2
```

```
def first(seq):
    """ The first element in a sequence

    >>> first('ABC')
    'A'
    """
    return next(iter(seq))
``` 
?

python-package/lightgbm/dask.py
```diff
@@ -0,0 +1,283 @@
...
+
+    if 'tree_learner' not in params or params['tree_learner'].lower() not in {'data', 'feature', 'voting'}:
+        logger.warning('Parameter tree_learner not set or set to incorrect value '
+                       '(%s), using "data" as default', params.get("tree_learner", None))
+        params['tree_learner'] = 'data'
+
+    # Tell each worker to train on the parts that it has locally
+    futures_classifiers = [client.submit(_train_part,
+                                         model_factory=model_factory,
+                                         params=assoc(params, 'num_threads', worker_ncores[worker]),
```
StrikerRUS, line 147: Aliases
https://lightgbm.readthedocs.io/en/latest/Parameters.html#num_threads

python-package/lightgbm/dask.py
```diff
@@ -0,0 +1,283 @@
...
+                                         list_of_parts=list_of_parts,
+                                         worker_addresses=list(worker_map.keys()),
+                                         local_listen_port=params.get('local_listen_port', 12400),
+                                         time_out=params.get('time_out', 120),
```
StrikerRUS, line 151: Aliases
https://lightgbm.readthedocs.io/en/latest/Parameters.html#network-parameters

python-package/lightgbm/dask.py
```diff
@@ -0,0 +1,283 @@
...
+    if isinstance(part, pd.DataFrame):
+        if proba:
+            result = pd.DataFrame(result, index=part.index)
+        else:
+            result = pd.Series(result, index=part.index, name='predictions')
+
+    return result
+
+
+def _predict(model, data, proba=False, dtype=np.float32, **kwargs):
```
StrikerRUS, line 180: Why is this not a method of `_LGBMModel`?

python-package/lightgbm/dask.py
```diff
@@ -0,0 +1,283 @@
...
+    @staticmethod
+    def _copy_extra_params(source, dest):
+        params = source.get_params()
+        attributes = source.__dict__
+        extra_param_names = set(attributes.keys()).difference(params.keys())
+        for name in extra_param_names:
+            setattr(dest, name, attributes[name])
+
+
+class LGBMClassifier(_LGBMModel, LocalLGBMClassifier):
```
StrikerRUS, line 235: Is it OK that users have to import this as `from lightgbm.dask import LGBMClassifier`?
Maybe it is better to give this class a unique name and allow to import from the root `from lightgbm import <dask_class>`?

python-package/setup.py
```diff
@@ -344,6 +344,15 @@ def run(self):
               'scipy',
               'scikit-learn!=0.22.0'
           ],
+          extras_require={
+              'dask': [
+                  'dask[array]>=2.0.0',
+                  'dask[dataframe]>=2.0.0'
+                  'dask[distributed]>=2.0.0',
+                  'pandas',
+                  'toolz'
```
StrikerRUS, line 353: See comment above about `toolz`.

tests/python_package_test/test_dask.py
```diff
@@ -0,0 +1,211 @@
...
+from sklearn.datasets import make_blobs, make_regression
+
+import lightgbm
+import lightgbm.dask as dlgbm
+
+data_output = ['array', 'scipy_csr_matrix', 'dataframe']
+data_centers = [[[-4, -4], [4, 4]], [[-4, -4], [4, 4], [-4, 4]]]
+
+pytestmark = [
+    pytest.mark.skipif(os.getenv("TASK", "") == "mpi", reason="Fails to run with MPI interface")
```
StrikerRUS, line 25: Any ideas why it happens?

tests/python_package_test/test_dask.py
```diff
@@ -0,0 +1,211 @@
...
+    pytest.skip("lightgbm.dask is currently supported in Linux environments", allow_module_level=True)
+
+import dask.array as da
+import dask.dataframe as dd
+import numpy as np
+import pandas as pd
+import scipy.sparse
+from dask.array.utils import assert_eq
+from dask_ml.metrics import accuracy_score, r2_score
+from distributed.utils_test import client, cluster_fixture, gen_cluster, loop  # noqa
```
StrikerRUS, line 15:
```suggestion
from distributed.utils_test import client, cluster_fixture, gen_cluster, loop
```

tests/python_package_test/test_dask.py
```diff
@@ -0,0 +1,211 @@
...
+        dw = da.from_array(w, chunk_size)
+    else:
+        raise ValueError("Unknown output type %s" % output)
+
+    return X, y, w, dX, dy, dw
+
+
+@pytest.mark.parametrize('output', data_output)
+@pytest.mark.parametrize('centers', data_centers)
+def test_classifier(output, centers, client, listen_port):  # noqa
```
StrikerRUS, line 70:
```suggestion
def test_classifier(output, centers, client, listen_port):
```

tests/python_package_test/test_dask.py
```diff
@@ -0,0 +1,211 @@
...
+    assert_eq(s1, s2)
+
+    assert_eq(p1, p2)
+    assert_eq(y, p1)
+    assert_eq(y, p2)
+
+
+@pytest.mark.parametrize('output', data_output)
+@pytest.mark.parametrize('centers', data_centers)
+def test_classifier_proba(output, centers, client, listen_port):  # noqa
```
StrikerRUS, line 93:
```suggestion
def test_classifier_proba(output, centers, client, listen_port):
```

tests/python_package_test/test_dask.py
```diff
@@ -0,0 +1,211 @@
...
+    p1 = p1.compute()
+
+    b = lightgbm.LGBMClassifier()
+    b.fit(X, y, sample_weight=w)
+    p2 = b.predict_proba(X)
+
+    assert_eq(p1, p2, atol=0.3)
+
+
+def test_classifier_local_predict(client, listen_port):  # noqa
```
StrikerRUS, line 108:
```suggestion
def test_classifier_local_predict(client, listen_port):
```

tests/python_package_test/test_dask.py
```diff
@@ -0,0 +1,211 @@
...
+    b.fit(X, y, sample_weight=w)
+    p2 = b.predict(X)
+
+    assert_eq(p1, p2)
+    assert_eq(y, p1)
+    assert_eq(y, p2)
+
+
+@pytest.mark.parametrize('output', data_output)
+def test_regressor(output, client, listen_port):  # noqa
```
StrikerRUS, line 125:
```suggestion
def test_regressor(output, client, listen_port):
```

tests/python_package_test/test_dask.py
```diff
@@ -0,0 +1,211 @@
...
+        assert_eq(s1, s2, atol=.01)
+
+    # Predictions should be roughly the same
+    assert_eq(y, p1, rtol=1., atol=100.)
+    assert_eq(y, p2, rtol=1., atol=50.)
+
+
+@pytest.mark.parametrize('output', data_output)
+@pytest.mark.parametrize('alpha', [.1, .5, .9])
+def test_regressor_quantile(output, client, listen_port, alpha):  # noqa
```
StrikerRUS, line 151:
```suggestion
def test_regressor_quantile(output, client, listen_port, alpha):
```

tests/python_package_test/test_dask.py
```diff
@@ -0,0 +1,211 @@
...
+    b.fit(X, y, sample_weight=w)
+    p2 = b.predict(X)
+    q2 = np.count_nonzero(y < p2) / y.shape[0]
+
+    # Quantiles should be right
+    np.isclose(q1, alpha, atol=.1)
+    np.isclose(q2, alpha, atol=.1)
+
+
+def test_regressor_local_predict(client, listen_port):  # noqa
```
StrikerRUS, line 169:
```suggestion
def test_regressor_local_predict(client, listen_port):
```

tests/python_package_test/test_dask.py
```diff
@@ -0,0 +1,211 @@
...
+
+    return X, y, w, dX, dy, dw
+
+
+@pytest.mark.parametrize('output', data_output)
+@pytest.mark.parametrize('centers', data_centers)
+def test_classifier(output, centers, client, listen_port):  # noqa
+    X, y, w, dX, dy, dw = _create_data('classification', output=output, centers=centers)
+
+    a = dlgbm.LGBMClassifier(time_out=5, local_listen_port=listen_port)
```
StrikerRUS, line 73: Please do not use one-letter variable names except some known cases when it is common (loops, list comprehensions, etc.)

tests/python_package_test/test_dask.py
```diff
@@ -0,0 +1,211 @@
...
+    a = a.fit(dX, dy, sample_weight=dw, client=client)
+    p1 = a.predict_proba(dX)
+    p1 = p1.compute()
+
+    b = lightgbm.LGBMClassifier()
+    b.fit(X, y, sample_weight=w)
+    p2 = b.predict_proba(X)
+
+    assert_eq(p1, p2, atol=0.3)
```
StrikerRUS, line 105: Is it possible to increase precision?

tests/python_package_test/test_dask.py
```diff
@@ -0,0 +1,211 @@
...
+    q1 = np.count_nonzero(y < p1) / y.shape[0]
+
+    b = lightgbm.LGBMRegressor(seed=42, objective='quantile', alpha=alpha)
+    b.fit(X, y, sample_weight=w)
+    p2 = b.predict(X)
+    q2 = np.count_nonzero(y < p2) / y.shape[0]
+
+    # Quantiles should be right
+    np.isclose(q1, alpha, atol=.1)
+    np.isclose(q2, alpha, atol=.1)
```
StrikerRUS, line 166:
```suggestion
    np.testing.assert_allclose(q1, alpha, atol=.1)
    np.testing.assert_allclose(q2, alpha, atol=.1)
```

tests/python_package_test/test_dask.py
```diff
@@ -0,0 +1,211 @@
...
+    a = a.fit(dX, dy, sample_weight=dw, client=client)
+    p1 = a.predict(dX)
+    p2 = a.to_local().predict(X)
+    s1 = r2_score(dy, p1)
+    p1 = p1.compute()
+    s2 = a.to_local().score(X, y)
+
+    # Predictions and scores should be the same
+    assert_eq(p1, p2)
+    np.isclose(s1, s2)
```
StrikerRUS, line 182:
```suggestion
    np.testing.assert_allclose(s1, s2)
```

StrikerRUS (changes requested):
@SfinxCZ Thanks a lot for your contribution! I like the general concept and this PR as initial stage of Dask support. Please check some my comments below:

tests/python_package_test/test_dask.py
```diff
@@ -0,0 +1,211 @@
+import os
```
StrikerRUS, line 1:
```suggestion
# coding: utf-8
import os
```

tests/python_package_test/test_dask.py
StrikerRUS: Does it fail on every MPI job or only on `MPI source` and **NOT** on `MPI pip`?

python-package/lightgbm/dask.py
jameslamb: ^ I agree with this. Was one of the things that I thought we could do as a follow-up maintenance PR since it's a small, targeted change.

.ci/test.sh
```diff
@@ -70,7 +70,7 @@ if [[ $TASK == "if-else" ]]; then
     exit 0
 fi
 
-conda install -q -y -n $CONDA_ENV joblib matplotlib numpy pandas psutil pytest python-graphviz scikit-learn scipy
+conda install -q -y -n $CONDA_ENV joblib matplotlib numpy pandas psutil pytest python-graphviz scikit-learn scipy dask distributed dask-ml
```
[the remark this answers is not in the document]
SfinxCZ: fixed

python-package/lightgbm/dask.py
```diff
@@ -0,0 +1,283 @@
+"""Distributed training with LightGBM and Dask.distributed.
```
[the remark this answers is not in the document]
SfinxCZ: fixed

python-package/lightgbm/dask.py
SfinxCZ: fixed

tests/python_package_test/test_dask.py
jameslamb: I think we can resolve this after this PR

tests/python_package_test/test_dask.py
jameslamb: I think we can resolve this after this PR

tests/python_package_test/test_dask.py
jameslamb: I think we can resolve this after this PR

python-package/lightgbm/dask.py
jameslamb: I agree  with you, I'm in favor of matching the approach `xgboost` took.

So we'd have `DaskLGBMClassifier` and `DaskLGBMRegressor`.

python-package/lightgbm/dask.py
jameslamb: I agree but also I think this is something we can resolve after this PR, since it's not part of the user-facing API.

python-package/lightgbm/dask.py
jameslamb: I think this can be resolved after this PR

python-package/lightgbm/dask.py
jameslamb: I think this can be resolved after this PR

python-package/lightgbm/dask.py
jameslamb: I think this can be resolved after this PR

python-package/lightgbm/dask.py
```diff
@@ -0,0 +1,283 @@
...
+
+def _concat(seq):
+    if isinstance(seq[0], np.ndarray):
+        return np.concatenate(seq, axis=0)
+    elif isinstance(seq[0], (pd.DataFrame, pd.Series)):
+        return pd.concat(seq, axis=0)
+    elif ss and isinstance(seq[0], ss.spmatrix):
+        return ss.vstack(seq, format='csr')
+    else:
+        raise TypeError('Data must be one of: numpy arrays, pandas dataframes, sparse matrices (from scipy). Got %s.' % str(type(seq[0])))
```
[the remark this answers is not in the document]
jameslamb: I think this can be resolved after this PR

jameslamb:
@StrikerRUS I appreciate the thorough review! But I have a strong preference for merging this PR and then addressing other concerns in multiple smaller, follow-up PRs. I'm prepared to put a lot of time and energy into the Dask interface, but I can't start on that work until we've done this initial large merge. I have dedicated time in my current job to work on this, where I believe @SfinxCZ is (graciously) donating time on nights and weekends, so I think I can move much more quickly.

I commented on the remaining unresolved conversations above that I think can be done in a follow-up PR.

I propose that the only ones @SfinxCZ needs to take action on are https://github.com/microsoft/LightGBM/pull/3515#discussion_r541800359 and the collection of `#noqa` removals since those are quick and can be accepted in the browser.

What do you think?

StrikerRUS (approved):
@jameslamb OK, as we are in a kind of a rush, feel free to merge. I'm totally OK with the general ideas of this PR and my unaddressed concerns can be covered by following up PRs.

jameslamb:
Thanks!

@SfinxCZ could you please change the public class names (https://github.com/microsoft/LightGBM/pull/3515#discussion_r546310507) and accept the suggestions about removing `# noqa` comments?

Then we can merge this and I'll add some issues to #2302 based on the other review comments here and start working on them.

SfinxCZ:
@jameslamb I've fixed part of the comments (noqa + class renaming).

jameslamb:
thanks very much for all your hard work @SfinxCZ ! And to the other `dask-lightgbm` maintainers for starting us on this path.

I'm going to merge this, and later tonight I'll add some issues to #2302 based on the review comments here and some other ideas I have for the Dask integration.

python-package/lightgbm/dask.py
jameslamb: added https://github.com/microsoft/LightGBM/issues/3671

python-package/lightgbm/dask.py
jameslamb: added https://github.com/microsoft/LightGBM/issues/3671

python-package/lightgbm/dask.py
jameslamb: added https://github.com/microsoft/LightGBM/issues/3671

jameslamb:
Sorry, got busy with other things today. Will add more Dask issues to the backlog tomorrow.

pseudotensor:
FYI: 

Fix for early stopping: https://github.com/h2oai/dask-lightgbm/commit/8cc8e83b0235050286a89dbbe2fd9d6b2e08271b
Fix for pred_contribs: https://github.com/h2oai/dask-lightgbm/commit/c19000b57c1239b2a6de65cf172163c8684ce7dc
Don't listen to worker num_threads, unrelated to n_jobs for lightgbm in typical 1 worker 1 process mode: https://github.com/h2oai/dask-lightgbm/commit/d4136987b03d164d73da49a73e75a56d6c2a3189 (ignore DAI specific things)

jameslamb:
@pseudotensor this is the first time I've seen `h2oai/dask-lightgbm`. Could you please explain the purpose of that project?

I'd prefer that you create pull requests here instead of making changes like that on a fork, so that LightGBM users have a single place to go for LightGBM-on-Dask, and so we don't duplicate efforts (the goal of https://github.com/dask/community/issues/104).

The changes you linked above look general enough that they could be contributed directly to LightGBM.

pseudotensor:
@jameslamb Purpose is to actually use dask-lightgbm.  I wasn't getting nearly any responses for bugs I was reporting, so I didn't have much hope for making PRs.  I only right now became aware that dask-lightgbm is being put into lightgbm, which is excellent, so I'm posting those changes so someone can make a PR to lightgbm directly.

pseudotensor:
> The changes you linked above look general enough that they could be contributed directly to LightGBM.

Ya, apart from the 3rd commit that has some import line, yes, all 3 should be applicable to lightgbm.  I can make a PR in a few days if that is requested.

jameslamb:
Thanks very much!

> I'm posting those changes so someone can make a PR to lightgbm directly.
> I can make a PR in a few days if that is requested.

I'd be happy to take these changes and PR them, if you'd like! Then you don't have to deal with tests and CI. But if you'd rather contribute them I'm also happy to let you do it.

> I wasn't getting nearly any responses for bugs I was reporting to the dask-lightgbm repo, so I didn't have much hope for making PRs.

You should expect much more active development on the integration now that the code is here in LightGBM. So if you're using it and run into problems or feature gaps, we'd very much appreciate you creating [issues](https://github.com/microsoft/LightGBM/issues) and pull requests so we know what's missing.

Thanks for helping us improve LightGBM!

pseudotensor:
Yes please make the relevant PRs if you can, thanks!  I maybe didn't do the cleanest/shorted code, but I didn't like the indexing by raw index that wouldn't work in general for the parts.

ffineis:
> Yes please make the relevant PRs if you can, thanks! I maybe didn't do the cleanest/shorted code, but I didn't like the indexing by raw index that wouldn't work in general for the parts.

Hey @pseudotensor heads up, I've opened #3708, noticed your changes to support validation sets. Happy to take these over for you so that we both get what we want - support for validation sets and `group`/pairwise ranking. Lmk!

pseudotensor:
Ya, go for it.  Just coordinate with @jameslamb since he also mentioned he would do some stuff.

jameslamb:
I've added the suggestions from https://github.com/microsoft/LightGBM/pull/3515#issuecomment-752709132 as individual issues, so let's continue the conversation there if either of you have further thoughts.

* early stopping: #3712 
* `pred_contrib` fix: #3713 
* not overwriting `num_thread`: #3714 

I'll start working on these this week and will review @ffineis 's ranker PR

lostmygithubaccount:
@jameslamb apologies for the ping on the PR, not sure where a better location would be

I'm trying to build a distributed example on Azure Machine Learning using ``lightgbm.dask``.

But more importantly, an internal Bing team is looking to move a large training workload from an internal tool to the public LightGBM package. We are prototyping the move now, and hoping to use an official release of ``lightgbm`` which contains the dask submodule.

When can we expect the release? Is there a place this is documented?

jameslamb:
@lostmygithubaccount no problem! That's so exciting!

Short answer: there is no place where we've documented when the next release (3.2.0) will come out.

Long answer: We've recently been trying to do LightGBM releases at least once every two months. This time of year, with people out for various holidays, you can expect it to be a little longer. The most recent release (3.1.1) was in early December, so I think it's reasonable to expect that 3.2.0 will be out in February. But that's only a best guess, not a commitment...it depends on the availability of maintainers and on discussions about what we want to be in that release, which have not been had yet.

I know it's not ideal, but if you want to start testing the new Dask integration you can try to adopt my testing code from https://github.com/microsoft/LightGBM/pull/3515#pullrequestreview-543175003. I'd actually really really appreciate aany feedback and bug reports your team could offer on it. I'm putting a lot of attention into it over the next few weeks so we can feel confident in it when we release next.

If you have additional questions, please open a new issue and ask there. Other maintainers and contributors will see it better in an issue than on a closed PR.

jameslamb:
@lostmygithubaccount in preparation for some meetup talks, I recently put together an example for using LightGBM with Dask that I think is a lot easier than the comment I linked you to above.

https://github.com/jameslamb/talks/tree/main/recent-developments-in-lightgbm

I'd love to hear more about what your team is working on and what features really matter for you in the new Dask module. Feel free to contact me at any of the places documented in https://github.com/jameslamb, and maybe we could set up some time to talk.

PR #3524, opened 2020-11-03
[ci] upgrade GitHub Actions third-party actions to newest versions

jameslamb:
I've been noticing a bunch of these warnings in GitHub Actions jobs:

> The `add-path` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

> The `set-env` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

For example, you can see these annotations on https://github.com/microsoft/LightGBM/actions/runs/343069981.

To be sure that these changes from GitHub don't disrupt our CI, this PR updates the GitHub Actions workflows in this project to use the newest versions of third-party actions. You can see, specifically, that updates to deal with this deprecation were made in the most recent release of `actions/checkout`: https://github.com/actions/checkout/releases/tag/v2.3.4

It also fixes a mistake I made setting them up....we should be using https://github.com/conda-incubator/setup-miniconda, not this fork of it: https://github.com/goanpeca/setup-miniconda.

### DIFF
--- a/.github/workflows/r_package.yml
+++ b/.github/workflows/r_package.yml
@@ -165,6 +165,11 @@ jobs:
     runs-on: ubuntu-latest
     container: rhub/rocker-gcc-san
     steps:
+      - name: Install Git before checkout
+        shell: bash
+        run: |
+          apt-get update
+          apt-get install -y git
       - name: Checkout repository
         uses: actions/checkout@v2.3.4
         with:
@@ -173,8 +178,6 @@ jobs:
       - name: Install packages
         shell: bash
         run: |
-          apt-get update
-          apt-get install -y git
           Rscript -e "install.packages(c('R6', 'data.table', 'jsonlite', 'testthat'), repos = 'https://cran.r-project.org')"
           sh build-cran-package.sh
           Rdevel CMD INSTALL lightgbm_*.tar.gz || exit -1
--- a/.github/workflows/r_package.yml
+++ b/.github/workflows/r_package.yml
@@ -173,6 +173,8 @@ jobs:
       - name: Install packages
         shell: bash
         run: |
+          apt-get update
+          apt-get install -y git
           Rscript -e "install.packages(c('R6', 'data.table', 'jsonlite', 'testthat'), repos = 'https://cran.r-project.org')"
           sh build-cran-package.sh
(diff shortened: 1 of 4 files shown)


StrikerRUS (changes requested):
Please update this line as well.
https://github.com/microsoft/LightGBM/blob/12e220b79d459396ce74f60f260786ad67a8fbb9/.github/workflows/cuda.yml#L41

jameslamb:
> Please update this line as well.
> https://github.com/microsoft/LightGBM/blob/12e220b79d459396ce74f60f260786ad67a8fbb9/.github/workflows/cuda.yml#L41

ah missed that one!

.github/workflows/r_valgrind.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     container: wch1/r-debug
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v1
+        uses: actions/checkout@v2.3.4
```
jameslamb, line 16: /gha run r-valgrind

.github/workflows/r_package.yml
```diff
@@ -165,8 +165,13 @@ jobs:
     runs-on: ubuntu-latest
     container: rhub/rocker-gcc-san
     steps:
+      - name: Install Git before checkout
+        shell: bash
+        run: |
+          apt-get update
+          apt-get install -y git
```
StrikerRUS, line 172:
```suggestion
          apt-get install --no-install-recommends -y git
```

StrikerRUS (approved):
Thanks! Just one minor comment below.

.github/workflows/r_valgrind.yml
jameslamb: looks like valgrind is happy with this change. I did check before that the image we use for valgrind tests had a recent version of `git` in it, and it does.

https://github.com/microsoft/LightGBM/runs/1349685476?check_suite_focus=true

guolinke (approved):

PR #3527, opened 2020-11-04
ignore files from local Dask work

jameslamb:
When you work locally with `dask`, by default it will create a directory for itself called `dask-worker-space/`. You can see this by running a small example like this

```python
from dask.distributed import Client
client = Client()
```

We'll start doing more local development with Dask in this project soon, based on #3515, so I think this directory `dask-worker-space/` should be added to `.gitignore`.


StrikerRUS (approved):
OK, let's merge it ahead of main Dask-migrating PR. And I don't see any other items we need take from https://github.com/dask/dask-lightgbm/blob/master/.gitignore.

PR #3528, opened 2020-11-04
[R-package] remove Makefile hack in install.libs.R

jameslamb:
Way back in #2530 , we added some code to `R-package/src/install.libs.R` which removes `.NOTPARALLEL` directives from the makefiles generated by `CMake`. This was done to appease `R CMD CHECK` to get us closer to a CRAN-ready package.

Now that a CRAN-ready version of the R package exists and doesn't rely on CMake or `install.libs.R` at all, I think that that hack can be removed. We don't need to care that `R CMD CHECK` is complaining about this, and it might make installation with CMake faster: https://github.com/microsoft/LightGBM/pull/2530/files#r353195984.

We've already established that the CRAN package's main goal is portability / ease-of-installation and the CMake-created package's main goal is performance. I think this change is in line with that. I also think this is a good change because it reduces the complexity in `install.libs.R` (which is already quite complex) and reduces the difference between the R package and other LightGBM wrappers.

Let's see what our CI says :grinning:

### DIFF
--- a/R-package/src/install.libs.R
+++ b/R-package/src/install.libs.R
@@ -216,29 +216,6 @@ if (!makefiles_already_generated) {
   .run_shell_command("cmake", c(cmake_args, ".."))
 }
 
-# R CMD check complains about the .NOTPARALLEL directive created in the cmake
-# Makefile. We don't need it here anyway since targets are built serially, so trying
-# to remove it with this hack
-generated_makefile <- file.path(
-  build_dir
-  , "Makefile"
-)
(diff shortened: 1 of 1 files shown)


jameslamb:
I checked all of the logs of CMake builds, and don't see any new `R CMD CHECK` warnings or notes from this PR! So maybe the issue that was originally "fixed" by this hack is no longer even relevant. Makes me feel even more confident that we can remove this code :grinning:

StrikerRUS (approved):
Nice simplification!

Issue #2947, opened 2020-03-26
[python] Migrate to parametrize_with_checks for scikit-learn integration tests


rth:
Minor maintenance point but it could be worth migrating to `parametrize_with_checks` instead of using `check_estimator` directly in scikit-learn integration tests (or rather the private `_yield_all_checks` function from scikit-learn):

https://github.com/microsoft/LightGBM/blob/c870c57a14b61329bf493b3b1a9a2893a62405cf/tests/python_package_test/test_sklearn.py#L285-L292

It particular, this would make each pair of estimator-check an independent pytest check which will make them easier to skip or debug issues like https://github.com/microsoft/LightGBM/issues/2628

Reference : https://scikit-learn.org/stable/developers/develop.html#rolling-your-own-estimator

StrikerRUS:
Thanks for letting us know about `parametrize_with_checks`! I believe that with estimator tags mechanism it'll simplify some code.

I'm adding this to our requests hub.

StrikerRUS:
Closed in favor of being in #2302. We decided to keep all feature requests in one place.

Welcome to contribute this feature! Please re-open this issue (or post a comment if you are not a topic starter) if you are actively working on implementing this feature.

PR #3533, opened 2020-11-06
TST make sklearn integration test compatible with 0.24

glemaitre:
Modify the integration test for scikit-learn. Some deprecations have been raised in 0.23 and the upcoming release of 0.24 will break this test.

### DIFF
--- a/tests/python_package_test/test_sklearn.py
+++ b/tests/python_package_test/test_sklearn.py
@@ -4,20 +4,17 @@ import joblib
 import math
 import os
 import unittest
-import warnings
 
 import lightgbm as lgb
 import numpy as np
 from sklearn import __version__ as sk_version
 from sklearn.base import clone
 from sklearn.datasets import load_svmlight_file, make_multilabel_classification
-from sklearn.exceptions import SkipTestWarning
 from sklearn.metrics import log_loss, mean_squared_error
 from sklearn.model_selection import GridSearchCV, RandomizedSearchCV, train_test_split
 from sklearn.multioutput import (MultiOutputClassifier, ClassifierChain, MultiOutputRegressor,
                                  RegressorChain)
 from sklearn.utils.estimator_checks import (
-     SkipTest,
     check_parameters_default_constructible,
     check_estimator,
 )
--- a/tests/python_package_test/test_sklearn.py
+++ b/tests/python_package_test/test_sklearn.py
@@ -15,8 +15,8 @@ from sklearn.model_selection import GridSearchCV, RandomizedSearchCV, train_test
 from sklearn.multioutput import (MultiOutputClassifier, ClassifierChain, MultiOutputRegressor,
                                  RegressorChain)
 from sklearn.utils.estimator_checks import (
-    check_parameters_default_constructible,
     check_estimator,
+    check_parameters_default_constructible,
 )
 from sklearn.utils.validation import check_is_fitted
 
--- a/tests/python_package_test/test_sklearn.py
+++ b/tests/python_package_test/test_sklearn.py
@@ -455,7 +455,6 @@ class TestSklearn(unittest.TestCase):
     # sklearn <0.19 cannot accept instance, but many tests could be passed only with min_data=1 and min_data_in_bin=1
     @unittest.skipIf(sk_version < '0.19.0', 'scikit-learn version is less than 0.19')
     def test_sklearn_integration(self):
-        # we cannot use `check_estimator` directly since there is no skip test mechanism
         for name, estimator in ((lgb.sklearn.LGBMClassifier.__name__, lgb.sklearn.LGBMClassifier),
                                 (lgb.sklearn.LGBMRegressor.__name__, lgb.sklearn.LGBMRegressor)):
             # we cannot leave default params (see https://github.com/microsoft/LightGBM/issues/833)
--- a/tests/python_package_test/test_sklearn.py
+++ b/tests/python_package_test/test_sklearn.py
@@ -7,6 +7,7 @@ import unittest
 
 import lightgbm as lgb
 import numpy as np
+import pytest
 from sklearn import __version__ as sk_version
 from sklearn.base import clone
 from sklearn.datasets import load_svmlight_file, make_multilabel_classification
@@ -15,8 +16,8 @@ from sklearn.model_selection import GridSearchCV, RandomizedSearchCV, train_test
 from sklearn.multioutput import (MultiOutputClassifier, ClassifierChain, MultiOutputRegressor,
                                  RegressorChain)
 from sklearn.utils.estimator_checks import (
-    check_estimator,
     check_parameters_default_constructible,
+    parametrize_with_checks,
 )
 from sklearn.utils.validation import check_is_fitted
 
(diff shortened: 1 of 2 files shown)


tests/python_package_test/test_sklearn.py
```diff
@@ -458,22 +458,10 @@ def test_sklearn_integration(self):
         # we cannot use `check_estimator` directly since there is no skip test mechanism
         for name, estimator in ((lgb.sklearn.LGBMClassifier.__name__, lgb.sklearn.LGBMClassifier),
                                 (lgb.sklearn.LGBMRegressor.__name__, lgb.sklearn.LGBMRegressor)):
-            check_parameters_default_constructible(name, estimator)
             # we cannot leave default params (see https://github.com/microsoft/LightGBM/issues/833)
             estimator = estimator(min_child_samples=1, min_data_in_bin=1)
-            for check in _yield_all_checks(name, estimator):
-                check_name = check.func.__name__ if hasattr(check, 'func') else check.__name__
-                if check_name == 'check_estimators_nan_inf':
-                    continue  # skip test because LightGBM deals with nan
```
glemaitre, line 467: This should be handled by the tag `allow_nan`.

tests/python_package_test/test_sklearn.py
```diff
@@ -458,22 +458,10 @@ def test_sklearn_integration(self):
...
-                elif check_name == "check_no_attributes_set_in_init":
```
glemaitre, line 468: This should be handle by the `xfail` that I added

tests/python_package_test/test_sklearn.py
```diff
@@ -458,22 +458,10 @@ def test_sklearn_integration(self):
...
-                    # skip test because scikit-learn incorrectly asserts that
-                    # private attributes cannot be set in __init__
-                    # (see https://github.com/microsoft/LightGBM/issues/2628)
-                    continue
-                try:
-                    check(name, estimator)
-                except SkipTest as message:
-                    warnings.warn(message, SkipTestWarning)
+            check_parameters_default_constructible(name, estimator)
```
glemaitre, line 463: The parameter `estimator` needs to be an instance and not a class as announce in the deprecation in 0.23

tests/python_package_test/test_sklearn.py
```diff
@@ -455,25 +455,12 @@ def test_feature_importances_type(self):
     # sklearn <0.19 cannot accept instance, but many tests could be passed only with min_data=1 and min_data_in_bin=1
     @unittest.skipIf(sk_version < '0.19.0', 'scikit-learn version is less than 0.19')
```
glemaitre, line 456: I am not sure yet but we might need `0.23` to use the latest testing mechanism.

tests/python_package_test/test_sklearn.py
```diff
@@ -1149,3 +1127,24 @@ def test_check_is_fitted(self):
...
+
+def _tested_estimators():
+    for Estimator in [lgb.sklearn.LGBMClassifier, lgb.sklearn.LGBMRegressor]:
+        yield Estimator()
+
+
+@pytest.mark.skipif(
+    sk_version < "0.23.0", reason="scikit-learn version is less than 0.23"
+)
+@parametrize_with_checks(list(_tested_estimators()))
```
glemaitre, line 1140: Only the `parametrize_with_checks` was using `_xfail` in 0.23. It has the advantage to show all the test from `_yield` as with `parametrize` of `pytest`.

glemaitre: However, it needs to be a pure function not embedded within a class

tests/python_package_test/test_sklearn.py
```diff
@@ -4,24 +4,37 @@
...
-from sklearn.exceptions import SkipTestWarning
 from sklearn.metrics import log_loss, mean_squared_error
 from sklearn.model_selection import GridSearchCV, RandomizedSearchCV, train_test_split
 from sklearn.multioutput import (MultiOutputClassifier, ClassifierChain, MultiOutputRegressor,
                                  RegressorChain)
-from sklearn.utils.estimator_checks import (_yield_all_checks, SkipTest,
-                                            check_parameters_default_constructible)
 from sklearn.utils.validation import check_is_fitted
 
+from sklearn.utils.estimator_checks import check_parameters_default_constructible
```
StrikerRUS, line 20: Please move this line before `from sklearn.utils.validation ...` to keep alphabetical order for imports.

tests/python_package_test/test_sklearn.py
```diff
@@ -4,24 +4,37 @@
...
+
+try:
+    from pkg_resources import parse_version  # type: ignore
+except ImportError:
+    # setuptools not installed
+    parse_version = LooseVersion  # type: ignore
```
StrikerRUS, line 26: Seems we should import `LooseVersion` first, shouldn't we?

tests/python_package_test/test_sklearn.py
```diff
@@ -1149,3 +1139,49 @@ def test_check_is_fitted(self):
...
+        sk_version < parse_version("0.21"), reason="scikit-learn version is less than 0.21"
+    )
+    @pytest.mark.parametrize(
+        "estimator, check",
+        _generate_checks_per_estimator(_yield_all_checks, _tested_estimators()),
+    )
+    def test_sklearn_integration(estimator, check):
+        xfail_checks = estimator._get_tags()["_xfail_checks"]
+        check_name = check.__name__ if hasattr(check, "__name__") else check.func.__name__
+        if xfail_checks and check_name in xfail_checks:
```
StrikerRUS, line 1166: With `get()` it looks a bit safer.
```suggestion
        xfail_checks = estimator._get_tags().get("_xfail_checks", {})
        check_name = check.__name__ if hasattr(check, "__name__") else check.func.__name__
        if check_name in xfail_checks:
```

tests/python_package_test/test_sklearn.py
```diff
@@ -1149,3 +1139,49 @@ def test_check_is_fitted(self):
...
+        check(name, estimator)
+else:
+    @parametrize_with_checks(list(_tested_estimators()))
+    def test_sklearn_integration(estimator, check, request):
+        estimator.set_params(min_child_samples=1, min_data_in_bin=1)
+        check(estimator)
+
+
+@pytest.mark.skipif(
+    not (sk_version < parse_version("0.24")),
```
StrikerRUS, line 1180: I believe this is more readable.
```suggestion
    (sk_version >= parse_version("0.24")),
```

StrikerRUS (commented):
@glemaitre Thank you very much for updating LightGBM to follow the most recent scikit-learn API changes! Really appreciate it!

Especially thanks a lot for preserving tests for old versions, because we try to target the most wide range of versions.

Generally LGTM! Just please consider checking a few my minor comments below.

tests/python_package_test/test_sklearn.py
glemaitre: `"_xfail_checks"` is always defined because it is a default tag. At least, if there is a breaking change in the future, one could detect it with the failure. WDYT?

tests/python_package_test/test_sklearn.py
glemaitre: Actually, you are importing setuptools in `setup.py` so we can safely import from `pkg_resources`.

tests/python_package_test/test_sklearn.py
StrikerRUS: Ah, so `_xfail_checks` is defined (with empty dict?) in BaseEstimator, if I got it correctly? Then it makes sense!

Then another question. Why do we need `if xfail_checks and ...` guard?

StrikerRUS:
Fixes #2947.

StrikerRUS (approved):
Manny thanks @glemaitre !

PR #3523, opened 2020-11-03
[R-package] [ci] add correct path for mingw32-make in CI

jameslamb:
In #3469, I've been trying to figure out why this CI job on GitHub Actions often fails:

```text
- os: windows-latest
  task: r-package
  compiler: MINGW
  toolchain: MINGW
  r_version: 3.6
  build_type: cmake
```

I don't think I've solved it yet, but I did find one problem we definitely need to fix. `Rtools35` stores `mingw32-make.exe` at `C:\Rtools\mingw_64\bin\mingw32-make.exe`, but that path isn't currently on `PATH` in our CI. I think it was just lost accidentally in some refactoring.

This PR doesn't fix #3469 but based on my testing, it does seem to make that issue happen less often. So maybe it will  at least make the contributor experience better.

PR #3535, opened 2020-11-07
[ci] simplify R CI scripts

jameslamb:
I've been learning about [`shellcheck`](https://github.com/koalaman/shellcheck), a linter for shell scripts.

This PR has a small simplification to `test_r_package.sh` that it suggested: replacing `grep "something | wc -l` with `grep --count "something"`. This might make that script a tiny tiny bit faster, but more importantly makes it a little easier to understand.

This PR also simplifies away a duplicated line in `test_r_package_windows.ps1` that I noticed while working on #3523

### DIFF
--- a/.ci/test_r_package.sh
+++ b/.ci/test_r_package.sh
@@ -108,7 +108,7 @@ if [[ $TASK == "r-package-check-docs" ]]; then
     Rscript --vanilla -e "install.packages('roxygen2', repos = '${CRAN_MIRROR}', lib = '${R_LIB_PATH}', dependencies = c('Depends', 'Imports', 'LinkingTo'))" || exit -1
     Rscript --vanilla -e "roxygen2::roxygenize('R-package/', load = 'installed')" || exit -1
     num_doc_files_changed=$(
-        git diff --name-only | grep -E "\.Rd|NAMESPACE" | wc -l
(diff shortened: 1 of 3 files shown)


.ci/test_r_package.sh
```diff
@@ -222,8 +221,7 @@ fi
 if [[ $R_BUILD_TYPE == "cran" ]]; then
     pragma_warning_present=$(
         cat $BUILD_LOG_FILE \
-        | grep -E "warning: unknown pragma ignored" \
-        | wc -l
+        | grep --count -E "warning: unknown pragma ignored"
```
StrikerRUS, line 224: Can we apply the same for
https://github.com/microsoft/LightGBM/blob/5dbd1df4b62a8fd33d93fe0a3c5f1bb9c3660e7c/.github/workflows/r_package.yml#L190
?

jameslamb: oh yep!

jameslamb: updated in https://github.com/microsoft/LightGBM/pull/3535/commits/cdabc9784234486563a41d2ce43aaf5c9521bf34

StrikerRUS (approved):
Thanks! LGTM!

Issue #3525, opened 2020-11-03
[R-package] invalid read detected by valgrind


jameslamb:
Based on the result of a recent run of the valgrind job (https://github.com/microsoft/LightGBM/runs/1349685476?check_suite_focus=true), it looks like recent commits have introduced an illegal read.

```text
==2055== Invalid read of size 1
==2055==    at 0x15335D0F: LightGBM::LeafSplits::Init(float const*, float const*) (leaf_splits.hpp:74)
==2055==    by 0x15329D9E: LightGBM::SerialTreeLearner::BeforeTrain() (serial_tree_learner.cpp:264)
==2055==    by 0x153296B3: LightGBM::SerialTreeLearner::Train(float const*, float const*) (serial_tree_learner.cpp:166)
==2055==    by 0x151A0A56: LightGBM::GBDT::TrainOneIter(float const*, float const*) (gbdt.cpp:379)
==2055==    by 0x15458612: LightGBM::Booster::TrainOneIter(float const*, float const*) (c_api.cpp:364)
==2055==    by 0x15449C4E: LGBM_BoosterUpdateOneIterCustom (c_api.cpp:1634)
==2055==    by 0x1546C469: LGBM_BoosterUpdateOneIterCustom_R (lightgbm_R.cpp:417)
==2055==    by 0x4941BDA: R_doDotCall (dotcode.c:610)
==2055==    by 0x494CCC6: do_dotcall (dotcode.c:1281)
==2055==    by 0x499FB01: bcEval (eval.c:7078)
==2055==    by 0x498B67F: Rf_eval (eval.c:727)
==2055==    by 0x498E414: R_execClosure (eval.c:1895)
==2055==  Address 0x8570974 is 260 bytes inside a block of size 1,512 free'd
==2055==    at 0x483CFBF: operator delete(void*) (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==2055==    by 0x151AA06F: std::default_delete<LightGBM::Config>::operator()(LightGBM::Config*) const (unique_ptr.h:85)
==2055==    by 0x151AA6FA: std::__uniq_ptr_impl<LightGBM::Config, std::default_delete<LightGBM::Config> >::reset(LightGBM::Config*) (unique_ptr.h:182)
[...]

PR #3536, opened 2020-11-07
[ci] Make valgrind check stricter

jameslamb:
An issue caught by `valgrind` was recently introduced onto `master` because the `valgrind` CI job didn't catch it: #3525 .

This PR fixes that so similar problems are caught in the future. Specifically, it will now fail builds where valgrind produces "invalid read" (like https://github.com/microsoft/LightGBM/issues/3525) or "invalid write" (like https://www.stats.ox.ac.uk/pub/bdr/memtests/valgrind/lightgbm/tests/testthat.Rout) errors.

### DIFF
--- a/.ci/test_r_package_valgrind.sh
+++ b/.ci/test_r_package_valgrind.sh
@@ -64,3 +64,21 @@ if [[ ${bytes_possibly_lost} -gt 352 ]]; then
     echo "valgrind found ${bytes_possibly_lost} bytes possibly lost"
     exit -1
 fi
+
+invalid_reads=$(
+  cat ${VALGRIND_LOGS_FILE} \
+    | grep --count -i "Invalid read"
+)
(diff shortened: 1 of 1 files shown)


.ci/test_r_package_valgrind.sh
```diff
@@ -64,3 +64,21 @@ if [[ ${bytes_possibly_lost} -gt 352 ]]; then
...
+fi
+
+invalid_writes=$(
+  cat ${VALGRIND_LOGS_FILE} \
+    | grep --count -i "Invalid write"
+)
+if [[ ${invalid_writes} -gt 0 ]]; then
+    echo "valgrind found invalid writes: ${invalid_writes}"
+    exit -1
+fi
```
jameslamb, line 84: /gha run r-valgrind

guolinke (approved):

StrikerRUS (approved):

PR #3539, opened 2020-11-08
[refactor] Reduce code duplication in c_api.cpp

AlbertoEAF:
Remove code duplication in c_api.cpp through use of templates.

### DIFF
--- a/src/c_api.cpp
+++ b/src/c_api.cpp
@@ -2392,136 +2392,78 @@ RowPairFunctionFromDenseRows(const void** data, int num_col, int data_type) {
   };
 }
 
+template<typename T, typename T1, typename T2>
+std::function<std::vector<std::pair<int, double>>(T idx)>
+RowFunctionFromCSR_helper(const void* indptr, const int32_t* indices, const void* data) {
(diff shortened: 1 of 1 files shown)


AlbertoEAF:
Can someone review? :)

AlbertoEAF:
I wonder if we should move the helper methods to a new file `c_api_helpers.cpp` + `c_api_helpers.h` and declare only the methods that we need from such helpers as the CSC_RowIterator and not the auxiliary methods to implement those such as the low level template methods.

AlbertoEAF (commented):
I wonder if we should add those 2 extra files (helpers apart from public API methods), or keep everything in the same file as it is - simpler, and less code.

AlbertoEAF:
I thought again and don't see splitting this into separate files as that useful.

@guolinke can you review?

jameslamb:
Could you please update this PR to the latest `master`? We recently learned that our CI was missing one of the checks that replicates CRAN (#3585 ), and added a new R CI job (#3588). Sorry for the inconvenience.

guolinke (approved):
LGTM

PR #3550, opened 2020-11-09
[ci][R-package] fix cleanup permissions

StrikerRUS:
From RTD builds
```
Warning: file 'lightgbm/cleanup' did not have execute permissions: corrected
```


jameslamb (changes requested):
This failure in the cmake builds:

```
** using staged installation
** libs
Warning: no source files found
```

Could be fixed by adding `cleanup` to this list, I think: https://github.com/microsoft/LightGBM/blob/master/build_r.R#L128

StrikerRUS:
OK, this change breaks every build. So weird, but I have no time to dig deeper. Closing, that original warning doesn't look scary.

StrikerRUS:
@jameslamb Ah, so nice timing for comment! But you was 0.00001sec faster! 😄

jameslamb:
Haha no problem. I can open another PR to remove that warning some time this week. I have some time to experiment with it.

StrikerRUS:
@jameslamb Thanks! I think it's suuuper minor 🙂

PR #3552, opened 2020-11-10
[R-package] fix install warning

jameslamb:
This PR fixes this warning produced while installing the R package:

> Warning: file 'lightgbm/cleanup' did not have execute permissions: corrected

Since the GitHub UI doesn't make it clear...the  change to `R-package/cleanup` is from running `chmod +x R-package/cleanup`


guolinke (approved):

jameslamb:
It seems from the logs that this warning is gone now :grinning: 

Thanks for the nudge @StrikerRUS

StrikerRUS:
Thanks a lot @jameslamb ! 🙂

PR #3553, opened 2020-11-10
Add note on LightGBM/OpenMP hanging on AWS Batch

cvsmith:
Saw this problem on AWS Batch...
- LightGBM ran fine in standalone Batch jobs (one container per instance)
- LightGBM ran really slowly/hanged in Batch array jobs (multiple containers per instance)

### DIFF
--- a/docs/FAQ.rst
+++ b/docs/FAQ.rst
@@ -172,6 +172,11 @@ For C/C++ users, any OpenMP feature cannot be used before the fork happens. If a
 fork happens (example: using OpenMP for forking), OpenMP will hang inside the forked sessions. Use new processes instead
 and copy memory as required by creating new processes instead of forking (or, use Intel compilers).
 
+Cloud platform container services may cause LightGBM to hang, if they use Linux fork to run multiple containers on a 
+single instance. For example, LightGBM hangs in AWS Batch array jobs, which `use the ECS agent 
+<https://aws.amazon.com/batch/faqs/#Features>`__ to manage multiple running jobs. Setting ``nthreads=1`` to mitigates 
+the issue.
+
 12. Why is early stopping not enabled by default in LightGBM?
 -------------------------------------------------------------
 
--- a/docs/FAQ.rst
+++ b/docs/FAQ.rst
@@ -174,8 +174,7 @@ and copy memory as required by creating new processes instead of forking (or, us
 
 Cloud platform container services may cause LightGBM to hang, if they use Linux fork to run multiple containers on a 
 single instance. For example, LightGBM hangs in AWS Batch array jobs, which `use the ECS agent 
-<https://aws.amazon.com/batch/faqs/#Features>`__ to manage multiple running jobs. Setting ``nthreads=1`` to mitigates 
-the issue.
+<https://aws.amazon.com/batch/faqs/#Features>`__ to manage multiple running jobs. Setting ``nthreads=1`` mitigates the issue.
 
 12. Why is early stopping not enabled by default in LightGBM?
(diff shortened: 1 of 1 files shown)


jameslamb (changes requested):
Thanks for using LightGBM and taking the time to contribute!

Can you clarify a few things for me?

* are you using Batch for training or scoring?
* when you say `nthreads=1`, do you mean the LightGBM parameter? Or are you describing something about AWS Batch?
* could you share a CloudFormation template or a screenshot of the AWS Batch console? I'm not really sure what you mean by "if they use Linux fork to run multiple containers on a single instance"

Laurae2:
`nthreads=1` is necessary when forking and using OpenMP (which LightGBM uses). Otherwise, OpenMP ends with an undefined behavior which seems it won't get fixed.

In AWS Batch it is possible to end up with some jobs isolated in a single instance (with only 1 container), which in this case forks without OpenMP activation beforehand: in this scenario, OpenMP will work correctly in the fork (for OpenMP to work in a fork, it must not have been used in the parent process).

See the following: https://github.com/microsoft/LightGBM/blob/master/docs/FAQ.rst#11-lightgbm-hangs-when-multithreading-openmp-and-using-forking-in-linux-at-the-same-time

docs/FAQ.rst
```diff
@@ -172,6 +172,11 @@ For C/C++ users, any OpenMP feature cannot be used before the fork happens. If a
 fork happens (example: using OpenMP for forking), OpenMP will hang inside the forked sessions. Use new processes instead
 and copy memory as required by creating new processes instead of forking (or, use Intel compilers).
 
+Cloud platform container services may cause LightGBM to hang, if they use Linux fork to run multiple containers on a 
+single instance. For example, LightGBM hangs in AWS Batch array jobs, which `use the ECS agent 
+<https://aws.amazon.com/batch/faqs/#Features>`__ to manage multiple running jobs. Setting ``nthreads=1`` to mitigates 
+the issue.
```
jameslamb, line 178:
```suggestion
<https://aws.amazon.com/batch/faqs/#Features>`__ to manage multiple running jobs. Setting ``nthreads=1`` mitigates the issue.
```

jameslamb (approved):
Thanks for the explanation @Laurae2 !

Ok this seems like a good addition to me then. Hopefully mentioning Batch by name will help people find this from search engines.

jameslamb (changes requested):
forgot I wanted to mark this "request changes" for the small typo fix

cvsmith:
Thanks @jameslamb! Just to followup on your questions:

> * are you using Batch for training or scoring?

Training - This problem shows up with `LGBMRegressor.fit()`. 

> * when you say `nthreads=1`, do you mean the LightGBM parameter? Or are you describing something about AWS Batch?

Yup, the LightGBM parameter. Specifically `LGBMRegressor(n_jobs=1)` in our case (`n_jobs` is the alias of `nthreads` in the `LGBMRegressor` docs, so that's what we used in our code, but `nthreads` is the parameter name mentioned in this FAQ response already, so that's what I used for this note).

> * could you share a CloudFormation template or a screenshot of the AWS Batch console? I'm not really sure what you mean by "if they use Linux fork to run multiple containers on a single instance"

I think this is covered in @Laurae2's comment 👍

cvsmith:
@Laurae2 really interesting that this issue also shows up in Azure Batch!

jameslamb (approved):
Thanks again for taking the time to contribute @cvsmith !  We hope you're enjoying using LightGBM, and will come back to contribute more in the future!

PR #3554, opened 2020-11-10
[docs] simplify RTD config and use latest Sphinx

StrikerRUS:
Build successful: https://readthedocs.org/projects/lightgbm/builds/12308946/.

I hope in the future we will be able to drop `requirements_rtd.txt` completely, but now `pip` still fails to resolve conflicts with preinstalled old (`1.8.5`) Sphinx and latest Breathe.

### DIFF
--- a/docs/requirements_base.txt
+++ b/docs/requirements_base.txt
@@ -1,3 +1,3 @@
-sphinx < 3.2
+sphinx
 sphinx_rtd_theme >= 0.3
 mock; python_version < '3'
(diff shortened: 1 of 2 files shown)


jameslamb (approved):
Oh great! Nice step towards simplifying this.

But UGH I cannot believe the R MINGW job failed on this. I guess #3541 was not enough. Maybe I got very very lucky when it built successfully 12 times in a row.

I'll re-open #3469

PR #3555, opened 2020-11-11
Add CMake option to enable sanitizers and build gtest

hcho3:
Following up with https://github.com/microsoft/LightGBM/pull/3338#issuecomment-703063297. Quote:
> My suggestion is to use Address and Leak Sanitizers. The sanitizers add extra instrumentation code to detect bad memory accesses and leaks at runtime, and unlike Valgrind have relatively light impact on performance.
> XGBoost's CI system runs C++ tests with sanitizers enabled, and so far it has caught a few memory issues (example: dmlc/xgboost#3557). I highly recommend it.

TODOs
- [x] Add CMake option to enable sanitizers (address, leak, undefined, thread)
- [x] Add CMake option to link with `gtest` (Google C++ testing framework)
- [ ] Add C++ tests. The goal is to produce a standalone testing executable that can be run with sanitizers.

Why gtest? In my experience, a shared library linked with AddressSanitizer cannot be loaded into Python or R session. It results in error
```
==48458==ASan runtime does not come first in initial library list; you should either link runtime to your application or manually preload it with LD_PRELOAD.
```
The most convenient setup is to build a standalone executable with sanitizers enabled. An additional benefit is that the C++ test executable tends to run quite fast, even when sanitizers are enabled.

An example: https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/master/604/pipeline/57#step-122-log-1, which ran the suite of C++ tests hosted in [this directory](https://github.com/dmlc/xgboost/tree/master/tests/cpp). The suite completed in about one minute.


AlbertoEAF:
Hi there, what's the state of this PR? @hcho3 

I'd like to start adding C++ tests too with gtest in another MR, so it would be immensely useful to have this merged already :)

hcho3:
I have not gotten around adding actual test cases in gtest. Right now this PR only adds gtest as an optional Camera target.

AlbertoEAF:
Not sure how others feel about it, but I think we can add tests in a separate MR and add support for gtest already so other MR's can start using gtest ;)

@StrikerRUS ?

StrikerRUS:
> Not sure how others feel about it, but I think we can add tests in a separate MR and add support for gtest already so other MR's can start using gtest ;)

I support this idea!
Let's merge PR with infrastructure setup and then everyone will be able to add tests.

@hcho3 Is this PR completed from your side and ready for review excluding "Add C++ tests." from TODOs?

hcho3:
Marking it ready for review.

CMakeLists.txt
```diff
@@ -451,6 +465,25 @@ if(__BUILD_FOR_R)
...
+      googletest
+      GIT_REPOSITORY https://github.com/google/googletest.git
+      GIT_TAG        release-1.10.0
+    )
+    FetchContent_MakeAvailable(googletest)
+    add_library(GTest::GTest ALIAS gtest)
+  endif()
+  file(GLOB CPP_TEST_SOURCES tests/cpp_test/*.cpp)
+  add_executable(testlightgbm ${CPP_TEST_SOURCES} ${SOURCES})
+  target_link_libraries(testlightgbm PRIVATE GTest::GTest)
```
StrikerRUS, line 484: Do new target names mean that we cannot combine other options like `USE_MPI` or `USE_OPENMP` with `BUILD_CPP_TEST`?

cmake/Sanitizer.cmake
```diff
@@ -0,0 +1,69 @@
+# Set appropriate compiler and linker flags for sanitizers.
+#
+# Usage of this module:
+#  enable_sanitizers("address;leak")
+
+# Add flags
+macro(enable_sanitizer sanitizer)
+  if(${sanitizer} MATCHES "address")
+    find_package(ASan)
```
StrikerRUS, line 9: Why not `REQUIRED`?

cmake/Sanitizer.cmake
```diff
@@ -0,0 +1,69 @@
...
+    else ()
+      if (${_use_thread_sanitizer})
+        message(FATAL_ERROR
+          "${_san} sanitizer is not compatible with thread sanitizer.")
+      endif()
+      set(_use_other_sanitizers 1)
+    endif()
+  endforeach()
+
+  message("Sanitizers: ${SANITIZERS}")
```
StrikerRUS, line 60:
```suggestion
  message(STATUS "Sanitizers: ${SANITIZERS}")
```

cmake/Sanitizer.cmake
```diff
@@ -0,0 +1,69 @@
...
+
+  foreach( _san ${SANITIZERS} )
+    string(TOLOWER ${_san} _san)
+    enable_sanitizer(${_san})
+  endforeach()
+  message("Sanitizers compile flags: ${SAN_COMPILE_FLAGS}")
```
StrikerRUS, line 66:
```suggestion
  message(STATUS "Sanitizers compile flags: ${SAN_COMPILE_FLAGS}")
```

cmake/modules/FindASan.cmake
```diff
@@ -0,0 +1,13 @@
...
+find_library(ASan_LIBRARY
+  NAMES libasan.so libasan.so.5 libasan.so.4 libasan.so.3 libasan.so.2 libasan.so.1 libasan.so.0
+  PATHS ${SANITIZER_PATH} /usr/lib64 /usr/lib /usr/local/lib64 /usr/local/lib ${CMAKE_PREFIX_PATH}/lib)
+
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(ASan DEFAULT_MSG
+  ASan_LIBRARY)
+
+mark_as_advanced(
+  ASan_LIBRARY
```
StrikerRUS, line 12: Sorry, could you please explain why `ASan_LIBRARY` is here and in `REQUIRED_VARS` of `find_package_handle_standard_args()` at the same time?

cmake/modules/FindASan.cmake
```diff
@@ -0,0 +1,13 @@
+set(ASan_LIB_NAME ASan)
+
+find_library(ASan_LIBRARY
+  NAMES libasan.so libasan.so.5 libasan.so.4 libasan.so.3 libasan.so.2 libasan.so.1 libasan.so.0
```
StrikerRUS, line 4:
```suggestion
  NAMES libasan.so libasan.so.5 libasan.so.4 libasan.so.3 libasan.so.2 libasan.so.1 libasan.so.0 libasan.so.0.0.0
```

cmake/modules/FindUBSan.cmake
```diff
@@ -0,0 +1,13 @@
+set(UBSan_LIB_NAME UBSan)
+
+find_library(UBSan_LIBRARY
+  NAMES libubsan.so libubsan.so.5 libubsan.so.4 libubsan.so.3 libubsan.so.2 libubsan.so.1 libubsan.so.0
```
StrikerRUS, line 4: Seems there are no other versions yet.
```suggestion
  NAMES libubsan.so libubsan.so.1 libubsan.so.0 libubsan.so.0.0.0
```

tests/cpp_test/test_main.cpp
```diff
@@ -0,0 +1,11 @@
+/*!
+ * Copyright (c) 2020 Microsoft Corporation. All rights reserved.
```
StrikerRUS, line 2:
```suggestion
 * Copyright (c) 2021 Microsoft Corporation. All rights reserved.
```

StrikerRUS (commented):
@hcho3 Thank you very much for the awesome contribution! Please consider checking some my comments/questions and suggestions below.

CMakeLists.txt
hcho3: For now, that's the case. To apply these options, we'll need to create an object library target `objlightgbm` to which we'd apply the set of common options, and then have `lightgbm` and `testlightgbm` depend on `objlightgbm`.

CMakeLists.txt
```diff
@@ -6,6 +6,12 @@ OPTION(USE_HDFS "Enable HDFS support (EXPERIMENTAL)" OFF)
 OPTION(USE_TIMETAG "Set to ON to output time costs" OFF)
 OPTION(USE_CUDA "Enable CUDA-accelerated training (EXPERIMENTAL)" OFF)
 OPTION(USE_DEBUG "Set to ON for Debug mode" OFF)
+OPTION(USE_SANITIZER "Use santizer flags" OFF)
+OPTION(SANITIZER_PATH "Path to sanitizer libs")
+SET(ENABLED_SANITIZERS "address" "leak" "undefined" CACHE STRING
```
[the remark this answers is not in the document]
hcho3: Can an option be a list? Or should we create a separate option for each sanitizer?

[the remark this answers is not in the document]
StrikerRUS: I think current behaviour (string with `Semicolon separated list of sanitizer names`) will be just fine.

CMakeLists.txt
StrikerRUS: OK, got it! Thank a lot for the detailed explanation and possible future workaround!

hcho3: Creating an object library will require a large refactor of the CMake file, so it won't be done in this PR.

cmake/modules/FindASan.cmake
hcho3: `find_package_handle_standard_args()` is the one that sets the variable `ASan_FOUND`. The `mark_as_advanced` is mostly cosmetic: it hides the variable `ASan_LIBRARY` from the CMake GUI, unless the advance mode is enabled. See https://stackoverflow.com/a/47866572

CMakeLists.txt
StrikerRUS: Yeah, I understand.

cmake/modules/FindASan.cmake
StrikerRUS: OK, but **why** do we need to hide it?.. I think building with GTest will be done by quite experienced users so it's OK to have some additional variables to show in GUI.
Maybe just omit this part to simplify the code?

hcho3:
I'm on vacation until March 3. I will revisit this PR when I am back.

cmake/modules/FindASan.cmake
hcho3: Sure, let me try removing it and see if the sanitizer still works.

CMakeLists.txt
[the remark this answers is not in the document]
hcho3: Actually, this has to be `set`, since `option` is meant to be Boolean: https://stackoverflow.com/a/36359003

cmake/Sanitizer.cmake
hcho3: Added

CMakeLists.txt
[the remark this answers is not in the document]
StrikerRUS: OK, I see. Thanks for taking a try!
I didn't realize that `OPTION` is only a "shortcut syntax" for boolean `SET` with `CACHE`:
 > Since bool cached variables are so common for builds, there is a shortcut syntax for making one using `option`:
    https://hsf-training.github.io/hsf-training-cmake-webpage/05-variables/index.html#cached-variables

StrikerRUS:
@hcho3 Could you please sync your branch with current `master`?

image


I'm not sure but guess this is the reason of failing Dask tests.

cc @jameslamb

cmake/Sanitizer.cmake
```diff
@@ -0,0 +1,69 @@
...
+#
+# Usage of this module:
+#  enable_sanitizers("address;leak")
+
+# Add flags
+macro(enable_sanitizer sanitizer)
+  if(${sanitizer} MATCHES "address")
+    find_package(ASan REQUIRED)
+    set(SAN_COMPILE_FLAGS "${SAN_COMPILE_FLAGS} -fsanitize=address")
+    if (ASan_FOUND)
```
StrikerRUS, line 11: I believe we can omit these `if (xxx_FOUND)`s due to recently added `REQUIRED` flag to `find_package` commands.

CMakeLists.txt
```diff
@@ -6,6 +6,12 @@ OPTION(USE_HDFS "Enable HDFS support (EXPERIMENTAL)" OFF)
 OPTION(USE_TIMETAG "Set to ON to output time costs" OFF)
 OPTION(USE_CUDA "Enable CUDA-accelerated training (EXPERIMENTAL)" OFF)
 OPTION(USE_DEBUG "Set to ON for Debug mode" OFF)
+OPTION(USE_SANITIZER "Use santizer flags" OFF)
+OPTION(SANITIZER_PATH "Path to sanitizer libs")
```
StrikerRUS, line 10: Based on https://github.com/microsoft/LightGBM/pull/3555#discussion_r588814820, this should be `SET`, right?

CMakeLists.txt
```diff
@@ -6,6 +6,12 @@ OPTION(USE_HDFS "Enable HDFS support (EXPERIMENTAL)" OFF)
 OPTION(USE_TIMETAG "Set to ON to output time costs" OFF)
 OPTION(USE_CUDA "Enable CUDA-accelerated training (EXPERIMENTAL)" OFF)
 OPTION(USE_DEBUG "Set to ON for Debug mode" OFF)
+OPTION(USE_SANITIZER "Use santizer flags" OFF)
+SET(SANITIZER_PATH "Path to sanitizer libs")
```
StrikerRUS, line 10:
```suggestion
SET(SANITIZER_PATH "" CACHE STRING "Path to sanitizer libs")
```

StrikerRUS (approved):
Many thanks for the great help!
LGTM for CMake part except one minor suggestion below.

Someone please check cpp part. Maybe @shiyu1994 @btrotta ?

shiyu1994:
Tested the BUILD_CPP_TEST option. The output is just fine, except that there's no actual tests. But as mentioned above, adding cpp tests will be a future work.

Thanks @hcho3 .

StrikerRUS:
@hcho3 Sorry for the inconvenience, but could you please sync the branch one more time, so that tests will pass and we will be able to merge the PR? We had a lot of CI issues this and past week 🙁 .

jameslamb:
Thank you so much for the help @hcho3 ! Awesome addition.

PR #3558, opened 2020-11-12
[ci] Use different PATH based on Rtools version

jameslamb:
This is another attempt at fixing #3469 . I looked through the source code of `r-lib/actions/setup-r` to see how they handle setting up R on GitHub Actions Windows runners, and noticed that they choose different additions to PATH depending on the Rtools version.

https://github.com/r-lib/actions/blob/9598b8eeb6d88de7d76d580d84443542bbfdffce/setup-r/src/installer.ts#L318-L324

```
  if (rtools4) {
    core.addPath(`C:\\rtools40\\usr\\bin`);
    core.addPath(`C:\\rtools40\\mingw64\\bin`);
  } else {
    core.addPath(`C:\\Rtools\\bin`);
    core.addPath(`C:\\Rtools\\mingw_64\\bin`);
  }
```

I also noticed that they use backslashes for their paths. Before this PR, we were mixing slash types (e.g. `C:\rtools40/usr/bin`).

I don't understand how either of those things would lead to the random failures we've seen, but that action is maintained by some very experienced R developers, and with the changes on this PR I was able to get the problematic MINGW + R  3.6 job to succeed 7 times: https://github.com/jameslamb/LightGBM/pull/37/checks?check_run_id=1388996512

image

### DIFF
--- a/.ci/test_r_package_windows.ps1
+++ b/.ci/test_r_package_windows.ps1
@@ -36,14 +36,14 @@ if ($env:R_MAJOR_VERSION -eq "3") {
   # Rtools 3.x has to be installed at C:\Rtools\
   #     * https://stackoverflow.com/a/46619260/3986677
   $RTOOLS_INSTALL_PATH = "C:\Rtools"
+  $env:RTOOLS_BIN = "$RTOOLS_INSTALL_PATH\bin"
   $env:RTOOLS_MINGW_BIN = "$RTOOLS_INSTALL_PATH\mingw_64\bin"
-  $env:RTOOLS_PATHS = "$RTOOLS_INSTALL_PATH\bin;" + "$env:RTOOLS_MINGW_BIN;"
   $env:RTOOLS_EXE_FILE = "rtools35-x86_64.exe"
   $env:R_WINDOWS_VERSION = "3.6.3"
 } elseif ($env:R_MAJOR_VERSION -eq "4") {
   $RTOOLS_INSTALL_PATH = "C:\rtools40"
+  $env:RTOOLS_BIN = "$RTOOLS_INSTALL_PATH\usr\bin"
   $env:RTOOLS_MINGW_BIN = "$RTOOLS_INSTALL_PATH\mingw64\bin"
-  $env:RTOOLS_PATHS = "$RTOOLS_INSTALL_PATH\usr\bin;" + "$env:RTOOLS_MINGW_BIN;"
   $env:RTOOLS_EXE_FILE = "rtools40-x86_64.exe"
   $env:R_WINDOWS_VERSION = "4.0.3"
 } else {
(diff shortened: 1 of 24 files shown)


jameslamb:
well that's a new error....

```text
R CMD check build logs:
* installing *source* package 'lightgbm' ...
** using staged installation
checking whether MM_PREFETCH works...yes
checking whether MM_MALLOC works...yes
** libs
mingw32-make: *** No rule to make target '$(CXX11STD)''.  Stop.
ERROR: compilation failed for package 'lightgbm'
* removing 'C:/tmp-r-cmd-check/lightgbm.Rcheck/lightgbm'
Looking for issues with R CMD check results
ERRORs have been found by R CMD check!
```

StrikerRUS:
> well that's a new error....

And in R4 environment. Previous error was in 3.6 one, if I'm not mistaken.

.ci/test_r_package_windows.ps1
```diff
@@ -36,12 +36,14 @@ if ($env:R_MAJOR_VERSION -eq "3") {
...
-  $env:RTOOLS_MINGW_BIN = "$RTOOLS_INSTALL_PATH/mingw_64/bin"
+  $env:RTOOLS_MINGW_BIN = "$RTOOLS_INSTALL_PATH\mingw_64\bin"
+  $env:RTOOLS_PATHS = "$RTOOLS_INSTALL_PATH\usr\bin;" + "$env:RTOOLS_MINGW_BIN;"
   $env:RTOOLS_EXE_FILE = "rtools35-x86_64.exe"
   $env:R_WINDOWS_VERSION = "3.6.3"
 } elseif ($env:R_MAJOR_VERSION -eq "4") {
   $RTOOLS_INSTALL_PATH = "C:\rtools40"
-  $env:RTOOLS_MINGW_BIN = "$RTOOLS_INSTALL_PATH/mingw64/bin"
+  $env:RTOOLS_MINGW_BIN = "$RTOOLS_INSTALL_PATH\mingw64\bin"
+  $env:RTOOLS_PATHS = "$RTOOLS_INSTALL_PATH\bin;" + "$env:RTOOLS_MINGW_BIN;"
```
jameslamb, line 46: ha! I did these backwards. Bad copy-paste, because this block is

```
if (rtools3) {
   ...
} elseif (rtool4) {
   ...
}
```

And the `r-lib/actions` on is

```
if (rtools4) {
   ...
} elseif (rtool4) {
   ...
}
```

jameslamb:
I'm going to try re-running 10 times, lets see how many succeed  / fail

1. [logs](https://github.com/microsoft/LightGBM/pull/3558/checks?check_run_id=1393722126) ✅ 
2.
3.
4.
5.
6.
7.
8.
9.
10.

jameslamb:
@StrikerRUS the R jobs passed 10 consecutive times with this change!

After the first re-run, I started pushing empty commits to re-trigger this because I realized there might be some things that are deterministic based on the commit you're building (like which runner you end up on or something).

I HOPE that means that the combination of this change and my other attempts have solved #3469

StrikerRUS:
@jameslamb 

> After the first re-run, I started pushing empty commits to re-trigger this because I realized there might be some things that are deterministic based on the commit you're building (like which runner you end up on or something).

Brilliant idea!

.ci/test_r_package_windows.ps1
```diff
@@ -36,12 +36,14 @@ if ($env:R_MAJOR_VERSION -eq "3") {
   # Rtools 3.x has to be installed at C:\Rtools\
   #     * https://stackoverflow.com/a/46619260/3986677
   $RTOOLS_INSTALL_PATH = "C:\Rtools"
-  $env:RTOOLS_MINGW_BIN = "$RTOOLS_INSTALL_PATH/mingw_64/bin"
+  $env:RTOOLS_MINGW_BIN = "$RTOOLS_INSTALL_PATH\mingw_64\bin"
+  $env:RTOOLS_PATHS = "$RTOOLS_INSTALL_PATH\bin;" + "$env:RTOOLS_MINGW_BIN;"
```
StrikerRUS, line 40: Looks like `$env:RTOOLS_PATHS` is only used for `PATH` enhancement. I think we can add `$env:RTOOLS_MINGW_BIN` directly to `PATH` without proxy `$env:RTOOLS_PATHS` strings storage. And rename `$env:RTOOLS_PATHS` to `$env:RTOOLS_BIN` or something like that. I believe these changes will make the scrip clearer.

jameslamb: I don't think that's any better or worse. But since you have a preference and I don't, I'll change it to that.

jameslamb: updated in https://github.com/microsoft/LightGBM/pull/3558/commits/16da5c0539caf26942015995a676508f69096e67

StrikerRUS: Thanks!

StrikerRUS (approved):

PR #3565, opened 2020-11-15
[python] mark 3.1 version is the last one with Python 2 support

StrikerRUS:
Refer to https://github.com/microsoft/LightGBM/pull/3484#issuecomment-723484953.


guolinke (approved):

jameslamb (approved):
Exciting!

I'd like to add that I think after 3.1.0, we should just go straight to only supporting `python >= 3.6`, since Python 3.5  is already end of life: https://www.python.org/downloads/release/python-3510/#

StrikerRUS:
@jameslamb 
> to only supporting python >= 3.6,

Already done! 🙂 

#3395

Issue #3519, opened 2020-11-02
[RFC] Travis adding much stricter limits on free users


jameslamb:
Noticed in https://github.com/dask/community/issues/107 this morning. Travis is putting much much stricter limits on free-tier use of its CI services.

https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing

> For those of you who have been building on public repositories (on travis-ci.com, with no paid subscription), we will upgrade you to our trial (free) plan with a 10K credit allotment (which allows around 1000 minutes in a Linux environment).

> We will be offering an allotment of OSS minutes that will be reviewed and allocated on a case by case basis. Should you want to apply for these credits please open a request with Travis CI support stating that you’d like to be considered for the OSS allotment. 

The blog post also makes it sound like macOS jobs will no longer be available for free-tier users, but honestly the blog post's discussion of that is very confusing.

I'm not sure yet what we should do with out existing Travis jobs, need to do some more research on how long they take to run and what that means in relation to these limits. :tired_face:

StrikerRUS:
Thanks a lot for bringing this up @jameslamb !

Yeah, seems that macOS builds will be paid...
Also, please notice we use `travis-ci.org`, not `.com`. Looks like this also matters.

StrikerRUS:
https://docs.travis-ci.com/user/billing-faq#what-if-i-am-building-open-source
> Each of the Travis CI Plans can have an amount of special OSS credits per month assigned to run builds only on public repositories. To find out more about it please contact the Travis CI support team.

Very "informative"! 😃

StrikerRUS:
> Also, please notice we use `travis-ci.org`, not `.com`. Looks like this also matters.

OK, `org` will be archived 31st December: https://travis-ci.community/t/org-to-com-migration-deadline/10260.

dgw:
> Love for OSS has gone: https://github.com/travis-ci/travis-web/pull/2569/files

I found this thread while googling for more info about the "OSS only credits". That linked change is disheartening.

Looks like my own OSS project will need to discuss the pros and cons of switching over to GitHub Actions instead of migrating our travis-ci.org repos to travis-ci.com. If you go through the process of requesting OSS credits, I'd love to see the results here. It would help us with our own decision-making process!

jameslamb:
Just got this error on a personal project of mine that runs on travis-ci.org

[...]

PR #3572, opened 2020-11-16
[ci] Remove Python 2.7 CI jobs

jameslamb:
Now that v3.1.0 has been released (#3484) and marked as the final release that supports Python 2.7 (#3565), I'd like to propose removing the Python 2.7 CI jobs.

This would unblock the work to move `dask-lightgbm` into this project (https://github.com/microsoft/LightGBM/pull/3515#issuecomment-728232698) and could help us save some credits on Travis while we deal with the issues caused by Travis's new announcements around ending support for open source projects (#3519).

I know there are other code-level things we could do to simplify `lightgbm` now that Python 2.7 support is being dropped, but I think cutting out the CI jobs is valuable enough that it's worth doing as a simple, standalone pull request.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,6 @@ env:
     - PYTHON_VERSION=3.8
   matrix:
     - TASK=regular PYTHON_VERSION=3.6
-    - TASK=sdist PYTHON_VERSION=2.7
     - TASK=bdist
     - TASK=if-else
     - TASK=lint
--- a/.vsts-ci.yml
+++ b/.vsts-ci.yml
@@ -115,9 +115,6 @@ jobs:
       regular:
         TASK: regular
         PYTHON_VERSION: 3.6
-      sdist:
-        TASK: sdist
-        PYTHON_VERSION: 2.7
       bdist:
         TASK: bdist
   steps:


StrikerRUS (changes requested):
You can't just remove `sdist` job. This job is very important as it emulates installation from PyPI source.

Instead, simply remove `PYTHON_VERSION=2.7` part.

StrikerRUS:
Also, I'd like to wait some time before actually dropping Python 2.7, because we might need to produce hotfix `3.1.1` version with Python 2.7 support.

jameslamb:
> You can't just remove `sdist` job. This job is very important as it emulates installation from PyPI source.
> 
> Instead, simply remove `PYTHON_VERSION=2.7` part.

oh ok, does this mean that right now on `master` we don't have another job that emulates "installation from PyPi source" for Python 3.x?

jameslamb:
> Also, I'd like to wait some time before actually dropping Python 2.7, because we might need to produce hotfix `3.1.1` version with Python 2.7 support.

alright I can just close this for now, I guess.

StrikerRUS:
> oh ok, does this mean that right now on `master` we don't have another job that emulates "installation from PyPi source" for Python 3.x?

We have one for macOS
https://github.com/microsoft/LightGBM/blob/0b5e60dd10338daa821bbb38cc022df8c0bc857e/.vsts-ci.yml#L81-L82

PR #3573, opened 2020-11-17
[ci] remove conda from R CI jobs

jameslamb:
On https://github.com/microsoft/LightGBM/pull/3405#issuecomment-728353296, I see the Windows CI jobs on GitHub Actions failing with this output

```text
==> WARNING: A newer version of conda exists. <==
  current version: 4.8.3
  latest version: 4.9.2

Please update conda by running

    $ conda update -n base -c defaults conda

#
# To activate this environment, use
#
#     $ conda activate test
#
# To deactivate an active environment, use
#
#     $ conda deactivate
```

I'm worried that this is another case where programs writing to stderr in GitHub Actions Windows jobs causes those jobs to fail. I don't know if this recently started failing because of software updates in GitHub Actions images (https://github.com/actions/virtual-environments/commits/main) or something else.

This PR proposes just updating `conda` to the latest version every time, to hopefully avoid this warning.

### DIFF
--- a/.ci/setup.sh
+++ b/.ci/setup.sh
@@ -48,8 +48,10 @@ else  # Linux
     fi
 fi
 
-if [[ $TRAVIS == "true" ]] || [[ $GITHUB_ACTIONS == "true" ]] || [[ $OS_NAME == "macos" ]]; then
-    sh conda.sh -b -p $CONDA
+if [[ "${TASK:0:9}" != "r-package" ]]; then
+    if [[ $TRAVIS == "true" ]] || [[ $OS_NAME == "macos" ]]; then
+        sh conda.sh -b -p $CONDA
+    fi
+    conda config --set always_yes yes --set changeps1 no
+    conda update -q -y conda
 fi
-conda config --set always_yes yes --set changeps1 no
-conda update -q -y conda
--- a/.ci/test.sh
+++ b/.ci/test.sh
@@ -8,6 +8,11 @@ elif [[ $OS_NAME == "linux" ]] && [[ $COMPILER == "clang" ]]; then
     export CC=clang
 fi
 
+if [[ "${TASK:0:9}" == "r-package" ]]; then
+    bash ${BUILD_DIRECTORY}/.ci/test_r_package.sh || exit -1
+    exit 0
+fi
+
 conda create -q -y -n $CONDA_ENV python=$PYTHON_VERSION
 source activate $CONDA_ENV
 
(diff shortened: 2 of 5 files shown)


jameslamb:
UPDATE I see the same failures on the Windows jobs in #3515 too:https://github.com/microsoft/LightGBM/pull/3515/checks?check_run_id=1408108173

But not on builds from last night. https://github.com/microsoft/LightGBM/actions/runs/365273180

So whatever changed, I think it was in the last 24 hours.

jameslamb:
Ok, that did not work. https://github.com/microsoft/LightGBM/pull/3573/checks?check_run_id=1409528379

Trying to upgrade to the action to 2.0 next.

jameslamb:
Seems like the combination of always updating `conda` + using v2 of this action fixes it

StrikerRUS (approved):

jameslamb:
@StrikerRUS let's move https://github.com/microsoft/LightGBM/pull/3405#issuecomment-728642475 here, since #3405 already has a lot of comments.

That's a good point! There is actually an `r-lib/setup-pandoc`  action we could try: https://github.com/r-lib/actions/tree/master/setup-pandoc

Let me try on my fork tonight. If that works, I'll just change this PR.

jameslamb:
Seems ok on my fork!

https://github.com/jameslamb/LightGBM/pull/38/checks?check_run_id=1409970898

I just pushed some changes that remove `conda` completely for R jobs

.ci/test.sh
```diff
@@ -8,6 +8,11 @@ elif [[ $OS_NAME == "linux" ]] && [[ $COMPILER == "clang" ]]; then
     export CC=clang
 fi
 
+if [[ "${TASK:0:9}" == "r-package" ]]; then
+    bash ${BUILD_DIRECTORY}/.ci/test_r_package.sh || exit -1
+    exit 0
+fi
```
jameslamb, line 14: I had to move this up so it happens before `conda create`

.ci/setup.sh
```diff
@@ -48,8 +48,10 @@ else  # Linux
     fi
 fi
 
-if [[ $TRAVIS == "true" ]] || [[ $GITHUB_ACTIONS == "true" ]] || [[ $OS_NAME == "macos" ]]; then
-    sh conda.sh -b -p $CONDA
+if [[ "${TASK:0:9}" != "r-package" ]]; then
```
jameslamb, line 51: @StrikerRUS I was torn between doing this `!=` thing of just adding  this:

```shell
if [[ "${TASK:0:9}" != "r-package" ]]; then
    exit 0
fi
```

Do you have a strong preference?

jameslamb:
I've requested a new approval, @StrikerRUS , since I totally changed this PR since the last time you reviewed and approved

.ci/setup.sh
StrikerRUS: @jameslamb 

> or just adding this before the conda steps:

I think the second way will be more readable.

StrikerRUS (approved):
LGTM!

.ci/setup.sh
StrikerRUS: Never mind! I changed my mind 😄

jameslamb: haha ok. Easy to change if you ever want to.

PR #3574, opened 2020-11-17
[R-package] Use command-line args to customize CMake builds (fixes #2441)

### LINKED ISSUE #2441: [R-package] use commandArgs instead of hardcoded stuff in the installation script

jameslamb:
This PR tries to fix #2441. As of this PR, users will no longer need to manually edit `R-package/src/install.libs.R` for customized CMake-based builds of the R package.

```shell
# force use of MINGW
Rscript build_r.R --skip-mingw

# force use of MSYS2
Rscript build_r.R --skip-msys2

# build for GPU
Rscript build_r.R --skip-gpu
```

I'm proposing that we do this by adding some string replacement in `build_r.R`. That way, if people have existing code that automatically changes `install.libs.R`, it won't be broken by this change.

I think this will make the user experience with `{lightgbm}` better, and sets up a better pattern if we want to offer more customization in the future (like a `--use-cuda` flag).


jameslamb:
To know if this worked for sure, I think we'll have to wait until #3573 is merged, so Windows CI jobs pass

R-package/README.md
```diff
@@ -150,15 +150,24 @@ cd LightGBM
...
 
-The `build_r.R` script builds the package in a temporary directory called `lightgbm_r`. It will destroy and recreate that directory each time you run the script.
+The `build_r.R` script builds the package in a temporary directory called `lightgbm_r`. It will destroy and recreate that directory each time you run the script. That scripts supports the following command-line options:
+
+- `--skip-install`: Build the package tarball, but do not install it.
+- `--use-gpu`: Build a GPU-enabled versioon of the library
+- `--use-mingw`: Force the use of MinGW toolchain, regardless of R version.
+- `--use-msys2`: Force the use of MSYS2 toolchain, regardless of R version.
 
 Note: for the build with Visual Studio/VS Build Tools in Windows, you should use the Windows CMD or Powershell.
```
StrikerRUS, line 160:
```suggestion
Note: for the build with Visual Studio/VS Build Tools in Windows, you should use the Windows CMD or PowerShell.
```

StrikerRUS (commented):
Changes in CI script looks OK to me. Just some typos below.

R-part reviewers @guolinke @Laurae2 , please pay attention to the following architectural decision.

> I'm proposing that we do this by adding some string replacement in `build_r.R`.

guolinke (approved):
LGTM

jameslamb:
thanks for the review! I'll merge in `master` to this, and then merge this if it builds successfully.

Since this is non-breaking, I think it's ok for it to be part of a patch (3.1.1) release. Yes it is technically a "feature" since it makes the installation flow for the R package easier, but it's not really a "feature" from the perspective of LightGBM as an entire project. So I think it's ok to bundle in a patch release, and better to do that than to have too many open PRs all waiting until after 3.1.1.

PR #3581, opened 2020-11-20
[python] Drop Python 2 support

StrikerRUS:
Do not include in any **patch** `3.1.x` release. Should be merged only in `3.2.0` version.


docker/dockerfile-r
```diff
@@ -4,7 +4,9 @@ FROM rocker/verse:${R_VERSION}
 WORKDIR /lgbm
 
 RUN apt-get update && \
-    apt-get install -y build-essential cmake && \
+    apt-get install -y --no-install-recommends \
+        build-essential \
+        cmake && \
```
jameslamb, line 9: I support this change, but it seems unrelated. Could you put it in a separate PR that we could approve and merge more quickly (since this PR will have to wait for a while until we know if the next release will be 3.1.1 or 3.2.0)?

tests/python_package_test/utils.py
```diff
@@ -1,23 +1,7 @@
...
-        cache = {}
-
-        def _lru_wrapper(user_function):
-            def wrapper(*args, **kwargs):
-                arg_key = (args, tuple(kwargs.items()))
-                if arg_key not in cache:
-                    cache[arg_key] = user_function(*args, **kwargs)
-                return cache[arg_key]
-            return wrapper
-        return _lru_wrapper
```
jameslamb, line 20: I'm very happy to see this go away :grinning:

jameslamb (changes requested):
These changes look great to me, thanks! I left one suggestion for an unrelated change to be moved.

docker/dockerfile-r
StrikerRUS: OK, sure!

jameslamb (approved):
Really well-done! I think you got everything that's dependent on Python 2. I can't think of anything else.

Tagging @aldanor @synapticarbors as maintainers of https://github.com/conda-forge/lightgbm-feedstock, to let them know that the next LightGBM minor release (3.2.0) will drop Python 2 support.

StrikerRUS:
@jameslamb 
> I can't think of anything else.

One more thing is string formatting unification (`str.format()` at least or cute f-strings). But I think it is a great candidate for Hacktoberfest 2021. 🙂

python-package/lightgbm/compat.py
```diff
@@ -162,10 +105,3 @@ def _check_sample_weight(sample_weight, X, dtype=None):
     _LGBMAssertAllFinite = None
     _LGBMCheckClassificationTargets = None
     _LGBMComputeSampleWeight = None
-
-
-# DeprecationWarning is not shown by default, so let's create our own with higher level
-class LGBMDeprecationWarning(UserWarning):
```
aldanor, line 168: Could there be some user code downstream catching/filtering this one?

StrikerRUS: Sorry, could you elaborate on what code you'd expect?

aldanor: It's probably ok, I don't think anything needs to be done :) What I meant is that you could say it's *technically* a breaking change since there can be user code like below:

```python
from lightgbm.compat import LGBBDeprecationWarning
import warnings
warnings.simplefilter('ignore', LGBMDeprecationWarning)
```

StrikerRUS: Ah, OK, got it!

Yeah, after this PR is merged, one should change 
```
from lightgbm.compat import LGBMDeprecationWarning
```
to 

```
from lightgbm.basic import LGBMDeprecationWarning
```

But we are planing to include this PR in `3.2.0` (not `3.1.1` patch) version and it has `breaking` label in the changelog. So, I believe it is OK...

StrikerRUS:
@guolinke Are you OK with this PR?

guolinke (approved):
LGTM

Issue #3210, opened 2020-07-07
RFC: version changes + more frequent releases


jameslamb:
I'd like to open this request for comment to discuss a proposal.

After releasing v 3.0.0 (#3071 ), I'd like to propose that we use 4-part version numbers for language wrappers, broken down like this:

![image](https://user-images.githubusercontent.com/7608904/86708875-f9a1d900-bfde-11ea-8285-6cad0b036b7e.png)

So, for example, if you see version `3.1.0.8` of the R package, that means *"the 8th released version of the R package which wraps `LightGBM` version 3.1.0"*.

## Example

The examples below don't propose that every new merge to `master` becomes a release, but the changes below are examples used to show what might cause different components of a 4-part version number to change.

**Event 1: 3.0.0 is release**

* `LightGBM` version set to 3.0.0
* `lightgbm` (Python) 3.0.0.0 released to PyPi
* `{lightgbm}` (R) 3.0.0.0 released to CRAN
* `LightGBM` (lib for .NET extensions) 3.0.0.0 released to NuGet

**Event 2: bug fix to LightGBM, like fixing #3209**

* `LightGBM` version set to 3.0.1
* `lightgbm` (Python) 3.0.1.0 released to PyPi
* `{lightgbm}` (R) 3.0.1.0 released to CRAN
* `LightGBM` (lib for .NET extensions) 3.0.1.0 released to NuGet

**Event 3: bug fix in {lightgbm} (R), like #3117**

* `{lightgbm}` (R) 3.0.1.1 released to CRAN

**Event 4: LightGBM adds a new type of boosting, like #2644**

* `LightGBM` version set to 3.1.0
* `lightgbm` (Python) 3.1.0.0 released to PyPi
* `{lightgbm}` (R) 3.1.0.0 released to CRAN
* `LightGBM` (lib for .NET extensions) 3.1.0.0 released to NuGet

## How this makes LightGBM better

This approach would allow us to release fixes to individual components of LightGBM more frequently.

This would allow us to avoid the current situation, where the PyPi package (for example), has not had an update in 7 months: https://pypi.org/project/lightgbm/#history. More frequent updates allow our users to rely on package managers more, instead of building from GitHub, which I think is a better user experience.

Releasing more frequently would also reduce the gap between the current state of this repo and the documentation at https://lightgbm.readthedocs.io/en/latest/, so that that documentation is more likely to answer a user's questions accurately.

[...]

PR #3586, opened 2020-11-22
WIP: [R-package] new release to fix CRAN error on 32-bit Windows

jameslamb:
The R package is currently at risk of being dropped from CRAN, because one of the learning-to-rank tests is failing on 32-bit Windows + R 3.6.

image

#3585 documents the problem, and describes a "quick fix" to get the package fixed on CRAN as well as a "long-term fix" that will better solve this problem. This PR has the "quick fix".

### Notes for Reviewers

I branched off of `master` at tag `v3.1.0`, to avoid including anything that has been merged in since that release. I think we should keep this PR open until `{lightgbm}` is fully accepted on CRAN with no issues. Once that happens and this can be merged, I'll update it to latest `master` and undo the changes to `VERSION.txt`.

I found that I'm able to reproduce the errors from CRAN using R Hub (https://github.com/microsoft/LightGBM/issues/3585#issuecomment-731654494), and I've confirmed with those same steps that this branch's changes resolve the problem.

~Once either @guolinke or @Laurae2 or @StrikerRUS approves, I'll submit 3.1.0.1 to CRAN with an explanation about why we are submitting again so soon.~ https://github.com/microsoft/LightGBM/pull/3586#issuecomment-731669043


R-package/tests/testthat/test_learning_to_rank.R
```diff
@@ -48,14 +49,17 @@ test_that("learning-to-rank with lgb.train() works as expected", {
     }
     expect_identical(sapply(eval_results, function(x) {x$name}), eval_names)
     expect_equal(eval_results[[1L]][["value"]], 0.775)
-    if (!ON_SOLARIS) {
+    if (!(ON_SOLARIS || ON_32_BIT_WINDOWS)) {
```
guolinke, line 52: maybe it is better to relax the constraint values? in case there are more tests may fail

jameslamb: do you mean change the numeric tolerance?

I think it's better to just completely skip these tests for now to be sure we don't get kicked off of CRAN or waste their time with too many submissions. I don't know the exact amount that the values are different by, and I'm not sure I can figure that out in a way that's guaranteed to be the same as CRAN's environment.

By the next LightGBM release, we can try to completely fix this instead of just skipping the tests.

To be clear, in GitHub Actions here these tests are still running on Linux, Mac, and 64-bit Windows, with multiple different compilers, and for R 3.6 and R 4.0. So this feature is still being thoroughly tested.

R-package/cran-comments.md
```diff
@@ -1,9 +1,21 @@
 # CRAN Submission History
 
-## v3.0.1 - Submission 1 - (November 15, 202)
+## v3.1.0.1 - Submission 1 - (November 21, 2020)
+
+### Cran response
```
StrikerRUS, line 5:
```suggestion
### CRAN response
```

R-package/cran-comments.md
```diff
@@ -1,9 +1,21 @@
 # CRAN Submission History
 
-## v3.0.1 - Submission 1 - (November 15, 202)
+## v3.1.0.1 - Submission 1 - (November 21, 2020)
```
StrikerRUS, line 3: > I'll submit 3.1.0.1 to CRAN with an explanation about why we are submitting again so soon.

I think it can be `3.1.1` regular release for the whole project as we haven't agreed on #3210 yet (4-part versioning).

StrikerRUS (approved):

R-package/cran-comments.md
jameslamb: I'm fine with doing a 3.1.1 release of the whole project, but that generates some extra work for you and @guolinke , so I thought it would be simpler to do an R-only version change.

adding a 4th version number just for the purpose of getting to CRAN is what we did in the last submission cycle (https://github.com/microsoft/LightGBM/pull/3338#issuecomment-698724866).

Let's let @guolinke be the tie-break (choose either 3.1.0.1 or 3.1.1 for the new CRAN submission). If it's 3.1.1, I can update this PR accodingly.

StrikerRUS:
> I'll submit 3.1.0.1 to CRAN with an explanation about why we are submitting again so soon.

BTW, shouldn't we wait for Sanitizers and Valgrind results to not bother CRAN maintainers with frequent submissions?

R-package/cran-comments.md
StrikerRUS: > adding a 4th version number just for the purpose of getting to CRAN is what we did in the last submission cycle

Please note that CI cannot be run due to merge conflicts. You simply cannot generate artifact that should be uploaded to CRAN (you can do it only manually on your machine, but then what was the reason to add `r_artifacts.yml` workflow?..)

jameslamb:
> > I'll submit 3.1.0.1 to CRAN with an explanation about why we are submitting again so soon.
> 
> BTW, shouldn't we wait for Sanitizers and Valgrind results to not bother CRAN maintainers with frequent submissions?

I think that they might have already run. Unfortunately I don't know how to tell for sure. When the package was posted on November 18, under the "check results" at https://cran.r-project.org/web/checks/check_results_lightgbm.html you could still see an entry for valgrind, but it was the old entry.

Now that entry has been removed, and there's no mention of valgrind.

But now that you make me think about it, you're right....let's wait to re-submit until they email me or @guolinke (I'm not sure who will get the email this time) telling us what we have to do to stay on CRAN.

R-package/cran-comments.md
jameslamb: > you can do it only manually on your machine, but then what was the reason to add r_artifacts.yml workflow?

So that checks that need to be run manually, like the Solaris ones, don't have to always be done by me, in case I'm unavailable. Anyone could generate the package on GitHub Actions with that comment, then upload it to https://builder.r-hub.io/.

I just looked at the list of recently-merged PRs and it doesn't seem like there are any new features, just small fixes, so I'm fine with making this 3.1.1. I'll update this branch to the latest `master` and change the version.

guolinke: we can update it to 3.1.1, but only update it cran.

StrikerRUS: > we can update it to 3.1.1, but only submit it cran.

I guess it will be confusing for users. And we will have a bunch of issues with a content like "When `3.1.1` version will be released at PyPI?"

guolinke: I just check the commit log, there are only “maintenance” change, not any bug fixes or new features. Therefore, I don't think we should release a new version.

guolinke: besides, releasing a new version in such a short time seems we have an emergence bug fix.

jameslamb: > I don't think we should release a new version.

Ok, then I think we should use 3.1.0.1 as the version number for CRAN. We have to increment the version number when we re-submit.

By the way @guolinke , has CRAN emailed you? I think they should be done running all their checks.

There are 12 checks done at  https://cran.r-project.org/web/checks/check_results_lightgbm.html

image

which is how many all other packages have (for example, `{jsonlite}`)

image

But I agree with @StrikerRUS 's comment in https://github.com/microsoft/LightGBM/pull/3586#issuecomment-731668240 that we shouldn't submit again until CRAN asks.

guolinke:
@jameslamb I didn't receive any emails from CRAN yet.

R-package/cran-comments.md
StrikerRUS: > I just check the commit log, there are only “maintenance” change, not any bug fixes or new features. 

Today we merged 2 bugfixes. And there can be more "fix" or "features" PRs if we wait some days for CRAN response.

image

jameslamb:
I just moved the test changes in this PR to #3588 , so we can get that fix onto `master` soon and have other PRs integrate it.

I've marked this PR work-in-progress and changed the title, since it will have to stay open until CRAN asks us for a new submission

guolinke:
I think we can set a timeline for this, for example, if CRAN still doesn't ask us in two weeks, we can directly release 3.1.1, and submit it to cran as well.

jameslamb:
> I think we can set a timeline for this, for example, if CRAN still doesn't ask us in two weeks, we can directly release 3.1.1, and submit it to cran as well.

I agree! 

If CRAN hasn't asked us to resubmit by the end of this week, I'd feel comfortable documenting that in the repo and publicizing it on social media. Since even if they surprise us in a few days and ask for a resubmission, we're very confident that we know how to fix it (unlike last time when I wasn't really sure).

StrikerRUS:
I think CRAN team hasn't plans to response this time. Can we start releasing `3.1.1` version?
I'd like to include 
- [ ] #3592
- [ ] #3598

in `3.1.1` as well.

jameslamb:
> I think CRAN team hasn't plans to response this time. Can we start releasing `3.1.1` version?
> I'd like to include
> 
> * [ ]  #3592
> * [ ]  #3598
> 
> in `3.1.1` as well.

I agree! Unfortunately I opened this PR as a branch from my fork (because originally it was just the test fix). I'm going to close this PR and open the release 3.1.1 PR on a LightGBM branch we can all push to.

PR #3587, opened 2020-11-22
[ci] Fix missing git install in R artifact job (fixes #3567)

### LINKED ISSUE #3567: [ci] R artifact build job is broken
See https://github.com/microsoft/LightGBM/runs/1404264739?check_suite_focus=true

The job defined in https://github.com/microsoft/LightGBM/blob/master/.github/workflows/r_artifacts.yml is currently breaking, with this error

> To create a local Git repository instead, add Git 2.18 or higher to the PATH
Error: Input 'submodules' not supported when falling back to download using the GitHub REST API. To create a local Git repository instead, add Git 2.18 or higher to the PATH.

jameslamb:
This PR fixes #3567 . When third-party actions were updated in #3524 I think it broke the job that is used to create a source distribution of the CRAN package.

### DIFF
--- a/.github/workflows/r_artifacts.yml
+++ b/.github/workflows/r_artifacts.yml
@@ -12,6 +12,11 @@ jobs:
     runs-on: ubuntu-latest
     container: rocker/r-base
     steps:
+      - name: Install Git before checkout
+        shell: bash
+        run: |
+          apt-get update
+          apt-get install --no-install-recommends -y git
       - name: Checkout repository
         uses: actions/checkout@v2.3.4
         with:


.github/workflows/r_artifacts.yml
```diff
@@ -12,6 +12,11 @@ jobs:
     runs-on: ubuntu-latest
     container: rocker/r-base
     steps:
+      - name: Install Git before checkout
+        shell: bash
+        run: |
+          apt-get update
+          apt-get install --no-install-recommends -y git
```
jameslamb, line 19: /gha build r-artifacts

jameslamb: that comment created this build: https://github.com/microsoft/LightGBM/actions/runs/376731177

jameslamb: ✔️ this succeeded

image

StrikerRUS (approved):

PR #3592, opened 2020-11-24
Check max_bin, etc. match config when using binary

cyfdecyf:
To avoid binary file with non-matching parameters from being used.

### DIFF
--- a/include/LightGBM/dataset_loader.h
+++ b/include/LightGBM/dataset_loader.h
@@ -44,7 +44,7 @@ class DatasetLoader {
 
   void SetHeader(const char* filename);
 
-  void CheckDataset(const Dataset* dataset);
+  void CheckDataset(const Dataset* dataset, bool is_load_from_binary);
 
   std::vector<std::string> LoadTextDataToMemory(const char* filename, const Metadata& metadata, int rank, int num_machines, int* num_global_data, std::vector<data_size_t>* used_data_indices);
 
--- a/src/io/dataset_loader.cpp
+++ b/src/io/dataset_loader.cpp
@@ -186,6 +186,7 @@ Dataset* DatasetLoader::LoadFromFile(const char* filename, int rank, int num_mac
   data_size_t num_global_data = 0;
   std::vector<data_size_t> used_data_indices;
   auto bin_filename = CheckCanLoadFromBin(filename);
+  bool is_load_from_binary = false;
   if (bin_filename.size() == 0) {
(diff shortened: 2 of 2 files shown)


guolinke:
Thank you @cyfdecyf

guolinke (approved):

StrikerRUS:
@guolinke What about other Dataset params from https://lightgbm.readthedocs.io/en/latest/Parameters.html#dataset-parameters, e.g. `max_bin_by_feature`, `bin_construct_sample_cnt`? Should they also be checked?

guolinke:
@StrikerRUS yes, thank you.
@cyfdecyf can you add these parameters too?

cyfdecyf:
@guolinke OK. I will add checking for those parameters too.

StrikerRUS:
> I will add checking for those parameters too.

I think there are more params in [`Dataset Parameters`](https://lightgbm.readthedocs.io/en/latest/Parameters.html#dataset-parameters) section that can be checked. 


Also, please refer to https://github.com/microsoft/LightGBM/blob/3e8e24b374fbfe6522e39862a79c81b2bea8b259/src/c_api.cpp#L186.

cyfdecyf:
I focused only on checking parameters that are stored in `Dataset` https://github.com/microsoft/LightGBM/blob/3e8e24b374fbfe6522e39862a79c81b2bea8b259/include/LightGBM/dataset.h#L636

What's the desired way to check parameters that's not in `Dataset` now? e.g. `feature_pre_filter`, `data_random_seed` etc.
Storing them in the binary file and only load them for checking? That would be a bigger change than this PR.

> > I will add checking for those parameters too.
> 
> I think there are more params in [`Dataset Parameters`](https://lightgbm.readthedocs.io/en/latest/Parameters.html#dataset-parameters) section that can be checked.

guolinke:
@cyfdecyf yes, I think only check the stored parameter in the Dataset binary object is enough. 
you can ping me once the PR is ready.

cyfdecyf:
@guolinke I've already added checking for `max_bin_by_feature`, `bin_construct_sample_cnt` in https://github.com/microsoft/LightGBM/pull/3592/commits/9d19083aef6ccc131fcc5bd8833bb390c89a2db0.

I've checked parameters listed in https://lightgbm.readthedocs.io/en/latest/Parameters.html#dataset-parameters against private members in `Dataset`, but not sure if there's any other parameters that still need to be checked.

guolinke:
current implementation looks good to me. ping @StrikerRUS to confirm.

StrikerRUS:
OK, I agree!

Thanks a lot @cyfdecyf for contribution!

Issue #1863, opened 2018-11-21
bdist_wheel doesn't allow any flags


KOLANICH:
In fact some of these flags are unneeded, instead distutils prefs should be used. So maybe custom steps are unneeded too.

StrikerRUS:
@KOLANICH Please elaborate.

Are you unable to install lightgbm from Python wheel? What error is raised?

KOLANICH:
>Are you unable to install lightgbm from Python wheel?

I'm unable to build a wheel using the prebuilt library, since no custom command class is implemented for that command.

https://github.com/Microsoft/LightGBM/blob/d8d01e39843c4e93d6e9cb9d2c17bc6d138f4975/python-package/setup.py#L275L279

IMHO these "custom cmd classes" inherited from the original ones are too dirty hacky approaches, instead setuptools should be redesigned into a graph of stages with possibility to insert an own stage between any 2.

IMHO these

https://github.com/Microsoft/LightGBM/blob/d8d01e39843c4e93d6e9cb9d2c17bc6d138f4975/python-package/setup.py#L185L198

flags shouldn't be used, instead it should always use the lib prebuilt with CMake.

StrikerRUS:
> I'm unable to build a wheel using the prebuilt library

Please share your use-case where you need this.

> instead it should always use the lib prebuilt with CMake.

How to build, say, MPI version with `pip` in this case?

KOLANICH:
>Please share your use-case where you need this.

It is more convenient for me to use CMake for building. I don't use Visual Studio for building, instead I use MinGW-w64 and Ninja as a low level build tool, all the libraries are situated in unexpected folders.

>How to build, say, MPI version with pip in this case?

Just enable that option in CMakeCache.txt ?

StrikerRUS:
> . I don't use Visual Studio for building, instead I use MinGW-w64 and Ninja as a low level build tool

Just use `--mingw` option with correctly configured PATH to build with MinGW.
But why you need wheel file for this?


> Just enable that option in CMakeCache.txt ?

Please elaborate. At present, I can just execute `pip install lightgbm --install-option=--mpi`. How to do it without those options?

StrikerRUS:
Also, we already have the option to use **prebuilt library file**.
```
python setup.py install --precompile
```
 Have you seen it?

KOLANICH:
>But why you need wheel file for this?

[...]

PR #3594, opened 2020-11-24
[python] Allow to build Python wheel package

dolfinus:
Other ML libraries like XGBoost allows user to build wheel package to deliver it then to a cluster nodes.

LightGBM can be also build into wheel package, but with default options only (like no --gpu and no --cuda). If one wants to build custom wheel, he/she will face with ``option --gpu not recognized`` error and so on.

In this PR I've added custom `bdist_wheel` script which accepts just the same arguments as `install`.

### DIFF
--- a/python-package/README.rst
+++ b/python-package/README.rst
@@ -172,6 +172,12 @@ Run ``python setup.py install --bit32``, if you want to use 32-bit version. All 
 
 If you get any errors during installation or due to any other reasons, you may want to build dynamic library from sources by any method you prefer (see `Installation Guide <https://github.com/microsoft/LightGBM/blob/master/docs/Installation-Guide.rst>`__) and then just run ``python setup.py install --precompile``.
 
+
+Build Wheel File
+****************
+
+You can use ``python setup.py bdist_wheel`` instead of ``python setup.py install`` to build wheel file and use it for installation later. This might be useful for systems with restricted or completely without network access.
+
 Troubleshooting
 ---------------
 
--- a/python-package/setup.py
+++ b/python-package/setup.py
@@ -5,7 +5,6 @@ from __future__ import absolute_import
 import io
 import logging
 import os
-import shutil
 import struct
 import subprocess
 import sys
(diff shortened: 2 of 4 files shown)


python-package/README.rst
```diff
@@ -172,6 +172,12 @@ Run ``python setup.py install --bit32``, if you want to use 32-bit version. All
 
 If you get any errors during installation or due to any other reasons, you may want to build dynamic library from sources by any method you prefer (see `Installation Guide <https://github.com/microsoft/LightGBM/blob/master/docs/Installation-Guide.rst>`__) and then just run ``python setup.py install --precompile``.
 
+
+Build wheel from GitHub
+'''''''''''''''''''''''''
```
StrikerRUS, line 177:
```suggestion
Build Wheel File
****************

```

python-package/README.rst
```diff
@@ -172,6 +172,12 @@ Run ``python setup.py install --bit32``, if you want to use 32-bit version. All
...
+You can use ``python setup.py bdist_wheel`` instead of ``python setup.py install`` to build wheel package instead of installing it into system.
+
+
```
StrikerRUS, line 180:
```suggestion
You can use ``python setup.py bdist_wheel`` instead of ``python setup.py install`` to build wheel file and use it for installation later. This might be useful for systems with restricted or completely without network access.

```

python-package/setup.py
```diff
@@ -35,7 +36,13 @@ def copy_files_helper(folder_name):
         src = os.path.join(CURRENT_DIR, os.path.pardir, folder_name)
         if os.path.exists(src):
             dst = os.path.join(CURRENT_DIR, 'compile', folder_name)
-            shutil.rmtree(dst, ignore_errors=True)
+            if os.path.exists(dst):
+                if os.path.isdir:
+                    # see https://github.com/pypa/distutils/pull/21
```
StrikerRUS, line 41: Should other calls to `shutil.rmtree` be replaced too?
https://github.com/microsoft/LightGBM/blob/b360700411a31085af2319294a5553e7c667a242/python-package/setup.py#L76
https://github.com/microsoft/LightGBM/blob/b360700411a31085af2319294a5553e7c667a242/python-package/setup.py#L101
https://github.com/microsoft/LightGBM/blob/b360700411a31085af2319294a5553e7c667a242/python-package/setup.py#L260
https://github.com/microsoft/LightGBM/blob/b360700411a31085af2319294a5553e7c667a242/python-package/setup.py#L262

python-package/setup.py
```diff
@@ -251,6 +258,66 @@ def run(self):
...
+        ('hdfs', 'h', 'Compile HDFS version'),
+        ('bit32', None, 'Compile 32-bit version'),
+        ('precompile', 'p', 'Use precompiled library'),
+        ('boost-root=', None, 'Boost preferred installation prefix'),
+        ('boost-dir=', None, 'Directory with Boost package configuration file'),
+        ('boost-include-dir=', None, 'Directory containing Boost headers'),
+        ('boost-librarydir=', None, 'Preferred Boost library directory'),
+        ('opencl-include-dir=', None, 'OpenCL include directory'),
+        ('opencl-library=', None, 'Path to OpenCL library')
+    ]
```
StrikerRUS, line 279: Please deduplicate code by creating global variable `LIGHTGBM_OPTIONS` in `setup.py` and using it here and in `CustomInstall` class
https://github.com/microsoft/LightGBM/blob/b360700411a31085af2319294a5553e7c667a242/python-package/setup.py#L197-L213

StrikerRUS (changes requested):
@dolfinus Thanks a lot for your contribution! Please check my comments below:

StrikerRUS:
Linking #1863 here.

python-package/README.rst
StrikerRUS: Feel free to add more use cases here.

dolfinus:
All comments are fixed, thanks!

StrikerRUS (approved):
@dolfinus Thank you very much! LGTM!

Does anyone else from reviewers want to take a look at this PR?

jameslamb (approved):
this is a really nice addition, thanks for taking the time to contribute to LightGBM!

PR #3597, opened 2020-11-25
[docs] add link to lightgbm4j to readme

shuttie:
LightGBM4j is a high-level Java wrapper for the LightGBM library. It can be nice to have it referenced from the main docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,8 @@ Dask-LightGBM (distributed and parallel Python-package): https://github.com/dask
 
 Ruby gem: https://github.com/ankane/lightgbm
 
+LightGBM4j (Java high-level binding): https://github.com/metarank/lightgbm4j
+
 Support
 -------
 


jameslamb (commented):
Thanks for your interest in LightGBM, @shuttie !

It looks like this project is very new ([first commit was 2 days ago](https://github.com/metarank/lightgbm4j/commit/96fc0e566a202a6cb7e12389cb647955ee1b61db)). What settings has it been used in?

My first instinct here is that if you see usability challenges with LightGBM's SWIG interface, it would be better for your to open issues / submit pull requests here to improve that interface, instead of creating a separate project.

However I'm not the expert here on LightGBM-on-the-JVM. I think we should get an opinion from @imatiach-msft .

imatiach-msft:
@shuttie this looks great!  As @jameslamb said this is very new, and it would be nice to have more documentation in the project.  I agree the SWIG API is very low-level, and it makes sense to have a nicer API for Java-only users that don't use spark (note for Java users who use spark they can just use the mmlspark API implementation).  I think eventually it makes sense to merge the nicer APIs to lightgbm repository directly, but this is ok for now too.

jameslamb (approved):
Thanks for the quick help @imatiach-msft ! Based on that reply, I'm +1 on this PR.

jameslamb:
thanks for adding the label @StrikerRUS , sorry I forgot

shuttie:
@imatiach-msft I agree, the wrapper library is quite new, but it's a clean-room reimplementation of a similar internal wrapper (but in Scala) I was using in my day job. The general idea is to iterate a bit on the library and LightGBM method coverage and when it's good enough, merge it into LightGBM as an official out-of-the-box binding. 

As for it's production usage, the original internal code was used in the Findify (https://findify.io) backend to do LTR tasks. The lightgbm4j is not yet used anywhere, but there is a plan to switch to this library instead of the internal one. Also I'm adding some simple integration tests there to prove that the wrapper works for basic tasks. For example in https://github.com/metarank/lightgbm4j/blob/main/src/test/java/io/github/metarank/lightgbm4j/CancerIntegrationTest.java there is a binary classification task on a UCI Breast Cancer dataset, and it seems to match the output of the vanilla LightGBM binary.

PR #3602, opened 2020-11-27
[ci] remove unnecessary environment variables in R-package CI

jameslamb:
I realized tonight that we're setting environment variables `R_*_VERSION` unnecessarily in the GitHub Actions config for the R-package tests. It's set directly in https://github.com/microsoft/LightGBM/blob/6be3e57d21446022d33266605375660a9b35d9af/.ci/test_r_package.sh#L16-L24,  and I think that's a better place because we should try to keep the CI config files as simple as possible, in case they ever need to move to different CI providers.

This PR removes those unnecessary statements.


StrikerRUS (approved):
Thanks! Absolutely agree with this statement

> I think that's a better place because we should try to keep the CI config files as simple as possible, in case they ever need to move to different CI providers.

PR #3615, opened 2020-11-30
clarify warning about small target values with MAPE (fixes #3608)

### LINKED ISSUE #3608: [LightGBM] [Warning] Met 'abs(label) < 1', will convert them to '1' in MAPE objective and metric
I'm doing Regeression with MAPE (Mean Absolute Percentage Error) objective and eva_metric.
What is the meaning, of following warning? 
```
[LightGBM] [Warning] Met 'abs(label) < 1', will convert them to '1' in MAPE objective and metric
```

jameslamb:
See #3608 for more details.

If you use LightGBM's built-in MAPE metric, target values where `abs(label) < 1` are rounded to 1.0 when calculating that metric. LightGBM checks for such values and warns about them.

This PR proposes a change to that warning message that I think makes it a bit clearer.


guolinke (approved):
Thank you!

PR #3618, opened 2020-12-01
[python] more detailed docs for trees_to_dataframe(), create_tree_digraph(), plot_tree()

jameslamb:
I've been looking at the `[lightgbm]` tag on Stacck Overflow recently, and have seen a few questions about how to interpret the output of some visualization methods in the Python Package.

I think there is an opportunity to improve the documentation for those methods. This PR proposes more detailed docs on `trees_to_dataframe()`, `create_tree_digraph()`, and `plot_tree()` to help people understand them better. It's very possible that I've also misunderstood the meaning of some of the these things, so I appreciate your thorough reviews.

### Notes for Reviewers

I opened this from a LightGBM branch so that we can hopefully enable RTD builds and see if I made any formatting mistakes. @StrikerRUS if you agree with the spirit of this PR, could you enable RTD builds for this branch?

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -2207,42 +2207,42 @@ class Booster(object):
         result : pandas DataFrame
             Returns a pandas DataFrame of the parsed model, with one row per node.
             This dataframe has the following columns:
-              * `tree_index`: int64, which tree a node belongs to.
-                0-based, so a value of ``6``, for example, means "this node is in the
-                7th tree".
-              * ``node_depth``: int64, how far a node is from the root of the tree.
-                The root node has a value of ``1``, its direct children are ``2``, etc.
-              * ``node_index``: string, unique identifier for a node
-              * ``left_child``: string, ``node_index`` of the child node to the left
-                of a split. ``None`` for leaf nodes.
-              * ``right_child``: string, ``node_index`` of the child node to the right
-                of a split. ``None`` for leaf nodes.
-              * ``parent_index``: string, ``node_index`` of this node's parent. ``None``
-                for the root node.
-              * ``split_feature``: string, identifier for the feature used for splitting.
-                This is of the form ``"Column_i"``, where ``i`` refers to the number of the
-                feature. For example, the first feature would be ``"Column_0"``. ``None``
-                for leaf nodes.
-              * ``split_gain``: float64, gain from adding this split to the tree. ``NaN``
-                for leaf nodes.
-              * ``threshold``: float64, value of the feature used to decide which side of
-                the split a record will go down. ``NaN`` for leaf nodes.
-              * ``decision_type``: string, logical operator describing how to compare a value
-                to ``threshold``. For example, ``split_feature = "Column_10",
-                threshold = 15, decision_type = "<="`` means that records where ``Column_10 <= 15``
-                follow the leftt side of the split, otherwise follows the right side of the split.
-                ``None`` for leaf nodes.
-              * ``missing_direction``: string, split direction that missing values should go to.
-                ``None`` for leaf nodes.
-              * ``missing_type``: object, describes what types of values are treated
-                as missing.
-              * ``value``: float64, predicted value for this leaf node, multiplied by
-                the learning rate.
-              * ``weight`: float64, proportion of records in training data that fall
-                into this node. This is a value between 0 and 100, so ``65.4`` means
-                "65.4%  of the training data falls into this node".
-              * ``count`: int64, number of records in the training data that fall into
-                this node.
+            * ``tree_index``: int64, which tree a node belongs to.
+              0-based, so a value of ``6``, for example, means "this node is in the
+              7th tree".
+            * ``node_depth``: int64, how far a node is from the root of the tree.
+              The root node has a value of ``1``, its direct children are ``2``, etc.
+            * ``node_index``: string, unique identifier for a node
+            * ``left_child``: string, ``node_index`` of the child node to the left
+              of a split. ``None`` for leaf nodes.
+            * ``right_child``: string, ``node_index`` of the child node to the right
+              of a split. ``None`` for leaf nodes.
+            * ``parent_index``: string, ``node_index`` of this node's parent. ``None``
+              for the root node.
+            * ``split_feature``: string, identifier for the feature used for splitting.
+              This is of the form ``"Column_i"``, where ``i`` refers to the number of the
+              feature. For example, the first feature would be ``"Column_0"``. ``None``
+              for leaf nodes.
+            * ``split_gain``: float64, gain from adding this split to the tree. ``NaN``
+              for leaf nodes.
+            * ``threshold``: float64, value of the feature used to decide which side of
+              the split a record will go down. ``NaN`` for leaf nodes.
+            * ``decision_type``: string, logical operator describing how to compare a value
+              to ``threshold``. For example, ``split_feature = "Column_10",
+              threshold = 15, decision_type = "<="`` means that records where ``Column_10 <= 15``
+              follow the left side of the split, otherwise follows the right side of the split.
+              ``None`` for leaf nodes.
+            * ``missing_direction``: string, split direction that missing values should go to.
+              ``None`` for leaf nodes.
+            * ``missing_type``: object, describes what types of values are treated
+              as missing.
+            * ``value``: float64, predicted value for this leaf node, multiplied by
+              the learning rate.
+            * ``weight``: float64, proportion of records in training data that fall
+              into this node. This is a value between 0 and 100, so ``65.4`` means
+              "65.4%  of the training data falls into this node".
+            * ``count``: int64, number of records in the training data that fall into
+              this node.
         """
         if not PANDAS_INSTALLED:
             raise LightGBMError('This method cannot be run without pandas installed')
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -2204,42 +2204,21 @@ class Booster(object):
 
         The returned DataFrame has the following columns.
 
-            - ``tree_index``: int64, which tree a node belongs to.
-              0-based, so a value of ``6``, for example, means "this node is in the
-              7th tree".
-            - ``node_depth``: int64, how far a node is from the root of the tree.
-              The root node has a value of ``1``, its direct children are ``2``, etc.
-            - ``node_index``: string, unique identifier for a node.
-            - ``left_child``: string, ``node_index`` of the child node to the left
-              of a split. ``None`` for leaf nodes.
-            - ``right_child``: string, ``node_index`` of the child node to the right
-              of a split. ``None`` for leaf nodes.
-            - ``parent_index``: string, ``node_index`` of this node's parent. ``None``
-              for the root node.
-            - ``split_feature``: string, identifier for the feature used for splitting.
-              This is of the form ``"Column_i"``, where ``i`` refers to the number of the
-              feature. For example, the first feature would be ``"Column_0"``. ``None``
-              for leaf nodes.
-            - ``split_gain``: float64, gain from adding this split to the tree. ``NaN``
-              for leaf nodes.
-            - ``threshold``: float64, value of the feature used to decide which side of
-              the split a record will go down. ``NaN`` for leaf nodes.
-            - ``decision_type``: string, logical operator describing how to compare a value
-              to ``threshold``. For example, ``split_feature = "Column_10",
-              threshold = 15, decision_type = "<="`` means that records where ``Column_10 <= 15``
-              follow the left side of the split, otherwise follows the right side of the split.
-              ``None`` for leaf nodes.
-            - ``missing_direction``: string, split direction that missing values should go to.
-              ``None`` for leaf nodes.
-            - ``missing_type``: string, describes what types of values are treated
-              as missing.
-            - ``value``: float64, predicted value for this leaf node, multiplied by
-              the learning rate.
-            - ``weight``: float64 or int64, proportion of records in training data that fall
-              into this node. This is a value between 0 and 100, so ``65.4`` means
-              "65.4%  of the training data falls into this node".
-            - ``count``: int64, number of records in the training data that fall into
-              this node.
+            - ``tree_index`` : int64, which tree a node belongs to. 0-based, so a value of ``6``, for example, means "this node is in the 7th tree".
+            - ``node_depth`` : int64, how far a node is from the root of the tree. The root node has a value of ``1``, its direct children are ``2``, etc.
+            - ``node_index`` : string, unique identifier for a node.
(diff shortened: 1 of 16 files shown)


StrikerRUS:
@jameslamb 
>  could you enable RTD builds for this branch?

Sure, done!
https://readthedocs.org/projects/lightgbm/builds/

jameslamb:
There are some warnings, will fix them tonight

>  /basic.py:docstring of lightgbm.Booster.trees_to_dataframe:6: WARNING: Unexpected indentation.
 /basic.py:docstring of lightgbm.Booster.trees_to_dataframe:8: WARNING: Block quote ends without a blank line; unexpected unindent.
 /plotting.py:docstring of lightgbm.create_tree_digraph:26: WARNING: Unexpected indentation.
 /plotting.py:docstring of lightgbm.create_tree_digraph:27: WARNING: Block quote ends without a blank line; unexpected unindent.
 /plotting.py:docstring of lightgbm.plot_tree:31: WARNING: Unexpected indentation.
 /plotting.py:docstring of lightgbm.plot_tree:32: WARNING: Block quote ends without a blank line; unexpected unindent.
looking for now-outdated files... none found

https://readthedocs.org/projects/lightgbm/builds/12463523/

python-package/lightgbm/plotting.py
```diff
@@ -487,9 +497,20 @@ def create_tree_digraph(booster, tree_index=0, show_info=None, precision=3,
         The index of a target tree to convert.
     show_info : list of strings or None, optional (default=None)
         What information should be shown in nodes.
-        Possible values of list items:
-        'split_gain', 'internal_value', 'internal_count', 'internal_weight',
-        'leaf_count', 'leaf_weight', 'data_percentage'.
+
```
jameslamb, line 500: I think this blank line is necessary.

**with blank line**

image

**without blank line**

image

jameslamb:
build is passing (https://readthedocs.org/projects/lightgbm/builds/12467879/) and I think the docs look ok, so this is ready for review.

* https://lightgbm.readthedocs.io/en/docs-trees-to-dataframe/pythonapi/lightgbm.plot_tree.html#lightgbm.plot_tree
* https://lightgbm.readthedocs.io/en/docs-trees-to-dataframe/pythonapi/lightgbm.create_tree_digraph.html#lightgbm.create_tree_digraph
* https://lightgbm.readthedocs.io/en/docs-trees-to-dataframe/pythonapi/lightgbm.Booster.html#lightgbm.Booster.trees_to_dataframe

Please check that my descriptions are correct, I'm not sure about all of them.

python-package/lightgbm/basic.py
```diff
@@ -2202,6 +2202,45 @@ def free_network(self):
...
         """Parse the fitted model and return in an easy-to-read pandas DataFrame.
 
+        The returned DataFrame has the following columns.
+
+            - ``tree_index``: int64, which tree a node belongs to.
+              0-based, so a value of ``6``, for example, means "this node is in the
+              7th tree".
+            - ``node_depth``: int64, how far a node is from the root of the tree.
+              The root node has a value of ``1``, its direct children are ``2``, etc.
+            - ``node_index``: string, unique identifier for a node
```
StrikerRUS, line 2212:
```suggestion
            - ``node_index``: string, unique identifier for a node.
```

python-package/lightgbm/basic.py
```diff
@@ -2202,6 +2202,45 @@ def free_network(self):
...
+            - ``left_child``: string, ``node_index`` of the child node to the left
+              of a split. ``None`` for leaf nodes.
+            - ``right_child``: string, ``node_index`` of the child node to the right
+              of a split. ``None`` for leaf nodes.
+            - ``parent_index``: string, ``node_index`` of this node's parent. ``None``
+              for the root node.
+            - ``split_feature``: string, identifier for the feature used for splitting.
+              This is of the form ``"Column_i"``, where ``i`` refers to the number of the
+              feature. For example, the first feature would be ``"Column_0"``. ``None``
+              for leaf nodes.
```
StrikerRUS, line 2222: This description is not always true. Feature names can have custom values passed by user.
```suggestion
            - ``split_feature``: string, name of the feature used for splitting.
               ``None`` for leaf nodes.
```

python-package/lightgbm/basic.py
```diff
@@ -2202,6 +2202,45 @@ def free_network(self):
...
+              threshold = 15, decision_type = "<="`` means that records where ``Column_10 <= 15``
+              follow the left side of the split, otherwise follows the right side of the split.
+              ``None`` for leaf nodes.
+            - ``missing_direction``: string, split direction that missing values should go to.
+              ``None`` for leaf nodes.
+            - ``missing_type``: object, describes what types of values are treated
+              as missing.
+            - ``value``: float64, predicted value for this leaf node, multiplied by
+              the learning rate.
+            - ``weight``: float64, proportion of records in training data that fall
```
StrikerRUS, line 2238:
```suggestion
            - ``weight``: float64 or int64, proportion of records in training data that fall
```

python-package/lightgbm/basic.py
```diff
@@ -2202,6 +2202,45 @@ def free_network(self):
...
+            - ``threshold``: float64, value of the feature used to decide which side of
+              the split a record will go down. ``NaN`` for leaf nodes.
+            - ``decision_type``: string, logical operator describing how to compare a value
+              to ``threshold``. For example, ``split_feature = "Column_10",
+              threshold = 15, decision_type = "<="`` means that records where ``Column_10 <= 15``
+              follow the left side of the split, otherwise follows the right side of the split.
+              ``None`` for leaf nodes.
+            - ``missing_direction``: string, split direction that missing values should go to.
+              ``None`` for leaf nodes.
+            - ``missing_type``: object, describes what types of values are treated
```
StrikerRUS, line 2234:
```suggestion
            - ``missing_type``: string, describes what types of values are treated
```

python-package/lightgbm/basic.py
StrikerRUS, line 2238: Seems that the description is not true.

#2037
#2269

image

python-package/lightgbm/basic.py
```diff
@@ -2202,6 +2202,45 @@ def free_network(self):
...
+              as missing.
+            - ``value``: float64, predicted value for this leaf node, multiplied by
+              the learning rate.
+            - ``weight``: float64, proportion of records in training data that fall
+              into this node. This is a value between 0 and 100, so ``65.4`` means
+              "65.4%  of the training data falls into this node".
+            - ``count``: int64, number of records in the training data that fall into
```
StrikerRUS, line 2241:
```suggestion
            - ``tree_index`` : int64, which tree a node belongs to.
              0-based, so a value of ``6``, for example, means "this node is in the
              7th tree".
            - ``node_depth`` : int64, how far a node is from the root of the tree.
              The root node has a value of ``1``, its direct children are ``2``, etc.
            - ``node_index`` : string, unique identifier for a node
            - ``left_child`` : string, ``node_index`` of the child node to the left
              of a split. ``None`` for leaf nodes.
            - ``right_child`` : string, ``node_index`` of the child node to the right
              of a split. ``None`` for leaf nodes.
            - ``parent_index`` : string, ``node_index`` of this node's parent. ``None``
              for the root node.
            - ``split_feature`` : string, identifier for the feature used for splitting.
              This is of the form ``"Column_i"``, where ``i`` refers to the number of the
              feature. For example, the first feature would be ``"Column_0"``. ``None``
              for leaf nodes.
            - ``split_gain`` : float64, gain from adding this split to the tree. ``NaN``
              for leaf nodes.
            - ``threshold`` : float64, value of the feature used to decide which side of
              the split a record will go down. ``NaN`` for leaf nodes.
            - ``decision_type`` : string, logical operator describing how to compare a value
              to ``threshold``. For example, ``split_feature = "Column_10",
              threshold = 15, decision_type = "<="`` means that records where ``Column_10 <= 15``
              follow the left side of the split, otherwise follows the right side of the split.
              ``None`` for leaf nodes.
            - ``missing_direction`` : string, split direction that missing values should go to.
              ``None`` for leaf nodes.
            - ``missing_type`` : object, describes what types of values are treated
              as missing.
            - ``value`` : float64, predicted value for this leaf node, multiplied by
              the learning rate.
            - ``weight`` : float64, proportion of records in training data that fall
              into this node. This is a value between 0 and 100, so ``65.4`` means
              "65.4%  of the training data falls into this node".
            - ``count`` : int64, number of records in the training data that fall into
```

python-package/lightgbm/plotting.py
```diff
@@ -474,6 +474,16 @@ def create_tree_digraph(booster, tree_index=0, show_info=None, precision=3,
                         orientation='horizontal', **kwargs):
     """Create a digraph representation of specified tree.
 
+    Each node in the graph represents a node in the tree.
+
+    Non-leaf nodes have labels like ``Column_10 <= 875.9``, which means
+    "this node splits on the 11th feature, with threshold 875.9".
```
StrikerRUS, line 480:
```suggestion
    "this node splits on the feature named "Column_10", with threshold 875.9".
```

python-package/lightgbm/plotting.py
```diff
@@ -536,6 +557,14 @@ def plot_tree(booster, ax=None, tree_index=0, figsize=None, dpi=None,
               show_info=None, precision=3, orientation='horizontal', **kwargs):
     """Plot specified tree.
 
+    Non-leaf nodes have labels like ``Column_10 <= 875.9``, which means
+    "this node splits on the 11th feature, with threshold 875.9".
```
StrikerRUS, line 561:
```suggestion
    "this node splits on the feature named "Column_10", with threshold 875.9".
```

python-package/lightgbm/plotting.py
```diff
@@ -487,9 +497,20 @@ def create_tree_digraph(booster, tree_index=0, show_info=None, precision=3,
...
+            - ``'internal_count'`` : number of records from the training data \
+                that fall into this non-leaf node
+            - ``'internal_weight'`` : total weight of all nodes that \
+                fall into this non-leaf node
+            - ``'leaf_count'`` : number of records from the training data \
+                that fall into this leaf node
+            - ``'leaf_weight'`` : total weight of all nodes that \
+                fall into this leaf node
+            - ``'data_percentage'`` : percentage of training data that fall into \
+                this node
```
StrikerRUS, line 513: FYI, we don't have line length limit for Python files. So next time there is no need to split such short lines in "unexpected" places. 🙂 

```suggestion
            - ``'split_gain'`` : gain from adding this split to the model
            - ``'internal_value'`` : raw predicted value that would be produced
              by this node if it was a leaf node
            - ``'internal_count'`` : number of records from the training data
              that fall into this non-leaf node
            - ``'internal_weight'`` : total weight of all nodes that
              fall into this non-leaf node
            - ``'leaf_count'`` : number of records from the training data
              that fall into this leaf node
            - ``'leaf_weight'`` : total weight of all nodes that
              fall into this leaf node
            - ``'data_percentage'`` : percentage of training data that fall into
              this node
```

python-package/lightgbm/plotting.py
```diff
@@ -556,9 +585,20 @@ def plot_tree(booster, ax=None, tree_index=0, figsize=None, dpi=None,
...
```
StrikerRUS, line 601:
```suggestion
            - ``'split_gain'`` : gain from adding this split to the model
            - ``'internal_value'`` : raw predicted value that would be produced
              by this node if it was a leaf node
            - ``'internal_count'`` : number of records from the training data
              that fall into this non-leaf node
            - ``'internal_weight'`` : total weight of all nodes that
              fall into this non-leaf node
            - ``'leaf_count'`` : number of records from the training data
              that fall into this leaf node
            - ``'leaf_weight'`` : total weight of all nodes that
              fall into this leaf node
            - ``'data_percentage'`` : percentage of training data that fall into
              this node
```

StrikerRUS (changes requested):
@jameslamb Thanks for this great PR! I believe it will help a lot users to better understand returned results of these methods.

Please check my comments below.

python-package/lightgbm/basic.py
jameslamb: ooooo ok thank you!

python-package/lightgbm/plotting.py
jameslamb: oh ok, I didn't realize. I'll just change them to single lines right now in this PR.

I originally had this written like your suggestion in https://github.com/microsoft/LightGBM/pull/3618/files#r534262617, but that causes sphinx to shade the first line gray.

python-package/lightgbm/basic.py
jameslamb: ok, I'll change this then. But I don't think just saying "name of the feature used for splitting" is sufficient. We should document both cases.

What do you think about this?

> string, name of the feature used for splitting. If `feature_name` was not provided to the `fit()` call that produced this model, this will be of the form `"Column_i"`, where `i` refers to the number of the feature. So for example, the first feature would be `"Column_0"`.

I think it's important to specify that these `Column_i` names are 0-based (since that might not be obvious), and that they're based on the order of the input data.

python-package/lightgbm/basic.py
jameslamb: Is this correct?

> float64 or int64, sum of hessian (second-order derivative of objective), summed over observations that fall in this node.

I think it is, because I tried with `LGBMClassifier` and saw that the values of `weight` and `count` were the same.

python-package/lightgbm/basic.py
StrikerRUS: @jameslamb 

> was not provided to the `fit()` call that produced this model

There are a lot of methods where feature names can be set, not only `fit()` from sklearn wrapper.

I believe it is enough to say that default feature names are in a form of `"Column_i"` (zero-based). But please my next paragraph.

>  We should document both cases.

Yeah, I support this idea! But I think that this can be documented in a more "general" place, because default feature names are common among all wrappers.

jameslamb: oh great points! Ok I'll change it, I agree with you

python-package/lightgbm/plotting.py
```diff
@@ -536,6 +551,14 @@ def plot_tree(booster, ax=None, tree_index=0, figsize=None, dpi=None,
               show_info=None, precision=3, orientation='horizontal', **kwargs):
     """Plot specified tree.
 
+    Non-leaf nodes have labels like ``Column_10 <= 875.9``, which means
```
StrikerRUS, line 554: To unify descriptions for `create_tree_digraph` and `plot_tree`.
```suggestion
    Each node in the graph represents a node in the tree.

    Non-leaf nodes have labels like ``Column_10 <= 875.9``, which means
```

StrikerRUS (approved):
@jameslamb Great improvement! Very clean description!

@guolinke Could you please also check that field descriptions are correct?

guolinke (approved):
LGTM

PR #3627, opened 2020-12-04
[docs] update documentation with AUC-mu, average precision

jameslamb:
I was reading https://lightgbm.readthedocs.io/en/latest/Features.html today and realized that a list of metrics is maintained there. That list is missing some of the newer things that @btrotta has added (average precision, AUC-mu). This PR proposes adding them.

I'm also proposing something I think we should do more of...adding a label like `new in v3.0.0`. Since we only host one version of the documentation, based on `master`, I think these types of annotations are useful. This is a practice I really appreciate in the `scikit-learn` docs, for example.

For example, https://scikit-learn.org/stable/modules/generated/sklearn.compose.ColumnTransformer.html#sklearn-compose-columntransformer

image

### DIFF
--- a/docs/Features.rst
+++ b/docs/Features.rst
@@ -205,9 +205,9 @@ LightGBM supports the following metrics:
 
 -  Multi-class error rate
 
--  AUC-mu **(--- new in v3.0.0 ---)**
+-  AUC-mu ``(new in v3.0.0)``
 
--  Average precision **(new in v3.1.0)**
+-  Average precision ``(new in v3.1.0)``
 
 -  Fair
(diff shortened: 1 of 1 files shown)


StrikerRUS (approved):

jameslamb:
> Good idea of adding version info!
> 
> For other items next time we can use a special Sphinx directive [`versionadded`](https://www.sphinx-doc.org/en/1.7/markup/para.html#directive-versionadded). (We cannot use it here because this directive is paragraph-level.)

oh cool! I did not know about that!

Issue #2919, opened 2020-03-16
regularization for categorical variables


devilwing0723:
Hi, I am using lightgbm to train model with some categorical variables. I am wondering where can I find more detailed documentation/introduction on the hyper parameters that can regularize the categorical variables, such as max_cat_threshold, max_cat_to_onehot, cat_l2, cat_smooth? Current documentation did not really specify how these parameters regularize, and what range of values is appropriate for hyper-tuning.

guolinke:
refer to the code here: https://github.com/microsoft/LightGBM/blob/9654f16528d2d1e3462a298106cb1fc6ed43cbac/src/treelearner/feature_histogram.hpp#L259-L483

cat_l2 is the additional l2 reg for categorical features: https://github.com/microsoft/LightGBM/blob/9654f16528d2d1e3462a298106cb1fc6ed43cbac/src/treelearner/feature_histogram.hpp#L349
cat_smooth is the smooth term when calculating the target encoding: https://github.com/microsoft/LightGBM/blob/9654f16528d2d1e3462a298106cb1fc6ed43cbac/src/treelearner/feature_histogram.hpp#L352

`max_cat_to_onehot` is too force one-vs-rest splits, when total number of categories is smaller than `max_cat_to_onehot`. 

`max_cat_threshold` is the max number of thresholds will be chosen in split condition: https://github.com/microsoft/LightGBM/blob/master/src/treelearner/feature_histogram.hpp#L364-L365

Issue #2617, opened 2019-12-06
Documentation Request: More information on finding the optimal ranges for parameter tuning?


duckladydinh:
Hi,

I start to play with LightGBM and want to optimize its parameters according to the tuning documentation. I would like to ask if there is a way to determine a good range for each parameter? Like bagging fraction for example? The range [0, 1] is understandable but is there a way to limit it? Like [0.8, 0.9] or something like that?

And also, I wonder why there are only a few (around a dozen or so) parameters mentioned in the tuning section? What about the rest? There are a lot of others in the parameters section, aren't they?

Thank you very much! I hope someone can share your experience!

Happy Coding!
Thuan

StrikerRUS:
@duckladydinh Hi!
It is very task- and dataset-dependent. There is no silver bullet.
For example, you can read this comment for unbalanced datasets: https://github.com/microsoft/LightGBM/issues/695#issuecomment-315591634.
In addition, you may want to get familiar with this awesome site: https://sites.google.com/view/lauraepp/parameters.
Also, XGBoost recently had a suggestion to revisit their default param values: https://github.com/dmlc/xgboost/issues/4986

I remember we had an internal conversation about the need to change default params in order to respect small datasets. At least, we might enhance the current guide with suggested ranges for different types of datasets.

StrikerRUS:
Closed in favor of being in #2302. We decided to keep all feature requests in one place.

Welcome to contribute this feature! Please re-open this issue (or post a comment if you are not a topic starter) if you are actively working on implementing this feature.

Issue #2411, opened 2019-09-15
Unreliable categorical groups


nirupamkar:
I am dealing with a classification problem with categorical features. While doing so I have encountered a strange situation. I have trimmed down the situation to a base case.
 
1.Say I have 100M rows with 1 feature having more than 1M categories. The actual target rate of this categories can range from 0 to 1.  

2.While predicting on the train dataset itself I observed there are categories with (say ~500 count and 0 target)  and the prediction is coming to be close to the mean target of the train data

3.It could be because of regularisation or the categorical split. I reduced the regularisation to practically zero and it was still similar

4.So it appeared it might be how the categorical splits have been made for the feature. What I did was I created 1 tree with say 150 nodes and with max_cat_thresold=150 and boosting_from_average=True, so that I can broadly check the categorical splits. (right now the split cannot be accessed from python interface )

5.I predicted the leaf index and calculated the target rate for each leaf index. There were almost 150 leaves so I can assume categorical histogram has got ~150 bins.

6.Coming to the observations

- There was a leaf(say L) with almost 50-60% of data with target rate almost same as the overall data
- There were leaves with ~200-300 counts and 0 target rate. So bins were made for these.
- However there was a significant number of categories(say C) with ~ 500 count and 0 target being clubbed in leaf L and thereby predicting average value for those

My question is why these categories are not being combined with 0 target rate bins. I wanted to upload the data over here, however the size is very big and the issue is less significant for lower volume of data. However I can give clarification of any questions if needed.

I am using server and it has got memory of 256GB.

@guolinke any help is really appreciated.

guolinke:
I think you misunderstood the categorical split algorithm.
There is no "merge bins" for categorical features, all categories are unique, except the long tail (the last 1% sorted by frequence) will be in the one bin. 
Therefore, in your case, with `max_cat_thresold=150`, only 300 (two side scanning, the first 150 and the last 150) are evaluated in categorical splits. And as a result, the categorical without counts=500 may in the middle, and be ignored. 

For the categorical features with many categories, I recommender to use the target encoding outside of GBDT.

nirupamkar:
@guolinke thanks for the quick reply. Really appreciate it. I misunderstood  the algorithm of categorical split of lgb. However if i put cat threshold=150, so there will be 1 bin with more than 1 category and rest will be having single category. Please confirm this.
Now if I put num_leaves=150, I am getting around 148 leaves with most having more than 1 categories. Am I missing something?

[...]

Issue #2261, opened 2019-07-12
`max_cat_threshold` Hyperparameter


pford221:
Hi, 

Is there another way to find out what is happening with `max_cat_threhsold` without trying to find it in the C++ source code? To save anyone kind enough to reply some time, I'll share my hypothesis and hopefully you can just correct me if I'm wrong?

Suppose there is a categorical variable of cardinality 10 (i.e. 10 unique levels). Also suppose that we set `max_cat_threshold = 1`. At each split opportunity, the algorithm aggregates the sum(gradients) / sum(hessians) for all the records in each of the 10 categories and then sorts the 10 categories by that ratio from lowest to highest (or highest to lowest...doesn't matter). Then because `max_cat_threshold = 1`, the algorithm only has one split point to evaluate and this split point is as close to the median (as determined by number of observations or weighted observations) as possible?

I appreciate any help on this!

guolinke:
@pford221 yeah, you are right. max_cat_threshold is to limit the categorical split points.

pford221:
Thank you, @guolinke. I realize that `max_cat_threshold` limits split points, but would you mind commenting on if my description of how it does it more-or-less correct? I really appreciate it.

guolinke:
@pford221 
I think most of them are correct.
But for 

> Then because max_cat_threshold = 1, the algorithm only has one split point to evaluate and this split point is as close to the median (as determined by number of observations or weighted observations) as possible?

, I want to clarify that, the split point will be the one with the highest (or lowest) `sum(gradients) / sum(hessians) `.

PR #3628, opened 2020-12-04
[docs] Add details on improving training speed

jameslamb:
This PR proposes a large update to the documentation on tuning hyperparameters (https://lightgbm.readthedocs.io/en/latest/Parameters-Tuning.html). Today, that section contains simple suggestions like "use small `max_depth`". In this PR, I'm proposing adding some context to each recommendation so users can understand more of the intuition for *how* these parameters affect training or dataset construction. I think these descriptions can also help users to better understand how LightGBM works generally.

image

### Notes for Reviewers

* I didn't want this PR to get too big so I just covered the section `For Faster Speed`. If reviewers like this contribution, I'd be happy to add similar contributions like it in future PRs for `For Better Accuracy` and `Deal with Overfitting`.
* @StrikerRUS I pushed this as a LightGBM doc so we can test readthedocs builds. Can you please enable builds for this branch?
* Similar to #3618 , I thought about this from reading Stack Overflow answers on the `[lightgbm]`. I might have a few more PRs in the coming weeks with more documentation change proposals like this.


docs/Parameters-Tuning.rst
```diff
@@ -36,15 +36,143 @@ To get good results using a leaf-wise tree, these are some important parameters:
...
+Increase ``min_data_in_bin`` to reduce training time.
+
+Decrease ``feature_fraction``
+*****************************
+
+By default, LightGBM considers all features in a ``Dataset`` during the training process. This behavior can be changed by setting ``feature_fraction`` to a value ``> 0`` and ``<= 1.0``. Setting ``feature_fraction`` to ``0.5``, for example, tells LightGBM to randomly select ``50%`` of features at the beginning of constructing each tree. This reduces the total number of splits that have to be evaluated to add each tree node.
+
+Decrease ``feature_fraction`` to reduce training time.
+
+Decrease ``max_cat_threshold``
```
jameslamb, line 152: I based this paragraph on the explanations in https://github.com/microsoft/LightGBM/issues/2261, https://github.com/microsoft/LightGBM/issues/2411, and https://github.com/microsoft/LightGBM/issues/2919

StrikerRUS:
> Can you please enable builds for this branch?

Done!

docs/Parameters-Tuning.rst
```diff
@@ -36,15 +36,143 @@ To get good results using a leaf-wise tree, these are some important parameters:
...
+'''''''''''''''''''''''''''''''''''''
+
+You might find that training is faster using a GPU-enabled build of LightGBM. See the `GPU Tutorial <./GPU-Tutorial.rst>`__ for details.
+
+Grow Shallower Trees
+''''''''''''''''''''
+
+The total training time for LightGBM increases with the total number of tree nodes added. LightGBM comes with several parameters that can be used to control the number of nodes per tree.
+
+The suggestions below will speed up training, but might hurt statistical performance.
```
btrotta, line 58: Maybe replace "statistical performance" with "training accuracy"? "Performance" is often used to mean "speed" so I think this would be less ambiguous. Also it might be good to mention that these options will also reduce overfitting, which can improve accuracy on the validation set.

docs/Parameters-Tuning.rst
```diff
@@ -36,15 +36,143 @@ To get good results using a leaf-wise tree, these are some important parameters:
...
+'''''''''''''''
+
+Decrease ``num_iterations``
+***************************
+
+The ``num_iterations`` parameter controls the number of boosting rounds that will be performed. Since LightGBM uses decision trees as the learners, this can also be thought of as "number of trees".
+
+Choosing the right value of ``num_iterations`` is highly dependent on the data and objective, so this parameter is often chosen from a set of possible values through hyperparameter tuning.
+
+Decrease ``num_iterations`` to reduce training time.
```
btrotta, line 100: It might be good to mention that `num_iterations` is often adjusted in conjunction with `learning_rate`. See, e.g. docs by @Laurae2 here: https://sites.google.com/view/lauraepp/parameters (under "Learning rate" section).

docs/Parameters-Tuning.rst
```diff
@@ -36,15 +36,143 @@ To get good results using a leaf-wise tree, these are some important parameters:
...
+
+Use Early Stopping
+******************
+
+If early stopping is enabled, after each boosting round the model's performance is evaluated against a validation set that contains data not available to the training process. That performance is then compared to the performance as of the previous boosting round. If the model's performance fails to improve for some number of consecutive rounds, LightGBM stops the training process.
```
btrotta, line 105: Similar to comment above, maybe clearer to use "accuracy" instead of "performance".

btrotta:
@jameslamb This is great! I think it will make LightGBM a lot more accessible for people new to tree-boosting algorithms. I left a few small suggestions.

docs/Parameters-Tuning.rst
jameslamb: oh good point! Ok I like the idea to use the term "training accuracy".

> Also it might be good to mention that these options will also reduce overfitting

I actually intentionally chose NOT to say that. There is a separate section in this document on how to improve training accuracy and another on how to fight overfitting. I wanted to keep this section only focused on runtime. I think being very straightforward like that makes it easier for people to reason about the direction and size of effects from changing these parameters.

The docs at https://lightgbm.readthedocs.io/en/latest/Parameters.html do have entries under each parameter like "can be used to reduce overfitting", "can be used to speed up training", etc.

docs/Parameters-Tuning.rst
jameslamb: good suggestion! Updated in https://github.com/microsoft/LightGBM/pull/3628/commits/d615ad46df1c07eba298aab7d37ad95f1d513b1c

Also I never noticed before that that is a Google site :laughing:

docs/Parameters-Tuning.rst
jameslamb: makes sense. I just replaced all of these with "accuracy" or "training accuracy": https://github.com/microsoft/LightGBM/pull/3628/commits/d615ad46df1c07eba298aab7d37ad95f1d513b1c

btrotta (approved):

StrikerRUS:
@jameslamb Unfortunately, you couldn't modify `Parameters.rst` directly. You should include your fixes in `config.h` and then run `helpers/parameter_generator.py` script. Refer to the comment in `Parameters.rst`:
> ..  List of parameters is auto generated by LightGBM\helpers\parameter_generator.py from LightGBM\include\LightGBM\config.h file.

docs/Parameters-Tuning.rst
```diff
@@ -36,15 +36,145 @@ To get good results using a leaf-wise tree, these are some important parameters:
...
+Choosing the right value of ``num_iterations`` and ``learning_rate`` is highly dependent on the data and objective, so these parameters are often chosen from a set of possible values through hyperparameter tuning.
+
+Decrease ``num_iterations`` to reduce training time.
+
+Use Early Stopping
+******************
+
+If early stopping is enabled, after each boosting round the model's training accuracy is evaluated against a validation set that contains data not available to the training process. That accuracy is then compared to the accuracy as of the previous boosting round. If the model's accuracy fails to improve for some number of consecutive rounds, LightGBM stops the training process.
+
+That "number of consecutive rounds" is controlled by the parameter ``early_stopping_rounds``. For example, ``early_stopping_rounds=1`` says *the first time accuracy on the validation set does not improve, stop training*.
```
StrikerRUS, line 109:
```suggestion
That "number of consecutive rounds" is controlled by the parameter ``early_stopping_rounds``. For example, ``early_stopping_rounds=1`` says "the first time accuracy on the validation set does not improve, stop training".
```

docs/Parameters-Tuning.rst
```diff
@@ -36,15 +36,145 @@ To get good results using a leaf-wise tree, these are some important parameters:
 For Faster Speed
 ----------------
 
--  Use bagging by setting ``bagging_fraction`` and ``bagging_freq``
+Add More Computational Resources
+''''''''''''''''''''''''''''''''
 
--  Use feature sub-sampling by setting ``feature_fraction``
+On systems where it is available, LightGBM uses OpenMP to parallelize many operations. The maximum number of threads used by LightGBM is controlled by the parameter ``num_threads``. By default, this will defer to the default behavior of OpenMP (one thread per real CPU core or the value in environment variable ``OMP_NUM_THREADS``, if it is set). For best performance, set this to the number of real CPU cores available.
```
StrikerRUS, line 42:
```suggestion
On systems where it is available, LightGBM uses OpenMP to parallelize many operations. The maximum number of threads used by LightGBM is controlled by the parameter ``num_threads``. By default, this will defer to the default behavior of OpenMP (one thread per real CPU core or the value in environment variable ``OMP_NUM_THREADS``, if it is set). For best performance, set this to the number of **real** CPU cores available.
```

docs/Parameters.rst
```diff
@@ -312,7 +312,7 @@ Learning Control Parameters
 
    -  frequency for bagging
 
-   -  ``0`` means disable bagging; ``k`` means perform bagging at every ``k`` iteration
+   -  ``0`` means disable bagging; ``k`` means perform bagging at every ``k`` iteration. Every ``k``-th iteration, LightGBM will randomly select ``bagging_fraction%`` of the data to use for the next ``k`` iterations.
```
StrikerRUS, line 315:
```suggestion
   -  ``0`` means disable bagging; ``k`` means perform bagging at every ``k`` iteration. Every ``k``-th iteration, LightGBM will randomly select ``bagging_fraction * 100 %`` of the data to use for the next ``k`` iterations
```

docs/Parameters.rst
```diff
@@ -358,11 +360,13 @@ Learning Control Parameters
...
 
-   -  ``<= 0`` means disable
+   -  can be used to speed up training
+
+   -  ``<= 0`` means disable (the default)
 
 -  ``first_metric_only`` :raw-html:`<a id="first_metric_only" title="Permalink to this parameter" href="#first_metric_only">&#x1F517;&#xFE0E;</a>`, default = ``false``, type = bool
 
-   -  set this to ``true``, if you want to use only the first metric for early stopping
+   -  LightGBM allows you to provide multiple evaluation metrics. Set this to ``true``, if you want to use only the first metric for early stopping.
```
StrikerRUS, line 369:
```suggestion
   -  LightGBM allows you to provide multiple evaluation metrics. Set this to ``true``, if you want to use only the first metric for early stopping
```

docs/Parameters.rst
```diff
@@ -382,6 +386,8 @@ Learning Control Parameters
 
 -  ``min_gain_to_split`` :raw-html:`<a id="min_gain_to_split" title="Permalink to this parameter" href="#min_gain_to_split">&#x1F517;&#xFE0E;</a>`, default = ``0.0``, type = double, aliases: ``min_split_gain``, constraints: ``min_gain_to_split >= 0.0``
 
+   -  can be used to speed up training
+
    -  the minimal gain to perform split
```
StrikerRUS, line 391: Swap these points please as we use `can be used ...` at the very end of param description.

docs/Parameters.rst
```diff
@@ -442,7 +448,9 @@ Learning Control Parameters
 
    -  used for the categorical features
 
-   -  limit the max threshold points in categorical features
+   -  limit number of split points considered for categorical features. See `the documentation on how LightGBM finds optimal splits for categorical features <./Features.rst#optimal-split-for-categorical-features>`_ for more details.
```
StrikerRUS, line 451:
```suggestion
   -  limit number of split points considered for categorical features. See `the documentation on how LightGBM finds optimal splits for categorical features <./Features.rst#optimal-split-for-categorical-features>`_ for more details
```

StrikerRUS (changes requested):
Awesome addition! It greatly contributes to #2617, I believe. Just some minor typos below:

jameslamb:
> @jameslamb Unfortunately, you couldn't modify `Parameters.rst` directly. You should include your fixes in `config.h` and then run `helpers/parameter_generator.py` script. Refer to the comment in `Parameters.rst`:
> 
> > ..  List of parameters is auto generated by LightGBM\helpers\parameter_generator.py from LightGBM\include\LightGBM\config.h file.

ohhhh sorry I forgot this

docs/Parameters.rst
jameslamb: done!

StrikerRUS (approved):

PR #3611, opened 2020-11-29
release 3.1.1

jameslamb:
PR for release 3.1.1. This is a bugfix release, so feature or breaking PRs like #3405, #3515, and #3581 should not be merged into it.

I created this on a LightGBM branch so any maintainer can push to it.

### Release checklist:

The PRs added to the checklist below are based on https://github.com/microsoft/LightGBM/pull/3586#issuecomment-735417087

- [x] Update [VERSION.txt](https://github.com/microsoft/LightGBM/blob/master/VERSION.txt) number.
- [x] Update version in [Appveyor config file](https://github.com/microsoft/LightGBM/blob/master/.appveyor.yml).
- [x] Update version in [`configure` file of R-package](https://github.com/microsoft/LightGBM/blob/master/R-package/configure).
- [x] merge #3598
- [x] merge #3592
- [x] fix #3629
- [x] run R valgrind checks after all PRs are merged
- [x] run R Solaris checks after all PRs are merged
- [ ] Push commit with `v*` tag to trigger `GitHubRelease` action at Azure Pipelines.
- [ ] Convert automatic GitHub release from `Draft` to normal one.
- [ ] Update [`stable` tag at GitHub](https://github.com/microsoft/LightGBM/releases/tag/stable).
- [ ] Upload release to [PyPI](https://pypi.org/project/lightgbm/).
- [ ] Upload release to [CRAN](https://cran.r-project.org/package=lightgbm).
- [ ] Update `cran-comments.md` when 3.1.1 is accepted on CRAN
- [ ] Upload release to [NuGet](https://www.nuget.org/packages/LightGBM/).
- [ ] Update version and hash in [Homebrew formula](https://github.com/Homebrew/homebrew-core/blob/master/Formula/lightgbm.rb).

PR #3630, opened 2020-12-05
[R-package] fix partial matching of keyword arguments in lgb.cv() (fixes #3629)

### LINKED ISSUE #3616: [ci] [R-package] R CI jobs on mac, linux do not always fail if unit tests fail
Look at this build from #3598 : https://github.com/microsoft/LightGBM/pull/3598/checks?check_run_id=1477128773. It was marked successful in GitHub Actions, even though a unit test is failing.

image

> Status: 1 ERROR

Similar jobs on Mac have the same test error, but are marked as failures by GitHub Actions. For example, https://github.com/microsoft/LightGBM/pull/3598/checks?check_run_id=1477128812.

image

It looks like the ones that failed only failed because they also had NOTEs.

So I think that the current R CI scripts aren't correctly accounting for the possibility of an `R CMD check` run that yields ERRORs but no NOTEs or WARNINGs. `R CMD check` *should* yield a non-0 exit code on ERRORs.

Windows jobs all correctly fail under these conditions, with a message like this

> ERRORs have been found by R CMD check!

### How to Fix This

Something in this code is not correctly capturing ERRORs on Mac and Linux: https://github.com/microsoft/LightGBM/blob/c02917e493c36f3b1e349338f1087fed33126576/.ci/test_r_package.sh#L156-L163.

You can add a `stop("hello")` in any of the R unit tests in `R-package/tests/testthat` to simulate the situation where `R CMD check` results in `Status: 1 ERROR`, and then try to fix that script. It might be the case that the trickery with putting `R CMD check` into the background and polling (to avoid timeouts) results in some mistakes, and the solution could be to just grep for `ERROR` in the output of `R CMD check`, just like we do for `WARNING`:

https://github.com/microsoft/LightGBM/blob/c02917e493c36f3b1e349338f1087fed33126576/.ci/test_r_package.sh#L184-L187

### LINKED ISSUE #3629: [R-package] R CMD CHECK note 'partial argument match'
Moving this discussion from https://github.com/microsoft/LightGBM/pull/3611#discussion_r536890212.

`R CMD CHECK` on the R package currently produces the following note:

> checking R code for possible problems ... NOTE
partial argument match of 'data' to 'dataset'
lgb.cv : : warning in getinfo(data = data, name =
"init_score"): partial argument match of 'data' to 'dataset'
lgb.cv : : warning in getinfo(data = data, name = "weight"):

## What this NOTE means and how to fix it

R supports partial argument matching. If the name of a keyword argument is close to an actual keyword argument, R will allow it.

```r
f <- function(data = 5){return(data + 1)}

f(dat = 11)
# [1] 12
```

These `getinfo()` keyword arguments were introduced in the work to close #3390 . The immediate fix for them is to just change those uses to `dataset =`

## Why wasn't this caught by CI?

When testing the R package today, `R CMD CHECK` produces few NOTEs that we know we can safely ignore because they aren't relevant to CRAN.

* "installed package size..."
* "maintainer is..."

The mechanism we use for this is to allow a maximum number of NOTEs in CI

* https://github.com/microsoft/LightGBM/blob/2c958dd47142b90993736ba82d7cae2f33e38245/.ci/test_r_package.sh#L189
* https://github.com/microsoft/LightGBM/blob/2c958dd47142b90993736ba82d7cae2f33e38245/.ci/test_r_package_windows.ps1#L171

When the R package was successfully released to CRAN in version 3.1.0, the "maintainer is..." NOTE was no longer thrown. Because we didn't notice this and didn't change the `ALLOWED_CHECK_NOTES` threshold in CI, this "partial argument match" was note caught by CI on #3617 

## How to close this issue

Open a PR with the following changes

* fix `getinfo(data = ...` calls to `get_info(dataset = ...`
* change the CI scripts for the R package to not allow ANY notes, and ignore specific ones using the environment variables available for configuring R CMD CHECK: https://cran.r-project.org/doc/manuals/r-devel/R-ints.html#Tools

jameslamb:
This PR fixes #3629 , and fixes the R package CI to be sure that similar issues don't arise in the future.

Also fixes #3616

For CI, I'm proposing that instead of having a limit on the NUMBER of R CMD CHECK NOTES, we just ignore very specific NOTEs that we know are not problematic for CRAN. I think that will be less susceptible to issues like #3629 and give us more confidence that CI is catching problems that would be caught by CRAN.

This is blocking release 3.1.1 (#3611)

### DIFF
--- a/.ci/test_r_package.sh
+++ b/.ci/test_r_package.sh
@@ -190,16 +190,8 @@ if [[ $check_succeeded == "no" ]]; then
     exit -1
 fi
 
-if grep -q -R "NOTE" "$LOG_FILE_NAME"; then
-    echo "NOTEs have been found by R CMD check!"
-    exit -1
-fi
-if grep -q -R "WARNING" "$LOG_FILE_NAME"; then
-    echo "WARNINGS have been found by R CMD check!"
-    exit -1
-fi
-if grep -q -R "ERROR" "$LOG_FILE_NAME"; then
-    echo "ERRORs have been found by R CMD check!"
+if grep -q -E "NOTE|WARNING|ERROR" "$LOG_FILE_NAME"; then
+    echo "NOTEs, WARNINGs, or ERRORs have been found by R CMD check"
     exit -1
 fi
 
--- a/.ci/test_r_package.sh
+++ b/.ci/test_r_package.sh
@@ -190,19 +190,16 @@ if [[ $check_succeeded == "no" ]]; then
     exit -1
 fi
 
+if grep -q -R "NOTE" "$LOG_FILE_NAME"; then
+    echo "NOTEs have been found by R CMD check!"
+    exit -1
+fi
 if grep -q -R "WARNING" "$LOG_FILE_NAME"; then
     echo "WARNINGS have been found by R CMD check!"
     exit -1
 fi
-
-ALLOWED_CHECK_NOTES=0
-NUM_CHECK_NOTES=$(
-    cat ${LOG_FILE_NAME} \
-        | grep -e '^Status: .* NOTE.*' \
-        | sed 's/[^0-9]*//g'
-)
-if [[ ${NUM_CHECK_NOTES} -gt ${ALLOWED_CHECK_NOTES} ]]; then
-    echo "Found ${NUM_CHECK_NOTES} NOTEs from R CMD check. Only ${ALLOWED_CHECK_NOTES} are allowed"
+if grep -q -R "ERROR" "$LOG_FILE_NAME"; then
+    echo "ERRORs have been found by R CMD check!"
     exit -1
 fi
 
--- a/.ci/test_r_package.sh
+++ b/.ci/test_r_package.sh
@@ -11,6 +11,15 @@ export PATH="$R_LIB_PATH/R/bin:$PATH"
 # https://stat.ethz.ch/pipermail/r-package-devel/2020q3/005930.html
 export _R_CHECK_SYSTEM_CLOCK_=0
 
+# ignore R CMD CHECK NOTE checking how long it has
+# been since the last submission
+export _R_CHECK_CRAN_INCOMING_REMOTE_=0
+
+# CRAN ignores the "installed size is too large" NOTE,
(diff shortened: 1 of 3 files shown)


.ci/test_r_package.sh
```diff
@@ -186,7 +195,7 @@ if grep -q -R "WARNING" "$LOG_FILE_NAME"; then
     exit -1
 fi
 
-ALLOWED_CHECK_NOTES=2
+ALLOWED_CHECK_NOTES=0
```
StrikerRUS, line 198: Great to see that we can simplify NOTEs checking by just some `export`s!

I think we can now change this to simple `if grep -q -R "NOTE ..."`, can't we?

jameslamb: oh sure, I can do that

jameslamb: while we're thinking about it, I think we should add a fix for https://github.com/microsoft/LightGBM/issues/3616 in this PR too:

```shell
if grep -q -R "ERROR" "$LOG_FILE_NAME"; then
    echo "ERRORs have been found by R CMD check!"
    exit -1
fi
```

StrikerRUS (approved):
Thanks!

Just one minor suggestion to simplify the code for your consideration: use one `grep -E "NOTE|WARNING|ERROR" ...` instead of three separated `if`s (and [similar for PowerShell](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/select-string?view=powershell-7.1#parameters)).

jameslamb:
> Thanks!
> 
> Just one minor suggestion to simplify and speedup the code for your consideration: use one `grep -E "NOTE|WARNING|ERROR" ...` instead of three separated `if`s (and [similar for PowerShell](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/select-string?view=powershell-7.1#parameters)).

I like it! Added in https://github.com/microsoft/LightGBM/pull/3630/commits/d9ca1c4c73a8652be67445eb054378f5c8c98bd4. I tested on may Mac with `bash` and `pwsh`.

guolinke (approved):

PR #3638, opened 2020-12-10
[docs] added daal4py to related projects

StrikerRUS:
Found it from here: https://medium.com/intel-analytics-software/improving-the-performance-of-xgboost-and-lightgbm-inference-3b542c03447e.


jameslamb (approved):
nice find!

PR #2976, opened 2020-04-06
Support UTF-8 characters in feature name again

henry0312:
This commit reverts 0d59859c670b9de37bffa8a6e536497c88d9f25d.
Also see:
- https://github.com/microsoft/LightGBM/issues/2226
- https://github.com/microsoft/LightGBM/issues/2478
- https://github.com/microsoft/LightGBM/pull/2229

I reproduced the issue and as @kidotaka gave us a great survey in #2226,
I don't conclude that the cause is UTF-8, but "an empty string (character)".
Therefore, I revert "throw error when meet non ascii (#2229)" whose commit hash
is 0d59859c670b9de37bffa8a6e536497c88d9f25d, and add support feature names as UTF-8 again.

----

### reprodude the issue

The below code raises `lightgbm.basic.LightGBMError: Wrong size of feature_names`

```python
import lightgbm
import numpy
from matplotlib import pyplot

numpy.random.seed(42)

train_x= numpy.random.normal(size=(1000, 4))
valid_x= numpy.random.normal(size=(100, 4))
train_t = numpy.random.random(1000)
valid_t = numpy.random.random(100)

train_lgb = lightgbm.Dataset(train_x, train_t)
valid_lgb = lightgbm.Dataset(valid_x, valid_t, reference=train_lgb)

# This has non-ascii strings but an empty string.
feature_names = ['F_零', 'F_一', 'F_二', '']

params = {
    'boosting_type': 'gbdt',
    'objective': 'regression',
    'metric': {'l2', 'l1'},
    'num_leaves': 31,
    'learning_rate': 0.05,
    'feature_fraction': 0.9,
    'bagging_fraction': 0.8,
    'bagging_freq': 5,
    'verbose': 0,
    'seed': 42,
}

print('Starting training...')
evals_result = {}
gbm = lightgbm.train(params,
                     train_lgb,
                     num_boost_round=20,
                     valid_sets=valid_lgb,
                     feature_name=feature_names,
                     evals_result=evals_result,
                     early_stopping_rounds=5)

print('Plotting feature importances...')
ax = lightgbm.plot_importance(gbm, ignore_zero=False)
pyplot.show()
```

### use utf-8 characters

You can see there is no problem with using utf-8 in feature names.

```python
import lightgbm
import numpy
from matplotlib import pyplot

numpy.random.seed(42)

train_x= numpy.random.normal(size=(1000, 4))
valid_x= numpy.random.normal(size=(100, 4))
train_t = numpy.random.random(1000)
[...]

PR #3647, opened 2020-12-15
[R-package] add support for non-ASCII feature names (fixes #2983)

### LINKED ISSUE #2983: [R-package] Add support for non-ASCII feature names
## Summary

See #2976 for background. That PR adds support for non-ASCII features back to the C++ and Python libraries. A test in the R package was added in that PR but was breaking for reasons that weren't obvious, see for example https://github.com/microsoft/LightGBM/pull/2976#issuecomment-609905640.

I _think_ that closing this issue will mean changing `lgb.encode.char()`. You'll know that a f ix is working if you remove the  `testthat::skip()` call added in #2976 .

jameslamb:
I started working on adding support for non-ASCII features in the R package tonight, and I think this might have already been fixed! I'm not sure how, but my bet is @AlbertoEAF 's work in #3405 , since that touched model reading / writing code.

I tested with R Hub and this is even now working on Solaris!

### Notes for Reviewers

The change in this PR is just to stop skipping a test that is no longer broken. However, I gave this the label `feature` because I think it makes sense for "you can now use non-ASCII features names" to show up in the "New Features" part of the changelog for the next release, even if this PR really didn't do the work to add them and @AlbertoEAF actually did all the hard work 😀


jameslamb:
ok it looks like this is still failing on Windows :confused: 

I'll change this to work in progress and work on it tonight.

AlbertoEAF:
Hi James :) I deserve no credits here, this came in the v3.0 release, I think it was this PR from @henry0312 https://github.com/microsoft/LightGBM/pull/2976/files.

jameslamb:

I learned a lot from ["UTF-8 Support on Windows"](https://developer.r-project.org/Blog/public/2020/05/02/utf-8-support-on-windows/) and ["Windows/UTF-8 Build of R and CRAN Packages"](https://developer.r-project.org/Blog/public/2020/07/30/windows/utf-8-build-of-r-and-cran-packages/index.html) by @kalibera.

> "R internally allows strings to be represented in the current native encoding, in UTF-8 and in Latin 1. When interacting with the operating system or external libraries, all these representations have to be converted to native encoding. On Linux and macOS today this is not a problem, because the native encoding is UTF-8, so all Unicode characters are supported. On Windows, the native encoding cannot be UTF-8 nor any other that could represent all Unicode characters."

I think that on Windows, we cannot assume that non-ASCII feature names will survive the round trip between C++ and R ( specifically with `Booster$dump_model()`, which relies on [json11 code that calls `encode_utf8()`](https://github.com/microsoft/LightGBM/blob/967b45c68665488ac872f77848f516d84a70726c/src/io/json11.cpp#L494)).

Before this PR, the test on non-ASCII feature names was skipped completely on all operating systems.

Now, that test runs on all operating systems, but on Windows we expect the feature names in `Booster$dump_model()` to be changed. 

I re-submitted on R Hub and can confirm these tests pass on 32-bit Windows and Solaris.

guolinke (approved):

R-package/tests/testthat/test_basic.R
```diff
@@ -1185,10 +1186,21 @@ test_that("lgb.train() supports non-ASCII feature names", {
...
   dumped_model <- jsonlite::fromJSON(bst$dump_model())
-  expect_identical(
-    dumped_model[["feature_names"]]
-    , feature_names
-  )
+
+  # UTF-8 strings are not well-supported on Windows
+  # * https://developer.r-project.org/Blog/public/2020/05/02/utf-8-support-on-windows/
+  # * https://developer.r-project.org/Blog/public/2020/07/30/windows/utf-8-build-of-r-and-cran-packages/index.html
+  if (!ON_WINDOWS) {
```
jameslamb, line 1193: /gha run r-valgrind

jameslamb: https://github.com/microsoft/LightGBM/actions/runs/458658890

jameslamb: valgrind is happy

```text
==1913== LEAK SUMMARY:
==1913==    definitely lost: 0 bytes in 0 blocks
==1913==    indirectly lost: 0 bytes in 0 blocks
==1913==      possibly lost: 336 bytes in 1 blocks
==1913==    still reachable: 309,428,667 bytes in 55,367 blocks
==1913==                       of which reachable via heuristic:
==1913==                         newarray           : 4,264 bytes in 1 blocks
==1913==         suppressed: 0 bytes in 0 blocks
==1913== Reachable blocks (those to which a pointer was found) are not shown.
==1913== To see them, rerun with: --leak-check=full --show-leak-kinds=all
==1913== 
==1913== For lists of detected and suppressed errors, rerun with: -s
==1913== ERROR SUMMARY: 3 errors from 2 contexts (suppressed: 0 from 0)
```

PR #3651, opened 2020-12-16
Add new task type: "save_binary".

cyfdecyf:
The main motivation of adding `save_binary` task type is to save memroy.

The python function `Dataset.save_binary()` has two drawbacks:

1. Uses much more memory compared to data itself
    - triple memory usage to be more exact, but I didn't investigate this too deep
2. Can not use the `two_round` parameter compared to command line verison to reduce memory usage

If you have interest to merge this PR, I'd also update the corresponding docs.

### DIFF
--- a/include/LightGBM/config.h
+++ b/include/LightGBM/config.h
@@ -27,7 +27,7 @@ namespace LightGBM {
 
 /*! \brief Types of tasks */
 enum TaskType {
-  kTrain, kPredict, kConvertModel, KRefitTree
+  kTrain, kPredict, kConvertModel, KRefitTree, kSaveBinary
 };
 const int kDefaultNumLeaves = 31;
 
--- a/include/LightGBM/config.h
+++ b/include/LightGBM/config.h
@@ -102,6 +102,7 @@ struct Config {
   // desc = ``predict``, for prediction, aliases: ``prediction``, ``test``
   // desc = ``convert_model``, for converting model file into if-else format, see more information in `Convert Parameters <#convert-parameters>`__
   // desc = ``refit``, for refitting existing models with new data, aliases: ``refit_tree``
+  // desc = ``save_binary``, load train (and validation) data then save dataset to binary file. Typical usage: ``save_binary`` first, then run multiple ``train`` tasks in parallel using the saved binary file
   // desc = **Note**: can be used only in CLI version; for language-specific packages you can use the correspondent functions
   TaskType task = TaskType::kTrain;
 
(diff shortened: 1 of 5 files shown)


cyfdecyf:
I've looked at this [failed check](https://github.com/microsoft/LightGBM/pull/3651/checks?check_run_id=1561488879), which is a compilation failure caused by missing `cstdint` header in `meta.h`. Guess this is not caused by my change?

jameslamb:
> I've looked at this [failed check](https://github.com/microsoft/LightGBM/pull/3651/checks?check_run_id=1561488879), which is a compilation failure caused by missing `cstdint` header in `meta.h`. Guess this is not caused by my change?

no you can ignore that one, it's very unreliable. Sorry! I'll re-run manually

cyfdecyf (commented):
Comment for some logging tweaks.

guolinke (approved):

include/LightGBM/config.h
```diff
@@ -27,7 +27,7 @@ namespace LightGBM {
 
 /*! \brief Types of tasks */
 enum TaskType {
-  kTrain, kPredict, kConvertModel, KRefitTree
+  kTrain, kPredict, kConvertModel, KRefitTree, kSaveBinary
```
StrikerRUS, line 30: @cyfdecyf Thanks for your contribution! Could you please resolve all conflicts and document new enum option?
https://github.com/microsoft/LightGBM/blob/ed651e86729477e7d2699534d63a5cace87f7878/include/LightGBM/config.h#L95-L106
After adding description of `save_binary` task, run this script to regenerate documentation.
https://github.com/microsoft/LightGBM/blob/master/helpers/parameter_generator.py

cyfdecyf: Sorry for the late reply. Conflicts are resolved and added document for the new enum option.

include/LightGBM/config.h
```diff
@@ -102,6 +102,7 @@ struct Config {
   // desc = ``predict``, for prediction, aliases: ``prediction``, ``test``
   // desc = ``convert_model``, for converting model file into if-else format, see more information in `Convert Parameters <#convert-parameters>`__
   // desc = ``refit``, for refitting existing models with new data, aliases: ``refit_tree``
+  // desc = ``save_binary``, load train (and validation) data then save dataset to binary file. Typical usage: ``save_binary`` first, then run multiple ``train`` tasks in parallel using the saved binary file
```
StrikerRUS, line 105: Many thanks for `Typical usage`! 👍

StrikerRUS (approved):
LGTM! Thanks for your contribution!

PR #3656, opened 2020-12-18
[docs] fix R documentation builds (fixes #3655)

### LINKED ISSUE #3655: [R-package] pkgdown docs for individual functions are broken
The R docs at https://lightgbm.readthedocs.io/en/latest/R/index.html are rendering correctly:

image

as is the reference page at https://lightgbm.readthedocs.io/en/latest/R/reference/index.html

image

But if you click any of the specific function names on the reference page, the resulting links are broken. For example, https://lightgbm.readthedocs.io/en/latest/R/reference/lgb.train.html

image

jameslamb:
This PR attempts to fix the broken R documentation (see #3655). Building the docs requires installing the R package. Now that #3405 has been merged, LightGBM's submodules have to be initialized for the package to be built successfully, and I think they were not on readthedocs builds.

latest build: https://readthedocs.org/projects/lightgbm/builds/12567523/

```text
* checking for LF line-endings in source and make files and shell scripts
WARNING: directory ‘lightgbm/src/external_libs/fast_double_parser’ is empty
WARNING: directory ‘lightgbm/src/external_libs/fmt’ is empty
```

and then later

```text
* installing to library ‘/home/docs/.conda/envs/r_env/lib/R/library’
* installing *source* package ‘lightgbm’ ...
** using staged installation
** libs
installing via 'install.libs.R' to /home/docs/.conda/envs/r_env/lib/R/library/00LOCK-lightgbm/00new/lightgbm
Building lib_lightgbm
In file included from /tmp/Rtmp4mfEea/R.INSTALL30a33e035de/lightgbm/src/include/LightGBM/config.h:16,
                 from /tmp/Rtmp4mfEea/R.INSTALL30a33e035de/lightgbm/src/include/LightGBM/boosting.h:8,
                 from /tmp/Rtmp4mfEea/R.INSTALL30a33e035de/lightgbm/src/src/boosting/boosting.cpp:5:
/tmp/Rtmp4mfEea/R.INSTALL30a33e035de/lightgbm/src/include/LightGBM/utils/common.h:35:10: fatal error: ../../../external_libs/fmt/include/fmt/format.h: No such file or directory
   35 | #include "../../../external_libs/fmt/include/fmt/format.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### Notes for Reviewers

@StrikerRUS could you please enable readthedocs builds for this branch?

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -248,8 +248,9 @@ def generate_r_docs(app):
     source /home/docs/.conda/bin/activate r_env
     export TAR=/bin/tar
     cd {0}
+    git submodule update --init --recursive
     export R_LIBS="$CONDA_PREFIX/lib/R/library"
-    Rscript build_r.R
+    Rscript build_r.R || exit 1
     cd {1}
     Rscript -e "roxygen2::roxygenize(load = 'installed')"
     Rscript -e "pkgdown::build_site( \


docs/conf.py
```diff
@@ -248,8 +248,9 @@ def generate_r_docs(app):
     source /home/docs/.conda/bin/activate r_env
     export TAR=/bin/tar
     cd {0}
+    git submodule update --init --recursive
```
StrikerRUS, line 251: Thanks for the fix! I think it will be better to use config file for such things.
https://docs.readthedocs.io/en/stable/config-file/v2.html#submodules

Could you please try include the following lines 
```
submodules:
  include: all
  recursive: true
```
into
https://github.com/microsoft/LightGBM/blob/master/.readthedocs.yml?

StrikerRUS: P.S. I've enabled RTD builds for the `docs/jlamb` branch and will keep it turned on as you'd asked.

docs/conf.py
```diff
@@ -248,8 +248,9 @@ def generate_r_docs(app):
...
     export R_LIBS="$CONDA_PREFIX/lib/R/library"
-    Rscript build_r.R
+    Rscript build_r.R || exit 1
```
StrikerRUS, line 253: Please use `-1` for the consistency. It may help to simply a search query in the repo if we decide to change something in the future.

Shouldn't `Rscript -e "pkgdown::build_site()` cause builds to fail as well?

StrikerRUS:
You are right in the root cause of the failures! Builds are now passing:

image

docs/conf.py
jameslamb: yeah you're right

docs/conf.py
jameslamb: sure, I can do that. And thanks for keeping the branch up!

StrikerRUS (approved):
Thanks for updating the config and forcing build to fail!

docs/conf.py
StrikerRUS: > And thanks for keeping the branch up!

image

This will auto-disable RTD builds! 🤣

PR #3661, opened 2020-12-19
[ci] update boost version

StrikerRUS:


StrikerRUS:
OK, sorry! I didn't check but all builds were actually failed:

image

Issue #2613, opened 2019-12-05
Load back saved parameters with save_model to Booster object


everdark:
## Environment info

Operating System:
Windows 10 (Same result on both Windows and WSL)

CPU/GPU model:
Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz

C++/Python/R version: 
Python 3.7

LightGBM version or commit hash:
2.3.1 installed by pip

## Reproducible examples

```python
import json
import lightgbm as lgb
import pandas as pd
import numpy as np
from sklearn.metrics import mean_squared_error

try:
    import cPickle as pickle
except BaseException:
    import pickle

print('Loading data...')
# load or create your dataset
df_train = pd.read_csv('../binary_classification/binary.train', header=None, sep='\t')
df_test = pd.read_csv('../binary_classification/binary.test', header=None, sep='\t')
W_train = pd.read_csv('../binary_classification/binary.train.weight', header=None)[0]
W_test = pd.read_csv('../binary_classification/binary.test.weight', header=None)[0]

y_train = df_train[0]
y_test = df_test[0]
X_train = df_train.drop(0, axis=1)
X_test = df_test.drop(0, axis=1)

num_train, num_feature = X_train.shape

# create dataset for lightgbm
# if you want to re-use data, remember to set free_raw_data=False
lgb_train = lgb.Dataset(X_train, y_train,
                        weight=W_train, free_raw_data=False)
lgb_eval = lgb.Dataset(X_test, y_test, reference=lgb_train,
                       weight=W_test, free_raw_data=False)

# specify your configurations as a dict
params = {
    'boosting_type': 'gbdt',
    'objective': 'binary',
    'metric': 'binary_logloss',
    'num_leaves': 31,
    'learning_rate': 0.05,
    'feature_fraction': 0.9,
    'bagging_fraction': 0.8,
    'bagging_freq': 5,
    'verbose': 0
}

# generate feature names
feature_name = ['feature_' + str(col) for col in range(num_feature)]

print('Starting training...')
# feature_name and categorical_feature
gbm = lgb.train(params,
                lgb_train,
                num_boost_round=10,
                valid_sets=lgb_train,  # eval training data
                feature_name=feature_name,
                categorical_feature=[21])

print(gbm.params) # Check params.
gbm.save_model('model.txt')

gbm2 = lgb.Booster(model_file='model.txt')
print(gbm2.params) # Nothing.
```

Coding example above is directly borrowed from official example `advanced_example.py`
[...]

PR #3662, opened 2020-12-20
[R-package] allow access to params in Booster

jameslamb:
I've been trying to check LightGBM questions on Stack Overflow more frequently. A search today led me to this question from two years ago: https://stackoverflow.com/questions/49674112/accessing-lightgbm-model-parameters/50115384#50115384.

It pointed out that there's no way to access the parameters from `Booster` object. This is something that the Python package has had for a while.

Storing the parameters in the Booster can be helpful for cases where you've saved a model and then later load it in a session that doesn't have the original training code available. For example, you might want to be sure that the parameters you use for training continuation are the same as those you used in the original training. Or you might want to use parameters as metadata in a model monitoring system.

This pull request adds a new public attribute, `Booster$params`. Just like `Booster.params` in the Python package, it includes all passed-in parameters AND any parameters on the Dataset used for training (such as `max_bin`).

This PR also adds the first set of tests on `saveRDS.lgb.booster()` and `readRDS.lgb.booster()`, which were previously not covered by any tests.


StrikerRUS:
(Shouldn't prevent to merge this PR) but I just want to say that a more general fix will be loading params from cpp side: https://github.com/microsoft/LightGBM/issues/2613#issuecomment-562216072

src/c_api.cpp
```diff
@@ -294,13 +294,15 @@ class Booster {
...
+    if (param.count("num_class") && new_config.num_class != config_.num_class) {
       Log::Fatal("Cannot change num_class during training");
     }
-    if (param.count("boosting")) {
+    if (param.count("boosting") && new_config.boosting != config_.boosting) {
       Log::Fatal("Cannot change boosting during training");
     }
-    if (param.count("metric")) {
+    if (param.count("metric") && new_config.metric != config_.metric) {
       Log::Fatal("Cannot change metric during training");
```
jameslamb, line 306: I originally had this PR only changing the R package, but then ran into this error

> Error: [LightGBM] [Fatal] Cannot change metric during training

This is thrown even if you aren't actually CHANGING `metric`. I think the change here in `c_api` is closer to the desired behavior, only throwing an error if the parameter is being changed.

R-package/tests/testthat/test_custom_objective.R
```diff
@@ -39,6 +39,8 @@ param <- list(
 )
 num_round <- 10L
 
+# this is failing because of the checks in src/c_api.cpp#294,
+# which says "if you pass params into reset-params, throw an error"
```
jameslamb, line 43:
```suggestion
```

guolinke (approved):

src/c_api.cpp
jameslamb: /gha run r-valgrind

jameslamb: https://github.com/microsoft/LightGBM/actions/runs/458548423

jameslamb: ✅ valgrind is happy

PR #3682, opened 2020-12-26
[ci] speed up files downloading for Windows CI

StrikerRUS:
So weird...
Refer to https://stackoverflow.com/a/43477248.

@tpboudreau Now everything is OK with OpenCL downloading speed 😄 
image


tpboudreau:
That's a great find @StrikerRUS -- what a horrible bug.

The changes look good to me.

jameslamb (approved):
wow great find! This is so weird haha

PR #3686, opened 2020-12-27
[R-package] remove unused R_AS_INT64

jameslamb:
As of refactoring for 3.1.0 (not sure exactly which PR), `R_AS_INT64` is unused. This PR removes it.


R-package/src/R_object_helper.h
```diff
@@ -108,8 +108,6 @@ typedef union { VECTOR_SER s; double align; } SEXPREC_ALIGN;
 
 #define R_AS_INT(x)    (*(reinterpret_cast<int*> DATAPTR(x)))
 
-#define R_AS_INT64(x)  (*(reinterpret_cast<int64_t*> DATAPTR(x)))
-
```
jameslamb, line 112: /gha run r-valgrind

jameslamb: https://github.com/microsoft/LightGBM/runs/1613261778?check_suite_focus=true

guolinke (approved):

PR #3688, opened 2020-12-27
[ci] remove unused CI checks, other small simplifications

jameslamb:
This PR proposes a few of the small changes from #3672, which I think should be non-controversial. Hopefully this will make that PR's diff smaller and make it easier to review.

* removes `[[ $TRAVIS == "true" ]]` checks on `check-docs` and `lint` tasks
    - the fact that we happen to currently run those tasks on Travis doesn't need to be guarded in the code
* sets `AZURE=true` as a global variable in Azure DevOps config, reducing duplication
* changes the condition `if [[ $TRAVIS == "true" ]] || [[ $GITHUB_ACTIONS == "true" ]]` to simply `if [[ $SETUP_CONDA != "false" ]];`
     - this condition was capturing "Linux jobs on Azure run in a container that already has conda, and all other jobs need to set it up". I think it's more explicit and less error-prone to just have `setup.sh` install conda by default, and allow turning that off with a variable `SETUP_CONDA`

### DIFF
--- a/.ci/setup.sh
+++ b/.ci/setup.sh
@@ -43,13 +43,13 @@ else  # Linux
         chmod +x cmake.sh
         ./cmake.sh --prefix=/usr/local --exclude-subdir
     fi
-    if [[ $TRAVIS == "true" ]] || [[ $GITHUB_ACTIONS == "true" ]]; then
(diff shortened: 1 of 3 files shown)


StrikerRUS (approved):
Thanks!

PR #3690, opened 2020-12-28
[docs] add doc on min_data_in_leaf approximation (fixes #3634)

### LINKED ISSUE #3634: min_data_in_leaf not exactly honoured

LightGBM component:  Python package

## Environment info

Operating System: MacOS 11.0.1

CPU/GPU model:2,3 GHz Dual-Core Intel Core i5 / Intel Iris Plus Graphics 640 1536 MB

Python version: 3.8.6

---

When I train a model with `min_data_in_leaf = <n>` and then check the output with `trees_to_dataframe()` filtering only leaf nodes (i.e. `df.split_feature.isna()`), many times the `count` column is < `<n>`. Changing `<n>` shows positive correlation with the reported count, but it's not behaving as a hard limit. I'm pretty sure this is caused by [this code](https://github.com/microsoft/LightGBM/blob/0655d67cc1ee28d536a773a6c5d978d38a0e56c1/src/treelearner/serial_tree_learner.cpp#L297):

```
  // no enough data to continue
  if (num_data_in_right_child < static_cast<data_size_t>(config_->min_data_in_leaf * 2)
      && num_data_in_left_child < static_cast<data_size_t>(config_->min_data_in_leaf * 2)) {
```

So unbalanced splits in which a child has count less than `min_data_in_leaf`  still may happen. This also explains the correlation.

If this is expected to work so, then I think the documentation should be made more precise. If not, then it seems a bug to me.

If what I explained above is not enough for you to reproduce or recognize the issue, I may provide a reproducible example latter, but I think the referenced code is quite clear.

jameslamb:
This proposes a change to the `min_data_in_leaf` docs, explaining that splits can produced leaf nodes with smaller counts, per the discussion in #3634.

### DIFF
--- a/docs/Parameters.rst
+++ b/docs/Parameters.rst
@@ -282,7 +282,7 @@ Learning Control Parameters
 
    -  minimal number of data in one leaf. Can be used to deal with over-fitting
 
-   -  **Note**: this is an approximation based on the hessian, so occasionally you may observe splits which produce leaf nodes that have less than this many observations
+   -  **Note**: this is an approximation based on the Hessian, so occasionally you may observe splits which produce leaf nodes that have less than this many observations
 
 -  ``min_sum_hessian_in_leaf`` :raw-html:`<a id="min_sum_hessian_in_leaf" title="Permalink to this parameter" href="#min_sum_hessian_in_leaf">&#x1F517;&#xFE0E;</a>`, default = ``1e-3``, type = double, aliases: ``min_sum_hessian_per_leaf``, ``min_sum_hessian``, ``min_hessian``, ``min_child_weight``, constraints: ``min_sum_hessian_in_leaf >= 0.0``
 
--- a/docs/Parameters.rst
+++ b/docs/Parameters.rst
@@ -282,6 +282,8 @@ Learning Control Parameters
 
    -  minimal number of data in one leaf. Can be used to deal with over-fitting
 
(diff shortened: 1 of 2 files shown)


include/LightGBM/config.h
```diff
@@ -273,6 +273,7 @@ struct Config {
   // alias = min_data_per_leaf, min_data, min_child_samples
   // check = >=0
   // desc = minimal number of data in one leaf. Can be used to deal with over-fitting
+  // desc = **Note**: this is an approximation based on the hessian, so occasionally you may observe splits which produce leaf nodes that have less than this many observations
```
StrikerRUS, line 276:
```suggestion
  // desc = **Note**: this is an approximation based on the Hessian, so occasionally you may observe splits which produce leaf nodes that have less than this many observations
```

docs/Parameters.rst
```diff
@@ -282,6 +282,8 @@ Learning Control Parameters
 
    -  minimal number of data in one leaf. Can be used to deal with over-fitting
 
+   -  **Note**: this is an approximation based on the hessian, so occasionally you may observe splits which produce leaf nodes that have less than this many observations
```
StrikerRUS, line 285:
```suggestion
   -  **Note**: this is an approximation based on the Hessian, so occasionally you may observe splits which produce leaf nodes that have less than this many observations
```

StrikerRUS (approved):
LGTM, thanks!

PR #3522, opened 2020-11-03
Optimization of row-wise histogram construction

shiyu1994:
This PR is to optimize row-wise histogram construction. For dense multi-value bin, the original bins, without offsets added, are stored, which saves memory when the max bin number per feature is small. For both dense and sparse multi-value, we try to use single precision floating point in histogram buffers, and use avx-256 instructions to speedup histogram construction.

PR #3694, opened 2020-12-30
Fix bug in corner case with histogram bin mismatch

shiyu1994:
This is to fix corner case bug due to histogram bin mismatch, which is introduced by a previous PR #3522. An error can occur when all the following conditions are satisfied
1. Row-wise histogram construction is used or feature subsampling is used.
2. The first feature group is a multi-value feature group.
3. The first group needs an extra padding bin.


guolinke (approved):

guolinke:
@shiyu1994 can you add a simple test for this corner case?

guolinke:
@shiyu1994 You can mark the test case as a todo item. I will merge this first.

Issue #3292, opened 2020-08-10
The result of feature parallel algorithm is not correct


songchuangyuan:
1. Directly use the stand-alone version of the binary_classification example test, the number of rows of data is 7000, the feature is 28, set the training parameters, calculate the AUC and loss value.
2. Using the example test of two parallel machines, the data in the first step is divided vertically. The number of data rows of the first machine is 7000 and feature 14, and the number of data rows of the second machine is 7000 and feature 14. Set the same parameters as the first step, calculate the AUC and loss values.

Result: Find that the results are not consistent, and there is a big difference. if you want to do parallelism, you should get the same result as the two steps. Why not consistent?

guolinke:
Some settings will break the consistency of parallel algorithm.
could you provide your settings?

guolinke:
BTW, for the feature palallel in lightgbm, you need to pass the whole dataset (with 28 featues), to both two nodes.

songchuangyuan:
> BTW, for the feature palallel in lightgbm, you need to pass the whole dataset (with 28 featues), to both two nodes.

Thank you for your answer. I tested it again according to your method, and the result is correct. In fact, the feature parallel mode is only to speed up the training, not because the file is too large to load into the memory. What about data parallelism? 

I did the first test:
1. Directly use the stand-alone version of the binary_classification example test, the number of rows of data is 7000, the feature is 28, set the training parameters, calculate the AUC and loss value.
2. Using the example test of two parallel machines, the data in the first step is divided horizontally. The number of data rows of the first machine is 3500 and feature 28, and the number of data rows of the second machine is 3500 and feature 28. Set the same parameters as the first step, calculate the AUC and loss values.
Result: It is found that the results of the two steps are not consistent..

I did the second test:
1. Directly use the stand-alone version of the binary_classification example test, the number of rows of data is 7000, the feature is 28, set the training parameters, calculate the AUC and loss value.
2. Using the example test of two parallel machines, both machines have the complete data from the first step. Set the same parameters as the first step, calculate the AUC and loss values.
Result: It is found that the results of the two steps are still inconsistent, and the results on the two machines are also inconsistent.

So how is the data parallel mode implemented?
It also needs to have the whole dataset(7000 rows) on both two nodes, but each node only reads a part of the dataset, create bins and construct histograms respectively, and then aggregates the histogram generated on different nodes using reduce scatter? In this step of creating bins, do you only need to use part of the local dataset or the whole dataset?

[...]

PR #3708, opened 2021-01-02
[python-package] [dask] Add DaskLGBMRanker

ffineis:
Hello,

This is a PR to include support for a DaskLGBMRanker. Previously only `DaskLGBMClassifier` and `DaskLGBMRegressor` were supported. This PR was [originally](https://github.com/dask/dask-lightgbm/pull/26/files) for the dask-lightgbm repo, but is migrated here after the incorporation of the recent. lightgbm.dask module. Tests added and were passing from an image built from a modification of dockerfile-python.

Note: it's easier to use dask dataframes with the DaskLGBMRanker than it is arrays through the use of `df.set_index`, although it's possible to use Dask arrays if the trainset is built in group-wise chunks. The constraints of the ranking task force us to not break up contiguous `groups`, which is why chunks/partitions need to be constructed from `groups` somehow. No tests for sparse arrays and DaskLGBMRanker at the moment.

Thanks! Excited for this new module.
- Frank


jameslamb (commented):
sorry it's taking me so long to review and test this! I'm planning to review it tomorrow.

ffineis:
Hey yeah no problem. Unfortunately, I can't get the test_dask.py tests passing 100% reliably. The errors are random, sometimes passing, they're hard to reproduce. Failures seem limited to GPU builds and the `ranker` tests. But the `[LightGBM] [Fatal] Socket recv error, code: 104` issue I believe [predates](https://github.com/dask/dask-lightgbm/pull/23/checks?check_run_id=1402948145) these new `ranker` tests, affecting regressor/classifier.

The strange part is that when I execute the container (i.e. instead of through a bash shell), the tests have passed every time. It's only upon running `pytest test_dask.py` within a REPL in interactive mode that exceptions pop up.

Dockerfile:
```
FROM ubuntu:16.04

ARG CONDA_DIR=/opt/conda
ENV PATH $CONDA_DIR/bin:$PATH

RUN apt-get update && \
    apt-get install -y --no-install-recommends \
        ca-certificates \
        cmake \
        build-essential \
        gcc \
        g++ \
        git \
        vim \
        wget && \
    # python environment
    wget https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh && \
    /bin/bash Miniconda3-latest-Linux-x86_64.sh -f -b -p $CONDA_DIR && \
    export PATH="$CONDA_DIR/bin:$PATH" && \
    conda config --set always_yes yes --set changeps1 no && \
    # lightgbm
    conda install -q -y numpy scipy scikit-learn pandas pytest dask[array]>=2.0.0 dask[dataframe]>=2.0.0 dask[distributed]>=2.0.0 && \
    git clone --recursive --branch dask/ranker --depth 1 https://github.com/ffineis/LightGBM && \
    cd LightGBM/python-package && python setup.py install && \
    # clean
    apt-get autoremove -y && apt-get clean && \
    conda clean -a -y && \
    rm -rf /usr/local/src/*

CMD pytest -s /LightGBM/tests/python_package_test/test_dask.py -k "ranker" --log-cli-level DEBUG
```

- Instability appears to be networking-related. When I run tests from within an interactive shell on the container (not in `-it` mode), I often run into a socket-related `LightGBMError:"Biniding port 13xxx failed"` running locally. On Travis the errors include `[LightGBM] [Fatal] Socket recv error, code: 104`. Both of these issues have been hard to get answers on.
- I've tried using my own `client` fixture built from a `distributed.LocalCluster` via `distributed.Client` (below) as well as [upping the default Client timeout](https://stackoverflow.com/questions/60088134/dask-distributed-client-error-failed-to-reconnect-to-scheduler-after-10-00-s), no dice on either - I'll still occasionally get the "Binding port" exception on a random regressor/ranker/classifier test:
```
@pytest.fixture()
def client(request, loop):
    print('setting up client')
    cli = Client(loop=loop, timeout='120s')

    def teardown():
        cli.close()
        print('closing client')

    request.addfinalizer(teardown)

    return cli
```
- The test test Exceptions on Travis are being triggered on `client` teardown. I think there's a worker that hangs and it causes the `disconnect` coroutine to wait for more than the [default of 3 seconds](https://github.com/dask/distributed/blob/658fccc3268bd16226cc5815bf00eed349fe1553/distributed/utils_test.py#L595), and the Exception seems second-hand to the test Failure - i.e. I think these would go away if the underlying test failure went away.

StrikerRUS:
@ffineis 

> Wondering if you might had any thoughts on networking port/socket issues?

I'm sorry, have no ideas...

tests/python_package_test/test_dask.py
```diff
@@ -36,6 +44,102 @@ def listen_port():
...
 
 
+def r2_score(dy_true, dy_pred):
+    """Helper function taken from dask_ml.metrics: computes coefficient of determination."""
+    numerator = ((dy_true - dy_pred) ** 2).sum(axis=0)
+    denominator = ((dy_true - dy_pred.mean(axis=0)) ** 2).sum(axis=0)
+    return (1 - numerator / denominator).compute()
+
+
+def _make_ranking(n_samples=100, n_features=20, n_informative=5, gmax=1, random_gs=False, avg_gs=10, random_state=0):
```
jameslamb, line 54: Could you add docs explaining what each of these parameters means? There are a lot to reason about here and I want to be sure that people editing or adding to these tests in the future understand what's going on.

I think it's:

* `n_samples` = total number of documents (records) in the dataset
* `n_features` = total number of features in the dataset
* `n_informative` = how many of the features should be informative. If this is greater than `n_features`, the dataset will have `n_features` features, and all of them will be informative. Features are guaranteed to be "informative" for ranking by creating their values as `bias + beta*(actual_relevance_scores)`, with a random `bias` and `beta`.
* `gmax` = Maximum relevance value. Relevance scores are integers. If you set this to 2, for example, all documents in a group will get relevant scores of either 0, 1, or 2.
* `random_gs` = True if you want group size to be random, False if you want all groups to be the same size
* `avg_gs` = number of documents (records) in each group

tests/python_package_test/test_dask.py
```diff
@@ -74,15 +178,14 @@ def test_classifier(output, centers, client, listen_port):
     dask_classifier = dlgbm.DaskLGBMClassifier(time_out=5, local_listen_port=listen_port)
     dask_classifier = dask_classifier.fit(dX, dy, sample_weight=dw, client=client)
     p1 = dask_classifier.predict(dX)
-    s1 = accuracy_score(dy, p1)
+    s1 = da.average(dy == p1).compute()
```
jameslamb, line 181: why did this test have to change? If it's because of something nondeterministic in `accuracy_score()` or some precision problem that makes this test flaky, could you make that a separate PR?

python-package/lightgbm/dask.py
```diff
@@ -68,15 +68,22 @@ def _train_part(params, model_factory, list_of_parts, worker_addresses, return_m
...
+    is_ranker = model_factory.__qualname__ == 'LGBMRanker'
+
     # Concatenate many parts into one
-    parts = tuple(zip(*list_of_parts))
-    data = _concat(parts[0])
-    label = _concat(parts[1])
-    weight = _concat(parts[2]) if len(parts) == 3 else None
+    data = _concat([d['X'] for d in list_of_parts])
+    label = _concat([d['y'] for d in list_of_parts])
+    weight = _concat([d['weight'] for d in list_of_parts]) if 'weight' in list_of_parts[0] else None
```
jameslamb, line 76: I love this change, thank you so much!

Since `list_of_parts` is already guaranteed to all be data sitting on this worker at this point, I don't think we were getting any value from the `tuple()` thing.

Can you make this a separate PR? I think we could review and merge that pretty quickly, and that it's not directly related to the ranker changes.

python-package/lightgbm/dask.py
```diff
@@ -102,20 +109,30 @@ def _train(client, data, label, params, model_factory, weight=None, **kwargs):
     y : dask array of shape = [n_samples]
         The target values (class labels in classification, real numbers in regression).
     params : dict
-    model_factory : lightgbm.LGBMClassifier or lightgbm.LGBMRegressor class
+    model_factory : lightgbm.LGBMClassifier, lightgbm.LGBMRegressor, or lightgbm.LGBMRanker class
     sample_weight : array-like of shape = [n_samples] or None, optional (default=None)
-            Weights of training data.
+        Weights of training data.
+    group : array-like
+        Group/query data, used for ranking task. sum(group) = n_samples.
```
jameslamb, line 116: Can you expand this a bit more? I know that the first time I did learning-to-rank tasks on LightGBM, I was really confused about how to set up `group`.

```suggestion
        Group/query data, only used for ranking task. sum(group) = n_samples. For example,
            if you have a 100-record dataset with `group = [10, 20, 40, 10, 10]`, that means that you have
            5 groups, where the first 10 records are in the first group, records 11-30 are the second group, etc.
```

python-package/lightgbm/dask.py
```diff
@@ -102,20 +109,30 @@ def _train(client, data, label, params, model_factory, weight=None, **kwargs):
...
+        for i, d in enumerate(parts):
+            parts[i] = {**d, 'weight': weight_parts[i]}
+
+    if group is not None:
+        group_parts = _split_to_parts(group, is_matrix=False)
+        for i, d in enumerate(parts):
+            parts[i] = {**d, 'group': group_parts[i]}
 
     # Start computation in the background
+    parts = list(map(delayed, parts))
```
jameslamb, line 135: :star_struck: cool idea, I like this! I think it can be in a separate PR though, together with the "put parts into dicts" change.

tests/python_package_test/test_dask.py
```diff
@@ -36,6 +44,102 @@ def listen_port():
 listen_port.port = 13000
 
 
+def r2_score(dy_true, dy_pred):
+    """Helper function taken from dask_ml.metrics: computes coefficient of determination."""
+    numerator = ((dy_true - dy_pred) ** 2).sum(axis=0)
+    denominator = ((dy_true - dy_pred.mean(axis=0)) ** 2).sum(axis=0)
+    return (1 - numerator / denominator).compute()
```
jameslamb, line 51: why was it necessary to replace the dask-ml one with this function?

If that's just to cut out our testing dependency on `dask-ml` (which would be much appreciated!), could you make that a separate PR with this change plus removing `dask-ml` from here:

* https://github.com/microsoft/LightGBM/blob/f2695daba2506d047831e39aeb521e94adec5e8d/.ci/test.sh#L71

I think that PR would be small and non-controversial, so we could review and merge it quickly.

tests/python_package_test/test_dask.py
```diff
@@ -180,7 +283,49 @@ def test_regressor_local_predict(client, listen_port):
 
     # Predictions and scores should be the same
     assert_eq(p1, p2)
-    assert_eq(s1, s2)
+    assert np.isclose(s1, s2, rtol=1e-4)
+
+
+@pytest.mark.parametrize('output', ['array', 'dataframe'])
+def test_ranker(output, client, listen_port):
+    X, y, w, g, dX, dy, dw, dg = _create_ranking_data(output=output)
```
jameslamb, line 291: could you add one more test like this, but where the groups have different numbers of records? Like 100 records with `group = [5, 15, 25, 5, 5, 5, 10, 10, 10, 10]` or something? There's such a tight relationship here between the group setup and the way the Dask collection is partitioned, I think a test like that could catch some types of issues not caught by the current tests.

jameslamb (changes requested):
This is looking awesome, thank you! Sorry it took so long to get you a first review. Now that I have a testing framework set up for reviewing LightGBM Dask stuff (https://github.com/jameslamb/lightgbm-dask-testing), I'll be able to give you feedback a lot more quickly.

I tested this PR on that setup and everything looks good! I like the way you set up the unit tests, thank you so much for all the effort you put into that!

I left some recommended changes. I also have one more quesiton below.

> although it's possible to use Dask arrays if the trainset is built in group-wise chunks. The constraints of the ranking task force us to not break up contiguous groups, which is why chunks/partitions need to be constructed from groups somehow.

Does this mean "one chunk / partition per group" or do you mean "all documents for a group need to be in the same chunk / partition"? Like, if I have 10 groups could I split them into 2 partitions?

.ci/test.sh
```diff
@@ -52,7 +52,7 @@ if [[ $TRAVIS == "true" ]] && [[ $TASK == "lint" ]]; then
             "r-lintr>=2.0"
     pip install --user cpplint
     echo "Linting Python code"
-    pycodestyle --ignore=E501,W503 --exclude=./compute,./eigen,./.nuget,./external_libs . || exit -1
+    pycodestyle --ignore=E402,E501,W503 --exclude=./compute,./eigen,./.nuget,./external_libs . || exit -1
```
StrikerRUS, line 55: Why was it needed? I can't find any new imports other that ones at the top of modules... Am I missing something?

ffineis:
Hey thanks for the review! I'll respond/make changes tomorrow.

.ci/test.sh
ffineis: Ah, this was for test_dask. The check for whether or not the test is running on Linux system occurs before the remaining imports.

StrikerRUS: Hmmm, strange... We have this Linux-test line in the `master` branch and there are no warnings.
https://github.com/microsoft/LightGBM/blob/f2695daba2506d047831e39aeb521e94adec5e8d/tests/python_package_test/test_dask.py#L4-L11

ffineis: weird! Huh. I'll remove E402 and see how the lint testing goes during test builds today.

ffineis:
> Does this mean "one chunk / partition per group" or do you mean "all documents for a group need to be in the same chunk / partition"? Like, if I have 10 groups could I split them into 2 partitions?

Yep, the latter - you can place whole separate groups into different partitions as long as all documents in a group are in the same partition. The partitioning needs to work like a ` GROUP BY` statement. Basically, if documents in the same group were broken up across separate partitions, then the value of the `group` vector passed to the LightGBMRanker constructor on each local worker would not be accurate, and this would degrade the quality of the gradients calculated under the lambdarank objective.

Dask.DataFrame is amenable to preserving group IDs by [using set_index](https://stackoverflow.com/questions/49532824/dask-dataframe-split-partitions-based-on-a-column-or-function). Dask.array does not seem to have a similar "group partitions according to this column" functionality, hence the use of `da.concatenate`.

tests/python_package_test/test_dask.py
```diff
@@ -180,7 +283,49 @@ def test_regressor_local_predict(client, listen_port):
...
+    local_ranker.fit(X, y, sample_weight=w, group=g)
+    rnkvec_local = local_ranker.predict(X)
+
+    # distributed ranker should be able to rank decently well.
+    dcor = spearmanr(rnkvec_dask, y).correlation
+    assert dcor > 0.6
+
+    # relative difference between distributed ranker and local ranker spearman corr should be small.
+    lcor = spearmanr(rnkvec_local, y).correlation
+    assert np.abs(dcor - lcor) / lcor < 0.03
```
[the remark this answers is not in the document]
ffineis: Agree this can just be an absolute difference.

tests/python_package_test/test_dask.py
ffineis: Yep, absolutely. Didn't know how much docstrings in the test modules were emphasized. Heads up, I'm also going to add a `group` parameter that's analogous to the LightGBMRanker's `group` parameter, so that we can specify group sizes directly (as requested). When specified, the `group` vector will override `n_samples` and `avg_gs` and `#data = sum(group)`.

tests/python_package_test/test_dask.py
ffineis: Nope, had to do with eliminating the `dask-ml` dependency.

I saw that `dask-ml` was missing from setup.py, so I just tried to remove that dependency since we're only using it for two `sklearn.metrics`-emulating functions anyway (`accuracy_score` and `r2_score`). Since dask-ml is now part of test.sh I can just keep this as `accuracy_score`.

tests/python_package_test/test_dask.py
ffineis: Cool, yeah I'll follow up 3708 with the removal of dask-ml as a test dependency. See reply to comment re: accuracy_score (above).

python-package/lightgbm/dask.py
ffineis: Sure, sorry this is my first OS PR and I guess the scope of the changes here is more than just "add ranker support" and is a non-insignificant stray from how xgboost.dask is written. Will do.

python-package/lightgbm/dask.py
ffineis: Oh yeah it's definitely not particularly well-documented. This docstring was copied from the docs for LGBMRanker's `group` argument.

`group` is not a reference to a column index or anything, but a vector of group sizes. Your explanation in the suggested changes are spot-on.

At the end of the day, LightGBM has an internal check that looks something like `assert X.shape[0] == np.sum(group)`. Just giving a heads up that the DaskLGBMRanker's docs would now be different from LGBMRanker's docs.

tests/python_package_test/test_dask.py
ffineis: Yeah definitely good idea. A lot of the state of my test_dask.py reflects me dumbing down the tests trying to get them to pass the "Binding port" bug.

tests/python_package_test/test_dask.py
jameslamb: ok awesome, thanks! In general we don't put a ton of emphasis on docstrings for functions in tests, but this one has a lot of parameters so I thought it would be good, to be sure future contributors understand how to use it

tests/python_package_test/test_dask.py
jameslamb: ok cool, thank you

python-package/lightgbm/dask.py
jameslamb: no prob! I have a preference for smaller PRs that we can review quickly, since I'm also actively working on the Dask module right now. Reduces the risk that we cause merge conflicts for each other, yenno

python-package/lightgbm/dask.py
jameslamb: > Just giving a heads up that the DaskLGBMRanker's docs would now be different from LGBMRanker's docs.

Ah, thanks for that! I think I'll go make a separate PR for the non-Dask places this is documented.

jameslamb: I added https://github.com/microsoft/LightGBM/pull/3772 to improve the docs on `group` in the Python and R packages.

tests/python_package_test/test_dask.py
```diff
@@ -180,7 +283,49 @@ def test_regressor_local_predict(client, listen_port):
...
-    assert_eq(s1, s2)
+    assert np.isclose(s1, s2, rtol=1e-4)
+
+
+@pytest.mark.parametrize('output', ['array', 'dataframe'])
+def test_ranker(output, client, listen_port):
+    X, y, w, g, dX, dy, dw, dg = _create_ranking_data(output=output)
+
+    # Avoid lightgbm.basic.LightGBMError: Binding port 13xxx failed exceptions.
+    time.sleep(10)
```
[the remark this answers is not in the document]
ffineis: Addressed in d6e5209

tests/python_package_test/test_dask.py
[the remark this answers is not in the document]
ffineis: Addressed in d6e5209, although I happened upon cases where the difference was as high as 0.015.... Upped to 0.03.

tests/python_package_test/test_dask.py
ffineis: Addressed in d6e5209

tests/python_package_test/test_dask.py
ffineis: Addressed in 78358e2

python-package/lightgbm/dask.py
ffineis: Addressed in d6e5209

python-package/lightgbm/dask.py
ffineis: Slick! Addressed in d6e5209

python-package/lightgbm/dask.py
ffineis: Reverted to tuples in d6e5209

tests/python_package_test/test_dask.py
ffineis: Addressed in d6e5209

tests/python_package_test/test_dask.py
ffineis: Addressed in d6e5209

ffineis:
Hmmm ok so latest test issues (only on Azure pipeline):
- Getting random teardown Exceptions (note: Exceptions, not test failures) caused by timeouts in the `client` fixture. These are random re: regressor/classifier/ranker.
- On GPU builds, getting generic ["Invalid value" exceptions](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=8563&view=logs&j=6ee41f30-52b9-5fb2-cd2b-d0121164dc51&t=feed3427-4866-55c6-1937-183bc0e12687&l=2255) intermixed with ["Socket recv error, code: 104
" exceptions](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=8563&view=logs&j=6ee41f30-52b9-5fb2-cd2b-d0121164dc51&t=feed3427-4866-55c6-1937-183bc0e12687&l=2375) *only for ranking* task tests. I'm going to try out `lightgbm-dask-testing` a little bit to see if I can reproduce, but probably unlikely as this seems limited to GPU builds.

tests/python_package_test/test_dask.py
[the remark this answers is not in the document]
jameslamb: ok makes sense, no problem. I think 0.0.3 is ok.

jameslamb:
ok I can also try to look into those errors

ffineis:
Hey so updates on this. I've been able to find is this - https://github.com/dask/dask-ml/issues/611 - looks like the same exception triggered by a timeout during `do_disconnect`.

I'm thinking that we just install a wrapper that uses `pytest.skip` if the TimeoutError triggered from do_disconnect, basically like this:
```
import os

import functools
import pytest

import dask.array as da
import numpy as np

from sklearn.linear_model import Lasso

num_tests = 5
alpha_vec = np.linspace(1e-5, stop=2, num=num_tests)

def handle_async_timeout_exceptions(f):
    @functools.wraps(f)
    def wrapper(client, **kwargs):
        try:
            return f(client, **kwargs)
        except ValueError as e:
            print(str(e))
            if 'my fake exception' in str(e):
                msg = 'Ignoring do_disconnect error in fixture teardown: ' + str(e) + ' kwargs: ' + str(kwargs)
                print(msg)
                pytest.skip(msg)
            else:
                raise(e)
    return wrapper

def bootstrap_lasso(X, y, alpha):
    n = X.shape[0]
    boot_idx = np.random.choice(range(n), size=n, replace=True)
    oob_idx = np.array(np.setdiff1d(range(n), boot_idx))

    reg = Lasso(alpha=alpha)
    reg.fit(X[boot_idx, :], y[boot_idx])

    return reg.score(X[oob_idx, :], y[oob_idx])


@pytest.mark.parametrize('alpha', alpha_vec)
@handle_async_timeout_exceptions
def test_lasso(client, alpha):
    if alpha >= 1:
        raise ValueError('my fake exception')

    X, y = make_regression(n_samples=int(1e5), n_features=10, random_state=123)
    r2_futures = client.map(lambda i: bootstrap_lasso(X=X, y=y, alpha=alpha), range(10))
    r2_values = client.gather(r2_futures)

    print('alpha = {:.6f} -- avg(r2) = {:.2f}'.format(alpha, np.mean(r2_values)))

    client.close() # maybe will this help avoid errors in teardown...?
```
FWIW I ran this with num_tests = 100 and couldn't reproduce the teardown exception.

jameslamb:
> Hey so updates on this. I've been able to find is this - [dask/dask-ml#611](https://github.com/dask/dask-ml/issues/611) - looks like the same exception triggered by a timeout during `do_disconnect`.
> 
> I'm thinking that we just install a wrapper that uses `pytest.skip` if the TimeoutError triggered from do_disconnect, basically like this:
> FWIW I ran this with num_tests = 100 and couldn't reproduce the teardown exception.

Thanks for the investigation @ffineis . Since this is only an error in fixture teardown, I'm good with your proposal. I want to go as fast as possible on Dask changes between now and LightGBM 3.2.0, so if this helps get this PR through I think it's ok and we can try to fix the upstream issue later as a maintenance task. I'm convinced by what you wrote and my own tests that this wouldn't cover up a legitimate problem with `lightgbm` itself.

Can you just add a comment or docstring (whichever you want) on that fixture, linking to the dask-ml issue?

tests/python_package_test/test_dask.py
```diff
@@ -200,8 +341,46 @@ def test_regressor_local_predict(client, listen_port):
...
     assert_eq(s1, s2)
+    assert_eq(p1, p2)
+
+
+@pytest.mark.parametrize('output', ['array', 'dataframe'])
+@pytest.mark.parametrize('group', [None, group_sizes])
+def test_ranker(output, client, listen_port, group):
+    X, y, w, g, dX, dy, dw, dg = _create_ranking_data(output=output, group=group)
+
+    dask_ranker = dlgbm.DaskLGBMRanker(time_out=5, local_listen_port=listen_port, seed=42, min_child_samples=1)
```
jameslamb, line 353: I just made some changes in https://github.com/microsoft/LightGBM/pull/3786 to speed up the tests in this module. If it'll work for the tests you're adding in this PR, can you please make every `DaskLGBMRanker` and `LGBMRanker` use `n_iterations=10, num_leaves=10`? I found that training smaller models was able to make the tests faster without sacrificing any coverage.

jameslamb (changes requested):

ffineis:
Yep for sure, good call! Getting to work this evening.

tests/python_package_test/test_dask.py
jameslamb: hey @ffineis is it cool if I pick up the "remove testing dependency on `dask-ml`" thing?

If you want to make the contribution I'll totally respect that, but if it's just extra work you'd rather not do, I'm happy to make a PR right now for it. Wanna be respectful of your time.

ffineis:
Hey so updates - sorry this has been so drawn out. Cannot for the life of me figure out why the ranker-dask tests are struggling so much in comparison to reg/clf tests.

1. I'm going to eliminate GPU tests for `test_ranker*` tests. Tests keep failing on GPU tasks with ominous LightGBMExceptions, and I can't silence them with the try/catch decorator. I think it deserves additional investigation, figuring out what the issue is with distributed ranker training on GPUs.
2. When `num_leaves` and `num_iterations` are small (10 vs 31), I believe the effects of `parallel` learning are creeping into create discrepancies between dask/local models. Once in a while the discrepancy between spearman correlation from dask model vs local model jumps to 0.3. See [here](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=8601&view=logs&j=e3619e06-e539-5956-1ae7-e8c2c29d8e91&t=4b3d6e5d-f8c3-577f-235f-b68c5ed4d3ab&l=1528).

I dug into the cases when this large discrepancy happened. I think it's related to [this](https://github.com/microsoft/LightGBM/issues/3292#issuecomment-671288210) issue - when the data split among workers has different distributions (for small data), you can learn pretty different splits between the parallelized and serial boosters. Basically, the dask model is still legit (its features are focused in features 0, 1, 2, 3, 4 (the "important" features), and the predictions between the dask-trained booster and the local-trained booster on a dummy test set are highly rank-correlated. Screenshot:

Screen Shot 2021-01-19 at 4 08 55 PM

A couple of ways I'm going to resolve this:

- Make even dumber fake ranking data. The "important" features will change monotonically with `y`. The model really should not mess this up. Imagine data the looks like this: `(y, x, qid)`: [(1, b + c*1, 0), (0, b, 0), (2, b + c *2, 0), (1, b + c*1, 1), (2, b + c*2, 1), (0, b, 1), ...]. 
- Use slightly more overfitting models, `num_estimators = 50` and `num_leaves=20`, for ranking tests. This helps the rankers overfit so much that the dask model and local model reach parity (I think).

jameslamb:
Thanks for such a thorough investigation @ffineis ! That all sounds totally fine to me.

The smaller `n_iterations` and `num_leaves` was a tiny optimization to save CI time. If it's causing issues with checking correctness, it's totally not worth it. I don't want you to do a ton of work just to make it work with those smaller settings just to save a few seconds of test time.

For the GPU builds...I think it's ok to skip them for LGBMRanker. You can rely on `os.environ["TASK"] == "gpu"` to detect it (https://github.com/microsoft/LightGBM/blob/master/.vsts-ci.yml#L47). Those tests aren't even checking that `lightgbm.dask` works with GPU-accelerated training. It's just doing regular CPU training using a version of LightGBM that happens to have been compiled with GPU support. So we don't lose anything by skipping the ranker tests there, and can figure out the issues in a later PR.

By the way, I have a feature request up to legit support GPU-based training in the Dask interface, like allowing you to train on data in `dask-cudf` dataframes (#3776 ). But until that gets picked up, we're not making any guarantees about the Dask interface working with GPU training.

jameslamb:
ayyy those changes seem to have worked! Nice idea closing down the cluster.

Seeing that that works, maybe in the future I'll just cut every test over to `@gen_cluster`. I just learned about https://distributed.dask.org/en/latest/develop.html#writing-tests tonight, and in there they describe `@gen_cluster` like this

> easier to debug and cause the fewest problems with shutdowns.

But now that this seems to be working ok, don't touch it in this PR 😆 

I'll give this another review right now. Thanks for working through it!

python-package/lightgbm/dask.py
```diff
@@ -166,20 +175,32 @@ def _train(client, data, label, params, model_factory, weight=None, **kwargs):
...
     sample_weight : array-like of shape = [n_samples] or None, optional (default=None)
-            Weights of training data.
+        Weights of training data.
+    group : array-like where sum(group) = [n_samples] or None for non-ranking objectives (default=None)
+        Group/query data, only used for ranking task. sum(group) = n_samples. For example,
+        if you have a 100-record dataset with `group = [10, 20, 40, 10, 10]`, that means that you have
+        5 groups, where the first 10 records are in the first group, records 11-30 are the second group, etc.
     """
-    # Split arrays/dataframes into parts. Arrange parts into tuples to enforce co-locality
+    # Split arrays/dataframes into parts. Arrange parts into dicts to enforce co-locality
```
jameslamb, line 186:
```suggestion
    # Split arrays/dataframes into parts. Arrange parts into tuples to enforce co-locality
```

This should be changed back since you pull the `dicts` change back out.

python-package/lightgbm/dask.py
```diff
@@ -166,20 +175,32 @@ def _train(client, data, label, params, model_factory, weight=None, **kwargs):
...
     data_parts = _split_to_parts(data, is_matrix=True)
     label_parts = _split_to_parts(label, is_matrix=False)
-    if weight is None:
-        parts = list(map(delayed, zip(data_parts, label_parts)))
+    weight_parts = _split_to_parts(sample_weight, is_matrix=False) if sample_weight is not None else None
+    group_parts = _split_to_parts(group, is_matrix=False) if group is not None else None
+
+    # choose between four options of (sample_weight, group) being (un)specified
+    if weight_parts is None and group_parts is None:
```
jameslamb, line 193: I'd usually say this is overcomplicated and we're gonna make mistakes like this, but I'm ok with it because

1. we'll have a PR soon after this to replace it with dicts that have keys like `weight`, `group`, etc. (https://github.com/microsoft/LightGBM/pull/3708#discussion_r558569333)
2. you didn't have to change anything in the regression and classification tests and they're passing

So no change required, in my opinion. Just commenting for the benefit of other reviewers.

tests/python_package_test/test_dask.py
```diff
@@ -37,6 +46,134 @@ def listen_port():
...
+    group : array-like, optional (default=None)
+        1-d array or list of group sizes. When `group` is specified, this overrides n_samples, random_gs, and
+        avg_gs by simply creating groups with sizes group[0], ..., group[-1].
+    gmax : int (default=2)
+        Maximum graded relevance value for creating relevance/target vector. If you set this to 2, for example, all
+        documents in a group will have relevance scores of either 0, 1, or 2.
+    random_gs : bool (default=False)
+        True will make group sizes ~ Poisson(avg_gs), False will make group sizes == avg_gs.
+    avg_gs : int (default=10)
+        Average number of documents (records) in each group
```
jameslamb, line 74: these docs are awesome, thank you!

tests/python_package_test/test_dask.py
```diff
@@ -118,6 +257,27 @@ def test_training_does_not_fail_on_port_conflicts(client):
             )
             assert dask_classifier.booster_
 
+    client.close()
+
+
+@pytest.mark.parametrize('output', data_output)
+@pytest.mark.parametrize('centers', data_centers)
+def test_classifier_proba(output, centers, client, listen_port):
```
jameslamb, line 265: I think this test might have come from a mistake resolving a merge conflict. I merged this into the `test_classifier` test in #3786 . Could you please remove it?

tests/python_package_test/test_dask.py
```diff
@@ -220,8 +386,62 @@ def test_regressor_local_predict(client, listen_port):
     s2 = dask_regressor.to_local().score(X, y)
 
     # Predictions and scores should be the same
-    assert_eq(p1, p2)
     assert_eq(s1, s2)
+    assert_eq(p1, p2)
```
jameslamb, line 390: can you please move this line back up above `assert_eq(s1, s2)`, so it drops out of the diff?

jameslamb (changes requested):
This is looking great, thank you! I really really like the functions you've added for creating a ranking dataset in tests.

I added a few minor comments. After you address those, I'd like to request another review from @StrikerRUS for general Python things.

For the other TODOs that got pulled out of this PR, I've created the following issues:

* use dicts for `parts`: https://github.com/microsoft/LightGBM/issues/3795 
* factor out testing dependency on `dask-ml`: https://github.com/microsoft/LightGBM/issues/3796

ffineis:
> ayyy those changes seem to have worked! Nice idea closing down the cluster.
> 
> Seeing that that works, maybe in the future I'll just cut every test over to `@gen_cluster`. I just learned about https://distributed.dask.org/en/latest/develop.html#writing-tests tonight, and in there they describe `@gen_cluster` like this
> 
> > easier to debug and cause the fewest problems with shutdowns.
> 
> But now that this seems to be working ok, don't touch it in this PR 😆
> 
> I'll give this another review right now. Thanks for working through it!

Haha thx, man I was ready to just stop using the `distributed` fixtures and use a localcluster setup/teardown. This was a guess but glad it worked. Yep absolutely will address the outstanding issues.

python-package/lightgbm/dask.py
ffineis: Cool, yeah this should be changed to dicts in follow-up.

tests/python_package_test/test_dask.py
ffineis: oops my bad!

StrikerRUS:
> After you address those, I'd like to request another review from @StrikerRUS for general Python things.

Sure! I'd be happy to do this.

tests/python_package_test/test_dask.py
ffineis: Addressed in d1d4abf

tests/python_package_test/test_dask.py
ffineis: Addressed in d1d4abf

python-package/lightgbm/dask.py
ffineis: Addressed in d1d4abf

ffineis:
> Thanks for such a thorough investigation @ffineis ! That all sounds totally fine to me.
> 
> The smaller `n_iterations` and `num_leaves` was a tiny optimization to save CI time. If it's causing issues with checking correctness, it's totally not worth it. I don't want you to do a ton of work just to make it work with those smaller settings just to save a few seconds of test time.
> 
> For the GPU builds...I think it's ok to skip them for LGBMRanker. You can rely on `os.environ["TASK"] == "gpu"` to detect it (https://github.com/microsoft/LightGBM/blob/master/.vsts-ci.yml#L47). Those tests aren't even checking that `lightgbm.dask` works with GPU-accelerated training. It's just doing regular CPU training using a version of LightGBM that happens to have been compiled with GPU support. So we don't lose anything by skipping the ranker tests there, and can figure out the issues in a later PR.
> 
> By the way, I have a feature request up to legit support GPU-based training in the Dask interface, like allowing you to train on data in `dask-cudf` dataframes (#3776 ). But until that gets picked up, we're not making any guarantees about the Dask interface working with GPU training.

Hey sorry never got back to you on this - yeah totally agree that special dask-GPU support is the way to go! Just a heads up - I got the teardown error on a DaskLGBMRegressor task during a GPU test (logs [here](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=8625&view=logs&j=582743ec-83fe-58c4-937e-4197b5816801&t=63b016c8-8965-547a-fab6-395af921c96a&l=535)) so I added the pytest.skip for all GPU tasks in 
56bb2c9.

tests/python_package_test/test_dask.py
jameslamb: thanks!

ffineis:
Hey can we merge? Just want to unblock the several PR's I feel are waiting in the wings, haha.

jameslamb (approved):
Thanks for all the hard work! From my perspective, this is ready to be merged.

However, I'd still like @StrikerRUS to do one more review for general Python things or similarity of patterns between here and the non-Dask `sklearn` interface.

tests/python_package_test/test_dask.py
```diff
@@ -37,6 +47,134 @@ def listen_port():
...
+
+    Returns
+    ----------
+    X : 2-d np.ndarray of shape = [n_samples (or np.sum(group), n_features]
+        Input feature matrix for ranking objective
+    y : 1-d np.array of shape = [n_samples (or np.sum(group))]
+        integer-graded relevance scores
+    group_ids: 1-d np.array of shape = [n_samples (or np.sum(group))]
+        vector of group ids, each value indicates to which group each record belongs
+    """
```
StrikerRUS, line 85:
```suggestion
    """Generate a learning-to-rank dataset - feature vectors grouped together with
    integer-valued graded relevance scores. Replace this with a sklearn.datasets function
    if ranking objective becomes supported in sklearn.datasets module.

    Parameters
    ----------
    n_samples : int, optional (default=100)
        Total number of documents (records) in the dataset.
    n_features : int, optional (default=20)
        Total number of features in the dataset.
    n_informative : int, optional (default=5)
        Number of features that are "informative" for ranking, as they are bias + beta * unif(min=y, max=y + 1),
        where bias and beta are standard normal variates. If this is greater than n_features, the dataset will have
        n_features features, all will be informative.
    group : array-like, optional (default=None)
        1-d array or list of group sizes. When `group` is specified, this overrides n_samples, random_gs, and
        avg_gs by simply creating groups with sizes group[0], ..., group[-1].
    gmax : int, optional (default=2)
        Maximum graded relevance value for creating relevance/target vector. If you set this to 2, for example, all
        documents in a group will have relevance scores of either 0, 1, or 2.
    random_gs : bool, optional (default=False)
        True will make group sizes ~ Poisson(avg_gs), False will make group sizes == avg_gs.
    avg_gs : int, optional (default=10)
        Average number of documents (records) in each group.

    Returns
    -------
    X : 2-d np.ndarray of shape = [n_samples (or np.sum(group)), n_features]
        Input feature matrix for ranking objective.
    y : 1-d np.array of shape = [n_samples (or np.sum(group))]
        Integer-graded relevance scores.
    group_ids: 1-d np.array of shape = [n_samples (or np.sum(group))]
        Array of group ids, each value indicates to which group each record belongs.
    """
```

tests/python_package_test/test_dask.py
```diff
@@ -37,6 +47,134 @@ def listen_port():
...
+
+            # groups should contain > 1 element for pairwise learning objective.
+            if gsize < 1:
+                continue
+
+            y_vec = np.append(y_vec, rnd_generator.choice(relvalues, size=gsize, replace=True))
+            group_id_vec = np.append(group_id_vec, [gid] * gsize)
+            gid += 1
+
+        y_vec, group_id_vec = y_vec[0:n_samples], group_id_vec[0:n_samples]
```
StrikerRUS, line 115:
```suggestion
        y_vec, group_id_vec = y_vec[:n_samples], group_id_vec[:n_samples]
```

tests/python_package_test/test_dask.py
```diff
@@ -37,6 +47,134 @@ def listen_port():
...
+        X[:, j] = bias + coef * y_vec
+
+    return X, y_vec, group_id_vec
+
+
+def _create_ranking_data(n_samples=100, output='array', chunk_size=50, **kwargs):
+    X, y, g = _make_ranking(n_samples=n_samples, random_state=42, **kwargs)
+    rnd = np.random.RandomState(42)
+    w = rnd.rand(X.shape[0]) * 0.01
+    g_rle = np.array([sum([1 for _ in grp]) for _, grp in itertools.groupby(g)])
```
StrikerRUS, line 132:
```suggestion
    g_rle = np.array([len(grp) for _, grp in itertools.groupby(g)])
```

tests/python_package_test/test_dask.py
```diff
@@ -37,6 +47,134 @@ def listen_port():
...
+
+    if output == 'dataframe':
+
+        # add target, weight, and group to DataFrame so that partitions abide by group boundaries.
+        X_df = pd.DataFrame(X, columns=[f'feature_{i}' for i in range(X.shape[1])])
+        X = X_df.copy()
+        X_df = X_df.assign(y=y, g=g, w=w)
+
+        # set_index ensures partitions are based on group id. See https://bit.ly/3pAWyNw.
```
StrikerRUS, line 141: I don't like URL shorteners. For example, links might expire...
```suggestion
        # set_index ensures partitions are based on group id.
        # See https://stackoverflow.com/questions/49532824/dask-dataframe-split-partitions-based-on-a-column-or-function.
```

tests/python_package_test/test_dask.py
```diff
@@ -37,6 +47,134 @@ def listen_port():
...
+
+        # encode group identifiers into run-length encoding, the format LightGBMRanker is expecting
+        # so that within each partition, sum(g) = n_samples.
+        dg = dg.map_partitions(lambda p: p.groupby('g', sort=False).apply(lambda z: z.shape[0]))
+
+    elif output == 'array':
+
+        # ranking arrays: one chunk per group. Each chunk must include all columns.
+        p = X.shape[1]
+        dX, dy, dw, dg = list(), list(), list(), list()
```
StrikerRUS, line 159:
```suggestion
        dX, dy, dw, dg = [], [], [], []
```

tests/python_package_test/test_dask.py
```diff
@@ -37,6 +47,134 @@ def listen_port():
...
+            dw.append(da.from_array(w[lhs:rhs]))
+            dg.append(da.from_array(np.array([g_rle[g_idx]])))
+
+        dX = da.concatenate(dX, axis=0)
+        dy = da.concatenate(dy, axis=0)
+        dw = da.concatenate(dw, axis=0)
+        dg = da.concatenate(dg, axis=0)
+
+    else:
+        raise ValueError('ranking data creation only supported for Dask arrays and dataframes')
```
StrikerRUS, line 173:
```suggestion
        raise ValueError('Ranking data creation only supported for Dask arrays and dataframes')
```

tests/python_package_test/test_dask.py
```diff
@@ -96,6 +234,8 @@ def test_classifier(output, centers, client, listen_port):
     assert_eq(y, p2)
     assert_eq(p1_proba, p2_proba, atol=0.3)
 
+    client.close()
```
StrikerRUS, line 237: @jameslamb Are you agree with this unrelated to Ranker itself change?

tests/python_package_test/test_dask.py
```diff
@@ -226,6 +374,54 @@ def test_regressor_local_predict(client, listen_port):
...
+
+
+@pytest.mark.parametrize('output', ['array', 'dataframe'])
+@pytest.mark.parametrize('group', [None, group_sizes])
+def test_ranker(output, client, listen_port, group):
+
+    X, y, w, g, dX, dy, dw, dg = _create_ranking_data(output=output, group=group)
+
+    # -- use many trees + leaves to overfit, help ensure that dask data-parallel strategy matches that of
+    # -- serial learner. See https://github.com/microsoft/LightGBM/issues/3292#issuecomment-671288210.
```
StrikerRUS, line 387:
```suggestion
    # use many trees + leaves to overfit, help ensure that dask data-parallel strategy matches that of
    # serial learner. See https://github.com/microsoft/LightGBM/issues/3292#issuecomment-671288210.
```

tests/python_package_test/test_dask.py
```diff
@@ -226,6 +374,54 @@ def test_regressor_local_predict(client, listen_port):
...
+
+    dask_ranker = dlgbm.DaskLGBMRanker(time_out=5, local_listen_port=listen_port, tree_learner='data',
+                                       n_estimators=10, num_leaves=10, seed=42, min_child_samples=1)
+    dask_ranker = dask_ranker.fit(dX, dy, group=dg, client=client)
+    rnkvec_dask = dask_ranker.predict(dX)
+    rnkvec_dask = rnkvec_dask.compute()
+    rnkvec_local = dask_ranker.to_local().predict(X)
+
+    # distributed and to-local scores should be the same.
+    assert_eq(rnkvec_dask, rnkvec_local)
```
StrikerRUS, line 421: @jameslamb I like this test very much! Can we add the same for other estimators? Seems that Classifier tests do not asserts real Dask `predict()` equals to `.to_local().predict()`.

StrikerRUS (commented):
@ffineis Thank you very much for awesome contribution! Below are just my minor comments, please check.

python-package/lightgbm/dask.py
```diff
@@ -380,3 +401,26 @@ def to_local(self):
         model : lightgbm.LGBMRegressor
         """
         return self._to_local(LGBMRegressor)
+
+
+class DaskLGBMRanker(_LGBMModel, LGBMRanker):
+    """Docstring is inherited from the lightgbm.LGBMRanker."""
+
+    def fit(self, X, y=None, sample_weight=None, group=None, client=None, **kwargs):
```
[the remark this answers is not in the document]
jameslamb: we should have a feature request. I'll write it up and add a link here.

@ffineis, could you add `init_score=None` here between `sample_weight` and `group`, so the order matches the sklearn interface for `LGBMRanker`? (https://github.com/microsoft/LightGBM/blob/1c60cfa6a6dc8f24cfc2f37b0555cf1252366370/python-package/lightgbm/sklearn.py#L962). That way, if people have existing sklearn code with positional arguments to `fit()`, they won't accidentally have their `init_score` interpreted as `group`.

And can you just then add a check like this?

```python
if init_score is not None:
    raise RuntimeError("init_score is not currently supported in lightgbm.dask")
```

@StrikerRUS speaking of positional arguments, I'll open another issue where we can discuss how to handle the `client` argument. But let's leave that out of this PR.

tests/python_package_test/test_dask.py
```diff
@@ -1,30 +1,40 @@
 # coding: utf-8
+"""Tests for lightgbm.dask module
+
+An easy way to run these tests is from the (python) docker container.
+Also see lightgbm-dask-testing repo: https://github.com/jameslamb/lightgbm-dask-testing
```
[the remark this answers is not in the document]
jameslamb: I was torn about these. Since it seems like there will be another round of commits anyway, @ffineis can you please remove them?

tests/python_package_test/test_dask.py
jameslamb: oh thank you, missed this

tests/python_package_test/test_dask.py
jameslamb: Yes, I'm ok with this. I'm not 100% sure that it is unrelated, even though it's touching non-ranker tests. Since this PR is close to being merged and since there aren't other PRs waiting on this fix (the timeout / shutdown issues in Dask tests have not been as severe elsewhere), I'm ok leaving it.

tests/python_package_test/test_dask.py
jameslamb: the classifier tests do check that

* comparison between Dask `predict()` and non-Dask `predict()`: https://github.com/microsoft/LightGBM/blob/1c60cfa6a6dc8f24cfc2f37b0555cf1252366370/tests/python_package_test/test_dask.py#L94
* comparison between Dask `predict()` and `.to_local().predict()`: https://github.com/microsoft/LightGBM/blob/1c60cfa6a6dc8f24cfc2f37b0555cf1252366370/tests/python_package_test/test_dask.py#L138

I'd be open to a PR that makes it more explicit and cuts out the total number of tests by combining `test_classifier_local_predict` (https://github.com/microsoft/LightGBM/blob/1c60cfa6a6dc8f24cfc2f37b0555cf1252366370/tests/python_package_test/test_dask.py#L122) into `test_classifier` (https://github.com/microsoft/LightGBM/blob/1c60cfa6a6dc8f24cfc2f37b0555cf1252366370/tests/python_package_test/test_dask.py#L72) though!

Not an action we need on this PR, I'll write it up as a "good first issue".

python-package/lightgbm/dask.py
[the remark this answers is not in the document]
StrikerRUS: @jameslamb 
Yes, sure! Agree with all your intents.

tests/python_package_test/test_dask.py
StrikerRUS: @jameslamb 
> comparison between Dask `predict()` and `to_local().predict()`:

Actually this checks `to_local().predict(dX)` and **local** `predict()`. It is not the same as `to_local().predict(dX)` and **Dask** `predict()`, given that asserts you pointing are placed in two different (in theory unrelated) tests, I think.

python-package/lightgbm/dask.py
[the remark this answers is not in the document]
jameslamb: init_score: https://github.com/microsoft/LightGBM/issues/3807

client placement: https://github.com/microsoft/LightGBM/issues/3808

tests/python_package_test/test_dask.py
jameslamb: it's confusing because it's two tests, so I think we should merge the tests in the future like I said above (not anything you need to do on here, @ffineis).

I'm saying that taken together, they show *transitively* that `dask_model.predict() == dask_model.to_local().predict()`, even though neither one does that direct comparison.

* (line 92) `dask_model.predict() == sklearn_model.predict()`
* (line 138) `dask_model.to_local().predict() == sklearn_model.predict()`

But that's only partially true, since `test_local_predict` only does binary classification, and only on array data. So combining them would give more coverage and would show whether that result holds up for multi-class classification or for other input data types.

StrikerRUS: > it's confusing because it's two tests, so I think we should merge the tests in the future like I said above

Totally agree!

> I'm saying that taken together, they show *transitively* that `dask_model.predict() == dask_model.to_local().predict()`

Also agree. But two different tests can take different inputs (params, data), can be modified in future separately, etc. So it can be not fully fair comparison. Merging them into one test will solve all issues. 🙂

python-package/lightgbm/dask.py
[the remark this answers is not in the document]
ffineis: Yep, no prob

tests/python_package_test/test_dask.py
ffineis: 100%

ffineis: Hey heads up on this - I changed the `label` values from y[i] = unif(label[i], label[i] + 1) to make the test data simpler to get more reliable models on unit tests. So I'm changing ```Number of features that are "informative" for ranking, as they are bias + beta * unif(min=y, max=y + 1),```

To

```Number of features that are "informative" for ranking, as they are bias + beta * y```

File: tests/python_package_test/test_dask.py (hunk 19 above)
Comment (ffineis) [reply tests/python_package_test/test_dask.py:132]: ```TypeError: object of type 'itertools._grouper' has no len()``
Thing we have to stick with the ugly list comprehension unfortunately. Want me to find an alternative?

**Review by StrikerRUS — APPROVED** [review]
I lost one my suggestion during previous review about `issubclass` and docstring about `group` param still doesn't match the same docstring in other places.

Everything else looks good to me. Thanks!

I'm approving this PR to not delay it anymore and let @jameslamb merge it finally.
You can commit my current suggestions right from the web interface. Just click `Commit suggestion` button if you agree with suggestion. No need to waste time re-typing them locally and then pushing with `git`.

File: python-package/lightgbm/dask.py [hunk 28]
```diff
@@ -167,22 +176,35 @@ def _train(client, data, label, params, model_factory, weight=None, **kwargs):
...
-    model_factory : lightgbm.LGBMClassifier or lightgbm.LGBMRegressor class
+    model_factory : lightgbm.LGBMClassifier, lightgbm.LGBMRegressor, or lightgbm.LGBMRanker class
     sample_weight : array-like of shape = [n_samples] or None, optional (default=None)
-            Weights of training data.
+        Weights of training data.
+    group : array-like or None, optional (default=None)
+        Group/query data.
+        Only used in the learning-to-rank task.
+        sum(group) = n_samples.
+        For example, if you have a 100-document dataset with ``group = [10, 20, 40, 10, 10, 10]``, that means that you have 6 groups, where the first 10 records are in the first group, records 11-30 are in the second group, etc.
```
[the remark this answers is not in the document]
ffineis: Ah, weird! I was just going off of https://github.com/microsoft/LightGBM/pull/3772/files, but yeah you're totally right, the 31-70 thing is in master. Thx!

jameslamb:
All the tests are passing! 🎉 

I just took one more look at the diff, and all looks good. So I'm going to merge this. Thanks so much for this fantastic contribution @ffineis , and to @StrikerRUS for the help with thorough reviews.

ffineis:
Thanks @jameslamb and @StrikerRUS for bearing with me!! Glad we could make it work.

PR #3715, opened 2021-01-04
[python][tests] small Python tests cleanup

StrikerRUS:
- Rename temp model file to not overcomplicate `.gitignore`.
- Reduce code duplication in params definition.
- Use asserts from `unittest` module for consistency with the rest codebase.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -423,7 +423,7 @@ lightgbm.Rcheck/
 miktex*.zip
 *.def
 
-# Files created by R examples and tests
+# Files created by R and Python examples and tests
 **/lgb-Dataset.data
 **/lgb.Dataset.data
 **/model.txt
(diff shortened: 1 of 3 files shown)


jameslamb (approved):
I personally have a preference for `assert` with an arbitrary expression instead of the added indirection of the unittest builtins, but I wont block this PR

StrikerRUS:
@jameslamb 
This looks like a "good first issue" to rewrite **all** tests removing `unittest` dependency and referring [`test_dask.py`](https://github.com/microsoft/LightGBM/blob/master/tests/python_package_test/test_dask.py) and [`test_dual.py`](https://github.com/microsoft/LightGBM/blob/master/tests/python_package_test/test_dual.py) as examples. WDYT?

jameslamb:
I agree! I can write it up

StrikerRUS:
Thanks! Please do it. 🙂

jameslamb:
> Thanks! Please do it.

added #3732. Feel free to edit it if there's anything you'd like to change.

StrikerRUS:
Thank you!

PR #3723, opened 2021-01-04
[ci] remove flaky MinGW job for R

StrikerRUS:
TBH, I'm tired to rerun each PR multiple times to pass this flaky job and finally be able to merge PR. In addition, GitHub doesn't allow to rerun each job separately, you can rerun only all jobs. It takes a lot of time. So I'd suggest comment out flaky job till issue #3469 will be fixed.

image

### DIFF
--- a/.github/workflows/r_package.yml
+++ b/.github/workflows/r_package.yml
@@ -65,12 +65,6 @@ jobs:
             compiler: clang
             r_version: 4.0
             build_type: cmake
-#          - os: windows-latest
-#            task: r-package
-#            compiler: MINGW
-#            toolchain: MINGW
-#            r_version: 3.6
-#            build_type: cmake
           - os: windows-latest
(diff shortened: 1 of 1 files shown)


jameslamb (changes requested):
I've been thinking about this as well. Totally agree.

But I'd prefer to just remove it instead of commenting it out. We have git and this PR will be linked to #3469 based on your comment, so whenever we make another attempt at #3469 it won't be hard to find this PR and look at the diff.

I have a preference for not having any commented-out code, since it can lead to false positives when you're searching through the repo for things.

StrikerRUS:
@jameslamb Addressed in 93ce5c31a811f69f4a383d3e9226864f9a4a1051.

jameslamb (approved):

jameslamb:
thanks again!

PR #3726, opened 2021-01-04
[ci] Move check-docs and lint jobs off Travis

jameslamb:
Because of #3519 , we're in the process of removing all CI jobs that run on Travis and moving them to GitHub Actions.

This PR moves `check-docs` and `lint` to GitHub Actions. I wanted to do this one by itself to make #3672 smaller. These two jobs are also the easiest to move.

This PR proposes putting `check-docs` and `lint` into a separate GitHub Actions workflow called "Static Analysis". Today, GitHub Actions does not allow you to re-run individual failed jobs. So if one job fails for random or temporary reasons, you have to re-run all jobs in the workflow. Since the `check-docs` tasks fails somewhat regularly due to temporary network problems, I think it is a good idea to keep it + `lint` separate, so that the cost of re-running it is low.

Thanks for reviewing all of these @StrikerRUS . I know it's a lot of PRs but hopefully the end result will be that our CI is more stable.

### DIFF
--- a/.ci/setup.sh
+++ b/.ci/setup.sh
@@ -48,7 +48,7 @@ else  # Linux
     fi
 fi
 
-if [[ "${TASK:0:9}" != "r-package" ]]; then
+if [[ "${TASK}" != "r-package" ]]; then
     if [[ $SETUP_CONDA != "false" ]]; then
         sh conda.sh -b -p $CONDA
     fi
--- a/.github/workflows/static_analysis.yml
+++ b/.github/workflows/static_analysis.yml
@@ -63,17 +63,18 @@ jobs:
       - name: Test documentation
         shell: bash
         run: |
-          Rscript --vanilla -e "roxygen2::roxygenize('R-package/', load = 'installed')" || exit -1
+          Rscript --vanilla -e "roxygen2::roxygenize('R-package/', load = 'installed')" || exit 100
           num_doc_files_changed=$(
               git diff --name-only | grep --count -E "\.Rd|NAMESPACE"
           )
           if [[ ${num_doc_files_changed} -gt 0 ]]; then
               echo "Some R documentation files have changed. Please re-generate them and commit those changes."
               echo ""
-              echo "    Rscript build_r.R"
+              echo "    sh build-cran-package.sh"
+              echo "    R CMD INSTALL --with-keep.source lightgbm_*.tar.gz"
               echo "    Rscript -e \"roxygen2::roxygenize('R-package/', load = 'installed')\""
               echo ""
-              exit -1
+              exit 101
           fi
   all-successful:
     # https://github.community/t/is-it-possible-to-require-all-github-actions-tasks-to-pass-without-enumerating-them/117957/4?u=graingert
--- a/.github/workflows/static_analysis.yml
+++ b/.github/workflows/static_analysis.yml
@@ -11,8 +11,10 @@ on:
     - master
 
 env:
+  COMPILER: 'gcc'
   CONDA_ENV: test-env
   GITHUB_ACTIONS: 'true'
+  OS_NAME: 'linux'
   PYTHON_VERSION: 3.8
 
 jobs:
@@ -36,8 +38,6 @@ jobs:
         shell: bash
         run: |
           export TASK="${{ matrix.task }}"
-          export OS_NAME="linux"
-          export COMPILER="gcc"
           export BUILD_DIRECTORY="$GITHUB_WORKSPACE"
           export CONDA=${HOME}/miniconda
           export PATH=${CONDA}/bin:$HOME/.local/bin:${PATH}
--- a/.github/workflows/static_analysis.yml
+++ b/.github/workflows/static_analysis.yml
@@ -61,7 +61,7 @@ jobs:
           sh build-cran-package.sh || exit -1
           R CMD INSTALL --with-keep.source lightgbm_*.tar.gz || exit -1
       - name: Test documentation
-        shell: bash --noprofile --norc -e {0}
+        shell: bash --noprofile --norc {0}
         run: |
           Rscript --vanilla -e "roxygen2::roxygenize('R-package/', load = 'installed')" || exit 100
           num_doc_files_changed=$(
--- a/.github/workflows/static_analysis.yml
+++ b/.github/workflows/static_analysis.yml
@@ -59,7 +59,7 @@ jobs:
         run: |
           Rscript --vanilla -e "install.packages(c('R6', 'data.table', 'jsonlite', 'roxygen2', 'testthat'), repos = 'https://cran.r-project.org')"
           sh build-cran-package.sh || exit -1
-          R CMD INSTALL lightgbm_*.tar.gz || exit -1
+          R CMD INSTALL --with-keep.source lightgbm_*.tar.gz || exit -1
       - name: Test documentation
         shell: bash
         run: |
--- a/.github/workflows/static_analysis.yml
+++ b/.github/workflows/static_analysis.yml
@@ -1,5 +1,5 @@
-# contains tests that can run without compiling LightGBM, like
-# linters and documentation checks.
+# contains non-functional tests, like checks on docs
+# and code style
 name: Static Analysis
 
 on:
@@ -45,7 +45,7 @@ jobs:
           $GITHUB_WORKSPACE/.ci/test.sh || exit -1
   r-check-docs:
     name: r-package-check-docs
-    timeout-minutes: 120
+    timeout-minutes: 60
     runs-on: ubuntu-latest
     container: rocker/verse
     steps:
--- a/.github/workflows/static_analysis.yml
+++ b/.github/workflows/static_analysis.yml
@@ -1,0 +1,52 @@
+# contains tests that can run without compiling LightGBM, like
+# linters and documentation checks.
+name: Static Analysis
+
+on:
+  push:
+    branches:
+    - master
+  pull_request:
+    branches:
+    - master
+
+env:
+  CONDA_ENV: test-env
+  GITHUB_ACTIONS: 'true'
+  PYTHON_VERSION: 3.8
+
+jobs:
+  test:
+    name: ${{ matrix.task }}
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - task: lint
+          - task: check-docs
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2.3.4
+        with:
+          fetch-depth: 5
+          submodules: true
+      - name: Setup and run tests
+        shell: bash
+        run: |
+          export TASK="${{ matrix.task }}"
+          export OS_NAME="linux"
+          export BUILD_DIRECTORY="$GITHUB_WORKSPACE"
+          export LGB_VER=$(head -n 1 VERSION.txt)
+          export CONDA=${HOME}/miniconda
+          export PATH=${CONDA}/bin:$HOME/.local/bin:${PATH}
+          $GITHUB_WORKSPACE/.ci/setup.sh
+          $GITHUB_WORKSPACE/.ci/test.sh
+  all-successful:
+    # https://github.community/t/is-it-possible-to-require-all-github-actions-tasks-to-pass-without-enumerating-them/117957/4?u=graingert
+    runs-on: ubuntu-latest
+    needs: [test]
+    steps:
+    - name: Note that all tests succeeded
+      run: echo "🎉"
--- a/.github/workflows/static_analysis.yml
+++ b/.github/workflows/static_analysis.yml
@@ -57,11 +57,11 @@ jobs:
       - name: Install packages
         shell: bash
         run: |
-          Rscript --vanilla -e "install.packages(c('R6', 'data.table', 'jsonlite', 'roxygen2', 'testthat'), repos = 'https://cran.r-project.org')"
+          Rscript -e "install.packages(c('R6', 'data.table', 'jsonlite', 'roxygen2', 'testthat'), repos = 'https://cran.r-project.org')"
           sh build-cran-package.sh || exit -1
           R CMD INSTALL --with-keep.source lightgbm_*.tar.gz || exit -1
       - name: Test documentation
-        shell: bash
+        shell: bash --noprofile --norc -e {0}
         run: |
           Rscript --vanilla -e "roxygen2::roxygenize('R-package/', load = 'installed')" || exit 100
           num_doc_files_changed=$(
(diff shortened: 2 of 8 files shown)


.github/workflows/static_analysis.yml
```diff
@@ -0,0 +1,52 @@
+# contains tests that can run without compiling LightGBM, like
+# linters and documentation checks.
```
StrikerRUS, line 2: Maybe move R lint check here for better logical grouping? Does it require all our R stuff installation routine or we can isolate this check in Docker like ASAN/UBSAN?
https://github.com/microsoft/LightGBM/blob/fc1b930026a2a376d8d9174c2c7662b7185885b5/.github/workflows/r_package.yml#L44-L47

.github/workflows/static_analysis.yml
```diff
@@ -0,0 +1,52 @@
...
+      matrix:
+        include:
+          - task: lint
+          - task: check-docs
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2.3.4
+        with:
+          fetch-depth: 5
+          submodules: true
```
StrikerRUS, line 34: To speedup. I believe empty submodule folders will not break anything.
```suggestion
          submodules: false
```

.github/workflows/static_analysis.yml
```diff
@@ -0,0 +1,52 @@
...
+        shell: bash
+        run: |
+          export TASK="${{ matrix.task }}"
+          export OS_NAME="linux"
+          export BUILD_DIRECTORY="$GITHUB_WORKSPACE"
+          export LGB_VER=$(head -n 1 VERSION.txt)
+          export CONDA=${HOME}/miniconda
+          export PATH=${CONDA}/bin:$HOME/.local/bin:${PATH}
+          $GITHUB_WORKSPACE/.ci/setup.sh
+          $GITHUB_WORKSPACE/.ci/test.sh
```
StrikerRUS, line 45:
```suggestion
          $GITHUB_WORKSPACE/.ci/setup.sh || exit -1
          $GITHUB_WORKSPACE/.ci/test.sh || exit -1
```

.github/workflows/static_analysis.yml
```diff
@@ -0,0 +1,52 @@
...
+  all-successful:
```
StrikerRUS, line 46: Need to mark this job `Required` in repo settings.
cc @guolinke

.github/workflows/static_analysis.yml
```diff
@@ -0,0 +1,52 @@
+# contains tests that can run without compiling LightGBM, like
+# linters and documentation checks.
+name: Static Analysis
```
StrikerRUS, line 3: Maybe add a badge for this new workflow?
https://github.com/microsoft/LightGBM/blame/master/README.md#L6

StrikerRUS (commented):
Thanks for doing this! Please check some my initial comments below:

.github/workflows/static_analysis.yml
jameslamb: `r-package-check-docs` does require all of our stuff, unfortunately. The library has to be installed (https://github.com/microsoft/LightGBM/blob/master/.ci/test_r_package.sh#L109) before you can re-build the docs with `roxygen2::roxygenize()`

.github/workflows/static_analysis.yml
jameslamb: oooo yeah you're correct. I was thinking about the problem we had in https://github.com/microsoft/LightGBM/issues/3655, but that was only for the R package because the R package has to be installed for its site to be built.

.github/workflows/static_analysis.yml
StrikerRUS: OK, I see. But seems that other R jobs also require building but are run inside a Docker. For instance, 
https://github.com/microsoft/LightGBM/blob/98e6f719c5111a8bc631877c3d3d9c55e360ae5d/.github/workflows/r_package.yml#L173-L180
or 
https://github.com/microsoft/LightGBM/blob/98e6f719c5111a8bc631877c3d3d9c55e360ae5d/.github/workflows/r_valgrind.yml#L23-L25

I wonder whether we can do the same for docs? I think the only thing we lose is finer control of used R version and CRAN/CMake type of installation. But is it important for docs check?

jameslamb: oh true, that's a good point! Ok let me try that, I like that idea.

.github/workflows/static_analysis.yml
jameslamb: good idea, added!

.github/workflows/static_analysis.yml
jameslamb: ok yeah that worked! That job only takes 7 minutes, I'm very happy with that and very comfortable grouping that one with `check-docs` and `lint`.

.github/workflows/static_analysis.yml
jameslamb: it looks like this was done, thanks @guolinke !

image

jameslamb:
ok I think this is ready for another review

README.md
```diff
@@ -3,7 +3,8 @@
 Light Gradient Boosting Machine
 ===============================
 
-[![GitHub Actions Build Status](https://github.com/microsoft/LightGBM/workflows/R%20GitHub%20Actions/badge.svg?branch=master)](https://github.com/microsoft/LightGBM/actions)
+[![R GitHub Actions Build Status](https://github.com/microsoft/LightGBM/workflows/R%20GitHub%20Actions/badge.svg?branch=master)](https://github.com/microsoft/LightGBM/actions)
```
StrikerRUS, line 6: I think we can rename this workflow to simple `R-package`. It will be consistent with other our workflows.
https://github.com/microsoft/LightGBM/blob/415c0cb57794d184ea8639aa5a98a83610639976/.github/workflows/r_package.yml#L1
```suggestion
[![R GitHub Actions Build Status](https://github.com/microsoft/LightGBM/workflows/R-package/badge.svg?branch=master)](https://github.com/microsoft/LightGBM/actions)
```

.github/workflows/static_analysis.yml
```diff
@@ -0,0 +1,85 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v2.3.4
+        with:
+          fetch-depth: 5
+          submodules: false
+      - name: Setup and run tests
+        shell: bash
+        run: |
+          export TASK="${{ matrix.task }}"
+          export OS_NAME="linux"
```
StrikerRUS, line 39:
```suggestion
          export OS_NAME="linux"
          COMPILER="gcc"
```

.github/workflows/static_analysis.yml
```diff
@@ -0,0 +1,85 @@
...
+          export BUILD_DIRECTORY="$GITHUB_WORKSPACE"
+          export LGB_VER=$(head -n 1 VERSION.txt)
```
StrikerRUS, line 41:
```suggestion
```

StrikerRUS (commented):
Generally this R looks OK to me. Thanks! Please check one more round of review below:

.github/workflows/static_analysis.yml
```diff
@@ -0,0 +1,85 @@
...
+          export CONDA=${HOME}/miniconda
+          export PATH=${CONDA}/bin:$HOME/.local/bin:${PATH}
```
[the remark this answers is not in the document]
jameslamb: this is where stuff that is installed with `pip install --user` was installed. That applies to things using in linting, like `cpplint` and `rstcheck`

.ci/setup.sh
```diff
@@ -48,7 +48,7 @@ else  # Linux
     fi
 fi
 
-if [[ "${TASK:0:9}" != "r-package" ]]; then
```
StrikerRUS, line 51: Same change here
https://github.com/microsoft/LightGBM/blob/415c0cb57794d184ea8639aa5a98a83610639976/.ci/test.sh#L11

.github/workflows/static_analysis.yml
```diff
@@ -0,0 +1,85 @@
...
+          $GITHUB_WORKSPACE/.ci/setup.sh || exit -1
+          $GITHUB_WORKSPACE/.ci/test.sh || exit -1
+  r-check-docs:
+    name: r-package-check-docs
+    timeout-minutes: 120
```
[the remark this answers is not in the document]
jameslamb: ah, just copy-paste from the valgrind template. Will reduce it to 60

.github/workflows/static_analysis.yml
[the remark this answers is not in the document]
StrikerRUS: Hmm, OK. Thanks for the explanation!

.github/workflows/static_analysis.yml
```diff
@@ -0,0 +1,85 @@
...
+          fetch-depth: 5
+          submodules: true
+      - name: Install packages
+        shell: bash
+        run: |
+          Rscript -e "install.packages(c('R6', 'data.table', 'jsonlite', 'roxygen2', 'testthat'), repos = 'https://cran.r-project.org')"
+          sh build-cran-package.sh || exit -1
+          R CMD INSTALL --with-keep.source lightgbm_*.tar.gz || exit -1
+      - name: Test documentation
+        shell: bash --noprofile --norc {0}
```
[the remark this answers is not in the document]
jameslamb: when you just use `shell: bash`, GitHub Actions actually uses `bash --noprofile --norc -e -o pipefail {0}` (where `{0}`  is a placeholder for your code from the `run:` block). This causes this command to fail if no changes are found in the docs:

```shell
num_doc_files_changed=$(
    git diff --name-only | grep --count -E "\.Rd|NAMESPACE"
)
```

If `git diff`  doesn't match anything, the full command returns a non-zero exit code.

You can try this yourself. Run the following in any git repo with no changes (i.e. `git status` says "nothing to commit, working tree clean").

```shell
git diff --name-only | grep --count -E "abc"
echo $?
```

This isn't a problem in the existing r-package-check-docs task on `master` because we run it in a script that does not have `-e` or `-o pipefail` set.


Using `bash --noprofile --norc {0}` ensures that the job only fails when we want it to (by calling `exit` explicitly).

.github/workflows/static_analysis.yml
jameslamb: why is it necessary to set  a compiler for `check-docs` and `lint`?

.github/workflows/static_analysis.yml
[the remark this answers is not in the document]
StrikerRUS: Excellent explanation, thank you!

.github/workflows/static_analysis.yml
StrikerRUS: Although compiler settings are not used directly in `lint` and `check-docs`, a lot of pre-steps rely on `COMPILER` env variable. For example here 
https://github.com/microsoft/LightGBM/blob/415c0cb57794d184ea8639aa5a98a83610639976/.ci/test.sh#L3-L9
and the whole `setup.sh` script. I believe it will be better to explicitly specify compiler to not lose anything during pre-steps in the future that will rely on it.

jameslamb: ok sure, no problem

.github/workflows/static_analysis.yml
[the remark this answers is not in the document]
jameslamb:
```suggestion
    timeout-minutes: 60
```

README.md
jameslamb: sure

.ci/setup.sh
jameslamb: fixed in https://github.com/microsoft/LightGBM/pull/3726/commits/f886b142b272f099373f5cef7c03661c0c53a8a6

README.md
jameslamb: fixed in https://github.com/microsoft/LightGBM/pull/3726/commits/f886b142b272f099373f5cef7c03661c0c53a8a6 and https://github.com/microsoft/LightGBM/pull/3726/commits/6496d84c1a60d4c9ce0fefcfd19c0ffe0608e130

.github/workflows/static_analysis.yml
```diff
@@ -0,0 +1,85 @@
+# contains tests that can run without compiling LightGBM, like
+# linters and documentation checks.
```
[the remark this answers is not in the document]
jameslamb: fixed in https://github.com/microsoft/LightGBM/pull/3726/commits/6496d84c1a60d4c9ce0fefcfd19c0ffe0608e130. New description:

> contains non-functional tests, like checks on docs and code style

.github/workflows/static_analysis.yml
```diff
@@ -0,0 +1,85 @@
...
+        uses: actions/checkout@v2.3.4
+        with:
+          fetch-depth: 5
+          submodules: false
+      - name: Setup and run tests
+        shell: bash
+        run: |
+          export TASK="${{ matrix.task }}"
+          export OS_NAME="linux"
+          COMPILER="gcc"
```
jameslamb, line 40:
```suggestion
          export COMPILER="gcc"
```

README.md
```diff
@@ -3,7 +3,8 @@
 Light Gradient Boosting Machine
 ===============================
 
-[![GitHub Actions Build Status](https://github.com/microsoft/LightGBM/workflows/R%20GitHub%20Actions/badge.svg?branch=master)](https://github.com/microsoft/LightGBM/actions)
+[![R-package Build Status](https://github.com/microsoft/LightGBM/workflows/R-package/badge.svg?branch=master)](https://github.com/microsoft/LightGBM/actions)
```
StrikerRUS, line 6:
```suggestion
[![R-package GitHub Actions Build Status](https://github.com/microsoft/LightGBM/workflows/R-package/badge.svg?branch=master)](https://github.com/microsoft/LightGBM/actions)
```

.github/workflows/static_analysis.yml
```diff
@@ -0,0 +1,85 @@
...
+        uses: actions/checkout@v2.3.4
+        with:
+          fetch-depth: 5
+          submodules: false
+      - name: Setup and run tests
+        shell: bash
+        run: |
+          export TASK="${{ matrix.task }}"
+          export OS_NAME="linux"
+          export COMPILER="gcc"
```
StrikerRUS, line 40: I think we can move these variables to `env` section of this config file closer to `PYTHON_VERSION` and other constants.

StrikerRUS (approved):
Great! Thank you for reorganizing CI!
I even had a chance to test `rerun-all` button for new workflow - Wikipedia was not answering. Everything was OK.

Just two nits.

.github/workflows/static_analysis.yml
jameslamb: fixed in https://github.com/microsoft/LightGBM/pull/3726/commits/1af56f4c9048084ab746c802a7957f350d3465d0

PR #3728, opened 2021-01-04
[docs][R-package] change obsolete wordings in R README

StrikerRUS:
Sorry, noticed this inconsistency only after merging #3724.
Previous link led to the customization of installation with CMake.


jameslamb (approved):
ah yes you're totally right, thank you

PR #3738, opened 2021-01-07
[ci][docs] replace file extensions in docs during Sphinx build

StrikerRUS:
@jameslamb I remember you proposed this change some time ago. But I was against because was not bought of idea to rewrite all docs source files during the build process. Now I know that we can do it via internal Sphinx process which doesn't require rewriting files on a disk. Also, this approach safer than rewriting files because we are sure our regex is applied **only** to links in docs (`self.document.traverse(reference)`) not to the whole document.

Live demo: https://lightgbm.readthedocs.io/en/sphinx/

### DIFF
--- a/.ci/test.sh
+++ b/.ci/test.sh
@@ -29,8 +29,6 @@ if [[ $TASK == "check-docs" ]]; then
     rstcheck --report warning --ignore-directives=autoclass,autofunction,doxygenfile `find . -type f -name "*.rst"` || exit -1
     # build docs and check them for broken links
     make html || exit -1
-    find ./_build/html/ -type f -name '*.html' -exec \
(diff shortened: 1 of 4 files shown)


jameslamb (approved):
ooooo yeah this is great! I just tried clicking through https://lightgbm.readthedocs.io/en/sphinx/ and it seems like internal links are all doing what we'd expect (including in the R package).

StrikerRUS:
@jameslamb Thanks!

> all doing what we'd expect (including in the R package)

Another +1 for this way of replacing extensions because R-package docs will be unaffected as they are outside of the Sphinx build tree. 😉  (We simply pick up their folder after Sphinx is finished.)
https://github.com/microsoft/LightGBM/blob/eb9bbfb327815e843b808f5e2b4d73ee3201b4ac/docs/conf.py#L305-L307

PR #3740, opened 2021-01-08
[ci] improve experience with optional GitHub workflows

StrikerRUS:
This PR was created by the inspiration of idea to require optional workflows such as [`R valgrind tests`](https://github.com/microsoft/LightGBM/blob/master/.github/workflows/r_valgrind.yml) (those which takes a lot of time and we run them from time to time) https://github.com/microsoft/LightGBM/pull/3439#issuecomment-704441623 either not be run at all or be completed successfully.

Unfortunately, it is impossible as of my knowledge. Two main reasons are:
- GitHub Actions doesn't allow to pass secret values (we need PAT for the most REST API calls) to forks from public repo ([1](https://github.com/peter-evans/create-pull-request/blob/master/docs/concepts-guidelines.md#restrictions-on-repository-forks), [2](https://github.community/t/make-secrets-available-to-builds-of-forks/16166)).
- GitHub actions doesn't allow to rerun successfully completed workflow ([1](https://github.community/t/is-it-possible-to-manually-force-an-action-workflow-to-be-re-run/2127/22), [2](https://stackoverflow.com/questions/56435547/how-do-i-re-run-github-actions#comment104289650_56810175)). I hope this feature request will be implemented in the near future and we will be able to make some improvements.

OK, probably it **is possible**, but will require switching to [GitHub Checks](https://docs.github.com/en/free-pro-team@latest/rest/reference/checks) from [GitHub Statuses](https://docs.github.com/en/free-pro-team@latest/rest/reference/repos#statuses) and creating own GitHub App. But I think this is an overkill for us.

Due to the limitations described above this PR tries to implement desired behavior as close as possible but there are two cases when maintainer **can** merge a PR with actually failing option workflow. So please **DO NOT do the following scenarios**.
- Due to big queue or long container initialisation phase workflow run was not able to send initial status to the PR from which it was triggered right after a triggering comment. So there is a time delta between triggering comment and initial responce from CI in which maintainer can click `Merge` button. Please after triggering a workflow be patient and allow it to send initial feedback.
- As optional workflows are not marked as `Required` (if they were, it is impossible to implement a case when we do not want to start a particular workflow for a given PR) the first reported status should be checked manually. I believe it is not a very big deal because merge button it case of failure of optional workflow will be grey not green and statuses section will be uncollapsed (see pic. below). But this is applicable only for the **first** reported status, all further statuses will be counted by `Optional checks` **Required** workflow so that any failures will prevent merging.
  image

Also, of course there can happen any network issues that prevent a workflow to send a feedback. But this problem is common to any CI service.

In addition, now optional workflows will be triggered by **ordinary comments** in PRs, not a review ones. This change is required due to secret values limitation described above. GitHub emits runs triggered by review comments associated with a fork PR made from, and runs "belong" to upstream repo in case of issue comment.

Here are some screens from my fork to show how it will look like:

image

image

image

image

Unfortunately, I'm afraid that reviewers are not able to test this PR properly, because workflow configs **should be in `master` branch** to start working. So, please perform real tests of new chatOps functionality in a follow-up fake PR. Just propose any change from a **personal fork** (it is important because of secrets issue described above) and try to trigger optional workflows.

I have been struggling with this for three months, so for sure this my initial description is lacking some details you want to know about. So feel free to ask, I'll recollect them in memory 🙂 .


.github/workflows/optional_checks.yml
```diff
@@ -0,0 +1,25 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2.3.4
+        with:
+          fetch-depth: 5
+          submodules: false
+      - name: Check that all tests succeeded
+        run: |
+            workflows=("R valgrind tests")
```
StrikerRUS, line 20: Here we should add optional workflows which failures should block PR merging.

.github/workflows/r_artifacts.yml
```diff
@@ -37,3 +49,9 @@ jobs:
         with:
           name: ${{ steps.build_package.outputs.artifact_name }}
           path: ${{ steps.build_package.outputs.artifact_path }}
+      - name: Send final status
+        if: ${{ always() }}
+        run: |
+          $GITHUB_WORKSPACE/.ci/append_comment.sh \
+            "${{ github.event.client_payload.comment_number }}" \
+            "Status: ${{ job.status }}."
```
StrikerRUS, line 57: Yes, I know that direct link for artifact downloading will be more useful. But unfortunately, it is impossible: 
> Gets a redirect URL to download an archive for a repository. This URL expires after 1 minute.
   https://docs.github.com/en/free-pro-team@latest/rest/reference/actions#download-an-artifact

jameslamb: ha it's ok, we don't use this job that often and it's only one or two more clicks to get to the artifact

.ci/append_comment.sh
```diff
@@ -0,0 +1,46 @@
...
+fi
+
+comment_id=$1
+body=$2
+
+old_comment_body=$(curl -sL \
+  -H "Accept: application/vnd.github.v3+json" \
+  -H "Authorization: token $SECRETS_WORKFLOW" \
+  "${GITHUB_API_URL}/repos/microsoft/LightGBM/issues/comments/$comment_id" | \
+  jq '.body')
```
jameslamb, line 32:
```suggestion
old_comment_body=$(
  curl -sL \
      -H "Accept: application/vnd.github.v3+json" \
      -H "Authorization: token $SECRETS_WORKFLOW" \
      "${GITHUB_API_URL}/repos/microsoft/LightGBM/issues/comments/$comment_id" | \
      jq '.body'
)
```

I think having the first line end with `(` and having `)` on a line by itself makes it easier to see where this begins and where it ends

jameslamb, line 32:
```suggestion
old_comment_body=$(
  curl -sL \
      -H "Accept: application/vnd.github.v3+json" \
      -H "Authorization: token $SECRETS_WORKFLOW" \
      "${GITHUB_API_URL}/repos/microsoft/LightGBM/issues/comments/$comment_id" | \
      jq '.body'
)
```

I think having the first line end with `(` and having `)` on a line by itself makes it easier to see where this begins and where it ends

.ci/append_comment.sh
```diff
@@ -0,0 +1,46 @@
...
+body=${body/failure/failure ❌}
+body=${body/error/failure ❌}
+body=${body/cancelled/failure ❌}
+body=${body/timed_out/failure ❌}
+body=${body/success/success ✔️}
+data=$(jq -n \
+  --argjson body "${old_comment_body%?}\r\n\r\n$body\"" \
+  '{"body":$body}')
```
jameslamb, line 40: same suggestion here about style for multi-line `$()`

.ci/get_workflow_status.py
```diff
@@ -0,0 +1,82 @@
+# coding: utf-8
+"""Get the most recent status of workflow for the current PR."""
+import json
+from os import environ
+from sys import argv, exit
+from time import sleep
+try:
+    from urllib import request
+except ImportError:
+    import urllib2 as request
```
jameslamb, line 10: why was this necessary? I have a lot more experience with `requests` than `urllib`, so I haven't seen this pattern

jameslamb (changes requested):
This is a great! Thanks for all the hard work. I think this will be really useful.

I read through all the scripts and [the docs on repository_dispatch events](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#repository_dispatch) and I think I understand what's going on here. Left some very minor comments, but after those are addressed I'm +1 on merging this and testing it out.

.ci/append_comment.sh
StrikerRUS: @jameslamb Thanks a lot for your review! I think I've addressed this readability problem in 84a6394705d80868c722305b0a514aedaa7324c4. Please give it a look.

.ci/get_workflow_status.py
StrikerRUS: This for Python 2/3 compatibility. For the same reason I'm not using `with ...` statement but manually close connection. I know we dropped Python 2 support some time ago. But default installed Python version in `ubuntu-latest` container we run our actions in is **2**. I really don't want to install conda/new Python version/additional packages because I want to keep `Optional checks` workflow where this script is used as lightweight as possible with the aim of the fastest execution time. So this Python-agnostic and dependency-free solution will work fine on any default Python version that can be installed in container.

jameslamb: ok I see, thanks. Fine with me

jameslamb (approved):
looks great, thanks! Let's merge it and try it out 🚀

Issue #3450, opened 2020-10-10
CUDA: multi GPUs issue


Brian0906:
I'm trying to use multi-GPUs to train the model. When I increase the number of data, this issue happens. 

Everything goes well if the size of train set is less than 10000.

Operating System:  Linux

CPU/GPU model:  GPU

![image](https://user-images.githubusercontent.com/26130954/95642532-a5296080-0adb-11eb-8d6f-ddbc2494ebfa.png)

StrikerRUS:
@Brian0906 Thanks a lot using experimental CUDA implementation! I observe the same error even with 1 GPU executing [`simple_example.py`](https://github.com/microsoft/LightGBM/blob/master/examples/python-guide/simple_example.py): https://github.com/microsoft/LightGBM/pull/3424#issuecomment-702189313.

Brian0906:
hi @StrikerRUS  what's the size of trainset in the simple_example.py? I fount that only if the size of dataset is large, this issue happens.

StrikerRUS:
Dataset is very small in the example: 7000x28
https://github.com/microsoft/LightGBM/blob/master/examples/regression/regression.train

austinpagan:
Is this the current issue for this problem?  I'm one of the members of the team that ported this CUDA code, and I'm ready to try to reproduce this problem in my environment, if someone can teach me how (preferably with the most simple possible dataset)...

StrikerRUS:
Hello @austinpagan !

Please refer to https://github.com/microsoft/LightGBM/pull/3428#issuecomment-747676987 for the self-contained repro via Docker. Please let me know if you need any additional details.

austinpagan:
Let me apologize if my questions are seen as somehow inappropriate, as I'm rather new to the open source environment...

OK, so three things I'd like to understand, please:
(1) I could not find, in your link, any guidance that I could understand, can you point me more specifically to the recreate scenario?
(2) Is this problem only seen within Docker containers?
(3) I cannot see in this documentation whether you found this problem on a Power system or on an X86 system.  I know it's only happening when you run on CUDA, but we'd still like to understand the environment.

StrikerRUS:
@austinpagan No need to apologize! Let me try to be more precise and do my best to answer your questions.

> (2) Is this problem only seen within Docker containers?

No, this error can be reproduced w/ and w/o Docker. But I believe Docker is the easiest way to reproduce the error on your side as it ensures we are using the same environment.

[...]

PR #3748, opened 2021-01-11
Ensure CUDA vector length is consistent with AlignedSize

ChipKerchner:
@StrikerRUS @austinpagan Fix for [#3450](https://github.com/microsoft/LightGBM/issues/3450)


ChipKerchner:
Compilation errors are due to outstanding fixes in master for CUDA (?)

StrikerRUS (approved):
Brilliant! Thanks a lot for the fix!


> Possible compilation errors are due to outstanding fixes in master related to CUDA (?)

We need one more fix to get CUDA work in `master` (https://github.com/microsoft/LightGBM/issues/3450#issuecomment-756209798). I will submit it right after merging your PR.

PR #3750, opened 2021-01-11
Update CUDA treelearner according to changes introduced for linear trees

StrikerRUS:
Refer to 
- https://github.com/microsoft/LightGBM/issues/3450#issuecomment-756188587
- #3299 
- https://github.com/microsoft/LightGBM/pull/3748#pullrequestreview-565432556

### DIFF
--- a/src/treelearner/cuda_tree_learner.cpp
+++ b/src/treelearner/cuda_tree_learner.cpp
@@ -534,8 +534,8 @@ void CUDATreeLearner::InitGPU(int num_gpu) {
   copyDenseFeature();
 }
 
-Tree* CUDATreeLearner::Train(const score_t* gradients, const score_t *hessians) {
-  Tree *ret = SerialTreeLearner::Train(gradients, hessians);
+Tree* CUDATreeLearner::Train(const score_t* gradients, const score_t *hessians, bool is_first_tree) {
+  Tree *ret = SerialTreeLearner::Train(gradients, hessians, is_first_tree);
   return ret;
 }
 
--- a/.github/workflows/cuda.yml
+++ b/.github/workflows/cuda.yml
@@ -47,6 +47,7 @@ jobs:
             export ROOT_DOCKER_FOLDER=/LightGBM
             cat > docker.env <<EOF
             TASK=cuda
+            METHOD=source
             COMPILER=gcc
             GITHUB_ACTIONS=true
             OS_NAME=linux
--- a/src/treelearner/cuda_tree_learner.h
+++ b/src/treelearner/cuda_tree_learner.h
@@ -45,7 +45,7 @@ class CUDATreeLearner: public SerialTreeLearner {
     ~CUDATreeLearner();
     void Init(const Dataset* train_data, bool is_constant_hessian) override;
     void ResetTrainingDataInner(const Dataset* train_data, bool is_constant_hessian, bool reset_multi_val_bin) override;
-    Tree* Train(const score_t* gradients, const score_t *hessians);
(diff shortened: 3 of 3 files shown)


src/treelearner/cuda_tree_learner.cpp
```diff
@@ -534,8 +534,8 @@ void CUDATreeLearner::InitGPU(int num_gpu) {
   copyDenseFeature();
 }
 
-Tree* CUDATreeLearner::Train(const score_t* gradients, const score_t *hessians) {
-  Tree *ret = SerialTreeLearner::Train(gradients, hessians);
+Tree* CUDATreeLearner::Train(const score_t* gradients, const score_t *hessians, bool is_first_tree) {
```
StrikerRUS, line 537: /gha run cuda-builds

ChipKerchner: Is CUDA not being built automictically with each pull request?

StrikerRUS: You are right. We didn't set it to be built automatically as it was constantly failing.
Now I think we can set it as ordinary test which will be required to be passed to merge new pull requests.

StrikerRUS: /gha run cuda-builds

StrikerRUS: Unfortunately, the error is still here. But with fixes from #3748 **SOME** tests are now passed, while without it all tests were failing.
The error is (I believe the same as before)
```
[LightGBM] [Fatal] [CUDA] invalid argument /LightGBM/src/treelearner/cuda_tree_learner.cpp 414
```
Full test report (`.` - success, `F` - failure, `s` - skipped, `x` - expected to fail):
```
platform linux -- Python 3.8.2, pytest-6.2.1, py-1.10.0, pluggy-0.13.1
rootdir: /LightGBM
collected 238 items

../tests/c_api_test/test_.py ..                                          [  0%]
../tests/python_package_test/test_basic.py ...F..FF.....                 [  6%]
../tests/python_package_test/test_consistency.py ......                  [  8%]
../tests/python_package_test/test_dask.py FFFFFFFFFFFFFFFFFFFFFFFFFF..   [ 20%]
../tests/python_package_test/test_dual.py s                              [ 21%]
../tests/python_package_test/test_engine.py FF...F.....FFFF...FFF..FFFF. [ 32%]
.FFFF.FFFF.....F...FFFFFF...FF.FFF.FF.F                                  [ 49%]
../tests/python_package_test/test_plotting.py F...F                      [ 51%]
../tests/python_package_test/test_sklearn.py .F.F.F.FFFFFFFF.FF..FF...FF [ 62%]
FF.FF.xFF..F..FF...FF.F..F.FFF..F.F..FF.....FF.FxFF..F..FF...FF.F......F [ 92%]
.F..FF.....FF.F..                                                        [100%]
```
https://github.com/microsoft/LightGBM/runs/1682722249?check_suite_focus=true

@guolinke @jameslamb @btrotta I think we need to merge this PR first to sync CUDA code with the rest codebase. Meantime I'll prepare new minimal reproducible example to show it @ChipKerchner .

jameslamb (approved):
seems ok to me. I definitely prefer to work like this, several small PRs that are incremental steps towards enabling CUDA tests automatically on every PR.

src/treelearner/cuda_tree_learner.cpp
ChipKerchner: I'll provide a pull request to fix the above failures.

btrotta (approved):
Looks fine.

PR #3752, opened 2021-01-12
Don't copy more than has been allocated to device_features.

ChipKerchner:
@StrikerRUS Don't copy more than has been allocated to device_features.

Fix CUDA failures [#3450](https://github.com/microsoft/LightGBM/issues/3450)

Possible compilation issues related to [#3750](https://github.com/microsoft/LightGBM/pull/3750)

### DIFF
--- a/src/treelearner/cuda_tree_learner.cpp
+++ b/src/treelearner/cuda_tree_learner.cpp
@@ -408,7 +408,7 @@ void CUDATreeLearner::copyDenseFeature() {
     // looking for dword_features_ non-sparse feature-groups
     if (!train_data_->IsMultiGroup(i)) {
       dense_feature_group_map_.push_back(i);
-      auto sizes_in_byte = train_data_->FeatureGroupSizesInByte(i);
+      auto sizes_in_byte = std::min(train_data_->FeatureGroupSizesInByte(i), static_cast<size_t>(num_data_));
       void* tmp_data = train_data_->FeatureGroupData(i);
       Log::Debug("Started copying dense features from CPU to GPU - 2");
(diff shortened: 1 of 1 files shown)


StrikerRUS:
@guolinke @shiyu1994 @btrotta Could you please help to review this tiny PR with #3750? I'd like to start improving CUDA CI part ASAP but I can't without two these PRs are merged. Sorry for the rush, thanks!

src/treelearner/cuda_tree_learner.cpp
```diff
@@ -408,7 +408,7 @@ void CUDATreeLearner::copyDenseFeature() {
     // looking for dword_features_ non-sparse feature-groups
     if (!train_data_->IsMultiGroup(i)) {
       dense_feature_group_map_.push_back(i);
-      auto sizes_in_byte = train_data_->FeatureGroupSizesInByte(i);
+      auto sizes_in_byte = std::min(train_data_->FeatureGroupSizesInByte(i), static_cast<size_t>(num_data_));
```
guolinke, line 411: should we assert for the type for `FeatureGroupData` ? I think it should be 1-Byte type.

guolinke (approved):

src/treelearner/cuda_tree_learner.cpp
ChipKerchner: I don't think an assert would be meaningful since FeatureGroupSizesInByte is returning a padded size (at least after 3.0.x it is).

StrikerRUS: @guolinke Can we merge or this discussions should be resolved before that?

guolinke: I mean that using `static_cast<size_t>(num_data_)` as the size, the type should be 1-Byte type, right? 
FeatureGroup could be 4-bit, 1Byte, 2Byte and 4Byte. 

However, I think we can merge this first.

PR #3755, opened 2021-01-13
[dask] [python-package] use keyword args for internal function calls

jameslamb:
When reviewing #3515 , I held back some PR review suggestions about code style. I didn't want to hold up the PR for minor style things.

This PR proposes some style changes that I think will make the Dask code easier to read and change.

* uses only keyword arguments for multi-argument function calls
* breaks up long function calls into one argument per line, to more easily see parameters and values
* adds blank line after the end of every method definition
* updates the `lightgbm.dask` module docstring to make it clear that this module originally came from `dask-lightgbm`, which itself originally was based on `dask-xgbooost`


StrikerRUS:
> When reviewing #3515 , I held back some PR review suggestions about code style. I didn't want to hold up the PR for minor style things.

I have some improvements in my mind that I haven't written in my review too. But all massive rewrites of dask.py module will introduce a lot of merge conflicts with #3708. WDYT, is it OK or we should merge #3708 first?

python-package/lightgbm/dask.py
```diff
@@ -262,12 +322,25 @@ class DaskLGBMRegressor(_LGBMModel, LGBMRegressor):
 
     def fit(self, X, y=None, sample_weight=None, client=None, **kwargs):
         """Docstring is inherited from the lightgbm.LGBMRegressor.fit."""
-        return self._fit(LGBMRegressor, X, y, sample_weight, client, **kwargs)
+        return self._fit(
+            model_factoory=LGBMRegressor,
```
StrikerRUS, line 326:
```suggestion
            model_factory=LGBMRegressor,
```

StrikerRUS (approved):
LGTM except 1 typo!

But pleeeease do not adopt old-fashioned 80-char line length limit 😃 !
https://github.com/microsoft/LightGBM/pull/2437#discussion_r329357479

jameslamb:
> LGTM except 1 typo!
> 
> But pleeeease do not adopt old-fashioned 80-char line length limit  !
> [#2437 (comment)](https://github.com/microsoft/LightGBM/pull/2437#discussion_r329357479)

Thanks!

> But all massive rewrites of dask.py module will introduce a lot of merge conflicts with #3708. WDYT, is it OK or we should merge #3708 first?

I looked at the diff in #3708 and I don't think the changes in this PR are going to cause too many issues. Some of these changes are actually changes that are already made in that PR, like adding keyword args instead of positional on some function calls.

I agree that more invasive rewrites that involves moving a lot of code around, like making `_train()` and `_predict()` methods on `_LGBMModel`, should not be done until #3708 is merged.

@ffineis can you look at the diff in this PR and let me know if you're ok with me merging it? If you say "please don't, this would be really annoying for me" we can pause this style-only PR until later.

jameslamb:
Ok I've merged the latest `master` into this and resolved conflicts. I'll do that again after I merge #3755, then merge this.

StrikerRUS (approved):
Re-approving 🙂 .

PR #3757, opened 2021-01-13
[ci] run artifacts packaging job on Ubuntu

StrikerRUS:
I noticed that the availability of Ubuntu machines is greater than Windows ones in general (jobs start faster). Also, Windows machines cost more (see screenshot below, it is for GitHub Actions but they utilize Azure). I think it makes sense to run packaging job on the most cheap and available VM type.

image

### DIFF
--- a/.vsts-ci.yml
+++ b/.vsts-ci.yml
@@ -105,7 +105,7 @@ jobs:
 - job: Windows
 ###########################################
   pool:
-    vmImage: 'ubuntu-latest'
+    vmImage: 'vs2017-win2016'
   strategy:
     matrix:
       regular:
@@ -143,7 +143,7 @@ jobs:
   - Windows
   condition: and(succeeded(), not(startsWith(variables['Build.SourceBranch'], 'refs/pull/')))
   pool:
-    vmImage: 'vs2017-win2016'
+    vmImage: 'ubuntu-latest'
   steps:
   # Download all agent packages from all previous phases
   - task: DownloadBuildArtifacts@0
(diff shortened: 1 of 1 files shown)


jameslamb (approved):
Your reasoning makes sense to me! I don't understand nuget so I don't know if not being on Windows would cause any issues for https://github.com/microsoft/LightGBM/blob/master/.nuget/create_nuget.py. But I think you do understand it well and are proposing this change, so I support it.

StrikerRUS:
Hmm, seems you are right that this change may break NuGet. I thought that `NuGetCommand@2` uses dotNet Core by default and in consequence is cross-platform. But there is a separate [`.NET Core CLI` task](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/build/dotnet-core-cli?view=azure-devops). I'll investigate this later.

StrikerRUS:
OK, I tried to run new pipeline. I can confirm that `NuGetCommand@2` task works fine on Ubuntu, so we don't need to change it to `DotNetCoreCLI@2`.

Here are two archives: one from `master` branch and another one from new modified to run on Ubuntu pipeline. The only "significant" difference I can see is in `LightGBM.3.1.1.99\package\services\metadata\core-properties\<some_sha_here>.psmdcp` file.

from `master` branch:
```
  <lastModifiedBy>NuGet, Version=4.1.0.2450, Culture=neutral, PublicKeyToken=31bf3856ad364e35;Microsoft Windows NT 6.2.9200.0;.NET Framework 4.5</lastModifiedBy>
```

from Ubuntu image:
```
  <lastModifiedBy>NuGet, Version=5.4.0.3, Culture=neutral, PublicKeyToken=31bf3856ad364e35;Unix 5.4.0.1036;.NET Framework 4.7.2</lastModifiedBy>
```

So I think it is safe to use Ubuntu image for packaging task. But I'd like to hear @guolinke's opinion before merging.

[NuGet_Ubuntu.zip](https://github.com/microsoft/LightGBM/files/5817801/NuGet_Ubuntu.zip)
[NuGet_master.zip](https://github.com/microsoft/LightGBM/files/5817802/NuGet_master.zip)

jameslamb:
oh great! thanks for testing

StrikerRUS:
@guolinke Please take a look when have time.

guolinke (approved):

PR #3762, opened 2021-01-14
[ci] Slightly optimize optional workflows checks

StrikerRUS:
By passing code phase instead of workflow name to `get_workflow_status.py` we can filter triggering comments by this phrase. It allows to workaround time delta described in first bad scenario in https://github.com/microsoft/LightGBM/pull/3740#issue-551423557. In other words, even for big queue for optional workflows in case commit was made nearly in the same time as triggering comment, required `Optional checks` workflow will wait for the latest reported status of just triggered optional workflow.


jameslamb (approved):
Awesome work! This is a really nice change, thank you.

Issue #3756, opened 2021-01-13
[python-package] type hints in python package


jameslamb:
## Request for Comment

#3581 officially ended support for Python 2 in the Python package.

Now that Python 2 support has been ended, I'd like to propose adding type hints to the library (https://docs.python.org/3/library/typing.html).

Benefits of adding type hints:

* serves as additional documentation about the types of arguments and return types of functions
* allows us to enable `mypy`, a static type checker for Python that will catch issues like "this function call can return None but you're accessing a property of the result as if it isn't None"
* allows other libraries that integrate with LightGBM (like `optuna`) to catch issues in their own code using `mypy`

I'm opening this Request for Comment to get thoughts on this narrow proposal:

> would you support a PR where I add type hints to the code in the https://github.com/microsoft/LightGBM/blob/master/python-package/lightgbm/dask.py?

## Status: Accepted

~If maintainers agree to this proposal, we could add a "good first issue" about adding hints for other modules and could talk about `mypy` separately, at some point in the future.~

This was originally opened as a request for comment, but it's been accepted and we're now accepting contributions to add type hints in the Python package!

## How to contribute

Anyone is welcome to submit a pull request adding type hints to places in the Python package that are missing them!

Please limit pull requests to 1 function/method per pull request, to make them easier to review.

* ~`.ci/get_workflow_status.py`~ (shouldn't be touched because this file is executed by default Python 2 installed on GitHub Actions workers)
* [x] `docs/conf.py` (#4526)
* [x] `helpers/check_dynamic_dependencies.py` (#4382)
* [x] `helpers/parameter_generator.py` (#4474)
* [ ] `python-package/lightgbm/basic.py` (#7140)
* [x] `python-package/lightgbm/callback.py` (#4093)
* [x] `python-package/lightgbm/compat.py`
* [x] `python-package/lightgbm/dask.py` (#3866)
* [x] `python-package/lightgbm/engine.py`
* [x] `python-package/lightgbm/libpath.py` (#4070)
* [x] `python-package/lightgbm/plotting.py` (#4367)
* [ ] `python-package/lightgbm/sklearn.py` (#7140)
* [x] `python-package/setup.py` (#4376)

### Handling line breaks

If adding hints to a method or function greatly increases the length of its lines, break each argument onto a separate line and have the closing parenthesis begin its own line, vertically aligned with the keyword `def`. Like this:

```python
# before
def make_ranking(n_samples=100, n_features=20, n_informative=5, gmax=2,
[...]

PR #3766, opened 2021-01-14
[dask] [python-package] Search for available ports when setting up network (fixes #3753)

### LINKED ISSUE #3753: [dask] [python-package] DaskLGBMRegressor training error: 'binding port 12402 failed'

## Environment info

Operating System: Ubuntu 18.04

C++ compiler version: 9.3.0

CMake version: 3.16.3

Python version: 3.8.5

LightGBM version or commit hash: `master` (https://github.com/microsoft/LightGBM/commit/78d31d9ae34c9c8ab2b0f8704c6962c8d3510062)

Training with `DaskLGBMRegressor` often fails with an error like this:

> LightGBMError: Binding port 12402 failed

The error doesn't ALWAYS happen, and training sometimes succeeds. It also doesn't always reference port 12402. I've found that the first time I call `DaskLGBMRegressor.fit()`, I don't see this error. After that, subsequent tries often result in the error.

Here's the result of 10 calls of `.fit()`, with `client.restart()` run after each one.

1. SUCCESS
2. FAIL (Binding port 12403 failed)
3. FAIL (Binding port 12402 failed)
4. SUCCESS
5. FAIL (Binding port 12402 failed)
6. FAIL (Binding port 12403 failed)
7. FAIL (Binding port 12401 failed)
8. FAIL (Binding port 12403 failed)
9. FAIL (Binding port 12403 failed)
10. SUCCESS
full traceback:

```text
---------------------------------------------------------------------------
LightGBMError                             Traceback (most recent call last)
<ipython-input-17-aea611c2313b> in <module>
     27 )
     28 
---> 29 dask_reg.fit(
     30     client=client,
     31     X=data,

/opt/conda/lib/python3.8/site-packages/lightgbm/dask.py in fit(self, X, y, sample_weight, client, **kwargs)
    263     def fit(self, X, y=None, sample_weight=None, client=None, **kwargs):
    264         """Docstring is inherited from the lightgbm.LGBMRegressor.fit."""
--> 265         return self._fit(LGBMRegressor, X, y, sample_weight, client, **kwargs)
    266     fit.__doc__ = LGBMRegressor.fit.__doc__
    267 

/opt/conda/lib/python3.8/site-packages/lightgbm/dask.py in _fit(self, model_factory, X, y, sample_weight, client, **kwargs)
    209 
    210         params = self.get_params(True)
--> 211         model = _train(client, X, y, params, model_factory, sample_weight, **kwargs)
    212 
    213         self.set_params(**model.get_params())

/opt/conda/lib/python3.8/site-packages/lightgbm/dask.py in _train(client, data, label, params, model_factory, weight, **kwargs)
    151                            for worker, list_of_parts in worker_map.items()]
    152 
--> 153     results = client.gather(futures_classifiers)
    154     results = [v for v in results if v]
    155     return results[0]

/opt/conda/lib/python3.8/site-packages/distributed/client.py in gather(self, futures, errors, direct, asynchronous)
   1983             else:
   1984                 local_worker = None
-> 1985             return self.sync(
   1986                 self._gather,
   1987                 futures,

/opt/conda/lib/python3.8/site-packages/distributed/client.py in sync(self, func, asynchronous, callback_timeout, *args, **kwargs)
    829             return future
    830         else:
--> 831             return sync(
    832                 self.loop, func, *args, callback_timeout=callback_timeout, **kwargs
    833             )

/opt/conda/lib/python3.8/site-packages/distributed/utils.py in sync(loop, func, callback_timeout, *args, **kwargs)
    338     if error[0]:
    339         typ, exc, tb = error[0]
--> 340         raise exc.with_traceback(tb)
    341     else:
    342         return result[0]

/opt/conda/lib/python3.8/site-packages/distributed/utils.py in f()
    322             if callback_timeout is not None:
    323                 future = asyncio.wait_for(future, callback_timeout)
--> 324             result[0] = yield future
    325         except Exception as exc:
    326             error[0] = sys.exc_info()

/opt/conda/lib/python3.8/site-packages/tornado/gen.py in run(self)
    733 
    734                     try:
--> 735                         value = future.result()
    736                     except Exception:
    737                         exc_info = sys.exc_info()

/opt/conda/lib/python3.8/site-packages/distributed/client.py in _gather(self, futures, errors, direct, local_worker)
   1848                             exc = CancelledError(key)
   1849                         else:
-> 1850                             raise exception.with_traceback(traceback)
   1851                         raise exc
   1852                     if errors == "skip":

/opt/conda/lib/python3.8/site-packages/lightgbm/dask.py in _train_part()
     77     try:
     78         model = model_factory(**params)
---> 79         model.fit(data, label, sample_weight=weight, **kwargs)
     80     finally:
     81         _safe_call(_LIB.LGBM_NetworkFree())

/opt/conda/lib/python3.8/site-packages/lightgbm/sklearn.py in fit()
    768             callbacks=None, init_model=None):
    769         """Docstring is inherited from the LGBMModel."""
--> 770         super().fit(X, y, sample_weight=sample_weight, init_score=init_score,
    771                     eval_set=eval_set, eval_names=eval_names, eval_sample_weight=eval_sample_weight,
    772                     eval_init_score=eval_init_score, eval_metric=eval_metric,

/opt/conda/lib/python3.8/site-packages/lightgbm/sklearn.py in fit()
    610             init_model = init_model.booster_
    611 
--> 612         self._Booster = train(params, train_set,
    613                               self.n_estimators, valid_sets=valid_sets, valid_names=eval_names,
    614                               early_stopping_rounds=early_stopping_rounds,

/opt/conda/lib/python3.8/site-packages/lightgbm/engine.py in train()
    226     # construct booster
    227     try:
--> 228         booster = Booster(params=params, train_set=train_set)
    229         if is_valid_contain_train:
    230             booster.set_train_data_name(train_data_name)

/opt/conda/lib/python3.8/site-packages/lightgbm/basic.py in __init__()
   2060                     else:
   2061                         raise ValueError("Invalid machines in params.")
-> 2062                     self.set_network(machines,
   2063                                      local_listen_port=params.get("local_listen_port", 12400),
   2064                                      listen_time_out=params.get("listen_time_out", 120),

/opt/conda/lib/python3.8/site-packages/lightgbm/basic.py in set_network()
   2196             Booster with set network.
   2197         """
-> 2198         _safe_call(_LIB.LGBM_NetworkInit(c_str(machines),
   2199                                          ctypes.c_int(local_listen_port),
   2200                                          ctypes.c_int(listen_time_out),

/opt/conda/lib/python3.8/site-packages/lightgbm/basic.py in _safe_call()
     50     """
     51     if ret != 0:
---> 52         raise LightGBMError(_LIB.LGBM_GetLastError().decode('utf-8'))
     53 
     54 

LightGBMError: Binding port 12401 failed
```

## Steps to reproduce

This is based on https://github.com/jameslamb/talks/tree/main/recent-developments-in-lightgbm, but copying in the steps in case that repo changes in the future.

1. Create a file `Dockerfile` with the following contents

```dockerfile
FROM daskdev/dask-notebook:latest

# based on
# https://github.com/microsoft/LightGBM/pull/3515#pullrequestreview-543175003
ENV LIGHTGBM_DIR=${HOME}/LightGBM
ENV LIGHTGBM_URL="https://github.com/microsoft/LightGBM.git"
ENV LIGHTGBM_COMMIT="78d31d9ae34c9c8ab2b0f8704c6962c8d3510062"

RUN apt update && \
    apt-get install -y \
        build-essential \
        cmake

RUN git clone --recursive ${LIGHTGBM_URL} ${LIGHTGBM_DIR} && \
    cd ${LIGHTGBM_DIR} && \
    git checkout ${LIGHTGBM_COMMIT}

RUN conda install -y -c conda-forge \
    dask-cloudprovider \
    scikit-learn \
    numpy \
    pandas && \
    # install LightGBM
    cd ${LIGHTGBM_DIR}/python-package && \
    python setup.py sdist && \
    pip install dist/lightgbm*.tar.gz && \
    conda clean --all

WORKDIR /home/jovyan/testing
```

2. Build an image from that file

```shell
docker build --no-cache -t dask-lgbm-testing:1 -f Dockerfile .
```

3. Run Jupyter Lab in a container built with that image.

```shell
docker run \
    -v $$(pwd):/home/jovyan/testing \
    -p 8888:8888 \
    -p 8787:8787 \
    --name dask-lgbm-test \
   dask-lgbm-testing:1
```

4. When that starts up, it will print a message like the one shown below. Copy the URL starting with 127.0.0.1 and paste it into a web browser, to access Jupyter Lab.

```text
[C 22:10:48.337 LabApp] 
    
    To access the notebook, open this file in a browser:
        file:///home/jovyan/.local/share/jupyter/runtime/nbserver-19-open.html
    Or copy and paste one of these URLs:
        http://4cfa66d00717:8888/?token=0122fc992c4f4485937224a29d0d65e798e2168c5e20286e
     or http://127.0.0.1:8888/?token=0122fc992c4f4485937224a29d0d65e798e2168c5e20286e
```

5. In a notebook, run the following code one time.

```python
import dask.array as da
from lightgbm.dask import DaskLGBMRegressor

n_workers = 3
client = Client(cluster)

```

That should succeed, and if you click the printed link, you should see the Dask diagnostic dashboard.

image

6. In that same notebook, run the code below multiple times. This code starts with a `client.restart()`, which clears the memory on all worker processes and removes any work from the scheduler.

```python
client.restart()

num_rows = 1e6
num_features = 1e2
num_partitions = 10
rows_per_chunk = num_rows / num_partitions

data = da.random.random((num_rows, num_features), (rows_per_chunk, num_features))

labels = da.random.random((num_rows, 1), (rows_per_chunk, 1))

data = data.persist()
labels = labels.persist()
_ = wait(data)
_ = wait(labels)

dask_reg = DaskLGBMRegressor(
    silent=False,
    max_depth=5,
    random_state=708,
    objective="regression_l2",
    learning_rate=0.1,
    tree_learner="data",
    n_estimators=10,
    min_child_samples=1,
    n_jobs=-1,
)

dask_reg.fit(
    client=client,
    X=data,
    y=labels,
)
```

I expect that you'll see a similar pattern as the one noted above. Training will sometimes succeed, but often fail with an error like "cannot bind port XXXX".

## Other information

I've noticed that often when this happens, it seems like maybe some of the worker processes were killed and restarted. I don't see messages about that in the logs, but the memory utilization for the workers is really uneven.

image

I've observed this behavior on `FargateCluster`s from `dask-cloudprovider` and on the `dask-kubernetes` clusters from [Saturn Cloud](https://www.saturncloud.io/s/). So I don't think this issue is specific to the Dask docker image I used in the example below, or to the use of `LocalCluster`.

I've also observed this behavior using [dask-lightgbm](https://github.com/dask/dask-lightgbm) built from current `master`, with LightGBM 3.0.0.

Given all of that, my best guess is that there is some race condition where workers join the LightGBM cluster in a nondeterministic order, or maybe where two workers claim the same rank.

jameslamb:
See #3753 for additional details.

## The Problem

By default, LightGBM's Python package is built to use TCP sockets for distributed learning (`-DUSE_SOCKET=1`). When performing distributed learning, LightGBM requires a list of machines and ports, as is documented [here](https://lightgbm.readthedocs.io/en/latest/Parallel-Learning-Guide.html#socket-version).

```text
machine1_ip 12345
machine2_ip 12345
```

`lightgbm.dask` currently creates this list by assigning a fixed range of ports. Assuming that you pass parameters `local_listen_port=12400`, for example, the first worker will open a socket `{first_worker_ip}: 12400`, the second will open one `{second_worker_ip}:12401`, and so on.

This approach is fragile, since it makes the assumption that the ports you request will definitely be available. It can lead to an error like this:

> Binding to port 12402 failed

This can happen either because of connections not quite cleaned up after previous LightGBM training runs (https://github.com/microsoft/LightGBM/issues/3753#issuecomment-759683870) or because of any other processes running on that machine which might already be using a port that LightGBM wants.

## Changes in this PR

* `lightgbm.dask` will now search for open ports on each worker when constructing the network. It uses `local_listen_port` as a starting point, but doesn't assume that that port or any other specific port is available.
    - this is now done one time in `_train()`, before setting up the network. On `master` today, the code to set up the parameters is re-run in `_train_part()` (once on each worker) even though it produces the same results
 * refactors the internals for constructing network parameters
 * adds tests that you can call `.fit()` with a Dask model many times in a row with no sleeping in between...this is the primary way that the bug in #3753 shows up
 * adds type hints to only new code added in this PR (per https://github.com/microsoft/LightGBM/issues/3756)

## Notes for reviewers

I tested this with both a `LocalCluster` with 3 workers and a [`dask_cloudprovider.aws.FargateCluster`](https://cloudprovider.dask.org/en/latest/aws.html#fargate), using the steps described in https://github.com/microsoft/LightGBM/issues/3753#issue-784647300 and the notebooks in https://github.com/jameslamb/talks/tree/main/recent-developments-in-lightgbm. ~I think I'll set up a separate small repo with my testing setup so that this can be more easily re-used by contributors in the future.~ This can be tested with the setup in https://github.com/jameslamb/lightgbm-dask-testing.

I captured the screenshot below to show some more evidence that this worked. I had the code print out its decision for ports on each worker, and I ran training 10 times in a row in a for loop like this:

```python
for i in range(10):
    client.restart()
    ...
    dask_reg.fit(...)
```

You can see that on some runs, it chose 12400, 12401, and 12402. But on others, LightGBM chose to skip up a few ports. That's the "search until you find one" stuff working! The picture on the right side of the screenshot shows the individual training runs succeeding one after another.

Screenshot from 2021-01-14 14-42-02

Since this worked well on both `LocalCluster` (where all workers have the same IP address) and `FargateCluster` (where each worker is a different physical machine, with a unique IP address), I feel confident in this fix.

special thanks to @imatiach-msft and @ffineis for their help figuring this out!


python-package/lightgbm/dask.py
```diff
@@ -23,33 +25,89 @@
...
+    for worker_address in worker_addresses:
+        port = client.submit(
+            func=_find_open_port,
+            workers=[worker_address],
+            worker_ip=urlparse(worker_address).hostname,
+            local_listen_port=local_listen_port,
+            ports_to_skip=lightgbm_ports
+        ).result()
+        lightgbm_ports.add(port)
+        worker_ip_to_port[worker_address] = port
```
jameslamb, line 108: I want to draw reviewers' attention to this block. This setup checks for ports to use sequentially. So it will first run `_find_open_port()` on worker 1's machine, then worker 2's machine, then work 3's, etc.

That means there is a small startup time introduced. I saw that this process took about 5 seconds per worker when I was running a cluster on AWS Fargate. So that means you might expect around 1 minute of startup time for a cluster with 20 workers.

I don't think this is too bad, and I think it's a fairly clean way to ensure that if you have multiple work processes on the same machine, they're assigned different ports in the network parameters

jameslamb:
hmmm, never saw this error before. Spotted on [the `macOS-latest`, `TASK=REGULAR` build](https://github.com/microsoft/LightGBM/pull/3766/checks?check_run_id=1705073676). Will try rebuilding, maybe it's a temporary thing.

```text
[NbConvertApp] Converting notebook interactive_plot_example.ipynb to notebook
  File "/Users/runner/miniconda/envs/test-env/bin/jupyter-nbconvert", line 11, in <module>
    sys.exit(main())
  File "/Users/runner/miniconda/envs/test-env/lib/python3.6/site-packages/jupyter_core/application.py", line 254, in launch_instance
    return super(JupyterApp, cls).launch_instance(argv=argv, **kwargs)
  File "/Users/runner/miniconda/envs/test-env/lib/python3.6/site-packages/traitlets/config/application.py", line 664, in launch_instance
    app.start()
  File "/Users/runner/miniconda/envs/test-env/lib/python3.6/site-packages/nbconvert/nbconvertapp.py", line 350, in start
    self.convert_notebooks()
  File "/Users/runner/miniconda/envs/test-env/lib/python3.6/site-packages/nbconvert/nbconvertapp.py", line 524, in convert_notebooks
    self.convert_single_notebook(notebook_filename)
  File "/Users/runner/miniconda/envs/test-env/lib/python3.6/site-packages/nbconvert/nbconvertapp.py", line 489, in convert_single_notebook
    output, resources = self.export_single_notebook(notebook_filename, resources, input_buffer=input_buffer)
  File "/Users/runner/miniconda/envs/test-env/lib/python3.6/site-packages/nbconvert/nbconvertapp.py", line 418, in export_single_notebook
    output, resources = self.exporter.from_filename(notebook_filename, resources=resources)
  File "/Users/runner/miniconda/envs/test-env/lib/python3.6/site-packages/nbconvert/exporters/exporter.py", line 181, in from_filename
    return self.from_file(f, resources=resources, **kw)
  File "/Users/runner/miniconda/envs/test-env/lib/python3.6/site-packages/nbconvert/exporters/exporter.py", line 199, in from_file
    return self.from_notebook_node(nbformat.read(file_stream, as_version=4), resources=resources, **kw)
  File "/Users/runner/miniconda/envs/test-env/lib/python3.6/site-packages/nbformat/__init__.py", line 143, in read
    return reads(buf, as_version, **kwargs)
  File "/Users/runner/miniconda/envs/test-env/lib/python3.6/site-packages/nbformat/__init__.py", line 77, in reads
    validate(nb)
  File "/Users/runner/miniconda/envs/test-env/lib/python3.6/site-packages/nbformat/validator.py", line 270, in validate
    relax_add_props=relax_add_props):
  File "/Users/runner/miniconda/envs/test-env/lib/python3.6/site-packages/nbformat/validator.py", line 303, in iter_validate
    validator = get_validator(version, version_minor, relax_add_props=relax_add_props)
  File "/Users/runner/miniconda/envs/test-env/lib/python3.6/site-packages/nbformat/validator.py", line 57, in get_validator
    schema_json = _get_schema_json(v, version=version, version_minor=version_minor)
  File "/Users/runner/miniconda/envs/test-env/lib/python3.6/site-packages/nbformat/validator.py", line 94, in _get_schema_json
    with open(schema_path) as f:
FileNotFoundError: [Errno 2] No such file or directory: '/Users/runner/miniconda/envs/test-env/lib/python3.6/site-packages/nbformat/v4/nbformat.v4.2.schema.json'
```

python-package/lightgbm/dask.py
```diff
@@ -138,13 +207,22 @@ def _train(client, data, label, params, model_factory, weight=None, **kwargs):
                        '(%s), using "data" as default', params.get("tree_learner", None))
         params['tree_learner'] = 'data'
 
+    # find an open port on each worker. note that multiple workers can run
+    # on the same machine, so this needs to ensure that each one gets its
+    # own port
+    local_listen_port = params.get('local_listen_port', 12400)
+    worker_address_to_port = _find_ports_for_workers(
+        client=client,
+        worker_addresses=list(worker_map.keys()),
```
StrikerRUS, line 216: `dict_keys` class suits in `Iterable[str]`. Maybe no need to convert?
```suggestion
        worker_addresses=worker_map.keys(),
```

tests/python_package_test/test_dask.py
```diff
@@ -89,6 +90,27 @@ def test_classifier(output, centers, client, listen_port):
     assert_eq(y, p2)
 
 
+def test_training_does_not_fail_on_port_conflicts(client):
+    X, y, w, dX, dy, dw = _create_data('classification', output='array')
```
StrikerRUS, line 94:
```suggestion
    _, _, _, dX, dy, dw = _create_data('classification', output='array')
```

tests/python_package_test/test_dask.py
```diff
@@ -183,21 +205,30 @@ def test_regressor_local_predict(client, listen_port):
...
+            ports_to_skip=set()
+        )
+        assert new_port == 12401
+
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+        s.bind((worker_ip, 12400))
+        s.listen()
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+            s.bind((worker_ip, 12401))
+            s.listen()
```
StrikerRUS, line 225: For readability.
```suggestion
    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s_1:
        s_1.bind((worker_ip, 12400))
        s_1.listen()
        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s_2:
            s_2.bind((worker_ip, 12401))
            s_2.listen()
```

StrikerRUS (commented):
@jameslamb Just to clarify: I have literally zero experience with Dask. Sorry.

However, all your thoughts sound reasonable to me, so I think I can approve this PR.
Just please take a look at some general Python comments I left.

python-package/lightgbm/dask.py
```diff
@@ -23,33 +25,89 @@
...
+    """
+    max_tries = 1000
+    out_port = None
+    found_port = False
+    for i in range(max_tries):
+        out_port = local_listen_port + i
+        try:
+            out_port = local_listen_port + i
+            if out_port in ports_to_skip:
+                continue
```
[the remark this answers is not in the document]
jameslamb: I agree

python-package/lightgbm/dask.py
```diff
@@ -23,33 +25,89 @@
...
+    result : int
+        A free port on the machine referenced by ``worker_ip``.
+    """
+    max_tries = 1000
+    out_port = None
+    found_port = False
+    for i in range(max_tries):
+        out_port = local_listen_port + i
+        try:
+            out_port = local_listen_port + i
```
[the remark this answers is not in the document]
jameslamb: ah! nope, that's a mistake. Thank you

python-package/lightgbm/dask.py
```diff
@@ -23,33 +25,89 @@
...
+            if out_port in ports_to_skip:
```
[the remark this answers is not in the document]
jameslamb: it will already be a set when it's passed in, so I don't think that's necessary

https://github.com/jameslamb/LightGBM/blob/fix/dask-network-setup/python-package/lightgbm/dask.py#L97

https://github.com/jameslamb/LightGBM/blob/fix/dask-network-setup/python-package/lightgbm/dask.py#L107

python-package/lightgbm/dask.py
```diff
@@ -23,33 +25,89 @@
...
+
+
+def _find_ports_for_workers(client: Client, worker_addresses: Iterable[str], local_listen_port: int) -> Dict[str, int]:
+    """Find an open port on each worker.
+
+    LightGBM distributed training uses TCP sockets by default, and this method is used to
+    identify open ports on each worker so LightGBM can reliable create those sockets.
+
+    Returns a dictionary where keys are
+    worker addresses and values are an open port for LightGBM to use.
```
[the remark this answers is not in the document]
jameslamb: oh! my mistake, thanks for catching that

python-package/lightgbm/dask.py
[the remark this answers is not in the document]
StrikerRUS: OK, as it is a private API, I'm fine with it then.

python-package/lightgbm/dask.py
```diff
@@ -23,33 +25,89 @@
...
 
     Parameters
     ----------
-    worker_addresses : iterable of str - collection of worker addresses in `<protocol>://<host>:port` format
-    local_worker_ip : str
+    client : dask.distributed.Client
+        Dask client.
+    worker_addresses : Iterable[str]
+        An iterable of addresses for workers in the cluster. These are strings of the form ```<protocol>://<host>:port```
```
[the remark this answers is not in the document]
jameslamb: whoops! no that was meant to be 2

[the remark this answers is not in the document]
jameslamb:
```suggestion
        An iterable of addresses for workers in the cluster. These are strings of the form ``<protocol>://<host>:port``
```

python-package/lightgbm/dask.py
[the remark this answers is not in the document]
jameslamb: moved in https://github.com/microsoft/LightGBM/pull/3766/commits/7aca0673c2ced0c1f24ccdb76314bfa6a92b1a66

python-package/lightgbm/dask.py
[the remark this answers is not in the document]
jameslamb: fixed in https://github.com/microsoft/LightGBM/pull/3766/commits/7aca0673c2ced0c1f24ccdb76314bfa6a92b1a66

jameslamb:
> @jameslamb Just to clarify: I have literally zero experience with Dask. Sorry.

@StrikerRUS it's ok, I understand! I'm not sure if any other maintainers have Dask experience. If there's ever a Dask PR that you feel is large or controversial, we can probably get some reviewing help from our friends at XGBoost or I can ask others in the Dask community.

python-package/lightgbm/dask.py
```diff
@@ -23,33 +25,85 @@
...
+    out_port = None
+    found_port = False
+    for i in range(max_tries):
+        out_port = local_listen_port + i
+        if out_port in ports_to_skip:
+            continue
+        try:
+            with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+                s.bind((worker_ip, out_port))
+                s.listen()
```
ffineis, line 61: Point of calling `s.listen()` here? Is it necessary to check for ability to bind socket?

python-package/lightgbm/dask.py
ffineis: This is probably overkill, but if you wanted to speed it up, maybe could group by IP address and apply the search in parallel to all workers on the same node:

```
def _find_ports_for_workers(client: Client, worker_addresses: Iterable[str], local_listen_port: int) -> Dict[str, int]:
    worker_ip_to_port = {}

    # group workers by ip addr
    worker_hostname_map = defaultdict(list())
    for worker_address in worker_addresses:
    	hostname = urlparse(worker_address).hostname
    	worker_hostname_map[hostname].append(worker_address)

    # run search for ports on groups of workers
   	lgbm_machines = list()
   	for hostname in worker_hostname_map:
   		machines_on_node = client.submit(
            func=_port_search_for_ip,
            worker_ip=hostname,
            desired_ports=len(worker_hostname_map[hostname]),
            local_listen_port=local_listen_port
        )

        # add more futures
        lgbm_machines.append(machines_on_node)

    # wait for search across nodes to complete.
    _ = wait(lgbm_machines)
    lgbm_machines = lgbm_machines.results()


def _port_search_for_ip(worker_ip: str, local_listen_port: int, n_desired_ports: int) -> Iterable[str]:
    max_tries = 1000
    out_ports = list()
    found_all_ports = False
    for i in range(max_tries):
        out_port = local_listen_port + i
        try:
                s.bind((worker_ip, out_port))
                s.listen()

            out_ports.append(out_port)
            if len(out_ports) == n_desired_ports:
            	found_all_ports = True
            	break

        except OSError:
            continue

    if not found_all_ports:
        raise RuntimeError()

    return [hostname + ':' + str(x) for x in out_ports]
```

Otherwise, the contraints of lightgbm + dask are that: 1) lightgbm requires that each worker knows about all of the other workers while at the same time 2) prior to distributing work to each worker we know which workers are going to be listening for lightgbm traffic on which ports. Seems like your fix nails this!

StrikerRUS:
@jameslamb 
> we can probably get some reviewing help from our friends at XGBoost or I can ask others in the Dask community.

Asking for a help from outside reviewers or maybe even find someone interested and experienced in LightGBM and Dask to collaborate on improvements in Dask-package on a regular basis supporting you will be just awesome.

tests/python_package_test/test_dask.py
```diff
@@ -183,21 +205,30 @@ def test_regressor_local_predict(client, listen_port):
...
+            ports_to_skip=set()
+        )
+        assert new_port == 12401
+
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s_1:
+        s_1.bind((worker_ip, 12400))
+        s_1.listen()
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s_2:
+            s_2.bind((worker_ip, 12401))
+            s_2.listen()
```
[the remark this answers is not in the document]
jameslamb: I don't think it should be a problem, because the context manager closes the socket it creates.

In other words, when Python exits the block `with socket.socket(...) as s:`, `socket` closes the connection `s`. This is just like how `with open("file.txt", "r") as f` closes a file connection for you.

python-package/lightgbm/dask.py
jameslamb: ah! you're totally right, it can be removed. I tested that like this:

```python
    s.bind(('127.0.0.1', 12400))
    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s_2:
        s_2.bind(('127.0.0.1', 12400))
```

You can see that without `listen()`, you'll still get an `OSError: Address already in use` from that second socket.

Thanks for catching it. Removed in https://github.com/microsoft/LightGBM/pull/3766/commits/4ee555786fe5ba00859c6215e694a9e0100daa6a

python-package/lightgbm/dask.py
jameslamb: > This is probably overkill, but if you wanted to speed it up, maybe could group by IP address and apply the search in parallel across separate nodes (IP addresses)? So each the search for ports gathers as many ports as there are workers on that server

That's a great idea, I like it! I think I'll push it to a later PR though, since it would take me some time to test and I really want to unblock you for #3708

jameslamb:
I've moved my test setup out into its own repo: https://github.com/jameslamb/lightgbm-dask-testing. I'm going to test this on `LocalCluster` and `FargateCluster` one more time, to be sure I didn't make any mistakes responding to review comments. If all goes well, I'll merge after that and go review #3708

python-package/lightgbm/dask.py
ffineis: GH needs more emojis but if I could, I'd give you the prayer "bless up" hands

ffineis (approved):

python-package/lightgbm/dask.py
jameslamb: haha thank you

jameslamb:
happy to say this worked on `LocalCluster` and `FargateCluster`. Thanks for the help and reviews!

python-package/lightgbm/dask.py
jameslamb: I just documented this in https://github.com/microsoft/LightGBM/issues/3768, thanks for taking the time to type out that possible implementation. Super helpful.

PR #3764, opened 2021-01-14
[python][tests] Migrates test_basic.py to use pytest

thomasjpfan:
Related to https://github.com/microsoft/LightGBM/issues/3732

This PR migrates the unitest-based test to use pytest. Additionally, this PR uses the `tmp_path` fixture when the test needs to create a temporary file.

PR #3767, opened 2021-01-15
[tests][python] completely remove tempfile from test_basic

StrikerRUS:
This is the last tempfile left after the main work on replacing it with pytest fixture in #3764.


jameslamb (approved):
Nice! Yeah I think the fixture is better than doing this ourselves with `NamedTemporaryFile`. Thanks for the change.

PR #3770, opened 2021-01-15
[dask] make pandas optional for lightgbm.dask

StrikerRUS:
We use this approach for all other our modules.

### DIFF
--- a/python-package/lightgbm/dask.py
+++ b/python-package/lightgbm/dask.py
@@ -17,7 +17,7 @@ from dask import delayed
 from dask.distributed import Client, default_client, get_worker, wait
 
 from .basic import _LIB, _safe_call
-from .compat import DataFrame, Series
+from .compat import PANDAS_INSTALLED, DataFrame, Series
 from .sklearn import LGBMClassifier, LGBMRegressor
 
 import scipy.sparse as ss
--- a/python-package/lightgbm/dask.py
+++ b/python-package/lightgbm/dask.py
@@ -11,13 +11,13 @@ from typing import Dict, Iterable
 from urllib.parse import urlparse
 
 import numpy as np
-import pandas as pd
 from dask import array as da
 from dask import dataframe as dd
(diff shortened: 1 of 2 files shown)


python-package/lightgbm/dask.py
```diff
@@ -11,13 +11,13 @@
 from urllib.parse import urlparse
 
 import numpy as np
-import pandas as pd
 from dask import array as da
 from dask import dataframe as dd
```
jameslamb, line 15: If you do this, then you need to make `dask.dataframe` a conditional import as well.

`pandas` is a required dependency of dask dataframe, so we can't require dask dataframe without requiring pandas. 

Try this

```shell
pip install --upgrade dask
pip uninstall -y pandas
python -c "from dask import dataframe as dd"
```

You'll get this error

```text
  File "/home/jlamb/miniconda3/lib/python3.7/site-packages/dask/dataframe/__init__.py", line 2, in <module>
    from .core import (
  File "/home/jlamb/miniconda3/lib/python3.7/site-packages/dask/dataframe/core.py", line 11, in <module>
    import pandas as pd
ModuleNotFoundError: No module named 'pandas'

The above exception was the direct cause of the following exception:

  File "/home/jlamb/miniconda3/lib/python3.7/site-packages/dask/dataframe/__init__.py", line 57, in <module>
    raise ImportError(msg) from e
ImportError: Dask dataframe requirements are not installed.

Please either conda or pip install as follows:

  conda install dask                     # either conda install
  python -m pip install "dask[dataframe]" --upgrade  # or python -m pip install
```

I don't think it's problematic to say that `pandas` is required to use the Dask interface.

jameslamb (changes requested):

python-package/lightgbm/dask.py
StrikerRUS: Ah, I didn't know that it is a transitive dependency. Thanks for pointing to it!

Issue #3751, opened 2021-01-12
Issue with call to c_api Predict functions in multi threaded way from golang


tarunreddy1018:
**How we are using LightGBM:**

We are using the LightGBM c_api in our service. But this c_api is being called from golang. Basically we have a golang wrapper around the c_api. And we call the c_api functions from golang using cgo.

We get the “lib_lightgbm.so”  library file from the Github release section and use them.

**Version of LightGBM being used:**

3.1.1 (Observed with all versions >= 3.0.0)

**LightGBM component:** 

C++ Library

**Environment info:**

Operating System:  Observed on both Linux and MacOS

Architecture: x86_64

CPU model: Intel(R) Xeon(R) Platinum 8259CL CPU @ 2.50GHz

C++ compiler version: gcc version 7.2.0 (Debian 7.2.0-1)

CMake version: 3.9.0

GLIBC version: ldd (Debian GLIBC 2.28-10) 2.28

**Context:**

Starting with LightGbm 3.0.0 version, there is an update in Predict functions where only the necessary parts of the predict function have been given with "unique_lock", and rest of the parts where they just do computation were given with "shared_lock".

Unlike the previous version like 2.3.1, where a "lock_guard" was used in the beginning of the predict function itself. This led to the situation where only one thread can execute this predict function at a time. This could not scale better in production environments where we do many predictions in a parallel way.

So, we decided to upgrade to a newer version of LightGbm 3.1.1. Which really improved the performance. Since now multiple threads can execute the predict function in a parallel way because of the "shared_lock". 

**Issue:**

But, Now we saw that the predictions were not consistent. That is if there are many threads which are invoking this predict function in a parallel way on a large scale, And if we give the same input multiple times to the service, it turns out that we are getting different predictions each time we make a call. 

This was not the issue with earlier LightGBM version that we used 2.3.1. All the predictions were consistent with that version. It could probably be because of the lock that it had in the predict function because of which we did not notice this issue.

To test this out quickly, we had to put a lock before we made a call to predict function. And now all the predictions were as expected and consistent.

So, we thought that there could be some race condition in the code or LightGbm was not meant to do predictions in a parallel way. To investigate it further we started looking into the c_api of predict function. And found out that it has a vector of vectors from which each thread gets its vector using thread id (tid) and copies the input data into that vector and passes it to predict function. Once the predict call is completed it clears the vector.

Definition of Vector of Vectors:  
[...]

Issue #3675, opened 2020-12-24
Getting random outputs when calling LightGBM Predict for single row function in a multi threaded environment (LGBM_BoosterPredictForMatSingleRowFast) method


tarunreddy1018:
Recently I have been using LightGBM 2.1.0 c_api library in production. And It was working fine. And now we have decided to update the version to 3.1.0 and check the performance gain. But we faced an issue where we were getting different results on different runs of predict function for single row (LGBM_BoosterPredictForMatSingleRowFast) method for the same input. But this problem occurs only when this function is being called in multithreaded environment. That means we can have different threads calling this function for different inputs. Could this mean that the predict function is no longer a thread safe?

Sample Code:
Config Setup: LGBM_BoosterPredictForMatSingleRowFastInit(m.model, C.C_API_PREDICT_NORMAL, C.int(0), C.int(0), 
                                                                                                              C.C_API_DTYPE_FLOAT64, C.int(len(m.featureNames)), 
                                                                                                              C.CString("num_threads=1"), &m.config)

Predict Call:   LGBM_BoosterPredictForMatSingleRowFast(m.config, cinputArray, &outLen, cresult)

The above methods are being called using cgo in golang

guolinke:
cc @AlbertoEAF

AlbertoEAF:
Hello @tarunreddy1018 yes this is not thread safe (nor the remaining predict methods). In my case we had to protect the predict calls with a mutex+lock scheme.

LGBM_BoosterPredictForMatSingleRowFast differs from the LGBM_BoosterPredictForMatSingleRow in the sense that it keeps your "prediction settings" and booster in your FastConfig (`m.config`) which is only used "read-only" by `LGBM_BoosterPredictForMatSingleRowFast`. Besides that, it internally uses the same machinery as the non-fast version:

```cpp
int LGBM_BoosterPredictForMatSingleRowFast(FastConfigHandle fastConfig_handle,
                                           const void* data,
                                           int64_t* out_len,
                                           double* out_result) {
  API_BEGIN();
  FastConfig *fastConfig = reinterpret_cast<FastConfig*>(fastConfig_handle);
  // Single row in row-major format:
  auto get_row_fun = RowPairFunctionFromDenseMatric(data, 1, fastConfig->ncol, fastConfig->data_type, 1);
  fastConfig->booster->PredictSingleRow(fastConfig->predict_type, fastConfig->ncol,
                                        get_row_fun, fastConfig->config,
[...]

PR #3771, opened 2021-01-16
Fix thread-safety in C API's PredictSingleRow

AlbertoEAF:
The single row predictor methods in the C APi are not thread safe as reported in https://github.com/microsoft/LightGBM/issues/3751 and https://github.com/microsoft/LightGBM/issues/3675. This fixes both reports. Thank you https://github.com/tarunreddy1018 for the detailed reports!

By using a unique lock instead of the shared lock the timings are very similar, but predictions are correct.

Remarks for future work:

- Even so, by designing a small C++ benchmark with a very simple LGBM model, more threads on a simple model are slower than the single-thread case. This is probably due to very small work units, the lock contention overhead increases.
- We should in the future benchmark with more complex models to see if supporting
threading on these calls is worth it in performance gains.
- If not, then we could choose to not to provide thread-safety and remove the locks altogether for maximal throughput (though people that didn't read the documentation carefully might be surprised that such methods were not thread-safe).
  - We should in that case use non-shared resources so there is no synchronization required and multiple calls can be issued in parallel with no lock contention.

See https://github.com/microsoft/LightGBM/issues/3751 for timings of the small benchmark and the benchmark gist https://gist.github.com/AlbertoEAF/5972db15a27c294bab65b97e1bc4c315

### DIFF
--- a/src/c_api.cpp
+++ b/src/c_api.cpp
@@ -391,7 +391,7 @@ class Booster {
       Log::Fatal("The number of features in data (%d) is not the same as it was in training data (%d).\n"\
                  "You can set ``predict_disable_shape_check=true`` to discard this error, but please be aware what you are doing.", ncol, boosting_->MaxFeatureIdx() + 1);
     }
-    SHARED_LOCK(mutex_)
+    UNIQUE_LOCK(mutex_)
     const auto& single_row_predictor = single_row_predictor_[predict_type];
     auto one_row = get_row_fun(0);
     auto pred_wrt_ptr = out_result;


AlbertoEAF:
Hello, can anyone merge this fix?

guolinke (approved):

guolinke:
Hi @AlbertoEAF 
Is that possible to expose single row related APIs in python/R package?

AlbertoEAF:
Hello @guolinke :)
It's totally possible, their API are very similar to the non-single row methods so it shouldn't be difficult at all. 

In python's case we could replace the input pandas dataframe by something lighter like a numpy array.

Issue #3571, opened 2020-11-16
How exactly LightGBM predictions are obtained?


maksymiuks:
Hi

In the beginning, I'd like to appreciate your work covering that package. I consider it a great tool. That's why I'm working on a dedicated R interface for tree ensemble models that allows calculating shap values fast using C++ code via Rcpp. LightGBM package is one in the scope of my interest. But for that, I need the information on how exactly the committee of trees is aggregated. From my inspection of the code, I have a hunch that the final prediction is a sum of prediction for all trees with some intercept subtracted. Am I correct? If so how to find that intercept because I wasn't able to in the model object. The goal for me is to acquire a plain sum of predictions for all trees.

Best Regards
Szymon Maksymiuk

guolinke:
except for mutli-class tasks, the prediction is the sum of prediction of each tree. 
For some tasks, like binary classification, there could be a transformation after sum, like sigmoid.
For multi-class, you need to sum the prediction by class first (trees are organized as tree[i * K + j], where i is iteration, j is class-id, and K is the number of class), and use softmax to get the probabilities for classes.

btrotta:
I'm not sure if I'm understanding your question correctly, but I think when you talk about the "intercept" you mean something like the baseline constant prediction? E.g. for a binary prediction problem where the training labels are 90% ones and 10% zeros, we would start with a constant prediction of 0.9 and then add trees to improve the accuracy. This is indeed how LightGBM works, and this constant value is added to the leaf values of the first tree. So if you use `Booster.save_model()` (https://lightgbm.readthedocs.io/en/latest/pythonapi/lightgbm.Booster.html?highlight=save%20model#lightgbm.Booster.save_model) the leaf values for the first tree include this baseline value. 
The relevant part of the C++ code is in `TrainOneIter` https://github.com/microsoft/LightGBM/blob/f38f118ce2fc6451b75363689fc06e011d69cf33/src/boosting/gbdt.cpp#L350 In the first iteration (when `gradients` and `hessians` are `nullptr`), it calls `BoostFromAverage` which calculates the constant initial prediction. Then it calculates the optimal tree (fitting to the error from the constant prediction), and later it calls `AddBias` to add the constant to the individual leaf values.

Issue #1969, opened 2019-01-25
LightGBM Predict Function returns the Logit Base value rather than row probabilities


sarahboufelja:
## Environment info

Operating System: Windows

CPU/GPU model:

C++/Python/R version: Python3

LightGBM version or commit hash: 2.2.2

## Error 

LightGBM predict method produces log odds Base Value (as indicated in the Shap Forces Graph) rather than probabilities and the output is exaclty the same for all rows. # Reproducible examples

## Predicting Probabilities 
y_proba = pd.DataFrame(lgbm_ult.predict(test_x_enc, raw_score=False, num_iteration=None, pred_leaf=False, pred_contrib=True))
![image](https://user-images.githubusercontent.com/8659897/51776283-dc35c380-20f8-11e9-9046-91f8c07bb40f.png)

![image](https://user-images.githubusercontent.com/8659897/51776314-fa9bbf00-20f8-11e9-85f1-42841f52745c.png)

guolinke:
if you want the probabilities, I think you should set `pred_contrib=False`

sarahboufelja:
Ok thank you guolinke. So when pred_contrib=True, the last columns corresponds to the "Base Estimate" rather than the element-wide prediction, right? I can process the probabilities myself by summing the Shap values and applying a sigmoid function?

maximemerabet:
@sarahboufelja Setting pred_contrib=True makes your predict return the shap values with the last item as the expected value (equivalent to shap_values + shap.TreeExplainer(tree_estimator).expected_value).

@slundberg Some inconsistency here in how expected_value is handled in LGBM vs shap (LGBM still includes the expected value in the shap array)

StrikerRUS:
@maximemerabet According to the @slundberg's answer here https://github.com/Microsoft/LightGBM/issues/1350#issuecomment-395903942, it's more likely that new features and updates will be released only in `shap` package, no update for LightGBM codebase.

maximemerabet:
@StrikerRUS Yup that makes sense. I'm more trying to point that lgbm's shap_values are not identical to shap's shap_values. I think some form of documentation around that would be preferred, or simply switch off the LGBM modules if it won't be maintained

slundberg:
I think having a working version in LightGBM is good to keep of course. But trying to keep the API's perfectly aligned is tricky, since it requires deprecating LightGBM API's at the moment.

StrikerRUS:
[...]

PR #3774, opened 2021-01-17
[dask] Support pred_contrib in Dask predict() methods (fixes #3713)

### LINKED ISSUE #3713: [dask] support 'pred_contrib' in predict()
## Summary

LightGBM's `Booster.predict()` can be use to get feature contributions, by passing `pred_contrib=True`. This should be supported for model objects from the Dask interface as well.

## Motivation

This is useful for inspecting the relative importance of features in a LightGBM model.

## Description

The check at https://github.com/microsoft/LightGBM/blob/aae4fe401b316e156bf1d4aa8bbc213a606aab5e/python-package/lightgbm/dask.py#L169 should also check whether `pred_contrib=True` has been passed, and return the full DataFrame of the result.

## References

* this change in a fork of `dask-lightgbm` shows what the relevant change might look like: https://github.com/h2oai/dask-lightgbm/commit/c19000b57c1239b2a6de65cf172163c8684ce7dc

jameslamb:
The `predict()` method on Dask model objects doesn't correctly handle `pred_contrib=True` today. It fails to return the full matrix of feature contributions.

Thanks to @pseudotensor for pointing out this bug (#3713).

This PR fixes that.

## Notes for Reviewers

 I found that the results of `predict(pred_contrib=True)` are different between the Dask interface and `sklearn` model objects, if `n_workers` in the Dask cluster is greater than 1. I observed that both feature contribution values and the "base value" in the pred_contrib output are often different. This is true for regression, binary classification, and multi-class classification. The differences are larger than what I think could be attributed to numeric precision issues.

@guolinke , should I expect that pred_contrib outputs are different between multi-machine training and single-machine training? I'm unsure if this is because of an issue in the Dask interface or if it's something that is LightGBM-wide.

## References

I found these conversations useful while working through this:

* https://github.com/microsoft/LightGBM/issues/1969
* https://github.com/microsoft/LightGBM/issues/3571
* https://github.com/slundberg/shap/issues/352


tests/python_package_test/test_dask.py
```diff
@@ -90,6 +91,57 @@ def test_classifier(output, centers, client, listen_port):
...
+        time_out=5,
+        local_listen_port=listen_port,
+        tree_learner='data'
+    )
+    dask_classifier = dask_classifier.fit(dX, dy, sample_weight=dw, client=client)
+    preds_with_contrib = dask_classifier.predict(dX, raw_score=True, pred_contrib=True).compute()
+
+    local_classifier = lightgbm.LGBMClassifier()
+    local_classifier.fit(X, y, sample_weight=w)
+    local_preds_with_contrib = local_classifier.predict(X, raw_score=True, pred_contrib=True)
```
StrikerRUS, line 109: Why do we need `raw_score=True` here? I believe `raw_score` and `pred_contrib` are mutually exclusive. Seems it is just luck that `if pred_contrib` is the last one check and this test is passed.
https://github.com/microsoft/LightGBM/blob/ed651e86729477e7d2699534d63a5cace87f7878/python-package/lightgbm/basic.py#L571-L577

jameslamb: oh ok, I'll remove it

jameslamb: removed in https://github.com/microsoft/LightGBM/pull/3774/commits/440a30bfe3a7f18ab989c61387b44e68cf3bd1e8

StrikerRUS:
@jameslamb 

> The `predict()` method on Dask model objects doesn't correctly handle `pred_contrib=True` today.

Just to clarify: are `raw_score=True` and `pred_leaf=True` fully supported now? If yes, I think we need tests, if no then feature requests.
https://github.com/microsoft/LightGBM/blob/d2c554540d86fa3ccd1013cf4f9263c654ebf74a/python-package/lightgbm/sklearn.py#L871-L880

python-package/lightgbm/dask.py
```diff
@@ -239,7 +239,7 @@ def _predict_part(part, model, proba, **kwargs):
         result = model.predict(data, **kwargs)
 
     if isinstance(part, pd.DataFrame):
-        if proba:
+        if proba or kwargs.get('pred_contrib', False):
```
StrikerRUS, line 242: I'd better rename `proba` to `pred_proba` and make other available in sklearn wrapper `pred_*` params ordinary function arguments to unify interfaces.

jameslamb: totally agree! I initially didn't do that because I was trying to keep the diffs small. But I'll do that on this PR.

jameslamb: ok I tried this in https://github.com/microsoft/LightGBM/pull/3774/commits/98a787860dbe4e10879e76cd79c9ad4a55e44d03, is that what you meant?

StrikerRUS: Yeah, thanks! Exactly what I meant. Love it!

tests/python_package_test/test_dask.py
```diff
@@ -97,6 +98,57 @@ def test_classifier(output, centers, client, listen_port):
...
+@pytest.mark.parametrize('output', data_output)
+@pytest.mark.parametrize('centers', data_centers)
+def test_classifier_pred_contrib(output, centers, client, listen_port):
+    X, y, w, dX, dy, dw = _create_data('classification', output=output, centers=centers)
+
+    dask_classifier = dlgbm.DaskLGBMClassifier(
+        time_out=5,
+        local_listen_port=listen_port,
+        tree_learner='data'
+    )
```
StrikerRUS, line 110: Add `n_estimators=10` and `num_leaves=10`?
#3786.

tests/python_package_test/test_dask.py
```diff
@@ -97,6 +98,57 @@ def test_classifier(output, centers, client, listen_port):
...
+        preds_with_contrib = np.array(preds_with_contrib.todense())
+
+    # shape depends on whether it is binary or multiclass classification
+    num_classes = len(centers)
+    if num_classes == 2:
+        expected_num_cols = dX.shape[1] + 1
+    else:
+        expected_num_cols = (dX.shape[1] + 1) * num_classes
+
+    if isinstance(dX, dask.dataframe.core.DataFrame):
```
StrikerRUS, line 128: I might be wrong, but according to docs and sources, we can use it without `core` part to not depend on inner implementation.
https://docs.dask.org/en/latest/dataframe-api.html#dask.dataframe.DataFrame
https://github.com/dask/dask/blob/72304a94c98ace592f01df91e3d9e89febda307c/dask/dataframe/__init__.py#L3
```suggestion
    if isinstance(dX, dask.dataframe.DataFrame):
```

tests/python_package_test/test_dask.py
```diff
@@ -170,6 +222,32 @@ def test_regressor(output, client, listen_port):
...
 
+@pytest.mark.parametrize('output', data_output)
+def test_regressor_pred_contrib(output, client, listen_port):
+    X, y, w, dX, dy, dw = _create_data('regression', output=output)
+
+    dask_regressor = dlgbm.DaskLGBMRegressor(
+        time_out=5,
+        local_listen_port=listen_port,
+        tree_learner='data'
+    )
```
StrikerRUS, line 233: Add `n_estimators=10` and `num_leaves=10`?
#3786.

StrikerRUS (commented):
Thanks for so actively improving Dask module!
Please check my minor comments below.

tests/python_package_test/test_dask.py
jameslamb: oh good idea

tests/python_package_test/test_dask.py
jameslamb: ooo that's a good idea, let me try that

python-package/lightgbm/dask.py
```diff
@@ -228,47 +228,67 @@ def _train(client, data, label, params, model_factory, weight=None, **kwargs):
     return results[0]
 
 
-def _predict_part(part, model, proba, **kwargs):
+def _predict_part(part, model, pred_proba, pred_leaf, pred_contrib, **kwargs):
```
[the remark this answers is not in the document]
jameslamb: added in https://github.com/microsoft/LightGBM/pull/3774/commits/64285895ec11c4a23af838c7cdfe62889a78b1a5. I'll answer in a separate comment

tests/python_package_test/test_dask.py
jameslamb: added in https://github.com/microsoft/LightGBM/pull/3774/commits/64285895ec11c4a23af838c7cdfe62889a78b1a5

tests/python_package_test/test_dask.py
jameslamb: added in https://github.com/microsoft/LightGBM/pull/3774/commits/64285895ec11c4a23af838c7cdfe62889a78b1a5 and it worked ok

tests/python_package_test/test_dask.py
jameslamb: added in https://github.com/microsoft/LightGBM/pull/3774/commits/64285895ec11c4a23af838c7cdfe62889a78b1a5

jameslamb:
> Just to clarify: are raw_score=True and pred_leaf=True fully supported now? If yes, I think we need tests, if no then feature requests.

I spent about an hour today trying to get the tests for `raw_score` and `pred_leaf` working. Both are only somewhat supported (for some combinations of task and input type). I faced several errors related to what I think is one root cause: when using `map_blocks()` or `map_partitions()`, you have to provide some metadata to help Dask understand the shape of the function call result. This is described in the `meta` param of https://docs.dask.org/en/latest/dataframe-api.html#dask.dataframe.DataFrame.map_partitions, for example.

I'll write up feature requests and link them here and in #2302. I'm not sure if it's that "raw_score and pred_leaf are not supported", or more if I was just making mistakes in the tests.

So as of this PR, those parameters will be understood by `predict()`, but whether or not it succeeds won't be guaranteed.

StrikerRUS:
> UPDATE: added #3792 and #3793

Thank you so much!

tests/python_package_test/test_dask.py
```diff
@@ -171,6 +223,32 @@ def test_regressor(output, client, listen_port):
     assert_eq(y, p2, rtol=1., atol=50.)
 
 
+@pytest.mark.parametrize('output', data_output)
+def test_regressor_pred_contrib(output, client, listen_port):
+    X, y, w, dX, dy, dw = _create_data('regression', output=output)
+
+    dask_regressor = dlgbm.DaskLGBMRegressor(
```
StrikerRUS, line 230: Add `n_estimators=10` and `num_leaves=10`?
#3786.

tests/python_package_test/test_dask.py
```diff
@@ -97,6 +98,57 @@ def test_classifier(output, centers, client, listen_port):
...
+        preds_with_contrib = np.array(preds_with_contrib.todense())
+
+    # shape depends on whether it is binary or multiclass classification
+    num_classes = len(centers)
+    if num_classes == 2:
+        expected_num_cols = dX.shape[1] + 1
+    else:
+        expected_num_cols = (dX.shape[1] + 1) * num_classes
+
+    if isinstance(dX, dask.dataframe.DataFrame):
```
StrikerRUS, line 128: Why don't use already imported `dd` module but import `dask` separately?
`dd.DataFrame`

StrikerRUS (changes requested):
Seems there were some merge conflicts because some my previous comments are not addressed but you said they were.

jameslamb:
> Seems there were some merge conflicts because some my previous comments are not addressed but you said they were.

so weird! Yeah, maybe a bad resolution of a merge conflict, sorry.

tests/python_package_test/test_dask.py
jameslamb: we definitely should. I'll make that change

tests/python_package_test/test_dask.py
```diff
@@ -97,6 +98,57 @@ def test_classifier(output, centers, client, listen_port):
...
+
+    local_classifier = lightgbm.LGBMClassifier()
+    local_classifier.fit(X, y, sample_weight=w)
+    local_preds_with_contrib = local_classifier.predict(X, pred_contrib=True)
+
+    if output == 'scipy_csr_matrix':
+        preds_with_contrib = np.array(preds_with_contrib.todense())
+
+    # shape depends on whether it is binary or multiclass classification
+    num_classes = len(centers)
```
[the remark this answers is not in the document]
jameslamb: Those features are already supported, because `DaskLGBMClassifier` inherits from `lightgbm.sklearn.LGBMClassifier` and `DaskLGBMRegressor` inherits from `lightgbm.sklearn.LGBMRegressor`. It even has a `.booster_`, so you don't have to use `to_local()` if you just want the booster directly.

image

I can change this test to use `.n_classes_`, that's fine.

tests/python_package_test/test_dask.py
jameslamb: yep! sorry about the bad merge conflict resolution

python-package/lightgbm/dask.py
```diff
@@ -257,47 +257,81 @@ def _train(client, data, label, params, model_factory, weight=None, **kwargs):
...
 
-def _predict(model, data, proba=False, dtype=np.float32, **kwargs):
+def _predict(model, data, raw_score=False, pred_proba=False, pred_leaf=False, pred_contrib=False,
+             dtype=np.float32, **kwargs):
     """Inner predict routine.
 
     Parameters
     ----------
     model :
     data : dask array of shape = [n_samples, n_features]
```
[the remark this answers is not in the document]
jameslamb: `dask.dataframe._Frame` is the superclass for both `dask.dataframe.DataFrame` and `dask.dataframe.Series`: https://github.com/dask/dask/blob/30c9a3a34240e08cccd7e38b115141661ca198cf/dask/dataframe/core.py#L272.

Can this be updated in another PR? It's not related to the changes in this PR and not code that I touched here. It seems like a better fit for the PR to add type hints to the Dask module (#3756 ), which I'm waiting until #3708 is done to do.

tests/python_package_test/test_dask.py
jameslamb: ok I just added this again (was lost because of a bad merge conflict resolution, sorry)

tests/python_package_test/test_dask.py
jameslamb: ok added this back again (now as `dd.DataFrame`)

tests/python_package_test/test_dask.py
jameslamb: added back again

tests/python_package_test/test_dask.py
```diff
@@ -97,6 +98,57 @@ def test_classifier(output, centers, client, listen_port):
     assert_eq(p1_proba, p2_proba, atol=0.3)
 
 
+@pytest.mark.parametrize('output', data_output)
+@pytest.mark.parametrize('centers', data_centers)
+def test_classifier_pred_contrib(output, centers, client, listen_port):
+    X, y, w, dX, dy, dw = _create_data('classification', output=output, centers=centers)
+
+    dask_classifier = dlgbm.DaskLGBMClassifier(
```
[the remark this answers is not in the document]
jameslamb: added back again

python-package/lightgbm/dask.py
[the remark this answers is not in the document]
ffineis: @jameslamb can we merge 3708?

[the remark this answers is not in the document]
StrikerRUS: > Can this be updated in another PR?
Sure!

tests/python_package_test/test_dask.py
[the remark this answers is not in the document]
StrikerRUS: Wow, awesome indeed! Thanks for clarification!

tests/python_package_test/test_dask.py
```diff
@@ -97,6 +97,57 @@ def test_classifier(output, centers, client, listen_port):
...
+    else:
+        expected_num_cols = (num_features + 1) * num_classes
+
+    # * shape depends on whether it is binary or multiclass classification
+    # * matrix for binary classification is of the form [feature_contrib, base_value],
+    #   for multi-class it's [feat_contrib_class1, base_value_class1, feat_contrib_class2, base_value_class2, etc.]
+    # * contrib outputs for distributed training are different than from local training, so we can just test
+    #   that the output has the right shape and base values are in the right position
+    num_features = dX.shape[1]
+    num_classes = len(centers)
```
StrikerRUS, line 139: Can we do the following?
```suggestion
    num_features = dask_classifier.n_features_
    num_classes = dask_classifier.n_classes_
    if num_classes == 2:
        expected_num_cols = num_features + 1
    else:
        expected_num_cols = (num_features + 1) * num_classes

    # * shape depends on whether it is binary or multiclass classification
    # * matrix for binary classification is of the form [feature_contrib, base_value],
    #   for multi-class it's [feat_contrib_class1, base_value_class1, feat_contrib_class2, base_value_class2, etc.]
    # * contrib outputs for distributed training are different than from local training, so we can just test
    #   that the output has the right shape and base values are in the right position
```

jameslamb: sure, that works too

StrikerRUS (approved):
Thanks, LGTM!

I'm afraid merging this PR will cause conflicts with #3708, so I'm not touching anything.

jameslamb:
now that #3708 has been merged, I'll fix merge conflicts here and then merge this

StrikerRUS:
@jameslamb Can we remove this strange branch?

image

jameslamb:
yep definitely! I just deleted it, sorry

StrikerRUS:
@jameslamb 
Ah, it is here again! 😄 

Can I remove it?

image

jameslamb:
WHAT I'm so confused. Yes please remove it, I'm sorry. maybe my remotes are set up wrong on some local clone.

StrikerRUS:
No problem, thanks!
> maybe my remotes are set up wrong on some local clone.

Perhaps... It is 2281 commits behind `master`, should be VERY old setup. 🙂

PR #3699, opened 2020-12-31
[R-package] enable use of trees with linear models at leaves (fixes #3319)

jameslamb:
This PR closes #3319, adding support for the new "leaves with linear models" feature @btrotta added in #3299.

Thankfully, the concerns from https://github.com/microsoft/LightGBM/issues/3319#issue-681647268 about the order in constructing the dataset weren't an issue! `lgb.train()` in the R package calls `update_params()` on the Dataset object and constructs it (https://github.com/microsoft/LightGBM/blob/967b45c68665488ac872f77848f516d84a70726c/R-package/R/lgb.train.R#L146-L147), so it will respect `linear_trees = true`.  If you've already called `$construct()` on the dataset before passing it in to `lgb.train()`, you'll get an informative error telling you that you can't use an already-constructed Dataset with `linear_trees = false` in training with `linear_trees = true`.

The unit tests I added for `lgb.train()` were based on those added in https://github.com/microsoft/LightGBM/pull/3299/files#diff-263bfdd4e44ba860ef61b458adfab0d5f391766a27a3fa7e9f221afc5f1e5bad, except `refit()` because the R package does not currently support `refit()` (#2369).

PR #3783, opened 2021-01-18
[R-package][tests] remove unused lines in tests for the speedup

StrikerRUS:
Validation results are not asserts to anything in these tests. So I believe we can remove `valids` param for clearness and to not waste time computing validation scores.

Sorry, I should have mentioned this during review of #3699.


jameslamb (approved):
good idea, thanks

PR #3785, opened 2021-01-18
[ci][docs] set limit for max requests per second to one host in linkchecker

StrikerRUS:
Try to workaround `Valid: 429 too many requests` to GitHub.
This is the most frequent error I see.

https://github.com/linkchecker/linkchecker/issues/193
https://github.com/linkchecker/linkchecker/blob/7db035d1bc124d531370ac25179af3e7a08a40ff/config/linkcheckerrc#L170-L171

### DIFF
--- a/docs/.linkcheckerrc
+++ b/docs/.linkcheckerrc
@@ -1,5 +1,5 @@
 [checking]
-maxrequestspersecond=3
+maxrequestspersecond=1
 recursionlevel=1
 anchors=1
 sslverify=0
--- a/docs/.linkcheckerrc
+++ b/docs/.linkcheckerrc
@@ -1,4 +1,5 @@
 [checking]
+maxrequestspersecond=3
 recursionlevel=1
 anchors=1
 sslverify=0


StrikerRUS:
In 10+ consecutive runs of `check-docs` test I didn't see `429` error.

jameslamb (approved):
Awesome change, thank you! Good idea to just slow linkchecker down, I didn't even think of that.

The `check-docs` task is so fast anyway, relative to our other tasks, that I think it's ok to slow it down like this. Let's try this and see how it goes. If we see another 429 show up, I'd even be ok with this being 3 seconds between links. If we did that, the total test time for `check docs` would still probably only 5-6 minutes, faster than most other

StrikerRUS:
Yeah, seems that this change is simply required given that many "serious" sites have DDoS protection now.

> I'd even be ok with this being 3 seconds between links.

Unfortunately, we cannot set `maxrequestspersecond` to float value (<1). Probably some another param will help. But let's think about it when/if needed 🙂 .

StrikerRUS:
OK, this limit didn't completely fixed the problem.
I see in `master`:
```
Warning    [url-rate-limited] Rate limited (Retry-After: 60)
Result     Valid: 429 too many requests
```
GitHub again.

Looking though the rest linkchecker params I believe the only option left that can help is 
```
# number of threads
#threads=10
```

https://github.com/linkchecker/linkchecker/blob/7db035d1bc124d531370ac25179af3e7a08a40ff/config/linkcheckerrc#L126-L127

Because I suspect that `maxrequestspersecond` should be read as `maxrequestspersecondPERTHREAD` actually, though I'm totally unsure and can be wrong.

Another option which I'd prefer is simply add `url-rate-limited` to `ignorewarnings` list:
https://github.com/microsoft/LightGBM/blob/787780851f911fb96fa2895b05ac792291954cf5/docs/.linkcheckerrc#L11

Or simply continue re-runing jobs (hopefully with less frequency) 🙂 .

jameslamb:
What do you think about bumping the parameter changed in this PR to 3 seconds? https://github.com/microsoft/LightGBM/pull/3785#pullrequestreview-570776482

StrikerRUS:
> What do you think about bumping the parameter changed in this PR to 3 seconds?

TBH, I don't know what to say. I can't find GitHub limits. I've found only GitHub **REST API** limits:
> For API requests using Basic Authentication or OAuth, you can make up to 5,000 requests per hour.
    https://docs.github.com/en/rest/overview/resources-in-the-rest-api#rate-limiting

It is `1.3(8)` requests per second.

Also, I've found some other examples for limits of famous site APIs.

Twitter: https://developer.twitter.com/en/docs/twitter-api/rate-limits#table

Google Analytics:

> 10 queries per second (QPS) per IP address.
   https://developers.google.com/analytics/devguides/config/mgmt/v3/limits-quotas

Slack

> Rate limits do apply to posting messages or other write events to the Real Time Messaging websocket. Please limit writes to 1 per second.
  https://api.slack.com/docs/rate-limits#rate-limits__rtm-apis__posting-messages

Trello:
> There is a limit of 300 requests per 10 seconds for each API key and 100 requests per 10 second interval for each token.

StrikerRUS:
Do you think that 1 RPS is too slow?

jameslamb:
Looking at the logs in https://github.com/microsoft/LightGBM/runs/1728534013, I see that we make 570 requests (if I understand that "links" and "URLs" are different things for linkchecker).

> That's it. 274 links in 296 URLs checked. 0 warnings found. 0 errors found.

So single-threaded 1 request per second would be 9.5 minutes. Add about a minute of time for `setup.sh`, that would make the check-docs job 10-11 minutes. I actually think that would be ok in exchange for not having to regularly manually re-run the job. That's shorter than our longest CI jobs.

I have another idea for this one...we could also change `check-docs` from something required on every PR to something that runs on a schedule, like every 6 hours. https://github.com/saturncloud/prefect-saturn/blob/ded98bf24db7da7bb1f2b59e7ad31e4a149b73a3/.github/workflows/main.yml#L12-L14

That way, we'd still regularly catch broken link issues, but this check wouldn't block PRs and it would have only a small effect on our CI capacity during times where multiple people are pushing commits. In my experience, most PRs in this project do not introduce new links.

StrikerRUS:
>I see that we make 570 requests (if I understand that "links" and "URLs" are different things for linkchecker).

I assume we don't need to sum these numbers and we make 274 requests overall.
`296 - 274 = 22` are ignored by our config.
However, this is just my guesses.

> I actually think that would be ok in exchange for not having to regularly manually re-run the job.

Parallel jobs from different PRs also may contribute to RPS we produce in case workers (agents, or how they are called in GitHub Actions...) share the same IP address.

> we could also change `check-docs` from something required on every PR to something that runs on a schedule, like every 6 hours.

I'm OK with CRON jobs. But only for `linkchecker` execution, not for the whole `check-docs` job. We check other important things there. But I don't think that anyone will check results of new CRON job on a regular basis.

jameslamb:
ooo ok! 274 is a lot lower. It's a good point you make about multiple jobs on different PRs all contributing.

I like the idea of splitting just `linkchecker` into a cron job on GitHub Actions. I think it's ok that it won't be checked regularly. I'm ok with the tradeoff of higher risk that a broken link exists in the docs for a few days or even weeks, in exchange for not needing to manually re-run CI jobs as much on pull requests and on `master`. 

And since that job would not block PRs, we could also slow it down to 1 thread, 1 request per second.

What do you think? I'd be happy to open a PR for this if you think it's a good idea.

StrikerRUS:
I agree with you.

> I'd be happy to open a PR for this if you think it's a good idea.

Thank you very much!

Please also add `on: workflow_dispatch` trigger to create a possibility for maintainers to run that workflow with `linkchecker` manually by clicking a button. It will be useful in case someone come to check a status and would like to rerun the workflow right now to make sure error is or is not temporary.
https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/

PR #3772, opened 2021-01-16
[docs] expand documentation on 'group' for ranking task

jameslamb:
Based on the discussion in https://github.com/microsoft/LightGBM/pull/3708#discussion_r558570787, this PR proposes adding more details to the explanation of `group` in the Python and R packages.

I think that the current description doesn't give enough details for someone unfamiliar with LightGBM to get started. I know that I personally really struggled to understand what "group" should look like the first time I did learning-to-rank with LightGBM.

PR #3787, opened 2021-01-19
[docs] fix current RTD failures

StrikerRUS:
After merging #3772 RTD builds started to fail with very strange error during creating PDF.
```
[143
! pdfTeX error (ext4): \pdfendlink ended up in different nesting level than \pd
fstartlink.
\AtBegShi@Output ...ipout \box \AtBeginShipoutBox 
                                                  \fi \fi 
l.13061 \item[{Return type}]
                             \leavevmode
!  ==> Fatal error occurred, no output PDF file produced!
```

image

I guess we hit the following issue: https://tex.stackexchange.com/a/424262.
I guess that the change of docstrings caused that some reference has been broken up by new page break position.

By trial and errors I found that either significant reduction of new `group` param description can change page breaks and solve th issue or some small addition to the current description. So I added `..., records 31-70 are in the third group ...`.

Also, I broke extremely huge docstring lines so that there is no terrible horizontal scrolling bar at GitHub for corresponding files.

### DIFF
--- a/docs/Parameters.rst
+++ b/docs/Parameters.rst
@@ -1243,7 +1243,7 @@ For wrapper libraries like in Python and R, this information can also be provide
 
 ::
 
-   [27, 18, 67, ...]
+    [27, 18, 67, ...]
 
 For example, if you have a 112-document dataset with ``group = [27, 18, 67]``, that means that you have 3 groups, where the first 27 records are in the first group, records 28-45 are in the second group, and records 46-112 are in the third group.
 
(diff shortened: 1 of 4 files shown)


jameslamb (approved):
oh weird, sorry 😬

StrikerRUS:
@jameslamb Totally not your fault! 🙂 We do not check PDF creation. This error is unpredictable.

StrikerRUS:
Well, can confirm that `master` is fixed now:

image

PR #3788, opened 2021-01-19
[dask] fix Dask import order

jameslamb:
This PR has two small cleanup actions for the Dask module.

1. moves `scipy` import above relative imports like `from .basic`, matching the ordering we use in other modules
2. replaces unused variables in tests with `_`

I found both of these small things with `pylint`.


StrikerRUS (approved):
LGTM!

Issue #3714, opened 2021-01-03
[dask] allow customization of num_threads


jameslamb:
## Summary

LightGBM training with Dask currently ignores the `num_threads` setting in user-provided parameters, and overwrites it with the number of CPU cores on each worker contributing to training.

This behavior is a good default, but it should be possible to override it.

## Motivation

There might be cases where it's desirable not to use as many threads as cores on each worker during training, and that should be possible.

## References

* See https://github.com/microsoft/LightGBM/blob/aae4fe401b316e156bf1d4aa8bbc213a606aab5e/python-package/lightgbm/dask.py#L144 for a proposed fix
* The current behavior comes from https://github.com/dask/dask-lightgbm

This behavior was recommended in https://github.com/microsoft/LightGBM/pull/3515#issuecomment-752709132.

jameslamb:
Closing this in favor of being in #2302 with other features. Please leave a comment here if you'd like to work on it.

jameslamb:
I'm going to open this because I'm actively working on it.

<hr>

@guolinke could tell me more about this note in the parameter docs for `num_threads`?

https://lightgbm.readthedocs.io/en/latest/Parameters.html#num_threads

> for parallel learning, do not use all CPU cores because this will cause poor performance for the network communication

I'm wondering if this means that I should change the default behavior of the Dask module to use `n_cores - 1` threads on each worker for this parameter.

StrikerRUS:
@jameslamb 

> I'm going to open this because I'm actively working on it.

Just want to make sure you're aware of that `nthreads` can/should be used among additional params for `predict` methods as well as for model training.

Here are two examples where users are not happy with the default values:

https://github.com/microsoft/LightGBM/issues/1534#issuecomment-406485465
https://github.com/microsoft/LightGBM/issues/2225#issuecomment-501220332

jameslamb:
I'm going to close this again for now, currently focusing on other Dask items. This can be done after 3.2.0 (#3872 ).

StrikerRUS:
Related: https://github.com/dmlc/xgboost/pull/7337.

StrikerRUS:
The following code of `joblib`'s [`DaskDistributedBackend`](https://github.com/joblib/joblib/blob/master/joblib/_dask.py) and [`MultiprocessingBackend`](https://github.com/joblib/joblib/blob/7742f5882273889f7aaf1d483a8a1c72a97d57e3/joblib/_parallel_backends.py#L433-L443) might be useful.

PR #3789, opened 2021-01-19
[dask] allow parameter aliases for local_listen_port, num_threads, tree_learner (fixes #3671)

### LINKED ISSUE #3671: [dask] support parameter aliases in Dask API
## Summary

LightGBM supports aliases for most parameter. For example, you can use `bagging_fraction` or `sub_row` to say "% of rows sampled during bagging".

The new dask API added in #3515 should be changed to support such aliases for the parameters it references.

See these comments for details on a few parameters that need to be updated to use aliases.

* https://github.com/microsoft/LightGBM/pull/3515#discussion_r541797986
* https://github.com/microsoft/LightGBM/pull/3515#discussion_r541798881
* https://github.com/microsoft/LightGBM/pull/3515#discussion_r541798968

A fix should re-use `_ConfigAliases`, and update it with any missing parameters (https://github.com/microsoft/LightGBM/blob/d90a16d520fa12d84ecd983fba323727348616a4/python-package/lightgbm/basic.py#L200).

## Motivation

This would make the Dask API more consistent with the non-Dask API in the Python package. The goal is for these to have interfaces that are as simiilar as possible.

jameslamb:
This PR closes #3671. It updates the code in `lightgbm.dask` so that it respects [parameter aliases](https://lightgbm.readthedocs.io/en/latest/Parameters.html).

So, for example, as of this PR you can use `tree_learner`, `tree`, `tree_type`, or `tree_learner_type` to refer to the `tree_learner` parameter that controls which [type of distributed training](https://lightgbm.readthedocs.io/en/latest/Parallel-Learning-Guide.html#choose-appropriate-parallel-algorithm) is performed.


StrikerRUS:
Hmmm,  itdoesn't look like GPU error is related to these Dask changes:
```
_________________ TestEngine.test_missing_value_handle_more_na _________________

self = <python_package_test.test_engine.TestEngine testMethod=test_missing_value_handle_more_na>

    def test_missing_value_handle_more_na(self):
        X_train = np.ones((100, 1))
        y_train = np.ones(100)
        trues = random.sample(range(100), 80)
        for idx in trues:
            X_train[idx, 0] = np.nan
            y_train[idx] = 0
        lgb_train = lgb.Dataset(X_train, y_train)
        lgb_eval = lgb.Dataset(X_train, y_train)
    
        params = {
            'metric': 'l2',
            'verbose': -1,
            'boost_from_average': False
        }
        evals_result = {}
        gbm = lgb.train(params, lgb_train,
                        num_boost_round=20,
                        valid_sets=lgb_eval,
                        verbose_eval=False,
                        evals_result=evals_result)
        ret = mean_squared_error(y_train, gbm.predict(X_train))
>       self.assertLess(ret, 0.005)
E       AssertionError: 0.008207460416181563 not less than 0.005
```

python-package/lightgbm/dask.py
```diff
@@ -197,21 +200,47 @@ def _train(client, data, label, params, model_factory, weight=None, **kwargs):
...
+    allowed_tree_learners = {
+        'data',
+        'data_parallel',
+        'feature',
+        'feature_parallel',
+        'voting',
+        'voting_parallel'
+    }
+    if tree_learner is None:
+        logger.warning('Parameter tree_learner not set. Using "data" as default"')
```
StrikerRUS, line 218:
```suggestion
        logger.warning('Parameter tree_learner not set. Using "data" as default')
```

python-package/lightgbm/dask.py
```diff
@@ -197,21 +200,47 @@ def _train(client, data, label, params, model_factory, weight=None, **kwargs):
...
-    local_listen_port = params.get('local_listen_port', 12400)
     worker_address_to_port = _find_ports_for_workers(
         client=client,
         worker_addresses=worker_map.keys(),
         local_listen_port=local_listen_port
     )
 
+    # num_threads is set below, so remove it and all aliases of it from params
+    for num_thread_alias in _ConfigAliases.get('num_threads'):
+        _ = params.pop(num_thread_alias, None)
```
StrikerRUS, line 242:
```suggestion
        params.pop(num_thread_alias, None)
```

python-package/lightgbm/dask.py
```diff
@@ -197,21 +200,47 @@ def _train(client, data, label, params, model_factory, weight=None, **kwargs):
...
     # on the same machine, so this needs to ensure that each one gets its
     # own port
-    local_listen_port = params.get('local_listen_port', 12400)
     worker_address_to_port = _find_ports_for_workers(
         client=client,
         worker_addresses=worker_map.keys(),
         local_listen_port=local_listen_port
     )
 
+    # num_threads is set below, so remove it and all aliases of it from params
```
StrikerRUS, line 240: Why don't we respect user value for `num_threads` parameter?

StrikerRUS (changes requested):
Thanks a lot for implementing my feature request! 🙂 

Could please do the same with aliases here?
https://github.com/microsoft/LightGBM/blob/4007b34fa83043cc30c2ec3c8cf2c7b059f07fec/python-package/lightgbm/basic.py#L2078-L2081
I think it OK to use the same PR for implementing network aliases in different placed of codebase. But feel free to disagree with me and say that it requires a separate PR or/and feature request.

python-package/lightgbm/dask.py
jameslamb: to be clear, this is existing behavior and this PR isn't changing it. There is a feature request to relax this a bit and make it the default behavior, but allow people to control `num_threads`: #3714. I'm still unsure about whether that's a good idea.

Ok given that, here's the reasoning for the current behavior...

When you set up Dask, each worker is a process and you tell Dask how many threads that worker should be able to use. Setting `num_threads` to `worker.nthreads` ensures that LightGBM respects that. It avoids the situation where, for example, you have 2 worker processes on the same 4-core machine both running with `num_threads = 4`, which would result in LightGBM using 8 threads.

XGBoost does this overriding and does not allow such customization, for example: https://github.com/dmlc/xgboost/blob/d6d72de339e3172143b1923a780c1674f4248fcd/python-package/xgboost/dask.py#L802-L809

It's also important that we set this worker-by-worker instead of accepting a single integer, since it's possible in Dask to run worker processes on a cluster of machines that are not all the same size.

When I pick up #3714 we can have more of a discussion about this and can involve others from the Dask community.

My best guess is that the solution we'll end up is:

* if you pass `num_threads = -1` or don't include `num_threads`, you get this behavior of `num_threads = worker.nthreads`
* if you pass `num_threads` with any value other than -1, we respect that but raise a warning explaining the issues that might cause in distributed training

jameslamb:
> Could please do the same with aliases here?

I'm happy to do this soon, but I'd like to do it as a separate PR. I have a preference for keeping the dask PRs small and quick, to prevent too much pain for @ffineis on #3708, and to make sure that the Dask interface keeps progressing towards something we're comfortable releasing with 3.2.0. You only get one chance to release code as "new" :grinning:

python-package/lightgbm/dask.py
StrikerRUS: OK, got it! Thank you for the super detailed response as always! 🙂 

Sorry, I didn't notice the existence of #3714.

StrikerRUS:
Looks like Dask tests are quite flaky:
```
/home/AzDevOps_azpcontainer/miniconda/envs/test-env/lib/python3.8/asyncio/tasks.py:490: TimeoutError
```

> but I'd like to do it as a separate PR.

Sure, thanks! 👍

jameslamb:
> Looks like Dask tests are quite flaky:

I'm realizing that too :tired_face: . @ffineis faced the same thing on #3708, and I think he's come up with a good solution for now on that PR. Also so my link and comments in https://github.com/microsoft/LightGBM/pull/3708#issuecomment-763365845...I think we might get better reliability from switching to the `@gen_cluster` fixture from Dask `distributed`. I'm planning to test that once #3708 is merged.

StrikerRUS (approved):
Great, thanks!

Issue #3732, opened 2021-01-05
[python] [ci] Refactor tests to use pytest style


jameslamb:
## Summary

The tests on the Python package currently use two different styles.

[The `dask` tests](https://github.com/microsoft/LightGBM/blob/3fd48864da10baa6b96abc1bde219e4956e4f0a7/tests/python_package_test/test_dask.py) for example, use the pytest style, where each test case is a function and conditions are checked by `assert`-ing

```python
def test_classifier(output, centers, client, listen_port):
    ....
    assert_eq(s1, s2)
    assert_eq(p1, p2)
```

While [the 'engine' tests](https://github.com/microsoft/LightGBM/blob/3fd48864da10baa6b96abc1bde219e4956e4f0a7/tests/python_package_test/test_engine.py) use the traditional unittest style, where each test case is a method on a class that inherits from `unittest.TestCase` and conditions are checked with unittest builtin methods.

```python
class TestEngine(unittest.TestCase):
    def test_binary(self):
        ...
        self.assertLess(ret, 0.14)
        self.assertEqual(len(evals_result['valid_0']['binary_logloss']), 50)
```

To close this issue, change all of the tests to the `pytest` style. You'll know you've been successful if none of the test files need to `import unittest`.

## Motivation

This consistency will make contributing on LightGBM easier. This is especially important in the Python project, which is often used to indirectly test changes at the C++ side (for example, #3299 ).

## Description

This change should be applied for all test modules in `tests/python_package_test`. Please avoid changing any of the test logic and just focus on style.

If you get stuck, `@` me here and I can help.

StrikerRUS:
- [ ] `test_basic.py`
- [ ] `test_consistency.py`
- [ ] `test_dask.py`
- [ ] `test_dual.py`
- [ ] `test_engine.py`
- [ ] `test_plotting.py`
- [ ] `test_sklearn.py`

PR #3798, opened 2021-01-20
TST Migrates test_consistency.py to pytest

thomasjpfan:
Towards https://github.com/microsoft/LightGBM/issues/3732


jameslamb (approved):
This looks great, thank you! Thanks for introducing [`pytest.importorskip`](https://docs.pytest.org/en/latest/reference.html#pytest-importorskip-ref).

I reviewed with https://www.textcompare.org/index.html?id=6000f45b143e0d0017c81ac3

PR #3799, opened 2021-01-20
[docs] fix min CMake version in docs for macOS

StrikerRUS:
No exceptions for threadless version.

https://github.com/microsoft/LightGBM/blob/d107872a5342442420edaa1d1849901f865f2331/CMakeLists.txt#L47-L51


jameslamb (approved):
looks good, thanks

PR #3803, opened 2021-01-21
[docs] bump min sphinx_rtd_theme version

StrikerRUS:
Needed to fix (probably [this one](https://github.com/readthedocs/sphinx_rtd_theme/pull/941)?) the following problem with lists inside other elements where colon is missed in version < 0.5 and words are joined.

image

Unfortunately, output from `breathe` still is not fixed. But there tooltip appeared on hover can help to split words.

image


jameslamb (approved):
oh great, thank you

Issue #3684, opened 2020-12-26
reduce sdist size


StrikerRUS:
Current `sdist` size in `master` is more than 3 Mb. This is due to adding new dependency `eigen` in #3299.

@jameslamb Could you please help with this like you've already done in #3639?

jameslamb:
sure! Would be happy to

PR #3809, opened 2021-01-21
Move compute and eigen libraries to external_libs folder

### LINKED ISSUE #3717: [RFC] Move compute and eigen libraries to external_libs folder
As the number of third-party dependencies is growing, I think it will be good to keep their source files in one place. `external_libs` folder looks good for this purpose. We already store `fast_double_parser` and `fmt` there.

StrikerRUS:
Fixed #3717.

I checked sdist size (#3684) with changes from this PR: approximately the same (not exactly due to longer paths in `SOURCES.txt` file and some other minor reasons) as for `master`.


guolinke (approved):

jameslamb (approved):
looks great to me. Thanks for checking the `sdist` size again! I think this is a really nice reorganization

PR #3811, opened 2021-01-22
[python-package] migrate test_plotting.py to pytest

thomasjpfan:
Towards https://github.com/microsoft/LightGBM/issues/3732

### DIFF
--- a/tests/python_package_test/test_plotting.py
+++ b/tests/python_package_test/test_plotting.py
@@ -88,8 +88,8 @@ def test_plot_split_value_histogram(params, breast_cancer_split, train_data):
                                          title='Histogram for feature @index/name@ @feature@',
                                          xlabel='x', ylabel='y', color='r')
     assert isinstance(ax1, matplotlib.axes.Axes)
-    assert (ax1.get_title() ==
-            'Histogram for feature name {}'.format(gbm1.booster_.feature_name()[27]))
+    title = 'Histogram for feature name {}'.format(gbm1.booster_.feature_name()[27])
+    assert ax1.get_title() == title
     assert ax1.get_xlabel() == 'x'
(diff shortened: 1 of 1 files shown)


tests/python_package_test/test_plotting.py
```diff
@@ -14,164 +13,188 @@
...
+
+    gbm1 = lgb.LGBMClassifier(n_estimators=10, num_leaves=3, silent=True)
+    gbm1.fit(X_train, y_train)
+
+    ax1 = lgb.plot_split_value_histogram(gbm1, gbm1.booster_.feature_name()[27], figsize=(10, 5),
+                                         title='Histogram for feature @index/name@ @feature@',
+                                         xlabel='x', ylabel='y', color='r')
+    assert isinstance(ax1, matplotlib.axes.Axes)
+    assert (ax1.get_title() ==
+            'Histogram for feature name {}'.format(gbm1.booster_.feature_name()[27]))
```
jameslamb, line 92:
```suggestion
    title = 'Histogram for feature name {}'.format(gbm1.booster_.feature_name()[27])
    assert ax1.get_title() == title
```

I think this change would address the issue from linting ():

> Linting Python code
./tests/python_package_test/test_plotting.py:91:29: W504 line break after binary operator

jameslamb (changes requested):
Thanks so much for all your help! I just left one small suggestion about linting.

I reviewed with https://www.textcompare.org/index.html?id=600a657520516c001761fc4c.

I like the use of module-scoped features to replicated what we were doing in `setUp()`, thank you.

StrikerRUS (approved):
LGTM!
Thank you!

jameslamb (approved):

PR #3813, opened 2021-01-22
[python-package] respect parameter aliases for network params

jameslamb:
Based on https://github.com/microsoft/LightGBM/pull/3789#pullrequestreview-571573162. This PR proposes some changes in the `Booster` constructor to respect aliases for [network parameters](https://lightgbm.readthedocs.io/en/latest/Parameters.html#network-parameters).

This PR also has the following small changes

* replaces mistaken parameter name "listen_time_out" with "time_out"
* adds a rule to .gitignore to ignore `.pytest_cache/` directories  (I found these were created but not ignored while I was testing)

### Notes for Reviewers

* `time_out` does not have any aliases (https://lightgbm.readthedocs.io/en/latest/Parameters.html#time_out)


python-package/lightgbm/basic.py
```diff
@@ -2081,16 +2081,39 @@ def __init__(self, params=None, train_set=None, model_file=None, model_str=None,
...
+                            break
+
+                    # if num_machines is in params, use that. Otherwise, take
+                    # num_machines implied by "machines"
+                    num_machines_from_params = None
+                    for machine_param in _ConfigAliases.get("num_machines"):
+                        val = params.pop(machine_param)
+                        if num_machines_from_params is None:
+                            num_machines_from_params = val
+                            break
```
StrikerRUS, line 2106: Remove all aliases from `params` to keep `params` clean like here
https://github.com/microsoft/LightGBM/blob/d9a96c90cb479cef87047ba20517d97982b563eb/python-package/lightgbm/engine.py#L146-L150
```suggestion
                    for port_param in _ConfigAliases.get("local_listen_port"):
                        val = params.get(port_param)
                        if val is not None:
                            local_listen_port = val
                            params["local_listen_port"] = val

                    # if num_machines is in params, use that. Otherwise, take
                    # num_machines implied by "machines"
                    num_machines_from_params = None
                    for machine_param in _ConfigAliases.get("num_machines"):
                        val = params.pop(machine_param)
                        if num_machines_from_params is None:
                            num_machines_from_params = val
```

jameslamb: @StrikerRUS since this pattern is repeated in a few places, what do you think about me changing this PR to introduce a function that does this?

Like this

```python
def _choose_param_value(main_param_name: str, params: Dict[str, Any], default_value: Any):
    params = deepcopy(params)

    # find a value, and remove other aliases with .pop()
    found_value = None
    for param in _ConfigAliases.get(main_param_name):
        val = params.pop(param)
        if found_value is None and val is not None:
            found_value = val

    if found_value is not None:
        params[main_param_name] = found_value
    else:
        params[main_param_name] = default_value

    return params


params = _choose_param_value(
    main_param_name="local_listen_port",
    params=params,
    default_value=12400
)
params = _choose_param_value(
    main_param_name="num_machines",
    params=params,
    default_value=num_machines_from_machine_list
)
```

Then we could re-use this everywhere that we want to do this type of operation. That would allow us to test it independently and to add logging in one place if we wanted (like a warning about duplicated parameters).

StrikerRUS: Пкуфе швуфб Ш дшлув ше!

jameslamb: thanks! Adding this right now. Ignore the diff here for now, I'll re-request a review when this is ready

jameslamb: ok @StrikerRUS I think this is ready for another review.

I'm also proposing moving all of the parameter stuff in `lightgbm.dask._train()` up near the beginning of that function. That will avoid the case where you have to wait for LightGBM to do all the data setup stuff, only to find out afterwards that training fails because of an issue in parameters.

python-package/lightgbm/basic.py
```diff
@@ -2144,16 +2186,29 @@ def __init__(self, params=None, train_set=None, model_file=None, model_str=None,
...
+                        params=params,
+                        default_value=12400
+                    )
+                    self.set_network(
+                        machines=machines,
+                        local_listen_port=params["local_listen_port"],
+                        listen_time_out=params.get("listen_time_out", 120),
+                        num_machines=params["num_machines"]
+                    )
                     break
```
StrikerRUS, line 2212: I think we should remove all aliases of `machines` from `params`. This `break` prevent doing this.

python-package/lightgbm/dask.py
```diff
@@ -197,6 +197,43 @@ def _train(client, data, label, params, model_factory, sample_weight=None, group
...
+    if params["tree_learner"] not in allowed_tree_learners:
+        _log_warning('Parameter tree_learner set to %s, which is not allowed. Using "data" as default' % tree_learner)
+        params['tree_learner'] = 'data'
+
+    if params['tree_learner'] not in {'data', 'data_parallel'}:
+        _log_warning(
+            'Support for tree_learner %s in lightgbm.dask is experimental and may break in a future release. Use "data" for a stable, well-tested interface.' % params['tree_learner']
+        )
+
+    # Some passed-inparameters can be removed:
```
StrikerRUS, line 228:
```suggestion
    # Some passed-in parameters can be removed:
```

tests/python_package_test/test_basic.py
```diff
@@ -329,3 +329,49 @@ def check_asserts(data):
...
+    )
+    assert params["learning_rate"] == 0.789
+
+    # all changes should be made on copies and not modify the original
+    expected_params = {
+        "local_listen_port": 1234,
+        "port": 2222,
+        "metric": "auc",
+        "num_trees": 81
+    }
```
StrikerRUS, line 376: I think `expected_params` can be gotten by copying `original_params` at the very beginning of the test to not re-type every value again here.
```suggestion
    expected_params = copy(original_params)
```

StrikerRUS (changes requested):
Thanks for improving alias handling! Please check my comments:

tests/python_package_test/test_basic.py
jameslamb: `copy()` performs a shallow clone, so some modifications to `original_params` can still impact the copy. https://docs.python.org/3/library/copy.html#copy.copy

I think it's preferable for this test to re-type the literal values, so that the test doesn't get a false positive because of subtle mistakes where we assumed something was a copy but actually was a reference

python-package/lightgbm/basic.py
```diff
@@ -352,6 +354,46 @@ def get(cls, *args):
         return ret
 
 
+def _choose_param_value(main_param_name: str, params: Dict[str, Any], default_value: Any) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
jameslamb: I like this name very much and think it's descriptive of what is happening. Given a dictionary of parameters, it chooses a single value to use for a single parameter.

python-package/lightgbm/basic.py
```diff
@@ -6,9 +6,11 @@
 import os
 import warnings
 from collections import OrderedDict
+from copy import deepcopy
```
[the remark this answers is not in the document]
jameslamb: oh I see, ok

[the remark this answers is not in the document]
jameslamb: updated in https://github.com/microsoft/LightGBM/pull/3813/commits/481a5b4faef0eae39ebf5085a24fba85b90a613f

python-package/lightgbm/basic.py
jameslamb: ok, I proposed something for this in https://github.com/microsoft/LightGBM/pull/3813/commits/63387db2992ce13f1490a8b1202e30e21142190f

I think the cleanest way to do it is to use `_choose_param_value()` to get a single value for `machines`. Then we don't need a separate `for` loop with a `break`.

StrikerRUS:
@jameslamb 
Could you please resolve conflicts?

jameslamb:
> @jameslamb
> Could you please resolve conflicts?

done

python-package/lightgbm/dask.py
```diff
@@ -196,6 +196,43 @@ def _train(client, data, label, params, model_factory, sample_weight=None, group
...
+        'voting',
+        'voting_parallel'
+    }
+    if params["tree_learner"] not in allowed_tree_learners:
+        _log_warning('Parameter tree_learner set to %s, which is not allowed. Using "data" as default' % tree_learner)
+        params['tree_learner'] = 'data'
+
+    if params['tree_learner'] not in {'data', 'data_parallel'}:
+        _log_warning(
+            'Support for tree_learner %s in lightgbm.dask is experimental and may break in a future release. Use "data" for a stable, well-tested interface.' % params['tree_learner']
```
StrikerRUS, line 224:
```suggestion
            'Support for tree_learner %s in lightgbm.dask is experimental and may break in a future release. \n'
            'Use "data" for a stable, well-tested interface.' % params['tree_learner']
```

tests/python_package_test/test_basic.py
```diff
@@ -329,3 +329,49 @@ def check_asserts(data):
...
+
+    # should choose a value from an alias and set that value on main param
+    # if only an alias is used
+    params = lgb.basic._choose_param_value(
+        main_param_name="num_iterations",
+        params=params,
+        default_value=17
+    )
+    assert params["num_iterations"] == 81
+    assert "random_state" not in params
```
StrikerRUS, line 360: ?
```suggestion
    assert "num_trees" not in params
```

StrikerRUS (approved):
@jameslamb Thanks a lot for cleaning up aliases handling logic!
LGTM except one typo that you've already mentioned earlier
> replaces mistaken parameter name "listen_time_out" with "time_out"

tests/python_package_test/test_basic.py
jameslamb: ah!! I originally was testing `seed` and its aliases, but then realized we don't use `seed` in `_ConfigAliases`. Thank you

Issue #3817, opened 2021-01-22
[dask] flaky test_ranker test


StrikerRUS:
Just seen in #3755.

```
assert 0.8655947285611406 > 0.9
```

https://github.com/microsoft/LightGBM/blob/477cbf373ea2138a186568ac88ef221ac74c7c71/tests/python_package_test/test_dask.py#L480

<details>

<summary>Full logs:</summary>

```
2021-01-22T16:37:09.2139516Z ============================= test session starts ==============================
2021-01-22T16:37:09.2141377Z platform linux -- Python 3.8.5, pytest-6.2.1, py-1.10.0, pluggy-0.13.1
2021-01-22T16:37:09.2141965Z rootdir: /__w/1/s
2021-01-22T16:37:09.2142341Z collected 248 items
2021-01-22T16:37:09.2142606Z 
2021-01-22T16:37:09.5775745Z ../tests/python_package_test/test_basic.py .............                 [  5%]
2021-01-22T16:37:18.0619061Z ../tests/python_package_test/test_consistency.py ......                  [  7%]
2021-01-22T16:39:33.9324482Z ../tests/python_package_test/test_dask.py .............................. [ 19%]
2021-01-22T16:40:13.7140067Z ..F.......                                                               [ 23%]
2021-01-22T16:40:13.7150510Z ../tests/python_package_test/test_dual.py s                              [ 24%]
2021-01-22T16:40:16.8375077Z ../tests/python_package_test/test_engine.py ............................ [ 35%]
2021-01-22T16:40:58.6316816Z .......................................                                  [ 51%]
2021-01-22T16:40:59.4355458Z ../tests/python_package_test/test_plotting.py .....                      [ 53%]
2021-01-22T16:41:01.1371760Z ../tests/python_package_test/test_sklearn.py ........................... [ 64%]
2021-01-22T16:41:04.0277169Z ......x.........................................x....................... [ 93%]
2021-01-22T16:41:04.2180367Z .................                                                        [100%]
2021-01-22T16:41:04.2181120Z 
2021-01-22T16:41:04.2186375Z =================================== FAILURES ===================================
2021-01-22T16:41:04.2188206Z __________________________ test_ranker[group1-array] ___________________________
2021-01-22T16:41:04.2188660Z 
2021-01-22T16:41:04.2189171Z output = 'array'
[...]

PR #3819, opened 2021-01-22
[dask] Address flaky test_ranker tests

ffineis:
Addresses issue #3817. Correlation between dask's ranker scores and locally-trained ranker scores is stochastic, despite our best efforts to make deterministic `_make_ranking` data and using `seed=42` throughout DaskLGBM* tests.

I ran the test 100 times and this is the observed distribution:
spearmanr_dist

@StrikerRUS observed a run achieving only 0.86 correlation.
So lowering the threshold to a more conservative 0.75.

This is a stopgap - ideally there is a way to distribute data deterministically to dask workers in tests.

### DIFF
--- a/tests/python_package_test/test_dask.py
+++ b/tests/python_package_test/test_dask.py
@@ -512,7 +512,7 @@ def test_ranker(output, client, listen_port, group):
     # have high rank correlation with scores from serial ranker.
     dcor = spearmanr(rnkvec_dask, y).correlation
     assert dcor > 0.6
-    assert spearmanr(rnkvec_dask, rnkvec_local).correlation > 0.9
+    assert spearmanr(rnkvec_dask, rnkvec_local).correlation > 0.75
 
     client.close()
 


jameslamb:
don't worry about the appveyor failures (https://ci.appveyor.com/project/guolinke/lightgbm/builds/37402833), they're unrelated and I'll do a rebuild which should fix them

jameslamb (approved):
Thanks! I agree with this stopgap for now, to not block other non-Dask development on this project. Azure DevOps doesn't allow us to re-run individual failed jobs (you have to re-run ALL of them), so having a flaky test there is really expensive.

StrikerRUS:
@jameslamb 

> Azure DevOps doesn't allow us to re-run individual failed jobs (you have to re-run ALL of them), so having a flaky test there is really expensive.

Actually, you can re-run **failed** jobs. It's really useful. 🙂  Thanks Azure, and shame on GitHub Actions where you really need to re-run **all** jobs.

But agree that any re-run costs a lot, because each job runs ~15 mins.

image

StrikerRUS (approved):
@ffineis Thanks a lot for prompt workaround!

Hope we will be able to find
> a way to distribute data deterministically to dask workers in tests.

jameslamb:
Guess my account on Azure DevOps doesn't have permission to do that, because I don't see that button. I tried to access recent failed jobs (while signed in with my GitHub, using "sign in with GitHub"), and keep getting kicked with this error

> {"$id":"1","innerException":null,"message":"TF400813: The user 'Windows Live ID\\JayLamb20@gmail.com' is not authorized to access this resource.","typeName":"Microsoft.TeamFoundation.Framework.Server.UnauthorizedRequestException, Microsoft.TeamFoundation.Framework.Server","typeKey":"UnauthorizedRequestException","errorCode":0,"eventId":3000}

Issue #3156, opened 2020-06-10
Jupyter notebook output is flooded by LGBM output


Dronablo:
## How you are using LightGBM?

LightGBM component: Python package

## Environment info

Operating System: Windows 10

CPU/GPU model: i7-9800X / RTX 2060

C++ compiler version: Microsoft (R) C/C++ version 19.25.28610.4 for x64

CMake version: 3.16.19112601-MSVC_2

Python version: 3.8.3

LightGBM version or commit hash: ce95d9ca8aab8fd007a2682d4cc80157b4e1b801

## Error message and / or logs
I use Lightgbm in Jyputer notebooks.
Before this recompile all the Lightgbm output went to stdout/stderr -> console.
Now it all goes to notebook output, which, in turn, must be displayed by browser. Due to the amount of data it fall into near-hang state very soon.

Sample output:
```
[LightGBM] [Warning] Find whitespaces in feature_names, replace with underlines
[LightGBM] [Info] Number of positive: 8504, number of negative: 56791
[LightGBM] [Info] This is the GPU trainer!!
[LightGBM] [Debug] Dataset::GetMultiBinFromSparseFeatures: sparse rate 0.944056
[LightGBM] [Info] Total Bins 381824
[LightGBM] [Info] Number of data points in the train set: 65295, number of used features: 1661
[LightGBM] [Info] Using GPU Device: GeForce RTX 2060 SUPER, Vendor: NVIDIA Corporation
[LightGBM] [Info] Compiling OpenCL Kernel with 256 bins...
[LightGBM] [Info] GPU programs have been built
[LightGBM] [Info] Size of histogram bin entry: 8
[LightGBM] [Info] 1575 dense feature groups (98.14 MB) transferred to GPU in 0.049000 secs. 1 sparse feature groups
[LightGBM] [Info] [binary:BoostFromScore]: pavg=0.130240 -> initscore=-1.187673
[LightGBM] [Info] Start training from score -1.187673
[LightGBM] [Debug] Trained a tree with leaves = 9 and max_depth = 5
```
Is there any way to redirect it back to console or, which is probably better, somehow handle with the default python logging configs?

I believe these changes are related to #3090.

PR #3820, opened 2021-01-23
[python] Allow to register custom logger in Python-package

### LINKED ISSUE #3805: Logging to Python Log
As a result of https://github.com/microsoft/LightGBM/pull/3090 LGBM now seems to happily redirect it's output to the standard python logger which is great. At present it seems that the output from any custom feval function passed to the .train() method, still goes to stdout, is this expected behaviour? It would be nice if it too got redirected to the python logger!

Thanks!

StrikerRUS:
Centralize all logging logic in one place.
Fixed #3805.
Related #3156.


guolinke (approved):
LGTM

StrikerRUS:
@jameslamb 

Can these two logger calls (this is the all logs we have from Dask so far) be replaced with new `_log()` function?

https://github.com/microsoft/LightGBM/blob/ac706e10e461d1c5987a3975cc48aa82fde33566/python-package/lightgbm/dask.py#L28
https://github.com/microsoft/LightGBM/blob/ac706e10e461d1c5987a3975cc48aa82fde33566/python-package/lightgbm/dask.py#L260
https://github.com/microsoft/LightGBM/blob/ac706e10e461d1c5987a3975cc48aa82fde33566/python-package/lightgbm/dask.py#L263

Please check this: https://github.com/dmlc/xgboost/pull/6609#discussion_r561641058.
> The logging module should be preferred over the simple print, since Dask has a good integration with the logging module. See related discussion at rapidsai/cuml#3350.

python-package/lightgbm/basic.py
```diff
@@ -15,9 +17,64 @@
...
+
+    return wrapper
+
+
+def _log(msg, is_warning=False):
+    """Output LightGBM's logs."""
+    if is_warning:
+        _LOGGER.warning(msg)
+    else:
+        _LOGGER.info(msg)
```
jameslamb, line 67: instead of using a flag for the log level, I'd prefer functions `_log_warn()` and `_log_info()`. I think that makes the code easier to understand, and is easier to extend to other log levels (error, fatal, and debug). That is how most loggers are structured.

jameslamb (changes requested):
I really like this! To answer your question...yes I think this can and should be used by the code in `dask.py` as well.

I left a suggestion for the a change to the API, though.

python-package/lightgbm/basic.py
StrikerRUS: I don't think that adding new arguments like `is_debug`, `is_fatal` is harder and less clear than writing new functions. I believe it's more a matter of taste. But I don't have any strong preferences, so I'm going to implement your suggestion.

StrikerRUS: @jameslamb 

> yes I think this can and should be used by the code in `dask.py` as well.

Done in 48c405828db7071afabf43f1e68a789d5575b10d.

> instead of using a flag for the log level, I'd prefer functions `_log_warn()` and `_log_info()`

Done in 7247790078184646271e61b3bfe1d762bb57b985.

jameslamb (approved):
looks great! Really nice addition

PR #3821, opened 2021-01-23
[python][tests] transfer test_save_and_load_linear to test_engine

StrikerRUS:
`test_basic.py` contains tests only for `Dataset` and `Booster` classes. There is `train()` function which we test in `test_engine.py` file in `test_save_and_load_linear` test.

<No changes, simple copy-paste.>

### DIFF
--- a/tests/python_package_test/test_basic.py
+++ b/tests/python_package_test/test_basic.py
@@ -110,31 +110,6 @@ def test_chunked_dataset_linear():
     valid_data.construct()
 
 
-def test_save_and_load_linear(tmp_path):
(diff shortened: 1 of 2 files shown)


jameslamb (approved):
sounds good

jameslamb:
Azure DevOps might be experiencing an outage

image

even though there are no issues on https://status.dev.azure.com/

StrikerRUS:
> Azure DevOps might be experiencing an outage

Microsoft-hosted agents are running right now... Self-hosted ones are stuck.

Issue #3814, opened 2021-01-22
[dask] tutorial documentation


jameslamb:
## Summary

The docs at https://lightgbm.readthedocs.io/en/latest/Parallel-Learning-Guide.html should be updated with information on the Dask interface in the Python package.

References to https://github.com/dask/dask-lightgbm should also be removed from this repo.

## References

This could draw heavily from this tutorial I wrote on `dask-lightgbm` (removing any references to paid services, of course): https://www.saturncloud.io/docs/tutorials/lightgbm/.

jameslamb:
Adding this to #2302 alongside other feature requests. I'll re-open this when I pick it up.

jameslamb:
I'm reopening this and assigning to myself since I'm actively working on it, as a next step toward release 3.2.0 (#3872)

PR #3822, opened 2021-01-23
[dask][docs] initial setup for Dask docs

StrikerRUS:
Towards #3814.

Maybe we can ask to [archive](https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/archiving-repositories) old [dask-lightgbm](https://github.com/dask/dask-lightgbm) repo?

Live demo: https://lightgbm.readthedocs.io/en/dask_docs/Python-API.html#dask-api.

Why don't we have the most general class `DaskLGBMModel`?

### DIFF
--- a/python-package/lightgbm/dask.py
+++ b/python-package/lightgbm/dask.py
@@ -387,7 +387,7 @@ def _predict(model, data, raw_score=False, pred_proba=False, pred_leaf=False, pr
 class _LGBMModel:
     def __init__(self):
         if not all((DASK_INSTALLED, PANDAS_INSTALLED, SKLEARN_INSTALLED)):
-            raise LightGBMError('Dask, Pandas and Scikit-learn are required for this module')
+            raise LightGBMError('dask, pandas and scikit-learn are required for lightgbm.dask')
 
     def _fit(self, model_factory, X, y=None, sample_weight=None, group=None, client=None, **kwargs):
         """Docstring is inherited from the LGBMModel."""
--- a/python-package/lightgbm/dask.py
+++ b/python-package/lightgbm/dask.py
@@ -22,7 +22,8 @@ from dask import dataframe as dd
 from dask import delayed
 from dask.distributed import Client, default_client, get_worker, wait
 
-from .basic import _ConfigAliases, _LIB, _safe_call
+from .basic import _ConfigAliases, _LIB, _safe_call, LightGBMError
+from .compat import DASK_INSTALLED, PANDAS_INSTALLED, SKLEARN_INSTALLED
 from .sklearn import LGBMClassifier, LGBMRegressor, LGBMRanker
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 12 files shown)


jameslamb:
> Maybe we can ask to archive old dask-lightgbm repo?

After LightGBM 3.2.0 is released, I'd like to make a pull request in `dask-lightgbm` that adds a warning to all public functions and constructors saying "dask-lightgbm will not receive further updates. Please use lightgbm instead ('pip install lightgbm[dask]')". If they agree, then after that release we can ask for that repo to be archived.

> Why don't we have the most general class DaskLGBMModel?

What is the value of `LGBMModel`? I don't understand why someone would use that instead of one of the task-specific classes (`DaskLGBMRanker`, `DaskLGBMRegressor`, `DaskLGBMClassifier`). Could you help me understand that?

python-package/lightgbm/dask.py
```diff
@@ -384,6 +385,9 @@ def _predict(model, data, raw_score=False, pred_proba=False, pred_leaf=False, pr
 
 
 class _LGBMModel:
+    def __init__(self):
+        if not all((DASK_INSTALLED, PANDAS_INSTALLED, SKLEARN_INSTALLED)):
+            raise LightGBMError('Dask, Pandas and Scikit-learn are required for this module')
```
jameslamb, line 390:
```suggestion
            raise LightGBMError('dask, pandas and scikit-learn are required for lightgbm.dask')
```

Instead of "this module", could you use the specific name? I think that makes the log message a little more useful standalone. It can be helpful for cases where people don't have direct access to the stack trace, which is required to understand what "this module" refers to.

For example, user  code or other frameworks might write things like this

```python
try:
    dask_reg = DaskLGBMClassifier()
except LightGBMError as err:
    log.fatal(err)
    raise SomeOtherException("LightGBM training failed")
```

I also think packages should be referenced by their exact package names, not capitalized names.

python-package/setup.py
```diff
@@ -344,7 +344,7 @@ def run(self):
           extras_require={
               'dask': [
                   'dask[array]>=2.0.0',
-                  'dask[dataframe]>=2.0.0'
+                  'dask[dataframe]>=2.0.0',
```
jameslamb, line 347: oh wow, thank you!

jameslamb (changes requested):
thank you for this! I left some small suggestions, but I agree with these changes.

python-package/setup.py
StrikerRUS: TBH, first time I noticed that was at LGTM site:
https://lgtm.com/projects/g/microsoft/LightGBM?mode=tree

python-package/lightgbm/dask.py
StrikerRUS: Thank you! Addressed in acac78fe852cef265a9dda4e4c0c3ddf30b91a72.

StrikerRUS:
@jameslamb 
> After LightGBM 3.2.0 is released, I'd like to make a pull request in

Great!

> What is the value of `LGBMModel`?

I believe this class is needed for extending scikit-learn features by supporting more objectives. For example, right now LightGBM supports [cross-entropy application](https://lightgbm.readthedocs.io/en/latest/Parameters.html#objective). LGBMClassifier cannot be used with this objective because scikit-learn checks targets to be `int`s
https://github.com/microsoft/LightGBM/blob/ac706e10e461d1c5987a3975cc48aa82fde33566/python-package/lightgbm/sklearn.py#L818
One can workaround this with `LGBMRegressor` and `cross_entropy` objective but I don't think this is semantically correct.

Some other checks can be added in `RegressorMixin` and `ClassifierMixin` at scikit-learn side that will prevent using some potential new LightGBM features.

So in general `LGBMModel` is for everything beyond classification and regression.

jameslamb (approved):
these changes look good, thanks very much!

jameslamb:
> So in general LGBMModel is for everything beyond classification and regression.

Ok, thanks for the explanation! I've created #3845 for the feature request.

Issue #3768, opened 2021-01-15
[dask] Search for open ports only once per IP


jameslamb:
## Summary

In LightGBM distributed training (documented [here](https://lightgbm.readthedocs.io/en/latest/Parallel-Learning-Guide.html) and [here](https://lightgbm.readthedocs.io/en/latest/Features.html#optimization-in-parallel-learning)), each worker needs access to a list of all other workers' IPs + a port to communicate with them over.

#3766 updated `lightgbm.dask` to search for open ports on each worker when creating this list, instead of just assuming a fixed range of ports would be available.

This works well, but it's a blocking operation that has to be done sequentially, so it slows down training.

The following pseudocode illustrates the process:

```python
worker_to_port = {}
ports_seen_so_far = set()

for worker_address in worker_addresses:
    port = client.submit(
        _find_an_open_port,
       worker_address,
       ports_seen_so_far
    )
    worker_to_port[worker_address] = port
    ports_seen_so_far.add(port)
```

This is done sequentially because multiple Dask worker processes can live on the same IP address. So if you use a `LocalCluster` with 3 workers, for example, all 3 of those workers will be on your local machine. Or if you use a multi-machine cluster with `nprocs > 1`, multiple worker processes will run on each physical machine in the cluster.

As a result of this change, the time complexity of that "find open ports" step is `O(num_worker_processes)`. If instead we only did the search once per IP address, then this check could be safely parallelized, and the time complexity would be more like `O(nprocs)`.

To close this issue, change `lightgbm.dask._find_ports_for_workers()` (https://github.com/microsoft/LightGBM/blob/f6d2dce4ee4aa1df11892b4eeeff1a67c33095d4/python-package/lightgbm/dask.py#L72) to instead dispatch a function to each worker machine that returns a list of as many open ports as there are Dask workers on that machine.

## Motivation

This optimization would reduce the overhead introduced by using Dask for distributed training, which should make training faster.

## References

This could be done following something like the code @ffineis provided in https://github.com/microsoft/LightGBM/pull/3766#discussion_r557927569.

jameslamb:
Adding this to #2302 and closing it, per our practice for managing features. Anyone is welcome to contribute this feature! Please leave a comment here if you're interested in contributing it.

jameslamb:
This was fixed by https://github.com/microsoft/LightGBM/pull/3823

PR #3823, opened 2021-01-23
[dask] use random ports in network setup

jmoralez:
This is intended to solve the problem that @jameslamb defined in #3768 by running `socket.bind('', 0)` with `client.run`, which finds a random port for each worker. I've included a small test to check that the found ports are indeed different for a `LocalCluster`. Happy to receive any feedback.


jameslamb (commented):
Thanks for your contribution and interest in LightGBM @jmoralez !

It does seem like this produces the desired behavior on `LocalCluster`, but I'm concerned that this might be disruptive for users on multi-machine clusters, such as those that can be created with [`dask-cloudprovider`](https://github.com/dask/dask-cloudprovider/). With the approach proposed in this PR, you lose the ability to control the range of ports that are used by lightgbm, which I think would be problematic for people working in environments that are limited by firewall rules. What do you think?

jmoralez:
Thank you for your feedback, @jameslamb! I don't think there will be a problem by using a different type of cluster, since `client.run` executes the function in every worker, so the found ports will be open in that (possibly remote) worker, I will try this out tomorrow using coiled. I understand your point about loosing control of the range of ports found, however in a remote setting the client only communicates to the scheduler and I believe the ports chosen to run each worker are defined in the same way as I proposed (I haven't been able to confirm this), so I think the scheduler should be able to access the ports found for lightgbm without a problem. I'm by no means an expert in this matter and could be wrong, so I'm happy to get your thoughts on this and discuss it further.

jameslamb:
> I understand your point about loosing control of the range of ports found, however in a remote setting the client only communicates to the scheduler and I believe the ports chosen to run each worker are defined in the same way as I proposed (I haven't been able to confirm this), so I think the scheduler should be able to access the ports found for lightgbm without a problem

Ah! Let me add more information on this.

By default Dask does find ports for that worker-to-scheduler communication randomly, but it exposes options to control that more tightly: https://docs.dask.org/en/latest/setup/cli.html#handling-ports.

The ports for LightGBM aren't related to Dask directly. LightGBM has its own distributed learning framework written in C++ (https://github.com/microsoft/LightGBM/tree/master/src/network). This Dask package just wraps that framework. In the Dask package here, we set up one long-running Dask task per worker, and that task is running a LightGBM worker. The LightGBM workers talk to each other directly over TCP sockets (by default), and Dask has no knowledge of that communication.

>  I will try this out tomorrow using coiled

Thanks very much! If it's easier, I have a testing setup that uses `dask-cloudprovider` to run a cluster on AWS Fargate: https://github.com/jameslamb/lightgbm-dask-testing.

When I was working on #3766, I found that there were several fixes I tried which worked well on LocalCluster that didn't hold up in a multi-machine setting.

jmoralez:
Hi @jameslamb. I've successfully ran the `test_training_does_not_fail_on_port_conflicts` test on a coiled cluster with 4 machines and 2 cpus each, however I'm not sure how to share that with you. I have a software environment called `jmoralez/lightgbm-test` there if you want to try it out (however you'll still need the local version which is the part I'm not sure about). I had to set `sample_size=10_000` because I was getting the _Connecting to rank x failed_ error. Do you have any suggestions for sharing the tests or how to proceed?

jmoralez:
Hi James. I've set up the version of my branch [here](https://cloud.coiled.io/jmoralez/notebooks). I think this random port assignment could be a fallback when specifying `local_listen_port=0` and otherwise use the current implementation, what do you think?

As a side note I've noticed that with few samples I get `Connecting to rank x failed` like crazy. For example, by running the notebook as is I got stuck at the second fit and had to do a `client.restart`, however with `sample_size=10_000` it runs the 5 fits in 4 seconds. I believe this is related to #3797 and I experienced this a couple of months ago as well as the TCP fail with dask-lightgbm and made me drop it. I'll dig deeper into this.

jameslamb:
> As a side note I've noticed that with few samples I get Connecting to rank x failed like crazy. For example, by running the notebook as is I got stuck at the second fit and had to do a client.restart, however with sample_size=10_000 it runs the 5 fits in 4 seconds

Can you please open a separate issue for this so we can discuss there? I have some theories.

<hr>

> I've set up the version of my branch here. I think this random port assignment could be a fallback when specifying local_listen_port=0 and otherwise use the current implementation, what do you think?

really cool, thank you! I clicked into the notebook and ran it. I really really appreciate the effort you've put into this PR and into setting up this nice test harness in Coiled. I can see that the 5 consecutive runs worked!

image

I even tried changing the test and running  50 consecutive trainings back to back. No problems :).

image

I can see that this cluster's workers are on different IPs, so I think this is a great test that this PR will work in other multi-machine settings.

image

I like the idea of possibly using both this implementation and the current one, but I'd rather switch the order. I think it makes sense to have your proposed (faster) approach be the default behavior, and for the current behavior to only be used if people opt in to it because they're working in an environment with networking constraints like firewall rules that only permit specific port ranges. I think users will be ok with paying a small performance penalty to comply with their security setup, and that anyone NOT constrained in that way should get the faster method by default.

However, I don't think we should support setting `local_listen_port` to `0` or any other special value. I want the result of `model.get_params()` to return valid LightGBM parameters that you could take and use in other non-Dask LightGBM APIs, and `0` is not a valid value for `local_listen_port`.https://lightgbm.readthedocs.io/en/latest/Parameters.html#local_listen_port.

Ok sorry, that was a lot of information. Given all that, here's my proposal

* if `local_listen_port` is missing from parameters, use your approach from this PR
* if `local_listen_port` is set to 12400 (LightGBM's default), use your approach from this PR
* if `local_listen_port` is set to a value other than 12400, use the existing approach where we specifically search 1000 ports from `local_listen_port` to `local_listen_port + 999`.

What do you think?

jmoralez:
Hi James. That sounds good, I'll work on that and update this PR. Thank you!

jmoralez:
Hi James. Since `local_listen_port` is set to `12400` in case it is `None` then if its 12400 I search for a random open port and otherwise use the current implementation. Looking forward for your feedback.

tests/python_package_test/test_dask.py
```diff
@@ -234,26 +234,34 @@ def test_classifier_pred_contrib(output, centers, client, listen_port):
...
-            n_estimators=5,
-            num_leaves=5
-        )
-        for _ in range(5):
-            dask_classifier.fit(
-                X=dX,
-                y=dy,
-                sample_weight=dw,
-                client=client
+        for local_listen_port in (None, 12400, 13000):
```
jameslamb, line 250: could you please push this up into a `pytest.mark.parametrize`? See https://github.com/microsoft/LightGBM/blob/56fc036def837aa6a754ceb5cdecde212892edcd/tests/python_package_test/test_dask.py#L341 for an example if you've never used that before.

python-package/lightgbm/dask.py
```diff
@@ -107,6 +107,22 @@ def _find_ports_for_workers(client: Client, worker_addresses: Iterable[str], loc
     return worker_ip_to_port
 
 
+def _find_random_open_port():
```
jameslamb, line 110: we've been trying to use type hints everywhere in the Dask module, in a step towards #3756 

can you please add this return hint?

```suggestion
def _find_random_open_port() -> int:
```

python-package/lightgbm/dask.py
```diff
@@ -296,11 +312,18 @@ def _train(
...
     # own port
-    worker_address_to_port = _find_ports_for_workers(
-        client=client,
-        worker_addresses=worker_map.keys(),
-        local_listen_port=params["local_listen_port"]
-    )
+    if params["local_listen_port"] == 12400:
+        # If local_listen_port is set to LightGBM's default value
+        # then we find a random open port for each worker
+        worker_address_to_port = client.run(_find_random_open_port)
```
jameslamb, line 318: Can you please add `workers=worker_map.keys()` to this call? That would avoid running this on workers that don't hold any of the training data.

jameslamb (changes requested):
Thanks! Really like the changes. I agree with the decision you made about `params["local_listen_port"] == 12400` inside `_train()`, since by that point that parameter is guaranteed to exist (https://github.com/microsoft/LightGBM/blob/56fc036def837aa6a754ceb5cdecde212892edcd/python-package/lightgbm/dask.py#L226-L230).

I left some new suggestions. Once you've addressed those, I'll test this manually on a FargateCluster with `dask-cloudprovider`, and ask @StrikerRUS for a review.

jmoralez:
Thank you for your feedback, James. I've included your comments.

jameslamb (changes requested):
Just left one more style thing, otherwise looks ok to me! I'll pull this and test on AWS.

<hr>

@StrikerRUS whenever you have time, could you give your opinion? I can give you the relevant context so you don't have to re-read all the discussion here and in #3768. 

In #3766, I added a function `_find_open_port()` that the Dask package uses to build the `machines` list. But as I noted in https://github.com/microsoft/LightGBM/pull/3766#discussion_r557736104, the approach adds some overhead. It currently runs that check sequentially. So for a 3-worker cluster, it will go "find port on `w1` then find port on `w2` then find port on `w3`". This means there is O(n_workers) overhead introduced.

This PR proposes running a function to find a random port on every worker simultaneously, so in theory the overhead of setting up `machines` becomes O(1). This is faster, but it also means the ports used for `machines` are totally unrelated to `local_listen_port`.

In https://github.com/microsoft/LightGBM/pull/3823#issuecomment-768816075 I told @jmoralez that in my opinion, this faster behavior should be the default, but that we need to still give people who are limited by firewall rules the ability to specify a `local_listen_port` and know that only that port and those near it will be searched.

So we settled on the current state of this PR, which is:

1. if local_listen_port is missing from parameters, use the faster approach
1. if local_listen_port is set to 12400 (LightGBM's default), use the faster approach
1. if local_listen_port is set to a value other than 12400, use the `_find_open_port()` approach where we specifically search 1000 ports from `local_listen_port` to `local_listen_port + 999`.

python-package/lightgbm/dask.py
```diff
@@ -296,11 +312,19 @@ def _train(
...
-    worker_address_to_port = _find_ports_for_workers(
-        client=client,
-        worker_addresses=worker_map.keys(),
-        local_listen_port=params["local_listen_port"]
-    )
+    if params["local_listen_port"] == 12400:
+        # If local_listen_port is set to LightGBM's default value
+        # then we find a random open port for each worker
+        worker_address_to_port = client.run(_find_random_open_port,
+                                            workers=worker_map.keys())
```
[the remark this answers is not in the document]
jameslamb: I just changed my suggestion to `list(worker_map.keys())`, based on these test failures: https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=8954&view=logs&j=fb919173-48b9-522d-5342-9a59a13eb10b&t=7fcbeee5-d4c1-571c-8072-1d5874591254

> TypeError: can not serialize 'dict_keys' object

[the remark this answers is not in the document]
jmoralez: my bad for not running the tests locally. thanks for the indentation comment, I hadn't noticed the wrapping.

jameslamb:
Sorry for the inconvenience, but can you update this PR and #3897 to `master`? I just merged a PR that changes the Dask interface to move `client` from `.fit()` and `.predict()` into the estimator constructors (https://github.com/microsoft/LightGBM/pull/3883).

python-package/lightgbm/dask.py
```diff
@@ -126,6 +126,22 @@ def _find_ports_for_workers(client: Client, worker_addresses: Iterable[str], loc
...
+    """Find a random open port on the machine.
+
+    Returns
+    -------
+    port : int
+        A free port on the machine.
+    """
+    import socket
+
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
```
StrikerRUS, line 139: Remove this line. `socket` is already imported in appropriate place
https://github.com/microsoft/LightGBM/blob/c3ac77b570b687e948d63b2078adcf4c37f3dec5/python-package/lightgbm/dask.py#L9

```suggestion
    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
```

StrikerRUS (commented):
@jmoralez Thanks a lot for the contribution!
@jameslamb Many thanks for the detailed overview of the problem behind this PR!

I worrying about only one thing. After this PR port `12400` is treated as "not set". This is not true from the user point of view. Imagine the following scenario. One reads LightGBM docs and finds that port for distributed training can be set via [`local_listen_port`](https://lightgbm.readthedocs.io/en/latest/Parameters.html#local_listen_port) parameter and its default value is `12400`. User doesn't have any preferences for ports and decides to leave default value (or more critical problem: harcodes `12400` in params). The user goes to firewall setting and opens `12400` port with the hope that everything will work. But it won't work because actually **random**, not `12400`, port will be used inside LightGBM.

Things are getting worse due to that there is no dedicated `port` param in Dask wrapper that can be set to say `None` and take priority over `local_listen_port` is kwargs. I have only one solution for this inconsistency for now:
document in Dask docs that actually **only for Dask** port `12400 === random`.
Somehow like the following with `warning` directive:

image

jameslamb:
Thanks @StrikerRUS . @jmoralez , don't make any changes yet. I need to go re-read the code in `network` again to be sure I understand the implications of the choice we make here. Will report back soon.

python-package/lightgbm/dask.py
jmoralez: I did this on purpose because I thought `client.run` would fail without importing the module inside the function but it works, so thank you for pointing it out. I'll remove it in my next push.

jameslamb:
I haven't forgotten about this PR. I did a lot of research this weekend, reading through LightGBM's [`network/` source code](https://github.com/microsoft/LightGBM/tree/master/src/network) to understand exactly how the params `local_listen_port` and `machine_list` are used. I'll have a recommendation to share soon.

jameslamb:
Ok @jmoralez we just merged https://github.com/microsoft/LightGBM/pull/3994, so this PR should now be unblocked.

Could you please update this to the latest `master` and then replace my use `_find_ports_for_workers()` with your implementation here?

https://github.com/microsoft/LightGBM/blob/1f73f55938a6bfc84eb18b0f04288a5f039eef8a/python-package/lightgbm/dask.py#L418

jmoralez:
Thank you, James! Should I remove `_find_ports_for_workers`, `_find_open_port` and modify `test_find_open_port_works`?

jameslamb:
> Thank you, James! Should I remove `_find_ports_for_workers`, `_find_open_port` and modify `test_find_open_port_works`?

yes please

tests/python_package_test/test_dask.py
```diff
@@ -343,6 +335,19 @@ def test_classifier_pred_contrib(output, centers, client):
     client.close(timeout=CLIENT_CLOSE_TIMEOUT)
 
 
+def test_find_random_open_port(client):
```
jameslamb, line 338: excellent test! I really like this

jameslamb (approved):
Excellent work! This is a nice change that simplifies the code AND makes it faster.

The tests look good to me as well. Thanks very much!

jmoralez:
Thanks a lot for all your help with this, James!

Issue #3776, opened 2021-01-17
[dask] GPU support 


jameslamb:
## Summary

The Dask interface in https://github.com/microsoft/LightGBM/blob/706f2af7badc26f6ec68729469ec6ec79a66d802/python-package/lightgbm/dask.py currently only supports CPU-based training.

It should support GPU-based training as well, for those who want to use `lightgbm.dask` to perform multi-GPU training (https://github.com/dask/community/issues/104#issuecomment-718755797).

## Motivation

Adding this feature would allow users to use `lightgbm.dask` as a way to more easily perform multi-GPU training.

## Notes for Reviewers

This issue is intentionally vague on details, because "GPU support" could be done in different ways.

For example, we could consider allowing the Dask model objects to take in [`dask-cudf`](https://github.com/rapidsai/cudf/tree/e86cc6544a1e68bc194f69ef63c1890cbfe25681/python/dask_cudf) data frames if `lightgbm` has been built with CUDA-based GPU support.

## References

This issue was originally posted as https://github.com/dask/dask-lightgbm/issues/22, and more background is available there. Moving it here as part of https://github.com/dask/community/issues/104.

jameslamb:
Closing this, as we use #2302 to track feature requests. Leave a comment below if you'd like to contribute this feature, and we'll be happy to re-open it!

StrikerRUS:
Another action required to address this issue is bringing support of OpenCL-based GPU training or documenting why it can't be supported.

> At some point (maybe when investigating #3776 ), we should look again at why the Dask tests fail on non-CUDA GPU builds.
   https://github.com/microsoft/LightGBM/pull/3825#issuecomment-766271497

PR #3825, opened 2021-01-23
[ci] improve and run CUDA jobs for every commit and PR

StrikerRUS:
Due to 
> It seems currently both azure pipeline and github actions cannot support the power-off/turn-on machines.
Even with the VM scale set in azure pipeline, it needs at least one machine stand by.
I know there are some external solutions, but they seems to require my Azure (Microsoft) account permission, which seems to violate our policy.
    https://github.com/microsoft/LightGBM/pull/3424#issuecomment-701723558

machine with NVIDIA GPU is always turned on and I don't see any reasons to not run CUDA jobs on it for every PR in this case. Running CUDA jobs on a regular basis will make us sure that no one new PR breaks CUDA source code.

CUDA jobs cannot be run in parallel but the duration of all 3 jobs I set in this PR are about 20 min. Is it OK?

### DIFF
--- a/.ci/setup.sh
+++ b/.ci/setup.sh
@@ -77,6 +77,7 @@ else  # Linux
         echo libamdocl64.so > $OPENCL_VENDOR_PATH/amdocl64.icd
     fi
     if [[ $TASK == "cuda" ]]; then
+        echo 'debconf debconf/frontend select Noninteractive' | sudo debconf-set-selections
         apt-get update
         apt-get install --no-install-recommends -y \
             cmake \
--- a/.ci/setup.sh
+++ b/.ci/setup.sh
@@ -22,7 +22,7 @@ else  # Linux
     if [[ $IN_UBUNTU_LATEST_CONTAINER == "true" ]]; then
         # fixes error "unable to initialize frontend: Dialog"
         # https://github.com/moby/moby/issues/27988#issuecomment-462809153
-        echo 'debconf debconf/frontend select Noninteractive' | sudo debconf-set-selections
+        echo 'debconf debconf/frontend select Noninteractive' | debconf-set-selections
 
         sudo apt-get update
         sudo apt-get install -y --no-install-recommends \
--- a/.ci/setup.sh
+++ b/.ci/setup.sh
@@ -78,10 +78,10 @@ else  # Linux
     fi
     if [[ $TASK == "cuda" ]]; then
         apt-get update
-        apt-get install --no-install-recommends -y curl wget
-        curl -sL https://cmake.org/files/v3.18/cmake-3.18.1-Linux-x86_64.sh -o cmake.sh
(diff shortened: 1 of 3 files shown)


StrikerRUS:
@jameslamb 
I'm very surprised that even Dask tests are passed with CUDA version!

jameslamb:
> @jameslamb
> I'm very surprised that even Dask tests are passed with CUDA version!

that's great! I'd be surprised if it actually took advantage of the GPU, but at least the tests passing there is good. At some point (maybe when investigating #3776 ), we should look again at why the Dask tests fail on non-CUDA GPU builds.

> CUDA jobs cannot be run in parallel but the duration of all 3 jobs I set in this PR are about 20 min. Is it OK?

I think this is totally fine!

I looked at the most recent one and it took about 24 minutes to run (https://github.com/microsoft/LightGBM/actions/runs/506269302).

image

That's quicker than other collections of jobs, so I don't think it would have too big of an impact on the peak level of activity in this repo. Because it gives us extra coverage of a part of the codebase that is very new (in terms of how many times it has been modified in its lifetime), I think it is very worth it.

.ci/setup.sh
```diff
@@ -77,11 +77,12 @@ else  # Linux
         echo libamdocl64.so > $OPENCL_VENDOR_PATH/amdocl64.icd
     fi
     if [[ $TASK == "cuda" ]]; then
+        echo 'debconf debconf/frontend select Noninteractive' | debconf-set-selections
         apt-get update
-        apt-get install --no-install-recommends -y curl wget
-        curl -sL https://cmake.org/files/v3.18/cmake-3.18.1-Linux-x86_64.sh -o cmake.sh
-        chmod +x cmake.sh
-        ./cmake.sh --prefix=/usr/local --exclude-subdir
```
jameslamb, line 84: I'm really glad to see building CMake from source replaced with an installation from a package manager 🙌

jameslamb (approved):
I love this change! I think this will give us some new automatic test coverage that's really needed. This is going to help the pace of development a lot. Thank you for your efforts 🚀

.ci/setup.sh
StrikerRUS: Huh, I decided that it is enough to have the most recent CMake for our other Linux jobs run inside [Ubuntu 14.04 container](https://github.com/guolinke/lightgbm-ci-docker). Thanks to default CMake from apt-get in Ubuntu 20.04 is quite fresh and passes our minimum version check!
https://github.com/microsoft/LightGBM/blob/ac706e10e461d1c5987a3975cc48aa82fde33566/CMakeLists.txt#L21-L22

As a bonus, we don't need to update it in one more place regularly 🙂

StrikerRUS:
@guolinke Please take a look when have time.

guolinke (approved):
Thank you, LGTM

PR #3826, opened 2021-01-23
[ci] improve optional checks

StrikerRUS:
Do not waste time (~20s) to clone the repo.

Skip everything for comments not starting with `/gha run`

image


jameslamb (approved):
great idea

PR #3827, opened 2021-01-23
[dask][tests] move make_ranking into utils

StrikerRUS:
This ~can speed up tests by caching and also~ makes `make_ranking` function available for non-Dask tests.

**UPD:** OK, we cannot cache it due to unhashable `group` argument.

Diff for `make_ranking` (docstring fixed):

https://www.textcompare.org/index.html?id=600cacc151271d00170ec70c

image


jameslamb (approved):
oh awesome! good idea. Thanks again for contributing this @ffineis

PR #3828, opened 2021-01-24
[dask][tests] reduce code duplication in Dask tests

StrikerRUS:
Ensure that local and Dask estimators share the same parameters.

Argument should be preferred to param alias (`seed` -> `random_state`) because internally both values are stored and then cpp code produces warning for ambiguous values.


jameslamb (approved):
this is great, thank you

Issue #3834, opened 2021-01-24
Support all LightGBM parallel tree learners in Dask


StrikerRUS:
According to this piece of code 
https://github.com/microsoft/LightGBM/blob/ac706e10e461d1c5987a3975cc48aa82fde33566/python-package/lightgbm/dask.py#L251-L264
Dask module supports [all available tree learner types](https://lightgbm.readthedocs.io/en/latest/Parameters.html#tree_learner).
Refer to https://lightgbm.readthedocs.io/en/latest/Parallel-Learning-Guide.html#choose-appropriate-parallel-algorithm for more details about difference among `tree_learner` param values.

However, our CI tests are run only for `data` parallel tree learner type.
https://github.com/microsoft/LightGBM/blob/ac706e10e461d1c5987a3975cc48aa82fde33566/tests/python_package_test/test_dask.py#L182

I believe that parametrization of all tests with different tree learners will improve tests and make us confident in Dask module quality.

How tests can be parametrized:
https://github.com/microsoft/LightGBM/blob/ac706e10e461d1c5987a3975cc48aa82fde33566/tests/python_package_test/test_dask.py#L31
https://github.com/microsoft/LightGBM/blob/ac706e10e461d1c5987a3975cc48aa82fde33566/tests/python_package_test/test_dask.py#L166

jameslamb:
@StrikerRUS I disagree with calling this a "good first issue", and believe it should be a feature request. Using a different tree learner isn't as simple as just changing the tree_learner parameter. You have to be sure that the way that your data (either Dask DataFrame or Dask Array) is partitioned matches the method you chose.

For a dataset with `n` observations and `k`  columns:

* feature parallel = partitions have shape `(n, k_i)`. All rows but a subset of features.
* data parallel = partitions have shape `(n_i, k)`. All features but a subset of rows

Implementing this will require some non-trivial changes  to the tests and the Dask module.

jameslamb:
I'm going to close this and add it to #2302. Thank you very much for writing this up, and for all of the Dask issues you've written up! Having this all documented in the backlog takes a lot of work but it's also critical to getting other contributors to come help.

StrikerRUS:
> Implementing this will require some non-trivial changes to the tests and the Dask module.

Huh, I was suspecting this but shy to say it out loud! 🙂 

Given that only `data` tree learner is actually supported right now, I strongly believe we should make it explicit and raise errors for these values
[...]

PR #3848, opened 2021-01-25
[dask] warn if attempting to use tree_learner other than data parallel

jameslamb:
Based on https://github.com/microsoft/LightGBM/issues/3834#issuecomment-766351436.

Today, LightGBM's distributed learning supports three methods (feature parallel, data parallel, and voting parallel) --> https://lightgbm.readthedocs.io/en/latest/Features.html#optimization-in-parallel-learning

The Dask interface in `lightgbm.dask` is currently only tested with data parallel, where each chunk of the dataset is a row-wise chunk (all features, subset of rows).

The other methods MIGHT work with `lightgbm.dask`, but significant testing and documentation is needed. Supporting those methods might include adding extra validation to the Dask package as well.

## Changes in this PR

The other parallel learning methods are optimizations, and I don't think the work to support them (#3834) needs to make it into the next LightGBM release (3.2.0). So that testing and documentation won't happen in the next few weeks.

This PR proposes adding a warning if you use one of these methods, to tell users that support for methods other than "data" is experimental and might break.

@StrikerRUS I know you recommended an error in https://github.com/microsoft/LightGBM/issues/3834#issuecomment-766351436, but in my opinion it would be better to make this only a warning so that expert users (who know how to do the Dask partitioning for the different methods) are still able to experiment with the other methods.

### DIFF
--- a/python-package/lightgbm/dask.py
+++ b/python-package/lightgbm/dask.py
@@ -244,6 +244,7 @@ def _train(client, data, label, params, model_factory, sample_weight=None, group
     for tree_learner_param in _ConfigAliases.get('tree_learner'):
         tree_learner = params.get(tree_learner_param)
         if tree_learner is not None:
+            params['tree_learner'] = tree_learner
             break
 
     allowed_tree_learners = {
@@ -263,7 +264,7 @@ def _train(client, data, label, params, model_factory, sample_weight=None, group
 
     if params['tree_learner'] not in {'data', 'data_parallel'}:
         _log_warning(
-            'Support for tree_learner %s in lightgbm.dask is currently untested, and should be considered experimental' % params['tree_learner']
+            'Support for tree_learner %s in lightgbm.dask is experimental and may break in a future release. Use "data" for a stable, well-tested interface.' % params['tree_learner']
         )
 
     local_listen_port = 12400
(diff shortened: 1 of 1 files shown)


python-package/lightgbm/dask.py
```diff
@@ -244,6 +244,7 @@ def _train(client, data, label, params, model_factory, sample_weight=None, group
     for tree_learner_param in _ConfigAliases.get('tree_learner'):
         tree_learner = params.get(tree_learner_param)
         if tree_learner is not None:
+            params['tree_learner'] = tree_learner
```
jameslamb, line 247: I know this is a little awkward in the case where aliases are used, because multiple aliases for tree_learner might exist in params.

That change will only live on `master` for a short time, because I'm cleaning it up in https://github.com/microsoft/LightGBM/pull/3813#discussion_r562754584 right now

StrikerRUS (approved):
LGTM!

> I know you recommended an error in #3834 (comment), but in my opinion it would be better to make this only a warning so that expert users (who know how to do the Dask partitioning for the different methods) are still able to experiment with the other methods.

OK, but please be prepared to that it is **very common** to use 
```
import warnings

warnings.filterwarnings("ignore")
```
in Python scripts with LightGBM 😃

PR #3849, opened 2021-01-25
[dask] factor dask-ml out of tests (fixes #3796)

### LINKED ISSUE #3796: [dask] remove testing dependency on dask-ml
## Summary

See https://github.com/microsoft/LightGBM/pull/3708#discussion_r558573080. LightGBM's tests currently rely on installing `dask-ml`, but only for a few lightweight metrics functions.

`dask-ml` should be factored out of the tests and CI scripts.

## Motivation

This would reduce the risk of dependency conflicts in CI setup and speed up builds a bit.

## References

@ffineis originally did this in https://github.com/microsoft/LightGBM/pull/3708#discussion_r558573080, but I asked him to pull it out into a separate future PR.

jameslamb:
This PR fixes #3796.

We currently install `dask-ml` in CI, but only for the use of two small metrics functions in tests. To reduce CI times and other issues that dependencies can introduce (e.g. incompatibilities that lead to failed or slower conda environment solves, breaking API changes), this proposes removing `dask-ml` and just using our own small implementations of the two metrics needed by tests.

This PR uses @ffineis 's simple R^2 originally proposed in https://github.com/microsoft/LightGBM/pull/3708/commits/b39a1617af2eff0b618f4aa93a2d45ac7f2db963#diff-25c5c42e96fce8e0eb89cc7c991691d224c8b238cd91d2f795efe9fe7151750c.

### DIFF
--- a/tests/python_package_test/test_dask.py
+++ b/tests/python_package_test/test_dask.py
@@ -17,7 +17,6 @@ import pandas as pd
 from scipy.stats import spearmanr
 import scipy.sparse
 from dask.array.utils import assert_eq
-from dask_ml.metrics import accuracy_score, r2_score
 from distributed.utils_test import client, cluster_fixture, gen_cluster, loop
 from sklearn.datasets import make_blobs, make_regression
 from sklearn.utils import check_random_state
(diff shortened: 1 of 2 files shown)


tests/python_package_test/test_dask.py
```diff
@@ -124,6 +123,16 @@ def _create_data(objective, n_samples=100, centers=2, output='array', chunk_size
     return X, y, weights, dX, dy, dw
 
 
+def _r2_score(dy_true, dy_pred):
+    numerator = ((dy_true - dy_pred) ** 2).sum(axis=0, dtype="f8")
+    denominator = ((dy_true - dy_pred.mean(axis=0)) ** 2).sum(axis=0, dtype="f8")
```
StrikerRUS, line 128: Can `dtype` be provided in a following way (or `np.dtype(np.float64)`)? I think using common type names makes code more readable.
```suggestion
    numerator = ((dy_true - dy_pred) ** 2).sum(axis=0, dtype=np.float64)
    denominator = ((dy_true - dy_pred.mean(axis=0)) ** 2).sum(axis=0, dtype=np.float64)
```

StrikerRUS (approved):
Very nice! Thank you!

PR #3850, opened 2021-01-25
[R-package] Remove duplicated return values from docs

### LINKED ISSUE #3739: [R-package] duplicated return values in docs
There are duplicated lines in return statements in our rendered [R-package documentation](https://lightgbm.readthedocs.io/en/latest/R/reference/index.html). For instance,

https://lightgbm.readthedocs.io/en/latest/R/reference/slice.html#value
image

While normal return statement should look like the following
https://lightgbm.readthedocs.io/en/latest/R/reference/lgb.Dataset.construct.html#value
image

I might be wrong (@jameslamb can correct me), but I believe this happens due to duplicated `@return` directives in the source files.
https://github.com/microsoft/LightGBM/blob/c2dd54f13a58072090e8aaf7dd653f5c5ec1bc22/R-package/R/lgb.Dataset.R#L955
https://github.com/microsoft/LightGBM/blob/c2dd54f13a58072090e8aaf7dd653f5c5ec1bc22/R-package/R/lgb.Dataset.R#L973

To close this issue one should make `pkgdown` (library our R-package documentation is built by) render return value only once for each function/method.

philip-khor:
Resolves #3739

### DIFF
--- a/R-package/R/lgb.Dataset.R
+++ b/R-package/R/lgb.Dataset.R
@@ -970,7 +970,6 @@ slice <- function(dataset, ...) {
 }
 
 #' @rdname slice
-#' @return constructed sub dataset
 #' @export
 slice.lgb.Dataset <- function(dataset, idxset, ...) {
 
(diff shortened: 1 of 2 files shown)


jameslamb (approved):
thank you!

PR #3852, opened 2021-01-25
[dask][tests] skip Dask tests when Dask is not installed and improve imports in Dask tests

StrikerRUS:
Use benefits of #3822 and import `DaskLGBM*` classes from the root.

Import only things that are really used.

Skip `test_dask.py` when Dask is not installed. This is necessary to not force non-Dask contributors install all Dask stuff to run tests locally.


jameslamb (approved):
looks good, thanks!

Issue #3795, opened 2021-01-20
[dask] use dictionaries instead of tuples for parts in Dask training


jameslamb:
## Summary

The Dask interface allows you to train a LightGBM model on data stored in a Dask Array or Dask DataFrame. As part of this, it zips together corresponding chunks into something called "parts". So for example, if the training data are like this:

* `X`: a Dask DataFrame with data for features
* `y`: a Dask Array with labels

Then that interface might zip together the chunk of `X` with the first 1000 rows and the chunk of `y` with labels for those 1000 rows.

If this was just features + labels, using tuples would be ok. But it can also include sample weights and, for learning-to-rank tasks, groups. In the future, it might include `init_score` as well.

The Dask interface should stitch things together into *dictionaries* instead, keyed with understandable keys like `"data"`, `"labels"`, `"group"`, etc.

## Motivation

This change would reduce the risks of mistakes in the Dask interface and would make the code easier to read and change.

## References

@ffineis originally implemented this in #3708, but I asked him to pull it out into a separate PR. See https://github.com/microsoft/LightGBM/pull/3708#discussion_r558569333.

ffineis:
Dibs please!

jameslamb:
all yours!

StrikerRUS:
Closed via #3853.

PR #3853, opened 2021-01-25
[dask] [python] Store co-local data parts as dicts instead of lists

ffineis:
Addresses #3795. The `_train` function in `dask.py` places delayed "parts," i.e. `data`, `label`, `sample_weight`, and `group` data into lists (which get distributed to workers) to enforce co-locality of data.  Each worker is distributed a `list_of_parts`, a list of lists. Once the worker receives its lists of parts, it infers the `data`, `label`, and then optionally `sample_weight` and `group` from each `list_of_parts` by assuming that `data` is in position 0, `label` is in position 1, and if the model is *not* a DaskLGBMRanker, the second position is reserved for `sample_weight`.

This is an issue of readability + scalability - implying the presence of `sample_weight` and `group` based on the length of each sublist of `list_of_parts` in conjunction with the estimator type is... hard to understand and doesn't lend itself well to storing additional `parts` like `eval_sets`. Instead, let's just store individual sets of parts as dicts instead of lists. Now each worker will receive its portion of the overall dataset as a list of dicts instead of a list of lists.

### DIFF
--- a/python-package/lightgbm/dask.py
+++ b/python-package/lightgbm/dask.py
@@ -203,13 +203,13 @@ def _train(client, data, label, params, model_factory, sample_weight=None, group
 
     if sample_weight is not None:
         weight_parts = _split_to_parts(data=sample_weight, is_matrix=False)
-        for i, d in enumerate(parts):
-            parts[i] = {**d, 'weight': weight_parts[i]}
+        for i in range(len(parts)):
+            parts[i]['weight'] = weight_parts[i]
 
     if group is not None:
         group_parts = _split_to_parts(data=group, is_matrix=False)
-        for i, d in enumerate(parts):
-            parts[i] = {**d, 'group': group_parts[i]}
+        for i in range(len(parts)):
+            parts[i]['group'] = group_parts[i]
 
     # Start computation in the background
     parts = list(map(delayed, parts))
--- a/python-package/lightgbm/dask.py
+++ b/python-package/lightgbm/dask.py
@@ -136,25 +136,24 @@ def _train_part(params, model_factory, list_of_parts, worker_address_to_port, re
     is_ranker = issubclass(model_factory, LGBMRanker)
 
     # Concatenate many parts into one
-    parts = tuple(zip(*list_of_parts))
-    data = _concat(parts[0])
-    label = _concat(parts[1])
+    data = _concat([x['data'] for x in list_of_parts])
+    label = _concat([x['label'] for x in list_of_parts])
+
+    if 'weight' in list_of_parts[0]:
+        weight = _concat([x['weight'] for x in list_of_parts])
+    else:
+        weight = None
+
+    if 'group' in list_of_parts[0]:
+        group = _concat([x['group'] for x in list_of_parts])
+    else:
+        group = None
 
     try:
         model = model_factory(**params)
-
         if is_ranker:
-            group = _concat(parts[-1])
-            if len(parts) == 4:
-                weight = _concat(parts[2])
-            else:
-                weight = None
             model.fit(data, y=label, sample_weight=weight, group=group, **kwargs)
         else:
-            if len(parts) == 3:
-                weight = _concat(parts[2])
-            else:
-                weight = None
             model.fit(data, y=label, sample_weight=weight, **kwargs)
 
     finally:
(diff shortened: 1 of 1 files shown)


python-package/lightgbm/dask.py
```diff
@@ -197,29 +196,20 @@ def _train(client, data, label, params, model_factory, sample_weight=None, group
...
     data_parts = _split_to_parts(data=data, is_matrix=True)
     label_parts = _split_to_parts(data=label, is_matrix=False)
+    parts = [{'data': x, 'label': y} for (x, y) in zip(data_parts, label_parts)]
 
     if sample_weight is not None:
         weight_parts = _split_to_parts(data=sample_weight, is_matrix=False)
-    else:
-        weight_parts = None
+        for i, d in enumerate(parts):
+            parts[i] = {**d, 'weight': weight_parts[i]}
```
jameslamb, line 207: ~Will this result in each of the existing parts (data, label) in `d` being copied when they're passed back in? I think it might.~ Ok ignore that, I tested and was wrong about how `**` worked.

Still, I have a slight preference just modifying `parts[i]` directory, taking advantage of the fact that we know it's already a dictionary. 

I think this is a little easier to read, what do you think?

```suggestion
        for i in len(parts):
            parts[i]['weight'] = weight_parts[i]
```

jameslamb (changes requested):
This is great, thank you! I just have one very small nitpicky style thing.

python-package/lightgbm/dask.py
StrikerRUS: I absolutely agree with James! But it should be `for i in range(len(parts)):`

jameslamb: ha! typed too fast, thanks. Edited it.

python-package/lightgbm/dask.py
```diff
@@ -197,29 +196,20 @@ def _train(client, data, label, params, model_factory, sample_weight=None, group
...
-    if weight_parts is None and group_parts is None:
-        parts = zip(data_parts, label_parts)
-    elif weight_parts is not None and group_parts is None:
-        parts = zip(data_parts, label_parts, weight_parts)
-    elif weight_parts is None and group_parts is not None:
-        parts = zip(data_parts, label_parts, group_parts)
-    else:
-        parts = zip(data_parts, label_parts, weight_parts, group_parts)
+        for i, d in enumerate(parts):
+            parts[i] = {**d, 'group': group_parts[i]}
```
jameslamb, line 212:
```suggestion
        for i in range(len(parts)):
            parts[i]['group'] = group_parts[i]
```

python-package/lightgbm/dask.py
```diff
@@ -136,25 +136,24 @@ def _train_part(params, model_factory, list_of_parts, worker_address_to_port, re
...
     try:
         model = model_factory(**params)
-
         if is_ranker:
-            group = _concat(parts[-1])
-            if len(parts) == 4:
-                weight = _concat(parts[2])
-            else:
-                weight = None
             model.fit(data, y=label, sample_weight=weight, group=group, **kwargs)
```
StrikerRUS, line 155: I believe this whole `try` context can be simplified to single call of `model.fit(data, y=label, sample_weight=weight, group=group, **kwargs)`.
`group` param can be passed to the base `LGBMModel` class without the need to check whether it is ranker or not (`is_ranker`).

python-package/lightgbm/dask.py
ffineis: yep yeah, this is easier to understand, all for it

python-package/lightgbm/dask.py
ffineis: Yep this is true, good catch. This was a holdover from the world before `fit` was estimator-type-agnostic and each Dask model subclass had its own fit method. Will change!

ffineis: Wait no - sorry I take prior comment back. This is a call to the *local*/serial model `fit` method. Classifier/Regressor don't accept `type` argument: ```Exception: TypeError("fit() got an unexpected keyword argument 'group'")```. So I don't think that `group` can be safely passed as a kwarg to all model types.

StrikerRUS: Oh sorry, indeed! I forgot about that `LGBMClassifier` and `LGBMRegressor` overwrite `fit` method and `group` is not available there.

python-package/lightgbm/dask.py
ffineis: Addressed in 0d36cc4

python-package/lightgbm/dask.py
ffineis: Addressed in 0d36cc4

jameslamb (approved):
This is much cleaner and I think it will help reduce the risk of mistakes in the future. Thanks so much!

StrikerRUS (approved):
Awesome! Thanks! Code is much cleaner now.

PR #3854, opened 2021-01-25
[R-package][docs] improve grouping in R API reference

StrikerRUS:
I think `predict()` should be grouped together with `train()`, as the most primitive ML (name of the section in docs) can be described as fit+predict 😄 .


jameslamb (approved):
sure, no problem

PR #3862, opened 2021-01-26
[python] use better names for imported classes from extra libraries

StrikerRUS:
Use common library import aliases as a prefixes in class names for better readability and to avoid possible name collisions.
Refer to https://github.com/microsoft/LightGBM/pull/3855#discussion_r564556639.


jameslamb (approved):
love this, thank you!

PR #3864, opened 2021-01-26
[python] simplify param aliases handling

StrikerRUS:
`get()` can handle multiple param names at one call. 
https://github.com/microsoft/LightGBM/blob/066720eff680606abca17a9b34d95905337d8afd/python-package/lightgbm/basic.py#L348-L353


jameslamb (approved):
Oh nice! First time I've seen `|=` before.

PR #3865, opened 2021-01-26
[dask] add tests on warnings, fix incorrect variable in log

jameslamb:
While working on #3756 , I've been running `mypy python-package/`. This caught a bug in my refactoring from recent PRs (not sure which one).

> python-package/lightgbm/dask.py:251: error: Name 'tree_learner' is not defined

This PR fixes that, and adds unit tests to cover the behavior under the two warnings in `lightgbm.dask._train()`.

### DIFF
--- a/tests/python_package_test/test_dask.py
+++ b/tests/python_package_test/test_dask.py
@@ -446,7 +446,7 @@ def test_warns_and_continues_on_unrecognized_tree_learner(client):
     assert dask_regressor.fitted_
 
 
-def test_warns_but_makes_no_changes_for_feature_or_voting_parallel(client):
+def test_warns_but_makes_no_changes_for_feature_or_voting_tree_learner(client):
     X = da.random.random((1e3, 10))
     y = da.random.random((1e3, 1))
     for tree_learner in ['feature_parallel', 'voting']:
--- a/tests/python_package_test/test_dask.py
+++ b/tests/python_package_test/test_dask.py
@@ -430,6 +430,40 @@ def test_find_open_port_works():
             assert new_port == 12402
 
 
+def test_warns_and_continues_on_unrecognized_tree_learner(client):
+    X = da.random.random((1e3, 10))
+    y = da.random.random((1e3, 1))
+    dask_regressor = lgb.DaskLGBMRegressor(
+        time_out=5,
+        local_listen_port=1234,
+        tree_learner='some-nonsense-value',
+        n_estimators=1,
+        num_leaves=2
+    )
+    with pytest.warns(UserWarning, match='Parameter tree_learner set to some-nonsense-value'):
+        dask_regressor = dask_regressor.fit(X, y, client=client)
+
+    assert dask_regressor.fitted_
+
+
+def test_warns_but_makes_no_changes_for_feature_or_voting_parallel(client):
+    X = da.random.random((1e3, 10))
+    y = da.random.random((1e3, 1))
+    for tree_learner in ['feature_parallel', 'voting']:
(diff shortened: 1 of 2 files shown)


tests/python_package_test/test_dask.py
```diff
@@ -430,6 +430,40 @@ def test_find_open_port_works():
...
+        n_estimators=1,
+        num_leaves=2
+    )
+    with pytest.warns(UserWarning, match='Parameter tree_learner set to some-nonsense-value'):
+        dask_regressor = dask_regressor.fit(X, y, client=client)
+
+    assert dask_regressor.fitted_
+
+
+def test_warns_but_makes_no_changes_for_feature_or_voting_parallel(client):
```
StrikerRUS, line 449:
```suggestion
def test_warns_but_makes_no_changes_for_feature_or_voting_tree_learner(client):
```

StrikerRUS:
> I've been running `mypy python-package/`. 

Could you please setup `mypy` in `lint` task but allow it to fail (`|| exit 0`) so that everyone will be able to see logs and possibly prevent new issues to appear. We had the same strategy with `cpplint`. We had been checking logs manually but at some moment we simply removed `|| exit 0`.

https://github.com/microsoft/LightGBM/blob/066720eff680606abca17a9b34d95905337d8afd/.ci/test.sh#L42

tests/python_package_test/test_dask.py
```diff
@@ -430,6 +430,40 @@ def test_find_open_port_works():
...
+            local_listen_port=1234,
+            tree_learner=tree_learner,
+            n_estimators=1,
+            num_leaves=2
+        )
+        with pytest.warns(UserWarning, match='Support for tree_learner %s in lightgbm' % tree_learner):
+            dask_regressor = dask_regressor.fit(X, y, client=client)
+
+        assert dask_regressor.fitted_
+        assert dask_regressor.get_params()['tree_learner'] == tree_learner
```
StrikerRUS, line 464: Just wonder: is it possible to access params via attributes like it can be done in sklearn-wrapper?
```
dask_regressor.tree_learner
```

StrikerRUS (approved):
Thanks! Just one proposal for even more informative name.

tests/python_package_test/test_dask.py
jameslamb: maybe! I haven't tested. I didn't actually know the sklearn interface supported that.

But either way, I have a preference for `get_params()` here because it also gives us some confidence that this is working: https://github.com/microsoft/LightGBM/blob/59153b28750137bb2730920fdcedf1a7a8560b30/python-package/lightgbm/dask.py#L419-L420

jameslamb:
> > I've been running `mypy python-package/`.
> 
> Could you please setup `mypy` in `lint` task but allow it to fail (`|| exit 0`) so that everyone will be able to see logs and possibly prevent new issues to appear. We had the same strategy with `cpplint`. We had been checking logs manually but at some moment we simply removed `|| exit 0`.
> 
> https://github.com/microsoft/LightGBM/blob/066720eff680606abca17a9b34d95905337d8afd/.ci/test.sh#L42

sure! Sounds like an excellent `good first issue` idea for people to fix warnings.

tests/python_package_test/test_dask.py
StrikerRUS: Exactly `set_params()` should set attributes btw
https://github.com/microsoft/LightGBM/blob/b4b1b75ac582e99202062ec784269bd5b77f2018/python-package/lightgbm/sklearn.py#L358-L376

Issue #3830, opened 2021-01-24
Support DataTable in Dask


StrikerRUS:
## Summary

Dask estimators should support input in a form of H2O DataTable .
## Motivation

This change would bring the Dask interface closer to full feature parity with the non-Dask interface.

## Description

Initial step can be supporting DataTable via converting it into Numpy array.

## References

https://github.com/microsoft/LightGBM/pull/3515#discussion_r541797207
https://github.com/microsoft/LightGBM/blob/da44387176316632f126114182f6754a68a1d557/python-package/lightgbm/basic.py#L946-L948
https://github.com/microsoft/LightGBM/blob/da44387176316632f126114182f6754a68a1d557/python-package/lightgbm/basic.py#L619-L620

StrikerRUS:
Closed in favor of being in #2302. We decided to keep all feature requests in one place.

Welcome to contribute this feature! Please re-open this issue (or post a comment if you are not a topic starter) if you are actively working on implementing this feature.

jameslamb:
I'm -1 on this change. I believe that `.fit()` in the Dask module should only accept Dask collections (Dask DataFrame and Dask Array).

Type hints and type decisions for Dask (#3756) is my next priority, and in the PR for that I'll propose that we raise an error in `.fit()` if `X`, `y`, or `sample_weight` are not Dask collections. This is what XGBoost does as well, and I think it's a very good pattern: https://github.com/dmlc/xgboost/blob/a275f4026728ed14fbc70da142ef7a4a1d3de04d/python-package/xgboost/dask.py#L258-L263.

If we don't put in such limitatioons, `lightgbm` will have to take on responsibility for how to move a non-Dask input out to the Dask cluster. That will introduce a lot of maintenance for not a lot of gain to users. The task of taking a non-Dask input and turning it into a Dask collection is theoretically simple, but there isn't a single "right" way to do it and the best way can depend on the shape of your data and the nature of your task.  Consider https://github.com/dask/dask/issues/6833#issuecomment-726517622 and the rest of the discussion on that issue and the linked ones for some examples of how this can be a rough part of the Dask experience.

[...]

Issue #3861, opened 2021-01-26
[dask] DaskRegressor.predict() fails on DataFrame / Series input


jameslamb:
## How you are using LightGBM?

LightGBM component: Python-package

## Environment info

Operating System: Ubuntu 18.04

C++ compiler version: gcc 8.3.0

CMake version: 3.13.4

Python version:

<details><summary>output of 'conda info'</summary>

```text
     active environment : saturn
    active env location : /opt/conda/envs/saturn
            shell level : 0
       user config file : /home/jovyan/.condarc
 populated config files : /opt/conda/.condarc
          conda version : 4.8.2
    conda-build version : not installed
         python version : 3.7.7.final.0
       virtual packages : __glibc=2.28
       base environment : /opt/conda  (writable)
           channel URLs : https://conda.saturncloud.io/pkgs/linux-64
                          https://conda.saturncloud.io/pkgs/noarch
                          https://conda.anaconda.org/conda-forge/linux-64
                          https://conda.anaconda.org/conda-forge/noarch
                          https://repo.anaconda.com/pkgs/main/linux-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/r/linux-64
                          https://repo.anaconda.com/pkgs/r/noarch
          package cache : /opt/conda/pkgs
                          /home/jovyan/.conda/pkgs
       envs directories : /opt/conda/envs
                          /home/jovyan/.conda/envs
               platform : linux-64
             user-agent : conda/4.8.2 requests/2.22.0 CPython/3.7.7 Linux/4.14.203-156.332.amzn2.x86_64 debian/10 glibc/2.28
                UID:GID : 1000:100
             netrc file : None
           offline mode : False
```

</details>

LightGBM version or commit hash: https://github.com/microsoft/LightGBM/tree/9f70e9685dfb5c82f2ee87176a8433a6b7a4b98f

## Error message and / or logs

Training with `lightgbm.dask.DaskLGBMRegressor` succeeds, and `.predict()` fails with this error.

```text
ValueError: Metadata inference failed in `_predict_part`.

You have supplied a custom function and Dask is unable to 
determine the type of output that that function returns. 

To resolve this please provide a meta= keyword.
The docstring of the Dask function you ran should have more information.

Original error is below:
------------------------
TypeError('Unknown type of parameter:y, got:Series')

Traceback:
---------
[...]

Issue #3829, opened 2021-01-24
[dask] Random failures in Dask tests during teardown


StrikerRUS:
Need to make Dask tests less flaky and ideally remove all randomness from them.
Log from one of failed job:
```
2021-01-24T00:16:23.5034764Z ============================= test session starts ==============================
2021-01-24T00:16:23.5036640Z platform linux -- Python 3.8.5, pytest-6.2.1, py-1.10.0, pluggy-0.13.1
2021-01-24T00:16:23.5037201Z rootdir: /__w/1/s
2021-01-24T00:16:23.5037554Z collected 248 items
2021-01-24T00:16:23.5037809Z 
2021-01-24T00:16:23.8406259Z ../tests/python_package_test/test_basic.py ............                  [  4%]
2021-01-24T00:16:32.3036057Z ../tests/python_package_test/test_consistency.py ......                  [  7%]
2021-01-24T00:18:49.4493918Z ../tests/python_package_test/test_dask.py .........................E.... [ 18%]
2021-01-24T00:19:32.8918399Z ...........                                                              [ 23%]
2021-01-24T00:19:32.8930118Z ../tests/python_package_test/test_dual.py s                              [ 23%]
2021-01-24T00:19:36.0702660Z ../tests/python_package_test/test_engine.py ............................ [ 35%]
2021-01-24T00:20:19.2796809Z ........................................                                 [ 51%]
2021-01-24T00:20:20.0033215Z ../tests/python_package_test/test_plotting.py .....                      [ 53%]
2021-01-24T00:20:21.7294635Z ../tests/python_package_test/test_sklearn.py ........................... [ 64%]
2021-01-24T00:20:24.6758838Z ......x.........................................x....................... [ 93%]
2021-01-24T00:20:24.8587472Z .................                                                        [100%]
2021-01-24T00:20:24.8588302Z 
2021-01-24T00:20:24.8590837Z ==================================== ERRORS ====================================
2021-01-24T00:20:24.8593230Z ______ ERROR at teardown of test_regressor_quantile[0.5-scipy_csr_matrix] ______
2021-01-24T00:20:24.8595011Z 
2021-01-24T00:20:24.8595723Z loop = <tornado.platform.asyncio.AsyncIOLoop object at 0x7f41dc8e2070>
2021-01-24T00:20:24.8596248Z 
2021-01-24T00:20:24.8596718Z     @pytest.fixture
2021-01-24T00:20:24.8597247Z     def cluster_fixture(loop):
[...]

PR #3866, opened 2021-01-26
[dask] Add type hints in Dask package

jameslamb:
As part of #3756 , this PR proposes type hints for the Dask package.

This PR attempts to document the current state of the package, and isn't a proposal about the final state for release 3.2.0. So, for example, right now `X` and `y` in `.predict()` have to be Dask collections, but I could see it being useful to allow other data types in the future. I'd like to defer that discussion to other issues / PRs.

### Notes for Reviewers

* I'm pushing this on the `docs/jlamb` branch because that branch already has readthedocs build enabled, and I want to see what effect, if any, this change has on the documentation
* I want to add `isinstance()` checks on the types of arguments to `fit()` and `predict()` that should be Dask collections, as mentioned in https://github.com/microsoft/LightGBM/issues/3830#issuecomment-766285843, but I'm proposing that that be a separate PR
* It might be weird to see that the hint for `**kwargs` is not `Dict[str, Any]`. I've learned that the hint for `*` and `**` stuff is expected to be the set of allowable types for the VALUES. https://www.python.org/dev/peps/pep-0484/#arbitrary-argument-lists-and-default-argument-values
* The annotation `Type[]` in a hint says "this should be a class, not an instance". So `df_class: Type[pd.DataFrame]` means you pass `pd.DataFrame` or a sub-class, where `df: pd.DataFrame` means a constructed data frame is expected.

I ran `mypy python-package/lightgbm/dask.py`. There are still some `mypy` issues but I don't think any are a direct result of this PR and some come from LightGBM's dependencies. mypy can also get easily confused when you use mixin classes, which we do a lot in the Dask package and the sklearn classes it imports. So I'm not worrying too much about the mypy stuff below right now (other than the one I fixed in #3865 ).

<details><summary>mypy results for the curious</summary>

```text
python-package/lightgbm/dask.py:15: error: Skipping analyzing 'numpy': found module but no type hints or library stubs
python-package/lightgbm/dask.py:15: note: See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports
python-package/lightgbm/dask.py:16: error: Skipping analyzing 'scipy.sparse': found module but no type hints or library stubs
python-package/lightgbm/dask.py:16: error: Skipping analyzing 'scipy': found module but no type hints or library stubs
python-package/lightgbm/dask.py:71: error: Incompatible return value type (got "Optional[int]", expected "int")
python-package/lightgbm/dask.py:94: error: Need type annotation for 'lightgbm_ports' (hint: "lightgbm_ports: Set[<type>] = ...")
python-package/lightgbm/dask.py:111: error: Value of type "Iterable[Union[Any, Any, Any, Any]]" is not indexable
python-package/lightgbm/dask.py:113: error: Value of type "Iterable[Union[Any, Any, Any, Any]]" is not indexable
python-package/lightgbm/dask.py:115: error: Value of type "Iterable[Union[Any, Any, Any, Any]]" is not indexable
python-package/lightgbm/dask.py:118: error: Value of type "Iterable[Union[Any, Any, Any, Any]]" is not indexable
python-package/lightgbm/dask.py:246: error: Name 'tree_learner' is not defined
python-package/lightgbm/dask.py:285: error: "Dict[str, Any]" has no attribute "status"
python-package/lightgbm/dask.py:286: error: Incompatible return value type (got "Dict[str, Any]", expected "LGBMModel")
python-package/lightgbm/dask.py:289: error: "Dict[str, Any]" has no attribute "key"; maybe "keys"?
python-package/lightgbm/dask.py:455: error: "_DaskLGBMModel" has no attribute "get_params"
python-package/lightgbm/dask.py:468: error: "_DaskLGBMModel" has no attribute "set_params"
python-package/lightgbm/dask.py:474: error: "_DaskLGBMModel" has no attribute "get_params"
python-package/lightgbm/dask.py:480: error: "_DaskLGBMModel" has no attribute "get_params"
python-package/lightgbm/dask.py:490: error: Signature of "fit" incompatible with supertype "LGBMClassifier"
python-package/lightgbm/dask.py:490: error: Signature of "fit" incompatible with supertype "LGBMModel"
python-package/lightgbm/dask.py:499: error: Incompatible return value type (got "_DaskLGBMModel", expected "DaskLGBMClassifier")
python-package/lightgbm/dask.py:509: error: Item "None" of "Optional[str]" has no attribute "partition"
python-package/lightgbm/dask.py:515: error: Signature of "predict" incompatible with supertype "LGBMClassifier"
python-package/lightgbm/dask.py:515: error: Signature of "predict" incompatible with supertype "LGBMModel"
python-package/lightgbm/dask.py:526: error: Signature of "predict_proba" incompatible with supertype "LGBMClassifier"
python-package/lightgbm/dask.py:550: error: Signature of "fit" incompatible with supertype "LGBMRegressor"
python-package/lightgbm/dask.py:550: error: Signature of "fit" incompatible with supertype "LGBMModel"
python-package/lightgbm/dask.py:559: error: Incompatible return value type (got "_DaskLGBMModel", expected "DaskLGBMRegressor")
python-package/lightgbm/dask.py:569: error: Item "None" of "Optional[str]" has no attribute "partition"
python-package/lightgbm/dask.py:575: error: Signature of "predict" incompatible with supertype "LGBMModel"
python-package/lightgbm/dask.py:598: error: Signature of "fit" incompatible with supertype "LGBMRanker"
python-package/lightgbm/dask.py:598: error: Signature of "fit" incompatible with supertype "LGBMModel"
python-package/lightgbm/dask.py:612: error: Incompatible return value type (got "_DaskLGBMModel", expected "DaskLGBMRanker")
python-package/lightgbm/dask.py:623: error: Item "None" of "Optional[str]" has no attribute "partition"
python-package/lightgbm/dask.py:629: error: Signature of "predict" incompatible with supertype "LGBMModel"
```
</details>


jameslamb:
@ffineis if you have time, could you take a look at this diff too? Since you've recently worked with the internals of the Dask module

ffineis:
> @ffineis if you have time, could you take a look at this diff too? Since you've recently worked with the internals of the Dask module

Yep on it tonight

python-package/lightgbm/dask.py
```diff
@@ -169,7 +181,16 @@ def _split_to_parts(data, is_matrix):
     return parts
 
 
-def _train(client, data, label, params, model_factory, sample_weight=None, group=None, **kwargs):
+def _train(
+    client: Client,
+    data: _DaskCollection,
+    label:_DaskCollection,
```
jameslamb, line 187:
```suggestion
    label: _DaskCollection,
```

ffineis (changes requested):

python-package/lightgbm/dask.py
```diff
@@ -535,7 +596,16 @@ def to_local(self):
...
     """Distributed version of lightgbm.LGBMRanker."""
 
-    def fit(self, X, y, sample_weight=None, init_score=None, group=None, client=None, **kwargs):
+    def fit(
+        self,
+        X: _DaskCollection,
+        y: _DaskCollection,
+        sample_weight: Optional[_DaskCollection] = None,
+        init_score: Optional[_DaskCollection] = None,
+        group: Optional[_1DArrayLike] = None,
```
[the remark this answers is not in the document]
jameslamb: oooooo ok! I misunderstood what was happening. Awesome, I'll look again and then change that hint.

I'll also write up a feature request for this. I think it's something that's non-breaking and additive, that could be done after 3.2.0. But like you said, since `group` is often fairly small, I think it could be a nice thing to be able to specify it as a list or lil numpy array.

python-package/lightgbm/dask.py
```diff
@@ -169,7 +181,16 @@ def _split_to_parts(data, is_matrix):
...
 
-def _train(client, data, label, params, model_factory, sample_weight=None, group=None, **kwargs):
+def _train(
+    client: Client,
+    data: _DaskCollection,
+    label: _DaskCollection,
+    params: Dict[str, Any],
+    model_factory: Type[LGBMModel],
+    sample_weight: Optional[_DaskCollection] = None,
+    group: Optional[_1DArrayLike] = None,
```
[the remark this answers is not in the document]
jameslamb: I'll update the docs in this PR too, I think that's directly related to this scope. Thanks!

python-package/lightgbm/dask.py
```diff
@@ -422,13 +469,13 @@ def _fit(self, model_factory, X, y, sample_weight=None, group=None, client=None,
...
 
-    def _to_local(self, model_factory):
+    def _to_local(self, model_factory: Type[LGBMModel]) -> LGBMModel:
         model = model_factory(**self.get_params())
         self._copy_extra_params(self, model)
         return model
 
     @staticmethod
-    def _copy_extra_params(source, dest):
+    def _copy_extra_params(source: "_DaskLGBMModel", dest: "_DaskLGBMModel") -> None:
```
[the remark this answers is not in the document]
jameslamb: shoot, you're totally right. good eye

python-package/lightgbm/dask.py
```diff
@@ -102,7 +107,7 @@ def _find_ports_for_workers(client: Client, worker_addresses: Iterable[str], loc
     return worker_ip_to_port
 
 
-def _concat(seq):
+def _concat(seq: Iterable[_DaskPart]) -> _DaskPart:
```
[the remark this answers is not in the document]
jameslamb: I didn't have any special reason for choosing `Iterable` instead of `List`. I was just looking at the code and didn't see any list-specific stuff so I thought it made sense to use the more general thing.

But since this is a totally-internally function, where we control the input, and since `Iterable` does weird stuff as that S/O post points out, I'll change this to `List[]`

python-package/lightgbm/dask.py
```diff
@@ -334,8 +363,16 @@ def _predict_part(part, model, raw_score, pred_proba, pred_leaf, pred_contrib, *
     return result
 
 
-def _predict(model, data, raw_score=False, pred_proba=False, pred_leaf=False, pred_contrib=False,
-             dtype=np.float32, **kwargs):
+def _predict(
+    model: LGBMModel,
+    data: _DaskCollection,
```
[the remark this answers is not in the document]
jameslamb: yup will do in this PR, thanks!

python-package/lightgbm/dask.py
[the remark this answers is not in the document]
jameslamb: Thinking through this, I realized that with the way we do doc inheritance, it will actually be a bit tricky to override the docs for `group`.

I wrote up my thoughts in https://github.com/microsoft/LightGBM/issues/3871, but I won't update the docs (except those on internal functions) in this PR

python-package/lightgbm/dask.py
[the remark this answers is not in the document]
jameslamb: fixed in https://github.com/microsoft/LightGBM/pull/3866/commits/bfd9dc0a7438a45901afe916c095bfb818b1dba9

python-package/lightgbm/dask.py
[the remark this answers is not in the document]
jameslamb: fixed in https://github.com/microsoft/LightGBM/pull/3866/commits/bfd9dc0a7438a45901afe916c095bfb818b1dba9

python-package/lightgbm/dask.py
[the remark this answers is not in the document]
jameslamb: fixed in https://github.com/microsoft/LightGBM/pull/3866/commits/bfd9dc0a7438a45901afe916c095bfb818b1dba9

python-package/lightgbm/dask.py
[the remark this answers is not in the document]
jameslamb: fixed the hint in https://docs.dask.org/en/latest/dataframe-api.html#dask.dataframe.to_csv

python-package/lightgbm/dask.py
```diff
@@ -333,15 +361,23 @@ def _predict_part(part, model, raw_score, pred_proba, pred_leaf, pred_contrib, *
...
+    raw_score: bool = False,
+    pred_proba: bool = False,
+    pred_leaf: bool = False,
+    pred_contrib: bool = False,
+    dtype: _PredictionDtype = np.float32,
+    **kwargs: Any
+) -> _DaskCollection:
```
jameslamb, line 373:
```suggestion
) -> dask_Array:
```

Based on https://github.com/microsoft/LightGBM/issues/3861#issuecomment-767970848, I realize now that this will always be a Dask array. That's the same behavior as the scikit-learn API.

```python
import numpy as np
import pandas as pd
from lightgbm import LGBMRegressor

reg = LGBMRegressor()

np.random.random(10)

num_features = 20
num_rows = 1000

X = pd.DataFrame({
    "col" + str(i): np.random.random(num_rows)
    for i in range(num_features)
})

y = np.random.random(num_rows)
reg.fit(X, y)

preds = reg.predict(X)

print(f"input type: {type(X)}, \npred type: {type(preds)}")
```

StrikerRUS:
@jameslamb 
Could you please replace all occurrences of `dask array` with `dask Array` fort the consistency with `dask DataFrame, or dask Series`?

jameslamb:
> @jameslamb
> Could you please replace all occurrences of `dask array` with `dask Array` for the consistency with `dask DataFrame, or dask Series`?

oh sure, no problem. done in https://github.com/microsoft/LightGBM/pull/3866/commits/b20ac3782639fc9f0d0c1e7aa9d0e7a543ca8d5a

jameslamb:
The type hints are currently shown in method signatures on the docs site:

image

from https://lightgbm.readthedocs.io/en/docs-jlamb/pythonapi/lightgbm.DaskLGBMClassifier.html#lightgbm.DaskLGBMClassifier.fit

I think this makes the method signature really hard to read. I'm going to try disabling this by setting `autodoc_typehints = "none"` in `conf.py` (https://stackoverflow.com/a/63560720/3986677).

StrikerRUS:
> I'm going to try disabling this by setting `autodoc_typehints = "none"` in `conf.py`

Don't forget to bump minimum required Sphinx version then

https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_typehints

https://github.com/microsoft/LightGBM/blob/b4b1b75ac582e99202062ec784269bd5b77f2018/docs/conf.py#L77

jameslamb:
> > I'm going to try disabling this by setting `autodoc_typehints = "none"` in `conf.py`
> 
> Don't forget to bump minimum required Sphinx version then
> 
> https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_typehints
> 
> https://github.com/microsoft/LightGBM/blob/b4b1b75ac582e99202062ec784269bd5b77f2018/docs/conf.py#L77

Oh thanks! Didn't know about that. I though the version was only managed in https://github.com/microsoft/LightGBM/blob/b4b1b75ac582e99202062ec784269bd5b77f2018/docs/requirements_rtd.txt#L2

I'll update that. But happy to say the builds on RTD did pass, and the type hints were successfully hidden: https://lightgbm.readthedocs.io/en/docs-jlamb/pythonapi/lightgbm.DaskLGBMClassifier.html#lightgbm.DaskLGBMClassifier

image

python-package/lightgbm/dask.py
```diff
@@ -169,24 +180,33 @@ def _split_to_parts(data, is_matrix):
     return parts
 
 
-def _train(client, data, label, params, model_factory, sample_weight=None, group=None, **kwargs):
+def _train(
+    client: Client,
+    data: _DaskCollection,
```
StrikerRUS, line 185: Is it OK that `_DaskCollection = Union[dask_Array, dask_Frame]` but `data` can be only `dask DataFrame`, but not `dask Series`?

python-package/lightgbm/dask.py
```diff
@@ -169,24 +180,33 @@ def _split_to_parts(data, is_matrix):
...
         Dask client.
-    data : dask array of shape = [n_samples, n_features]
+    data : dask Array or dask DataFrame of shape = [n_samples, n_features]
         Input feature matrix.
-    label : dask array of shape = [n_samples]
+    label : dask Array, dask DataFrame, or dask Series of shape = [n_samples]
         The target values (class labels in classification, real numbers in regression).
     params : dict
         Parameters passed to constructor of the local underlying model.
     model_factory : lightgbm.LGBMClassifier, lightgbm.LGBMRegressor, or lightgbm.LGBMRanker class
```
StrikerRUS, line 205: Maybe simplify?
```suggestion
    model_factory : subclass of lightgbm.LGBMModel
```
And here (I can't add suggestion for that line):
https://github.com/microsoft/LightGBM/blob/c4b51317f8f722f60105335e5c799877a7dd80bd/python-package/lightgbm/dask.py#L200

python-package/lightgbm/dask.py
```diff
@@ -169,24 +180,33 @@ def _split_to_parts(data, is_matrix):
...
 
     Parameters
     ----------
     client : dask.distributed.Client
         Dask client.
-    data : dask array of shape = [n_samples, n_features]
+    data : dask Array or dask DataFrame of shape = [n_samples, n_features]
         Input feature matrix.
-    label : dask array of shape = [n_samples]
+    label : dask Array, dask DataFrame, or dask Series of shape = [n_samples]
```
StrikerRUS, line 201:
```suggestion
    label : dask Array, dask DataFrame or dask Series of shape = [n_samples]
```

python-package/lightgbm/dask.py
```diff
@@ -169,24 +180,33 @@ def _split_to_parts(data, is_matrix):
...
         The target values (class labels in classification, real numbers in regression).
     params : dict
         Parameters passed to constructor of the local underlying model.
     model_factory : lightgbm.LGBMClassifier, lightgbm.LGBMRegressor, or lightgbm.LGBMRanker class
         Class of the local underlying model.
-    sample_weight : array-like of shape = [n_samples] or None, optional (default=None)
+    sample_weight : dask Array or dask DataFrame or Dask Series or None of shape = [n_samples] or None, optional (default=None)
```
StrikerRUS, line 207:
```suggestion
    sample_weight : dask Array, dask DataFrame, Dask Series of shape = [n_samples] or None, optional (default=None)
```

python-package/lightgbm/dask.py
```diff
@@ -169,24 +180,33 @@ def _split_to_parts(data, is_matrix):
...
         Weights of training data.
-    group : array-like or None, optional (default=None)
+    group : dask Array or dask DataFrame or Dask Series of shape = [n_samples] or None, optional (default=None)
```
StrikerRUS, line 209:
```suggestion
    group : dask Array, dask DataFrame, Dask Series of shape = [n_samples] or None, optional (default=None)
```

python-package/lightgbm/dask.py
```diff
@@ -333,15 +361,23 @@ def _predict_part(part, model, raw_score, pred_proba, pred_leaf, pred_contrib, *
...
+    pred_leaf: bool = False,
+    pred_contrib: bool = False,
+    dtype: _PredictionDtype = np.float32,
+    **kwargs: Any
+) -> _DaskCollection:
     """Inner predict routine.
 
     Parameters
     ----------
     model : lightgbm.LGBMClassifier, lightgbm.LGBMRegressor, or lightgbm.LGBMRanker class
```
StrikerRUS, line 378:
```suggestion
    model : subclass of lightgbm.LGBMModel
```

python-package/lightgbm/dask.py
```diff
@@ -333,15 +361,23 @@ def _predict_part(part, model, raw_score, pred_proba, pred_leaf, pred_contrib, *
     return result
 
 
-def _predict(model, data, raw_score=False, pred_proba=False, pred_leaf=False, pred_contrib=False,
-             dtype=np.float32, **kwargs):
+def _predict(
+    model: LGBMModel,
+    data: _DaskCollection,
```
StrikerRUS, line 366: Same about inability to pass `dask Series` as above.

python-package/lightgbm/dask.py
```diff
@@ -358,11 +394,11 @@ def _predict(model, data, raw_score=False, pred_proba=False, pred_leaf=False, pr
...
     Returns
     -------
-    predicted_result : dask array of shape = [n_samples] or shape = [n_samples, n_classes]
+    predicted_result : dask Array of shape = [n_samples] or shape = [n_samples, n_classes]
         The predicted values.
-    X_leaves : dask array of shape = [n_samples, n_trees] or shape = [n_samples, n_trees * n_classes]
+    X_leaves : dask Array of shape = [n_samples, n_trees] or shape = [n_samples, n_trees * n_classes]
         If ``pred_leaf=True``, the predicted leaf of every tree for each sample.
-    X_SHAP_values : dask array of shape = [n_samples, n_features + 1] or shape = [n_samples, (n_features + 1) * n_classes] or list with n_classes length of such objects
+    X_SHAP_values : dask Array of shape = [n_samples, n_features + 1] or shape = [n_samples, (n_features + 1) * n_classes] or list with n_classes length of such objects
```
StrikerRUS, line 401: > ... or list with n_classes length of such objects

Is it supported?

python-package/lightgbm/dask.py
```diff
@@ -393,11 +429,21 @@ def _predict(model, data, raw_score=False, pred_proba=False, pred_leaf=False, pr
             **kwargs
         )
     else:
-        raise TypeError('Data must be either Dask array or dataframe. Got %s.' % str(type(data)))
+        raise TypeError('Data must be either dask Array or dataframe. Got %s.' % str(type(data)))
```
StrikerRUS, line 432:
```suggestion
        raise TypeError('Data must be either dask Array or dask DataFrame. Got %s.' % str(type(data)))
```

python-package/lightgbm/dask.py
```diff
@@ -393,11 +429,21 @@ def _predict(model, data, raw_score=False, pred_proba=False, pred_leaf=False, pr
...
 
 
 class _DaskLGBMModel:
-    def _fit(self, model_factory, X, y, sample_weight=None, group=None, client=None, **kwargs):
+
+    def _fit(
+        self,
+        model_factory: Type[LGBMModel],
+        X: _DaskCollection,
```
StrikerRUS, line 440: Same about inability to pass `dask Series` as above.

python-package/lightgbm/dask.py
```diff
@@ -438,7 +484,14 @@ def _copy_extra_params(source, dest):
 class DaskLGBMClassifier(LGBMClassifier, _DaskLGBMModel):
     """Distributed version of lightgbm.LGBMClassifier."""
 
-    def fit(self, X, y, sample_weight=None, client=None, **kwargs):
+    def fit(
+        self,
+        X: _DaskCollection,
```
StrikerRUS, line 489: Same about inability to pass `dask Series` as above.

python-package/lightgbm/dask.py
```diff
@@ -456,7 +509,7 @@ def fit(self, X, y, sample_weight=None, client=None, **kwargs):
                    + ' ' * 12 + 'Dask client.\n'
                    + ' ' * 8 + _init_score + _after_init_score)
 
-    def predict(self, X, **kwargs):
+    def predict(self, X: _DaskCollection, **kwargs: Any) -> _DaskCollection:
```
StrikerRUS, line 512: Based on https://github.com/microsoft/LightGBM/pull/3866/files#r565691112.
```suggestion
    def predict(self, X: _DaskCollection, **kwargs: Any) -> dask_Array:
```

StrikerRUS, line 512: Same about inability to pass `dask Series` as above.

python-package/lightgbm/dask.py
```diff
@@ -467,7 +520,7 @@ def predict(self, X, **kwargs):
 
     predict.__doc__ = LGBMClassifier.predict.__doc__
 
-    def predict_proba(self, X, **kwargs):
+    def predict_proba(self, X: _DaskCollection, **kwargs: Any) -> _DaskCollection:
```
StrikerRUS, line 523: Same about inability to pass `dask Series` as above.

StrikerRUS, line 523: Based on https://github.com/microsoft/LightGBM/pull/3866/files#r565691112.
```suggestion
    def predict_proba(self, X: _DaskCollection, **kwargs: Any) -> dask_Array:
```

python-package/lightgbm/dask.py
```diff
@@ -492,7 +545,14 @@ def to_local(self):
 class DaskLGBMRegressor(LGBMRegressor, _DaskLGBMModel):
     """Distributed version of lightgbm.LGBMRegressor."""
 
-    def fit(self, X, y, sample_weight=None, client=None, **kwargs):
+    def fit(
+        self,
+        X: _DaskCollection,
```
StrikerRUS, line 550: Same about inability to pass `dask Series` as above.

python-package/lightgbm/dask.py
```diff
@@ -510,7 +570,7 @@ def fit(self, X, y, sample_weight=None, client=None, **kwargs):
                    + ' ' * 12 + 'Dask client.\n'
                    + ' ' * 8 + _init_score + _after_init_score)
 
-    def predict(self, X, **kwargs):
+    def predict(self, X: _DaskCollection, **kwargs) -> _DaskCollection:
```
StrikerRUS, line 573: Same about inability to pass `dask Series` as above.

StrikerRUS, line 573: Based on https://github.com/microsoft/LightGBM/pull/3866/files#r565691112.
```suggestion
    def predict(self, X: _DaskCollection, **kwargs) -> dask_Array:
```

python-package/lightgbm/dask.py
```diff
@@ -534,7 +594,16 @@ def to_local(self):
 class DaskLGBMRanker(LGBMRanker, _DaskLGBMModel):
     """Distributed version of lightgbm.LGBMRanker."""
 
-    def fit(self, X, y, sample_weight=None, init_score=None, group=None, client=None, **kwargs):
+    def fit(
+        self,
+        X: _DaskCollection,
```
StrikerRUS, line 599: Same about inability to pass `dask Series` as above.

python-package/lightgbm/dask.py
```diff
@@ -556,13 +625,13 @@ def fit(self, X, y, sample_weight=None, init_score=None, group=None, client=None
                    + ' ' * 12 + 'Dask client.\n'
                    + ' ' * 8 + _eval_set + _after_eval_set)
 
-    def predict(self, X, **kwargs):
+    def predict(self, X: _DaskCollection, **kwargs: Any) -> _DaskCollection:
```
StrikerRUS, line 628: Same about inability to pass `dask Series` as above.

StrikerRUS, line 628: Based on https://github.com/microsoft/LightGBM/pull/3866/files#r565691112.
```suggestion
    def predict(self, X: _DaskCollection, **kwargs: Any) -> dask_Array:
```

StrikerRUS (commented):
Not sure my comments will be useful, but anyway: 🙂

python-package/lightgbm/dask.py
jameslamb: you're absolutely right, thank you. I'll make a new type `_DaskArrayLike` for all of these

python-package/lightgbm/dask.py
jameslamb: That would be less words, but I disagree that it would be "simpler". I think listing the three classes is easier to understand than saying "subclass of LGBMModel", especially since outside contributors might not know that `LGBMClassifier` (for example) is a child of `LGBMModel` without looking it up first.

Since I'm not aware of any plan to add a new, 4th child of `LGBMModel` any time soon, I'd rather keep this as-is.

python-package/lightgbm/dask.py
jameslamb: it looks like `dask-lightgbm` probably just copied that from https://github.com/microsoft/LightGBM/blob/c4b51317f8f722f60105335e5c799877a7dd80bd/python-package/lightgbm/sklearn.py#L691.

How does the sklearn model object produce an output like that?

StrikerRUS: Yes, I simply copied this line.

It comes from here:
https://github.com/microsoft/LightGBM/blob/c4b51317f8f722f60105335e5c799877a7dd80bd/python-package/lightgbm/basic.py#L678
https://github.com/microsoft/LightGBM/blob/c4b51317f8f722f60105335e5c799877a7dd80bd/python-package/lightgbm/basic.py#L805-L849
Contributed in #3000.

python-package/lightgbm/dask.py
jameslamb: fixed in https://github.com/microsoft/LightGBM/pull/3866/commits/8b19e04b05f83a38f5213c1103040f49d0a20988

python-package/lightgbm/dask.py
jameslamb: fixed in https://github.com/microsoft/LightGBM/pull/3866/commits/8b19e04b05f83a38f5213c1103040f49d0a20988

python-package/lightgbm/dask.py
jameslamb: fixed in https://github.com/microsoft/LightGBM/pull/3866/commits/8b19e04b05f83a38f5213c1103040f49d0a20988

python-package/lightgbm/dask.py
jameslamb: fixed in https://github.com/microsoft/LightGBM/pull/3866/commits/8b19e04b05f83a38f5213c1103040f49d0a20988

python-package/lightgbm/dask.py
jameslamb: fixed in https://github.com/microsoft/LightGBM/pull/3866/commits/8b19e04b05f83a38f5213c1103040f49d0a20988

python-package/lightgbm/dask.py
jameslamb: fixed in https://github.com/microsoft/LightGBM/pull/3866/commits/8b19e04b05f83a38f5213c1103040f49d0a20988

python-package/lightgbm/dask.py
jameslamb: fixed in https://github.com/microsoft/LightGBM/pull/3866/commits/8b19e04b05f83a38f5213c1103040f49d0a20988

python-package/lightgbm/dask.py
jameslamb: fixed in https://github.com/microsoft/LightGBM/pull/3866/commits/8b19e04b05f83a38f5213c1103040f49d0a20988

python-package/lightgbm/dask.py
jameslamb: fixed in https://github.com/microsoft/LightGBM/pull/3866/commits/8b19e04b05f83a38f5213c1103040f49d0a20988

python-package/lightgbm/dask.py
jameslamb: I decided on `_DaskMatrixLike` to reduce confusion. This change also means we don't need `dask_Frame` any more.

Added in https://github.com/microsoft/LightGBM/pull/3866/commits/8b19e04b05f83a38f5213c1103040f49d0a20988

python-package/lightgbm/dask.py
jameslamb: Since this is an internal function though, I don't feel strongly. So if you want me to change it I will.

StrikerRUS: Nope, I'm fine with your reasoning. Thanks!

python-package/lightgbm/dask.py
```diff
@@ -332,15 +361,23 @@ def _predict_part(part, model, raw_score, pred_proba, pred_leaf, pred_contrib, *
...
+def _predict(
+    model: LGBMModel,
+    data: _DaskMatrixLike,
+    raw_score: bool = False,
+    pred_proba: bool = False,
+    pred_leaf: bool = False,
+    pred_contrib: bool = False,
+    dtype: _PredictionDtype = np.float32,
+    **kwargs: Any
+) -> _DaskCollection:
```
StrikerRUS, line 373:
```suggestion
) -> dask_Array:
```

python-package/lightgbm/dask.py
```diff
@@ -466,7 +520,7 @@ def predict(self, X, **kwargs):
 
     predict.__doc__ = LGBMClassifier.predict.__doc__
 
-    def predict_proba(self, X, **kwargs):
+    def predict_proba(self, X: _DaskCollection, **kwargs: Any) -> dask_Array:
```
StrikerRUS, line 523:
```suggestion
    def predict_proba(self, X: _DaskMatrixLike, **kwargs: Any) -> dask_Array:
```

StrikerRUS (approved):
LGTM except two comment below and one unresolved question from previous review: https://github.com/microsoft/LightGBM/pull/3866#discussion_r566106219.

Thanks a lot!

python-package/lightgbm/dask.py
jameslamb: blegh, thought I got them all

python-package/lightgbm/dask.py
jameslamb: From that PR, it looks like this is specific to sparse inputs? I'll have to test that with a sparse Dask Array. Can try that later tonight and report back.

StrikerRUS: Here are corresponding tests you can refer to (some were added later in separate RPs):
https://github.com/microsoft/LightGBM/blob/42d1633aebe124821cff42c728a42551db715168/tests/python_package_test/test_engine.py#L1025-L1123

jameslamb: Thanks for the links!

Ok, I tested some stuff tonight. I found that right now, if you provide a Dask Array whose pieces are CSC matrices to `.predict(pred_contrib = TRUE)`, you get back a Dask Array whose individual partitions are also CSC matrices. Calling `.compute()` on that Dask array returns a `scipy.sparse.coo.coo_matrix`.

That returned matrix does have the expected dimensions. For an input of shape (100, 2), for `n_classes=3`, you get back a matrix of shape `(100, 9)`.

So I think the "or list with n_classes..." part should be removed from this docstring. What do you think?

```python
import dask.array as da
import numpy as np
from lightgbm.dask import DaskLGBMClassifier
from lightgbm.sklearn import LGBMClassifier
from scipy.sparse import csc_matrix
from sklearn.datasets import make_blobs

n_workers = 3
client = Client(cluster)

chunk_size=50
X, y = make_blobs(n_samples=100, centers=3, random_state=42)
rnd = np.random.RandomState(42)
dX = da.from_array(X, chunks=(chunk_size, X.shape[1])).map_blocks(csc_matrix)
dy = da.from_array(y, chunks=chunk_size)

#  Dask predict() creates a Dask Array, whose `.compute()` is a 'scipy.sparse.coo.coo_matrix'
dask_clf = DaskLGBMClassifier(n_estimators=5, num_leaves=2, tree_learner="data")
dask_clf.fit(client, dX, dY)

preds = dask_clf.predict(dX, pred_contrib=True)
preds_computed = preds.compute()

print(type(preds), type(preds_computed), f"{dask_clf.n_classes_} classes, {dX.shape[1]} features")
print("---")
print(dX.partitions[0].compute())
print("---")
preds.compute().shape

# local predict() returns a list of 3 CSC matrices
X = dX.compute()
y = dy.compute()

local_clf = LGBMClassifier()
local_clf.fit(X=dX.compute(), y=y)
local_preds = local_clf.predict(dX.compute().tocsc(), pred_contrib=True)

print(local_clf.n_classes_, type(local_preds))
print("---")
print(local_preds)
```

jameslamb: I made this change in https://github.com/microsoft/LightGBM/pull/3866/commits/7b3a0e569df5c5db0f6454e9c85f313cec35889c, but won't merge until you have a chance to comment

StrikerRUS:
@jameslamb 
Well, I see this branch is in sync with `master` and there is a Dask teardown error in `regular` job.
https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=8861&view=logs&j=c28dceab-947a-5848-c21f-eef3695e5f11&t=fa158246-17e2-53d4-5936-86070edbaacf

image

I'm not re-running the failed job to let you collect all details from log you may want to get familiar.

According to https://github.com/microsoft/LightGBM/issues/3829#issuecomment-769412131 I'm going to increase `timeout` value for Dask tests.

python-package/lightgbm/dask.py
StrikerRUS: I'm sorry, I don't have an easy access to Dask environment where I can run tests with LightGBM due to that Windows isn't supported right now. So, I'm OK with whatever you decide. By my initial comment I just wanted to bring to your attention that there is a special case of return type for sparse matrices in the underlying API.

I just wonder where Dask merges `n_classes` sparse matrices into 1 like here
https://github.com/microsoft/LightGBM/blob/42d1633aebe124821cff42c728a42551db715168/tests/python_package_test/test_engine.py#L1083-L1086

Because 
> ```#  Dask predict() creates a Dask Array, whose `.compute()` is a 'scipy.sparse.coo.coo_matrix'```

`coo_matrix` should be casted to `csr` one here
https://github.com/microsoft/LightGBM/blob/c4b51317f8f722f60105335e5c799877a7dd80bd/python-package/lightgbm/basic.py#L721-L727
and inner reshaping should not be happen:
https://github.com/microsoft/LightGBM/blob/c4b51317f8f722f60105335e5c799877a7dd80bd/python-package/lightgbm/basic.py#L730-L736

jameslamb: thanks for pointing it out! I'm not sure why that's happening. I think having a return type of `List[array]` is going to be really challenging for Dask, but either way we can figure it out in a separate issue.

Recall that for this PR, the goal is

> This PR attempts to document the current state of the package, and isn't a proposal about the final state for release 3.2.0

And the current state is that if you pass a Dask Array whose parts are CSC matrices into `.predict(pred_contrib=True)`, you will get back a Dask Array.

I added a feature request to resolve the inconsistency. I think it's going to be really difficult to do, but can be figured out at a later time and doesn't need to block this PR: https://github.com/microsoft/LightGBM/issues/3881

jameslamb:
> @jameslamb
> Well, I see this branch is in sync with `master` and there is a Dask teardown error in `regular` job.
> https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=8861&view=logs&j=c28dceab-947a-5848-c21f-eef3695e5f11&t=fa158246-17e2-53d4-5936-86070edbaacf
> 
> image
> 
> I'm not re-running the failed job to let you collect all details from log you may want to get familiar.
> 
> According to [#3829 (comment)](https://github.com/microsoft/LightGBM/issues/3829#issuecomment-769412131) I'm going to increase `timeout` value for Dask tests.

Thanks. I moved the discussion over to #3829 and re-opened it.

I pushed an empty commit here to re-run CI, since I don't have rights to re-run jobs in Azure.

python-package/lightgbm/dask.py
StrikerRUS: Thank you! Agree.

jameslamb:
Ok since https://github.com/microsoft/LightGBM/pull/3866#discussion_r566106219 has been moved to #3881 , I think this PR can be merged. Thanks for the reviews!

Issue #3873, opened 2021-01-28
R-lintr failures at CI


StrikerRUS:
Our R-lintr CI is broken. I see this error in `master`, #3866 and #3872.

https://github.com/microsoft/LightGBM/runs/1783657796

```
Warning message:
package ‘lintr’ was built under R version 3.6.3 
[1] 49 R files need linting
Error in loadNamespace(j <- i[[1L]], c(lib.loc, .libPaths()), versionCheck = vI[[j]]) : 
  namespace ‘xfun’ 0.15 is being loaded, but >= 0.19 is required
Calls: <Anonymous> ... namespaceImportFrom -> asNamespace -> loadNamespace
Execution halted
```

StrikerRUS:
cc @jameslamb

jameslamb:
ok looking into this one now

jameslamb:
The most recent release of `r-lintr` on conda-forge was 8 months ago. https://anaconda.org/conda-forge/r-lintr/files

So one of its dependencies must have broken it :confused:

PR #3874, opened 2021-01-28
[ci] ignore CUDA-related strings in Python logger test

StrikerRUS:
Fix current CUDA workflow failures in `master` due to merge conflict.

PR #3876, opened 2021-01-28
[ci] fall back to source installs of R packages if binary missing (fixes #3875)

### LINKED ISSUE #3875: macOS CI R jobs fail
Can be observed in #3874.
Refer to https://github.com/microsoft/LightGBM/pull/3874/checks?check_run_id=1784106648

```
* checking tests ...
  Running ‘testthat.R’
  ERROR
Running the tests in ‘tests/testthat.R’ failed.
Complete output:
  > library(testthat)
  > library(lightgbm)
  Loading required package: R6
  > 
  > test_check(
  +     package = "lightgbm"
  +     , stop_on_failure = TRUE
  +     , stop_on_warning = FALSE
  + )
  Error in loadNamespace(name) : there is no package called 'withr'
  Calls: test_check ... loadNamespace -> withRestarts -> withOneRestart -> doWithOneRestart
  Execution halted
```

jameslamb:
I believe this will fix #3875. Right now in the Mac CI jobs for the R package, we set `type = "binary"` in `install.packages()`. This can lead to failures shortly after a new package needed in those jobs is uploaded to CRAN, when only source versions are available. Precompiled binaries from CRAN usually take 1-2 days to become available, in my experience.

the package `{withr}` was updated yesterday (https://cran.r-project.org/web/packages/withr/index.html), and I think the failures observed in #3875 happened because of this phenomenon.

This PR changes the strategy for installation on Mac to "both". From `?install.packages`

> An alternative (and the current default) is "both" which means ‘use binary if available and current, otherwise try source’. The action if there are source packages which are preferred but may contain code which needs to be compiled is controlled by getOption("install.packages.compile.from.source"). type = "both" will be silently changed to "binary" if either contriburl or available is specified.

So as of this PR, Mac builds will look for binaries (which are faster to install), but will install source packages from CRAN if the source package is newer.

This should prevent future issues like this from blocking LightGBM's CI.

### DIFF
--- a/.ci/test_r_package.sh
+++ b/.ci/test_r_package.sh
@@ -103,7 +103,7 @@ packages="c('data.table', 'jsonlite', 'Matrix', 'R6', 'testthat')"
 if [[ $OS_NAME == "macos" ]]; then
     packages+=", type = 'both'"
 fi
-Rscript --vanilla -e "options(install.packages.compile.from.source = 'both'); install.packages(${packages}, repos = '${CRAN_MIRROR}', lib = '${R_LIB_PATH}', dependencies = c('Depends', 'Imports', 'LinkingTo'))" || exit -1
+Rscript --vanilla -e "options(install.packages.compile.from.source = 'always'); install.packages(${packages}, repos = '${CRAN_MIRROR}', lib = '${R_LIB_PATH}', dependencies = c('Depends', 'Imports', 'LinkingTo'))" || exit -1
 
 cd ${BUILD_DIRECTORY}
 
--- a/.ci/test_r_package.sh
+++ b/.ci/test_r_package.sh
@@ -97,13 +97,18 @@ if [[ $OS_NAME == "macos" ]]; then
     fi
 fi
 
+# hack to fix https://github.com/microsoft/LightGBM/pull/3876#issuecomment-769187766
+if [[ $OS_NAME == "macos" ]]; then
+    Rscript --vanilla -e "install.packages('withr', repos = '${CRAN_MIRROR}', type = 'source', lib = '${R_LIB_PATH}', dependencies = c('Depends', 'Imports', 'LinkingTo'))"
+fi
+
 # Manually install Depends and Imports libraries + 'testthat'
 # to avoid a CI-time dependency on devtools (for devtools::install_deps())
 packages="c('data.table', 'jsonlite', 'Matrix', 'R6', 'testthat')"
 if [[ $OS_NAME == "macos" ]]; then
     packages+=", type = 'both'"
 fi
-Rscript --vanilla -e "options(install.packages.compile.from.source = 'always'); install.packages(${packages}, repos = '${CRAN_MIRROR}', lib = '${R_LIB_PATH}', dependencies = c('Depends', 'Imports', 'LinkingTo'))" || exit -1
+Rscript --vanilla -e "options(install.packages.compile.from.source = 'both'); install.packages(${packages}, repos = '${CRAN_MIRROR}', lib = '${R_LIB_PATH}', dependencies = c('Depends', 'Imports', 'LinkingTo'))" || exit -1
 
 cd ${BUILD_DIRECTORY}
 
--- a/.ci/test_r_package.sh
+++ b/.ci/test_r_package.sh
@@ -103,7 +103,7 @@ packages="c('data.table', 'jsonlite', 'Matrix', 'R6', 'testthat')"
 if [[ $OS_NAME == "macos" ]]; then
     packages+=", type = 'both'"
 fi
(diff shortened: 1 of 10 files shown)


.ci/test_r_package.sh
```diff
@@ -101,9 +101,9 @@ fi
 # to avoid a CI-time dependency on devtools (for devtools::install_deps())
 packages="c('data.table', 'jsonlite', 'Matrix', 'R6', 'testthat')"
 if [[ $OS_NAME == "macos" ]]; then
-    packages+=", type = 'binary'"
+    packages+=", type = 'both'"
```
jameslamb, line 104: as a reminder, this flag cannot be used on Linux, and that's why it's in this `if`.

From inside a `rocker/verse` container:

```text
> install.packages("jsonlite", type = "both")
Installing package into ‘/usr/local/lib/R/site-library’
(as ‘lib’ is unspecified)
Error in install.packages("jsonlite", type = "both") : 
  type == "both" can only be used on Windows or a CRAN build for macOS
```

StrikerRUS (approved):
Many thanks!

jameslamb:
@StrikerRUS we'll have to use admin rights to merge this, since I haven't fixed #3873 yet. Is that ok?

Otherwise I can remove the linting job temporarily from `all-successful` in the GitHub Actions config, then put it back in the fix for #3873

StrikerRUS:
@jameslamb 
> we'll have to use admin rights to merge this, 

And this one #3874 to fix **CUDA** tests.

I think admin rights is better than mess in CI code between PRs.

StrikerRUS:
I see in macOS R4:
```
 Error in loadNamespace(name) : there is no package called 'withr'
  Calls: test_check ... loadNamespace -> withRestarts -> withOneRestart -> doWithOneRestart
  Execution halted
```

Seems having `both` is not enough.

jameslamb:
> I see in macOS R4:
> 
> ```
>  Error in loadNamespace(name) : there is no package called 'withr'
>   Calls: test_check ... loadNamespace -> withRestarts -> withOneRestart -> doWithOneRestart
>   Execution halted
> ```
> 
> Seems having `both` is not enough.

ok seeing that, now I think the issue is that something somewhere in `{testthat}`'s many recursive dependencies (it was more than 100, the last time I checked), needs `{withr}` but didn't correctly specify it in their metadata. I can just add an `install.packages("withr")` to try to fix it.

I still think the `"both"` fix is a good one and will also improve stability, even if it didn't end up fixing this specific issue.

jameslamb:
hmmmm interestingly, it DOES look like R tried to install `{withr}` (https://github.com/microsoft/LightGBM/pull/3876/checks?check_run_id=1785068087).

> trying URL 'https://cloud.r-project.org/bin/macosx/contrib/4.0/withr_2.4.1.tgz'
  cannot open URL 'https://cloud.r-project.org/bin/macosx/contrib/4.0/withr_2.4.1.tgz'
In addition: Warning message:
In download.file(url, destfile, method, mode = "wb", ...) :
  cannot open URL 'https://cloud.r-project.org/bin/macosx/contrib/4.0/withr_2.4.1.tgz': HTTP status was '404 Not Found'
Warning in download.packages(pkgs, destdir = tmpd, available = available,  :
  download of package ‘withr’ failed
trying URL 'https://cloud.r-project.org/bin/macosx/contrib/4.0/data.table_1.13.6.tgz'
Content type 'application/x-gzip' length 2335683 bytes (2.2 MB)

But it still only looked for the binary. Ok, let me experiment a little. I commented out all other CIs so I don't hurt our capacity too much while testing.

jameslamb:
> I think admin rights is better than mess in CI code between PRs.

I agree!

jameslamb:
Ah!!! I can see on CRAN that the `{withr}` `r-release` binaries exist, but for 2.4.0 only (the package is on 2.4.1). This is why the builds are only failing for R 4.x and not R 3.6.

image

I think R is getting confused by this.

No amount of trickery with `install.packages()` can fix this and still preserve our binary installs, I don't think. So I guess we do have to add a separate installation of `{withr}`, from source, to resolve this.

StrikerRUS:
@jameslamb 

> So I guess we do have to add a separate installation of `{withr}`, from source, to resolve this.

This will be reverted after a day or two, right? When version will be in a sync state at CRAN, I mean.

jameslamb:
It probably could be reverted, yes. I don't know a more permanent fix for this problem :/

`type = "both"` is supposed to exactly solve this problem, but it looks like it isn't.

jameslamb:
This seems to have worked: https://github.com/microsoft/LightGBM/pull/3876/checks?check_run_id=1785421659

jameslamb:
#3874 and #3877 have been merged into `master` and I've merged `master` into this PR, so all CI jobs should now pass on this PR

PR #3880, opened 2021-01-29
[CUDA] Support CUDA 9 and test different CUDA versions at CI

StrikerRUS:
I think testing different CUDA versions will greatly help us to be confident in CUDA code.

Linking #3431.

CUDA < 10.0 fails to recognize `7.5` architecture.
```
[  7%] Building CUDA object CMakeFiles/histo_16_64_256-fulldata_sp_const.dir/src/treelearner/kernels/histogram_16_64_256.cu.o
nvcc fatal   : Unsupported gpu architecture 'compute_75'
```
Refer to https://github.com/Kitware/CMake/blob/36bb0e32d7e3976eed424078cf5cac459651f0f1/Modules/FindCUDA/select_compute_arch.cmake#L50-L60.


.github/workflows/cuda.yml
```diff
@@ -26,10 +26,13 @@ jobs:
         include:
           - method: source
             python_version: 3.6
+            cuda_version: "11.2.0"
           - method: pip
             python_version: 3.7
+            cuda_version: "10.0"
```
StrikerRUS, line 32: GitHub Actions treats values written as `x.0` like `int`s, so we need explicitly set them to strings.

CMakeLists.txt
```diff
@@ -156,7 +156,14 @@ if(USE_CUDA)
...
     LIST(APPEND CMAKE_CUDA_FLAGS -Xcompiler=${OpenMP_CXX_FLAGS} -Xcompiler=-fPIC -Xcompiler=-Wall)
-    CUDA_SELECT_NVCC_ARCH_FLAGS(CUDA_ARCH_FLAGS 6.0 6.1 6.2 7.0 7.5+PTX)
+
+    set(CUDA_ARCHS "6.0" "6.1" "6.2" "7.0")
+    if(CUDA_VERSION VERSION_GREATER_EQUAL "10.0")
+        list(APPEND CUDA_ARCHS "7.5")
+    endif()
+    list(POP_BACK CUDA_ARCHS CUDA_LAST_SUPORTED_VERSION)
+    list(APPEND CUDA_ARCHS "${CUDA_LAST_SUPORTED_VERSION}+PTX")
+    CUDA_SELECT_NVCC_ARCH_FLAGS(CUDA_ARCH_FLAGS ${CUDA_ARCHS})
```
StrikerRUS, line 166: cc @ChipKerchner 
I hope this change will not break anything on POWER machines.

ChipKerchner: I don't think so

.ci/setup.sh
```diff
@@ -80,9 +80,15 @@ else  # Linux
...
-            cmake \
             curl \
+            lsb-release \
+            software-properties-common \
             wget
+        curl -sL https://apt.kitware.com/keys/kitware-archive-latest.asc | apt-key add -
+        apt-add-repository "deb https://apt.kitware.com/ubuntu/ $(lsb_release -cs) main" -y
+        apt-get update
+        apt-get install --no-install-recommends -y \
+            cmake
```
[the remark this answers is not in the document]
jameslamb: blegh ok, that's alright

.github/workflows/cuda.yml
jameslamb: :scream: that's horrible haha

jameslamb (approved):
good idea, thanks

CMakeLists.txt
StrikerRUS: @ChipKerchner I'm so sorry for my poor English skills, but could you please rephrase? Do you think that
(1) this change can potentially hurt compilation on POWER systems
OR
(2) this change should be OK for POWER systems
?

And if (1), how this change can be harmful to POWER machines?

ChipKerchner: I think this change is OK for POWER systems.

StrikerRUS: OK, got it! Thanks a lot for finding time to comment!

Issue #3838, opened 2021-01-24
[docs] [dask] Document how to save a Dask model


jameslamb:
## Summary

There are multiple ways to save a model trained with LightGBM's Dask module. These might include the following (I have not tested if a fitted instance of `DaskLGBM***` is pickle-able):

* pickle the model object directly
* get the non-Dask, `lightgbm.sklearn` equivalent from `dask_model.to_local()` and save that
* get the `lightgbm` booster from `dask_model.booster_`, and use one of the supported methods for saving that Booster object

This documentation should also explain how to load these models, and should make it clear that using Dask for training does not mean that Dask packages are required at scoring time.

## Motivation

I believe this is different from #3814 , as it relies on some not-yet-investigated feature of the Dask module and some not-yet-decided design decisions.

## References

`cuml` has a nice, simple example for this: https://docs.rapids.ai/api/cuml/stable/pickling_cuml_models.html#Distributed-Model-Pickling

jameslamb:
Added to #2302 with other feature requests.

This can be contributed by anyone! Please leave a comment if you'd like to contribute this feature, and this issue will be re-opened.

jameslamb:
I'm reopening this and assigning to myself since I'm actively working on it, as a next step toward release 3.2.0 (#3872)

PR #3883, opened 2021-01-31
[dask] remove 'client' kwarg from fit() and predict() (fixes #3808)

### LINKED ISSUE #3808: [RFC] [dask] decide and document how users should provide a Dask client at training time
## Summary

The model objects in https://github.com/microsoft/LightGBM/blob/master/python-package/lightgbm/dask.py currently inherit from the classes in https://github.com/microsoft/LightGBM/blob/master/python-package/lightgbm/sklearn.py. For example, `lightgbm.dask.DaskLGBMRegressor` is a child of `lightgbm.sklearn.LGBMRegressor`.

These Dask estimators perform training on data stored in Dask collections (DataFrame and Array). To do this, they require a Dask client to talk to the cluster that those collections are stored on.

This issue describes the current state and possible alternatives for how LightGBM chooses the client to use for that task.

### Option 1 - provide `client` in a keyword arg to `.fit()` (current state)

In Dask training, users provide a client to the cluster that they'd like to use for training. Currently, this is done in the `.fit()` method (which overrides the parent method from the `sklearn` interface).

https://github.com/microsoft/LightGBM/blob/9cc3777c8a892c7ca253864839a926fd35f2b775/python-package/lightgbm/dask.py#L337-L339

**Good properties of this pattern:**

- allows full user control over the cluster used for training
- only used at runtime of `.fit()`, so `DaskLGBMClassifier` and `DaskLGBMRegressor` can be serialized with `pickle` / `cloudpickle` (~I need to confirm that...but I know for sure that~ you cannot pickle a Dask client)

**Bad properties of this pattern**

* makes the signature of the `.fit()` method look different from the `sklearn` equivalent, because of the extra parameter `client`

### Option 2 - don't allow users to customize client

We could choose not to expose `client` at all as a parameter, and just use the default client that's available, for example by asking people to run training in a context manager (shown below), more generally, by using [`distributed.get_client()`](https://distributed.dask.org/en/latest/api.html#distributed.get_client)

```python
from distributed import Client
with Client(cluster) as client:
    dask_reg = DaskLGBMRegressor(...)
    dask_reg.fit(X, y)
```

**Good properties**

- keeps the constructor AND `.fit()` methods identical to the `sklearn` equivalents
- can be generated at runtime of `.fit()`, so `DaskLGBMClassifier` and `DaskLGBMRegressor` can be serialized with `pickle` / `cloudpickle` (I need to confirm that...but I know for sure that you cannot pickle a Dask client)

**Bad properties**

- users could end up with the default client without knowing it, and this could cause issues that lead to confusion or poor performance

### Option 3 - allow users to set `.client` after construction

In this option, we could allow users to set `.client` after constructing an estimator, and use that property if it's set. Otherwise, use default client.

 ```python
dask_reg = DaskLGBMRegressor(...)
dask_reg.client = Client(cluster) 
dask_reg.fit(X, y) 
```

This is what `xgboost` does in their Dask interface --> https://github.com/dmlc/xgboost/blob/7bc56fa0eda6984197152ed4aa98d2fc9e9e8cc8/python-package/xgboost/dask.py#L171

**Good properties**

- keeps the constructor AND `.fit()` methods identical to the `sklearn` equivalents

**Bad properties**

- users could end up with the default client without knowing it, and this could cause issues that lead to confusion or poor performance
- this extra step of setting an attribute might feel unnatural to those who are used to the scikit-learn pattern of constructing an estimator then calling `.fit()`
- might have to do a bit of extra work to be sure that a `DaskLGBMRegressor` object (for example) can be pickled

### Option 4 - optionally pass a client in the model object's constructor

This is what `cuml` does for their Dask interface, for example: https://docs.rapids.ai/api/cuml/stable/api.html#cuml.dask.cluster.KMeans. If `client = None`, then in `cuml` you get the default client based on the context.

```python
dask_reg = DaskLGBMRegressor(client=Client(cluster))
dask_reg.fit(X, y) 
```

**Good properties**

- allows tight control over the client used, while also providing a you-don't-have-to-think-about-it option as the default

**Bad properties**

- constructor has different API than the scikit-learn equivalent
- constructed model object cannot be pickled directly (see https://docs.rapids.ai/api/cuml/stable/pickling_cuml_models.html#Distributed-Model-Pickling)

## Other related questions

This project's Dask interface is very new, and there are other unresolved questions that might be related to this one. Like

* should `.predict()` on Dask estimators only accept inputs that are Dask collections? `[will insert issue link here soon]`
    - today, it doesn't allow you to choose a client because it expects that the input will be local data (`pandas` or `numpy`)
* what set of options do we want to support for saving models trained with the Dask estimators? #3838

jameslamb:
This PR proposes a different pattern for how users can provide a Dask client to model objects in the Dask module.

It removes the keyword argument `client`  from `fit()` and `predict()`. As of this PR, users can optionally pass a client into the model's constructor. If they do that, it is saved on the model object and used by `fit()` and `predict()`. Otherwise, [`distributed.default_client()`](https://github.com/dask/distributed/blob/1297b18ff09276f6ad1553d40ab3ce77acf0fc0e/distributed/client.py#L4408) is used at runtime.

Given that you've set up a cluster and client...

```python
import lightgbm as lgb
from distributed import LocalCluster, Client

cluster = LocalCluster()
client = Client(cluster)
```

...any of the following would work as of this PR.

```python
# keyword arg in constructor
clf = lgb.DaskLGBMClassifier(client=client)
clf.fit(X, y)

# updating after construction with `.set_params(client=client)`
clf = lgb.DaskLGBMClassifier()
clf.set_params(client = client)
clf.fit(X, y)

# never telling LightGBM to use a specific client
clf = lgb.DaskLGBMClassifier()
clf.fit(X, y)
```

## Why prefer this pattern?

Based on feedback from @jsignell and @martindurant in https://github.com/microsoft/LightGBM/issues/3808#issuecomment-766835811 and https://github.com/microsoft/LightGBM/issues/3808#issuecomment-766904696, who felt comfortable with the use of `default_client()` when none is given, and who made these excellent points:

> I think the dasky-est approach would be to ... assume that they have initiated a client already

> I do think that allowing an optional argument to the model constructor is a reasonable and useful addition; you might just want to have separate things fitting on different clusters simultaneously

This pattern of an optional keyword-only argument in the constructor makes the signatures of `.predict()` and `.fit()` identical to those in the scikit-learn model objects.

I was also about to work around not being able to pickle a model object that had a Dask client as an attribute, which was my main concern with that approach.

This is very close to what `xgboost` chose (https://github.com/dmlc/xgboost/blob/d8ec7aad5a9a3eb580c55680aee8ad1a975cba20/python-package/xgboost/dask.py#L1386-L1394), except that today I believe they require users to set `.client` after constructing a model object.

`cuml` also allows for passing `client` into the model constructor (https://github.com/rapidsai/cuml/blob/a22681c89630926c48f98ac3cb54d39bd2b91026/python/cuml/dask/common/base.py#L41-L45).

### Changes in this PR

* adds optional keyword-only argument `client` to constructor for Dask model objects
* adds option for users to set  a model object's Dask client by using `.client` after construction
* removes keyword argument `client` from `.fit()`  and `.predict()` for Dask model objects
* adds a `__getstate__()` method on `lightgbm.sklearn.LGBMModel`, to handle removing a stored Dask client from a model before serializing it. (see [the `pickle` docs](https://docs.python.org/3/library/pickle.html#object.__getstate__) for information on how this works)
* adds unit tests that Dask model objects can be successfully serialized and read back in with `pickle`, `joblib`, and `cloudpickle`
* copies the constructor from `lightgbm.sklearn`  equivalents into the Dask module, so that customizing documentation works correctly (see comment on the relevant part of the diff). Adds a unit test that the signatures for Dask model objects and their sklearn equivalent are identical.

### Notes for Reviewers

* using the `docs/jlamb` branch so we can see readthedocs builds
* I chose not to label this `breaking` because there has never been a `lightgbm` release with the dask module. This change does make `lightgbm`'s API different from the one in `dask-lightgbm`, but that shouldn't be considered a breaking change from LightGBM's perspective
* I believe we should strive to make the Dask model objects pickleable (can be documented more fully in https://github.com/microsoft/LightGBM/issues/3838), so that users can store and load them just like they do other model objects. This PR makes that explicit with tests.

I know this is a small PR but with a lot of explanation and considerations. Thank you for your time and energy reviewing it!


tests/python_package_test/test_dask.py
```diff
@@ -415,6 +451,165 @@ def test_ranker(output, client, listen_port, group):
...
+
+    local_model = dask_model.to_local()
+    assert local_model._client is None
+    with pytest.raises(AttributeError):
+        local_model.client
+
+    client.close(timeout=CLIENT_CLOSE_TIMEOUT)
+
+
+@pytest.mark.parametrize('serializer', ['pickle', 'joblib', 'cloudpickle'])
```
jameslamb, line 526: I chose to test all three of these because:

* `pickle` is part of the standard library, and very popular
* `joblib` is popular with `scikit-learn`'s ecosystem, and tested in LightGBM's sklearn tests: https://github.com/microsoft/LightGBM/blob/ea8e47ea24fb942155f16da6720b32de976d5731/tests/python_package_test/test_sklearn.py#L393
* `cloudpickle` is the preferred alternative to `pickle` in the Dask ecosystem
    - `distributed` already has a strong dependency on it (https://github.com/dask/distributed/blob/08ea96890674d48b90f4e1f92959957e5e362a18/requirements.txt#L2), so this doesn't introduce any new dependencies for LightGBM's tests and any users of the Dask module will already have `cloudpickle`

python-package/lightgbm/dask.py
```diff
@@ -478,18 +495,58 @@ def _copy_extra_params(source: Union["_DaskLGBMModel", LGBMModel], dest: Union["
...
         for name in extra_param_names:
-            setattr(dest, name, attributes[name])
+            if name != "_client":
+                setattr(dest, name, attributes[name])
 
 
 class DaskLGBMClassifier(LGBMClassifier, _DaskLGBMModel):
     """Distributed version of lightgbm.LGBMClassifier."""
 
+    def __init__(self, boosting_type='gbdt', num_leaves=31, max_depth=-1,
```
jameslamb, line 505: This seems to have the desired effect! When I built the docs locally, I saw that `__init__()` docs for `DaskLGBMClassifier`, `DaskLGBMRegressor`, and `DaskLGBMRanker` have the `client` doc added, and the docs for their scikit-learn equivalents do not.

**DaskLGBMClassifier.__init__()**

image

**LGBMClassifier.__init__()**

image

### Why I think copying is the best alternative

I tried several other ways to update the docstrings, but none of them quite worked. I'll explain in terms of `*Classifier`, but this applies for `*Regressor` and `*Ranker` as well.

1. Editing `DaskLGBMClassifier.__init__.doc`

```python
class DaskLGBMClassifier(LGBMClassifier, _DaskLGBMModel):
    ...

_base_doc = LGBMClassifier.__init__.__doc__
_before_kwargs, _kwargs, _after_kwargs = _base_doc.partition('**kwargs')
DaskLGBMClassifier.__init__.__doc__ = (
    _before_kwargs
    + 'client : dask.distributed.Client or None, optional (default=None)\n'
    + ' ' * 12 + 'Dask client. If ``None``, ``distributed.default_client()`` will be used at runtime. This client will not be saved if the model object is pickled.\n'
    + ' ' * 8 + _kwargs + _after_kwargs
)
```

Doing this for each of the 3 model objects, the API docs show 3 copies of the doc for `client`. Those 3 copies also show up on the docs for `lightgbm.sklearn.LGBMClassifier`.

Screen Shot 2021-01-30 at 11 23 24 PM


2. Setting up a pass-through `__init__()`

```python
class DaskLGBMClassifier(LGBMClassifier, _DaskLGBMModel):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)

    _base_doc = __init__.__doc__
    _before_kwargs, _kwargs, _after_kwargs = _base_doc.partition('**kwargs')
    __init__.__doc__ = (
        _before_kwargs
        + 'client : dask.distributed.Client or None, optional (default=None)\n'
        + ' ' * 12 + 'Dask client. If ``None``, ``distributed.default_client()`` will be used at runtime. This client will not be saved if the model object is pickled.\n'
        + ' ' * 8 + _kwargs + _after_kwargs
    )
```

This results in this error at runtime.

```text
doing this to avoid:
RuntimeError: scikit-learn estimators should always specify their parameters in
the signature of their __init__ (no varargs).
<class 'lightgbm.dask.DaskLGBMClassifier'> with constructor (self, *args, **kwargs)
doesn't  follow this convention.
```

3. Just copying `__init__`

Both of these variations:

```python
class DaskLGBMClassifier(LGBMClassifier, _DaskLGBMModel):
     __init__ = LGBMClassifier.__init__

    _base_doc = __init__.__doc__
    _before_kwargs, _kwargs, _after_kwargs = _base_doc.partition('**kwargs')
    __init__.__doc__ = (
        _before_kwargs
        + 'client : dask.distributed.Client or None, optional (default=None)\n'
        + ' ' * 12 + 'Dask client. If ``None``, ``distributed.default_client()`` will be used at runtime. This client will not be saved if the model object is pickled.\n'
        + ' ' * 8 + _kwargs + _after_kwargs
    )
```

Doing this for each of the 3 model objects, the API docs show 3 copies of the doc for `client`. Those 3 copies also show up on the docs for `lightgbm.sklearn.LGBMClassifier`. (same as in the image above)

StrikerRUS (changes requested):
@jameslamb I'm afraid this is an antipattern in the scikit-learn world! Our classes will be incompatible with scikit-learn dask ecosystem.
This is why you got 
```
RuntimeError: scikit-learn estimators should always specify their parameters in
the signature of their __init__ (no varargs).
```
I believe the best option will be passing `client` in `__init__` as normal argument with `None` as default value. I mean the following:
```
class DaskLGBMClassifier(LGBMClassifier, _DaskLGBMModel):
    def __init__(self, ..., client=None, **kwargs):
        self.client = client
        super().__init__(...)
```

Also, do not suggest users setting `client` directly via attribute. `set_params(client=new_client)` should be used.

python-package/lightgbm/sklearn.py
```diff
@@ -325,6 +328,13 @@ def __init__(self, boosting_type='gbdt', num_leaves=31, max_depth=-1,
         self._n_classes = None
         self.set_params(**kwargs)
 
+    def __getstate__(self):
+        """Remove un-picklable attributes before serialization."""
+        client = self.__dict__.pop("_client", None)
+        out = copy.deepcopy(self.__dict__)
+        self._client = client
+        return out
```
StrikerRUS, line 336: I believe this can be done in Dask estimators, not in parent class.

tests/python_package_test/test_dask.py
```diff
@@ -13,6 +16,7 @@
 if not lgb.compat.DASK_INSTALLED:
     pytest.skip('Dask is not installed', allow_module_level=True)
 
+import cloudpickle
```
StrikerRUS, line 19: I believe `cloudpickle` should be added here as we importing it directly, we shouldn't rely on that it is a second-order dependency of some package.
https://github.com/microsoft/LightGBM/blob/ea8e47ea24fb942155f16da6720b32de976d5731/.ci/test.sh#L93

python-package/lightgbm/sklearn.py
jameslamb: If we do it there, this code would have to be copied 3 times. Because `_DaskLGBMModel` comes second in MRO, it wouldn't be safe to just put this on that mixin. Are you ok with that duplication?

StrikerRUS: Yeah, probably `__getstate__()` is overridden somewhere above in pure scikit-learn branch of inheritance. But we can name this method as `_getstate()` in `_DaskLGBMModel` and call it, so no duplication will be required.
```
def __getstate__(self):
   return self._get_state()
```
Hust like we currently do for all other methods.

jameslamb:
> I believe the best option will be passing client in __init__ as normal argument with None as default value. I mean the following:

> class DaskLGBMClassifier(LGBMClassifier, _DaskLGBMModel):
    def __init__(self, ..., client=None, **kwargs):
        self.client = client
        super().__init__(...)
> Also, do not suggest users setting client directly via attribute. set_params(client=new_client) should be used.

I'm really struggling to adopt this pattern. `sklearn.BaseEstimator` (the parent class that `get_params()` comes from), assumes that every keyword argument from the constructor counts as a parameter and returns it from `.get_params()`.

https://github.com/scikit-learn/scikit-learn/blob/b3ea3ed6a/sklearn/base.py#L152

So treating `client` like this is causing all of the tests to fail, with complaints like

> TypeError: cannot pickle '_asyncio.Task' object

Probably because `get_params()` is called in https://github.com/microsoft/LightGBM/blob/198a15138e94844b73a1a9ef88d909731160899a/python-package/lightgbm/dask.py#L452 and then passed through to `_train()`. The client is in it and not pickleable.

I pushed https://github.com/microsoft/LightGBM/pull/3883/commits/344376bee1a7a3dbe97f7078a782d091d1084c30 so you can see what I tried.

<hr>

I really think this is a sign that we should not expose `client` as a keyword argument in the signature, and should revert that part of https://github.com/microsoft/LightGBM/pull/3883/commits/344376bee1a7a3dbe97f7078a782d091d1084c30 and use my original proposal to `.pop()` it off of kwargs. I also believe we should not use `.set_params()` for `client`.

I understand that you said my original scikit-learn antipattern, but I don't think it's appropriate to treat `client` like "just another parameter". scikit-learn's preference for having explicit kwargs and having each argument be an instance attribute are described in ["developing scikit-learn estimators"](https://scikit-learn.org/stable/developers/develop.html#instantiation)

>  every keyword argument accepted by `__init__` should correspond to an attribute on the instance. Scikit-learn relies on this to find the relevant attributes to set on an estimator when doing model selection.

> There should be no logic, not even input validation, and the parameters should not be changed...The reason for postponing the validation is that the same validation would have to be performed in `set_params`, which is used in algorithms like `GridSearchCV`.

Given this, I really don't think the pattern I proposed violates the intention from scikit-learn.

* you would never search over values of `client` in hyperparameter tuning
* `client` isn't required to be the same to reproduce a model result
* `client` shouldn't ever be saved in a model file, even if it could be pickled, because it refers to temporary infrastructure that isn't guaranteed to be there when you load that model in the future

tests/python_package_test/test_dask.py
jameslamb: sure, done in https://github.com/microsoft/LightGBM/pull/3883/commits/344376bee1a7a3dbe97f7078a782d091d1084c30

python-package/lightgbm/sklearn.py
jameslamb: ok, good suggestion! did that in https://github.com/microsoft/LightGBM/pull/3883/commits/344376bee1a7a3dbe97f7078a782d091d1084c30.

I called it `_lgb_get_state` because I'm nervous about using general-sounding method names in `_DaskLGBMModel` when there are so many classes upstream of `LGBMClassifier`. I'm nervous about some version of scikit-learn introducing a `_get_state` that would override this. We might not catch that in tests because we don't test against many different versions of scikit-learn.

I think I'll propose a PR to do that for other methods on the `_DaskLGBMModel` mixin as well.

StrikerRUS:
>Given this, I really don't think the pattern I proposed violates the intention from scikit-learn.

Scikit-learn is really complex system with many undocumented "features", rules and assumptions about inherited classes. I guess you've already known this. So it is very likely that violating some of their rules can break a lot of integrations. Of course, inability to use hyperparameter search for `client` argument is not a serious problem, but there is no guaranty that it is not breaking a whole hyperparameter tunability.

Also, I'm strongly against interfering into `sklearn.py` module with changes required for child Dask module. Making a parent aware of its children is another serious programming antipattern.

I believe that we can achieve the goal to make `_DaskLGBMModel` class serialisable without much harm to sklearn compatibility and hacky workarounds. I'll try some ideas in a next days.

jameslamb:
>  I'll try some ideas in a next days

Don't worry about it, I know you've mentioned you can't test Dask features in your local dev environment. I'll continue down this path you've suggested and we can look at the diff at the end.

StrikerRUS:
> Don't worry about it, I know you've mentioned you can't test Dask features in your local dev environment.

But I could use our Linux CI job 😉 . Not as comfortable as own local setup, but better than nothing.

Only one thing is making me puzzled. 

```
# fitted model should survive pickling round trip, and pickling
# shouldn't have side effects on the model object
```

Given that `dask.distributed.Client` is unpickable, how `fitted_model.client === model_loaded_from_disk.client` can be achieved without overwriting `__setstate__` and finding a way to save client metainfo that will help to re-create it after loading from disk?
For example, `Booster` object is saved into a string and recreated from it during unpickling process:
https://github.com/microsoft/LightGBM/blob/fd33199c7b7b46dae8fd10b710a97a284cc4bb6f/python-package/lightgbm/basic.py#L2284-L2303

jameslamb:
Please don't spend time on it, I want to be respectful of your time and I ithink I can go faster testing locally.

For your questiin about pickling...it's acceptable and expected that the model loaded from a file doesn't have .client set. A client is a live connection to a specfic Dask scheduler and you have no guarantee that exact same schefuler will still exist or be accessible from wherever you load the model. This has no impact on the correctness of the model. Load the same `model.pkl` and called `.predict()` on the same data on two different Dask clusters, you will get identical results.

StrikerRUS:
Yeah, you'll definitely be faster implementing this, but I'd like to help.

Then I'm listing here some initial ideas for your consideration I wanted to try first:
- move `if client is None: client = default_client()` into `_train()` function;
- simply use `self.client = client` in `__init__`s; `set_params` should be used **outside** class implementation by users, not by us in the internals;
- pop `self.client` in `__getstate__` and replace it with something that will help to "clone" current client;
- in `__setstate__` use saved metainfo about client to create a new one and save in `self.client`.

StrikerRUS:
For the last two points in my list of initial ideas I meant to find something close to the following:
https://github.com/kichappa/QC/blob/ad9f756d8fafac3916688fd6dfb3786ae7c2aa14/distributed/variable.py#L203-L213

jameslamb:
Ok @StrikerRUS I think I've come up with something that addresses your comments. Thanks for being patient with me. Can you please take another look?

> simply use self.client = client in __init__s; set_params should be used outside class implementation by users, not by us in the internals;

done

> move if client is None: client = default_client() into `_train()` function;

I disagree with this. I think using an `@property` on the estimator is a better pattern, because then user code can access `model.client` and know for sure that that will always return the client to be used in training or prediction (either one they set or `distributed.default_client()`).

> pop self.client in `__getstate__` and replace it with something that will help to "clone" current client; in `__setstate__` use saved metainfo about client to create a new one and save in self.client.

I really think we should not try to recreate the client that was active at the time that a model was saved. I expressed some thoughts on this in https://github.com/microsoft/LightGBM/pull/3883#issuecomment-770468202.

Dask clusters shouldn't be thought of as very-long-lived infrastructure like a database. Dask clusters are intended to be temporary, and the dominant pattern for Dask code is connect to whatever the "active" cluster is at runtime (this is what @jsignell was explaining in https://github.com/microsoft/LightGBM/issues/3808#issuecomment-766835811). So I think it is absolutely fine, and actually desirable, for a `DaskLGBM*` loaded from a file to have `self._client = None`. Users whoo want tight control over the client can use `.set_params(client=client)` after loading.

The code you linked in https://github.com/microsoft/LightGBM/pull/3883#issuecomment-770472008, for example, saves a scheduler address (a specific IP + port) and would fail if there isn't a Dask scheduler running on that IP and listening on that port.

StrikerRUS:
@jameslamb Seems pickling tests are still failing. Please check it.

> I disagree with this. I think using an `@property` on the estimator is a better pattern, because then user code can access `model.client` and know for sure that that will always return the client to be used in training or prediction (either one they set or `distributed.default_client()`).

Agree, makes sense! But please rename this property to `client_` according to scikit-learn rules. All normal attributes (without trailing and leading underscores) should be set in `__init__` and their values should never be changed. For any computed during `fit` stage attributes should attributes with trailing underscores be used.

https://scikit-learn.org/stable/developers/develop.html#parameters-and-init

Please check the following example with `objective` argument.
https://github.com/microsoft/LightGBM/blob/763b5f3cdc180e0ade87d2c4ad785cb9a47287ce/python-package/lightgbm/sklearn.py#L211-L214
https://github.com/microsoft/LightGBM/blob/763b5f3cdc180e0ade87d2c4ad785cb9a47287ce/python-package/lightgbm/sklearn.py#L295
https://github.com/microsoft/LightGBM/blob/763b5f3cdc180e0ade87d2c4ad785cb9a47287ce/python-package/lightgbm/sklearn.py#L318
https://github.com/microsoft/LightGBM/blob/763b5f3cdc180e0ade87d2c4ad785cb9a47287ce/python-package/lightgbm/sklearn.py#L494-L502
https://github.com/microsoft/LightGBM/blob/763b5f3cdc180e0ade87d2c4ad785cb9a47287ce/python-package/lightgbm/sklearn.py#L735-L740

jameslamb:
Ok I think I've made the requested changes in https://github.com/microsoft/LightGBM/pull/3883/commits/af269b56f88b4d2d0f9179dee471c869aa9713b7. Add `client_` so now we have `client`, `_client`, and `_client`. And fixed the pickling tests.

python-package/lightgbm/dask.py
```diff
@@ -478,18 +505,82 @@ def _copy_extra_params(source: Union["_DaskLGBMModel", LGBMModel], dest: Union["
...
+            importance_type=importance_type,
+            **kwargs
+        )
+
+    _base_doc = LGBMClassifier.__init__.__doc__
+    _before_kwargs, _kwargs, _after_kwargs = _base_doc.partition('**kwargs')
+    __init__.__doc__ = (
+        _before_kwargs
+        + 'client : dask.distributed.Client or None, optional (default=None)\n'
+        + ' ' * 12 + 'Dask client. If ``None``, ``distributed.default_client()`` will be used at runtime. This client will not be saved if the model object is pickled.\n'
```
StrikerRUS, line 572: TBH, it was not clear enough that **any** client will not be saved. "This" may refer to `distributed.default_client()`, I guess, and confuse users that custom client will be saved...

tests/python_package_test/test_dask.py
```diff
@@ -137,6 +141,32 @@ def _accuracy_score(dy_true, dy_pred):
...
+    if serializer == 'pickle':
+        with open(filepath, 'wb') as f:
+            pickle.dump(obj, f)
+    elif serializer == 'joblib':
+        joblib.dump(obj, filepath)
+    elif serializer == 'cloudpickle':
+        with open(filepath, 'wb') as f:
+            cloudpickle.dump(obj, f)
+    else:
+        raise ValueError(f'unrecognized serializer type: {serializer}')
```
StrikerRUS, line 154:
```suggestion
        raise ValueError(f'Unrecognized serializer type: {serializer}')
```

tests/python_package_test/test_dask.py
```diff
@@ -137,6 +141,32 @@ def _accuracy_score(dy_true, dy_pred):
...
+    if serializer == 'pickle':
+        with open(filepath, 'rb') as f:
+            return pickle.load(f)
+    elif serializer == 'joblib':
+        return joblib.load(filepath)
+    elif serializer == 'cloudpickle':
+        with open(filepath, 'rb') as f:
+            return cloudpickle.load(f)
+    else:
+        raise ValueError(f'unrecognized serializer type: {serializer}')
```
StrikerRUS, line 167:
```suggestion
        raise ValueError(f'Unrecognized serializer type: {serializer}')
```

tests/python_package_test/test_dask.py
```diff
@@ -424,6 +460,239 @@ def test_ranker(output, client, listen_port, group):
...
+        "local_listen_port": listen_port,
+        "n_estimators": 1,
+        "num_leaves": 2
+    }
+
+    # fit should work if client isn't provided
+    dask_model = model_factory(**params)
+    assert dask_model._client is None
+    assert dask_model.client is None
+    assert dask_model.client_ == client
```
StrikerRUS, line 493: This should raise `UnfittedError` according to sklearn policy.

tests/python_package_test/test_dask.py
```diff
@@ -424,6 +460,239 @@ def test_ranker(output, client, listen_port, group):
...
+
+    local_model = dask_model.to_local()
+    with pytest.raises(AttributeError):
+        local_model._client
+        local_model.client
+        local_model.client_
+
+    # should be able to set client after construction
+    dask_model = model_factory(**params)
+    dask_model.set_params(client=client)
```
StrikerRUS, line 516: Can we set here some custom client to ensure that path branch `else` for condition `if self._client is None:` works?

tests/python_package_test/test_dask.py
```diff
@@ -424,6 +460,239 @@ def test_ranker(output, client, listen_port, group):
...
+
+    params = {
+        "time_out": 5,
+        "local_listen_port": listen_port,
+        "n_estimators": 1,
+        "num_leaves": 2
+    }
+
+    if set_client:
+        params.update({"client": client})
```
StrikerRUS, line 573: Same here about custom client to check difference between `default_client()` and provided by user.

tests/python_package_test/test_dask.py
```diff
@@ -501,3 +772,23 @@ def f(part):
             model_factory=lgb.LGBMClassifier
         )
         assert 'foo' in str(info.value)
+
+
+def test_dask_classes_and_sklearn_equivalents_have_identical_constructors_except_client_arg():
```
StrikerRUS, line 777: 👍

tests/python_package_test/test_dask.py
```diff
@@ -501,3 +772,23 @@ def f(part):
...
+
+        # "client" should be the only different, and the final argument
+        assert dask_spec.args[:-1] == sklearn_spec.args
+        assert dask_spec.defaults[:-1] == sklearn_spec.defaults
+        assert dask_spec.args[-1] == 'client'
+        assert dask_spec.defaults[-1] is None
+
+    _compare_spec(lgb.DaskLGBMClassifier, lgb.LGBMClassifier)
+    _compare_spec(lgb.DaskLGBMRegressor, lgb.LGBMRegressor)
+    _compare_spec(lgb.DaskLGBMRanker, lgb.LGBMRanker)
```
StrikerRUS, line 794: Maybe `@pytest.mark.parametrize`?

StrikerRUS (commented):
@jameslamb 
Awesome work!
Thanks a lot for digging deep in sklearn internals and finding way not to make a parent aware of its children. I left some comments for your consideration.

python-package/lightgbm/dask.py
jameslamb: Ok I can change it

python-package/lightgbm/dask.py
```diff
@@ -434,25 +434,49 @@ def _predict(
...
+    _client: Optional[Client] = None
+
+    @property
+    def client_(self) -> Client:
+        """Dask client.
+
+        This property can be passed in the constructor or updated
+        with ``model.set_params(client=client)``.
+        """
+        if self._client is None:
```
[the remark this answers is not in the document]
jameslamb: Oh I see, ok. Why should the check be on `._n_features`? That seems kind of indirect. Shouldn't it be on `.fitted_` directly for a check of whether or not the model has been fit?

I'm also confused how you would like `_fit()` to work. I'm currently passing `client=self.client_` into `self._fit()`, and relying on that to resolve whether there is a client stored on the object or `get_default_client()` should be used. This suggested change would break that behavior, because of course the model is probably not fitted yet at the time you call `.fit()`. What do you suggest I do for that?

> BTW, I think it will be great to setup sklearn integration tests at our CI for Dask classes. It will not allow us to be sure that our classes fully compatible with sklearn but at least will check basic compatibility. WDYT

Sure, I think that's fine. I think it's an ok fit for this PR and will add it here.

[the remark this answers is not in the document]
StrikerRUS: > Why should the check be on `._n_features`?

For historical reasons. `fitted_` was introduced quite recently in our sklearn wrapper. But now you can use it, indeed 👍 .

> I'm also confused how you would like `_fit()` to work.

Oh I see now! Can you pass `self.client` into `_fit()` and there assign result of a check to `self._client`? While `client_` will always return `self._client` or raise error? Just like all other properties in sklearn wrapper.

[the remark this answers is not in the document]
StrikerRUS: > I think it's an ok fit for this PR and will add it here.

Maybe in a follow-up PR? I'm afraid that it can fail now and stop merging `client` argument migration.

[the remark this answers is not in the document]
jameslamb: I want to be really sure that there is exactly one place where resolve which client to use (https://github.com/microsoft/LightGBM/pull/3883#issuecomment-770554762).

So I don't want to put this code into the body of `_fit()`.

```python
if self.client is None:
    client = default_client()
else:
    client = self.client
```

I think it would work to pull that out into a small function like this:

```python
def _choose_client(client: Optional[Client]):
    if self.client is None:
        return default_client()
    else:
        return self.client
```

Then use that in both the `client_` property (after the fitted check) and in `_fit()`. That would give us confidence that accessing `.client_` returns the same client that will be used in training.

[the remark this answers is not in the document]
StrikerRUS: OK, sounds good!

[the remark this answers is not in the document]
jameslamb: Added https://github.com/microsoft/LightGBM/issues/3894 to capture the task of adding a scikit-learn compatibility test. I made this a "good first issue" because I think it could be done without deep knowledge of LightGBM, but I also think it's ok for you or me to pick up in the near future (we don't need to reserve it for new contributors).

python-package/lightgbm/dask.py
jameslamb: updated in https://github.com/microsoft/LightGBM/pull/3883/commits/555a57a47345e6b5b1a205b88a4c794bbf096e50 to this

> The Dask client used by this class will not be saved if the model object is pickled.

python-package/lightgbm/dask.py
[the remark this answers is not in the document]
jameslamb: Ok I think I've addressed these suggestions in https://github.com/microsoft/LightGBM/pull/3883/commits/555a57a47345e6b5b1a205b88a4c794bbf096e50

* added an `LGBMNotFittedError` if accessing `.client_` for a not-yet-fitted model
* added an internal function `_get_dask_cluster()` to hold the logic of using `default_client()` if `client` is None
* changed the unit tests to use two different clusters and clients

Note that the change to have two clusters active in the same test will result in a lot of this warning in the logs:

> test_dask.py::test_model_and_local_version_are_picklable_whether_or_not_client_set_explicitly[False-ranking-cloudpickle]
  /home/jlamb/miniconda3/lib/python3.7/site-packages/distributed/node.py:155: UserWarning: Port 8787 is already in use.
  Perhaps you already have a cluster running?
  Hosting the HTTP server on port 40743 instead
    http_address["port"], self.http_server.port

This is not something we have to worry about, and I'd rather leave it alone and let Dask do the right thing (picking a random port for the scheduler when the default one is unavailable) than add more complexity to the tests by adding our own logic to set the scheduler port.

python-package/lightgbm/dask.py
```diff
@@ -434,25 +453,49 @@ def _predict(
 
 class _DaskLGBMModel:
 
+    # self._client is set in the constructor of classes that use this mixin
+    _client: Optional[Client] = None
```
StrikerRUS, line 457: Looks like `_client` was replaced with `_get_dask_client()` function in the latest commit and is not needed anymore.

tests/python_package_test/test_dask.py
```diff
@@ -501,3 +821,29 @@ def f(part):
...
+    [
+        (lgb.DaskLGBMClassifier, lgb.LGBMClassifier),
+        (lgb.DaskLGBMRegressor, lgb.LGBMRegressor),
+        (lgb.DaskLGBMRanker, lgb.LGBMRanker)
+    ]
+)
+def test_dask_classes_and_sklearn_equivalents_have_identical_constructors_except_client_arg(classes):
+    def _compare_spec(dask_cls, sklearn_cls):
+        dask_spec = inspect.getfullargspec(dask_cls)
+        sklearn_spec = inspect.getfullargspec(sklearn_cls)
```
StrikerRUS, line 837: I think it is possible to simplify here by removing inner function and setting directly 
```
dask_spec = inspect.getfullargspec(classes[0])
sklearn_spec = inspect.getfullargspec(classes[1])
```

StrikerRUS (approved):
@jameslamb Great job! And I think this API is quite intuitive.
Just two nits.

tests/python_package_test/test_dask.py
jameslamb: oh right! didn't think about how using `parametrize` meant this function was useless

python-package/lightgbm/dask.py
jameslamb: oh good point! Remove in https://github.com/microsoft/LightGBM/pull/3883/commits/876bfe5f6c756e946594e3316fec716ae358de18

tests/python_package_test/test_dask.py
jameslamb: removed this function in https://github.com/microsoft/LightGBM/pull/3883/commits/876bfe5f6c756e946594e3316fec716ae358de18

jameslamb:
> Great job! And I think this API is quite intuitive.

Thanks for talking me through it! Learned a lot more about scikit-learn doing this, you're a good teacher.

I've addressed the last two comments in https://github.com/microsoft/LightGBM/pull/3883/commits/876bfe5f6c756e946594e3316fec716ae358de18. Will merge after I see if CI + readthedocs passes, and after checking the docs site.

jameslamb:
readthedocs looks ok to me!

image

https://lightgbm.readthedocs.io/en/docs-jlamb/

no change to `LGBMClassifier` docs

image

dask docs have client in them (I checked classifier, regressor, and ranker)

image

`client_` shows up in attributes table

image

with the expected docs

image

`client` is gone from the `fit()` docs

image

StrikerRUS:
@jameslamb Thank you! Sorry my help was in a form of only words not of actual code 🙁 .

PR #3885, opened 2021-01-31
[CUDA][docs] explicitly require CUDA 9.0 or higher

StrikerRUS:


jameslamb (approved):
makes sense, thank you

guolinke (approved):

PR #3892, opened 2021-02-02
[dask] [python] client.rebalance on dask ranker test

ffineis:
Further addresses #3817 by making `test_dask.py` a little bit more predictable. Previously the `test_ranker` test while using `output='array'` provided very uneven `group` allocations among its two test workers, unlike in the case of dask.dataframe input. Because `#data` is so small (e.g. 100 rows), when one worker gets a small amount of data relative to the other(s), this can cause rather significant discrepancies between the Dask ranker and the standard `LGBMRanker`. [See comment](https://github.com/microsoft/LightGBM/issues/3817#issuecomment-770426671) for more background.

Applying [`client.rebalance`](https://distributed.dask.org/en/latest/api.html#distributed.Client.rebalance) in this case (small data and uneven worker data distributions) helps make the tests a bit more predictable, tightening the distribution of observed spearman correlations with `LGBMRanker`. Thanks for the patience to the maintainers!

### DIFF
--- a/tests/python_package_test/test_dask.py
+++ b/tests/python_package_test/test_dask.py
@@ -19,10 +19,10 @@ import numpy as np
 import pandas as pd
 from scipy.stats import spearmanr
 from dask.array.utils import assert_eq
+from dask.distributed import wait
 from distributed.utils_test import client, cluster_fixture, gen_cluster, loop
 from scipy.sparse import csr_matrix
 from sklearn.datasets import make_blobs, make_regression
-from sklearn.utils import check_random_state
 
 from .utils import make_ranking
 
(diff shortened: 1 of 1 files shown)


tests/python_package_test/test_dask.py
```diff
@@ -409,7 +418,7 @@ def test_ranker(output, client, listen_port, group):
     # have high rank correlation with scores from serial ranker.
     dcor = spearmanr(rnkvec_dask, y).correlation
     assert dcor > 0.6
-    assert spearmanr(rnkvec_dask, rnkvec_local).correlation > 0.75
+    assert spearmanr(rnkvec_dask, rnkvec_local).correlation > 0.8
```
jameslamb, line 421: nice! I'm comfortable with this, given the distributed you saw in https://github.com/microsoft/LightGBM/issues/3817#issuecomment-770426671

tests/python_package_test/test_dask.py
```diff
@@ -19,10 +19,10 @@
 import pandas as pd
 from scipy.stats import spearmanr
 from dask.array.utils import assert_eq
+from dask.distributed import wait
 from distributed.utils_test import client, cluster_fixture, gen_cluster, loop
 from scipy.sparse import csr_matrix
 from sklearn.datasets import make_blobs, make_regression
-from sklearn.utils import check_random_state
```
jameslamb, line 25: oh thanks for removing this. Didn't realize it was unused.

jameslamb (approved):
This is awesome! This is one of those fun changes that looks super simple, but I know a ton of good research went into it in https://github.com/microsoft/LightGBM/issues/3817#issuecomment-770426671.

Thanks very much 🚀

tests/python_package_test/test_dask.py
ffineis: all credit goes to pycharm

PR #3895, opened 2021-02-02
[SWIG] add size_t manipulation functions

shuttie:
I'm a maintainer of the [lightgbm4j](https://github.com/metarank/lightgbm4j) library, and would like to expose the `LGBM_DatasetGetFeatureNames` function in a more java-friendly way. 

The problem with the `LGBM_DatasetGetFeatureNames` function is that it requires a `size_t` [type param](https://lightgbm.readthedocs.io/en/latest/C-API.html#c.LGBM_DatasetGetFeatureNames), which is not possible to instantiate at all, as SWIG definition file only exposes int/long/double/float/int64_t/int32_t types.

This one-liner PR also exposes a `size_t` function family in SWIG wrapper, so `lightgbmlibJNI.java` will also include these new ones:

```
>   public final static native long new_size_tp();
>   public final static native long copy_size_tp(long jarg1);
>   public final static native void delete_size_tp(long jarg1);
>   public final static native void size_tp_assign(long jarg1, long jarg2);
>   public final static native long size_tp_value(long jarg1);
```

### DIFF
--- a/swig/pointer_manipulation.i
+++ b/swig/pointer_manipulation.i
@@ -21,6 +21,7 @@
 %pointer_functions(float, floatp)
 %pointer_functions(int64_t, int64_tp)
 %pointer_functions(int32_t, int32_tp)
+%pointer_functions(size_t, size_tp)
 
 %pointer_cast(int64_t *, long *, int64_t_to_long_ptr)
 %pointer_cast(int64_t *, double *, int64_t_to_double_ptr)


jameslamb:
Thanks for contributing!

@AlbertoEAF or @imatiach-msft , could you comment on this?

AlbertoEAF:
Hello @shuttie :)

Thanks for contributing.There are already several LightGBM Java providers out there which are using such functions in Java through a int64_t when size_t pointers are needed. However, I agree that is more correct to use a proper size_t pointer.

LGTM ;)

shuttie:
Using `int64_t` as a `size_t` can break code in a case if someone will try to build this library on 32-bit systems (like an Android phone not on Aarch64): AFAIK size_t is [defined as `unsigned int` on 32-bit ARM](https://developer.arm.com/documentation/dui0491/i/C-and-C---Implementation-Details/Basic-data-types). 

So I see two options here:
* force SWIG to mess with datatypes in a way that int64_t will be casted to size_t, which may break on 32-bit
* change the `LGBM_DatasetGetFeatureNames` to use platform-independent types for their arguments, which will break API compatibility
* allow SWIG to correctly use the size_t type, so it will both work on 32bit and preserve API compat at the same time.

AlbertoEAF:
Sure shuttie, I'm not defending using int64_t, that's why I said:

> I agree that is more correct to use a proper size_t pointer.
>
> LGTM ;)

imatiach-msft (approved):
LGTM!

StrikerRUS (approved):
Merging based on @AlbertoEAF's and @imatiach-msft's approvals. Thank you both guys!
@shuttie Thanks a lot got your contribution!

Issue #3882, opened 2021-01-30
[dask] Having workers without data kills the training process


jmoralez:
## How you are using LightGBM?

LightGBM component: Python package

## Environment info

Operating System: Ubuntu 20.04

CPU/GPU model: Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz

C++ compiler version: g++ 9.3.0

CMake version: 3.18.2

Python version: 3.7.9

LightGBM version or commit hash: b6386842d1b24d59928c9bf67333ddafdbda0da9

## Error message and / or logs

[LightGBM] [Warning] Set TCP_NODELAY failed                                                                                                       
[LightGBM] [Warning] Set TCP_NODELAY failed                                                                                                       
[LightGBM] [Info] Trying to bind port 38927...                                                                                                    
[LightGBM] [Info] Trying to bind port 44733...                                                                                                    
[LightGBM] [Info] Binding port 38927 succeeded                                                                                                    
[LightGBM] [Info] Binding port 44733 succeeded                                                                                                    
[LightGBM] [Warning] Set TCP_NODELAY failed                                                                                                       
[LightGBM] [Warning] Set TCP_NODELAY failed                                                                                                       
[LightGBM] [Info] Listening...                                                                                                                    
[LightGBM] [Info] Listening...                                                                                                                    
[LightGBM] [Warning] Connecting to rank 2 failed, waiting for 200 milliseconds                                                                    
[LightGBM] [Warning] Connecting to rank 2 failed, waiting for 200 milliseconds                                                                    
[LightGBM] [Warning] Connecting to rank 2 failed, waiting for 260 milliseconds
[LightGBM] [Warning] Connecting to rank 2 failed, waiting for 260 milliseconds
[LightGBM] [Warning] Connecting to rank 2 failed, waiting for 338 milliseconds
[LightGBM] [Warning] Connecting to rank 2 failed, waiting for 338 milliseconds
[LightGBM] [Warning] Connecting to rank 2 failed, waiting for 439 milliseconds
[LightGBM] [Warning] Connecting to rank 2 failed, waiting for 570 milliseconds
[LightGBM] [Warning] Connecting to rank 2 failed, waiting for 570 milliseconds
[LightGBM] [Warning] Connecting to rank 2 failed, waiting for 741 milliseconds
[LightGBM] [Warning] Connecting to rank 2 failed, waiting for 741 milliseconds
[...]

PR #3897, opened 2021-02-03
[dask] test training when a worker has no data

jmoralez:
This includes a test to check that the training process succeeds even when a worker has no data, e.g. verify that #3882 is solved. 

I wanted to persist the collections to a specific worker with something like:
```python
for worker in workers:
    dX, dy, dw = client.persist([dX, dy, dw], workers=worker)
```
However I got: `TypeError: unhashable type: 'Array'`.

So in the test I just create a collection with one chunk, persist it and check that it is only in one worker. Any feedback to make this test more robust is welcome.

### DIFF
--- a/tests/python_package_test/test_dask.py
+++ b/tests/python_package_test/test_dask.py
@@ -5,7 +5,7 @@ import inspect
 import joblib
 import pickle
 import socket
-from itertools import groupby
+from itertools import groupby, product
 from os import getenv
 from sys import platform
 
@@ -35,9 +35,12 @@ from .utils import make_ranking
 # see https://distributed.dask.org/en/latest/api.html#distributed.Client.close
 CLIENT_CLOSE_TIMEOUT = 120
 
+tasks = ['classification', 'regression', 'ranking']
 data_output = ['array', 'scipy_csr_matrix', 'dataframe']
 data_centers = [[[-4, -4], [4, 4]], [[-4, -4], [4, 4], [-4, 4]]]
 group_sizes = [5, 5, 5, 10, 10, 10, 20, 20, 20, 50, 50]
+tasks_outputs = [(task, output) for task, output in product(tasks, data_output)
+                 if (task, output) != ('ranking', 'scipy_csr_matrix')]
 
 pytestmark = [
     pytest.mark.skipif(getenv('TASK', '') == 'mpi', reason='Fails to run with MPI interface'),
@@ -460,7 +463,7 @@ def test_ranker(output, client, listen_port, group):
     client.close(timeout=CLIENT_CLOSE_TIMEOUT)
 
 
-@pytest.mark.parametrize('task', ['classification', 'regression', 'ranking'])
+@pytest.mark.parametrize('task', tasks)
 def test_training_works_if_client_not_provided_or_set_after_construction(task, listen_port, client):
     if task == 'ranking':
         _, _, _, _, dX, dy, _, dg = _create_ranking_data(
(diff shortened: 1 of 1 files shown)


jameslamb (changes requested):
Thanks very much! I have some ideas to simplify this, without needing to introduce a new function to check that the data is only on one worker.

1. Assert that `client.nworkers > 1`, just to be sure the client fixture is for a multi-worker cluster.
2. Just repartition the data to 1 partition (`.repartition(npartitions=1)` for data frame, `.rechunk((X.shape[0], X.shape[1]))`  for array). assert that `dX.npartitions == 1`. If you have only 1 partition, you know for sure only one worker has data!

**other requests**

Please make sure this test is run for ranking, classification, and regression. You can borrow from https://github.com/microsoft/LightGBM/blob/876bfe5f6c756e946594e3316fec716ae358de18/tests/python_package_test/test_dask.py#L463-L487 to see how to do that.

For training on a single worker, the result should be identical to non-Dask training. So please also train a regular `lightgbm.sklearn.LGBM[Classifier/Regressor/Ranker]` and then  check that `assert_eq(dask_model.predict(dX).compute(), local_model.predict(X))`. You can see the other tests in this file for references on how to do that, or `@` me if you have any questions.

jmoralez:
Hi James. I've included the different tasks and outputs. I just saw the PR you made for the categoricals and saw the new output type, I'll wait til that gets merged to include it.

tests/python_package_test/test_dask.py
```diff
@@ -805,6 +808,67 @@ def f(part):
         assert 'foo' in str(info.value)
 
 
+@pytest.mark.parametrize('task,output', tasks_outputs)
+def test_training_having_workers_without_data(client, task, output):
```
jameslamb, line 812:
```suggestion
def test_training_succeeds_even_if_some_workers_do_not_have_any_daata(client, task, output):
```

I'd prefer a more informative name here.

tests/python_package_test/test_dask.py
```diff
@@ -805,6 +808,67 @@ def f(part):
...
+
+    def repartition(collection):
```
jameslamb, line 814:
```suggestion
    def repartition(collection):
        """Merge the parts of a Dask collection into a single partition"""
```

I love this function! But it has a very general name, so I'd like to recommend adding this docstring so others working on this code in the future will understand what its purpose is.

tests/python_package_test/test_dask.py
```diff
@@ -805,6 +808,67 @@ def f(part):
...
+            group=None
+        )
+        dask_model_factory = lgb.DaskLGBMRanker
+        local_model_factory = lgb.LGBMRanker
+    else:
+        X, y, w, dX, dy, dw = _create_data(
+            objective=task,
+            output=output
+        )
+        g = dg = None
```
jameslamb, line 833:
```suggestion
        g = None
        dg = None
```

Please put these on separate lines.

jameslamb (changes requested):
Thanks so much! I like the changes you've made. We're starting to have enough of these tests where `tasks` is parameterized, makes sense to concentrate it at the top of `tetst_dask.py` like you did.

I left a few minor suggestions. Other than those, I agree with your proposal to wait until #3908 is merged.

jmoralez:
Hi James. I've included your comments and the `dataframe-with-categorical` data output. Looking forward to your review.

jameslamb (approved):
This look great! Thanks very much for the help

PR #3898, opened 2021-02-03
[docs] fix typo: one-hot coding should be one-hot encoding

gauravchopracg:


docs/Python-Intro.rst
```diff
@@ -97,7 +97,7 @@ In LightGBM, the validation data should be aligned with training data.
     train_data = lgb.Dataset(data, label=label, feature_name=['c1', 'c2', 'c3'], categorical_feature=['c3'])
 
 LightGBM can use categorical features as input directly.
-It doesn't need to convert to one-hot coding, and is much faster than one-hot coding (about 8x speed-up).
+It doesn't need to convert to one-hot coding, and is much faster than one-hot encoding (about 8x speed-up).
```
StrikerRUS, line 100:
```suggestion
It doesn't need to convert to one-hot encoding, and is much faster than one-hot encoding (about 8x speed-up).
```

StrikerRUS (changes requested):
Thanks @gauravchopracg !
I think there is one more place in this sentence where `coding -> encoding` can be applyed.

docs/Python-Intro.rst
jameslamb: I reopened this and committed the suggestion. @gauravchopracg please note in the future that you need to click "commit" to apply a suggestion. Marking the comment thread "resolved" just hides it and does not apply changes to the code. It's confusing 😄

jameslamb (approved):
Thanks for taking the time to contribute! We hope you'll contribute more in the future.

StrikerRUS (approved):

Issue #2638, opened 2019-12-18
Support ingestion of training data from parquet files in the CLI version


ogru:
## Summary
Implement a feature for ingesting the training data from multiple parquet files with the CLI version. 

## Motivation

In memory efficient model fitting with big data we are often utilizing the CLI version to read the data and then convert it into binaries in order to reduce the memory footprint. There are two bottlenecks with such approach: 1) The data needs to be in one big csv file which often requires time consuming file concatenation 2) If using pandas to write the files into many chunks, the csv format writing tends to be inefficient. Since pandas can write into parquet quite efficiently, it would be awesome if we could read such parquet files directly into the LightGBM CLI version. This would remove the need to do csv file writing and concatenation.

## Description
Have the LightGBM CLI version read one or many parquet files. In the case of many, have the code read all parquet files inside a folder. 

## References
None, it is a productivity hack.

StrikerRUS:
@ogru Thank you very much for posting this request here!

Just one question: can't [MMLSpark with native parquet support](https://github.com/microsoft/LightGBM/issues/1286#issuecomment-517380041) fit for your scenario?

ogru:
Hi @StrikerRUS, 

Thank you for the quick reply. I am aware of the MMLSpark solution you referenced. The problem is not that we cannot provide parquet data to LightGBM, we can always load it into python/pandas and proceed in the conventional manner. However, this approach is insanely memory expensive, mostly due to pandas. I doubt that spark would be more efficient than pandas on a single machine (i.e. GCP VMs). 

To overcome memory bottlenecks with pandas, we utilize the CLI version which is super efficient. The problem is that the CLI version wants ONE training file as input. If the training data is in many files (csv of pq), which it usually is, concatenating them is a huge pain in the neck. If it is all going to end up inside LightGBM anyways, why not reading individual files and concatenating the data inside LightGBM? 

I am not fixated on parquet, reading multiple csv files would also be a huge improvement. 

Many thanks!

StrikerRUS:
@ogru Thanks for the clarification! 

We already have a feature request for possibility to read data from multiple files. So I'm going to reference your experience from there.

Issue #2031, opened 2019-02-26
Multi file as input for LightGBM


sosososo:
When I want to use LightGBM on 'Aether' ( a platform in Microsoft), multi files as input will be faster to upload or set up folder, but current LightGBM don't support multi files or dataset as input. Besides that, If we want to use it, we have to merge multi files as one file, which would be time-consuming especially when data is bigger. I wonder will we support it in future?

andrewliuxxx:
good advice，i‘m supporting file folder as input in my private lgb。

StrikerRUS:
@andrewliuxxx Great! Would you mind creating a PR?

snoe925:
I have also been working on fixing pipe read support  e.g. ```data=<(process on the fly) valid_data=<(process on the fly ..)```

StrikerRUS:
Closed in favor of being in #2302. We decided to keep all feature requests in one place.

Welcome to contribute this feature! Please re-open this issue (or post a comment if you are not a topic starter) if you are actively working on implementing this feature.

StrikerRUS:
One more request for using multiple files (preferably in parquet format) on one machine as input: #2638.

zwqjoy:
@StrikerRUS  Hi, Does have a detail date to support the training data from multiple parquet files with the CLI version ?

StrikerRUS:
@zwqjoy Hey! As of my knowledge, no one has picked this feature request up yet. Maybe [MMLSpark](https://github.com/microsoft/LightGBM/issues/1286#issuecomment-517380041) or [dataset creation from multiple files in Python](https://github.com/microsoft/LightGBM/pull/4089) will fit your needs?

zwqjoy:
@StrikerRUS Thanks,   do you mean if I want train dist lightgbm(data in hadoop:   lots of part,   I need use MML Spark). not use CLI  DIST method.  because input not support   lots of part file ?

StrikerRUS:
@zwqjoy Yeah, exactly. CLI distributed version requires that entire file with training data should be presented on each machine. Please try MMLSpark for chunked data.

shiyu1994:
Actually, distributed training with CLI supports partitioning data into each machine. See `pre_partition`
https://lightgbm.readthedocs.io/en/latest/Parameters.html#pre_partition

StrikerRUS:
@shiyu1994 Ah, forgot about this option, thanks for correcting me!

zwqjoy:
[...]

PR #3900, opened 2021-02-03
Support multiple train data on single machine

cyfdecyf:
This is a quick and dirty hack which I implemented a month ago.

As there are requests #2031 #2638 for this feature, I'd like to send a PR and see if you are interested in adding this feature.

I've tested correctness by running on higgs dataset with similary config from [boosting_tree_benchmarks](https://github.com/guolinke/boosting_tree_benchmarks). With `deterministic=true` and compared the output model file and prediction result.

### DIFF
--- a/include/LightGBM/dataset_loader.h
+++ b/include/LightGBM/dataset_loader.h
@@ -15,11 +15,17 @@ namespace LightGBM {
 
 class DatasetLoader {
  public:
+  LIGHTGBM_EXPORT DatasetLoader(const Config& io_config, const PredictFunction& predict_fun, int num_class, const std::vector<const char*>& filenames);
+
   LIGHTGBM_EXPORT DatasetLoader(const Config& io_config, const PredictFunction& predict_fun, int num_class, const char* filename);
 
   LIGHTGBM_EXPORT ~DatasetLoader();
 
-  LIGHTGBM_EXPORT Dataset* LoadFromFile(const char* filename, int rank, int num_machines);
+  LIGHTGBM_EXPORT Dataset* LoadFromFile(const std::vector<const char*>& filenames, int rank, int num_machines);
+
+  LIGHTGBM_EXPORT Dataset* LoadFromFile(const char* filename, int rank, int num_machines) {
+    return LoadFromFile(std::vector<const char*>{filename}, rank, num_machines);
+  }
 
   LIGHTGBM_EXPORT Dataset* LoadFromFile(const char* filename) {
     return LoadFromFile(filename, 0, 1);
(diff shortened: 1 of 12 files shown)


guolinke:
@shiyu1994  can you help to review this PR?

include/LightGBM/dataset_loader.h
```diff
@@ -40,25 +46,37 @@ class DatasetLoader {
                                                         const std::unordered_set<int>& categorical_features);
 
  private:
-  Dataset* LoadFromBinFile(const char* data_filename, const char* bin_filename, int rank, int num_machines, int* num_global_data, std::vector<data_size_t>* used_data_indices);
+  Dataset* LoadFromBinFile(const std::vector<const char*>& data_filename, const char* bin_filename, int rank, int num_machines, int* num_global_data, std::vector<data_size_t>* used_data_indices);
 
-  void SetHeader(const char* filename);
+  void SetHeader(const char* filenames);
```
shiyu1994, line 51: Change `filenames` back to `filename`? Since only the first file is used for `SetHeader`.

include/LightGBM/dataset_loader.h
```diff
@@ -40,25 +46,37 @@ class DatasetLoader {
...
+  std::vector<std::string> LoadTextDataToMemory(const char* filename, const Metadata& metadata, int rank, int num_machines, int* num_global_data, std::vector<data_size_t>* used_data_indices) {
+    return LoadTextDataToMemory(std::vector<const char*>{filename}, metadata, rank, num_machines, num_global_data, used_data_indices);
+  }
 
   std::vector<std::string> SampleTextDataFromMemory(const std::vector<std::string>& data);
 
-  std::vector<std::string> SampleTextDataFromFile(const char* filename, const Metadata& metadata, int rank, int num_machines, int* num_global_data, std::vector<data_size_t>* used_data_indices);
+  std::vector<std::string> SampleTextDataFromFile(const std::vector<const char*>& filenames, const Metadata& metadata, int rank, int num_machines, int* num_global_data, std::vector<data_size_t>* used_data_indices);
+
+  std::vector<std::string> SampleTextDataFromFile(const char* filename, const Metadata& metadata, int rank, int num_machines, int* num_global_data, std::vector<data_size_t>* used_data_indices) {
```
shiyu1994, line 65: It seems that this function is now unused. Maybe we can remove this.

cyfdecyf:
@shiyu1994 Thanks for taking time to review this PR.

As I said this is a quick and dirty hack. As I ignored many use cases, I guess this PR is far from being acceptable to be merged. If that's the case, I'd rather close this PR instead of continue working on this feature.

I'm now adding new features for `Dataset` related python bindings as this would be a more general approach for my problems.

The core idea is to support `two_round` in Python:

1. Load input data in Python, so I'm not limited to use text data
2. Do data sampling in Python, feed sampled data to C++ to construct bin mapper
3. Fead data in many batches to construct the final `Dataset`
    - So we can read data from multiple file
    - Hope this can also avoid triple memory usage I've seen when using Python interface

PR #3901, opened 2021-02-03
[docs] make link checks badge clickable

StrikerRUS:
For use-case: I open LightGBM repo -> I see last CRON link checks workflow is failing -> I want to rerun that workflow to see whether error was temporary or we need update broken link.


jameslamb (approved):
makes sense to me

PR #3904, opened 2021-02-03
[dask] remove unused private _client attribute

StrikerRUS:
Refer to https://github.com/microsoft/LightGBM/pull/3883#discussion_r569033103.

### DIFF
--- a/.vsts-ci.yml
+++ b/.vsts-ci.yml
@@ -1,7 +1,7 @@
 trigger:
   branches:
     include:
-    - master
+    - dask_cleanup
   tags:
     include:
     - v*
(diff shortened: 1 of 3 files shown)


jameslamb (commented):
Why is `_client` not required? The other properties on scikit-learn estimators seem to all have an attribute like that with a single leading underscore:

https://github.com/microsoft/LightGBM/blob/c324e891be58789f0423cd5c986e7396a524b3ce/python-package/lightgbm/sklearn.py#L320-L325

StrikerRUS:
> The other properties on scikit-learn estimators seem to all have an attribute like that with a single leading underscore:

Yeah, but this is not required by sklearn. It is made for storing underlying value of a property (with training `_`). You chose to use function for this purpose.

jameslamb (approved):
ok. I thought that https://github.com/microsoft/LightGBM/pull/3883#discussion_r569033103 referred to literally that one line you commented on, not the property completely.

PR #3906, opened 2021-02-03
[dask] remove extra 'client' kwarg in DaskLGBMRegressor

jameslamb:
Looks like I missed removing the `client` argument in `DaskLGBMRegressor` in #3883 , sorry!

I think in the future, we might be able to use `mypy` to catch API differences between the Dask package and its scikit-learn equivalents. For now, this PR removes that argument.

### DIFF
--- a/python-package/lightgbm/dask.py
+++ b/python-package/lightgbm/dask.py
@@ -710,7 +710,6 @@ class DaskLGBMRegressor(LGBMRegressor, _DaskLGBMModel):
         X: _DaskMatrixLike,
         y: _DaskCollection,
         sample_weight: Optional[_DaskCollection] = None,
-        client: Optional[Client] = None,
         **kwargs: Any
     ) -> "DaskLGBMRegressor":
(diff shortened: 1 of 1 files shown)


StrikerRUS:
> I think in the future, we might be able to use mypy to catch API differences between the Dask package and its scikit-learn equivalents. For now, this PR removes that argument.

I believe this approach you used for `init` with `inspect` is more reliable.
https://github.com/microsoft/LightGBM/blob/b1e000c04590b3437fd12da244528326aa0cc914/tests/python_package_test/test_dask.py#L816

StrikerRUS (approved):

jameslamb:
> I believe this approach you used for init with inspect is more reliable.

It is, but at the expense of a bit more maintenance burden. Because the test will have to explicitly articulate which keyword args are currently in the signature, and which are passed through with `**kwargs**`.

Anyway, for now the inspect() thing is probably good to do. I'll open a PR in a bit with tests like that for `fit()` and `predict()`.

PR #3911, opened 2021-02-05
[dask] Add unit tests that signatures are the same between Dask and scikit-learn estimators 

### LINKED ISSUE #3907: [dask] Add unit tests that signatures are the same between Dask and scikit-learn estimators
## Summary

In a refactoring in #3883, I made a silly mistake and forgot to remove a keyword argument from a method in `DaskLGBMRegressor`.

We should have unit tests on the Dask module that check that estimators' `.fit()` and `.predict()` methods have a similar signature to their scikit-learn equivalents.

## Description

The exposed keyword arguments in `.fit()` and `.predict()` in the Dask estimators are a subset of all the available keyword arguments in their scikit-learn equivalents.

For example,

```python
# lightgbm.dask.DaskLGBMClassifier
def fit(self, X, y, sample_weight=None, **kwargs)

# lightgbm.sklearn.LGBM
def fit(self, X, y,
        sample_weight=None, init_score=None,
        eval_set=None, eval_names=None, eval_sample_weight=None,
        eval_init_score=None, eval_metric=None, early_stopping_rounds=None,
        verbose=True, feature_name='auto', categorical_feature='auto',
        callbacks=None, init_model=None)
```

These tests should check the following:

* every keyword argument in the signature of the Dask estimator's method is in the sklearn estimator's method
* every keyword argument in the signature of the Dask estimator's method is in the same **position** as in the sklearn estimator's method (e.g. `X` is the first argument after `self`)
* every keyword argument in the signature of the Dask estimator's method has the same default value (or lack of default) as in the sklearn estimator's method

## References

Carried over from  this suggestion from @StrikerRUS https://github.com/microsoft/LightGBM/pull/3906#issuecomment-772853820.

See https://github.com/microsoft/LightGBM/blob/b1e000c04590b3437fd12da244528326aa0cc914/tests/python_package_test/test_dask.py#L816 for a reference on how to use `inspect` to do this.

GOusignu:
I tried to check for a sublist in sklearn's args which matches dask's (this ensures that order is checked), the same goes for the
default values. 

fixes #3907

### DIFF
--- a/tests/python_package_test/test_dask.py
+++ b/tests/python_package_test/test_dask.py
@@ -837,7 +837,6 @@ def test_dask_classes_and_sklearn_equivalents_have_identical_constructors_except
         (lgb.DaskLGBMRegressor.predict, lgb.LGBMRegressor.predict),
         (lgb.DaskLGBMRanker.fit, lgb.LGBMRanker.fit),
         (lgb.DaskLGBMRanker.predict, lgb.LGBMRanker.predict)
-
     ]
 )
 def test_dask_methods_and_sklearn_equivalents_have_similar_signatures(methods):
@@ -846,9 +845,11 @@ def test_dask_methods_and_sklearn_equivalents_have_similar_signatures(methods):
     dask_params = inspect.signature(methods[0]).parameters
     sklearn_params = inspect.signature(methods[1]).parameters
     assert dask_spec.args == sklearn_spec.args[:len(dask_spec.args)]
-    assert dask_spec.defaults == sklearn_spec.defaults[:len(dask_spec.defaults)]
-    assert dask_spec.varargs == sklearn_spec.varargs[:len(dask_spec.varargs)]
-    assert dask_spec.varkw == sklearn_spec.varkw[:len(dask_spec.varkw)]
-    assert dask_spec.kwonlyargs == sklearn_spec.kwonlyargs[:len(dask_spec.kwonlyargs)]
-    assert dask_spec.kwonlydefaults == sklearn_spec.kwonlydefaults[:len(dask_spec.kwonlydefaults)]
-    assert all([dask_params[key].default == sklearn_params[key].default for key in dask_params]) is True
+    assert dask_spec.varargs == sklearn_spec.varargs
+    if sklearn_spec.varkw:
+        assert dask_spec.varkw == sklearn_spec.varkw[:len(dask_spec.varkw)]
+    assert dask_spec.kwonlyargs == sklearn_spec.kwonlyargs
+    assert dask_spec.kwonlydefaults == sklearn_spec.kwonlydefaults
+    for key in dask_params:
+        error_msg = f"param '{key}' has different default values in the methods"
+        assert dask_params[key].default == sklearn_params[key].default, error_msg
--- a/tests/python_package_test/test_dask.py
+++ b/tests/python_package_test/test_dask.py
@@ -831,13 +831,13 @@ def test_dask_classes_and_sklearn_equivalents_have_identical_constructors_except
 @pytest.mark.parametrize(
     "methods",
     [   
-        (lgb.DaskLGBMClassifier.fit, lgb.LGBMClassifier.fit)
-        (lgb.DaskLGBMClassifier.predict, lgb.LGBMClassifier.predict)
+        (lgb.DaskLGBMClassifier.fit, lgb.LGBMClassifier.fit),
+        (lgb.DaskLGBMClassifier.predict, lgb.LGBMClassifier.predict),
 
-        (lgb.DaskLGBMRegressor.fit, lgb.LGBMRegressor.fit)
-        (lgb.DaskLGBMRegressor.predict, lgb.LGBMRegressor.predict)
+        (lgb.DaskLGBMRegressor.fit, lgb.LGBMRegressor.fit),
+        (lgb.DaskLGBMRegressor.predict, lgb.LGBMRegressor.predict),
 
-        (lgb.DaskLGBMRanker.fit, lgb.LGBMRanker.fit)
+        (lgb.DaskLGBMRanker.fit, lgb.LGBMRanker.fit),
         (lgb.DaskLGBMRanker.predict, lgb.LGBMRanker.predict)
 
     ]
--- a/tests/python_package_test/test_dask.py
+++ b/tests/python_package_test/test_dask.py
@@ -847,6 +847,6 @@ def test_dask_methods_and_sklearn_equivalents_have_similar_signatures(methods):
     sklearn_spec = inspect.getfullargspec(methods[1])
 
     # checks for a sublist in sklearn's kwargs which matches dask's
-    assert dark_spec.args is not None and sklearn_spec.args is not None
-    assert any(sklearn_spec.args[i: i + len(dask_spec.args)] == dask_spec.args for i in range(len(sklearn_spec.args) - len(dask_spec.args) + 1)) is True
-    assert any(sklearn_spec.defaults[i: i + len(dask_spec.defaults)] == dask_spec.defaults for i in range(len(sklearn_spec.defaults) - len(dask_spec.defaults) + 1)) is True
+    if dark_spec.args is not None and sklearn_spec.args is not None:
+        assert any(sklearn_spec.args[i: i + len(dask_spec.args)] == dask_spec.args for i in range(len(sklearn_spec.args) - len(dask_spec.args) + 1)) is True
+        assert any(sklearn_spec.defaults[i: i + len(dask_spec.defaults)] == dask_spec.defaults for i in range(len(sklearn_spec.defaults) - len(dask_spec.defaults) + 1)) is True
--- a/tests/python_package_test/test_dask.py
+++ b/tests/python_package_test/test_dask.py
@@ -852,5 +852,5 @@ def test_dask_methods_and_sklearn_equivalents_have_similar_signatures(methods):
     assert dask_spec.kwonlyargs == sklearn_spec.kwonlyargs
     assert dask_spec.kwonlydefaults == sklearn_spec.kwonlydefaults
     for param in dask_spec.args:
-        error_msg = f"param '{key}' has different default values in the methods"
+        error_msg = f"param '{param}' has different default values in the methods"
         assert dask_params[param].default == sklearn_params[param].default, error_msg
--- a/tests/python_package_test/test_dask.py
+++ b/tests/python_package_test/test_dask.py
@@ -833,7 +833,7 @@ def test_dask_classes_and_sklearn_equivalents_have_identical_constructors_except
     [
         (lgb.DaskLGBMClassifier.fit, lgb.LGBMClassifier.fit),
         (lgb.DaskLGBMClassifier.predict, lgb.LGBMClassifier.predict),
-        (lgb.DaskLGBMClassifier.predict_proba, lgb.LGBMClassifier.predict_proba),    
+        (lgb.DaskLGBMClassifier.predict_proba, lgb.LGBMClassifier.predict_proba), 
         (lgb.DaskLGBMRegressor.fit, lgb.LGBMRegressor.fit),
         (lgb.DaskLGBMRegressor.predict, lgb.LGBMRegressor.predict),
         (lgb.DaskLGBMRanker.fit, lgb.LGBMRanker.fit),
@@ -854,5 +854,5 @@ def test_dask_methods_and_sklearn_equivalents_have_similar_signatures(methods):
     for key in dask_params:
         error_msg1 = f"param '{key}' has different default values in the methods"
         error_msg2 = f"param '{key}' in dask's method has no sklearn equivalent"
-        assert key in sklearn_params is True, error_msg2
+        assert key in sklearn_params == True, error_msg2
         assert dask_params[key].default == sklearn_params[key].default, error_msg1
--- a/tests/python_package_test/test_dask.py
+++ b/tests/python_package_test/test_dask.py
@@ -847,6 +847,6 @@ def test_dask_methods_and_sklearn_equivalents_have_similar_signatures(methods):
     sklearn_spec = inspect.getfullargspec(methods[1])
 
     # checks for a sublist in sklearn's kwargs which matches dask's
-    if dark_spec.args is not None and sklearn_spec.args is not None:
+    if dask_spec.args is not None and sklearn_spec.args is not None:
         assert any(sklearn_spec.args[i: i + len(dask_spec.args)] == dask_spec.args for i in range(len(sklearn_spec.args) - len(dask_spec.args) + 1)) is True
         assert any(sklearn_spec.defaults[i: i + len(dask_spec.defaults)] == dask_spec.defaults for i in range(len(sklearn_spec.defaults) - len(dask_spec.defaults) + 1)) is True
--- a/tests/python_package_test/test_dask.py
+++ b/tests/python_package_test/test_dask.py
@@ -849,4 +849,6 @@ def test_dask_methods_and_sklearn_equivalents_have_similar_signatures(methods):
     # checks for a sublist in sklearn's kwargs which matches dask's
     if dask_spec.args is not None and sklearn_spec.args is not None:
         assert any(sklearn_spec.args[i: i + len(dask_spec.args)] == dask_spec.args for i in range(len(sklearn_spec.args) - len(dask_spec.args) + 1)) is True
+    
+    if dask_spec.defaults is not None and sklearn_spec.defaults is not None:
         assert any(sklearn_spec.defaults[i: i + len(dask_spec.defaults)] == dask_spec.defaults for i in range(len(sklearn_spec.defaults) - len(dask_spec.defaults) + 1)) is True
--- a/tests/python_package_test/test_dask.py
+++ b/tests/python_package_test/test_dask.py
@@ -854,4 +854,3 @@ def test_dask_methods_and_sklearn_equivalents_have_similar_signatures(methods):
         if key in sklearn_params:
             error_msg = f"param '{key}' has different default values in the methods"
             assert dask_params[key].default == sklearn_params[key].default, error_msg
-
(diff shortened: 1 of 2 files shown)


GOusignu:
bunch of syntax and styling mistakes, my bad

tests/python_package_test/test_dask.py
```diff
@@ -826,3 +826,29 @@ def test_dask_classes_and_sklearn_equivalents_have_identical_constructors_except
...
     assert dask_spec.args[-1] == 'client'
     assert dask_spec.defaults[-1] is None
+
+
+@pytest.mark.parametrize(
+    "methods",
+    [
+        (lgb.DaskLGBMClassifier.fit, lgb.LGBMClassifier.fit),
+        (lgb.DaskLGBMClassifier.predict, lgb.LGBMClassifier.predict),
+
```
jameslamb, line 836:
```suggestion
```

Please eliminate these unnecessary lines between entries

tests/python_package_test/test_dask.py
```diff
@@ -826,3 +826,29 @@ def test_dask_classes_and_sklearn_equivalents_have_identical_constructors_except
...
+        (lgb.DaskLGBMRegressor.fit, lgb.LGBMRegressor.fit),
+        (lgb.DaskLGBMRegressor.predict, lgb.LGBMRegressor.predict),
+
```
jameslamb, line 839:
```suggestion
```

tests/python_package_test/test_dask.py
```diff
@@ -826,3 +826,29 @@ def test_dask_classes_and_sklearn_equivalents_have_identical_constructors_except
...
+def test_dask_methods_and_sklearn_equivalents_have_similar_signatures(methods):
+    dask_spec = inspect.getfullargspec(methods[0])
+    sklearn_spec = inspect.getfullargspec(methods[1])
+
+    # checks for a sublist in sklearn's kwargs which matches dask's
+    if dask_spec.args is not None and sklearn_spec.args is not None:
+        assert any(sklearn_spec.args[i: i + len(dask_spec.args)] == dask_spec.args for i in range(len(sklearn_spec.args) - len(dask_spec.args) + 1)) is True
+
+    if dask_spec.defaults is not None and sklearn_spec.defaults is not None:
+        assert any(sklearn_spec.defaults[i: i + len(dask_spec.defaults)] == dask_spec.defaults for i in range(len(sklearn_spec.defaults) - len(dask_spec.defaults) + 1)) is True
```
jameslamb, line 854: Thanks very much for this! I think this code could be simplified a lot, and needs some additional comparison.

You can remove the `if dask_spec.args is not None`... blocks. All the methods being considered will have a non-empty set of arguments.

I don't think this list comprehension is necessary. Something like this should work:

```python
num_dask_args = len(dask_spec.args)
assert dask_spec.args == sklearn_spec.args[:num_dask_args]
assert dask_spec.defaults == sklearn_spec.defaults[:num_dask_args]
```

Basically, you can rely on the fact that the built-in `==` method for list objects will take care of most of what you've written out by hand.

<hr>

Can you please also include checks on all the other characteristics in the spec? https://github.com/microsoft/LightGBM/blob/fc2da2ec84e2dd9aa06421edcb5a625808eb546e/tests/python_package_test/test_dask.py#L819-L822

* `varargs`
* `varkw`
* `kwonlyargs`
* `kwonlydefaults`

If you need help inspecting these characteristics, you can use the `_asdict()` method on the results of `inspect.getfullargspec()`, like this:

```python
inspect.getfullargspec(methods[1])._asdict()
```

That produces a nice output like this

```text
OrderedDict([('args',
              ['self',
               'X',
               'y',
               'sample_weight',
               'init_score',
               'eval_set',
               'eval_names',
               'eval_sample_weight',
               'eval_init_score',
               'eval_metric',
               'early_stopping_rounds',
               'verbose',
               'feature_name',
               'categorical_feature',
               'callbacks',
               'init_model']),
             ('varargs', None),
             ('varkw', None),
             ('defaults',
              (None,
               None,
               None,
               None,
               None,
               None,
               None,
               None,
               True,
               'auto',
               'auto',
               None,
               None)),
             ('kwonlyargs', []),
             ('kwonlydefaults', None),
             ('annotations', {})])
```

</hr>

I did some of my own research while reviewing this, and I found a confusing behavior. It seem defaults for these specs are not being parsed the way I'd expect. For example, `lgb.DaskLGBMClassifier.fit()` takes 4 arguments. Its `.args` from `inspect.getfullargspec()` correctly has 4 items, but its `.defaults` only has 1!

```python
import inspect
import lightgbm as lgb
dask_spec = inspect.getfullargspec(lgb.DaskLGBMClassifier.fit)
sklearn_spec = inspect.getfullargspec(lgb.LGBMClassifier.fit)

print(dask_spec.args)
# ['self', 'X', 'y', 'sample_weight']

print(dask_spec.defaults)
# (None,)

print(sklearn_spec.args)
# ['self', 'X', 'y', 'sample_weight', 'init_score', 'eval_set', 'eval_names', 'eval_sample_weight', 'eval_init_score', 'eval_metric', 'early_stopping_rounds', 'verbose', 'feature_name', 'categorical_feature', 'callbacks', 'init_model']

print(sklearn_spec.defaults)
# (None, None, None, None, None, None, None, None, True, 'auto', 'auto', None, None)
```

After some investigation, I found that the `.defaults` list is an ordered list of defaults **only for the arguments that have defaults**. This makes it really hard to compare two different functions.

For the comparison, could you try looping over the values of `dask_spec.args`, and then checking that the defaults are the same as the equivalent sklearn default by using `inspect.signature()`?

For example:

```python
dask_params = inspect.signature(lgb.DaskLGBMClassifier.fit).parameters
sklearn_params = inspect.signature(lgb.LGBMClassifier.fit).parameters

print(dask_params['X'].default)
print(sklearn_params['X'].default)

dask_params['X'].default == sklearn_params['X'].default
```

jameslamb (changes requested):
Thanks for picking this up! I like the way you set this up with `pytest.mark.parametrize`. I left some suggestions for your consideration.

jameslamb:
I've edited your pull request description to include `fixes #3907`. This creates a link to that issue and will automatically close that issue when the PR is merged.

tests/python_package_test/test_dask.py
GOusignu: Thanks for the feedback! The only reason i put `if dask_spec.args is not None` , was that it was failing the code checks,
because of the error `TypeError: object of type 'NoneType' has no len()` , so I figured I had to put that.
 
As for the defaults, yeah it makes sense, because the `.defaults` list is containing only those args that have defaults, my approach wouldn't work as there could be arguments in between that don't have defaults. 

Will make the necessary changes.

jameslamb: Please let me know if you hit that NoneType error again. I'd be surprised to learn that `.args` for any of the `fit()` or `predict()` methods returns `None`.

Thanks for contributing!

tests/python_package_test/test_dask.py
```diff
@@ -826,3 +826,29 @@ def test_dask_classes_and_sklearn_equivalents_have_identical_constructors_except
...
+    sklearn_spec = inspect.getfullargspec(methods[1])
+    dask_params = inspect.signature(methods[0]).parameters
+    sklearn_params = inspect.signature(methods[1]).parameters
+    assert dask_spec.args == sklearn_spec.args[:len(dask_spec.args)]
+    assert dask_spec.defaults == sklearn_spec.defaults[:len(dask_spec.defaults)]
+    assert dask_spec.varargs == sklearn_spec.varargs[:len(dask_spec.varargs)]
+    assert dask_spec.varkw == sklearn_spec.varkw[:len(dask_spec.varkw)]
+    assert dask_spec.kwonlyargs == sklearn_spec.kwonlyargs[:len(dask_spec.kwonlyargs)]
+    assert dask_spec.kwonlydefaults == sklearn_spec.kwonlydefaults[:len(dask_spec.kwonlydefaults)]
+    assert all([dask_params[key].default == sklearn_params[key].default for key in dask_params]) is True
```
jameslamb, line 854: could you refactor this to give us a more informative error if one fails?

```suggestion
    for param in dask_spec.args:
         error_msg = f"param '{param}' has different default values"
         assert dask_params[param].default == sklearn_params[param], error_msg
```

tests/python_package_test/test_dask.py
```diff
@@ -826,3 +826,29 @@ def test_dask_classes_and_sklearn_equivalents_have_identical_constructors_except
...
+
+    ]
+)
+def test_dask_methods_and_sklearn_equivalents_have_similar_signatures(methods):
+    dask_spec = inspect.getfullargspec(methods[0])
+    sklearn_spec = inspect.getfullargspec(methods[1])
+    dask_params = inspect.signature(methods[0]).parameters
+    sklearn_params = inspect.signature(methods[1]).parameters
+    assert dask_spec.args == sklearn_spec.args[:len(dask_spec.args)]
+    assert dask_spec.defaults == sklearn_spec.defaults[:len(dask_spec.defaults)]
```
jameslamb, line 849: based on my explanation in https://github.com/microsoft/LightGBM/pull/3911#discussion_r571111631, you cannot rely on `.defaults` to work like this.

The first element of `.defaults` isn't "the default value of the first argument", it's "the default of the first argument with a default"

```suggestion
```

tests/python_package_test/test_dask.py
```diff
@@ -826,3 +826,29 @@ def test_dask_classes_and_sklearn_equivalents_have_identical_constructors_except
...
+    assert dask_spec.varargs == sklearn_spec.varargs[:len(dask_spec.varargs)]
```
jameslamb, line 850:
```suggestion
    assert dask_spec.varargs == sklearn_spec.varargs
```

Neither method allows variable positional args (with `*`), so this check can be simplified. This will make the tests fail if we ever add `*` on either side of this relationship in the future.

tests/python_package_test/test_dask.py
```diff
@@ -826,3 +826,29 @@ def test_dask_classes_and_sklearn_equivalents_have_identical_constructors_except
...
+    assert dask_spec.varkw == sklearn_spec.varkw[:len(dask_spec.varkw)]
```
jameslamb, line 851: I'm very sorry, I realize now we shouldn't check `.varkw`, since the sklearn functions don't have `**kwargs` pass-through.

https://github.com/microsoft/LightGBM/blob/fc2da2ec84e2dd9aa06421edcb5a625808eb546e/python-package/lightgbm/sklearn.py#L378-L384

Can you please change this check to this? That way, if we add the ability to pass kwargs through `.fit()` or `.predict()` on the sklearn estimators in the future, this test will check that the same name is used with `**` as the Dask estimators use.

```suggestion
    if sklearn_spec.varkw:
        assert dask_spec.varkw == sklearn_spec.varkw
```

tests/python_package_test/test_dask.py
```diff
@@ -826,3 +826,29 @@ def test_dask_classes_and_sklearn_equivalents_have_identical_constructors_except
...
+    assert dask_spec.kwonlyargs == sklearn_spec.kwonlyargs[:len(dask_spec.kwonlyargs)]
+    assert dask_spec.kwonlydefaults == sklearn_spec.kwonlydefaults[:len(dask_spec.kwonlydefaults)]
```
jameslamb, line 853: These are identical and from this point forward, should both change together. So this check can be simplified.

```suggestion
    assert dask_spec.kwonlyargs == sklearn_spec.kwonlyargs
    assert dask_spec.kwonlydefaults == sklearn_spec.kwonlydefaults
```

tests/python_package_test/test_dask.py
```diff
@@ -826,3 +826,29 @@ def test_dask_classes_and_sklearn_equivalents_have_identical_constructors_except
...
+@pytest.mark.parametrize(
+    "methods",
+    [
+        (lgb.DaskLGBMClassifier.fit, lgb.LGBMClassifier.fit),
+        (lgb.DaskLGBMClassifier.predict, lgb.LGBMClassifier.predict),
+        (lgb.DaskLGBMRegressor.fit, lgb.LGBMRegressor.fit),
+        (lgb.DaskLGBMRegressor.predict, lgb.LGBMRegressor.predict),
+        (lgb.DaskLGBMRanker.fit, lgb.LGBMRanker.fit),
+        (lgb.DaskLGBMRanker.predict, lgb.LGBMRanker.predict)
+
```
jameslamb, line 840: please remove this unnecessary empty line

```suggestion
```

jameslamb (changes requested):
Thanks! Please see my newest comments.

tests/python_package_test/test_dask.py
GOusignu: Thanks again! I'll make the changes

GOusignu:
Now I understand why I was getting the `TypeError: object of type 'NoneType' has no len()`.
It was because of that same issue that sklearn's `fit()` methods don't have `**kwargs` pass-through.

tests/python_package_test/test_dask.py
```diff
@@ -826,3 +826,31 @@ def test_dask_classes_and_sklearn_equivalents_have_identical_constructors_except
...
+    dask_params = inspect.signature(methods[0]).parameters
+    sklearn_params = inspect.signature(methods[1]).parameters
+    assert dask_spec.args == sklearn_spec.args[:len(dask_spec.args)]
+    assert dask_spec.varargs == sklearn_spec.varargs
+    if sklearn_spec.varkw:
+        assert dask_spec.varkw == sklearn_spec.varkw[:len(dask_spec.varkw)]
+    assert dask_spec.kwonlyargs == sklearn_spec.kwonlyargs
+    assert dask_spec.kwonlydefaults == sklearn_spec.kwonlydefaults
+    for key in dask_params:
+        if key in sklearn_params:
```
jameslamb, line 854: Please remove this if statement. If the Dask side adds an argument that is not in the equivalent scikit-learn side, we want this test to fail.

jameslamb (changes requested):
thanks! Please see my newest suggestion

tests/python_package_test/test_dask.py
```diff
@@ -826,3 +826,31 @@ def test_dask_classes_and_sklearn_equivalents_have_identical_constructors_except
     assert dask_spec.defaults[:-1] == sklearn_spec.defaults
     assert dask_spec.args[-1] == 'client'
     assert dask_spec.defaults[-1] is None
+
+
+@pytest.mark.parametrize(
+    "methods",
+    [
+        (lgb.DaskLGBMClassifier.fit, lgb.LGBMClassifier.fit),
+        (lgb.DaskLGBMClassifier.predict, lgb.LGBMClassifier.predict),
```
StrikerRUS, line 835: I believe we can add `predict_proba` here as well.
https://lightgbm.readthedocs.io/en/latest/pythonapi/lightgbm.LGBMClassifier.html#lightgbm.LGBMClassifier.predict_proba
https://lightgbm.readthedocs.io/en/latest/pythonapi/lightgbm.DaskLGBMClassifier.html#lightgbm.DaskLGBMClassifier.predict_proba

jameslamb: ooo thank you! for got about that one

tests/python_package_test/test_dask.py
GOusignu: Yeah I was having a doubt on this one. Isn't the test already failing for the `fit()` method? Because the dask version has `**kwargs` pass-through , but the sklearn version does not. When I run the code with the if statement, it gives the error:

`Keyerror: 'kwargs'` 
Which makes sense as the `fit` methods in sklearn's estimators have no variable keyword arguments.

If I get any `Keyerror` in the mentioned line, it means that the particular dask param has no sklearn coumterpart.
I propose to just catch the `Keyerror` and the `AssertionError` and display different error messages for them, for example

```suggestion
        try:
            assert dask_params[key].defaults == sklearn_params[key].defaults
        except AssertionError:
            print(error_msg1)
        except Keyerror:
            print(f"param {key} in dask's method has no sklearn equivalent")
```

GOusignu: Yeah sure. Its already failing the test though right? Because the `fit()` method of dask has `**kwargs` passthrough, but the sklearn method does not. When I run the assertion, it gives the following error for only the `fit` methods:

`Keyerror: 'kwargs'`

Which makes sense considering that sklearn `fit()` has no variable keyword arguments

GOusignu: I tried to use an additional assertion to catch this , something like 
`error_msg2 = f"param '{key}' in dask's method has no sklearn equivalent"`
`assert key in sklearn_params is True, error_msg2`

jameslamb: Are you sure? I just ran the following:

```python
import lightgbm as lgb
import inspect

dask_spec = inspect.getfullargspec(lgb.DaskLGBMClassifier.fit)
print(dask_spec.args)
```

and `kwargs` was not included in the list

> ['self', 'X', 'y', 'sample_weight']

GOusignu: I meant the `inspect.signature` method, 
```
import lightgbm as lgb
import inspect

dask_params = inspect.signature(lgb.LGBMClassifier.predict).parameters
print(dask_params)
```

The code above gives the following output
```
OrderedDict([('self', <Parameter "self">), ('X', <Parameter "X">), ('raw_score', <Parameter "raw_score=False">), ('start_iteration', <Parameter "start_iteration=0">), ('num_iteration', <Parameter "num_iteration=None">), ('pred_leaf', <Parameter "pred_leaf=False">), ('pred_contrib', <Parameter "pred_contrib=False">), ('kwargs', <Parameter "**kwargs">)])
```

`kwargs` is included

GOusignu: Oh yeah you mentioned it before, I'll change the loop to loop over `dask_spec.args` instead of `dask_params`

jameslamb: ah I see! `inspect` has several functions with names that sound like they do the same thing, so it can get confusing. If you loop over `.args` like I suggested in https://github.com/microsoft/LightGBM/pull/3911#discussion_r571111631, this check won't be necessary and we can rely on getting a KeyError if there are arguments in the Dask inreface not found in the sklearn equivalent

GOusignu: Yes exactly, I realised that

GOusignu: We only needed to check the positional arguments for their defaults, as we already had checked for keyword only defaults earlier

jameslamb (approved):
this looks great, thanks very much for the contribution!

GOusignu:
Thanks a lot for the guidance!

PR #3891, opened 2021-02-02
[python-package] fix some warnings from mypy

tara-jawahar:
Fixed some mypy type warnings in setup.py, dask.py, and sklearn.py

PR #3916, opened 2021-02-06
[python-package] Some mypy fixes

AlbertoEAF:
Related to https://github.com/microsoft/LightGBM/issues/3867.

Fixes some mypy false-positives.
Mypy is not smart enough to detect some things like alternative definitions to imports. Some type ignores are required.

### DIFF
--- a/python-package/lightgbm/callback.py
+++ b/python-package/lightgbm/callback.py
@@ -26,7 +26,7 @@ class EarlyStopException(Exception):
 
 # Callback environment used by callbacks
 CallbackEnv = collections.namedtuple(
-    "LightGBMCallbackEnv",
+    "CallbackEnv",
     ["model",
      "params",
      "iteration",
--- a/python-package/lightgbm/compat.py
+++ b/python-package/lightgbm/compat.py
@@ -16,12 +16,10 @@ except ImportError:
 
         pass
 
-
     class pd_DataFrame:  # type: ignore
         """Dummy class for pandas.DataFrame."""
 
         pass
-
 
     concat = None
     is_dtype_sparse = None
@@ -130,12 +128,10 @@ except ImportError:
 
         pass
 
-
     class dask_DataFrame:  # type: ignore
         """Dummy class for dask.dataframe.DataFrame."""
 
         pass
-
 
     class dask_Series:  # type: ignore
         """Dummy class for dask.dataframe.Series."""
--- a/python-package/lightgbm/compat.py
+++ b/python-package/lightgbm/compat.py
@@ -14,8 +14,14 @@ except ImportError:
     class pd_Series:  # type: ignore
         """Dummy class for pandas.Series."""
 
+        pass
+
+
     class pd_DataFrame:  # type: ignore
         """Dummy class for pandas.DataFrame."""
+
+        pass
+
 
     concat = None
     is_dtype_sparse = None
@@ -47,6 +53,8 @@ except ImportError:
 
     class dt_DataTable:  # type: ignore
         """Dummy class for datatable.DataTable."""
+
+        pass
 
 
 """sklearn"""
--- a/python-package/lightgbm/compat.py
+++ b/python-package/lightgbm/compat.py
@@ -11,10 +11,10 @@ try:
 except ImportError:
     PANDAS_INSTALLED = False
 
-    class pd_Series: # type: ignore
+    class pd_Series:  # type: ignore
         """Dummy class for pandas.Series."""
 
-    class pd_DataFrame: # type: ignore
+    class pd_DataFrame:  # type: ignore
         """Dummy class for pandas.DataFrame."""
 
     concat = None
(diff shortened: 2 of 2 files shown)


python-package/lightgbm/callback.py
```diff
@@ -26,7 +26,7 @@ def __init__(self, best_iteration, best_score):
 
 # Callback environment used by callbacks
 CallbackEnv = collections.namedtuple(
-    "LightGBMCallbackEnv",
+    "CallbackEnv",
```
jameslamb, line 29: what mypy warning does this address? I'm confused by this change.

python-package/lightgbm/sklearn.py
```diff
@@ -791,7 +791,7 @@ def fit(self, X, y,
                     categorical_feature=categorical_feature, callbacks=callbacks, init_model=init_model)
         return self
 
-    _base_doc = LGBMModel.fit.__doc__
+    _base_doc = LGBMModel.fit.__doc__ or ""
     _base_doc = (_base_doc[:_base_doc.find('group :')]
```
jameslamb, line 795: can you please revert these `_base_doc` fixes? Another contributor is already helping with them on #3891

python-package/setup.py
```diff
@@ -322,7 +322,7 @@ def run(self):
     if os.path.isfile(os.path.join(CURRENT_DIR, os.path.pardir, 'VERSION.txt')):
         copy_file(os.path.join(CURRENT_DIR, os.path.pardir, 'VERSION.txt'),
                   os.path.join(CURRENT_DIR, 'lightgbm', 'VERSION.txt'),
-                  verbose=0)
+                  verbose=0) # type: ignore
```
jameslamb, line 325: can you please revert this too? #3891 is already fixing this

jameslamb (changes requested):
Thanks! Please see my comments. Can you also please share the specific mypy warnings your PR addresses? That will help us to understand the changes.

python-package/lightgbm/callback.py
AlbertoEAF: The declared type `LightGBMCallbackEnv` is not the same name as the variable that holds that definition. That is unnecessary and confusing. Mypy complains about that.

python-package/lightgbm/sklearn.py
AlbertoEAF: done

python-package/lightgbm/callback.py
jameslamb: I was asking for the specific warning. I ran `mypy --ignore-missing imports python-package/`, and looks like it is this

> python-package/lightgbm/callback.py:28: error: First argument to namedtuple() should be 'CallbackEnv', not 'LightGBMCallbackEnv'

python-package/lightgbm/compat.py
```diff
@@ -11,16 +11,12 @@
 except ImportError:
     PANDAS_INSTALLED = False
 
-    class pd_Series:
+    class pd_Series: # type: ignore
```
jameslamb, line 14:
```suggestion
    class pd_Series:  # type: ignore
```

please add two spaces before each of these comments, to match our linting rules

https://github.com/microsoft/LightGBM/pull/3916/checks?check_run_id=1894201789

> ./python-package/lightgbm/compat.py:14:21: E261 at least two spaces before inline comment
./python-package/lightgbm/compat.py:17:24: E261 at least two spaces before inline comment
./python-package/lightgbm/compat.py:48:24: E261 at least two spaces before inline comment
./python-package/lightgbm/compat.py:122:22: E261 at least two spaces before inline comment
./python-package/lightgbm/compat.py:125:26: E261 at least two spaces before inline comment
./python-package/lightgbm/compat.py:128:23: E261 at least two spaces before inline comment

python-package/lightgbm/compat.py
```diff
@@ -11,16 +11,12 @@
...
         """Dummy class for pandas.Series."""
 
-        pass
```
jameslamb, line 17: can you please put these `pass` statements back? I understand that they aren't strictly necessary and that some linters like `pylint` will tell you to remove them, but we have a preference for using `pass` statements in cases like this to make it clear more explicit that the classes don't do anything by themselves.

`mypy` does not care about these, as far as I know, so I don't think the change has an impact on `mypy`'s output

jameslamb (changes requested):
Thanks! I left a few additional suggestions.

jameslamb:
@AlbertoEAF are you still interested in helping out on this?

AlbertoEAF:
Thanks for the ping James, done and rebased ;)

jameslamb (changes requested):
Looks good, thanks! Can you please just revert the extra lines you've added?

That will resolve these linting errors.

```text
./python-package/lightgbm/compat.py:20:5: E303 too many blank lines (2)
./python-package/lightgbm/compat.py:26:5: E303 too many blank lines (2)
./python-package/lightgbm/compat.py:134:5: E303 too many blank lines (2)
./python-package/lightgbm/compat.py:140:5: E303 too many blank lines (2)
```

jameslamb (approved):
looks good, thanks for the help!

StrikerRUS (approved):

PR #3920, opened 2021-02-07
[ci] encode token in the same way as e-mail address for RHub

StrikerRUS:
Sorry, I have no idea why I didn't do this earlier...


.ci/run_rhub_solaris_checks.R
```diff
@@ -9,22 +9,30 @@ email <- c(
...
 } else {
     null_file <- "/dev/null"
 }
+
 sink(file = null_file)
+rhub::validate_email(
+    email = intToUtf8(email - 42L)
+    , token = intToUtf8(token - 42L)
+)
+sink()
```
StrikerRUS, line 28: Prevent e-mail is being printed in logs. I believe that bots can also parse logs for e-mails printed in pain text.

image

jameslamb: ooo good catch, didn't think about that

jameslamb (approved):
Seems ok to me, thanks

PR #3032, opened 2020-04-29
[docs] updated bug report template

jameslamb:
I'd like to update the bug report template. I have seen the following things in recent issues that I think we should try to nudge people to avoid.

* ambiguous environment information
* using screenshots instead of code snippets

Some examples of "ambiguous environment information" from recent issues:

image

image

image

image

PR #3921, opened 2021-02-07
[docs] simplify bug report template

jameslamb:
I've observed that people reporting bugs in this project very very rarely use the issue template. I think that we're asking for too much information, and that people just get overwhelmed and ignore it entirely.

I've noticed that other projects like https://github.com/PrefectHQ/prefect/issues and https://github.com/dask/distributed/issues get a much higher % of issues following their template, and I think this is because their templates are so much simpler.

This PR proposes simplifying the issue template for bug reports greatly, to hopefully get higher-quality reports.

image

### DIFF
--- a/.github/ISSUE_TEMPLATE/BUG_REPORT.md
+++ b/.github/ISSUE_TEMPLATE/BUG_REPORT.md
@@ -7,7 +7,7 @@ about: Something isn't working as expected? Here is the right place to report.
 <!-- A clear description of the bug -->
 
 ## Reproducible example
-<!-- Minimal code that exhibits this behavior.-->
+<!-- Minimal code that exhibits this behavior -->
 
 ## Environment info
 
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/BUG_REPORT.md
```diff
@@ -3,65 +3,24 @@ name: Bug Report 🐞
...
-* R package
-* Python package
-* C++ library
-* Command Line Interface (CLI)
-* Java via SWIG
-* Other (please specify)
--->
-LightGBM component: 
+## Reproducible example
+<!-- Minimal code that exhibits this behavior.-->
```
StrikerRUS, line 10:
```suggestion
<!-- Minimal code that exhibits this behavior -->
```

StrikerRUS (approved):
Thanks!
> I think that we're asking for too much information, and that people just get overwhelmed and ignore it entirely.

Yeah, I was afraid of this: https://github.com/microsoft/LightGBM/pull/3032#pullrequestreview-402606107.

PR #3927, opened 2021-02-08
Optimize array-from-ctypes in basic.py

asford:
Approximately %80 of runtime when loading "low column count, high row
count" DataFrames into Datasets is consumed in `np.fromiter`, called
as part of the `Dataset.get_field` method.

This is particularly pernicious hotspot, as unlike other ctypes-based
methods this is a hot loop over a python iterator loop and causes
significant GIL-contention in multi-threaded applications.

Replace `np.fromiter` with a direct call to `np.ctypeslib.as_array`,
which allows a single-shot `copy` of the underlying array.

This reduces the load time of a ~35 million row categorical dataframe
with 1 column from ~5 seconds to ~1 second, and allows multi-threaded
execution.

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -97,7 +97,7 @@ def list_to_1d_numpy(data, dtype=np.float32, name='list'):
 def cfloat32_array_to_numpy(cptr, length):
     """Convert a ctypes float pointer array to a numpy array."""
     if isinstance(cptr, ctypes.POINTER(ctypes.c_float)):
-        return np.fromiter(cptr, dtype=np.float32, count=length)
+        return np.ctypeslib.as_array(cptr, shape=(length,)).copy()
     else:
         raise RuntimeError('Expected float pointer')
 
@@ -105,7 +105,7 @@ def cfloat32_array_to_numpy(cptr, length):
 def cfloat64_array_to_numpy(cptr, length):
     """Convert a ctypes double pointer array to a numpy array."""
     if isinstance(cptr, ctypes.POINTER(ctypes.c_double)):
-        return np.fromiter(cptr, dtype=np.float64, count=length)
+        return np.ctypeslib.as_array(cptr, shape=(length,)).copy()
     else:
         raise RuntimeError('Expected double pointer')
 
@@ -113,7 +113,7 @@ def cfloat64_array_to_numpy(cptr, length):
 def cint32_array_to_numpy(cptr, length):
     """Convert a ctypes int pointer array to a numpy array."""
     if isinstance(cptr, ctypes.POINTER(ctypes.c_int32)):
-        return np.fromiter(cptr, dtype=np.int32, count=length)
+        return np.ctypeslib.as_array(cptr, shape=(length,)).copy()
     else:
         raise RuntimeError('Expected int32 pointer')
 
@@ -121,7 +121,7 @@ def cint32_array_to_numpy(cptr, length):
 def cint64_array_to_numpy(cptr, length):
     """Convert a ctypes int pointer array to a numpy array."""
     if isinstance(cptr, ctypes.POINTER(ctypes.c_int64)):
-        return np.fromiter(cptr, dtype=np.int64, count=length)
+        return np.ctypeslib.as_array(cptr, shape=(length,)).copy()
     else:
         raise RuntimeError('Expected int64 pointer')
 


asford:
Profile shot, v3.1.1 via `conda-forge`:

image

Post-patch, v3.1.1:

image

jameslamb:
Thanks so much for taking the time to contribute @asford ! I think we'll need a little time to research this and might have some followup questions, but the benchmarking graphic and the fact that all the tests are passing is really encouraging.

asford:
My pleasure, thanks for all the effort on y'alls part maintaining the tool.

Not sure what would be useful from an review perspective, but this is actually quite easy to evaluate by monkeypatching the associated functions. If it'd be useful I'd be happy to setup a colab or binder with the head-to-head demo or setup a pytest-benchmark test case between the two code paths.

guolinke:
Will the copy increase the memory usage?

asford:
The existing `np.fromiter` based implementation is creating a copy of the underlying float array, just in a highly inefficient way by iterating over the pointer via the ctypes interface. This change preserves the same semantics by copying the contents of the internal buffer.

Though there could be a *minor* reduction is memory consumption by making this return a view, rather than copy, of the ctypes pointer but this opens a *huge* can of worms wrt lifetime of the underlying buffer. I've no idea idea how lightgbm's internal memory management strategy is structured, but t's far-safer to perform the copy so that the standard memory management semantics for numpy are obeyed. (ie. the array contents are valid for the lifetime of the ndarray object).

As you can see from the second, post fix, profiling result there are a lot of big existing gains to be had in tuning the "Dataset from DataFrame" story. For example, the current categorical normalization performs at-least-one full-size copy of the source dataframe and performs a slow single-threaded normalization pass.

@jameslamb given your work with dask dataframes you may find this strategy useful. I was seeing some GME-circa-January '21 gains in dataset construction performance. 🚀 

<details>

<summary> Code snippet for threaded, multi-column Dataset construction </summary>

```python
def merge_dataset(target: lgb.Dataset, other: lgb.Dataset) -> lgb.Dataset:
    """In-place merge of lightgbm dataset features.
    Args:
        target: Dataset, updated in place.
        other: Dataset, features added to target.
    Returns:
        In-place updated target.
    """

    # LightGBM contains a booster reference, in C, which is handled via
    # "add_features_for".

    # pandas-categorical metadata, needed to ensure proper cat encoding,
    # is stored on the python-level object and must be manually merged

    if target.pandas_categorical or other.pandas_categorical:
        pandas_categorical = (
            target.pandas_categorical if target.pandas_categorical else []
        ) + (other.pandas_categorical if other.pandas_categorical else [])
    else:
        pandas_categorical = None

    categorical_column = target.params.get("categorical_column", []) + [
        c + target.num_feature() for c in other.params.get("categorical_column", [])
    ]

    target.add_features_from(other)

    target.pandas_categorical = pandas_categorical
    if categorical_column:
        target.params["categorical_column"] = categorical_column

    return target


def construct_dataset_threaded(
    df: pd.DataFrame, pre_apply: Callable[[pd.DataFrame], pd.DataFrame] = lambda df: df
) -> lgb.Dataset:
    """Multi-threaded Dataset setup.
    Args:
        df: Target dataframe, multi-threaded construct on a per-column basis.
        pre_apply: Pre-process a single-column frame before Dataset constructor.
    Returns:
        Dataset, with feature-per-column.
    """

    @dask.delayed
    def d_construct(col_df: pd.DataFrame) -> lgb.Dataset:
        return lgb.Dataset(pre_apply(col_df)).construct()

    with dask.config.set(scheduler="threads"):
        (d_lgb_datasets,) = dask.persist([d_construct(df[[c]]) for c in df.columns])

        dataset = d_lgb_datasets[0].compute()
        for d_other in d_lgb_datasets[1:]:
            merge_dataset(dataset, d_other.compute())

        return dataset
```

</details>

StrikerRUS:
@asford Thanks a lot for your explanation!

For the reference, source code of `fromiter` function:
https://github.com/numpy/numpy/blob/b4491833d636344ef295f1609bcf4742c1d94d10/numpy/core/src/multiarray/multiarraymodule.c#L2206-L2231
https://github.com/numpy/numpy/blob/72be9ca10b5391fc036c4cd61830e168712b51ea/numpy/core/src/multiarray/ctors.c#L3731-L3848

I'm afraid only about that we are loosing `dtype` information with new interface. Will `numpy` preserve original type of `cptr` and will not cast for example `float` to default `float64`? According to the [source code of `as_array`](https://github.com/numpy/numpy/blob/fb215c76967739268de71aa4bda55dd1b062bc2e/numpy/ctypeslib.py#L499-L517), it will preserve, but I'd like to be 100% sure.

asford:
As far as I can tell, the requested types in these functions were probably to workaround the fact that `np.fromiter` may disregard the type of the input `ctypes` pointer.

As you mention, `as_array` infers the array dtype from the source pointer, so the ctype dtype is passed through. You could be extremely defensive and `assert` on the expected dtype before returning, however I'm not a yuge fan of asserts "on the way out".

The pull is is set to "Allow edits by maintainers". I'm totally fine with edits here to match the expectations of the project; will likely not be online to update this pull for several days.

StrikerRUS (approved):
LGTM based on great comments from @asford !

jameslamb (approved):
great contribution, thanks so much for the time and effort you put into improving LightGBM!

guolinke (approved):
Thank you!

PR #3872, opened 2021-01-28
v3.2.0 release

guolinke:
This is a stable release.
The next release may is 4.0.0, for many breaking changes. And it may take a long time to finish. 
So it better to fix the critical bugs in 3.2.0.

Please list the PR need to be merged in this release.

PR #3908, opened 2021-02-04
[dask] Support Dask dataframes with 'category' columns (fixes #3861)

jameslamb:
This PR fixes https://github.com/microsoft/LightGBM/issues/3861.

## Root Cause

Thanks to @jmoralez for figuring out that the root issue was `pandas` "category" columns for data frames! `lightgbm.dask._predict_part()` did some conversions thatt did not handle these columns correctly.

## Changes in this PR

1. Removes code in `_predict_part()` that converts a chunk of data to array before passing it through to `.predict()` on the underlying scikit-learn model.
    - LightGBM's scikit-learn estimators already support scoring directly on pandas data frames, so this was useless. And it was the root of the problem! Because when you use `.values` on a pandas "category" column, the results are strings. Passing a numpy array with string data into `.predict()` methods in LightGBM leads to errors.

2. Adds data frames with a "category" column to many of the Dask unit tests

## Reproducible example

I've added a reproducible example in https://github.com/microsoft/LightGBM/issues/3861#issuecomment-773011094. If you run the code on current `master`, it will fail with the error described in #3861. If you run that code on this branch, it will succeed 😀 .

## References

I found a few issues and PRs related to "category" columns in `pandas` being supported in the scikit-learn interface. Putting them here for anyone who finds this PR from search in the future.

* https://github.com/microsoft/LightGBM/pull/1979
* https://github.com/microsoft/LightGBM/pull/1766
* https://github.com/microsoft/LightGBM/pull/1764
* https://github.com/microsoft/LightGBM/blob/8ef874bd4804b54b48ef4d796c8110be671a6950/python-package/lightgbm/basic.py#L497-L500

Issue #3636, opened 2020-12-09
Wrong shape of predictions


pseudotensor:
Lightgbm 2.2.3 -> 3.1.1

Long term bug.

```
import lightgbm as lgb
import pickle
model, Xslice, kwargs_predict = pickle.load(open("tmp/h2oai/newlgbm.pkl", "rb"))
preds = model.predict_proba(Xslice, **kwargs_predict)
print("version: %s" % lgb.__version__)
assert preds.shape[0] == Xslice.shape[0], "%s %s" % (preds.shape[0], Xslice.shape[0])
```

[newlgbm.pkl.zip](https://github.com/microsoft/LightGBM/files/5664010/newlgbm.pkl.zip)

gives

```
version: 3.1.1
Traceback (most recent call last):
  File "lgbm_shapeissue.py", line 7, in <module>
    assert preds.shape[0] == Xslice.shape[0], "%s %s" % (preds.shape[0], Xslice.shape[0])
AssertionError: 4 163
```

Basically, if one has objective multiclass but have only 2 label types, then lgbm messes up and gives totally wrong results (wrong shape and preds nan).  Here preds.shape[1] is 326, which is wild, just normal predictions not pred_contribs etc.

guolinke:
@pseudotensor can this be reproduced from scratch, without pickle?

pseudotensor:
The pickle is pretty trivial file of model and data.  I'm sure it can be reproduced from scratch, but is it not enough?

guolinke:
@pseudotensor I think the training (how to get the model) is the root cause for this issue, not in the prediction stage. 
So it would be better to have an end-to-end example.

pseudotensor:
I don't have easy way to get full training case.  But as I mentioned, the model was set to multiclass objective but training data target only had 2 label types.

guolinke:
As you use `predict_proba`, which is from sklearn interface, not LightGBM native python interface. So I assume you use the `LGBMClassifier` to perform training.
however, In `LGBMClassifier`, we will first check the label, if it only has 2 label types, it will use binary loss. 

https://github.com/microsoft/LightGBM/blob/44a6fb7ffa646b469fc10475b3526c61239682ac/python-package/lightgbm/sklearn.py#L808-L812

So I am not sure how you perform the multi-class training, with the 2 label types. If you use the native LightGBM interface, you cannot use `predict_proba`.

rabyj:
[...]

PR #3930, opened 2021-02-09
[dask] [docs] Fix inaccuracies in API docs for Dask module (fixes #3871)

### LINKED ISSUE #3871: [RFC] [dask] [docs] Override details in API docs that differ between dask and sklearn interfaces
## Summary

The module `lightgbm.dask` contains model objects which have very similar APIs to those from `lightgbm.sklearn`, and which directly inherit from them. For example, `lightgbm.dask.DaskLGBMRegressor` inherits from `lightgbm.sklearn.DaskLGBMRegressor`.

We currently take advantage of this inheritance to also inherit doc strings. For example: https://github.com/microsoft/LightGBM/blob/b4b1b75ac582e99202062ec784269bd5b77f2018/python-package/lightgbm/dask.py#L452-L457

That code reads the parent class's `__doc__` and then patches it with some Dask-specific details.

There are still many places in the API docs which should be updated to show information specific to the Dask module. For example, the docs for the `X`, `y`, `group`, and `sample_weight` arguments to `.fit()` and `.predict()` should be updated to say that they expect Dask inputs (Dask Array, Dask DataFrame, Dask Series).

image

## Motivation

This change would make the API docs a reliable source of information for how to structure inputs for the Dask module, which would make it easier for users to get started. This has value as a complement to a long-form tutorial (#3814).

## Description

I can think of a few different ways to accomplish this.

### Option 1: custom templates

We could write documentation templates that are shared between the classes. Here's an oversimplified example.

```python
__fit_params_template = """
Build a gradient boosting model from the training set (X, y).

    Parameters
    ----------
    X : {x_type} of shape = [n_samples, n_features]
        Input feature matrix.
    y : {y_type} shape = [n_samples]
    sample_weight : {weight_type} of shape = [n_samples] or None, optional (default=None)
        Weights of training data.
    init_score : {init_score_type} of shape = [n_samples] or None, optional (default=None)
        Init score of training data.
"""

__DaskCollectionDescription = "A Dask DataFrame, Dask Array, or Dask Series"
__LocalArrayDescription = "A pandas dataframe, numpy array, or sparse matrix"

class LGBMClassifier:
    def fit(...):
        pass

    fit.__doc__ = __fit_params_template.format(
        x_type=__LocalArrayDescription,
        y_type=__LocalArrayDescription,
        weight_type=__LocalArrayDescription,
        init_score_type=__LocalArrayDescription,
    )

class DaskLGBMClassifier:
    def fit(...):
        pass

    fit.__doc__ = __fit_params_template.format(
        x_type=__DaskCollectionDescription,
        y_type=__DaskCollectionDescription,
        weight_type=__DaskCollectionDescription,
        init_score_type=__DaskCollectionDescription,
    )
```

Something similar is done in the R package to avoid writing the same documentation in multiple places in source control, even though it is used in multiple places in the 

* https://github.com/microsoft/LightGBM/blob/b4b1b75ac582e99202062ec784269bd5b77f2018/R-package/R/lightgbm.R#L1
* https://github.com/microsoft/LightGBM/blob/b4b1b75ac582e99202062ec784269bd5b77f2018/R-package/R/lgb.train.R#L4

### Option 2: use `sphinx-autodoc-typehints`

If the only difference we anticipate in the API docs is about the types of inputs, we could stop relying on typehints in docstrings, and instead use [sphinx-autodoc-typehints](https://pypi.org/project/sphinx-autodoc-typehints/). This extension creates the typehints in Sphinx docs automatically based on the hints in the code, without you needing to write them in docstrings. I think that might allow us to just inherit docstrings and get the hinting stuff for free.

### Option 3: just copy the docstrings and manually keep them in sync

We could stop having the Dask interface inherit the docstrings from `lightgbm.sklearn`, and instead just literally copy the sklearn docs into the Dask module. Then the type hints and any other information could be changed in the Dask module freely and easily.

## References

* DaskLGBMClassifier docs, for reference: https://lightgbm.readthedocs.io/en/latest/pythonapi/lightgbm.DaskLGBMClassifier.html#lightgbm.DaskLGBMClassifier .
* this issue was inspired by this comment: https://github.com/microsoft/LightGBM/pull/3866#discussion_r565016128

jameslamb:
This PR proposes a fix to https://github.com/microsoft/LightGBM/issues/3871.

* corrects type information (e.g. to say "dask Array" instead of "array-like")
* removes docs for keyword args that aren't currently supported (like `init_score` and various `eval_*` arguments)
* removes docs for custom objective, since that isn't supported yet (TODO: I'll write up a feature request tomorrow)

To minimize duplication, this PR moves the `sklearn` docstrings into template strings that are then formatted with `.format()`. This indirection is a little confusing, but we can at least rely on getting CI failures in the `check-docs` task, since `.format()` has to fill all template values.

```python
"{a} - {b}".format(a="hello")

# KeyError: 'b'
```

## Notes for Reviewers

* I opened this PR on the `docs/jlamb` branch so we can check on readthedocs builds: https://lightgbm.readthedocs.io/en/docs-jlamb/Python-API.html.
* there are two places in `LGBMModel.__init__` that have references that are incorrect for Dask, but when I tried to templatize it I got errors from Sphinx like `LGBMClassifier.__init__:66:Inline strong start-string without end-string.`. I'll try some more things, but it's ok if we can't figure out how to fix that for this PR since we're trying to get a release out (#3872 ) and the inaccuracies are minor.
    - `learning_rate` docs that says "you can use callbacks..."
    - `objective` doc that references scikit-learn classes ("Default: ‘regression’ for LGBMRegressor...")

### DIFF
--- a/python-package/lightgbm/dask.py
+++ b/python-package/lightgbm/dask.py
@@ -1,8 +1,8 @@
 # coding: utf-8
-"""Distributed training with LightGBM and Dask.distributed.
+"""Distributed training with LightGBM and dask.distributed.
 
 This module enables you to perform distributed training with LightGBM on
-Dask.Array and Dask.DataFrame collections.
+dask.Array and dask.DataFrame collections.
 
 It is based on dask-lightgbm, which was based on dask-xgboost.
 """
@@ -223,17 +223,17 @@ def _train(
     ----------
     client : dask.distributed.Client
         Dask client.
-    data : dask Array or dask DataFrame of shape = [n_samples, n_features]
+    data : Dask Array or Dask DataFrame of shape = [n_samples, n_features]
         Input feature matrix.
-    label : dask Array, dask DataFrame or dask Series of shape = [n_samples]
+    label : Dask Array, Dask DataFrame or Dask Series of shape = [n_samples]
         The target values (class labels in classification, real numbers in regression).
     params : dict
         Parameters passed to constructor of the local underlying model.
     model_factory : lightgbm.LGBMClassifier, lightgbm.LGBMRegressor, or lightgbm.LGBMRanker class
         Class of the local underlying model.
-    sample_weight : dask Array, dask DataFrame, Dask Series of shape = [n_samples] or None, optional (default=None)
+    sample_weight : Dask Array, Dask DataFrame, Dask Series of shape = [n_samples] or None, optional (default=None)
         Weights of training data.
-    group : dask Array, dask DataFrame, Dask Series of shape = [n_samples] or None, optional (default=None)
+    group : Dask Array, Dask DataFrame, Dask Series of shape = [n_samples] or None, optional (default=None)
         Group/query data.
         Only used in the learning-to-rank task.
         sum(group) = n_samples.
@@ -403,7 +403,7 @@ def _predict(
     ----------
     model : lightgbm.LGBMClassifier, lightgbm.LGBMRegressor, or lightgbm.LGBMRanker class
         Fitted underlying model.
-    data : dask Array or dask DataFrame of shape = [n_samples, n_features]
+    data : Dask Array or Dask DataFrame of shape = [n_samples, n_features]
         Input feature matrix.
     raw_score : bool, optional (default=False)
         Whether to predict raw scores.
@@ -420,11 +420,11 @@ def _predict(
 
     Returns
     -------
-    predicted_result : dask Array of shape = [n_samples] or shape = [n_samples, n_classes]
+    predicted_result : Dask Array of shape = [n_samples] or shape = [n_samples, n_classes]
         The predicted values.
-    X_leaves : dask Array of shape = [n_samples, n_trees] or shape = [n_samples, n_trees * n_classes]
+    X_leaves : Dask Array of shape = [n_samples, n_trees] or shape = [n_samples, n_trees * n_classes]
         If ``pred_leaf=True``, the predicted leaf of every tree for each sample.
-    X_SHAP_values : dask Array of shape = [n_samples, n_features + 1] or shape = [n_samples, (n_features + 1) * n_classes]
+    X_SHAP_values : Dask Array of shape = [n_samples, n_features + 1] or shape = [n_samples, (n_features + 1) * n_classes]
         If ``pred_contrib=True``, the feature contributions for each sample.
     """
     if not all((DASK_INSTALLED, PANDAS_INSTALLED, SKLEARN_INSTALLED)):
@@ -455,7 +455,7 @@ def _predict(
             **kwargs
         )
     else:
-        raise TypeError('Data must be either dask Array or dask DataFrame. Got %s.' % str(type(data)))
+        raise TypeError('Data must be either Dask Array or Dask DataFrame. Got %s.' % str(type(data)))
 
 
 class _DaskLGBMModel:
@@ -616,10 +616,10 @@ class DaskLGBMClassifier(LGBMClassifier, _DaskLGBMModel):
         )
 
     _base_doc = _lgbmmodel_doc_fit.format(
-        X_shape="dask Array or dask DataFrame of shape = [n_samples, n_features]",
-        y_shape="dask Array, dask DataFrame or dask Series of shape = [n_samples]",
-        sample_weight_shape="dask Array, dask DataFrame, Dask Series of shape = [n_samples] or None, optional (default=None)",
-        group_shape="dask Array, dask DataFrame, Dask Series of shape = [n_samples] or None, optional (default=None)"
+        X_shape="Dask Array or Dask DataFrame of shape = [n_samples, n_features]",
+        y_shape="Dask Array, Dask DataFrame or Dask Series of shape = [n_samples]",
+        sample_weight_shape="Dask Array, Dask DataFrame, Dask Series of shape = [n_samples] or None, optional (default=None)",
+        group_shape="Dask Array, Dask DataFrame, Dask Series of shape = [n_samples] or None, optional (default=None)"
     )
 
     # DaskLGBMClassifier does not support init_score, evaluation data, or early stopping
@@ -644,11 +644,11 @@ class DaskLGBMClassifier(LGBMClassifier, _DaskLGBMModel):
 
     predict.__doc__ = _lgbmmodel_doc_predict.format(
         description="Return the predicted value for each sample.",
-        X_shape="dask Array or dask DataFrame of shape = [n_samples, n_features]",
+        X_shape="Dask Array or Dask DataFrame of shape = [n_samples, n_features]",
         output_name="predicted_result",
-        predicted_result_shape="dask Array of shape = [n_samples] or shape = [n_samples, n_classes]",
-        X_leaves_shape="dask Array of shape = [n_samples, n_trees] or shape = [n_samples, n_trees * n_classes]",
-        X_SHAP_values_shape="dask Array of shape = [n_samples, n_features + 1] or shape = [n_samples, (n_features + 1) * n_classes]"
+        predicted_result_shape="Dask Array of shape = [n_samples] or shape = [n_samples, n_classes]",
+        X_leaves_shape="Dask Array of shape = [n_samples, n_trees] or shape = [n_samples, n_trees * n_classes]",
+        X_SHAP_values_shape="Dask Array of shape = [n_samples, n_features + 1] or shape = [n_samples, (n_features + 1) * n_classes]"
     )
 
     def predict_proba(self, X: _DaskMatrixLike, **kwargs: Any) -> dask_Array:
@@ -662,11 +662,11 @@ class DaskLGBMClassifier(LGBMClassifier, _DaskLGBMModel):
 
     predict_proba.__doc__ = _lgbmmodel_doc_predict.format(
         description="Return the predicted probability for each class for each sample.",
-        X_shape="dask Array or dask DataFrame of shape = [n_samples, n_features]",
+        X_shape="Dask Array or Dask DataFrame of shape = [n_samples, n_features]",
         output_name="predicted_probability",
-        predicted_result_shape="dask Array of shape = [n_samples, n_classes]",
-        X_leaves_shape="dask Array of shape = [n_samples, n_trees] or shape = [n_samples, n_trees * n_classes]",
-        X_SHAP_values_shape="dask Array of shape = [n_samples, n_features + 1] or shape = [n_samples, (n_features + 1) * n_classes]"
+        predicted_result_shape="Dask Array of shape = [n_samples, n_classes]",
+        X_leaves_shape="Dask Array of shape = [n_samples, n_trees] or shape = [n_samples, n_trees * n_classes]",
+        X_SHAP_values_shape="Dask Array of shape = [n_samples, n_features + 1] or shape = [n_samples, (n_features + 1) * n_classes]"
     )
 
     def to_local(self) -> LGBMClassifier:
@@ -767,10 +767,10 @@ class DaskLGBMRegressor(LGBMRegressor, _DaskLGBMModel):
         )
 
     _base_doc = _lgbmmodel_doc_fit.format(
-        X_shape="dask Array or dask DataFrame of shape = [n_samples, n_features]",
-        y_shape="dask Array, dask DataFrame or dask Series of shape = [n_samples]",
-        sample_weight_shape="dask Array, dask DataFrame, Dask Series of shape = [n_samples] or None, optional (default=None)",
-        group_shape="dask Array, dask DataFrame, Dask Series of shape = [n_samples] or None, optional (default=None)"
+        X_shape="Dask Array or Dask DataFrame of shape = [n_samples, n_features]",
+        y_shape="Dask Array, Dask DataFrame or Dask Series of shape = [n_samples]",
+        sample_weight_shape="Dask Array, Dask DataFrame, Dask Series of shape = [n_samples] or None, optional (default=None)",
+        group_shape="Dask Array, Dask DataFrame, Dask Series of shape = [n_samples] or None, optional (default=None)"
     )
 
     # DaskLGBMRegressor does not support init_score, evaluation data, or early stopping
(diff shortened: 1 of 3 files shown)


python-package/lightgbm/sklearn.py
```diff
@@ -937,6 +929,15 @@ def predict_proba(self, X, raw_score=False, start_iteration=0, num_iteration=Non
         else:
             return np.vstack((1. - result, result)).transpose()
 
+    predict_proba.__doc__ = _lgbmmodel_doc_predict.format(
+        description="Return the predicted probability for each class for each sample.",
+        X_shape="array-like or sparse matrix of shape = [n_samples, n_features]",
+        output_name="predicted_probability",
+        predicted_result_shape="array-like of shape = [n_samples] or shape = [n_samples, n_classes]",
```
StrikerRUS, line 936:
```suggestion
        predicted_result_shape="array-like of shape = [n_samples, n_classes]",
```

python-package/lightgbm/dask.py
```diff
@@ -626,7 +660,14 @@ def predict_proba(self, X: _DaskMatrixLike, **kwargs: Any) -> dask_Array:
             **kwargs
         )
 
-    predict_proba.__doc__ = LGBMClassifier.predict_proba.__doc__
+    predict_proba.__doc__ = _lgbmmodel_doc_predict.format(
+        description="Return the predicted probability for each class for each sample.",
+        X_shape="dask Array or dask DataFrame of shape = [n_samples, n_features]",
+        output_name="predicted_probability",
+        predicted_result_shape="dask Array of shape = [n_samples] or shape = [n_samples, n_classes]",
```
StrikerRUS, line 667: Not sure, but guess it is the same as for sklearn.
```suggestion
        predicted_result_shape="dask Array of shape = [n_samples, n_classes]",
```

python-package/lightgbm/dask.py
```diff
@@ -731,7 +792,14 @@ def predict(self, X: _DaskMatrixLike, **kwargs) -> dask_Array:
             **kwargs
         )
 
-    predict.__doc__ = LGBMRegressor.predict.__doc__
+    predict.__doc__ = _lgbmmodel_doc_predict.format(
+        description="Return the predicted value for each sample.",
+        X_shape="dask Array or dask DataFrame of shape = [n_samples, n_features]",
+        output_name="predicted_result",
+        predicted_result_shape="dask Array of shape = [n_samples] or shape = [n_samples, n_classes]",
```
StrikerRUS, line 799:
```suggestion
        predicted_result_shape="dask Array of shape = [n_samples]",
```

python-package/lightgbm/dask.py
```diff
@@ -731,7 +792,14 @@ def predict(self, X: _DaskMatrixLike, **kwargs) -> dask_Array:
...
+        X_leaves_shape="dask Array of shape = [n_samples, n_trees] or shape = [n_samples, n_trees * n_classes]",
```
StrikerRUS, line 800:
```suggestion
        X_leaves_shape="dask Array of shape = [n_samples, n_trees]",
```

python-package/lightgbm/dask.py
```diff
@@ -731,7 +792,14 @@ def predict(self, X: _DaskMatrixLike, **kwargs) -> dask_Array:
...
+        X_SHAP_values_shape="dask Array of shape = [n_samples, n_features + 1] or shape = [n_samples, (n_features + 1) * n_classes]"
```
StrikerRUS, line 801:
```suggestion
        X_SHAP_values_shape="dask Array of shape = [n_samples, n_features + 1]"
```

python-package/lightgbm/dask.py
```diff
@@ -832,13 +904,39 @@ def fit(
...
     def predict(self, X: _DaskMatrixLike, **kwargs: Any) -> dask_Array:
         """Docstring is inherited from the lightgbm.LGBMRanker.predict."""
         return _predict(self.to_local(), X, **kwargs)
 
-    predict.__doc__ = LGBMRanker.predict.__doc__
+    predict.__doc__ = _lgbmmodel_doc_predict.format(
+        description="Return the predicted value for each sample.",
+        X_shape="dask Array or dask DataFrame of shape = [n_samples, n_features]",
+        output_name="predicted_result",
+        predicted_result_shape="dask Array of shape = [n_samples] or shape = [n_samples, n_classes]",
```
StrikerRUS, line 936:
```suggestion
        predicted_result_shape="dask Array of shape = [n_samples]",
```

python-package/lightgbm/dask.py
```diff
@@ -832,13 +904,39 @@ def fit(
...
+        X_leaves_shape="dask Array of shape = [n_samples, n_trees] or shape = [n_samples, n_trees * n_classes]",
```
StrikerRUS, line 937:
```suggestion
        X_leaves_shape="dask Array of shape = [n_samples, n_trees]",
```

python-package/lightgbm/dask.py
```diff
@@ -832,13 +904,39 @@ def fit(
...
+        X_SHAP_values_shape="dask Array of shape = [n_samples, n_features + 1] or shape = [n_samples, (n_features + 1) * n_classes]"
```
StrikerRUS, line 938:
```suggestion
        X_SHAP_values_shape="dask Array of shape = [n_samples, n_features + 1]"
```

python-package/lightgbm/sklearn.py
```diff
@@ -176,6 +176,170 @@ def __call__(self, preds, dataset):
...
+        If 'auto' and data is pandas DataFrame, data columns names are used.
+    categorical_feature : list of strings or int, or 'auto', optional (default='auto')
+        Categorical features.
+        If list of int, interpreted as indices.
+        If list of strings, interpreted as feature names (need to specify ``feature_name`` as well).
+        If 'auto' and data is pandas DataFrame, pandas unordered categorical columns are used.
+        All values in categorical features should be less than int32 max value (2147483647).
+        Large values could be memory consuming. Consider using consecutive integers starting from zero.
+        All negative values in categorical features will be treated as missing values.
+        The output cannot be monotonically constrained with respect to a categorical feature.
```
StrikerRUS, line 250: I believe default values for these three params (`verbose`, `feature_name`, `categorical_feature`) are not true for Dask. Refer to https://github.com/microsoft/LightGBM/pull/3908#discussion_r571343767
> As far as I can tell, "category" columns will only be automatically treated as categorical features if you set categorical_feature = 'auto', which is not the default, and of course you can force it to happen by setting categorical_featue to the column names / indices in the parameters.

StrikerRUS (changes requested):
@jameslamb I love how you introduced shared templates for docstrings! However, I think there are some mistakes, that I noted below:

python-package/lightgbm/sklearn.py
jameslamb: There isn't any special handling in the Dask interface for categorical features.

It's passed through directly to sklearn. So whatever is true for sklearn should be true for Dask. That comment you linked is about the behavior of the sklearn estimators.

Since `verbose`, `feature_name`, and `categorical_feature` aren't in the signatures of the Dask estimators, but just passed to the `sklearn` estimators through `**kwargs`, their defaults are identical to those from sklearn

python-package/lightgbm/dask.py
jameslamb: Oh thank you, my mistake. I thought I remembered that there was an inconsistency in the case of binary classification, where `predict_proba()` only returned an array of shape `[n_samples]`.

I was thinking about https://github.com/microsoft/LightGBM/issues/3636, but I guess I remembered it incorrectly.

python-package/lightgbm/sklearn.py
StrikerRUS: OK, got it!

Will `'auto'` work for dask DataFrame?

python-package/lightgbm/dask.py
```diff
@@ -604,7 +615,23 @@ def fit(
             **kwargs
         )
 
-    fit.__doc__ = LGBMClassifier.fit.__doc__
+    _base_doc = _lgbmmodel_doc_fit.format(
+        X_shape="dask Array or dask DataFrame of shape = [n_samples, n_features]",
+        y_shape="dask Array, dask DataFrame or dask Series of shape = [n_samples]",
+        sample_weight_shape="dask Array, dask DataFrame, Dask Series of shape = [n_samples] or None, optional (default=None)",
+        group_shape="dask Array, dask DataFrame, Dask Series of shape = [n_samples] or None, optional (default=None)"
```
StrikerRUS, line 622:
```suggestion
        sample_weight_shape="dask Array, dask DataFrame, dask Series of shape = [n_samples] or None, optional (default=None)",
        group_shape="dask Array, dask DataFrame, dask Series of shape = [n_samples] or None, optional (default=None)"
```

python-package/lightgbm/dask.py
```diff
@@ -721,7 +766,23 @@ def fit(
             **kwargs
         )
 
-    fit.__doc__ = LGBMRegressor.fit.__doc__
+    _base_doc = _lgbmmodel_doc_fit.format(
+        X_shape="dask Array or dask DataFrame of shape = [n_samples, n_features]",
+        y_shape="dask Array, dask DataFrame or dask Series of shape = [n_samples]",
+        sample_weight_shape="dask Array, dask DataFrame, Dask Series of shape = [n_samples] or None, optional (default=None)",
+        group_shape="dask Array, dask DataFrame, Dask Series of shape = [n_samples] or None, optional (default=None)"
```
StrikerRUS, line 773:
```suggestion
        sample_weight_shape="dask Array, dask DataFrame, dask Series of shape = [n_samples] or None, optional (default=None)",
        group_shape="dask Array, dask DataFrame, dask Series of shape = [n_samples] or None, optional (default=None)"
```

python-package/lightgbm/dask.py
```diff
@@ -832,13 +904,39 @@ def fit(
             **kwargs
         )
 
-    fit.__doc__ = LGBMRanker.fit.__doc__
+    _base_doc = _lgbmmodel_doc_fit.format(
+        X_shape="dask Array or dask DataFrame of shape = [n_samples, n_features]",
+        y_shape="dask Array, dask DataFrame or dask Series of shape = [n_samples]",
+        sample_weight_shape="dask Array, dask DataFrame, Dask Series of shape = [n_samples] or None, optional (default=None)",
+        group_shape="dask Array, dask DataFrame, Dask Series of shape = [n_samples] or None, optional (default=None)"
```
StrikerRUS, line 911:
```suggestion
        sample_weight_shape="dask Array, dask DataFrame, dask Series of shape = [n_samples] or None, optional (default=None)",
        group_shape="dask Array, dask DataFrame, dask Series of shape = [n_samples] or None, optional (default=None)"
```

StrikerRUS (commented):
For the consistency:

python-package/lightgbm/dask.py
jameslamb: @StrikerRUS I just checked the Dask docs for guidance, and I think based on https://marketing.dask.org/en/latest/refer.html that we should actually prefer capitalized `Dask` when talking about the project, and lowercase `dask` when referring to the specific Python package `dask`. So for example:

* `Dask DataFrame`
* `dask.distributed client`

I'll make that change everywhere in this PR, including in lines not currently touched by it. I think it makes sense in the scope of this PR.

h/t to @rikturr for pointing this out to me (https://github.com/dask/dask.github.io/pull/44)

StrikerRUS: OK, thank you!
We don't have a lot of docs about Dask yet, so I think it is OK to do it right here in this PR.

python-package/lightgbm/sklearn.py
jameslamb: I expect that it will, since the chunks of a Dask DataFrame are each pandas dataframes, and since we saw debugging #3861 that Dask DataFrame preserves the "category" type of the underlying dataframes.

However, I haven't actually tested that yet. I think it could be tested by removing `categorical_feature` from `params` in the Dask tests, like https://github.com/microsoft/LightGBM/blob/846b512d892426d34fac0749ccfd2d7a31eef877/tests/python_package_test/test_dask.py#L464 and instead setting it to `"categorical_feature": "auto"`. I can try that in a bit.

jameslamb: interesting! When I tried setting `categorical_feature = "auto"` in params, I got this error:

> E           lightgbm.basic.LightGBMError: categorical_feature is not a number,
E           if you want to use a column name,
E           please add the prefix "name:" to the column name

But that error is coming from `lgb.LGBMClassifier.fit()`. `lgb.DaskLGBMClassifier.fit()` is ok with it

image

When I switch to passing `categorical_feature="auto"` in `.fit()` (instead of in the constructor), all the tests pass!

So:

1. yes, `auto` seems to work for the Dask interface so this doc is accurate
2. I should probably change the tests (in a separate PR) to pass `categorical_feature` to `.fit()`, not in the model constructor. I can see that that's what the sklearn tests do: https://github.com/microsoft/LightGBM/blob/846b512d892426d34fac0749ccfd2d7a31eef877/tests/python_package_test/test_sklearn.py#L497-L508. Looking through the `sklearn` module, I see that `categorical_feature` is used in `fit()` everywhere.

python-package/lightgbm/dask.py
jameslamb: updated this everywhere I could find in https://github.com/microsoft/LightGBM/pull/3930/commits/8cceecf2f0b184f22559da58c8c12891e8bcd6e3

I used `git grep -i dask` from the root of the repo to check for other cases and don't see any that need their capitalizations changed

python-package/lightgbm/sklearn.py
StrikerRUS: > and instead setting it to "categorical_feature": "auto". I can try that in a bit.

> I should probably change the tests (in a separate PR) to pass categorical_feature to .fit(), not in the model constructor

But `'auto'` is default value, so you can just omit `categorical_feature argument in tests at all. And yes, let's leave this for a separate PR please.

tests/python_package_test/test_dask.py
```diff
@@ -575,7 +575,7 @@ def test_ranker(output, client, listen_port, group):
             group=group,
         )
 
-    # rebalance small dask.array dataset for better performance.
+    # rebalance small Dask.array dataset for better performance.
```
StrikerRUS, line 578:
```suggestion
    # rebalance small dask.Array dataset for better performance.
```

python-package/lightgbm/dask.py
StrikerRUS: Great! Many thanks!

StrikerRUS (approved):
LGTM! Thank you!

python-package/lightgbm/sklearn.py
jameslamb: good point! added https://github.com/microsoft/LightGBM/pull/3932

jameslamb:
the most recent build on RTD did pass: https://readthedocs.org/projects/lightgbm/builds/12979086/

And docs look ok to me! https://lightgbm.readthedocs.io/en/docs-jlamb/Python-API.html

For example, here's `DaskLGBMRanker.fit()`

image

And the sklearn docs look unchanged.

Issue #1287, opened 2018-03-28
Windows GPU CMAKE Generator Error x64


Whamp:
I'm new to cmake so this may just be user error but when installing for gpu on windows according to the instructions here:

 [https://github.com/Microsoft/LightGBM/blob/master/docs/Installation-Guide.rst#build-gpu-version](url)

I get the following error message when running:

`cmake -DCMAKE_GENERATOR_PLATFORM=x64 -DUSE_GPU=1 ..`

## Environment info
Operating System: Windows 10 Pro
CPU: i7-5820K 
GPU: Nvidia 1080 Ti
C++ Microsoft Visual C++ 2017 Redistributable (x64) - 14.13.26020
Python 3.6.4 Anaconda
CMAKE version 3.10.3
Visual Studio 2017

## Error Message:
```
(base) C:\Users\Will\LightGBM\build>cmake -DCMAKE_GENERATOR_PLATFORM=x64 -DUSE_GPU=1 ..
CMake Error at CMakeLists.txt:9 (PROJECT):
  Generator

    NMake Makefiles

  does not support platform specification, but platform

    x64

  was specified.


CMake Error: CMAKE_C_COMPILER not set, after EnableLanguage
CMake Error: CMAKE_CXX_COMPILER not set, after EnableLanguage
-- Configuring incomplete, errors occurred!
See also "C:/Users/Will/LightGBM/build/CMakeFiles/CMakeOutput.log".
```

## Attempted Resolution

I have tried changing the generator to `Visual Studio 15 2017 Win64` but get the same error message with 'Visual Studio 15 2017 Win64' in place of  'NMake Makefiles'

Whamp:
No idea what solved it but trying the install fresh from the top worked

PR #3931, opened 2021-02-09
Improve GPU build instructions

nhirschey:
Until I added the -G flag I kept getting errors like

```
CMake Error at CMakeLists.txt:30 (PROJECT):
  Generator

    NMake Makefiles

  does not support platform specification, but platform

    x64

  was specified.
```

The solution came form the [xgboost gpu build instructions](https://xgboost.readthedocs.io/en/latest/build.html)

### DIFF
--- a/docs/Installation-Guide.rst
+++ b/docs/Installation-Guide.rst
@@ -566,9 +566,12 @@ Following procedure is for the **MSVC** (Microsoft Visual C++) build.
      cd LightGBM
      mkdir build
      cd build
-     cmake -A x64 -DUSE_GPU=1 -DBOOST_ROOT=C:/local/boost_1_63_0 -DBOOST_LIBRARYDIR=C:/local/boost_1_63_0/lib64-msvc-14.0 ..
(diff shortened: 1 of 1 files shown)


StrikerRUS (changes requested):
@nhirschey Thanks for your contribution!

However, normally, CMake is able to build the project without explicitly specified "Visual Studio" generator on Windows. I guess you have some modified environment, so you should tell CMake what to choose via `-G` flag: maybe multiple VS versions are installed on your machine or you have confusing/complicated PATH environment variable.
See the following line which allows us to build LightGBM on numerous CI services (Appveyor, GitHub Actions, Azure Pipelines) just fine for different CMake/VS combinations.
https://github.com/microsoft/LightGBM/blob/846b512d892426d34fac0749ccfd2d7a31eef877/.ci/test_windows.ps1#L48

These docs are already very detailed and have a lot of info. So I'd prefer to keep instructions as general and simple as possible. And I'm against including additional commands that are relevant only for some uncommon (read "non-default") environments as required ones for everybody.

nhirschey:
If it's not useful, go ahead and reject the request. I agree that simplicity is good and the build process instructions are quite complex.

I saw other unresolved build issues in the repo with the same error (https://github.com/microsoft/LightGBM/issues/647, https://github.com/microsoft/LightGBM/issues/1287) and thought this might help.

I do have multiple VS installs (regular and preview). Perhaps that was the issue. I seem to now not need the -G flag, so maybe it is only needed the first time (if you have multiple installs).

PR #3932, opened 2021-02-09
[dask] test that Dask automatically treats 'category' columns as categorical features

jameslamb:
Based on thread starting at https://github.com/microsoft/LightGBM/pull/3930#discussion_r573107195.

#3908 added some tests that the `lightgbm.dask` estimators correctly handled Dask DataFrames where some columns are `pandas` "category" columns. That PR used the parameter `categorical_feature` to explicitly provide a list of which features should be treated as categorical.

This is unnecessary, since `auto` is the default behavior and `auto` tells LightGBM "treat 'category' columns as categorical features" (https://github.com/microsoft/LightGBM/blob/846b512d892426d34fac0749ccfd2d7a31eef877/python-package/lightgbm/basic.py#L513-L514).

This PR removes the unnecessary `categorical_features` parameter in tests. As of this PR, we'll now be testing not only that `lightgbm.dask` estimators correctly train on categorical features, but also that it also automatically

### DIFF
--- a/tests/python_package_test/test_dask.py
+++ b/tests/python_package_test/test_dask.py
@@ -210,11 +210,6 @@ def test_classifier(output, centers, client, listen_port):
         "num_leaves": 10
     }
 
-    if output == 'dataframe-with-categorical':
-        params["categorical_feature"] = [
-            i for i, col in enumerate(dX.columns) if col.startswith('cat_')
-        ]
-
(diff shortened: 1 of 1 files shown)


StrikerRUS (approved):
Nice!

jameslamb:
> Nice!

thanks for pointing it out!

PR #3933, opened 2021-02-09
[docs][python] fix shape description of returned result for predict_proba

StrikerRUS:
I was wrong in my suggestion previously https://github.com/microsoft/LightGBM/pull/3930#discussion_r573036384.
Shape can be `[n_samples]` in case of binary problem and `raw_score=True` or custom objective function.


jameslamb (approved):
ah thanks for catching that!

PR #3936, opened 2021-02-10
Fix access violation exception that can occur during invocation of loop lambda function when inner_start >= inner_end in 'For' template.

mjmckp:
In particular, this can occur in `Tree::AddPredictionToScore` on line 291 of src\io\tree.cpp where the loop lambda function body (created by the `PredictionFun` macro) dereferences `used_data_indices[start]`.

For reference, the particular case which triggered this exception in my case was:
  * `start` = 0
  * `end` = 93,203
  * `min_block_size` = 512
for which the `BlockInfo` method gave:
  * `n_block` = 56
  * `num_inner` = 1,696
and so, for the case `i=55` (i.e., the last case in the loop), we get
  * `inner_start = start + num_inner * i` = 93,280
which is greater than 'end' and hence triggers the exception.


guolinke:
cc @shiyu1994 
I think there are multiple places could have this problem, could you check them carefully?

guolinke (approved):
Thank you!

include/LightGBM/utils/threading.h
```diff
@@ -78,7 +78,7 @@ class Threading {
       OMP_LOOP_EX_BEGIN();
       INDEX_T inner_start = start + num_inner * i;
       INDEX_T inner_end = std::min(end, inner_start + num_inner);
-      inner_fun(i, inner_start, inner_end);
+      if (inner_start < inner_end) inner_fun(i, inner_start, inner_end);
```
StrikerRUS, line 81: @mjmckp 
Could you please make code more readable?
```suggestion
      if (inner_start < inner_end) {
        inner_fun(i, inner_start, inner_end);
      }
```

shiyu1994:
I've checked all the `#pragma omp` directives. I think this is the only place with such problem. In other cases even when `inner_start` is no less than `inner_end`, there'll be no problem.

Issue #3941, opened 2021-02-12
inf/nan values in model are silently converted to zero when model is round-tripped to file


mjmckp:
When compiled with VS2017, a round trip of a LightGBM model to string (or file) via the `GBDT::SaveModelToString` and `GBDT::LoadModelFromString` functions silently replaces all inf/nan values in the model with zeros.

This happens because the `GBDT::LoadModelFromString` calls `CommonC::StringToArray<double>` (see below, from line 1105 of utils\common.h) to convert the strings containing various parts of the model (such as the thresholds etc):
```
template<typename T>
struct __StringToTHelper<T, true> {
  T operator()(const std::string& str) const {
    double tmp;

    // Fast (common) path: For numeric inputs in RFC 7159 format:
    const bool fast_parse_succeeded = fast_double_parser::parse_number(str.c_str(), &tmp);

    // Rare path: Not in RFC 7159 format. Possible "inf", "nan", etc. Fallback to standard library:
    if (!fast_parse_succeeded) {
      std::stringstream ss;
      Common::C_stringstream(ss);
      ss << str;
      ss >> tmp;
    }

    return static_cast<T>(tmp);
  }
};
```

When the input to this method is "inf", "-inf" or "nan", the second half of the method is meant to handle these cases, but in fact it actually silently sets the return value `tmp` to zero.

This bug appears to have been introduced in 792c930305a818db1463911c2cbee92d462eaab9 in Dec 2020.

## Reproduction
1) Calibrate a LightGBM model which contains a non-finite value (e.g., inf) in one of the trees
2) Save the model to file
3) Load the model from file
4) Inspect the loaded model and observe all the non-finite values have been replaced by zero

## Environment info
Windows 10, Visual Studio 2017.

StrikerRUS:
Fixed via #3942.

PR #3942, opened 2021-02-12
Fix parsing of non-finite values

mjmckp:
Proposed solution for #3941.  Current implementation silently returns zero when input string is "inf", "-inf", or "nan" when compiled with VS2017, so instead just explicitly check for these values and fail if there is no match. No attempt to optimise string allocations in this implementation since it is usually rarely invoked.


StrikerRUS:
@AlbertoEAF Could you please help to review?

AlbertoEAF (changes requested):
Great catch!

1. C++98 does not define aspect of inf/nan values in text, but C99 does, and so, unlike `stod`, apparently stringstreams might not work in all platforms (specially MSVC) to parse `inf`, and `nan` values: https://www.boost.org/doc/libs/1_70_0/libs/math/doc/html/math_toolkit/fp_facets/facets_intro.html

2. The old implementation used `stod` (which is not locale-safe for our purposes): https://en.cppreference.com/w/cpp/string/basic_string/stof. See the old version of of this very function here: https://github.com/microsoft/LightGBM/commit/792c930305a818db1463911c2cbee92d462eaab9#diff-ddbdb79f73cc159561b86b759b0542fec75ca9f4869a7eac5790e8f9ebaf7ef4R382

3. We shouldn't stop parsing numbers when `fast_double_parser::parse_number` fails. There might be some numerical values that besides inf, or nan which are not parsed either. Hence we should keep the stringstream parsing for numerical values on the slow path instead of giving a fatal error.

include/LightGBM/utils/common.h
```diff
@@ -1082,12 +1082,18 @@ struct __StringToTHelper<T, true> {
     // Fast (common) path: For numeric inputs in RFC 7159 format:
     const bool fast_parse_succeeded = fast_double_parser::parse_number(str.c_str(), &tmp);
 
-    // Rare path: Not in RFC 7159 format. Possible "inf", "nan", etc. Fallback to standard library:
+    // Rare path: Not in RFC 7159 format. Possible "inf", "nan", etc.
     if (!fast_parse_succeeded) {
-      std::stringstream ss;
-      Common::C_stringstream(ss);
-      ss << str;
-      ss >> tmp;
```
AlbertoEAF, line 1090: Keep this code on the `else` branch instead of raising a fatal error.

AlbertoEAF (changes requested):

include/LightGBM/utils/common.h
```diff
@@ -1082,12 +1082,18 @@ struct __StringToTHelper<T, true> {
...
+      std::string strlower(str); 
+      std::transform(strlower.begin(), strlower.end(), strlower.begin(), [](int c) -> char { return static_cast<char>(::tolower(c)); });
+      if (strlower == std::string("inf"))
+        tmp = std::numeric_limits<double>::infinity();
+      else if (strlower == std::string("-inf"))
+        tmp = -std::numeric_limits<double>::infinity();
+      else if (strlower == std::string("nan"))
```
AlbertoEAF, line 1093: Important: Missing -nan handling.

Probably best to halve the string comparisons by parsing first the "-" sign.

include/LightGBM/utils/common.h
```diff
@@ -1082,12 +1082,18 @@ struct __StringToTHelper<T, true> {
...
 
-    // Rare path: Not in RFC 7159 format. Possible "inf", "nan", etc. Fallback to standard library:
+    // Rare path: Not in RFC 7159 format. Possible "inf", "nan", etc.
     if (!fast_parse_succeeded) {
-      std::stringstream ss;
-      Common::C_stringstream(ss);
-      ss << str;
-      ss >> tmp;
+      std::string strlower(str); 
+      std::transform(strlower.begin(), strlower.end(), strlower.begin(), [](int c) -> char { return static_cast<char>(::tolower(c)); });
```
AlbertoEAF, line 1088: Great clean code @mjmckp ;)

Instead of allocating a string, and since you already have a lambda, what about defining a lower-case comparison lambda and use `std::equal` to check the "inf" and "nan" values below?

Although this is the rare branch there might be longer strings than inf or nan which might be parsed here and might slow down our parsing without need.

AlbertoEAF: Example: https://thispointer.com/c-case-insensitive-string-comparison-using-stl-c11-boost-library/

include/LightGBM/utils/common.h
mjmckp: But that code silently fails, which is completely unacceptable.

In reality we are only parsing strings generated by LightGBM itself when the model was written out to file, so we should ensure there are robust round-trip tests which include models that contain inf and nan values. There are no other possible non-finite values defined for IEEE 754 floating point numbers, and if at some point in the future the standard changed, this would be quickly picked up by the round-trip tests and easily addressed.

I am shocked that such tests don't already exist.

include/LightGBM/utils/common.h
mjmckp: Ok thanks, I'll add -nan.  I don't think it's worth obfuscating this rarely executed branch with optimisations until profiling shows it is a bottleneck.

include/LightGBM/utils/common.h
mjmckp: Again, this hardly seems worth it, this branch is rarely invoked, meanwhile a colossal amount of strings are being allocated in splitting and parsing the input file, so these few extra allocations are a drop in the ocean.

I think our time is better spent adding robust round-trip tests to ensure major bugs like this don't occur again...

AlbertoEAF: > I think our time is better spent adding robust round-trip tests to ensure major bugs like this don't occur again...

Agreed. Will you add such tests?
I actually run such tests but on an external lgbm provider and didn't have nan nor inf on my model, but would prefer to see them in lgbm's CI so any breakage is detected immediately.

include/LightGBM/utils/common.h
AlbertoEAF: I meant doing number parsing after the nan and inf checks. Your logic makes sense regarding IEEE-754, although I wouldn't recommend dropping that parsing at the end without adding said tests first, not 100.0% sure fast_double_parser parses all numbers.

include/LightGBM/utils/common.h
```diff
@@ -1082,12 +1082,20 @@ struct __StringToTHelper<T, true> {
...
     const bool fast_parse_succeeded = fast_double_parser::parse_number(str.c_str(), &tmp);
 
-    // Rare path: Not in RFC 7159 format. Possible "inf", "nan", etc. Fallback to standard library:
+    // Rare path: Not in RFC 7159 format. Possible "inf", "nan", etc.
     if (!fast_parse_succeeded) {
-      std::stringstream ss;
-      Common::C_stringstream(ss);
-      ss << str;
-      ss >> tmp;
+      std::string strlower(str);
```
StrikerRUS, line 1087: To pass linting tests
```
include/LightGBM/utils/common.h:1087:  Line ends in whitespace.  Consider deleting these extra spaces.  [whitespace/end_of_line] [4]
```
```suggestion
      std::string strlower(str);
```

include/LightGBM/utils/common.h
mjmckp: I could assist in adding the tests, any idea where these should go and how best to implement them?

StrikerRUS: @mjmckp Thank you! For tests we need to merge #3555 first, I guess.

StrikerRUS:
@shiyu1994 @btrotta @guolinke Could you please review this to include in the upcoming release?

include/LightGBM/utils/common.h
StrikerRUS: @mjmckp #3555 has been just merged. We will really appreciate new GTest-compatible tests in this or in a follow-up PR.

mjmckp: Sure, would you mind pointing me towards a similar kind of test that I can use as a starting point please?

StrikerRUS: Unfortunately, we don't have any tests yet. This is something that we should concentrate on in the near future. For now, I think you can take a look at tests from @AlbertoEAF in #3997. 
https://github.com/microsoft/LightGBM/pull/3997/files#diff-c363eba6eda99d9e560f8341a1fc8fe02e885d2256db2482e1c543430a25666d

guolinke (approved):
LGTM

include/LightGBM/utils/common.h
StrikerRUS: @mjmckp I've merged this PR with the aim to not delay the upcoming release. Please feel free to add tests in a new PR. We'll be very grateful! And thanks a lot for the bug fix!

mjmckp: @StrikerRUS Thanks a lot, I'll get up to speed on how the new tests work and add some tests for this in a new PR soon.

Issue #3909, opened 2021-02-04
[dask] Cannot use a mix of Array and DataFrame types in training


jameslamb:
## Summary

Using the estimators in `lightgbm.dask`, it is not possible to train on a mix of Dask Array and Dask DataFrame. I found, for example, that if `X` is a Dask DataFrame and `y` is a Dask Array, training fails with an error like this

> TypeError: Wrong type(ndarray) for label.
It should be list, numpy 1-D array or pandas Series

<details><summary>full logs</summary>

```text
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-28-e17ddcb3ecf3> in <module>
----> 1 reg.fit(X=ddf, y=label)

~/miniconda3/lib/python3.7/site-packages/lightgbm/dask.py in fit(self, X, y, sample_weight, client, **kwargs)
    720             y=y,
    721             sample_weight=sample_weight,
--> 722             **kwargs
    723         )
    724 

~/miniconda3/lib/python3.7/site-packages/lightgbm/dask.py in _fit(self, model_factory, X, y, sample_weight, group, **kwargs)
    498             sample_weight=sample_weight,
    499             group=group,
--> 500             **kwargs
    501         )
    502 

~/miniconda3/lib/python3.7/site-packages/lightgbm/dask.py in _train(client, data, label, params, model_factory, sample_weight, group, **kwargs)
    337     ]
    338 
--> 339     results = client.gather(futures_classifiers)
    340     results = [v for v in results if v]
    341     return results[0]

~/miniconda3/lib/python3.7/site-packages/distributed/client.py in gather(self, futures, errors, direct, asynchronous)
   1990                 direct=direct,
   1991                 local_worker=local_worker,
-> 1992                 asynchronous=asynchronous,
   1993             )
   1994 

~/miniconda3/lib/python3.7/site-packages/distributed/client.py in sync(self, func, asynchronous, callback_timeout, *args, **kwargs)
    831         else:
    832             return sync(
--> 833                 self.loop, func, *args, callback_timeout=callback_timeout, **kwargs
    834             )
    835 

~/miniconda3/lib/python3.7/site-packages/distributed/utils.py in sync(loop, func, callback_timeout, *args, **kwargs)
    338     if error[0]:
    339         typ, exc, tb = error[0]
--> 340         raise exc.with_traceback(tb)
    341     else:
    342         return result[0]

[...]

PR #3943, opened 2021-02-12
[dask] [python-package] include support for column array as label

jmoralez:
This solves #3909 by allowing to have an `(n_samples, 1)` array as label when `data` is a `pandas.DataFrame`.

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -134,8 +134,8 @@ def is_numpy_column_array(data):
     return len(shape) == 2 and shape[1] == 1
 
 
-def numpy_1d_array_to_dtype(array, dtype):
-    """Convert 1d array to dtype."""
+def cast_numpy_1d_array_to_dtype(array, dtype):
+    """Cast numpy 1d array to given dtype."""
     if array.dtype == dtype:
         return array
     return array.astype(dtype=dtype, copy=False)
@@ -149,11 +149,11 @@ def is_1d_list(data):
 def list_to_1d_numpy(data, dtype=np.float32, name='list'):
     """Convert data to numpy 1-D array."""
     if is_numpy_1d_array(data):
-        return numpy_1d_array_to_dtype(data, dtype)
+        return cast_numpy_1d_array_to_dtype(data, dtype)
     elif is_numpy_column_array(data):
         _log_warning('Converting column-vector to 1d array')
         array = data.ravel()
-        return numpy_1d_array_to_dtype(array, dtype)
+        return cast_numpy_1d_array_to_dtype(array, dtype)
     elif is_1d_list(data):
         return np.array(data, dtype=dtype, copy=False)
     elif isinstance(data, pd_Series):
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -135,7 +135,7 @@ def is_numpy_column_array(data):
 
 
 def numpy_1d_array_to_dtype(array, dtype):
-    """Convert 1d array to dtype"""
+    """Convert 1d array to dtype."""
     if array.dtype == dtype:
         return array
     return array.astype(dtype=dtype, copy=False)
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -151,7 +151,7 @@ def list_to_1d_numpy(data, dtype=np.float32, name='list'):
     if is_numpy_1d_array(data):
         return numpy_1d_array_to_dtype(data, dtype)
     elif is_numpy_column_array(data):
-        _log_warning('Converting column vector to 1d array')
+        _log_warning('Converting column-vector to 1d array')
         array = data.ravel()
         return numpy_1d_array_to_dtype(array, dtype)
     elif is_1d_list(data):
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -126,6 +126,20 @@ def is_numpy_1d_array(data):
     return isinstance(data, np.ndarray) and len(data.shape) == 1
 
 
+def is_numpy_column_array(data):
+    """Check whether data is a column numpy array."""
+    if not isinstance(data, np.ndarray):
+        return False
+    shape = data.shape
+    return len(shape) == 2 and shape[1] == 1
+
+
+def numpy_1d_array_to_dtype(array, dtype):
+    if array.dtype == dtype:
+        return array
+    return array.astype(dtype=dtype, copy=False)
+
+
 def is_1d_list(data):
     """Check whether data is a 1-D list."""
     return isinstance(data, list) and (not data or is_numeric(data[0]))
@@ -134,10 +148,11 @@ def is_1d_list(data):
 def list_to_1d_numpy(data, dtype=np.float32, name='list'):
     """Convert data to numpy 1-D array."""
     if is_numpy_1d_array(data):
-        if data.dtype == dtype:
-            return data
-        else:
-            return data.astype(dtype=dtype, copy=False)
+        return numpy_1d_array_to_dtype(data, dtype)
+    elif is_numpy_column_array(data):
+        _log_warning('Converting column vector to 1d array')
+        array = data.ravel()
+        return numpy_1d_array_to_dtype(array, dtype)
     elif is_1d_list(data):
         return np.array(data, dtype=dtype, copy=False)
     elif isinstance(data, pd_Series):
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -135,6 +135,7 @@ def is_numpy_column_array(data):
 
 
 def numpy_1d_array_to_dtype(array, dtype):
+    """Convert 1d array to dtype"""
     if array.dtype == dtype:
         return array
     return array.astype(dtype=dtype, copy=False)
--- a/tests/python_package_test/test_basic.py
+++ b/tests/python_package_test/test_basic.py
@@ -379,7 +379,7 @@ def test_choose_param_value():
 
 
 @pytest.mark.parametrize(
-    'y', 
+    'y',
     [
         np.random.rand(10),
         np.random.rand(10, 1),
--- a/tests/python_package_test/test_basic.py
+++ b/tests/python_package_test/test_basic.py
@@ -8,6 +8,7 @@ from sklearn.datasets import dump_svmlight_file, load_svmlight_file
 from sklearn.model_selection import train_test_split
 
 import lightgbm as lgb
+from lightgbm.compat import PANDAS_INSTALLED, pd_Series
 
 from .utils import load_breast_cancer
 
@@ -377,11 +378,14 @@ def test_choose_param_value():
     assert original_params == expected_params
 
 
+@pytest.mark.skipif(not PANDAS_INSTALLED, reason='pandas is not installed')
 @pytest.mark.parametrize(
     'y',
     [
         np.random.rand(10),
         np.random.rand(10, 1),
+        pd_Series(np.random.rand(10)),
+        pd_Series(['a', 'b']),
         [1] * 10,
         [[1], [2]]
     ])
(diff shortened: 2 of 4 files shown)


jmoralez:
Hi @jameslamb. Can you give me your thoughts on these changes?

jameslamb:
Thanks so much for looking into this!

>  Can you give me your thoughts on these changes?

Can you please add tests (one in `test_sklearn.py`, one in `test_dask.py`) testing that training is successful and correct when `data` is a `pandas` DataFrame and `label` is a numpy column array? Once you've written tests like that proving that this PR fixes the issue described in #3909, I'd be happy to give a more thorough review on the approach.

tests/python_package_test/test_basic.py
```diff
@@ -3,6 +3,7 @@
 
 import lightgbm as lgb
 import numpy as np
+import pandas as pd
```
StrikerRUS, line 6: @jmoralez Please make `pandas` optional for basic tests like you did in sklearn tests:
```
pd = pytest.importorskip("pandas")
```

tests/python_package_test/test_dask.py
```diff
@@ -1108,3 +1108,26 @@ def test_dask_methods_and_sklearn_equivalents_have_similar_signatures(methods):
     for param in dask_spec.args:
         error_msg = f"param '{param}' has different default values in the methods"
         assert dask_params[param].default == sklearn_params[param].default, error_msg
+
+
+def test_training_succeeds_when_data_is_dataframe_and_label_is_column_array(
+    client,
+    listen_port
+):
```
jameslamb, line 1116: thanks for this! I think test could be a bit stronger. Instead of just smoke-testing that training runs without error, could you make this test check that `label` vectors of shape `[n_samples, 1]` and `[n_samples]` result in identical models?

I think that could look something like this

```python
    params = {
        'n_estimators': 1,
        'num_leaves': 3,
        'local_listen_port': listen_port,
        'time_out': 5
    }
    _, _, _, dX, dy, dw = _create_data(
        objective='regression',
        output='dataframe'
    )
    dy = dy.to_dask_array(lengths=True)
    dy_col_array = dy.reshape(-1, 1)
    assert len(dy_col_array.shape) == 2 and dy_col_array.shape[1] == 1

    model1 = lgb.DaskLGBMRegressor(**params)
    model1.fit(dX, dy, sample_weight=dw)
    p1 = model1.predict(dX).compute()
    assert model1.fitted_
    
    dy_vector = dy.reshape(-1)
    assert len(dy_col_array.shape) == 1
    model2 = lgb.DaskLGBMRegressor(**params)
    model2.fit(dX, dy, sample_weight=dw)
    p2 = model2.predict(dX).compute()
    assert model2.fitted_
    
    assert_eq(p1, p2)
```

Notice in the above that I also changed the check from `assert model.booster_` to `assert model.fitted_`. I think that's a more direct check of the thing we're trying to check ("is the model fitted").

tests/python_package_test/test_dask.py
```diff
@@ -1108,3 +1108,26 @@ def test_dask_methods_and_sklearn_equivalents_have_similar_signatures(methods):
     for param in dask_spec.args:
         error_msg = f"param '{param}' has different default values in the methods"
         assert dask_params[param].default == sklearn_params[param].default, error_msg
+
+
+def test_training_succeeds_when_data_is_dataframe_and_label_is_column_array(
```
jameslamb, line 1113: can you please run this test for all three of `DaskLGBMClassifier`, `DaskLGBMRanker`, and `DaskLGBMRegressor`? Similar to https://github.com/microsoft/LightGBM/blob/84c4b7504a49eed63fa819a92db7d0b18292f3f4/tests/python_package_test/test_dask.py#L966

jameslamb (changes requested):
Thanks! The fix looks good to me. I left some suggestions for making the tests a bit stronger.

tests/python_package_test/test_basic.py
jmoralez: Since I had some `pd.Series` in the parametrize for `y` I changed it to this:
```python
    ('y', 'is_series'),
    [
        (np.random.rand(10), False),
        (np.random.rand(10, 1), False),
        ([1] * 10, False),
        (np.random.rand(10), True),
        (['a', 'b'], True),
        ([[1], [2]], False)
    ])
```
and then I do this inside the function
```python
    if is_series:
        pd = pytest.importorskip('pandas')
        y = pd.Series(y)
```
Is that the best way to do it?

tests/python_package_test/test_dask.py
jmoralez: This only seems to impact the regressor, the ranker and the classifier go through scikit-learns check_X_y

jameslamb: testing all three will prevent us from introducing a change in the future that makes that not true

jmoralez: Yeah that sounds reasonable. I'll include it in test_sklearn as well

jmoralez: Is there a way to make the model deterministic? I'm calling repeatedly `model_1d.fit(dX, dy, sample_weight=dw).predict(dX).compute()` and the results differ from time to time. I've set `random_state` in the parameters.

jameslamb: you can try adding "deterministic: true" to the parameters: https://lightgbm.readthedocs.io/en/latest/Parameters.html#deterministic

jmoralez: I had to switch to single worker training for the test to pass everytime

jameslamb: ok, I'd prefer removing the `assert_eq()` then and going back to multi-worker training, sorry. It can be difficult to guarantee consistent results when doing distributed training on such tiny data sizes. I had hoped it would work but if it doesn't, we can go back to just training one model and checking `.fitted_`. Sorry for sending you down a road to a dead end.

jmoralez: Sorry hadn't seen this. Even though the training is only happening in one worker the idea is testing that both the dask and scikit-learn models can train with a column array and get the same results, isn't it?

jameslamb: For the Dask interface, having data in a single partition is basically just falling back to training with the sklearn estimators.

The test you've added in `test_sklearn.py` that ends in `assert np.allclose(preds_1d, preds_2d)` is already checking that, for non-distributed training with the sklearn interface, the model output is the same when a label has shape `[n_samples, 1]` or `[n_samples]`.

Given that, I'd rather have the Dask test be a smoke test that multi-worker distributed training works when data has a column-array shape.

tests/python_package_test/test_dask.py
```diff
@@ -196,6 +196,15 @@ def _unpickle(filepath, serializer):
         raise ValueError(f'Unrecognized serializer type: {serializer}')
 
 
+def collection_to_single_partition(collection):
```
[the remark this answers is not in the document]
jameslamb: you can ignore this comment. Please just put this method back, based on https://github.com/microsoft/LightGBM/pull/3943/files#r575617861

tests/python_package_test/test_basic.py
StrikerRUS: Maybe it will be easier to just mark the whole test for skipping if `pandas` is not installed?
```
@pytest.mark.skipif(not lgb.compat.PANDAS_INSTALLED, reason='pandas is not installed')
```
like here
https://github.com/microsoft/LightGBM/blob/50e061f396d7c1b704e79371dea0ddf07680ba75/tests/python_package_test/test_plotting.py#L132

Yeah, it will skip the whole test if pandas is not installed but it looks much easier to read and maintain than one more parametrization argument.

jameslamb: I agree with @StrikerRUS , I think that's simpler than having this flag and it's ok. We install `pandas` in CI so the tests won't be skipped there.

jmoralez: But how would I define the series within the parametrize arguments? I can maybe make a second test that runs for series only and skip that one if pandas isn't installed. Or maybe just skip the `test_list_to_1d_numpy` altogether and just test the training with the column array?

StrikerRUS: I believe you could use `pd_Series` class from our `compat.py` module.
https://github.com/microsoft/LightGBM/blob/4ae59494ab9091a8405a61ae7b13bdedbdbed93e/python-package/lightgbm/compat.py#L4-L17

Will this work?
```py
@pytest.mark.skipif(not PANDAS_INSTALLED, reason='pandas is not installed') 
@pytest.mark.parametrize(
    'y',
    [
        (np.random.rand(10)),
        (np.random.rand(10, 1)),
        ([1] * 10),
        (pd_Series(np.random.rand(10))),
        (pd_Series(['a', 'b'])),
        ([[1], [2]])
    ])
@pytest.mark.parametrize('dtype', [np.float32, np.float64])
def test_list_to_1d_numpy(y, dtype):
```

jmoralez: I hadn't seen that, that definitely works, thank you!

jameslamb (approved):
changes look good to me, thanks for picking this up!

jmoralez:
Thank you for your help, James! I just realized I wasn't fitting to the column array in the test for dask so I fixed that.

StrikerRUS:
@jmoralez 
Sorry for the inconvenience. We just merged #3947. Could you please resolve conflicts?

jameslamb:
@jmoralez sorry again, but can you please merge `master`  into this? We just fixed some additional CI issues 😬

python-package/lightgbm/basic.py
```diff
@@ -126,6 +126,21 @@ def is_numpy_1d_array(data):
...
+def is_numpy_column_array(data):
+    """Check whether data is a column numpy array."""
+    if not isinstance(data, np.ndarray):
+        return False
+    shape = data.shape
+    return len(shape) == 2 and shape[1] == 1
+
+
+def numpy_1d_array_to_dtype(array, dtype):
+    """Convert 1d array to dtype."""
```
StrikerRUS, line 138:
```suggestion
def numpy_1d_array_cast_to_dtype(array, dtype):
    """Cast 1d array to given dtype."""
```

python-package/lightgbm/basic.py
```diff
@@ -134,10 +149,11 @@ def is_1d_list(data):
...
     if is_numpy_1d_array(data):
-        if data.dtype == dtype:
-            return data
-        else:
-            return data.astype(dtype=dtype, copy=False)
+        return numpy_1d_array_to_dtype(data, dtype)
+    elif is_numpy_column_array(data):
+        _log_warning('Converting column-vector to 1d array')
+        array = data.ravel()
+        return numpy_1d_array_to_dtype(array, dtype)
```
StrikerRUS, line 156:
```suggestion
        return numpy_1d_array_cast_to_dtype(array, dtype)
```

tests/python_package_test/test_sklearn.py
```diff
@@ -1192,3 +1192,36 @@ def test_parameters_default_constructible(estimator):
...
+        if task == 'ranking':
+            model_1d = model_factory(**params).fit(X, y, group=g)
+            model_2d = model_factory(**params).fit(X, y_col_array, group=g)
+        else:
+            model_1d = model_factory(**params).fit(X, y)
+            model_2d = model_factory(**params).fit(X, y_col_array)
+
+    preds_1d = model_1d.predict(X)
+    preds_2d = model_2d.predict(X)
+    assert np.allclose(preds_1d, preds_2d)
```
StrikerRUS, line 1227: We prefer `numpy.testing` module for asserting arrays. Also, I think it is important to test for exact equal because models should be identical given the difference in only shape of `y`.
```suggestion
    numpy.testing.assert_array_equal(preds_1d, preds_2d)
```

tests/python_package_test/test_dask.py
```diff
@@ -1084,6 +1084,45 @@ def test_dask_methods_and_sklearn_equivalents_have_similar_signatures(methods):
         assert dask_params[param].default == sklearn_params[param].default, error_msg
 
 
+@pytest.mark.parametrize('task', tasks)
+def test_training_succeeds_when_data_is_dataframe_and_label_is_column_array(
+    task,
+    client,
+    listen_port
```
StrikerRUS, line 1091: @jameslamb Should `listen_port` be removed here given that #3994 is merged now?

python-package/lightgbm/basic.py
jmoralez: How about `cast_numpy_1d_array_to_dtype`?

StrikerRUS: Sounds good!

tests/python_package_test/test_dask.py
jameslamb: yes it should

StrikerRUS (approved):
Nice job! Thank you very much!

jmoralez:
Thank you for your comments!

Issue #3807, opened 2021-01-21
[dask] support init_score


jameslamb:
## Summary

LightGBM allows you to provide initial scores to boost from (https://lightgbm.readthedocs.io/en/latest/pythonapi/lightgbm.Dataset.html?highlight=init_score#lightgbm.Dataset.set_init_score).

This option should be supported in the Dask interface. Dask model classes should accept a dask array, dask dataframe, or Dask dataframe series as input for `init_score`.

## Motivation

This change would bring the Dask interface closer to full feature parity with the non-Dask interface, so that users who'd otherwise like to use Dask don't have to avoid it because `init_score` is missing.

## References

Created from this conversation: https://github.com/microsoft/LightGBM/pull/3708#discussion_r562024425

jameslamb:
Closing this in favor of being in #2302 with other feature requests. Anyone is welcome to contribute this feature. Leave a comment below and it can be re-opened.

jmoralez:
I see this is a method of `lgb.Dataset`, should I attempt to make a Dask equivalent? I guess this will have to be done eventually to avoid having lots of parts all over the place.

jameslamb:
Since you're working on this, I'll open the issue back up. Thanks for looking into it!

> I see this is a method of lgb.Dataset, should I attempt to make a Dask equivalent

I'm not sure what you mean by "method of Dataset", sorry. For a training dataset of shape `[n, k]`, `init_score` is an array of shape `[n]` with initial scores for each sample.

https://github.com/microsoft/LightGBM/blob/84c4b7504a49eed63fa819a92db7d0b18292f3f4/python-package/lightgbm/sklearn.py#L194

Since the Dask interface only supports LightGBM's scikit-learn interface today, you don't need to do anything with the `Dataset` object to add this feature. I think it'll work to handle `init_score` exactly the way we handle `sample_weight`: https://github.com/microsoft/LightGBM/blob/84c4b7504a49eed63fa819a92db7d0b18292f3f4/python-package/lightgbm/dask.py#L293-L296

* it should be passed to `.fit()` as a Dask Array, Dask DataFrame, or Dask series
* it should have the same partitioning as `X` and `y`
* pass it around in `parts` just like we currently do with `sample_weight`

jmoralez:
Oh I meant the link in the summary points to `lgb.Dataset.set_init_score`. Ok, will try your proposed approach.

jmoralez:
[...]

Issue #3793, opened 2021-01-19
[dask] support 'raw_score' in predict()


jameslamb:
## Summary

See https://github.com/microsoft/LightGBM/pull/3774#issuecomment-763136825 for background.

To close this issue, add tests to https://github.com/microsoft/LightGBM/blob/master/tests/python_package_test/test_dask.py confirming that `.predict(X, raw_score=True)` works for `DaskLGBMClassifier` and `DaskLGBMRegressor` . For `DaskLGBMClassifier`, this should work with both `.predict()` and `.predict_proba()`.

## Motivation

Adding this feature would allow users to get raw predictions from LightGBM instead of converted output, which might be used as inputs to another model or for other exploratory purposes.

This would bring the Dask interface closer to parity with `lightgbm.sklearn`.

## References

When I first attempted this in #3774, I hit some issues related to how LightGBM uses `dask.DataFrame.map_partitions()` and `dask.Array.map_blocks()`. These function allow you to apply a function to each partition of a distributed data collection, like `X.map(some_func)`. In some situations, they require you to provide additional information about the shape and data types of the return from `some_func()`.

You can learn more about this in the `map_partitions()` docs: https://docs.dask.org/en/latest/dataframe-api.html#dask.dataframe.DataFrame.map_partitions

For testing this feature, you may find https://github.com/jameslamb/lightgbm-dask-testing useful.

jameslamb:
Closing this in favor of #2302, where we store all feature requests. Anyone is welcome to contribute this feature. Leave a comment on this issue if you'd like to work on it.

jmoralez:
I'd like to take this as well, James.

jameslamb:
sure, thanks! It's possible that this already works and that all we need are tests. Look to test_sklearn.py for inspiration

jmoralez:
I've repeated part in the tests that sets up the model factory and data from task and output a couple of times now. I want to extract that to a function like `_model_factory_and_data_from_task_and_output`. Should I open a different PR for that kind of refactor?

jameslamb:
I'd prefer to keep the data stuff as-is for now, please. There are already several layers of indirection in the data loading for those tests and I'd prefer not to add more right now.

You could reduce duplication in the `model_factory` stuff simply with a dictionary.

```shell
task_to_dask_factory = {
    "regression": lgb.DaskLGBMRegressor,
    "classification": lgb.DaskLGBMClassifier,
    "ranking": lgb.DaskLGBMRanker
[...]

PR #3950, opened 2021-02-13
[dask] Include support for init_score

jmoralez:
This solves #3807 by including the option to specify `init_score` as a dask collection to the `fit` methods of the dask estimators. A test is included to check that using `init_score` with dask and scikit-learn estimators results in the same result and that not using it results in a different one.

### DIFF
--- a/tests/python_package_test/test_dask.py
+++ b/tests/python_package_test/test_dask.py
@@ -1129,20 +1129,22 @@ def test_init_score(
     }
     init_score = np.repeat(init_score, y.size)
     local_model = local_model_factory(**params)
+    local_model_no_init = local_model_factory(**params)
     if task == 'ranking':
         local_model.fit(X, y, sample_weight=w, init_score=init_score, group=g)
+        local_model_no_init.fit(X, y, sample_weight=w, group=g)
     else:
         local_model.fit(X, y, sample_weight=w, init_score=init_score)
+        local_model_no_init.fit(X, y, sample_weight=w)
     local_preds = local_model.predict(X)
+    local_no_init_preds = local_model_no_init.predict(X)
 
     dask_init_score = da.from_array(init_score, chunks=y.size)
+    if output == 'dataframe':
+        dask_init_score = dd.from_array(dask_init_score)
     dask_model = dask_model_factory(client=client, **params)
     dask_model.fit(dX, dy, sample_weight=dw, init_score=dask_init_score, group=dg)
     dask_preds = dask_model.predict(dX).compute()
 
-    dask_model_no_init = dask_model_factory(client=client, **params)
-    dask_model_no_init.fit(dX, dy, sample_weight=dw, group=dg)
-    dask_no_init_preds = dask_model_no_init.predict(dX).compute()
-
     assert assert_eq(local_preds, dask_preds)
-    assert ~np.allclose(dask_preds, dask_no_init_preds)  # default init is 0
+    assert ~np.allclose(local_preds, local_no_init_preds)
--- a/tests/python_package_test/test_dask.py
+++ b/tests/python_package_test/test_dask.py
@@ -197,15 +197,6 @@ def _unpickle(filepath, serializer):
         raise ValueError(f'Unrecognized serializer type: {serializer}')
 
 
-def collection_to_single_partition(collection):
-    """Merge the parts of a Dask collection into a single partition."""
-    if collection is None:
-        return
-    if isinstance(collection, da.Array):
-        return collection.rechunk(*collection.shape)
-    return collection.repartition(npartitions=1)
-
-
 @pytest.mark.parametrize('output', data_output)
 @pytest.mark.parametrize('centers', data_centers)
 def test_classifier(output, centers, client, listen_port):
@@ -979,6 +970,14 @@ def test_training_succeeds_even_if_some_workers_do_not_have_any_data(client, tas
     if task == 'ranking' and output == 'scipy_csr_matrix':
         pytest.skip('LGBMRanker is not currently tested on sparse matrices')
 
+    def collection_to_single_partition(collection):
+        """Merge the parts of a Dask collection into a single partition."""
+        if collection is None:
+            return
+        if isinstance(collection, da.Array):
+            return collection.rechunk(*collection.shape)
+        return collection.repartition(npartitions=1)
+
     if task == 'ranking':
         X, y, w, g, dX, dy, dw, dg = _create_ranking_data(
             output=output,
@@ -1084,7 +1083,7 @@ def test_dask_methods_and_sklearn_equivalents_have_similar_signatures(methods):
 
 @pytest.mark.parametrize('task', tasks)
 @pytest.mark.parametrize('output', data_output)
-@pytest.mark.parametrize('init_score', [-1, 1])
+@pytest.mark.parametrize('init_score', [0.25, 0.75])
 def test_init_score(
     task,
     output,
@@ -1100,8 +1099,8 @@ def test_init_score(
             output=output,
             group=None
         )
-        dask_model_factory = lgb.DaskLGBMRanker
         local_model_factory = lgb.LGBMRanker
+        dask_model_factory = lgb.DaskLGBMRanker
     else:
         X, y, w, dX, dy, dw = _create_data(
             objective=task,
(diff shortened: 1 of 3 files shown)


tests/python_package_test/test_dask.py
```diff
@@ -1078,3 +1079,72 @@ def test_dask_methods_and_sklearn_equivalents_have_similar_signatures(methods):
...
+            local_model_factory = lgb.LGBMClassifier
+        elif task == 'regression':
+            dask_model_factory = lgb.DaskLGBMRegressor
+            local_model_factory = lgb.LGBMRegressor
+    # training on multiple workers can lead to different results
+    # we make sure to train on only one worker to compare the results
+    dX = collection_to_single_partition(dX)
+    dy = collection_to_single_partition(dy)
+    dw = collection_to_single_partition(dw)
+    dg = collection_to_single_partition(dg)
```
jameslamb, line 1121: please use multi-worker training like the other tests, for the same reasons I gave in https://github.com/microsoft/LightGBM/pull/3943/files#r575609015 and https://github.com/microsoft/LightGBM/pull/3943/files#r575617861

If that means that `assert_eq()` doesn't quite work, it's enough to just check that the predictions are different when init_score is used vs. when it is not used.

jameslamb (changes requested):
Thanks so much for picking this up! Please see my first round of comments.

<hr>

The `fit()` docs will also need to be updated to include `init_score`. They currently don't show it, intentionally.

https://lightgbm.readthedocs.io/en/latest/pythonapi/lightgbm.DaskLGBMClassifier.html#lightgbm.DaskLGBMClassifier.fit

image

Updating the API docs for the Dask estimators is a bit complicated:

* this template is used by both the Dask and sklearn estimators: https://github.com/microsoft/LightGBM/blob/89da9902657a172e7dd1a188e1b0afe059d075ba/python-package/lightgbm/sklearn.py#L192

Here's where we cut that and other arguments out of the docs for one of the estimators: https://github.com/microsoft/LightGBM/blob/89da9902657a172e7dd1a188e1b0afe059d075ba/python-package/lightgbm/dask.py#L626. There are 3 lines like that in `dask.py`. We do this slicing of the doc so that the Dask estimators can still inherit most documentation language from the sklearn estimators.

If you're interested in updating those, could you try it? You can build and test the docs locally like this:

```shell
cd docs/
make clean html
open _build/html/index.html
```

If you don't want to mess with the docs that is TOTALLY fine, and I can do it in a follow-up PR.

tests/python_package_test/test_dask.py
jmoralez: would this be something like fitting a model without init score and then fitting one with init score, computing the predictions for each one and doing `assert not np.allclose(regular_model_predictions, model_with_init_score_predictions)`?

jameslamb: I mean that you can change the tolerances in `assert assert_eq(local_preds, dask_preds)` if you need to.

And then yeah, `assert not assert_eq(regular_model_predictions, model_with_init_score_predictions)` is probably good enough for now. Until we support `raw_score` and we can make a more direct comparison (#3793 )

jmoralez:
Hi James. I'm having trouble with the tests for this, the results aren't very consistent between runs. Is there something else we can check more easily? I can see that when `init_score` is set, `value` for the root node is 0

jameslamb:
> Hi James. I'm having trouble with the tests for this, the results aren't very consistent between runs. Is there something else we can check more easily? I can see that when `init_score` is set, `value` for the root node is 0

for now, I think it's fine to just test that `dask_model.booster_.trees_to_dataframe()['value'][0] == 0.`. As far as I could tell, we don't have any tests on the `sklearn` module that use init_score, that you could use as a reference. I've been testing some things for the last two days but can't come up with a clean, more effective test. It requires some knowledge of the default objective functions that is a bit beyond my current knowledge.

tests/python_package_test/test_dask.py
```diff
@@ -1224,6 +1224,51 @@ def test_training_succeeds_when_data_is_dataframe_and_label_is_column_array(
...
+        'time_out': 5
+    }
+    if output.startswith('dataframe'):
+        init_scores = dy.map_partitions(lambda x: pd.Series([init_score] * x.size))
+    else:
+        init_scores = da.full_like(dy, fill_value=init_score, dtype='float')
+    model = model_factory(client=client, **params)
+    model.fit(dX, dy, sample_weight=dw, init_score=init_scores, group=dg)
+    # value of the root node is 0 when init_score is set
+    assert model._Booster.trees_to_dataframe()['value'][0] == 0
```
jameslamb, line 1267:
```suggestion
    assert model._booster.trees_to_dataframe()['value'][0] == 0
```

tests/python_package_test/test_dask.py
```diff
@@ -1224,6 +1224,51 @@ def test_training_succeeds_when_data_is_dataframe_and_label_is_column_array(
...
+
+    params = {
+        'n_estimators': 1,
+        'num_leaves': 2,
+        'time_out': 5
+    }
+    if output.startswith('dataframe'):
+        init_scores = dy.map_partitions(lambda x: pd.Series([init_score] * x.size))
+    else:
+        init_scores = da.full_like(dy, fill_value=init_score, dtype='float')
```
jameslamb, line 1263: oh cool! did not know about https://docs.dask.org/en/latest/array-api.html#dask.array.full_like

tests/python_package_test/test_dask.py
```diff
@@ -1224,6 +1224,51 @@ def test_training_succeeds_when_data_is_dataframe_and_label_is_column_array(
     client.close(timeout=CLIENT_CLOSE_TIMEOUT)
 
 
+@pytest.mark.parametrize('task', tasks)
+@pytest.mark.parametrize('output', data_output)
+@pytest.mark.parametrize('init_score', [0.25, 0.75])
```
jameslamb, line 1229: what is the purpose of testing two different values?

jameslamb (changes requested):

tests/python_package_test/test_dask.py
jmoralez: I had that initially because I was comparing to the local model and kind of wanted to avoid having an init_score equal to whatever it would choose automatically based on the label. I guess that can go away now

tests/python_package_test/test_dask.py
jmoralez: I think it's `_Booster`, I get this error: `AttributeError: 'DaskLGBMRegressor' object has no attribute '_booster'`.

jameslamb: sorry, typo
```suggestion
    assert model.booster_.trees_to_dataframe()['value'][0] == 0
```

tests/python_package_test/test_dask.py
jameslamb: ok then yes please just set it to one value. It can be some random nonsense values like `0.12345`

tests/python_package_test/test_dask.py
jmoralez: Yeah I wanted something like that for the dataframe too haha. Maybe the `if` statement would be more consistent if I used `map_blocks(lambda x: np.asarray([init_score] * x.size))`. What do you think?

jameslamb: yeah fair point! I have no strong preference either way, so whichever you prefer is good with me.

jmoralez:
I've included the requested changes, please let me know if you have more comments.

jameslamb (commented):
This looks great!

Thanks for taking on the documentation as well. I cloned your fork and built the docs to check, and they look good to me.

I'm leaving a "comment" review instead of approving because I'd like to request a review from @StrikerRUS as well before we merge, but from my perspective these changes are approved.

tests/python_package_test/test_dask.py
```diff
@@ -1224,6 +1225,50 @@ def test_training_succeeds_when_data_is_dataframe_and_label_is_column_array(
...
+    params = {
+        'n_estimators': 1,
+        'num_leaves': 2,
+        'time_out': 5
+    }
+    init_score = random.random()
+    if output.startswith('dataframe'):
+        init_scores = dy.map_partitions(lambda x: pd.Series([init_score] * x.size))
+    else:
+        init_scores = da.full_like(dy, fill_value=init_score, dtype='float')
```
StrikerRUS, line 1263: I'd prefer not to mix standard types and numpy ones.
```suggestion
        init_scores = da.full_like(dy, fill_value=init_score, dtype=np.float64)
```

StrikerRUS (approved):
Many thanks for your contribution!
Just one nit below.

jameslamb (approved):

Issue #3712, opened 2021-01-03
[dask] Add support for early stopping in Dask interface


jameslamb:
## Summary

`DaskLGBMClassifier` and `DaskLGBMRegressor` in the Python package should support early stopping.

## Motivation

Early stopping is generally useful with gradient boosting algorithms, to avoid wasted training iterations or unnecessary growth in the model size once desirable performance has been achieved. This feature is available in the non-Dask interfaces for LightGBM, and should be available with the Dask one.

## Description

This should mimic the approach XGBoost took (https://github.com/dmlc/xgboost/blob/516a93d25c3b6899558700430ffc99a29ea21e1a/python-package/xgboost/dask.py#L1386), where `eval_set` contains Dask collection s(Dask Array or Dask DataFrame).

## References

* relevant XGBoost PR (see linked issues and PRs): https://github.com/dmlc/xgboost/issues/5495
* one proposal: https://github.com/microsoft/LightGBM/pull/3515#issuecomment-752709132, https://github.com/h2oai/dask-lightgbm/commit/8cc8e83b0235050286a89dbbe2fd9d6b2e08271b

jameslamb:
Closing in favor of being in #2302 with other feature requests. Please leave a comment here if you'd like to work on this.

ffineis:
Hey I'd like to take this if it's cool - planning mainly base this off of [changes](https://github.com/h2oai/dask-lightgbm/commit/8cc8e83b0235050286a89dbbe2fd9d6b2e08271b) called out in #3515 and try to bring this in line with xgboost.dask.

jameslamb:
sure, thank you! I'm really close to having a small reproducible example for the random "cannot bind to port XXXX" issue, will link that here when I've written it up.

ffineis:
> sure, thank you! I'm really close to having a small reproducible example for the random "cannot bind to port XXXX" issue, will link that here when I've written it up.

Omigod lifesaver, thank you!

jameslamb:
@ffineis do you think you'll have time to work on this this week? We're planning to do a 3.2.0 release of LightGBM in the next week or two. I didn't include this in my list of must-have Dask features for the next release  (https://github.com/microsoft/LightGBM/pull/3872#issuecomment-768794191), but I'd love to try to get this change in if we can since it can have such a big impact on training runtime.

If you don't have time this week, could I take this back from you and try it out?

Thanks so much for all your help with the Dask module so far!!

ffineis:
[...]

Issue #3841, opened 2021-01-24
Write tests for parallel code


StrikerRUS:
I'm marking this issue with `question` label but hope to convert it to `feature request` after some discussions.

Right now LightGBM lacks any tests related to parallel (in located in [`network` folder](https://github.com/microsoft/LightGBM/tree/master/src/network)) code. `MPI` job which run on some our CIs simply execute ordinary tests with `serial` single machine tree learner but with dynamic library compiled with MPI support. In other words, we just check that MPI code can be compiled.

Recently added Dask module are getting more and more tests and this is very good but they are 
- run only on Linux so far;
- limited to Dask functionality;
- not very clear about errors in terms what is going wrong: either some Dask internal processes or LightGBM underlying code.

I believe it will be good to write some basic tests that will cover low-level LightGBM cpp code. Ideally it should cover both [socket](https://github.com/microsoft/LightGBM/blob/master/src/network/linkers_socket.cpp) and [MPI](https://github.com/microsoft/LightGBM/blob/master/src/network/linkers_mpi.cpp) implementations.

Adding such tests will help to improve standalone LightGBM parallel library and in consequence Dask-package built on top of it.

Linking #261.

Refer to https://github.com/microsoft/LightGBM/issues/3839#issue-792746004 for test example.

StrikerRUS:
OK, I see "thumb-up" reactions and based on them marking this issue with "feature request" label 🙂

StrikerRUS:
Closed in favor of being in #2302. We decided to keep all feature requests in one place.

Welcome to contribute this feature! Please re-open this issue (or post a comment if you are not a topic starter) if you are actively working on implementing this feature.

jmoralez:
Hi. I recently used the CLI interface to do "distributed" training on my local machine, so I just set the machine list to 127.0.0.1 with different ports, two config files and ran the lightgbm binary in two terminals. Would this be something along those lines?

jameslamb:
Yep exactly! We'd want a way to do that automatically in CI, but would support a PR that even just adds the tests and proposes a way to easily run and extend them.

jmoralez:
I've been trying this out and currently I have a python script that uses threads to spawn the machines from the CLI. I'm using the data and configurations provided in `examples/parallel_learning`. I have a couple of questions:

1. Is this a good approach?
2. Would the CI stuff run in github actions or on azure?
[...]

PR #3952, opened 2021-02-13
[dask] Early stopping

ffineis:
Attempt to address #3712 - lightgbm.dask support for early stopping. Implemented this to work with multiple eval sets (i.e. multiple (X, y) pairs), sample weights, group lengths, and implemented so that when an individual `eval_set`, `eval_sample_weights`, or `eval_group` is the same as (`data`, `label`), `sample_weights`, or `group`, just use the latter instead of having to recompute the training set/weights/group lengths.

This is all that's going on, making little mini eval sets out of delayed parts in a consistent manner:
dask_lgbm_eval_sets

Note that our test cases actually uncovered a small data issue - when one worker isn't distributed any `eval_set` data (e.g. because the validation set has fewer parts than workers, or because the data distribution is very unbalanced), then LightGBM throws an exception (because other workers do have eval_set data). This is why I added the `RuntimeError` - check for whether a worker has not received any eval_set parts when `_train` has been provided an `eval_set`.

### DIFF
--- a/python-package/lightgbm/dask.py
+++ b/python-package/lightgbm/dask.py
@@ -251,8 +251,13 @@ def _train_part(
                 **kwargs
             )
         else:
-            model.fit(data, label, sample_weight=weight, eval_set=local_eval_set,
-                      eval_sample_weight=local_eval_sample_weight, **kwargs)
+            model.fit(
+                data,
+                label,
+                sample_weight=weight,
+                eval_set=local_eval_set,
+                eval_sample_weight=local_eval_sample_weight, **kwargs
+            )
 
     finally:
         _safe_call(_LIB.LGBM_NetworkFree())
--- a/python-package/lightgbm/dask.py
+++ b/python-package/lightgbm/dask.py
@@ -186,10 +186,11 @@ def _train_part(
     if n_evals:
 
         local_eval_set = []
+        eval_names = []
         if has_eval_weights:
-            local_eval_sample_weight = list()
+            local_eval_sample_weight = []
         if is_ranker:
-            local_eval_group = list()
+            local_eval_group = []
 
         # consolidate parts of each individual eval component.
         for i in range(n_evals):
@@ -229,6 +230,11 @@ def _train_part(
                     else:
                         g_e.extend(eval_group[i])
 
+                local_eval_names = part.get('eval_names')
+                if local_eval_names:
+                    if len(local_eval_names) > len(eval_names):
+                        eval_names = local_eval_names
+
             # _concat each eval component.
             local_eval_set.append((_concat(x_e), _concat(y_e)))
             if w_e:
@@ -253,6 +259,7 @@ def _train_part(
                 eval_set=local_eval_set,
                 eval_sample_weight=local_eval_sample_weight,
                 eval_group=local_eval_group,
+                eval_names=eval_names if eval_names else None,
                 **kwargs
             )
         else:
@@ -261,7 +268,9 @@ def _train_part(
                 label,
                 sample_weight=weight,
                 eval_set=local_eval_set,
-                eval_sample_weight=local_eval_sample_weight, **kwargs
+                eval_sample_weight=local_eval_sample_weight,
+                eval_names=eval_names if eval_names else None,
+                **kwargs
             )
 
     finally:
@@ -290,6 +299,7 @@ def _train(
     sample_weight: Optional[_DaskCollection] = None,
     group: Optional[_DaskCollection] = None,
     eval_set: Optional[List[Tuple[_DaskCollection, _DaskCollection]]] = None,
+    eval_names: Optional[List[str]] = None,
     eval_sample_weight: Optional[List[_DaskCollection]] = None,
     eval_group: Optional[List[_DaskCollection]] = None,
     **kwargs: Any
@@ -318,6 +328,8 @@ def _train(
         where the first 10 records are in the first group, records 11-30 are in the second group, records 31-70 are in the third group, etc.
     eval_set : List of (X, y) tuples of Dask data collections, or None, optional (default=None)
         List of (X, y) tuple pairs to use as validation sets.
+    eval_names: List of strings or None, optional (default=None))
+        Names of eval_set.
     eval_sample_weight: List of Dask data collections or None, optional (default=None)
         List of Dask Array or Dask Series, weights for each validation set in eval_set.
     eval_group: List of Dask data collections or None, optional (default=None)
@@ -396,7 +408,7 @@ def _train(
         for i, (X, y) in enumerate(eval_set):
 
             # when individual eval set is equivalent to training data, skip recomputing parts.
-            if X is data and y is label
+            if X is data and y is label:
                 for parts_idx in range(n_parts):
                     eval_sets[parts_idx].append('__train__')
 
@@ -461,6 +473,8 @@ def _train(
         # assign sub-eval_set components to worker parts.
         for parts_idx, e_set in eval_sets.items():
             parts[parts_idx]['eval_set'] = e_set
+            if eval_names:
+                parts[parts_idx]['eval_names'] = [eval_names[i] for i in range(len(e_set))]
             if eval_sample_weight:
                 parts[parts_idx]['eval_sample_weight'] = eval_sample_weights[parts_idx]
             if eval_group:
@@ -654,17 +668,33 @@ class _DaskLGBMModel:
         X: _DaskMatrixLike,
         y: _DaskCollection,
         sample_weight: Optional[_DaskCollection] = None,
+        init_score: Optional[List[_DaskCollection]] = None,
         group: Optional[_DaskCollection] = None,
         eval_set: Optional[List[Tuple[_DaskCollection, _DaskCollection]]] = None,
+        eval_names: Optional[List[str]] = None,
         eval_sample_weight: Optional[List[_DaskCollection]] = None,
+        eval_class_weight: Optional[Union[dict, str]] = None,
+        eval_init_score: Optional[List[_DaskCollection]] = None,
         eval_group: Optional[List[_DaskCollection]] = None,
+        eval_metric: Optional[Union[Callable, str, List[Union[Callable, str]]]] = None,
+        eval_stopping_rounds: Optional[int] = None,
         **kwargs: Any
     ) -> "_DaskLGBMModel":
         if not all((DASK_INSTALLED, PANDAS_INSTALLED, SKLEARN_INSTALLED)):
             raise LightGBMError('dask, pandas and scikit-learn are required for lightgbm.dask')
 
+        not_supported = ['init_score', 'eval_init_score', 'eval_class_weight']
+        for ns in not_supported:
+            if eval(ns) is not None:
+                raise RuntimeError(f'{ns} is not currently supported in lightgbm.dask')
+
         params = self.get_params(True)
         params.pop("client", None)
+
+        if eval_metric:
+            kwargs['eval_metric'] = eval_metric
+        if eval_stopping_rounds:
+            kwargs['eval_stopping_rounds'] = eval_stopping_rounds
 
         model = _train(
             client=_get_dask_client(self.client),
@@ -675,6 +705,7 @@ class _DaskLGBMModel:
             sample_weight=sample_weight,
             group=group,
             eval_set=eval_set,
+            eval_names=eval_names,
             eval_sample_weight=eval_sample_weight,
             eval_group=eval_group,
             **kwargs
@@ -775,13 +806,24 @@ class DaskLGBMClassifier(LGBMClassifier, _DaskLGBMModel):
         sample_weight: Optional[_DaskCollection] = None,
         init_score: Optional[_DaskCollection] = None,
         eval_set: Optional[List[Tuple[_DaskCollection, _DaskCollection]]] = None,
+        eval_names: Optional[List[str]] = None,
         eval_sample_weight: Optional[List[_DaskCollection]] = None,
+        eval_class_weight: Optional[List[_DaskCollection]] = None,
         eval_init_score: Optional[List[_DaskCollection]] = None,
+        eval_metric: Optional[Union[Callable, str, List[Union[Callable, str]]]] = None,
+        eval_stopping_rounds: Optional[int] = None,
         **kwargs: Any
     ) -> "DaskLGBMClassifier":
         """Docstring is inherited from the lightgbm.LGBMClassifier.fit."""
-        if init_score is not None or eval_init_score is not None:
-            raise RuntimeError('init_score and eval_init_score are not currently supported in lightgbm.dask')
+        not_supported = ['init_score', 'eval_init_score', 'eval_class_weight']
+        for ns in not_supported:
+            if eval(ns) is not None:
+                raise RuntimeError(f'{ns} is not currently supported in lightgbm.dask')
+
+        if eval_metric:
+            kwargs['eval_metric'] = eval_metric
+        if eval_stopping_rounds:
+            kwargs['eval_stopping_rounds'] = eval_stopping_rounds
 
         return self._fit(
             model_factory=LGBMClassifier,
@@ -789,6 +831,7 @@ class DaskLGBMClassifier(LGBMClassifier, _DaskLGBMModel):
             y=y,
             sample_weight=sample_weight,
             eval_set=eval_set,
+            eval_names=eval_names,
             eval_sample_weight=eval_sample_weight,
             **kwargs
         )
@@ -901,13 +944,24 @@ class DaskLGBMRegressor(LGBMRegressor, _DaskLGBMModel):
         sample_weight: Optional[_DaskCollection] = None,
         init_score: Optional[_DaskCollection] = None,
         eval_set: Optional[List[Tuple[_DaskCollection, _DaskCollection]]] = None,
+        eval_names: Optional[List[str]] = None,
         eval_sample_weight: Optional[List[_DaskCollection]] = None,
+        eval_class_weight: Optional[List[_DaskCollection]] = None,
         eval_init_score: Optional[List[_DaskCollection]] = None,
+        eval_metric: Optional[Union[Callable, str, List[Union[Callable, str]]]] = None,
+        eval_stopping_rounds: Optional[int] = None,
         **kwargs: Any
     ) -> "DaskLGBMRegressor":
         """Docstring is inherited from the lightgbm.LGBMRegressor.fit."""
-        if init_score is not None or eval_init_score is not None:
-            raise RuntimeError('init_score and eval_init_score are not currently supported in lightgbm.dask')
+        not_supported = ['init_score', 'eval_init_score', 'eval_class_weight']
+        for ns in not_supported:
+            if eval(ns) is not None:
+                raise RuntimeError(f'{ns} is not currently supported in lightgbm.dask')
+
+        if eval_metric:
+            kwargs['eval_metric'] = eval_metric
+        if eval_stopping_rounds:
+            kwargs['eval_stopping_rounds'] = eval_stopping_rounds
 
         return self._fit(
             model_factory=LGBMRegressor,
@@ -915,6 +969,7 @@ class DaskLGBMRegressor(LGBMRegressor, _DaskLGBMModel):
             y=y,
             sample_weight=sample_weight,
             eval_set=eval_set,
+            eval_names=eval_names,
             eval_sample_weight=eval_sample_weight,
             **kwargs
         )
@@ -1016,14 +1071,25 @@ class DaskLGBMRanker(LGBMRanker, _DaskLGBMModel):
         init_score: Optional[_DaskCollection] = None,
         group: Optional[_DaskCollection] = None,
         eval_set: Optional[List[Tuple[_DaskCollection, _DaskCollection]]] = None,
+        eval_names: Optional[List[str]] = None,
         eval_sample_weight: Optional[List[_DaskCollection]] = None,
+        eval_class_weight: Optional[List[_DaskCollection]] = None,
         eval_init_score: Optional[List[_DaskCollection]] = None,
         eval_group: Optional[List[_DaskCollection]] = None,
+        eval_metric: Optional[Union[Callable, str, List[Union[Callable, str]]]] = None,
+        eval_stopping_rounds: Optional[int] = None,
         **kwargs: Any
     ) -> "DaskLGBMRanker":
         """Docstring is inherited from the lightgbm.LGBMRanker.fit."""
-        if init_score is not None or eval_init_score is not None:
-            raise RuntimeError('init_score and eval_init_score are not currently supported in lightgbm.dask')
+        not_supported = ['init_score', 'eval_init_score', 'eval_class_weight']
+        for ns in not_supported:
+            if eval(ns) is not None:
+                raise RuntimeError(f'{ns} is not currently supported in lightgbm.dask')
+
+        if eval_metric:
+            kwargs['eval_metric'] = eval_metric
+        if eval_stopping_rounds:
+            kwargs['eval_stopping_rounds'] = eval_stopping_rounds
 
         return self._fit(
             model_factory=LGBMRanker,
@@ -1032,6 +1098,7 @@ class DaskLGBMRanker(LGBMRanker, _DaskLGBMModel):
             sample_weight=sample_weight,
             group=group,
             eval_set=eval_set,
+            eval_names=eval_names,
             eval_sample_weight=eval_sample_weight,
             eval_group=eval_group,
             **kwargs
--- a/python-package/lightgbm/dask.py
+++ b/python-package/lightgbm/dask.py
@@ -408,7 +408,6 @@ def _train(
                         eval_sets[parts_idx].append(([x_e], [y_e]))
 
                     else:
-                        # n_evals = len(eval_sets[parts_idx]) - 1
                         eval_sets[parts_idx][-1][0].append(x_e)
                         eval_sets[parts_idx][-1][1].append(y_e)
 
--- a/python-package/lightgbm/dask.py
+++ b/python-package/lightgbm/dask.py
@@ -193,7 +193,10 @@ def _train_part(
 
         # consolidate parts of each individual eval component.
         for i in range(n_evals):
-            x_e, y_e, w_e, g_e = [], [], [], []
+            x_e = []
+            y_e = []
+            w_e = []
+            g_e = []
             for part in list_of_parts:
 
                 if not part.get('eval_set'):
--- a/python-package/lightgbm/dask.py
+++ b/python-package/lightgbm/dask.py
@@ -144,6 +144,7 @@ def _train_part(
     worker_address_to_port: Dict[str, int],
     return_model: bool,
     time_out: int = 120,
+    evals_provided: bool = False,
     **kwargs: Any
 ) -> Optional[LGBMModel]:
     local_worker_address = get_worker().address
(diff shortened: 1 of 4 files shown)


python-package/lightgbm/dask.py
```diff
@@ -282,17 +354,100 @@ def _train(
     data_parts = _split_to_parts(data=data, is_matrix=True)
     label_parts = _split_to_parts(data=label, is_matrix=False)
     parts = [{'data': x, 'label': y} for (x, y) in zip(data_parts, label_parts)]
+    n_parts = len(parts)
 
     if sample_weight is not None:
         weight_parts = _split_to_parts(data=sample_weight, is_matrix=False)
-        for i in range(len(parts)):
+        for i in range(n_parts):
```
jameslamb, line 361: I totally support ths simplification. Could you make this a separate PR? Doesnt require new tests and we can merge it immediately

jameslamb (commented):
Thanks so much for this!! I'll give it a thorough review tonight or tomorrow

python-package/lightgbm/dask.py
ffineis: Hey yes, you're just referring to the swapping of `len(parts)` for `n_parts`, right?

tests/python_package_test/test_dask.py
```diff
@@ -647,6 +647,98 @@ def test_ranker(output, client, listen_port, group):
...
+            print(f'_accuracy_score(dy, p1) = {_accuracy_score(dy, p1)}')
+            _accuracy_score(dy, p1) > 0.8
+
+        elif task == 'regression':
+            print(f'_r2_score(dy, p1) = {_r2_score(dy, p1)}')
+            _r2_score(dy, p1) > 0.8
+
+        else:
+            print(f'spearmanr(p1.compute(), y).correlation = {spearmanr(p1.compute(), y).correlation}')
+            assert spearmanr(p1.compute(), y).correlation > 0.8
```
jameslamb, line 737:
```suggestion
            p1_cor = spearmanr(p1.compute(), y).correlation
            msg = f'spearman correlation of predictions with actuals was <= 0.8 ({p1_cor})'
            assert p1_cor > 0.8, msg
```

`print()` statements can get easily lost in `pytest`'s output, but the text of a failure is always easy to see because it gets collected together with the test failures at the end.

Could you change this and the tests above to show that information if the test fails, instead of printing it?

And can you also store the result used in these messages in a variable, instead of computing it twice?

tests/python_package_test/test_dask.py
```diff
@@ -647,6 +647,98 @@ def test_ranker(output, client, listen_port, group):
...
+
+    client.close(timeout=CLIENT_CLOSE_TIMEOUT)
+
```
jameslamb, line 740: will `.evals_set_` also be set on the returned model?

https://github.com/microsoft/LightGBM/blob/fd094d2719a04b984b376a665b6c3e5e1aff2967/tests/python_package_test/test_sklearn.py#L155

If so can you add some test logic checking that that looks like you'd expect? i.e. correct metric(s), correct shape.

tests/python_package_test/test_dask.py
```diff
@@ -647,6 +647,98 @@ def test_ranker(output, client, listen_port, group):
...
+        p1 = dask_model.predict(dX)
+        if task == 'classification':
+            print(f'_accuracy_score(dy, p1) = {_accuracy_score(dy, p1)}')
+            _accuracy_score(dy, p1) > 0.8
+
+        elif task == 'regression':
+            print(f'_r2_score(dy, p1) = {_r2_score(dy, p1)}')
+            _r2_score(dy, p1) > 0.8
+
+        else:
```
jameslamb, line 735: Could you please add some additional checks on other attributes that will be set  when early stopping is enabled?

`best_score_` and `best_iteration_`? You can look in any of the sklearn tests that use `early_stopping_rounds` for reference: https://github.com/microsoft/LightGBM/blob/fd094d2719a04b984b376a665b6c3e5e1aff2967/tests/python_package_test/test_sklearn.py#L114

python-package/lightgbm/dask.py
```diff
@@ -593,14 +755,23 @@ def fit(
         X: _DaskMatrixLike,
         y: _DaskCollection,
         sample_weight: Optional[_DaskCollection] = None,
+        init_score: Optional[_DaskCollection] = None,
+        eval_set: Optional[List[Tuple[_DaskCollection, _DaskCollection]]] = None,
```
jameslamb, line 759: Could you please also pass through `eval_names`? Ideally I'd like these methods to have the same argument position as `lgb.sklearn.LGBMModel.fit()`.

https://github.com/microsoft/LightGBM/blob/fd094d2719a04b984b376a665b6c3e5e1aff2967/python-package/lightgbm/sklearn.py#L544

and can you please do this in `DaskLGBMRegressor.fit()` and `DaskLGBMRanker.fit()` as well?

python-package/lightgbm/dask.py
```diff
@@ -593,14 +755,23 @@ def fit(
...
+        eval_sample_weight: Optional[List[_DaskCollection]] = None,
```
jameslamb, line 760: could you add `eval_class_weight` right here, and then raise an exception if it's non-None and says that it isn't yet supported? I don't think it should be in this PR and I can write up a feature request for it.

But I'd like these methods to have the same argument positions as `lgb.sklearn.LGBMModel.fit()`:

https://github.com/microsoft/LightGBM/blob/fd094d2719a04b984b376a665b6c3e5e1aff2967/python-package/lightgbm/sklearn.py#L545

and can you please do this in `DaskLGBMRegressor.fit()` and `DaskLGBMRanker.fit()` as well?

python-package/lightgbm/dask.py
```diff
@@ -176,12 +177,74 @@ def _train_part(
     else:
         group = None
 
+    # construct local eval_set data.
+    local_eval_set, local_eval_sample_weight, local_eval_group = None, None, None
```
jameslamb, line 181:
```suggestion
    local_eval_set = None
    local_eval_sample_weight = None
    local_eval_group = None
```

can you please move these to individual lines? I think that's a bit easier to read, but more importantly it makes [the git blame output](https://github.com/microsoft/LightGBM/blame/fd094d2719a04b984b376a665b6c3e5e1aff2967/python-package/lightgbm/dask.py) more useful

python-package/lightgbm/dask.py
```diff
@@ -176,12 +177,74 @@ def _train_part(
...
+
+        local_eval_set = []
+        if has_eval_weights:
+            local_eval_sample_weight = list()
+        if is_ranker:
+            local_eval_group = list()
+
+        # consolidate parts of each individual eval component.
+        for i in range(n_evals):
+            x_e, y_e, w_e, g_e = [], [], [], []
```
jameslamb, line 194:
```suggestion
            x_e = []
            y_e = []
            w_e = []
            g_e = []
```

Same comment as above. I have a preference for one assignment per line

python-package/lightgbm/dask.py
```diff
@@ -232,6 +298,12 @@ def _train(
         sum(group) = n_samples.
         For example, if you have a 100-document dataset with ``group = [10, 20, 40, 10, 10, 10]``, that means that you have 6 groups,
         where the first 10 records are in the first group, records 11-30 are in the second group, records 31-70 are in the third group, etc.
+    eval_set : List of (X, y) tuples of Dask data collections, or None, optional (default=None)
+        List of (X, y) tuple pairs to use as validation sets.
+    eval_sample_weight: List of Dask data collections or None, optional (default=None)
+        List of dask Array or dask Series, weights for each validation set in eval_set.
```
jameslamb, line 304:
```suggestion
        List of Dask Array or Dask Series, weights for each validation set in eval_set.
```

We try to use "Dask" (capitalized) when referring to the package and "dask" (lowercase) when referring to a module in the library, based on https://marketing.dask.org/en/latest/refer.html

python-package/lightgbm/dask.py
```diff
@@ -232,6 +298,12 @@ def _train(
...
+    eval_group: List of Dask data collections or None, optional (default=None)
+        List of dask Array or dask Series, group/query for each validation set in eval_set.
```
jameslamb, line 306:
```suggestion
        List of Dask Array or Dask Series, group/query for each validation set in eval_set.
```

python-package/lightgbm/dask.py
```diff
@@ -282,17 +354,100 @@ def _train(
...
+        eval_sets = defaultdict(list)
+        if eval_sample_weight:
+            eval_sample_weights = defaultdict(list)
+        if eval_group:
+            eval_groups = defaultdict(list)
+
+        for i, (X, y) in enumerate(eval_set):
+
+            # when individual eval set is equivalent to training data, skip recomputing parts.
+            if id(X) == id(data) and id(y) == id(label):
```
jameslamb, line 381:
```suggestion
            if X is data and y is label
```

Is there a particular reason to prefer `id()` to using `is`? I honestly don't know if there's any meaningful difference.

For this check in `sklearn` we use `is`, so if you don't have a strong preference I'd prefer to change to `is` for consistency and because I think people who might come contribute are more likely to be familiar with `is` than they are with `id()`.

https://github.com/microsoft/LightGBM/blob/fd094d2719a04b984b376a665b6c3e5e1aff2967/python-package/lightgbm/sklearn.py#L663

python-package/lightgbm/dask.py
```diff
@@ -282,17 +354,100 @@ def _train(
...
+                for j in range(len(eval_x_parts)):
+                    parts_idx = j % n_parts
+
+                    x_e, y_e = eval_x_parts[j], eval_y_parts[j]
+
+                    if j < n_parts:
+                        eval_sets[parts_idx].append(([x_e], [y_e]))
+
+                    else:
+                        # n_evals = len(eval_sets[parts_idx]) - 1
```
jameslamb, line 398:
```suggestion
```

looks like this was left over from development

python-package/lightgbm/dask.py
```diff
@@ -282,17 +354,100 @@ def _train(
...
+                    eval_sets[parts_idx].append('__train__')
+
+            else:
+                eval_x_parts = _split_to_parts(data=X, is_matrix=True)
+                eval_y_parts = _split_to_parts(data=y, is_matrix=False)
+
+                for j in range(len(eval_x_parts)):
+                    parts_idx = j % n_parts
+
+                    x_e, y_e = eval_x_parts[j], eval_y_parts[j]
```
jameslamb, line 392:
```suggestion
                    x_e = eval_x_parts[j]
                    y_e = eval_y_parts[j]
```

python-package/lightgbm/dask.py
```diff
@@ -331,6 +486,7 @@ def _train(
             worker_address_to_port=worker_address_to_port,
             time_out=params.get('time_out', 120),
             return_model=(worker == master_worker),
+            evals_provided=eval_set is not None,
```
jameslamb, line 489: instead of passing through a flag, could this just be computed inside `_train`? I'd like to limit the use of passed-through flags to control internal behavior, if possible. I've found in the past that that creates unnecessary risk of mistakes.

python-package/lightgbm/dask.py
jameslamb: yep exactly, just that little bit

jameslamb (changes requested):
Ok, I took a first look through this. I think the approach you took to knitting the parts together makes sense, and thanks for dealing with the added complexity of the fact that every worked has to have at least a little bit of evaluation data!

I still need to test this and go through it again more carefully, but I left an initial round of suggestions for your consideration.

<hr>

We also do some weird stuff in LightGBM's docs for the Dask estimators, and right now all of the `eval_*` things are hidden in those docs. I described this in https://github.com/microsoft/LightGBM/pull/3950#pullrequestreview-589943673. If you see that and are brave enough, I'd appreciate if you could update the docs in this PR as well. But if you look and it seems too complicated based on the amount of time you available to contribute, no problem at all and I can do it in a later PR.

StrikerRUS:
@jameslamb 
> If you see that and are brave enough, I'd appreciate if you could update the docs in this PR as well.

Just want to remind you that you will not be able to see the final result at RTD because this PR is from a fork repo.

jameslamb:
> @jameslamb
> 
> > If you see that and are brave enough, I'd appreciate if you could update the docs in this PR as well.
> 
> Just want to remind you that you will not be able to see the final result at RTD because this PR is from a fork repo.

I understand. The instructions at https://github.com/microsoft/LightGBM/blob/master/docs/README.rst#build have worked well for me in testing the docs locally.

python-package/lightgbm/dask.py
```diff
@@ -176,12 +177,74 @@ def _train_part(
...
+        if evals_provided:
+            msg = "eval_set was provided but worker %s was not allocated validation data. Try rebalancing data across workers."
+            raise RuntimeError(msg % local_worker_address)
+
     try:
         model = model_factory(**params)
         if is_ranker:
-            model.fit(data, label, sample_weight=weight, group=group, **kwargs)
+            model.fit(data, label, sample_weight=weight, group=group, eval_set=local_eval_set,
+                      eval_sample_weight=local_eval_sample_weight, eval_group=local_eval_group, **kwargs)
```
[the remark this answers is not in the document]
ffineis: Yeah I'm totally with this, this is how I normally write params but didn't want to abuse style!

tests/python_package_test/test_dask.py
```diff
@@ -647,6 +647,98 @@ def test_ranker(output, client, listen_port, group):
...
+            dask_model.fit(dX, dy, group=dg, eval_set=eval_set,
+                           eval_sample_weight=eval_sample_weight, eval_group=eval_group,
+                           eval_metric=eval_metric, early_stopping_rounds=5)
+
+    else:
+        dask_model = dask_model.fit(dX, dy, group=dg, eval_set=eval_set,
+                                    eval_sample_weight=eval_sample_weight, eval_group=eval_group,
+                                    eval_metric=eval_metric, early_stopping_rounds=5)
+        fitted_trees = dask_model.to_local().booster_.num_trees()
+        assert fitted_trees < full_trees
```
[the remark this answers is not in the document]
ffineis: Ok yeah that makes sense. Definitely.

tests/python_package_test/test_dask.py
```diff
@@ -647,6 +647,98 @@ def test_ranker(output, client, listen_port, group):
...
+        with pytest.raises(RuntimeError, match='eval_set was provided but worker .* was not allocated validation data'):
+            dask_model.fit(dX, dy, group=dg, eval_set=eval_set,
+                           eval_sample_weight=eval_sample_weight, eval_group=eval_group,
+                           eval_metric=eval_metric, early_stopping_rounds=5)
+
+    else:
+        dask_model = dask_model.fit(dX, dy, group=dg, eval_set=eval_set,
+                                    eval_sample_weight=eval_sample_weight, eval_group=eval_group,
+                                    eval_metric=eval_metric, early_stopping_rounds=5)
+        fitted_trees = dask_model.to_local().booster_.num_trees()
```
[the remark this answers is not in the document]
ffineis: TIL

tests/python_package_test/test_dask.py
ffineis: No yeah this was an oversight, sorry! The test stuff I got out in a rush.

python-package/lightgbm/dask.py
ffineis: https://github.com/microsoft/LightGBM/pull/3985

tests/python_package_test/test_dask.py
ffineis: It does come with `evals_result`, good call will add!

python-package/lightgbm/dask.py
ffineis: TIL!

tests/python_package_test/test_dask.py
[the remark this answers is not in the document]
jameslamb: oooooooo yeah I like that! One other thing I thought of tonight we could be even more sure by setting `early_stopping_rounds=1`

tests/python_package_test/test_dask.py
jameslamb: awesome! That feature is valuable even outside of just early stopping, to see how much improvement you were getting from each additional iteration (for example).

Could you add a unit test that passes eval sets but NOT `early_stopping_rounds`? Just to be sure that those two things aren't tightly coupled here

python-package/lightgbm/dask.py
jameslamb: my teammate @rikturr taught me about this last week 😀

python-package/lightgbm/dask.py
jameslamb: merged, thanks. Whenever you merge `master` back into this branch, those lines should drop out of the diff.

python-package/lightgbm/dask.py
ffineis: Yeah I just took this from xgboost_dask.py - https://github.com/dmlc/xgboost/blob/master/python-package/xgboost/dask.py#L865

I mean `is` is easier to read, let's do it.

python-package/lightgbm/dask.py
ffineis: Yeah for sure. Should I add `eval_metric` and `early_stopping_rounds` too, or just assume they're part of `kwargs`?

jameslamb: oh right! Yes please, adding them here would be good and I think it's in scope for this PR

python-package/lightgbm/dask.py
ffineis: Yeah so this flag is trying to communicate to each worker whether or not an `eval_set` was provided to the dask estimator, so that if any worker was not provided any eval_data, it should fail out (each worker needs at least a little bit). So I can add this flag as an element of each `part` instead, but just an FYI it's not as clean as a single flag one-liner.

jameslamb: OH I see, ok I misunderstood. I think it's ok to leave here

ffineis:
> > @jameslamb
> > > If you see that and are brave enough, I'd appreciate if you could update the docs in this PR as well.
> > 
> > 
> > Just want to remind you that you will not be able to see the final result at RTD because this PR is from a fork repo.
> 
> I understand. The instructions at https://github.com/microsoft/LightGBM/blob/master/docs/README.rst#build have worked well for me in testing the docs locally.

Hey yeah I think I can handle opening up the docs for the eval* params. I mean the current kind of surgery on the doc strings is just eliminating a whole section of parameters, so I need to open them back up, but probably still keep... `eval_init_score` silent for the time being. I'm going to do all of these more code-related tune-ups first and the come back to address the docs once the functional parts + tests are looking good.

python-package/lightgbm/dask.py
ffineis: Addressed in 6737acd5

python-package/lightgbm/dask.py
ffineis: Addressed in 18cd42c

tests/python_package_test/test_dask.py
ffineis: Addressed in 7099204

tests/python_package_test/test_dask.py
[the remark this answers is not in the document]
ffineis: Addressed in 7099204

tests/python_package_test/test_dask.py
ffineis: Addressed in 1335591

tests/python_package_test/test_dask.py
ffineis: Addressed in 1335591

jameslamb:
@ffineis I can review this again whenever you've merged in `master` and fixed conflicts

ffineis:
> @ffineis I can review this again whenever you've merged in `master` and fixed conflicts

Awesome, thank you would appreciate that.

Hey re: docstrings - I'm trying to implement docs for the `eval*` parameters, but sklearn's `_lgbmmodel_doc_fit` does not provide a template for any parameter besides `X`, `y`, `sample_weight`, and `group`, so I think that when I bring back to life a parameter like `eval_sample_weight`, the docs will still read "list of **arrays** or None, optional (default=None)".

How should we handle this? Should [sklearn's](https://github.com/microsoft/LightGBM/blob/89da9902657a172e7dd1a188e1b0afe059d075ba/python-package/lightgbm/sklearn.py#L707) `_lgbmmodel_doc_fit` be made more flexible, or should we just use this kind of short-circuiting approach, like

```
  fit.__doc__ = (
      _base_doc[:_base_doc.find('eval_sample_weight :')]
      + 'eval_sample_weight :  list of Dask Array, Dask DataFrame or Dask Series, or None, optional (default=None).\n'
)
```
for each parameter where the sklearn docs mention numpy arrays?

jameslamb:
> > @ffineis I can review this again whenever you've merged in `master` and fixed conflicts
> 
> Awesome, thank you would appreciate that.
> 
> Hey re: docstrings - I'm trying to implement docs for the `eval*` parameters, but sklearn's `_lgbmmodel_doc_fit` does not provide a template for any parameter besides `X`, `y`, `sample_weight`, and `group`, so I think that when I bring back to life a parameter like `eval_sample_weight`, the docs will still read "list of **arrays** or None, optional (default=None)".
> 
> How should we handle this? Should [sklearn's](https://github.com/microsoft/LightGBM/blob/89da9902657a172e7dd1a188e1b0afe059d075ba/python-package/lightgbm/sklearn.py#L707) `_lgbmmodel_doc_fit` be made more flexible by adding more templates so that it doesn't explicitly refer to numpy arrays? Or should we just use this kind of short-circuiting approach, like
> 
> ```
>   fit.__doc__ = (
>       _base_doc[:_base_doc.find('eval_sample_weight :')]
>       + 'eval_sample_weight :  list of Dask Array, Dask DataFrame or Dask Series, or None, optional (default=None).\n'
>       + ' ' * 12 + 'Weights of eval data.\n'
> )
> ```

you can edit `_lgbmmodel_doc_fit` in sklearn and aadd more template strings like `{eval_weight_shape}`. I think that's easier to understand and modify than the short-circuiting thing. And it means that you only have to edit descriptions like "weights of eval data" in one place.

ffineis:
Hey, sorry for the delay in communication. The code and tests are basically how I want them, but I've been testing this out over the last week via pytest and in a jupyterlab notebook. When I `fit` a DaskLGBM estimator using `eval_set` one of three things happen:

1. Model trains without failures.
2. Mysterious `Socket recv error code, 104`. In this case, the worker graph shows one worker that has successfully-stored `results = client.gather(futures_classifiers)` in memory (I'm guessing the worker *not* chosen to receive the fitted estimator) while the other worker has an `erred` task.
Screen Shot 2021-02-22 at 10 57 09 PM
socket_recv_error_104_graph
3. Indefinite hanging. In the worker graph I've noticed a few peculiar things are happing while DaskLightGBM is hanging: a) one worker remains pre-occupied with something - one worker's CPU usage is 1-3%, while the other's CPU is ~100%. I don't think memory is the issue - 1GB allocated to each worker who each rock < 200Mb each. The occupied worker has tasks that are in the `processing` state. When I hit "stop" in jupyterlab, this triggers a KeyboardInterrupt Error, but then this seems to unblocks the log-jam and model training all of the sudden continues through to early stopping (still with an error due to having to interrupt processing though). Inspecting the hanging task - it's a 'suspicious' task and clearly has something to do with a LightGBMException that I think may not be getting raised correctly?
task_hanging_graph
Screen Shot 2021-02-22 at 10 21 59 PM
Screen Shot 2021-02-24 at 10 14 52 PM
Screen Shot 2021-02-24 at 10 15 11 PM

After googling "Dask task hanging," I saw the note about [oversubscribed threads](tests/python_package_test/test_dask.py). So I attempted a fix where I set `num_threads: 1` (instead of the default `client.ncores` for each worker). This did not resolve anything.

If it's any use, both the `Socket recv error code, 104` and the hanging seem to be related as they both happen right after the moment when LightGBM has decided "yep, time to stop training and assemble the final model," basically the equivalent of a Spark `collect` step.
Screen Shot 2021-02-20 at 12 03 07 AM

Totally open to any/all advice or insights here!

jameslamb (changes requested):
Man sorry it took SO LONG to get back to you. I've finished the known issues I wanted to get done for the next release (#3872 ), so now I can fully focus on this.

I cloned your branch and ran some tests locally, and I was able to exactly reproduce the behavior you thoroughly documented in https://github.com/microsoft/LightGBM/pull/3952#issuecomment-785606610. Thanks for describing that so well!

I just focused on the regression tests with a single eval set, since that was enough to reproduce the issue.

<details><summary>testing code</summary>

```python
import lightgbm as lgb

import dask.array as da
import numpy as np
from dask.array.utils import assert_eq
from dask.distributed import Client, LocalCluster


def _create_data(objective, n_samples=100, n_features=None, centers=2, output='array', chunk_size=50, random_weights=True):
    n_features = n_features if n_features else 100
    X, y = make_regression(n_samples=n_samples, n_features=n_features, random_state=42)
    rnd = np.random.RandomState(42)
    weights = rnd.random(X.shape[0]) * 0.01
    if not random_weights:
        weights = np.ones([X.shape[0]])
    dX = da.from_array(X, (chunk_size, X.shape[1]))
    dy = da.from_array(y, chunk_size)
    dw = da.from_array(weights, chunk_size)

    return X, y, weights, dX, dy, dw


def _r2_score(dy_true, dy_pred):
    denominator = ((dy_true - dy_pred.mean(axis=0)) ** 2).sum(axis=0, dtype=np.float64)
    return (1 - numerator / denominator).compute()


def _accuracy_score(dy_true, dy_pred):
    return da.average(dy_true == dy_pred).compute()

cluster = LocalCluster(n_workers=2)
client = Client(cluster)

task = 'regression'
eval_sizes = [0.9]
eval_names_prefix = 'specified'

client.restart()
# use larger number of samples to prevent faux early stopping whereby
# boosting stops on accident because each worker has few data points and achieves 0 loss.
n_samples = 1000
n_features = 10
n_eval_sets = len(eval_sizes)
early_stopping_rounds = 1
eval_set = []
eval_sample_weight = []

eval_names = [eval_names_prefix + f'_{i}' for i in range(len(eval_sizes))]

X, y, w, dX, dy, dw = _create_data(
    n_samples=n_samples,
    n_features=n_features,
    objective=task,
    output='array',
    chunk_size=10,
    random_weights=False
)
dg = None
eval_at = None
eval_group = None
model_factory = lgb.DaskLGBMRegressor
eval_metrics = ['rmse']

for eval_size in eval_sizes:
    _, _, _, dX_e, dy_e, dw_e = _create_data(
        n_samples=max(10, int(n_samples * eval_size)),
        n_features=n_features,
        objective=task,
        output='array',
        chunk_size=10,
        random_weights=False
    )

    eval_set.append((dX_e, dy_e))
    eval_sample_weight.append(dw_e)

full_trees = 100
params = {
    "random_state": 42,
    "n_estimators": full_trees,
    "num_leaves": 31,
    "first_metric_only": True
}

dask_model = model_factory(
    client=client,
    **params
)

dask_model = dask_model.fit(
    dX,
    dy,
    group=dg,
    eval_set=eval_set,
    eval_names=eval_names,
    eval_sample_weight=eval_sample_weight,
    eval_group=eval_group,
    eval_metric=eval_metrics,
    early_stopping_rounds=early_stopping_rounds,
    eval_at=eval_at,
    verbose=True
)
fitted_trees = dask_model.booster_.num_trees()
assert fitted_trees < full_trees
assert dask_model.best_iteration_ < full_trees

# be sure that model still produces decent output.
p1 = dask_model.predict(dX)
p1_r2 = _r2_score(dy, p1)
msg = f'r2 score of predictions with actuals was <= 0.8 ({p1_r2})'
assert p1_r2 > 0.8, msg

# check that evals_result contains expected eval_set names when provided.
n_rounds_tried = dask_model.best_iteration_ + early_stopping_rounds
evals_result = dask_model.evals_result_
assert len(evals_result) == n_eval_sets
evals_result_names = list(evals_result.keys())
if eval_names:
    assert all(x in eval_names for x in evals_result_names)

# check that evals_result names default to "training" or "valid_xx" without eval_names.
for evals_result_name in evals_result_names:
    if not eval_names:
        assert evals_result_name.startswith('training') or evals_result_name.startswith('valid')

    # check that eval_metric(s) are contained in evals_result dicts.
    for i, metric in enumerate(eval_metrics):

        assert metric in evals_result[evals_result_name]

        # len of each eval_metric should be number of fitted trees + early_stopping_rounds.
        assert len(evals_result[evals_result_name][metric]) == n_rounds_tried

        # stopping decision should have been made based on the best score of the first of eval_metrics.
        if i == 0:
            best_score = dask_model.best_score_[evals_result_name][metric]
            best_iter_zero_indexed = dask_model.best_iteration_ - 1

            assert_eq(best_score, min(evals_result[evals_result_name][metric]), atol=0.03)
            assert abs(best_iter_zero_indexed - np.argmin(evals_result[evals_result_name][metric])) \
                   <= early_stopping_rounds

```

</details>

**A Theory**

I have a theory about what's going on. I think that it might be the case that once early stopping has been detected, workers are shutting down and not carefully coordinating the shutdown. That could create a race condition where the order that they shut down matters. I suspect this because you were getting `socket recv error: 104`.

That comes from https://github.com/microsoft/LightGBM/blob/e5c3f7e755659c120b94c109b515569fa6765b9c/src/network/socket_wrapper.hpp#L286, and the specific integer value comes from https://github.com/microsoft/LightGBM/blob/e5c3f7e755659c120b94c109b515569fa6765b9c/src/network/socket_wrapper.hpp#L168. I saw answers on Stack Overflow saying that 104 means connection reset by peer, and based on https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/errno.h.html I can confirm that's true. You could run the code below on http://cpp.sh/ (or locally with the C++ compiler of your choice) to confirm.

**UPDATE (3/13): found a better list of all error codes at https://www-numi.fnal.gov/offline_software/srt_public_context/WebDocs/Errors/unix_system_errors.html**

```cpp
#include <iostream>
#include <cmath>
#include <cerrno>
#include <cstring>
#include <clocale>
#include <cstdio>
 
int main()
{
    std::cout << "message: " << std::strerror(ECONNRESET) << '\n';
    printf("error code: %d\n", ECONNRESET);
}
```

> message: Connection reset by peer
error code: 104

I think this type of problem could explain all three behaviors you're seeing:

* sometimes it works
* sometimes training fails with this "connection reset by peer"
* sometimes it hangs forever

We could try to figure this out by looking at how [mmlspark](https://github.com/Azure/mmlspark) handles early stopping, and by `@`-ing in other maintainers here. We could also try to see if the problem shows up in non-Dask distributed training, like with the LightGBM CLI, by picking up #3841.

HOWEVER, before we do that, I have a request.

**A Request**

Can you open a separate PR with only the non-early-stopping eval changes? Being able to use evaluation sets is valuable even without early stopping, because that allows you to get a deeper understanding of how the training process is progressing. The `test_eval_set_without_early_stopping` tests passed every time I ran them, so I'm fairly confident that they could be separated out.

```shell
pytest test_dask.py::test_eval_set_without_early_stopping
```

Doing the evaluation sets separately would also make this PR smaller and make it a bit easier for us to focus in on the source of instability uncovered by your tests.

Thank you SO SO SO much for all of your hard work on this. Sorry the PR has dragged on for a while. Now that the next release is almost ready, I can be a lot more responsive and help move this forward.

ffineis:
Hey @jameslamb no problem, seriously this was **worth** the wait. Thanks for all of your effort in actually going and reproducing the issue! The worst part about bugs is I never really know if it's just my bad code, if there really is a bug, or if I'm just going nuts.

Yeah I totally agree with you on both fronts - this is probably a deeper issue than it makes sense to tackle in this PR or even before the lightgbm.dask release, and what you're proposing about just tackling eval_sets makes sense. Totally happy make a simpler PR to get eval_sets out the door but not early stopping.

Questions about that PR:
1. Should `early_stopping_rounds` be marked as "not supported" just like "init_score" is and fail out if provided by user?
2. I'm just going to delete the early stopping code, not comment anything. Should I leave TODOs or like leave comments "insert early stopping here" or anything like that?

python-package/lightgbm/dask.py
```diff
@@ -170,12 +172,107 @@ def _train_part(
...
+            local_eval_set.append((_concat(x_e), _concat(y_e)))
+            if w_e:
+                local_eval_sample_weight.append(_concat(w_e))
+            if g_e:
+                local_eval_group.append(_concat(g_e))
+
+    else:
+        # when a worker receives no eval_set while other workers have eval data, causes LightGBMExceptions.
+        if evals_provided:
+            local_worker_address = get_worker().address
```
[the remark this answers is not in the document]
ffineis: Ha yeah I've had the same thought - "why am I waiting to do this here?" I was just too lazy to pry into the worker-parts map and go check before launching `client.submit`s. This is a good call and should be done prior to worker dispatch.

jameslamb:
> Totally happy make a simpler PR to get eval_sets out the door but not early stopping.
> 
> Questions about that PR:
> 
> 1. Should `early_stopping_rounds` be marked as "not supported" just like "init_score" is and fail out if provided by user?

Yeah I would accept a PR that adds `early_stopping_rounds` in the kwargs and raises a "not supported" error.

> 2. I'm just going to delete the early stopping code, not comment anything. Should I leave TODOs or like leave comments "insert early stopping here" or anything like that?

I have a preference for not leaving TODO comments in the code, since they can become out of date as things get refactored and since they're not as visible in the backlog as issues.

Whether you choose to update this branch or close this PR and create a new one is totally up to you! If it was me, I'd do this:

1. branch off of `dask/early_stopping` to like `dask/eval_sets`
2. delete all the early stopping code from `dask/eval_sets`
3. merge `master` until `dask/eval_sets`
4. close this PR
5. open a new PR from `dask/eval_sets`

I would do that if it was me so you could always come back to this PR and easily look at the diff to see where early stopping changes fit in. But up to you!

ffineis:
Awesome, thanks for these tips, on it

ffineis:
Hey I'm currently fixing merge conflicts and noticed #3950, so I'm going to address eval_init_score in the upcoming PR for eval_sets as well

jameslamb:
> Hey I'm currently fixing merge conflicts and noticed #3950, so I'm going to address eval_init_score in the upcoming PR for eval_sets as well

sounds good!

ffineis:
Hey I just opened up #4101 so I'm going to close this out! Thanks for all the patience, and sorry eval_set/ES couldn't make it into the first lightgbm.dask release!

PR #3955, opened 2021-02-14
[R-package] fix duplicated return values in docs

philip-khor:
Addresses #3739


jameslamb (approved):
thanks so much for the contribution! Looks good

Issue #3958, opened 2021-02-14
Apply isort to python_package and corresponding tests


zyxue:
## Summary

Apply [isort](https://github.com/PyCQA/isort) to all python files, including tests

## Motivation

Consistent imports order improves readability

## Description

Per discussion at https://github.com/microsoft/LightGBM/pull/3957#issuecomment-778792279, isort will be applied to the following files, which are selected with

```
fd --exclude external_libs -F '.py' | sort
```

one PR at a time for ease of review:

- [x] docs/conf.py
- [x] examples/python-guide/advanced_example.py
- [x] examples/python-guide/logistic_regression.py
- [x] examples/python-guide/plot_example.py
- [x] examples/python-guide/simple_example.py
- [x] examples/python-guide/sklearn_example.py
- [x] helpers/check_dynamic_dependencies.py
- [x] helpers/parameter_generator.py
- [x] python-package/lightgbm/\_\_init\_\_.py
- [x] python-package/lightgbm/basic.py
- [x] python-package/lightgbm/callback.py
- [x] python-package/lightgbm/compat.py
- [x] python-package/lightgbm/dask.py
- [x] python-package/lightgbm/engine.py
- [x] python-package/lightgbm/libpath.py
- [x] python-package/lightgbm/plotting.py
- [x] python-package/lightgbm/sklearn.py
- [x] python-package/setup.py
- [x] tests/c_api_test/test_.py
- [x] tests/cpp_test/test.py
- [x] tests/python_package_test/\_\_init\_\_.py
- [x] tests/python_package_test/test_basic.py
- [x] tests/python_package_test/test_consistency.py
- [x] tests/python_package_test/test_dask.py
- [x] tests/python_package_test/test_dual.py
- [x] tests/python_package_test/test_engine.py
- [x] tests/python_package_test/test_plotting.py
- [ ] tests/python_package_test/test_sklearn.py
- [x] tests/python_package_test/test_utilities.py
- [x] tests/python_package_test/utils.py

StrikerRUS:
@zyxue Thank you very much for your great contribution and original feature request!

PR #3959, opened 2021-02-14
[ci][python] apply isort to docs/conf.py (#3958)

zyxue:
see https://github.com/microsoft/LightGBM/issues/3958


StrikerRUS (approved):
Thank you very much!
LGTM!

PR #3969, opened 2021-02-14
[ci][python] apply isort to python-package/lightgbm/dask.py #3958

zyxue:
see #3958


jameslamb (approved):
looks ok to me, thanks

PR #3983, opened 2021-02-14
[ci][python] apply isort to tests/python_package_test/test_sklearn.py #3958

zyxue:
see #3958

### DIFF
--- a/tests/python_package_test/test_sklearn.py
+++ b/tests/python_package_test/test_sklearn.py
@@ -1,30 +1,35 @@
 # coding: utf-8
 import itertools
-import joblib
 import math
 import os
 
-import lightgbm as lgb
+import joblib
 import numpy as np
 import pytest
 from pkg_resources import parse_version
 from sklearn import __version__ as sk_version
 from sklearn.base import clone
 from sklearn.datasets import load_svmlight_file, make_multilabel_classification
-from sklearn.utils.estimator_checks import check_parameters_default_constructible
 from sklearn.metrics import log_loss, mean_squared_error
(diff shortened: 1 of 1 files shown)


tests/python_package_test/test_sklearn.py
```diff
@@ -1,30 +1,35 @@
...
 from sklearn.metrics import log_loss, mean_squared_error
-from sklearn.model_selection import GridSearchCV, RandomizedSearchCV, train_test_split
-from sklearn.multioutput import (MultiOutputClassifier, ClassifierChain, MultiOutputRegressor,
-                                 RegressorChain)
+from sklearn.model_selection import (GridSearchCV, RandomizedSearchCV,
+                                     train_test_split)
+from sklearn.multioutput import (ClassifierChain, MultiOutputClassifier,
+                                 MultiOutputRegressor, RegressorChain)
+from sklearn.utils.estimator_checks import \
+    check_parameters_default_constructible
```
StrikerRUS, line 19: We don't use controversial rule of 80 char line length limit in LightGBM.
```suggestion
from sklearn.utils.estimator_checks import check_parameters_default_constructible
```

StrikerRUS (changes requested):

tests/python_package_test/test_sklearn.py
StrikerRUS: If this change was made by `isort` and not by `black`, we can mark it `# isort:skip`.

StrikerRUS: Never mind! I created #3990.

PR #3985, opened 2021-02-15
[dask] reuse len(parts) as n_parts

ffineis:
Address [comment](https://github.com/microsoft/LightGBM/pull/3952#discussion_r575713223) in #3952 to reuse `len(parts)` as `n_parts`. cc @jameslamb


jameslamb (approved):
thanks very much!

PR #3987, opened 2021-02-16
Fix evalution of linear trees with a single leaf.

mjmckp:
Note that trees without linear models at the leaf always handle `num_leaves = 1` as a special case and directly output the leaf value.  Linear trees were missing this special case handling, and hence would have the following issues:
 * Calling `Tree::Predict` or `Tree::PredictByMap` would cause an access violation exception attempting to access the first value of the empty `split_feature_` array in `GetLeaf`.
 * `PredictionFunLinear` would either cause an access violation or go into an infinite loop when attempting to do the equivalent of `GetLeaf`.

Note also that `PredictionFun` does not need the same changes as `PredictionFunLinear`, since both are only called by `Tree::AddPredictionToScore`, which has a special case for `(!is_linear_ && num_leaves_ <= 1)` that precludes calling `PredictionFun`.


guolinke:
cc @btrotta

btrotta:
@mjmckp Thanks for finding and fixing this! I reproduced the error you describe with the following script:
```
import lightgbm as lgb

X_train, y_train = load_breast_cancer(return_X_y=True)
params = {
    "objective": "binary",
    "linear_tree": True,
    "min_sum_hessian": 5000
}
bst = lgb.train(params, train_data, num_boost_round=5, valid_sets={train_data}, valid_names='train')
bst.predict(X_train)
```

Your fix looks good to me!

btrotta (approved):

jameslamb:
Thanks for the fix @mjmckp !

PR #3989, opened 2021-02-16
WIP: [R-package] Add support for specifying training indices in lgb.cv()

### LINKED ISSUE #3924: [R-package] add support for specifying training indices in lgb.cv()
## Summary

The addition of a `train_folds` argument in `lgb.cv()` would allow for more fine-grained folds generation that is useful in some scenarios, such as time series forecasting (just like xgboost R package does).

## Motivation

The R function `lgb.cv()` currently has got an argument that allows you to specify manual folds through the `folds` argument. This argument expects a `list` of indices that should go to the test set for each fold, and all the other indices will go to the train set.

However, in some types of datasets and tasks (such as in time series), you may actually want to have different folds where some indices are just not used for that specific fold, neither in the train or test sets (just to avoid leaking information from the future).

## Description

The Xgboost R package included this feature a while back, which essentially consists of adding one more argument to the `cv()` function called `train_folds`. This way, `train_folds`, if specified, makes sure that only those indices will go to the train set in each fold. If it is not specified, the train indices will just be the opposite of the ones in the `folds` argument, just like `lgb.cv()` works right now.

## References

Please see the `train_folds` argument in `xgb.cv()` [here](https://www.rdocumentation.org/packages/xgboost/versions/1.3.2.1/topics/xgb.cv), and the relevant code in xgb can be found [here](https://github.com/dmlc/xgboost/blob/master/R-package/R/xgb.cv.R#L196)

Thank you!

julioasotodv:
fixes #3924

### DIFF
--- a/R-package/R/lgb.cv.R
+++ b/R-package/R/lgb.cv.R
@@ -306,7 +306,7 @@ lgb.cv <- function(params = list()
         test_indices <- folds[[k]]
       }
       
-      # Generate train_indices from either the train_folds argument 
+      # Generate train_indices from either the train_folds argument
       # or as the opposite of (test)folds argument:
       if (!is.null(train_folds)) {
         train_indices <- train_folds[[k]]
--- a/R-package/R/lgb.cv.R
+++ b/R-package/R/lgb.cv.R
@@ -305,7 +305,7 @@ lgb.cv <- function(params = list()
       } else {
         test_indices <- folds[[k]]
       }
-      
+
       # Generate train_indices from either the train_folds argument
       # or as the opposite of (test)folds argument:
       if (!is.null(train_folds)) {
--- a/R-package/R/lgb.cv.R
+++ b/R-package/R/lgb.cv.R
@@ -32,7 +32,7 @@ CVBooster <- R6::R6Class(
 #' @param folds \code{list} provides a possibility to use a list of pre-defined CV folds
 #'              (each element must be a vector of test fold's indices). When folds are supplied,
 #'              the \code{nfold} and \code{stratified} parameters are ignored.
-#' @param train_folds \code{list} specifying which indicies to use for training. If \code{NULL}
+#' @param train_folds \code{list} specifying which indices to use for training. If \code{NULL}
 #'                    (the default) all indices not specified in \code{folds} will be used for training.
 #' @param colnames feature names, if not null, will use this to overwrite the names in dataset
 #' @param categorical_feature categorical features. This can either be a character vector of feature
--- a/R-package/R/lgb.cv.R
+++ b/R-package/R/lgb.cv.R
@@ -85,13 +85,13 @@ lgb.cv <- function(params = list()
                    , showsd = TRUE
                    , stratified = TRUE
                    , folds = NULL
-                   , train_folds = NULL
                    , init_model = NULL
                    , colnames = NULL
                    , categorical_feature = NULL
                    , early_stopping_rounds = NULL
                    , callbacks = list()
                    , reset_data = FALSE
+                   , train_folds = NULL
                    , ...
                    ) {
 
--- a/R-package/R/lgb.cv.R
+++ b/R-package/R/lgb.cv.R
@@ -32,6 +32,8 @@ CVBooster <- R6::R6Class(
 #' @param folds \code{list} provides a possibility to use a list of pre-defined CV folds
 #'              (each element must be a vector of test fold's indices). When folds are supplied,
 #'              the \code{nfold} and \code{stratified} parameters are ignored.
+#' @param train_folds \code{list} specifying which indicies to use for training. If \code{NULL}
+#'                    (the default) all indices not specified in \code{folds} will be used for training.
 #' @param colnames feature names, if not null, will use this to overwrite the names in dataset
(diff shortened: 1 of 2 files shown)


julioasotodv:
No idea why Azure pipelines are unhappy. It looks like it was an issue with the instances while building the containers... Shall we force rebuild?

StrikerRUS:
@julioasotodv 
> It looks like it was an issue with the instances while building the containers... Shall we force rebuild?

Thanks, done!

R-package/R/lgb.cv.R
```diff
@@ -83,6 +85,7 @@ lgb.cv <- function(params = list()
                    , showsd = TRUE
                    , stratified = TRUE
                    , folds = NULL
+                   , train_folds = NULL
```
jameslamb, line 88: Can you please move this down to after `reset_data`? Adding new arguments in the middle of the signature is a breaking change for anyone using positional arguments.

R-package/R/lgb.cv.R
```diff
@@ -32,6 +32,8 @@ CVBooster <- R6::R6Class(
 #' @param folds \code{list} provides a possibility to use a list of pre-defined CV folds
 #'              (each element must be a vector of test fold's indices). When folds are supplied,
 #'              the \code{nfold} and \code{stratified} parameters are ignored.
+#' @param train_folds \code{list} specifying which indicies to use for training. If \code{NULL}
```
jameslamb, line 35:
```suggestion
#' @param train_folds \code{list} specifying which indices to use for training. If \code{NULL}
```

jameslamb (changes requested):
Thanks for taking the time to contribute! This is a really nice addition.

Could you please add a unit test in the `lgb.cv()` section, checking that this behavior works as expected? It would be great if you can set up a small time-series cross validation example as the test. https://github.com/microsoft/LightGBM/blob/master/R-package/tests/testthat/test_basic.R#L262

jameslamb:
I changed the PR's description to say "fixes" instead of "as seen in". That way when we merge this, the issue will be closed automatically.

R-package/R/lgb.cv.R
julioasotodv: Yes, you are right. Didn't think about that.

julioasotodv: However, should I keep the docs in the current order (this is, right after `folds`?) This way we don't break compatibility but feels more natural for users

julioasotodv: Indeed, it does fail. So the new argument should go at the end in both the docs and the function signature.

julioasotodv:
> 
> 
> Thanks for taking the time to contribute! This is a really nice addition.
> 
> Could you please add a unit test in the `lgb.cv()` section, checking that this behavior works as expected? It would be great if you can set up a small time-series cross validation example as the test. https://github.com/microsoft/LightGBM/blob/master/R-package/tests/testthat/test_basic.R#L262

Sure! I will try to come up with a meaningful test.

R-package/R/lgb.cv.R
jameslamb: yep, thanks for investigating!

jameslamb:
@julioasotodv are you still interested in contributing this? I'm available to help if you have any questions.

julioasotodv:
@jameslamb Indeed! Sorry had a couple of busy weeks...

Yes, I was thinking on how to make a test for this, as I don't see it straightforward... Since `lgb.cv()` basically returns the results list, I could only think about seeding everything (model through params and nfold) and check that result changes with and without `train_folds`. 

But perhaps there is a better way I did not think of...

jameslamb:
No problem, and no rush at all!

Let me know if it's too simplistic, but what about this?

1. Construct a dataset for a regression tasks where the relationship between the features and target should be easy for LightGBM to figure out. Like maybe where you have a few features that are all `some_constant * y + some_noise()`
2. Construct a second, much larger dataset with the same number of features that is all random noise
3. Set `folds` to only slices of the "real" data (so you aren't evaluating on any of the noise data)
4. Run `lgb.cv()` twice, once without `train_folds` and once with `train_folds` set to only contain indices from the "real data". Make `num_iterations` and `num_leaves` fairly small (like `num_iterations = 5L, num_leaves = 5L`).
5. Compare model performance between the two cases, evaluated only against the "real" data. The model trained on only the "real data" (where you specified `train_folds`) should have much better performance than the one trained on mostly noise.

jameslamb:
gently pinging @julioasotodv , is there anything I can do to help you with this?

jameslamb:
For now, I'm going to close this PR due to lack of response. @julioasotodv, thank you again for your interest in LightGBM and a great issue write-up in #3924! If you have the time and interest to contribute this in the future, we'd welcome a new pull request.

I think providing better support for customizing of the cross-validation process (like for cases mentioned in #3924 where you want to do time-based splits) is a valuable contribution to the R package, but not one that will be done by maintainers before the next release (#4310).

I'm going to lock discussion on this PR for now to focus all discussion of this feature back on #3924. Anyone who is interested in contributing this feature (or wants to ask maintainers to give it higher priority and implement it) is welcome to comment on #3924.

PR #3993, opened 2021-02-16
[tests][dask] make find-open-port test more reliable

jameslamb:
Working on a proposal to address https://github.com/microsoft/LightGBM/pull/3823#pullrequestreview-582002723, I found that sometimes the "find an open port" test in `test_dask.py` failed with an error.

> assert new_port == 12401
> E assert 12404 == 12401

I think that test is too strict. Right now it checks "if port 12400 is unavailable, `_find_open_port()` returns 12401". If 12401 is busy for some unrelated reason, this test fails.

This PR proposes loosening the condition a bit, to reduce the risk of test failures like that. It isn't important that that method finds 12401 specifically. Just some port that is `>` the port number you passed in.

### DIFF
--- a/tests/python_package_test/test_dask.py
+++ b/tests/python_package_test/test_dask.py
@@ -885,27 +885,27 @@ def test_model_and_local_version_are_picklable_whether_or_not_client_set_explici
                     assert_eq(preds_orig_local, preds_loaded_model_local)
 
 
-def test_find_open_port_works():
+def test_find_open_port_works(listen_port):
     worker_ip = '127.0.0.1'
     with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
-        s.bind((worker_ip, 12400))
+        s.bind((worker_ip, listen_port))
         new_port = lgb.dask._find_open_port(
             worker_ip=worker_ip,
-            local_listen_port=12400,
+            local_listen_port=listen_port,
             ports_to_skip=set()
         )
-        assert new_port >= 12401
+        assert new_port > listen_port
 
     with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s_1:
-        s_1.bind((worker_ip, 12400))
+        s_1.bind((worker_ip, listen_port))
         with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s_2:
-            s_2.bind((worker_ip, 12401))
+            s_2.bind((worker_ip, listen_port + 1))
             new_port = lgb.dask._find_open_port(
                 worker_ip=worker_ip,
-                local_listen_port=12400,
+                local_listen_port=listen_port,
                 ports_to_skip=set()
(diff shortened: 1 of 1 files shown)


tests/python_package_test/test_dask.py
```diff
@@ -885,27 +885,27 @@ def test_model_and_local_version_are_picklable_whether_or_not_client_set_explici
                     assert_eq(preds_orig_local, preds_loaded_model_local)
 
 
-def test_find_open_port_works():
+def test_find_open_port_works(listen_port):
     worker_ip = '127.0.0.1'
     with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
-        s.bind((worker_ip, 12400))
+        s.bind((worker_ip, listen_port))
```
StrikerRUS, line 891: > I think that test is too strict.

I think now test is too soft 😃 . What about checking some meaningful intervals? Something like
```
worker_ip = '127.0.0.1'
port = listen_port
    s.bind((worker_ip, port))
    ...
assert port < new_port < port + 10
```

jameslamb: ok I can change it to `port < new_port < port + 1000`, since that is the range that the function covers: https://github.com/microsoft/LightGBM/blob/5321fef67b5cc4711dd7b2de985d5fe2b76d90cf/python-package/lightgbm/dask.py#L72

tests/python_package_test/test_dask.py
```diff
@@ -885,27 +885,27 @@ def test_model_and_local_version_are_picklable_whether_or_not_client_set_explici
...
-            s_2.bind((worker_ip, 12401))
+            s_2.bind((worker_ip, listen_port + 1))
             new_port = lgb.dask._find_open_port(
                 worker_ip=worker_ip,
-                local_listen_port=12400,
+                local_listen_port=listen_port,
                 ports_to_skip=set()
             )
-            assert new_port == 12402
+            assert new_port > listen_port + 1
```
jameslamb, line 908:
```suggestion
            assert listen_port < new_port < listen_port + 1000
```

tests/python_package_test/test_dask.py
```diff
@@ -885,27 +885,27 @@ def test_model_and_local_version_are_picklable_whether_or_not_client_set_explici
...
-        s.bind((worker_ip, 12400))
+        s.bind((worker_ip, listen_port))
         new_port = lgb.dask._find_open_port(
             worker_ip=worker_ip,
-            local_listen_port=12400,
+            local_listen_port=listen_port,
             ports_to_skip=set()
         )
-        assert new_port == 12401
+        assert new_port > listen_port
```
jameslamb, line 897:
```suggestion
        assert listen_port < new_port < listen_port + 1000
```

StrikerRUS (approved):
Thanks!

Issue #3847, opened 2021-01-25
[dask] add cv() function


jameslamb:
## Summary

As of this writing, `lightgbm.dask` only supports  model classes that mimic the scikit-learn API. It should also support a function, equivalent to `lightgbm.engine.cv`.

https://github.com/microsoft/LightGBM/blob/36322ceeae1912f15ebb005c77cd9bf416087e78/python-package/lightgbm/engine.py#L391

Because `cv()` expects to be given a LightGBM `Dataset` object, this also implies creating a new class `lightgbm.dask.DaskDataset`. `cv()` should take in `train_set` as a `lightgbm.dask.DaskDataset`, and should return a regular LightGBM Booster.

## Motivation

Having a functional interface would make the transition from non-Dask to Dask easier for users who are already using `lightgbm.engine.cv)`.

## References

See the `DaskDMatrix` in `xgboost.dask` for some inspiration on how `DaskDataset` might be implemented.

https://github.com/dmlc/xgboost/blob/a275f4026728ed14fbc70da142ef7a4a1d3de04d/python-package/xgboost/dask.py#L181-L186

`DaskDataset` may be implemented outside of this feature, to support #3846

jameslamb:
Closing this in favor of putting it in #2302 with other feature requests. Comment below if you'd like to work on this, and it can be re-opened.

PR #3994, opened 2021-02-17
[dask] allow tight control over ports

jameslamb:
In https://github.com/microsoft/LightGBM/pull/3823, @jmoralez proposed a change for the Dask module that would speed up training by taking a bit of setup overhead that is `O(n_workers)` and make it `O(1)`. That discussion led to a conversation about how to give tight control over the ports used by LightGBM, for users who are using a cluster of machines whose communication is limited by firewall rules.

I promised to go do some research in LightGBM's source code and come back with a proposal. This PR is that proposal.

## Proposed Changes

This PR proposes that `lightgbm.dask` should handle ports as follows.

### If no network params are given in ``params`` (the default)

```python
model = DaskLGBMRegressor()
model.fit(data, labels)
```

LightGBM chooses ports randomly. As of this PR, it will use the current approach on `master` (searching 1000 ports starting with the default value of `local_listen_port`), but after this PR we should replace that with @jmoralez 's proposal in #3823.

### If ``machines`` is provided in ``params``

```python
machines = "10.0.0.1:15000,10.0.0.1:15001,10.0.1.1:15000,10.0.1.2:15005"
model = DaskLGBMRegressor(
    machines=machines
)
model.fit(data, labels)
```

LightGBM respects ``machines`` and does not do any searching. This gives people who are in a constrained environment total control over the ports used, without needing to invent a new parameter specific to the Dask interface.

This can work for cases where you run multiple Dask worker processes on the same host. For example, for a `FargateCluster` with `n_workers=4, nprocs=2`, you might use ``machines`` like this:

```text
10.0.0.1:15000,10.0.0.1:15001,10.0.1.1:15000,10.0.1.2:15005
```

In a follow-up PR after this, to make this easier, I'd propose adding a function `dask_collection_to_machines_param()`, which takes in your training data and a list of ports that you've allowed in your firewall settings, and returns a `machines` string you can put into `params`.

```python
def dask_collection_to_machines_param(
    data: Union[dask.DataFrame, dask.Array],
    allowed_ports: Set[int]
) -> str:
```

### If ``local_listen_port`` is provided in ``params``

```python
model = DaskLGBMRegressor(
    local_listen_port=16000
)
model.fit(data, labels)
```

Create ``machines`` by looking at which worker addresses have a piece of the training data, and assume that each of them will use the same port (``local_listen_port``). This will only work in the case where you have 1 Dask worker process per physical host.

## How this PR improves `lightgbm.dask`

This gives an official answer to the question "how I should I used distributed LightGBM training with Dask if my Dask cluster is constrained by firewall rules?", while also making it possible to take advantage of the speedups from #3823 .

It does this in a way that only uses existing LightGBM network parameters, and which doesn't require any notes in documentation that say "hey this parameter has a slightly different meaning in Dask than everywhere else".

## How LightGBM's network setup works

This is optional background that will help explain the proposal I'm making. Everything below refers only to the socket-based build of LightGBM.

<details><summary>notes on how the code in src/network works</summary>

Each worker runs [`Network::Init()`](https://github.com/microsoft/LightGBM/blob/75b9b0d3c8081cbe8ddedaac08c4213df55efe29/src/network/network.cpp#L30) initialize a LightGBM network. The following happens when that method is called.

1. "What is my ID (rank) and what other workers are in this network?" --> Create a [`Linkers`](https://github.com/microsoft/LightGBM/blob/75b9b0d3c8081cbe8ddedaac08c4213df55efe29/src/network/linkers_socket.cpp#L24)
    1. "My ID will be an integer corresponding to my place in the ``machines`` list" --> https://github.com/microsoft/LightGBM/blob/75b9b0d3c8081cbe8ddedaac08c4213df55efe29/src/network/linkers_socket.cpp#L42
    2. "I can open a TCP listener socket to accept connections from other workers" --> https://github.com/microsoft/LightGBM/blob/75b9b0d3c8081cbe8ddedaac08c4213df55efe29/src/network/linkers_socket.cpp#L51-L52
    3. "I'll set up some data structures to store information about the other workers in the network" --> https://github.com/microsoft/LightGBM/blob/75b9b0d3c8081cbe8ddedaac08c4213df55efe29/src/network/linkers_socket.cpp#L54-L60
    4. "I need to establish a socket connection with each other worker in the network" --> call [`Linkers::Construct()`](https://github.com/microsoft/LightGBM/blob/75b9b0d3c8081cbe8ddedaac08c4213df55efe29/src/network/linkers_socket.cpp#L166) to set up these connections
        1. "Since these are two-way communications, I'll just initiate communications with all workers that have a `rank` greater than mine. I know from ``machines`` which IP addresses to look for and what ``local_listen_port`` they'll be listening on. The workers with `rank` less than me will initiate communications with me. Each time I communicate successfully, I'll save that new TCP socket so I can use it during training to talk to that specific worker" --> https://github.com/microsoft/LightGBM/blob/75b9b0d3c8081cbe8ddedaac08c4213df55efe29/src/network/linkers_socket.cpp#L196-L216
     5. "Ok, now that I have sockets to talk to all the other workers, I can destroy the listener socket. I'm not accepting any more new connections" --> https://github.com/microsoft/LightGBM/blob/75b9b0d3c8081cbe8ddedaac08c4213df55efe29/src/network/linkers_socket.cpp#L65-L66
2. "Now that I know how to talk to other workers, I should also set up some space in memory to store data I want to send to them and data I receive from them" --> https://github.com/microsoft/LightGBM/blob/75b9b0d3c8081cbe8ddedaac08c4213df55efe29/src/network/network.cpp#L37-L40

You can learn more about this from https://lightgbm.readthedocs.io/en/latest/Parallel-Learning-Guide.html#preparation, https://lightgbm.readthedocs.io/en/latest/Parallel-Learning-Guide.html#id1, and https://lightgbm.readthedocs.io/en/latest/Parameters.html#network-parameters.

</details>

## Notes for Reviewers

I'll devote a section to this topic in the documentation introduced for #3814 .

The one Dask unit test that's failing will pass if #3993 is accepted and merged.


jameslamb:
I thought I left a comment on here but I guess I navigated away before sending it haha.

@jmoralez @ffineis if you have time and interest, I'd really appreciate your feedback on this PR.

tests/python_package_test/test_dask.py
```diff
@@ -173,6 +173,14 @@ def _accuracy_score(dy_true, dy_pred):
     return da.average(dy_true == dy_pred).compute()
 
 
+def _find_random_open_port() -> int:
```
jameslamb, line 176: borrowing this from #3823 . Could be imported from `lightgbm.dask` in that PR.

jameslamb:
I tested this on a `FargateCluster` on AWS tonight using `dask-cloudprovider`: https://github.com/jameslamb/lightgbm-dask-testing/blob/main/notebooks/aws.ipynb

I tried all three possible configurations described above, and all three trainings succeeded. So I'm confident this will work on either `LocalCluster` (what we use in tests) or a true distributed cluster.

python-package/lightgbm/dask.py
```diff
@@ -238,14 +266,42 @@ def _train(
...
+    Note
+    ----
+
+    This method handles setting up the following network parameters based on information
+    about the Dask cluster referenced by ``client``.
+
+    * ``local_listen_port``: port that each LightGBM worker opens a listening socket on,
+        to accept connections from other workers. This can be differ from LightGBM worker
+        to LightGBM worker, but does not have to.
+    * ``machines``: a list of all machines in the cluster, plus a port to communicate
```
ffineis, line 279: This totally just my opinion, but `"machines": a list of all machines in the cluster, plus a port to communicate..."` is kind of circular - the second use of "machines" is referring to IP addresses, but the original use of `machines` (IMO) is an IP:port combo

jameslamb: oh yeah I probably should not use the word "machines" again in the description haha, thank you

python-package/lightgbm/dask.py
```diff
@@ -312,14 +367,60 @@ def _train(
...
+        _log_info("Using passed-in 'machines' parameter")
+        worker_address_to_port = _machines_to_worker_map(
+            machines=machines,
+            worker_addresses=worker_addresses
+        )
+    else:
+        if listen_port_in_params:
+            _log_info("Using passed-in 'local_listen_port' for all workers")
+            unique_hosts = set([urlparse(a).hostname for a in worker_addresses])
+            if len(unique_hosts) < len(worker_addresses):
```
ffineis, line 398: Ah slick, very slick

ffineis:
> I thought I left a comment on here but I guess I navigated away before sending it haha.
> 
> @jmoralez @ffineis if you have time and interest, I'd really appreciate your feedback on this PR.

Hey I think these three options make sense and that they're easy to understand! Nice work! Really, I don't have much to add.

python-package/lightgbm/dask.py
jameslamb: ok updated in https://github.com/microsoft/LightGBM/pull/3994/commits/25462ea00d39cc034fac030415800d615730ae0e, thanks for catching this

StrikerRUS:
@jameslamb I haven't looked at the diff yet, just read the detailed explanation you've provided.

I think I found one inconsistency.

> If no network params are given in `params` (the default)
   LightGBM chooses ports randomly.

> It does this in a way that only uses existing LightGBM network parameters, and which doesn't require any notes in documentation that say "hey this parameter has a slightly different meaning in Dask than everywhere else".

Default value of the `local_listen_port` is [`12400`](https://lightgbm.readthedocs.io/en/latest/Parameters.html#local_listen_port) and not "any random value starting from 12400 (or even simply 'any random value' after #3823)". And some users which are fine with **default** port and are under firewall constraints don't expect that ports will be **random**, they expect all ports on all machines will be exactly `12400` (default value). So, I still think that we need some clarification of this inconsistency in Dask docs.
Refer to https://github.com/microsoft/LightGBM/pull/3823#pullrequestreview-582342932.

jameslamb:
> @jameslamb I haven't looked at the diff yet, just read the detailed explanation you've provided.
> 
> I think I found one inconsistency.
> 
> > If no network params are given in `params` (the default)
> > LightGBM chooses ports randomly.
> 
> > It does this in a way that only uses existing LightGBM network parameters, and which doesn't require any notes in documentation that say "hey this parameter has a slightly different meaning in Dask than everywhere else".
> 
> Default value of the `local_listen_port` is [`12400`](https://lightgbm.readthedocs.io/en/latest/Parameters.html#local_listen_port) and not "any random value starting from 12400 (or even simply 'any random value' after #3823)". And some users which are fine with **default** port and are under firewall constraints don't expect that ports will be **random**, they expect all ports on all machines will be exactly `12400` (default value). So, I still think that we need some clarification of this inconsistency in Dask docs.
> Refer to [#3823 (review)](https://github.com/microsoft/LightGBM/pull/3823#pullrequestreview-582342932).

ok, fair enough. I just added a warning in https://github.com/microsoft/LightGBM/pull/3994/commits/deeab63f6e9841c5be643cf9cee07454e2ff5e2b.

This isn't a LightGBM branch so you won't be able to see RTD builds, but this worked fine for me running `cd docs/ && make clean html` locally.

image

StrikerRUS:
@jameslamb Thank you! I just thought about it a little bit more and come up to that it probably will be easier to just write `default = 12400 (random for Dask-package)` because right now users must read docs very carefully in two places and we must duplicate this warning in multiple (at least 3) places after implementing #3846 and #3847.
Refer to 
https://github.com/microsoft/LightGBM/blob/7880b79fde36873b95839c0029ea9d649397da51/include/LightGBM/config.h#L98
for how to set custom default value in params docs.

jameslamb:
> @jameslamb Thank you! I just thought about it a little bit more and come up to that it probably will be easier to just write `default = 12400 (random for Dask-package)` because right now users must read docs very carefully in two places and we must duplicate this warning in multiple (at least 3) places after implementing #3846 and #3847.
> Refer to
> https://github.com/microsoft/LightGBM/blob/7880b79fde36873b95839c0029ea9d649397da51/include/LightGBM/config.h#L98
> 
> 
> for how to set custom default value in params docs.

ok sure, no problem. Done in https://github.com/microsoft/LightGBM/pull/3994/commits/0c81f6012c794d657b4ef4022e102c322621003c

image

python-package/lightgbm/dask.py
```diff
@@ -238,14 +266,44 @@ def _train(
...
         Returns fitted underlying model.
+
+    Note
+    ----
+
+    This method handles setting up the following network parameters based on information
+    about the Dask cluster referenced by ``client``.
+
+    * ``local_listen_port``: port that each LightGBM worker opens a listening socket on,
+            to accept connections from other workers. This can be differ from LightGBM worker
```
jameslamb, line 277:
```suggestion
            to accept connections from other workers. This can differ from LightGBM worker
```

tests/python_package_test/test_dask.py
```diff
@@ -1033,6 +1030,84 @@ def collection_to_single_partition(collection):
...
+        machines=",".join([
+            "127.0.0.1:" + str(_find_random_open_port())
+            for _ in range(n_workers)
+        ]),
+    )
+    if task == 'ranking':
+        dask_model2.fit(dX, dy, group=dg)
+    else:
+        dask_model2.fit(dX, dy)
+    assert dask_model2.fitted_
```
StrikerRUS, line 1093: We should really assert that 
1. `assert dask_model2.get_params()['machines'] == machines`
2. check somehow that passed workers from `machines` were used and not any other as it is stated in the test name.

StrikerRUS (changes requested):
Great job! But I have some comments below.

python-package/lightgbm/dask.py
```diff
@@ -312,14 +369,60 @@ def _train(
...
+        )
+    else:
+        if listen_port_in_params:
+            _log_info("Using passed-in 'local_listen_port' for all workers")
+            unique_hosts = set([urlparse(a).hostname for a in worker_addresses])
+            if len(unique_hosts) < len(worker_addresses):
+                msg = (
+                    "'local_listen_port' was provided in Dask training parameters, but at least one "
+                    "machine in the cluster has multiple Dask worker processes running on it. Please omit "
+                    "'local_listen_port' or pass 'machines'."
```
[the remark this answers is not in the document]
jameslamb: Could you explain why you think we need to include a newline?

I'm concerned that in logs, it will look like an exception with only the text before the newline followed by a separate print statement.

image

tests/python_package_test/test_dask.py
jameslamb: oh good point! ok yes, I can do that.

> check somehow that passed workers from machines were used and not any other as it is stated in the test name.

I can test this by binding one of the ports mentioned in `machines`, and asserting that training fails with the "failed to bind port" error

python-package/lightgbm/dask.py
```diff
@@ -271,10 +329,9 @@ def _train(
         )
 
     # Some passed-in parameters can be removed:
-    #   * 'machines': constructed automatically from Dask worker list
     #   * 'num_machines': set automatically from Dask worker list
     #   * 'num_threads': overridden to match nthreads on each Dask process
-    for param_alias in _ConfigAliases.get('machines', 'num_machines', 'num_threads'):
+    for param_alias in _ConfigAliases.get('num_machines', 'num_threads'):
         params.pop(param_alias, None)
```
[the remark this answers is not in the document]
jameslamb: sorry, I accepted this suggestion but I think now that we should only apply it to `num_machines`, not `num_threads`.

This results in a warning that users cannot suppress.

>   /opt/conda/lib/python3.8/site-packages/lightgbm/dask.py:338: UserWarning: Parameter n_jobs will be ignored.
    _log_warning(f"Parameter {param_alias} will be ignored.")

Caused by the fact that `n_jobs` is an alias of `num_threads`

* https://github.com/microsoft/LightGBM/blob/646267d2656cbd13d86e498e96c6db95a9bae5d8/python-package/lightgbm/dask.py#L489
* https://github.com/microsoft/LightGBM/blob/646267d2656cbd13d86e498e96c6db95a9bae5d8/python-package/lightgbm/sklearn.py#L516

I believe that every warning should be something that can be changed by user code changes. Otherwise, we're just adding noise to logs that might cause people to start filtering out ALL warnings.

python-package/lightgbm/dask.py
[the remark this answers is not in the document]
StrikerRUS: OK, it is up to you. Feel free to revert new line. I personally don't like long line warnings/errors.

python-package/lightgbm/dask.py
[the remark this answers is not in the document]
StrikerRUS: I think it is important to notify users about `num_threads` as well before implementing #3714. Silently ignore parameter is more serious problem compared to unfixable warning, I believe.

jmoralez:
Sorry for being late to the party, I've had a lot of stuff at work.
> Default value of the local_listen_port is 12400 and not "any random value starting from 12400 (or even simply 'any random value' after #3823)". And some users which are fine with default port and are under firewall constraints don't expect that ports will be random, they expect all ports on all machines will be exactly 12400 (default value)

I agree with @StrikerRUS here and I really like the proposal of including the `machines` argument because I think it makes it very easy for the user to be explicit about the desired distributed training configuration. Right now I think when `machines` is specified it isn't being verified that the ports are actually open, so I think we could add something like:

```python
def _check_port_is_open(port: int) -> int:
"""Check that port is open on the machine or find a random open one if port is 0."""
    try:
              s.bind(('', port))
              port = s.getsockname()[1]
    except OSError:
        # handle port already in use
    return port
```
to `_machines_to_worker_map` and that way for `machines="ip1:12600,ip2:12700"` the function would check that `12600` is available in `ip1` and `12700` is available in `ip2`. And using random ports for training could be achieved with `machines="ip1:0,ip2:0"`. This wouldn't change LightGBM's default behavior in the dask interface and should only be documented in the `machines` argument.

jameslamb:
No problem @jmoralez , thanks for all your help so far with the Dask package.

> Right now I think when machines is specified it isn't being verified that the ports are actually open, so I think we could add something like...

I'm against adding additional code for this in the Python package. You'll already get the informative error "Binding port 12345 failed" if one of the ports you chose is not open. https://github.com/microsoft/LightGBM/blob/5005b7beb6e30229c787abecab962ce12dff58c0/src/network/linkers_socket.cpp#L131.

> And using random ports for training could be achieved with machines="ip1:0,ip2:0"

I don't want to encourage or support this pattern. If the communication between workers is not governed by strict firewall rules,  `lightgbm.dask` should "just work",  just like every library in the Dask ecosystem works without you ever needing to think about IP addresses or ports.

The "list IPs explicitly with `machines`" approach is fragile because if any of those machines don't also have a piece of the training data, training will fail. And because those values have to be recomputed each time you connect to a new cluster, or whenever a Dask worker is restarted. It's worth paying the price of that fragility in exchange for being compliant with your organization's networking constraints, but I wouldn't encourage anyone who is comfortable with getting random ports to use `machines`.

So I'm against adding a 4th option that is like "if you use `:0` in `machines`, `lightgbm.dask` will search randomly for ports". I think the 3 options explained in this pull request's description are sufficient.

python-package/lightgbm/dask.py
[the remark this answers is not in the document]
jameslamb: alright I'm not going to accept this suggestion, then.

tests/python_package_test/test_dask.py
```diff
@@ -1033,6 +1030,142 @@ def collection_to_single_partition(collection):
...
+            dask_model3.fit(dX, dy, group=dg)
+        else:
+            dask_model3.fit(dX, dy)
+
+    client.close(timeout=CLIENT_CLOSE_TIMEOUT)
+
+
+@pytest.mark.parametrize('task', tasks)
+@pytest.mark.parametrize('output', data_output)
+def test_machines_should_be_used_if_provided(task, output):
```
jameslamb, line 1117: I had to make this a separate test from `test_network_params_not_required_but_respected_if_given` because all the different context managers (`client` fixture + `with pytests.raises()` + `with socket.socket()`) were interacting in a strange way, and I was getting frequent teardown errors.

python-package/lightgbm/dask.py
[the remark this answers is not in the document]
jameslamb: I disagree in this specific case about the meaning of "ignore", since this is a parameter default and not something explicitly passed in. However, since `num_threads` isn't directly related to the purpose of this PR and since I don't want to delay this PR too long because I'd like to merge #3823 soon after it, I'll leave this warning in for now and propose another PR in a few days where we could discuss it further.

tests/python_package_test/test_dask.py
jameslamb: ok I think I've addressed both oof these in my most recent commit. I'm glad you mentioned `#1`, because it's something I didn't think about correctly. To fix that, I captured whether `machines` was provided explicitly in parameters, and if it was I had `_train()` preserve it in the fitted model object. Like this

```python
if not machines_in_params:
    for param in _ConfigAliases.get('machines'):
        model._other_params.pop(param, None)
```

tests/python_package_test/test_dask.py
jameslamb: this is also why I had to use `LocalCluster` instead of the `client` fixture for this test

python-package/lightgbm/dask.py
```diff
@@ -312,14 +373,60 @@ def _train(
...
+    if machines is not None:
+        _log_info("Using passed-in 'machines' parameter")
+        worker_address_to_port = _machines_to_worker_map(
+            machines=machines,
+            worker_addresses=worker_addresses
+        )
+    else:
+        if listen_port_in_params:
+            _log_info("Using passed-in 'local_listen_port' for all workers")
+            unique_hosts = set([urlparse(a).hostname for a in worker_addresses])
```
StrikerRUS, line 403: For efficiency.
```suggestion
            unique_hosts = set(urlparse(a).hostname for a in worker_addresses)
```

tests/python_package_test/test_dask.py
```diff
@@ -1033,6 +1030,142 @@ def collection_to_single_partition(collection):
...
+        if task == 'classification':
+            dask_model_factory = lgb.DaskLGBMClassifier
+        elif task == 'regression':
+            dask_model_factory = lgb.DaskLGBMRegressor
+
+    # rebalance data to be sure that each worker has a piece of the data
+    if output == 'array':
+        client.rebalance()
+
+    # model 1 - no network parameters givem
```
StrikerRUS, line 1062:
```suggestion
    # model 1 - no network parameters given
```

tests/python_package_test/test_dask.py
```diff
@@ -1033,6 +1030,142 @@ def collection_to_single_partition(collection):
...
+
+
+@pytest.mark.parametrize('task', tasks)
+@pytest.mark.parametrize('output', data_output)
+def test_machines_should_be_used_if_provided(task, output):
+    if task == 'ranking' and output == 'scipy_csr_matrix':
+        pytest.skip('LGBMRanker is not currently tested on sparse matrices')
+
+    with LocalCluster(n_workers=2) as cluster:
+        with Client(cluster) as client:
```
StrikerRUS, line 1122: Just letting you know that you can use one-line `with` for multiple variables to not make the code over-indented:
```
with LocalCluster(n_workers=2) as cluster, Client(cluster) as client:
```

StrikerRUS (approved):
LGTM!

tests/python_package_test/test_dask.py
jameslamb: oooo I didn't know you could do that when the second `with` references the `as` name from the first one, nice!

python-package/lightgbm/dask.py
jameslamb: thanks! I never thought through the fact that writing it like this means you have to first create a list and then construct a set from the list.

tests/python_package_test/test_dask.py
jameslamb: I just made that change in https://github.com/microsoft/LightGBM/pull/3994/commits/b3c8a2cf88142147fcb58a4dfa2c87f51d5e6eb5, thanks

PR #4004, opened 2021-02-19
[dask] use more specific method names on _DaskLGBMModel

jameslamb:
The Dask sklearn estimators in `lightgbm.dask` inherit from both their `lightgbm.sklearn` equivalent and a mixin called `_DaskLGBMModel`.

https://github.com/microsoft/LightGBM/blob/7880b79fde36873b95839c0029ea9d649397da51/python-package/lightgbm/dask.py#L525

Those equivalents from `lightgbm.sklearn` inherit from a mix of LightGBM internal classes and some mixins brought in from `sklearn`.

```python
from lightgbm.sklearn import DaskLGBMClassifier
for c in DaskLGBMClassifier.__mro__:
    print(c)

# <class 'lightgbm.dask.DaskLGBMClassifier'>
# <class 'lightgbm.sklearn.LGBMClassifier'>
# <class 'lightgbm.sklearn.LGBMModel'>
# <class 'sklearn.base.BaseEstimator'>
# <class 'sklearn.base.ClassifierMixin'>
# <class 'lightgbm.dask._DaskLGBMModel'>
# <class 'object'>
```

Since `LGBMClassifier` is the first parent class in the method resolution order ("MRO") for `DaskLGBMClassifier`, if it or any of the parent classes from `sklearn` introduce a method that conflicts with one of the method names in `_DaskLGBMModel`, the `_DaskLGBMModel` method won't be used, which could break `DaskLGBMClassifier`.

To protect against this happening, this PR proposes changing the method names on `_DaskLGBMModel` to names that are less likely to conflict with `sklearn` in the future. For example, it proposes changing `_fit()` to `_lgb_dask_fit()`, so that if an `sklearn.base.BaseEstimator._fit()` is introduced in some future version of `sklearn`, `DaskLGBMClassifier` will still use the correct method.

All of the methods changed here are internal, so this should not have any user-facing impact.

### Notes for reviewers

I also considered the pattern below for absolute certainty:

```python
class DaskLGBMClassifier(LGBMClassifier, _DaskLGBMModel)
    def fit(self, ...):
        _DaskLGBMModel._fit(self, ...)
```

But I don't understand the full implications of doing that, so I decided against it.

### DIFF
--- a/python-package/lightgbm/dask.py
+++ b/python-package/lightgbm/dask.py
@@ -465,7 +465,7 @@ class _DaskLGBMModel:
 
         return _get_dask_client(client=self.client)
 
-    def _lgb_getstate(self) -> Dict[Any, Any]:
+    def _lgb_dask_getstate(self) -> Dict[Any, Any]:
         """Remove un-picklable attributes before serialization."""
         client = self.__dict__.pop("client", None)
         self._other_params.pop("client", None)
@@ -474,7 +474,7 @@ class _DaskLGBMModel:
         self.client = client
         return out
 
-    def _fit(
+    def _lgb_dask_fit(
         self,
         model_factory: Type[LGBMModel],
         X: _DaskMatrixLike,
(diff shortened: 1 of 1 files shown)


StrikerRUS (approved):
Makes sense! Thank you!

Issue #4002, opened 2021-02-19
SHAP feature contribution for linear trees


SpeckledJim2:
## Description
When using linear_tree = TRUE and predict() with predcontrib = TRUE, the sum of the feature contributions does not equal the predicted value.

## Reproducible example
library(lightgbm)
x <- matrix(data = sample(rnorm(100L), size = 100L), ncol = 1L)
y <- 2L * x + runif(nrow(x), 0L, 0.1)

lgb_params_1 <- list(
    objective = "regression"
    , linear_tree = FALSE
    , verbose = -1L
    , metric = "mse"
    , seed = 0L
    , num_leaves = 2L
    , bagging_freq = 1L
    , subsample = 1.0
)
  
dtrain <- lgb.Dataset(data = x, label = y)
bst_lin_1 <- lgb.train(data = dtrain, nrounds = 10L, params = lgb_params_1, valids = list("train" = dtrain))

lgb_params_2 <- lgb_params_1
lgb_params_2$linear_tree <- TRUE # this is the only parameter that has changed
dtrain <- lgb.Dataset(data = x, label = y)
bst_lin_2 <- lgb.train(data = dtrain, nrounds = 10L, params = lgb_params_2, valids = list("train" = dtrain))

pred_1 <- predict(bst_lin_1, x, predcontrib = FALSE) # predict on model 1
pred_contrib_1 <- rowSums(predict(bst_lin_1, x, predcontrib = TRUE)) # predict on model 1 with feature contribs
diff_1 <- pred_1 - pred_contrib_1
sd(diff_1) # very close to zero as expected 

pred_2 <- predict(bst_lin_2, x, predcontrib = FALSE) # predict on model 2
pred_contrib_2 <- rowSums(predict(bst_lin_2, x, predcontrib = TRUE)) # predict on model 2 with feature contribs
diff_2 <- pred_2 - pred_contrib_2
sd(diff_2) # not zero - rowSums do not total to predicted values

## Environment info
R version 4.0.2 (2020-06-22)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 18.04.5 LTS

Matrix products: default
BLAS:   /usr/lib/x86_64-linux-gnu/openblas/libblas.so.3
LAPACK: /usr/lib/x86_64-linux-gnu/libopenblasp-r0.2.20.so

LightGBM installed from source (following R installation instructions), version 3.1.1.99

## Additional Comments
This is possibly the same as issue  #3998

[...]

PR #4006, opened 2021-02-20
[DOCS] Update docs to note that pred_contrib is not available for linear trees

btrotta:
As noted in #4002, `pred_contrib` does not work for linear trees.

### DIFF
--- a/src/application/predictor.hpp
+++ b/src/application/predictor.hpp
@@ -85,6 +85,10 @@ class Predictor {
         }
       };
     } else if (predict_contrib) {
+      if (boosting_->IsLinear()) {
+        Log::Warning("Predicting SHAP feature contributions is not implemented for linear trees. "
+          "Results will be incorrect.");
+      }
       predict_fun_ = [=](const std::vector<std::pair<int, double>>& features,
                          double* output) {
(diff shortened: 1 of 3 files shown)


jameslamb (commented):
@btrotta could you add a warning in code, preferably at the C++ level? For other frameworks that LightGBM models are used with, there might not be any manual review of the `pred_contrib` output, and that might mean that wrong values are just silently used. I think raising a warning would be helpful, because then at least there is a chance of it being noticed in logs.

src/application/predictor.hpp
```diff
@@ -85,6 +85,10 @@ class Predictor {
         }
       };
     } else if (predict_contrib) {
+      if (boosting_->IsLinear()) {
+        Log::Warning("Predicting SHAP feature contributions is not implemented for linear trees. "
```
StrikerRUS, line 89: Maybe raising error will be better? I believe it is better than returning nonsense results.
```suggestion
        Log::Fatal("Predicting SHAP feature contributions is not implemented for linear trees. "
```

StrikerRUS (approved):
Thanks a lot!

Issue #3944, opened 2021-02-12
[dask] add a DaskDataset


jameslamb:
## Summary

`lightgbm.dask` should include a `DaskDataset`, a version of the LightGBM `Dataset` object where array-like inputs are provided as Dask collections.

## Motivation

This would greatly simplify the internals for of `lightgbm.dask`'s training code.  It could consolidate the work currently being done by [`_concat()`](https://github.com/microsoft/LightGBM/blob/84c4b7504a49eed63fa819a92db7d0b18292f3f4/python-package/lightgbm/dask.py#L136) + [`_split_to_parts()`](https://github.com/microsoft/LightGBM/blob/84c4b7504a49eed63fa819a92db7d0b18292f3f4/python-package/lightgbm/dask.py#L199) + [all of this `parts` stuff](https://github.com/microsoft/LightGBM/blob/84c4b7504a49eed63fa819a92db7d0b18292f3f4/python-package/lightgbm/dask.py#L289-L317).

This feature is also a necessary prerequisite to supporting a functional, non-sklearn interface (#https://github.com/microsoft/LightGBM/issues/3846). The non-Dask `train()` function takes in a Dataset object (https://github.com/microsoft/LightGBM/blob/84c4b7504a49eed63fa819a92db7d0b18292f3f4/python-package/lightgbm/engine.py#L27).

## Description

Some of LightGBM's most powerful features come from data pre-processing before the training process begins. This includes binning continuous data into histograms, bundling sparse features, and many other pre-processing steps. In LightGBM core, these steps are handled in the `Dataset` object.

* Python: https://github.com/microsoft/LightGBM/blob/84c4b7504a49eed63fa819a92db7d0b18292f3f4/python-package/lightgbm/basic.py#L1035
* C++: https://github.com/microsoft/LightGBM/blob/84c4b7504a49eed63fa819a92db7d0b18292f3f4/src/io/dataset.cpp
* R: https://github.com/microsoft/LightGBM/blob/84c4b7504a49eed63fa819a92db7d0b18292f3f4/R-package/R/lgb.Dataset.R

## References

I described the `Dataset` object at a high level in https://stackoverflow.com/questions/65924856/lightgbm-intent-of-lightgbm-dataset/66004139#66004139.

`xgboost.dask` has already done something similar with their dataset object, the DMatrix

[...]

PR #4024, opened 2021-02-26
[dask] Include support for raw_score in predict (fixes #3793)

### LINKED ISSUE #3793: [dask] support 'raw_score' in predict()

jmoralez:
This attempts to solve #3793 by testing the `predict` method for `DaskLGBMRegressor` and `DaskLGBMClassifier` with `raw_score=True`. The tests for `DaskLGBMClassifier` are performed with binary and multi-class classification since `predict_proba=True` and `raw_score=True` for binary classification return a 1d array/pandas Series as opposed to the multi-class variants (which return the same number of columns as classes). 

I found that using `drop_axis=1` for `dask.array.map_blocks` works everytime and avoids having many `if` statements testing the conditions described above but I'm happy to discuss the best way to do this. I also replaced the decision to return a `pd.Series` or `pd.DataFrame` by checking the shape of the input array, which I think should be discussed as well.


tests/python_package_test/test_dask.py
```diff
@@ -1255,3 +1255,59 @@ def test_sklearn_integration(estimator, check, client):
 def test_parameters_default_constructible(estimator):
     name, Estimator = estimator.__class__.__name__, estimator.__class__
     sklearn_checks.check_parameters_default_constructible(name, Estimator)
+
+
+@pytest.mark.parametrize('task', ['binary_classification', 'multi-class_classification', 'regression'])
```
jameslamb, line 1260: can you please adding ranking tests as well?

jameslamb: actually this change also makes me realize that the other tests using `@pytest.mark.parametrize('task', tasks)` are not currently testing multi-class classification.

Are you interested in submitting a separate PR to add `multi-class-classification` to `tasks`? https://github.com/microsoft/LightGBM/blob/6356e659af74909aa77eb87d54b79f0373772509/tests/python_package_test/test_dask.py#L41

Doing that would improve our test coverage AND reduce the risk of mistakes like this where a task if forgotten.

tests/python_package_test/test_dask.py
```diff
@@ -1255,3 +1255,59 @@ def test_sklearn_integration(estimator, check, client):
...
+@pytest.mark.parametrize('output', data_output)
+def test_predict_with_raw_score(task, output, client):
+    if task.endswith('classification'):
+        objective = 'classification'
+        model_factory = lgb.DaskLGBMClassifier
+    elif task == 'regression':
+        objective = 'regression'
+        model_factory = lgb.DaskLGBMRegressor
+
+    if task.startswith('multi'):
```
jameslamb, line 1270:
```suggestion
    if task == 'multi-class_classification':
```

There is only one task that this matters for, so I'd rather be explicit and check that it is multi-class classification. That reduces indirection and makes the test code easier to follow.

tests/python_package_test/test_dask.py
```diff
@@ -1255,3 +1255,59 @@ def test_sklearn_integration(estimator, check, client):
...
+        for i in range(3):  # we test for three classes
+            class_df = trees_df[trees_df.tree_index == i]
+            expected_class_mean = compute_expected_predictions_mean(class_df)
+            expected_mean.append(expected_class_mean)
+    else:
+        expected_mean = compute_expected_predictions_mean(trees_df)
+    np.testing.assert_almost_equal(computed_mean, expected_mean)
+
+    if task.endswith('classification'):
+        raw_probabilities = model.predict_proba(dX, raw_score=True).compute()
```
jameslamb, line 1310:
```suggestion
        pred_proba_raw = model.predict_proba(dX, raw_score=True).compute()
```

The name "probabilities" is misleading here, because the results are not actually probabilities. If you pass `raw_score=True`, you get back the raw leaf values.

tests/python_package_test/test_dask.py
```diff
@@ -1255,3 +1255,59 @@ def test_sklearn_integration(estimator, check, client):
...
+    model.fit(dX, dy)
+    raw_predictions = model.predict(dX, raw_score=True).compute()
+    computed_mean = np.mean(raw_predictions, axis=0)
+
+    def compute_expected_predictions_mean(tree_df: pd.DataFrame):
+        """Computes the expected mean of the predictions from the dataframe of a tree with two leaves."""
+        leafs_df = tree_df.iloc[1:]  # remove root node
+        leafs_weights = leafs_df['weight'].values
+        leafs_values = leafs_df['value'].values
+        return leafs_values @ leafs_weights / leafs_weights.sum()
```
jameslamb, line 1296:
```suggestion
    def compute_expected_predictions_mean(tree_df: pd.DataFrame):
        """Computes the expected mean of the predictions from a model with one tree and two leaves."""
        leaf_values = tree_df['value']
        leaf_weights = tree_df['weight']
        sum_weights = leaf_weights[1] + leaf_weights[2]
        return leaf_values[1] * (leaf_weights[1]/sum_weights) + leaf_values[2] * (leaf_weights[2]/sum_weights)
```

Unless you disagree with my suggestion from https://github.com/microsoft/LightGBM/issues/3793#issuecomment-785613951 because you see a correctness or performance issue, please just use that code or something similarly straightforward.

I'm opposed to adding uncommon operators like [the `@` operator](https://www.python.org/dev/peps/pep-0465/)  unless there is a correctness of performance reason to do so. In my opinion, doing that makes this code harder to understand for other contributors.

tests/python_package_test/test_dask.py
```diff
@@ -1255,3 +1255,59 @@ def test_sklearn_integration(estimator, check, client):
...
+    trees_df = model.booster_.trees_to_dataframe()
+    if task == 'multi-class_classification':
+        expected_mean = []
+        for i in range(3):  # we test for three classes
+            class_df = trees_df[trees_df.tree_index == i]
+            expected_class_mean = compute_expected_predictions_mean(class_df)
+            expected_mean.append(expected_class_mean)
+    else:
+        expected_mean = compute_expected_predictions_mean(trees_df)
+    np.testing.assert_almost_equal(computed_mean, expected_mean)
```
jameslamb, line 1307: is this guaranteed to fail if the shape of `computed_mean` is incorrect? Like it is a `(3,)` array instead of `(1, 3)`? If not, I think we should add a specific test on shape to be sure mistakes aren't made in multiclass classification

tests/python_package_test/test_dask.py
```diff
@@ -1255,3 +1255,59 @@ def test_sklearn_integration(estimator, check, client):
 def test_parameters_default_constructible(estimator):
     name, Estimator = estimator.__class__.__name__, estimator.__class__
     sklearn_checks.check_parameters_default_constructible(name, Estimator)
+
+
+@pytest.mark.parametrize('task', ['binary_classification', 'multi-class_classification', 'regression'])
+@pytest.mark.parametrize('output', data_output)
+def test_predict_with_raw_score(task, output, client):
+    if task.endswith('classification'):
+        objective = 'classification'
```
jameslamb, line 1264: please change this to match the other tests, where `task` is passed directly into `_create_data()`: https://github.com/microsoft/LightGBM/blob/6356e659af74909aa77eb87d54b79f0373772509/tests/python_package_test/test_dask.py#L641.

Creating another layer of indirection here is unnecessary and could lead to mistakes. The names you use for `task` and the names accepted by the `objective` argument of `_create_data()` should exactly match.

If you agree with my proposal in another comment to add a `multiclass-classification` task, I think it would be appropriate in that PR to add this in `_create_data()`:

```python
if object == 'multiclass-classification' and centers < 3:
    centers = 3
```

jameslamb (changes requested):
Thanks for taking this on! Please see my first round of comments

python-package/lightgbm/dask.py
```diff
@@ -505,7 +501,7 @@ def _predict(
             pred_leaf=pred_leaf,
             pred_contrib=pred_contrib,
             dtype=dtype,
-            **kwargs
+            drop_axis=1
```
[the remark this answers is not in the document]
jmoralez: Hi James. This was necessary for `predict_proba` with `raw_score=True` for binary classification, because the current handling of this is:
```python
if pred_proba:
    kwargs['chunks'] = (data.chunks[0], (model.n_classes_,))
else:
    kwargs['drop_axis'] = 1
```
however, for binary classification with raw_score the result doesn't have `model.n_classes_` (2) columns but just 1. So I had to include more conditions to check if this was the raw_score of binary classification. I messed up the conditions once and saw that it went to the else and it worked. Same for the dataframe case where I changed it to just check the shape of the `result` array.

tests/python_package_test/test_dask.py
jmoralez: Sure. Should this modify the `_create_data` function to add an objective of `multi-class-classification`?

tests/python_package_test/test_dask.py
jmoralez: Yes, this fails if one array is `(3,)` and the other is either `(1,3) or (3,1)`.

tests/python_package_test/test_dask.py
jmoralez: I agree the data creation for this test is pretty annoying. Should I open that PR modifying the tasks and update this when that gets merged?

python-package/lightgbm/dask.py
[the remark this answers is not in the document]
jameslamb: ah now I understand better what you meant in the PR description. Sounds good, thanks very much for the fix!

tests/python_package_test/test_dask.py
jameslamb: excellent, thank you

tests/python_package_test/test_dask.py
jameslamb: yes please! That would be great. I think that one will be non-controversial and should be a quick review.

tests/python_package_test/test_dask.py
jameslamb: yes please! based on this discussion: https://github.com/microsoft/LightGBM/pull/4024#discussion_r583795293

when you do that, I think it would make sense to remove the `centers` argument from that function, and just use the `objective` to set `centers=2` or `centers=3`. We don't have any tests right now that use more than 3 classes or that really care about the exact numeric values of the cluster centers coming out of `make_blobs()`.

jmoralez: Hi James. I'm having trouble with the tests for ranking. The check for the expected mean is because if you have 25 samples in one leaf with value 0 and 75 samples in another leaf with value 1 you expect your mean to be (25 * 0 + 75 * 1) / (25 + 75), right? Which I guess would be roughly equal to checking that there are 25 predictions with the value 0 and 75 with the value 1. But for ranking I'm seeing that these don't match:
image

jmoralez: btw this doesn't happen with local ranker.

jameslamb: I don't understand what you mean by "don't match" and "having trouble", sorry. Where are you getting the numbers 75 and 25 from?

jmoralez: By having trouble I mean the tests aren't passing because the expected mean isn't the actual mean. The 75 and 25 in the image are the actual counts for the prediction values, so we expect to have 77 samples with prediction -0.025374 and 23 samples with prediction 0.072166 but we get 75 and 25 respectively.

jameslamb: oh wow! Ok yeah that was not clear to me, thanks for rephrasing it. That definitely seems problematic.

However, I want to clarify something...you're using `weight`, not `count`, to get the expected mean, right?

jmoralez: Yes, I used the counts here for clarity but the normalized weights are roughly the same (0.772, 0.228)

jameslamb: Ok, got it! This definitely seems like a bug. Does it happen on every test for the ranker, and only for the ranker?

If so, can you open a separate issue with a minimal reproducible example? You could use https://github.com/microsoft/LightGBM/blob/master/examples/python-guide/dask/ranking.py if you need a simpler data setup for ranking.

This seems like a bug and I have a few ideas of where it could be but I don't expect that it's specific to Dask or to the changes in this PR, so I think it would be better to discuss it in a separate issue.

jmoralez: Every test for the ranker fails, but I just realized that I wasn't adding sample weight to the other tasks and if I do then they all fail. This isn't right, is it?

jameslamb: Is the reverse true? If you do unweighted training for all tasks, do the tests pass?

If so, then I think these tests should use unweighted training, and that the "counts of unique preds disagrees with the results of `trees_to_dataframe()`" thing is a separate issue.

jmoralez: Yes, without using weights the tasks succeed except for the ranker.

jameslamb: oh that's different than what I understood before. I thought your last comment meant "I noticed that ranker was the only task that I was doing weighted training for".

Ok, I think the challenge here is that the `weight` column in the tree node metadata and the sample weights are related but not identical concepts. For this test, let's simplify by not using weighted training.

For the ranking test, can you just use a tolerance like `assert abs(expected_mean - preds_mean) < threshold`? And then open the separate issue from https://github.com/microsoft/LightGBM/pull/4024#discussion_r592870901 documenting the inconsistency you saw in the results and small reproducible example if possible. Based on what you've said so far I suspect you've uncovered a bug in distributed LightGBM, and I don't think that bug needs to block this PR.

jmoralez: Should I compare the normalized weights from the dataframe vs the normalized counts from the unique predictions in that issue?

jmoralez: Or I could compare directly the count column vs the unique counts of the predictions. It's not very clear to me why we don't use that here, why are we using the weights?

jameslamb: you know what now that I think about it...it would be enough to compare the unique values of the leaf nodes to the unique values of the preds. If those are the same, you know you got the raw scores

jmoralez: So the tests would check `set(raw_predictions) == set(trees_df['value'])`?

jameslamb: yes

jmoralez: done. please let me know what you think

jameslamb (approved):
Looks great to me, thanks very much! I'm going to wait to merge this until after LightGBM 3.2.0 is released (#3872 ), assuming that happens in the next day or two.

jmoralez:
Sounds good. What would be a good next contribution? I'd like to work on the dataset but I'm happy to contribute something else if you consider it more important

jameslamb:
> Sounds good. What would be a good next contribution? I'd like to work on the dataset but I'm happy to contribute something else if you consider it more important

thanks very much! Could you try adding  tests for `voting_parallel` learning? As part  of https://github.com/microsoft/LightGBM/issues/3834. I added some notes on how to get started at https://github.com/microsoft/LightGBM/issues/3834#issuecomment-802541296.

I don't want to start on `DaskLGBMDataset` yet (https://github.com/microsoft/LightGBM/issues/3944) since I think that will take a while and will probably conflict with @ffineis 's work on adding eval sets (https://github.com/microsoft/LightGBM/pull/3952#issuecomment-797862596).

jmoralez:
Wow I hadn't seen the voting parallel learning, it looks awesome. Will definitely work on that and hope that it doesn't suffer from what I've seen in #4026.

StrikerRUS:
@jmoralez Could you please sync with latest `master` branch?

Issue #4028, opened 2021-03-01
[docs] Add alt text on images


jameslamb:
## Summary

For every image in LightGBM's documentation, alternative text that describes the image should be added.

## Motivation

People with limited vision use screen readers to view web pages. When these screen readers encounter an image, they will only read some placeholder like "IMAGE" by default. However, if [the "alt" tag](https://www.w3schools.com/tags/att_img_alt.asp) is specified, many screen readers will use that instead.

Adding alt text for images would improve the accessibility of LightGBM's documentation for users who use screen readers.

## Description

To help with this issue, submit a pull request that adds alt text for images in LightGBM's documentation. Please limit pull requests to only 1 file, to simplify reviews.

As of this writing, the following documentation files have images (based on `git grep -E "\.\. image" docs`)

* [x] `docs/Features.rst` (#4036)
* [x] `docs/GPU-Performance.rst` (#4036)
* [x] `docs/GPU-Windows.rst` (#4036)
* [x] `docs/Parallel-Learning-Guide.rst` (#4030)
* [x] `docs/Parameters-Tuning.rst` (#4035)
* [x] `docs/gcc-Tips.rst` (#4187)
* [x] `docs/index.rst` (#4029)

When adding alt text, follow the advice in https://accessibility.psu.edu/images/alttext/.

> ...include what you might relay over the phone

> it should describe the image. It should NOT include information about copyright or any extra information that would be relevant to a seeing audience.

This can be done with the `:alt:` parameter in Sphinx `..image` directives (see [the Sphinx docs](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/images.html) for details).

For example:

```rst
.. image:: ./_static/images/some-image.png
  :alt: A screenshot with build logs from GCC.
```

## References

While working on this, you might also consider some of the recommendations from https://webaim.org/techniques/screenreader/#how. For example, things that should be read out loud as acronyms, like `gcc`, should be spelled out all in capital letters ("GCC").

akshitadixit:
Hello, may I please work on this?

jameslamb:
Sure, thanks @akshitadixit ! Please feel free to submit a pull request based on the description above. Let me know if you have any questions!

subhamagrawal7:
I would like to try and add alt text to Parameters-Tuning.rst. Could you assign this?

jameslamb:
Thanks @subhamagrawal7 ! Please feel free to submit a pull request based on the description above, and let me know if you have any questions.

[...]

PR #4035, opened 2021-03-01
[docs] Add alt text to image in Parameters-Tuning.rst

subhamagrawal7:
Add alt text to Leaf-wise growth image, as part of #4028

### DIFF
--- a/docs/Parameters-Tuning.rst
+++ b/docs/Parameters-Tuning.rst
@@ -70,6 +70,7 @@ LightGBM adds nodes to trees based on the gain from adding that node, regardless
 
 .. image:: ./_static/images/leaf-wise.png
    :align: center
+   :alt: Figure of Leaf-wise tree growth from Features documentation which grows trees leaf-wise (best-first)
 
 Because of this growth strategy, it isn't straightforward to use ``max_depth`` alone to limit the complexity of trees. The ``num_leaves`` parameter sets the maximum number of nodes per tree. Decrease ``num_leaves`` to reduce training time.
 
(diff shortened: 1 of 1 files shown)


docs/Parameters-Tuning.rst
```diff
@@ -70,6 +70,7 @@ LightGBM adds nodes to trees based on the gain from adding that node, regardless
 
 .. image:: ./_static/images/leaf-wise.png
    :align: center
+   :alt: Figure of Leaf-wise tree growth from Features documentation which grows trees leaf-wise (best-first)
```
jameslamb, line 73: I don't think this would be very clear for someone who cannot already see the image. Alternative text is intended to describe an image for someone who cannot see it, either because the image cannot be loaded or because the person uses a screen reader to read web pages.

What do you think about this instead?

```suggestion
   :alt: Three consecutive images of decision trees, where each shows the tree with an additional two leaf nodes added. Shows that leaf-wise growth can result in trees that have some branches which are longer than other.
```

jameslamb (changes requested):
Thanks! Please see my suggestion

docs/Parameters-Tuning.rst
subhamagrawal7: Ok I thought that a too long alt text may not be allowed else your suggestion is perfect for explaining the image

subhamagrawal7: Should I commit the same??

jameslamb: Yes, if you agree with the suggestion you can click "Commit suggestion" right here in the GitHub user interface. That way, you don't have to manually change local files and push a new commit.

subhamagrawal7:
@jameslamb could you please review this and tell if anything else is needed..

jameslamb:
@subhamagrawal7 sorry for the inconvenience, but could you please merge in the latest changes from `master`? We recently had to fix some issues in this project's continuous integration.

subhamagrawal7:
@jameslamb I have merged the upstream/master into my branch now what should I do??

jameslamb (approved):
thanks very much!

PR #4027, opened 2021-03-01
[docs] update description of deterministic parameter

shiyu1994:
Update the description of `deterministic` parameter. To guarantee that `deterministic=true` does produce stable results. We should also set `force_col_wise=true` or `force_row_wise=true`.

Issue #3763, opened 2021-01-14
List of dependencies that should be updated from time to time


StrikerRUS:
Here is a list of different places in LightGBM's GitHub repo where we specify some dependencies or helpers. Quite often we should specify a particular version of such software. And these versions are tend to obsolete with time. If you see that there is a newer version comparing to what we have, please feel free to propose a PR with updates.

- SHAs of commits of third-party dependencies in the [`external_libs` folder](https://github.com/microsoft/LightGBM/tree/master/external_libs). If there are some checkpoints (markers of stable versions) of these libraries, e.g. Releases at GitHub, the most recent stable release should be used (not the unstable latest `master` commit). If there are no such indicators of stable commit, use the latest available one. *How to [update GitHub submodules](https://stackoverflow.com/questions/5828324/update-git-submodule-to-latest-commit-on-origin).*
- SHAs of commits OpenCL libraries integration (same notes about versions as above):
1.https://github.com/microsoft/LightGBM/blob/f997a0692ca0f26740d2bdef2695c3e881d4e918/CMakeIntegratedOpenCL.cmake#L5-L6
2.https://github.com/microsoft/LightGBM/blob/f997a0692ca0f26740d2bdef2695c3e881d4e918/CMakeIntegratedOpenCL.cmake#L8-L9
3.https://github.com/microsoft/LightGBM/blob/f997a0692ca0f26740d2bdef2695c3e881d4e918/CMakeIntegratedOpenCL.cmake#L11-L12
- Visual Studio version for OpenCL libraries integration https://github.com/microsoft/LightGBM/blob/f997a0692ca0f26740d2bdef2695c3e881d4e918/CMakeIntegratedOpenCL.cmake#L66-L68
- MinGW w64 version at Appveyor https://github.com/microsoft/LightGBM/blob/f997a0692ca0f26740d2bdef2695c3e881d4e918/.appveyor.yml#L26
- [Read the Docs config file](https://github.com/microsoft/LightGBM/blob/master/.readthedocs.yml) (version of config file; adding new and removing deprecated config values)
- [`PublishBuildArtifacts` Azure Pipelines task](https://github.com/microsoft/LightGBM/blob/master/.vsts-ci.yml) (multiple occurrences, do search the file for `PublishBuildArtifacts` keyword)
[...]

PR #4042, opened 2021-03-03
[ci] [R-package] upgrade to R 4.0.4 in CI

jameslamb:
R Windows CI jobs are currently failing on `master` ([logs](https://github.com/microsoft/LightGBM/runs/2018204127)) and #4027 ([logs](https://github.com/microsoft/LightGBM/pull/4027/checks?check_run_id=2018832067)) with the following:

```text
* checking whether package 'lightgbm' can be installed ... WARNING

  Warning: package 'R6' was built under R version 4.0.4

See 'C:/tmp-r-cmd-check/lightgbm.Rcheck/00install.out' for details.

Status: 1 WARNING
```

There has not been a release of `{R6}` since October 2020 (https://cran.r-project.org/web/packages/R6/index.html), but CRAN does create new package binaries when a new version of R is released. R version 4.0.4 was released about two weeks ago: https://twitter.com/pdalgd/status/1361246911378960391, but right now LightGBM's CI uses 4.0.3.

This PR proposes resolving the warning by updating R to 4.0.4.

### Notes for Reviewers

I did look into whether this warning could be disabled in `R CMD CHECK` with an environment variable, and it doesn't look like it.

* https://github.com/wch/r-source/blob/613bdfd0e1d3fc9984142d5da3da448adf2438c7/src/library/base/R/library.R#L321
* https://github.com/wch/r-source/blob/613bdfd0e1d3fc9984142d5da3da448adf2438c7/src/library/base/R/library.R#L119-L122

Linking to #3763

### DIFF
--- a/.ci/test_r_package.sh
+++ b/.ci/test_r_package.sh
@@ -27,8 +27,8 @@ if [[ "${R_MAJOR_VERSION}" == "3" ]]; then
     export R_LINUX_VERSION="3.6.3-1bionic"
     export R_APT_REPO="bionic-cran35/"
 elif [[ "${R_MAJOR_VERSION}" == "4" ]]; then
-    export R_MAC_VERSION=4.0.3
-    export R_LINUX_VERSION="4.0.3-1.1804.0"
+    export R_MAC_VERSION=4.0.4
+    export R_LINUX_VERSION="4.0.4-1.1804.0"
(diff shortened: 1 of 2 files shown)


shiyu1994 (approved):

StrikerRUS:
Thanks a lot for the prompt fix!

PR #4043, opened 2021-03-04
enforce interaction constraints with monotone_constraints_method = intermediate/advanced

ChristophAymannsQC:
First attempt to address #4044 .


tests/python_package_test/test_interaction_constraints.py
```diff
@@ -0,0 +1,134 @@
...
+        "monotone_constraints": [1, 1, 0],
+        "monotone_constraints_method": "basic",
+    }
+    return boosting_params
+
+
+@pytest.mark.parametrize(
+    "monotone_constraints_method", ["basic", "intermediate", "advanced"]
+)
+def test_interaction_constraints(
```
StrikerRUS, line 103: Please merge this new test into already existent one:
https://github.com/microsoft/LightGBM/blob/37e987828d3141aded1a6e356fdcc7954a6f2f5c/tests/python_package_test/test_engine.py#L1255

For the reference: here is `test_interaction_constraints`
https://github.com/microsoft/LightGBM/blob/37e987828d3141aded1a6e356fdcc7954a6f2f5c/tests/python_package_test/test_engine.py#L2465

ChristophAymannsQC: Thanks for the comments! Just to clarify, you asked that I add the test to `test_monotone_constraints()` or `test_interaction_constraints()`? Seems to me that `test_interaction_constraints()` would be the better place for it. Alternatively, I could just add a new function `test_monotone_and_interaction_constraints()` to  `test_engine.py` that does what I'm currently doing in `test_interaction_constraints.py`.

ChristophAymannsQC: ok. had closer look at the tests. I think I can merge into `test_monotone_constraints()`.

StrikerRUS: I think any variant will be fine, but better will be one that reuses existing test infrastructure as much as possible. Thank you!

ChristophAymannsQC: I updated `test_monotone_constraints()` as discussed.

btrotta (approved):
Looks fine to me

tests/python_package_test/test_engine.py
```diff
@@ -1273,28 +1274,72 @@ def is_correctly_constrained(learner, x3_to_category=True):
...
+
+        def has_interaction(treef):
+            n = 0
+            for fs in feature_sets:
+                if len(treef.intersection(fs)) > 0:
+                    n += 1
+            if n > 1:
+                return True
+            else:
+                return False
```
StrikerRUS, line 1313:
```suggestion
            return n > 1
```

StrikerRUS (changes requested):
Thanks a lot for your contribution!
I have only few minor comments for tests code.

tests/python_package_test/test_interaction_constraints.py
StrikerRUS: Thanks a lot!

ChristophAymannsQC:
> Thanks a lot for your contribution!
> I have only few minor comments for tests code.

Thanks a lot for the comments. Added as suggested.

StrikerRUS (approved):
Python test part looks fine to me! Thank you!

src/treelearner/serial_tree_learner.cpp
```diff
@@ -768,7 +768,7 @@ double SerialTreeLearner::GetParentOutput(const Tree* tree, const LeafSplits* le
   return parent_output;
 }
 
-void SerialTreeLearner::RecomputeBestSplitForLeaf(int leaf, SplitInfo* split) {
+void SerialTreeLearner::RecomputeBestSplitForLeaf(Tree* tree, int leaf, SplitInfo* split) {
```
jameslamb, line 771: sorry for not reveiewing sooner, but this is a breaking change to the library's API. is that ok?

Just want to be sure we consider it before this is merged

jameslamb (changes requested):

src/treelearner/serial_tree_learner.cpp
StrikerRUS: I guess it is internal API not public C API, isn't it?

jameslamb: that method is marked `protected` (https://github.com/microsoft/LightGBM/blob/fcfd4132e6d40a22d52023396329c41fd3de4a42/src/treelearner/serial_tree_learner.h#L121).

I'm not that experienced with C++, but from https://docs.microsoft.com/en-us/cpp/cpp/protected-cpp?view=msvc-160#remarks it seems `protected` methods aren't exactly "internal".

> Protected members are not as private as private members, which are accessible only to members of the class in which they are declared, but they are not as public as public members, which are accessible in any function.

> Class members declared as protected can be used only by the following....Classes derived with public or protected access from the class that originally declared these members.

Reading that, it seems to me that someone writing their own extension of LightGBM could be relying on this API.

@btrotta could you comment on this? If I'm totally wrong here I apologize for slowing down the PR and would be happy to remove my blocking review.

StrikerRUS: @jameslamb 
> Reading that, it seems to me that someone writing their own extension of LightGBM could be relying on this API.

The only API for which we are trying to provide backward compatibility is C API and all functions from it are listed in this file: https://github.com/microsoft/LightGBM/blob/master/include/LightGBM/c_api.h
All other functions are treated as internal and we do not make any promises regarding their signatures.

jameslamb: alright, my review can be dismissed then. Sorry to slow the process down.

ChristophAymannsQC: thanks you nevertheless for your comments.

StrikerRUS:
I think @CharlesAuguste is interested in this.

StrikerRUS:
Gently ping @CharlesAuguste

CharlesAuguste:
I completely missed this, I am very sorry. I will have a look at this PR in the next few days and write back.

CharlesAuguste:
@StrikerRUS I just had a look at the code, and I agree with @ChristophAymannsQC that everything should be working well with this fix. I also think it was a good idea to parametrize the test to keep the old test, because the test with the interaction constraints may not be able to catch some issues relating to monotone constraints.

StrikerRUS:
@CharlesAuguste Thanks a lot for your feedback!

@ChristophAymannsQC Thank you very much!

PR #4056, opened 2021-03-09
Set `is_linear_` to false when it is absent from the model file

### LINKED ISSUE #3778: [bug] Creating multiple boosters from file results in broken scores.
Using the master branch, if I create 2 boosters from file using the C API, I get broken scores from all but the first created booster:

```cpp

BoosterHandle bh1, bh2;  

LGBM_BoosterCreateFromModelfile("./LightGBM_model.txt", &num_iterations, &bh1); // comment this & you get good results
LGBM_BoosterCreateFromModelfile("./LightGBM_model.txt", &num_iterations, &bh2);

// Score is only correct if I predict with bh1, or predict with bh2 but don't initialize bh1 before initializing bh2:
predict(bh2, values             , NUM_FEATURES, num_iterations, &out_len, &ref_scores[0]); // score changes depending on bh2 being the first Booster created or not.



```

Meaning, only the first booster created works properly.

I'm on Ubuntu 18.04, x86_64.

Thank you

shiyu1994:
This is to fix #3778. When loading booster from a model file produced by old version of LightGBM without linear tree option, we should set `is_linear_` of a tree to `false`, to avoid uninitialized `is_linear_` value, which can cause undefined behavior.

### DIFF
--- a/src/io/tree.cpp
+++ b/src/io/tree.cpp
@@ -698,6 +698,8 @@ Tree::Tree(const char* str, size_t* used_len) {
     int is_linear_int;
     Common::Atoi(key_vals["is_linear"].c_str(), &is_linear_int);
     is_linear_ = static_cast<bool>(is_linear_int);
+  } else {
+    is_linear_ = false;
   }
 
   if ((num_leaves_ <= 1) && !is_linear_) {


btrotta (approved):

cyfdecyf:
I've been biten by this bug and just figured out the problem.

Loading a model generated by previous version of LightGBM which does not have `is_linear=0`. This leads to uninitialized `is_linear_` instance variable and leads to wrong predict values.

PR #3913, opened 2021-02-06
[ci][R-package] run R CRAN checks on Solaris by optional workflow

StrikerRUS:
Refer to #3740.

How it looks like: https://github.com/StrikerRUS/LightGBM/pull/5#issuecomment-773700926.

image

PR #4061, opened 2021-03-10
[ci] set 'pending' commit status for running R Solaris optional workflow

StrikerRUS:
I forgot to add one line that sets yellow 'pending' status for running R Solaris workflow in #3913.

Similar to R valgrind workflow:
https://github.com/microsoft/LightGBM/blob/15853a7a02621347aa74bdacb97bbbdc80b25cab/.github/workflows/r_valgrind.yml#L31-L37

Noticed that in https://github.com/microsoft/LightGBM/pull/3872#issuecomment-795972470.

### DIFF
--- a/.github/workflows/r_solaris.yml
+++ b/.github/workflows/r_solaris.yml
@@ -31,6 +31,7 @@ jobs:
       - name: Send init status
         if: ${{ always() }}
         run: |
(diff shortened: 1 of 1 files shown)


jameslamb (approved):
ohhh nice, sorry that I missed this in review.

Issue #3782, opened 2021-01-18
[dask] Support more operating systems


StrikerRUS:
## Summary

The Dask interface in https://github.com/microsoft/LightGBM/blob/706f2af7badc26f6ec68729469ec6ec79a66d802/python-package/lightgbm/dask.py currently only supports Linux OS.

It should support other operating systems that are supported by LightGBM itself (macOS and Windows) as well.

## Motivation

Adding this feature would allow users to use lightgbm.dask on their favorite OS.

## References

To close this issue modify `lightgbm.dask` codebase so that it will be possible to remove the following `skip` directive.
https://github.com/microsoft/LightGBM/blob/ed651e86729477e7d2699534d63a5cace87f7878/tests/python_package_test/test_dask.py#L7-L8

StrikerRUS:
Closing this, as we use #2302 to track feature requests. Leave a comment below if you'd like to contribute this feature, and we'll be happy to re-open it!

jameslamb:
I tried testing today whether #3840 would allow use of the Dask package on Mac (based on https://github.com/microsoft/LightGBM/issues/3839#issuecomment-766410406).

I found that most tests from `pytest tests/python_package_test/test_dask.py` pass, but the ranker tests regularly fail.


```text
==================================== short test summary info ====================================
------------------------------------- Captured stderr call --------------------------------------
[LightGBM] [Fatal] Socket recv error, code: 54
distributed.scheduler - INFO - Remove worker <Worker 'tcp://127.0.0.1:53872', name: tcp://127.0.0.1:53872, memory: 5, processing: 1>
distributed.core - INFO - Removing comms to tcp://127.0.0.1:53872
distributed.worker - WARNING -  Compute Failed
Function:  _train_part
args:      ()
[...]

PR #4064, opened 2021-03-12
[ci] run Dask examples on CI

StrikerRUS:
Dask examples are placed in the [`python-guide/dask`](https://github.com/microsoft/LightGBM/tree/master/examples/python-guide/dask) subdirectory.

Also this PR adds some punctuation to Dask guide.

### DIFF
--- a/docs/Parallel-Learning-Guide.rst
+++ b/docs/Parallel-Learning-Guide.rst
@@ -65,7 +65,7 @@ LightGBM's Python package supports distributed learning via `Dask`_. This integr
 Dask Examples
 '''''''''''''
 
-For sample code using ``lightgbm.dask``, see `these Dask examples`_ .
+For sample code using ``lightgbm.dask``, see `these Dask examples`_.
 
 Training with Dask
 ''''''''''''''''''
@@ -121,9 +121,9 @@ While setting up for training, ``lightgbm`` will concatenate all of the partitio
 
 When setting up data partitioning for LightGBM training with Dask, try to follow these suggestions:
 
-* ensure that each worker in the cluster has some of the training data
-* try to give each worker roughly the same amount of data, especially if your dataset is small
-* if you plan to train multiple models (for example, to tune hyperparameters) on the same data, use ``client.persist()`` before training to materialize the data one time
+* ensure that each worker in the cluster has some of the training data;
+* try to give each worker roughly the same amount of data, especially if your dataset is small;
+* if you plan to train multiple models (for example, to tune hyperparameters) on the same data, use ``client.persist()`` before training to materialize the data one time.
 
 Using a Specific Dask Client
 ****************************
@@ -195,9 +195,9 @@ If you are running multiple Dask worker processes on physical host in the cluste
 
   Providing ``machines`` gives you complete control over the networking details of training, but it also makes the training process fragile. Training will fail if you use ``machines`` and any of the following are true:
 
-  * any of the ports mentioned in ``machines`` are not open when training begins
-  * some partitions of the training data are held by machines that that are not present in ``machines``
-  * some machines mentioned in ``machines`` do not hold any of the training data
+  * any of the ports mentioned in ``machines`` are not open when training begins;
+  * some partitions of the training data are held by machines that that are not present in ``machines``;
+  * some machines mentioned in ``machines`` do not hold any of the training data.
 
 **Option 2: specify one port to use on every worker**
 
@@ -223,8 +223,8 @@ You could edit your firewall rules to allow communication between any of the wor
 
   Providing ``local_listen_port`` is slightly less fragile than ``machines`` because LightGBM will automatically figure out which workers have pieces of the training data. However, using this method, training can fail if any of the following are true:
 
-  * the port ``local_listen_port`` is not open on any of the worker hosts
-  * any machine has multiple Dask worker processes running on it
+  * the port ``local_listen_port`` is not open on any of the worker hosts;
+  * any machine has multiple Dask worker processes running on it.
 
 Prediction with Dask
 ''''''''''''''''''''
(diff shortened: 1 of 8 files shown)


StrikerRUS:
Good news! I accidentally forgot to exclude running Dask examples on macOS and they succeeded! Sorry, no logs. Linking #3782.

examples/python-guide/dask/binary-classification.py
```diff
@@ -1,10 +1,15 @@
+import sys
+
 import dask.array as da
 from distributed import Client, LocalCluster
 from sklearn.datasets import make_blobs
 
 import lightgbm as lgb
 
 if __name__ == "__main__":
+    if not sys.platform.startswith('linux'):
```
jameslamb, line 10: If the examples were working on Mac, can we remove this check until something breaks? These are supposed to be examples that teach users how to use `lightgbm.dask`, and I'm really uncomfortable with including details only necessary for CI in them.

docs/Parallel-Learning-Guide.rst
```diff
@@ -121,9 +121,9 @@ While setting up for training, ``lightgbm`` will concatenate all of the partitio
 
 When setting up data partitioning for LightGBM training with Dask, try to follow these suggestions:
 
-* ensure that each worker in the cluster has some of the training data
-* try to give each worker roughly the same amount of data, especially if your dataset is small
-* if you plan to train multiple models (for example, to tune hyperparameters) on the same data, use ``client.persist()`` before training to materialize the data one time
+* ensure that each worker in the cluster has some of the training data;
```
jameslamb, line 124: Why is it necessary to add all these semicolons?

jameslamb (changes requested):
Thanks for the fix, but I disagree with adding CI-only details into the examples, and I'm confused by the punctuation changes in the docs

examples/python-guide/dask/binary-classification.py
StrikerRUS: > including details only necessary for CI in them.

I don't think these details are CI-related only. Imagine, user is using macOS and trying to get familiar with LightGBM's Dask-package. There is no info about that we only support Linux for now anywhere in the docs. So user runs example, gets a crash and feels disappointed (maybe posts a new issue here).

> If the examples were working on Mac

I guess that according to #3782, this is just a happy coincidence.

However, this is not my strong opinion and if you still think that it is better to allow users execute examples that we are know **may** crash on their systems, I'll be happy to revert these `if` guards.

docs/Parallel-Learning-Guide.rst
StrikerRUS: Sorry, as you know I'm not a native English speaker, I thought these semicolons were just accidentally missed.

> Semicolons were traditionally used in lists to separate each bullet point, but although this is still correct, they are not used as often today. 
   https://www.onlinegrammar.com.au/punctuation-in-lists/

Now I see that I was wrong. But it looks like we need a full stop at the end at least, right?

> A common style is to use lower case for the first letter in each bullet point, and a full stop after the last bullet point.

examples/python-guide/dask/binary-classification.py
jameslamb: I think they are CI-specific because of the use of `sys.exit(0)`. If a user copies that code and runs it in a Python REPL, their Python session will end. If they run it in a Jupyter notebook, they'll get an error like this

> SystemExit: 0
/Users/jlamb/miniconda3/lib/python3.8/site-packages/IPython/core/interactiveshell.py:3425: UserWarning: To exit: use 'exit', 'quit', or Ctrl-D.
  warn("To exit: use 'exit', 'quit', or Ctrl-D.", stacklevel=1)

On this point:

> There is no info about that we only support Linux for now anywhere in the docs.
> I guess that according to #3782, this is just a happy coincidence

I think that's an oversight, and that we should have a note at https://lightgbm.readthedocs.io/en/latest/Parallel-Learning-Guide.html#dask saying that Dask features are only tested on Linux.

While I was developing these examples in #3814 , I tested them on my Mac. So I do actually expect them to continue working on Mac. They're much simpler than the tests in `test_dask.py`, which is why I think we're unable to run `test_dask.py` in Mac CI jobs.

I'm in favor of reverting these `if` guards and instead adding a note to the documentation saying that `lightgbm.dask` is only tested on Linux.

docs/Parallel-Learning-Guide.rst
jameslamb: No problem, totally understood!

I don't think you need to worry about this. For native speakers (in American English, at least...that website you linked to is from Australia 😁 ) bulleted lists don't really have formal rules.  Any mix of the following is equally valid:

* capitalizing vs. not capitalizing the first letter of each bullet point
* using a period at the end of each bullet point
* using semicolons at the end of the first `n-1` bullet points and a period at the end of the final one

So it's not that the current state on `master` is "right" and your proposal is "wrong". They're all equally valid ways to write a bulleted list, and I don't think it's worth spending time worrying about it.

examples/python-guide/dask/binary-classification.py
StrikerRUS: > If a user copies that code and runs it in a Python REPL, their Python session will end. If they run it in a Jupyter notebook, they'll get an error like this

It is suggested to run these examples only from command line and from nothing else.

> After installing the package and its dependencies, any of the examples here can be run with a command like this:
    `python binary-classification.py`

> I'm in favor of reverting these if guards and instead adding a note to the documentation saying that lightgbm.dask is only tested on Linux.

Sure, done!

docs/Parallel-Learning-Guide.rst
StrikerRUS: Got it! Thank you very much for the lesson! 🙂

python-package/README.rst
```diff
@@ -204,6 +204,10 @@ You can use ``python setup.py bdist_wheel`` instead of ``python setup.py install
 Install Dask-package
 ''''''''''''''''''''
 
+.. warning::
+
+    Dask-package is only tested on Linux.
```
jameslamb, line 209: When this file is viewed on GitHub, this warning doesn't render the way it would in Sphinx.

image

Are you ok with that? It doesn't bother me because I hope we'll be able to remove this soon anyway, but just wanted you to be aware.

jameslamb (approved):
looks good, thanks

python-package/README.rst
StrikerRUS: Yeah, like all other `Warning`s and `Note`s in other RSTs:

image

I believe there should be no problem with rendering on PyPI because http://rst.ninjs.org/ shows no errors when I paste raw content of `python-package/README.rst`.
FYI, PyPI used to show blank page in case of any errors in RST formatting of uploaded README, but I don't know what's the current policy of error handling there.

jameslamb: ok sounds good!

PR #4070, opened 2021-03-14
[python-package] Add type hint to the libpath file

deddyjobson:
- Added type hints for libpath.py as per https://github.com/microsoft/LightGBM/issues/3756.
- Verified that there were no additional errors when running mypy.


jameslamb (approved):
Looks good to me, thanks for the contribution!

PR #4068, opened 2021-03-14
[python-package] add type hints on Booster.set_network()

jameslamb:
While working on `lightgbm.dask`, I noticed that the docstring for `machines` in `Booster.set_network()` seems to be incorrect.

It says the following:

> machines : list, set or string
    Names of machines.

However, `machines` is passed to `lightgbm.basic.c_str()`, which only accepts strings: https://github.com/microsoft/LightGBM/blob/bcf443b568052eac05cb8b11b17be9926b862f8f/python-package/lightgbm/basic.py#L2359.

If you pass a set or list, you'll get an error like.

> AttributeError: 'set' object has no attribute 'encode'

```python
import lightgbm as lgb
machines = set(["10.1.1.1:12400", "10.1.1.2:12400"])
lgb.basic.c_str(machines)
```

This PR proposes updating that docstring to make it clear that only strings are supported. It also adds type hints on the other arguments in `Booster.set_network()` and a return type hint, as part of #3756.

PR #4071, opened 2021-03-15
[dask] [ci] fix flaky network-setup test

jameslamb:
This PR attempts to fix the occasional failure in the Dask tests mentioned in https://github.com/microsoft/LightGBM/pull/4068#issuecomment-798972634. One of the test cases in `test_dask.py` explicitly requires that at least two workers join the cluster and have pieces of the training data.

This PR tries to prevent that error by using [`Client.wait_for_workers(2)`](https://distributed.dask.org/en/latest/api.html#distributed.Client.wait_for_workers). That call will block until at least 2 workers are up and have joined the cluster, which should prevent errors caused by one worker process being a little slow to start sometimes.

It's difficult to reliably test that this PR would fix the problem mentioned in https://github.com/microsoft/LightGBM/pull/4068#issuecomment-798972634, since it happens so rarely and randomly. I tried running the test in question 11 times on `master` and 11 times after this change (locally on my Mac) and it passed every time.

### DIFF
--- a/tests/python_package_test/test_dask.py
+++ b/tests/python_package_test/test_dask.py
@@ -1024,6 +1024,8 @@ def test_network_params_not_required_but_respected_if_given(client, task, output
     if task == 'ranking' and output == 'scipy_csr_matrix':
         pytest.skip('LGBMRanker is not currently tested on sparse matrices')
 
+    client.wait_for_workers(2)
+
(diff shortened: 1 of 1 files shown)


StrikerRUS (approved):
Your explanation sounds reasonable to me! Thanks!

PR #3494, opened 2020-10-27
Support deterministic 

guolinke:
partial fix #3372

PR #3492, opened 2020-10-27
avoid min_data and min_hessian are zeros at the same time

guolinke:
to fix #3489

Issue #3463, opened 2020-10-17
splitting for nan value looks wrong


chloe-wang:
hi,

I have a dataset which have a feature with value between -0.5 and 0.5. This feature also contains some missing value with nan.

However, I found the model split on some extremely big value like 1e300 which is very close to the infinity in python. 

![image](https://user-images.githubusercontent.com/22490931/96334080-a0971600-10a0-11eb-882a-159fa65818c8.png)

Per my understanding, missing value will not be used for splitting, so it's weird that the tree will split on that big value.

chloe-wang:
This issue may be related to #2811 , but the issue still exits after I upgraded the lightgbm to v3.0.0.

guolinke:
use 1e300 as split value, mean that non-missing goes to left, and missing goes to right. Is it wrong?

chloe-wang:
Using 1e300 as split value shall not be a big issue, but we found the importance for the feature will increase a lot. We choose a feature in random, and set the half of its value to np.nan, the importance of this feature will be among top 10 while it's not informative at all.

chloe-wang:
And one more thing, in xgboost, I think nan value will not be used for finding split point. Instead nan value will be mapped to the left node or right node after the split point has been decided by those non-missing values. It maybe more reasonable. I not sure for that though.

guolinke:
for nan algorithm, lightgbm and xgboost use the same algorithm. it will try to put nan to left and right, and use the one with the larger gain.
Therefore, if one feature has lots of nan, it has more "freedom" than normal features, and thus, its importance may is higher.

chloe-wang:
Then, why we can see the split value at 1e300, as the non-missing value range only from -0.5 to 0.5. So the split value shall not be bigger than 0.5....

chloe-wang:
Maybe the nan is used for bin construction?

guolinke:
`1e300` is not missing value. `1e300` is the upper bound of non-missing value.

guolinke:
yeas, there is special bin for missing value. 
and in nan finding algorithm, this bin could be in left side or right side. 
In your above example, the nan bin is in the right side, and all non-missing values are in left side. 

In you want more details, you can check the value of decision_type, it contains the information of missing value handle in that node.

chloe-wang:
@guolinke I think you are right. Thanks a lot. I will close the issue.

shiyuangu:
[...]

PR #4081, opened 2021-03-18
Precise text file parsing

cyfdecyf:
When comparing prediction result using command line version and Python API, I noticed some prediction values differs starting at the 5th non-zero digits. I suspect the difference is caused by differnet float number parsing algorithms used in LightGBM and pandas. (For reference, I used `pandas.read_csv(fname, float_precision="round_trip")` to load csv file in my Python code.)

So I added precise text file parsing with `fast_double_parser` and the result confirms my guess.

This patch also contains a simple benchmark which shows `Common::Atof` is much faster than using `fast_double_parser`.


cyfdecyf:
@jameslamb The benchmark result doesn't show `fast_double_parser` being faster than `Comon::Atof`. So this PR is more useful for people who wants to verify problems caused by precision lost of when parsing floating number.

include/LightGBM/utils/common.h
```diff
@@ -330,6 +331,26 @@ inline static const char* Atof(const char* p, double* out) {
   return p;
 }
 
+// Use fast_double_parse and strtod (if parse failed) to parse double.
+inline static const char* AtofPrecise(const char* p, double* out) {
```
shiyu1994, line 335: Does that mean we can replace
https://github.com/microsoft/LightGBM/blob/1d2f3e11018dcba56a8670f07c5d91f642e0266e/include/LightGBM/utils/common.h#L1079-L1102
which was added in #3942 with a call to this `AtofPrecise`? Do they behave exactly the same?

tests/benchmark/parser/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
+cmake_minimum_required(VERSION 3.0)
```
shiyu1994, line 1: I'm not sure whether it is appropriate to add these benchmark scripts to the master branch. Why do you think they are necessary?

shiyu1994 (commented):
Thanks for your contribution. Just two comments that address my concerns.

tests/benchmark/parser/CMakeLists.txt
cyfdecyf: This is not necessary to merge into the master branch. I'm just curious about the actual performance of `Common::Atof` and `fast_double_parser`.

And maybe this benchmark can be helpful for people who want to improve the performance of text parser.

I can remove this commit if you decide to not include it.

include/LightGBM/utils/common.h
cyfdecyf: Here a test program https://gist.github.com/cyfdecyf/63f4e7339bbe5a5a23474fda66375742

`strtod` returns signaling nan for "nan" and the like. For other case the behavior is the same.

I'll update `AtofPrecise` to return quiet nan.

Refer to: [1](http://www.open-std.org/jtc1/sc22/wg14/www/docs/n975.htm#7.20.1.3%20the%20strtod,%20strtof,%20and%20strtold%20functions) [2](http://port70.net/~nsz/c/c11/n1570.html#7.22.1.3)

cyfdecyf: I'm not an expert on floating number. So I write a test program https://gist.github.com/cyfdecyf/63f4e7339bbe5a5a23474fda66375742

The only difference is that `strtod` would not return negative NaN. So the `Atof` function in this gist handles this special case.

Please help take a look at the gist and check if there's any problem. I'll update `Common::Atof` and replace the change added in #3492.

Refer to: [1](http://www.open-std.org/jtc1/sc22/wg14/www/docs/n975.htm#7.20.1.3%20the%20strtod,%20strtof,%20and%20strtold%20functions) [2](http://port70.net/~nsz/c/c11/n1570.html#7.22.1.3)

cyfdecyf: The special handling for -NaN in gist (revision 5) the has one problem though: it's incorrect to handle input like " -nan" (note beginning space). But I'm wonder if there's need for the special handling of this case? From this mailling list [thread](https://newlib.sourceware.narkive.com/CpbSqKGq/strtod-nan-returns-negative-nan), it seems like different C library have different treatment for parsing "-nan".

I suggest just leave `Common::AtofPrecise` as is and don't add special handling for "-nan" and the like.

CMakeLists.txt
```diff
@@ -1,3 +1,4 @@
+OPTION(USE_PRECISE_TEXT_PARSER "Use precise double parser for text input file" OFF)
```
StrikerRUS, line 1: Why do we need new compilation option only for one function? Why not simply use `AtofPrecise` instead of `Atof` by default?
Having a lot of functions doing the same work is very confusing, greatly increases maintenance burden and hurts overall development process.

cyfdecyf: 1. If `AtofPrecise` is faster (or not much slower) than `Atof`, I'd like to use it by default. But it's actually much slower in my simple benchmark
    - When I see `fast_double_parser` mentioned in the commit log, I thought it's been used for text file parsing too. But it's actually not which confused me at first. I guess someone might have done the performance test and thus not using it for text parsing
    - With this performance difference, I'd choose precise version only when precision is required
2. `AtofPrecise` does not behave exactly the same with `Atof`. For some of my test models, prediction results for csv input can differ starting from the 5th non-zero digit
    - As comment in `utils/common.h` says both `StringToHelperFast` and `StringToHelper` are kept to maintain bit-for-bit legacy LightGBM behavior for precision, I guess you'd prefer to keep the old behavior by default
    - I noticed this when working on #4089 and some followups. Result verification shows this difference which makes me upset about the correctness

I've also been biten by ClickHouse's choice of non-precise float parsing by default. They choose to sacrifice 1 or 2 bits of precision to keep good float parsing performance https://github.com/ClickHouse/ClickHouse/issues/1665 (3 years ago, not sure whether still true now). For me, I choose to change the float parsing function to precise and the compile the package myself.

tests/benchmark/parser/CMakeLists.txt
cyfdecyf: The benchmark commit is now reverted. It's better to create microbenchmarks for this.

include/LightGBM/utils/common.h
cyfdecyf: Code introduced in #3942 is replaced by `AtofPrecise` in commit 498090dd.

CMakeLists.txt
StrikerRUS: OK, I see now that the main reason is that `AtofPrecise` is much slower than current solution.
But I still strongly against adding new compilation option for this because of maintenance burden and not many users will compile the library on their own to get precise file parsing.
However, I believe that new **config param** will be a good workaround for this situation. Just like recently added (see #3494 and #3578) [`deterministic` param](https://lightgbm.readthedocs.io/en/latest/Parameters.html#deterministic). Users don't have to re-compile the library every time they are switching "performance/accuracy" scenarios but do it in the runtime.
Is it possible to add new param for precise parsing? WDYT? Thanks!

cyfdecyf: Thanks for the feedback. I'll add a config parameter and remove the compile time option to do this.

BTW, `deterministic` parameter is very useful for verifying result when change code in LightGBM.

include/LightGBM/utils/common.h
shiyu1994: Great! LGTM.

shiyu1994 (approved):

cyfdecyf:
I've been busy with other features these days. I'll finish adding new option to enable precise float parsing soon.

shiyu1994:
Sorry, I didn't noticed that @StrikerRUS 's comment hasn't been addressed before approving this.

cyfdecyf:
I rebased this PR to latest master and made a force push.

The latest commit adds new option `precise_float_parser` for dataset parameters.

CMakeLists.txt
shiyu1994: Now that we have the config parameter for precise float parsing, shall we remove this compile option?

shiyu1994 (commented):
Almost done. Just a question about keeping the compilation option for precise float parsing.

cyfdecyf:
@shiyu1994 Sorry I forgot to remove that compile option.

Fix pushed. Thanks for spotting this.

shiyu1994 (approved):

StrikerRUS:
@cyfdecyf Thank you very much for converting compilation option into config one!

Is it possible to write some (very) basic tests to ensure that precise parser is working (including inf, nan, etc.)?
Something like this https://github.com/dmlc/xgboost/blob/master/tests/cpp/common/test_charconv.cc.
Please refer to https://github.com/microsoft/LightGBM/blob/master/tests/cpp_tests/test_chunked_array.cpp as an example of cpp test infrastructure.

cyfdecyf:
@StrikerRUS OK. Hope I can find time to add some tests tomorrow.

cyfdecyf:
@StrikerRUS Test case for `Common::Atof` added. In order to get the `test_chunked_array.cpp`, I did a rebase and force-push.

The tests are mainly for inf, nan. One thing to note is that -NaN parsing may have different behavior as mention before in code review.

Please let me know if there's any other cases that should be added.

tests/cpp_tests/test_common.cpp
```diff
@@ -0,0 +1,104 @@
...
+      { "1E0", 1.0 },
+      { "-1", -1.0 },
+      { "-1E0", -1.0 },
+      { "123456.0", 123456.0 },
+      { "432E1", 432E1 },
+      { "1.2345678", 1.2345678 },
+      { "2.4414062E-4", 2.4414062E-4 },
+      { "3.0540412E5", 3.0540412E5 },
+      { "3.355445E7", 3.355445E7 },
+      { "1.1754944E-38", 1.1754944E-38 },
```
StrikerRUS, line 48: Please add `FLT_MAX`, `DBL_MIN/MAX` and just a few corner cases, for instance [from here](https://github.com/Alexhuszagh/rust-lexical/blob/master/data/test-parse-unittests/strtod_tests.toml).

StrikerRUS (commented):
@cyfdecyf Thank you so so much for finding time to write tests!
Please consider checking a few minor comments below.

StrikerRUS:
Gently ping @AlbertoEAF and @mjmckp as persons recently contributed fixes to this domain. We will really appreciate your feedback for this PR.

AlbertoEAF (changes requested):
LGTM, there are only two things that I request changing, I put a thumbsup in StrikerRus' comments:
1. Revert the change in the chunked array tests
2. Add more tests for numerical corner cases

include/LightGBM/config.h
```diff
@@ -712,6 +712,11 @@ struct Config {
   // desc = **Note**: can be used only in CLI version; for language-specific packages you can use the correspondent function
   bool save_binary = false;
 
+  // [no-save]
```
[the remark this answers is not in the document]
cyfdecyf: My initial motivation for this PR is to ease comparision of prediction result with cmdline and other language (e.g. python). So I didn't thought about saving this parameter in the model file.

To be more specific

1. the model is trained with cmdline
2. prediction is done with both python and cmdline
    - this requires precise float parsing in order to make both prediction have the same result

As this option can also be used for training, I think it's better to include it in the model file. At least this can add information about what float parsing mode is used during train.

I'll remove `no-save`  here.

tests/cpp_tests/test_chunked_array.cpp
```diff
@@ -65,11 +65,6 @@ class ChunkedArrayTest : public testing::Test {
 };
 
 
-/*! ChunkedArray cannot be built from chunks of size 0. */
-TEST_F(ChunkedArrayTest, constructorWithChunkSize0Throws) {
-  ASSERT_THROW(ChunkedArray<int> ca(0), std::runtime_error);
-}
-
```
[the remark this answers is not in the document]
cyfdecyf: Sorry this is a mistake. The fatal message made me thought the test is not running correctly. I'll revert this back.

docs/Parameters.rst
```diff
@@ -820,6 +820,12 @@ Dataset Parameters
 
    -  **Note**: can be used only in CLI version; for language-specific packages you can use the correspondent function
 
+-  ``precise_float_parser`` :raw-html:`<a id="precise_float_parser" title="Permalink to this parameter" href="#precise_float_parser">&#x1F517;&#xFE0E;</a>`, default = ``false``, type = bool
+
+   -  Use precise floating point number parsing for text parser (e.g. CSV, TSV, LibSVM input).
+
+   -  **Note**: setting this to ``true`` may lead to much slower text parsing.
```
StrikerRUS, line 827:
```suggestion
   -  use precise floating point number parsing for text parser (e.g. CSV, TSV, LibSVM input)

   -  **Note**: setting this to ``true`` may lead to much slower text parsing
```

include/LightGBM/config.h
```diff
@@ -714,6 +714,10 @@ struct Config {
   // desc = **Note**: can be used only in CLI version; for language-specific packages you can use the correspondent function
   bool save_binary = false;
 
+  // desc = Use precise floating point number parsing for text parser (e.g. CSV, TSV, LibSVM input).
+  // desc = **Note**: setting this to ``true`` may lead to much slower text parsing.
```
StrikerRUS, line 718:
```suggestion
  // desc = use precise floating point number parsing for text parser (e.g. CSV, TSV, LibSVM input)
  // desc = **Note**: setting this to ``true`` may lead to much slower text parsing
```

StrikerRUS (approved):
LGTM except one nit below to keep all params descriptions in consistent style.

As this parser is not used for model files but only for datasets, I believe there will be no any inconsistency issues with default parser, right?
https://github.com/microsoft/LightGBM/issues/3463#issuecomment-711115057
https://github.com/microsoft/LightGBM/blob/d517ba12f2e7862ac533908304dddbd770655d2b/include/LightGBM/utils/common.h#L314-L321

cyfdecyf:
@StrikerRUS Thank you for taking time to review this PR. The added corner test cases indeed found one problem which is not setting `errno` to 0 before calling `strtod`.

Regarding your question about loading model files, `AtofPrecise` is actually used in some places. For example [`leaf_value_`](https://github.com/microsoft/LightGBM/blob/5014f19f150e5a6ebe236584c52519ee11446f8a/src/io/tree.cpp#L686) parsing will call [`AtofPrecise`](https://github.com/cyfdecyf/LightGBM/blob/precise-text-parse/include/LightGBM/utils/common.h#L1104).

[`__StringToTHelper`](https://github.com/microsoft/LightGBM/blob/5014f19f150e5a6ebe236584c52519ee11446f8a/include/LightGBM/utils/common.h#L1086-L1099) in the latest commit in master branch does not handle input `na`, `infinity` either. I replaced those code with `AtofPrecise` and the behavior should be the same on NaN and Inf parsing.

cyfdecyf:
BTW, why does model loading uses both precise and non-precise version of floating point number parsing function?

For example:

https://github.com/microsoft/LightGBM/blob/5014f19f150e5a6ebe236584c52519ee11446f8a/src/io/tree.cpp#L686
https://github.com/microsoft/LightGBM/blob/5014f19f150e5a6ebe236584c52519ee11446f8a/src/io/tree.cpp#L710

Is this for keeping backward compatibility?

shiyu1994:
The second one is not doing floating point number parsing. It just parses an integer.

StrikerRUS:
@shiyu1994 
> Using `StringToArray` or `StringToArrayFast` should make no difference.

I thought that `StringToArrayFast` is used in places where some information loss in rare cases is acceptable for the aim of speedup. For example, 
https://github.com/microsoft/LightGBM/blob/d517ba12f2e7862ac533908304dddbd770655d2b/src/io/tree.cpp#L751-L761
, https://github.com/microsoft/LightGBM/pull/3938/files.

Was I wrong?

shiyu1994:
@StrikerRUS Sorry, I did not made it clear. I mean because here the value is an integer, so both methods won't cause information loss.

BTW, can we merge this PR?

StrikerRUS:
@shiyu1994 

> I mean because here the value is an integer, so both methods won't cause information loss.

Thanks, got it! But for the linked case with the `double` type in both cases, is my intuition correct?

> BTW, can we merge this PR?

I don't have any objections. I think we can merge if you don't have any comments for code changed after your previous review.

AlbertoEAF:
> Thanks, got it! But for the linked case with the `double` type in both cases, is my intuition correct?

Exactly @StrikerRUS and we should be very careful if we were to switch methods when parsing doubles or we can end up with subtly different model scores just by upgrading LightGBM. Hence the decision at the time when the big model read/write was done, to keep the old bit-for-bit behaviour, independently of the speed of the parsing when switching to the fast_double_parser.

cyfdecyf:
> @StrikerRUS Sorry, I did not made it clear. I mean because here the value is an integer, so both methods won't cause information loss.

I linked to the wrong example in the first place. In fact I was planing to include an example like @StrikerRUS has showed.

shiyu1994:
@StrikerRUS Yes, that's correct. @AlbertoEAF thanks for your explanation. @cyfdecyf  That's OK. We can merge this PR now. Thanks for your contribution.

PR #4087, opened 2021-03-19
store all CMake files in one place

StrikerRUS:
Similarly to #3809 with external dependencies, it is better to store all CMake-related files in one place.


CMakeLists.txt
```diff
@@ -100,7 +100,6 @@ include_directories(${EIGEN_DIR})
 ADD_DEFINITIONS(-DEIGEN_MPL2_ONLY)
 
 if(__BUILD_FOR_R)
-    list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake/modules")
```
StrikerRUS, line 103: Already done at line 35
https://github.com/microsoft/LightGBM/blob/971b5486873339bcaa2bbe0948c95e1fff7246c7/CMakeLists.txt#L35

jameslamb (approved):
seems good to me! Thanks

guolinke (approved):

Issue #3896, opened 2021-02-03
[dask] Support all LightGBM boosting types


StrikerRUS:
## Summary

Right now the Dask interface in https://github.com/microsoft/LightGBM/blob/706f2af7badc26f6ec68729469ec6ec79a66d802/python-package/lightgbm/dask.py  is tested only with `gbdt` (default) boosting type.
https://lightgbm.readthedocs.io/en/latest/Parameters.html#boosting

I believe that parametrization of core tests (`test_classifier`, `test_regressor`, `test_ranker`) with different boosting types (`rf`, `dart`, `goss`) will improve tests and make us confident in Dask module quality.
Please pay attention to that for some boosting types some parameters cannot be left with their default values. For example, `rf` performs the following checks
https://github.com/microsoft/LightGBM/blob/fcfd4132e6d40a22d52023396329c41fd3de4a42/src/boosting/rf.hpp#L35-L36

How tests can be parametrized:
https://github.com/microsoft/LightGBM/blob/ac706e10e461d1c5987a3975cc48aa82fde33566/tests/python_package_test/test_dask.py#L31
https://github.com/microsoft/LightGBM/blob/ac706e10e461d1c5987a3975cc48aa82fde33566/tests/python_package_test/test_dask.py#L166

StrikerRUS:
Closed in favor of being in #2302. We decided to keep all feature requests in one place.

Welcome to contribute this feature! Please re-open this issue (or post a comment if you are not a topic starter) if you are actively working on implementing this feature.

jmoralez:
Hi, I'd like to work on this. I'm just starting out and saw all the tests fail for the other boosting_types, so I digged a bit and I'm ashamed to say I hadn't really seen the testing data we generate and just saw that for regression X is a (100, 100) array which gets split into two partitions, so each worker trains on a (50, 100) collection, which I think isn't very good. I tried generating 5 features with 2 informatives and 1,000 samples, so each worker gets (500, 5) and now all the tests pass. @jameslamb what are your thoughts regarding this? I believe 1,000 samples are ok and I don't think we need that many features, I think the tests should check that the distributed training achieves the same result as local, but not make it overly difficult by providing few samples and many features.

jameslamb:
a data size like (1000, 5) or something is totally fine for tests

jmoralez:
[...]

PR #4088, opened 2021-03-20
[tests][dask] Add voting_parallel algorithm in tests (fixes #3834)

### LINKED ISSUE #3834: Support all LightGBM parallel tree learners in Dask

jmoralez:
This includes the `voting_parallel` `tree_learner` for `test_regressor`, `test_classifier` and `test_ranker` in the tests for the dask module and removes the warning about experimental support that was previously triggered (because it wasn't tested).


tests/python_package_test/test_dask.py
```diff
@@ -44,6 +44,7 @@
 CLIENT_CLOSE_TIMEOUT = 120
 
 tasks = ['binary-classification', 'multiclass-classification', 'regression', 'ranking']
+distributed_training_algorithms = ['data_parallel', 'voting_parallel']
```
jameslamb, line 47: All 4 of these strings are valid for this argument (https://lightgbm.readthedocs.io/en/latest/Parameters.html#tree_learner):

* `data`
* `data_parallel`
* `voting`
* `voting_parallel`

But I don't think it's worth it to have all four choices here because that would create a lot of new tests (`2 * len(tasks) * len(data_output)`).

So could you do the following?

1. Change `distributed_training_algorithms` to the "main" values from https://lightgbm.readthedocs.io/en/latest/Parameters.html#tree_learner (`['data', 'voting_parallel']`)
2. add a new test like `test_training_respects_tree_learner_aliases()` that uses `['data_parallel', 'voting_parallel']`. It can just use `output="array"`, to reduce the total number of tests.

tests/python_package_test/test_dask.py
```diff
@@ -905,22 +910,18 @@ def test_warns_and_continues_on_unrecognized_tree_learner(client):
...
+    X = da.random.random((1_000, 10), chunks=(500, 10))
+    y = da.random.random((1_000, 1), chunks=500)
+    dask_regressor = lgb.DaskLGBMRegressor(
+        client=client,
+        time_out=5,
+        tree_learner='feature_parallel',
+        n_estimators=1,
+        num_leaves=2
+    )
+    with pytest.raises(lgb.basic.LightGBMError):
```
jameslamb, line 923:
```suggestion
    with pytest.raises(lgb.basic.LightGBMError, match="Do not support feature parallel in c api"):
```

Can you please add the specific error message in https://github.com/microsoft/LightGBM/issues/3834#issuecomment-782269920 here? In tests that catch errors, I think it's valuable to check that you got the exact error you expected. Otherwise, unexpected errors introduced by code changes can slip past the tests.

jameslamb (changes requested):
Thanks for doing this! I'm glad it was as easy as just changing the tests. I have a few suggested changes for the tests.

jmoralez:
Hi, James. Do you have any suggestions on what to work on next?

jmoralez:
I noticed a failing test due to graphviz I believe. This time it failed in [Linux_latest_gpu_source](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=9528&view=logs&j=582743ec-83fe-58c4-937e-4197b5816801&t=63b016c8-8965-547a-fab6-395af921c96a&l=628). Here are the errors:
```
Warning: Could not load "/home/AzDevOps_azpcontainer/miniconda/envs/test-env/lib/graphviz/libgvplugin_pango.so.6" - It was found, so perhaps one of its dependents was not.  Try ldd.
Warning: Could not load "/home/AzDevOps_azpcontainer/miniconda/envs/test-env/lib/graphviz/libgvplugin_gd.so.6" - It was found, so perhaps one of its dependents was not.  Try ldd.
```
I believe this test was also failing on some builds before I merged the latest master.

The other one is related to https://github.com/microsoft/LightGBM/pull/4095#issuecomment-804523725.

jameslamb:
> Hi, James. Do you have any suggestions on what to work on next?

are you mostly interested in Dask? If so, this issue would be a good one to pick up next: https://github.com/microsoft/LightGBM/issues/3896. If not, let me know and I can recommend something else.

tests/python_package_test/test_dask.py
jameslamb: looks like this was addressed in https://github.com/microsoft/LightGBM/pull/4088/commits/76b99941a8e8c016be9f375f09fe6d0af8e1108e

tests/python_package_test/test_dask.py
jameslamb: looks like this was addressed in https://github.com/microsoft/LightGBM/pull/4088/commits/76b99941a8e8c016be9f375f09fe6d0af8e1108e

jameslamb (changes requested):
Thanks for making that most recent round of changes. I don't have any other suggestions. Could you please update to the latest `master`?

tests/python_package_test/test_dask.py
```diff
@@ -273,7 +276,7 @@ def test_classifier(output, task, boosting_type, client):
     p2_proba = local_classifier.predict_proba(X)
     s2 = local_classifier.score(X, y)
 
-    if boosting_type == 'rf' and output == 'dataframe-with-categorical':
```
StrikerRUS, line 276: IS this change related to this PR?

tests/python_package_test/test_dask.py
```diff
@@ -930,22 +935,36 @@ def test_warns_and_continues_on_unrecognized_tree_learner(client):
...
-        dask_regressor = lgb.DaskLGBMRegressor(
-            client=client,
-            time_out=5,
-            tree_learner=tree_learner,
-            n_estimators=1,
-            num_leaves=2
-        )
-        with pytest.warns(UserWarning, match='Support for tree_learner %s in lightgbm' % tree_learner):
-            dask_regressor = dask_regressor.fit(X, y)
+@pytest.mark.parametrize('task', tasks)
```
StrikerRUS, line 938: Can we not parametrize this test with `tasks` and not loose code coverage? It will be helpful to save some CI time.

tests/python_package_test/test_dask.py
```diff
@@ -930,22 +935,36 @@ def test_warns_and_continues_on_unrecognized_tree_learner(client):
...
 
-        assert dask_regressor.fitted_
-        assert dask_regressor.get_params()['tree_learner'] == tree_learner
+    assert dask_model.fitted_
+    assert dask_model.get_params()['tree_learner'] == tree_learner
+
+
+def test_error_on_feature_parallel_tree_learner(client):
+    X = da.random.random((1_000, 10), chunks=(500, 10))
+    y = da.random.random((1_000, 1), chunks=500)
```
StrikerRUS, line 958: To save CI time.
```suggestion
    X = da.random.random((100, 10), chunks=(50, 10))
    y = da.random.random((100, 1), chunks=50)
```

StrikerRUS (commented):
Great to know that voting parallel is working without any modifications!
Please consider checking a few my minor comments below:

tests/python_package_test/test_dask.py
jmoralez: Yes. Because of https://github.com/microsoft/LightGBM/issues/4118, rf for classification isn't very reliable. When using `tree_learner='data'` these tests would pass, but with voting they all fail, so I moved all of the rf tests to the "soft check" instead of just the ones for `dataframe-with-categorical`.

tests/python_package_test/test_dask.py
jmoralez: addressed in aa78ea3

tests/python_package_test/test_dask.py
jmoralez: addressed in aa78ea3

jmoralez:
@jameslamb do you know what's the advantage of using the `client` fixture? (Apart from avoiding the client construction). I just tried running `test_regressor` by initializing the client at the top and the time goes from 137s down to 98s.
Here's the profile with the fixture:
Screenshot from 2021-03-30 20-22-14
And without:
image
So basically with the fixture 70% of the time is spent in setting up and tearing down the cluster. Without it only 60% of the time is spent there and seems to be faster.

tests/python_package_test/test_dask.py
StrikerRUS: Got it, thanks for the explanation!

StrikerRUS:
@jmoralez 
> do you know what's the advantage of using the `client` fixture?

Awesome research! Could you please copy-paste your comment here: https://github.com/microsoft/LightGBM/issues/3829#issuecomment-769442538. I believe that place is the best one to continue the discussion.

StrikerRUS (approved):
LGTM!

jameslamb (approved):
thanks for this!

PR #4093, opened 2021-03-22
[python-package] Add type hints to the callback file

deddyjobson:
## What I did
- Added type hints for callback.py as per https://github.com/microsoft/LightGBM/issues/3756.
- Verified that there were no additional errors when running mypy.

## Note
- I ran into a problem from mypy where function attributes raise errors. I implemented a workaround as suggested in the discussion thread (https://github.com/python/mypy/issues/2087).


deddyjobson:
@jameslamb Can you please confirm if I need to resolve these failed tests or if they're not caused by my patch?

jameslamb:
Some of the failures are unrelated to this PR's changes and should be fixed by merging the most recent `master` (https://github.com/microsoft/LightGBM/pull/4095).

However, I think the `check-docs` failure is probably related to this pull request.

> [autosummary] generating autosummary for: Python-API.rst
                                                                                        
> Warning, treated as error:
> [autosummary] failed to import 'lightgbm.Booster': no module named lightgbm.Booster

That error from Sphinx is not very informative but it usually is caused by a legitimate underlying problem, for example https://github.com/microsoft/LightGBM/pull/3192#issuecomment-754231716.

Can you please provide the specific `mypy` error(s) that this PR attempts to fix with `CallbackWithAttributes`?

deddyjobson:
Here is an example of a mypy error that CallbackWithAttributes resolves:

```python-package\lightgbm\callback.py:86: error: "Callable[[CallbackEnv], None]" has no attribute "order"```

I did import an additional package (typing_extensions) which I guess is what caused the problem. Removing the package and one associated type hint resulted in no errors.

File: python-package/lightgbm/callback.py [hunk 1]
```diff
@@ -63,15 +76,16 @@ def print_evaluation(period=1, show_stdv=True):
...
+    def _callback(env: CallbackEnv) -> None:
         if period > 0 and env.evaluation_result_list and (env.iteration + 1) % period == 0:
             result = '\t'.join([_format_eval_result(x, show_stdv) for x in env.evaluation_result_list])
             _log_info('[%d]\t%s' % (env.iteration + 1, result))
     _callback.order = 10
     return _callback
 
 
-def record_evaluation(eval_result):
+def record_evaluation(eval_result: dict) -> CallbackWithAttributes:
```
jameslamb, line 88:
```suggestion
def record_evaluation(eval_result: Dict[str, Dict[str, List[Any]]]) -> CallbackWithAttributes:
```

I think this could be a bit more specific, and being more specific will get us more informative errors from `mypy`.

python-package/lightgbm/callback.py
```diff
@@ -2,14 +2,27 @@
 """Callbacks library."""
 import collections
 from operator import gt, lt
+from typing import Any, Callable, List, Union
 
 from .basic import _ConfigAliases, _log_info, _log_warning
 
 
+class CallbackWithAttributes():
```
jameslamb, line 10:
```suggestion
class _CallbackWithAttributes:
```

Could you please prefix this with `_`? Since this is just an internal detail for type-checking, I don't want to encourage `lightgbm` users to import this in their own code.

python-package/lightgbm/callback.py
```diff
@@ -2,14 +2,27 @@
...
+    """Allows the creation of function attributes without making mypy raise errors."""
+
+    order: int
+    before_iteration: bool
+
+
+def callback_attr_decorator(func: Any) -> CallbackWithAttributes:
```
jameslamb, line 17:
```suggestion
def _callback_attr_decorator(func: Any) -> CallbackWithAttributes:
```

python-package/lightgbm/callback.py
```diff
@@ -2,14 +2,27 @@
...
+    """Prevents function attributes from raising mypy errors."""
```
jameslamb, line 18:
```suggestion
    """Adds more specific type annotation for callback functions.
    
    This decorator informs type checkers like mypy that callback functions have attributes
    like ``before_iteration`` and ``order``.
    """
```

Could you make this more specific? I think this proposed description makes it clearer what is going on in this code.

python-package/lightgbm/callback.py
```diff
@@ -2,14 +2,27 @@
 """Callbacks library."""
 import collections
 from operator import gt, lt
+from typing import Any, Callable, List, Union
 
 from .basic import _ConfigAliases, _log_info, _log_warning
 
 
+class CallbackWithAttributes():
+    """Allows the creation of function attributes without making mypy raise errors."""
```
jameslamb, line 11:
```suggestion
    """Bare type used to tell type checkers which attributes callback functions have"""
```

Similar suggestion here. What do you think about this proposed description? I think that "prevents mypy errors" is too limited of an explanation and doesn't explain exactly what's happening. The immediate reason to add these changes is to address warnings found by `mypy`, but `mypy` is not the only thing in the Python ecosystem that takes advantage of libraries' type hints if they're defined.

jameslamb (changes requested):
Thanks for doing this @deddyjobson ! This one is tricky, really appreciate you working through it.

I left some initial comments, but I'd like @StrikerRUS to comment as well. I'm ok with the approach you've proposed of introducing a decorator to document the fact that we add attributes to the callback functions, I think that's a nice, non-invasive approach to improve the value we get from `mypy`.

python-package/lightgbm/callback.py
deddyjobson: I see. Thanks for the suggestion; the proposed description makes it much clearer.

deddyjobson: I saw this linting error from a failed test. 
```./python-package/lightgbm/callback.py:18 in private function `_callback_attr_decorator`:
        D401: First line should be in imperative mood (perhaps 'Add', not 'Adds')```

It seems to me that we should start the docstring with "Adds" like you suggested and not "Add" for a docstring which describes what the function does. I'll change the first word to "Add" like the error message suggested, but I thought I'd make sure that it's really the right thing to do.

**Review by jameslamb — APPROVED** [review]
Looks good to me! I can see in the `lint` job's logs that there are no `mypy` warnings for `callback.py`, and I think you've covered everything that needs a hint.

https://github.com/microsoft/LightGBM/pull/4093/checks?check_run_id=2207644984

Thanks very much! I won't merge until @StrikerRUS has a chance to look as well.

**Comment by StrikerRUS** [comment]
I'm afraid decorator will hurt performance as callback functions are called on each iteration.

**Comment by deddyjobson** [comment]
Removing the decorator results in 5 additional errors from mypy:
```
python-package\lightgbm\callback.py:71: error: "Callable[[CallbackEnv], None]" has no attribute "order"
python-package\lightgbm\callback.py:102: error: "Callable[[CallbackEnv], None]" has no attribute "order"
python-package\lightgbm\callback.py:142: error: "Callable[[CallbackEnv], None]" has no attribute "before_iteration"
python-package\lightgbm\callback.py:143: error: "Callable[[CallbackEnv], None]" has no attribute "order"
python-package\lightgbm\callback.py:240: error: "Callable[[CallbackEnv], None]" has no attribute "order"
```

**Comment by jameslamb** [comment]
Thanks! I agree with @StrikerRUS and I didn't think about the overhead before. In this first round of adding type hints to the library, I think it's desirable to only make changes that don't negatively affect the performance of `lightgbm`.

Can you please add `# type: ignore` comments for the lines you've mentioned in https://github.com/microsoft/LightGBM/pull/4093#issuecomment-811067338?

**Comment by deddyjobson** [comment]
Cool, didn't know I could do that.

**Comment by jameslamb** [comment]
> Cool, didn't know I could do that.

oh, yep! This is super useful. Python is a really flexible language, so there are a lot of situations that are hard for `mypy` to understand. It's just a static analyzer and doesn't run your code.

By the way, I think there needs to be two spaces between the end of the line and the comment

```python
x = something()  # type: ignore
```

jameslamb (approved):
Thanks very much for the changes! This looks good to me. I can see in https://github.com/microsoft/LightGBM/pull/4093/checks?check_run_id=2242197161 that there are no `mypy` warnings for `callback.py` and I agree with the choices of type hints you made.

StrikerRUS (approved):

jameslamb:
We resolved a few blocking issues in continuous integration today, so I've updated this with the changes from the latest `master`. Will merge once it builds. Thanks again for the help!

PR #4092, opened 2021-03-22
[ci] apply cpplint to cpp tests

StrikerRUS:

PR #4095, opened 2021-03-22
[ci] fix flaky Azure Pipelines jobs

StrikerRUS:
Today I had to re-run Azure builds more than 5 times to make it create new release. Also, many current PRs are failing with 
```
Executing transaction: ...working... g_module_open() failed for /home/AzDevOps_azpcontainer/miniconda/envs/test-env/lib/gdk-pixbuf-2.0/2.10.0/loaders/libpixbufloader-svg.so: libXrender.so.1: cannot open shared object file: No such file or directory

g_module_open() failed for /home/AzDevOps_azpcontainer/miniconda/envs/test-env/lib/gdk-pixbuf-2.0/2.10.0/loaders/libpixbufloader-svg.so: libXrender.so.1: cannot open shared object file: No such file or directory
```
For example, #4093, #4092.

I hope system-wide installation of `libxau` along with `libxrender1` will solve the issue.

Refer to https://github.com/conda-forge/graphviz-feedstock/issues/35#issuecomment-786368065:
> So it is possible to get away without installing `libXau` at the OS level ... but for reliable results, you should either install `libXau` at the OS level, or always explicitly install `xorg-libxau`.

### DIFF
--- a/.ci/setup.sh
+++ b/.ci/setup.sh
@@ -44,6 +44,7 @@ else  # Linux
             libicu66 \
             libssl1.1 \
             libunwind8 \
+            libxau6 \
             locales \
             netcat \
             unzip \
--- a/.ci/setup.sh
+++ b/.ci/setup.sh
@@ -44,7 +44,7 @@ else  # Linux
             libicu66 \
             libssl1.1 \
             libunwind8 \
-            libxau6 \
+            libxau-dev \
             locales \
             netcat \
             unzip \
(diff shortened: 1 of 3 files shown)


jameslamb (approved):
thanks for figuring this out!

jameslamb:
seems quay.io is down :/

https://status.quay.io/#past-incidents

image

That would explain the failing qemu builds

Issue #4108, opened 2021-03-25
[R-package] Warnings in R package tests


jameslamb:
## Description

The R package tests currently have the following warnings which seem problematic. Contributions to fix these would be welcomed!

```text
[LightGBM] [Warning] Unknown parameter: valids
[LightGBM] [Warning] Unknown parameter: 0x0000000016067830>
[LightGBM] [Warning] Unknown parameter: valids
[LightGBM] [Warning] Unknown parameter: 0x0000000016067830>
[LightGBM] [Warning] Unknown parameter: valids
[LightGBM] [Warning] Unknown parameter: 0x0000000016067830>
[LightGBM] [Warning] Unknown parameter: valids
[LightGBM] [Warning] Unknown parameter: 0x0000000016067830>
[LightGBM] [Warning] Unknown parameter: valids
[LightGBM] [Warning] Unknown parameter: 0x0000000016067830>
[LightGBM] [Warning] Unknown parameter: valids
[LightGBM] [Warning] Unknown parameter: 0x0000000016067830>
[LightGBM] [Warning] Unknown parameter: valids
[LightGBM] [Warning] Unknown parameter: 0x0000000016067830>
[LightGBM] [Warning] Unknown parameter: valids
[LightGBM] [Warning] Unknown parameter: 0x0000000016067830>
[LightGBM] [Warning] Unknown parameter: valids
[LightGBM] [Warning] Unknown parameter: 0x0000000016067830>
[LightGBM] [Warning] Unknown parameter: valids
[LightGBM] [Warning] Unknown parameter: 0x0000000016067830>
[LightGBM] [Warning] Unknown parameter: valids
[LightGBM] [Warning] Unknown parameter: 0x0000000016067830>
[LightGBM] [Warning] Unknown parameter: valids
[LightGBM] [Warning] Unknown parameter: 0x0000000016067830>
[LightGBM] [Warning] Unknown parameter: valids
[LightGBM] [Warning] Unknown parameter: 0x0000000016067830>
[LightGBM] [Warning] Unknown parameter: valids
[LightGBM] [Warning] Unknown parameter: 0x0000000016067830>
[LightGBM] [Warning] There are no meaningful features, as all feature values are constant.
[LightGBM] [Warning] Unknown parameter: 2,3
[LightGBM] [Warning] Unknown parameter: 127,3
[LightGBM] [Warning] Unknown parameter: 2,3
[LightGBM] [Warning] Unknown parameter: 2,3
[...]

Issue #4105, opened 2021-03-24
[R-package] test on non-ASCII feature names fails on debian + clang CRAN check


jameslamb:
## Description

`{lightgbm}` 3.2.0 was recently released to CRAN (#3872 , https://cran.r-project.org/web/packages/lightgbm/index.html). One of the CRAN checks is currently failing while running unit tests.

https://cran.r-project.org/web/checks/check_results_lightgbm.html

![image](https://user-images.githubusercontent.com/7608904/112347477-e336cd00-8c94-11eb-8397-c6843b7dcc3d.png)

```text
  == Failed tests ================================================================
  -- Failure (test_basic.R:1233:5): lgb.train() supports non-ASCII feature names --
  dumped_model[["feature_names"]] not identical to `feature_names`.
  4/4 mismatches
  x[1]: "F_é\u009b¶"
  y[1]: "F_<U+96F6>"
  
  x[2]: "F_äž\u0080"
  y[2]: "F_<U+4E00>"
  
  x[3]: "F_äº\u008c"
  y[3]: "F_<U+4E8C>"
  
  x[4]: "F_äž\u0089"
  y[4]: "F_<U+4E09>"
  
  [ FAIL 1 | WARN 0 | SKIP 3 | PASS 633 ]
  Error: Test failures
  Execution halted
checking PDF version of manual ... OK
checking for non-standard things in the check directory ... OK
DONE
Status: 1 ERROR
````

Full logs are too large to post, but they'll be available at https://www.r-project.org/nosvn/R.check/r-devel-linux-x86_64-debian-clang/lightgbm-00check.html for a few days.

## Reproducible example

N/A - this failure happened on CRAN. I haven't had a chance yet to try it on the equivalent R Hub environment.

## Environment info

Details on this test can be found at https://cran.r-project.org/web/checks/check_flavors.html#r-devel-linux-x86_64-debian-clang.

## Additional Comments

`{lightgbm}`'s CRAN package is currently only tested with with `gcc`: https://github.com/microsoft/LightGBM/blob/ab474dc4035416ef512961ea14595a729b55e1ef/.github/workflows/r_package.yml#L104-L108

although the R package built with CMake is tested on Linux with `clang`: https://github.com/microsoft/LightGBM/blob/ab474dc4035416ef512961ea14595a729b55e1ef/.github/workflows/r_package.yml#L33-L42

jameslamb:
Some additional notes:

[...]

PR #4109, opened 2021-03-25
[R-package] fix typo in linear learner test

jameslamb:
This fixes a typo in params on one of the R unit tests, which led to the warning noted in https://github.com/microsoft/LightGBM/issues/4105#issuecomment-806259143

> [LightGBM] [Warning] Unknown parameter: categorical_featurs

That test specifically checks that training a model in R using linear models at the leaves works if you also have categorical features. So this typo might have actually prevented that test from running correctly and covering what it was expected to.

I also looked through the logs from a recent build of the `r-package (windows-2019, MSVC, R 4, cmake)` CI job. Those jobs use `Rscript testthat.R` instead of `R CMD CHECK`, so you can see all of the logs from the tests. For the other warnings that look problematic, I've opened #4108 with the tag `good first issue` so that it might be addressed by other contributors.

### DIFF
--- a/R-package/tests/testthat/test_basic.R
+++ b/R-package/tests/testthat/test_basic.R
@@ -1883,7 +1883,7 @@ test_that("lgb.train() works with linear learners when Dataset has categorical f
     , metric = "mse"
     , seed = 0L
     , num_leaves = 2L
-    , categorical_featurs = 1L
+    , categorical_features = 1L
   )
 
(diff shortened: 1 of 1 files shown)


StrikerRUS (approved):
Thanks! Will merge after re-runnning Azure due to Clang installation error and our current "best friend" `104` socket error.

PR #4111, opened 2021-03-26
[ci] attach archives with complete source code of the repo to releases

StrikerRUS:
Refer to https://github.com/microsoft/LightGBM/pull/3872#issuecomment-804087682.

I believe that adding archives with repo source code to artifacts that we produce for each commit in `master` is pointless, so I'm not adding `PublishBuildArtifacts@1` step for archives.


jameslamb (approved):
oh excellent, thank you!

Issue #3518, opened 2020-11-02
[RFC] Docker Hub subscription


StrikerRUS:
Docker has introduced paid subscriptions and has limited pulls for users.
https://www.docker.com/blog/scaling-docker-to-serve-millions-more-developers-network-egress/

We use Docker in our Azure Pipelines with our own Docker image from https://github.com/guolinke/lightgbm-ci-docker: 
https://github.com/microsoft/LightGBM/blob/cf69591b68c68a78ea4064e6941b8270b8a00eef/.vsts-ci.yml#L15-L16
and for some R builds 
https://github.com/microsoft/LightGBM/blob/cf69591b68c68a78ea4064e6941b8270b8a00eef/.github/workflows/r_package.yml#L166
https://github.com/microsoft/LightGBM/blob/cf69591b68c68a78ea4064e6941b8270b8a00eef/.github/workflows/r_valgrind.yml#L13

With new Docker subscription model we have the following.

Free tier allows **100 pulls per 6 hours**: https://www.docker.com/pricing.

Some special options are available for Open Source projects but requires applying Google Form and not listed publicly: https://www.docker.com/community/open-source/application.

To be honest, I'm not sure whether 100 pulls are enough for us as some CIs use Docker internally even if we do not specify containers and require Docker jobs/actions/steps. Also, it is very unclear for me how CI can "log in"  to use 200 free pulls instead of default 100 ones. 
> Authenticated users (Docker ID)

For example, NVIDIA suggests migrating to `NGC` from Docker Hub for CUDA Docker users: https://twitter.com/NVIDIAAI/status/1323133675417051145.

jameslamb:
Since this is a Microsoft project, can we use [Azure Container Registry](https://azure.microsoft.com/en-us/services/container-registry/) ? We could push the images that we care about there.

> it is very unclear for me how CI can "log in" to use 200 free pulls instead of default 100 ones.

You can use the [`docker login`](https://docs.docker.com/engine/reference/commandline/login/) command to make authenticated requests to a container registry.

StrikerRUS:
Hmm. sounds reasonable! But what about images that we do not maintain and want to use the latest version of them? I'm speaking about R helper images mostly.

Anyway, I think we can wait a little bit and see what 100 pulls per 6 hours mean in real life. Also, it worth to keep an eye on open source community movements.

jameslamb:
[...]

PR #4059, opened 2021-03-10
[dask] raise more informative error for duplicates in 'machines' (fixes #4057)

jameslamb:
As of #3994 , it's possible to give the estimators in `lightgbm.dask` an exact set of worker IPs + ports to use for distributed training. This is done by providing a parameter `machines`, a string with a comma-delimited list like `ip1:port1,ip2:port2`.

If you accidentally include duplicates in this list, the error you get today is not very informative:

>  KeyError: 'pop from an empty set'

This PR proposes catching this case and throwing a more informative error.

## Notes for Reviewers

In this case, "duplicates" refers to duplicate `ip-port` pairs. It is possible and very common to run multiple Dask workers on one physical machine. In those cases, it's `machines` might look like this

```text
ip1:12400
ip1:12405
ip1:12475
```

In #4057, I mentioned that it was probably possible to even further reduce the risk of the type of job failure that cause #4057. This could be done with something like this: https://github.com/microsoft/LightGBM/pull/4059/commits/9442bdf00f193a19a923dc0deb46b7822cb6f601.

However, after thinking about it, I now think we shouldn't do that. If a change results in this error happening a lot more frequently than I expect it to (1 in ever 4 billion runs), then we should see CI failing more frequently. I think it's desirable for that to happen so we know something is wrong.

PR #4112, opened 2021-03-26
[ci] bump CUDA version and pull dockers from NVIDIA NGC

StrikerRUS:
Why switch to NVIDIA NGC? Due to #3518.

### DIFF
--- a/.github/workflows/cuda.yml
+++ b/.github/workflows/cuda.yml
@@ -86,7 +86,7 @@ jobs:
             $ROOT_DOCKER_FOLDER/.ci/setup.sh || exit -1
             $ROOT_DOCKER_FOLDER/.ci/test.sh || exit -1
             EOF
-            docker run --env-file docker.env -v "$GITHUB_WORKSPACE":"$ROOT_DOCKER_FOLDER" --rm --gpus all "nvidia/cuda:${{ matrix.cuda_version }}-devel" /bin/bash $ROOT_DOCKER_FOLDER/docker-script.sh
+            docker run --env-file docker.env -v "$GITHUB_WORKSPACE":"$ROOT_DOCKER_FOLDER" --rm --gpus all "nvcr.io/nvidia/cuda:${{ matrix.cuda_version }}-devel" /bin/bash $ROOT_DOCKER_FOLDER/docker-script.sh
   all-successful:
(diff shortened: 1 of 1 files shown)


jameslamb:
ugh I'm surprised to see this error from https://github.com/microsoft/LightGBM/issues/4057 again (error message was introduced in https://github.com/microsoft/LightGBM/pull/4059)

https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=9565&view=logs&j=c28dceab-947a-5848-c21f-eef3695e5f11&t=fa158246-17e2-53d4-5936-86070edbaacf

```text
machines = '127.0.0.1:53229,127.0.0.1:53229'
worker_addresses = dict_keys(['tcp://127.0.0.1:43843', 'tcp://127.0.0.1:42213'])

    def _machines_to_worker_map(machines: str, worker_addresses: List[str]) -> Dict[str, int]:
        """Create a worker_map from machines list.
    
        Given ``machines`` and a list of Dask worker addresses, return a mapping where the keys are
        ``worker_addresses`` and the values are ports from ``machines``.
    
        Parameters
        ----------
        machines : str
            A comma-delimited list of workers, of the form ``ip1:port,ip2:port``.
        worker_addresses : list of str
            A list of Dask worker addresses, of the form ``{protocol}{hostname}:{port}``, where ``port`` is the port Dask's scheduler uses to talk to that worker.
    
        Returns
        -------
        result : Dict[str, int]
            Dictionary where keys are work addresses in the form expected by Dask and values are a port for LightGBM to use.
        """
        machine_addresses = machines.split(",")
    
        if len(set(machine_addresses)) != len(machine_addresses):
>           raise ValueError(f"Found duplicates in 'machines' ({machines}). Each entry in 'machines' must be a unique IP-port combination.")
E           ValueError: Found duplicates in 'machines' (127.0.0.1:53229,127.0.0.1:53229). Each entry in 'machines' must be a unique IP-port combination.
```

StrikerRUS:
I have already seen this error one time after #4059. Guess this is somehow related to the initial seed of pseudo-random generator.
I believe more fair test from https://github.com/microsoft/LightGBM/issues/4057#issuecomment-794889388 will be to run `lgb.dask._find_random_open_port()` with Python restart.

jameslamb (approved):
oh really good idea to switch to NGC, thanks

PR #4113, opened 2021-03-26
[tests][dask] Create an informative categorical feature

jmoralez:
This modifies the `dataframe-with-categorical` output type to create only two categorical features and makes one of them informative for the task. 

In order to have more consistent results I've modified the `_create_data` function to generate 1,000 samples instead of 100, which caused problems with categoricals and reduced the accuracy of the distributed training. Also, for regression 100 features were being generated, I changed this to generate only 4, with 2 informative.

Adding more samples allows the tolerances in the array checks to be lower, but I had to increase a bit the complexity of the models for `test_regressor` and `test_classifier`.

I didn't change anything in `_create_ranking_data` because I haven't seen problems with those tests, but could take a look as well.

### DIFF
--- a/tests/python_package_test/test_dask.py
+++ b/tests/python_package_test/test_dask.py
@@ -131,7 +131,7 @@ def _create_ranking_data(n_samples=100, output='array', chunk_size=50, **kwargs)
     return X, y, w, g_rle, dX, dy, dw, dg
 
 
-def _create_data(objective, n_samples=100, output='array', chunk_size=50, **kwargs):
+def _create_data(objective, n_samples=1_000, output='array', chunk_size=500, **kwargs):
     if objective.endswith('classification'):
         if objective == 'binary-classification':
             centers = [[-4, -4], [4, 4]]
@@ -141,7 +141,7 @@ def _create_data(objective, n_samples=100, output='array', chunk_size=50, **kwar
             raise ValueError(f"Unknown classification task '{objective}'")
         X, y = make_blobs(n_samples=n_samples, centers=centers, random_state=42)
     elif objective == 'regression':
-        X, y = make_regression(n_samples=n_samples, random_state=42)
+        X, y = make_regression(n_samples=n_samples, n_features=4, n_informative=2, random_state=42)
     elif objective == 'ranking':
         return _create_ranking_data(
             n_samples=n_samples,
@@ -161,7 +161,7 @@ def _create_data(objective, n_samples=100, output='array', chunk_size=50, **kwar
     elif output.startswith('dataframe'):
         X_df = pd.DataFrame(X, columns=['feature_%d' % i for i in range(X.shape[1])])
         if output == 'dataframe-with-categorical':
-            num_cat_cols = 5
+            num_cat_cols = 2
             for i in range(num_cat_cols):
                 col_name = "cat_col" + str(i)
                 cat_values = rnd.choice(['a', 'b'], X.shape[0])
@@ -172,13 +172,15 @@ def _create_data(objective, n_samples=100, output='array', chunk_size=50, **kwar
                 X_df[col_name] = cat_series
                 X = np.hstack((X, cat_series.cat.codes.values.reshape(-1, 1)))
 
-            # for the small data sizes used in tests, it's hard to get LGBMRegressor to choose
-            # categorical features for splits. So for regression tests with categorical features,
-            # _create_data() returns a DataFrame with ONLY categorical features
+            # make one categorical feature relevant to the target
+            cat_col_is_a = X_df['cat_col0'] == 'a'
             if objective == 'regression':
-                cat_cols = [col for col in X_df.columns if col.startswith('cat_col')]
-                X_df = X_df[cat_cols]
-                X = X[:, -num_cat_cols:]
+                y = np.where(cat_col_is_a, y, 2 * y)
+            elif objective == 'binary-classification':
+                y = np.where(cat_col_is_a, y, 1 - y)
+            elif objective == 'multiclass-classification':
+                n_classes = 3
+                y = np.where(cat_col_is_a, y, (1 + y) % n_classes)
         y_df = pd.Series(y, name='target')
         dX = dd.from_pandas(X_df, chunksize=chunk_size)
         dy = dd.from_pandas(y_df, chunksize=chunk_size)
@@ -238,8 +240,8 @@ def test_classifier(output, task, client):
     )
 
     params = {
-        "n_estimators": 10,
-        "num_leaves": 10
+        "n_estimators": 50,
+        "num_leaves": 31
     }
 
     dask_classifier = lgb.DaskLGBMClassifier(
@@ -265,7 +267,7 @@ def test_classifier(output, task, client):
     assert_eq(p1, p2)
     assert_eq(y, p1)
     assert_eq(y, p2)
-    assert_eq(p1_proba, p2_proba, atol=0.3)
+    assert_eq(p1_proba, p2_proba, atol=0.01)
     assert_eq(p1_local, p2)
     assert_eq(y, p1_local)
 
@@ -407,7 +409,8 @@ def test_regressor(output, client):
 
     params = {
         "random_state": 42,
-        "num_leaves": 10
+        "num_leaves": 31,
+        "n_estimators": 20,
     }
 
     dask_regressor = lgb.DaskLGBMRegressor(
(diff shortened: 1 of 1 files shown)


tests/python_package_test/test_dask.py
```diff
@@ -432,9 +434,8 @@ def test_regressor(output, client):
     p2 = local_regressor.predict(X)
 
     # Scores should be the same
-    if not output.startswith('dataframe'):
-        assert_eq(s1, s2, atol=.01)
-        assert_eq(s1, s1_local, atol=.003)
+    assert_eq(s1, s2, atol=.01)
+    assert_eq(s1, s1_local, atol=0.003)
```
jameslamb, line 438: when you tested, did you try running these tests multiple times in a row? I'd like to know how close these tolerances are to the maximum difference that you saw. It's good to push these tolerances down to catch big differences between distributed training and non-distributed training, but not so close to the edge that the tests might fail sometimes due to randomness in how the chunks are distributed across the workers.

jameslamb (commented):
overall, looks great! I totally support this change, and it's ok to not touch the ranker tests for now. I'm glad it allowed for some simplifications, and will enable us to go deeper on testing the statistical behavior of the Dask module (https://github.com/microsoft/LightGBM/issues/3896#issuecomment-807877365).

I looked at the logs to see if making the models more complex had a big impact on total runtime of the tests. It doesn't seem like it!

See the results below for the `linux_regular` job. A 20 second difference is not problematic given the run time of this project's CI jobs. Just comparing one job isn't a precise estimate, but it's enough to give me confidence that this change won't 

**latest build on master ([link](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=9554&view=logs&j=c28dceab-947a-5848-c21f-eef3695e5f11&t=fa158246-17e2-53d4-5936-86070edbaacf))**

```text
===== 388 passed, 7 skipped, 2 xfailed, 357 warnings in 696.97s (0:11:36) ======
```

**this PR ([link](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=9567&view=logs&j=c28dceab-947a-5848-c21f-eef3695e5f11&t=fa158246-17e2-53d4-5936-86070edbaacf))**

```text
===== 388 passed, 7 skipped, 2 xfailed, 357 warnings in 720.08s (0:12:00) ======
```

tests/python_package_test/test_dask.py
```diff
@@ -432,9 +434,8 @@ def test_regressor(output, client):
     p2 = local_regressor.predict(X)
 
     # Scores should be the same
-    if not output.startswith('dataframe'):
-        assert_eq(s1, s2, atol=.01)
-        assert_eq(s1, s1_local, atol=.003)
+    assert_eq(s1, s2, atol=.01)
```
StrikerRUS, line 437:
```suggestion
    assert_eq(s1, s2, atol=0.01)
```

StrikerRUS (approved):

tests/python_package_test/test_dask.py
jmoralez: For the test above the maximum absolute difference I observed was 0.0025, so 0.01 should be fine. For this one I believe we can even remove the atol, since this checks the score for the same model, doesn't it?

jameslamb: great, thank you. Yes, I think you're right. `s1` and `s1_local` should be identical since they're from the same model.

jmoralez: I played around with the checks for the predictions and the target and found that for `atol=50` we could go to `rtol=0.5` (the maximum observed was 0.25). But maybe we could instead check `p1` vs `p2`? They're very similar so maybe that'd be a better check and could go very low in the tolerances as well. Also I switched the order in the comparison for this one because `assert_eq(a, b)` checks `abs(a - b) <= atol + rtol * abs(b)`, so I think it makes more sense that `b = y` in this case. What do you think?

jameslamb: ok I'm a little confused by your notation, sorry. Are you referring to this change specifically?

image

To be honest, I don't understand what you mean. Can you give a small reproducible example that shows `assert_eq()` passing or failing based only on the order that you pass things in?

To me, it looks like the checks in that function are symmetric: https://github.com/dask/dask/blob/913a192fd3b35696883bdd4c4eeba2418b946b3a/dask/dataframe/utils.py#L841-L847.

jose-moralez: Haha, I'm sorry. Yes, I meant those checks exactly. The asymmetric part is in https://github.com/dask/dask/blob/913a192fd3b35696883bdd4c4eeba2418b946b3a/dask/dataframe/utils.py#L847, which calls [`np.allclose`](https://github.com/numpy/numpy/blob/fb215c76967739268de71aa4bda55dd1b062bc2e/numpy/core/numeric.py#L2256), and [`np.isclose`](https://github.com/numpy/numpy/blob/fb215c76967739268de71aa4bda55dd1b062bc2e/numpy/core/numeric.py#L2346)

 Here's an example:
```python
import numpy as np

a = 100
b = 120
np.allclose(a, b, rtol=0.1, atol=9)  # (20 - 9) / 0.1 = 110 < 120 (True)
np.allclose(b, a, rtol=0.1, atol=9)  # (20 - 9) / 0.1 = 110 < 100 (False)
```

jose-moralez:
There's a failing test in [linux regular](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=9573&view=logs&j=c28dceab-947a-5848-c21f-eef3695e5f11&t=fa158246-17e2-53d4-5936-86070edbaacf&l=3384). I've seen this happen a couple of times now, the environment [finds conflicts](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=9573&view=logs&j=c28dceab-947a-5848-c21f-eef3695e5f11&t=fa158246-17e2-53d4-5936-86070edbaacf&l=1386) and then [gives up](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=9573&view=logs&j=c28dceab-947a-5848-c21f-eef3695e5f11&t=fa158246-17e2-53d4-5936-86070edbaacf&l=1602).

tests/python_package_test/test_dask.py
jameslamb: :exploding_head: I had no idea that these were not symmetric! Thanks for the example. I see now that there's a note about this in the `np.isclose()` docs:

> Unlike the built-in `math.isclose`, the above equation is not symmetric
    in `a` and `b` -- it assumes `b` is the reference value -- so that
    `isclose(a, b)` might be different from `isclose(b, a)`

Ok, since this is checking that we get a "good enough" model result, I agree with the change to treat `y` (the actual value of the target) as the "reference value". Thanks for the attention to detail.

jameslamb: also now I know @jmoralez and @jose-moralez are the same person :stuck_out_tongue:

jameslamb (approved):
Looks great to me, thanks very much for the suggestions.

jameslamb:
> There's a failing test in [linux regular](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=9573&view=logs&j=c28dceab-947a-5848-c21f-eef3695e5f11&t=fa158246-17e2-53d4-5936-86070edbaacf&l=3384). I've seen this happen a couple of times now, the environment [finds conflicts](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=9573&view=logs&j=c28dceab-947a-5848-c21f-eef3695e5f11&t=fa158246-17e2-53d4-5936-86070edbaacf&l=1386) and then [gives up](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=9573&view=logs&j=c28dceab-947a-5848-c21f-eef3695e5f11&t=fa158246-17e2-53d4-5936-86070edbaacf&l=1602).

yeah, I've seen this a few times too :/

We try not to use pinned dependencies for Python packages in this project's CI, to be sure the project is tested against new releases. Sometimes this can result in failures like that where `conda` thinks it cannot resolve an environment. Usually 1 rebuild fixes this, and I just restarted that job.

tests/python_package_test/test_dask.py
jose-moralez: haha, yes. This is the account I use at work. I usually contribute on my free time using my personal account, but feel free to ping me on this one at work hours.

PR #4117, opened 2021-03-26
[ci] build CRAN R-package on Azure with every commit and attach it to releases

StrikerRUS:
Contributes to #3283.

Right now maintainer has to type `/gha run build-r-artifacts` to prepare R-package for submission to CRAN. This PR moves CRAN package building from GitHub Actions to Azure Pipelines where we create all other artifacts.

New job takes about 1 minute, so it doesn't significantly impact overall CI time. But with this job it is possible to attach CRAN artifact to releases automatically (refer to https://github.com/microsoft/LightGBM/pull/3872#issuecomment-804094123).

image

image

@jameslamb Could you please help to summarize what else should be done to close issue #3283?


jameslamb (approved):
This is awesome, thanks!

> @jameslamb Could you please help to summarize what else should be done to close issue #3283?

I think the only manual part left is updating `configure.ac`. It's possible for GitHub Actions to submit pull requests, so I think it's possible to create a workflow where you could, for example, run `/gha run update-r-configure`,  where it would run something like the `docker run` command in https://github.com/microsoft/LightGBM/tree/8cc6eefcefb398ce5ae4f0884e12c918c9e4f1e4/R-package#changing-the-cran-package.

However, to be honest I consider that a "nice to have". It only takes me (or anyone with access to docker) about 10 seconds to update `configure` on release branches and push a new commit. So I'm ok with closing #3283 as of this PR!

I know #3283 mentioned creating binaries but I don't think LightGBM should do that, now that the R package is on CRAN and CRAN prepares binaries automatically.

StrikerRUS:
> I know #3283 mentioned creating binaries but I don't think LightGBM should do that, now that the R package is on CRAN and CRAN prepares binaries automatically.

Totally support this! Let's not confuse users with a lot of binaries, especially with ones that differ from CRAN binaries.


> It's possible for GitHub Actions to submit pull requests ...

OK, good idea! I think I can implement this, but this will be limited to PRs only within our repo and will not work from forked personal repos (I learned it during working on #3740). I believe this is OK because `configure.ac` is updated only with releases which are done by maintainers with write access to the main repo.

Issue #4118, opened 2021-03-27
rf boosting type for classification


jmoralez:
## Description

I found that the rf boosting type struggles for classification tasks, even trivial ones. I'd like to know how its `predict_proba` method works, I'd expect it to be the proportion of trees that vote for each class, but I think the implementation may be derived from boosting.

## Reproducible example

```python
import lightgbm as lgb
import matplotlib.pyplot as plt
import numpy as np
from sklearn.datasets import make_blobs

centers = [[-4, -4], [4, 4]]
X, y = make_blobs(n_samples=100, centers=centers, random_state=42)
plt.scatter(X[:, 0], X[:, 1], c=y)
```
This is the data:
![Screenshot from 2021-03-26 22-41-29](https://user-images.githubusercontent.com/8473587/112710122-77b35200-8e84-11eb-8f76-f43b43501644.png)

So a single split at 0 does the job, which is what the trees do:
```python
params = {
    'boosting_type': 'rf',
    'n_estimators': 2,
    'bagging_freq': 1,
    'bagging_fraction': 0.8
}
clf = lgb.LGBMClassifier(**params).fit(X, y)
use_cols = ['tree_index', 'node_depth', 'split_feature', 'threshold', 
            'split_gain', 'decision_type', 'value', 'count']
print(clf.booster_.trees_to_dataframe()[use_cols].to_markdown(index=False))
```
|   tree_index |   node_depth | split_feature   |   threshold |   split_gain | decision_type   |   value |   count |
|-------------:|-------------:|:----------------|------------:|-------------:|:----------------|--------:|--------:|
|            0 |            1 | Column_0        |       1e-35 |       82.988 | <=              |       0 |      83 |
|            0 |            2 |                 |     nan     |      nan     |                 |      -2 |      42 |
|            0 |            2 |                 |     nan     |      nan     |                 |       2 |      41 |
|            1 |            1 | Column_0        |       1e-35 |       79.95  | <=              |       0 |      80 |
|            1 |            2 |                 |     nan     |      nan     |                 |      -2 |      41 |
|            1 |            2 |                 |     nan     |      nan     |                 |       2 |      39 |

However, the probabilities don't reflect that:
```python
prob = clf.predict_proba(X)
prob[:5]
# array([[0.11920292, 0.88079708],
#        [0.88079708, 0.11920292],
#        [0.11920292, 0.88079708],
#        [0.88079708, 0.11920292],
#        [0.11920292, 0.88079708]])
```
[...]

PR #4119, opened 2021-03-27
[tests][dask] test all boosting types (fixes #3896)

### LINKED ISSUE #3896: [dask] Support all LightGBM boosting types

jmoralez:
This includes all boosting types (gbdt, rf, goss, dart) in `test_classifier`, `test_regressor` and `test_ranker`

### DIFF
--- a/tests/python_package_test/test_dask.py
+++ b/tests/python_package_test/test_dask.py
@@ -254,7 +254,6 @@ def test_classifier(output, task, boosting_type, client):
     elif boosting_type == 'goss':
         params['top_rate'] = 0.5
 
-
     dask_classifier = lgb.DaskLGBMClassifier(
         client=client,
         time_out=5,
--- a/tests/python_package_test/test_dask.py
+++ b/tests/python_package_test/test_dask.py
@@ -278,12 +278,13 @@ def test_classifier(output, task, boosting_type, client):
         assert_eq(s1, s2, atol=0.01)
         assert_eq(p1_proba, p2_proba, atol=0.8)
     else:
+        assert_eq(s1, s2)
         assert_eq(p1, p2)
+        assert_eq(p1, y)
+        assert_eq(p2, y)
         assert_eq(p1_proba, p2_proba, atol=0.03)
-        assert_eq(y, p1)
-        assert_eq(y, p2)
         assert_eq(p1_local, p2)
-        assert_eq(y, p1_local)
+        assert_eq(p1_local, y)
 
     # pref_leaf values should have the right shape
     # and values that look like valid tree nodes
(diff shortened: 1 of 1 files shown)


jameslamb (approved):
This looks great, thanks so much for the contribution amd for documenting the issues you faced with random forest mode in #4118. @StrikerRUS are you ok with this increase in the number of tests? I think these are providing valuable new coverage and are worth it, but want to get your opinion.

tests/python_package_test/test_dask.py
```diff
@@ -263,13 +273,17 @@ def test_classifier(output, task, client):
...
-    assert_eq(y, p2)
-    assert_eq(p1_proba, p2_proba, atol=0.01)
-    assert_eq(p1_local, p2)
-    assert_eq(y, p1_local)
+    if boosting_type == 'rf' and output == 'dataframe-with-categorical':
+        # https://github.com/microsoft/LightGBM/issues/4118
+        assert_eq(s1, s2, atol=0.01)
+        assert_eq(p1_proba, p2_proba, atol=0.8)
+    else:
+        assert_eq(p1, p2)
```
StrikerRUS, line 281: Why are we dropping `assert_eq(s1, s2)` from the `else` branch?

StrikerRUS (commented):
> @StrikerRUS are you ok with this increase in the number of tests? I think these are providing valuable new coverage and are worth it, but want to get your opinion.

Thank you. I support this and think this is valuable.
Just one question below.

tests/python_package_test/test_dask.py
jmoralez: That was totally my bad, I lost it when splitting the statements. I've included it in 
5d7001d

jameslamb: thanks for catching it @StrikerRUS ! I missed it too

StrikerRUS (approved):
Great, many thanks!

PR #4121, opened 2021-03-27
Update test_chunked_array to please cpplint

AlbertoEAF:
This should fix the remaining cpplint issues in test_chunked_array.

This also cleans up a bit the last test.

Requested by @StrikerRUS in https://github.com/microsoft/LightGBM/pull/4092.


StrikerRUS:
Thank you very much!

PR #4123, opened 2021-03-27
[ci] remove output parametrization from two Dask tests

StrikerRUS:
Save CI time.

I think these tests are for specific network params and we don't test something principally new by feeding different data types.


jameslamb (approved):
oooo yes I agree! Nice reduction in the tests without losing coverage, thank you

PR #4127, opened 2021-03-28
update CODEOWNERS

jameslamb:
This PR proposes some changes to CODEOWNERS based on my experience of the last year with reviews in this project. My hope is that this change will reduce the number of times that we have to manually add reviewers, and will reduce notifications for other maintainers who don't need to be on some PRs.

For example, I have no idea why we had the rule "only @Laurae2 is added by default if `.vsts-ci.yml` is touched" haha.


StrikerRUS (approved):
Thanks for the update!

Issue #4099, opened 2021-03-23
[Dask] Expected error randomly not raised in Dask test


StrikerRUS:
```
2021-03-23T11:32:00.3655481Z         error_msg = "has multiple Dask worker processes running on it"
2021-03-23T11:32:00.3656048Z         with pytest.raises(lgb.basic.LightGBMError, match=error_msg):
2021-03-23T11:32:00.3656573Z >           dask_model3.fit(dX, dy, group=dg)
2021-03-23T11:32:00.3657297Z E           Failed: DID NOT RAISE <class 'lightgbm.basic.LightGBMError'>
```

https://github.com/microsoft/LightGBM/blob/77d54b3d22e6b55371e25f962f14d00d50f9e55c/tests/python_package_test/test_dask.py#L1058-L1060

Refer to https://github.com/microsoft/LightGBM/pull/4068#issuecomment-798972044 and https://github.com/microsoft/LightGBM/pull/4068#issuecomment-804857749 for full logs.

jameslamb:
hmmm interesting. In the logs mentioned in those comments, it looks like this is a different root cause from what was fixed in #4071. I think that' what's happening here is that the data is still all ending up on one worker somehow. This is possibly the same underlying problem as https://github.com/microsoft/LightGBM/issues/4074, actually.

Error code 104 means "connection reset by peer" ([link](https://www-numi.fnal.gov/offline_software/srt_public_context/WebDocs/Errors/unix_system_errors.html)), which could occur in distributed training if one of the Dask workers dies and is restarted.

Similarly here, if one of the workers died before training started, then it's possible that Dask would have moved the training data back to the other worker, and that then `dask_model3.fit()` was fitting on only a single worker.

It's possible that one of the workers died because the two previous `.fit()` calls before this one left behind objects that caused it to run out of memory or caused other errors on the workers.

There's no reason that this test has to be in the same test case as the other network params tests. I just did that to try to minimize the total runtime of tests (number of times we call `_create_data()`). I think that moving this to its own standalone test case might reduce the risk of this specific failure.

StrikerRUS:
> It's possible that one of the workers died because the two previous `.fit()` calls before this one left behind objects that caused it to run out of memory or caused other errors on the workers.

```
client = <Client: 'tcp://127.0.0.1:36289' processes=2 threads=2, memory=16.70 GB>
```
I believe 16Gb should be enough for toy datasets we use for tests...

[...]

PR #4132, opened 2021-03-29
[dask] run one training task on each worker

jameslamb:
This PR resolves a critical source of instability in `lightgbm.dask`, which I think might be the cause of some of the issues we have observed sporadically in tests (#4057, #4074, #4099).

## Changes in this PR

* specifies arguments in `lightgbm.dask`'s use of [`distributed.Client.submit()`](https://distributed.dask.org/en/latest/api.html#distributed.Client.submit) to more tightly control how Dask schedules LightGBM training tasks

## How this improves `lightgbm`

I believe this PR will fix a critical source of instability that can cause training to fail (either throwing an error or hanging indefinitely) if multiple training tasks are accidentally scheduled onto the same worker. I think that the risk of this issue increases as you have more workers in the Dask cluster and as there is more other work going on in the cluster alongside LightGBM training.

This PR should also improve the stability of LightGBM's Dask unit tests.

## Overview

In distributed LightGBM training, users create `n` workers, each of which runs LightGBM training on a local chunk of the full training data. Estimators in `lightgbm.dask` expect training data to be given in [Dask collections](https://docs.dask.org/en/latest/delayed-collections.html). Those estimators look at which Dask workers hold each partition of the training data, and run one LightGBM training process per worker.

The logic for scheduling those training processes out onto Dask workers is currently incorrect.

https://github.com/microsoft/LightGBM/blob/9cab93a9f3c0b46301a2cd11631223f4ab1daf16/python-package/lightgbm/dask.py#L383-L397

The current code follows this pseudocode, which is like

> **"given `n` workers with data, run `n` training processes on the cluster"**

```python
for work in workers:
    client.submit(
        _train_part
    )
```

However, the desired behavior is slightly different. It should be

> **"given `n` workers with data, run exactly 1 training process on each worker"**

```python
for work in workers:
    client.submit(
        _train_part,
        workers=worker
    )
```

This more specific definition is important. If the training data are split across two workers (for example) and LightGBM sets `machines` to a list with those two workers, but then two `_train_part()` tasks are scheduled on the same worker, training will fail.

* if both scheduled onto the worker that is `rank 0`, the worker will be killed and restarted.
    - > distributed.nanny - INFO - Worker process 23623 was killed by signal 11
* if both scheduled onto a worker that is not `rank 0`, I saw that LightGBM just hangs indefinitely. If you have task-level timeouts set in Dask, Dask might eventually kill that process. Either way, training will not succeed.

Per [the Dask docs](https://distributed.dask.org/en/latest/api.html#distributed.Client.submit), the following parameters should be used in `lightgbm.dask` to achieve tight control over where tasks run.

* `workers`: specify EXACTLY which worker(s) the task being submitted can run on
* `allow_other_workers` (default: ~True~ False): indicate whether it is ok to run the task on other workers not mentioned in `workers`
* `pure` (default: True): whether or not the function is a "pure function", which [according to Wikipedia](https://en.wikipedia.org/wiki/Pure_function), means that it always returns the same outputs given the same inputs, does not modify any global state (including the file system), and does not have side effects on output streams

## More Background

The bug fixed by this PR is a general class of problem in `lightgbm.dask` that we should be aware of. By default, Dask assumes that tasks can be scheduled onto any worker, and even that t[hey can be "stolen" from one worker by another](https://distributed.dask.org/en/latest/work-stealing.html)).

These default assumptions don't hold for distributed training of LightGBM or XGBoost.

## Notes for Reviewers

I checked the blame in `dask/distributed`, and it looks like the three keyword arguments to `client.submit()` that I'm specifying here have been a part of the client for at least 3 years. So I don't think they cause any compatibility issues we need to worry about.

https://github.com/dask/distributed/blame/1b32bd30201ef6ced5029180143d2c37b393b586/distributed/client.py#L1234-L1240

### DIFF
--- a/python-package/lightgbm/dask.py
+++ b/python-package/lightgbm/dask.py
@@ -391,6 +391,8 @@ def _train(
             num_machines=num_machines,
             time_out=params.get('time_out', 120),
             return_model=(worker == master_worker),
+            workers=[worker],
+            allow_other_workers=False,
             **kwargs
         )
         for worker, list_of_parts in worker_map.items()
--- a/python-package/lightgbm/dask.py
+++ b/python-package/lightgbm/dask.py
@@ -380,6 +380,13 @@ def _train(
     num_machines = len(worker_address_to_port)
 
     # Tell each worker to train on the parts that it has locally
+    #
+    # This code treates _train_part()` calls as not "pure" because:
+    #     1. there is randomness in the training process unless parameters ``seed``
+    #        and ``is_deterministic`` are set
+    #     2. even with those parameters set, the output of one ``_train_part()`` call
+    #        relies on global state (it and all the other LightGBM training processes
+    #        coordinate with each other)
     futures_classifiers = [
         client.submit(
             _train_part,
@@ -393,6 +400,7 @@ def _train(
             return_model=(worker == master_worker),
             workers=[worker],
             allow_other_workers=False,
+            pure=False,
             **kwargs
         )
         for worker, list_of_parts in worker_map.items()
--- a/python-package/lightgbm/dask.py
+++ b/python-package/lightgbm/dask.py
@@ -381,7 +381,7 @@ def _train(
 
     # Tell each worker to train on the parts that it has locally
     #
-    # This code treates _train_part()` calls as not "pure" because:
+    # This code treates ``_train_part()`` calls as not "pure" because:
     #     1. there is randomness in the training process unless parameters ``seed``
     #        and ``is_deterministic`` are set
     #     2. even with those parameters set, the output of one ``_train_part()`` call


jameslamb:
@ffineis and @jmoralez , if you have time this week could you take a look at this and check if my Dask understanding is correct?

ffineis:
Awesome!! Wow. Yes I'll review this tomorrow PM.

jameslamb:
welp, `test_model_and_local_version_are_picklable_whether_or_not_client_set_explicitly` on `linux_latest regular` still just failed with the following exception:

> Exception: LightGBMError('Socket send error, code: 104',)

https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=9613&view=logs&j=275189f9-c769-596a-7ef9-49fb48a9ab70&t=3a9e7a4a-04e6-52a0-67ea-6e8f6cfda74f

HOWEVER, I still think this PR fixes a critical source of instability. Maybe we just have multiple stability problems in the `lightgbm.dask` / the Dask tests :joy:

jameslamb:
Added #4133 which I think should fix the "sometimes we get a `machines` list with duplicate ports" problem.

Pushing an empty commit to re-trigger CI, but leaving the build from https://github.com/microsoft/LightGBM/pull/4132#issuecomment-809066586 undisturbed in case anyone wants to look at the logs.

jameslamb:
The most recent `cuda 11.2.2 source (linux, gcc, Python 3.7)` build failed with what looks like a temporary issue from the main anaconda channel

```text
Collecting package metadata (current_repodata.json): ...working... failed
                                                                                         
CondaHTTPError: HTTP 502 BAD GATEWAY for url <https://repo.anaconda.com/pkgs/r/noarch/current_repodata.json>
Elapsed: 00:00.056802
CF-RAY: 637a79459a561382-YVR

A remote server error occurred when trying to retrieve this URL.

A 500-type error (e.g. 500, 501, 502, 503, etc.) indicates the server failed to
fulfill a valid request.  The problem may be spurious, and will resolve itself if you
try your request again.  If the problem persists, consider notifying the maintainer
of the remote server.
```

I've manually restarted that build.

StrikerRUS (approved):
@jameslamb Your detailed description makes a lot of sense!

> `allow_other_workers (default: True): indicate whether it is ok to run the task on other workers not mentioned in `workers`

But it looks like this param is already `False` by default:

image

jameslamb:
> But it looks like this param is already False by default:

It is, but I think we should be explicit about that. In case it wasn't clear, the lack of `workers=[worker]` was the actual BUG fixed here. The other two parameters (`pure` and `allow_other_workers`) added are just for completeness, so you can look at that call and understand exactly how the training tasks should be treated.

jameslamb:
oh I see I wrote "Default: True` in the PR description. Edited :joy:

PR #4135, opened 2021-03-29
valid_sets should equal lgb_eval in line 53

benmuhlmann:


examples/python-guide/advanced_example.py
```diff
@@ -50,7 +50,7 @@
 gbm = lgb.train(params,
                 lgb_train,
                 num_boost_round=10,
-                valid_sets=lgb_train,  # eval training data
+                valid_sets=lgb_eval,  # eval training data
```
jameslamb, line 53: Thanks for the contribution! But this line is not a mistake. For the first training in this example, the goal is to print evaluation results on the training data to get an idea of how the fit is improving at each iteration.

The comment `# eval training data` was left here to try to say "yes, it's intentional that the training set is being used".

Do you think there is some rewording of that comment that would make it clearer?

jameslamb (changes requested):

jameslamb:
I'm going to close this for now based on https://github.com/microsoft/LightGBM/pull/4135#discussion_r603551281. @benmuhlmann thanks again for your interest in LightGBM, and please come back and contribute any time!

Issue #4131, opened 2021-03-28
[R Package] autoconf.ac OpenMP test ignores LDFLAGS (fix provided).


jlisic:
## Description
The LDFLAGS R variable "R CMD config LDFLAGS" is not included as part of the OpenMP configure.ac test in the R package.  This causes the compilation to not find OpenMP libraries due to ignoring the -L directories in .R/Makevars.  This is a larger issue on the Mac where -L is usually needed to find the OpenMP libraries.  Furthermore, since LDFLAGS are used for the package compilation, it would be a more reasonable test if LDFLAGS was used.

I wrote a fixed autoconf.ac, but I"m not sure how you would want to merge it due to the need to re-run autoconf.

## Reproducible example

From the Mac 

#### Current version does not look at the LDFLAGS when compiling the OpenMP test, therefore the test fails
/opt/homebrew/Cellar/llvm/11.1.0/bin/clang -o conftest conftest.c -lomp -Xclang -fopenmp

#### After adding the changes to autoconf.ac and running autoconf I was able to fix the issue and pass the OpenMP test by adding ${LDFLAGS} to the OpenMP test.

https://github.com/jlisic/LightGBM/commit/565dd56627b304392b21f205de2fa33069bc08cb#diff-9363f71bbcd7d64bb47ed62564d0bada0c8073646aaeecfe838068aa5bb17d2d

/opt/homebrew/Cellar/llvm/11.1.0/bin/clang -o conftest conftest.c -L/usr/local/lib -L/opt/homebrew/lib -lomp -L/opt/homebrew/opt/llvm/lib -Wl,-rpath,/opt/homebrew/opt/llvm/lib -lomp -Xclang -fopenmp

## Environment info

LightGBM version or commit hash: 3.2.0.99

M1  Mac running big-sur (R 4.0.3.). 
home-brew llvm 11.1.0 + libomp

cd lightgbm
R CMD INSTALL .

#### ~/.R/Makevars

CC=/opt/homebrew/Cellar/llvm/11.1.0/bin/clang
CXX=/opt/homebrew/Cellar/llvm/11.1.0/bin/clang++
CXX11=/opt/homebrew/Cellar/llvm/11.1.0/bin/clang++
CXX14=/opt/homebrew/Cellar/llvm/11.1.0/bin/clang++
CXX17=/opt/homebrew/Cellar/llvm/11.1.0/bin/clang++
[...]

PR #4140, opened 2021-03-30
[ci] use GitHub Actions to re-generate R configure

### LINKED ISSUE #3283: [R-package] automate release process for R

StrikerRUS:
Closed #3283.

Refer to https://github.com/microsoft/LightGBM/pull/4117#pullrequestreview-622620015 for the initial discussion.

Refer to https://github.com/StrikerRUS/LightGBM/pull/1 for the usage example.

~Push from a workflow won't trigger our regular workflows:~
> ~For example, if a workflow run pushes code using the repository's `GITHUB_TOKEN`, a new workflow will not run even when the repository contains a workflow configured to run when `push` events occur.~
  ~https://docs.github.com/en/actions/reference/events-that-trigger-workflows#triggering-new-workflows-using-a-personal-access-token~

~I don't think that this a problem for us, because anyway we need a follow-up commit with reverting back to `~~VERSION~~` in `configure.ac`, right?~

### DIFF
--- a/.github/workflows/r_configure.yml
+++ b/.github/workflows/r_configure.yml
@@ -1,0 +1,38 @@
+name: R generate configure
+
+on:
+  repository_dispatch:
+    types: [gha_run_r_configure]
+
+jobs:
+  r-configure:
+    name: r-configure
+    timeout-minutes: 60
+    runs-on: ubuntu-latest
+    container: "ubuntu:20.04"
+    steps:
+      - name: Install essential software before checkout
+        run: |
+          apt-get update
+          apt-get install --no-install-recommends -y \
+            ca-certificates \
+            git
+      - name: Checkout repository
+        uses: actions/checkout@v2.3.4
+        with:
+          fetch-depth: 5
+          submodules: true
+          repository: microsoft/LightGBM
+          ref: "refs/heads/${{ fromJSON(github.event.client_payload.pr_branch) }}"
+          token: ${{ secrets.WORKFLOW }}
+          persist-credentials: true
+      - name: Update configure
+        shell: bash
+        run: ./R-package/recreate-configure.sh || exit -1
+      - name: Push changes
+        run: |
+          git config --global user.name "GitHub Actions Bot"
+          git config --global user.email "githubactionsbot@users.noreply.github.com"
+          git add "./R-package/configure"
+          git commit --allow-empty -m "Auto-update configure"
+          git push
--- a/R-package/README.md
+++ b/R-package/README.md
@@ -308,7 +308,7 @@ At build time, `configure` will be run and used to create a file `Makevars`, usi
 
 3. Edit `src/Makevars.in`.
 
-Alternatively, GitHub Actions can re-generate this file for you. On a pull request (only on internal one, not works for ones from forks), create a comment with this phrase:
+Alternatively, GitHub Actions can re-generate this file for you. On a pull request (only on internal one, does not work for ones from forks), create a comment with this phrase:
 
 > /gha run r-configure
 
--- a/R-package/README.md
+++ b/R-package/README.md
@@ -308,6 +308,10 @@ At build time, `configure` will be run and used to create a file `Makevars`, usi
 
 3. Edit `src/Makevars.in`.
 
+Alternatively, GitHub Actions can re-generate this file for you. On a pull request (only on internal one, not works for ones from forks), create a comment with this phrase:
+
+> /gha run r-configure
+
 **Configuring for Windows**
 
 At build time, `configure.win` will be run and used to create a file `Makevars.win`, using `Makevars.win.in` as a template.
--- a/.github/workflows/triggering_comments.yml
+++ b/.github/workflows/triggering_comments.yml
@@ -32,3 +32,11 @@ jobs:
           "${{ github.event.issue.pull_request.url }}" \
           "${{ github.event.comment.id }}" \
           "gha_run_r_solaris"
+
+    - name: Trigger update R configure
+      if: github.event.comment.body == '/gha run r-configure'
+      run: |
+        $GITHUB_WORKSPACE/.ci/trigger_dispatch_run.sh \
+          "${{ github.event.issue.pull_request.url }}" \
+          "${{ github.event.comment.id }}" \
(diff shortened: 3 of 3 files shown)


.github/workflows/r_configure.yml
```diff
@@ -0,0 +1,38 @@
...
+          ref: "refs/heads/${{ fromJSON(github.event.client_payload.pr_branch) }}"
+      - name: Update configure
+        shell: bash
+        run: ./R-package/recreate-configure.sh || exit -1
+      - name: Push changes
+        run: |
+          git config --global user.name "GitHub Actions Bot"
+          git config --global user.email "githubactionsbot@users.noreply.github.com"
+          git add "./R-package/configure"
+          git commit --allow-empty -m"Auto-update configure"
```
jameslamb, line 37:
```suggestion
          git commit --allow-empty -m "Auto-update configure"
```

jameslamb (changes requested):
>  because anyway we need a follow-up commit with reverting back to ~~VERSION~~ in configure.ac, right?

No, no followup commit is required because there is never a commit that removes the `~~VERSION~~` in `configure.ac`. I'm confused, how is `configure.ac` getting updated and showing up in the diff for https://github.com/StrikerRUS/LightGBM/pull/2? Looking at the workflow template, I don't understand how `configure.ac` is getting changed OR how that change is getting committed.

Notice it was unchanged in the diff for the most recent release: https://github.com/microsoft/LightGBM/pull/3872/files, and `recreate-configure.sh` does that templating with a **temporary** `configure.ac`: https://github.com/microsoft/LightGBM/blob/f879018b50c2b323ca0409fb5cc338e67ff648f4/R-package/recreate-configure.sh#L25-L26.

StrikerRUS:
> I'm confused, how is `configure.ac` getting updated and showing up in the diff for StrikerRUS#2?

Just like it is suggested in https://github.com/microsoft/LightGBM/tree/master/R-package#changing-the-cran-package (**Linux or Mac** section):
1. edit `configure.ac`
2. comment `/gha run r-configure`
3. revert step `#1`

How this process should be done right?

jameslamb:
"Edit `configure.ac`" there is not about changing the version. It's about making *functional* changes, for example like the one suggested in #4131.

If the only thing that is changing is the version of LightGBM, you do not need to touch `configure.ac` at all.

StrikerRUS:
> If the only thing that is changing is the version of LightGBM, you do not need to touch `configure.ac` at all.

OK, got it. What should be touched then?

jameslamb:
> > If the only thing that is changing is the version of LightGBM, you do not need to touch `configure.ac` at all.
> 
> OK, got it. What should be touched then?

you don't need to manually edit any files. Just running `recreate-configure.sh` will create a new `configure` script.

StrikerRUS:
> you don't need to manually edit any files.

Ah, seems I'm getting what's going on! Just manually edit `VERSION.txt`, right?

jameslamb:
Yep! Just manually edit `VERSION.txt` if that's changing. Which I think is ok.

Sorry if the instructions weren't clear. When I wrote https://github.com/microsoft/LightGBM/tree/master/R-package#changing-the-cran-package , the main audience I had in mind was contributors making changes like #4131, not our usual maintenance like doing a new release. If there are changes to the language there that you think would make it more clear, I'd be happy to review those as part of this PR.

StrikerRUS:
OK, @jameslamb please take one more look. I hope I understood it right this time.
Here is new demo: https://github.com/StrikerRUS/LightGBM/pull/1.

R-package/README.md
```diff
@@ -308,6 +308,10 @@ At build time, `configure` will be run and used to create a file `Makevars`, usi
 
 3. Edit `src/Makevars.in`.
 
+Alternatively, GitHub Actions can re-generate this file for you. On a pull request (only on internal one, not works for ones from forks), create a comment with this phrase:
```
jameslamb, line 311: small suggested change to this language

```suggestion
Alternatively, GitHub Actions can re-generate this file for you. On a pull request (only on internal one, does not work for ones from forks), create a comment with this phrase:
```

.github/workflows/r_configure.yml
```diff
@@ -0,0 +1,38 @@
...
+          token: ${{ secrets.WORKFLOW }}
+          persist-credentials: true
+      - name: Update configure
+        shell: bash
+        run: ./R-package/recreate-configure.sh || exit -1
+      - name: Push changes
+        run: |
+          git config --global user.name "GitHub Actions Bot"
+          git config --global user.email "githubactionsbot@users.noreply.github.com"
+          git add "./R-package/configure"
```
jameslamb, line 36: I can tell your main OS is Windows because of your defensive use of quotes :stuck_out_tongue:

jameslamb (approved):
This looks great, thanks! The diff in https://github.com/StrikerRUS/LightGBM/pull/1/files#diff-dadfc37f082f218f0e2f0adfc1affc7c55044bcdf8dff110aab9cdfda842d6f3 looks exactly like what I'd expect. Just one minor grammar comment, but I think the workflow is perfect.

I think it is ok that the commit from this bot won't trigger a new CI build. Usually on release PRs (the main place where we update `configure`) there are additional commits after updating `configure` that don't have any impact on `configure`.

For cases like #4131 , this will have to be something that we "just know" to re-run if the contributor doesn't manually regenerate `configure`. I don't think that's problematic at all, as that file very rarely receives updates that are not part of the release process: https://github.com/microsoft/LightGBM/commits/master/R-package/configure.

StrikerRUS:
Thanks!

> looks exactly like what I'd expect. 

Great!

> I think it is ok that the commit from this bot won't trigger a new CI build

It **does** with the latest code! 🙂 Thanks to `token` and `persist-credentials` arguments of `actions/checkout`.


image


> For cases like #4131 , ...

Yeah, unfortunately, GitHub Actions cannot push into forked repos. So we can't do anything with it.

PR #4145, opened 2021-03-30
[python] added f-strings to python-package/setup.py

NovusEdge:


python-package/setup.py
```diff
@@ -183,7 +183,7 @@ def compile_cpp(use_mingw=False, use_gpu=False, use_cuda=False, use_mpi=False,
                 arch = "Win32" if bit32 else "x64"
                 vs_versions = ("Visual Studio 16 2019", "Visual Studio 15 2017", "Visual Studio 14 2015")
                 for vs in vs_versions:
-                    logger.info("Starting to compile with %s (%s).", vs, arch)
+                    logger.info(f"Starting to compile with %s (%s).", vs, arch)
```
StrikerRUS, line 186:
```suggestion
                    logger.info(f"Starting to compile with {vs} ({arch}).")
```

StrikerRUS (changes requested):
One mistake below, everything else looks fine!

NovusEdge:
> One mistake below, everything else looks fine!

Could you please mention the line where the mistake is.

jameslamb:
> > One mistake below, everything else looks fine!
> 
> Could you please mention the line where the mistake is.

it is right here: https://github.com/microsoft/LightGBM/pull/4145#discussion_r604399154

StrikerRUS:
@NovusEdge 
> Could you please mention the line where the mistake is.

GitHub review comments are posted with a small visible piece of code. Most often, the last line of that codepiece is the line review comment is referring to.

image


Also, GitHub has recently introduced "Review suggestions". So that reviewers now can suggest code changes directly in the review comments:

image

Contributors, in case they are agree with a suggestion, are able to commit the suggested change directly in the browser by clicking "Commit suggestion" button or commit multiple suggestion at one time ("Add suggestion to batch" button).

StrikerRUS (approved):
@NovusEdge Thanks for your contribution!

Issue #4026, opened 2021-02-27
[dask] Distributed training sometimes produces very high leaf values


jose-moralez:
## Description

Hi. I'm trying to use LightGBM in time series forecasting and it works fine with local models. However, when using dask the predictions sometimes are huge (my time series are in the range [0, 160] and the predictions are 10^38). I've included a reproducible example (I'm sorry it's not that minimal but I couldn't reproduce it with some regular datasets). By running the example below sometimes the values are kind of big but other times they just really explode. With bigger datasets I've seen that the first tree is pretty much the same between local and dask but by the second tree the values in dask become increasingly large (so it may have something to do with the gradients). I haven't been able to track down the issue and was hoping someone could point me in the right direction.

## Reproducible example

```python
from itertools import chain
from math import ceil, log10

import dask.dataframe as dd
import lightgbm as lgb
import numpy as np
import pandas as pd
from dask.distributed import Client

def generate_daily_series(n_series: int, 
                          min_length: int = 50, 
                          max_length: int = 500):
    series_lengths = np.random.randint(min_length, max_length+1, n_series)
    total_length = series_lengths.sum()
    n_digits = ceil(log10(n_series))
    
    dates = pd.date_range('2000-01-01', periods=max_length, freq='D').values
    uids = [[f'id_{i:0{n_digits}}'] * serie_length for i, serie_length in enumerate(series_lengths)]
    ds = [dates[-serie_length:] for serie_length in series_lengths]
    y = np.arange(7)[np.arange(total_length) % 7] + np.random.rand(total_length) * 0.5
    series = pd.DataFrame({
        'unique_id': list(chain.from_iterable(uids)),
        'ds': list(chain.from_iterable(ds)),
        'y': y
    })
    series = series.set_index('unique_id')
    return series

def generate_features(df, lags, expanding, rolling,
                      ewm_alpha, date_features):
    df = df.copy()
    for lag in lags:
        df[f'lag_{lag}'] = df.y.shift(lag)
    grouped_lag1 = df.groupby('unique_id')['lag_1']
    for op in expanding:
        df[f'expanding_{op}'] = grouped_lag1.transform(
            lambda y: y.expanding().agg(op)
        )
    for op in rolling:
        df[f'rolling_{op}'] = grouped_lag1.transform(
            lambda y: y.rolling(7).agg(op)
        )
    dates = df.ds.dt
    for feature in date_features:
        df[feature] = getattr(dates, feature)
    df = df.dropna()
    return df

[...]

PR #4149, opened 2021-04-01
[tests][dask] Increase number of partitions in data

jmoralez:
This increases the default partitions of the collections returned by `_create_data` to 20. The purpose is to make it less likely that a worker gets all the partitions and be more confident that distributed training is being performed across all tests.


tests/python_package_test/test_dask.py
```diff
@@ -198,7 +198,7 @@ def _create_data(objective, n_samples=1_000, output='array', chunk_size=500, **k
 
 def _r2_score(dy_true, dy_pred):
     numerator = ((dy_true - dy_pred) ** 2).sum(axis=0, dtype=np.float64)
-    denominator = ((dy_true - dy_pred.mean(axis=0)) ** 2).sum(axis=0, dtype=np.float64)
+    denominator = ((dy_true - dy_true.mean(axis=0)) ** 2).sum(axis=0, dtype=np.float64)
```
jameslamb, line 201: I expect the other changes in this PR to generate a bit of discussion, given that you said you saw other tests failing as a result of them. Please make this `_r2_score()` fix a separate PR so it doesn't get stuck waiting on that discussion to be resolved.

jameslamb:
Linking the discussion this came from: https://github.com/microsoft/LightGBM/issues/3829#issuecomment-811591145

tests/python_package_test/test_dask.py
```diff
@@ -273,7 +273,7 @@ def test_classifier(output, task, boosting_type, client):
     p2_proba = local_classifier.predict_proba(X)
     s2 = local_classifier.score(X, y)
 
-    if boosting_type == 'rf' and output == 'dataframe-with-categorical':
```
StrikerRUS, line 276: I believe this part has already been removed in #4088. Please sync with the latest `master`.

tests/python_package_test/test_dask.py
jameslamb: fixed in #4151

tests/python_package_test/test_dask.py
```diff
@@ -255,7 +255,7 @@ def test_classifier(output, task, boosting_type, tree_learner, client):
             'bagging_fraction': 0.9,
         })
     elif boosting_type == 'goss':
-        params['top_rate'] = 0.5
+        params['top_rate'] = 0.7
```
jameslamb, line 258: it looks like this was added since I last reviewed (https://github.com/microsoft/LightGBM/pull/4149/commits/981084f5360b55daae97fe54d0679e0a12fc1703). Can you please explain why it's necessary?

jmoralez: `test_classifier` became flaky in this PR. I assume it's because previously we weren't performing distributed training or at least not everytime, so adding this generated some fails in multiclass classification for [data_parallel-dart](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=9683&view=logs&j=275189f9-c769-596a-7ef9-49fb48a9ab70&t=3a9e7a4a-04e6-52a0-67ea-6e8f6cfda74f&l=533), [voting_parallel-rf](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=9683&view=logs&j=275189f9-c769-596a-7ef9-49fb48a9ab70&t=3a9e7a4a-04e6-52a0-67ea-6e8f6cfda74f&l=779) (this one is very surprising, given that the atol is 0.8), [voting_parallel-gbdt](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=9683&view=logs&j=e3619e06-e539-5956-1ae7-e8c2c29d8e91&t=4b3d6e5d-f8c3-577f-235f-b68c5ed4d3ab&l=1569), [voting_parallel-dart](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=9683&view=logs&j=e3619e06-e539-5956-1ae7-e8c2c29d8e91&t=4b3d6e5d-f8c3-577f-235f-b68c5ed4d3ab&l=1800), [voting_parallel-goss](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=9683&view=logs&j=b463fb3a-b487-5cfc-fc07-6d216464ba86&t=c16b9cdb-c4c8-59f9-2ae2-b81f62c3a9d0&l=547). Most of them are for dataframe with categoricals but there are a couple with sparse matrices. I have to debug them to see what's actually happening, this is a very simple classification problem and I'd expect to get a perfect score with little effort. I'll ping you here once I'm done but it could take a bit haha.

jameslamb: got it, thanks! Let me know if you need any help

jameslamb:
@jmoralez we fixed a few more CI issues today (#4168 , #4158 , #4167 ). Whenever you return to this, please update to the latest `master`. Sorry for the inconvenience.

jmoralez:
Will do. I'm actually looking into this right now, seems to be related to the amount of data each worker gets. So both get data using more partitions but I believe it may not be balanced sometimes and those times the tests fail because adding a `client.rebalance()` and bumping up the threshold for the probas from 0.03 to 0.05 makes the tests pass. What are your thoughts about using rebalance here?

jameslamb:
> Adding a rebalance kind of beats the purpose of this PR haha. Let me just keep moving the chunksize a bit, 100 looks promising.

Yeah, exactly haha. You could also try increasing `n_samples` from 1e3 to 1e4 or something! It's possible that you're running into some problems that are more severe with tiny amounts of data. `pytest` will report the total timing so you should be able to see (and share with reviewers) the impact of using more data on runtime.

jameslamb:
And the same goes for increasing `n_estimators` or `num_leaves`. If training a slightly larger model improves the stability of the tests, that's totally fine.

jmoralez:
It's very strange because it's always the same point that gets a lower score and it doesn't seem logical. The local model gives 99.8% and the distributed one 93.4%.
image
The red dot is the one that causes the test to fail.

jameslamb:
Sorry, I don't understand the axes in that plot or what you mean by "the local model gives 99.8%".

jmoralez:
Haha, sorry. This is a zoom of the lower left section of the data that gets generated for classification. These points all correspond to the same class (centered at [-4, -4]), the axes are the continuous features. The percentages are the probabilities that each model gives to the class (class 1 in this case). It seems strange that the red dot gets a lower probability given that it's not that far away from the center and there are others further away.

tests/python_package_test/test_dask.py
jameslamb: > I'd expect to get a perfect score with little effort

Given the small dataset sizes we use in tests, I think it would be useful to set  `min_data_in_leaf: 0` everywhere. That might improve the predictability of the results.

jmoralez: Sorry this is taking so long, I haven't had much time and I'm really confused by this. The same data point makes the test fail even for `data_parallel` and `gbdt`, I'm trying to figure out what's exactly going on here, I have the test in a while loop and it eventually fails because of that data point, I'm not sure what's wrong with it haha.

jmoralez: Btw, setting `min_data_in_leaf=0` gives this error: `LightGBMError: Check failed: (best_split_info.right_count) > (0) at /hdd/github/LightGBM/src/treelearner/serial_tree_learner.cpp, line 663`. Do you think this could be related to #4026? This data is shuffled but I think forcing few samples in a leaf gives more chance to getting an empty split in one of the workers.

jmoralez: Here's `abs(local_probas - dask_probas)` per iteration for `data_parallel` `gbdt` for just that one sample (index 377):
image
So from the 7th iteration onwards the probabilities start to increasingly differ, I think there's definitely something strange going on here.

jmoralez:
@jameslamb I just tried this again today and that single point still makes the test fail. Should I close this? Or I can try to make a notebook for you to debug and maybe you can find something else, I'm not sure if #4220 is the reason or if it's something else.

jameslamb:
> @jameslamb I just tried this again today and that single point still makes the test fail. Should I close this? Or I can try to make a notebook for you to debug and maybe you can find something else, I'm not sure if #4220 is the reason or if it's something else.

so weird! Thanks for all your investigation so far.

Could you merge the latest `master` into this branch? I can pick it up from here and see if I find anything else. Sometimes when you've been staring at the same problem for this long, it just requires a second set of eyes. I want to see if we can figure this out, because I think there's a chance we'll uncover a bug in distributed training similar to #4026.

jmoralez:
I have a notebook that I've been using for this, I can maybe upload it here. Do you think that'd help you?

jmoralez:
I uploaded my notebook [here](https://cloud.coiled.io/jmoralez/notebooks), I forgot to specify the cpus so it only has two but changing the `threads_per_worker` to 1 in the `Client` allows you to replicate the issue there.

jameslamb:
Perfect, thanks!

Issue #4046, opened 2021-03-05
[python-package] `init_score` and data structures in custom functions shape for multiclass classification


jmoralez:
## Description

When using `init_score` in multiclass classification it would be very intuitive to use an `(n_samples, n_classes)` collection, as is suggested in https://github.com/microsoft/LightGBM/issues/2595#issuecomment-558561739, however this isn't currently supported in the python package, so you have to reshape this to an `(n_samples * n_classes, )` collection.

This probably isn't that big of a deal with local models, however since dask partitions the collections by rows, this could produce problems with the partitioning.

## Reproducible example

```python
import lightgbm as lgb
import numpy as np

n_samples = 100
n_features = 2
n_classes = 3

X = np.random.rand(n_samples, n_features)
y = np.random.randint(0, n_classes, n_samples)
init_score = np.random.rand(n_samples, n_classes)
lgb.LGBMClassifier().fit(X, y, init_score=init_score)
```

```
Traceback (most recent call last):
  File "init_score.py", line 7, in <module>
    lgb.LGBMClassifier().fit(X, y, init_score=init_score)
  File "/hdd/github/LightGBM/python-package/lightgbm/sklearn.py", line 895, in fit
    callbacks=callbacks, init_model=init_model)
  File "/hdd/github/LightGBM/python-package/lightgbm/sklearn.py", line 688, in fit
    callbacks=callbacks, init_model=init_model)
  File "/hdd/github/LightGBM/python-package/lightgbm/engine.py", line 228, in train
    booster = Booster(params=params, train_set=train_set)
  File "/hdd/github/LightGBM/python-package/lightgbm/basic.py", line 2229, in __init__
    train_set.construct()
  File "/hdd/github/LightGBM/python-package/lightgbm/basic.py", line 1472, in construct
    categorical_feature=self.categorical_feature, params=self.params)
  File "/hdd/github/LightGBM/python-package/lightgbm/basic.py", line 1294, in _lazy_init
    self.set_init_score(init_score)
  File "/hdd/github/LightGBM/python-package/lightgbm/basic.py", line 1843, in set_init_score
    init_score = list_to_1d_numpy(init_score, np.float64, name='init_score')
  File "/hdd/github/LightGBM/python-package/lightgbm/basic.py", line 165, in list_to_1d_numpy
    "It should be list, numpy 1-D array or pandas Series".format(type(data).__name__, name))
TypeError: Wrong type(ndarray) for init_score.
[...]

PR #4150, opened 2021-04-01
[python-package] Support 2d collections as input for `init_score` in multiclass classification task

jmoralez:
This intends to solve #4046 by allowing an `(n_samples, n_classes)` collection to be provided as `init_score` in `lightgbm.LGBMClassifier`'s `fit` method.

I'd also like investigate the possibility to allow `grad` and `hess` to be 2d collections as well for custom objectives.

I'm opening this to get some feedback on my current approach.


jmoralez:
@StrikerRUS I'd appreciate your feedback on this when you have time.

StrikerRUS:
@jmoralez Thanks a lot for picking this up!
As an early feedback I can say that I support the way you do it. But
- I'd better replace `try/except` with complex `if ***-1d` check
- `self.params` is not reliable way to retrieve any info (for example, refer to #2613). If we were in a `Booster` class, number of classes should be gotten via https://lightgbm.readthedocs.io/en/latest/C-API.html#c.LGBM_BoosterGetNumClasses C API function.  But we are in `Dataset` class, so I guess we can just blindly flatten the data and allow cpp code deal with wrong resulted data points.

jmoralez:
Thank you. I've removed the try/except and replaced it with checks for the dimension of the input collection and I flatten without looking for the classes. Please let me know what you think.

jmoralez:
@StrikerRUS should I keep going with this?

StrikerRUS:
@jmoralez Sorry I've missed this PR! Yes, I think more intuitive shape of arguments is good enhancement. Will get back to this PR soon.

tests/python_package_test/test_basic.py
```diff
@@ -8,7 +8,7 @@
 from sklearn.model_selection import train_test_split
 
 import lightgbm as lgb
-from lightgbm.compat import PANDAS_INSTALLED, pd_Series
+from lightgbm.compat import PANDAS_INSTALLED, pd_DataFrame, pd_Series
```
StrikerRUS, line 11: Update docstring here
https://github.com/microsoft/LightGBM/blob/237ac299fcffa2b6ed9c68917dfdf0767e88e59c/python-package/lightgbm/basic.py#L1073

tests/python_package_test/test_basic.py
```diff
@@ -8,7 +8,7 @@
...
```
StrikerRUS, line 12: Update docstring here
https://github.com/microsoft/LightGBM/blob/237ac299fcffa2b6ed9c68917dfdf0767e88e59c/python-package/lightgbm/basic.py#L1490

tests/python_package_test/test_basic.py
```diff
@@ -8,7 +8,7 @@
...
 
 from .utils import load_breast_cancer
```
StrikerRUS, line 13: Should we reshape `init_score` for multiclass task before returning it?
https://github.com/microsoft/LightGBM/blob/237ac299fcffa2b6ed9c68917dfdf0767e88e59c/python-package/lightgbm/basic.py#L1927

python-package/lightgbm/basic.py
```diff
@@ -1840,7 +1883,17 @@ def set_init_score(self, init_score):
...
+            elif is_2d_collection(init_score):
+                init_score = data_to_2d_numpy(init_score, np.float64, name='init_score')
+                init_score = init_score.ravel()
+                self.init_score = init_score
+            else:
+                raise TypeError(
+                    'init_score must be list, numpy 1-D array or pandas Series.\n'
+                    'In multiclass classification init_score must be list, numpy 2-D array or pandas DataFrame.'
+                )
             self.set_field('init_score', init_score)
```
StrikerRUS, line 1897: What about `set_field()` method? Should it also support 2-D collections for multiclass task?

python-package/lightgbm/basic.py
```diff
@@ -1840,7 +1883,17 @@ def set_init_score(self, init_score):
         """
         self.init_score = init_score
         if self.handle is not None and init_score is not None:
-            init_score = list_to_1d_numpy(init_score, np.float64, name='init_score')
+            if is_1d_collection(init_score):
+                init_score = list_to_1d_numpy(init_score, np.float64, name='init_score')
+            elif is_2d_collection(init_score):
+                init_score = data_to_2d_numpy(init_score, np.float64, name='init_score')
+                init_score = init_score.ravel()
```
StrikerRUS, line 1890: I'm not sure that simple `ravel()` can be used here. I guess we should change the order (`ravel(order='F')`).
> For multi-class task, the y_pred is group by class_id first, then group by row_id. If you want to get i-th row y_pred in j-th class, the access way is y_pred[j * num_data + i] and you should group grad and hess in this way as well.

https://github.com/microsoft/LightGBM/blob/237ac299fcffa2b6ed9c68917dfdf0767e88e59c/python-package/lightgbm/basic.py#L1182-L1187

StrikerRUS (commented):
As I already said, I think this is useful enhancement, but it seems it requires deeper integration and more docs updates (see comments below).

python-package/lightgbm/basic.py
jmoralez: Seems like the dataset fields are vectors, or do you mean making the collection 1-D inside that method?

python-package/lightgbm/basic.py
```diff
@@ -165,6 +175,39 @@ def list_to_1d_numpy(data, dtype=np.float32, name='list'):
...
 
 
+def is_numpy_2d_array(data):
+    """Check whether data is a numpy 2-D array."""
+    return isinstance(data, np.ndarray) and len(data.shape) == 2 and data.shape[1] > 1
+
+
+def is_2d_list(data):
+    """Check whether data is a 2-D list."""
+    return isinstance(data, list) and is_1d_list(data[0])
```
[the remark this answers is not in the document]
jmoralez: Should I add a `len(data) > 0` there? I think if someone sends an empty list as init score there's going to be an error somewhere, not sure if this should let it go through though.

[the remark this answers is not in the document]
StrikerRUS: I think `len(data) > 0` here would be good to have right before passing `dat[0]` to `is_1d_list()`.

python-package/lightgbm/basic.py
StrikerRUS: Yeah, cpp side accents only flatten vectors even for multiclass task. I mean, should `set_field()` method accept 2D matrix as init score for multiclass task? Actually, it means just move all (or some) your new logic from `set_init_score` to `if field_name == 'init_score':` branch of  `set_field()` method. So that users will be able pass 2D matrices to both methods.

python-package/lightgbm/basic.py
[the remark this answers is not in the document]
jameslamb: Just want to note something I learned recently, that `scipy` sparse matrices don't support `len()`.

```python
X = csr_matrix((1,1))
len(X)
```

  File "<stdin>", line 1, in <module>
  File "/home/jlamb/miniconda3/lib/python3.7/site-packages/scipy/sparse/base.py", line 291, in __len__
    raise TypeError("sparse matrix length is ambiguous; use getnnz()"
TypeError: sparse matrix length is ambiguous; use getnnz() or shape[0]

I don't think it would be a problem in this specific case since the check comes after `is isinstance(data, list) and` so `len(data)` wouldn't be evaluated in that statement if `data` is not a list, but just wanted to note it in case other refactorings here result in calling `len(data)` when the type of `data` hasn't been checked yet.

tests/python_package_test/test_basic.py
```diff
@@ -8,7 +8,7 @@
 from sklearn.model_selection import train_test_split
 
 import lightgbm as lgb
```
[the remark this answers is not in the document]
jmoralez: Should I override this for the classifier and replace with `shape = [n_samples * n_classes] or shape = [n_samples, n_classes]`?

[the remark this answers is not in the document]
StrikerRUS: I think it can be `"array-like of shape = [n_samples] or shape = [n_samples * n_classes] (for multi-class task) or shape = [n_samples, n_classes] (for multi-class task) or None, optional (default=None)"`

python-package/lightgbm/basic.py
```diff
@@ -1071,6 +1114,7 @@ def __init__(self, data, label=None, reference=None,
             For example, if you have a 100-document dataset with ``group = [10, 20, 40, 10, 10, 10]``, that means that you have 6 groups,
             where the first 10 records are in the first group, records 11-30 are in the second group, records 31-70 are in the third group, etc.
         init_score : list, numpy 1-D array, pandas Series or None, optional (default=None)
+            For multiclass classification can also be list of lists, numpy 2-D array or pandas DataFrame.
```
StrikerRUS, line 1117: This should be written after `init_score :` in allowed types description.
I guess something like this will be OK...
```suggestion
        init_score : list, list of lists (for multi-class task), numpy array, pandas Series, pandas DataFrame (for multi-class task) or None, optional (default=None)
```

tests/python_package_test/test_basic.py
```diff
@@ -8,7 +8,7 @@
 from sklearn.model_selection import train_test_split
```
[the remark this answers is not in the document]
jmoralez: Don't think so, if its a 1-D array then `parts.ndim == 1` and for a 2-D array `parts.shape[1] == 1`

https://github.com/microsoft/LightGBM/blob/3a657c8bf245777bb3cdc6de535b851674ec278c/python-package/lightgbm/dask.py#L132

python-package/lightgbm/basic.py
jmoralez: I don't have a preference, if you think it makes more sense for this to be inside `set_field` I can move it there.

tests/python_package_test/test_basic.py
jmoralez: Is there a way to know the number of classes when this method is called?

tests/python_package_test/test_basic.py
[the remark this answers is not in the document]
StrikerRUS: OK. What does `is_matrix=False` in `_split_to_parts()` mean?

tests/python_package_test/test_basic.py
StrikerRUS: At Dataset level we don't number of classes. But we can get number of data points
https://lightgbm.readthedocs.io/en/latest/C-API.html#c.LGBM_DatasetGetNumData. I guess for non-multiclass task that number should be equal to `len(self.init_score)`. Number of classes can be gotten by dividing `len(self.init_score)` to #rows of Dataset.

python-package/lightgbm/basic.py
StrikerRUS: Why users are able to pass 2-D object to `set_init_score()` and not able to do so in `set_field('init_score', ...)`? I believe it makes sense to unify all interfaces.

python-package/lightgbm/basic.py
```diff
@@ -1823,7 +1866,7 @@ def set_init_score(self, init_score):
 
         Parameters
         ----------
-        init_score : list, numpy 1-D array, pandas Series or None
+        init_score : list, list of lists (for multi-class task), numpy array, pandas Series, pandas DataFrame (for multi-class task) or None, optional (default=None)
```
StrikerRUS, line 1869: It is not optional.
```suggestion
        init_score : list, list of lists (for multi-class task), numpy array, pandas Series, pandas DataFrame (for multi-class task) or None
```

python-package/lightgbm/basic.py
jmoralez: Done in ad4495942d486435a7bfa2a5e382dc661c2e6f7a

tests/python_package_test/test_basic.py
jmoralez: Done in ad4495942d486435a7bfa2a5e382dc661c2e6f7a. I put the logic in `get_field` instead of `get_init_score` following the discussion in https://github.com/microsoft/LightGBM/pull/4150#discussion_r666954862, let me know what you think.

tests/python_package_test/test_basic.py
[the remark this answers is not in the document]
jmoralez: Haha, not sure. Seems to come from https://github.com/dask/dask-lightgbm/blob/dbb66ca423ff26d9b3423092e76275dd00a5b47a/dask_lightgbm/core.py#L89. The `to_delayed` method of arrays returns a 2-D array if the input was 2-D, however setting `is_matrix` to either `True` or `False` passes that assert.

tests/python_package_test/test_basic.py
```diff
@@ -515,3 +515,19 @@ def test_list_to_1d_numpy(y, dtype):
     result = lgb.basic.list_to_1d_numpy(y, dtype=dtype)
     assert result.size == 10
     assert result.dtype == dtype
+
+
+@pytest.mark.parametrize('init_score_type', ['array', 'dataframe', 'list'])
+def test_init_score_for_multiclass_classification(init_score_type):
+    init_score = [[0, 1, 2] for _ in range(10)]
```
StrikerRUS, line 522: I'd prefer to have all unique values in the array to make it totally clear that order argument for `ravel()` is chosen correctly in the source code. I mean, when all values are unique, it's more intuitive that different orders will produce different results.
```suggestion
    init_score = [[i * 10 + j for j in range(3)] for i in range(10)]
```

tests/python_package_test/test_basic.py
```diff
@@ -515,3 +515,19 @@ def test_list_to_1d_numpy(y, dtype):
...
+    if init_score_type == 'array':
+        init_score = np.array(init_score)
+    elif init_score_type == 'dataframe':
+        if not PANDAS_INSTALLED:
+            pytest.skip('Pandas is not installed.')
+        init_score = pd_DataFrame(init_score)
+    data = np.random.rand(10, 2)
+    ds = lgb.basic.Dataset(data, init_score=init_score)
+    ds.construct()
```
StrikerRUS, line 531:
```suggestion
    ds = lgb.basic.Dataset(data, init_score=init_score).construct()
```

tests/python_package_test/test_basic.py
```diff
@@ -515,3 +515,19 @@ def test_list_to_1d_numpy(y, dtype):
...
+    np.testing.assert_equal(ds.init_score, init_score)
+    np.testing.assert_equal(ds.get_field('init_score'), init_score)
```
StrikerRUS, line 533: cpp code that can alter data field should be triggered first, and after that we can check Python attribute. Without calling `get_field()` we just check that

```
ds.init_score = init_score
np.testing.assert_equal(ds.init_score, init_score)
```
Refer to #2390.
```suggestion
    np.testing.assert_equal(ds.get_field('init_score'), init_score)
    np.testing.assert_equal(ds.init_score, init_score)
```

StrikerRUS (changes requested):
@jmoralez 
I'm so sorry for the delay again!
I like the PR in it's current state and I hope this is the last round of review. Please check my comments below.

tests/python_package_test/test_basic.py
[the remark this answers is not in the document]
StrikerRUS: > however setting `is_matrix` to either `True` or `False` passes that assert.

Hmm, maybe we can refactor this legacy code then (in a separate PR, of course)?

python-package/lightgbm/basic.py
```diff
@@ -1908,12 +1951,19 @@ def set_field(self, field_name, data):
...
+            if is_1d_collection(data):
+                data = list_to_1d_numpy(data, np.float64, name='init_score')
+            elif is_2d_collection(data):
+                data = data_to_2d_numpy(data, np.float64, name='init_score')
+                data = data.ravel(order='F')
+            else:
+                raise TypeError(
+                    'init_score must be list, numpy 1-D array or pandas Series.\n'
+                    'In multiclass classification init_score can also be a list of lists, numpy 2-D array or pandas DataFrame.'
+                )
```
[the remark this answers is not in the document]
jmoralez: Thanks for this! I changed your suggestion a bit in 2c7ef3c. Let me know what you think.

tests/python_package_test/test_dask.py
```diff
@@ -1583,17 +1583,14 @@ def test_init_score(task, output, cluster):
             'time_out': 5
         }
         init_score = random.random()
-        # init_scores must be a 1D array, even for multiclass classification
```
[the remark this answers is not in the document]
jmoralez: Addressed in 2c7ef3c. I think the docstrings maybe ended up a bit too verbose, let me know what you think.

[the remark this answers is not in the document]
StrikerRUS: @jmoralez Thanks a lot! I like your explicit wordings.
I'm so sorry, I merged #4558 and introduced conflicts.

Also, I just understood... There shouldn't be any updates for DaskLGBMRegressor's and DaskLGBMRanker's docstrings, `(for multi-class task)` is not applicable there.

python-package/lightgbm/basic.py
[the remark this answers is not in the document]
StrikerRUS: Perfect, thanks!

tests/python_package_test/test_basic.py
```diff
@@ -538,3 +538,18 @@ def test_list_to_1d_numpy(y, dtype):
...
+def test_init_score_for_multiclass_classification(init_score_type):
+    init_score = [[i * 10 + j for j in range(3)] for i in range(10)]
+    if init_score_type == 'array':
+        init_score = np.array(init_score)
+    elif init_score_type == 'dataframe':
+        if not PANDAS_INSTALLED:
+            pytest.skip('Pandas is not installed.')
+        init_score = pd_DataFrame(init_score)
+    data = np.random.rand(10, 2)
+    ds = lgb.basic.Dataset(data, init_score=init_score).construct()
```
StrikerRUS, line 553:
```suggestion
    ds = lgb.Dataset(data, init_score=init_score).construct()
```

StrikerRUS (commented):
@jmoralez Many thanks for addressing all previous comments. I don't have any more except one new below and https://github.com/microsoft/LightGBM/pull/4150#discussion_r700649703.

python-package/lightgbm/basic.py
```diff
@@ -1145,7 +1188,7 @@ def __init__(self, data, label=None, reference=None,
             sum(group) = n_samples.
             For example, if you have a 100-document dataset with ``group = [10, 20, 40, 10, 10, 10]``, that means that you have 6 groups,
             where the first 10 records are in the first group, records 11-30 are in the second group, records 31-70 are in the third group, etc.
-        init_score : list, numpy 1-D array, pandas Series or None, optional (default=None)
+        init_score : list, list of lists (for multi-class task), numpy array, pandas Series, pandas DataFrame (for multi-class task) or None, optional (default=None)
```
jmoralez, line 1191: @StrikerRUS do you think this should have a comma before `or None`? I did include it in the dask docstrings but I just realized this doesn't have it. I'll make them consistent but would like to know what you think is the correct one.

StrikerRUS (approved):
Thank you very much for this PR! LGTM, except one typo below:

python-package/lightgbm/basic.py
StrikerRUS: I like how you did it. I believe a comma before `or None` prevents users from thinking that it possible to include `None`s into a list: https://github.com/microsoft/LightGBM/pull/4557#issue-720040131. However, I'm not sure whether it is grammatically correct or not. @jameslamb should know for sure 🙂

StrikerRUS:
@jameslamb Would you like to provide your review?

python-package/lightgbm/basic.py
jameslamb: I have a personally prefer `a, b, c, or None, optional (default=None) (with the `,` before `or None`),  but both are equally valid and I don't think you need to change anything

jameslamb:
> @jameslamb Would you like to provide your review?

yes please, thanks! I can provide a review later tonight

jameslamb (changes requested):
Looks pretty good to me! The logic makes sense and I think the added test (plus the fact that the changes to existing tests were so minimal) gives me extra confidence that this change is working.

I just left some minor comments around two areas:

* when introduced new module-level objects in the Python package that are only intended for internal use, I think we should prefix them with a `_` to make that a bit clearer
* new functions / methods / classes in the Python package should have type hints added, in pursuit of #3756 . PR authors and reviewers have the most context about the expected types right now, during the PR introducing new code

python-package/lightgbm/basic.py
```diff
@@ -180,6 +190,39 @@ def list_to_1d_numpy(data, dtype=np.float32, name='list'):
...
 
 
+def is_numpy_2d_array(data):
+    """Check whether data is a numpy 2-D array."""
+    return isinstance(data, np.ndarray) and len(data.shape) == 2 and data.shape[1] > 1
+
+
+def is_2d_list(data):
+    """Check whether data is a 2-D list."""
+    return isinstance(data, list) and len(data) > 0 and is_1d_list(data[0])
```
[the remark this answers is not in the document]
jmoralez: The `is_1d_list` function was there before so you didn't suggest adding the `_` prefix and the type hints. Do you want me to add them?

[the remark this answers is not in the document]
jameslamb: oh sorry that's a mistake, I only meant for these suggestions to apply to newly-added code

jameslamb (approved):
Looks like all the suggestions I made were addressed, thanks very much.

This PR should remove a bit of friction for users, really nice addition!

StrikerRUS:
@jmoralez Could you please sync with `master` to merge recent CI fixes into this branch?

jmoralez:
Hi. I have covid so I'll be inactive for a couple of weeks, feel free to make any changes to my PRs.

jameslamb:
Ah! I'm so sorry to hear that @jmoralez !!!

Don't worry at all about notifications in LightGBM, focus on resting up and hopefully feeling better soon! Let us know in the maintainer Slack whenever you're feeling better, and until then I'll avoid `@`-ing you.

StrikerRUS:
Ah, get well soon!

PR #4151, opened 2021-04-01
[tests][dask] use dy_true mean in denominator for _r2_score

jmoralez:
Reference: https://github.com/scikit-learn/scikit-learn/blob/95119c13af77c76e150b753485c662b7c52a41a2/sklearn/metrics/_regression.py#L693-L694


jameslamb:
Linking that this came from https://github.com/microsoft/LightGBM/pull/4149#discussion_r605371439

jameslamb (approved):
thanks for fixing this!

Issue #4034, opened 2021-03-01
[R-package] predict() breaks when using a Dataset stored in a file


ticarki:
## Description
On Windows R crashes using Dataset.lgb.save, without error message.
On Linux I am able to save the dataset, but lgb.predict can not find saved dataset 

## Reproducible example

For the Windows bug (the example given by lightgbm::lgb.Dataset.save)
```{R}
data(agaricus.train, package = "lightgbm")
train <- agaricus.train
dtrain <- lgb.Dataset(train$data, label = train$label)
train_file =  tempfile(fileext = ".bin")
lgb.Dataset.save(dtrain, train_file)
```

For the Linux bug (Example given by lightgbm::lgb.load + predict using a file as input)
```{R}
data(agaricus.train, package = "lightgbm")
train <- agaricus.train
dtrain <- lgb.Dataset(train$data, label = train$label)
data(agaricus.test, package = "lightgbm")
test <- agaricus.test
dtest <- lgb.Dataset.create.valid(dtrain, test$data, label = test$label)

test_file =  file.path(getwd(), "test.bin")
lgb.Dataset.save(dtest, test_file)

params <- list(objective = "regression", metric = "l2")
valids <- list(test = dtest)
model <- lgb.train(params = params, data = dtrain, nrounds = 5L, 
                   valids = valids, learning_rate = 1.0, 
                   early_stopping_rounds = 3L)
model_file <- tempfile(fileext = ".txt")
lgb.save(model, model_file)
load_booster <- lgb.load(filename = model_file)
model_string <- model$save_model_to_string(NULL) # saves best iteration
load_booster_from_str <- lgb.load(model_str = model_string)

model$predict(test_file)
```

The error:
``` {R}
Error in lgb.call(fun_name = "LGBM_BoosterPredictForFile_R", ret = NULL,  : 
  [LightGBM] [Fatal] Data file �?�V doesn't exist.
```

## Environment info

LightGBM version or commit hash:

lightgbm_3.1.1 

Command(s) you used to install LightGBM

```R
install.packages('lightgbm')
```

jameslamb:
Thanks very much for using `{lightgbm}` and for the excellent bug report! It's possible that the Windows part of this is related to another not-yet-solved issue (#4007), but I'm not sure yet.

[...]

PR #4155, opened 2021-04-02
[R-package] prevent symbol lookup conflicts (fixes #4045)

### LINKED ISSUE #4007: [R-package] R session crashes on Windows when {lightgbm} is loaded after tidyverse
I am using lightgbm in a `tidymodels` workflow using the [treesnip](https://github.com/curso-r/treesnip) package. Data loading and wrangling uses functions from the `tidyverse`.

If `library(lightgbm)` is loaded after `library(tidyverse)`, the R session in RStudio crashes during model fit without further error.

If `library(lightgbm)` is loaded before `library(tidyverse)`, this does not happen.

This results in a crash:

```
library(tidyverse)

library(tidymodels)
library(lightgbm)
library(treesnip)

workflow <- workflow() %>% 
  add_recipe(recipe(formula = Sepal.Length ~ ., data = iris) ) %>% 
  add_model(boost_tree() %>% 
              set_mode("regression") %>% 
              set_engine("lightgbm"))

fit(workflow, data = iris)
```

This works:

```
library(tidymodels)
library(lightgbm)
library(treesnip)

library(tidyverse)

workflow <- workflow() %>% 
  add_recipe(recipe(formula = Sepal.Length ~ ., data = iris) ) %>% 
  add_model(boost_tree() %>% 
              set_mode("regression") %>% 
              set_engine("lightgbm"))

fit(workflow, data = iris)
```

I installed lightgbm v3.1.1 from CRAN.

Session info:

```
R version 4.0.3 (2020-10-10)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 18363)

Matrix products: default

locale:
[1] LC_COLLATE=English_United Kingdom.1252  LC_CTYPE=English_United Kingdom.1252   
[3] LC_MONETARY=English_United Kingdom.1252 LC_NUMERIC=C                           
[5] LC_TIME=English_United Kingdom.1252    

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
 [1] forcats_0.5.1       stringr_1.4.0       readr_1.4.0         tidyverse_1.3.0     treesnip_0.1.0.9000
 [6] lightgbm_3.1.1      R6_2.5.0            yardstick_0.0.7     workflows_0.2.1     tune_0.1.2         
[11] tidyr_1.1.2         tibble_3.0.6        rsample_0.0.9       recipes_0.1.15      purrr_0.3.4        
[16] parsnip_0.1.5       modeldata_0.1.0     infer_0.5.4         ggplot2_3.3.3       dplyr_1.0.4        
[21] dials_0.0.9         scales_1.1.1        broom_0.7.4         tidymodels_0.1.2   

 [1] httr_1.4.2         jsonlite_1.7.2     splines_4.0.3      foreach_1.5.1      modelr_0.1.8       prodlim_2019.11.13
 [7] vip_0.3.2          assertthat_0.2.1   GPfit_1.0-8        cellranger_1.1.0   globals_0.14.0     ipred_0.9-9       
[13] pillar_1.4.7       backports_1.2.1    lattice_0.20-41    glue_1.4.2         pROC_1.17.0.1      digest_0.6.27     
[19] hardhat_0.1.5      rvest_0.3.6        snakecase_0.11.0   colorspace_2.0-0   Matrix_1.2-18      plyr_1.8.6        
[25] timeDate_3043.102  pkgconfig_2.0.3    lhs_1.1.1          DiceDesign_1.9     listenv_0.8.0      haven_2.3.1       
[31] gower_0.2.2        lava_1.6.8.1       generics_0.1.0     tictoc_1.0         ellipsis_0.3.1     withr_2.4.1       
[37] janitor_2.1.0      furrr_0.2.2        nnet_7.3-14        cli_2.3.0          readxl_1.3.1       survival_3.2-7    
[43] magrittr_2.0.1     crayon_1.4.1       fs_1.5.0           future_1.21.0      parallelly_1.23.0  MASS_7.3-53       
[49] xml2_1.3.2         class_7.3-17       tools_4.0.3        data.table_1.13.6  hms_1.0.0          lifecycle_1.0.0   
[55] reprex_1.0.0       munsell_0.5.0      compiler_4.0.3     tinytex_0.29       rlang_0.4.10       grid_4.0.3        
[61] iterators_1.0.13   rstudioapi_0.13    gtable_0.3.0       codetools_0.2-16   DBI_1.1.1          gridExtra_2.3     
[67] lubridate_1.7.9.2  stringi_1.5.3      parallel_4.0.3     Rcpp_1.0.6         vctrs_0.3.6        rpart_4.1-15      
[73] dbplyr_2.1.0       tidyselect_1.1.0   xfun_0.21         
```

### LINKED ISSUE #4045: [R-package] FATAL Error when run train()
## Description
getting Fatal Error when i run lgb.train()

## Reproducible example
model <- lgb.train(params = params,data = dtrain) 

## Environment info

error

jameslamb:
This PR fixes a critical issue affecting Windows users of the R package. In #4045, several users reported a problem like "if I load certain R libraries before `{lightgbm}`, `{lightgbm}` can cause the R session to crash when constructing a dataset".

After some investigation (detailed in https://github.com/microsoft/LightGBM/issues/4045#issuecomment-812289182), I believe I found the root cause for this. The routine `R_registerRoutines()` at https://github.com/microsoft/LightGBM/blob/d517ba12f2e7862ac533908304dddbd770655d2b/R-package/src/lightgbm_R.cpp#L724 isn't actually being called by R!

The key piece of documentation is at https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Converting-a-package-to-use-registration.

> Construct the registration table ... if included in a C++ file the ‘R_init’ function would need to be declared extern "C"):

This PR adds that declaration to `lightgbm_R.cpp` and adds a unit test to prevent this type of mistake from being re-introduced in the future.

Fixes #4045.
Addresses the Windows half of #4034.

I need to test if this would fix #4007. Will update this description once I try those.

<hr>

## Background

R provides an interface for "native routine registration", a mechanism to tell R exactly which symbols it should expect to find in a shared library (`.dll` / `.so`). This makes calls to compiled code with `.Call()` safer because R looks in a specific place.

If that type of registration isn't done, then it can create the possibility for conflicts with other loaded DLLs. From https://cran.r-project.org/doc/manuals/r-release/R-exts.html#dyn_002eload-and-dyn_002eunload:

> If a shared object/DLL is loaded more than once the most recent version is used. More generally, if the same symbol name appears in several shared objects, the most recently loaded occurrence is used. The PACKAGE argument and registration (see the next section) provide good ways to avoid any ambiguity in which occurrence is meant.

This note from https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Registering-native-routines is relevant as well.

> In calls to .C, .Call, .Fortran and .External, R must locate the specified native routine by looking in the appropriate shared object/DLL. By default, R uses the operating-system-specific dynamic loader to lookup the symbol in all132 loaded DLLs and the R executable or libraries it is linked to. Alternatively, the author of the DLL can explicitly register routines with R and use a single, platform-independent mechanism for finding the routines in the DLL. One can use this registration mechanism to provide additional information about a routine, including the number and type of the arguments, and also make it available to R programmers under a different name.

<hr>

## How to test this

You can run the following code to check if registration was successful.

```r
library(lightgbm)
dll_info <- getLoadedDLLs()[["lightgbm"]]

# check that dynamic lookup has been disabled
!isTRUE(dll_info[["dynamicLookup"]])

# check that all the expected symbols show up
getDLLRegisteredRoutines(dll_info[["path"]])
```

```text
                               .Call .Call.numParameters
1                LGBM_GetLastError_R                   3
2       LGBM_DatasetCreateFromFile_R                   5
3        LGBM_DatasetCreateFromCSC_R                  10
4        LGBM_DatasetCreateFromMat_R                   7
5            LGBM_DatasetGetSubset_R                   6
6      LGBM_DatasetSetFeatureNames_R                   3
7      LGBM_DatasetGetFeatureNames_R                   5
8           LGBM_DatasetSaveBinary_R                   3
9                 LGBM_DatasetFree_R                   2
10            LGBM_DatasetSetField_R                   5
11        LGBM_DatasetGetFieldSize_R                   4
12            LGBM_DatasetGetField_R                   4
13 LGBM_DatasetUpdateParamChecking_R                   3
14          LGBM_DatasetGetNumData_R                   3
15       LGBM_DatasetGetNumFeature_R                   3
16              LGBM_BoosterCreate_R                   4
17                LGBM_BoosterFree_R                   2
18 LGBM_BoosterCreateFromModelfile_R                   3
19 LGBM_BoosterLoadModelFromString_R                   3
20               LGBM_BoosterMerge_R                   3
21        LGBM_BoosterAddValidData_R                   3
22   LGBM_BoosterResetTrainingData_R                   3
23      LGBM_BoosterResetParameter_R                   3
24       LGBM_BoosterGetNumClasses_R                   3
25       LGBM_BoosterUpdateOneIter_R                   2
26 LGBM_BoosterUpdateOneIterCustom_R                   5
27     LGBM_BoosterRollbackOneIter_R                   2
28 LGBM_BoosterGetCurrentIteration_R                   3
29  LGBM_BoosterGetUpperBoundValue_R                   3
30  LGBM_BoosterGetLowerBoundValue_R                   3
31        LGBM_BoosterGetEvalNames_R                   5
32             LGBM_BoosterGetEval_R                   4
33       LGBM_BoosterGetNumPredict_R                   4
34          LGBM_BoosterGetPredict_R                   4
35      LGBM_BoosterPredictForFile_R                  11
36      LGBM_BoosterCalcNumPredict_R                   9
37       LGBM_BoosterPredictForCSC_R                  15
38       LGBM_BoosterPredictForMat_R                  12
39           LGBM_BoosterSaveModel_R                   5
40   LGBM_BoosterSaveModelToString_R                   7
41           LGBM_BoosterDumpModel_R                   
```


R-package/tests/testthat/test_registration.R
```diff
@@ -0,0 +1,17 @@
+ON_WINDOWS <- .Platform$OS.type == "windows"
+
+# this test is used to catch silent errors in routine registration,
+# like the one documented in
+# https://github.com/microsoft/LightGBM/issues/4045#issuecomment-812289182
+test_that("lightgbm routine registration worked", {
+    testthat::skip_if_not(ON_WINDOWS)
```
jameslamb, line 7: this `skip_if_not()` is to work around this error I saw on Linux jobs in CI

```text
* checking tests ...
  Running ‘testthat.R’
   ERROR
Running the tests in ‘tests/testthat.R’ failed.
Last 13 lines of output:
  ══ Failed tests ════════════════════════════════════════════════════════════════
  ── Failure (test_registration.R:8:5): lightgbm routine registration worked ─────
  dll_info[["dynamicLookup"]] is not FALSE
  
  `actual` is NULL
  `expected` is a logical vector (FALSE)
  ── Error (test_registration.R:11:5): lightgbm routine registration worked ──────
  Error: no applicable method for 'getDLLRegisteredRoutines' applied to an object of class "NULL"
  Backtrace:
      █
   1. └─base::getDLLRegisteredRoutines(dll_info[["path"]]) test_registration.R:11:4
  
  [ FAIL 2 | WARN 0 | SKIP 3 | PASS 634 ]
  Error: Test failures
  Execution halted
```

jameslamb:
hmmm, seeing a few things in CI that I didn't see locally.

```text
* checking foreign function calls ... NOTE

Registration problems:

  symbol 'fun_name' not in namespace:

   .Call(fun_name, ..., ret, call_state, PACKAGE = "lightgbm")

  symbol 'fun_name' not in namespace:

   .Call(fun_name, ..., call_state, PACKAGE = "lightgbm")

See chapter 'System and foreign language interfaces' in the 'Writing R

Extensions' manual.
```

I'm going to remove reviewers and move this back to draft for now

StrikerRUS:
@jameslamb 

> move this back to draft for now

Please share how did you do that. Where is the corresponding button?

jameslamb:
> Please share how did you do that. Where is the corresponding button?

Ha sure! It's very very hidden in the UI, because the text color is the same as all of the other text.

image

You can click "convert to draft" under "Reviewers" (top right corner of the page) to convert back.

StrikerRUS:
@jameslamb Thanks a lot! I was lacking this button multiple times!

StrikerRUS:
Can resolving #3016 help with this issue as well?

jameslamb:
> Can resolving #3016 help with this issue as well?

Right now, I don't think so.

I think the problem is that `{lightgbm}` wraps `.Call()` calls in a function to handle errors called `lgb.call()` (https://github.com/microsoft/LightGBM/blob/d517ba12f2e7862ac533908304dddbd770655d2b/R-package/R/utils.R#L62), but if you are using routine registration then `R CMD check --as-cran` expects that the first argument to `.Call()` is one of the symbols you've registered. Like this: https://github.com/Rdatatable/data.table/blob/ec1259af1bf13fc0c96a1d3f9e84d55d8106a9a4/R/bmerge.R#L181.

I'm experimenting with this today. I think error handling can be moved into the C++ side and then `lgb.call()` can be factored out. It's an internal function so there wouldn't be any user-facing impact.

R-package/tests/testthat/test_registration.R
jameslamb: ah! The problem with this test isn't actually about operating system. This test will fail with an error like the one above on LightGBM's CMake-based builds and pass on CRAN builds. That's because, by using `install.libs.R` to build `lib_lightgbm` instead of R's built-in machinery, the routine registration stuff doesn't work.

Given that as of this PR the package will now use literal R symbols (e.g. `.Call(LGBM_BoosterGetCurrentIteration_R,...`), I don't think this test is necessary. Any failures in the registration will result in "object not found errors".

I'm going to remove this test, and will provide more explanation in a separate comment for how to make this work with CMake-based builds.

jameslamb:
Ok, I think that this PR is ready for review! I know the description has a lot of information and this requires some deep knowledge of how R's build system works, so I'll summarize here:

1. Today, `{lightgbm}` code that calls C++ functions with `.Call()` is using an unsafe strategy to search for symbols, and that can lead to serious conflicts with other loaded packages (see #4045)
2. The preferred, and safer, way to do this is to use "routine registration" (documented in the PR description)
3. When you use registration, R will create R objects in the package's namespace and expect you to use them. So, for example, for the C++ function [`LGBM_DatasetCreateFromCSC_R`](https://github.com/microsoft/LightGBM/blob/b6c71e5e941b885b1a5169d460c430df9d392fa9/R-package/src/lightgbm_R.h#L55), you have to use `.Call(LGBM_DatasetCreateFromCSC_R, ...)`, not `.Call("LGBM_DatasetCreateFromCSC_R", ..., PACKAGE = "lightgbm")`.
4. When you use `install.libs.R` to build a custom `.so`/`.dll` (as `{lightgbm}` does for CMake-based builds), this automatic creation of R objects will not work. To force that to be done, you can use code like `useDynLib(lib_lightgbm, LGBM_DatasetCreateFromCSC_R, ...)`

As a result of the changes in this PR:

1. `{lightgbm}` should be safer to use with other packages, and the critical issue #4045 should be fixed
2. `{lightgbm}`'s CRAN package will be structured in a way that is closer to the recommended practices documented in https://cran.r-project.org/doc/manuals/r-release/R-exts.html

R-package/R/utils.R
```diff
@@ -59,53 +57,6 @@ lgb.last_error <- function() {
 
 }
 
-lgb.call <- function(fun_name, ret, ...) {
```
jameslamb, line 62: These `lgb.call()` functions can be removed because once routine registration is used, referencing routines using strings does not work: https://github.com/microsoft/LightGBM/pull/4155#issuecomment-812315498.

R-package/R/lgb.Dataset.R
```diff
@@ -13,7 +13,12 @@ Dataset <- R6::R6Class(
       if (!lgb.is.null.handle(x = private$handle)) {
 
         # Freeing up handle
-        lgb.call(fun_name = "LGBM_DatasetFree_R", ret = NULL, private$handle)
+        call_state <- 0L
```
jameslamb, line 16: Even though error-handling has now been moved into the C++ side (#4163), as of this PR passing in `call_state` is still needed. The functions in https://github.com/microsoft/LightGBM/blob/b6c71e5e941b885b1a5169d460c430df9d392fa9/R-package/src/lightgbm_R.h has a return type `LGBM_SE`, which is the LightGBM equivalent of R's built-in `SEXP` type.

If you change those routines to all be `void` (since they modify their inputs by reference and don't technically have to refer to anything), errors are thrown about not being able to convert a null to `SEXP`.

So the `call_state` thing is a simple way to allocate an `LGBM_SE` (an R integer, in this case) that can then be returned.

As part of #3016 we'll be able to remove these unnecessary `call_state` things, but I believe they're harmless for now.

build_r.R
```diff
@@ -369,6 +369,41 @@ description_contents <- gsub(
 )
 writeLines(description_contents, DESCRIPTION_FILE)
 
+# CMake-based builds can't currently use R's builtin routine registration,
+# so have to update NAMESPACE manually, with a statement like this:
+#
+# useDynLib(lib_lightgbm, LGBM_GetLastError_R, LGBM_DatasetCreateFromFile_R, ...)
```
StrikerRUS, line 375: Maybe it will be easier to have this content by default in the file, and **remove** it for CRAN build?

R-package/R/lgb.Booster.R
```diff
@@ -474,14 +498,43 @@ Booster <- R6::R6Class(
         num_iteration <- self$best_iter
       }
 
-      # Return model string
-      return(
-        lgb.call.return.str(
-          fun_name = "LGBM_BoosterSaveModelToString_R"
+      # Create buffer
```
StrikerRUS, line 501: Maybe write a new helper function for calls with buffer allocation? I believe it will improve the code readability.

StrikerRUS (approved):
Excellent improvement for the package! I'm approving based on the detailed PR description, linked paragraph of article in comments and green CI checks.
Just two minor comments below.

R-package/R/lgb.Booster.R
jameslamb: That was exactly what `lgb.call.return.str()` was. But since you cannot pass a string with a function name to `.Call()` once registration is used (https://github.com/microsoft/LightGBM/pull/4155#issuecomment-812315498), I couldn't see a good way to keep a shared helper function.

I also think we'll be able to factor this code away completely soon, as part of #3016, by just creating the R character vector on the C++ side and returning it. See #4242 for a first example of that.

StrikerRUS: Ah, OK, I see!

build_r.R
jameslamb: hmmm I'm not sure. This looks like more code than having this in the file by default and removing it for CRAN builds, but I thought this R code would be easier to write than the equivalent code with `sed` in `build-cran-package.sh`. Just because it involves replacing / writing literal `.` and literal `)` / `(` and I was intimidated by figuring out how to do the escaping correctly in `sed` haha.

Do you have a strong preference for this change? Otherwise, I'd prefer to leave this as-is for now and maybe make that change as some maintenance work in the future. Just because I'd like to keep pushing forward on #3016 and it doesn't seem like there is anything *wrong* about the current approach.

StrikerRUS: Not really, this was just a proposal to simplify the code. I'm totally fine with the working current approach.

jameslamb: ok great, thank you very much!

PR #4157, opened 2021-04-02
[docs] update link to Boost binaries

StrikerRUS:
Refer to https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/.

> May 1st, 2021 | Bintray services will no longer be available.


jameslamb (approved):
good find, thank you

PR #4147, opened 2021-03-30
[python] added f-strings to docs/conf.py

NovusEdge:

PR #4104, opened 2021-03-24
[ci] restrict CMake version for CUDA+Clang jobs

StrikerRUS:
Our CUDA+Clang builds are incopatible with yesterday-released CMake 3.20.0 version:

Found Clang, but using gcc.

```
-- The C compiler identification is Clang 6.0.0
-- The CXX compiler identification is Clang 6.0.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/clang - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/clang++ - skipped

...

gcc: error: unrecognized command line option ‘-fopenmp=libomp’; did you mean ‘-fopenmp-simd’?
```

```
-- The C compiler identification is Clang 6.0.0
-- The CXX compiler identification is Clang 6.0.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/clang - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/clang++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- The CUDA compiler identification is NVIDIA 10.0.130
-- Detecting CUDA compiler ABI info
-- Detecting CUDA compiler ABI info - done
-- Check for working CUDA compiler: /usr/local/cuda/bin/nvcc - skipped
-- Detecting CUDA compile features
-- Detecting CUDA compile features - done
-- Found OpenMP_C: -fopenmp=libomp (found version "3.1") 
-- Found OpenMP_CXX: -fopenmp=libomp (found version "3.1") 
-- Found OpenMP: TRUE (found version "3.1")  
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Failed
-- Looking for pthread_create in pthreads
-- Looking for pthread_create in pthreads - not found
-- Looking for pthread_create in pthread
-- Looking for pthread_create in pthread - found
-- Found Threads: TRUE  
[...]

PR #4158, opened 2021-04-03
[ci] upgrade R to 4.0.5 in CI

jameslamb:
R 4.0.5 ("Shake and Throw") was released two days ago (https://stat.ethz.ch/pipermail/r-announce/2021/000668.html).

This PR updates this project's continuous integration environments to use the new version. Linking #3763.

Release notes: https://cran.r-project.org/bin/windows/base/NEWS.R-4.0.5.html

### DIFF
--- a/.ci/test_r_package.sh
+++ b/.ci/test_r_package.sh
@@ -34,8 +34,8 @@ if [[ "${R_MAJOR_VERSION}" == "3" ]]; then
     export R_APT_REPO="bionic-cran35/"
 elif [[ "${R_MAJOR_VERSION}" == "4" ]]; then
     export R_MAC_VERSION=4.0.5
-    export R_LINUX_VERSION="4.0.5-1.1804.0"
-    export R_APT_REPO="bionic-cran40/"
+    export R_LINUX_VERSION="4.0.5-1.2004.0"
+    export R_APT_REPO="focal-cran40/"
 else
     echo "Unrecognized R version: ${R_VERSION}"
     exit -1
--- a/.ci/test_r_package.sh
+++ b/.ci/test_r_package.sh
@@ -77,6 +77,7 @@ fi
 
 # Installing R precompiled for Mac OS 10.11 or higher
 if [[ $OS_NAME == "macos" ]]; then
+    brew update-reset && brew update
     if [[ $R_BUILD_TYPE == "cran" ]]; then
         brew install automake
     fi
--- a/.ci/test_r_package.sh
+++ b/.ci/test_r_package.sh
@@ -33,8 +33,8 @@ if [[ "${R_MAJOR_VERSION}" == "3" ]]; then
     export R_LINUX_VERSION="3.6.3-1bionic"
     export R_APT_REPO="bionic-cran35/"
 elif [[ "${R_MAJOR_VERSION}" == "4" ]]; then
-    export R_MAC_VERSION=4.0.4
-    export R_LINUX_VERSION="4.0.4-1.1804.0"
+    export R_MAC_VERSION=4.0.5
+    export R_LINUX_VERSION="4.0.5-1.1804.0"
     export R_APT_REPO="bionic-cran40/"
 else
     echo "Unrecognized R version: ${R_VERSION}"
--- a/.ci/test_r_package_windows.ps1
+++ b/.ci/test_r_package_windows.ps1
@@ -66,7 +66,7 @@ if ($env:R_MAJOR_VERSION -eq "3") {
   $env:RTOOLS_BIN = "$RTOOLS_INSTALL_PATH\usr\bin"
   $env:RTOOLS_MINGW_BIN = "$RTOOLS_INSTALL_PATH\mingw64\bin"
   $env:RTOOLS_EXE_FILE = "rtools40-x86_64.exe"
-  $env:R_WINDOWS_VERSION = "4.0.4"
+  $env:R_WINDOWS_VERSION = "4.0.5"
 } else {
   Write-Output "[ERROR] Unrecognized R version: $env:R_VERSION"
   Check-Output $false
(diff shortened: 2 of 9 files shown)


jameslamb:
On Mac, this is failing with the error observed in https://github.com/microsoft/LightGBM/pull/4147#issuecomment-812599863

```text
==> Downloading from http://ctan.math.utah.edu/ctan/tex-archive/systems/mac/mactex/mactex-basictex-20200407.pkg
```

On ubuntu, looks I'm not sure what the root problem is. I tried removing `--no-install-recommends` and installing `r-base-core` explicitly, neither worked. Will have to experiment this weekend.

```text
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 r-base-dev : Depends: r-base-core (>= 4.0.5-1.1804.0) but 4.0.4-1.2004.0 is to be installed
E: Unable to correct problems, you have held broken packages.
```

StrikerRUS:
@jameslamb 
> On ubuntu, looks I'm not sure what the root problem is. 

This is the same issue I faced during restricting CMake version recently.
https://github.com/microsoft/LightGBM/pull/4104#discussion_r600506347
You have to (hope temporarily) explicitly specify dependency (`r-base-core` in this case) to resolve the issue.

jameslamb:
> You have to (hope temporarily) explicitly specify dependency (r-base-core in this case) with version to resolve the issue.

Right, I tried this and it did not work. Let me add more details on what I meant by "doesn't work".

Adding `r-base-core` 4.0.5 explicitly yields this:

```text
The following packages have unmet dependencies:
 r-base-core : Depends: libicu60 (>= 60.1-1~) but it is not installable
               Depends: libreadline7 (>= 6.0) but it is not installable
E: Unable to correct problems, you have held broken packages.
```

Trying to add `libicu60` and `libreadline7` explicitly yields this

```text
Package libreadline7 is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

Package libicu60 is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

E: Package 'libicu60' has no installation candidate
E: Package 'libreadline7' has no installation candidate
```

So maybe we just cannot update to R 4.0.5 on Ubuntu 20.04 yet. `libicu60` and `libreadline7` are not available on Ubuntu 20.04 (https://packages.ubuntu.com/search?suite=all&searchon=names&keywords=libicu60, https://packages.ubuntu.com/search?suite=all&section=all&arch=any&keywords=libreadline7&searchon=names).

Hopefully if we check back in a few days, it will be possible.

StrikerRUS (approved):

jameslamb:
Ok I just retried this locally and it seems like just installing `r-base-dev` should now work without any workarounds! Let's see if CI agrees with me.

To test locally, I replicated LightGBM's CI environment with the commands below. It isn't identical (since R jobs run in [the Azure DevOps `ubuntu-latest` VM image](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md)) which is very different from the `ubuntu:latest` docker image, but I'm hoping that those differences are not meaningful for the purpose of `apt` resolving dependencies.

Pasting my testing setup here just in case it's useful as a reference in the future.

```shell
docker run \
    -v $(pwd):/opt/LightGBM \
    -w /opt/LightGBM \
    --env BUILD_DIRECTORY=/opt/LightGBM \
    --env COMPILER=gcc \
    --env GITHUB_ACTIONS="true" \
    --env IN_UBUNTU_LATEST_CONTAINER="true" \
    --env OS_NAME=linux \
    --env R_VERSION=4.0 \
    --env R_BUILD_TYPE=cran \
    --env SETUP_CONDA="true" \
    --env TASK=r-package \
    -it \
    ubuntu:latest \
    /bin/bash

apt update -y
apt install -y sudo
sudo apt-get install -y dirmngr
./.ci/setup.sh
./.ci/test.sh
```

jameslamb:
Sad, this is still failing with the same error. Ok, I'll keep investigating locally.

jameslamb:
Ok, I think I got this working! The trick was to switch the repository for R 4.0.5 to `focal-cran40/`.

<hr>

All 3 CUDA jobs failed with no logs after running for more than hour. I just restarted them.

image

StrikerRUS (approved):

StrikerRUS:
@jameslamb What do you think about the following trick?
https://github.com/BayesWitnesses/m2cgen/commit/6eabad16de5ab4396eb4fb5d88109f9b45302f43
It allows to not change codename of Ubuntu in repo address each time we bump/downgrade/switch it.

jameslamb:
> @jameslamb What do you think about the following trick?
> [BayesWitnesses/m2cgen@6eabad1](https://github.com/BayesWitnesses/m2cgen/commit/6eabad16de5ab4396eb4fb5d88109f9b45302f43)
> It allows to not change codename of Ubuntu in repo address each time we bump/downgrade/switch it.

Nice trick! But that only works if you are not pinning the version of R (further down in that file, you have an unpinned `r-base`). That's problematic if you want to test on environments other than the most recent R, I think, since the version numbers have distribution information in them.

For example. `3.6.3-1bionic` (https://github.com/microsoft/LightGBM/blob/b674439aec933e0c1b72efc7dda482a9c7e75283/.ci/test_r_package.sh#L33)

So I think it makes sense for `mc2gen` since there's only a single R job and it always wants to get the newest-released R, but I don't think it would work cleanly here in LightGBM.

jameslamb:
ok merging this one with admin rights, since it's starting to cause others to fail (https://github.com/microsoft/LightGBM/pull/4167#issuecomment-816981625) and since if we rebuild it, it will fail on the `check-docs` task (fixed in https://github.com/microsoft/LightGBM/pull/4168)

PR #4161, opened 2021-04-04
[ci] run 'brew update' in macOS jobs (fixes #4160)

### LINKED ISSUE #4160: [ci] installing 'basictex' from homebrew failing in macOS builds
Opening an issue for this because the problem has been going on for a few days.

This was first noticed two days ago, in https://github.com/microsoft/LightGBM/pull/4147#pullrequestreview-627053285. R-package jobs on GitHub Actions (for example, [this one](https://github.com/microsoft/LightGBM/pull/4147/checks?check_run_id=2254678244)) are failing with the following errors from `R CMD check`.

```text
* checking PDF version of manual ... WARNING
* checking PDF version of manual without hyperrefs or index ... ERROR
Re-running with no redirection of stdout/stderr.
Hmm ... looks like a package
Error in texi2dvi(file = file, pdf = TRUE, clean = clean, quiet = quiet,  : 
  pdflatex is not available
Error in texi2dvi(file = file, pdf = TRUE, clean = clean, quiet = quiet,  : 
  pdflatex is not available
Error in running tools::texi2pdf()
You may want to clean up by 'rm -Rf /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T//RtmppA4pCU/Rd2pdf86631a98164'
* checking for non-standard things in the check directory ... NOTE
```

I believe the root case is that installing `basictex` is failing.

```text
==> Downloading from https://ctan.math.illinois.edu/systems/mac/mactex/mactex-basictex-20200407.pkg
curl: (22) The requested URL returned error: 404 
sudo: tlmgr: command not found
sudo: tlmgr: command not found
```

This happens at https://github.com/microsoft/LightGBM/blob/d517ba12f2e7862ac533908304dddbd770655d2b/.ci/test_r_package.sh#L86

## More Context

This is not about the `ctan.math.illinois.edu` mirror specifically. `brew` is picking a random mirror, and I've seen others mentioned in other builds, and locally on my own Mac.

It looks like these failures started around the same time as this relevant-looking PR was merged: https://github.com/Homebrew/homebrew-cask/pull/103001. It's interesting that version `20200407` is mentioned in those logs, when according that to that PR the new version should be `2021.0325`.

jameslamb:
As documented in #4160, LightGBM's macOS jobs for the R package started failing two days ago, roughly around the same time as the version of `basictex` was updated at https://github.com/Homebrew/homebrew-cask/pull/103001.

I was going to file a bug report with Homebrew, and in that process I was directed to that project's "how to report bugs" guidelines: https://github.com/Homebrew/homebrew-cask#reporting-bugs. That includes the suggestion that you should run `brew update-reset && brew update`. After trying that on my Mac, I was then able to `brew install --cask basictex`.

This PR proposes adding a similar to step to LightGBM's CI. This is similar to how this project uses `sudo apt-get update` to update repository information for the `apt` package manager on Linux jobs.

* `brew update-reset`: "Fetch and reset Homebrew and all tap repositories (or any specified
repository) using git(1) to their latest origin/HEAD."
* `brew update`: "Fetch the newest version of Homebrew and all formulae from GitHub using git(1)
and perform any necessary migrations"

I believe this will fix #4160 and prevent similar disruptions in the future, and that it should add only a few seconds to CI jobs.

### DIFF
--- a/.ci/setup.sh
+++ b/.ci/setup.sh
@@ -1,7 +1,6 @@
 #!/bin/bash
 
 if [[ $OS_NAME == "macos" ]]; then
-    brew update-reset && brew update
     if  [[ $COMPILER == "clang" ]]; then
         brew install libomp
         if [[ $AZURE == "true" ]]; then
--- a/.ci/setup.sh
+++ b/.ci/setup.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 if [[ $OS_NAME == "macos" ]]; then
+    brew update-reset && brew update
     if  [[ $COMPILER == "clang" ]]; then
         brew install libomp
         if [[ $AZURE == "true" ]]; then
--- a/.ci/test_r_package.sh
+++ b/.ci/test_r_package.sh
@@ -77,6 +77,7 @@ fi
 
 # Installing R precompiled for Mac OS 10.11 or higher
 if [[ $OS_NAME == "macos" ]]; then
+    brew update-reset && brew update
     if [[ $R_BUILD_TYPE == "cran" ]]; then
         brew install automake
     fi


.ci/setup.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 if [[ $OS_NAME == "macos" ]]; then
+    brew update-reset && brew update
```
StrikerRUS, line 4: Please move this temp line to 
https://github.com/microsoft/LightGBM/blob/d517ba12f2e7862ac533908304dddbd770655d2b/.ci/test_r_package.sh#L79
with the aim to not slow down non-R jobs.

StrikerRUS (changes requested):

.ci/setup.sh
jameslamb: oh I don't really think of it as temporary or as R-specific. What is currently happening with basictex could just as easily happen with anything else installed from brew.

StrikerRUS: Agree, I remember we had to use `brew update` for `gcc` for a long time. But I think we should avoid `brew update` when it is possible due to it increases jobs time.

jameslamb: Alright, moved it in https://github.com/microsoft/LightGBM/pull/4161/commits/f93d7aae3f085c0e0c564db305694936773e4e44 since this is blocking all other PRs.

But I disagree with treating it as temporary and only doing this in the R package's tests. I think having every Mac build be 15 seconds longer is worth a reduced risk of problems like #4160 blocking PR merges in LightGBM for multiple days.

StrikerRUS: Hmm, I'm very surprised to see 15sec for `brew update`! I remember it took several minutes...
If this command is really executed within <1min nowadays, I'm totally fine to have `brew update` for **each build**.

StrikerRUS (approved):
LGTM! Feel free to move it back to `setup.sh` before merging given https://github.com/microsoft/LightGBM/pull/4161#discussion_r607049389.

.ci/setup.sh
jameslamb: I'll merge this for now to unblock other PRs, I can test the timings and submit another PR in the future to move this back to `setup.sh`. The update took only a few seconds on my Mac but that's not a fair test of how long it would take in CI.

PR #4164, opened 2021-04-06
[ci] Add debian-clang-devel CI job for the R package

jameslamb:
The R package failed on CRAN check on its most recent submission (documented in #4105).

This PR proposes a CI job that attempts to replicate that environment, so that similar problems can be caught before submitting to CRAN in the future.

Following the suggestion in https://github.com/microsoft/LightGBM/issues/4105#issuecomment-809004895, this tries to replicate CRAN's `r-devel-linux-x86_64-debian-clang` test, using the same container image that R Hub uses for that purpose.


jameslamb:
I was able to replicate the error from CRAN locally with this approach.

image

Pushing now to see if it's replicated on GitHub Actions as well. If it is, then I'll add another commit to try to actually patch the test so it will pass.

jameslamb:
Seems like this check is working! CUDA jobs are failing for probably-unrelated reasons (https://github.com/microsoft/LightGBM/pull/4158#issuecomment-813812070), but I can see that that new debian-clang-devel job failed with exactly the same error as seen on CRAN.

https://github.com/microsoft/LightGBM/pull/4164/checks?check_run_id=2275252015

jameslamb:
@StrikerRUS I received an email from Guolin yesterday. CRAN is asking us fix this problem and re-submit.

```text
Dear maintainer,

Please see the problems shown on
https://cran.r-project.org/web/checks/check_results_lightgbm.html.

Apparently your package fails its checks in a strict Latin-1* locale,
e.g. under Linux using LANG=en_US.iso88591 (see the debian-clang
results).

Please correct before 2021-04-21 to safely retain your package on CRAN.

Best,
-k
```

I think it's very important that `{lightgbm}` not get kicked off of CRAN even for a day, as that would break any users' automation that relies on installing `{lightgbm}` from CRAN.

I'll update this PR to make sure the package passes tests on `r-devel-linux-x86_64-debian-clang`.

@StrikerRUS which of these options would you prefer? I'm ok with whichever one you want, so we can come to a quick resolution well before April 21.

1. Submit the R package to CRAN as version `3.2.0.1` (this idea was rejected in #3210 but we could consider this a special case)
2. Do a 3.2.1 release of all of LightGBM

StrikerRUS:
@jameslamb I'd prefer variant 2. Luckly, we haven't merged any breaking changes that doesn't fit for minor release into `master` yet.

jameslamb:
> @jameslamb I'd prefer variant 2. Luckly, we haven't merged any breaking changes that doesn't fit for minor release into `master` yet.

ok sounds good! I'll get this PR working today, and I'll open a separate PR for the release with the usual release checklist.

jameslamb:
Ok I think the debian check will pass now!

jameslamb:
Ok I've updated this with all the blocking CI fixes that are now on `master` (#4161, #4167, #4168, #4158), so hopefully now it should be ready for a normal review with no need to use any admin privileges.

StrikerRUS:
I've pushed empty commit with the aim to re-run CI and not overwrite [important logs for Dask ranker error](https://github.com/microsoft/LightGBM/issues/3817#issuecomment-817126135) at the same time.

.github/workflows/r_package.yml
```diff
@@ -181,10 +181,33 @@ jobs:
...
+        with:
+          fetch-depth: 5
+          submodules: true
+      - name: Install packages and run tests
+        shell: bash
+        run: |
+          export PATH=/opt/R-devel/bin/:${PATH}
+          Rscript -e "install.packages(c('R6', 'data.table', 'jsonlite', 'testthat'), repos = 'https://cran.r-project.org')"
+          sh build-cran-package.sh
+          R CMD check --as-cran --run-donttest lightgbm_*.tar.gz || exit -1
```
StrikerRUS, line 206: Why don't we check for `ERRORS/WARNINGS/NOTES` here?
https://github.com/microsoft/LightGBM/blob/55a31bfe1a84f26bc65a4c6af3717155f0a6c8d7/.ci/test_r_package.sh#L184-L187

Especially, this looks critical given #3616.

BTW, the same question for `GCC ASAN/UBSAN` above.

R-package/tests/testthat/test_basic.R
```diff
@@ -1229,7 +1234,7 @@ test_that("lgb.train() supports non-ASCII feature names", {
   # UTF-8 strings are not well-supported on Windows
   # * https://developer.r-project.org/Blog/public/2020/05/02/utf-8-support-on-windows/
   # * https://developer.r-project.org/Blog/public/2020/07/30/windows/utf-8-build-of-r-and-cran-packages/index.html
-  if (!ON_WINDOWS) {
+  if (UTF8_LOCALE && !ON_WINDOWS) {
```
StrikerRUS, line 1237: 👍

StrikerRUS (changes requested):

.github/workflows/r_package.yml
jameslamb: **short answer**

it's an oversight, and we definitely should! I'll push that change here

**long answer**

All of that indirection where we decide whether or not to fail builds based on grepping through the R CMD check logs is primarily done so we have the chance to echo all of the more detailed logs out first: https://github.com/microsoft/LightGBM/blob/55a31bfe1a84f26bc65a4c6af3717155f0a6c8d7/.ci/test_r_package.sh#L178

When you run `R CMD check --as-cran`, any ERROR will result in a non-0 exit code.
 https://github.com/microsoft/LightGBM/pull/4164#issuecomment-814232054 demonstrates that this worked for a build with an ERROR. 

You're right that these things would miss WARNINGs or NOTEs on those other builds. I'll push a fix shortly.

jameslamb: Oh sorry and one other thing...for the UBSAN tests we're just running the unit tests directly with `Rscript testthat.R`, not running all of `R CMD check`.  So that test doesn't need to care about other R CMD check stuff.

jameslamb: I just pushed https://github.com/microsoft/LightGBM/pull/4164/commits/1ff93a6fce40856406d7f60893a0b514d09f6e9f, which adds code to fail the debian-clang build on WARNINGs or NOTEs, and intentionally added a documentation mistake that will cause a WARNING, so we can see that work

jameslamb: From the descriptions of CRAN's "extra" checks at https://cran.r-project.org/web/checks/check_issue_kinds.html, it looks like the UBSAN/ASAN tests run `R CMD check --use-valgrind`. So it would probably good for us to do that too.

But the feedback cycle when testing a change like that is very slow because R runs so much more slowly when instrumented with valgrind (those jobs often take around an hour to run). I'll write that up in a separate feature request. It isn't critical for this PR or the 3.2.1 release.

jameslamb: Ok I'm pretty confident the new change will fail builds on warnings: https://github.com/microsoft/LightGBM/pull/4164/checks?check_run_id=2313684939

```text
Status: 1 WARNING
See
  '/__w/LightGBM/LightGBM/lightgbm.Rcheck/00check.log'
for details.


NOTEs, WARNINGs, or ERRORs have been found by R CMD check
```

I'll push one change that will throw a NOTE to test that it also fails on NOTEs.

jameslamb: Here is a build failing on a NOTE:
https://github.com/microsoft/LightGBM/pull/4164/checks?check_run_id=2313911147

```text
Status: 1 NOTE
See
  '/__w/LightGBM/LightGBM/lightgbm.Rcheck/00check.log'
for details.


NOTEs, WARNINGs, or ERRORs have been found by R CMD check
```

StrikerRUS: Thank you very much for doing this!

.github/workflows/r_package.yml
```diff
@@ -8,6 +8,18 @@ on:
...
+  # hack to get around this:
+  # https://stat.ethz.ch/pipermail/r-package-devel/2020q3/005930.html
+  _R_CHECK_SYSTEM_CLOCK_: 0
+  # ignore R CMD CHECK NOTE checking how long it has
+  # been since the last submission
+  _R_CHECK_CRAN_INCOMING_REMOTE_: 0
+  # CRAN ignores the "installed size is too large" NOTE,
+  # so our CI can too. Setting to a large value here just
+  # to catch extreme problems
+  _R_CHECK_PKG_SIZES_THRESHOLD_: 100
```
StrikerRUS, line 21: I believe we can now drop these lines

https://github.com/microsoft/LightGBM/blob/9e1d7fa1bb395a9c4e5f7af9b13b9eb315e360fa/.ci/test_r_package_windows.ps1#L82-L93

StrikerRUS (changes requested):

.github/workflows/r_package.yml
jameslamb: oops yes! Forgot about the powershell script, thank you

jameslamb: removed in https://github.com/microsoft/LightGBM/pull/4164/commits/7f8547a7d361e5bbc5da6f032c8919282a86b5da

R-package/cran-comments.md
```diff
@@ -1,11 +1,27 @@
...
 
 ### CRAN response
 
 ### Maintainer Notes
 
+## v3.2.0 - Submission 1 - (March 22, 2021)
+
+### CRAN response
+
+Package is failing checks in the `r-devel-linux-x86_64-debian-clang` environment (described [here](https://cran.r-project.org/web/checks/check_flavors.html#r-devel-linux-x86_64-debian-clang). Specifically, one unit test on the use of non-ASCII feature names in `Booster$dump_model()` fails.
```
StrikerRUS, line 13:
```suggestion
Package is failing checks in the `r-devel-linux-x86_64-debian-clang` environment (described [here](https://cran.r-project.org/web/checks/check_flavors.html#r-devel-linux-x86_64-debian-clang)). Specifically, one unit test on the use of non-ASCII feature names in `Booster$dump_model()` fails.
```

StrikerRUS (approved):
Thanks a lot for improving testing CRAN environment coverage!

PR #4167, opened 2021-04-09
[ci] Stop running CUDA tests at CI

StrikerRUS:
Unblock LightGBM development process.
Refer to #4165.

I hope in the near future we will just revert this commit and restore running CUDA at our CI.

PR #4172, opened 2021-04-13
[ci] Restore CUDA jobs at CI

### LINKED ISSUE #4165: [ci] CUDA builds failing to start
## Description

For the last day, I've observed that CUDA builds (defined in https://github.com/microsoft/LightGBM/blob/b674439aec933e0c1b72efc7dda482a9c7e75283/.github/workflows/cuda.yml) are failing without logs.

Jobs just failed with "This check failed".

image

## Reproducible example

I've seen this across several pull requests so I don't think it's specific to any one PR's changes. See https://github.com/microsoft/LightGBM/actions/workflows/cuda.yml.

The most recent successful run was 23 hours ago, the first failed run was 21 hours ago.

image

## Additional Comments

I can't debug this because I don't think I have access to wherever the machine for this job is managed (https://github.com/microsoft/LightGBM/issues/3402#issuecomment-699198779).

I've labeled this `blocking` because it seems to be happening across all recent PRs.

StrikerRUS:
This reverts commit 36454a3b74c4941027d14df76687993f566efbe7.

Closed #4165.
Refer to #4167.

### DIFF
--- a/.github/workflows/cuda.yml
+++ b/.github/workflows/cuda.yml
@@ -1,0 +1,96 @@
+name: CUDA Version
+
+on:
+  push:
+    branches:
+    - master
+  pull_request:
+    branches:
+    - master
+
+env:
+  github_actions: 'true'
+  os_name: linux
+  task: cuda
+  conda_env: test-env
+
+jobs:
+  test:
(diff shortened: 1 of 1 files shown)


jameslamb (approved):
🚀🚀🚀

Issue #4136, opened 2021-03-29
Migrate to f-strings for the Python code


StrikerRUS:
Migrate all Python code from old-fashioned `format()` functions, formatting `%` operators and simple concatenations (`+`) to modern [`f-strings`](https://www.python.org/dev/peps/pep-0498/) ([brief guide](https://realpython.com/python-f-strings/)). They are known to be the fastest approach and also increase code readability.

![image](https://user-images.githubusercontent.com/25141164/112898582-aed66d00-90e9-11eb-8cdb-0bc67e69a643.png)
Source: https://cito.github.io/blog/f-strings/.

They are supported starting from Python 3.6 which is the oldest Python version LightGBM officially supports:
https://github.com/microsoft/LightGBM/blob/9cab93a9f3c0b46301a2cd11631223f4ab1daf16/python-package/setup.py#L374-L378

If you're interested in contributing, you do not need to fix all files at once. Pull requests that address all replacements in **one file** are mostly welcomed!

Refer to #4137 for the example PR.

- [x] ``.\.ci\get_workflow_status.py`` (https://github.com/microsoft/LightGBM/pull/4436/files#r663508231)
- [x] ``.\.nuget\create_nuget.py`` (#4162)
- [x] ``.\docs\conf.py`` (#4147)
- [x] ``.\examples\python-guide\advanced_example.py`` (#4386)
- [x] ``.\examples\python-guide\logistic_regression.py`` (#4356)
- [x] ``.\examples\python-guide\plot_example.py`` (#4421)
- [x] ``.\examples\python-guide\simple_example.py`` (#4396)
- [x] ``.\examples\python-guide\sklearn_example.py`` (#4396)
- [x] ``.\examples\python-guide\dask\binary-classification.py``
- [x] ``.\examples\python-guide\dask\multiclass-classification.py``
- [x] ``.\examples\python-guide\dask\prediction.py``
- [x] ``.\examples\python-guide\dask\ranking.py``
- [x] ``.\examples\python-guide\dask\regression.py``
- [x] ``.\helpers\check_dynamic_dependencies.py``
- [x] ``.\helpers\parameter_generator.py`` (#4146)
- [x] ``.\python-package\setup.py`` (#4145)
- [x] ``.\python-package\lightgbm\basic.py`` (#4143)
- [x] ``.\python-package\lightgbm\callback.py`` (#4142)
- [x] ``.\python-package\lightgbm\compat.py``
[...]

PR #945, opened 2017-09-29
[docs] move wiki to Read the Docs

StrikerRUS:
The main aim of this PR is to move remained wiki pages to `docs` folder and to **Read the Docs** in consequence. So all documentation is in one place and wiki could be deleted now.
Also some references have been added to easier navigation through the documentation.
In all README files headings have been refactored to the same type (all capital letters).

Please review `Installation-Guide.rst` with more attention - I added some sections about GUI compilation with Visual Studio there.
It would be great if someone could help me find out how to made Read the Docs deal with links with anchors and not to break everything.
Also, there is file `gcc-tips.Rmd` in `docs` folder on which there are no references. Where it's better to insert the reference to it?
And I want to ask @Laurae2 continue updating `LightGBM/docs/Key-Events.md` file - it's very useful thing.

PR #4175, opened 2021-04-14
Add alt text for gcc-tips

### LINKED ISSUE #4028: [docs] Add alt text on images

akshitadixit:
Closes #4028 

Added alt text for images in gcc-tips.rst, sorry for taking this long, was reading about things since I did not fully understand all of the images. Kindly review and suggest in case I missed on something.

### DIFF
--- a/docs/gcc-Tips.rst
+++ b/docs/gcc-Tips.rst
@@ -42,22 +42,22 @@ Some explanatory pictures:
 .. image:: ./_static/images/gcc-chart.png
    :align: center
    :target: ./_static/images/gcc-chart.png
-   :alt: this picture is a rendition of first where y-axis is cost and x-axis is number of threads comparing the four algorithms
+   :alt: a grid of vertical bar charts comparing run time for different combinations of max depth, compiler flags, and number of threads. The charts show that for shallow trees, using more threads is always expected to provide some reduction in run time. But for deeper trees (max depth greater than 10), using a value of num threads that is too high can actually result in slower training.
 
 .. image:: ./_static/images/gcc-comparison-1.png
    :align: center
    :target: ./_static/images/gcc-comparison-1.png
-   :alt: a bar chart comapring Light G.B.M performance versus compilation flags which is maximum at depth 8 with Light G.B.M version 2, flag O 3
+   :alt: a horizontal bar chart comparing Light G B M performance versus compilation flags. For most settings of max depth, best performance was achieved with flags dash O 3, dash M T U N E equals native.
 
 .. image:: ./_static/images/gcc-comparison-2.png
    :align: center
    :target: ./_static/images/gcc-comparison-2.png
-   :alt: a bar chart comapring Light G.B.M cumulative performance versus compilation flags which is maximum at depth 8 with Light G.B.M version 2, flag O 3
+   :alt: a set of 4 vertical bar charts comparing Light G B M performance versus compilation flags. For most settings of max depth, best performance was achieved with flags dash O 3, dash M T U N E equals native.
 
 .. image:: ./_static/images/gcc-meetup-1.png
    :align: center
    :target: ./_static/images/gcc-meetup-1.png
-   :alt: comparison of cumulative speed versus the slowest algorithm at various depths with v-2, O-3 remaining constant almost in all cases
+   :alt: grid of line charts showing the relative speed of training for different combinations of max depth, number of threads, and compilation flags. The grid shows that for models with max depth greater than 5, compiling Light G B M with default compiler flags produces faster training time than any of the customizations explored.
 
 .. image:: ./_static/images/gcc-meetup-2.png
    :align: center
--- a/docs/gcc-Tips.rst
+++ b/docs/gcc-Tips.rst
@@ -32,27 +32,34 @@ Some explanatory pictures:
 .. image:: ./_static/images/gcc-table.png
    :align: center
    :target: ./_static/images/gcc-table.png
+   :alt: picture comparing depth with number of threads with each implementation where red color means bad combination and greener areas mean good combination
 
 .. image:: ./_static/images/gcc-bars.png
    :align: center
    :target: ./_static/images/gcc-bars.png
+   :alt: picture of a simple bar chart against running time
 
 .. image:: ./_static/images/gcc-chart.png
    :align: center
    :target: ./_static/images/gcc-chart.png
+   :alt: this picture is a rendition of first where y-axis is cost and x-axis is number of threads comparing the four algorithms
 
 .. image:: ./_static/images/gcc-comparison-1.png
    :align: center
    :target: ./_static/images/gcc-comparison-1.png
+   :alt: a bar chart comapring Light G.B.M performance versus compilation flags which is maximum at depth 8 with Light G.B.M version 2, flag O 3
 
 .. image:: ./_static/images/gcc-comparison-2.png
    :align: center
    :target: ./_static/images/gcc-comparison-2.png
+   :alt: a bar chart comapring Light G.B.M cumulative performance versus compilation flags which is maximum at depth 8 with Light G.B.M version 2, flag O 3
 
 .. image:: ./_static/images/gcc-meetup-1.png
    :align: center
    :target: ./_static/images/gcc-meetup-1.png
+   :alt: comparison of cumulative speed versus the slowest algorithm at various depths with v-2, O-3 remaining constant almost in all cases
 
 .. image:: ./_static/images/gcc-meetup-2.png
    :align: center
    :target: ./_static/images/gcc-meetup-2.png
+   :alt: comparison of cumulative speed versus the slowest density of each algorithm at various depths with v-2, O-3 remaining constant almost in all cases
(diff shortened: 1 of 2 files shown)


docs/gcc-Tips.rst
```diff
@@ -32,27 +32,34 @@ Some explanatory pictures:
...
 
 .. image:: ./_static/images/gcc-chart.png
    :align: center
    :target: ./_static/images/gcc-chart.png
+   :alt: this picture is a rendition of first where y-axis is cost and x-axis is number of threads comparing the four algorithms
 
 .. image:: ./_static/images/gcc-comparison-1.png
    :align: center
    :target: ./_static/images/gcc-comparison-1.png
+   :alt: a bar chart comapring Light G.B.M performance versus compilation flags which is maximum at depth 8 with Light G.B.M version 2, flag O 3
```
jameslamb, line 50:
```suggestion
   :alt: a horizontal bar chart comparing Light G B M performance versus compilation flags. For most settings of max depth, best performance was achieved with flags dash O 3, dash M T U N E equals native.
```

I think the description for this one needs to be reworked. I don't think it's important to mention that the best performance on this particular problem was achieved with `max_depth=8`, since that's highly dependent on the dataset. The most important conclusion from this plot is that you should expect compiling with `-O3 -mtune=native` to give the best performance results for different settings of `max_depth`.

docs/gcc-Tips.rst
```diff
@@ -32,27 +32,34 @@ Some explanatory pictures:
...
 
 .. image:: ./_static/images/gcc-comparison-2.png
    :align: center
    :target: ./_static/images/gcc-comparison-2.png
+   :alt: a bar chart comapring Light G.B.M cumulative performance versus compilation flags which is maximum at depth 8 with Light G.B.M version 2, flag O 3
```
jameslamb, line 55:
```suggestion
   :alt: a set of 4 vertical bar charts comparing Light G B M performance versus compilation flags. For most settings of max depth, best performance was achieved with flags dash O 3, dash M T U N E equals native.
```

I think this could have the same description as the image above it, just with "horizontal" switched to "vertical".

docs/gcc-Tips.rst
```diff
@@ -32,27 +32,34 @@ Some explanatory pictures:
...
 
 .. image:: ./_static/images/gcc-meetup-1.png
    :align: center
    :target: ./_static/images/gcc-meetup-1.png
+   :alt: comparison of cumulative speed versus the slowest algorithm at various depths with v-2, O-3 remaining constant almost in all cases
```
jameslamb, line 60:
```suggestion
   :alt: grid of line charts showing the relative speed of training for different combinations of max depth, number of threads, and compilation flags. The grid shows that for models with max depth greater than 5, compiling Light G B M with default compiler flags produces faster training time than any of the customizations explored.
```

I think the story here is a little different. Will you consider this re-write?

docs/gcc-Tips.rst
```diff
@@ -32,27 +32,34 @@ Some explanatory pictures:
...
 
 .. image:: ./_static/images/gcc-meetup-2.png
    :align: center
    :target: ./_static/images/gcc-meetup-2.png
+   :alt: comparison of cumulative speed versus the slowest density of each algorithm at various depths with v-2, O-3 remaining constant almost in all cases
```
jameslamb, line 65: I think these stacked area plots (https://lightgbm.readthedocs.io/en/latest/_static/images/gcc-meetup-2.png) are really difficult to understand, and that this one should just be removed.

@StrikerRUS what do you think?

jameslamb (changes requested):
Thank you so much for taking this on @akshitadixit ! This page has received very little attention in the 3+ years since this content was first introduced, and I understand that it's pretty difficult to interpret some of the plots there, which don't have any explanatory text. This is the first time I've actually read this page and even I was pretty confused by some of it.

I left some suggestions below, but we'll also need to wait for another maintainer to comment on my questions about possibly removing some of these plots.

By the way...I noticed that you made this PR from the `master` branch on your fork. Since you're planning to help us with other things (https://github.com/microsoft/LightGBM/issues/4136#issuecomment-819274063), in the future you'll probably find it useful to create a new branch on your fork for each pull request. That way, you can have multiple pull requests into this project active at the same time.

StrikerRUS:
`gcc Tips` is @Laurae2's baby: #511, https://github.com/microsoft/LightGBM/pull/945#issuecomment-334923417. Maybe he would like to comment or update this doc.

python-package/lightgbm/sklearn.py
```diff
@@ -76,7 +76,7 @@ def __call__(self, preds, dataset):
         elif argc == 3:
             grad, hess = self.func(labels, preds, dataset.get_group())
         else:
-            raise TypeError("Self-defined objective function should have 2 or 3 arguments, got %d" % argc)
+            raise TypeError(f"Self-defined objective function should have 2 or 3 arguments, got {argc}")
```
jameslamb, line 79: Can you please remove these unrelated changes from this PR? I think they were meant for https://github.com/microsoft/LightGBM/pull/4181.

On whichever branch you use (based on my recommendation in https://github.com/microsoft/LightGBM/pull/4181), you can remove them with a command like `git reset --soft HEAD~1` followed by a force push.

Let me know if you get stuck!

akshitadixit:
What a relief to know I am not the only person who was confused after seeing the page. I almost gave up on the PR afraid to pull naive alt texts to such a project. I read your suggestions and realized what I misinterpreted. Thank you!

jameslamb:
> What a relief to know I am not the only person who was confused after seeing the page. I almost gave up on the PR afraid to pull naive alt texts to such a project. I read your suggestions and realized what I misinterpreted. Thank you!

Yes, sorry about that! I think that page could use some work. Thanks for pushing through it! Without your PR we probably would have not noticed that it needed some next explanatory text.

jameslamb:
ok @akshitadixit I have an update for you! I talked to @Laurae2 in another chat about this. Copying those comments below:

> * first pic = with different depth hyperparameter values and different number of threads, the difference between using -O2 -mtune=core2, -O2 -mtune=native, -O3 -mtune=native, and -O3 -fast-math -mtune=native is negligible, although -mtune=native is slightly faster (+2%)
> * second pic = same as first pic (smaller bar = faster, it is cumulated time)
> * third pic = -O2 -mtune=core2 mean time is larger than the others (hence slower), -O3 -mtune=native is the fastest
> * fourth pic = using -O2 -mtune=core2 as the baseline, all 3 other options are faster, with -O3 -mtune=native being the fastest
> * fifth pic = same as fourth pic, but chart is grouped by compiler set of options instead of depth
> * sixth pic = overall all performed tests, how much cumulated time (in percentage) you would save by using which compiler set of options
> * seventh pic = against overall possible time saved, how much each compiler set of options will save you in percentage

> the pics can be deleted since the most important is to tell gcc users to use -O3 -mtune=native and link to the experiment (the pics are only the details to not have to sift through 100+ slides, if really need a pic then 5th pic is best to keep


So given that, could you please change this pull request to remove all of the images EXCEPT for `gcc-comparison-2.png`? 

Sorry that you spent time on them and we're going to delete them but know that we wouldn't have made this change unless you brought it up! So this improvement of the docs is thanks to your efforts.

akshitadixit:
Closing this to fix all of the mess I created :sweat_smile:

PR #4176, opened 2021-04-14
[ci] cmake: remove linking to sanitizer library

cyfdecyf:
Adding compiler flag will automatically link to the required libraries.

Besides, clang static links to sanitizer libraries by default. Thus linking to dynamic library must be removed in order to use clang.

For reference:

- [Difference of AddressSanitizer in Clang and GCC](https://github.com/google/sanitizers/wiki/AddressSanitizerClangVsGCC-(5.0-vs-7.1))
- [StackOverflow answer](https://stackoverflow.com/a/40215639/306935) about `-lasan` option has been discouraged. It directs to mailing list answer by ASan developer
- [LeakSanitizer doc](https://github.com/google/sanitizers/wiki/AddressSanitizerLeakSanitizer#stand-alone-mode) mentions that standalone mode can work without the slowdown of ASan. This mode will link to "a runtime library containing just the bare necessities required for LeakSanitizer to work". For GCC,
  - when turning on both ASan and LSan, the executable only links with `libasan` without `liblsan`
  - when disabling ASan and enabling LSan, `liblsan` will be linked

### DIFF
--- a/cmake/Sanitizer.cmake
+++ b/cmake/Sanitizer.cmake
@@ -6,24 +6,16 @@
 # Add flags
 macro(enable_sanitizer sanitizer)
   if(${sanitizer} MATCHES "address")
-    find_package(ASan REQUIRED)
     set(SAN_COMPILE_FLAGS "${SAN_COMPILE_FLAGS} -fsanitize=address")
-    link_libraries(${ASan_LIBRARY})
 
   elseif(${sanitizer} MATCHES "thread")
-    find_package(TSan REQUIRED)
     set(SAN_COMPILE_FLAGS "${SAN_COMPILE_FLAGS} -fsanitize=thread")
-    link_libraries(${TSan_LIBRARY})
 
   elseif(${sanitizer} MATCHES "leak")
-    find_package(LSan REQUIRED)
     set(SAN_COMPILE_FLAGS "${SAN_COMPILE_FLAGS} -fsanitize=leak")
-    link_libraries(${LSan_LIBRARY})
 
   elseif(${sanitizer} MATCHES "undefined")
-    find_package(UBSan REQUIRED)
     set(SAN_COMPILE_FLAGS "${SAN_COMPILE_FLAGS} -fsanitize=undefined -fno-sanitize-recover=undefined")
-    link_libraries(${UBSan_LIBRARY})
 
   else()
     message(FATAL_ERROR "Santizer ${sanitizer} not supported.")
--- a/cmake/modules/FindASan.cmake
+++ b/cmake/modules/FindASan.cmake
@@ -1,9 +1,0 @@
-set(ASan_LIB_NAME ASan)
-
-find_library(ASan_LIBRARY
-  NAMES libasan.so libasan.so.5 libasan.so.4 libasan.so.3 libasan.so.2 libasan.so.1 libasan.so.0 libasan.so.0.0.0
-  PATHS ${SANITIZER_PATH} /usr/lib64 /usr/lib /usr/local/lib64 /usr/local/lib ${CMAKE_PREFIX_PATH}/lib)
-
-include(FindPackageHandleStandardArgs)
-find_package_handle_standard_args(ASan DEFAULT_MSG
-  ASan_LIBRARY)
--- a/cmake/modules/FindLSan.cmake
+++ b/cmake/modules/FindLSan.cmake
@@ -1,9 +1,0 @@
-set(LSan_LIB_NAME lsan)
-
-find_library(LSan_LIBRARY
-  NAMES liblsan.so liblsan.so.0 liblsan.so.0.0.0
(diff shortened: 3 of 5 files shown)


cyfdecyf:
The "Python-package / mpi source (macOS-latest, Python 3.9)" failure should not be caused by this PR.

jameslamb:
> The "Python-package / mpi source (macOS-latest, Python 3.9)" failure should not be caused by this PR.

hmmm ok yeah I suspect you're right.

When commenting on CI jobs, it's usually desirable to leave a link to the logs so maintainers can go look: https://github.com/microsoft/LightGBM/pull/4176/checks?check_run_id=2342784529.

```text
E   ImportError: 
E   
E   IMPORTANT: PLEASE READ THIS FOR ADVICE ON HOW TO SOLVE THIS ISSUE!
E   
E   Importing the numpy C-extensions failed. This error can happen for
E   many reasons, often due to issues with your setup or how NumPy was
E   installed.
E   
E   We have compiled some common reasons and troubleshooting tips at:
E   
E       https://numpy.org/devdocs/user/troubleshooting-importerror.html
E   
E   Please note and check the following:
E   
E     * The Python version is: Python3.9 from "/Users/runner/miniconda/envs/test-env/bin/python"
E     * The NumPy version is: "1.19.2"
E   
E   and make sure that they are the versions you expect.
E   Please carefully study the documentation linked above for further help.
E   
E   Original error was: dlopen(/Users/runner/miniconda/envs/test-env/lib/python3.9/site-packages/numpy/core/_multiarray_umath.cpython-39-darwin.so, 2): Library not loaded: @rpath/libopenblas.dylib
E     Referenced from: /Users/runner/miniconda/envs/test-env/lib/python3.9/site-packages/numpy/core/_multiarray_umath.cpython-39-darwin.so
E     Reason: image not found
```

Can you please push an empty commit to this branch to re-trigger CI? So it will rebuild but we won't lose the logs.

cyfdecyf:
@jameslamb Didn't realise that I can put a link commenting about CI.

Empty commit pushed now.

jameslamb:
> Didn't realise that I can put a link commenting about CI.

No problem! If you ever need to look, you can click on "Details" next to a check to see its logs.

image

> Empty commit pushed now

Thanks, seems like all checks are passing now 🎉

jameslamb:
I'm removing my request-for-review on this one, I don't think I'm qualified to approve it. Thanks very much for the help, @cyfdecyf !

If you have time, could you please update this to the latest `master`? Someone from the team will review it soon.

StrikerRUS:
Hey @hcho3 @trivialfis ! Could you please take a look at https://github.com/microsoft/LightGBM/pull/4176#issue-615182364?

trivialfis:
I think the changes make sense.  In older gcc versions linking those libraries was required.  But the requirement was removed in newer versions.  In fact, with the latest gcc distribution, those libraries are hidden deeply as internal shared objects so those CMake scripts will stop working.

cyfdecyf:
@jameslamb OK. I just did a rebase to latest master commit and force-pushed.

StrikerRUS:
@trivialfis Thanks a lot for your comment!

@cyfdecyf Could you please share versions of compilers you tested your changes from this PR with? We need to support quite wide range of versions of gcc and Clang. But gcc 4.8.4 is must-have because with this version we create public artifacts.

cyfdecyf:
@StrikerRUS Versions of compiler tested (all from Ubuntu 18.04):

- Clang 7.0.0
- GCC 7.5.0
- GCC 4.8.5
  - Note: seems only AddressSanitizer is supported for GCC 4.8.5. Must run cmake as `cmake -DUSE_SANITIZER=on -DENABLED_SANITIZERS=address`. Otherwise cmake will fail because unrecongnized command line option.
  - I guess the public artifacts build options won't enable sanitizers, so this PR shouldn't cause any problems in this regard
  - Besides, as it's better to use newer version of compiler when using sanitizers, I suggest just ignore this issue

StrikerRUS:
@cyfdecyf Thanks for the clarification!

> Note: seems only AddressSanitizer is supported for GCC 4.8.5.

Are you speaking about the current `master` branch or version in this PR?

> I guess the public artifacts build options won't enable sanitizers, so this PR shouldn't cause any problems in this regard

Yes, sure! Public artifacts are built with optimization flags and without any sanitizers.

> Besides, as it's better to use newer version of compiler when using sanitizers, I suggest just ignore this issue

Initially I thought about testing with sanitizers the same version of compiler with which public releases are made. To make sure that version of compiler doesn't find any issues. Now I believe your proposal to prefer testing with sanitizers only the latest compiler version makes sense. Let me think about it a little bit more time.

cyfdecyf:
> @cyfdecyf Thanks for the clarification!
> 
> > Note: seems only AddressSanitizer is supported for GCC 4.8.5.
> 
> Are you speaking about the current `master` branch or version in this PR?
>

It's the problem of GCC 4.8.5 itself. If we don't limit to only ASan during cmake invocation when using GCC 4.8.5, building will result compiling error because g++ doesn't recognize `-fsanitize=leak`, `-fsanitize=undefined` options.

StrikerRUS (approved):

PR #4177, opened 2021-04-14
fix: Dataset::CreateValid init fields which saves to binary.

cyfdecyf:
These fields are saved to binary file. Currently they are uninitialized and (maybe) not used.

Undefined Behavior Sanitizer will complain about assigning to bool variables (`use_missing_`, `zero_as_missing_`) whose value is not 0 or 1 in `DatasetLoader::LoadFromBinFile`.

### DIFF
--- a/src/io/dataset.cpp
+++ b/src/io/dataset.cpp
@@ -727,6 +727,11 @@ void Dataset::CreateValid(const Dataset* dataset) {
   feature_groups_.clear();
   num_features_ = dataset->num_features_;
   num_groups_ = num_features_;
+  max_bin_ = dataset->max_bin_;
+  min_data_in_bin_ = dataset->min_data_in_bin_;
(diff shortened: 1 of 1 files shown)


shiyu1994 (approved):

Willian-Zhang:
Note `examples/regression` would generate different bin for validation set before this commit.


handy script to verify result. 
```
rm -r build
mkdir build && cd build
cmake .. || exit 125
make -j8 || exit 125
cd -

cd examples/regression
rm -f *.bin
"../../lightgbm" config=train.conf task=save_binary
cd -
shasum ./examples/regression/*.bin > bin.sha
shasum -c bin.sha

cd examples/regression
rm -f *.bin
"../../lightgbm" config=train.conf task=save_binary
cd -
shasum -c bin.sha

```

Issue #4178, opened 2021-04-15
[mmlspark] how to deal with "Stopped training because there are no more leaves that meet the split requirements"


imatiach-msft:
## Description
Customer is seeing issue where LightGBM regressor in mmlspark is giving bad outputs with default parameters.  Investigating the issue, I found that LightGBM is outputting "[Warning] Stopped training because there are no more leaves that meet the split requirements".  The problem seems to be that we hit this code:

https://github.com/microsoft/LightGBM/blob/master/src/boosting/gbdt.cpp#L441

But then this last tree seems to cause junk predictions, values close to infinity:
[LightGBM] [Debug] Trained a tree with leaves = 1 and max_depth = 1

For example (true label vs what lightgbm predicted for target):
![image](https://user-images.githubusercontent.com/24683184/114812406-ebf35e00-9d7d-11eb-8f7f-e5d924451565.png)

Setting fewer iterations than the last iteration seems to resolve the issue.
Although I can detect that LGBM_BoosterUpdateOneIter is returning false when it hits this case, should I actually be setting the iteration on the booster to be the one before last, or is that something that lightgbm should be doing?  Or, should the last tree with a single node not be producing values close to infinity?

In python API I see we return that function's result:
https://github.com/microsoft/LightGBM/blob/master/python-package/lightgbm/basic.py#L2647
but it doesn't look like it is used or checked anywhere, including the dask implementation:
https://github.com/microsoft/LightGBM/blob/master/python-package/lightgbm/dask.py#L90
or python implementation:
https://github.com/microsoft/LightGBM/blob/master/python-package/lightgbm/engine.py#L249

## Reproducible example
Since this is in spark it will be difficult to reproduce, but I have access to a cluster with dummy data that reproduces the issue so I can validate any fixes.
Full output from lightgbm logs in spark:
[LightGBM] [Info] Trying to bind port 12400...
[LightGBM] [Info] Binding port 12400 succeeded
[LightGBM] [Info] Listening...
[LightGBM] [Info] Connected to rank 1
[LightGBM] [Info] Connected to rank 2
[LightGBM] [Info] Local rank: 0, total number of machines: 3
2021-04-15T03:15:39.704+0000: [GC (Allocation Failure) [PSYoungGen: 10960185K->213453K(11862528K)] 13872080K->3125356K(16915456K), 0.0432352 secs] [Times: user=0.31 sys=0.01, real=0.04 secs] 
[...]

PR #4185, opened 2021-04-16
[fix] Fix bug in data distributed learning with local empty leaf

shiyu1994:
In data distributed learning (including both `data` and `voting` modes), when the value distribution of a feature differs too much across machines (for example, two machines have totally nonoverlapped ranges of the same feature), there can be a leaf which contains 0 local data on a machine during training.

(Note that this happens more easily when `pre_partition=true`. With `pre_partition=false`, the data partition will be random and feature value distributions across machines are likely to be similar.)

In that case, the histogram buffers of that leaf won't be cleared because the `ConstructHistograms` method in `Dataset` will simply exit.
https://github.com/microsoft/LightGBM/blob/6ad3e6e9e306d3657a0a3d0201862d1f471b2a5e/include/LightGBM/dataset.h#L470-L478
Thus the histogram clear code in this method is not executed.
https://github.com/microsoft/LightGBM/blob/6ad3e6e9e306d3657a0a3d0201862d1f471b2a5e/src/io/dataset.cpp#L1192-L1193
But still, with data distributed learning, the histograms will be transferred to other machines to get the global histograms, for example, in `data` mode, we have
https://github.com/microsoft/LightGBM/blob/fba18e488ccd82b25c4e0788b2fec1f75bcde46a/src/treelearner/data_parallel_tree_learner.cpp#L159-L170
Since the histogram buffer is not cleared, content from previous iteration remains in the buffer, which results in a wrong global histogram after synchronization.

This PR is to guarantee that under data distributed learning, when an empty local leaf appears, its histogram content is cleared before synchronization. And it should fix issue #4026 and potentially issue #4178.

### DIFF
--- a/src/treelearner/data_parallel_tree_learner.cpp
+++ b/src/treelearner/data_parallel_tree_learner.cpp
@@ -155,6 +155,22 @@ template <typename TREELEARNER_T>
 void DataParallelTreeLearner<TREELEARNER_T>::FindBestSplits(const Tree* tree) {
   TREELEARNER_T::ConstructHistograms(
       this->col_sampler_.is_feature_used_bytree(), true);
+  const int smaller_leaf_index = this->smaller_leaf_splits_->leaf_index();
+  const data_size_t local_data_on_smaller_leaf = this->data_partition_->leaf_count(smaller_leaf_index);
+  if (local_data_on_smaller_leaf <= 0) {
+    // clear histogram buffer before synchronizing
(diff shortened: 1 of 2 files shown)


jameslamb (approved):
Explanation and changes make sense to me, thank you very much for the thorough explanation!

freetz-tiplu:
This fix does fix the case for data-parallel training but there are still some errors in the voting-parallel case as seen in #4414

PR #4187, opened 2021-04-16
[docs] Add changes to gcc-tips

akshitadixit:
* Removed all images except gcc-cpmparison-2.png
* Added alt text to it


docs/gcc-Tips.rst
```diff
@@ -27,32 +27,9 @@ You can find more details on the experimentation below:
...
    :target: ./_static/images/gcc-comparison-2.png
-
-.. image:: ./_static/images/gcc-meetup-1.png
-   :align: center
-   :target: ./_static/images/gcc-meetup-1.png
-
-.. image:: ./_static/images/gcc-meetup-2.png
-   :align: center
-   :target: ./_static/images/gcc-meetup-2.png
+   :alt: picture with a chart grouped by compiler set of options using O2 mtune=core2 as the baseline. All the other 3 options are faster, with O3 mtune=native being the fastest.
```
jameslamb, line 35:
```suggestion
   :alt: picture with a chart grouped by compiler set of options using O2 mtune=core2 as the baseline. All the other 3 options are faster, with O3 M tune equals native being the fastest.
```

I think it might help screen readers to break this phrase up, what do you think?

docs/gcc-Tips.rst
```diff
@@ -27,32 +27,9 @@ You can find more details on the experimentation below:
...
-   :target: ./_static/images/gcc-bars.png
-
-.. image:: ./_static/images/gcc-chart.png
-   :align: center
-   :target: ./_static/images/gcc-chart.png
-
-.. image:: ./_static/images/gcc-comparison-1.png
-   :align: center
-   :target: ./_static/images/gcc-comparison-1.png
+Here is an explanatory picture:
```
jameslamb, line 30:
```suggestion
The image below compares the runtime for training with different compiler options to a baseline using LightGBM compiled with `-O2 --mtune=core2`. All three options are faster than that baseline. The best performance was achieved with `-O3 --mtune=native`.
```

Could you expand this description? For figures like this, I like to have explanatory text that tells people what conclusion they're supposed to draw.

jameslamb (changes requested):
Thanks for doing this! Please see my comments below.

docs/gcc-Tips.rst
akshitadixit: I checked with a screen reader and only separating the 'M' works well, the equals to operator is being read aloud. Your suggestion is good to go if both occurences of mtune be changed to M tune.

jameslamb: ah right, thank you! I just edited the suggestion to work for both uses of `mtune`.

jameslamb:
The two CI failures were unrelated to this PR, and I just restarted both of them manually. So I think this can be merged once https://github.com/microsoft/LightGBM/pull/4187#discussion_r614971908 is resolved.

akshitadixit:
I just wish I can submit a clean PR for once. Thank you for the patience. I learnt why every senior dev always told me to create new branches and avoid working on master directly.

jameslamb (approved):
This looks good to me, thanks very much!

@StrikerRUS in case you didn't see on the previous PR, the deletions in this PR were done with @Laurae2 's permission: https://github.com/microsoft/LightGBM/pull/4175#issuecomment-820748957

> I just wish I can submit a clean PR for once.

Don't even worry about it! If you look back through mine or other maintainers' pull requests here, you'll see that even we (people who have been working on this project for years) very rarely have a PR that is merged without any requested changes. It's just a part of the process.

StrikerRUS:
@akshitadixit Please remove actually files from the following folder references for which were removed in this PR.
https://github.com/microsoft/LightGBM/tree/master/docs/_static/images

StrikerRUS (approved):
LGTM! @akshitadixit We really appreciate your great help!

PR #4144, opened 2021-03-30
[python] added f-strings to python-package/lightgbm/dask.py

NovusEdge:

PR #4188, opened 2021-04-16
[python] Migrate to f-strings in python-package/lightgbm/sklearn.py

akshitadixit:
Switches some string formatting in `sklearn.py` to use f-strings, as part of #4136.

### DIFF
--- a/python-package/lightgbm/sklearn.py
+++ b/python-package/lightgbm/sklearn.py
@@ -532,9 +532,8 @@ class LGBMModel(_LGBMModelBase):
         """
         for key, value in params.items():
             setattr(self, key, value)
-            under_score = "_"
-            if hasattr(self, f"{under_score} {key}"):
-                setattr(self, f"{under_score} {key}", value)
+            if hasattr(self, f"_{key}"):
+                setattr(self, f"_{key}", value)
             self._other_params[key] = value
         return self
 
@@ -720,7 +719,7 @@ class LGBMModel(_LGBMModelBase):
             X = _LGBMCheckArray(X, accept_sparse=True, force_all_finite=False)
         n_features = X.shape[1]
         if self._n_features != n_features:
-            raise ValueError(f"Number of features of the model must "
+            raise ValueError("Number of features of the model must "
                              f"match the input. Model n_features_ is {self._n_features} and "
                              f"input n_features is {n_features}")
         return self._Booster.predict(X, raw_score=raw_score, start_iteration=start_iteration, num_iteration=num_iteration,
@@ -919,10 +918,9 @@ class LGBMClassifier(LGBMModel, _LGBMClassifierBase):
         """Docstring is set after definition, using a template."""
         result = super().predict(X, raw_score, start_iteration, num_iteration, pred_leaf, pred_contrib, **kwargs)
         if callable(self._objective) and not (raw_score or pred_leaf or pred_contrib):
-            new_line = "\n"
-            _log_warning(f"Cannot compute class probabilities or labels "
-                         f"due to the usage of customized objective function.{new_line}"
-                         f"Returning raw scores instead.")
+            _log_warning("Cannot compute class probabilities or labels "
+                         "due to the usage of customized objective function.\n"
+                         "Returning raw scores instead.")
             return result
         elif self._n_classes > 2 or raw_score or pred_leaf or pred_contrib:
             return result
@@ -993,7 +991,7 @@ class LGBMRanker(LGBMModel):
                    + _base_doc[_base_doc.find('eval_init_score :'):])  # type: ignore
     _base_doc = fit.__doc__
     _before_early_stop, _early_stop, _after_early_stop = _base_doc.partition('early_stopping_rounds :')
-    fit.__doc__ = (f"{_before_early_stop}"
-                   f" eval_at : iterable of int, optional (default=(1, 2, 3, 4, 5))\n"
-                   f"{' ':12} The evaluation positions of the specified metric.\n"
-                   f"{' ':8} {_early_stop} {_after_early_stop}")
+    fit.__doc__ = (_before_early_stop
+                   "eval_at : iterable of int, optional (default=(1, 2, 3, 4, 5))\n"
+                   f"{' ':12}The evaluation positions of the specified metric.\n"
+                   f"{' ':8}{_early_stop}{_after_early_stop}")
--- a/python-package/lightgbm/sklearn.py
+++ b/python-package/lightgbm/sklearn.py
@@ -532,8 +532,9 @@ class LGBMModel(_LGBMModelBase):
         """
         for key, value in params.items():
             setattr(self, key, value)
-            if hasattr(self, '_' + key):
-                setattr(self, '_' + key, value)
+            under_score = "_"
+            if hasattr(self, f"{under_score} {key}"):
+                setattr(self, f"{under_score} {key}", value)
             self._other_params[key] = value
         return self
 
@@ -652,7 +653,7 @@ class LGBMModel(_LGBMModelBase):
                 elif isinstance(collection, dict):
                     return collection.get(i, None)
                 else:
-                    raise TypeError('{} should be dict or list'.format(name))
+                    raise TypeError(f"{name} should be dict or list")
 
             if isinstance(eval_set, tuple):
                 eval_set = [eval_set]
@@ -719,9 +720,9 @@ class LGBMModel(_LGBMModelBase):
             X = _LGBMCheckArray(X, accept_sparse=True, force_all_finite=False)
         n_features = X.shape[1]
         if self._n_features != n_features:
-            raise ValueError(f"Number of features of the model must " \
-                             f"match the input. Model n_features_ is {self._n_features} and " \
-                             f"input n_features is {n_features} ")
+            raise ValueError(f"Number of features of the model must "
+                             f"match the input. Model n_features_ is {self._n_features} and "
+                             f"input n_features is {n_features}")
         return self._Booster.predict(X, raw_score=raw_score, start_iteration=start_iteration, num_iteration=num_iteration,
                                      pred_leaf=pred_leaf, pred_contrib=pred_contrib, **kwargs)
 
@@ -919,8 +920,8 @@ class LGBMClassifier(LGBMModel, _LGBMClassifierBase):
         result = super().predict(X, raw_score, start_iteration, num_iteration, pred_leaf, pred_contrib, **kwargs)
         if callable(self._objective) and not (raw_score or pred_leaf or pred_contrib):
             new_line = "\n"
-            _log_warning(f"Cannot compute class probabilities or labels " \
-                         f"due to the usage of customized objective function.{new_line}" \
+            _log_warning(f"Cannot compute class probabilities or labels "
+                         f"due to the usage of customized objective function.{new_line}"
                          f"Returning raw scores instead.")
             return result
         elif self._n_classes > 2 or raw_score or pred_leaf or pred_contrib:
@@ -992,7 +993,7 @@ class LGBMRanker(LGBMModel):
                    + _base_doc[_base_doc.find('eval_init_score :'):])  # type: ignore
     _base_doc = fit.__doc__
     _before_early_stop, _early_stop, _after_early_stop = _base_doc.partition('early_stopping_rounds :')
-    fit.__doc__ = (_before_early_stop
-                   + 'eval_at : iterable of int, optional (default=(1, 2, 3, 4, 5))\n'
-                   + ' ' * 12 + 'The evaluation positions of the specified metric.\n'
-                   + ' ' * 8 + _early_stop + _after_early_stop)
+    fit.__doc__ = (f"{_before_early_stop}"
+                   f" eval_at : iterable of int, optional (default=(1, 2, 3, 4, 5))\n"
+                   f"{' ':12} The evaluation positions of the specified metric.\n"
+                   f"{' ':8} {_early_stop} {_after_early_stop}")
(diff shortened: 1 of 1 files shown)


python-package/lightgbm/sklearn.py
```diff
@@ -719,10 +719,9 @@ def predict(self, X, raw_score=False, start_iteration=0, num_iteration=None,
             X = _LGBMCheckArray(X, accept_sparse=True, force_all_finite=False)
         n_features = X.shape[1]
         if self._n_features != n_features:
-            raise ValueError("Number of features of the model must "
-                             "match the input. Model n_features_ is %s and "
-                             "input n_features is %s "
-                             % (self._n_features, n_features))
+            raise ValueError(f"Number of features of the model must " \
+                             f"match the input. Model n_features_ is {self._n_features} and " \
+                             f"input n_features is {n_features} ")
```
jameslamb, line 724:
```suggestion
            raise ValueError("Number of features of the model must "
                             f"match the input. Model n_features_ is {self._n_features} and "
                             f"input n_features is {n_features} ")
```

Can you please remove these backslashes? They are not necessary. This is why the linting job is complaining. Also, the first string (which doesn't have any templating stuff with `{}`) does not need to be an f-string.

https://github.com/microsoft/LightGBM/pull/4188/checks?check_run_id=2362245578

```text
Linting Python code
./python-package/lightgbm/sklearn.py:722:71: E502 the backslash is redundant between brackets
./python-package/lightgbm/sklearn.py:723:95: E502 the backslash is redundant between brackets
./python-package/lightgbm/sklearn.py:922:75: E502 the backslash is redundant between brackets
./python-package/lightgbm/sklearn.py:923:90: E502 the backslash is redundant between brackets
```

In case you're not familiar with it, multiple consecutive string literals within parentheses will be interpreted as a single string by Python.

```python
one_line = "it is cloudy"
multi_line = (
    "it "
    "is "
    "cloudy"
)

assert one_line == multi_line
print(multi_line)
```

python-package/lightgbm/sklearn.py
```diff
@@ -919,9 +918,10 @@ def predict_proba(self, X, raw_score=False, start_iteration=0, num_iteration=Non
         """Docstring is set after definition, using a template."""
         result = super().predict(X, raw_score, start_iteration, num_iteration, pred_leaf, pred_contrib, **kwargs)
         if callable(self._objective) and not (raw_score or pred_leaf or pred_contrib):
-            _log_warning("Cannot compute class probabilities or labels "
-                         "due to the usage of customized objective function.\n"
-                         "Returning raw scores instead.")
+            new_line = "\n"
+            _log_warning(f"Cannot compute class probabilities or labels " \
+                         f"due to the usage of customized objective function.{new_line}" \
+                         f"Returning raw scores instead.")
```
jameslamb, line 924:
```suggestion
            _log_warning("Cannot compute class probabilities or labels "
                         "due to the usage of customized objective function.\n"
                         "Returning raw scores instead.")
```

I don't think this one needs any changes, actually. Since nothing is being templated in it.

jameslamb (changes requested):
Thanks for picking this up! Please see my comments below.

I also made one small change to the PR. I added a note to the description mentioning #4136, so that there is a link connecting this PR to a previous issue and vice versa. We use that linking feature heavily in this project, to keep track of the connections between different discussions.

python-package/lightgbm/sklearn.py
akshitadixit: I was trying to use a multi-line f-string, but I do see where it wasn't required.

python-package/lightgbm/sklearn.py
```diff
@@ -919,6 +918,7 @@ def predict_proba(self, X, raw_score=False, start_iteration=0, num_iteration=Non
         """Docstring is set after definition, using a template."""
         result = super().predict(X, raw_score, start_iteration, num_iteration, pred_leaf, pred_contrib, **kwargs)
         if callable(self._objective) and not (raw_score or pred_leaf or pred_contrib):
+            new_line = "\n"
```
jameslamb, line 921: I should have included this in my suggestion, sorry. I think this can be removed as well.

```suggestion
```

jameslamb (approved):
looks good, thank you!

python-package/lightgbm/sklearn.py
```diff
@@ -720,9 +720,8 @@ def predict(self, X, raw_score=False, start_iteration=0, num_iteration=None,
         n_features = X.shape[1]
         if self._n_features != n_features:
             raise ValueError("Number of features of the model must "
-                             "match the input. Model n_features_ is %s and "
-                             "input n_features is %s "
-                             % (self._n_features, n_features))
+                             f"match the input. Model n_features_ is {self._n_features} and "
+                             f"input n_features is {n_features} ")
```
StrikerRUS, line 724:
```suggestion
                             f"input n_features is {n_features}")
```

python-package/lightgbm/sklearn.py
```diff
@@ -720,9 +720,8 @@ def predict(self, X, raw_score=False, start_iteration=0, num_iteration=None,
...
         return self._Booster.predict(X, raw_score=raw_score, start_iteration=start_iteration, num_iteration=num_iteration,
                                      pred_leaf=pred_leaf, pred_contrib=pred_contrib, **kwargs)
```
StrikerRUS, line 727: Please update the following lines within this PR as well
https://github.com/microsoft/LightGBM/blob/211ef7878f0c8a72757fd2b0050029908bc3091a/python-package/lightgbm/sklearn.py#L535-L536
https://github.com/microsoft/LightGBM/blob/211ef7878f0c8a72757fd2b0050029908bc3091a/python-package/lightgbm/sklearn.py#L997-L998 (part about `' ' * x`)
https://github.com/microsoft/LightGBM/blob/211ef7878f0c8a72757fd2b0050029908bc3091a/python-package/lightgbm/sklearn.py#L655

StrikerRUS (changes requested):

python-package/lightgbm/sklearn.py
akshitadixit: @StrikerRUS do I only change the `' '*x` part or do I also modify the concatenation taking place there into an f-string as below?

<pre>
space = " "
    fit.__doc__ = (f"{_before_early_stop}" \
                  f"eval_at : iterable of int, optional (default=(1, 2, 3, 4, 5))\n" \
                  f"{space*12} The evaluation positions of the specified metric.\n" \
                  f"{space*8} {_early_stop} {_after_early_stop}")
</pre>

StrikerRUS: Originally I thought about only `' '*x` part, but just saw that @jameslamb would prefer modifying concatenation as well: https://github.com/microsoft/LightGBM/pull/4144#discussion_r610993091. So, please modify it with concatenation.

BTW, there are some other places where concatenation can be eliminated with f-strings:
https://github.com/microsoft/LightGBM/blob/211ef7878f0c8a72757fd2b0050029908bc3091a/python-package/lightgbm/sklearn.py#L705-L711
https://github.com/microsoft/LightGBM/blob/211ef7878f0c8a72757fd2b0050029908bc3091a/python-package/lightgbm/sklearn.py#L825-L831
https://github.com/microsoft/LightGBM/blob/211ef7878f0c8a72757fd2b0050029908bc3091a/python-package/lightgbm/sklearn.py#L898-L902
https://github.com/microsoft/LightGBM/blob/211ef7878f0c8a72757fd2b0050029908bc3091a/python-package/lightgbm/sklearn.py#L990-L992
But I don't think we should replace concatenation with f-strings there because concatenation is more readable in those cases in my opinion. @jameslamb WDYT?


Also, please check the following comment about padding https://github.com/microsoft/LightGBM/pull/4144#discussion_r610994830.

jameslamb: > But I don't think we should replace concatenation with f-strings there because concatenation is more readable in those cases in my opinion

Yes I agree with you.

akshitadixit: So I infer that no concatenations be migrated then.

jameslamb: yes, cases like the ones @StrikerRUS mentioned (where `.format()` was used to replace a shared template like `_lgbmmodel_doc_fit` or where there are just variables being added to each other but not string literals) do not need to be changed for this pull request

python-package/lightgbm/sklearn.py
```diff
@@ -532,8 +532,9 @@ def set_params(self, **params):
         """
         for key, value in params.items():
             setattr(self, key, value)
-            if hasattr(self, '_' + key):
-                setattr(self, '_' + key, value)
+            under_score = "_"
+            if hasattr(self, f"{under_score} {key}"):
+                setattr(self, f"{under_score} {key}", value)
```
jameslamb, line 537:
```suggestion
            if hasattr(self, f"_{key}"):
                setattr(self, f"_{key}", value)
```

I don't think it's necessary to pull underscore out into its own variable. f-strings can be a mix of literal strings and Python expressions.

Also note that the `+` operator for strings does not add a space. So the equivalent of `'_' + key` is `f"_{key}"`, not `f"_ {key}"`

python-package/lightgbm/sklearn.py
```diff
@@ -719,10 +720,9 @@ def predict(self, X, raw_score=False, start_iteration=0, num_iteration=None,
             X = _LGBMCheckArray(X, accept_sparse=True, force_all_finite=False)
         n_features = X.shape[1]
         if self._n_features != n_features:
-            raise ValueError("Number of features of the model must "
-                             "match the input. Model n_features_ is %s and "
-                             "input n_features is %s "
-                             % (self._n_features, n_features))
+            raise ValueError(f"Number of features of the model must "
```
jameslamb, line 723:
```suggestion
            raise ValueError("Number of features of the model must "
```

This first string is a string literal without any templating, so it doesn't require an `f`

python-package/lightgbm/sklearn.py
```diff
@@ -919,9 +919,10 @@ def predict_proba(self, X, raw_score=False, start_iteration=0, num_iteration=Non
         """Docstring is set after definition, using a template."""
         result = super().predict(X, raw_score, start_iteration, num_iteration, pred_leaf, pred_contrib, **kwargs)
         if callable(self._objective) and not (raw_score or pred_leaf or pred_contrib):
-            _log_warning("Cannot compute class probabilities or labels "
-                         "due to the usage of customized objective function.\n"
-                         "Returning raw scores instead.")
+            new_line = "\n"
+            _log_warning(f"Cannot compute class probabilities or labels "
+                         f"due to the usage of customized objective function.{new_line}"
+                         f"Returning raw scores instead.")
```
jameslamb, line 925:
```suggestion
            _log_warning("Cannot compute class probabilities or labels "
                         "due to the usage of customized objective function.\n"
                         "Returning raw scores instead.")
```

It looks like the unnecessary `new_line` variable from https://github.com/microsoft/LightGBM/pull/4188/files#r615043256 was added back. I think it can be removed, and strings without any templating do not need an `f`.

python-package/lightgbm/sklearn.py
```diff
@@ -992,7 +993,7 @@ def fit(self, X, y,
...
     _base_doc = fit.__doc__
     _before_early_stop, _early_stop, _after_early_stop = _base_doc.partition('early_stopping_rounds :')
-    fit.__doc__ = (_before_early_stop
-                   + 'eval_at : iterable of int, optional (default=(1, 2, 3, 4, 5))\n'
-                   + ' ' * 12 + 'The evaluation positions of the specified metric.\n'
-                   + ' ' * 8 + _early_stop + _after_early_stop)
+    fit.__doc__ = (f"{_before_early_stop}"
+                   f" eval_at : iterable of int, optional (default=(1, 2, 3, 4, 5))\n"
+                   f"{' ':12} The evaluation positions of the specified metric.\n"
+                   f"{' ':8} {_early_stop} {_after_early_stop}")
```
jameslamb, line 999:
```suggestion
    fit.__doc__ = (_before_early_stop
                   "eval_at : iterable of int, optional (default=(1, 2, 3, 4, 5))\n"
                   f"{' ':12}The evaluation positions of the specified metric.\n"
                   f"{' ':8}{_early_stop}{_after_early_stop}")
```

I think these changes have introduced some extra spaces. It's also not necessary to add an f-string around a value that is a single variable, e.g. `_before_early_stop` does not need to be changed to `f"{_before_early_stop}"`. In that case, the templating with f-strings isn't changing anything about the string.

jameslamb (changes requested):
Please see my new round of comments, and let me know if you have any questions! If it seems like we're giving you confusing or contradictory suggestions I apologize.

python-package/lightgbm/sklearn.py
```diff
@@ -993,6 +992,6 @@ def fit(self, X, y,
     _base_doc = fit.__doc__
     _before_early_stop, _early_stop, _after_early_stop = _base_doc.partition('early_stopping_rounds :')
     fit.__doc__ = (_before_early_stop
-                   + 'eval_at : iterable of int, optional (default=(1, 2, 3, 4, 5))\n'
-                   + ' ' * 12 + 'The evaluation positions of the specified metric.\n'
-                   + ' ' * 8 + _early_stop + _after_early_stop)
+                   "eval_at : iterable of int, optional (default=(1, 2, 3, 4, 5))\n"
```
jameslamb, line 995: ah! I'm so sorry @akshitadixit , I misled you in my previous comment. To concatenate like this, I guess you do need to do `f"{_before_early_stop}"`. It isn't possible to concatenate a variable and a string literal like this.

```python
a = "hello"
x = (
    a
    " world"
)
```

Could you please change `_before_early_stop` to `f"{_before_early_stop}"`?

Sorry again, totally my fault for asking you to change that.

jameslamb (changes requested):

python-package/lightgbm/sklearn.py
akshitadixit: Not a problem, I did think of checking it up but then laid back on it.

jameslamb (approved):
looks great, thank you!

StrikerRUS (approved):
LGTM!

PR #4190, opened 2021-04-17
[docs] add working dir to R package docker run examples

jameslamb:
There are a few examples in the R-package documentation which rely on the use of Docker containers. These involve mounting in the current working directory, and then running `cd` to change directories to that location.

This PR proposes a minor change to remove that manual use of `cd`, just to make those examples slightly easier to use.

```shell
# before: run the container and then cd
docker run \
    -v $(pwd):/opt/LightGBM \
    -it ubuntu:latest \
    cd /opt/LightGBM && do-something

# after: run the container with working directory set to the location LightGBM is mounted in at
docker run \
    -v $(pwd):/opt/LightGBM \
    -w /opt/LightGBM \
    -it ubuntu:latest \
    do-something
```

### DIFF
--- a/R-package/README.md
+++ b/R-package/README.md
@@ -300,8 +300,9 @@ At build time, `configure` will be run and used to create a file `Makevars`, usi
     ```shell
     docker run \
         -v $(pwd):/opt/LightGBM \
+        -w /opt/LightGBM \
         -t ubuntu:20.04 \
-        /bin/bash -c "cd /opt/LightGBM && ./R-package/recreate-configure.sh"
+        ./R-package/recreate-configure.sh
     ```
 
     The version of `autoconf` used by this project is stored in `R-package/AUTOCONF_UBUNTU_VERSION`. To update that version, update that file and run the commands above. To see available versions, see https://packages.ubuntu.com/search?keywords=autoconf.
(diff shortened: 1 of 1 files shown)


R-package/README.md
```diff
@@ -300,8 +300,9 @@ At build time, `configure` will be run and used to create a file `Makevars`, usi
     ```shell
     docker run \
         -v $(pwd):/opt/LightGBM \
+        -w /opt/LightGBM \
         -t ubuntu:20.04 \
-        /bin/bash -c "cd /opt/LightGBM && ./R-package/recreate-configure.sh"
+        ./R-package/recreate-configure.sh
```
StrikerRUS, line 305: Is it working without `/bin/bash`?

StrikerRUS (approved):
Thanks!

R-package/README.md
jameslamb: yep! The file is executable and has the shebang `#!/bin/bash` at the top. https://github.com/microsoft/LightGBM/blob/dac774935cbadf4082bdc431b18b96c3fa21c8b1/R-package/recreate-configure.sh#L1

I tested this before pushing the PR, so I'm confident in it.

PR #4194, opened 2021-04-18
[R-package] Make returned feature importances from lgb.importance() visible by default

david-cortes:
Feature importances are returned as invisible. I don't get why would anyone want them invisible. This PR fixes it.

### DIFF
--- a/R-package/R/lgb.importance.R
+++ b/R-package/R/lgb.importance.R
@@ -75,6 +75,7 @@ lgb.importance <- function(model, percentage = TRUE) {
     )]
   }
 
-  return(withVisible(tree_imp_dt)$value)
+  # adding an empty [] to ensure the table is printed the first time print.data.table() is called
+  return(tree_imp_dt[])
 
 }
--- a/R-package/R/lgb.importance.R
+++ b/R-package/R/lgb.importance.R
@@ -75,6 +75,6 @@ lgb.importance <- function(model, percentage = TRUE) {
     )]
   }
 
-  return(tree_imp_dt)
+  return(withVisible(tree_imp_dt)$value)
 
 }


jameslamb (changes requested):
Thank you for your interest in LightGBM! But I am -1 on this change. `withVisible()` adds overhead that is unnecessary to meet the behavior you're trying to achieve.

From `?withVisible`

> This function evaluates an expression, returning it in a two element list containing its value and a flag showing whether it would automatically print.

I'm opposed to adding the overhead of constructing that list when it isn't clear how the behavior as of this pull request would be different than the current behavior of `{lightgbm}`.

`lgb.importance()` already returns a `data.table` of importances without using `invisible()`. If you've found that `data.table`'s default printing behavior is not what you want, you can assign the result of `lgb.importance()` to a variable in your code and call `print()` on it directly.

jameslamb:
@david-cortes please correct me if I've misunderstood the effect or intention of this PR.

david-cortes:
Thing is, even you assign the result to an object, it will not print it the first time you try to do so, because it is invisible. I'm attaching a video.
lgbm_invisible


BTW, the overhead is measured in microseconds:
```r
library(data.table)
library(microbenchmark)
set.seed(1)
X <- as.data.table(rnorm(1e4*1e3), nrow=1e4)
make_inv <- function(x) invisible(x)
make_vis <- function(x) withVisible(x)$value
Xinv <- make_inv(X)
microbenchmark({Xvis <- make_vis(Xinv)})
```
```
Unit: microseconds
                           expr   min    lq     mean median     uq      max neval
 {     Xvis <- make_vis(Xinv) } 1.473 1.503 16.09723 1.5075 1.5325 1444.489   100
```

jameslamb:
Thanks for that investigation!

I just tested and I don't think this PR will actually provide the behavior you're looking for. When you use `:=` expressions in a `data.table`, it will not be printed the first time it's returned.

Try running the code below:

```r
f <- function(){
    someDT <- data.table::data.table(
        x = rnorm(1e4)
        , y = rnorm(1e4)
    )
    someDT[, z := rnorm(.N)]
    return(someDT)
}

f2 <- function(){
    someDT <- data.table::data.table(
        x = rnorm(1e4)
        , y = rnorm(1e4)
    )
    someDT[, z := rnorm(.N)]
    return(withVisible(someDT)$value)
}

f3 <- function(){
    someDT <- data.table::data.table(
        x = rnorm(1e4)
        , y = rnorm(1e4)
    )
    someDT[, z := rnorm(.N)]
    return(someDT[])
}

# does not print
f()

# does not print
f2()

# prints
f3()
```

I'm ok with adding a workaround to ensure printing of the result for this function. Since that result is a `data.table` we can trust that it will only print a handful of rows no matter how big it is, and the number of columns will always be the same 4 (`Gain`, `Cover`, `Frequency`, `Feature`).

I'll leave a suggestion on how to achieve the behavior you're looking for.

R-package/R/lgb.importance.R
```diff
@@ -75,6 +75,6 @@ lgb.importance <- function(model, percentage = TRUE) {
     )]
   }
 
-  return(tree_imp_dt)
+  return(withVisible(tree_imp_dt)$value)
```
jameslamb, line 78: I don't think this will have the desired behavior (see https://github.com/microsoft/LightGBM/pull/4194#issuecomment-822023021).

I believe it will do what you want with this suggestion, based on https://stackoverflow.com/a/32989328/3986677

```suggestion
  # adding an empty [] to ensure the table is printed the first time print.data.table() is called
  return(tree_imp_dt[])
```

jameslamb (changes requested):

jameslamb (approved):
Can confirm that this is working. I built `{lightgbm}` locally from this branch, like so:

```shell
sh build-cran-package.sh
R CMD INSTALL lightgbm_*.tar.gz
```

Then ran the following sample code

```r
library(lightgbm)

train <- agaricus.train

model <- lgb.train(
    params = list(
        objective = "binary"
        , max_depth = -1L
        , num_leaves = 31L
        , min_data_in_leaf = 1L
    )
    , data = dtrain
    , nrounds = 5L
)

lgb.importance(model, percentage = TRUE)
```

And I can see that the importance table has the correct shape and contents, and is printed immediately.

image

Thanks very much for bringing it up, @david-cortes .

PR #4197, opened 2021-04-18
[R-package] change default nrounds to 100 to match LightGBM core library default

### LINKED ISSUE #4196: Default arguments differ between interfaces
In the Python version of lightgbm, the default number of boosting rounds is 100 (which is reasonable), whereas in the R version the default is 10 (which is not reasonable).

Would be nice if both could have the same default arguments.

(I haven't checked the rest of the arguments so I don't know if there are more discrepancies)

david-cortes:
Fixes #4196

The R interface has some default values for arguments related to training and evaluation which differ from Python, namely:
* Number of boosting rounds (100 in python, 10 in R).
* Number of cross-validation folds (5 in python, 3 in R).

This PR changes the default values in R to match those of Python.


R-package/R/lgb.cv.R
```diff
@@ -71,8 +71,8 @@ CVBooster <- R6::R6Class(
 #' @export
 lgb.cv <- function(params = list()
                    , data
-                   , nrounds = 10L
-                   , nfold = 3L
+                   , nrounds = 100L
+                   , nfold = 5L
```
jameslamb, line 75:
```suggestion
                   , nfold = 3L
```

Please revert this change to `nfold`.

I'm open to updating the R package's defaults to match the defaults from LightGBM's core library (https://lightgbm.readthedocs.io/en/latest/Parameters.html), but the number of folds to use in k-fold cross-validation should be left to individual wrapper packages. It would be a nice side benefit to have closer alignment between the R and Python package but I don't support making breaking changes for only that purpose.

In other words, I support this PR not as "make the R package more closely match the Python package" but "make the R package more closely match the LightGBM core library".

I've changed the PR title accordingly.

jameslamb (changes requested):
Thanks very much for your interest in LightGBM and attention to detail! Please see my comment below about `nfold`.

jameslamb (approved):
Thanks for this!

The failing CI job is not related to your changes. I've manually restarted the build, and will merge this when it passes.

jameslamb:
Before I merge, I want to check. @StrikerRUS are you ok with this change?

StrikerRUS:
I'm totally fine with this fix! I remember I spent quite large amount of time syncing Python, native and docs param values several years ago. I'm glad that someone has started working of R-package now!
I'd like to note that the next release of LightGBM will be major `4.0.0`, so we're free to change anything we need/want.

jameslamb:
Ok thanks!

PR #4199, opened 2021-04-19
[docs][R-package] Update the explanation of num_threads (fixes #4192)

### LINKED ISSUE #4192: [R-package] Quick question about num of thread
Hi, Thank you for making lightGBM in R!

I am using LightGBM in R and have a quick question about the `num_thread`. 

According to the [manual](https://lightgbm.readthedocs.io/en/latest/R/reference/lightgbm.html), the number of threads is the physical core of CPU. But usually what I have seen in my R code, set the num thread is equal to num_thread - 1 such as

```r
cores <- parallel::detectCores() -1
cores
```

So if we have 4 core, use 3 for parallel and 1 for the controller. Is this applied to LightGBM too? So if I have 4 physical core of CPU and set the number of thread as 3?

issactoast:
Hi @jameslamb 

Here is the PR related to the issue #4192. Please check out the update!

Best,
Issac

### DIFF
--- a/R-package/R/lgb.cv.R
+++ b/R-package/R/lgb.cv.R
@@ -46,8 +46,9 @@ CVBooster <- R6::R6Class(
 #'                \item{\code{max_depth}: Limit the max depth for tree model. This is used to deal with
 #'                                 overfit when #data is small. Tree still grow by leaf-wise.}
 #'                \item{\code{num_threads}: Number of threads for LightGBM. For the best speed, set this to
-#'                                   the number of real CPU cores, not the number of threads (most
-#'                                   CPU using hyper-threading to generate 2 threads per CPU core).}
+#'                             the number of real CPU cores(\code{parallel::detectCores(logical = FALSE)}),
+#'                             not the number of threads (most CPU using hyper-threading to generate 2 threads
+#'                             per CPU core).}
 #'            }
 #' @inheritSection lgb_shared_params Early Stopping
 #' @return a trained model \code{lgb.CVBooster}.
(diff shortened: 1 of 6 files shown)


jameslamb (changes requested):
Thanks very much for this!

Sorry I didn't check earlier, but there are two other places with this same bit of documentation. Can you please update these as well?

* https://github.com/microsoft/LightGBM/blob/69d02e387a6f4516b991665c9f6ac1378c695492/R-package/R/lgb.cv.R#L48
* https://github.com/microsoft/LightGBM/blob/69d02e387a6f4516b991665c9f6ac1378c695492/R-package/R/lgb.train.R#L21

issactoast:
@jameslamb Sure, will do that, but just wonder what's the difference between LightGBM/lightgbm_r and LightGBM/R-package? I know the latter is the r package but what's the lightgbm_r? That folder has the same three files lgb.cv.R, lgb.train.R, and lightgbm.R

jameslamb:
Oh sure, I can explain. The R package wraps C/C++ code stored in `src/`, `include/`, and `external_libs/`. So to build the R package, you have to copy all that code into the correct location inside the R package's file structure.

To do that, the R build scripts (`build-cran-package.sh` for autoconf-based builds and `build_r.R` for CMake-based builds) create a new folder, `lightgbm_r`, and copy `R-package/` plus all those C/C++ files into it at build time. I think it will be clearer if you inspect those build scripts.

`lightgbm_r` is not actually checked into the repo, as it is ignored by `gitignore`. So you do not need to change aything in that directory.

Let me know if you have other questions! We'd love to have you contribute more to the R package in the future of you're interested, and understanding this setup might help you do that.

issactoast:
Thank you for the explanation! Yes, I would like to add some contributions to the LGBM if I could! :) I will ask if I have some questions related to the package!

jameslamb (approved):
Looks great, thanks very much!

PR #4202, opened 2021-04-19
[R-package] clarify parameter documentation (fixes #4193)

### LINKED ISSUE #4193: [R-package] docs for lgb.Dataset in R are incomplete
In the R package, the function `lgb.Dataset` has a parameter `categorical_feature`, with this description:
```
categorical features
```
https://github.com/microsoft/LightGBM/blob/7ea2bc4dbca53c03eaae2aa2b22f9b30447caea4/R-package/R/lgb.Dataset.R#L736

It does not describe what kind of input it needs to be, in contrast to `lgb.Dataset.set.categorical`.

jameslamb:
This PR proposes some clarifications in the R documentation, based on the feedback in #4193.

### Notes for Reviewers

Pushing to the branch  `docs/jlamb` so we can check the rendered docs on readthedocs.

### DIFF
--- a/R-package/R/lgb.train.R
+++ b/R-package/R/lgb.train.R
@@ -5,19 +5,20 @@
 #' @param valids a list of \code{lgb.Dataset} objects, used for validation
 #' @param record Boolean, TRUE will record iteration message to \code{booster$record_evals}
 #' @param colnames feature names, if not null, will use this to overwrite the names in dataset
-#' @param categorical_feature list of str or int
-#'                            type int represents index,
-#'                            type str represents feature names
+#' @param categorical_feature categorical features. This can either be a character vector of feature
+#'                            names or an integer vector with the indices of the features (e.g.
+#'                            \code{c(1L, 10L)} to say "the first and tenth columns").
 #' @param callbacks List of callback functions that are applied at each iteration.
 #' @param reset_data Boolean, setting it to TRUE (not the default value) will transform the
 #'                   booster model into a predictor model which frees up memory and the
 #'                   original datasets
-#' @param ... other parameters, see Parameters.rst for more information. A few key parameters:
+#' @param ... other parameters, see \href{https://lightgbm.readthedocs.io/en/latest/Parameters.html}{
+#'            The "Parameters" section of the documentation} for more information. A few key parameters:
 #'            \itemize{
 #'                \item{\code{boosting}: Boosting type. \code{"gbdt"}, \code{"rf"}, \code{"dart"} or \code{"goss"}.}
 #'                \item{\code{num_leaves}: Maximum number of leaves in one tree.}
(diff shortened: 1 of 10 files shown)


R-package/R/lgb.train.R
```diff
@@ -5,19 +5,20 @@
...
-#'                            type str represents feature names
+#' @param categorical_feature categorical features. This can either be a character vector of feature
+#'                            names or an integer vector with the indices of the features (e.g.
+#'                            \code{c(1L, 10L)} to say "the first and tenth columns").
 #' @param callbacks List of callback functions that are applied at each iteration.
 #' @param reset_data Boolean, setting it to TRUE (not the default value) will transform the
 #'                   booster model into a predictor model which frees up memory and the
 #'                   original datasets
-#' @param ... other parameters, see Parameters.rst for more information. A few key parameters:
+#' @param ... other parameters, see \href{https://lightgbm.readthedocs.io/en/latest/Parameters.html}{
```
StrikerRUS, line 15: Just a question: how this differs from `params` (the first one) argument?

StrikerRUS (approved):

R-package/R/lgb.train.R
jameslamb: unfortunately, they both can take the exact same information. And then they get merged together: https://github.com/microsoft/LightGBM/blob/8e126c80ba72d63c078b1c166e2b0b04126fee67/R-package/R/lgb.train.R#L85-L86.

That has been a part of this project's R package since before I joined LightGBM, so for several years. Now that we're saying that the next release will be 4.0.0 and breaking changes will be allowed, I would love to get rid of support for `...` and force people to be more explicit. I think it introduces extra complexity for no benefit.

jameslamb: There is a very real functional downside of allowing `...`, I should have mentioned it in my comment. If you misspell one of the keyword arguments, it will just be assumed that you are referring to a LightGBM parameter.

So if you use, say, `lgb.train(resetting_data = TRUE)`, training will proceed with `reset_data` still set to the default (`TRUE`) and you'll get a warning in logs like `[LightGBM ] [WARN] unrecognized parameter: 'resetting_data'`.

If we didn't support `...`, then you'd get an error (not a warning) in R immediately like "unrecognized argument: `resetting_data`", which I think is desirable.

StrikerRUS: > I would love to get rid of support for `...`

Yeah, good idea I think!

jameslamb: I'll add a feature request. And `@` some of the people who've opened R issues here in the past for their thoughts.

jameslamb: I just created https://github.com/microsoft/LightGBM/issues/4226

PR #4205, opened 2021-04-19
added aliases to params

StrikerRUS:
I always forget which param is plural and which is not. 😄


jameslamb (approved):
Looks ok to me! I used `git grep` to search for any cases in wrappers where we have checks like `if "extra_trees" in params` or something (which would require adding code to be sure that aliases are respected) and didn't find any.

PR #4101, opened 2021-03-23
[dask] add support for eval sets and custom eval functions

ffineis:
Followup PR regarding #3952 - implements `eval_set` functionality for lightgbm.dask but *without* early stopping.

This is implemented this to work with all eval-* parameters:
- multiple eval sets (i.e. multiple (X, y) pairs in `eval_set`)
- `eval_names`
- `eval_sample_weight`
- `eval_class_weight` (for `DaskLGBMClassifier` only)
- `eval_init_score` 
- `eval_group` (for `DaskLGBMRanker` only)

When an individual eval_set, eval_sample_weight, eval_init_score, or eval_group is the same as (data, label)/sample_weight/init_score/group, just use the latter instead of having to `compute` the training set/weights/init_score/group multiple times.

This is all that's going on, making little mini eval sets out of delayed parts in a consistent manner:
distributed_eval_sets

Other things to know:
- Raises warning when a worker does not receive any `eval_set` parts. This check is now performed prior to `client.submit` found in `_train` function. Model training still completes in this scenario, but depending on which worker returns its `futures_classifier`, `best_score_` and `evals_result_` attributes can be empty or contain data. Moreover, when a worker is missing `eval_set` entirely, this will fail out once `early_stopping_rounds` becomes supported - local worker calls to `model.fit(..., eval_data=None, early_stopping_rounds=x)` will throw a exception. 
- See [comment](https://github.com/microsoft/LightGBM/pull/3952#pullrequestreview-610403805) as to why early_stopping_rounds is explicitly omitted from this PR.
- This adds 32 tests! I'm happy to trim it down a little bit to 16.

PR #4214, opened 2021-04-21
[dask] Fix typo mentioned in 4101

ffineis:
Fixes typo in dask _train as mentioned in [this comment](https://github.com/microsoft/LightGBM/pull/4101#discussion_r616275252).

### DIFF
--- a/python-package/lightgbm/dask.py
+++ b/python-package/lightgbm/dask.py
@@ -422,7 +422,7 @@ def _train(
 
     # Tell each worker to train on the parts that it has locally
     #
-    # This code creates ``_train_part()`` calls as not "pure" because:
+    # This code treats ``_train_part()`` calls as not "pure" because:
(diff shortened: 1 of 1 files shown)


python-package/lightgbm/dask.py
```diff
@@ -422,7 +422,7 @@ def _train(
 
     # Tell each worker to train on the parts that it has locally
     #
-    # This code treates ``_train_part()`` calls as not "pure" because:
+    # This code creates ``_train_part()`` calls as not "pure" because:
```
jameslamb, line 425: thank you! Actually, I meant to write "treats". Like "it treats the calls as not pure"

```suggestion
    # This code treats ``_train_part()`` calls as not "pure" because:
```

jameslamb (approved):
thank you!

Issue #4216, opened 2021-04-22
(R) Functions that replace R headers are not compatible with all R versions


david-cortes:
Migrated from here: https://github.com/microsoft/LightGBM/pull/4207

The R interface for LightGBM has replacements for R's internal headers which re-define the structs used by R's variables. These underlying structures behind R objects are however not constant between versions, and currently the LighGBM replacements will not work correctly with R 3.6 for windows and mac (which from the requirements is meant to be supported, and is tested in the CI pipelines in github here).

Example:
```r
library(lightgbm)
data(mtcars)
X <- as.matrix(mtcars[,-1])
y = mtcars[,1,drop=TRUE]
m = lightgbm(X, y, params=list(objective="regression", min_data=1))
pred_before = predict(m, X)
mode(X) = "double"
pred_after = predict(m, X)
```
```r
print(pred_before)
```
```
 [1] 20.68292 20.68292 21.85530 20.94345 19.18488 18.79409 16.31907 22.89741 21.85530 19.51054 18.59869 17.68684 18.27303
[14] 16.90526 13.77891 13.77891 16.57959 28.10799 26.80534 29.08497 21.00858 17.10065 16.90526 15.66774 19.51054 24.78625
[27] 23.93953 26.80534 17.29605 19.83620 16.77499 20.94345
```
```r
print(pred_after)
```
```
 [1] 26.80534 26.80534 26.80534 26.80534 26.80534 26.80534 26.80534 26.80534 26.80534 26.80534 26.80534 26.80534 26.80534
[14] 26.80534 26.80534 26.80534 26.80534 26.80534 26.80534 26.80534 26.80534 26.80534 26.80534 26.80534 26.80534 26.80534
[27] 26.80534 26.80534 26.80534 26.80534 26.80534 26.80534
```

(This was tested with R 3.6.3 and RTools35 on windows, but the problem will not be present if tested with R 4.0.4 in linux)

More concretely, what's failing there is the function `R_REAL_PTR(lightbm_struct)` (meant to obtain a pointer to the array of `double`s underneath an R matrix or vector), which does not output the same as the R C function `REAL(SEXP)`, returning instead a pointer to the address of some other array or struct. This not only causes incorrect results but can also lead to segmentation faults.

For now I think an easy solution would be to require R >= 4.0.0, but this is not a real solution as the structs can change at any time as the R headers are not meant to be replaced by libraries.

jameslamb:
[...]

PR #4225, opened 2021-04-26
[R-package] fix warnings in unit tests

jameslamb:
This PR addresses several (but not all) of the R unit test warnings documented in #4108.

**`Unknown parameter: valids / {memory_address}`**

```text
[LightGBM] [Warning] Unknown parameter: 0x7fb4ee5c84d8>
[LightGBM] [Warning] Unknown parameter: valids
```

These warnings were caused by calls that were passing `valids` to `lgb.cv()`.
`lgb.cv()` does not accept validation data...it creates holdout sets using k-fold cross-validation.

This is an example of something that is currently a warning which would be an error if `...`
was not allowed (https://github.com/microsoft/LightGBM/pull/4202#discussion_r616075433, ~will write up a feature request soon~ https://github.com/microsoft/LightGBM/issues/4226).

**`Unknown parameter: categorical_features`**

```text
[LightGBM] [Warning] Unknown parameter: categorical_features
```

**`There are no meaningful features`**

```text
[LightGBM] [Warning] There are no meaningful features, as all feature values are constant
```

This came from the test `lgb.train() works with linear learners and data where a feature has only 1 non-NA value`. That test should have added an additional feature with enough variability to be usable during training.

### Notes for Reviewers

#4108 was originally opened as a `good first issue` about a month ago, with the hope that an outside contributor would pick it up. Normally I'd leave such an issue for non-maintainers. However, I want to focus on the R package over the next few weeks because of some issues that have been building up about it.

* #4007
* #4034
* #4045
* #4216 (from https://github.com/microsoft/LightGBM/pull/4207#issuecomment-824455972)
* https://github.com/tlverse/sl3/issues/344 & https://github.com/tlverse/sl3/pull/345
* https://github.com/curso-r/treesnip/issues/38

Given that I expect a lot of activity on the R package in the near future, I think it's very important that the R package's unit tests cover the package as reliably as possible.

### DIFF
--- a/R-package/tests/testthat/test_basic.R
+++ b/R-package/tests/testthat/test_basic.R
@@ -1582,9 +1582,6 @@ test_that("If first_metric_only is TRUE, lgb.cv() decides to stop early based on
     , data = DTRAIN_RANDOM_REGRESSION
     , nfold = nfolds
     , nrounds = nrounds
-    , valids = list(
-      "valid1" = DVALID_RANDOM_REGRESSION
-    )
     , eval = list(
       .increasing_metric
       , .constant_metric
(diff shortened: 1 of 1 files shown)


StrikerRUS (approved):
LGTM!

PR #4089, opened 2021-03-21
[python-package] Create Dataset from multiple data files 

cyfdecyf:
This PR implements creating Dataset from multiple data files.

We add a `Sequence` interface which requires data providers implement:

- Return total number of data
- Random access
  - so we can sample data without going through data in two round
- Range access
  - used to read data in batch thus avoid loading all data into memory

With this interface, we can create `Dataset` from `List[Sequence]`. Also refer this to #2789 as I think this interface allows we use other types of efficient binary data format. I closed #3900 because it's an ugly hack and data loading features are much easier to implement in Python.

An example of creating `Dataset` from multiple HDF5 files is also given. As I mentioned in #2788, I'm more familiar with HDF5 file so it's choosen as the example.

 #2031 also mentions the feature request, guess this is a common feature request.

PR #4228, opened 2021-04-26
Regression Tests and gitlab CI support

Willian-Zhang:
> This PR depends on https://github.com/microsoft/LightGBM/pull/4089

> `Regression` here stands for testing method as described [here](https://en.wikipedia.org/wiki/Regression_testing) . To not be confused with term `regression` tasks in ML realm. 

This PR:
- Add basic regression test supports
  - create report for all tests 
  - many to many relationship between testing scripts and verified datas
  - multiple regression tests
  - sha1 and color report
- Gitlab CI Support
  - to build lightgbm on the fly 
  - run regression tests

Example Result: [on gitlab](https://gitlab.com/Willian-Zhang/LightGBM/-/jobs/1210982418)

Caveats/Future Plans:
- Gitlab CI Support depends on a [image](https://github.com/Willian-Zhang/LightGBM-build-docker/blob/main/Dockerfile) I created [willianz/lightgbm-build-env:latest](https://hub.docker.com/repository/docker/willianz/lightgbm-build-env), this won't effect current upstream CI (Microsoft) whose CI were fulled on GitHub.
- Regression Tests not yet added to GitHub CI
- Current check implementation relies on local data generation for referencing data
  - as regression logics (caching and moving data around) should be more dependent on CI tool side instead of checking side
  - and there would be a pending decision on how Add-Remove-Change effects CI reports (are they strictly prohibited)
- This PR depends on https://github.com/microsoft/LightGBM/pull/4089
  - Considering there are some local regression tests yet under development against Sequence feature


jameslamb (changes requested):
Thanks for your interest in LightGBM!

But I don't understand this pull request, sorry. Is there some previous discussion that you've had with maintainers here that describes this work?

My initial observations:

* I'm -1 on supporting new CI pipelines on GitLab. This project already has extensive testing on GitHubAction, AppVeyor, and Azure DevOps and I'd like to understand why it's necessary to add GitLab.
* It's not clear to me why the new tests added in this PR need to be in their own folder called `regression/` instead of being included in LightGBM's existing testing infrastructure. I'd like to understand why that is being proposed.

Willian-Zhang:
@jameslamb 
Sorry for not making this clear.


Intention for this PR is to purpose what LightGBM seems to be lacking: **Regression test that ensures LightGBM produces same outputs (intermediate results, models, predictions, etc) over commits or releases**. 
This is especially useful on project of this magnitude, specially when building, testing with large datasets takes huge amount of time or resources, while CI could step up offload those heavy tasks asynchronously during development cycle.


GitLab CI here serves as a demo. We have been testing legally protected datasets (e.g. Yahoo), as well as large ones (e.g. Higgs) on self hosted server. GitLab CI integrates well to our current testing infrastructure, also does not affect current public Github CI.


However considerations are: Regressions tests on a ML framework could be very different to traditional ones that requires each in-output to be  strictly unchanged, restrictions could be added where needed or instead prompting user for change if not fatal.
Despite running procedure could be arbitrary (C++, Python, R or any combination of those routine) and reporting mechanism is strongly sighted to CI tool adopted, It's really up to your design decision where restrictions or suggestion to put on regression changes.




### TL;DR
1. Gitlab is just a proposal, it does not affect current GitHub CI procedure.
2. Please consider adding Regression tests.

jameslamb:
> Intention for this PR is to purpose what LightGBM seems to be lacking: Regression test that ensures LightGBM produces same outputs (intermediate results, models, predictions, etc) over commits or releases.

Thanks for the clarification.

Even though there isn't a separate directory in this project called `regression_tests`, many of the project's tests meet that definition...they are intended to catch breaking changes. For example, consider these tests that check for exact measures of metrics from training on a fixed  dataset:

* https://github.com/microsoft/LightGBM/blob/26cde5f513d58ad567da7817d76da1418087d3df/R-package/tests/testthat/test_basic.R#L172-L173
* https://github.com/microsoft/LightGBM/blob/26cde5f513d58ad567da7817d76da1418087d3df/R-package/tests/testthat/test_basic.R#L572-L573

We'd welcome a discussion about **specific** types of tests that the project is missing today, and then small, focused pull requests to add them one at a time.

Could you please close this pull request and open a feature request at https://github.com/microsoft/LightGBM/issues describing the types of tests that you believe LightGBM is missing?

Please be as specific as possible, for example:

* tests that a model file written in version `X` can be used in version `Y`
* tests that a Dataset file written by version `X` can be read by version `Y`
* etc.

This pull request in its current state needs significant changes before it is likely to be accepted by maintainers, so I think this topic would benefit from some discussion first.

StrikerRUS:
I agree with @jameslamb . There is no doubt that regression tests are very important. However, according to the provided definition, 
> **Regression testing** (rarely non-regression testing) is re-running functional and non-functional tests to ensure that previously developed and tested software still performs after a change.

such tests can be written in quite different formats, not necessary as a special GitLab jobs.

I believe we already have regression tests for
**metrics**, 
https://github.com/microsoft/LightGBM/blob/4530deda28063003c86925d9c5303b325492524f/tests/python_package_test/test_engine.py#L169
https://github.com/microsoft/LightGBM/blob/4530deda28063003c86925d9c5303b325492524f/tests/python_package_test/test_engine.py#L2662
**internal model structure**,
https://github.com/microsoft/LightGBM/blob/4530deda28063003c86925d9c5303b325492524f/tests/python_package_test/test_basic.py#L47-L48
**printed outputs**
https://github.com/microsoft/LightGBM/blob/4530deda28063003c86925d9c5303b325492524f/tests/python_package_test/test_utilities.py#L97

All these asserts check that LightGBM performs the same way it did in previous commit. I guess this is what regression tests are about.

I believe we should add more such tests with hardcoded expected results and make them more strict (for metrics on test datasets, for example) and pay more attention to PRs which change that values (#4349). Contributions are very welcome 😉 !

Sorry, but I think this PR in its' current state is not going to be merged.

jameslamb:
I've created two issues to track specific pieces of work that I think are relevant to the discussion in this pull request.

* #4406: regression tests on binary dataset files
* #4407: regression tests on model files

Since it has been almost two months since https://github.com/microsoft/LightGBM/pull/4228#issuecomment-830706993 with no comment from non-maintainers on this pull request, I am closing this pull request and locking the conversation.

For anyone arriving at this discussion who is interested in contributing such regression tests, please comment on #4406 / #4407 or open a new issue.

Issue #4065, opened 2021-03-12
Include init_score on the Python Booster class


JoshuaC3:
## Summary

Add `init_score_values` as a parameter on the Python [Booster](https://github.com/microsoft/LightGBM/blob/master/python-package/lightgbm/basic.py#L2158) class. Currently, it is only available on the [Dataset](https://github.com/microsoft/LightGBM/blob/master/python-package/lightgbm/basic.py#L1051) class.

## Motivation

Currently, it is a little tricky to find `init_score` after a model is trained. While it can be set, and therefore known, when [`boost_from_average`](https://lightgbm.readthedocs.io/en/latest/Parameters.html#boost_from_average) is isn't. When saving a model, the easy access to this presumably gets lost. Therefore, it would be useful to have it as a parameter on the model.

It then allows us to access it via the sklearn API.

## Description

See [here](https://github.com/microsoft/LightGBM/issues/3905) for further details and initial motivation.

JoshuaC3:
Possibly something like,

```
    for dataset_name, eval_name, score, _ in evaluation_result_list:
        booster.best_score[dataset_name][eval_name] = score

    booster.init_score_value = trainset.get_init_score()

    if not keep_training_booster:
        booster.model_from_string(booster.model_to_string(), False).free_dataset()
```

[here](https://github.com/microsoft/LightGBM/blob/master/python-package/lightgbm/engine.py#L269).

StrikerRUS:
Closed in favor of being in #2302. We decided to keep all feature requests in one place.

Welcome to contribute this feature! Please re-open this issue (or post a comment if you are not a topic starter) if you are actively working on implementing this feature.

Issue #4234, opened 2021-04-27
Add BoostFromAverage to Python model.


JoshuaC3:
## Summary

Add the BoostFromAverage output to the init_score on the train_set Dataset. It seems to me that they serve the same function - that is, reducing training time by allowing the model to have some previous knowledge of the data before the first boosting round. BoostingFromAverage is a simply a subset of the `init_score` functionality.

## Motivation

This allows us to recover the mean of the target variable after training and then when saving the model. There are many instances where we may not have access to the label data (and therefore the mean) after saving a model.

This is also the equivalent to the intercept in EBMs so aids in comparisons of LightGBM and EBM. See https://github.com/microsoft/LightGBM/issues/3905#issuecomment-797136404 and [EBM intercept calculations](https://github.com/interpretml/interpret/blob/ac5e727498cbdcb6911444420343454b0f328a46/python/interpret-core/interpret/glassbox/ebm/ebm.py#L1021).

Finally, and maybe most importantly, this would allow a user better insights into how the model is working at each stage and would help with model interpretability #4065 and #3095.

## Description

Something like, writing the results from `BoostFromAverage` back to the model here
https://github.com/microsoft/LightGBM/blob/d517ba12f2e7862ac533908304dddbd770655d2b/src/boosting/gbdt.cpp#L344

StrikerRUS:
Superseded by #4313 as a more general approach.

Issue #3905, opened 2021-02-03
Investigate possibility of borrowing some features from Explainable Boosted Machines


JoshuaC3:
## Summary

Borrow ideas from InterpretMLs Explainable Boosting Machine to make LGBM more interpretable as well as more comparable to their EBM.

## Description

I am sure you are somewhat aware of [InterpretMLs Explainable Boosting Machine](https://github.com/interpretml/interpret) - also a Microsoft innovation!

I have been a long time user of LGBM and now an recent fan of EBM. Clearly, they are similar in ways and each have their strong points and weaknesses. There are trade-offs of choosing either. That said, I have been playing around with some settings in LGBM to make the final models behave more like the EBMs. The reasons for this are two fold:
1) Increase LGBM interpretability
2) Allow more direct comparisons of results/functionality.

Some tips, tricks and findings are as follows:

Setting `n_estimators` high, `learning_rate` low, and `num_leaves` low, begins to mimic some of the behaviours of the EBM. That is, iteratively building LOTS of VERY shallow trees, VERY slowly. I add feature constraints so that each tree is univariate. Pairwise interactions could also be added where needed. This allows the model to learn incrementally small amounts from each feature, rather like the EBM.  Essentially, replicating a model of the form:

`lgb = F0(X0) + F1(X1) + ... + Fn(Xn)`

which is effectively a GLM/GAM.

```python
lgb = LGBMRegressor(
    n_estimators=5000, #large
    learning_rate=0.01, #small
    num_leaves=4, #shallow
    interaction_constraints=[[i] for i in range(X.shape[1])], #[[i, j] for i in range(X.shape[1]) for j in range(X.shape[1]) if i != j]
#     feature_contri=[10 for i in range(X.shape[1])], #use to reduce single feature dependence?
#     monotone_constraints=[1 for i in range(X.shape[1] - 1)] + [-1], #use for expert knowledge. I used in my temperature vs gas use case. temp -> gas monotonically decreasing.
)
``` 

To me, however, there seems to be two relatively simple features that could be added to facilitate this further.
1) The addition of a Constant Term.
2) Schedulable Feature Cycling.

#### The Constant Term

This would allow the model to remove the extra/redundant initial "value"* from the first set of splits on the first feature. This is the same as shifting your y-variable (e.g. lgb.fit(X, y - C), but until the initial run, you do not know what C is.

#### Schedulable Feature Cycling

[...]

PR #4235, opened 2021-04-27
Add init_score attr to python booster class.

JoshuaC3:
Save `init_score` to the python booster class so that it is accessible after training, after saving and as part of model interpretability: #4065

### DIFF
--- a/python-package/lightgbm/engine.py
+++ b/python-package/lightgbm/engine.py
@@ -267,6 +267,7 @@ def train(params, train_set, num_boost_round=100,
             evaluation_result_list = earlyStopException.best_score
             break
     booster.best_score = collections.defaultdict(collections.OrderedDict)
+    booster.init_score = train_set.get_init_score()
     for dataset_name, eval_name, score, _ in evaluation_result_list:
         booster.best_score[dataset_name][eval_name] = score
     if not keep_training_booster:


jameslamb (commented):
Thanks @JoshuaC3 ! I've read the attached issue and the other stuff it links to so I think I understand what this is trying to accomplish, but I'm concerned.

`init_score` can be an array with shape `(n_observations, )` or even `(n_observations, n_classes)` for multiclass classification ([docs](https://lightgbm.readthedocs.io/en/latest/pythonapi/lightgbm.Dataset.html#lightgbm.Dataset)), so this could be a fairly large amount of data. This might substantially increase the memory footprint of the Booster object if you used `init_score` during training, which would have at least two negative side effects:

* increases the memory requirements for deploying a LightGBM model
* increases the amount of data that needs to be sent over the wire from Dask workers back to the client in distributed training with `lightgbm.dask` (https://github.com/microsoft/LightGBM/blob/d5c2c55682cdf52b9371dbead9bd551e2acbffdc/python-package/lightgbm/dask.py)

Is it absolutely necessary for #3905?

I'm just leaving a "Comment" review to bring this to the attention of other reviewers, for their consideration.

JoshuaC3:
Thanks for reviewing.

Yes, I was concerned about this as well. It certainly isn't essential and is somewhat of a workaround for the following: #4234 (which would be, or at least some variant to create an "intercept"). That said, I do think `init_score` is useful in its own right, but might have too much over an overhead in some cases.

Would a parameter such as `keep_init_score` be an option?

jameslamb:
> Would a parameter such as keep_init_score be an option?

I'm personally opposed to adding such a flag to the Booster object for this purpose. I think anyone choosing to set that flag could just as easily choose to separately save `init_score` (or the entire `Dataset` object). I worry that each parameter added to LightGBM's public API increases the complexity of the API, which can make it harder for users to learn and harder for maintainers to anticipate and test all combinations of parameters.

However, that's just one opinion! We should hear what @shiyu1994 and @StrikerRUS think.

shiyu1994:
Thanks @JoshuaC3 for your work. Keep the init_score available can be useful in some cases. But I have the following concerns,
1. As with @jameslamb, I don't think adding one more parameter `keep_init_score` would be a good idea, which increases the complexity of the pakcage.
2. We need to make it clear that `init_score` provided by users through 'Dataset' and the `"init_score"` caused by `boost_from_average=True` are two different concepts which does not interfere with each in the core of LightGBM. The former one can be get and set through Python API. But for the later one, currently it is not stored in anywhere of the `Booster` or `Dataset` object, except in first tree model. The only place that this `boost_from_average` score is saved is in the first tree (or first `K` trees if `K` classification). In the first tree, value for `boost_from_average` is simply added to all leaves, so it is mixed with the original leaf prediction values and cannot be recover. So even with the booster model saving the `init_score` from `Dataset`, as is done in this PR, still the `boost_from_average` value cannot be get. Without setting `init_score` in `Dataset` constructor, `get_init_score` simply returns `None`, regardless of `boost_from_average`.
3. For interpretability mentioned in #3905, I think what we want is the `boost_from_average` value. We can have an API to provide access to this value. That requires recording this value from the `gbdt.cpp` code and returns through new C and Python API functions.
https://github.com/microsoft/LightGBM/blob/d5c2c55682cdf52b9371dbead9bd551e2acbffdc/src/boosting/gbdt.cpp#L371-L376
So to make us closer to the interpretability goal in #3905, I think we should focus on the `boost_from_average` score instead of the user specified `init_score` for dataset.
I'm not sure I have fully understand your plan @JoshuaC3 for #3905. If I made any mistake, please feel free to respond.

JoshuaC3:
@shiyu1994 I think you do understand my plan, and maybe a little better than I did. Your comment has made it clear now!

You point 3. is spot on and highlights one of the reasons why I tried to access `init_score` instead of `boost_from_average`:
`init_score` is removed and is not used/forced into the leaves of the first tree unlike `boost_from_average`. This behaviour highly problematic when working out the feature importance in the [EBM style](https://github.com/interpretml/interpret/issues/12#issuecomment-493635372) as it biases the feature with the first split massively. Fixing this would involved something like having the average removed from the first trees leaves, saved on the booster model and added to the scores at prediction.

The other reason was, often don't wish to boost from "average", but wish to boost from some other value instead: min, max, median, 20th quantile, etc. This is because EBMs mean-centre the feature-scores in favour of the intercept being the mean of the target, thus it is rarely is the mean. Potentially superseding `boost_from_mean` with a `boost_from_value='mean'` parameter would be an nice end goal. We could then have `boost_from_value='median'`, `boost_from_value='min'` or `boost_from_value=53.4`, for example. I feel this could be appealing in other applications, possibly for optimising median centred metrics (MAE). That said, it is not essential in comparison to the previous point.

Is it worth copying this to an issue/feature request for further discussion and closing this PR?

shiyu1994:
@JoshuaC3 OK. So the final goal is to provide the EBM style interpretability.

> Fixing this would involved something like having the average removed from the first trees leaves, saved on the booster model and added to the scores at prediction.

That would require modifying the prediction logic. When prediction starts from the first tree, we should add back the average manually.

A simpler work around would be to make the average score available to the Python API for some cases the user want to know its value, but without removing the average from the first tree. And the EBM style interpretability can be calculated internally by subtracting in C++ or Python part. With the average being stored, it won't be difficult to calculate the EBM style interpretability. In other words, we subtract the average from the model only when calculating the EBM style interpretability, but leave the prediction logic and model file unchanged.

Do you think which strategy better suits the goal?

JoshuaC3:
My gut feeling is that the first proposal - modifying the prediction logic - seems more robust and more easily used by other languages, not just Python.

That said, option two means fewer changes to the prediction logic. I have created some examples in this [notebook](https://gist.github.com/JoshuaC3/fd129b2ba7d1244f9f7a365b709575bf#file-elgb-regression-example-ipynb) on my gist. The EBM-FI calculation and Plots is current written in Python and is a little slow - I am sure it could be sped up a lot though! However, it doesn't seem to make much difference if we subtract the mean before training, at the start of training or after training, so this option seems OK.

I think compared to this `init_score` PR it is a MUCH better solution!

Should I close this PR and open an issue detailing this?

shiyu1994:
@JoshuaC3 Sure! That would be great.

JoshuaC3:
Close in favour of

PR #4242, opened 2021-04-30
[R-package] Convert LGBM_GetLastError_R to use R built-in types

jameslamb:
This is a small step towards #3016. This PR proposes converting `LGBM_GetLastError_R` (and the R function that calls it, `lgb.last_error()`) to use R's built-in types instead of the custom type defined in https://github.com/microsoft/LightGBM/blob/91f72e2a8c8b1e5433d30d3891923b3a067402a6/R-package/src/R_object_helper.h.

### Notes for Reviewers

~This PR is not dependent on #4155.~ #4155 has been merged to `master` and this PR has been updated with it.

You can see the following to learn about the `SEXP`type.

* usage:
    - https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Handling-R-objects-in-C
    - https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Garbage-Collection
    - https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Allocating-storage
    - https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Handling-character-data
* low-level details:
    - https://cran.r-project.org/doc/manuals/r-release/R-ints.html#SEXPs

### DIFF
--- a/R-package/src/lightgbm_R.cpp
+++ b/R-package/src/lightgbm_R.cpp
@@ -56,8 +56,12 @@ LGBM_SE EncodeChar(LGBM_SE dest, const char* src, LGBM_SE buf_len, LGBM_SE actua
   return dest;
 }
 
-LGBM_SE LGBM_GetLastError_R(LGBM_SE buf_len, LGBM_SE actual_len, LGBM_SE err_msg) {
-  return EncodeChar(err_msg, LGBM_GetLastError(), buf_len, actual_len, std::strlen(LGBM_GetLastError()) + 1);
+SEXP LGBM_GetLastError_R() {
+  SEXP out;
+  out = PROTECT(Rf_allocVector(STRSXP, 1));
+  SET_STRING_ELT(out, 0, Rf_mkChar(LGBM_GetLastError()));
+  UNPROTECT(1);
+  return out;
 }
 
 LGBM_SE LGBM_DatasetCreateFromFile_R(LGBM_SE filename,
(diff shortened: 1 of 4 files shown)


jameslamb:
Ok CI seems to be happy, so I've converted this to "ready for review".

R-package/src/lightgbm_R.cpp
```diff
@@ -56,8 +56,12 @@ LGBM_SE EncodeChar(LGBM_SE dest, const char* src, LGBM_SE buf_len, LGBM_SE actua
   return dest;
 }
 
-LGBM_SE LGBM_GetLastError_R(LGBM_SE buf_len, LGBM_SE actual_len, LGBM_SE err_msg) {
-  return EncodeChar(err_msg, LGBM_GetLastError(), buf_len, actual_len, std::strlen(LGBM_GetLastError()) + 1);
+SEXP LGBM_GetLastError_R() {
+  SEXP out;
+  out = PROTECT(Rf_allocVector(STRSXP, 1));
+  SET_STRING_ELT(out, 0, Rf_mkChar(LGBM_GetLastError()));
```
StrikerRUS, line 62: Just a question: what is the difference between `Rf_allocVector`/`allocVector` and `Rf_mkChar`/`mkChar`?
I could find only the following link: https://www.cs.kent.ac.uk/projects/cxxr/doc/html/Rinternals_8h.html. And it looks like they are just aliases...

StrikerRUS (approved):
Very nice simplification!

R-package/src/lightgbm_R.cpp
jameslamb: yep, just aliases! This `#define` says "use the symbols with the `Rf_` prefix" https://github.com/microsoft/LightGBM/blob/1a367c65d182c0cc8d47e29e8f3d7a478dca3695/R-package/src/lightgbm_R.cpp#L15

I like that because it makes it a little more obvious which symbols are coming from R's headers.

Issue #3018, opened 2020-04-23
LGBMRanker query group setting when using gridsearchcv


chouisgiser:
https://github.com/microsoft/LightGBM/issues/1137

## Environment info

Operating System: Mac OS 10.15

CPU/GPU model: no

C++/Python/R version: python 3.7

LightGBM version or commit hash:

## Error message

lightgbm.basic.LightGBMError: Sum of query counts is not same with #data

## Reproducible examples

estimator_params = {'boosting_type': 'gbdt',
                    'objective': 'lambdarank',
                    'min_child_samples': 5,
                    'importance_type': 'gain',
                    }

gbm = lgb.LGBMRanker(**estimator_params)

params_grid = {'n_estimators': [10, 20],
               'num_leaves': [10, 20],
               'max_depth': [10],
               'learning_rate': [0.1],
               }

cv_group_info = query_train.astype(int)
flatted_group = np.repeat(range(len(cv_group_info)), repeats=cv_group_info)

logo = LeaveOneGroupOut()
cv = logo.split(X_train, y_train, groups=flatted_group)
cv_group = logo.split(X_train, groups=flatted_group)

grid = GridSearchCV(gbm, params_grid, cv=cv, verbose=2,
                          scoring=make_scorer(ndcg_score, greater_is_better=True), refit=False)

def group_gen(flatted_group, cv):
    for train, test in cv:
        yield np.unique(flatted_group[train], return_counts=True)[1]

gen = group_gen(flatted_group, cv_group)
params_fit = {
    'eval_set': [(X_test, y_test)],
    'eval_group': [query_test],
    'eval_metric': 'ndcg',
    'early_stopping_rounds': 100,
    'eval_at': [1, 2, 3],
}

grid.fit(X_train, y_train, group= next(gen), **params_fit)

## Steps to reproduce

1. Data is from the sample data in lambdarank directory
2. Use the LeaveOneGroup out
3. When I used my own data in which the query count of each group is the same, the code works. But if I use the data with different query counts in the query file, it reports the error.

guolinke:
ping @StrikerRUS for the https://github.com/microsoft/LightGBM/issues/1137#issuecomment-410216479

StrikerRUS:
Ah, it's a pity that workaround doesn't work fine anymore.

Maybe `cv` and `cv_group` generators produce different indices for some reason?..

[...]

PR #4243, opened 2021-04-30
[docs][python][scikit-learn] added note for LGBMRanker

StrikerRUS:
Address https://github.com/microsoft/LightGBM/issues/3018#issuecomment-828867332.

image


jameslamb (approved):
I think it's a very hepful addition, thank you.

PR #4244, opened 2021-05-01
[R-package] remove pre-allocated call_state in C++ calls

jameslamb:
This is another step towards #3016.

As of #4155 and #4163, the R package use `.Call()` to call routines from the C++ library. Currently, an integer called `call_state` is initialized on the R side before each such call, its value is manipulated during the call, and its value is returned by each call. Initializing this integer and passing it around is unnecessary, because nothing is done with it on the R side after the call completes. All uses of `.Call()` in the R package (except in calls to `LGBM_GetLastError_R`) modify their inputs by reference and do not return anything.

This PR proposes simplifying the R package and removing unnecessary memory usage and computation by removing the `call_state` concept. Instead, the functions defined in https://github.com/microsoft/LightGBM/blob/b27dcfa48f76e77c9e5033dcdefc1f3a76d99c8b/R-package/src/lightgbm_R.h are converted to return a `SEXP` (R's built-in type for passing data between R and C). This is initialized on the C++ side.

### Notes for Reviewers

If you want to learn more about the R `SEXP` object, see https://cran.r-project.org/doc/manuals/r-release/R-ints.html#SEXPs.

### DIFF
--- a/R-package/R/lgb.Booster.R
+++ b/R-package/R/lgb.Booster.R
@@ -16,11 +16,9 @@ Booster <- R6::R6Class(
       if (!lgb.is.null.handle(x = private$handle)) {
 
         # Freeing up handle
-        call_state <- 0L
         .Call(
           LGBM_BoosterFree_R
           , private$handle
-          , call_state
         )
         private$handle <- NULL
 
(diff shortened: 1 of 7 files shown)


StrikerRUS (approved):
Great!

PR #4245, opened 2021-05-01
[docs][python] update some docs related to custom objective

### LINKED ISSUE #3149: [Improvement] [Python] Improve docs relating to custom objective functions

StrikerRUS:
Supersedes #3502.

Closed #3149.

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -2600,14 +2600,17 @@ class Booster:
 
                 preds : list or numpy 1-D array
                     The predicted values.
+                    Predicted values are returned before any transformation,
+                    e.g. they are raw margin instead of probability of positive class for binary task.
                 train_data : Dataset
                     The training dataset.
                 grad : list or numpy 1-D array
-                    The value of the first order derivative (gradient) for each sample point.
+                    The value of the first order derivative (gradient) of the loss
+                    with respect to the elements of preds for each sample point.
                 hess : list or numpy 1-D array
-                    The value of the second order derivative (Hessian) for each sample point.
+                    The value of the second order derivative (Hessian) of the loss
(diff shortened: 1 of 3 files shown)


jameslamb (approved):
Nice! This improved my understanding of using custom objective functions.

PR #4246, opened 2021-05-01
[R-package] use R standard routine to access read-only ints passed to C++

jameslamb:
This is another step towards #3016. It proposed using R's standard interface to handle integers passed into the C++ side (for things like array size, buffer size, number of iterations, etc.).

Changes in this PR:

* use R builtin `SEXP` type for any arguments passed from R to C++ that should be read-only integers
* use R builtin `Rf_asInteger()` to convert those SEXP objects to ints (replacing LightGBM's `R_AS_INT`)
* fix minor inconsistencies between `lightgbm_R.h` and `lightgbm_R.cpp` (e.g. parameter named `nrow` in the header file and `num_row` in the implementation)

### Background

There is not a scalar integer type in R. When you run some code like `x <- 1L`, R creates a length-one **array** (referred to in R as a "vector").

To treat such data as an integer, instead of an integer array, in C/C++, it's necessary to extract the first element of that array into a new `int`. R provides a convenience function, `Rf_asInteger()`, for exactly that purpose.

See https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Some-convenience-functions for more details.

### Notes for Reviewers

This does not rely on #4242

### DIFF
--- a/R-package/src/lightgbm_R.cpp
+++ b/R-package/src/lightgbm_R.cpp
@@ -48,7 +48,7 @@ LGBM_SE EncodeChar(LGBM_SE dest, const char* src, LGBM_SE buf_len, LGBM_SE actua
     Log::Fatal("Don't support large string in R-package");
   }
   R_INT_PTR(actual_len)[0] = static_cast<int>(str_len);
-  if (R_AS_INT(buf_len) < static_cast<int>(str_len)) {
+  if (Rf_asInteger(buf_len) < static_cast<int>(str_len)) {
     return dest;
   }
   auto ptr = R_CHAR_PTR(dest);
@@ -76,9 +76,9 @@ LGBM_SE LGBM_DatasetCreateFromFile_R(LGBM_SE filename,
 LGBM_SE LGBM_DatasetCreateFromCSC_R(LGBM_SE indptr,
   LGBM_SE indices,
   LGBM_SE data,
-  LGBM_SE num_indptr,
-  LGBM_SE nelem,
-  LGBM_SE num_row,
+  SEXP num_indptr,
+  SEXP nelem,
+  SEXP num_row,
   LGBM_SE parameters,
   LGBM_SE reference,
   LGBM_SE out,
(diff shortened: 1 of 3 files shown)


R-package/src/lightgbm_R.cpp
```diff
@@ -566,78 +566,78 @@ SEXP LGBM_BoosterPredictForCSC_R(LGBM_SE handle,
...
+  SEXP feature_importance_type,
+  SEXP buffer_len,
   LGBM_SE actual_len,
   LGBM_SE out_str) {
   R_API_BEGIN();
   int64_t out_len = 0;
-  std::vector<char> inner_char_buf(R_AS_INT(buffer_len));
-  CHECK_CALL(LGBM_BoosterSaveModelToString(R_GET_PTR(handle), 0, R_AS_INT(num_iteration), R_AS_INT(feature_importance_type), R_AS_INT(buffer_len), &out_len, inner_char_buf.data()));
+  std::vector<char> inner_char_buf(Rf_asInteger(buffer_len));
+  CHECK_CALL(LGBM_BoosterSaveModelToString(R_GET_PTR(handle), 0, Rf_asInteger(num_iteration), Rf_asInteger(feature_importance_type), Rf_asInteger(buffer_len), &out_len, inner_char_buf.data()));
```
StrikerRUS, line 626:
```suggestion
  int64_t buf_len = static_cast<int64_t>(Rf_asInteger(buffer_len));
  std::vector<char> inner_char_buf(buf_len);
  CHECK_CALL(LGBM_BoosterSaveModelToString(R_GET_PTR(handle), 0, Rf_asInteger(num_iteration), Rf_asInteger(feature_importance_type), buf_len, &out_len, inner_char_buf.data()));
```

R-package/src/lightgbm_R.cpp
```diff
@@ -566,78 +566,78 @@ SEXP LGBM_BoosterPredictForCSC_R(LGBM_SE handle,
...
+  SEXP feature_importance_type,
+  SEXP buffer_len,
   LGBM_SE actual_len,
   LGBM_SE out_str) {
   R_API_BEGIN();
   int64_t out_len = 0;
-  std::vector<char> inner_char_buf(R_AS_INT(buffer_len));
-  CHECK_CALL(LGBM_BoosterDumpModel(R_GET_PTR(handle), 0, R_AS_INT(num_iteration), R_AS_INT(feature_importance_type), R_AS_INT(buffer_len), &out_len, inner_char_buf.data()));
+  std::vector<char> inner_char_buf(Rf_asInteger(buffer_len));
+  CHECK_CALL(LGBM_BoosterDumpModel(R_GET_PTR(handle), 0, Rf_asInteger(num_iteration), Rf_asInteger(feature_importance_type), Rf_asInteger(buffer_len), &out_len, inner_char_buf.data()));
```
StrikerRUS, line 640:
```suggestion
  int64_t buf_len = static_cast<int64_t>(Rf_asInteger(buffer_len));
  std::vector<char> inner_char_buf(buf_len);
  CHECK_CALL(LGBM_BoosterDumpModel(R_GET_PTR(handle), 0, Rf_asInteger(num_iteration), Rf_asInteger(feature_importance_type), buf_len, &out_len, inner_char_buf.data()));
```

StrikerRUS (approved):
I think we can cast `buffer_len` only once in two places below.

jameslamb:
> I think we can cast `buffer_len` only once in two places below.

nice! valgrind seems happy with that change (https://github.com/microsoft/LightGBM/pull/4246#issuecomment-831312528) and so does the rest of CI, so I'll merge this.

PR #4247, opened 2021-05-02
[R-package] Use R standard routines to access numeric and integer array data in C++

### LINKED ISSUE #4216: (R) Functions that replace R headers are not compatible with all R versions

jameslamb:
Another step towards #3016.

Fixes #4216.

Changes in this PR:

* removes LightGBM-custom pointers to R data (`R_REAL_PTR()`, `R_INT_PTR()`) with the standard equivalents from `Rinternals.h` (`REAL()`, `INTEGER()`)
* converts arguments on the C++ side that are accessed that way from `LGBM_SE` to `SEXP`

### Description

R is a dynamically-typed language. You can run code like `x <- c(1, 2, 3)` without declaring that `x` is a numeric array, and R will just figure it out.

R makes this possible by storing data for an object in a structure called a `SEXPREC`. That structure will contain different data based on the R class (integer, character, function, etc.). Libraries can reference that data in C code using a type, `SEXP`, provided by `Rinternals.h`. Libraries can also get a pointer to a particular type of data within a `SEXPREC` using other functions provided by `Rinternals.h`, for example `REAL()` to get a pointer to its numeric data.

As described in #4216, the internal details of the `SEXPREC` struct can change between different versions of R. `SEXP` is the official R API into that struct, and by using it libraries can reliably stay compatible with multiple R versions.

This PR's changes fix a bug that currently exists when using `{lightgbm}` with R 3.6, and protects `{lightgbm}` from similar bugs in the future.

For more details on this, see https://cran.r-project.org/doc/manuals/r-release/R-ints.html#SEXPs.

### Notes for Reviewers

This PR does not depend on #4242 or #4247

### DIFF
--- a/R-package/tests/testthat/test_basic.R
+++ b/R-package/tests/testthat/test_basic.R
@@ -1247,6 +1247,38 @@ test_that("lgb.train() supports non-ASCII feature names", {
   }
 })
 
+test_that("lgb.train() works with integer, double, and numeric data", {
+  data(mtcars)
+  X <- as.matrix(mtcars[, -1L])
+  y <- mtcars[, 1L, drop = TRUE]
+  expected_mae <- 4.263667
+  for (data_mode in c("numeric", "double", "integer")) {
+    mode(X) <- data_mode
+    nrounds <- 10L
+    bst <- lightgbm(
+      data = X
+      , label = y
+      , params = list(
+        objective = "regression"
+        , min_data = 1L
+        , learning_rate = 0.01
+        , seed = 708L
+      )
+      , nrounds = nrounds
+    )
+
+    # should have trained for 10 iterations and found splits
+    modelDT <- lgb.model.dt.tree(bst)
+    expect_equal(modelDT[, max(tree_index)], nrounds - 1L)
+    expect_gt(nrow(modelDT), nrounds * 3L)
+
+    # should have achieved expected performance
+    preds <- predict(bst, X)
+    mae <- mean(abs(y - preds))
+    expect_true(abs(mae - expected_mae) < TOLERANCE)
+  }
+})
+
 test_that("when early stopping is not activated, best_iter and best_score come from valids and not training data", {
   set.seed(708L)
   trainDF <- data.frame(
--- a/R-package/src/R_object_helper.h
+++ b/R-package/src/R_object_helper.h
@@ -102,8 +102,6 @@ typedef union { VECTOR_SER s; double align; } SEXPREC_ALIGN;
 
 #define R_CHAR_PTR(x)  (reinterpret_cast<char*>DATAPTR(x))
 
-#define R_INT_PTR(x)   (reinterpret_cast<int*> DATAPTR(x))
-
 #define R_AS_INT(x)    (*(reinterpret_cast<int*> DATAPTR(x)))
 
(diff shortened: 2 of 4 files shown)


R-package/tests/testthat/test_basic.R
```diff
@@ -1247,6 +1247,38 @@ test_that("lgb.train() supports non-ASCII feature names", {
...
+    expect_equal(modelDT[, max(tree_index)], nrounds - 1L)
+    expect_gt(nrow(modelDT), nrounds * 3L)
+
+    # should have achieved expected performance
+    preds <- predict(bst, X)
+    mae <- mean(abs(y - preds))
+    expect_true(abs(mae - expected_mae) < TOLERANCE)
+  }
+})
+
```
jameslamb, line 1281: Thanks to @david-cortes for providing the write-up in https://github.com/microsoft/LightGBM/issues/4216, which included the reproducible example that inspired this unit test.

I tested this unit test tonight on my Mac, with R 3.6 and `{lightgbm}` 3.2.1 installed from CRAN, and confirmed that it failed in exactly the way described in #4216. So I'm confident that this test passing means the changes in this PR fix #4216.

StrikerRUS (approved):
Thank you very much for the iterative improvement!

PR #4251, opened 2021-05-03
[ci] Delete lock.yml

StrikerRUS:
lockbot has been not working for months. I don't know how to fix it.

Maybe we should update bot access settings, but I'm not allowed to do this:

image


jameslamb:
Oh wow, thank you for noticing this! I forgot we had lock bot in addition to the no-response bot.

I just tried to update it at https://github.com/apps/lock/installations/206869 and got the exact same response you did.

> This action must be performed by an organization owner

I looked at the source of the bot, https://github.com/dessant/lock-threads-app, which has had the following in its README since February 2020.

> This project is in maintenance mode, migrating to the version built on GitHub Actions is recommended: https://github.com/dessant/lock-threads

Are you interested in trying to switch to that version at https://github.com/dessant/lock-threads? It looks like it can all be done from GitHub Actions without needing repo-owner access. If you'd rather just remove this bot, that's ok with me too. I have no strong preference.

jameslamb (approved):

PR #4252, opened 2021-05-04
[R-package] Use R standard routines to access character data in C++

jameslamb:
Another step towards #3016.

Similar to the changes for numeric data in #4247, this proposes replacing moost uses of the LightGBM-custom `R_CHAR_PTR` with R's standard interface for accessing character data. Unlike that PR, this requires two steps...extracting a `CHARSXP` from the object passed in from R, then getting a pointer to its data (`char *`). That results in a pattern like this:

```cpp
CHAR(Rf_asChar(filename))
```

As a result of these changes, the utility function `lgb.c_str()` is no longer necessary.

### Notes for Reviewers

1. #4034 is not fixed by this change, but it is partially fixed. As of this PR, the reproducible examples in that issue yield `[LightGBM] [Fatal] Unknown format of training data.` instead of `[LightGBM] [Fatal] Data file ��?��V doesn't exist.`. That can be addressed in a followup PR.

2. `R_CHAR_PTR` cannot be completely removed in this PR, because it is still used in `EncodeChar`. That can be addressed in a future PR.

https://github.com/microsoft/LightGBM/blob/bb88d92ef27acc0731bf3b348920560410aa2e78/R-package/src/lightgbm_R.cpp#L46-L57

### References

* https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Handling-character-data
* https://github.com/dmlc/xgboost/blob/b35dd76dca6dfa9151394da847567ce99fa76085/R-package/src/xgboost_R.cc#L72

### DIFF
--- a/R-package/R/lgb.Booster.R
+++ b/R-package/R/lgb.Booster.R
@@ -452,7 +452,7 @@ Booster <- R6::R6Class(
         , private$handle
         , as.integer(num_iteration)
         , as.integer(feature_importance_type)
-        , lgb.c_str(x = filename)
+        , filename
       )
 
       return(invisible(self))
--- a/R-package/R/lgb.Booster.R
+++ b/R-package/R/lgb.Booster.R
@@ -90,7 +90,7 @@ Booster <- R6::R6Class(
           # Create booster from model
           .Call(
             LGBM_BoosterCreateFromModelfile_R
-            , lgb.c_str(x = modelfile)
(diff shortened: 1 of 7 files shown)


StrikerRUS (approved):

jameslamb:
Getting closer! Hopefully just a few more PRs and we can finally be done with #3016.

PR #4250, opened 2021-05-03
Correct spelling

az0:
Most changes were in comments, and there were a few changes to literals for log output.

There were no changes to variable names, function names, IDs, or functionality.

PR #4253, opened 2021-05-04
fix param name

StrikerRUS:
Correction of #4250.

`hessians` here is a param name (`const score_t* hessians`) and should be written from a small letter.


jameslamb (approved):
ah! thank you, sorry for the oversight

jameslamb:
Ugh sorry, looking back through #4250 it looks like I made the same mistake in one other place

https://github.com/microsoft/LightGBM/blob/ec87432591e8c7c22528ac896818e466207aa681/src/boosting/gbdt.h#L144

could you please fix that too as part of this PR?

StrikerRUS:
Thanks for spotting another similar place!

Issue #4206, opened 2021-04-20
lightgbm.cv strange results in Random Forest mode


candalfigomoro:
This is basically a continuation of this closed issue: https://github.com/microsoft/LightGBM/issues/3387

With lightgbm's cv method, if I use 300 trees in Random Forest mode, I get 300 scores.

In comparison, when I use xgboost's cv method in Random Forest mode, I get just 1 score (this naturally arises from the fact that in xgboost we have decoupled parameters for boosted/parallel trees, see https://github.com/microsoft/LightGBM/issues/3128).

What is the intepretation of the 300 scores? Is the 1st score the one I'd get with 1 tree and the 300th score the one I get with a Random Forest of 300 trees?

While for boosting iterations this kind of output makes sense, I think the interpretation of lightgbm's cv results in Random Forest mode is a bit confusing.

shiyu1994:
@candalfigomoro Thanks for using LightGBM.

> What is the intepretation of the 300 scores? Is the 1st score the one I'd get with 1 tree and the 300th score the one I get with a Random Forest of 300 trees?

Yes, that's correct. The `i`th value in the list represents the averaged `mean` or `std` across the folds using the first `i` iterations.
This is because `rf` is treated as a special type of `boosting` in LightGBM, which does not update the gradients between iterations.

candalfigomoro:
@shiyu1994 Thank you!

PR #4166, opened 2021-04-06
[ci] run cpp tests at CI

StrikerRUS:
Linking #3555 and #261.
Tested with both `gcc` and `(Apple)Clang`.

PR #4254, opened 2021-05-04
[tests][cli] distributed training

### LINKED ISSUE #3841: Write tests for parallel code

jmoralez:
This ~aims to close~ contributes to #3841 by proposing a way to run tests on distributed training calling the CLI from python. 

Here I define a class `DistributedMockup` (suggestions for a better name are more than welcome) that implements a `fit` method that creates the required configuration files, saves the data and then runs the lightgbm binary using a `ThreadPoolExecutor`. The training logs are taken from the first machine in `mlist.txt` and are piped to stdout.

There's also a `predict` method that computes the predictions using the trained model on a single call to the binary (don't think it's necessary here to make a call on each partition).

I tested this on my fork using github actions, however this will run on azure pipelines so I'll be working on that following the very detailed explanation here: https://github.com/microsoft/LightGBM/issues/3841#issuecomment-812959028


jmoralez:
Hi @jameslamb. Will this: https://github.com/microsoft/LightGBM/blob/fcd2453514a6ecd780121e4ea31a227559145451/.ci/test.sh#L202
trigger the tests?
In that case I think I have to fix the paths, everything right now is designed to work by calling `pytest` from `tests/distributed/`

jmoralez:
This ran successfully ([logs](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=9986&view=logs&j=cd496dfc-df55-5b41-9968-b20563c04f1a&t=04af574d-e756-55c4-f10f-8a87043a256d&l=457)) so I think it's ready for review. Right now it's only running two simple tests but could be extended to perform the tests that are run for the dask interface, we could probably borrow most of the data creation and the tests themselves.

jameslamb (changes requested):
This is great, thank you for the effort to make this extensible and to fit it into this project's existing continuous integration setup!

<hr>

Can you please add rules for the files that would be created by these tests to the top-level [`.gitignore`](https://github.com/microsoft/LightGBM/blob/c1d2dbe2c546e3d0c6f9c66ef5dcef6428541df2/.gitignore)?

I pulled this branch and ran the tests tonight (worked perfectly 🎉 ) , and several files were untracked.

```text
On branch tests/distributed
Your branch is up to date with 'origin/tests/distributed'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)

	mlist.txt
	predictions.txt
	train.txt
	train0.conf
	train0.txt
	train1.conf
	train1.txt

nothing added to commit but untracked files present (use "git add" to track)
```

<hr>

I see that right now, the logic for `_write_data()` is written like "evenly split the training data between the workers". In my opinion, I think it would be better to have `DistributedMockup.fit()` expect `data` to be a list of data chunks. In the few months since `lightgbm.dask` was first introduced, we've seen a lot of issues related to uneven distribution of the training data across workers (for example #3817, #3882, https://github.com/microsoft/LightGBM/pull/4101#issuecomment-824972826, #4206).

If `fit()` took in `List[np.ndarray]`, then it would be easy to write tests on the behavior of distributed training in situations where, for example, feature distributions are very different on different partitions, or where the dataset sizes are very different.

What do you think?

<hr>

Can you please add docstrings and type hints (including return type hints) on the methods of `DistributedMockup`, `create_data()`, and `run_and_log()`? I think that will help others to be able to contribute changes to these tests in the future, and will give us a chance of finding subtle bugs with `mypy`.

As of #3756, we've been trying to encourage the use of type hints in this project's Python code.

### Notes for Other Reviewers

I support merging this PR with only a single job on `Linux_latest` (this is what I asked for in https://github.com/microsoft/LightGBM/issues/3841#issuecomment-812959028) and with only minimal tests like those in `test_classifier` and `test_regressor`. I have a preference for setting up the structure here and then incrementally adding tests later, like @jmoralez has proposed.

jmoralez:
Hi @jameslamb, I've included your comments, I think sending the data already partitioned to the `fit` method makes a lot of sense. 

I noticed that `mypy` is only ran for the `python-package` folder in the `lint` task so I included it in the `cli-distributed` task to check the `_test_distributed.py` file.

[This](https://ci.appveyor.com/project/guolinke/lightgbm/builds/39100892/job/wh9p7x32rem3wlks) job failed in the CI but seems to be testing on windows, so I don't think is related to this PR. Looking forward to your thoughts on these changes.

tests/distributed/_test_distributed.py
```diff
@@ -0,0 +1,156 @@
...
+import subprocess
+import sys
+from concurrent.futures import ThreadPoolExecutor
+from typing import Dict, List
+
+import numpy as np
+from sklearn.datasets import make_blobs, make_regression
+from sklearn.metrics import accuracy_score
+
+import lightgbm as lgb
```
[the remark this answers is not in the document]
jmoralez: The compilation step takes the most time and once that's done using the `--precompile` flag makes the installation very fast. I thought we may use the python package later for loading the booster and checking some stuff like what's currently done for the categoricals here: https://github.com/microsoft/LightGBM/blob/0a172d9e3b292110d90878ba7f0932623b80cbe2/tests/python_package_test/test_dask.py#L315-L323 for example.

One thing that does take a bit more time is this: https://github.com/microsoft/LightGBM/blob/0a172d9e3b292110d90878ba7f0932623b80cbe2/.ci/test.sh#L107-L108 I placed these tests below that because they need numpy and scikit-learn but could maybe move this a bit up to just install those and avoid all the other stuff that gets installed there. What do you think?

tests/distributed/_test_distributed.py
```diff
@@ -0,0 +1,156 @@
...
+        y_pred = np.loadtxt('predictions.txt')
+        return y_pred
+
+    def write_train_config(self, i: int) -> None:
+        """Create a file train{i}.txt with the required configuration to train.
+
+        Each worker gets a different port and piece of the data, the rest are the
+        model parameters contained in `self.config`.
+        """
+        with open(f'train{i}.conf', 'wt') as f:
```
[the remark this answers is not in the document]
jmoralez: I think they both need the same file and the data gets partitioned internally, I see this using the current approach:
```
[LightGBM] [Info] Number of data points in the train set: 248, number of used features: 4
[LightGBM] [Info] Number of data points in the train set: 252, number of used features: 4
```
Whereas using `lightgbm.dask`
```
[LightGBM] [Info] Number of data points in the train set: 500, number of used features: 4
[LightGBM] [Info] Number of data points in the train set: 500, number of used features: 4
```

Not sure if there's a way to tell each machine to use all of its data, I think right now it's splitting the data of the main machine.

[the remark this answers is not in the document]
jmoralez: Ah: https://lightgbm.readthedocs.io/en/latest/Parameters.html#pre_partition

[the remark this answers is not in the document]
jameslamb: haha ah I JUST found that too. Yeah, I think that is necessary for the CLI. I just tried adding `pre_partition: True` to the config and found that the returned model used all 1000 data points.

Ok, I think doing that should be enough. It would still probably be good to write two separate model files, to avoid any issues that might be caused by both processes trying to write to the same file.

tests/distributed/_test_distributed.py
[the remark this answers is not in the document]
jameslamb: oooo ok yeah, that makes sense. I wasn't thinking about how quick it is with `--precompile`.

I'd definitely support moving these tests above that line and just `conda install`-ing what you need inside the task's code in `test.sh`.

We do the same thing for other tasks like `check-docs` and `if-else`.

https://github.com/microsoft/LightGBM/blob/0a172d9e3b292110d90878ba7f0932623b80cbe2/.ci/test.sh#L31

https://github.com/microsoft/LightGBM/blob/0a172d9e3b292110d90878ba7f0932623b80cbe2/.ci/test.sh#L80

.ci/test.sh
```diff
@@ -107,6 +107,16 @@ fi
 conda install -q -y -n $CONDA_ENV cloudpickle dask distributed joblib matplotlib numpy pandas psutil pytest scikit-learn scipy
 pip install graphviz  # python-graphviz from Anaconda is not allowed to be installed with Python 3.9
 
+if [[ $TASK == "cli-distributed" ]]; then
+    pip install --user mypy
+    mypy --ignore-missing-imports tests/distributed/
```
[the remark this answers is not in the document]
jmoralez: Removed in 9f51ad8

tests/distributed/_test_distributed.py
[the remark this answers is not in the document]
jmoralez: Adressed in 9f51ad8

tests/distributed/_test_distributed.py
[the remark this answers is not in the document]
jmoralez: Addressed in 9f51ad8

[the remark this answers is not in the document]
jameslamb: excellent, thanks!

jameslamb (approved):
This looks great, thank you!

@shiyu1994 could you also please review? I think you're more familiar with the CLI than I am (and @StrikerRUS is currently reviewing a lot of other PRs 😀 )

.vsts-ci.yml
```diff
@@ -119,6 +119,8 @@ jobs:
         TASK: gpu
         METHOD: wheel
         PYTHON_VERSION: 3.7
+      distributed:
+        TASK: cli-distributed
```
StrikerRUS, line 123: Sorry for writing this so late!
Do you @jameslamb really think that parallel tests should be executed as a separate CI job? I believe that environment initialization overhead is quite big in this case.
From my point of view, we should add new CI jobs in cases when new job requires something special to be done we haven't done before as a part of our CI scripts.
For example,
- `dask` is tested within our ordinary (mostly Python-package) tests (we just added some dependencies in conda env), we don't have separate `dask` CI job
- `mpi` requires MPI installation and is mutually exclusive with ordinary installation, hence we have a separate `mpi` CI job
- `gpu` requires installation of `boost`, GPU drivers and is mutually exclusive with ordinary installation, hence we have a separate `gpu` CI job
- `regular`, `sdist`, `bdist` jobs are presented because they all test different ways of Python-package installation and its' correctness
- `swig` only compiles SWIG wrapper and doesn't run any tests (even for such scenario we used to compile SWIG artifacts within `sdist` job before got self-running agents to balance CI loading)
- `cpp_tests` requires special CMake flags and doesn't require Python ecosystem initialization, hence we have a separate `cpp-tests` CI job

Also, please note that C API tests are run with each Python-package-related CI job.
https://github.com/microsoft/LightGBM/blob/master/tests/c_api_test/test_.py

I believe that basic distributed tests which are not requiring MPI installation or other special things can be run with each Python-package-related CI job. Please share your opinion on this.

.ci/test.sh
```diff
@@ -85,6 +85,15 @@ if [[ $TASK == "if-else" ]]; then
     exit 0
 fi
 
+if [[ $TASK == "cli-distributed" ]]; then
+    conda install -q -y -n $CONDA_ENV numpy pytest scikit-learn
+    mkdir $BUILD_DIRECTORY/build && cd $BUILD_DIRECTORY/build && cmake .. && make -j4 || exit -1
```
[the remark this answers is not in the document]
jose-moralez: The executable is definitely needed. The shared library is needed for the python package which is installed afterwards using the `--precompile` flag. Right now the python package doesn't get used but *may* be used in the future, do you think we should wait until it's actually used to build the shared library and install the python package?

.vsts-ci.yml
jameslamb: > I believe that environment initialization overhead is quite big in this case.

Specifically, the most recent run of this job took 4m15s, about 1m40s of which was spent compiling `lightgbm` and running the tests (based on timestamps in logs)

```text
2021-05-12T03:28:03.4369397Z -- The C compiler identification is Clang 10.0.0
...
2021-05-12T03:29:43.5905794Z ============================== 2 passed in 1.04s ===
```

reference: https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=10053&view=logs&j=cd496dfc-df55-5b41-9968-b20563c04f1a

> I believe that basic distributed tests which are not requiring MPI installation or other special things can be run with each Python-package-related CI job

They are "basic" today because I've asked @jmoralez to focus this first PR on only setting up the structure and not on adding a lot of additional tests. But I expect that a lot more tests will be added in the future. Every time we see an issue with `lightgbm.dask` or from `mmlspark` and wonder *"is this a problem at the level of those interfaces or something from LightGBM's core library?"*, I think we'll want to add tests here.

I think it's possible that a few months from now, these distributed training tests might take 3 or 4 minutes to run (just the test part, not all of the environment setup). Instead of adding 3 to 4 minutes to each Python CI job, I thought it made sense to add only a single job as a starting point.

If you don't find that convincing, I don't feel strongly enough about it to block the PR until we agree. If you still think these tests should be run on every Python job instead of as a standalone, please advise @jmoralez on specifically what changes should be made (for example).

.ci/test.sh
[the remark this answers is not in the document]
StrikerRUS: Yeah, I think so.

.vsts-ci.yml
StrikerRUS: > focus this first PR on only setting up the structure and not on adding a lot of additional tests.

I do support this.

> Every time we see an issue with `lightgbm.dask` or from `mmlspark` and wonder *"is this a problem at the level of those interfaces or something from LightGBM's core library?"*, I think we'll want to add tests here.

For sure, that was the reason I created #3841.


You are right that separate CI jobs have their own advantages, but I think that the main benefit of embedding distributed tests into CI jobs with main tests is that in this case you get good coverage and balance for different OSes, compilers and environments in general. You don't need to think about where to setup new job: on Azure, GitHub Action, Appveyor; what compiler to choose on one CI service and what compiler on another one; invent new `if` rules for `setup.py` or something else. It is already done and you get it for free.

> If you don't find that convincing, I don't feel strongly enough about it to block the PR until we agree.

I believe it's a bad idea to merge PRs that are "convenient" for only one of the maintainers.
We are not speaking about just theoretical things, we are discussing this particular PR and changes it proposes to merge into the `master` branch. I don't agree with that this is "delaying" PR somehow. This is what PRs are designed for - for discussing changes they contain.

With new info uncovered

> The executable is definitely needed.
   https://github.com/microsoft/LightGBM/pull/4254#discussion_r631955547

some previous statements may become irrelevant, for example. But this should be discussed before the merge to prevent code reverts in the future, I suppose.

jameslamb: > This is what PRs are designed for

I don't disagree. Sorry if it sounded like I was complaining, that was not my intention.

> You don't need to think about where to setup new job: on Azure, GitHub Action, Appveyor; what compiler to choose on one CI service and what compiler on another one; invent new if rules for setup.py or something else. It is already done and you get it for free.

In this case, @jmoralez WILL have to think about those things, all at once, right now. Since distributed training with the CLI isn't currently tested on Mac or Windows, for example, if failures are experienced adding those tests to jobs for those operating systems then they'll have to be fixed (or ignored with `if` or `pytest.skip`) before this can be merged. The same goes for other combinations, like different Python version.

So, said another way, I have a preference for taking smaller, incremental steps for adding this particular type of testing. I prefer adding this type of testing in a single job right now because then two types of work (changing which configurations the job runs on, adding new tests) can happen in parallel.

I think you are saying you'd prefer to add broader coverage in this first PR. I am ok with that and happy to support it.

StrikerRUS: > In this case, @jmoralez WILL have to think about those things, all at once, right now.

Oh no, definitely no! Only in case he'd love to do it.

As an example, please remember adding cpp tests with GTest. First, possibility of tests compilation was added in #3555 and was checked for correctness: https://github.com/microsoft/LightGBM/pull/3555#issuecomment-796365895. Then, we actually added CI jobs with cpp tests: #4166. And right now different tests are proposed to be added in pending PRs. I think this is a good example of
 > taking smaller, incremental steps for adding this particular type of testing.

As the core code of this PR is independent to the way we will execute it, I think we can go the same direction here. Leave only testing code in this PR and add CI jobs calling it in a follow-up one(s). WDYT?

shiyu1994:
> This looks great, thank you!
> 
> @shiyu1994 could you also please review? I think you're more familiar with the CLI than I am (and @StrikerRUS is currently reviewing a lot of other PRs 😀 )

Sure. Will do that soon.

.vsts-ci.yml
jameslamb: > Oh no, definitely no

Oh ok, I guess I misunderstood. I thought that your comment at the beginning of this thread was saying that you wanted this PR to add these tests to all existing Python jobs. I think now that maybe you just meant "for us to consider #3841 **fixed**, it should be added to all of those jobs".

> Leave only testing code in this PR and add CI jobs calling it in a follow-up one(s)

Sure! Let's do that, I agree. So then I think this PR would need the following changes:

* remove `fixes` from the PR title and `close` from the PR description
* remove the changes to `.vsts-ci.yml` and `.ci/test.sh`

@StrikerRUS do you agree?

tests/distributed/_test_distributed.py
```diff
@@ -0,0 +1,157 @@
...
+
+def test_classifier():
+    """Test the classification task."""
+    num_machines = 2
+    data = create_data(task='binary-classification')
+    partitions = np.array_split(data, num_machines)
+    params = {
+        'objective': 'binary',
+    }
+    clf = DistributedMockup(params)
```
shiyu1994, line 139: Shall we pass the `num_machines` to `DistributedMockup`, to guarantee the consistence between the number of machines in `array_split` and distributed training? Just a suggestion and for convenience for future modification.

tests/distributed/_test_distributed.py
```diff
@@ -0,0 +1,157 @@
...
+
+def test_regressor():
+    """Test the regression task."""
+    num_machines = 2
+    data = create_data(task='regression')
+    partitions = np.array_split(data, num_machines)
+    params = {
+        'objective': 'regression',
+    }
+    reg = DistributedMockup(params)
```
shiyu1994, line 154: Same with the `clf` case.

tests/distributed/_test_distributed.py
```diff
@@ -0,0 +1,157 @@
...
+    reg.fit(partitions)
+    y_pred = reg.predict()
+    np.testing.assert_allclose(y_pred, reg.label_, rtol=0.5, atol=50.)
```
shiyu1994, line 157: Won't `rtol=0.5` be too large to ensure sensible training results?

.vsts-ci.yml
StrikerRUS: > I think now that maybe you just meant "for us to consider #3841 **fixed**, it should be added to all of those jobs".

Yeah, #3841 should be "fixed" only after adding some meaningful number of tests for distributed code.

> do you agree?

Totally agree.

jameslamb: Great. @jmoralez when you have time, could you please remove the changes from `.vsts-ci.yml` and `.ci/test.sh`?

I've removed "fixes" from the PR title.

tests/distributed/_test_distributed.py
jmoralez: Those were the values used in the tests for dask. I tested a bit and using `rtol=0.2` still passes, below that the test starts to fail.
https://github.com/microsoft/LightGBM/blob/f076ca58b382fa2fb0a6d8d7367983083aed03a6/tests/python_package_test/test_dask.py#L518-L519

tests/distributed/_test_distributed.py
jmoralez: Added in a424374, now `num_machines` is passed in the configuration

shiyu1994 (approved):

tests/distributed/predict.conf
```diff
@@ -0,0 +1,4 @@
+task = predict
```
StrikerRUS, line 1: I think we can write this file on the fly in `__init__()` or `predict()` method like all other helper files.

StrikerRUS (commented):
Thanks for addressing fundamental comments! Please now check some minor comments below.

tests/distributed/_test_distributed.py
```diff
@@ -0,0 +1,159 @@
...
+        X, y = make_blobs(n_samples, centers=centers, random_state=42)
+    elif task == 'regression':
+        X, y = make_regression(n_samples, n_features=4, n_informative=2, random_state=42)
+    dataset = np.hstack([y.reshape(-1, 1), X])
+    return dataset
+
+
+def run_and_log(cmd: List[str]) -> None:
+    """Run `cmd` in another process and pipe its logs to this process' stdout."""
+    process = subprocess.Popen(cmd, stdout=subprocess.PIPE)
```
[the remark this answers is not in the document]
jmoralez: When I was implementing this I read [here](https://docs.python.org/3/library/subprocess.html#subprocess.run) that when using `subprocess.run` stdout and stderr weren't captured by default, so I added this. However I'm testing this now and it actually pipes them to the calling process, so I think capturing means like actually getting them as bytes. I think its best to just use `subprocess.run` as is. Do you agree?

tests/distributed/_test_distributed.py
```diff
@@ -0,0 +1,159 @@
...
+
+    def worker_train(self, i: int) -> None:
+        """Start the training process on the `i`-th worker.
+
+        If this is the first worker, its logs are piped to stdout.
+        """
+        cmd = f'./lightgbm config=train{i}.conf'.split()
+        if i == 0:
+            return run_and_log(cmd)
+        subprocess.run(cmd)
```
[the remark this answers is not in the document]
jmoralez: Following the idea in https://github.com/microsoft/LightGBM/pull/4254#discussion_r640233072, I believe we can just call `result = subprocess.run(cmd)` and `if result.returncode != 0: raise RuntimeError`. The stderr will be there because it gets piped. What do you think?

tests/distributed/predict.conf
jmoralez: added in f5e9d496715cf409f5bdf38e38092e72d0764a3d

tests/distributed/_test_distributed.py
```diff
@@ -0,0 +1,159 @@
...
+
+    def _set_ports(self) -> None:
+        """Randomly assign a port for training to each worker and save all ports to mlist.txt."""
+        self.listen_ports = [lgb.dask._find_random_open_port() for _ in range(self.n_workers)]
```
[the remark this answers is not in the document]
jmoralez: added in f5e9d496715cf409f5bdf38e38092e72d0764a3d

tests/distributed/_test_distributed.py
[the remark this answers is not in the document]
StrikerRUS: Agree. As we don't use output in any checks, we don't need to capture it.

tests/distributed/_test_distributed.py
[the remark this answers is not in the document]
StrikerRUS: Yes, I guess so.

tests/distributed/_test_distributed.py
```diff
@@ -0,0 +1,178 @@
...
+    def fit(self, partitions: List[np.ndarray], train_config: Dict = {}) -> None:
+        """Run the distributed training process on a single machine.
+
+        For each worker i:
+            1. The i-th partition is saved as train{i}.txt.
+            2. A random port is assigned for training.
+            3. A configuration file train{i}.conf is created.
+            4. The lightgbm binary is called with config=train{i}.conf in another thread.
+            5. The trained model is saved as model{i}.txt. Each model file only differs in data and local_listen_port.
+        The whole training set is saved as train.txt and the logs from the first worker are piped to stdout.
```
StrikerRUS, line 100: I think part about stdout is not actual any more after the latest refactoring.

tests/distributed/_test_distributed.py
```diff
@@ -0,0 +1,178 @@
...
+        for result in results:
+            if result.returncode != 0:
+                raise RuntimeError
+
+    def predict(self, predict_config: Dict = {}) -> np.ndarray:
+        """Compute the predictions using the model created in the fit step.
+
+        model0.txt is used to predict the training set train.txt using predict.conf.
+        The predictions are saved as predictions.txt and are then loaded to return them as a numpy array.
+        The logs are piped to stdout.
```
StrikerRUS, line 124: I think this line is not actual any more after the latest refactoring.

.gitignore
```diff
@@ -430,6 +430,11 @@ miktex*.zip
 **/lgb.Dataset.data
 **/model.txt
 **/lgb-model.txt
+tests/distributed/mlist.txt
+tests/distributed/train*
+tests/distributed/model*
+tests/distributed/predictions*.txt
```
StrikerRUS, line 436:
```suggestion
tests/distributed/predictions*.txt
tests/distributed/predict.conf
```

.gitignore
```diff
@@ -430,6 +430,11 @@ miktex*.zip
 **/lgb.Dataset.data
 **/model.txt
 **/lgb-model.txt
+tests/distributed/mlist.txt
```
StrikerRUS, line 433: We need to replace all `open(smth)` calls with
```
CURR_DIR = os.path.abspath(os.path.dirname(__file__))

...

open(os.path.join(CURR_DIR, smth))
```
to be sure that these lines in `.gitignore` are correct. Because now all files are created in the directory from which `pytest` is called.

StrikerRUS (commented):
Thanks for addressing my previous comments! Here are some new comments about outdated docstrings and `.gitignore` update.

.gitignore
jmoralez: I had thought about this as well. Can we define if `pytest` is going to be called from the top level directory (`BUILD_DIRECTORY` in [test.sh](https://github.com/microsoft/LightGBM/blob/master/.ci/test.sh)) or from some other path? Right now everything happens in `tests/distributed`, I even had to copy the binary there. I think it may make more sense to call `pytest` from the top level directory and modify the paths here so that `pytest` is called where the binary is and the files get created at `tests/distributed`. Is that what you suggested?

jmoralez: I added this idea in f7c6fddc2e82ea55773e72f7639ff5d6514a0cb4, please let me know what you think. `BINARY_DIR` would be `BUILD_DIRECTORY` and `TESTS_DIR` would be `BUILD_DIRECTORY/tests/distributed`

StrikerRUS: I suppose we can add an argument with `lightgbm[.exe]` executable file location and don't force that file to be in `distributed` directory.

https://stackoverflow.com/a/55301318

jose-moralez: In f7c6fdd the binary is assumed to be where it gets built (`BUILD_DIRECTORY`), so there's no need to move it anymore and the tests can be run with `pytest tests/distributed/_test_distributed.py`

StrikerRUS: Executable file is not always located in the `BUILD_DIRECTORY` folder. On Windows, for example, it is in `BUILD_DIRECTORY/windows/x64/Release` folder or `BUILD_DIRECTORY/Release` folder depending on how (via command line or Visual Studio GUI) executable was built.

jmoralez: I thought this was going to be tested on linux only. Will look into passing the executable location to pytest

StrikerRUS: Since this PR can be treated as some kind of framework for testing distributed training, I believe more general solutions here are important.

> Will work in passing the executable location to pytest

Thank you very much!

jmoralez: Added in 56113de, the argument is `--execdir`, so the tests could be run with `pytest tests/distributed/_test_distributed.py --execdir BUILD_DIRECTORY/Release`. Looking forward to your thoughts. I haven't used lightgbm on windows so I'm not sure if the executable has the `.exe` suffix

StrikerRUS: > I'm not sure if the executable has the `.exe` suffix

You're right - it has. But suffix can be omitted in calls. Both lines below result in opening Edge browser:

image


Just a question: why did you decide to introduce `execdir` argument and deduce full path to the executable later in tests but not directly `execfile`?

jose-moralez: Oh I didn't think of that haha, I'll change it to execfile.

StrikerRUS (approved):
Thank you very much for as always valuable contribution! LGTM, except one typo and one simplification.

jmoralez:
@StrikerRUS I've added your latest comments, let me know what you think.

StrikerRUS:
@jmoralez Yeah, I saw. Many thanks!
I've already approved this PR (https://github.com/microsoft/LightGBM/pull/4254#pullrequestreview-682344989).

@jameslamb There have been several changes since your latest approval. Would you like to take another look before merging?

jameslamb:
> @jameslamb There have been several changes since your latest approval. Would you like to take another look before merging?

yes please, will look soon

tests/distributed/_test_distributed.py
```diff
@@ -0,0 +1,198 @@
...
+        futures = []
+        with ThreadPoolExecutor(max_workers=self.n_workers) as executor:
+            for i in range(self.n_workers):
+                self.write_train_config(i)
+                train_future = executor.submit(self.worker_train, i)
+                futures.append(train_future)
+            results = [f.result() for f in futures]
+        for result in results:
+            if result.returncode != 0:
+                raise RuntimeError
```
jameslamb, line 135: can you please add a specific error message here? Otherwise, if this error is triggered during testing we'll just see `pytest`'s logs end with an empty `RuntimeError` and will have to work backwards from the line number in the stack trace to figure out what happened.

tests/distributed/_test_distributed.py
```diff
@@ -0,0 +1,198 @@
...
+
+    def predict(self, predict_config: Dict = {}) -> np.ndarray:
```
jameslamb, line 137:
```suggestion
    def predict(self, predict_config: Dict[str, Any] = {}) -> np.ndarray:
```

I think all of the keys in the config will be strings, right?

jameslamb (changes requested):
Overall this looks good to me. I just left some minor suggestions.

And one more general one... Can you please replace all uses of `pathlib` with `os.path`?

This project does not use `pathlib` anywhere else (you can confirm this with `git grep pathlib`), and keeping such things consistent reduces the maintenance burden for maintainers and friction for new contributors.

tests/distributed/_test_distributed.py
jmoralez: Since stderr from each worker is captured, pytest logs look like this:
image
Should I add something like `'Error in training'`?

jmoralez:
I think `pathlib` should be considered instead of `os.path`, for example this: https://github.com/microsoft/LightGBM/blob/0701a32da9fae7b2de8a01d702fa7d6abf36e836/python-package/lightgbm/__init__.py#L26-L29
becomes this:
```python
dir_path = Path(__file__).parent
version_file = dir_path / 'VERSION.txt'
if version_file.is_file():
    with version_file.open() as f:
```

StrikerRUS:
> I think `pathlib` should be considered instead of `os.path`, for example this:

I believe we can start migrating from `os.path` to `pathlib` similarly to #4136. @jameslamb WDYT?

jameslamb:
> I believe we can start migrating from os.path to pathlib

Unless there is a functional reason to prefer `pathlib` (e.g. efficiency, compatibility), I don't support transitioning this project's Python code from `os.path` to `pathlib`. I don't personally think that the example in https://github.com/microsoft/LightGBM/pull/4254#issuecomment-868880164 is very compelling.

Is there some functional reason to prefer it?

> similarly to #4136

I've observed that some other "good first issue" types of efforts (like #4136) have attracted contributions that required a lot of maintainer effort. In my opinion, that is worthwhile if the project gets some functional benefit from the changes, but not something that we should encourage for a change whose primary benefit, as I understand it, is a style preference.

If you'd like to make this change, @StrikerRUS , I won't oppose that. But I think the issue to document it should not use the `good first issue` tag, and that maintainers like you should do some of the work and not worry about preserving it for new contributors.

tests/distributed/_test_distributed.py
jameslamb: yes please

StrikerRUS:
> Is there some functional reason to prefer it?

It makes code **much** clearer and it's portable across platforms. With `os.path` module, ideally, we should use `os.pardir` instead of `..` and `os.path.sep` instead of `/` inside `os.path.join()`, for example. But we often forget about it:
https://github.com/microsoft/LightGBM/blob/bb39bc995f92d3ef19dab1a18dcb2cb8f3fae817/tests/c_api_test/test_.py#L17-L20

One good article:
https://treyhunner.com/2018/12/why-you-should-be-using-pathlib/


> I've observed that some other "good first issue" types of efforts (like #4136) have attracted contributions that required a lot of maintainer effort.

To be honest, I've never been a fan of "good first issue"s due to the reasons you've written above.

> If you'd like to make this change, @StrikerRUS , I won't oppose that.

Great! I'll prepare an issue then.

tests/distributed/_test_distributed.py
jameslamb:
```suggestion
                raise RuntimeError('Error in training')
```

Since this was the only remaining suggestion on this PR, since it is so small, and since I'm agreeing with your suggestion, I feel comfortable using maintainer permissions to commit this change directly. I'll approve and merge once this builds one more time.

jameslamb (approved):
Looks good, thanks very much for the help!

StrikerRUS:
@jameslamb Can we merge this?

jameslamb:
> @jameslamb Can we merge this?

yep! Sorry, should have merged it right after my review. I got distracted by something else and forgot to come back to this PR. Thanks for the `@`.

PR #4256, opened 2021-05-05
[R-package] move creation of character vectors in some methods to C++ side

jameslamb:
Another step towards #3016.

Resolves https://github.com/microsoft/LightGBM/pull/4155#discussion_r623899919.

This PR affects the following functions that return an R character vector:

* `LGBM_DatasetGetFeatureNames_R()`
* `LGBM_BoosterGetEvalNames_R()`
* `LGBM_BoosterSaveModelToString_R()`
* `LGBM_BoosterDumpModel_R()`

Currently, a buffer and an R character vector are allocated on the R side, then the corresponding C++ functions are called to write model data to that buffer and, eventually, copy it into that character vector.

This PR proposes simplifying that interaction by just creating that character vector from the C++ side and returning it to R. This has the following benefits:

* eliminates some unnecessary computation, like joining all eval names into one tab-separated string and then splitting it back apart on the R side
* removes code from the R package that was involved in managing buffers
* removes unnecessary code on the C++ side (`EncodeChar`, `R_CHAR_PTR`), replacing some LightGBM-custom stuff with standard routines available from R

### References

* https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Handling-character-data

### DIFF
--- a/R-package/src/lightgbm_R.cpp
+++ b/R-package/src/lightgbm_R.cpp
@@ -460,8 +460,6 @@ SEXP LGBM_BoosterGetEvalNames_R(LGBM_SE handle) {
   // if any eval names were larger than allocated size,
   // allow for a larger size and try again
   if (required_string_size > reserved_string_size) {
-    std::vector<std::vector<char>> names(len);
-    std::vector<char*> ptr_names(len);
     for (int i = 0; i < len; ++i) {
       names[i].resize(required_string_size);
       ptr_names[i] = names[i].data();
@@ -476,7 +474,6 @@ SEXP LGBM_BoosterGetEvalNames_R(LGBM_SE handle) {
         ptr_names.data()));
   }
   CHECK_EQ(out_len, len);
-  CHECK_GE(reserved_string_size, required_string_size);
   eval_names = PROTECT(Rf_allocVector(STRSXP, len));
   for (int i = 0; i < len; ++i) {
     SET_STRING_ELT(eval_names, i, Rf_mkChar(ptr_names[i]));
--- a/R-package/src/lightgbm_R.cpp
+++ b/R-package/src/lightgbm_R.cpp
@@ -641,12 +641,14 @@ SEXP LGBM_BoosterSaveModelToString_R(LGBM_SE handle,
   R_API_BEGIN();
   int64_t out_len = 0;
   int64_t buf_len = 1024 * 1024;
+  int64_t num_iter = Rf_asInteger(num_iteration);
+  int64_t importance_type = Rf_asInteger(feature_importance_type);
   std::vector<char> inner_char_buf(buf_len);
-  CHECK_CALL(LGBM_BoosterSaveModelToString(R_GET_PTR(handle), 0, Rf_asInteger(num_iteration), Rf_asInteger(feature_importance_type), buf_len, &out_len, inner_char_buf.data()));
+  CHECK_CALL(LGBM_BoosterSaveModelToString(R_GET_PTR(handle), 0, num_iter, importance_type, buf_len, &out_len, inner_char_buf.data()));
   // if the model string was larger than the initial buffer, allocate a bigger buffer and try again
   if (out_len > buf_len) {
     inner_char_buf.resize(out_len);
-    CHECK_CALL(LGBM_BoosterSaveModelToString(R_GET_PTR(handle), 0, Rf_asInteger(num_iteration), Rf_asInteger(feature_importance_type), out_len, &out_len, inner_char_buf.data()));
+    CHECK_CALL(LGBM_BoosterSaveModelToString(R_GET_PTR(handle), 0, num_iter, importance_type, out_len, &out_len, inner_char_buf.data()));
   }
   model_str = PROTECT(Rf_allocVector(STRSXP, 1));
   SET_STRING_ELT(model_str, 0, Rf_mkChar(inner_char_buf.data()));
@@ -662,12 +664,14 @@ SEXP LGBM_BoosterDumpModel_R(LGBM_SE handle,
   R_API_BEGIN();
   int64_t out_len = 0;
   int64_t buf_len = 1024 * 1024;
+  int64_t num_iter = Rf_asInteger(num_iteration);
+  int64_t importance_type = Rf_asInteger(feature_importance_type);
   std::vector<char> inner_char_buf(buf_len);
-  CHECK_CALL(LGBM_BoosterDumpModel(R_GET_PTR(handle), 0, Rf_asInteger(num_iteration), Rf_asInteger(feature_importance_type), buf_len, &out_len, inner_char_buf.data()));
+  CHECK_CALL(LGBM_BoosterDumpModel(R_GET_PTR(handle), 0, num_iter, importance_type, buf_len, &out_len, inner_char_buf.data()));
   // if the model string was larger than the initial buffer, allocate a bigger buffer and try again
   if (out_len > buf_len) {
     inner_char_buf.resize(out_len);
-    CHECK_CALL(LGBM_BoosterDumpModel(R_GET_PTR(handle), 0, Rf_asInteger(num_iteration), Rf_asInteger(feature_importance_type), out_len, &out_len, inner_char_buf.data()));
+    CHECK_CALL(LGBM_BoosterDumpModel(R_GET_PTR(handle), 0, num_iter, importance_type, out_len, &out_len, inner_char_buf.data()));
   }
   model_str = PROTECT(Rf_allocVector(STRSXP, 1));
   SET_STRING_ELT(model_str, 0, Rf_mkChar(inner_char_buf.data()));
--- a/R-package/src/lightgbm_R.cpp
+++ b/R-package/src/lightgbm_R.cpp
@@ -39,22 +39,8 @@
     return R_NilValue; \
   }
 
-using LightGBM::Common::Join;
 using LightGBM::Common::Split;
 using LightGBM::Log;
-
-LGBM_SE EncodeChar(LGBM_SE dest, const char* src, SEXP buf_len, SEXP actual_len, size_t str_len) {
-  if (str_len > INT32_MAX) {
-    Log::Fatal("Don't support large string in R-package");
-  }
-  INTEGER(actual_len)[0] = static_cast<int>(str_len);
-  if (Rf_asInteger(buf_len) < static_cast<int>(str_len)) {
-    return dest;
-  }
-  auto ptr = R_CHAR_PTR(dest);
-  std::memcpy(ptr, src, str_len);
-  return dest;
-}
 
 SEXP LGBM_GetLastError_R() {
   SEXP out;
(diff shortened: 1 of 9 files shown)


R-package/src/lightgbm_R.h
```diff
@@ -389,14 +385,11 @@ LIGHTGBM_C_EXPORT SEXP LGBM_BoosterGetLowerBoundValue_R(
 
 /*!
 * \brief Get Name of eval
-* \param eval_names eval names
+* \param handle Handle of booster
 * \return 0 when succeed, -1 when failure happens
```
StrikerRUS, line 389: Should be updated?

R-package/src/lightgbm_R.h
```diff
@@ -564,34 +557,28 @@ LIGHTGBM_C_EXPORT SEXP LGBM_BoosterSaveModel_R(
 
 /*!
 * \brief create string containing model
-* \param handle handle
+* \param handle Booster handle
 * \param num_iteration, <= 0 means save all
-* \param out_str string of model
-* \return 0 when succeed, -1 when failure happens
+* \param feature_importance_type type of feature importance, 0: split, 1: gain
+* \return length-1 R character vector with model string
```
StrikerRUS, line 563: What does mean `length-1`?..

R-package/src/lightgbm_R.cpp
```diff
@@ -616,31 +606,33 @@ SEXP LGBM_BoosterSaveModel_R(LGBM_SE handle,
...
-  SEXP feature_importance_type,
-  SEXP buffer_len,
-  SEXP actual_len,
-  LGBM_SE out_str) {
+  SEXP feature_importance_type) {
+  SEXP model_str;
   R_API_BEGIN();
   int64_t out_len = 0;
-  int64_t buf_len = static_cast<int64_t>(Rf_asInteger(buffer_len));
+  int64_t buf_len = 1024 * 1024;
```
StrikerRUS, line 613: What if `1024 * 1024` is not enough to save some big model? With removed "try with default len, repeat with actual if not enough" (`if (act_len > buf_len)`) this now looks like a regression compared to the current fully correct implementation.

C API docs says:
> **buffer_len** – String buffer length, if `buffer_len < out_len`, you should re-allocate buffer

StrikerRUS (commented):
I'm confused with removed reallocating buffer code...

R-package/src/lightgbm_R.cpp
jameslamb: OHHHH I see now, thank you for that explanation. I misunderstood the purpose of the code on the R side that was calling this function twice.

Ok yes you're right, that work needs to be done here. Will update it.

R-package/src/lightgbm_R.h
jameslamb:
```suggestion
* \return R character vector (length=1) with model string
```

Is this re-write clearer? It's common when talking about R to say `length-n vector` to mean "A vector with `n` elements", but if that's not clear to you it probably will be unclear to others reading this code.

This is one of those places where the fact that R doesn't have scalar types makes the language a little awkward. In most other languages we'd just say this is a string.

jameslamb:
I'm still experimenting with this, so marked it WIP to block merging even though other checks passed.

R-package/src/lightgbm_R.h
jameslamb: yes, thanks! updated

R-package/src/lightgbm_R.h
jameslamb: I updated to this, but will leave the thread unresolved in case you have a different suggestion

R-package/src/lightgbm_R.cpp
jameslamb: Alright, I've made this changes in recent commits.

I'm really glad you point this out, because it also made me realize an opportunity to allow larger feature names! Right now, the code in `LGBM_DatasetGetFeatureNames_R` on `master` will not allow any feature names longer than 256 characters.

https://github.com/microsoft/LightGBM/blob/f83180883a51729af9562ccac7b450c6da79a04a/R-package/src/lightgbm_R.cpp#L163

https://github.com/microsoft/LightGBM/blob/f83180883a51729af9562ccac7b450c6da79a04a/R-package/src/lightgbm_R.cpp#L179

The code below throws an error on `{lightgbm}` 3.2.1, but works as of this branch.

```r
library(lightgbm)

feature_names <- names(iris)
long_name <- paste0(rep("a", 1000L), collapse = "")
feature_names[1L] <- long_name
names(iris) <- feature_names
# check that feature name survived the trip from R to C++ and back
dtrain <- lgb.Dataset(
    data = as.matrix(iris[, -5L])
    , label = as.numeric(iris$Species) - 1L
)
dtrain$construct()
col_names <- dtrain$get_colnames()

# Error in lgb.call(fun_name = fun_name, ret = buf, ..., buf_len, act_len) : 
#  [LightGBM] [Fatal] Check failed: (reserved_string_size) >= (required_string_size) at lightgbm_R.cpp, line 177 .
```

But it should be possible to! Based on https://github.com/microsoft/LightGBM/blob/f83180883a51729af9562ccac7b450c6da79a04a/include/LightGBM/c_api.h#L295.

So I've updated the calls to `LGBM_DatasetGetFeatureNames_R` and `LGBM_BoosterGetEvalNames_R` to retry with a larger buffer on long names.

<hr>

I've added tests to this PR to check that this is working as expected.

* `LGBM_DatasetGetFeatureNames_R()`: `test_dataset.R`
* `LGBM_BoosterGetEvalNames_R()`:  I could not find a way to generate a large string value for this, but I might misunderstand how `LGBM_BoosterGetEvalNames` works. Opened #4264 with a question.
* `LGBM_BoosterSaveModelToString_R()`: `test_lgb.Booster.R`
* `LGBM_BoosterDumpModel_R()`: `test_lgb.Booster.R`

R-package/src/lightgbm_R.h
StrikerRUS: I tried to find a variable named `length` described somewhere earlier in the description and figure out why this vector contains `length-1` ("`length minus 1`") elements. 😄

R-package/src/lightgbm_R.cpp
StrikerRUS: > So I've updated the calls to `LGBM_DatasetGetFeatureNames_R` and `LGBM_BoosterGetEvalNames_R` to retry with a larger buffer on long names.

Great!
I think we can do the same for Python wrapper.
https://github.com/microsoft/LightGBM/blob/f83180883a51729af9562ccac7b450c6da79a04a/python-package/lightgbm/basic.py#L1896-L1900
https://github.com/microsoft/LightGBM/blob/f83180883a51729af9562ccac7b450c6da79a04a/python-package/lightgbm/basic.py#L3273-L3277
https://github.com/microsoft/LightGBM/blob/f83180883a51729af9562ccac7b450c6da79a04a/python-package/lightgbm/basic.py#L3472-L3476

R-package/src/lightgbm_R.h
```diff
@@ -388,14 +384,11 @@ LIGHTGBM_C_EXPORT SEXP LGBM_BoosterGetLowerBoundValue_R(
 
 /*!
 * \brief Get Name of eval
```
StrikerRUS, line 386:
```suggestion
* \brief Get names of eval metrics
```

R-package/src/lightgbm_R.h
```diff
@@ -388,14 +384,11 @@ LIGHTGBM_C_EXPORT SEXP LGBM_BoosterGetLowerBoundValue_R(
...
-* \param eval_names eval names
-* \return 0 when succeed, -1 when failure happens
+* \param handle Handle of booster
+* \return R character vector with names of eval sets
```
StrikerRUS, line 388: According to #4264.
```suggestion
* \return R character vector with names of eval metrics
```

R-package/src/lightgbm_R.h
```diff
@@ -583,34 +576,28 @@ LIGHTGBM_C_EXPORT SEXP LGBM_BoosterSaveModel_R(
...
   SEXP num_iteration,
-  SEXP feature_importance_type,
-  SEXP buffer_len,
-  SEXP actual_len,
-  LGBM_SE out_str
+  SEXP feature_importance_type
 );
 
 /*!
 * \brief dump model to json
```
StrikerRUS, line 591:
```suggestion
* \brief dump model to JSON
```

R-package/tests/testthat/test_lgb.Booster.R
```diff
@@ -240,6 +240,68 @@ test_that("Loading a Booster from a string works", {
...
+        , label = train$label
+        , num_leaves = 100L
+        , learning_rate = 0.01
+        , nrounds = 500L
+        , objective = "binary"
+        , save_name = tempfile(fileext = ".model")
+        , verbose = -1L
+    )
+
+    pred <- predict(bst, train$data)
```
StrikerRUS, line 258: Nice test!
I believe that with `rawscore = TRUE` or `predleaf = TRUE` this test will become even more strict and reliable.

R-package/tests/testthat/test_lgb.Booster.R
```diff
@@ -240,6 +240,68 @@ test_that("Loading a Booster from a string works", {
...
+        , nrounds = 200L
+        , objective = "binary"
+        , save_name = tempfile(fileext = ".model")
+        , verbose = -1L
+    )
+
+    model_json <- bst$dump_model()
+
+    # make sure this test is still producing a model bigger than the default
+    # buffer size used in LGBM_BoosterSaveModelToString
```
StrikerRUS, line 297:
```suggestion
    # buffer size used in LGBM_BoosterDumpModel_R
```

R-package/src/lightgbm_R.cpp
```diff
@@ -616,31 +636,43 @@ SEXP LGBM_BoosterSaveModel_R(LGBM_SE handle,
...
   std::vector<char> inner_char_buf(buf_len);
   CHECK_CALL(LGBM_BoosterSaveModelToString(R_GET_PTR(handle), 0, Rf_asInteger(num_iteration), Rf_asInteger(feature_importance_type), buf_len, &out_len, inner_char_buf.data()));
-  EncodeChar(out_str, inner_char_buf.data(), buffer_len, actual_len, static_cast<size_t>(out_len));
+  // if the model string was larger than the initial buffer, allocate a bigger buffer and try again
+  if (out_len > buf_len) {
+    inner_char_buf.resize(out_len);
+    CHECK_CALL(LGBM_BoosterSaveModelToString(R_GET_PTR(handle), 0, Rf_asInteger(num_iteration), Rf_asInteger(feature_importance_type), out_len, &out_len, inner_char_buf.data()));
```
StrikerRUS, line 649: Cast `num_iteration` and `feature_importance_type` only once.

R-package/src/lightgbm_R.cpp
```diff
@@ -616,31 +636,43 @@ SEXP LGBM_BoosterSaveModel_R(LGBM_SE handle,
...
   int64_t out_len = 0;
-  int64_t buf_len = static_cast<int64_t>(Rf_asInteger(buffer_len));
+  int64_t buf_len = 1024 * 1024;
   std::vector<char> inner_char_buf(buf_len);
   CHECK_CALL(LGBM_BoosterDumpModel(R_GET_PTR(handle), 0, Rf_asInteger(num_iteration), Rf_asInteger(feature_importance_type), buf_len, &out_len, inner_char_buf.data()));
-  EncodeChar(out_str, inner_char_buf.data(), buffer_len, actual_len, static_cast<size_t>(out_len));
+  // if the model string was larger than the initial buffer, allocate a bigger buffer and try again
+  if (out_len > buf_len) {
+    inner_char_buf.resize(out_len);
+    CHECK_CALL(LGBM_BoosterDumpModel(R_GET_PTR(handle), 0, Rf_asInteger(num_iteration), Rf_asInteger(feature_importance_type), out_len, &out_len, inner_char_buf.data()));
```
StrikerRUS, line 670: Cast `num_iteration` and `feature_importance_type` only once.

StrikerRUS (changes requested):
Thank you very much! Looks great overall!
I left a few minor comments below.

R-package/src/lightgbm_R.cpp
jameslamb: > I think we can do the same for Python wrapper.

Sure! But I think it should be a separate PR. I'd really like to focus on finishing #3016 soon.

R-package/tests/testthat/test_lgb.Booster.R
jameslamb: updated in https://github.com/microsoft/LightGBM/pull/4256/commits/6430f62dfc683449f6ac01d51a67505c7947efbc

R-package/src/lightgbm_R.cpp
jameslamb: updated in https://github.com/microsoft/LightGBM/pull/4256/commits/6430f62dfc683449f6ac01d51a67505c7947efbc

R-package/src/lightgbm_R.cpp
jameslamb: updated in https://github.com/microsoft/LightGBM/pull/4256/commits/6430f62dfc683449f6ac01d51a67505c7947efbc

StrikerRUS (approved):
Thanks for addressing comments!

StrikerRUS:
Just posted issue in Sphinx repository for failing `Static Analysis / check-docs (pull_request)` job: https://github.com/sphinx-doc/sphinx/issues/9188.

Issue #4208, opened 2021-04-21
[R-package] R handles produce segmentation faults when de-serialized


david-cortes:
The R interface uses a "handle" object which I guess stores a pointer to a C++ object. The objects these handles point to do not survive serialization in R, and will cause segmentation faults if they are serialized and de-serialized - which happens for example when restarting an R session.

Example:
```r
library(lightgbm)
data(agaricus.train, package = "lightgbm")
train <- agaricus.train
dtrain <- lgb.Dataset(train$data, label = train$label)
data(agaricus.test, package = "lightgbm")
test <- agaricus.test
dtest <- lgb.Dataset.create.valid(dtrain, test$data, label = test$label)
params <- list(objective = "regression", metric = "l2")
valids <- list(test = dtest)
model <- lgb.train(
    params = params
    , data = dtrain
    , nrounds = 5L
    , valids = valids
    , min_data = 1L
    , learning_rate = 1.0
    , early_stopping_rounds = 3L
)
```
After running it, restart the R session (assuming it's configured to save the environment between restarts), and execute these two lines again:
```r
library(lightgbm)
model <- lgb.train(
    params = params
    , data = dtrain
    , nrounds = 5L
    , valids = valids
    , min_data = 1L
    , learning_rate = 1.0
    , early_stopping_rounds = 3L
)
```
At that point it will produce a segmentation fault, crashing the R process.

I suppose the correct solution would be to use R's own external pointer object class and leave the destructor/free-er to the same R external pointer object. Those objects will also reset to `nullptr` upon de-serialization, which can then be checked inside the functions beforehand to avoid producing segmentation faults.

PR #4265, opened 2021-05-08
[R-package] manage Dataset and Booster handles as R external pointers (fixes #3016)

### LINKED ISSUE #3016: [R-package] Factor out custom R interface to lib_lightgbm

jameslamb:
Fixes #3016.

Contributes to #4208, might help with #4034.

## Description

When the R package calls functions from the C++ side that create persistent objects in memory, like a Dataset  or Booster, it stores a pointer to the memory address of such objects in a "handle". Today, these handles are stored in a double on 64-bit systems and as an int32 on 32-bit systems.

https://github.com/microsoft/LightGBM/blob/f83180883a51729af9562ccac7b450c6da79a04a/R-package/R/utils.R#L9-L15

Today, the R package uses LightGBM's custom R-to-C interface to manage these handles. This can lead to issues like #4208, and generally exposes LightGBM to compatibility problems if new versions of R introduce changes that require corresponding changes in that custom interface.

This PR proposes replacing that custom management of "handles" with R's built-in pattern for such things. R's C API has a concept called "external pointers" designed for exactly this purpose.

From https://cran.r-project.org/doc/manuals/r-release/R-exts.html#External-pointers-and-weak-references

> External pointer SEXPs are intended to handle references to C structures such as ‘handles’

## Notes for Reviewers

This only fixes #3016 if is is merged after #4256.

The "References" section below contains links to the resources that helped me to understand these concepts better, and they might help reviewers to understand these changes.

### References

* https://cran.r-project.org/doc/manuals/r-release/R-exts.html#External-pointers-and-weak-references
* https://cran.r-project.org/doc/manuals/r-release/R-exts.html#An-external-pointer-example
* https://cran.r-project.org/doc/manuals/r-devel/R-ints.html#SEXPTYPEs
* http://adv-r.had.co.nz/C-interface.html#c-data-structures
* https://github.com/dmlc/xgboost/blob/e41619b1fc6152d4392896c6caf9a0dc960c5558/R-package/src/xgboost_R.cc
* https://stackoverflow.com/a/26671781/3986677


StrikerRUS:
Wonderful!
Unfortunately, this PR doesn't fix the following compilation warning
```
/home/runner/work/LightGBM/LightGBM/lightgbm.Rcheck/00_pkg_src/lightgbm/src/src/lightgbm_R.cpp:720:52: warning: cast between incompatible function types from ‘SEXPREC* (*)(SEXP)’ {aka ‘SEXPREC* (*)(SEXPREC*)’} to ‘DL_FUNC’ {aka ‘void* (*)()’} [-Wcast-function-type]
```
as it was suspected in https://github.com/microsoft/LightGBM/issues/3016#issue-605212323
> I *suspect*, though I don't know for sure, that this would also resolve these warnings we're getting from some compilers:

Any ideas why it doesn't and whether we should create a separate issue for these warnings as this PR is closing #3016?

jameslamb:
> Any ideas why it doesn't and whether we should create a separate issue for these warnings as this PR is closing #3016?

Please create a separate issue for it and I can investigate it (and document which specific combinations of operating system + R version + compiler + installation method it occurs for) in the future.

jameslamb:
> Please create a separate issue for it and I can investigate it

@StrikerRUS I'm back at a computer now so I can make this issue.

jameslamb:
created https://github.com/microsoft/LightGBM/issues/4273 to track the compiler warning

StrikerRUS (changes requested):
Very nice PR!
Please consider checking some my questions and cleanup suggestions below.

StrikerRUS:
Hmmm, valgrind test is failing due to the time limit.
```
The job running on runner GitHub Actions 78 has exceeded the maximum execution time of 120 minutes.
```
Rerunning...

jameslamb:
whew, seems that it succeeded.

Total duration of https://github.com/microsoft/LightGBM/actions/runs/833251558 was 1 hour, 50 minutes, 46 seconds.

I think that if we see timeout issues again in the future, we should increase the allowed time by another 30 minutes.

StrikerRUS:
> I think that if we see timeout issues again in the future, we should increase the allowed time by another 30 minutes.

Agree.

StrikerRUS (approved):
Woo-hoo! Thank you!

jameslamb:
Excited about this one!!

PR #4279, opened 2021-05-12
Migrate to f-strings in python-package\lightgbm\plotting.py (#4136)

akshitadixit:
Related to #4136

### DIFF
--- a/python-package/lightgbm/plotting.py
+++ b/python-package/lightgbm/plotting.py
@@ -13,11 +13,11 @@ from .sklearn import LGBMModel
 def _check_not_tuple_of_2_elements(obj, obj_name='obj'):
     """Check object is not tuple or does not have 2 elements."""
     if not isinstance(obj, tuple) or len(obj) != 2:
-        raise TypeError('%s must be a tuple of 2 elements.' % obj_name)
+        raise TypeError(f"{obj_name} must be a tuple of 2 elements.")
 
 
 def _float2str(value, precision=None):
-    return ("{0:.{1}f}".format(value, precision)
+    return (f"{value:.{precision}f}"
             if precision is not None and not isinstance(value, str)
             else str(value))
 
@@ -206,8 +206,8 @@ def plot_split_value_histogram(booster, feature, bins=None, ax=None, width_coef=
 
     hist, bins = booster.get_split_value_histogram(feature=feature, bins=bins, xgboost_style=False)
     if np.count_nonzero(hist) == 0:
-        raise ValueError('Cannot plot split value histogram, '
-                         'because feature {} was not used in splitting'.format(feature))
+        raise ValueError("Cannot plot split value histogram, "
+                         f"because feature {feature} was not used in splitting")
     width = width_coef * (bins[1] - bins[0])
     centred = (bins[:-1] + bins[1:]) / 2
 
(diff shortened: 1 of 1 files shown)


python-package/lightgbm/plotting.py
```diff
@@ -13,11 +13,11 @@
 def _check_not_tuple_of_2_elements(obj, obj_name='obj'):
     """Check object is not tuple or does not have 2 elements."""
     if not isinstance(obj, tuple) or len(obj) != 2:
-        raise TypeError('%s must be a tuple of 2 elements.' % obj_name)
+        raise TypeError(f"{obj_name} must be a tuple of 2 elements.")
 
 
 def _float2str(value, precision=None):
-    return ("{0:.{1}f}".format(value, precision)
+    return (f"{value:.{precision}f}"
```
jameslamb, line 20: nice work! Did not know you could do string interpolation on the right-hand side of the `:` in an f-string.

I tested a few cases and confirmed that this still has the expected behavior.

```python
assert _float2str(0.0000001) == '1e-07'
assert _float2str(0.0000001, precision=1) == '0.0'
assert _float2str(0.0000001, precision=10) == '0.0000001000'
```

python-package/lightgbm/plotting.py
```diff
@@ -206,8 +206,8 @@ def plot_split_value_histogram(booster, feature, bins=None, ax=None, width_coef=
 
     hist, bins = booster.get_split_value_histogram(feature=feature, bins=bins, xgboost_style=False)
     if np.count_nonzero(hist) == 0:
-        raise ValueError('Cannot plot split value histogram, '
-                         'because feature {} was not used in splitting'.format(feature))
+        raise ValueError("Cannot plot split value histogram, "
+                         f"because feature {feature} was not used in splitting")
```
jameslamb, line 210:
```suggestion
        raise ValueError('Cannot plot split value histogram, '
                         f'because feature {feature} was not used in splitting')
```

Can you please keep single quotes here? I know that the codebase is somewhat inconsistent about this, but that's an issue for another day. For now, I'd prefer to keep the diff for this pull request as small as possible to minimize the impact on [the "git blame" view](https://docs.github.com/en/github/managing-files-in-a-repository/tracking-changes-in-a-file)

python-package/lightgbm/plotting.py
```diff
@@ -422,15 +422,15 @@ def add(root, total_count, parent=None, decision=None):
...
             if 'leaf_weight' in show_info:
-                label += '<br/>{0} weight'.format(_float2str(root['leaf_weight'], precision))
+                label += f"<br/>{_float2str(root['leaf_weight'], precision)} weight"
             if 'leaf_count' in show_info:
-                label += '<br/>count: {0}'.format(root['leaf_count'])
+                label += f"<br/>count: {root['leaf_count']}"
             if "data_percentage" in show_info:
-                label += '<br/>{0}% of data'.format(_float2str(root['leaf_count'] / total_count * 100, 2))
+                label += f"<br/>{_float2str(root['leaf_count'] / total_count * 100, 2)}% of data"
             label = "<" + label + ">"
```
jameslamb, line 434:
```suggestion
            label = f"<{label}>"
```

can you also please change this line? Concatenation with `+` is another type of string formatting we'd like to replace as part of #4136.

jameslamb: And can you please make the same change on line 420? I'm not able to comment on it.

jameslamb: Everything using `+=` is fine to leave just as you have, but for the first assignments with just `=`, we'd like to replace those with f-strings.

jameslamb (changes requested):
Thanks very much for taking the time to contribute! Please see a few suggestions below.

python-package/lightgbm/plotting.py
akshitadixit: Yes, even I learnt this today :)

python-package/lightgbm/plotting.py
akshitadixit: Sure.

StrikerRUS (approved):
LGTM!

jameslamb (approved):

PR #4283, opened 2021-05-13
fix calculation of weighted gamma loss (fixes #4174)

### LINKED ISSUE #4174: Regression Gamma Loss Gradient calculation is incorrect with weight
## Description
In src/objective/regression_objective.hpp, line 698 the weight term is only being applied to a single term, rather than the entire expression.  As a result, the weighted calculation of gamma objective is incorrect : 
gradients[i] = static_cast<score_t>(1.0 - label_[i] / std::exp(score[i]) * weights_[i]);

## Reproducible example
Lightgbm simplified code (incorrect) : (1.0 - y / exp(p) * w)
Xgboost simplified code (correct) : (1 - y / exp(p)) * w
Lightgbm simplifed code (correct) : (1-y/exp(p))*w
Tweedie, rho = 2 (gamma) simplified code : ((-y * exp(-1 * p) + exp(0 * p)) * w === (1 – y / exp(p)) * w

mayer79:
Fixes https://github.com/microsoft/LightGBM/issues/4174


jameslamb:
Thanks for this! I've edited the pull request title to be more descriptive. Pull request titles become changelog entries in this project (please see https://github.com/microsoft/LightGBM/releases/tag/v3.2.1).

jameslamb:
I don't think the CUDA build failures are related to these changes. All of the Dask tests are failing there with the following error

```text
>       from distributed.protocol.core import dumps_msgpack
E       ImportError: cannot import name 'dumps_msgpack' from 'distributed.protocol.core' (/root/miniconda/envs/test-env/lib/python3.7/site-packages/distributed/protocol/core.py)
```

I can look into this in a few hours. I suspect that this is a result of some new version of a dependency being released, will have to check them later to see.

src/objective/regression_objective.hpp
```diff
@@ -695,7 +695,7 @@ class RegressionGammaLoss : public RegressionPoissonLoss {
     } else {
       #pragma omp parallel for schedule(static)
       for (data_size_t i = 0; i < num_data_; ++i) {
-        gradients[i] = static_cast<score_t>(1.0 - label_[i] / std::exp(score[i]) * weights_[i]);
+        gradients[i] = static_cast<score_t>((1.0 - label_[i] / std::exp(score[i])) * weights_[i]);
```
lorentzenchr, line 698:
```suggestion
        gradients[i] = static_cast<score_t>((1.0 - label_[i] * std::exp(-score[i])) * weights_[i]);
```
Could be a tiny little bit faster.

lorentzenchr (approved):
This fixed the bug. What about a unit test for this?

src/objective/regression_objective.hpp
mayer79: @lorentzenchr 
Regarding timing: the same pattern (i.e. y/exp(p) instead of y*exp(-p) appears multiple time in the same file. My suggestion would be to open a new issue/PR that would change this consistently, along with 1-2 experiments on timings.

Regarding unit tests: I added some in a new R script. On the current master, it fails. With the gamma weight fix it passes.

lorentzenchr: the `exp` dominates timing by far.

shiyu1994: I agree that we should make the formula in the same file consistent. If we want to change division into multiplication, we should modify all these
https://github.com/microsoft/LightGBM/blob/adf36d766afb6261276d0624f4f5b216d51880c8/src/objective/regression_objective.hpp#L689-L700

jameslamb:
Ok @mayer79 , now that we've merged #4288, if you update to the latest `master` I think the failing checks should pass. Sorry for the inconvenience!

R-package/tests/testthat/test_weighted_loss.R
```diff
@@ -0,0 +1,66 @@
+context("Case weights are respected")
```
jameslamb, line 1: This test looks good to me, thank you! This might be the first unit test we've added that covers weighted training at all in the R package, so really appreciate it.

It's ok with me for this to live in a new test file as you've set it up. `test_basic.R` (where most of the other `lgb.train()` tests live) has gotten kind of big, and there are already uses of `lgb.train()` outside of that file in their own files (such as https://github.com/microsoft/LightGBM/blob/c629cb0b6bd2830894b710cbd4d8241b82ac3105/R-package/tests/testthat/test_learning_to_rank.R or https://github.com/microsoft/LightGBM/blob/c629cb0b6bd2830894b710cbd4d8241b82ac3105/R-package/tests/testthat/test_custom_objective.R).

jameslamb (commented):
Tests and changes look good to me! But another maintainer who's more familiar with gamma loss should probably review as well, so I'm just leaving a "comment" review.

src/objective/regression_objective.hpp
lorentzenchr: Done in #4289.

lorentzenchr:
@shiyu1994 After having reviewed #4289, maybe you'd like to have another look here as it fixes an actual bug.

shiyu1994 (approved):
The changes LGTM.

PR #4291, opened 2021-05-14
[R-package] Handle integer types more accurate in R-to-C interface

StrikerRUS:
Fix compilation warnings like the following one
```
C:\Users\runneradmin\RtmpMzXGiG\R.INSTALL11384d93526b\lightgbm\src\src\lightgbm_R.cpp(670,3): warning C4244: 'argument': conversion from 'int64_t' to 'int', possible loss of data [C:\Users\runneradmin\RtmpMzXGiG\R.INSTALL11384d93526b\lightgbm\src\build\_lightgbm.vcxproj]
```

Checked all `Rf_asInteger` and `INTEGER` calls and tried to bring data types in the consistency with original ones from 
https://github.com/microsoft/LightGBM/blob/master/include/LightGBM/c_api.h.

PR #4292, opened 2021-05-15
[python] Handle integer types more accurate in Python-to-C interface

StrikerRUS:
The same as for R-package in #4291.

Match function signatures from https://github.com/microsoft/LightGBM/blob/master/include/LightGBM/c_api.h.

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -786,8 +786,8 @@ class _InnerPredictor:
                 self.handle,
                 ptr_data,
                 ctypes.c_int(type_ptr_data),
-                ctypes.c_int(mat.shape[0]),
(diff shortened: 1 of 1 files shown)


jameslamb (approved):
I went through this this morning. My review strategy was to pull up the Python package from this PR's branch and then the source of `c_api.h` in an editor, side-by-side.

image

I looked for each `_LIB.something()` call in the Python package and at the corresponding signature from the `c_api` header. All the signatures seem to match now, thanks very much!

StrikerRUS:
@jameslamb Many thanks! My strategy was to search for `ctypes.` in `basic.py` file and track all calls of this module. I hope that two slightly different strategies give better chance to not overlook a mistake.

PR #4143, opened 2021-03-30
[python] added f-string to python-package/lightgbm/basic.py

NovusEdge:

PR #4293, opened 2021-05-15
[python] handle arbitrary length feature names in Python-package

StrikerRUS:
Refer to https://github.com/microsoft/LightGBM/pull/4256#discussion_r628743652.

### DIFF
--- a/tests/python_package_test/test_basic.py
+++ b/tests/python_package_test/test_basic.py
@@ -16,7 +16,9 @@ from .utils import load_breast_cancer
 def test_basic(tmp_path):
     X_train, X_test, y_train, y_test = train_test_split(*load_breast_cancer(return_X_y=True),
                                                         test_size=0.1, random_state=2)
-    train_data = lgb.Dataset(X_train, label=y_train)
+    feature_names = [f"Column_{i}" for i in range(X_train.shape[1])]
+    feature_names[1] = "a" * 1000  # set one name to a value longer than default buffer size
+    train_data = lgb.Dataset(X_train, label=y_train, feature_name=feature_names)
     valid_data = train_data.create_valid(X_test, label=y_test)
 
     params = {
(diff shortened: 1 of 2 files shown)


tests/python_package_test/test_basic.py
```diff
@@ -37,6 +39,8 @@ def test_basic(tmp_path):
         if i % 10 == 0:
             print(bst.eval_train(), bst.eval_valid())
 
+    assert train_data.get_feature_name() == feature_names
```
StrikerRUS, line 42: Current `master` fails here with
```
BufferError: Allocated feature name buffer size (255) wasinferior to the needed size (1001).
```

jameslamb: >  I overlooked bad string construction in #4143 

ah!! I see now

```python
required_string_vuffer_size.value
```

Sorry about that. I must have looked at that diff 10 times and never noticed that mistake (probably my mistake). I think we should consider adding `pylint` in CI. I see this error running `pylint python-package/ | grep vuffer`.

```text
python-package/lightgbm/basic.py:3462:124: E0602: Undefined variable 'required_string_vuffer_size' (undefined-variable)
python-package/build/lib/lightgbm/basic.py:3460:124: E0602: Undefined variable 'required_string_vuffer_size' (undefined-variable)
```

I mentioned that last year and then forgot about it :grimacing: https://github.com/microsoft/LightGBM/pull/3377#issuecomment-691286374. I should have written up an issue...I'll do that right now.

jameslamb: I just created https://github.com/microsoft/LightGBM/issues/4308

StrikerRUS: Wow! Very nice catch! I haven't noticed this too 🙁 .

 I was speaking about "no whitespace between words" issue.
```
BufferError: Allocated feature name buffer size (255) wasinferior to the needed size (1001).
                                                   ------^------
```

jameslamb (approved):
Looks great, thank you! The tests make sense to me, and I think these calls are correct. I got a bit confused about the difference between `actual_*`, `required_*`, and `reserved_*`, but I can't think of better names.

I compared these changes to the equivalent code in `c_api` and in `R-package/src/lightgbm_R.cpp`, and they all seem to match (except for `LGBM_BoosterGetFeatureNames()`, which is not used in the R package today).

tests/python_package_test/test_basic.py
jameslamb: haha oh! Then two mistakes :laughing:

PR #4297, opened 2021-05-16
[tests][python] Handle data types more accurate in C API test

StrikerRUS:
Linking #4292.

Match function signatures from https://github.com/microsoft/LightGBM/blob/master/include/LightGBM/c_api.h.

### DIFF
--- a/tests/c_api_test/test_.py
+++ b/tests/c_api_test/test_.py
@@ -51,10 +51,6 @@ dtype_int32 = 2
 dtype_int64 = 3
 
 
-def c_array(ctype, values):
-    return (ctype * len(values))(*values)
-
-
 def c_str(string):
     return ctypes.c_char_p(string.encode('utf-8'))
 
@@ -70,9 +66,9 @@ def load_from_file(filename, reference):
         ref,
         ctypes.byref(handle))
     print(LIB.LGBM_GetLastError())
-    num_data = ctypes.c_long()
+    num_data = ctypes.c_int(0)
     LIB.LGBM_DatasetGetNumData(handle, ctypes.byref(num_data))
-    num_feature = ctypes.c_long()
+    num_feature = ctypes.c_int(0)
     LIB.LGBM_DatasetGetNumFeature(handle, ctypes.byref(num_feature))
     print('#data: %d #feature: %d' % (num_data.value, num_feature.value))
     return handle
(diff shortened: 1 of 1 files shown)


tests/c_api_test/test_.py
```diff
@@ -99,22 +95,27 @@ def load_from_csr(filename, reference):
         ref = reference
 
     LIB.LGBM_DatasetCreateFromCSR(
-        c_array(ctypes.c_int, csr.indptr),
-        dtype_int32,
-        c_array(ctypes.c_int, csr.indices),
-        csr.data.ctypes.data_as(ctypes.POINTER(ctypes.c_void_p)),
-        dtype_float64,
+        csr.indptr.ctypes.data_as(ctypes.POINTER(ctypes.c_int32)),
```
jameslamb, line 98: could this use `.byref()` instead of `POINTER()`?

From https://docs.python.org/3/library/ctypes.html#passing-pointers-or-passing-parameters-by-reference

> ctypes exports the `byref()` function which is used to pass parameters by reference. The same effect can be achieved with the `pointer()` function, although pointer() does a lot more work since it constructs a real pointer object, so it is faster to use `byref()` if you don’t need the pointer object in Python itself.

I'm very unfamiliar with `ctypes` so apologies if that doesn't make sense.

tests/c_api_test/test_.py
```diff
@@ -127,31 +128,36 @@ def load_from_csc(filename, reference):
             values = line.split('\t')
             data.append([float(x) for x in values[1:]])
             label.append(float(values[0]))
-    mat = np.array(data)
+    mat = np.array(data, dtype=np.float64)
     label = np.array(label, dtype=np.float32)
-    csr = sparse.csc_matrix(mat)
+    csc = sparse.csc_matrix(mat)
```
jameslamb, line 133: thanks for catching this!

jameslamb (approved):
Looks good! Thanks very much for breaking long lines like into one-argument-per-line. I think that makes it easier to read, which is really important since some of the functions take so many positional arguments.

I just left one question about how we're using `ctypes` but it isn't directly related to this PR and could be addressed in a later PR if something needs to change.

tests/c_api_test/test_.py
StrikerRUS: I wish I knew... The cited paragraph looks like we can benefit from using `byref` instead of `pointer`. I will create a new issue for this idea to not lose in the merged PR and read about it.

jameslamb: Ok thanks very much!

PR #4302, opened 2021-05-19
[python-package] Improve Graphviz import error message (fixes #4299)

### LINKED ISSUE #4299: [python-package] Graphviz ImportError confusion
When trying to plot the tree, the ImportError regarding Graphviz 'You must install graphviz to plot tree.' can be a little confusing since one also must restart their session in addition to installing it. This has been brought up before but the error message wasn't changed - @jameslamb was it intentionally left out or may I create a PR to add that clarification?

---

'I think we could add a better message like "if you've recently installed graphviz, be sure to restart your session". We'd welcome a PR to add that if you'd like to contribute it! Otherwise I will probably add it in a few days.'
_Originally posted by @jameslamb in https://github.com/microsoft/LightGBM/issues/2428#issuecomment-533693087_

AngelikaAntsmae:
Added a specification to restart the session to the import error message regarding Graphviz.

### DIFF
--- a/python-package/lightgbm/plotting.py
+++ b/python-package/lightgbm/plotting.py
@@ -379,7 +379,7 @@ def _to_graphviz(tree_info, show_info, feature_names, precision=3,
     if GRAPHVIZ_INSTALLED:
         from graphviz import Digraph
     else:
-        raise ImportError('You must install graphviz to plot tree.')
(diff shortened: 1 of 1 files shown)


jameslamb (approved):
Looks good to me, thanks very much for taking the time to contribute!

We hope you'll consider coming back to contribute more in the future :grinning:

PR #4304, opened 2021-05-20
[python] improve error message for required packages

StrikerRUS:
Based on #4302.

It turned out that some users don't know that python cannot find packages installed during the current session by default. So it will be good to tell them about this fact in error messages.


jameslamb (approved):
Nice, I think this will be helpful for beginners.

PR #4322, opened 2021-05-26
[fix] fix GatherInfoForThresholdNumerical boundary (fix #4286)

### LINKED ISSUE #4286: "Check failed: (best_split_info.left_count) > (0)" in forced_split mode
I met this when using "forcedsplit_filename". following data and code will repeat this issue.  The only clue I can provide is that  the num of nan in features may affect some split process with forcedsplits.

lgbversion:3.2.1
xdata:
[checkdata.zip](https://github.com/microsoft/LightGBM/files/6477570/checkdata.zip)

```
import lightgbm as lgb
import numpy as np
import json
x=np.load($xdata)
y=np.random.rand(x.shape[0])
forcesplit_json={'feature':1,'threshold':0.3,'right':{'feature':1,'threshold':0.7}}
tmppath=$somerandpath
with open(tmppath,'w') as f:
    json.dump(forcesplit_json,f)
params={'objective':'mse','forcedsplit_filepath':tmppath,'max_bin':64}
ds=lgb.Dataset(data=x,label=y,params=params)
lgb.train(ds,params=params)

```

shiyu1994:
This is to fix #4286. In `GatherInfoForThresholdNumerical `, since it is calculating the right sum statistics, it should break when `static_cast<uint32_t>(t + offset) <= threshold` instead of `static_cast<uint32_t>(t + offset) < threshold`. Because data with bin value `<= threshold` are split into the left child node.

### DIFF
--- a/src/treelearner/feature_histogram.hpp
+++ b/src/treelearner/feature_histogram.hpp
@@ -572,7 +572,7 @@ class FeatureHistogram {
     const double cnt_factor = num_data / sum_hessian;
     // from right to left, and we don't need data in bin0
     for (; t >= t_end; --t) {
-      if (static_cast<uint32_t>(t + offset) < threshold) {
+      if (static_cast<uint32_t>(t + offset) <= threshold) {
         break;
       }
 


jameslamb (approved):
nice, thank you!

shiyu1994:
BTW, currently our force splits method only produces splits with `default_left=true`, so only sweeping from the right side of the histogram is implemented in the GatherInfoForThresholdNumerical. Meanwhile, for categorical feature, the split condition can only be a single category value, instead of a set of category values as in normal categorical splits. So I think it requires further efforts to complete the functionality of force splits.

shiyu1994:
It seems that forced splits functionality is not covered by any test cases. I think we may need to add tests for it.

PR #4325, opened 2021-05-26
[fix] skip empty bins when calculating cnt_in_bin in BinMapper::FindBin (fix #4301)

### LINKED ISSUE #4301: Running fit method on LGBMRegressor kills Jupyter Kernel
Hello team,

I'm facing a weird behavior of LGBMRegressor. Sometimes it causes Jupyter Kernel to die on some datasets.

I was able to minimize the test case.

The code itself is pretty simple:
import pandas as pd
import numpy as np
import lightgbm
test = pd.read_pickle('weird.pkl')
lightgbm.LGBMRegressor().fit(test.drop(columns=['y']), test['y'])

In order to reproduce you will need to launch it for the specific dataset:
[weird.zip](https://github.com/microsoft/LightGBM/files/6508547/weird.zip)

versions:
pandas - '1.1.5', numpy - '1.19.5', lightgbm - '3.2.1'
I was able to reproduce on Windows and on Ubuntu

pip install lighgbm

Could you please advise how would it be possible to tackle this point?

Thank you!

shiyu1994:
As described in https://github.com/microsoft/LightGBM/issues/4301#issuecomment-848622040, this is to fix #4301 by skipping the empty bins when calculating cnt_in_bin in `BinMapper::FindBin`. So that we can correctly decides the `most_freq_bin_`.

### DIFF
--- a/src/io/bin.cpp
+++ b/src/io/bin.cpp
@@ -411,7 +411,7 @@ namespace LightGBM {
         cnt_in_bin.resize(num_bin_, 0);
         int i_bin = 0;
         for (int i = 0; i < num_distinct_values; ++i) {
-          if (distinct_values[i] > bin_upper_bound_[i_bin]) {
+          while (distinct_values[i] > bin_upper_bound_[i_bin] && i_bin < num_bin_ - 1) {
             ++i_bin;
           }
           cnt_in_bin[i_bin] += counts[i];


jameslamb:
@shiyu1994 sorry for the inconvenience, could you please merge `master` into this? That should get you the fix from #4326, which I think should avoid timeout issues from QEMU builds.

jameslamb (approved):
thank you very much!

I built this branch locally and ran a modified version of the reproducible example from https://github.com/microsoft/LightGBM/issues/4301#issuecomment-847521799 10 times tonight.

```python
import zipfile
from io import BytesIO

import lightgbm
import pandas as pd
import numpy as np
import requests

from sklearn.metrics import mean_absolute_error

data_url = "https://github.com/microsoft/LightGBM/files/6508547/weird.zip"

zipdata = BytesIO()
zipdata.write(requests.get(data_url, headers={"Accept": "application/octet-stream"}).content)
zip_contents = zipfile.ZipFile(zipdata)
data_file = zip_contents.open("weird.pkl")

test = pd.read_pickle(data_file)

bst = lightgbm.LGBMRegressor(verbose=1, num_iterations=5000).fit(test.drop(columns=["y"]), test["y"])

mae = mean_absolute_error(bst.predict(test.drop(columns=["y"])), test["y"])

print(f"target mean: {np.mean(test['y'])}")
print(f"MAE: {mae}")
```

Training succeeded every time, and increasing `num_iterations` led to a better fit to the training data. Given that and the fact that all CI jobs are passing, I'm confident in this fix.

PR #3996, opened 2021-02-17
[dask] run Dask tests on aarch64 architecture

StrikerRUS:
Refer to https://github.com/microsoft/LightGBM/pull/3948#discussion_r575694158.

Now that QEMU is setup and proved to be stable on our CI, let's run Dask tests on aarch64 architecture and see what happens.

PR #4326, opened 2021-05-26
[ci] Increase timeout value for QEMU builds

StrikerRUS:
`120` is too close to values reported in https://github.com/microsoft/LightGBM/pull/3996#issuecomment-845068961 and can sometimes be not enough to finish all tests (for example, here https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=10157&view=logs&j=c2f9361f-3c13-57db-3206-cee89820d5e3). Sorry, I had to think about this in #3996.


jameslamb (approved):
thanks!

Issue #4315, opened 2021-05-24
[CUDA] illegal memory access when using CUDA and small max_bin


RobinDong:
## Description
By using CUDA histogram of the master branch, the simple python code report memory error if it uses `max_bin: 7`

### Build it with CUDA ###
```cmake -DCMAKE_CUDA_COMPILER=/usr/local/cuda/bin/nvcc -DUSE_CUDA=ON -DUSE_DEBUG=ON -DCUDA_DEBUG=3 ..```
### Run python snippet below ###
```
import numpy as np
import pandas as pd
import lightgbm as lgb

# Generate fake data
NUM_DATA = 100000

data_map = {}
for index in range(200):
    col = f"col{index}"
    data_map[col] = np.random.rand(NUM_DATA)
    if index == 0:
        col0 = data_map[col]

data_map["target"] = col0

df = pd.DataFrame(data_map)
train = df[NUM_DATA//10:]
test = df[0:NUM_DATA//10]

train_target = train['target']
test_target = test['target']
train = train.drop(['target'], axis=1)
test = test.drop(['target'], axis=1)
valid_set = lgb.Dataset(test.values, label=test_target.values)
train_set = lgb.Dataset(train.values, label=train_target.values)

# Set parameters and start to train
params = {
    'boosting_type': 'gbdt',
    'objective': 'regression',
    'metric': {'l2', 'l1'},
    'device_type': 'cuda',
    'verbose': 2,
    'max_bin': 7,
}

print('LightGBM training started')
print('Parameters: %s' % params)

trained_model = lgb.train(
    params,
    train_set,
    valid_sets=valid_set)

print(trained_model)
```
And it will report error:
```
[LightGBM] [Fatal] [CUDA] an illegal memory access was encountered LightGBM/src/treelearner/cuda_tree_learner.cpp 239

terminate called after throwing an instance of 'std::runtime_error'
  what():  [CUDA] an illegal memory access was encountered LightGBM/src/treelearner/cuda_tree_learner.cpp 239

Aborted (core dumped)
```

## Environment info
Operating System: Linux

CPU Model: Intel(R) Xeon(R) CPU @ 2.30GHz
GPU model: Nvidia T4
CUDA: 11.0.182
Python 3.6.9
LightGBM version or commit hash: `master` branch

Command(s) you used to install LightGBM

```shell
mkdir build
cd build
[...]

PR #4327, opened 2021-05-28
[CUDA] fix CUDA memory error by reducing block number (#4315)

RobinDong:
By using the log from https://github.com/microsoft/LightGBM/blob/master/src/treelearner/cuda_tree_learner.cpp#L334, we could see the CUDA memory allocated is 18000000 bytes,  when running the example code of https://github.com/microsoft/LightGBM/issues/4315. But by adding a log on https://github.com/microsoft/LightGBM/blob/master/src/treelearner/kernels/histogram_16_64_256.cu#L148, I found that the access offset defined by `feature_id * feature_size` is much larger than 18000000 bytes. This is the cause of the memory accessing error.
Actually, the `feature_id` is already the identity number of these dense features, so we don't need, and can't multiply the `num_workgroups` for CUDA kernel in https://github.com/microsoft/LightGBM/blob/master/src/treelearner/cuda_kernel_launcher.cu

I have already tested some self-made cases for LightGBM(CUDA) with and without multiply of `num_workgroups` in the CUDA kernel. The test results show that they could reach the same model accuracy or loss.

### DIFF
--- a/src/treelearner/cuda_kernel_launcher.cu
+++ b/src/treelearner/cuda_kernel_launcher.cu
@@ -38,20 +38,20 @@ void cuda_histogram(
     if (leaf_num_data == num_data) {
       if (use_all_features) {
         if (!is_constant_hessian)
-          histogram16<<<16*num_workgroups, 16, 0, stream>>>(arg0, arg1, arg2,
+          histogram16<<<num_workgroups, 16, 0, stream>>>(arg0, arg1, arg2,
                   arg3, arg4, arg5,
                   arg6, arg7, arg8, static_cast<acc_type*>(arg9), exp_workgroups_per_feature);
         else
-           histogram16<<<16*num_workgroups, 16, 0, stream>>>(arg0, arg1, arg2,
+           histogram16<<<num_workgroups, 16, 0, stream>>>(arg0, arg1, arg2,
                   arg3, arg4, arg5,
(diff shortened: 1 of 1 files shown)


shiyu1994:
@RobinDong Thanks for your fix. I think the authors originally intended to increase the number of blocks when the threads per block is smaller, to better exploit parallelism. This is a quick fix, but before merging this, let me test the change in efficiency.

src/treelearner/cuda_kernel_launcher.cu
```diff
@@ -38,41 +38,41 @@ void cuda_histogram(
     if (leaf_num_data == num_data) {
       if (use_all_features) {
         if (!is_constant_hessian)
-          histogram16<<<16*num_workgroups, 16, 0, stream>>>(arg0, arg1, arg2,
+          histogram16<<<num_workgroups, 16, 0, stream>>>(arg0, arg1, arg2,
```
shiyu1994, line 41: Perhaps a better fix would be to change how the `group_id` is calculated in the kernel, instead of simple remove of the multiplications. I need to perform some evaluation to see how the efficiency changes with fewer number of blocks.

shiyu1994:
After testing the alternative solution mentioned above, I found it has no advantage over the current modification in this PR. Thanks for your nice fix @RobinDong !

shiyu1994 (approved):

PR #4330, opened 2021-05-29
[tests][ci] run cpp tests with sanitizers on Linux and macOS

StrikerRUS:
Current way of running `testlightgbm.exe` on Windows doesn't return actual status code. Take a look at the logs in `master`:
```
...

[----------] Global test environment tear-down
[==========] 17 tests from 2 test suites ran. (13 ms total)
[  PASSED  ] 16 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] AtofPreciseTest.CornerCases
```
https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=10169&view=logs&j=1f4df553-f999-5fff-c6fe-71123c872ae2&t=5236b51f-5710-5c9e-d20e-7548daf49042
Proposed call in this PR makes `master` Windows job red.

---

Looks like AppleClang doesn't have `leak` sanitizer. So set `SANITIZERS` to only `"address;undefined"` for macOS.
```
-- The C compiler identification is AppleClang 12.0.0.12000032
-- The CXX compiler identification is AppleClang 12.0.0.12000032

...

clang: error: unsupported option '-fsanitize=leak' for target 'x86_64-apple-darwin19.6.0'
```
> It seems that the Clang/LLVM shipped by Apple does not have `-fsanitize=leak` support.
   https://stackoverflow.com/a/55778432

https://developer.apple.com/documentation/xcode/diagnosing-memory-thread-and-crash-issues-early

---

MSVC is just about getting support of ASan
https://devblogs.microsoft.com/cppblog/addresssanitizer-asan-for-windows-with-msvc/
https://devblogs.microsoft.com/cppblog/address-sanitizer-for-msvc-now-generally-available/
https://docs.microsoft.com/en-us/cpp/sanitizers/?view=msvc-160

> Current support is limited to x86 and x64 on Windows 10. Send us feedback on what you'd like to see in future releases. Your feedback helps us prioritize other sanitizers for the future, such as `/fsanitize=thread`, `/fsanitize=leak`, `/fsanitize=memory`, `/fsanitize=undefined`, or `/fsanitize=hwaddress`. You can report bugs here if you run into issues.
   https://docs.microsoft.com/en-us/cpp/sanitizers/asan?view=msvc-160

```
...

cl : Command line warning D9002: ignoring unknown option '-fsanitize=address' [D:\a\1\s\build\_deps\googletest-build\googletest\gtest.vcxproj]
  gtest-all.cc
[...]

PR #4336, opened 2021-06-03
Log warning instead of fatal when parsing float get under/overflow.

cyfdecyf:
For texts that resolve to infinity, under or overflow should be accepted.

This should resolve unit test failure mention in #4330.

### DIFF
--- a/include/LightGBM/utils/common.h
+++ b/include/LightGBM/utils/common.h
@@ -347,7 +347,7 @@ inline static const char* AtofPrecise(const char* p, double* out) {
     Log::Fatal("no conversion to double for: %s", p);
   }
   if (errno == ERANGE) {
-    Log::Fatal("convert to double got underflow or overflow: %s", p);
+    Log::Warning("convert to double got underflow or overflow: %s", p);
   }
   return end2;
 }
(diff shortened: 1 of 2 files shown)


StrikerRUS:
Please update cpp tests code.
https://github.com/microsoft/LightGBM/blob/master/tests/cpp_tests/test_common.cpp
Right now UnderOverFlow test is failing with:
```
...

[ RUN      ] AtofPreciseTest.UnderOverFlow
[LightGBM] [Warning] convert to double got underflow or overflow: 1e+400
/Users/runner/work/1/s/tests/cpp_tests/test_common.cpp:108: Failure
Expected: LightGBM::Common::AtofPrecise("1e+400", &got) throws an exception of type std::runtime_error.
  Actual: it throws nothing.

...
```

cyfdecyf:
@StrikerRUS Thanks for spotting this. I ran a single test case and didn't notice there's outdated test case.

Fix pushed.

StrikerRUS (commented):
Thank you!

I don't think I'm qualified to actually approve this PR, so just commenting.

shiyu1994 (approved):
@cyfdecyf Thanks for your contribution. The changes LGTM.

Issue #4310, opened 2021-05-20
release 3.3.0


jameslamb:
## Summary

I'd like to start a discussion on this question:

> Should we do a v3.3.0 release of LightGBM some time in June?

## Motivation

My primary motivation for this proposal is that user reports in the last two months have revealed critical issues with the R package (#4007, #4045, #4216, #4259, #4305), and these are fixed by recent PRs related to #3016 (especially #4155 and #4247).

But looking at the PRs that have been merged since LightGBM 3.2.1, there are a lot of other useful fixes, including:

* fix for issue in data parallel learning when feature distributions on partitions don't overlap (#4185)
* more accurate handoffs between C++ side and wrapper packages (#4291, #4292)
* changes to MRO to match scikit-learn's preferences (#3192)
* fix warnings from CUDA builds (#4268)

There has only been one technically "breaking" PR merged, and I think it's ok to include in a minor release (#4197).

## Proposal

I'd like to get other maintainers' thoughts on the following proposal.

1. Do not merge #3234 yet
2. Prepare a 3.3.0 release, to be released some time in June
3. Complete any fixes / feature requests that maintainers suggest before that release

Personally, I'd like to include at least the following in a 3.3.0 release:

* C++
    - [x] #4272
    - [x] incorrect calculation of gamma metric in weighted training (#4174 / #4283)
* Dask
    - [x] eval sets (#4101)
    - [ ] *(not critical)* inconsistencies in leaf counts when performing multiclass classification (#4220)
* Python package
    - [x] fix unexplained jupyter kernel restarts (#4301)
* R package

    - [x] prevent segfaults when objects are serialized / deserialized (#4208)
        - [x] https://github.com/microsoft/LightGBM/pull/4586
    - [x] a deprecation warning for `...` in the R package (https://github.com/microsoft/LightGBM/issues/4226#issuecomment-829473584)
        - [x] https://github.com/microsoft/LightGBM/pull/4571
        - [x] https://github.com/microsoft/LightGBM/pull/4581
        - [x] https://github.com/microsoft/LightGBM/pull/4522
        - [x] https://github.com/microsoft/LightGBM/pull/4523
        - [x] https://github.com/microsoft/LightGBM/pull/4548
        - [x] https://github.com/microsoft/LightGBM/pull/4554
        - [x] https://github.com/microsoft/LightGBM/pull/4572
        - [x] https://github.com/microsoft/LightGBM/pull/4573
    - [x] fix memory leaks (#4282)
[...]

Issue #4226, opened 2021-04-26
[RFC] [R-package] Remove support for passing parameters through '...'


jameslamb:
## Question

I'm opening this request for comment (RFC) to get feedback on the following proposal, originally mentioned in https://github.com/microsoft/LightGBM/pull/4202#discussion_r616083553.

> **In its next major release, should `{lightgbm}` remove support for passing parameters through `...` in function calls?**

This would apply to:

* `lgb.cv()`
* `lgb.Dataset()`
* `lgb.Dataset.create.valid()`
* `lgb.train()`
* `lightgbm()`

## Description

LightGBM's core library supports many parameters, documented at https://lightgbm.readthedocs.io/en/latest/Parameters.html. The exported functions in `{lightgbm}` (the R package), support at least 3 different ways to pass values for these parameters.

* in a named list called `params`
* (for some parameters) via explicit keyword arguments
* via implicit keyword arguments with `...`

For example, the following calls to `lightgbm::lightgbm()` are all equally valid ways to say "perform 14 boosting rounds".

```r
library(lightgbm)

data(agaricus.train, package = "lightgbm")
train <- agaricus.train

# option 1 - params
bst <- lightgbm(
    data = train$data
    , label = train$label
    , objective = "binary"
    , params = list(
        num_iterations = 14L
    )
)

# option 2 - explicit keyword argument ("nrounds")
bst <- lightgbm(
    data = train$data
    , label = train$label
    , objective = "binary"
    , nrounds = 14L
)

# option 3 - passing params through ...
bst <- lightgbm(
    data = train$data
    , label = train$label
    , objective = "binary"
    , num_iterations = 14L
)
```

## Motivation

**pros**

* reduces friction for new users getting familiar with LightGBM
* turns some situations that are currently warnings into failures (https://github.com/microsoft/LightGBM/pull/4202#discussion_r616097010)
* reduces maintenance burden of the R package by limiting the number of combinations of inputs that have to be tested

**cons**

* breaks existing code that relies on the `...` behavior
    - NOTE: maintainers have agreed that the next LightGBM release will be a 4.0.0 release, so this breaking change by itself wouldn't force a new major version of LightGBM

## References

* `{catboost}` does not support passing parameters through `...`: https://github.com/catboost/catboost/blob/80fa998c355fb9e5e937893e8b7d67286352965c/catboost/R-package/R/catboost.R#L1478
[...]

PR #4337, opened 2021-06-04
[R-package] remove unused code in lgb.params2str()

jameslamb:
While starting to work on adding a deprecation warning for uses of `...` in the R package (as part of #4310), I found some unused code that I think can be removed.

`lgb.params2str()`, an internal-only utility function, is only ever called with the single keyword argument `params`.

This PR proposes removing `...` from that function's signature.

Contributes to #4226


StrikerRUS (approved):
LGTM!

PR #4338, opened 2021-06-04
[R-package] remove uses of ... in Predictor constructor

jameslamb:
Contributes to #4226

This PR proposes removing the use of `...` from the constructor for the `Predictor` class in the R package. That class is not exported in the package's namespace, so this is not a user-facing breaking change.

Removing this use of `...` makes the package's code slightly stricter, which reduces the risk of bugs.


StrikerRUS (approved):
LGTM!

PR #4340, opened 2021-06-04
[docs] make building of C++ tests section collapsable

StrikerRUS:
Result:

image


jameslamb (approved):
ah yes, I think that looks great. Thank you!

PR #4341, opened 2021-06-05
[R-package] resolve test warning about is.na() and handles

jameslamb:
This PR resolves the following warning in the R package's tests, which was probably introduced in #4265

> Warning message:
    In is.na(handle) :
    is.na() applied to non-(list or vector) of type 'externalptr'

### DIFF
--- a/R-package/tests/testthat/test_dataset.R
+++ b/R-package/tests/testthat/test_dataset.R
@@ -82,7 +82,8 @@ test_that("lgb.Dataset: Dataset should be able to construct from matrix and retu
     , lightgbm:::lgb.params2str(params = list())
     , ref_handle
   )
-  expect_false(is.na(handle))
+  expect_is(handle, "externalptr")
+  expect_false(is.null(handle))
   .Call(LGBM_DatasetFree_R, handle)
   handle <- NULL
 })


StrikerRUS (approved):
LGTM! Was that warning shown somewhere in our CI logs (I can't find)?

jameslamb:
> LGTM! Was that warning shown somewhere in our CI logs (I can't find)?

I didn't see the warning message itself in any of our CI logs, but for any jobs where we run `Rscript testthat.R` instead of `R CMD CHECK`, `{testthat}`  produces the following summary at the end:

```text
[ FAIL 0 | WARN 1 | SKIP 3 | PASS 655 ]
```

For example, you can see that at the end of the test with sanitizers: https://github.com/microsoft/LightGBM/runs/2736608112?check_suite_focus=true

I noticed that "WARN 1" and then ran the tests locally to see what it was.

I'm not sure if there is a way to get the warning text to display in our CI logs. In that job with sanitizers (for example), we are already redirecting both stderr and stdout

https://github.com/microsoft/LightGBM/blob/441f38bb8140cab9ea3076f903e66541a5c48785/.github/workflows/r_package.yml#L193

Issue #2479, opened 2019-09-30
[R-package] Export callback functions


jameslamb:
See [this comment](https://github.com/microsoft/LightGBM/pull/2443#issuecomment-536321691) and [this comment thread](https://github.com/microsoft/LightGBM/pull/2443#issuecomment-536314106) from #2443 . Currently the functions in `callback.R` are not exported in the R package. This means that the use of `callbacks` keyword argument in `lgb.train()` is unclear.

To close this issue, submit a PR with the following features

* [ ] Roxygen documentation on relevant functions in `R-package/callbacks.R`, including `@export` tag. Consider also using `@family` tag to auto-generate `see-also` links.
* [ ] regenerated `.Rd` documentation (see [this comment](https://github.com/microsoft/LightGBM/pull/2437#issuecomment-536244608) for how to do this). This should also regenerate `R-package/NAMESPACE`.
* [ ] updated documentation in any other functions that expose `callbacks`, linking to the relevant documentation

jameslamb:
Closing this in favor of being in #2302 . We keep all feature requests in #2302 . If you'd like to pick this issue up and submit a PR, please do and it can be re-opened.

jameslamb:
PR #5018 has been closed. Since there's no active work going on for this feature, I'm going to close this issue.

Anyone interested in this feature is welcome to contribute it! If you'd like to contribute this, comment here and the issue can be re-opened. Maintainers will be happy to help you.

StrikerRUS:
@jameslamb I see you
>self-assigned this on Jan 3

Are you still working on this feature request?

jameslamb:
Yes, thanks for the reminder, I'd like to get to this within the next release cycle. Not having it is blocking #6172. Let's leave it open please.

PR #4342, opened 2021-06-05
[R-package] remove unused argument in early stopping callback

jameslamb:
Noticed tonight that the function returned by `cb.early.stop()` has an argument `finalize = FALSE`.

That argument's value is never referenced, and no existing code expects to pass it. This PR proposes removing it.

The places where callbacks are executed only pass a single argument, `env`:

https://github.com/microsoft/LightGBM/blob/28c3c45d3be82f115630c4a887a0cf5e611ef5e0/R-package/R/lgb.train.R#L288-L290

https://github.com/microsoft/LightGBM/blob/28c3c45d3be82f115630c4a887a0cf5e611ef5e0/R-package/R/lgb.train.R#L329-L331

https://github.com/microsoft/LightGBM/blob/28c3c45d3be82f115630c4a887a0cf5e611ef5e0/R-package/R/lgb.cv.R#L367-L369

https://github.com/microsoft/LightGBM/blob/28c3c45d3be82f115630c4a887a0cf5e611ef5e0/R-package/R/lgb.cv.R#L393-L395

I'm also confident in this change because there are multiple unit tests checking that early stopping works correctly in the R package, for example:

https://github.com/microsoft/LightGBM/blob/28c3c45d3be82f115630c4a887a0cf5e611ef5e0/R-package/tests/testthat/test_basic.R#L576

### Notes for Reviewers

This is not a user-facing breaking change because all callbacks are internal to the package (since #2479 is still open).

### DIFF
--- a/R-package/R/callback.R
+++ b/R-package/R/callback.R
@@ -324,7 +324,7 @@ cb.early.stop <- function(stopping_rounds, first_metric_only = FALSE, verbose = 
   }
 
   # Create callback
-  callback <- function(env, finalize = FALSE) {
+  callback <- function(env) {
 
     # Check for empty evaluation
     if (is.null(eval_len)) {


StrikerRUS (approved):
LGTM!

PR #4343, opened 2021-06-05
[R-package] factor out lgb.check.r6.class()

jameslamb:
The R package has many places where it needs to check that an object is a special LightGBM R6 class (`lgb.Booster`, `lgb.Dataset`, or `lgb.Predictor`). Today, it is inconsistent in the way that it does this. Some code uses special functions like `lgb.is.Booster()` which call a one-line utility function `lgb.check.r6.class()`. Other code calls `lgb.check.r6.class()`.

This code proposes factoring out `lgb.check.r6.class()` completely. I think this has the following benefits:

* removes a layer of indirection / opportunity for inconsistency
* makes the code easier to read and understand

```r
# before
if (!lgb.check.r6.class(object = train_set, name = "lgb.Dataset"))

# after
if (!lgb.is.Dataset(train_set))
```

Since what `lgb.check.r6.class()` does is so simple, I think that the added indirection it introduces is not worth it, and that the R package's code will be a bit simpler as a result of this change.

### DIFF
--- a/R-package/R/lgb.Booster.R
+++ b/R-package/R/lgb.Booster.R
@@ -45,7 +45,7 @@ Booster <- R6::R6Class(
         # Check if training dataset is not null
         if (!is.null(train_set)) {
           # Check if training dataset is lgb.Dataset or not
-          if (!lgb.check.r6.class(object = train_set, name = "lgb.Dataset")) {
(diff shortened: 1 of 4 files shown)


StrikerRUS (approved):
LGTM!

PR #4344, opened 2021-06-05
[R-package] remove lgb.last_error() and LGBM_GetLastError_R()

jameslamb:
It seems that #4163 removed the last uses of the internal helper function `lgb.last_error()` in the R package. This PR proposes removing that function and its C++ counterpart, `LGBM_GetLastError_R()`. It's no longer necessary for the R package to reach into the C++ side and get the last error, since as of #4163 and other refactoring, C++ calls will raise R errors whenever one is thrown.

https://github.com/microsoft/LightGBM/blob/28c3c45d3be82f115630c4a887a0cf5e611ef5e0/R-package/src/lightgbm_R.cpp#L36-L38

Removing this code should have all the usual benefits of removing unused code...faster compilation, smaller library size, reduced maintenance burden, etc.

### DIFF
--- a/R-package/R/utils.R
+++ b/R-package/R/utils.R
@@ -15,16 +15,6 @@ lgb.is.null.handle <- function(x) {
   )
 }
 
-# [description] Get the most recent error stored on the C++ side and raise it
-#               as an R error.
-lgb.last_error <- function() {
-  err_msg <- .Call(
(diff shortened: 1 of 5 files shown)


StrikerRUS (approved):
LGTM! Nice to see one test that caused valgrind to fail is gone.

PR #4346, opened 2021-06-07
[docs] document how to pass multi-value params from Python and R fixes #4345)

### LINKED ISSUE #4345: Clarify how to specify monotonic constraints in docu
## Description

In https://lightgbm.readthedocs.io/en/latest/Parameters.html, it is unclear how to pass the `monotone_constraints` parameter.

> monotone_constraints 🔗︎, default = None, type = multi-int, aliases: mc, monotone_constraint
> 
> - used for constraints of monotonic features
> - 1 means increasing, -1 means decreasing, 0 means non-constraint
> - you need to specify all features in order. For example, mc=-1,0,1 means decreasing for 1st feature, non-constraint for 2nd feature and increasing for the 3rd feature

In R and Python, we can pass neither a "multi-int" like `mc=-1,0,1` nor a list, but rather a string like "-1,0,1". 

## Suggestion

Add to the docu a sentence like: "In R and Python, pass a string of the form '-1,0,1'." (If this is correct.)

jameslamb:
Fixed #4345 

The documentation for `monotone_constraints` does not currently describe the expected data format for that parameter in the Python and R packages.

Changes in this PR:

* adds documentation notes that `[-1, 0, 1]` is valid in the Python package and `c(-1, 0, 1)` is valid in the R package
* adds R tests based on these tests in the Python package (to be sure the documentation changes proposed here are correct)

https://github.com/microsoft/LightGBM/blob/20e633875f3c843e25f3877071c950e673b879b6/tests/python_package_test/test_engine.py#L1256

### Notes for Reviewers

This PR is still a work in progress. Will finish getting the tests working before asking for reviews.


docs/Parameters.rst
```diff
@@ -514,6 +514,12 @@ Learning Control Parameters
 
    -  you need to specify all features in order. For example, ``mc=-1,0,1`` means decreasing for 1st feature, non-constraint for 2nd feature and increasing for the 3rd feature
 
+   -  in the CLI or C++, use a string like ``"-1,0,1"``
+
+   -  in the Python package, can use either a string or a list like ``[-1, 0, 1]``
```
mayer79, line 519:
```suggestion
   -  in the Python package, can use either a string as in the CLI or a list like ``[-1, 0, 1]``
```

docs/Parameters.rst
```diff
@@ -514,6 +514,12 @@ Learning Control Parameters
...
+
+   -  in the R package, can use either a string or a vector like ``c(-1, 0, 1)``
```
mayer79, line 521:
```suggestion
   -  in the R package, can use either a string as in the CLI or a vector like ``c(-1, 0, 1)``
```

include/LightGBM/config.h
```diff
@@ -471,6 +471,9 @@ struct Config {
   // desc = used for constraints of monotonic features
   // desc = ``1`` means increasing, ``-1`` means decreasing, ``0`` means non-constraint
   // desc = you need to specify all features in order. For example, ``mc=-1,0,1`` means decreasing for 1st feature, non-constraint for 2nd feature and increasing for the 3rd feature
+  // desc = in the CLI or C++, use a string like ``"-1,0,1"``
+  // desc = in the Python package, can use either a string or a list like ``[-1, 0, 1]``
```
mayer79, line 475:
```suggestion
  // desc = in the Python package, can use either a string as in the CLI or a list like ``[-1, 0, 1]``
```

include/LightGBM/config.h
```diff
@@ -471,6 +471,9 @@ struct Config {
...
+  // desc = in the R package, can use either a string or a vector like ``c(-1, 0, 1)``
```
mayer79, line 476:
```suggestion
  // desc = in the R package, can use either a string as in the CLI or a vector like ``c(-1, 0, 1)``
```

docs/Parameters.rst
StrikerRUS, line 519: @jameslamb I'd like to keep this important doc in s consistent state. Changes proposed in this this PR are actually applicable for any `multi-***`-type parameter. We have a lot of such params, e.g. `max_bin_by_feature `, `cegb_penalty_feature_lazy`, `cegb_penalty_feature_coupled`, `categorical_feature`, `label_gain`, etc. I think it will be better to write a separate paragraph about how `multi-***` params can be passed to a program, if you think there should be some clarification for this.

Also, I'm against documenting internal string format for language wrappers. Actually, all params are passed via a string internally.
https://github.com/microsoft/LightGBM/pull/4101#discussion_r639267831
I don't think we should expose this and it's better to encourage users to use native language structures to pass params.

jameslamb: ok, I agree with both of these points. I'll split the unit tests part of this into a separate PR.

> if you think there should be some clarification for this.

I definitely do. If an expert LightGBM user like @mayer79 wasn't aware (https://github.com/microsoft/LightGBM/issues/4345#issuecomment-855659189) then I think many others will be unaware or will spend time trying to figure it out from unit tests / example code.

jameslamb: Ok, I've changed this to just document the concept generally. I think it's ready for review.

image

include/LightGBM/config.h
jameslamb: based on https://github.com/microsoft/LightGBM/pull/4346#discussion_r646507680, I think we're going to add general documentation on the fact that you can use a list in Python / vector in R, instead of specifically adding notes like this on each parameter. So I've reverted the changes to the specific `monotone_constraints` docs.

include/LightGBM/config.h
jameslamb: resolving, see https://github.com/microsoft/LightGBM/pull/4346#discussion_r647050589

docs/Parameters.rst
jameslamb: resolving, see https://github.com/microsoft/LightGBM/pull/4346#discussion_r647050589

docs/Parameters.rst
jameslamb: resolving, see https://github.com/microsoft/LightGBM/pull/4346#discussion_r647050589

docs/Parameters.rst
```diff
@@ -28,6 +28,26 @@ By using config files, one line can only contain one parameter. You can use ``#`
 
 If one parameter appears in both command line and config file, LightGBM will use the parameter from the command line.
 
+For the Python and R packages, any parameters that accept a list of values can be specified in those languages' default array types.
```
StrikerRUS, line 31: Do you think this info is worth to mention?
```suggestion
For the Python and R packages, any parameters that accept a list of values (usually they have ``multi-xxx`` type, e.g. ``multi-int`` or ``multi-double``) can be specified in those languages' default array types.
```

jameslamb: yes that's a nice addition! Thank you.

StrikerRUS (approved):
Thanks a lot, very clear description!

PR #4352, opened 2021-06-08
[ci] [R-package] add unit tests on monotone constraints

jameslamb:
Contributes to #4345, see https://github.com/microsoft/LightGBM/pull/4346#discussion_r647033503.

This PR proposes adding unit tests on monotone constraints to the R package, since there currently are not any.

This test is basically a translation of the corresponding Python test into R, but simplified by leaving out details about interaction constraints. The goal of this PR is to have at least basic coverage of monotone constraints from the R package.

https://github.com/microsoft/LightGBM/blob/bab58d0e90b2ee93d7e0656957e73e92d46554bc/tests/python_package_test/test_engine.py#L1256

### DIFF
--- a/R-package/tests/testthat/test_basic.R
+++ b/R-package/tests/testthat/test_basic.R
@@ -1746,7 +1746,6 @@ test_that("lgb.train() fit on linearly-relatead data improves when using linear 
 
 
 test_that("lgb.train() w/ linear learner fails already-constructed dataset with linear=false", {
-  testthat::skip("Skipping this test because it causes issues for valgrind")
   set.seed(708L)
   params <- list(
     objective = "regression"
--- a/R-package/tests/testthat/test_basic.R
+++ b/R-package/tests/testthat/test_basic.R
@@ -2129,11 +2129,17 @@ context("monotone constraints")
   return(any(diff(y) < 0.0) & any(diff(y) > 0.0))
 }
 
+# R equivalent of numpy.linspace()
+.linspace <- function(start_val, stop_val, num) {
+  weights <- (seq_len(num) - 1L) / (num - 1L)
+  return(start_val + weights * (stop_val - start_val))
+}
+
 .is_correctly_constrained <- function(learner, x3_to_categorical) {
   iterations <- 10L
   n <- 1000L
-  variable_x <- seq_len(n) / n
-  fixed_xs_values <- seq_len(n) / n
+  variable_x <- .linspace(0L, 1L, n)
+  fixed_xs_values <- .linspace(0L, 1L, n)
   for (i in seq_len(iterations)) {
     fixed_x <- fixed_xs_values[i] * rep(1.0, n)
     monotonically_increasing_x <- matrix(
@@ -2207,7 +2213,7 @@ for (x3_to_categorical in c(TRUE, FALSE)) {
         params = params
         , data = dtrain
         , obj = "regression_l2"
-        , nrounds = 10L
+        , nrounds = 100L
       )
       expect_true({
         .is_correctly_constrained(
--- a/R-package/tests/testthat/test_basic.R
+++ b/R-package/tests/testthat/test_basic.R
@@ -2042,3 +2042,147 @@ test_that(paste0("lgb.train() gives same results when using interaction_constrai
   expect_equal(pred1, pred2)
 
 })
+
+context("monotone constraints")
+
+.generate_trainset_for_monotone_constraints_tests <- function(x3_to_categorical) {
+  n_samples <- 3000L
+  x1_positively_correlated_with_y <- rnorm(n = n_samples)
+  x2_negatively_correlated_with_y <- rnorm(n = n_samples)
+  x3_negatively_correlated_with_y <- rnorm(n = n_samples)
+  if (x3_to_categorical) {
+    x3_negatively_correlated_with_y <- as.integer(abs(runif(n_samples) / 0.25))
+    categorical_features <- 3L
+  } else {
+    x3_negatively_correlated_with_y <- runif(n_samples)
+    categorical_features <- NULL
+  }
+  X <- data.matrix(
+    data.frame(
+      list(
+        x1_positively_correlated_with_y
+        , x2_negatively_correlated_with_y
+        , x3_negatively_correlated_with_y
+      )
+    )
+  )
+  zs <- rnorm(n = n_samples, mean = 0.0, sd = 0.01)
+  scales <- 10.0 * rnorm(6L + 0.5)
+  y <- (
+    scales[1L] * x1_positively_correlated_with_y
+    + sin(scales[2L] * pi * x1_positively_correlated_with_y)
+    - scales[3L] * x2_negatively_correlated_with_y
+    - cos(scales[4L] * pi * x2_negatively_correlated_with_y)
+    - scales[5L] * x3_negatively_correlated_with_y
+    - cos(scales[6L] * pi * x3_negatively_correlated_with_y)
+    + zs
+  )
+  return(lgb.Dataset(
+    data = X
+    , label = y
+    , categorical_feature = categorical_features
+    , free_raw_data = FALSE
+  ))
+}
+
+.is_increasing <- function(y) {
+  return(all(diff(y) >= 0.0))
+}
+
+.is_decreasing <- function(y) {
+  return(all(diff(y) <= 0.0))
+}
+
+.is_non_monotone <- function(y) {
+  return(any(diff(y) < 0.0) & any(diff(y) > 0.0))
+}
+
+.is_correctly_constrained <- function(learner, x3_to_categorical) {
+  iterations <- 10L
+  n <- 1000L
+  variable_x <- seq_len(n) / n
+  fixed_xs_values <- seq_len(n) / n
+  for (i in seq_len(iterations)) {
+    fixed_x <- fixed_xs_values[i] * rep(1.0, n)
+    monotonically_increasing_x <- matrix(
+      data = c(variable_x, fixed_x, fixed_x)
+      , ncol = 3L
+    )
+    monotonically_increasing_y <- predict(
+      learner
+      , monotonically_increasing_x
+    )
+
+    monotonically_decreasing_x <- matrix(
+      data = c(fixed_x, variable_x, fixed_x)
+      , ncol = 3L
+    )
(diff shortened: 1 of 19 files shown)


jameslamb:
@StrikerRUS I wanted to test the results of the `valgrind`  job when removing `testthat::skip()` calls from the R package's tests, to compare it to the results I'm seeing in #4353. I'm doing that here on this unrelated PR, with https://github.com/microsoft/LightGBM/pull/4352#issuecomment-857726568, because I couldn't get the optional checks working on my fork.

Could you please tell me what secrets I need to set up on my fork to be able to run the `triggering_comments` jobs there, like you did in https://github.com/microsoft/LightGBM/pull/3740#issue-551423557?

I tried setting a secret `SECRETS_WORKFLOW` in the settings on my fork to a GitHub personal access token and to the client token for a GitHub oauth app. In both cases, when I left a comment to trigger valgrind (https://github.com/jameslamb/LightGBM/pull/43) the jobs failed with this error:

> {
  "message": "Bad credentials",
  "documentation_url": "https://docs.github.com/rest"
}

Thank you!

jameslamb:
Ok I've reverted the temporary change from https://github.com/microsoft/LightGBM/pull/4352/commits/27147777e33e55da74978f156ba594b708860e0a, and this is now ready for review again.

StrikerRUS:
@jameslamb 

> Could you please tell me what secrets I need to set up on my fork

You need to set one secret named `WORKFLOW` with your [PAT](https://docs.github.com/en/github/authenticating-to-github/keeping-your-account-and-data-secure/creating-a-personal-access-token) with at least `repo:status` and `workflow` scopes (selecting `workflow` auto-chooses the whole `repo` section unfortunately).

image


After that don't forget to replace all `microsoft/LightGBM` occurrences with `jameslamb/LightGBM` in repo's code, e.g. here
https://github.com/microsoft/LightGBM/blob/24ac920879c82c5dcd1ff4ebc6379596c7ebf18e/.github/workflows/r_solaris.yml#L29
https://github.com/microsoft/LightGBM/blob/24ac920879c82c5dcd1ff4ebc6379596c7ebf18e/.ci/trigger_dispatch_run.sh#L51

jameslamb:
aha!!

> After that don't forget to replace all microsoft/LightGBM occurrences with jameslamb/LightGBM in repo's code, e.g. here

This was the part I was missing. I'll try that, thank you very much.

R-package/tests/testthat/test_basic.R
```diff
@@ -2042,3 +2042,147 @@ test_that(paste0("lgb.train() gives same results when using interaction_constrai
...
+.generate_trainset_for_monotone_constraints_tests <- function(x3_to_categorical) {
+  n_samples <- 3000L
+  x1_positively_correlated_with_y <- rnorm(n = n_samples)
+  x2_negatively_correlated_with_y <- rnorm(n = n_samples)
+  x3_negatively_correlated_with_y <- rnorm(n = n_samples)
+  if (x3_to_categorical) {
+    x3_negatively_correlated_with_y <- as.integer(abs(runif(n_samples) / 0.25))
+    categorical_features <- 3L
+  } else {
+    x3_negatively_correlated_with_y <- runif(n_samples)
```
StrikerRUS, line 2057: Why does this line needed?

R-package/tests/testthat/test_basic.R
```diff
@@ -2042,3 +2042,147 @@ test_that(paste0("lgb.train() gives same results when using interaction_constrai
...
+    data.frame(
+      list(
+        x1_positively_correlated_with_y
+        , x2_negatively_correlated_with_y
+        , x3_negatively_correlated_with_y
+      )
+    )
+  )
+  zs <- rnorm(n = n_samples, mean = 0.0, sd = 0.01)
+  scales <- 10.0 * rnorm(6L + 0.5)
```
StrikerRUS, line 2070: Different bounds compared to Python original test.

R-package/tests/testthat/test_basic.R
```diff
@@ -2042,3 +2042,147 @@ test_that(paste0("lgb.train() gives same results when using interaction_constrai
...
+      "lgb.train() supports monotone constraints ("
+      , "categoricals="
+      , x3_to_categorical
+      , ", method="
+      , monotone_constraints_method
+      , ")"
+    )
+    test_that(test_msg, {
+      set.seed(708L)
+      dtrain <- .generate_trainset_for_monotone_constraints_tests(
```
StrikerRUS, line 2164: `dtain` can be constructed only 3 times instead of current 6 if moved out of the inner loop.

R-package/tests/testthat/test_basic.R
```diff
@@ -2042,3 +2042,147 @@ test_that(paste0("lgb.train() gives same results when using interaction_constrai
...
+
+context("monotone constraints")
+
+.generate_trainset_for_monotone_constraints_tests <- function(x3_to_categorical) {
+  n_samples <- 3000L
+  x1_positively_correlated_with_y <- rnorm(n = n_samples)
+  x2_negatively_correlated_with_y <- rnorm(n = n_samples)
+  x3_negatively_correlated_with_y <- rnorm(n = n_samples)
+  if (x3_to_categorical) {
+    x3_negatively_correlated_with_y <- as.integer(abs(runif(n_samples) / 0.25))
```
StrikerRUS, line 2054: Different categorization compared to Python original test.

R-package/tests/testthat/test_basic.R
```diff
@@ -2042,3 +2042,147 @@ test_that(paste0("lgb.train() gives same results when using interaction_constrai
...
+    monotonically_decreasing_y <- predict(
+      learner
+      , monotonically_decreasing_x
+    )
+
+    if (x3_to_categorical) {
+      non_monotone_data <- c(
+        fixed_x
+        , fixed_x
+        , as.integer(abs(variable_x / 0.25))
```
StrikerRUS, line 2129: Different categorization compared to Python original test.

StrikerRUS (commented):
Generally LGTM, but I leaved some comments that highlight the difference with the original test. I think that it's OK to have not two identical tests for R and Python.

Also, I have one comment about how to speed up the test.

R-package/tests/testthat/test_basic.R
jameslamb: oh! It isn't, you're that I have duplicate definitions. The one a few lines up can be removed.

```r
x3_negatively_correlated_with_y <- rnorm(n = n_samples)
```

R-package/tests/testthat/test_basic.R
jameslamb: ah I see, my parentheses were wrong

https://github.com/microsoft/LightGBM/blob/53ffba7c5c4a4d64853ad6c5409735b986283d2d/tests/python_package_test/test_engine.py#L1240

R-package/tests/testthat/test_basic.R
jameslamb: fixed in https://github.com/microsoft/LightGBM/pull/4352/commits/e7b2048a6f8b9e42e5dda4b9012d8c9dcc06a9d7. I think that as of that commit, it should be the same.

Made the following changes

* use `runif(min =0, max = 1)` to match `np.random.random()` in Python
* used `as.integer(x / 0.01)` to match `np.digitize(bins=np.arange(0, 1, 0.01))`

R-package/tests/testthat/test_basic.R
jameslamb: updated, see https://github.com/microsoft/LightGBM/pull/4352#discussion_r650626970

R-package/tests/testthat/test_basic.R
jameslamb: nice, thanks! Made that change in https://github.com/microsoft/LightGBM/commit/e7b2048a6f8b9e42e5dda4b9012d8c9dcc06a9d7

R-package/tests/testthat/test_basic.R
jameslamb: fixed in https://github.com/microsoft/LightGBM/commit/e7b2048a6f8b9e42e5dda4b9012d8c9dcc06a9d7

R-package/tests/testthat/test_basic.R
jameslamb: fixed in https://github.com/microsoft/LightGBM/commit/e7b2048a6f8b9e42e5dda4b9012d8c9dcc06a9d7

R-package/tests/testthat/test_basic.R
StrikerRUS: Still different bounds. I guess you need `runif(n = 6L, min = 0.0, max = 1.0)` to mimic `np.random.random(6)` part.

R-package/tests/testthat/test_basic.R
```diff
@@ -2042,3 +2042,146 @@ test_that(paste0("lgb.train() gives same results when using interaction_constrai
...
+  X <- matrix(
+    data = c(
+        x1_positively_correlated_with_y
+        , x2_negatively_correlated_with_y
+        , x3_negatively_correlated_with_y
+    )
+    , ncol = 3L
+  )
+  zs <- rnorm(n = n_samples, mean = 0.0, sd = 0.01)
+  scales <- 10.0 * (rnorm(6L) + 0.5)
```
jameslamb, line 2068:
```suggestion
  scales <- 10.0 * (runif(n = 6L, min = 0.0, max = 1.0) + 0.5)
```

R-package/tests/testthat/test_basic.R
jameslamb: fixed in https://github.com/microsoft/LightGBM/pull/4352/commits/a15596a5b4d44f8c0ac9e8ab5069c73a28669924

jameslamb:
Tests started failing after https://github.com/microsoft/LightGBM/pull/4352/commits/a15596a5b4d44f8c0ac9e8ab5069c73a28669924, which exposed a few more places where these tests differed from the Python equivalents.

More changes in https://github.com/microsoft/LightGBM/pull/4352/commits/4950ed30c0330e585c17502aa176e66280c57b40 to make these tests consistent with the Python ones.

1. Replace `seq_len(n) / n` with code that produces an output identical to `numpy.linspace()`
2. train for 100 rounds, not 10

This is ready for another review.

StrikerRUS (approved):

PR #4354, opened 2021-06-08
[R-package] remove unnecessary library() calls in tests

jameslamb:
This PR proposes removing two unnecessary `library()` calls in the R package's tests. Using `{testthat}`, it's rarely necessary to call `library()` for dependencies from inside test scripts.

### Notes for Reviewers

I thought it was fun and interesting to note that the two calls removed in this PR have been there since the very first PR that introduced the R package!

#168 (January 2017)


StrikerRUS (approved):
LGTM since CI is green! 🙂

PR #4355, opened 2021-06-08
[python-package] use f-strings for concatenation in examples/python-guide/logistic_regression.py

sagnik1511:
Contributes to #4136.

Updated  `.\examples\python-guide\logistic_regression.py`  file

PR #4356, opened 2021-06-08
[python-package] use f-strings for concatenation in examples/python-guide/logistic_regression.py

sagnik1511:
This PR contributes to issue #4136 .
I've updated the file as I stated in https://github.com/microsoft/LightGBM/issues/4136#issuecomment-856897054.

In this PR I have updated the file `.\examples\python-guide\logistic_regression.py` where I have updated 4 lines of code with the f-string format.

I have previously updated 2 lines and updated two lines after your feedback @jameslamb.
Do let me know if there's an issue. Thank You.

### DIFF
--- a/examples/python-guide/logistic_regression.py
+++ b/examples/python-guide/logistic_regression.py
@@ -1,7 +1,9 @@
 # coding: utf-8
 """Comparison of `binary` and `xentropy` objectives.
+
 BLUF: The `xentropy` objective does logistic regression and generalizes
 to the case where labels are probabilistic (i.e. numbers between 0 and 1).
+
 Details: Both `binary` and `xentropy` minimize the log loss and use
 `boost_from_average = TRUE` by default. Possibly the only difference
 between them with default settings is that `binary` may achieve a slight
@@ -52,6 +54,7 @@ def log_loss(preds, labels):
 
 def experiment(objective, label_type, data):
     """Measure performance of an objective.
+    
     Parameters
     ----------
     objective : string 'binary' or 'xentropy'
@@ -60,6 +63,7 @@ def experiment(objective, label_type, data):
         Type of the label.
     data : dict
         Data for training.
+        
     Returns
     -------
     result : dict
--- a/examples/python-guide/logistic_regression.py
+++ b/examples/python-guide/logistic_regression.py
@@ -54,7 +54,7 @@ def log_loss(preds, labels):
 
 def experiment(objective, label_type, data):
     """Measure performance of an objective.
-    
+
     Parameters
     ----------
     objective : string 'binary' or 'xentropy'
@@ -63,7 +63,7 @@ def experiment(objective, label_type, data):
         Type of the label.
     data : dict
         Data for training.
-        
+
     Returns
     -------
     result : dict
--- a/examples/python-guide/logistic_regression.py
+++ b/examples/python-guide/logistic_regression.py
@@ -113,5 +113,5 @@ A = [experiment('binary', label_type='binary', data=DATA)['time']
      for k in range(K)]
 B = [experiment('xentropy', label_type='binary', data=DATA)['time']
      for k in range(K)]
-print('Best `binary` time: ' + str(min(A)))
-print('Best `xentropy` time: ' + str(min(B)))
+print(f"Best `binary` time: {str(min(A))}")
+print(f"Best `xentropy` time: {str(min(B))}")
--- a/examples/python-guide/logistic_regression.py
+++ b/examples/python-guide/logistic_regression.py
@@ -1,9 +1,7 @@
 # coding: utf-8
 """Comparison of `binary` and `xentropy` objectives.
-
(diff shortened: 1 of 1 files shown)


examples/python-guide/logistic_regression.py
```diff
@@ -1,9 +1,7 @@
 # coding: utf-8
 """Comparison of `binary` and `xentropy` objectives.
-
```
jameslamb, line 3: please revert all of these unrelated whitespace changes

examples/python-guide/logistic_regression.py
```diff
@@ -113,5 +109,5 @@ def experiment(objective, label_type, data):
      for k in range(K)]
 B = [experiment('xentropy', label_type='binary', data=DATA)['time']
      for k in range(K)]
-print('Best `binary` time: ' + str(min(A)))
-print('Best `xentropy` time: ' + str(min(B)))
+print(f"Best `binary` time: {str(min(A))}")
+print(f"Best `xentropy` time: {str(min(B))}")
```
jameslamb, line 113:
```suggestion
print(f"Best `binary` time: {min(A)}")
print(f"Best `xentropy` time: {min(B)}")
```

`str()` is no longer necessary inside an f-string, so I think this can be simplified.

jameslamb, line 113:
```suggestion
print(f"Best `binary` time: {min(A)}")
print(f"Best `xentropy` time: {min(B)}")
```

`str()` is no longer necessary inside an f-string, so I think this can be simplified.

jameslamb (changes requested):
Thanks! Please see my newest comments.

<hr>

A few other notes, since I think you might be new to creating pull requests on GitHub.

1. You do not need to (and should not) close a pull request and open a new one in response to review comments. Once you have opened a pull request from a branch, every commit you push to that branch will get added to the pull request for as long as the pull request is open.
2. I've changed the title for this pull request back to the one I originally had written in #4355. It is not necessary to include your username in the title, and we'd prefer that you did not. I think you may have been confused looking at the changelog link I included in https://github.com/microsoft/LightGBM/pull/4355#pullrequestreview-678755908...the username and PR number are added automatically to that changelog. I recommend that you look through some previous PRs in this project to get some ideas about how titles are usually written: https://github.com/microsoft/LightGBM/pulls?q=is%3Apr+is%3Aclosed
3. You should not use language like "fixes" or "resolves" an issue in your PR title or comments unless you are fully fixing it. I removed the word "resolves" from the PR description, because if it was left then GitHub would automatically close #4136 when this PR was merged. We don't want that in this case, since this is only a partial step toward that issue.

examples/python-guide/logistic_regression.py
jameslamb: I see that you've marked this "resolved", but these changes are still in the PR. You can see them on the "Files Changed" tab, at https://github.com/microsoft/LightGBM/pull/4356/files.

image

Please add blank lines back to these places where they were removed.

jameslamb (changes requested):

sagnik1511:
I have copied the whole file and changed that into this 
https://colab.research.google.com/drive/1Gnf41gU5c6ZP8rDajMexF4CWepr1D5vi?usp=sharing
Can you tell me if this is okay or not??

examples/python-guide/logistic_regression.py
sagnik1511: I don't know why that happened. I just copied the whole code and modified those 4 lines with f-string formation.

jameslamb: You may be using an editor that automatically removes such lines or something like that. Thanks for removing them.

jameslamb:
> I have copied the whole file and changed that into this
> https://colab.research.google.com/drive/1Gnf41gU5c6ZP8rDajMexF4CWepr1D5vi?usp=sharing
> Can you tell me if this is okay or not??

Sorry, I am not going to click that. It's security risk for us as maintainers to click links from people we don't know.

This project uses GitHub to manage code review, so please push any changes you'd like to make to this pull request and we'd be happy to review them here.

sagnik1511:
> > Okay, sir, I've updated the file in your required format. Should I create a new PR then??
> 
> I can see that you've pushed changes to `examples/python-guide/logistic_regression.py` to this pull request. So I'm confused, what other file are you referring to?

Sorry, Sir, I was confused a bit.
I think now the PR is in the required format. So, this can be merged, right?

sagnik1511:
> > So, this can be merged, right?
> 
> Maybe. A set of automated checks runs on every commit to this project. You can see them near the bottom of the PR.
> 
> image
> 
> At a minimum, all of these will need to pass before the PR is merged.

Okay, Sir. So now I should wait for the completion of those reviews, right? Or are there any changes left from my side?
And also sorry for bothering you this much :(

StrikerRUS (approved):
LGTM, thanks!

jameslamb (approved):
looks good, thank you very much for the help!

PR #4359, opened 2021-06-09
Updated tests/python_package_test/test_plotting.py to use f-strings

WestonKing-Leatham:
Contributes to #4136.


jameslamb (approved):
Thanks very much! I double-checked that file and don't see any other uses of concatenation that need to be changed to f-strings.

PR #4360, opened 2021-06-09
[R-package] avoid unnecessary computation of std deviations in lgb.cv()

jameslamb:
`lgb.cv()` in the R package performs k-fold cross-validation, and can optionally compute the standard deviation of evaluation metrics across all folds. This option is controlled by the argument `showsd` (default TRUE).

Today, even when `showsd = FALSE`, an internal function called by `lgb.cv()` still calculates error standard deviations and then `lgb.cv()` just ignores them.

https://github.com/microsoft/LightGBM/blob/9143003df6eb1a17b6e697159c3c870d5a5e7219/R-package/R/lgb.cv.R#L388-L390

This PR proposes a fix that avoids this unnecessary work, which should slightly speed up `lgb.cv()` for users who are not interested in the standard deviation of error estimates. It also introduces a test confirming that the output format of `lgb.cv()` changes based on the value of the `showsd` argument.


StrikerRUS (approved):
Nice catch! LGTM!

PR #4361, opened 2021-06-09
[R-package] remove defaults in internal functions

jameslamb:
The bug discovered in #4360 was missed for a long time because `lgb.cv()` relied on an internal function which set a default for a keyword argument. That default meant that it was possible for a bug of the form "an argument from `lgb.cv()` is not being passed through to this internal function" was possible.

This PR proposes removing other defaults in internal functions, to prevent similar issues in the future. I did not find any cases where the functions changed in this PR were being called without an explicit value for every keyword argument, so this is just defensive and should not change any current behavior for users.

### DIFF
--- a/R-package/R/callback.R
+++ b/R-package/R/callback.R
@@ -111,7 +111,7 @@ cb.reset.parameters <- function(new_params) {
 }
 
 # Format the evaluation metric string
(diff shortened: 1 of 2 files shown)


StrikerRUS (approved):
Thanks!

Issue #3949, opened 2021-02-12
[Question] Builtin metrics from python API


memeplex:
I would like to score my models using metrics provided by lightgbm instead of generic scores like `r2_score` or other provided by sklearn. The closer I got was calling `est.booster_.eval(Dataset(X, y), "arbitrary_name", fobj)` but this:

1. Throws `'Booster' object has no attribute 'train_set'`, at least when `est` was fit using the sklearn API.
2. According to the Python API docs `fobj` is a callable or `None`, so it doesn't seem to support named builtin metrics.
3. Minor quirk: forces me to use an arbitrary name.

Maybe I'm misunderstanding the usage of the method. Maybe `name` is intended to refer to one of the already set validation datasets, but then I don't understand what the first argument is for. And I don't see why the error.

Notice that I don't want to use metrics in fit, for early stopping or whatever, but to score already fit models, for example to call metrics from the `score` method during CV.

Is there any way to access those metrics from python?

StrikerRUS:
@memeplex Sorry for missing this issue!

Is this what you are looked for?

```
import lightgbm as lgb
from sklearn.datasets import load_boston

X, y = load_boston(return_X_y=True)
lgb_data = lgb.Dataset(X, y)
est = lgb.train({'objective': 'regression', 'metric': ['mape', 'rmse']},
                lgb_data, num_boost_round=10, keep_training_booster=True)

new_data = lgb.Dataset(X[:10, :], y[:10], reference=lgb_data)
est.eval(new_data, 'new_data')
```
```
>>> [('new_data', 'mape', 0.13074529807365748, False),
     ('new_data', 'rmse', 3.8224721524651835, False)]
```

PR #4364, opened 2021-06-09
[docs][python] enhance `keep_training_booster` param description

### LINKED ISSUE #3580: unexpected results for predicting on validation sets in python
If I understand correctly, Lightgbm makes it difficult to predict on a validation dataset (without access to raw say numpy data). There are functions `Booster.{eval,eval_train,eval_valid}`, but they are not correctly documented and are not user-friendly. 

incorrect doc: By default, `Booster.valid_sets` will be cleared at the end of training making it impossible to predict on training or validation sets using the above methods. The parameter `keep_training_booster` for `train` has to be set to `True` to keep the link. In the documentation for this parameter, there is no word about this effect. I can get around this by defining my own `Booster` and calling `Booster.update`, but this feels like a hack. 

not user-friendly: there is no straightforward way to get the actual predictions - I can `feval` scored on them in a direct way. Again, I can hack an `feval` that changes a global object to get around this or use `Booster._Booster__inner_predict`. 

Please correct me and accept my apologies if any of the statements above are incorrect. Is it a reasonable ask to keep the prediction possibility whenever possible and document releasing datasets accurately?

StrikerRUS:
Close #3580.
Refer to https://github.com/microsoft/LightGBM/issues/3949#issuecomment-808810679.

### DIFF
--- a/python-package/lightgbm/engine.py
+++ b/python-package/lightgbm/engine.py
@@ -128,6 +128,7 @@ def train(params, train_set, num_boost_round=100,
     keep_training_booster : bool, optional (default=False)
         Whether the returned Booster will be used to keep training.
         If False, the returned value will be converted into _InnerPredictor before returning.
+        This means you won't be able to use ``eval``, ``eval_train`` or ``eval_valid`` methods of the returned Booster.
         When your model is very large and cause the memory error,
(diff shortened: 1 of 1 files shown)


jameslamb (approved):
makes sense to me

PR #4365, opened 2021-06-09
[docs] document sanitizers

StrikerRUS:
Refer to https://xgboost.readthedocs.io/en/latest/contrib/unit_tests.html#sanitizers-detect-memory-errors-and-data-races, #3555, #4330.

### DIFF
--- a/docs/Installation-Guide.rst
+++ b/docs/Installation-Guide.rst
@@ -13,6 +13,21 @@ Users who want to perform benchmarking can make LightGBM output time costs for d
 
 It is possible to build LightGBM in debug mode. In this mode all compiler optimizations are disabled and LightGBM performs more checks internally. To enable debug mode you can add ``-DUSE_DEBUG=ON`` to CMake flags or choose ``Debug_*`` configuration (e.g. ``Debug_DLL``, ``Debug_mpi``) in Visual Studio depending on how you are building LightGBM.
 
+.. _sanitizers:
+
+In addition to the debug mode, LightGBM can be built with compiler sanitizers.
+To enable them add ``-DUSE_SANITIZER=ON -DENABLED_SANITIZERS="san_1;san_2;...;san_n"`` to CMake flags.
+``san_1`` ... ``san_n`` can refer to the following supported sanitizers:
+
+- ``address`` - AddressSanitizer (ASan);
+- ``leak`` - LeakSanitizer (LSan);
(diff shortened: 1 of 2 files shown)


docs/Installation-Guide.rst
```diff
@@ -13,6 +13,21 @@ Users who want to perform benchmarking can make LightGBM output time costs for d
 
 It is possible to build LightGBM in debug mode. In this mode all compiler optimizations are disabled and LightGBM performs more checks internally. To enable debug mode you can add ``-DUSE_DEBUG=ON`` to CMake flags or choose ``Debug_*`` configuration (e.g. ``Debug_DLL``, ``Debug_mpi``) in Visual Studio depending on how you are building LightGBM.
 
+.. _sanitizers:
+
+In addition to the debug mode, LightGBM can be built with compiler sanitizers.
+To enable them add ``-DUSE_SANITIZER=ON -DENABLED_SANITIZERS="san_1;san_2;...;san_n"`` to CMake flags.
+``san_1`` ... ``san_n`` can refer to the following supported sanitizers:
```
jameslamb, line 20: I was confused by this notation. I think it would be easier to use if we made this documentation copy-pastable (using the actual flag names).

```suggestion
To enable them add ``-DUSE_SANITIZER=ON -DENABLED_SANITIZERS="address;leak;undefined;thread"`` to CMake flags.
These values refer to the following supported sanitizers:
```

jameslamb (changes requested):
Thanks for this useful addition to the documentation! Please see my comments about making this copy-pastable.

StrikerRUS:
@jameslamb Thanks for your review!

> Please see my comments about making this copy-pastable.

I thought about it a lot and couldn't find a good variant. Unfortunately, your proposal is not fully copy-pastable as well due to that `thread` sanitizer is not compatible with others. Maybe you have any other variants?

jameslamb:
> Maybe you have any other variants?

I would prefer then to keep my suggestion but just not include `thread` in the list passed to `-DENABLED_SANITIZERS`

like

```text
To enable them add ``-DUSE_SANITIZER=ON -DENABLED_SANITIZERS="address;leak;undefined"`` to CMake flags.
```

StrikerRUS:
TBH, I don't like this variant very much. But because I'm out of ides for more general and clear wording, I'm going to accept your suggestion.

jameslamb (approved):

PR #4376, opened 2021-06-14
[python] add type hints to python-package/setup.py

greyhere:
Added type hints for python-package/setup.py as per #3756

### DIFF
--- a/python-package/setup.py
+++ b/python-package/setup.py
@@ -216,7 +216,7 @@ def compile_cpp(
 
 class CustomInstallLib(install_lib):
 
-    def install(self) -> None:
+    def install(self) -> List[str]:
         outfiles = install_lib.install(self)
         src = find_lib()[0]
         dst = os.path.join(self.install_dir, 'lightgbm')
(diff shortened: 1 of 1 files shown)


python-package/setup.py
```diff
@@ -204,7 +216,7 @@ def compile_cpp(use_mingw=False, use_gpu=False, use_cuda=False, use_mpi=False,
 
 class CustomInstallLib(install_lib):
 
-    def install(self):
+    def install(self) -> None:
```
jameslamb, line 219: This one returns a list of strings (see `return outfiles` a few lines below).

```suggestion
    def install(self) -> List[str]:
```

jameslamb (changes requested):
Thanks for taking this on! Please see my suggestions below.

greyhere:
> Thanks for taking this on! Please see my suggestions below.

Thanks! I applied your suggestion. Is there anything more to change?

python-package/setup.py
greyhere: Thanks for pointing out my mistake! I applied your suggestion. Is there any thing more to change?

jameslamb (approved):
These changes look good to me, thanks for the help!

StrikerRUS (approved):
LGTM!

PR #4377, opened 2021-06-14
[tests] replace pytest.parametrize

StrikerRUS:
Avoid unnecessary trainset generation.

### DIFF
--- a/tests/python_package_test/test_engine.py
+++ b/tests/python_package_test/test_engine.py
@@ -1252,8 +1252,8 @@ def generate_trainset_for_monotone_constraints_tests(x3_to_category=True):
     return trainset
 
 
-@pytest.mark.parametrize("test_with_interaction_constraints", [True, False])
-def test_monotone_constraints(test_with_interaction_constraints):
+@pytest.mark.parametrize("test_with_categorical_variable", [True, False])
+def test_monotone_constraints(test_with_categorical_variable):
     def is_increasing(y):
         return (np.diff(y) >= 0.0).all()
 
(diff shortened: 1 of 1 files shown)


tests/python_package_test/test_engine.py
```diff
@@ -1316,10 +1316,10 @@ def has_interaction(treef):
...
         return not has_interaction_flag.any()
 
-    for test_with_categorical_variable in [True, False]:
-        trainset = generate_trainset_for_monotone_constraints_tests(
-            test_with_categorical_variable
-        )
+    trainset = generate_trainset_for_monotone_constraints_tests(
+        test_with_categorical_variable
+    )
+    for test_with_interaction_constraints in [True, False]:
```
jameslamb, line 1322: I'm ok with this change, but could you update the assert on line 1334 with an error message that will tell us which case it failed on if it fails?

```python
error_msg = f"Model not correctly constrained (test_with_interaction_constraints={test_with_interaction_constraints})"

assert is_correctly_constrained(
    constrained_model, test_with_categorical_variable
), error_msg
```

Without that, that particular test could fail and we wouldn't know which case it failed on just from the logs.

I think it would also be useful too include that value of `monotone_constraints_method` in that message too (since it is similarly from an inner `for` loop), but that's not directly related to this PR so up to you if you want to do that at the same time or separately.

tests/python_package_test/test_engine.py
```diff
@@ -1399,8 +1399,9 @@ def test_monotone_penalty_max():
     }
 
     unconstrained_model = lgb.train(params_unconstrained_model, trainset_unconstrained_model, 10)
-    unconstrained_model_predictions = unconstrained_model.\
-        predict(x3_negatively_correlated_with_y.reshape(-1, 1))
+    unconstrained_model_predictions = unconstrained_model.predict(
+        x3_negatively_correlated_with_y.reshape(-1, 1)
+    )
```
jameslamb, line 1404: I like this formatting a lot more, thank you

jameslamb (changes requested):
nice change, thanks! Please see one suggestion I left about making it a bit easier to tell which tests failed from the logs.

tests/python_package_test/test_engine.py
StrikerRUS: Good point! Done in 44454821f1c5624fe5ed809f048e47adf47112e1.

jameslamb (approved):
looks great, thanks for speeding up the tests a bit!

PR #4380, opened 2021-06-15
[dask] Dask Vector types for group, init_score, sample_weights (fixes #4375)

### LINKED ISSUE #4375: Group, init_score, and sample_weight cannot be type Dask DataFrame
## Description
The `group`, `sample_weight`, and `init_score` Dask estimator parameters cannot be `Dask DataFrames`. They are currently described as "Dask Array, Dask DataFrame, Dask Series..." (e.g. [here](https://lightgbm.readthedocs.io/en/latest/pythonapi/lightgbm.DaskLGBMRanker.html#lightgbm.DaskLGBMRanker.fit)). Serial LightGBM estimators do not accept DataFrame values for these parameters - they should be array-like and not 2-D - which means that the distributed versions of these params also cannot be 2-D.

In `python-package/lightgbm/dask.py` these args' `typing` needs to reflect this in dask.py by adding a `_DaskAarrayLike` constant equal to `Union[dask_Array, dask_Series]` and then changing any required docstrings to remove the mention that they can be "Dask DataFrame" type.

## Reproducible example
```
import dask.dataframe as dd
from dask.distributed import Client, LocalCluster
import lightgbm as lgb
import numpy as np
import pandas as pd

X = pd.DataFrame(np.random.normal(size = (100, 5)))
y = pd.Series(np.random.choice([0, 1], size = 100))
init_score = pd.DataFrame(np.ones(100))

# -- get TypeError because init_score, group are 2-D.
clf = lgb.LGBMClassifier()
clf.fit(X, y)

g = pd.DataFrame([10] * 10, columns = ['group'])
rnk = lgb.LGBMRanker()
rnk.fit(X, y, group=g)

# -- similarly, in the distributed case...
client = Client(LocalCluster())

dX = dd.from_pandas(X, chunksize=10)
dy = dd.from_pandas(y, chunksize=10)
dinit_score = dd.from_pandas(pd.DataFrame(np.ones(100)), chunksize=10)

# -- still TypeError because distributed init_score is 2-D.
clf = lgb.DaskLGBMClassifier()
clf.fit(dX, dy, init_score=dinit_score)
```

## Environment info

LightGBM version or commit hash: 3.2.0

```shell
pip install lightgbm
```

## Additional Comments
First raised here: https://github.com/microsoft/LightGBM/pull/4101#discussion_r650628028

ffineis:
Addresses #4375 - removes `Dask DataFrame` type option for DaskLGBM estimators' `group`, `init_score`, and `sample_weights` parameters. These must be 1-d distributed vectors, cannot be 2-d.

To be honest I was kind of torn grammatically between "`Dask Array or Dask Series or None, optional (default=None)`" and "`Dask Array, Dask Series, or None, optional (default=None)`". Welcome an opinion on this!

### DIFF
--- a/python-package/lightgbm/dask.py
+++ b/python-package/lightgbm/dask.py
@@ -22,6 +22,7 @@ from .sklearn import LGBMClassifier, LGBMModel, LGBMRanker, LGBMRegressor, _lgbm
 
 _DaskCollection = Union[dask_Array, dask_DataFrame, dask_Series]
 _DaskMatrixLike = Union[dask_Array, dask_DataFrame]
+_DaskVectorLike = Union[dask_Array, dask_Series]
 _DaskPart = Union[np.ndarray, pd_DataFrame, pd_Series, ss.spmatrix]
 _PredictionDtype = Union[Type[np.float32], Type[np.float64], Type[np.int32], Type[np.int64]]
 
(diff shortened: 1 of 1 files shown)


jameslamb (approved):
I like the phrasing you went with. Thanks very much for doing this and for the excellent write-up in #4375!

PR #4381, opened 2021-06-15
[R-package] consolidate duplicate lists of Dataset info keys

jameslamb:
As part of adding a deprecation warning about the removal of `...` support in the R package (per https://github.com/microsoft/LightGBM/issues/4226#issuecomment-829473584), I wanted to be able to give different warnings in items passed through `...` in `lgb.Dataset()` based on whether they are params (e.g. `max_bin`) or "info" elements (e.g. `init_score`).

I realized that there are three duplicate places where the R package stores the list of valid keys for "info".

This PR proposes consolidating them into a single place, to reduce the risk of inconsistencies and simplify the code a bit.


StrikerRUS (approved):
Like it!

PR #3948, opened 2021-02-12
Run tests and build Python wheels for aarch64 architecture

StrikerRUS:
Contributes to #1129.

Fixed #3456.
Fixed #3517.
(Superceded) closed #3421.

I run 10x10 consecutive runs and all of them were successful, so I'm quite sure in stability of new QEMU job.

image

It can be seen from the picture above that sometimes QEMU job takes ~30min and sometimes ~1h. I thought that maybe there is any bottlenecking test that freezes whole job. But unfortunately, long runs are related to QEMU itself (it is common practice to set `timeout=360` for QEMU builds) because in those jobs all tests contribute to slowness. Here are timings from `pytest`.

#### ~30min job:

```
2021-02-12T14:26:32.5928137Z ============================== slowest durations ===============================
2021-02-12T14:26:32.5928779Z 555.37s call     tests/python_package_test/test_engine.py::test_int32_max_sparse_contribs
2021-02-12T14:26:32.5929439Z 76.17s call     tests/python_package_test/test_consistency.py::test_multiclass
2021-02-12T14:26:32.5930097Z 37.81s call     tests/python_package_test/test_engine.py::test_predict_with_start_iteration
2021-02-12T14:26:32.5930761Z 22.77s call     tests/python_package_test/test_consistency.py::test_binary_linear
2021-02-12T14:26:32.5931383Z 19.36s call     tests/python_package_test/test_engine.py::test_model_size
2021-02-12T14:26:32.5932015Z 18.50s call     tests/python_package_test/test_consistency.py::test_binary
2021-02-12T14:26:32.5932641Z 17.16s call     tests/python_package_test/test_engine.py::test_metrics
2021-02-12T14:26:32.5933266Z 13.51s call     tests/python_package_test/test_consistency.py::test_lambdarank
2021-02-12T14:26:32.5933901Z 12.67s call     tests/python_package_test/test_consistency.py::test_xendcg
2021-02-12T14:26:32.5934537Z 12.45s call     tests/python_package_test/test_consistency.py::test_regression
2021-02-12T14:26:32.5935161Z 12.15s call     tests/python_package_test/test_engine.py::test_cv
2021-02-12T14:26:32.5936002Z 10.82s call     tests/python_package_test/test_engine.py::test_monotone_constraints
2021-02-12T14:26:32.5936621Z 8.56s call     tests/python_package_test/test_engine.py::test_auc_mu
[...]

PR #4385, opened 2021-06-16
[ci] Increase timeout value for QEMU builds

StrikerRUS:
Seem that previous bump (#4326) is not enough.

> ##[error]The job running on agent Hosted Agent ran longer than the maximum time of 150 minutes.
https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=10344&view=logs&j=c2f9361f-3c13-57db-3206-cee89820d5e3

### DIFF
--- a/.vsts-ci.yml
+++ b/.vsts-ci.yml
@@ -150,7 +150,7 @@ jobs:
     PRODUCES_ARTIFACTS: 'true'
   pool:
     vmImage: ubuntu-latest
-  timeoutInMinutes: 150
+  timeoutInMinutes: 180
   strategy:
     matrix:
       bdist:


jameslamb (approved):
wow 3 hours :grimacing: 

Ok I still think that won't be a problem at the current pace of development in this repo. If it becomes a problem maybe we should consider moving some of these to optional jobs triggered by comments in the future.

But for now, approved!

StrikerRUS:
> wow 3 hours 😬

Agree... 🙁 

Typical pattern is simply set max allowed value
> timeoutInMinutes: 360
    https://github.com/search?l=YAML&p=3&q=qemu+timeoutInMinutes&type=Code

> But unfortunately, long runs are related to QEMU itself (it is common practice to set timeout=360 for QEMU builds) because in those jobs all tests contribute to slowness.
    https://github.com/microsoft/LightGBM/pull/3948#issue-572704305


> If it becomes a problem maybe we should consider moving some of these to optional jobs triggered by comments in the future.

... or just skip some tests there.

PR #4386, opened 2021-06-17
[python] Improving the syntax of the fstrings in the file: .\examples\python-guide\advanced_example.py

sayantan1410:
I have changed a string that had + concatenation to fstring format. 
Contributing to the issue #4136

### DIFF
--- a/examples/python-guide/advanced_example.py
+++ b/examples/python-guide/advanced_example.py
@@ -193,6 +193,7 @@ print('Finished 50 - 60 rounds with self-defined objective function and multiple
 
 print('Starting a new training job...')
 
+
 # callback
 def reset_metrics():
     def callback(env):
--- a/examples/python-guide/advanced_example.py
+++ b/examples/python-guide/advanced_example.py
@@ -56,7 +56,7 @@ gbm = lgb.train(params,
 
 print('Finished first 10 rounds...')
 # check feature name
-print('7th feature name is:', lgb_train.feature_name[6])
+print(f'7th feature name is: {lgb_train.feature_name[6]}')
 
 print('Saving model...')
 # save model to file
@@ -70,10 +70,10 @@ with open('model.json', 'w+') as f:
     json.dump(model_json, f, indent=4)
 
 # feature names
-print('Feature names:', gbm.feature_name())
+print(f'Feature names: {gbm.feature_name()}')
 
 # feature importances
-print('Feature importances:', list(gbm.feature_importance()))
+print(f'Feature importances: {list(gbm.feature_importance())}')
 
 print('Loading model to predict...')
 # load model to predict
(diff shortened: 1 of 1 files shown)


jameslamb (changes requested):
Thanks! I think that there are some more places in this file that could be changed. Can you please update them?

https://github.com/microsoft/LightGBM/blob/4167649cd50035b8237d121f706dc9f1e1b79ed0/examples/python-guide/advanced_example.py#L59

https://github.com/microsoft/LightGBM/blob/4167649cd50035b8237d121f706dc9f1e1b79ed0/examples/python-guide/advanced_example.py#L73

https://github.com/microsoft/LightGBM/blob/4167649cd50035b8237d121f706dc9f1e1b79ed0/examples/python-guide/advanced_example.py#L76

https://github.com/microsoft/LightGBM/blob/4167649cd50035b8237d121f706dc9f1e1b79ed0/examples/python-guide/advanced_example.py#L84

https://github.com/microsoft/LightGBM/blob/4167649cd50035b8237d121f706dc9f1e1b79ed0/examples/python-guide/advanced_example.py#L96

This one doesn't need to be an f-string, but could be combined into one line (since this project doesn't enforce a line length in Python code and since the full content of this one isn't THAT long): https://github.com/microsoft/LightGBM/blob/4167649cd50035b8237d121f706dc9f1e1b79ed0/examples/python-guide/advanced_example.py#L190-L191

For the examples with `mean_squared_error(y_test, y_pred) ** 0.5`, please store the result of that computation in a variable called `mse` and then reference `mse` in the f-string.

examples/python-guide/advanced_example.py
```diff
@@ -187,12 +189,10 @@ def accuracy(preds, train_data):
                 feval=[binary_error, accuracy],
                 valid_sets=lgb_eval)
 
-print('Finished 50 - 60 rounds with self-defined objective function '
-      'and multiple self-defined eval metrics...')
+print('Finished 50 - 60 rounds with self-defined objective function and multiple self-defined eval metrics...')
 
 print('Starting a new training job...')
 
-
```
jameslamb, line 195: Please revert this whitespace change by adding back a blank line.

jameslamb (changes requested):

examples/python-guide/advanced_example.py
sayantan1410: Oh I see, that's why it was not passing one check. 
Thank you for mentioning I will surely do this.

examples/python-guide/advanced_example.py
```diff
@@ -70,18 +70,19 @@
...
 
 print('Loading model to predict...')
 # load model to predict
 bst = lgb.Booster(model_file='model.txt')
 # can only predict with the best iteration (or the saving iteration)
 y_pred = bst.predict(X_test)
 # eval with loaded model
-print("The rmse of loaded model's prediction is:", mean_squared_error(y_test, y_pred) ** 0.5)
+rmse_loaded_model = mean_squared_error(y_test, y_pred) ** 0.5
+print(f"The rmse of loaded model's prediction is: {rmse_loaded_model}")
```
StrikerRUS, line 85:
```suggestion
print(f"The RMSE of loaded model's prediction is: {rmse_loaded_model}")
```

StrikerRUS (approved):
Thanks very much!
Just one minor suggestion about metric name.

jameslamb (approved):
thanks very much!

jameslamb:
closing and re-opening this to re-trigger CI. It seems Azure got stuck.

sayantan1410:
Now it seems, that all checks have been passed. 
Let me know if everything is alright otherwise I will make another PR for this one and will close it. 
@jameslamb

sayantan1410:
Thanks for merging it !!

PR #4387, opened 2021-06-17
fix param aliases

StrikerRUS:
- add `precise_float_parser` to Dataset params at language wrappers level and raise error in case of new value in config for constructed Dataset
- move `linear_tree` param to Dataset section (**open to discussion** about another way how indicate the fact that `linear_tree` config value is used at Dataset creation step)

### DIFF
--- a/R-package/R/aliases.R
+++ b/R-package/R/aliases.R
@@ -4,7 +4,7 @@
 # [description] List of respected parameter aliases specific to lgb.Dataset. Wrapped in a function to
 #               take advantage of lazy evaluation (so it doesn't matter what order
 #               R sources files during installation).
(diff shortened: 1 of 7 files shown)


jameslamb (approved):
makes sense to me, thanks

StrikerRUS:
@btrotta What do you think about moving `linear_tree` param into `Dataset Parameters` section?

chivee (approved):

btrotta:
> @btrotta What do you think about moving `linear_tree` param into `Dataset Parameters` section?

@StrikerRUS this seems fine to me, since choosing `linear_tree` changes the how the dataset is constructed and stored.

PR #4389, opened 2021-06-18
fix: typo in python class _InnerPredictor docstring

cyfdecyf:
Fix typo.


jameslamb (approved):
Thank you very much for the help!

PR #4395, opened 2021-06-20
[docs][R-package] add authors in R-package description

StrikerRUS:
As per suggestion https://github.com/microsoft/LightGBM/pull/4291#issuecomment-841701659.

### DIFF
--- a/R-package/DESCRIPTION
+++ b/R-package/DESCRIPTION
@@ -24,7 +24,7 @@ Authors@R: c(
     person("Alberto", "Ferreira", role = c("ctb")),
     person("Daniel", "Lemire", role = c("ctb")),
     person("Victor", "Zverovich", role = c("cph")),
-    person("IBM Corporation", role = c("ctb")),
+    person("IBM Corporation", role = c("ctb"))
     )
 Description: Tree based algorithms can be improved by introducing boosting frameworks. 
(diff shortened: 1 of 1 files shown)


R-package/DESCRIPTION
```diff
@@ -23,7 +24,7 @@ Authors@R: c(
     person("Alberto", "Ferreira", role = c("ctb")),
     person("Daniel", "Lemire", role = c("ctb")),
     person("Victor", "Zverovich", role = c("cph")),
-    person("IBM Corporation", role = c("ctb"))
+    person("IBM Corporation", role = c("ctb")),
```
jameslamb, line 27:
```suggestion
    person("IBM Corporation", role = c("ctb"))
```

Trailing commas are not valid in R code the way they are in Python. This is why R jobs are failing for this PR with an error like this:

```text
* checking DESCRIPTION meta-information ... ERROR
Malformed Authors@R field:
  argument 22 is empty
```

for example, https://github.com/microsoft/LightGBM/pull/4395/checks?check_run_id=2870678011

jameslamb (approved):
Thanks very much 😀

R-package/DESCRIPTION
StrikerRUS: Many thanks!

> Trailing commas are not valid in R code the way they are in Python.

Ah, sad 🙁

PR #4396, opened 2021-06-21
[python] Improving the syntax of `print`s in `simple_example.py` and `sklearn_example.py`

StrikerRUS:
Based on comments https://github.com/microsoft/LightGBM/pull/4386#pullrequestreview-686376988 and https://github.com/microsoft/LightGBM/pull/4386#discussion_r655295441.

These files are already marked as done in #4136 because there were no old string formatting methods there, only `print`s.


jameslamb (approved):

jameslamb:
Thanks very much!

PR #4391, opened 2021-06-18
[R-package] fix protection stack imbalance and unprotected objects (fixes #4390)

fabsig:
Fixes  #4390.

PR #4397, opened 2021-06-22
[ci] [R-package] increase timeout on valgrind job

jameslamb:
See https://github.com/microsoft/LightGBM/pull/4391#issuecomment-864273568 and this failed job with a timeout of 180 minutes (https://github.com/microsoft/LightGBM/actions/runs/959047226).

It seems recently-added R package tests (or maybe some combination of that and changes in the C++ side) has led to the valgrind jobs exceeding 120 minutes. As a reminder, R runs substantially slower under valgrind (https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Using-valgrind).

### DIFF
--- a/.github/workflows/r_valgrind.yml
+++ b/.github/workflows/r_valgrind.yml
@@ -7,7 +7,7 @@ on:
 jobs:
   test-r-valgrind:
     name: r-package (ubuntu-latest, R-devel, valgrind)
-    timeout-minutes: 300
+    timeout-minutes: 180
     runs-on: ubuntu-latest
     container: wch1/r-debug
     env:
--- a/.github/workflows/r_valgrind.yml
+++ b/.github/workflows/r_valgrind.yml
@@ -7,7 +7,7 @@ on:
 jobs:
   test-r-valgrind:
     name: r-package (ubuntu-latest, R-devel, valgrind)
-    timeout-minutes: 120
+    timeout-minutes: 300
     runs-on: ubuntu-latest
     container: wch1/r-debug
     env:


jameslamb:
This job is failing with an error message that suggests that the job config is being sourced from `master`, not this branch.

> The job running on runner GitHub Actions 125 has exceeded the maximum execution time of 120 minutes.

https://github.com/microsoft/LightGBM/actions/runs/959283121

(notice that it says "120 minutes")

@StrikerRUS , does that mean this needs to be merged to be tested?

StrikerRUS:
@jameslamb 
> does that mean this needs to be merged to be tested?

Yes, you're right.

https://github.com/microsoft/LightGBM/blob/47208894d819f1b8cef193e71f78e9a8c2f38462/.github/workflows/r_valgrind.yml#L3-L5
With `repository_dispatch` event trigger, workflow config is taken from the default branch (`master` in our case).
https://docs.github.com/en/actions/reference/events-that-trigger-workflows#repository_dispatch

.github/workflows/r_valgrind.yml
```diff
@@ -7,7 +7,7 @@ on:
 jobs:
   test-r-valgrind:
     name: r-package (ubuntu-latest, R-devel, valgrind)
-    timeout-minutes: 120
+    timeout-minutes: 300
```
StrikerRUS, line 10: Is `300` really needed? Maybe try with `180` for the first time?

jameslamb: ok 180 is fine

```suggestion
    timeout-minutes: 180
```

StrikerRUS (approved):
Thank you.

StrikerRUS:
Do you have rights to merge this PR with failing test?

jameslamb:
It seems I don't have sufficient permissions to merge when a check is failing. Notice "squash and merge" is grayed out.

image

So I guess I'll have to open a new PR with these changes.

PR #4303, opened 2021-05-19
[docs] replace broken mmlspark notebook link in docs

jameslamb:
The most recent `link checks` job failed.

```text
URL        `https://github.com/Azure/mmlspark/blob/master/notebooks/samples/LightGBM%20-%20Quantile%20Regression%20for%20Drug%20Discovery.ipynb'
Name       `this MMLSpark example'
Parent URL file:///home/runner/work/LightGBM/LightGBM/docs/_build/html/Parallel-Learning-Guide.html, line 273, col 8
Real URL   https://github.com/Azure/mmlspark/blob/master/notebooks/samples/LightGBM%20-%20Quantile%20Regression%20for%20Drug%20Discovery.ipynb
Check time 0.464 seconds
Result     Error: 404 Not Found
```

reference: https://github.com/microsoft/LightGBM/runs/2618067011?check_suite_focus=true

Seems that `mmlspark` removed their LightGBM Quantile Regression example yesterday. See https://github.com/Azure/mmlspark/pull/1043.

This PR proposes updating the documentation to point to a different LightGBM `mmlspark` notebook.

PR #4401, opened 2021-06-23
[docs] update link to LightGBM example in MMLSpark repo

StrikerRUS:


jameslamb (approved):
😫 it was only a month ago that we had to update this link (#4303). Seems they did another re-organization in that repo: https://github.com/Azure/mmlspark/pull/1073

I hope they stop moving files around soon 😆

PR #4408, opened 2021-06-26
[docs] Add shapash to the list of related projects

StrikerRUS:
> Compatible with most of sklearn, **lightgbm**, catboost, xgboost models
    https://github.com/MAIF/shapash/blob/master/docs/index.rst#shapash-features


jameslamb (approved):
nice, thanks!

PR #4410, opened 2021-06-26
[python] replace numpy.zeros with numpy.empty for the speedup

StrikerRUS:
> `empty`, unlike `zeros`, does not set the array values to zero, and may therefore be marginally faster. On the other hand, it requires the user to manually set all the values in the array, and should be used with caution.
    https://numpy.org/doc/stable/reference/generated/numpy.empty.html


jameslamb (approved):
Makes sense to me!

I checked the blame and `np.empty()` has been a part of `numpy` for 15 years, so I don't think this introduces any compatibility problems: https://github.com/numpy/numpy/blame/623bc1fae1d47df24e7f1e29321d0c0ba2771ce0/numpy/matlib.py

Issue #3326, opened 2020-08-20
Prediction for the Learn-to-Rank Application


A4Ayou:
Hello,

I developed a learning-to-rank (LTR) model by setting the objective to `lambdarank`, and training the model using the `.fit()` function with an LGB Data Set where I passed (1) the features data, (2) the label data, and (3) the `group `of the query. 

To predict the results, I am using the `.predict()` function. But this function only takes as input the array of features, and do not consider the argument `group` to indicate the query. 

My question is then: **when using prediction for an LTR model, should we pass data for one query at a time in the .predict() function, or can we directly pass a set of different queries?**

For example, let's say I want to predict the ranking from 2 distinct queries A and B. I get my features for the documents in A `X(A)` and for the documents in B `X(B)`. Should I run` .predict() `separately on `X(A)` and `X(B)` or can I just concatenate both `X(A)` and `X(B) `into a single numpy object `X `and run `X`? 

My intuition would be to go with the first option, but it may also not be the most convenient. For my applications, I am dealing with thousands of queries. Does that mean that I have to run the prediction thousands of time to get the results? 

This is a follow up on the issue #1398 which was closed (https://github.com/Microsoft/LightGBM/issues/1398)

Please let me know if something is unclear

Thanks!

Have a nice day!

guolinke:
Group is not needed in prediction time. The prediction results are pointwise, and you can rank them by yourself.

A4Ayou:
> Group is not needed in prediction time. The prediction results are pointwise, and you can rank them by yourself.

Thank you for the quick answer ! 

Follow-up question (just by curiosity): would it be more efficient to run the prediction with the pairwise approach (as it is done for training)? Or is it not feasible?

guolinke:
I think the current solution (listwise) is more efficient than pairwise.

A4Ayou:
But the lambdarank implemented in LightGBM training module is pairwise and not listwise, correct? 

And independent from this question, why are we using pointwise approach instead of pairwise/listwise for prediction?

guolinke:
@A4Ayou 
it is listwise, as all pairs in one list are used in training. And then, based on the scores of all pairs, the pointwise score is aggregated. In short, the lambdarank will calculate the score listwise, then reduce it pointwise. and the tree is learned to aggregated pointwise score. Therefore, it does not need to use listwise in prediction.

A4Ayou:
> @A4Ayou
[...]

PR #4411, opened 2021-06-26
[docs] clarify description of prediction early stopping

StrikerRUS:
https://github.com/microsoft/LightGBM/blob/f83180883a51729af9562ccac7b450c6da79a04a/src/application/predictor.hpp#L43-L58

https://github.com/microsoft/LightGBM/blob/d517ba12f2e7862ac533908304dddbd770655d2b/include/LightGBM/boosting.h#L282-L283

https://github.com/microsoft/LightGBM/blob/8116d880f79967d339783e5e9597351639232015/include/LightGBM/objective_function.h#L61-L62

### DIFF
--- a/docs/Parameters.rst
+++ b/docs/Parameters.rst
@@ -909,7 +909,11 @@ Predict Parameters
 
    -  used only in ``prediction`` task
 
+   -  used only in ``classification`` and ``ranking`` applications
+
    -  if ``true``, will use early-stopping to speed up the prediction. May affect the accuracy
+
+   -  **Note**: cannot be used with ``rf`` boosting type or custom objective function
 
 -  ``pred_early_stop_freq`` :raw-html:`<a id="pred_early_stop_freq" title="Permalink to this parameter" href="#pred_early_stop_freq">&#x1F517;&#xFE0E;</a>`, default = ``10``, type = int
 
--- a/include/LightGBM/config.h
+++ b/include/LightGBM/config.h
@@ -767,7 +767,9 @@ struct Config {
 
   // [no-save]
   // desc = used only in ``prediction`` task
+  // desc = used only in ``classification`` and ``ranking`` applications
   // desc = if ``true``, will use early-stopping to speed up the prediction. May affect the accuracy
+  // desc = **Note**: cannot be used with ``rf`` boosting type or custom objective function
   bool pred_early_stop = false;
 
   // [no-save]


docs/Parameters.rst
```diff
@@ -909,8 +909,12 @@ Predict Parameters
 
    -  used only in ``prediction`` task
 
+   -  used only in ``classification`` and ``ranking`` applications
```
StrikerRUS, line 912: ### Classification
https://github.com/microsoft/LightGBM/blob/0701a32da9fae7b2de8a01d702fa7d6abf36e836/src/objective/binary_objective.hpp#L188
https://github.com/microsoft/LightGBM/blob/d517ba12f2e7862ac533908304dddbd770655d2b/src/objective/multiclass_objective.hpp#L153
https://github.com/microsoft/LightGBM/blob/d517ba12f2e7862ac533908304dddbd770655d2b/src/objective/multiclass_objective.hpp#L259

### Ranking
https://github.com/microsoft/LightGBM/blob/8116d880f79967d339783e5e9597351639232015/src/objective/rank_objective.hpp#L80

docs/Parameters.rst
```diff
@@ -909,8 +909,12 @@ Predict Parameters
...
+
    -  if ``true``, will use early-stopping to speed up the prediction. May affect the accuracy
 
+   -  **Note**: cannot be used with ``rf`` boosting type or custom objective function
```
StrikerRUS, line 916: https://github.com/microsoft/LightGBM/blob/8116d880f79967d339783e5e9597351639232015/src/boosting/rf.hpp#L205-L208
https://github.com/microsoft/LightGBM/blob/a8ee487aca35363fafa027e6b7695976045096b3/src/boosting/gbdt.h#L159-L169

StrikerRUS:
Does prediction early stopping really work with ranking applications?

jameslamb (approved):
Looks good to me based on the links provided, thanks for doing that!

I don't know the answer to https://github.com/microsoft/LightGBM/pull/4411#issuecomment-869059869 though.

StrikerRUS:
@cbecker @guolinke Could you please help with this question https://github.com/microsoft/LightGBM/pull/4411#issuecomment-869059869?

cbecker:
@StrikerRUS It depends on what the scores look for ranking.
For multi-class, each instance is given N scores [c1, c2, c3 c4, .. cn], and we compute the margin (see `prediction_early_stop.cpp`, `CreateMulticlass()`). The idea is that when a class is further than all other by a large margin (score difference), then we can stop.

I see that for ranking LightGBM implements regression. In this case, what is the prediction for each sample?
If it is a relative 'rank' to other values as a difference, as in multi-class, then one could apply the multi-class early stopping computation above (though I never tried that in practice)

StrikerRUS:
@cbecker Thank you so much for quick response!
LightGBM's prediction values in ranking are pointwise (https://github.com/microsoft/LightGBM/issues/3326#issuecomment-677957226).

Issue #4416, opened 2021-06-28
Migrate to `pathlib` for the Python code


StrikerRUS:
Refer to https://github.com/microsoft/LightGBM/pull/4254#issuecomment-869163563 for the rationale.

- [x] ``.\.nuget\create_nuget.py``
- [x] ``.\docs\conf.py``
- [x] ``.\examples\python-guide\advanced_example.py``
- [x] ``.\examples\python-guide\logistic_regression.py``
- [x] ``.\examples\python-guide\plot_example.py``
- [x] ``.\examples\python-guide\simple_example.py``
- [x] ``.\examples\python-guide\sklearn_example.py``
- [x] ``.\examples\python-guide\dask\binary-classification.py``
- [x] ``.\examples\python-guide\dask\multiclass-classification.py``
- [x] ``.\examples\python-guide\dask\prediction.py``
- [x] ``.\examples\python-guide\dask\ranking.py``
- [x] ``.\examples\python-guide\dask\regression.py``
- [x] ``.\helpers\check_dynamic_dependencies.py``
- [x] ``.\helpers\parameter_generator.py``
- [x] ``.\python-package\setup.py``
- [x] ``.\python-package\lightgbm\basic.py``
- [x] ``.\python-package\lightgbm\callback.py``
- [x] ``.\python-package\lightgbm\compat.py``
- [x] ``.\python-package\lightgbm\dask.py``
- [x] ``.\python-package\lightgbm\engine.py``
- [x] ``.\python-package\lightgbm\libpath.py``
- [x] ``.\python-package\lightgbm\plotting.py``
- [x] ``.\python-package\lightgbm\sklearn.py``
- [x] ``.\python-package\lightgbm\__init__.py``
- [x] ``.\tests\cpp_test\test.py``
- [x] ``.\tests\c_api_test\test_.py``
- [x] ``.tests\distributed\conftest.py``
- [x] ``.tests\distributed\_test_distributed.py``
- [x] ``.\tests\python_package_test\test_basic.py``
- [x] ``.\tests\python_package_test\test_consistency.py``
- [x] ``.\tests\python_package_test\test_dask.py``
- [x] ``.\tests\python_package_test\test_dual.py``
- [x] ``.\tests\python_package_test\test_engine.py``
- [x] ``.\tests\python_package_test\test_plotting.py``
- [x] ``.\tests\python_package_test\test_sklearn.py``
- [x] ``.\tests\python_package_test\test_utilities.py``
[...]

PR #4428, opened 2021-07-02
[python] migrate to pathlib in python examples

StrikerRUS:
Contributes to #4416.


jameslamb (approved):
looks ok to me, thanks

PR #4429, opened 2021-07-02
[ci] add h5 files into `.gitignore`

StrikerRUS:
Due to
https://github.com/microsoft/LightGBM/blob/c359896e9bdbbbf89af5e186dd07080a40545452/examples/python-guide/dataset_from_multi_hdf5.py#L89-L90


jameslamb (approved):
ah good idea, thanks

PR #4432, opened 2021-07-03
[ci] ensure interactive_plot_example notebook is run in interactive mode at CI

StrikerRUS:


jameslamb (approved):
Ah ok, I see why we're doing this. Seems good to me!

PR #4440, opened 2021-07-04
[python] allow to pass some params as pathlib.Path objects

StrikerRUS:
Allow to pass filenames as `libpath.Path` objects. Now path-like objects can be used in both named arguments and `params` dictionary (`kwargs`).

Contributes to #4416.

### DIFF
--- a/tests/python_package_test/test_engine.py
+++ b/tests/python_package_test/test_engine.py
@@ -2261,9 +2261,8 @@ def test_forced_bins():
     x[:, 0] = np.arange(0, 1, 0.01)
     x[:, 1] = -np.arange(0, 1, 0.01)
     y = np.arange(0, 1, 0.01)
-    forcedbins_filename = str(
-        Path(__file__).absolute().parents[2] / 'examples' / 'regression' / 'forced_bins.json'
-    )
+    forcedbins_filename = (Path(__file__).absolute().parents[2] / 'examples'
+                           / 'regression' / 'forced_bins.json')
     params = {'objective': 'regression_l1',
               'max_bin': 5,
               'forcedbins_filename': forcedbins_filename,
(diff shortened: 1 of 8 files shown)


tests/python_package_test/test_engine.py
```diff
@@ -2261,9 +2261,8 @@ def test_forced_bins():
     x[:, 0] = np.arange(0, 1, 0.01)
     x[:, 1] = -np.arange(0, 1, 0.01)
     y = np.arange(0, 1, 0.01)
-    forcedbins_filename = str(
-        Path(__file__).absolute().parents[2] / 'examples' / 'regression' / 'forced_bins.json'
-    )
+    forcedbins_filename = (Path(__file__).absolute().parents[2] / 'examples'
+                           / 'regression' / 'forced_bins.json')
```
jameslamb, line 2265: I think the original indentation currently on `master` is easier to read than this (here and on the similar one below). Would you consider just removing the `str()` but otherwise leaving the original indentation?

jameslamb (approved):
If you are ok with supporting this expansion of the API and think the usability is worth the added maintenance burden, it's fine with me.

I left one very small suggestion about formatting but I'm approving because I'm ok with this being merged whether or not you agree with that suggestion.

StrikerRUS:
> If you are ok with supporting this expansion of the API and think the usability is worth the added maintenance burden, it's fine with me.

I don't think it actually costs a lot us to support `pathlib.Path` as a valid type of parameter. All we need to do, is just add `str()` calls in some places.

What is important, it helps users easier switch between XGBoost and LightGBM because XGBoost has been supporting path-like params for some time.
https://github.com/dmlc/xgboost/blob/f937f514aafdc05515355e35158532d7e593ae74/python-package/xgboost/core.py#L713

tests/python_package_test/test_engine.py
StrikerRUS: Yes, agree with you! I don't know why I chose to change indentation 🙂 .

StrikerRUS:
```suggestion
    forcedbins_filename = (
        Path(__file__).absolute().parents[2] / 'examples' / 'regression' / 'forced_bins.json'
    )
```

tests/python_package_test/test_engine.py
```diff
@@ -2285,9 +2284,8 @@ def test_forced_bins():
     est = lgb.train(params, lgb_x, num_boost_round=20)
     predicted = est.predict(new_x)
     assert len(np.unique(predicted)) == 3
-    params['forcedbins_filename'] = str(
-        Path(__file__).absolute().parents[2] / 'examples' / 'regression' / 'forced_bins2.json'
-    )
+    params['forcedbins_filename'] = (Path(__file__).absolute().parents[2] / 'examples'
+                                     / 'regression' / 'forced_bins2.json')
```
StrikerRUS, line 2288:
```suggestion
    params['forcedbins_filename'] = (
        Path(__file__).absolute().parents[2] / 'examples' / 'regression' / 'forced_bins2.json'
    )
```

PR #4441, opened 2021-07-04
[docs][python] add versionadded to Sequence class in Python wrapper

StrikerRUS:
Linking #4089.


jameslamb (approved):
ah yeah great idea, thanks

PR #4444, opened 2021-07-05
[python] migrate to pathlib in setup.py and use `absolute()` on paths first

StrikerRUS:
Contributes to #4416.

### DIFF
--- a/python-package/setup.py
+++ b/python-package/setup.py
@@ -1,14 +1,13 @@
 # coding: utf-8
 """Setup lightgbm package."""
 import logging
-import os
 import struct
 import subprocess
 import sys
-from distutils.dir_util import copy_tree, create_tree, remove_tree
-from distutils.file_util import copy_file
+from pathlib import Path
 from platform import system
-from typing import List, Optional
+from shutil import copyfile, copytree, rmtree
+from typing import List, Optional, Union
 
 from setuptools import find_packages, setup
 from setuptools.command.install import install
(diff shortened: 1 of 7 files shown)


python-package/setup.py
```diff
@@ -129,15 +116,15 @@ def compile_cpp(
...
-
-    if os.path.exists(os.path.join(CURRENT_DIR, "build_cpp")):
-        remove_tree(os.path.join(CURRENT_DIR, "build_cpp"))
-    os.makedirs(os.path.join(CURRENT_DIR, "build_cpp"))
-    os.chdir(os.path.join(CURRENT_DIR, "build_cpp"))
+    build_dir = CURRENT_DIR / "build_cpp"
+    rmtree(build_dir, ignore_errors=True)
+    build_dir.mkdir(parents=True)
+    original_dir = Path.cwd()
+    chdir(build_dir)
```
StrikerRUS, line 123: Unfortunately, we have to `chdir` due to old CMake doesn't support specifying building directory (I don't want to use undocumented flag):
https://stackoverflow.com/a/24435795
https://stackoverflow.com/a/20611964

> ` cmake [<options>] -S <path-to-source> -B <path-to-build>`
   https://cmake.org/cmake/help/latest/manual/cmake.1.html#synopsis

and no such possibility in CMake < 3.13
https://cmake.org/cmake/help/v3.12/manual/cmake.1.html#synopsis

jameslamb: thanks for calling this out, makes sense!

jameslamb (approved):
looks good, thanks!

PR #4446, opened 2021-07-06
[python-package] use toarray() instead of todense() in tests and examples

jameslamb:
Based on the advice in https://github.com/microsoft/LightGBM/pull/4378#issuecomment-873805964, this PR proposes changing a few uses of `.todense()` called on sparse arrays to `.toarray()`.

This method converts a sparse array to a `np.array`, a better-supported object type than `np.matrix` (the type returned by `.todense()`).

See https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.spmatrix.toarray.html.

### Notes for Reviewers

There are a few other uses of `.todense()` in the Dask tests, but those will be changed as part of #4378.

### DIFF
--- a/examples/python-guide/dask/ranking.py
+++ b/examples/python-guide/dask/ranking.py
@@ -30,7 +30,7 @@ if __name__ == "__main__":
 
     # make this array dense because we're splitting across
     # a sparse boundary to partition the data
-    X = X.todense()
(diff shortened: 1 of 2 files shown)


StrikerRUS (approved):
Nice change, thanks a lot!

PR #4448, opened 2021-07-06
Log warnings for number of bins of categorical features

### LINKED ISSUE #3735: Ignoring set max_bin for categorical features should log a warning to avoid confusing log messages
When training on GPU and setting max_bin (or max_bin_by_feature), the parameter might be silently ignored for categorical features with number of categories >= max_bin.

Still, the gpu_tree_learner might issue a warning that max_bin should be set for best performance. This behavior can confuse users as they are not given any hint why the set max_bin (as suggested by the warning) has been ignored.

Suggested behavior:

When max_bin (or max_bin_by_feature) has been set, but has to be ignored (e.g. due to a higher number of categories) a warning should be logged that indicates the problem.

* Python package
LightGBM component: treelearner

## Environment info

Operating System: Linux
CPU/GPU model: GPU
C++ compiler version: gcc 8.3.0
CMake version: 3.13.4
Python version: 3.7.8

LightGBM version or commit hash: 3fd4886

## Reproducible example(s)
```
import lightgbm as lgb
import numpy as np
import pandas as pd
params = {
    'max_bin': 15,
    'objective': 'binary',
    'boosting_type': 'gbdt',
    'device': 'gpu',
    'gpu_platform_id': 0,
    'gpu_device_id': 0
}
row_count = 10000
category_count = 15
X_train = pd.DataFrame({
    'x_int': np.random.randint(category_count, size=row_count),
    'x_float': np.random.normal(size=row_count)
})
y_train = np.random.randint(2, size=row_count)

# this will issue a warning that max_bin should be set to 15 (which is already the case) and
# compile a kernel with 64 bins
X_train['x_int'] = X_train['x_int'].astype('category')
model = lgb.train(params, train_data, 10)

# this will compile a kernel with 16 bins
X_train['x_int'] = X_train['x_int'].astype('int64')
model = lgb.train(params, train_data, 10)
```

shiyu1994:
This is to fix #3735, by
1. Log warnings when number of bins of categorical features exceeds the configured maximum number of bins.
2. When checking the bin number of feature groups in GPU versions (both GPU and CUDA) for better performance (to avoid bin numbers 17 and 65), we only consider the feature groups which contain no categorical features.

### DIFF
--- a/include/LightGBM/bin.h
+++ b/include/LightGBM/bin.h
@@ -156,9 +156,10 @@ class BinMapper {
   * \param use_missing True to enable missing value handle
   * \param zero_as_missing True to use zero as missing value
   * \param forced_upper_bounds Vector of split points that must be used (if this has size less than max_bin, remaining splits are found by the algorithm)
+  * \param feature_name Feature name of this feature, for displaying warning information
   */
   void FindBin(double* values, int num_values, size_t total_sample_cnt, int max_bin, int min_data_in_bin, int min_split_data, bool pre_filter, BinType bin_type,
-               bool use_missing, bool zero_as_missing, const std::vector<double>& forced_upper_bounds);
+               bool use_missing, bool zero_as_missing, const std::vector<double>& forced_upper_bounds, const std::string& feature_name);
 
   /*!
   * \brief Use specific number of bin to calculate the size of this class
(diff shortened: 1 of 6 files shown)


shiyu1994:
After a through thinking, it would be better to produce only one warning information for all categorical features which exceed the configured maximum bin number. Because it is very often that categorical features contain categories greater then `max_bin`. Producing one warning for each categorical feature will mess up the log output.

jameslamb:
I support  this PR! But don't think I'm qualified to review this code.

Could you update it to the latest `master`? And maybe one of the new maintainers for C++ code (@tongwu-msft or @hzy46) could help provide a review.

StrikerRUS:
@guolinke Could you please help to review this PR?

guolinke (approved):

PR #4449, opened 2021-07-06
[ci] Add CI job running rchk on the R package (fixes #4400)

### LINKED ISSUE #4400: [R-package] [ci] Add a CI job testing the R package with rchk
## Summary

A continuous integration (CI) job should be added which tests the R package with [`rchk`](https://github.com/kalibera/rchk).

From that project's description,

> This project consists of several bug-finding tools that look for memory protection errors in C source code using R API, that is in the source code of R itself and packages. The tools perform whole-program static analysis on LLVM bitcode and run on Linux. 

## Motivation

`rchk` can automatically find issues in C/C++ code used in an R package. Adding a CI job in this repository which runs `rchk` should provide higher confidence in releases and aid maintainers in catching user-facing issues before they make it into releases.

For example, I found that `rchk` run on the most recent CRAN release (3.2.1) would have caught the bug that led to #4045 (fixed in #4155).

> ERROR: initialization function R_init_lightgbm(_DllInfo*) in C++ will not be used by R

Following the instructions at https://github.com/kalibera/rchk/blob/master/doc/DOCKER.md#checking-a-package-from-a-tarball to check the latest version of `{lightgbm}` on CRAN.

```shell
wget https://cran.r-project.org/src/contrib/lightgbm_3.2.1.tar.gz
mkdir packages
cp lightgbm_3.2.1.tar.gz packages
docker run -v `pwd`/packages:/rchk/packages kalibera/rchk:latest /rchk/packages/lightgbm_3.2.1.tar.gz
```

Results in logs that end as follows.

```text
Installed libraries of package  lightgbm 
[1] "/rchk/packages/libsonly/lightgbm/libs/lightgbm.so"

Library name (usually package name): lightgbm
ERROR: initialization function R_init_lightgbm(_DllInfo*) in C++ will not be used by R
ERROR: did not find initialization function R_init_lightgbm
ERROR: too many states (abstraction error?) in function strptime_internal
Analyzed 39643 functions, traversed 106572 states.

Rchk version: 3d653b7c8f92dac912532856b55f44d2986c6553
R version: 79889/R Under development (unstable) (2021-01-27 r79889)
LLVM version: 10.0.0
```

## Description

If this job can be run on CI in less than 20 minutes, I think it should be included in the standard R checks run on every commit (https://github.com/microsoft/LightGBM/blob/47208894d819f1b8cef193e71f78e9a8c2f38462/.github/workflows/r_package.yml).

If not, it should be added as a standalone job that can be triggered by a comment like `/gha run r-rchk`. See https://github.com/microsoft/LightGBM/blob/47208894d819f1b8cef193e71f78e9a8c2f38462/.github/workflows/r_valgrind.yml for an example.

NOTE: when implementing this, the error `too many states (abstraction error) in strptime_internal` can safely be ignored. See https://github.com/kalibera/rchk/issues/22#issuecomment-656036156.

## References

Many thanks to @fabsig for bringing this up in #4390.

See https://github.com/microsoft/LightGBM/issues/4390#issuecomment-866527566 for some details on the current state of this tool's output against `master` of LightGBM.

jameslamb:
Fixes #4400

This PR proposes adding a new R CI job that tests the R package with [`rchk`](https://github.com/kalibera/rchk/commits/master). Such a check can be used to catch bugs of the form "C/C++ code failed to protect an object from R's garbage collection".

Please see #4400 for a detailed description and links to other documentation explaining the value of this check.

### Notes for Reviewers

I tested this job on my fork before opening this PR. https://github.com/jameslamb/LightGBM/pull/45

This job is taking around 13 minutes to run, so I'm proposing that it just be added to the regular set of checks run on every commit and not require manual triggering by a comment.

As an added bonus, this PR tests that LightGBM can be compiled with [`wllvm++`](https://github.com/travitch/whole-program-llvm).

* job succeeding on current `master`: https://github.com/jameslamb/LightGBM/pull/45/checks?check_run_id=2995447785
* job failing on a commit that intentionally re-introduced one of the bugs fixed by #4391
    - commit: https://github.com/jameslamb/LightGBM/pull/45/commits/c0a49cc81beaa1a828118e2b91635932b1682759
    - logs: https://github.com/jameslamb/LightGBM/pull/45/checks?check_run_id=2995519824

Example error caught by this job:

> Function LGBM_DatasetCreateFromFile_R
  [PB] has negative depth /rchk/packages/build/nfA5DRc4/lightgbm/src/lightgbm_R.cpp:67
  [UP] attempt to unprotect more items (3) than protected (2), results will be incomplete /rchk/packages/build/nfA5DRc4/lightgbm/src/lightgbm_R.cpp:67
  [PB] has possible protection stack imbalance /rchk/packages/build/nfA5DRc4/lightgbm/src/lightgbm_R.cpp:68

### DIFF
--- a/.ci/test_r_package.sh
+++ b/.ci/test_r_package.sh
@@ -94,7 +94,12 @@ fi
 
 # Manually install Depends and Imports libraries + 'testthat'
 # to avoid a CI-time dependency on devtools (for devtools::install_deps())
-packages="c('data.table', 'jsonlite', 'Matrix', 'R6', 'testthat')"
+# NOTE: testthat is not required when running rchk
+if [[ "${TASK}" == "r-rchk" ]]; then
+    packages="c('data.table', 'jsonlite', 'Matrix', 'R6')"
+else
+    packages="c('data.table', 'jsonlite', 'Matrix', 'R6', 'testthat')"
+fi
 compile_from_source="both"
 if [[ $OS_NAME == "macos" ]]; then
     packages+=", type = 'binary'"
--- a/.ci/test_r_package.sh
+++ b/.ci/test_r_package.sh
@@ -147,8 +147,8 @@ elif [[ $R_BUILD_TYPE == "cran" ]]; then
         # https://github.com/kalibera/rchk/issues/22#issuecomment-656036156
         exit $(
             cat ${RCHK_LOG_FILE} \
-            grep -v "in function strptime_internal" \
-            grep --count -E '\[PB\]|ERROR'
+            | grep -v "in function strptime_internal" \
+            | grep --count -E '\[PB\]|ERROR'
         )
     fi
 
--- a/.ci/test_r_package.sh
+++ b/.ci/test_r_package.sh
@@ -130,6 +130,21 @@ elif [[ $R_BUILD_TYPE == "cran" ]]; then
 
     ./build-cran-package.sh || exit -1
 
+    if [[ "${TASK}" == "r-rchk" ]]; then
+        echo "Checking R package with rchk"
+        mkdir -p packages
+        cp ${PKG_TARBALL} packages
+        RCHK_LOG_FILE="rchk-logs.txt"
+        docker run \
+            -v $(pwd)/packages:/rchk/packages \
+            kalibera/rchk:latest \
+            "/rchk/packages/${PKG_TARBALL}" \
+        2>&1 > ${RCHK_LOG_FILE} \
+        || (cat ${RCHK_LOG_FILE} && exit -1)
+        cat ${RCHK_LOG_FILE}
+        exit $(grep --count -E '\[PB\]|ERROR' < ${RCHK_LOG_FILE})
+    fi
+
     # Test CRAN source .tar.gz in a directory that is not this repo or below it.
     # When people install.packages('lightgbm'), they won't have the LightGBM
     # git repo around. This is to protect against the use of relative paths
--- a/.ci/test_r_package.sh
+++ b/.ci/test_r_package.sh
@@ -142,7 +142,14 @@ elif [[ $R_BUILD_TYPE == "cran" ]]; then
         2>&1 > ${RCHK_LOG_FILE} \
         || (cat ${RCHK_LOG_FILE} && exit -1)
         cat ${RCHK_LOG_FILE}
-        exit $(grep --count -E '\[PB\]|ERROR' < ${RCHK_LOG_FILE})
+
+        # the exception below is from R itself and not LightGBM:
+        # https://github.com/kalibera/rchk/issues/22#issuecomment-656036156
+        exit $(
+            cat ${RCHK_LOG_FILE} \
+            grep -v "in function strptime_internal" \
+            grep --count -E '\[PB\]|ERROR'
+        )
     fi
 
     # Test CRAN source .tar.gz in a directory that is not this repo or below it.
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,43 +1,43 @@
-version: 3.2.1.99.{build}
+# version: 3.2.1.99.{build}
 
-image: Visual Studio 2015
-platform: x64
-configuration:  # a trick to construct a build matrix with multiple Python versions
-  - 3.7
+# image: Visual Studio 2015
+# platform: x64
+# configuration:  # a trick to construct a build matrix with multiple Python versions
+#   - 3.7
 
-# only build pull requests and
-# commits to 'master'
-branches:
-  only:
-    - master
+# # only build pull requests and
+# # commits to 'master'
+# branches:
+#   only:
+#     - master
 
-environment:
-  matrix:
-    - COMPILER: MSVC
-      TASK: python
-    - COMPILER: MINGW
-      TASK: python
+# environment:
+#   matrix:
+#     - COMPILER: MSVC
+#       TASK: python
+#     - COMPILER: MINGW
+#       TASK: python
 
-clone_depth: 5
+# clone_depth: 5
 
-install:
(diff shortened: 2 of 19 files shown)


.ci/test_r_package.sh
```diff
@@ -94,7 +94,12 @@ fi
 
 # Manually install Depends and Imports libraries + 'testthat'
 # to avoid a CI-time dependency on devtools (for devtools::install_deps())
-packages="c('data.table', 'jsonlite', 'Matrix', 'R6', 'testthat')"
+# NOTE: testthat is not required when running rchk
```
jameslamb, line 97: `{testthat}` has a lot of dependencies, and since the `rchk` job runs on Linux they have to be compiled from source (since CRAN does not provide Linux binaries).

This additional `if` statement adds a bit of new complexity to this CI script, but I think it's worth it because I found in testing on my fork that this saves 1-2 minutes on the run time of the job.

.ci/test_r_package.sh
```diff
@@ -130,6 +135,28 @@ elif [[ $R_BUILD_TYPE == "cran" ]]; then
...
     ./build-cran-package.sh || exit -1
 
+    if [[ "${TASK}" == "r-rchk" ]]; then
+        echo "Checking R package with rchk"
+        mkdir -p packages
+        cp ${PKG_TARBALL} packages
+        RCHK_LOG_FILE="rchk-logs.txt"
+        docker run \
+            -v $(pwd)/packages:/rchk/packages \
+            kalibera/rchk:latest \
```
StrikerRUS, line 145: This docker image doesn't have any already configured R environment, right?

jameslamb: I think that it does, but the way that it is set up, it expects that you've built the package outside and then passed it into a container running the image.

https://github.com/kalibera/rchk/blob/master/doc/DOCKER.md#checking-a-package-from-a-tarball

that's why I chose to use `docker run` here after our usual R setup stuff on an `ubuntu-latest` VM, instead of running a containerized job like

https://github.com/microsoft/LightGBM/blob/ec1debcee84e6b02f749282d9a501021370e74a9/.github/workflows/r_package.yml#L167-L171

I think it's ok since `docker` is available by default in the `ubuntu-latest` image: https://github.com/actions/virtual-environments/blob/7d0d3aa82ab8565cc43684bd08bf72fb6f5cd9f7/images/linux/Ubuntu2004-README.md

StrikerRUS: Thanks! That's exactly what I wanted to clarify.

StrikerRUS: Did you see that?
https://hub.docker.com/r/rhub/ubuntu-rchk
https://github.com/r-hub/rhub-linux-builders/blob/master/ubuntu-rchk/Dockerfile
Maybe this image can help us to run containerized job or we don't want to do that?..

jameslamb: Oh interesting, no I didn't know about that. I can try it tonight or tomorrow.

jameslamb: Alright I tested this tonight but couldn't quite get it working.

I tried the following from the root of the `LightGBM` repo.

```shell
# get a shell inside an ubuntu-rchk container
docker run \
    -v $(pwd):/opt/LightGBM \
    -w /opt/LightGBM \
    -it rhub/ubuntu-rchk \
    /bin/bash

# run the following inside the container
sh build-cran-package.sh
Rscript -e "install.packages(c('R6', 'data.table', 'jsonlite', 'Matrix'), repos = 'https://cran.r-project.org')"
rchk.sh lightgbm_*.tar.gz
```

Compilation of `{lightgbm}` and its dependencies succeeded, but running this failed with an error I don't understand.

Here is the tail of the logs:

> /usr/local/bin/wllvm++ -stdlib=libc++ -std=gnu++11 -shared -L/usr/local/lib -o lightgbm.so boosting/boosting.o boosting/gbdt.o boosting/gbdt_model_text.o boosting/gbdt_prediction.o boosting/prediction_early_stop.o io/bin.o io/config.o io/config_auto.o io/dataset.o io/dataset_loader.o io/file_io.o io/json11.o io/metadata.o io/parser.o io/train_share_states.o io/tree.o metric/dcg_calculator.o metric/metric.o objective/objective_function.o network/ifaddrs_patch.o network/linker_topo.o network/linkers_mpi.o network/linkers_socket.o network/network.o treelearner/data_parallel_tree_learner.o treelearner/feature_parallel_tree_learner.o treelearner/gpu_tree_learner.o treelearner/linear_tree_learner.o treelearner/serial_tree_learner.o treelearner/tree_learner.o treelearner/voting_parallel_tree_learner.o c_api.o lightgbm_R.o -pthread
installing to /home/docker/R-svn/packages/lib/00LOCK-lightgbm/00new/lightgbm/libs
** R
** data
** demo
** byte-compile and prepare package for lazy loading
** help
*** installing help indices
*** copying figures
** building package indices
** testing if installed package can be loaded from temporary location
** checking absolute paths in shared objects and dynamic libraries
** testing if installed package can be loaded from final location
** testing if installed package keeps a record of temporary installation path
>* DONE (lightgbm)
/home/docker/rchk/scripts/check_package.sh: line 100:  2577 Killed                  $RCHK/src/$T $RBC $F > $FOUT 2>&1

So I think this PR should be merged in its current state. It would have been great to have a fully-containerized job so that we wouldn't have to spend CI time installing R for this, but I don't think investigating this further is the best use of our time, given that the current run time of the job is already not too bad (around 13 minutes).

StrikerRUS: OK, got it! Thanks a lot for the try! If you have some time, I guess it worth to report this error to https://github.com/r-hub/rhub-linux-builders with example of successful run directly in https://github.com/kalibera/rchk.

PR #4453, opened 2021-07-07
[ci] bump CUDA version from 11.2.2 to 11.4.0 at CI

StrikerRUS:


jameslamb (approved):
I'm ok with this if CI passes!

I think that it would be valuable to explicitly mention the version in the PR title, so that people who are looking at the changelog of the next release will see 11.4.0 directly.

Maybe like this

> [ci] bump CUDA version from 11.2.2 to 11.4.0 at CI

PR #4457, opened 2021-07-09
[tests][python] added tests for early stop in prediction in ranking task

StrikerRUS:
Refer to https://github.com/microsoft/LightGBM/pull/4411#issuecomment-869208342.

Currently, prediction early stopping is never tested with ranking task.


jameslamb (approved):
Hey great, thanks very much for digging through the source code for prediction early stopping and discovering that it's expected to work for ranking too!

PR #4462, opened 2021-07-10
[python] parallelize MinGW make similarly to Unix make command

StrikerRUS:
Refer to #1745.


jameslamb (approved):
Oh wow, very nice change?

Do you think that CMake-based builds of the R package (https://github.com/microsoft/LightGBM/blob/ef76db45c73338156a1f722914f910f82e60dcec/R-package/src/install.libs.R) would benefit from this too?

StrikerRUS:
@jameslamb Yeah, I think so. Nice catch!

PR #4468, opened 2021-07-13
Fix undefined behavior with NaN input in CategoricalDecision()

hcho3:
Currently, the `CategoricalDecision()` function casts the input `fval` into `int`, when `fval` is possibly a NaN (Not-a-Number):

https://github.com/microsoft/LightGBM/blob/5b7a6f3e7150aeb704d1dd2b852d246af3e913a3/include/LightGBM/tree.h#L368-L379

This cast causes undefined behavior when `fval` is NaN. According to https://stackoverflow.com/a/10366541:

> When a finite value of real floating type is converted to an integer type other than _Bool, the fractional part is discarded (i.e., the value is truncated toward zero). **If the value of the integral part cannot be represented by the integer type, the behavior is undefined.**

Casting NaN into `int` may or may not produce a negative value.

Minimal reproducible example, where this is a problem:
```python
import lightgbm as lgb
import numpy as np

X = np.array(30*[[1]] + 30*[[2]] + 30*[[0]])
y = np.array(60 * [5] + 30*[10])
train_data = lgb.Dataset(X, label=y, categorical_feature=[0])
bst = lgb.train({}, train_data, 1)

print(bst.predict(np.array([[np.NaN], [0.0]])))
# prints array([6.5       , 6.99999999])
```
(Example taken from dmlc/treelite#277)
Model plot:
image

The model produced by the example has `missing_type` field set to `MissingType::None`, indicating that all missing-value inputs should be mapped to 0. However, undefined behavior when casting NaN into `int` makes it uncertain how `CategoricalDecision()` would behave:
* If NaN gets casted into a negative value: `CategoricalDecision()` will return the right child when given the NaN input.
* If NaN gets casted into a positive value: given the NaN input, `CategoricalDecision()` will behave as if it's given an input of 0.

On my machine, casting NaN into `int` produces -2147483648.

**Proposed Fix**. In  `CategoricalDecision()`, check the input `fval` against NaN **first**. The input should be casted into `int` only when it's known to be not NaN.
* The fix eliminates the undefined behavior.
* The fix also ensures that the missing type setting is honored.

### DIFF
--- a/include/LightGBM/tree.h
+++ b/include/LightGBM/tree.h
@@ -366,13 +366,9 @@ class Tree {
   }
 
   inline int CategoricalDecision(double fval, int node) const {
-    uint8_t missing_type = GetMissingType(decision_type_[node]);
     int int_fval;
     if (std::isnan(fval)) {
-      if (missing_type == MissingType::NaN) {
-        return right_child_[node];
-      }
-      int_fval = 0;
+      return right_child_[node];
     } else {
       int_fval = static_cast<int>(fval);
       if (int_fval < 0) {
--- a/include/LightGBM/tree.h
+++ b/include/LightGBM/tree.h
@@ -367,15 +367,17 @@ class Tree {
 
   inline int CategoricalDecision(double fval, int node) const {
     uint8_t missing_type = GetMissingType(decision_type_[node]);
-    int int_fval = static_cast<int>(fval);
-    if (int_fval < 0) {
-      return right_child_[node];;
-    } else if (std::isnan(fval)) {
-      // NaN is always in the right
+    int int_fval;
+    if (std::isnan(fval)) {
       if (missing_type == MissingType::NaN) {
         return right_child_[node];
       }
       int_fval = 0;
+    } else {
+      int_fval = static_cast<int>(fval);
+      if (int_fval < 0) {
+        return right_child_[node];
+      }
     }
     int cat_idx = static_cast<int>(threshold_[node]);
     if (Common::FindInBitset(cat_threshold_.data() + cat_boundaries_[cat_idx],


StrikerRUS:
@jameslamb Could you please take a look at the failing Dask test? I don't believe the error is related to this PR, but I don't want to lose corresponding logs by clicking "rerun".

shiyu1994:
@hcho3 Thank you! Nice fix. But I notice there's another problem in this case.
>If NaN gets casted into a positive value: given the NaN input, CategoricalDecision() will behave as if it's given an input of 0.

After the fix, `NaN` still behaves as `0` if the missing type of the categorical feature is `None`. But the raw value `0` can be a valid category in the original dataset.
So, if we have a dataset contains `NaN`, if it is used as validation set, `NaN` behaves as `-1` (maps to bin 0) according to
https://github.com/microsoft/LightGBM/blob/5b7a6f3e7150aeb704d1dd2b852d246af3e913a3/include/LightGBM/bin.h#L464-L468
Note in the above code that this is *bin* 0, not the *original feature* value 0.
However, when we treat the dataset as test set in `predict` methods, the `NaN` will be treated as *original feature* value 0 according to current logic, which is a mismatch.

So I suggest we should directly return right child when `nan` of categorical feature is encountered during prediction.

Following is an minimal example which reproduces the problem stated above. The evaluation result and the prediction result differ.

```python
import lightgbm as lgb
import numpy as np

X = np.array(30*[[1]] + 30*[[2]] + 30*[[0]])
y = np.array(60 * [5] + 30*[10])
X_test = np.array([[np.NaN], [0.0]])
y_test = [0, 1]

train_data = lgb.Dataset(X, label=y, categorical_feature=[0])
test_data = lgb.Dataset(X_test, label=y_test, reference=train_data)

bst = lgb.train(params={"metric":"rmse", "objective":"regression"},
                train_set=train_data,
                num_boost_round=1,
                keep_training_booster=True,
                valid_sets=[test_data],
                valid_names=["test"])

eval_result = bst.eval(test_data, "test")
print("eval rmse: %f" % eval_result[0][2])

pred = bst.predict(X_test)
print("predict rmse: %f" % np.sqrt(np.mean((pred - y_test) ** 2)))
```

This outputs
```
eval rmse: 6.254998
predict rmse: 6.519202
```
if directly return right child whenever `NaN` is encountered, it gives
```
eval rmse: 6.254998
predict rmse: 6.254998
```

hcho3:
@shiyu1994 

> NaN behaves as -1

> we should directly return right child when nan of categorical feature is encountered during prediction.

Do you mean that `MissingType` setting is only applicable to numerical splits and not categorical splits? So `Nan` should always map to the right child node, irrespective of the value of the missing type field?

shiyu1994:
@hcho3 Yes.
According to
https://github.com/microsoft/LightGBM/blob/5b7a6f3e7150aeb704d1dd2b852d246af3e913a3/src/io/bin.cpp#L465-L468
categorical feature will always has bin 0 for `NaN` and unseen categories.
And when finding best split for categorical features
https://github.com/microsoft/LightGBM/blob/5b7a6f3e7150aeb704d1dd2b852d246af3e913a3/src/treelearner/feature_histogram.hpp#L306-L308
https://github.com/microsoft/LightGBM/blob/5b7a6f3e7150aeb704d1dd2b852d246af3e913a3/src/treelearner/feature_histogram.hpp#L374-L379
Here `offset` is a mark for whether bin 0 is actually stored in the histogram. So we can see that bin 0 is always ignored when finding split for categorical features.
So the data in bin 0 of categorical features always go to the right child.

hcho3:
@shiyu1994 Thanks for your reply. How does LightGBM handle sparse arrays? Does missing values in a Scipy arrays get mapped to zeros?

shiyu1994:
@hcho3 Yes. They are mapped to zeros.

shiyu1994 (approved):

StrikerRUS:
Azure Pipelines is stuck. Sorry, I have to close-reopen PR to re-trigger CI.

Issue #4403, opened 2021-06-23
Take `Sequence` into consideration in `get_data()` and `add_features_from()` methods of `Dataset` object


StrikerRUS:
This issue is actual only after merging #4089.

Refer to https://github.com/microsoft/LightGBM/pull/4089#discussion_r654721270.

cyfdecyf:
I've started working on this. As I don't uses these two APIs, I need some guidence.

PR #4472 adds support `Sequence` input for `Dataset.get_data`. I'm wondering what's the use case of creating a subset with an existing `Dataset`.

For `Dataset.add_features_from`, I'm reluctant to support this naively for two reasons:

- It defeats the saving memory benefit of `Sequence` interface
  - The current implementation seems combing raw data together
- We can combine features by `Sequence` interface easily
  - Thus don't need to use `add_features_from`

Here's a simple combine Sequence by columns example:

```python
class CombineSequenceByColumn(lgb.Sequence):
    def __init__(self, seqlst: List[lgb.Sequence], batch_size):
        self.seqlst = seqlst
        self.batch_size = batch_size

    def __getitem__(self, idx):
        if isinstance(idx, numbers.Integral):
            return np.hstack([arr[idx] for arr in self.seqlst])
        elif isinstance(idx, slice):
            return np.hstack([arr[idx.start:idx.stop] for arr in self.seqlst])
        else:
            raise TypeError(f"Sequence Index must be an integer/list/slice, got {type(idx).__name__}")

    def __len__(self):
        return len(self.seqlst[0])
```

If we really want to support `Sequence` input for `add_features_from`, I suggest first create a class `CombineSequenceByRow` to combine `List[Sequence]` as a single `Sequence` object. With those two `Sequence` combining classes, we can support `Sequence` input for `add_features_from` like this:

- If both datasets' input are `Sequence`, then create an object of `CombineSequenceByColumn`
- If only one dataset is `Sequence`, then adapt the other as `Sequence` and then combine as above case
- For input of `List[Sequence]`, create an object of `CombineSequenceByRow` and then handle as above cases

This implementation can still retain the benefit of saving memory, but it's a lot of work as we have to adapt various existing supprted types to `Sequence`. It would be much easier to ask the user to combine features with `Sequence` interface.

StrikerRUS:
@cyfdecyf Thanks a lot for taking this issue!
> It would be much easier to ask the user to combine features with `Sequence` interface.

[...]

PR #4472, opened 2021-07-14
[python] support Dataset.get_data for Sequence input.

cyfdecyf:
Implement features for #4403.

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -641,9 +641,9 @@ class Sequence(abc.ABC):
         .. code-block:: python
 
             if isinstance(idx, numbers.Integral):
-                return self.__get_one_line__(idx)
+                return self._get_one_line(idx)
             elif isinstance(idx, slice):
-                return np.stack(self.__get_one_line__(i) for i in range(idx.start, idx.stop))
+                return np.stack(self._get_one_line(i) for i in range(idx.start, idx.stop))
             else:
                 raise TypeError(f"Sequence index must be integer or slice, got {type(idx).__name__}")
 
@@ -1515,7 +1515,8 @@ class Dataset:
         # set feature names
         return self.set_feature_name(feature_name)
 
-    def __yield_row_from(self, seqs: List[Sequence], indices: Iterable[int]):
+    @staticmethod
+    def _yield_row_from_seqlist(seqs: List[Sequence], indices: Iterable[int]):
         offset = 0
         seq_id = 0
         seq = seqs[seq_id]
@@ -1541,7 +1542,7 @@ class Dataset:
         indices = self._create_sample_indices(total_nrow)
 
         # Select sampled rows, transpose to column order.
-        sampled = np.array([row for row in self.__yield_row_from(seqs, indices)])
+        sampled = np.array([row for row in self._yield_row_from_seqlist(seqs, indices)])
         sampled = sampled.T
 
         filtered = []
@@ -2236,7 +2237,7 @@ class Dataset:
 
         Returns
         -------
-        data : string, pathlib.Path, numpy array, pandas DataFrame, H2O DataTable's Frame, scipy.sparse, list of numpy arrays or None
+        data : string, pathlib.Path, numpy array, pandas DataFrame, H2O DataTable's Frame, scipy.sparse, Sequence, list of Sequences or list of numpy arrays or None
             Raw data used in the Dataset construction.
         """
         if self.handle is None:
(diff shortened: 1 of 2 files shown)


python-package/lightgbm/basic.py
```diff
@@ -2250,6 +2251,10 @@ def get_data(self):
                     self.data = self.data.iloc[self.used_indices].copy()
                 elif isinstance(self.data, dt_DataTable):
                     self.data = self.data[self.used_indices, :]
+                elif isinstance(self.data, Sequence):
+                    self.data = np.array([self.data[i] for i in self.used_indices])
```
StrikerRUS, line 2255: I thought we require from `Sequence` implementations to be able pass list as index.
https://github.com/microsoft/LightGBM/blob/5b7a6f3e7150aeb704d1dd2b852d246af3e913a3/python-package/lightgbm/basic.py#L636
https://github.com/microsoft/LightGBM/blob/5b7a6f3e7150aeb704d1dd2b852d246af3e913a3/python-package/lightgbm/basic.py#L645
So maybe we can simplify to `self.data[self.used_indices]`?

python-package/lightgbm/basic.py
```diff
@@ -2363,6 +2368,10 @@ def add_features_from(self, other):
         if other.data is None:
             self.data = None
         elif self.data is not None:
+            if isinstance(self.data, Sequence) or isinstance(other.data, Sequence):
+                raise ValueError('Add features from Sequence is not supported.'
+                                 'Please combine features in your Sequence implementation.')
```
StrikerRUS, line 2373: If you want to prohibit using `add_features_from` with `Sequence` at all, you need to move this error early than calling combining features at cpp side.
https://github.com/microsoft/LightGBM/blob/5b7a6f3e7150aeb704d1dd2b852d246af3e913a3/python-package/lightgbm/basic.py#L2360

Alternatively, I think it will be enough to yield a default warning 
https://github.com/microsoft/LightGBM/blob/5b7a6f3e7150aeb704d1dd2b852d246af3e913a3/python-package/lightgbm/basic.py#L2419-L2424
which informs users that features were combined at cpp side but they are not allowed to call `get_data()` anymore at Python side.

StrikerRUS (commented):
@cyfdecyf Many thanks for this PR! Please check my initial thoughts below:

python-package/lightgbm/basic.py
```diff
@@ -1515,7 +1515,8 @@ def _lazy_init(self, data, label=None, reference=None,
         # set feature names
         return self.set_feature_name(feature_name)
 
-    def __yield_row_from(self, seqs: List[Sequence], indices: Iterable[int]):
+    @staticmethod
```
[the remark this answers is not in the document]
cyfdecyf: Because we don't use `self` in this method and the logic is quite general.

I considered moving it to `Sequence` interface as a static method when writing this PR. This would exposes the ability of indexing from list of `Sequence` to user.

python-package/lightgbm/basic.py
cyfdecyf: I've made the change as suggested. The requirement is mentioned as a comment in `Sequence.__getitem__` docstring.

https://github.com/cyfdecyf/LightGBM/blob/sequence-dataset/python-package/lightgbm/basic.py#L648

python-package/lightgbm/basic.py
cyfdecyf: OK. As the cpp side is doing actual combining, I removed the raise statement.

python-package/lightgbm/basic.py
```diff
@@ -2250,6 +2254,10 @@ def get_data(self):
                     self.data = self.data.iloc[self.used_indices].copy()
                 elif isinstance(self.data, dt_DataTable):
                     self.data = self.data[self.used_indices, :]
+                elif isinstance(self.data, Sequence):
+                    self.data = self.data[self.used_indices]
+                elif isinstance(self.data, list) and len(self.data) > 0 and all(isinstance(x, Sequence) for x in self.data):
+                    self.data = np.array([row for row in self._yield_row_from_seqlist(self.data, self.used_indices)])
```
StrikerRUS, line 2260: Please add test cases with `Sequence` and `List[Sequence]` for `add_features_from()` here
https://github.com/microsoft/LightGBM/blob/90342e929b1801fb2a20b410e2db5a69a9063b48/tests/python_package_test/test_basic.py#L325

StrikerRUS: Also, please enhance this test (or just add a new one)
https://github.com/microsoft/LightGBM/blob/d05f54701ee88a1373bd9e0a1ea4317827015c9c/tests/python_package_test/test_basic.py#L133
with new `get_data()` functionality for `Sequence(s)` raw data.

cyfdecyf: Thanks for the review. I'll add test cases later this week.

cyfdecyf:
Test cases are added now. It's a little hacky because I'm not familiar with the use cases.

StrikerRUS (approved):
@cyfdecyf Thank you very much for continuing to enhance `Sequence` support!

> Test cases are added now. It's a little hacky because I'm not familiar with the use cases.

No problem! I can slightly refactor them later. Don't want to delay this PR.

PR #4485, opened 2021-07-24
[docs] document CLI behavior when label_column is omitted

jameslamb:
Answering [this stack overflow question](https://stackoverflow.com/questions/68482580/in-the-example-regression-which-is-the-target-column/68506958#68506958) tonight, I wasn't able to figure out from LightGBM's documentation that, when using the CLI, if `label_column` is omitted from `train.conf`, LightGBM defaults to using the first column in the training data as the target.

I had to figure this out from source code

* `Dataset.label_idx_`  is initialized to 0: https://github.com/microsoft/LightGBM/blob/a8ee487aca35363fafa027e6b7695976045096b3/src/io/dataset_loader.cpp#L22
* `Dataset.label_idx_` is only overridden if label_column is provided in the config: https://github.com/microsoft/LightGBM/blob/a8ee487aca35363fafa027e6b7695976045096b3/src/io/dataset_loader.cpp#L48-L57

This PR proposes two small changes to make this clearer for new users of the CLI:

* adds a line to the docs for https://lightgbm.readthedocs.io/en/latest/Parameters.html#label_column explaining this behavior
* adds `label_column` to all of the training configs in `examples/`, to nudge users who copy those configs as a reference to explicitly define `label_column` in their own configs

### DIFF
--- a/docs/Parameters.rst
+++ b/docs/Parameters.rst
@@ -762,6 +762,8 @@ Dataset Parameters
 
    -  add a prefix ``name:`` for column name, e.g. ``label=name:is_click``
 
+   -  if omitted, the first column in the training data is used as the label
+
    -  **Note**: works only in case of loading data directly from file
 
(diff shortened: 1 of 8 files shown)


StrikerRUS (approved):
Nice clarification!

PR #4486, opened 2021-07-24
remove examples/.gitignore

jameslamb:
Working on #4485, I was looking at the CLI examples in https://github.com/microsoft/LightGBM/tree/master/examples tonight and noticed there is a `.gitignore` left there (unmodified since October 2016, https://github.com/microsoft/LightGBM/commit/351b3d7e2cd900d6c61d652e376c00ca210cdaba).

This PR proposes removing it. As of this PR, the single root-level `.gitignore` would be the only place managing ignoring files in source control, which I think is a bit easier to maintain and reason about.


StrikerRUS (approved):
Thanks! Nice catch!

PR #4488, opened 2021-07-26
[docs][ci] bump versions of R-package dependencies at RTD

### LINKED ISSUE #4487: RTD builds are failing due to time out
```
Build took 999 seconds

Build exited due to time out
```

image

image

StrikerRUS:
Fixed #4487.

image

image

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -316,12 +316,12 @@ def setup(app):
         app.connect("builder-inited", generate_doxygen_xml)
     else:
         app.add_directive('doxygenfile', IgnoredDirective)
-    if RTD:  # build R docs only on Read the Docs site
-        if first_run:
-            app.connect("builder-inited", generate_r_docs)
-        app.connect("build-finished",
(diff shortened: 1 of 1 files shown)


StrikerRUS:
@jameslamb Yeah, there were no logs because we accumulate stdout and stderr first and then dump the combined result
https://github.com/microsoft/LightGBM/blob/fdc582ea6ba13faf15ee6707c7c7542790c8821d/docs/conf.py#L286-L301
But due to the time out error we didn't reach "dump" step.

First, I disabled R docs completely to narrow the root cause.
d3324602b2c95fd6c29180ca66e82596a79ebc01
That build without R docs was successful and took only 100 seconds. After that I was sure that something is wrong with R-package installation or R docs generation process. Then I just started with the simplest idea of upgrading deps because I remember one case when some particular package build had been removed from anaconda.org repo instead of pushing new build number/hash.

jameslamb:
> Then I just started with the simplest idea of upgrading deps because I remember one case when some particular package build had been removed from anaconda.org repo instead of pushing new build number/hash.

ah ok great, thanks for the explanation!

PR #4496, opened 2021-08-01
move Network method implementations from network.h to network.cpp (fixes #4464)

### LINKED ISSUE #4464: [R-package] R package crashes on windows when loaded together with {fansi} or anything that depends on it
This is probably related to:

- https://github.com/microsoft/LightGBM/issues/4045 and
- https://github.com/microsoft/LightGBM/issues/4007

## Description

Using lightgbm while `parsnip` is loaded crashes the R session with: Exited with status -1073741819.

## Reproducible example

Calling:

```
library(parsnip)
library(lightgbm)
data(agaricus.train, package='lightgbm')
train <- agaricus.train
model <- lgb.cv(
       params = list(
       objective = "regression", 
       metric = "l2"
       ) , 
data = dtrain
)
```

## Environment info

I am using the dev version of LightGBM as suggested in https://github.com/microsoft/LightGBM/issues/4007#issuecomment-869080432
The error only occurs on Windows.

Here's a GitHub actions run that shows the behavior. 
This shows that it works fine if parsnip is not loaded: https://github.com/curso-r/treesnip/runs/3037580458?check_suite_focus=true#step:9:1
And this one shows the error message: https://github.com/curso-r/treesnip/runs/3037580458?check_suite_focus=true#step:10:21

I could also reproduce it locally on a Windows machine, but I am not sure what's the best way to get a stack trace.
Let me know if I can help with further debugging.

jameslamb:
## Short Description

I *think* this PR fixes #4464. It moves implementation of `Network::num_machines()` and `Network::rank()` from the `network` header to its implementation source file, to prevent an issue where minGW compiles that code in a way that it's possible for `Network::num_machines()` to cause a crash.

## Long Description

I installed the CRAN build of the R package from latest `master` (https://github.com/microsoft/LightGBM/commit/661bde103ac474a04bd3958ac1dfb75307c7871d as of today), with the following commands.

```shell
sh build-cran-package.sh
```

Put the following (from #4464) in a script called `test.R`.

```r
library(fansi)
library(lightgbm)

data(agaricus.train, package='lightgbm')
train <- agaricus.train
model <- lgb.cv(
    params = list(
        objective = "regression"
        , metric = "l2"
    )
    , data = dtrain
)  
print("done training")
```

This produces a crash. Installing the CRAN-style package from source with this PR's branch and running the same code does not produce a crash.

I believe that this issue is minGW-specific, and that that compiler (or at least the version of it bundled with RTools) can create a situation where when `Network::num_machines()` is run here

https://github.com/microsoft/LightGBM/blob/661bde103ac474a04bd3958ac1dfb75307c7871d/src/io/dataset_loader.cpp#L626

the property `Network.num_machines_`  has not been been initialized yet.

https://github.com/microsoft/LightGBM/blob/661bde103ac474a04bd3958ac1dfb75307c7871d/src/network/network.cpp#L17

Something about loading the `{fansi}` R package before `{lightgbm}` causes something (maybe the loading of some other DLL)  which leads to this access-before-initialization problem.

**NOTE:** On `master`, the reproducible example from #4464 does not cause a crash on Windows when `{lightgbm}` is compiled with Visual Studio compilers (`Rscript build_r.R`). This crash also does not occur on macOS or Linux (for CMake-based or CRAN-style installation). I think this is another piece of evidence supporting the idea that the core issue is specific to minGW.


StrikerRUS (commented):
✔️ 
Awesome research and neat fix! I don't think that moving `rank()` and `num_machines()` functions to `.cpp` file can break anything. "Approving" based on reports https://github.com/microsoft/LightGBM/issues/4464#issuecomment-890458819 and https://github.com/microsoft/LightGBM/issues/4464#issuecomment-890507571.

jameslamb:
Thanks very much! Did you leave a "comment" review because you want @shiyu1994 to review as well or was it an accident?

StrikerRUS:
@jameslamb Yeah, I'd like any cpp maintainer leave their formal approval (maybe inlining was critical here, idk...)

jameslamb:
@shiyu1994 or @btrotta could you please take a look at this whenever you have time?

jameslamb:
@shiyu1994 or @btrotta could you please take a look at this whenever you have time?

shiyu1994 (approved):
@jameslamb Thanks for the nice fix! It looks good to me.

jameslamb:
Thanks @shiyu1994 ! I'll update this to latest `master` and re-run the valgrind and solaris R tests, just to be sure it's still working with other recent changes on the R package.

jameslamb:
after updating to the most recent `master`, all CI jobs are passing (including valgrind and Solaris). That plus two approvals makes me very confident in this PR, so I think it's ok to merge.

Thanks everyone for your help! I'm excited about this one, will be really helpful for users of the R package on Windows.

Issue #4259, opened 2021-05-05
R session crashes on Windows (fatal error), on basic tutorial example


actuariallonewolf:
## Description

R session has a fatal crash when trying to run lightgbm, on Windows.  
   
I had this behavior earlier this year and noticed it crashed whenever lightgbm was loaded last.  
My quick fix back then, was to load lightgbm first, and then everything worked out good.
  
Interestingly, it was later reported [here](https://github.com/microsoft/LightGBM/issues/4007).    
    

However, I'm opening a bug report this time around, because no matter what I try, lightgbm crashes all the time.  My trick no longer works...  
  
Thanks!

## Reproducible example

```
install.packages("lightgbm")
library(lightgbm)

data(agaricus.train, package='lightgbm')
train <- agaricus.train
dtrain <- lgb.Dataset(train$data, label = train$label)
model <- lgb.cv(
  params = list(
    objective = "regression"
    , metric = "l2"
  )
  , data = dtrain
)  
```

## Environment info

sessionInfo()

```
R version 4.0.5 (2021-03-31)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 18363)

Matrix products: default

locale:
[1] LC_COLLATE=English_Canada.1252  LC_CTYPE=English_Canada.1252    LC_MONETARY=English_Canada.1252 LC_NUMERIC=C                    LC_TIME=English_Canada.1252    

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] lightgbm_3.2.1 R6_2.4.1      

loaded via a namespace (and not attached):
 [1] Rcpp_1.0.5           lubridate_1.7.9.2    lattice_0.20-41      prettyunits_1.1.1    class_7.3-18         ps_1.5.0             assertthat_0.2.1    
 [8] rprojroot_2.0.2      digest_0.6.25        ipred_0.9-9          foreach_1.5.1        plyr_1.8.6           stats4_4.0.5         ggplot2_3.3.2       
[15] pillar_1.4.6         rlang_0.4.9          caret_6.0-86         rstudioapi_0.13      data.table_1.13.4    callr_3.5.1          rpart_4.1-15        
[22] Matrix_1.3-2         reticulate_1.19      desc_1.2.0           devtools_2.3.2       splines_4.0.5        gower_0.2.2          stringr_1.4.0       
[...]

PR #3946, opened 2021-02-12
[R-package] [docs] add intro vignette

mayer79:
First step towards https://github.com/microsoft/LightGBM/issues/1944. 

This initial vignette shows the very basic workflow of LightGBM. As such, it would *not* replace the demo with the corresponding name as that seems rather a guide how to deal with sparse data.

The html looks as in this zip here:

[basic_walkthrough.zip](https://github.com/microsoft/LightGBM/files/5970298/basic_walkthrough.zip)

### Some notes on R vignettes in general

- CRAN does not rebuild the vignette html, but it still runs its R code.
- While it is possible to use a logo, it is non-trivial.
- Would you stick to the same code style as in the package code? I am asking because LightGBM is the only project that use R code with comma first.
- The vignette html will be created automatically when building the package. Since the whole building process of LGB is complex, I have no idea if this works out as I hope... still, our colleagues here https://cran.r-project.org/web/packages/xgboost/index.html were successful.

PR #4499, opened 2021-08-02
[R-package] quote path variables in build-cran-package.sh

jameslamb:
LightGBM provides a script, `build-cran-package.sh`, which can be used to produce a CRAN-style source distribution of the R package.

This is effectively user-facing code, since we encourage users to run it while developing or building from `master` to get unreleased features (e.g. https://github.com/microsoft/LightGBM/issues/4259#issuecomment-832900302, https://github.com/microsoft/LightGBM/issues/4464#issuecomment-890454199).

I realized yesterday, working on #3946 , that that script can break if called from a path that has spaces in it.

Consider the following:

```shell
mkdir -p './some files'
cd 'some files'
cd LightGBM
sh build-cran-package.sh
```

This errors out like so:

> build-cran-package.sh: line 15: test: /Users/James.Lamb/repos/open-source/some: binary operator expected
cp: /Users/James.Lamb/repos/open-source/some is a directory (not copied).

This PR proposes quoting all shell variables in that script which include paths, which prevents this problem.

### DIFF
--- a/build-cran-package.sh
+++ b/build-cran-package.sh
@@ -9,13 +9,13 @@
 
 set -e
 
-ORIG_WD=$(pwd)
-TEMP_R_DIR=$(pwd)/lightgbm_r
+ORIG_WD="$(pwd)"
+TEMP_R_DIR="$(pwd)/lightgbm_r"
 
-if test -d ${TEMP_R_DIR}; then
-    rm -r ${TEMP_R_DIR}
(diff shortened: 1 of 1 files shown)


StrikerRUS (approved):
Thanks! I searched for `$` in that file and can confirm there are no unquoted path variables anymore.

PR #4498, opened 2021-08-01
[dask] find all needed ports in each host at once (fixes #4458)

jmoralez:
This replaces my initial proposal (#3823) of using `client.run` to find an open port in each worker, since that could produce collisions (#4057, #4458) when one host had more than one worker process running in it (as is the case with `LocalCluster`). The collisions problem was solved in #4133, however it would be desirable to not have any collisions at all, which is what this (hopefully) achieves.

PR #4501, opened 2021-08-04
[tests][dask] reduce number of collisions tests

jmoralez:
The recent change of the search for open ports for the dask interface (#4498) removed the collisions and to make sure we never got any the test ran 1,000 times, which is too much for the QEMU CI job (https://github.com/microsoft/LightGBM/pull/4498#discussion_r682665570).

This reduces the number of times the collisions are tested to 25 (as suggested in https://github.com/microsoft/LightGBM/pull/4498#discussion_r682809148).

### DIFF
--- a/.ci/test.sh
+++ b/.ci/test.sh
@@ -148,7 +148,7 @@ elif [[ $TASK == "bdist" ]]; then
         fi
     fi
     pip install --user $BUILD_DIRECTORY/python-package/dist/*.whl || exit -1
-    pytest --durations=0 $BUILD_DIRECTORY/tests || exit -1
+    pytest $BUILD_DIRECTORY/tests || exit -1
     exit 0
 fi
 
--- a/.ci/test.sh
+++ b/.ci/test.sh
@@ -207,7 +207,7 @@ fi
 make _lightgbm -j4 || exit -1
 
 cd $BUILD_DIRECTORY/python-package && python setup.py install --precompile --user || exit -1
-pytest $BUILD_DIRECTORY/tests || exit -1
+pytest --durations=0 $BUILD_DIRECTORY/tests || exit -1
 
 if [[ $TASK == "regular" ]]; then
     if [[ $PRODUCES_ARTIFACTS == "true" ]]; then
--- a/.ci/test.sh
+++ b/.ci/test.sh
@@ -148,7 +148,7 @@ elif [[ $TASK == "bdist" ]]; then
         fi
     fi
     pip install --user $BUILD_DIRECTORY/python-package/dist/*.whl || exit -1
-    pytest $BUILD_DIRECTORY/tests || exit -1
+    pytest --durations=0 $BUILD_DIRECTORY/tests || exit -1
     exit 0
 fi
 
@@ -207,7 +207,7 @@ fi
 make _lightgbm -j4 || exit -1
 
 cd $BUILD_DIRECTORY/python-package && python setup.py install --precompile --user || exit -1
-pytest --durations=0 $BUILD_DIRECTORY/tests || exit -1
+pytest $BUILD_DIRECTORY/tests || exit -1
 
 if [[ $TASK == "regular" ]]; then
     if [[ $PRODUCES_ARTIFACTS == "true" ]]; then
--- a/tests/python_package_test/test_dask.py
+++ b/tests/python_package_test/test_dask.py
@@ -465,7 +465,7 @@ def test_assign_open_ports_to_workers(cluster):
         workers = client.scheduler_info()['workers'].keys()
         n_workers = len(workers)
         host_to_workers = lgb.dask._group_workers_by_host(workers)
-        for _ in range(1_000):
+        for _ in range(25):
             worker_address_to_port = lgb.dask._assign_open_ports_to_workers(client, host_to_workers)
             found_ports = worker_address_to_port.values()
             assert len(found_ports) == n_workers


jmoralez:
Seems like it's still taking too long (53 minutes)
```
2021-08-04T18:30:40.3348177Z ../tests/python_package_test/test_dask.py .............................. [ 14%]
2021-08-04T18:39:07.3817934Z ........................................................................ [ 26%]
2021-08-04T18:54:31.2839865Z ........................................................................ [ 37%]
2021-08-04T19:11:17.5745812Z ......s...............s...............s...............s................. [ 49%]
2021-08-04T19:21:33.5172290Z ..................................s..................................... [ 61%]
2021-08-04T19:23:08.2142881Z s.......................s........                                        [ 66%]
2021-08-04T19:23:08.2245588Z ../tests/python_package_test/test_dual.py s                              [ 66%]
```

jmoralez:
@jameslamb would you support adding `--durations=0` to the `pytest` call to get the running time of each test? It may be another one that's taking this long.

StrikerRUS:
> ... which is too much for the QEMU CI job

We can set number of runs depending on the architecture tests are run on.
```python
from platform import machine

...

n_runs = 25 if machine() != 'x86_64' else 1000
for _ in range(n_runs):

...
```

jmoralez:
@StrikerRUS do you think it should still be run that many times in the other jobs? I think I may have got too excited with the "should never collide" idea.

jameslamb:
>  I think I may have got too excited with the "should never collide" idea.

Haha yeah I personally would prefer to just run it, say, 25 times in each job (like you currently have in the PR) and not introduce a difference in different test environments. I think that's more than enough to catch issues, given the current level of activity in this repo and the number of concurrent CI jobs running on each commit.

> @jameslamb would you support adding --durations=0 to the pytest call to get the running time of each test

I at least would support you pushing a commit to this PR right now so we can see those results in logs! I'd want to see how verbose that output is to make a decision of whether or not to merge such a change, but it would at least help us to understand where time is being spent.

jmoralez:
Hmm I thought changing https://github.com/microsoft/LightGBM/blob/5fe27d594285eeaf31d2f1ebfe29453f8c6c6ad3/.ci/test.sh#L210 affected the QEMU job, however it didn't print the times (I can see them in other jobs though).

Does that job run a different call to `pytest`?

StrikerRUS:
OK, please forget [my suggestion](https://github.com/microsoft/LightGBM/pull/4501#issuecomment-892939147).

StrikerRUS:
@jmoralez 

> Does that job run a different call to `pytest`?

QEMU uses the following call of `pytest`
https://github.com/microsoft/LightGBM/blob/5fe27d594285eeaf31d2f1ebfe29453f8c6c6ad3/.ci/test.sh#L151
We produce artifacts from QEMU job and hence set `bdist` task.
https://github.com/microsoft/LightGBM/blob/5fe27d594285eeaf31d2f1ebfe29453f8c6c6ad3/.vsts-ci.yml#L158

jmoralez:
`1.76s call     tests/python_package_test/test_dask.py::test_assign_open_ports_to_workers` 🤣
[logs](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=10656&view=logs&j=c2f9361f-3c13-57db-3206-cee89820d5e3&t=bf5b33f9-9072-549e-d8b1-cab79a1dd2a9&l=1617)

jameslamb:
Ha ok, thanks for looking into it!

I think that's strong evidence that #4498 was not the cause of timeouts on the QEMU builds. Could you please remove the changes to `test.sh`? Then I think we should still merge this (reducing number of checks from 1000 to 25) anyway, even if the impact on timing will be small.

jameslamb (approved):
thanks for the investigation!

I just observed another QEMU timeout on #4504 (https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=10669&view=logs&j=c2f9361f-3c13-57db-3206-cee89820d5e3).

If you have ideas from these timings on how we could reduce the runtime for those jobs without sacrificing too much test coverage, we'd welcome them!

StrikerRUS (approved):

jmoralez:
Some jobs just seem to take longer in all the tests, for example [here](https://dev.azure.com/lightgbm-ci/8461a79b-5dce-4085-ad70-4410b7135276/_apis/build/builds/10656/logs/279):
```
2021-08-06T18:36:11.2971161Z ../tests/python_package_test/test_basic.py ............................. [  5%]
2021-08-06T18:36:14.1860271Z ......................                                                   [  8%]
2021-08-06T18:38:54.7440546Z ../tests/python_package_test/test_consistency.py ......                  [  9%]
2021-08-06T18:42:39.7003057Z ../tests/python_package_test/test_dask.py .............................. [ 14%] 
```
2 minutes `test_basic` and 4 minutes `test_consistency`.
The [last timeout](https://dev.azure.com/lightgbm-ci/8461a79b-5dce-4085-ad70-4410b7135276/_apis/build/builds/10669/logs/279):
```
2021-08-09T02:45:03.2870447Z ../tests/python_package_test/test_basic.py ............................. [  5%]
2021-08-09T02:45:07.6525079Z ......................                                                   [  8%]
2021-08-09T02:59:01.0798201Z ../tests/python_package_test/test_consistency.py ......                  [  9%]
2021-08-09T03:04:47.1295927Z ../tests/python_package_test/test_dask.py .............................. [ 14%]
```
14 minutes `test_basic`, 6 minutes `test_consistency`.

PR #4198, opened 2021-04-19
[ci] parallelize R package installs in CI jobs

jameslamb:
This week, @jnolis taught me that the R function `install.packages()` can support parallel package installations! That function supports an argument `Ncpus`. If set to a value greater than 1, R will install multiple packages at the same time.

See https://stat.ethz.ch/R-manual/R-patched/library/utils/html/install.packages.html for more.

This PR proposes setting that argument to the value of `parallel::detectCores()` in this project's CI jobs and documentation of how to run them manually. I think this should reduce the time it takes R CI jobs to run, especially on Linux where CRAN does not prepare precompiled binaries. `{parallel}` comes installed in all standard installations of R, so this does not introduce a new dependency for LightGBM's CI jobs.

Windows and Linux runners from GitHub Actions have a single 2-core CPU and macOS runners have a single 3-core CPU: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources.

<hr>

### Testing whether this actually makes builds faster

The table below compares run times for the R jobs on the 3 most recent builds of `master` and 3 builds of this PR

* `master` build 1: https://github.com/microsoft/LightGBM/actions/runs/759494394
* `master` build 2: https://github.com/microsoft/LightGBM/actions/runs/759907488
* `master` build 3: https://github.com/microsoft/LightGBM/actions/runs/761721481
* build 1: https://github.com/microsoft/LightGBM/actions/runs/762022067
* build 2: https://github.com/microsoft/LightGBM/actions/runs/762081831
* build 3: https://github.com/microsoft/LightGBM/actions/runs/762122535

**Linux**

| job                                                                | master 1 | master 2 |master  3  | build 1 | build 2 | build 3 |
|:-----------------------------------------:|:--------:|:---------:|:---------:|:-------:|:------:|:-------:|
|ubuntu-latest, gcc, R 3.6, cmake                | 11m12s | 11m14s | 10m28s | 10m43s | **{08m32s}** | 10m01s|
|ubuntu-latest, gcc, R 4, cmake                   | 11m35s | 10m31s | 12m15s   | **{08m56s}** | 09m50s | 09m34s|
|ubuntu-latest, clang, R 3.6, cmake             | 11m18s | 11m02s | 12m06s | 09m02s | **{08m42s}** | 09m00s|
|ubuntu-latest, clang, R 4, cmake                | 09m50s | 10m22s | 09m59s| 09m05s | 08m53s | **{07m54s}**|
[...]

PR #4503, opened 2021-08-09
[ci] move Solaris and valgrind test steps into scripts

jameslamb:
#3946 introduces two new `Suggests`-level dependencies for the R package.

To run the Solaris and valgrind tests successfully, those dependencies need to be added to the jobs for those tests (https://github.com/microsoft/LightGBM/blob/master/.github/workflows/r_solaris.yml and https://github.com/microsoft/LightGBM/blob/master/.github/workflows/r_valgrind.yml).

This creates an interesting challenge. Right now changes like that need to be made in the workflow files in `.github/workflows/`, but when those jobs are triggered by comments they source those configs from `master` (https://github.com/microsoft/LightGBM/pull/4397#issuecomment-866790489).

This PR proposes moving more of the logic for those checks into scripts in `.ci/`. That will make it possible to test the following types of changes to the Solaris and valgrind jobs before merging them:

* adding or removing dependencies installed with `install.packages()`
* changing arguments to `install.packages()` (e.g. #4198)
* changing `apt-get install`'d dependencies for the Solaris checks
* changing arguments passed to `Rdvalgrind CMD INSTALL` in the valgrind checks

### DIFF
--- a/.ci/test_r_package_solaris.sh
+++ b/.ci/test_r_package_solaris.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+sh build-cran-package.sh || exit -1
+
+apt-get install --no-install-recommends -y \
+  libcurl4-openssl-dev \
+  libxml2-dev \
+  libssl-dev
+
+log_file="$GITHUB_WORKSPACE/rhub_logs.txt"
+Rscript -e "install.packages('rhub', dependencies = c('Depends', 'Imports', 'LinkingTo'), repos = 'https://cran.r-project.org', Ncpus = parallel::detectCores())" || exit -1
+
+Rscript $GITHUB_WORKSPACE/.ci/run_rhub_solaris_checks.R $(pwd)/lightgbm_*.tar.gz $log_file || exit -1
--- a/.ci/test_r_package_solaris.sh
+++ b/.ci/test_r_package_solaris.sh
@@ -7,7 +7,7 @@ apt-get install --no-install-recommends -y \
   libxml2-dev \
   libssl-dev
 
-log_file="$GITHUB_WORKSPACE/rhub_logs.txt"
+log_file="rhub_logs.txt"
 Rscript -e "install.packages('rhub', dependencies = c('Depends', 'Imports', 'LinkingTo'), repos = 'https://cran.r-project.org', Ncpus = parallel::detectCores())" || exit -1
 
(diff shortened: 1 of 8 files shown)


.ci/test_r_package_solaris.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+sh build-cran-package.sh || exit -1
+
+apt-get install --no-install-recommends -y \
+  libcurl4-openssl-dev \
+  libxml2-dev \
+  libssl-dev
+
+log_file="$GITHUB_WORKSPACE/rhub_logs.txt"
```
StrikerRUS, line 10: We try to not use CI service dependable env variables in bash script (try to search for `GITHUB_WORKSPACE`: it is used only in config `yml` files).

I guess we can switch to relative paths here just like in `test_r_package_valgrind.sh`
https://github.com/microsoft/LightGBM/blob/2f3764265cc2c43ffaa250e89c1f2efc859404ae/.ci/test_r_package_valgrind.sh#L5-L13

.github/workflows/r_solaris.yml
```diff
@@ -37,15 +37,7 @@ jobs:
...
-        run: |
-          sh build-cran-package.sh || exit -1
-          apt-get install --no-install-recommends -y \
-              libcurl4-openssl-dev \
-              libxml2-dev \
-              libssl-dev
-          log_file="$GITHUB_WORKSPACE/rhub_logs.txt"
-          Rscript -e "install.packages('rhub', dependencies = c('Depends', 'Imports', 'LinkingTo'), repos = 'https://cran.r-project.org', Ncpus = parallel::detectCores())"
-          Rscript $GITHUB_WORKSPACE/.ci/run_rhub_solaris_checks.R $(pwd)/lightgbm_*.tar.gz $log_file || exit -1
+        run: ./.ci/test_r_package_valgrind.sh
```
StrikerRUS, line 40:
```suggestion
        run: ./.ci/test_r_package_solaris.sh
```

StrikerRUS (changes requested):
Really useful refactoring! Thanks!

Please check my initial comments below.

.ci/test_r_package_solaris.sh
jameslamb: good point! fixed in https://github.com/microsoft/LightGBM/pull/4503/commits/10deff608ebcd95b60abd310472bad181d9768e2

StrikerRUS (approved):
LGTM! Let's check these changes in a next PR.

PR #4504, opened 2021-08-09
[R-package] use C++ compiler for pre-compile checks on Windows

jameslamb:
Looking into #4131 as part of #4310, I've noticed some other areas for possible improvement in the `configure` and `configure.win` scripts.

This PR fixes one in `configure.win`. I noticed that the mm_prefetch and mm_malloc tests are being run with a C compiler on Windows. They should be done with a C++ compiler, as they are on Linux/Mac

https://github.com/microsoft/LightGBM/blob/5fe27d594285eeaf31d2f1ebfe29453f8c6c6ad3/R-package/configure.ac#L41

and for all CMake-based builds

https://github.com/microsoft/LightGBM/blob/5fe27d594285eeaf31d2f1ebfe29453f8c6c6ad3/CMakeLists.txt#L253

I'm not aware of any problems caused by the fact that these checks were being done by a C compiler, but a C++ compiler is going to be used for compiling `lib_lightgbm`.

### Notes for Reviewers

To understand the difference between `CXXFLAGS` and `CPPFLAGS` and why they are both included, see [the documentation from `R CMD config`](https://svn.r-project.org/R/trunk/src/scripts/config).

> **CPPFLAGS**      C/C++ preprocessor flags, e.g. `-I<dir>` if you have headers in a nonstandard directory `<dir>`
> **CXXFLAGS**      compiler flags for CXX

### DIFF
--- a/R-package/configure.win
+++ b/R-package/configure.win
@@ -10,7 +10,7 @@ CXX11=`"${R_EXE}" CMD config CXX11`
 CXX11STD=`"${R_EXE}" CMD config CXX11STD`
 CXX="${CXX11} ${CXX11STD}"
 CXXFLAGS=`"${R_EXE}" CMD config CXX11FLAGS`
-CPPFLAGS=`"${R_EXE}" CMD config CPPFLAGS
+CPPFLAGS=`"${R_EXE}" CMD config CPPFLAGS`
 
 # LightGBM-specific flags
 LGB_CPPFLAGS=""
--- a/R-package/configure.win
+++ b/R-package/configure.win
@@ -6,7 +6,7 @@
 ###########################
 
 R_EXE="${R_HOME}/bin${R_ARCH_BIN}/R"
-CXX=`"${R_EXE}" CMD config CXX`
+CXX=`"${R_EXE}" CMD config CXX11`
 CXXFLAGS=`"${R_EXE}" CMD config CXXFLAGS`
 CPPFLAGS=`"${R_EXE}" CMD config CPPFLAGS`
 
--- a/R-package/configure.win
+++ b/R-package/configure.win
@@ -7,7 +7,9 @@
 
 R_SCRIPT="${R_HOME}/bin${R_ARCH_BIN}/Rscript"
 R_EXE="${R_HOME}/bin${R_ARCH_BIN}/R"
-CC=`"${R_EXE}" CMD config CC`
+CXX=`"${R_EXE}" CMD config CXX`
+CXXFLAGS=`"${R_EXE}" CMD config CXXFLAGS`
+CPPFLAGS=`"${R_EXE}" CMD config CPPFLAGS`
 
 # LightGBM-specific flags
 LGB_CPPFLAGS=""
@@ -24,7 +26,7 @@ LGB_CPPFLAGS="${LGB_CPPFLAGS} -DEIGEN_MPL2_ONLY"
 
 ac_mm_prefetch="no"
 
-cat > conftest.c <<EOL
+cat > conftest.cpp <<EOL
 #include <xmmintrin.h>
 int main() {
   int a = 0;
@@ -33,7 +35,7 @@ int main() {
 }
 EOL
 
-${CC} -o conftest conftest.c 2>/dev/null && ./conftest && ac_mm_prefetch="yes"
+${CXX} ${CXXFLAGS} ${CPPFLAGS} -o conftest conftest.cpp 2>/dev/null && ./conftest && ac_mm_prefetch="yes"
 echo "checking whether MM_PREFETCH works...${ac_mm_prefetch}"
 
 if test "${ac_mm_prefetch}" = "yes";
@@ -46,7 +48,7 @@ fi
 ############
 ac_mm_malloc="no"
 
-cat > conftest.c <<EOL
+cat > conftest.cpp <<EOL
 #include <mm_malloc.h>
 int main() {
     char *a = (char*)_mm_malloc(8, 16);
@@ -55,7 +57,7 @@ int main() {
 }
 EOL
 
-${CC} -o conftest conftest.c 2>/dev/null && ./conftest && ac_mm_malloc="yes"
+${CXX} ${CXXFLAGS} ${CPPFLAGS} -o conftest conftest.cpp 2>/dev/null && ./conftest && ac_mm_malloc="yes"
 echo "checking whether MM_MALLOC works...${ac_mm_malloc}"
 
 if test "${ac_mm_malloc}" = "yes";
--- a/R-package/configure.win
+++ b/R-package/configure.win
@@ -6,9 +6,11 @@
 ###########################
 
 R_EXE="${R_HOME}/bin${R_ARCH_BIN}/R"
-CXX=`"${R_EXE}" CMD config CXX11`
-CXXFLAGS=`"${R_EXE}" CMD config CXXFLAGS`
-CPPFLAGS=`"${R_EXE}" CMD config CPPFLAGS`
+CXX11=`"${R_EXE}" CMD config CXX11`
+CXX11STD=`"${R_EXE}" CMD config CXX11STD`
+CXX="${CXX11} ${CXX11STD}"
+CXXFLAGS=`"${R_EXE}" CMD config CXX11FLAGS`
+CPPFLAGS=`"${R_EXE}" CMD config CPPFLAGS
 
 # LightGBM-specific flags
 LGB_CPPFLAGS=""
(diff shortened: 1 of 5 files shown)


jameslamb:
I'm seeing several errors in CI today about connecting to GitHub.

For example, from an R package job: https://github.com/microsoft/LightGBM/pull/4504/checks?check_run_id=3282998279

```text
  "C:\Program Files\Git\bin\git.exe" -c protocol.version=2 fetch --no-tags --prune --progress --no-recurse-submodules --depth=5 origin +d08d1f3c51a3f4876ca39d591097269b68cc9bc3:refs/remotes/pull/4504/merge
  Error: fatal: unable to access 'https://github.com/microsoft/LightGBM/': Failed to connect to github.com port 443: Timed out
  The process 'C:\Program Files\Git\bin\git.exe' failed with exit code 128
  Waiting 11 seconds before trying again
  "C:\Program Files\Git\bin\git.exe" -c protocol.version=2 fetch --no-tags --prune --progress --no-recurse-submodules --depth=5 origin +d08d1f3c51a3f4876ca39d591097269b68cc9bc3:refs/remotes/pull/4504/merge
  Error: fatal: unable to access 'https://github.com/microsoft/LightGBM/': Failed to connect to github.com port 443: Timed out
  The process 'C:\Program Files\Git\bin\git.exe' failed with exit code 128
  Waiting 14 seconds before trying again
  "C:\Program Files\Git\bin\git.exe" -c protocol.version=2 fetch --no-tags --prune --progress --no-recurse-submodules --depth=5 origin +d08d1f3c51a3f4876ca39d591097269b68cc9bc3:refs/remotes/pull/4504/merge
  Error: fatal: unable to access 'https://github.com/microsoft/LightGBM/': Failed to connect to github.com port 443: Timed out
  Error: The process 'C:\Program Files\Git\bin\git.exe' failed with exit code 128
```

And a Python-package job: https://github.com/microsoft/LightGBM/pull/4504/checks?check_run_id=3282997602

```text
  Submodule path 'external_libs/fast_double_parser': checked out 'ace60646c02dc54c57f19d644e49a61e7e7758ec'
  Error: fatal: unable to access 'https://github.com/fmtlib/fmt.git/': Failed to connect to github.com port 443: Operation timed out
  Fetched in submodule path 'external_libs/fmt', but it did not contain cc09f1a6798c085c325569ef466bcdcffdc266d4. Direct fetching of that commit failed.
  Error: The process '/usr/local/bin/git' failed with exit code 1
```

It does not seem like there were any incidents on GitHub today (https://www.githubstatus.com/history). Azure is not reporting any recent incidents either (https://status.azure.com/en-us/status/history/).

I'm not sure what's happening :/

jameslamb:
Solaris checks are failing with the following error

> ERROR: dependency ‘R6’ is not available for package ‘lightgbm’

and

> downloaded 59 KB
trying URL 'https://cloud.r-project.org/src/contrib/praise_1.0.0.tar.gz'
cannot open URL 'https://cloud.r-project.org/src/contrib/praise_1.0.0.tar.gz'
1: In download.file(url, destfile, method, mode = "wb", ...) :
downloaded length 0 != reported length 63
2: In download.file(url, destfile, method, mode = "wb", ...) :
cannot open URL 'https://cloud.r-project.org/src/contrib/praise_1.0.0.tar.gz': HTTP status was '503 Service Unavailable'
Warning in download.packages(pkgs, destdir = tmpd, available = available, :
download of package ‘praise’ failed

and

> Error: Bioconductor version cannot be validated; no internet connection?
1: In file(con, "r") :
cannot open URL 'https://bioconductor.org/config.yaml': HTTP status was '503 Service Unavailable'
2: In file(con, "r") :
cannot open URL 'http://bioconductor.org/config.yaml': HTTP status was '503 Service Unavailable'
Execution halted
ERROR: lazy loading failed for package ‘R6’

and

> ERROR: dependency ‘R6’ is not available for package ‘processx’

<hr>

I'm going to run it again, and if we see similar failures I'll report them at https://github.com/r-hub/rhub.

jameslamb:
Just pushed https://github.com/microsoft/LightGBM/pull/4504/commits/4884a695355574db9cece6c22ca1fe6806154c35, adding `{Matrix}` to the setup for the valgrind tests. It is an `Imports`-level dependency so it should be installed. Maybe one of the other dependencies was introducing it before and that's why it was ok for us to not explicitly `install.packages()` it.

jameslamb:
Installing `{Matrix}` seemed to fix the failures in the `valgrind` job!

https://github.com/microsoft/LightGBM/pull/4504#issuecomment-896043897

StrikerRUS:
Should we add `Matrix` in other places for the consistency, e.g. in config for sanitizers or debian clang?

jameslamb:
> Should we add `Matrix` in other places for the consistency, e.g. in config for sanitizers or debian clang?

I think so, yeah. I'll be away from my computer for a few hours. You can push a commit here directly if you'd like, otherwise I'll do it later tonight.

jameslamb:
Added `{Matrix}` explicitly in the other places where we install R dependencies: https://github.com/microsoft/LightGBM/pull/4504/commits/b43b6c71d6b01e738b71aaad44aaa759f026ee7c

R-package/configure.win
```diff
@@ -6,7 +6,9 @@
 ###########################
 
 R_EXE="${R_HOME}/bin${R_ARCH_BIN}/R"
-CC=`"${R_EXE}" CMD config CC`
```
StrikerRUS, line 9: Why do we keep `CC` for Unix (considering #4506 as well)?
https://github.com/microsoft/LightGBM/blob/b43b6c71d6b01e738b71aaad44aaa759f026ee7c/R-package/configure.ac#L23

R-package/configure.win
```diff
@@ -6,7 +6,9 @@
...
+CXX=`"${R_EXE}" CMD config CXX`
```
StrikerRUS, line 9: Why do we use `CXX11` for Unix and `CXX` here to configure `CXX` variable (considering #4506 as well)??
https://github.com/microsoft/LightGBM/blob/b43b6c71d6b01e738b71aaad44aaa759f026ee7c/R-package/configure.ac#L24

StrikerRUS (commented):
Good sign!
```
checking whether MM_PREFETCH works...yes
checking whether MM_MALLOC works...yes
```
Please check two my questions about consistency with Unix version of `configure` below

R-package/configure.win
jameslamb: I wanted to keep this PR's scope limited to Windows.

A C compiler is only used on the unix-alike side (`configure.ac`) for the OpenMP test, but #4507 proposes changing that to a C++ compiler.

R-package/configure.win
jameslamb: ah! thanks for noticing, that is an oversight. It should be CXX11 since we explicitly set a C++ standard

https://github.com/microsoft/LightGBM/blob/926526c838196a7497a85b6b8cf07657a88b69e6/R-package/src/Makevars.win.in#L1

From https://cran.r-project.org/doc/manuals/R-exts.html#Configure-and-cleanup

>  If the package specifies a non-default C++ standard, use the config variable names (such as `CXX17`) appropriate to the standard

From R's perspective, our use of C++ is "non-default". See https://cran.r-project.org/doc/manuals/R-exts.html#Using-C_002b_002b-code.

> As from R 4.0.0 a C++ compiler will be selected only if it conforms to the 2011 standard (‘C++11’). A minor update43 (‘C++14’) was published in December 2014 and will be used by default as from R 4.1.0 if supported

updated in https://github.com/microsoft/LightGBM/pull/4504/commits/b576a4f580f6b16f75a77462fb688c5df1b942ca

R-package/configure.win
jameslamb: Based on this comment thread, I just removed the changes for OpenMP checks from #4506, so now there won't be overlap between this PR, #4506, and #4507.

R-package/configure.win
StrikerRUS: Seems we are setting less variables than it's done in the example from https://cran.r-project.org/doc/manuals/R-exts.html#Using-C_002b_002b-code.
```
CXX11=`"${R_EXE}" CMD config CXX11`
CXX11STD=`"${R_EXE}" CMD config CXX11STD`
CXX="${CXX11} ${CXX11STD}"
CXXFLAGS=`"${R_EXE}" CMD config CXX11FLAGS`
CPPFLAGS=`"${R_EXE}" CMD config CPPFLAGS`
```

Also note that `CXX11FLAGS` is used to configure `CXXFLAGS`.

StrikerRUS: I was able to find a similar code in the wild by searching for `CXX11FLAGS filename:configure.ac`. For example,
https://github.com/cbg-ethz/MC-CBN/blob/6cd55b8289d22809a5029981f87cd1cbe5be27ea/configure.ac#L21-L38
https://github.com/cran/spp/blob/2603673566115871405f5842e489454805b9f5c0/configure.ac#L15-L22
https://github.com/akhikolla/ClusterTests/blob/1c7702ba4035511ffe7d4dd27e9021b6962718fe/BART/configure.ac#L19-L22

jameslamb: oooo interesting, I missed that! Thanks for pointing it out. I just made those changes here and on #4506 .

StrikerRUS (approved):
Thank you very much for this PR! LGTM!

PR #4506, opened 2021-08-09
[R-package] pass R-configured compiler flags to checks in configure

jameslamb:
Investigating #4131 , I've been re-reading [the section of "Writing R Extensions" that discusses `configure` scripts](https://cran.r-project.org/doc/manuals/R-exts.html#Configure-and-cleanup).

I believe the R package's current `configure` script is not currently following some of that manual's advice. When it does configure tests (e.g. to check for an included `mm_malloc`), it isn't passing through the compiler flags that R will use when compiling `lib_lightgbm`.

> You will very likely need to ensure that the same C compiler and compiler flags are used in the configure tests as when compiling R or your package

> You can use `R CMD config` to get the value of the basic configuration variables... If you do, it is essential that you use both the command and the appropriate flags, so that for example `CC` must always be used with `CFLAGS`

> Packages written with C++ need to pick up the details for the C++ compiler and switch the current language to C++ 

This PR ensures that the relevant flags are passed through to those checks.

### DIFF
--- a/R-package/configure.ac
+++ b/R-package/configure.ac
@@ -21,7 +21,10 @@ if test -z "${R_HOME}"; then
     exit 1
 fi
 CC=`"${R_HOME}/bin/R" CMD config CC`
+CFLAGS=`"${R_HOME}/bin/R" CMD config CFLAGS`
 CXX=`"${R_HOME}/bin/R" CMD config CXX11`
+CPPFLAGS=`"${R_HOME}/bin/R" CMD config CPPFLAGS`
+CXXFLAGS=`"${R_HOME}/bin/R" CMD config CXXFLAGS`
 
 # LightGBM-specific flags
 LGB_CPPFLAGS=""
@@ -53,7 +56,7 @@ AC_LANG_CONFTEST(
         )
     ]
 )
-${CXX} -o conftest conftest.cpp 2>/dev/null && ./conftest && ac_mmprefetch=yes
+${CXX} ${CPPFLAGS} ${CXXFLAGS} -o conftest conftest.cpp 2>/dev/null && ./conftest && ac_mmprefetch=yes
 AC_MSG_RESULT([${ac_mmprefetch}])
 if test "${ac_mmprefetch}" = yes; then
     LGB_CPPFLAGS+=" -DMM_PREFETCH=1"
@@ -80,7 +83,7 @@ AC_LANG_CONFTEST(
         )
     ]
 )
-${CXX} -o conftest conftest.cpp 2>/dev/null && ./conftest && ac_mm_malloc=yes
+${CXX} ${CPPFLAGS} ${CXXFLAGS} -o conftest conftest.cpp 2>/dev/null && ./conftest && ac_mm_malloc=yes
 AC_MSG_RESULT([${ac_mm_malloc}])
 if test "${ac_mm_malloc}" = yes; then
     LGB_CPPFLAGS+=" -DMM_MALLOC=1"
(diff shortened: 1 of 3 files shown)


jameslamb:
Ah I forgot, `valgrind` tests here will fail until we merge #4504 (https://github.com/microsoft/LightGBM/pull/4504#issuecomment-896442336).

I'll just add `{Matrix}` here as well (only for the `valgrind` test) so this PR can be tested without waiting on #4504 to be merge.

jameslamb:
triggering new `valgrind` and `solaris` checks now that I've merged #4504 into `master` and `master` into this branch

R-package/configure.ac
```diff
@@ -21,7 +21,11 @@ if test -z "${R_HOME}"; then
     exit 1
 fi
 CC=`"${R_HOME}/bin/R" CMD config CC`
-CXX=`"${R_HOME}/bin/R" CMD config CXX11`
+CXX11=`"${R_HOME}/bin/R" CMD config CXX11`
+CXX11STD=`"${R_HOME}/bin/R" CMD config CXX11STD`
+CXX="${CXX11} ${CXX11STD}"
+CPPFLAGS=`"${R_HOME}/bin/R" CMD config CPPFLAGS`
+CXXFLAGS=`"${R_HOME}/bin/R" CMD config CXX11FLAGS`
```
StrikerRUS, line 28: I believe we should add the following line here (and in `configure.win` as well btw) according to the example in https://cran.r-project.org/doc/manuals/R-exts.html#Configure-and-cleanup.
```
Packages written with C++ need to pick up the details for the C++ compiler and switch the current language to C++ by something like

CXX=`"${R_HOME}/bin/R" CMD config CXX`
if test -z "$CXX"; then
  AC_MSG_ERROR([No C++ compiler is available])
fi
CXXFLAGS=`"${R_HOME}/bin/R" CMD config CXXFLAGS`
CPPFLAGS=`"${R_HOME}/bin/R" CMD config CPPFLAGS`
AC_LANG(C++)
```
The same line appears in https://cran.r-project.org/doc/manuals/R-exts.html#Using-C_002b_002b-code.


```suggestion
CXXFLAGS=`"${R_HOME}/bin/R" CMD config CXX11FLAGS`
AC_LANG(C++)
```

jameslamb: `AC_LANG(C++)` is already included in all the relevant places in `configure.ac`.

https://github.com/microsoft/LightGBM/blob/86ead2050c006b984e88396e0e380304bac114d1/R-package/configure.ac#L41

https://github.com/microsoft/LightGBM/blob/86ead2050c006b984e88396e0e380304bac114d1/R-package/configure.ac#L68

I guess it could be pulled up to this line to slightly reduce duplication, but that won't change the behavior of `configure` at all.

I don't think it's appropriate for `configure.win`. `AC_LANG` is an `autoconf` directive ([see here](https://www.gnu.org/software/autoconf/manual/autoconf-2.63/html_node/Language-Choice.html)), and `configure.win` is not used as an input to `autoconf` the way `configure.ac` is. `configure.win` is a regular shell script.

jameslamb: I've done that consolidation in `configure.ac` in https://github.com/microsoft/LightGBM/pull/4506/commits/70fcec0224d409520ed63fc4b7d77097a5b903ab.

StrikerRUS (approved):
Great job, thanks!

R-package/configure.ac
StrikerRUS: > I don't think it's appropriate for `configure.win`

Ah, indeed, you're right!

jameslamb:
QEMU build hit the 3-hour timeout

https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=10728&view=results

image

Here's how far it made it through the tests:

```text
Successfully installed lightgbm-3.2.1.99
WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv
platform linux -- Python 3.9.6, pytest-6.2.4, py-1.10.0, pluggy-0.13.1
rootdir: /LightGBM
collected 618 items

../tests/c_api_test/test_.py ..                                          [  0%]
../tests/python_package_test/test_basic.py ............................. [  5%]
......................                                                   [  8%]
../tests/python_package_test/test_consistency.py ......                  [  9%]
../tests/python_package_test/test_dask.py .............................. [ 14%]
........................................................................ [ 26%]
........................................................................ [ 37%]
......s...............s...............s...............s................. [ 49%]
##[error]The operation was canceled.
Finishing: Setup and run tests
```

I'm going to re-run it.

StrikerRUS:
> Here's how far it made it through the tests:

Just 49%? 😨

PR #4507, opened 2021-08-09
[R-package] fix OpenMP checking on macOS (fixes #4131)

### LINKED ISSUE #4131: [R Package] autoconf.ac OpenMP test ignores LDFLAGS (fix provided).

jameslamb:
Fixes #4131.

It makes a few improvements to the way that CRAN-style builds of the R package's `configure` script checks for the presence of OpenMP on macOS.

As a result, **OpenMP detection in the R package should be more reliable, meaning more users can benefit from multi-threading speedups**.

changes:

* passes the result of `R CMD config LDFLAGS` when running the OpenMP check, to ensure that that check uses the same linker flags that R itself will use when compiling `lib_lightgbm`
    - notably, this puts `-L/usr/local/lib` into the statement used to compile a test program, so brew-installed `libomp` will be found
    - this also allows users to install `{lightgbm}` with non-standard install locations for `libomp` by modifying `LDFLAGS` in `~/.R/Makevars`
* adds an additional check using only `-fopenmp` if compilation with `-Xclang -fopenmp` fails
    - this makes OpenMP detection work when compiling a CRAN package using `g++` (on `master`, it fails with a complaint that flag `-Xclang` is not recognized)
    - see https://stackoverflow.com/a/12002492/3986677
* compiles the test program with a C++ compiler instead of a C compiler

### Notes for Reviewers

CRAN prepares precompiled binaries for macOS, so this change should only affect people building from source using the repo, or choosing to do custom compilation from CRAN with something like `install.packages("lightgbm", type = "source")`.

As noted above, this should also help with detection of OpenMP when not using `clang` on Mac.

<details><summary>click here to learn how I tested using g++ on Mac</summary>

I have OpenMP at the following loccations on my laptop.

```shell
find /usr -name 'libomp.dylib'
```

```text
/usr/local/lib/libomp.dylib
/usr/local/Cellar/libomp/12.0.0/lib/libomp.dylib
```

My `~/.R/Makevars` contains only the following.

```shell
CXX=/usr/local/bin/g++-11
CXX11=/usr/local/bin/g++-11
CC=/usr/local/bin/gcc-11
```

When I build the CRAN package from source on `master`, OpenMP isn't found.

```shell
Rscript -e "remove.packages('lightgbm')"
sh build-cran-package.sh
R CMD INSTALL lightgbm_3.2.1.99.tar.gz > out.txt 2>&1
cat out.txt | grep -E "checking whether OpenMP"
```

> checking whether OpenMP will work in a package... no

On this PR, it's found.

</details>

### DIFF
--- a/R-package/configure
+++ b/R-package/configure
@@ -1695,6 +1695,8 @@ if test -z "${R_HOME}"; then
 fi
 CC=`"${R_HOME}/bin/R" CMD config CC`
 CXX=`"${R_HOME}/bin/R" CMD config CXX11`
+CPPFLAGS=`"${R_HOME}/bin/R" CMD config CPPFLAGS`
+CXXFLAGS=`"${R_HOME}/bin/R" CMD config CXXFLAGS`
 LDFLAGS=`"${R_HOME}/bin/R" CMD config LDFLAGS`
 
 # LightGBM-specific flags
@@ -1801,6 +1803,7 @@ fi
 
 if test `uname -s` = "Darwin"
 then
+    OPENMP_CXXFLAGS='-Xclang -fopenmp'
     OPENMP_LIB='-lomp'
     ac_pkg_openmp=no
     { $as_echo "$as_me:${as_lineno-$LINENO}: checking whether OpenMP will work in a package" >&5
@@ -1830,17 +1833,17 @@ main ()
 
 
 _ACEOF
+    ${CXX} ${CPPFLAGS} ${CXXFLAGS} ${LDFLAGS} ${OPENMP_CXXFLAGS} ${OPENMP_LIB} -o conftest conftest.cpp 2>/dev/null && ./conftest && ac_pkg_openmp=yes
 
-    # test without any additional flags
-    ${CXX} ${LDFLAGS} ${OPENMP_CXXFLAGS} ${OPENMP_LIB} -o conftest conftest.cpp 2>/dev/null && ./conftest && ac_pkg_openmp=yes
+    # -Xclang is not portable (it is clang-specific)
+    # if compilation above failed, try without that flag
     if test "${ac_pkg_openmp}" = no; then
         if test -f "./conftest"; then
             rm ./conftest
         fi
-        OPENMP_CXXFLAGS="-Xclang -fopenmp"
-        ${CXX} ${LDFLAGS} ${OPENMP_CXXFLAGS} ${OPENMP_LIB} -o conftest conftest.cpp 2>/dev/null && ./conftest && ac_pkg_openmp=yes
+        OPENMP_CXXFLAGS="-fopenmp"
+        ${CXX} ${CPPFLAGS} ${CXXFLAGS} ${LDFLAGS} ${OPENMP_CXXFLAGS} ${OPENMP_LIB} -o conftest conftest.cpp 2>/dev/null && ./conftest && ac_pkg_openmp=yes
     fi
-    echo "${OPENMP_CXXFLAGS}" > /Users/jlamb/Desktop/flags.txt
 
     { $as_echo "$as_me:${as_lineno-$LINENO}: result: ${ac_pkg_openmp}" >&5
 $as_echo "${ac_pkg_openmp}" >&6; }
--- a/R-package/configure
+++ b/R-package/configure
@@ -1693,10 +1693,11 @@ if test -z "${R_HOME}"; then
     echo "could not determine R_HOME"
     exit 1
 fi
-CC=`"${R_HOME}/bin/R" CMD config CC`
-CXX=`"${R_HOME}/bin/R" CMD config CXX11`
+CXX11=`"${R_HOME}/bin/R" CMD config CXX11`
+CXX11STD=`"${R_HOME}/bin/R" CMD config CXX11STD`
+CXX="${CXX11} ${CXX11STD}"
 CPPFLAGS=`"${R_HOME}/bin/R" CMD config CPPFLAGS`
-CXXFLAGS=`"${R_HOME}/bin/R" CMD config CXXFLAGS`
+CXXFLAGS=`"${R_HOME}/bin/R" CMD config CXX11FLAGS`
 LDFLAGS=`"${R_HOME}/bin/R" CMD config LDFLAGS`
 
 # LightGBM-specific flags
--- a/R-package/configure
+++ b/R-package/configure
@@ -1695,6 +1695,7 @@ if test -z "${R_HOME}"; then
 fi
 CC=`"${R_HOME}/bin/R" CMD config CC`
 CXX=`"${R_HOME}/bin/R" CMD config CXX11`
+LDFLAGS=`"${R_HOME}/bin/R" CMD config LDFLAGS`
 
 # LightGBM-specific flags
 LGB_CPPFLAGS=""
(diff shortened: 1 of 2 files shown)


jameslamb:
The `valgrind` test is failing on complaints about a missing dependency

> * installing to library ‘/usr/local/RDvalgrind/lib/R/site-library’
ERROR: dependency ‘Matrix’ is not available for package ‘lightgbm’

PROBABLY not related to this PR's changes, but it'll be easy enough to test after #4503 is merged.

jameslamb:
Solaris checks are failing with the following error

> ERROR: dependency ‘R6’ is not available for package ‘lightgbm’

and

> downloaded 59 KB
trying URL 'https://cloud.r-project.org/src/contrib/praise_1.0.0.tar.gz'
cannot open URL 'https://cloud.r-project.org/src/contrib/praise_1.0.0.tar.gz'
1: In download.file(url, destfile, method, mode = "wb", ...) :
downloaded length 0 != reported length 63
2: In download.file(url, destfile, method, mode = "wb", ...) :
cannot open URL 'https://cloud.r-project.org/src/contrib/praise_1.0.0.tar.gz': HTTP status was '503 Service Unavailable'
Warning in download.packages(pkgs, destdir = tmpd, available = available, :
download of package ‘praise’ failed

and

> Error: Bioconductor version cannot be validated; no internet connection?
1: In file(con, "r") :
cannot open URL 'https://bioconductor.org/config.yaml': HTTP status was '503 Service Unavailable'
2: In file(con, "r") :
cannot open URL 'http://bioconductor.org/config.yaml': HTTP status was '503 Service Unavailable'
Execution halted
ERROR: lazy loading failed for package ‘R6’

and

> ERROR: dependency ‘R6’ is not available for package ‘processx’

<hr>

I'm going to run it again, and if we see similar failures I'll report them at https://github.com/r-hub/rhub.

jameslamb:
> I'm going to run it again, and if we see similar failures I'll report them at https://github.com/r-hub/rhub.

I decided not to run these again, since other PRs will need to get merged before this one anyway (like #4506).

@StrikerRUS I see you're starting to merge things now that the blocking CI issue (#4524) is fixed. Don't worry about updating this PR. I can do that tonight, since it might require resolving conflicts with #4506.

StrikerRUS (approved):
Makes sense to me, LGTM!
Thanks a lot for great improvements in `configure.ac`!

R-package/configure
```diff
@@ -1828,7 +1822,18 @@ main ()
 
 
 _ACEOF
-    ${CC} -o conftest conftest.c ${OPENMP_LIB} ${OPENMP_CXXFLAGS} 2>/dev/null && ./conftest && ac_pkg_openmp=yes
+    ${CXX} ${CPPFLAGS} ${CXXFLAGS} ${LDFLAGS} ${OPENMP_CXXFLAGS} ${OPENMP_LIB} -o conftest conftest.cpp 2>/dev/null && ./conftest && ac_pkg_openmp=yes
+
+    # -Xclang is not portable (it is clang-specific)
+    # if compilation above failed, try without that flag
```
StrikerRUS, line 1828: @jameslamb Sorry for commenting in closed PR.
I just thought about this a little bit.
This if branch is mostly for "not AppleClang compiler", right? Then why do we test only `-fopenmp` flag? Maybe it's worth checking some other flags, e.g. `-qopenmp` and `-openmp` for Intel compiler?
Possible values: https://github.com/autoconf-archive/autoconf-archive/blob/6c25681d78d963708362ea0d4d63ee9b508782b6/m4/ax_openmp.m4#L82

https://www.gnu.org/software/autoconf-archive/ax_openmp.html

jameslamb: oooo interesting! Yes, this branch is for "not clang", and was inspired by what I saw other `data.table` doing (since I know those maintainers have spent a lot of time struggling with OpenMP detection in R packages).

For example, 

https://github.com/Rdatatable/data.table/blob/4cf92893742a0359ce3b3ddfd279a9e69627d52b/configure#L91

```shell
# Compilation failed -- try forcing -fopenmp instead.
R_NO_OPENMP=0
"${CC}" "${CFLAGS}" -fopenmp test-omp.c || R_NO_OPENMP=1
```

I didn't know about these other flags that you've mentioned, but seems like a good idea to try for even more portability! If you open an issue listing those other flags (with the links you've included here) and and which compilers they are recognized by, I'd be happy to work on adding and testing that support.

StrikerRUS: > If you open an issue ...

Sure, just created #4537.

PR #4510, opened 2021-08-09
[python] Use double type for `init_score` array when set by predictor

StrikerRUS:
I'm not sure but looks like `init_score` is supposed to be only `double` type.
-  `LGBM_BoosterPredictFor*` functions return predictions in double type
- https://lightgbm.readthedocs.io/en/latest/C-API.html#c.LGBM_DatasetSetField
   > *init_score* only works for `C_API_DTYPE_FLOAT64`
- https://github.com/microsoft/LightGBM/blob/cfe8eb17c9f104a1681de85839f9fe6bd4d6692b/python-package/lightgbm/basic.py#L446
- https://github.com/microsoft/LightGBM/blob/cfe8eb17c9f104a1681de85839f9fe6bd4d6692b/python-package/lightgbm/basic.py#L1919-L1920

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -1394,7 +1394,7 @@ class Dataset:
             if used_indices is not None:
                 assert not self.need_slice
                 if isinstance(data, (str, Path)):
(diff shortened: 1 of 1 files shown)


jameslamb (approved):
makes sense to me

Issue #4484, opened 2021-07-22
Python wheels not PEP 599 compliant; wheels should include libgomp.so.1


AWSjswinney:
## Description
PEP 599 and PEP 513 describe a spec for broadly compatible wheels. They include a list of dynamic libraries that wheels are allowed to dynamically link against without bundling the `so` in the wheel. The LightGBM wheel depends on `libgomp.so.1` but does not include in the wheel package. To fix this, you can use [auditwheel](https://github.com/pypa/auditwheel). I have not tested this, but it probably involves adding line like this, [here](https://github.com/microsoft/LightGBM/blob/48257d45e577df69abbc86542d0515159e524eb2/.ci/test.sh#L145).
```
auditwheel repair dist/lightgbm-$LGB_VER-py3-none-$PLATFORM.whl
```

Specs:
[PEP 599 -- The manylinux2014 Platform Tag](https://www.python.org/dev/peps/pep-0599/)
[PEP 513 -- A Platform Tag for Portable Linux Built Distributions](https://www.python.org/dev/peps/pep-0513/)

## Reproducible example
This works on Linux on `x86_64` and `aarch64`.
```shell
docker run -it --rm amazonlinux sh -c "yum install -y python3-pip && pip3 install lightgbm && python3 -c 'import lightgbm'"
```

```
...
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/local/lib/python3.7/site-packages/lightgbm/__init__.py", line 8, in <module>
    from .basic import Booster, Dataset, register_logger
  File "/usr/local/lib/python3.7/site-packages/lightgbm/basic.py", line 95, in <module>
    _LIB = _load_lib()
  File "/usr/local/lib/python3.7/site-packages/lightgbm/basic.py", line 86, in _load_lib
    lib = ctypes.cdll.LoadLibrary(lib_path[0])
  File "/usr/lib64/python3.7/ctypes/__init__.py", line 437, in LoadLibrary
    return self._dlltype(name)
  File "/usr/lib64/python3.7/ctypes/__init__.py", line 359, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: libgomp.so.1: cannot open shared object file: No such file or directory
```

## Environment info

LightGBM version or commit hash: lightgbm-3.2.1

Command(s) you used to install LightGBM

```shell
pip3 install lightgbm
```

jameslamb:
[...]

PR #3378, opened 2020-09-11
[docs] Simplify the python installation instruction

guolinke:

PR #4517, opened 2021-08-12
[docs] add information about installing libgomp

AWSjswinney:
On the python instructions, add information about installing libgomp
because it may not be installed by default and is required for
lightgbm to run.

This is to address #4484

### DIFF
--- a/docs/Python-Intro.rst
+++ b/docs/Python-Intro.rst
@@ -28,6 +28,23 @@ To verify your installation, try to ``import lightgbm`` in Python:
 
     import lightgbm as lgb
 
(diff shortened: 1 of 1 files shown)


StrikerRUS:
@AWSjswinney Thanks a lot for getting back to this issue!

We've recently moved all prerequisites to the installation guide to avoid a duplication and make intro guide clearer: #3378.

Also, this important note about libomp should be in the CLI version installation guide, not only in the Python one. I'll prepare a new PR today and ask you to review with the aim to not waste your time cleaning our docs.
> Also, if I find the the time, I may test and submit a pull request myself.
    https://github.com/microsoft/LightGBM/issues/4484#issuecomment-887731216

AWSjswinney:
Closed in favor of #4520

PR #4519, opened 2021-08-14
[ci] replace uses of backticks in test.sh with $()

jameslamb:
I've been experimenting with [`shellcheck`](https://github.com/koalaman/shellcheck) recently in some projects.

I ran it over all of this project's shell scripts tonight, like this:

```shell
brew install shellcheck
shell_scripts=$(
    git ls-files \
    | grep -E "\.sh$"
)

shellcheck \
    --color=never \
    --format=gcc \
    $shell_scripts
```

This PR proposes fixing one type of warning raised by `shellcheck`. 

```text
.ci/test.sh:43:31: note: Use $(...) notation instead of legacy backticked `...`. [SC2006]
.ci/test.sh:45:86: note: Use $(...) notation instead of legacy backticked `...`. [SC2006]
```

See https://github.com/koalaman/shellcheck/wiki/SC2006 for some background on why `$()` should be preferred to backticks for inline expressions.

### DIFF
--- a/.ci/test.sh
+++ b/.ci/test.sh
@@ -40,9 +40,9 @@ if [[ $TASK == "check-docs" ]] || [[ $TASK == "check-links" ]]; then
     pip install --user -r requirements.txt
     # check reStructuredText formatting
     cd $BUILD_DIRECTORY/python-package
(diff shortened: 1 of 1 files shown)


StrikerRUS (approved):
Thanks!

PR #4520, opened 2021-08-14
[docs] Add notes in installation guide, including ones about OpenMP

### LINKED ISSUE #4484: Python wheels not PEP 599 compliant; wheels should include libgomp.so.1

StrikerRUS:
Supersedes #4517 (https://github.com/microsoft/LightGBM/pull/4517#issuecomment-898448540).
Closes #4484.

### DIFF
--- a/docs/Installation-Guide.rst
+++ b/docs/Installation-Guide.rst
@@ -129,6 +129,8 @@ On Linux LightGBM can be built using **CMake** and **gcc** or **Clang**.
 
 **Note**: glibc >= 2.14 is required.
 
+**Note**: In some rare cases you may need to install OpenMP runtime library separately (use your package manager and search for ``lib[g|i]omp`` for doing this).
+
 Also, you may want to read `gcc Tips <./gcc-Tips.rst>`__.
 
 macOS
@@ -437,6 +439,8 @@ On Linux an MPI version of LightGBM can be built using **Open MPI**, **CMake** a
 
 **Note**: glibc >= 2.14 is required.
 
+**Note**: In some rare cases you may need to install OpenMP runtime library separately (use your package manager and search for ``lib[g|i]omp`` for doing this).
+
 macOS
 ^^^^^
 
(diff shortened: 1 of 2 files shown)


StrikerRUS:
@AWSjswinney Will really appreciate your review!

jameslamb (approved):
Good clarification, thanks.

AWSjswinney (approved):
Looks good to me. I will close my own PR, #4517

StrikerRUS:
Many thanks for your reviews guys!

PR #4522, opened 2021-08-15
[R-package] add deprecation warnings about some uses of '...'

jameslamb:
Contributes to #4310 (based on https://github.com/microsoft/LightGBM/issues/4226#issuecomment-829473584).

This PR proposes adding some deprecation warnings on uses of `...` for parameters in the following functions:

* `dim.lgb.Dataset()`
* `getinfo.lgb.Dataset()`
* `lgb.cv()`
* `lgb.train()`
* `setinfo.lgb.Dataset()`

### Notes for reviewers

Given this R code

```r
library(lightgbm)

data(agaricus.train, package = "lightgbm")
train <- agaricus.train
dtrain <- lgb.Dataset(train$data, label = train$label)

model <- lgb.train(
    params = list(
        objective = "regression"
    )
    , data = dtrain
    , nrounds = 5L
    , bagging_fraction = 0.5
    , bagging_freq = 2L
)
```

This is the warning shown:

> Warning message:
In lgb.train(params = list(objective = "regression"), data = dtrain,  :
  lgb.train: Found the following passed through '...': bagging_fraction, bagging_freq. These will be used, but in future releases of lightgbm, this warning will become an error. Add these to 'params' instead. See ?lgb.train for documentation on how to call this function.

image

Deprecation warnings for other uses of `...` will be addressed in other pull requests.

PR #4524, opened 2021-08-15
[ci] use flag '--allow-releaseinfo-change' in some 'apt-get update' calls

jameslamb:
On PRs I opened within the last hour (#4522, #4523), the `r-package (ubuntu-latest, R-devel, GCC ASAN/UBSAN) (pull_request)` job is failing.

> E: Repository 'http://deb.debian.org/debian testing InRelease' changed its 'Codename' value from 'bullseye' to 'bookworm'
Error: Process completed with exit code 100.

https://github.com/microsoft/LightGBM/pull/4523/checks?check_run_id=3332073359

After some research, I think this could be fixed by preferring `apt update` to `apt-get update`. A similar error was reported a few hours ago in https://www.mail-archive.com/debian-user@lists.debian.org/msg772749.html, and responses in that thread suggest that running `apt update` in place of `apt-get update` can fix it (e.g. https://www.mail-archive.com/debian-user@lists.debian.org/msg772763.html).

As far as I can tell from that thread, this is happening because of this combination of factors:

1. there was a new Debian release today: https://wiki.debian.org/DebianReleases#Production_Releases
2. some package(s) shipped in that release with bad metadata related to that new release
3. `apt-get update` does not install new dependencies of upgradeable software (https://linuxhint.com/diff_apt_vs_aptget/)

Opening this PR to replace calls to `apt-get update` with `apt update` in R jobs using `rhub/` container images.

### Notes for Reviewers

It's possible similar changes will need to be made for other jobs in the future, and maybe even that we should completely prefer `apt` to `apt-get`, but that requires more research and I think it's outside of the scope of this PR.

### DIFF
--- a/.github/workflows/r_package.yml
+++ b/.github/workflows/r_package.yml
@@ -181,7 +181,7 @@ jobs:
       - name: Install Git before checkout
         shell: bash
         run: |
-          apt update -y
+          apt-get update --allow-releaseinfo-change
           apt-get install --no-install-recommends -y git
       - name: Checkout repository
         uses: actions/checkout@v2.3.4
@@ -210,7 +210,7 @@ jobs:
       - name: Install Git before checkout
         shell: bash
         run: |
-          apt update -y
+          apt-get update --allow-releaseinfo-change
           apt-get install --no-install-recommends -y git
       - name: Checkout repository
         uses: actions/checkout@v2.3.4
--- a/.github/workflows/r_package.yml
+++ b/.github/workflows/r_package.yml
@@ -181,7 +181,7 @@ jobs:
       - name: Install Git before checkout
         shell: bash
         run: |
-          apt-get update
+          apt update -y
           apt-get install --no-install-recommends -y git
       - name: Checkout repository
         uses: actions/checkout@v2.3.4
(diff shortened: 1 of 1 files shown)


jameslamb:
Doesn't seem like this worked 😬

I'm going to try just rebuilding #4522, maybe it was fixed upstrean.

jameslamb:
> I'm going to try just rebuilding #4522, maybe it was fixed upstrean.

That did not work, sadly. https://github.com/microsoft/LightGBM/pull/4522/checks?check_run_id=3334336822

Investigating some other options locally right now.

jameslamb:
Following another suggestion from https://www.mail-archive.com/debian-user@lists.debian.org/msg772749.html worked for me locally (in a container using the `rhub/rocker-gcc-san:latest` image).

> For the 'apt-get' front-end, the necessary solution is to run the
command 'apt-get update --allow-releaseinfo-change'.

Just pushed https://github.com/microsoft/LightGBM/pull/4524/commits/890feeabc93574a1c1f964f327b9b17bf927ae09, which I hope will resolve this issue.

StrikerRUS (approved):
Thank you for fast fix!

I think we can add this flag to other `apt-get update` calls.

jameslamb:
> I think we can add this flag to other apt-get update calls.

Sure! I didn't do that here because I think it needs a bit of investigation. I'm not sure what version of `apt-get` that flag was added in and this project's CI jobs cover a wide range of Ubuntu versions.

PR #4525, opened 2021-08-15
[R-package] parallelize compilation in CMake-based builds

jameslamb:
Inspired by #4462, this proposes passing `-j4` when building the R package with `CMake`, to parallelize compilation and make building faster.

Ran the following on my Mac to benchmark and found that this did make compilation noticeably faster.

```shell
# test this PR
time Rscript build_r.R
rm -r ./lightgbm_r

# test master
git checkout master
time Rscript build_r.R
```

**this PR**

```text
real	1m45.165s
user	4m36.421s
sys	0m13.545s
```

**`master`**

```text
real	3m17.079s
user	3m3.012s
sys	0m9.107s
```

### Notes for Reviewers

CI should pass here once #4524 is merged.


Laurae2:
What happens on computers with fewer than 4 CPU thread? (cloud VMs, VPS, embedded devices)

In R I remember we didn't allow to use more than 1 thread to compile LightGBM by default (the user had to manually add a parameter to perform multithreaded compilation), as to not overload those machines (still common in businesses).

Previous solution 1, in a R configuration file:

```
MAKEFLAGS = -j4
```

Previous solution 2, in a prompt:

```
export MAKE="make -j8"
```

Previous solution 3, directly from R:

```
Sys.setenv(MAKE = "make -j4")
```

Is there a way to have something similar instead of an hardcoded 4 thread?

jameslamb:
> Is there a way to have something similar instead of an hardcoded 4 thread?

Sure, we could expose a command-line argument that you'd pass like this:

```shell
Rscript build_r.R \
    --parallel-build-jobs 4
```

Similar to the way that other customizations have been exposed as command-line flags.

https://github.com/microsoft/LightGBM/blob/86ead2050c006b984e88396e0e380304bac114d1/build_r.R#L44-L51

Do you think that added complexity is worth it?

> I remember we didn't allow to use more than 1 thread to compile LightGBM by default, as to not overload those machines

Is it your opinion that the changes in #4462 should be reverted and that such customization should be exposed in a command-line argument in the Python package too?

Laurae2:
> Do you think that added complexity is worth it?

If this just requires adding `-j` parameter in the list you mentioned, yes (users wanting to compile explicitly in parallel). Otherwise, it would need to longer rework, and instead `-j4` would be a temporary harmonious solution with what we have on the Python side.

In my opinion we should expose that parameter in the command line as `make` `-j` parameter was intended for the user to set it explicitly (because it linearly consumes more RAM to compile with more threads).

Laurae2 (approved):
Can be merged to temporarily get Pythion and R compilation behaving similarly for the number of threads used.

R-package/src/install.libs.R
```diff
@@ -132,7 +132,7 @@ if (WINDOWS && use_visual_studio) {
 # Prepare installation steps
 cmake_args <- NULL
 build_cmd <- "make"
-build_args <- "_lightgbm"
+build_args <- "_lightgbm -j4"
```
StrikerRUS, line 135:
```suggestion
build_args <- c("_lightgbm", "-j4")
```

R-package/src/install.libs.R
```diff
@@ -194,7 +194,7 @@ if (WINDOWS) {
     cmake_args <- c(cmake_args, "-G", shQuote(windows_makefile_generator))
     .run_shell_command("cmake", c(cmake_args, ".."), strict = FALSE)
     build_cmd <- windows_build_tool
-    build_args <- "_lightgbm"
+    build_args <- "_lightgbm -j4"
```
StrikerRUS, line 197:
```suggestion
    build_args <- c("_lightgbm", "-j4")
```

R-package/src/install.libs.R
```diff
@@ -203,7 +203,7 @@ if (WINDOWS) {
       cmake_args <- c(cmake_args, "-G", shQuote(windows_makefile_generator))
       .run_shell_command("cmake", c(cmake_args, ".."), strict = FALSE)
       build_cmd <- windows_build_tool
-      build_args <- "_lightgbm"
+      build_args <- "_lightgbm -j4"
```
StrikerRUS, line 206:
```suggestion
      build_args <- c("_lightgbm", "-j4")
```

StrikerRUS (commented):
I believe passing args as a vector will be more correct, especially on Windows with `processx`
> **args**    Character vector, arguments to the command.
    https://processx.r-lib.org/reference/run.html#arguments

jameslamb:
Sorry for taking so long to get back to this! I'd like to not merge this, and take @Laurae2 's suggestion of allowing for `Rscript build_r.R -j4`.

jameslamb:
Ok I just updated this. `build_r.R` will now accept `-j`, and still defaults to single-threaded compilation.

Tested on my Mac, which has two physical CPUs and 4 logical CPUs (processor = "2.3 GHz Intel Core i5"). The results were what @Laurae2 said we should expect...going from single-threaded to double-threaded compilation gives a noticeable speedup, but beyond that, compilation doesn't get any faster.

```text
time Rscript build_r.R -j1

# real    3m2.032s
# user    2m53.028s
# sys 0m6.873s

time Rscript build_r.R -j2

# real    1m54.311s
# user    3m14.167s
# sys 0m7.428s

time Rscript build_r.R -j4

# real    1m40.045s
# user    4m22.629s
# sys 0m9.230s

time Rscript build_r.R -j8

# real    1m38.459s
# user    4m16.870s
# sys 0m9.590s
```

@Laurae2 what doo you think?

.ci/test_r_package.sh
```diff
@@ -112,7 +112,7 @@ cd ${BUILD_DIRECTORY}
 PKG_TARBALL="lightgbm_*.tar.gz"
 LOG_FILE_NAME="lightgbm.Rcheck/00check.log"
 if [[ $R_BUILD_TYPE == "cmake" ]]; then
-    Rscript build_r.R --skip-install || exit -1
+    Rscript build_r.R -j2 --skip-install || exit -1
```
StrikerRUS, line 115: Why `j2`? We use `j4` in all our CI jobs, e.g.
https://github.com/microsoft/LightGBM/blob/aafedd8a1c57432ac3e38e4af236c00c6c54e84c/.ci/test.sh#L105
https://github.com/microsoft/LightGBM/blob/aafedd8a1c57432ac3e38e4af236c00c6c54e84c/.ci/test.sh#L216
Refer to #1745.

jameslamb: Based on https://github.com/microsoft/LightGBM/pull/4525#issuecomment-904186660 and the investigation I did last night (https://github.com/microsoft/LightGBM/pull/4525#issuecomment-962540858), I found that using more threads than you have available CPUs doesn't provide much speedup compared to using exactly as many threads as CPUs, but does lead to higher memory consumption.

According to https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources, the VMs we get from GitHub Actions come with a single 2-core CPU. This is why I chose `-j2`.

All that said though, I think most of the benefit for CI comes from moving from single-threaded to multi-threaded, and the difference between `-j2` and  `-j4` shouldn't be large in that environment. And I don't think the memory consumption for compilation is going to be a problem, given your point about all other jobs in this project's CI using `-j4` already. I'll change these to `-j4` for consistency with the other CI jobs.

.ci/test_r_package_windows.ps1
```diff
@@ -153,10 +153,10 @@ if ($env:COMPILER -ne "MSVC") {
...
     if ($env:TOOLCHAIN -eq "MINGW") {
       Write-Output "Telling R to use MinGW"
-      $env:BUILD_R_FLAGS = "c('--skip-install', '--use-mingw')"
+      $env:BUILD_R_FLAGS = "c('--skip-install', '--use-mingw', '-j4')"
     } elseif ($env:TOOLCHAIN -eq "MSYS") {
       Write-Output "Telling R to use MSYS"
-      $env:BUILD_R_FLAGS = "c('--skip-install', '--use-msys2')"
+      $env:BUILD_R_FLAGS = "c('--skip-install', '--use-msys2', '-j4')"
     } elseif ($env:TOOLCHAIN -eq "MSVC") {
       $env:BUILD_R_FLAGS = "'--skip-install'"
```
StrikerRUS, line 161: I guess we are missing `-j` option here in case of fallback to MinGW/MSYS
https://github.com/microsoft/LightGBM/blob/b1facf50502fb51a1e60c05b3ec83f68289df497/R-package/src/install.libs.R#L200-L206

R-package/README.md
```diff
@@ -152,6 +152,9 @@ Rscript build_r.R
 
 The `build_r.R` script builds the package in a temporary directory called `lightgbm_r`. It will destroy and recreate that directory each time you run the script. That script supports the following command-line options:
 
+- `-j[jobs]`: number of threads to use when compiling LightGBM. e.g. `-j4` will try to compile 4 objects at a time.
+    - by default, this script uses single-thread compilation
+    - for best results, set `-j` to the number of physical CPUs
```
StrikerRUS, line 157: Also, I believe `by` and `for` should be capitalized, but I'm not sure.
```suggestion
- `-j[jobs]`: number of threads to use when compiling LightGBM. E.g., `-j4` will try to compile 4 objects at a time.
    - by default, this script uses single-thread compilation
    - for best results, set `-j` to the number of physical CPUs
```

jameslamb: both are equally correct. The rules for bulleted lists like this are very vague: https://github.com/microsoft/LightGBM/pull/4064#discussion_r593826093

.ci/test_r_package_windows.ps1
jameslamb: I didn't include a `-j` in this CI script because if the jobs that are supposed to use MSVC end up falling back to MinGW/MSYS, they'll fail anyway:

https://github.com/microsoft/LightGBM/blob/b1facf50502fb51a1e60c05b3ec83f68289df497/.ci/test_r_package_windows.ps1#L215-L224

For users (not CI scripts), passing through `-j` options to MinGW/MSYS is handled in that part of `install.libs.R` that you linked to

https://github.com/microsoft/LightGBM/blob/3d0274caa8e188c75a7d77655b74f5cb8bdc3135/R-package/src/install.libs.R#L207

It looks like the link you shared is from a commit on `master`, not this PR.

StrikerRUS: > I didn't include a `-j` in this CI script because if the jobs that are supposed to use MSVC end up falling back to MinGW/MSYS, they'll fail anyway:

Ah, completely forgot about that check! Thanks for the reminder.

R-package/README.md
StrikerRUS: Sorry, this time I was confused by a full stop before the list. 😃

StrikerRUS (approved):
LGTM! Thanks for making `-j` flag publicly accessible.

PR #4526, opened 2021-08-15
[python] add type hints in docs/conf.py

jameslamb:
Adds type hints to code in `docs/conf.py`, as part of #3756.

### Notes for Reviewers

#3756 has been open for about 8 months now, so I think it's ok for maintainers to start picking up the work to finish it and not wait any longer for outside contributors to find it from searches for `good first issue`.

Better type hints will improve the ability of `mypy` to catch issues (#3868 ).

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -24,6 +24,7 @@ from pathlib import Path
 from re import compile
 from shutil import copytree
 from subprocess import PIPE, Popen
+from typing import Any, List
 from unittest.mock import Mock
 
 import sphinx
@@ -51,7 +52,7 @@ class InternalRefTransform(Transform):
     default_priority = 210
     """Numerical priority of this transform, 0 through 999."""
 
-    def apply(self, **kwargs):
+    def apply(self, **kwargs: Any) -> None:
         """Apply the transform to the document tree."""
         for section in self.document.traverse(reference):
             if section.get("refuri") is not None:
@@ -63,7 +64,7 @@ class IgnoredDirective(Directive):
 
     has_content = True
 
-    def run(self):
+    def run(self) -> List:
         """Do nothing."""
         return []
 
@@ -197,7 +198,7 @@ htmlhelp_basename = 'LightGBMdoc'
 latex_logo = str(CURR_PATH / 'logo' / 'LightGBM_logo_black_text_small.png')
 
 
-def generate_doxygen_xml(app):
+def generate_doxygen_xml(app: Any) -> None:
     """Generate XML documentation for C API by Doxygen.
 
     Parameters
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -197,7 +198,7 @@ def run(self):
 latex_logo = str(CURR_PATH / 'logo' / 'LightGBM_logo_black_text_small.png')
 
 
-def generate_doxygen_xml(app):
+def generate_doxygen_xml(app: Any) -> None:
```
StrikerRUS, line 201: Maybe use `sphinx.application.Sphinx` for `app` (here and at L246 and L308)?
https://www.sphinx-doc.org/en/master/extdev/appapi.html

jameslamb: Sure, can do. I was just following the documentation below.

```text
    Parameters
    ----------
    app : object
        The application object representing the Sphinx process.
```

StrikerRUS: I think we can update docstring as well.

jameslamb: sure! Updated in https://github.com/microsoft/LightGBM/pull/4526/commits/81f0a8259078df7abe12f92b0bcee811129db127

docs/conf.py
```diff
@@ -197,12 +199,12 @@ def run(self):
...
 
 
-def generate_doxygen_xml(app):
+def generate_doxygen_xml(app: Sphinx) -> None:
     """Generate XML documentation for C API by Doxygen.
 
     Parameters
     ----------
-    app : object
+    app : Sphinx
```
StrikerRUS, line 207: I think we can be more precise here.
```suggestion
    app : sphinx.application.Sphinx
```

docs/conf.py
```diff
@@ -242,12 +244,12 @@ def generate_doxygen_xml(app):
...
 
 
-def generate_r_docs(app):
+def generate_r_docs(app: Sphinx) -> None:
     """Generate documentation for R-package.
 
     Parameters
     ----------
-    app : object
+    app : Sphinx
```
StrikerRUS, line 252:
```suggestion
    app : sphinx.application.Sphinx
```

docs/conf.py
```diff
@@ -304,12 +306,12 @@ def generate_r_docs(app):
...
 
 
-def setup(app):
+def setup(app: Sphinx) -> None:
     """Add new elements at Sphinx initialization time.
 
     Parameters
     ----------
-    app : object
+    app : Sphinx
```
StrikerRUS, line 314:
```suggestion
    app : sphinx.application.Sphinx
```

StrikerRUS (approved):
LGTM!

PR #4527, opened 2021-08-15
[python] add type hints to logging functions in basic.py

jameslamb:
Adds type hints to logging code in `python-package/lightgbm/basic.py`, as part of #3756.

### Notes for Reviewers

See https://github.com/microsoft/LightGBM/issues/3756#issuecomment-899115595.

I ran `mypy --ignore-missing-imports python-package/lightgbm` and don't see any new warnings about `basic.py` or code imported from there

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -63,7 +63,7 @@ def _normalize_native_string(func: Callable[[str], None]) -> Callable[[str], Non
     msg_normalized: List[str] = []
 
     @wraps(func)
-    def wrapper(msg):
+    def wrapper(msg: str) -> None:
         nonlocal msg_normalized
         if msg.strip() == '':
             msg = ''.join(msg_normalized)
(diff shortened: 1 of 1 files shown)


python-package/lightgbm/basic.py
```diff
@@ -58,9 +58,9 @@ def register_logger(logger):
...
 
 
-def _normalize_native_string(func):
+def _normalize_native_string(func: Callable[[str], None]) -> Callable[[str], None]:
     """Join log messages from native library which come by chunks."""
-    msg_normalized = []
+    msg_normalized: List[str] = []
 
     @wraps(func)
     def wrapper(msg):
```
StrikerRUS, line 66: Was `wrapper()` simply missed or you haven't annotated it due to problems with `Union` in return type?
https://github.com/python/mypy/issues/1693

jameslamb: oh just missed

jameslamb: updated in https://github.com/microsoft/LightGBM/pull/4527/commits/a1bec4ed1b1cd99290996dd7f4435855c14abb37, `mypy` didn't have any complaints so I think it's ok

StrikerRUS (approved):
Nicely done, LGTM!

PR #4531, opened 2021-08-18
ranking algorithm: position unbiased option

robert-howley-zocdoc:
* adds support for the position unbiased adjustments described in the [Unbiased LambdaMART](https://arxiv.org/pdf/1809.05818.pdf) paper
  * this methodology attempts to correct for position bias in the result set
  * implementation assumes queries are fed into training in the order in which they appeared
* note for fellow practitioners ... 
  * you'll often see lower ndcg@1 but higher ndcg at lower cutoffs relative to standard (no bias correction) training
  * this isn't an unreasonable outcome of reducing position bias
* this pr is loosely inspired by the author's implementation [here](https://github.com/acbull/Unbiased_LambdaMart)
  * warning: interpreting that repo can be difficult bc some parameters were either implicitly or explicitly fixed
  * free to talk in more detail about the implications of fixed sigmoid in that repo and how it translates
* created a branch [position_unbiased_notes](https://github.com/Zocdoc/LightGBM/blob/position_unbiased_notes/src/objective/rank_objective.hpp#L161) that has more detailed comments on how the formulas of the paper line up with the code changes


src/objective/rank_objective.hpp
```diff
@@ -253,9 +306,86 @@ class LambdarankNDCG : public RankingObjective {
...
+      }
+    }
+
+    for (int i = 0; i < truncation_level_; i++) {
+      // Update bias
+      i_biases_pow_[i] = pow(i_costs_[i] / i_costs_[0], position_bias_regularizer);
+      j_biases_pow_[i] = pow(j_costs_[i] / j_costs_[0], position_bias_regularizer);
+    }
+
+    LogDebugPositionBiases();
```
robert-howley-zocdoc, line 355: while i found being able to track evolution and end point of position bias adjustment values useful in debug mode, happy to remove this since it doesn't have any new feature benefit during normal usage.

jameslamb:
Thanks very much for the contribution @robert-howley-zocdoc !

Once you sign the CLA, we'd be happy to start providing a review. Ask here if you have any questions or concerns!

.gitignore
```diff
@@ -462,3 +462,6 @@ dask-worker-space/
 *.pub
 *.rdp
 *_rsa
+
+# swig jni
+*_swig.jnilib
```
jameslamb, line 467: Interesting, were you using SWIG builds to test your changes?

Anyway, can you please open a separate PR proposing this `.gitignore` change, and explaining how files like this get created and why they shouldn't be checked into the repo?

We have a strong preference in this project for small, focused pull requests, and this change doesn't seem tightly related to the core purpose of this PR.

src/objective/rank_objective.hpp
```diff
@@ -156,6 +182,7 @@ class LambdarankNDCG : public RankingObjective {
     std::stable_sort(
         sorted_idx.begin(), sorted_idx.end(),
         [score](data_size_t a, data_size_t b) { return score[a] > score[b]; });
+
```
jameslamb, line 185: Can you please remove these and other whitespace-only changes?

image

If you feel that they improve the readability of the code, we'd welcome a separate pull request proposing them. That separate pull request could also include some of the other changes you've made below where you added or updated comments on existing code (where those comment changes don't relate to the feature of "add position-unbiased option for lambdarank").

like these:

image

image

----

Those types of PRs that only change comments and whitespace take minimal effort for reviewers to review, and shrink the size of changes like this one that are introducing more involved enhancements.

tests/python_package_test/test_sklearn.py
```diff
@@ -112,21 +112,33 @@ def test_multiclass():
     assert gbm.evals_result_['valid_0']['multi_logloss'][gbm.best_iteration_ - 1] == pytest.approx(ret)
 
 
-def test_lambdarank():
+def lambdarank_test_runner(lambdarank_unbiased=False, **kwargs):
```
jameslamb, line 115: thanks very much for adding tests!

I think we could have higher confidence that these changes are backwards-compatible if the existing tests were not touched at all. That sort of acts as a proxy for users' existing code.

I appreciate the effort at reducing duplicated code here, but could you please instead just add new tests for this feature and leave the others untouched? It's ok if that means copying and pasting the contents of `test_lambdarank`.

src/objective/rank_objective.hpp
```diff
@@ -135,19 +146,34 @@ class LambdarankNDCG : public RankingObjective {
     }
     // construct Sigmoid table to speed up Sigmoid transform
     ConstructSigmoidTable();
+
+    // initialize position bias vectors
+    InitPositionBiasesAndGradients();
```
jameslamb, line 151: should this initialization be conditional on the value of `unbiased_`? Otherwise, the default behavior will be to initialize them and then not use them, right?

(apologies if I've misunderstood how this works)

jameslamb (changes requested):
Thanks very much! Please see my first round of minor comments. Other reviewers are more qualified than me to comment on the algorithmic correctness of these changes, and whether or not such a modification should be supported in LightGBM.

src/objective/rank_objective.hpp
robert-howley-zocdoc: they get initialized to 1 and only updated if `unbiased_` is true. so any time you see them when `unbiased_` is false, you're just dividing by 1.

.gitignore
robert-howley-zocdoc: removed it from the change set.

src/objective/rank_objective.hpp
robert-howley-zocdoc: i think i undid all the whitespace changes

robert-howley-zocdoc:
> Thanks very much for the contribution @robert-howley-zocdoc !
> 
> Once you sign the CLA, we'd be happy to start providing a review. Ask here if you have any questions or concerns!

btw sorry for the long lapse. getting approval on the cla took a bit and then i was on vacation for 2+ weeks. back and will be responsive.

jameslamb:
> back and will be responsive.

no problem at all, we understand! If you're able to submit a separate PR with some of the recommended small changes like the `.gitignore` one (https://github.com/microsoft/LightGBM/pull/4531#discussion_r695369261), you might be able to go faster here because you won't have to wait for maintainers to manually approve Continuous Integration runs (GitHub only applies that gate to first-time contributors in a repo).

I also want to set the expectation that I personally am probably not qualified to review that the submitted code is algorithmically correct, and that we'll need a review from @guolinke or @shiyu1994 or @btrotta for that.

Thanks again for working with us on this great addition!

.gitignore
robert-howley-zocdoc: i think this ended up here bc of the way i built it locally. 70/30 chance it was due to me not having looked/thought about c++ for 10 years. ignore the .gitignore

jameslamb:
It's been about two months since this PR had any activity.

@shiyu1994 @guolinke @btrotta could one of you please provide a review? I'd like to bring this to a resolution before the 4.0 release.

---------------

@robert-howley-zocdoc could you update this to the latest `master`? I tried to do that myself tonight, but I think you've opted out of allowing maintainers to push to branches on your fork.

<details><summary>what I tried (click me)</summary>

```shell
git clone \
    git@github.com:Zocdoc/LightGBM.git \
    zocdoc-lgb

cd zocdoc-lgb
git remote add upstream git@github.com:microsoft/LightGBM.git
git pull upstream master
git fetch origin position_unbiased
git checkout position_unbiased
git merge master
git push origin position_unbiased
```

> ERROR: Permission to Zocdoc/LightGBM.git denied to jameslamb.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.

</details>

shiyu1994:
Sorry for the late response. Will review this soon.

thvasilo:
Hi @robert-howley-zocdoc could you clarify this assumption a bit?

> implementation assumes queries are fed into training in the order in which they appeared

I was trying to determine the document position appears in the calculation, does the assumption mean that for a query/document group like:

```
idx query_id      var1      var2      var3  relevance
0           0  0.624776  0.191463  0.598358          0
1           0  0.258280  0.658307  0.148386          0
2           0  0.893683  0.059482  0.340426          0
3           0  0.879514  0.526022  0.712648          1
4           0  0.188580  0.279471  0.062942          0
```

the implementation assumes that the record with idx=0 appeared first in the results, idx=1 second etc? So it assume that the input dataset was created with presentation order in mind correct? Otherwise we'd need to include another column indicating position in the calculation.

thvasilo:
Hi @shiyu1994 based on the comment above and in my limited testing of this PR, reading through the paper/code it looks correct to me. I'd like to still run a couple of tests with the same datasets as in the paper to see how close this gets to the results there.

One question I had: After this gets merged one extension I'd like to work on is allowing the position to be a separate column that is available to the optimization process. The reason I'm suggesting this is that in real world data you often have missing rows/data and other logging errors that would break the contiguous assumption of this implementation, leading to incorrect position biases being updated.

I wanted to get your early opinion on this potential contribution. Position data could be added as a metadata attribute, similar to query boundaries.

robhowley:
@thvasilo 
> the implementation assumes that the record with idx=0 appeared first in the results, idx=1 second etc? So it assume that the input dataset was created with presentation order in mind correct? Otherwise we'd need to include another column indicating position in the calculation.

Correct on all 3. I use `lightgbm` via `mmlspark` so I added a `sortWithinPartitions` call to zocdoc's fork of that repo to take care of it. your point on real world data potentially having incomplete sets is completely valid.

thvasilo:
Hi @robhowley one thing I'm noticing in the debug logs is that while the position biases monotonically decrease for most positions, the cutoff position (truncation level) has a large bias, I even see it go above 1 for `bias_i`, whereas the biases for the positions before are very low as expected. Any idea what could be causing that? What's special about the bias value of the cutoff?

Another thing I'm observing is that the `bias_j` values for positions  2 and 3 `> 1`. Have you observed that or could this be a data issue (because of the real-world data issues I mentioned)?

src/objective/rank_objective.hpp
```diff
@@ -199,15 +222,26 @@ class LambdarankNDCG : public RankingObjective {
...
-        if (norm_ && best_score != worst_score) {
+        if ((norm_ || unbiased_) && best_score != worst_score) {
           delta_pair_NDCG /= (0.01f + fabs(delta_score));
         }
         // calculate lambda for this pair
         double p_lambda = GetSigmoid(delta_score);
         double p_hessian = p_lambda * (1.0f - p_lambda);
+
+        int debias_high_rank = static_cast<int>(std::min(high, truncation_level_ - 1));
+        int debias_low_rank = static_cast<int>(std::min(low, truncation_level_ - 1));
```
thvasilo, line 233: Hi I am trying to investigate what the issue is with the biases for the truncation cutoff spot that I'm observing and I'm wondering if the thresholding here is the culprit.

Within the context of this implementation, the case where `debias_high_rank` gets the value `truncation_level - 1` is when `high > truncation_level - 1`, i.e. `sorted_idx[high_rank] > truncation_level - 1` i.e. `sorted_idx[j] > truncation_level - 1`. 

Correct me if I'm wrong but the only case the above happens is when `j` is the `high_rank`, and `j` starts from `i+1` which can take a `>truncation_level` value right?

I was wondering if it's better to check the truncation level for both high and low ranks and only update the position biases when both are lower than `truncation_level`, instead of thresholding their values.

As I'm looking for the update to introduce the position column as a metadata value, what I'm thinking is along the lines of:

```c++
const data_size_t high = sorted_idx[high_rank];
const data_size_t low = sorted_idx[low_rank];
data_size_t high_position, low_position;
if (unbiased_) {
  // record_positions_ is a vector of length data_size, with the position of each 
  // record in the data. start is the offset for the current query
  high_position = record_positions_[start + sorted_idx[high_rank]];
  low_position = record_positions_[start + sorted_idx[low_rank]];
} else {
  high_position = high;
  low_position = low;
}

if (unbiased_) {
  double p_cost = log(1.0f / (1.0f - p_lambda)) * delta_pair_NDCG;

  // more relevant (clicked) gets debiased by less relevant (unclicked), only if within truncation levels
  if (high_position <= truncation_level_ && low_position <= truncation_level_) {
    i_costs_buffer_[tid][high_position] += p_cost / j_biases_pow_[low_position];
    j_costs_buffer_[tid][low_position] += p_cost / i_biases_pow_[high_position];  // and vice versa
  }
}
// By default we set values of 1.0 as no-ops
double i_bias_pow = 1.0;
double j_bias_pow = 1.0;
// We only use actual bias values if they are both within the truncation limits
if (high_position <= truncation_level_ && low_position <= truncation_level_) {
  i_bias_pow = i_biases_pow_[high_position];
  j_bias_pow = j_biases_pow_[low_position];
}
// update, either with 1.0 values if at least one of data points ended up outside the truncation threshold or the actual biases
p_lambda *= -sigmoid_ * delta_pair_NDCG / i_bias_pow / j_bias_pow;
p_hessian *= sigmoid_ * sigmoid_ * delta_pair_NDCG / j_bias_pow / j_bias_pow;
```

Does this make sense? The main suggestion is to not "clamp" the bias positions between `[0, truncation_level - 1]` but rather not perform bias updates when one of the document-pair items falls outside the truncation level (making all bias positions valid).

It's a bit unclear to me still the difference between `high/low` and `high_rank/low_rank`. IIRC We sort records in a query by their score and to establish high/low rank we compare their label values.

What I want to ensure is that: since my `record_positions_` will have the original data order, the extracted position corresponds to the expected record, which I think is what we accomplish by `record_positions_[start + sorted_idx[high_rank]]`.

src/objective/rank_objective.hpp
shiyu1994: Yes. Please remove this or wrap the debug code with `#ifdef DEBUG` macro.

src/objective/rank_objective.hpp
```diff
@@ -276,6 +387,35 @@ class LambdarankNDCG : public RankingObjective {
...
+   * \brief Should use lambdarank with position bias correction
+   * [arxiv.org/pdf/1809.05818.pdf]
+   */
+  bool unbiased_;
+
+  /*! \brief Position bias regularizer norm */
+  double bias_p_norm_;
+
+  /*! \brief Position bias regularizer exponent, 1 / (1 + bias_p_norm_) */
+  double position_bias_regularizer;
```
shiyu1994, line 415: We'd better add an underscore after the member's name to be consistent with the whole code base, i.e. `position_bias_regularizer_`.

shiyu1994 (commented):
@robert-howley-zocdoc Thanks for the great contribution. Please kindly check my comments. The implementation is ok overall, except some minor mistakes and inconsistence with the Unbiased Lambdarank paper.

thvasilo:
Hi @shiyu1994 if @robert-howley-zocdoc doesn't have the time to make requested changes I could do the final few changes on top of @robert-howley-zocdoc's great work to push this through. The I can do a follow up PR with my extensions that enable a custom position column.

StrikerRUS:
@thvasilo Sounds great!

robhowley:
yea, sorry for not being on top of this. i don't work at zocdoc anymore which makes time allocation to this a little tricky. feel free to takeover as needed.

shiyu1994:
I'm picking up this. But it seems that I have no permission to push to Zocdoc/LightGBM. So I'll merge this into a branch in our repo and open a new PR.

jameslamb:
Thanks @shiyu1994 , yeah I agree with that approach! Thanks @robhowley for getting this work to this point, and for agreeing to let us pick it up and carry it forward. Really really appreciate it!

shiyu1994:
Close this and move to #5393.

PR #4533, opened 2021-08-19
[python] add parameter object_hook to method dump_model

xadupre:
Method dump_model parses a string to create a JSON structure. If a function needs to modify this structure, it needs to walk through the whole structure a second time. The object_hook can be used to do both in one pass. This trick was used to significantly speed up the conversion of an heavy model (>40.000 trees) into ONNX (https://github.com/onnx/onnxmltools/pull/491). This PR would avoid duplication of the code of dump_model in [onnxmltools](https://github.com/onnx/onnxmltools).


python-package/lightgbm/basic.py
```diff
@@ -3357,6 +3357,7 @@ def dump_model(self, num_iteration=None, start_iteration=0, importance_type='spl
             What type of feature importance should be dumped.
             If "split", result contains numbers of times the feature is used in a model.
             If "gain", result contains total gains of splits which use the feature.
+        object_hook: function given to `json.loads`
```
StrikerRUS, line 3360:
```suggestion
        object_hook : callable or None, optional (default=None)
            Detailed description of the argument ended with full stop.
```

StrikerRUS (changes requested):
@xadupre Thank you very much for the contribution!
Very interesting addition! TBH, I've never used this argument of `json.loads()`.
Impressive results in the ONNX speedup! I believe new argument might be useful for other users and it's worth exposing `object_hook` in the `dump_model()` signature.

Generally LGTM, just one comment about numpy doc style below.

jameslamb (commented):
Thanks very much for this! Really nice contribution and we appreciate you taking the time to upstream it into LightGBM.

I agree with @StrikerRUS , this is a great addition and just needs a slight documentation update

xadupre:
Thanks for the feedback. I improved the documentation.

python-package/lightgbm/basic.py
```diff
@@ -3357,6 +3357,15 @@ def dump_model(self, num_iteration=None, start_iteration=0, importance_type='spl
...
             If "gain", result contains total gains of splits which use the feature.
+        object_hook : callable or None, optional (default=None)
+            If not None, object_hook is a function called while parsing the json
+            string returned by the C API. It may be used to alter the json, to store
+            specific values while building the json structure. It avoids
+            walking through the structure again. It saves a significant amount
+            of time if the number of trees is huge.
+            Signature is `def object_hook(node: dict) -> dict`.
+            None is equivalent to `lambda node: node`.
+            See documentation of `json.loads` for further details.
```
StrikerRUS, line 3368: I think code is more appropriate than italic here.
```suggestion
            If not None, ``object_hook`` is a function called while parsing the json
            string returned by the C API. It may be used to alter the json, to store
            specific values while building the json structure. It avoids
            walking through the structure again. It saves a significant amount
            of time if the number of trees is huge.
            Signature is ``def object_hook(node: dict) -> dict``.
            None is equivalent to ``lambda node: node``.
            See documentation of ``json.loads()`` for further details.
```

StrikerRUS (approved):
Thanks a lot! LGTM, except one minor nit about emphasizing style in the docstring.

Ah, forgot to say special thanks for adding test! 🙂

jameslamb (approved):

PR #4536, opened 2021-08-19
[ci] Add checks that MM_PREFETCH and MM_MALLOC are used in CRAN builds

StrikerRUS:
Make sure we provide efficient version of LightGBM via CRAN distribution.


jameslamb (approved):
good idea! I like it

PR #4540, opened 2021-08-20
[ci] Check for MM_PREFETCH and MM_MALLOC not only in CRAN builds

StrikerRUS:
Expands #4536.


jameslamb (approved):
excellent! I think this will make us even more confident in changes to `configure` scripts in the future

Issue #4539, opened 2021-08-20
[ci] warnings from cppcheck


jameslamb:
## Summary

Warnings caught by `cppcheck` should be resolved, if they are determined to be genuine issues and not false positives.

## Motivation

[`cppcheck`](http://cppcheck.sourceforge.net/) is a static analyzer for C++ code.

This tool can catch some classes of issues that are not caught by other linters or tests with sanitizers, and it does that using a lightweight approach that doesn't require compiling an instrumented version of the library.

See https://sourceforge.net/p/cppcheck/wiki/ListOfChecks/ for a full list of checks. 

## Description

Similar to the approach taken in LightGBM with `cpplint` (#1990) and `mypy` (#3867), for a time these checks might be run in continuous integration without blocking merges, just as extra information about the state of the code.

See http://cppcheck.sourceforge.net/ for instructions on how to install `cppcheck` on different operating systems.

LightGBM is a large, complex project and its codebase contains many possible combinations of `#ifdef` conditions. Each additional `#ifdef` adds another combination of things to try, which can slow down checking substantially and have a big impact on the amount of logs produced.

To ensure that `cppcheck` does not take too long and that you can see all of the logs after it's done, I recommend doing the following:

1. direct output to a file
2. remove the third-party libraries included as submodules in LightGBM before running this check
    - NOTE: you can restore these at any time with `git submodulue init && git submodule update --recursive`
3. undefine some preprocessor defines with `-U` to limit the number of combinations searched (you can always selectively remove these `-U` checks to do more thorough checks!)

```shell
rm -r ./external_libs

cppcheck \
    --force \
    --enable=all \
    --std=c++11 \
    -I include/ \
    -UDEBUG \
    -ULABEL_T_USE_DOUBLE \
    -ULGB_R_BUILD \
    -USCORE_T_USE_DOUBLE \
    -Usun \
    -U__sun \
    -U__SVR4 \
    -U__svr4__ \
    -UTIMETAG \
    -UUSE_CUDA \
    -UUSE_GPU \
    -UUSE_MPI \
    src/ \
> cppcheck.txt 2>&1
```

### How to Contribute

Look at the current list of warnings by running `cppcheck` yourself locally (following the steps in the "Description" section above).

Small pull requests focused on a single part of the codebase or type of `cppcheck` warning are greatly appreciated. Please do not submit a pull request that attempts to fix many of the `cppcheck` warnings all at once.

When submitting a pull request, include at least the following in the description:

[...]

PR #4542, opened 2021-08-21
simplify and speed up comparisons for splits with identical gains

### LINKED ISSUE #4414: Trainings binary logloss increasing in some iterations in voting-parallel setting
## Description

Some time ago I encountered the problem that when I did not use min_data_in_leaf with a higher value than default, that the training's binary logloss would increase in some iterations. Spikes would occur which varied in size. Some small but some really big as shown in the example. 
distributed_metrics
I always used `gbdt`. The spikes only occured in a distributed setting and i could observe them for `data` and `voting`-parallel training. I did not test it in a feature parallel setting. After some testing I could see that the source of the error had to lie in LightGBMs C++-Code, but could not find the specific location.
Then I saw this issue #4026. I observed that in my trees the leaf-values became really high as well in the iterations where spikes occured. After installing the fix #4185. This solved the problem for the `data`-parallel case and improved solutions for `voting`-parallel. But small spikes would still occur in the `voting`-parallel case. 
distributed_metrics_voting

The issue is mainly there to point out that #4026 is not yet fully resolved for the `voting`-parallel case. Perhaps the increase in binary_loglosses has another cause in the `voting`-parallel case as well

## Reproducible example

I did not manage to create a reproducible example for the error does not happen every time.

## Environment info

first version: 3.1.1.99
version with fix: 3.2.1.99

```shell
git clone --recursive https://github.com/microsoft/LightGBM
cd LightGBM
mkdir build
cd build
cmake ..
make -j4
```

jameslamb:
## Short Summary

This PR fixes a bug in the `==` operator for `SplitInfo` and `LightSplitInfo` in `src/treelearner/split_info.hpp`. In the event that two candidate splits have different gains, `==` currently incorrectly says that those splits are identical.

I'm not certain, but I believe this is a bug that could lead to incorrect behavior in voting parallel learning, at the stage where learners select the top `k` features to consider for future splits. I'm not sure if this fix would fix #4414, but if I'm right then it might at least help.

## Longer Description

The `SplitInfo` class is used to store information about a candidate tree split, including the gain of the split and the index of the feature it refers to. There is also a lighter-weight alternative to this class, called `LightSplitInfo`.

LightGBM's training process involves calculating the gain for different splits and comparing those splits to each other to determine the best one. To do this, it defines `>` and `==` operators for comparing two `SplitInfo` or `LightSplitInfo` instances.

Running `cppcheck` (#4539), I found what looks like a bug in the definition of the `==` operator for `SplitInfo` and `LightSplitInfo`!

```text
src/treelearner/split_info.hpp:177:25: warning: Opposite inner 'return' condition leads to a dead code block. [oppositeInnerCondition]
      return local_gain == other_gain;
src/treelearner/split_info.hpp:271:25: warning: Opposite inner 'return' condition leads to a dead code block. [oppositeInnerCondition]
      return local_gain == other_gain;
```

This code (introduced in https://github.com/microsoft/LightGBM/commit/4f77bd2860913a9a210c72a367fc88b4a9c13b3d), has what looks like a mistake:

```cpp
if (local_gain != other_gain) {
      return local_gain == other_gain;
}
```

That code is saying "if the gain of this split and the split being compared to it are different, `this_split == that_split` should be `true`". 

### How this fix affects `LightGBM`

I tried commenting out the definitions of those `==` operator in `src/treelearner/split_info.hpp` to see where it's used. It's used much less frequently than the `<` operator, but I found at least one place...in the stage where voting parallel training evaluates splits to pick the top `k` features to consider!

https://github.com/microsoft/LightGBM/blob/a926d4fe939118b6377455220e44ca9ded5b964e/src/treelearner/voting_parallel_tree_learner.cpp#L173

https://github.com/microsoft/LightGBM/blob/a926d4fe939118b6377455220e44ca9ded5b964e/include/LightGBM/utils/array_args.h#L160

https://github.com/microsoft/LightGBM/blob/a926d4fe939118b6377455220e44ca9ded5b964e/include/LightGBM/utils/array_args.h#L137

https://github.com/microsoft/LightGBM/blob/a926d4fe939118b6377455220e44ca9ded5b964e/include/LightGBM/utils/array_args.h#L118-L119

A reminder from [the original LightGBM paper](https://proceedings.neurips.cc/paper/2016/file/10a5ab2db37feedfdeaab192ead4ac0e-Paper.pdf) of how this affects training:

> For local voting, the top-k attributes are selected from each machine according to its local data. Then, globally top-2k attributes are determined by a majority voting among these local candidates. Finally, the full-grained histograms of the globally top-2k attributes are collected from local machines in order to identify the best (most informative) attribute and its split point.


StrikerRUS (commented):
I believe now these `if (local_gain != other_gain)` statements can be moved upper (right before `int local_feature = this->feature;` line) for the speedup.

jameslamb:
> I believe now these `if (local_gain != other_gain)` statements can be moved upper (right before `int local_feature = this->feature;` line) for the speedup.

oooo yeah, good idea! I'll do that here

jameslamb:
updated in https://github.com/microsoft/LightGBM/pull/4542/commits/b40452e75424732381da07adcf7377f28ef719f2

great suggestion @StrikerRUS , this is going to make the majority of split comparisons a little bit faster, very nice 😎

jameslamb:
oops, forgot `LightSplitInfo` too. Updated in https://github.com/microsoft/LightGBM/pull/4542/commits/5a60b7b3b1ddf6e89fc50371109dea8f1f3586f9

shiyu1994 (approved):
A good rewrite to make the code much cleaner. But I think the original implementation is also correct. In the old implementation of `operator==`, it is saying, when `local_gain != other_gain`, the result of `==` is `false`. Since when `local_gain != other_gain` is `true`. `local_gain==other_gain` is false.
But I still think this modification is quite valuable.

jameslamb:
> But I think the original implementation is also correct.

ahhh ok yes @shiyu1994 you're right! I got confused by this 

```cpp
if (local_gain != other_gain) {
     return local_gain == other_gain;
}
```

I'll change the title of the PR before we merge, to make it clear that this is a cleanup / small optimization and that it isn't actually fixing anything "incorrect".

PR #4544, opened 2021-08-22
[python] add type hints on train() in engine.py

jameslamb:
Proposes adding type hints on `train()`, as part of #3756.

### DIFF
--- a/python-package/lightgbm/engine.py
+++ b/python-package/lightgbm/engine.py
@@ -4,6 +4,7 @@ import collections
 import copy
 from operator import attrgetter
 from pathlib import Path
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
 import numpy as np
 
@@ -11,14 +12,34 @@ from . import callback
 from .basic import Booster, Dataset, LightGBMError, _ConfigAliases, _InnerPredictor, _log_warning
 from .compat import SKLEARN_INSTALLED, _LGBMGroupKFold, _LGBMStratifiedKFold
 
+_LGBM_CustomObjectiveFunction = Callable[
+    [Union[List, np.ndarray], Dataset],
+    Tuple[Union[List, np.ndarray], Union[List, np.ndarray]]
+]
+_LGBM_CustomMetricFunction = Callable[
+    [Union[List, np.ndarray], Dataset],
+    Tuple[str, float, bool]
+]
 
-def train(params, train_set, num_boost_round=100,
-          valid_sets=None, valid_names=None,
-          fobj=None, feval=None, init_model=None,
-          feature_name='auto', categorical_feature='auto',
-          early_stopping_rounds=None, evals_result=None,
-          verbose_eval=True, learning_rates=None,
-          keep_training_booster=False, callbacks=None):
+
+def train(
+    params: Dict[str, Any],
+    train_set: Dataset,
+    num_boost_round: int = 100,
+    valid_sets: Optional[List[Dataset]] = None,
+    valid_names: Optional[List[str]] = None,
+    fobj: Optional[_LGBM_CustomObjectiveFunction] = None,
+    feval: Optional[_LGBM_CustomMetricFunction] = None,
+    init_model: Optional[Union[str, Path, Booster]] = None,
+    feature_name: Union[List[str], str] = 'auto',
+    categorical_feature: Union[List[str], str] = 'auto',
+    early_stopping_rounds: Optional[int] = None,
+    evals_result: Optional[Dict[str, Any]] = None,
+    verbose_eval: bool = True,
+    learning_rates: Optional[Union[List[float], Callable[[int], float]]] = None,
+    keep_training_booster: bool = False,
+    callbacks: Optional[List[Callable]] = None
+) -> Booster:
     """Perform the training with given parameters.
 
     Parameters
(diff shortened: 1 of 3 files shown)


StrikerRUS (changes requested):
Oh, that was a hard work, thanks!

sklearn-wrapper and hence Dask-package use callables with another signature for custom objective and metric functions:
https://github.com/microsoft/LightGBM/blob/4e18c60b4d3bbcdb39f88c12f2001a1606cac268/python-package/lightgbm/sklearn.py#L23-L49

https://github.com/microsoft/LightGBM/blob/4e18c60b4d3bbcdb39f88c12f2001a1606cac268/python-package/lightgbm/sklearn.py#L112-L144

jameslamb:
> sklearn-wrapper and hence Dask-package use callables with another signature for custom objective and metric functions

oh interesting, all this time I didn't know that the interfaces were different!

I'll remove those changes from this PR and make them in a separate one.

StrikerRUS:
> I'll remove those changes from this PR and make them in a separate one.

Nice plan, thanks!

python-package/lightgbm/engine.py
```diff
@@ -4,21 +4,42 @@
...
+
+
+def train(
+    params: Dict[str, Any],
+    train_set: Dataset,
+    num_boost_round: int = 100,
+    valid_sets: Optional[List[Dataset]] = None,
+    valid_names: Optional[List[str]] = None,
+    fobj: Optional[_LGBM_CustomObjectiveFunction] = None,
+    feval: Optional[_LGBM_CustomMetricFunction] = None,
```
StrikerRUS, line 32:
```suggestion
    feval: Optional[Union[_LGBM_CustomMetricFunction, List[_LGBM_CustomMetricFunction]]] = None,
```

python-package/lightgbm/engine.py
```diff
@@ -4,21 +4,42 @@
...
+    init_model: Optional[Union[str, Path, Booster]] = None,
+    feature_name: Union[List[str], str] = 'auto',
+    categorical_feature: Union[List[str], str] = 'auto',
```
StrikerRUS, line 35:
```suggestion
    categorical_feature: Union[List[str], List[int], str] = 'auto',
```

python-package/lightgbm/engine.py
```diff
@@ -4,21 +4,42 @@
...
+    early_stopping_rounds: Optional[int] = None,
+    evals_result: Optional[Dict[str, Any]] = None,
```
StrikerRUS, line 37: This dict should be empty when passing to the function. Can we reflect this fact with typing?
```suggestion
    evals_result: Optional[Dict] = None,
```

python-package/lightgbm/engine.py
```diff
@@ -4,21 +4,42 @@
...
+    verbose_eval: bool = True,
```
StrikerRUS, line 38:
```suggestion
    verbose_eval: Union[bool, int] = True,
```

StrikerRUS (changes requested):

python-package/lightgbm/engine.py
jameslamb: ah, right! got it

https://github.com/microsoft/LightGBM/blob/11d7608f2d755899a88bfa1c0fbc779b9e9b3505/python-package/lightgbm/engine.py#L58

python-package/lightgbm/engine.py
jameslamb: Oh interesting, I thought it could be anything but I see now that that dictionary is going to be emptied.

https://github.com/microsoft/LightGBM/blob/11d7608f2d755899a88bfa1c0fbc779b9e9b3505/python-package/lightgbm/callback.py#L90

Yes, there is a way to say in typing that something has a specific value, via the `Literal` type: https://www.python.org/dev/peps/pep-0586.

However, that type was only added to the standard library in Python 3.8, so I don't think it should be used in `lightgbm`.

```text
docker run -t python:3.7-slim python -c "from typing import Literal"
```

ImportError: cannot import name 'Literal' from 'typing' (/usr/local/lib/python3.7/typing.py)

I think a better solution would be to leave the hint as `Dict[str, Any]`, and add to the docstring that anything in that dictionary is going to be removed by `lightgbm`.

That's still a correct documentation of what the eventual value will be. Using an inner type in containers like `Dict[]`, `List[]`, etc. doesn't necessarily imply that they'll be initialized with any contents.

e.g. this pattern is valid and common:

```python
def is_in_list(number: int, numbers: List[int]) -> bool:
    return number in numbers
```

even though that function might be called with an empty list, like `is_in_list(5, [])`

jameslamb: I've proposed that docs update in https://github.com/microsoft/LightGBM/pull/4544/commits/2120d7135ba67e860c5b31b0be3498b8816bd2e4

StrikerRUS: Perfect, thanks!

StrikerRUS (approved):
Great job, many thanks!

python-package/lightgbm/engine.py
```diff
@@ -101,7 +122,9 @@ def train(params, train_set, num_boost_round=100,
         The index of iteration that has the best performance will be saved in the ``best_iteration`` field
         if early stopping logic is enabled by setting ``early_stopping_rounds``.
     evals_result: dict or None, optional (default=None)
-        This dictionary used to store all evaluation results of all the items in ``valid_sets``.
+        Dictionary used to store all evaluation results of all the items in ``valid_sets``.
+        This should be initialized outside of your call to ``train()`` and should be empty.
+        Any initial contents of the dictionary will be deleted by `train()`.
```
jameslamb, line 127:
```suggestion
        Any initial contents of the dictionary will be deleted by ``train()``.
```

ugh I missed some backticks here haha

PR #4545, opened 2021-08-22
[docs] Clarify the fact that predict() on a file does not support saved Datasets (fixes #4034)

### LINKED ISSUE #4034: [R-package] predict() breaks when using a Dataset stored in a file

jameslamb:
Fixes #4034.
Contributes to #4310.

`predict()` methods in LightGBM can be used with data stored in a text file (CSV, TSV, or LibSVM). Those methods do not support creating predictions on other formats like LightGBM `Dataset` objects saved in a binary file.

This PR proposes the following changes to fix #4034:

* clarifies the supported file formats for predict methods in the R and Python package documentation
* adds a unit test in the R package on the behavior of calling `Booster$eval()` on a constructed `Dataset` stored in a file (https://github.com/microsoft/LightGBM/issues/4034#issuecomment-843003438)
* updates the relevant error message in the C++ code used to determine file type when predicting on a file, to make it clearer that only certain formats of text file are supported

### DIFF
--- a/R-package/R/lgb.Dataset.R
+++ b/R-package/R/lgb.Dataset.R
@@ -711,7 +711,7 @@ Dataset <- R6::R6Class(
 #' @description Construct \code{lgb.Dataset} object from dense matrix, sparse matrix
 #'              or local file (that was created previously by saving an \code{lgb.Dataset}).
 #' @param data a \code{matrix} object, a \code{dgCMatrix} object or
-#'             a character representing path to a text file (CSV, TSV, or LibSVM)
+#'             a character representing a path to a text file (CSV, TSV, or LibSVM)
 #'             or a LightGBM Dataset binary file
 #' @param params a list of parameters. See
 #'               \href{https://lightgbm.readthedocs.io/en/latest/Parameters.html#dataset-parameters}{
--- a/R-package/R/lgb.Dataset.R
+++ b/R-package/R/lgb.Dataset.R
@@ -776,8 +776,9 @@ lgb.Dataset <- function(data,
 #' @title Construct validation data
 #' @description Construct validation data according to training data
 #' @param dataset \code{lgb.Dataset} object, training data
-#' @param data a \code{matrix} object, a \code{dgCMatrix} object or
-#'             a character representing a path to a text file (CSV, TSV, or LibSVM)
+#' @param data a \code{matrix} object, a \code{dgCMatrix} object,
+#'             a character representing a path to a text file (CSV, TSV, or LibSVM),
+#'             or a character representing a path to a binary \code{Dataset} file
 #' @param info a list of information of the \code{lgb.Dataset} object
 #' @param ... other information to pass to \code{info}.
 #'
--- a/R-package/R/lgb.Dataset.R
+++ b/R-package/R/lgb.Dataset.R
@@ -710,7 +710,9 @@ Dataset <- R6::R6Class(
 #' @title Construct \code{lgb.Dataset} object
 #' @description Construct \code{lgb.Dataset} object from dense matrix, sparse matrix
 #'              or local file (that was created previously by saving an \code{lgb.Dataset}).
-#' @param data a \code{matrix} object, a \code{dgCMatrix} object or a character representing a filename
+#' @param data a \code{matrix} object, a \code{dgCMatrix} object or
+#'             a character representing path to a text file (CSV, TSV, or LibSVM)
+#'             or a LightGBM Dataset binary file
 #' @param params a list of parameters. See
 #'               \href{https://lightgbm.readthedocs.io/en/latest/Parameters.html#dataset-parameters}{
 #'               The "Dataset Parameters" section of the documentation} for a list of parameters
@@ -774,7 +776,8 @@ lgb.Dataset <- function(data,
 #' @title Construct validation data
 #' @description Construct validation data according to training data
 #' @param dataset \code{lgb.Dataset} object, training data
-#' @param data a \code{matrix} object, a \code{dgCMatrix} object or a character representing a filename
+#' @param data a \code{matrix} object, a \code{dgCMatrix} object or
+#'             a character representing a path to a text file (CSV, TSV, or LibSVM)
 #' @param info a list of information of the \code{lgb.Dataset} object
 #' @param ... other information to pass to \code{info}.
 #'
--- a/R-package/R/lgb.Dataset.R
+++ b/R-package/R/lgb.Dataset.R
@@ -711,8 +711,8 @@ Dataset <- R6::R6Class(
 #' @description Construct \code{lgb.Dataset} object from dense matrix, sparse matrix
 #'              or local file (that was created previously by saving an \code{lgb.Dataset}).
 #' @param data a \code{matrix} object, a \code{dgCMatrix} object,
-#'             a character representing a path to a text file (CSV, TSV, or LibSVM)
-#'             or a LightGBM Dataset binary file
+#'             a character representing a path to a text file (CSV, TSV, or LibSVM),
+#'             or a character representing a path to a binary \code{lgb.Dataset} file
 #' @param params a list of parameters. See
 #'               \href{https://lightgbm.readthedocs.io/en/latest/Parameters.html#dataset-parameters}{
 #'               The "Dataset Parameters" section of the documentation} for a list of parameters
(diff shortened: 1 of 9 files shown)


R-package/R/lgb.Dataset.R
```diff
@@ -774,7 +776,8 @@ lgb.Dataset <- function(data,
 #' @title Construct validation data
 #' @description Construct validation data according to training data
 #' @param dataset \code{lgb.Dataset} object, training data
-#' @param data a \code{matrix} object, a \code{dgCMatrix} object or a character representing a filename
+#' @param data a \code{matrix} object, a \code{dgCMatrix} object or
+#'             a character representing a path to a text file (CSV, TSV, or LibSVM)
```
StrikerRUS, line 780: In Python, it can also be a binary file. I guess it's true and for the R-package.
```py
import lightgbm as lgb
from sklearn.datasets import load_boston

X, y = load_boston(return_X_y=True)
lgb_data = lgb.Dataset(X, y)
lgb_data.save_binary('dataset.bin')
lgb_valid = lgb_data.create_valid('dataset.bin')
lgb.train({'objective': 'regression'}, lgb_valid, num_boost_round=5)
lgb_valid.get_field('label')
```

R-package/R/lgb.Dataset.R
```diff
@@ -710,7 +710,9 @@ Dataset <- R6::R6Class(
 #' @title Construct \code{lgb.Dataset} object
 #' @description Construct \code{lgb.Dataset} object from dense matrix, sparse matrix
 #'              or local file (that was created previously by saving an \code{lgb.Dataset}).
-#' @param data a \code{matrix} object, a \code{dgCMatrix} object or a character representing a filename
+#' @param data a \code{matrix} object, a \code{dgCMatrix} object or
```
StrikerRUS, line 713: One `or` is already used for a binary file:
> or a LightGBM Dataset binary file
```suggestion
#' @param data a \code{matrix} object, a \code{dgCMatrix} object,
```

python-package/lightgbm/basic.py
```diff
@@ -3460,7 +3460,7 @@ def refit(self, data, label, decay_rate=0.9, **kwargs):
         ----------
         data : string, pathlib.Path, numpy array, pandas DataFrame, H2O DataTable's Frame or scipy.sparse
             Data source for refit.
-            If string or pathlib.Path, it represents the path to txt file.
+            If string or pathlib.Path, it represents the path to a text file (CSV, TSV, or LibSVM) or a LightGBM Dataset binary file.
```
StrikerRUS, line 3463: I don't think `refit()` supports binary data due that this data is used in `predict()` internally.

https://github.com/microsoft/LightGBM/blob/4e18c60b4d3bbcdb39f88c12f2001a1606cac268/python-package/lightgbm/basic.py#L3481

```py
import numpy as np
import lightgbm as lgb
from sklearn.datasets import load_boston

X, y = load_boston(return_X_y=True)
lgb_data = lgb.Dataset(X, y)
lgb_data.save_binary('dataset.bin')

bst = lgb.train({'objective': 'regression'}, lgb_data, num_boost_round=5)
bst.refit('dataset.bin', label=np.ones_like(y))
```

src/io/parser.cpp
```diff
@@ -235,7 +235,7 @@ Parser* Parser::CreateParser(const char* filename, bool header, int num_features
   int num_col = 0;
   DataType type = GetDataType(filename, header, lines, &num_col);
   if (type == DataType::INVALID) {
-    Log::Fatal("Unknown format of training data.");
+    Log::Fatal("Unknown format of training data. Only CSV, TSV, and LibSVM formats are supported.");
```
StrikerRUS, line 238: Please update wording here to something like
```
LibSVM (zero-based) / TSV / CSV formatted raw text file
```
image

https://github.com/microsoft/LightGBM/blob/master/docs/Python-Intro.rst#data-interface

StrikerRUS (changes requested):
Thank you for picking this up! I have some comment below:

src/io/parser.cpp
jameslamb: what is "TXT format"? It seems like that  line is confusing the difference between file format and the extension in the file's name.

StrikerRUS: Yeah, I agree, that's why I asked to update it omitting file extension.

jameslamb: OH I misunderstood. I thought by "here" you meant "update this line in the diff to match that line in the docs", but I see now that you meant the opposite. My mistake, I'll update the docs accordingly, thanks!

python-package/lightgbm/basic.py
jameslamb: ah you're right, bad copy-paste, sorry

src/io/parser.cpp
jameslamb: Updated in https://github.com/microsoft/LightGBM/pull/4545/commits/33448b6d2a396447e6515b8357d69fa1cfc7c351. If I still misunderstood what you're requesting, it's ok with me if you push directly to this branch.

python-package/lightgbm/basic.py
jameslamb: updated in https://github.com/microsoft/LightGBM/pull/4545/commits/33448b6d2a396447e6515b8357d69fa1cfc7c351, thanks for the sample code!

R-package/R/lgb.Dataset.R
jameslamb: updated in https://github.com/microsoft/LightGBM/pull/4545/commits/33448b6d2a396447e6515b8357d69fa1cfc7c351, thanks for the sample code!

src/io/parser.cpp
StrikerRUS: I'm sorry for initial confusing suggestion! Thanks, you've made exactly what I wanted to change.

R-package/R/lgb.Dataset.R
```diff
@@ -710,7 +710,9 @@ Dataset <- R6::R6Class(
 #' @title Construct \code{lgb.Dataset} object
 #' @description Construct \code{lgb.Dataset} object from dense matrix, sparse matrix
 #'              or local file (that was created previously by saving an \code{lgb.Dataset}).
-#' @param data a \code{matrix} object, a \code{dgCMatrix} object or a character representing a filename
+#' @param data a \code{matrix} object, a \code{dgCMatrix} object,
+#'             a character representing a path to a text file (CSV, TSV, or LibSVM)
+#'             or a LightGBM Dataset binary file
```
StrikerRUS, line 715: Let's make identical descriptions really identical without paraphrasing. Users might spend their time trying to find a difference where there is actually no any difference. Also, it will help us to not miss all occurrences of identical parameters during possible future updates.
```suggestion
#'             a character representing a path to a text file (CSV, TSV, or LibSVM),
#'             or a character representing a path to a binary \code{Dataset} file
```

StrikerRUS (approved):
Thank you! LGTM, except one minor unifying suggestion.

R-package/R/lgb.Dataset.R
jameslamb: oh sure, seems fine to me!

I'll do this one manually instead of applying in the browser, since the corresponding `lgb.Dataset.Rd` will also have to be regenerated.

jameslamb: updated in https://github.com/microsoft/LightGBM/pull/4545/commits/de189a677a0eb5a69536dfe409744871b95055be, thanks as always for being thorough! You're right, that will make it easier to catch all such phrases in the future.

Issue #4195, opened 2021-04-18
R: passing train parameters is inconvenient


david-cortes:
In the R package, when one wants to fit a model, the functions have an argument `params`, where one passes an R list.

This is inconvenient and makes it hard to find the right parameter names. Would be better if either:
* There was a function `lgb.Params` which would simply be a constructor wrapper returning a list of the parameters but with documented arguments, so that one could get autocomplete and pop-up documentation in IDEs by doing something like this:
```r
model <- lightgbm(
  data,
  params=lgb.Params(<tab here>)
)
```
* The functions accepted and documented their arguments without needing to pass them as a list.

jameslamb:
Thanks very much for the recommendation!

I'm strongly opposed to removing the interface of passing parameters as a list to `params`. At this point, that would be a very impactful breaking change that would affect most users of LightGBM.

However, I'm open to more discussion on the idea of a function which can make the process of setting parameters easier. So far we've been hesitant to take that on because it introduces some maintenance burden and because copying the details of the LightGBM core library into wrapper packages introduces a risk of inconsistencies (like the one you've already noted in #4196).

Some questions about such a function:

* would you want every parameter in https://lightgbm.readthedocs.io/en/latest/Parameters.html to have a corresponding keyword argument in this function?
* would you want the arguments in this function to all have the default values documented in https://lightgbm.readthedocs.io/en/latest/Parameters.html, or would you rather that they have no defaults so you have to explicitly choose a value for each?
* would you want Dataset parameters like `max_bin` (https://lightgbm.readthedocs.io/en/latest/Parameters.html#dataset-parameters) to be handled in this function too?

david-cortes:
But it doesn't need to be a breaking change. What I suggest would be simply to make a function with documented arguments that would return a list with the same elements and values - e.g.
```r
Parameters <- function(arg1=10, arg2="a") {
    return(as.list(environment()))
}
```
But with those arguments documented and already set to their defaults (or perhaps to NULL, guess the effect is the same), so that it would play along with RStudio and other IDEs documentation browser, tooltips, and autocomplete.

I think it should be possible to auto-generate such a file from the html or the documentation system of the C++ library, maybe adding a note that it's autogenerated.

And yes, would be nice if the R package documentation would mention everything that's possible to pass.

jameslamb:
> But it doesn't need to be a breaking change

[...]

PR #4548, opened 2021-08-23
[R-package] add deprecation warnings on uses of '...' in predict() and reset_parameter()

jameslamb:
Contributes to #4310 (based on https://github.com/microsoft/LightGBM/issues/4226#issuecomment-829473584).

This PR proposes adding some deprecation warnings on uses of `...` for parameters in the following functions:

* `predict.lgb.Booster()`
* `Booster$predict()`
* `Booster$reset_parameter()`

### Notes for reviewers

Removing `...` from `reset_parameter()` in release 4.0.0 would make that method's signature identical to the Python package's implementation.

https://github.com/microsoft/LightGBM/blob/4db10d86dcccc856cc42e9f66aef424590997f14/python-package/lightgbm/basic.py#L2870

However, removing `...` from the signature of `Booster$predict()` and `predict.lgb.Booster()` would be different from the behavior in the Python package, which allows additional parameters passed through `**kwargs`.

https://github.com/microsoft/LightGBM/blob/c65a2e3363cdff5ddb31689c6d555ec345932d98/python-package/lightgbm/basic.py#L3399-L3401

I think not allowing `...`, and instead encouraging the use of `params`, for additional prediction parameters is a better interface for the R package. It makes that interface consistent with all the other parts of the R package which use `params` (like Dataset construction and training), and leaves open the possibility of an improved user experience with "other" prediction parameters by implementing something like a prediction-parameters-specific version of the feature described in https://github.com/microsoft/LightGBM/issues/4195#issuecomment-822072038.

But @StrikerRUS , if you feel strongly that consistency with the Python package is more important than enabling that eventual "generate a `params` list from a function" feature for the specific case of `predict()` methods, I'd be ok with leaving the `...` for them.


jameslamb:
I'll look into these CI failures on Mac and Linux builds tomorrow:

```text
predict.lgb.Booster: no visible global function definition for
  ‘modifyList’
  modifyList
Consider adding
  importFrom("utils", "modifyList")
to your NAMESPACE file.
```

Maybe `modifyList()` now needs to be explicitly imported and used with `::`. Very weird, given that it's used in several other places on `master` without causing any issues.

jameslamb:
I just pushed https://github.com/microsoft/LightGBM/pull/4548/commits/799bd013d112a114752a78cfa559638c1b188ba3 to address https://github.com/microsoft/LightGBM/pull/4548#issuecomment-903455208

I'm still not sure why this NOTE is only showing up now, but that commit should fix it.

StrikerRUS:
TBH, I'm not sure what consistency is better: consistency with the Python-package or consistency across the R-package. Probably the latter, but will defer to other maintainers' opinion.

R-package/R/lgb.Booster.R
```diff
@@ -469,8 +480,19 @@ Booster <- R6::R6Class(
...
+                       params = list(),
                        ...) {
 
+      additional_params <- list(...)
+      if (length(additional_params) > 0L) {
+        warning(paste0(
+          "Booster$predict(): Found the following passed through '...': "
+          , paste(names(additional_params), collapse = ", ")
+          , ". These will be used, but in future releases of lightgbm, this warning will become an error. "
+          , "Add these to 'params' instead. See ?predict.lgb.Booster for documentation on how to call this function."
```
StrikerRUS, line 492: > See ?predict.lgb.Booster for ...

Sorry, I'm not familiar with this, but it looks like a copy-paste typo from `predict.lgb.Booster`.

R-package/R/lgb.Booster.R
```diff
@@ -469,8 +480,19 @@ Booster <- R6::R6Class(
                        predcontrib = FALSE,
                        header = FALSE,
                        reshape = FALSE,
+                       params = list(),
```
StrikerRUS, line 483: I suppose we should update all our tests (and examples, but if I'm not mistaken, we don't have any yet) and use `params` in them to silence new warning.

R-package/R/lgb.Booster.R
jameslamb: ah, good question! Nope, this one was not a mistake.

Since the `R6` class `Booster` is not currently exported https://github.com/microsoft/LightGBM/blob/11d7608f2d755899a88bfa1c0fbc779b9e9b3505/R-package/NAMESPACE, users of the R package cannot view any documentation for it with something like `?Booster`.

So I think that in the current state of the package, `?predict.lgb.Booster` is the best place to direct them.

R-package/R/lgb.Booster.R
jameslamb: Have you seen any such warnings or places where prediction parameters are passed through in our examples or tests? I checked through the tests and didn't find any uses of extra parameters that didn't match the other keyword arguments already.

To be clear, this would only apply for passing the parameters listed at https://lightgbm.readthedocs.io/en/latest/Parameters.html#predict-parameters, such as `pred_early_stop`.

I searched for those parameters with `git grep` and didn't find any uses in the R package. For example, with 

```shell
git grep pred_early_stop R-package/
```

jameslamb:
> TBH, I'm not sure what consistency is better: consistency with the Python-package or consistency across the R-package. Probably the latter, but will defer to other maintainers' opinion.

Ok thank you! I think consistency within the R package is probably more important, and consistency between R and Python is valuable but not AS important.

So I think we should move forward with introducing this `params` pattern for extra prediction parameters in the R package.

R-package/R/lgb.Booster.R
StrikerRUS: Ah, OK, thanks a lot for the explnation!

R-package/R/lgb.Booster.R
StrikerRUS: Sorry, didn't know that you had already done it! Then everything is OK with this PR I think. But it's a good sign to add a test for those parameters in R-package 😉

StrikerRUS (approved):
LGTM, thanks!

R-package/R/lgb.Booster.R
jameslamb: No problem, I should have mentioned it. And definitely agree that it's a sign of a missing test.

PR #4552, opened 2021-08-25
WIP: get CI timings on long-running jobs

jameslamb:
I've noticed recently that QEMU builds are regularly timing out after 3 hours.  This slows down development on the repository to a level that I'm now really starting to feel, whiich is especially impactful given the timezone difference between maintainers.

Opening this PR to continue the investigation in #4501 into which tests are taking the longest time to run, so I can propose possibly skipping some of them for QEMU builds.

I tried to do this testing on my fork, but Azure DevOps was not picking up PRs there (https://github.com/jameslamb/LightGBM/pull/48).


StrikerRUS:
Thanks for starting a work on this! I wanted to create an issue for this problem this week.

Just my observations: QEMU either works fine within 2 hours, or approximately doubles this time. I have a weak guess that this might be a hardware issue or deep system-wide software one.

jameslamb:
> QEMU either works fine within 2 hours, or approximately doubles this time

Yeah I've seen this too! This most recent build only took 1h23m:

https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=10794&view=logs&j=c2f9361f-3c13-57db-3206-cee89820d5e3&t=bf5b33f9-9072-549e-d8b1-cab79a1dd2a9

> I have a weak guess that this might be a hardware issue or deep system-wide software one.

I agree with you, that's my guess too. But I think we still might be able to at least temporarily avoid timeouts by skipping or reducing the size of some tests (https://github.com/microsoft/LightGBM/pull/4501#issuecomment-892939147).

jameslamb:
I'm not going to push anything for a while since there are several builds currently pending and waiting on QEMU to finish, but want to document some things I've been researching and am planning to try, maybe later tonight.

### Specifying `platform`  in `docker run`

I see the following warnings in all QEMU builds here (those that succeed and those that timeout).

> WARNING: The requested image's platform (linux/arm64) does not match the detected host platform (linux/amd64) and no specific platform was requested

After some research, I think this warning can be suppressed by explicitly telling Docker which platform you're attempting to emulate, and want to test specifying `--platform linux/arm/v7` right here

https://github.com/microsoft/LightGBM/blob/ee5636f16be636e11dbe3e2960996f52cb846aec/.vsts-ci.yml#L193-L199

some links I referenced while investigating this

* originally learned about this from https://github.com/dani-garcia/vaultwarden/discussions/1264
* https://www.docker.com/blog/multi-platform-docker-builds/

### Running fewer Dask tests on aarch64 builds

Thinking we can define something like the following in `tests/python_package_test/test_dask.py`

```python
on_non_x86_arch = machine() != 'x86_64'

# run fewer tests on aarch64 builds, since they are slower in CI
if on_non_x86_arch:
    tasks = ['binary-classification', 'regression', 'ranking']
    distributed_training_algorithms = ['data']
    data_output = ['array', 'dataframe-with-categorical']
    boosting_types = ['gbdt']
```

I've been testing different versions of these choices by running the following, which reports how many tests will be run but doesn't actually run them.

```shell
cd tests/python_package_test
pytest --collect-only test_dask.py
```

The configuration above would reduce the number of Dask tests in QEMU builds from to 351 to 127

### Pinning to a specific `pypa/manylinux` tag

I've noticed that `pypa/manylinux` publishes images every few days. Many of those updates seems like minor dependency bumps using a bot like dependabot.

https://github.com/pypa/manylinux/commits/master

It's possible that pinning to a specific tag instead of using `:latest` could eliminate one source of variability in builds for this CI job.

jameslamb:
@StrikerRUS since the QEMU timeouts problems seems to be really affecting the development velocity in this project recently, I think we should consider substantially reducing the number of Dask tests run in those builds, with something like the suggestion I left in "Running fewer Dask tests on aarch64 builds" in https://github.com/microsoft/LightGBM/pull/4552#issuecomment-907701734.

Would you support that if I did some work tonight to repurpose this PR to add such skips (or some similar combination of skips)?

I'd prefer to not lose any test coverage across compiler / Python version / operating system / architecture combinations, but in this case I think that it's necessary for a while to get through the backlog of open pull requests and try to get a 3.3.0 release out soon (#4310).

StrikerRUS:
Yes, I support this.

Also please consider the following option: revert a372ed5032572edd66711e534390289e61e37cf7, create a new issue and investigate more why Dask tests stuck during QEMU builds.

jameslamb:
> Yes, I support this.
> 
> Also please consider the following option: revert [a372ed5](https://github.com/microsoft/LightGBM/commit/a372ed5032572edd66711e534390289e61e37cf7), create a new issue and investigate more why Dask tests stuck during QEMU builds.

ok sure, can work on it tonight!

I'm not really sure that it's that "the Dask tests get stuck" and I think it might be more like "MOST tests take significantly longer occasionally, and this is worst for Dask because there are just a lot more Dask tests". But I think you're right, this is having such a significant impact on development that it might be better to just completely turn off the Dask tests there for now and investigate this separately.

StrikerRUS:
We can try try to set `timeoutInMinutes: 360` and check whether percentage of completed tests will be different from current 49% on timed out builds.

jameslamb:
Since we already have so much discussion here, I decided to keep this PR for testing ideas and opened #4600 to completely skip the Dask tests on other architectures for now.

jameslamb:
I'm going to close this PR for now. I'd like to resume running the Dask tests on QEMU CI builds in the future, but I'm not actively working on that.

PR #4564, opened 2021-08-27
[docs] update links to SynapseML (former MMLSpark)

StrikerRUS:
MMLSpark was renamed into SynapseML and moved from Azure to Microsoft GitHub account.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ Shapash (model visualization and interpretation): https://github.com/MAIF/shapas
 
 dtreeviz (decision tree visualization and model interpretation): https://github.com/parrt/dtreeviz
 
-MMLSpark (LightGBM on Spark): https://github.com/Azure/mmlspark
+SynapseML (LightGBM on Spark): https://github.com/microsoft/SynapseML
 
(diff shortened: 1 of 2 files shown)


jameslamb (approved):
whoa! Seems that the `mmlspark` GitHub links are still automatically redirecting for now, but definitely agree with updating all of these references.

I checked with `git grep -i mmlspark` and found all the same lines you changed.

StrikerRUS:
> Seems that the `mmlspark` GitHub links are still automatically redirecting for now, ...

Yeah, GitHub creates redirects for renamed and even transferred repositories:

> When you rename a repository, all existing information, with the exception of project site URLs, is automatically redirected to the new name
   https://docs.github.com/en/github/administering-a-repository/managing-repository-settings/renaming-a-repository

> All links to the previous repository location are automatically redirected to the new location. When you use `git clone`, `git fetch`, or `git push` on a transferred repository, these commands will redirect to the new repository location or URL. However, to avoid confusion, we strongly recommend updating any existing local clones to point to the new repository URL. 
   https://docs.github.com/en/github/administering-a-repository/managing-repository-settings/transferring-a-repository#whats-transferred-with-a-repository

PR #4565, opened 2021-08-27
[python][docs] Refer to string type as `str` in docstrings

StrikerRUS:
I'm sorry, I missed this one in #4557.


jameslamb (approved):
Ah thanks! Sorry for not catching this in my review on #4557.

I just checked like this on `master`

```shell
git grep ': string'
```

and the lines in this PR are the only ones that match.

PR #4568, opened 2021-08-28
[R-package] fix warnings in examples

jameslamb:
The deprecation warnings added as part of #4226 introduced some warnings in the examples in `{lightgbm}`'s documentation.

For example, see https://lightgbm.readthedocs.io/en/latest/R/reference/lgb.train.html

image

This PR fixes them and converts examples to the new pattern we want to encourage as of v3.3.0 (#4310), passing all parameters through `params`.

### DIFF
--- a/R-package/R/lgb.Booster.R
+++ b/R-package/R/lgb.Booster.R
@@ -759,7 +759,7 @@ Booster <- R6::R6Class(
 #' preds <- predict(model, test$data)
 #'
 #' # pass other prediction parameters
-#' predict(
+#' preds <- predict(
 #'     model,
 #'     test$data,
 #'     params = list(
--- a/R-package/R/lgb.Booster.R
+++ b/R-package/R/lgb.Booster.R
@@ -743,15 +743,18 @@ Booster <- R6::R6Class(
 #' data(agaricus.test, package = "lightgbm")
 #' test <- agaricus.test
 #' dtest <- lgb.Dataset.create.valid(dtrain, test$data, label = test$label)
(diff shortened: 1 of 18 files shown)


StrikerRUS (approved):
Thank you very much for fixing warnings!
I've checked all `@examples` in all `.R` files and found only two issues. I think we can update them in this PR as one of them is super minor and another one is directly related to `...` params.

I think we need to assign the following prediction result to some variable
https://github.com/microsoft/LightGBM/blob/ee5636f16be636e11dbe3e2960996f52cb846aec/R-package/R/lgb.Booster.R#L758-L759

Looks like a typo because I can't find `test` parameter. Probably should be `label = test$label`? I guess there is no warning about `test` keyword argument because we haven't deprecated `...` params in Dataset constructor yet.
https://github.com/microsoft/LightGBM/blob/f5925c3f69f1a40684fa4c397283452ee5507175/R-package/R/lgb.Dataset.R#L1153

jameslamb:
> I think we can update them in this PR

Agree with both of these, thanks! Just pushed https://github.com/microsoft/LightGBM/pull/4568/commits/ea5ab13646947e67e0f35270456a9af9817da616 to fix these.

> Looks like a typo because I can't find `test` parameter. Probably should be `label = test$label`?

Thanks for catching this! Just want to note this is a good example of the value of #4226 . In `{lightgbm}` 4.0.0, when `...` is no longer respected, mistakes like this will become loud errors instead of unrecognized parameters being silently ignored.

PR #4569, opened 2021-08-28
[R-package] fix warnings in demos

jameslamb:
Similar to the changes made to examples in the R package's API docs in #4568, this PR fixes warnings in demos related to the deprecation warnings added as part of #4226.

For example, if you run `Rscript R-package/demo/basic_walkthrough.R` on current `master`, you'll see the following on stdout.

> Warning message:
In lgb.train(data = dtrain2, num_leaves = 4L, learning_rate = 1,  :
  lgb.train: Found the following passed through '...': num_leaves, learning_rate, nthread, objective. These will be used, but in future releases of lightgbm, this warning will become an error. Add these to 'params' instead. See ?lgb.train for documentation on how to call this function.

This PR fixes them and converts examples to the new pattern we want to encourage as of v3.3.0 (#4310), passing all parameters through `params`.

### Notes for Reviewers

By release 4.0.0, the goal is to have these demos converted to R package vignettes (#1944). However, until then I think it's at least worth updating these to the new recommended pattern, in case users reference the demos as examples.

### DIFF
--- a/R-package/demo/multiclass.R
+++ b/R-package/demo/multiclass.R
@@ -41,7 +41,6 @@ my_preds <- predict(model, test[, 1L:4L])
 params <- list(
     min_data = 1L
     , learning_rate = 1.0
-    , early_stopping_rounds = 10L
     , objective = "multiclass"
     , metric = "multi_error"
     , num_class = 3L
@@ -51,6 +50,7 @@ model <- lgb.train(
     , dtrain
     , 100L
     , valids
+    , early_stopping_rounds = 10L
 )
 
 # We can predict on test data, identical
--- a/R-package/demo/multiclass.R
+++ b/R-package/demo/multiclass.R
@@ -18,14 +18,18 @@ dtest <- lgb.Dataset.create.valid(dtrain, data = test[, 1L:4L], label = test[, 5
 valids <- list(test = dtest)
 
 # Method 1 of training
-params <- list(objective = "multiclass", metric = "multi_error", num_class = 3L)
+params <- list(
+    objective = "multiclass"
+    , metric = "multi_error"
+    , num_class = 3L
+    , min_data = 1L
+    , learning_rate = 1.0
+)
 model <- lgb.train(
     params
     , dtrain
     , 100L
     , valids
-    , min_data = 1L
-    , learning_rate = 1.0
     , early_stopping_rounds = 10L
 )
 
@@ -34,17 +38,19 @@ model <- lgb.train(
 my_preds <- predict(model, test[, 1L:4L])
 
 # Method 2 of training, identical
-model <- lgb.train(
-    list()
-    , dtrain
-    , 100L
-    , valids
-    , min_data = 1L
+params <- list(
+    min_data = 1L
     , learning_rate = 1.0
     , early_stopping_rounds = 10L
     , objective = "multiclass"
     , metric = "multi_error"
     , num_class = 3L
+)
+model <- lgb.train(
+    params
+    , dtrain
+    , 100L
+    , valids
 )
 
 # We can predict on test data, identical
--- a/R-package/demo/multiclass_custom_objective.R
+++ b/R-package/demo/multiclass_custom_objective.R
@@ -28,7 +28,7 @@ params <- list(
     , metric = "multi_logloss"
 )
 model_builtin <- lgb.train(
-    params = params
+    params
     , dtrain
     , 100L
     , valids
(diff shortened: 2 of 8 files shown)


R-package/demo/multiclass_custom_objective.R
```diff
@@ -20,18 +20,20 @@ valids <- list(train = dtrain, test = dtest)
...
+params <- list(
+    min_data = 1L
+    , learning_rate = 1.0
+    , num_class = 3L
+    , boost_from_average = FALSE
+    , metric = "multi_logloss"
+)
 model_builtin <- lgb.train(
-    list()
+    params = params
```
StrikerRUS, line 31: Is it possible to pass keyword arguments before positional ones in R?
```suggestion
    params
```

R-package/demo/multiclass_custom_objective.R
```diff
@@ -92,17 +94,19 @@ custom_multiclass_metric <- function(preds, dtrain) {
     ))
 }
 
+params <- list(
+    min_data = 1L
+    , learning_rate = 1.0
+    , num_class = 3L
+)
 model_custom <- lgb.train(
     list()
```
StrikerRUS, line 103:
```suggestion
    params
```

R-package/demo/weight_param.R
```diff
@@ -34,14 +34,14 @@ params <- list(
     , num_leaves = 7L
     , max_depth = 3L
     , nthread = 1L
+    , min_data = 1L
+    , learning_raate = 1.0
```
StrikerRUS, line 38:
```suggestion
    , learning_rate = 1.0
```

StrikerRUS (changes requested):
Please fix duplicated `min_data` param here:
https://github.com/microsoft/LightGBM/blob/ee5636f16be636e11dbe3e2960996f52cb846aec/R-package/demo/categorical_features_rules.R#L86-L88

And update `multiclass.R` demo in the following two places:
https://github.com/microsoft/LightGBM/blob/ee5636f16be636e11dbe3e2960996f52cb846aec/R-package/demo/multiclass.R#L27-L28
https://github.com/microsoft/LightGBM/blob/ee5636f16be636e11dbe3e2960996f52cb846aec/R-package/demo/multiclass.R#L42-L47

R-package/demo/multiclass_custom_objective.R
jameslamb: yes, R doesn't care about that difference. R's argument parsing will first match all values passed as keyword args, then pass remaining positional arguments to remaining unmatched parameters.

```r
f <- function(x1, x2,  x3) {
    print(paste0("x1: ", x1, " | x2: ", x2, " | x3: ", x3))
}

f("a", "b", "c")
# [1] "x1: a | x2: b | x3: c"

f(x2 = "a", "b", "c")
# [1] "x1: b | x2: a | x3: c"

f("b", x3 = "a", "c")
# [1] "x1: b | x2: c | x3: a"
```

I'm fine with the suggestion, but just wanted to show R isn't opinionated about that the way Python is.

R-package/demo/multiclass_custom_objective.R
jameslamb: ah! Knew I'd miss one of these, thank you.

jameslamb:
> Please fix duplicated `min_data` param here...and update `multiclass.R`

Fixed in https://github.com/microsoft/LightGBM/pull/4569/commits/3838d33e9c0a8c090838ed432ff167ee09b20720, thanks.

R-package/demo/multiclass_custom_objective.R
StrikerRUS: Oh, interesting! Thanks for the explanation and sorry for the false alarm.

R-package/demo/multiclass.R
```diff
@@ -34,18 +38,20 @@ model <- lgb.train(
...
     , early_stopping_rounds = 10L
     , objective = "multiclass"
     , metric = "multi_error"
     , num_class = 3L
 )
+model <- lgb.train(
+    params
+    , dtrain
+    , 100L
+    , valids
```
StrikerRUS, line 53: Diff of this suggestion looks terrible, but it just proposes moving `early_stopping_rounds` from `params` to `train()` according to all other demos.
```suggestion
    , objective = "multiclass"
    , metric = "multi_error"
    , num_class = 3L
)
model <- lgb.train(
    params
    , dtrain
    , 100L
    , valids
    , early_stopping_rounds = 10L
```

StrikerRUS (approved):
Nice, thank you! Just one minor suggestion below:

PR #4570, opened 2021-08-29
[python] add 'auto' value for `importance_type` param in plotting

StrikerRUS:
Respect `importance_type` attribute of LGBMModel in `plot_importance()`.

Marking this as `breaking`, but I don't think this change will actually affect many users. Only those will be affected who for some reason specify `LGBMModel(importance_type="gain")` and leave default value (`importance_type="split"`) in `plot_importance()`.


jameslamb (approved):
seems ok to me, and I agree this is the type of "breaking" change that is ok for a minor release

PR #3234, opened 2020-07-16
Target and Count encodings for categorical features

shiyu1994:
This pull request is to support converting categorical features into CTR and count values. The CTR values are calculated by dividing the training data into folds, with a cross-validation style. 

The performance evaluation has been added in https://github.com/microsoft/LightGBM/pull/3234#issuecomment-691792107. Note that this is a basic version without ensemble trick. We will have a separate PR to implement ensemble trick to boost the performance of this categorical encoding method.

Detailed descriptions and guideline for reviews can be found below.

### Idea
LightGBM can handle categorical features internally, without any manual encoding into numerical values by users. The current approach that LightGBM handles categorical features can be found here https://lightgbm.readthedocs.io/en/latest/Features.html#optimal-split-for-categorical-features.
However, this approach has two known drawbacks:
1. Strong regularization is often required for this approach to be effective when the number of categories are high. The 3 hyperparameters `cat_l2` (https://lightgbm.readthedocs.io/en/latest/Parameters.html#cat_l2), `cat_smooth` (https://lightgbm.readthedocs.io/en/latest/Parameters.html#cat_smooth) and `max_cat_to_onehot` (https://lightgbm.readthedocs.io/en/latest/Parameters.html#max_cat_to_onehot) need to be tuned carefully.
2. The current approach requires sorting over call categories for every categorical feature on every node, which can be slow with a large number of categories.

Inspired by CatBoost, we implement two new approach for internal encoding of categorical features:
1. Encode with the categories by target (label) values (Unknown as target encoding). A category value `c` of feature `j` is encoded as
<a href="https://www.codecogs.com/eqnedit.php?latex=\frac{\sum_i&space;I[x_{ij}=c]&space;y_i&space;&plus;&space;w&space;\cdot&space;p}{\sum_i&space;I[x_{ij}=c]&space;&plus;&space;w}" target="_blank"></a>
where <a href="https://www.codecogs.com/eqnedit.php?latex=x_{ij}" target="_blank"></a> is the value of feature `j` for data `i` in the training set, <a href="https://www.codecogs.com/eqnedit.php?latex=I" target="_blank"></a> is the indicator function, <a href="https://www.codecogs.com/eqnedit.php?latex=y_i" target="_blank"></a> is the label for data `i`, `p` is the prior value (which is the mean of labels over the whole dataset, by default), and `w` is the weight for the prior value.
2. Encode the categories by the number of appearances in the whole training data, i.e.,
[...]

Issue #2526, opened 2019-10-24
Feature Request: customizable early_stopping_tolerance


kryptonite0:
I have a situation where the default numerical tolerance (0.001) for early stopping is too large. My target has a gamma distribution and the LGB Regressor reaches convergence too early, when the numerous low target values are well approximated by the model, but the few large values are still underestimated. When I deactivate early stopping, I can see the loss metric still improving at the 4th or more decimal digit, past the best iteration reached during early stopping. It would be great to be able to set manually the tolerance.

StrikerRUS:
@kryptonite0 Can you please provide a reproducible example or training logs at least?

As I know, we do not have any **default** numerical tolerance. Refer to:
https://github.com/microsoft/LightGBM/blob/fc991c9d7e07fd8e460ea24d1f5609f61fd2af3f/python-package/lightgbm/callback.py#L227
https://docs.python.org/3/library/operator.html#operator.lt

Linking https://github.com/dmlc/xgboost/issues/4982 here.

StrikerRUS:
Ping @kryptonite0 

As you can see from examples, there is no "default numerical tolerance (0.001)":
https://github.com/microsoft/LightGBM/blob/785e477856b9e9e9e69123caf175dbb8455c3b11/examples/python-guide/advanced_example.py#L52-L59

```
[1]	training's binary_logloss: 0.680894
[2]	training's binary_logloss: 0.672151
[3]	training's binary_logloss: 0.664753
[4]	training's binary_logloss: 0.656185
[5]	training's binary_logloss: 0.648174
[6]	training's binary_logloss: 0.641671
[7]	training's binary_logloss: 0.635597
[8]	training's binary_logloss: 0.628874
[9]	training's binary_logloss: 0.622432
[10]	training's binary_logloss: 0.616403
Finished first 10 rounds...
```

StrikerRUS:
At present we do not have any "default numerical tolerance". But having customizable early stopping tolerance might be useful in some cases.

StrikerRUS:
Closed in favor of being in #2302. We decided to keep all feature requests in one place.

Welcome to contribute this feature! Please re-open this issue (or post a comment if you are not a topic starter) if you are actively working on implementing this feature.

StrikerRUS:
#4580 implemented this feature request for Python-package. Thank you very much @jmoralez !
[...]

PR #4574, opened 2021-08-29
[RFC][python] deprecate advanced args of `train()` and `cv()`  functions and sklearn wrapper

StrikerRUS:
Proposing this PR to include in #4310
> I also think this release could be a good opportunity for maintainers to think carefully about what breaking changes might be made (in addition to #3234) in a 4.0.0 release, and to add deprecation warnings for them in the 3.3.0 release.

Encourage users to use callbacks. With removed in `4.0.0` release advanced arguments codebase will become a little bit clearer, confusing cases when both callback and corresponding advanced argument are used will be eliminated, there will be no need to expose new callbacks or new arguments of existing callbacks (#2526) via new arguments of high-level functions/classes making signature of functions incredibly difficult to understand especially for newcomers.


jameslamb:
I'm supportive of this idea, but could preserve the current behavior where `.best_iteration_` and `.best_score_` are `None` if early stopping is not used?

Having those now be 0 is a breaking change and I think we should only make that breaking change in 4.0.0. There could be user code out there that is using `_best_iteration_ is None` as a way to detect whether not early stopping was used.

python-package/lightgbm/callback.py
```diff
@@ -50,19 +50,27 @@ def _format_eval_result(value: list, show_stdv: bool = True) -> str:
 
 
 def print_evaluation(period: int = 1, show_stdv: bool = True) -> Callable:
```
jmoralez, line 52: maybe we could rename this to `log_evaluation` as well? Seems like it is only used in `engine` right now so it could be a good time before it is used explicitly as a callback.

StrikerRUS:
@jameslamb

> could we preserve the current behavior

Sure, will revert some lines of code.

python-package/lightgbm/callback.py
StrikerRUS: @jmoralez 

Good idea, I like it! Unfortunately, `print_evaluation()` callback is already in public API, so we cannot simply rename it.
https://lightgbm.readthedocs.io/en/latest/Python-API.html#callbacks

I'll prepare a separate PR for that if you don't mind.

StrikerRUS:
@jameslamb 
> could we preserve the current behavior

Addressed in 07046ecafff4f949d96883a2088487c269e27ac5.
> or :obj:`None`

in `best_score_`'s docstring was just a mistake. It's always a dict, possibly an empty one.

Also, fixed a bug when `best_iteration_` was not set if early stopping was specified via keyword arguments or a callback, which is now a recommended way of specifying early stop.
In addition, fixed a bug when `best_iteration_` and `evals_result_` were preserved (not updated) after a previous call to `fit()` method. This happens when the first call was with `early_stopping_rounds=xx` and the seconnd one without `early_stopping_rounds=None`.

I can extract bug fixes into separate PRs to make this one easier to review if needed.

jameslamb (approved):
> I can extract bug fixes into separate PRs to make this one easier to review if needed.

I can see where the bug fixes are and understand them, think it's ok to leave them in this PR. Thanks very much for those!

This looks great to me, I have no other comments. Sorry it took a few days to come review this.

Really looking forward to the point in the future where we come back and remove all these deprecation warnings and backwards-compatibility things in the Python and R packages 😀

PR #4575, opened 2021-08-29
[docs][python] Refer to functions as `callable` in docstrings

StrikerRUS:
Also, use singular form of a word in a `list of ...` type descriptions to match ones with simple types (`list of int` or `list of str`).


jameslamb (approved):
nice standardization, thanks!

PR #4577, opened 2021-08-30
[RFC][python] deprecate `silent` and standalone `verbose` args. Prefer global `verbose` param

StrikerRUS:
Proposing this PR to include in #4310

> I also think this release could be a good opportunity for maintainers to think carefully about what breaking changes might be made (in addition to #3234) in a 4.0.0 release, and to add deprecation warnings for them in the 3.3.0 release.

Having multiple params and arguments that control verbosity is quite confusing, see https://github.com/microsoft/LightGBM/issues/1157#issuecomment-395257431. I believe leaving only `verbose` keyword argument, that is described in our [general parameters guide](https://lightgbm.readthedocs.io/en/latest/Parameters.html#verbosity) and which one that a lot of users is used to simply set to `-1`, will simplify user experience and decrease the maintainance burden for us.

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -1440,7 +1440,7 @@ class Dataset:
                              f'Please use {key} argument of the Dataset constructor to pass this parameter.')
         # user can set verbose with params, it has higher priority
         if silent != "warn":
-            _log_warning("'silent' argument is deprecated and will be removed in 4.0.0 release. "
+            _log_warning("'silent' argument is deprecated and will be removed in a future release of LightGBM. "
                          "Pass 'verbose' parameter via 'params' instead.")
         else:
             silent = False
@@ -2494,8 +2494,8 @@ class Booster:
         params = {} if params is None else deepcopy(params)
         # user can set verbose with params, it has higher priority
         if silent != 'warn':
-            _log_warning("'silent' argument is deprecated and will be removed in 4.0.0 release. "
-                         "Pass 'verbose' parameter via keyword arguments instead.")
+            _log_warning("'silent' argument is deprecated and will be removed in a future release of LightGBM. "
+                         "Pass 'verbose' parameter via 'params' instead.")
         else:
             silent = False
         if not any(verbose_alias in params for verbose_alias in _ConfigAliases.get("verbosity")) and silent:
@@ -3296,7 +3296,7 @@ class Booster:
         if verbose in {'warn', '_silent_false'}:
             verbose = verbose == 'warn'
         else:
-            _log_warning("'verbose' argument is deprecated and will be removed in 4.0.0 release.")
+            _log_warning("'verbose' argument is deprecated and will be removed in a future release of LightGBM.")
         if verbose:
             _log_info(f'Finished loading model, total used {int(out_num_iterations.value)} iterations')
         self.__num_class = out_num_class.value
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -1123,7 +1123,7 @@ class Dataset:
     """Dataset in LightGBM."""
 
     def __init__(self, data, label=None, reference=None,
-                 weight=None, group=None, init_score=None, silent=False,
+                 weight=None, group=None, init_score=None, silent='warn',
                  feature_name='auto', categorical_feature='auto', params=None,
                  free_raw_data=True):
         """Initialize Dataset.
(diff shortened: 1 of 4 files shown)


python-package/lightgbm/basic.py
```diff
@@ -3283,6 +3293,10 @@ def model_from_string(self, model_str, verbose=True):
         _safe_call(_LIB.LGBM_BoosterGetNumClasses(
             self.handle,
             ctypes.byref(out_num_class)))
+        if verbose != 'warn':
+            _log_warning("'verbose' argument is deprecated and will be removed in 4.0.0 release.")
+        else:
+            verbose = True
         if verbose:
             _log_info(f'Finished loading model, total used {int(out_num_iterations.value)} iterations')
```
StrikerRUS, line 3301: For this one, I don't think one log line deserves a standalone argument. Users are able to get number of iterations after constructing a Booster if needed.

python-package/lightgbm/basic.py
```diff
@@ -2574,7 +2584,7 @@ def __init__(self, params=None, train_set=None, model_file=None, model_str=None,
             self.__num_class = out_num_class.value
             self.pandas_categorical = _load_pandas_categorical(file_name=model_file)
         elif model_str is not None:
-            self.model_from_string(model_str, not silent)
+            self.model_from_string(model_str)
```
StrikerRUS, line 2587: Without touching this line each call to `train()` ends with 
> 'verbose' argument is deprecated and will be removed in 4.0.0 release.

jameslamb:
I support this change, good idea!

But I have a mild preference for these deprecation warnings to use language like "in a future release of lightgbm" instead of "in the 4.0.0 release". This way if something happens and we end up not making all of these changes in exactly 4.0.0 (i.e. in case we make mistakes), that warning would still be accurate.

Since LightGBM doesn't do long-term support release (i.e. there wouldn't be a 3.3.1 release after 4.0.0 is out), I don't think it's necessary to list a specific future version in deprecation warnings.

StrikerRUS:
@jameslamb 
> But I have a mild preference for these deprecation warnings to use language like "in a future release of lightgbm" instead of "in the 4.0.0 release".

I think it's better to be explicit and specify exact version we're targeting. I understand your worries, but these changes are quite easy to implement and  I hope nothing can prevent us to include them in `4.0.0`. As opposite, we cannot be 100% sure that there will be no `3.3.1` release before `4.0.0`. As you've said, we can end up having serious bug in `3.3.0` that will require releasing hotfix version `3.3.1`. In that case "in a future release of lightgbm" will point to `3.3.1` version. It seems to me misleading. WDYT?

jameslamb:
>  In that case "in a future release of lightgbm" will point to 3.3.1 version. It seems to me misleading. WDYT?

It wouldn't be misleading in that example, because in that situation I expect that these deprecations warnings would still be left in the library in a 3.3.1 release. So both 3.3.0 code and 3.3.1 code would say "in a future release of lightgbm" if you are using the deprecated behavior.

I think that "a future release" is general enough that it won't / shouldn't be interpreted to mean "the next release".

As a general rule, I just think that log messages in libraries shouldn't be used as a source of information about very specific future versions, since we can't say with certainty right now whether the release that these deprecations will be removed in will be 4.0.0 or 4.0.1 or 4.1.0 or 5.0.0. We have a specific plan and it seems that *the most likely* situation is that these deprecations will become errors in exactly 4.0.0, but I don't think that's guaranteed to be true.

StrikerRUS:
> I think that "a future release" is general enough that it won't / shouldn't be interpreted to mean "the next release".

Ah, here is where we are thinking differently. For me personally "a future release" === "the next release".

> As a general rule, I just think that log messages in libraries shouldn't be used as a source of information about very specific future versions, ...

This statement is quite controversial, I think. For instance, scikit-learn has very strict and formal rules of a deprecation process and always specify target version for a removal.

> As in these examples, the warning message should always give both the version in which the deprecation happened and the version in which the old behavior will be removed. If the deprecation happened in version 0.x-dev, the message should say deprecation occurred in version 0.x and the removal will be in 0.(x+2), so that users will have enough time to adapt their code to the new behaviour.
   https://scikit-learn.org/stable/developers/contributing.html#deprecation

Example: https://github.com/scikit-learn/scikit-learn/blob/786c4e5e39648127f41913d4e0c53672a44306f0/sklearn/tree/_classes.py#L338-L346

So, I believe it depends on development policy whether warnings should be explicit or not.

However, I don't have a strong preference and will rephrase all warnings.

python-package/lightgbm/basic.py
jameslamb: agree! (sorry I missed commenting on this in my previous review)

jameslamb:
> However, I don't have a strong preference and will rephrase all warnings.

Totally hear your concerns, and I think the `scikit-learn` example is an excellent counterpoint. Thanks very much for making this change.

jameslamb (approved):
looks great, thanks for doing this before 4.0.0!

PR #4579, opened 2021-09-01
[ci] explicitly set --platform when running aarch64 image in QEMU builds

jameslamb:
I see the following warnings in all QEMU builds here (those that succeed and those that timeout).

> WARNING: The requested image's platform (linux/arm64) does not match the detected host platform (linux/amd64) and no specific platform was requested

After some research, I think this warning can be suppressed by explicitly telling Docker which platform we're attempting to emulate. I'm not sure if this change will resolve the issues we've had with some QEMU builds being much slower than others (https://github.com/microsoft/LightGBM/pull/4552#issuecomment-905521220), but it seems plausible that this could help. Explicitly telling Docker that we're trying to emulate the aarch64 architecture on an amd64 machine might unlock some different behavior in Docker that prevents performance issues like those we've experienced.

I can at least confirm, from investigation on #4552 , that this change does suppress that warning shown above.

Some links I referenced while investigating this

* originally learned about this from https://github.com/dani-garcia/vaultwarden/discussions/1264
* https://www.docker.com/blog/multi-platform-docker-builds/
* https://devblogs.microsoft.com/premier-developer/running-docker-windows-and-linux-containers-simultaneously/

### Notes for Reviewers

I checked the platform of both images used in the QEMU job using the following code.

```shell
get_platform() {
    local image=${1}
    docker pull ${image}
    echo "--- platform (${image}) ---"
    docker image inspect ${image} \
    | jq -r '.[0] | "\(.Os)/\(.Architecture)"'
}

get_platform multiarch/qemu-user-static:latest
get_platform quay.io/pypa/manylinux2014_aarch64:latest
```

Which produced the following.

```text
--- platform (multiarch/qemu-user-static) ---
linux/amd64

--- platform (quay.io/pypa/manylinux2014_aarch64:latest) ---
linux/arm64
```

### Testing

Editing this section to record the timings of a few consecutive runs (triggered by empty commits).

| status | runtime | logs |
|:------:|:-------:|:-----:|
| ✅       | 1h36m  | [link](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=10868&view=logs&j=c2f9361f-3c13-57db-3206-cee89820d5e3&t=bf5b33f9-9072-549e-d8b1-cab79a1dd2a9) |
| ✅       | 1h43m  | [link](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=10873&view=logs&j=c2f9361f-3c13-57db-3206-cee89820d5e3) |
| ✅  | 1h35m | [link](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=10877&view=logs&j=c2f9361f-3c13-57db-3206-cee89820d5e3) |
| :x: | 3h00m | [link](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=10878&view=logs&j=c2f9361f-3c13-57db-3206-cee89820d5e3) |
| ✅ | 1h42m | [link](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=10880&view=logs&j=c2f9361f-3c13-57db-3206-cee89820d5e3) |


jameslamb:
QEMU build passed in 1h36m without raising the warning: https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=10868&view=logs&j=c2f9361f-3c13-57db-3206-cee89820d5e3&t=bf5b33f9-9072-549e-d8b1-cab79a1dd2a9.

I'm going to push an empty commit to try triggering another build, just to get a little more evidence that this might be helping with the timeouts problem.

jameslamb:
since there isn't much other activity on the repo at the moment, I'm going to push another empty commit to trigger another run. I'll also consolidate these links and runtimes by editing the PR description, so they can be read more easily.

StrikerRUS:
Thanks a lot for your efforts in investigating the cause of QEMU builds instability! We should merge this PR regardless it decreases tests duration or not. At least it eliminates the warning you've mentioned above.

Please don't hardcode `linux/arm64` value. The `QEMU_multiarch` Azure Pipelines job and CI scripts were designed to respect any valid value in the `ARCH` env variable.
https://github.com/microsoft/LightGBM/blob/4ee6399db460644c14847662e1ccd55ebb026c17/.vsts-ci.yml#L159

Right now one can extend the job matrix to something like the following
```yml
  strategy:
    matrix:
      bdist_1:
        TASK: bdist
        ARCH: aarch64
      bdist_2:
        TASK: bdist
        ARCH: ppc64
      bdist_3:
        TASK: bdist
        ARCH: s390x
```
and it will work fine. In terms of environment setup, not in terms of successful LightGBM compilation 🤣 .
Here are all possible platform tags: https://github.com/pypa/manylinux#manylinux.
Can we preserve this feature and set `platform` flag based on the `ARCH` env variable?
At least we can append `PLATFORM` item into list of environmental variables associated with each job in the job matrix:
```yml
  strategy:
    matrix:
      bdist:
        TASK: bdist
        ARCH: aarch64
        PLATFORM: linux/arm64
```
If you ask me, I'd personally prefer detecting a platform from the `ARCH` variable 🙂 .

jameslamb:
> If you ask me, I'd personally prefer detecting a platform from the ARCH variable

For sure, I didn't think about using this approach to emulate other architectures. That's a good point.

Which of these would you prefer?

* read platform from a separate `PLATFORM` variable we set
* get it programmatically based on the chosen image, using `docker image inspect` (like in the description of this PR)?

I don't have a strong preference between those. Setting `PLATFORM` manually would be easier right now since we only have one such job, but doing it programmatically would reduce the amount of configuration to think about if we want to add more platforms in the future.

StrikerRUS:
Did you noticed how long `docker pull` and `docker image inspect` take?
I prefer 
> get it programmatically based 

if it doesn't take a significant amount of CI time.

Also, I can suggest defining a mapping from `ARCH` env variable to correct platform value. But unfortunately this approach won't automatically handle new platforms that possibly will be added in the future.

jameslamb:
> Did you noticed how long docker pull and docker image inspect take?

`docker inspect` takes a negligible amount of time (< 1 seconds). It's just reading a little JSON that's already bundled in the image.

And however long `docker pull` takes should add a very small amount of time to the job (also < 1 second), since the image will then be present and match the cache, so it won't be re-pulled when running `docker run` immediately after.

For example

```shell
docker pull quay.io/pypa/manylinux2014_aarch64:latest
docker run --platform linux/arm64 --rm quay.io/pypa/manylinux2014_aarch64:latest /bin/bash
```

you'll see `docker run` immediately enters a container running that image without re-pulling it

I'll try the programmatic way.

StrikerRUS:
> the image will then be present and match the cache, so it won't be re-pulled when running `docker run` immediately after.

Ah, I forgot about caching! Thanks!

jameslamb:
Looks like this most recent one is still going to timeout and fail: https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=10878&view=logs&j=c2f9361f-3c13-57db-3206-cee89820d5e3

But I agree with https://github.com/microsoft/LightGBM/pull/4579#issuecomment-910745714, I think this PR is worth merging even if it doesn't totally fix the problem. I'll push the changes to set `--platform` programmatically soon.

.vsts-ci.yml
```diff
@@ -190,12 +191,20 @@ jobs:
       $ROOT_DOCKER_FOLDER/.ci/setup.sh || exit -1
       $ROOT_DOCKER_FOLDER/.ci/test.sh || exit -1
       EOF
+      IMAGE_URI="quay.io/pypa/manylinux2014_${ARCH}"
+      docker pull "${IMAGE_URI}" || exit -1
+      PLATFORM=$(
+        docker image inspect "${IMAGE_URI}"\
```
StrikerRUS, line 197:
```suggestion
        docker image inspect "${IMAGE_URI}" \
```

.vsts-ci.yml
```diff
@@ -190,12 +191,20 @@ jobs:
...
+        | jq -r '.[0] | "\(.Os)/\(.Architecture)"'
+      ) || exit -1
+      echo "detected image platform: ${PLATFORM}"
```
StrikerRUS, line 200: Would you prefer leaving this debugging message?

jameslamb: yes definitely. It will be helpful if this job fails in the future with cryptic errors, since there's no other place in the logs to see the value of `--platform` that was used.

jameslamb:
hmmm this is a new CI failure I haven't seen before!

https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=10900&view=logs&j=1f4df553-f999-5fff-c6fe-71123c872ae2&t=5236b51f-5710-5c9e-d20e-7548daf49042&l=181

From `Windows cpp_tests`

```text
...
[ RUN      ] ChunkedArrayTest.doubleFreeSafe
[LightGBM] [Fatal] no conversion to double for: x1
[       OK ] ChunkedArrayTest.doubleFreeSafe (0 ms)
...
[----------] 11 tests from ChunkedArrayTest (9 ms total)

[----------] 5 tests from AtofPreciseTest
[ RUN      ] AtofPreciseTest.CornerCases
[LightGBM] [Warning] convert to double got underflow or overflow: 179769313486231580793728971405303415079934132710037826936173778980444968292764750946649017977587207096330286416692887910946555547851940402630657488671505820681908902000708383676273854845817711531764475730270069855571366959622842914819860834936475292719074168444365510704342711559699508093042880177904174497792
...
[----------] 5 tests from AtofPreciseTest (1 ms total)

[----------] Global test environment tear-down
[==========] 20 tests from 3 test suites ran. (11 ms total)
[  PASSED  ] 20 tests.
```

StrikerRUS:
Looks like a wrong exit code. All tests are passed
```
[==========] 20 tests from 3 test suites ran. (11 ms total)
[  PASSED  ] 20 tests.
```

jameslamb:
Oh I see. I thought this looked like a problem

```text
[LightGBM] [Fatal] no conversion to double for: x1
```

But I guess it must not be because I see it in the logs of other successful builds, like this one from `master`: https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=10902&view=logs&j=1f4df553-f999-5fff-c6fe-71123c872ae2&t=5236b51f-5710-5c9e-d20e-7548daf49042

Just manually restarted the failed build.

StrikerRUS (approved):
LGTM!

PR #4581, opened 2021-09-01
[R-package] preserve uses of '...' in Dataset slice() method

jameslamb:
As mentioned in https://github.com/microsoft/LightGBM/pull/4573#discussion_r698897132, the changes in #4573 (removing `...` from `Dataset$new()`) represent a small regression, since they break the currently-supported behavior of passing additional parameters or attributes like `weight` through `Dataset$slice()`.

This PR restores that behavior, makes the deprecation warning about `...` clearer, and adds unit tests to confirm that that behavior still works.

### Note for Reviewers

When the breaking changes in #4226 are implemented in release 4.0.0, the unit tests introduced in this PR will break and should be removed.

### DIFF
--- a/R-package/R/lgb.Dataset.R
+++ b/R-package/R/lgb.Dataset.R
@@ -546,21 +546,37 @@ Dataset <- R6::R6Class(
     # Slice dataset
     slice = function(idxset, ...) {
 
-      additional_params <- list(...)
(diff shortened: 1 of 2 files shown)


StrikerRUS (approved):
LGTM, thank you!

PR #4582, opened 2021-09-01
[R-package] fix inaccurate comments, remove unnecessary comments

jameslamb:
This PR proposes a few changes to comments in the source code of the R package. Some things I've noticed while looking closely at that code a lot recently, working on #4310.

* removes reference to `lgb.convert()` (was removed about a year ago, in #3269)
   - I also checked for other uses with `git grep -E "lgb\.convert"`...did not find any
* removes comments that do not add any additional information about what the code is doing or why it's doing it (similar to changes in #4383)

for example:

```r
# Check if dataset is not a dataset
if (!lgb.is.Dataset(x = dataset)) {
```

### DIFF
--- a/R-package/R/callback.R
+++ b/R-package/R/callback.R
@@ -365,17 +365,13 @@ cb.early.stop <- function(stopping_rounds, first_metric_only, verbose) {
           # Check if early stopping is required
           if (cur_iter - best_iter[i] >= stopping_rounds) {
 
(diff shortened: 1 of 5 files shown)


StrikerRUS (approved):
Thanks, nice cleanup!
> I also checked for other uses with `git grep -E "lgb\.convert"`...did not find any

Confirmed.

PR #4587, opened 2021-09-02
[R-package] avoid unnecessary computation and add tests for Dataset set_reference() method

jameslamb:
Noticed while working on #4586.

`Dataset$set_reference()` allows changing the reference Dataset that a given `Dataset` is based on, which affects (for example) how features are binned.

Using `Dataset$set_reference()` to set `reference` to the existing object doesn't have any effect on the `Dataset` object. This PR proposes moving the existing check for that all the way up to the beginning of that method, to avoid any other unnecessary computation and slightly speed up this method for that case.

### DIFF
--- a/R-package/tests/testthat/test_dataset.R
+++ b/R-package/tests/testthat/test_dataset.R
@@ -185,6 +185,9 @@ test_that("Dataset$set_reference() updates categorical_feature, colnames, and pr
     dtest$get_colnames()
     , dtrain$get_colnames()
   )
+  expect_false(
+    identical(dtest$get_colnames(), test_original_feature_names)
+  )
 })
 
 test_that("lgb.Dataset: colnames", {
--- a/R-package/tests/testthat/test_dataset.R
+++ b/R-package/tests/testthat/test_dataset.R
@@ -1,5 +1,9 @@
 context("testing lgb.Dataset functionality")
 
+data(agaricus.train, package = "lightgbm")
+train_data <- agaricus.train$data[seq_len(1000L), ]
+train_label <- agaricus.train$label[seq_len(1000L)]
+
 data(agaricus.test, package = "lightgbm")
 test_data <- agaricus.test$data[1L:100L, ]
 test_label <- agaricus.test$label[1L:100L]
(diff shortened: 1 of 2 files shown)


StrikerRUS (commented):
Why do we leave `# Check for empty data` and `# Check for non-existing reference` untouched placed after setting `# Set known references`. I think it is misleading that some attributes are set (`categorical_feature`, `colnames` and `predictor`) but the whole method fails with fatal error. Doesn't it leave object in a corrupted, ok maybe inconsistent, state?

Moreover, I might be wrong but 
```
self$set_categorical_feature(categorical_feature = reference$.__enclos_env__$private$categorical_feature)
```
will simply crash without 
```
        # Reference is unknown
        if (!lgb.is.Dataset(reference)) {
          stop("set_reference: Can only use lgb.Dataset as a reference")
        }
```
placed before in case of wrong reference type.

jameslamb:
Ah yes, you're right! Didn't notice that those should also be moved up sooner. I'll make that change here.

jameslamb:
This refactoring is also making me realize that that method doesn't protect against the case where you run `set_reference(reference = NULL)`. Right now on `master`, that results in

```text
Error in self$set_colnames(colnames = reference$get_colnames()) : 
  attempt to apply non-function
```

I'll fix that here too

jameslamb:
Ok, just moved these validations up further. I realized while doing this that there are currently no direct tests on `Dataset$set_reference()`, so added some of to be more confident in this change.

You can confirm that there are no direct tests on that method by running the following on `master`.

```shell
git grep -E "set.*ref" R-package/tests
```

R-package/tests/testthat/test_dataset.R
```diff
@@ -74,6 +78,115 @@ test_that("Dataset$slice() supports passing Dataset attributes through '...'", {
   expect_identical(dsub1$getinfo("init_score"), init_score)
 })
 
+test_that("Dataset$set_reference() on a constructed datasetfails if raw data has been freed", {
```
StrikerRUS, line 81:
```suggestion
test_that("Dataset$set_reference() on a constructed Dataset fails if raw data has been freed", {
```

R-package/tests/testthat/test_dataset.R
```diff
@@ -74,6 +78,115 @@ test_that("Dataset$slice() supports passing Dataset attributes through '...'", {
...
+    , dtrain$.__enclos_env__$private$predictor$.__enclos_env__$private$handle
+  )
+  expect_identical(
+    dtest$.__enclos_env__$private$categorical_feature
+    , dtrain$.__enclos_env__$private$categorical_feature
+  )
+  expect_identical(
+    dtest$get_colnames()
+    , dtrain$get_colnames()
+  )
```
StrikerRUS, line 187: Should we also check that column names differ from `test_original_feature_names` at this point?

StrikerRUS (approved):
With this PR, the order of code lines in `set_reference()` function looks good to me, thanks!
Just two minor comments regarding new tests.

R-package/tests/testthat/test_dataset.R
jameslamb: good idea! That would make this a bit safer. Added:

```suggestion
  expect_identical(
    dtest$get_colnames()
    , dtrain$get_colnames()
  )
  expect_false(
    identical(dtest$get_colnames(), test_original_feature_names)
  )
```

PR #4588, opened 2021-09-02
[R-package] fix inaccurate error message in Dataset get_colnames() method

jameslamb:
Noticed while working on #4586.

An error message in `Dataset$get_colnames()` currently references `Dataset$dim()`, which is inaccurate and might lead to confusion.

This PR fixes that error message.

### DIFF
--- a/R-package/R/lgb.Dataset.R
+++ b/R-package/R/lgb.Dataset.R
@@ -410,7 +410,7 @@ Dataset <- R6::R6Class(
 
       } else {
 
-        # Trying to work with unknown dimensions is not possible
+        # Trying to work with unknown formats is not possible
         stop(
           "Dataset$get_colnames(): cannot get column names before dataset has been constructed, please call "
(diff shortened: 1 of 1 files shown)


R-package/R/lgb.Dataset.R
```diff
@@ -412,8 +412,8 @@ Dataset <- R6::R6Class(
 
         # Trying to work with unknown dimensions is not possible
```
StrikerRUS, line 413: Left after `Dataset$dim()` too?
```suggestion
        # Trying to work with unknown formats is not possible
```

StrikerRUS (approved):
Keen eye, LGTM!

R-package/R/lgb.Dataset.R
jameslamb: blegh yes, definitely. Thank you! I shouldn't work on these things late at night 😆

PR #4592, opened 2021-09-04
[ci] simplify docker info parsing in QEMU builds

StrikerRUS:
Use `--format` option insted of parsing the whole output with `jq` tool.
https://docs.docker.com/engine/reference/commandline/inspect/#options


jameslamb (approved):
ooooo really nice change! I always forget that the `docker` commands support this result formatting. Checked the QEMU build logs and confirmed this is working.

PR #4604, opened 2021-09-12
[python] rename `print_evaluation()` into `log_evaluation()`

StrikerRUS:
In response to https://github.com/microsoft/LightGBM/pull/4574#discussion_r699411060.


jameslamb (approved):
works for me! Thanks for adding the deprecation warning.

PR #4606, opened 2021-09-15
[docs] update link to MinGW-w64 site

StrikerRUS:
Current link is down (404).
Found this by checking https://github.com/microsoft/LightGBM/actions/workflows/linkchecker.yml.


jameslamb (approved):
thanks for checking!

StrikerRUS:
@shiyu1994 Could you please help with CUDA machine if you have a direct access to it? All CUDA builds are failing with 
```
docker: Error response from daemon: OCI runtime create failed: container_linux.go:380: starting container process caused: process_linux.go:545: container init caused: Running hook #0:: error running hook: exit status 1, stdout: , stderr: nvidia-container-cli: initialization error: nvml error: driver not loaded: unknown.
```

I can't find anything helpful that is not related to WLS, but guess we can start from updating the driver with rebooting.

PR #4608, opened 2021-09-16
allow inclusion in C programs

drewmiller:
Including "c_api.h" from programs compiled with C compilers errors due to required inclusion of C++ headers. These edits allow c_api.h to be included directly in C programs.


jameslamb:
Thanks for this @drewmiller !

Is it possible to share a link to the C project you're working on that needs to link against LightGBM? No worries if not, we'll understand if it's proprietary.

Also would you be willing to write up a feature request at https://github.com/microsoft/LightGBM/issues explaining why it would be valuable for C projects to be able to link against LightGBM directly, with specific examples of such projects if you have them? If this is a feature that people care about, we'll want to add a continuous integration job that tests that linking such a project works, to ensure that compatibility issues aren't accidentally introduced in the future.

drewmiller:
> Thanks for this @drewmiller !
> 
> Is it possible to share a link to the C project you're working on that needs to link against LightGBM? No worries if not, we'll understand if it's proprietary.

Unfortunately, it's proprietary.

> 
> Also would you be willing to write up a feature request at https://github.com/microsoft/LightGBM/issues explaining why it would be valuable for C projects to be able to link against LightGBM directly, with specific examples of such projects if you have them? If this is a feature that people care about, we'll want to add a continuous integration job that tests that linking such a project works, to ensure that compatibility issues aren't accidentally introduced in the future.

Sure.

drewmiller:
Thanks for the great product, MS team. I've added a simple feature request, and I'm happy to help write a toy example for inclusion in a unit testing pipeline if helpful. I'm not super familiar with the code base yet, so I recognize my "trivial" commit may have cascading effects. However, this software would be more useful to me with this feature, so I thought I'd open the door with some proposed changes, and I'm happy to help bring them into existence however I can. That includes bulking out the feature request as necessary. Warm regards.

jameslamb:
Thanks very much for opening #4609. I think it's something we should consider supporting and should having  tests for, but that work doesn't need to be done on this PR.

Could you please merge the latest `master` into this branch? We recently had to turn off the CUDA CI jobs due to some infrastructure problems (#4611). Sorry for the inconvenience.

drewmiller:
> Thanks very much for opening #4609. I think it's something we should consider supporting and should having tests for, but that work doesn't need to be done on this PR.
> 
> Could you please merge the latest `master` into this branch? We recently had to turn off the CUDA CI jobs due to some infrastructure problems (#4611). Sorry for the inconvenience.

Done; although the CI seems to be hanging at `all-successful`

StrikerRUS:
> although the CI seems to be hanging at `all-successful`.

This is a GitHub protection from miners: https://github.com/microsoft/LightGBM/pull/4266#issuecomment-837346143. We have to click `Approve and run` button for PRs from a first-time contributors after each their commit.

include/LightGBM/c_api.h
```diff
@@ -1321,7 +1328,9 @@ LIGHTGBM_C_EXPORT int LGBM_NetworkInitWithFunctions(int num_machines,
                                                     void* reduce_scatter_ext_fun,
                                                     void* allgather_ext_fun);
 
-#if defined(_MSC_VER)
+#ifndef __cplusplus
+#define THREAD_LOCAL  /*!< \brief Thread local specifier no-op in C builds. */
```
guolinke, line 1332: It seems C11 also has thread_local, refer to https://en.wikipedia.org/wiki/C11_(C_standard_revision)

drewmiller: The follow-up changes take advantage of `_Thread_local` from C11. They also make the C API header compatible with ANSI C. If the changes to the comments disrupt doc generation, they could be rolled back in favor of abandoning support for C versions prior to C99. The attached [test.c.gz](https://github.com/microsoft/LightGBM/files/7229809/test.c.gz) can be gunzipped in `c_api_test/` and used to test both code generation and compilation with the below:

Code generation:
`gcc test.c -ansi -I../../include/ -E` - should exclude `inline` and `_Thread_local`
`gcc test.c -std=c99 -I../../include/ -E` - should include `inline` and exclude `_Thread_local`
`gcc test.c -std=c11 -I../../include/ -E` - should include `inline` and `_Thread_local`
`g++ test.c -I../../include/ -E` - should include `inline` and `thread_local`

Remove the `-E` switch on the above to confirm compilation for each setting.

guolinke:
@shiyu1994  can you take a look on this?

shiyu1994 (approved):
@drewmiller Thanks for your contribution. The changes LGTM.

Issue #4610, opened 2021-09-17
CUDA CI jobs are broken in `master`


StrikerRUS:
Initially reported in https://github.com/microsoft/LightGBM/pull/4606#issuecomment-921374740.
```
docker: Error response from daemon: OCI runtime create failed: container_linux.go:380: starting container process caused: process_linux.go:545: container init caused: Running hook #0:: error running hook: exit status 1, stdout: , stderr: nvidia-container-cli: initialization error: nvml error: driver not loaded: unknown.
```

<details>

<summary>Full logs:</summary>

```
2021-09-15T23:12:00.6215072Z Can't find any online and idle self-hosted runner in the current repository, account/organization and enterprise that matches the required labels: 'self-hosted , linux'
2021-09-15T23:12:00.7256411Z Waiting for a self-hosted runner to pickup this job...
2021-09-15T23:12:47.4157030Z Job is about to start running on the runner: nv6-01 (repository)
2021-09-15T23:12:52.0695982Z Current runner version: '2.282.0'
2021-09-15T23:12:52.0700468Z Runner name: 'nv6-01'
2021-09-15T23:12:52.0701070Z Runner group name: 'Default'
2021-09-15T23:12:52.0702313Z Machine name: 'nv6-01'
2021-09-15T23:12:52.0705689Z ##[group]GITHUB_TOKEN Permissions
2021-09-15T23:12:52.0706967Z Actions: write
2021-09-15T23:12:52.0707553Z Checks: write
2021-09-15T23:12:52.0708038Z Contents: write
2021-09-15T23:12:52.0708549Z Deployments: write
2021-09-15T23:12:52.0709026Z Discussions: write
2021-09-15T23:12:52.0709541Z Issues: write
2021-09-15T23:12:52.0710017Z Metadata: read
2021-09-15T23:12:52.0710512Z Packages: write
2021-09-15T23:12:52.0710989Z PullRequests: write
2021-09-15T23:12:52.0711583Z RepositoryProjects: write
2021-09-15T23:12:52.0712180Z SecurityEvents: write
2021-09-15T23:12:52.0712745Z Statuses: write
2021-09-15T23:12:52.0713361Z ##[endgroup]
2021-09-15T23:12:52.0716281Z Prepare workflow directory
2021-09-15T23:12:52.1453868Z Prepare all required actions
2021-09-15T23:12:52.1463663Z Getting action download info
[...]

PR #4611, opened 2021-09-17
[ci] Stop running CUDA tests at CI

StrikerRUS:
Unblock LightGBM development process.
Refer to #4610.


jameslamb (approved):
Thanks for this and for opening the issue!

I agree with this change, since it at least allows us to make forward progress until @shiyu1994 is available to help work through the issues on those CUDA runners.

I do think though that we should consider #4610 a blocker to the 3.3.0 release (#4310). I hate to delay that release even further but I wouldn't feel comfortable with doing a new release that contains changes which haven't been tested with CUDA-based builds.

If the self-hosted runners continues to be a problem, maybe we can talk with @quasiben and friends at NVIDIA about getting some help for LightGBM's CI similar to the stuff discussed in https://github.com/dask/community/issues/138.

PR #4613, opened 2021-09-17
[R-package] prevent memory leak if pointer fails to allocate 

david-cortes:
As a follow up from https://github.com/microsoft/LightGBM/pull/4597, this PR adds preventions in case an R external object pointer object fails to allocate. This should be an incredibly uncommon situation (failing to allocate 8 bytes), but it is in any case preventable.

### DIFF
--- a/R-package/src/lightgbm_R.cpp
+++ b/R-package/src/lightgbm_R.cpp
@@ -94,7 +94,7 @@ SEXP LGBM_DatasetCreateFromFile_R(SEXP filename,
   SEXP parameters,
   SEXP reference) {
   R_API_BEGIN();
-  SEXP ret;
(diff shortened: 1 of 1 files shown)


david-cortes:
The failing checks don't show  any errors.

jameslamb:
On https://github.com/microsoft/LightGBM/runs/3669905286?check_suite_focus=true, I see

> Error: The operation was canceled.

And that the job runtime was 3 hours, which is the maximum we allow.

https://github.com/microsoft/LightGBM/blob/c3b60c289b27cb8ded7dd67a2aa18017e9e1ac64/.github/workflows/r_valgrind.yml#L10

This happens sometimes and is usually something transient related to the infrastructure behind GitHub Actions. I'll re-run the valgrind test.

jameslamb (approved):
Looks like all the tests are passing, and I understand what's going on in this change. Thanks for the help!

PR #4614, opened 2021-09-18
[python] add placeholders to titles in plotting functions

StrikerRUS:
Similarly to `title` argument in `plot_split_value_histogram()` function:

https://github.com/microsoft/LightGBM/blob/d10f9d43c490d3e49afc4fdcb72a129cfaf70777/python-package/lightgbm/plotting.py#L199-L205

### DIFF
--- a/python-package/lightgbm/plotting.py
+++ b/python-package/lightgbm/plotting.py
@@ -62,6 +62,7 @@ def plot_importance(
     xlabel : str or None, optional (default="Feature importance")
         X-axis title label.
         If None, title is disabled.
+        @importance_type@ placeholder can be used, and it will be replaced with the value of ``importance_type`` parameter.
     ylabel : str or None, optional (default="Features")
         Y-axis title label.
         If None, title is disabled.
@@ -150,6 +151,7 @@ def plot_importance(
     if title is not None:
         ax.set_title(title)
     if xlabel is not None:
+        xlabel = xlabel.replace('@importance_type@', importance_type)
         ax.set_xlabel(xlabel)
     if ylabel is not None:
         ax.set_ylabel(ylabel)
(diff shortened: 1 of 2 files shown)


python-package/lightgbm/plotting.py
```diff
@@ -370,14 +373,17 @@ def plot_metric(
         if metric not in metrics_for_one:
             raise KeyError('No given metric in eval results.')
         results = metrics_for_one[metric]
-    num_iteration, max_result, min_result = len(results), max(results), min(results)
+    num_iteration = len(results)
+    max_result = max(results)
+    min_result = min(results)
```
jameslamb, line 378: thank you for breaking this up!

tests/python_package_test/test_plotting.py
```diff
@@ -68,6 +68,13 @@ def test_plot_importance(params, breast_cancer_split, train_data):
     assert ax2.patches[2].get_facecolor() == (0, .5, 0, 1.)  # g
     assert ax2.patches[3].get_facecolor() == (0, 0, 1., 1.)  # b
 
+    ax3 = lgb.plot_importance(gbm0, title='t @importance_type@', xlabel='x @importance_type@', ylabel='y @importance_type@')
+    assert isinstance(ax3, matplotlib.axes.Axes)
+    assert ax3.get_title() == 't @importance_type@'
```
jameslamb, line 73: I'm confused by these tests. Shouldn't the template string `@importance_type@` have been replaced with the actual value of `importance_type`?

jameslamb (commented):
I support this change but I don't understand the tests, left a question about it.

tests/python_package_test/test_plotting.py
StrikerRUS: Only in `xlabel` argument, not in `title` or any other.
https://github.com/microsoft/LightGBM/blob/6714a0f7fcfb2f5f0ea1adfa4c45b4ef0501098c/python-package/lightgbm/plotting.py#L62-L65

jameslamb: ah I see, think I was confused by changes to the multiple different functions. Thanks for the explanation.

jameslamb (approved):
very nice addition, thanks!

PR #4615, opened 2021-09-18
[python] fix mypy error in `dask.py`

StrikerRUS:
Contributes to #3867.

Fixed
```
python-package\lightgbm\dask.py:33: error: First argument to namedtuple() should be "_HostWorkers", not "HostWorkers"
```


jmoralez (approved):

jameslamb (approved):
ah nice, thanks!

PR #4624, opened 2021-09-23
[python] deprecate "auto" value of `ylabel` argument of `plot_metric()` function

StrikerRUS:
Continuation of #4614.
With `@metric@` placeholder we don't need to maintain `'auto'`.


jameslamb (approved):
excellent!

guolinke (approved):

PR #4628, opened 2021-09-25
[ci] bump CUDA version from 11.4.0 to 11.4.2 at CI

StrikerRUS:
Starting from `11.4.1` version you have to specify OS as a part of image.
https://ngc.nvidia.com/catalog/containers/nvidia:cuda/tags


jameslamb (approved):
looks good, thanks for the link explaining the extra changes

PR #4636, opened 2021-09-30
[python][sklearn] add `__sklearn_is_fitted__()` method to be better compatible with scikit-learn API

StrikerRUS:
Refer to scikit-learn 1.0 release notes and corresponding PR:
> `Enhancement` `utils.validation.check_is_fitted` now uses `__sklearn_is_fitted__` if available, instead of checking for attributes ending with an underscore.
   https://scikit-learn.org/stable/whats_new/v1.0.html#sklearn-utils
   https://github.com/scikit-learn/scikit-learn/pull/20657

After doing some search over GitHub repos, it looks like that only XGBoost has already adopted that new API:
https://github.com/dmlc/xgboost/pull/7230

### DIFF
--- a/python-package/lightgbm/sklearn.py
+++ b/python-package/lightgbm/sklearn.py
@@ -529,6 +529,9 @@ class LGBMModel(_LGBMModelBase):
             }
         }
 
+    def __sklearn_is_fitted__(self) -> bool:
+        return getattr(self, "fitted_", False)
+
     def get_params(self, deep=True):
         """Get parameters for this estimator.
 


python-package/lightgbm/sklearn.py
```diff
@@ -529,6 +529,9 @@ def _more_tags(self):
             }
         }
 
+    def __sklearn_is_fitted__(self) -> bool:
+        return getattr(self, "fitted_", False)
```
StrikerRUS, line 533: We set `fitted_` attribute at the end of the `fit()` method
https://github.com/microsoft/LightGBM/blob/a77260f03eb735923125024b0f36df9f63a6e0f3/python-package/lightgbm/sklearn.py#L770

jameslamb:
> Are we facing the following issue at our CI services? https://news.yahoo.com/millions-old-phones-laptops-smart-210600565.html

oh wow! maybe!

> Err:10 https://apt.kitware.com/ubuntu bionic Release
  Certificate verification failed: The certificate is NOT trusted. The certificate chain uses expired certificate.  Could not handshake: Error in the certificate verification. [IP: 66.194.253.25 443]

python-package/lightgbm/sklearn.py
jameslamb: I noticed in the https://github.com/dmlc/xgboost/pull/7230, they also replaced some other internal checks of estimator attributes with this method call.

I think that's a good idea, so only this method needs to know about the `fitted_` property. What do you think?

That would mean changing the following to `if self.__sklearn_is_fitted__()`:

https://github.com/microsoft/LightGBM/blob/a77260f03eb735923125024b0f36df9f63a6e0f3/python-package/lightgbm/dask.py#L1004-L1005

jameslamb (approved):
thanks for catching this in the `scikit-learn` release notes! Totally agree.

Just left one comment for your consideration.

python-package/lightgbm/sklearn.py
StrikerRUS: I agree. I remember you were asking why do we use `if self._n_features is None` everywhere for checking that estimator is fitted when we have `self.fitted_`.
https://github.com/microsoft/LightGBM/blob/a77260f03eb735923125024b0f36df9f63a6e0f3/python-package/lightgbm/sklearn.py#L857
https://github.com/microsoft/LightGBM/pull/3883#discussion_r568859996
My answer was that it's because `self.fitted_` was introduced much later. Now we can replace everything with `self.__sklearn_is_fitted__()`. And I think it will be clearer.

But I'd prefer to do it in a follow-up PR, if you do not mind.

jameslamb: ok yep, I'm fine with it being a followup

jameslamb:
@StrikerRUS I've open #4646 to consolidate the discussion about the certificate issues for CUDA CI jobs (since they aren't specific to this PR).

jk7ss:
This [change](https://github.com/microsoft/LightGBM/commit/4b140bcc97090abf353fb681b9de8807d367f70f) 

```
    def __sklearn_is_fitted__(self) -> bool:
        return getattr(self, "fitted_", False)
```

that defaults to `False`, together with [this test](https://github.com/microsoft/LightGBM/blame/2394b41ac2bc64c36115c4eb314633dede0c6b9c/python-package/lightgbm/sklearn.py#L794):

```python
    def predict(...):
        if not self.__sklearn_is_fitted__():
            raise LGBMNotFittedError("Estimator not fitted, call fit before exploiting the model.")
```

renders all models that were trained and serialized prior to `v3.3.0` failing prediction after deserialization, because they are missing the `fitted_` attribute.
Shouldn't it either
a) cause `AttributeError` if it is missing -> meaning that the model object is not code compatible? or
b) silently expect it's fitted to be backward compatible? 

It's a reasonable assumption that a deserialized model was already fitted?

StrikerRUS:
@jk7ss Thank you very for your report! This is indeed important to be backward compatible with serialized models fitted in previous versions of LightGBM. But `fitted_` attribute was added in #3329 and that PR was included in the[ `v3.1.0` release](https://github.com/microsoft/LightGBM/releases/tag/v3.1.0) (16 Nov 2020). So I guess everything should be OK for models fitted and serialized with versions started from `v3.1.0`. Do you think we should do something to be backward compatible even for models trained and serialized prior to `v3.1.0`?

PR #4639, opened 2021-10-01
[tests][python-package] refactor list_to_1d_numpy test to run without pandas installed

### LINKED ISSUE #4638: [python-package] Tests not running without panda  (tests.python_package_test/test_basic.py)
## Description
If running pytest without an installation of pandas no test is executed and the run is stopped with the following error message:

```python_package_test/test_basic.py:None (python_package_test/test_basic.py)
python_package_test/test_basic.py:519: in <module>
    ???
E   TypeError: pd_Series() takes no arguments
```

In that line, we can find a test that seems to be supposedly skipped if pandas is not installed. But as far as I know pytest.mark.parametrize is evaluated anyways, making the whole test suit crash. I'm honestly not sure if this is an error on my side. But it looks like an easy thing to oversee if you have pandas installed by default, so I thought I should mention it here. 

```python
@pytest.mark.skipif(not PANDAS_INSTALLED, reason='pandas is not installed')
@pytest.mark.parametrize(
    'y',
    [
        np.random.rand(10),
        np.random.rand(10, 1),
        pd_Series(np.random.rand(10)),
        pd_Series(['a', 'b']),
        [1] * 10,
        [[1], [2]]
    ])
@pytest.mark.parametrize('dtype', [np.float32, np.float64])
def test_list_to_1d_numpy(y, dtype):
    if isinstance(y, np.ndarray) and len(y.shape) == 2:
        with pytest.warns(UserWarning, match='column-vector'):
            lgb.basic.list_to_1d_numpy(y)
        return
    elif isinstance(y, list) and isinstance(y[0], list):
        with pytest.raises(TypeError):
            lgb.basic.list_to_1d_numpy(y)
        return
    elif isinstance(y, pd_Series) and y.dtype == object:
        with pytest.raises(ValueError):
            lgb.basic.list_to_1d_numpy(y)
        return
    result = lgb.basic.list_to_1d_numpy(y, dtype=dtype)
    assert result.size == 10
    assert result.dtype == dtype
```

## Reproducible example

```shell
pytest tests/python_package_test
```
## Environment info

3.2.1.99

```shell
# install the Python package
git submodule init
git submodule update --recursive
cd python-package
python setup.py install
```

jmoralez:
Currently `tests/python-package-test/test_basic.py::test_list_to_1d_numpy` uses `pd_Series` within the parametrize calls, which fails when pandas isn't installed. This moves the series creation after checking that pandas is indeed installed and we can be sure that `pd_Series(collection)` will succeed.

Fixes #4638.

### DIFF
--- a/tests/python_package_test/test_basic.py
+++ b/tests/python_package_test/test_basic.py
@@ -510,19 +510,23 @@ def test_choose_param_value():
     assert original_params == expected_params
 
 
-@pytest.mark.skipif(not PANDAS_INSTALLED, reason='pandas is not installed')
-@pytest.mark.parametrize(
-    'y',
-    [
-        np.random.rand(10),
-        np.random.rand(10, 1),
-        pd_Series(np.random.rand(10)),
-        pd_Series(['a', 'b']),
-        [1] * 10,
-        [[1], [2]]
-    ])
+@pytest.mark.parametrize('collection', ['1d_np', '2d_np', 'pd_float', 'pd_str', '1d_list', '2d_list'])
 @pytest.mark.parametrize('dtype', [np.float32, np.float64])
-def test_list_to_1d_numpy(y, dtype):
+def test_list_to_1d_numpy(collection, dtype):
+    collection2y = {
+        '1d_np': np.random.rand(10),
+        '2d_np': np.random.rand(10, 1),
+        'pd_float': np.random.rand(10),
+        'pd_str': ['a', 'b'],
+        '1d_list': [1] * 10,
+        '2d_list': [[1], [2]],
+    }
+    y = collection2y[collection]
+    if collection.startswith('pd'):
+        if not PANDAS_INSTALLED:
+            pytest.skip('pandas is not installed')
+        else:
+            y = pd_Series(y)
     if isinstance(y, np.ndarray) and len(y.shape) == 2:
         with pytest.warns(UserWarning, match='column-vector'):
             lgb.basic.list_to_1d_numpy(y)


tests/python_package_test/test_basic.py
```diff
@@ -510,19 +510,23 @@ def test_choose_param_value():
...
-    'y',
-    [
-        np.random.rand(10),
-        np.random.rand(10, 1),
-        pd_Series(np.random.rand(10)),
-        pd_Series(['a', 'b']),
-        [1] * 10,
-        [[1], [2]]
-    ])
+@pytest.mark.parametrize('collection', ['1d_np', '2d_np', 'pd_float', 'pd_str', '1d_list', '2d_list'])
```
strobelTha, line 513: I fear this construct might be hard to understand if the reason behind it is unknown. 
Would it be an option to not change the test at all, but adapt the dummy implementation for pandas in lightgbm/compat.py

Current:
```python
try:
    from pandas import DataFrame as pd_DataFrame
    from pandas import Series as pd_Series
    from pandas import concat
    from pandas.api.types import is_sparse as is_dtype_sparse
    PANDAS_INSTALLED = True
except ImportError:
    PANDAS_INSTALLED = False

    class pd_Series:  # type: ignore
        """Dummy class for pandas.Series."""

        pass

    class pd_DataFrame:  # type: ignore
        """Dummy class for pandas.DataFrame."""

        pass
```

Proposal:
```python
try:
    from pandas import DataFrame as pd_DataFrame
    from pandas import Series as pd_Series
    from pandas import concat
    from pandas.api.types import is_sparse as is_dtype_sparse
    PANDAS_INSTALLED = True
except ImportError:
    PANDAS_INSTALLED = False

    class pd_Series:  # type: ignore
        """Dummy class for pandas.Series."""
        def __init__(self, *args):
            pass

    class pd_DataFrame:  # type: ignore
        """Dummy class for pandas.DataFrame."""
        def __init__(self, *args):
            pass
```

StrikerRUS: I agree with @strobelTha . I believe that making all our dummy classes from `compat.py` creatable with any combination of arguments by adding catch-all constructor will be useful
```python
class pd_Series:  # type: ignore
    """Dummy class for pandas.Series."""
    def __init__(self, *args, **kwargs):
        pass
```

jmoralez: I like the refactor because of the following reasons:
* Moving the collection definition inside the function instead of having that big parametrize seems cleaner.
* Now the test will run for lists and numpy arrays when pandas isn't installed, whereas previously the whole test was (supposed to be) skipped.
* In case the test fails the message is more descriptive, i.e. `FAILED tests/python_package_test/test_basic.py::test_list_to_1d_numpy[float32-2d_np]` vs `FAILED tests/python_package_test/test_basic.py::test_list_to_1d_numpy[float64-y1]`.

> ```
> class pd_Series:  # type: ignore
>     """Dummy class for pandas.Series."""
>     def __init__(self, *args, **kwargs):
>         pass
> ```

@strobelTha would you like to contribute those changes?

StrikerRUS: @jmoralez I liked reasons `#2` and `#3` you listed above. Thanks a lot for emphasizing advantages of the current refactor.

And indeed, we can add constructors for our dummy classes in a separate PR.

StrikerRUS (approved):
LGTM!
Refer to https://github.com/microsoft/LightGBM/pull/4639#discussion_r721757280 for my more detailed comment.

PR #4528, opened 2021-08-16
WIP: [CUDA] New CUDA version

shiyu1994:
Current GPU and CUDA versions of LightGBM only calculate histograms on GPU. This is a new CUDA version for LightGBM which runs the whole training process on GPU.

PR #2485, opened 2019-10-01
reduce the buffer when using high dimensional data in distributed mode.

guolinke:
to fix #2484

PR #4643, opened 2021-10-02
remove unused BinMapper::SizeForSpecificBin()

jameslamb:
`cppcheck` shows the following warning (https://github.com/microsoft/LightGBM/issues/4539#issuecomment-903036082)

> src/io/bin.cpp:523:0: style: The function 'SizeForSpecificBin' is never used. [unusedFunction]

I confirmed that this method is not called anywhere in the library, using `git grep SizeForSpecificBin`. I think it's safe to remove, since only the code in `src/c_api.cpp` is considered parrt of the public API for LightGBM.

I also double-checked and don't see this method referenced in the in-progress CUDA work (#4528, #4630).

### Notes for Reviewers

I checked the git blame to see if this method had been added for a specific purpose in the past, but don't see anything like that. It was part of the first commit on the project.

https://github.com/microsoft/LightGBM/blame/7a82ba4ffcca8530e53b75f148429978653a57c3/include/LightGBM/bin.h#L153


StrikerRUS (approved):
Thanks for the cleanup!
The usage of this method was dropped in #2485:
https://github.com/microsoft/LightGBM/pull/2485/files#diff-85e23cbc42ffa687f2c8a01ec3cb414cfe99b95a72d3df5d45b10bdc523f6784L684

shiyu1994 (approved):

PR #4650, opened 2021-10-05
remove unused DCGCalculator::CalDCGAtK()

jameslamb:
`cppcheck` shows the following warning (https://github.com/microsoft/LightGBM/issues/4539#issuecomment-903036082)

> src/metric/dcg_calculator.cpp:110:0: style: The function 'CalDCGAtK' is never used. [unusedFunction]

I confirmed that this method is not called anywhere in the library, using `git grep CalDCGAtK`. I think it's safe to remove, since only the code in `src/c_api.cpp` is considered parrt of the public API for LightGBM.

I also double-checked and don't see this method referenced in the in-progress CUDA work (#4528, #4630).

### Notes for Reviewers

I checked the git blame to see if this method had been added for a specific purpose in the past, but don't see anything like that. It was part of the first commit on the project.

https://github.com/microsoft/LightGBM/blame/e95d5ab849f046b8fdf755441e5767094aeb8de2/src/metric/dcg_calculator.cpp#L56


shiyu1994 (approved):
Approve given all the checks have been passed.

StrikerRUS (approved):

PR #4668, opened 2021-10-09
[doc] Update GPU-Windows.rst, missing bootstrap.bat gcc

tmacleod:
Added line 190, "bootstrap.bat gcc"
This step is necessary in order to complete the following steps of "b2 install...". b2 will not be present in C:\boost\boost_1_63_0 unless "bootstrap.bat gcc" is run there first (this is in addition to the "bootstrap.bat gcc" in line 187).
I have also added another "bootstrap" to line 241 (and missing parenthesis) to make it match the update.

### DIFF
--- a/docs/GPU-Windows.rst
+++ b/docs/GPU-Windows.rst
@@ -187,6 +187,7 @@ We can now start downloading and compiling the required Boost libraries:
        bootstrap.bat gcc
        b2 install --prefix="C:\boost\boost-build" toolset=gcc
        cd C:\boost\boost_1_63_0
+       bootstrap.bat gcc
 
 To build the Boost libraries, you have two choices for command prompt:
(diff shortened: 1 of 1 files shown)


jameslamb:
@huanzhang12 are you able to help review this? I don't have easy access to an environment to test these suggested changes.

StrikerRUS:
I just checked this and was able to successfully build LightGBM GPU with and without the proposed line.
`b2.exe` is created with the first call of `bootstrap.bat gcc` and is located in `C:\boost\boost-build\bin`.

image

According to the guide, that location should be added to the `PATH` variable, so `b2.exe` is accessible from any folder after that.

image

jameslamb:
Ah great, thanks @StrikerRUS

PR #4678, opened 2021-10-14
[ci] introduce CI jobs that mimic CRAN gcc-ASAN and clang-ASAN tests (fixes #4674)

jameslamb:
#4674 documents an issue in LightGBM discovered by CRAN's checks using the address sanitizer for `gcc` and `clang`.

This PR proposes introducing CI jobs that I believe can reproduce that issue, and which I hope will help us to catch similar issues during development in the future (to reduce the risk of CRAN rejections).

* comment explaining the design choices in these tests: https://github.com/microsoft/LightGBM/issues/4674#issuecomment-942922289
* builds on my fork showing these checks reproducing #4674
    - `gcc`: https://github.com/jameslamb/LightGBM/pull/49/checks?check_run_id=3890682813
    - `clang`: https://github.com/jameslamb/LightGBM/pull/49/checks?check_run_id=3890682831

PR #4686, opened 2021-10-15
[R-package] Add `print()` and `summary()` methods for Booster

david-cortes:
Currently, when examining a lightgbm object, it shows information which is typically not relevant to the user, while missing important pieces of information that one would want to see, such as the objective function or the model dimensions. This PR adds a `print` and a `summary` method with shorter outputs and more relevant information. The old `print` can still be accessed through `str`.


jameslamb:
That message from this project's linter exactly mirrors direct feedback we received from CRAN.

In one submission of `{lightgbm}` 3.0.0 to CRAN in September 2020, CRAN's response included the following (documented [in cran-comments.md](https://github.com/microsoft/LightGBM/blob/master/R-package/cran-comments.md#v300---submission-4---september-4-2020.))

> Please replace cat() by message() or warning() in your functions (except for print() and summary() functions). Messages and warnings can be suppressed if needed.

Since this PR is adding `print()` and `summary()` methods, I think uses of `cat()` in those specific cases will be ok with CRAN.

According to https://github.com/jimhester/lintr#project-configuration, it's possible to tell `{lintr}` to exclude specific lines by adding the comment `# nolint`. Please try adding such comments to the lines in this PR that use `cat()`.

jameslamb (approved):
This looks great, thanks so much for the idea and for the effort you put into the implementation!

## Notes for other reviewers

I just updated this to latest `master`, to integrate other recent changes, especially the new CI jobs mimicking CRAN's tests with sanitizers (#4678).

If those pass, I'm comfortable merging this now that v3.3.1 has been released (#4715). Even if CRAN rejects v3.3.1 for some reason and we end up needing a v3.3.2 release, I'd be comfortable with these changes being part of such a release, since I think they're low risk and do not contain any breaking changes.

StrikerRUS:
@jameslamb Can we merge this now?

jameslamb:
yep! I'll merge it. Thanks again for the help @david-cortes !

PR #4687, opened 2021-10-15
[R-package] Expand user paths in file names

david-cortes:
Many R functions  take file paths as inputs, but will not expand paths like `~/myfile`, which this PR fixes. It comes without any tests because generating files in the user folder in examples and tests would be against CRAN policies.


jameslamb:
@StrikerRUS I'm going to merge this one.

I think it's a good candidate for a potential 3.3.1 release (since it seems #4674 will probably require such a release) as it improves the usability of the R package in a way that doesn't introduce breaking changes, large new functionality, new dependencies, or substantial risk of submission issues with CRAN.

PR #4688, opened 2021-10-16
[docs] fix C API docs rendering

StrikerRUS:
Refer to 
- https://www.doxygen.nl/manual/docblocks.html#memberdoc
- https://www.doxygen.nl/manual/preprocessing.html

I checked all variants of `brief` description and only this one results in no weird artifacts in the `LGBM_SetLastError()` function.

Force `__cplusplus` in docs to get more general description.

Live demo: https://lightgbm.readthedocs.io/en/fix_docs/C-API.html.

#### Before

image

image

#### After

image

image


drewmiller:
@StrikerRUS Note that the use of `//` isn't supported by ANSI C. So if there's no workaround to using `//` in 'c_api.h', ANSI C support will need to be dropped.

Issue #4537, opened 2021-08-19
Improve portability of OpenMP checks in R-package configure on macOS


StrikerRUS:
Originally discussed in https://github.com/microsoft/LightGBM/pull/4507#discussion_r692334224.

Right now, `configure.ac` checks compilation on macOS with `-Xclang -fopenmp` OpenMP flags. In case of failure, it tries to compile one more time with default `-fopenmp` flag. I think we can improve that check by trying to compile with some other OpenMP flags with the aim to support more compilers.

https://github.com/microsoft/LightGBM/blob/c65a2e3363cdff5ddb31689c6d555ec345932d98/R-package/configure.ac#L103

Possible flags to check: https://github.com/autoconf-archive/autoconf-archive/blob/6c25681d78d963708362ea0d4d63ee9b508782b6/m4/ax_openmp.m4#L82.

StrikerRUS:
Closing in favor of being in #2302. We decided to keep all feature requests in one place.

Welcome to contribute this feature! Please re-open (or post a comment if you are not a topic starter) this issue if you are actively working on implementing this feature.

PR #4692, opened 2021-10-17
[R-package] Link to optimized linear algebra libraries

david-cortes:
LightGBM uses Eigen for solving linear systems. Under the default configuration for the R build, it makes it use Eigen's own linear algebra operations, which might not be as efficient as an optimized BLAS/LAPACK library, especially under the default R build options which do **not** use `-march=native` and instead limit to `-msse2` on amd64 computers.

It can be confirmed that it is not linking to either of those by downloading the windows binary build from CRAN and running something like this:
```
dumpbin /dependents lightgbm.dll
```

This PR adds the necessary linkage to use BLAS and LAPACK in the R builds, which can be confirmed by building for windows and re-running the command above.


david-cortes:
The mac setup that is failing is because it cannot link to `gfortran`. Probably needs to install it in order to build the package:
https://cran.r-project.org/bin/macosx/tools/
https://www.cynkra.com/blog/2021-03-16-gfortran-macos/

jameslamb:
Ok, coming back to this!

Given the comments so far, I don't support adding unconditional linking to BLAS/LAPACK.

* this feature would only provide a performance improvement when using linear learners
* the expected improvement is unknown ([comment](https://github.com/microsoft/LightGBM/pull/4692#issuecomment-952983807)), 
* adding this linking unconditionally might require Mac users installing from source to install `gfortran` if they don't have it ([comment](https://github.com/microsoft/LightGBM/pull/4692#issuecomment-945201066))

If we were to add this support in the CRAN package, I'd prefer to do it conditionally, similar to the way this project conditionally uses OpenMP: https://github.com/microsoft/LightGBM/blob/5fa887bb79d0e4d1961c8adafce9832448535e71/R-package/configure.ac#L108.

=================================

Please open a separate feature request documenting related ideas for the Python package, since you've said you don't want to attempt such changes in this PR (https://github.com/microsoft/LightGBM/pull/4692#issuecomment-949930771). To keep the conversation here focused on only the changes in this PR.

david-cortes:
> Ok, coming back to this!
> 
> Given the comments so far, I don't support adding unconditional linking to BLAS/LAPACK.
> 
> * this feature would only provide a performance improvement when using linear learners
> * the expected improvement is unknown ([comment](https://github.com/microsoft/LightGBM/pull/4692#issuecomment-952983807)),
> * adding this linking unconditionally might require Mac users installing from source to install `gfortran` if they don't have it ([comment](https://github.com/microsoft/LightGBM/pull/4692#issuecomment-945201066))
> 
> If we were to add this support in the CRAN package, I'd prefer to do it conditionally, similar to the way this project conditionally uses OpenMP:
> 
> https://github.com/microsoft/LightGBM/blob/5fa887bb79d0e4d1961c8adafce9832448535e71/R-package/configure.ac#L108
> 
> .
> =================================
> 
> Please open a separate feature request documenting related ideas for the Python package, since you've said you don't want to attempt such changes in this PR ([#4692 (comment)](https://github.com/microsoft/LightGBM/pull/4692#issuecomment-949930771)). To keep the conversation here focused on only the changes in this PR.

I think that workaround which you added for openmp on mac is no longer necessary in the latest R versions. It is BTW also not a portable solution, since a user having GCC or ICC as the default compiler will be able to use openmp through the R-provided flags but would fail that test.

jameslamb:
>  also not a portable solution, since a user having GCC or ICC as the default compiler will be able to use openmp through the R-provided flags but would fail that test.

Thanks for the note. For ICC, that's a known issue (@StrikerRUS mentioned it in https://github.com/microsoft/LightGBM/pull/4507#discussion_r692334224), and LightGBM has an item in it's backlog to address it: #4537 .

I don't think that statement is completely correct though, that `gcc` users on Mac wouldn't be able to use openmp with the LightGBM R package.

That file explicitly accounts for that:

https://github.com/microsoft/LightGBM/blob/5fa887bb79d0e4d1961c8adafce9832448535e71/R-package/configure.ac#L123-L131

This project doesn't have tests using `gcc` on Mac for the CRAN package (due to limited CI capacity), but that combination was explicitly tested when this OpenMP check was added: #4507.

If you can produce a reproducible example showing that compiling the CRAN package with `gcc` on Mac fails to find OpenMP, we'd welcome a comment with that evidence on #4507, as it'll be helpful to whoever eventually picks up that feature.

Laurae2:
> adding this linking unconditionally might require Mac users installing from source to install gfortran if they don't have it

Mac users who compile R packages from source are expected to have gfortran installed and usable for R.

> Actually, there's something strange with these timings: if I run it multiple times, further runs then take <1s (wether this is ran with eigen or with blas).

Destroy the model and garbage collect the memory before retraining each model.

Note that you might need to run them once for hot loading memory.

> the expected improvement is unknown

BLAS, Intel MKL, and tuned BLASes should be faster. However it depends on the CPU, available CPU cache, what's currently using the CPU caches, and many other factors.

> > ... R provides an internal BLAS which is well-tested and will be adequate for most uses of R.
https://cran.r-project.org/doc/manuals/r-devel/R-admin.html#Linear-algebra
> 
> However, some articles over the internet state that the real speedup can be provided by external BLAS implementations.

Correct, the R default BLAS is quite stable and sort of battletested for the results (but not for speed, obviously).

There are known cases where some BLASes don't get the answers right (especially SVDs combined with ill-conditional matrix which chain matrix operations, where getting a number flip can go wrong all down the line). R default shipped BLAS also has some of those quirks but are very hard to find.

Some results below on HIGGS, which MKL probably won't improve well due to how small the number of columns is.

Linux + ICC + Eigen ran 11 times, took average run:

```r
    user   system  elapsed 
2513.352   1.991   71.483 

0.8363038
```

Linux + ICC + Intel MKL (on Dual Intel Xeon 6154) ran 11 times, took average run:

```r
    user   system  elapsed 
2339.294    3.775   65.341

0.8363038
```

Reproducible code below. Note that I'm running LightGBM with 36 threads and forcing them to spawn during `fread` (which, in my case, uses exactly 36 threads) in order to enforce first touch policy for OpenMP threads to maintain the highest memory locality possible (I have multiple NUMA nodes).
On UMA setups (no NUMA nodes), just rerun from speed everytime to get new measurements. 1st run will be slightly slower than subsequent runs, therefore discard first run.
On NUMA setups, clear the RAM cache with `sync; echo 3 > /proc/sys/vm/drop_caches` and rerun from scratch the script, as different NUMA setups might arise and lead to large differences (it shouldn't if you run the R with CPU affinity). Results might be overall slower than UMA since memory is fully wiped everytime, and no cache abuse can happen.

```r
library(lightgbm)
library(data.table)

setwd("Downloads/HIGGS.csv") # Where HIGGS.csv is : https://archive.ics.uci.edu/ml/machine-learning-databases/00280/
setDTthreads(36) # Make sure this is equal to the number of threads you are requesting, so if you are in NUMA you will have memory locality for max performance
data <- fread("HIGGS.csv") # Force spawn all the requested threads
labels <- data$V1
data[, V1 := NULL]
data <- as.matrix(data)

data_train <- data[1:8800000, ]
data_valid <- data[8800001:11000000, ]

labels_train <- labels[1:8800000]
labels_valid <- labels[8800001:11000000]

dtrain_lgb <- lgb.Dataset(data_train, label = labels_train)
dvalid_lgb <- lgb.Dataset.create.valid(dtrain_lgb, data_valid, label = labels_valid)
valids_lgb <- list(valid = dvalid_lgb)


auc <- function(preds, labels) {
  x1 = as.numeric(preds[labels == 1])
  n1 = as.numeric(length(x1))
  x2 = as.numeric(preds[labels == 0])
  n2 = as.numeric(length(x2))
  r = rank(c(x1,x2))
  return((sum(r[1:n1]) - n1 * (n1 + 1) / 2) / (n1 * n2))
}

# 36 CPU threads training
speed <- system.time({
  model <- lgb.train(
    params = list(max_depth = 6,
                  num_leaves = 2^8,
                  learning_rate = 0.1,
                  min_data_in_leaf = 1,
                  min_sum_hessian_in_leaf = 1,
                  lambda_l1 = 0,
                  lambda_l2 = 1,
                  min_gain_to_split = 0,
                  max_bin = 255,
                  force_row_wise = TRUE,
                  linear_tree = TRUE,
                  boosting = "gbdt",
                  objective = "binary",
                  metric = "binary_logloss",
                  device = "cpu"),
    data = dtrain_lgb,
    valids = valids_lgb,
    nrounds = 500,
    verbose = 1,
    num_thread = 36
  )
})
model_predict <- predict(model, data_valid)
auc(model_predict, labels_valid)
rm(model)
invisible(gc(verbose = FALSE))
speed
```

@david-cortes can you please check on HIGGS the speed with Eigen and with OpenBLAS using my example?

david-cortes:
Did a bit more testing. After measuring the execution time for only the second time that I run the code, I get roughly the same timings for eigen and openblas (186-190 seconds, on an 8c/16t ryzen 7 2700 using gcc11) and roughly the same AUROCs, whether using the default compilation options from R or something higher (O3/march=native).

However the first run on a fresh system restart seems to be more variable. Guess it was just a matter of luck seeing better timings for openblas, and for this specific use case eigen is well optimized.

jameslamb:
Ah ok, @david-cortes thank you so much for bringing this up and for the time and effort spent investigating it!!

And @Laurae2 for the excellent explanation in and examples in https://github.com/microsoft/LightGBM/pull/4692#issuecomment-973304576.

Issue #4657, opened 2021-10-08
ignore_column does not work when loading a dataset from a binary file


karunrao97:
## Description

I need to load select columns of a LightGBM dataset from a binary file. I am trying to use the ignore_column parameter for this, but this results in the following (where c1 is the first column name I'm trying to ignore):

> LightGBMError: Could not find ignore column c1 in data file

## Reproducible example

```python
import lightgbm as lgb
import numpy as np

n_samples = 100
feature_names = [f'c{i}' for i in range(10)]
ds_path = 'ds.bin'

ds = lgb.Dataset(np.random.random((n_samples, len(feature_names))), label=np.random.random(n_samples), feature_name=feature_names)
ds.construct()
ds.save_binary(ds_path)

ds = lgb.Dataset(ds_path)
ds.construct()  # this works fine

ds = lgb.Dataset(ds_path, params={'ignore_column': f'name:{",".join(feature_names[1:])}'})
ds.construct()  # this throws the above-mentioned exception
```

## Environment info

LightGBM version or commit hash: 5b7a6f3e7150aeb704d1dd2b852d246af3e913a3

Command(s) you used to install LightGBM: I built it from source

## Additional Comments

I also tried the following:

```python
ds = lgb.Dataset(ds_path, params={'ignore_column': ",".join([str(i) for i in range(1, len(feature_names))])})
ds.construct()
```

This doesn't crash, but it just loads all the features, and when training a model on this dataset, all the features are used, so it doesn't seem like ignore_column had any effect in this case.

jmoralez:
This happens with the CLI as well, both the error when passing `name:c1` and using them for training when passing integers like `0,1,2`.

PR #4701, opened 2021-10-21
Fix some paramater hints when loading from binary file

hzy46:
#4657 describes `ignore_column` will cause an error when loading from binary file. I think here're two related issues in the code:

1. `SetHeader` should only be called when loading from text file. Now, it is also called when loading the data from binary file. This is why error mentioned in issue 4657 happens.

https://github.com/microsoft/LightGBM/blob/13ed38ca55dc7d454712d794281c628f62301a0f/src/io/dataset_loader.cpp#L25

2. Apart from `ignore_column`, parameters `two_round`, `header`, `label_column`, `weight_column`, `group_column` also have no effect when loading from binary. I changed the doc. Need to confirm whether we should add warning for these parameters.

### DIFF
--- a/src/io/dataset_loader.cpp
+++ b/src/io/dataset_loader.cpp
@@ -22,10 +22,7 @@ DatasetLoader::DatasetLoader(const Config& io_config, const PredictFunction& pre
   label_idx_ = 0;
   weight_idx_ = NO_SPECIFIC;
   group_idx_ = NO_SPECIFIC;
-  if (filename != nullptr && CheckCanLoadFromBin(filename) == "") {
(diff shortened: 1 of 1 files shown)


StrikerRUS:
@hzy46 Thanks for your contribution! `Parameters.rst` shouldn't be touched directly. Instead, make your changes in the [`config.h` file](https://github.com/microsoft/LightGBM/blob/master/include/LightGBM/config.h) and then run `python helpers/parameter_generator.py` to update `Parameters.rst`.
Refer to https://github.com/microsoft/LightGBM/blame/master/docs/Parameters.rst#L1.

PR #4716, opened 2021-10-26
[docs] [dask] Add return information to Dask fit() docs (fixes #4402)

### LINKED ISSUE #4402: [python] Add `Returns` section to docstrings of `fit()` methods of Dask estimators
Refer to https://github.com/microsoft/LightGBM/pull/4101#issuecomment-866375793.

> Also, during rendering and checking my current suggestions I noticed that there are no Returns sections for fit() methods of Dask estimators. Is it OK?

> I guess they should have a return block similar to those in the equivalent scikit-learn estimators ...

jameslamb:
Fixes #4402, adding `Returns` blocks to the documentation for `fit()` method on estimators in `lightgbm.dask`.

### How I tested this

```shell
cd docs
make clean html
open _build/html/index.html
```

Looks like the docs are rendering correctly.

image

I also clicked through the links next to `Return type` and confirmed that they link back to the constructor docs for the relevant class.


jameslamb:
Thanks for the reviews! I just updated this to latest `master`. If it passes, I'll merge this.

@jmoralez I agree with your suggestion to update the same information from the sklearn docstrings. I'll propose those changes in a separate PR.

StrikerRUS:
What do you think about locating `Returns` section **before** a note about custom function, like it's done in sklearn?

#### sklearn

image

#### dask

image

jameslamb:
> What do you think about locating Returns section before a note about custom function, like it's done in sklearn?

Sure, works for me!

StrikerRUS (approved):
Thank you!

Issue #3679, opened 2020-12-25
Avoiding Exception "Check failed: (best_split_info.right_count) > (0) at ..." with a regression task


ch3rn0v:
## How you are using LightGBM?
* Python package

## Environment info

* Operating System: Ubuntu 20.04.1 LTS
* Python version: 3.8.5
* GCC 7.3.0
* LightGBM version or commit hash: 3.1.1

## Steps to reproduce

1. In jupyter lab's notebook, prepare train and validation datasets. (They are huge and private, so can't share a reproducible example).
2. Train lgbm with the data with different sets of features.
3. Observe an exception looking like this:

`Check failed: (best_split_info.right_count) > (0) at [...]`
Sometimes it says `left_count` instead of `right_count`.
Other times it doesn't occur at all, depending on the features I use.

## Other details

Apparently this is the start of the piece of code initiating the exception: https://github.com/microsoft/LightGBM/blob/master/src/treelearner/serial_tree_learner.cpp#L652.

I tried setting `min_data_in_leaf` to a value greater than zero. It helps sometimes, but not reliably. Same with `feature_fraction`. I also tried changing `min_sum_hessian_in_leaf`, to no avail. Also tried setting `min_data_in_leaf` and `min_sum_hessian_in_leaf` simultaneously, no difference.

This (or a similar) issue is mentioned a few times here:
- https://stackoverflow.com/questions/60161691/best-split-info-check-failure-encountered-while-fitting-lightgbm-classifier
- https://github.com/microsoft/LightGBM/issues/3603
- https://github.com/microsoft/LightGBM/issues/2742

None of them suggests an approach that allowed me to avoid these exceptions. Would you please share any ideas how to fix this, or at least why does this issue happen at all? If I understand correctly, one could simply trim the split leading to this error and stop branching further. Please correct me if I'm wrong. Thank you.

guolinke:
you can use larger `min_data_per_leaf` or `min_hessian_per_leaf`.

ch3rn0v:
>you can use larger `min_data_per_leaf` or `min_hessian_per_leaf`
Sure, I tried a few values. If I increase these too much though it results in the model being under-fitted.

I use `"regression"` as the value for the `"objective"` param. Metric is `"l1"`.

>you may use the sample weight?

Do you mean specifying different weights for different samples? If so, I do not use this in the example we are discussing here.

Thank you very much for the rapid reply!

guolinke:
[...]

PR #4725, opened 2021-10-27
Always respect forced splits, even when feature_fraction < 1.0 (fixes #4601)

### LINKED ISSUE #4601: Using feature_fraction and forced splits together can cause training failures
## Description

Using forced splits together with `feature_fraction` can lead to the following error

> [LightGBM] [Fatal] Check failed: (best_split_info.left_count) > (0) at treelearner/serial_tree_learner.cpp, line 653 .

I've found that this error:

* does not occur if using `feature_fraction=1.0` (the default)
* occurs randomly (even with a fixed dataset and setting `random_state` and `feature_fraction_seed`)
* tends to less often with more verbose logging (i.e. occurs less frequently with `verbose = -10` and least frequently with `verbose = 10`
* occurs even for datasets where the number of samples is well below `bin_construct_sample_cnt`

## Reproducible example

<details><summary>R example (click me)</summary>

The code below using the built-in `Chickweight` dataset can be used to reproduce this issue.

Trying different combinations of settings for parameters `VERBOSITY`, `FEATURE_FRACTION` and `USE_FORCED_SPLITS`, I found the following results.

*different settings of verbose*

```text
---- forced_splits=TRUE, feature_fraction=0.600000, verbose=10---
5 out of 10 succeeded
6 out of 10 succeeded
4 out of 10 succeeded
7 out of 10 succeeded
7 out of 10 succeeded

---- forced_splits=TRUE, feature_fraction=0.600000, verbose=1 ---
0 out of 10 succeeded
0 out of 10 succeeded
0 out of 10 succeeded
1 out of 10 succeeded
0 out of 10 succeeded

---- forced_splits=TRUE, feature_fraction=0.600000, verbose=0 ---
0 out of 10 succeeded
0 out of 10 succeeded
0 out of 10 succeeded
2 out of 10 succeeded
0 out of 10 succeeded

---- forced_splits=TRUE, feature_fraction=0.600000, verbose=-10 ---
0 out of 10 succeeded
1 out of 10 succeeded
0 out of 10 succeeded
0 out of 10 succeeded
0 out of 10 succeeded
```

*different settings of feature fraction (NOTE: this dataset only has 3 features)*

```text
---- forced_splits=TRUE, feature_fraction=0.400000, verbose=10 ---
0 out of 10 succeeded
0 out of 10 succeeded
0 out of 10 succeeded
0 out of 10 succeeded
0 out of 10 succeeded

---- forced_splits=TRUE, feature_fraction=0.600000, verbose=10 ---
8 out of 10 succeeded
6 out of 10 succeeded
6 out of 10 succeeded
4 out of 10 succeeded
6 out of 10 succeeded

---- forced_splits=TRUE, feature_fraction=0.800000, verbose=10 ---
5 out of 10 succeeded
9 out of 10 succeeded
8 out of 10 succeeded
8 out of 10 succeeded
8 out of 10 succeeded

---- forced_splits=TRUE, feature_fraction=0.800000, verbose=10 ---
10 out of 10 succeeded
10 out of 10 succeeded
10 out of 10 succeeded
10 out of 10 succeeded
10 out of 10 succeeded
```

```r
library(data.table)
library(lightgbm)

set.seed(321)

VERBOSITY <- 10L
FEATURE_FRACTION <- 0.6
USE_FORCED_SPLITS <- TRUE

data(ChickWeight)
trainDT <- data.table::as.data.table(ChickWeight)
target <- trainDT[["weight"]]
X <- data.matrix(trainDT[, .(Time, Chick, Diet)])

string = paste0('{
  "feature": 0,
  "threshold": 10.0,
  "right": {
    "feature": 0,
    "threshold": 16.0
  }
}')

write(string, file = "forced_split.json")

string = paste0('[
    {
        "feature": 0,
        "bin_upper_bound": [ 10.0, 16.0]
    }
]')

write(string, file = "forcedbins.json")

train_ds <- lgb.Dataset(
    data = X
    , label = target
    , params = list(
        forcedbins_filename = "forcedbins.json"
        , min_data_in_leaf = 1
    )
)

params <- list(
    objective = "regression"
    , feature_fraction = FEATURE_FRACTION
    , num_iterations = 10
    , num_leaves = 8
    , random_state = 708
    , feature_fraction_seed = 708
)

if (USE_FORCED_SPLITS) {
    params[["forcedsplits_filename"]] <- "forced_split.json"
}

# fails randomly, even setting random_state and feature_fraction_seed
results <- sapply(
    X = seq_len(10L)
    , FUN = function(...) {
        try({
            fit <- lgb.train(
                params = params
                , data = train_ds
                , verbose = VERBOSITY
            )
            return(1)
        })
        return(0)
    }
)
num_successes <- sum(results)
num_attempts <- length(results)
print("")
print(sprintf(
    "---- forced_splits=%s, feature_fraction=%f, verbose=%i---"
    , USE_FORCED_SPLITS
    , FEATURE_FRACTION
    , VERBOSITY
))
print(paste0(num_successes, " out of ", num_attempts, " succeeded"))
```

*whether or not feature_fraction is used (using settings of other params which led to the most errors)*

```text
---- forced_splits=TRUE, feature_fraction=0.400000, verbose=-1---
0 out of 10 succeeded
0 out of 10 succeeded
0 out of 10 succeeded
0 out of 10 succeeded
0 out of 10 succeeded

---- forced_splits=FALSE, feature_fraction=0.400000, verbose=-1---
10 out of 10 succeeded
10 out of 10 succeeded
10 out of 10 succeeded
10 out of 10 succeeded
10 out of 10 succeeded
```

</details>

I haven't been able to reproduce this issue in the Python package yet, although I'm not sure if it's R-specific.

<details><summary>Python example code (click me)</summary>

My attempt to replicate the R example in Python. For a variety of settings of `VERBOSITY` and `FEATURE_FRACTION`, I have not been able to replicate the errors observed in R.

Maybe there is some dataset-specific characteristic that's significant. One way to try to make this an even tighter comparison would be to write the `ChickWeight` data from the R package out to a CSV file and read that data into Python.

```python
import json
import lightgbm as lgb
import pandas as pd
import numpy as np

VERBOSITY = -10
FEATURE_FRACTION = 0.2
USE_FORCED_SPLITS = True

X, target = make_regression(
    n_samples=1_000,
    n_features=3,
    n_informative=1,
    random_state=321
)

threshold1 = np.percentile(X[:, 0], 0.4)
threshold2 = np.percentile(X[:, 0], 0.9)

forced_split = {
    "feature": 0,
    "threshold": threshold1,
    "right": {
        "feature": 0,
        "threshold": threshold2
    }
}
with open("forced_split.json", "w") as f:
    f.write(json.dumps(forced_split))
    
forced_bins = [
    {
        "feature": 0,
        "bin_upper_bound": [threshold1, threshold2]
    }
]
with open("forcedbins.json", "w") as f:
    f.write(json.dumps(forced_bins))
    
train_ds = lgb.Dataset(
    data=X,
    label=target,
    params={
        "forcedbins_filename": "forcedbins.json",
        "min_data_in_leaf": 1
    }
)

params = {
    "objective": "regression",
    "feature_fraction": FEATURE_FRACTION,
    "num_iterations": 10,
    "num_leaves": 8,
    "random_state": 708,
    "feature_fraction_seed": 708,
    "verbose": VERBOSITY
}

if USE_FORCED_SPLITS:
    params["forcedsplits_filename"] = "forced_split.json"
    
successes = 0
attempts = 0
for i in range(10):
    attempts += 1
    try:
        bst = lgb.train(
            params=params,
            train_set=train_ds,
        )
        successes += 1
    except Exception as err:
        print(err)

print(f"---- forced_splits={USE_FORCED_SPLITS}, feature_fraction={round(FEATURE_FRACTION, 6)}, verbose={VERBOSITY} ---")
print(f"{successes} out of {attempts} succeeded")
```

</details>

## Environment info

LightGBM version or commit hash: latest `master` (https://github.com/microsoft/LightGBM/commit/64f15005040aafa528bbec7f873364d94b7e36a4)

```shell
# R
sh build-cran-package.sh
R CMD INSTALL lightgbm_3.2.1.9.tar.gz

# python
cd python-package
python setup.py install
```

<details><summary>output of R sessionInfo() (click me)</summary>

```text
R version 4.1.0 (2021-05-18)
Platform: x86_64-apple-darwin17.0 (64-bit)
Running under: macOS Mojave 10.14.6

Matrix products: default
BLAS:   /Library/Frameworks/R.framework/Versions/4.1/Resources/lib/libRblas.dylib
LAPACK: /Library/Frameworks/R.framework/Versions/4.1/Resources/lib/libRlapack.dylib

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base

[1] compiler_4.1.0
```

</details>

<details><summary>output of 'conda info' (click me)</summary>

```text
     active environment : None
       user config file : /Users/jlamb/.condarc
 populated config files : /Users/jlamb/.condarc
          conda version : 4.10.3
    conda-build version : not installed
         python version : 3.8.8.final.0
       virtual packages : __osx=10.14.6=0
                          __unix=0=0
                          __archspec=1=x86_64
       base environment : /Users/jlamb/miniconda3  (writable)
      conda av data dir : /Users/jlamb/miniconda3/etc/conda
  conda av metadata url : None
           channel URLs : https://conda.anaconda.org/conda-forge/osx-64
                          https://conda.anaconda.org/conda-forge/noarch
                          https://repo.anaconda.com/pkgs/main/osx-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/r/osx-64
                          https://repo.anaconda.com/pkgs/r/noarch
          package cache : /Users/jlamb/miniconda3/pkgs
                          /Users/jlamb/.conda/pkgs
       envs directories : /Users/jlamb/miniconda3/envs
                          /Users/jlamb/.conda/envs
               platform : osx-64
             user-agent : conda/4.10.3 requests/2.25.1 CPython/3.8.8 Darwin/18.7.0 OSX/10.14.6
                UID:GID : 501:20
             netrc file : None
           offline mode : False
```

</details>

## Additional Comments

Created based on the example code provided in https://github.com/microsoft/LightGBM/issues/4591#issuecomment-914688417.

The fact that more verbose logging leads to this error occurring less often is very surprising to me. The only explanations I can think of for this are:

a. there is a race condition somewhere in LightGBM and adding more logging slows LightGBM down enough to reduce the risk of hitting it
b. something else about the logging code causes a sync-up in otherwise asynchronous code

It would probably be useful to investigate this race condition idea further by running the reproducible examples above under different values of `num_threads`.

tongwu-msft:
Fixes #4601.

Root cause for this issue: The histogram for the force split feature might not construct if feature_fraction < 1.0. And it would failed at SplitInner.

- Only col-wise histogram construction impacted.  row-wise histogram construction would go through all features.
- Only force split settings gain > 0 impacted. If no gain for force_split leaf, the force_split settings would be auto ignored.

### DIFF
--- a/src/treelearner/serial_tree_learner.cpp
+++ b/src/treelearner/serial_tree_learner.cpp
@@ -323,22 +323,29 @@ bool SerialTreeLearner::BeforeFindBestSplit(const Tree* tree, int left_leaf, int
 
 void SerialTreeLearner::FindBestSplits(const Tree* tree) {
   std::vector<int8_t> is_feature_used(num_features_, 0);
-  #pragma omp parallel for schedule(static, 256) if (num_features_ >= 512)
+#pragma omp parallel for schedule(static, 256) if (num_features_ >= 512)
   for (int feature_index = 0; feature_index < num_features_; ++feature_index) {
     if (!col_sampler_.is_feature_used_bytree()[feature_index]) continue;
     if (parent_leaf_histogram_array_ != nullptr
-        && !parent_leaf_histogram_array_[feature_index].is_splittable()) {
+      && !parent_leaf_histogram_array_[feature_index].is_splittable()) {
       smaller_leaf_histogram_array_[feature_index].set_is_splittable(false);
       continue;
     }
     is_feature_used[feature_index] = 1;
   }
+
+  FindBestSplits(tree, is_feature_used);
+}
+
+void SerialTreeLearner::FindBestSplits(const Tree* tree, const std::vector<int8_t>& is_feature_used)
+{
   bool use_subtract = parent_leaf_histogram_array_ != nullptr;
 
 #ifdef USE_CUDA
   if (LGBM_config_::current_learner == use_cpu_learner) {
     SerialTreeLearner::ConstructHistograms(is_feature_used, use_subtract);
-  } else {
+  }
+  else {
     ConstructHistograms(is_feature_used, use_subtract);
   }
 #else
@@ -463,11 +470,8 @@ int32_t SerialTreeLearner::ForceSplits(Tree* tree, int* left_leaf,
   std::unordered_map<int, SplitInfo> forceSplitMap;
   q.push(std::make_pair(left, *left_leaf));
   while (!q.empty()) {
-    // before processing next node from queue, store info for current left/right leaf
-    // store "best split" for left and right, even if they might be overwritten by forced split
-    if (BeforeFindBestSplit(tree, *left_leaf, *right_leaf)) {
-      FindBestSplits(tree);
-    }
+    FindBestSplitsForForceSplitLeaf(tree, left_leaf, right_leaf, left, right);
+
     // then, compute own splits
     SplitInfo left_split;
     SplitInfo right_split;
@@ -559,6 +563,31 @@ int32_t SerialTreeLearner::ForceSplits(Tree* tree, int* left_leaf,
     ++result_count;
   }
   return result_count;
+}
+
+void SerialTreeLearner::FindBestSplitsForForceSplitLeaf(Tree* tree, int* left_leaf, int* right_leaf, Json left_force_split_leaf_setting, Json right_force_split_leaf_setting)
+{
+  // before processing next node from queue, store info for current left/right leaf
+  // store "best split" for left and right, even if they might be overwritten by forced split
+  if (BeforeFindBestSplit(tree, *left_leaf, *right_leaf)) {
+    std::vector<int8_t> is_feature_used(num_features_, 0);
+
+    if (!left_force_split_leaf_setting.is_null()) {
+      const int left_feature = left_force_split_leaf_setting["feature"].int_value();
+      const int left_inner_feature_index = train_data_->InnerFeatureIndex(left_feature);
+
+      is_feature_used[left_inner_feature_index] = 1;
+    }
+
+    if (!right_force_split_leaf_setting.is_null()) {
+      const int right_feature = right_force_split_leaf_setting["feature"].int_value();
+      const int right_inner_feature_index = train_data_->InnerFeatureIndex(right_feature);
+
+      is_feature_used[right_inner_feature_index] = 1;
+    }
+
+    FindBestSplits(tree, is_feature_used);
+  }
 }
 
 void SerialTreeLearner::SplitInner(Tree* tree, int best_leaf, int* left_leaf,
--- a/src/treelearner/serial_tree_learner.cpp
+++ b/src/treelearner/serial_tree_learner.cpp
@@ -322,10 +322,15 @@ bool SerialTreeLearner::BeforeFindBestSplit(const Tree* tree, int left_leaf, int
 }
 
 void SerialTreeLearner::FindBestSplits(const Tree* tree) {
+  FindBestSplits(tree, nullptr);
+}
+
+void SerialTreeLearner::FindBestSplits(const Tree* tree, const std::set<int>* force_features)
+{
   std::vector<int8_t> is_feature_used(num_features_, 0);
 #pragma omp parallel for schedule(static, 256) if (num_features_ >= 512)
   for (int feature_index = 0; feature_index < num_features_; ++feature_index) {
-    if (!col_sampler_.is_feature_used_bytree()[feature_index]) continue;
+    if (!col_sampler_.is_feature_used_bytree()[feature_index] && (force_features == nullptr || force_features->find(feature_index) == force_features->end())) continue;
     if (parent_leaf_histogram_array_ != nullptr
       && !parent_leaf_histogram_array_[feature_index].is_splittable()) {
       smaller_leaf_histogram_array_[feature_index].set_is_splittable(false);
@@ -334,11 +339,6 @@ void SerialTreeLearner::FindBestSplits(const Tree* tree) {
     is_feature_used[feature_index] = 1;
   }
 
-  FindBestSplits(tree, is_feature_used);
-}
-
-void SerialTreeLearner::FindBestSplits(const Tree* tree, const std::vector<int8_t>& is_feature_used)
-{
   bool use_subtract = parent_leaf_histogram_array_ != nullptr;
 
 #ifdef USE_CUDA
@@ -570,23 +570,21 @@ void SerialTreeLearner::FindBestSplitsForForceSplitLeaf(Tree* tree, int* left_le
   // before processing next node from queue, store info for current left/right leaf
   // store "best split" for left and right, even if they might be overwritten by forced split
   if (BeforeFindBestSplit(tree, *left_leaf, *right_leaf)) {
-    std::vector<int8_t> is_feature_used(num_features_, 0);
+    std::set<int> force_features;
 
     if (!left_force_split_leaf_setting.is_null()) {
       const int left_feature = left_force_split_leaf_setting["feature"].int_value();
       const int left_inner_feature_index = train_data_->InnerFeatureIndex(left_feature);
-
-      is_feature_used[left_inner_feature_index] = 1;
+      force_features.insert(left_inner_feature_index);
     }
 
     if (!right_force_split_leaf_setting.is_null()) {
       const int right_feature = right_force_split_leaf_setting["feature"].int_value();
       const int right_inner_feature_index = train_data_->InnerFeatureIndex(right_feature);
-
-      is_feature_used[right_inner_feature_index] = 1;
+      force_features.insert(right_inner_feature_index);
     }
 
-    FindBestSplits(tree, is_feature_used);
+    FindBestSplits(tree, &force_features);
   }
 }
 
--- a/src/treelearner/serial_tree_learner.cpp
+++ b/src/treelearner/serial_tree_learner.cpp
@@ -468,7 +468,7 @@ int32_t SerialTreeLearner::ForceSplits(Tree* tree, int* left_leaf,
   std::unordered_map<int, SplitInfo> forceSplitMap;
   q.push(std::make_pair(left, *left_leaf));
 
-  // Histogram construction require parent features. 
+  // Histogram construction require parent features.
   std::set<int> force_split_features = FindAllForceFeatures(*forced_split_json_);
   while (!q.empty()) {
     if (BeforeFindBestSplit(tree, *left_leaf, *right_leaf)) {
--- a/src/treelearner/serial_tree_learner.cpp
+++ b/src/treelearner/serial_tree_learner.cpp
@@ -467,8 +467,13 @@ int32_t SerialTreeLearner::ForceSplits(Tree* tree, int* left_leaf,
   bool left_smaller = true;
   std::unordered_map<int, SplitInfo> forceSplitMap;
   q.push(std::make_pair(left, *left_leaf));
+
+  // Histogram construction require parent features. 
+  std::set<int> force_split_features = FindAllForceFeatures(*forced_split_json_);
   while (!q.empty()) {
-    FindBestSplitsForForceSplitLeaf(tree, left_leaf, right_leaf, left, right);
+    if (BeforeFindBestSplit(tree, *left_leaf, *right_leaf)) {
+      FindBestSplits(tree, &force_split_features);
+    }
 
     // then, compute own splits
     SplitInfo left_split;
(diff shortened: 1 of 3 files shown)


tests/python_package_test/test_engine.py
```diff
@@ -2887,3 +2888,33 @@ def hook(obj):
...
+    params = {
+        "objective": "regression",
+        "feature_fraction": 0.6,
+        "force_col_wise": True,
+        "feature_fraction_seed": 1,
+        "forcedsplits_filename": tmp_split_file
+    }
+
+    gbm = lgb.train(params, lgb_train)
+    treesDf = gbm.trees_to_dataframe()
```
StrikerRUS, line 2916: `trees_to_dataframe()` method requires that the `pandas` package should be installed. And `pandas` isn't a hard dependency of the LightGBM Python-package. So, this test should be skipped in case of `pandas` isn't installed.
https://github.com/microsoft/LightGBM/blob/d130bb198b2b9520f0eb6d2a2be3ca1b3bd22c42/tests/python_package_test/test_engine.py#L2505
https://github.com/microsoft/LightGBM/blob/d130bb198b2b9520f0eb6d2a2be3ca1b3bd22c42/python-package/setup.py#L338-L343
#4638

Given that the purpose of this test is to ensure correct behavior of core LightGBM functionality with `feature_fraction` and `forcedsplits_filename` parameters, and `trees_to_dataframe()` method is used only to ease the check of produced results, I suggest to drop the usage of `trees_to_dataframe()` and run this important test unconditionally to the `pandas` installation. This can be achieved by parsing the output of `dump_model()` method directly.

StrikerRUS (changes requested):

tests/python_package_test/test_engine.py
tongwu-msft: Thanks, updated to dump_model()

src/treelearner/serial_tree_learner.cpp
```diff
@@ -322,17 +323,22 @@ bool SerialTreeLearner::BeforeFindBestSplit(const Tree* tree, int left_leaf, int
 }
 
 void SerialTreeLearner::FindBestSplits(const Tree* tree) {
+  FindBestSplits(tree, nullptr);
+}
+
+void SerialTreeLearner::FindBestSplits(const Tree* tree, const std::set<int>* force_features) {
   std::vector<int8_t> is_feature_used(num_features_, 0);
-  #pragma omp parallel for schedule(static, 256) if (num_features_ >= 512)
+#pragma omp parallel for schedule(static, 256) if (num_features_ >= 512)
```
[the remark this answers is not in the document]
jameslamb: Thanks, but I still see several white-space only changes at https://github.com/microsoft/LightGBM/pull/4725/files.

For example:

image

Maybe you have some settings in your local editor that is automatically making changes on save?

tests/python_package_test/test_engine.py
```diff
@@ -2887,3 +2888,30 @@ def hook(obj):
...
+
+    params = {
+        "objective": "regression",
+        "feature_fraction": 0.6,
+        "force_col_wise": True,
+        "feature_fraction_seed": 1,
+        "forcedsplits_filename": "forced_split.json"
+    }
+
+    gbm = lgb.train(params, lgb_train)
```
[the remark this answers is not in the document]
jameslamb: oooo interesting!! I didn't consider that fact that the choice of col-wise vs. row-wise histogram construction could be a source of randomness for small datasets (because the timing differences are so small). Thanks very much for explaining that.

But this PR is intended to FIX #4601, regardless of whether or not `force_col_wise=True`, right? I think that's the ideal behavior, since the setting of `force_col_wise` should just be an optimization and not impact the correctness of results or risk of errors.

If it is, then I think the unit tests should be testing once with `force_col_wise=True` and once with `force_row_wise=True`.

tests/python_package_test/test_engine.py
```diff
@@ -2887,3 +2888,37 @@ def hook(obj):
     dumped_model_str = str(bst.dump_model(5, 0, object_hook=hook))
     assert "leaf_value" not in dumped_model_str
     assert "LV" in dumped_model_str
+
+
+def test_force_split_with_feature_fraction(tmp_path):
+    def check_force_splits(dumped_model):
```
jameslamb, line 2894: Since this is called only one time, inside this one test case, I don't think there's any added benefit to encapsulating this code in a function. To eliminate some indirection and make the test a bit easier for others to understand, can you please just include this code at the bottom of this test case?

```python
# ensure every tree used this split
tree_info = gbm.dump_model()["tree_info"]
assert len(tree_info) > 1
for tree in tree_info:
    tree_structure = tree["tree_structure"]
    assert tree_structure['split_feature'] == 0
```

jameslamb (changes requested):
Thanks for the changes. Please see a few suggestions I left about the tests. I think it's important to test with `force_col_wise=true` and `force_row_wise=true` to be confident that this fixes #4601.

tests/python_package_test/test_engine.py
tongwu-msft: Thanks, updated.

jameslamb: Thanks! Can you please also include the `assert len(tree_info) > 1` from my suggestion?

That's just a little extra defense in case `gbm.dump_model()["tree_info"]` is unexpectedly empty. If that was empty, none of the code under `for tree in tree_info` would be executed.

tests/python_package_test/test_engine.py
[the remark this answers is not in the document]
tongwu-msft: I find another very random issue related to force_row_wise=True.... Good news is now I can repro it with this test if I change force-row-wise to True with seed 1 (it is not related to feature fraction, and looks like also not related to force split...), I believe it caused by some issues in row wise histogram construction, but I still not able to find the root cause. Let me work on that separately and add tests for row-wise in that fix.

tests/python_package_test/test_engine.py
tongwu-msft: Thanks, updated.

tests/python_package_test/test_engine.py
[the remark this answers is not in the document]
tongwu-msft: It looks like the possible root cause of #4739 and #3679

src/treelearner/serial_tree_learner.cpp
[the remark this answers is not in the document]
tongwu-msft: Thanks, updated.

tests/python_package_test/test_engine.py
```diff
@@ -2887,3 +2888,36 @@ def hook(obj):
...
+
+    gbm = lgb.train(params, lgb_train)
+    ret = mean_absolute_error(y_test, gbm.predict(X_test))
+    assert ret < 2.0
+
+    tree_info = gbm.dump_model()["tree_info"]
+    assert len(tree_info) > 1
+    for tree in tree_info:
+        tree_structure = tree["tree_structure"]
+        assert tree_structure['split_feature'] == 0
```
StrikerRUS, line 2923:
```suggestion
        assert tree_structure['split_feature'] == 0
        assert tree_structure['threshold'] == pytest.approx(0.5, abs=1e-1)
```

StrikerRUS (approved):
LGTM, thank you very much for working on this!
I left only one suggestion for your consideration to make the test more strict.

tests/python_package_test/test_engine.py
tongwu-msft: the actual threshold is the nearest record is about 0.52..

shiyu1994:
> Thanks for the changes. Please see a few suggestions I left about the tests. I think it's important to test with `force_col_wise=true` and `force_row_wise=true` to be confident that this fixes #4601.

@jameslamb Hi James, we found that when `force_row_wise=true`, the test case fails due to other problem, which is not related to using forced splits with feature_fraction < 1.0. It is probably due to the same problem as in issues #3679 #4739.

Can we leave this in subsequent PRs to fix, and merge this for now, given that all tests are passed?

tongwu-msft:
> > Hi James, we found that when force_row_wise=true, the test case fails due to other problem
> 
> @shiyu1994 just to be sure I understand, which of these do you mean?
> 
> 1. the tests added in this PR fail if you run them on `master` with `force_row_wise=True`
> 2. the tests added in this PR fail on this PR branch (but not `master`) with `force_row_wise=True`
> 
> If it's number 1, then I support merging this PR as-is and fixing the other bug in a later PR.
> 
> If it's number 2, then that means that bug would be introduced by this PR, and I don't think it should be merg ed.

@jameslamb thanks for the check here. It is for #1, "force_row_wise=True" same test would also repro on current master branch.

jameslamb:
>  It is for #1

Got it, thanks! Then I think it's ok to merge this and fix that other bug separately. Thanks for explaining it to me 😄

jameslamb (approved):
Thanks for working through this!

Never considered, when I encountered #4601, that a source of randomness could be in the choice of row-wise vs. column-wise Dataset construction. So I learned something really important through reviewing this and through your explanations ❤️

PR #4732, opened 2021-10-28
[ci] indicate support of Monterey

StrikerRUS:
According to the policy, we indicate support only for the last 3 macOS versions in wheel name: https://github.com/microsoft/LightGBM/pull/2691#issue-549125519.

I believe that "the last 3 macOS  versions" means 3 last major releases like it was before 11.x 

Successfully built bottles for `arm64_monterey`, `arm64_big_sur`, `monterey`, `big_sur`, `catalina`, `mojave` by Homebrew team give some confidence in that LightGBM is indeed compatible with the wide range of macOS versions.
https://github.com/Homebrew/homebrew-core/blob/4060c0ab42d8b9b7168b34f55f12e16a91b27150/Formula/lightgbm.rb#L10-L15

Just FYI, it is now possible to provide one `universal2` wheel for both `x86_64` and `arm64`.
- https://discuss.python.org/t/apple-silicon-and-packaging/4516
- https://github.com/numpy/numpy/issues/18143

### DIFF
--- a/.ci/test.sh
+++ b/.ci/test.sh
@@ -133,7 +133,7 @@ if [[ $TASK == "sdist" ]]; then
 elif [[ $TASK == "bdist" ]]; then
     if [[ $OS_NAME == "macos" ]]; then
         cd $BUILD_DIRECTORY/python-package && python setup.py bdist_wheel --plat-name=macosx --python-tag py3 || exit -1
-        mv dist/lightgbm-$LGB_VER-py3-none-macosx.whl dist/lightgbm-$LGB_VER-py3-none-macosx_10_14_x86_64.macosx_10_15_x86_64.macosx_11_0_x86_64.whl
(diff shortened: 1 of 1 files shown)


jameslamb (approved):
Thanks for this. I agree with your interpretation of "last 3 major versions".

StrikerRUS:
> I agree with your interpretation of "last 3 major versions".

Just to clarify, right now due to the restrictions of our CI (https://github.com/microsoft/LightGBM/pull/4732#issuecomment-953987758) we are supporting "last 4 major versions", not 3.

We may consider migrating to a newer macOS builder like XGBoost recently did: https://github.com/dmlc/xgboost/pull/7382.
But it will require updating info about the Xcode version used to create release artifacts in two places
https://github.com/microsoft/LightGBM/tree/master/python-package#install-from-pypi
https://github.com/microsoft/LightGBM/blob/dcae8dd5696cb35f105680ecc1de15feede250a1/.ci/setup.sh#L7
because the minimum available version of Xcode at `macos-10.15` builder is `10.3`
https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md#xcode

And I don't know whether new artifacts will remain backward compatible.

jameslamb:
Got it, yep thanks for the further explanation. Makes sense to me.

PR #4637, opened 2021-09-30
add param aliases from scikit-learn

StrikerRUS:
With `1.0` release of scikit-learn `HistGradientBoostingClassifier` and `HistGradientBoostingRegressor` are no longer experimental.
https://scikit-learn.org/stable/whats_new/v1.0.html#sklearn-ensemble

Add names of their constructor arguments as aliases to LightGBM's params. This should help users play around with different estimators with minimal changes in their code.
https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.HistGradientBoostingRegressor.html

PR #4746, opened 2021-10-29
Add 'nrounds' as an alias for 'num_iterations' (fixes #4743)

### LINKED ISSUE #4743: [R-package] Add 'nrounds' as an alias for 'num_iterations'
## Description

In the R package, `lgb.train()` and `lightgbm()` expose a keyword argument `nrounds`, an integer indicating how many boosting rounds should be performed.

`{lightgbm}` uses the value of that argument to set `num_iteration`, unless `num_iterations` or another alias for it (https://lightgbm.readthedocs.io/en/latest/Parameters.html#num_iterations) is provided in the keyword argument `params`.

https://github.com/microsoft/LightGBM/blob/798dc1d4191b93fd34797d62b79c66cd95209406/R-package/R/lgb.train.R#L111-L115

Since `nrounds` is not a supported alias for `num_iterations` in LightGBM, it is not possible to override the default value of `nrounds` in `lgb.train()` or `lightgbm()` by passing `nrounds` as part of the list in `params`.

As @mikemahoney218 noted in https://github.com/microsoft/LightGBM/issues/4226#issuecomment-952269650, this is confusing behavior. Everywhere else in the R and Python packages, LightGBM treats values in `params` as higher-precedence than those passed through keyword arguments.

This behavior also adds friction to hyperparameter tuning (and will add even more once the suggestions from #4226 are fully implemented), as it makes `nrounds` a training parameter that cannot be altered by altering the list in `params`.

`nrounds` should be added as an alias for `num_iterations` in LightGBM.

## Reproducible example

```r
library(lightgbm)

    agaricus.train$data
)

bst <- lightgbm::lgb.train(
    params = list(
        "nrounds" = 17
        , "objective" = "regression"
    )
    , data = dtrain
)

# should be 17, but 100 boosting rounds were performed
bst$current_iter()

# [1] 100
```

## How to fix this

Add `nrounds` to the list of aliases for `num_iterations` at https://github.com/microsoft/LightGBM/blob/798dc1d4191b93fd34797d62b79c66cd95209406/R-package/R/aliases.R#L108-L119.

To ensure that other interfaces to LightGBM besides the R package respect this parameter alias, update the relevant C++ code and documentation. See https://github.com/microsoft/LightGBM/pull/4637/files for reference of which files should be changed. Do not edit `docs/Parameters.rst` directly... run `python helpers/parameter_generator.py` from the root of the repo after updating files in `include/` and `src/`.

Add a unit test to https://github.com/microsoft/LightGBM/blob/798dc1d4191b93fd34797d62b79c66cd95209406/R-package/tests/testthat/test_basic.R which confirms that `{lightgbm}` respects `nrounds` passed through `parameters`.

## Additional Comments

@StrikerRUS @Laurae2 please let me know if you disagree with this idea or have any additional thoughts to add.

mikemahoney218:
This PR addresses #4743, adding `nrounds` as an alias for `num_iterations` so that (within the R package) the top-level `nrounds` parameter may also be passed to the `params` argument of `lightgbm` and `lgb.train`. 

I followed the template of #4637 and the new tests work on my machine (:tm:). Please let me know if I missed anything!

### DIFF
--- a/R-package/tests/testthat/test_basic.R
+++ b/R-package/tests/testthat/test_basic.R
@@ -230,16 +230,23 @@ test_that("lightgbm() accepts nrounds as either a top-level argument or paramete
     )
     , save_name = tempfile(fileext = ".model")
   )
+  
+  top_level_l2 <- top_level_bst$eval_train()[[1L]][["value"]]
+  param_l2    <- param_bst$eval_train()[[1L]][["value"]]
+  both_l2      <- both_customized$eval_train()[[1L]][["value"]]
 
-  expect_equal(param_bst$current_iter(), top_level_bst$current_iter())
-  expect_equal(param_bst$best_score
-               , top_level_bst$best_score
-               , tolerance = TOLERANCE)
+  # check type just to be sure the subsetting didn't return a NULL
+  expect_true(is.numeric(top_level_l2))
+  expect_true(is.numeric(param_l2))
+  expect_true(is.numeric(both_l2))
+  
+  # check that model produces identical performance
+  expect_identical(top_level_l2, param_l2)
+  expect_identical(both_l2, param_l2)
 
-  expect_equal(param_bst$current_iter(), both_customized$current_iter())
-  expect_equal(param_bst$best_score
-               , both_customized$best_score
-               , tolerance = TOLERANCE)
+  expect_identical(param_bst$current_iter(), top_level_bst$current_iter())
+  expect_identical(param_bst$current_iter(), both_customized$current_iter())
+
 })
 
 test_that("lightgbm() performs evaluation on validation sets if they are provided", {
@@ -520,6 +527,75 @@ test_that("lgb.train() rejects negative or 0 value passed to nrounds", {
     }, "nrounds should be greater than zero")
   }
 })
+
+
+test_that("lgb.train() accepts nrounds as either a top-level argument or parameter", {
+  nrounds <- 15L
+  
+  set.seed(708L)
+  top_level_bst <- lgb.train(
+    data = lgb.Dataset(
+      train$data
+      , label = train$label
+    )
+    , nrounds = nrounds
+    , params = list(
+      objective = "regression"
+      , metric = "l2"
+      , num_leaves = 5L
+      , save_name = tempfile(fileext = ".model")
+    )
+  )
+  
+  set.seed(708L)
+  param_bst <- lgb.train(
+    data = lgb.Dataset(
+      train$data
+      , label = train$label
+    )
+    , params = list(
+      objective = "regression"
+      , metric = "l2"
+      , num_leaves = 5L
+      , nrounds = nrounds
+      , save_name = tempfile(fileext = ".model")
+    )
+  )
+  
+  set.seed(708L)
+  both_customized <- lgb.train(
+    data = lgb.Dataset(
+      train$data
+      , label = train$label
+    )
+    , nrounds = 20L
+    , params = list(
+      objective = "regression"
+      , metric = "l2"
+      , num_leaves = 5L
+      , nrounds = nrounds
+      , save_name = tempfile(fileext = ".model")
+    )
+  )
+  
+  top_level_l2 <- top_level_bst$eval_train()[[1L]][["value"]]
+  params_l2    <- param_bst$eval_train()[[1L]][["value"]]
+  both_l2      <- both_customized$eval_train()[[1L]][["value"]]
+  
+  # check type just to be sure the subsetting didn't return a NULL
+  expect_true(is.numeric(top_level_l2))
+  expect_true(is.numeric(params_l2))
+  expect_true(is.numeric(both_l2))
+  
+  # check that model produces identical performance
+  expect_identical(top_level_l2, params_l2)
+  expect_identical(both_l2, params_l2)
+  
+  expect_identical(param_bst$current_iter(), top_level_bst$current_iter())
+  expect_identical(param_bst$current_iter(), both_customized$current_iter())
+  
+})
+
 
 test_that("lgb.train() throws an informative error if 'data' is not an lgb.Dataset", {
   bad_values <- list(
--- a/R-package/tests/testthat/test_basic.R
+++ b/R-package/tests/testthat/test_basic.R
@@ -188,6 +188,60 @@ test_that("lightgbm() rejects negative or 0 value passed to nrounds", {
   }
 })
 
+test_that("lightgbm() accepts nrounds as either a top-level argument or parameter", {
+  nrounds <- 15L
+
+  set.seed(708L)
+  top_level_bst <- lightgbm(
+    data = train$data
+    , label = train$label
+    , nrounds = nrounds
+    , params = list(
+      objective = "regression"
+      , metric = "l2"
+      , num_leaves = 5L
+    )
+    , save_name = tempfile(fileext = ".model")
+  )
+
+  set.seed(708L)
+  param_bst <- lightgbm(
+    data = train$data
+    , label = train$label
+    , params = list(
+      objective = "regression"
+      , metric = "l2"
+      , num_leaves = 5L
+      , nrounds = nrounds
+    )
+    , save_name = tempfile(fileext = ".model")
+  )
+
+  set.seed(708L)
+  both_customized <- lightgbm(
+    data = train$data
+    , label = train$label
+    , nrounds = 20L
+    , params = list(
+      objective = "regression"
+      , metric = "l2"
+      , num_leaves = 5L
+      , nrounds = nrounds
+    )
+    , save_name = tempfile(fileext = ".model")
+  )
+
+  expect_equal(param_bst$current_iter(), top_level_bst$current_iter())
+  expect_equal(param_bst$best_score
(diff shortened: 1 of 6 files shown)


StrikerRUS:
@mikemahoney218 Thanks a lot for your contribution! Sorry for the inconvenience, but could you please sync this PR with the latest `master` branch to fix CI issues?

mikemahoney218:
@StrikerRUS No problem, should be up-to-date now.

R-package/tests/testthat/test_basic.R
```diff
@@ -188,6 +188,60 @@ test_that("lightgbm() rejects negative or 0 value passed to nrounds", {
...
+
+  expect_equal(param_bst$current_iter(), top_level_bst$current_iter())
+  expect_equal(param_bst$best_score
+               , top_level_bst$best_score
+               , tolerance = TOLERANCE)
+
+  expect_equal(param_bst$current_iter(), both_customized$current_iter())
+  expect_equal(param_bst$best_score
+               , both_customized$best_score
+               , tolerance = TOLERANCE)
```
StrikerRUS, line 242: I believe we should check here for the exact equality without any `tolerance`. Given the fixed environment and the same set of parameters, consecutive calls of lightgbm should produce bit-by-bit identical results. Just like here:
https://github.com/microsoft/LightGBM/blob/d62378b45314c2688deef7333ae16fcc9863d881/tests/python_package_test/test_basic.py#L66-L67

StrikerRUS (approved):
Thank you for adding new alias!
Generally LGTM, except one comment about asserting results in the test. But I'll defer the final decision to @jameslamb.

R-package/tests/testthat/test_basic.R
```diff
@@ -188,6 +188,60 @@ test_that("lightgbm() rejects negative or 0 value passed to nrounds", {
...
+      , num_leaves = 5L
+      , nrounds = nrounds
+    )
+    , save_name = tempfile(fileext = ".model")
+  )
+
+  expect_equal(param_bst$current_iter(), top_level_bst$current_iter())
+  expect_equal(param_bst$best_score
+               , top_level_bst$best_score
+               , tolerance = TOLERANCE)
```
jameslamb, line 237: I don't think `best_score` is the right thing to use here. For any LightGBM training where validation data isn't provided, this will always be `NA`.

```r
library(lightgbm)

train <- agaricus.train

nrounds <- 15L

top_level_bst <- lightgbm(
    data = train$data
    , label = train$label
    , nrounds = nrounds
    , params = list(
      objective = "regression"
      , metric = "l2"
      , num_leaves = 5L
    )
    , save_name = tempfile(fileext = ".model")
)

top_level_bst$best_score
# [1] NA
```

To test "the model produced is identical", to be more confident that the value of `num_iterations` was passed through, you can use `$eval_train()`. This creates predictions on the training data, using the full model, and then provides evaluation metrics based on those predictions.

```r
top_level_l2 <- top_level_bst$eval_train()[[1L]][["value"]]
params_l2    <- params_bst$eval_train()[[1L]][["value"]]

# check type just to be sure the subsetting didn't return a NULL
expect_true(is.numeric(top_level_2))
expect_true(is.numeric(params_l2))

# check that model produces identical performance
expect_equal(top_level_l2, params_l2)
```

Could you make this change?

R-package/tests/testthat/test_basic.R
jameslamb: I agree with @StrikerRUS . And just to be clear, that would mean using `testthat::expect_identical()`, not just removing `tolerance`.

Since `expect_equal()` already allows for small differences by default: https://rdrr.io/cran/testthat/man/equality-expectations.html

R-package/tests/testthat/test_basic.R
```diff
@@ -188,6 +188,60 @@ test_that("lightgbm() rejects negative or 0 value passed to nrounds", {
   }
 })
 
+test_that("lightgbm() accepts nrounds as either a top-level argument or parameter", {
```
jameslamb, line 191: can you please add an identical test (totally fine to copy and paste the code) that this also works with `lightgbm::lgb.train()`?

This package provides both interfaces that do the same thing, mainly for backwards-compatibility reasons, so we try to test that both are behaving as expected.

jameslamb (changes requested):
Thanks so much for picking this up! Changes look good to me, but I left a few requests on the additional unit tests.

R-package/tests/testthat/test_basic.R
mikemahoney218: I swapped this to `expect_identical`. When running interactively, this test was non-deterministic; it would sometimes fail with message ``top_level_l2` not identical to `params_l2`. Objects equal but not identical.` and sometimes succeed. My attempts to capture this using reprex succeeded every time, so I'm not sure if something about the interactive environment, R Studio, or (most likely) the way I was running tests was causing the issue, but I figured I'd flag all the same.

R-package/tests/testthat/test_basic.R
```diff
@@ -188,6 +188,67 @@ test_that("lightgbm() rejects negative or 0 value passed to nrounds", {
...
+  # check type just to be sure the subsetting didn't return a NULL
+  expect_true(is.numeric(top_level_l2))
+  expect_true(is.numeric(param_l2))
+  expect_true(is.numeric(both_l2))
+  
+  # check that model produces identical performance
+  expect_identical(top_level_l2, param_l2)
+  expect_identical(both_l2, param_l2)
+
+  expect_identical(param_bst$current_iter(), top_level_bst$current_iter())
```
jameslamb, line 247: sorry for missing this in my last review! Could you also add an assertion here somewhere comparing one of these `$current_iter()` values to `nrounds` directly?

That way, bugs in `$current_iter()` won't be able to silently pass this test.

For example, if `$current_iter()` always returned `100`, all these assertions would pass, even though you used `nrounds = 20`.

jameslamb (changes requested):

mikemahoney218:
I believe I've added commits that addressed all your comments, namely:

+ Use `eval_train()` rather than `best_score()`
+ Testing `lgb.train()` as well as `lightgbm()`
+ Explicit tests comparing `current_iter()` to `nrounds`
+ `expect_identical()` rather than `expect_equal()`

I want to mention that `expect_identical()` often failed when running these tests interactively; I don't know if it was something in how I ran the tests or in how my R Studio is configured that caused this issue. My attempts to capture this via `reprex::reprex()` always succeeded, so I'm expecting they'll work in a CI environment, but I wanted to flag this anyway.

I'm logging off for the weekend around now, so if you all have any other changes I'll look into them come Monday. Have a good weekend :smile:

jameslamb:
Thanks very much for the help! We may leave some additional comments here but no need to respond to them over the weekend. This is not urgent, and we really appreciate you taking the time so far to help out!

I'll try running the tests you've added on my machine and see if I also see nondeterministic behavior, thanks very much for flagging that.

StrikerRUS:
Kindly ping @mikemahoney218

mikemahoney218:
@StrikerRUS I believe that I'm waiting for you all to review the new changes :smile: At any rate, I'm not aware of any changes that need to be made on my end right now.

StrikerRUS:
@mikemahoney218 Ah, sorry, missed your the most recent commits! Then please fix the following linting errors:
```
[1] Total linting issues found: 10
/home/runner/work/LightGBM/LightGBM/R-package/tests/testthat/test_basic.R:233:1: style: Trailing whitespace is superfluous.
/home/runner/work/LightGBM/LightGBM/R-package/tests/testthat/test_basic.R:242:1: style: Trailing whitespace is superfluous.
/home/runner/work/LightGBM/LightGBM/R-package/tests/testthat/test_basic.R:535:1: style: Trailing whitespace is superfluous.
/home/runner/work/LightGBM/LightGBM/R-package/tests/testthat/test_basic.R:550:1: style: Trailing whitespace is superfluous.
/home/runner/work/LightGBM/LightGBM/R-package/tests/testthat/test_basic.R:565:1: style: Trailing whitespace is superfluous.
/home/runner/work/LightGBM/LightGBM/R-package/tests/testthat/test_basic.R:581:1: style: Trailing whitespace is superfluous.
/home/runner/work/LightGBM/LightGBM/R-package/tests/testthat/test_basic.R:585:1: style: Trailing whitespace is superfluous.
/home/runner/work/LightGBM/LightGBM/R-package/tests/testthat/test_basic.R:590:1: style: Trailing whitespace is superfluous.
/home/runner/work/LightGBM/LightGBM/R-package/tests/testthat/test_basic.R:594:1: style: Trailing whitespace is superfluous.
/home/runner/work/LightGBM/LightGBM/R-package/tests/testthat/test_basic.R:598:1: style: Trailing whitespace is superfluous.
```
https://github.com/microsoft/LightGBM/runs/4051576042?check_suite_focus=true#step:3:790

mikemahoney218:
@StrikerRUS We'll see what CI thinks but should be fixed in 0070402

jameslamb (approved):
Changes all look good to me, thanks very much!!

shiyu1994:
@mikemahoney218 @StrikerRUS @jameslamb Thank you all for the excellent work.

PR #4749, opened 2021-10-29
[python] Make dummy classes constructible with any arguments

StrikerRUS:
Refer to the following thread https://github.com/microsoft/LightGBM/pull/4639#discussion_r720068911. Making dummy classes constructible with any arguments may help to write tests.

I'm not overriding default `__init__()` method for three internal classes (prefixed with `_`) because doing that for two of them (`_LGBMRegressorBase` and `_LGBMClassifierBase`) makes `autosummary` Sphinx extension to fail because of inheritance issues.


jameslamb (approved):
very nice!

PR #4750, opened 2021-10-29
[docs][python] add `datatable` to the mocked modules during docs building process and sort them alphabetically

StrikerRUS:
I think it's better to add `datatable` because we try to import it here
https://github.com/microsoft/LightGBM/blob/d62378b45314c2688deef7333ae16fcc9863d881/python-package/lightgbm/compat.py#L43


jameslamb (approved):
I think this is ok, and consistent with the handling of all the other libraries we import. Thanks for also taking this opportunity to alphabetize `MOCK_MODULES` and split it to one library per line so it's easier to read and so the git blame will be more specific if future such changes are needed.

PR #4753, opened 2021-10-30
[python][sklearn] add `n_estimators_` and `n_iter_` post-fit attributes

StrikerRUS:
Ease access to the **real** number of boosted iterations performed during the training.
`n_estimators_` is added to highlight the possible difference between initially passed `n_estimators` and the actual number of iterations.
`n_iter_` is added to mimic `HistGradientBoosting[Regressor/Classifier]` interface.
https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.HistGradientBoostingRegressor.html#sklearn.ensemble.HistGradientBoostingRegressor.n_iter_

### DIFF
--- a/python-package/lightgbm/sklearn.py
+++ b/python-package/lightgbm/sklearn.py
@@ -848,15 +848,23 @@ class LGBMModel(_LGBMModelBase):
         return self._objective
 
     @property
-    def n_estimators_(self):
-        """:obj:`int`: The concrete number of boosted trees this model contains after fitting."""
+    def n_estimators_(self) -> int:
+        """:obj:`int`: True number of boosting iterations performed.
+
+        This might be less than parameter ``n_estimators`` if early stopping was enabled or
+        if boosting stopped early due to limits on complexity like ``min_gain_to_split``.
+        """
         if not self.__sklearn_is_fitted__():
             raise LGBMNotFittedError('No n_estimators found. Need to call fit beforehand.')
         return self._Booster.current_iteration()
 
     @property
-    def n_iter_(self):
-        """:obj:`int`: Number of iterations of the boosting process."""
+    def n_iter_(self) -> int:
+        """:obj:`int`: True number of boosting iterations performed.
+
+        This might be less than parameter ``n_estimators`` if early stopping was enabled or
+        if boosting stopped early due to limits on complexity like ``min_gain_to_split``.
+        """
         if not self.__sklearn_is_fitted__():
             raise LGBMNotFittedError('No n_iter found. Need to call fit beforehand.')
         return self._Booster.current_iteration()
--- a/python-package/lightgbm/sklearn.py
+++ b/python-package/lightgbm/sklearn.py
@@ -848,6 +848,20 @@ class LGBMModel(_LGBMModelBase):
         return self._objective
 
     @property
+    def n_estimators_(self):
+        """:obj:`int`: The concrete number of boosted trees this model contains after fitting."""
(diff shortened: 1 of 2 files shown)


python-package/lightgbm/sklearn.py
```diff
@@ -847,6 +847,20 @@ def objective_(self):
             raise LGBMNotFittedError('No objective found. Need to call fit beforehand.')
         return self._objective
 
+    @property
+    def n_estimators_(self):
+        """:obj:`int`: The concrete number of boosted trees this model contains after fitting."""
```
jameslamb, line 852: I think users who are not very familiar with LightGBM specifically and / or GBDTs generally might not understand what "concrete number of boosted trees" means, or why it would be different from parameter `num_iterations` and its aliases.

I also think that phrase could be misinterpreted in the case of multi-class classification, where LightGBM fits one tree per class and the total number of *trees* is actually `n_classes * n_iterations`.

Would you consider something like this, here and for `n_iter_`?

> True number of boosting iterations performed. This might be less than parameter `num_iterations` if early stopping was enabled or if boosting stopped early due to limits on complexity like `min_gain_to_split`.

The `min_gain_to_split` comment is a reference to https://github.com/microsoft/LightGBM/blob/9e1d7fa1bb395a9c4e5f7af9b13b9eb315e360fa/src/boosting/gbdt.cpp#L441

---------

Also want to note here that the `sklearn.ensemble.GradientBoostingClassifier.n_estimators_` doc also references early stopping explicitly.

https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.GradientBoostingClassifier.html

python-package/lightgbm/sklearn.py
```diff
@@ -847,6 +847,20 @@ def objective_(self):
             raise LGBMNotFittedError('No objective found. Need to call fit beforehand.')
         return self._objective
 
+    @property
+    def n_estimators_(self):
```
jameslamb, line 851:
```suggestion
    def n_estimators_(self) -> int:
```

I think using return type hints here can help `mypy` to catch potential bugs in other code referencing this property (both in `lightgbm` and in projects relying on `lightgbm`).

python-package/lightgbm/sklearn.py
```diff
@@ -847,6 +847,20 @@ def objective_(self):
...
+        """:obj:`int`: The concrete number of boosted trees this model contains after fitting."""
+        if not self.__sklearn_is_fitted__():
+            raise LGBMNotFittedError('No n_estimators found. Need to call fit beforehand.')
+        return self._Booster.current_iteration()
+
+    @property
+    def n_iter_(self):
```
jameslamb, line 858:
```suggestion
    def n_iter_(self) -> int:
```

jameslamb (changes requested):
I support this idea! Please see my proposals for changes to the documentation.

python-package/lightgbm/sklearn.py
StrikerRUS: Thank you! I agree with the proposed wording.
I believe that more useful description is better than keeping consistent with scikit-learn's wording for `n_iter_`.

python-package/lightgbm/sklearn.py
StrikerRUS: Ah, completely forgot about that we're trying to add type hints everywhere, sorry.

jameslamb (approved):
Thanks for the changes, looks good to me!

Issue #4706, opened 2021-10-22
[python] Changing `n_jobs` through scikit-learn interface has no effect


david-cortes:
If I set `n_jobs` in a fitted lightgbm model through the scikit-learn interface, the change has no effect.

Example:
```python
import numpy as np
from lightgbm import LGBMRegressor
from sklearn.datasets import fetch_california_housing
X, y = fetch_california_housing(return_X_y=True)

import numpy as np
X_long = np.repeat(X, 1000).reshape((-1, X.shape[1]))

model = LGBMRegressor(n_jobs=4).fit(X, y)
```

Now start watching the resource usage for the process and run this:
```python
model.set_params(n_jobs=1)
pred = model.predict(X_long)
```

PR #4756, opened 2021-10-30
[docs] improve docs about `nthreads` parameter

StrikerRUS:
Improve docs according to the suggestion in https://github.com/microsoft/LightGBM/issues/4706#issuecomment-949979204.


jameslamb (approved):
I think this clarification makes sense, thank you.

PR #4758, opened 2021-10-31
[python][sklearn] respect objective aliases

StrikerRUS:
I believe this is quite odd to add objective alias used in sklearn (`loss` #4637) and simply ignore any objective aliases at the same time.

### DIFF
--- a/python-package/lightgbm/sklearn.py
+++ b/python-package/lightgbm/sklearn.py
@@ -578,21 +578,32 @@ class LGBMModel(_LGBMModelBase):
             feature_name='auto', categorical_feature='auto',
             callbacks=None, init_model=None):
         """Docstring is set after definition, using a template."""
+        params = self.get_params()
+
+        params.pop('objective', None)
+        for alias in _ConfigAliases.get('objective'):
+            if alias in params:
+                self._objective = params.pop(alias)
+                _log_warning(f"Found `{alias}` in params. Will use it instead of argument")
         if self._objective is None:
             if isinstance(self, LGBMRegressor):
                 self._objective = "regression"
             elif isinstance(self, LGBMClassifier):
-                self._objective = "binary"
+                if self._n_classes > 2:
+                    self._objective = "multiclass"
+                else:
+                    self._objective = "binary"
             elif isinstance(self, LGBMRanker):
                 self._objective = "lambdarank"
             else:
                 raise ValueError("Unknown LGBMModel type.")
         if callable(self._objective):
             self._fobj = _ObjectiveFunctionWrapper(self._objective)
+            params['objective'] = 'None'  # objective = nullptr for unknown objective
         else:
             self._fobj = None
+            params['objective'] = self._objective
 
-        params = self.get_params()
         # user can set verbose with kwargs, it has higher priority
         if self.silent != "warn":
             _log_warning("'silent' argument is deprecated and will be removed in a future release of LightGBM. "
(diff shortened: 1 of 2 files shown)


tests/python_package_test/test_sklearn.py
```diff
@@ -910,10 +938,8 @@ def test_metrics():
     assert 'multi_logloss' in gbm.evals_result_['training']
     assert 'multi_error' in gbm.evals_result_['training']
 
-    # invalid objective is replaced with default multiclass one
-    # and invalid binary metric is replaced with multiclass alternative
-    gbm = lgb.LGBMClassifier(objective='invalid_obj',
-                             **params).fit(eval_metric='binary_error', **params_fit)
```
StrikerRUS, line 916: I believe this behavior is controversial and I'd better remove such silence replacement of objective function and instead raise error.

python-package/lightgbm/sklearn.py
```diff
@@ -621,20 +632,15 @@ def fit(self, X, y,
...
-        params['objective'] = self._objective
-        if self._fobj:
-            params['objective'] = 'None'  # objective = nullptr for unknown objective
 
         # Do not modify original args in fit function
         # Refer to https://github.com/microsoft/LightGBM/pull/2619
         eval_metric_list = copy.deepcopy(eval_metric)
         if not isinstance(eval_metric_list, list):
             eval_metric_list = [eval_metric_list]
-
```
jameslamb, line 633: I don't support this line and the three others below it removing blank lines. I think the resulting block of code is more difficult to read than the current state.

**`master`**

image


**this branch**

image

jameslamb (changes requested):
Totally support these changes and agree that this makes the scikit-learn API behave in a way that's closer to what we've told users to expect (where parameter aliases can always be used to override keyword arguments).

I just left one minor suggestion about some of the unrelated whitespace changes.

python-package/lightgbm/sklearn.py
```diff
@@ -578,21 +578,32 @@ def fit(self, X, y,
             feature_name='auto', categorical_feature='auto',
             callbacks=None, init_model=None):
         """Docstring is set after definition, using a template."""
+        params = self.get_params()
+
+        params.pop('objective', None)
+        for alias in _ConfigAliases.get('objective'):
+            if alias in params:
+                self._objective = params.pop(alias)
+                _log_warning(f"Found `{alias}` in params. Will use it instead of argument")
```
shiyu1994, line 587: Is this sentence complete? Should it be `Will use it instead of argument `objective``?

tests/python_package_test/test_sklearn.py
shiyu1994: I agree. We should explicitly tell the user what they intended to use is not valid. Or at least we should give a warning.

python-package/lightgbm/sklearn.py
StrikerRUS: Ah, OK. Reverted in 771cb5859c704f5e2ed7a58d8abb5a2744604b68.

python-package/lightgbm/sklearn.py
StrikerRUS: In this particular case I agree with you. Warning message was improved in 2d0392ef6c5e5a8670dcc22684c63afe9faaa2ba.

However, please note that in this similar warning
https://github.com/microsoft/LightGBM/blob/da98f24711a2faab17f94e5b2a636e6609c93fa6/python-package/lightgbm/engine.py#L571-L574

we cannot add specific argument name which is being overwritten because it depends on whether user uses `train()` function directly or indirectly from the sklearn-wrapper. Please consider the following example:
```python
import lightgbm as lgb

from sklearn.datasets import load_boston

X, y = load_boston(return_X_y=True)
lgb_train = lgb.Dataset(X, y)

lgb.train({'n_iter': 5}, lgb_train)  # here n_iter is used instead of num_boost_round argument

lgb.LGBMRegressor(n_iter=5).fit(X, y)  # here n_iter is used instead of n_estimators argument
```

jameslamb (approved):

StrikerRUS:
Should we block merging of this PR for one week similarly to https://github.com/microsoft/LightGBM/pull/4580#issuecomment-956509886?

jameslamb:
> Should we block merging of this PR for one week similarly to [#4580 (comment)](https://github.com/microsoft/LightGBM/pull/4580#issuecomment-956509886)?

I support not merging this for a few more days, yes.

shiyu1994 (approved):

PR #3549, opened 2020-11-09
[ci] fix CUDA CI builds

StrikerRUS:
I was wrong in https://github.com/microsoft/LightGBM/pull/3524#pullrequestreview-522772442. I guess to use v2 we need update Actions runner first: https://github.com/actions/checkout/issues/348. @guolinke Could you help please? For now rollback to v1.
```
Run actions/checkout@v2.3.4
  with:
    fetch-depth: 5
    submodules: true
    repository: microsoft/LightGBM
    token: ***
    ssh-strict: true
    persist-credentials: true
    clean: true
    lfs: false
Error: No such file or directory
```

PR #4763, opened 2021-11-02
[ci] upgrade actions/checkout to v2.4.0

jameslamb:
This project's CI jobs running on GitHub Actions use the `actions/checkout` action to clone the repo at the beginning of builds. Most jobs in the repo are currently using v2.3.4 of this action (CUDA jobs are using v1).

This PR proposes upgrading all jobs to v2.4.0, the newest release of that action. Just to keep up with changes there and take advantage of small bugfixes.

See https://github.com/actions/checkout/releases for details on the newest releases of this action.

### DIFF
--- a/.github/workflows/cuda.yml
+++ b/.github/workflows/cuda.yml
@@ -60,11 +60,9 @@ jobs:
             sudo chmod a+rw /var/run/docker.sock
             sudo systemctl restart docker
       - name: Remove old folder with repository
-        run: |
-          sudo rm -rf $GITHUB_WORKSPACE
-          mkdir -p $GITHUB_WORKSPACE
+        run: sudo rm -rf $GITHUB_WORKSPACE
       - name: Checkout repository
-        uses: actions/checkout@v2.4.0
+        uses: actions/checkout@v2.3.4
         with:
           fetch-depth: 5
           submodules: true
--- a/.github/workflows/cuda.yml
+++ b/.github/workflows/cuda.yml
@@ -62,7 +62,7 @@ jobs:
       - name: Remove old folder with repository
         run: sudo rm -rf $GITHUB_WORKSPACE
       - name: Checkout repository
-        uses: actions/checkout@v1
+        uses: actions/checkout@v2.4.0
         with:
           fetch-depth: 5
           submodules: true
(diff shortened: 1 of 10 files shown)


jameslamb:
hmmm, I guess this must be the reason CUDA builds were still on `actions/checkout@v1`:

> Syncing repository: microsoft/LightGBM
Getting Git version info
  Working directory is '/home/guoke/actions-runner/_work/LightGBM/LightGBM'
  /usr/bin/git version
  git version 2.17.1
Deleting the contents of '/home/guoke/actions-runner/_work/LightGBM/LightGBM'
The repository will be downloaded using the GitHub REST API
To create a local Git repository instead, add Git 2.18 or higher to the PATH
Error: Input 'submodules' not supported when falling back to download using the GitHub REST API. To create a local Git repository instead, add Git 2.18 or higher to the PATH.

https://github.com/microsoft/LightGBM/runs/4086585308?check_suite_focus=true

I'll revert the changes to those jobs in this PR.

StrikerRUS:
We cannot use `v2` on CUDA builds because of node installation issues: https://github.com/microsoft/LightGBM/pull/3549#issuecomment-724951057.

.github/workflows/cuda.yml
```diff
@@ -62,7 +62,7 @@ jobs:
       - name: Remove old folder with repository
         run: sudo rm -rf $GITHUB_WORKSPACE
       - name: Checkout repository
-        uses: actions/checkout@v1
+        uses: actions/checkout@v2.3.4
```
jameslamb, line 65:
```suggestion
        uses: actions/checkout@v1
```

jameslamb:
> We cannot use `v2` on CUDA builds because of node.js installation issues: [#3549 (comment)](https://github.com/microsoft/LightGBM/pull/3549#issuecomment-724951057).

ah ok, thanks! I wasn't able to find that from search but I had a feeling there must be a specific reason I'd forgotten.

StrikerRUS:
Linking #3763.

StrikerRUS:
Just a note: it is OK that solaris and valgrind checks are still using `2.3.4` version.
> Run actions/checkout@v2.3.4
   https://github.com/microsoft/LightGBM/runs/4086606076?check_suite_focus=true#step:4:1

That happens because ChatOps CI jobs are pulling GitHub Actions Workflow configuration file from the `master` branch. Refer to https://github.com/microsoft/LightGBM/pull/4503#issue-963553881.

jameslamb:
Ha right right! I keep forgetting that, sorry.

PR #4765, opened 2021-11-03
[ci] use wch1/r-debug image in Solaris tests

jameslamb:
As a result of the changes #3946 , which introduce `vignettes/` to the R package (part of #1944), building the R package with `R CMD build` will include installing the package, since it must be installed to create vignette outputs.

To make this work in the existing Solaris CI job, two changes are required:

* `pandoc` must be available in the container used to build the package before submitting it to R Hub
* `{lightgbm}`'s dependencies must be installed *before* running `sh build-cran-package.sh`.

This PR proposes both of those changes. It proposes the re-organization of `test_r_package_solaris.sh` just to make the diff of #3946 slightly smaller and easier to review.

Changing the image used in the Solaris job (to one that already has `pandoc` installed) needs to be done in a separate PR, since GitHub Actions configs for jobs triggered by comments are sourced from `master`  (https://github.com/microsoft/LightGBM/pull/4397#issuecomment-866790489).

### DIFF
--- a/.github/workflows/r_solaris.yml
+++ b/.github/workflows/r_solaris.yml
@@ -9,7 +9,7 @@ jobs:
     name: solaris-cran
     timeout-minutes: 120
     runs-on: ubuntu-latest
-    container: rocker/r-base
+    container: wch1/r-debug
     env:
       SECRETS_WORKFLOW: ${{ secrets.WORKFLOW }}
     steps:
@@ -19,7 +19,6 @@ jobs:
           apt-get update
           apt-get install --no-install-recommends -y \
             curl \
-            git \
             jq
       - name: Checkout repository
         uses: actions/checkout@v2.3.4
--- a/.ci/test_r_package_solaris.sh
+++ b/.ci/test_r_package_solaris.sh
@@ -1,13 +1,13 @@
 #!/bin/bash
 
-sh build-cran-package.sh || exit -1
-
 apt-get install --no-install-recommends -y \
   libcurl4-openssl-dev \
   libxml2-dev \
   libssl-dev
 
-log_file="rhub_logs.txt"
 Rscript -e "install.packages('rhub', dependencies = c('Depends', 'Imports', 'LinkingTo'), repos = 'https://cran.r-project.org', Ncpus = parallel::detectCores())" || exit -1
 
+sh build-cran-package.sh || exit -1
+
+log_file="rhub_logs.txt"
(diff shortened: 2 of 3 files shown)


.github/workflows/r_solaris.yml
```diff
@@ -19,7 +19,6 @@ jobs:
           apt-get update
           apt-get install --no-install-recommends -y \
             curl \
-            git \
```
jameslamb, line 22: `git` already exists in this image.

```shell
docker run --entrypoint="" -it wch1/r-debug which git
# /usr/bin/git
```

.github/workflows/r_solaris.yml
```diff
@@ -9,7 +9,7 @@ jobs:
     name: solaris-cran
     timeout-minutes: 120
     runs-on: ubuntu-latest
-    container: rocker/r-base
+    container: wch1/r-debug
```
jameslamb, line 12: just noting this is the same image used in sanitizer jobs:

https://github.com/microsoft/LightGBM/blob/da98f24711a2faab17f94e5b2a636e6609c93fa6/.github/workflows/r_package.yml#L175-L179

And those jobs are successfully building the package with vignettes.

https://github.com/microsoft/LightGBM/runs/4086356142?check_suite_focus=true

more evidence that this image has pandoc:

```shell
docker run --entrypoint="" -it wch1/r-debug pandoc --version

# pandoc 2.5
# Compiled with pandoc-types 1.17.5.4, texmath 0.11.2.2, skylighting 0.7.7
# Default user data directory: /root/.pandoc
```

.github/workflows/r_solaris.yml
StrikerRUS: Then this simplification can be done for `r_valgrind` job as well (not necessary to do it in the same PR):
https://github.com/microsoft/LightGBM/blob/8deb90948a481820ba2109166038b146ae63a5dd/.github/workflows/r_valgrind.yml#L12-L23

StrikerRUS (approved):

.github/workflows/r_solaris.yml
jameslamb: oh nice! I'll push that change here.

jameslamb:
Solaris job is working with this change! https://github.com/microsoft/LightGBM/pull/3946#issuecomment-959783925

`valgrind` timed out (https://github.com/microsoft/LightGBM/pull/3946#issuecomment-959784081), tried triggering it again (https://github.com/microsoft/LightGBM/pull/3946#issuecomment-960242417)

Issue #4771, opened 2021-11-05
[dask] [python] LightGBM hangs indefinitely after an error


jameslamb:
## Description

Shortly after `dask=2021.10.0` and `distributed=2021.10.0` were uploaded to the Anaconda default channels (used in LightGBM's CI), LightGBM's CI job started failing with timeouts (#4769).

Logs suggested that one of the unit tests on `lightgbm.dask` was hanging indefinitely. Specifically, the test after `test_errors`.

## Reproducible example

Run inside a container with `conda` available.

```shell
docker run \
    --rm \
    --workdir /usr/local/testing \
    -it \
    continuumio/miniconda3:4.10.3 \
    /bin/bash

apt-get update
apt-get install \
    --no-install-recommends \
    -y \
        build-essential
```

Using `dask` and `distributed` v2021.10.0, after a run which produces an error out on the cluster, the next test gets stuck indefinitely.

```shell
conda create \
    --name dask-test \
    --override-channels \
    -c conda-forge \
    --yes \
        cloudpickle=2.0.0 \
        cmake=3.21.3 \
        dask=2021.10.0 \
        distributed=2021.10.0 \
        joblib=1.1.0 \
        numpy=1.21.2 \
        pandas=1.3.4 \
        pytest=6.2.4 \
        scikit-learn=1.0.1 \
        scipy=1.7.1

source activate dask-test

git clone \
    --recursive \
    https://github.com/microsoft/LightGBM.git

cd LightGBM

pushd python-package/
    python setup.py install
popd

# this will succeed
pytest -v \
    tests/python_package_test/test_dask.py \
    -k 'test_errors'

# this will succeed
pytest -v \
    tests/python_package_test/test_dask.py \
    -k 'test_training_succeeds_even_if_some_workers_do_not_have_any_data'

# running both tests, the "training_succeeds" one hangs indefinitely
pytest -v \
    tests/python_package_test/test_dask.py \
    -k 'test_errors or test_training_succeeds_even_if_some_workers_do_not_have_any_data'
```

Downgrading to `dask` / `distributed` 2021.9.0 fixes the issue.

```shell
source deactivate
conda env remove --name test-dask
conda create \
    --name dask-test \
    --override-channels \
    -c conda-forge \
    --yes \
        cloudpickle=2.0.0 \
        cmake=3.21.3 \
        dask=2021.9.1 \
        distributed=2021.9.1 \
        joblib=1.1.0 \
        numpy=1.21.2 \
[...]

PR #4772, opened 2021-11-05
[tests][dask] fix hangs after exception in worker

jmoralez:
As described in #4771, with `dask==2021.10.0` and `distributed==2021.10.0` when a worker encounters an exception in a test as happens in https://github.com/microsoft/LightGBM/blob/2394b41ac2bc64c36115c4eb314633dede0c6b9c/tests/python_package_test/test_dask.py#L1309-L1310
the following tests hang indefinitely.

This adds a `client.restart` to restart the workers and get rid of the raised exception, since as of #4159 all of LightGBM's Dask unit tests share the same cluster.


StrikerRUS:
With added `client.restart()` only one job timed out.
```
../tests/python_package_test/test_dask.py .............................. [ 14%]
........................................................................ [ 25%]
........................................................................ [ 37%]
......s...............s...............s...............s................. [ 48%]
..........................
##[error]The operation was canceled.
```
https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=11463&view=logs&j=e3619e06-e539-5956-1ae7-e8c2c29d8e91&t=4b3d6e5d-f8c3-577f-235f-b68c5ed4d3ab&l=1238

But I guess it means that this workaround cannot be treated as reliable enough and it brings flakiness into our tests.

jameslamb:
> But I guess it means that this workaround cannot be treated as reliable enough

I agree, it seems like this doesn't totally fix the issue 😕 

we'll have to keep investigating in #4771

jmoralez:
Seems like it should work, I don't see what could be hanging with fresh workers. I can maybe try increasing the timeout in the restart or add a sleep after it. What do you guys think?

jameslamb:
>  can maybe try increasing the timeout in the restart or add a sleep after it

Sure, try it! We'd probably learn something from that exercise either way.

Feel free to use this PR to test strategies. One trick I like to use, when I need to use CI to test, is to have a single commit that removes all the other CI jobs that aren't relevant to what I'm working on, and then bring those CI configs back in later with `git revert`.

So for you, that might be like

```shell
git rm .appveyor.yaml
git rm .github/workflows/r_*.yaml
git add .
git commit -m "remove unrelated CI configs"
```

jmoralez:
Wow [this job](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=11470&view=logs&j=e3619e06-e539-5956-1ae7-e8c2c29d8e91&t=4b3d6e5d-f8c3-577f-235f-b68c5ed4d3ab) really broke

jmoralez:
Closing this as `client.restart` doesn't seem like the way to go.

jameslamb:
> Wow [this job](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=11470&view=logs&j=e3619e06-e539-5956-1ae7-e8c2c29d8e91&t=4b3d6e5d-f8c3-577f-235f-b68c5ed4d3ab) really broke

whoa! Ok thanks for investigating. We'll have to try to find another approach.

PR #4774, opened 2021-11-06
[docs] add specific estimator names to sklearn fit() docs

jameslamb:
This PR proposes adding links to the specific estimator class in the Return doc for `.fit()` methods in LightGBM's scikit-learn interface.

Based on @jmoralez 's suggestion: https://github.com/microsoft/LightGBM/pull/4716#issuecomment-952033373

### How I tested this

```shell
cd docs
make clean html
open _build/html/index.html
```

Looks like the docs are rendering correctly.

image

I also clicked through the links next to `Return type` and confirmed that they link back to the constructor docs for the relevant class.

### DIFF
--- a/python-package/lightgbm/sklearn.py
+++ b/python-package/lightgbm/sklearn.py
@@ -262,7 +262,7 @@ _lgbmmodel_doc_fit = (
 
     Returns
     -------
-    self : object
+    self : LGBMModel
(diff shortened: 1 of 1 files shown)


StrikerRUS (approved):
Thank you for improving these docstrings!

PR #4776, opened 2021-11-06
[ci] Revert temp workaround for `freetype` version

StrikerRUS:
This reverts commit d62378b45314c2688deef7333ae16fcc9863d881.

Something was changed in Windows environments and we don't need that workaround anymore.


jameslamb (approved):
great!

PR #4782, opened 2021-11-08
Add customized parser support

chjinche:
## Motivation
Currently, lightGBM can only support three text data formats (tsv, csv, libsvm), and data type check process is kind of coarse. For example, a tsv may be judged as libsvm if there is string feature containing colon. Besides, lightGBM does not support feature transformation, users need to do separate transformation themselves, which may be a bit inconvenient.

## Proposal
In this PR, we'd like to add customized parser to unlock free data preprocessing scenario. Users could use lightgbm as dependency, and follow `parser` interface to design their own parsers or could simply try [freeform parser](https://github.com/microsoft/lightgbm-transform/blob/main/examples/freeform2_parser.cpp). 
Compared to separate transformation. this way has several pros:

- More efficient. Data preprocessing can go with parsing each line, and take advantage of multi-processing designed by lightgbm naturally. No need to store whole transformed data in file/memory.
- More convenient for development and iteration. Built-in transformation could keep offline and online consistent well.

## Changes
The main code changes are:
- Add `parser_config_file` to config.
- Add parser factory and reflector to support customized parser.
- Save/load parser config content along with model for automatic data processing in prediction.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,8 @@ MLflow (experiment tracking, model monitoring framework): https://github.com/mlf
 
 `{mlr3extralearners}` (R `{mlr3}`-compliant interface): https://github.com/mlr-org/mlr3extralearners
 
+lightgbm-transform (feature transformation binding): https://github.com/microsoft/lightgbm-transform
+
 Support
 -------
 
--- a/docs/Parameters.rst
+++ b/docs/Parameters.rst
@@ -856,7 +856,7 @@ Dataset Parameters
 
    -  see `lightgbm-transform <https://github.com/microsoft/lightgbm-transform>`__ for usage examples
 
-   -  **Note**: ``lightgbm-transform`` is not maintained by LightGBM's maintainers. Bug reports or feature requests should go to `issue page <https://github.com/microsoft/lightgbm-transform/issues>`__
+   -  **Note**: ``lightgbm-transform`` is not maintained by LightGBM's maintainers. Bug reports or feature requests should go to `issues page <https://github.com/microsoft/lightgbm-transform/issues>`__
 
 Predict Parameters
 ~~~~~~~~~~~~~~~~~~
--- a/docs/Parameters.rst
+++ b/docs/Parameters.rst
@@ -850,6 +850,14 @@ Dataset Parameters
 
    -  **Note**: setting this to ``true`` may lead to much slower text parsing
 
+-  ``parser_config_file`` :raw-html:`<a id="parser_config_file" title="Permalink to this parameter" href="#parser_config_file">&#x1F517;&#xFE0E;</a>`, default = ``""``, type = string
+
+   -  path to a ``.json`` file that specifies customized parser initialized configuration
+
+   -  see `lightgbm-transform <https://github.com/microsoft/lightgbm-transform>`__ for usage examples
+
+   -  **Note**: ``lightgbm-transform`` is not maintained by LightGBM's maintainers. Bug reports or feature requests should go to `issue page <https://github.com/microsoft/lightgbm-transform/issues>`__
+
 Predict Parameters
 ~~~~~~~~~~~~~~~~~~
 
--- a/include/LightGBM/config.h
+++ b/include/LightGBM/config.h
@@ -723,7 +723,7 @@ struct Config {
 
   // desc = path to a ``.json`` file that specifies customized parser initialized configuration
   // desc = see `lightgbm-transform <https://github.com/microsoft/lightgbm-transform>`__ for usage examples
-  // desc = **Note**: ``lightgbm-transform`` is not maintained by LightGBM's maintainers. Bug reports or feature requests should go to `issue page <https://github.com/microsoft/lightgbm-transform/issues>`__
+  // desc = **Note**: ``lightgbm-transform`` is not maintained by LightGBM's maintainers. Bug reports or feature requests should go to `issues page <https://github.com/microsoft/lightgbm-transform/issues>`__
   std::string parser_config_file = "";
 
   #pragma endregion
--- a/include/LightGBM/config.h
+++ b/include/LightGBM/config.h
@@ -721,6 +721,11 @@ struct Config {
   // desc = **Note**: setting this to ``true`` may lead to much slower text parsing
   bool precise_float_parser = false;
 
+  // desc = path to a ``.json`` file that specifies customized parser initialized configuration
+  // desc = see `lightgbm-transform <https://github.com/microsoft/lightgbm-transform>`__ for usage examples
+  // desc = **Note**: ``lightgbm-transform`` is not maintained by LightGBM's maintainers. Bug reports or feature requests should go to `issue page <https://github.com/microsoft/lightgbm-transform/issues>`__
+  std::string parser_config_file = "";
+
   #pragma endregion
 
   #pragma region Predict Parameters
--- a/include/LightGBM/dataset.h
+++ b/include/LightGBM/dataset.h
@@ -15,6 +15,7 @@
 
 #include <string>
 #include <functional>
+#include <map>
 #include <memory>
 #include <mutex>
 #include <unordered_set>
@@ -254,6 +255,14 @@ class Parser {
  public:
   typedef const char* (*AtofFunc)(const char* p, double* out);
 
+  /*! \brief Default constructor */
+  Parser() {}
+
+  /*!
+  * \brief Constructor for customized parser. The constructor accepts content not path because need to save/load the config along with model string
+  */
+  explicit Parser(std::string) {}
+
   /*! \brief virtual destructor */
   virtual ~Parser() {}
 
@@ -271,12 +280,58 @@ class Parser {
   /*!
   * \brief Create an object of parser, will auto choose the format depend on file
   * \param filename One Filename of data
+  * \param header whether input file contains header
   * \param num_features Pass num_features of this data file if you know, <=0 means don't know
   * \param label_idx index of label column
   * \param precise_float_parser using precise floating point number parsing if true
   * \return Object of parser
   */
   static Parser* CreateParser(const char* filename, bool header, int num_features, int label_idx, bool precise_float_parser);
+
+  /*!
+  * \brief Create an object of parser, could use customized parser, or auto choose the format depend on file
+  * \param filename One Filename of data
+  * \param header whether input file contains header
+  * \param num_features Pass num_features of this data file if you know, <=0 means don't know
+  * \param label_idx index of label column
+  * \param precise_float_parser using precise floating point number parsing if true
+  * \param parser_config_str Customized parser config content
+  * \return Object of parser
+  */
+  static Parser* CreateParser(const char* filename, bool header, int num_features, int label_idx, bool precise_float_parser,
+                              std::string parser_config_str);
+
+  /*!
+  * \brief Generate parser config str used for custom parser initialization, may save values of label id and header
+  * \param filename One Filename of data
+  * \param parser_config_filename One Filename of parser config
+  * \param header whether input file contains header
+  * \param label_idx index of label column
+  * \return Parser config str
+  */
+  static std::string GenerateParserConfigStr(const char* filename, const char* parser_config_filename, bool header, int label_idx);
+};
+
+/*! \brief Interface for parser factory, used by customized parser */
+class ParserFactory {
+ private:
+  ParserFactory() {}
+  std::map<std::string, std::function<Parser*(std::string)>> object_map_;
+
+ public:
+  ~ParserFactory() {}
+  static ParserFactory& getInstance();
+  void Register(std::string class_name, std::function<Parser*(std::string)> objc);
+  Parser* getObject(std::string class_name, std::string config_str);
+};
+
+/*! \brief Interface for parser reflector, used by customized parser */
+class ParserReflector {
+ public:
+  ParserReflector(std::string class_name, std::function<Parser*(std::string)> objc) {
+    ParserFactory::getInstance().Register(class_name, objc);
+  }
+  virtual ~ParserReflector() {}
 };
 
 /*! \brief The main class of data set,
@@ -605,6 +660,9 @@ class Dataset {
   /*! \brief Get names of current data set */
   inline const std::vector<std::string>& feature_names() const { return feature_names_; }
 
+  /*! \brief Get content of parser config file */
+  inline const std::string parser_config_str() const { return parser_config_str_; }
+
   inline void set_feature_names(const std::vector<std::string>& feature_names) {
     if (feature_names.size() != static_cast<size_t>(num_total_features_)) {
       Log::Fatal("Size of feature_names error, should equal with total number of features");
@@ -722,6 +780,7 @@ class Dataset {
   /*! map feature (inner index) to its index in the list of numeric (non-categorical) features */
   std::vector<int> numeric_feature_map_;
   int num_numeric_features_;
+  std::string parser_config_str_;
 };
 
 }  // namespace LightGBM
--- a/include/LightGBM/utils/common.h
+++ b/include/LightGBM/utils/common.h
@@ -8,6 +8,7 @@
 #if ((defined(sun) || defined(__sun)) && (defined(__SVR4) || defined(__svr4__)))
 #include <LightGBM/utils/common_legacy_solaris.h>
 #endif
+#include <LightGBM/utils/json11.h>
 #include <LightGBM/utils/log.h>
 #include <LightGBM/utils/openmp_wrapper.h>
 
@@ -61,6 +62,8 @@
 namespace LightGBM {
 
 namespace Common {
+
+using json11::Json;
 
 /*!
 * Imbues the stream with the C locale.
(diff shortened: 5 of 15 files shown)


src/boosting/gbdt.h
```diff
@@ -483,6 +485,10 @@ class GBDT : public GBDTBase {
   std::vector<std::unique_ptr<Tree>> models_;
   /*! \brief Max feature index of training data*/
   int max_feature_idx_;
+  /*! \brief Parser config file content */
+  std::string parser_config_str_ = "";
+  /*! \brief Model path */
+  std::string model_path_;
```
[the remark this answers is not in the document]
chjinche: Good catch! Removed.

include/LightGBM/dataset.h
```diff
@@ -605,6 +661,9 @@ class Dataset {
   /*! \brief Get names of current data set */
   inline const std::vector<std::string>& feature_names() const { return feature_names_; }
 
+  /*! \brief Get content of parser config file */
+  inline const std::string parser_config_str() const {return parser_config_str_;}
```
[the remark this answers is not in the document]
chjinche: Fixed.

include/LightGBM/dataset.h
```diff
@@ -271,12 +281,58 @@ class Parser {
...
+  * \return parser config str
+  */
+  static std::string GenerateParserConfigStr(const char* filename, const char* parser_config_filename, bool header, int label_idx);
+};
+
+/*! \brief Interface for parser factory, used by customized parser */
+class ParserFactory {
+ private:
+  ParserFactory() {}
+  std::map<std::string, std::function<Parser*(std::string)>> _object_map;
```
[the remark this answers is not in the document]
chjinche: Fixed.

include/LightGBM/dataset.h
```diff
@@ -254,6 +255,15 @@ class Parser {
  public:
   typedef const char* (*AtofFunc)(const char* p, double* out);
 
+  /*! \brief default constructor*/
+  Parser() {}
+
+  /*!
+  * \brief constructor for customized parser.
+  * \param Parser config content, not path because need to save/load the config along with model string.
```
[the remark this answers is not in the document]
chjinche: The constructor is added for customized parser created outside LightGBM repo, so if add parameter name here, unused parameter warning will be raised in compilation. And looks no documentation about `Parser` class I think. So could I remove the param description here and add notes to brief part instead?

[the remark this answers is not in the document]
shiyu1994: Yes. It is ok to remove the param description.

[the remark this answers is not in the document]
chjinche: Done. Thanks

shiyu1994:
@jameslamb @StrikerRUS @guolinke Could you please have a look at this PR if you have time. Thanks.

StrikerRUS:
@chjinche From the first glance this PR looks really good and doesn't break backward compatibility. Thanks a lot for finding a way to support feature transformations in a non-invasive for core LightGBM API way! I'll provide more thoughtful review when have time.

chjinche:
@StrikerRUS Much appreciated for reviewing and glad that you like the implementation 😍  
As the feature is required urgently by Microsoft Internal products, and we have already spent more time on making the integration change 😅, it will be very grateful if you could provide feedbacks **before next week**. Your views are important to us, we're very happy to work on the feature with all of you! 😄

jameslamb (dismissed):
It looks like merging this PR as-is would break the ability to use the LightGBM R package on Solaris, which would result in that package being rejected and removed from CRAN.

Here is the relevant error from https://builder.r-hub.io/status/lightgbm_3.3.1.99.tar.gz-917bd78bb1fe442a8c8d245586e29b13

```text
#> In file included from ./include/LightGBM/config.h:16:0,
1782#> from ./include/LightGBM/boosting.h:8,
1783#> from boosting/boosting.cpp:5:
1784#> ./include/LightGBM/utils/common.h:207:82: error: ‘INT_MAX’ was not declared in this scope
1785#> inline static std::string LoadStringFromFile(const char* filename, int row_num = INT_MAX) {
1786#> ^
1787#> gmake: *** [/opt/csw/lib/R/etc/Makeconf:177: boosting/boosting.o] Error 1
1788#> ERROR: compilation failed for package ‘lightgbm’
1789#> * removing ‘/export/home/XjiOipx/R/lightgbm’
1790#> Warning message:
1791#> In i.p(...) :
1792#> installation of package ‘/export/home/XjiOipx/Rtemp/RtmpoRqZQc/file6c8076a9dbf/lightgbm_3.3.1.99.tar.gz’ had non-zero exit status
```

I think this is might be a compiler-specific problem, and maybe explicitly including `<climits>` would fix it? Based on [this Stack Overflow answer](https://stackoverflow.com/a/40700885/3986677) which seems related.

Could you try that and re-run the Solaris CI job? ([link to instructions](https://github.com/microsoft/LightGBM/tree/master/R-package#solaris))

Sorry for the inconvenience. I hope the fix is that simple 😬

jameslamb:
Other than the issue with the Solaris job, I agree with @StrikerRUS ... this looks like a good approach for the integration! Thanks very very much for your efforts there.

I'm definitely interested in trying this out once the relevant changes have been merged in LightGBM and the `lightgbm-transform` docs have been updated. We had good success at a previous company I worked at bundling feature engineering and inference code together. I think it's a really powerful idea.

shiyu1994:
We can also use `std::numeric_limits<int>::max()` as an alternative to `INT_MAX`. Since it is already used in other place of the code base, see
https://github.com/microsoft/LightGBM/blob/6cbb3586b92112fc1f27a179b201ac924e5600cc/src/boosting/prediction_early_stop.cpp#L21
so it should be safe to use.

Inclusion of `#include <limits>` is required for this.

jameslamb:
> We can also use std::numeric_limits<int>::max()

Oh nice! I like that even better, and since it's used in other places in the code I think it's likely to work on Solaris.

src/io/dataset_loader.cpp
```diff
@@ -42,6 +42,14 @@ void DatasetLoader::SetHeader(const char* filename) {
     if (config_.header) {
       std::string first_line = text_reader.first_line();
       feature_names_ = Common::Split(first_line.c_str(), "\t,");
+    } else if (!Common::LoadStringFromFile(config_.parser_config_file.c_str()).empty()) {
+      // support to get header from parser config, so could utilize following label name to id mapping logic.
+      std::string header_in_parser_config = Common::GetFromParserConfig(
+        Common::LoadStringFromFile(config_.parser_config_file.c_str()), "header");
```
StrikerRUS, line 48: Please load config file only once (`Common::LoadStringFromFile()`).

tests/python_package_test/test_utilities.py
```diff
@@ -97,3 +99,15 @@ def dummy_metric(_, __):
                 actual_log_wo_gpu_stuff.append(line)
 
     assert "\n".join(actual_log_wo_gpu_stuff) == expected_log
+
+
+def test_smoke_custom_parser(tmp_path):
+    data_path = Path(__file__).absolute().parents[2] / 'examples/binary_classification/binary.train'
```
StrikerRUS, line 105:
```suggestion
    data_path = Path(__file__).absolute().parents[2] / 'examples' / 'binary_classification' / 'binary.train'
```

tests/python_package_test/test_utilities.py
```diff
@@ -97,3 +99,15 @@ def dummy_metric(_, __):
...
+    parser_config_file = tmp_path / 'parser.ini'
+    with open(parser_config_file, 'w') as fout:
+        fout.write("{\"className\": \"dummy\", \"id\":\"1\"}")
```
StrikerRUS, line 108:
```suggestion
        fout.write('{"className": "dummy", "id": "1"}')
```

tests/python_package_test/test_utilities.py
```diff
@@ -97,3 +99,15 @@ def dummy_metric(_, __):
...
+
+    data = lgb.Dataset(data_path, params={"parser_config_file": parser_config_file})
+    with pytest.raises(lgb.basic.LightGBMError,
+                       match="Cannot find parser class 'dummy', please register first or check config format"):
+        lgb.train({}, data)
```
StrikerRUS, line 113:
```suggestion
        data.construct()
```

tests/python_package_test/test_utilities.py
```diff
@@ -97,3 +99,15 @@ def dummy_metric(_, __):
                 actual_log_wo_gpu_stuff.append(line)
 
     assert "\n".join(actual_log_wo_gpu_stuff) == expected_log
+
+
+def test_smoke_custom_parser(tmp_path):
```
StrikerRUS, line 104: This file is used to test helper public functions from Python-package (https://lightgbm.readthedocs.io/en/latest/Python-API.html#utilities).
Please move this test into
https://github.com/microsoft/LightGBM/blob/master/tests/python_package_test/test_basic.py
as it checks core functionality of LightGBM.

StrikerRUS (changes requested):
Thanks again for this PR!
Please check some my suggestions and questions below:

include/LightGBM/utils/common.h
```diff
@@ -200,6 +204,43 @@ inline static std::vector<std::string> Split(const char* c_str, const char* deli
   return ret;
 }
 
+inline static std::string LoadStringFromFile(const char* filename, int row_num = INT_MAX) {
```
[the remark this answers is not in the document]
chjinche: @StrikerRUS Thanks for pointing out. I've removed this function, and use `TextReader` instead. And add a func named `JoinedLines` to class `TextReader`. FYI, @shiyu1994 @jameslamb looks compilation error for r solaris also could be solved in this way.

src/boosting/gbdt.cpp
```diff
@@ -120,6 +120,8 @@ void GBDT::Init(const Config* config, const Dataset* train_data, const Objective
   feature_names_ = train_data_->feature_names();
   feature_infos_ = train_data_->feature_infos();
   monotone_constraints_ = config->monotone_constraints;
+  // get parser config file content
+  parser_config_str_ = train_data_->parser_config_str();
```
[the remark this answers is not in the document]
chjinche: Added.

src/boosting/gbdt_model_text.cpp
```diff
@@ -399,6 +399,11 @@ std::string GBDT::SaveModelToString(int start_iteration, int num_iteration, int
     ss << loaded_parameter_ << "\n";
     ss << "end of parameters" << '\n';
   }
+  if (!parser_config_str_.empty()) {
```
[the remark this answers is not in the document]
chjinche: Updated to "v4".

src/io/dataset_loader.cpp
chjinche: Done.

tests/python_package_test/test_utilities.py
chjinche: Thanks for comments of this test!

tests/python_package_test/test_utilities.py
chjinche: Done.

tests/python_package_test/test_utilities.py
chjinche: Done.

tests/python_package_test/test_utilities.py
chjinche: Done.

docs/Parameters.rst
```diff
@@ -850,6 +850,14 @@ Dataset Parameters
 
    -  **Note**: setting this to ``true`` may lead to much slower text parsing
 
+-  ``parser_config_file`` :raw-html:`<a id="parser_config_file" title="Permalink to this parameter" href="#parser_config_file">&#x1F517;&#xFE0E;</a>`, default = ``""``, type = string
+
+   -  path to a ``.json`` file that specifies customized parser initialized configuration
+
+   -  see `lightgbm-transform <https://github.com/microsoft/lightgbm-transform>`__ for usage examples
+
+   -  **Note**: ``lightgbm-transform`` is not maintained by LightGBM's maintainers. Bug reports or feature requests should go to `issue <https://github.com/microsoft/lightgbm-transform/issues>`__
```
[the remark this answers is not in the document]
chjinche: Done.

include/LightGBM/config.h
```diff
@@ -721,6 +721,11 @@ struct Config {
   // desc = **Note**: setting this to ``true`` may lead to much slower text parsing
   bool precise_float_parser = false;
 
+  // desc = path to a ``.json`` file that specifies customized parser initialized configuration
+  // desc = see `lightgbm-transform <https://github.com/microsoft/lightgbm-transform>`__ for usage examples
+  // desc = **Note**: ``lightgbm-transform`` is not maintained by LightGBM's maintainers. Bug reports or feature requests should go to `issue <https://github.com/microsoft/lightgbm-transform/issues>`__
```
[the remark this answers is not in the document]
chjinche: Done.

README.md
```diff
@@ -117,6 +117,8 @@ MLflow (experiment tracking, model monitoring framework): https://github.com/mlf
 
 `{mlr3extralearners}` (R `{mlr3}`-compliant interface): https://github.com/mlr-org/mlr3extralearners
 
+lightgbm-transform (transformation binding): https://github.com/microsoft/lightgbm-transform
```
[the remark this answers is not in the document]
chjinche: Done.

include/LightGBM/utils/common.h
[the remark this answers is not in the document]
shiyu1994: That's great.

jameslamb (commented):
Looks ok to me now that valgrind and solaris tests are passing, thanks! Please wait for a review from @shiyu1994 before merging, my review shouldn't be enough for a merge on this one.

By the way, we use squash merging in this project, so it's not necessary to rebase, squash commits, and force-push. In fact, as a reviewer I personally would prefer that you didn't do squashing within a pull request, so it's possible to see the changes in each subsequent commit pushed in response to reviewers' comments.

StrikerRUS:
I would kindly ask next time address review comments in a separate commits and don't perform force pushes in a such kind of big PRs. Separate commits speed up next round of reviews by allowing to check that only requested changes were done and nothing else were touched by a bad merge commit for example.

include/LightGBM/utils/text_reader.h
```diff
@@ -84,6 +84,17 @@ class TextReader {
   * \return Text data, store in std::vector by line
   */
   inline std::vector<std::string>& Lines() { return lines_; }
+  /*!
+  * \brief Get joined text data that read from file
+  * \return Text data, store in std::string, joined all lines by delimiter "\n"
+  */
+  inline std::string JoinedLines() {
```
StrikerRUS, line 91: I believe that delimiter can be an argument of this function given the general purpose of this class.

docs/Parameters.rst
```diff
@@ -850,6 +850,14 @@ Dataset Parameters
 
    -  **Note**: setting this to ``true`` may lead to much slower text parsing
 
+-  ``parser_config_file`` :raw-html:`<a id="parser_config_file" title="Permalink to this parameter" href="#parser_config_file">&#x1F517;&#xFE0E;</a>`, default = ``""``, type = string
+
+   -  path to a ``.json`` file that specifies customized parser initialized configuration
+
+   -  see `lightgbm-transform <https://github.com/microsoft/lightgbm-transform>`__ for usage examples
+
+   -  **Note**: ``lightgbm-transform`` is not maintained by LightGBM's maintainers. Bug reports or feature requests should go to `issue page <https://github.com/microsoft/lightgbm-transform/issues>`__
```
StrikerRUS, line 859:
```suggestion
   -  **Note**: ``lightgbm-transform`` is not maintained by LightGBM's maintainers. Bug reports or feature requests should go to `issues page <https://github.com/microsoft/lightgbm-transform/issues>`__
```

include/LightGBM/config.h
```diff
@@ -721,6 +721,11 @@ struct Config {
   // desc = **Note**: setting this to ``true`` may lead to much slower text parsing
   bool precise_float_parser = false;
 
+  // desc = path to a ``.json`` file that specifies customized parser initialized configuration
+  // desc = see `lightgbm-transform <https://github.com/microsoft/lightgbm-transform>`__ for usage examples
+  // desc = **Note**: ``lightgbm-transform`` is not maintained by LightGBM's maintainers. Bug reports or feature requests should go to `issue page <https://github.com/microsoft/lightgbm-transform/issues>`__
```
StrikerRUS, line 726:
```suggestion
  // desc = **Note**: ``lightgbm-transform`` is not maintained by LightGBM's maintainers. Bug reports or feature requests should go to `issues page <https://github.com/microsoft/lightgbm-transform/issues>`__
```

StrikerRUS (commented):
✔️ 
LGTM, except two very minor suggestions below! Thank you very for for this enhancement!

I'm leaving only "comment" because I believe that only review from cpp maintainer should count towards a merge.

include/LightGBM/utils/text_reader.h
shiyu1994: Thanks for your suggestions. Let's do this in a follow-up PR. Since this one is emergency for our product.

shiyu1994 (approved):
The changes LGTM!

shiyu1994:
@chjinche Thanks for this great work! And thank @StrikerRUS @jameslamb for your review. BTW, @StrikerRUS, could you please mark the request changes as resolved. It seems that the requested changes have been addressed, but they still block the merging. Thanks for your detailed review.

docs/Parameters.rst
chjinche: My bad, fixed.

include/LightGBM/utils/text_reader.h
chjinche: Done.

shiyu1994:
I'm going to merged this given that all the checks. The latest suggestions of @StrikerRUS have been addressed by @chjinche.

StrikerRUS:
> It seems that the requested changes have been addressed, but they still block the merging.

Sorry about that!

PR #4787, opened 2021-11-09
[python] Faster categorical column names selection

Neronuser:
* Faster categorical column names selection

Change slow and redundant dataframe query by select_dtypes into a dataframe.dtypes list comprehension

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -16,6 +16,7 @@ from typing import Any, Callable, Dict, Iterable, List, Optional, Set, Tuple, Un
 
 import numpy as np
 import scipy.sparse
+from pandas import CategoricalDtype
 
 from .compat import PANDAS_INSTALLED, concat, dt_DataTable, is_dtype_sparse, pd_DataFrame, pd_Series
 from .libpath import find_lib_path
@@ -566,7 +567,7 @@ def _data_from_pandas(data, feature_name, categorical_feature, pandas_categorica
             raise ValueError('Input data must be 2 dimensional and non empty.')
         if feature_name == 'auto' or feature_name is None:
             data = data.rename(columns=str)
-        cat_cols = list(data.select_dtypes(include=['category']).columns)
+        cat_cols = [col for col, dtype in zip(data.columns, data.dtypes) if isinstance(dtype, CategoricalDtype)]
         cat_cols_not_ordered = [col for col in cat_cols if not data[col].cat.ordered]
         if pandas_categorical is None:  # train dataset
             pandas_categorical = [list(data[col].cat.categories) for col in cat_cols]
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -16,9 +16,8 @@ from typing import Any, Callable, Dict, Iterable, List, Optional, Set, Tuple, Un
 
 import numpy as np
 import scipy.sparse
-from pandas import CategoricalDtype
 
-from .compat import PANDAS_INSTALLED, concat, dt_DataTable, is_dtype_sparse, pd_DataFrame, pd_Series
+from .compat import PANDAS_INSTALLED, concat, dt_DataTable, is_dtype_sparse, pd_CategoricalDtype, pd_DataFrame, pd_Series
 from .libpath import find_lib_path
 
 ZERO_THRESHOLD = 1e-35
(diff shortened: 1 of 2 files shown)


python-package/lightgbm/basic.py
```diff
@@ -16,6 +16,7 @@
 
 import numpy as np
 import scipy.sparse
+from pandas import CategoricalDtype
```
jmoralez, line 19:
```suggestion
```

python-package/lightgbm/basic.py
```diff
@@ -16,6 +16,7 @@
...
 
 from .compat import PANDAS_INSTALLED, concat, dt_DataTable, is_dtype_sparse, pd_DataFrame, pd_Series
```
jmoralez, line 21:
```suggestion
from .compat import PANDAS_INSTALLED, concat, dt_DataTable, is_categorical_dtype, is_dtype_sparse, pd_DataFrame, pd_Series
```

python-package/lightgbm/basic.py
```diff
@@ -566,7 +567,7 @@ def _data_from_pandas(data, feature_name, categorical_feature, pandas_categorica
             raise ValueError('Input data must be 2 dimensional and non empty.')
         if feature_name == 'auto' or feature_name is None:
             data = data.rename(columns=str)
-        cat_cols = list(data.select_dtypes(include=['category']).columns)
+        cat_cols = [col for col, dtype in zip(data.columns, data.dtypes) if isinstance(dtype, CategoricalDtype)]
```
jmoralez, line 570:
```suggestion
        cat_cols = [col for col, dtype in zip(data.columns, data.dtypes) if is_categorical_dtype(dtype)]
```

jmoralez (changes requested):
Thank you for your contribution @Neronuser! Since pandas isn't a hard dependency we use a `compat` module to import things from it. I've suggested the changes to the `basic.py` file and you'd have to add the required import in https://github.com/microsoft/LightGBM/blob/b1facf50502fb51a1e60c05b3ec83f68289df497/python-package/lightgbm/compat.py#L9
to add `is_categorical_dtype` and define it as `None` after https://github.com/microsoft/LightGBM/blob/b1facf50502fb51a1e60c05b3ec83f68289df497/python-package/lightgbm/compat.py#L27

python-package/lightgbm/basic.py
Neronuser: Thanks @jmoralez! I updated the PR with required changes, but instead of importing is_categorical_dtype from pandas I create a Dummy CategoricalDtype object. This is done because is_categorical_dtype also checks if arrays have categorical dtype which introduces minor overhead for this case. Is that OK, or do you insist on using is_categorical_dtype?

jmoralez (approved):

jmoralez:
Thank you @Neronuser, looks good to me! Gently ping @StrikerRUS for a review as well.

StrikerRUS (approved):
LGTM! Thanks for this improvement!
I also checked that `CategoricalDtype` has been available to be imported in a such way at least since pandas version `0.20` which was released in 2017. So, I believe we are good with backward compatibility.
https://github.com/pandas-dev/pandas/blob/0.20.x/pandas/api/types/__init__.py#L4

StrikerRUS (changes requested):
Sorry, I was wrong in my previous comment. You can import `CategoricalDtype` directly from the root only since version `0.24`. For previous versions you should specify the full path:
```
from pandas.api.types import CategoricalDtype
```

python-package/lightgbm/compat.py
```diff
@@ -6,6 +6,7 @@
     from pandas import DataFrame as pd_DataFrame
     from pandas import Series as pd_Series
     from pandas import concat
+    from pandas.api.types import CategoricalDtype as pd_CategoricalDtype
```
StrikerRUS, line 9: This is fine that this import works with the latest version. But will it make sense to try import like `from pandas import CategoricalDtype as pd_CategoricalDtype` in case they'll change internal structure of modules?
Refer to
https://github.com/microsoft/LightGBM/blob/99e0a4bd7b9e7c557e593ff9172799822abc4b7d/python-package/lightgbm/compat.py#L68-L73
WDYT?

Neronuser: Yes, this makes sense, thank you. Not sure, but it feels more likely that they are going to change `pandas.api.types` than the top-level import of their types `from pandas import CategoricalDtype`. Especially, given that their current top-level init goes into `pandas.core.api` for CategoricalDtype.

StrikerRUS (approved):
LGTM, thank you very much!

boris-saivahc:
I'd like to ask about the reason behind the merge of this change without subsequent release. I am currently dealing with a large dataset that consists of multiple categorical features. However, the implementation in version 3.3.5 results in an unnecessary increase in memory usage. It would greatly benefit me to have this change included in the released version.

jameslamb:
Subscribe to #5153 to be notified of the next release.

There's nothing specific to this change keeping it out of releases...in general we have some challenges with maintainer availability in this project that have led to such a long delay between releases. We're trying to get a release out in the next few months, sorry for the inconvenience.

Issue #4602, opened 2021-09-09
[ci] Run 'cmakelint' in CI


jameslamb:
## Summary

LightGBM's continuous integration (CI) should use [`cmakelint`](https://github.com/cmake-lint/cmake-lint) to enforce consistency in the formatting of `CMakeLists.txt`.

## Motivation

Following this tool's recommendations should make `CMakeLists.txt` a bit easier to read and maintain, by enforcing consistency in formatting.

You can run `cmakelint --filter` to see a list of the supported checks.

```text
convention/filename
linelength
package/consistency
readability/logic
readability/mixedcase
readability/wonkycase
syntax
whitespace/eol
whitespace/extra
whitespace/indent
whitespace/mismatch
whitespace/newline
whitespace/tabs
```

## Description

Similar to the approach taken in LightGBM with `cpplint` (#1990) and `mypy` (#3867), for a time these checks might be run in continuous integration without blocking merges, just as extra information about the state of the code.

## How to Contribute

To contribute to this issue, run the following commands

```shell
pip install --upgrade cmakelint
cmakelint --linelength=120
```

Make a pull request that addresses some of the warnings. In the pull request description, list the specific warnings your changes address.

Small, focused pull requests are appreciate. Do not try to fix all of the warnings in a single PR.

## References

* https://github.com/cmake-lint/cmake-lint
* this issue was created from the discussion in https://github.com/microsoft/LightGBM/pull/4593
* those interested in improving static analysis of C/C++ code in the project might also explore #4539

PR #4794, opened 2021-11-13
[ci] fix CMakeLint errors related to function naming case

StrikerRUS:
Contributes to #4602.

Use lower case for all functions to fix all `readability/mixedcase` errors.

There are false alerts for `readability/wonkycas` in some built-in CMake functions, e.g.
https://github.com/microsoft/LightGBM/blob/6cbb3586b92112fc1f27a179b201ac924e5600cc/CMakeLists.txt#L501
https://github.com/microsoft/LightGBM/blob/6cbb3586b92112fc1f27a179b201ac924e5600cc/CMakeLists.txt#L506
https://cmake.org/cmake/help/latest/module/FetchContent.html

So the option only is to not check `readability/wonkycase` type of errors. Because there is no `NOLINT` directive in the `cmakelint` package like it is in the `cpplint` package 🙁 .


jameslamb (changes requested):
> There are false alerts for readability/wonkycas in some built-in CMake 

@StrikerRUS I don't think we need to filter out the `readability/wonkycase` warnings.

I think the mixed-case commands you mentioned can be lower-cased. I tried that tonight and was able to run the tests successfully.

```shell
# lowercase commands
sed -i.bak -e "s/FetchContent_Declare/fetchcontent_declare/" CMakeLists.txt
sed -i.bak -e "s/FetchContent_MakeAvailable/fetchcontent_makeavailable/" CMakeLists.txt

# run tests
rm -rf build/
mkdir build/
cmake -DBUILD_CPP_TEST=ON -DUSE_OPENMP=OFF -DUSE_DEBUG=ON ..
make testlightgbm -j4
./../testlightgbm
```

That's consistent with https://cmake.org/cmake/help/latest/manual/cmake-language.7.html#command-invocations, which says

> Command names are case-insensitive

StrikerRUS:
Hmmm, thanks for clarifying that those commands **can** be lower-cased. But in the official docs they **are** mixed-cased. Don't you think we should stick to the official (which I personally consider as a recommended) CMake naming style but not to `cmakelint` one?

jameslamb:
> Don't you think we should stick to the official (which I personally consider as a recommended) CMake naming style but not to cmakelint one

Personally, I'd prefer to use the lower-cased one so `cmakelint` can continue to do the review work for us of telling contributors to lowercase commands.

Otherwise, we'll have to write comments like "can you please change `OPTION()` to `option()` for consistency", "can you please change `ADD_DEFINITIONS()` to `add_definitions()`", etc.

StrikerRUS:
> Otherwise, we'll have to write comments like "can you please change `OPTION()` to `option()` for consistency", "can you please change `ADD_DEFINITIONS()` to `add_definitions()`", etc.

These style issues will produce `readability/mixedcase` error. So we won't have say anything.
`cmakelint` will stop catching something like `OptIoN()` or `add_DEFINITIONS()`. Do you think this is a common problem we'll face?

jameslamb:
ohhhhh I understand, yeah ok I am ok with filtering out `readability/wonkycase` then.  Thanks for talking through it with me.

jameslamb (approved):

jameslamb:
Nice work! Just 3 warnings from `cmakelint` remaining after this!

https://github.com/microsoft/LightGBM/runs/4201144410?check_suite_focus=true

PR #4796, opened 2021-11-13
[ci] fix CMakeLint `linelength` errors

StrikerRUS:
Contributes to #4602.

Why we can't use normal indents for long strings.

> Problem with the CMake 3.0 approach is that it doesn't ignore indentation. Which means multi-line strings can't be indented with the rest of the code. 
   https://stackoverflow.com/questions/7637539/how-to-split-strings-across-multiple-lines-in-cmake#comment70641940_27983206

For example:

image

image


jameslamb (approved):
very nice, thanks! And thanks for the explanation about long strings, I definitely would have asked about that.

PR #4798, opened 2021-11-13
[ci] use pure `endif()` and `endfunction()` without expression inside in `CMakeLists.txt`

StrikerRUS:
Contributes to #4602.

Fix all
- `Expression repeated inside endif; better to use only endif() [readability/logic]`
- `Expression repeated inside endfunction; better to use only endfunction() [readability/logic]`

errors.


jameslamb (approved):
Checked the `lint` job logs and confirmed this eliminates all the cases of these warnings, thanks!

PR #4799, opened 2021-11-13
[ci] simplify processing of `CMAKE_CUDA_FLAGS` variable in `CMakeLists.txt`

StrikerRUS:
Work with `CMAKE_CUDA_FLAGS` as with a string, do not mix lists and strings in one variable.


jameslamb (approved):
Works for me, and CUDA CI jobs seem happy with this change. Thanks!

PR #4800, opened 2021-11-14
[python] Remove `silent` argument

StrikerRUS:
Refer to #4577.


jameslamb (approved):
Thanks! Glad we're starting to remove some of these deprecation warnings and simplifying things.

PR #4802, opened 2021-11-14
WIP: Load back parameters from saved model file (fixes #2613)

### LINKED ISSUE #2613: Load back saved parameters with save_model to Booster object

zyxue:
This is still just an idea to collect feedbacks.

The approach in this PR:

1. made `loaded_parameter_` an attribute of `Boosting` instead of `GBDT` so it can be accessed in `c_api.cpp`.
1. implemented `LGBM_BoosterGetConfig` to return `loaded_parameter_` as a string.
1. `loaded_parameter_` can be parsed in Python code into proper types for the `params` dictionary.

test `test_booster_load_params_when_passed_model_str` passes

This PR is related to https://github.com/microsoft/LightGBM/issues/2613

A couple of questions:

* why isn't `loaded_parameter_` a parameter of `Boosting` instead of `GBDT` already? The other boosting types don't have parameters?
* Do I understand correctly that `c_api.cpp` can only deal with `gbdt` as I see it's hard-coded in https://github.com/microsoft/LightGBM/blob/874e6359241594a7fdfef59924532361c236cf7f/src/c_api.cpp#L109

### DIFF
--- a/src/c_api.cpp
+++ b/src/c_api.cpp
@@ -1627,14 +1627,24 @@ int LGBM_BoosterGetConfig(
     while(std::getline(input_stream, line, '\n')){
       if (line.rfind("[boosting: ", 0) == 0) {
         obj["boosting"] = Json{extract_param("boosting", line)};
+      } else if (line.rfind("[boost_from_average: ", 0) == 0) {
+        obj["boost_from_average"] = Json{std::stod(extract_param("boost_from_average", line)) == 1 ? true : false};
+      } else if (line.rfind("[learning_rate: ", 0) == 0) {
+        obj["learning_rate"] = Json{std::stod(extract_param("learning_rate", line))};
+      } else if (line.rfind("[metric: ", 0) == 0) {
+        obj["metric"] = Json{extract_param("metric", line)};
+      } else if (line.rfind("[monotone_constraints: ", 0) == 0) {
+        obj["monotone_constraints"] = Json{LightGBM::Common::StringToArray<int8_t>(extract_param("monotone_constraints", line), ',')};
       } else if (line.rfind("[num_iterations: ", 0) == 0) {
         obj["num_iterations"] = Json{std::stoi(extract_param("num_iterations", line))};
       } else if (line.rfind("[num_leaves: ", 0) == 0) {
         obj["num_leaves"] = Json{std::stoi(extract_param("num_leaves", line))};
-      } else if (line.rfind("[learning_rate: ", 0) == 0) {
-        obj["learning_rate"] = Json{std::stod(extract_param("learning_rate", line))};
+      } else if (line.rfind("[objective: ", 0) == 0) {
+        obj["objective"] = Json{extract_param("objective", line)};
       } else if (line.rfind("[tree_learner: ", 0) == 0) {
         obj["tree_learner"] = Json{extract_param("tree_learner", line)};
+      } else if (line.rfind("[verbose: ", 0) == 0) {
+        obj["verbose"] = Json{std::stoi(extract_param("verbose", line))};
       }
       // TODO: more params to be extracted
     }
--- a/src/c_api.cpp
+++ b/src/c_api.cpp
@@ -13,6 +13,7 @@
 #include <LightGBM/objective_function.h>
 #include <LightGBM/prediction_early_stop.h>
 #include <LightGBM/utils/common.h>
+#include <LightGBM/utils/json11.h>
 #include <LightGBM/utils/log.h>
 #include <LightGBM/utils/openmp_wrapper.h>
 #include <LightGBM/utils/random.h>
@@ -29,6 +30,8 @@
 #include "application/predictor.hpp"
 #include <LightGBM/utils/yamc/alternate_shared_mutex.hpp>
 #include <LightGBM/utils/yamc/yamc_shared_lock.hpp>
+
+using json11::Json;
 
 namespace LightGBM {
 
@@ -1594,6 +1597,16 @@ int LGBM_BoosterLoadModelFromString(
   API_END();
 }
 
+// extracts the value for key in a parameter line. E.g. for the key boosting and
+// a line like `[boosting: gbdt]` (without newline character), the extracted value would be "gbdt".
+std::string extract_param(std::string key, std::string line) {
+  std::string prefix = "[" + key + ": ";
+  std::string suffix = "]";
+  std::string prefix_replaced = line.replace(line.begin(), line.begin() + prefix.length(), "");
+  std::string suffix_replaced = prefix_replaced.replace(prefix_replaced.end() - suffix.length(), prefix_replaced.end(), "");
+  return suffix_replaced;
+}
+
 int LGBM_BoosterGetConfig(
   BoosterHandle handle,
   int64_t buffer_len,
@@ -1602,12 +1615,37 @@ int LGBM_BoosterGetConfig(
 ) {
   API_BEGIN();
   Booster* ref_booster = reinterpret_cast<Booster*>(handle);
-  // GBDT* ref_gbdt = reinterpret_cast<GBDT*>(ref_booster);
   std::string params = ref_booster->GetBoosting()->loaded_parameter_;
-  // std::string params = "abc";
-  *out_len = static_cast<int64_t>(params.size()) + 1;
+
+  // Converts parameters to Json
+  Json::object obj = Json::object{};
+  std::stringstream input_stream(params);
+  std::string line;
+
+  if (out_str != NULL)
+  {
+    while(std::getline(input_stream, line, '\n')){
+      if (line.rfind("[boosting: ", 0) == 0) {
+        obj["boosting"] = Json{extract_param("boosting", line)};
+      } else if (line.rfind("[num_iterations: ", 0) == 0) {
+        obj["num_iterations"] = Json{std::stoi(extract_param("num_iterations", line))};
+      } else if (line.rfind("[num_leaves: ", 0) == 0) {
+        obj["num_leaves"] = Json{std::stoi(extract_param("num_leaves", line))};
+      } else if (line.rfind("[learning_rate: ", 0) == 0) {
+        obj["learning_rate"] = Json{std::stod(extract_param("learning_rate", line))};
+      } else if (line.rfind("[tree_learner: ", 0) == 0) {
+        obj["tree_learner"] = Json{extract_param("tree_learner", line)};
+      }
+      // TODO: more params to be extracted
+    }
+  }
+
+  Json params_json = Json{obj};
+  std::string params_json_str = params_json.dump();
+
+  *out_len = static_cast<int64_t>(params_json_str.size()) + 1;
   if (*out_len <= buffer_len) {
-    std::memcpy(out_str, params.c_str(), *out_len);
+    std::memcpy(out_str, params_json_str.c_str(), *out_len);
   }
   API_END();
 }
--- a/src/c_api.cpp
+++ b/src/c_api.cpp
@@ -1594,6 +1594,24 @@ int LGBM_BoosterLoadModelFromString(
   API_END();
 }
 
+int LGBM_BoosterGetConfig(
+  BoosterHandle handle,
+  int64_t buffer_len,
+  int64_t* out_len,
+  char* out_str
+) {
+  API_BEGIN();
+  Booster* ref_booster = reinterpret_cast<Booster*>(handle);
+  // GBDT* ref_gbdt = reinterpret_cast<GBDT*>(ref_booster);
+  std::string params = ref_booster->GetBoosting()->loaded_parameter_;
+  // std::string params = "abc";
+  *out_len = static_cast<int64_t>(params.size()) + 1;
+  if (*out_len <= buffer_len) {
+    std::memcpy(out_str, params.c_str(), *out_len);
+  }
+  API_END();
+}
+
 #ifdef _MSC_VER
   #pragma warning(disable : 4702)
 #endif
--- a/tests/python_package_test/test_basic.py
+++ b/tests/python_package_test/test_basic.py
@@ -648,7 +648,7 @@ def _fake_model() -> lgb.Booster:
 
     return gbm
 
-def test_booster__load_param_when_passed_model_file(fake_model: lgb.Booster) -> None:
+def test_booster__load_param_when_passed_model_file(fake_model):
     with tempfile.TemporaryDirectory() as temp_dir:
         model_file = Path(temp_dir) / "model.txt"
         fake_model.save_model(model_file)
@@ -658,7 +658,7 @@ def test_booster__load_param_when_passed_model_file(fake_model: lgb.Booster) -> 
     # TODO: needs parse more params
     assert loaded.params['boosting'] == 'gbdt'
 
-def test_booster__load_param_when_passed_model_str(fake_model: lgb.Booster) -> None:
+def test_booster__load_param_when_passed_model_str(fake_model):
     model_str = fake_model.model_to_string()
 
     loaded = lgb.Booster(model_str=model_str)
--- a/tests/python_package_test/test_basic.py
+++ b/tests/python_package_test/test_basic.py
@@ -612,55 +612,72 @@ def test_custom_objective_safety():
     with pytest.raises(ValueError, match=re.escape(f"number of models per one iteration ({nclass})")):
         bad_bst_multi.update(fobj=_bad_gradients)
 
+
+@pytest.fixture(name="fake_input_params")
+def _fake_input_params():
+    return {
+        "boosting": "gbdt",
+        "boost_from_average": True,
+        "learning_rate": 0.05,
+        "metric": "binary_logloss",
+        "monotone_constraints": [-1, 0, 1],
+        "num_iterations": 5,
+        "num_leaves": 31,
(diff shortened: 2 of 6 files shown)


jameslamb (changes requested):
Hi @zyxue , sorry for the very long delay without a review! We really appreciate you working on this. I'm ready to review this and help move it forward.

1. Please remove all commented-out code.
2. Please merge in the latest changes from `master`
2. Please try to get this to a state where at least one test in the Python package is passing for one parameter...then we can iterate from there based on review comments.

--------------------------

At first glance, I have one suggestion about a change to the approach...instead of passing the raw content of the `parameters` block from a text format back through `LGBM_BoosterGetConfig()`, I think it would be better to have code on the C/C++ side parse that information and pass back a JSON string.

That way, every interface to LightGBM (for example, the Python and R packages in this repo) doesn't need to have some version of this code (with one line per parameter):

```python
for line in io.StringIO(_params_str):
    if line.startswith('[boosting: '):
        self.params['boosting'] = line.strip().replace(f"[boosting: ", "").replace("]", "")
```

and could instead just pass the result of `LGBM_BoosterGetConfig()` to a JSON parser like `json` in Python or `{jsonlite}` in R.

jameslamb:
@zyxue thanks for coming back to this pull request!

As you work on it, please do not rebase + force push. Use merge commits instead (e.g. `git merge master`). Overwriting the commit history makes it more difficult for reviewers to understand the changes you've made in response to review comments.

This project squashes all pull request commits into a single commit on merge, so you don't need to worry about having too many commits here.

zyxue:
@jameslamb , gotcha, thank you for the tip.

zyxue:
Is https://github.com/microsoft/LightGBM/blob/master/include/LightGBM/utils/json11.h expected to be the same as https://raw.githubusercontent.com/dropbox/json11/master/json11.hpp ?

jameslamb:
> Is https://github.com/microsoft/LightGBM/blob/master/include/LightGBM/utils/json11.h expected to be the same as https://raw.githubusercontent.com/dropbox/json11/master/json11.hpp ?

That is where the `json11` code in LightGBM is originally from, but we consider it "vendored in", meaning that since it was first added to this project, LightGBM-specific modifications have sometimes been made to it.

zyxue:
@jameslamb , I've reimplemented `LGBM_BoosterGetConfig` to return parameters as a json string, please let me know what you think.

tests/python_package_test/test_basic.py
```diff
@@ -609,3 +611,57 @@ def test_custom_objective_safety():
...
+        lgb_train,
+        num_boost_round=3,
+        valid_sets=lgb_train,  # eval training data
+        feature_name=feature_name,
+        categorical_feature=[21],
+    )
+
+    return gbm
+
+def test_booster_load_params_when_passed_model_file(fake_model: lgb.Booster) -> None:
```
jameslamb, line 651:
```suggestion
def test_booster_load_params_when_passed_model_file(fake_model):
```

Thank you very much for adding type hints! But they are not necessary for test functions (anything in a `tests/` files that starts `def test_*`).

Please remove these, for consistency with the rest of the tests.

tests/python_package_test/test_basic.py
```diff
@@ -609,3 +611,57 @@ def test_custom_objective_safety():
...
+    with tempfile.TemporaryDirectory() as temp_dir:
+        model_file = Path(temp_dir) / "model.txt"
+        fake_model.save_model(model_file)
+
+        loaded = lgb.Booster(model_file=model_file)
+
+    # TODO: needs parse more params
+    assert loaded.params['boosting'] == 'gbdt'
```
jameslamb, line 659: As you continue working on this feature, please consider the following additional parameters as a minimal sample that should help catch some different types of parsing issues.

* `monotone_constraints`: an array of ints
    - https://github.com/microsoft/LightGBM/blob/a06fadfb7ac3fdb26da3d4afc061a8b976070c50/src/io/config_auto.cpp#L448
* `learning_rate`: a double
    - https://github.com/microsoft/LightGBM/blob/a06fadfb7ac3fdb26da3d4afc061a8b976070c50/src/io/config_auto.cpp#L332
* `boost_from_average`: a boolean
    - https://github.com/microsoft/LightGBM/blob/a06fadfb7ac3fdb26da3d4afc061a8b976070c50/src/io/config_auto.cpp#L583

tests/python_package_test/test_basic.py
```diff
@@ -609,3 +611,57 @@ def test_custom_objective_safety():
...
     with pytest.raises(ValueError, match=re.escape(f"number of models per one iteration ({nclass})")):
         bad_bst_multi.update(fobj=_bad_gradients)
+
+@pytest.fixture(name="fake_model")
+def _fake_model() -> lgb.Booster:
+    # TODO: maybe removed deps on data_dir later
+    data_dir = Path(__file__).parent.parent.parent / "examples/binary_classification"
+
+    df_train = pd.read_csv(data_dir / "binary.train", header=None, sep="\t")
+    weights = pd.read_csv(data_dir / "binary.train.weight", header=None)[0]
```
jameslamb, line 621: whether or not training uses weights shouldn't change the behavior of parsing parameters from the model text format, right? I think this could be removed, to simplify the tests a bit.

src/c_api.cpp
```diff
@@ -1594,6 +1597,59 @@ int LGBM_BoosterLoadModelFromString(
...
+}
+
+int LGBM_BoosterGetConfig(
+  BoosterHandle handle,
+  int64_t buffer_len,
+  int64_t* out_len,
+  char* out_str
+) {
+  API_BEGIN();
+  Booster* ref_booster = reinterpret_cast<Booster*>(handle);
```
jameslamb, line 1617: I'd like to propose a different organization of this code, but please don't make any changes to the organization until another maintainer like @StrikerRUS, @shiyu1994, or @tongwu-msft  comments.

Instead of all this logic being in `src/c_api.cpp`, I think most of the implementation for `LGBM_BoosterGetConfig()` should live somewhere else. Similar to how `LGBM_BoosterSaveModelToString()` is defined here in `src/c_api.cpp`.

https://github.com/microsoft/LightGBM/blob/cf38071b6ac7e238b1654286dbb5a9a98b2c320b/src/c_api.cpp#L2233

And then it is just a thin wrapper on `GBDT::SaveModelToString()`.

https://github.com/microsoft/LightGBM/blob/a06fadfb7ac3fdb26da3d4afc061a8b976070c50/src/boosting/gbdt_model_text.cpp#L410

I also think it's desirable for any code that needs to list every parameter by name should be in the automatically-generated methods in `src/config_auto.cpp`.

So I think this should be approached as following:

1. Add a method like `GBDT::GetLoadedConfig()` which reformats `loaded_parameter_` into the appropriate format a parameters string and then initializes a `Config` object by calling `Config::GetMembersFromString()`
    - https://github.com/microsoft/LightGBM/blob/cf38071b6ac7e238b1654286dbb5a9a98b2c320b/src/io/config_auto.cpp#L321
2. Add a method in `src/io/config_auto.cpp` like `Config::DumpConfigToJson()`, which returns a JSON representation of a `Config` object.
    - NOTE: all the code there is generated by https://github.com/microsoft/LightGBM/blob/a06fadfb7ac3fdb26da3d4afc061a8b976070c50/helpers/parameter_generator.py. But if I was working on this, my approach would be to just write code directly in `src/io/config_auto.cpp`, get the tests into a good state, and *then* make `parameter_generator.py` write that code.
    - this will be useful if LightGBM adopts the proposal at #4887 in the future
3. Alter `LGBM_BoosterGetConfig()` to get the output of `Config::DumpConfigToJson()`.

--------------

Sorry if this is overly complicated. Let's see what other maintainers say. And we are here to help! Thanks again for your help with this feature.

jameslamb (changes requested):
Thanks very much! I left some suggestions for a reorganization of the code on the C/C++ side. But let's see what other maintainers say.

tests/python_package_test/test_basic.py
zyxue: Why is test necessary in code but not in test?

zyxue: Thanks for the detailed explanation! My thought is just that with type hints, the code is much easier to understand in general. But I also prefer not to deviate from PR. I'll leave type hints out of test for now.

zyxue:
>please don't make any changes to the organization until another maintainer like @StrikerRUS, @shiyu1994, or @tongwu-msft comments.
Hey @jameslamb , should I update now or wait till more comments come in?

jameslamb:
As I said, please wait until another maintainer offers their opinion. While you wait on that, you can try working through the other suggestions like https://github.com/microsoft/LightGBM/pull/4802#discussion_r786419637.

src/c_api.cpp
```diff
@@ -1594,6 +1597,69 @@ int LGBM_BoosterLoadModelFromString(
...
+      } else if (line.rfind("[num_iterations: ", 0) == 0) {
+        obj["num_iterations"] = Json{std::stoi(extract_param("num_iterations", line))};
+      } else if (line.rfind("[num_leaves: ", 0) == 0) {
+        obj["num_leaves"] = Json{std::stoi(extract_param("num_leaves", line))};
+      } else if (line.rfind("[objective: ", 0) == 0) {
+        obj["objective"] = Json{extract_param("objective", line)};
+      } else if (line.rfind("[tree_learner: ", 0) == 0) {
+        obj["tree_learner"] = Json{extract_param("tree_learner", line)};
+      } else if (line.rfind("[verbose: ", 0) == 0) {
+        obj["verbose"] = Json{std::stoi(extract_param("verbose", line))};
```
guolinke, line 1647: I think we can maintain a Key-value map, and load parameters in a for-loop, rather than use many if-elsees.

guolinke:
@shiyu1994 can you take a look for this PR?

src/c_api.cpp
johnpaulett: Per #5244, I would find extracting the `categorical_feature` param helpful as well.

src/c_api.cpp
zyxue: >reformats `loaded_parameter_` into the appropriate format a parameters string

@jameslamb , do you actually mean reformatting `loaded_parameter_` into a `std::unordered_map<std::string, std::string>` object? 

I'm a bit confused by the name `GetMembersFromString`, it actually means GetMembersFrom an unordered_map, right?

jameslamb: I mean doing whatever is necessary to turn `loaded_parameter_` into a `Config` object, yes. You're right that the method `Config::GetMembersFromString()` takes an unordered map with string keys and string values.

jameslamb:
Due to a lack of activity on this pull request, we've decided to move forward with a separate PR for this feature in #5424, so I'm going to close this.

@zyxue thanks very much for your interest in LightGBM and for attempting this! If you have more time to work with us in the future, we'd welcome additional contributions. I'd be happy to suggest some smaller contributions which might not involve as much discussion and require as much of your time and energy.

src/c_api.cpp
jameslamb: > would find extracting the categorical_feature param helpful as well

this is now being addressed in #5424 . @johnpaulett if you're interested in influencing the shape that this support takes in LightGBM, you're welcome to leave review comments on that PR.

PR #4807, opened 2021-11-17
Load initial scores when use binary data files in CLI version

### LINKED ISSUE #4786: Initializing both the models and bin data file retrains the model, instead of continuing training.
## Description
I want to run distributed CLI training and I want to run the training twice: the second time initializing the model from the first run. 

So what I do is I use parameter `init_model` the second time. 

What I found is that depending on whether I use another parameter `is_save_binary_file` the results are different. In particular, **if I use `is_save_binary_file=False`, then I get the continual training, that I expect.** 

Here I have loss curves for two runs, 100 iterations each. So you can see the line is smooth, meaning that the second run picked up the first model. 

image

But, **if I use `is_save_binary_file=True`, then the second run retrains the model.**

image

I would expect that binary files should not affect initialization of the model. In particular, I would expect `is_save_binary_file=True` just creates convenient bin mappers for the future trees, that are independent of each run I use them for.

Is it a bug or expected behavior? How would I go about using both parameters `init_model` and `is_save_binary_file`? 

It would be useful to use binary files, as they are very quick to read, and being able to initialize the model as well. Is it difficult to do? Can you give some pointers where this code logic is? 

## Reproducible example
This example gives the setting for distributed CLI version. Just try to train the second run with and without bin files. 
In particular, 

Setting A. `is_save_binary_file=False`

Setting B. `is_save_binary_file=True`

In both settings, you need to have 2 runs, the first one that saves the model `output_model` and the second one that loads the model `input_model`.

## Environment info

LightGBM version or commit hash: 3.2.1

```shell
lightgbm config=train.conf
```

shiyu1994:
This is to fix #4786. Currently there's no way to load initial score when loading from binary data file in CLI version.

In `Metadata::LoadFromMemory` and `Metadata::SaveBinaryToFile`, other fields of `Metadata` is handled, except `init_score_`. The only way to set initial score when loading from binary data files is the call methods like `set_init_score` in Python, which is impossible for CLI version.

The implementation goes like this: after loading from binary data files, checks whether there are initial score files exists. If so, initial scores are loaded.

The corresponding documentation for continual training is also updated.

### DIFF
--- a/docs/Parameters.rst
+++ b/docs/Parameters.rst
@@ -1259,6 +1259,8 @@ The initial score file corresponds with data file line by line, and has per scor
 And if the name of data file is ``train.txt``, the initial score file should be named as ``train.txt.init`` and placed in the same folder as the data file.
 In this case, LightGBM will auto load initial score file if it exists.
 
+If binary data files exist for raw data file ``train.txt``, for example in the name ``train.txt.bin``, then the initial score file should be named as ``train.txt.bin.init``.
+
 Weight Data
 ~~~~~~~~~~~
 
--- a/include/LightGBM/dataset.h
+++ b/include/LightGBM/dataset.h
@@ -144,6 +144,9 @@ class Metadata {
     queries_[idx] = static_cast<data_size_t>(value);
   }
 
+  /*! \brief Load initial scores from file */
+  void LoadInitialScore(const std::string& data_filename);
+
   /*!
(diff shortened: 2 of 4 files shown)


StrikerRUS:
@guolinke Could you please help to review this PR?

guolinke (approved):

Issue #4804, opened 2021-11-15
Can LightGBM support more than max int32_t data points?


thvasilo:
## Summary

Hi I was wondering if all it would take for LightGBM to support more than the ~2B data points int32_t allows for, would be doing the work required to change `data_size_t` to int64 as proposed in https://github.com/microsoft/LightGBM/issues/2818 ?

Seems like that work has been dropped for now, but I guess it's open to be picked up?

## Motivation

For large-scale learning tasks we might want to use more than 2B data points.

## Description

The work would include refactoring the codebase to support data IDs that are int64-based.

## References

Seems like @guolinke  branch at https://github.com/microsoft/LightGBM/tree/sparse_bin_int64 is no longer available as a starting point?

shiyu1994:
According to the discussion in #2818, changing `data_size_t` from `int32_t` to `int64_t` seems enough (without considering efficiency). But I'm not sure whether there are some conversions between `int32_t` and `data_size_t` in repo now. If so, something unexpected may happen.

Contribution is welcome! I can help to check with the existence of the conversion. I'm personally also very interested in this.

thvasilo:
Well getting the code to compile and tests to pass after changing all refs to `data_size_t` to use `int64_t` explicitly wasn't much work.

I'm kinda worried because lots of times I saw `data_size_t` used interchangeably with `int` or `int32_t` in terms of returns types etc, and I don't have enough depth to know which references to `data_size_t` should be converted to `int64_t` and which should be converted to `int32_t` to avoid large memory costs. To work through quickly I replaced all references to `data_size_t` with `int64_t`, which means that wherever that was used with the assumption that it refers to `int32_t` it will likely incur a performance/memory penalty.

With some help from the devs though I could give this a shot, starting with a WIP PR and working on some performance regression tests.

shiyu1994:
Feel free to open a PR when you have time. Thanks!

Issue #2818, opened 2020-02-24
[feature request] Support int64_t data_size_t


guolinke:
As the `cnt` in the histogram is removed, the overhead of `int64_t` as `data_size_t` will be smaller.

There are several todo items:

- [ ] remove `data_size_t` completely in code, directly used `int64_t`
- [ ] Template types for the `indices` in `DataPartition`, to achieve the same performance as before when the data_size_t is smaller than int32_t. 
- [ ] update related c_api, and Python / R interfaces

guolinke:
I have a branch (https://github.com/microsoft/LightGBM/tree/sparse_bin_int64) working on this, will revisit it when have time.

Adam1679:
hey, can i help this one?

guolinke:
@Adam1679 yeah, very welcome

Adam1679:
sorry, this is my first time to make public contribution. Any suggestions about how to begin this one? Should I fork your branch and get familiar with the code and your modifications?

guolinke:
yes, you can fork the code and code in your cloned repo first, and create a PR here when it is almost ready.
If there are any problems/questions, you can ask me directly in this issue.
BTW, it is better to start from `master`, not the `sparse_bin_int64`, as there are many new changes in master branch.

Adam1679:
ok, i will try my best.

Adam1679:
hey, I read your commit "5a5b26d" of your branch and I saw that you changed data_size_t to from int32_t to int64_t directly. And also you changed some "int" to "data_size_t" and some "data_size_t" to "int". I understand that you cast some "data_size_t" to "int" since now "data_size_t" is bigger. But why do you change some "int" to "data_size_t"? I don't understand this part. Also, it would be highly appreciated if you can elaborate more on the "the overhead of int64_t as data_size_t will be smaller." Is it because that "int32_t" is not big enough in some cases but you have to compromise to use it since the overhead of using int64 in the `cnt` related part is unbearable?

guolinke:
@Adam1679 some of `int` should be `data_size_t`, so i fix them.

For the overheads:
1) the `cnt` in histogram, this is removed. So you don't need to care it any more.
[...]

PR #4809, opened 2021-11-18
[WIP] Add support for more than max(int_32_t) data points

thvasilo:
Aimed to address https://github.com/microsoft/LightGBM/issues/4804 and https://github.com/microsoft/LightGBM/issues/2818

This is a WIP PR to start work on allowing more than the ~2B data points that `int_32_t` allows for.

For now I've simply replaced all references to `data_size_t` with `int_64_t` and made some post-hoc fixes in certain places replacing `int64_t` with `int32_t` where it was needed for the project to compile and tests to run.

Currently I see 20/20 tests passing but I also see this output when running them:

>[LightGBM] [Fatal] ChunkedArray chunk size must be larger than 0!
>[LightGBM] [Fatal] no conversion to double for: x1

Due to the prevalence of `data_size_t` it might be a better strategy to selectively update certain files/classes to move to `int64_t` to make reviewing easier and the changes more gradual (splitting up this PR into multiple ones). To do that I would need some guidance from the core devs as to which references to `data_size_t` are the best candidates to change to enable the support for `int_64_t`  data points.

Another area where devs could help is suggesting stress tests and how to measure performance/memory cost impact, perhaps running benchmarks or other performance checks.

### DIFF
--- a/include/LightGBM/meta.h
+++ b/include/LightGBM/meta.h
@@ -25,7 +25,7 @@
 namespace LightGBM {
 
 /*! \brief Type of data size, it is better to use signed type*/
-typedef int32_t data_size_t;
+typedef int64_t data_size_t;
 
 // Enable following macro to use double for score_t
 // #define SCORE_T_USE_DOUBLE
--- a/src/c_api.cpp
+++ b/src/c_api.cpp
@@ -1384,11 +1384,11 @@ int LGBM_DatasetCreateFromCSC(const void* col_ptr,
 }
 
 int LGBM_DatasetGetSubset(
-  const DatasetHandle handle,
-  const int32_t* used_row_indices,
-  int32_t num_used_row_indices,
-  const char* parameters,
-  DatasetHandle* out) {
+        const DatasetHandle handle,
+        const data_size_t *used_row_indices,
+        int32_t num_used_row_indices,
+        const char* parameters,
+        DatasetHandle* out) {
   API_BEGIN();
   auto param = Config::Str2Map(parameters);
   Config config;
@@ -1398,7 +1398,7 @@ int LGBM_DatasetGetSubset(
   CHECK_GT(num_used_row_indices, 0);
   const int32_t lower = 0;
   const int32_t upper = full_dataset->num_data() - 1;
-  CheckElementsIntervalClosed(used_row_indices, lower, upper, num_used_row_indices, "Used indices of subset");
+  CheckElementsIntervalClosed<data_size_t>(used_row_indices, lower, upper, num_used_row_indices, "Used indices of subset");
   if (!std::is_sorted(used_row_indices, used_row_indices + num_used_row_indices)) {
     Log::Fatal("used_row_indices should be sorted in Subset");
   }
@@ -1481,7 +1481,7 @@ int LGBM_DatasetSetField(DatasetHandle handle,
   if (type == C_API_DTYPE_FLOAT32) {
     is_success = dataset->SetFloatField(field_name, reinterpret_cast<const float*>(field_data), static_cast<int32_t>(num_element));
   } else if (type == C_API_DTYPE_INT32) {
-    is_success = dataset->SetIntField(field_name, reinterpret_cast<const int*>(field_data), static_cast<int32_t>(num_element));
+    is_success = dataset->SetIntField(field_name, reinterpret_cast<const data_size_t*>(field_data), static_cast<int32_t>(num_element));
   } else if (type == C_API_DTYPE_FLOAT64) {
     is_success = dataset->SetDoubleField(field_name, reinterpret_cast<const double*>(field_data), static_cast<int32_t>(num_element));
   }
@@ -1491,7 +1491,7 @@ int LGBM_DatasetSetField(DatasetHandle handle,
 
 int LGBM_DatasetGetField(DatasetHandle handle,
                          const char* field_name,
-                         int* out_len,
+                         data_size_t* out_len,
                          const void** out_ptr,
                          int* out_type) {
   API_BEGIN();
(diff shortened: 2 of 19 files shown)


guolinke:
I think we should benchmark for the training speed regards to this change.

shiyu1994:
@thvasilo Thanks for opening this PR.
To replace `data_size_t` with `int64_t`, we only need to change the definition of `data_size_t` here https://github.com/microsoft/LightGBM/blob/06e3c4a9bb23e3ffdacf38a62373c53a2552f74b/include/LightGBM/meta.h#L28
There's no need to touch so many files.

@guolinke I can help to add templates for the `data_partition.cpp`, and check that all usages of `data_size_t` with `int64_t` is correct.

I think it would be better to make `int64_t` data size as an option, instead of the default setting. Do you think a compiler option or an hyperparameter is preferred?

shiyu1994:
It seems that it would be difficult to use a hyperparameter to enable `int64_t` without large changes.

guolinke:
I think compile option is much easier. 
Another solution is using template, and auto choose data_size_t according to #data .

thvasilo:
Hi @shiyu1994 I was trying to follow @guolinke tasklist from #2818 that had listed

> remove data_size_t completely in code, directly used int64_t

I can go back to simply changing the type of `data_size_t` if you think that's a better option. Let me know and I'll make the changes accordingly.

Agreed that a compile option or template that allows for runtime selection would be the best choices. I think for a first iteration a compile-time option would be easiest.

thvasilo:
Switched to simple substitution of the typedef for `data_size_t`. I'll try a few learning tasks and report back, if you have particular benchmarks you'd like to see please let me know.

thvasilo:
@shiyu1994 could you provide some guidance with resolving the Python/C API errors?

Seems like some symbols are now missing, getting errors like:

> E       AttributeError: /home/AzDevOps_azpcontainer/.local/lib/python3.9/site-packages/lightgbm/lib_lightgbm.so: undefined symbol: LGBM_DatasetGetField

I'm assuming the signature of the functions has changed as a result of the type change, but probably some parts of the code are left unchanged causing header/implementation mismatches. It's hard to figure out all parts of the code that need changing as CLion doesn't seem to recognize all targets (e.g. testlighgbm is not recognized).

What's the easiest way to run the tests that the CI runs locally?

shiyu1994:
@thvasilo Thanks for working on this. I just have a quick test on your branch. The declaration and implementation of `LGBM_DatasetGetField` is not compatible.

src/c_api.cpp
```diff
@@ -1491,7 +1491,7 @@ int LGBM_DatasetSetField(DatasetHandle handle,
 
 int LGBM_DatasetGetField(DatasetHandle handle,
                          const char* field_name,
-                         int* out_len,
+                         data_size_t* out_len,
```
shiyu1994, line 1494: We should change the declaration of `LGBM_DatasetGetField` in `c_api.h`.

include/LightGBM/meta.h
```diff
@@ -25,7 +25,7 @@
 namespace LightGBM {
 
 /*! \brief Type of data size, it is better to use signed type*/
-typedef int32_t data_size_t;
+typedef int64_t data_size_t;
```
shiyu1994, line 28: Add a compile option to switch between int64_t and int32_t.

src/c_api.cpp
shiyu1994: In addition, if we change the parameter types of these C APIs, we will break the Python and R-package code where these C APIs are called, and `int32_t*` is passed as the argument.

I think in the first step, it would be good enough to enable `int64_t` for CLI version, and keep the changes within C++ code. So we can keep these C APIs unchanged. That requires converting these `int32_t*` arrays into `int64_t*` arrays manually within these C APIs.

thvasilo: Not sure if I follow: Is is possible to make the switch from int32 to int64 for the C++ package, while isolating them from the Python package? My understanding is that the C_api changes are needed because that's what Python uses for its bindings. 

So moving to int64 causes the Python bindings to break, unless we also make the changes in the c_api as well? 

Is it possible to isolate changes to the CLI version alone?

include/LightGBM/meta.h
thvasilo: I'll try this, there's a lot of places where there are mixes of refs to `data_size_t` and `int32/64` hopefully, I'll be able to isolate them :)

src/c_api.cpp
shiyu1994: In the python/R code, the C API is called
https://github.com/microsoft/LightGBM/blob/946817a505eeedf7b99d975fb7dd7294dd3d4e9d/python-package/lightgbm/basic.py#L1843-L1848
And if we change the C API, we have to change these calls in python/R, for example the `ctypes.c_int` should be changed to `ctypes.c_int64` in line 1798.

So I think we can keep the C APIs unchanged, so that we don't have to make many changes in python/R package now.

But that would require some manual casting of `int32_t` into `int64_t` within the C APIs, for example, in
https://github.com/microsoft/LightGBM/blob/946817a505eeedf7b99d975fb7dd7294dd3d4e9d/src/c_api.cpp#L1386-L1410
we will do something like this to convert the `int32_t` array into `int64_t`,
```c++
std::vector<int64_t> used_row_indices_int64(num_used_row_indices);
for (data_size_t i = 0; i < num_used_row_indices; ++i) {
    used_row_indices_int64[i] = static_cast<int64_t>(used_row_indices[i]);
}
```
And pass `used_row_indices_int64.data()` in place of `used_row_indices` in subsequent function calls

sinhrks:
Hi, is this PR still active? Because I meet the same issue, I’m willing to work on this.

thvasilo:
Hi, I'm no longer working on this project so feel free to pick it up.

PR #4812, opened 2021-11-18
[R-package] Move R6 to Imports

david-cortes:
The R package has R6 as a 'Depends', which causes it to load it fully into the namespace when loading lightgbm. This is not actually required since only one function from this package is used in the R code, and this function is prefixed with `R6::`. This PR moves to 'Imports' instead so that it doesn't load its full namespace in user sessions.

Also it updates the dependency on R to 3.6, since after the latest PRs it won't compile with R 3.5 anymore due to headers not having all the function prototypes.

### DIFF
--- a/R-package/DESCRIPTION
+++ b/R-package/DESCRIPTION
@@ -28,7 +28,7 @@ Authors@R: c(
     person("IBM Corporation", role = c("ctb")),
     person("David", "Cortes", role = c("ctb"))
     )
-Description: Tree based algorithms can be improved by introducing boosting frameworks. 
+Description: Tree based algorithms can be improved by introducing boosting frameworks.
     'LightGBM' is one such framework, based on Ke, Guolin et al. (2017) <https://papers.nips.cc/paper/6907-lightgbm-a-highly-efficient-gradient-boosting-decision>.
     This package offers an R interface to work with it.
     It is designed to be distributed and efficient with the following advantages:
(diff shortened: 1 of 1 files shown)


jameslamb (changes requested):
Thanks for this!

I totally support moving R6 down to `Imports`, makes sense.

> updates the dependency on R to 3.6, since after the latest PRs it won't compile with R 3.5 anymore due to headers not having all the function prototypes.

Can you please provide some specific evidence for this? If not, then I'd like to test this assertion before bumping this floor.

R 3.6.0 came out in April 2019 ([link](https://cran.r-project.org/bin/windows/base/old/)), and R 3.6.3 is the lowest version we tested against in this project's CI 

https://github.com/microsoft/LightGBM/blob/5fa887bb79d0e4d1961c8adafce9832448535e71/.ci/test_r_package.sh#L18-L21

So if recent changes have introduced an incompatibility with R 3.5.x I'm fine increasing this floor. But I'd like to see some evidence of that before accepting this change.

---------------

Also note that I've changed the title of this pull request to be a bit more specific. PR titles in this project become release note items (see https://github.com/microsoft/LightGBM/releases/tag/v3.3.1 for an example), so we try to make them answers to the question "what would users want to know about this change?".

david-cortes:
Did a bit more research. `R_FlushConsole` was introduced in the R header somewhere between version 3.5.0 and 3.6.0. Don't want to find out which one specifically so I'll bump back the R version dependency.

jameslamb:
Ok thank you. I will look into the specific R versions stuff more deeply later, since you've said you're not interested in it at the moment. Thanks for bringing that to our attention!

I want to be protective of that version floor until it starts to cause maintenance problems, since I remember working in locked-down corporate environments where it was very difficult to get upgraded to newer versions of R.

jameslamb (approved):
I support this change, thanks very much!

PR #4815, opened 2021-11-19
[python] raise an informative error instead of segfaulting when custom objective produces incorrect output

yaxxie:
Using the following code can cause a segfault due to `LGBM_BoosterUpdateOneIterCustom` making an assumption that the passed float arrays match the length of the training data.

```python
import lightgbm
import numpy

X = numpy.random.randn(10_000_000, 5)
Y = numpy.random.choice([0, 1], 10_000_000)

ds = lightgbm.Dataset(X, Y)

def bad_grads(x, y):
    return numpy.random.randn(2), numpy.random.rand(2)

lightgbm.train({}, ds, fobj=bad_grads)
```

A faulty `fobj` function can cause at best totally incorrect boosting and at worst segmentation fault. This small patch prevents this from occurring. I noticed it while adding support for `LGBM_BoosterUpdateOneIterCustom` to the julia package (see https://github.com/IQVIA-ML/LightGBM.jl/pull/114/files)


StrikerRUS:
@yaxxie Thanks a lot for this PR!

 >making an assumption that the passed float arrays match the length of the training data.

This assumption is wrong. `grad` and `hess` should be length of `n_samples * num_model_per_iteration`. `num_model_per_iteration` equals to `1` for regression and binary classification and `#classes` for multiclass classification.

Also, what do you think about moving this check to cpp side
https://github.com/microsoft/LightGBM/blob/d517ba12f2e7862ac533908304dddbd770655d2b/src/boosting/goss.hpp#L59
https://github.com/microsoft/LightGBM/blob/b0137debe6e9cc92b65ec71b0fe8a56ea213c143/src/boosting/gbdt.cpp#L371
so that all language packages will benefit from it?

yaxxie:
Thanks @StrikerRUS I figured the assumption would be wrong when I saw some failing tests

If we move this to the CPP side, it would require an API change, right?

StrikerRUS:
> If we move this to the CPP side, it would require an API change, right?

Sorry, could you please clarify what API changes do you mean?

I believe something like
```cpp
int64_t total_size = static_cast<int64_t>(num_data_) * num_tree_per_iteration_;
CHECK_EQ(hessians.size(), hessians.size());
CHECK_EQ(hessians.size(), total_size);
```
would be enough.

yaxxie:
Happy to try this out, but I can't see how it would work. When the data array comes from outside the lib, how do we know its size?

StrikerRUS:
> When the data array comes from outside the lib, how do we know its size?

Sorry, didn't get your question.

yaxxie:
The `gradients` and `hessians` variables are arrays of float, not `std::vector` with `size` methods, and this makes sense because sometimes they are generated internally (so the `gradients_` and `hessians_` variables are `std::vector`), but when they come from outside the library (as in being passed the pointers from `LGBM_BoosterUpdateOneIterCustom`) we do not (cannot) know their sizes.

The other C-API points will pass size variables where they need to know them (such as when we construct from mat and we need a size variable there, or when we retrieve string names and we need to know how much was allocated to the underlying buffer so as not to overwrite past allocated memory).

This is why the API can segfault when passed incorrectly sized gradients, since, and underlying assumption is made about the size of the allocated buffers which is not (and cannot) be verified by the library itself. We could have the user pass the sizes to the API call, but this would then still require changes at all language implementations.

I'm happy for someone to point out where I've gone wrong with the above.

(I tried the code sample provided, it won't compile because float*'s don't have `size` method)
```
/home/yaxattax/git/LightGBM/include/LightGBM/utils/log.h:35:9: note: in definition of macro ‘CHECK’
   35 |   if (!(condition))                                                         \
      |         ^~~~~~~~~
/home/yaxattax/git/LightGBM/src/boosting/gbdt.cpp:385:3: note: in expansion of macro ‘CHECK_EQ’
  385 |   CHECK_EQ(hessians.size(), gradients.size());
      |   ^~~~~~~~
/home/yaxattax/git/LightGBM/src/boosting/gbdt.cpp:386:21: error: request for member ‘size’ in ‘hessians’, which is of non-class type ‘const score_t*’ {aka ‘const float*’}
  386 |   CHECK_EQ(hessians.size(), total_size);
      |                     ^~~~

```

shiyu1994:
@yaxxie Thanks for working on this! Yes, I think there should be a change in C API. Currently the C API only accepts the pointers to the gradients and hessians. If we want to know the length of the array allocated outside C API, we must add new parameters. But the change in C API may require changes in the code of every language packages.

So maybe doing the check in the Python and R side is preferable. @StrikerRUS WDYT.

yaxxie:
> So maybe doing the check in the Python and R side is preferable. 

We'd also want to update the docs for C-API to make it explicit that there is a length expectation for the array of floats passed.

StrikerRUS:
@yaxxie Sorry, I didn't notice what type `gradients` and `hessians` actually are. I confused them with `gradients_` and `hessians_`.

@shiyu1994 
> If we want to know the length of the array allocated outside C API, we must add new parameters. But the change in C API may require changes in the code of every language packages.

Can we use the fact that those arrays **should always be** the length of `static_cast<int64_t>(num_data_) * num_tree_per_iteration_`? I believe new parameters for actual lengths makes a little sense here, as we already know their size in correct API usage scenario.

> So maybe doing the check in the Python and R side is preferable.

If we cannot do any checks with raw pointers, I'm OK with this way.

shiyu1994:
>Can we use the fact that those arrays should always be the length of static_cast<int64_t>(num_data_) * num_tree_per_iteration_?

@StrikerRUS  I think fix itself is the guarantee this. To check the size in C API side, we must know the exact length of the allocated array in the Python side.

python-package/lightgbm/basic.py
```diff
@@ -3079,7 +3079,14 @@ def __boost(self, grad, hess):
         assert grad.flags.c_contiguous
         assert hess.flags.c_contiguous
         if len(grad) != len(hess):
-            raise ValueError(f"Lengths of gradient({len(grad)}) and hessian({len(hess)}) don't match")
+            raise ValueError(f"Lengths of gradient ({len(grad)}) and hessian ({len(hess)}) don't match")
+        num_train_data = self.train_set.num_data()
+        num_models = self.num_model_per_iteration()
```
StrikerRUS, line 3084: Use precomputed at the init phase value for the speedup.
https://github.com/microsoft/LightGBM/blob/67b4205c8043326553e294fa2c01ad1189784631/python-package/lightgbm/basic.py#L2645-L2649
```suggestion
        num_models = self.__num_class
```

python-package/lightgbm/basic.py
```diff
@@ -3079,7 +3079,14 @@ def __boost(self, grad, hess):
...
+        if len(grad) != (num_train_data * num_models):
+            raise ValueError(
+                f"Lengths of gradient ({len(grad)}) and hessian ({len(hess)}) "
+                + f"don't match training data length ({num_train_data}) * number of models ({num_models})"
+            )
```
StrikerRUS, line 3089:
```suggestion
        if len(grad) != num_train_data * num_models:
            raise ValueError(
                f"Lengths of gradient ({len(grad)}) and Hessian ({len(hess)}) "
                f"don't match training data length ({num_train_data}) * number of models per one iteration ({num_models})"
            )
```

python-package/lightgbm/basic.py
```diff
@@ -3079,7 +3079,14 @@ def __boost(self, grad, hess):
         assert grad.flags.c_contiguous
         assert hess.flags.c_contiguous
         if len(grad) != len(hess):
-            raise ValueError(f"Lengths of gradient({len(grad)}) and hessian({len(hess)}) don't match")
+            raise ValueError(f"Lengths of gradient ({len(grad)}) and hessian ({len(hess)}) don't match")
```
StrikerRUS, line 3082:
```suggestion
            raise ValueError(f"Lengths of gradient ({len(grad)}) and Hessian ({len(hess)}) don't match")
```

StrikerRUS (changes requested):
Please consider checking some my minor suggestions below:

yaxxie:
@StrikerRUS Thanks will get around to doing these properly soon
I want to add a simple test and also I think it would be worthwhile updating the C-API documentation to state this requirement upon the caller -- could you point me in the right place to make this change?

StrikerRUS:
@yaxxie Thank you!
Basic tests in which you call C API directly should be added in the following file:
https://github.com/microsoft/LightGBM/blob/master/tests/c_api_test/test_.py
For basic Python tests where you don't use `train()` or `cv()` function from `engine.py` file you should use this file
https://github.com/microsoft/LightGBM/blob/master/tests/python_package_test/test_basic.py
If you want to add cpp tests, they should go to somewhere in this folder
https://github.com/microsoft/LightGBM/tree/master/tests/cpp_tests

As for modifying C API docs, edit comments in Doxygen format in this file
https://github.com/microsoft/LightGBM/blob/master/include/LightGBM/c_api.h

include/LightGBM/c_api.h
```diff
@@ -572,6 +572,9 @@ LIGHTGBM_C_EXPORT int LGBM_BoosterRefit(BoosterHandle handle,
 /*!
  * \brief Update the model by specifying gradient and Hessian directly
  *        (this can be used to support customized loss functions).
+ * \note
+ * The length of the arrays referenced by ``grad`` and ``hess`` must be equal to
+ * ``num_class*num_train_data``, this is not verified by the library, the caller must ensure this.
```
StrikerRUS, line 577:
```suggestion
 * ``num_class * num_train_data``, this is not verified by the library, the caller must ensure this.
```

StrikerRUS (approved):
Thanks a lot getting back to this and adding tests!
LGTM! Just one nit to improve readability.

yaxxie:
@StrikerRUS Anything else needed?

StrikerRUS:
@jameslamb @shiyu1994 Would you like to be a second reviewer for this PR?

tests/python_package_test/test_basic.py
```diff
@@ -579,3 +579,30 @@ def test_param_aliases():
...
+    classes = [0, 1, 2]
+    y_multiclass = np.random.choice(classes, 100)
+    ds_binary = lgb.Dataset(X, y_binary).construct()
+    ds_multiclass = lgb.Dataset(X, y_multiclass).construct()
+    bad_bst_binary = lgb.Booster({'objective': "none"}, ds_binary)
+    good_bst_binary = lgb.Booster({'objective': "none"}, ds_binary)
+    bad_bst_multi = lgb.Booster({'objective': "none", "num_class": len(classes)}, ds_multiclass)
+    good_bst_multi = lgb.Booster({'objective': "none", "num_class": len(classes)}, ds_multiclass)
+    good_bst_binary.update(fobj=_good_gradients)
+    with pytest.raises(ValueError):
```
jameslamb, line 604:
```suggestion
    with pytest.raises(ValueError, match="number of models per one iteration (1)"):
```

Could you please use `match` to look for a specific error message? That way, this test won't silently pass if a change results in `lightgbm` raising a different, unrelated `ValueError`.

tests/python_package_test/test_basic.py
```diff
@@ -579,3 +579,30 @@ def test_param_aliases():
...
+        bad_bst_binary.update(fobj=_bad_gradients)
+    good_bst_multi.update(fobj=_good_gradients)
+    with pytest.raises(ValueError):
```
jameslamb, line 607:
```suggestion
    with pytest.raises(ValueError, match="number of models per one iteration (3)"):
```

jameslamb (changes requested):
Thanks very much for this!

I followed the conversation with @StrikerRUS and @shiyu1994 and understand why we've chosen to do this check on the Python / R side instead of in C/C++.

Please see two small suggestions to make the tests slightly stricter. Would you please also write up a feature request at https://github.com/microsoft/LightGBM/issues documenting the need to do this same work for the R package?

jameslamb:
@yaxxie @StrikerRUS I just changed the title of this PR to hopefully be a bit more informative for the purposes of release notes

yaxxie:
@jameslamb I opened #4905 and pushed commit to address your remarks. Please do let me know if anything else is required.

tests/python_package_test/test_basic.py
```diff
@@ -579,3 +579,30 @@ def test_param_aliases():
...
+    classes = [0, 1, 2]
+    y_multiclass = np.random.choice(classes, 100)
+    ds_binary = lgb.Dataset(X, y_binary).construct()
+    ds_multiclass = lgb.Dataset(X, y_multiclass).construct()
+    bad_bst_binary = lgb.Booster({'objective': "none"}, ds_binary)
+    good_bst_binary = lgb.Booster({'objective': "none"}, ds_binary)
+    bad_bst_multi = lgb.Booster({'objective': "none", "num_class": len(classes)}, ds_multiclass)
+    good_bst_multi = lgb.Booster({'objective': "none", "num_class": len(classes)}, ds_multiclass)
+    good_bst_binary.update(fobj=_good_gradients)
+    with pytest.raises(ValueError, match="number of models per one iteration \(1\)"):
```
jameslamb, line 604:
```suggestion
    with pytest.raises(ValueError, match="number of models per one iteration \\(1\\)"):
```

Please see these linting errors seen in https://github.com/microsoft/LightGBM/runs/4613473477?check_suite_focus=true

> ./tests/python_package_test/test_basic.py:604:78: W605 invalid escape sequence '\('
./tests/python_package_test/test_basic.py:604:81: W605 invalid escape sequence '\)'
./tests/python_package_test/test_basic.py:607:79: W605 invalid escape sequence '\('
./tests/python_package_test/test_basic.py:607:95: W605 invalid escape sequence '\)'

tests/python_package_test/test_basic.py
```diff
@@ -579,3 +579,30 @@ def test_param_aliases():
...
+        bad_bst_binary.update(fobj=_bad_gradients)
+    good_bst_multi.update(fobj=_good_gradients)
+    with pytest.raises(ValueError, match=f"number of models per one iteration \({len(classes)}\)"):
```
jameslamb, line 607:
```suggestion
    with pytest.raises(ValueError, match=f"number of models per one iteration \\({len(classes)}\\)"):
```

tests/python_package_test/test_basic.py
```diff
@@ -579,3 +579,30 @@ def test_param_aliases():
...
+
+def _good_gradients(labels, preds):
+    return np.random.randn(len(labels)), np.random.rand(len(labels))
+
+
+def test_custom_objective_safety():
+    X = np.random.randn(100, 5)
+    y_binary = np.random.choice([0, 1], 100)
+    classes = [0, 1, 2]
+    y_multiclass = np.random.choice(classes, 100)
```
jameslamb, line 596:
```suggestion
    X = np.random.randn(100, 5)
    y_binary = np.array([0] * 50 + [1] * 50)
    classes = [0, 1, 2]
    y_multiclass = np.random.choice([0] * 33 + [1] * 33 + [2] * 34)
```

Sorry, just thought of this...can you please remove the randomness from this data construction? Since you're using completely-random data and not testing the produced models, the values don't really matter.

Choosing randomly and using such a small amount of data makes it possible that these tests could fail randomly due to situations like "`y_binary` is all 0s". It may seem like a small probability, but consider that the Python tests run about 40 times on every commit to every pull request in this project.

jameslamb (changes requested):
Thanks for the test changes and for opening #4905! Please see a few more small suggestions.

tests/python_package_test/test_basic.py
StrikerRUS: Please don't escape any symbols for the readability purpose. Just add a `*` symbol at the end:
```suggestion
    with pytest.raises(ValueError, match="number of models per one iteration (1) *"):
```

yaxxie: The escape is necessary; `(` and `)` are characters which means something to the regular expression engine. I'll switch to using `re.escape`

yaxxie:
@jameslamb let me know if a946d28 addresses your concerns

jameslamb (approved):
looks ok to me, thanks very much for the help!

tests/python_package_test/test_basic.py
```diff
@@ -579,3 +580,32 @@ def test_param_aliases():
     assert all(len(i) >= 1 for i in aliases.values())
     assert all(k in v for k, v in aliases.items())
     assert lgb.basic._ConfigAliases.get('config', 'task') == {'config', 'config_file', 'task', 'task_type'}
+
+
+def _bad_gradients(labels, preds):
+    return np.random.randn(len(labels) + 1), np.random.rand(len(labels) + 1)
```
StrikerRUS, line 586: Sorry for the oversight in my previous review! Current signature is for sklearn wrapper, standard API accepts Dataset object instead of true labels.
https://lightgbm.readthedocs.io/en/latest/pythonapi/lightgbm.Booster.html#lightgbm.Booster.update

```suggestion
def _bad_gradients(preds, train_data):
    return np.random.randn(len(preds) + 1), np.random.rand(len(preds) + 1)
```

tests/python_package_test/test_basic.py
```diff
@@ -579,3 +580,32 @@ def test_param_aliases():
...
+
+
+def _good_gradients(labels, preds):
+    return np.random.randn(len(labels)), np.random.rand(len(labels))
```
StrikerRUS, line 590:
```suggestion
def _good_gradients(preds, train_data):
    return np.random.randn(len(preds)), np.random.rand(len(preds))
```

StrikerRUS (changes requested):

include/LightGBM/c_api.h
```diff
@@ -572,6 +572,9 @@ LIGHTGBM_C_EXPORT int LGBM_BoosterRefit(BoosterHandle handle,
 /*!
  * \brief Update the model by specifying gradient and Hessian directly
  *        (this can be used to support customized loss functions).
+ * \note
+ * The length of the arrays referenced by ``grad`` and ``hess`` must be equal to
+ * ``num_class * num_train_data``, this is not verified by the library, the caller must ensure this.
```
shiyu1994, line 577: Should this be `this IS verified by the library`, or simply delete this last sentence? Because we are actually verifying this through this pull request.

yaxxie: This is the C-API docs -- the context here is the `lightgbm.so` library, rather than python or R libraries. What we're saying is that a caller to this function `LGBM_BoosterUpdateOneIterCustom` is responsible to ensure that the condition is met. I'm happy to tweak the wording, but what the python library does as a convenience to the user is not applicable here.

tests/python_package_test/test_basic.py
yaxxie: 2f6b647

tests/python_package_test/test_basic.py
yaxxie: 2f6b647

StrikerRUS (approved):
Thanks for fixing custom objective function signature in recent commit.

PR #4547, opened 2021-08-23
[python] add type hints for custom objective and metric functions in scikit-learn interface

jameslamb:
Created in response to https://github.com/microsoft/LightGBM/pull/4544#pullrequestreview-735574138.

Contributes to #3756.

Proposes introducing more specific type hints for custom objective and metric functions in the scikit-learn and Dask interfaces in the Python package.

PR #4816, opened 2021-11-20
[python][docs] fix type hints for custom functions and remove vague `array-like` wording

StrikerRUS:
According to https://github.com/microsoft/LightGBM/issues/3756#issuecomment-907622141.
Follow-up PR after #4547.
Based on the investigation from the following thread: https://github.com/microsoft/LightGBM/pull/4547#discussion_r695989171.

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -112,6 +112,7 @@ _LIB = _load_lib()
 
 
 NUMERIC_TYPES = (int, float, bool)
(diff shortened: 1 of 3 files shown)


jameslamb (approved):
Looks great, and I support  the idea of consolidating the type hint into an internal constant, `_ArrayLike`.

The hints and docs look correct to me based on the linked conversations, and I don't see anything problematic in `mypy` checks in the `lint` task on this PR ([link](https://github.com/microsoft/LightGBM/runs/4270648514?check_suite_focus=true)).

PR #4822, opened 2021-11-22
[python][sklearn] respect parameters for predictions in `init()` and `set_params()` methods

### LINKED ISSUE #4706: [python] Changing `n_jobs` through scikit-learn interface has no effect

StrikerRUS:
Fixed #4706.


jameslamb (approved):
I like this approach, thanks!

PR #4823, opened 2021-11-22
[docs] document that `pred_early_stop` can be used only in normal and raw scores prediction

StrikerRUS:
Note that `early_stop_` is used only in `PredictRaw()` and `Predict()` cases:
https://github.com/microsoft/LightGBM/blob/bfb346c13e6ea44f8b663a66bfc41e189d39a1fb/src/application/predictor.hpp#L71-L138


jameslamb (approved):
makes sense to me, thanks

PR #4768, opened 2021-11-04
Suppress categorical warning (fixes #3379)

hzy46:
To fix #3379

PR #4824, opened 2021-11-23
[python] add type hints to `_compare_params_for_warning()` and make it reusable

StrikerRUS:
Pass keys that should be ignored during comparing to make `_compare_params_for_warning()` function reusable.

Function was introduced in #4768.


jameslamb (approved):
totally support this re-factoring, I think this is a bit easier to follow. Thanks!

PR #4830, opened 2021-11-28
[docs][python] simplify mocking in docs

StrikerRUS:
Refer to https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html.


jameslamb (approved):
nice, thanks!

PR #4832, opened 2021-11-28
[python][sklearn] remove `verbose` argument from `fit()` method

StrikerRUS:
Refer to #4574.


jameslamb (approved):
Looks good to me! I pulled the branch and re-built the docs to check, Dask and Python API docs look correct to me.

PR #4833, opened 2021-11-28
[tests][dask] fix argument names in custom eval function in Dask test

StrikerRUS:
RTD builds for visual checks: https://lightgbm.readthedocs.io/en/note_shape/.

### DIFF
--- a/docs/Parameters.rst
+++ b/docs/Parameters.rst
@@ -923,6 +923,8 @@ Predict Parameters
 
    -  used only in ``classification`` and ``ranking`` applications
 
+   -  used only for predicting normal or raw scores
+
    -  if ``true``, will use early-stopping to speed up the prediction. May affect the accuracy
 
    -  **Note**: cannot be used with ``rf`` boosting type or custom objective function
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -25,7 +25,6 @@ from re import compile
 from shutil import copytree
 from subprocess import PIPE, Popen
 from typing import Any, List
-from unittest.mock import Mock
 
 import sphinx
 from docutils.nodes import reference
@@ -39,22 +38,6 @@ LIB_PATH = CURR_PATH.parent / 'python-package'
 sys.path.insert(0, str(LIB_PATH))
 
 INTERNAL_REF_REGEX = compile(r"(?P<url>\.\/.+)(?P<extension>\.rst)(?P<anchor>$|#)")
-
-# -- mock out modules
-MOCK_MODULES = [
-    'dask',
-    'dask.distributed',
-    'datatable',
-    'graphviz',
-    'matplotlib',
-    'numpy',
-    'pandas',
-    'scipy',
-    'scipy.sparse',
-    'sklearn'
-]
-for mod_name in MOCK_MODULES:
-    sys.modules[mod_name] = Mock()
 
 
 class InternalRefTransform(Transform):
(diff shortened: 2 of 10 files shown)


jameslamb (changes requested):
Thanks for looking into this! But I don't think the proposed change is correct.

When using, for example, `DaskLGBMRegressor`, `lightgbm.dask` will create one task running `LGBMRegressor.fit()` on each Dask worker.

https://github.com/microsoft/LightGBM/blob/2f5d898520dfa7a09f9f73794e5a0586eea1d4f5/python-package/lightgbm/dask.py#L765-L767

https://github.com/microsoft/LightGBM/blob/2f5d898520dfa7a09f9f73794e5a0586eea1d4f5/python-package/lightgbm/dask.py#L164

https://github.com/microsoft/LightGBM/blob/2f5d898520dfa7a09f9f73794e5a0586eea1d4f5/python-package/lightgbm/dask.py#L319-L329

At that moment, that evaluation function will be called on only _local_ data in each worker process, which will be numpy arrays, not Dask arrays.

Here's some sample code proving that. Note that it includes an `assert` which will fail if `y_true` is anything other than a numpy array.

<details><summary>(sample code)</summary>

```python
import dask.array as da
import numpy as np
from lightgbm.dask import DaskLGBMRegressor

# set  up Dask cluster
n_workers = 2
client = Client(cluster)

# create training data and an additional eval set
def _make_dataset(n_samples):
    X, y = make_regression(n_samples=n_samples)
    dX = da.from_array(X, chunks=(1000, X.shape[1]))
    dy = da.from_array(y, chunks=1000)
    return dX, dy

# training data
dX, dy = _make_dataset(10_000)

# eval data
dX_e, dy_e = _make_dataset(2_000)

def _custom_metric(y_true, y_pred):
    metric_name = "custom_metric"
    is_higher_better = False
    metric_value = 0.708
    assert str(type(y_true)) == "<class 'numpy.ndarray'>"
    return metric_name, metric_value, is_higher_better

dask_reg = DaskLGBMRegressor(
    client=client,
    objective="regression_l2",
    tree_learner="data"
)
dask_reg.fit(
    X=dX,
    y=dy,
    eval_set=[
        (dX, dy),
        (dX_e, dy_e)
    ],
    eval_metric = ["rmse", _custom_metric]
)

# confirm that custom metric function was actually called
# (should be an array of value 0.708)
print(dask_reg.evals_result_["valid_0"]["custom_metric"])
```

</details>

StrikerRUS:
Ah, thank you very much! I was confused by the metrics in our tests:

https://github.com/microsoft/LightGBM/blob/2f5d898520dfa7a09f9f73794e5a0586eea1d4f5/tests/python_package_test/test_dask.py#L222-L236

They all takes arguments named `dy_true` and `dy_pred` where `d` stays for "Dask", I guess.
I think it's better to remove `d` prefix from the arguments in `_constant_metric()` custom eval function. WDYT?

jameslamb:
> where d stays for "Dask", I guess.

ah yep, you're right! That is confusing. I'd support removing the `d` from the arguments in `_constant_metric()`.

StrikerRUS:
@jameslamb 
> ah yep, you're right! That is confusing. I'd support removing the `d` from the arguments in `_constant_metric()`.

Great! I reverted docstring changes and pushed a fix for argument names.

jameslamb (approved):

PR #4837, opened 2021-11-30
[python-package] fix mypy errors in sklearn.py

jameslamb:
Contributes to #3867.

This PR fixes `mypy` errors in `sklearn.py`.

> python-package/lightgbm/sklearn.py:532: error: Need type annotation for "_other_params" (hint: "_other_params: Dict[<type>, <type>] = ...")
python-package/lightgbm/sklearn.py:879: error: "None" has no attribute "current_iteration"
python-package/lightgbm/sklearn.py:890: error: "None" has no attribute "current_iteration"
python-package/lightgbm/sklearn.py:944: error: Item "None" of "Optional[str]" has no attribute "replace"
python-package/lightgbm/sklearn.py:1011: error: Item "None" of "Optional[str]" has no attribute "replace"
python-package/lightgbm/sklearn.py:1110: error: Item "None" of "Optional[str]" has no attribute "replace"

```shell
mypy \
    --ignore-missing-imports \
    python-package/
```

### DIFF
--- a/python-package/lightgbm/sklearn.py
+++ b/python-package/lightgbm/sklearn.py
@@ -2,11 +2,11 @@
 """Scikit-learn wrapper interface for LightGBM."""
 import copy
 from inspect import signature
-from typing import Callable, Dict, List, Optional, Tuple, Union
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
 import numpy as np
 
-from .basic import Dataset, LightGBMError, _ArrayLike, _choose_param_value, _ConfigAliases, _log_warning
+from .basic import Booster, Dataset, LightGBMError, _ArrayLike, _choose_param_value, _ConfigAliases, _log_warning
 from .callback import log_evaluation, record_evaluation
 from .compat import (SKLEARN_INSTALLED, LGBMNotFittedError, _LGBMAssertAllFinite, _LGBMCheckArray,
                      _LGBMCheckClassificationTargets, _LGBMCheckSampleWeight, _LGBMCheckXY, _LGBMClassifierBase,
@@ -525,11 +525,11 @@ class LGBMModel(_LGBMModelBase):
         self.random_state = random_state
         self.n_jobs = n_jobs
         self.importance_type = importance_type
-        self._Booster = None
+        self._Booster: Optional[Booster] = None
         self._evals_result = None
         self._best_score = None
         self._best_iteration = None
-        self._other_params = {}
+        self._other_params: Dict[str, Any] = {}
         self._objective = objective
         self.class_weight = class_weight
         self._class_weight = None
@@ -874,7 +874,7 @@ class LGBMModel(_LGBMModelBase):
         This might be less than parameter ``n_estimators`` if early stopping was enabled or
         if boosting stopped early due to limits on complexity like ``min_gain_to_split``.
         """
-        if not self.__sklearn_is_fitted__():
+        if not self.__sklearn_is_fitted__() or self._Booster is None:
             raise LGBMNotFittedError('No n_estimators found. Need to call fit beforehand.')
         return self._Booster.current_iteration()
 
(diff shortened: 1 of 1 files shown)


python-package/lightgbm/sklearn.py
```diff
@@ -1008,7 +1008,7 @@ def fit(self, X, y,
                     callbacks=callbacks, init_model=init_model)
         return self
 
-    _base_doc = LGBMModel.fit.__doc__.replace("self : LGBMModel", "self : LGBMClassifier")
+    _base_doc = LGBMModel.fit.__doc__.replace("self : LGBMModel", "self : LGBMClassifier")  # type: ignore
```
jameslamb, line 1011: warnings about `.__doc__` being `Optional[str]` could be avoided by changing this to something like `getattr(LGBMModel.fit, "__doc__", "")`, but I think this is preferable, so that `.replace()` raises an exception in CI if something is accidentally changed that results in the docs being empty for a method.

StrikerRUS (changes requested):
Type hints and `#ignore` directives look OK to me, but I disagree with actual code complication for the sake of removing mypy false alarm.

python-package/lightgbm/sklearn.py
```diff
@@ -874,7 +874,7 @@ def n_estimators_(self) -> int:
         This might be less than parameter ``n_estimators`` if early stopping was enabled or
         if boosting stopped early due to limits on complexity like ``min_gain_to_split``.
         """
-        if not self.__sklearn_is_fitted__():
+        if not self.__sklearn_is_fitted__() or self._Booster is None:
```
StrikerRUS, line 877: `__sklearn_is_fitted__()` was introduced with the aim to avoid a such type of checks for initial values of attributes: #4654. I don't think that workaround for mypy error worth introducing nonsense code. I'm not supporting this particular change.

jameslamb: Consider the following:

```python
import lightgbm as lgb

data = lgb.Dataset(X, label=y)

# train model
clf = lgb.LGBMClassifier(
    objective="binary",
    n_estimators=5
)

clf.fit(X, y)
clf._Booster = None
clf.n_estimators_
```

On `master`, that code will yield this error:

> AttributeError: 'NoneType' object has no attribute 'current_iteration'

As of this PR, it would return this error

> No n_estimators found. Need to call fit beforehand.

So this is not a "false positive" from `mypy`. `mypy` is correctly complaining about the fact that this code path and others like it treat `self._Booster` as non-None even though it is possible for it to be `None`.

This is also why I don't see the proposed change here as "nonsense". With the change proposed on this line, in the unlikely and unsupported but **possible** scenario where `_Booster` has been set to `None`, you'll get a more informative error message.

--------------

Anyway, I'll remove this and use `# type: ignore` comments. But wanted you to understand that I wasn't just trying to do weird things to silence a linter.

jameslamb: updated in https://github.com/microsoft/LightGBM/pull/4837/commits/6e3420187f1115a7690a03b81fc417cbe28c6e76

StrikerRUS: > `clf._Booster = None`

Accessing private attributes, especially setting them to a custom value, from an outside code shouldn't happen in a normal scenario. I think it sounds quite weird to set private attribute to a random value and expect correct behavior. It's likely the same as
```python
...

clf._Booster = 42
# expect something meaningful
clf.predict(X)
```

As we set all private attributes in a `fit()` method and users shouldn't modify them, I still believe we shouldn't add more checks and use only `__sklearn_is_fitted__()`.

StrikerRUS (approved):
LGTM, thanks a lot!

python-package/lightgbm/sklearn.py
jameslamb: for sure, this is why I thought of this situation as "unlikely and unsupported but possible". I'm ok with not adding these checks and just using `# type: ignore`.

PR #4839, opened 2021-11-30
[python-package] fix mypy errors in engine.py

jameslamb:
Contributes to #3867.

This PR fixes `mypy` errors in `engine.py`

> python-package/lightgbm/engine.py:226: error: Incompatible types in assignment (expression has type "Set[<nothing>]", variable has type "Optional[List[Callable[..., Any]]]")
python-package/lightgbm/engine.py:230: error: Incompatible types in assignment (expression has type "Set[Callable[..., Any]]", variable has type "Optional[List[Callable[..., Any]]]")
python-package/lightgbm/engine.py:242: error: Item "List[Callable[..., Any]]" of "Optional[List[Callable[..., Any]]]" has no attribute "add"
python-package/lightgbm/engine.py:242: error: Item "None" of "Optional[List[Callable[..., Any]]]" has no attribute "add"
python-package/lightgbm/engine.py:244: error: Item "List[Callable[..., Any]]" of "Optional[List[Callable[..., Any]]]" has no attribute "add"
python-package/lightgbm/engine.py:244: error: Item "None" of "Optional[List[Callable[..., Any]]]" has no attribute "add"
python-package/lightgbm/engine.py:247: error: Item "List[Callable[..., Any]]" of "Optional[List[Callable[..., Any]]]" has no attribute "add"
python-package/lightgbm/engine.py:247: error: Item "None" of "Optional[List[Callable[..., Any]]]" has no attribute "add"
python-package/lightgbm/engine.py:252: error: Item "List[Callable[..., Any]]" of "Optional[List[Callable[..., Any]]]" has no attribute "add"
python-package/lightgbm/engine.py:252: error: Item "None" of "Optional[List[Callable[..., Any]]]" has no attribute "add"
python-package/lightgbm/engine.py:254: error: Item "None" of "Optional[List[Callable[..., Any]]]" has no attribute "__iter__" (not iterable)
python-package/lightgbm/engine.py:255: error: Unsupported left operand type for - ("List[Callable[..., Any]]")
python-package/lightgbm/engine.py:255: error: Unsupported left operand type for - ("None")
python-package/lightgbm/engine.py:255: note: Left operand is of type "Optional[List[Callable[..., Any]]]"
python-package/lightgbm/engine.py:256: error: Incompatible types in assignment (expression has type "List[Union[Callable[..., Any], Any]]", variable has type "Set[Union[Callable[..., Any], Any]]")
python-package/lightgbm/engine.py:300: error: Incompatible types in assignment (expression has type "float", variable has type "List[Any]")

```shell
mypy \
    --ignore-missing-imports \
    python-package/
```

### Notes for Reviewers

Some of these errors caught an issue in the type hints for `callbacks.EarlyStopException`. That exception has a property `best_score` which seems like it should be a float, but in every place where one of those is created, `best_score` is populated with a list of eval results similar to this:

```python
[
    ('training', 'binary_logloss', 0.014361246252423975, False),
    ('training', 'mape', 0.013889796906138775, False),
    ('valid', 'binary_logloss', 0.0561150102648419, False),
    ('valid', 'mape', 0.042410104106485866, False)
]
```

https://github.com/microsoft/LightGBM/blob/67b4205c8043326553e294fa2c01ad1189784631/python-package/lightgbm/engine.py#L300

https://github.com/microsoft/LightGBM/blob/67b4205c8043326553e294fa2c01ad1189784631/python-package/lightgbm/callback.py#L280

https://github.com/microsoft/LightGBM/blob/67b4205c8043326553e294fa2c01ad1189784631/python-package/lightgbm/callback.py#L307

### DIFF
--- a/python-package/lightgbm/callback.py
+++ b/python-package/lightgbm/callback.py
@@ -25,7 +25,7 @@ class EarlyStopException(Exception):
         ----------
         best_iteration : int
             The best iteration stopped.
-        best_score : list of (eval_name, metric_name, eval_result, is_higher_better) tuples
+        best_score : list of (eval_name, metric_name, eval_result, is_higher_better) tuple
             Scores for each metric, on each validation set, as of the best iteration.
         """
         super().__init__()
--- a/python-package/lightgbm/callback.py
+++ b/python-package/lightgbm/callback.py
@@ -2,7 +2,7 @@
 """Callbacks library."""
 import collections
 from functools import partial
-from typing import Any, Callable, Dict, List, Union
+from typing import Any, Callable, Dict, List, Tuple, Union
 
 from .basic import _ConfigAliases, _log_info, _log_warning
 
@@ -18,15 +18,15 @@ def _lt_delta(curr_score: float, best_score: float, delta: float) -> bool:
 class EarlyStopException(Exception):
     """Exception of early stopping."""
 
-    def __init__(self, best_iteration: int, best_score: float) -> None:
+    def __init__(self, best_iteration: int, best_score: List[Tuple[str, str, float, bool]]) -> None:
         """Create early stopping exception.
 
         Parameters
         ----------
         best_iteration : int
             The best iteration stopped.
-        best_score : float
-            The score of the best iteration.
+        best_score : list of (eval_name, metric_name, eval_result, is_higher_better) tuples
+            Scores for each metric, on each validation set, as of the best iteration.
         """
         super().__init__()
         self.best_iteration = best_iteration
--- a/python-package/lightgbm/callback.py
+++ b/python-package/lightgbm/callback.py
@@ -6,7 +6,6 @@ from typing import Any, Callable, Dict, List, Tuple, Union
 
 from .basic import _ConfigAliases, _log_info, _log_warning
 
-
 _EvalResultTuple = Union[
     List[Tuple[str, str, float, bool]],
     List[Tuple[str, str, float, bool, float]]
--- a/python-package/lightgbm/callback.py
+++ b/python-package/lightgbm/callback.py
@@ -7,6 +7,12 @@ from typing import Any, Callable, Dict, List, Tuple, Union
 from .basic import _ConfigAliases, _log_info, _log_warning
 
 
+_EvalResultTuple = Union[
+    List[Tuple[str, str, float, bool]],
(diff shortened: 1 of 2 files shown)


python-package/lightgbm/callback.py
```diff
@@ -18,15 +18,15 @@ def _lt_delta(curr_score: float, best_score: float, delta: float) -> bool:
...
+    def __init__(self, best_iteration: int, best_score: List[Tuple[str, str, float, bool]]) -> None:
         """Create early stopping exception.
 
         Parameters
         ----------
         best_iteration : int
             The best iteration stopped.
-        best_score : float
-            The score of the best iteration.
+        best_score : list of (eval_name, metric_name, eval_result, is_higher_better) tuples
```
StrikerRUS, line 28: Refer to #4575.
```suggestion
        best_score : list of (eval_name, metric_name, eval_result, is_higher_better) tuple
```

python-package/lightgbm/callback.py
```diff
@@ -18,15 +18,15 @@ def _lt_delta(curr_score: float, best_score: float, delta: float) -> bool:
 class EarlyStopException(Exception):
     """Exception of early stopping."""
 
-    def __init__(self, best_iteration: int, best_score: float) -> None:
+    def __init__(self, best_iteration: int, best_score: List[Tuple[str, str, float, bool]]) -> None:
```
StrikerRUS, line 21: It seems that this list can contain 5-element tuples.
https://github.com/microsoft/LightGBM/blob/67b4205c8043326553e294fa2c01ad1189784631/python-package/lightgbm/callback.py#L280
https://github.com/microsoft/LightGBM/blob/67b4205c8043326553e294fa2c01ad1189784631/python-package/lightgbm/callback.py#L275
https://github.com/microsoft/LightGBM/blob/67b4205c8043326553e294fa2c01ad1189784631/python-package/lightgbm/callback.py#L51

StrikerRUS (changes requested):

python-package/lightgbm/callback.py
jameslamb: oooo thank you! I checked the format of this with a debugger, but forgot you could ask to show standard deviation.

I'll update this hint.

StrikerRUS (approved):
Thank you very much for fixing callback typehints!

Issue #4005, opened 2021-02-20
MiKTeX setup error at CI in master


StrikerRUS:
```
Setting up MiKTeX
miktexsetup.exe: The remote package repository is outdated. You have to choose another repository.
miktexsetup.exe: Data: url="https://ctan.math.illinois.edu/systems/win32/miktex/tm/packages/"
```
https://github.com/microsoft/LightGBM/blob/646267d2656cbd13d86e498e96c6db95a9bae5d8/.ci/test_r_package_windows.ps1#L59
Only for Windows. 
Full logs: https://github.com/microsoft/LightGBM/runs/1941681116

cc @jameslamb

jameslamb:
haha NOOOOO we had been doing so well for so long!

I remember that the last time this happened, it went away on its own after a day.

https://github.com/microsoft/LightGBM/issues/3454#issuecomment-707474556

> I think this is just a thing that happens with CTAN a few times a year and we have to wait another day for it to hopefully improve :/

I'll try manually re-running the R jobs on `master` few times over the next few hours, let's see if it resolves itself. If

StrikerRUS:
> it's fine with me if you want to merge things with administrator powers

I don't have them.

jameslamb:
ah I see, my mistake. Well I'm busy for the next few hours, but can look at this tonight if it isn't resolved just by waiting.

jameslamb:
I just started https://github.com/microsoft/LightGBM/runs/1943021929 and I'm hopeful about it, because it looks like the `math.illinois.edu` mirror was updated within the last hour

http://dante.ctan.org/mirmon/

![image](https://user-images.githubusercontent.com/7608904/108605651-22fb5380-737b-11eb-8873-b6afda641ecd.png)

jameslamb:
nope, sad, same failure

> miktexsetup.exe: The remote package repository is outdated. You have to choose another repository.
miktexsetup.exe: Data: url="https://ctan.math.illinois.edu/systems/win32/miktex/tm/packages/"
Error: Process completed with exit code -1.

ok, I'll check back in a few hours

StrikerRUS:
Thank you! I'll try to re-run `master` too.

jameslamb:
@StrikerRUS looks like others have been reporting this too :/

https://github.com/MiKTeX/miktex/issues/720

first report on that issue was 16 hours ago, and maintainers there said 

[...]

PR #4840, opened 2021-11-30
[python-package] fix mypy error about missing type hint in dask.py

jameslamb:
Contributes to #3867.

This PR fixes one `mypy` error in `dask.py`.

> python-package/lightgbm/dask.py:933: error: Need type annotation for "out"

```shell
mypy \
    --ignore-missing-imports \
    python-package/
```

### DIFF
--- a/python-package/lightgbm/dask.py
+++ b/python-package/lightgbm/dask.py
@@ -930,7 +930,7 @@ def _predict(
             num_cols = model.n_features_ + 1
 
             nrows_per_chunk = data.chunks[0]
-            out: List[dask_Array] = [[] for _ in range(num_classes)]
+            out: List[List[dask_Array]] = [[] for _ in range(num_classes)]
 
             # need to tell Dask the expected type and shape of individual preds
             pred_meta = data._meta
--- a/python-package/lightgbm/dask.py
+++ b/python-package/lightgbm/dask.py
@@ -957,10 +957,12 @@ def _predict(
             # Concatenate them to return a list of Dask Arrays.
             out_arrays: List[dask_Array] = []
             for i in range(num_classes):
-                out_arrays[i] = dask_array_from_delayed(
-                    value=delayed(concat_fn)(out[i]),
-                    shape=(data.shape[0], num_cols),
-                    meta=pred_meta
+                out_arrays.append(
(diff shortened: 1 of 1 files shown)


jmoralez:
I think the first time `out` is populated in https://github.com/microsoft/LightGBM/blob/b94577d05cd08a6a9147bff16395d442866cb511/python-package/lightgbm/dask.py#L945 it's actually `List[List[dask_Array]]`, where each entry has as many items as there are partitions. Then the contents get overwritten in https://github.com/microsoft/LightGBM/blob/b94577d05cd08a6a9147bff16395d442866cb511/python-package/lightgbm/dask.py#L959 where the partitions get concatenated and the elements become `List[dask_Array]`.

I think variables changing types is a big no-no for `mypy` so maybe we could define two variables here?

jameslamb:
> it's actually`List[List[dask_Array]]`

ooooooo you're right! I missed that `.append()`. Fixed in https://github.com/microsoft/LightGBM/pull/4840/commits/006a271fe8b4d4352b751b385b0b8319a6357684.

> where the partitions get concatenated and the elements become `List[dask_Array]`.
> I think variables changing types is a big no-no for `mypy` so maybe we could define two variables here?

You're generally right about `mypy` not liking it when code overwrites a variable with a value of a different type than it was initialized with. Fixed in https://github.com/microsoft/LightGBM/pull/4840/commits/32f40dec3dc7813ca5649459598bcac3c2150f7c.

jmoralez (approved):

jameslamb:
just re-ran the R-package jobs, let's see if the MiKTeX issue (https://github.com/microsoft/LightGBM/issues/4005#issuecomment-983235895) was fixed overnight

jameslamb:
🎉 CI seems happy again! 

e.g. from [this Windows build](https://github.com/microsoft/LightGBM/runs/4383385794?check_suite_focus=true)

> The downloaded binary packages are in
	C:\Users\runneradmin\AppData\Local\Temp\RtmpwpHjaj\downloaded_packages
Downloading https://github.com/microsoft/LightGBM/releases/download/v2.0.12/miktexsetup-4.0-x64.zip
Setting up MiKTeX
Installing MiKTeX
Done installing MiKTeX

PR #4844, opened 2021-12-02
[python-package] [dask] fix mypy errors from dask.py docstrings

jameslamb:
Contributes to #3867.

This PR fixes a few `mypy` errors in `dask.py`

> python-package/lightgbm/dask.py:1144: error: Item "None" of "Optional[str]" has no attribute "partition"
python-package/lightgbm/dask.py:1327: error: Item "None" of "Optional[str]" has no attribute "partition"
python-package/lightgbm/dask.py:1490: error: Item "None" of "Optional[str]" has no attribute "partition"

```shell
mypy \
    --ignore-missing-imports \
    python-package/
```

### DIFF
--- a/python-package/lightgbm/dask.py
+++ b/python-package/lightgbm/dask.py
@@ -1141,7 +1141,7 @@ class DaskLGBMClassifier(LGBMClassifier, _DaskLGBMModel):
         )
 
     _base_doc = LGBMClassifier.__init__.__doc__
-    _before_kwargs, _kwargs, _after_kwargs = _base_doc.partition('**kwargs')
+    _before_kwargs, _kwargs, _after_kwargs = _base_doc.partition('**kwargs')  # type: ignore
     _base_doc = f"""
         {_before_kwargs}client : dask.distributed.Client or None, optional (default=None)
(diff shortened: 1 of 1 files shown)


python-package/lightgbm/dask.py
```diff
@@ -1141,7 +1141,7 @@ def __init__(
         )
 
     _base_doc = LGBMClassifier.__init__.__doc__
-    _before_kwargs, _kwargs, _after_kwargs = _base_doc.partition('**kwargs')
+    _before_kwargs, _kwargs, _after_kwargs = _base_doc.partition('**kwargs')  # type: ignore
```
jameslamb, line 1144: the warnings mentioned in the description come from the fact that `LGBMClassifier.__init__.__doc__` has type `Optional[str]`.

Proposing that these be ignored, for the reasons described in https://github.com/microsoft/LightGBM/pull/4837#discussion_r758927420

shiyu1994 (approved):
Thank you for working on this. LGTM according to the linked comment.

jmoralez (approved):

PR #4845, opened 2021-12-02
[python][sklearn] unify values of `best_iteration` for sklearn and standard APIs

StrikerRUS:
Do not replace default `0` with `None` in sklearn wrapper.
https://github.com/microsoft/LightGBM/blob/e048a6b409ff72cd2f563b05689ec082bd2e4635/python-package/lightgbm/engine.py#L270

Refer to https://github.com/microsoft/LightGBM/pull/4574#issuecomment-908810538.


jameslamb (approved):
great! Glad to be able to remove a little complexity. Thanks for waiting until 4.0 for this change.

PR #4849, opened 2021-12-03
[ci] migrate CI from macOS 10.14 to 10.15 and drop support of Mojave

### LINKED ISSUE #4797: `macOS-10.14` Azure Pipelines environment is deprecated
> The macOS-10.14 environment is deprecated and will be removed on December 10, 2021. For more details see https://devblogs.microsoft.com/devops/hosted-pipelines-image-deprecation/

We use this environment here
https://github.com/microsoft/LightGBM/blob/6e6fb14cdfdeb4dd0580bfe48f99da1251766bd2/.vsts-ci.yml#L220

The main problem is that that environment is used right now to create release artifacts: LightGBM dylib, SWIG jar and `macosx` Python wheel file.
https://github.com/microsoft/LightGBM/blob/6e6fb14cdfdeb4dd0580bfe48f99da1251766bd2/.vsts-ci.yml#L250-L251

So we should consider this upcoming migration as a possible breaking change: https://github.com/microsoft/LightGBM/pull/4732#issuecomment-955306714.

StrikerRUS:
Closed #4797.

### DIFF
--- a/.ci/setup.sh
+++ b/.ci/setup.sh
@@ -4,7 +4,7 @@ if [[ $OS_NAME == "macos" ]]; then
     if  [[ $COMPILER == "clang" ]]; then
         brew install libomp
         if [[ $AZURE == "true" ]]; then
-            sudo xcode-select -s /Applications/Xcode_9.4.1.app/Contents/Developer || exit -1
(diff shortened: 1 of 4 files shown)


jameslamb (approved):
Thanks for working on this! Given your explanation in #4797, I understand the implications of this change and I'm ok with it.

PR #4852, opened 2021-12-04
[python-package][dask] handle failures parsing worker host names

jameslamb:
Contributes to #3867.

This PR fixes `mypy` errors in `dask.py`

> python-package/lightgbm/dask.py:99: error: Invalid index type "Optional[str]" for "Dict[str, _HostWorkers]"; expected type "str"
python-package/lightgbm/dask.py:383: error: Invalid index type "Optional[str]" for "defaultdict[str, Set[int]]"; expected type "str"

This is `mypy` correctly pointing out the `urlparse(something).hostname` can be `None`, but `lightgbm.dask` treats as if it is guaranteed to be non-None.

```python

print(urlparse("https://127.0.0.1:123").hostname)
# 127.0.0.1

print(urlparse("127.0.0.1:123").hostname)
# None
```

The changes in this PR add explicit handling for that case, so that if it happens `lightgbm` will raise a more specific and helpful error.

Confirmed that this suppresses the error by running the following.

```shell
mypy \
    --ignore-missing-imports \
    python-package/
```

### DIFF
--- a/python-package/lightgbm/dask.py
+++ b/python-package/lightgbm/dask.py
@@ -95,6 +95,8 @@ def _group_workers_by_host(worker_addresses: Iterable[str]) -> Dict[str, _HostWo
     host_to_workers: Dict[str, _HostWorkers] = {}
     for address in worker_addresses:
         hostname = urlparse(address).hostname
+        if not hostname:
+            raise ValueError(f"Could not parse host name from worker address '{address}'")
         if hostname not in host_to_workers:
             host_to_workers[hostname] = _HostWorkers(default=address, all=[address])
         else:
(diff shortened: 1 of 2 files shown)


jmoralez (approved):
LGTM! I think right now the `_group_workers_by_host` function can't get invalid addresses since it gets them from dask, however it's nice to have that function tested in case it ever needs to be used somewhere else.

jameslamb:
> can't get invalid addresses since it gets them from dask

You're right, I don't know any specific case right now where a worker address could make it to this method that doesn't have a `hostname` in the parsing result from `urlparse()`.

But I think `mypy` is right here that it is POSSIBLE for `urlparse(address).hostname` to be `None`, and I think we'll be glad to have this specific error message there to make debugging faster.

Issue #4543, opened 2021-08-21
[RFC] [R-package] Replace "info" interface in lgb.Dataset with keyword arguments


jameslamb:
## Summary

The following changes  should be made to `lgb.Dataset()` in the R package.

"deprecated" = "supported, but raises a warning if used".

**In release 3.3.0 (#4310)**

* [x] deprecate keyword argument `info` in `lgb.Dataset()`
    - https://github.com/microsoft/LightGBM/pull/4573
* [x] add keyword arguments `group`, `weight`, `init_score`, and `label` to `lgb.Dataset()`
    - https://github.com/microsoft/LightGBM/pull/4573
* [x] deprecate passing anything through `...` in `lgb.Dataset()` (https://github.com/microsoft/LightGBM/issues/4226#issuecomment-829473584)
    - https://github.com/microsoft/LightGBM/pull/4573
* [x] deprecate `Dataset$getinfo()` (with a warning its name will be changed to `get_field()`)
* [x] deprecate `Dataset$setinfo()` (with a warning its name will be changed to `set_field()`)
* [x] add method `Dataset$get_field(field_name)` to `Dataset`, matching the Python package
    - https://github.com/microsoft/LightGBM/blob/8a90ea3f267a81a529e3f069cc13e0f6320e7989/python-package/lightgbm/basic.py#L1939
* [x] add method `Dataset$set_field(field_name, data)` should be added to `Dataset`, matching the Python package
    - https://github.com/microsoft/LightGBM/blob/8a90ea3f267a81a529e3f069cc13e0f6320e7989/python-package/lightgbm/basic.py#L1890

**In release 4.0.0**

* [x] remove `...` from `lgb.Dataset()`
    - #4874
* [x] remove all deprecation warnings added for release 3.3.0
    - several issues linked to this one
* [x] remove `Dataset$getinfo()`
    - #4864
* [x] remove `info` from `lgb.Dataset()`
    - #4866
* [x] remove `Dataset$setinfo()`
    - #4854

## Motivation

* reduces maintenance burden by making the R package more closely resemble the Python package
* improves usability, especially for users working in IDEs like RStudio
    - elevating properties to keyword arguments allows for tab-completion and inline documentation references (https://github.com/microsoft/LightGBM/issues/4226#issuecomment-826523570, https://github.com/microsoft/LightGBM/issues/4226#issuecomment-827145419)
[...]

PR #4854, opened 2021-12-04
[R-package] remove Dataset setinfo()

jameslamb:
Contributes to #4543.

This PR proposes removing `Dataset$setinfo()` and the corresponding S3 method. See #4543 for background on why.

### Notes for Reviewers

v3.3.0 and v3.3.1 contain a deprecation warning about this change.


docs/FAQ.rst
```diff
@@ -227,13 +227,7 @@ Run ``lgb.unloader(wipe = TRUE)`` in the R console, and recreate the LightGBM da
 Due to the pointers, choosing to not wipe variables will not fix the error.
 This is a known issue: `Microsoft/LightGBM#698 <https://github.com/microsoft/LightGBM/issues/698>`__.
 
-2. I used ``setinfo()``, tried to print my ``lgb.Dataset``, and now the R console froze!
```
StrikerRUS, line 230: Please don't break references to this FAQ used at third-party sites. Just replace the answer for this question with info about that this is not actual anymore. Refer to https://github.com/microsoft/LightGBM/blob/master/docs/FAQ.rst#1-error-setup-script-specifies-an-absolute-path-when-installing-from-github-using-python-setuppy-install or https://github.com/microsoft/LightGBM/blob/master/docs/FAQ.rst#6-bagging-is-not-reproducible-when-changing-the-number-of-threads as an example.

StrikerRUS (changes requested):

docs/FAQ.rst
jameslamb: ah ok, yes I agree with you. Fixed in https://github.com/microsoft/LightGBM/pull/4854/commits/0770c94801c660e9a49de9d9c730fac66d1fbf67.

I checked on v3.3.0 and latest `master`, and there are no longer issues printing a Dataset (maybe as a result of work in #3016 ).

```r

library(lightgbm)

train <- agaricus.train
print(dtrain)
dtrain$construct()
print(dtrain)
dtrain$setinfo("weight", rep(1.0, nrow(train$data)))
print(dtrain)
```

StrikerRUS (approved):
LGTM! Nice to see this cleanup!

PR #4846, opened 2021-12-02
[python][sklearn] Remove `early_stopping_rounds` argument of `fit()` method

StrikerRUS:
Corresponding RTD build for docstring correctness check: https://lightgbm.readthedocs.io/en/remove_early_stop/Python-API.html.

PR #4855, opened 2021-12-05
[python] remove workaround for `UnboundLocalError` in early stopping callback for Python 2

StrikerRUS:
Python 2 support was dropped for this project in #3581.

With Python 3 we can use [`nonlocal` statement](https://docs.python.org/3/reference/simple_stmts.html#the-nonlocal-statement) instead of wrapping variables into lists as a [workaround](https://stackoverflow.com/a/52667047) for `UnboundLocalError` in nested function scopes.

This workaround was introduced in https://github.com/microsoft/LightGBM/pull/1895#discussion_r240008496 when we had to support Python 2.

Found this during investigating the failures in #4846.

### DIFF
--- a/python-package/lightgbm/callback.py
+++ b/python-package/lightgbm/callback.py
@@ -220,13 +220,19 @@ def early_stopping(stopping_rounds: int, first_metric_only: bool = False, verbos
     best_iter = []
     best_score_list: list = []
     cmp_op = []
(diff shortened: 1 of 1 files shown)


jameslamb (approved):
oh interesting! First time I've seen `nonlocal` before, but your explanation makes sense to me.

PR #4857, opened 2021-12-05
[R-package] remove support for '...' in predict()

jameslamb:
Contributes to #4226.

This PR removes support for passing anything through `...` in `predict.lgb.Booster()` and `Booster$predict()`.

### Notes for Reviewers

v3.3.0 and v3.3.1 contain a deprecation warning about this change.

### DIFF
--- a/R-package/R/lgb.Booster.R
+++ b/R-package/R/lgb.Booster.R
@@ -764,6 +764,7 @@ Booster <- R6::R6Class(
 #'               \href{https://lightgbm.readthedocs.io/en/latest/Parameters.html#predict-parameters}{
 #'               the "Predict Parameters" section of the documentation} for a list of parameters and
 #'               valid values.
+#' @param ... ignored
 #' @return For regression or binary classification, it returns a vector of length \code{nrows(data)}.
 #'         For multiclass classification, either a \code{num_class * nrows(data)} vector or
 #'         a \code{(nrows(data), num_class)} dimension matrix is returned, depending on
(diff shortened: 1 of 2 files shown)


R-package/R/lgb.Booster.R
```diff
@@ -828,23 +815,12 @@ predict.lgb.Booster <- function(object,
                                 predcontrib = FALSE,
                                 header = FALSE,
                                 reshape = FALSE,
-                                params = list(),
-                                ...) {
+                                params = list()) {
```
jameslamb, line 818: It looks like `...` cannot be removed from `predict.lgb.Booster()`, since the `predict()` generic in base R has a `...`.

From `R CMD check` ([build link](build link: https://github.com/microsoft/LightGBM/runs/4420930711?check_suite_focus=true))

```text
* checking S3 generic/method consistency ... WARNING
predict:
  function(object, ...)
predict.lgb.Booster:
  function(object, data, start_iteration, num_iteration, rawscore,
           predleaf, predcontrib, header, reshape, params)

See section ‘Generic functions and methods’ in the ‘Writing R
Extensions’ manual.
```

From https://cran.r-project.org/doc/manuals/R-exts.html#Generic-functions-and-methods

> A method must have all the arguments of the generic, including … if the generic does.

jameslamb: Add a warning and updated docs in https://github.com/microsoft/LightGBM/pull/4857/commits/49c8368124f57a5946b226043052efc10faa316e, so that at least users are notified that `...` is not intended for use with any `{lightgbm}` parameters.

StrikerRUS (approved):
LGTM! Thanks for the explanation of generic `predict()` function.

PR #4858, opened 2021-12-05
[R-package] small simplifications in Booster and Predictor

jameslamb:
Working through #4857, I noticed a few functions in the R package doing something like this:

```r
some_function <- function(x = NULL) {
    if (is.null(x)) {
        x <- -1L
    }
}
```

This PR proposes simplifying that code by just changing the default values in function signatures.

```r
some_function <- function(x = -1L){
}
```

### Notes for Reviewers

Function defaults are lazily evaluated in R, so no memory is saved by using `NULL` instead of an integer.

```r
# this does not fail until the function is called
do_something <- function(m = stop("hey")){
     return TRUE
}
```

Function defaults are not stored in the global environment, so it isn't necessary to worry about side effects [like you have to with using mutable types as function defaults in Python](https://stackoverflow.com/a/26320938/3986677).

```r
do_something <- function(x = 10L) {
    x <- x + 1L
    print(x)
}

do_something()
# 11

do_something()
# 11

do_something()
# 11
```

### DIFF
--- a/R-package/R/lgb.Booster.R
+++ b/R-package/R/lgb.Booster.R
@@ -497,7 +497,7 @@ Booster <- R6::R6Class(
 
     # Predict on new data
     predict = function(data,
-                       start_iteration = NULL,
+                       start_iteration = 0L,
                        num_iteration = NULL,
                        rawscore = FALSE,
                        predleaf = FALSE,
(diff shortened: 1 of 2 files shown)


StrikerRUS (approved):
LGTM!
Are these functions in public API? If so, this PR should be marked as `breaking` because it drops possibility of using `NULL` at a user side.

jameslamb:
> Are these functions in public API? If so, this PR should be marked as breaking because it drops possibility of using `NULL` at a user side.

Ah good point! I was so focused on this being the mechanism for the default, I didn't think about users actually passing `NULL` explicitly.

Given that, I don't support these changes any more 😬 . I don't think this very small simplification is worth breaking users' code. I'm going to close this.

StrikerRUS:
OK, no problem! 🙂

Issue #3625, opened 2020-12-04
Categorical features in sklearn API warning


david26694:
## How you are using LightGBM?

LightGBM component: Python package

## Environment info

Operating System: macOS Catalina Version 10.15.2 (19C57)

CPU/GPU model: 1,4 GHz Quad-Core Intel Core i5

C++ compiler version: Apple clang version 11.0.3 (clang-1103.0.32.59)

Python version: 3.7

## Error message and / or logs

I use the `categorical_feature` option in `LGBMRegressor` and I get the following warning:

```UserWarning: Using categorical_feature in Dataset. warnings.warn('Using categorical_feature in Dataset.')```

Is this expected? Do you know why is this happening, what does it mean?

## Reproducible example(s)

```python
import pandas as pd
from sklearn.datasets import load_iris
from lightgbm import LGBMRegressor
from category_encoders import OrdinalEncoder

X = load_iris()['data']
y = load_iris()['target']

X = OrdinalEncoder(cols=[3]).fit_transform(X)

dt = LGBMRegressor()
dt.fit(X, y, categorical_feature=[3])
```

StrikerRUS:
Yes, it is expected. This warning was originally introduced to make users sure that their categorical features were not lost.

Please refer to #792.

And we already have a request to refactor group of warnings related to cat. features: #3379. Feel free to comment there and share your opinion on how these warnings can be improved.

Issue #3145, opened 2020-06-03
Lightgbm data in R 


mgzde:
Hi everyone,

I'm using lightgbm package in my Windows 10 computer, RStudio 3.6.3 version. I also asked how can I install lightgbm for R a few weeks ago then it was initiliazed successfuly (I guess)with help but when I run the given lightgbm data(bank dataset #887) it gives me a model;
type: NULL 
length:0 
size:0B 
value:NULL(empty).
 Also, I run lightgbm for my 10 years monthly univariate time series data, it also gives same output, empty. I don't know the reason for this issue, may it have been installed incorrectly or something went wrong? 

Thank you

jameslamb:
@mgzde Can you please provide the following information:

1. How did you install the LightGBM R package? Descibe this in as much detail as possible, including the version or commit hash used.
2. What is the exact code you ran?

mgzde:
I've installed git and CMake(3.16.6)
I've installed Visual Studio Code (April 2020 version) (Actually, I want to use lightgbm in RStudio, so I have RStudio 3.6.3)
I've also installed RTools35
I've added C:\Program Files\Rtools\bin to path
Then I've run the following code
git clone --recursive https://github.com/microsoft/LightGBM
cd LightGBM
Rscript build_r.R
![Ekran Görüntüsü (73)](https://user-images.githubusercontent.com/65132187/83756076-2dec4780-a677-11ea-989d-19a99327790e.png)
![Ekran Görüntüsü (74)](https://user-images.githubusercontent.com/65132187/83756068-2c228400-a677-11ea-8c8d-aeef5ffa1bbd.png)
![Ekran Görüntüsü (75)](https://user-images.githubusercontent.com/65132187/83756070-2d53b100-a677-11ea-896d-c5de74334808.png)
![Ekran Görüntüsü (76)](https://user-images.githubusercontent.com/65132187/83756074-2d53b100-a677-11ea-897a-e8a3471323a6.png)
![Ekran Görüntüsü (77)](https://user-images.githubusercontent.com/65132187/83756075-2dec4780-a677-11ea-92a3-a131341443cc.png)

Then I've tried in R with the following codes and outputs for agaricus data;
![Ekran Görüntüsü (111)](https://user-images.githubusercontent.com/65132187/83756961-76583500-a678-11ea-885b-94696a8e4b42.png)
[...]

PR #4868, opened 2021-12-07
[python] reset storages in early stopping callback after finishing training

StrikerRUS:
Right now `early_stopping()` callback cannot be used with scikit-learn's `GridSearchCV` tool due to that storages for metric results are not cleaned after early stopping occurs. This results in **comparing current results with the best iteration for the first grid iteration for the first fold**.

This was spotted by our `test_grid_search` test
https://github.com/microsoft/LightGBM/blob/fe535a0e9dfb251fcc85dd33fb0bf193d022fe90/tests/python_package_test/test_sklearn.py#L276
which fails with early stopping setup passed via callback with the following error
```
>       assert grid.best_estimator_.best_score_['valid_0']['multi_logloss'] < 0.25
E       assert 0.5915701709051111 < 0.25

../tests/python_package_test/test_sklearn.py:318: AssertionError
```

This PR proposes resetting all global variables in the `_init()` helper function and determining whether initialization is required by the special variable `inited` instead of checking `cmp_op` global variable for emptiness, which results right now in that `_init()` is called only once during the whole grid search routine.

Extracted from #4846.

### DIFF
--- a/python-package/lightgbm/callback.py
+++ b/python-package/lightgbm/callback.py
@@ -221,6 +221,7 @@ def early_stopping(stopping_rounds: int, first_metric_only: bool = False, verbos
     best_score_list: list = []
     cmp_op = []
     enabled = True
+    inited = False
     first_metric = ''
 
     def _init(env: CallbackEnv) -> None:
@@ -229,6 +230,7 @@ def early_stopping(stopping_rounds: int, first_metric_only: bool = False, verbos
         nonlocal best_score_list
         nonlocal cmp_op
         nonlocal enabled
+        nonlocal inited
         nonlocal first_metric
         enabled = not any(env.params.get(boost_alias, "") == 'dart' for boost_alias
                           in _ConfigAliases.get("boosting"))
@@ -241,6 +243,14 @@ def early_stopping(stopping_rounds: int, first_metric_only: bool = False, verbos
 
         if verbose:
             _log_info(f"Training until validation scores don't improve for {stopping_rounds} rounds")
+
+        # reset storages
+        best_score = []
+        best_iter = []
+        best_score_list = []
+        cmp_op = []
+        inited = True
+        first_metric = ''
 
         n_metrics = len(set(m[1] for m in env.evaluation_result_list))
         n_datasets = len(env.evaluation_result_list) // n_metrics
@@ -283,6 +293,7 @@ def early_stopping(stopping_rounds: int, first_metric_only: bool = False, verbos
     def _final_iteration_check(env: CallbackEnv, eval_name_splitted: List[str], i: int) -> None:
         nonlocal best_iter
         nonlocal best_score_list
+        nonlocal inited
         if env.iteration == env.end_iteration - 1:
             if verbose:
                 best_score_str = '\t'.join([_format_eval_result(x) for x in best_score_list[i]])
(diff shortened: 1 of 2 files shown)


python-package/lightgbm/callback.py
```diff
@@ -283,13 +293,15 @@ def _init(env: CallbackEnv) -> None:
...
         nonlocal best_score_list
+        nonlocal inited
         if env.iteration == env.end_iteration - 1:
             if verbose:
                 best_score_str = '\t'.join([_format_eval_result(x) for x in best_score_list[i]])
                 _log_info('Did not meet early stopping. '
                           f'Best iteration is:\n[{best_iter[i] + 1}]\t{best_score_str}')
                 if first_metric_only:
                     _log_info(f"Evaluated only: {eval_name_splitted[-1]}")
+            inited = False
```
shiyu1994, line 304: It seems that `inited` is reset to `False` only when early stopping is not triggered? Because
1. `inited` is reset to `False` only when `env.iteration == env.end_iteration - 1`
2. `_final_iteration_check` is not called when early stopping is triggered according to
https://github.com/microsoft/LightGBM/blob/00f87c52ec24b1f8358c230646685459e31b60b8/python-package/lightgbm/callback.py#L306-L327

shiyu1994 (commented):
Thank you for working on this. Just leave a comment for now since I'm not sure the `inited` is reset correctly.

python-package/lightgbm/callback.py
StrikerRUS: @shiyu1994 Thanks for your review!

> It seems that `inited` is reset to `False` only when early stopping is not triggered?

I believe `inited` is reset to `False` when early stopping is triggered here
https://github.com/microsoft/LightGBM/blob/dff622c26aceb5bc602a6279d4a45d7e99d389e5/python-package/lightgbm/callback.py#L339-L340
Triggering early stopping means to raise `EarlyStopException` and `inited` is reset to `False` right before that line.

Also, if it is not reset, the test `test_grid_search()` will fail because we use constant custom evaluation metric there which force early stopping to happen at `stopping_rounds` iteration because there is no improvement after the first iteration.

shiyu1994:
@StrikerRUS Thanks for the explanation. Sorry that I did not notice line 339.

shiyu1994 (approved):
The changes LGTM. Thank you.

jameslamb (approved):
thanks very much for the explanation! I think this fix makes sense.

PhillipMaire:
is this going to be merged to master? having a working grid search and other related packaged like OPTUNA work with light GBM is a must for many people and without this update they still don't work properly with light GBM!! from [OPTUNA #3145](https://github.com/optuna/optuna/issues/3145#issuecomment-1118012232) "LightGBM 3.3.2, which has been released yesterday, does not include the change. We need to wait for the fix..." and recently [OPTUNA #3625](https://github.com/optuna/optuna/issues/3625)

jameslamb:
> is this going to be merged to `master`

This PR was merged on December 9, 2021. I think you mean "when will this be released to package managed like PyPI".

Due to a lack of maintainer activity, it will probably still be several months until the next release of LightGBM. You can subscribe to #5153 for updates on the next release, and even comment on the linked issues if there are any you'd like to contribute, to move the project closer to that release.

PhillipMaire:
ahh yes thank you for clarifying this! I installed it on my [test notebook](https://colab.research.google.com/drive/1LigmaNSoK_FFi1BYTwhaDHZEBiLZskz0#scrollTo=wZQXbecNuArq) and it seems to run fine but I get an error with OPTUNA which I mentioned to them. I installed using the following in case anyone else needs this 
```
!git clone --recursive https://github.com/microsoft/LightGBM.git
!cd LightGBM/python-package && python setup.py install
```

PR #4871, opened 2021-12-09
[R-package] remove unused callback cb.reset.parameter

jameslamb:
Proposes removing the callback function `cb.reset.parameter()` in the R package's code.

### Notes for Reviewers

All callback functions in the R package are currently unexported (see #2479). Any that are unused can be safely removed without affecting users.

`cb.reset.parameter()` is unused in the package. It was added in the very first PR introducing the R package (https://github.com/microsoft/LightGBM/pull/168), I think probably just because the same callback existed in the Python package.

This PR proposes removing it to simplify the R package's code.


StrikerRUS:
Why remove but not expose (as per #2479)?

jameslamb:
> Why remove but not expose (as per #2479)?

The code for `cb.reset.parameter()` was added to this project around 5 years ago and hasn't been tested or used at any point during that time. So I don't know if it actually works with the current state of LightGBM, or even if it ever worked.

I think that completing #2479 will require a medium-sized effort...it will involve adding documentation and tests, for example. I'm personally not ready to put in that effort yet, as I'm prioritizing some other work on this project higher.

So I thought "as long as the callbacks are not actually accessible to users, might as well remove this one that has never been used, to simplify the package's code".

StrikerRUS:
>  So I don't know if it actually works with the current state of LightGBM, or even if it ever worked.

Ah, OK. My point was to preserve callbacks code with the aim to not reimplement them again when expose to public API. But thanks to GitHub cross-references, one still can copy-paste remove code from this PR if needed.

StrikerRUS (approved):
LGTM!

Issue #789, opened 2017-08-06
[CLI] Categorical: Read string and convert to int on the fly


AbdealiJK:
This is a feature request

It would be very useful to be able to read in Categorical values as strings ("abc", "def", etc) and convert that to integers internally.

This adds a bit of overhead, but would be much easier for users. There would probably be some overhead to do this, so a flag can be made if the user wants to do it, or we can automatically check if the column has any alphabet or not

limexp:
@AbdealiJK, @guolinke,

There are two points to consider.

1. Reading data. Is using categorical_column parameter is required to be filled? Or feature must became categorial as soon as any non-numeric value is found? I'm quite sure that decision must be based on categorial_columns parameter. Also string representations (like 'NA' or 'nan') for missing values must be defined. 

2. Saving and exporting. We can save additional block with strings to integers mapping. Or we can try to use intial string values. The latter approach could lead to many errors (for example, with legacy code).

AbdealiJK:
I think:
1. categorical_column would be the better and simpler approach to do it
1. string to integer mapping is the best way to do it. Using initial string values throughout would require a lot of refactoring and also integer mapping is more memory efficient

limexp:
@AbdealiJK
2. I didn't suggest to refactor everything and add strings to the kernel. The question was about representation of split value in the saved model. It could be integer index (with mapping saved apart) or a string value itself.

Laurae2:
@limexp string value would increase model size and is not efficient, it can also cause major issues (not including the potential name clashes and what to do with characters which are not compliant to the text format used by LightGBM)

limexp:
@Laurae2, I totally agree. And each decision has its pros and cons.
This feature would be great for CLI, so data can be used without preprocessing. Is it really important for python or R interfaces? 
In any case we lose control over NaN values.

Laurae2:
@limexp R and Python already have the conversion from categoricals to integers.

But if the model deployment in production is done using CLI, one must create a script to convert categoricals to their appropriate integers (usually done with SQL or any other data warehousing software).

dah33:
I was using categorical features on a Kaggle Kernel. Python converts NaNs to "-1" when you convert values within categoricals to integers. This causes LightGBM to bomb out with a fatal error. This is silent on Kaggle, and kills the kernel.

The solution is:

`for col in categorical_vars:
    df[col] = pd.Categorical(df[col].cat.codes+1)`

guolinke:
[...]

PR #4876, opened 2021-12-10
Strings as Categorical Values

yanchaomei:
The decision tree of lightgbm can only support the classification of integer labels. For string type data labels, users need to preprocess the labels themselves, convert the string type to integer type, and then use lightgbm.

We modify and support the direct input of string type data, establish a string to double mapping when reading the file in the training process, and then save it to the model. Restore the map from the model during the prediction process;


include/LightGBM/dataset.h
```diff
@@ -13,6 +13,8 @@
 #include <LightGBM/utils/random.h>
 #include <LightGBM/utils/text_reader.h>
 
+// #include "../../src/io/parser.hpp"
```
hippogr, line 16: >/ #include "../../src/io/parser.hpp" [](http://example.com/codeflow?start=1&length=36)

if you don't need this line, you can simply remove it instead of commenting it. :-)

include/LightGBM/dataset.h
```diff
@@ -739,6 +754,36 @@ class Dataset {
...
+        name = name + iter->first + "=" + std::to_string(iter->second) + '\n'; 
+        iter++;
+      }
+      name += '\n';
+      mapping.push_back(name);
+    }
+  }
+
+
+  inline std::vector< std::string >  get_mapping() const {
```
hippogr, line 783: > std::string  [](http://example.com/codeflow?start=21&length=13)

Looks like that we don't need the space between std::string

include/LightGBM/dataset.h
```diff
@@ -24,6 +26,19 @@
 
 namespace LightGBM {
 
+class Str2Num {
```
hippogr, line 29: >class [](http://example.com/codeflow?start=0&length=5)

since all of the members are public, so how about change it to structure instead of class?

include/LightGBM/dataset.h
```diff
@@ -781,6 +826,7 @@ class Dataset {
   std::vector<int> numeric_feature_map_;
   int num_numeric_features_;
   std::string parser_config_str_;
+  std::vector< std::string > mapping;
```
hippogr, line 829: >< std::string  [](http://example.com/codeflow?start=13&length=14)

same

include/LightGBM/dataset.h
```diff
@@ -739,6 +754,36 @@ class Dataset {
     return raw_data_[numeric_feature_map_[feat_ind]].data();
   }
 
+  inline void store_mapping()  {
```
hippogr, line 757: >   [](http://example.com/codeflow?start=29&length=2)

single space is OK

include/LightGBM/dataset.h
```diff
@@ -739,6 +754,36 @@ class Dataset {
...
+    // store maps to mapping
+    int offset = 0;
+    for(int i=0; i < maps.size(); i++){
```
hippogr, line 760: >for(i [](http://example.com/codeflow?start=4&length=5)

for<space>{

include/LightGBM/dataset.h
```diff
@@ -739,6 +754,36 @@ class Dataset {
     return raw_data_[numeric_feature_map_[feat_ind]].data();
   }
 
+  inline void store_mapping()  {
+    // store maps to mapping
+    int offset = 0;
+    for(int i=0; i < maps.size(); i++){    
+      std::string name;
+      
+      if(i == label_idx_){
```
hippogr, line 763: > if( [](http://example.com/codeflow?start=5&length=4)

same one, please review the other code to make sure you are following the code style

include/LightGBM/dataset.h
```diff
@@ -739,6 +754,36 @@ class Dataset {
     return raw_data_[numeric_feature_map_[feat_ind]].data();
   }
 
+  inline void store_mapping()  {
+    // store maps to mapping
+    int offset = 0;
+    for(int i=0; i < maps.size(); i++){    
+      std::string name;
```
hippogr, line 761: >std::string name; [](http://example.com/codeflow?start=6&length=17)

how about move this one out of loop?

include/LightGBM/dataset.h
```diff
@@ -739,6 +754,36 @@ class Dataset {
...
+        name = "label";
+        offset--;
+      }else{ 
+        name = feature_names_[i + offset];
+        }
+        
+      name = name + '\n';
+      std::map<std::string, double>::iterator iter;
+      iter = maps[i].Str2NumMap.begin();
+      while(iter != maps[i].Str2NumMap.end()){
```
hippogr, line 773: why not 
`for (auto i = maps[i].Str2NumMap..begin(); i != maps[i].Str2NumMap..end(); ++i)`?

include/LightGBM/dataset.h
```diff
@@ -739,6 +754,36 @@ class Dataset {
...
+        name = name + iter->first + "=" + std::to_string(iter->second) + '\n'; 
+        iter++;
+      }
+      name += '\n';
+      mapping.push_back(name);
```
hippogr, line 778: >push_back(name) [](http://example.com/codeflow?start=14&length=15)

push_back(std::move(name)) is better. You can check the other places to see if you can use move to avoid additional costs.

include/LightGBM/utils/common.h
```diff
@@ -31,6 +31,7 @@
 #include <unordered_map>
 #include <utility>
 #include <vector>
+#include <map>
```
hippogr, line 34: is this the only line to change it in this file? I suggest that you add this include in the cpp or header when it uses <map> instead of adding it in the common header

include/LightGBM/dataset.h
```diff
@@ -739,6 +754,36 @@ class Dataset {
...
+      std::string name;
+      
+      if(i == label_idx_){
+        name = "label";
+        offset--;
+      }else{ 
+        name = feature_names_[i + offset];
+        }
+        
+      name = name + '\n';
```
hippogr, line 770: >name = name + '\n'; [](http://example.com/codeflow?start=6&length=19)

I see you are using += when you process string, which is good. But why not this one? :-)

src/boosting/gbdt_model_text.cpp
```diff
@@ -621,6 +635,35 @@ bool GBDT::LoadModelFromString(const char* buffer, size_t len) {
...
+        is_inmaps = true;
+      } else if (cur_line == std::string("end of Maps")) {
+        p += line_len;
+        p = Common::SkipNewLine(p);
+        break;
+      } else if(cur_line == std::string("map")){
+        index++;
+        tmp.Str2NumMap.clear();
+      }
+       else if (is_inmaps) {
```
hippogr, line 655: > else [](http://example.com/codeflow?start=6&length=5)

keep the same style as the code above

src/boosting/gbdt_model_text.cpp
```diff
@@ -621,6 +635,35 @@ bool GBDT::LoadModelFromString(const char* buffer, size_t len) {
...
+         std::vector< std::string > key_value;
+         key_value = Common::Split(cur_line.c_str(), "=");
+         
+         tmp.Str2NumMap.insert(std::pair < std::string , double > (key_value[0], std::stod(key_value[1]) ) );
```
hippogr, line 659: >) ) ) [](http://example.com/codeflow?start=103&length=5)

same question about the style

src/boosting/gbdt_model_text.cpp
```diff
@@ -621,6 +635,35 @@ bool GBDT::LoadModelFromString(const char* buffer, size_t len) {
...
+        p += line_len;
+        p = Common::SkipNewLine(p);
+        break;
+      } else if(cur_line == std::string("map")){
+        index++;
+        tmp.Str2NumMap.clear();
+      }
+       else if (is_inmaps) {
+         std::vector< std::string > key_value;
+         key_value = Common::Split(cur_line.c_str(), "=");
```
hippogr, line 657: I guess here you need some checks. One is that you may need to check if the key_value vector has two elements so your accessing key_value[1] won't bring exceptions. Another one is to do check if std::stod throws exception or not.

src/io/dataset_loader.cpp
```diff
@@ -17,6 +18,8 @@ namespace LightGBM {
 
 using json11::Json;
 
+extern std::vector<Str2Num> maps;
```
hippogr, line 21: why do you need this?

src/boosting/gbdt_model_text.cpp
```diff
@@ -16,6 +17,8 @@
 
 namespace LightGBM {
 
+extern std::vector<Str2Num> maps;
+
```
hippogr, line 21: you don't really need this, right?

src/io/parser.cpp
```diff
@@ -12,6 +12,8 @@
 
 namespace LightGBM {
 
+std::vector<Str2Num> maps;
```
hippogr, line 15: same, seems like you don't need this

src/io/parser.hpp
```diff
@@ -15,6 +15,135 @@
 
 namespace LightGBM {
 
+// define a str2num map class
+
+extern std::vector<Str2Num> maps;
+
+const char* Atof_and_map(const char* p, double* out, int idx) {
```
hippogr, line 22: >Atof_and_map [](http://example.com/codeflow?start=12&length=12)

I guess you don't need to write a string to float function by yourself. How about read out the string, trim the spaces and tabs, then check if it is a nan or infinite, then you can use the API to do that. I don't see any necessary to implement a str2num here

tests/python_package_test/test_string.py
```diff
@@ -0,0 +1,102 @@
...
+import numbers
+from pathlib import Path
+
+import numpy as np
+# import pytest
+from scipy import sparse
+from sklearn.datasets import dump_svmlight_file, load_svmlight_file, load_breast_cancer
+from sklearn.model_selection import train_test_split
+import sys
+sys.path.append("python-package")
```
jameslamb, line 12: Please do not modify `sys.path`, as that could lead to a situation where some imports seem to work during tests and then actually fail when users run `lightgbm` on their systems.

tests/python_package_test/test_string.py
```diff
@@ -0,0 +1,102 @@
...
+        "pred_early_stop_freq": 5,
+        "pred_early_stop_margin": 1.5
+    }
+    pred_early_stopping = bst.predict(X_test, **pred_parameter)
+    # scores likely to be different, but prediction should still be the same
+    np.testing.assert_array_equal(np.sign(pred_from_matr),
+                                  np.sign(pred_early_stopping))
+
+
+test_basic(Path(r'tests\python_package_test\test_string'))
```
jameslamb:
```suggestion
```

It is not necessary to call test functions like this. `pytest` (the test runner used in LightGBM), will automatically detected and run all functions whose names start with `test_`.

Please remove this line, and ensure that every file has a newline at the end of it.

src/c_api.cpp
```diff
@@ -728,6 +728,7 @@ class Booster {
 
   std::string SaveModelToString(int start_iteration, int num_iteration,
                                 int feature_importance_type) const {
+
```
jameslamb, line 731: Please revert this and any other whitespace-only changes not related to this PR. Such changes increase the size of the diff and therefore the effort required to review.

src/io/dataset_loader.cpp
```diff
@@ -1172,7 +1175,7 @@ void DatasetLoader::ConstructBinMappersFromTextData(int rank, int num_machines,
   if (dataset->has_raw()) {
     dataset->ResizeRaw(static_cast<int>(sample_data.size()));
   }
-
+
```
jameslamb, line 1178:
```suggestion
```

src/io/dataset_loader.cpp
```diff
@@ -1364,6 +1368,7 @@ void DatasetLoader::ExtractFeaturesFromFile(const char* filename, const Parser*
       OMP_LOOP_EX_END();
     }
     OMP_THROW_EX();
+
```
jameslamb, line 1371:
```suggestion
```

tests/cpp_tests/test.conf
```diff
@@ -0,0 +1,171 @@
+# task type, support train and predict
+task = train
+
+
```
jameslamb, line 4:
```suggestion
```

Please avoid using more than one consecutive blank line in these `.conf` and `.config` files, to make them easier to read.

jameslamb (changes requested):
This PR currently is proposing adding several files to source control for this project which I don't approve of adding to this repo. Please remove them or explain why you believe they should be added.

* `LightGBM.pyproj`
* `CMakeSettings.json`
* `windows/LightGBM_model.txt`
* `windows/LightGBM_predict_result.txt`

I've also left several other comments. Please don't include whitespace-only changes or commented-out code in this pull request.

jameslamb:
Given that there hasn't been any activity on this PR in 3+ months and the CLA has not been signed, it seems to me that this PR has been abandoned. I'm going to close this.

If anyone reading this is interested in picking up this work for #789, please open a new pull request.

PR #4878, opened 2021-12-10
[python] remove `verbose_eval` argument of `train()` and `cv()` functions

StrikerRUS:
Refer to #4574.


jameslamb (approved):
Love these simplifications!

Issue #4862, opened 2021-12-06
[R-package] [tests] reduce verbosity in unit tests


jameslamb:
## Summary

Unit tests in the R package which are not explicitly testing `{lightgbm}`'s behavior for different values of the parameter `verbose` should use `verbose = -1L` to suppress all other log outputs.

## Motivation

Suppressing more log output in tests may make them faster by reducing the number of times the package has to write to stdout / stderr.

It will also improve the usability of the logs for quickly identifying where issues happened.

## Description

This work should be done in multiple small pull requests, not one huge one changing all tests.

It should be possible to override the value of `verbose` should be using an environment variable, so that developers can occasionally, easily run the tests in "debug" mode if additional logs are needed to identify issues.

Maybe like this:

```r
LGB_TEST_VERBOSITY <- as.integer(Sys.getenv("LGB_TEST_VERBOSITY", "-1"))

test_that("some test", {
    params = list(verbose = LGB_TEST_VERBOSITY)
    ...
})
```

## References

Created from this comment: https://github.com/microsoft/LightGBM/pull/4567#pullrequestreview-741025297

PR #4879, opened 2021-12-11
[R-package] reduce verbosity in some unit tests

jameslamb:
Contributes to #4862.

This PR proposes the first batch of changes to reduce the verbosity of the R package tests. I only changed a few tests, as this PR is mainly focused on getting a review of the proposed approach.

### How I Tested This

I used the following code to determine the impact of this change.

```shell
R CMD INSTALL --with-keep.source ./lightgbm_3.3.1.99.tar.gz
cd R-package/tests
```

On `master`, the R tests generate 3065 lines of logs.
As of this PR, that's reduced to 2984.

### Why allow changing this with an environment variable?

I think it would be useful to occasionally see all LightGBM logs from tests.

That can help to catch issues like #4225 that don't show up as test errors.

### DIFF
--- a/R-package/tests/testthat/test_lgb.plot.interpretation.R
+++ b/R-package/tests/testthat/test_lgb.plot.interpretation.R
@@ -87,7 +87,7 @@ test_that("lgb.plot.interepretation works as expected for multiclass classificat
         params = params
         , data = dtrain
         , nrounds = 3L
-        , verbosity = VERBOSITY
+        , verbose = VERBOSITY
     )
     num_trees <- 5L
     tree_interpretation <- lgb.interprete(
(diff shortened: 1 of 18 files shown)


R-package/tests/testthat/test_lgb.plot.interpretation.R
```diff
@@ -82,6 +87,7 @@ test_that("lgb.plot.interepretation works as expected for multiclass classificat
         params = params
         , data = dtrain
         , nrounds = 3L
+        , verbosity = VERBOSITY
```
jameslamb, line 90:
```suggestion
        , verbose = VERBOSITY
```

StrikerRUS (approved):
LGTM! I like this approach, thanks!

PR #4880, opened 2021-12-11
[python] allow to register any custom logger (fixes #4783)

### LINKED ISSUE #4783: Relax constraint on logger class in `register_logger`
## Summary

Allow any logger instance to be registered via `register_logger`, as long as it has `info` and `warning` methods defined.

## Motivation

Currently only instance of `logging.Logger` or its subclass can be registered. I'd like to use logger from [`loguru`](https://github.com/Delgan/loguru), which is more convenient, however it will trigger `TypeError` now. Since only `info` and `warning` are used to log messages, any logger class providing these two methods would suffice.

## Description

Instead of checking the class of logger, just check the existence of `info` and `warning` method.

Reference implementation:
```python
def register_logger(logger) -> None:
    def _has_method(method):
        return callable(getattr(logger, method, None))

    if not _has_method("info") or not _has_method("warning"):
        raise TypeError("Logger must provide 'info' and 'warning' method")
    global _LOGGER
    _LOGGER = logger
```

## References

N/A

RustingSword:
Allow to register any custom logger, as long as it provides `info`, `warning` and `error` method. Refer to #4783.


python-package/lightgbm/basic.py
```diff
@@ -41,20 +41,26 @@ def info(self, msg: str) -> None:
...
     Parameters
     ----------
-    logger : logging.Logger
+    logger : Any
         Custom logger.
     """
-    if not isinstance(logger, Logger):
-        raise TypeError("Logger should inherit logging.Logger class")
+    def _has_method(method_name):
+        return callable(getattr(logger, method_name, None))
```
jameslamb, line 60:
```suggestion
    def _has_method(logger: Any, method_name: str) -> bool:
        return getattr(logger, method_name, None) is not None
```

I'd expect a function called `_had_method()` to return a boolean, and I'd prefer if it didn't reach up into the calling environment.

Would you consider this format?

jameslamb (changes requested):
Thanks very much for this!

I've added `(fixes #4783)`  to the pull request title here so GitHub will automatically close #4783 if this is merged, and so the connection between this PR and that issue is clear to anyone reading the release notes (PR titles become release note items in this project, [see here for example](https://github.com/microsoft/LightGBM/releases/tag/v3.3.1)).

Can you please add a unit test covering this new behavior, to be sure it isn't accidentally broken by future refactorings? See the diff from https://github.com/microsoft/LightGBM/pull/3820/files#diff-57a0d4d593bc1ebac04a99d2e477a178c209de8c50e6cffdb5f9195b26dd6e68 for an example.

RustingSword:
Thanks for the thorough guidance! I considered updating the unit test, however not sure what to test. How about add test to make sure the method check is satisfied? Maybe something like this:

```python
def test_register_invalid_logger():
    class LoggerWithoutInfoMethod:
        def warning(self, msg: str) -> None:
            print(msg)

        def error(self, msg: str) -> None:
            print(msg)

    class LoggerWithoutWarningMethod:
        def info(self, msg: str) -> None:
            print(msg)

        def error(self, msg: str) -> None:
            print(msg)

    class LoggerWithoutErrorMethod:
        def info(self, msg: str) -> None:
            print(msg)

        def warning(self, msg: str) -> None:
            print(msg)

    def _test_raise_type_error(invalid_logger) -> None:
        try:
            lgb.register_logger(invalid_logger)
        except Exception as err:
            assert isinstance(err, TypeError)
            assert str(err) == "Logger must provide 'info', 'warning' and 'error' method"

    _test_raise_type_error(LoggerWithoutInfoMethod())
    _test_raise_type_error(LoggerWithoutWarningMethod())
    _test_raise_type_error(LoggerWithoutErrorMethod())
```

python-package/lightgbm/basic.py
jameslamb: oh! I just realized that `callable()` is checking whether or not that attribute obtained with `getattr()` exists and is callable! Sorry, that was my mistake reading this code.

I think just passing in `logger` and adding the `->` bool type hint is required.

```suggestion
    def _has_method(logger: Any, method_name: str) -> bool:
        return callable(getattr(logger, method_name, None))
```

jameslamb:
> not sure what to test

I think tests covering the following would give us confidence that this feature is working:

* given a logger class that implements `info()`, `warning()`, and `error()` but does not inherit from `logging.Logger`, that class's logging methods are used by correctly and successfully by `lightgbm`
* given a class that has properties `info`, `warning`, and `error` set to something non-callable (like integers), `lightgbm` throws the error introduced in this PR
* given a class that is missing `info()`, `warning()`, or `error()`, `lightgbm` throws the error introduced in this PR

For tests that check for errors, please use `pytest.raises()` for that instead of implementing `_test_raise_type_error()`, see the following for an example.

https://github.com/microsoft/LightGBM/blob/80662618996431c222e6c92cbba7e784f8a935cd/tests/python_package_test/test_dask.py#L481

RustingSword:
> * given a logger class that implements `info()`, `warning()`, and `error()` but does not inherit from `logging.Logger`, that class's logging methods are used by correctly and successfully by `lightgbm`

Is there any way to directly test the logging utility in `lightgbm`, or do I need to test it like https://github.com/microsoft/LightGBM/pull/3820/files#diff-57a0d4d593bc1ebac04a99d2e477a178c209de8c50e6cffdb5f9195b26dd6e68 to go through the training process? If so, I'm afraid there will be a large portion of duplicate code, which requires further refactoring.

> For tests that check for errors, please use `pytest.raises()` for that instead of implementing `_test_raise_type_error()`

OK, this is much better than my clumsy implementation.

python-package/lightgbm/basic.py
```diff
@@ -41,20 +41,26 @@ def info(self, msg: str) -> None:
     def warning(self, msg: str) -> None:
         warnings.warn(msg, stacklevel=3)
 
+    def error(self, msg: str) -> None:
+        warnings.warn(msg, stacklevel=3)
```
StrikerRUS, line 45: I guess this was a typo, right?
```suggestion
    def error(self, msg: str) -> None:
        warnings.error(msg, stacklevel=3)
```

python-package/lightgbm/basic.py
```diff
@@ -41,20 +41,26 @@ def info(self, msg: str) -> None:
...
+
 
-_LOGGER: Union[_DummyLogger, Logger] = _DummyLogger()
+_LOGGER: Any = _DummyLogger()
 
 
-def register_logger(logger: Logger) -> None:
+def register_logger(logger: Any) -> None:
```
StrikerRUS, line 51: I think we can make this API even more flexible by introducing three new arguments with default values: `info_method_name`, `warning_method_name` and `error_method_name`. This will allow, for example, use `warning` or `warn` method depending on logger backend
> Note There is an obsolete method `warn` which is functionally identical to `warning`. As `warn` is deprecated, please do not use it - use `warning` instead.
   https://docs.python.org/3/library/logging.html#logging.Logger.warning

StrikerRUS (changes requested):

python-package/lightgbm/basic.py
RustingSword: Seems there's no `error()` in [`warnings`](https://docs.python.org/3/library/warnings.html#available-functions)? I seldom use this module, so correct me if I'm wrong.

python-package/lightgbm/basic.py
RustingSword: I agree, will add these args.

RustingSword:
> Is there any way to directly test the logging utility in `lightgbm`

I find that maybe I can directly use `lgb._log_info(msg)` to test logging functions?

Also I'm wondering whether we should export `_log_info`, `_log_warning`, `_log_error` as public methods, say as `info`, `warning` and `error`. So once we registered our own logger, we can use `lgb.info()`, `lgb.warning()`, `lgb.error()` in user code, to make the usage and logging result consistent.

python-package/lightgbm/basic.py
StrikerRUS: Ah, indeed! I confused this with `error()` function from logging module. Sorry.
https://docs.python.org/3/library/logging.html#logging.Logger.error

For our dummy logger I think we can just raise an exception then, WDYT?

RustingSword: Since this is the default logger, in the future when `error()` is actually needed, we still need to implement it. So I'm inclined to leave a dummy yet usable implementation as default.

jameslamb:
@RustingSword are you still interested in pursuing this PR? If you can update this to the latest `master`, I'd be happy to provide another review.

--------

I'm sorry, I realized today that you were waiting for feedback from us on https://github.com/microsoft/LightGBM/pull/4880#issuecomment-992063235.

> Also I'm wondering whether we should export _log_info, _log_warning, _log_error as public methods, say as info, warning and error

I don't believe those methods should be exported or that calling LightGBM's logger like `lgb.info()` should be encouraged. I think the ability to pass in your own logger means that if you want to have other code that calls the same logger as LightGBM, you can instantiate that logger in your code and pass it into `register_logger()`.

RustingSword:
@jameslamb I rebased on master. If we don't expose logging methods, then there's no direct way to test logger. Should I duplicate the tests used in [test_register_logger ](https://github.com/microsoft/LightGBM/blob/master/tests/python_package_test/test_utilities.py#L9)? Personally I think the test is coupled with other functionalities, such as training, which have been tested in other module, e.g. test_engine.py.

jameslamb:
> I rebased on master

Since this project uses squash merging, just using `git merge master`  and pushing a merge commit is sufficient. This is also typically easier...something about the way you rebased has made GitHub think this PR is changing 100 files, which is going to make it really difficult to review.

Screen Shot 2022-03-19 at 10 08 23 PM

Can you please remove those commits and instead *merge* `master`, so the diff only shows what this PR is changing? If you're not sure how to, please let me know and I'd be happy to update this branch for you.

> If we don't expose logging methods, then there's no direct way to test logger

Since you are injecting a logger, I think it's possible to design one for a test which modifies a global variable, and then to test the content of those global variables.

pseudo-code showing what I mean `INFO`-level logs (which could be extended for `WARN` and `FATAL`).

```python
import lightgbm as lgb

info_logs = []

class _TestLogger:
    def info(self, msg: str) -> None:
        global info_logs
        info_logs.append(msg)

lgb.register_logger(_TestLogger())

# directly invoke `lgb.basic._log_info()`
lgb.basic._log_info("hello")

assert info_logs == ["hello"]
```

Besides that, you could have one similar test that performs training with `lgb.train()`, and just check that `info_logs` is non-empty. That way, the test wouldn't be as tightly coupled to LightGBM's training logic, but would still check "`lgb.train()` actually uses the logger registered with `register_logger()`".

RustingSword:
Sorry, I will do a force update on this branch to merge master instead of rebase. 

The suggested test method is great, will implement later.

jameslamb:
Sounds good, thanks! We are VERY sorry it took so long to respond to your questions. Thanks very much for still being willing to contribute!

StrikerRUS:
> I don't believe those methods should be exported or that calling LightGBM's logger like `lgb.info()` should be encouraged. I think the ability to pass in your own logger means that if you want to have other code that calls the same logger as LightGBM, you can instantiate that logger in your code and pass it into `register_logger()`.

I'm totally agree with this statement.

StrikerRUS:
@RustingSword 
Could you please fix one linting error?
```
./tests/python_package_test/test_utilities.py:147:1: E303 too many blank lines (3)
```

RustingSword:
> @RustingSword Could you please fix one linting error?
> 
> ```
> ./tests/python_package_test/test_utilities.py:147:1: E303 too many blank lines (3)
> ```

Sorry, missed this one, fixed.

RustingSword:
When I'm running pytest locally, the expected log messages in `test_register_logger` miss two lines at the beginning, which are
```
WARNING | categorical_feature in Dataset is overridden.
New categorical_feature is [1]
```

I'm not sure whether this should be updated, since there seems no related test error in the ci pipeline.

StrikerRUS:
@RustingSword 
> When I'm running pytest locally ...

Are you using the latest version of LightGBM?
Refer to #4768.

RustingSword:
> @RustingSword
> 
> > When I'm running pytest locally ...
> 
> Are you using the latest version of LightGBM? Refer to #4768.

I see, I'm using 3.3.2, and #4768 is not included per #4930.

python-package/lightgbm/basic.py
jameslamb: Why is a `.error()` method being added at all? Per your description in #4783 

> Since only `info` and `warning` are used to log messages, any logger class providing these two methods would suffice.

And I can confirm that such a method isn't used anywhere.

```shelll
git grep -E "\.error"
```

I have a strong preference for not having any "well we might need this in the future" code in the package, as it increases the size of the package and amount of code to be maintained without affecting the user experience. I think this PR should not add a `.error()` implementation at all. @StrikerRUS what do you think?

jameslamb (changes requested):
Tests look great, thank you very much! Please just see my one remaining comment about the `.error()` methood.

python-package/lightgbm/basic.py
StrikerRUS: OK, let's remove requirements to have `.error()` method in custom logger class in this PR. But please note that there will be a breaking change in case we'll want to require it in the future.

RustingSword: Removed `.error()` related code.

jameslamb: Thanks!

> please note that there will be a breaking change in case we'll want to require it in the future

Yes, I understand. I think the case where the `lightgbm` library would want to emit an ERROR-level log but not raise an exception is unlikely to come up.

Even the logger on the C++ side does not have a method for ERROR-level messages: https://github.com/microsoft/LightGBM/blob/17d4e007852b4f0b34b211a674cb26e6751c98e2/include/LightGBM/utils/log.h.

jameslamb (approved):
looks great to me, thanks very much for the idea and the contribution!

python-package/lightgbm/basic.py
```diff
@@ -41,21 +41,37 @@ def warning(self, msg: str) -> None:
...
 
     Parameters
     ----------
-    logger : logging.Logger
+    logger : Any
         Custom logger.
+    info_method_name: str, optional (default="info")
+        Method used to log info messages.
+    warning_method_name: str, optional (default="warning")
+        Method used to log warning messages.
```
StrikerRUS, line 61:
```suggestion
    info_method_name : str, optional (default="info")
        Method used to log info messages.
    warning_method_name : str, optional (default="warning")
        Method used to log warning messages.
```

StrikerRUS (approved):
@RustingSword Thank you so much for your work!

Just one nit about docstring style:

python-package/lightgbm/basic.py
RustingSword: Fixed. Maybe we can add a corresponding linter rule? Or even better, use a unified formatter.

RustingSword:
@jameslamb @StrikerRUS Thank you both for your excellent guidance!

python-package/lightgbm/basic.py
StrikerRUS: Nice idea! Do you know any one?
`pydocstyle` linter we use here doesn't catch such inconsistency.
https://github.com/microsoft/LightGBM/blob/d163c2c16629e3da41dc5d7b8dbb76b8fc6249db/.ci/test.sh#L81

RustingSword: Seems this is a [known issue](https://github.com/PyCQA/pydocstyle/issues/519) with pydocstyle. I tried `numpydoc` as mentioned in the comment, however it is not satisfying either. First, `numpydoc` is an extension to `sphinx`, and the validation rules cannot be set in command line options when used as a standalone tool. Second, `numpydoc` is not intelligent enough. We can see this in above issue, where a single missing whitespace will result in multiple false positives.

```
smalltest.no:PR01:Parameters {'a'} not documented
smalltest.no:PR02:Unknown parameters {'a: int'}
```

These two errors are due to parsing error caused by the missing whitespace before colon.

So, I don't have suggested tool suited for this 😞

jameslamb: Ah no problem, thanks very much for investigating!

PR #4882, opened 2021-12-12
[python] remove `evals_result` argument of `train()` function

StrikerRUS:
Refer to #4574.


jameslamb (approved):
I support this, thanks!

PR #4884, opened 2021-12-12
[python][sklearn] do not replace empty dict with `None` for `evals_result_`

StrikerRUS:
Simplify codebase and unify standard and sklearn interfaces.
Similarly to #4845.


jameslamb (approved):
excellent! Totally support  this simplification.

PR #4890, opened 2021-12-14
clear memory allocated for sampled data when constructing Dataset from text file

xuchuanyin:
Sample data is useless after BinMapper is constructed, but the corresponding memory is still there before feature extraction is finished.

### DIFF
--- a/src/io/dataset_loader.cpp
+++ b/src/io/dataset_loader.cpp
@@ -234,8 +234,9 @@ Dataset* DatasetLoader::LoadFromFile(const char* filename, int rank, int num_mac
       auto sample_data = SampleTextDataFromMemory(text_data);
       CheckSampleSize(sample_data.size(),
                       static_cast<size_t>(dataset->num_data_));
-      // construct feature bin mappers
+      // construct feature bin mappers & clear sample data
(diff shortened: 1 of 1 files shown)


jameslamb (commented):
excellent, thanks for this!

I support this change.

Since users can configure how much data is sampled during the calculation of bins by setting parameter `bin_construct_sample_cnt`, it's definitely possible for the sampled data to have a noticable memory footprint.

However, a reviewer more familiar with C++, like @shiyu1994 or @guolinke , should probably review this before it's merged.

https://github.com/microsoft/LightGBM/blob/9f13a9c897ce41beebb50c3e44b501f4a7a0145e/src/io/dataset_loader.cpp#L926-L927

jameslamb:
@xuchuanyin I hope you don't mind, I've proposed a re-wording of this pull request's title. Since pull request titles become items in this project's release notes, I've proposed a title that I think will be a bit easier to understand for users of LightGBM who aren't familiar with its internals.

xuchuanyin:
@jameslamb yeah, the optimized title is much better

shiyu1994 (approved):
Thanks for your contribution!

Sorry for the late response. I was busy with another project in previous days. And now I'm back.

PR #4891, opened 2021-12-16
[python] Drop Python 3.6 support

StrikerRUS:
Python 3.6 is reaching its End of Life next week and no need to support it anymore.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -28,7 +28,6 @@ install:
   - set CONDA_ENV="test-env"
   - ps: |
       switch ($env:PYTHON_VERSION) {
-          "3.6" {$env:MINICONDA = "C:\Miniconda36-x64"}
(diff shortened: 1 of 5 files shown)


jameslamb (approved):
I support it! Especially since other dependencies of `lightgbm` have already dropped support for Python 3.6, like `numpy` (https://github.com/numpy/numpy/blob/0302205105161c6bbae5d45fe4e7a5a624f0dbf4/setup.py#L410) and `pandas` (https://github.com/pandas-dev/pandas/blob/47eb2198896852e17005629ed82c18940a05e975/setup.cfg#L40).

StrikerRUS:
> other dependencies of `lightgbm` have already dropped support for Python 3.6

They have already dropped even Python 3.7 (EoL 2023-06-27), interesting...

PR #4893, opened 2021-12-18
[python] Start supporting Python 3.10

StrikerRUS:
Python 3.10 was released 2021-10-04.

Why enclose Python version in quotes: https://github.com/microsoft/LightGBM/pull/3880#discussion_r566841420.
```
...
Solving environment: ...working... failed                                                               
PackagesNotFoundError: The following packages are not available from current channels:
  - python=3.1
...
```

**WIP** Due to
- `dask=2021.9.1` isn't compatible with Python 3.10;
   https://github.com/microsoft/LightGBM/blob/8e729af38dc0d15baba47a28d552d049a96f27f3/.ci/test.sh#L117
- building `numpy` and other dependencies from sources takes more than 30 minutes.


StrikerRUS:
[advanced_example.py](https://github.com/microsoft/LightGBM/blob/master/examples/python-guide/advanced_example.py) fails during executing on [`Linux regular` CI job](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=12181&view=logs&j=c28dceab-947a-5848-c21f-eef3695e5f11&t=fa158246-17e2-53d4-5936-86070edbaacf&l=1012) with Python 3.10 with the following error
```
  File "/__w/1/s/examples/python-guide/advanced_example.py", line 8, in <module>
  File "/opt/miniforge/envs/test-env/lib/python3.10/site-packages/sklearn/__init__.py", line 83, in <module>
    from .utils._show_versions import show_versions
  File "/opt/miniforge/envs/test-env/lib/python3.10/site-packages/sklearn/utils/_show_versions.py", line 12, in <module>
    from ._openmp_helpers import _openmp_parallelism_enabled
ImportError: dlopen: cannot load any more object with static TLS
```
This fail occurs due to compatibility problems with old Ubuntu 14.04 (old glibc actually) and OpenMP.
Related discussions and workarounds (all about upgrading software versions):
- https://github.com/scikit-learn/scikit-learn/issues/14485
- https://github.com/pytorch/pytorch/issues/2575
- https://github.com/conda-forge/conda-forge.github.io/issues/1551
- https://finisky.github.io/2020/01/06/jupyterdlopen/

This error shouldn't occur with `glibc >=2.21`: https://sourceware.org/bugzilla/show_bug.cgi?id=17620.

StrikerRUS:
It's getting harder and harder to support a such old version of Ubuntu (14.04) ...

jameslamb:
> It's getting harder and harder to support a such old version of Ubuntu (14.04)

Thanks for the additional research and issues you linked to! https://github.com/conda-forge/conda-forge.github.io/issues/1551 is especially troubling 😞 

According to https://ubuntu.com/about/release-cycle#ubuntu-kernel-release-cycle, Ubuntu 14.04 was first released in 2014, starting in 2019 it went into "only receiving security patches" mode, and it will be in that mode until 2024.

image

Given all that, I'd like to try to keep supporting it, but also acknowledge that we have a limited amount of time and energy to spend on maintaining this project.

Since the issues you linked are all related to OpenMP, what do you think about switching Ubuntu 14.04 builds to using `--nomp`?

https://github.com/microsoft/LightGBM/blob/2d1caf140a3b0c68f908071277b67c2a842a297b/python-package/setup.py#L25

Maybe that could be a manageable compromise between "spend time trying to get this working on Ubuntu 14.04" and "stop testing on Ubuntu 14.04 entirely".

StrikerRUS:
> Since the issues you linked are all related to OpenMP, what do you think about switching Ubuntu 14.04 builds to using `--nomp`?

A lot of CI jobs with Ubuntu 14.04 are used to produce public artifacts
https://github.com/microsoft/LightGBM/blob/2d1caf140a3b0c68f908071277b67c2a842a297b/.vsts-ci.yml#L72-L73
I don't think it's a good idea to build artifacts with `--nomp` flag...

I'm thinking about to simply re-arange Python versions there because it seems that only 3.10 one has compatibility problems.

.vsts-ci.yml
```diff
@@ -35,17 +35,18 @@ jobs:
     matrix:
       regular:
         TASK: regular
+        PYTHON_VERSION: '3.9'
```
StrikerRUS, line 38: I tried to use different Python versions for the same `TASK` in `Linux` and `Linux_latest` jobs. The same is true for macOS configurations on Azure Pipelines and GitHub Actions.

StrikerRUS:
I believe this PR is ready for review.

jameslamb (approved):
excellent! thanks for doing this! And for the explanation about why it is necessary to add quotes in the config 😂

PR #4894, opened 2021-12-19
[python-package] support customizing Dataset creation in Booster.refit() (fixes #3038)

### LINKED ISSUE #3038: refit in Python does not support weights
## Summary

The [refit method](https://lightgbm.readthedocs.io/en/latest/pythonapi/lightgbm.Booster.html?highlight=refit#lightgbm.Booster.refit) in Python does not support weights (or in fact anything other than data and labels). This is because [here](https://github.com/microsoft/LightGBM/blob/44a9120197fbfc7f3bfb9349e4eb2e9443c1971b/python-package/lightgbm/basic.py#L2666), the training data set gets created using:
```python
train_set = Dataset(data, label, silent=True)
```

It would be great if refit accepted additional arguments (or kwargs) specifically for the `Dataset` call.

Passing in the `train_set` directly would also be an option. But since we need to predict first, we would also need to pass the training data in as data frame, which is not so nice.

## Motivation

I'm using re-fit in an application where weights are very important.

I'm happy to open a PR if this sounds useful.

TremaMiguel:
**Context**:
This PR aims to close #3038 

**Changes**
* refit method accepts `kwargs_for_dataset` parameter to pass weight parameter to Dataset initialization inside `refit()`.
* refit method accepts `kwargs_for_predict` parameter to pass original params to `predict` method.

**Tests**:
Based on @jtilly example on issue #3038 test that `refit` method returns a valid prediction value when passing `kwargs_for_dataset` or `kwargs_for_predict`.

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -3532,9 +3532,9 @@ class Booster:
             will use ``leaf_output = decay_rate * old_leaf_output + (1.0 - decay_rate) * new_leaf_output`` to refit trees.
         reference : Dataset or None, optional (default=None)
             reference for ``data``.
-            If this is Dataset for validation, training data should be used as reference.
         weight : list, numpy 1-D array, pandas Series or None, optional (default=None)
-            Weight for each ``data`` instance.
+            Weight for each ``data`` instance. Weight should be non-negative values because the Hessian
+            value multiplied by weight is supposed to be non-negative.
         group : list, numpy 1-D array, pandas Series or None, optional (default=None)
             Group/query size for ``data``.
         init_score : list, numpy 1-D array, pandas Series or None, optional (default=None)
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -3503,7 +3503,7 @@ class Booster:
                                  raw_score, pred_leaf, pred_contrib,
                                  data_has_header, is_reshape)
 
-    def refit(self, data, label, decay_rate=0.9, kwargs_for_predict=None, kwargs_for_dataset=None):
+    def refit(self, data, label, decay_rate=0.9, dataset_params=None, **kwargs):
         """Refit the existing Booster by new data.
 
         Parameters
@@ -3516,11 +3516,12 @@ class Booster:
         decay_rate : float, optional (default=0.9)
             Decay rate of refit,
             will use ``leaf_output = decay_rate * old_leaf_output + (1.0 - decay_rate) * new_leaf_output`` to refit trees.
-        kwargs_for_predict: dict, optional (default=None)
-            parameters passed to ``predict`` method.
-        kwargs_for_dataset: dict, optional (default=None)
+        dataset_params: dict, optional (default=None)
             additional parameters passed to ``Dataset`` class. If the parameters ``data, label, params`` are contained, they
             are removed.
+        **kwargs
+            Other parameters for refit.
+            These parameters will be passed to ``predict`` method.
 
         Returns
         -------
@@ -3529,9 +3530,8 @@ class Booster:
         """
         if self.__set_objective_to_none:
             raise LightGBMError('Cannot refit due to null objective function.')
-        kwargs_for_predict = {} if kwargs_for_predict is None else kwargs_for_predict
-        kwargs_for_dataset = {} if kwargs_for_dataset is None else kwargs_for_dataset
-        predictor = self._to_predictor(deepcopy(kwargs_for_predict))
+        dataset_params = {} if dataset_params is None else dataset_params
+        predictor = self._to_predictor(deepcopy(kwargs))
         leaf_preds = predictor.predict(data, -1, pred_leaf=True)
         nrow, ncol = leaf_preds.shape
         out_is_linear = ctypes.c_int(0)
@@ -3545,8 +3545,8 @@ class Booster:
         )
         new_params["linear_tree"] = bool(out_is_linear.value)
         for arg in ['data', 'label', 'params']:
-            kwargs_for_dataset.pop(arg, None)
-        train_set = Dataset(data, label, params=new_params, **kwargs_for_dataset)
+            dataset_params.pop(arg, None)
+        train_set = Dataset(data, label, params=new_params, **dataset_params)
         new_params['refit_decay_rate'] = decay_rate
         new_booster = Booster(new_params, train_set)
         # Copy models
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -3514,6 +3514,7 @@ class Booster:
         init_score=None,
         feature_name='auto',
         categorical_feature='auto',
+        dataset_params=None,
         free_raw_data=True,
         **kwargs
     ):
@@ -3550,7 +3551,9 @@ class Booster:
             Large values could be memory consuming. Consider using consecutive integers starting from zero.
             All negative values in categorical features will be treated as missing values.
             The output cannot be monotonically constrained with respect to a categorical feature.
-         free_raw_data : bool, optional (default=True)
+        dataset_params : dict or None, optional (default=None)
+            Other parameters for Dataset ``data``.
+        free_raw_data : bool, optional (default=True)
             If True, raw data is freed after constructing inner Dataset for ``data``.
         **kwargs
             Other parameters for refit.
@@ -3563,6 +3566,8 @@ class Booster:
         """
         if self.__set_objective_to_none:
             raise LightGBMError('Cannot refit due to null objective function.')
+        if dataset_params is None:
+            dataset_params = {}
         predictor = self._to_predictor(deepcopy(kwargs))
         leaf_preds = predictor.predict(data, -1, pred_leaf=True)
         nrow, ncol = leaf_preds.shape
@@ -3576,6 +3581,7 @@ class Booster:
             default_value=None
         )
         new_params["linear_tree"] = bool(out_is_linear.value)
+        new_params.update(dataset_params)
         train_set = Dataset(
             data=data,
             label=label,
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -3531,21 +3531,25 @@ class Booster:
             Decay rate of refit,
             will use ``leaf_output = decay_rate * old_leaf_output + (1.0 - decay_rate) * new_leaf_output`` to refit trees.
         reference : Dataset or None, optional (default=None)
-            reference for ``data``.
+            Reference for ``data``.
         weight : list, numpy 1-D array, pandas Series or None, optional (default=None)
             Weight for each ``data`` instance. Weight should be non-negative values because the Hessian
             value multiplied by weight is supposed to be non-negative.
         group : list, numpy 1-D array, pandas Series or None, optional (default=None)
             Group/query size for ``data``.
-        init_score : list, numpy 1-D array, pandas Series or None, optional (default=None)
+            Only used in the learning-to-rank task.
+            sum(group) = n_samples.
+            For example, if you have a 100-document dataset with ``group = [10, 20, 40, 10, 10, 10]``, that means that you have 6 groups,
+            where the first 10 records are in the first group, records 11-30 are in the second group, records 31-70 are in the third group, etc.
+        init_score : list, list of lists (for multi-class task), numpy array, pandas Series, pandas DataFrame (for multi-class task), or None, optional (default=None)
             Init score for ``data``.
-        feature_name : list of strings or 'auto', optional (default="auto")
+        feature_name : list of str or 'auto', optional (default="auto")
             Feature names for ``data``.
             If 'auto' and data is pandas DataFrame, data columns names are used.
-        categorical_feature : list of strings or int, or 'auto', optional (default="auto")
+        categorical_feature : list of str or int, or 'auto', optional (default="auto")
             Categorical features for ``data``.
             If list of int, interpreted as indices.
-            If list of strings, interpreted as feature names (need to specify ``feature_name`` as well).
+            If list of str, interpreted as feature names (need to specify ``feature_name`` as well).
             If 'auto' and data is pandas DataFrame, pandas unordered categorical columns are used.
             All values in categorical features should be less than int32 max value (2147483647).
             Large values could be memory consuming. Consider using consecutive integers starting from zero.
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -3503,7 +3503,7 @@ class Booster:
                                  raw_score, pred_leaf, pred_contrib,
                                  data_has_header, is_reshape)
 
-    def refit(self, data, label, decay_rate=0.9, **kwargs):
+    def refit(self, data, label, decay_rate=0.9, kwargs_for_predict=None, kwargs_for_dataset=None):
         """Refit the existing Booster by new data.
 
         Parameters
@@ -3516,9 +3516,10 @@ class Booster:
         decay_rate : float, optional (default=0.9)
             Decay rate of refit,
             will use ``leaf_output = decay_rate * old_leaf_output + (1.0 - decay_rate) * new_leaf_output`` to refit trees.
-        **kwargs
-            Other parameters for refit.
-            These parameters will be passed to ``predict`` method.
+        kwargs_for_predict: dict, optional (default=None)
+            parameters passed to ``predict`` method.
+        kwargs_for_dataset: dict, optional (default=None)
+            additional parameters passed to ``Dataset`` class. The keys data, label and params should not be contained.
 
         Returns
         -------
@@ -3527,7 +3528,9 @@ class Booster:
         """
         if self.__set_objective_to_none:
             raise LightGBMError('Cannot refit due to null objective function.')
-        predictor = self._to_predictor(deepcopy(kwargs))
+        kwargs_for_predict = {} if kwargs_for_predict is None else kwargs_for_predict
+        kwargs_for_dataset = {} if kwargs_for_dataset is None else kwargs_for_dataset
+        predictor = self._to_predictor(deepcopy(kwargs_for_predict))
         leaf_preds = predictor.predict(data, -1, pred_leaf=True)
         nrow, ncol = leaf_preds.shape
         out_is_linear = ctypes.c_int(0)
@@ -3540,7 +3543,7 @@ class Booster:
             default_value=None
         )
         new_params["linear_tree"] = bool(out_is_linear.value)
-        train_set = Dataset(data, label, params=new_params)
+        train_set = Dataset(data, label, params=new_params, **kwargs_for_dataset)
         new_params['refit_decay_rate'] = decay_rate
         new_booster = Booster(new_params, train_set)
         # Copy models
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -3543,7 +3543,7 @@ class Booster:
             where the first 10 records are in the first group, records 11-30 are in the second group, records 31-70 are in the third group, etc.
         init_score : list, list of lists (for multi-class task), numpy array, pandas Series, pandas DataFrame (for multi-class task), or None, optional (default=None)
             Init score for ``data``.
-        feature_name : list of str or 'auto', optional (default="auto")
+        feature_name : list of str, or 'auto', optional (default="auto")
             Feature names for ``data``.
             If 'auto' and data is pandas DataFrame, data columns names are used.
         categorical_feature : list of str or int, or 'auto', optional (default="auto")
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -3519,7 +3519,8 @@ class Booster:
         kwargs_for_predict: dict, optional (default=None)
             parameters passed to ``predict`` method.
         kwargs_for_dataset: dict, optional (default=None)
-            additional parameters passed to ``Dataset`` class. The keys data, label and params should not be contained.
+            additional parameters passed to ``Dataset`` class. If the parameters ``data, label, params`` are contained, they
+            are removed.
 
         Returns
         -------
@@ -3543,6 +3544,8 @@ class Booster:
             default_value=None
         )
         new_params["linear_tree"] = bool(out_is_linear.value)
+        for arg in ['data', 'label', 'params']:
+            kwargs_for_dataset.pop(arg, None)
         train_set = Dataset(data, label, params=new_params, **kwargs_for_dataset)
         new_params['refit_decay_rate'] = decay_rate
         new_booster = Booster(new_params, train_set)
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -3503,7 +3503,20 @@ class Booster:
                                  raw_score, pred_leaf, pred_contrib,
                                  data_has_header, is_reshape)
 
-    def refit(self, data, label, decay_rate=0.9, dataset_params=None, **kwargs):
+    def refit(
+        self,
+        data,
+        label,
+        decay_rate=0.9,
+        reference=None,
+        weight=None,
+        group=None,
+        init_score=None,
+        feature_name='auto',
+        categorical_feature='auto',
+        free_raw_data=True,
+        **kwargs
+    ):
         """Refit the existing Booster by new data.
 
         Parameters
@@ -3516,9 +3529,29 @@ class Booster:
         decay_rate : float, optional (default=0.9)
             Decay rate of refit,
             will use ``leaf_output = decay_rate * old_leaf_output + (1.0 - decay_rate) * new_leaf_output`` to refit trees.
-        dataset_params: dict, optional (default=None)
-            additional parameters passed to ``Dataset`` class. If the parameters ``data, label, params`` are contained, they
-            are removed.
+        reference : Dataset or None, optional (default=None)
+            reference for ``data``.
+            If this is Dataset for validation, training data should be used as reference.
+        weight : list, numpy 1-D array, pandas Series or None, optional (default=None)
+            Weight for each ``data`` instance.
+        group : list, numpy 1-D array, pandas Series or None, optional (default=None)
+            Group/query size for ``data``.
+        init_score : list, numpy 1-D array, pandas Series or None, optional (default=None)
+            Init score for ``data``.
+        feature_name : list of strings or 'auto', optional (default="auto")
+            Feature names for ``data``.
+            If 'auto' and data is pandas DataFrame, data columns names are used.
+        categorical_feature : list of strings or int, or 'auto', optional (default="auto")
+            Categorical features for ``data``.
+            If list of int, interpreted as indices.
+            If list of strings, interpreted as feature names (need to specify ``feature_name`` as well).
+            If 'auto' and data is pandas DataFrame, pandas unordered categorical columns are used.
+            All values in categorical features should be less than int32 max value (2147483647).
+            Large values could be memory consuming. Consider using consecutive integers starting from zero.
+            All negative values in categorical features will be treated as missing values.
+            The output cannot be monotonically constrained with respect to a categorical feature.
+         free_raw_data : bool, optional (default=True)
+            If True, raw data is freed after constructing inner Dataset for ``data``.
         **kwargs
             Other parameters for refit.
             These parameters will be passed to ``predict`` method.
@@ -3530,7 +3563,6 @@ class Booster:
         """
         if self.__set_objective_to_none:
             raise LightGBMError('Cannot refit due to null objective function.')
-        dataset_params = {} if dataset_params is None else dataset_params
         predictor = self._to_predictor(deepcopy(kwargs))
         leaf_preds = predictor.predict(data, -1, pred_leaf=True)
         nrow, ncol = leaf_preds.shape
@@ -3544,9 +3576,18 @@ class Booster:
             default_value=None
         )
         new_params["linear_tree"] = bool(out_is_linear.value)
-        for arg in ['data', 'label', 'params']:
-            dataset_params.pop(arg, None)
-        train_set = Dataset(data, label, params=new_params, **dataset_params)
+        train_set = Dataset(
+            data=data,
+            label=label,
+            reference=reference,
+            weight=weight,
+            group=group,
+            init_score=init_score,
+            feature_name=feature_name,
+            categorical_feature=categorical_feature,
+            params=new_params,
+            free_raw_data=free_raw_data,
+        )
         new_params['refit_decay_rate'] = decay_rate
         new_booster = Booster(new_params, train_set)
         # Copy models
--- a/tests/python_package_test/test_engine.py
+++ b/tests/python_package_test/test_engine.py
@@ -50,24 +50,6 @@ def categorize(continuous_x):
     return np.digitize(continuous_x, bins=np.arange(0, 1, 0.01))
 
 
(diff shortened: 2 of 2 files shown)


python-package/lightgbm/basic.py
```diff
@@ -3503,7 +3503,7 @@ def predict(self, data, start_iteration=0, num_iteration=None,
                                  raw_score, pred_leaf, pred_contrib,
                                  data_has_header, is_reshape)
 
-    def refit(self, data, label, decay_rate=0.9, **kwargs):
+    def refit(self, data, label, decay_rate=0.9, kwargs_for_predict=None, kwargs_for_dataset=None):
```
jameslamb, line 3506: The proposed approach is a breaking change. It means that code which previously passed any of the [prediction parameters](https://lightgbm.readthedocs.io/en/latest/Parameters.html#predict-parameters) would now have to be changed to bundle those up into a dictionary passed to `kwargs_for_predict`.

Will you consider this suggestion? It would allow the behavior that you want (fully customize the inner call to `Dataset()`) without breaking any users' existing code.

```suggestion
    def refit(
        self,
        data,
        label,
        decay_rate=0.9,
        reference=None,
        weight=None,
        group=None,
        init_score=None,
        feature_name='auto',
        categorical_feature='auto',
        dataset_params=None,
        free_raw_data=True,
        **kwargs
     ):
```

Specifically, I'm suggesting just exposing all of the keyword arguments from the `Dataset` construct ([link](https://lightgbm.readthedocs.io/en/latest/pythonapi/lightgbm.Dataset.html#lightgbm.Dataset)) and then passing them directly into it.

That would avoid breaking existing code and would make documentation and auto-completion for people working in IDEs.

@StrikerRUS @jmoralez what do you think?

tests/python_package_test/test_engine.py
```diff
@@ -50,6 +50,24 @@ def categorize(continuous_x):
     return np.digitize(continuous_x, bins=np.arange(0, 1, 0.01))
 
 
+@pytest.fixture
+def artifacts_for_refit_kwargs():
```
jameslamb, line 54: I think that the indirection added by using this fixture isn't worth the small reduction in duplicated code.

Can you please remove this fixture and just copy this code directly into each of the tests you're adding?

tests/python_package_test/test_engine.py
```diff
@@ -50,6 +50,24 @@ def categorize(continuous_x):
...
+    X = np.array([1, 2, 2]).reshape((3, 1))
+    label = np.array([1, 2, 3])
+    data = lgb.basic.Dataset(X, label)
+    booster = lgb.engine.train(
+        {
+            "min_data_in_bin": 1,
+            "min_data_in_leaf": 1,
+            "learning_rate": 1,
+            "boost_from_average": False,
```
jameslamb, line 63:
```suggestion
            "learning_rate": 1,
            "boost_from_average": False,
```

Why is it important, for these tests, to specify these two parameters? I don't understand how they impact the results that are being tested. If these are just left over from other code, please remove them.

tests/python_package_test/test_engine.py
jameslamb: Actually....as mentioned in https://github.com/microsoft/LightGBM/issues/3038#issuecomment-996211592, please base these new tests on

https://github.com/microsoft/LightGBM/blob/8a34b1af2d5ae734a446f97ee984508a55ecb2d4/tests/python_package_test/test_engine.py#L1531-L1533

Specifically, please use the existing datasets used in other tests.

As I think you've noticed (because of setting `min_data_in_bin=1, min_data_in_leaf=1`), LightGBM has parameter defaults which are not suitable to very very small datasets like this 3-row, 1-feature dataset you've created. Using such small datasets can lead to instability in tests.

tests/python_package_test/test_engine.py
```diff
@@ -1545,6 +1563,39 @@ def test_refit():
...
+        "init_score": None,
+        "feature_name": "auto",
+        "categorical_feature": "auto",
+        "free_raw_data": True
+    }
+    booster_refit = booster.refit(
+        X, label, kwargs_for_dataset=kwargs_for_dataset
+    )
+    pred = booster_refit.predict(X)
+    assert pred.shape == (3, )
```
jameslamb, line 1582: I think that this assertion would be true whether or not the `weight` passed above was used, wouldn't it? If I'm right about that, then this is not effectively testing the changes in this PR.

Please test that the results you get are different based on the weights provided.

You could use the following as an example:

https://github.com/microsoft/LightGBM/blob/8a34b1af2d5ae734a446f97ee984508a55ecb2d4/tests/python_package_test/test_engine.py#L528-L537

jameslamb (changes requested):
Thanks for picking this up! I left some initial comments based on the format of what you're proposing, but I'm not the best person to comment on whether or not #3038 should be adopted at all.

For example, I don't know if the comment mentioned at https://github.com/microsoft/LightGBM/issues/3038#issuecomment-621509323 means that `init_score`, `weight`, and `group` is still true.

Hopefully @guolinke @shiyu1994 or  @StrikerRUS can comment on whether we should proceed with this feature.

tests/python_package_test/test_engine.py
StrikerRUS: > Can you please remove this fixture and just copy this code directly into each of the tests you're adding?

I do support this.

python-package/lightgbm/basic.py
StrikerRUS: TBH, I'm OK with any of these two interfaces. Only one thing that I'm against of is param names. I believe `kwargs_for_dataset` and `kwargs_for_predict` should be `params_for_dataset` and `params_for_predict ` respectively.

tests/python_package_test/test_engine.py
StrikerRUS: Also I think these two new tests can be merged into one to reduce overheads.

python-package/lightgbm/basic.py
jameslamb: I think of `params` as anything that shows up in https://lightgbm.readthedocs.io/en/latest/Parameters.html and is an attribute of the `config` object on the C++ side.

That description doesn't apply to things like  `init_score`, `weight`, and `group`.

That's why I've proposed `dataset_params` as a keyword argument, to capture anything in https://lightgbm.readthedocs.io/en/latest/Parameters.html#dataset-parameters.

StrikerRUS: OK. My intention was just to get rid of `kwargs` word in argument names actually. 🙂

TremaMiguel: OK, agree on adding a single `dataset_params` argument that includes all other parameters to be passed to [Dataset](https://github.com/microsoft/LightGBM/blob/44a9120197fbfc7f3bfb9349e4eb2e9443c1971b/python-package/lightgbm/basic.py#L740) and leaving `kwargs` as it is now.

tests/python_package_test/test_engine.py
TremaMiguel: OK, thanks for the reference @jameslamb.

tests/python_package_test/test_engine.py
TremaMiguel: @jameslamb just to double check, the test should prove that a weighted sample gives different metric value than a non-weighted, but not necessarily a better score because weights are random `weight=np.abs(np.random.normal(size=y.shape))`, is that right?

jameslamb: yes, as long as choosing the weights is the only source of randomness. Just something that gives us confidence that weights are actually being used inside this method's code.

tests/python_package_test/test_engine.py
TremaMiguel: replace fixed data with breast cancer data.

python-package/lightgbm/basic.py
jameslamb: Since @StrikerRUS didn't have a strong preference between the interfaces, please use the one I proposed above with dedicated keyword arguments for each keyword argument of `lgb.Dataset()`, and using `dataset_params` to be passed to `params` in `lgb.Dataset()`.

I feel strongly that the term "params" has a special meaning in LightGBM and including things like `init_score`, `weight` in anything called `params` could cause confusion for users.

TremaMiguel: @jameslamb shouldn't it be all parameters from `Dataset` except `params` as `train_set = Dataset(data, label, params=new_params)` and `new_params` is defined inside the function

jameslamb: I do see that at

https://github.com/microsoft/LightGBM/blob/e4c0ca5f5d428a2497944e83c4db1efae6949b04/python-package/lightgbm/basic.py#L3537-L3542

But if the goal of this PR is to support fully customizing the construction of the `Dataset` inside `Booster.refit()`, then I think it's important to include `params`.

I think it could work like this:

```python
new_params = _choose_param_value( 
     main_param_name="linear_tree", 
     params=self.params, 
     default_value=None 
 ) 
if dataset_params is None:
    dataset_params = {}

new_params.update(dataset_params)
```

shiyu1994:
> For example, I don't know if the comment mentioned at #3038 (comment) means that init_score, weight, and group is still true.

I think it is useful that we want to change the weights of data points when refit. If we see `weight`, `init_score` and `group` as properties of a dataset, then when refitting on a new dataset, it is very meaningful to allow new settings of these properties.

For example, a new ranking dataset can have totally different `group`.

tests/python_package_test/test_engine.py
```diff
@@ -1545,6 +1545,34 @@ def test_refit():
...
+        'metric': 'binary_logloss',
+        'verbose': -1,
+        'min_data': 10
+    }
+    gbm = lgb.train(params, lgb_train, num_boost_round=20)
+    non_weight_err_pred = log_loss(y_test, gbm.predict(X_test))
+    new_gbm = gbm.refit(
+        data=X_train,
+        label=y_train,
+        weight=np.abs(np.random.normal(size=y_train.shape)),
```
shiyu1994, line 1564: It seems that weights can be negative with `np.random.normal`.
`np.random.rand` should be a better option.
https://numpy.org/doc/stable/reference/random/generated/numpy.random.rand.html

shiyu1994: We don't want negative weights because it many in many cases the `hessian` value of a data is supposed to be non-negative. And `hessian` is multiplied by `weight`.

shiyu1994:
@TremaMiguel Thanks for working on this!

tests/python_package_test/test_engine.py
TremaMiguel: @shiyu1994 ok, an additional explanation to `weight` parameter documentation could be added

```
weight : list, numpy 1-D array, pandas Series or None, optional (default=None)
            Weight for each ``data`` instance. Weight should be non-negative values because the Hessian
            is supposed to be non-negative.
```

python-package/lightgbm/basic.py
```diff
@@ -3516,6 +3530,31 @@ def refit(self, data, label, decay_rate=0.9, **kwargs):
         decay_rate : float, optional (default=0.9)
             Decay rate of refit,
             will use ``leaf_output = decay_rate * old_leaf_output + (1.0 - decay_rate) * new_leaf_output`` to refit trees.
+        reference : Dataset or None, optional (default=None)
+            reference for ``data``.
+            If this is Dataset for validation, training data should be used as reference.
```
jameslamb, line 3535:
```suggestion
```

Please remove this line, which I think was copied from the `train()` docs. There isn't a concept of "validation" datasets for the `refit()` method.

tests/python_package_test/test_engine.py
```diff
@@ -1545,6 +1545,59 @@ def test_refit():
...
+        'header': False,
+        'label_column': "",
+        'weight_column': "",
+        'group_column': "",
+        'ignore_column': "",
+        'forcedbins_filename': "",
+        'save_binary': False,
+        'precise_float_parser': False,
+        'parser_config_file': "",
+    }
```
jameslamb, line 1584: Please see my comment from https://github.com/microsoft/LightGBM/pull/4894#discussion_r773321120.

If these parameter values are not all necessary for this test, please remove them from this code. If they are necessary, could you please explain why?

tests/python_package_test/test_engine.py
```diff
@@ -1545,6 +1545,59 @@ def test_refit():
...
+        data=X_train,
+        label=y_train,
+        weight=np.random.rand(y_train.shape[0]),
+        reference=None,
+        group=None,
+        init_score=None,
+        feature_name="auto",
+        categorical_feature="auto",
+        dataset_params=dataset_params,
+        free_raw_data=True
```
jameslamb, line 1595:
```suggestion
        dataset_params=dataset_params,
```

Since this test code is just setting all these parameters to the default values in `Booster.refit()`, please remove them from this call.

Keeping the test code focused only on what is minimally necessary makes it easier to understand what specifically is being testing. It also tests the behavior that user code does not need to specify all of these arguments.

tests/python_package_test/test_engine.py
```diff
@@ -1545,6 +1545,59 @@ def test_refit():
...
+def test_refit_dataset_params():
+    # check refit accepts dataset_params
+    X, y = load_breast_cancer(return_X_y=True)
+    X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.1, random_state=42)
+    lgb_train = lgb.Dataset(X_train, y_train)
+    train_params = {
+        'objective': 'binary',
+        'metric': 'binary_logloss',
+        'verbose': -1,
+        'min_data': 10
```
jameslamb, line 1557:
```suggestion
        'min_data': 10,
        'seed': 123
```

Please set a seed here, to remove a source of randomness in the training of the initial Booster.

jameslamb (changes requested):
Thanks for the testing changes. Please see a few more suggestions I've provided.

tests/python_package_test/test_engine.py
TremaMiguel: I'm setting them as default to test that `Dataset` "accepts" them, think of just leaving these three as they influence the behavior
```
'max_bin': 255,
'min_data_in_bin': 3,
'data_random_seed': 1
```

tests/python_package_test/test_engine.py
TremaMiguel: got it!

tests/python_package_test/test_engine.py
TremaMiguel: got it!

python-package/lightgbm/basic.py
TremaMiguel: got it

tests/python_package_test/test_engine.py
```diff
@@ -1545,6 +1545,35 @@ def test_refit():
...
 
 
+def test_refit_dataset_params():
+    # check refit accepts dataset_params
+    X, y = load_breast_cancer(return_X_y=True)
+    X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.1, random_state=42)
+    lgb_train = lgb.Dataset(X_train, y_train)
+    train_params = {
+        'objective': 'binary',
+        'metric': 'binary_logloss',
```
jameslamb, line 1555:
```suggestion
```

Since you aren't providing a validation set, `metric` is not necessary. Please remove it.

tests/python_package_test/test_engine.py
```diff
@@ -1545,6 +1545,35 @@ def test_refit():
...
+        'verbose': -1,
+        'min_data': 10,
```
jameslamb, line 1557:
```suggestion
```

I was confused by the appearance of `min_data` in the parameters passed to `lgb.train()`, as it's only used in `lgb.Dataset`: https://lightgbm.readthedocs.io/en/latest/Parameters.html#min_data_in_leaf

Please remove it here.

tests/python_package_test/test_engine.py
```diff
@@ -1545,6 +1545,35 @@ def test_refit():
...
+        'seed': 123
+    }
+    gbm = lgb.train(train_params, lgb_train, num_boost_round=20)
```
jameslamb, line 1560:
```suggestion
    gbm = lgb.train(train_params, lgb_train, num_boost_round=10)
```

To save some CI time, please reduce `num_boost_round` to 10.

tests/python_package_test/test_engine.py
```diff
@@ -1545,6 +1545,35 @@ def test_refit():
...
+    }
+    new_gbm = gbm.refit(
+        data=X_train,
+        label=y_train,
+        weight=np.random.rand(y_train.shape[0]),
+        dataset_params=dataset_params,
+    )
+    weight_err_pred = log_loss(y_test, new_gbm.predict(X_test))
+    assert weight_err_pred != non_weight_err_pred
+
```
jameslamb, line 1575:
```suggestion
    train_set_params = new_gbm.train_set.get_params()
    assert train_set_params["max_bin"] == 260
    assert train_set_params["min_data_in_bin"] == 5
    assert train_set_params["data_random_seed"] == 123
```

Since you are specifying `Dataset` parameters here, can you please test that they're passed through correctly?

Adding such tests will increase confidence that this change is working as expected.

jameslamb: And actually, now that I think about it...could you also directly extract the weight and compare it to the one that was passed in?

Like

```python
refit_weight = np.random.rand(y_train.shape[0])
new_gbm = gbm.refit(
        data=X_train,
        label=y_train,
        weight=refit_weight,
        dataset_params=dataset_params,
)
stored_weights = new_gbm.train_set.get_weight()
assert stored_weights == refit_weight
```

jameslamb (changes requested):
Thanks very much. Please see some additional suggestions to make the tests a bit stronger.

tests/python_package_test/test_engine.py
TremaMiguel: here I'll use [np.assert_all_close](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html#numpy.testing.assert_allclose) because tests array of float numbers, or do you have another thing in mind?

jameslamb: yep, sounds good!

TremaMiguel: @jameslamb ci is green, let me know if anything else is needed

jameslamb (approved):
Looks good to me, thanks very much!

I've edited the PR description so it will be a bit more informative when it's used as a bullet point in the release notes.

Before this is merged, I'd like another Python maintainer like @jmoralez or @StrikerRUS to review.

python-package/lightgbm/basic.py
```diff
@@ -3516,6 +3530,31 @@ def refit(self, data, label, decay_rate=0.9, **kwargs):
         decay_rate : float, optional (default=0.9)
             Decay rate of refit,
             will use ``leaf_output = decay_rate * old_leaf_output + (1.0 - decay_rate) * new_leaf_output`` to refit trees.
+        reference : Dataset or None, optional (default=None)
+            reference for ``data``.
```
StrikerRUS, line 3534:
```suggestion
            Reference for ``data``.
```

python-package/lightgbm/basic.py
```diff
@@ -3516,6 +3530,31 @@ def refit(self, data, label, decay_rate=0.9, **kwargs):
...
+        weight : list, numpy 1-D array, pandas Series or None, optional (default=None)
+            Weight for each ``data`` instance. Weight should be non-negative values because the Hessian
+            value multiplied by weight is supposed to be non-negative.
```
StrikerRUS, line 3537: Would you like to prepare another PR that will add this note in all other places where `weight` is used?

python-package/lightgbm/basic.py
```diff
@@ -3516,6 +3530,31 @@ def refit(self, data, label, decay_rate=0.9, **kwargs):
...
+        group : list, numpy 1-D array, pandas Series or None, optional (default=None)
+            Group/query size for ``data``.
```
StrikerRUS, line 3539: We need to keep descriptions of the same param in different places consistent with aim to not miss any such place during any updates.

```suggestion
            Group/query size for ``data``.
            Only used in the learning-to-rank task.
            sum(group) = n_samples.
            For example, if you have a 100-document dataset with ``group = [10, 20, 40, 10, 10, 10]``, that means that you have 6 groups,
            where the first 10 records are in the first group, records 11-30 are in the second group, records 31-70 are in the third group, etc.
```

python-package/lightgbm/basic.py
```diff
@@ -3516,6 +3530,31 @@ def refit(self, data, label, decay_rate=0.9, **kwargs):
...
+        init_score : list, numpy 1-D array, pandas Series or None, optional (default=None)
```
StrikerRUS, line 3540:
```suggestion
        init_score : list, list of lists (for multi-class task), numpy array, pandas Series, pandas DataFrame (for multi-class task), or None, optional (default=None)
```

python-package/lightgbm/basic.py
```diff
@@ -3516,6 +3530,31 @@ def refit(self, data, label, decay_rate=0.9, **kwargs):
...
+            Init score for ``data``.
+        feature_name : list of strings or 'auto', optional (default="auto")
```
StrikerRUS, line 3542: We use the name of real Python type to reference string type in our docstrings, refer to #4557.

```suggestion
        feature_name : list of str, or 'auto', optional (default="auto")
```

python-package/lightgbm/basic.py
```diff
@@ -3516,6 +3530,31 @@ def refit(self, data, label, decay_rate=0.9, **kwargs):
...
+            Feature names for ``data``.
+            If 'auto' and data is pandas DataFrame, data columns names are used.
+        categorical_feature : list of strings or int, or 'auto', optional (default="auto")
```
StrikerRUS, line 3545:
```suggestion
        categorical_feature : list of str or int, or 'auto', optional (default="auto")
```

python-package/lightgbm/basic.py
```diff
@@ -3516,6 +3530,31 @@ def refit(self, data, label, decay_rate=0.9, **kwargs):
...
+            Categorical features for ``data``.
+            If list of int, interpreted as indices.
+            If list of strings, interpreted as feature names (need to specify ``feature_name`` as well).
```
StrikerRUS, line 3548:
```suggestion
            If list of str, interpreted as feature names (need to specify ``feature_name`` as well).
```

tests/python_package_test/test_engine.py
```diff
@@ -1545,6 +1545,40 @@ def test_refit():
...
+        dataset_params=dataset_params,
+    )
+    weight_err_pred = log_loss(y_test, new_gbm.predict(X_test))
+    train_set_params = new_gbm.train_set.get_params()
+    stored_weights = new_gbm.train_set.get_weight()
+    assert weight_err_pred != non_weight_err_pred
+    assert train_set_params["max_bin"] == 260
+    assert train_set_params["min_data_in_bin"] == 5
+    assert train_set_params["data_random_seed"] == 123
+    np.testing.assert_allclose(stored_weights, refit_weight, verbose=True)
```
StrikerRUS, line 1579: For the consistency with the rest codebase don't explicitly set this default value.
```suggestion
    np.testing.assert_allclose(stored_weights, refit_weight)
```

StrikerRUS (changes requested):
Thanks for working on this! Generally LGTM, just few minor fixes for the docstring.

tests/python_package_test/test_engine.py
```diff
@@ -1545,6 +1545,40 @@ def test_refit():
     assert err_pred > new_err_pred
 
 
+def test_refit_dataset_params():
+    # check refit accepts dataset_params
+    X, y = load_breast_cancer(return_X_y=True)
+    X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.1, random_state=42)
+    lgb_train = lgb.Dataset(X_train, y_train)
```
jmoralez, line 1552: I've never used `refit` before but it seems that it sets `init_score=0`. So if we use `refit` with the same data and `decay_rate=0.0` I'd expect to get the same results, however that seems to only be true if we set `init_score=0` in the first booster. So with the current test, the assertion about the different scores passes even if we send in the same data to `refit` without modifying any of the dataset arguments.
```suggestion
    lgb_train = lgb.Dataset(X_train, y_train, init_score=np.zeros(y_train.size))
```

tests/python_package_test/test_engine.py
```diff
@@ -1545,6 +1545,40 @@ def test_refit():
...
+        'max_bin': 260,
+        'min_data_in_bin': 5,
+        'data_random_seed': 123,
+    }
+    new_gbm = gbm.refit(
+        data=X_train,
+        label=y_train,
+        weight=refit_weight,
+        dataset_params=dataset_params,
+    )
```
jmoralez, line 1571:
```suggestion
        dataset_params=dataset_params,
        decay_rate=0.0,
    )
```

tests/python_package_test/test_engine.py
```diff
@@ -1545,6 +1545,40 @@ def test_refit():
     assert err_pred > new_err_pred
 
 
+def test_refit_dataset_params():
+    # check refit accepts dataset_params
+    X, y = load_breast_cancer(return_X_y=True)
+    X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.1, random_state=42)
```
jmoralez, line 1551: Do we really need this split? I think it would be enough to check that the predictions over the training set are different between the original and the refitted booster.

jmoralez (changes requested):

python-package/lightgbm/basic.py
TremaMiguel: got it

python-package/lightgbm/basic.py
TremaMiguel: yes, once this PR is done. Reference: Issue #4921

python-package/lightgbm/basic.py
TremaMiguel: got it

python-package/lightgbm/basic.py
TremaMiguel: got it

tests/python_package_test/test_engine.py
TremaMiguel: got it

tests/python_package_test/test_engine.py
TremaMiguel: Both work, we just want to test that predictions values or metric changes when calling `refit`, it can be in `train` or `test` set.

I think we can add an additional test to proof the change in the `train` set, what are your thoughts @jmoralez ?

tests/python_package_test/test_engine.py
TremaMiguel: To illustrate what you say

```python
from sklearn.metrics import log_loss

import lightgbm as lgb
import numpy as np

#lgb_train = lgb.Dataset(X_train, y_train)
lgb_train = lgb.Dataset(X_train, y_train, init_score=np.zeros(y_train.size))
train_params = {
    'objective': 'binary',
    'verbose': -1,
    'seed': 123
}
gbm = lgb.train(train_params, lgb_train, num_boost_round=10)
non_weight_err_pred = log_loss(y_test, gbm.predict(X_test))
#refit_weight = np.random.rand(y_train.shape[0])
dataset_params = {
    #'max_bin': 260,
    #'min_data_in_bin': 5,
    'data_random_seed': 123,
}
new_gbm  = gbm.refit(
    data=X_train,
    label=y_train,
    #weight=refit_weight,
    dataset_params=dataset_params,
    decay_rate=0.0,
)
weight_err_pred = log_loss(y_test, new_gbm.predict(X_test))
weight_err_pred
>>> 0.2682710111899916
non_weight_err_pred
>>> 0.2682710111899916
```

but if you set train data without init_score `lgb_train = lgb.Dataset(X_train, y_train)` you get different scores

```
weight_err_pred
>>> 0.1442188066230218
non_weight_err_pred
>>> 0.24365628110393137

Comment (TremaMiguel) [reply tests/python_package_test/test_engine.py:1552]: Not sure if this has something to do with how `init_score` is initialized for [_lazy_init](https://github.com/microsoft/LightGBM/blob/44a9120197fbfc7f3bfb9349e4eb2e9443c1971b/python-package/lightgbm/basic.py#L973). Apparently `init_score` will remain `None`.

File: tests/python_package_test/test_engine.py (hunk 24 above)
Comment (jameslamb) [reply tests/python_package_test/test_engine.py:1551]: > I think we can add an additional test to proof the change in the train set

You don't need to add a second test. I agree with @jmoralez ... calling `.predict()` on any data will be enough for this unit test. Please remove the `train_test_split()` and just create predictions on `X`.

File: python-package/lightgbm/basic.py [hunk 25]
```diff
@@ -3516,6 +3530,35 @@ def refit(self, data, label, decay_rate=0.9, **kwargs):
...
+            value multiplied by weight is supposed to be non-negative.
+        group : list, numpy 1-D array, pandas Series or None, optional (default=None)
+            Group/query size for ``data``.
+            Only used in the learning-to-rank task.
+            sum(group) = n_samples.
+            For example, if you have a 100-document dataset with ``group = [10, 20, 40, 10, 10, 10]``, that means that you have 6 groups,
+            where the first 10 records are in the first group, records 11-30 are in the second group, records 31-70 are in the third group, etc.
+        init_score : list, list of lists (for multi-class task), numpy array, pandas Series, pandas DataFrame (for multi-class task), or None, optional (default=None)
+            Init score for ``data``.
+        feature_name : list of str or 'auto', optional (default="auto")
```
StrikerRUS, line 3546: Just one missed part from my previous review. This comma makes new docstring consistent with all other similar docstrings across the package and prevents readers from treating this as `list of "auto" words`.
```suggestion
        feature_name : list of str, or 'auto', optional (default="auto")
```

python-package/lightgbm/basic.py
StrikerRUS: Thank you very much for documenting this!

tests/python_package_test/test_engine.py
TremaMiguel: @StrikerRUS @jameslamb any comment on this?, is this related to this pr ?

tests/python_package_test/test_engine.py
TremaMiguel: got it!

python-package/lightgbm/basic.py
TremaMiguel: got it

TremaMiguel:
@StrikerRUS @jameslamb @jmoralez is there any pending change to close this PR?

tests/python_package_test/test_engine.py
jameslamb: Please wait for @jmoralez to respond to you in this thread. This project is maintained by a team of mostly volunteers who review these PRs on nights and weekends, so please be patient with us.

jmoralez: I'd like these changes to be made to the current tests (setting init_score to zero and the decay rate to 0) so we can be sure that the changes in the refitted booster are only due to the different dataset. The fact that refit sets init_score to 0 might be a bug and may be worth opening an issue about it, but for this PR once these changes are made I think it's good to go.

TremaMiguel: ok @jmoralez I've made corresponding changes. For your information, issue #4951 is open if you would like to add more context.

jmoralez (approved):

StrikerRUS (approved):
LGTM! Thank you so much for your contribution!

Issue #4793, opened 2021-11-12
[tests][python-package] Modify tests using boston housing dataset


jmoralez:
## Description
It was recently brought to light that the boston housing dataset has an ethical problem (which is more thoroughly described in https://github.com/scikit-learn/scikit-learn/issues/16155) and thus will be removed from scikit-learn in version 1.2 ([Olivier Grisel's explanation](https://twitter.com/ogrisel/status/1442894248488046595)). Since some of the tests for the python package rely on this dataset ([sample logs with the warning](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=11536&view=logs&j=c28dceab-947a-5848-c21f-eef3695e5f11&t=fa158246-17e2-53d4-5936-86070edbaacf&l=553)) they should be changed to use a different dataset.

Tests currently using the boston dataset:
- [ ] test_engine::test_regression
- [ ] test_engine::test_continue_train
- [x] test_engine::test_continue_train_reused_dataset
- [ ] test_engine::test_continue_train_dart
- [x] test_engine::test_cv
- [x] test_engine::test_feature_name
- [x] test_engine::test_save_load_copy_pickle
- [ ] test_engine::test_mape_rf
- [ ] test_engine::test_mape_dart
- [x] test_engine::test_model_size
- [ ] test_engine::test_get_split_value_histogram
- [ ] test_engine::test_early_stopping_for_only_first_metric
- [x] test_engine::test_extra_trees
- [x] test_engine::test_path_smoothing
- [ ] test_engine::test_interaction_constraints
- [x] test_engine::test_predict_with_start_iteration
- [ ] test_engine::test_force_split_with_feature_fraction
- [ ] test_sklearn::test_regression
- [x] test_sklearn::test_objective_aliases
- [ ] test_sklearn::test_regression_with_custom_objective
- [ ] test_sklearn::test_dart
- [ ] test_sklearn::test_stacking_regressor
- [x] test_sklearn::test_clone_and_property
- [x] test_sklearn::test_joblib
- [x] test_sklearn::test_non_serializable_objects_in_callbacks
- [x] test_sklearn::test_evaluate_train_set
- [ ] test_sklearn::test_metrics
- [ ] test_sklearn::test_first_metric_only
- [x] test_sklearn::test_training_succeeds_when_data_is_dataframe_and_label_is_column_array

StrikerRUS:
Thanks for writing this up!
[...]

PR #4895, opened 2021-12-19
[tests][python-package] change boston dataset to synthetic dataset in tests that don't check score

jmoralez:
Contributes to #4793.

Replaces the `load_boston` function with `make_synthetic_regression` (which calls `sklearn.datasets.make_regression`) in the tests that don't make assertions about the score of the model.

### DIFF
--- a/tests/python_package_test/test_engine.py
+++ b/tests/python_package_test/test_engine.py
@@ -18,7 +18,7 @@ from sklearn.model_selection import GroupKFold, TimeSeriesSplit, train_test_spli
 
 import lightgbm as lgb
 
-from .utils import load_boston, load_breast_cancer, load_digits, load_iris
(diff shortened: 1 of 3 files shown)


tests/python_package_test/test_engine.py
```diff
@@ -2102,7 +2102,7 @@ def test_default_objective_and_metric():
 
 @pytest.mark.skipif(psutil.virtual_memory().available / 1024 / 1024 / 1024 < 3, reason='not enough RAM')
 def test_model_size():
-    X, y = load_boston(return_X_y=True)
+    X, y = make_synthetic_regression()
```
StrikerRUS, line 2105: I checked all CI jobs and there are no new skips compared to `master` due to dataset replacement.

StrikerRUS (approved):
Thanks a lot!

jameslamb (approved):
excellent, thanks!

Issue #4667, opened 2021-10-09
[R-package] `lgb.cv` ignores `verbose` argument


david-cortes:
In the R version, function `lgb.cv` has a parameter `verbose` with this description:
```
verbosity for output, if <= 0, also will disable the print of evaluation during training
```

However, regardless of what I pass for verbose, it will still print `[Info]` and higher-up messages:
```r
library(lightgbm)
data("agaricus.train")
pool = lgb.Dataset(agaricus.train$data, label=agaricus.train$label)
params = list(
    objective_type="binary",
    force_row_wise=TRUE,
    seed=123
)
lgb.cv(params, pool, verbose=0)
```
```
[LightGBM] [Info] [binary:BoostFromScore]: pavg=0.486412 -> initscore=-0.054366
[LightGBM] [Info] Start training from score -0.054366
[LightGBM] [Warning] No further splits with positive gain, best gain: -inf
[LightGBM] [Info] [binary:BoostFromScore]: pavg=0.476969 -> initscore=-0.092189
[LightGBM] [Info] Start training from score -0.092189
[LightGBM] [Warning] No further splits with positive gain, best gain: -inf
[LightGBM] [Info] [binary:BoostFromScore]: pavg=0.482957 -> initscore=-0.068198
[LightGBM] [Info] Start training from score -0.068198
[LightGBM] [Warning] No further splits with positive gain, best gain: -inf
...
```

jameslamb:
Assigning this to myself to signal that I'm working on it right now.

PR #4899, opened 2021-12-21
[R-package] update parameter 'verbosity' based on keyword arg 'verbose'

jameslamb:
Related to #4667.
Related to #4862.

While investigating #4667, I noticed a small inconsistency in `lgb.train()` and `lgb.cv()`. Those functions have a keyword argument `verbose` and update the parameter `verbose` in parameters passed down to `LGBM_BoosterCreate`.

https://github.com/microsoft/LightGBM/blob/7b10baffb6e600862b087a4207343b0b58eed2b5/R-package/R/lgb.cv.R#L105

https://github.com/microsoft/LightGBM/blob/7b10baffb6e600862b087a4207343b0b58eed2b5/R-package/R/lgb.train.R#L77

As a result, if you pass different values for `verbosity` (in `params`) and `verbose` (via keyword arguments), then the Booster's parameters will contain both `verbosity` and `verbose`.

Consider this example:

```r
library(lightgbm)
data("agaricus.train")
    data = agaricus.train$data
)
bst <- lightgbm::lgb.train(
    params = list(
        num_iterations = 2L
        , verbosity = 1L
    )
    , data = dtrain
    , obj = "binary"
    , verbose = -1L
)
bst$params
```

> $verbosity
[1] 1

> $verbose
[1] -1

Note that `verbosity` is the main parameter recognized by LightGBM (https://lightgbm.readthedocs.io/en/latest/Parameters.html#verbosity) and `verbose` is an alias for it.

This PR proposes two changes:

1. Update `params$verbosity` (the main parameter name, instead of an alias) based on the value of keyword argument `verbose`
2. Use internal function `lgb.check.wrapper_param()` to ensure consistent behavior when processing keyword arguments.

### DIFF
--- a/R-package/tests/testthat/test_basic.R
+++ b/R-package/tests/testthat/test_basic.R
@@ -1534,6 +1534,60 @@ test_that("lgb.train() works with integer, double, and numeric data", {
   }
 })
 
+test_that("lgb.train() updates params based on keyword arguments", {
+  dtrain <- lgb.Dataset(
+    data = matrix(rnorm(400L), ncol =  4L)
+    , label = rnorm(100L)
+  )
+
+  # defaults from keyword arguments should be used if not specified in params
+  invisible(
+    capture.output({
+      bst <- lgb.train(
+        data = dtrain
+        , obj = "regression"
+        , params = list()
+      )
+    })
+  )
+  expect_equal(bst$params[["verbosity"]], 1L)
+  expect_equal(bst$params[["num_iterations"]], 100L)
+
+  # main param names should be preferred to keyword arguments
+  invisible(
+    capture.output({
+      bst <- lgb.train(
+        data = dtrain
+        , obj = "regression"
+        , params = list(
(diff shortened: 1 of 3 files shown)


R-package/tests/testthat/test_basic.R
```diff
@@ -1534,6 +1534,60 @@ test_that("lgb.train() works with integer, double, and numeric data", {
...
 })
 
+test_that("lgb.train() updates params based on keyword arguments", {
+  dtrain <- lgb.Dataset(
+    data = matrix(rnorm(400L), ncol =  4L)
+    , label = rnorm(100L)
+  )
+
+  # defaults from keyword arguments should be used if not specified in params
+  invisible(
```
jameslamb, line 1544: these uses of `invisible(capture.output(...))` suppress log messages from being printed, per #4862

StrikerRUS (approved):
Nice improvement!

PR #4902, opened 2021-12-21
[R-package] remove unnecessary comments and fix typos in comments

jameslamb:
I've been looking more closely at the R package recently, and noticed a few comments that are either uninformative or provide typos. This PR proposes cleaning them up.

### DIFF
--- a/R-package/R/lgb.cv.R
+++ b/R-package/R/lgb.cv.R
@@ -550,7 +550,7 @@ lgb.stratified.folds <- function(y, k) {
 
       ## Create a vector of integers from 1:k as many times as possible without
       ## going over the number of samples in the class. Note that if the number
-      ## of samples in a class is less than k, nothing is producd here.
+      ## of samples in a class is less than k, nothing is produced here.
       seqVector <- rep(seq_len(k), numInClass[i] %/% k)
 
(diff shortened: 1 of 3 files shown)


R-package/R/lgb.cv.R
```diff
@@ -550,7 +550,7 @@ lgb.stratified.folds <- function(y, k) {
 
       ## Create a vector of integers from 1:k as many times as possible without
       ## going over the number of samples in the class. Note that if the number
-      ## of samples in a class is less than k, nothing is producd here.
+      ## of samples in a class is less than k, nothing is produced here.
```
StrikerRUS, line 553: Just out of my curiosity, why double `##` here?

StrikerRUS (approved):

R-package/R/lgb.cv.R
jameslamb: no reason at all! According to the git blame, these lines were added 5 years ago

https://github.com/microsoft/LightGBM/blame/0df38106b48161e12204de5621e406cc6c19a3e6/R-package/R/lgb.cv.R#L512-L519

I'd support a PR changing them (and other instances like this) to a single `#`

StrikerRUS: OK, got it! 🙂

PR #4903, opened 2021-12-21
[R-package] respect 'verbose' argument in lgb.cv() (fixes #4667)

### LINKED ISSUE #4667: [R-package] `lgb.cv` ignores `verbose` argument

jameslamb:
Fixes #4667.

On `master`, the tests added in this PR fail and the code below prints INFO and WARNING level logs.

```r
library(lightgbm)
data("agaricus.train")
dtrain <- lgb.Dataset(
    data = agaricus.train$data
)
lgb.cv(
    params = list()
    , data = dtrain
    , verbose = -1L
    , nrounds = 3L
    , nfold = 2L
    , obj = "binary"
)
```

As of this PR, the tests pass and no logs are produced.

### DIFF
--- a/R-package/R/lgb.cv.R
+++ b/R-package/R/lgb.cv.R
@@ -102,6 +102,7 @@ lgb.cv <- function(params = list()
   }
 
   # Setup temporary variables
+  params$verbose <- verbose
   params <- lgb.check.obj(params = params, obj = obj)
   params <- lgb.check.eval(params = params, eval = eval)
   fobj <- NULL
@@ -111,11 +112,6 @@ lgb.cv <- function(params = list()
   # in `params`.
   # this ensures that the model stored with Booster$save() correctly represents
   # what was passed in
-  params <- lgb.check.wrapper_param(
-    main_param_name = "verbosity"
-    , params = params
-    , alternative_kwarg_value = verbose
-  )
   params <- lgb.check.wrapper_param(
     main_param_name = "num_iterations"
     , params = params
@@ -290,7 +286,7 @@ lgb.cv <- function(params = list()
   # someDT$some_column returns NULL is 'some_column' does not exist in the data.table
   bst_folds <- lapply(
     X = seq_along(folds)
-    , FUN = function(k, params) {
+    , FUN = function(k) {
 
       # For learning-to-rank, each fold is a named list with two elements:
       #   * `fold` = an integer vector of row indices
@@ -336,15 +332,12 @@ lgb.cv <- function(params = list()
       }
 
       booster <- Booster$new(params = params, train_set = dtrain)
-      print(sprintf("params for fold %i", k))
-      print(params)
       booster$add_valid(data = dtest, name = "valid")
       booster$reset_parameter(params)
       return(
         list(booster = booster)
       )
     }
-    , params = params
   )
 
   # Create new booster
--- a/R-package/R/lgb.cv.R
+++ b/R-package/R/lgb.cv.R
@@ -12,8 +12,8 @@ CVBooster <- R6::R6Class(
       return(invisible(NULL))
     },
     reset_parameter = function(new_params) {
-      for (x in boosters) {
-        x$reset_parameter(params = new_params)
+      for (x in self$boosters) {
+        x[["booster"]]$reset_parameter(params = new_params)
       }
       return(invisible(self))
     }
@@ -551,7 +551,7 @@ lgb.stratified.folds <- function(y, k) {
 
       ## Create a vector of integers from 1:k as many times as possible without
       ## going over the number of samples in the class. Note that if the number
-      ## of samples in a class is less than k, nothing is producd here.
+      ## of samples in a class is less than k, nothing is produced here.
       seqVector <- rep(seq_len(k), numInClass[i] %/% k)
 
       ## Add enough random integers to get length(seqVector) == numInClass[i]
--- a/R-package/R/lgb.cv.R
+++ b/R-package/R/lgb.cv.R
@@ -102,7 +102,6 @@ lgb.cv <- function(params = list()
   }
 
   # Setup temporary variables
-  params$verbose <- verbose
   params <- lgb.check.obj(params = params, obj = obj)
   params <- lgb.check.eval(params = params, eval = eval)
   fobj <- NULL
@@ -112,6 +111,11 @@ lgb.cv <- function(params = list()
   # in `params`.
   # this ensures that the model stored with Booster$save() correctly represents
   # what was passed in
+  params <- lgb.check.wrapper_param(
+    main_param_name = "verbosity"
+    , params = params
+    , alternative_kwarg_value = verbose
+  )
   params <- lgb.check.wrapper_param(
     main_param_name = "num_iterations"
     , params = params
--- a/R-package/R/lgb.cv.R
+++ b/R-package/R/lgb.cv.R
@@ -344,7 +344,6 @@ lgb.cv <- function(params = list()
         booster$add_valid(data = dtrain, name = "train")
       }
       booster$add_valid(data = dtest, name = "valid")
-      booster$reset_parameter(params)
       return(
         list(booster = booster)
       )
--- a/R-package/R/lgb.cv.R
+++ b/R-package/R/lgb.cv.R
@@ -102,7 +102,6 @@ lgb.cv <- function(params = list()
   }
 
   # Setup temporary variables
-  params$verbose <- verbose
   params <- lgb.check.obj(params = params, obj = obj)
   params <- lgb.check.eval(params = params, eval = eval)
   fobj <- NULL
@@ -112,6 +111,11 @@ lgb.cv <- function(params = list()
   # in `params`.
   # this ensures that the model stored with Booster$save() correctly represents
   # what was passed in
+  params <- lgb.check.wrapper_param(
+    main_param_name = "verbosity"
+    , params = params
+    , alternative_kwarg_value = verbose
+  )
   params <- lgb.check.wrapper_param(
     main_param_name = "num_iterations"
     , params = params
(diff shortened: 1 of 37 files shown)


R-package/R/lgb.cv.R
```diff
@@ -333,6 +333,7 @@ lgb.cv <- function(params = list()
 
       booster <- Booster$new(params = params, train_set = dtrain)
       booster$add_valid(data = dtest, name = "valid")
+      booster$reset_parameter(params)
```
jameslamb, line 336: note to reviewers: I don't totally understand *why* this change fixes #4667. At first I was worried that this function might suffer from a larger problem like "`lgb.cv()` does not respect values passed to `params`", but that is not true.

Consider the following example. Setting `min_gain_to_split` to a large number for the binary objective results in no splits, as expected.

```r
library(lightgbm)
data("agaricus.train")
dtrain <- lgb.Dataset(
    data = agaricus.train$data
)
lgb.cv(
    params = list(
        min_gain_to_split = 100000.0
    )
    , data = dtrain
    , verbose = 0L
    , nrounds = 3L
    , nfold = 2L
    , obj = "binary"
)
```

The logs are full of the following

> [LightGBM] [Warning] No further splits with positive gain, best gain: -inf

Changing to `min_gain_to_split = 0.0`, those warnings go away and training produces trees with multiple leaves.

StrikerRUS: Maybe this related to the lack of corresponding code lines in R-package?
https://github.com/microsoft/LightGBM/blob/2ef3cb81ed66706ed12282857635f0a5941d92ad/python-package/lightgbm/engine.py#L222-L225

jameslamb: Maybe? I'm not really sure. The code in `lgb.train()` and `lgb.cv()` in the R package is very similar, and `lgb.train()` doesn't have this issue with the parameter `verbose`.

Anyway, I think this fix is still worth merging even if it isn't fully understood...I don't think adding another `$update_params()` with the passed-in params introduces any risk.

StrikerRUS: > I don't think adding another `$update_params()` with the passed-in params introduces any risk.

Agree. But this introduces performance degradation because this method calls C API function under the hood and is being executed for each fold.

jameslamb: > but this introduces performance degradation

Sure, but I think that in this case, the cost of calling this method once per fold (one time at set up, not on each iteration, just to be clear) is worth it in exchange for ensuring that `lgb.cv()`'s behavior is correct.

StrikerRUS (approved):
Despite that I'm not fully satisfied by this fix that causes performance degradation and believe there can be a way of specifying params without resetting them, I don't want to block this PR from merging. Please consider this my approval just as a comment that allows you to click "Merge" button. This isn't a "real" review because I don't understand why this fixes original issue (as you do https://github.com/microsoft/LightGBM/pull/4903#discussion_r772847332).

R-package/R/lgb.cv.R
StrikerRUS: OK, I'm letting you to decide.

jameslamb:
Ok thanks @StrikerRUS ! I'm going to try to investigate a bit more before merging this.

R-package/R/lgb.cv.R
jameslamb: Ok looking into this more closely tonight, I think I figured out:

1. why `verbose` is not being respected
2. why it seems like ONLY `verbose` is not being respected and all other parameters are working correctly

## Short Summary

Two bugs in the R package:

1. `Dataset$slice()` calls `LGBM_DatasetGetSubset()`, which will globally set the log level to INFO if `verbosity` isn't in the parameters passed to it. And `verbosity` is never passed through.
2. When `Dataset$update_params(new_params)` is called on a `Dataset` that has already been constructed, it doesn't store `new_params` in memory in the R object (`private$params`)

I just pushed https://github.com/microsoft/LightGBM/pull/4903/commits/619eaf484232642d9e231a5186a1d2cf2c48d74a which reverts the addition of `booster$reset_parameter()` and adds fixes for these two bugs.

## Much Longer Summary

I was able to reduce the internals of `lgb.cv()` to the following code that helps narrow down the issue.

<details><summary>sample code (click me)</summary>

```r
library(lightgbm)

data("agaricus.train", package = "lightgbm")

params <- list(
    max_bin = 123
    , verbose = -7L
)

dtrain <- lgb.Dataset(
    data = agaricus.train$data
    , params = params
)
dtest <- slice(
    dataset = dtrain
    , idxset = seq_len(100L)
)

# with a validation set, you get logs
booster_w_valid <- lightgbm:::Booster$new(
    params = params
    , train_set = dtrain
)
booster_w_valid$add_valid(data = dtest, name = "valid")
booster_w_valid$update()

# without a validation set, no logs!
booster_no_valid <- lightgbm:::Booster$new(
    params = params
    , train_set = dtrain
)
booster_no_valid$update()
```

</details>

root cause: The use of `slice()` on a `Dataset` resets LightGBM's log level globally to the default value (INFO).

More details:

1. `Dataset$slice()` gets params from `Dataset$get_params()`
    * https://github.com/microsoft/LightGBM/blob/af5b40e1f653e25a5f3470e09de4d1c68dc59663/R-package/R/lgb.Dataset.R#L538
2. `Dataset$get_params()` filters down to only Dataset-specific parameters.
    * https://github.com/microsoft/LightGBM/blob/af5b40e1f653e25a5f3470e09de4d1c68dc59663/R-package/R/lgb.Dataset.R#L583-L592
3. `verbose` is not in the Dataset-specific parameters.
    * https://github.com/microsoft/LightGBM/blob/af5b40e1f653e25a5f3470e09de4d1c68dc59663/R-package/R/aliases.R#L9
4. `Dataset$slice()` eventually calls `Dataset$construct()` which calls `LGBM_DatasetGetSubset()`, which runs `config.Set()`
    * https://github.com/microsoft/LightGBM/blob/af5b40e1f653e25a5f3470e09de4d1c68dc59663/src/c_api.cpp#L1407
5. `config.Set()` calls `Log::ResetLogLevel()` which resets log level **globally** for the current process.
    * https://github.com/microsoft/LightGBM/blob/af5b40e1f653e25a5f3470e09de4d1c68dc59663/src/io/config.cpp#L241-L249
6. If `verbosity` isn't found in `params`, it defaults to INFO-level
    * https://github.com/microsoft/LightGBM/blob/af5b40e1f653e25a5f3470e09de4d1c68dc59663/include/LightGBM/config.h#L543

`Dataset.subset()` in the Python package doesn't suffer from this issue because it references `self.params` (the dictionary of parameters stored as an attribute on the `Dataset` instance) instead of using `Dataset.get_params()`, so `verbose` is not filtered out.

https://github.com/microsoft/LightGBM/blob/af5b40e1f653e25a5f3470e09de4d1c68dc59663/python-package/lightgbm/basic.py#L1860

At first, I tried to just fix this by adding `"verbosity"` to the list of parameters returned by `Dataset$get_params()`. Unfortunately, that can lead to a new type of problem where the value of `verbosity` stored in a constructed Dataset will take precedence over values passed to `lgb.train()` / `lgb.cv()` if that Dataset is re-used multiple times.

https://github.com/microsoft/LightGBM/blob/af5b40e1f653e25a5f3470e09de4d1c68dc59663/R-package/R/lgb.Booster.R#L42

Next, I tried just changing the use of `Dataset$get_params()` to use `Dataset$private$params` in `Dataset$slice()`. That uncovered another bug. That still suffered from the problem that once the training Dataset was constructed, verbosity was locked in to the value from the last time it was constructed.

1. When calling `Dataset$construct()`, if it's already been constructed then R calls `LGBM_DatasetUpdateParamChecking()`
    * https://github.com/microsoft/LightGBM/blob/af5b40e1f653e25a5f3470e09de4d1c68dc59663/R-package/R/lgb.Dataset.R#L562
2. That call is try-catched. If it fails, then `Dataset$construct()` does the right thing based on whether or not raw data is still available. However, if that call does NOT fail, `private$params` is not updated!
    * https://github.com/microsoft/LightGBM/blob/af5b40e1f653e25a5f3470e09de4d1c68dc59663/R-package/R/lgb.Dataset.R#L559-L577

jameslamb:
>  I'm going to try to investigate a bit more

Happy to say I found the root cause for this bug, and was able to revert the additional call to `Booster$reset_parameter()`! See https://github.com/microsoft/LightGBM/pull/4903#discussion_r777294852 for a detailed description.

This taught me something important about LightGBM's internals. Any time `config.Set()` is called, it will globally reset LightGBM's log level.

https://github.com/microsoft/LightGBM/blob/af5b40e1f653e25a5f3470e09de4d1c68dc59663/src/io/config.cpp#L241-L249

This becomes important considering https://github.com/microsoft/LightGBM/issues/4904#issuecomment-1003805133

> throughout LightGBM, "params" isn't used to represent "all of the configuration for LightGBM" but more like "overrides to default configuration values"

Any code creating a `config` object on the C/C++ side based on a `params_str` that doesn't include `verbosity` will end up resetting the log level to INFO for the current process, until some other calls sets it to somethingg else.

StrikerRUS (approved):
Brilliant investigation! I'm super excited that you've uncovered deep underlying bug. Thank you very much!

Issue #4483, opened 2021-07-22
Warn when passing labels with missing values


ChristophAymannsQC:
## Summary

Currently (it seems to me) missing values (`np.nan`) are silently replaced by 0s when `objective=regression` (I haven't checked other objective functions), possibly related to [this](https://github.com/microsoft/LightGBM/blob/d517ba12f2e7862ac533908304dddbd770655d2b/src/io/metadata.cpp#L319). It would be nice if this was made transparent to the user who might be inadvertently passing missing values. Alternatively, observations with missing values could be dropped.

## Minimal example

```python
import lightgbm as lgb
import numpy as np
import pandas as pd
# %%
rng = np.random.default_rng(42)
x = rng.normal(size=100)
y = 3 * x + rng.normal(size=100)
X = pd.DataFrame(data={"x": x})
y[rng.uniform(size=100) < 0.1] = np.nan
# %%
data = lgb.Dataset(X, label=y)
print(data.get_label())
print(y)
# %%
eval_res = {}
booster = lgb.train(
    params={"objective": "regression", "verbosity": -1},
    train_set=data,
    valid_sets=[data],
    verbose_eval=False,
    evals_result=eval_res,
)
starting_loss = eval_res["training"]["l2"][0]
# %%
print(data.get_label())
# %%
y_dash = y.copy()
y_dash[np.isnan(y)] = 0
data_dash = lgb.Dataset(X, label=y_dash)
# %%
eval_res_dash = {}
booster = lgb.train(
    params={"objective": "regression", "verbosity": -1},
    train_set=data_dash,
    valid_sets=[data_dash],
    verbose_eval=False,
    evals_result=eval_res_dash,
)
starting_loss_dash = eval_res_dash["training"]["l2"][0]
# %%
assert starting_loss_dash == starting_loss
```

kurchi1205:
Hi , I want to make my first contribution . Can I work on this?

ChristophAymannsQC:
Sure. Please do give it a shot.

jameslamb:
Thanks for starting this discussion, @ChristophAymannsQC !

@kurchi1205 before you get too far into trying to implement this, I think we should get maintainers' opinions on this question. I have some initial thoughts.

[...]

Issue #4040, opened 2021-03-02
missing values during prediction should throw an Exception if missing data wasn't present during training


rmminusrslash:
## Summary

When we encounter missing values during prediction after training without any missing data, the model predicts these examples without logging a warning or an exception (see issue [this issue])(https://github.com/microsoft/LightGBM/issues/2921):

- Missing numerical values set to zero
- Missing categorical values are sent to the right leaf 

As a result, several or all features could be missing and the model would still return a prediction (of unknown quality). 

I propose to at least log a warning and allow the model to be configured in a strict mode where unexpected missing values lead to an exception (I would argue this should be the default, but it might not work for compatibility reasons).

## Motivation
Changing the current behavior is important for using lightgbm in production. When working with a train-test-split missing data in the testset is easily recognized and a difference between test and train is less likely than a difference between training and production data.

In production, data or code bugs can lead to one or multiple features being missing. In my experience, bugs that change the data happen as commonly as other bugs. 

The current behavior would silently impute them to zero (numerical case) or assign them to an existing leaf (categorical case). The model would silently misbehave and it could be hard to detect, especially if the bug is only on the inference side, but not on the training data (which is typical when the data is not coming from a common feature store).

## Description
Throw an exception when missing values are seen during inference but not during training. Value imputing should probably be done before calling the model, so I propose to make throwing an exception the default behaviour. 

You might not agree (hence the current implementation), so maybe it could be an option to log a warning?

It would also be great to improve the documentation on the use_missing=false flag:

```use_missing 🔗︎, default = true, type = bool
set this to false to disable the special handle of missing value
```
The doc string doesn't give an explanation of what is done instead during training and inference when disabling the special value handling. 

## References
https://github.com/microsoft/LightGBM/issues/2921
https://lightgbm.readthedocs.io/en/latest/Parameters.html

StrikerRUS:
Closed in favor of being in #2302. We decided to keep all feature requests in one place.

Welcome to contribute this feature! Please re-open this issue (or post a comment if you are not a topic starter) if you are actively working on implementing this feature.

PR #4909, opened 2021-12-24
[python-package] check feature names in predict with dataframe (fixes #812)

### LINKED ISSUE #812: [python package]: suggestion: lgb.Booster.predict() should check that the input X data makes sense

jmoralez:
This adds a check to verify that the feature names of the trained booster match the column names of a pandas dataframe when predicting with one and setting `validate_features=True`.

I believe this closes #812, since the input shape is already checked at the C++ side and the dataset object can't be constructed with repeated feature names.


jmoralez:
The tests are failing at https://github.com/microsoft/LightGBM/blob/cace5bbd8ee8185b96fb2f6734578db099f08a3a/examples/python-guide/advanced_example.py#L84 where the model was trained with feature names that are different than the column names of the `X_test` dataframe. This seems more like a feature, since even though in this case the features are correct, these are the types of cases that this PR aims to help detect and allow the user to handle. Should I change that file to make the column names equal to the feature names?

jameslamb (commented):
I don't support adding checks like this to `Booster.predict()`. This extra overhead might have a noticeable impact on time-per-prediction if you're serving a LightGBM model and doing single-row predictions.

I understand that "columns are in the wrong order" is a failure mode to guard against in deploying machine learning models, but I'm not convinced that the `Booster` object (or even the `lightgbm` library) is the right place for such checks. Doing this in the way currently proposed in this PR makes the overhead of this check unavoidable, which latency-sensitive users might not be happy about.

I feel the same way about requests like #4040.

If @StrikerRUS @shiyu1994 disagree and are fine with adding things like this, I won't block this PR.

I'll also raise an alternative idea here...what about something like the following?

* `Booster.predict()` = optimized for performance
* `Booster.predict_strict()` = same as `Booster.predict()`, but optimized for more input data checks. Raises informative errors and warnings when the data being scored is different from the training data.

jmoralez:
Thank you for your comments @jameslamb! I agree that the `Booster.predict` is a crucial part and that we would like the overhead to be a small as possible there. I like your suggestion and I think we could incorporate this as an option within the `Booster.predict` method, having an argument like `strict` that defaults to `False` and only run these checks if that argument was `True`.

Having said that I ran some benchmarks with the latest commit (e0827d0) that only changes the column order when necessary and I get 1.21 milliseconds in master and 1.23 milliseconds in this PR.
<details>
<summary>Benchmarking script</summary>

  ```python
import argparse
import timeit

import lightgbm as lgb
import numpy as np
import pandas as pd


parser = argparse.ArgumentParser()
parser.add_argument('--n_samples', type=int, default=10_000)
parser.add_argument('--n_features', type=int, default=100)
parser.add_argument('--repeats', type=int, default=20_000)
parser.add_argument('--predict_samples', type=int, default=1)
args = parser.parse_args()

X = np.random.rand(args.n_samples, args.n_features)
y = np.random.rand(args.n_samples)

df = pd.DataFrame(X, columns=[f'feature_{i}' for i in range(X.shape[1])])
df['y'] = y
ds = lgb.Dataset(df.drop(columns='y'), df['y'])
bst = lgb.train({'num_leaves': 31, 'verbosity': -1}, ds)
test_idxs = np.random.choice(range(X.shape[0]), args.predict_samples)
X_test = df.iloc[test_idxs, :-1]
print(f'Measuring time predicting {args.predict_samples} sample(s).')
print(f'{timeit.timeit("bst.predict(X_test)", number=args.repeats, globals=globals()) / args.repeats * 1_000:.2f} milliseconds')
  ```
</details>

Also I think users who want the minimum latency when computing predictions use numpy arrays, where the latency for this test was 0.22 milliseconds.

Happy to discuss this further, and merry Christmas @jameslamb!

StrikerRUS:
> * `Booster.predict_strict()` = same as `Booster.predict()`, but optimized for more input data checks.

>  I like your suggestion and I think we could incorporate this as an option within the `Booster.predict` method, having an argument like `strict` that defaults to `False` and only run these checks if that argument was `True`.

I'd better prefer to add new argument rather than new method.

XGBoost has `validate_features` parameter for such purporses:
https://github.com/dmlc/xgboost/pull/5191
https://github.com/dmlc/xgboost/pull/3653
https://github.com/dmlc/xgboost/pull/6605

shiyu1994:
I'd also prefer to add new parameters in `predict` method for feature name checks. Actually, we already have something like `predict_disable_shape_check`, see
https://lightgbm.readthedocs.io/en/latest/Parameters.html#predict_disable_shape_check

shiyu1994:
>Should I change that file to make the column names equal to the feature names?

Yes, I think we should do so.

jameslamb:
> I'd better prefer to add new argument rather than new method.

> XGBoost has validate_features parameter for such purposes:

Ok sure. Thanks for the links! I'm fine with adding such a parameter instead of a separate method. I think it should be specifically named `validate_features` to be consistent with XGBoost.

python-package/lightgbm/basic.py
```diff
@@ -512,12 +512,23 @@ def _get_bad_pandas_dtypes(dtypes):
     return bad_indices
 
 
-def _data_from_pandas(data, feature_name, categorical_feature, pandas_categorical):
+def _data_from_pandas(data, feature_name, categorical_feature, pandas_categorical, validate_features=False):
```
jameslamb, line 515:
```suggestion
def _data_from_pandas(data, feature_name, categorical_feature, pandas_categorical, validate_features):
```

In my experience, the use of default values in internal functions can lead to silent bugs of the form "higher-level things forgot to pass through a parameter passed in from the public API".

I think this default should be removed, forcing every call to `_data_from_pandas()` to explicitly choose a value.

jameslamb (changes requested):
I think these changes do address the request in #812 and the added tests cover that new behavior well. Nice work!

However, I have one more question. Still thinking about latency-sensitive users (https://github.com/microsoft/LightGBM/pull/4909#pullrequestreview-839729178), I feel that the default value for `validate_features` should be `False`.

As a user of a machine learning framework, I think I'd prefer to be told

> *"hey this method can now do some extra validation, change your code to opt into it"* 

instead of

> *"hey this method now does some extra validation by default, change your code to avoid the performance penalty of that validation"*

I feel like that specifically for `predict()` methods since those are necessary for serving models and therefore latency-sensitive. I expect other methods related to training, cross validation, or plotting (for example) to be much less latency-sensitive.

Before making changes, let's also hear what @StrikerRUS and @shiyu1994 hear about this topic.

StrikerRUS:
I have one general question before the review. Can this check be made at cpp side? Like reading feature names from file for `LGBM_BoosterPredictForFile` and optionally accepting feature names as new argument for all other `LGBM_BoosterPredict*` functions? Similarly to https://lightgbm.readthedocs.io/en/latest/Parameters.html#predict_disable_shape_check.

jmoralez:
> Can this check be made at cpp side?

I guess it's possible I'm just not sure how hard it would be, if it isn't too hard I can give it a shot. WDYT @shiyu1994?

jmoralez:
> Can this check be made at cpp side?

I gave this a shot and have an initial implementation. Let me know what you think.

StrikerRUS:
@jmoralez 
> I gave this a shot and have an initial implementation. 

Great, thanks for doing this!

Maybe next `4.0.0` release is a good time to not overcomplicate C API and add `names` argument (allowing it to be empty) directly to the existing `LGBM_BoosterPredictForMat()` function as a breaking change? WDYT?

shiyu1994:
> I gave this a shot and have an initial implementation. Let me know what you think.
@jmoralez Thank you! For this PR, let's keep the changes in the python side. And we can add the checks in C APIs in follow-up PRs so that other interfaces other than python can also benefits from this check.

StrikerRUS:
@shiyu1994 
> let's keep the changes in the python side. And we can add the checks in C APIs in follow-up PRs so that other interfaces other than python can also benefits from this check.

Do you mean to implement this feature at Python side in this PR and in a follow-up PR transfer it to cpp side? Actually it would mean remove Python code and added in this PR. Sorry it's confusing me, so I'd like to get the reason of such decision why not implement this feature at cpp side in the first round.

jmoralez:
@shiyu1994 WDYT? Should we focus only on the python package in this PR or should I restore the checks on the cpp side?

jmoralez:
Given that the logic for getting the feature names from the booster is already implemented on the Python and R sides it's easier to check them on each wrapper instead of defining the logic on the cpp side as well. So I can add the validation for the R side in this PR as well. I can also reduce the scope of the validation to just checking feature names and not trying to reorder them (that's probably too much and we should allow the user to do it instead) and disable the validation by default to not give unpleasant surprises to users. Please let me know what you think @jameslamb @StrikerRUS.

StrikerRUS:
>  I can also reduce the scope of the validation to just checking feature names and not trying to reorder them (that's probably too much and we should allow the user to do it instead)

Absolutely agree with this! We should check just all provided feature names for the exact equality with the stored ones.

> and disable the validation by default to not give unpleasant surprises to users.

I believe default behavior should be to raise an error if names used at training phase don't match names at inference. In most cases it should be an user-side error rather than intentional renaming. But it's definitely OK to have no names during inference while using some names during training and vice versa.

> it's easier to check them on each wrapper

But I still think that feature name check should be performed at cpp side to reduce the code duplication across wrappers and allow third-party packages to use new benefits as well. Especially given that the code was already written 91f2f8e55a173ed671a6172fee4466de73d3ab01, and that closely related [`predict_disable_shape_check`](https://lightgbm.readthedocs.io/en/latest/Parameters.html#predict_disable_shape_check) param is a global cpp-level param.

jameslamb:
> I can also reduce the scope of the validation to just checking feature names and not trying to reorder them (that's probably too much and we should allow the user to do it instead

+1, agree with this. I support keeping LightGBM stricter at its boundaries instead of trying to pull in code to fix issues in user input (similar to https://github.com/microsoft/LightGBM/issues/4483#issuecomment-909302041).

> disable the validation by default

I still feel as I did in https://github.com/microsoft/LightGBM/pull/4909#pullrequestreview-842363693, that this validation should be disabled by default, so that this change doesn't introduce new overhead to `predict()`.

> it's easier to check them on each wrapper

+1 to what @StrikerRUS said, I believe this is something that would be better handled on the C++ side, to avoid code duplication and so that all wrappers benefit from this behavior.

It's important to remember that when we refer to "wrappers", that doesn't only mean the R and Python packages...it also refers to Java interface via SWIG, SynapseML (LightGBM-on-Spark), and some of the projects in other languages mentioned at https://github.com/microsoft/LightGBM#external-unofficial-repositories.

jmoralez:
Thank you for your input. Do you have a suggestion on the interface? I previously created a new function but that seemed clunky and @StrikerRUS suggested: 

> Maybe next 4.0.0 release is a good time to not overcomplicate C API and add names argument (allowing it to be empty) directly to the existing LGBM_BoosterPredictForMat() function as a breaking change? WDYT?

But I also see the option of adding a new function like `LGBM_BoosterValidateFeatures` that can be called before the predict functions.

jmoralez:
I've added another proposal in bdce26b, please let me know what you think. I could do the same for the R package if you agree.

StrikerRUS:
I like new approach a lot! Thank you so much @jmoralez ! 🙂

jmoralez:
I believe this is ready for another review, I think we should leave the R-package part for a different PR.

python-package/lightgbm/basic.py
```diff
@@ -779,9 +779,6 @@ def predict(self, data, start_iteration=0, num_iteration=-1,
             Prediction result.
             Can be sparse or a list of sparse objects (each element represents predictions for one class) for feature contributions (when ``pred_contrib=True``).
         """
-        if isinstance(data, Dataset):
-            raise TypeError("Cannot use Dataset instance for prediction, please use raw data instead")
-        data = _data_from_pandas(data, None, None, self.pandas_categorical)[0]
```
StrikerRUS, line 784: Why was this moved from Predictor code? Now Predictor cannot accept pandas DataFrame which means, for example, `refit()` method cannot accept DataFrames anymore:
https://github.com/microsoft/LightGBM/blob/ab9236baa06c3f071acf6b45a771c2a09661f146/python-package/lightgbm/basic.py#L3615

and DataFrames cannot be used together with `init_model` argument during constructing Dataset:
https://github.com/microsoft/LightGBM/blob/ab9236baa06c3f071acf6b45a771c2a09661f146/python-package/lightgbm/basic.py#L1402-L1404

src/c_api.cpp
```diff
@@ -2129,6 +2129,31 @@ int LGBM_BoosterPredictForCSC(BoosterHandle handle,
...
 
+int LGBM_BoosterValidateFeatureNames(BoosterHandle handle,
+                                     const char** data_names,
+                                     int data_num_features) {
+  API_BEGIN();
+  int booster_num_features;
+  size_t out_buffer_len;
+  LGBM_BoosterGetFeatureNames(handle, 0, &booster_num_features, 0, &out_buffer_len, nullptr);
+  if (booster_num_features != data_num_features) {
+    Log::Fatal("Booster was trained on %d features, but got %d input features to predict.", booster_num_features, data_num_features);
```
StrikerRUS, line 2140: Maybe use more neutral word "model" instead of "booster" to not confuse users that might use high level abstractions built on Booster like sklearn estimators?
```suggestion
    Log::Fatal("Model was trained on %d features, but got %d input features to predict.", booster_num_features, data_num_features);
```

StrikerRUS (changes requested):
Thank you very much for rethinking the interface. Love it.

LGTM for the approach! However, I left some comments below.

python-package/lightgbm/basic.py
jameslamb: I agree with this comment, and will just add that it would be useful to have unit tests (in a separate PR) for them, so such regressions could be caught automatically in the future.

jmoralez: Addressed in 774a715

StrikerRUS: Thanks!

Should we add `validate_features` argument to those methods as well?

StrikerRUS (approved):
Thank you very much for adding this feature!

StrikerRUS:
@guolinke Could you please help with review of cpp part?

src/c_api.cpp
```diff
@@ -2129,6 +2129,31 @@ int LGBM_BoosterPredictForCSC(BoosterHandle handle,
...
+  size_t out_buffer_len;
+  LGBM_BoosterGetFeatureNames(handle, 0, &booster_num_features, 0, &out_buffer_len, nullptr);
+  if (booster_num_features != data_num_features) {
+    Log::Fatal("Model was trained on %d features, but got %d input features to predict.", booster_num_features, data_num_features);
+  }
+  std::vector<std::vector<char>> tmp_names(booster_num_features);
+  std::vector<char*> booster_names(booster_num_features);
+  for (int i = 0; i < booster_num_features; ++i) {
+    tmp_names[i].resize(out_buffer_len);
+    booster_names[i] = tmp_names[i].data();
```
guolinke, line 2146: you can use https://github.com/microsoft/LightGBM/blob/560c884dd153ac6a95429df8c749aae9632b8215/include/LightGBM/utils/common.h#L634-L642 this to simplify the code.

src/c_api.cpp
```diff
@@ -2129,6 +2129,31 @@ int LGBM_BoosterPredictForCSC(BoosterHandle handle,
...
+                                     const char** data_names,
+                                     int data_num_features) {
+  API_BEGIN();
+  int booster_num_features;
+  size_t out_buffer_len;
+  LGBM_BoosterGetFeatureNames(handle, 0, &booster_num_features, 0, &out_buffer_len, nullptr);
+  if (booster_num_features != data_num_features) {
+    Log::Fatal("Model was trained on %d features, but got %d input features to predict.", booster_num_features, data_num_features);
+  }
+  std::vector<std::vector<char>> tmp_names(booster_num_features);
```
guolinke, line 2142:
```suggestion
  std::vector<std::vector<char>> tmp_names(booster_num_features, std::vector<char>(out_buffer_len));
```

src/c_api.cpp
```diff
@@ -2129,6 +2129,31 @@ int LGBM_BoosterPredictForCSC(BoosterHandle handle,
...
+  std::vector<char*> booster_names(booster_num_features);
+  for (int i = 0; i < booster_num_features; ++i) {
```
guolinke, line 2144: this loop is not necessary.

src/c_api.cpp
jmoralez: Thanks! I've changed it in 3ff99ff.

guolinke (approved):
Thank you so much!

StrikerRUS:
@jameslamb Your previous requesting changes review doesn't allow to merge this PR. Please re-review.

jameslamb (approved):
Very nice work on this @jmoralez , thanks very much!

@StrikerRUS , thanks for the `@`. Apologies for holding this up due to my lack of review.

PR #4913, opened 2021-12-28
[R-package] respect aliases for objective and metric and lgb.train() and lgb.cv()

jameslamb:
Similar to #4903, this PR proposes changes to ensure that parameter aliases in `lgb.train()` and `lgb.cv()` are handled correctly in the R package.

Currently, internal helper functions access `params$objective` and `params$metric` directly, without considering aliases.

https://github.com/microsoft/LightGBM/blob/ce486e5b45a6f5e67743e14765ed139ff8d532e5/R-package/R/utils.R#L168

https://github.com/microsoft/LightGBM/blob/ce486e5b45a6f5e67743e14765ed139ff8d532e5/R-package/R/utils.R#L194

As a result, it's not possible today to use any of [the aliases for `objective`](https://lightgbm.readthedocs.io/en/latest/Parameters.html#objective) with `lgb.train()` and `lgb.cv()`.

```r
library(lightgbm)
data("agaricus.train", package = "lightgbm")
train <- agaricus.train
dtrain <- lgb.Dataset(
    data = train$data
    , label = train$label
)
bst <- lgb.train(
    data = dtrain
    , params = list(
        num_leaves = 5L
        , application = "binary"
        , num_iterations = 5L
    )
)
```

>  Error in lgb.check.obj(params = params, obj = obj) : 
  lgb.check.obj: objective should be a character or a function

This PR proposes the following changes to address that:

* populate `params$objective` and `params$metric` using internal helper function `lgb.check.wrapper_param()`, which handles reconciling parameters passed by main parameter name and aliases
* move code for handling the possibility of custom objective and metrics (which are R function objects, not strings) down so it happens _after_ aliases have been handled in `params`

### DIFF
--- a/R-package/tests/testthat/test_basic.R
+++ b/R-package/tests/testthat/test_basic.R
@@ -127,6 +127,7 @@ test_that("train and predict softmax", {
   expect_equal(length(pred), nrow(iris) * 3L)
 })
 
+
 test_that("use of multiple eval metrics works", {
   metrics <- list("binary_error", "auc", "binary_logloss")
   bst <- lightgbm(
--- a/R-package/tests/testthat/test_basic.R
+++ b/R-package/tests/testthat/test_basic.R
@@ -127,7 +127,6 @@ test_that("train and predict softmax", {
   expect_equal(length(pred), nrow(iris) * 3L)
 })
 
-
 test_that("use of multiple eval metrics works", {
   metrics <- list("binary_error", "auc", "binary_logloss")
   bst <- lightgbm(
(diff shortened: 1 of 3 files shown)


StrikerRUS:
Could you please resolve a conflict with `master`?

R-package/tests/testthat/test_basic.R
```diff
@@ -616,6 +662,52 @@ test_that("lgb.train() works as expected with multiple eval metrics", {
...
+    )
+    , valids = list(
+      "train" = dtrain
+    )
+  )
+  record_results <- bst$record_evals[["train"]]
+  expect_equal(sort(names(record_results)), c("auc", "binary_logloss"))
+  expect_length(record_results[["auc"]][["eval"]], nrounds)
+  expect_length(record_results[["binary_logloss"]][["eval"]], nrounds)
+  expect_equal(bst$params[["metric"]], list("auc", "binary_logloss"))
```
StrikerRUS, line 708: I like this assertion! Can we add the same assertion for objective into `test_that("lgb.train() respects parameter aliases for objective",` test?

StrikerRUS (approved):
LGTM, thank you!

R-package/tests/testthat/test_basic.R
jameslamb: ah yep, definitely! good point that `params` should be checked for that test too.

R-package/tests/testthat/test_basic.R
```diff
@@ -616,6 +662,52 @@ test_that("lgb.train() works as expected with multiple eval metrics", {
...
+      num_leaves = 5L
+      , application = "binary"
+      , num_iterations = nrounds
+    )
+    , valids = list(
+      "the_training_data" = dtrain
+    )
+  )
+  expect_named(bst$record_evals[["the_training_data"]], "binary_logloss")
+  expect_length(bst$record_evals[["the_training_data"]][["binary_logloss"]][["eval"]], nrounds)
```
jameslamb, line 683:
```suggestion
  expect_length(bst$record_evals[["the_training_data"]][["binary_logloss"]][["eval"]], nrounds)
  expect_equal(bst$params[["objective"]], "binary")
```

R-package/tests/testthat/test_basic.R
jameslamb: added in https://github.com/microsoft/LightGBM/pull/4913/commits/f90b2d68981bd875094127554734d1514fc708f7

PR #4914, opened 2021-12-28
[R-package] fix handling of duplicate parameters (fixes #4521)

### LINKED ISSUE #4521: [R-package] Possibly-incorrect handling of duplicate parameters
## Description

In the CLI, if you provide a training config with `bagging_fraction=0.8` and `bagging_fraction=0.5`, you will see a warning like this during training.

> bagging_fraction is set=0.8, bagging_fraction=0.5 will be ignored. Current value: bagging_fraction=0.8

If you provide a config with those duplicate values to `lgb.train()` in the R package, you'll see this instead:

> bagging_fraction is set=0.8, bagging_fraction=0.8 will be ignored. Current value: bagging_fraction=0.8

I need to do more investigation, but I think this means that the R package is not handling duplicates in parameters correctly.

## Reproducible example

This behavior cannot be reproduced in the Python package because Python code in this project uses dictionaries for parameters at all user-facing entrypoints and Python dictionaries enforce the uniqueness of keys.

However, the R package uses R lists, which can have duplicate keys (e.g. `list(a = 5, a = 6)`).

```r
library(lightgbm)

train <- agaricus.train

model <- lgb.train(
  params = list(
      objective = "regression",
      bagging_fraction = 0.8
  )
  , data = dtrain
  , nrounds = 5L
  , bagging_fraction = 0.5
)
```

> [LightGBM] [Warning] bagging_fraction is set=0.8, bagging_fraction=0.8 will be ignored. Current value: bagging_fraction=0.8

With the CLI, the same duplicated parameters results in the expected warning message.

```shell
cd ./examples/regression
echo "bagging_fraction = 0.5" >> train.conf
echo "" >> train.conf

"../../lightgbm" config=train.conf
```

> [LightGBM] [Warning] bagging_fraction is set=0.8, bagging_fraction=0.5 will be ignored. Current value: bagging_fraction=0.8

## Environment info

LightGBM version or commit hash: latest `master` (https://github.com/microsoft/LightGBM/commit/86ead2050c006b984e88396e0e380304bac114d1)

```shell
# R package
sh build-cran-package.sh

# CLI
mkdir build
cd build
cmake ..
make -j4
```

## Additional Comments

This is probably related to the R package's use of `append()` to handle parameters passed through `...` (R's equivalent of `**kwargs` in Python). Combining two lists in R with `append()` doesn't merge keys.

```r
append(list(a = 6), list(b = 7, a = 9))

$a
[1] 6

$b
[1] 7

$a
[1] 9
```

It's possible that changing these uses to `modifyList()` would help with this issue.

In Python package, in the scikit-learn interface, parameters passed through `**kwargs` take precedence over those passed in the `params` dictionary.

https://github.com/microsoft/LightGBM/blob/86ead2050c006b984e88396e0e380304bac114d1/python-package/lightgbm/sklearn.py#L517

https://github.com/microsoft/LightGBM/blob/86ead2050c006b984e88396e0e380304bac114d1/python-package/lightgbm/sklearn.py#L562-L566

However, since the current behavior of the R package is for parameters in `params` to take precedence over those from `...`, I think that that behavior should be preserved in the 3.3.0 release (#4310). All uses of `...` are going to be removed in LightGBM 4.0.0 (#4226).

jameslamb:
Fixes #4521.

When LightGBM's core library finds duplicates provided in params, it emits a warning-level log message and uses the value from the first instance of that parameter encountered while parsing parameters.

The R package doesn't enforce uniqueness of parameter names either, but due to a bug in how it converts an R list of parameters to a string, the warning message generated when duplicate parameters are passed is currently incorrect.

This PR proposes fixing that.

### DIFF
--- a/R-package/R/utils.R
+++ b/R-package/R/utils.R
@@ -25,6 +25,7 @@ lgb.params2str <- function(params) {
     stop("params must be a list")
   }
 
+  names(params) <- gsub("\\.", "_", names(params))
   param_names <- names(params)
   ret <- list()
 
--- a/R-package/R/utils.R
+++ b/R-package/R/utils.R
@@ -25,20 +25,18 @@ lgb.params2str <- function(params) {
     stop("params must be a list")
   }
 
-  # Split parameter names
-  names(params) <- gsub("\\.", "_", names(params))
-
+  param_names <- names(params)
   ret <- list()
 
   # Perform key value join
-  for (key in names(params)) {
+  for (i in seq_along(params)) {
 
     # If a parameter has multiple values, join those values together with commas.
     # trimws() is necessary because format() will pad to make strings the same width
     val <- paste0(
       trimws(
         format(
-          x = params[[key]]
+          x = unname(params[[i]])
           , scientific = FALSE
         )
       )
@@ -47,7 +45,7 @@ lgb.params2str <- function(params) {
     if (nchar(val) <= 0L) next # Skip join
 
     # Join key value
-    pair <- paste0(c(key, val), collapse = "=")
+    pair <- paste0(c(param_names[[i]], val), collapse = "=")
     ret <- c(ret, pair)
 
   }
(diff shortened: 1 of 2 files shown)


StrikerRUS (approved):
Thank you for fixing this!

PR #4915, opened 2021-12-28
[R-package] [tests] remove uses of testthat::context()

jameslamb:
Older versions of `{testthat}` required the use of a function, `testthat::context()`, to separate tests in reporting output.

`{testthat}` has been moving away from requiring that, instead preferring to group tests results by test file name, the way `pytest` does.

From the testthat docs ([link](https://testthat.r-lib.org/reference/context.html))

> Use of `context()` is no longer recommended. Instead omit it, and messages will use the name of the file instead. This ensures that the context and test file name are always in sync.

And from the vignette describing `{testthat}` 3.x ([link](https://github.com/r-lib/testthat/blob/f8ccf033a74d89e2d1d3cb2ec3c6c4f63ca69055/vignettes/third-edition.Rmd#L75-L76)), which was first released in October 2020 ([link](https://cran.r-project.org/src/contrib/Archive/testthat/))

> `context()` is formally deprecated. testthat has been moving away from `context()` in favour of file names for quite some time, and now you'll be strongly encouraged remove these calls from your tests.

This PR proposes removing uses of `context()` in the R package's tests.

### Notes for Reviewers

Contributes to #4862. Check the logs in the `r-package MSVC` tests to see what the new test logs look like (since those tests don't run `R CMD check`).

### DIFF
--- a/R-package/tests/testthat/test_Predictor.R
+++ b/R-package/tests/testthat/test_Predictor.R
@@ -2,8 +2,6 @@ VERBOSITY <- as.integer(
   Sys.getenv("LIGHTGBM_TEST_VERBOSITY", "-1")
 )
 
-context("Predictor")
-
 test_that("Predictor$finalize() should not fail", {
     X <- as.matrix(as.integer(iris[, "Species"]), ncol = 1L)
     y <- iris[["Sepal.Length"]]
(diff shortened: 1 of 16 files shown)


StrikerRUS (approved):
Thanks for the cleanup!

StrikerRUS:
Please resolve conflicts with `master`.

PR #4916, opened 2021-12-28
[R-package][tests] remove uses of `testthat::expect_is()`

jameslamb:
Similar to #4915, this PR proposes removing `{lightgbm}`'s use of a `{testthat}` function that is currently deprecated.

From the vignette describing `{testthat}` 3.x ([link](https://github.com/r-lib/testthat/blob/f8ccf033a74d89e2d1d3cb2ec3c6c4f63ca69055/vignettes/third-edition.Rmd#L78-L79)), which was first released in October 2020 ([link](https://cran.r-project.org/src/contrib/Archive/testthat/))

> `expect_is()` is deprecated in favour of the more specific `expect_type()`, `expect_s3_class()`, and `expect_s4_class()`.
    This ensures that you check the expected class along with the expected OO system.

This PR proposes removing uses of `expect_is()` in the R package's tests. It proposes replacing them with `expect_true(methods::is())` instead of the newest `{testthat}` because those 3 methods can't be used to test the class of an `R6` object.

### DIFF
--- a/R-package/tests/testthat/test_basic.R
+++ b/R-package/tests/testthat/test_basic.R
@@ -394,7 +394,7 @@ test_that("CVBooster$reset_parameter() works as expected", {
     , nrounds = 3L
     , nfold = n_folds
   )
-  expect_is(cv_bst, "lgb.CVBooster")
+  expect_true(methods::is(cv_bst, "lgb.CVBooster"))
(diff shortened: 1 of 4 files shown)


StrikerRUS (approved):
Thanks for the updates!

Issue #3381, opened 2020-09-11
custom the objective function for LTR


chloe-wang:
Hi there, 

I noticed that there is a sample for custom objective:
https://github.com/Microsoft/LightGBM/blob/master/R-package/demo/multiclass_custom_objective.R

However,  how shall we define a custom objective when the objective includes some other information other than preds and dtrain? For example, in the lambdarank case, we may have the query_id information as well.

chloe-wang:
I figure it out.

Issue #2239, opened 2019-06-24
performance of lambdarank using custom objective is poor when compared with in-built lambdarank


y-research-yu:
Based on the common used dataset, say MQ2008, the 5-fold CV performance of in-built lambdarank is:

nDCG@1:0.4319, nDCG@3:0.4574, nDCG@5:0.4953, nDCG@10:0.5844

By setting the hessian with constant values, say 1.0, the 5-fold CV performance of manually plug-in lambdarank via the fobj parameter is:

nDCG@1:0.3483, nDCG@3:0.4021, nDCG@5:0.4429, nDCG@10:0.5394

If using the manually computed hessian, the following issue is observed, namely the training can not conducted ...

[10]	valid_0's ndcg@1: 0.113095	valid_0's ndcg@2: 0.161686	valid_0's ndcg@3: 0.202154	valid_0's ndcg@4: 0.214604	valid_0's ndcg@5: 0.23572
[20]	valid_0's ndcg@1: 0.113095	valid_0's ndcg@2: 0.161686	valid_0's ndcg@3: 0.202154	valid_0's ndcg@4: 0.214604	valid_0's ndcg@5: 0.23572
[30]	valid_0's ndcg@1: 0.113095	valid_0's ndcg@2: 0.161686	valid_0's ndcg@3: 0.202154	valid_0's ndcg@4: 0.214604	valid_0's ndcg@5: 0.23572

## Environment info

Operating System: ubuntu 16.04

Python version: 3.7

LightGBM version or commit hash: 2.2.3

**Any comments on the above results and the following questions are highly appreciated**.
(1) why there is an obvious difference between in-built lambdarank and manual plug-in  lambdarank? Is it due to the inner parameter setting?
(2) what are the tips when using manually computed hessian, which seems to be quite sensitive and vulnerable?

guolinke:
I think it is related to your custom obj, could you provide it?
And what is "manually computed hessian", any differences with built-in hessian?

y-research-yu:
@guolinke so many thanks for your quick response.
I just created a gist snippet https://gist.github.com/y-research-yu/ae9d32dabce52d33073278c2a13dabac

I made the comparison by commenting fobj or not to test the custom objective, e.g., 
params = {'boosting_type': 'gbdt', # gbdt, dart
                  'objective': 'lambdarank',
                  'metric': 'ndcg',
[...]

Issue #1896, opened 2018-12-06
custom objective for lambdarank/ranking


xin-jin:
Is custom objective function supported for ranking models? I would like to tweak the lambdarank loss a little bit. Since the loss function needs to know the group information, what would the loss function signature be?

Thanks!

guolinke:
refer to https://github.com/Microsoft/LightGBM/blob/3ad9cba0321a8ac2e5f625da6dcc6f7c1516a750/examples/python-guide/advanced_example.py#L140-L145 
you can use the `train_data.get_group()`

1a1a11a:
is the function signature `def loglikelihood(preds, train_data)` or `def loglikelihood(y_true, pred)`? Because here https://towardsdatascience.com/custom-loss-functions-for-gradient-boosting-f79c1b40466d says the later and when I print the two input types are `<class 'numpy.ndarray'> <class 'numpy.ndarray'>`.

1a1a11a:
Maybe the function for different interfaces (sklearn and LightGBM) are different?

guolinke:
@1a1a11a yeah, they are different.

siamakz:
Hi @guolinke 
I am also trying to define a custom objective for ranking problem. Can you please elaborate more on how it can be done? I am just starting to work with LightGBM. Thanks!

PR #4920, opened 2021-12-30
[dask] add support for custom objective functions (fixes #3934)

### LINKED ISSUE #3934: [dask] Support custom objective functions
## Summary

The Dask estimators in `lightgbm.dask` should support the use of a custom objective function.

## Motivation

This feature would bring Dask estimators closer to parity with the sklearn estimators.

## Description

I haven't thought through this much yet, just writing up a placeholder issue for discussion. If you're reading this and have ideas, please comment and the issue can be re-opened.

## References

See https://github.com/microsoft/LightGBM/blob/15916a950476e2971aee9d9a0bd30ed0f43b9b56/python-package/lightgbm/sklearn.py#L430-L453 for an explanation of how this works in the sklearn estimators.

Can look at how `xgboost.dask` handles this for inspiration:

* https://github.com/dmlc/xgboost/blob/9b267a435e0aa1dfdff211d1a5b668538c51cd35/python-package/xgboost/dask.py#L1552-L1555
* https://github.com/dmlc/xgboost/blob/9b267a435e0aa1dfdff211d1a5b668538c51cd35/python-package/xgboost/dask.py#L849

jameslamb:
Fixes #3934.

This PR proposes changes to add support for the use of custom objective functions with `lightgbm.dask`. Nothing functionally needed to change in the package to support this, but this PR makes the following changes:

* adds unit tests on the use of custom objective functions for main types of tasks
* adds custom objective function note to the Dask API docs
* adds a section in the "Distributed Learning Guide" showing an example of how to use a custom objective function in `lightgbm.dask`
* updates the type hints in the Dask scikit-learn model objects (contributing to #3756)

### Where did the objective functions added in the tests come from?

**Binary classification (logistic regression)**

https://github.com/microsoft/LightGBM/blob/af5b40e1f653e25a5f3470e09de4d1c68dc59663/tests/python_package_test/test_sklearn.py#L57

**Multiclass classification (multi-logloss)**

Converted this code from the R demos to Python. I tested it manually against the R package to confirm that the two functions produce the same results.

https://github.com/microsoft/LightGBM/blob/af5b40e1f653e25a5f3470e09de4d1c68dc59663/R-package/demo/multiclass_custom_objective.R#L45

**Ranking and Regression (regression_l2)**

https://github.com/microsoft/LightGBM/blob/af5b40e1f653e25a5f3470e09de4d1c68dc59663/tests/python_package_test/test_sklearn.py#L51

I couldn't find an example of a custom objective function for the learning task. This is something that we've received feature requests about in the past (e.g. #1896,  #2239, #3381) but none of those discussions produced Python or R code that could be used as a custom objective function for learning-to-rank with LightGBM.

I believe it's fine, for the purposes of these tests, to use the L2 objective for the ranking task. I found that it produced good-enough results (measured by spearman correlation of the predicted scores with the actual labels) that improved as I increased `num_iterations`, which I took as a good sign.

### Why not just add a custom objective function test to `test_{task}`?

I considered adding something like the following to `test_classifier`, `test_regressor`, and `test_ranker` in the Dask tests:

```python
@pytest.mark.parametrize('objective', ["regression_l2", _objective_least_squares])
```

That would ensure that custom objective functions are tested for every combination of data type, boosting type, and distributed training strategy.

However, I decided against this because there are already so many Dask tests, and I didn't want to add too much incremental burden to this project's CI. I believe the tests added in this PR provide good enough confidence that the Dask estimators are passing through custom objective functions to the `lightgbm.sklearn` equivalents.


jameslamb:
One of the new tests, `test_classifier_custom_objective[multiclass-classification-dataframe-with-categorical]`, failed across some (but not all) of the Python jobs.

> AssertionError: found values in 'a' and 'b' which differ by more than the allowed amount

* `cuda 10.0` ([link](https://github.com/microsoft/LightGBM/runs/4664985085?check_suite_focus=true))
* `cuda 11.4.2` ([link](https://github.com/microsoft/LightGBM/runs/4664985070?check_suite_focus=true))
* `Linux_latest regular` ([link](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=11886&view=logs&j=275189f9-c769-596a-7ef9-49fb48a9ab70&t=3a9e7a4a-04e6-52a0-67ea-6e8f6cfda74f))
* `Linux bdist` ([link](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=11886&view=logs&j=fb919173-48b9-522d-5342-9a59a13eb10b&t=7fcbeee5-d4c1-571c-8072-1d5874591254))
* `Linux sdist` ([link](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=11886&view=logs&j=02a2c3ba-81f8-54e3-0767-5d5adbb0daa9&t=720ee3fa-96d4-5b47-dbf4-01607b74ade2))
* `Linux regular` ([link](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=11886&view=logs&j=c28dceab-947a-5848-c21f-eef3695e5f11&t=fa158246-17e2-53d4-5936-86070edbaacf))

Pushed https://github.com/microsoft/LightGBM/pull/4920/commits/8b784190e7569be242f516edd9d3a71e480e5f44 to increase `num_iterations` and `num_leaves`, matching the values used in the existing `test_classifier` test.

https://github.com/microsoft/LightGBM/blob/af5b40e1f653e25a5f3470e09de4d1c68dc59663/tests/python_package_test/test_dask.py#L279-L280

jameslamb:
@ffineis if you have time / interest I'd welcome your thoughts on this PR too 👀

tests/python_package_test/test_dask.py
```diff
@@ -262,6 +262,37 @@ def _unpickle(filepath, serializer):
...
+    grad = y_pred - y_true
+    hess = y_pred * (1.0 - y_pred)
+    return grad, hess
+
+
+def _objective_logloss(y_true, y_pred):
+    num_rows = len(y_true)
+    num_class = len(np.unique(y_true))
+    # operate on preds as [num_data, num_classes] matrix
+    y_pred = y_pred.T.reshape(-1, num_class)
```
jmoralez, line 282: The `y_pred` array is 1D and comes in the form [probs_class1 | probs_class2 | probs_class3] so we have to reshape it with `order='F'`.
```suggestion
    y_pred = y_pred.reshape(-1, num_class, order='F')
```

tests/python_package_test/test_dask.py
```diff
@@ -262,6 +262,37 @@ def _unpickle(filepath, serializer):
...
+    row_wise_max = np.max(y_pred, axis=1).reshape(num_rows, 1)
+    preds = y_pred - row_wise_max
+    prob = np.exp(preds) / np.sum(np.exp(preds), axis=1).reshape(num_rows, 1)
+    grad_update = np.zeros_like(preds)
+    grad_update[np.arange(num_rows), y_true.astype('int')] = -1.0
+    grad = prob + grad_update
```
jmoralez, line 288: I believe this should be something like:
```suggestion
    ohe_classes = np.zeros_like(prob)
    ohe_classes[np.arange(num_rows), y_true.astype(int)] = 1.0
    grad = ohe_classes * (prob - 1) + (1 - ohe_classes) * prob
```

tests/python_package_test/test_dask.py
```diff
@@ -262,6 +262,37 @@ def _unpickle(filepath, serializer):
...
+    hess = 2.0 * prob * (1.0 - prob)
```
jmoralez, line 289: Per #3256 the hessian gets multiplied by the factor `num_class / (num_class - 1)`
```suggestion
    factor = num_class / (num_class - 1)
    hess = factor * prob * (1 - prob)
```

jmoralez (changes requested):

jmoralez:
@jameslamb I left some suggestions on the multi-class objective function to try to get about the same score as the built-in one. Was that your intention?

tests/python_package_test/test_dask.py
jameslamb: ah!! you're absolutely right! thank you.

```python
import numpy as np

x = np.array([
    0.1, 0.1, 0.2, 0.2, 0.3, 0.3
])

x.T.reshape(-1, 3)

# array([
#    [0.1, 0.1, 0.2],
#    [0.2, 0.3, 0.3]
# ])

x.reshape(-1, 3, order='F')

# array([
#    [0.1, 0.2, 0.3],
#    [0.1, 0.2, 0.3]
# ])
```

tests/python_package_test/test_dask.py
jameslamb: ah ok yeah, thanks for catching that!

tests/python_package_test/test_dask.py
jameslamb: hmmm are you sure? It seems to me that the existing current logic is correct, based on

https://github.com/microsoft/LightGBM/blob/af5b40e1f653e25a5f3470e09de4d1c68dc59663/src/objective/multiclass_objective.hpp#L100-L103

Looking more closely, it also seems like your proposed change and the existing code in this PR are identical. Aren't both saying "the gradient is `prob` for classes not equal to the label and otherwise `prob - 1`"?

jmoralez: You're right, that's my bad haha.

jameslamb:
> I left some suggestions on the multi-class objective function to try to get about the same score as the built-in one. Was that your intention?

Thanks very much @jmoralez ! I think you caught some critical mistakes in my implementation. My goal isn't to perfectly replicate the built-in one, just to have something good enough to use in unit tests to build confidence that the Dask estimators are handling custom objective functions correctly.

tests/python_package_test/test_dask.py
jameslamb: ha no problem!

jmoralez (approved):

tests/python_package_test/test_dask.py
```diff
@@ -455,6 +487,74 @@ def test_classifier_pred_contrib(output, task, cluster):
...
+            time_out=5,
+            tree_learner='data',
+            **params
+        )
+        dask_classifier = dask_classifier.fit(dX, dy, sample_weight=dw)
+        dask_classifier_local = dask_classifier.to_local()
+        p1_proba = dask_classifier.predict_proba(dX).compute()
+        p1_proba_local = dask_classifier_local.predict_proba(X)
+
+        # with a custom objective, predictiion result is a raw score instead of predicted class
```
StrikerRUS, line 528:
```suggestion
        # with a custom objective, prediction result is a raw score instead of predicted class
```

tests/python_package_test/test_dask.py
```diff
@@ -455,6 +487,74 @@ def test_classifier_pred_contrib(output, task, cluster):
...
+        p2_class = p2_class.astype('int64')
+
+        if task == 'multiclass-classification':
+            p1_class = p1_class.argmax(axis=1)
+            p1_class_local = p1_class_local.argmax(axis=1)
+            p2_class = p2_class.argmax(axis=1)
+
+        # function should have been preserved
+        assert callable(dask_classifier.objective)
+        assert callable(dask_classifier_local.objective)
```
StrikerRUS, line 547: I think it's more idiomatic to check post-fit attribute:
https://lightgbm.readthedocs.io/en/latest/pythonapi/lightgbm.LGBMModel.html#lightgbm.LGBMModel.objective_
```suggestion
        assert callable(dask_classifier.objective_)
        assert callable(dask_classifier_local.objective_)
```

tests/python_package_test/test_dask.py
```diff
@@ -700,6 +800,55 @@ def test_regressor_quantile(output, alpha, cluster):
...
+        p1 = p1.compute()
+
+        local_regressor = lgb.LGBMRegressor(**params)
+        local_regressor.fit(X, y, sample_weight=w)
+        p2 = local_regressor.predict(X)
+        s2 = local_regressor.score(X, y)
+
+        # function should have been preserved
+        assert callable(dask_regressor.objective)
+        assert callable(dask_regressor_local.objective)
```
StrikerRUS, line 838:
```suggestion
        assert callable(dask_regressor.objective_)
        assert callable(dask_regressor_local.objective_)
```

tests/python_package_test/test_dask.py
```diff
@@ -808,6 +957,68 @@ def test_ranker(output, group, boosting_type, tree_learner, cluster):
...
+        # distributed ranker should be able to rank decently well with the least-squares objective
+        # and should have high rank correlation with scores from serial ranker.
+        dcor = spearmanr(rnkvec_dask, y).correlation
+        assert dcor > 0.6
+        assert spearmanr(rnkvec_dask, rnkvec_local).correlation > 0.8
+        assert_eq(rnkvec_dask, rnkvec_dask_local)
+
+        # function should have been preserved
+        assert callable(dask_ranker.objective)
+        assert callable(dask_ranker_local.objective)
```
StrikerRUS, line 1019:
```suggestion
        assert callable(dask_ranker.objective_)
        assert callable(dask_ranker_local.objective_)
```

tests/python_package_test/test_dask.py
```diff
@@ -262,6 +262,38 @@ def _unpickle(filepath, serializer):
...
+def _objective_logloss(y_true, y_pred):
+    num_rows = len(y_true)
+    num_class = len(np.unique(y_true))
+    # operate on preds as [num_data, num_classes] matrix
+    y_pred = y_pred.reshape(-1, num_class, order='F')
+    row_wise_max = np.max(y_pred, axis=1).reshape(num_rows, 1)
+    preds = y_pred - row_wise_max
+    prob = np.exp(preds) / np.sum(np.exp(preds), axis=1).reshape(num_rows, 1)
+    grad_update = np.zeros_like(preds)
+    grad_update[np.arange(num_rows), y_true.astype('int')] = -1.0
```
StrikerRUS, line 287: Can we please use numpy objects for referring data types instead of strings?
```suggestion
    grad_update[np.arange(num_rows), y_true.astype(np.int32)] = -1.0
```

tests/python_package_test/test_dask.py
```diff
@@ -455,6 +487,74 @@ def test_classifier_pred_contrib(output, task, cluster):
...
+        p1_class = (1.0 / (1.0 + np.exp(-p1_proba))) > 0.5
+        p1_class = p1_class.astype('int64')
+        p1_class_local = (1.0 / (1.0 + np.exp(-p1_proba_local))) > 0.5
+        p1_class_local = p1_class_local.astype('int64')
+
+        local_classifier = lgb.LGBMClassifier(**params)
+        local_classifier.fit(X, y, sample_weight=w)
+        p2_proba = local_classifier.predict_proba(X)
+        p2_class = (1.0 / (1.0 + np.exp(-p1_proba))) > 0.5
+        p2_class = p2_class.astype('int64')
```
StrikerRUS, line 538:
```suggestion
        p1_class = p1_class.astype(np.int64)
        p1_class_local = (1.0 / (1.0 + np.exp(-p1_proba_local))) > 0.5
        p1_class_local = p1_class_local.astype(np.int64)

        local_classifier = lgb.LGBMClassifier(**params)
        local_classifier.fit(X, y, sample_weight=w)
        p2_proba = local_classifier.predict_proba(X)
        p2_class = (1.0 / (1.0 + np.exp(-p1_proba))) > 0.5
        p2_class = p2_class.astype(np.int64)
```

StrikerRUS (commented):
Thanks for working on this! I have a few number of minor suggestions below:

StrikerRUS:
> One of the new tests, `test_classifier_custom_objective[multiclass-classification-dataframe-with-categorical]`, failed across some (but not all) of the Python jobs.

It has failed again in `Linux regular` CI job:
```
            # probability estimates should be similar
>           assert_eq(p1_proba, p2_proba, atol=0.03)
...
>           assert allclose(a, b, **kwargs), msg
E           AssertionError: found values in 'a' and 'b' which differ by more than the allowed amount
...
=========================== short test summary info ============================
FAILED ../tests/python_package_test/test_dask.py::test_classifier_custom_objective[multiclass-classification-dataframe-with-categorical]
= 1 failed, 649 passed, 12 skipped, 2 xfailed, 398 warnings in 407.40s (0:06:47) =

```

jameslamb:
> It has failed again in Linux regular CI job

I think this multiclass objective I've added has numeric stability problems. I'll experiment a bit.

jmoralez:
In #4925 I'm testing that objective function against the built-in and seems to be working ok. Maybe increasing the learning rate to 0.1 could mimic the results from `test_classifier` and achieve that atol of 0.03

StrikerRUS:
I noticed that Linux Azure Pipelines CI jobs started to timeout randomly two or three days ago.

jameslamb:
> I noticed that Linux Azure Pipelines CI jobs started to timeout randomly two or three days ago.

Thanks for letting me know! Looks like that is what happened here as well.

By the way... @jmoralez @StrikerRUS I pushed changes to this PR last night and I think it's ready for another review. See the diff in https://github.com/microsoft/LightGBM/pull/4920/commits/54df09e734cbc98f261bf4a108e781e41d591914.

I found the following issues that were causing the multiclass classification tests to fail:

1. `.predict_proba()` doesn't produce a probability when using a custom objective...it's identical to `.predict(raw_score=True)`. That made the test code confusing, and I was accidentally comparing raw scores with code like `assert_eq(p1_proba, p2_proba)`.
2. I forgot to change the logic for going from probabilities to classes to be different for binary classification vs. multi-class classification.
3. multiclass classification with categorical features was very sensitive to small differences in Dataset construction, probably as a result of the small number of samples used in these tests. I propose addressing this by removing more randomness by providing `seed`, `deterministic=True`, and `force_col_wise=True`.

I ran the `test_classifier_custom_objective` test as of this branch 50 consecutive times, using [my dockerized setup for testing `lightgbm.dask`](https://github.com/jameslamb/lightgbm-dask-testing/blob/72c9c3f9b3008e4c061a758a805c5e0cea53c08e/Makefile#L82), and all 50 runs succeeded.

tests/python_package_test/test_dask.py
```diff
@@ -700,6 +802,55 @@ def test_regressor_quantile(output, alpha, cluster):
...
+        # Scores should be the same
+        assert_eq(s1, s2, atol=0.01)
+        assert_eq(s1, s1_local)
+
+        # local and Dask predictions should be the same
+        assert_eq(p1, p1_local)
+
+        # predictions should be better than random
+        assert_eq(p1, y, rtol=0.5, atol=50.)
+        assert_eq(p2, y, rtol=0.5, atol=50.)
```
StrikerRUS, line 851: I believe it's important for test case semantics here to use the same settings for `assert_eq()` in these two lines. So this my suggestion is not for code deduplication but is for ensuring that possible future precision corrections will be done simultaneously.
```suggestion
        # predictions should be better than random
        assert_precision = {"rtol": 0.5, "atol": 50.}
        assert_eq(p1, y, **assert_precision)
        assert_eq(p2, y, **assert_precision)
```

tests/python_package_test/test_dask.py
```diff
@@ -808,6 +959,68 @@ def test_ranker(output, group, boosting_type, tree_learner, cluster):
...
+
+        local_ranker = lgb.LGBMRanker(**params)
+        local_ranker.fit(X, y, sample_weight=w, group=g)
+        rnkvec_local = local_ranker.predict(X)
+
+        # distributed ranker should be able to rank decently well with the least-squares objective
+        # and should have high rank correlation with scores from serial ranker.
+        dcor = spearmanr(rnkvec_dask, y).correlation
+        assert dcor > 0.6
+        assert spearmanr(rnkvec_dask, rnkvec_local).correlation > 0.8
```
StrikerRUS, line 1016: For better readability. `dcor` isn't used later anywhere.
```suggestion
        assert spearmanr(rnkvec_dask, y).correlation > 0.6
        assert spearmanr(rnkvec_dask, rnkvec_local).correlation > 0.8
```

StrikerRUS (commented):
Thank you very much for digging deep into multiclass objective problems!
I left some last minor comments below.

tests/python_package_test/test_dask.py
jameslamb: ok sure, makes sense.

tests/python_package_test/test_dask.py
```diff
@@ -455,6 +487,76 @@ def test_classifier_pred_contrib(output, task, cluster):
...
+        assert callable(dask_classifier.objective_)
+        assert callable(dask_classifier_local.objective_)
+
+        # should correctly classify every sample
+        assert_eq(p1_class, y)
+        assert_eq(p1_class_local, y)
+        assert_eq(p2_class, y)
+
+        # probability estimates should be similar
+        assert_eq(p1_proba, p2_proba, atol=0.04)
```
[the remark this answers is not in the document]
jameslamb:
```suggestion
        # probability estimates should be similar
        assert_eq(p1_proba, p2_proba, atol=0.04)
        assert_eq(p1_proba, p1_proba_local)
```

Good point! I chose to compare `p1_proba` and `p1_proba_local`. If those aren't *identical*, that suggests a problem with syncing model information across the workers.

StrikerRUS (approved):
LGTM!

jameslamb:
Thanks! @jmoralez do you want a chance to re-review? Since I made some nontrivial changes after your earlier ✅

StrikerRUS:
> I noticed that Linux Azure Pipelines CI jobs started to timeout randomly two or three days ago.

> Thanks for letting me know! Looks like that is what happened here as well.


I guess the following warning from timeouted CI job can help us to find a root cause:

```
2022-01-14T22:12:20.3133305Z tests/python_package_test/test_sklearn.py::test_sklearn_integration[LGBMRegressor()-check_regressors_train(readonly_memmap=True,X_dtype=float32)]
2022-01-14T22:12:20.3133965Z   /root/miniconda/envs/test-env/lib/python3.8/site-packages/threadpoolctl.py:546: RuntimeWarning: 
2022-01-14T22:12:20.3134409Z   Found Intel OpenMP ('libiomp') and LLVM OpenMP ('libomp') loaded at
2022-01-14T22:12:20.3134727Z   the same time. Both libraries are known to be incompatible and this
2022-01-14T22:12:20.3135026Z   can cause random crashes or deadlocks on Linux when loaded in the
2022-01-14T22:12:20.3135283Z   same Python program.
2022-01-14T22:12:20.3135545Z   Using threadpoolctl may cause crashes or deadlocks. For more
2022-01-14T22:12:20.3135827Z   information and possible workarounds, please see
2022-01-14T22:12:20.3136165Z       https://github.com/joblib/threadpoolctl/blob/master/multiple_openmp.md
2022-01-14T22:12:20.3136411Z   
2022-01-14T22:12:20.3136614Z     warnings.warn(msg, RuntimeWarning)
```

tests/python_package_test/test_dask.py
```diff
@@ -455,6 +487,77 @@ def test_classifier_pred_contrib(output, task, cluster):
...
+            **params
+        )
+        dask_classifier = dask_classifier.fit(dX, dy, sample_weight=dw)
+        dask_classifier_local = dask_classifier.to_local()
+        p1_raw = dask_classifier.predict(dX, raw_score=True).compute()
+        p1_raw_local = dask_classifier_local.predict(X, raw_score=True)
+
+        # with a custom objective, prediction result is a raw score instead of predicted class
+        p1_proba = 1.0 / (1.0 + np.exp(-p1_raw))
+        p1_proba_local = 1.0 / (1.0 + np.exp(-p1_raw_local))
```
jmoralez, line 531: Shouldn't this be sigmoid for binary and softmax for multiclass?

tests/python_package_test/test_dask.py
```diff
@@ -455,6 +487,77 @@ def test_classifier_pred_contrib(output, task, cluster):
...
+
+        local_classifier = lgb.LGBMClassifier(**params)
+        local_classifier.fit(X, y, sample_weight=w)
+        p2_raw = local_classifier.predict(X, raw_score=True)
+        p2_proba = 1.0 / (1.0 + np.exp(-p2_raw))
```
jmoralez, line 536: Same as above

tests/python_package_test/test_dask.py
jameslamb: ah!! yeah you are absolutely right, thanks for catching that!

updated in https://github.com/microsoft/LightGBM/pull/4920/commits/2e2752cbf4975e71f9945c8e8ac18047dd234abb

In that commit, I also reduced the tolerance when comparing `p*_proba` from 0.04 to 0.03, matching the threshold used in `test_classifier`. I think this mistake I made was the reason that I couldn't reliably get those to be within 0.03 of each other.

I re-ran this test 20 times in a row and it passed every time.

tests/python_package_test/test_dask.py
jameslamb: addressed in https://github.com/microsoft/LightGBM/pull/4920/commits/2e2752cbf4975e71f9945c8e8ac18047dd234abb

jmoralez (approved):

PR #4927, opened 2022-01-06
[python-package] add support for pandas nullable types

### LINKED ISSUE #4173: Add support for pandas nullable types to the sklearn api
## Summary

I would like to use lightgbm's sklearn api with the new pandas nullable types. Currently, lightgbm does not recognize these types as valid. Reproducer using `3.0.0` and pandas `0.24.1`:

```python
from lightgbm.sklearn import LGBMClassifier
import pytest
import pandas as pd

for dtype in ['Int64', 'Float64', 'boolean']:
    
    lr = LGBMClassifier()

    # ValueError if only features use nullable types
    df = pd.DataFrame({"a": pd.Series([True, False, True, False], dtype=dtype),
                       "b": pd.Series([4, 5, 6])})
    y = pd.Series([1, 0, 1, 0])
    
    with pytest.raises(ValueError):
        lr.fit(df, y)
    
    # ValueError if only target uses nullable types
    df = pd.DataFrame({"a": pd.Series([True, False, True, False])})
    y = pd.Series([1, 0, 1, 0], dtype=dtype)
    
    with pytest.raises(ValueError):
        lr.fit(df, y)

# No ValueError if use "old" types
for dtype in ['int', 'float', 'bool']:
    
    df = pd.DataFrame({"a": pd.Series([True, False, True, False], dtype=dtype)})
    y = pd.Series([1, 0, 1, 0])

    lr = LGBMClassifier()
    lr.fit(df, y)
```

## Motivation

As a user, I would like to build a data processing pipeline using the latest pandas features. I can work around lightgbm's limitations but this is cumbersome and some information could be lost in the conversion. 

Thank you!

## References
Pandas nullable integer dtype [documentation](https://pandas.pydata.org/docs/user_guide/integer_na.html).

jmoralez:
This broadens the accepted pandas dataframes dtypes to include the pandas nullable dtypes (Int, boolean and Float).

Closes #4173.


python-package/lightgbm/basic.py
```diff
@@ -502,14 +502,19 @@ def c_int_array(data):
 
 
 def _get_bad_pandas_dtypes(dtypes):
-    pandas_dtype_mapper = {'int8': 'int', 'int16': 'int', 'int32': 'int',
-                           'int64': 'int', 'uint8': 'int', 'uint16': 'int',
-                           'uint32': 'int', 'uint64': 'int', 'bool': 'int',
-                           'float16': 'float', 'float32': 'float', 'float64': 'float'}
```
StrikerRUS, line 508: Thanks a lot for this PR!
Could you please add one more test that checks LightGBM works with all these types (different `[u]int`s)?

tests/python_package_test/test_basic.py
```diff
@@ -609,3 +609,44 @@ def test_custom_objective_safety():
...
+    df2['x3'] = df2['x3'].astype('boolean')
+
+    # test training succeeds
+    ds_nullable_dtypes = lgb.Dataset(df2, y)
+    bst_nullable_dtypes = lgb.train(params, ds_nullable_dtypes, num_boost_round=5)
+
+    # test equal predictions
+    preds = bst.predict(df)
+    preds_nullable_dtypes = bst_nullable_dtypes.predict(df2)
+    np.testing.assert_allclose(preds, preds_nullable_dtypes)
```
jameslamb, line 652: I like the structure of this test, but I think it could pass even if this support wasn't working correctly. For example, with purely-random input data it's possible that both of the models trained here will just return the sample mean (e.g. not add any leaves, because all splits have 0 or too-low gain). In that situation, the types of issues you'd expect from a type-casting change (like loss of precision or incorrect data because of incorrect pointer math) might not be caught.

Could you modify this test to use one of the datasets used in the other tests here, and check things like the score on the training data being "ok" and the model having more leaves than `num_boost_round`?

jameslamb (changes requested):
Thanks for this! I left a few small suggestions.

python-package/lightgbm/basic.py
```diff
@@ -546,9 +551,8 @@ def _data_from_pandas(data, feature_name, categorical_feature, pandas_categorica
             raise ValueError("DataFrame.dtypes for data must be int, float or bool.\n"
                              "Did not expect the data types in the following fields: "
                              f"{bad_index_cols_str}")
-        data = data.values
-        if data.dtype != np.float32 and data.dtype != np.float64:
-            data = data.astype(np.float32)
+        target_dtype = _find_commmon_dtype(data.dtypes)
+        data = data.to_numpy(dtype=target_dtype, na_value=np.nan)
```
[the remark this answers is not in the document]
jmoralez: Ah good point. The problem with these dtypes is that they use `pd.NA` as their missing value, so using `.values` with this value present returns an array with object dtype because `pd.NA` has type `NAType` and numpy can't convert it to float. I'll try to find another approach.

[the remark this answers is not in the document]
jmoralez: Got it, `data.astype(target_dtype).values` works.

tests/python_package_test/test_basic.py
jmoralez: I've modified the test for the nullable dtypes in e1cf6c9cd759a735dcf5bf68f015e951b255c677. I used a synthetic dataset because I needed a boolean and some ints and most of the real datasets have floats but I made sure that all of the features were used and that the score was good. Let me know what you think.

python-package/lightgbm/basic.py
jmoralez: I've added a test to compare the results using all float64 and using combinations of uints, ints and floats in e1cf6c9cd759a735dcf5bf68f015e951b255c677. Let me know what you think.

jmoralez:
Hmm, got this error `/home/runner/miniconda/envs/test-env/lib/R/bin/exec/R: symbol lookup error: /home/runner/miniconda/envs/test-env/lib/R/bin/exec/../../lib/../../libreadline.so.8: undefined symbol: tputs` in the lint task [logs](https://github.com/microsoft/LightGBM/runs/4771044438?check_suite_focus=true#step:3:754).

python-package/lightgbm/basic.py
```diff
@@ -502,14 +502,19 @@ def c_int_array(data):
...
 def _get_bad_pandas_dtypes(dtypes):
-    pandas_dtype_mapper = {'int8': 'int', 'int16': 'int', 'int32': 'int',
-                           'int64': 'int', 'uint8': 'int', 'uint16': 'int',
-                           'uint32': 'int', 'uint64': 'int', 'bool': 'int',
-                           'float16': 'float', 'float32': 'float', 'float64': 'float'}
-    bad_indices = [i for i, dtype in enumerate(dtypes) if (dtype.name not in pandas_dtype_mapper
-                                                           and (not is_dtype_sparse(dtype)
-                                                                or dtype.subtype.name not in pandas_dtype_mapper))]
-    return bad_indices
+    return [i for i, dtype in enumerate(dtypes) if not is_numeric_dtype(dtype)]
```
StrikerRUS, line 505: By switching from fixed mapper to general `is_numeric_dtype()` function we are allowing to pass new types that were not allowed previously. For example, `np.complex64` and `np.float128`. In other words, every subclass of `numpy.number`:
```
>>> subdtypes(np.generic)

[numpy.generic,
 [[numpy.number,
   [[numpy.integer,
     [[numpy.signedinteger,
       [numpy.int8,
        numpy.int16,
        numpy.int32,
        numpy.int64,
        numpy.longlong,
        numpy.timedelta64]],
      [numpy.unsignedinteger,
       [numpy.uint8,
        numpy.uint16,
        numpy.uint32,
        numpy.uint64,
        numpy.ulonglong]]]],
    [numpy.inexact,
     [[numpy.floating,
       [numpy.float16, numpy.float32, numpy.float64, numpy.float128]],
      [numpy.complexfloating,
       [numpy.complex64, numpy.complex128, numpy.complex256]]]]]],
  [numpy.flexible,
   [[numpy.character, [numpy.bytes_, numpy.str_]],
    [numpy.void, [numpy.record]]]],
  numpy.bool_,
  numpy.datetime64,
  numpy.object_]]
```
https://pandas.pydata.org/docs/user_guide/basics.html#:~:text=All%20NumPy%20dtypes%20are%20subclasses%20of%20numpy.generic%3A

python-package/lightgbm/basic.py
```diff
@@ -546,9 +551,8 @@ def _data_from_pandas(data, feature_name, categorical_feature, pandas_categorica
             raise ValueError("DataFrame.dtypes for data must be int, float or bool.\n"
                              "Did not expect the data types in the following fields: "
                              f"{bad_index_cols_str}")
-        data = data.values
-        if data.dtype != np.float32 and data.dtype != np.float64:
```
StrikerRUS, line 550: This `if` statement was introduced with the aim to save memory in case of original data already have `float[32|64]` type: #2383. With new unconditional `data.astype(target_dtype).values` expression and [default argument `copy=True` of `astype()` function](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.astype.html), we are loosing that efficiency improvement. Am I wrong?

python-package/lightgbm/basic.py
jmoralez: Oh you're right, I hadn't thought about that. Do you think it'd be better to explicitly list all the allowed dtypes and just do something like: `[i for i, dtype in enumerate(dtypes) if isinstance(dtype, allowed_dtypes)`

python-package/lightgbm/basic.py
jmoralez: The `np.ndarray.astype`s `copy` argument also defaults to `True` ([docs](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.astype.html#numpy.ndarray.astype)) so I think the behaviour is the same as it previously was. We could set `copy=False` here if we won't modify the values, and if the dtype matches (all of the columns are either f32 or f64) there won't be any copies.

jmoralez: It'd be nice to test if any copies are made.

StrikerRUS: > The `np.ndarray.astype`s `copy` argument also defaults to `True` ([docs](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.astype.html#numpy.ndarray.astype)) so I think the behaviour is the same as it previously was.

Previously, `np.ndarray.astype()` **wasn't executed at all** if dtype is `float` already. So, I think the behaviour is changed here.

jmoralez: I added the `copy=False` argument and added a test to check that no copies are made for a single float dtype df in 98325e920def79a565b36fdcf7ede8a2b70c381e

python-package/lightgbm/basic.py
jmoralez: I used the idea of what [is_numeric_dtype](https://github.com/pandas-dev/pandas/blob/66e3805b8cabe977f40c05259cc3fcf7ead5687d/pandas/core/dtypes/common.py#L1188) does and checked for more specific subclasses (np.integer, np.floating and np.bool_) in 98325e920def79a565b36fdcf7ede8a2b70c381e. Running `{dtype: subdtypes(dtype) for dtype in (np.integer, np.floating, np.bool_)}` with the subdtypes function in the pandas link you posted returns:
```
{numpy.integer: [numpy.integer,
  [[numpy.signedinteger,
    [numpy.int8,
     numpy.int16,
     numpy.int32,
     numpy.int64,
     numpy.longlong,
     numpy.timedelta64]],
   [numpy.unsignedinteger,
    [numpy.uint8,
     numpy.uint16,
     numpy.uint32,
     numpy.uint64,
     numpy.ulonglong]]]],
 numpy.floating: [numpy.floating,
  [numpy.float16, numpy.float32, numpy.float64, numpy.float128]],
 numpy.bool_: numpy.bool_}
```
and I removed the np.float128 and np.timedelta64. Let me know what you think

jmoralez:
Hmm seems like there's no np.float128 in windows haha.

StrikerRUS:
> Hmm seems like there's no np.float128 in windows

Seems that's true: https://github.com/winpython/winpython/issues/613.

tests/python_package_test/test_basic.py
jameslamb: Ok yeah that looks good to me! Thanks

python-package/lightgbm/basic.py
StrikerRUS: Note for other reviewers.
Casting to float[32|64] is done later here
https://github.com/microsoft/LightGBM/blob/e6a2f7162cf0bae4cbad6223c2700d74f0d67c4b/python-package/lightgbm/basic.py#L1595-L1598
for creating dataset and here
https://github.com/microsoft/LightGBM/blob/e6a2f7162cf0bae4cbad6223c2700d74f0d67c4b/python-package/lightgbm/basic.py#L849-L852
for predicting.

tests/python_package_test/test_basic.py
```diff
@@ -609,3 +610,101 @@ def test_custom_objective_safety():
     good_bst_multi.update(fobj=_good_gradients)
     with pytest.raises(ValueError, match=re.escape(f"number of models per one iteration ({nclass})")):
         bad_bst_multi.update(fobj=_bad_gradients)
+
+
+def test_numpy_regular_dtypes():
```
StrikerRUS, line 615:
```suggestion
def test_pandas_with_numpy_regular_dtypes():
```

Also, please move this and `test_pandas_nullable_dtypes` tests to `test_engine.py` file because `lgb.train()` function that is defined in `engine.py` is used in these tests.

python-package/lightgbm/basic.py
StrikerRUS: > Looking at this a bit more closely I think I could add `np.float32` to ...

Sorry, didn't get how adding `np.float32` will help to avoid a loss of precision...

jmoralez: Sorry. If we add floats to that function it will cast to float64, i.e.
```python
import numpy as np

np.find_common_type([np.int32, np.float32], [])  # float64
```
That also avoids a copy if we have like [int16, int32], the common dtype would be int32 so a copy would be made and then when casting to float another copy would be made. By including float32 in there we can cast to the target dtype only once.

StrikerRUS: uhu, thanks for the explanation! I think it's great idea!

jmoralez: added in 530828b

tests/python_package_test/test_basic.py
jmoralez: addressed in 8bf1617

StrikerRUS (approved):
Thank you so much for this feature! I left two very minor stylish comments and one suggestion for checking model trained on nullable dtypes in test for a support of nullable dtypes.

Also, I think we should mark this PR as `breaking` due to new smart casting algorithm (https://github.com/microsoft/LightGBM/pull/4927#discussion_r790930269). Right now anything non-float is casted to float32 unconditionally.

StrikerRUS:
@jmoralez Please update this branch to unblock merging button for this PR.

@jameslamb Would you like to take a look at this PR?

jameslamb:
> @jameslamb Would you like to take a look at this PR?

Yes please. I would like to review this.

jameslamb (approved):
Amazing work, thank you! I read through the conversations and reviewed the diff, and agree with the decisions that were made.

I also learned some new `numpy` features from you through this PR! `np.shares_memory()` to detect if copies were made is awesome, I'll definitely use that in other projects in the future.

jameslamb:
Merging this since both @StrikerRUS  and I have approved. @jmoralez sorry I took so long to get back to this and provide a review.

Issue #4926, opened 2022-01-05
[gpu] Large dataset In LGBMRegressor Failed


jiapengwen:
## Description
use gpu version, I have a Large  dataset, but execute failed 

## Reproducible example
``` python
import pandas as pd
import numpy as np
import os
import time
from lightgbm import LGBMRegressor
from sklearn.datasets import make_classification
from sklearn.datasets import make_regression
from sklearn.model_selection import train_test_split
import lightgbm as lgbm




if __name__ == '__main__':
    time1=time.time()
    # X=np.random.random((4300000,2200)) #work
    # y=np.random.random((4300000)) 

    X=np.random.random((8000000,2200))
    y=np.random.random((8000000))
    # X=np.random.random((7000000,2200)) # work
    # y=np.random.random((7000000))
    time2 = time.time()
    print('contruct data cost:',time2-time1)
    print(X[2][2])
    print(y[2:10])
    time1=time.time()
    model = lgbm.LGBMRegressor(device="gpu",n_estimators=1000,verbose=4,max_bin=16,tree_learner = 'serial',gpu_use_dp='false',n_jobs=1,
                                max_depth=7,num_leaves=31 ,min_child_samples=17000)
    model.fit(X, y,callbacks=[lgbm.log_evaluation()])
    time2 = time.time()
    print('gpu cost:',time2-time1)


######################################
[LightGBM] [Warning] Accuracy may be bad since you didn't explicitly set num_leaves OR 2^max_depth > num_leaves. (num_leaves=31).
[LightGBM] [Info] This is the GPU trainer!!
[LightGBM] [Info] Total Bins 35200
[LightGBM] [Info] Number of data points in the train set: 8000000, number of used features: 2200
[LightGBM] [Info] Using GPU Device: Quadro RTX 6000, Vendor: NVIDIA Corporation
[LightGBM] [Info] Compiling OpenCL Kernel with 16 bins...
[LightGBM] [Info] GPU programs have been built
[LightGBM] [Info] Size of histogram bin entry: 8
terminate called after throwing an instance of 'boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::compute::opencl_error> >'
  what():  Memory Object Allocation Failure

```

## Environment info

LightGBM version or commit hash:

LightGBM version "3.3.1";

Command(s) you used to install LightGBM

```shell
pip3 install lightgbm --install=--gpu

```

python version python3.6.9

## Additional Comments
My machine has 227G memory,

[...]

PR #4928, opened 2022-01-06
gpu allocate memory overflow

jiapengwen:
num_data_ is int_32, when 8000000(800w)* 275, it's overflow,  please fix

### DIFF
--- a/src/treelearner/gpu_tree_learner.cpp
+++ b/src/treelearner/gpu_tree_learner.cpp
@@ -503,7 +503,7 @@ void GPUTreeLearner::AllocateGPUMemory() {
     }
     // copying the last 1 to (dword_features - 1) feature-groups in the last tuple
     queue_.enqueue_write_buffer(device_features_->get_buffer(),
-                        (num_dense_feature4_ - 1) * num_data_ * sizeof(Feature4), num_data_ * sizeof(Feature4), host4);
+                        (num_dense_feature4_ - 1) * (uint64_t)num_data_ * sizeof(Feature4), num_data_ * sizeof(Feature4), host4);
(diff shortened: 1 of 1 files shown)


src/treelearner/gpu_tree_learner.cpp
```diff
@@ -245,7 +245,7 @@ void GPUTreeLearner::AllocateGPUMemory() {
   }
   // allocate memory for all features (FIXME: 4 GB barrier on some devices, need to split to multiple buffers)
   device_features_.reset();
-  device_features_ = std::unique_ptr<boost::compute::vector<Feature4>>(new boost::compute::vector<Feature4>(num_dense_feature4_ * num_data_, ctx_));
+  device_features_ = std::unique_ptr<boost::compute::vector<Feature4>>(new boost::compute::vector<Feature4>((uint64_t)num_dense_feature4_ * num_data_, ctx_));
```
shiyu1994, line 248: @jiapengwen Thanks you for working on this! I think we should also have a fix here,
https://github.com/microsoft/LightGBM/blob/1d8bfd8c9b3fa7d56db81201adedb61cfc11669c/src/treelearner/gpu_tree_learner.cpp#L505-L506
could you help with this?

jiapengwen: yes, u r right, it may be overflow, I added for  robus

shiyu1994 (approved):

jameslamb:
Thanks for this!

I've just added "fixes #4926" to the description. In the future, when you open a pull request in this project that is related to an existing conversation in the project's backlog or other pull request, please link to it so that reviewers have sufficient context.

PR #4939, opened 2022-01-08
[ci] use `-1` for exit code in case of failure

StrikerRUS:
Just to keep the repository codebase consistent. We use either `0` or `-1` as exit code everywhere, see e.g.
https://github.com/microsoft/LightGBM/blob/master/.ci/test.sh
https://github.com/microsoft/LightGBM/blob/master/.github/workflows/cuda.yml
https://github.com/microsoft/LightGBM/blob/master/.github/workflows/r_package.yml

### DIFF
--- a/.github/workflows/r_package.yml
+++ b/.github/workflows/r_package.yml
@@ -209,7 +209,7 @@ jobs:
         run: |
           cd R-package/tests
           exit_code=0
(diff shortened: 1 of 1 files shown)


jameslamb (approved):
Sure, ok with me.

PR #4940, opened 2022-01-09
[tests][R-package] use one `#` symbol for comments

StrikerRUS:
Refer to https://github.com/microsoft/LightGBM/pull/4902#discussion_r773553163.


jameslamb (approved):
thanks for cleaning these up

PR #4941, opened 2022-01-09
[R-package] Fix custom objective detection in `print.lgb.Booster()`

StrikerRUS:
As the best of my knowledge, string comparison in R is case-sensitive. I think it's better to use lowercased `"none"` string value for objective parameter to avoid errors.
https://github.com/microsoft/LightGBM/blob/34f949782252d202097539124149e0d5f69000bd/R-package/R/lgb.Booster.R#L247-L250
https://github.com/microsoft/LightGBM/blob/34f949782252d202097539124149e0d5f69000bd/R-package/R/lgb.Booster.R#L864-L866

### DIFF
--- a/R-package/R/lgb.Booster.R
+++ b/R-package/R/lgb.Booster.R
@@ -861,7 +861,7 @@ print.lgb.Booster <- function(x, ...) {
 
   if (!handle_is_null) {
     obj <- x$params$objective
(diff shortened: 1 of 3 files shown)


jameslamb (approved):
I support this change and the simplification in `print.lgb.Booster()`, thanks!

PR #4947, opened 2022-01-14
[python-package] make record_evaluation compatible with cv (fixes #4943)

### LINKED ISSUE #4943: [python-package] record_evaluation callback doesn't work with cv function
## Description

The `record_evaluation` callback fails when trying to use it with the `cv` function. Since some integrations that log metrics may use that callback to save the training information and the `cv` function is very useful to do hyperparameter tuning I believe these two should work together.

## Reproducible example

```python
import lightgbm as lgb
import numpy as np

X = np.random.rand(100, 2)
y = np.random.rand(100)
ds = lgb.Dataset(X, y)
params = {'objective': 'l2', 'num_leaves': 5, 'verbose': -1}
eval_result = {}
callbacks = [lgb.record_evaluation(eval_result)]
bst = lgb.train(params, ds, num_boost_round=2, valid_sets=[ds], callbacks=callbacks)
print(eval_result)
# {'training': OrderedDict([('l2', [0.08664710410521784, 0.08580684200793261])])}

eval_result = {}
callbacks = [lgb.record_evaluation(eval_result)]
cv_hist = lgb.cv(
    params, ds, num_boost_round=2, stratified=False, callbacks=callbacks, eval_train_metric=True
)
# Traceback (most recent call last):
#   File "record_eval.py", line 14, in <module>
#     cv_hist = lgb.cv(params, ds, stratified=False, callbacks=callbacks)
#   File "/hdd/github/LightGBM/python-package/lightgbm/engine.py", line 582, in cv
#     cb(callback.CallbackEnv(model=cvfolds,
#   File "/hdd/github/LightGBM/python-package/lightgbm/callback.py", line 140, in _callback
#     _init(env)
#   File "/hdd/github/LightGBM/python-package/lightgbm/callback.py", line 134, in _init
#     for data_name, eval_name, _, _ in env.evaluation_result_list:
# ValueError: too many values to unpack (expected 4)
```
This can be easily fixed by capturing the extra content from `env.evaluation_result_list` with `*_` here: https://github.com/microsoft/LightGBM/blob/4aaeb22932b72aaff632b40ba596ca2533071ca2/python-package/lightgbm/callback.py#L134 and here: https://github.com/microsoft/LightGBM/blob/4aaeb22932b72aaff632b40ba596ca2533071ca2/python-package/lightgbm/callback.py#L141
which would yield:
`{'cv_agg': OrderedDict([('train l2', [0.06319766819578873, 0.06253713563921684]), ('valid l2', [0.06679350623951755, 0.06694167044391186])])}`

## Environment info

LightGBM version or commit hash: db045f40305dfeae0b1dcff0658dbaa172dcab37

## Additional Comments

I'm happy to make the described changes if other maintainers agree with my solution.

jmoralez:
Makes the record_evaluation callback work with the cv function and returns a similar output to the one returned when used with the train function, i.e.
```
{
  "train": {"metric1": [...], "metric2": [...]},
  "valid": {"metric1": [...], "metric2": [...]}
}
```

Closes #4943.

### DIFF
--- a/python-package/lightgbm/callback.py
+++ b/python-package/lightgbm/callback.py
@@ -131,14 +131,24 @@ def record_evaluation(eval_result: Dict[str, Dict[str, List[Any]]]) -> Callable:
 
     def _init(env: CallbackEnv) -> None:
         eval_result.clear()
-        for data_name, eval_name, _, _ in env.evaluation_result_list:
+        for item in env.evaluation_result_list:
+            if len(item) == 4:  # regular train
+                data_name, eval_name = item[:2]
+            else:  # cv
+                data_eval = item[1]
+                data_name, eval_name = data_eval.split()
             eval_result.setdefault(data_name, collections.OrderedDict())
             eval_result[data_name].setdefault(eval_name, [])
 
     def _callback(env: CallbackEnv) -> None:
         if env.iteration == env.begin_iteration:
             _init(env)
-        for data_name, eval_name, result, _ in env.evaluation_result_list:
+        for item in env.evaluation_result_list:
+            if len(item) == 4:
+                data_name, eval_name, result = item[:3]
+            else:
+                data_eval, result = item[1:3]
+                data_name, eval_name = data_eval.split()
             eval_result[data_name][eval_name].append(result)
     _callback.order = 20  # type: ignore
     return _callback
--- a/python-package/lightgbm/callback.py
+++ b/python-package/lightgbm/callback.py
@@ -144,7 +144,7 @@ def record_evaluation(eval_result: Dict[str, Dict[str, List[Any]]]) -> Callable:
             if len(item) == 4:  # regular train
                 data_name, eval_name = item[:2]
             else:  # cv
-                data_name, eval_name = _extract_data_and_eval_name(item[1])
+                data_name, eval_name = item[1].split()
             eval_result.setdefault(data_name, collections.OrderedDict())
             if len(item) == 4:
                 eval_result[data_name].setdefault(eval_name, [])
@@ -160,7 +160,7 @@ def record_evaluation(eval_result: Dict[str, Dict[str, List[Any]]]) -> Callable:
                 data_name, eval_name, result = item[:3]
                 eval_result[data_name][eval_name].append(result)
             else:
-                data_name, eval_name = _extract_data_and_eval_name(item[1])
+                data_name, eval_name = item[1].split()
                 res_mean, res_stdv = item[2], item[4]
                 eval_result[data_name][f'{eval_name}-mean'].append(res_mean)
                 eval_result[data_name][f'{eval_name}-stdv'].append(res_stdv)
--- a/python-package/lightgbm/callback.py
+++ b/python-package/lightgbm/callback.py
@@ -129,15 +129,6 @@ def record_evaluation(eval_result: Dict[str, Dict[str, List[Any]]]) -> Callable:
     if not isinstance(eval_result, dict):
         raise TypeError('eval_result should be a dictionary')
 
-    def _extract_data_and_eval_name(data_eval):
-        data_eval = data_eval.split()
-        if len(data_eval) == 1:  # no train metric
-            data_name = 'valid'
-            eval_name = data_eval[0]
-        else:
-            data_name, eval_name = data_eval
-        return data_name, eval_name
-
     def _init(env: CallbackEnv) -> None:
         eval_result.clear()
         for item in env.evaluation_result_list:
--- a/python-package/lightgbm/callback.py
+++ b/python-package/lightgbm/callback.py
@@ -129,16 +129,28 @@ def record_evaluation(eval_result: Dict[str, Dict[str, List[Any]]]) -> Callable:
     if not isinstance(eval_result, dict):
         raise TypeError('eval_result should be a dictionary')
 
+    def _extract_data_and_eval_name(data_eval):
+        data_eval = data_eval.split()
+        if len(data_eval) == 1:  # no train metric
+            data_name = 'valid'
+            eval_name = data_eval[0]
+        else:
+            data_name, eval_name = data_eval
+        return data_name, eval_name
+
     def _init(env: CallbackEnv) -> None:
         eval_result.clear()
         for item in env.evaluation_result_list:
             if len(item) == 4:  # regular train
                 data_name, eval_name = item[:2]
             else:  # cv
-                data_eval = item[1]
-                data_name, eval_name = data_eval.split()
+                data_name, eval_name = _extract_data_and_eval_name(item[1])
             eval_result.setdefault(data_name, collections.OrderedDict())
-            eval_result[data_name].setdefault(eval_name, [])
+            if len(item) == 4:
+                eval_result[data_name].setdefault(eval_name, [])
+            else:
+                eval_result[data_name].setdefault(f'{eval_name}-mean', [])
+                eval_result[data_name].setdefault(f'{eval_name}-stdv', [])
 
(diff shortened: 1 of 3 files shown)


tests/python_package_test/test_engine.py
```diff
@@ -3224,3 +3224,34 @@ def test_force_split_with_feature_fraction(tmp_path):
...
+
+
+def test_record_evaluation_with_cv():
+    X, y = make_synthetic_regression()
+    ds = lgb.Dataset(X, y)
+    eval_result = {}
+    callbacks = [lgb.record_evaluation(eval_result)]
+    metrics = ['l2', 'rmse']
+    params = {'objective': 'l2', 'num_leaves': 3, 'metric': metrics}
+    cv_hist = lgb.cv(params, ds, num_boost_round=5, stratified=False, callbacks=callbacks, eval_train_metric=True)
```
StrikerRUS, line 3253: Proposed solution crashes if you set `eval_train_metric=False` in this test case.
```
ValueError                                Traceback (most recent call last)
<ipython-input-5-e50e682e859d> in <module>
----> 1 cv_hist = lgb.cv(params, data_train, num_boost_round=5, stratified=False, callbacks=callbacks, eval_train_metric=False)

D:\Miniconda3\lib\site-packages\lightgbm\engine.py in cv(params, train_set, num_boost_round, folds, nfold, stratified, shuffle, metrics, fobj, feval, init_model, feature_name, categorical_feature, early_stopping_rounds, fpreproc, verbose_eval, show_stdv, seed, callbacks, eval_train_metric, return_cvbooster)
    645         try:
    646             for cb in callbacks_after_iter:
--> 647                 cb(callback.CallbackEnv(model=cvfolds,
    648                                         params=params,
    649                                         iteration=i,

D:\Miniconda3\lib\site-packages\lightgbm\callback.py in _callback(env)
    141     def _callback(env: CallbackEnv) -> None:
    142         if not eval_result:
--> 143             _init(env)
    144         for item in env.evaluation_result_list:
    145             if len(item) == 4:

D:\Miniconda3\lib\site-packages\lightgbm\callback.py in _init(env)
    135             else:  # cv
    136                 data_eval = item[1]
--> 137                 data_name, eval_name = data_eval.split()
    138             eval_result.setdefault(data_name, collections.OrderedDict())
    139             eval_result[data_name].setdefault(eval_name, [])

ValueError: not enough values to unpack (expected 2, got 1)
```

item in `env.evaluation_result_list` will look like
```
('cv_agg', 'l2', 0.077931728049798, False, 0.00858679769235092)
```

python-package/lightgbm/callback.py
```diff
@@ -131,14 +131,24 @@ def record_evaluation(eval_result: Dict[str, Dict[str, List[Any]]]) -> Callable:
 
     def _init(env: CallbackEnv) -> None:
         eval_result.clear()
-        for data_name, eval_name, _, _ in env.evaluation_result_list:
+        for item in env.evaluation_result_list:
+            if len(item) == 4:  # regular train
+                data_name, eval_name = item[:2]
+            else:  # cv
```
StrikerRUS, line 137: What do you think about recording information about stdv as well? Just like it is done in the returned value from `cv()` function.
https://github.com/microsoft/LightGBM/blob/4aaeb22932b72aaff632b40ba596ca2533071ca2/python-package/lightgbm/engine.py#L576-L579
https://github.com/microsoft/LightGBM/blob/4aaeb22932b72aaff632b40ba596ca2533071ca2/python-package/lightgbm/engine.py#L364
https://github.com/microsoft/LightGBM/blob/ce486e5b45a6f5e67743e14765ed139ff8d532e5/tests/python_package_test/test_engine.py#L2253-L2260

StrikerRUS (changes requested):

python-package/lightgbm/callback.py
jmoralez: Added in 22a7bbe0a74db7121887026b8a483b08df2b0409

tests/python_package_test/test_engine.py
jmoralez: I fixed this for this callback by setting the first level key to `valid` even when there's no train metric being computed to preserve the structure but maybe it'd be better if the second element from this result list was 'valid l2' to be consistent. WDYT?

jmoralez: By that I mean removing the if in https://github.com/microsoft/LightGBM/blob/a06fadfb7ac3fdb26da3d4afc061a8b976070c50/python-package/lightgbm/engine.py#L370-L373 and set `key = f"{one_line[0]} {one_line[1]}"` always.

StrikerRUS: > ... but maybe it'd be better if the second element from this result list was 'valid l2' to be consistent.

Yeah, I support this.

jmoralez: changed in 914818c4bc70142c159d9865e0b1583486947c8d

StrikerRUS: I'm going to mark this PR as `breaking` due to this change.

tests/python_package_test/test_engine.py
```diff
@@ -3224,3 +3224,42 @@ def test_force_split_with_feature_fraction(tmp_path):
...
+    params = {'objective': 'l2', 'num_leaves': 3}
+    num_boost_round = 5
+    bst = lgb.train(params, ds, num_boost_round=num_boost_round, valid_sets=[ds], callbacks=callbacks)
+    assert list(eval_result.keys()) == ['training']
+    train_mses = []
+    for i in range(num_boost_round):
+        pred = bst.predict(X, num_iteration=i + 1)
+        mse = mean_squared_error(y, pred)
+        train_mses.append(mse)
+    np.testing.assert_allclose(eval_result['training']['l2'], train_mses)
```
StrikerRUS, line 3243: Love this test! 👍

python-package/lightgbm/callback.py
```diff
@@ -129,17 +129,41 @@ def record_evaluation(eval_result: Dict[str, Dict[str, List[Any]]]) -> Callable:
     if not isinstance(eval_result, dict):
         raise TypeError('eval_result should be a dictionary')
 
+    def _extract_data_and_eval_name(data_eval):
```
StrikerRUS, line 132: I can't find where this function is used...

tests/python_package_test/test_engine.py
```diff
@@ -3224,3 +3224,42 @@ def test_force_split_with_feature_fraction(tmp_path):
...
+    if train_metric:
+        expected_datasets.add('train')
+    assert set(eval_result.keys()) == expected_datasets
+    for dataset in expected_datasets:
+        for metric in metrics:
+            for agg in ('mean', 'stdv'):
+                key = f'{dataset} {metric}-{agg}'
+                np.testing.assert_allclose(
+                    cv_hist[key], eval_result[dataset][f'{metric}-{agg}']
+                )
```
StrikerRUS, line 3265: And this test is nice! 🙂

python-package/lightgbm/callback.py
jmoralez: I forgot to remove this in 914818c4bc70142c159d9865e0b1583486947c8d, removed in 54754fb83cd2826275a0b81a40b7d4742cb51dd5

StrikerRUS (approved):
LGTM, thank you very much!

StrikerRUS:
@jmoralez Please update this branch to unblock merging button for this PR.

@jameslamb Would you like to take a look at this PR?

jameslamb (approved):
> @jameslamb Would you like to take a look at this PR?

Thanks for the `@` @StrikerRUS ! I just looked through this and don't have any additional suggestions.

PR #2954, opened 2020-03-27
[ci] changed CRAN mirror to fix failing installs

jameslamb:
This fixes the issue we are seeing in Mac builds  across all PRs right now.

@StrikerRUS I know you tried a similar change in https://github.com/microsoft/LightGBM/pull/2949#issuecomment-605034139 and said it didn't work. I don't see anything obviously wrong with your approach, but I do know the change I'm suggesting in this PR works on my Mac.

Changes:

1. Change from RStudio CRAN mirror to  https://cloud.r-project.org/
2. Make that explicit, instead of relying on `.Rprofile`

Since  we only have one `install.packages()` call, I don't think we need to deal with the complexity of when and if the `.Rprofile` file gets sourced.

on my Mac: 

Screen Shot 2020-03-27 at 10 23 28 AM

My best  guess about what happened is:

1. R binaries for macOS became temporarily unavailable on CRAN
2. this broke whatever process mirrors them from CRAN's main repository to https//cran.rstudio.com
3. CRAN repository was still broken when @StrikerRUS tried it (see that comment above)
4. CRAN is now fixed so it succeeded for me
5. RStudio still hasn't caught up (maybe CRAN fixed the issue in a way that breaks RStudio's mirroring or something)

Even if that is true and a few hours from now the RStudio repo will be back up, I _still_ think this PR should be accepted. It removes unnecessary indirection (with `.Rprofile`) and points our builds at the real source of truth for R packages.

PR #4953, opened 2022-01-16
[ci] use conda-forge in Linux and macOS CI jobs

### LINKED ISSUE #4948: Linux CI jobs hang forever after completing all Python tests successfully
This problem has started to timeout our CI jobs about 5 days ago. The most frequent CI jobs that run out of allowed 60min limit are `Linux_latest regular` and `Linux_latest sdist` at Azure Pipelines. Also, I just saw `CUDA Version / cuda 10.0 pip (linux, clang, Python 3.8)` encountered the same problem.

From test logs I guess that the root cause is connected to the following warning message from the `joblib/threadpoolctl` package:
```
2022-01-14T22:12:20.3133305Z tests/python_package_test/test_sklearn.py::test_sklearn_integration[LGBMRegressor()-check_regressors_train(readonly_memmap=True,X_dtype=float32)]
2022-01-14T22:12:20.3133965Z   /root/miniconda/envs/test-env/lib/python3.8/site-packages/threadpoolctl.py:546: RuntimeWarning: 
2022-01-14T22:12:20.3134409Z   Found Intel OpenMP ('libiomp') and LLVM OpenMP ('libomp') loaded at
2022-01-14T22:12:20.3134727Z   the same time. Both libraries are known to be incompatible and this
2022-01-14T22:12:20.3135026Z   can cause random crashes or deadlocks on Linux when loaded in the
2022-01-14T22:12:20.3135283Z   same Python program.
2022-01-14T22:12:20.3135545Z   Using threadpoolctl may cause crashes or deadlocks. For more
2022-01-14T22:12:20.3135827Z   information and possible workarounds, please see
2022-01-14T22:12:20.3136165Z       https://github.com/joblib/threadpoolctl/blob/master/multiple_openmp.md
2022-01-14T22:12:20.3136411Z   
2022-01-14T22:12:20.3136614Z     warnings.warn(msg, RuntimeWarning)
```

jameslamb:
Contributes to #4948.

This PR attempts to switch LightGBM's Linux and macOS CI jobs to use `conda-forge` instead of the anaconda default channels. Based on the discussion in #4948, this PR proposes doing that through the use of `mambaforge` ([link](https://github.com/conda-forge/miniforge#mambaforge)).

### Note for reviewers

~Opening this as a draft, to test changes on LightGBM's real CI setup.~


jameslamb:
Even cutting over to `conda-forge`, I'm still seeing one job fail with a timeout similar to #4948

`Linux sdist` ([build](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=12020&view=logs&j=02a2c3ba-81f8-54e3-0767-5d5adbb0daa9&t=720ee3fa-96d4-5b47-dbf4-01607b74ade2))

@StrikerRUS could you look at those logs and the changes I've made in this PR so far and let me know if you have any ideas for fixes to investigate? If we can't think of anything else, then next I'll try setting `MKL_THREADING_LAYER=GNU` to see if it helps (https://github.com/microsoft/LightGBM/issues/4948#issuecomment-1013585404)

StrikerRUS:
@jameslamb Thanks for your efforts!

#4948 is affecting only `Linux_latest` jobs family. `Linux sdist` is run inside [our Ubuntu-14 Docker](https://github.com/guolinke/lightgbm-ci-docker) and compiles LightGBM via gcc
https://github.com/microsoft/LightGBM/blob/4aaeb22932b72aaff632b40ba596ca2533071ca2/.vsts-ci.yml#L28

So this is another problem. According to the logs, 

```
platform linux -- Python 3.7.12[pypy-7.3.7-final], pytest-6.2.5, py-1.11.0, pluggy-1.0.0
rootdir: /__w/1/s
collected 647 items

../tests/python_package_test/test_basic.py F............................ [  4%]
F..FFFFFF...................                                             [  8%]
../tests/python_package_test/test_consistency.py FFFFFF                  [  9%]
../tests/python_package_test/test_dask.py 
```

a lot of tests fail and the job hangs during trying to run the first Dask test.

A guess the reason is here:
```
platform linux -- Python 3.7.12[pypy-7.3.7-final], pytest-6.2.5, py-1.11.0, pluggy-1.0.0
```

Somehow pypy implementation `Python 3.7.12[pypy-7.3.7-final]` was configured to run tests.

jameslamb:
> Somehow pypy implementation Python 3.7.12[pypy-7.3.7-final] was configured to run tests.

Ah! weird! Ok thank you for the help, I'll investigate that.

.ci/test.sh
```diff
@@ -114,8 +128,23 @@ if [[ $TASK == "swig" ]]; then
...
+    numpy \
+    pandas \
+    psutil \
+    pytest \
+    scikit-learn \
+    scipy || exit -1
+
+# python-graphviz has to be installed separately to prevent conda from downgrading to pypy
+${CONDA_INSTALL} -q -y -n $CONDA_ENV \
+    python-graphviz || exit -1
```
jameslamb, line 147: Only for the specific of case of `Linux sdist` (Python 3.7, ubuntu:14.04, gcc), I found that `conda` was downgrading Python from a cpython-based Python 3.7.12 to a PyPy-based Python 3.7.12: https://github.com/microsoft/LightGBM/pull/4953#issuecomment-1014707402

Was able to reproduce this locally in Docker.

```shell
docker run \
    --rm \
    --env CONDA=/opt/conda \
    --entrypoint="" \
    -it lightgbm/vsts-agent:ubuntu-14.04 \
    /bin/bash

# then run the following in the container
export PATH=${CONDA}/bin:${PATH}

conda create \
    --name test-env \
    -c conda-forge \
    --override-channels \
        python=3.7

conda config --set always_yes yes --set changeps1 no
conda update -q -y conda

source activate test-env
conda install -c conda-forge --override-channels -q -y -n test-env \
    cloudpickle \
    dask \
    distributed \
    joblib \
    matplotlib \
    numpy \
    pandas \
    psutil \
    python-graphviz \
    pytest \
    scikit-learn \
    scipy
```

This resulted in the following from the `conda install` logs

```text
The following packages will be DOWNGRADED:

  python                        3.7.12-hf930737_100_cpython --> 3.7.12-0_73_pypy
  python_abi                                    3.7-2_cp37m --> 3.7-2_pypy37_pp73
  setuptools                          60.5.0-py37h89c1867_0 --> 59.8.0-py37h9c2f6ca_0
```

So I started thinking "ok, what is changing in this PR? One thing I'm proposing changing in this PR is installing `python-graphviz` with conda instead of as a separate `pip install`, to reduce the risk of conflicts.

Funny enough, it seems like that change was the cause for `conda`  choosing PyPy. Using the sample code above, I found that installing `python-graphviz` by itself, separately, does not result in `conda` choosing to downgrade from a cpython-based Python to the PyPy one.

.ci/test.sh
```diff
@@ -29,13 +29,20 @@ if [[ "$TASK" == "cpp-tests" ]]; then
     exit 0
 fi
 
-conda create -q -y -n $CONDA_ENV python=$PYTHON_VERSION
+# on Ubuntu 14.04, gpu_source build segfaults with gcc 10.x or 11.x
+INITIAL_CONDA_CONSTRAINTS="python=${PYTHON_VERSION}"
+if [[ $TASK == "gpu" ]] && [[ $OS_NAME == "linux" ]] && [[ $IN_UBUNTU_LATEST_CONTAINER != "true" ]]; then
+    INITIAL_CONDA_CONSTRAINTS="${INITIAL_CONDA_CONSTRAINTS} libgcc-ng=9.3.0 libgomp=9.3.0"
+fi
```
jameslamb, line 36: I found the builds for `Linux gpu_source` segfault with newer versions of `libgcc-ng` and `libgomp` installed.

With Anaconda default channels, we've been getting v9.3.0 of those dependencies. See, for example, the most recent build on `master` ([build link](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=12029&view=logs&j=9ce42e5d-f31a-544a-a6de-dc42f8c013ed&t=b696def5-78ae-590c-4234-04332342ba2d)).

```text
The following NEW packages will be INSTALLED:

  ...
  libgcc-ng          pkgs/main/linux-64::libgcc-ng-9.3.0-h5101ec6_17
  libgomp            pkgs/main/linux-64::libgomp-9.3.0-h5101ec6_17
```

Switching to `conda-forge`, `mamba` / `conda` (I tried both) really wanted to install much newer versions of these dependencies. I tried pinning to `<11.0` ([link](https://github.com/microsoft/LightGBM/pull/4953/commits/2455e0d2c06807d0d31435e74d3dc199be2cc951)), and `<10.0` ([link](https://github.com/microsoft/LightGBM/pull/4953/commits/1cac6f8417a190efb41e0e261ff0d3a58d85cf42))...both still resulted in a segmentation fault in the tests in `tests/c_api_test/test_.py`.

I'm not sure what is happening here, but I think we just haven't seen this problem before because Anaconda default channels is giving us such an old version of `libgcc` and `libgomp` for the `Linux gpu_source` job.

.ci/test.sh
```diff
@@ -29,13 +29,20 @@ if [[ "$TASK" == "cpp-tests" ]]; then
...
+
+mamba create -q -y -n $CONDA_ENV ${INITIAL_CONDA_CONSTRAINTS}
 source activate $CONDA_ENV
 
 cd $BUILD_DIRECTORY
 
 if [[ $TASK == "check-docs" ]] || [[ $TASK == "check-links" ]]; then
     cd $BUILD_DIRECTORY/docs
+    # NOTE: this job uses conda instead of mamba for consistency with RTD builds
```
jameslamb, line 45: planning to address this in a separate PR for #4954

jameslamb:

.ci/test.sh
```diff
@@ -114,8 +119,23 @@ if [[ $TASK == "swig" ]]; then
     exit 0
 fi
 
-conda install -q -y -n $CONDA_ENV cloudpickle "dask=2021.9.1" "distributed=2021.9.1" joblib matplotlib numpy pandas psutil pytest scikit-learn scipy
-pip install graphviz  # python-graphviz from Anaconda is not allowed to be installed with Python 3.9
+
+mamba install -q -y -n $CONDA_ENV \
+    cloudpickle \
+    dask \
+    distributed \
```
jmoralez, line 126: maybe we could leave these just as `dask`? Since `dask` already [includes](https://github.com/conda-forge/dask-feedstock/blob/4d652a702b4a42155a89ed599e7dccbad8821f6e/recipe/meta.yaml#L21-L22) `dask-core` and `distributed`, and `dask-core` [includes](https://github.com/conda-forge/dask-core-feedstock/blob/8b195b724f280ffe795a597e8aaabd2a4a1530bd/recipe/meta.yaml#L24) `cloudpickle`

.ci/test.sh
StrikerRUS: Ah, I see. It's a pity!
Could you please also check that this isn't elated to the used Python version?

StrikerRUS:
Could you please trigger a test CI run with swapped compilers (`clang`->`gcc` and `gcc`->`clang`)? Small example: bc2f8ce3229d623541b4102651587c3a5dfecef7.

.ci/test.sh
jameslamb: On that recent `master` build I linked, I see we got:

```text
pkgs/main/linux-64::python-3.9.7-h12debd9_1
```

On one of the builds where `Linux gpu_source` segfaulted, I see

```text
python             conda-forge/linux-64::python-3.9.9-h543edf9_0_cpython
python_abi         conda-forge/linux-64::python_abi-3.9-2_cp39
```

I'm not sure if the presence of `python_abi` is significant or not. I can try researching this a bit more in an hour or so.

jameslamb: Interesting, on the most recent build of this PR (which passed), I see `python_abi` was installed but we got `python` 3.9.6.

([build link](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=12031&view=logs&j=9ce42e5d-f31a-544a-a6de-dc42f8c013ed&t=b696def5-78ae-590c-4234-04332342ba2d))

```text
  + python                3.9.6  h49503c6_1_cpython  conda-forge/linux-64      27 MB
  + python_abi              3.9  2_cp39              conda-forge/linux-64       4 KB
```

I'll try a build that pins `python` to 3.9.7 but doesn't pin `libgcc` and `libgomp`.

jameslamb: Interesting! Pinning to `python=3.9.6` and not pinning anything else for `Linux gpu_source`, the job succeeded!

([build link](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=12040&view=logs&j=9ce42e5d-f31a-544a-a6de-dc42f8c013ed&t=b696def5-78ae-590c-4234-04332342ba2d))

I then tried just removing any special conditions in https://github.com/microsoft/LightGBM/pull/4953/commits/30b5e3972a235b464771f0050943ce46d7ef6176, and the `Linux gpu_source` build failed with a segmentation fault ([build link](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=12041&view=logs&j=9ce42e5d-f31a-544a-a6de-dc42f8c013ed&t=b696def5-78ae-590c-4234-04332342ba2d)).

I compared the logs of the two jobs using PyCharm's file differ. The table below has the full set of differences in dependencies managed by `mamba`.

| package         | successful build           | failed build               |
|:---------------:|:--------------------------:|:--------------------------:|
|bzip2            | --                         | 1.0.8 (h7f98852_4)         |
|dbus             | 1.13.6 (h48d8840_2)        | 1.13.6 (h5008d03_3)        |
|gettext          | 0.19.8.1 (h0b5b191_1005)   | 0.19.8.1 (h73d1719_1008)   |
|glib             | 2.68.4 (h9c3ff4c_0)        | --                         |
|glib-tools       | 2.68.4 (h9c3ff4c_0)        | --                         |
|graphviz         | 2.49.1 (h85b4f2f_0)        | 2.50.0 (h85b4f2f_1)        |
|gst-plugins-base | 1.18.5 (hf529b03_0)        | 1.18.5 (hf529b03_3)        |
|gstreamer        | 1.18.5 (h76c114f_0)        | 1.18.5 (h9f60fe5_3)        |
|harfbuzz         | 3.0.0 (h83ec7ef_1)         | 3.1.1 (h83ec7ef_0)         |
|libglib          | 2.68.4 (h3e27bee_0)        | 2.70.2 (h174f98d_1)        |
|libffi           | 3.3 (h58526e2_2)           | 3.4.2 (h7f98852_5)         |
|libnsl           | --                         | 2.0.0 (h7f98852_0)         |
|librsvg          | 2.52.2 (hc3c00ef_0)        | 2.52.5 (hc3c00ef_1)        |
|python           | 3.9.6 (h49503c6_1_cpython) | 3.9.9 (h543edf9_0_cpython) |

I also noticed that `libstdcxx-ng` ends up being the same version in both builds, but that it was available as of `mamba create` for the successful one and wasn't included until the later `mamba install` for the failed build.

So I tried just explicitly adding `libstdcxx-ng` as a dependency in `conda create`, in https://github.com/microsoft/LightGBM/pull/4953/commits/c9ba0b285e4b3d33db75acbe0837866791e8a8b4. Unfortunately, the `Linux gpu_source` build still segfaults with that change ([build link](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=12043&view=logs&j=9ce42e5d-f31a-544a-a6de-dc42f8c013ed&t=b696def5-78ae-590c-4234-04332342ba2d)).

So I think for now, to get this working and unblock CI for the rest of the project, we should just pin that job to python=3.9.6.

jameslamb:
> trigger a test CI run with swapped compilers

Pushed https://github.com/microsoft/LightGBM/pull/4953/commits/4bf9af4adf8cb5ba73ff367bd49acfd5f005b3ad swapping compilers. The only job that failed was `Linux gpu_source`...segfaulted when using `clang` ([Azure DevOps build link](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=12044&view=results)).

The diff in that commit is a bit difficult to read because it also involves uncommenting some commented-out jobs in CI config files, so I took screenshots of the important parts of the diff.

`.github/workflows/python_package.yml`



`.vsts-ci.yml`



Pushed another commit switching the compilers back: https://github.com/microsoft/LightGBM/pull/4953/commits/5808c2422caa3bd396c5b37194d8efdb8590469a

.ci/test.sh
StrikerRUS: Oh, sorry for being not clear enough! I meant check different **minor** Python versions, 3.7 and 3.8.

StrikerRUS:
> The only job that failed was `Linux gpu_source`...segfaulted when using clang

OK, this was expected: #3475. I'm glad that no more jobs fail!

> so I took screenshots of the important parts of the diff.

Thanks for doing this! `CUDA` jobs already check different compilers, so no need to swap compilers there.

.ci/test.sh
jameslamb: Ah ok. I just pushed a commit testing `Linux gpu_source` with several different versions of Python: https://github.com/microsoft/LightGBM/pull/4953/commits/f6f63cbb9b4a9cf138f3580ae4223a8acdd0e94a

See the results at https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=12051&view=results. This job succeeds with `PYTHON_VERSION` set to 3.7, 3.8, and 3.9.6. It segfaults with `PYTHON_VERSION` set to 3.9 and 3.10.

image

.ci/setup.sh
```diff
@@ -18,7 +18,7 @@ if [[ $OS_NAME == "macos" ]]; then
         brew install swig
     fi
     brew install graphviz
-    curl -sL -o conda.sh https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-x86_64.sh
+    curl -sL -o conda.sh https://github.com/conda-forge/miniforge/releases/latest/download/Mambaforge-MacOSX-x86_64.sh
```
StrikerRUS, line 21:
```suggestion
    curl -sL -o mambaforge.sh https://github.com/conda-forge/miniforge/releases/latest/download/Mambaforge-MacOSX-x86_64.sh
```

.ci/setup.sh
```diff
@@ -121,11 +121,10 @@ else  # Linux
...
     fi
     if [[ $SETUP_CONDA != "false" ]]; then
-        if [[ $ARCH == "x86_64" ]]; then
-            curl -sL -o conda.sh https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh
-        else
-            curl -sL -o conda.sh https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-Linux-${ARCH}.sh
-        fi
+        curl \
+            -sL \
+            -o conda.sh \
```
StrikerRUS, line 126:
```suggestion
            -o mambaforge.sh \
```
and here
https://github.com/microsoft/LightGBM/blob/e6a2f7162cf0bae4cbad6223c2700d74f0d67c4b/.ci/setup.sh#L134

.ci/test.sh
```diff
@@ -61,17 +62,15 @@ if [[ $TASK == "check-docs" ]] || [[ $TASK == "check-links" ]]; then
 fi
 
 if [[ $TASK == "lint" ]]; then
-    conda install -q -y -n $CONDA_ENV \
+    mamba install -q -y -n $CONDA_ENV \
+        cmakelint \
+        cpplint \
+        isort \
+        libxml2 \
```
StrikerRUS, line 69: Could you please check whether we still need explicitly install `libxml2`?
https://github.com/microsoft/LightGBM/pull/2954#issuecomment-605233486

StrikerRUS (commented):
@jameslamb 
WDYT about changing `${HOME}/miniconda` in the following lines to `${HOME}/mamba`, similarly to https://github.com/guolinke/lightgbm-ci-docker/pull/22#discussion_r786243926?
https://github.com/microsoft/LightGBM/blob/e6a2f7162cf0bae4cbad6223c2700d74f0d67c4b/.github/workflows/static_analysis.yml#L42
https://github.com/microsoft/LightGBM/blob/e6a2f7162cf0bae4cbad6223c2700d74f0d67c4b/.github/workflows/linkchecker.yml#L30
https://github.com/microsoft/LightGBM/blob/e6a2f7162cf0bae4cbad6223c2700d74f0d67c4b/.github/workflows/cuda.yml#L84
https://github.com/microsoft/LightGBM/blob/e6a2f7162cf0bae4cbad6223c2700d74f0d67c4b/.github/workflows/python_package.yml#L69
https://github.com/microsoft/LightGBM/blob/e6a2f7162cf0bae4cbad6223c2700d74f0d67c4b/.vsts-ci.yml#L125
https://github.com/microsoft/LightGBM/blob/e6a2f7162cf0bae4cbad6223c2700d74f0d67c4b/.vsts-ci.yml#L186
https://github.com/microsoft/LightGBM/blob/e6a2f7162cf0bae4cbad6223c2700d74f0d67c4b/.vsts-ci.yml#L239

Also, I think it'll be good to migrate to `mamba` in Windows CI jobs as well. In a follow-up PR, of course. Do you agree?

Not sure, but shouldn't we change this
https://github.com/microsoft/LightGBM/blob/e6a2f7162cf0bae4cbad6223c2700d74f0d67c4b/.ci/setup.sh#L136-L137
to 
```
mamba config --set always_yes yes --set changeps1 no 
mamba update -q -y mamba
```
? This should be especially useful for `Linux` CI jobs that are run into our infrequently updated Docker.

In addition, could you please start trying to get 5 consecutive successful CI builds to make syre these changes don't make any jobs flaky?

jameslamb:
> WDYT about changing ${HOME}/miniconda in the following lines to ${HOME}/mamba, similarly to guolinke/lightgbm-ci-docker#22 (comment)?

Sure, I can do that

<hr>

> but shouldn't we change this...

There is no `mamba config` or `mamba update`. If you try running one of those commands, you'll see the following error message

> Currently, only install, create, list, search, run, info and clean are supported through mamba.

That comes from here: https://github.com/mamba-org/mamba/blob/234f2511808d6b3dcc843e6fb1bf7d7302fbeb31/mamba/mamba/mamba.py#L746-L749.

So the `conda config` line should be left as-is. I think it will still affect `mamba` correctly. The other line should be changed to `conda update -y mamba`, good catch!

<hr>

> Also, I think it'll be good to migrate to mamba in Windows CI jobs as well. In a follow-up PR, of course. Do you agree?

Yes, I agree. I focused on Linux and macOS first because #4948 is `blocking` and I wanted to try to resolve that as fast as possible.

.ci/test.sh
jameslamb: oh yeah, nice! I'll be very happy if this PR is able to remove this and other installation hacks 🙌 

tried removing it in https://github.com/microsoft/LightGBM/pull/4953/commits/bc750fbd3523bd6250f35fd4538c0a94c8c03f4b, let's see how it goes.

.ci/setup.sh
jameslamb: updated this in https://github.com/microsoft/LightGBM/pull/4953/commits/bc750fbd3523bd6250f35fd4538c0a94c8c03f4b instead of accepting in the browser, to batch all changes together and avoid duplicate CI runs.

.ci/setup.sh
jameslamb: updated this in https://github.com/microsoft/LightGBM/pull/4953/commits/bc750fbd3523bd6250f35fd4538c0a94c8c03f4b instead of accepting in the browser, to batch all changes together and avoid duplicate CI runs.

.ci/setup.sh
```diff
@@ -121,18 +124,17 @@ else  # Linux
...
-        fi
+        curl \
+            -sL \
+            -o mambaforge.sh \
+            https://github.com/conda-forge/miniforge/releases/latest/download/Mambaforge-Linux-${ARCH}.sh
     fi
 fi
 
-if [[ "${TASK}" != "r-package" ]]; then
+if [[ "${TASK}" != "r-package" ]] && [[ "${TASK}" != "r-rchk" ]]; then
```
jameslamb, line 134: really interesting! I guess we've been accidentally installing conda for the `r-rchk` job, but this never failed before because the GitHub Actions environments have an installation of `conda` already?

See https://github.com/microsoft/LightGBM/runs/4909427380?check_suite_focus=

> ERROR: File or directory already exists: '/usr/share/miniconda'
If you want to update an existing installation, use the -u option.

PackageNotInstalledError: Package is not installed in prefix.
  prefix: /usr/share/miniconda
  package name: mamba

See https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu1804-Readme.md#environment-variables....by default, GitHub Actions sets `CONDA=/usr/share/miniconda` in Linux environments.

jameslamb: looks like the fix I added in https://github.com/microsoft/LightGBM/pull/4953/commits/44de0babe86aaeb0f94995725f575fbe34a22ed6 worked: https://github.com/microsoft/LightGBM/runs/4909444516?check_suite_focus=true

.ci/test.sh
jameslamb: all linting tasks passed! https://github.com/microsoft/LightGBM/runs/4909444083?check_suite_focus=true

And I don't see any issues in the logs. So looks like it's ok to remove this `libxml2` install.

jameslamb:
> could you please start trying to get 5 consecutive successful CI builds to make syre these changes don't make any jobs flaky?

Sure! I'll push some empty commits and keep updating this comment.

1. https://github.com/microsoft/LightGBM/pull/4953/commits/44de0babe86aaeb0f94995725f575fbe34a22ed6 ✅ 
2.
3.
4.
5.

StrikerRUS (approved):
Thanks a lot for this PR with a lot of investigations!

StrikerRUS:
Ah, `QEMU_multiarch bdist` timed out 😢 
https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=12076&view=logs&j=c2f9361f-3c13-57db-3206-cee89820d5e3&t=bf5b33f9-9072-549e-d8b1-cab79a1dd2a9

jameslamb:
> Ah, QEMU_multiarch bdist timed out 😢

😭 😭 😭 😭 😭 we were so close! I have no idea, from those logs, whether or not that timeout is related to this PR or just a random infrastructure-level failure.

But I can at least say that it seems to happening at the point where `python-graphviz` is separately installed.

https://github.com/microsoft/LightGBM/blob/5ab178905474db54677664e4cb0578ba08d9d1b1/.ci/test.sh#L129-L131

I'm going to try making the logs at that step more verbose and running 5 more commits, let's see if we can reproduce the issue.

StrikerRUS:
> I'm going to try making the logs at that step more verbose and running 5 more commits, let's see if we can reproduce the issue.

Reproduced with the first commit. Looks like installation process hangs during this command
https://github.com/microsoft/LightGBM/blob/537fbd4f55b2f3a6da95c3fbd22bdf7a82002204/.ci/test.sh#L116-L127
```
  Downgrade:
───────────────────────────────────────────────────────────────────────────────────────────────────

  - setuptools            60.5.0  py39ha65689a_0            installed                            
  + setuptools            59.8.0  py39ha65689a_0            conda-forge/linux-aarch64     1018 KB
```

But previous commit passed that step and timed out during `graphviz` installation
```
  Downgrade:
───────────────────────────────────────────────────────────────────────────────────────────────────

  - setuptools            60.5.0  py39ha65689a_0            installed                            
  + setuptools            59.8.0  py39ha65689a_0            conda-forge/linux-aarch64     1018 KB

  Summary:

  Install: 73 packages
  Downgrade: 1 packages

  Total download: 152 MB

───────────────────────────────────────────────────────────────────────────────────────────────────

Preparing transaction: ...working... done
Verifying transaction: ...working... done
Executing transaction: ...working... done
```

StrikerRUS:
I checked durations of `QEMU` jobs for all previous empty testing commits and they all are fine (except the timed out last one). They took ~40-70 mins which is comparable with `master` branch. So we cannot say about general slowness of QEMU jobs after switching to mamba (like we observed with Dask tests in #4552).

StrikerRUS:
[`mamba 0.20.0`](https://github.com/mamba-org/mamba/releases/tag/2022.01.25) has been released recently. Maybe we can retry with new version?

jameslamb:
ah very nice! Ok yeah @StrikerRUS I'll try pushing a new set of commits (and updating this comment). Since we run `conda update -y mamba`, that should pull in the newest version.

StrikerRUS:
Nope, still times out...

jameslamb:
blegh. I looked through the issues in the `mamba` repo and don't see anything obviously related.

I'm going to test another theory. I just pushed a commit that prints the output of `mamba info`. I wonder if `mamba` is trying to write to a path that's mounted into the container

https://github.com/microsoft/LightGBM/blob/0075814f02f898ae4070d848078b9d0d4b510ceb/.vsts-ci.yml#L199-L200

I've seen that before using docker, that heavy I/O (writing/reading/deleting lots of small files) on a `:rw`-mounted volume can sometimes cause operations to be really slow. I want to see if maybe `mamba` is writing downloaded package files (for example) to a path that's mounted in, and if that is causing the slowness.

jameslamb:
> I want to see if maybe mamba is writing downloaded package files (for example) to a path that's mounted in

I see the following in `mamba info` ([build link](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=12124&view=logs&j=c2f9361f-3c13-57db-3206-cee89820d5e3&t=bf5b33f9-9072-549e-d8b1-cab79a1dd2a9))

```text
     active environment : None
       user config file : /root/.condarc
 populated config files : /root/mamba/.condarc
                          /root/.condarc
          conda version : 4.11.0
    conda-build version : not installed
         python version : 3.9.7.final.0
       virtual packages : __linux=5.11.0=0
                          __glibc=2.17=0
                          __unix=0=0
                          __archspec=1=aarch64
       base environment : /root/mamba  (writable)
      conda av data dir : /root/mamba/etc/conda
  conda av metadata url : None
           channel URLs : https://conda.anaconda.org/conda-forge/linux-aarch64
                          https://conda.anaconda.org/conda-forge/noarch
          package cache : /root/mamba/pkgs
                          /root/.conda/pkgs
       envs directories : /root/mamba/envs
                          /root/.conda/envs
               platform : linux-aarch64
             user-agent : conda/4.11.0 requests/2.27.1 CPython/3.9.7 Linux/5.11.0-1027-azure centos/7.9.2009 glibc/2.17
                UID:GID : 0:0
             netrc file : None
           offline mode : False
```

None of that looks obviously problematic to me.  The two volumes mounted in are at `/LightGBM` and `/home/vsts/work/1/a` in the container's filesystem.

jameslamb:
In more verbose build logs ([link](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=12128&view=logs&j=c2f9361f-3c13-57db-3206-cee89820d5e3&t=bf5b33f9-9072-549e-d8b1-cab79a1dd2a9&l=543)), I see errors like the following:

```text
WARNING: md5sum mismatch of tar archive
expected: 752675fba7a7da236dd1d1fe1f052337
     got: bffa730cc0732d7c7919549e7eb6eda2  -
Unpacking payload ...
ERROR   Truncated tar archive
...
Extracting "libiconv-1.16-h6dd45c4_0.tar.bz2"
ERROR   Error opening archive: truncated bzip2 input
ERROR   /root/mamba/pkgs/libiconv-1.16-h6dd45c4_0.tar.bz2: Could not open archive for reading.
```

So I wonder if part of the issue related to `mamba` doing something weird with the cache directory? For example, maybe it is timing out trying to read in files that have been corrupted by its writes.

I see some possibly-related issues in the `mamba` repo related to handling of the cache and the behavior of `mamba` when updating an existing environment.

* https://github.com/mamba-org/mamba/issues/852
* https://github.com/mamba-org/mamba/issues/1413
* https://github.com/mamba-org/mamba/issues/1402

jameslamb:
> To test this, I pushed changes (c82fe21) which run mamba clean after every step that installs packages.

That worked! QEMU build succeeded in 48m5s: https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=12129&view=logs&j=c2f9361f-3c13-57db-3206-cee89820d5e3&t=bf5b33f9-9072-549e-d8b1-cab79a1dd2a9.

Next, I'll start undoing the debugging changes I made to try to narrow it down to only the changes necessary to make these builds work.

jameslamb:
I just merged in #4989 with the fixes for R-package builds.

Next, I'm going to test whether "something related to the `mamba` cache" is to blame for the segfaults I observed in the `gpu_source` build (https://github.com/microsoft/LightGBM/pull/4953#discussion_r786357266). For example, maybe it led to linked libraries getting into a weird state.

jameslamb:
commits (I hope this works this time 😬 )

1. https://github.com/microsoft/LightGBM/pull/4953/commits/986dad6683c845f83bc546264c5a3fe76d6c526e
2.
3.
4.
5.

StrikerRUS:
Unfortunately, QEMU job timed out again at the same point 😢 

My suggestion is to stop investigation for now, create a feature request for migrating from conda to mamba and replace mambaforge with miniforge in this PR (this should be enough to fix original issue of libomp conflicts).

jameslamb:
> migrating from conda to mamba and replace mambaforge with miniforge in this PR

😫 yeah I agree with you. Will start pushing commits testing that. Based on my understanding of the README at https://github.com/conda-forge/miniforge, I think the `conda` executable you get with the Mambaforge installer is the same as what you'd get using a non-`mamba`  miniforge installer...so I believe just changing `mamba`  to `conda` in this PR will be enough and changes to the ubuntu:14.04 image won't be necessary.

StrikerRUS:
> Based on my understanding of the README at https://github.com/conda-forge/miniforge, I think the `conda` executable you get with the Mambaforge installer is the same as what you'd get using a non-`mamba`  miniforge installer...so I believe just changing `mamba`  to `conda` in this PR will be enough and changes to the ubuntu:14.04 image won't be necessary.

Probably you are right, but let's not leave confusing moments without any direct reasons for that. Also, identity between installers can be changed at any time in the future. I'll push necessary changes into `dev` branch of our CI Docker repo.

.ci/setup.sh
```diff
@@ -18,7 +18,10 @@ if [[ $OS_NAME == "macos" ]]; then
         brew install swig
     fi
     brew install graphviz
-    curl -sL -o conda.sh https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-x86_64.sh
+    curl \
+        -sL \
+        -o mambaforge.sh \
+        https://github.com/conda-forge/miniforge/releases/latest/download/Mambaforge-MacOSX-x86_64.sh
```
StrikerRUS, line 24:
```suggestion
    curl \
        -sL \
        -o miniforge.sh \
        https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-MacOSX-x86_64.sh
```

.ci/setup.sh
```diff
@@ -115,17 +118,16 @@ else  # Linux
...
     if [[ $SETUP_CONDA != "false" ]]; then
-        if [[ $ARCH == "x86_64" ]]; then
-            curl -sL -o conda.sh https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh
-        else
-            curl -sL -o conda.sh https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-Linux-${ARCH}.sh
-        fi
+        curl \
+            -sL \
+            -o mambaforge.sh \
+            https://github.com/conda-forge/miniforge/releases/latest/download/Mambaforge-Linux-${ARCH}.sh
```
StrikerRUS, line 124:
```suggestion
        curl \
            -sL \
            -o miniforge.sh \
            https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-Linux-${ARCH}.sh
```

.ci/setup.sh
```diff
@@ -115,17 +118,16 @@ else  # Linux
...
     fi
 fi
 
-if [[ "${TASK}" != "r-package" ]]; then
+if [[ "${TASK}" != "r-package" ]] && [[ "${TASK}" != "r-rchk" ]]; then
     if [[ $SETUP_CONDA != "false" ]]; then
-        sh conda.sh -b -p $CONDA
+        sh mambaforge.sh -b -p $CONDA
```
StrikerRUS, line 130:
```suggestion
        sh miniforge.sh -b -p $CONDA
```

.ci/test.sh
```diff
@@ -29,7 +29,7 @@ if [[ "$TASK" == "cpp-tests" ]]; then
     exit 0
 fi
 
-conda create -q -y -n $CONDA_ENV python=$PYTHON_VERSION
+conda create -q -y -n $CONDA_ENV python=${PYTHON_VERSION}
```
StrikerRUS, line 32: To simplify the diff by removing this unrelated line.
```suggestion
conda create -q -y -n $CONDA_ENV python=$PYTHON_VERSION
```

.ci/test.sh
```diff
@@ -114,8 +111,22 @@ if [[ $TASK == "swig" ]]; then
...
+    numpy \
+    pandas \
+    psutil \
+    pytest \
+    scikit-learn \
+    scipy || exit -1
+
+# python-graphviz has to be installed separately to prevent conda from downgrading to pypy
+conda install -q -y -n $CONDA_ENV \
+    python-graphviz || exit -1
```
StrikerRUS, line 129: Let's try to simplify with conda now. Maybe that issue was related only to mamba.
```suggestion
    psutil \
    pytest \
    python-graphviz \
    scikit-learn \
    scipy || exit -1
```

.github/workflows/cuda.yml
```diff
@@ -81,7 +81,7 @@ jobs:
             LGB_VER=$(head -n 1 VERSION.txt)
             EOF
             cat > docker-script.sh <<EOF
-            export CONDA=\$HOME/miniconda
+            export CONDA=\$HOME/mambaforge
```
StrikerRUS, line 84:
```suggestion
            export CONDA=\$HOME/miniforge
```

.github/workflows/linkchecker.yml
```diff
@@ -27,7 +27,7 @@ jobs:
       - name: Setup and run tests
         run: |
           export BUILD_DIRECTORY="$GITHUB_WORKSPACE"
-          export CONDA=${HOME}/miniconda
+          export CONDA=${HOME}/mambaforge
```
StrikerRUS, line 30:
```suggestion
          export CONDA=${HOME}/miniforge
```

.github/workflows/python_package.yml
```diff
@@ -66,7 +66,7 @@ jobs:
           fi
           export BUILD_DIRECTORY="$GITHUB_WORKSPACE"
           export LGB_VER=$(head -n 1 VERSION.txt)
-          export CONDA=${HOME}/miniconda
+          export CONDA=${HOME}/mambaforge
```
StrikerRUS, line 69:
```suggestion
          export CONDA=${HOME}/miniforge
```

.github/workflows/static_analysis.yml
```diff
@@ -39,7 +39,7 @@ jobs:
         run: |
           export TASK="${{ matrix.task }}"
           export BUILD_DIRECTORY="$GITHUB_WORKSPACE"
-          export CONDA=${HOME}/miniconda
+          export CONDA=${HOME}/mambaforge
```
StrikerRUS, line 42:
```suggestion
          export CONDA=${HOME}/miniforge
```

.vsts-ci.yml
```diff
@@ -122,7 +124,7 @@ jobs:
   - script: |
       echo "##vso[task.setvariable variable=BUILD_DIRECTORY]$BUILD_SOURCESDIRECTORY"
       echo "##vso[task.setvariable variable=LGB_VER]$(head -n 1 VERSION.txt)"
-      CONDA=$HOME/miniconda
+      CONDA=$HOME/mambaforge
```
StrikerRUS, line 127:
```suggestion
      CONDA=$HOME/miniforge
```

.vsts-ci.yml
```diff
@@ -183,7 +185,7 @@ jobs:
       BUILD_ARTIFACTSTAGINGDIRECTORY=$BUILD_ARTIFACTSTAGINGDIRECTORY
       EOF
       cat > docker-script.sh <<EOF
-      export CONDA=\$HOME/miniconda
+      export CONDA=\$HOME/mambaforge
```
StrikerRUS, line 188:
```suggestion
      export CONDA=\$HOME/miniforge
```

.vsts-ci.yml
```diff
@@ -236,7 +238,7 @@ jobs:
   - script: |
       echo "##vso[task.setvariable variable=BUILD_DIRECTORY]$BUILD_SOURCESDIRECTORY"
       echo "##vso[task.setvariable variable=LGB_VER]$(head -n 1 VERSION.txt)"
-      CONDA=$AGENT_HOMEDIRECTORY/miniconda
+      CONDA=$AGENT_HOMEDIRECTORY/mambaforge
```
StrikerRUS, line 241:
```suggestion
      CONDA=$AGENT_HOMEDIRECTORY/miniforge
```

.vsts-ci.yml
```diff
@@ -49,6 +49,8 @@ jobs:
       gpu_source:
         TASK: gpu
         METHOD: source
+        # on Ubuntu 14.04, gpu_source build segfaults with Python 3.9.9
+        PYTHON_VERSION: 3.9.6
```
StrikerRUS, line 53: Let's try to simplify with conda now. Maybe that issue was related only to mamba.
```suggestion
```

StrikerRUS (changes requested):

StrikerRUS:
> I'll push necessary changes into dev branch of our CI Docker repo.

Done: https://github.com/guolinke/lightgbm-ci-docker/pull/23#issuecomment-1030338339.

.vsts-ci.yml
jameslamb: fine with me. You can directly push any changes you want to this PR.

.ci/test.sh
StrikerRUS: Doesn't work. Reverted in 5e033adc726e8ceb34b35d731b019698045afd16.

.vsts-ci.yml
StrikerRUS: Doesn't work. Reverted in 5e033adc726e8ceb34b35d731b019698045afd16.

StrikerRUS:
Well, starting yet another round of empty commits for stability testing.

1.

jameslamb:
The changes you made all look good to me, @StrikerRUS . Thanks for doing that!

I forgot about the `libxau` stuff that was necessary for graphviz, really appreciate you removing that. And I'm glad that R jobs will no longer `brew install graphviz`.

If the empty commits all build successfully and you're confident in the changes, it's ok with me if you approve and merge this PR. Just noting that since we're in different timezones and the 5th empty commit might happen while I'm asleep 😄

StrikerRUS:
> ... and the 5th empty commit  ...

I'd like to push 10 empty commits for better confidence. After that I'll switch compilers and push 5 more empty commits. I hope to finish this by the end of this weekend.

StrikerRUS:
Huh, just noticed the following strange thing in the latest `master`'s QEMU job.
https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=12134&view=logs&j=c2f9361f-3c13-57db-3206-cee89820d5e3&t=bf5b33f9-9072-549e-d8b1-cab79a1dd2a9&l=565

Looks like `mamba` is used somewhere around here
https://github.com/microsoft/LightGBM/blob/0688f471fc303af816cd6564107f6d720ebb99ec/.ci/setup.sh#L127-L130

image
image
image



But the next line in our ci script is performed with conda
https://github.com/microsoft/LightGBM/blob/0688f471fc303af816cd6564107f6d720ebb99ec/.ci/setup.sh#L131
image

StrikerRUS:
@jmoralez Would you like to take a fresh look?

StrikerRUS:
OK, GitHub is lagging.

image

Feel free to continue [my plan](https://github.com/microsoft/LightGBM/pull/4953#issuecomment-1030611608) while I'm sleeping.

StrikerRUS:
I can see the only difference between successful and the most recemt failing Linux GPU jobs:
```
    scipy-1.7.3                |   py39hee8e79c_0        22.0 MB  conda-forge

    scipy-1.8.0                |   py39hee8e79c_0        25.4 MB  conda-forge
```

StrikerRUS:
@jameslamb @jmoralez Please check the following results: https://github.com/microsoft/LightGBM/pull/4953#issuecomment-1030468356.

StrikerRUS (approved):

StrikerRUS:
@jameslamb @jmoralez I'd like to push this blocking PR forward. Could you please take a look when have time?

jameslamb:
> Could you please take a look when have time?

Sorry, didn't realize you were waiting on me. Changes all look fine to me, ok with me if we merge them.

jmoralez (approved):

Issue #4859, opened 2021-12-06
[docs] [R-package] upgrade R documentation to {pkgdown} 2.0


jameslamb:
## Summary

This project uses `{pkgdown}` to generate a documentation website for the R package (https://lightgbm.readthedocs.io/en/latest/R/reference/).

A new major release of `{pkgdown}`, v2.0, was recently published to CRAN ([link](https://cran.r-project.org/web/packages/pkgdown/index.html)) and to conda-forge ([link](https://anaconda.org/conda-forge/r-pkgdown)).

This project's documentation should be updated to use that new version of `{pkgdown}`.

## Motivation

This release brings some significant new features, detailed at https://www.tidyverse.org/blog/2021/12/pkgdown-2-0-0.

I think the following would be useful for `{lightgbm}`:

* ability to specify redirects when pages are removed (extra important now that progress is being made on #1944)
* static site search with no external dependencies

## Description

To implement this, update the `conda` environment used in readthedocs ("RTD") builds so that it uses `pkgdown>=2.0.0`.

https://github.com/microsoft/LightGBM/blob/f7e39388d0327e687854f22d7e06aec302770994/docs/conf.py#L262-L275

And possibly make changes to the configuration in https://github.com/microsoft/LightGBM/tree/f7e39388d0327e687854f22d7e06aec302770994/R-package/pkgdown.

## References

`{pkgdown}` repo: https://github.com/r-lib/pkgdown

Vignette explaining how to add search to a `{pkgdown}` site: https://cran.r-project.org/web/packages/pkgdown/vignettes/search.html

jameslamb:
Per this project's policy, I've added this feature request to #2302. Anyone is welcome to contribute this! If you're interested, please comment here. The issue can be re-opened and maintainers will be happy to answer any questions you have.

jameslamb:
Sorry, this was locked accidentally. Just unlocked it. We'd still love help with this feature!

PR #4957, opened 2022-01-18
[ci] [docs] use miniforge for readthedocs builds (fixes #4954)

### LINKED ISSUE #4954: [ci] [docs] readthedocs configuration options are deprecated
## Description

This project uses readthedocs to build and host its documentation at https://lightgbm.readthedocs.io/en/latest/.

The way those docs are built is configured here: https://github.com/microsoft/LightGBM/blob/4aaeb22932b72aaff632b40ba596ca2533071ca2/.readthedocs.yaml#L1-L14

According to the documentation for that service, some of the configuration options used in this project's `.readthedocs.yaml` are deprecated.

*python.version* ([link](https://docs.readthedocs.io/en/stable/config-file/v2.html#python-version))

> This option is now deprecated and replaced by [build.tools.python](https://docs.readthedocs.io/en/stable/config-file/v2.html#build-tools-python). See [python.version (legacy)](https://docs.readthedocs.io/en/stable/config-file/v2.html#python-version-legacy) for the description of this option.

According to https://docs.readthedocs.io/en/stable/config-file/v2.html#python-version-legacy, using `version: 3` is also deprecated.

> Make sure to use quotes `(")` to make it a string. We previously supported using numbers here, but that approach is deprecated.

## Additional Comments

These deprecated configuration should be removed, so that documentation builds don't break when those features are removed in future changes to the readthedocs service.

Given the discussion in #4948, I think this should be accomplished by switching the project's readthedocs builds to use `mamba`, as documented at "Making builds faster with mamba" ([RTD docs link](https://docs.readthedocs.io/en/stable/guides/conda.html#making-builds-faster-with-mamba)).

jameslamb:
Fixes #4954.
Contributes to #4948.
Contributes to #4859.

This PR proposes updating LightGBM's readthedocs configuration to tell RTD to use `mamba` to manage dependencies. See the description of #4954 for relevant links to RTD documentation describing these new config options.

## Notes for Reviewers

This PR updates the `"check-docs"` CI job to use `conda`. I'm hoping this can be merged before #4953, and then #4953 could switch that CI check to use `mamba` so LightGBM's CI will more closely match what happens on `master` RTD builds.

@StrikerRUS could you please temporarily enable builds for this branch on RTD so we can test the new configuration?

### DIFF
--- a/docs/README.rst
+++ b/docs/README.rst
@@ -13,7 +13,7 @@ After each commit on ``master``, documentation is updated and published to `Read
 Build
 -----
 
-It is not necessary to re-build this documentatioon while modifying LightGBM's source code.
+It is not necessary to re-build this documentation while modifying LightGBM's source code.
 The HTML files generated using ``Sphinx`` are not checked into source control.
 However, you may want to build them locally during development to test changes.
 
@@ -28,6 +28,7 @@ Run the following from the root of this repository to pull the relevant image an
 
     docker run \
         --rm \
+        --user=0 \
         -v $(pwd):/opt/LightGBM \
         --env C_API=true \
         --env CONDA=/opt/conda \
@@ -35,9 +36,8 @@ Run the following from the root of this repository to pull the relevant image an
         --env READTHEDOCS=true \
         --workdir=/opt/LightGBM/docs \
         --entrypoint="" \
-        --user=0 \
         -it readthedocs/build:ubuntu-20.04-2021.09.23 \
-        /bin/bash --noprofile --norc build-docs.sh
+        /bin/bash build-docs.sh
 
 When that code completes, open ``docs/html/index.html`` in your browser.
 
--- a/docs/README.rst
+++ b/docs/README.rst
@@ -20,7 +20,7 @@ However, you may want to build them locally during development to test changes.
 Docker
 ^^^^^^
 
-The most reliable way to build the documentation locally is with Docker, using `the same images readthedocs uses <https://hub.docker.com/r/readthedocs/build>`_.
+The most reliable way to build the documentation locally is with Docker, using `the same images Read the Docs uses <https://hub.docker.com/r/readthedocs/build>`_.
 
 Run the following from the root of this repository to pull the relevant image and run a container locally.
 
@@ -51,7 +51,7 @@ You can build the documentation locally. Just install Doxygen and run in ``docs`
 
 .. code:: sh
 
-    pip install breathe sphinx 'sphinx_rtd_theme>0.5'
+    pip install breathe sphinx 'sphinx_rtd_theme>=0.5'
     make html
 
 Unfortunately, documentation for R code is built only on our site, and commands above will not build it for you locally.
@@ -61,6 +61,6 @@ If you faced any problems with Doxygen installation or you simply do not need do
 
 .. code:: sh
 
-    pip install sphinx 'sphinx_rtd_theme>0.5'
+    pip install sphinx 'sphinx_rtd_theme>=0.5'
     export C_API=NO || set C_API=NO
     make html
--- a/docs/README.rst
+++ b/docs/README.rst
@@ -38,7 +38,7 @@ Run the following from the root of this repository to pull the relevant image an
         -it readthedocs/build:ubuntu-20.04-2021.09.23 \
         /bin/bash build-docs.sh
 
-When that code completes, open ``docs/html/index.html`` in your browser.
+When that code completes, open ``docs/_build/html/index.html`` in your browser.
 
 .. note::
 
@@ -55,8 +55,8 @@ You can build the documentation locally without Docker. Just install Doxygen and
     make html
 
 Note that this will not build the R documentation.
-To build the R docuumentation locally, use the Docker-based approach described above.
 Consider using common R utilities for documentation generation, if you need it.
+Or use the Docker-based approach described above to build the R documentation locally.
 
 If you faced any problems with Doxygen installation or you simply do not need documentation for C code, it is possible to build the documentation without it:
 
--- a/docs/README.rst
+++ b/docs/README.rst
@@ -31,8 +31,7 @@ Run the following from the root of this repository to pull the relevant image an
         --user=0 \
         -v $(pwd):/opt/LightGBM \
         --env C_API=true \
-        --env CONDA=/opt/conda \
-        --env PYTHONUNBUFFERED=1 \
+        --env CONDA=/opt/miniforge \
         --env READTHEDOCS=true \
         --workdir=/opt/LightGBM/docs \
         --entrypoint="" \
--- a/docs/README.rst
+++ b/docs/README.rst
@@ -13,11 +13,46 @@ After each commit on ``master``, documentation is updated and published to `Read
 Build
 -----
 
+It is not necessary to re-build this documentatioon while modifying LightGBM's source code.
+The HTML files generated using ``Sphinx`` are not checked into source control.
+However, you may want to build them locally during development to test changes.
+
+Docker
+^^^^^^
+
+The most reliable way to build the documentation locally is with Docker, using `the same images readthedocs uses <https://hub.docker.com/r/readthedocs/build>`_.
+
+Run the following from the root of this repository to pull the relevant image and run a container locally.
+
+.. code:: sh
+
+    docker run \
+        --rm \
+        -v $(pwd):/opt/LightGBM \
+        --env C_API=true \
+        --env CONDA=/opt/conda \
+        --env PYTHONUNBUFFERED=1 \
+        --env READTHEDOCS=true \
+        --workdir=/opt/LightGBM/docs \
+        --entrypoint="" \
+        --user=0 \
+        -it readthedocs/build:ubuntu-20.04-2021.09.23 \
+        /bin/bash --noprofile --norc build-docs.sh
+
+When that code completes, open ``docs/html/index.html`` in your browser.
+
+.. note::
+
+    The navigation in locally-built docs does not link to the local copy of the R documentation. To view the local version of the R docs, open ``docs/_build/html/R/index.html`` in your browser.
+
+Locally
+^^^^^^^
+
 You can build the documentation locally. Just install Doxygen and run in ``docs`` folder
 
 .. code:: sh
 
-    pip install -r requirements.txt
+    pip install breathe sphinx 'sphinx_rtd_theme>0.5'
     make html
 
 Unfortunately, documentation for R code is built only on our site, and commands above will not build it for you locally.
--- a/docs/README.rst
+++ b/docs/README.rst
@@ -62,6 +62,6 @@ If you faced any problems with Doxygen installation or you simply do not need do
 
 .. code:: sh
 
-    pip install -r requirements_base.txt
+    pip install breathe sphinx 'sphinx_rtd_theme>0.5'
     export C_API=NO || set C_API=NO
     make html
--- a/docs/README.rst
+++ b/docs/README.rst
@@ -42,19 +42,20 @@ When that code completes, open ``docs/html/index.html`` in your browser.
 
 .. note::
 
-    The navigation in locally-built docs does not link to the local copy of the R documentation. To view the local version of the R docs, open ``docs/_build/html/R/index.html`` in your browser.
+    The navigation in these locally-built docs does not link to the local copy of the R documentation. To view the local version of the R docs, open ``docs/_build/html/R/index.html`` in your browser.
 
-Locally
-^^^^^^^
+Without Docker
+^^^^^^^^^^^^^^
 
-You can build the documentation locally. Just install Doxygen and run in ``docs`` folder
+You can build the documentation locally without Docker. Just install Doxygen and run in ``docs`` folder
 
 .. code:: sh
 
     pip install breathe sphinx 'sphinx_rtd_theme>=0.5'
     make html
 
-Unfortunately, documentation for R code is built only on our site, and commands above will not build it for you locally.
+Note that this will not build the R documentation.
+To build the R docuumentation locally, use the Docker-based approach described above.
 Consider using common R utilities for documentation generation, if you need it.
 
 If you faced any problems with Doxygen installation or you simply do not need documentation for C code, it is possible to build the documentation without it:
--- a/docs/README.rst
+++ b/docs/README.rst
@@ -62,6 +62,6 @@ If you faced any problems with Doxygen installation or you simply do not need do
 
 .. code:: sh
 
-    pip install breathe sphinx 'sphinx_rtd_theme>0.5'
+    pip install sphinx 'sphinx_rtd_theme>0.5'
     export C_API=NO || set C_API=NO
     make html
--- a/docs/build-docs.sh
+++ b/docs/build-docs.sh
@@ -2,15 +2,14 @@
 
 rm -f ./_FIRST_RUN.flag
 
-ARCH=$(uname -m)
 export PATH="${CONDA}/bin:${PATH}"
 
 curl \
     -L \
-    -o ${HOME}/conda.sh \
-    "https://github.com/conda-forge/miniforge/releases/latest/download/Mambaforge-Linux-${ARCH}.sh"
+    -o ${HOME}/miniforge.sh \
+    "https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-MacOSX-x86_64.sh"
 
-/bin/bash ${HOME}/conda.sh -b -p ${CONDA}
+/bin/bash ${HOME}/miniforge.sh -b -p ${CONDA}
 
 conda config --set always_yes yes --set changeps1 no
 
@@ -20,16 +19,6 @@ mamba create \
     --file env.yml || exit -1
 
 source activate docs-env
-
-python \
-    -m sphinx \
-    -T \
-    -E \
-    -W \
-    --keep-going \
-    -b html \
-    -d _build/doctrees \
-    -D language=en \
-    . _build/html || exit -1
+make clean html || exit -1
 
 echo "Done building docs. Open docs/_build/html/index.html in a web browser to view them."
--- a/docs/build-docs.sh
+++ b/docs/build-docs.sh
@@ -29,6 +29,6 @@ ${CONDA}/envs/docs-env/bin/python \
     -b html \
     -d _build/doctrees \
     -D language=en \
-    . _build/html
+    . _build/html || exit -1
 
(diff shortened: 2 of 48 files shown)


StrikerRUS:
> could you please temporarily enable builds for this branch on RTD so we can test the new configuration?

Sure, done!
https://readthedocs.org/projects/lightgbm/builds/15811372/

docs/README.rst
```diff
@@ -13,11 +13,46 @@ After each commit on ``master``, documentation is updated and published to `Read
...
+
+Locally
+^^^^^^^
+
 You can build the documentation locally. Just install Doxygen and run in ``docs`` folder
 
 .. code:: sh
 
-    pip install -r requirements.txt
+    pip install breathe sphinx 'sphinx_rtd_theme>0.5'
```
StrikerRUS, line 55: Please remove the mention of deleted `requirements_base.txt` file later in this document.

jameslamb: ah missed that one, thanks! Fixed in https://github.com/microsoft/LightGBM/pull/4957/commits/d53e14e993aff2946e32483291de66afaced32d3.

jameslamb:
> Sure, done!
https://readthedocs.org/projects/lightgbm/builds/15811372/

Thanks for that! I'm really happy to see that the build succeeded, docs look correct (I checked Python, C API, and the R docs), and that the build didn't take longer than usual.

StrikerRUS:
Hmm... `check-docs` job failed with
```
Running Sphinx v4.4.0
making output directory... done
                                                                                             
Exception occurred:
  File "/home/runner/miniconda/envs/docs-env/lib/python3.9/site-packages/sphinx/theming.py", line 207, in load_external_theme
    theme_entry_points = entry_points(group='sphinx.html_themes')
TypeError: entry_points() got an unexpected keyword argument 'group'
The full traceback has been saved in /tmp/sphinx-err-bcgclosn.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
make: *** [Makefile:20: html] Error 2
```
https://github.com/microsoft/LightGBM/runs/4863199597?check_suite_focus=true

docs/build-docs.sh
```diff
@@ -0,0 +1,35 @@
...
+conda config --set always_yes yes --set changeps1 no
+
+mamba create \
+    --yes \
+    --name docs-env \
+    --file env.yml || exit -1
+
+source activate docs-env
+
+${CONDA}/envs/docs-env/bin/python \
```
StrikerRUS, line 24: Why after `source activate docs-env` we can't simply use `python` and need to specify full path?

docs/build-docs.sh
```diff
@@ -0,0 +1,35 @@
...
+    -m sphinx \
+    -T \
+    -E \
+    -W \
+    --keep-going \
+    -b html \
+    -d _build/doctrees \
+    -D language=en \
+    . _build/html || exit -1
```
StrikerRUS, line 33: Why do we need this long command and cannot use `make html`?
https://github.com/microsoft/LightGBM/blob/master/docs/Makefile

docs/README.rst
```diff
@@ -13,11 +13,46 @@ After each commit on ``master``, documentation is updated and published to `Read
...
 -----
 
+It is not necessary to re-build this documentation while modifying LightGBM's source code.
+The HTML files generated using ``Sphinx`` are not checked into source control.
+However, you may want to build them locally during development to test changes.
+
+Docker
+^^^^^^
+
+The most reliable way to build the documentation locally is with Docker, using `the same images readthedocs uses <https://hub.docker.com/r/readthedocs/build>`_.
```
StrikerRUS, line 23: Hmm... This link gives me 404 error. And Docker Hub says `readthedocs` organization has no repositories.

image

image


But I can see specific image in that repository. Maybe that repository is "hidden"?..

image

https://hub.docker.com/layers/readthedocs/build/ubuntu-20.04-2021.09.23/images/sha256-09630e9fa6553d602756a73b63907475f59cb3667f073578f88b857693f3d359?context=explore

docs/README.rst
```diff
@@ -27,6 +62,6 @@ If you faced any problems with Doxygen installation or you simply do not need do
 
 .. code:: sh
 
-    pip install -r requirements_base.txt
+    pip install breathe sphinx 'sphinx_rtd_theme>0.5'
```
StrikerRUS, line 65:
```suggestion
    pip install sphinx 'sphinx_rtd_theme>0.5'
```

docs/build-docs.sh
jameslamb: I got this from the RTD builds, e.g. https://readthedocs.org/projects/lightgbm/builds/15824338/. Was trying to mimic exactly what is run in that environment.

Screen Shot 2022-01-19 at 10 39 14 PM

If you think that's not necessary and is the same as running `make html` I'd be happy to change it to `make html`.

docs/README.rst
jameslamb: Hmmm, I don't get a 404 when I go there in my browser.

image

Or using `curl -I https://hub.docker.com/r/readthedocs/build`

image

Notice that in both cases, I'm not authenticated with DockerHub. So I'm pretty confident that that page is publicly-accessible.

Could you try again? Maybe Docker Hub had some temporary issues around the time you tried this link.

jameslamb:
> TypeError: entry_points() got an unexpected keyword argument 'group'

I think I understand what's happening.

I've found reports claiming that this error is about a difference in `importlib-metadata` versions.

* https://github.com/apache/airflow/issues/19804#issuecomment-979363879
* https://github.com/python/importlib_metadata/issues/331#issuecomment-881704671

`importlib.metadata` (note the `.`) has been a part of the Python standard library since Python 3.8 (https://docs.python.org/3/library/importlib.metadata.html#entry-points).

`importlib-metadata` (note the `-`) is an external library providing `importlib.metadata`'s functionality. From PyPI ([link](https://pypi.org/project/importlib-metadata/)):

> This package supplies third-party access to the functionality of importlib.metadata including improvements added to subsequent Python versions.

> New features are introduced in this third-party library and later merged into CPython.

The `conda-forge` feedstock for `importlib-metadata` was archived on `conda-forge` in 2019 (https://github.com/conda-forge/importlib-metadata-feedstock).

It looks like `sphinx` will first try to use external package `importlib_metadata` if available, and otherwise will use the standardlib package. 

https://github.com/sphinx-doc/sphinx/blob/9a42e42a401444dc833261409fb55a5067409b8e/sphinx/theming.py#L19-L22

```python
try:  # Python < 3.10 (backport)
    from importlib_metadata import entry_points
except ImportError:
    from importlib.metadata import entry_points
```

I suspect that the issue here is related to mixing conda channels. Notice in https://github.com/microsoft/LightGBM/runs/4863199597?check_suite_focus=true that `conda create` is getting packages from the anaconda defaults channels, and then `conda install` is getting `conda-forge` channels.

```text
python             pkgs/main/linux-64::python-3.9.5-h12debd9_4
readline           pkgs/main/linux-64::readline-8.1-h27cfd23_0
requests           pkgs/main/noarch::requests-2.25.1-pyhd3eb1b0_0
...
...
...
doxygen            conda-forge/linux-64::doxygen-1.9.2-hb166930_0
rstcheck           conda-forge/noarch::rstcheck-3.3.1-pyh9f0ad1d_0
```

I think that specifying `nodefaults` in `env.yml` is either not working as the documentation says it should or I've misunderstood that documentation ([docs link](https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html?highlight=nodefaults#creating-an-environment-file-manually)).

> You can exclude the default channels by adding `nodefaults` to the channels list.

I tried that approach because `conda env create --file` does not accept flags like `--channel`. I'm going to try pushing a commit with a different approach. Note that this issue should go away, and this code should be simplified, once we're using `mamba[forge]`.

docs/build-docs.sh
jameslamb: ah this is left over from debugging. I ran into an issue similar to https://github.com/conda/conda/issues/3200 where `source activate` can fail when used in a bash script with `set  -e`.

This can be reverted to just `python`.

jameslamb:
I haven't been able to figure out the right combination of `conda` commands to fix this issue.

Since the builds are working correctly on RTD ([build link](https://readthedocs.org/projects/lightgbm/builds/15825981/)) and in Docker locally (both using `mamba[forge]`), I'm confident that just using `mamba[forge]` will resolve these issues.

So I'd like to request changing the order of merging PRs I'd originally suggested in this PR's description. @StrikerRUS could we merge #4953 *before* this one? Then I'll update this one to use `mamba[forge]`.

docs/build-docs.sh
jameslamb: updated in https://github.com/microsoft/LightGBM/pull/4957/commits/0c312eef03d1600853d4e35f8198ee79b6514cb0

docs/build-docs.sh
StrikerRUS: I think `make html` should be enough. If we really want to mimic `readthedocs`, we should update `make` files, but not duplicate code.
https://github.com/microsoft/LightGBM/blob/master/docs/Makefile
https://github.com/microsoft/LightGBM/blob/master/docs/make.bat

docs/README.rst
StrikerRUS: Today everything is OK 😎

StrikerRUS:
> So I'd like to request changing the order of merging PRs I'd originally suggested in this PR's description. @StrikerRUS could we merge #4953 *before* this one?

Sure!

jameslamb:
Now that #4953 has been merged, I'll work on updating this.

https://docs.readthedocs.io/en/stable/config-file/v2.html#build-tools-python says it's possible to set `python.version: miniconda3-4.7`. I'll switch from mamba to that, based on https://github.com/microsoft/LightGBM/pull/4953#issuecomment-1026766270 and https://github.com/microsoft/LightGBM/pull/4953#issuecomment-1030317210

jameslamb:

1. `checks-docs` build succeeded ([build link](https://github.com/microsoft/LightGBM/runs/5210954621?check_suite_focus=true)) 
    - logs show all packages are being installed only from `conda-forge`
2. RTD build succeeded ([build link](https://readthedocs.org/projects/lightgbm/builds/16100853/))
    - logs show all packages are being installed only from `conda-forge`
    - build took 564s (most recent builds on `master` took 700+ seconds)
    - Python package docs look correct: https://lightgbm.readthedocs.io/en/fix-rtd-configuration/Python-API.html
    - long-form docs like Distributed Learning Guide look correct: https://lightgbm.readthedocs.io/en/fix-rtd-configuration/Parallel-Learning-Guide.html
    - R docs were look correct: https://lightgbm.readthedocs.io/en/fix-rtd-configuration/R/index.html
3. I tested the `docker run` command introduced in `docs/README.rst` in this PR on my Mac tonight. It ran successfully and the docs rendered locally look correct.

docs/README.rst
```diff
@@ -13,11 +13,45 @@ After each commit on ``master``, documentation is updated and published to `Read
...
+        --workdir=/opt/LightGBM/docs \
+        --entrypoint="" \
+        -it readthedocs/build:ubuntu-20.04-2021.09.23 \
+        /bin/bash build-docs.sh
+
+When that code completes, open ``docs/html/index.html`` in your browser.
+
+.. note::
+
+    The navigation in locally-built docs does not link to the local copy of the R documentation. To view the local version of the R docs, open ``docs/_build/html/R/index.html`` in your browser.
```
StrikerRUS, line 45: I think we have some confusing wordings in this doc about R docs.

There are two subsections `Docker` and `Locally` in this doc. TBH, I think that Docker can be treated as "building locally" actually. But it's just a nit. The problem is that in the `Docker` section we use `in locally-built docs` phrase which can be easily treated as a reference to the `Locally` section. Let me suggest to replace `Locally` with `Manually`.

docs/README.rst
```diff
@@ -13,11 +13,45 @@ After each commit on ``master``, documentation is updated and published to `Read
...
+
 You can build the documentation locally. Just install Doxygen and run in ``docs`` folder
 
 .. code:: sh
 
-    pip install -r requirements.txt
+    pip install breathe sphinx 'sphinx_rtd_theme>0.5'
     make html
 
 Unfortunately, documentation for R code is built only on our site, and commands above will not build it for you locally.
```
StrikerRUS, line 57: Not true anymore. R docs are built if one choose `Docker` way to generate docs.

docs/README.rst
```diff
@@ -13,11 +13,45 @@ After each commit on ``master``, documentation is updated and published to `Read
...
+
+Locally
+^^^^^^^
+
 You can build the documentation locally. Just install Doxygen and run in ``docs`` folder
 
 .. code:: sh
 
-    pip install -r requirements.txt
+    pip install breathe sphinx 'sphinx_rtd_theme>0.5'
```
StrikerRUS, line 54:
```suggestion
    pip install breathe sphinx 'sphinx_rtd_theme>=0.5'
```

docs/README.rst
```diff
@@ -27,6 +61,6 @@ If you faced any problems with Doxygen installation or you simply do not need do
 
 .. code:: sh
 
-    pip install -r requirements_base.txt
+    pip install sphinx 'sphinx_rtd_theme>0.5'
```
StrikerRUS, line 64:
```suggestion
    pip install sphinx 'sphinx_rtd_theme>=0.5'
```

docs/build-docs.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+rm -f ./_FIRST_RUN.flag
+
+export PATH="${CONDA}/bin:${PATH}"
+
+curl \
+    -L \
```
StrikerRUS, line 8:
```suggestion
    -sL \
```

docs/build-docs.sh
```diff
@@ -0,0 +1,23 @@
...
+    -o ${HOME}/miniforge.sh \
+    https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-Linux-x86_64.sh
+
+/bin/bash ${HOME}/miniforge.sh -b -p ${CONDA}
+
+conda config --set always_yes yes --set changeps1 no
```
StrikerRUS, line 14: Maybe we need to update `conda` itself here?
https://github.com/microsoft/LightGBM/blob/d31346f6c265e994c54ebddd19ed1fb6445ba0c9/.ci/setup.sh#L115

docs/env.yml
```diff
@@ -0,0 +1,17 @@
...
+  - breathe
+  - python=3.9
+  - r-base=4.1.2
+  - r-data.table=1.14.2
+  - r-jsonlite=1.7.2
+  - r-knitr=1.37
+  - r-matrix=1.4_0
+  - r-pkgdown=1.6.1
+  - r-rmarkdown=2.11
+  - r-roxygen2=7.1.2
```
StrikerRUS, line 15: We don't want to pin R package versions anymore?.. https://github.com/microsoft/LightGBM/pull/2176#issuecomment-499058519

StrikerRUS (changes requested):
Thank you so much for working on this! Please consider checking my fresh comments below.

docs/env.yml
jameslamb: The comment you've linked refers to a commit, https://github.com/microsoft/LightGBM/pull/2176/commits/014bee124102f4a1762e4be5316fab52e852f143, that changed the list of package from "not pinned at all" to "pin to aa specific BUILD of a specific VERSION".

This PR's changes are still pinning them to specific package VERSIONS, just not to individual BUILDS. Sometimes new builds of the same package version are pushed to conda channels, to fix issues like, for example, mistakes in metadata about declared dependencies. We should want to get such fixes automatically, as they make it more likely that the environment will solve correctly.

docs/build-docs.sh
jameslamb: sure

jameslamb: updated in https://github.com/microsoft/LightGBM/pull/4957/commits/8010f99fda097fa9702b9c900f31d5d078eaf1fc

docs/README.rst
jameslamb: what do you think about just calling the other section `Without Docker`, to eliminate any confusion?

I tried that here: https://github.com/microsoft/LightGBM/pull/4957/commits/8010f99fda097fa9702b9c900f31d5d078eaf1fc

StrikerRUS: I love it! 🙂

docs/README.rst
```diff
@@ -13,20 +13,55 @@ After each commit on ``master``, documentation is updated and published to `Read
...
+        -v $(pwd):/opt/LightGBM \
+        --env C_API=true \
+        --env CONDA=/opt/miniforge \
+        --env READTHEDOCS=true \
+        --workdir=/opt/LightGBM/docs \
+        --entrypoint="" \
+        -it readthedocs/build:ubuntu-20.04-2021.09.23 \
+        /bin/bash build-docs.sh
+
+When that code completes, open ``docs/html/index.html`` in your browser.
```
StrikerRUS, line 41: I believe `_build` was missed here.
```suggestion
When that code completes, open ``docs/_build/html/index.html`` in your browser.
```

docs/README.rst
```diff
@@ -13,20 +13,55 @@ After each commit on ``master``, documentation is updated and published to `Read
...
 .. code:: sh
 
-    pip install -r requirements.txt
+    pip install breathe sphinx 'sphinx_rtd_theme>=0.5'
     make html
 
-Unfortunately, documentation for R code is built only on our site, and commands above will not build it for you locally.
+Note that this will not build the R documentation.
+To build the R docuumentation locally, use the Docker-based approach described above.
 Consider using common R utilities for documentation generation, if you need it.
```
StrikerRUS, line 59:
```suggestion
Note that this will not build the R documentation.
Consider using common R utilities for documentation generation, if you need it.
Or use the Docker-based approach described above to build the R documentation locally.
```

docs/env.yml
StrikerRUS: > The comment you've linked refers to a commit, https://github.com/microsoft/LightGBM/commit/014bee124102f4a1762e4be5316fab52e852f143, that changed the list of package from "not pinned at all" to "pin to aa specific BUILD of a specific VERSION".

More relevant past discussion: https://github.com/microsoft/LightGBM/pull/2725#discussion_r373726849.

OK, let's not pin build numbers for now. But if RTD starts to fail, we know what to try first.

StrikerRUS (approved):
Thanks a lot! Just two minor suggestions below and I think this PR can be merged.

Issue #4952, opened 2022-01-15
OOM with device_type = cuda when using lgb.cv(...) for parameters grid search


denmoroz:
## Description
I am using `lgb.cv` with `device_type = cuda` to find the best parameters set on validation data using grid search.

LightGBM fails with "Out Of Memory" after few training runs on the same fixed dataset and different parameter sets that are not dependent on data (`num_leaves`, `lambda_l1`, `lambda_l2` and some others).

## Reproducible example
Let's take [simple example](https://github.com/microsoft/LightGBM/blob/master/examples/python-guide/simple_example.py) as a starting point and update it a bit to emulate cross-validation over `num_leaves` using `cuda` device:

```
# coding: utf-8
from pathlib import Path

import pandas as pd
from sklearn.metrics import mean_squared_error

import lightgbm as lgb

print('Loading data...')
# load or create your dataset
regression_example_dir = Path(__file__).absolute().parents[1] / 'regression'
df_train = pd.read_csv(str(regression_example_dir / 'regression.train'), header=None, sep='\t')
df_test = pd.read_csv(str(regression_example_dir / 'regression.test'), header=None, sep='\t')

y_train = df_train[0]
y_test = df_test[0]
X_train = df_train.drop(0, axis=1)
X_test = df_test.drop(0, axis=1)

# create dataset for lightgbm
lgb_train = lgb.Dataset(X_train, y_train)
lgb_eval = lgb.Dataset(X_test, y_test, reference=lgb_train)

# specify your configurations as a dict
params = {
    'device_type': 'cuda',
    'boosting_type': 'gbdt',
    'objective': 'regression',
    'metric': {'l2', 'l1'},
    'learning_rate': 0.05,
    'feature_fraction': 0.9,
    'bagging_fraction': 0.8,
    'bagging_freq': 5,
    'verbose': 0
}

for num_leaves in [31, 63, 255]:
    print('Starting training...')

    training_params = params.copy()
    training_params["num_leaves"] = num_leaves

    gbm = lgb.train(params,
                    lgb_train,
                    num_boost_round=1000,
                    valid_sets=lgb_eval,
                    callbacks=[lgb.early_stopping(stopping_rounds=100)])

    print('Starting predicting...')
    # predict
    y_pred = gbm.predict(X_test, num_iteration=gbm.best_iteration)
    # eval
    rmse_test = mean_squared_error(y_test, y_pred) ** 0.5
    print(f'The RMSE of prediction is: {rmse_test}')
[...]

PR #4963, opened 2022-01-20
CUDATreeLearner: free GPU memory in destructor if any allocated

denmoroz:
Fixes memory freeing issue described [here](https://github.com/microsoft/LightGBM/issues/4952): allocated GPU memory is not released after training cycle is complete, which leads to constant memory growth while using multiple train() invocations inside the same process (like hyperparameters search).

### DIFF
--- a/src/treelearner/cuda_tree_learner.cpp
+++ b/src/treelearner/cuda_tree_learner.cpp
@@ -63,6 +63,47 @@ CUDATreeLearner::CUDATreeLearner(const Config* config)
 }
 
 CUDATreeLearner::~CUDATreeLearner() {
+ #pragma omp parallel for schedule(static, num_gpu_)
+
+  for (int device_id = 0; device_id < num_gpu_; ++device_id) {
+    // do nothing it there is no gpu feature
+    int num_gpu_feature_groups = num_gpu_feature_groups_[device_id];
+    if (num_gpu_feature_groups) {
+      CUDASUCCESS_OR_FATAL(cudaSetDevice(device_id));
+
+      if (device_features_[device_id] != NULL) {
+        CUDASUCCESS_OR_FATAL(cudaFree(device_features_[device_id]));
+      }
+
+      if (device_gradients_[device_id] != NULL) {
+        CUDASUCCESS_OR_FATAL(cudaFree(device_gradients_[device_id]));
+      }
+
+      if (device_hessians_[device_id] != NULL) {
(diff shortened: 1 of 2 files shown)


denmoroz:
Ups, seems like some tests are not passing as they are cancelled by timeout. Trying to restart.

denmoroz:
Still getting
image

jmoralez:

> Ups, seems like some tests are not passing as they are cancelled by timeout.

That is due to #4948 and will be solved by #4953, once that is merged we'll let you know to update this branch and everything should succeed.

denmoroz:
@jmoralez 
Oh, I see.

> once that is merged we'll let you know to update this branch and everything should succeed
Thanks 🙏

jmoralez:
Hi @denmoroz. #4953 has been merged, can you please update this branch to include those changes?

src/treelearner/cuda_tree_learner.cpp
```diff
@@ -63,6 +63,43 @@ CUDATreeLearner::CUDATreeLearner(const Config* config)
 }
 
 CUDATreeLearner::~CUDATreeLearner() {
+  #pragma omp parallel for schedule(static, num_gpu_)
```
shiyu1994, line 66: With a static schedule, and `num_gpu_` chunk size, I think there will be only 1 thread being used. So I can we simply remove this line?

denmoroz: @shiyu1994 
To be honest I've just copy-pasted simplified code from [here](https://github.com/microsoft/LightGBM/blob/master/src/treelearner/cuda_tree_learner.cpp#L319) which uses `omp parallel`, so I decided that probably it is done intentionally. Do you think we should actually remove this?

denmoroz:
> Hi @denmoroz. #4953 has been merged, can you please update this branch to include those changes?

@jmoralez Sure! Sorry for delayed response, just figured out that.

src/treelearner/cuda_tree_learner.cpp
shiyu1994: After experimenting, I find that schedule(static, num_gpu_) will only result in a single thread. With `num_gpu_` block size, and with `num_gpu_` total iterations. So this would not parallelize the for loop.
I think it is ok to keep the `omp parallel` in this PR. And we can fix this with another PR. Thank you!

shiyu1994 (approved):
LGTM. Thanks for your contribution!

PR #4964, opened 2022-01-21
Target and Count encodings for categorical features - Iteration 1

tongwu-msft:
Split PR: [Target and Count encodings for categorical features](https://github.com/microsoft/LightGBM/pull/3234) to multiple part. 

## Implementation
1. CategoryFeatureEncoder: abstract class for encoders.
2. CategoryFeatureEncoderManager: encode value for aggregated multiple encoders.
3. CategoryFeatureTargetInformationCollector: collect data information before training.

### DIFF
--- a/src/feature_engineering/category_feature_target_encoder.cpp
+++ b/src/feature_engineering/category_feature_target_encoder.cpp
@@ -1,0 +1,39 @@
+/*!
+  * Copyright (c) 2016 Microsoft Corporation. All rights reserved.
+  * Licensed under the MIT License. See LICENSE file in the project root for license information.
+  */
+
+#include "encoder.hpp"
+
+namespace LightGBM {
+  double CategoryFeatureTargetEncoder::Encode(double feature_value) {
+    int category = static_cast<int>(feature_value);
+
+    if (count_information_.find(category) != count_information_.end()) {
+      return (count_information_[category] + prior_ * prior_weight_) / (total_count_ + prior_weight_);
+    }
+
+    return default_value;
+  }
+
+  json11::Json::object CategoryFeatureTargetEncoder::DumpToJsonObject()
+  {
+    json11::Json::object result = CategoryFeatureEncoder::DumpToJsonObject();
+    result[encoder_type_key] = json11::Json(target_encoder_type);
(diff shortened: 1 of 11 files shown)


src/feature_engineering/category_feature_target_encoder.cpp
```diff
@@ -0,0 +1,79 @@
...
+// property name keys
+const std::string count_information_key = "count_information";
+const std::string label_information_key = "label_information";
+const std::string category_key = "cat";
+const std::string value_key = "value";
+const std::string count_prior_key = "prior";
+const std::string count_prior_weight_key = "prior_weight";
+
+namespace LightGBM {
+  double CategoryFeatureTargetEncoder::Encode(double feature_value) {
```
shiyu1994, line 17: Is the fold ID considered for encoding of training dataset?

shiyu1994: Ok. I see that the encoder is per-fold. So there's no question.

shiyu1994:
@tongwu-msft Thank you for working on this. The design overall looks good to me. Can we move on to integrating with the dataset classes?

tongwu-msft:
> @tongwu-msft Thank you for working on this. The design overall looks good to me. Can we move on to integrating with the dataset classes?

Yes, let me split the big PR into multiple iterations, this is it1 and we can do it parallel.

jameslamb:
I'm very confused. After 7 months of inactivity, this PR was closed without any comment...and it was supposed to replace #3234 which has not received a commit in about 8 months.

@shiyu1994 @tongwu-msft @guolinke what is happening with the large changes to LightGBM's handling of categorical values? Is someone working on this and should it still be considered a requirement for v4.0.0 (#5153)?

cc @jmoralez @StrikerRUS @btrotta

shiyu1994:
I can pick this up. But I'll first focus on the CUDA parts before v4.0.0. Another option would be extracting the categorical feature handling into a separate module for LightGBM dataset preprocessing (something like https://github.com/microsoft/LightGBM-transform). So that we can focus on the core training algorithm of GBDT in our codebase.

PR #4967, opened 2022-01-22
[ci] bump CUDA version from `11.5.0` to `11.5.1` at CI

StrikerRUS:
https://catalog.ngc.nvidia.com/orgs/nvidia/containers/cuda/tags

Also, revert #4648 according to https://gitlab.com/nvidia/container-images/cuda/-/issues/140#note_806446655.

### DIFF
--- a/.ci/setup.sh
+++ b/.ci/setup.sh
@@ -98,15 +98,9 @@ else  # Linux
                 clang \
                 libomp-dev
         fi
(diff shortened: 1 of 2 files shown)


jameslamb (approved):
Excellent! Glad we were able to revert that `apt-get` change and simplify `setup.sh` a bit.

Issue #4295, opened 2021-05-16
[R-package] Request: provide more idiomatic interface


david-cortes:
LightGBM for R has an interface which requires creating a dataset object from an R native type, and passing this custom type to the model training function. The prediction functions then take native R objects instead.

This interface is inconvenient to use, and is very different from what base R or typical R packages for decision trees (or classification/regresssion in general) have as interface.

Would be better if lightgbm could offer a more idiomatic interface, like `ranger` for example, which among others:
* Allows a formula interface as well as an X/y interface, both of which accept R's own types (including data frames).
* Accepts non-standard evaluation for column names in data frames.
* Automatically determines categorical features when using data frames.

jameslamb:
Thank you for your interest in LightGBM, and for writing this up! In the future, please link to previous conversations when you create new issues from them, as this helps us to keep track of the many conversations in this project. Specifically, I believe the content of this issue is very closely related to #42007 and ttps://github.com/microsoft/LightGBM/pull/4207#issuecomment-840895032.

> * Allows a formula interface as well as an X/y interface, both of which accept R's own types (including data frames).
>
> * Accepts non-standard evaluation for column names in data frames.

Have you seen `{treesnip}`'s LightGBM integration?

https://github.com/curso-r/treesnip/blob/bf27cd871b7fb663a76818f66ea0a5f9bf09f444/tests/testthat/test-lightgbm.R#L16

Does that package's interface satisfy the first two items (accept a formula, use non-standard evaluation for column names)? If not, it would be very helpful if you could provide code examples describing how you would like `{lightgbm}` to work. More specifics on the interface you would like this `{lightgbm}` to provide are necessary for maintainers or other contributors to do the work to satisfy this request.

> could offer a more idiomatic interface

Is there a standard in the R ecosystem that you recommend `{lightgbm}` try to follow? Similar to https://scikit-learn.org/stable/developers/develop.html?

david-cortes:
Yes, `treesnip` provides an interface with handling of data frames, categorical features, and non-standard evaluation; but it is a `parsnip` connector.

A good standard to follow would be Tidymodel's guidelines for modeling packages:
https://tidymodels.github.io/model-implementation-principles/

`ranger` is a very good practical example to copy from. Lots of other R packages for decision trees follow similar conventions (e.g. `randomForest`, `gbm`, `party`, `C5.0`, among many others) - not sure if I'd call it an ecosystem though.

PR #4685, opened 2021-10-14
[R-package] enable saving Booster with saveRDS() and loading it with readRDS() (fixes #4296)

david-cortes:
Fixes #4296

This PR adds changes as necessary for model objects to be serializable through functions such as `saveRDS` and `qsave` by keeping raw bytes in the model object, from which a C++ object can be reconstructed when such object is de-serialized and the pointer reset to null.

PR #4968, opened 2022-01-23
[R-package] User-friendly redesign for `lightgbm()`

david-cortes:
ref #4295

This PR redesigns the R interface for `lightgbm()` so as to make it similar to those of other packages and base R's, while keeping the old `lgb.train` interface as-is. This redesign includes changes such as:
* Allowing a formula interface as well as an X/y interface with non-standard evaluation for arguments.
* Allowing `data.frame` inputs and handling the encoding of categorical variables in the X/y interface.
* Auto-detecting binary/multiclass classification according to the type of the target variable.
* Renaming function arguments to make them more in line with what is used in core R packages.
* Minor details such as ensuring that the prediction outputs have appropriate row and colum names.
* Using base-1 indexing in the outputs.
* Changing default arguments to match those of Python (e.g. number of threads).
* Expanding a bit the documentation.


jameslamb (changes requested):
Thanks very much for your continued interest in LightGBM!

However, I'm not willing to review this pull request in its current form.

As we have politely mentioned to you in the past, such large pull requests with multiple unrelated changes are very difficult to review. Some examples: https://github.com/microsoft/LightGBM/pull/4207#pullrequestreview-659266142, https://github.com/microsoft/LightGBM/pull/4685#pullrequestreview-788846414, . 

As we've also politely mentioned to you several times in the past, making comments like "similar to those of other packages" or "more in line with what is used in core R packages" without providing reference links or specific evidence unnecessarily increases the effort required to review such proposals. Some examples: https://github.com/microsoft/LightGBM/pull/4692#issuecomment-949981898, https://github.com/microsoft/LightGBM/pull/4692#pullrequestreview-790114309, https://github.com/microsoft/LightGBM/pull/4207#issuecomment-824460193, https://github.com/microsoft/LightGBM/pull/4812#pullrequestreview-810231067, https://github.com/microsoft/LightGBM/pull/4597#issuecomment-913922705, https://github.com/microsoft/LightGBM/issues/4681#issuecomment-943730750.

This pull request is proposing several very large breaking changes to the R package's interface, each of which requires careful consideration and conversation about what the benefits of those changes are and whether they are worth the cost of breaking users' existing code. We have an obligation to LightGBM's existing users to understand and carefully consider changes to this project, especially changes that break existing code, and can't do that effectively for so many changes at one time.

We really appreciate all of your help with LightGBM, but need to be protective of the limited time that this project's small group of mostly-volunteer maintainers can devote to maintaining the project.

<hr>

Could you please make each of the related sets of changes a separate pull request or proposal in an issue, which could be reviewed and discussed independently, and in each please provide specific evidence describing the benefit of the proposed change?

Looking through these changes tonight, I see the following sets of changes which should each be their own pull request / issue:

1. Moving parameters `objective`, `threads`, and `metric` out of `params` list in examples and tests and passing it through keyword arguments.
2. Other changes in tests that move parameters out of `params` and into keyword arguments of `lightgbm::lightgbm()`.
3. Changes to argument names in `predict.lgb.Booster()`.
4. Changing `lightgbm::lightgbm()` from a regular function to a set of S3 methods that dispatch on the type of `data` (WITHOUT introducing the new `formula` support or changing any argument names).
5. Adding support for formulas.
6. Automatically setting `categorical_features` based on data frame column types (as part of #4323).
7. Handling `factor` columns in `data.frame` inputs (as part of #4323)
8. Changing argument names `data` and `label` to `X` and `y` in `lightgbm::lightgbm()`.
9. non-standard evaluation of arguments using `eval.parent()`, `deparse1()`, `substitute()`, etc.
10. adding column names to predictions
11. addition of a "class" type of prediction, where `predict()` returns the predicted class instead of probabilities for classification tasks

StrikerRUS:
@david-cortes Thank you so much for your huge work!

@jameslamb I absolutely agree with you that this PR should be spit into smaller ones. I just want to add that open one big PR with all changes first is a good idea because maintainers can assess overall idea and understand a final state that the proposing changes move the project into. Big PR can be treated as a playground for initial and general discussions (demonstrative alternative to original feature request). Also, it's good to test how all proposed connected changes fit together.

See for example https://github.com/dmlc/xgboost/pull/7545#issuecomment-1019271818 or https://github.com/dmlc/xgboost/pull/7309#issuecomment-957323649.

PR #4970, opened 2022-01-23
[R-package] Add base-1 indexing option for prediction outputs

david-cortes:
ref #4968

The prediction function in the R package produces outputs that have a base-0 numeration. R uses base-1 numeration however, which makes the result a bit confusing and harder to use on the outside.

What's even more confusing, it applies base-0 numeration also to `start_iteration` without any warning about it either.

This PR adds an option to make the inputs and result have a base-1 numeration instead, and along the way corrects type `None` (which does not exist in R) in the docs about these parameters to say `NULL`.


david-cortes:
The mac CI failures are due to dependencies that failed to install and not related to this PR.

jameslamb:
I haven't forgotten about this PR, but I'm not ready to review it yet. There are many places in the R package which deal with references to numeration (i.e. where users have to provide a number to refer to "the `n`th thing"), and those references use a mix of 1-based and 0-based indexing.

Before considering this proposal to control that behavior with a boolean flag in one limited place, I'd like to list out all such places in the package's public API and see if we can come up with a general standard for when the R package uses 1-based numeration and when it uses 0-based numeration.

I'll try to complete that list and open an a "request for comment (RFC)" issue with a proposal soon.

jameslamb (changes requested):
I haven't forgotten about this PR, but I'm not ready to review it yet. There are many places in the R package which deal with references to numeration (i.e. where users have to provide a number to refer to "the `n`th thing"), and those references use a mix of 1-based and 0-based indexing.

Before considering this proposal to control that behavior with a boolean flag in one limited place, I'd like to list out all such places in the package's public API and see if we can come up with a general standard for when the R package uses 1-based numeration and when it uses 0-based numeration.

I'll try to complete that list and open an a "request for comment (RFC)" issue with a proposal soon.

Issue #4425, opened 2021-07-01
Question: why setting num_threads = cpu cores is better


LittleLittleCloud:
I'm running a few experiments using lgbm in docker container, and I find a SIGNIFICANT speed improvement when setting `num_threads` equals to cpu number instead of thread number (which is cpu number * 2). Just wonder if it is expected?

FYI, the cpu limit is set to 1 for that docker container. So the old `num_threads` is 2 while the new `num_threads` is 1. And the new `num_thread` improves training speed by 20X approximately.

shiyu1994:
The physical core numbers of CPU is the actual number of cores. But the thread number, which is twice of the CPU core numbers in some intel CPUs, is due to the hyperthreading design. Hyperthreading means that a CPU core can be allocated 2 threads so that when one thread waits for data being fetched from memory, another thread can be scheduled to the core. In this way, we can hide memory latency.

Using hyperthreading maybe helpful in some cases. But it can also hurt the performance in other cases. For example, when each thread use a lot of cache, two threads in the same core may compete for the cache space, so there can be more cache misses with hyperthreading.

It is hard to determine from the algorithm itself whether hyperthreading is useful or not. But I observed that in GBDT applications, using hyperthreading is often a bad choice.

StrikerRUS:
> Just wonder if it is expected?

Yeah, it's expected and have been already documented:

> for the best speed, set this to the number of **real CPU cores**, not the number of threads (most CPUs use hyper-threading to generate 2 threads per CPU core)
   https://lightgbm.readthedocs.io/en/latest/Parameters.html#num_threads

jameslamb:
Agree with everything said above. Linking this related issue that provides a little bit more context: https://github.com/microsoft/LightGBM/issues/4192#issuecomment-822023560

PR #4972, opened 2022-01-23
[R-package] Promote number of threads to top-level argument in `lightgbm()` and change default to number of cores

david-cortes:
ref #4968

The Python interface for `lightgbm` allows passing the number of threads to use as a top-level parameter, while the R interface does not. The Python interface also defaults to using all available threads, while the R interface doesn't.

This PR promotes the number of threads towards top-level parameter in `lightgbm()`, changes the default to match Python's, and explains that a number of "zero" means using the configured OMP threads.

IMHO the number of threads is a very common thing to want to control (probably more so than other top-level parameters such as `callbacks`), and its default value should be explicit to the user.


R-package/R/lightgbm.R
```diff
@@ -114,12 +126,14 @@ NULL
...
 #' @inheritSection lgb_shared_params Early Stopping
 #' @return a trained \code{lgb.Booster}
+#' @importFrom parallel detectCores
 #' @export
 lightgbm <- function(data,
                      label = NULL,
                      weight = NULL,
                      params = list(),
                      nrounds = 100L,
+                     nthreads = parallel::detectCores(),
```
jameslamb, line 136: Inserting this argument in the middle of the function signature is a breaking change, from the perspective of anyone calling `lightgbm::lightgbm()` using positional arguments.

Please move this to the bottom of the argument list (just before `...`) and move the corresponding parameter documentation in the roxygen comments to the same position in the function's documentation.

------

Per the conversation in #4192, https://github.com/microsoft/LightGBM/issues/4425#issuecomment-872943848, and elsewhere, LightGBM tends to run fastest when `num_threads` is set to the number of *physical* (not *logical*) CPUs. For `lightgbm::lightgbm()`, I recommend changing this to `parallel::detectCores(logical = FALSE)`.

---

Since this is being introduced as a top-level argument in the R package for the first time, please use the name `num_threads` instead of `nthreads`. I think it's desirable to have these keyword arguments match the "main" LightGBM parameters: https://lightgbm.readthedocs.io/en/latest/Parameters.html#num_threads

Existing cases where that isn't true (e.g. `nrounds` instead of `num_iterations`) have just been preserved for backwards compatibility.

R-package/NAMESPACE
```diff
@@ -52,6 +52,7 @@ importFrom(graphics,barplot)
 importFrom(graphics,par)
 importFrom(jsonlite,fromJSON)
 importFrom(methods,is)
+importFrom(parallel,detectCores)
```
jameslamb, line 55: please add `parallel` to the `Imports:` section in `DESCRIPTION`, immediately following `methods`.

https://github.com/microsoft/LightGBM/blob/d163c2c16629e3da41dc5d7b8dbb76b8fc6249db/R-package/DESCRIPTION#L57-L64

R-package/R/lightgbm.R
```diff
@@ -92,6 +92,18 @@ NULL
 #' @inheritParams lgb_shared_params
 #' @param label Vector of labels, used if \code{data} is not an \code{\link{lgb.Dataset}}
 #' @param weight vector of response values. If not NULL, will set to dataset
+#' @param nthreads Number of parallel threads to use. For best speed, this should be set to the number of
+#'                 physical cores in the CPU - in a typical x86-64 machine, this corresponds to half the
+#'                 number of maximum threads (e.g. `nthreads = max(parallel::detectCores() / 2L, 1L)` as
+#'                 a shorthand for the optimal value).
```
jameslamb, line 98:
```suggestion
#' @param nthreads Number of parallel threads to use. For best speed, this should be set to the number of
#'                 physical cores in the CPU - (e.g. \code{parallel::detectCores(logical = FALSE)})
```

As mentioned in https://github.com/microsoft/LightGBM/issues/4192#issuecomment-822023560, it's possible to get this information directly from `parallel::detectCores()` without using any heuristics. Please refer to that in the documentation instead.

R-package/R/lightgbm.R
```diff
@@ -92,6 +92,18 @@ NULL
...
+#'
+#'                 Be aware that using too many threads can result in speed degradation in smaller datasets
+#'                 (see the parameters documentation for more details).
+#'
+#'                 If passing zero, will use the default number of threads configured for OpenMP
+#'                 (typically controlled through an environment variable `OMP_NUM_THREADS`).
+#'
+#'                 This parameter overrides `num_threads` in `params` if it exists there.
```
jameslamb, line 106:
```suggestion
```

Across LightGBM's R and Python packages, we try to enforce the behavior that elements in `params` (a list in R and a dictionary in Python) always take precedence over keyword arguments to individual functions.

Please remove this line in the documentation.

R-package/R/lightgbm.R
```diff
@@ -135,6 +149,8 @@ lightgbm <- function(data,
     stop("nrounds should be greater than zero")
   }
 
+  params$num_threads <- nthreads
```
jameslamb, line 152:
```suggestion
  params <- lgb.check.wrapper_param(
    main_param_name = "num_threads"
    , params = params
    , alternative_kwarg_value = num_threads
  )
```

Across LightGBM's Python and R packages, we try to enforce the behavior in `params` overrides LightGBM's defaults set in https://github.com/microsoft/LightGBM/blob/master/include/LightGBM/config.h and anything passed in via keyword arguments.

Please change this line to enforce that behavior, and to ensure that only a single value of `num_threads` and its aliases (https://lightgbm.readthedocs.io/en/latest/Parameters.html#num_threads) is passed down into `lgb.train()`.

R-package/tests/testthat/test_Predictor.R
```diff
@@ -68,6 +68,7 @@ test_that("start_iteration works correctly", {
     bst <- lightgbm(
         data = as.matrix(train$data)
         , label = train$label
+        , nthreads = 1L
```
jameslamb, line 71: This change and similar changes I see in this PR (like in the vignettes) where `nthreads = 1L` is being added don't seem necessary to implement support for passing through `num_threads` as a top-level parameter in `lightgbm::lightgbm()`.

Please explain why these are necessary or remove them.

jameslamb (changes requested):
Thank you for the proposal! I'm ok with adding `num_threads` as a top-level argument to `lightgbm::lightgbm()`, and with using the built-in `{parallel}` package to set a default value for it.

Please see my first round of review comments.

In addition, please add unit tests checking this behavior. At least the following:

* "`lightgbm()` does not require parameter num_threads"
* "if num_threads in `params` and keyword argument to lightgbm() have different values, the value in params is used"
* "if num_threads is not found in `params`, `lightgbm()` uses the value passed to keyword argument `num_threads`"

Whenever you propose user-facing changes to this project, please add tests on those changes to improve maintainers' confidence and to ensure that future developments in the project don't break the proposed behavior inadvertently.

R-package/tests/testthat/test_Predictor.R
david-cortes: They are necessary, because CRAN policies say that the examples, tests, and vignettes should only use 1 or 2 threads. Otherwise, there would have to be a version for running on them on CRAN and another version for running it locally, or it'd have to run a purrr wrapper or similar that detects whether it's being run on CRAN servers or not.

R-package/NAMESPACE
david-cortes: `parallel` is part of base R. Base packages are not supposed to be mentioned there.

R-package/R/lightgbm.R
david-cortes: Ok, will change it to `parallel::detectCores(logical = FALSE)`, however:
* In my own setup, I get better timings when fitting models with the full number of threads, despite the recommendation.
* The default for Python uses the full number of threads.
* Passing `logical=FALSE` doesn't always do what you're thinking, and it doesn't work in all operating systems.

david-cortes:
Updated.

R-package/tests/testthat/test_Predictor.R
StrikerRUS: Can we then set this param to `2L` with the aim to slightly speed up our CI tests?
Also, I think it would be nice to have `NTHREADS` global variable similarly to the `VERBOSITY` one
https://github.com/microsoft/LightGBM/blob/820ae7e65e90860061ec89cc4a1ad0a39b96e051/R-package/tests/testthat/test_lgb.Booster.R#L1-L3

R-package/tests/testthat/test_basic.R
```diff
@@ -2928,6 +2949,36 @@ test_that("lightgbm() defaults to 'regression' objective if objective not otherw
...
+  expect_equal(bst$params$num_threads, 1L)
+
+  bst <- lightgbm(
+    data = train$data
+    , label = train$label
+    , nrounds = 5L
+    , verbose = VERBOSITY
+    , params = list(num_threads = 1L)
+  )
+  expect_equal(bst$params$num_threads, 1L)
```
StrikerRUS, line 2969: I believe this check is quite "weak". It will be a lot better to inspect dumped LightGBM model to ensure that `num_threads` parameter reaches cpp side and is properly treated.

Here is an example in Python:
```
import numpy as np
import lightgbm as lgb

X = np.random.random((50, 10))
y = np.random.random(50)

est = lgb.LGBMRegressor(n_estimators=5, n_jobs=7).fit(X, y)
est.booster_.save_model('model.txt')
```
You can find `[num_threads: 7]` line in the `model.txt` file under the `parameters:` section.

R-package/R/lightgbm.R
jameslamb: > In my own setup, I get better timings when fitting models with the full number of threads, despite the recommendation.

That's different from what we've observed in the past (e.g. https://github.com/microsoft/LightGBM/issues/4192#issuecomment-822023560, https://github.com/microsoft/LightGBM/issues/4425#issuecomment-872661139), but I guess it's dependent on the type of hardware you have, your operating system, version of relevant libraries, and where LightGBM is spending time in a particular training run.

> doesn't always do what you're thinking, and doesn't work in all operating systems

Oh interesting! I went researching for what you meant by this statement, and see the following in `?parallel::detectCores()`

```text
logical | if possible, use the number of physical CPUs/cores (if FALSE) or logical CPUs (if TRUE).
Currently this is honoured only on macOS, Solaris and Windows.
```

II also noticed this:

```text
An integer, NA if the answer is unknown.
```

Now that I see that it's possible for this function to return `NA`, I think that for maximum portability the default value should be set to `0L` (meaning the value of `omp_get_num_threads()`), instead of using any of R's mechanisms for detecting the available CPUs. That way, we wouldn't need to worry about an `NA` being passed through to that parameter (I'm not sure how that would be handled on the C++ side).

Setting the default to `0L` would also be consistent with the default on the C++ side:

* https://lightgbm.readthedocs.io/en/latest/Parameters.html#num_threads
* https://github.com/microsoft/LightGBM/blob/60e72d5f4ec5ec2a4d30a53633e1e2d59c81309b/include/LightGBM/config.h#L205

And technically consistent with the default in the `scikit-learn` estimators in the Python package, where a value of `-1` is passed, since any value `<= 0` is treated by LightGBM as "take the result of `omp_get_num_threads()`".

* https://github.com/microsoft/LightGBM/blob/60e72d5f4ec5ec2a4d30a53633e1e2d59c81309b/python-package/lightgbm/sklearn.py#L365
* https://github.com/microsoft/LightGBM/blob/60e72d5f4ec5ec2a4d30a53633e1e2d59c81309b/include/LightGBM/utils/openmp_wrapper.h#L30-L34
* 

Please just change the default to `0L` and remove `{parallel}` from this PR. Thanks for pointing me towards the fact that `detectCores(logical = FALSE)` has different behavior on different systems!

R-package/NAMESPACE
jameslamb: I believe, based on https://github.com/microsoft/LightGBM/pull/4972#discussion_r838027789, that `{parallel}` should be removed from this PR, but in case this comes up in the future I'll explain here the evidence in favor of putting it in `Imports`.

"Writing R Extensions", in the ["Package Dependencies"](https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Package-Dependencies) sub-section in the section about `DESCRIPTION`, explicitly states that package authors should include "all the standard packages that are used".

> The ‘Imports’ field lists packages whose namespaces are imported from (as specified in the NAMESPACE file) but which do not need to be attached... Ideally this field will include all the standard packages that are used

The only package which it says not to include is `{base}`.

> It makes no sense to declare a dependence on R without a version specification, nor on the package **base**: this is an R package and package **base** is always available

As shown in the link above, `{lightgbm}` includes other default packages like `{graphics}` and `{utils}` in `Imports:`. Those have been there since the first CRAN release in October 2020, and CRAN has never complained about them.

The CRAN mirror on GitHub shows many packages on CRAN which list `{parallel}` in their `Imports`: https://github.com/search?p=2&q=org%3Acran+filename%3ADESCRIPTION+%27parallel%2C%27&type=Code.

Including packages written by R-core members, like `{iotools}` ([link](https://github.com/cran/iotools/blob/714368d3f0785d5e3e6b16d174c80cc4cc57be70/DESCRIPTION#L7)).

`@`-ing other maintainers so they're aware of this for the purpose of future PR reviews: @jmoralez @StrikerRUS @guolinke @shiyu1994

R-package/tests/testthat/test_Predictor.R
jameslamb: At https://cran.r-project.org/web/packages/policies.html, I see the following

> If running a package uses multiple threads/cores it must never use more than two simultaneously: the check farm is a shared resource and will typically be running many checks simultaneously.

Thanks for bringing it to our attention. If you're able to share links to specific documentation for such claims in the future, it would save maintainers some time and be really appreciated.

I'm open to limiting all of the examples / tests to 2 threads, but don't believe it needs to be done in this PR. I've documented the work to limit the number of threads used in #5102, and will open a separate PR doing that in a way that works with LightGBM's CI process.

Please remove these changes where you are passing `num_threads = 1L` from this PR, and limit this PR to only exposing `num_threads` as a keyword argument to `lightgbm::lightgbm()`.

R-package/R/lightgbm.R
jameslamb: Can you tell us *why* you think that function (which would mean adding a new dependency to `{lightgbm}`) is preferable to setting this to `0L` and deferring to LightGBM's C++ code to pick a default?

david-cortes: Because it ends up picking the actual number of physical cores even in setups in which `detectCores` would get it incorrectly, and never returns `NA`. I guess that the most typical scenario is to want to parallelize as much as possible, and in such case it'd be better to have the number of physical cores as the default and auto-detected correctly.

jameslamb: Got it, thanks for that. I'm nervous about taking on a new dependency just for the purpose of setting this default, but I do think it's valuable to have `lightgbm::lightgbm()` by default choose the same `num_threads` as you'd find recommended in the documentation or issues page.

And since `{RhpcBLASctl}` is on CRAN (https://cran.r-project.org/web/packages/RhpcBLASctl/index.html), we at least know it meets CRAN's rigorous standards for portability.

I'm willing to try this. To avoid imposing a new *required* dependency on users of the R package, I'm suggesting the following:

* make `num_threads = NULL` the default in `lightgbm::lightgbm()`
* add a function in https://github.com/microsoft/LightGBM/blob/master/R-package/R/utils.R like the following:
    - ```r
       .suggest_default_num_threads <- function(){
           if (require("RhcpBLASctl")) {
              return(RhcpBLASctl::get_num_cores())
          } else {
              warning("Optional package RhcpBLASctl not found. Consider
          }
          return(0L)
       }
       ```
* in `lightgbm::lightgbm()`, have code like the following
    - ```r
       if (is.null(num_threads)) {
           num_threads <- .suggest_default_num_threads()
       }
       params <- lgb.check.wrapper_param(
          main_param_name = "num_threads"
          , params = params
          , alternative_kwarg_value = num_threads
        )
       ```
* add `{RhcpBLASctl}` as a `Suggests`-level import in https://github.com/microsoft/LightGBM/blob/master/R-package/DESCRIPTION

This would avoid adding a new required dependency, and would mean that the behavior of passing `num_threads = -1` or `num_threads = 0` through keyword arg `num_threads` would have the exact same meaning, the one documented inhttps://lightgbm.readthedocs.io/en/latest/Parameters.html#num_threads and described in my comments above in this thread.

That would also encapsulate this logic so that it could optionally be re-used in the future in other parts of the package (like in setting default `num_threads` for `Dataset` construction).

david-cortes: But that would still require you to put that package under `Suggests`, and if something were to go wrong with that package on CRAN, you'd get the same kinds of issues as if it were a dependency.

Having the value wrapped like that is IMO less transparent to the user and less intuitive to see in a function signature.

jameslamb: > and if something were to go wrong with that package on CRAN, you'd get the same kinds of issues as if it were a dependency

Sure, *on CRAN* a `Suggests` dependency imposes the same risks as an `Imports` or `Depends` one since CRAN will install all `Suggests` dependencies when running `R CMD check`.

But putting that into `Suggests` means that for users of the package, `install.packages("lightgbm")` won't require the installation of `{RhcpBLASctl}`, so `{RhcpBLASctl}` being incompatible with their environments wouldn't prevent the use of `{lightgbm}`.

I'm willing to take on a bit more maintenance burden (increased risk of CRAN failures caused by a dependency) in exchange for a better user experience with the package's default values, but I'm not willing to add a new unconditional dependency to `{lightgbm}` just to set a better default value for `num_threads`.

david-cortes:
Updated, with a fallback to `parallel::detectCores` when appropriate.

jameslamb (changes requested):
Thanks very much for the changes, and the idea to fall back to `parallel::detectCores()` if `RhcpBLASctl::get_num_cores()` is not available.

* please remove all the added `num_threads = 1` in docs / examples /tests not otherwise touched by this PR, per the request in https://github.com/microsoft/LightGBM/pull/4972#discussion_r838043521
* please preserve the alphabetical ordering of imports in `DESCRIPTION`
    - `{RhpcBLASctl}` should be between `{processx}` and `{rmarkdown}`
    - `{parallel}` should be between `{methods}` and `{utils}`

jameslamb:
regarding this linter warning ([build link](https://github.com/microsoft/LightGBM/runs/5760533651?check_suite_focus=true))

> /home/runner/work/LightGBM/LightGBM/R-package/R/utils.R:223:7: warning: Function "require" is undesirable. As an alternative, library() is preferred to require() because it will raise an error immediately if a package is missing..

Please suppress it with a `# nolint` comment on the line using `require()`.

david-cortes:
> Thanks very much for the changes, and the idea to fall back to `parallel::detectCores()` if `RhcpBLASctl::get_num_cores()` is not available.
> 
> * please remove all the added `num_threads = 1` in docs / examples /tests not otherwise touched by this PR, per the request in [#4972 (comment)](https://github.com/microsoft/LightGBM/pull/4972#discussion_r838043521)
> * please preserve the alphabetical ordering of imports in `DESCRIPTION`
>   
>   * `{RhpcBLASctl}` should be between `{processx}` and `{rmarkdown}`
>   * `{parallel}` should be between `{methods}` and `{utils}`

Aren't capitalized letters supposed to come before small-case letters?

jameslamb:
> Aren't capitalized letters supposed to come before small-case letters?

maybe in some types of sorting, but alphabetizing that list is mainly intended to benefit developers on this project visually inspecting that file, and I don't always remember if, for example, `{rmarkdown}` begins with a capital or lowercase "r".

david-cortes:
Updated.

jameslamb:
> Looks like RhpcBLASctl needs to be added somewhere in the CI configuration.

Yes please. I think these are all the places:

https://github.com/microsoft/LightGBM/blob/0688f471fc303af816cd6564107f6d720ebb99ec/.ci/test_r_package_windows.ps1#L125

https://github.com/microsoft/LightGBM/blob/eb686a7658ea5af87dd5bf96cd547e8197a69ab4/.github/workflows/r_package.yml#L191

https://github.com/microsoft/LightGBM/blob/cb8c61e4d3ac08b3d2c65c91bde483397f861cd3/.github/workflows/static_analysis.yml#L60

https://github.com/microsoft/LightGBM/blob/5fa887bb79d0e4d1961c8adafce9832448535e71/.ci/test_r_package_solaris.sh#L10

https://github.com/microsoft/LightGBM/blob/0688f471fc303af816cd6564107f6d720ebb99ec/.ci/test_r_package.sh#L114

https://github.com/microsoft/LightGBM/blob/5fa887bb79d0e4d1961c8adafce9832448535e71/.ci/test_r_package_valgrind.sh#L3

https://github.com/microsoft/LightGBM/blob/d4cdbcfe1b1db2d48e95665a8b358028c496cdb2/.vsts-ci.yml#L316

R-package/R/utils.R
```diff
@@ -217,3 +217,27 @@ lgb.check.wrapper_param <- function(main_param_name, params, alternative_kwarg_v
   params[[main_param_name]] <- alternative_kwarg_value
   return(params)
 }
+
+#' @importFrom parallel detectCores
+lgb.get.default.num.threads <- function() {
+  # nolint
+  if (require("RhpcBLASctl", quietly = TRUE)) {
```
jameslamb, line 224:
```suggestion
  if (require("RhpcBLASctl", quietly = TRUE)) {. # nolint
```

The `# nolint` needs to be on this line for `{lintr}` to respect it.

R-package/R/utils.R
```diff
@@ -217,3 +217,27 @@ lgb.check.wrapper_param <- function(main_param_name, params, alternative_kwarg_v
   params[[main_param_name]] <- alternative_kwarg_value
   return(params)
 }
+
+#' @importFrom parallel detectCores
+lgb.get.default.num.threads <- function() {
+  # nolint
```
jameslamb, line 223:
```suggestion
```

R-package/tests/testthat/test_basic.R
```diff
@@ -189,6 +189,7 @@ test_that("lightgbm() rejects negative or 0 value passed to nrounds", {
         data = dtrain
         , params = params
         , nrounds = nround_value
+        , save_name = tempfile(fileext = ".model")
```
jameslamb, line 192: This change looks unrelated to this PR. Can you please explain why this was necessary? If it's not related to this PR, please remove it.

jameslamb (changes requested):
I left some suggestions for how to get the R linter to pass, and a suggestion to please remove unrelated changes from this PR.

david-cortes:
Updated.

jameslamb (approved):
Looks great to me! Thanks very much for your help with this, and with the corresponding changes to the Python package in #5105.

PR #4973, opened 2022-01-23
[R-package] Rename `data` -> `newdata` in `predict`

david-cortes:
ref #4968

Typically, the prediction function for model objects in R gets passed the prediction data as an argument named `newdata`, while this package instead names it `data`. All of base R's prediction functions (for example from the `stats` package which is part of base R and around which most other alike packages are modeled), such as `predict.lm` or `predict.glm`, name the argument `newdata`.

From a quick look at popular decision tree packages, most of them also name their argument as `newdata` - examples off the top of m mind that I searched for: `xgboost`, `randomForest`, `gbm`, `C5.0`, `rpart`, `randomForestSRC`. The only outlier out there is `ranger` which does not name it `newdata`. I guess the ratio of `newdata` to something else would grow considerably higher if packages for linear models were considered.

This PR renames the argument to `newdata` to be more in line with other packages that R users are likely to be familiar with.


jameslamb (approved):
Thanks very much for this proposal and for the list of examples!

I did some research tonight and agree with your proposal... even though `stats::predict()` is not opinionated about the name of the keyword argument referring to "data to create predictions from" in S3 methods that extend it, `newdata` is the most popular name used.

In addition to the packages mentioned in this PR's description, I also looked at a few other packages in the CRAN Machine Learning Task View ([link](https://cran.r-project.org/web/views/MachineLearning.html)).

| package | signature of `predict()` S3 method | code link   |
|:--------:|-----------------------------------:|:-----:|
| `{arules}` | `predict.itemMatrix(object, newdata, ...)` | [link](https://github.com/mhahsler/arules/blob/dfe0a7b7a0d68be5aa4c4ee4a1cac2c574d3270b/R/predict.R#L66-L67) |
| `{caret}` | `predict.train(object, newdata, ...)`        | [link](https://github.com/topepo/caret/blob/be921481bcdd5492beee4f8a59ae57e69a2c8047/pkg/caret/R/predict.train.R#L109) |
| `{catboost}` | `catboost.caret$predict(modelFit, newdata, ...)` | [link](https://github.com/catboost/catboost/blob/7d293bfb103d3891f1839ccb7d799a8e6d4a36ad/catboost/R-package/R/catboost.caret.R#L106) |
| `{C5.0}` | `predict.C5.0(object, newdata, ...)` | [link](https://github.com/topepo/C5.0/blob/807161d42af421ae6700e4b608d8b65edc9c8a08/R/predict.C5.0.R#L71-L73) |
| `{gbm}` | `predict.gbm(object, newdata, ...)` | [link](https://github.com/gbm-developers/gbm/blob/18e595597e19c289328f5eb0b73945a1ea72fc7d/R/predict.gbm.R#L5) |
| `{mboost}` | `predict.mboost(object, newdata, ...)` | [link](https://github.com/cran/mboost/blob/128a48aa7fc77ec6419ef05f0b29789337fc6fa0/R/methods.R#L25) |
| `{randomForest}` | `predict.randomForest(object, newdata, ...)` | [link](https://github.com/cran/randomForest/blob/6cedb5fbf26ca857afda687f42c5ba535d61eaf0/R/predict.randomForest.R#L2) |
| `{rpart}` | `predict.rpart(object, newdata, ...)` | [link](https://github.com/cran/rpart/blob/9994543c62ae212b3cd582c50985e0748f8ead1b/R/predict.rpart.R#L1) |
| `{stats}` | `predict.glm(object, newdata = NULL, ...)` | [link](https://github.com/wch/r-source/blob/79298c499218846d14500255efd622b5021c10ec/src/library/stats/R/predict.glm.R#L20) |
| `{stats}` | `predict.lm(object, newdata, ...)` | [link](https://github.com/wch/r-source/blob/79298c499218846d14500255efd622b5021c10ec/src/library/stats/R/lm.R#L667) | `{xgboost}` | `predict.xgb.Booster(object, newdata, ...)` | [link](https://github.com/dmlc/xgboost/blob/18a4af63aa2f99f5941fe36cb147829d68aa42d1/R-package/R/xgb.Booster.R#L327) |

Examples of other patterns:

| package | signature of `predict()` S3 method | code link   |
|:--------:|-----------------------------------:|:-----:|
| `{catboost}` | `catboost.predict(model, pool, ...)` | [link](https://github.com/catboost/catboost/blob/7d293bfb103d3891f1839ccb7d799a8e6d4a36ad/catboost/R-package/R/catboost.R#L1833) |
| `{elasticnet}` | `predict.ent(model, newx, ...)` | [link](https://github.com/cran/elasticnet/blob/42e1465dfd6298b2d099852a1b2b7991ee054bd9/R/enet_funcs.R#L193-L194) | 
| `{lars}` | `predict.lars(object, newx, ...)` | [link](https://github.com/cran/lars/blob/5b6af0bcd469ee7fc9cffb4d87594fbec84e4a8c/R/predict.lars.R#L2) |
| `{lightgbm}` | `predict.lgb.Booster(object, data, ...)` | [link](https://github.com/microsoft/LightGBM/blob/5631366175f68a28bd174982cca16975f7fb48af/R-package/R/lgb.Booster.R#L781-L782) |
| `{parsnip}` | `predict.model_fit(object, new_data, ...)` | [link](https://github.com/tidymodels/parsnip/blob/3e2447c1bad3e5c056d804ad45646ca129d11c87/R/predict.R#L134) |
| `{RGF}` | a `$predict(x)` method on an R6  object | [link](https://github.com/RGF-team/rgf/blob/effad8bbcc6820c937bda95ea5bb68855b696dc7/R-package/tests/testthat/test-RGF_package.R#L54) |

<hr>

Given that...I support changing this argument name in `predict.lgb.Booster()`, as you've proposed. And I'm ok with this being a breaking change from the perspective of anyone using keyword arguments instead of positional arguments. Since `newdata` doesn't have a default argument, any existing code like (`predict(bst, data = X)`) will break with a loud error like `argument "newdata" is missing, with no default`.

Normally I'd say this kind of breaking change should go through a round of deprecation warnings, but support making it now for the same reasons mentioned in https://github.com/microsoft/LightGBM/pull/5007#issue-1138103040.

@StrikerRUS or @Laurae2 at your earliest convenience, could you give your opinion on whether or not we should merge this breaking change as-in?

StrikerRUS:
@jameslamb 
> could you give your opinion on whether or not we should merge this breaking change **as-in**?

Sorry, could you please clarify which other possible ways we have? Did you mean 
> a round of deprecation warnings

by this?

StrikerRUS (approved):
Provided examples make me think `newdata` is a kind of "stardard" name in the R-lang world, thus I support this renaming.

jameslamb:
@StrikerRUS I'm sorry for not being clear. I meant:

**Option 1:** Just change second argument name from `data` to `newdata`.

**Option 2:** Change second argument name from `data` to `newdata`, add `data = NULL` at the end of the signature, and if any non-NULL value is passed to `data`, use that value but raise a deprecation warning. Then in a later release (like 4.1.x), remove support for keyword argument `data` completely.

Which has the following implications:

```r
# will not break with either optio
preds <- predict(booster, X)

# Option 1: would fail, with error 'argument "newdata" is missing, with no default'
# Option 2: would succeed, with warning like 'using "data" as a keyword argument is deprecated, use "newdata"'
preds <- predict(booster, data = X)
```

Since `newdata` seems like a widely-accepted standard and since this project hasn't been doing releases very often and seems like it might still be a long time until 4.0, I support  **Option 1** because of the reduced maintenance burden and because the breaking change results in a loud error that will be easy for users to understand and react to.

StrikerRUS:
@jameslamb Thanks for the extremely detailed explanation!

> I support **Option 1**

Me too.

StrikerRUS:
@david-cortes Sorry for the inconvenience, but could you please sync this PR with `master` branch to pass CI tests?

david-cortes:
Updated.

jameslamb:
Closing and re-opening to re-trigger CI. I'm not sure why several of the CI builds failed (across different operating systems and with different errors).

PR #4974, opened 2022-01-23
[R-package] remove behavior where lightgbm() saves model to disk

david-cortes:
ref #4968

The `lightgbm()` function defaults towards dumping the model to disk in a file named `lightgbm.model`. After the last series of changes to the serialization logic, this is perhaps not a reasonable default anymore, since now model objects can be serialized natively and will for example survive session restarts in RStudio and similar, so a user fitting a model would not be prone to inadvertently lose it nor need to use this format for model serialization, thus decreasing the need for having a file in disk.

Among modeling packages, it is rather uncommon to produce a file in disk by default - off the top of my head, the only such R libraries that I can think of are `xgboost` and `catboost`, while most other packages (which rely on Rds for serialization) do not even have the option of dumping in the function call.

### DIFF
--- a/R-package/R/lightgbm.R
+++ b/R-package/R/lightgbm.R
@@ -92,8 +92,6 @@ NULL
 #' @inheritParams lgb_shared_params
 #' @param label Vector of labels, used if \code{data} is not an \code{\link{lgb.Dataset}}
 #' @param weight vector of response values. If not NULL, will set to dataset
-#' @param save_name File name to use when writing the trained model to disk. Should end in ".model".
-#'                  If passing `NULL`, will not save the trained model to disk.
 #' @param ... Additional arguments passed to \code{\link{lgb.train}}. For example
 #'     \itemize{
 #'        \item{\code{valids}: a list of \code{lgb.Dataset} objects, used for validation}
@@ -120,7 +118,6 @@ lightgbm <- function(data,
                      verbose = 1L,
                      eval_freq = 1L,
                      early_stopping_rounds = NULL,
-                     save_name = NULL,
                      init_model = NULL,
                      callbacks = list(),
                      serializable = TRUE,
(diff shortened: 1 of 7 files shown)


R-package/R/lightgbm.R
```diff
@@ -120,7 +120,7 @@ lightgbm <- function(data,
                      verbose = 1L,
                      eval_freq = 1L,
                      early_stopping_rounds = NULL,
-                     save_name = "lightgbm.model",
+                     save_name = NULL,
```
jameslamb, line 123: This change would break existing code that relies on this behavior _silently_. Code relying on this behavior might fail with a "file not found" error after `lightgbm::lightgbm()` has been called, resulting in possibly-significant time spent debugging to figure out that `{lightgbm}`'s default behavior has changed.

I agree with the argument in the pull request description that model `lightgbm::lightgbm()` shouldn't be responsible for persisting models to disk. I support just removing the argument `save_name` from `lightgbm()` entirely.  I think we should make that change right now, in this PR, without waiting for a round of deprecation warnings, because:

1.  it's already been almost 4 months since the most recent LightGBM release ([v3.3.1](https://github.com/microsoft/LightGBM/releases/tag/v3.3.1), excluding the special v3.3.2 that only had one small patch) and I expect it to be at least 2 months until the next, 4.0.0, release
2. this breaking change would result in a loud training-time error, which I consider much less expensive than than a silent breaking change or than one that happens at scoring time with a deployed model

@StrikerRUS @Laurae2 , what do you think?

jameslamb (changes requested):
Thanks for this proposal! I left one comment, please wait for another maintainer to respond to me before making any changes.

R-package/R/lightgbm.R
StrikerRUS: I agree with both points above. Thanks a lot for the detailed explanation!

david-cortes: Removed.

jameslamb (approved):
Thanks very much! I'm really glad to see this simplification.

StrikerRUS:
I'm going to close and instantly re-open this PR to re-trigger Microsoft CLA bot that stuck. Sorry for the inconvenience!

PR #4976, opened 2022-01-23
[R-package] Promote objective and init_score to top-level arguments in `lightgbm()`

david-cortes:
ref #4968

This PR promotes `objective` and `init_score` to top-level parameters in `lightgbm()`. This is for 3 reasons:
* `objective` is probably one of the most common parameters to want to change. It's also accepted as a top-level parameter in the `lgb.train` interface and in the scikit-learn interface.
* Having `objective` as a top-level parameter will allow an option `"auto"` in the future (as part of #4968) to adjust it automatically for binary and multi-class classification depending on the input.
* Having `init_score` as a top-level parameter will allow to take it as a column from the input data in the future (as part of #4968), even if it right now looks out of place.

### DIFF
--- a/R-package/R/lightgbm.R
+++ b/R-package/R/lightgbm.R
@@ -92,6 +92,13 @@ NULL
 #' @inheritParams lgb_shared_params
 #' @param label Vector of labels, used if \code{data} is not an \code{\link{lgb.Dataset}}
 #' @param weight vector of response values. If not NULL, will set to dataset
+#' @param init_score Initial values for each observation from which the boosting process will
+#'                   be started (e.g. as the result of some previous model). If not passing it (the default),
+#'                   will start from a blank state.
+#' @param objective Optimization objective (e.g. `"regression"`, `"binary"`, etc.).
+#'                  For a list of accepted objectives, see
+#'                  \href{https://lightgbm.readthedocs.io/en/latest/Parameters.html}{
+#'                  the "Parameters" section of the documentation}.
 #' @param save_name File name to use when writing the trained model to disk. Should end in ".model".
 #'                  If passing `NULL`, will not save the trained model to disk.
 #' @param ... Additional arguments passed to \code{\link{lgb.train}}. For example
@@ -115,6 +122,8 @@ NULL
 lightgbm <- function(data,
                      label = NULL,
                      weight = NULL,
+                     init_score = NULL,
+                     objective = "regression",
                      params = list(),
                      nrounds = 100L,
                      verbose = 1L,
@@ -136,13 +145,14 @@ lightgbm <- function(data,
 
   # Check whether data is lgb.Dataset, if not then create lgb.Dataset manually
   if (!lgb.is.Dataset(x = dtrain)) {
-    dtrain <- lgb.Dataset(data = data, label = label, weight = weight)
+    dtrain <- lgb.Dataset(data = data, label = label, weight = weight, init_score = init_score)
   }
 
   train_args <- list(
     "params" = params
     , "data" = dtrain
     , "nrounds" = nrounds
+    , "obj" = objective
     , "verbose" = verbose
     , "eval_freq" = eval_freq
     , "early_stopping_rounds" = early_stopping_rounds
--- a/R-package/tests/testthat/test_basic.R
+++ b/R-package/tests/testthat/test_basic.R
@@ -74,9 +74,9 @@ test_that("train and predict binary classification", {
   bst <- lightgbm(
     data = train$data
     , label = train$label
+    , objective = "binary"
     , params = list(
         num_leaves = 5L
-        , objective = "binary"
         , metric = "binary_error"
     )
     , nrounds = nrounds
@@ -104,12 +104,12 @@ test_that("train and predict softmax", {
   bst <- lightgbm(
     data = as.matrix(iris[, -5L])
     , label = lb
+    , objective = "multiclass"
     , params = list(
         num_leaves = 4L
         , learning_rate = 0.05
         , min_data = 20L
         , min_hessian = 10.0
-        , objective = "multiclass"
         , metric = "multi_error"
         , num_class = 3L
     )
@@ -131,10 +131,10 @@ test_that("use of multiple eval metrics works", {
   bst <- lightgbm(
     data = train$data
     , label = train$label
+    , objective = "binary"
     , params = list(
         num_leaves = 4L
         , learning_rate = 1.0
-        , objective = "binary"
         , metric = metrics
     )
     , nrounds = 10L
@@ -155,9 +155,9 @@ test_that("lgb.Booster.upper_bound() and lgb.Booster.lower_bound() work as expec
   bst <- lightgbm(
     data = train$data
     , label = train$label
+    , objective = "binary"
     , params = list(
         num_leaves = 5L
-        , objective = "binary"
         , metric = "binary_error"
     )
     , nrounds = nrounds
@@ -173,9 +173,9 @@ test_that("lgb.Booster.upper_bound() and lgb.Booster.lower_bound() work as expec
   bst <- lightgbm(
     data = train$data
     , label = train$label
+    , objective = "regression"
     , params = list(
         num_leaves = 5L
-        , objective = "regression"
         , metric = "l2"
     )
     , nrounds = nrounds
@@ -192,6 +192,7 @@ test_that("lightgbm() rejects negative or 0 value passed to nrounds", {
     expect_error({
       bst <- lightgbm(
         data = dtrain
+        , objective = "regression"
         , params = params
         , nrounds = nround_value
         , save_name = tempfile(fileext = ".model")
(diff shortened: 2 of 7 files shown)


R-package/R/lightgbm.R
```diff
@@ -92,6 +92,13 @@ NULL
 #' @inheritParams lgb_shared_params
 #' @param label Vector of labels, used if \code{data} is not an \code{\link{lgb.Dataset}}
 #' @param weight vector of response values. If not NULL, will set to dataset
+#' @param init_score Initial values for each observation from which the boosting process will
```
jameslamb, line 95: Please add a new test that passing `init_score` to `lightgbm::lightgbm()` works as expected. Any user-facing changes to behavior should be accompanied by unit tests to ensure that the functionality is working as expected and to reduce the risk that future changes accidentally break that functionality.

jameslamb (changes requested):
Thanks for this. I support adding `init_score` and `objective`  as keyword arguments in `lightgbm()`. Please see the requested changes.

david-cortes:
Updated.

R-package/R/lightgbm.R
```diff
@@ -92,6 +92,11 @@ NULL
 #' @inheritParams lgb_shared_params
 #' @param label Vector of labels, used if \code{data} is not an \code{\link{lgb.Dataset}}
 #' @param weight vector of response values. If not NULL, will set to dataset
+#' @param init_score initial score is the base prediction lightgbm will boost from
```
jameslamb, line 95: Can you please re-arrange these so the order of the parameters in the roxygen comments matches the function signature? Specifically, `objective` should come before `init_score`, and both should be after `save_name`.

R-package/R/lightgbm.R
```diff
@@ -136,13 +143,24 @@ lightgbm <- function(data,
...
+
+  if (length(intersect(names(params), .PARAMETER_ALIASES()[["objective"]]))) {
+    obj_names <- .PARAMETER_ALIASES()[["objective"]]
+    for (obj_name in obj_names) {
+      if (!is.null(params[[obj_name]])) {
+        objective <- params[[obj_name]]
+        break
+      }
+    }
   }
```
jameslamb, line 157: Why is this code necessary? It shouldn't be, since `lgb.train()` already resolves parameter aliases for `objective` 

https://github.com/microsoft/LightGBM/blob/15266f1d613a37fc6d59bc0462854a99eeee3b0b/R-package/R/lgb.train.R#L95-L99

and ensures that entries in `params` take precedence over the keyword argument

https://github.com/microsoft/LightGBM/blob/15266f1d613a37fc6d59bc0462854a99eeee3b0b/R-package/R/lgb.train.R#L107-L113

R-package/tests/testthat/test_basic.R
```diff
@@ -2758,3 +2758,68 @@ for (x3_to_categorical in c(TRUE, FALSE)) {
     })
   }
 }
+
+test_that("lightgbm() accepts objective as function argument and under params", {
```
jameslamb, line 2762: Thanks for these tests!

Can you please add one more, `lightgbm() defaults to "regression" objective if objective not provided as function argument or under params`?

Since this PR is proposing adding a default for `objective`, I'd like to have a test explicitly capturing that behavior. User-facing changes to the package's behavior should be accompanied by tests, so that we can automatically catch the case where future proposed changes break existing behavior.

R-package/vignettes/basic_walkthrough.Rmd
```diff
@@ -62,10 +62,10 @@ X <- data.matrix(bank[, c("age", "balance")])
 fit <- lightgbm(
   data = X
   , label = y
+  , objective = "binary"
```
jameslamb, line 65: Please revert this vignette change. I support making is possible to pass `objective` as a keyword argument, but not recommending it as the preferred way to specify that parameter.

jameslamb (changes requested):
Thanks very much for the tests! Please see my most recent requested changes.

R-package/R/lightgbm.R
david-cortes: Those get rearranged automatically by roxygen. Rearranging them there doesn't guarantee that they'll appear in the same order in the final Rd.

R-package/R/lightgbm.R
david-cortes: If that's how `lgb.train` is supposed to work, then it is broken there too.

R-package/tests/testthat/test_basic.R
david-cortes: Will add it, but the easy solution in such case would be to not have a default value so that the user would always need to supply it. That's, of course, a breaking change.

R-package/vignettes/basic_walkthrough.Rmd
david-cortes: Reverted.

R-package/R/lightgbm.R
jameslamb: I'm glad `{roxygen2}` does that, but even just for the benefit of contributors modifying this code, I'd prefer to keep the documentation in the same order as the function signature.

R-package/R/lightgbm.R
jameslamb: Ah you're right!

I just installed the R package from source on the latest `master` (https://github.com/microsoft/LightGBM/commit/2d1caf140a3b0c68f908071277b67c2a842a297b).

```shell
R CMD INSTALL ./lightgbm_3.3.2.99.tar.gz
```

Then tested this as follows.

```r
library(lightgbm)

data("EuStockMarkets")

bst <- lightgbm::lgb.train(
    data = lightgbm::lgb.Dataset(
        data = EuStockMarkets[, c("SMI", "CAC", "FTSE")]
        , label = EuStockMarkets[, "DAX"]
    )
    , params = list(objective = "regression_l2")
    , nrounds = 5L
    , obj = "regression_l1"
)
print(bst$params$objective)
```

I'll open a PR to fix that.

Such things should be handled in `lgb.train()`, not `lightgbm()`. Thanks for bringing it to our attention.

jameslamb: created #5007

R-package/R/lightgbm.R
david-cortes: Moved.

R-package/R/lightgbm.R
david-cortes: Removed. But it will require merging the other PR first in order to pass the tests.

R-package/R/lightgbm.R
jameslamb: thank you for moving these, but they are still out of order. `objective` comes before `init_score` in the function signature, but the docs for `init_score` come before `objective. Please make the orders consistent.

R-package/tests/testthat/test_basic.R
```diff
@@ -2780,3 +2780,78 @@ for (x3_to_categorical in c(TRUE, FALSE)) {
...
+
+  bst2 <- lightgbm(
+    data = train$data
+    , label = train$label
+    , objective = "regression"
+    , params = list(loss = "regression_l1")
+    , nrounds = 5L
+    , verbose = -1L
+  )
+  expect_equal(bst2$params$objective, "regression_l1")
```
jameslamb, line 2823: As @StrikerRUS pointed out to me in https://github.com/microsoft/LightGBM/pull/5007#discussion_r807440458, just testing the value in the attribute `$params$objective` in the `Booster`  object in R doesn't fully test that the expected objective is actually passed through to training.

For all tests where you're testing the value of `$params$objective`, can you please also add checks like those I added in #5007?

https://github.com/microsoft/LightGBM/blob/2f27d4b226815016402f4633876790663a0ed8b5/R-package/tests/testthat/test_basic.R#L703-L710

jameslamb (changes requested):
Thanks very much! I think this is almost ready to merge, just a few more recommended changes.

david-cortes:
Updated.

R-package/tests/testthat/test_basic.R
```diff
@@ -2780,3 +2780,102 @@ for (x3_to_categorical in c(TRUE, FALSE)) {
...
+})
+
+test_that("lightgbm() defaults to 'regression' objective if objective not otherwise provided", {
+  bst <- lightgbm(
+    data = train$data
+    , label = train$label
+    , nrounds = 5L
+    , verbose = -1L
+  )
+  expect_equal(bst$params$objective, "regression")
```
jameslamb, line 2880: Please update this test based on https://github.com/microsoft/LightGBM/pull/4976/files#r810442651 as well, to test that actual objective used by the `Booster`.

jameslamb (changes requested):
thanks! Just one more suggestion and this can be merged.

R-package/tests/testthat/test_basic.R
david-cortes: Updated.

R-package/tests/testthat/test_basic.R
```diff
@@ -2812,3 +2812,109 @@ for (x3_to_categorical in c(TRUE, FALSE)) {
...
+  expect_equal(bst$params$objective, "regression")
```
jameslamb, line 2913:
```suggestion
```

jameslamb (approved):
I just updated this with the latest changes from `master`, now that #5021 has been merged.

I'll merge this once it builds. Thanks very much for the contribution!

PR #4977, opened 2022-01-23
[R-package] Keep row names in output from `predict`

david-cortes:
ref #4968

Calling `predict` on a lightgbm model object will produce a result without row names. If the data on which predictions are being made has row names, chances are that the user will want to keep those in the prediction output. This PR keeps the row names in the prediction output when the input data has them.


david-cortes:
Failing CI checks on mac are due to a dependency that fails to install and not related to the changes in this PR.

jameslamb (changes requested):
Thanks very much for this proposal.

I don't support adding this feature. I'm not convinced that the (in my opinion) slight convenience is worth the additional code complexity and maintenance burden.

I also want to be especially protective of `predict()` code paths, since I expect user code creating predictions to be much more latency-sensitive than training code. (related conversation: https://github.com/microsoft/LightGBM/pull/4909#pullrequestreview-842363693).

We can get some other opinions though. @mayer79 @StrikerRUS @Laurae2 if any of you are interested, could you give your opinion on this proposed change?

david-cortes:
The latency added by this feature is around 30 microseconds in my setup (~40 when objects have names), which is very low compared to the time it takes to make a single-row prediction which can reach up to a few miliseconds depending on the choice of hyperparameters and arguments (and lower than the overhead added from syntax-related coding choices such as usage of R6).

StrikerRUS:
I'm sorry, I very far away from the R-lang world, so can't write thoughtful comment here. If keeping row names is something like a standard behavior for other R packages, then we can add this feature. If not, then I agree with @jameslamb .

david-cortes:
Row names are kept in base R and most recommended/core packages for decision trees (e.g. `rpart`, `randomForest`, etc.).

Did a bit of testing with popular non-core packages, and found that many decision-tree packages do not keep them (e.g. `ranger`, `randomForestSRC`), although their output is not exactly a numeric  vector/matrix in the same way as in `lightgbm` or base R.

OTOH tried a few packages for linear models off the top of my mind, and all of them kept the row names (`glmnet`, `picasso`, `ncvreg`, `grpreg`, `glmx`, `VGAM`, `lme4`).

jameslamb:
Sorry for the delayed response, I recently returned from traveling.

Thanks for checking other packages! Since projects like `{randomForest}`, `{rpart}`, `{stats}`, and the linear model projects you mentioned support this functionality, I'm convinced that it's standard enough that `{lightgbm}` should support it too.

However, `predict()` in `{lightgbm}` is more complex than most of those other packages...it doesn't only produce predicted values of the target, but is also used (for example) to get feature contributions or information about which leaves records match.

I'd only feel confident taking on this behavior if unit tests were added confirming that row names are handled correctly for every combination of the following:

* row names in `newdata` (has rownames, does not have rownames)
* `reshape` (`TRUE` or `FALSE`)
* `newdata` (matrix, `dgCMatrix`, or filepath to a CSV)
* task in the model (regression, binary classification, multiclass classification)
    - this is important because the output shape is different for multiclass classification
    - https://github.com/microsoft/LightGBM/blob/9a4e70687d5c0732ca895959f418c3f923f2e85a/R-package/R/lgb.Booster.R#L739-L745
* type of output requested (`rawscore`, `predleaf`, `predcontrib`, none of those)

If you have time and interest in adding such tests, I'd support this PR.

If not, then I think this PR should be closed and we could convert it to a feature request to be added to #2302 and picked up at some point in the future.

david-cortes:
Added more tests, but left out CSV tests because predicting on them is currently broken.

david-cortes:
The failing checks are due to the vignette. It fails to make a prediction at L75:
```r
summary(predict(fit, X))
```
```
Error in numeric(npred) : vector size cannot be NA
```
Probably coming from here:
https://github.com/microsoft/LightGBM/blob/6b56a90cd1324a6dbac2afa0a352c9355b0dc3cf/R-package/R/lgb.Predictor.R#L151

It is not related to this PR since this error happens before reaching the code that was changed.

david-cortes:
> The failing checks are due to the vignette. It fails to make a prediction at L75:
> 
> ```r
> summary(predict(fit, X))
> ```
> 
> ```
> Error in numeric(npred) : vector size cannot be NA
> ```
> 
> Probably coming from here:
> 
> https://github.com/microsoft/LightGBM/blob/6b56a90cd1324a6dbac2afa0a352c9355b0dc3cf/R-package/R/lgb.Predictor.R#L151
> 
> It is not related to this PR since the error happens before reaching the code that was changed.

Nevermind this comment, there was another change I didn't notice in the merge conflict.

R-package/tests/testthat/test_Predictor.R
```diff
@@ -111,3 +113,117 @@ test_that("start_iteration works correctly", {
     pred_leaf2 <- predict(bst, test$data, start_iteration = 0L, num_iteration = end_iter + 1L, predleaf = TRUE)
     expect_equal(pred_leaf1, pred_leaf2)
 })
+
+test_that("predictions keep row names from the data", {
```
jameslamb, line 117: these tests look great, thank you! To speed up debugging if they break due to changes in future PRs, can you please break them into 3 test cases?

* "predict() keeps row names from data (binary classification)"
* "predict() keeps row names from data (multi-class classification)"
* "predict() keeps row names from data (regression)"

To avoid duplicating test helper code and to make it clear that those methods were written just for these tests, please move them out of the test case and rename them to `.expect_has_row_names()` and `.expect_doesnt_have_row_names()`.

R-package/tests/testthat/test_Predictor.R
```diff
@@ -111,3 +113,117 @@ test_that("start_iteration works correctly", {
...
+    data("mtcars")
+    X <- as.matrix(mtcars[, -1L])
+    y <- as.numeric(mtcars[, 1L])
+    dtrain <- lgb.Dataset(X, label = y, params = list(max_bins = 5L))
+    bst <- lgb.train(
+        data = dtrain
+        , obj = "regression"
+        , nrounds = 5L
+        , verbose = -1L
```
jameslamb, line 126:
```suggestion
        , verbose = VERBOSITY
```

Please use the constant defined at the top of this file to set verbosity.

https://github.com/microsoft/LightGBM/blob/17d4e007852b4f0b34b211a674cb26e6751c98e2/R-package/tests/testthat/test_Predictor.R#L1-L3

Per #4862, we're slowly working on reducing the logs produced by these tests, to make it easier to visually inspect test results. That code I've linked allows easily running the tests with different verbosity levels by passing an environment variables.

R-package/tests/testthat/test_Predictor.R
```diff
@@ -111,3 +113,117 @@ test_that("start_iteration works correctly", {
...
+    y <- as.numeric(iris$Species) - 1.0
+    X <- as.matrix(iris[, names(iris) != "Species"])
+    row.names(X) <- paste("rname", seq(1L, nrow(X)), sep = "")
+    dtrain <- lgb.Dataset(X, label = y, params = list(max_bins = 5L))
+    bst <- lgb.train(
+        data = dtrain
+        , obj = "multiclass"
+        , params = list(num_class = 3L)
+        , nrounds = 5L
+        , verbose = -1L
```
jameslamb, line 185:
```suggestion
        , verbose = VERBOSITY
```

R-package/tests/testthat/test_Predictor.R
```diff
@@ -111,3 +113,117 @@ test_that("start_iteration works correctly", {
...
+    data(agaricus.train, package = "lightgbm")
+    X <- agaricus.train$data
+    y <- agaricus.train$label
+    row.names(X) <- paste("rname", seq(1L, nrow(X)), sep = "")
+    dtrain <- lgb.Dataset(X, label = y, params = list(max_bins = 5L))
+    bst <- lgb.train(
+        data = dtrain
+        , obj = "binary"
+        , nrounds = 5L
+        , verbose = -1L
```
jameslamb, line 212:
```suggestion
        , verbose = VERBOSITY
```

jameslamb (changes requested):
Thanks very much for adding tests and documenting the CSV issue in #5093.

The tests made me realize that for input data with row names, the value of `reshape` isn't sufficient to determine whether or not the output should have row names.

`reshape`'s value only matters for multi-class classification...for regression and binary classification tasks, the number of rows in the output from `predict()` should be the same as the number of rows in `newdata`, so for those objectives row names should be preserved _regardless of the value of `reshape`_.

Could you please do the following:

* move the code you've added into `predict.lgb.Booster()` into `Predictor$predict()`
* ensure that for binary classification and regression, row names are always preserved

For reference, here's the relevant code in `Predictor$predict()` that is figuring out whether or not the predictions have the same number of elements as the input data has rows.

https://github.com/microsoft/LightGBM/blob/17d4e007852b4f0b34b211a674cb26e6751c98e2/R-package/R/lgb.Predictor.R#L215-L231

david-cortes:
Updated.

R-package/tests/testthat/test_Predictor.R
```diff
@@ -111,3 +113,132 @@ test_that("start_iteration works correctly", {
...
+    pred <- predict(bst, X, reshape = TRUE, predleaf = TRUE)
+    .expect_has_row_names(pred, X)
+    pred <- predict(bst, X, reshape = TRUE, predcontrib = TRUE)
+    .expect_has_row_names(pred, X)
+    pred <- predict(bst, X, reshape = FALSE)
+    if (!multiclass || NROW(X) == NROW(pred)) {
+        .expect_has_row_names(pred, X)
+    } else {
+        .expect_doesnt_have_row_names(pred)
+    }
```
jameslamb, line 150: I'm really struggling to understand these test cases, since it requires reasoning about what was passed in to `.expect_row_names_kept()` from within each of the `test_that(...)` calls, and since these `if` statements rely on the fact that `pred` here happens to reference the predictions produced with `reshape = FALSE` and none of `rawscore` / `predleaf` / `predcontrib`.

To make this easier to follow (both for reviewing and for future contributors), can you please remove the use of this `.expect_row_names_kept()`  function and instead explicitly include checks in each of the `test_that(...)` calls? Those checks should not need to use `NROW(pred)` at all. For test code, I'd prefer some code duplication in exchange for it being easier to understand what is being tested.

jameslamb (changes requested):
Thanks very much for the changes! But please see my suggestion about simplifying the test code. In its current state, the indirection with `.expect_row_names_kept()` makes it difficult to understand what is being tested.

R-package/tests/testthat/test_Predictor.R
david-cortes: Not sure what you want me to do there. That `pred` *does* reference predictions with different combinations of `reshape` / `rawscore` / `predcontrib` / `predleaf` (all predictions generated there are named `pred`).

You asked me to add tests for different values of `reshape`, but `reshape` might get ignored depending on the rest of the parameters, and thus whether the output is meant to have row names or not, depends on the shape of what's actually returned in the end by the prediction function. If I were to adjust the tests by hard-coding the reshape logic in them, they then would test both the row names and the reshape logic together, which I don't think is ideal. The most logical way of testing the row names in a way that's independent of `reshape` would be by checking the number of rows in the data and the predictions.

jameslamb: > they then would test both the row names and the reshape logic together, which I don't think is ideal

"hard-coding the reshape logic", i.e. not changing the test expectation based on a condition like ` NROW(X) == NROW(pred)`, is exactly what I'm asking for.

We know, for example, that for `reshape = FALSE` and a multi-class objective, the output of `predict()` is expected to have `NROW(X) * number_of_classes` elements, and therefore shouldn't have row names. It's desirable for the tests to break if that changes.

As a maintainer on this project, I believe I'm able to push commits onto branches from forks associated with open PRs. Would you like me to just directly push the testing changes I'm proposing to this branch?

david-cortes: Yes, would be better if you modify the tests yourself.

jameslamb: Ok no prob! I'll push a commit in the next day or two.

jameslamb: Updated in https://github.com/microsoft/LightGBM/pull/4977/commits/04b743fd94e6d398909209adf306275ab67816fd.

Thankfully, #4971  helped simplify these tests (and therefore users' code). Thanks for that!

jameslamb (approved):
Very nice change, thanks for proposing it and for helping by adding such comprehensive tests!

jmoralez (commented):
Could the testing logic
```R
    # dense matrix with row names
    pred <- predict(bst, X)
    .expect_has_row_names(pred, X)
    pred <- predict(bst, X, rawscore = TRUE)
    .expect_has_row_names(pred, X)
    pred <- predict(bst, X, predleaf = TRUE)
    .expect_has_row_names(pred, X)
    pred <- predict(bst, X, predcontrib = TRUE)
    .expect_has_row_names(pred, X)

    # dense matrix without row names
    Xcopy <- X
    row.names(Xcopy) <- NULL
    pred <- predict(bst, Xcopy)
    .expect_doesnt_have_row_names(pred)

    # sparse matrix with row names
    Xcsc <- as(X, "CsparseMatrix")
    pred <- predict(bst, Xcsc)
    .expect_has_row_names(pred, Xcsc)
    pred <- predict(bst, Xcsc, rawscore = TRUE)
    .expect_has_row_names(pred, Xcsc)
    pred <- predict(bst, Xcsc, predleaf = TRUE)
    .expect_has_row_names(pred, Xcsc)
    pred <- predict(bst, Xcsc, predcontrib = TRUE)
    .expect_has_row_names(pred, Xcsc)

    # sparse matrix without row names
    Xcopy <- Xcsc
    row.names(Xcopy) <- NULL
    pred <- predict(bst, Xcopy)
    .expect_doesnt_have_row_names(pred)
```
 be extracted into a function?

david-cortes:
> Could the testing logic
> 
> ```r
>     # dense matrix with row names
>     pred <- predict(bst, X)
>     .expect_has_row_names(pred, X)
>     pred <- predict(bst, X, rawscore = TRUE)
>     .expect_has_row_names(pred, X)
>     pred <- predict(bst, X, predleaf = TRUE)
>     .expect_has_row_names(pred, X)
>     pred <- predict(bst, X, predcontrib = TRUE)
>     .expect_has_row_names(pred, X)
> 
>     # dense matrix without row names
>     Xcopy <- X
>     row.names(Xcopy) <- NULL
>     pred <- predict(bst, Xcopy)
>     .expect_doesnt_have_row_names(pred)
> 
>     # sparse matrix with row names
>     Xcsc <- as(X, "CsparseMatrix")
>     pred <- predict(bst, Xcsc)
>     .expect_has_row_names(pred, Xcsc)
>     pred <- predict(bst, Xcsc, rawscore = TRUE)
>     .expect_has_row_names(pred, Xcsc)
>     pred <- predict(bst, Xcsc, predleaf = TRUE)
>     .expect_has_row_names(pred, Xcsc)
>     pred <- predict(bst, Xcsc, predcontrib = TRUE)
>     .expect_has_row_names(pred, Xcsc)
> 
>     # sparse matrix without row names
>     Xcopy <- Xcsc
>     row.names(Xcopy) <- NULL
>     pred <- predict(bst, Xcopy)
>     .expect_doesnt_have_row_names(pred)
> ```
> 
> be extracted into a function?

Well, it was in a function before...

jameslamb:
> could the testing logic be extracted into a function

@jmoralez I specifically wanted to NOT do that (https://github.com/microsoft/LightGBM/pull/4977#discussion_r838005773) to make it more explicit exactly what the expectation is for each of these settings.

For this many combinations of possibilities, I think having each test be self-contained is preferable to the indirection introduced by putting that logic into a function.

jmoralez:
I think with the `reshape` argument it was a bit harder but now it's identical, isn't it? I actually spent some time trying to figure out whether or not something was different in one of them.

jameslamb:
Ok @jmoralez , updated in https://github.com/microsoft/LightGBM/pull/4977/commits/ec1c802d3ea85e941ad7d22af33e7e309e550bd8

jmoralez (approved):

PR #4978, opened 2022-01-23
[tests][python] remove compatibility code for old versions in tests

StrikerRUS:
Compatibility code was needed for versions of dependencies released more than year(s) ago. I think we can decrease maintenance burden by dropping support for quite old versions of dependencies. BTW, new PRs often don't consider support of old versions at all, e.g. #4927.


jameslamb (approved):
I'm ok with it.

`scikit-learn` 0.23.2 (the last version before 0.24) was realased in August 2020. ([release history](https://pypi.org/project/scikit-learn/#history))

`pandas` 0.23.4 (the last version before 0.24) was released in August 2018 ([release history](https://pypi.org/project/pandas/#history))

PR #4979, opened 2022-01-25
Correct documentation for sparse predictions

david-cortes:
The documentation says that the parameter `nindptr` for `LGBM_BoosterPredictSparseOutput` should be the number of rows plus one, but this is incorrect when the input type is CSC.

It also says that parameter `out_len` will at exit indicate the length of the indices and data, which is also incorrect since the first element it writes is the length of the index pointer.

This PR fixes both.

### DIFF
--- a/include/LightGBM/c_api.h
+++ b/include/LightGBM/c_api.h
@@ -839,7 +839,7 @@ LIGHTGBM_C_EXPORT int LGBM_BoosterPredictForCSR(BoosterHandle handle,
  * \param num_iteration Number of iterations for prediction, <= 0 means no limit
  * \param parameter Other parameters for prediction, e.g. early stopping for prediction
  * \param matrix_type Type of matrix input and output, can be ``C_API_MATRIX_TYPE_CSR`` or ``C_API_MATRIX_TYPE_CSC``
- * \param[out] out_len Length of output indptr and data (pointer to an array of two elements)
+ * \param[out] out_len Length of output data and indptr (pointer to an array of two elements)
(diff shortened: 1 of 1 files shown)


jameslamb:
@shiyu1994 or @tongwu-msft could you help review this one?

shiyu1994:
Sure.

include/LightGBM/c_api.h
```diff
@@ -839,7 +839,7 @@ LIGHTGBM_C_EXPORT int LGBM_BoosterPredictForCSR(BoosterHandle handle,
  * \param num_iteration Number of iterations for prediction, <= 0 means no limit
  * \param parameter Other parameters for prediction, e.g. early stopping for prediction
  * \param matrix_type Type of matrix input and output, can be ``C_API_MATRIX_TYPE_CSR`` or ``C_API_MATRIX_TYPE_CSC``
- * \param[out] out_len Length of output indices and data
+ * \param[out] out_len Length of output data and indptr (pointer to an array of two elements)
```
shiyu1994, line 842: Thanks for doing this! Maybe replacing `two elements` with `two lengths` will make it more clear, given that `Length` was previously mentioned in this line.

david-cortes: I think that'd make it more confusing but I'll try to reword it.

david-cortes:
Updated.

shiyu1994:
@david-cortes Thanks for the rewording!

shiyu1994 (approved):

Issue #4980, opened 2022-01-26
`LGBM_BoosterPredictForCSRSingleRowFastInit` behavior differs from documentation


david-cortes:
Calling C function `LGBM_BoosterPredictForCSRSingleRowFastInit` has undocumented side effects in the booster object that it takes as input. According to the documentation, this function should produce a `FastConfigHandle` object which is to then be passed to the prediction function `LGBM_BoosterPredictForCSRSingleRowFast`, but additionally:

* Calling this function will modify the `BoosterHandle` input that it is passed, changing its behavior also in calls to `LGBM_BoosterPredictForCSRSingleRow` (which do not used the produced `FastConfigHandle`).
* Calling this function twice reusing the same `BoosterHandle` but with different parameters (such as `start_iteration`) will in the second call produce an object which still has the configuration from the earlier object.
* Calling `LGBM_BoosterPredictForCSRSingleRow` with a different `predict_type` will affect the configuration from the `FastConfigHandle` object in further calls to `LGBM_BoosterPredictForCSRSingleRowFast`, even though `FastConfigHandle` is not passed to `LGBM_BoosterPredictForCSRSingleRow`.

StrikerRUS:
@AlbertoEAF Could you please take a look?

StrikerRUS:
@AlbertoEAF 
> I'm not available this month and probably not the next either.

No problem! Thank you very much for your detailed input!

david-cortes:
Once this gets sorted out, the R docs need to be updated to reflect the correct behavior: https://github.com/microsoft/LightGBM/pull/5312#event-7244449800

PR #4982, opened 2022-01-26
[R-package] Add missing prediction functions to R interface

### LINKED ISSUE #4966: [R-package] Cannot pass CSR single-row for prediction
When making predictions on a single row with sparse data, the Python interface for lightgbm allows passing the input as a sparse CSR matrix, and making predictions this way is faster/more efficient than with CSC format when the input is large. The R interface does not allow passing prediction-time inputs in CSR formats (package `Matrix` has classes `dsparseVector` and `dgRMatrix` for that).

david-cortes:
Fixes #4966

This PR adds R wrappers over important C-level functions for predictions which are currently missing from the R interface - particularly, functions dealing with single-row predictions and CSR data.

After issue https://github.com/microsoft/LightGBM/issues/4980 is sorted out, the documentation for `lgb.configure_fast_predict` needs to be updated to reflect the details about it.


david-cortes:
The sanitizer issues are not from this PR and should be fixed in this other PR: https://github.com/microsoft/LightGBM/pull/4984

jameslamb (changes requested):
Thanks for taking this on!

I believe this PR has two large and unrelated sets of changes (supporting single-row predictions and supporting predictions on more sparse array types).

Please break off the "support predicting on more sparse array types" changes into a separate PR to reduce the effort required to review these changes, or explain to us why you feel these two sets of changes need to be made together.

jameslamb:
> if it were to be done in sequential PRs, the first one would need slightly different code to leave something usable, which the second would then have to undo to arrive at the current code.

> it's more work

Reviewing a PR this large is, in my experience, more work for reviewers than the sum of reviewing smaller changesets that add up to this PR's changes. Maintainers in this project regularly context-switch between many different issues and PRs (here and in other open source projects), and PRs that are this large require a substantial amount of dedicated time to re-gather the right context to be able to thoroughly review changes.

Looking at the diff tonight, it seems to me that it should be possible to make these changes in the following sequence without any merge conflicts or half-solutions:

1. support `predcontrib` predictions for all sparse types
2. support non-`predcontrib` predictions for all sparse types (using only the multi-row prediction interfaces, e.g. `LGBM_BoosterPredictSparseOutput_R`)
3. support fast single-row predictions (e.g. `LGBM_BoosterPredictForMatSingleRow`) predictions for all types supported by `predict()`

If you're willing to do that, we'd appreciate it. If you're still not open to that after reading the suggestion above, then we can leave this as one PR but I expect there will be a much longer review cycle.

david-cortes:
Let's break it into 4-5 PRs then.

jameslamb:
Great, thank you very much!

jameslamb:
I believe this can be closed, based on the description of #5312.

> This PR adds the remainder of prediction functions from PR #4982... The idea in #4982 was to split the additions into multiple PRs, but ... this PR just adds the rest of C prediction functions that are not in the R interface together.

@david-cortes correct me if I'm wrong about that and this can be re-opened.

PR #4985, opened 2022-01-26
Pass train dataset parser config to valid dataset loading parser

chjinche:
Issue: We received [an issue](https://github.com/microsoft/lightgbm-transform/issues/16) in [lightgbm-transform](https://github.com/microsoft/lightgbm-transform), in which user complained that custom parser can not take effect on valid dataset.

Solution: With some investigation, we find it is a bug and could be fixed by passing train dataset parser config to valid dataset loading parser.


chjinche:
Kindly ping, much appreciated if any experts could look into the timeout ci failure? Looks unrelated to my change.

jameslamb:
> timeout ci failure

@chjinche please update this branch to the latest changes from `master`. We recently merged a change, #4953, that should fix the CI timeouts.

shiyu1994 (approved):
Thank you for the fix. LGTM!

PR #4989, opened 2022-01-30
[ci] upgrade to R 4.1.2 in CI, change approach to macOS-latest R CI builds (fixes #4988)

### LINKED ISSUE #4988: [ci] [R-package] macOS builds failing to install R
## Description

Starting some time in the last few days, R-package CI jobs on macOS have been failing with errors like the following.

Error seen on R 3.6.x builds

> installer: Package name is R 3.6.3 for Mac OS X 10.11 or higher (El Capitan build)
> /Users/runner/work/LightGBM/LightGBM/.ci/test_r_package.sh: line 108: Rscript: command not found

Error seen on R 4.1.x builds

> installer: Error - the package path specified was invalid: '/Users/runner/work/LightGBM/LightGBM/R.pkg'.

All R jobs on macOS are failing with this error, so it doesn't seem limited to specific R versions or compiler.

## Reproducible example

See, for example, https://github.com/microsoft/LightGBM/runs/4994115921?check_suite_focus=true.

This doesn't seem related to the changes on any specific PR, as these jobs are failing this way across multiple PRs which are not changing anything about how R is set up. For example, #4953, #4970, #4977.

See https://github.com/microsoft/LightGBM/actions/workflows/r_package.yml.

## Notes

See https://cran.r-project.org/doc/manuals/r-release/R-admin.html#Installing-R-under-macOS for guidance from CRAN on how to install R on Mac. Maybe the tips there could help in identifying the issue here.

jameslamb:
Fixes #4988.

See #4988 for a description of recent CI issues with R macOS builds.

I believe it could be related to GitHub Actions upgrading its `macOS-latest` environment to macOS 11.x 2 days ago. See https://github.com/actions/virtual-environments/pull/4048.

Searched around GitHub a bit and found a discussion in `r-lib/actions` that seems relevant. https://github.com/r-lib/actions/issues/412 describes an issue where older R installers can't create the necessary symlinks to `R` and `Rscript` on macOS 11.x.

The changes in this PR are based on the changes that that project used to fix that issue: https://github.com/r-lib/actions/commit/a1c2c67891672688569dddc7c8fc31c8230fe1a8


jameslamb:
It looks like these symlink changes fixed the issue for R 3.6.x, but not R 4.1.x: https://github.com/microsoft/LightGBM/runs/4994766651?check_suite_focus=true

I noticed at https://cran.r-project.org/bin/macosx/base/ that R 4.1.2 was published in November 2021. This project's CI currently uses R 4.1.1, published in August 2021.

I pushed https://github.com/microsoft/LightGBM/pull/4989/commits/8117a0594f81b0e0a13e3fca6a6581afc7465259 updated the R version to 4.1.2, hoping that that version's installer would work with macOS 11.x. Probably a good idea to do this anyway, per #3763.

jameslamb:
Ok I think this is ready for review! Sorry that the research is split between comments here and on #4988, I should have tried to keep them all in one place.

Summary of changes:

* pin R 4.x builds to `macOS-10.15` environment (the environment they were running on until 3 days ago)
    - reference: https://github.com/microsoft/LightGBM/issues/4988#issuecomment-1025069374
* for R 3.x builds, continue using `macOS-latest` environment but manually set symlinks to `R` and `Rscript` necessary to make that work
     - references: https://github.com/r-lib/actions/issues/412 + https://github.com/r-lib/actions/commit/a1c2c67891672688569dddc7c8fc31c8230fe1a8

StrikerRUS:
Thanks for looking into this!

Why can't we use the latest versions of macOS and R together? I believe this is the most popular scenario among users.

Maybe we can use Homebrew installation for R?

jameslamb:
> Why can't we use the latest versions of macOS and R together?

I don't know, but I can see in https://github.com/microsoft/LightGBM/issues/4988#issuecomment-1025069374 that GitHub itself was also not able to build the latest version of R on the `macOS-latest` environment.

When I tried that with R 4.1.2 from CRAN, I saw the following error ([build link](https://github.com/microsoft/LightGBM/runs/4994838445?check_suite_focus=true))

```text
installer: Error - the package path specified was invalid: '/Users/runner/work/LightGBM/LightGBM/R.pkg'.
```

> Maybe we can use Homebrew installation for R?

Maybe. I've never used R from homebrew before (in this project, other projects, or personally) so I can't say how that differs from the `.pkg` packages provided by CRAN or what other issues we might face with it.

If you want to try that you're welcome to push commits to this branch. But I don't have much time to devote to LightGBM this week so I can't commit to trying that. I think this PR in its current state is sufficient to unblock CI in this project without losing any coverage of OS-compiler-R_version combinations, as until a few days ago we were also not testing the R package with R 4.1.2 + macOS 11.x.

StrikerRUS:
Also linking related feature request: https://github.com/actions/virtual-environments/issues/4960.

StrikerRUS:
> When I tried that with R 4.1.2 from CRAN, I saw the following error (build link)

Could you please document this in a new our internal feature request?

.ci/test_r_package.sh
```diff
@@ -20,8 +20,8 @@ if [[ "${R_MAJOR_VERSION}" == "3" ]]; then
     export R_LINUX_VERSION="3.6.3-1bionic"
     export R_APT_REPO="bionic-cran35/"
 elif [[ "${R_MAJOR_VERSION}" == "4" ]]; then
-    export R_MAC_VERSION=4.1.1
-    export R_LINUX_VERSION="4.1.1-1.2004.0"
+    export R_MAC_VERSION=4.1.2
+    export R_LINUX_VERSION="4.1.2-1.2004.0"
```
StrikerRUS, line 24: I guess we need update all R versions in this list (`4.0` -> `4.1`), not only for macOS
https://github.com/microsoft/LightGBM/blob/0075814f02f898ae4070d848078b9d0d4b510ceb/.github/workflows/r_package.yml#L30-L140

StrikerRUS (approved):
LGTM except one suggestion below.
Thank you for unblocking CI!

.ci/test_r_package.sh
jameslamb: ah yep you're right! Updated in https://github.com/microsoft/LightGBM/pull/4989/commits/7cb0fd0e4428c57c8fef7a31a263e0ff6ee921b0

jameslamb:
> could you please document this in a new feature request?

Added #4990 and added it to #2302

PR #5001, opened 2022-02-11
[ci] use `latest` tag again for Ubuntu 14 CI Docker

StrikerRUS:
According to https://github.com/guolinke/lightgbm-ci-docker/pull/22#issuecomment-1014778489.

image


jameslamb (approved):
excellent! Thanks for your help with that.

Issue #5004, opened 2022-02-14
[python-package] “python_requires” should be set with “>=3.5”, as lightgbm 3.3.2 is not compatible with all Python versions.


PyVCEchecker:
## Description
Currently, the keyword argument **python_requires** of **setup()** is not set, and thus it is assumed that this distribution is compatible with all Python versions. 
However, I found it is not compatible with **Python 2**. My local Python version is 2.7, and I encounter the following error when executing _“pip install lightgbm”_ 

```
Collecting lightgbm
  Downloading lightgbm-3.3.2.tar.gz (1.5 MB)
     |████████████████████████████████| 1.5 MB 233 kB/s 
    ERROR: Command errored out with exit status 1:
     command: /usr/local/bin/python -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-O0hSCR/lightgbm/setup.py'"'"'; __file__='"'"'/tmp/pip-install-O0hSCR/lightgbm/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-pip-egg-info-TZ3SVN
         cwd: /tmp/pip-install-O0hSCR/lightgbm/
    Complete output (6 lines):
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-O0hSCR/lightgbm/setup.py", line 38
        def find_lib() -> List[str]:
                       ^
    SyntaxError: invalid syntax
    ----------------------------------------
ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
```
I noticed that **setup.py** used the syntax feature **Type Hints**. The feature is only compatible with Python 3, resulting in installation failure of lightgbm in Python 2.7.

Way to fix:
modify **setup()** in **setup.py,** add **python_requires** keyword argument:
```python
setup(…
     python_requires=">=3.5"
     …)
```
Thanks for your attention.
Best regrads,
PyVCEchecker

jameslamb:
Thanks very much for using LightGBM and for the report, @PyVCEchecker !

[...]

PR #5005, opened 2022-02-14
[python] add "python_requires" keyword argument for setup()

PyVCEchecker:
Add `python_requires='>=3.5'` for **setup()** #5004


python-package/setup.py
```diff
@@ -340,6 +340,7 @@ def run(self) -> None:
           version=version,
           description='LightGBM Python Package',
           long_description=readme,
+          python_requires='>=3.5',
```
StrikerRUS, line 343: Thanks for this contribution!
But I don't think that LightGBM is compatible with Python 3.5 due to using f-strings e.g. here
https://github.com/microsoft/LightGBM/blob/15266f1d613a37fc6d59bc0462854a99eeee3b0b/python-package/setup.py#L134
that were introduced in Python 3.6, 
https://www.python.org/dev/peps/pep-0498/
```suggestion
          python_requires='>=3.6',
```

StrikerRUS (changes requested):

StrikerRUS:
Also, I'm re-targeting this PR to the `master` branch from `py310` one.

StrikerRUS:
> Also, I'm re-targeting this PR to the `master` branch from `py310` one.

Ah, commits from `py310` were included into this PR by the re-targeting.
Also, CLA bot wasn't triggered for some reason.

@PyVCEchecker Could you please re-create a new PR targeting LightGBM's `master` branch and including all recent commits from it?

PyVCEchecker:
You are right. LightGBM is not compatible with Python 3.5. I have re-create a new PR here #5006

PR #5006, opened 2022-02-14
[python] enforce a floor of Python 3.6 (fixes #5004)

### LINKED ISSUE #5004: [python-package] “python_requires” should be set with “>=3.5”, as lightgbm 3.3.2 is not compatible with all Python versions.

PyVCEchecker:
Fixes #5004.
Replaces #5005.

Add python_requires='>=3.6' for setup().

### DIFF
--- a/python-package/setup.py
+++ b/python-package/setup.py
@@ -340,6 +340,7 @@ if __name__ == "__main__":
           version=version,
           description='LightGBM Python Package',
           long_description=readme,
+          python_requires='>=3.6',
           install_requires=[
               'wheel',
               'numpy',


jameslamb (approved):
Makes sense to me! Thanks for doing this!

I have two small notes for when you make future open source contributions.

1. I recommend creating a new branch from your fork, instead of directly committing to `master` there. That makes it possible to work on multiple pull requests at once, and ensures that there will never be commits on `master` of your fork that don't exist on the upstream project.
2. I've added `(fixes #5004)` to the title of this PR, so that the corresponding issue will be automatically closed when this is merged.

StrikerRUS (approved):
@PyVCEchecker Thanks a lot for this contribution, LGTM!

StrikerRUS:
@jmoralez My position is that LightGBM should be as much permissive as possible. This is by the way not only my personal position, but historically supported general policy in this repo. Absence of artificial restrictions allows to run LightGBM in a wide range of different environments and hence increase its' popularity among users. As an example, I can name Dockerized Linux artifacts building in Ubuntu 14, CI jobs with Visual Studio 2015a and CUDA 9.0. There are a lot of cases of restricted environments which users cannot update for different reasons. I believe it's important to let users at least try to run LightGBM in such environments.

jameslamb:
> My position is that LightGBM should be as much permissive as possible. 

I 100% agree with @StrikerRUS on this. Libraries like LightGBM should use as few ceilings / floors on language version and dependencies as possible.

In this case, I supported a floor of `python>=3.6` because with the use of type hints and f-strings throughout the package, older versions of Python don't even recognize the package as syntactically-valid Python code.

In case it helps describe my personal opinion on this better, I summarized some thoughts on this for library dependencies (not language version) over here: https://github.com/snowflakedb/snowflake-connector-python/pull/604#discussion_r562821349

PR #5007, opened 2022-02-15
[R-package] prefer params to keyword argument in lgb.train()

jameslamb:
From https://github.com/microsoft/LightGBM/pull/4976#discussion_r806384554.

Relates to #4913.

Today, when using `lightgbm::lgb.train()`, if you you pass different values to keyword argument `obj` and parameter `"objective"` in the `params` list, the value passed to the keyword argument will be preferred.

That is inconsistent with the way that the rest of the R package and the Python package works. Across LightGBM language-specific wrapper packages, it's expected that the values from `params` will take precedence over keyword arguments.

This PR proposes changing the R package to ensure that if `objective` or its aliases are passed through `params`, that value takes precedence over the keyword argument.

### Why make this breaking change now?

Normally I'd propose that a change like this go through a deprecation cycle. However, it's been almost 4 months since the previous release (v3.3.1, not counting v3.3.2 since that was just a small patch to satisfy CRAN), and given the current state of open large PRs like #3234, #4630, and #4827 I expect it will be at least 2 months until the next, 4.0.0 release.

### DIFF
--- a/R-package/R/utils.R
+++ b/R-package/R/utils.R
@@ -117,7 +117,7 @@ lgb.check_interaction_constraints <- function(interaction_constraints, column_na
 
 }
 
-lgb.check.obj <- function(params, obj) {
+lgb.check.obj <- function(params) {
 
   # List known objectives in a vector
   OBJECTIVES <- c(
@@ -158,24 +158,17 @@ lgb.check.obj <- function(params, obj) {
     , "xendcg_mart"
   )
 
-  # Check whether the objective is empty or not, and take it from params if needed
-  if (!is.null(obj)) {
-    params$objective <- obj
+  if (is.null(params$objective)) {
+    stop("lgb.check.obj: objective should be a character or a function")
   }
 
-  # Check whether the objective is a character
   if (is.character(params$objective)) {
 
-    # If the objective is a character, check if it is a known objective
     if (!(params$objective %in% OBJECTIVES)) {
 
       stop("lgb.check.obj: objective name error should be one of (", paste0(OBJECTIVES, collapse = ", "), ")")
 
     }
-
-  } else if (!is.function(params$objective)) {
-
-    stop("lgb.check.obj: objective should be a character or a function")
 
   }
 
--- a/R-package/tests/testthat/test_basic.R
+++ b/R-package/tests/testthat/test_basic.R
@@ -578,7 +578,7 @@ test_that("lgb.cv() prefers objective in params to keyword argument", {
     )
     , params = list(
       application = "regression_l1"
-      , verbosity = -1L
+      , verbosity = VERBOSITY
     )
     , nrounds = 5L
     , obj = "regression_l2"
@@ -721,7 +721,7 @@ test_that("lgb.train() prefers objective in params to keyword argument", {
     )
     , params = list(
         loss = "regression_l1"
-        , verbosity = -1L
+        , verbosity = VERBOSITY
     )
     , nrounds = 5L
     , obj = "regression_l2"
--- a/R-package/tests/testthat/test_basic.R
+++ b/R-package/tests/testthat/test_basic.R
@@ -569,6 +569,23 @@ test_that("lgb.cv() respects parameter aliases for objective", {
   expect_length(cv_bst$boosters, nfold)
 })
 
+test_that("lgb.cv() prefers objective in params to keyword argument", {
+  data("EuStockMarkets")
+  bst <- lgb.train(
+    data = lgb.Dataset(
+      data = EuStockMarkets[, c("SMI", "CAC", "FTSE")]
+      , label = EuStockMarkets[, "DAX"]
+    )
+    , params = list(
+      objective = "regression_l2"
+      , verbosity = -1L
+    )
+    , nrounds = 5L
+    , obj = "regression_l1"
+  )
+  expect_equal(bst$params$objective, "regression_l2")
+})
+
 test_that("lgb.cv() respects parameter aliases for metric", {
   nrounds <- 3L
   nfold <- 4L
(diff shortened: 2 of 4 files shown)


R-package/tests/testthat/test_basic.R
```diff
@@ -569,6 +569,23 @@ test_that("lgb.cv() respects parameter aliases for objective", {
   expect_length(cv_bst$boosters, nfold)
 })
 
+test_that("lgb.cv() prefers objective in params to keyword argument", {
+  data("EuStockMarkets")
+  bst <- lgb.train(
```
StrikerRUS, line 574: Should be `cv()` here.

R-package/tests/testthat/test_basic.R
```diff
@@ -684,6 +701,23 @@ test_that("lgb.train() respects parameter aliases for objective", {
...
+      , label = EuStockMarkets[, "DAX"]
+    )
+    , params = list(
+        objective = "regression_l2"
+        , verbosity = -1L
+    )
+    , nrounds = 5L
+    , obj = "regression_l1"
+  )
+  expect_equal(bst$params$objective, "regression_l2")
```
StrikerRUS, line 718: TBH, it doesn't look like a reliable test. Simple preserving params object "as-is" as a field of a Booster will pass this check. Maybe we can at least use some non-trivial `objective` param alias in `params`?

jameslamb: I agree with you. I just pushed https://github.com/microsoft/LightGBM/pull/5007/commits/297a0a561c37f4facbcf7c77b0328c4fc096d40c to make this a lot stricter.

* test that `objective` in the model dumped by by `$save_model_to_string()` matches the one from `params`
    - a lot stricter, since it relies on a code path that will extract the objective from the `Booster` object on the C++ side, instead of just the `params` list in the R object
    - https://github.com/microsoft/LightGBM/blob/9fc348af6f5ca60b91c9447d813a401567a4738a/R-package/src/lightgbm_R.cpp#L867
* use aliases for `objective` in params

R-package/tests/testthat/test_basic.R
jameslamb: ah!! bad copy-pasting, thanks for catching that 😬 

updated in https://github.com/microsoft/LightGBM/pull/5007/commits/297a0a561c37f4facbcf7c77b0328c4fc096d40c

jameslamb:
@shiyu1994 can you please help us diagnose why this `license/cla` check is here and is stuck in pending?

Screen Shot 2022-02-16 at 8 33 55 AM

#4985 and #5009 are also blocked by it. I believe the automation that checks whether contributors have signed the Contributor License Agreement is a Microsoft-wide configuration, not something specific to this repo, and I suspect something about it has changed.

R-package/tests/testthat/test_basic.R
```diff
@@ -569,6 +569,34 @@ test_that("lgb.cv() respects parameter aliases for objective", {
...
+test_that("lgb.cv() prefers objective in params to keyword argument", {
+  data("EuStockMarkets")
+  cv_bst <- lgb.cv(
+    data = lgb.Dataset(
+      data = EuStockMarkets[, c("SMI", "CAC", "FTSE")]
+      , label = EuStockMarkets[, "DAX"]
+    )
+    , params = list(
+      application = "regression_l1"
+      , verbosity = -1L
```
StrikerRUS, line 581:
```suggestion
      , verbosity = VERBOSITY
```

R-package/tests/testthat/test_basic.R
```diff
@@ -684,6 +712,31 @@ test_that("lgb.train() respects parameter aliases for objective", {
...
+test_that("lgb.train() prefers objective in params to keyword argument", {
+  data("EuStockMarkets")
+  bst <- lgb.train(
+    data = lgb.Dataset(
+      data = EuStockMarkets[, c("SMI", "CAC", "FTSE")]
+      , label = EuStockMarkets[, "DAX"]
+    )
+    , params = list(
+        loss = "regression_l1"
+        , verbosity = -1L
```
StrikerRUS, line 724:
```suggestion
        , verbosity = VERBOSITY
```

StrikerRUS (changes requested):

R-package/R/utils.R
```diff
@@ -158,25 +158,18 @@ lgb.check.obj <- function(params, obj) {
...
+    stop("lgb.check.obj: objective should be a character or a function")
   }
 
-  # Check whether the objective is a character
   if (is.character(params$objective)) {
 
-    # If the objective is a character, check if it is a known objective
     if (!(params$objective %in% OBJECTIVES)) {
 
       stop("lgb.check.obj: objective name error should be one of (", paste0(OBJECTIVES, collapse = ", "), ")")
```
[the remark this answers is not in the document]
jameslamb: Looks to me like `lgb.check.obj()` has been a part  of the R package since the very first PR introducing the R package, back in 2017.

https://github.com/microsoft/LightGBM/blame/9259a5318cd03951f2eefe378d9fb8db74457ee3/R-package/R/utils.R#L120

https://github.com/microsoft/LightGBM/pull/168

I agree that in the current version of the R package, this isn't really necessary any more. And I'd support removing it, to remove the need to maintain a hard-coded list of objective names in the R package.

```r
library(lightgbm)

data("EuStockMarkets")
lightgbm:::Booster$new(
    train_set = lightgbm::lgb.Dataset(
        data = EuStockMarkets[, c("SMI", "CAC", "FTSE")]
        , label = EuStockMarkets[, "DAX"]
    )
    , params = list(
        objective = "no_such_objective"
    )
)
```

Raises the following exception:

> [LightGBM] [Fatal] Unknown objective type name: no_such_objective
Error in try({ : Unknown objective type name: no_such_objective
Error in initialize(...) : lgb.Booster: cannot create Booster handle

I'll open a separate PR tonight factoring it out.

jameslamb:
> Seems that has recovered now.

ah great, glad just closing and re-opening things fixed it. Thanks for checking @shiyu1994 and @StrikerRUS

R-package/R/utils.R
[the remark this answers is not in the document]
StrikerRUS: Great, thanks!

StrikerRUS (approved):
LGTM, thank you!

PR #4959, opened 2022-01-18
[docs] clarify that categorical features will be converted to integers internally

jmoralez:
Emphasizes that categorical features **will** be converted to `int` internally, to avoid the confusion that they should be provided as integers in the input data, as was asked in #4932. Also adds information on what will happen if negative values are found.

Here's how that section looks like:
image

Closes #4932.

PR #5009, opened 2022-02-16
[docs] document rounding behavior of floating point numbers in categorical features

### LINKED ISSUE #4898: [RFC] Round to the nearest integer when predicting with categorical value
## Summary
When predicting with a categorical feature, LightGBM truncates the floating-point value into an integer using `static_cast<int>`, which truncates the value towards 0:
https://github.com/microsoft/LightGBM/blob/8a34b1af2d5ae734a446f97ee984508a55ecb2d4/include/LightGBM/tree.h#L368-L377

So for example, 3.3 becomes 3, and 2.9 becomes 2.

I propose that the input should be rounded to the nearest integer instead (e.g. 2.9 gets rounded to 3).

## Motivation

Users are expected to perform pre-processing to map categorical features to integer values. Floating-point errors may cause some category values to be a little off: 2.99999 for example. Truncating 2.99999 to 2 will cause a surprise.

## Description

The line
```cpp
int_fval = static_cast<int>(fval); 
```
should be replaced with
```cpp
int_fval = static_cast<int>(std::round(fval)); 
```

## References

Related: #4468

StrikerRUS:
Continuation of 2d1caf140a3b0c68f908071277b67c2a842a297b.

Closes #4898.

Similarly to my comment in https://github.com/microsoft/LightGBM/pull/4959#issuecomment-1016996904, there are a lot of places with categorical features description which we should keep in a consistent state.

### DIFF
--- a/docs/Advanced-Topics.rst
+++ b/docs/Advanced-Topics.rst
@@ -25,6 +25,7 @@ Categorical Feature Support
 
 -  Categorical features must be encoded as non-negative integers (``int``) less than ``Int32.MaxValue`` (2147483647).
    It is best to use a contiguous range of integers started from zero.
+   Floating point numbers in categorical features will be rounded towards 0.
 
 -  Use ``min_data_per_group``, ``cat_smooth`` to deal with over-fitting (when ``#data`` is small or ``#category`` is large).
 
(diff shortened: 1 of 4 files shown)


jameslamb (approved):
Completely agree! Thanks very much for doing this.

Issue #4719, opened 2021-10-26
[python] Neptune LightGBM callback yields an error with the new logic introduced in v3.3.0


aptlin:
## Description

Neptune-LightGBM integration is broken due to changes in handling the callbacks in v3.3.0.

## Reproducible example

1. Install the dependencies 
```bash
pip install lightgbm==3.3.0 neptune-client==0.10.10 neptune-lightgbm==0.9.13
```

2. Run the example:
```python
import copy
from neptune.new.integrations.lightgbm import NeptuneCallback as LGBMNeptuneCallback
import neptune.new as npt
import numpy as np
import pandas as pd
from scipy.special import expit

import lightgbm as lgb

#################
# Taken from https://github.com/microsoft/LightGBM/blob/master/examples/python-guide/logistic_regression.py
# Simulate some binary data with a single categorical and
#   single continuous predictor
np.random.seed(0)
N = 1000
X = pd.DataFrame({
    'continuous': range(N),
    'categorical': np.repeat([0, 1, 2, 3, 4], N / 5)
})
CATEGORICAL_EFFECTS = [-1, -1, -2, -2, 2]
LINEAR_TERM = np.array([
    -0.5 + 0.01 * X['continuous'][k]
    + CATEGORICAL_EFFECTS[X['categorical'][k]] for k in range(X.shape[0])
]) + np.random.normal(0, 1, X.shape[0])
TRUE_PROB = expit(LINEAR_TERM)
Y = np.random.binomial(1, TRUE_PROB, size=N)
#################

run = npt.init(mode='debug')
callback = LGBMNeptuneCallback(run)
callbacks = [callback]
model = lgb.LGBMClassifier()
model.fit(X, Y, callbacks=callbacks)

```

This fails with

```
TypeError: cannot pickle '_io.TextIOWrapper' object
```

## Environment info

LightGBM version: 3.3.0

## Additional Comments

neptune-lightgbm version: 0.9.13 
neptune-client version: 0.10.10

The reason why this happens is because of using a deepcopy for callbacks:

https://github.com/microsoft/LightGBM/blob/fa4ecf4c4da57b1889e39c872eb5449080f1e02e/python-package/lightgbm/sklearn.py#L733

Indeed, the error is the same for the code below:

```
import copy
copy.deepcopy(callbacks)
```

PR #5012, opened 2022-02-16
[python] make `early_stopping` callback pickleable

Yard1:
Hey, I'm a maintainer of [LightGBM-Ray](https://github.com/ray-project/lightgbm_ray). I was working on early stopping support for that library, however the early stopping callback on LightGBM's master branch raises exceptions in [our tests (see `test_linux_cutting_edge (3.7)`)](https://github.com/ray-project/lightgbm_ray/runs/5222443116?check_suite_focus=true#logs).

I assumed this is due to the `nonlocal` variables present in the callback and them not interacting well with Ray serialization using cloudpickle. I would imagine there would be similar issues with other distributed computing or serialization libraries.

In order to solve that issue, I took the liberty of rewriting that callback into a callable class, with the `nonlocal` variables becoming attributes. This solved the issues with serialization. No functionality has been affected.

Please let me know if this approach is suitable, and if not, I would love to discuss other potential solutions to the problem I am facing.


StrikerRUS:
@Yard1 Hey, thank you very much for the interest in LightGBM and this contribution!

As a early opinion I would say I support conversion complex functions with closures into classes. I think this will make callbacks usage more user friendly and LightGBM codebase simpler.

Just a note: we have received a feedback about the need in callbacks serializability from Neptune maintainers recently #4719. Also, XGBoost migrated to callback classes: https://github.com/dmlc/xgboost/pull/6199. So I believe this is the right direction.

jameslamb:
I agree with @StrikerRUS , I definitely support the proposed direction. I'll mark this PR as "in-progress" for now. Please get the unit tests working again, and then we'll be able to provide a review. Let us know if you need help with the tests.

jameslamb:
@Yard1 since you're a new contributor, we'll have to manually approve CI runs on each commit

image

This is something GitHub added last year (I think?) to combat abuse of GitHub Actions.

Sorry for the inconvenience!

Yard1:
@jameslamb All's good, was on the accepting side myself many times 😂 The tests pass locally for me now, so it should be good!

python-package/lightgbm/callback.py
```diff
@@ -231,124 +231,135 @@ def early_stopping(stopping_rounds: int, first_metric_only: bool = False, verbos
...
+                    continue
+                elif env.iteration - self.best_iter[i] >= self.stopping_rounds:
+                    if self.verbose:
+                        eval_result_str = '\t'.join(
+                            [_format_eval_result(x) for x in self.best_score_list[i]])
+                        _log_info(
+                            f"Early stopping, best iteration is:\n[{self.best_iter[i] + 1}]\t{eval_result_str}")
+                        if self.first_metric_only:
+                            _log_info(
+                                f"Evaluated only: {eval_name_splitted[-1]}")
```
jameslamb, line 360: Could you please revert indentation changes like these? That would simplify the diff and make it easier for reviewers to understand what is being changed.

For example, on `master` this project has

```python
eval_result_str = '\t'.join([_format_eval_result(x) for x in best_score_list[i]])
```

and on this PR you have

```python
eval_result_str = '\t'.join(
                            [_format_eval_result(x) for x in self.best_score_list[i]])
```

We don't enforce a limit on line length in this project's Python code, and prefer to leave style-only changes to maintainers in most cases (to ensure that the git blame is easy to use as a source of information about when lines changed).

python-package/lightgbm/callback.py
```diff
@@ -231,124 +231,135 @@ def early_stopping(stopping_rounds: int, first_metric_only: bool = False, verbos
...
-        # split is needed for "<dataset type> <metric>" case (e.g. "train l1")
-        first_metric = env.evaluation_result_list[0][1].split(" ")[-1]
-        for eval_ret, delta in zip(env.evaluation_result_list, deltas):
-            best_iter.append(0)
-            best_score_list.append(None)
-            if eval_ret[3]:  # greater is better
-                best_score.append(float('-inf'))
-                cmp_op.append(partial(_gt_delta, delta=delta))
+    class _early_stopping_callback:
+        def __init__(self, stopping_rounds: int, first_metric_only: bool = False, verbose: bool = True, min_delta: Union[float, List[float]] = 0.0) -> None:
```
jameslamb, line 236:
```suggestion
        def __init__(
            self,
            stopping_rounds: int,
            first_metric_only: bool = False,
            verbose: bool = True,
            min_delta: Union[float, List[float]] = 0.0
        ) -> None:
```

Thanks for adding type hints here! Could you please use this suggested indentation? We've slowly been adopting type hints in this package (see the discussion and PRs linked to #3756), and have used the convention of this one-argument-per-line formatting when there is more than one argument.

For example:

https://github.com/microsoft/LightGBM/blob/717631af1836be9b03444ece485f3a3b5dbae5da/python-package/lightgbm/dask.py#L167-L177

jameslamb (changes requested):
Thanks for this! The fact that tests are passing without needing any modification is a very good sign 🙌 

Please see my first round of suggestions. I think it would be easier to review this PR if some of the style changes you've propose were reverted.

python-package/lightgbm/callback.py
Yard1: Hey @jameslamb, I couldn't find any information on what formatting style to use so I just ran autopep8. Let me try to make the formatting as close to original as possible.

Yard1:
@jameslamb I've applied your feedback, please take a look. Thanks!

jmoralez (approved):
Thank you for this! LGTM

jameslamb (changes requested):
Thanks very much for doing this!

Could you please add a test in https://github.com/Yard1/LightGBM/blob/patch-1/tests/python_package_test/test_engine.py that ensures that this callback is pickleable? 

The primary purpose of this PR, as I understand it, is to make the early stopping callback pickleable so that it can be used by frameworks like Ray that are serializing the arguments to `lgb.train()` to send them around to different process. If I'm right about that, I think it would be valuable to explicitly test that so that future changes to LightGBM don't accidentally break that behavior.

Ideally this would work with `pickle`, `joblib`, and `cloudpickle`...could you try that? See the Dask tests for inspiration.

https://github.com/microsoft/LightGBM/blob/f185695617d29e446e021fed1261ba84f09ff992/tests/python_package_test/test_dask.py#L237-L260

https://github.com/microsoft/LightGBM/blob/f185695617d29e446e021fed1261ba84f09ff992/tests/python_package_test/test_dask.py#L1296

---------------------

@jmoralez I suspect we'll need to make changes similar to this PR to for all other callbacks to support `callbacks` in `lightgbm.dask`. I can write up a feature request for that.

Yard1:
@jameslamb I have added a test, please take a look!

Yard1:
The tests are failing due to a scheduled Windows brownout in CI.

Yard1:
Hey @jameslamb the Windows CI env doesn't seem to have cloudpickle installed. How can we fix that?

python-package/lightgbm/callback.py
```diff
@@ -199,156 +199,160 @@ def _callback(env: CallbackEnv) -> None:
...
-        nonlocal best_score
-        nonlocal best_iter
-        nonlocal best_score_list
-        nonlocal cmp_op
-        nonlocal enabled
-        nonlocal first_metric
-        enabled = not any(env.params.get(boost_alias, "") == 'dart' for boost_alias
-                          in _ConfigAliases.get("boosting"))
-        if not enabled:
+class _EarlyStoppingCallback:
```
jameslamb, line 202: can you please explain why it was necessary to pull this class out of the function? Was that required to make the callback function pickleable?

jameslamb (changes requested):
Great test! Thanks very much for adding that, and for moving the pickling / unpickling helper functions into `utils.py`.

Please see one question I have about related changes in `callback.py`.

python-package/lightgbm/callback.py
Yard1: Yes - cloudpickle can pickle local functions and classes, but pickle and joblib require them to be accessible from module level.

jameslamb: Ah interesting, thanks for that. Glad we added tests!

Yard1:
@jameslamb Had to merge with master, please rerun tests

tests/python_package_test/test_engine.py
```diff
@@ -18,8 +18,9 @@
 
 import lightgbm as lgb
 
-from .utils import (load_boston, load_breast_cancer, load_digits, load_iris, make_synthetic_regression,
-                    sklearn_multiclass_custom_objective, softmax)
+from .utils import (load_boston, load_breast_cancer, load_digits, load_iris, make_synthetic_regression, pickle_obj,
+                    unpickle_obj, sklearn_multiclass_custom_objective, softmax)
+
```
Yard1, line 23:
```suggestion
from .utils import (load_boston, load_breast_cancer, load_digits, load_iris, make_synthetic_regression, pickle_obj,
                    unpickle_obj, sklearn_multiclass_custom_objective, softmax)
```

python-package/lightgbm/callback.py
```diff
@@ -199,156 +199,160 @@ def _callback(env: CallbackEnv) -> None:
...
+        By default, standard output resource is used.
+        Use ``register_logger()`` function to register a custom logger.
+    min_delta : float or list of float, optional (default=0.0)
+        Minimum improvement in score to keep training.
+        If float, this single value is used for all metrics.
+        If list, its length should match the total number of metrics.
+
+    Returns
+    -------
+    callback : callable
```
StrikerRUS, line 352:
```suggestion
    callback : _EarlyStoppingCallback
```

python-package/lightgbm/callback.py
```diff
@@ -199,156 +199,160 @@ def _callback(env: CallbackEnv) -> None:
...
+        The callback that activates early stopping.
+    """
+    return _EarlyStoppingCallback(stopping_rounds=stopping_rounds, first_metric_only=first_metric_only, verbose=verbose, min_delta=min_delta)
+
+
+_EarlyStoppingCallback.__doc__ = early_stopping.__doc__
```
StrikerRUS, line 358: I think we don't need this line for the internal class.

python-package/lightgbm/callback.py
```diff
@@ -199,156 +199,160 @@ def _callback(env: CallbackEnv) -> None:
...
         for eval_ret, delta in zip(env.evaluation_result_list, deltas):
-            best_iter.append(0)
-            best_score_list.append(None)
+            self.best_iter.append(0)
+            self.best_score_list.append(None)
             if eval_ret[3]:  # greater is better
-                best_score.append(float('-inf'))
-                cmp_op.append(partial(_gt_delta, delta=delta))
+                self.best_score.append(float('-inf'))
+                self.cmp_op.append(partial(_gt_delta, delta=delta))
```
StrikerRUS, line 279: I think we can encapsulate `_gt_delta()` and `_lt_delta()` functions in this class.

tests/python_package_test/test_engine.py
```diff
@@ -3424,3 +3424,19 @@ def test_pandas_nullable_dtypes():
 
     # test equal predictions
     np.testing.assert_allclose(preds, preds_nullable_dtypes)
+
+
+@pytest.mark.parametrize('serializer', ["pickle", "joblib", "cloudpickle"])
+def test_early_stopping_callback_is_picklable(serializer, tmp_path):
```
StrikerRUS, line 3430: Please create a new file named `test_callback.py` in the `python_package_test` folder and move this test there.

StrikerRUS (changes requested):
Thank you very much for your work! Please consider checking some very minor suggestions below from me.

StrikerRUS (approved):
Awesome work! Thank you very much for this PR!

Yard1:
@jameslamb Gentle ping :)

Yard1:
Let me merge master again, maybe that will fix the CI

Yard1:
@StrikerRUS @jameslamb CI is green now 🙏

StrikerRUS:
@jameslamb Could you please take another look before merging?

jameslamb (approved):
Looks great!

Thanks so much for your work on this, and for adding tests. 🚀

jameslamb:
Thanks again for this!

I changed the title to "make `early_stopping` callback pickleable" because PR titles in this project automatically become items in the release notes, and I think the fact that `lgb.early_stopping` is now pickleable will be more important to users than the particular mechanism used to achieve that.

I've also added #5080 documenting the work to ensure that all other callbacks are pickleable. Thanks for bringing the issue to our attention.

Good luck with `lightgbm-ray`, and we hope we'll see more contributions from you in the future 👋

Yard1:
Thanks!

PR #5013, opened 2022-02-17
[Docs] Weights non-negative for train data

### LINKED ISSUE #4921: [python-package] [docs]: weights params description should read weights should be non-negative
## Description

Following conversation on PR #4894, any method using `weight` parameter should describe the following 

> Weight should be non-negative values because the Hessian value multiplied by weight is supposed to be non-negative.

TremaMiguel:
## Context

This PR aims to close #4921 

## Changes

Update `weight` parameter documentation for training data to read: 

```markdown

Weight should be non-negative values because the Hessian
value multiplied by weight is supposed to be non-negative.
```


StrikerRUS (commented):
Please replace `Weight should be ...` with `Weights should be ...`.
And here too:
https://github.com/microsoft/LightGBM/blob/b99a8f01a0de3f4ddc3f254f568216b2a95abbeb/python-package/lightgbm/basic.py#L3545

Add this clarification for non-negative weights to the following line in params description as well:
https://github.com/microsoft/LightGBM/blob/b99a8f01a0de3f4ddc3f254f568216b2a95abbeb/include/LightGBM/config.h#L673

Also, WDYT about weights for validation data?
https://github.com/microsoft/LightGBM/blob/b99a8f01a0de3f4ddc3f254f568216b2a95abbeb/python-package/lightgbm/sklearn.py#L160
https://github.com/microsoft/LightGBM/blob/b99a8f01a0de3f4ddc3f254f568216b2a95abbeb/python-package/lightgbm/sklearn.py#L240
https://github.com/microsoft/LightGBM/blob/b99a8f01a0de3f4ddc3f254f568216b2a95abbeb/python-package/lightgbm/sklearn.py#L295
https://github.com/microsoft/LightGBM/blob/b99a8f01a0de3f4ddc3f254f568216b2a95abbeb/python-package/lightgbm/dask.py#L444
https://github.com/microsoft/LightGBM/blob/b99a8f01a0de3f4ddc3f254f568216b2a95abbeb/python-package/lightgbm/basic.py#L1821

We don't compute Hessians there, but I think it would be good to ask to set non-negative weights as a general rule there as well. For example, in metrics calculation we use sum of weights as a denominator. Negative values may cause division by zero.
https://github.com/microsoft/LightGBM/blob/b99a8f01a0de3f4ddc3f254f568216b2a95abbeb/src/metric/regression_metric.hpp#L97-L99

cc @shiyu1994 @jameslamb @jmoralez

docs/Parameters.rst
```diff
@@ -1275,7 +1275,9 @@ LightGBM supports weighted training. It uses an additional file to store weight
     0.8
     ...
 
-It means the weight of the first data row is ``1.0``, second is ``0.5``, and so on.
+It means the weight of the first data row is ``1.0``, second is ``0.5``, and so on. Weight should be non-negative
+values because the Hessian value multiplied by weight is supposed to be non-negative.
```
shiyu1994, line 1279: Thank you for dong this. I think it would be enough to say `weighs should be non-negative`. Exposing details about Hessian values may make the users confusing if they don't understand how Hessian values are used in GBDT algorithms.

shiyu1994:
> We don't compute Hessians there, but I think it would be good to ask to set non-negative weights as a general rule there as well. For example, in metrics calculation we use sum of weights as a denominator. Negative values may cause division by zero.

I think we only need to mention that sample weights should be non-negative, leaving out details about Hessian values.

TremaMiguel:
@StrikerRUS with validation data the justification can read

```python
Weights should be non-negative because it may cause division by zero
in metrics calculation (sum of weights as denominator).
```

Otherwise, it may not be clear to the user why weights should be non-negative.

docs/Parameters.rst
TremaMiguel: Ok, here the simple explanation works because you can already infer it from the example.

TremaMiguel:
@StrikerRUS @shiyu1994 we have the option whether or not to detail a brief explanation about non-negative weights.

1. For training data

```python
weights should be non-negative values because the Hessian value multiplied by weight is supposed to be non-negative
```

2. For validation data

```python
Weights should be non-negative because it may cause division by zero
in metrics calculation (sum of weights as denominator).
```

I'm more inclined to the brief explanation so that the user has a general idea about why weights should be non-negative, what are your thoughts?

StrikerRUS:
Hmmm, it has just come to my mind! Maybe instead of docs we can simply prohibit negative values in weights in cpp code just like we avoid inf here
https://github.com/microsoft/LightGBM/blob/d517ba12f2e7862ac533908304dddbd770655d2b/src/io/metadata.cpp#L337-L340
or like it's already done in xentropy objective?
https://github.com/microsoft/LightGBM/blob/820ae7e65e90860061ec89cc4a1ad0a39b96e051/src/objective/xentropy_objective.hpp#L64-L74

@shiyu1994

shiyu1994:
@StrikerRUS Sure. But in addition with prohibition in C++ code, documentation is also a good hint so that users won't feel surprised when their code with negative weights fails. Maybe we can have another PR for checking weights for non-negativity in C++ code.

@TremaMiguel Thanks for the proposal! After a more thorough think, I feel that the reason that weights should be non-negative is not only because Hessian values. It is unreasonable to set an instance in training dataset with negative weight by nature. Since the instance will be considered to harm the performance of the model. In that case, maybe the user should modify the label of that instance instead of setting a negative weight value.
So I think a description that the weights should be non-negative is good enough. The users can get the motivation behind the hint.

docs/Parameters.rst
```diff
@@ -1275,8 +1275,7 @@ LightGBM supports weighted training. It uses an additional file to store weight
     0.8
     ...
 
-It means the weight of the first data row is ``1.0``, second is ``0.5``, and so on. Weight should be non-negative
-values because the Hessian value multiplied by weight is supposed to be non-negative.
+It means the weight of the first data row is ``1.0``, second is ``0.5``, and so on. Weights should be non-negative.
```
shiyu1994, line 1278: Thank you for the contribution. You may use the script
https://github.com/microsoft/LightGBM/blob/master/helpers/parameter_generator.py
to generate Parameters.rst after modifying the description of parameters in config.h. In this way, the documentation will be consistent with the description in the code, and the static checks won't fail due to inconsistency.

python-package/lightgbm_test.ipynb
```diff
@@ -0,0 +1,165 @@
+{
```
shiyu1994, line 1: Is this file added by accident?

TremaMiguel: yes, sorry on that

docs/Parameters.rst
TremaMiguel: thanks, it would be difficult to figure that on my own

TremaMiguel:
@shiyu1994 ok let's go with a shorter description. Issue #5023 with c++ change is open.

StrikerRUS (approved):
Nicely done, thanks a lot for this PR and for opening new issue!

shiyu1994:
Close and re-open to include the latest fix for CI.

PR #5014, opened 2022-02-17
[R-package] raise informative errors directly when Booster creation fails

jameslamb:
Based on https://github.com/microsoft/LightGBM/pull/5007#discussion_r808601126, I started working on a PR to remove `lgb.check.obj()` from the R package.

While doing that, I found that the `try()` in `Booster$initialize()` made it difficult to catch specific error messages in tests. I think that that `try()` is left over from prior to the work for #3016. Now that errors from the C++ side are raised directly as R errors, I no longer think it's necessary.

This PR doesn't change anything about when the package raises an error, just the content of the error message.

### Note for Reviewers

I recommend using "Hide whitespace" when reviewing this diff.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -3,7 +3,7 @@ version: 3.3.2.99.{build}
 image: Visual Studio 2015
 platform: x64
 configuration:  # a trick to construct a build matrix with multiple Python versions
-  - 3.7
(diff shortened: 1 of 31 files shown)


StrikerRUS (approved):
Great, thank you!

PR #5015, opened 2022-02-18
[ci] Enclose Python versions in quotes in config yml files

StrikerRUS:
Use strings instead of floats for exact matching.
Refer to https://github.com/microsoft/LightGBM/pull/3880#discussion_r566841420.


jameslamb (approved):
thanks!

PR #5021, opened 2022-02-20
[R-package] remove internal function lgb.check.obj()

jameslamb:
Based on https://github.com/microsoft/LightGBM/pull/5007#discussion_r808401098. The R package has an internal function, `lgb.check.obj()`, that has been there since the very beginning of the R package (#168).

All of the things it does or once did are already handled by other parts of the code base.

1. resolve parameter aliases for `objective` and keyword argument
    - https://github.com/microsoft/LightGBM/blob/820ae7e65e90860061ec89cc4a1ad0a39b96e051/R-package/R/lgb.train.R#L95-L99
    - #4913, #5007
2. extract custom objective functions written in R
    - https://github.com/microsoft/LightGBM/blob/820ae7e65e90860061ec89cc4a1ad0a39b96e051/R-package/R/lgb.train.R#L109-L113
3. raise an informative error for unrecognized objectives
    - https://github.com/microsoft/LightGBM/blob/d517ba12f2e7862ac533908304dddbd770655d2b/src/objective/objective_function.cpp#L51

This PR proposes removing `lgb.check.obj()` from the R package, to reduce maintenance burden for the package and remove the risk of bugs caused by that code.

### Notes for Reviewers

Should be merged prior to #4976.

### DIFF
--- a/R-package/R/lgb.cv.R
+++ b/R-package/R/lgb.cv.R
@@ -127,7 +127,7 @@ lgb.cv <- function(params = list()
   params <- lgb.check.wrapper_param(
     main_param_name = "objective"
     , params = params
-    , alternative_kwarg_value = NULL
+    , alternative_kwarg_value = obj
   )
   params <- lgb.check.wrapper_param(
     main_param_name = "early_stopping_round"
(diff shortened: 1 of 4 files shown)


StrikerRUS (approved):
LGTM, thanks! Nice to see this simplification! 🙂

jameslamb:
Just updated this branch to get the CI fixes from #5022

PR #5029, opened 2022-02-23
[docs][R-package] Use direct link for the list of supported objectives

StrikerRUS:
Linking #4976.

We can provide a direct reference to any parameter name instead of making users to search for it through the huge "Parameters" web page.


jameslamb (approved):
love it, thanks!

PR #5034, opened 2022-02-27
[tests][python] move tests that use `train()` function defined in `engine.py` from `test_basic.py` to `test_engine.py`

StrikerRUS:
No changes applied to the code, just pure copy-paste.


jmoralez (approved):
Thank you!

jameslamb (approved):
I agree, thanks for this! Sorry for not catching it in earlier reviews.

Issue #4708, opened 2021-10-22
trees with one leaf can have strong bias in predictions, far outside observed range


arnocandel:
## Description
Regression model predicts outside y range, by a lot
```
3.3.0
ymin=996.5
ymax=1034.3

n_estimators=1
pred_min=1019.4970124657474
pred_max=1019.4970124657474

n_estimators=10
pred_min=437.80879810966866
pred_max=440.58117669151034

n_estimators=100
pred_min=-3.610830259193446
pred_max=4.928009330393595
```

## Reproducible example
```
import lightgbm as lgb
import numpy as np
from lightgbm.sklearn import LGBMRegressor

X = np.array(
    [
        [1021.0589, 1018.9578],
        [1023.85754, 1018.7854],
        [1024.5468, 1018.88513],
        [1019.02954, 1018.88513],
        [1016.79926, 1018.88513],
        [1007.6, 1018.88513],
        [1014.86957, 1018.88513],
        [1016.6986, 1018.9578],
        [1011.2, 1018.88513],
        [1007.6, 1018.9578],
        [1016.8388, 1018.9578],
        [1021.7486, 1018.9578],
        [1016.8388, 1018.9578],
        [1016.6986, 1018.9578],
        [1016.8388, 1018.9578],
        [1017.6715, 1018.7854],
        [1007.7, 1018.7854],
        [1021.17773, 1018.88513],
        [1016.8388, 1018.9578],
        [1017.6715, 1018.7854],
        [1016.8388, 1018.9578],
        [1016.8388, 1018.9578],
        [1017.3, 1018.7854],
        [1021.17773, 1018.88513],
        [1021.17773, 1018.88513],
        [996.5, 1018.88513],
        [1017.5635, 1018.7854],
        [1016.6986, 1018.9578],
        [1021.3, 1018.88513],
        [1024.6957, 1018.88513],
        [1021.3, 1018.88513],
        [1017.3, 1018.7854],
        [999.4, 1018.7854],
        [1017.5635, 1018.7854],
        [1016.8388, 1018.9578],
        [1021.17773, 1018.88513],
        [1007.9, 1018.88513],
        [1016.8, 1018.9578],
        [1010.2, 1018.7854],
        [1030.2, 1018.88513],
        [1021.7486, 1018.9578],
        [1016.6986, 1018.9578],
        [1013.09265, 1018.9578],
        [1016.8388, 1018.9578],
        [1026.4463, 1018.88513],
[...]

PR #5050, opened 2022-03-03
[fix] fix duplicate added initial scores for single-leaf trees

shiyu1994:
In example of #4708, the initial scores can be added twice when for single-leaf trees. This is to fix the duplicate addition.

### DIFF
--- a/src/boosting/gbdt.cpp
+++ b/src/boosting/gbdt.cpp
@@ -419,20 +419,15 @@ bool GBDT::TrainOneIter(const score_t* gradients, const score_t* hessians) {
     } else {
       // only add default score one-time
       if (models_.size() < static_cast<size_t>(num_tree_per_iteration_)) {
-        double output = 0.0;
-        if (!class_need_train_[cur_tree_id]) {
-          if (objective_function_ != nullptr) {
-            output = objective_function_->BoostFromScore(cur_tree_id);
+        if (objective_function_ != nullptr && !config_->boost_from_average && !train_score_updater_->has_init_score()) {
+          init_scores[cur_tree_id] = ObtainAutomaticInitialScore(objective_function_, cur_tree_id);
+          // updates scores
+          train_score_updater_->AddScore(init_scores[cur_tree_id], cur_tree_id);
+          for (auto& score_updater : valid_score_updater_) {
+            score_updater->AddScore(init_scores[cur_tree_id], cur_tree_id);
           }
-        } else {
-          output = init_scores[cur_tree_id];
         }
-        new_tree->AsConstantTree(output);
-        // updates scores
-        train_score_updater_->AddScore(output, cur_tree_id);
-        for (auto& score_updater : valid_score_updater_) {
-          score_updater->AddScore(output, cur_tree_id);
-        }
+        new_tree->AsConstantTree(init_scores[cur_tree_id]);
       }
     }
     // add model
--- a/tests/python_package_test/test_engine.py
+++ b/tests/python_package_test/test_engine.py
@@ -3427,8 +3427,10 @@ def test_pandas_nullable_dtypes():
 
 
 def test_boost_from_average_with_single_leaf_trees():
-    X = np.array(
-        [[1021.0589, 1018.9578],
+    # test data are taken from bug report
+    # https://github.com/microsoft/LightGBM/issues/4708
+    X = np.array([
+        [1021.0589, 1018.9578],
         [1023.85754, 1018.7854],
         [1024.5468, 1018.88513],
         [1019.02954, 1018.88513],
@@ -3440,7 +3442,7 @@ def test_boost_from_average_with_single_leaf_trees():
         "min_data_in_bin": 1,
         "extra_seed": 7,
         "objective": "regression",
-        "verbose": -2,
+        "verbose": -1,
         "boost_from_average": True,
         "min_data_in_leaf": 1,
     }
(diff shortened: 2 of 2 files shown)


src/boosting/gbdt.cpp
```diff
@@ -419,20 +419,15 @@ bool GBDT::TrainOneIter(const score_t* gradients, const score_t* hessians) {
     } else {
       // only add default score one-time
       if (models_.size() < static_cast<size_t>(num_tree_per_iteration_)) {
-        double output = 0.0;
-        if (!class_need_train_[cur_tree_id]) {
```
shiyu1994, line 423: Note that in `GBDT::BoostFromAverage`, the average score is calculated and added to the training/validation scorers regardless of `class_need_train_`, so this should be remove.

StrikerRUS:
I believe the example from #4708 can be added as a regression test.

guolinke (approved):

jameslamb:
> I believe the example from #4708 can be added as a regression test.

I agree with this. Every time we fix a bug, that PR should also include a test to prevent that bug from being re-introduced.

shiyu1994:
@StrikerRUS test case added. Could you please help to check the python part? Thank you!

shiyu1994:
The test case fails on master branch.

tests/python_package_test/test_engine.py
```diff
@@ -18,7 +18,7 @@
 
 import lightgbm as lgb
 
-from .utils import (load_boston, load_breast_cancer, load_digits, load_iris, make_synthetic_regression,
+from utils import (load_boston, load_breast_cancer, load_digits, load_iris, make_synthetic_regression,
```
StrikerRUS, line 21: @shiyu1994 Sure! Thanks a lot for adding the test!

Let's fix imports first.

```suggestion
from .utils import (load_boston, load_breast_cancer, load_digits, load_iris, make_synthetic_regression,
```

tests/python_package_test/test_engine.py
```diff
@@ -3424,3 +3424,29 @@ def test_pandas_nullable_dtypes():
 
     # test equal predictions
     np.testing.assert_allclose(preds, preds_nullable_dtypes)
+
+
+def test_boost_from_average_with_single_leaf_trees():
+    X = np.array(
```
StrikerRUS, line 3430:
```suggestion
    # test data are taken from bug report
    # https://github.com/microsoft/LightGBM/issues/4708
    X = np.array(
```

tests/python_package_test/test_engine.py
```diff
@@ -3424,3 +3424,29 @@ def test_pandas_nullable_dtypes():
...
+        [1019.02954, 1018.88513],
+        [1016.79926, 1018.88513],
+        [1007.6, 1018.88513]], dtype=np.float32)
+    y = np.array([1023.8, 1024.6, 1024.4, 1023.8, 1022.0, 1014.4], dtype=np.float32)
+    params = {
+        "extra_trees": True,
+        "min_data_in_bin": 1,
+        "extra_seed": 7,
+        "objective": "regression",
+        "verbose": -2,
```
StrikerRUS, line 3443: Let's use `-1` for the consistency with the rest codebase.
```suggestion
        "verbose": -1,
```

tests/python_package_test/test_engine.py
```diff
@@ -3424,3 +3424,29 @@ def test_pandas_nullable_dtypes():
...
+        "boost_from_average": True,
+        "min_data_in_leaf": 1,
+    }
+    train_set = lgb.Dataset(X, y)
+    model = lgb.train(params=params, train_set=train_set, num_boost_round=10)
+
+    preds = model.predict(X)
+    mean_preds = np.mean(preds)
+    assert mean_preds >= y.min() and mean_preds <= y.max()
```
StrikerRUS, line 3452:
```suggestion
    assert y.min() <= mean_preds <= y.max()
```

tests/python_package_test/test_engine.py
```diff
@@ -3424,3 +3424,29 @@ def test_pandas_nullable_dtypes():
 
     # test equal predictions
     np.testing.assert_allclose(preds, preds_nullable_dtypes)
+
+
+def test_boost_from_average_with_single_leaf_trees():
+    X = np.array(
+        [[1021.0589, 1018.9578],
```
StrikerRUS, line 3431:
```suggestion
    X = np.array([
        [1021.0589, 1018.9578],
```

StrikerRUS (commented):
Thanks a lot for the test! I can confirm it cannot be passed with `v3.3.2`.

I just left some minor lint and readability suggestions.

tests/python_package_test/test_engine.py
shiyu1994: Resolved via 8412bb3

tests/python_package_test/test_engine.py
shiyu1994: Resolved via 8412bb3

tests/python_package_test/test_engine.py
shiyu1994: Resolved via 8412bb3

tests/python_package_test/test_engine.py
shiyu1994: Resolved via 8412bb3

StrikerRUS (approved):
LGTM, thank you so much for the fix!

jameslamb (approved):
thanks very much for this!

PR #5052, opened 2022-03-04
[python-package] remove 'fobj' in favor of passing custom objective function in params (fixes #3244)

### LINKED ISSUE #3244: [Python] Allowing custom objective / metric function with "objective" and "metric" parameters
## Summary

It would be nice if one could register custom objective and loss functions, so that these can be passed into the LightGBM's  `train` function via the `param` argument. 

## Motivation

I believe that this would be a nice feature as this allows for easier hyperparameter tuning. Currently one would need to write logic that retrieves the custom function based on the name in a config file and pass those into the train function's `feval` and `fobj` arguments. It would be much easier and cleaner if one could just register these custom functions s.t. they can be passed in like the standard objective functions and eval metrics.
## Description

Currently one would do something like this to use custom objective functions and evaluation metrics:

```python
import lightgbm as lgb
from custom import custom_objective, custom_metric

params = {}
gbm = lgb.train(params=params,
                train_set=train_set,
                valid_sets=eval_set,
                fobj=custom_objective,
                feval=custom_metric
                )
```

I would propose to also allow something similar to this:

```python
import lightgbm as lgb
from custom import custom_objective, custom_metric

lgb.register_metric(name="custom_metric", function=custom_metric)
lgb.register_objective(name="custom_objective", function=custom_objective)

params = {
    'objective': 'custom_objective',
    'metric': ['custom_metric'],
}

gbm = lgb.train(params=params,
                train_set=train_set,
                valid_sets=eval_set
                )

```

TremaMiguel:
**Context**

This PR ~closes~ contributes to #3244 

**Changes**

* ~Any custom evaluation passed to `train` through `params['metric'] or `feval` is accepted.~
* Allow passing custome objective functions through `params['objective']`. ~In the case that `params['objective']` and `fobj` are both callable the first one is taken into account.~ `fobj` argument is removed.

**Test**

* Functional tests with both classification and regression cases.


python-package/lightgbm/engine.py
```diff
@@ -125,6 +127,18 @@ def train(
     """
     # create predictor first
     params = copy.deepcopy(params)
+    eval_metric = params.get('metric')
+    # merge metrics through params and feval
+    if eval_metric is not None:
+        eval_metric_list = copy.deepcopy(eval_metric)
+        if not isinstance(eval_metric, list):
+            eval_metric_list = [eval_metric_list] if isinstance(eval_metric_list, str) else list(eval_metric_list)
```
shiyu1994, line 135: When eval_metric is a single callable object, (e.g., when use a callable function for `metric` in parameter), this will result in an error `TypeError: 'function' object is not iterable` since we are converting a function to a list.

tests/python_package_test/test_engine.py
```diff
@@ -2267,6 +2267,92 @@ def test_multiple_feval_train():
     assert 'decreasing_metric' in evals_result['valid_0']
 
 
+def test_params_metric_feval_callable_train():
+    # Test classification
+    X, y = load_breast_cancer(return_X_y=True)
+    params = {'verbose': -1, 'objective': 'binary', 'metric': ['binary_logloss', constant_metric]}
```
shiyu1994, line 2273: Shall we add a case where `metric` is a single customized metric function?

shiyu1994 (changes requested):

shiyu1994:
@TremaMiguel Thanks for working on this! Could you please check my comments above?

python-package/lightgbm/engine.py
```diff
@@ -125,6 +127,18 @@ def train(
...
+            params['metric'] = eval_metric_list
+        eval_metrics_callable, eval_metrics_builtin = _separate_metrics_list(eval_metric_list)
+        params = _concat_params_metrics(params, eval_metrics_builtin)
+        feval = _concat_metric_feval_callables(eval_metrics_callable, feval)
+    # objective can be passed either through params or fobj
```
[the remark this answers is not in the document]
TremaMiguel: got it

python-package/lightgbm/engine.py
TremaMiguel: I see here three cases to handle:

1. `eval_metric_list` is str -> `[eval_metric_list]`
2. `eval_metric_list` is callable -> `[eval_metric_list]`
3.  `eval_metric_list` is a set -> `list(eval_metric_list)`

in case `eval_metric_list` is a tuple then `list(eval_metric_list)` works.

tests/python_package_test/test_engine.py
TremaMiguel: I think we can test single callable with one case either classification or regression to avoid duplicating code 

```python
# for classification
'metric': [constant_metric]
```

TremaMiguel:
@shiyu1994 thanks for the time for reviewing, I've provided some answer to open threads.

shiyu1994:
Close and reopen to retriever CI.

tests/python_package_test/test_engine.py
shiyu1994: I agree. Shall we do this in this PR? Thanks for the fix. This PR should now work for single callable.

TremaMiguel: Got it, just to be consistent `classification` tests single callable from `objective`

shiyu1994 (approved):
Thank you! The changes LGTM now. I'm not sure if other reviewers would like to take a look before this is merging. So gently ping @StrikerRUS @jameslamb @jmoralez @guolinke.

StrikerRUS:
@TremaMiguel Thanks for your work!

Since scikit-learn wrapper internally calls `train()` function,
https://github.com/microsoft/LightGBM/blob/9a4e70687d5c0732ca895959f418c3f923f2e85a/python-package/lightgbm/sklearn.py#L753-L764
we can remove duplicated code that is in `engine.py` now, right?
https://github.com/microsoft/LightGBM/blob/9a4e70687d5c0732ca895959f418c3f923f2e85a/python-package/lightgbm/sklearn.py#L668-L675

StrikerRUS:
Also, please make the same changes for `cv()` function.

python-package/lightgbm/engine.py
```diff
@@ -125,6 +127,20 @@ def train(
     """
     # create predictor first
     params = copy.deepcopy(params)
+    eval_metric = params.get('metric')
```
StrikerRUS, line 130: There are some other names (aliases) for the `metric` param. We should check all of them
https://lightgbm.readthedocs.io/en/latest/Parameters.html#metric

Also, the same is true for the `objective` param.
https://lightgbm.readthedocs.io/en/latest/Parameters.html#objective

StrikerRUS:
These changes is quite complicated and requires more work that it seems from the first glance. So I suggest to split this PR into two ones with changes related to custom objective and metric respectively.

TremaMiguel:
@StrikerRUS thanks for the feedback, I'll split it in three PR:

1. This one address #3244 to support custom objective function in `train` and `cv`.
2. Second one address #3244 to support custom evaluation function in `train` and `cv`.
3. Third one address #5075.

python-package/lightgbm/engine.py
TremaMiguel: got it

StrikerRUS:
@TremaMiguel Nice plan! 👍

TremaMiguel:
Code is ready for a second review please @StrikerRUS @jameslamb

python-package/lightgbm/basic.py
```diff
@@ -294,6 +294,51 @@ def to_string(x):
...
+    params = deepcopy(params)
+    params_metric = deepcopy(params.get('metric'))
+    params_metric = [params_metric] if isinstance(params_metric, (str, type(None))) else params_metric
+    params_metric = [e for e in metrics_builtin if e not in params_metric] + params_metric
+    params_metric = [metric for metric in params_metric if metric is not None]
+    params['metric'] = params_metric
+    return params
+
+
+def _concat_metric_feval_callables(metrics_callable, feval_callable):
```
jameslamb, line 315: Based on https://github.com/microsoft/LightGBM/pull/5052#issuecomment-1065949317, this PR should only contain changes related to callable objective functions, not anything related to `metric`.

Can you please remove this and other changes that are specific to `metric` from this PR?

As you've said in https://github.com/microsoft/LightGBM/pull/5052#issuecomment-1067492446, a separate PR will deal with changes related to `metric`.

python-package/lightgbm/engine.py
```diff
@@ -125,6 +128,7 @@ def train(
     """
     # create predictor first
     params = copy.deepcopy(params)
+    fobj = _objective_is_callable(params, fobj)
```
jameslamb, line 131: I don't believe that adding a new internal function like this is necessary.

I think it would be sufficient, and simpler to add the following here

```python
params = _choose_param_value(
    main_param_name='objective',
    params=params,
    default_value=fobj
)
if callable(params["objective"]):
    fobj = params["objective"]
    params["objective"] = 'none'
```

And remove this code right below iit.

```python
if fobj is not None:
    for obj_alias in _ConfigAliases.get("objective"):
        params.pop(obj_alias, None)
    params['objective'] = 'none'
```

`_choose_param_value()`:

* handles resolving parameter aliases for `objective`
    - which, as of #4829, are sourced via the C API, to keep the Python package and C++ library consistent
* enforces the behavior "values in `params` should be preferred to keyword arguments"

@StrikerRUS what do you think?

python-package/lightgbm/engine.py
```diff
@@ -383,7 +387,8 @@ def cv(params, train_set, num_boost_round=100,
     Parameters
     ----------
     params : dict
-        Parameters for Booster.
+        Parameters for Booster. If callable objective functions is passed through ``params`` and ``fobj``
+        the Booster uses the first one.
```
jameslamb, line 391:
```suggestion
        Parameters for Booster. Values passed through `params` take precedence over those
        supplied via keyword arguments.
```

It's ok with me to add documentation on the `params` argument here, but I think it should be more general (since that behavior isn't specific just to `objective`), and should be more explicit about params vs. keyword arguments (as "the first one" is unclear).

tests/python_package_test/test_engine.py
```diff
@@ -2273,6 +2273,81 @@ def test_multiple_feval_train():
...
+        params=params,
+        train_set=train_dataset,
+        valid_sets=validation_dataset,
+        num_boost_round=5,
+        feval=decreasing_metric,
+        callbacks=[lgb.record_evaluation(evals_result)]
+    )
+    assert len(evals_result['valid_0']) == 2
+    assert 'binary_logloss' in evals_result['valid_0']
+    assert 'decreasing_metric' in evals_result['valid_0']
```
jameslamb, line 2294: Per https://github.com/microsoft/LightGBM/pull/5052#issuecomment-1067492446, please remove tests related to custom metrics, and ensure that the tests added in this PR only contain the code necessary to test the following:

* if a callable is found in `params["objective"]` (or any of the aliases of `"objective"`), that objective is used
* if a callable is found in `params["objective"]` (or any of the aliases of `"objective"`) AND another one is passed via `fobj`, the objective provided to `params` is preferred to the one provided to `fobj`

jameslamb (changes requested):
Thanks very much for this!

Please see my comments about simplifying the implementation, and please remove remaining details left in this PR that are specific to `metric` (per https://github.com/microsoft/LightGBM/pull/5052#issuecomment-1067492446).

python-package/lightgbm/basic.py
TremaMiguel: copy that

python-package/lightgbm/engine.py
StrikerRUS: I like this proposal. The code will look much better I think.

TremaMiguel: @jameslamb I like more that implementation is more clean and consistent and works the same

python-package/lightgbm/engine.py
TremaMiguel: got it

tests/python_package_test/test_engine.py
TremaMiguel: ok @jameslamb, I've added a test `test_choose_param_value_objective` to test this functionality within `_choose_param_value`. 

The reason for this is that `cv` returns a `Booster`, which has a `params` attribute, inside params `objective` should be `none`.

```python
cv_res = lgb.cv(
      params_with_metric,
      lgb_train,
      num_boost_round=10,
      nfold=3,
      stratified=False,
      return_cvbooster=True
)

cv_booster = cv_res['cvbooster'].boosters
cv_objs = [
    cb.params['objective'] == 'none' for cb in cv_booster
]
assert all(cv_objs)
```

TremaMiguel: @jameslamb there's one caveat with `_choose_param_value`, when `objective` is string and `fobj` callable, the second one should take precedence.

python-package/lightgbm/engine.py
```diff
@@ -486,12 +492,17 @@ def cv(params, train_set, num_boost_round=100,
...
     if not isinstance(train_set, Dataset):
         raise TypeError("Training only accepts Dataset object")
-
     params = copy.deepcopy(params)
-    if fobj is not None:
-        for obj_alias in _ConfigAliases.get("objective"):
-            params.pop(obj_alias, None)
-        params['objective'] = 'none'
+    if isinstance(params.get('objective'), str) and callable(fobj):
+        params['objective'] = fobj
```
jameslamb, line 497: The GitHub UI is not allowing me to add a new comment under https://github.com/microsoft/LightGBM/pull/5052#discussion_r830513129, so I'll continue that thread over here on the relevant code.

> @jameslamb there's one caveat with _choose_param_value, when objective is string and fobj callable, the second one should take precedence

I see you're referring to the code below, where if `fobj` is not `None`, it is preferred to anything found in `params`.

https://github.com/microsoft/LightGBM/blob/8e721c564d4534820a7b7febf81382ccf247b5c9/python-package/lightgbm/engine.py#L127-L131

I personally support making the breaking change in this PR that if `fobj` is not `None` and if a value for `objective` or its aliases is found in `params`, the value from `params` is preferred. I think that across all places in the Python and R packages, LightGBM should try to provide the behavior "values provided via `params` override keyword arguments".

Let's get others' opinions before making changes though.

-------


@StrikerRUS @shiyu1994 @guolinke  what do you think? Given the following:

```python
lgb.train(
    params={"objective": _other_custom_objective_function},
    fobj=_some_custom_objective_function,
    ...
)
```

- **option 1**: use the function passed to `fobj` (current behavior)
- **option 2**: use the function passed to `params` (breaking change)
- **option 3**: raise an error about conflicting objectives (breaking change)
- **option 4**: something else

My preference is for **option 2**, for consistency with how `params` is used in most other places in the Python and R packages.

jameslamb (changes requested):
Thanks for your work on this! Please see my next round of suggestions.

tests/python_package_test/test_engine.py
StrikerRUS: Can we remove `fjob` separate argument from `train()` and `cv()` functions right now in this PR given that our next `4.0.0` release will contain a lot of breaking changes?
https://github.com/microsoft/LightGBM/issues/3244#issuecomment-1056004279
@jameslamb @jmoralez @shiyu1994 @guolinke

python-package/lightgbm/engine.py
StrikerRUS: Oh, I'm sorry, I hadn't noticed this comment before writing my question here https://github.com/microsoft/LightGBM/pull/5052#discussion_r830537218.
I think we can remove `fjob` argument and work only with `objective` parameter (and its' aliases).

jameslamb: Sorry to split it somewhere else. I don't know why GitHub wasn't letting me add a new comment in that other thread (it seems like I can now).

Let's resolve this and continue the conversation in that other thread.

tests/python_package_test/test_engine.py
jameslamb: I support removing `fobj` right now, as part of this PR.

At this point, it still seems like v4.0.0 is a long way away and might contain many breaking changes. Given that it might be a long time until the 4.0.0 release, I don't think this project is able right now to do a deprecation warning followed by a removal in a separate release.

python-package/lightgbm/engine.py
StrikerRUS: Yeah, I noticed that GitHub behaves differently (oddly) depending on where you're trying to write comments: on the `Conversation` or `Files changed` tab.

image

tests/python_package_test/test_engine.py
StrikerRUS: Just a note: we should preserve this clarification of callable objective signature
https://github.com/microsoft/LightGBM/blob/d4cdbcfe1b1db2d48e95665a8b358028c496cdb2/python-package/lightgbm/engine.py#L54-L72

I think we can move to the bottom of the docstring and convert into a `Note` section similarly how it's done in the sklearn wrapper.

https://github.com/microsoft/LightGBM/blob/d4cdbcfe1b1db2d48e95665a8b358028c496cdb2/python-package/lightgbm/sklearn.py#L442

jameslamb: Agree! @TremaMiguel can you please remove `fobj` from `train()` and `cv()`? You can update any tests that use it to instead take in a callable objective function via `params`.

jameslamb:
Changed the label for this to `breaking`, based on https://github.com/microsoft/LightGBM/pull/5052#discussion_r830537218.

python-package/lightgbm/engine.py
```diff
@@ -41,7 +41,8 @@ def train(
     Parameters
     ----------
     params : dict
-        Parameters for training.
+        Parameters for Booster. Values passed through `params` take precedence over those
```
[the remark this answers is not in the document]
TremaMiguel: copy that

python-package/lightgbm/engine.py
```diff
@@ -383,7 +388,8 @@ def cv(params, train_set, num_boost_round=100,
     Parameters
     ----------
     params : dict
-        Parameters for Booster.
+        Parameters for Booster. Values passed through `params` take precedence over those
```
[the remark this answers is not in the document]
TremaMiguel: copy that

tests/python_package_test/test_basic.py
```diff
@@ -511,6 +519,36 @@ def test_choose_param_value():
...
+        return np.ones(preds.shape), np.ones(preds.shape)
+
+    def mse_obj(y_pred, dtrain):
+        y_true = dtrain.get_label()
+        grad = (y_pred - y_true)
+        hess = np.ones(len(grad))
+        return grad, hess
+
+    # If callable is found in objective
+    params = {'objective': dummy_obj}
```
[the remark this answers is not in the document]
TremaMiguel: Thanks for spotting this one

tests/python_package_test/test_basic.py
```diff
@@ -511,6 +519,36 @@ def test_choose_param_value():
...
+    params = lgb.basic._choose_param_value(
+        main_param_name="objective",
+        params=params,
+        default_value=None
+    )
+    assert params['objective'] == dummy_obj
+
+    # If callable in objective and fobj prefer 'objective'
```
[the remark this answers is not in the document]
TremaMiguel: copy that

tests/python_package_test/test_basic.py
```diff
@@ -511,6 +519,36 @@ def test_choose_param_value():
...
+    params = {'objective': dummy_obj}
+    params = lgb.basic._choose_param_value(
+        main_param_name="objective",
+        params=params,
+        default_value=mse_obj
+    )
+    assert params['objective'] == dummy_obj
+
```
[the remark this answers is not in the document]
TremaMiguel: got it

tests/python_package_test/test_engine.py
```diff
@@ -2273,6 +2280,75 @@ def test_multiple_feval_train():
...
+    )
+    cv_booster = cv_res['cvbooster'].boosters
+    cv_objs = [
+        cb.params['objective'] == 'none' for cb in cv_booster
+    ]
+    assert all(cv_objs)
+
+    # Test regression
+    X_train, y_train = make_synthetic_regression()
+    params = {'verbose': -1}
```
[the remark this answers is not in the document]
TremaMiguel: copy that

tests/python_package_test/test_engine.py
```diff
@@ -2273,6 +2280,75 @@ def test_multiple_feval_train():
...
+    booster = lgb.train(
+        params=params,
+        train_set=train_dataset,
+        valid_sets=validation_dataset,
+        num_boost_round=5
+    )
+    assert booster.params['objective'] == 'none'
+
+    # Test regression
+    X, y = load_boston(return_X_y=True)
```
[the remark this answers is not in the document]
TremaMiguel: got it

tests/python_package_test/test_engine.py
```diff
@@ -2273,6 +2280,75 @@ def test_multiple_feval_train():
     assert 'decreasing_metric' in evals_result['valid_0']
 
 
+def test_objective_callable_train():
+    # Test classification
+    X, y = load_breast_cancer(return_X_y=True)
+    params = {'verbose': -1, 'objective': dummy_obj, 'metric': 'binary_logloss'}
+    X_train, X_validation, y_train, y_validation = train_test_split(X, y, test_size=0.2)
```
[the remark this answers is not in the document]
TremaMiguel: got it

tests/python_package_test/test_engine.py
```diff
@@ -2273,6 +2280,75 @@ def test_multiple_feval_train():
...
+    X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.1, random_state=42)
+    lgb_train = lgb.Dataset(X_train, y_train)
+    lgb_eval = lgb.Dataset(X_test, y_test, reference=lgb_train)
+    booster = lgb.train(
+        params,
+        lgb_train,
+        num_boost_round=5,
+        valid_sets=lgb_eval,
+    )
+    assert booster.params['objective'] == 'none'
```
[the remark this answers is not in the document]
TremaMiguel: @jameslamb ok, I've added the changes for all 4 cases, though, not sure they should assert a fix value, i.e. 

```python
logloss_error = log_loss(y, y_pred)
rocauc_error = roc_auc_score(y, y_pred)
assert booster.params['objective'] == 'none'
assert logloss_error == pytest.approx(0.25, 0.1)
assert rocauc_error == pytest.approx(0.99, 0.5)
```

tests/python_package_test/test_engine.py
TremaMiguel: @jameslamb @StrikerRUS what is the ETA for `v4.0.0`? will this PR remain open until then?

StrikerRUS: @TremaMiguel 
Unfortunately, there is no ETA yet. But because `v4.0.0` release will be the **next release** this PR will be merged into the `master` branch right after approving.

TremaMiguel: got it thanks for the info

TremaMiguel:
Ready for another feedback loop @jameslamb @StrikerRUS

tests/python_package_test/test_engine.py
```diff
@@ -2278,6 +2311,99 @@ def test_multiple_feval_train():
     assert 'decreasing_metric' in evals_result['valid_0']
 
 
+def test_objective_callable_train():
```
jameslamb, line 2314: can you please split these new tests up by learning task?

* `test_objective_callable_train_regression`
* `test_objective_callable_train_binary_classification`
* `test_objective_callable_cv_regression`
* `test_objective_callable_cv_binary_classification`

Since I don't see any logic shared between the regression and classification code proposed in these new tests, I think splitting them up is preferable. Finer-grained tests make debugging a bit faster, as "all of the custom-objective-function tests are failing" has different possible root causes than, say, "only the regression custom-objective-function tests are failing".

That pattern has been really helpful for me in working on `lightgbm.dask`. For example

https://github.com/microsoft/LightGBM/blob/17d4e007852b4f0b34b211a674cb26e6751c98e2/tests/python_package_test/test_dask.py#L443

means that I've been able to run _just that test_ like

```shell
pytest tests/python_package_tests/test_dask.py::test_classifier_custom_objective
```

which has really sped up the testing cycle for some types of changes

python-package/lightgbm/sklearn.py
```diff
@@ -597,7 +597,9 @@ def _process_params(self, stage: str) -> Dict[str, Any]:
         if callable(self._objective):
             if stage == "fit":
                 self._fobj = _ObjectiveFunctionWrapper(self._objective)
-            params['objective'] = 'None'  # objective = nullptr for unknown objective
+                params['objective'] = self._fobj
+            elif stage == "predict":
+                params['objective'] = 'None'
```
jameslamb, line 602:
```suggestion
            else:
                params['objective'] = 'None'
```

Instead of adding a new `if` branch here, is it possible to just indent the following line over so that it's only run when `stage == "fit"`? Since I don't think `objective` should ever need to be passed through `objective` when predicting.

https://github.com/microsoft/LightGBM/blob/17d4e007852b4f0b34b211a674cb26e6751c98e2/python-package/lightgbm/sklearn.py#L604

That would make this logic a bit simpler, by avoiding adding another `if` branch.

jameslamb (changes requested):
Thanks very much for the changes.  Please see a few minor comments below. I think this PR is close!

tests/python_package_test/test_engine.py
TremaMiguel: makes sense that tests check a single test case, thanks for providing references, it helps with the learning.

python-package/lightgbm/sklearn.py
TremaMiguel: ok, agree on it.

The idea is to avoid the `_ObjectiveFunctionWrapper` type to be passed as an `objective` parameter when calling `predict`. If you try to convert these type to string to be passed to C api it fails.

tests/python_package_test/test_engine.py
jameslamb: Great! I just marked this thread "unresolved". Please don't mark cobversations resolved before you've puhed the related changes, as that mkes it difficult for reviewers to understand the relationship between comments and commits, and to confirm that all suggestions have been addressed.

TremaMiguel: got it

tests/python_package_test/test_engine.py
```diff
@@ -29,6 +30,34 @@ def dummy_obj(preds, train_data):
...
+def logloss_obj(preds, train_data):
+    """Taken from https://maxhalford.github.io/blog/lightgbm-focal-loss/"""
+    y = train_data.get_label()
+    p = special.expit(preds)
+    grad = p - y
+    hess = p * (1 - p)
+    return grad, hess
+
+
+def logloss_metric(preds, train_data):
```
jameslamb, line 49: as mentioned in https://github.com/microsoft/LightGBM/pull/5052#pullrequestreview-913851568, please remove anything related to custom *metric* from this PR, to keep it focused only on custom objective functions

tests/python_package_test/test_engine.py
```diff
@@ -2278,6 +2311,99 @@ def test_multiple_feval_train():
...
+        'verbose': -1,
+        'objective': logloss_obj,
+        'learning_rate': 0.01
+    }
+    train_dataset = lgb.Dataset(X, y)
+    booster = lgb.train(
+        params=params,
+        train_set=train_dataset,
+        num_boost_round=100,
+        feval=logloss_metric
```
jameslamb, line 2326: as mentioned in https://github.com/microsoft/LightGBM/pull/5052#pullrequestreview-913851568, please remove any changes from this PR related to custom *metric* and keep it focused only on custom objective functions

tests/python_package_test/test_engine.py
```diff
@@ -29,6 +30,34 @@ def dummy_obj(preds, train_data):
...
 
 
+def mse_obj(y_pred, dtrain):
+    y_true = dtrain.get_label()
+    grad = (y_pred - y_true)
+    hess = np.ones(len(grad))
+    return grad, hess
+
+
+def logloss_obj(preds, train_data):
```
jameslamb, line 40: to avoid introducing a new import from a dependency (`scipy.special`), could you please just copy over the existing implementation from the Dask tests which uses only `numpy`?

https://github.com/microsoft/LightGBM/blob/d163c2c16629e3da41dc5d7b8dbb76b8fc6249db/tests/python_package_test/test_dask.py#L241-L245

(but modified to take in `(preds, train_data)` instead of `(y_true, y_pred)`)

jameslamb (changes requested):
Thank you very much for breaking up the tests. Please see my most recent comments about some other things I missed in previous reviews.

tests/python_package_test/test_engine.py
TremaMiguel: copy that

tests/python_package_test/test_engine.py
TremaMiguel: copy that

tests/python_package_test/test_engine.py
TremaMiguel: copy that

python-package/lightgbm/engine.py
```diff
@@ -486,12 +492,16 @@ def cv(params, train_set, num_boost_round=100,
...
-
     params = copy.deepcopy(params)
-    if fobj is not None:
-        for obj_alias in _ConfigAliases.get("objective"):
-            params.pop(obj_alias, None)
-        params['objective'] = 'none'
+    params = _choose_param_value(
+        main_param_name='objective',
+        params=params,
+        default_value=None
```
StrikerRUS, line 499: Same question as above for `default_value=None`.

python-package/lightgbm/sklearn.py
```diff
@@ -597,7 +597,9 @@ def _process_params(self, stage: str) -> Dict[str, Any]:
         if callable(self._objective):
             if stage == "fit":
                 self._fobj = _ObjectiveFunctionWrapper(self._objective)
```
StrikerRUS, line 599: We don't need `self._fobj` anymore.
```suggestion
```

P.S. GitHub doesn't allow to make suggestion for the piece of code where lines were both removed and added, so my suggestion is in a form of three consecutive comments.

python-package/lightgbm/sklearn.py
```diff
@@ -597,7 +597,9 @@ def _process_params(self, stage: str) -> Dict[str, Any]:
...
-            params['objective'] = 'None'  # objective = nullptr for unknown objective
+                params['objective'] = self._fobj
```
StrikerRUS, line 600:
```suggestion
                params['objective'] = _ObjectiveFunctionWrapper(self._objective)
```

python-package/lightgbm/sklearn.py
```diff
@@ -597,7 +597,9 @@ def _process_params(self, stage: str) -> Dict[str, Any]:
...
+            else:
+                params['objective'] = 'None'
         else:
             if stage == "fit":
                 self._fobj = None
```
StrikerRUS, line 605:
```suggestion
```

StrikerRUS (changes requested):
Nice work! 💪 
Please consider checking the first round of my review comments.

python-package/lightgbm/sklearn.py
StrikerRUS: > Since I don't think `objective` should ever need to be passed through `objective` when predicting.

Then I strongly prefer just `pop()` `objective` key because `'None'` has special meaning in LightGBM.

tests/python_package_test/test_engine.py
```diff
@@ -2278,6 +2297,98 @@ def test_multiple_feval_train():
...
+    params = {
+        'verbose': -1,
+        'objective': logloss_obj,
+        'learning_rate': 0.01
+    }
+    train_dataset = lgb.Dataset(X, y)
+    booster = lgb.train(
+        params=params,
+        train_set=train_dataset,
+        num_boost_round=100
```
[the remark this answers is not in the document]
TremaMiguel: Ok, checking out the other examples from `test_engine.py`, a reasonable value can be 20 or 25.

python-package/lightgbm/engine.py
```diff
@@ -28,7 +28,6 @@ def train(
     num_boost_round: int = 100,
     valid_sets: Optional[List[Dataset]] = None,
     valid_names: Optional[List[str]] = None,
-    fobj: Optional[_LGBM_CustomObjectiveFunction] = None,
```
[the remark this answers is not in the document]
TremaMiguel: copy that

python-package/lightgbm/engine.py
```diff
@@ -41,7 +40,8 @@ def train(
     Parameters
     ----------
     params : dict
-        Parameters for training.
+        Parameters for Booster. Values passed through ``params`` take precedence over those
```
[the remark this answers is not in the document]
TremaMiguel: copy that

python-package/lightgbm/engine.py
```diff
@@ -41,7 +40,8 @@ def train(
...
+        supplied via keyword arguments.
```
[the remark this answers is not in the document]
TremaMiguel: got it

python-package/lightgbm/engine.py
```diff
@@ -50,27 +50,6 @@ def train(
...
-                The value of the first order derivative (gradient) of the loss
-                with respect to the elements of preds for each sample point.
-            hess : numpy 1-D array or numpy 2-D array (for multi-class task)
-                The value of the second order derivative (Hessian) of the loss
-                with respect to the elements of preds for each sample point.
-
-        For multi-class task, preds are numpy 2-D array of shape = [n_samples, n_classes],
-        and grad and hess should be returned in the same format.
-
     feval : callable, list of callable, or None, optional (default=None)
```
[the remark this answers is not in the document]
TremaMiguel: makes sense, thanks for catching this

python-package/lightgbm/engine.py
```diff
@@ -118,17 +97,43 @@ def train(
...
     # create predictor first
     params = copy.deepcopy(params)
-    if fobj is not None:
-        for obj_alias in _ConfigAliases.get("objective"):
-            params.pop(obj_alias, None)
-        params['objective'] = 'none'
+    params = _choose_param_value(
+        main_param_name='objective',
+        params=params,
+        default_value=None
```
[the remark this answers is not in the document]
TremaMiguel: @StrikerRUS with `default_value='none'` user will be warned about it

> lightgbm.basic.LightGBMError: No object function provided

so he knows that he should provide an objective function to train.

python-package/lightgbm/engine.py
```diff
@@ -383,7 +388,8 @@ def cv(params, train_set, num_boost_round=100,
     Parameters
     ----------
     params : dict
-        Parameters for Booster.
+        Parameters for Booster. Values passed through ``params`` take precedence over those
+        supplied via keyword arguments.
```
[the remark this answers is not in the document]
TremaMiguel: got it

python-package/lightgbm/engine.py
```diff
@@ -403,27 +409,6 @@ def cv(params, train_set, num_boost_round=100,
     metrics : str, list of str, or None, optional (default=None)
         Evaluation metrics to be monitored while CV.
         If not None, the metric in ``params`` will be overridden.
-    fobj : callable or None, optional (default=None)
```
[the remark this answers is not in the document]
TremaMiguel: ups, thanks for catching this one

python-package/lightgbm/engine.py
```diff
@@ -403,27 +409,6 @@ def cv(params, train_set, num_boost_round=100,
...
-                The value of the first order derivative (gradient) of the loss
-                with respect to the elements of preds for each sample point.
-            hess : numpy 1-D array or numpy 2-D array (for multi-class task)
-                The value of the second order derivative (Hessian) of the loss
-                with respect to the elements of preds for each sample point.
-
-        For multi-class task, preds are numpy 2-D array of shape = [n_samples, n_classes],
-        and grad and hess should be returned in the same format.
-
     feval : callable, list of callable, or None, optional (default=None)
```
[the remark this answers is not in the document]
TremaMiguel: copy that

python-package/lightgbm/engine.py
TremaMiguel: this could work the same as before with `default_value='none'`, so that user knows he needs an `object function`. 

One possibility could be that `CVBooster` handles a similar logic as `LGBMModel._process_params`

```python
if self._objective is None:
    if isinstance(self, LGBMRegressor):
        self._objective = "regression"
    elif isinstance(self, LGBMClassifier):
        if self._n_classes > 2:
            self._objective = "multiclass"
        else:
            self._objective = "binary"
    elif isinstance(self, LGBMRanker):
        self._objective = "lambdarank"
    else:
        raise ValueError("Unknown LGBMModel type.")
```

python-package/lightgbm/sklearn.py
TremaMiguel: copy that

python-package/lightgbm/sklearn.py
TremaMiguel: got it

python-package/lightgbm/sklearn.py
TremaMiguel: copy that

python-package/lightgbm/engine.py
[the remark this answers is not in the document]
TremaMiguel: with `default_value=None`  the log is

 > [LightGBM] [Warning] Auto-choosing col-wise multi-threading, the overhead of testing was 0.000036 seconds.You can set `force_col_wise=true` to remove the overhead.
[LightGBM] [Info] Total Bins 90
[LightGBM] [Info] Number of data points in the train set: 50, number of used features: 5
[LightGBM] [Info] Start training from score 0.620336
[LightGBM] [Warning] No further splits with positive gain, best gain: -inf
[LightGBM] [Warning] No further splits with positive gain, best gain: -inf[LightGBM] [Warning] No further splits with positive gain, best gain: -inf
[LightGBM] [Warning] No further splits with positive gain, best gain: -inf
[LightGBM] [Warning] No further splits with positive gain, best gain: -inf
[LightGBM] [Warning] No further splits with positive gain, best gain: -inf
[LightGBM] [Warning] No further splits with positive gain, best gain: -inf
[LightGBM] [Warning] No further splits with positive gain, best gain: -inf[LightGBM] [Warning] No further splits with positive gain, best gain: -inf
[LightGBM] [Warning] No further splits with positive gain, best gain: -inf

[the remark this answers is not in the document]
StrikerRUS: I think we should preserve the current behavior then user is allowed to not specify exact objective and get [`regression` by default](https://lightgbm.readthedocs.io/en/latest/Parameters.html#objective).

Looks like `None` doesn't overwrite default cpp-level `regression` value, right?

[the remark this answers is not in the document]
TremaMiguel: yes, `None` values are ignored by `param_dict_to_str`

```python
from lightgbm.basic import param_dict_to_str
from lightgbm.basic import _choose_param_value

params = {}
params = _choose_param_value(
         main_param_name='objective',
         params=params,
         default_value=None
)
>>> params
{'objective': None}
>>> param_dict_to_str(params)
''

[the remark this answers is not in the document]
Comment (StrikerRUS) [reply python-package/lightgbm/engine.py:127]: Great, thanks for checking that!

**Comment by TremaMiguel** [comment]
@StrikerRUS ready for a second round of comments.

File: tests/python_package_test/test_engine.py [hunk 37]
```diff
@@ -2278,6 +2297,98 @@ def test_multiple_feval_train():
...
+def test_objective_callable_cv_regression():
+    X, y = make_synthetic_regression()
+    lgb_train = lgb.Dataset(X, y)
+    params_with_metric = {
+        'verbose': -1,
+        'objective': mse_obj,
+        'metric': 'l2'
+    }
+    cv_res = lgb.cv(
+        params_with_metric,
```
StrikerRUS, line 2374: Looks like `metric` param isn't used in this test.
```suggestion
    params = {
        'verbose': -1,
        'objective': mse_obj
    }
    cv_res = lgb.cv(
        params,
```

tests/python_package_test/test_engine.py
```diff
@@ -2334,7 +2445,8 @@ def custom_obj(y_pred, ds):
     builtin_obj_bst = lgb.train(params, ds, num_boost_round=10)
     builtin_obj_preds = builtin_obj_bst.predict(X)
 
-    custom_obj_bst = lgb.train(params, ds, num_boost_round=10, fobj=custom_obj)
+    params = {'objective': custom_obj, 'num_class': 3, 'num_leaves': 7}
```
StrikerRUS, line 2448:
```suggestion
    params['objective'] = custom_obj
```

StrikerRUS (changes requested):
Thanks a lot for addressing my previous comments!
Here is a second round of review related to tests.

tests/python_package_test/test_basic.py
```diff
@@ -513,6 +513,45 @@ def test_choose_param_value():
...
+@pytest.mark.parametrize("objective_alias", lgb.basic._ConfigAliases.get("objective"))
+def test_choose_param_value_objective(objective_alias):
+    def dummy_obj(preds, train_data):
+        return np.ones(preds.shape), np.ones(preds.shape)
+
+    def mse_obj(y_pred, dtrain):
+        y_true = dtrain.get_label()
+        grad = (y_pred - y_true)
+        hess = np.ones(len(grad))
+        return grad, hess
```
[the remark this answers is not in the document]
TremaMiguel: copy that

tests/python_package_test/test_engine.py
```diff
@@ -1882,7 +1897,7 @@ def test_metrics():
     lgb_valid = lgb.Dataset(X_test, y_test, reference=lgb_train)
 
     evals_result = {}
-    params_verbose = {'verbose': -1}
+    params_verbose = {'verbose': -1, 'objective': dummy_obj}
```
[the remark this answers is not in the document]
TremaMiguel: copy that

tests/python_package_test/test_engine.py
```diff
@@ -1891,11 +1906,11 @@ def test_metrics():
                                        'metric': ['binary_logloss', 'binary_error'],
                                        'verbose': -1}
     params_obj_metric_none_verbose = {'objective': 'binary', 'metric': 'None', 'verbose': -1}
-    params_metric_log_verbose = {'metric': 'binary_logloss', 'verbose': -1}
+    params_metric_log_verbose = {'objective': dummy_obj, 'metric': 'binary_logloss', 'verbose': -1}
```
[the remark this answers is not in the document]
TremaMiguel: copy that

tests/python_package_test/test_engine.py
```diff
@@ -1891,11 +1906,11 @@ def test_metrics():
...
     params_metric_err_verbose = {'metric': 'binary_error', 'verbose': -1}
-    params_metric_inv_verbose = {'metric_types': 'invalid_metric', 'verbose': -1}
-    params_metric_multi_verbose = {'metric': ['binary_logloss', 'binary_error'], 'verbose': -1}
-    params_metric_none_verbose = {'metric': 'None', 'verbose': -1}
+    params_metric_inv_verbose = {'objective': dummy_obj, 'metric_types': 'invalid_metric', 'verbose': -1}
+    params_metric_multi_verbose = {'objective': dummy_obj, 'metric': ['binary_logloss', 'binary_error'], 'verbose': -1}
+    params_metric_none_verbose = {'objective': dummy_obj, 'metric': 'None', 'verbose': -1}
```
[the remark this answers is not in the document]
TremaMiguel: copy that

tests/python_package_test/test_engine.py
```diff
@@ -2190,9 +2205,13 @@ def train_booster(params=params_obj_verbose, **kwargs):
 
     obj_multi_aliases = ['multiclass', 'softmax', 'multiclassova', 'multiclass_ova', 'ova', 'ovr']
     for obj_multi_alias in obj_multi_aliases:
+        # Custom objective replaces multiclass
         params_obj_class_3_verbose = {'objective': obj_multi_alias, 'num_class': 3, 'verbose': -1}
+        params_obj_class_3_custom_obj = {'objective': dummy_obj, 'num_class': 3, 'verbose': -1}
         params_obj_class_1_verbose = {'objective': obj_multi_alias, 'num_class': 1, 'verbose': -1}
+        params_obj_class_1_custom_obj = {'objective': dummy_obj, 'num_class': 1, 'verbose': -1}
         params_obj_verbose = {'objective': obj_multi_alias, 'verbose': -1}
+        params_obj_custom_obj = {'objective': dummy_obj, 'verbose': -1}
```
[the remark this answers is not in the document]
TremaMiguel: got it

tests/python_package_test/test_engine.py
TremaMiguel: got it

tests/python_package_test/test_engine.py
TremaMiguel: got it

python-package/lightgbm/sklearn.py
```diff
@@ -596,11 +596,10 @@ def _process_params(self, stage: str) -> Dict[str, Any]:
                     raise ValueError("Unknown LGBMModel type.")
         if callable(self._objective):
             if stage == "fit":
-                self._fobj = _ObjectiveFunctionWrapper(self._objective)
-            params['objective'] = 'None'  # objective = nullptr for unknown objective
+                params['objective'] = _ObjectiveFunctionWrapper(self._objective)
+            else:
+                params['objective'] = 'None'
```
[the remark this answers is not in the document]
TremaMiguel: @StrikerRUS  If we `pop` objective the following error in `test_sklearn.test_multiclass_custom_objective` raises 

> lightgbm.basic.LightGBMError: Number of classes must be 1 for non-multiclass training

I think the objective task is lost somewhere when predicting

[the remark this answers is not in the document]
TremaMiguel: We could pass the value of `objective` to params based on the class instance, that is,

```python
if callable(self._objective):
    if stage == "fit":
        params['objective'] = _ObjectiveFunctionWrapper(self._objective)
    else:
        # params['objective'] = 'None'
        if isinstance(self, LGBMRegressor):
            params['objective'] =  "regression"
        elif isinstance(self, LGBMClassifier):
            if self._n_classes > 2:
                params['objective'] = "multiclass"
            else:
                params['objective'] = "binary"
        elif isinstance(self, LGBMRanker):
            params['objective'] = "lambdarank"
        else:
            raise ValueError("Unknown LGBMModel type.")
```

but is more clean just to write

```
params['objective'] = 'None'
```

what's the special meaning of `None` ?

tests/python_package_test/test_engine.py
```diff
@@ -1959,32 +1963,32 @@ def train_booster(params=params_obj_verbose, **kwargs):
 
     # fobj, no feval
     # no default metric
-    res = get_cv_result(params=params_verbose, fobj=dummy_obj)
+    res = get_cv_result(params=params_dummy_obj_verbose)
     assert len(res) == 0
 
     # metric in params
-    res = get_cv_result(params=params_metric_err_verbose, fobj=dummy_obj)
```
StrikerRUS, line 1966: Add `dummy_obj` to `params_metric_err_verbose` and rename them into `params_dummy_obj_metric_err_verbose`

tests/python_package_test/test_engine.py
```diff
@@ -2042,47 +2046,47 @@ def train_booster(params=params_obj_verbose, **kwargs):
...
     # fobj, feval
     # no default metric, only custom one
-    res = get_cv_result(params=params_verbose, fobj=dummy_obj, feval=constant_metric)
+    res = get_cv_result(params=params_dummy_obj_verbose, feval=constant_metric)
     assert len(res) == 2
     assert 'valid error-mean' in res
 
     # metric in params with custom one
-    res = get_cv_result(params=params_metric_err_verbose, fobj=dummy_obj, feval=constant_metric)
+    res = get_cv_result(params=params_metric_err_verbose, feval=constant_metric)
```
StrikerRUS, line 2054: Add `dummy_obj` to `params_metric_err_verbose` and rename them into `params_dummy_obj_metric_err_verbose`

tests/python_package_test/test_engine.py
```diff
@@ -2190,9 +2194,13 @@ def train_booster(params=params_obj_verbose, **kwargs):
 
     obj_multi_aliases = ['multiclass', 'softmax', 'multiclassova', 'multiclass_ova', 'ova', 'ovr']
     for obj_multi_alias in obj_multi_aliases:
+        # Custom objective replaces multiclass
         params_obj_class_3_verbose = {'objective': obj_multi_alias, 'num_class': 3, 'verbose': -1}
+        params_dummy_obj_class_3_verbose = {'objective': dummy_obj, 'num_class': 3, 'verbose': -1}
         params_obj_class_1_verbose = {'objective': obj_multi_alias, 'num_class': 1, 'verbose': -1}
```
StrikerRUS, line 2200: Looks like this variable is unused now.

tests/python_package_test/test_engine.py
```diff
@@ -2278,6 +2286,97 @@ def test_multiple_feval_train():
...
+
+def test_objective_callable_cv_regression():
+    X, y = make_synthetic_regression()
+    lgb_train = lgb.Dataset(X, y)
+    params_with_metric = {
+        'verbose': -1,
+        'objective': mse_obj
+    }
+    cv_res = lgb.cv(
+        params_with_metric,
```
StrikerRUS, line 2362: metric was removed.
```suggestion
    params = {
        'verbose': -1,
        'objective': mse_obj
    }
    cv_res = lgb.cv(
        params,
```

python-package/lightgbm/sklearn.py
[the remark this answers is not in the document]
StrikerRUS: OK, I see now.
Seems that LightGBM checks the consistency for objective and number of classes even during predicting.
https://github.com/microsoft/LightGBM/blob/6b56a90cd1324a6dbac2afa0a352c9355b0dc3cf/src/io/config.cpp#L261
Full logs:
```
==================================================================================================== FAILURES ====================================================================================================
________________________________________________________________________________________ test_multiclass_custom_objective ________________________________________________________________________________________

    def test_multiclass_custom_objective():
        centers = [[-4, -4], [4, 4], [-4, 4]]
        X, y = make_blobs(n_samples=1_000, centers=centers, random_state=42)
        params = {'n_estimators': 10, 'num_leaves': 7}
        builtin_obj_model = lgb.LGBMClassifier(**params)
        builtin_obj_model.fit(X, y)
        builtin_obj_preds = builtin_obj_model.predict_proba(X)

        custom_obj_model = lgb.LGBMClassifier(objective=sklearn_multiclass_custom_objective, **params)
        custom_obj_model.fit(X, y)
>       custom_obj_preds = softmax(custom_obj_model.predict(X, raw_score=True))

test_sklearn.py:1299:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
d:\miniconda3\lib\site-packages\lightgbm\sklearn.py:1050: in predict
    result = self.predict_proba(X, raw_score, start_iteration, num_iteration,
d:\miniconda3\lib\site-packages\lightgbm\sklearn.py:1063: in predict_proba
    result = super().predict(X, raw_score, start_iteration, num_iteration, pred_leaf, pred_contrib, **kwargs)
d:\miniconda3\lib\site-packages\lightgbm\sklearn.py:813: in predict
    return self._Booster.predict(X, raw_score=raw_score, start_iteration=start_iteration, num_iteration=num_iteration,
d:\miniconda3\lib\site-packages\lightgbm\basic.py:3538: in predict
    return predictor.predict(data, start_iteration, num_iteration,
d:\miniconda3\lib\site-packages\lightgbm\basic.py:813: in predict
    preds, nrow = self.__pred_for_np2d(data, start_iteration, num_iteration, predict_type)
d:\miniconda3\lib\site-packages\lightgbm\basic.py:903: in __pred_for_np2d
    return inner_predict(mat, start_iteration, num_iteration, predict_type)
d:\miniconda3\lib\site-packages\lightgbm\basic.py:873: in inner_predict
    _safe_call(_LIB.LGBM_BoosterPredictForMat(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

ret = -1

    def _safe_call(ret: int) -> None:
        """Check the return value from C API call.

        Parameters
        ----------
        ret : int
            The return value from C API calls.
        """
        if ret != 0:
>           raise LightGBMError(_LIB.LGBM_GetLastError().decode('utf-8'))
E           lightgbm.basic.LightGBMError: Number of classes must be 1 for non-multiclass training

d:\miniconda3\lib\site-packages\lightgbm\basic.py:142: LightGBMError
---------------------------------------------------------------------------------------------- Captured stdout call ----------------------------------------------------------------------------------------------
[LightGBM] [Warning] Auto-choosing col-wise multi-threading, the overhead of testing was 0.000075 seconds.
[LightGBM] [Info] Total Bins 510
[LightGBM] [Info] Number of data points in the train set: 1000, number of used features: 2
[LightGBM] [Info] Start training from score -1.096614
[LightGBM] [Info] Start training from score -1.099613
[LightGBM] [Info] Start training from score -1.099613
[LightGBM] [Warning] No further splits with positive gain, best gain: -inf
[LightGBM] [Warning] No further splits with positive gain, best gain: -inf
[LightGBM] [Warning] No further splits with positive gain, best gain: -inf
[LightGBM] [Warning] Auto-choosing col-wise multi-threading, the overhead of testing was 0.000071 seconds.
[LightGBM] [Info] Total Bins 510
[LightGBM] [Info] Number of data points in the train set: 1000, number of used features: 2
[LightGBM] [Warning] No further splits with positive gain, best gain: -inf
[LightGBM] [Warning] No further splits with positive gain, best gain: -inf
[LightGBM] [Warning] No further splits with positive gain, best gain: -inf
[LightGBM] [Warning] No further splits with positive gain, best gain: -inf
[LightGBM] [Warning] No further splits with positive gain, best gain: -inf
---------------------------------------------------------------------------------------------- Captured stderr call ----------------------------------------------------------------------------------------------
[LightGBM] [Fatal] Number of classes must be 1 for non-multiclass training
```

Let's then pass `'None'` for the custom object during `predict` phase.

StrikerRUS (changes requested):
Some more minor suggestions for tests.

python-package/lightgbm/sklearn.py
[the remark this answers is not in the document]
TremaMiguel: makes sense

tests/python_package_test/test_engine.py
TremaMiguel: got it

tests/python_package_test/test_engine.py
TremaMiguel: copy that

tests/python_package_test/test_engine.py
TremaMiguel: copy that

tests/python_package_test/test_engine.py
TremaMiguel: copy that

TremaMiguel:
@StrikerRUS thanks for the time taken to review, I've addressed your comments. Let me know if something else is missing.

StrikerRUS (approved):
Thank you so much for this hard work with API and test changes! LGTM!

StrikerRUS:
@jameslamb @shiyu1994 @jmoralez @guolinke Please help with another pair of eyes.

TremaMiguel:
Perhaps a helping hand here @jameslamb ?

jameslamb (approved):
I reviewed the changes and threads tonight, I have no additional comments. Thanks very much for all the hard work on this @TremaMiguel , and as always for the very thorough review @StrikerRUS .

jameslamb:
I changed the PR title to explicitly include `fobj`, so it's easier to find for people looking through the release notes trying to understand why their code is broken.

TremaMiguel:
@jameslamb @StrikerRUS @shiyu1994 thank you very much for reviewing this.

StrikerRUS:
@TremaMiguel Let's go further with `feval` according to your [plan](https://github.com/microsoft/LightGBM/pull/5052#issuecomment-1067492446)! 😃

bfan1256:
Is there a specific way to use this with .train() I've tried to specifically set params['objective'] but get Unknown type parameter:objective got:function

And I'm simply passing a function reference

jmoralez:
Hey @bfan1256, you can find an example on how to use it [here](https://github.com/microsoft/LightGBM/blob/20975badd7ea3083c89f08a1f4dfda1a9789f6bc/tests/python_package_test/test_engine.py#L2633-L2654). If that doesn't work for you could you please open a new issue?

PR #5059, opened 2022-03-07
Build Windows artifacts in `windows-2019` image instead of `vs2017-win2016`

### LINKED ISSUE #4748: `windows-2016` Azure Pipelines environment is deprecated
> The windows-2016 environment will be deprecated on November 15, 2021, and removed on March 15, 2022. Migrate to windows-latest instead. For more details see https://github.com/actions/virtual-environments/issues/4312

We use this environment here
https://github.com/microsoft/LightGBM/blob/96ecab6fb16df02c084656a3625d513d8dfde551/.vsts-ci.yml#L260

The main problem is that that environment is used right now to create release artifacts: LightGBM executable, dll, SWIG jar and `win_amd64` Python wheel file.
https://github.com/microsoft/LightGBM/blob/96ecab6fb16df02c084656a3625d513d8dfde551/.vsts-ci.yml#L287-L288
So we should be careful during migration to not accidentally change LightGBM installation requirements (refer to #1708, #2889 as an examples of problems we may face).

StrikerRUS:
Closed #4748.

Refer to https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml and https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md.

I'm not sure but it looks like our new Windows artifacts won't be compatible with VC Redistributable 2015 and 2017.

> Visual Studio versions since Visual Studio 2015 share the same redistributable files. For example, any apps built by the Visual Studio 2015, 2017, 2019, or 2022 toolsets can use the latest Microsoft Visual C++ Redistributable. However, the version of the Microsoft Visual C++ redistributable installed on the machine must be the same or higher than the version of the Visual C++ toolset used to create your application. 
   https://docs.microsoft.com/en-us/cpp/windows/latest-supported-vc-redist?view=msvc-170#notes

> For example, Visual Studio 2019 redistributable files may be used by apps built by using the Visual Studio 2017 or 2015 toolset. While they may be compatible, we don't support using older redistributable files in apps built by using a newer toolset. For example, using the 2017 redistributable files in apps built by using the 2019 toolset isn't supported.
   https://docs.microsoft.com/en-us/cpp/windows/determining-which-dlls-to-redistribute?view=msvc-170

### DIFF
--- a/.vsts-ci.yml
+++ b/.vsts-ci.yml
@@ -262,7 +262,7 @@ jobs:
 - job: Windows
 ###########################################
   pool:
-    vmImage: 'vs2017-win2016'
+    vmImage: 'windows-2019'
   strategy:
     matrix:
       regular:


StrikerRUS:
This PR should be merged before March 15 to not break our CI services.

StrikerRUS:
There is an ongoing discussion for supporting Visual Studio 2017 in `windows-2019` image. Probably worth a subscription:
- https://github.com/actions/virtual-environments/issues/4312#issuecomment-1059415201
- https://github.com/actions/virtual-environments/issues/5177

jameslamb (approved):
Thanks for this! I agree that it seems we have no other option, given the significant effort it seems would be required to use Windows containers and the lack of interest from Microsoft/GitHub in the various issues linked to this PR.

guolinke (approved):

PR #5065, opened 2022-03-10
[python-package] Better column dtype logging when column has "bad dtype"

### LINKED ISSUE #5064: [Python] Improve bad pandas dtype error messages
# Motivation:
`_get_bad_pandas_dtypes` is used in a number of places in the codebase to get the indices of `pd.DataFrame` dtypes that do not conform to certain requirements.

The messages of the errors that are then raised provide information that "bad dtypes" were found but not what those dtypes were. Furthermore, only in 1 of 4 instances does the error message provide info on the column(s) the "bad dtypes" were found in.

# Ideas:
Just add something to the error message that shows a link between each `(bad_column, bad_column_dtype)` pair

hsorsky:
Closes #5064

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -513,7 +513,7 @@ def _check_for_bad_pandas_dtypes(pandas_dtypes_series):
         if not is_allowed_numpy_dtype(pandas_dtype.type)
     ]
     if bad_pandas_dtypes:
-        raise ValueError('DataFrame.dtypes must be int, float or bool.\n'
+        raise ValueError('pandas dtypes must be int, float or bool.\n'
                          f'Fields with bad pandas dtypes: {", ".join(bad_pandas_dtypes)}')
 
 
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -184,7 +184,8 @@ def list_to_1d_numpy(data, dtype=np.float32, name='list'):
         return np.array(data, dtype=dtype, copy=False)
     elif isinstance(data, pd_Series):
         if _get_bad_pandas_dtypes([data.dtypes]):
-            raise ValueError('Series.dtypes must be int, float or bool')
+            raise ValueError(f'Series.dtypes must be int, float or bool. Series '
+                             f'\'{data.name}\' found to have dtype \'{data.dtypes}\'')
         return np.array(data, dtype=dtype, copy=False)  # SparseArray should be supported as well
     else:
         raise TypeError(f"Wrong type({type(data).__name__}) for {name}.\n"
@@ -217,8 +218,12 @@ def _data_to_2d_numpy(data: Any, dtype: type = np.float32, name: str = 'list') -
     if _is_2d_list(data):
         return np.array(data, dtype=dtype)
     if isinstance(data, pd_DataFrame):
-        if _get_bad_pandas_dtypes(data.dtypes):
-            raise ValueError('DataFrame.dtypes must be int, float or bool')
+        bad_indices = _get_bad_pandas_dtypes(data.dtypes)
+        if bad_indices:
+            bad_dtypes = data.dtypes.iloc[bad_indices]
+            bad_dtypes_str = str(list(bad_dtypes.items()))
+            raise ValueError('DataFrame.dtypes must be int, float or bool.\n'
+                             f'Fields with bad data types: {bad_dtypes_str}')
         return cast_numpy_array_to_dtype(data.values, dtype)
     raise TypeError(f"Wrong type({type(data).__name__}) for {name}.\n"
                     "It should be list of lists, numpy 2-D array or pandas DataFrame")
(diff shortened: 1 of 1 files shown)


hsorsky:
Current failing tests look (seemingly) unrelated to this PR AFAICT. It's failing on `test_create_tree_digraph`

jmoralez:
Thank you for your contribution! What do you think about changing the `_get_bad_pandas_dtypes` function instead? It could return something like `[f'{name}: {dtype}' for name, dtype in dtypes.iteritems() if not is_allowed_numpy_dtype(dtype.type)]` instead and we could use that list like:
```python
bad_dtypes = _get_bad_pandas_dtypes(df.dtypes)
if bad_dtypes:
    raise ValueError(f"Bad dtypes: {', '.join(bad_dtypes)}")
```

hsorsky:
> What do you think about changing the _get_bad_pandas_dtypes function instead?

The reason I didn't initially do that was that `_get_bad_pandas_dtypes` is used both on the output of `DataFrame.dtypes` (i.e. a Series) but also on `[Series.dtypes]` (i.e. a list of `dtype` objects), in which case there, unfortunately, isn't an attached column name. I wanted to start off with the PR being as unintrusive to the rest of the code as possible, but if maintainers are happy to update either `_get_bad_pandas_dtypes` to handle that case, or the code that calls it like `_get_bad_pandas_dtypes([Series.dtypes])` to ensure the data is passed in as a series of dtypes (with just one element), then I'm happy to make those changes. 

My opinion is that the latter is cleaner and that we should integrate that with your approach.

My thinking for making the func get called on a series would be to do `_get_bad_pandas_dtypes(Series.to_frame().dtypes)` instead of `_get_bad_pandas_dtypes([Series.dtypes])`

jmoralez:
I'm +1 on using `Series.to_frame().dtypes` for the label case, it seems to be very fast and allows the function to stay the same. Let's wait for @StrikerRUS's opinion though.

StrikerRUS:
I'm for moving duplicated post-processing code into the `_get_bad_pandas_dtypes()` function.

StrikerRUS:
@hsorsky 
> Current failing tests look (seemingly) unrelated to this PR AFAICT. It's failing on `test_create_tree_digraph`

Sorry for the inconvenience! Fails should be fixed by #5068.

hsorsky:
> I'm for moving duplicated post-processing code into the _get_bad_pandas_dtypes() function.

just to confirm before I do any more work on this tomorrow - you mean moving everything, including error raising, inside the function? I'd be fine with that as it reduces a large amount of similar code, and we don't lose much in the error message (we can just treat them all like the DataFrame based error messages and it should be obvious from the stack trace that it was in fact a series in the singular series case)

StrikerRUS:
Sounds like a good plan! Also, maybe with the aim to not confuse users with DataFrame/Series words, we can refer to dtypes as "pandas dtypes" or something similar?

python-package/lightgbm/basic.py
```diff
@@ -509,7 +507,14 @@ def is_allowed_numpy_dtype(dtype):
             and not issubclass(dtype, (np.timedelta64, float128))
         )
 
-    return [i for i, dtype in enumerate(dtypes) if not is_allowed_numpy_dtype(dtype.type)]
+    bad_pandas_dtypes = [
+        f'{column_name}: {pandas_dtype}'
+        for column_name, pandas_dtype in pandas_dtypes_series.iteritems()
+        if not is_allowed_numpy_dtype(pandas_dtype)
```
jmoralez, line 513:
```suggestion
        if not is_allowed_numpy_dtype(pandas_dtype.type)
```

python-package/lightgbm/basic.py
```diff
@@ -509,7 +507,14 @@ def is_allowed_numpy_dtype(dtype):
...
         )
 
-    return [i for i, dtype in enumerate(dtypes) if not is_allowed_numpy_dtype(dtype.type)]
+    bad_pandas_dtypes = [
+        f'{column_name}: {pandas_dtype}'
+        for column_name, pandas_dtype in pandas_dtypes_series.iteritems()
+        if not is_allowed_numpy_dtype(pandas_dtype.type)
+    ]
+    if bad_pandas_dtypes:
+        raise ValueError('DataFrame.dtypes must be int, float or bool.\n'
```
StrikerRUS, line 516: Maybe replace concrete `DataFrame` with just `pandas` for more general wording that wouldn't be confusing for the case of label as a `Series`?
https://github.com/microsoft/LightGBM/pull/5065#issuecomment-1064708046
```suggestion
        raise ValueError('pandas dtypes must be int, float or bool.\n'
```

StrikerRUS (approved):
LGTM, thank you so much for the help!
Just one nit below:

python-package/lightgbm/basic.py
jameslamb: ^ I agree with this

jameslamb (approved):
Very nice change, thanks for the help with this! I also approve, and think this should be merged once @StrikerRUS 's one additional comment has been addressed.

jmoralez (approved):

hsorsky:
thanks for the reviews!

PR #5066, opened 2022-03-10
[python-package] [R-package] propagate the best iteration of cvbooster into the individual boosters

### LINKED ISSUE #4777: [python] CVBooster.predict does not use the best iteration
I don't know if it is a bug. The `predict` of CVBooster will not use the best iteration as the `num_iteration` when `num_iteration=None`. It will use the best iterations of underlying Boosters, but they are all -1.

jmoralez:
This propagates the `best_iteration` attribute in the Python package and the `best_iter` attribute in the R package from the CV Booster to each individual booster in the boosters list. In the Python package this ensures that calling `CVBooster.predict` uses the best iteration in all boosters (closes #4777).

### DIFF
--- a/R-package/R/lgb.cv.R
+++ b/R-package/R/lgb.cv.R
@@ -434,6 +434,10 @@ lgb.cv <- function(params = list()
     )
     cv_booster$best_score <- cv_booster$record_evals[["valid"]][[first_metric]][[.EVAL_KEY()]][[cv_booster$best_iter]]
   }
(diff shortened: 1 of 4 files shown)


StrikerRUS (approved):
LGTM, thanks a lot!

jameslamb (approved):
looks great, thanks!

Can you please add `[python-package] [R-package]` to the description so people reading the release notes will be able to identify that this PR affected those components?

PR #5071, opened 2022-03-13
[ci] increase timeout on R Hub Solaris job to 4 hours

jameslamb:
Proposes increasing the timeout on the R-package CI job testing on Solaris via R Hub, from 2 hours to 4 hours.

I recently observed a run that took 3h32m, which led to the GitHub Actions job being marked failed, blocking merging of a PR.

https://github.com/microsoft/LightGBM/pull/4827#issuecomment-1066184033

### DIFF
--- a/.github/workflows/r_solaris.yml
+++ b/.github/workflows/r_solaris.yml
@@ -7,7 +7,7 @@ on:
 jobs:
   test:
     name: solaris-cran
-    timeout-minutes: 120
+    timeout-minutes: 240
     runs-on: ubuntu-latest
     container: wch1/r-debug
     env:


StrikerRUS:
Looks like it won't help.

image

Actually, I think the job fails because the second Solaris job isn't starting.
https://github.com/microsoft/LightGBM/blob/f834cab49a3e394969d087bb209d500b70f428dd/.ci/run_rhub_solaris_checks.R#L31-L36

jameslamb:
> Looks like it won't help.

oh sad! Ok yes, looks like you're right. I'll investigate as soon as possible. I did check and can confirm both of those environments are still active, according to `rhub::platforms()`.

> run r-solaris

This job you just triggered won't read the new timeout value, right? I thought that the triggering-comments jobs had to read their configuration only from `master` (https://github.com/microsoft/LightGBM/pull/4397#issuecomment-866790489, #4503).

StrikerRUS:
> This job you just triggered won't read the new timeout value, right?

Yep! Just wonted to check again. Maybe there were temporary problems on the R-Hub side.

StrikerRUS:
> Maybe there were temporary problems on the R-Hub side.

This time CI job (both Solaris R-Hub jobs) passed in 1h...

jameslamb:
oh wow, interesting! Ok, maybe it was just an infrastructure issue with R Hub then!

We can close this PR and try again on #4827. Thanks for investigating.

PR #5083, opened 2022-03-20
clarify no-meaningful-features warning in Dataset construction (fixes #5081)

### LINKED ISSUE #5081: "Warning: no meaningful features" for very small datasets
## Description

For small datasets (e.g. 5 or fewer features, less than 40 observations), `Dataset` construction says that features are "constant" and and not "meaningful", even when features do have variability and are related to the target.

> [LightGBM] [Warning] There are no meaningful features, as all feature values are constant.

## Reproducible example

Observed this while using the Python package.

```python
import lightgbm as lgb

for num_rows in range(10, 100, 5):
    print(f"--- {num_rows} ---")
    X, y = make_regression(
        n_samples=num_rows,
        n_informative=5,
        n_features=5,
        random_state=708
    )
    dtrain = lgb.Dataset(
        data=X,
        label=y,
        params={
            "verbosity": 1,
            "min_data_in_bin": 1
        }
    )
    dtrain.construct()
```

Produces a log like the following:

```text
--- 10 ---
[LightGBM] [Warning] There are no meaningful features, as all feature values are constant.
--- 15 ---
[LightGBM] [Warning] There are no meaningful features, as all feature values are constant.
--- 20 ---
[LightGBM] [Warning] There are no meaningful features, as all feature values are constant.
--- 25 ---
[LightGBM] [Warning] There are no meaningful features, as all feature values are constant.
--- 30 ---
[LightGBM] [Warning] There are no meaningful features, as all feature values are constant.
--- 35 ---
[LightGBM] [Warning] There are no meaningful features, as all feature values are constant.
--- 40 ---
--- 45 ---
--- 50 ---
--- 55 ---
--- 60 ---
--- 65 ---
--- 70 ---
--- 75 ---
--- 80 ---
--- 85 ---
--- 90 ---
--- 95 ---
```

This is surprising, given that those features _are_ informative for the target.
For example, a plain random forest achieves a good fit to the target even for just 20 observations from that same dataset.

```python
from sklearn.ensemble import RandomForestRegressor
from sklearn.metrics import r2_score

X, y = make_regression(
    n_samples=20,
    n_informative=5,
    n_features=5,
    random_state=708
)

reg = RandomForestRegressor(n_estimators=10).fit(X, y)
r2_score(y, reg.predict(X))
# 0.92602
```

I'm able to reproduce this behavior with a different dataset in the R package.

<details><summary>R example (click me)</summary>

```r
library(lightgbm)

data(ChickWeight)

target_col <- "weight"
feature_cols <- c("Time", "Chick")

for (num_rows in seq(10, 100, 5)){
    print(paste0("--- ", num_rows, " ---"))
    X <- data.matrix(ChickWeight[seq_len(num_rows), feature_cols])
    y <- ChickWeight[seq_len(num_rows), target_col]
        data = X
        , label = y
        , params = list(
            verbosity = 1
            , min_data_in_bin = 1
        )
    )
    dtrain$construct()
}
```

```text
[1] "--- 10 ---"
[LightGBM] [Warning] There are no meaningful features, as all feature values are constant.
[1] "--- 15 ---"
[LightGBM] [Warning] There are no meaningful features, as all feature values are constant.
[1] "--- 20 ---"
[LightGBM] [Warning] There are no meaningful features, as all feature values are constant.
[1] "--- 25 ---"
[LightGBM] [Warning] There are no meaningful features, as all feature values are constant.
[1] "--- 30 ---"
[LightGBM] [Warning] There are no meaningful features, as all feature values are constant.
[1] "--- 35 ---"
[LightGBM] [Warning] There are no meaningful features, as all feature values are constant.
[1] "--- 40 ---"
[LightGBM] [Warning] There are no meaningful features, as all feature values are constant.
[1] "--- 45 ---"
[1] "--- 50 ---"
[1] "--- 55 ---"
[1] "--- 60 ---"
[1] "--- 65 ---"
[1] "--- 70 ---"
[1] "--- 75 ---"
[1] "--- 80 ---"
[1] "--- 85 ---"
[1] "--- 90 ---"
[1] "--- 95 ---"
[1] "--- 100 ---"
```

</details>

## Environment info

LightGBM version or commit hash: `v3.3.2` and latest `master` (https://github.com/microsoft/LightGBM/commit/8e721c564d4534820a7b7febf81382ccf247b5c9)

```shell
cd python-package
python setup.py install

cd ..
```

## Additional Comments

I discovered this while attempting to answer https://stackoverflow.com/questions/71285022/why-lightgbm-python-package-gives-bad-prediction-using-for-regression-task.

I understand that supporting very small datasets is not a high priority for LightGBM, but I still think investigating this issue is valuable because:

1. I've observed in issues and Stack Overflow posts that users often use very small datasets to test LightGBM and improve their understanding of it
2. the root issue here might be one that affects larger datasets as well (it might just be less likely)

jameslamb:
Fixes #5081.

During `Dataset` construction, LightGBM may drop features for the following reasons:

* all data would end up in one bin (e.g. all missing or otherwise constant)
* not possible to create any bins the satisfy `min_data_in_bin`
* not possible to create any bins that satisfy `min_data_in_leaf`

https://github.com/microsoft/LightGBM/blob/d4cdbcfe1b1db2d48e95665a8b358028c496cdb2/src/io/bin.cpp#L493-L502

If all features in the raw data are dropped for these reasons, the following warning is issued

> [LightGBM] [Warning] There are no meaningful features, as all feature values are constant.

Note that that warning only cites one of the possible reasons a feature is excluded (all values are constant). As a result, sometimes this warning can be confusing or misleading. See the example in the description of #5081, where for a very small dataset (<40 rows), LightGBM excludes all features and issues this warning unless `min_data_in_bin=1, min_data_in_leaf=1` are set.

This PR proposes clarifying the language in this warning.

## Notes for Reviewers

@StrikerRUS @shiyu1994 @guolinke why is this a warning and not a fatal error? What is the purpose for allowing a `Dataset` to be constructed which has 0 features? Is it related to distributed training, or maybe to the ability to use `forced_splits`?

### DIFF
--- a/src/io/dataset.cpp
+++ b/src/io/dataset.cpp
@@ -333,8 +333,9 @@ void Dataset::Construct(std::vector<std::unique_ptr<BinMapper>>* bin_mappers,
   }
   if (used_features.empty()) {
     Log::Warning(
-        "There are no meaningful features, as all feature values are "
-        "constant.");
+        "There are no meaningful features which satisfy the provided configuration. "
+        "Decreasing Dataset parameters min_data_in_bin or min_data_in_leaf and re-constructing "
+        "Dataset might resolve this warning.");
   }
   auto features_in_group = NoGroup(used_features);
 
--- a/tests/python_package_test/test_utilities.py
+++ b/tests/python_package_test/test_utilities.py
@@ -43,7 +43,7 @@ def test_register_logger(tmp_path):
     lgb.plot_metric(eval_records)
 
     expected_log = r"""
-INFO | [LightGBM] [Warning] There are no meaningful features, as all feature values are constant.
+INFO | [LightGBM] [Warning] There are no meaningful features which satisfy the provided configuration. Decreasing Dataset parameters min_data_in_bin or min_data_in_leaf and re-constructing Dataset might resolve this warning.
 INFO | [LightGBM] [Info] Number of positive: 2, number of negative: 2
 INFO | [LightGBM] [Info] Total Bins 0
 INFO | [LightGBM] [Info] Number of data points in the train set: 4, number of used features: 0


shiyu1994:
> Is it related to distributed training

I don't think so, since this warning is given after all feature bin mappers are synced, see
https://github.com/microsoft/LightGBM/blob/d4cdbcfe1b1db2d48e95665a8b358028c496cdb2/src/io/dataset_loader.cpp#L791

> or maybe to the ability to use forced_splits

That is the possible reason.

jameslamb:
> But at least, without any feature, we can provide an average score for the whole dataset, which is meaningful in some sense

That makes sense to me! I guess a no-feature model is still technically a valid (even if unlikely to be useful) LightGBM model, and supporting it allows users to not need to account for that possibility in training code. For example, in grid search if you set `min_data_in_leaf` too high and all features have a single bin, you'll still get a valid model (with just very bad performance on the hold-out set) instead of an error.

Thanks for the explanation @shiyu1994 !

guolinke (approved):

Issue #4946, opened 2022-01-13
Check failed: (best_split_info.left_count) > (0)


chixujohnny:
Hi, I found a bug when training with large X_train.

lgb-gpu version: 3.3.2
CUDA=11.1 
CentOS
ram=2TB
GPU=A100-40G

when the X_train is more than (1800w, 1000), lgb-gpu will has a bug like this:
**[LightGBM] [Fatal] Check failed: (best_split_info.left_count) > (0) at LightGBM/src/treelearner/serial_tree_learner.app, line 686**

When I use LGB==3.2.1 , I have the same problem as #4480 : when the GPU memory more than 8.3G will Memory Object Allocation Failure

in this LGB version==3.3.2 , LGB can't load more than 17G GPU memory (GPU has 40G memory), it seems like something problem occur in the tree split step and only happened when GPU memory loaded more than 17G.

Another colleague has this same problem, his lgb version is 3.3.1

jameslamb:
Thanks very much for using LightGBM!

I suspect that this is the same as other issues that have been reported (e.g. #4739, #3679), but it's hard to say without more details.

Are you able to provide any (and hopefully all) of the following?

* specific commands you used to install LightGBM
* the most minimal possible complete version of the code you're running which can reproduce this issue

That would be very helpful. Without such information and with only an error message, significant investigation will probably be required to figure out why you encountered this error.

chixujohnny:
> 

Thanks for your reply!

I build the source code in this command:
mkdir build ; cd build ; cmake  -DUSE_GPU=1 .. && make -j

Sorry I can't paste my code because the code is on my comany machine. I can't screenshots or copy the code, but I can descript my code:
X = np.random.rand(18000000, 1000).astype('float32')
y = np.random.rand(18000000).astype('float32')
model = LGBMRegressor(**params)
model.fit(X, y)

chixujohnny:
One more thing to add.
when I only use 【num_leaves】，the defult is : Memory Object Allocation Failure

jameslamb:
That's ok, completely understand that the code might be sensitive.

Are you able to share the values for `params`? Configuration you're using might give us some clues to help narrow this down.

chixujohnny:
params = {
‘n_estimators’ : 500,
'learning_rate' : 0.05,
'subsample_freq' : 6 ,
'subsample' : 0.91,
'colsample_bytree' : 0.83,
'colsample_bynode' : 0.78,
'num_leaves': 64,
[...]

PR #5087, opened 2022-03-21
Fix program stop when split data count equals zero

GinkoBalboa:
Recently we have discovered a couple of bugs preventing us to train models. The first bug is 
connected with the abrupt stopping of the training because a node split is produced in which the data count is 
zero. The bug makes the LightGBM exit so that the model is lost. The problem arises since we lose
our training progress, and we cannot use the LightGBM in these cases. We have also discovered that
other people had similar problems, for example:

- https://github.com/microsoft/LightGBM/issues/4946

The first thing was to produce the data that causes crashes. We've done this by generating random 
sequences and capturing the sequence that causes the desired error. Because of the, somewhat
different numerics during the training processes when dealing with GPU or CPU we gathered
two sets of data, one that breaks on the CPU, and one that breaks on the GPU:

- `tests/data/data_fail_leaf_count_zero_cpu.csv`
- `tests/data/data_fail_leaf_count_zero_gpu.csv`
- `tests/data/data_fail_num_machines_gt_one.csv`

When these sequences are ran through the test code given in the additional test cases for this
problem:

- `tests/python_package_test/test_engine.py:test_training_leaf_count_zero`
- `tests/python_package_test/test_engine.py:test_training_num_machines_gt_one`

the training stops with one of the two following errors:

```console
    Check failed: (best_split_info.left_count) > (0) at /home/user/LightGBM/src/treelearner/serial_tree_learner.cpp, line 687
```

```console
    Check failed: (best_split_info.right_count) > (0) at /home/user/LightGBM/src/treelearner/serial_tree_learner.cpp, line 697
```

## The solution

The most simple solution for us is not to stop the training at this point and just return a 
warning. At least, this allows us to save the trained model and examine it later. When we 
implemented the solution there was another break that manifested by the `num_machine` greater
than one check failed, while running the second test example (`test_training_num_machines_gt_one`).

```console
    Check failed: (num_machines) > (1) at /home/user/LightGBM/src/treelearner/serial_tree_learner.cpp, line 740
```

So in this place, we implemented the same simple logic - just log a warning and continue the 
training. Maybe the patch proposed here is not the most elegant one, but at least it gives us the 
possibility not to lose the trained model.


jameslamb (commented):
@guolinke or @shiyu1994 I'm not sure about the changes in `serial_tree_learner.cpp`. I think you're better qualified to review that proposal.

If you agree with the proposed changes, I'll come back and give a more specific review about changes to the tests.

guolinke:
Hi @shiyu1994 , is  `best_split_info.left_count == 0` still happening in the latest commits?
Although it is more like a bug, this PR provides a workaround.

shiyu1994:
> is best_split_info.left_count == 0 still happening in the latest commits?

Yes, since we are still getting this bug reports from the users, see e.g., https://github.com/microsoft/LightGBM/issues/3679#issuecomment-938652811.

I believe we need a fix instead of a workaround. I'll debug with the above reproducible example today.

jameslamb:
> I believe we need a fix instead of a workaround. I'll debug with the above reproducible example today.

@shiyu1994 this comment means we should close this pull request, in favor of a fix, right?

guolinke:
it depends how soon we can fix it. If it is quick, we can have a fix, otherwise we can merge this workaround first.

StrikerRUS:
I believe we should have a real fix for this error in `v4.0.0`, not a removed constrained.

shiyu1994:
@jameslamb 

> this comment means we should close this pull request, in favor of a fix, right?

Definitely we want to fix this before 4.0.0. But maybe we can keep this open before the fix? Or even directly modify this PR for a fix since it provides a test case which should be included together with the fix.

shiyu1994:
> But maybe we can keep this open before the fix?

This PR can be closed. The root cause of the example provided in #3679 @mshivers (which is the test case added in this PR) is due to cost efficient gradient boosting. Please refer to #5164.

@GinkoBalboa Thank you for opening this PR and bring the issue to us again!

PR #5097, opened 2022-03-26
Fixing treeshap calculation when data is weighted (fixes #5095)

### LINKED ISSUE #5095: SHAP attributions erronously ignore training data weights
## Description

When predicting SHAP contributions (`model.predict(X, pred_contrib=True)`), the resulting calculation doesn't consider the weight that the individual rows had in the training dataset, and is instead based on the number of records that ended up in the given leaf. This causes incorrect attributions (as best demonstrated by the expected value column).

## Reproducible example
Let's train a model on a toy example:

```python
import numpy as np
import lightgbm

X=np.array([[1,2,3,4,5,6]*100, [1, 1.2, 0.3, 0.8, 0.4, 5]*100], dtype=np.float).T
y=X.dot([0.5, 8])
w=np.array([1,1,1,1,1,10]*100, dtype=np.float)

data = lightgbm.Dataset(X)
data.set_label(y)
data.set_weight(w)

model = lightgbm.train(
    {
        'objective':'regression',
        'metric':'mse',
        'boosting':'gbdt',
        'num_leaves': 7,
        'learning_rate':0.001
    },
    data,
    num_boost_round=100,
)
```

If we ask the model for the contributions of individual features for each prediction, we get a result where the last column is supposed to be the (possibly weighted) expected value of the model predictions on the training set:
```python
model.predict(X, pred_contrib=True)
```

Result:
```python
array([[-0.57021568,  0.1084576 , 29.44625229],
       [-0.55553781,  0.29371625, 29.44625229],
       [-0.64439847, -0.25531573, 29.44625229],
       ...,
       [-0.57688026,  0.10560135, 29.44625229],
       [-0.52634076, -0.20199933, 29.44625229],
       [ 2.69676242,  0.12615042, 29.44625229]])
```

However, it is instead always the *unweighted* expected value of the predictions on the training set. All relevant statistics:

```python
print(np.average(y)) # 13.35
print(np.average(y, weights=w)) # 31.14
print(np.average(model.predict(X))) # 29.4462522901535
print(np.average(model.predict(X), weights=w)) # 31.139999991989136
```

In fact, the result should be the same if we train with weights, or we just repeat the training instances according to their weights. In the latter case, the result is as expected:

```python
X_expand = np.repeat(X, w.astype(int), axis=0)
y_expand = np.repeat(y, w.astype(int), axis=0)
data_expand = lightgbm.Dataset(X_expand)
data_expand.set_label(y_expand)

model2 = lightgbm.train(
    {
        'objective':'regression',
        'metric':'mse',
        'boosting':'gbdt',
        'num_leaves': 7,
        'learning_rate':0.001
    },
    data_expand,
    num_boost_round=100,
)

model2.predict(X_expand, pred_contrib=True)
```

Results in:
```python
array([[-2.23944737,  0.08394158, 31.13999999],
       [-2.17359526,  0.218026  , 31.13999999],
       [-2.40050732, -0.19295458, 31.13999999],
       ...,
       [ 1.07870497,  0.05046017, 31.13999999],
       [ 1.07870497,  0.05046017, 31.13999999],
       [ 1.07870497,  0.05046017, 31.13999999]])
```

Also statistics:

```python
print(np.average(model2.predict(X_expand))) # 31.139999992370605
```

Again, the error is easiest to see when comparing expected values, but (since the attributions sum up to the prediction) must affect other attributions as well:

```python
print(model.predict(X_expand, pred_contrib=True)[0]) # array([-2.23944737,  0.08394158, 31.13999999])
print(model2.predict(X_expand, pred_contrib=True)[0]) # array([-0.57021568,  0.1084576 , 29.44625229])
```

## Environment info

3.3.2

I'm testing under Linux in a jupyter notebook, but this should be irrelevant.

## Additional Comments

I've created a fix for this in my fork (https://github.com/proto-n/LightGBM) and I'm in the process of preparing a pull request.

proto-n:
Fixing Issue #5095

The bug is described in the issue. Continuing with the same example, after this commit, we have:

```python
import numpy as np
import lightgbm

X=np.array([[1,2,3,4,5,6]*100, [1, 1.2, 0.3, 0.8, 0.4, 5]*100], dtype=np.float).T
y=X.dot([0.5, 8])
w=np.array([1,1,1,1,1,10]*100, dtype=np.float)

data = lightgbm.Dataset(X)
data.set_label(y)
data.set_weight(w)

model = lightgbm.train(
    {
        'objective':'regression',
        'metric':'mse',
        'boosting':'gbdt',
        'num_leaves': 7,
        'learning_rate':0.001
    },
    data,
    num_boost_round=100,
)

model.predict(X, pred_contrib=True)
```
resulting in:
```python
array([[-2.23944737,  0.08394158, 31.13999999],
       [-2.17359526,  0.218026  , 31.13999999],
       [-2.40050732, -0.19295458, 31.13999999],
       ...,
       [-2.24611195,  0.08108533, 31.13999999],
       [-2.28244961, -0.13963819, 31.13999999],
       [ 1.07870497,  0.05046017, 31.13999999]])
```

Which is the correct behavior, now the same as the expanded version (see the related issue for expanded version):
```
print(np.average(model.predict(X), weights=w)) # 31.139999991989136
```

Also, the contributions now equal those of the expanded version:
```
# see related issue for X_expand and model2
model.predict(X, pred_contrib=True)[0] # array([-2.23944737,  0.08394158, 31.13999999])
model2.predict(X_expand, pred_contrib=True)[0] # array([-2.23944737,  0.08394158, 31.13999999])
```

-----

The commit modifies the codebase in two places. First, it modifies the SHAP calculation to reference a newly created `data_weight` function. Second, I needed to also fix the weight calculation for internal nodes *and* the expected value calculation as well, as the weights were calculated incorrectly for some reason, and the expected value didn't consider weights either.

Funny enough, beyond SHAP calculations, I couldn't find any other reference to either expected value or the internal weights storage (other than serialization and cuda->cpu / cpu->cuda transfer). I may have missed something though.

Also, it's not clear to me if anything else (like tests of some kind) is required to merge the PR. Please advise.

### DIFF
--- a/include/LightGBM/tree.h
+++ b/include/LightGBM/tree.h
@@ -180,7 +180,7 @@ class Tree {
   inline int data_count(int node) const { return node >= 0 ? internal_count_[node] : leaf_count_[~node]; }
 
   /*! \brief Get the summed weights of data points that fall at or below this node*/
-  inline int data_weight(int node) const { return node >= 0 ? internal_weight_[node] : leaf_weight_[~node]; }
+  inline double data_weight(int node) const { return node >= 0 ? internal_weight_[node] : leaf_weight_[~node]; }
 
   /*!
   * \brief Shrinkage for the tree's output
--- a/include/LightGBM/tree.h
+++ b/include/LightGBM/tree.h
@@ -179,6 +179,9 @@ class Tree {
   /*! \brief Get the number of data points that fall at or below this node*/
   inline int data_count(int node) const { return node >= 0 ? internal_count_[node] : leaf_count_[~node]; }
 
+  /*! \brief Get the summed weights of data points that fall at or below this node*/
+  inline int data_weight(int node) const { return node >= 0 ? internal_weight_[node] : leaf_weight_[~node]; }
+
   /*!
   * \brief Shrinkage for the tree's output
   *        shrinkage rate (a.k.a learning rate) is used to tune the training process
@@ -563,7 +566,7 @@ inline void Tree::Split(int leaf, int feature, int real_feature,
   leaf_parent_[leaf] = new_node_idx;
   leaf_parent_[num_leaves_] = new_node_idx;
   // save current leaf value to internal node before change
-  internal_weight_[new_node_idx] = leaf_weight_[leaf];
+  internal_weight_[new_node_idx] = left_weight + right_weight;
   internal_value_[new_node_idx] = leaf_value_[leaf];
   internal_count_[new_node_idx] = left_cnt + right_cnt;
   leaf_value_[leaf] = std::isnan(left_value) ? 0.0f : left_value;
--- a/src/io/tree.cpp
+++ b/src/io/tree.cpp
@@ -952,9 +952,9 @@ void Tree::TreeSHAP(const double *feature_values, double *phi,
   } else {
     const int hot_index = Decision(feature_values[split_feature_[node]], node);
     const int cold_index = (hot_index == left_child_[node] ? right_child_[node] : left_child_[node]);
-    const double w = data_count(node);
-    const double hot_zero_fraction = data_count(hot_index) / w;
-    const double cold_zero_fraction = data_count(cold_index) / w;
+    const double w = data_weight(node);
+    const double hot_zero_fraction = data_weight(hot_index) / w;
+    const double cold_zero_fraction = data_weight(cold_index) / w;
     double incoming_zero_fraction = 1;
     double incoming_one_fraction = 1;
 
@@ -1004,9 +1004,9 @@ void Tree::TreeSHAPByMap(const std::unordered_map<int, double>& feature_values, 
   } else {
     const int hot_index = Decision(feature_values.count(split_feature_[node]) > 0 ? feature_values.at(split_feature_[node]) : 0.0f, node);
     const int cold_index = (hot_index == left_child_[node] ? right_child_[node] : left_child_[node]);
-    const double w = data_count(node);
-    const double hot_zero_fraction = data_count(hot_index) / w;
-    const double cold_zero_fraction = data_count(cold_index) / w;
+    const double w = data_weight(node);
+    const double hot_zero_fraction = data_weight(hot_index) / w;
+    const double cold_zero_fraction = data_weight(cold_index) / w;
     double incoming_zero_fraction = 1;
     double incoming_one_fraction = 1;
 
@@ -1033,10 +1033,12 @@ void Tree::TreeSHAPByMap(const std::unordered_map<int, double>& feature_values, 
 
 double Tree::ExpectedValue() const {
   if (num_leaves_ == 1) return LeafOutput(0);
-  const double total_count = internal_count_[0];
+
+  const double total_weight = internal_weight_[0];
+
   double exp_value = 0.0;
   for (int i = 0; i < num_leaves(); ++i) {
-    exp_value += (leaf_count_[i] / total_count)*LeafOutput(i);
+    exp_value += (leaf_weight_[i] / total_weight)*LeafOutput(i);
   }
   return exp_value;
 }


proto-n:
Apparently, tests are failing: in the test cases, the contributions don't seem to sum up to the predictions. In my toy example they do:
```python
print(np.linalg.norm(model.predict(X, pred_contrib=True).sum(axis=1)-model.predict(X))) # 6.018678656295655e-13
```

I will investigate why.

StrikerRUS:
Kindly ping @slundberg for the help.

proto-n:
I've done some digging. The tests are unweighted, and the error seems to come from the fact that in the case of classification, the weights don't seem to be identical to the number of nodes.

Apparently the [left_weight and right_weight arguments](https://github.com/proto-n/LightGBM/blob/e9b3aeb1ec8d1846572bf3931b4aed318546ad14/include/LightGBM/tree.h#L544) don't simply always mean the summed weight of the relevant data subsets. [They come](https://github.com/proto-n/LightGBM/blob/e9b3aeb1ec8d1846572bf3931b4aed318546ad14/src/treelearner/]serial_tree_learner.cpp#L629) from the left_sum_hessian and right_sum_hessian fields of a [SplitInfo object](https://github.com/proto-n/LightGBM/blob/e9b3aeb1ec8d1846572bf3931b4aed318546ad14/src/treelearner/split_info.hpp#L22).

While this appears to be indeed the summed data weight with a regression task, in the case of a binary classification task (which the failing tests use) it's a different value.

Example tree from (unweighted) regression, note the `leaf_weight` and `leaf_count` lines:
```
Tree=0
num_leaves=6
num_cat=0
split_feature=0 1 1 0 0
split_gain=421979 2288.13 308.167 162 0.500004
threshold=5.5000000000000009 0.6000000000000002 1.1000000000000003 4.5000000000000009 3.5000000000000004
decision_type=2 2 2 2 2
left_child=1 3 4 -1 -3
right_child=-2 2 -4 -5 -6
leaf_value=31.112759993871052 31.151859993298849 31.117359994252521 31.119459994633992 31.114559993108113 31.117259993871052
leaf_weight=100 1000 100 100 100 100
leaf_count=100 1000 100 100 100 100
internal_value=31.14 31.1163 31.118 31.1137 31.1173
internal_weight=1500 500 300 200 200
internal_count=1500 500 300 200 200
is_linear=0
shrinkage=1
``` 

Example tree from (unweighted) classification:
```
Tree=8
num_leaves=2
num_cat=0
split_feature=0
split_gain=1470.49
threshold=5.5000000000000009
decision_type=2
left_child=-1
right_child=-2
leaf_value=-0.0029528957469360378 0.001494046189980353
leaf_weight=111.98318749666214 221.32959961891174
leaf_count=500 1000
internal_value=0
internal_weight=333.313
internal_count=1500
is_linear=0
shrinkage=0.001
```

Note that while close, the ratio of `111.98318749666214` to `221.32959961891174` is not the same as `500` to `1000`. Also note that my code only modifies the calculation of the `internal_weight` values and doesn't touch the `leaf_weight` values.


Unfortunately, the splitting part of the codebase is a bit esoteric to me and I'm afraid I'd need a significant time investment to understand it. I wonder if someone familiar with that part could clarify the role of hessians vs weights?

proto-n:
It seems to me there hasn't been any activity regarding this. I'm wondering if I can do anything to help?

jameslamb:
Sorry for the delay and thank you for your help and patience!

@guolinke or @shiyu1994 are you able to help review this PR?

guolinke (dismissed):
LGTM

StrikerRUS:
@guolinke Sorry, I guess you've overlooked the part of this PR about failing tests: https://github.com/microsoft/LightGBM/pull/5097#issuecomment-1079659424 and https://github.com/microsoft/LightGBM/pull/5097#issuecomment-1080659340.

guolinke:
It seems the failed tests are related to shap calculation. @proto-n can you help to fix the tests?

proto-n:
@guolinke I would love to, but some of the relevant code falls outside of my expertise. I've detailed the reasons for the failing tests in [this comment](https://github.com/microsoft/LightGBM/pull/5097#issuecomment-1080659340). TL;DR: the fix works for regression, but the weight variables seem to mean something else for classification (and I'm not sure what, and whether that's correct behavior).

guolinke:
It is not trivial to store data weights in tree structure, as it will affect the training efficiency.

Although hessian is not weight, but it is very close. So I prefer to make the hessian-based SHAP. So we can believe current output is correct, and fix ground truth values in tests. Besides, documents about the hessian based SHAP should be added as well.

proto-n:
The thing is, this way the contribution values will violate the rule that `sum(contributions)=prediction`, which is one of the (four) fundamental requirements of SHAP. So I don't particularly like the idea, but ultimately it shouldn't be my decision.

guolinke:
I am not familiar with SHAP, so cannot make the decision. Kindly ping @slundberg for help in data weights.

StrikerRUS:
Maybe it's worth creating a cross-reference issue in the [shap repository](https://github.com/slundberg/shap) to make maintainers there aware of the problem?

Seems that there are already a lot of issues about using sample weights with shap, e.g. https://github.com/slundberg/shap/issues/2508, https://github.com/slundberg/shap/issues/1328, https://github.com/slundberg/shap/issues/2395.

jameslamb:
> Sorry for the possibly dumb question, but how do I ping

Not a dumb question! `@`-mentions on GitHub are the main way we alert external contributors. To be clear, @guolinke wasn't instructing you to ping @slundberg ... he was sending such a notification via his comment.

jameslamb:
It has been 3 months since the last activity on this PR. What needs to happen for it to move forward? We need an opinion from someone who is an expert in SHAP?

If that's the case, maybe we could tag in our friends from XGBoost for an opinion, or look at how XGBoost handles this?

proto-n:
> It has been 3 months since the last activity on this PR. What needs to happen for it to move forward? We need an opinion from someone who is an expert in SHAP?
> 
> If that's the case, maybe we could tag in our friends from XGBoost for an opinion, or look at how XGBoost handles this?

Unfortunately I don't think XGBoost can help here, since we know the problem and the possible solutions, it's just that both are inconvenient.

The main problem is that LightGBM doesn't store the exact per-leaf (and per-node) data-weights in the model in the case of classification. Instead, it stores the hessians which are close but not the same. Storing the actual weights is not trivial according to @guolinke. However, these are required for correct SHAP calculation.

Somebody (I'm not sure who) has to make a decision: either SHAP can be calculated based on hessians (which results in breaking one of the fundamental properties of SHAP), or the actual weights have to be stored somehow (which is a nontrivial modification).

PR #5103, opened 2022-03-30
[CUDA] Fix row-wise histogram construction with dense data matrix

shiyu1994:
Currently row-wise histogram construction cannot run successfully with dense data matrix. The check `row_ptr_bit_type` should be skipped here for dense data matrix,
https://github.com/microsoft/LightGBM/blob/60e72d5f4ec5ec2a4d30a53633e1e2d59c81309b/src/treelearner/cuda/cuda_histogram_constructor.cu#L280-L288

In addition, we disable the use of multi-value feature group in `cuda_exp` for now in this PR. Because we don't have an implementation of sparse bin currently, thus using multi-value feature group can be too costly for GPU memory. We force the feature groups to be single-value.

### DIFF
--- a/src/io/dataset.cpp
+++ b/src/io/dataset.cpp
@@ -342,7 +342,7 @@ void Dataset::Construct(std::vector<std::unique_ptr<BinMapper>>* bin_mappers,
   auto is_sparse = io_config.is_enable_sparse;
   if (io_config.device_type == std::string("cuda") || io_config.device_type == std::string("cuda_exp")) {
       LGBM_config_::current_device = lgbm_device_cuda;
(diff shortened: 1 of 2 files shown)


guolinke (approved):

jameslamb:
Thanks for this fix! Two things:

1. Why was this change so urgent that it needed to be merged with admin powers, even when one CI job was failing? ([build link](https://github.com/microsoft/LightGBM/runs/5747933267?check_suite_focus=true))
2. In the future, can you please try to ensure that "fix" type PRs are accompanied by tests that would prevent the issue from being re-introduced in the future?

PR #5105, opened 2022-03-30
[python-package] Use scikit-learn interpretation of negative `n_jobs` and change default to number of cores

david-cortes:
The scikit-learn interface for lightgbm has a parameter `n_jobs` with a default value of `-1`, which is interpreted as using the default number of OMP threads.

According to the scikit-learn glossary, a negative `n_jobs` has a different meaning (uses joblib's formula), so `-1` means to use all available threads:
https://scikit-learn.org/stable/glossary.html#term-n_jobs

This PR changes the meaning of negative `n_jobs` to match that of scikit-learn, which I think is how most people would expect scikit-learn compatible libraries to behave.

I'm not sure what kind of requirements are there for the tests that are run on Python. This change *could* be tested, but it would imply fitting models using multiple threads per function call - not sure if there's any issue with that like in the R interface.

### DIFF
--- a/python-package/lightgbm/compat.py
+++ b/python-package/lightgbm/compat.py
@@ -179,7 +179,7 @@ except ImportError:
 try:
     from joblib import cpu_count
 
-    def _LGBMCpuCount(only_physical_cores: bool=True):
+    def _LGBMCpuCount(only_physical_cores: bool = True):
         return cpu_count(only_physical_cores=only_physical_cores)
 except ImportError:
     try:
--- a/python-package/lightgbm/compat.py
+++ b/python-package/lightgbm/compat.py
@@ -174,3 +174,21 @@ except ImportError:
 
         def __init__(self, *args, **kwargs):
             pass
+
+"""cpu_count()"""
+try:
+    from joblib import cpu_count
+
+    def _LGBMCpuCount(only_physical_cores: bool=True):
+        return cpu_count(only_physical_cores=only_physical_cores)
+except ImportError:
+    try:
+        from psutil import cpu_count
+
+        def _LGBMCpuCount():
+            return cpu_count(logical=only_physical_cores)
+    except ImportError:
+        from multiprocessing import cpu_count
+
+        def _LGBMCpuCount():
+            return cpu_count()
--- a/python-package/lightgbm/sklearn.py
+++ b/python-package/lightgbm/sklearn.py
@@ -443,8 +443,8 @@ class LGBMModel(_LGBMModelBase):
             For better performance, it is recommended to set this to the number of physical cores
             in the CPU.
 
-            Negative integers are interpreted as following joblib's formula, just like scikit-learn
-            (so e.g. -1 means using all threads). A value of zero corresponds the default number of
+            Negative integers are interpreted as following joblib's formula (n_cpus + 1 + n_jobs), just like
+            scikit-learn (so e.g. -1 means using all threads). A value of zero corresponds the default number of
             threads configured for OpenMP in the system. A value of ``None`` (the default) corresponds
             to using the number of physical cores in the system (its correct detection requires
             either the ``joblib`` or the ``psutil`` util libraries to be installed).
--- a/python-package/lightgbm/sklearn.py
+++ b/python-package/lightgbm/sklearn.py
@@ -654,7 +654,7 @@ class LGBMModel(_LGBMModelBase):
         return params
 
     def _process_n_jobs(self, n_jobs: int):
-        """Convert special values of n_jobs to their actual values according to the formulas that apply
+        """Convert special values of n_jobs to their actual values according to the formulas that apply.
 
         Parameters
         ----------
--- a/python-package/lightgbm/sklearn.py
+++ b/python-package/lightgbm/sklearn.py
@@ -2,7 +2,6 @@
 """Scikit-learn wrapper interface for LightGBM."""
 import copy
 from inspect import signature
-from joblib import cpu_count
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
 import numpy as np
@@ -14,6 +13,17 @@ from .compat import (SKLEARN_INSTALLED, LGBMNotFittedError, _LGBMAssertAllFinite
                      _LGBMComputeSampleWeight, _LGBMLabelEncoder, _LGBMModelBase, _LGBMRegressorBase, dt_DataTable,
                      pd_DataFrame)
 from .engine import train
+
+try:
+    from joblib import cpu_count
+    cpu_count_lib = "joblib"
+except ImportError:
+    try:
+        from psutil import cpu_count
+        cpu_count_lib = "psutil"
+    except ImportError:
+        from multiprocessing import cpu_count
+        cpu_count_lib = "multiprocessing"
 
 _EvalResultType = Tuple[str, float, bool]
 
@@ -363,7 +373,7 @@ class LGBMModel(_LGBMModelBase):
         reg_alpha: float = 0.,
         reg_lambda: float = 0.,
         random_state: Optional[Union[int, np.random.RandomState]] = None,
-        n_jobs: int = 0,
+        n_jobs: Union[int, None] = None,
         importance_type: str = 'split',
         **kwargs
     ):
@@ -426,14 +436,18 @@ class LGBMModel(_LGBMModelBase):
             If int, this number is used to seed the C++ code.
             If RandomState object (numpy), a random integer is picked based on its state to seed the C++ code.
             If None, default seeds in C++ code are used.
-        n_jobs : int, optional (default=0)
-            Number of parallel threads to use for training (can be changed at prediction time).
-            Negative integers are interpreted as following joblib's formula, just like scikit-learn
-            (so e.g. -1 means using all threads). A value of zero corresponds the default number of
-            threads configured for OpenMP in the system.
+        n_jobs : int or None, optional (default=None)
+            Number of parallel threads to use for training (can be changed at prediction time by
+            passing it as an extra keyword argument).
 
             For better performance, it is recommended to set this to the number of physical cores
             in the CPU.
+
+            Negative integers are interpreted as following joblib's formula, just like scikit-learn
+            (so e.g. -1 means using all threads). A value of zero corresponds the default number of
+            threads configured for OpenMP in the system. A value of ``None`` (the default) corresponds
+            to using the number of physical cores in the system (its correct detection requires
+            either the ``joblib`` or the ``psutil`` util libraries to be installed).
         importance_type : str, optional (default='split')
             The type of feature importance to be filled into ``feature_importances_``.
             If 'split', result contains numbers of times the feature is used in a model.
@@ -643,9 +657,18 @@ class LGBMModel(_LGBMModelBase):
         params = _choose_param_value("metric", params, original_metric)
 
         # use joblib conventions for negative n_jobs, just like scikit-learn
-        num_threads_aliases = _ConfigAliases.get("num_threads")
-        if self.n_jobs < 0 and not any([k in num_threads_aliases for k in params.keys()]):
-            params["num_threads"] = max(cpu_count() + 1 + self.n_jobs, 1)
+        if self.n_jobs is None or self.n_jobs < 0:
+            num_threads_aliases = _ConfigAliases.get("num_threads")
+            if not any([k in num_threads_aliases for k in params.keys()]):
+                if self.n_jobs is None:
+                    if cpu_count_lib == "joblib":
+                        params["num_threads"] = cpu_count(only_physical_cores=True)
+                    elif cpu_count_lib == "psutil":
+                        params["num_threads"] = cpu_count(logical=False)
+                    else:
+                        params["num_threads"] = cpu_count()
+                elif self.n_jobs < 0:
+                    params["num_threads"] = max(cpu_count() + 1 + self.n_jobs, 1)
 
         return params
 
--- a/python-package/lightgbm/sklearn.py
+++ b/python-package/lightgbm/sklearn.py
@@ -852,8 +852,11 @@ class LGBMModel(_LGBMModelBase):
 
         # number of threads can have values with special meaning which is only applied
         # in the scikit-learn interface, these should not reach the c++ side as-is
-        predict_params = _choose_param_value("num_threads", predict_params, self.n_jobs)
-        predict_params["num_threads"] = self._process_n_jobs(predict_params["num_threads"])
+        num_threads_aliases = _ConfigAliases.get("num_threads")
+        for alias in num_threads_aliases:
+            if alias in predict_params:
+                n_jobs = predict_params.pop(alias)
+        predict_params["num_threads"] = self._process_n_jobs(n_jobs)
 
         return self._Booster.predict(X, raw_score=raw_score, start_iteration=start_iteration, num_iteration=num_iteration,
                                      pred_leaf=pred_leaf, pred_contrib=pred_contrib, **predict_params)
--- a/python-package/lightgbm/sklearn.py
+++ b/python-package/lightgbm/sklearn.py
@@ -373,7 +373,7 @@ class LGBMModel(_LGBMModelBase):
         reg_alpha: float = 0.,
         reg_lambda: float = 0.,
         random_state: Optional[Union[int, np.random.RandomState]] = None,
-        n_jobs: Union[int, None] = None,
+        n_jobs: Optional[int] = None,
         importance_type: str = 'split',
         **kwargs
     ):
--- a/python-package/lightgbm/sklearn.py
+++ b/python-package/lightgbm/sklearn.py
@@ -1,6 +1,7 @@
 # coding: utf-8
 """Scikit-learn wrapper interface for LightGBM."""
 import copy
+import multiprocessing
 from inspect import signature
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
@@ -362,7 +363,7 @@ class LGBMModel(_LGBMModelBase):
         reg_alpha: float = 0.,
         reg_lambda: float = 0.,
         random_state: Optional[Union[int, np.random.RandomState]] = None,
-        n_jobs: int = -1,
+        n_jobs: int = 0,
         importance_type: str = 'split',
         **kwargs
     ):
@@ -425,8 +426,14 @@ class LGBMModel(_LGBMModelBase):
             If int, this number is used to seed the C++ code.
             If RandomState object (numpy), a random integer is picked based on its state to seed the C++ code.
             If None, default seeds in C++ code are used.
-        n_jobs : int, optional (default=-1)
+        n_jobs : int, optional (default=0)
             Number of parallel threads to use for training (can be changed at prediction time).
+            Negative integers are interpreted as following joblib's formula, just like scikit-learn
+            (so e.g. -1 means using all threads). A value of zero corresponds the default number of
+            threads configured for OpenMP in the system.
+
+            For better performance, it is recommended to set this to the number of physical cores
+            in the CPU.
         importance_type : str, optional (default='split')
             The type of feature importance to be filled into ``feature_importances_``.
             If 'split', result contains numbers of times the feature is used in a model.
@@ -634,6 +641,11 @@ class LGBMModel(_LGBMModelBase):
 
         # overwrite default metric by explicitly set metric
         params = _choose_param_value("metric", params, original_metric)
+
+        # use joblib conventions for negative n_jobs, just like scikit-learn
+        num_threads_aliases = _ConfigAliases.get("num_threads")
+        if self.n_jobs < 0 and not any([k in num_threads_aliases for k in params.keys()]):
+            params["num_threads"] = multiprocessing.cpu_count() + 1 - self.n_jobs
 
         return params
 
--- a/python-package/lightgbm/sklearn.py
+++ b/python-package/lightgbm/sklearn.py
@@ -8,10 +8,23 @@ import numpy as np
 
 from .basic import Booster, Dataset, LightGBMError, _choose_param_value, _ConfigAliases, _log_warning
 from .callback import record_evaluation
-from .compat import (SKLEARN_INSTALLED, LGBMNotFittedError, _LGBMAssertAllFinite, _LGBMCheckArray,
-                     _LGBMCheckClassificationTargets, _LGBMCheckSampleWeight, _LGBMCheckXY, _LGBMClassifierBase,
-                     _LGBMComputeSampleWeight, _LGBMCpuCount, _LGBMLabelEncoder, _LGBMModelBase,
-                     _LGBMRegressorBase, dt_DataTable, pd_DataFrame)
+from .compat import (
+    SKLEARN_INSTALLED,
+    LGBMNotFittedError,
+    _LGBMAssertAllFinite,
+    _LGBMCheckArray,
+    _LGBMCheckClassificationTargets,
+    _LGBMCheckSampleWeight,
+    _LGBMCheckXY,
+    _LGBMClassifierBase,
+    _LGBMComputeSampleWeight,
+    _LGBMCpuCount,
+    _LGBMLabelEncoder,
+    _LGBMModelBase,
+    _LGBMRegressorBase,
+    dt_DataTable,
+    pd_DataFrame,
+)
 from .engine import train
 
 _EvalResultType = Tuple[str, float, bool]
--- a/python-package/lightgbm/sklearn.py
+++ b/python-package/lightgbm/sklearn.py
@@ -851,6 +851,7 @@ class LGBMModel(_LGBMModelBase):
 
         # number of threads can have values with special meaning which is only applied
         # in the scikit-learn interface, these should not reach the c++ side as-is
+        n_jobs = self.n_jobs
         num_threads_aliases = _ConfigAliases.get("num_threads")
         for alias in num_threads_aliases:
             if alias in predict_params:
--- a/python-package/lightgbm/sklearn.py
+++ b/python-package/lightgbm/sklearn.py
@@ -654,8 +654,7 @@ class LGBMModel(_LGBMModelBase):
         return params
 
     def _process_n_jobs(self, n_jobs: int):
-        """
-        Convert special values of n_jobs to their actual values according to the formulas that apply
+        """Convert special values of n_jobs to their actual values according to the formulas that apply
 
         Parameters
         ----------
--- a/python-package/lightgbm/sklearn.py
+++ b/python-package/lightgbm/sklearn.py
@@ -657,19 +657,23 @@ class LGBMModel(_LGBMModelBase):
         params = _choose_param_value("metric", params, original_metric)
 
         # use joblib conventions for negative n_jobs, just like scikit-learn
-        if self.n_jobs is None or self.n_jobs < 0:
-            num_threads_aliases = _ConfigAliases.get("num_threads")
-            num_threads_aliases = [alias for alias in num_threads_aliases if alias != "n_jobs"]
-            if not any([k in num_threads_aliases for k in params.keys()]):
-                if self.n_jobs is None:
-                    if cpu_count_lib == "joblib":
-                        params["num_threads"] = cpu_count(only_physical_cores=True)
-                    elif cpu_count_lib == "psutil":
-                        params["num_threads"] = cpu_count(logical=False)
-                    else:
-                        params["num_threads"] = cpu_count()
-                elif self.n_jobs < 0:
-                    params["num_threads"] = max(cpu_count() + 1 + self.n_jobs, 1)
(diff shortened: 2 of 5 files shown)


python-package/lightgbm/sklearn.py
```diff
@@ -635,6 +642,11 @@ def _process_params(self, stage: str) -> Dict[str, Any]:
         # overwrite default metric by explicitly set metric
         params = _choose_param_value("metric", params, original_metric)
 
+        # use joblib conventions for negative n_jobs, just like scikit-learn
+        num_threads_aliases = _ConfigAliases.get("num_threads")
+        if self.n_jobs < 0 and not any([k in num_threads_aliases for k in params.keys()]):
+            params["num_threads"] = multiprocessing.cpu_count() + 1 - self.n_jobs
```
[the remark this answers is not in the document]
david-cortes: Corrected.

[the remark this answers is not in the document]
StrikerRUS: Just checked how `joblib` determines number of CPU cores:
https://github.com/joblib/joblib/blob/7742f5882273889f7aaf1d483a8a1c72a97d57e3/joblib/externals/loky/backend/context.py#L110

Ughhh! 😬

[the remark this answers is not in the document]
david-cortes: Corrected again.

david-cortes:
@StrikerRUS What would you think of changing the default to the number of physical cores instead?

StrikerRUS:
> What would you think of changing the default to the number of physical cores instead?

TBH, I'm not sure. I see some inconsistency. scikit-learn in their glossary says about CPUs (e.g. "If set to -1, all CPUs are used") but under the hood joblib detects number of threads (according to default [`only_physical_cores=False`](https://github.com/joblib/joblib/blob/a64b808dec003f21e00d07a11bff371523dc19f1/joblib/parallel.py#L307))
https://github.com/scikit-learn/scikit-learn/blob/582fa30a31ffd1d2afc6325ec3506418e35b88c2/sklearn/ensemble/_forest.py#L444-L454

david-cortes:
I'll BTW point out that the linter is complaining about the order of the imports. According to the docs [here](https://github.com/microsoft/LightGBM/tree/master/python-package), the python code should be PEP8-compliant save for a few exceptions. What this linter is complaining about is outside the scope of PEP8 and not mentioned in the docs. I'm also not sure what would be the right order of imports according to the linter (the logs don't mention which linter is complaining about it).

david-cortes:
> > What would you think of changing the default to the number of physical cores instead?
> 
> TBH, I'm not sure. I see some inconsistency. scikit-learn in their glossary says about CPUs (e.g. "If set to -1, all CPUs are used") but under the hood joblib detects number of threads (according to default [`only_physical_cores=False`](https://github.com/joblib/joblib/blob/a64b808dec003f21e00d07a11bff371523dc19f1/joblib/parallel.py#L307)) https://github.com/scikit-learn/scikit-learn/blob/582fa30a31ffd1d2afc6325ec3506418e35b88c2/sklearn/ensemble/_forest.py#L444-L454

I mean, what would you think about setting the default value to the output from `joblib.cpu_count(only_physical_cores=True)` (while still interpreting negative numbers the same way as scikit-learn).

python-package/lightgbm/sklearn.py
```diff
@@ -2,6 +2,7 @@
 """Scikit-learn wrapper interface for LightGBM."""
 import copy
 from inspect import signature
+from joblib import cpu_count
```
StrikerRUS, line 5: We don't have `joblib` in our Python-package requirements. It's available just because it's a dependency of scikit-learn. I'm afraid it can be changed in the future.
https://github.com/microsoft/LightGBM/blob/417c732cc05978276d533c43070d24dfd6c892b7/python-package/setup.py#L351-L356

Moreover, we try to keep other LightGBM interfaces working even if there is no scikit-learn installed in environment.
https://github.com/microsoft/LightGBM/blob/417c732cc05978276d533c43070d24dfd6c892b7/python-package/lightgbm/sklearn.py#L471-L473

I really like how `joblib` determines number of CPUs in their complex way. I guess we can have something the following in [our compatibility module](https://github.com/microsoft/LightGBM/blob/master/python-package/lightgbm/compat.py)

```python
try:
    from joblib import cpu_count
except ImportError:
    try:
        from psutil import cpu_count
    except ImportError:
        from multiprocessing import cpu_count
```

WDYT?

StrikerRUS:
> I mean, what would you think about setting the default value to the output returned from `joblib.cpu_count(only_physical_cores=True)` (while still interpreting negative numbers the same way as scikit-learn).

According to the scikit-learn docs and our own recommendations, we should set `only_physical_cores=True`. But looks like scikit-learn under the hood uses `only_physical_cores=False`, so we won't match their behavior ideally.

I'm for using `only_physical_cores=True` in our package.

david-cortes:
Updated.

david-cortes:
Well, it seems joblib's CPU count function wasn't very robust after all.

tests/python_package_test/test_sklearn.py
```diff
@@ -1301,3 +1302,26 @@ def test_multiclass_custom_objective():
     np.testing.assert_allclose(builtin_obj_preds, custom_obj_preds, rtol=0.01)
     assert not callable(builtin_obj_model.objective_)
     assert callable(custom_obj_model.objective_)
+
+
+def test_negative_n_jobs():
+    n_threads = joblib.cpu_count()
+    if n_threads <= 1:
+        return None
+    val_minus_two = n_threads - 1
```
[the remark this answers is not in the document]
david-cortes: Nevermind this comment, it's something else that was failing, hope it's fixed this time.

david-cortes:
Looks like the failing tests are because of the dask interface having different default arguments from the scikit-learn interface. Not familiar with dask so I'll ask: should the dask interface also default to the same `n_jobs` and have the same interpretation of negative `n_jobs`?

david-cortes:
From the failing tests, it seems [this one](https://github.com/microsoft/LightGBM/blob/3ed0027bffe820ba6c81f32b82f02ef8302c6aad/tests/python_package_test/test_consistency.py#L134) passes a string for number of threads, which according to the docs, is not within the allowed types.

StrikerRUS:
@david-cortes 
Sorry for the delayed response.
> From the failing tests, it seems [this one](https://github.com/microsoft/LightGBM/blob/3ed0027bffe820ba6c81f32b82f02ef8302c6aad/tests/python_package_test/test_consistency.py#L134) passes a string for number of threads, which according to the docs, is not within the allowed types.

Please replace this line 
https://github.com/microsoft/LightGBM/blob/3d25e373ac09fa0cd5ce75a9decbc195e0bd5220/tests/python_package_test/test_consistency.py#L24
with
```python
self.params[key] = value if key not in {'num_trees', 'num_threads'} else int(value)
```
to fix CI failures.

python-package/lightgbm/sklearn.py
```diff
@@ -14,6 +14,17 @@
...
+try:
+    from joblib import cpu_count
+    cpu_count_lib = "joblib"
+except ImportError:
+    try:
+        from psutil import cpu_count
+        cpu_count_lib = "psutil"
+    except ImportError:
+        from multiprocessing import cpu_count
+        cpu_count_lib = "multiprocessing"
```
StrikerRUS, line 26: Please move all this helper stuff to the [`compat.py`](https://github.com/microsoft/LightGBM/blob/master/python-package/lightgbm/compat.py) file with the following encapsulation:
```python
"""cpu_count()"""
try:
    from joblib import cpu_count

    def _LGBMCpuCount():
        return cpu_count(only_physical_cores=True)
except ImportError:
    try:
        from psutil import cpu_count

        def _LGBMCpuCount():
            return cpu_count(logical=False)
    except ImportError:
        from multiprocessing import cpu_count

        def _LGBMCpuCount():
            return cpu_count()
```

python-package/lightgbm/sklearn.py
```diff
@@ -362,7 +373,7 @@ def __init__(
         reg_alpha: float = 0.,
         reg_lambda: float = 0.,
         random_state: Optional[Union[int, np.random.RandomState]] = None,
-        n_jobs: int = -1,
+        n_jobs: Union[int, None] = None,
```
StrikerRUS, line 376:
```suggestion
        n_jobs: Optional[int] = None,
```

python-package/lightgbm/sklearn.py
```diff
@@ -425,8 +436,18 @@ def __init__(
...
-        n_jobs : int, optional (default=-1)
-            Number of parallel threads to use for training (can be changed at prediction time).
+        n_jobs : int or None, optional (default=None)
+            Number of parallel threads to use for training (can be changed at prediction time by
+            passing it as an extra keyword argument).
+
+            For better performance, it is recommended to set this to the number of physical cores
+            in the CPU.
+
+            Negative integers are interpreted as following joblib's formula, just like scikit-learn
```
StrikerRUS, line 446: > Negative integers are interpreted as following joblib's formula

Formula is lost.

python-package/lightgbm/sklearn.py
```diff
@@ -635,6 +656,25 @@ def _process_params(self, stage: str) -> Dict[str, Any]:
         # overwrite default metric by explicitly set metric
         params = _choose_param_value("metric", params, original_metric)
 
+        # use joblib conventions for negative n_jobs, just like scikit-learn
+        n_jobs = self.n_jobs
+        if stage == "fit":
```
StrikerRUS, line 661: Could you please clarify, why this `if` guard is needed here?
`n_jobs` can be passed for the prediction methods as well.

python-package/lightgbm/sklearn.py
```diff
@@ -635,6 +656,25 @@ def _process_params(self, stage: str) -> Dict[str, Any]:
...
+        if n_jobs is None or n_jobs < 0:
+            if n_jobs is None:
+                if cpu_count_lib == "joblib":
+                    params["num_threads"] = cpu_count(only_physical_cores=True)
+                elif cpu_count_lib == "psutil":
+                    params["num_threads"] = cpu_count(logical=False)
+                else:
+                    params["num_threads"] = cpu_count()
+            elif n_jobs < 0:
+                params["num_threads"] = max(cpu_count() + 1 + self.n_jobs, 1)
```
StrikerRUS, line 676: I think this can be done with less lines of code:
```suggestion
        if stage == "fit":
            params.pop('n_jobs', None)
        params = _choose_param_value("num_threads", params, self.n_jobs)
        n_jobs = params["num_threads"]
        if n_jobs is None:
            params["num_threads"] = _LGBMCpuCount()
        elif n_jobs < 0:
            params["num_threads"] = max(_LGBMCpuCount() + 1 + n_jobs, 1)
```

tests/python_package_test/test_sklearn.py
StrikerRUS, line 1311:
```suggestion
    val_minus_two = n_threads - 2
```

tests/python_package_test/test_sklearn.py
```diff
@@ -1301,3 +1302,26 @@ def test_multiclass_custom_objective():
...
+    X, y = load_breast_cancer(return_X_y=True)
+    gbm = lgb.LGBMClassifier(n_estimators=2, verbose=-1, n_jobs=-2).fit(X, y)
+    gbm.booster_.save_model("model.txt")
+    with open("model.txt", "r") as f:
+        model_txt = f.read()
+    assert bool(re.search(r"\[num_threads: %d\]" % val_minus_two, model_txt))
```
StrikerRUS, line 1317:
```suggestion
    assert bool(re.search(rf"\[num_threads: {val_minus_two}\]", model_txt))
```

tests/python_package_test/test_sklearn.py
```diff
@@ -1301,3 +1302,26 @@ def test_multiclass_custom_objective():
...
+
+
+def test_default_n_jobs():
+    n_cores = joblib.cpu_count(only_physical_cores=True)
+    X, y = load_breast_cancer(return_X_y=True)
+    gbm = lgb.LGBMClassifier(n_estimators=2, verbose=-1, n_jobs=None).fit(X, y)
+    gbm.booster_.save_model("model.txt")
+    with open("model.txt", "r") as f:
+        model_txt = f.read()
+    assert bool(re.search(r"\[num_threads: %d\]" % n_cores, model_txt))
```
StrikerRUS, line 1327:
```suggestion
    assert bool(re.search(rf"\[num_threads: {n_cores}\]", model_txt))
```

StrikerRUS (changes requested):
Thanks a lot for working on this!
Please consider checking some my suggestions and questions below.

tests/python_package_test/test_sklearn.py
david-cortes: Joblib's formula is `n_cpus + 1 + n_jobs`. Thus, passing `-2` means "use all but one thread".

python-package/lightgbm/sklearn.py
david-cortes: `n_jobs` is always passed at `fit`, but not always passed at `predict`. Ommitting it at prediction time means using the default.

tests/python_package_test/test_sklearn.py
StrikerRUS: Please add an inline comment explaining this. Right this line is extremely confusing and can be read as "val_minus_two = val - 1" which looks like a bug.

python-package/lightgbm/sklearn.py
StrikerRUS: Right now at prediction time two aliases (default `n_jobs` set in init of fit stage and manually calculated `num_threads`) of the same parameter are passed to cpp code which is incorrect.

david-cortes: It's passing only one actually. But the issue is, `None` at fit-time means "use the number of cores", while at predict-time it means "use the default number of omp threads".

python-package/lightgbm/sklearn.py
david-cortes: Added.

tests/python_package_test/test_sklearn.py
StrikerRUS: I mean to add a comment for very confusing `val_minus_two = n_threads - 1` line exactly, not for `n_jobs=-2`.

StrikerRUS:
> Right now we have the issue that `n_jobs=None` at construction time means "use all the cores", while `n_jobs=None` at prediction time means "use the default number of OMP threads".

But can't we fix this issue by processing `n_jobs` param at predict time the same way as during the fit stage?

StrikerRUS:
The failing `test_consistency.py::test_binary_linear` test makes me think that a result of training a model with linear trees depends on number of threads (possibly eigen dependency is a root cause).
https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=12751&view=logs&j=c28dceab-947a-5848-c21f-eef3695e5f11&t=fa158246-17e2-53d4-5936-86070edbaacf&l=650
@btrotta Could you please help with investigation?

david-cortes:
> The failing `test_consistency.py::test_binary_linear` test makes me think that a result of training a model with linear trees depends on number of threads (possibly eigen dependency is a root cause). https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=12751&view=logs&j=c28dceab-947a-5848-c21f-eef3695e5f11&t=fa158246-17e2-53d4-5936-86070edbaacf&l=650 @btrotta Could you please help with investigation?

Those differences are far too large to be caused only by randomness in eigen's multithreading scheduling. There is something else going on there.

david-cortes:
Updated to use the same interpretation of `n_jobs==None` at fit and predict time, and using the suggested helper function to simplify the code. However, I'd say the helper function and the way in which parameters are processed is rather inefficient and one could decrease the time it takes to make a single-row prediction by making these steps more efficien.

david-cortes:
Actually, it's still not working. The issue is that the functions that process the paramters take `None` values as if the parameter were not passed.

david-cortes:
Not sure what the linter is complaining about. The lines for the imports that were modified in this PR are taken from what `black` suggests.

StrikerRUS:
> Those differences are far too large to be caused only by randomness in eigen's multithreading scheduling. There is something else going on there.

Yeah, differences are indeed quite significant. But I can't imagine any other sources of inconsistency between standard and sklearn interfaces in this PR. I don't believe this touced line can bring a such effect given that ``
```python
self.params[key] = value if key not in {'num_trees', 'num_threads'} else int(value)
```

StrikerRUS:
> Not sure what the linter is complaining about. The lines for the imports that were modified in this PR are taken from what `black` suggests.

Sorry for the inconvenience with the linter! We use `isort` linter. It suggests (`isort . --diff`) the following import ordering:
```python
from .compat import (SKLEARN_INSTALLED, LGBMNotFittedError, _LGBMAssertAllFinite, _LGBMCheckArray,
                     _LGBMCheckClassificationTargets, _LGBMCheckSampleWeight, _LGBMCheckXY, _LGBMClassifierBase,
                     _LGBMComputeSampleWeight, _LGBMCpuCount, _LGBMLabelEncoder, _LGBMModelBase, _LGBMRegressorBase,
                     dt_DataTable, pd_DataFrame)
```

python-package/lightgbm/compat.py
```diff
@@ -174,3 +174,21 @@ class dask_Series:  # type: ignore
...
+    try:
+        from psutil import cpu_count
+
+        def _LGBMCpuCount():
+            return cpu_count(logical=only_physical_cores)
+    except ImportError:
+        from multiprocessing import cpu_count
+
+        def _LGBMCpuCount():
+            return cpu_count()
```
StrikerRUS, line 194: Please unify the function signature for all fallback cases.
```suggestion
        def _LGBMCpuCount(only_physical_cores: bool = True):
            return cpu_count(logical=not only_physical_cores)
    except ImportError:
        from multiprocessing import cpu_count

        def _LGBMCpuCount(only_physical_cores: bool = True):
            return cpu_count()
```

python-package/lightgbm/sklearn.py
```diff
@@ -8,10 +8,23 @@
...
+    _LGBMCheckXY,
+    _LGBMClassifierBase,
+    _LGBMComputeSampleWeight,
+    _LGBMCpuCount,
+    _LGBMLabelEncoder,
+    _LGBMModelBase,
+    _LGBMRegressorBase,
+    dt_DataTable,
+    pd_DataFrame,
+)
```
StrikerRUS: To make `isort` happy.
```suggestion
from .compat import (SKLEARN_INSTALLED, LGBMNotFittedError, _LGBMAssertAllFinite, _LGBMCheckArray,
                     _LGBMCheckClassificationTargets, _LGBMCheckSampleWeight, _LGBMCheckXY, _LGBMClassifierBase,
                     _LGBMComputeSampleWeight, _LGBMCpuCount, _LGBMLabelEncoder, _LGBMModelBase, _LGBMRegressorBase,
                     dt_DataTable, pd_DataFrame)
```

python-package/lightgbm/sklearn.py
```diff
@@ -635,8 +658,34 @@ def _process_params(self, stage: str) -> Dict[str, Any]:
...
 
+        # use joblib conventions for negative n_jobs, just like scikit-learn
+        # at predict time, this is handled later due to the order of parameter updates
+        if stage == "fit":
+            params = _choose_param_value("num_threads", params, self.n_jobs)
+            params["num_threads"] = self._process_n_jobs(params["num_threads"])
+
         return params
 
+    def _process_n_jobs(self, n_jobs: int):
```
StrikerRUS, line 656:
```suggestion
    def _process_n_jobs(self, n_jobs: Optional[int]) -> int:
```

python-package/lightgbm/sklearn.py
```diff
@@ -635,8 +658,34 @@ def _process_params(self, stage: str) -> Dict[str, Any]:
...
+        """Convert special values of n_jobs to their actual values according to the formulas that apply.
+
+        Parameters
+        ----------
+        n_jobs : int
```
StrikerRUS, line 661:
```suggestion
        n_jobs : int or None
```

python-package/lightgbm/sklearn.py
```diff
@@ -635,8 +658,34 @@ def _process_params(self, stage: str) -> Dict[str, Any]:
...
+            The original value of n_jobs, Potentially having special values such as 'None' or
```
StrikerRUS, line 662:
```suggestion
            The original value of n_jobs, potentially having special values such as 'None' or
```

python-package/lightgbm/sklearn.py
```diff
@@ -812,6 +861,16 @@ def predict(self, X, raw_score=False, start_iteration=0, num_iteration=None,
         ):
             predict_params.pop(alias, None)
         predict_params.update(kwargs)
+
+        # number of threads can have values with special meaning which is only applied
+        # in the scikit-learn interface, these should not reach the c++ side as-is
+        n_jobs = self.n_jobs
+        num_threads_aliases = _ConfigAliases.get("num_threads")
+        for alias in num_threads_aliases:
```
StrikerRUS, line 856: Just to decrease the number of code lines without hurting the code readability.
```suggestion
        for alias in _ConfigAliases.get("num_threads"):
```

StrikerRUS:
> But I can't imagine any other sources of inconsistency between standard and sklearn interfaces in this PR.

Now that Linux jobs at Azure Pipelines fail always but Windows jobs fail irregularly (32b73a289ad85b15a0739146118b88aca20dc076: [all successful](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=12751&view=results); a31fcedcc2260fb20492a8b80fa38bc778dc929c: [Windows regular](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=12765&view=logs&j=e0c3ed1e-0c79-508f-2647-33e12d6ed3d4&t=754e6cbc-265f-5c76-c93e-c776f753cc46); 21cd3b207a67841a8ad7e92b889700cbdea04afa: [Windows sdist](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=12771&view=logs&j=67ae2ff8-4b33-578b-0813-9c0440091667&t=aed7750d-29a7-50d7-d7f0-fc7145577059)) I almost sure that the reason is in randomness caused by different number of threads.

StrikerRUS:
Well, the issue of non-deterministic results of the linear tree model given different number of threads isn't related to this PR anyway.

@david-cortes Let's unblock this PR by specifying LightGBM [default number of threads](https://lightgbm.readthedocs.io/en/latest/Parameters.html#num_threads) (0) here:
https://github.com/microsoft/LightGBM/blob/23da5fc56abceac9ccb8ca2b09f945684e0f769e/tests/python_package_test/test_consistency.py#L87
```python
gbm = lgb.LGBMClassifier(**fd.params, n_jobs=0)
```

david-cortes:
> ```python
> , n_jobs=0
> ```

Updated, but I think you should open an issue so as not to forget about those differences. There might be some big bug underneath.

StrikerRUS:
> Updated, but I think you should open an issue so as not to forget about those differences. There might be some big bug underneath.

Thanks a lot! Sure, will open a separate issue for investigating this.

StrikerRUS:
@jmoralez @jameslamb Will really appreciate your review for this PR.

python-package/lightgbm/sklearn.py
```diff
@@ -812,6 +848,15 @@ def predict(self, X, raw_score=False, start_iteration=0, num_iteration=None,
...
             predict_params.pop(alias, None)
         predict_params.update(kwargs)
+
+        # number of threads can have values with special meaning which is only applied
+        # in the scikit-learn interface, these should not reach the c++ side as-is
+        n_jobs = self.n_jobs
+        for alias in _ConfigAliases.get("num_threads"):
+            if alias in predict_params:
+                n_jobs = predict_params.pop(alias)
+        predict_params["num_threads"] = self._process_n_jobs(n_jobs)
```
StrikerRUS, line 858: This can be simplified with `_choose_param_value()`:
```suggestion
        predict_params = _choose_param_value("num_threads", predict_params, self.n_jobs)
        predict_params["num_threads"] = self._process_n_jobs(predict_params["num_threads"])
```

tests/python_package_test/test_sklearn.py
```diff
@@ -1301,3 +1302,29 @@ def test_multiclass_custom_objective():
...
+    # 'val_minus_two' here is the expected number of threads for n_jobs=-2
+    val_minus_two = n_threads - 1
+    X, y = load_breast_cancer(return_X_y=True)
+    # Note: according to joblib's formula, a value of n_jobs=-2 means
+    # "use all but one thread" (formula: n_cpus + 1 + n_jobs)
+    gbm = lgb.LGBMClassifier(n_estimators=2, verbose=-1, n_jobs=-2).fit(X, y)
+    gbm.booster_.save_model("model.txt")
+    with open("model.txt", "r") as f:
+        model_txt = f.read()
+    assert bool(re.search(rf"\[num_threads: {val_minus_two}\]", model_txt))
```
StrikerRUS, line 1320: Save file into a temp directory, not in the current working one.
```suggestion
def test_negative_n_jobs(tmp_path):
    n_threads = joblib.cpu_count()
    if n_threads <= 1:
        return None
    # 'val_minus_two' here is the expected number of threads for n_jobs=-2
    val_minus_two = n_threads - 1
    X, y = load_breast_cancer(return_X_y=True)
    # Note: according to joblib's formula, a value of n_jobs=-2 means
    # "use all but one thread" (formula: n_cpus + 1 + n_jobs)
    gbm = lgb.LGBMClassifier(n_estimators=2, verbose=-1, n_jobs=-2).fit(X, y)
    gbm.booster_.save_model(tmp_path / "model.txt")
    with open(tmp_path / "model.txt", "r") as f:
        model_txt = f.read()
    assert bool(re.search(rf"\[num_threads: {val_minus_two}\]", model_txt))
```

tests/python_package_test/test_sklearn.py
```diff
@@ -1301,3 +1302,29 @@ def test_multiclass_custom_objective():
...
+
+
+def test_default_n_jobs():
+    n_cores = joblib.cpu_count(only_physical_cores=True)
+    X, y = load_breast_cancer(return_X_y=True)
+    gbm = lgb.LGBMClassifier(n_estimators=2, verbose=-1, n_jobs=None).fit(X, y)
+    gbm.booster_.save_model("model.txt")
+    with open("model.txt", "r") as f:
+        model_txt = f.read()
+    assert bool(re.search(rf"\[num_threads: {n_cores}\]", model_txt))
```
StrikerRUS, line 1330:
```suggestion
def test_default_n_jobs(tmp_path):
    n_cores = joblib.cpu_count(only_physical_cores=True)
    X, y = load_breast_cancer(return_X_y=True)
    gbm = lgb.LGBMClassifier(n_estimators=2, verbose=-1, n_jobs=None).fit(X, y)
    gbm.booster_.save_model(tmp_path / "model.txt")
    with open(tmp_path / "model.txt", "r") as f:
        model_txt = f.read()
    assert bool(re.search(rf"\[num_threads: {n_cores}\]", model_txt))
```

StrikerRUS (approved):
Thank you so much for the hard work on this PR!
LGTM except one minor simplifying the codebase suggestion and using temp folder in tests.

StrikerRUS:
> Sure, will open a separate issue for investigating this.

Created: #5266

python-package/lightgbm/sklearn.py
david-cortes: Problem is, `_choose_param_value` treats `None` as if it were not passed, whereas after this PR, `None` has a special value. For example, one can pass `n_jobs=2` in the constructor, then `n_jobs=None` in the call to `predict`, and if that function were used, it would not have the desired effect.

StrikerRUS: Ah, I see now! Thanks for the explanation!

StrikerRUS: @jameslamb Maybe we can refactor `_choose_param_value()` function so that it will handle `None` value of a param correctly?

StrikerRUS:
Kindly pink @jameslamb and @jmoralez for https://github.com/microsoft/LightGBM/pull/5105#issuecomment-1146696790.

python-package/lightgbm/sklearn.py
jameslamb: Ok yeah, I think I understand. Will put up a PR shortly.

jameslamb: created #5289

jameslamb (approved):
Changes look good to me! I have no additional comments, and it's ok with me if we merge this and then later come back and apply the suggestion from https://github.com/microsoft/LightGBM/pull/5105/files#r889631120 after #5289 is merged.

jmoralez (approved):

Issue #5080, opened 2022-03-17
[python-package] ensure that all callbacks are pickleable


jameslamb:
## Summary

All callback functions should be serializable with `pickle`, `clooudpickle`, and `joblib`.

## Motivation

As described in #5012, in some interfaces to LightGBM, it's necessary to broadcast LightGBM parameters or objects from the `lightgbm` Python library to multiple processes / machines. The primary mechanism for this in Python is to serialize objects with a library like `pickle`, `cloudpickle`, or `joblib`.

#5012 made `lgb.callback.early_stopping` serializable, for the benefit of [lightgbm-ray](https://github.com/ray-project/lightgbm_ray). I expect that would also be necessary to use callbacks in `lightgbm.dask`, and for any other settings where users want to pickle/unpickle LightGBM objects.

## Description

Functions in `lightgbm.callback`:

* [x] `early_stopping()` (#5012)
* [ ] `log_evaluation()` (#5101)
* [ ] `record_evaluation()`
* [ ] `reset_parameter()`

Tests similar to the following should be added for each of these functions, to ensure that they're pickleable and that that remains true after future changes to this project.

https://github.com/microsoft/LightGBM/blob/f77e0adf59bfff6b64e2d04e84a0da3079cbe145/tests/python_package_test/test_callback.py#L9-L22

It's possible that just adding such tests will be enough, and that the remaining functions are already pickleable. But if not, then changes will need to be made to support this.

jameslamb:
Closing this issue and adding it to #2302, per this project's policy for feature requests.

Anyone is welcome to contribute this, and a fix can be done across multiple pull requests. If you're interested in working on this, please comment below and maintainers will re-open the issue.

StrikerRUS:
@jameslamb Thanks a lot for writing this up!

> It's possible that just adding such tests will be enough, and that the remaining functions are already pickleable. But if not, then changes will need to be made to support this.

Let me kindly disagree with this statement. I believe we should refactor current interface converting it into classes anyway for the sake of code readability and to avoid any unintended side effects.

StrikerRUS:
Also, I think it worth mentioning that there shouldn't be any code refactoring that assumes that **all callbacks are pickleable**.

We should allow to use third-party callbacks that are unpickleable: #4719.

jameslamb:
> I believe we should refactor current interface converting it into classes anyway for the sake of code readability and to avoid any unintended side effects.

[...]

PR #5107, opened 2022-03-30
[python] make `record_evaluation` callback pickleable

StrikerRUS:
Contributes to #5080.


jameslamb (approved):
excellent, thanks! And thanks very much for simplifying the "pickle then unpickle" code in tests, now that there are several cases of a similarly-structured test.

jmoralez (approved):
Nice!

Issue #2921, opened 2020-03-17
What happens with missing values during prediction?


AlbertoEAF:
Hello,

Suppose I stick to `zero_as_missing=false`, `use_missing=true`. Can you explain what happens during prediction if there are missing values?

I read a bit of the code but those parameters are only used in training, not scoring.

The only reference I saw in the documentation regarding missing values was:

1. https://lightgbm.readthedocs.io/en/latest/Advanced-Topics.html#missing-value-handle
2. https://www.kaggle.com/c/home-credit-default-risk/discussion/57918
3. http://mlexplained.com/2018/01/05/lightgbm-and-xgboost-explained/

According to those sources, nulls are allocated to the bins that reduce the loss during training.

Is that true? And if so, what if there are no missing values in training?

guolinke:
refer to the https://github.com/microsoft/LightGBM/blob/9654f16528d2d1e3462a298106cb1fc6ed43cbac/include/LightGBM/tree.h#L527-L539

AlbertoEAF:
Hello @guolinke I already read a bit and still have some doubts which I can't get through.

During train:
- Missing values for numericals are assigned to the side of the ongoing split that most reduces the loss. This occurs for numerical features alone right **(1)?** 
- Missing values for categoricals are always alloted to the right side of the split **(2)?**

During prediction when a missing is found for a numerical field:
- If there were missing values during train for that split, we allocate to that side of the split the missing feature, correct **(3)?**
- If there were no missing values in train what do we do? I assume we assign by default to the left side **(4)?** _I'm not sure because I couldn't track all the code through._

During prediction when a missing is found for a categorical field:
- We always assign to the right side of the split right? Is there a particular reason **(5)?**


Thank you!

guolinke:
(1) yes
(2) yes
(3) yes, by default left
(4) It will be converted to zero.  refer to https://github.com/microsoft/LightGBM/blob/a8c1e0a11a5cbfac62fb57d16901fea16de95412/include/LightGBM/tree.h#L260-L264
(5) for categorical features, the split is unorder (both {(1, 3), (2, 4, nan)} ati {(2, 4, nan), (1, 3) } are possible for categorical feature, but not for numerical feature). there are, force the missing to tthe right side is okay.

AlbertoEAF:
Sorry but I got confused reading **(3)** & **(4)**.

> (3) yes, by default left


Just to be sure, you're saying that during scoring, missing values for a numerical split:
[...]

PR #5108, opened 2022-03-30
[R-package] Add sparse feature contribution predictions

david-cortes:
ref #4982

This PR adds a missing C-level function for predictions on sparse inputs to the R interface.

Docs are not updated in order to avoid future merge conflicts with following PRs.


jameslamb (changes requested):
Thanks very much for this!

To fix errors like this in CI:

```text
── 1. Error (test_Predictor.R:85:5): start_iteration works correctly ───────────
Error in `predictor$predict(data = data, start_iteration = start_iteration, 
    num_iteration = num_iteration, rawscore = rawscore, predleaf = predleaf, 
    predcontrib = predcontrib, header = header, reshape = reshape)`: object 'LGBM_BoosterPredictSparseOutput_R' not found
Backtrace:
 1. stats::predict(bst, test$data, predcontrib = TRUE)
      at test_Predictor.R:85:4
 2. lightgbm:::predict.lgb.Booster(bst, test$data, predcontrib = TRUE)
 3. object$predict(...)
 4. predictor$predict(...)

── 2. Error (test_Predictor.R:130:5): Feature contributions from sparse inputs p
Error in `predictor$predict(data = data, start_iteration = start_iteration, 
    num_iteration = num_iteration, rawscore = rawscore, predleaf = predleaf, 
    predcontrib = predcontrib, header = header, reshape = reshape)`: object 'LGBM_BoosterPredictSparseOutput_R' not found
Backtrace:
 1. stats::predict(bst, Xcsc, predcontrib = TRUE)
      at test_Predictor.R:130:4
 2. lightgbm:::predict.lgb.Booster(bst, Xcsc, predcontrib = TRUE)
 3. object$predict(...)
 4. predictor$predict(...)
 ```

I think you need to add `LGBM_BoosterPredictSparseOutput_R()` in this registration table: https://github.com/microsoft/LightGBM/blob/60244e4a410693be6a8785b428174730aedd2334/R-package/src/lightgbm_R.cpp#L937

**Comment by david-cortes** [comment]
Added.

File: R-package/tests/testthat/test_Predictor.R [hunk 1]
```diff
@@ -111,3 +113,28 @@ test_that("start_iteration works correctly", {
...
+    pred_csc <- predict(bst, Xcsc, predcontrib = TRUE)
+    expect_s4_class(pred_csc, "dgCMatrix")
+
+    Xcsr <- as(X, "RsparseMatrix")
+    pred_csr <- predict(bst, Xcsr, predcontrib = TRUE)
+    expect_s4_class(pred_csr, "dgRMatrix")
+
+    Xspv <- as(X[1L, , drop = FALSE], "sparseVector")
+    pred_spv <- predict(bst, Xspv, predcontrib = TRUE)
+    expect_s4_class(pred_spv, "dsparseVector")
```
jameslamb, line 139: Beyond just testing the *type* of the returned objects, can you also please add assertions that the predicted *values* are the same for all of these cases, and that they're the same as those predicted for a regular R matrix?

Those `.Call()` calls involve passing a lot of positional arguments with similar values, so such assertions would give us greater confidence that this is working correctly.

R-package/src/lightgbm_R.cpp
```diff
@@ -851,6 +867,72 @@ SEXP LGBM_BoosterPredictForMat_R(SEXP handle,
   R_API_END();
 }
 
+struct SparseOutputPointers {
+  void* indptr; int32_t* indices; void* data; int indptr_type; int data_type;
```
jameslamb, line 871:
```suggestion
  void* indptr;
  int32_t* indices;
  void* data;
  int indptr_type;
  int data_type;
```

Could you please break these declarations onto individual lines, to make it a bit easier to understand this object's structure when reading the code?

jameslamb, line 871:
```suggestion
  void* indptr;
  int32_t* indices;
  void* data;
  int indptr_type;
  int data_type;
```

Could you please break these declarations onto individual lines, to make it a bit easier to understand this object's structure when reading the code?

jameslamb (changes requested):
Looks great to me! Just left two small sugestions.

Can you please merge in the latest changes from `master`? After that, I'll run this project's comment-triggered valgrind and Solaris tests.

I'd also like a second opinion from a maintainer who's more experience with C/C++. @shiyu1994 and/or @guolinke can you help review this PR?

R-package/tests/testthat/test_Predictor.R
david-cortes: I thought missing data was handled the same way as xgboost, which means predictions for sparse outputs **should** be different from those of dense inputs.

jameslamb: I believe the sparse and dense data structures here are just different representations in memory of the exact same matrices, and that specialized methods for them in LightGBM are just intended to allow that sparse data to stay sparse throughout training + scoring.

And I believe that's not directly related to the handling of missing data (which is described in more detail in the discussion at https://github.com/microsoft/LightGBM/issues/2921#issuecomment-607215320 and at https://lightgbm.readthedocs.io/en/latest/Advanced-Topics.html?highlight=missing#missing-value-handle).

Consider the following example:

```r
library(lightgbm)
library(Matrix)

set.seed(708L)

data("EuStockMarkets")

stockDF <- as.data.frame(EuStockMarkets)
feature_names <- c("SMI", "CAC", "FTSE")
target_name <- "DAX"

# randomly set a portion of each feature to NA or 0
for (col_name in feature_names) {
    stockDF[
        sample(
            x = seq_len(nrow(stockDF))
            , size = as.integer(0.01 * nrow(stockDF))
            , replace = FALSE
        )
        , col_name
    ] <- NA_real_
    stockDF[
        sample(
            x = seq_len(nrow(stockDF))
            , size = as.integer(0.01 * nrow(stockDF))
            , replace = FALSE
        )
        , col_name
    ] <- 0.0
}

X_mat <- data.matrix(stockDF[, feature_names])
y <- stockDF[[target_name]]
X_dgCMatrix <- as(X_mat, "dgCMatrix")

bst_mat <- lightgbm::lightgbm(
    data = X_mat
    , label = y
    , objective = "regression"
    , nrounds = 10L
)

bst_dgCMatrix <- lightgbm::lightgbm(
    data = X_dgCMatrix
    , label = y
    , objective = "regression"
    , nrounds = 10L
)


# predicted values don't depend on input type from training time or the type of newdata
preds_mat_mat <- predict(bst_mat, X_mat)
preds_dgCMatrix_mat <- predict(bst_mat, X_dgCMatrix)
preds_mat_dgCMatrix <- predict(bst_dgCMatrix, X_mat)
preds_dgCMatrix_dgCMatrix <- predict(bst_dgCMatrix, X_dgCMatrix)

stopifnot(
    all(
        all(preds_mat_mat == preds_dgCMatrix_mat)
        , all(preds_dgCMatrix_mat == preds_mat_dgCMatrix)
        , all(preds_mat_dgCMatrix == preds_dgCMatrix_dgCMatrix)
    )
)
```

If you find a case where this is not true and LightGBM is creating different predictions for sparse and, I'd consider that a bug worth addressing.

@shiyu1994 @guolinke @StrikerRUS please correct me if I've misspoken.

R-package/src/lightgbm_R.cpp
jameslamb: I see that this was addressed in https://github.com/microsoft/LightGBM/pull/5108/commits/631c935ecdea77d3399e01d5ca3df876d1bdebcf, thanks.

R-package/tests/testthat/test_Predictor.R
david-cortes: Good to know, would be useful to have that the docs, since xgboost works differently (treats non-present sparse entries as missing instead of as zeros) and one might assume both libraries would work the same way.

david-cortes:
Since PR #4977 is going to be merged, I've now modified this PR to also keep row names when applicable, and added a comparison against dense inputs in the tests.

david-cortes:
Failing CI checks are not related to this PR:
https://github.com/microsoft/LightGBM/runs/5819840473?check_suite_focus=true

jameslamb:
> failing CI checks are not related to this PR

I agree, we see this occasionally with our CI jobs that use Homebrew to set up LaTeX.

```text
LaTeX errors found:
! LaTeX Error: File `inconsolata.sty' not found.

Type X to quit or <RETURN> to proceed,
or enter new name. (Default extension: sty)

! Emergency stop.
<read *> 
         
l.287 ^^M
         
```

Will manually re-run them.

R-package/R/lgb.Predictor.R
```diff
@@ -126,6 +127,110 @@ Predictor <- R6::R6Class(
...
+        ncols_out <- integer(1L)
+        .Call(LGBM_BoosterGetNumClasses_R, private$handle, ncols_out)
+        ncols_out <- (ncols + 1L) * max(ncols_out, 1L)
+        if (!inherits(data, "dsparseVector") && ncols_out > .Machine$integer.max) {
+          stop("Resulting matrix of feature contributions is too large for R to handle.")
+        }
+
+        if (inherits(data, "dsparseVector")) {
+
+          if (length(data) > ncols) {
```
jameslamb, line 142: For every `stop()` that you've included in this PR, can you please add a unit test checking that that specific error is raised, using `testhat::expect_error()`?

Here's an example:

https://github.com/microsoft/LightGBM/blob/3798f8720f4fbc4f1249717fb8f2f037e0075a46/R-package/tests/testthat/test_basic.R#L523-L537

I've found tests like that are especially useful for methods like this one with many `if - else if` branches, to be sure that refactorings don't result in unreachable code paths or accidentally failing to raise an informative error.

R-package/tests/testthat/test_Predictor.R
jameslamb: Oh interesting, I did not know that. https://xgboost.readthedocs.io/en/stable/faq.html#why-do-i-see-different-results-with-sparse-and-dense-data

> “Sparse” elements are treated as if they were “missing” by the tree booster, and as zeros by the linear booster. For tree models, it is important to use consistent data formats during training and scoring.

--------------

> would be useful to have that in the docs

LightGBM's documentation does already describe this behavior directly. Please see https://lightgbm.readthedocs.io/en/latest/Advanced-Topics.html#missing-value-handle

> - LightGBM uses NA (NaN) to represent missing values by default. Change it to use zero by setting `zero_as_missing=true`
> - When `zero_as_missing=false` (default), the unrecorded values in sparse matrices (and LightSVM) are treated as zeros.

jameslamb (changes requested):
This is great, thanks so much! I left a few more suggested changes.

I'm very happy that the Solaris and valgrind tests are passing 🎉 

Since this PR introduces so much new C++ code, I'd like my review to not be the only one that counts towards a merge. @shiyu1994 @guolinke could one of you please help review this as well?

R-package/R/lgb.Predictor.R
```diff
@@ -126,6 +127,110 @@ Predictor <- R6::R6Class(
...
+                       , "dgRMatrix"
+                       , "dgCMatrix"
+                       , paste(class(data)
+                       , collapse = ", ")))
+
+        }
+
+        if (NROW(row.names(data))) {
+          out@Dimnames[[1L]] <- row.names(data)
+        }
```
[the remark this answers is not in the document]
david-cortes: Those are already tested in the tests from the PR for row names.

R-package/R/lgb.Predictor.R
david-cortes: I don't think that's a desirable behavior. The package doesn't support other types of sparse inputs *right now*, but if in the future it were to add support for e.g. COO matrices, one would not want to find out that there are tests out there specifically testing that it fails for COO matrices.

Testing all of these conditions would also imply fitting the model to a CSC matrix with close to 2^30 columns.

shiyu1994:
@jameslamb Sure, will review this soon.

R-package/R/lgb.Predictor.R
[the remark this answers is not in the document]
jameslamb: They are not. The links in my comment point to tests on `"CsparseMatrix"` objects, but if you click them you won't see tests on the types referenced in this PR: `"dsparseMatrix"`, `"dsparseVector"`, `"dgRMatrix"`, `"dgCMatrix"`.

Is there something I've misunderstood about the relationship between these classes?

R-package/R/lgb.Predictor.R
jameslamb: > Testing all of these conditions would also imply fitting the model to a CSC matrix with close to 2^30 columns

You're right, saying "every `stop()`" was too much. It's ok to not test the conditions like `.Machine$integer.max`. 

But please add these types of tests as requested for each of the "shape of input data" `stop()`  calls (`if (ncol(data) > ncols) `) and for the one on supported types like `stop(sprintf("Predictions on sparse inputs are only allowed for...`.

R-package/R/lgb.Predictor.R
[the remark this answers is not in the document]
david-cortes: > They are not. The links in my comment point to tests on `"CsparseMatrix"` objects, but if you click them you won't see tests on the types referenced in this PR: `"dsparseMatrix"`, `"dsparseVector"`, `"dgRMatrix"`, `"dgCMatrix"`.
> 
> Is there something I've misunderstood about the relationship between these classes?

There's a class hierarchy...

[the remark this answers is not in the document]
jameslamb: Please tell us specifically what you mean by "there's a class hierarchy", and why it means that you don't want to add the tests I'm asking you to add.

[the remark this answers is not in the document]
david-cortes: `dgCMatrix` is a subclass of `CsparseMatrix`, which is a subclass of `sparseMatrix`, and so on. Classes like `dsparseMatrix` are abstract.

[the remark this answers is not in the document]
jameslamb: Thanks for that. I'm still struggling to understand how that means that the tests I'm asking for shouldn't be added.

Consider the case added in this PR beginning with `} else if (inherits(data, "dgRMatrix")) {`. It doesn't contain a `return()` statement, so that at the end of the `if - else if` block, the "possibly add row names" code (the line this comment thread is on) will run.

If someone were to add `return(out)` on line 193, that "possibly add row names" code wouldn't be reached. I think it's desirable for a test to fail in that case, to inform us that adding that `return` statement is a breaking change that causes row names to not be set on the predictions.

The tests I linked test a regular dense R matrix and a `CsparseMatrix`. As the example below shows, that means they don't currently cover the cases where the input to `predict()` is a `"dgRMatrix"` or a `"dsparseVector"`.

```r
library(Matrix)

# the first batch of test cases use a regular R dense matrix
X <- matrix(rnorm(100), ncol = 4)
inherits(X, "dgRMatrix")         # FALSE
inherits(X, "dsparseMatrix")  # FALSE
inherits(X, "dsparseVector")  # FALSE
inherits(X, "dgCMatrix")         # FALSE

# the second batch of test cases converts that to a CsparseMatrix
Xcsc <- as(X, "CsparseMatrix")
inherits(Xcsc, "dgRMatrix")         # FALSE
inherits(Xcsc, "dsparseMatrix")  # TRUE
inherits(Xcsc, "dsparseVector")  # FALSE
inherits(Xcsc, "dgCMatrix")         # TRUE
```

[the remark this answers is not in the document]
david-cortes: Ok, if you *really* want to have tests for everything, I've added a test for CSR matrices. A vector representing a single row cannot have row names so I left that out of tests.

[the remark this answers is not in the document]
jameslamb: > if you really want to have tests for everything

Thank you, yes I do. The more of this project's behaviors are reflected in tests, the less likely it is that future changes silently break that behavior. This project is too large for all of these concerns to just be kept in maintainers' heads and enforced through PR comments.

R-package/R/lgb.Predictor.R
david-cortes: In that case it'd make more sense to have general tests about unexpected number of columns in the data for all prediction types. The current code for other prediction inputs and types also throws such errors, and those errors don't have tests.

david-cortes: I also don't think it'd be appropriate to have tests checking that it fails for some class like symmetric unit-diagonal sparse CSR matrices just for this - it'd make more to have those tests for all prediction possibilities.

david-cortes:
Failing CI checks like this are definitely not from this PR: https://github.com/microsoft/LightGBM/runs/6029069414?check_suite_focus=true

R-package/R/lgb.Predictor.R
jameslamb: > The current code for other prediction inputs and types also throws such errors, and those errors don't have tests.

I am trying to establish the practice that *changes* to this project are accompanied by tests on those *changes*, so that at least the amount of untested code in the project isn't increased by new PRs. You're not wrong about there being other code paths in the project that are untested or under-tested, but that fact isn't a reason to avoid adding tests in this or other new PRs.

> to have those tests for all prediction possibilities

I don't understand this second point, sorry. I'm not suggesting that LightGBM's tests need an exhaustive list of possible unsupported data types that might be passed into it.

david-cortes: You are asking to add tests for `predict(<sparse matrix>, predcontrib=TRUE)`. The conditions to test are not specific to `<sparse matrix>`, nor to `predcontrib=TRUE`. If tests like those were to be added, the logical structure would be to add them for different possible call types to `predict(<data>, <types>)`, not just for `predict(<sparse matrix>, predcontrib=TRUE)`.

jameslamb: I'm limiting what I'm asking for to only `predict(<sparse matrix>, predcontrib=TRUE)` because those are the only cases being added in this PR and as I said, I'm trying to avoid any *new* untested code being added to the project.

If you want to create a separate PR adding tests on these types of shape checks for all combinations of `<data>` and `<type>` you're welcome to and it would be helpful, but I'm not asking you to commit that time and effort for the purpose of this PR.

jameslamb:
> Failing CI checks like this are definitely not from this PR

Yep, a recent security patch to `git` broke some of our containerized CI jobs (#5151). Once #5152 is merged, they should be fixed.

R-package/R/lgb.Predictor.R
david-cortes: Added the requested tests, but I'd suggest a big refactor after these PRs are merged.

david-cortes:
The linter raised a curious comment:
```
warning: Function "cbind" is undesirable. As an alternative, cbind is an unsafe way to build up a data frame. merge() or direct column assignment is preferred..
```

Says that "cbind is an unsafe way to build up a data frame", but the code is not building a data.frame, nor is it meant to do so.

It then suggests `merge`, which does not do the same operation as `cbind`.

Then it suggests direct column assignment, which will not work if the LHS is a matrix instead of data.frame.

The linter thus leaves one out of efficient options for concatenating matrices or other non-data-frame objects.

jameslamb (changes requested):
Thanks very much for adding tests. I left one more suggestion for those tests...otherwise, I approve of this from the R side.

Before merging though, I would still like some help from @guolinke or @shiyu1994 to review the changes in `lightgbm_R.cpp` (https://github.com/microsoft/LightGBM/pull/5108#pullrequestreview-929585537).

------------

> The linter raised a curious comment ... The linter thus leaves one out of efficient options for concatenating matrices

Since you haven't mentioned the specific place where you wanted to use `cbind()` and what you did instead, it's not possible for maintainers to judge whether whatever pattern you used instead is less efficient than using `cbind()`.

The warning you've mentioned was added a while ago in this project, to nudge contributors who were not familiar with R into using safer patterns when creating data frames. If you have a *specific* proposal for why that restriction should be removed, with a *specific* example of a place in `{lightgbm}`'s code that would benefit from the use of `cbind()`, I'd be happy to consider removing that restriction.

david-cortes:
> Thanks very much for adding tests. I left one more suggestion for those tests...otherwise, I approve of this from the R side.
> 
> Before merging though, I would still like some help from @guolinke or @shiyu1994 to review the changes in `lightgbm_R.cpp` ([#5108 (review)](https://github.com/microsoft/LightGBM/pull/5108#pullrequestreview-929585537)).
> 
> > The linter raised a curious comment ... The linter thus leaves one out of efficient options for concatenating matrices
> 
> Since you haven't mentioned the specific place where you wanted to use `cbind()` and what you did instead, it's not possible for maintainers to judge whether whatever pattern you used instead is less efficient than using `cbind()`.
> 
> The warning you've mentioned was added a while ago in this project, to nudge contributors who were not familiar with R into using safer patterns when creating data frames. If you have a _specific_ proposal for why that restriction should be removed, with a _specific_ example of a place in `{lightgbm}`'s code that would benefit from the use of `cbind()`, I'd be happy to consider removing that restriction.

Any usage of `cbind` with CSC matrices would be more efficient and take fewer lines of code than creating an empty matrix and assigning columns from the cbinded objects. Don't see any usage for that right now, but something like a test checking that it throws an error with inputs of incorrect shape would benefit from calling `cbind`.

jameslamb:
@shiyu1994 @guolinke can you please provide a review?

guolinke:
C++ part looks good to me

guolinke (approved):

jameslamb (approved):
Thanks for looking @guolinke !

Ok @david-cortes , I think we can move forward with this! Since it's been a few weeks since the last commit on this PR, can you please update it to the latest state of `master`? If CI passes after that, I'll merge this up.

Thanks for all the hard work!

david-cortes:
Updated.

david-cortes:
The linter is interpreting regular expressions as file paths:
```
[[1]]
/home/runner/work/LightGBM/LightGBM/R-package/tests/testthat/test_Predictor.R:167:71: warning: [non_portable_path] Use file.path() to construct portable file paths.
    expect_error(predict(bst, X_wrong, predcontrib = TRUE), regexp = "input data has \\d+ columns")
                                                                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~
```

jameslamb:
> The linter is interpreting regular expressions as file paths:

I noticed something similar upgrading to `{lintr}` 3.0, in #5294 . In that PR, I used a `# nolint` comment. But for this PR, could you please just replace `\\d+` with the literal number of columns expected? Since you'll have to touch that line anyway, might as well also make the test a bit stricter.

PR #5109, opened 2022-03-30
[python] make `reset_parameter` callback pickleable

### LINKED ISSUE #5080: [python-package] ensure that all callbacks are pickleable

StrikerRUS:
Fixed #5080.

### DIFF
--- a/python-package/lightgbm/callback.py
+++ b/python-package/lightgbm/callback.py
@@ -130,7 +130,8 @@ class _RecordEvaluationCallback:
                 self.eval_result[data_name][eval_name].append(result)
             else:
                 data_name, eval_name = item[1].split()
-                res_mean, res_stdv = item[2], item[4]
+                res_mean = item[2]
+                res_stdv = item[4]
                 self.eval_result[data_name][f'{eval_name}-mean'].append(res_mean)
                 self.eval_result[data_name][f'{eval_name}-stdv'].append(res_stdv)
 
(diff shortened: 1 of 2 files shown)


python-package/lightgbm/callback.py
```diff
@@ -130,7 +130,8 @@ def __call__(self, env: CallbackEnv) -> None:
                 self.eval_result[data_name][eval_name].append(result)
             else:
                 data_name, eval_name = item[1].split()
-                res_mean, res_stdv = item[2], item[4]
+                res_mean = item[2]
+                res_stdv = item[4]
```
jameslamb, line 134: thank you so much for doing this 😂

jameslamb (approved):
Thanks for the changes, and very cool example using `boosting_round`!

jmoralez (approved):

PR #5111, opened 2022-03-31
integrate MT-GBM into LightGBM (fixes #5047)

### LINKED ISSUE #5047: A external repository : Multi-Tasks GBDT
## Summary

We propose Multi-task Gradient Boosting Machine (MT-GBM), a GBDT-based method for multi-task learning. The MT-GBM can find the shared tree structures and split branches according to multi-task losses.  It can be widely used in realistic scenes.

We hope to add our repository here: https://github.com/microsoft/LightGBM#external-unofficial-repositories.  
Looking forward for your reply!

Repository: https://github.com/antmachineintelligence/mtgbmcode
Paper:https://arxiv.org/abs/2201.06239

antmachineintelligence:
Context

Issue https://github.com/microsoft/LightGBM/issues/5047

Changes

I add a new class of tree_learner and add many new version functions while trying not to break the old code structure and make the reviewer see the multi-task implements. But I don't think it should be updated like my request. It should be done by merging my code into serial_tree_learner by someone who knows the repo better.

Here are some desc about my code:
1. The multi-value tree is implied by copying the same structure tree and assigning them different values.
2. The Gradient and Hessian in fobj are split into two as the paper show. 
3. I have added some functions like savemodel2 to transform the model trained by multi-tasks into a single value model if you need not predict the sub tasks.


src/boosting/gbdt.cpp
```diff
@@ -15,6 +15,12 @@
 #include <ctime>
 #include <sstream>
 
+#ifdef __APPLE__
+#include <sys/malloc.h>
+#else
+#include <malloc.h>
+#endif
```
jameslamb, line 22: I'm really curious, why was this necessary? Is there something missing from LightGBM's existing checks, that should be added to LightGBM independent of this PR?

https://github.com/microsoft/LightGBM/blob/4ae3d1387de73fbadf6194faff816a3b446e506c/CMakeLists.txt#L287-LL300

https://github.com/microsoft/LightGBM/blob/b0137debe6e9cc92b65ec71b0fe8a56ea213c143/include/LightGBM/utils/common.h#L46-L60

jameslamb (changes requested):
Thanks very much for opening this for discussion.

Please sign the contributor license agreement (https://github.com/microsoft/LightGBM/pull/5111#issuecomment-1084035570).

In addition, please make the following changes to reduce the size of this PR, so maintainers can better understand what you're proposing.

* since your example code is written in Python, please move it into https://github.com/microsoft/LightGBM/tree/master/examples/python-guide
    - if you want to keep these examples in notebook form, please open the notebooks and click "restart kernel and clear output" to remove all the outputs like log messages from those files
    - please remove `raw_data.csv` and use something like the datasets from `sklearn.datasets`  or custom data created with `numpy` / `pandas`, to prevent adding to much to the size of this repo
    - please remove the git submodule pointing to https://github.com/white-bird/kaggle-ieee, as that unmaintained repo being removed in the future might prevent the successful cloning of LightGBM with `git clone --recursive`
* update this branch with the changes from the latest version of LightGBM on the `master` branch

---------------

@shiyu1994 @guolinke (and any other maintainers reading this) I'll wait until you are able to provide a comment on whether or not LightGBM should try to support multi-task learning. My initial reaction is that LightGBM shouldn't pursue integrating this at this time, as it might significantly complicate this project at a time when the maintenance burden of the project's existing scope is already very difficult for the existing group of maintainers to keep up with.

But I'm not sure how popular / powerful multi-task gradient boosting is or has the potential to be.

src/boosting/gbdt.cpp
antmachineintelligence: You are right. That should be deleted.

guolinke:
Thank you @antmachineintelligence  for the contribution, and also thank @jameslamb for the reviews. 
In my opinion, as long as this feature is independent, only has minor changes for most existing files, and won't change the efficiency and performance of the normal training, I am okay with the PR. 
If @antmachineintelligence you can take care of the future maintenance of this feature, it would be better. 
also cc @shiyu1994

shiyu1994:
@antmachineintelligence Thanks for your contribution!

> In my opinion, as long as this feature is independent, only has minor changes for most existing files, and won't change the efficiency and performance of the normal training, I am okay with the PR.

I agree. I need some time to review this PR to see if minor changes are enough to integrate it.

jameslamb:
I'm closing this PR due to lack of activity. It has been 2+ months since the last commit or comment on this PR, and maintainers here have significant other work to be done for the next release of LightGBM (#5153).

@shiyu1994 I will leave it up to you whether you want to take supporting multi-task learning as a feature request, or just say that LightGBM won't support this functionality.

PR #5116, opened 2022-04-01
[python-package][R-package] allow using feature names when retrieving number of bins

jmoralez:
Improves on #5048 by allowing to use the feature names as well as the feature indices when retrieving the constructed number of bins. This follows the proposal described https://github.com/microsoft/LightGBM/pull/5048#pullrequestreview-904566843.


python-package/lightgbm/basic.py
```diff
@@ -2386,20 +2386,22 @@ def num_feature(self):
...
+            Index or name of the feature.
 
         Returns
         -------
         number_of_bins : int
             The number of constructed bins for the feature in the Dataset.
         """
         if self.handle is not None:
+            if isinstance(feature, str):
+                feature = self.feature_name.index(feature)
```
StrikerRUS, line 2404: What if I want to use default feature names assigned on cpp side (`['Column_0', 'Column_1', ...]`)?

Also, feature names might be overwritten at cpp side:
https://github.com/microsoft/LightGBM/blob/248fbfa6e8fbf3c284a713681eed4d988e16131d/src/io/dataset.cpp#L1465-L1479
```python
import numpy as np
import lightgbm as lgb

X = np.random.random((50, 5))
y = np.random.random((50))

feature_names=[f'x{i}' for i in range(X.shape[1])]
lgb_data1 = lgb.Dataset(X, y, feature_name=feature_names).construct()
lgb_data2 = lgb.Dataset(X, y, feature_name=feature_names).construct()
lgb_data1.add_features_from(lgb_data2)
lgb_data1.add_features_from(lgb_data2)

bst = lgb.train({}, lgb_data1, num_boost_round=10)

print(bst.dump_model()['feature_names'])
#> ['x0', 'x1', 'x2', 'x3', 'x4', 'D2_x0', 'D2_x1', 'D2_x2', 'D2_x3', 'D2_x4', 'D3_x0', 'D3_x1', 'D3_x2', 'D3_x3', 'D3_x4']
```

I believe we should use the same code as here to handle the above mentioned cases
https://github.com/microsoft/LightGBM/blob/5631366175f68a28bd174982cca16975f7fb48af/python-package/lightgbm/basic.py#L3784-L3785

jmoralez: The `add_features_from` method updates the feature names https://github.com/microsoft/LightGBM/blob/248fbfa6e8fbf3c284a713681eed4d988e16131d/python-package/lightgbm/basic.py#L2523 so in the example you posted `lgb_data1.feature_num_bin('D3_x4')` works.

When not using feature names in the constructor I don't think people can expect that getting the number of bins by name works, or even know what column names are assigned. I could replace `self.feature_name` with `self.get_feature_name()` when looking for the feature names but I'm not sure if that use case is worth it.

StrikerRUS: I think it's not a big deal to replace `self.feature_name` with `self.get_feature_name()` for the sake of increasing the coverage of function use cases. Are you confused by potential performance degradation or what?

StrikerRUS: > Do you think maybe we can try to keep the feature names in sync with the cpp side?

I think it's great idea!
Seems it will be enough to just set `self.feature_name = self.get_feature_name()` after Dataset construction.

jmoralez: added in 87b57ca6558bd42b1aea8b437fae18dad15f1411

R-package/R/lgb.Dataset.R
```diff
@@ -381,6 +385,12 @@ Dataset <- R6::R6Class(
       if (lgb.is.null.handle(x = private$handle)) {
         stop("Cannot get number of bins in feature before constructing Dataset.")
       }
+      if (is.character(feature)) {
+        feature <- which(colnames(self) == feature)
+        if (length(feature) == 0L) {
+          stop("feature not found")
+        }
```
jameslamb, line 392:
```suggestion
        feature_name <- feature
        feature <- which(colnames(self) == feature)
        if (length(feature) == 0L) {
          stop(sprintf("feature '%s' not found", feature_name))
        }
```

In my experience, including information about the input in error messages like this is really helpful for figuring out why you hit it. For example, if you misspelled a feature name that might be clearer when looking at an error message like `feature 'num_clickss' not found`.

Would you consider making this a bit more informative?

R-package/R/lgb.Dataset.R
```diff
@@ -288,6 +288,10 @@ Dataset <- R6::R6Class(
         self$set_colnames(colnames = private$colnames)
       }
 
+      # If the data didn't have feature names we take the ones defined at cpp side
+      # otherwise we just overwrite them
+      self$get_colnames()
```
jameslamb, line 293: I think understand the discussion in https://github.com/microsoft/LightGBM/pull/5116/files#r843276288 that lead to this change, but I don't understand why adding this call here is necessary for the R package and I don't think it's achieving what you want (avoiding a call to the C++ side for performance reasons in `Dataset$get_feature_num_bin()`).

The use of `colnames(self)` down in this PR's proposed change to `Dataset$get_feature_num_bin()` will still make a call to the C++ side.

`colnames()` called on a `Dataset` calls `dimnames.lgb.Dataset()`, which calls `self$get_colnames()`, which gets feature names from the C++ side for a constructed Dataset.

https://github.com/microsoft/LightGBM/blob/b462d0a40f6978c0a66a47af4522c2ae12532316/R-package/R/lgb.Dataset.R#L983

If you want to avoid that call to the C++ side, I think the code in `Dataset$get_feature_num_bin()` could just reference `private$colnames` directly.

Consider the following example:

```r
library(lightgbm)
data("EuStockMarkets")

feature_names <- c("SMI", "CAC", "FTSE")
target_name <- "DAX"
stockDF <- as.data.frame(EuStockMarkets)
X_mat <- data.matrix(stockDF[, feature_names])
y <- stockDF[[target_name]]

# colnames are stored in private$colnames when initializing Dataset
    data = X_mat
    , label = y
    , colnames = feature_names
)
print(dtrain$.__enclos_env__$private$colnames)

# colnames are in private$colnames after construction
dtrain$construct()
print(dtrain$.__enclos_env__$private$colnames)

# set_colnames() updates the C++ side AND private$colnames
dtrain$set_colnames(c("A", "B", "C"))
print(dtrain$.__enclos_env__$private$colnames)
.Call(
    lightgbm:::LGBM_DatasetGetFeatureNames_R
    , dtrain$.__enclos_env__$private$handle
)
```

R-package/tests/testthat/test_dataset.R
```diff
@@ -547,4 +547,15 @@ test_that("lgb.Dataset$get_feature_num_bin() works", {
...
   expect_identical(actual_num_bins, expected_num_bins)
+  # test using defined feature names
+  bins_by_name <- sapply(colnames(raw_mat), ds$get_feature_num_bin)
+  expect_identical(unname(bins_by_name), expected_num_bins)
+  # test using default feature names
+  no_names_mat <- raw_mat
+  colnames(no_names_mat) <- NULL
+  ds_no_names <- lgb.Dataset(no_names_mat, params = list(min_data_in_bin = min_data_in_bin))
+  ds_no_names$construct()
+  default_names <- lapply(seq(1L, ncol(raw_mat)), function(i) sprintf("Column_%d", i - 1L))
```
jameslamb, line 558:
```suggestion
  default_names <- lapply(
    X = seq(1L, ncol(raw_mat))
    , FUN = function(i){
      sprintf("Column_%d", i - 1L)
    }
  )
```

Could you please reformat this like this?

It's a personal style preference, but I've found that `lapply()`, `sapply()`, etc. calls in R are easier to read if the function definition uses `{}` and if the body of the function isn't on the same line as the function signature.

jameslamb (changes requested):
Sorry it took me a few days to provide a review. I left a few suggestions for your consideration.

StrikerRUS (approved):
LGTM, thanks a lot for adding this feature!

Just one suggestion: how about adding a test with categorical features?

R-package/R/lgb.Dataset.R
jmoralez: Thank you. You're right, the call in the `Dataset$get_feature_num_bin()` should use `private$colnames`. The line here however tries to mimic the one on the python side when after constructing the dataset calls `self.feature_name = self.get_feature_name()` to make sure that the feature names on the python side are the same as the ones on the C++ side. This is for the case when no `colnames` are provided in the constructor but the default names are assigned on the C++ side and we want to retrieve the feature bins by the default names, because in that case `private$colnames` is `NULL` until we call `self$get_colnames()` once.

R-package/R/lgb.Dataset.R
jmoralez: added in fe6c9b4b6cf0f60af366dc402acaad9c182ceb31

R-package/tests/testthat/test_dataset.R
jmoralez: added in fe6c9b4b6cf0f60af366dc402acaad9c182ceb31

StrikerRUS (approved):
Re-aproving. Thanks for adding the test!

R-package/R/lgb.Dataset.R
jameslamb: > to make sure that the feature names on the python side are the same as the ones on the C++ side

Ok, I see. I still think it is not obvious, when looking at this line, that that's what is happening. It just looks like a call that got left behind, whose result isn't assigned.

Since you already know with certainty that at this point in `Dataset$construct()` that `Dataset` has been constructed, I think it would be clearer to directly set `private$colnames` like this:

```r
private$colnames <- .Call(
    LGBM_DatasetGetFeatureNames_R
    , private$handle
)
```

Instead of relying on the fact that `self$get_colnames()` happens to do that while it's also getting the features from the C++ side.

Looking at this again, I also recommend reformatting this comment to the following:

```text
# Ensure that private$colnames matches the feature names on the C++ side. This line is necessary
# in cases like constructing from a file or from a matrix with no column names.
```

So that it's a bit clearer when this can happen.

jameslamb (changes requested):
Thanks for this! Please just see my one new suggestion.

R-package/R/lgb.Dataset.R
jmoralez: Done in 302a6a1

R-package/R/lgb.Dataset.R
jameslamb: thanks very much!

jameslamb (approved):
Looks great to me, thanks for working on this!

PR #5117, opened 2022-04-01
[python-package] remove is_reshape argument in Booster.predict (fixes #5115)

### LINKED ISSUE #5115: [python-package] remove is_reshape in Booster.predict()
## Summary

Argument `is_reshape` should be removed from `Booster.predict()` in the Python package.

In all cases where LightGBM returns more than one predicted value per row in the input data, `Booster.predict()` should just return a 2D matrix with `num_rows` rows.

## Motivation

Would simplify the prediction code, reducing the risk of mistakes made by refactorings.

Would simplify the public API of the Python package, making it easier for other code that wraps `lightgbm` to integrate with it.

Now, when there is an impending v4.0.0 release with many breaking changes, is a good opportunity to make such changes, since users upgrading to 4.0.0 will have to deal with other breaking changes already.

## Description

For some tasks, LightGBM's `predict()` routine produces a single predicted value per row in the input data. In those situations, the Python package returns a 1D array.

In other situations, `predict()` produces more than one number per row in the input data. For example, in multiclass classification, for each row in the input data `predict()` produces one class probability or raw value of the objective function per class in the target.

In such situations, it's often desirable to work with those predictions as a `(num_rows, num_classes)` matrix, instead of as a 1-dimensional row-major array.

That is the Python package's default behavior, but it can be overridden by passing argument `is_reshape` to `Booster.predict()`.

https://github.com/microsoft/LightGBM/blob/33eb03767de025ec71d5ac9db35f0936637c1b27/python-package/lightgbm/basic.py#L3492

When that is set to `True` (the default), `lightgbm` converts the predictions into such a matrix.

https://github.com/microsoft/LightGBM/blob/33eb03767de025ec71d5ac9db35f0936637c1b27/python-package/lightgbm/basic.py#L835-L839

If set to `False`, `lightgbm` will return a row-major 1D vector.

## References

Created based on similar changes in the R package, #4971.

jmoralez:
Fixes #5115.

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -752,8 +752,7 @@ class _InnerPredictor:
         return this
 
     def predict(self, data, start_iteration=0, num_iteration=-1,
-                raw_score=False, pred_leaf=False, pred_contrib=False, data_has_header=False,
-                is_reshape=True):
+                raw_score=False, pred_leaf=False, pred_contrib=False, data_has_header=False):
         """Predict logic.
 
         Parameters
@@ -774,8 +773,6 @@ class _InnerPredictor:
         data_has_header : bool, optional (default=False)
             Whether data has header.
             Used only for txt data.
-        is_reshape : bool, optional (default=True)
-            Whether to reshape to (nrow, ncol).
 
         Returns
         -------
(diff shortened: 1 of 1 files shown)


jmoralez:
`is_reshape` seemed like it could be removed from the `_InnerPredictor` as well but if the idea was only removing it from the `Booster` I'll undo those changes.

StrikerRUS (approved):
LGTM, thanks a lot!

jameslamb (approved):
looks great, thanks!

>  from the _InnerPredictor as well but if the idea was only removing it from the Booster I'll undo those changes.

I totally support removing this in `_InnerPredictor.predict()` too. I only mentioned `Booster`  explicitly in #5115 because I thought the description should be focused on the proposed changes to the package's public API.

PR #5118, opened 2022-04-01
[R-package] Add missed packages into dependencies list

StrikerRUS:
Install `RhpcBLASctl` during docs check.

Match list of dependencies in README with one really used in the corresponding CI jobs.
ASAN/UBSAN: https://github.com/microsoft/LightGBM/blob/248fbfa6e8fbf3c284a713681eed4d988e16131d/.github/workflows/r_package.yml#L191
Valgrind: https://github.com/microsoft/LightGBM/blob/248fbfa6e8fbf3c284a713681eed4d988e16131d/.ci/test_r_package_valgrind.sh#L3

### DIFF
--- a/.github/workflows/static_analysis.yml
+++ b/.github/workflows/static_analysis.yml
@@ -57,7 +57,7 @@ jobs:
       - name: Install packages
         shell: bash
         run: |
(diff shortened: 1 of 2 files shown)


jameslamb (approved):
ah! I forgot these `install.packages()` are also documented in the README. Sorry for not catching that in #4972.

PR #5119, opened 2022-04-02
[python-package] highlight the path a sample takes through a tree in `plot_tree` and `create_tree_digraph` (fixes #4784)

### LINKED ISSUE #4784: Add a function to plot tree with a case
## Summary
When we use LGBM to predict, we may find some bad case. I wanna a function to plot the tree, and the nodes and path of the case will be lighted(may be colored with “orange”); 

## Motivation

It’s useful to find the reason of the badcase;
## Description
The plot_tree function can be used. Just add one param ‘ test_data’, default None, if the param is None, the function plot the tree. If the param is the instance of DataFrame of np.array, the function can plot the tree and light the nodes and path of the badcase;
[](url)

jmoralez:
This adds an argument `x` (or we can maybe call it `sample`) that takes a single sample and highlights the path that sample takes through a tree in the tree plotting functions. The path is highlighted by making the edges of the nodes as well as the edges blue and bold. Here's an example for different tree sizes.

<details>
<summary>Sample script</summary>

```python
import lightgbm as lgb

X, y = make_regression(1_000, n_features=4, n_informative=2, random_state=0)
ds = lgb.Dataset(X, y)

fig, ax = plt.subplots(nrows=2, ncols=2, figsize=(16, 6))
leaves = [7, 15, 31, 63]
for i, (axi, num_leaves) in enumerate(zip(ax.flat, leaves)):
    bst = lgb.train({'num_leaves': num_leaves, 'verbose': -1}, ds, num_boost_round=5)
    lgb.plot_tree(bst, x=X[-1], ax=axi)
```
</details>

image

Closes #4784.


jameslamb (changes requested):
Thanks for this! I think this is a really powerful feature, and I support adding it by adding new arguments to `create_tree_digraph()` and `plot_tree()`.

I agree with @StrikerRUS 's suggestion to use color on the arrows and edges, I think it looks great.

A few other recommendations:

* Instead of `x`, would you consider calling this argument `example_case` or `example_observation`?
    - I think the word `sample`  should be avoided, since in other parts of LightGBM that's used to mean "randomly choose".
* could you please add some tests on this new behavior? See the existing tests in `test_plotting.py` for reference, for example:
    - https://github.com/microsoft/LightGBM/blob/7820746266a9033294365a9129ecdd8a91928a02/tests/python_package_test/test_plotting.py#L159
* what will the behavior of this code be if `x` has more than one row in it? Would you consider adding some validation that raises an exception with an informative error if `x` has more than one row?

And some questions (some of which might be about these plotting functions generally and outside the scope of this PR, I'm not sure)

* will this work with categorical features? If I remember correctly, `decision_type` can include "or" rules like `||`
* will this work with categorical features *stored as `pandas` categorical types*?
    - I think it might not, since the proposed code does a direct comparison to the values in `x`, so if the data is a `pd.Series` it isn't passed through the `_data_from_pandas()` logic
    - https://github.com/microsoft/LightGBM/blob/7820746266a9033294365a9129ecdd8a91928a02/python-package/lightgbm/basic.py#L785
    - https://github.com/microsoft/LightGBM/blob/7820746266a9033294365a9129ecdd8a91928a02/python-package/lightgbm/basic.py#L549-L554

jameslamb (changes requested):
Thanks for adding additional tests and categorical support! Really nice work!

I left one small suggestion. I'd still like to test this locally a little bit more, will do that tomorrow. But overall I'm really excited about this 🤩

python-package/lightgbm/plotting.py
```diff
@@ -594,8 +616,18 @@ def create_tree_digraph(
     if show_info is None:
         show_info = []
 
+    if example_case is not None:
+        if example_case.shape[0] > 1:
```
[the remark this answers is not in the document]
jmoralez: Addressed in 6c0c233

jameslamb (approved):
VERY nice work @jmoralez ! I tested tonight in more depth and found everything worked really well. These plots are awesome 🤩 .

I was looking for an example real-world dataset that had informative categorical features, and found that you can filter the UCI Machine Learning Repository by "contains categorical features": https://archive.ics.uci.edu/ml/datasets.php?format=&task=&att=cat&area=&numAtt=&numIns=&type=&sort=nameUp&view=table.

Found that this one worked well for my investigation: https://archive.ics.uci.edu/ml/machine-learning-databases/solar-flare/.

That dataset has ONLY categorical features, so it's useful for testing categorical-specific stuff. I think I'll turn to it in the future when experimenting with `lightgbm` and other ML libraries.

<details><summary>example code (click me)</summary>

```python
import lightgbm as lgb
import pandas as pd
data_url = "https://archive.ics.uci.edu/ml/machine-learning-databases/solar-flare/flare.data2"
df = pd.read_csv(
    filepath_or_buffer=data_url,
    sep=" ",
    header=1,
    skip_blank_lines=True,
    names=[
        "class_code",
        "largest_spot_size_code",
        "spot_distribution_code",
        "activity",
        "evolution",
        "prev_flare_24h",
        "historically_complex",
        "became_complex_on_this_pass",
        "area",
        "largest_spot_area",
        "c_class_flare_count",
        "m_class_flare_count",
        "x_class_flare_count"
    ]
)

for col in df.columns:
    if pd.api.types.is_object_dtype(df[col]):
        df[col] = df[col].astype("category")

y = (df[["c_class_flare_count"]] > 0).astype("int").values.ravel()
feature_names = [c for c in df.columns if not c.endswith("flare_count")]
X = df[feature_names]

dtrain = lgb.Dataset(
    X,
    y,
    feature_name=feature_names,
    categorical_feature="auto",
    params={
        "min_data_in_bin": 5,
        "min_data_per_group": 1
    }
)
bst = lgb.train(
    params={
        'num_leaves': 7,
        'objective': 'binary',
        'min_data_in_leaf': 1,
        'verbose': 1
    },
    train_set=dtrain,
    num_boost_round=3
)

example_case = X[:1]
pd.set_option('display.width', 1000)
print(example_case)
print("---")
print(X["class_code"].cat.categories)
print("---")
lgb.create_tree_digraph(bst, example_case=example_case, tree_index=1)
```

</details>

Tried with a few records, and can see that the records go the correct way in the split, including the handling of `||` for categoricals!

image

-----

In a future PR, could you please add an example using this new function to https://github.com/microsoft/LightGBM/blob/1cc9f9dcee15586aefdd8271a1c04ad010d3c53a/examples/python-guide/plot_example.py?

jmoralez:
> In a future PR, could you please add an example using this new function

Sure.

Thanks for the great suggestions on your review, as always!

python-package/lightgbm/plotting.py
```diff
@@ -594,8 +616,18 @@ def create_tree_digraph(
...
 
+    if example_case is not None:
+        if example_case.shape[0] != 1:
+            raise ValueError('example_case must have a single row.')
+        if isinstance(example_case, pd_DataFrame):
+            example_case = _data_from_pandas(example_case, None, None, booster.pandas_categorical)[0][0]
+        elif isinstance(example_case, np.ndarray) and example_case.ndim == 2:
+            example_case = example_case[0]
+        else:
+            raise ValueError('example_case must be a numpy 2-D array or a pandas Dataframe')
```
StrikerRUS, line 627:
```suggestion
            raise ValueError('example_case must be a numpy 2-D array or a pandas DataFrame')
```

python-package/lightgbm/plotting.py
```diff
@@ -656,6 +689,9 @@ def plot_tree(
     orientation : str, optional (default='horizontal')
         Orientation of the tree.
         Can be 'horizontal' or 'vertical'.
+    example_case : numpy 2-D array or pandas DataFrame, optional (default=None)
```
StrikerRUS, line 692:
```suggestion
    example_case : numpy 2-D array, pandas DataFrame or None, optional (default=None)
```

python-package/lightgbm/plotting.py
```diff
@@ -563,6 +582,9 @@ def create_tree_digraph(
...
```
StrikerRUS, line 585:
```suggestion
    example_case : numpy 2-D array, pandas DataFrame or None, optional (default=None)
```

python-package/lightgbm/plotting.py
```diff
@@ -433,23 +434,42 @@ def _to_graphviz(
...
+            direction = None
+            if example_case is not None:
+                if root['decision_type'] == '==':
+                    thresholds = {int(x) for x in root['threshold'].split('||')}
+                    if example_case[split_feature] in thresholds:
+                        direction = 'left'
+                    else:
+                        direction = 'right'
+                else:
+                    direction = 'left' if example_case[split_feature] <= root['threshold'] else 'right'
```
StrikerRUS, line 472: Does this code cover all combinations of [`use_missing`](https://lightgbm.readthedocs.io/en/latest/Parameters.html#use_missing), [`zero_as_missing`](https://lightgbm.readthedocs.io/en/latest/Parameters.html#zero_as_missing) params and NaN values in `example_case`?

Refer to #2921.

StrikerRUS (commented):
Looks great!
But I'm afraid that this PR doesn't cover of possible scenarios of params and values in `example_case` (see inline comment below).

python-package/lightgbm/plotting.py
```diff
@@ -1,13 +1,14 @@
 # coding: utf-8
 """Plotting library."""
+import math
 from copy import deepcopy
 from io import BytesIO
-from typing import Any, Dict, List, Optional, Tuple, Union
+from typing import Any, Dict, List, Literal, Optional, Tuple, Union
```
StrikerRUS, line 6: Nice improvement compared to general `: str` typing!
But unfortunately, `Literal` is only available in Python 3.8 and newer while LightGBM should support Python 3.7 as well.
> New in version 3.8.
   https://docs.python.org/3/library/typing.html#typing.Literal

https://github.com/microsoft/LightGBM/blob/560c884dd153ac6a95429df8c749aae9632b8215/python-package/setup.py#L386-L390

I believe we should update `: str` type hints with `Literal` in many places in the codebase when Python 3.7 will reach its EOL ([`2023-06-27`](https://devguide.python.org/#status-of-python-branches)).

jmoralez: We could maybe consider using [typing-extensions](https://pypi.org/project/typing-extensions/)

python-package/lightgbm/plotting.py
jmoralez: I added the decisions for numerical splits in 4c19af5 and for categorical in 9f65d63

python-package/lightgbm/plotting.py
StrikerRUS: I'm personally against introducing new package dependencies just for typehints.

tests/python_package_test/test_plotting.py
```diff
@@ -187,6 +188,129 @@ def test_create_tree_digraph(breast_cancer_split):
     assert 'count' not in graph_body
 
 
+@pytest.mark.parametrize('use_missing', [True, False])
+@pytest.mark.parametrize('zero_as_missing', [True, False])
+def test_numeric_split_direction(use_missing, zero_as_missing):
```
StrikerRUS, line 193: Excellence tests!

StrikerRUS (commented):
Awesome contribution!
Just some minor comments below:

python-package/lightgbm/plotting.py
```diff
@@ -414,13 +415,40 @@ def plot_metric(
...
+    return direction
+
+
+def _determine_direction_for_categorical_split(fval: float, thresholds: str, missing_type: str) -> str:
+    if missing_type == 'None':
+        int_fval = -1 if math.isnan(fval) else int(fval)
+    else:
+        int_fval = 0 if math.isnan(fval) else int(fval)
+    int_thresholds = {int(t) for t in thresholds.split('||')}
+    return 'left' if int_fval in int_thresholds else 'right'
```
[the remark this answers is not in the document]
jmoralez: I updated this in 1e6f95d to do the same as #4468. -1 are replaced by nans in https://github.com/microsoft/LightGBM/blob/df14e6077e07259163c9200d2c570a022e2625cd/python-package/lightgbm/basic.py#L559 so we only get nan or a non-negative integer.

python-package/lightgbm/plotting.py
```diff
@@ -414,13 +415,40 @@ def plot_metric(
...
+    else:
+        if default_left:
+            direction = 'left' if le_threshold or math.isnan(fval) else 'right'
+        else:
+            direction = 'left' if le_threshold and not math.isnan(fval) else 'right'
+    return direction
+
+
+def _determine_direction_for_categorical_split(fval: float, thresholds: str, missing_type: str) -> str:
+    if missing_type == 'None':
```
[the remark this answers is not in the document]
jmoralez: Per https://github.com/microsoft/LightGBM/pull/4468#issuecomment-883745908 the categorical split ignores the missing type so I modified the code as I explain in https://github.com/microsoft/LightGBM/pull/5119#discussion_r907933191

python-package/lightgbm/plotting.py
```diff
@@ -414,13 +415,40 @@ def plot_metric(
     return ax
 
 
+def _determine_direction_for_numeric_split(fval: float, threshold: float, missing_type: str, default_left: bool) -> str:
+    le_threshold = fval <= threshold
+    if missing_type == 'None' or (missing_type == 'Zero' and default_left and ZERO_THRESHOLD < threshold):
```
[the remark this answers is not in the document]
jmoralez: Added in 1e6f95d

python-package/lightgbm/plotting.py
```diff
@@ -594,8 +643,18 @@ def create_tree_digraph(
...
 
+    if example_case is not None:
+        if example_case.shape[0] != 1:
+            raise ValueError('example_case must have a single row.')
+        if isinstance(example_case, pd_DataFrame):
+            example_case = _data_from_pandas(example_case, None, None, booster.pandas_categorical)[0][0]
+        elif isinstance(example_case, np.ndarray) and example_case.ndim == 2:
+            example_case = example_case[0]
+        else:
+            raise ValueError('example_case must be a numpy 2-D array or a pandas DataFrame')
```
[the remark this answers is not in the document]
jmoralez: I moved both checks of the input to the top in 1e6f95d. Please let me know if you prefer this suggestion so I update it.

python-package/lightgbm/plotting.py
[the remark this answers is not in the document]
StrikerRUS: Thanks for doing this!

Do we need to update the following code as well?
https://github.com/microsoft/LightGBM/blob/fb37e5074ad0163afd9d88366b58075899775cfa/src/io/tree.cpp#L549-L553

python-package/lightgbm/plotting.py
```diff
@@ -414,13 +415,38 @@ def plot_metric(
     return ax
 
 
+def _determine_direction_for_numeric_split(fval: float, threshold: float, missing_type: str, default_left: bool) -> str:
+    le_threshold = fval <= threshold
+    if missing_type == _MissingType.NONE or (missing_type == _MissingType.ZERO and default_left and ZERO_THRESHOLD < threshold):
```
StrikerRUS, line 420: You cannot compare string with Enum this way. Such comparison will always return `False.`
```python
missing_type = 'Zero'
missing_type == _MissingType.ZERO  # False
```
`missing_type` should either be converted to Enum first or you should compare in the following way: `missing_type == _MissingType.ZERO.value`.

Also, this incorrect comparison shows that added tests are unreliable unfortunately 😢 They should fail, but all CI is green right now.

jmoralez: I updated the numerical split definition to the one here: https://github.com/microsoft/LightGBM/blob/f94050a4cc94909e10a0064baff11cec795eb250/include/LightGBM/tree.h#L335-L353.
Now the test fails if I comment out the enum conversion. Let me know what you think, it looks different than the previous one.

guolinke:
It seems azure pipeline is broken 😢, but I don't have permission to fix it now.  @shiyu1994 , can you take a look?

python-package/lightgbm/plotting.py
StrikerRUS: Thanks for doing this!

Do we need to sync the code in `if/else` dump as well (in a separate PR)?
https://github.com/microsoft/LightGBM/blob/11110c540eb84f519da299ea9eb76cc23a0ca570/src/io/tree.cpp#L520-L560

python-package/lightgbm/plotting.py
```diff
@@ -414,13 +415,38 @@ def plot_metric(
...
+            or (missing_type == _MissingType.NAN and math.isnan(fval))):
+        direction = 'left' if default_left else 'right'
+    else:
+        direction = 'left' if fval <= threshold else 'right'
+    return direction
+
+
+def _determine_direction_for_categorical_split(fval: float, thresholds: str) -> str:
+    if math.isnan(fval):
+        return 'right'
```
StrikerRUS, line 437: https://github.com/microsoft/LightGBM/blob/f94050a4cc94909e10a0064baff11cec795eb250/include/LightGBM/tree.h#L374-L381
```suggestion
    if math.isnan(fval) or int(fval) < 0:
        return 'right'
```

python-package/lightgbm/plotting.py
jmoralez: Yes, I think that code may need to be updated. I'll take a look.

shiyu1994:
@guolinke I'm working on this and will fix this soon. Sorry for the delay.

StrikerRUS (approved):
Great work! Thank you very much!

python-package/lightgbm/plotting.py
StrikerRUS: Kindly ping @jmoralez for possible following-up PR.

PR #5120, opened 2022-04-03
[ci] use lee-dohm/no-response to close stale issues (fixes #5060)

### LINKED ISSUE #5060: [ci] switch to a supported no-response bot
## Description

This project uses `probot/no-response` to automatically close issues where the original poster has stops interacting with the issue.

I noticed today that that project is no longer supported.

As of July 2021, the README at https://github.com/probot/no-response reads as follows:

> ⚠️ We no longer recommend using this app
> Use the action instead: https://github.com/lee-dohm/no-response

This project should switch to https://github.com/lee-dohm/no-response, to limit the risk of this useful automation failing due to incompatibilities with future GitHub Actions changes.

## Additional Comments

Noticed this while considering a contribution to https://github.com/dask/community/issues/60 and reading the discussion there.

jameslamb:
Fixes #5060.

See #5060 for details...the bot that LightGBM currently uses to close stale issues (https://github.com/probot/no-response) was archived by its maintainer in July 2021. That maintainer recommends switching to https://github.com/lee-dohm/no-response.

This PR proposes that change.

## Notes for Reviews

The proposed change would use GitHub Actions to handle this task.

LightGBM's existing solution for this, `probot/no-response`, is instead a GitHub App, which I think means it had to be enabled at https://github.com/microsoft/LightGBM/settings/installations.

If this PR is approved and merged, @shiyu1994 we'll need you to remove the `no-response`  bot from the repo's settings. I don't have permission to do that.


.github/workflows/no-response.yml
```diff
@@ -0,0 +1,23 @@
+name: No Response Bot
+
+on:
+  issue_comment:
+    types: [created]
+  schedule:
+    # "every 6 hours"
+    - cron: '0 */6 * * *'
```
StrikerRUS, line 8: To decrease CI overload and match our current behavior, please change this to "every 24h".

.github/workflows/no-response.yml
```diff
@@ -0,0 +1,23 @@
...
+
+jobs:
+  noResponse:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: lee-dohm/no-response@v0.5.0
```
StrikerRUS, line 14: Thanks for working on this! However, this action looks abandoned as well: last commit was 21 Oct 2021.

StrikerRUS: And our current no response bot was archived on on 13 Jul 2021:
https://github.com/probot/no-response/commit/74cbc5393669059479e157084e2b3cb2ed8421c2

jameslamb: The one we currently use, `probot/noresponse` is explicitly archived, so I'm 100% confident it won't be receiving future updates and that it should be replaced.

I don't think `lee-dohm/no-response` not having a commit since October 2021 means it is "abandoned". That action is very simple, so I just don't think it requires much active development.

Consider the official "checkout" action, managed directly by GitHub: https://github.com/actions/checkout/commits/main. Even that very popular action, which is used across MANY repos (https://github.com/search?l=YAML&q=%22uses%3A+actions%2Fcheckout%22&type=Code), went 4 months (November 2021 to February 2022) without any commits.

image

.github/workflows/no-response.yml
jameslamb: oh sure! Updated in https://github.com/microsoft/LightGBM/pull/5120/commits/d256695133ec16cb348fc203240910bfe46f1eb7.

I picked an arbitrary hour (04:00 UTC). If you have any strong opinion about running at a particular time of day, I'd be happy to change it.

.github/workflows/no-response.yml
StrikerRUS: Thanks for the investigation!
I just thought that it's possible to find a more actively developing and popular action (the proposed one has only 15 stars). But I haven't managed to find the alternative.

StrikerRUS (approved):
LGTM!

jameslamb:
alright, let's see how it works!

@shiyu1994 can you please remove the no-response bot at https://github.com/microsoft/LightGBM/settings/installations?

shiyu1994:
@jameslamb Seems that I don't have the permission to operate that either.
image
I did not see an uninstall button.
According to
https://github.community/t/how-to-remove-a-github-app/127567
It seems that we should operate from the organization side?
image

But I don't have the permission to change the settings of the organization.

It is weird that we cannot operate on this. Did I miss something?

jameslamb:
Oh interesting! I'm not sure, maybe @guolinke will remember from when this was set up originally in #2843. I don't even have the access to see that list you've shared, so I'm just guessing at how this works 😬

guolinke:
I cannot open the https://github.com/microsoft/LightGBM/settings/installations , @shiyu1994 it seems you didn't join your GitHub account to Microsoft organization, maybe you need to join to change the setting.

PR #5121, opened 2022-04-03
[R-package] ensure boosting happens in tests on small datasets

jameslamb:
The R package contains a few tests using the `mtcars` dataset that comes built into R.

That dataset only has 32 observations in it. From `?mtcars`.

```r
data(mtcars)
dim(mtcars)
# [1] 32 11
```

As a result, the calls to `lightgbm()` and `lgb.train()` in tests using that dataset are not currently performing any boosting, for reasons described in #5081.

If setting `verbosity` to something low enough to allow `INFO` and `WARNING` level logs, those tests contain logs like the following:

```text
[LightGBM] [Warning] There are no meaningful features which satisfy the provided configuration. Decreasing Dataset parameters min_data_in_bin or min_data_in_leaf and re-constructing Dataset might resolve this warning.
[LightGBM] [Info] Total Bins 0
[LightGBM] [Info] Number of data points in the train set: 32, number of used features: 0
[LightGBM] [Info] Start training from score 20.090625
```

This PR proposes setting `min_data_in_bin = 1` and `min_data_in_leaf = 1` in those examples, to ensure that boosting occurs. I believe this will improve the test coverage LightGBM gets from these tests, by ensuring that the tests use models that actually generate trees with splits.

### DIFF
--- a/R-package/tests/testthat/test_Predictor.R
+++ b/R-package/tests/testthat/test_Predictor.R
@@ -116,12 +116,20 @@ test_that("predictions for regression and binary classification are returned as 
     data(mtcars)
     X <- as.matrix(mtcars[, -1L])
     y <- as.numeric(mtcars[, 1L])
-    dtrain <- lgb.Dataset(X, label = y, params = list(max_bins = 5L))
+    dtrain <- lgb.Dataset(
+      X
+      , label = y
+      , params = list(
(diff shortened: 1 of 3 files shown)


StrikerRUS (approved):
LGTM, thanks!

PR #5122, opened 2022-04-03
[R-package] ensure values in params override keyword arguments to predict() (fixes #4670)

### LINKED ISSUE #4670: [R-package] `predict` ignores `predict_raw_score`
After fitting a binary classification model, calling `predict` with `predict_raw_score=TRUE` or its aliases as explained in [the docs](https://lightgbm.readthedocs.io/en/latest/Parameters.html#predict-parameters) will not output the raw score, only a probability as it would without this parameter:
```r
library(lightgbm)
data("agaricus.train")
pool = lgb.Dataset(agaricus.train$data, label=agaricus.train$label)
params = list(
    objective_type="binary",
    force_row_wise=TRUE,
    seed=123
)
model = lgb.train(params, pool, verbose=0)
summary(predict(model, agaricus.train$data, params=list(predict_raw_score=TRUE)))
```
```
     Min.   1st Qu.    Median      Mean   3rd Qu.      Max. 
0.0000201 0.0000212 0.0000232 0.4821127 0.9999774 0.9999774 
```
```r
summary(predict(model, agaricus.train$data))
```
```
     Min.   1st Qu.    Median      Mean   3rd Qu.      Max. 
0.0000201 0.0000212 0.0000232 0.4821127 0.9999774 0.9999774 
```

jameslamb:
Fixes #4670.

Currently, `predict.lgb.Booster()` and `Booster$predict()` in the R package have the following keyword arguments allowing control of the type of prediction LightGBM produces:

* `rawscore`: leaf values in terms of the objective function, without any postprocessing (e.g. no conversion to probabilities for classification)
* `predleaf`: `(num_row, num_tree)` matrix where the `(i, j)` entry describes which leaf node in tree `j` input observation `i` falls into
* `predcontrib`: `(num_row, num_feature + 1)`  matrix with the `(i, j)` entry describes the feature contribution of feature `j` to the predicted value for input observation `i`, and the final column contains the Shapley "base value"

LightGBM's C++ library supports parameters for these same values, but per the example in #4670, those are currently ignored if provided to `predict()` in the R package.

This PR fixes that, and ensures that values in `params` override keyword arguments.


StrikerRUS:
Maybe we can simply rename corresponding keyword arguments to match the names from https://lightgbm.readthedocs.io/en/latest/Parameters.html#predict-parameters?

In addition, I will also support the same breaking change for the Python-package.

jameslamb:
> Maybe we can simply rename corresponding keyword arguments to match the names from https://lightgbm.readthedocs.io/en/latest/Parameters.html#predict-parameters?

hmmm I'm not opposed to that, but I think it would still be important to deal with the possibility of this configuration also showing up in `params`, since prediction parameters are part of the public API for `predict()`.

Since the set of prediction parameters changes so infrequently, and since we're considering breaking changes for the upcoming release anyway, what do you think about this set of changes?

* just expose ALL of the parameters at https://lightgbm.readthedocs.io/en/latest/Parameters.html#predict-parameters as keyword arguments in `predict.lgb.Booster()`
* remove the `params` list from `predict.lgb.Booster()` and `Booster$predict()` entirely

good things about this:

* makes `predict()` interface stricter, eliminating the risk of bugs like #4670
* makes the R interface more closely match the documentation at https://lightgbm.readthedocs.io/en/latest/Parameters.html#predict-parameters
* reduces overhead in `predict()` calls related to checking the `params` list
* is the type of breaking change that will be loud and obvious (resulting in and error like "unused argument `params`")

bad things about this:

* breaks some existing user code
* removes support for aliases of the prediction parameters, such as `is_pred_contrib` and `contrib`, in the R package
    - (I suspect these are very rarely used)
* requires changes to the R package to support newly-added prediction parameters in the C API

StrikerRUS:
Ah, I forgot that we [switched from `...` to `params` in `predict()`](f7e39388d0327e687854f22d7e06aec302770994) for the R-package! This makes my previous suggestion less relevant for the R-package...

StrikerRUS:
TBH, I don't support exposing **ALL** predict-level parameters...

You've already mentioned the main disadvantages in "bad things about this:", but I'm also afraid of the number of unimportant params users will see in the `predict()` function signature. I strongly believe it will be confusing for them and they will waste their time getting familiar with each param description (and those descriptions are definitely not perfect and detailed).

jameslamb:
Yeah makes sense. Then I think we should keep this PR as-is, with the checking of `params`. And I think we should avoid changing the argument names (e.g. changing `predleaf` to `predict_leaf_index)`.

R-package/tests/testthat/test_Predictor.R
```diff
@@ -112,6 +112,167 @@ test_that("start_iteration works correctly", {
...
+      , seed = 708L
+    )
+    , nrounds = 10L
+    , verbose = VERBOSITY
+  )
+
+  # check that predictions really look like leaf index predictions
+  preds_leaf_s3_keyword <- predict(bst, X, predleaf = TRUE)
+  expect_true(is.matrix(preds_leaf_s3_keyword))
+  expect_equal(dim(preds_leaf_s3_keyword), c(32L, 10L))
```
jmoralez, line 193:
```suggestion
  expect_equal(dim(preds_leaf_s3_keyword), c(nrow(X), bst$current_iter()))
```

R-package/tests/testthat/test_Predictor.R
```diff
@@ -112,6 +112,167 @@ test_that("start_iteration works correctly", {
...
+  expect_true(min(preds_leaf_s3_keyword) >= 0L)
+  max_leaf_node_index <- max(
+    model = lgb.model.dt.tree(bst)[["leaf_index"]]
+    , na.rm = TRUE
+  )
+  expect_true(max(preds_leaf_s3_keyword) <= max_leaf_node_index)
```
jmoralez, line 199:
```suggestion
  trees_dt <- lgb.model.dt.tree(bst)
  max_leaf_by_tree_from_dt <- trees_dt[, .(idx = max(leaf_index, na.rm = TRUE)), by = tree_index]$idx
  max_leaf_by_tree_from_preds <- apply(preds_leaf_s3_keyword, 2, max, na.rm = TRUE)
  expect_equal(max_leaf_by_tree_from_dt, max_leaf_by_tree_from_preds)
```

R-package/tests/testthat/test_Predictor.R
jameslamb: can you explain why you think the suggested change is preferable to what I currently have? I think testing against literals is a powerful way to catch subtle bugs, for exampleaccidental in-place modifications of `X`

R-package/tests/testthat/test_Predictor.R
jameslamb: Nice! This is even stricter than what i currently have.

R-package/tests/testthat/test_Predictor.R
jmoralez: I thought it was easier to understand what is being tested, for example if I don't know how many rows `mtcars` has.  It's also similar to the `predcontrib` test: `expect_equal(dim(preds_contrib_s3_keyword), c(nrow(X), num_features + 1L))`

StrikerRUS:
> Then I think we should keep this PR as-is

OK, that's fine with me.

R-package/tests/testthat/test_Predictor.R
jameslamb: Ok sure, I'll make that change. I like sometimes mixing in these tests with literals since I've found they can catch subtle bugs, but I understand what you're saying about this one being hard to understand when reading it.

jmoralez (approved):

R-package/R/lgb.Booster.R
```diff
@@ -486,6 +486,34 @@ Booster <- R6::R6Class(
         start_iteration <- 0L
       }
 
+      # possibly override keyword arguments with parameters
+      #
+      # NOTE: this length() check minimizes the latency introduced by these checks,
+      #       for the common case where params is empty
+      #
+      # NOTE: doing this here instead of in Predictor$predict() to keep
+      #       Predictor$predict() as fast as possible
```
StrikerRUS, line 495: Why was this needed?

R-package/tests/testthat/test_Predictor.R
```diff
@@ -114,6 +114,166 @@ test_that("start_iteration works correctly", {
...
+    )
+    , nrounds = 10L
+    , verbose = VERBOSITY
+  )
+
+  # check that the predictions from predict.lgb.Booster() really look like raw score predictions
+  preds_prob <- predict(bst, X)
+  preds_raw_s3_keyword <- predict(bst, X, rawscore = TRUE)
+  preds_prob_from_raw <- 1.0 / (1.0 + exp(-preds_raw_s3_keyword))
+  expect_equal(preds_prob, preds_prob_from_raw, tolerance = 1e-6)
```
StrikerRUS, line 144: Please a global variable for the default tolerance:
https://github.com/microsoft/LightGBM/blob/b462d0a40f6978c0a66a47af4522c2ae12532316/R-package/tests/testthat/test_basic.R#L17

R-package/tests/testthat/test_Predictor.R
```diff
@@ -114,6 +114,166 @@ test_that("start_iteration works correctly", {
...
+test_that("predict() params should override keyword argument for raw-score predictions", {
+  data(agaricus.train, package = "lightgbm")
+  X <- agaricus.train$data
+  y <- agaricus.train$label
+  bst <- lgb.train(
+    data = lgb.Dataset(
+      data = X
+      , label = y
+      , params = list(
+        verbosity = VERBOSITY
```
StrikerRUS, line 126: I think `VERBOSITY` from `train()` function is passed to unconstructed `Dataset` class automatically, no need to specify this param twice

https://github.com/microsoft/LightGBM/blob/b462d0a40f6978c0a66a47af4522c2ae12532316/R-package/R/lgb.train.R#L80-L84
https://github.com/microsoft/LightGBM/blob/b462d0a40f6978c0a66a47af4522c2ae12532316/R-package/R/lgb.train.R#L152-L154

R-package/R/lgb.Booster.R
jameslamb: Like I mentioned in https://github.com/microsoft/LightGBM/pull/4909#pullrequestreview-842363693 and https://github.com/microsoft/LightGBM/pull/4977#pullrequestreview-894601022, I think we should be overly-protective of `predict()` functions / methods, as those may be used in latency-sensitive applications like serving a model as a REST API.

Each call to `lgb.check.wrapper_param()` results in a call to get all aliases from the C++ side and deserialize them from JSON into an R list: 

https://github.com/microsoft/LightGBM/blob/b462d0a40f6978c0a66a47af4522c2ae12532316/R-package/R/utils.R#L193

https://github.com/microsoft/LightGBM/blob/b462d0a40f6978c0a66a47af4522c2ae12532316/R-package/R/aliases.R#L40-L45

Since I expect that the use of additional `params` in `predict()` is not common, I added this `if length(` statement here to avoid imposing the overhead of those 3 `lgb.check.wrapper_param()` calls on existing user code if it wasn't necessary.

R-package/tests/testthat/test_Predictor.R
jameslamb: ah good point! updated in https://github.com/microsoft/LightGBM/pull/5122/commits/d8efeaab40d5bcfa1b082304ca47f0bab20005f5

R-package/tests/testthat/test_Predictor.R
jameslamb: oh sure, updated in https://github.com/microsoft/LightGBM/pull/5122/commits/d8efeaab40d5bcfa1b082304ca47f0bab20005f5

R-package/R/lgb.Booster.R
StrikerRUS: Hmm, thanks for the detailed as always explanation! Are you aware of any `Predictor` usage examples in a third-party packages?

Is it possible to refactor `.PARAMETER_ALIASES()` in a way it will call cpp function only once for the session. Refer to the Python-package implementation as an example:
https://github.com/microsoft/LightGBM/blob/b462d0a40f6978c0a66a47af4522c2ae12532316/python-package/lightgbm/basic.py#L340

StrikerRUS (approved):
LGTM!

R-package/R/lgb.Booster.R
jameslamb: > Are you aware of any Predictor usage examples in a third-party packages?

I'm not aware of any such usage. It is *technically* possible for user code to get a `Predictor` using the public API of the R package.

You'd have to use `lgb.train()` or `lightgbm()` to get a `Booster`, and then `Booster$to_predictor()` to get a `Predictor`.

https://github.com/microsoft/LightGBM/blob/b462d0a40f6978c0a66a47af4522c2ae12532316/R-package/R/lgb.Booster.R#L509-L511

But I would be surprised if many users do that, so for all intents and purposes `Predictor` is private today.

> Is it possible to refactor `.PARAMETER_ALIASES()` in a way it will call cpp function only once for the session

Yes, it is. Briefly, I can think of at least 3 options:

1. Have an environment in the R package's namespace, and have calls to `.PARAMETER_ALIASES()` check that environment for a list like `parameter_aliases`, and only re-populate it if its missing
    - this could also possibly be loaded at the time the package is loaded, using the special `.onLoad()` hook (see ["load hooks" in Writing R extensions](https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Load-hooks))
2. Use the `{memoise}` package ([link](https://cran.r-project.org/web/packages/memoise/index.html)), which has functionality similar to `functools.lru_cache()` in Python.

This issue is larger than just this PR, since `.PARAMETER_ALIASES()` is used many times in the R package. I'll write up a separate issue describing this and link it here.

jameslamb: > I'll write up a separate issue

Ok actually it was easy enough to just put my preferred proposal directly into a pull request, so I did that instead of creating an issue.

I've opened #5141.

Note that even if that PR was merged, I'd still recommend keeping this `if (length(params) > 0L)` check, to avoid running the other R-side for loops and checks in `lgb.check.wrapper_param()` on calls to `predict()`.

https://github.com/microsoft/LightGBM/blob/b462d0a40f6978c0a66a47af4522c2ae12532316/R-package/R/utils.R#L193-L203

raghavthind2005 (approved):

R-package/R/lgb.Booster.R
StrikerRUS: > Note that even if that PR was merged, I'd still recommend keeping this `if (length(params) > 0L)` check, to avoid running the other R-side for loops and checks in `lgb.check.wrapper_param()` on calls to `predict()`.

I don't have anything against `if (length(params) > 0L)` check! 🙂 
I was just thinking about the reasons to keep it in `Booster` instead of `Predictor`.

jameslamb: Ah ok!

I think it's preferable to keep it in the `Booster` for now, since `Predictor$predict()`  doesn't support taking `params`

https://github.com/microsoft/LightGBM/blob/9a964fef17c5e20587b5d4f3120011065ad1b08b/R-package/R/lgb.Predictor.R#L81-L87

and since it needs to pass boolean values (well, integer representations of them) for raw_score, leaf_index, and pred_contrib to the `LGBM_Predict*`  functions

https://github.com/microsoft/LightGBM/blob/9a964fef17c5e20587b5d4f3120011065ad1b08b/R-package/R/lgb.Predictor.R#L110-L122

Since these are all internal details, I think we can change it in the future without breaking any user code if there's a compelling reason to move this into the `Predictor`.

StrikerRUS: > `Predictor$predict()`  doesn't support taking `params`

Yeah, it takes them in a constructor
https://github.com/microsoft/LightGBM/blob/f860282658d079f517cce7b614a1f08e6ccc6d3e/R-package/R/lgb.Predictor.R#L29
https://github.com/microsoft/LightGBM/blob/9a964fef17c5e20587b5d4f3120011065ad1b08b/python-package/lightgbm/basic.py#L707-L708

jameslamb: Right, it takes them in the constructor and then immediately converts them to a string to be passed to the C++ side and stores that string.

https://github.com/microsoft/LightGBM/blob/0a4851f5fed79fc8815e97c23e92756ab77bef1c/R-package/R/lgb.Predictor.R#L30

Since `Predictor$predict()` accepts things like `rawscore` as keyword arguments, supporting passing something like `predict_raw_score` in `params` would require one of the following changes to `Predictor`:

* extract `rawscore`, `predleaf`, etc. in the constructor and store them as private attributes, before converting to string with `lgb.params2str()`
* accept a `params` list in `Predictor$predict()` instead of in the constructor
* accept a `params` list in `Predictor$predict()` and merge it with whatever `params` were passed into the constructor and stored at `private$params`

I think it'd be better to do this work up in `Booster` than to make any of those changes to `Predictor` to support this.

That would match the behavior of LightGBM's C API anyway...the root cause of #4670 and the reason for this PR in the first place is that functions like `LGBM_BoosterPredictForMat()` on the C++ side ignore parameters like `predict_raw_score` passed through `params`, and instead rely on `predict_type` being resolved outside and passed through a function argument.

https://github.com/microsoft/LightGBM/blob/0a4851f5fed79fc8815e97c23e92756ab77bef1c/R-package/R/lgb.Predictor.R#L159-L172

https://github.com/microsoft/LightGBM/blob/0a4851f5fed79fc8815e97c23e92756ab77bef1c/R-package/src/lightgbm_R.cpp#L839

https://github.com/microsoft/LightGBM/blob/0a4851f5fed79fc8815e97c23e92756ab77bef1c/R-package/src/lightgbm_R.cpp#L846-L848

https://github.com/microsoft/LightGBM/blob/0a4851f5fed79fc8815e97c23e92756ab77bef1c/src/c_api.cpp#L2133

I think parameters like `predict_raw_score` only exist on the C++ side for the CLI.

https://github.com/microsoft/LightGBM/blob/6b56a90cd1324a6dbac2afa0a352c9355b0dc3cf/src/application/application.cpp#L248-L251

jameslamb: @StrikerRUS since you and @jmoralez have approved this PR and since in this thread we're only discussing the internal API of the package (so decisions are reversible without impacting users), I've updated this to latest `master` and am going to merge it once CI passes.

If you come back to this thread and want something to be changed, please let me know.

StrikerRUS: Sure, fine with me!

jameslamb: Thanks! This was my plan last night....then all CI started breaking, with the issues being discussed on #5169 😂

PR #5123, opened 2022-04-03
rename internal callback functions

mayer79:
Following https://github.com/microsoft/LightGBM/pull/5018#pullrequestreview-892006092, 
changing the internal callback names:

- cb.early.stop() --> cb_early_stop()
- cb.print.evaluation() --> cb_print_evaluation()
- cb.record.evaluation() --> cb_record_evaluation()

### DIFF
--- a/R-package/R/callback.R
+++ b/R-package/R/callback.R
@@ -71,7 +71,7 @@ merge.eval.string <- function(env) {
 
 }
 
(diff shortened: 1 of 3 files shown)


jameslamb (approved):
Thanks for doing this!

I double-checked like this:

```shell
git grep -E "cb\."
```

and didn't find any other references.

PR #5068, opened 2022-03-11
[ci] fix current `master` fails with graphviz-related error

StrikerRUS:
To workaround the following error on Windows:
```
__________________________ test_create_tree_digraph ___________________________
cmd = ['dot', '-Kdot', '-Tpdf', '-O', 'Tree4.gv'], input_lines = None
encoding = None, capture_output = True, quiet = False
kwargs = {'cwd': None, 'startupinfo': <subprocess.STARTUPINFO object at 0x000000F539A7A988>, 'stderr': -1, 'stdout': -1}
proc = CompletedProcess(args=['dot', '-Kdot', '-Tpdf', '-O', 'Tree4.gv'], returncode=1, stdout=b'', stderr=b'Warning: Could n... json json0 metafile mp pdf pic plain plain-ext png pov ps ps2 svg tif tiff tk vml xdot xdot1.2 xdot1.4 xdot_json\r\n')
    def run_check(cmd: typing.Sequence[typing.Union[os.PathLike, str]], *,
                  input_lines: typing.Optional[BytesOrStrIterator] = None,
                  encoding: typing.Optional[str] = None,
                  capture_output: bool = False,
                  quiet: bool = False,
                  **kwargs) -> subprocess.CompletedProcess:
        """Run the command described by ``cmd``
            with ``check=True`` and return its completed process.
    
        Raises:
            CalledProcessError: if the returncode of the subprocess is non-zero.
        """
        log.debug('run %r', cmd)
    
        cmd = list(map(_compat.make_subprocess_arg, cmd))
    
        if not kwargs.pop('check', True):  # pragma: no cover
            raise NotImplementedError('check must be True or omited')
    
        if capture_output:  # Python 3.6 compat
            kwargs['stdout'] = kwargs['stderr'] = subprocess.PIPE
    
        if encoding is not None:
            kwargs['encoding'] = encoding
    
        kwargs.setdefault('startupinfo', _compat.get_startupinfo())
    
        try:
            if input_lines is not None:
                assert kwargs.get('input') is None
                assert iter(input_lines) is input_lines
                proc = _run_input_lines(cmd, input_lines, kwargs=kwargs)
            else:
                proc = subprocess.run(cmd, **kwargs)
        except OSError as e:
            if e.errno == errno.ENOENT:
                raise ExecutableNotFound(cmd) from e
            raise
    
        if not quiet and proc.stderr:
            _write_stderr(proc.stderr)
    
        try:
>           proc.check_returncode()
[...]

PR #5126, opened 2022-04-05
[ci] Remove temp workaround for graphviz installation

StrikerRUS:
Reverts #5068.


jameslamb (approved):
excellent! I'm glad we could remove this.

Issue #5128, opened 2022-04-05
[ci] R macOS CI jobs failing on 'brew install basictex'


jameslamb:
## Description

For at least the last 24 hours, macOS CI jobs have been failing with the following error:

```text
==> Caveats
You must restart your terminal window for the installation of MacTex CLI tools to take effect.
Alternatively, Bash and Zsh users can run the command:

  eval "$(/usr/libexec/path_helper)"

==> Downloading http://mirror.ctan.org/systems/mac/mactex/mactex-basictex-20210325.pkg
==> Downloading from https://ctan.math.illinois.edu/systems/mac/mactex/mactex-basictex-20210325.pkg
curl: (22) The requested URL returned error: 404 
Error: Download failed on Cask 'basictex' with message: Download failed: http://mirror.ctan.org/systems/mac/mactex/mactex-basictex-20210325.pkg
sudo: tlmgr: command not found
sudo: tlmgr: command not found
installer: Error - the package path specified was invalid: '/Users/runner/work/LightGBM/LightGBM/R.pkg'.
```

Which comes from here:

https://github.com/microsoft/LightGBM/blob/d811f3a074d30924a88aebe93eea66df70ecb7dc/.ci/test_r_package.sh#L74-L77

This results in the following related error during `R CMD check`:

```text
* checking PDF version of manual ...  WARNING
LaTeX errors when creating PDF version.
This typically indicates Rd problems.
* checking PDF version of manual without hyperrefs or index ... ERROR
Re-running with no redirection of stdout/stderr.
Hmm ... looks like a package
Error in texi2dvi(file = file, pdf = TRUE, clean = clean, quiet = quiet,  : 
  pdflatex is not available
Error in texi2dvi(file = file, pdf = TRUE, clean = clean, quiet = quiet,  : 
  pdflatex is not available
Error in running tools::texi2pdf()
You may want to clean up by 'rm -Rf /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T//RtmpJ2Qw8R/Rd2pdf2a345d471e9'
* checking for non-standard things in the check directory ... NOTE
Found the following files/directories:
  ‘lightgbm-manual.tex’
```

## Reproducible example

Example build from #4811: https://github.com/microsoft/LightGBM/runs/5826112507?check_suite_focus=true.

But this is failing on `master` and multiple PRs.

This is affecting all macOS R CI jobs.

jameslamb:
[...]

Issue #4990, opened 2022-02-01
[ci] [R-package] use macOS 11.x in R 4.x CI jobs


jameslamb:
## Summary

Created from https://github.com/microsoft/LightGBM/pull/4989#issuecomment-1025743555.

LightGBM's CI jobs for the R package should use the newest available macOS version (i.e. `macOS-latest` environment from GitHub Actions) together with jobs testing the newest available R version.

## Motivation

macOS 11 ("Big Sur") was first introduced in June 2020 (about 19 months ago). It's likely that a large portion of the LightGBM users working on Macs are using R 4.1.x along with macOS 11. Testing this combination would prevent changes that break compatibility with such environments from being merged.

## Description

To address this issue, update `os:` in https://github.com/microsoft/LightGBM/blob/master/.github/workflows/r_package.yml to `macOS-latest` for R 4.x macOS jobs.

Note that when I tried this in #4989, I saw the following error ([build link](https://github.com/microsoft/LightGBM/runs/4994838445?check_suite_focus=true)):

> installer: Error - the package path specified was invalid: '/Users/runner/work/LightGBM/LightGBM/R.pkg'.

Possible solutions to that:

1. maybe R 4.2.0 will produce a `.pkg` (at https://cran.r-project.org/bin/macosx/base/) that works with Big Sur
2. installing R from Homebrew (https://formulae.brew.sh/formula/r)
3. maybe the `macOS-latest` environment will become compatible with R 4.1.2 in response to https://github.com/actions/virtual-environments/issues/4960

## References

* https://github.com/actions/virtual-environments/pull/4048
* https://github.com/actions/virtual-environments/issues/4960
* https://github.com/microsoft/LightGBM/issues/4988
* https://github.com/r-lib/actions/issues/412

PR #5129, opened 2022-04-05
[ci] fail R macOS CI jobs earlier when installations fail

jameslamb:
#5128 describes an issue where:

* `brew install` of a required dependency fails
* the CI job does not fail
* building LightGBM, running tests, etc. all happen
* a much later step (`R CMD check` for the R package) fails, because it's missing that dependency from Homebrew

This PR proposes adding some `exit -1` calls to the main CI script for for R-package jobs, so that those jobs will fail earlier when such issues occur in the future.

Benefits of this:

* less work required to find the root cause of issues (since the problem will be near the end of logs)
* jobs will exit sooner, meaning less time occupying a spot in the repo's limited GitHub Actions queue (https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits)


jameslamb:
Going to move this back to "draft" while investigating the latest error

> installer: Error - the package path specified was invalid: '/Users/runner/work/LightGBM/LightGBM/R.pkg'.

But hey, at least in https://github.com/microsoft/LightGBM/runs/5826439843?check_suite_focus=true, `basictex` was installed successfully, so maybe #5128 is fixed!

jameslamb:
😱  it looks like the `R.pkg` we were installing was removed!

```shell
curl -I https://cran.r-project.org/bin/macosx/R-4.1.2.pkg
```

```text
HTTP/1.1 404 Not Found
Date: Tue, 05 Apr 2022 01:51:31 GMT
Server: Apache
Vary: accept-language,accept-charset
Accept-Ranges: bytes
Content-Type: text/html; charset=utf-8
Content-Language: en
```

But because there was no `|| exit -1`  on the line installing it, CI jobs didn't fail.

https://github.com/microsoft/LightGBM/blob/d811f3a074d30924a88aebe93eea66df70ecb7dc/.ci/test_r_package.sh#L79

And those jobs just silently fell back to the R version already available from GitHub Actions (https://github.com/actions/virtual-environments/blob/main/images/macos/macos-11-Readme.md).

For example, look at this `r-package (macOS-10.15, clang, R 4.1, cmake)` job from two days ago: https://github.com/microsoft/LightGBM/runs/5800165206?check_suite_focus=true

R installation failed:

```text
...
installer: Error - the package path specified was invalid: '/Users/runner/work/LightGBM/LightGBM/R.pkg'.
...
```

But the job continued, and you can see that the job used R 4.1.3, the version installed by default by GitHub Actions for this runner ([link](https://github.com/actions/virtual-environments/blob/main/images/macos/macos-12-Readme.md)).

```text
* using R version 4.1.3 (2022-03-10)
```

Pushed https://github.com/microsoft/LightGBM/pull/5129/commits/43d60f658d6103c98dcbd516a28fb8ff31489eb6, targeting https://cran.r-project.org/bin/macosx/base/ for R 4.x builds so this doesn't happen again.

StrikerRUS:
yikes!

Thanks for fixing this!

Can this error be related to #4990?

StrikerRUS (approved):

jameslamb:
> Can this error be related to https://github.com/microsoft/LightGBM/issues/4990?

ah!!! I bet you're right! I can try that later (updating those jobs to `macOS-latest`). I think it's very very likely that they're related.

Issue #5131, opened 2022-04-06
LGBMRegressor with linear_tree has constant leafs


Pablo-Davila:
## Description
I am trying to build a gradient boosting tree with linear models in its leafs. I am using the parameter [linear_tree](https://lightgbm.readthedocs.io/en/latest/Parameters.html#linear_tree) for this purpose.

To check wether the leafs are actually linear I have built a sample dataset and a regression model with a single estimator with the code below. The results after running the script suggest that the leafs are constant values:

![meh](https://user-images.githubusercontent.com/59798058/162086223-684408d0-8a9c-4824-b72e-56d03bb0f204.png)

Am I missing something?

## Reproducible example
```python
from lightgbm import LGBMRegressor
from sklearn.model_selection import train_test_split
import matplotlib.pyplot as plt
import numpy as np


# Sample "dataset"
x = np.arange(0,30,0.1).reshape(-1,1)
y = np.sin(x).reshape(-1)

data = np.stack([
    y[i:i+6]
    for i in range(y.shape[0] - 6)
])
x_train, x_test, y_train, y_test = train_test_split(
    data[:, :-1],
    data[:, -1],
    test_size=0.25,
    shuffle=False
)

# Model
params = {
    'objective': 'regression',
    'learning_rate': 1,
    'n_estimators': 1,
    'max_depth': 2,
    'linear_tree': True,
}
model = LGBMRegressor(**params)

model.fit(x_train, y_train)

# Plotting
plt.plot(x,y)
plt.plot(x[5:x_train.shape[0]+5].reshape(-1), model.predict(x_train))
plt.plot(x[-x_test.shape[0]:].reshape(-1), model.predict(x_test))
plt.legend(["real", "prediction (train)", "prediction (test)"])
plt.show()
```

## Environment info

LightGBM version or commit hash: v3.3.2

Pablo-Davila:
Hi @jmoralez, thank you very much for your quick and accurate response. This definitely solves my issue.

PR #5132, opened 2022-04-07
[docs] Document behaviour of the first linear estimator

Pablo-Davila:
Related to #5131

### DIFF
--- a/docs/Parameters.rst
+++ b/docs/Parameters.rst
@@ -663,6 +663,8 @@ Dataset Parameters
       -  tree splits are chosen in the usual way, but the model at each leaf is linear instead of constant
 
       -  the linear model at each leaf includes all the numerical features in that leaf's branch
+      
+      -  the first tree has constant leaf values
 
       -  categorical features are used for splits as normal but are not used in the linear models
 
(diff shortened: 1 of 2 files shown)


jameslamb:
Thanks so much for taking the time to improve LightGBM, @Pablo-Davila ! I support this change to the docs.

In this project, the file `Parameters.rst` is automatically generated from comments in LightGBM's C API code. Could you please add this change at https://github.com/microsoft/LightGBM/blob/3798f8720f4fbc4f1249717fb8f2f037e0075a46/include/LightGBM/config.h#L588, and then run the following from the root of the repo?

```shell
python helpers/parameter_generator.py
```

I noticed the branch from your fork is called `patch-1`, which I'm guessing means you made this edit in the GitHub UI. If you'd prefer not to clone LightGBM to make this change, please let me know and I'd be happy to push this requested change to this branch.

jameslamb (approved):
Thanks very much for this!

And for providing such a complete reproducible example in #5131. We really really appreciate it.

Pablo-Davila:
Thank you for your guidance 😊

PR #5133, opened 2022-04-07
[R-package] Use `type` argument to control prediction types

david-cortes:
The `predict` function for lightgbm model objects has three mutually-exclusive parameters (`rawscore`, `predcontrib`, `predleaf`) which control the type of output to produce in the predictions.

The convention in R is to make these types controllable through a `type` parameter, as done in base R's `predict.glm` and in many core packages for decision trees such as `rpart`, `randomForest`, `gbm`, and other popular decision tree packages such as `ranger`, `C50`, etc.

Some popular packages do take booleans, but mostly when they output more than one prediction type per function call, which lightgbm doesn't do.

In order to make the R interface more similar to base R and to other packages, this PR replaces those boolean arguments with a `type` argument.

Additionally, it adds a `"response"` type which will output the predicted class in classification objectives (something that's present in the Python interface but missing from the R interface).


R-package/R/lgb.Booster.R
```diff
@@ -799,18 +811,36 @@ predict.lgb.Booster <- function(object,
...
-      data = newdata
-      , start_iteration = start_iteration
-      , num_iteration = num_iteration
-      , rawscore = rawscore
-      , predleaf =  predleaf
-      , predcontrib =  predcontrib
-      , header = header
-      , params = params
-    )
+  type <- head(type, 1L)
```
jameslamb, line 814: What is the purpose of having `type` default to a vector with 5 elements, and then always taking only the first thing provided? It seems to me that it would be simpler to just default to `"link"`.

jameslamb (changes requested):
Thanks very much for this proposal!

I'm open to considering this breaking change, but since the goal is just consistency with other projects, I need some more information.

Can you please tell me specifically which packages you're referring to in this statement?

> Some popular packages do take booleans

david-cortes:
> Thanks very much for this proposal!
> 
> I'm open to considering this breaking change, but since the goal is just consistency with other projects, I need some more information.
> 
> Can you please tell me specifically which packages you're referring to in this statement?
> 
> > Some popular packages do take booleans

Amonf others, `randomForest` works like that.

R-package/R/lgb.Booster.R
david-cortes: The purpose is to have the allowed values visible in the function signature so that they are easily seen by the user and easy to autocomplete, in the same way as for example base R's `predict.glm`.

StrikerRUS:
Cross-reference: https://github.com/dmlc/xgboost/pull/7947.

R-package/R/lgb.Booster.R
```diff
@@ -713,6 +713,23 @@ Booster <- R6::R6Class(
 #' @param object Object of class \code{lgb.Booster}
 #' @param newdata a \code{matrix} object, a \code{dgCMatrix} object or
 #'                a character representing a path to a text file (CSV, TSV, or LibSVM)
+#' @param type Type of prediction to output. Allowed types are:\itemize{
```
jameslamb, line 716: Can you please add a note to this documentation that when choosing `"link"` and `"response"`, if you're using a custom objective function they'll be ignored and `"raw"` predictions will be returned?

On the Python side, `lightgbm` raises a warning in such situations.

https://github.com/microsoft/LightGBM/blob/f715645725bcfd7750b9f0a535ef9fd67454eb16/python-package/lightgbm/sklearn.py#L1064-L1067

The R package should probably do that to, but that could be deferred to a later PR.

R-package/R/lgb.Booster.R
```diff
@@ -770,15 +784,13 @@ Booster <- R6::R6Class(
...
-#' @importFrom utils modifyList
+#' @importFrom utils modifyList head
 #' @export
 predict.lgb.Booster <- function(object,
                                 newdata,
+                                type = c("link", "response", "raw", "leaf", "contrib"),
                                 start_iteration = NULL,
                                 num_iteration = NULL,
-                                rawscore = FALSE,
-                                predleaf = FALSE,
```
jameslamb, line 780: Down in the section where this function catches arguments that fall into `...`

https://github.com/microsoft/LightGBM/blob/f715645725bcfd7750b9f0a535ef9fd67454eb16/R-package/R/lgb.Booster.R#L822-L824

Please add the following:

```r
if (isTRUE(additional_params[["rawscore"]])) {
    stop("Argument 'rawscore' is no longer supported. Use type = 'raw' instead.")
}
if (isTRUE(additional_params[["predleaf"]])) {
    stop("Argument 'predleaf' is no longer supported. Use type = 'leaf' instead.")
}
if (isTRUE(additional_params[["predcontrib"]])) {
    stop("Argument 'predcontrib' is no longer supported. Use type = 'contrib' instead.")
}
```

I'm ok with breaking users' code in the next release in exchange for making the package's interface more compatible with other packages for modeling in R, but I think we should provide specific, actionable error messages when possible to reduce the effort required for affected users to alter their code.

jameslamb (changes requested):
Thanks for this!  I'm generally supportive of this proposal, but left a few suggestions on the implementation.

Ideally we'd do a round of deprecation warnings first (as `{xgboost}` is planning to do, https://github.com/dmlc/xgboost/pull/7947#issuecomment-1140672894), but the reality in this project right now is that due to a lack of maintainer attention, we're only doing 1-2 releases a year. In addition, `{lightgbm}` has very few reverse dependencies on CRAN, and the next release will be a large major-version release with breaking changes (#5153).

@jmoralez @StrikerRUS what do you think?

@mayer79 we'd also welcome your opinion on this if you have one.

mayer79:
@jameslamb : the proposal is as it should have been from the beginning. Unfortunately, now it seems too late to change key arguments of a key function. It will break multiple R packages (SHAPforxgboost, fastshap, plus one I am currently writing) and a lot of existing code, especially in the field of XAI. For what Version is the change planned?

david-cortes:
Updated.

StrikerRUS:
@jameslamb 
> what do you think?

I think we can omit deprecation cycle in the upcoming `v4.0.0` release.

jameslamb:
> now it seems too late
>
> For what Version is the change planned?

@mayer79 thanks very much for your feedback! Hearing you say that this is "how it should have been from the beginning" echoes how I was feeling and I think what @david-cortes is thinking as well.

I understand that the breaking change will cause some pain, but I think it is necessary to get the package to the state we that we want. The reality we're facing in LightGBM right now is a significant lack of maintainer attention in this project + the fact that it has been almost 8 months since the previous substantial release (v3.3.1, in October 2021). We simply don't have the development velocity in this project to do deprecation cycles the way that healthier projects like scikit-learn can.

These changes are planned for a v4.0.0 major release (#5153), after which we will be much more skeptical of breaking changes anywhere in the project.

> break multiple R packages (SHAPforxgboost, fastshap, plus one I am currently writing)

I'm willing to go submit PRs to all of those projects listed as reverse dependencies of `{lightgbm}` (https://cran.r-project.org/web/packages/lightgbm/index.html) to make the forward- and backward-compatible so the new release won't break them. For example, this PR is not changing the public method `Booster$predict()` at all, so it's possible for `{SHAPforxgboost}`, `{fastshap}`, etc. to switch to using that method and not be broken by a new `{lightgbm}` release. I will go open issues on those projects and any other you'd like to point me to.

R-package/tests/testthat/test_Predictor.R
```diff
@@ -442,8 +436,52 @@ test_that("predictions for multiclass classification are returned as matrix", {
...
     expect_equal(nrow(pred), nrow(X))
     expect_equal(ncol(pred), 3L)
-    pred <- predict(model, X, rawscore = TRUE)
+    pred <- predict(model, X, type = "raw")
     expect_true(is.matrix(pred))
     expect_equal(nrow(pred), nrow(X))
     expect_equal(ncol(pred), 3L)
 })
+
+test_that("predict type='response' returns integers for classification objectives", {
```
jameslamb, line 445:
```suggestion
test_that("predict type='response' returns predicted class for classification objectives", {
```

The significant thing here is that the result is the predicted classes, not just that they're integers. Would you please consider this rewording?

R-package/tests/testthat/test_Predictor.R
```diff
@@ -442,8 +436,52 @@ test_that("predictions for multiclass classification are returned as matrix", {
...
+      , obj = "multiclass"
+      , nrounds = 5L
+      , verbose = VERBOSITY
+      , params = list(num_class = 3L)
+    )
+    pred <- predict(model, X, type = "response")
+    expect_true(all(pred %in% c(0L, 1L, 2L)))
+})
+
+test_that("predict type='response' returns decimals for regression objectives", {
```
jameslamb, line 474:
```suggestion
test_that("predict type='response' returns values in the target's range for regression objectives", {
```

jameslamb (approved):
Two more very small suggestions, otherwise I am good with these changes. Thanks very much for the work!

Since this is such a significant breaking change, @jmoralez would you also please help with a review?

jmoralez:
Sure. I'll add my review tomorrow

mayer79:
> > now it seems too late
> > For what Version is the change planned?
> 
> @mayer79 thanks very much for your feedback! Hearing you say that this is "how it should have been from the beginning" echoes how I was feeling and I think what @david-cortes is thinking as well.
> 
> I understand that the breaking change will cause some pain, but I think it is necessary to get the package to the state we that we want. The reality we're facing in LightGBM right now is a significant lack of maintainer attention in this project + the fact that it has been almost 8 months since the previous substantial release (v3.3.1, in October 2021). We simply don't have the development velocity in this project to do deprecation cycles the way that healthier projects like scikit-learn can.
> 
> These changes are planned for a v4.0.0 major release (#5153), after which we will be much more skeptical of breaking changes anywhere in the project.
> 
> > break multiple R packages (SHAPforxgboost, fastshap, plus one I am currently writing)
> 
> I'm willing to go submit PRs to all of those projects listed as reverse dependencies of `{lightgbm}` (https://cran.r-project.org/web/packages/lightgbm/index.html) to make the forward- and backward-compatible so the new release won't break them. For example, this PR is not changing the public method `Booster$predict()` at all, so it's possible for `{SHAPforxgboost}`, `{fastshap}`, etc. to switch to using that method and not be broken by a new `{lightgbm}` release. I will go open issues on those projects and any other you'd like to point me to.

Thanks for your detailed answer, James. The packages I have mentioned use `predcontrib = TRUE`, so they will be affected. As soon as this PR is merged, I can fix SHAPforxgboost (I already drafted a fix) and `shapviz` (a package that does not load LGB but still works with it...) plus ping Brandon from fastshap. I think we only need a little bit of time (3-4 weeks) between this PR and switching to LGB 4 on CRAN.

mayer79 (commented):
@jameslamb @david-cortes @jmoralez  I like the new "type" argument. However, if I correctly understood the intended changes, they are in gross contrast what usual prediction functions involving link functions do. Sorry for this late input. I think it is important to consider it anyway.

According to `predict.glm()`, type "response" returns values on the scale of the response.  A Poisson regression with log link would return positive values, a binary logistic regression would return probabilities.

Type "link" means, on the other hand, that predictions are made in the link space. For a logistic regression, it would be the logit of the probabilities, for a Poisson regression the log expected counts etc. 

**LightGBM seems to use "link" in the exact opposite, i.e., in the sense that the inverse link is being applied. But this is what is usually be called "response"...**

My suggestions: 

- The default type is "response". Binary logistic will return probabilities. 
- There is either "raw" or "link" but not both. For historic reasons, I tend to "raw", but "link" is fine as well.
- There should be a type = "class" in order to get the majority class for those objectives where it makes sense. `glm.predict()`does not have this, but for instance "predict.glmnet()".
- Type "leaf" and "contrib" are okay.

Side note: `predict.glm()` uses `type = "link"` as the default, i.e., it predicts log odds instead of probabilities for a logistic regression. We should not do it similarly because then every logistic/Poisson/Gamma/Tweedie model would be affected from the change.

david-cortes:
@mayer79 Thanks for spotting the mismatch w.r.t. base R. I've updated it by changing the names to "response" for what's obtained after applying the link function and "class" for the type that predicts class in classification objectives.

david-cortes:
The linter check doesn't say what exactly failed, don't know what it is complaining about: https://github.com/microsoft/LightGBM/runs/6887268528?check_suite_focus=true

jameslamb:
@david-cortes that error you're facing comes from `{lintr}`.

```text
Linting R code
[1] 52 R files need linting
Error in match.arg(semicolon, several.ok = TRUE) : 
  'arg' must be NULL or a character vector
Calls: <Anonymous> ... flatten_list -> assign_item -> linter_fun -> match.arg
In addition: There were 25 warnings (use warnings() to see them)
Execution halted
```

There was a new major release of `{lintr}` to CRAN today. I suspect rebuilding will fix this, since I just merged #5290. I'll re-trigger this job.

jmoralez (approved):

mayer79 (approved):

R-package/tests/testthat/test_Predictor.R
```diff
@@ -240,9 +240,7 @@ test_that("predict() params should override keyword argument for raw-score predi
         , nm = rawscore_alias
       )
     )
-    preds_raw_s3_param <- predict(bst, X, params = params)
     preds_raw_r6_param <- bst$predict(X, params = params)
-    expect_equal(preds_raw_s3_keyword, preds_raw_s3_param)
```
jameslamb, line 245: Sorry, I just noticed this. @david-cortes , why are you proposing removing support for passing something like `"predict_raw_score"` through params?

That effectively reverts part of #5122, which was a fix for an issue you opened (#4670).

I feel that `{lightgbm}` should continue to support changing the type of prediction via `params` entries, and that `params` entries should take precedence over the `type` keyword argument in `predict.lgb.Booster()`, unless you can provide a compelling version to remove that support as part of this PR.

R-package/tests/testthat/test_Predictor.R
```diff
@@ -524,8 +518,52 @@ test_that("predictions for multiclass classification are returned as matrix", {
...
+    y <- agaricus.train$label
+    dtrain <- lgb.Dataset(X, label = y, params = list(max_bins = 5L))
+    bst <- lgb.train(
+        data = dtrain
+        , obj = "regression"
+        , nrounds = 5L
+        , verbose = VERBOSITY
+    )
+    pred <- predict(bst, X, type = "class")
+    expect_true(all(!(pred %in% c(0.0, 1.0))))
```
jameslamb, line 568:
```suggestion
    expect_true(!any(pred %in% c(0.0, 1.0)))
```

Since the last time a commit on this PR successfully passed CI, LightGBM has upgraded its `lint` CI job to `{lintr}` v3.0.0 and added several new linters (#5303).

Please accept this suggestion to resolve the one linter warning raised in https://github.com/microsoft/LightGBM/runs/6995392727?check_suite_focus=true.

```text
[[1]]
/home/runner/work/LightGBM/LightGBM/R-package/tests/testthat/test_Predictor.R:568:17: warning: [outer_negation] !any(x) is better than all(!x). The former applies negation only once after aggregation instead of many times for each element of x.
    expect_true(all(!(pred %in% c(0.0, 1.0))))
                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

jameslamb (changes requested):
Thanks very much to this, and to @mayer79 for the excellent review comments!

I re-reviewed tonight. Please see a few more suggestions that I believe should be addressed before this is merged.

R-package/tests/testthat/test_Predictor.R
david-cortes: Restored them.

jameslamb (approved):
Changes look good to me, thanks very much! If you can resolve the merge conflicts with `master` and get this updated, we'll merge it.

david-cortes:
Updated.

Issue #4705, opened 2021-10-22
Calling multithreaded functions sets global number of OMP threads


david-cortes:
Calling lightgbm functions which use multithreading will forcibly change the configured number of openmp threads in the whole processes from where lightgbm was called, which is a bad practice (for example, depending on how numpy was built, it might change the number of threads that matrix multiplications in numpy will use afterwards). This can be checked in lines such as this: https://github.com/microsoft/LightGBM/blob/a2b60e84c17462d6e467f4363a0fae203ca7fee5/src/c_api.cpp#L2144

The fix is easy but requires some design changes in the overall code structure: don't set the number of openmp threads, pass them instead as a parameter to the openmp pragmas.

Example:
```python
import sklearn
import threadpoolctl
threadpoolctl.threadpool_info()
```
```
[{'filepath': '/home/david/anaconda3/envs/py3/lib/libopenblasp-r0.3.13.so',
  'prefix': 'libopenblas',
  'user_api': 'blas',
  'internal_api': 'openblas',
  'version': '0.3.13',
  'num_threads': 16,
  'threading_layer': 'pthreads',
  'architecture': 'Zen'},
 {'filepath': '/home/david/anaconda3/envs/py3/lib/libgomp.so.1.0.0',
  'prefix': 'libgomp',
  'user_api': 'openmp',
  'internal_api': 'openmp',
  'version': None,
  'num_threads': 16}]
```
```python
from lightgbm import LGBMRegressor
from sklearn.datasets import fetch_california_housing
X, y = fetch_california_housing(return_X_y=True)

model = LGBMRegressor(n_jobs=3).fit(X, y)
p = model.predict(X)
threadpoolctl.threadpool_info()
```
```
[{'filepath': '/home/david/anaconda3/envs/py3/lib/libopenblasp-r0.3.13.so',
  'prefix': 'libopenblas',
  'user_api': 'blas',
  'internal_api': 'openblas',
  'version': '0.3.13',
  'num_threads': 16,
  'threading_layer': 'pthreads',
  'architecture': 'Zen'},
 {'filepath': '/home/david/anaconda3/envs/py3/lib/libgomp.so.1.0.0',
  'prefix': 'libgomp',
  'user_api': 'openmp',
  'internal_api': 'openmp',
  'version': None,
  'num_threads': 3}]
```

StrikerRUS:
Related: https://github.com/dmlc/xgboost/pull/7537.

guolinke:
[...]

PR #5134, opened 2022-04-07
[R-package] Temporary workaround for not modifying global OMP config

david-cortes:
ref #4705

This is a temporary workaround to avoid modifying the global OpenMP user configuration for the whole R process when running `lgb.train`. This PR should be reverted once issue #4705 is fixed properly.


jameslamb (changes requested):
Thanks for your proposal, but I don't support introducing a new required dependency to the R package for this purpose.

Instead of this fix that benefits only the R package, and even then only one part of it (i.e. not other multi-threaded operations like prediction and `Dataset` construction), if you're interested in seeing LightGBM's behavior here change, please work with @guolinke and @shiyu1994 on #4705 to implement something at the C++ level that all users of LightGBM will benefit from.

david-cortes:
In #4705, @guolinke suggested a workaround like this. It does BTW fix the issue for dataset construction when it happens inside `lgb.train`.

shiyu1994:
I agree that such changes should be made in C++ so that both R and Python packages can leverage the workaround. I can help to do that.

jameslamb:
> both R and Python packages

Thanks @shiyu1994 . For the benefit of everyone reading this PR, I just want to clarify...when I refer to wrappers of LightGBM's C++ code, I don't only mean the R and Python package in this repo. I also mean SynapseML (formerly MMLSpark), the Java interface generated with SWIG, `LightGBM.NET`, and the other projects listed at https://github.com/microsoft/LightGBM#external-unofficial-repositories.

Based on this discussion, I'm going to close this PR.

@david-cortes thanks very much for documenting #4705 and for helping this project try to address it!

PR #5141, opened 2022-04-11
[R-package] reduce cost of repeated parameter alias checks

jameslamb:
Created in response to https://github.com/microsoft/LightGBM/pull/5122#discussion_r846855854.

For many of LightGBM's supported  parameters (https://lightgbm.readthedocs.io/en/latest/Parameters.html), LightGBM respects multiple *aliases*. For example, `num_iterations=10`, `nrounds=10`, and `num_tree=10` all result in the same behavior.

As a result of this, the R package frequently needs to deal with the possibility that some relevant configuration might be provided via any of these aliases, and sometimes via special keyword arguments in the R package's public API.

Earlier versions of `{lightgbm}` included a hard-coded list of string literals with a mapping from select "main" parameter names to their aliases. To ensure that the package is always in sync with changes on the C++ side, #4829 replaced that list with a call to an entrypoint in LightGBM's C++ code which returns a JSON string containing all parameters and their aliases.

As noted in https://github.com/microsoft/LightGBM/pull/5122#discussion_r846818117, each of those calls results in re-generating that JSON string and deserializing it into an R list.

Since the parameter mapping will never change within an R session, it should only be necessary to fetch it once. This PR proposes introducing an internal cache in the R package and changing `.PARAMETER_ALIASES()` to check that cache before going to the C++ side to get parameter aliases.

### DIFF
--- a/R-package/R/aliases.R
+++ b/R-package/R/aliases.R
@@ -33,11 +33,18 @@
     )])
 }
 
+# [description] Non-exported environment, used for caching details that only need to be
+#               computed once per R session.
+.lgb_session_cache_env <- new.env()
+
 # [description] List of respected parameter aliases. Wrapped in a function to take advantage of
 #               lazy evaluation (so it doesn't matter what order R sources files during installation).
 # [return] A named list, where each key is a main LightGBM parameter and each value is a character
 #          vector of corresponding aliases.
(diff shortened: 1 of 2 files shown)


jmoralez (approved):

StrikerRUS (approved):
Very nice indeed!

PR #5148, opened 2022-04-13
[python] Use predefined constant in feature importance type comparison instead of raw int literal

StrikerRUS:
https://github.com/microsoft/LightGBM/blob/b462d0a40f6978c0a66a47af4522c2ae12532316/python-package/lightgbm/basic.py#L453


jameslamb (approved):
ah nice, thank you!

PR #5149, opened 2022-04-13
[ci] bump CUDA version from `11.5.1` to `11.6.2` at CI

StrikerRUS:
https://catalog.ngc.nvidia.com/orgs/nvidia/containers/cuda/tags

Linking #3763.


jameslamb (approved):
🎉 excellent!

PR #5160, opened 2022-04-20
[python-package] support saving and loading CVBooster (fixes #3556)

### LINKED ISSUE #3556: [python] support saving and loading CVBooster
The two most immediately apparent ways to save a cvbooster (to me at least) would be to use pickle, or save_model() as is done with basic booster models. However, neither of these appear to work:

```
import pandas as pd
import numpy as np
import lightgbm as lgb
import dill

data = pd.DataFrame(np.random.rand(100, 10))
data.columns = [str(i) for i in range(len(data.columns))]
target = data.pop("9")
params = {
    'learning_rate': 0.1,
    'min_data_in_leaf':3,
    'verbose': -1
}
dtrain = lgb.Dataset(data=data)
lgbcv = lgb.cv(params,train_set=dtrain,return_cvbooster=True)
cvbooster = lgbcv.pop("cvbooster")

# Using save_model()
cvbooster.save_model("File.txt")

# Fails
cvbooster_from_file = lgb.CVBooster(model_file = "File.txt")

# Returns a single booster
cvbooster_from_file = lgb.Booster(model_file = "File.txt")


######## Using pickle (dill)
# Save
with open("File.pkl","wb") as f:
    dill.dump(cvbooster,f)

# Fails
with open("File.pkl","rb") as f:
    cvbooster_from_pickle = dill.load(f)
```

Would it be possible to get a `.save_model()` method for CVBooster?

nyanp:
This PR resolves #3556 by implementing the following method in `CVBooster`:

- save_model
- model_from_string
- model_to_string


python-package/lightgbm/engine.py
```diff
@@ -276,18 +278,42 @@ class CVBooster:
         The best iteration of fitted model.
     """
 
-    def __init__(self):
+    def __init__(self, model_file=None):
```
jameslamb, line 281:
```suggestion
    def __init__(
        self,
        model_file: Optional[str, pathlib.Path] = None
    ):
```

Per #3756, we are trying to encourage more uses of type hints in the Python package, to improve the ability of type checkers like `mypy` to catch bugs. Can you please add hints here?

python-package/lightgbm/engine.py
```diff
@@ -276,18 +278,42 @@ class CVBooster:
...
 
+        if model_file is not None:
+            with open(str(model_file), "r") as file:
+                self._from_dict(json.load(file))
+
     def _append(self, booster):
         """Add a booster to CVBooster."""
         self.boosters.append(booster)
 
+    def _from_dict(self, models):
```
jameslamb, line 302:
```suggestion
    def _from_dict(self, models: Dict[str, any]) -> None:
```

python-package/lightgbm/engine.py
```diff
@@ -276,18 +278,42 @@ class CVBooster:
...
+        """Load CVBooster from dict."""
+        self.best_iteration = models["best_iteration"]
+        self.boosters = []
+        for model_str in models["boosters"]:
+            self.boosters.append(Booster(model_str=model_str))
+
+    def _to_dict(self, num_iteration, start_iteration, importance_type):
```
jameslamb, line 309:
```suggestion
    def _to_dict(self, num_iteration: int, start_iteration: int, importance_type: str) -> Dict[str, Any]:
```

python-package/lightgbm/engine.py
```diff
@@ -298,6 +324,73 @@ def handler_function(*args, **kwargs):
             return ret
         return handler_function
 
+    def model_from_string(self, model_str):
```
jameslamb, line 327:
```suggestion
    def model_from_string(self, model_str: str) -> "CVBooster":
```

python-package/lightgbm/engine.py
```diff
@@ -298,6 +324,73 @@ def handler_function(*args, **kwargs):
...
+
+        Returns
+        -------
+        self : CVBooster
+            Loaded CVBooster object.
+        """
+        self._from_dict(json.loads(model_str))
+        return self
+
+    def model_to_string(self, num_iteration=None, start_iteration=0, importance_type='split'):
```
jameslamb, line 343:
```suggestion
    def model_to_string(
        self,
        num_iteration: Optional[int] = None,
        start_iteration: int = 0,
        importance_type: str = 'split'
    ) -> str:
```

python-package/lightgbm/engine.py
```diff
@@ -298,6 +324,73 @@ def handler_function(*args, **kwargs):
...
+            If "gain", result contains total gains of splits which use the feature.
+
+        Returns
+        -------
+        str_repr : str
+            String representation of CVBooster.
+        """
+        return json.dumps(self._to_dict(num_iteration, start_iteration, importance_type))
+
+    def save_model(self, filename, num_iteration=None, start_iteration=0, importance_type='split'):
```
jameslamb, line 366:
```suggestion
    def save_model(
        self,
        filename: Union[str, pathlib.Path],
        num_iteration: Optional[int] = None,
        start_iteration: int = 0,
        importance_type: str = 'split'
    ) -> "CVBooster":
```

tests/python_package_test/test_engine.py
```diff
@@ -1068,6 +1068,39 @@ def test_cvbooster():
...
+        'verbose': -1,
+    }
+    nfold = 3
+    lgb_train = lgb.Dataset(X_train, y_train)
+
+    def predict(cv_booster):
+        return np.array(cv_booster.predict(X_test))
+
+    cv_res = lgb.cv(params, lgb_train,
+                    num_boost_round=25,
```
jameslamb, line 1086:
```suggestion
                    num_boost_round=10,
```

We try to keep these tests as inexpensive as possible, to ensure that they run quickly. Could `num_boost_round` be reduced to 10?

tests/python_package_test/test_engine.py
```diff
@@ -1068,6 +1068,39 @@ def test_cvbooster():
...
+    ret_origin = predict(cvbooster)
+    other_ret = []
+
+    cvbooster.save_model('lgb.model')
+
+    other_ret.append(predict(lgb.CVBooster(model_file='lgb.model')))
+    other_ret.append(predict(lgb.CVBooster().model_from_string(cvbooster.model_to_string())))
+
+    for ret in other_ret:
+        np.testing.assert_array_equal(ret_origin, ret)
```
jameslamb, line 1101:
```suggestion
    preds = predict(cvbooster)

    cvbooster.save_model('lgb.model')
    preds_from_file = predict(lgb.CVBooster(model_file='lgb.model')))
    preds_from_string =  predict(lgb.CVBooster().model_from_string(cvbooster.model_to_string())))

    np.testing.assert_array_equal(preds, preds_from_file)
    np.testing.assert_array_equal(preds, preds_from_string)
```

I think the indirection added by using this empty list and then a `for` loop over it makes it a bit difficult to understand what is being tested here. Will you please consider a rewrite like this?

jameslamb: Also, to ensure that files written by these tests are automatically cleaned up, please use the `pytest` `tmp_path` fixture.

For example:

https://github.com/microsoft/LightGBM/blob/6b56a90cd1324a6dbac2afa0a352c9355b0dc3cf/tests/python_package_test/test_engine.py#L2922

https://github.com/microsoft/LightGBM/blob/6b56a90cd1324a6dbac2afa0a352c9355b0dc3cf/tests/python_package_test/test_engine.py#L3023-L3024

jameslamb (changes requested):
Thanks very much for taking this on!

I will provide a more thorough review in the next few days, for now please see my initial round of suggestions.

Also note that this does not totally resolve #3556. As mentioned in the discussion there, it should also be possible to pickle a `CVBooster` and load one from a pickle file. I think that work should be picked up in a follow-up PR after this one.

python-package/lightgbm/engine.py
nyanp: @jameslamb 
I think that is a great idea. Thanks for the suggestion!

tests/python_package_test/test_engine.py
nyanp: @jameslamb 
I wrote it with reference to the other test case (`test_save_load_copy_pickle`), but it seems that your suggestion is more concise and readable. Thanks ;)

tests/python_package_test/test_engine.py
nyanp: 👍

nyanp:
> Also note that this does not totally resolve #3556. As mentioned in the discussion there, it should also be possible to pickle a `CVBooster` and load one from a pickle file. I think that work should be picked up in a follow-up PR after this one.

This can be solved by simply defining `__getstate__` and `__setstate__` in `CVBooster`, which I will try to solve together in this PR.

tests/python_package_test/test_engine.py
```diff
@@ -1072,6 +1072,46 @@ def test_cvbooster():
...
+                    return_cvbooster=True)
+    cvbooster = cv_res['cvbooster']
+
+    preds = predict(cvbooster)
+
+    model_path_txt = str(tmp_path / 'lgb.model')
+    model_path_pkl = str(tmp_path / 'lgb.pkl')
+
+    cvbooster.save_model(model_path_txt)
+    with open(model_path_pkl, 'wb') as f:
```
jameslamb, line 1102: This project tries to support `pickle`, `joblib`, and `cloudpickle`.

Could you please modify this test to use all 3 of those serializers?

Some references to similar tests:

https://github.com/microsoft/LightGBM/blob/56ccea4243aba4c538ebca9de8c67ee0c9bc60eb/tests/python_package_test/test_callback.py#L33

https://github.com/microsoft/LightGBM/blob/56ccea4243aba4c538ebca9de8c67ee0c9bc60eb/tests/python_package_test/test_callback.py#L8-L22

tests/python_package_test/test_engine.py
```diff
@@ -1072,6 +1072,46 @@ def test_cvbooster():
...
+        pickle.dump(cvbooster, f)
+
```
jameslamb, line 1104:
```suggestion
del cvboost
```

Right here, after the `cvbooster` object is no longer needed, could you please delete it?

That would call the destructors of each of the underlying `Booster` objects, and would make me more confident that a `CVBooster` could be pickled in one processed and then unpickled in another.

Said another way...after training, the `CVBooster` contains multiple `Booster` objects. Those are Python objects which contain a `handle`, a memory address to a `Booster` object allocated by LightGBM's C++ code. If just that memory address is stored in the pickle file, then re-loading such a file *in the same process* might work because there's still a `Booster` at that same address in memory...but that would probably not be true in a new process.

tests/python_package_test/test_engine.py
```diff
@@ -1072,6 +1072,46 @@ def test_cvbooster():
...
+    params = {
+        'objective': 'binary',
+        'metric': 'binary_logloss',
+        'verbose': -1,
+    }
+    nfold = 3
+    lgb_train = lgb.Dataset(X_train, y_train)
+
+    def predict(cv_booster):
+        return np.array(cv_booster.predict(X_test))
```
jameslamb, line 1087: I'm concerned that having this function reach up into the calling environment for an object that was not passed to it (`X_test`) makes it a bit hard to understand what these tests are predicting on.

In addition, this method is very thin (just provides a functional API to `CVBooster.predict()`), so I don't think it's actually providing a consistency benefit to the tests, and therefore isn't worth the indirection.

Could you please remove this inner function, and just modify the `preds` lines like this:

```python
preds = cvbooster.predict(X_test)
preds_from_txt_file = lgb.CVBooster(model_file=model_path_txt).predict(X_test)
model_str = cvbooster.model_to_string()
preds_from_string = lgb.CVBooster().model_from_string(model_str).predict(X_test)
with open(model_path_pkl, 'rb') as f:
    preds_from_pkl = pickle.load(f).predict(X_test)
``

File: tests/python_package_test/test_engine.py [hunk 12]
```diff
@@ -1072,6 +1072,46 @@ def test_cvbooster():
...
+    preds = predict(cvbooster)
+
+    model_path_txt = str(tmp_path / 'lgb.model')
+    model_path_pkl = str(tmp_path / 'lgb.pkl')
+
+    cvbooster.save_model(model_path_txt)
+    with open(model_path_pkl, 'wb') as f:
+        pickle.dump(cvbooster, f)
+
+    preds_from_file = predict(lgb.CVBooster(model_file=model_path_txt))
```
jameslamb, line 1105:
```suggestion
    preds_from_txt_file = predict(lgb.CVBooster(model_file=model_path_txt))
```

Could you please rename this `preds_from_txt_file`? To make it clear how this is different from `preds_from_pkl` (since those are also produced based on a model stored in a file).

jameslamb (changes requested):
Thanks for the test changes! I'd originally suggested splitting pickling into a separate PR to keep this one small, but now that I see how small the changes are I think it's ok to leave all in this PR.

Please see a few more recommendations on the tests. I'll provide a more thorough review once I have a chance to run this code myself and inspect the outputs.

tests/python_package_test/test_engine.py
nyanp: @jameslamb 
Thanks! I'll give it a try. Can I move `SERIALIZERS` and `pickle_and_unpickle_object` from `test_serializers.py` to `tests/python_package_test/utils.py` to reuse them? I think these do not appear to be callback-specific functions, but rather generic codes.

tests/python_package_test/test_engine.py
nyanp: That makes sense. I've fixed it at [12c2fca](https://github.com/microsoft/LightGBM/pull/5160/commits/12c2fca07c9ca65f763ebc96144a18d46fa441ca).

tests/python_package_test/test_engine.py
jameslamb: Yep, moving those and re-using them sounds good to me!

nyanp: Thanks, done in  [3ec7fd0](https://github.com/microsoft/LightGBM/pull/5160/commits/3ec7fd02ebefa64b560135b14fcd3f44ab0affda). Now it seemed too many things to write everything in the single test, so I've extracted the pickle test into a separate function.

nyanp:
I signed the CLA and once the status was OK, but at some point, the bot status changed again to "not signed yet".
When I access the CLA page (https://cla.opensource.microsoft.com/microsoft/LightGBM?pullRequest=5160) again, it says "You have signed the CLA for multiple repositories or organizations" and I cannot re-sign. 
Can anyone help me?

StrikerRUS:
@nyanp Sorry about this inconvenience!
I believe we can "fix" it via closing-reopenning this PR.

nyanp:
@StrikerRUS 
The CLA bot is still complaining, but CI seems to have passed. Thanks a lot!

@jameslamb 
All you pointed out have been fixed. Could you please review them again?

StrikerRUS:
> The CLA bot is still complaining, but CI seems to have passed. Thanks a lot!

Yeah, I see. Check mark is green

image

I believe this is enough.

python-package/lightgbm/engine.py
```diff
@@ -299,6 +328,90 @@ def handler_function(*args, **kwargs):
...
+        self._from_dict(json.loads(model_str))
+        return self
+
+    def model_to_string(
+        self,
+        num_iteration: Optional[int] = None,
+        start_iteration: int = 0,
+        importance_type: str = 'split'
+    ) -> str:
+        """Save CVBooster to string.
```
jameslamb, line 359:
```suggestion
        """Save CVBooster to JSON string.
```

Can you please clarify that this method writes out JSON?

I totally understand (and support!) the decision to use JSON for this purpose, as that's a direction we'd like to move in this project (#4887).

But `CVBooster.model_to_string()` returning JSON does make it different from `Booster.model_to_string()`, which returns a LightGBM-custom-formatted text file.

https://github.com/microsoft/LightGBM/blob/c000b8cc689a9b3ee8bf4f294749d2f848f7176f/python-package/lightgbm/basic.py#L3401-L3408

Given that, I think it's important to over-communicate the fact that `CVBooster` uses JSON strings in its docs, since users may have been conditioned to think that the interface for `CVBooster` is very similar to the one for `Booster`.

python-package/lightgbm/engine.py
```diff
@@ -299,6 +328,90 @@ def handler_function(*args, **kwargs):
...
+            Start index of the iteration that should be saved.
+        importance_type : str, optional (default="split")
+            What type of feature importance should be saved.
+            If "split", result contains numbers of times the feature is used in a model.
+            If "gain", result contains total gains of splits which use the feature.
+
+        Returns
+        -------
+        str_repr : str
+            String representation of CVBooster.
```
jameslamb, line 377:
```suggestion
            JSON string representation of CVBooster.
```

python-package/lightgbm/engine.py
```diff
@@ -299,6 +328,90 @@ def handler_function(*args, **kwargs):
...
+        return json.dumps(self._to_dict(num_iteration, start_iteration, importance_type))
+
+    def save_model(
+        self,
+        filename: Union[str, Path],
+        num_iteration: Optional[int] = None,
+        start_iteration: int = 0,
+        importance_type: str = 'split'
+    ) -> "CVBooster":
+        """Save CVBoosters to file.
```
jameslamb, line 388:
```suggestion
        """Save CVBoosters to a file as JSON text.
```

jameslamb (changes requested):
Sorry it took so long to provide another review. I took a look tonight, and this looks great!

I just left one more set of minor comments.

Please update this branch to the latest `master`, and I'd like either @StrikerRUS or @jmoralez to also provide a review before we merge this.

Thanks again for all your hard work!

python-package/lightgbm/engine.py
```diff
@@ -299,6 +328,90 @@ def handler_function(*args, **kwargs):
...
+        model_str : str
+            Model will be loaded from this string.
+
+        Returns
+        -------
+        self : CVBooster
+            Loaded CVBooster object.
+        """
+        self._from_dict(json.loads(model_str))
+        return self
```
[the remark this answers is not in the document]
nyanp: @jameslamb 
These method names are given to override `model_from_string` and `model_to_string` of `Booster`. I agree that the class method is cleaner code for these use cases, but users may be confused because both `CVBooster.from_string` and `CVBooster.model_from_string` are callable.

I think it would be better to leave the method names as they are and specify in the docstring which methods are actually overridden, what do you think?


draft of docstring:
```
Auxiliary data structure to hold and redirect all boosters of ``cv`` function.
This class has the same methods as Booster class.
All method calls, except for the following methods, are actually performed for underlying Boosters and 
then all returned results are returned in a list.

- model_from_string
- model_to_string
- save_model
```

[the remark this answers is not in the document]
jameslamb: I agree with @nyanp and assumed that was the reason when I reviewed this...`Booster` has these methods as instance methods, so I think `CVBooster` should as well.

If we were re-designing these classes from scratch I'd agree that class methods are preferable, but I don't think such aa breaking change is worth doing now.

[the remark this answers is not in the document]
jameslamb: And I agree with the proposed docstring @nyanp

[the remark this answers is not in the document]
nyanp: Oops, sorry, I was replying to another account... Thanks for the suggestion @jmoralez , and thanks for your comment @jameslamb !

> If we were re-designing these classes from scratch I'd agree that class methods are preferable, but I don't think such aa breaking change is worth doing now.

Yes, I agree. I will modify the docstring ;)

nyanp:
@jameslamb 
Hi, could you please review it again? I believe all review points have been resolved. Thanks :)

python-package/lightgbm/engine.py
```diff
@@ -267,7 +268,12 @@ class CVBooster:
 
     Auxiliary data structure to hold and redirect all boosters of ``cv`` function.
```
StrikerRUS, line 269:
```suggestion
    Auxiliary data structure to hold and redirect all boosters of ``cv()`` function.
```

python-package/lightgbm/engine.py
```diff
@@ -267,7 +268,12 @@ class CVBooster:
...
     This class has the same methods as Booster class.
-    All method calls are actually performed for underlying Boosters and then all returned results are returned in a list.
+    All method calls, except for the following methods, are actually performed for underlying Boosters and
+    then all returned results are returned in a list.
+
+    - model_from_string
+    - model_to_string
+    - save_model
```
StrikerRUS, line 276:
```suggestion
    - ``model_from_string()``
    - ``model_to_string()``
    - ``save_model()``
```

python-package/lightgbm/engine.py
```diff
@@ -277,18 +283,45 @@ class CVBooster:
...
     """
 
-    def __init__(self):
+    def __init__(
+        self,
+        model_file: Optional[Union[str, Path]] = None
+    ):
         """Initialize the CVBooster.
 
         Generally, no need to instantiate manually.
```
StrikerRUS, line 292: Is this statement still valid?..

python-package/lightgbm/engine.py
```diff
@@ -277,18 +283,45 @@ class CVBooster:
...
+        Parameters
+        ----------
+        model_file : str, pathlib.Path or None, optional (default=None)
+            Path to the CVBooster model file.
         """
         self.boosters = []
         self.best_iteration = -1
 
+        if model_file is not None:
+            with open(str(model_file), "r") as file:
```
StrikerRUS, line 303:
```suggestion
            with open(model_file, "r") as file:
```

python-package/lightgbm/engine.py
```diff
@@ -277,18 +283,45 @@ class CVBooster:
...
     def _append(self, booster):
         """Add a booster to CVBooster."""
         self.boosters.append(booster)
 
+    def _from_dict(self, models: Dict[str, Any]) -> None:
+        """Load CVBooster from dict."""
+        self.best_iteration = models["best_iteration"]
+        self.boosters = []
+        for model_str in models["boosters"]:
+            self.boosters.append(Booster(model_str=model_str))
```
StrikerRUS, line 315:
```suggestion
            self._append(Booster(model_str=model_str))
```

python-package/lightgbm/engine.py
```diff
@@ -277,18 +283,45 @@ class CVBooster:
...
+
+    def _to_dict(self, num_iteration: int, start_iteration: int, importance_type: str) -> Dict[str, Any]:
```
StrikerRUS, line 317:
```suggestion
    def _to_dict(self, num_iteration: Optional[int], start_iteration: int, importance_type: str) -> Dict[str, Any]:
```

python-package/lightgbm/engine.py
```diff
@@ -299,6 +332,90 @@ def handler_function(*args, **kwargs):
...
+        num_iteration: Optional[int] = None,
+        start_iteration: int = 0,
+        importance_type: str = 'split'
+    ) -> "CVBooster":
+        """Save CVBoosters to a file as JSON text.
+
+        Parameters
+        ----------
+        filename : str or pathlib.Path
+            Filename to save Booster.
```
StrikerRUS, line 397:
```suggestion
            Filename to save CVBooster.
```

python-package/lightgbm/engine.py
StrikerRUS, line 276: What about [`dump_model()`](https://lightgbm.readthedocs.io/en/latest/pythonapi/lightgbm.Booster.html#lightgbm.Booster.dump_model) method?

tests/python_package_test/test_engine.py
```diff
@@ -1072,6 +1073,64 @@ def test_cvbooster():
     assert ret < 0.15
 
 
+def test_cvbooster_save_load(tmp_path):
```
StrikerRUS, line 1076: Can we also check that `best_iteration` attribute was saved and loaded correctly?

StrikerRUS (commented):
Thanks a lot for this PR!
Please consider checking some my comments below.

python-package/lightgbm/engine.py
jameslamb: +1 I think this line should be removed. Since the methods introduced in this PR are instance methods and not class methods.

jameslamb (dismissed):
Thanks very much for this! I agree with all of @StrikerRUS 's suggestions, so please address those before we can merge this.

I have no further comments right now.

StrikerRUS:
Kindly ping @nyanp

jameslamb:
@nyanp are you interested in continuing this pull request? This is a really useful addition to the library and we're very grateful for your work so far. I'd love to see this completed.

nyanp:
@jameslamb 
Sorry for the late reply, I have been off for a while. I will resume today!

python-package/lightgbm/engine.py
nyanp: @StrikerRUS 
Currently cvbooster does not override `dump_model`; it returns a list of JSON dictionaries, which I think is consistent with user expectations.

tests/python_package_test/test_engine.py
nyanp: @StrikerRUS 
I’ve added the test in 26365003532865bdf751684f7beb06b4074279e2.

python-package/lightgbm/engine.py
nyanp: The statement has been removed in 
[cf52d45](https://github.com/microsoft/LightGBM/pull/5160/commits/cf52d45098621771de0f6d5bed3c4fc835af4d3a).

nyanp:
I believe all review content has now been resolved.
Please re-review @StrikerRUS @jameslamb

python-package/lightgbm/engine.py
StrikerRUS: Agree, thanks!

python-package/lightgbm/engine.py
```diff
@@ -305,6 +336,90 @@ def handler_function(*args: Any, **kwargs: Any) -> List[Any]:
...
+        return json.dumps(self._to_dict(num_iteration, start_iteration, importance_type))
+
+    def save_model(
+        self,
+        filename: Union[str, Path],
+        num_iteration: Optional[int] = None,
+        start_iteration: int = 0,
+        importance_type: str = 'split'
+    ) -> "CVBooster":
+        """Save CVBoosters to a file as JSON text.
```
StrikerRUS, line 396:
```suggestion
        """Save CVBooster to a file as JSON text.
```

tests/python_package_test/test_engine.py
```diff
@@ -1073,6 +1074,66 @@ def test_cvbooster():
...
+    X, y = load_breast_cancer(return_X_y=True)
+    X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.1, random_state=42)
```
StrikerRUS, line 1079:
```suggestion
    X_train, X_test, y_train, _ = train_test_split(X, y, test_size=0.1, random_state=42)
```

tests/python_package_test/test_engine.py
```diff
@@ -1073,6 +1074,66 @@ def test_cvbooster():
...
+    cvbooster_from_string = lgb.CVBooster().model_from_string(model_string)
+    for cvbooster_loaded in [cvbooster_from_txt_file, cvbooster_from_string]:
+        assert best_iteration == cvbooster_loaded.best_iteration
+        np.testing.assert_array_equal(preds, cvbooster_loaded.predict(X_test))
+
+
+@pytest.mark.parametrize('serializer', SERIALIZERS)
+def test_cvbooster_picklable(serializer):
+    X, y = load_breast_cancer(return_X_y=True)
+    X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.1, random_state=42)
```
StrikerRUS, line 1113:
```suggestion
    X_train, X_test, y_train, _ = train_test_split(X, y, test_size=0.1, random_state=42)
```

tests/python_package_test/test_engine.py
```diff
@@ -1073,6 +1074,66 @@ def test_cvbooster():
...
+                    num_boost_round=10,
+                    nfold=nfold,
+                    callbacks=[lgb.early_stopping(stopping_rounds=5)],
+                    return_cvbooster=True)
+    cvbooster = cv_res['cvbooster']
+    preds = cvbooster.predict(X_test)
+
+    cvbooster_from_disk = pickle_and_unpickle_object(obj=cvbooster, serializer=serializer)
+    del cvbooster
+
```
StrikerRUS, line 1132:
```suggestion

    assert best_iteration == cvbooster_from_disk.best_iteration

```

tests/python_package_test/test_engine.py
```diff
@@ -1073,6 +1074,66 @@ def test_cvbooster():
...
+    lgb_train = lgb.Dataset(X_train, y_train)
+
+    cv_res = lgb.cv(params, lgb_train,
+                    num_boost_round=10,
+                    nfold=nfold,
+                    callbacks=[lgb.early_stopping(stopping_rounds=5)],
+                    return_cvbooster=True)
+    cvbooster = cv_res['cvbooster']
+    preds = cvbooster.predict(X_test)
+
```
StrikerRUS, line 1129:
```suggestion
best_iteration = cvbooster.best_iteration

```

StrikerRUS (approved):
Great job, thank you!

LGTM!

Just some minor comments below.

tests/python_package_test/test_engine.py
```diff
@@ -1073,6 +1074,69 @@ def test_cvbooster():
...
+    lgb_train = lgb.Dataset(X_train, y_train)
+
+    cv_res = lgb.cv(params, lgb_train,
+                    num_boost_round=10,
+                    nfold=nfold,
+                    callbacks=[lgb.early_stopping(stopping_rounds=5)],
+                    return_cvbooster=True)
+    cvbooster = cv_res['cvbooster']
+    preds = cvbooster.predict(X_test)
+best_iteration = cvbooster.best_iteration
```
StrikerRUS, line 1129: Sorry, bad indent in my previous suggestion.
```suggestion
    best_iteration = cvbooster.best_iteration
```

jameslamb (approved):
I just re-reviewed. All of my and other reviewers' comments have been addressed, and I don't have any other suggestions.

EXCELLENT work @nyanp 🎉 🎉 🎉 

We really appreciate your hard work on this very impactful feature, and your patience with the review process. I hope you'll consider contributing to LightGBM more in the future.

nyanp:
> I just re-reviewed. All of my and other reviewers' comments have been addressed, and I don't have any other suggestions.
> 
> EXCELLENT work @nyanp 🎉 🎉 🎉
> 
> We really appreciate your hard work on this very impactful feature, and your patience with the review process. I hope you'll consider contributing to LightGBM more in the future.

@jameslamb 
Thank you very much for the kind and polite review! I am very happy to contribute to my favorite library. I have enjoyed this long journey ;)

PR #5157, opened 2022-04-17
Fix number of features mismatching in continual training (fix #5156)

shiyu1994:
This is to fix #5156.

PR #5164, opened 2022-04-21
Clear split info buffer in cost efficient gradient boosting before every iteration (fix partially #3679)

### LINKED ISSUE #4969: Count number of leaves in the final ensemble
## Summary

During the training, the model builds a decision tree regressor with a different number of leaves.

The question here is that each tree has how many leaves?

## Motivation
To compare the lightGBM model size with different models such as XGBoost

We can check the size of the model

## Description

The model already has the dump method, is there any feature to return the leaves from the dumped model?

shiyu1994:
This is to fix the case provided by @mshivers in #3679. It turns out that the bug is not related to the core algorithm of LightGBM, but due to the cost-efficient gradient boosting module.

In the code snippet below, the best split of `SerialTreeLearner` can be updated by `CostEfficientGradientBoosting`,
https://github.com/microsoft/LightGBM/blob/fc0c8fd4521ab4814fe3f59999115b70984e76c1/src/treelearner/cost_effective_gradient_boosting.hpp#L85-L110

However, `CostEfficientGradientBoosting` did not clear its buffer of splits `splits_per_leaf_` before a new boosting iteration starts, which causes it to contain splits from previous trees. And these splits with wrong split information (sum of gradients and hessians, gain, etc.) will be mixed into the current tree.

### DIFF
--- a/src/treelearner/cost_effective_gradient_boosting.hpp
+++ b/src/treelearner/cost_effective_gradient_boosting.hpp
@@ -10,6 +10,7 @@
 #include <LightGBM/dataset.h>
 #include <LightGBM/utils/common.h>
 #include <LightGBM/utils/log.h>
+#include <LightGBM/utils/threading.h>
 
 #include <vector>
 
@@ -67,12 +68,12 @@ class CostEfficientGradientBoosting {
 
   void BeforeTrain() {
     // clear the splits in splits_per_leaf_
-    const int num_total_splits = static_cast<int>(splits_per_leaf_.size());
-    const int num_threads = OMP_NUM_THREADS();
-    #pragma omp parallel for schedule(static) num_threads(num_threads)
-    for (int i = 0; i < num_total_splits; ++i) {
-      splits_per_leaf_[i].Reset();
-    }
+    Threading::For<size_t>(0, splits_per_leaf_.size(), 1024,
+      [this] (int /*thread_index*/, size_t start, size_t end) {
+      for (size_t i = start; i < end; ++i) {
+        splits_per_leaf_[i].Reset();
+      }
+    });
   }
 
   double DeltaGain(int feature_index, int real_fidx, int leaf_index,
--- a/src/treelearner/cost_effective_gradient_boosting.hpp
+++ b/src/treelearner/cost_effective_gradient_boosting.hpp
@@ -32,6 +32,7 @@ class CostEfficientGradientBoosting {
       return true;
     }
   }
+
   void Init() {
     auto train_data = tree_learner_->train_data_;
     if (!init_) {
@@ -63,6 +64,17 @@ class CostEfficientGradientBoosting {
     }
     init_ = true;
   }
+
+  void BeforeTrain() {
+    // clear the splits in splits_per_leaf_
+    const int num_total_splits = static_cast<int>(splits_per_leaf_.size());
+    const int num_threads = OMP_NUM_THREADS();
+    #pragma omp parallel for schedule(static) num_threads(num_threads)
+    for (int i = 0; i < num_total_splits; ++i) {
+      splits_per_leaf_[i].Reset();
+    }
+  }
+
   double DetlaGain(int feature_index, int real_fidx, int leaf_index,
                    int num_data_in_leaf, SplitInfo split_info) {
     auto config = tree_learner_->config_;
@@ -82,6 +94,7 @@ class CostEfficientGradientBoosting {
                      feature_index] = split_info;
     return delta;
   }
+
   void UpdateLeafBestSplits(Tree* tree, int best_leaf,
                             const SplitInfo* best_split_info,
                             std::vector<SplitInfo>* best_split_per_leaf) {
--- a/src/treelearner/cost_effective_gradient_boosting.hpp
+++ b/src/treelearner/cost_effective_gradient_boosting.hpp
@@ -75,7 +75,7 @@ class CostEfficientGradientBoosting {
     }
   }
 
-  double DetlaGain(int feature_index, int real_fidx, int leaf_index,
+  double DeltaGain(int feature_index, int real_fidx, int leaf_index,
                    int num_data_in_leaf, SplitInfo split_info) {
     auto config = tree_learner_->config_;
     double delta =
--- a/tests/python_package_test/test_basic.py
+++ b/tests/python_package_test/test_basic.py
@@ -430,6 +430,52 @@ def test_cegb_scaling_equalities(tmp_path):
         assert p1txt == p2txt
 
 
+def test_cegb_split_buffer_clean():
+    # modified from https://github.com/microsoft/LightGBM/issues/3679#issuecomment-938652811
+    # and https://github.com/microsoft/LightGBM/pull/5087
+    # test that the ``splits_per_leaf_`` of CEGB is cleaned before training a new tree
+    # which is done in the fix #5164
+    # without the fix:
+    #    Check failed: (best_split_info.left_count) > (0)
+
+    R, C = 1000, 100
+    seed = 29
+    np.random.seed(seed)
+    data = np.random.randn(R, C)
+    for i in range(1, C):
+        data[i] += data[0] * np.random.randn()
+
+    N = int(0.8 * len(data))
+    train_data = data[:N]
+    test_data = data[N:]
+    train_y = np.sum(train_data, axis=1)
+    test_y = np.sum(test_data, axis=1)
+
+    train = lgb.Dataset(train_data, train_y, free_raw_data=True)
+    test = lgb.Dataset(test_data, test_y, free_raw_data=True, reference=train)
+
+    # The test is run twice, on cpu and gpu
+    params = {
+        'device': "cpu",
+        'boosting_type': 'gbdt',
+        'objective': 'regression',
+        'max_bin': 255,
+        'num_leaves': 31,
+        'seed': 0,
+        'learning_rate': 0.1,
+        'min_data_in_leaf': 0,
+        'verbose': 2,
+        'min_split_gain': 1000.0,
+        'cegb_penalty_feature_coupled': 5 * np.arange(C),
+        'cegb_penalty_split': 0.0002,
+        'cegb_tradeoff': 10.0,
+        'num_threads': 16,
+        'force_col_wise': True,
+    }
+
+    lgb.train(params, train, num_boost_round=20, valid_sets=test)
+
+
 def test_consistent_state_for_dataset_fields():
 
     def check_asserts(data):
--- a/tests/python_package_test/test_basic.py
+++ b/tests/python_package_test/test_basic.py
@@ -430,52 +430,6 @@ def test_cegb_scaling_equalities(tmp_path):
         assert p1txt == p2txt
 
 
-def test_cegb_split_buffer_clean():
-    # modified from https://github.com/microsoft/LightGBM/issues/3679#issuecomment-938652811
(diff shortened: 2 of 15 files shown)


jameslamb (changes requested):
Did you accidentally tag the wrong issue in the description? This doesn't seem related to #4969 at all.

Also, is it possible to create a test that reproduces the bug this PR addresses, to ensure this fix is working and prevent it from being accidentally re-introduced in the future?

shiyu1994:
> Did you accidentally tag the wrong issue in the description? This doesn't seem related to https://github.com/microsoft/LightGBM/issues/4969 at all.

Sorry, it should be #4946. Corrected.

bluesummerv:
n1 = 800000
n2 = 2200
x = np.random.random((n1,n2))
y= np.random.random((n1))
model = lgb.LGBMRegressor(num_leaves=100, n_estimators=20, device='gpu')
model.fix(x,y)

Segmentation fault (core dumped)

shiyu1994:
@bluesummerv Thanks for using LightGBM. It seems that your example is not related with this PR. Could you please open a new issue for your bug report?

shiyu1994:
> Sorry, it should be https://github.com/microsoft/LightGBM/issues/4946. Corrected.

More accurately, this PR only fixes the case provided by @mshivers in #3679.

shiyu1994:
> Also, is it possible to create a test that reproduces the bug this PR addresses, to ensure this fix is working and prevent it from being accidentally re-introduced in the future?

@jameslamb Thanks for the suggestion. But TBH I don't think we need a test case for this fix. It just fixes a logical mistake in the code of CEGB, which is not a corner case that is likely to be introduced again by future modification. I think it make sense to add test cases to guarantee the code to work in some tricky scenarios. But perhaps it is not necessary to add a test case for obvious logical mistakes in programming. WDYT?

jameslamb:
> WDYT?

I believe a test should be added in this PR.

I strongly believe that PRs which fix bugs in software should also introduce tests confirming that the software no longer exhibits those bugs, unless the effort involved in creating the tests or the cost of running the tests is too large.

Given that we have already have a clear reproducible example of a bug in LightGBM (https://github.com/microsoft/LightGBM/issues/3679#issuecomment-938652811), the effort required to convert it to a test case should not be too much, and I think it is worth it for the increased confidence that this bug won't be reintroduced. That test would also give us some additional code coverage of using CEGB, which isn't tested thoroughly in the project today.

In my opinion, tests should not just be used for "corner cases" or "tricky scenarios". Tests should try to cover the ways that users might use the software, and check that the software behaves as those uses would expect it to. Testing gives us confidence that once a bug is fixed, it won't be reintroduced.

This project is very large and there are many, many possible code paths based on e.g. different combinations of training parameters, compilation options, data characteristics, operating systems, etc. That set of combinations is far too large for us to just rely on maintainers' knowledge and pull request reviews to catch regressions.

shiyu1994:
@jameslamb Done with adding the test case. Please check.

src/treelearner/cost_effective_gradient_boosting.hpp
```diff
@@ -63,6 +64,17 @@ class CostEfficientGradientBoosting {
     }
     init_ = true;
   }
+
+  void BeforeTrain() {
+    // clear the splits in splits_per_leaf_
+    const int num_total_splits = static_cast<int>(splits_per_leaf_.size());
```
StrikerRUS, line 70: I believe this can overflow on wide datasets:
https://github.com/microsoft/LightGBM/blob/3d25e373ac09fa0cd5ce75a9decbc195e0bd5220/src/treelearner/cost_effective_gradient_boosting.hpp#L38-L40
https://github.com/microsoft/LightGBM/blob/3d25e373ac09fa0cd5ce75a9decbc195e0bd5220/include/LightGBM/config.h#L181-L183
More precisely, on datasets with `16 384` features and maximum allowed `num_leaves` (`131072`) value.
I believe 16k features are not something unreachable in a user scenarios and some autoML/hyperopt software may use maximum allowed value for any parameter.

StrikerRUS:
@guolinke This is very important bugfix for our long living bug. Could you please help to review it?

src/treelearner/cost_effective_gradient_boosting.hpp
shiyu1994: Thanks for spotting this! The size is cast into signed int because omp does not support unsigned int as index in for loop. 
I've switched to `Threading::For` to avoid this.

shiyu1994: Resolved via 45de0a4

shiyu1994:
> bugfix for our long living bug

Note that this **only** fixes the case when CEGB is used (that's why the word *partially* is used in the title of this PR). I need further verification to see if the same bug (Check failed: (best_split_info.left_count) > (0)) still happens in normal cases.

src/treelearner/cost_effective_gradient_boosting.hpp
StrikerRUS: Thanks a lot for doing this!

StrikerRUS:
> Note that this **only** fixes the case when CEGB is used (that's why the word *partially* is used in the title of this PR).

Thanks for the clarification!

However, GitHub autoclosing mechanism doesn't understand this and will close #3679 after merging this PR.
https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword
I propose changing the title to "fix partially #3679" to not close that issue automatically.

tests/python_package_test/test_basic.py
```diff
@@ -430,6 +430,52 @@ def test_cegb_scaling_equalities(tmp_path):
         assert p1txt == p2txt
 
 
+def test_cegb_split_buffer_clean():
```
StrikerRUS, line 433: Please move this test to `test_engine.py` file because `lgb.train()` function that is defined in `engine.py` is used in this tests.

tests/python_package_test/test_basic.py
```diff
@@ -430,6 +430,52 @@ def test_cegb_scaling_equalities(tmp_path):
...
+    params = {
+        'device': "cpu",
+        'boosting_type': 'gbdt',
+        'objective': 'regression',
+        'max_bin': 255,
+        'num_leaves': 31,
+        'seed': 0,
+        'learning_rate': 0.1,
+        'min_data_in_leaf': 0,
+        'verbose': 2,
```
StrikerRUS, line 467: Refer to https://github.com/microsoft/LightGBM/pull/5157#discussion_r868528652.
```suggestion
        'verbose': -1,
```

tests/python_package_test/test_basic.py
StrikerRUS, line 433: Add some asserts. Refer to https://github.com/microsoft/LightGBM/pull/5157#discussion_r868526686.

tests/python_package_test/test_basic.py
```diff
@@ -430,6 +430,52 @@ def test_cegb_scaling_equalities(tmp_path):
...
+        'min_split_gain': 1000.0,
+        'cegb_penalty_feature_coupled': 5 * np.arange(C),
+        'cegb_penalty_split': 0.0002,
+        'cegb_tradeoff': 10.0,
+        'num_threads': 16,
+        'force_col_wise': True,
+    }
+
+    lgb.train(params, train, num_boost_round=20, valid_sets=test)
```
StrikerRUS, line 476: Either check that validation scores contain expected values, or remove unused test dataset creation.

shiyu1994: Resolved via a04d004

tests/python_package_test/test_basic.py
shiyu1994: Resolved via a04d004

tests/python_package_test/test_basic.py
shiyu1994: Resolved via a04d004

tests/python_package_test/test_basic.py
shiyu1994: Resolved via a04d004

tests/python_package_test/test_engine.py
```diff
@@ -3566,3 +3566,50 @@ def test_boost_from_average_with_single_leaf_trees():
...
+        'num_leaves': 31,
+        'seed': 0,
+        'learning_rate': 0.1,
+        'min_data_in_leaf': 0,
+        'verbose': -1,
+        'min_split_gain': 1000.0,
+        'cegb_penalty_feature_coupled': 5 * np.arange(C),
+        'cegb_penalty_split': 0.0002,
+        'cegb_tradeoff': 10.0,
+        'num_threads': 16,
```
StrikerRUS, line 3608: The majority of CI workers have around 2 virtual CPU cores. I don't think setting `num_threads` to `16` (or any other non-default value) is reasonable.

tests/python_package_test/test_engine.py
```diff
@@ -3566,3 +3566,50 @@ def test_boost_from_average_with_single_leaf_trees():
...
+        'force_col_wise': True,
+    }
+
+    model = lgb.train(params, train, num_boost_round=10)
+    predicts = model.predict(test_data)
+    rmse = np.sqrt(np.mean((predicts - test_y) ** 2))
```
StrikerRUS, line 3614: Can we reuse `mean_squared_error` for the simplicity?
https://github.com/microsoft/LightGBM/blob/0018206742544cfd0b2c1e156f177ab5f468a998/tests/python_package_test/test_engine.py#L17

tests/python_package_test/test_engine.py
shiyu1994: Resolved via [f8e8170](https://github.com/microsoft/LightGBM/pull/5164/commits/f8e81700e6e4425d9cd9145b1173217855fd8d76)

tests/python_package_test/test_engine.py
shiyu1994: Resolved via [f8e8170](https://github.com/microsoft/LightGBM/pull/5164/commits/f8e81700e6e4425d9cd9145b1173217855fd8d76)

tests/python_package_test/test_engine.py
```diff
@@ -3566,3 +3566,49 @@ def test_boost_from_average_with_single_leaf_trees():
...
+    N = int(0.8 * len(data))
+    train_data = data[:N]
+    test_data = data[N:]
+    train_y = np.sum(train_data, axis=1)
+    test_y = np.sum(test_data, axis=1)
+
+    train = lgb.Dataset(train_data, train_y, free_raw_data=True)
+
+    params = {
+        'device': "cpu",
```
StrikerRUS, line 3595: My last concern regarding this test. Can we remove `device` parameter to run this test against both cpu and gpu devices at our CI? Or the fix in this PR is CPU-exclusive and doesn't fix GPU code?

shiyu1994: Thanks. That should be removed. Resoved via [5235131](https://github.com/microsoft/LightGBM/pull/5164/commits/5235131e90539115206f6e168863ddbec7bdb007)

StrikerRUS (approved):
LGTM, thanks!

guolinke (approved):

shiyu1994:
@StrikerRUS 
> LGTM, thanks!

Could you please remove the change requests so that we can merge this RP? Currently it is blocked by the change requests. Thanks!

jameslamb (approved):
> Could you please remove the change requests

Looks like that's from my initial review.

image

I've removed it here. Thanks for the fix!

PR #5167, opened 2022-04-22
[CUDA] Fix integer overflow in cuda row-wise data

shiyu1994:
This is to fix integer overflow in `cuda_exp`. When data index is multiplied with the number of features, integer overflow can easily occur if 32-bit integers are used. In this PR, we cast the indices into 64-bit integers whenever necessary.

### DIFF
--- a/src/io/cuda/cuda_row_data.cpp
+++ b/src/io/cuda/cuda_row_data.cpp
@@ -318,15 +318,15 @@ void CUDARowData::GetDenseDataPartitioned(const BIN_TYPE* row_wise_data, std::ve
     [this, num_total_columns, row_wise_data, out_data] (int /*thread_index*/, data_size_t start, data_size_t end) {
       for (size_t i = 0; i < feature_partition_column_index_offsets_.size() - 1; ++i) {
         const int num_prev_columns = static_cast<int>(feature_partition_column_index_offsets_[i]);
-        const data_size_t offset = num_data_ * num_prev_columns;
+        const size_t offset = static_cast<size_t>(num_data_) * static_cast<size_t>(num_prev_columns);
         const int partition_column_start = feature_partition_column_index_offsets_[i];
         const int partition_column_end = feature_partition_column_index_offsets_[i + 1];
         const int num_columns_in_cur_partition = partition_column_end - partition_column_start;
         for (data_size_t data_index = start; data_index < end; ++data_index) {
-          const data_size_t data_offset = offset + data_index * num_columns_in_cur_partition;
-          const data_size_t read_data_offset = data_index * num_total_columns;
+          const size_t data_offset = offset + data_index * num_columns_in_cur_partition;
+          const size_t read_data_offset = static_cast<size_t>(data_index) * num_total_columns;
           for (int column_index = 0; column_index < num_columns_in_cur_partition; ++column_index) {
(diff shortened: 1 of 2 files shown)


SiNZeRo:
@shiyu1994 awesome!

guolinke (approved):

jameslamb (commented):
@shiyu1994 is it possible to add a test for this, that would fail on current `master`  and pass as of these changes?

I feel that every bug fix should be accompanied by a test if possible, to prevent accidentally re-introducing that bug in the future.

jameslamb:
Also @shiyu1994 please remember that every pull request needs to be labeled with one of the labels mentioned in https://github.com/microsoft/LightGBM/blob/fc0c8fd4521ab4814fe3f59999115b70984e76c1/.github/release-drafter.yml, so it'll be correctly classified in release notes.

shiyu1994:
> @shiyu1994 is it possible to add a test for this, that would fail on current master and pass as of these changes?

I think it would be difficult to include such test cases in ci. Because that would require generating a dataset with #data x #feature greater than the range of 32-bit integers, which is quite large.

shiyu1994:
> Also @shiyu1994 please remember that every pull request needs to be labeled with one of the labels

Thanks for the reminder! Will tag every PR from now on.

jameslamb:
> I think it would be difficult to include such test cases in ci.

alright, no problem

shiyu1994:
The job `R-package / r-package (macOS-latest, clang, R 4.1, cran) (pull_request)`
```
==> Downloading https://mirror.ctan.org/systems/mac/mactex/mactex-basictex-20220314.pkg
[137](https://github.com/microsoft/LightGBM/runs/6143699319?check_suite_focus=true#step:5:137)
==> Downloading from https://ctan.math.washington.edu/tex-archive/systems/mac/mactex/mactex-basictex-20220314.pkg
[138](https://github.com/microsoft/LightGBM/runs/6143699319?check_suite_focus=true#step:5:138)
curl: (60) SSL certificate problem: certificate has expired
[139](https://github.com/microsoft/LightGBM/runs/6143699319?check_suite_focus=true#step:5:139)
More details here: https://curl.haxx.se/docs/sslcerts.html
[140](https://github.com/microsoft/LightGBM/runs/6143699319?check_suite_focus=true#step:5:140)

[141](https://github.com/microsoft/LightGBM/runs/6143699319?check_suite_focus=true#step:5:141)
curl failed to verify the legitimacy of the server and therefore could not
[142](https://github.com/microsoft/LightGBM/runs/6143699319?check_suite_focus=true#step:5:142)
establish a secure connection to it. To learn more about this situation and
[143](https://github.com/microsoft/LightGBM/runs/6143699319?check_suite_focus=true#step:5:143)
how to fix it, please visit the web page mentioned above.
[144](https://github.com/microsoft/LightGBM/runs/6143699319?check_suite_focus=true#step:5:144)
Error: Download failed on Cask 'basictex' with message: Download failed: https://mirror.ctan.org/systems/mac/mactex/mactex-basictex-20220314.pkg
[145](https://github.com/microsoft/LightGBM/runs/6143699319?check_suite_focus=true#step:5:145)
```
fails again with the above message.

Close and reopen to rerun the tests.

shiyu1994:
It is weird that the same error still happens. I've check the link for download in the error message and it works well.

PR #5173, opened 2022-04-23
[tests] replace `fobj` with `custom objective` in test comments and make tests stricter

StrikerRUS:
Linking #5052.
More strict tests can help to spot some bugs or inconsistencies between different LightGBM versions like https://github.com/microsoft/LightGBM/pull/4630#discussion_r825490038.


jameslamb (approved):
totally agree, thanks very much for this!

PR #5169, opened 2022-04-22
[ci] fix git checkout for comment-triggered CI jobs

jameslamb:
See https://github.com/microsoft/LightGBM/pull/5159#issuecomment-1105981238.

A patch was recently pushed in `git` which fixes a newly-discovered security vulnerability. That patch causes the `actions/checkout` GitHub Action to fail for containerized jobs. See #5151 for details.

#5152 sought to fix that for LightGBM's CI jobs, but missed optional jobs triggered by comments.

This PR applies that fix to those jobs.

## Note for Reviewers

Since the configs for these comment-triggered jobs are sourced from `master`, I think this PR will have to be merged and then we'll have to use comments on another open PR to test.

Issue #5106, opened 2022-03-30
Python package looks for library in wrong path


david-cortes:
Trying to build the python package from source without installing it will somehow try to pick the wrong path for system libraries. I get an error about being unable to import `scipy.sparse`, even though I can import that library in the same session (this is after successfully building `lib_lightgbm` through the cmake system):
```python
>>> import scipy.sparse
>>> import lightgbm
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/david/del/py_lightgbm/python-package/lightgbm/__init__.py", line 8, in <module>
    from .basic import Booster, Dataset, Sequence, register_logger
  File "/home/david/del/py_lightgbm/python-package/lightgbm/basic.py", line 126, in <module>
    _LIB = _load_lib()
  File "/home/david/del/py_lightgbm/python-package/lightgbm/basic.py", line 117, in _load_lib
    lib = ctypes.cdll.LoadLibrary(lib_path[0])
  File "/home/david/anaconda3/envs/py3/lib/python3.9/ctypes/__init__.py", line 460, in LoadLibrary
    return self._dlltype(name)
  File "/home/david/anaconda3/envs/py3/lib/python3.9/ctypes/__init__.py", line 382, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: /home/david/anaconda3/envs/py3/lib/python3.9/site-packages/scipy/sparse/../../../../libstdc++.so.6: version `GLIBCXX_3.4.30' not found (required by /home/david/del/py_lightgbm/lib_lightgbm.so)

```

jmoralez:
Hi @david-cortes, thank you for your report. Can you provide some more details on how you installed the package? I've had this happened to me when I installed `lightgbm` from conda and then from pip and somehow the dlls get mixed. If you've already compiled the package yourself you can try `python setup.py install --precompile` which uses the produced artifacts. If you can provide some more details on your installation procedure we might be able to help a bit more.

david-cortes:
I didn't install it, I compiled it from source from the current master branch using the cmake system and tried to import it in python from its folder. Doing that imports it correctly in other setups.

jmoralez:
I see. Did you have it installed already in your environment? I just tried:
```bash
conda create -n lgb_test -y python=3.9 scipy
conda activate lgb_test
python -c 'import lightgbm'
```
and it works correctly.

jmoralez:
[...]

PR #5174, opened 2022-04-23
[ci] temporarily pin Azure Devops jobs to ubuntu 20.04

jameslamb:
All `Linux_latest` jobs on Azure Devops started failing a few days ago, because the `ubuntu:latest` tag on DockerHub moved from Ubuntu 20.04 to 22.04.

https://github.com/microsoft/LightGBM/pull/5169#issuecomment-1106413298
https://github.com/microsoft/LightGBM/pull/5169#issuecomment-1107139143

While attempting to fix this on #5169, I ran into the issue from #5106 (https://github.com/microsoft/LightGBM/issues/5106#issuecomment-1107346250). I believe I've identified the root cause there, but I expect fixing it to require significant discussion and effort.

For now, to unblock other development in this project, this PR proposes **temporarily** pinning the `Linux_latest` jobs back to Ubuntu 20.04.

### DIFF
--- a/.vsts-ci.yml
+++ b/.vsts-ci.yml
@@ -16,7 +16,7 @@ resources:
   - container: ubuntu1404
     image: lightgbm/vsts-agent:ubuntu-14.04
   - container: ubuntu-latest
-    image: 'ubuntu:latest'
+    image: 'ubuntu:20.04'
     options: "--name ci-container -v /usr/bin/docker:/tmp/docker:ro"
   - container: rbase
     image: wch1/r-debug


StrikerRUS (approved):
Agree! Thanks for working on this!
Should we also pin Ubuntu version for GitHub Actions (they share the same environments with Azure Pipelines)?

jameslamb:
I don't think we need to yet.

On Azure DevOps, these jobs are running inside an `ubuntu:latest` docker container.

On GitHub Actions, they run on the `ubuntu-latest` VM image (not inside a container), which is still on 20.04 according to https://github.com/actions/virtual-environments.

I think that's why we've only seen job failures on Azure DevOps.

PR #5183, opened 2022-04-27
[c-api] check number of features when retrieving number of bins

jmoralez:
When retrieving the number of bins for a given feature the input wasn't checked so if we input an invalid feature index (>= number of features) we could get garbage values or even a segfault, so this adds a check to see if we're in the valid range of feature indices.


StrikerRUS (approved):
LGTM, thanks!

guolinke (approved):

Issue #5051, opened 2022-03-03
[RFC] Inconsistent behavior when a single-leaf tree is encountered


shiyu1994:
## Description
With a single-leaf tree, CLI version stops training at once. But Python API continues to train.

## Reproducible example
Example by @arnocandel in #4708.

## Additional Comments
We have two choices:
1. Stopped training at once when a single leaf tree is encountered in all APIs.
2. Continue to train, add the same prediction value to all training samples. The prediction value should be `- sum_of_gradients/sum_of_hessians` in the root node, which is now not calculated.

Gently ping @guolinke @jameslamb @StrikerRUS @hzy46 @btrotta for your opinion.

guolinke:
I think option 2 is better.

jameslamb:
Thanks for writing this up and for the description in https://github.com/microsoft/LightGBM/issues/4708#issuecomment-1058276074!

I think option 2 is preferable.

Encountering a single-leaf tree doesn't necessarily mean that training should stop. I'd expect future boosting rounds could still further informative splits in some situations, like:

* using [`bagging_fraction`](https://lightgbm.readthedocs.io/en/latest/Parameters.html#bagging_fraction) to re-sample rows
* using [`feature_fraction`](https://lightgbm.readthedocs.io/en/latest/Parameters.html#feature_fraction) to randomly choose features
* using a custom objective that has some sort of randomness in it or some behavior dependent on the number of iterations

shiyu1994:
@jameslamb Thanks, then let's go to option 2.

PR #5193, opened 2022-05-02
[python-package] Use update() finish condition on booster loop

Samsagax:
Currently the boosting loop in python does not stop if the booster says it is finished. This patch captures the output of the booster.update() call and if it is finished it breaks the loop at the end of current iteration.

### DIFF
--- a/python-package/lightgbm/engine.py
+++ b/python-package/lightgbm/engine.py
@@ -224,6 +224,7 @@ def train(
             valid_set._reverse_update_params()
     booster.best_iteration = 0
 
+    is_finished = False
     # start training
     for i in range(init_iteration, init_iteration + num_boost_round):
         for cb in callbacks_before_iter:
@@ -234,7 +235,7 @@ def train(
                                     end_iteration=init_iteration + num_boost_round,
                                     evaluation_result_list=None))
 
-        booster.update(fobj=fobj)
+        is_finished = booster.update(fobj=fobj)
 
         evaluation_result_list = []
         # check evaluation result.
@@ -253,6 +254,8 @@ def train(
         except callback.EarlyStopException as earlyStopException:
             booster.best_iteration = earlyStopException.best_iteration + 1
             evaluation_result_list = earlyStopException.best_score
+            break
+        if is_finished:
             break
     booster.best_score = collections.defaultdict(collections.OrderedDict)
     for dataset_name, eval_name, score, _ in evaluation_result_list:


StrikerRUS:
Hey @Samsagax , thanks for your interest in LightGBM!

`is_finished` indicates that **only one current iteration** has been finished successfully. You patch makes booster to stop after the first iteration which is wrong behavior.

Samsagax:
Strange. Works for me with this change. If i don't apply this, then when a booster trains and finds no splittable features I get a spam of warnings and a constant output metric.

The warning says "Stopped training because there are no more leaves that meet the split requirements"

Looking up the code I find I'm hitting this line https://github.com/microsoft/LightGBM/blob/da9072fde2b5cd7160866e7e9e49a14e7fba8bf3/src/boosting/gbdt.cpp#L438
If I understood correctly that if will return `true` to the caller when training stops. Following the python code I see the C++ function `TrainOneIter()` is called by `booster.update()` python function and that is passed to the loop. Maybe is specific to GBDT booster, anyway.

Thanks for having the time to review the patch.

StrikerRUS:
> Works for me with this change. 

This may mean that due to incorrect parameters your training actually builds only one tree and that's why there's no difference for your case.

This mechanism shouldn't work as an early stopping. That warning should help to understand that something is wrong in users' setup. Refer to #4178 for more details.

jameslamb:
I just want to add one more comment, for those finding this PR from searching that warning.

It's related to https://github.com/microsoft/LightGBM/issues/5051#issuecomment-1062524473.

Another reason this shouldn't be used as an early stopping mechanism is that "can't find any more splits" in one iteration doesn't mean LightGBM automatically won't be able to in the next iteration. For example, if you're using [`feature_fraction`](https://lightgbm.readthedocs.io/en/latest/Parameters.html#feature_fraction) to randomly choose a subset of features at each iteration, then it's possible that even if no informative splits are found in iteration `i`, some could be in iteration `i + 1`.

Samsagax:
> I just want to add one more comment, for those finding this PR from searching that warning.
> 
> It's related to [#5051 (comment)](https://github.com/microsoft/LightGBM/issues/5051#issuecomment-1062524473).
> 
> Another reason this shouldn't be used as an early stopping mechanism is that "can't find any more splits" in one iteration doesn't mean LightGBM automatically won't be able to in the next iteration. For example, if you're using [`feature_fraction`](https://lightgbm.readthedocs.io/en/latest/Parameters.html#feature_fraction) to randomly choose a subset of features at each iteration, then it's possible that even if no informative splits are found in iteration `i`, some could be in iteration `i + 1`.

Thanks for the clarification and I myself experienced that situation (with feature_fraction < 1). I agree it should continue and this patch is not desired behaviour. I was just replicating the behaviour of the `Train()` method of the CLI version here: https://github.com/microsoft/LightGBM/blob/da9072fde2b5cd7160866e7e9e49a14e7fba8bf3/src/boosting/gbdt.cpp#L266

I still find a discrepancy on how the CLI version works since its implementation of GBDT `Train()` method does stop on the first tree that can't be split (https://github.com/microsoft/LightGBM/blob/da9072fde2b5cd7160866e7e9e49a14e7fba8bf3/src/boosting/gbdt.cpp#L271). Maybe it should not use the return of `TrainOneIter()` as `is_finished` condition to bring both implementations in line?

jameslamb:
> Maybe it should not use the return of TrainOneIter() as is_finished condition to bring both implementations in line?

I think that's exactly what the conclusion of the discussion in #5051 was, yep!

If you'd like to attempt a pull request to implement that change in behavior, please comment on #5051 and see if @shiyu1994 or @guolinke will agree to help review and answer questions.

Thanks for your continued interest in LightGBM!

PR #5197, opened 2022-05-04
[ci] Prevent Python downgrading to pypy on Windows

StrikerRUS:
Fix current `master` failures at Appveyor that happen due to pytest uses `pypy3` as a Python interpreter.
```
The following packages will be DOWNGRADED:
  python                        3.7.12-h900ac77_100_cpython --> 3.7.12-0_73_pypy
  python_abi                                    3.7-2_cp37m --> 3.7-2_pypy37_pp73

...

============================= test session starts =============================
platform win32 -- Python 3.7.12[pypy-7.3.7-final], pytest-7.1.1, pluggy-1.0.0

...
_________________________________ test_basic __________________________________
self = <class 'ctypes.c_char_p'>, value = 1047415529008
    def from_param(self, value):
        if isinstance(value, self):
            return value
        try:
>           as_parameter = value._as_parameter_
E           AttributeError: 'int' object has no attribute '_as_parameter_'
C:\Miniconda3-x64\envs\test-env\lib_pypy\_ctypes\basics.py:56: AttributeError
```
https://ci.appveyor.com/project/guolinke/lightgbm/builds/43430960/job/smunq6b4pcq936b5?fullLog=true

Looks like `libxml2` installation is an universal workaround for any CI error 😄 
Related #4993, #5068, #5126.

PR #5198, opened 2022-05-05
[ci] Prevent Python downgrading to pypy on Windows (2)

StrikerRUS:
Ah, `libxml2` isn't omnipotent as [we thought](https://github.com/microsoft/LightGBM/pull/5197#pullrequestreview-962369155) 🙁 

Appveyor at `master` fails again with the same error as yesterday in #5197 (but only MSVC job this time).
https://ci.appveyor.com/project/guolinke/lightgbm/builds/43451448


jameslamb (approved):
😭 😭 😭

PR #5199, opened 2022-05-07
[R-package] ensure that callbacks respect verbosity from params

jameslamb:
Contributes to #4862.
Fixes an issue introduced by #4899 .

In the R package, `lightgbm()`, `lgb.train()`, and `lgb.cv()` have a keyword argument `verbose`. #4899 added changes to ensure that that verbosity parameters passed through `params` took precedence over that keyword argument.

However, evaluation and and early stopping callbacks still reference that keyword argument, which can lead to inconsistent behavior where some printed messages' verbosity can only be controlled by the keyword argument.

While working on this, I also discovered another closely-related bug...parameter `verbose` in calls to `cb_early_stop()` is being passed an integer, but that value is then checked with `isTRUE()`...which means that no messages are ever printed from that callback.

https://github.com/microsoft/LightGBM/blob/dad391574137958402bec4b417bc594fa1c5fbed/R-package/R/callback.R#L207

This PR proposes changes to ensure that it's possible to silence early stopping messages and evaluation messages using `params`.

### Notes for Reviewers

This PR reduces the number of logs printed by the R package's tests from 2095 lines to 1616 lines.

```shell
R CMD INSTALL --with-keep.source ./lightgbm_3.3.1.99.tar.gz
cd R-package/tests
```

### DIFF
--- a/R-package/R/lightgbm.R
+++ b/R-package/R/lightgbm.R
@@ -161,6 +161,11 @@ lightgbm <- function(data,
     , params = params
     , alternative_kwarg_value = num_threads
   )
+  params <- lgb.check.wrapper_param(
+    main_param_name = "verbosity"
+    , params = params
+    , alternative_kwarg_value = verbose
+  )
 
   # Set data to a temporary variable
   dtrain <- data
@@ -175,7 +180,7 @@ lightgbm <- function(data,
     , "data" = dtrain
     , "nrounds" = nrounds
     , "obj" = objective
-    , "verbose" = verbose
+    , "verbose" = params[["verbosity"]]
     , "eval_freq" = eval_freq
     , "early_stopping_rounds" = early_stopping_rounds
     , "init_model" = init_model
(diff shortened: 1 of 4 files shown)


jmoralez (approved):
Very thorough tests!

R-package/R/lightgbm.R
```diff
@@ -189,7 +194,7 @@ lightgbm <- function(data,
   }
 
   # Set validation as oneself
-  if (verbose > 0L) {
+  if (params[["verbosity"]] > 0L) {
     train_args[["valids"]][["train"]] <- dtrain
   }
```
StrikerRUS, line 199: Quite strange behavior, to be honest...
Maybe remove it in `v4.0.0`?

StrikerRUS (approved):
LGTM! 👍

R-package/R/lightgbm.R
jameslamb: I totally agree! I definitely think we should consider removing this prior to v4.0.0.

This has been in `{lightgbm}` since the very first version of the R package, in #168.

Wrote up a separate issue for discussion of that: #5203.

StrikerRUS: Thanks a lot for doing this!

Issue #5153, opened 2022-04-14
[RFC] 4.0.0 Release


jameslamb:
## Summary

@StrikerRUS @guolinke @shiyu1994 @jmoralez @btrotta @Laurae2 could you please try to list out what you all feel is required before a v4.0.0 release of LightGBM is prepared?

Please edit this description and add issues that you feel should be fixed prior to a v4.0.0 release. I've proposed an initial list below.

**Python package**

* [ ] #4220
* [ ] #4771
* [x] #4788
* [x] #4793
* [x] #5313
* [x] #4518
* [x] #3556
* [x] #5354
* [x] #5027
* [x] #5105
* [x] #4596
* [x] #5052

**R package**

* [ ] #4323
* [ ] #4904
* [ ] #4970
* [ ] #5102
* [ ] #5135
* [x] #4881
* [x] #5036
* [x] #4966
* [x] #4982
* [x] #4813
* [x] #5108
* [x] #5116
* [x] #5122
* [x] #4670

**CUDA**

* [ ] #4781
* [ ] #5062
* [ ] #5076
* [ ] #5086
* [ ] #5163
* [x] #4785

**GPU (non-CUDA)**

* [ ] #3339
* [ ] #3475
* [ ] #3645
* [ ] #3870
* [ ] #4082
* [ ] #4618
* [ ] #4612
* [ ] #4820
* [x] #4497

**Quantized Training**
* [ ] #5606 

**Other**

* [ ] #3234
* [ ] #3679
* [ ] #4156
* [ ] #4286
* [ ] #4331
* [ ] #4405
* [ ] #4414
* [ ] #4739
* [ ] #4780
* [ ] #4946
* [ ] #5093
* [ ] #5124
* [ ] #5091 
* [x] #4827
* [x] #5391
* [x] #5344
* [x] #4649

## Motivation

The most recent release of LightGBM, v3.3.1, was about 6 months ago ([October 27, 2021](https://github.com/microsoft/LightGBM/releases/tag/v3.3.1).

*There was a v3.3.2  release on January 7, 2022, but it just contained a single small patch to satisfy CRAN*

We have been talking for even longer than that about putting out a 4.0.0 release of LightGBM. Many many fixes, features, and breaking changes have been merged since v3.3.1, and the state of this project on `master` is now significantly different from what users will get running `pip install lightgbm`.

[...]

Issue #4919, opened 2021-12-29
[R-package] [docs] instructions for generating code coverage produce incorrect results


jameslamb:
## Description

The R package's README contains a section describing how to generate a code coverage report using `{covr}`.

https://github.com/microsoft/LightGBM/blob/ce486e5b45a6f5e67743e14765ed139ff8d532e5/R-package/README.md#code-coverage

Those instructions run without error and do produce _some_ code coverage, but the coverage results are incorrect.

For example, look at how many files are reported as having 0 coverage.

```text
R/lgb.convert_with_rules.R: 0.00%
R/lgb.cv.R: 0.00%
R/lgb.drop_serialized.R: 0.00%
R/lgb.interprete.R: 0.00%
R/lgb.make_serializable.R: 0.00%
R/lgb.plot.importance.R: 0.00%
R/lgb.plot.interpretation.R: 0.00%
R/lgb.restore_handle.R: 0.00%
R/readRDS.lgb.Booster.R: 0.00%
R/saveRDS.lgb.Booster.R: 0.00
```

Seeing 0.00% coverage for `lgb.cv()`, for example, is very surprising, given that there are around 30 calls to that function in the R package's tests.

https://github.com/microsoft/LightGBM/blob/ce486e5b45a6f5e67743e14765ed139ff8d532e5/R-package/tests/testthat/test_basic.R#L363

It isn't like all coverage results are 0.0, as can sometimes happen with `pytest-cov` for example. Other files do show some test coverage.

```text
R/lgb.importance.R: 96.00%
R/lightgbm.R: 96.43%
R/aliases.R: 100.00%
R/lgb.model.dt.tree.R: 100.00%
R/metrics.R: 100.00%
```

<details><summary>full coverage results (click me)</summary>

```text
lightgbm Coverage: 52.00%
R/lgb.convert_with_rules.R: 0.00%
R/lgb.cv.R: 0.00%
R/lgb.drop_serialized.R: 0.00%
R/lgb.interprete.R: 0.00%
R/lgb.make_serializable.R: 0.00%
R/lgb.plot.importance.R: 0.00%
R/lgb.plot.interpretation.R: 0.00%
R/lgb.restore_handle.R: 0.00%
R/readRDS.lgb.Booster.R: 0.00%
R/saveRDS.lgb.Booster.R: 0.00%
src/boosting/goss.hpp: 0.00%
src/boosting/rf.hpp: 0.00%
[...]

Issue #3445, opened 2020-10-08
[R-package] R session crashes on iris example


OldMortality:
R crashes when I get to lgb.train statement.

This is happening in my project, but I can recreate it in the iris example (attached)

I will attach the code, and a screenshot of the error.
This has worked for me before, I don't know what has changed.

[lightgbm R fatal error.zip](https://github.com/microsoft/LightGBM/files/5351548/lightgbm.R.fatal.error.zip)

The FAQ says I may ping a member with the relevant knowledge, so I will attempt to do this here: @guolinke  @Laurae2

Many thanks for your attention.

I am using Windows 10, R version 4.0.2.
I installed the lightgbm package like this:

```
# PKG_URL <- "https://github.com/microsoft/LightGBM/releases/download/v3.0.0/lightgbm-3.0.0-r-cran.tar.gz"
# 
# remotes::install_url(PKG_URL)
```

jameslamb:
Hi @OldMortality , I can help you with this. Thanks for the repo and for using `LightGBM`!

I've copied your code below, as that makes it a little easier for others to see and improves the relevance of search results in this repo

<details><summary>reproducible example (click me)</summary>

```r
rm(list=ls())
library(lightgbm)

# We load the default iris dataset shipped with R
data(iris)


# We must convert factors to numeric
# They must be starting from number 0 to use multiclass
# For instance: 0, 1, 2, 3, 4, 5...
iris$Species <- as.numeric(as.factor(iris$Species)) - 1L
# prevent R-session from crashing: response must be a factor
# iris$Species <- as.factor(iris$Species)

# We cut the data set into 80% train and 20% validation
# The 10 last samples of each class are for validation



train <- as.matrix(iris[c(1L:40L, 51L:90L, 101L:140L), ])
test <- as.matrix(iris[c(41L:50L, 91L:100L, 141L:150L), ])
dtrain <- lgb.Dataset(data = train[, 1L:4L], label = train[, 5L])
dtest <- lgb.Dataset.create.valid(dtrain, data = test[, 1L:4L], label = test[, 5L])
valids <- list(test = dtest)

# Method 1 of training
params <- list(objective = "multiclass", metric = "multi_error", num_class = 3L)
model <- lgb.train(
  params
  , dtrain
  , 100L
  , valids
  , min_data = 1L
  , learning_rate = 1.0
  , early_stopping_rounds = 10L
)

s# We can predict on test data, outputs a 90-length vector
[...]

PR #5204, opened 2022-05-10
[R-package] remove lgb.unloader()

jameslamb:
Proposes removing `lgb.unloader()` from the R package.

Reasons I support removing this:

* simplifies the public API of the R package, reducing the risk that users spend time investigating a function they don't need
* reduces the run time of the package's tests
* removes some tests that might break in future versions of R or `{testthat}` because of the unorthodox things this function does with environments
    - see, for example, https://github.com/microsoft/LightGBM/issues/4919#issuecomment-1003267201
* I don't think it's used much, and that it's unnecessary as of #3016
    - searching on GitHub (not perfect, I know), the only uses of this function I found were in forks of LightGBM
    - https://github.com/search?q=%22lgb.unloader%22&type=code

### Notes for Reviewers

`lgb.unloader()` was introduced 5+ years ago, in #378. At that time, `{lightgbm}` was using a custom R-to-C interface, which could result in some situations where `Dataset` and `Booster` objects lifecycles' weren't handled in the standard way and users could experience session crashes or freezes when errors occurred (e.g.  #2088, #3445, #4208).

As of the changes tied to #3016 (released in v3.3.0), `{lightgbm}`'s R-to-C interface is now the standard one described in https://cran.r-project.org/doc/manuals/r-release/R-exts.html, and we haven't received reports of such issues.

Now, when the next release will be a `v4.0.0` (#5153) with breaking changes, seems like a good time to make this change.

### How I tested this

In addition to running the package's unit tests, I also ran the one affected demo.

```shell
Rscript R-package/demo/weight_param.R
```

That ran without issue.

I also searched for other uses of `lgb.unloader()` like this:

```shell
git grep -E 'lgb\.unload'
```


StrikerRUS (approved):
Great simplification, thank you!

jmoralez (commented):
I agree with this. Just a minor suggestion on the `R-package/demo/weight_param.R` demo, I think we should change the lines a bit after the unloader part (76-84) to something like the following:

```
# We now create the dataset without specifying weights 
# which makes each sample weigh 1 and thus the sum of weights equal to 6513
valids <- list(test = dtest)
```
Which makes me think also line 11 should be:
`# - Run 3: sum of weights equal to 6513 with adjusted regularization (learning)`
or
`# - Run 3: sum of weights equal to 0.06513 (x 1e5) with adjusted regularization (learning)`

jameslamb:
> I think we should change the lines a bit after the unloader part (76-84) to something like the following

Sure, I'd be happy to review a PR with proposed changes to the demos. But I'm going to merge this PR, as I don't think your suggestion is directly related to `lgb.unloader()`.

jmoralez:
They were kind of left over after removing the unloader, since they reload the library and the data

jameslamb:
Ah ok, sorry. I thought that since you left a "comment" instead of "request changes" review, that meant you were just saying something else you noticed and not requesting changes for this specific PR.

jmoralez:
They're not really necessary but make it a bit confusing without the unloader part. I can address that in a follow up PR

jameslamb:
Ah I see ok. Yeah please do, thanks!

PR #5159, opened 2022-04-20
Check existence of inet_pton for win32 in CMakeLists.txt (fixes #5019)

shiyu1994:
This is to fix #5019. We manually check the existence of `inet_pton` in `CMakeLists.txt` to avoid `inet_pton` being redefined in `src/network/socket_wrapper.hpp`.

PR #5210, opened 2022-05-12
[ci] fix R Hub token

jameslamb:
I'm so sorry, @StrikerRUS you were absolutely right in https://github.com/microsoft/LightGBM/pull/5159#issuecomment-1124339616...the way I manually submitted the R package to R Hub yesterday, I overwrite this project's R Hub authentication token.

I had thought that a given email could have multiple auth tokens with R Hub, but I see now that that's not correct. When I tried submitting today with the existing token in `.ci/run_rhub_solaris_checks.R`, `rhub::check()` failed with the following error.

```text
─  Uploading package
Error in query("SUBMIT PACKAGE", data = list(email = unbox(email), token = unbox(email_get_token(email)),  : 
  Email address not validated
```

This PR proposes updating that script with the token from https://github.com/microsoft/LightGBM/pull/5159#issuecomment-1124339616.


StrikerRUS (approved):
Oh, no problem! That wasn't obvious at all from R Hub docs!

Thanks for updating the token.

Issue #5207, opened 2022-05-10
[python] `Dataset` cannot be constructed from sparse `Sequence`


tony-theorem:
## Description
A `Dataset` cannot be constructed from a `Sequence` that returns sparse data. This behavior is unexpected as `Dataset` generally supports sparse data .

## Reproducible example
```
from __future__ import annotations
import numbers
from typing import Iterable

import lightgbm as lgbm
import numpy as np
from scipy import sparse

class SparseSequence(lgbm.Sequence):
    def __init__(self, sparse_data: sparse.csr_array) -> None:
        assert sparse_data.ndim == 2
        self.sparse_data = sparse_data
        
    def __len__(self) -> None:
        return self.sparse_data.shape[0]
    
    def __getitem__(
        self,
        idx: numbers.Integral | slice | Iterable[int],
    ) -> sparse.csr_array:
        if isinstance(idx, numbers.Integral):
            return self._get_row(int(idx))
        elif isinstance(idx, (slice, Iterable)):
            iter_idx = range(len(self))[idx] if isinstance(idx, slice) else idx
            rows = [self._get_row(i) for i in iter_idx]
            return (
                sparse.csr_array(sparse.vstack(rows, format="csr"))
                if len(rows) != 0 else
                sparse.csr_array((0, self.sparse_data.shape[1]))
            )
        else:
            raise TypeError(
                f"Sequence index must be integer, slice or iterable, got {type(idx).__name__}"
            )
    
    def _get_row(self, idx: int) -> sparse.csr_array:
        return self.sparse_data[[idx], :]

sparse_array = sparse.csr_array([[0, 0], [0, 1], [1, 0], [1, 1]], dtype=np.float64)
labels = [0, 1, 1, 0]

# Succeeds
lgbm.Dataset(data=sparse_array, label=labels).construct()
# Fails
lgbm.Dataset(data=SparseSequence(sparse_array), label=labels).construct()
```

## Environment info
Python version 3.9.12
```
lightgbm == 3.3.2
numpy == 1.22.3
scipy == 1.8.0
```

## Additional Comments

[...]

PR #5211, opened 2022-05-13
[python-package] allow custom weighing in fobj for scikit-learn API (closes #5027)

### LINKED ISSUE #5027: [python] Weights is ignored when customized objective function is used
## Description
Weights are not multiplied to the gradients and hessians from customized objective function in Python API.

## Reproducible example
```python
import numpy as np
import lightgbm as lgb

def fobj(preds, train_data):
    labels = train_data.get_label()
    return preds - labels, np.ones_like(labels)

def test():
    np.random.seed(123)
    num_data = 10000
    num_feature = 100
    train_X = np.random.randn(num_data, num_feature)
    train_y = np.mean(train_X, axis=-1)
    valid_X = np.random.randn(num_data, num_feature)
    valid_y = np.mean(valid_X, axis=-1)
    weights = np.random.rand(num_data)
    train_data = lgb.Dataset(train_X, train_y, weight=weights) # comment out weights will get the same output
    valid_data = lgb.Dataset(valid_X, valid_y)
    params = {
        "verbose": 2,
        "metric": "rmse",
        "learning_rate": 0.2,
        "num_trees": 20,
    }
    booster = lgb.train(train_set=train_data, valid_sets=[valid_data], valid_names=["valid"], params=params, fobj=fobj)

if __name__ == "__main__":
    test()
```

Version 3.3.2

Install from source

jmoralez:
This removes the automatic weighing for `grad` and `hess` when using a custom objective in the scikit-learn API and makes the signature equivalent to the custom eval `(y_true, y_pred, weight, group)`.

Also adds tests to check that the behavior using the built-in objectives and the custom ones is consistent, i.e. if we give the weights to the dataset/fit using the built-in objectives we get the same results as if we use custom objectives and do the weighing ourselves. 

Closes #5027.


tests/python_package_test/test_sklearn.py
```diff
@@ -1273,18 +1273,64 @@ def test_training_succeeds_when_data_is_dataframe_and_label_is_column_array(task
...
+        weight = np.full_like(y, 2)
+        X_train, X_valid, y_train, y_valid, weight_train, weight_valid = train_test_split(
+            X, y, weight, test_size=0.2, random_state=0
+        )
+    else:
+        X_train, X_valid, y_train, y_valid = train_test_split(
+            X, y, test_size=0.2, random_state=0
+        )
+        weight_train = None
+        weight_valid = None
```
StrikerRUS, line 1313: I think this is a little bit more readable by avoiding code duplication.
```suggestion
    train_test_split_func = partial(train_test_split, test_size=0.2, random_state=0)
    X_train, X_valid, y_train, y_valid = train_test_split_func(X, y)
    if use_weight:
        weight = np.full_like(y, 2)
        weight_train, weight_valid = train_test_split_func(weight)
    else:
        weight_train = None
        weight_valid = None
```

tests/python_package_test/test_sklearn.py
```diff
@@ -1273,18 +1273,64 @@ def test_training_succeeds_when_data_is_dataframe_and_label_is_column_array(task
...
+    model.fit(
+        X_train,
+        y_train,
+        sample_weight=weight_train,
+        eval_set=[(X_train, y_train), (X_valid, y_valid)],
+        eval_names=['train', 'valid'],
+        eval_sample_weight=[weight_train, weight_valid],
+        eval_metric=custom_eval,
+        callbacks=[lgb.record_evaluation(eval_result)],
+    )
```
StrikerRUS, line 1326: We have a built-in `eval_result` in sklearn estimators.
https://github.com/microsoft/LightGBM/blob/2871b53e8e5e0ef23590410d7596d91173322ae4/python-package/lightgbm/sklearn.py#L749-L750
```suggestion
    model = lgb.LGBMClassifier(**params)
    model.fit(
        X_train,
        y_train,
        sample_weight=weight_train,
        eval_set=[(X_train, y_train), (X_valid, y_valid)],
        eval_names=['train', 'valid'],
        eval_sample_weight=[weight_train, weight_valid],
        eval_metric=custom_eval,
    )
    eval_result = model.evals_result_
```

StrikerRUS (changes requested):
Thank you so much for picking up this issue!
Please consider checking some my minor comments below:

python-package/lightgbm/sklearn.py
```diff
@@ -54,7 +54,10 @@ def __init__(self, func: _LGBM_ScikitCustomObjectiveFunction):
         Parameters
         ----------
         func : callable
-            Expects a callable with signature ``func(y_true, y_pred)`` or ``func(y_true, y_pred, group)``
+            Expects a callable with following signatures:
+            ``func(y_true, y_pred)``,
+            ``func(y_true, y_pred, weight)``
+            or ``func(y_true, y_pred, weight, group)``
```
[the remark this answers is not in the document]
jmoralez: Addressed in 71b3ba4

StrikerRUS:
@shiyu1994 Would you like to take a look at this PR as you were a creator of #5027?

StrikerRUS:
@shiyu1994

StrikerRUS:
@shiyu1994

StrikerRUS:
@jameslamb Please help with a second review.

jameslamb (approved):
Thanks for this! I believe I understand the changes, prior discussion on this PR, and previous discussion in #5207.

It's unfortunate that we have to make the breaking change where providing a callable that takes 3 arguments will now be treated as `(y_true, y_pred, weight)` when previous releases treated it as `(y_true, y_pred, group)`, since the values in `weight` and `group` can be similar and therefore code that worked in `lightgbm` v3.3.2 could **silently** start producing incorrect results in v4.0.0.

But I agree that that's worth it in exchange for consistency with the existing interface for scikit-learn **eval** functions, as @StrikerRUS mentioned in 

https://github.com/microsoft/LightGBM/issues/5027#issuecomment-1049283105

Since v4.0.0 will contain many breaking changes intended to reduce inconsistencies, and since the project's current state doesn't really support doing a round of deprecation warnings first (e.g. https://github.com/microsoft/LightGBM/pull/5133#pullrequestreview-995964570).

PR #5212, opened 2022-05-13
[R-package] minor improvements on weight demo

jmoralez:
Removes some extra lines that aren't necessary after #5204 and makes minor modifications to the weight descriptions.

### DIFF
--- a/R-package/demo/weight_param.R
+++ b/R-package/demo/weight_param.R
@@ -6,13 +6,13 @@
 library(lightgbm)
 
 # We will train a model with the following scenarii:
-# - Run 1: sum of weights equal to 0.06513 without adjusted regularization (not learning)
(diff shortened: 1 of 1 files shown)


StrikerRUS:
While we're on it, maybe transform this demo into a [vignette](https://github.com/microsoft/LightGBM/tree/master/R-package/vignettes) as a part of #1944?

jameslamb (approved):
Thanks for doing this! I agree with @StrikerRUS 's suggestion to consider converting this demo to a vignette, but I also think we should just merge this and that that can be a separate PR if you have time / interest.

jmoralez:
Yeah I have #1944 on my todo list, mainly to get more familiar with the full R-package. But I'll probably pick that up later, I'm currently focusing on the items of #5153.

jameslamb:
Thank you! Totally agree with that approach @jmoralez

PR #5215, opened 2022-05-15
Cleanup codeowners

StrikerRUS:
Remove inactive persons from the CODEOWNERS file.

Removed persons have been inactive for a long time (or simply never reviewed any PR) and unlikely will return to review new PRs.
CODEOWNERS file is used to automatically assign reviewers for PRs. Keeping this file clean and actual might be important for two reasons:
* outside contributors will be more patient for long delays in reviews when they understand the actual number of persons who are able to give a review;
* active maintainers will be more motivated to provide their reviews and feel greater responsibility for in time replies when they see only a couple of persons among assigned reviewers for a particular PR or that there is no "backup" reviewer for that code changes at all.

However, active maintainers can still ping these removed from the file persons manually in special cases according to this ["person - code area" mapping](https://lightgbm.readthedocs.io/en/latest/FAQ.html#critical-issues).

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -11,18 +11,18 @@
 
 # other catch-alls that will get matched if specific rules below are not matched
 *.R    @jameslamb @jmoralez
-*.py    @StrikerRUS @jmoralez @jameslamb @shiyu1994 @hzy46 @tongwu-msft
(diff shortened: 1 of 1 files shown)


jameslamb (approved):
Completely agree, thank you very much for doing this!

guolinke (approved):

PR #5225, opened 2022-05-19
[python-package] make a shallow copy when replacing categorical features with codes (fixes #4596)

### LINKED ISSUE #4596: Avoid dataframe deep copy while creating dataset
## Summary

The python API does the following while creating a dataset from a dataframe with cat cols:

```python
        if len(cat_cols):  # cat_cols is list
            data = data.copy()  # not alter origin DataFrame
```

By default `copy` does a deep copy (copies all data and indices) but AFAICS there is no need of that here and it could trigger a memory expensive operation.

## Motivation

Avoid extra memory usage. The sequence of steps here will end up with three copies of the data:
* the original dataframe
* the duplicated dataframe (transient)
* the dataset

## Description

Use `copy(deep=False)`:

```python
df = pd.DataFrame(dict(
    x=pd.Categorical(["a", "b", "b"]),
    y=[0, 1., 0],
))
df2 = df.copy(deep=False)
df2[["x"]] = df2[["x"]].apply(lambda x: x.cat.codes).replace({-1: np.nan})
print(df)
print(df2)
print(df2.values)
```

```
   x    y
0  a  0.0
1  b  1.0
2  b  0.0

   x    y
0  0  0.0
1  1  1.0
2  1  0.0

[[0. 0.]
 [1. 1.]
 [1. 0.]]
```

jmoralez:
Fixes #4596.

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -549,7 +549,7 @@ def _data_from_pandas(data, feature_name, categorical_feature, pandas_categorica
                 if list(data[col].cat.categories) != list(category):
                     data[col] = data[col].cat.set_categories(category)
         if len(cat_cols):  # cat_cols is list
-            data = data.copy()  # not alter origin DataFrame
+            data = data.copy(deep=False)  # not alter origin DataFrame
             data[cat_cols] = data[cat_cols].apply(lambda x: x.cat.codes).replace({-1: np.nan})
         if categorical_feature is not None:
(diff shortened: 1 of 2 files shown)


guolinke (approved):
Thank you!

jameslamb (approved):
very nice tests! Thanks for the excellent ork on this, especially using profiling to prove this works.

StrikerRUS (approved):
LGTM, thank you!

PR #5227, opened 2022-05-21
[R-package] silence more logs in unit tests

jameslamb:
Contributes to #4862.

This PR proposes additional changes to silence logs in the R package's tests.

### Notes for Reviewers

As of #5209, it's safe to change the verbosity in calls to `lightgbm()` without worrying about validation and early stopping.

Using the following code, I found that this PR reduces the output of the R package's tests from 636 lines to 73 lines.

```shell
cd R-package/tests
```


R-package/tests/testthat/test_dataset.R
```diff
@@ -404,6 +421,8 @@ test_that("lgb.Dataset: should be able to run lgb.cv() immediately after using l
     , metric = "binary_logloss"
     , num_leaves = 5L
     , learning_rate = 1.0
+    , num_iterations = 5L
```
jameslamb, line 424: note for you @jmoralez : here's the one line in this PR not *strictly* related to logging. Adding `verbosity` below is sufficient to suppress all log lines from this test.

But while looking at it, I noticed that this was running for 100 iterations (because I saw all the logs before silencing them). Reducing that to 5 just to save CI time.

jmoralez (approved):

StrikerRUS (approved):

jameslamb:
I see what looks like a bunch of networking-related issues across different R CI jobs.

`r-package (windows-latest, MINGW, R 3.6, cmake)` ([build link](https://github.com/microsoft/LightGBM/runs/6541216853?check_suite_focus=true#logs))

```text
Downloading https://cran.r-project.org/bin/windows/base/old/3.6.3/R-3.6.3-win.exe
Invoke-WebRequest: D:\a\LightGBM\LightGBM\.ci\test_r_package_windows.ps1:12
Line |
  12 |      Invoke-WebRequest -Uri $url -OutFile $destfile
     |      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | No connection could be made because the target machine actively refused it.
```

`r-sanitizers (ubuntu-latest, R-devel, gcc ASAN/UBSAN)` ([build link](https://github.com/microsoft/LightGBM/runs/6541216493?check_suite_focus=true#logs))

```text
Warning: unable to access index for repository https://cran.r-project.org/src/contrib:
  cannot open URL 'https://cran.r-project.org/src/contrib/PACKAGES'
```

And I see CRAN unavailable from my laptop as well.

```shell
curl -I cran.r-project.org
# curl: (7) Failed to connect to cran.r-project.org port 80: Connection refused

curl -I http://cran.r-project.org
# curl: (7) Failed to connect to cran.r-project.org port 80: Connection refused

curl -I https://cran.r-project.org
# curl: (7) Failed to connect to cran.r-project.org port 443: Connection refused
```

image

I think CRAN may be down.

StrikerRUS:
> I think CRAN may be down.

Yeah, it seems so. Still down...

jameslamb:
Looks like it is STILL down 😫 

Dirk makes a good point in https://stat.ethz.ch/pipermail/r-devel/2022-May/081749.html...other mirrors exist. I'm going to push a commit proposing switching to https://cran.rstudio.com/, which is up right now and which I expect might continue to have higher availability, since it is backed by a for-profit company and presumably supported by an on-call support staff.

jameslamb:
Just update this to latest `master` now that #5230 has been merged. That should help with CI. Will merge this if CI jobs all pass.

PR #5230, opened 2022-05-23
[ci] switch CRAN mirror to RStudio

jameslamb:
Created based on https://github.com/microsoft/LightGBM/pull/5227#issuecomment-1133817758.

This project's CI has been blocked for almost a day because cran.r-project.org has been down. According to one of CRAN's maintainers, this is because of an issue with an on-prem apache webserver running in a datacenter in Austria, and it won't be fixed until tomorrow because it's currently the weekend: https://stat.ethz.ch/pipermail/r-devel/2022-May/081750.html.

This PR proposes switching LightGBM's CI to use cran.rstudio.com instead for all its R CI jobs. That mirror is backed by a for-profit company and (I assume) supported by an on-call rotation of engineers who will quickly respond to availability issues. This blog post from 2013 hints how it is architected: https://www.rstudio.com/blog/rstudio-cran-mirror/. This is the default mirror set for everyone using R in the RStudio IDE, so I strongly expect it to have higher availability than cran.r-project.org.

### Shouldn't we change links in documentation too?

I intentionally limited the scope of this PR to only LightGBM's CI jobs, and left its documentation pointing to `cran.r-project.org` untouched. I recall CRAN complaining in the past about "non-canonical URLs", and I see the following in https://cran.rstudio.com/doc/manuals/r-release/R-exts.html#Specifying-URLs

> The canonical URL for a CRAN package is `https://cran.r-project.org/package=pkgname`

Writing R Extensions doesn't say whether CRAN is opinionated about which mirror is referenced in such links, but I don't want to risk a rejection by changing the existing README links.

My primary goal with this PR is to unblock LightGBM's development.

### DIFF
--- a/.ci/test_r_package.sh
+++ b/.ci/test_r_package.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # set up R environment
-CRAN_MIRROR="https://cloud.r-project.org/"
+CRAN_MIRROR="https://cran.rstudio.com"
 R_LIB_PATH=~/Rlib
 mkdir -p $R_LIB_PATH
 export R_LIBS=$R_LIB_PATH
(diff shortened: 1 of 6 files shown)


StrikerRUS (approved):
Awesome! Thank you for the quick fix!

PR #5237, opened 2022-05-24
[R-package] silence logs in print(), show(), summary() tests

jameslamb:
Contributes to #4862.

This PR proposes capturing the logs printed from tests on `show()`, `summary()`, and `print()` methods for `Booster`, added in #4686.

Using the following code, I found that this PR reduces the output of the R package's tests from 73 lines to 28 lines.

```shell
cd R-package/tests
```

### DIFF
--- a/R-package/tests/testthat/test_lgb.Booster.R
+++ b/R-package/tests/testthat/test_lgb.Booster.R
@@ -1253,24 +1253,64 @@ test_that("Booster's print, show, and summary work correctly", {
        )
     }
 
+    .has_expected_content_for_fitted_model <- function(printed_txt) {
+      expect_true(any(grepl("^LightGBM Model", printed_txt)))
+      expect_true(any(grepl("^Fitted to dataset", printed_txt)))
+    }
+
+    .has_expected_content_for_finalized_model <- function(printed_txt) {
+      expect_true(any(grepl("^LightGBM Model$", printed_txt)))
+      expect_true(any(grepl("Booster handle is invalid", printed_txt)))
+    }
+
     .check_methods_work <- function(model) {
 
-        # should work for fitted models
-        ret <- print(model)
-        .have_same_handle(ret, model)
-        ret <- show(model)
-        expect_null(ret)
-        ret <- summary(model)
-        .have_same_handle(ret, model)
+        #--- should work for fitted models --- #
 
-        # should not fail for finalized models
-        model$finalize()
-        ret <- print(model)
+        # print()
+        log_txt <- capture.output({
+          ret <- print(model)
+        })
         .have_same_handle(ret, model)
-        ret <- show(model)
+        .has_expected_content_for_fitted_model(log_txt)
+
+        # show()
+        log_txt <- capture.output({
+          ret <- show(model)
+        })
         expect_null(ret)
-        ret <- summary(model)
+        .has_expected_content_for_fitted_model(log_txt)
+
+        # summary()
+        log_text <- capture.output({
+          ret <- summary(model)
+        })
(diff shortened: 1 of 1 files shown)


jmoralez (approved):
Nice! I see only the test files names and the dots. What are the tasks left to close #4862?

jameslamb:
> What are the tasks left to close https://github.com/microsoft/LightGBM/issues/4862

It's really weird to me that the logs at https://pipelines.actions.githubusercontent.com/serviceHosts/1f53ab1f-1bd7-45b1-9237-cc6b054df090/_apis/pipelines/1/runs/30984/signedlogcontent/17?urlExpires=2022-05-24T16%3A18%3A43.5279800Z&urlSigningMethod=HMACV1&urlSignature=3rzC%2F8PrYEI46MUXIzdWrTHDcMPMfcLJPuT7wmriySk%3D (for example) are getting truncated after the `dataset` tests.

```text
basic: ............................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................
custom_objective: ........
dataset: ................................No issues were found checking the R package
```

There should actually be many more. Here's what I see running `Rscript testthat.R` on my laptop.

```text
Predictor: ..................................................................................................................................................................
basic: ...........................................................[LightGBM] [Fatal] Unknown objective type name: not_a_real_objective
..........................[LightGBM] [Fatal] Unknown objective type name: not_a_real_objective
..................................................................................................................................................................................................................................[LightGBM] [Fatal] Cannot change linear_tree after constructed Dataset handle.
.....................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................
custom_objective: ........
dataset: ................................[LightGBM] [Fatal] Initial score size doesn't match data size
..............................................
learning_to_rank: ...............................................
lgb.Booster: .........................................................................[LightGBM] [Fatal] Unknown importance type: only support split=0 and gain=1
...................................[LightGBM] [Fatal] Unknown objective type name: not_a_real_objective
...[LightGBM] [Fatal] Could not open file-that-does-not-exist.model
.[LightGBM] [Fatal] Unknown model format or submodel type in model file /var/folders/b4/rzfqc68n4nj5flhvf3041ycr0000gr/T//Rtmpo8uaJX/file83d75faa23ce.model
.[LightGBM] [Fatal] Model file doesn't specify the number of classes
............................................................................................
lgb.convert_with_rules: .........................................................................................................................................
lgb.importance: ...........
lgb.interprete: ........
lgb.plot.importance: .........
lgb.plot.interpretation: ...
metrics: ...
parameters: ..........................
utils: ......................
weighted_loss: ...
```

The last piece for #4862 is to eliminate those Fatal logs that make it through. Those happen because for the R package, `Log::Fatal()` on the C++ side writes to stderr 

https://github.com/microsoft/LightGBM/blob/f2e1ad6f50f1795987ad92b4f4dc61c3bbb0c8f3/include/LightGBM/utils/log.h#L121-L130

AND raises an R exception (which again writes to stderr).

https://github.com/microsoft/LightGBM/blob/f2e1ad6f50f1795987ad92b4f4dc61c3bbb0c8f3/R-package/src/lightgbm_R.cpp#L37-L42

The R exception is trapped and its output suppressed by `testthat::expect_error()`, but not that first print.

jmoralez:
Oh I was looking at the [MSVC logs](https://github.com/microsoft/LightGBM/runs/6566568032?check_suite_focus=true#step:6:1958) and they look perfect but I see the same as you do on my laptop, I thought it was something on my side because it has M1 but I run it with Rosetta so I figured it didn't work correctly.

PR #5239, opened 2022-05-24
[python] add more type hints on LGBMModel methods

jameslamb:
Contributes to #3756.

Adds type hints on some methods in `lightgbm.sklearn.LGBMModel` that don't currently have them.

### DIFF
--- a/python-package/lightgbm/sklearn.py
+++ b/python-package/lightgbm/sklearn.py
@@ -505,7 +505,7 @@ class LGBMModel(_LGBMModelBase):
         self._n_classes = None
         self.set_params(**kwargs)
 
-    def _more_tags(self):
+    def _more_tags(self) -> Dict[str, Any]:
         return {
             'allow_nan': True,
             'X_types': ['2darray', 'sparse', '1dlabels'],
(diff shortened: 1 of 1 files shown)


StrikerRUS:
I remember the following discussion about these mypy false alarms: https://github.com/microsoft/LightGBM/pull/4837#pullrequestreview-819683612.

jameslamb:
Thanks for finding those links @StrikerRUS ! I remembered those conversations before, and am ok with the conclusion we reached there, that `# type: ignore` comments could be used to handle these cases where we "just know" that an attribute like `n_classes_` will always either be an integer (if model has been fitted) or raise an exception (if not fitted).

I didn't add such comments in this PR because I just wanted to focus on getting the type hints correct, and wanted to leave `mypy` warnings for separate PRs targeting #3867.

jmoralez (approved):

StrikerRUS:
> I didn't add such comments in this PR because I just wanted to focus on getting the type hints correct, and wanted to leave `mypy` warnings for separate PRs targeting #3867.

Great, totally agree!

PR #5226, opened 2022-05-21
remove support for Solaris (fixes #5216)

jameslamb:
Fixes #5216 .
Related to #3513.

Per the discussion in https://github.com/microsoft/LightGBM/pull/5217#issuecomment-1127856547, this PR proposes removing support for the Solaris operating system in LightGBM.

That support was added only to get `{lightgbm}` (the R package) onto CRAN, so that it would be easier for R users to install. 
There is strong evidence (provided in that comment) that CRAN no longer requires packages to be installable on Solaris.

I'm proposing removing this support for the following reasons:

1. R Hub is the only easy-to-use option we have found for testing that LightGBM is portable to Solaris, and the fact that CRAN no longer requires such portability means that incidents like #5216 which prevent us from testing on R Hub could easily happen again
2. LightGBM is already struggling from a lack of maintainer activity, and removing a source of maintenance burden will allow us to focus our limited time and attention on higher-value areas of the project.

PR #5248, opened 2022-05-28
Remove leftovers after the drop of Solaris support

StrikerRUS:
Refer to #5226.


jameslamb (approved):
Ah right, sorry about that! I did `git grep -i solaris` but forgot to also search for `sun`.

Issue #5228, opened 2022-05-22
[R-package] [ci] static analysis CI job not compatible with {lintr} 3.0.0


jameslamb:
## Description

This project uses `{lintr}` to catch some types of efficiency and correctness issues and to enforce standard style across all its R code.

I noticed on @MichaelChirico 's Twitter tonight that a new major release of `{lintr}` will be published to CRAN soon.

https://twitter.com/michael_chirico/status/1528022047502217216

I tested `{lightgbm}` with that new release tonight, and found that this project's current linting setup raises some errors and warnings with that new version. If no changes are made, once `{lintr}` 3.0.0 is published to CRAN, LightGBM's linting CI job will start to break.

## Reproducible example

Installed `{lintr}` from source on my Mac (R 4.1.0).

```shell
Rscript -e "remotes::install_github('https://github.com/r-lib/lintr.git')"
```

Then ran this project's linting.

```shell
Rscript .ci/lint_r_code.R $(pwd)/R-package
```

That produced the following errors and warnings.

```text
Error: `fun` must be a function taking exactly one argument.
In addition: Warning messages:
1:  The use of linters of class 'function' was deprecated in lintr version 2.0.1.9001. Use linters classed as 'linter' (see ?Linter) instead.
2:  The use of linters of class 'function' was deprecated in lintr version 2.0.1.9001. Use linters classed as 'linter' (see ?Linter) instead.
Execution halted
```

After modifying LightGBM's CI code to get past that error, I saw the following errors.

```text
Warning messages:
1: Linter closed_curly_linter was deprecated in lintr version 2.0.1.9001. Use brace_linter instead.
2: Linter open_curly_linter was deprecated in lintr version 2.0.1.9001. Use brace_linter instead.
3: Linter paren_brace_linter was deprecated in lintr version 2.0.1.9001. Use brace_linter instead.
4: Linter semicolon_terminator_linter was deprecated in lintr version 2.0.1.9001. Use semicolon_linter instead.
```

Also found that some of the R package's code will now fail the set of linters it's using. At first glance, these look to me like things I'd expected these linters to already catch, so they must be the benefit of bugfixes in `{lintr}`!

<details><summary>14 new linter failures (click me)</summary>

```text
/Users/jlamb/repos/LightGBM/R-package/inst/make-r-def.R:74:20: warning: Use file.path() to construct portable file paths.
        pattern = "[Ordinal/Name Pointer] Table"
[...]

PR #5249, opened 2022-05-29
[R-package] address linter warnings about portable paths

jameslamb:
Contributes to #5228.

Resolves the remaining linter warnings that would be raised by `{lintr}` 3.0.0 once it's released (https://github.com/microsoft/LightGBM/issues/5228#issuecomment-1140355470).

```text
[1] Total linting issues found: 4
[[1]]
/Users/jlamb/repos/LightGBM/R-package/inst/make-r-def.R:74:20: warning: Use file.path() to construct portable file paths.
        pattern = "[Ordinal/Name Pointer] Table"
                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~

[[2]]
/Users/jlamb/repos/LightGBM/R-package/src/install.libs.R:99:4: warning: Use file.path() to construct portable file paths.
  "../inst/bin/CMakeLists.txt"
   ^~~~~~~~~~~~~~~~~~~~~~~~~~

[[3]]
/Users/jlamb/repos/LightGBM/R-package/src/install.libs.R:125:6: warning: Use file.path() to construct portable file paths.
    "../../inst/make-r-def.R"
     ^~~~~~~~~~~~~~~~~~~~~~~

[[4]]
/Users/jlamb/repos/LightGBM/R-package/tests/testthat/test_basic.R:2428:12: warning: Use file.path() to construct portable file paths.
test_that("lgb.train() w/ linear learner fails already-constructed dataset with linear=false", {
           ^~
```

### DIFF
--- a/R-package/inst/make-r-def.R
+++ b/R-package/inst/make-r-def.R
@@ -71,7 +71,7 @@ invisible(file.remove(OBJDUMP_FILE))
 # see https://www.cs.colorado.edu/~main/cs1300/doc/mingwfaq.html
 start_index <- which(
     grepl(
-        pattern = "[Ordinal/Name Pointer] Table"
+        pattern = "[Ordinal/Name Pointer] Table"  # nolint
         , x = objdump_results
         , fixed = TRUE
     )
--- a/R-package/src/install.libs.R
+++ b/R-package/src/install.libs.R
@@ -23,6 +23,11 @@ if (!(R_int_UUID == "0310d4b8-ccb1-4bb8-ba94-d36a55f60262"
   warning("Warning: unmatched R_INTERNALS_UUID, may not run normally.")
 }
 
+# Get some paths
+source_dir <- file.path(R_PACKAGE_SOURCE, "src", fsep = "/")
+build_dir <- file.path(source_dir, "build", fsep = "/")
+inst_dir <- file.path(R_PACKAGE_SOURCE, "inst", fsep = "/")
+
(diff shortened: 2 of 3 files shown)


jmoralez (approved):

R-package/inst/make-r-def.R
```diff
@@ -71,7 +71,7 @@ invisible(file.remove(OBJDUMP_FILE))
 # see https://www.cs.colorado.edu/~main/cs1300/doc/mingwfaq.html
 start_index <- which(
     grepl(
-        pattern = "[Ordinal/Name Pointer] Table"
+        pattern = "[Ordinal/Name Pointer] Table"  # nolint
```
MichaelChirico, line 74: recommend using `# nolint: portable_paths_linter.` for this to help prevent other lints from cropping up on the same line & being hidden :)

jameslamb: oooo I didn't know that it was possible to target specific linters like that, [the way you can with flake8](https://flake8.pycqa.org/en/3.1.1/user/ignoring-errors.html#in-line-ignoring-errors).

Thanks for letting us know! I'll put up a PR doing that.

MichaelChirico: It won't actually do anything yet -- it's part of the 3.0.0 release. But it doesn't hurt to start using it now!

jameslamb: Got it, thanks for letting us know about this!

PR #5250, opened 2022-05-29
[R-package] [ci] silence more logs in tests (fixes #4862)

### LINKED ISSUE #4862: [R-package] [tests] reduce verbosity in unit tests

jameslamb:
Fixes #4862.

This removes the last remaining log lines from the R tests (other than dots for test statuses).

As described in https://github.com/microsoft/LightGBM/pull/5237#issuecomment-1136142589, when LightGBM's C++ code calls `Log::Fatal()`, the same log message comes through stderr twice:

1. as a log message printed with `REprintf()`
2. as an R exception

image

I think it makes sense to have both, so that if user code try-catches `{lightgbm}` stuff, suppressing those exceptions, they'll still at least see a log message indicating that the exception happened.

This PR proposes just using `capture.output()` in the R package's tests to keep those logs out of the test output.

### DIFF
--- a/R-package/tests/testthat/test_basic.R
+++ b/R-package/tests/testthat/test_basic.R
@@ -561,13 +561,15 @@ test_that("lgb.cv() raises an informative error for unrecognized objectives", {
     , label = train$label
   )
   expect_error({
-    bst <- lgb.cv(
-      data = dtrain
-      , params = list(
-        objective_type = "not_a_real_objective"
-        , verbosity = VERBOSITY
+    capture.output({
+      bst <- lgb.cv(
+        data = dtrain
+        , params = list(
+          objective_type = "not_a_real_objective"
+          , verbosity = VERBOSITY
+        )
       )
-    )
+    }, type = "message")
   }, regexp = "Unknown objective type name: not_a_real_objective")
 })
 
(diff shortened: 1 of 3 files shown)


jmoralez (approved):
Nice! I just ran this locally and see only the tests and their status

jameslamb:
Great! Thanks for double-checking.

Issue #3641, opened 2020-12-10
Warning shown with verbosity=-1


memeplex:
According to the documentation `< 0: Fatal, = 0: Error (Warning), = 1: Info, > 1: Debug` but I'm unable to get rid  of the `2^max_depth > num_leaves` warning.

## How you are using LightGBM?

LightGBM component: Python package

## Environment info

Operating System: MacOS 11.0.1 (Big Sur)

CPU/GPU model: 2,3 GHz Dual-Core Intel Core i5 / Intel Iris Plus Graphics 640 1536 MB

Python version: 3.8.6

## Error message and / or logs

```
[LightGBM] [Warning] Accuracy may be bad since you didn't explicitly set num_leaves OR 2^max_depth > num_leaves. (num_leaves=31).
```

## Reproducible example(s)

```
import lightgbm as lgb
import pandas as pd
import numpy as np

X = pd.DataFrame({
    "x1": np.arange(100),
    "x2": np.arange(100),
})
y = X.x1 + X.x2
lgb.train({"max_depth": 10, "verbose": -1}, lgb.Dataset(X, y))
```

memeplex:
Besides, the UserWarning mentioned in https://github.com/microsoft/LightGBM/issues/3640 and https://github.com/microsoft/LightGBM/issues/3379 is also shown when verbosity = -1. This is a python warning and as such probably works at a different level, but I believe it would be better to treat all kind of warnings in a consistent way.

memeplex:
Some additional info:

1. It only happens the first time, I've to reset the kernel to reproduce it one more time. As in another recent report of mine, some global state seems to be persisten between invocations (probably config, since it's global).

2. Looking at the Booster class initializer the problem seems to happen here:

```
            train_set.construct()                                         <--- shows the warning
            # copy the parameters from train_set
            params.update(train_set.get_params())
            params_str = param_dict_to_str(params)
            self.handle = ctypes.c_void_p()
            _safe_call(_LIB.LGBM_BoosterCreate(          <--- but config is initialized here (too late?).
```

memeplex:
Oh, no, I think it's much simpler! It's all in config.cpp:

```
 // check for conflicts
  CheckParamConflict();  <---- shows the warning

  if (verbosity == 1) {.  <---- sets log level
    LightGBM::Log::ResetLogLevel(LightGBM::LogLevel::Info);
  } else if (verbosity == 0) {
[...]

PR #5253, opened 2022-05-30
suppress alias warnings with verbosity<0 (fixes #4518)

### LINKED ISSUE #4518: Unable to suppress parameter alias related warnings
## Description

Many parameter aliases trigger warnings that may be annoying and repetitive, specially in the context of model selection using grid search. These warnings can't be silenced except by changing the parameter to some specific alias. Often, this specific alias is not the one "recommended" (most used or first listed) in the documentation.

## Reproducible example

See https://github.com/microsoft/LightGBM/issues/3641#issuecomment-898092405

## Environment info

LightGBM version or commit hash: 3.2.1

jmoralez:
Sets the logging level before resolving the parameter aliases so that the warnings related to them can be suppressed using the verbosity parameter.

Fixes #4518.


tests/python_package_test/test_engine.py
```diff
@@ -3578,3 +3578,26 @@ def test_boost_from_average_with_single_leaf_trees():
     preds = model.predict(X)
     mean_preds = np.mean(preds)
     assert y.min() <= mean_preds <= y.max()
+
+
+@pytest.mark.parametrize('verbosity', [-1, 0])
```
jameslamb, line 3583: Thanks for this test!

I'm ok with hard-coding `"verbose"` followed by `"verbosity"` in `config.cpp`, as I think that's more efficient than calling  `LGBM_DumpParamAliases()` (introduced in #4829) or something like it to get all parameter aliases that control verbosity.

But could you modify this test such that it would catch the case "LightGBM doesn't respect all aliases for managing verbosity"? That way, we'd know to modify `config.cpp` if 

I think that could be accomplished by changing the test structure to something like the following:

```python
@pytest.mark.parametrize('verbosity_param', _ConfigAliases.get("verbose"))
@pytest.mark.parametrize('verbosity_value', [-1, 0])
def test_verbosity_can_suppress_alias_warnings(capsys, verbosity_param, verbosity_value):
    # ...
    params = {
        # ...
        verbosity_param: verbosity_value
        # ...
    }
    # ... 
```

src/io/config.cpp
```diff
@@ -42,6 +42,19 @@ std::unordered_map<std::string, std::string> Config::Str2Map(const char* paramet
...
+  GetInt(params, "verbosity", &verbosity);
+  if (verbosity == 1) {
+    LightGBM::Log::ResetLogLevel(LightGBM::LogLevel::Info);
+  } else if (verbosity == 0) {
+    LightGBM::Log::ResetLogLevel(LightGBM::LogLevel::Warning);
+  } else if (verbosity >= 2) {
+    LightGBM::Log::ResetLogLevel(LightGBM::LogLevel::Debug);
+  } else {
+    LightGBM::Log::ResetLogLevel(LightGBM::LogLevel::Fatal);
+  }
```
jameslamb, line 57: Super minor thing, but would you consider re-ordering this such that it is order by verbosity (either from most to least verbose or vice-versa)? Right now it's jumping back and forth.

e.g., I think it's be slightly easier to understand if it went `DEBUG --> INFO --> WARN --> FATAL` or `FATAL --> WARN --> INFO --> DEBUG`

tests/python_package_test/test_engine.py
```diff
@@ -3578,3 +3578,26 @@ def test_boost_from_average_with_single_leaf_trees():
...
+        'verbosity': verbosity,
+    }
+    lgb.train(params, ds, num_boost_round=1)
+    expected_msg = (
+        '[LightGBM] [Warning] bagging_fraction is set=0.8, subsample=0.75 will be ignored. '
+        'Current value: bagging_fraction=0.8'
+    )
+    stdout = capsys.readouterr().out
+    if verbosity == -1:
+        assert stdout == ''
```
jameslamb, line 3601: Instead of testing that `stdout` is empty, I think this should be changed to checking that there are 0 LightGBM-generated messages (i.e. those containing `[LightGBM]`) emitted.

That would prevent the case where this test fails when one of `lightgbm`'s dependencies emits some log messages (like a deprecation warning or something) that `lightgbm` does not try to suppress based on the value of its `verbosity` setting.

I proposed something similar in the R package's tests when working on #4862, e.g. https://github.com/microsoft/LightGBM/blob/fb37e5074ad0163afd9d88366b58075899775cfa/R-package/tests/testthat/test_basic.R#L2312-L2315

jameslamb (changes requested):
Thanks for doing this! Like I mentioned in https://github.com/microsoft/LightGBM/issues/4518#issuecomment-1140363388, I totally support this change. But I'd like to hear @StrikerRUS's perspective on this one as well before this is merged.

Until them, please see a few small suggestions below.

src/io/config.cpp
```diff
@@ -240,16 +253,6 @@ void Config::Set(const std::unordered_map<std::string, std::string>& params) {
     save_binary = true;
   }
 
-  if (verbosity == 1) {
```
jameslamb, line 243: I was surprised to see this being removed, but I *think* it's ok. I couldn't find any cases of `Config.Set()` being called without `Config::Str2Map()` being called, and we have many tests on verbosity in the R and Python package which are passing.

Just wanted to say that out loud, for the benefit of other reviewers.

jameslamb (approved):
Changes look good to me, thanks! Since `Config` is so central to the codebase, I'd like one other maintainer to review and approve before we merge this.

src/io/config.cpp
```diff
@@ -42,6 +42,19 @@ std::unordered_map<std::string, std::string> Config::Str2Map(const char* paramet
   for (auto arg : args) {
     KV2Map(&params, Common::Trim(arg).c_str());
   }
+  // define verbosity and set logging level
+  int verbosity = Config().verbosity;
+  GetInt(params, "verbose", &verbosity);
+  GetInt(params, "verbosity", &verbosity);
```
StrikerRUS, line 48: If params have both `verbose` and `verbosity` keys I think we should still emit the following warning for this case
https://github.com/microsoft/LightGBM/blob/1cc9f9dcee15586aefdd8271a1c04ad010d3c53a/include/LightGBM/config.h#L1145-L1147
to inform user that `verbosity` has higher precedence over `verbose`.

StrikerRUS, line 48: I believe we should pop `verbose` and `verbosity` here to not process them twice later
https://github.com/microsoft/LightGBM/blob/1cc9f9dcee15586aefdd8271a1c04ad010d3c53a/src/io/config.cpp#L45

src/io/config.cpp
```diff
@@ -42,6 +42,19 @@ std::unordered_map<std::string, std::string> Config::Str2Map(const char* paramet
   for (auto arg : args) {
     KV2Map(&params, Common::Trim(arg).c_str());
   }
+  // define verbosity and set logging level
+  int verbosity = Config().verbosity;
```
StrikerRUS, line 46: Shouldn't we update the following function as well?
https://github.com/microsoft/LightGBM/blob/1cc9f9dcee15586aefdd8271a1c04ad010d3c53a/src/application/application.cpp#L50

src/io/config.cpp
```diff
@@ -42,6 +42,19 @@ std::unordered_map<std::string, std::string> Config::Str2Map(const char* paramet
   for (auto arg : args) {
     KV2Map(&params, Common::Trim(arg).c_str());
```
StrikerRUS, line 43: Looks like the following warnings in `KV2Map()` function are still cannot be suppressed by passing `verbosity` param because `// define verbosity and set logging level` happens later in the code
https://github.com/microsoft/LightGBM/blob/1cc9f9dcee15586aefdd8271a1c04ad010d3c53a/src/io/config.cpp#L28-L35

tests/python_package_test/test_engine.py
```diff
@@ -3578,3 +3579,27 @@ def test_boost_from_average_with_single_leaf_trees():
...
+    lgb.train(params, ds, num_boost_round=1)
+    expected_msg = (
+        '[LightGBM] [Warning] bagging_fraction is set=0.8, subsample=0.75 will be ignored. '
+        'Current value: bagging_fraction=0.8'
+    )
+    stdout = capsys.readouterr().out
+    if verbosity == -1:
+        assert re.search(r'\[LightGBM\]', stdout) is None
+    else:
+        assert expected_msg in stdout
```
StrikerRUS, line 3605: I believe this will be more correct.
```suggestion
    if verbosity >= 0:
        assert expected_msg in stdout
    else:
        assert re.search(r'\[LightGBM\]', stdout) is None
```

tests/python_package_test/test_engine.py
```diff
@@ -3578,3 +3579,27 @@ def test_boost_from_average_with_single_leaf_trees():
     preds = model.predict(X)
     mean_preds = np.mean(preds)
     assert y.min() <= mean_preds <= y.max()
+
+
+@pytest.mark.parametrize('verbosity_param', lgb.basic._ConfigAliases.get("verbosity"))
+@pytest.mark.parametrize('verbosity', [-1, 0])
+def test_verbosity_can_suppress_alias_warnings(capsys, verbosity_param, verbosity):
```
StrikerRUS, line 3586: Since the original issue is quite "sensitive", could you please add a test that checks behavior when user passes multiple aliases for the `verbosity` param?
Refer to .

StrikerRUS (commented):
Thanks for this PR!
I'm sorry, config initialization routine is still not clear enough for me. However, I left some comments that may help to spot uncovered by this PR cases.

src/io/config.cpp
jmoralez: These warnings come from duplicated parameters, not aliases. I can try to address them as well in this PR if its in the scope.

src/io/config.cpp
jmoralez: I believe if I move the logic into `KeyAliasTransform` this function will work as expected as well, except for the duplicate argument warnings.

src/io/config.cpp
jmoralez: I wanted to leave them because I'm not sure if they're used some place after but I can drop them here if this is the only place they're used.

src/io/config.cpp
StrikerRUS: I think it's good idea to move the logic into `KeyAliasTransform`.

src/io/config.cpp
StrikerRUS: I believe it's related.

jmoralez:
I've changed the approach, now:
* Every parameter is saved to an `unordered_multimap` which can have several values for the same key, this is so we can see if there are any duplicate parameters after parsing them.
* Once we have all the parameters in the multimap we check for `verbose` and `verbosity` and set the logging level.
* We then move the multimap to a regular map and warn about duplicate arguments. This step takes a random one if two or more are provided (we can discuss this).
* Once the parameters are in a map we check for the aliases as before.
This of course uses more memory now but I couldn't think of a better way to handle the possible duplicate parameters. I'm open to adapt this to a better idea.

jmoralez:
There are some issues right now with this because it prints the warnings several times and there seems to be a weird interaction with the dataset both in the python and R-packages where if you don't re-construct the dataset, verbosity -1 doesn't remove the alias warnings. I have to look into that.

jmoralez:
Seems like the duplicate warnings about aliases is on master as well. I'll open an issue.

src/io/config.cpp
StrikerRUS: I agree with you. I believe each warning is helpful, but unfortunately LightGBM users want to completely silence their program (according to the issues they create in the repo).

guolinke (approved):
Thank you!

src/application/application.cpp
```diff
@@ -68,16 +67,16 @@ void Application::LoadParameters(int argc, char** argv) {
         if (line.size() == 0) {
           continue;
         }
-        Config::KV2Map(&params, line.c_str());
+        Config::KV2Map(&all_params, line.c_str());
       }
     } else {
-      Log::Warning("Config file %s doesn't exist, will ignore",
-                   params["config"].c_str());
+      Log::Warning("Config file %s doesn't exist, will ignore", all_params["config"][0].c_str());
```
StrikerRUS, line 73: Ah, this warning will be emitted despite the verbosity setting.

StrikerRUS:
Could you please resolve merge conflicts?

src/io/config.cpp
```diff
@@ -22,26 +22,61 @@ void Config::KV2Map(std::unordered_map<std::string, std::string>* params, const
...
-          key.c_str(), value_search->second.c_str());
-      }
+      params->operator[](key).emplace_back(value);
     }
   } else {
     Log::Warning("Unknown parameter %s", kv);
   }
 }
 
+void RetrieveFirstValueFromKey(const std::unordered_map<std::string, std::vector<std::string>>& params, std::string key, int* out) {
```
StrikerRUS, line 32: I believe it may be confusing that this function designed for the general purpose (according to its name) checks that value should be `int` typed.

StrikerRUS (commented):
Thanks a lot for this PR!
I don't have any other additional comments except a one below about the function naming.

But I think I'm not qualified to formally approve these mostly cpp changes.
✔️

src/io/config.cpp
jmoralez: I modified the name to `GetFirstValue`. Given that the `out` param is an `int*` I think it's expected to check that the value is an `int`.

src/io/config.cpp
```diff
@@ -22,26 +22,61 @@ void Config::KV2Map(std::unordered_map<std::string, std::string>* params, const
...
-          key.c_str(), value_search->second.c_str());
-      }
+      params->operator[](key).emplace_back(value);
     }
   } else {
     Log::Warning("Unknown parameter %s", kv);
   }
 }
 
+void GetFirstValue(const std::unordered_map<std::string, std::vector<std::string>>& params, std::string key, int* out) {
```
guolinke, line 32: maybe rename to "GetFirstIntValue" ? as it only supports int type?

guolinke (approved):
Looks good to me, except a minor naming

src/io/config.cpp
jmoralez: changed to `GetFirstValueAsInt` in e9a7a0f

StrikerRUS:
Can we merge this PR?

jameslamb:
> Seems like the duplicate warnings about aliases is on master as well. I'll open an issue.

did this happen? I couldn't find an issue like that from search and don't see one linked here.

> Can we merge this PR?

Based on what I see in the diff for unit tests, I think this can be merged. It does seem like the warnings described in https://github.com/microsoft/LightGBM/issues/3641#issuecomment-898092405 (which led to #4518) can now be suppressed with `verbosity`.

But I'll defer to @jmoralez . To be honest, there have been significant changes since I first approved and it's difficult for me to understand the current state of this PR a bit.

For example, the conversation in https://github.com/microsoft/LightGBM/pull/5253#discussion_r889728704 is unresolved.

jmoralez:
> did this happen?

Yes, I opened https://github.com/microsoft/LightGBM/issues/5332.

The duplicate parameter warnings were also addressed in this PR. The `KeepFirstValues` function keeps the first value defined for each parameter and warns about the rest (if there are any). This allows to override parameters defined in a configuration file through arguments in the CLI, or if defined twice in the parameter list in the R-package, the first one takes precedence and a warning is printed for the second one.

My opinion is that this is ready to be merged, but of course I'm biased haha.

src/io/config.cpp
jmoralez: Addressed by having an unordered_map of vectors where we store all parameters for as many times as they're defined, set the verbosity taking the first values of `verbose` and `verbosity` and then de-duplicating with the `KeepFirstValues` function, where warning level logs are printed but after the verbosity has been set, so these can be disabled as well.

src/io/config.cpp
jmoralez: I left them so that we can warn about duplicates later.

jmoralez: This is done in the `KeyAliasTransform` function after setting the verbosity and de-duplicating.

jameslamb (changes requested):
> Yes, I opened https://github.com/microsoft/LightGBM/issues/5332.

ah ok, thanks for pointing that out! I just got unlucky with the search terms I chose 😅 

If that issue had mentioned this PR with `#5253` or if a comment had been posted on this PR saying "created `#5332`", it would have been easier for me to find.

> The duplicate parameter warnings were also addressed in this PR

Sorry, I'm still confused...do you mean specifically that as of this PR being merged, we could also close #5332?

If so, please:

* add "fixes #5332" to the description
* add one more test confirming that the duplicate warnings are no longer raised

jmoralez:
> do you mean specifically that as of this PR being merged, we could also close https://github.com/microsoft/LightGBM/issues/5332?

No, I meant that this was originally only going to remove warnings for parameter aliases, i.e. you set `bagging_fraction` and `subsample`, but in https://github.com/microsoft/LightGBM/pull/5253#discussion_r889728704 it was suggested that this also addressed warnings for duplicate parameters, i.e. you set `bagging_fraction=0.7` and `bagging_fraction=0.8`, so that conversation was resolved because it was also addressed.

The duplicate prints (#5332) is not solved yet.

jameslamb (approved):
Thanks for this and sorry for taking so long to come back and re-review! I understand now what his PR does and doesn't fix.

Let's merge this...it's only changing internal details, so easy to revise any of these design decisions in the future if we decide a different pattern is preferable for changing other logging-related stuff.

Awesome contribution, @jmoralez ! People have been asking for this for a long time 😬

PR #5254, opened 2022-05-30
[python-package] make a shallow copy on dataframe rename (fixes #4596)

### LINKED ISSUE #4596: Avoid dataframe deep copy while creating dataset

jmoralez:
Adds on #5225. Fixes #4596.

### DIFF
--- a/tests/python_package_test/test_basic.py
+++ b/tests/python_package_test/test_basic.py
@@ -648,9 +648,7 @@ def test_no_copy_when_single_float_dtype_dataframe(dtype):
     pd = pytest.importorskip('pandas')
     X = np.random.rand(10, 2).astype(dtype)
     df = pd.DataFrame(X)
-    # feature names are required to not make a copy (rename makes a copy)
-    feature_name = ['x1', 'x2']
-    built_data = lgb.basic._data_from_pandas(df, feature_name, None, None)[0]
+    built_data = lgb.basic._data_from_pandas(df, 'auto', None, None)[0]
     assert built_data.dtype == dtype
     assert np.shares_memory(X, built_data)
 
@@ -659,7 +657,7 @@ def test_categorical_code_conversion_doesnt_modify_original_data():
     pd = pytest.importorskip('pandas')
     X = np.random.choice(['a', 'b'], 100).reshape(-1, 1)
     df = pd.DataFrame(X.copy(), columns=['x1'], dtype='category')
-    data = lgb.basic._data_from_pandas(df, ['x1'], None, None)[0]
+    data = lgb.basic._data_from_pandas(df, 'auto', None, None)[0]
     # check that the original data wasn't modified
     np.testing.assert_equal(df['x1'], X[:, 0])
     # check that the built data has the codes
--- a/tests/python_package_test/test_basic.py
+++ b/tests/python_package_test/test_basic.py
@@ -644,20 +644,22 @@ def test_custom_objective_safety():
 
 
 @pytest.mark.parametrize('dtype', [np.float32, np.float64])
-def test_no_copy_when_single_float_dtype_dataframe(dtype):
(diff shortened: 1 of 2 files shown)


jameslamb:
@jmoralez I want to remind you that every PR needs to have a label in https://github.com/microsoft/LightGBM/blob/59f59c9fbe060eae6114b21d6aa5f7b0ad80b4e7/.github/release-drafter.yml, so that it can be correctly categorized in the release notes.

tests/python_package_test/test_basic.py
```diff
@@ -648,9 +648,7 @@ def test_no_copy_when_single_float_dtype_dataframe(dtype):
     pd = pytest.importorskip('pandas')
     X = np.random.rand(10, 2).astype(dtype)
     df = pd.DataFrame(X)
-    # feature names are required to not make a copy (rename makes a copy)
```
jameslamb, line 651: This test currently tests "when you pass a `pandas` DataFrame and explicit list of feature names, a copy is not made".

This PR is proposing changing it to "when you pass a `pandas` DataFrame and `feature_names="auto"`, a copy is not made".

To test the broader behavior "when you pass a `pandas` DataFrame, a copy is not made, regardless of how you specify feature names", would you consider instead changing the tests touched in this PR such that both cases (`feature_name` and `"auto"`) are tested?

jameslamb (changes requested):
Thanks for the very thorough investigation with profiling!

I have one question...does this change mean that `lightgbm` will now *modify* the input `pandas` DataFrame in place? Like if I bring a `pandas` DataFrame with integer column names to `lightgbm`, and I use `feature_name="auto"`, will those now be changed to strings?

If no, then I totally support this change.

If yes then I still support it (I think the memory savings is worth the potential challenges caused by silent side-effects like changing the type of column names), but think we should label this PR `breaking` and maybe include a documentation change and/or warning in that situation. I think "changing your DataFrame's column names when `lightgbm` previously used to leave it unmodified" is a breaking change that could affect users' code that works on that DataFrame post-training.

jmoralez:
> does this change mean that lightgbm will now modify the input pandas DataFrame in place?

No, the argument `copy` to `rename` is actually passed as the `deep` argument to `DataFrame.copy`, links here: https://github.com/microsoft/LightGBM/issues/4596#issuecomment-1133784728. So it will still make a copy but a shallow one, i.e. it won't copy the underlying arrays and the index, but things like renaming the columns will only modify the copy.

jameslamb (approved):
Changes look great, and thanks for the answer to my question about whether the rename will have side effects outside of LightGBM.

Looks good!

StrikerRUS (approved):
Thanks a lot for the nice improvement and test enhancements!

PR #5255, opened 2022-05-31
[python] add more type hints in basic.py

jameslamb:
Contributes to #3756.

Proposes adding a few more hints in `basic.py`. Just quickly scanned through this file tonight and chose a few hints that looked easy to determine just from looking at the code and docstrings, trying to slowly make progress on #3756 as I have time.

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -3957,7 +3957,7 @@ class Booster:
                     name.startswith(('auc', 'ndcg@', 'map@', 'average_precision')) for name in self.__name_inner_eval
                 ]
 
-    def attr(self, key: str) -> Any:
+    def attr(self, key: str) -> Optional[str]:
         """Get attribute string from the Booster.
 
         Parameters
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -3684,7 +3684,7 @@ class Booster:
         predictor.pandas_categorical = self.pandas_categorical
         return predictor
 
-    def num_feature(self):
+    def num_feature(self) -> int:
         """Get number of features.
 
         Returns
@@ -3698,7 +3698,7 @@ class Booster:
             ctypes.byref(out_num_feature)))
         return out_num_feature.value
 
-    def feature_name(self):
+    def feature_name(self) -> List[str]:
         """Get names of features.
 
         Returns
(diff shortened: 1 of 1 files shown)


python-package/lightgbm/basic.py
```diff
@@ -3953,7 +3957,7 @@ def __get_eval_info(self):
                     name.startswith(('auc', 'ndcg@', 'map@', 'average_precision')) for name in self.__name_inner_eval
                 ]
 
-    def attr(self, key):
+    def attr(self, key: str) -> Any:
```
StrikerRUS, line 3960:
```suggestion
    def attr(self, key: str) -> Optional[str]:
```

jameslamb: Oh interesting, I didn't realize this could only be set by `Booster.set_attr()`.

Can you help me understand the purpose of these methods? It seems they were added back in 2016  (https://github.com/microsoft/LightGBM/commit/eba6d200a81776b0062971ee3f03a5486fc97825#diff-5e0fb2a7b1ec4988ccde5e74a6c4b609841c1143f281da2afc85e8c464d37b3b) and today are only used in one place, a test.

```shell
git grep -E '\.attr\('
```

https://github.com/microsoft/LightGBM/blob/1cc9f9dcee15586aefdd8271a1c04ad010d3c53a/tests/python_package_test/test_sklearn.py#L483

```shell
git grep -E '\.set_attr\('
```

https://github.com/microsoft/LightGBM/blob/1cc9f9dcee15586aefdd8271a1c04ad010d3c53a/tests/python_package_test/test_sklearn.py#L32

jameslamb: I agree either way about the type hint so I don't think this conversation needs to block this PR. But I'm wondering if there's an opportunity here to remove these methods and the internal object `Booster.__attr`.

StrikerRUS: > Can you help me understand the purpose of these methods?

I'm sorry, I can't.

I *guess* it's a legacy code that was initially borrowed from XGBoost. In XGBoost those methods are used for distributed training with Rabit: https://github.com/dmlc/xgboost/commit/ecb3a271bed151252fb048528ce5a90ad75bb68f.

StrikerRUS: I have no objection to remove `set_attr()` and `attr()` methods in a separate PR to make LightGBM codebase clearer.

StrikerRUS (approved):
LGTM!

python-package/lightgbm/basic.py
jameslamb: Ok thanks! I'll make a separate PR that proposes removing them.

jameslamb: created #5272

PR #5259, opened 2022-06-01
Support compiling with ClangCL and Visual Studio

buresu:
build with the following commands.
```
cmake .. -G "Visual Studio 16 2019" -T ClangCL -A x64
cmake --build . --config Release
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -498,7 +498,7 @@ if(USE_MPI)
 endif()
 
 if(USE_OPENMP)
-  if(CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang")
+  if(CMAKE_CXX_COMPILER_ID MATCHES "Clang|AppleClang")
     target_link_libraries(lightgbm_objs PUBLIC OpenMP::OpenMP_CXX)
     # c_api headers also includes OpenMP headers, thus compiling
     # lightgbm_capi_objs needs include directory for OpenMP.


jameslamb (commented):
Thanks very much for this proposal! I hadn't heard of `clang-cl` until today. Will need to do a bit of research on this, and maybe test with a new CI job.

Links I'm finding useful (for the benefit of other reviewers):

* https://gitlab.kitware.com/cmake/cmake/-/issues/22668
* https://gitlab.kitware.com/cmake/cmake/-/issues/20415#note_707011
* https://stackoverflow.com/a/64189112/3986677
* https://docs.microsoft.com/en-us/cpp/build/clang-support-msbuild?view=msvc-170

buresu:
Thank you for reviewing.
This is a common clang build problem, but I can't reproduce on linux clang, so I exemplified on visual studio clang.

StrikerRUS:
What I've learned from links in https://github.com/microsoft/LightGBM/pull/5259#pullrequestreview-995646690 so far (thanks a lot @jameslamb for them!):
- Only Visual Studio 2019 and newer is supported;
- `ClangCL` is a toolset value, so it can be added in 
   https://github.com/microsoft/LightGBM/blob/1cc9f9dcee15586aefdd8271a1c04ad010d3c53a/python-package/setup.py#L172

jameslamb:
> This is a common clang build problem, but I can't reproduce on linux clang

This project is tested with `clang` on Linux and macOS on every commit, so I'd be surprised to learn that an issue of the form "LightGBM cannot be compiled with any version of `clang`" exists.

https://github.com/microsoft/LightGBM/blob/27d9ad2e8e7ea2d00dbab7b17b0a34b5d9184fe0/.vsts-ci.yml#L88-LL91

https://github.com/microsoft/LightGBM/blob/27d9ad2e8e7ea2d00dbab7b17b0a34b5d9184fe0/.vsts-ci.yml#L225-LL228

https://github.com/microsoft/LightGBM/blob/27d9ad2e8e7ea2d00dbab7b17b0a34b5d9184fe0/.ci/test.sh#L3-L9

https://github.com/microsoft/LightGBM/blob/27d9ad2e8e7ea2d00dbab7b17b0a34b5d9184fe0/.github/workflows/python_package.yml#L60-LL66

That is probably why you haven't been able to reproduce this on Linux. I suspect the issue you've faced is specific to using `ClangCL` + `MSVC` on Winddows.

buresu:
@jameslamb I beleave this problem is also occurd on linux (in some cases).
Clang openmp requires a link path to libomp, but on linux libomp is usually installed on the system (e.g. libomp-dev).
Because this path is already added to link path, linker can find libomp.
If we use libomp outside the system path, I think we need to a link path to libomp.
I think apple clang probably uses libomp installed by brew, which causes same problem.

buresu:
After noticing the above, I built on the vc prompt and the build went through without this change. 
Sorry for the confusion. I will close this PR.

jameslamb:
@buresu sorry, could you clarify? Are you saying that the following worked for you with with 0 modifications to LightGBM?

```shell
cmake .. -G "Visual Studio 16 2019" -T ClangCL -A x64
cmake --build . --config Release
```

Or did you have to modify `CMakeLists.txt`  or something in `python-package/` to compile successfully?

buresu:
@jameslamb Yes, the build successful with zero changes using vc prompt.
When using powershell the build succeeded with this PR change, but it was linking llvm openmp installed by scoop. (This is my mistake).
When built using powershell with no changes, vc's openmp was not found.
I think this is a lack of support from cmake.

jameslamb:
ahhh ok, thanks for the additional information @buresu !

PR #5267, opened 2022-06-05
[tests][python] Make test that checks original pandas data isn't modified more strict

StrikerRUS:
Sorry, should have proposed that in #5254.

Check that `_data_from_pandas()` doesn't modify column names (doesn't convert int to string) of original data.
Refer to https://github.com/microsoft/LightGBM/pull/5254#pullrequestreview-989794258.

### DIFF
--- a/tests/python_package_test/test_basic.py
+++ b/tests/python_package_test/test_basic.py
@@ -654,12 +654,11 @@ def test_no_copy_when_single_float_dtype_dataframe(dtype, feature_name):
     assert np.shares_memory(X, built_data)
 
 
-@pytest.mark.parametrize('feature_name', [[42], 'auto'])
+@pytest.mark.parametrize('feature_name', [['x1'], [42], 'auto'])
 def test_categorical_code_conversion_doesnt_modify_original_data(feature_name):
     pd = pytest.importorskip('pandas')
     X = np.random.choice(['a', 'b'], 100).reshape(-1, 1)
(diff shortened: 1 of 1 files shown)


tests/python_package_test/test_basic.py
```diff
@@ -654,16 +654,18 @@ def test_no_copy_when_single_float_dtype_dataframe(dtype, feature_name):
     assert np.shares_memory(X, built_data)
 
 
-@pytest.mark.parametrize('feature_name', [['x1'], 'auto'])
+@pytest.mark.parametrize('feature_name', [[42], 'auto'])
```
jameslamb, line 657:
```suggestion
@pytest.mark.parametrize('feature_name', [['x1'], [42], 'auto'])
```

Instead of replacing a string with a number here, would you consider *adding* a case here? That way, this test would still cover the case "`feature_name` provided, and names are strings".

jameslamb (approved):
Thanks for doing this! I left one suggestion for your consideration, but marking "approve" because I'm fine with whichever you decide about it.

tests/python_package_test/test_basic.py
StrikerRUS: Thanks for the proposal! Added in 2b107f6d883f0931456cbdb973fbe0ef626d0f0d.

jameslamb: changes look great, thank you!

PR #5271, opened 2022-06-07
[python-package] add type hints on cv()

jameslamb:
Contributes to #3756.

This PR proposes adding type hints to `lightgbm.cv()`, to improve `mypy`'s ability to catch issues.

### DIFF
--- a/python-package/lightgbm/compat.py
+++ b/python-package/lightgbm/compat.py
@@ -75,9 +75,9 @@ try:
     from sklearn.utils.validation import assert_all_finite, check_array, check_X_y
     try:
         from sklearn.exceptions import NotFittedError
-        from sklearn.model_selection import GroupKFold, StratifiedKFold
+        from sklearn.model_selection import BaseCrossValidator, GroupKFold, StratifiedKFold
     except ImportError:
-        from sklearn.cross_validation import GroupKFold, StratifiedKFold
+        from sklearn.cross_validation import BaseCrossValidator, GroupKFold, StratifiedKFold
         from sklearn.utils.validation import NotFittedError
     try:
         from sklearn.utils.validation import _check_sample_weight
@@ -90,6 +90,7 @@ try:
             return sample_weight
 
     SKLEARN_INSTALLED = True
+    _LGBMBaseCrossValidator = BaseCrossValidator
     _LGBMModelBase = BaseEstimator
     _LGBMRegressorBase = RegressorMixin
     _LGBMClassifierBase = ClassifierMixin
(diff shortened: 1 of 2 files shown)


python-package/lightgbm/engine.py
```diff
@@ -373,12 +378,25 @@ def _agg_cv_result(raw_results):
...
-       folds=None, nfold=5, stratified=True, shuffle=True,
-       metrics=None, feval=None, init_model=None,
-       feature_name='auto', categorical_feature='auto',
-       fpreproc=None, seed=0, callbacks=None, eval_train_metric=False,
-       return_cvbooster=False):
+def cv(
+    params: Dict[str, Any],
+    train_set: Dataset,
+    num_boost_round: int = 100,
+    folds: Optional[Union[Iterable[Tuple[np.ndarray, np.ndarray]], _LGBMBaseCrossValidator]] = None,
```
jameslamb, line 385: according to https://docs.python.org/3/library/typing.html#typing.Generator, it's ok to use `Iterable[return_type]` to document a generator or iterator. So it isn't necessary to use `typing.Generator` or `typing.Iterator`.

python-package/lightgbm/compat.py
```diff
@@ -75,9 +75,9 @@ def __init__(self, *args, **kwargs):
     from sklearn.utils.validation import assert_all_finite, check_array, check_X_y
     try:
         from sklearn.exceptions import NotFittedError
-        from sklearn.model_selection import GroupKFold, StratifiedKFold
+        from sklearn.model_selection import BaseCrossValidator, GroupKFold, StratifiedKFold
```
jameslamb, line 78: `sklearn.model_selection.BaseCrossValidator` has been in `scikit-learn` for at least 7 years, so I don't think this will cause any portability issues.

https://github.com/scikit-learn/scikit-learn/blame/159cb46c5672b0ba87d0ba80c6b80ec1aa5fda32/sklearn/model_selection/__init__.py#L3

jmoralez (approved):

StrikerRUS (changes requested):
Just two comments.

python-package/lightgbm/engine.py
```diff
@@ -4,19 +4,24 @@
...
 from .basic import Booster, Dataset, LightGBMError, _choose_param_value, _ConfigAliases, _InnerPredictor, _log_warning
-from .compat import SKLEARN_INSTALLED, _LGBMGroupKFold, _LGBMStratifiedKFold
+from .compat import SKLEARN_INSTALLED, _LGBMBaseCrossValidator, _LGBMGroupKFold, _LGBMStratifiedKFold
 
 _LGBM_CustomMetricFunction = Callable[
     [np.ndarray, Dataset],
     Tuple[str, float, bool]
 ]
 
+_LGBM_PreprocCallable = Callable[
```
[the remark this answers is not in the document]
jameslamb: ah yeah ok, makes sense

StrikerRUS (approved):
LGTM!

PR #5272, opened 2022-06-07
[python-package] remove Booster.set_attr() and Booster.attr()

jameslamb:
Proposes removing `Booster.set_attr()` and `Booster.attr()`.

Per https://github.com/microsoft/LightGBM/pull/5255#discussion_r889648818 those functions are not used anywhere. It looks like they may have been copied from `xgboost` around when LightGBM's Python package was first created.

This PR proposes removing them to simplify the interface for `Booster` and reduce the maintenance burden of the project.

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -2570,7 +2570,6 @@ class Booster:
         self.network = False
         self.__need_reload_eval_info = True
         self._train_data_name = "training"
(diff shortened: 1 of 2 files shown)


guolinke (approved):
LGTM

StrikerRUS (approved):
LGTM, thanks for simplifying the codebase!

PR #5273, opened 2022-06-07
[R-package] [docs] document difference between lightgbm() and lgb.train() (fixes #5203)

### LINKED ISSUE #5203: [RFC] [R-package] decouple verbosity and validation sets in lightgbm()
## Summary

`lightgbm::lightgbm()` in the R package currently has the behavior "if logging verbosity is set higher than 0, add the training data to the validation sets used to compute eval metrics and for early stopping".

This behavior should be removed.

## Motivation

I think this coupling of "what level of logs do I want?" and "what validation sets do I want to use?" is confusing and surprising.

## Description

To do this, I think the following condition should be removed:

https://github.com/microsoft/LightGBM/blob/14ca8fc07e3411fac215c4009ed47aa019fad980/R-package/R/lightgbm.R#L196-L199

And `valids` should be promoted to a top-level keyword argument.

This is a breaking change, and if it's done should be done prior to the v4.0.0 release (#5153).

## References

Created based on https://github.com/microsoft/LightGBM/pull/5199#discussion_r868583048.

jameslamb:
Fixes #5203.

This PR proposes making the explanation in https://github.com/microsoft/LightGBM/issues/5203#issuecomment-1124502018 explicit in the R package's documentation.

### Notes for Reviewers

The main *functional* change proposed in #5203 (decoupling `verbosity` and validation sets in `lightgbm::lightgbm()`) was implemented in #5209.

### DIFF
--- a/R-package/R/lightgbm.R
+++ b/R-package/R/lightgbm.R
@@ -92,7 +92,7 @@ NULL
 #' @description High-level R interface to train a LightGBM model. Unlike \code{\link{lgb.train}}, this function
 #'              is focused on compatibility with other statistics and machine learning interfaces in R.
 #'              This focus on compatibility means that this interface may experience more frequent breaking API changes
-#'              than \code{link{lgb.train}}.
+#'              than \code{\link{lgb.train}}.
 #'              For efficiency-sensitive applications, or for applications where breaking API changes across releases
 #'              is very expensive, use \code{\link{lgb.train}}.
 #' @inheritParams lgb_shared_params
--- a/R-package/R/lightgbm.R
+++ b/R-package/R/lightgbm.R
@@ -89,7 +89,11 @@ NULL
 
 #' @name lightgbm
 #' @title Train a LightGBM model
-#' @description Simple interface for training a LightGBM model.
+#' @description High-level R interface to train a LightGBM model. Unlike \code{\link{lgb.train}}, this function
+#'              is focused on compatibility with other statistics and machine learning interfaces in R.
+#'              It is optimized for compatibility and usability, sometimes at the expense of efficiency or API stability.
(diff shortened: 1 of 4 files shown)


guolinke (approved):

R-package/R/lightgbm.R
```diff
@@ -89,7 +89,12 @@ NULL
 
 #' @name lightgbm
 #' @title Train a LightGBM model
-#' @description Simple interface for training a LightGBM model.
+#' @description High-level R interface to train a LightGBM model. Unlike \code{\link{lgb.train}}, this function
+#'              is focused on compatibility with other statistics and machine learning interfaces in R.
+#'              This focus on compatibility means that this interface may experience more frequent breaking API changes
+#'              than \code{link{lgb.train}}.
```
StrikerRUS, line 95:
```suggestion
#'              than \code{\link{lgb.train}}.
```

R-package/man/lightgbm.Rd
```diff
@@ -99,7 +99,12 @@ the "objective" item of the "Parameters" section of the documentation}.}
 a trained \code{lgb.Booster}
 }
 \description{
-Simple interface for training a LightGBM model.
+High-level R interface to train a LightGBM model. Unlike \code{\link{lgb.train}}, this function
+             is focused on compatibility with other statistics and machine learning interfaces in R.
+             This focus on compatibility means that this interface may experience more frequent breaking API changes
+             than \code{link{lgb.train}}.
```
StrikerRUS, line 105:
```suggestion
             than \code{\link{lgb.train}}.
```

StrikerRUS (commented):
I would also like to wait @jmoralez's review before merging, because I'm not an R user at all. 🙂

jmoralez (approved):

PR #5274, opened 2022-06-07
WIP: [ci] use R 4.2 in Windows R-package CI jobs (fixes #4881)

### LINKED ISSUE #4881: [R-package] Support R 4.2
## Summary

R 4.2 is going to include some significant changes to how packages are built on Windows. I've created this "feature request" to document those changes in this project's backlog and so that we have something to link conversations to if users report issues on R 4.2.x, but R 4.2 has not yet been released.

## Motivation

`{lightgbm}`, the R package for LightGBM, should continue to work with new versions of R.

## Description

See https://developer.r-project.org/Blog/public/2021/12/07/upcoming-changes-in-r-4.2-on-windows/. This was shared by Tomas Kalibera on the r-pkg-devel mailing list.

Some notable quotes

> R 4.2 for Windows will support UTF-8 as native encoding

> ...will require at least Windows 10 (version 1903) on desktop systems, Windows Server 2022 on long-term support server systems or Windows Server 1903 from the semi-annual channel

> R 4.2 will also drop support for 32-bit builds on Windows

> A new compiler toolchain, RTools42, will be used for building Windows binaries of R and R packages from source

According to that blog post, there is now support for testing on R Hub with the UCRT toolchain. See https://builder.r-hub.io/advanced.

Screen Shot 2021-12-11 at 10 57 30 PM

So I think it would be useful to add a new comment-triggered workflow, similar to the existing Solaris one.

https://github.com/microsoft/LightGBM/blob/80662618996431c222e6c92cbba7e784f8a935cd/.github/workflows/r_solaris.yml

## References

Since March of this year, CRAN has been running checks with this toolchain on existing packages. `{lightgbm}` is currently passing those checks without WARNINGs or ERRORs!

Screen Shot 2021-12-11 at 11 00 18 PM

jameslamb:
Fixes #4881.

Attempts to add support for R 4.2 in most of the R package's CI jobs.

### Notes for Reviewers

#4881 refers to creating a comment-triggered workflow for the UCRT toolchain for Windows builds, but I don't think that's necessary any more now that R 4.2.0 has actually been released and Rtools42 contains the UCRT build tools.

References for some concerns about Windows builds:

* Rtools42
    - https://cran.r-project.org/bin/windows/Rtools/rtools42/rtools.html
* CRAN still requires 32-bit builds for `r-oldrel` (which is R 4.1)
    - https://cran.r-project.org/web/checks/check_flavors.html#r-devel-windows-x86_64
* different paths for different versions, per `r-lib/actions` code
    - https://github.com/r-lib/actions/blob/8e8f7edba0e5bda9318f9377c0dc9c2f5d0fb786/setup-r/src/installer.ts#L431-L457

### References

https://github.com/r-lib/actions/issues/574#issue-1265095474

https://cran.r-project.org/bin/windows/base/howto-R-4.2.html

> To make the use of Rtools42 simpler, when R is installed via the binary installer it by default uses Rtools42 for the compilers and libraries. PATH will be set by R (inside front-ends like RGui and RTerm, but also R CMD) to include the build tools (e.g. make) and the compilers (e.g. gcc). In addition, R installed via the binary installer will automatically set R_TOOLS_SOFT (and LOCAL_SOFT for backwards compatibility) to the Rtools42 location for building R packages. This feature is only present in the installer builds of R, not when R is installed from source.

Related fix in `r-lib/actions`: https://github.com/r-lib/actions/commit/7d98418c6b5cfca592b100e2f348fbe90c18bcf3


StrikerRUS:
I'd like to propose checking passed R version to the `FindLibR.cmake` module from the logs
https://github.com/microsoft/LightGBM/blob/f1328d5c5fee1da5907371e292dda3e8344b4b63/cmake/modules/FindLibR.cmake#L76
similarly to
https://github.com/microsoft/LightGBM/blob/f1328d5c5fee1da5907371e292dda3e8344b4b63/.ci/test_r_package.sh#L237-L253

That will prevent situations like https://github.com/microsoft/LightGBM/pull/5129#issuecomment-1088190413.

jameslamb:
ooooo, I think that the `inet_pton` detection check in `configure.win` might be part of the problem!

I tried running the following shell script in the CI builds for the Windows jobs.

```sh
echo "------ testing test program ----"

cat > conftest.cpp <<EOL
#include <ws2tcpip.h>
int main() {
  void* p = inet_pton;
  return 0;
}
EOL

g++ -std=gnu++11 -I"c:/rtools42/x86_64-w64-mingw32.static.posix/include" -o conftest conftest.cpp || exit 123
echo "------ done compiling ----"

./conftest || exit 123
rm -f ./conftest
rm -f ./conftest.cpp
echo "------ done running ----"
```

And got the following

```text
------ testing test program ----
conftest.cpp: In function 'int main()':
conftest.cpp:3:13: error: invalid conversion from 'INT (*)(INT, LPCSTR, PVOID)' {aka 'int (*)(int, const char*, void*)'} to 'void*' [-fpermissive]
    3 |   void* p = inet_pton;
      |             ^~~~~~~~~
      |             |
      |             INT (*)(INT, LPCSTR, PVOID) {aka int (*)(int, const char*, void*)}
Building R package
```

To me, that looks like `inet_pton` was FOUND, but was incorrectly detected as NOT FOUND because this code can't be compiled.

jameslamb:
This also seems relevant for the CMake side: https://cmake.org/cmake/help/latest/module/CheckCXXSymbolExists.html

> `CheckCXXSymbolExists` command is unreliable when <symbol> is (potentially) an overloaded function. Since there is no reliable way to predict whether a given function in the system environment may be defined as an overloaded function or may be an overloaded function on other systems or will become so in the future, it is generally advised to use the [CheckCXXSourceCompiles](https://cmake.org/cmake/help/latest/module/CheckCXXSourceCompiles.html#module:CheckCXXSourceCompiles) module for checking any function symbol (unless somehow you surely know the checked function is not overloaded on other systems or will not be so in the future).

jameslamb:
Closing this in favor of #5503 , which includes the fixes discovered through this investigation.

PR #5286, opened 2022-06-13
[ci] Install PoCL in Docker for old Ubuntu 14.04

StrikerRUS:
In cooperation with https://github.com/guolinke/lightgbm-ci-docker/pull/26 (that PR should be merged first, already merged).

### DIFF
--- a/.ci/setup.sh
+++ b/.ci/setup.sh
@@ -74,17 +74,6 @@ else  # Linux
         if [[ $IN_UBUNTU_LATEST_CONTAINER == "true" ]]; then
             sudo apt-get install --no-install-recommends -y \
                 pocl-opencl-icd
-        else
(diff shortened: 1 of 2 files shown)


StrikerRUS:
GPU job fails with Clang (error: [`ImportError: dlopen: cannot load any more object with static TLS`](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=12909&view=logs&j=9ce42e5d-f31a-544a-a6de-dc42f8c013ed&t=b696def5-78ae-590c-4234-04332342ba2d)).

Current `master` also fails with Clang but with different error: [`.Fatal Python error: Segmentation fault`](https://github.com/microsoft/LightGBM/issues/3475).

So I don't think this should be a blocker.

jameslamb (approved):
Very nice, thanks!

PR #5289, opened 2022-06-14
[python] preserve None in _choose_param_value()

jameslamb:
Addresses https://github.com/microsoft/LightGBM/pull/5105/files#r889683078.

`_choose_param_value()` in the Python package is used to ensure that parameters with multiple possible values are resolved with the following precedence rules:

1. main parameter name found in `params`
2. alias found in `params`
3. a default value (e.g. the type sometimes provided by keyword arguments from other interfaces)

As noted in the linked comment, that function currently does not preserve a literal `None` in some situations. This PR fixes that.


python-package/lightgbm/basic.py
```diff
@@ -408,21 +408,27 @@ def _choose_param_value(main_param_name: str, params: Dict[str, Any], default_va
...
+            break
 
-    if found_value is not None:
-        params[main_param_name] = found_value
-    else:
-        params[main_param_name] = default_value
+    if main_param_name in params.keys():
+        for param in aliases:
+            params.pop(param, None)
+        return params
```
jmoralez, line 428: Isn't this the same as lines 414-417?

python-package/lightgbm/basic.py
```diff
@@ -408,21 +408,27 @@ def _choose_param_value(main_param_name: str, params: Dict[str, Any], default_va
...
 
-    for param in _ConfigAliases.get(main_param_name):
-        val = params.pop(param, None)
-        if found_value is None and val is not None:
-            found_value = val
+    # if main param name was not found, search for an alias
+    for param in aliases:
```
jmoralez, line 420: This is probably for a different PR but shouldn't we try to match the behavior of the C++ side with respect to aliases and try to sort them using the same criteria than here? https://github.com/microsoft/LightGBM/blob/fc0c8fd4521ab4814fe3f59999115b70984e76c1/include/LightGBM/config.h#L1141-L1144

python-package/lightgbm/basic.py
jameslamb: If `main_param_name in params.keys():` is true in this block, it means "a value was provided via an alias". This needs to be here to clear out the other aliases.

That's different from lines 414-417, which is like "a value was provided via the main parameter name".

jmoralez: Ah I see. Maybe after not finding the `main_param_name` we could do the following:
```
aliases_in_params = list(params.keys() & aliases)
# if any alias was found, set the value of the first one
if aliases_in_params:
    params[main_param_name] = params[aliases_in_params[0]]
# remove remaining aliases
for aliases_in_params[1:]:
    params.pop(param, None)
```
I'm approving since this is just a suggestion. Thanks for the explanation!

jmoralez (approved):

python-package/lightgbm/basic.py
```diff
@@ -408,21 +408,27 @@ def _choose_param_value(main_param_name: str, params: Dict[str, Any], default_va
     # avoid side effects on passed-in parameters
     params = deepcopy(params)
 
-    # find a value, and remove other aliases with .pop()
-    # prefer the value of 'main_param_name' if it exists, otherwise search the aliases
-    found_value = None
+    aliases = set(_ConfigAliases.get(main_param_name)) - {main_param_name}
```
StrikerRUS, line 411: This is already `set`
https://github.com/microsoft/LightGBM/blob/2f5baa3d39efb518cd13a7932fe4d8602c36762f/python-package/lightgbm/basic.py#L369
```suggestion
    aliases = _ConfigAliases.get(main_param_name) - {main_param_name}
```

StrikerRUS (approved):
Thank you for refactoring this!

I'm +1 for consistent alias choosing method with cpp code.

python-package/lightgbm/basic.py
jameslamb: ah right! I forgot, thank you very much for the reminder.

python-package/lightgbm/basic.py
jameslamb: @StrikerRUS @jmoralez I just created #5304 to track the work of making alias-resolution consistent and reproducible.

Please add any other details there that you think I might have missed.

PR #5290, opened 2022-06-14
[ci] Pin lintr to <3.0

jgiannuzzi:
Lint jobs started to fail in the CI because a new version of `r-lintr` was released. This PR pins the version to be 2.x.


jameslamb (approved):
Oy, sorry about that!

Note for other maintainers: now that `{lintr}` 3.0.0 is out on CRAN, I'll try to put up a PR in the next few days to revert this change (linking #5228).

PR #5292, opened 2022-06-15
[ci][dask][gpu] Run Dask tests with LightGBM GPU version

jgiannuzzi:
As it was suggested in https://github.com/microsoft/LightGBM/pull/5282#issue-1268671683 (previous attempt: #5285).

Using findings from https://github.com/microsoft/LightGBM/pull/5282#issuecomment-1155440483:
> My understanding is that the `dask` tests only fail with this error on Ubuntu 14.04 with Python 3.10. I ran them locally with Python 3.8 and 3.9 and they work fine. Do we maybe want to set `PYTHON_VERSION` again on the Linux `gpu_source` job?
>
> On Ubuntu 20.04, the tests seem to fail because of a timeout. Enabling those tests on a fake GPU is expected to take more time. Maybe setting `timeoutInMinutes` like we do for the aarch64 job would help?


jameslamb:
@jgiannuzzi I just updated this branch with the changes from latest `master`. I'm hoping that these changes that have accumulated since this PR was first opened will allow us to re-enable the Dask tests for `gpu` builds:

* #5396
* #5544
* #5555
* #5580

Thanks for experimenting with this and pushing it forward! The Dask tests are the most comprehensive tests on distributed training that this project has, so enabling them on more environment and installation types is really valuable to the project.

.vsts-ci.yml
```diff
@@ -36,6 +36,7 @@ jobs:
     PRODUCES_ARTIFACTS: 'true'
   pool: sh-ubuntu
   container: linux-artifact-builder
+  timeoutInMinutes: 60
```
jameslamb, line 39:
```suggestion
```

.vsts-ci.yml
```diff
@@ -89,6 +90,7 @@ jobs:
     SETUP_CONDA: 'true'
   pool: sh-ubuntu
   container: ubuntu-latest
+  timeoutInMinutes: 60
```
jameslamb, line 93:
```suggestion
```

jameslamb (approved):
Happy to say that this is working!

I think that the timeout observed was actually not about the tests taking too long to run, but was related some combination of:

* the library-loading issues described in #5584, and fixed by upgrading the `Linux_*` jobs to an environment with a newer GLIBC in #5580
* the network-cleanup strategy in the Python package (https://github.com/microsoft/LightGBM/pull/5510#issuecomment-1270990841), fixed by #5510

Given that, I pushed https://github.com/microsoft/LightGBM/pull/5292/commits/06c2d3b82d073443c2f5c364fc89ce3cea2d256b removing the proposed timeouts for Azure DevOps from this PR.

The GPU jobs are taking less than 30 mins ([build link](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=14101&view=results)).

Screen Shot 2023-01-02 at 7 38 28 PM

Thanks very much for improving the test coverage @jgiannuzzi !

PR #5300, opened 2022-06-19
[R-package] Specify concrete linting rule that is being ignored via `nolint` mark

StrikerRUS:
Refer to https://github.com/microsoft/LightGBM/pull/5249#discussion_r889447818.

Continuation of #5294.


jameslamb (approved):
Excellent, thank you!

PR #5301, opened 2022-06-19
[python-package] add more type hints in engine.py

jameslamb:
Contributes to #3756.

This PR proposes adding more type hints in `engine.py`. I believe this finishes the `engine.py` work for #3756.


python-package/lightgbm/engine.py
```diff
@@ -365,7 +374,9 @@ def _make_n_folds(full_data, folds, nfold, params, seed, fpreproc=None, stratifi
     return ret
 
 
-def _agg_cv_result(raw_results):
+def _agg_cv_result(
+    raw_results: List[Tuple[str, str, float, bool]]
```
StrikerRUS, line 378: https://github.com/microsoft/LightGBM/blob/706540483a142efdb5be4d2670c1ebd75704ce7b/python-package/lightgbm/engine.py#L594
I think inner list is from `eval_valid()`
https://github.com/microsoft/LightGBM/blob/706540483a142efdb5be4d2670c1ebd75704ce7b/python-package/lightgbm/basic.py#L3261
https://github.com/microsoft/LightGBM/blob/706540483a142efdb5be4d2670c1ebd75704ce7b/python-package/lightgbm/basic.py#L3285-L3288
and the outer one is from `__getattr__()`:
https://github.com/microsoft/LightGBM/blob/706540483a142efdb5be4d2670c1ebd75704ce7b/python-package/lightgbm/engine.py#L297
https://github.com/microsoft/LightGBM/blob/706540483a142efdb5be4d2670c1ebd75704ce7b/python-package/lightgbm/engine.py#L301-L304
```suggestion
    raw_results: List[List[Tuple[str, str, float, bool]]]
```

jameslamb: Ah yep, you are absolutely right. I was using a debugger to check some of these types in complicated code paths with many levels of function calls, and made a mistake in where I place the debugger and how I interpreted its results.

StrikerRUS (approved):
Thank you!

Issue #5303, opened 2022-06-19
[ci] [R-package] Add new linters from {lintr} 3.0


jameslamb:
Now that `{lintr}` 3.0 has been released and this project is using it (#5294), I believe the project should take advantage of additional linters added in that new version.

These are documented in "New and improved features" at https://cran.r-project.org/web/packages/lintr/news/news.html.

Reading through those new linters, I believe at least the following would be useful to add  in LightGBM's CI:

* [x] `any_duplicated_linter()` (#5311)
* [x] `any_is_na_linter()` (#5306)
* [x] `class_equals_linter()` (#5307)
* [x] `duplicate_argument_linter()` (#5310)
* [x] `inner_combine_linter()` (#5311)
* [x] `literal_coercion_linter()` (#5311)
* [x] `missing_argument_linter()` (#5311)
* [x] `numeric_leading_zero_linter()` (#5311)
* [x] `outer_negation_linter()` (#5311)
* [x] `package_hooks_linter()` (#5311)
* [x] `paste_linter()` (#5320)
* [x] `sprintf_linter()` (#5311)
* [x] `regex_subset_linter()` (#5311)
* [x] `string_boundary_linter()` (#5324)
* [x] `unreachable_code_linter()` (#5311)
* [x] `vector_logic_linter()` (#5311)

### Motivation

Pushing more of the work of standardizing the project's R code into automation reduces the effort required by both reviewers and contributors in pull request reviews.

Some of these checks are not just about style...they catch efficiency or correctness issues. Doing so automatically reduces the risk of such issues being missed by contributors and reviewers.

### Approach

Pull requests for this issue are welcomed from both maintainers and outside contributors!  Please only do one or two linters per pull request, to make them easier to review and spread the work out more evenly over time.

SaumyaBhushan:
Can you please tell me how can I do this ?

jameslamb:
> Can you please tell me how can I do this ?

@SaumyaBhushan you can see #5306 for a good example, and the comments in https://github.com/microsoft/LightGBM/pull/5307#pullrequestreview-1011638019 for more details on how to make a successful pull request for this issue.

SaumyaBhushan:
Thanks @jameslamb for the help.

jameslamb:
This work is now complete. Thanks very much @CuriousCorrelation and @SaumyaBhushan ! We'd welcome additional contributions in the future if you're interested!

PR #5307, opened 2022-06-19
[ci] [R-package] Add class_equals linter

CuriousCorrelation:
Once merged, closes `class_equals_linter()` task of https://github.com/microsoft/LightGBM/issues/5303

Resolves linter warnings raised by {lintr} 3.0.0

``` sh
[[1]]
/home/ccr-gentoo/Projects/LightGBM/R-package/R/utils.R:2:14: warning: [class_equals_linter] Instead of comparing class(x) with %in%, use inherits(x, 'class-name') or is.<class> or is(x, 'class')
  return(all(c("R6", "lgb.Booster") %in% class(x)))
             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

[[2]]
/home/ccr-gentoo/Projects/LightGBM/R-package/R/utils.R:6:14: warning: [class_equals_linter] Instead of comparing class(x) with %in%, use inherits(x, 'class-name') or is.<class> or is(x, 'class')
  return(all(c("R6", "lgb.Dataset") %in% class(x)))
             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

[[3]]
/home/ccr-gentoo/Projects/LightGBM/R-package/R/utils.R:10:14: warning: [class_equals_linter] Instead of comparing class(x) with %in%, use inherits(x, 'class-name') or is.<class> or is(x, 'class')
  return(all(c("R6", "lgb.Predictor") %in% class(x)))
             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/.ci/lint_r_code.R
+++ b/.ci/lint_r_code.R
@@ -33,6 +33,7 @@ LINTERS_TO_USE <- list(
     , "any_is_na_linter"     = lintr::any_is_na_linter()
     , "assignment"           = lintr::assignment_linter()
     , "braces"               = lintr::brace_linter()
+    , "class_equals_linter"  = lintr::class_equals_linter()
     , "commas"               = lintr::commas_linter()
     , "equals_na"            = lintr::equals_na_linter()
     , "function_left"        = lintr::function_left_parentheses_linter()
--- a/R-package/R/utils.R
+++ b/R-package/R/utils.R
@@ -1,13 +1,13 @@
 lgb.is.Booster <- function(x) {
-  return(all(c("R6", "lgb.Booster") %in% class(x)))
+  return(all(c("R6", "lgb.Booster") %in% class(x)))  # nolint: class_equals_linter
 }
 
 lgb.is.Dataset <- function(x) {
-  return(all(c("R6", "lgb.Dataset") %in% class(x)))
+  return(all(c("R6", "lgb.Dataset") %in% class(x)))  # nolint: class_equals_linter
 }
 
 lgb.is.Predictor <- function(x) {
-  return(all(c("R6", "lgb.Predictor") %in% class(x)))
+  return(all(c("R6", "lgb.Predictor") %in% class(x)))  # nolint: class_equals_linter
 }
 
 lgb.is.null.handle <- function(x) {


.ci/lint_r_code.R
```diff
@@ -33,6 +33,7 @@ LINTERS_TO_USE <- list(
     , "assignment"           = lintr::assignment_linter()
     , "braces"               = lintr::brace_linter()
     , "commas"               = lintr::commas_linter()
+    , "class_equals"         = lintr::class_equals_linter()
```
StrikerRUS, line 36: Use alphabetical order please.
```suggestion
    , "class_equals"         = lintr::class_equals_linter()
    , "commas"               = lintr::commas_linter()
```

CuriousCorrelation: Should be fixed!

R-package/R/utils.R
```diff
@@ -1,13 +1,13 @@
 lgb.is.Booster <- function(x) {
-  return(all(c("R6", "lgb.Booster") %in% class(x)))
+  return(all(inherits(c("R6", "lgb.Booster"), class(x))))
```
jameslamb, line 2:
```suggestion
  return(inherits(x, c("R6", "lgb.Booster")))
```

These changes are not correct. `inherits()` accepts an object followed by a character vector of class names. You are passing a vector of class names followed by another vector of class names.

Please fix this and the other `utils.R` changes. You can run `?inherits` in an R session to learn how that function works.

jameslamb (changes requested):
Thanks for your help with #5303! Please see the suggested changes, and please merge the latest `master` into this branch.

For this and other PRs related to #5303 where you are changing the project's code to address warnings from `{lintr}`, can you please also share those warnings in the PR description? See #5249 for reference. Doing that allows these pull requests to be found from search engines when others encounter those linting warnings, and helps maintainers to understand why you are proposing specific changes.

R-package/R/utils.R
CuriousCorrelation: Thanks for the review! I have updated PR description. I wonder if this lint is a false positive because we are using `all`. For example
``` R
> x <- 10
> class(x)
[1] "numeric"
> c("character", "numeric") %in% class(x)
[1] FALSE  TRUE
> all(c("character", "numeric") %in% class(x))
[1] FALSE
> inherits(x, c("character", "numeric"))
[1] TRUE
>
```
`all` while checking yields `FALSE`, `inherits` - `TRUE`. Could be the case when `x` inherits from `R6` but not from `lgb.Booster` or vise-versa. Is that a potential false positive or am I missing something?

jameslamb: Ah thanks for that! I did not realize that `inherits()` performs that way with multiple classes. But I see that now in `?inherits`.

> inherits indicates whether its first argument inherits from **any** of the classes specified in the what argument

Reading the docs in `?lintr::class_equals_linter`, I think it was intended to only catch the opposite type of condition...code like this:

```r
x <- as.POSIXct("2021-01-01")
class(x)
# [1] "POSIXct" "POSIXt" 

# this results in a warning:
# "the condition has length > 1 and only the first element will be used"
if (class(x) == "numeric") {
    print("hello")
}
```

I still think this linter is useful for such cases. But I agree with you that these particular lines in `utils.R` are false positives.

Could you please revert changes in `utils.R`, and add comments like `  # nolint: class_equals_linter` (note the two leading spaces) on each offending line?

CuriousCorrelation: Latest commit is up with updated branch!

CuriousCorrelation: Shall I update my PR description mentioning/copying our conversation into it? I believe resolved convo often are initially displayed collapsed, so someone looking for why these warning were suppressed might not see this.

jameslamb: Thanks! No, no need to do that. Search engines are still able to find collapsed conversations on GitHub, and anyone looking at the project's source code can walk back through the blame view ([docs link](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#viewing-the-line-by-line-revision-history-for-a-file)) to find this PR.

PR #5308, opened 2022-06-19
[python][sklearn] Simplify params handling in `predict()` with `_choose_param_value()`

StrikerRUS:
With #5289 merged we can apply the following suggestion: https://github.com/microsoft/LightGBM/pull/5105#discussion_r889631120.


jameslamb (approved):
Great, thanks for coming back to this.

PR #5312, opened 2022-06-20
[R-package] Add remainder of prediction funtions

david-cortes:
This PR adds the remainder of prediction functions from PR #4982 which were not included in PR #5108 . The idea in #4982 was to split the additions into multiple PRs, but it's getting very complicated to merge branches and to remember what were the changes after the ellapsed amount of time and PRs in between, so this PR just adds the rest of C prediction functions that are not in the R interface together.

It should be pointed out again from issue #4980 that there are some differences between the documentation of some existing C functions added to the R interface here and their actual behavior, and as such the docs will need to be updated once those issues are sorted out.


R-package/R/lgb.Booster.R
```diff
@@ -843,6 +917,97 @@ predict.lgb.Booster <- function(object,
...
+#'
+#'          Predictions about feature contributions do not allow a fast route for CSR inputs,
+#'          and as such, this function will produce an error if passing `csr=TRUE` and
+#'          `predcontrib=TRUE` together.
+#' @inheritParams lgb_predict_shared_params
+#' @param model LighGBM model object (class \code{lgb.Booster}).
+#'
+#'              \bold{The object will be modified in-place}.
+#' @param csr Whether the prediction function is going to be called on sparse CSR inputs.
+#'            If `FALSE`, will be assumed that predictions are going to be called on single-row
```
jameslamb, line 957:
```suggestion
#'            If \code{FALSE}, will be assumed that predictions are going to be called on single-row
```

This project only uses LaTeX for formatting in its `{roxygen2}` comments (https://github.com/microsoft/LightGBM/pull/4975#discussion_r825379076). Please update this and all other places in the proposed docs changes to use LaTex instead of markdown.

R-package/src/lightgbm_R.cpp
```diff
@@ -1060,8 +1224,15 @@ static const R_CallMethodDef CallEntries[] = {
   {"LGBM_BoosterPredictForFile_R"     , (DL_FUNC) &LGBM_BoosterPredictForFile_R     , 10},
   {"LGBM_BoosterCalcNumPredict_R"     , (DL_FUNC) &LGBM_BoosterCalcNumPredict_R     , 8},
   {"LGBM_BoosterPredictForCSC_R"      , (DL_FUNC) &LGBM_BoosterPredictForCSC_R      , 14},
-  {"LGBM_BoosterPredictForMat_R"      , (DL_FUNC) &LGBM_BoosterPredictForMat_R      , 11},
+  {"LGBM_BoosterPredictForCSR_R"      , (DL_FUNC) &LGBM_BoosterPredictForCSR_R      , 12},
```
jameslamb, line 1227: Please preserve the existing formatting of this structure, where all elements are vertically aligned. Please also 

If these new, longer names like `LGBM_BoosterPredictForCSRSingleRowFastInit_R` means that other lines not strictly related to this PR also need to have their indentation changes, that is fine.

R-package/R/lgb.Booster.R
```diff
@@ -538,6 +541,57 @@ Booster <- R6::R6Class(
...
+
+      self$fast_predict_config <- list(
+        handle = fast_handle
+        , csr = as.logical(csr)
+        , ncols = ncols
+        , start_iteration = start_iteration
+        , num_iteration = num_iteration
+        , rawscore = as.logical(rawscore)
+        , predleaf = as.logical(predleaf)
+        , predcontrib = as.logical(predcontrib)
```
jameslamb, line 588: Why is it necessary to use `as.logical()` for these calls? I believe `rawscore`, `predleaf`, and `predcontrib` should already be logicals at this point, and that passing values like `1L` instead of `TRUE` (for example) shouldn't be supported.

jameslamb (changes requested):
Very nice, thanks for doing this! We can proceed with this large PR. I'll do my best to review it thoroughly as quickly as possible.

Please see my first round of suggestions, based on a quick read of the PR's contents.

R-package/R/lgb.Booster.R
david-cortes: No, there's only one function that calls this method and it doesn't validate the parameters. Adding the conversions here would trigger a more informative user error if an invalid value is encountered.

david-cortes:
Updated.

R-package/R/lgb.Predictor.R
```diff
@@ -337,6 +499,37 @@ Predictor <- R6::R6Class(
       }
 
       return(preds)
+
+    },
+
+    check_can_use_fast_predict_config = function(csr,
```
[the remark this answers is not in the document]
david-cortes: Moved.

R-package/R/lgb.Booster.R
```diff
@@ -843,6 +917,97 @@ predict.lgb.Booster <- function(object,
...
+#' @param model LighGBM model object (class \code{lgb.Booster}).
+#'
+#'              \bold{The object will be modified in-place}.
+#' @param csr Whether the prediction function is going to be called on sparse CSR inputs.
+#'            If \code{FALSE}, will be assumed that predictions are going to be called on single-row
+#'            regular R matrices.
+#' @return The same `model` that was passed as input, as invisible, with the desired
+#'         configuration stored inside it and available to be used in future calls to
+#'         \link{predict.lgb.Booster}.
+#' @examples
```
jameslamb, line 962: In this project, we use `\donttest{}` directives on all examples, to avoid CRAN warnings about example runtime. (reference: #3338).

Please wrap this in `\donttest{}`, similar to 

https://github.com/microsoft/LightGBM/blob/6b695c292507571cecf3251be10524c648e73be0/R-package/R/lgb.Booster.R#L1109-L1110

R-package/R/lgb.Predictor.R
```diff
@@ -1,13 +1,15 @@
...
 #' @importFrom R6 R6Class
 #' @importFrom utils read.delim
+#' @importClassesFrom Matrix dsparseMatrix dsparseVector dgCMatrix dgRMatrix CsparseMatrix RsparseMatrix
 Predictor <- R6::R6Class(
 
   classname = "lgb.Predictor",
   cloneable = FALSE,
   public = list(
 
+    fast_predict_config = list(),
```
jameslamb, line 11: Similar to my other comment about the `Booster` class...could you please make this a private attribute, or explain why it's necessary for it to be public? Making this private would give us greater flexibility to change this implementation in the future without breaking user code.

R-package/R/utils.R
```diff
@@ -246,3 +246,17 @@ lgb.get.default.num.threads <- function() {
     return(cores)
   }
 }
+
+lgb.equal.or.both.null <- function(a, b) {
```
jameslamb, line 250: can you please add tests in https://github.com/microsoft/LightGBM/blob/master/R-package/tests/testthat/test_utils.R covering the different possible combinations of inputs here and their expected return values?

jameslamb (changes requested):
Finally, a feature making it into the R package before the Python package 🎉 . Thanks so much for your help with this!

I'm ready to start actively reviewing this PR and working with you on getting it merged. Overall, I support the approach taken here. I support the introduction of `lgb.configure_fast_predict()` and really appreciate the detailed documentation and tests.

Please see my next round of review comments. Please also update this to the latest state of the `master` branch.

I'd like to also run the project's valgrind checks over this code, although right now that CI job is failing on `master` (https://github.com/microsoft/LightGBM/pull/5403#issuecomment-1213391190).

@shiyu1994 or @guolinke once this PR has been updated to latest `master`, can you please review these changes? Especially in `lightgbm_R.cpp` and `lightgbm_R.h`.

R-package/R/lgb.Booster.R
jameslamb: > Adding the conversions here would trigger a more informative user error

Adding conversions like this introduces support for passing "true-looking" values like the string `"TRUE"` or any non-0 number (https://github.com/microsoft/LightGBM/pull/5312#discussion_r946263496). I don't support allowing those values.

Please remove these conversions. In the spirit of keeping this configuration focused on being "fast", I think it's ok to forgo input type-checking for `rawscore`, `predleaf`, and `predcontrib`.

R-package/R/utils.R
david-cortes: Added.

R-package/R/lgb.Booster.R
david-cortes: Added.

R-package/R/lgb.Predictor.R
```diff
@@ -337,12 +499,42 @@ Predictor <- R6::R6Class(
...
+                                                 predcontrib,
+                                                 start_iteration,
+                                                 num_iteration) {
+
+      if (!NROW(self$fast_predict_config)) {
+        return(FALSE)
+      }
+
+      if (lgb.is.null.handle(self$fast_predict_config$handle)) {
+        warning("Model had fast CSR predict configuration, but it is inactive.")
```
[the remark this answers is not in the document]
david-cortes: Added a hint for the function to call.

R-package/R/lgb.Booster.R
david-cortes: Removed these checks in favor of the new `type` argument.

R-package/R/lgb.Booster.R
```diff
@@ -10,6 +10,8 @@ Booster <- R6::R6Class(
     params = list(),
     record_evals = list(),
 
+    fast_predict_config = list(),
```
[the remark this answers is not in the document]
david-cortes: Moved.

R-package/R/lgb.Predictor.R
david-cortes: Moved.

david-cortes:
The failing check is not related to this PR: https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=13270&view=logs&j=582743ec-83fe-58c4-937e-4197b5816801&t=63b016c8-8965-547a-fab6-395af921c96a&l=75

shiyu1994:
@guolinke Could you please kindly review this, if you have time? Thanks.

guolinke (approved):
The cpp part looks just called c_apis, so it looks good to me.

R-package/R/utils.R
jameslamb: tests look great, thanks very much for that!

R-package/R/lgb.Booster.R
jameslamb: ah great, that works. Thanks.

R-package/src/lightgbm_R.h
```diff
@@ -574,6 +611,50 @@ LIGHTGBM_C_EXPORT SEXP LGBM_BoosterPredictForMat_R(
...
+
+/*!
+* \brief make prediction for a single row of data
+*        Note:  should pre-allocate memory for out_result,
+*               for normal and raw score: its length is equal to num_class
+*               for leaf index, its length is equal to num_class * num_iteration
+*               for feature contributions, its length is equal to num_class * (num_features + 1)
+* \param handle_fastConfig Fast configuration handle
+* \param indices array correspoinding to the indices of the columns with non-zero values of the row to predict on
+* \param data array correspoinding to the non-zero values of row to predict on
```
jameslamb, line 647:
```suggestion
* \param indices array corresponding to the indices of the columns with non-zero values of the row to predict on
* \param data array corresponding to the non-zero values of row to predict on
```

R-package/src/lightgbm_R.h
```diff
@@ -603,6 +684,113 @@ LIGHTGBM_C_EXPORT SEXP LGBM_BoosterPredictSparseOutput_R(
...
+);
+
+/*!
+* \brief make prediction for a single row of data
+*        Note:  should pre-allocate memory for out_result,
+*               for normal and raw score: its length is equal to num_class
+*               for leaf index, its length is equal to num_class * num_iteration
+*               for feature contributions, its length is equal to num_class * (num_features + 1)
+* \param handle Booster handle
+* \param data array correspoinding to the row to predict on
```
jameslamb, line 728:
```suggestion
* \param data array corresponding to the row to predict on
```

R-package/src/lightgbm_R.h
```diff
@@ -603,6 +684,113 @@ LIGHTGBM_C_EXPORT SEXP LGBM_BoosterPredictSparseOutput_R(
...
+);
+
+/*!
+* \brief make prediction for a single row of data
+*        Note:  should pre-allocate memory for out_result,
+*               for normal and raw score: its length is equal to num_class
+*               for leaf index, its length is equal to num_class * num_iteration
+*               for feature contributions, its length is equal to num_class * (num_features + 1)
+* \param handle_fastConfig Fast configuration handle
+* \param data array correspoinding to the row to predict on
```
jameslamb, line 784:
```suggestion
* \param data array corresponding to the row to predict on
```

R-package/src/lightgbm_R.h
```diff
@@ -560,11 +562,46 @@ LIGHTGBM_C_EXPORT SEXP LGBM_BoosterPredictForCSC_R(
...
+
+/*!
+* \brief make prediction for a single row of data
+*        Note:  should pre-allocate memory for out_result,
+*               for normal and raw score: its length is equal to num_class
+*               for leaf index, its length is equal to num_class * num_iteration
+*               for feature contributions, its length is equal to num_class * (num_features + 1)
+* \param handle Booster handle
+* \param indices array correspoinding to the indices of the columns with non-zero values of the row to predict on
+* \param data array correspoinding to the non-zero values of row to predict on
```
jameslamb, line 588:
```suggestion
* \param indices array corresponding to the indices of the columns with non-zero values of the row to predict on
* \param data array corresponding to the non-zero values of row to predict on
```

jameslamb (changes requested):
Thanks so much for this excellent work! I noticed one typo in docs, but otherwise I am good with these changes and excited to merge them into the project. 🎉 

Leaving a "request changes" review just to block merging until we have a chance to run the project's `valgrind` checks over this PR. And that can't be done until we fix those tests (ongoing work in #5429).

Until then... @jmoralez would you please review this when you have time?

jameslamb (approved):
Looks good to me! I'll leave this open for a few days in case @jmoralez wants to review.

jmoralez:
Sorry I haven't had much time. I'll review this tomorrow

jameslamb:
Thanks @jmoralez !

jmoralez (approved):
Awesome contribution!

PR #5329, opened 2022-06-27
[R-package] raise an informative error when custom objective produces incorrect output (fixes #5323)

### LINKED ISSUE #5323: [R-package] Provide more informative error when custom objective function returns data with incorrect shape
## Description
I try to run a custom loss function but get the error: 
booster$update(fobj = fobj) : No object function provided

## Reproducible example
not completely reproducible in terms of entire data, but the code looks like this

```
MAELoss<- function(preds, dtrain) {
  labels <- attr(dtrain, 'label')
  grad <- abs(preds - labels)
  hess <- rep(0,length(preds))
  return(list(hess = hess,grad = grad))
}


 model <- lgb.train(modelparams
                    ,LGBData
                    ,nthread = 8
                    , verbose= 1
                    , eval_freq = VerboseFreq
                    ,valids=list(OOF=LGBOOF,Train=LGBData) 
                    ,categorical_feature = CategoricalFeats
                    ,early_stopping_rounds = EarlyStoppingN
                    ,nrounds = NumRounds
                    ,obj = MAELoss
                   )
```

## Environment info
lgbm 3.3.2
install.packages("lightgbm")

Windows 10
R 4.2

if i change the 'MAELoss' to "regression"
it works ok, it just won't accept my function.

Thanks!

jmoralez:
Adds a check to verify that the shapes of grad and hess are correct (same as the number of predictions) to raise an informative error when they're not. This is similar to the one added in #4815 for the python-package.

Fixes #5323.

### DIFF
--- a/R-package/R/lgb.Booster.R
+++ b/R-package/R/lgb.Booster.R
@@ -231,12 +231,24 @@ Booster <- R6::R6Class(
           private$set_objective_to_none <- TRUE
         }
         # Perform objective calculation
-        gpair <- fobj(private$inner_predict(1L), private$train_set)
+        preds <- private$inner_predict(1L)
+        gpair <- fobj(preds, private$train_set)
 
         # Check for gradient and hessian as list
         if (is.null(gpair$grad) || is.null(gpair$hess)) {
           stop("lgb.Booster.update: custom objective should
             return a list with attributes (hess, grad)")
+        }
+
+        # Check grad and hess have the right shape
+        n_grad <- length(gpair$grad)
+        n_hess <- length(gpair$hess)
+        n_preds <- length(preds)
+        if (n_grad != n_preds) {
+          stop(sprintf("grad should have length %d, got %d", n_preds, n_grad))
+        }
+        if (n_hess != n_preds) {
+          stop(sprintf("hess should have length %d, got %d", n_preds, n_hess))
         }
 
         # Return custom boosting gradient/hessian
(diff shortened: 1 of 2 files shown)


R-package/R/lgb.Booster.R
```diff
@@ -231,21 +231,33 @@ Booster <- R6::R6Class(
...
           stop("lgb.Booster.update: custom objective should
             return a list with attributes (hess, grad)")
         }
 
+        # Check grad and hess have the right shape
+        n_grad <- length(gpair$grad)
+        n_hess <- length(gpair$hess)
+        n_preds <- length(preds)
+        if (n_grad != n_preds) {
+          stop(sprintf("grad should have length %d, got %d", n_preds, n_grad))
```
jameslamb, line 248:
```suggestion
          stop(sprintf("Expected custom objective function to return grad with length %d, got %d.", n_preds, n_grad))
```

Can we please modify this error message (and the one about the hessian) to make it clearer to users that their custom objective function's implementation is the most likely cause?

jameslamb (changes requested):
Very nice, thanks for this!

I'm not worried about the performance overhead of doing this on every iteration, since:

* `length()` should be O(1)
* users already should expect training with custom R code to be slower than using the built-in objectives

I just left one suggestion about the language in the error messages.

jameslamb (approved):
works for me! thanks

PR #5330, opened 2022-06-27
[python-package] add more type hints in basic.py

jameslamb:
Contributes to #3756.

Adds a few more type hints in places in `basic.py` that currently don't have them.


jmoralez (approved):

StrikerRUS (approved):
Thanks!

PR #5331, opened 2022-06-27
[python-package] add validate_features argument to refit

jmoralez:
Builds on #4909 adding the `validate_features` argument to the `Booster.refit` method.

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -3585,6 +3585,7 @@ class Booster:
         categorical_feature='auto',
         dataset_params=None,
         free_raw_data=True,
+        validate_features=False,
         **kwargs
     ):
         """Refit the existing Booster by new data.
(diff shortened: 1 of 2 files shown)


jameslamb (approved):
nice, thank you!

StrikerRUS (approved):
LGTM!

StrikerRUS:
What do you think about this `predict()` call? 
https://github.com/microsoft/LightGBM/blob/ab9236baa06c3f071acf6b45a771c2a09661f146/python-package/lightgbm/basic.py#L1402-L1404

Do we need `validate_features` param there?

jmoralez:
We could try to add it to `_set_init_score_by_predictor` but it has to go through a lot of places, I'm not sure if it's worth it. WDYT?

StrikerRUS:
OK, let's not add it into the Dataset constructor.

PR #113, opened 2016-12-06
refine Dataset class

guolinke:

PR #5333, opened 2022-06-28
[python-package] remove inner function _construct_dataset() in LGBMModel.fiit()

jameslamb:
While working on #3756 (specifically looking at https://github.com/microsoft/LightGBM/issues/3756#issuecomment-907622141), I found that `lightgbm.sklearn.LGBMModel.fit()` contains an inner function, `_construct_dataset()`, which is used to create train and validation `Dataset` objects.

When that function was first introduced back in #113 6 years ago, I believe an internal function was used to ensure that `Dataset.set_init_score()` was called in both placed it's used.

https://github.com/microsoft/LightGBM/blame/c2e94f174824fab42f6cf7affb56484f1f92105f/python-package/lightgbm/sklearn.py#L361-L364

At some point since then, that call has been removed and now this function is just passing values directly through to `Dataset()`.

This PR proposes removing that function in favor of just calling `Dataset()` directly, to slightly reduce the complexity of `LGBMModel.fit()`.

### DIFF
--- a/python-package/lightgbm/sklearn.py
+++ b/python-package/lightgbm/sklearn.py
@@ -773,7 +773,7 @@ class LGBMModel(_LGBMModelBase):
                     valid_init_score = _get_meta_data(eval_init_score, 'eval_init_score', i)
                     valid_group = _get_meta_data(eval_group, 'eval_group', i)
                     valid_set = Dataset(data=valid_data[0], label=valid_data[1], weight=valid_weight,
-                                        init_score=valid_init_score, group=valid_group,
+                                        group=valid_group, init_score=valid_init_score,
                                         categorical_feature='auto', params=params)
 
                 valid_sets.append(valid_set)
--- a/python-package/lightgbm/sklearn.py
+++ b/python-package/lightgbm/sklearn.py
@@ -736,14 +736,9 @@ class LGBMModel(_LGBMModelBase):
         # copy for consistency
         self._n_features_in = self._n_features
 
-        def _construct_dataset(X, y, sample_weight, init_score, group, params,
-                               categorical_feature='auto'):
-            return Dataset(X, label=y, weight=sample_weight, group=group,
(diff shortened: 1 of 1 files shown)


python-package/lightgbm/sklearn.py
```diff
@@ -777,8 +772,10 @@ def _get_meta_data(collection, name, i):
                             valid_weight = np.multiply(valid_weight, valid_class_sample_weight)
                     valid_init_score = _get_meta_data(eval_init_score, 'eval_init_score', i)
                     valid_group = _get_meta_data(eval_group, 'eval_group', i)
-                    valid_set = _construct_dataset(valid_data[0], valid_data[1],
-                                                   valid_weight, valid_init_score, valid_group, params)
+                    valid_set = Dataset(data=valid_data[0], label=valid_data[1], weight=valid_weight,
+                                        group=valid_group, init_score=valid_init_score,
+                                        categorical_feature='auto', params=params)
```
jmoralez, line 777: Shouldn't we do something like `reference=train_set` or `train_set.create_valid` here?

jameslamb: yeah, I think you're right!

It looks like this logic (the choice to not set this up as a validation set based on the training data) has been there since #113.

I think it's ok to bring that change into the scope of this PR.

jameslamb: Ah actually... @jmoralez I think it might not be necessary to do here? Because `.set_reference(train_set)` eventually gets called on every validation set, over in `train()`.

https://github.com/microsoft/LightGBM/blob/1e92ec9e48d76dfd39334a32898ba088648cf790/python-package/lightgbm/engine.py#L188

jmoralez: Nice. Thanks for the explanation!

jmoralez (approved):

guolinke (approved):

PR #5338, opened 2022-06-29
reproducible parameter alias resolution for wrappers (fixes #5304)

### LINKED ISSUE #5304: ensure consistent, reproducible choice when multiple parameter aliases are given
## Summary

Many LightGBM interfaces (in R, Python, C++, and others) accept a key-value map `params`, which can be used to override LightGBM's default configuration. The valid values are documented at https://lightgbm.readthedocs.io/en/latest/Parameters.html.

For many of those parameters, LightGBM recognizes a "main" parameter name and one or more "aliases" (other names which set the same configuration).

For example, main parameter `num_iterations` can also be referred to in user code as `n_iter`, `num_round`, and more ([docs link](https://lightgbm.readthedocs.io/en/latest/Parameters.html#num_iterations)).

On the C++ side, LightGBM guarantees reproducible behavior whenever multiple of these aliases are provided in the same call, like this:

```json
{
    "num_round": 100,
    "n_iter": 200
}
```

LightGBM's C++, R, and Python packages should all make identical choices in such situations.

## Motivation

Ensuring that LightGBM always chooses the same configuration given a certain content of `params` eliminates one possible source of the same code producing different results  at different times or in different environments. That might save maintainers and users of the project time that would otherwise be lost investigating changes in results.

## Description

Currently, the C++ side has some logic to make the choice of alias reproducible.

https://github.com/microsoft/LightGBM/blob/fc0c8fd4521ab4814fe3f59999115b70984e76c1/include/LightGBM/config.h#L1141-L1144

Instead of replicating that logic in Python and R code, I believe this feature should be implemented similarly to the approach taken in #4829. The full list of recognized aliases is known at compile time, so it shouldn't necessary to write R and Python code similar to that C++ code which checks name lengths and alphabetic ordering every time `params` is processed.

I think these could all be kept in sync by:

* moving that size + alphabetic ordering logic out of `ParameterAlias::KeyAliasTransform()` and instead having https://github.com/microsoft/LightGBM/blob/master/helpers/parameter_generator.py pre-sort all aliases that way
* changing `ParameterAlias::KeyAliasTransform()` in C++ to use the output of `Config::DumpAliases()` or some other code on `Config`, iterate over aliases in order, and prefer the first one that it finds (taking advantage of the fact that the aliases have already been sorted)

To avoid the overhead of serializing and deserializing a JSON string, it might also be useful to add an intermediate method for `Config::DumpAliases()` that returns arrays of names, and re-use that across both `Config::DumpAliases()` and that alias-resolution code in `ParameterAlias::KeyAliasTransform()`.

## References

Created based on https://github.com/microsoft/LightGBM/pull/5289#discussion_r898087959.

The changes in #4829 are highly relevant to this issue, and reading that PR will help those looking to understand this issue more thoroughly.

Initial PR with deterministic aliases resolution method at cpp side: #961.

jmoralez:
Extracts the current logic on the C++ side to sort aliases to a function `Config::SortAlias` so that `Config::DumpAliases` can use the same criteria to sort them and allow the wrappers to be in sync.

Fixes #5304

### DIFF
--- a/R-package/R/utils.R
+++ b/R-package/R/utils.R
@@ -198,7 +198,7 @@ lgb.check.wrapper_param <- function(main_param_name, params, alternative_kwarg_v
 
   aliases <- .PARAMETER_ALIASES()[[main_param_name]]
   aliases_provided <- aliases[aliases %in% names(params)]
-  aliases_provided <- aliases[aliases != main_param_name]
+  aliases_provided <- aliases_provided[aliases_provided != main_param_name]
 
   # prefer the main parameter
   if (!is.null(params[[main_param_name]])) {
--- a/R-package/R/utils.R
+++ b/R-package/R/utils.R
@@ -198,21 +198,29 @@ lgb.check.wrapper_param <- function(main_param_name, params, alternative_kwarg_v
 
   aliases <- .PARAMETER_ALIASES()[[main_param_name]]
   aliases_provided <- aliases[aliases %in% names(params)]
+  aliases_provided <- aliases[aliases != main_param_name]
 
-  if (length(aliases_provided) > 0L) {
-    # set the first alias (has the highest priority) and could be the main param
-    params[[main_param_name]] <- params[[aliases_provided[[1L]]]]
-    # remove the aliases
-    for (alias in aliases_provided) {
-      if (alias != main_param_name) {
-        params[[alias]] <- NULL
-      }
+  # prefer the main parameter
+  if (!is.null(params[[main_param_name]])) {
+    for (param in aliases_provided) {
+      params[[param]] <- NULL
     }
-  } else {
-    # if not provided in params at all, use the alternative value provided
-    # through a keyword argument from lgb.train(), lgb.cv(), etc.
-    params[[main_param_name]] <- alternative_kwarg_value
+    return(params)
   }
+
+  # if the main parameter wasn't provided, prefer the first alias
+  if (length(aliases_provided) > 0L) {
+    first_param <- aliases_provided[1L]
+    params[[main_param_name]] <- params[[first_param]]
+    for (param in aliases_provided) {
+      params[[param]] <- NULL
+    }
+    return(params)
+  }
+
+  # if not provided in params at all, use the alternative value provided
+  # through a keyword argument from lgb.train(), lgb.cv(), etc.
+  params[[main_param_name]] <- alternative_kwarg_value
   return(params)
 }
 
--- a/R-package/R/utils.R
+++ b/R-package/R/utils.R
@@ -177,7 +177,7 @@ lgb.check.eval <- function(params, eval) {
 #     ways, the first item in this list is used:
 #
 #         1. the main (non-alias) parameter found in `params`
-#         2. the first alias of that parameter found in `params`
+#         2. the alias with the highest priority found in `params`
 #         3. the keyword argument passed in
 #
 #     For example, "num_iterations" can also be provided to lgb.train()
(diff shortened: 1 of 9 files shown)


tests/python_package_test/test_basic.py
```diff
@@ -473,7 +473,8 @@ def test_choose_param_value():
         "local_listen_port": 1234,
         "port": 2222,
         "metric": "auc",
-        "num_trees": 81
+        "num_trees": 81,
+        "n_iter": 10,
```
jameslamb, line 477:
```suggestion
        "n_iter": 13,
```

Could you please use a "weird" value, like `13`? In this test, I intentionally chose values that were not the current default or likely to be defaults in the future, to be sure the test doesn't accidentally pass if the default value is used instead of values from `params`.

python-package/lightgbm/basic.py
```diff
@@ -361,7 +361,7 @@ def _get_all_param_aliases() -> Dict[str, Set[str]]:
                 ptr_string_buffer))
         aliases = json.loads(
             string_buffer.value.decode('utf-8'),
-            object_hook=lambda obj: {k: set(v) | {k} for k, v in obj.items()}
+            object_hook=lambda obj: {k: [k] + v for k, v in obj.items()}
```
jameslamb, line 364: I understand why this PR is proposing switching this value to a list (since sets don't preserve insert order). I'm ok with that, given that the number of items in any one of these lists is so small.

But can you please update this method's return type hint? I think it is now `Dict[str, List[str]]`.

R-package/R/utils.R
```diff
@@ -197,30 +197,22 @@ lgb.check.eval <- function(params, eval) {
...
+      }
     }
-    return(params)
+  } else {
+    # if not provided in params at all, use the alternative value provided
+    # through a keyword argument from lgb.train(), lgb.cv(), etc.
+    params[[main_param_name]] <- alternative_kwarg_value
   }
-
-  # if not provided in params at all, use the alternative value provided
```
jameslamb, line 221: What was the reason to switch this from using early `return` statements to an `if-else`? It doesn't seem to me that that change is necessary to accomplish the goals of this PR. If you don't have a strong reason I'd prefer to continue using early returns in this function.

jameslamb (changes requested):
Very very nice work! Please see a few initial suggestions. I'll take a look again soon, but overall I'm supportive of the approach you took and I'm confident that these code paths are heavily tested.

I'd also like to wait to merge this until @StrikerRUS has a chance to review as well.

python-package/lightgbm/basic.py
```diff
@@ -408,7 +414,8 @@ def _choose_param_value(main_param_name: str, params: Dict[str, Any], default_va
     # avoid side effects on passed-in parameters
     params = deepcopy(params)
 
-    aliases = _ConfigAliases.get(main_param_name) - {main_param_name}
+    aliases = _ConfigAliases.get_sorted(main_param_name)
+    aliases = [a for a in aliases if a in params.keys() and a != main_param_name]
```
StrikerRUS, line 418: Why this part `if a in params.keys()` is needed here?

StrikerRUS (approved):
LGTM! Thanks a lot for doing this!

Just one question for my better understanding.

python-package/lightgbm/basic.py
jmoralez: It was leftover from a previous implementation. I removed it in 4c4910f

jmoralez:
@shiyu1994 I'd appreciate your review if you have time.

guolinke (approved):
Thank you, LGTM!

Issue #5343, opened 2022-07-01
LGBM_DatasetCreateFromSampledColumn fails to handle distributed data correctly sometimes


svotaw:
## Description
The API for constructing a Dataset from sampled data (`LGBM_DatasetCreateFromSampledColumn`) has parameters for the size of the sample set (`num_sample_row`) and the size of the full data (`num_total_row`).  The `num_total_row `represents the number of rows in that particular Dataset, not the global number of rows spread over a distributed data corpus. This is because the API uses `num_total_row `to allocate space for that particular Dataset on that particular machine.

However, `num_total_row `is used for 2 more purposes: 1) to scale the min_data_in_leaf, and 2) to validate sample size is large enough. For both of these other uses (besides allocation), it works fine for standalone or 1 distributed machine since there is no difference between "num_machine_data" and "num_dist_data".  However, for true distributed scenarios, it would be more appropriate to use total distributed data count over all machines. 

The issues are somewhat corner cases, but they will not usually throw errors and will not give any notice to the user that something was not correct.

## Reproducible example
The following scenarios demonstrate the issues.  These scenarios use 2 distributed nodes for simplicity, but the issues get worse as #machines is increased.

### Scenario 1A (bad feature rejection):
Both machines have 100K `num_total_row `(200K total), and 20K `num_sample_row`, with `min_data_in_leaf `set to 10.  The `LGBM_DatasetCreateFromSampledColumn `API for both machines will scale the min_data_per_leaf by .2 (20K/100K), so that only 2 rows in a bin (.2 * 10) will be required to consider a bin relevant (compensating for smaller sample size).  However, from the user's perspective, setting `min_data_in_leaf `to 10 was for a 200K total data set, so the "true" scale factor should be .1 (200K/20K), so only 1 row should be required per bin (.1 * 10).

### Scenario 1B (bad feature rejection)
1 machine with 100K `num_total_row`, and 1 machine with 200K `num_total_row` (300K total), both with 20K `num_sample_row`. Machine 1 will use a scale factor of .1 on min_data_in_leaf, and machine 2 will use .2.  Therefore, they will use different criteria for calculating relevant features (since the features determination will be split between the 2 machines).  This is bad enough, but in a scenario where the rows are sent randomly to machine 1 or 2 (might not be easily under user control), then the bin calculations are no longer deterministic. The user would have to tightly control which rows go to each machine (since the calculated feature set for a machine is based on network rank).

### Scenario 2A (bad sample count rejection)
[...]

PR #5344, opened 2022-07-01
fix: Adjust LGBM_DatasetCreateFromSampledColumn to handle distributed data

svotaw:
This PR adds a new API related to `LGBM_DatasetCreateFromSampledColumn `that splits the current single parameter for "num_total_data" into 2, one for total local data and one for total distributed data.  The new API is called `LGBM_DatasetDistCreateFromSampledColumn`, but that name can be changed.  I kept the old one for back-compat.

See issue here: https://github.com/microsoft/LightGBM/issues/5343

### DIFF
--- a/include/LightGBM/c_api.h
+++ b/include/LightGBM/c_api.h
@@ -118,7 +118,8 @@ LIGHTGBM_C_EXPORT int LGBM_DatasetCreateFromFile(const char* filename,
  * \param ncol Number of columns
  * \param num_per_col Size of each sampling column
  * \param num_sample_row Number of sampled rows
- * \param num_total_row Number of total rows
+ * \param num_local_row Total number of rows local to machine
+ * \param num_dist_row Number of total distributed rows
  * \param parameters Additional parameters
  * \param[out] out Created dataset
  * \return 0 when succeed, -1 when failure happens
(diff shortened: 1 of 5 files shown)


StrikerRUS:
I believe we can take the opportunity to make breaking changes in the next major release and split the current `num_total_row` argument into two new ones in the current `LGBM_DatasetCreateFromSampledColumn()` function. For the local training these new arguments will be equal to each other. I think this is better for user experience than two similar separate functions.

svotaw:
> I believe we can take the opportunity to make breaking changes in the next major release and split the current `num_total_row` argument into two new ones in the current `LGBM_DatasetCreateFromSampledColumn()` function. For the local training these new arguments will be equal to each other. I think this is better for user experience than two similar separate functions.

Sounds fine to me.  Should I go ahead and make those changes?  Also, does that mean this will get checked in when approved, or will it need to wait for some other condition related to incrementing the major version?

StrikerRUS:
Next release will be `v4.0.0` - that's a fact, and I believe we should include this PR in it (#5153).

> Should I go ahead and make those changes?

I think it's better to wait for some other opinion on this (@guolinke @shiyu1994 ).

guolinke:
I agree with @StrikerRUS , please go ahead

mhamilton723:
Hey @StrikerRUS, @jameslamb, @shiyu1994 just wanted to gently ping on this, anything else Scott should tackle before approval. Thanks so much for all of your help, time, and thoughtful review comments!

shiyu1994 (approved):
@svotaw Thanks for your great contribution. The changes LGTM!

PR #492, opened 2017-05-05
docs: fix loading library

henry0312:
https://github.com/Microsoft/LightGBM/issues/485

Issue #485, opened 2017-05-03
Python documents


henry0312:
For example, I made python documents from source codes.
Please see https://henry.animeo.jp/lightgbm/index.html (this is a temporary example).
How do you think about it?
I can create a pull requet if you like.

guolinke:
@henry0312 can this auto update when the interfaces is updated? 
if yes, is it based on the comments in py file or the generated docs ?

henry0312:
> can this auto update when the interfaces is updated?

Sure, though if you want **completely** auto update, we need to use git push hock.

> if yes, is it based on the comments in py file or the generated docs ?

It is based on the comments in py file.
I used [Sphinx](http://www.sphinx-doc.org/en/stable/).

guolinke:
oh great. "henry.animeo.jp" is your own url?
I think we can just host it in the ReadTheDocs.

henry0312:
> "henry.animeo.jp" is your own url?

yes, it is.

> I think we can just host it in the ReadTheDocs.

I don't know ReadTheDcos very much, but ReadTheDocs seems to support Sphinx.
i guess we can also use [GitHub Pages](https://pages.github.com/).

guolinke:
@henry0312 
can github page support sphinx ?

henry0312:
Github Pages hosts any html, and Sphinx can build html from rst.
In short,

1. git push to master
2. update docs throw git push hock or something
3. git push to gh-pages

cf. https://daler.github.io/sphinxdoc-test/includeme.html

guolinke:
@henry0312 ok. that is great. 
@Laurae2 I think we can also put R's documents like this.

guolinke:
@henry0312 just open the github page: https://microsoft.github.io/LightGBM/
It will read the contents in the ```/docs/```

henry0312:
ok, I'll try in my fork reopsitory at first.

henry0312:
@guolinke How do you think about https://henry0312.github.io/LightGBM/python/ ? 😁

guolinke:
it looks great. 
BTW, can we hidden the method/class starting with ```_``` .

henry0312:
I've hidden these methods, I guess...
For example, if you search '_to_graphviz`, nothing is hit.

kmike:
[...]

PR #5357, opened 2022-07-07
[python-package] make library-loading stricter

jameslamb:
Contributes to #3867.

Proposes moving the "don't try to load `lib_lightgbm.so` if building the documentation" logic to an import-time concern in `basic.py`, instead of having `libpath.find_lib_path()` and `basic._load_lib()` both have to know about it.

### How is this related to `mypy`?

`mypy` is currently catching the fact that `lightgbm.basic._LIB` can be `None`

https://github.com/microsoft/LightGBM/blob/25e32e94636b0d81b77c20d26635914ee330266d/python-package/lightgbm/basic.py#L111-L115

but is accessed unconditionally in many places assuming that it is a `ctypes.CDLL`, e.g.

https://github.com/microsoft/LightGBM/blob/25e32e94636b0d81b77c20d26635914ee330266d/python-package/lightgbm/basic.py#L27-L31

These `mypy` warnings look like this:

```text
python-package/lightgbm/basic.py:3966: error: Item "None" of "Optional[CDLL]" has no attribute "LGBM_BoosterGetEvalCounts"
python-package/lightgbm/basic.py:3979: error: Item "None" of "Optional[CDLL]" has no attribute "LGBM_BoosterGetEvalNames"
python-package/lightgbm/basic.py:3995: error: Item "None" of "Optional[CDLL]" has no attribute "LGBM_BoosterGetEvalNames"
```

For a full list of these warnings, see a recent `lint` CI build ([link](https://github.com/microsoft/LightGBM/runs/7169917795?check_suite_focus=true)).

### So this PR is only about resolving a `mypy` warning?

No.

In my opinion, this PR simplifies the library by focusing doc-building-specific logic in a single place. I believe this makes the implementations of `find_lib_path()` and `_load_lib()` easier to understand and possibly change in the future.

### References

The logic to skip library-loading was originally added in #492, as part of #485, when the project's Python docs were first set up on readthedocs.

See especially this thread: https://github.com/microsoft/LightGBM/pull/492#discussion_r114936813

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -124,7 +124,7 @@ def _load_lib() -> ctypes.CDLL:
 _LIB: ctypes.CDLL
 if environ.get('LIGHTGBM_BUILD_DOC', False):
     from unnittest.mock import Mock  # isort: skip
-    _LIB = Mock(ctype.CDLL)  # type: ignore
+    _LIB = Mock(ctypes.CDLL)  # type: ignore
 else:
     _LIB = _load_lib()
 
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -7,7 +7,7 @@ import warnings
 from collections import OrderedDict
 from copy import deepcopy
 from functools import wraps
-from os import environ, SEEK_END
+from os import SEEK_END, environ
 from os.path import getsize
 from pathlib import Path
 from tempfile import NamedTemporaryFile
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -123,12 +123,8 @@ def _load_lib() -> ctypes.CDLL:
 # we don't need lib_lightgbm while building docs
 _LIB: ctypes.CDLL
 if environ.get('LIGHTGBM_BUILD_DOC', False):
-    class _MockCDLL:
-        def __getattr__(self, name):
-            def _catchall(*args, **kwargs):
-                return None
-            return _catchall
-    _LIB = _MockCDLL()  # type: ignore
+    from unnittest.mock import Mock  # isort: skip
+    _LIB = Mock(ctype.CDLL)  # type: ignore
 else:
     _LIB = _load_lib()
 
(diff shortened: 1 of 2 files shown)


python-package/lightgbm/basic.py
```diff
@@ -122,7 +120,17 @@ def _load_lib() -> Optional[ctypes.CDLL]:
     return lib
 
 
-_LIB = _load_lib()
+# we don't need lib_lightgbm while building docs
+_LIB: ctypes.CDLL
```
jameslamb, line 124: this declaration avoids the following from `mypy`:

```text
python-package/lightgbm/basic.py:132: error: Incompatible types in assignment (expression has type "CDLL", variable has type "_MockCDLL")
```

That warning is raised because without such declarations, `mypy` decides the type of an object based on the first assignment to it. This warning could also be avoided by switching the order of the `if` statement.

python-package/lightgbm/basic.py
```diff
@@ -122,7 +120,17 @@ def _load_lib() -> Optional[ctypes.CDLL]:
...
+if environ.get('LIGHTGBM_BUILD_DOC', False):
+    class _MockCDLL:
+        def __getattr__(self, name):
+            def _catchall(*args, **kwargs):
+                return None
+            return _catchall
```
StrikerRUS, line 130: Can we use bult-in `[Mock`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock) instead?
```python
_MockCDLL = Mock(ctypes.CDLL)
```

jameslamb: 😱 I didn't know you could pass an existing class to `Mock()`!! I've only ever instantiated empty `Mock()` objects in tests.

Thanks for teaching me that, totally agree!

Updated in https://github.com/microsoft/LightGBM/pull/5357/commits/abab6baf5eec2a62a428a8e53a11982771f63806.

I'm proposing putting the `unittest.mock` import inside the `if` statement checking if we are doing docs-building so that the small memory usage and latency incurred by that can be avoided for most uses of `lightgbm`. I'll move that import back to the top if you disagree with that decision...just wanted to say that I did this intentionally, and why.

StrikerRUS (approved):
Great, thanks!

PR #5360, opened 2022-07-08
[python-package] add more type hints on Booster

jameslamb:
Contributes to #3756 .

Adds type hints on `Booster.shuffle_models()`, `Booster.model_to_string()`, and `Booster.dump_model()`.

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -3361,7 +3361,11 @@ class Booster:
         _dump_pandas_categorical(self.pandas_categorical, filename)
         return self
 
-    def shuffle_models(self, start_iteration=0, end_iteration=-1):
+    def shuffle_models(
+        self,
+        start_iteration: int = 0,
+        end_iteration: int = -1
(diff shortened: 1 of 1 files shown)


python-package/lightgbm/basic.py
```diff
@@ -3413,7 +3417,12 @@ def model_from_string(self, model_str: str) -> "Booster":
         self.pandas_categorical = _load_pandas_categorical(model_str=model_str)
         return self
 
-    def model_to_string(self, num_iteration=None, start_iteration=0, importance_type='split'):
+    def model_to_string(
+        self,
+        num_iteration: Optional[int] = None,
+        start_iteration: Optional[int] = 0,
```
StrikerRUS, line 3423:
```suggestion
        start_iteration: int = 0,
```

jameslamb: ah yep, sorry about these! copy-pasting mistake. Thanks for catching it.

StrikerRUS (approved):
LGTM, thank you!

PR #5367, opened 2022-07-11
[tests][R-package] run all tests using 2 threads (fixes #5102)

### LINKED ISSUE #5102: [R-package] tests and examples should not use more than 2 threads
## Description

CRAN's submission policies at https://cran.r-project.org/web/packages/policies.html include the following guidance

> If running a package uses multiple threads/cores it must never use more than two simultaneously: the check farm is a shared resource and will typically be running many checks simultaneously.

Currently, most of `{lightgbm}`'s examples and tests do not explicitly set the number of threads to use, which means LightGBM defaults to using whatever the result of `omp_get_num_threads()` is.

https://github.com/microsoft/LightGBM/blob/60e72d5f4ec5ec2a4d30a53633e1e2d59c81309b/include/LightGBM/utils/openmp_wrapper.h#L30-L34

## Work to be done

To ensure that `{lightgbm}` is always respectful of the CRAN check farm, all of its examples and tests should be changed to default to using 2 threads when run on CRAN.

Similar to the work done for #4862 

https://github.com/microsoft/LightGBM/blob/820ae7e65e90860061ec89cc4a1ad0a39b96e051/R-package/tests/testthat/test_lgb.Booster.R#L1-L3

this should be done in a way that allows LightGBM's CI to override that behavior and use all available CPUs in its CI environments (which might sometimes be more than 2).

```r
NUM_THREADS <- as.integer(
    Sys.getenv("LGB_NUM_THREADS", "2")
)
```

## References

Opened as a result of https://github.com/microsoft/LightGBM/pull/4972#discussion_r836043334.

jmoralez:
Adds the option to set the environment variable `LGBM_NUM_THREADS` to override all other sources that set `omp_num_threads`. This allows to globally set the number of threads for all operations from LightGBM, which will be used to limit the number of threads to 2 for the R-package tests. This is meant only for internal use so it isn't documented but we could if someone thinks it would be useful.

Fixes #5102.

### DIFF
--- a/R-package/tests/testthat.R
+++ b/R-package/tests/testthat.R
@@ -1,4 +1,4 @@
-Sys.setenv(OMP_NUM_THREADS = "2")
+Sys.setenv(LGBM_DEFAULT_NUM_THREADS = "2")
 library(testthat)
 library(lightgbm)
 
--- a/R-package/tests/testthat.R
+++ b/R-package/tests/testthat.R
@@ -1,3 +1,4 @@
+Sys.setenv(OMP_NUM_THREADS = "2")
 library(testthat)
 library(lightgbm)
 
--- a/R-package/tests/testthat/test_basic.R
+++ b/R-package/tests/testthat/test_basic.R
@@ -1,6 +1,9 @@
 VERBOSITY <- as.integer(
   Sys.getenv("LIGHTGBM_TEST_VERBOSITY", "-1")
 )
+NUM_THREADS <- as.integer(
+  Sys.getenv("LGB_NUM_THREADS", "2")
+)
 
 ON_WINDOWS <- .Platform$OS.type == "windows"
 
@@ -79,6 +82,7 @@ test_that("train and predict binary classification", {
         , objective = "binary"
         , metric = "binary_error"
         , verbose = VERBOSITY
+        , num_threads = NUM_THREADS
     )
     , nrounds = nrounds
     , valids = list(
@@ -120,6 +124,7 @@ test_that("train and predict softmax", {
         , metric = "multi_error"
         , num_class = 3L
         , verbose = VERBOSITY
+        , num_threads = NUM_THREADS
     )
     , nrounds = 20L
     , valids = list(
@@ -150,6 +155,7 @@ test_that("use of multiple eval metrics works", {
         , objective = "binary"
         , metric = metrics
         , verbose = VERBOSITY
+        , num_threads = NUM_THREADS
     )
     , nrounds = 10L
     , valids = list(
@@ -179,6 +185,7 @@ test_that("lgb.Booster.upper_bound() and lgb.Booster.lower_bound() work as expec
         , objective = "binary"
         , metric = "binary_error"
         , verbose = VERBOSITY
+        , num_threads = NUM_THREADS
     )
     , nrounds = nrounds
   )
@@ -197,6 +204,7 @@ test_that("lgb.Booster.upper_bound() and lgb.Booster.lower_bound() work as expec
         , objective = "regression"
         , metric = "l2"
         , verbose = VERBOSITY
+        , num_threads = NUM_THREADS
     )
     , nrounds = nrounds
   )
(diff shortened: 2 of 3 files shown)


jmoralez:
@jameslamb trying the suggestion from https://github.com/microsoft/LightGBM/issues/5102#issuecomment-1183972529 I found that we needed to set the parameters in many more places, and for example `predict` doesn't take parameters so it involved more changes than expected. I took a different approach here but with the same idea, I've updated the description accordingly. Please let me know what you think.

jameslamb:
@jmoralez ah ok, the point about `predict()` makes sense.

But to be honest, I don't understand how the new approach using `Sys.setenv(LGBM_DEFAULT_NUM_THREADS = "2")` at the top of `testthat.R` is working when `Sys.setenv("OMP_NUM_THREADS" = 2)` didn't (https://github.com/microsoft/LightGBM/issues/5102#issuecomment-1183440229). I thought the reason `Sys.setenv()` didn't work for you before was because it didn't affect the environment for test cases, as a result of how `{testthat}` invokes your test code.

How exactly are you testing that this change restricts LightGBM to using 2 threads?

-----

Seeing how much effort this is taking, and how many different places have to be modified, I think we might want to consider closing this and instead focusing on #4705, then using the solution to #4705 to satisfy #5102.

jmoralez:
> how the new approach using Sys.setenv(LGBM_DEFAULT_NUM_THREADS = "2") at the top of testthat.R is working when Sys.setenv("OMP_NUM_THREADS" = 2) didn't

I was relying on the OMP to check for the env variable, this time I'm manually reading it.

> I thought the reason Sys.setenv() didn't work for you before was because it didn't affect the environment for test cases, as a result of how {testthat} invokes your test code.

It was due to how OMP works, it just reads the env variable at program startup, not at runtime.

> How exactly are you testing that this change restricts LightGBM to using 2 threads?

Looking at the htop while running `Rscript testthat.R` and seeing that the CPU usage is below 200%. But that's just to make sure, these changes make it so that whenever we try to set the omp_num_threads we actually set the content of that env variable, so all other sources are ignored.

> Seeing how much effort this is taking, and how many different places have to be modified, I think we might want to consider closing this and instead focusing on https://github.com/microsoft/LightGBM/issues/4705, then using the solution to https://github.com/microsoft/LightGBM/issues/4705 to satisfy https://github.com/microsoft/LightGBM/issues/5102.

I think that #4705 has two possible solutions:
1. Do something like a context manager as suggested in https://github.com/microsoft/LightGBM/issues/4705#issuecomment-1057525376, where whenever we want to modify the number of threads we set the omp_num_threads to that number but afterwards we restore it to its original value.
2. Use a different variable to store the number of threads and pass it to the pragmas, that'd be something like what data.table does but it's hard because there are +200 parallel for.

In either case I think it wouldn't help with #5102 because for that we want to globally set the number of threads to a number or set an upper bound for them. And in each of them we could use this approach because if we do 1. we can leave the proposed code here as is and it would work, and in 2. we could modify this to set the default number of threads for lightgbm instead of OMP.

jameslamb:
ok @jmoralez now that we got v3.3.3 out, I'm ready to come back and help move this forward.

I'm sorry, but I don't support expanding LightGBM's API this way, or adding another layer of complexity to how LightGBM chooses how many threads to use. We can say "this environment variable is internal only", but we can't actually enforce that, and this PR would add the first use of `std::getenv()` in the package which might slightly increase the risk of running into portability problems (e.g. that working differently on different operating systems, or in different locales).

I don't think that's worth it for the narrow purpose of "more closely follow CRAN's *suggested* guidelines, which they've never complained about before".

Given that CRAN has never rejected a submission for `{lightgbm}`'s use of more than 2 threads, I think it is ok to use an approach which possibly misses a few cases, in exchange for not adding a new entrypoint (e.g. the first LightGBM-specific environment variable) in the actual code in `lib_lightgbm` or the R package.

I recommend that we just do what #5102 initially proposed in its description, and follow the same approach that was used for verbosity in #4862.

Thank you again for investigating these other different options in this PR! But I think we should just move forward with something that is simple and known to work for this case, even though it feels inelegant and might miss a few cases.

jameslamb:
@jmoralez I saw you closed this.

Are you planning to open another PR implementing the approach I suggested? If not, would you support one if I did?

Totally understand if you feel like you don't want to put any more effort into this particular issue.

jameslamb:
Ok thanks @jmoralez ! And great points about `predict` and Dataset construction.

I will pick this up at some point (in a new PR).

PR #5379, opened 2022-07-18
[R-package] [docs] fix typo in documentation on init_model

### LINKED ISSUE #5377: [R-package] init_model should accept booster
## Summary

It would be great if one could pass the lgb.Booster directly as `init_model` and not have to save it to disk first.

sebffischer:
Fixes #5377.

### DIFF
--- a/R-package/R/lightgbm.R
+++ b/R-package/R/lightgbm.R
@@ -44,7 +44,7 @@
 #'                 }
 #'             }
 #' @param eval_freq evaluation output frequency, only effective when verbose > 0 and \code{valids} has been provided
-#' @param init_model path or model file of \code{lgb.Booster} object, will continue training from this model
+#' @param init_model path of model file or \code{lgb.Booster} object, will continue training from this model
 #' @param nrounds number of training rounds
 #' @param obj objective function, can be character or custom objective function. Examples include
 #'            \code{regression}, \code{regression_l1}, \code{huber},
--- a/R-package/R/lightgbm.R
+++ b/R-package/R/lightgbm.R
@@ -44,7 +44,7 @@
 #'                 }
 #'             }
 #' @param eval_freq evaluation output frequency, only effective when verbose > 0 and \code{valids} has been provided
-#' @param init_model path of model file of \code{lgb.Booster} object, will continue training from this model
+#' @param init_model path or model file of \code{lgb.Booster} object, will continue training from this model
 #' @param nrounds number of training rounds
 #' @param obj objective function, can be character or custom objective function. Examples include
 #'            \code{regression}, \code{regression_l1}, \code{huber},
--- a/R-package/man/lgb.cv.Rd
+++ b/R-package/man/lgb.cv.Rd
@@ -101,7 +101,7 @@ by the values of outcome labels.}
 (each element must be a vector of test fold's indices). When folds are supplied,
 the \code{nfold} and \code{stratified} parameters are ignored.}
 
-\item{init_model}{path of model file of \code{lgb.Booster} object, will continue training from this model}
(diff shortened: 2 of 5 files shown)


jameslamb (changes requested):
Thanks very much for this!

First, can you please re-generate the documentation for the package? We keep the `.Rd` files (these are what R uses to power the help docs shown with e.g. `?lightgbm::lightgbm`) in source control.

You could do that by running the following:

```shell
cd R-package
Rscript --vanilla -e "roxygen2::roxygenize(load = 'installed')"
```

If you don't want to do that, please give me access to your fork and I'd be happy to make those changes and push them to this PR.

-----

Second, I've added `fixes #5377` to the description. Whenever you open a pull request here, please provide a description that helps maintainers to understand why you're proposing what you're proposing. And whenever some prior discussions on GitHub led to the PR or are relevant to it, please link them.

-----

In the future when you contribute here or in other projects on GitHub, I encourage you to create new branches on your fork instead of modifying `master` / `main`. That way:

* you can work on multiple pull requests at the same time
* it's easier to ensure your fork's `master` / `main` branch will have the same history as the one in the upstream project when changes are made upstream

-----

Let me know if you have any questions, I'm here to help 👋

R-package/R/lightgbm.R
```diff
@@ -44,7 +44,7 @@
 #'                 }
 #'             }
 #' @param eval_freq evaluation output frequency, only effective when verbose > 0 and \code{valids} has been provided
-#' @param init_model path of model file of \code{lgb.Booster} object, will continue training from this model
+#' @param init_model path or model file of \code{lgb.Booster} object, will continue training from this model
```
jmoralez, line 47:
```suggestion
#' @param init_model path of model file or \code{lgb.Booster} object, will continue training from this model
```

jameslamb: ha didn't notice this! Yes, agree with @jmoralez , please make this change.

sebffischer:
Thanks! Maybe the instructions on  how to build the docs could be added [here](https://lightgbm.readthedocs.io/en/latest/README.html
)

I ran `devtools::document()` which did not work and then I - incorrectly - assumed the docs might be build automatically in the CI.

Also one has to run `git submodule update --init --recursive` before to get the external libs

jameslamb:
> Maybe the instructions on how to build the docs could be added

Yep you're right! Sorry about that. I promised @jmoralez I'd do that a long time ago and never did 🙈 . Will do that right now.

> one has to run git submodule update --init --recursive

Yes that's right. That is mentioned in multiple places in https://github.com/microsoft/LightGBM/tree/master/R-package and https://lightgbm.readthedocs.io/en/v3.3.2/Installation-Guide.html, although in a slightly different way.

LightGBM's documentation recommends setting up all the submodules at the time that you clone, like this:

```shell
git clone \
    --recursive \
    https://github.com/microsoft/LightGBM
```

But if you've already cloned without `--recursive`, then you're correct that updating them for your existing repo requires running

```shell
git submodule init
git submodule update --recursive
```

jameslamb (approved):
Changes look great to me, thanks very much for your contribution and for using LightGBM 🎉

StrikerRUS (approved):

jameslamb:
Sorry for the delay! We had some continuous integration (CI) problems to fix, which have been resolved in #5388 .

I just updated this PR to the latest changes in `master`, and once CI runs successfully I'll merge it.

jameslamb:
Thanks again for the contribution @sebffischer !

Now that this is merged, `master` here in LightGBM will not match `master` on your fork. I recommend deleting your fork of the repo. If you come back to contribute in the future (we hope you will!), use a branch other than `master` and you won't have to worry about it.

PR #5386, opened 2022-07-24
[python-package] Improve readme on experimental cuda usage

thomasaarholt:
I've edited the Readme to make clear a few minor things that caught me by surprise.

### DIFF
--- a/python-package/README.rst
+++ b/python-package/README.rst
@@ -37,7 +37,7 @@ For **macOS** (we provide wheels for 3 newest macOS versions) users:
 - For version smaller than 2.1.2, **gcc-7** with **OpenMP** is required.
 
 Build from Sources
-******************
+''''''''''''''''''
 
 .. code:: sh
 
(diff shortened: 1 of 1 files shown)


python-package/README.rst
```diff
@@ -37,7 +37,7 @@ For **macOS** (we provide wheels for 3 newest macOS versions) users:
 - For version smaller than 2.1.2, **gcc-7** with **OpenMP** is required.
 
 Build from Sources
-******************
+''''''''''''''''''
```
jameslamb, line 40: Can you please explain why this change to the heading level is necessary?

jameslamb (changes requested):
Thanks for using LightGBM. I left one question, will review again after that's addressed.

python-package/README.rst
thomasaarholt: Sure. On current master, the header "Build from Sources" is a level below ["Install from [PyPI]"](https://github.com/microsoft/LightGBM/tree/master/python-package#install-from-pypi). 
Building from source does use pip, but does not use PyPi. This change places the two headers at the same level.

If you would rather I omit that change, that's fine.

thomasaarholt: Before writing this issue I wanted to link to one of the sub-headings in the documentation so I could refer to it in a different issue. For some reason I got the impression that the lower-level headers didn't let me hyperlink to them directly. That wasn't right (I've just hyperlinked to ["Build from Sources"](https://github.com/microsoft/LightGBM/tree/master/python-package#build-from-sources) in the comment above), but that is what led me to make this change.

jameslamb: > Building from source does use pip, but does not use PyPi.

It's true that using `pip` generally doesn't rely on a package repository like PyPI and that it can be used with local files. However, all of the commands in the section you've linked are specifically about how to build from source with a package served by PyPI.

Notice that none of the commands in that section include a filepath, like `.` or `python-package/`. And also notice that if you try one of those commands, you'll see a warning like the following:

> UserWarning: Disabling all use of wheels due to the use of --build-option / --global-option / --install-option

So a command like `pip install lightgbm --install-option=--cuda` means "download the source (non-wheel) distribution from PyPI, and build a version of the Python package using a `lib_lightgbm.so` compiled with `-DUSE_CUDA=ON`".

If you scroll down a bit further, you'll see that there is a separate section, https://github.com/microsoft/LightGBM/tree/master/python-package#install-from-github, that covers building from source in a local clone of this git repo.

Please revert this change.

thomasaarholt: Don't worry - I don't feel strongly about this 😃 Change reverted.

shiyu1994:
Close and reopen to retrigger ci tests.

python-package/README.rst
jameslamb: thanks very much

jameslamb (commented):
Thanks! I'm going to dismiss my review so it doesn't block merging, but will defer to @shiyu1994 's opinion about the changes since he is the author oof the `cuda_exp` implementation.

shiyu1994 (approved):
LGTM. Thanks! We will add an option to use single copy of memory for the new CUDA version.

Issue #5390, opened 2022-07-26
[ci] [dask] CI jobs failing with Dask 2022.7.1


jameslamb:
## Description

Created from https://github.com/microsoft/LightGBM/pull/5388#issuecomment-1195848451.

All CUDA CI jobs and several Linux jobs are failing with the following.

```text
FAILED ../tests/python_package_test/test_dask.py::test_machines_should_be_used_if_provided[binary-classification]
FAILED ../tests/python_package_test/test_dask.py::test_machines_should_be_used_if_provided[multiclass-classification]
FAILED ../tests/python_package_test/test_dask.py::test_machines_should_be_used_if_provided[regression]
FAILED ../tests/python_package_test/test_dask.py::test_machines_should_be_used_if_provided[ranking]
= 4 failed, 700 passed, 10 skipped, 2 xfailed, 395 warnings in 655.51s (0:10:55) =
```

`client.restart()` calls in that test are resulting in the following:

> raise TimeoutError(f"{len(bad_nannies)}/{len(nannies)} nanny worker(s) did not shut down within {timeout}s")
> E   asyncio.exceptions.TimeoutError: 1/2 nanny worker(s) did not shut down within 120s

<details><summary>traceback (click me)</summary>

```text
/root/miniforge/envs/test-env/lib/python3.9/site-packages/distributed/client.py:3360: in restart
    return self.sync(
/root/miniforge/envs/test-env/lib/python3.9/site-packages/distributed/utils.py:338: in sync
    return sync(
/root/miniforge/envs/test-env/lib/python3.9/site-packages/distributed/utils.py:405: in sync
    raise exc.with_traceback(tb)
/root/miniforge/envs/test-env/lib/python3.9/site-packages/distributed/utils.py:378: in f
    result = yield future
/root/miniforge/envs/test-env/lib/python3.9/site-packages/tornado/gen.py:762: in run
    value = future.result()
/root/miniforge/envs/test-env/lib/python3.9/site-packages/distributed/client.py:3329: in _restart
    await self.scheduler.restart(timeout=timeout, wait_for_workers=wait_for_workers)
/root/miniforge/envs/test-env/lib/python3.9/site-packages/distributed/core.py:1153: in send_recv_from_rpc
    return await send_recv(comm=comm, op=key, **kwargs)
/root/miniforge/envs/test-env/lib/python3.9/site-packages/distributed/core.py:943: in send_recv
    raise exc.with_traceback(tb)
[...]

PR #5398, opened 2022-08-02
[ci] temporarily pin scipy version and fix Python linting error

StrikerRUS:
Unblock CI in this repo.

Temp workaround for #5390.

### DIFF
--- a/.ci/test.sh
+++ b/.ci/test.sh
@@ -120,9 +120,9 @@ fi
 
 # temporary fix for https://github.com/microsoft/LightGBM/issues/5390
 if [[ $PYTHON_VERSION == "3.7" ]]; then
-    DASK_DEPENDENCIES="dask distributed"
+    DEPENDENCIES="dask distributed"
 else
-    DASK_DEPENDENCIES="dask=2022.7.0 distributed=2022.7.0"
+    DEPENDENCIES="dask=2022.7.0 distributed=2022.7.0 scipy<1.9"
 fi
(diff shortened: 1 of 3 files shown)


jameslamb (approved):
Thanks for this! I can work on trying to remove these pins in the future.

----

I checked, and I'm guessing the new linting error ([build link](https://github.com/microsoft/LightGBM/runs/7639284498?check_suite_focus=true)) is caused by the release of `pycodestyle` 2.9.0 about 3 days ago: https://pypi.org/project/pycodestyle/#history. Maybe it just made it to `conda-forge`.

```text
Linting Python code
./python-package/lightgbm/basic.py:1510:31: E275 missing whitespace after keyword
```

I don't need to re-approve when you fix that...I'm in favor of altering the code in `basic.py` to comply with what `pycodestyle` wants, but I trust whatever decision you make about that.

StrikerRUS:
At first, I thought that that linting CI job fail is a false positive, but after a precise check I see we really have a linting error in our codebase. No whitespace after `not` operator:
https://github.com/microsoft/LightGBM/blob/cdc83a5be585e26091c45d68d5acc6d95e7ce31f/python-package/lightgbm/basic.py#L1510

https://stackoverflow.com/a/59395093

jameslamb:
> No whitespace after not operator:

Wow! I stared at that line for a long time and also wasn't sure what it was complaining about. Great work.

PR #5406, opened 2022-08-07
[python-package] add type hints on Booster.save_model()

jameslamb:
Contributes to #3756.

Adds type hints on `Booster.save_model()`.


StrikerRUS (approved):
Thanks!

Issue #5313, opened 2022-06-21
[python-package] prefix all internal object names with "_"


jameslamb:
## Short Description

I'm opening this RFC to propose the following:

> **all objects in the Python package which are intended to be internal should have names beginning with `_`**

The convention "anything starting with `_` is internal" is common in Python programming. If LightGBM's Python package followed that convention consistently, it would simplify LightGBM's position on which parts of the Python package are considered "internal" and subject to breaking changes in any release.

## Detailed Description

In pursuit of the v4.0.0 release (#5153), we have been accepting breaking changes which simplify the project and reduce its maintenance burden, so that limited maintainer time can be focused on providing a good experience with LightGBM's main functionality.

I think "all internal objects should begin with `_`" is one such type of change. Making it clearer which parts of the Python package are considered internal reduces the surface area of the public API which users need to consider, and increases contributors' flexibility to change implementations without breaking user code.

R packages support a "namespace" concept (see ["Writing R extensions"](https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Package-namespaces)), which clearly separates exported and internal objects, and I think that mechanism has been really helpful. The closest thing that Python has, to my knowledge, is the `__all__` member for modules (https://docs.python.org/3/tutorial/modules.html#importing-from-a-package). `__all__` controls the behavior of "star imports".

For example, if you run the following:

```python
from lightgbm import *
dir()
```

the only LightGBM objects you'll see are those in the package's `__all__`.

This mechanism is a good start, but prefixing all internal objects with `_` would be an even stronger signal about which things are internal. It would also improve the consistency of the Python package's code.

## Specific Proposal

<details><summary>expand to see original proposal (click me)</summary>

I am proposing that the following be prefixed with a `_` and treated as internal:

* https://github.com/microsoft/LightGBM/blob/ff947bfac5148a9b94d045f2879aee183e02e174/python-package/lightgbm/basic.py#L128
* https://github.com/microsoft/LightGBM/blob/ff947bfac5148a9b94d045f2879aee183e02e174/python-package/lightgbm/basic.py#L144
* https://github.com/microsoft/LightGBM/blob/ff947bfac5148a9b94d045f2879aee183e02e174/python-package/lightgbm/basic.py#L155
* https://github.com/microsoft/LightGBM/blob/ff947bfac5148a9b94d045f2879aee183e02e174/python-package/lightgbm/basic.py#L160
[...]

PR #5409, opened 2022-08-08
[python-package] prefix NUMERIC_TYPES with _ to make it a internal object

thomasjpfan:
Contributes to #5313


jameslamb (approved):
Thanks so much for helping out @thomasjpfan !

I altered your PR title slightly, hope you don't mind. PR titles become release note items in this project, and we use the convention `[python-package]` for Python-specific PRs.

PR #5412, opened 2022-08-10
[python-package] ignore training set on early stopping callback (fixes #5354)

### LINKED ISSUE #5354: [python-package] Early Stopping does not work as expected
## Description

I found that when early stopping is enabled and there're multiple validation sets in valid_sets, the LightGBM might won't save the best model as we expected.

## Reproducible example

```
np.random.seed(2022)
toy_pd = pd.DataFrame({'feature1':np.random.randn(1000),'feature2':np.random.randn(1000),'label':np.concatenate((np.ones(500), np.zeros(500)))})
feats = [f for f in toy_pd if f not in ['label']]
print(feats)


params = {
    'learning_rate': 0.07,
    'boosting_type': 'gbdt',
    'objective': 'binary',
    'metric': ['auc'],
    'is_unbalance': False,
    'lambda_l1':0.5,
    'lambda_l2':10,
    'num_leaves':32,
    'max_depth':-1,
    'min_child_samples': 15, 
    'verbose': -1,
    'seed': 42,
    'n_jobs': 8,
}

fold_num = 5
kf = KFold(n_splits=fold_num, shuffle=True, random_state=42)
for fold, (train_idx, val_idx) in enumerate(kf.split(toy_pd[feats], toy_pd['label'])):
    print('-----------', fold, '-----------')
    train = lgb.Dataset(toy_pd.loc[train_idx, feats],
                        toy_pd.loc[train_idx,'label'])
    val = lgb.Dataset(toy_pd.loc[val_idx, feats],
                      toy_pd.loc[val_idx, 'label'])
    model = lgb.train(params, train, valid_sets=[train, val], num_boost_round=4,
                      callbacks=[lgb.early_stopping(2), lgb.log_evaluation(1)])


```
output:
```
----------- 0 -----------
[1]	training's auc: 0.63893	valid_1's auc: 0.582732
Training until validation scores don't improve for 2 rounds
[2]	training's auc: 0.639539	valid_1's auc: 0.580629
[3]	training's auc: 0.652984	valid_1's auc: 0.588742
[4]	training's auc: 0.678193	valid_1's auc: 0.570563
Did not meet early stopping. Best iteration is:
[4]	training's auc: 0.678193	valid_1's auc: 0.570563
...
----------- 3 -----------
[1]	training's auc: 0.649097	valid_1's auc: 0.575294
Training until validation scores don't improve for 2 rounds
[2]	training's auc: 0.657224	valid_1's auc: 0.566957
[3]	training's auc: 0.666932	valid_1's auc: 0.565524
Early stopping, best iteration is:
[1]	training's auc: 0.649097	valid_1's auc: 0.575294
```
As you can see, I clearly hope lgb could save the model that has the best auc on val because val is the last one in valid_sets list. However, in this example, it sometimes returns model which has the best performance on valid_sets[0], sometimes valid_sets[-1]. 

It seems that when it does not meet early stopping, something would go wrong.

I'm very confused about this. I fixed all random seeds so you can easily reproduce it.

### LightGBM version or commit hash:
'3.3.2'

### Command(s) you used to install LightGBM
```shell
pip install lightgbm
```

jmoralez:
Fixes #5354

### DIFF
--- a/python-package/lightgbm/callback.py
+++ b/python-package/lightgbm/callback.py
@@ -266,9 +266,9 @@ class _EarlyStoppingCallback:
         only_train_set = (
             len(env.evaluation_result_list) == 1
             and self._is_train_set(
-                env.evaluation_result_list[0][0],
-                env.evaluation_result_list[0][1].split(" ")[0],
-                env.model._train_data_name)
+                ds_name=env.evaluation_result_list[0][0],
+                eval_name=env.evaluation_result_list[0][1].split(" ")[0],
+                train_name=env.model._train_data_name)
         )
         self.enabled = not is_dart and not only_train_set
         if not self.enabled:
--- a/python-package/lightgbm/callback.py
+++ b/python-package/lightgbm/callback.py
@@ -258,11 +258,24 @@ class _EarlyStoppingCallback:
     def _lt_delta(self, curr_score: float, best_score: float, delta: float) -> bool:
         return curr_score < best_score - delta
 
+    def _is_train_set(self, ds_name: str, eval_name: str, train_name: str) -> bool:
+        return (ds_name == "cv_agg" and eval_name == "train") or ds_name == train_name
+
     def _init(self, env: CallbackEnv) -> None:
-        self.enabled = not any(env.params.get(boost_alias, "") == 'dart' for boost_alias
-                               in _ConfigAliases.get("boosting"))
+        is_dart = any(env.params.get(alias, "") == 'dart' for alias in _ConfigAliases.get("boosting"))
+        only_train_set = (
+            len(env.evaluation_result_list) == 1
+            and self._is_train_set(
+                env.evaluation_result_list[0][0],
+                env.evaluation_result_list[0][1].split(" ")[0],
+                env.model._train_data_name)
+        )
+        self.enabled = not is_dart and not only_train_set
         if not self.enabled:
-            _log_warning('Early stopping is not available in dart mode')
+            if is_dart:
+                _log_warning('Early stopping is not available in dart mode')
+            elif only_train_set:
+                _log_warning('Only training set found, disabling early stopping.')
             return
         if not env.evaluation_result_list:
             raise ValueError('For early stopping, '
@@ -339,9 +352,7 @@ class _EarlyStoppingCallback:
             eval_name_splitted = env.evaluation_result_list[i][1].split(" ")
             if self.first_metric_only and self.first_metric != eval_name_splitted[-1]:
                 continue  # use only the first metric for early stopping
-            if ((env.evaluation_result_list[i][0] == "cv_agg" and eval_name_splitted[0] == "train"
-                    or env.evaluation_result_list[i][0] == env.model._train_data_name)):
-                self._final_iteration_check(env, eval_name_splitted, i)
+            if self._is_train_set(env.evaluation_result_list[i][0], eval_name_splitted[0], env.model._train_data_name):
                 continue  # train data for lgb.cv or sklearn wrapper (underlying lgb.train)
             elif env.iteration - self.best_iter[i] >= self.stopping_rounds:
                 if self.verbose:
--- a/tests/python_package_test/test_engine.py
+++ b/tests/python_package_test/test_engine.py
@@ -796,7 +796,9 @@ def test_early_stopping_ignores_training_set(use_valid):
         assert eval_result['valid']['l2'][1] > eval_result['valid']['l2'][0]  # valid didn't
     else:
         with pytest.warns(UserWarning, match='Only training set found, disabling early stopping.'):
-            train_fn()
+            bst = train_fn()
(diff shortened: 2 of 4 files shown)


tests/python_package_test/test_utilities.py
```diff
@@ -29,17 +29,18 @@ def dummy_metric(_, __):
...
-    lgb_data = lgb.Dataset(X, y)
+    lgb_train = lgb.Dataset(X, y)
+    lgb_valid = lgb.Dataset(X, y)  # different object for early-stopping
 
     eval_records = {}
     callbacks = [
         lgb.record_evaluation(eval_records),
         lgb.log_evaluation(2),
-        lgb.early_stopping(4)
+        lgb.early_stopping(10)
```
jmoralez, line 39: I changed this because otherwise it would stop at the 4th iteration, this was due to the training set being ignored for early stopping but now early stopping is disabled if you only provide the training set as valid, so I had to provide a dummy validation set.

tests/python_package_test/test_sklearn.py
```diff
@@ -1124,11 +1124,6 @@ def fit_and_check(eval_set_names, metric_names, assumed_iteration, first_metric_
     iter_min = min([iter_min_l1, iter_min_l2])
     iter_min_valid1 = min([iter_valid1_l1, iter_valid1_l2])
 
-    # training data as eval_set
```
jmoralez, line 1127: I think we shouldn't test this anymore but I'm open to modify it by adding an extra validation set so that we can check the training results

StrikerRUS (approved):
Thank you!

python-package/lightgbm/callback.py
```diff
@@ -258,11 +258,24 @@ def _gt_delta(self, curr_score: float, best_score: float, delta: float) -> bool:
...
     def _init(self, env: CallbackEnv) -> None:
-        self.enabled = not any(env.params.get(boost_alias, "") == 'dart' for boost_alias
-                               in _ConfigAliases.get("boosting"))
+        is_dart = any(env.params.get(alias, "") == 'dart' for alias in _ConfigAliases.get("boosting"))
+        only_train_set = (
+            len(env.evaluation_result_list) == 1
+            and self._is_train_set(
+                env.evaluation_result_list[0][0],
+                env.evaluation_result_list[0][1].split(" ")[0],
+                env.model._train_data_name)
```
jameslamb, line 271:
```suggestion
                ds_name=env.evaluation_result_list[0][0],
                eval_name=env.evaluation_result_list[0][1].split(" ")[0],
                train_name=env.model._train_data_name)
```

Can you please use keyword arguments here? I think this makes it easier to read this code and understand what it's doing, and that habit prevents mistakes of the form "accidentally passed arguments in the wrong order", which is extra important in cases like this where the arguments have the same type and overlapping ranges of valid values.

tests/python_package_test/test_engine.py
```diff
@@ -764,6 +764,41 @@ def test_early_stopping():
...
+    valid_sets = [train_ds]
+    valid_names = ['train']
+    if use_valid:
+        valid_sets.append(valid_ds)
+        valid_names.append('valid')
+    eval_result = {}
+
+    def train_fn():
+        return lgb.train(
+            {'num_leaves': 5},
```
jameslamb, line 785:
```suggestion
            params={'num_leaves': 5, 'num_iterations': 10},
```

To save some CI time, would you consider reducing the number of boosting rounds? ([the default is 100](https://lightgbm.readthedocs.io/en/latest/Parameters.html#num_iterations))

tests/python_package_test/test_engine.py
```diff
@@ -764,6 +764,41 @@ def test_early_stopping():
...
+            callbacks=[lgb.early_stopping(1), lgb.record_evaluation(eval_result)]
+        )
+    if use_valid:
+        bst = train_fn()
+        assert bst.best_iteration == 1
+        assert eval_result['train']['l2'][1] < eval_result['train']['l2'][0]  # train improved
+        assert eval_result['valid']['l2'][1] > eval_result['valid']['l2'][0]  # valid didn't
+    else:
+        with pytest.warns(UserWarning, match='Only training set found, disabling early stopping.'):
+            train_fn()
```
jameslamb, line 799:
```suggestion
            bst = train_fn()
        assert bst.current_iteration() == 10
        assert bst.best_iteration == 0
```

Instead of only checking that this warning is raised, can you also please add some checks that early stopping was not actually triggered?

That would catch cases where this warning is raised but early stopping is still accidentally enabled.

jameslamb (changes requested):
Tests and logic changes look good to me. Thanks for working on this!

Please see a few small suggestions.

tests/python_package_test/test_engine.py
jameslamb: oh sorry, missed that!

jameslamb (approved):
very nice, thanks for this!

Issue #4904, opened 2021-12-22
Loading binary dataset does not consider the creation non-default params


OfekShilon:
## Description
When we create a dataset with non default parameters, save it and load it - `construct()` breaks.

## Reproducible example
```r
library(lightgbm)
nn <- 1e5
xx <- cbind(x=seq.int(nn)%%5, y=seq.int(nn)%%17)
yy <- sin(seq.int(nn))

## lgb.dataset.params <- list(bin_construct_sample_cnt=floor(nn/2), max_bin=16)
lgb.dataset.params <- list(bin_construct_sample_cnt=floor(nn/2))

ref.data <- lgb.Dataset(data=xx, label=yy, params=lgb.dataset.params)
cat("#constructing ref.data with bin_construct_sample_cnt params, [SUCCEEDS]\n")
ref.data$construct()
ref.data.fn <- tempfile()
ref.data$save_binary(ref.data.fn)
ref.data2 <- lgb.Dataset(ref.data.fn)    # In real life scenarios it is hard to pass here the original construction params, and they're available in the file anyway.
ref.data2$construct()
```
Gives error:
```
[LightGBM] [Fatal] Dataset bin_construct_sample_cnt 50000 != config 200000
Error in ref.data2$construct() : 
  Dataset bin_construct_sample_cnt 50000 != config 200000
```
## Environment info
```
> sessionInfo()
R version 4.0.5
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 18.04.5 LTS

Matrix products: default
BLAS:   /opt/R-4.0.5.mkl2020/lib64/R/lib/libRblas.so
LAPACK: /opt/R-4.0.5.mkl2020/lib64/R/lib/libRlapack.so

Random number generation:
 RNG:     Mersenne-Twister 
 Normal:  Inversion 
 Sample:  Rounding 
 
locale:
 [1] LC_CTYPE=en_US.UTF-8 LC_NUMERIC=C         LC_TIME=C            LC_COLLATE=C         LC_MONETARY=C        LC_MESSAGES=C        LC_PAPER=C          
 [8] LC_NAME=C            LC_ADDRESS=C         LC_TELEPHONE=C       LC_MEASUREMENT=C     LC_IDENTIFICATION=C 

attached base packages:
[1] datasets  utils     stats     graphics  grDevices methods   base     

other attached packages:
[...]

PR #5416, opened 2022-08-11
include parameters from reference dataset on subset (fixes #5402)

### LINKED ISSUE #5402: Failed to load Dataset.subset() back after Dataset.save_binary()
## Description
I'd like to split a large dataset into several subsets and use them later, via Dataset.subset() API.
I really like the LightGBM binary dataset, that saves the memory and disk, so I tried to use it everywhere. But I found the following workflow doesn't work:

1. Construct and save a binary dataset.
2. Load a subset from the binary dataset.
3. Save the subset.
4. [Failed] Load the subset back.

## Reproducible example
```python
import lightgbm as lgb
import numpy as np

# Create and save the data
data = np.random.random((100,10))
ds = lgb.Dataset(data).construct()
ds.save_binary('train.bin')

# Load, create, and save a subset
ds = lgb.Dataset('train.bin')
subset = ds.subset([1,2,3,5,8]).construct()
print(f'Got {subset.num_data()} samples from {ds.num_data()} samples')
subset.save_binary('subset.bin')

# Load but failed
subset = lgb.Dataset('subset.bin').construct()
```

Error message:
```
[LightGBM] [Info] Saving data to binary file train.bin
[LightGBM] [Info] Load from binary file train.bin
Got 5 samples from 100 samples
[LightGBM] [Info] Saving data to binary file subset.bin
[LightGBM] [Info] Load from binary file subset.bin
[LightGBM] [Fatal] Dataset max_bin 140839269 != config 255
  File "load_bin.py", line 16, in <module>
    subset = lgb.Dataset('subset.bin').construct()
  File "D:\yuri_env\lib\site-packages\lightgbm\basic.py", line 1815, in construct
    self._lazy_init(self.data, label=self.label,
  File "D:\yuri_env\lib\site-packages\lightgbm\basic.py", line 1528, in _lazy_init
    _safe_call(_LIB.LGBM_DatasetCreateFromFile(
  File "D:\yuri_env\lib\site-packages\lightgbm\basic.py", line 125, in _safe_call
    raise LightGBMError(_LIB.LGBM_GetLastError().decode('utf-8'))
lightgbm.basic.LightGBMError: Dataset max_bin 140839269 != config 255
```

## Environment info

LightGBM 3.3.2

```shell
pip install lightgbm
```
Other environments:
- Python 3.8.8
- numpy 1.21.5

jmoralez:
Fixes #5402

### DIFF
--- a/tests/python_package_test/test_basic.py
+++ b/tests/python_package_test/test_basic.py
@@ -243,6 +243,13 @@ def test_chunked_dataset_linear():
     valid_data.construct()
 
 
+def test_save_dataset_subset_and_load_from_file(tmp_path):
+    data = np.random.rand(100, 2)
+    ds = lgb.Dataset(data)
+    ds.subset([1,2,3,5,8]).save_binary(tmp_path / 'subset.bin')
+    lgb.Dataset(tmp_path / 'subset.bin').construct()
+
+
 def test_subset_group():
     rank_example_dir = Path(__file__).absolute().parents[2] / 'examples' / 'lambdarank'
     X_train, y_train = load_svmlight_file(str(rank_example_dir / 'rank.train'))
--- a/tests/python_package_test/test_basic.py
+++ b/tests/python_package_test/test_basic.py
@@ -245,9 +245,10 @@ def test_chunked_dataset_linear():
 
 def test_save_dataset_subset_and_load_from_file(tmp_path):
     data = np.random.rand(100, 2)
-    ds = lgb.Dataset(data)
+    params = {'max_bin': 50, 'min_data_in_bin': 10}
+    ds = lgb.Dataset(data, params=params)
     ds.subset([1, 2, 3, 5, 8]).save_binary(tmp_path / 'subset.bin')
-    lgb.Dataset(tmp_path / 'subset.bin').construct()
+    lgb.Dataset(tmp_path / 'subset.bin', params=params).construct()
 
 
 def test_subset_group():
--- a/tests/python_package_test/test_basic.py
+++ b/tests/python_package_test/test_basic.py
@@ -246,7 +246,7 @@ def test_chunked_dataset_linear():
 def test_save_dataset_subset_and_load_from_file(tmp_path):
     data = np.random.rand(100, 2)
     ds = lgb.Dataset(data)
-    ds.subset([1,2,3,5,8]).save_binary(tmp_path / 'subset.bin')
+    ds.subset([1, 2, 3, 5, 8]).save_binary(tmp_path / 'subset.bin')
     lgb.Dataset(tmp_path / 'subset.bin').construct()
 
 
--- a/src/io/dataset.cpp
+++ b/src/io/dataset.cpp
@@ -740,6 +740,11 @@ void Dataset::CopyFeatureMapperFrom(const Dataset* dataset) {
   group_feature_cnt_ = dataset->group_feature_cnt_;
   forced_bin_bounds_ = dataset->forced_bin_bounds_;
   feature_need_push_zeros_ = dataset->feature_need_push_zeros_;
+  max_bin_ = dataset->max_bin_;
+  min_data_in_bin_ = dataset->min_data_in_bin_;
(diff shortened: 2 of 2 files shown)


StrikerRUS:
Should we also add  `device_type_` and `gpu_device_id_`?
https://github.com/microsoft/LightGBM/blob/5ef46f93e2fa095b3f93b3b5728ecf024a2caaad/src/io/dataset.cpp#L789-L790

jmoralez:
`CopySubrow` is called after `CopyFeatureMapperFrom` when creating the subset
https://github.com/microsoft/LightGBM/blob/6b695c292507571cecf3251be10524c648e73be0/src/c_api.cpp#L1554
and those attributes are set there https://github.com/microsoft/LightGBM/blob/5ef46f93e2fa095b3f93b3b5728ecf024a2caaad/src/io/dataset.cpp#L845-L846
https://github.com/microsoft/LightGBM/blob/5ef46f93e2fa095b3f93b3b5728ecf024a2caaad/src/io/dataset.cpp#L857-L858

tests/python_package_test/test_basic.py
```diff
@@ -243,6 +243,13 @@ def test_chunked_dataset_linear():
     valid_data.construct()
 
 
+def test_save_dataset_subset_and_load_from_file(tmp_path):
+    data = np.random.rand(100, 2)
+    ds = lgb.Dataset(data)
```
jameslamb, line 248: Could you please try overriding the defaults (e.g. setting `max_bin: 7` or something) and check that those non-default values survive the round trip to disk?

I think that would increase our confidence that this is working as expected. Otherwise, I think a bug of the form "all parameter values are lost when writing to binary Dataset file" could make it through this test.

jameslamb (changes requested):
Thanks for this! Please see my suggestion about making the test stricter.

tests/python_package_test/test_basic.py
jmoralez: I think that will raise an error like the one in #4902, but I'll try it and confirm here.

jameslamb: Ok I see. I expected a PR called "include parameters" to test that parameters on either side of an operation had the same values.

I think maybe I got confused by the presence of writing and reading a binary file, and thought the issue was specific to storing a `Dataset` to disk. Is it like *"when taking a subset, not all parameters are copied from the reference Dataset to the subset...and this can show up as an error loading the Dataset from file"*?

If it is, then it would be great to be able to reach into the Dataset (on the C++ side, not the Python object) and check that attributes like `max_bin_` are the same and a non-default value. But I'm not sure how to do that without introducing a new `c_api` entrypoint.

So if adding the test I suggested does hit the error from #4904, then I think this test in its current state is ok. It is still an improvement that fixes a bug.

jmoralez: I built the dataset with non default parameters and check that loading it with the same ones succeeds in 6a2fd1f.

jameslamb: Ok nice! Thank you for that and for the explanation.

I tried running this PR's test code on latest `master`, and can see based on the error message that those non-default parameter values are being respected.

```python
import lightgbm as lgb
import numpy as np

data = np.random.rand(100, 2)
params = {'max_bin': 50, 'min_data_in_bin': 10}
ds = lgb.Dataset(data, params=params).construct()
ds.subset([1, 2, 3, 5, 8]).save_binary('subset.bin')
lgb.Dataset('subset.bin', params=params).construct()
```

> [LightGBM] [Fatal] Dataset was constructed with parameter max_bin=32649. It cannot be changed to **50** when loading from binary file.

jameslamb (approved):
Thanks for this!

I ran the test code on latest `master` (https://github.com/microsoft/LightGBM/commit/865c126a1e3ccdd77ec205b9dde46e5f3c5b6b21) and confirmed that this test fails with the error

> [LightGBM] [Fatal] Dataset was constructed with parameter max_bin=32649. It cannot be changed to 50 when loading from binary file.

The fix here looks correct to me.

StrikerRUS (approved):
Thank you!

PR #5246, opened 2022-05-27
[python] Reapply the trained categorical columns when predicting

johnpaulett:
Fixes #5244.  During prediction, force any columns that were categorical during training to dtype `category` again.  Useful when hosted via kserve and the user is sending a HTTP JSON POST that will not natively get translated to a categorical column in the DataFrame.

Initially tried coding this change in `_data_from_pandas`, but elected to pull it into a separate method that is only called by `predict()`.  I'm open to any feedback or suggestion on how to better implement this change.

Issue #4323, opened 2021-05-26
[R-package] Accept data frames as inputs


jameslamb:
## Summary

It should be possible to use R data frames directly as inputs in `{lightgbm}`, without converting them to matrices.

## Motivation

The `data.frame` is a very common data structure in R, and many statistics and machine learning projects accept data frames as inputs, including but not limited to `{caret}`, `{randomForest}`, and the packages listed in https://github.com/microsoft/LightGBM/issues/4295#issuecomment-841753885.

Allowing the use of `data.frame` objects would reduce friction for R users working with LightGBM.

## Description

I think this work can be broken down into the following components

* [ ] accept `data.frame` inputs for `data` argument to `lgb.Dataset()`
* [ ] accept `data.frame` inputs for `data` argument to `lightgbm()`
* [ ] accept `data.frame` inputs for `data` argument to `lgb.cv()`
* [ ] automatically encode categorical-looking column types (factor, character) in a LightGBM-friendly way
* [ ] accept `data.frame` inputs for `predict.lgb.Booster()` / `Booster$predict()` / `Predictor$predict()`
    - once "automatically encode categorical-looking column types" is implemented, this should also including converting categorical-looking columns with the same rules
* [ ] allow passing a column name for `init_score` in `lgb.Dataset()` if `data` is a `data.frame`
* [ ] allow passing a column name for `label` in `lgb.Dataset()` if `data` is a `data.frame`
* [ ] allow passing a column name for `weight` in `lgb.Dataset()` if `data` is a `data.frame`

## References

See #4207 for a prior proposal and https://github.com/microsoft/LightGBM/pull/4207#pullrequestreview-659266142 for reference on this issue.

This feature precedes #4295.

jameslamb:
I've added this feature request to #2302, per this project's standard approach for managing its backlog.

If you are interested in contributing to this feature or have additional information to add, please leave a comment and the issue can be re-opened.

jameslamb:
Want to add that when this feature is picked up, the R package should support categorical features for data frames using the same interface as the Python package:

* setting `categorical_feature = "auto"` in `Dataset` (the default) means "automatically detect categorical features"
* if `categorical_feature` in `Dataset` is a vector of feature indices or names, use that list of features for `categorical_feature` instead

https://github.com/microsoft/LightGBM/blob/8a90ea3f267a81a529e3f069cc13e0f6320e7989/python-package/lightgbm/basic.py#L1127

[...]

PR #5424, opened 2022-08-16
[python-package][R-package] load parameters from model file (fixes #2613)

### LINKED ISSUE #2613: Load back saved parameters with save_model to Booster object

jmoralez:
Given that #4802 seems to have stalled, this intends to supersede it, mainly to make #4323 easier by having access to the features that would have to be turned into factors and also unblock #5246.

Fixes #2613.


jmoralez:
@StrikerRUS @jameslamb I'd appreciate your take on this approach. If you agree I can make the required changes to the R-package as well. The main idea is:
1. Parse the contents of `loaded_parameter_` into a JSON string.
2. Parse the parameter types from the config.
3. Load the parameters as strings in the wrappers and parse them to their corresponding types.
4. Assign the result as the parameters of the created Booster when calling the load_from_string/load_from_file methods.

python-package/lightgbm/basic.py
```diff
@@ -156,6 +156,28 @@ def _safe_call(ret: int) -> None:
         raise LightGBMError(_LIB.LGBM_GetLastError().decode('utf-8'))
 
 
+def _get_string_from_c_api(func: Callable, booster_handle: Optional[ctypes.c_void_p] = None) -> str:
```
StrikerRUS, line 159: Could you please make a separate PR with this code deduplication change only?
We have other places with string buffer creation, e.g. in
https://github.com/microsoft/LightGBM/blob/e063dad261d06049a461c3e2a364350e2b20b9f7/python-package/lightgbm/basic.py#L2315
https://github.com/microsoft/LightGBM/blob/e063dad261d06049a461c3e2a364350e2b20b9f7/python-package/lightgbm/basic.py#L3510
https://github.com/microsoft/LightGBM/blob/e063dad261d06049a461c3e2a364350e2b20b9f7/python-package/lightgbm/basic.py#L3568
https://github.com/microsoft/LightGBM/blob/e063dad261d06049a461c3e2a364350e2b20b9f7/python-package/lightgbm/basic.py#L3865
https://github.com/microsoft/LightGBM/blob/e063dad261d06049a461c3e2a364350e2b20b9f7/python-package/lightgbm/basic.py#L4085

StrikerRUS:
I'm afraid some floating point values will not survive float->string->float round trip...
Do we need something like the following?
https://github.com/dmlc/xgboost/pull/5772

python-package/lightgbm/basic.py
jmoralez: Removed in 0e6591b. d4e781b also adds a lot of duplicated for string buffers in `lightgbm_R.cpp`, could address that in a follow up PR as well.

jmoralez:
> I'm afraid some floating point values will not survive float->string->float round trip...

Yes, this will probably loose precision on some cases. We can address loading the parameters here and maybe have a feature request for making sure we get matches that are as close as possible for floating point values.

StrikerRUS:
Well, I don't think there can be a better solution other than a proposed one.

jmoralez:
I don't think precision loss in the parameters could be that big of a deal. There's already some on the thresholds and leaf values when writing the model file, so the parameters should be ok.

R-package/tests/testthat/test_lgb.Booster.R
```diff
@@ -172,15 +172,22 @@ test_that("Loading a Booster from a text file works", {
...
+        , boosting = "rf"
+        , bagging_fraction = 0.8
+        , bagging_freq = 1L
+        , force_col_wise = TRUE
+        , categorical_feature = c(1L, 2L)
+        , interaction_constraints = list(c(1L, 2L), 1L)
+        , learning_rate = 1.0
+        , objective = "binary"
+        , verbosity = VERBOSITY
+    )
```
jameslamb, line 186: I haven't reviewed this PR too thoroughly yet, but want to make one comment right now about testing as I think it'll help you.

Similar to https://github.com/microsoft/LightGBM/pull/4802#discussion_r786419637.

I believe that both the R package and Python packages should have tests that check at least one parameter per type, to catch type-specific issues.

In `config_auto.cpp`, I see the following types

* `bool`
* `double`
* `int`
* `string`
* `vector<double>`
* `vector<int>`
* `vector<string>`

I think this PR is missing tests for these types:

* `vector<double>`
* `vector<string>`

Can you please modify the R and Python tests to check parameters with those types? Since they are sometimes very specialized (for example, the only `vector<double>` is `cegb_*` parameters), I think it would make sense to add *new* test cases narrowly checking those types.

jameslamb (changes requested):
Left one quick comment about testing which might help in your work on this. I'll try to review more thoroughly later this week.

jmoralez:
@shiyu1994 does cuda_exp set `force_col_wise` to false?

StrikerRUS:
@jmoralez 
> does cuda_exp set `force_col_wise` to false?

Yes. Some GPU parameters are hardcoded:
https://github.com/microsoft/LightGBM/blob/83627ff0d6baf49e3b18991f47b12d9fa5724cbc/src/io/config.cpp#L336-L355

shiyu1994:
@jmoralez 
Yes, cuda_exp forces `force_col_wise=false` and `force_row_wise=true`.

R-package/R/lgb.Booster.R
```diff
@@ -727,6 +728,20 @@ Booster <- R6::R6Class(
...
 
+    get_loaded_param = function(handle) {
+      params_str <- .Call(
+        LGBM_BoosterGetLoadedParam_R
+        , handle
+      )
+      params <- jsonlite::fromJSON(params_str)
+      if ("interaction_constraints" %in% names(params)) {
+        params[["interaction_constraints"]] <- lapply(params[["interaction_constraints"]], function(x) x + 1L)
+      }
```
jmoralez, line 739: Given that this was only necessary for interaction constraints I'm afraid categorical feature's indices aren't being adjusted. Can that be the case?

jmoralez: It happens on dataset:
https://github.com/microsoft/LightGBM/blob/2b8fe8b4bdc00a2611442fdee4c45316f08b1c4b/R-package/R/lgb.Dataset.R#L159
https://github.com/microsoft/LightGBM/blob/2b8fe8b4bdc00a2611442fdee4c45316f08b1c4b/R-package/R/lgb.Dataset.R#L184
We may want to work on this later because in this case the saved model's categorical_feature is wrong.

jameslamb: As I mentioned in https://github.com/microsoft/LightGBM/pull/4970#pullrequestreview-937407074

1. there are many places where LightGBM's C++ code expects 0-based indexes
2. because R is 1-based, there have been some places where the R package has translated from 1-based inputs to 0-based values
3. this has been done very inconsistently
4. this requires careful thought and a comprehensive plan

For example.... should 0-based indices be converted back to 1-based in `Booster$save_model()`? Or is it expected that the model file produced by R code could be directly read by Python code?

What about in `params`? Is it expected that you could save a dictionary from Python code to a file like `params.json`, then read it in R and immediately pass it to `lgb.train(..., params)`?

It's not obvious to me which choice in these and other situations is "right" or "wrong", and I think that deserves a comprehensive plan with guidelines for when users should expect to see 0-based vs. 1-based values. I'd like to do that work at some point, but it is really difficult and a large amount of effort right.

For the purpose of this PR, I recommend just focusing on preserving the current behavior.

jmoralez: I was thinking that maybe the CLI would get the wrong predictions but it doesn't seem like it. IMO I would expect the saved model to be the same regardless of whether it was trained on R or Python, so converting to 0-based indices when saving makes sense to me.

> For the purpose of this PR, I recommend just focusing on preserving the current behavior.

Yes, I don't intend to make changes on this PR, just wanted to highlight that I found it odd.

tests/python_package_test/test_engine.py
```diff
@@ -1203,6 +1203,30 @@ def test_feature_name_with_non_ascii():
...
+        'metric': ['l2', 'rmse'],
+        'num_leaves': 5,
+        'num_threads': 1,
+    }
+    model_file = tmp_path / 'model.txt'
+    lgb.train(params, ds, num_boost_round=1, categorical_feature=[1, 2]).save_model(model_file)
+    bst = lgb.Booster(model_file=model_file)
+    set_params = {k: bst.params[k] for k in params.keys()}
+    assert set_params == params
+    assert bst.params['categorical_feature'] == [1, 2]
```
jameslamb, line 1227: These tests are checking that the `params` attribute on the Python `Booster` object matches what is in the file...but that being true doesn't necessarily guarantee that if you used this `Booster` for continued training that it would still actually use those parameters at the C++ side, right?

I'm bringing this up because I'm struggling to understand the relationship between `GBDT::loaded_parameter_`

https://github.com/microsoft/LightGBM/blob/9e89ee7fac7dfe18f0f4ab3ea7542854e3d0b936/src/boosting/gbdt_model_text.cpp#L600

and `GBDT::config_`

https://github.com/microsoft/LightGBM/blob/9e89ee7fac7dfe18f0f4ab3ea7542854e3d0b936/src/boosting/gbdt.h#L461-L462

In https://github.com/microsoft/LightGBM/issues/2613#issuecomment-955948360, I'd recommended using `GBDT::config_` as the source of the parameter information, because it seems to me like the parameters section of the model file is loaded up into that property `loaded_parameter_`, but not actually used to configure the `GBDT` object. But now I'm not sure that that's right either haha.

For example....if you tried continued training after loading a model from text file like this, would LightGBM respect `feature_fraction=0.7`? Or would it fall back to the default of `1.0`?

jameslamb (changes requested):
I was able to review this thoroughly today.

I love that you were able to accomplish this with only *private* changes to the R and Python package, and one new public entrypoint in `c_api`. Nice work! That makes me confident that we could change this implementation in the future without breaking users.

I don't have any specific suggestions at the moment (tests look awesome!), but I asked two questions that I feel I need a better understanding of before I can approve this.

tests/python_package_test/test_engine.py
jmoralez: > that being true doesn't necessarily guarantee that if you used this Booster for continued training that it would still actually use those parameters at the C++ side, right?

I believe the `loaded_parameter_` attribute was added in #1495 to have the model parameters stored, but it isn't actually used anywhere other than when writing the parameters back to a file when there's no config. https://github.com/microsoft/LightGBM/blob/d78b6bc2fdf96c87e3cb61f2d497a962e3270c91/src/boosting/gbdt_model_text.cpp#L393-L401 Not sure when this happens though, maybe from the CLI using refit task.

So I see these parameters more as "informative" of how the Booster was trained, rather than to be used to continue training. The original issue wanted to have `objective` be loaded back to be used in SHAP and I think having them may help trying to replicate previous trainings or having the `categorical_feature` for inference. I actually thought of maybe having an argument to decide whether or not to load them, they may not be useful and that computation could be skipped.

python-package/lightgbm/basic.py
```diff
@@ -2764,6 +2764,7 @@ def __init__(
                 ctypes.byref(out_num_class)))
             self.__num_class = out_num_class.value
             self.pandas_categorical = _load_pandas_categorical(file_name=model_file)
+            params = self._get_loaded_param()
```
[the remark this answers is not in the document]
jmoralez: I prefer ignoring them, since they won't be used by the loaded booster and it could cause confusion, we can raise a warning when passing both params and model_file to the Booster constructor make this more evident. When using the loaded booster as init model users can override the parameters for the new iterations in the train function, so that should be enough.

I haven't done much incremental learning so I hadn't given much thought to this haha, please feel free to correct me.

tests/python_package_test/test_engine.py
StrikerRUS: @guolinke Could you please help to understand the connection between `config_` and  `loaded_parameter_`?

python-package/lightgbm/basic.py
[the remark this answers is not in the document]
StrikerRUS: > We can raise a warning when passing both params and model_file to the Booster constructor to make this more evident.

I strongly believe we need a warning for users here.

And I think it's better to override loaded params from model string by the params passed in via `kwargs`.

tests/python_package_test/test_engine.py
guolinke: IMO, I think the "loaded_parameter_"  is used to check the parameters used for the model file. For continued training, the current "config_" is used.

jameslamb: > to check the parameters used for the model file

ok but @guolinke does `loaded_parameter_` actually get used to update `config_`?

If it does, then wrappers like the R and Python package shouldn't ever need to know about `loaded_parameter_`, and could just rely on `config_`.

If if doesn't, then I think it means that if you load a model from a text file and then perform continued training, you'll need to explicitly `.update_parameters()` (at least in the Python and R package), or otherwise the additional training rounds will just use all of the default values of LightGBM's parameters.

In my opinion, ensuring that a model loaded from a text file looks as similar as possible to one created in-memory by training (e.g. that both `Booster.params` in the Python object and the `config_` property on the C++ booster are identical to what they'd be after running `lgb.train()`) would simplify this feature and other features related to using models that were loaded from text files.

guolinke: @jameslamb I think the `config_` is used to update  `loaded_parameter_`.
User can still set new parameters when "loaded_parameter_", when continue training from a model. In this case, I think the new parameters have higher priority. 
So it is like, model_file -> loaed_parameters_ -> load_config -> update by user's new parameters.
The problem here is the default parameter values.

python-package/lightgbm/basic.py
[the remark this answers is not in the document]
jmoralez: Sure I can add the warning. What would be an example use case for overriding the parameters from the model file when loading the booster?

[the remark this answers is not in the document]
jameslamb: Anything around continued training might result in loading a model from a file and then performing additional training with different parameters.

For examples:

* my example in this thread, about wanting to change `machines` (since the IP addresses of machines used for distributed training might be different from when the model file was created)
* wanting to use a different `num_leaves` or `learning_rate` when performing an update like "train a few more iterations on newly-arrived data" (see my explanation in [this Stack Overflow answer](https://stackoverflow.com/questions/73664093/lightgbm-train-vs-update-vs-refit/73669068#73669068))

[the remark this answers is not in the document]
jmoralez: But I'm struggling to understand when `Booster.update` would be called on a model loaded from file. It requires a bit of setup like defining the training set, and if you pass the training set to the constructor it takes a different path where it calls BoosterCreate with only the training set and the parameters.

The other case would be passing it to lgb.train as init_model and that already allows you to define the parameters for the new iterations.

tests/python_package_test/test_engine.py
guolinke: @jmoralez @jameslamb any further comments about this?

jmoralez: I added a warning about passing parameters to the booster constructor in 9467814 because in order to be able to call `Booster.update` you need a training set, and if you initialize a booster with the training set `lgb.Booster(model_file=..., train_set=...)` the `model_file` argument gets ignored, so I don't think there's a way to do incremental training based on the `Booster` object. Thus, I don't think it really matters what the parameters for the loaded booster are on the cpp side, because the incremental training would be through `lgb.train` that already supports overriding the parameters for the new iterations. WDYT?

guolinke: it sounds good to me

jameslamb: Thanks @jmoralez 

It's an excellent point that `Booster`'s constructor has logic like

```python
if train_set is not None:
    ...
elif model_file is not None:
    ...
```

Meaning that you can't have both. I agree with the warning you've added and appreciate you adding a test for it as well!

I'm sorry for holding up this PR for so long over this. I'm still confused about the relationship between `config_` and `loaded_parameter_`, but I don't have any more concerns about this specific PR.

jameslamb (approved):
Thanks for the awesome work, and sorry I held this up for so long. The logic for how parameters flow through LightGBM is fairly complicated, and I wanted to be sure I understood the full implications of this change.

I think we should merge this 😀

tests/python_package_test/test_engine.py
jameslamb: I should also add, as I mentioned in https://github.com/microsoft/LightGBM/pull/5424#pullrequestreview-1092403723 ... I think we should merge this because future changes probably wouldn't require any user-facing breaking changes. The fact that the new function you've added in `c_api` takes in a `BoosterHandle` means that we could swap it to using a different property of the `Booster` (e.g. `config_` instead of `loaded_parameter_`) in the future without an API-level breaking change. Very nice!

Issue #5163, opened 2022-04-21
[CUDA] Objective functions and metrics for cuda_exp


shiyu1994:
## Summary

The `cuda_exp` in current master branch includes only a new tree learner version for CUDA. Objectives (boosting) and metrics (evaluation) should be implemented on CUDA.

jameslamb:
Thanks for writing this up @shiyu1994 !

Can you add some more details here, to help others from the community who might want to help contribute to this?

For example...would it be possible to implement metrics one at a time (e.g. a PR like "add CUDA implementation of MAPE") and have LightGBM choose whether to evaluate it on the GPU or CPU? Or would this require a single PR implementing all metrics?

shiyu1994:
> For example...would it be possible to implement metrics one at a time

Sure. That's what I have planned. Each PR should contain only one or several related objectives/metrics, to keep the PRs as small as possible.

jameslamb:
Hey @shiyu1994 , are there and objective functions or metrics which still need to be added? If so, could you list them here?

If not, could we close this?

shiyu1994:
Yes, there are still a few, which are marked in these two files:
https://github.com/microsoft/LightGBM/blob/16ca335b4a0d7a83be2b0e05133d0919b96bf35e/src/objective/objective_function.cpp#L46-L64
https://github.com/microsoft/LightGBM/blob/16ca335b4a0d7a83be2b0e05133d0919b96bf35e/src/metric/metric.cpp#L38-L70

jameslamb:
Ahhh right, ok I was searching in the wrong place. Thank you.

PR #5425, opened 2022-08-16
[CUDA] Add binary objective for cuda_exp

shiyu1994:
Add binary objective for `cuda_exp`. (#5163)


guolinke:
it seems the cuda self-hosted machine in queue? BTW, do we have a test for the cuda binary objective?

shiyu1994:
> BTW, do we have a test for the cuda binary objective?

I think this can be covered by existing tests for binary objective on CPUs, for example,
https://github.com/microsoft/LightGBM/blob/9489f878b3568e70b441e5df602483e116f24cc6/tests/python_package_test/test_engine.py#L61-L83

because we are replacing the default device with `cuda_exp` in
https://github.com/microsoft/LightGBM/blob/9489f878b3568e70b441e5df602483e116f24cc6/.ci/test.sh#L204
when testing for `cuda_exp` version.

shiyu1994:
It is weird that this PR should now passes all the tests. But my local tests are failing with `cuda_exp` enabled. I'm taking a deeper check into this.

src/objective/cuda/cuda_binary_objective.cu
```diff
@@ -0,0 +1,227 @@
...
+  const double sigmoid, const data_size_t num_data,
+  score_t* cuda_out_gradients, score_t* cuda_out_hessians) {
+  const data_size_t data_index = static_cast<data_size_t>(blockDim.x * blockIdx.x + threadIdx.x);
+  if (data_index < num_data) {
+    const label_t cuda_label = static_cast<int>(cuda_labels[data_index]);
+    const int label = IS_OVA ? (cuda_label == ova_class_id ? 1 : -1) : (cuda_label > 0 ? 1 : -1);
+    const double response = -label * sigmoid / (1.0f + exp(label * sigmoid * cuda_scores[data_index]));
+    const double abs_response = fabs(response);
+    if (!USE_WEIGHT) {
+      if (USE_LABEL_WEIGHT) {
```
guolinke, line 129: should we use `constexpr` for these template conditions ?

shiyu1994: constexpr if is only available with c++17 or higher version.

PR #5299, opened 2022-06-16
feature: Add true streaming APIs to reduce client-side memory usage

svotaw:
### Background
Currently, Azure SynapseML wraps the LightGBM library to make it much easier for customers to use it in a distributed Spark environment. The initial approach taken by SynapseML uses the `LGBM_DatasetCreateFromMat `API to create a `Dataset `for either each Spark partition or each executor (user option). However, this method requires the SynapseML Scala code to load the entire uncompressed data into memory before sending it to LightGBM, and merge partition data together manually. This amounts to using an order of magnitude or more memory (raw double arrays and multiple copies) over what LightGBM `Datasets `use internally (binned data). This requires larger Spark clusters than are really needed, and often causes OOM issues.

In order to improve the memory performance of SynapseML-wrapped LightGBM, we decided to convert the `Dataset `creation into more of a “_streaming_” scenario (as opposed to the above “_bulk_” input matrices), where we do NOT create large arrays on client side. After initial investigation, there were existing LightGBM APIs that seemed to fit this purpose: `LGBM_DatasetCreateFromSampledColumn` and `LGBM_DatasetPushRows[byCSR]`.  This seemed to allow creation of a “reference” `Dataset `with defined feature groups and bins, and to push small micro-batches of data into the set. However, these APIs suffered from several significant issues as currently implemented:

1.	Not thread safe for parallelism, since `FinishDataset `is called when the literal last index is pushed.
2.	Only pushes feature data, and `Metadata `must still be handled client side

### Changes
This PR adds APIs to LightGBM to fix these issues and create a true “streaming” flow where microbatches of full rows, including metadata, can be pushed directly into LightGBM `Dataset `format. Control over FinishLoad() is given to client to eliminate problem with push order.

The new general streaming flow uses the following APIs:
1. `DatasetCreateFromSampledColumn `(final number of rows) with samples from partition
2.`DatasetInitStreaming`, to initialize allocation and other setup
3.`DatasetPushRows[byCSR]WithMetadata` over the Spark Row `Iterator`
4.`DatasetMarkFinished`

Note that there shouldn’t be any impact of this PR on training iteration code.  The changes are mostly limited to initial `Dataset `creation and loading.

### Testing
I added C++ tests for all of the above so we could test the functionality intensively. There are now C++ test files streaming.  Both dense and sparse scenarios are covered, microbatch sizes of 1 and 2+, and all types of Metadata.
[...]

PR #5437, opened 2022-08-23
Add streaming concurrency tests

svotaw:
This PR adds concurrency testing to new streaming APIs added in PR https://github.com/microsoft/LightGBM/pull/5299.

It also attempts to fix a concurrency issue in `SparseBin `PushRow API. WIthout this fix, concurrent calls often fail to the `LGBM_DatasetPushRowsByCSRWithMetadata `API.

### DIFF
--- a/src/io/sparse_bin.hpp
+++ b/src/io/sparse_bin.hpp
@@ -92,9 +92,6 @@ class SparseBin : public Bin {
   void Push(int tid, data_size_t idx, uint32_t value) override {
     auto cur_bin = static_cast<VAL_T>(value);
     if (cur_bin != 0) {
-      if (push_buffers_.size() <= tid) {
-        push_buffers_.resize(tid + 1);
-      }
       // Splitting this operation into separate calls avoids a concurrency exception. Do not simplify.
       auto buffer = push_buffers_[tid];
       buffer.emplace_back(idx, cur_bin);
--- a/src/io/sparse_bin.hpp
+++ b/src/io/sparse_bin.hpp
@@ -91,11 +91,12 @@ class SparseBin : public Bin {
 
   void Push(int tid, data_size_t idx, uint32_t value) override {
     auto cur_bin = static_cast<VAL_T>(value);
-    if (push_buffers_.size() <= tid) {
-      push_buffers_.resize(tid + 1);
-    }
     if (cur_bin != 0) {
-      push_buffers_[tid].emplace_back(idx, cur_bin);
+      if (push_buffers_.size() <= tid) {
+        push_buffers_.resize(tid + 1);
+      }
+      auto buffer = push_buffers_[tid];
+      buffer.emplace_back(idx, cur_bin);
     }
   }
 
--- a/src/io/sparse_bin.hpp
+++ b/src/io/sparse_bin.hpp
@@ -95,6 +95,7 @@ class SparseBin : public Bin {
       if (push_buffers_.size() <= tid) {
         push_buffers_.resize(tid + 1);
       }
+      // Splitting this operation into separate calls avoids a concurrency exception. Do not simplify.
       auto buffer = push_buffers_[tid];
       buffer.emplace_back(idx, cur_bin);
     }
--- a/src/io/sparse_bin.hpp
+++ b/src/io/sparse_bin.hpp
@@ -91,6 +91,9 @@ class SparseBin : public Bin {
 
   void Push(int tid, data_size_t idx, uint32_t value) override {
     auto cur_bin = static_cast<VAL_T>(value);
+    if (push_buffers_.size() <= tid) {
+      push_buffers_.resize(tid + 1);
+    }
     if (cur_bin != 0) {
       push_buffers_[tid].emplace_back(idx, cur_bin);
     }
--- a/src/io/sparse_bin.hpp
+++ b/src/io/sparse_bin.hpp
@@ -92,9 +92,7 @@ class SparseBin : public Bin {
   void Push(int tid, data_size_t idx, uint32_t value) override {
     auto cur_bin = static_cast<VAL_T>(value);
     if (cur_bin != 0) {
-      // Splitting this operation into separate calls avoids a concurrency exception. Do not simplify.
-      auto buffer = push_buffers_[tid];
-      buffer.emplace_back(idx, cur_bin);
+      push_buffers_[tid].emplace_back(idx, cur_bin);
     }
   }
 
--- a/tests/cpp_tests/test_stream.cpp
+++ b/tests/cpp_tests/test_stream.cpp
@@ -198,6 +198,7 @@ void test_stream_sparse(
 
     dataset = static_cast<Dataset*>(dataset_handle);
 
+    Log::Info("Streaming sparse dataset, %d rows sparse data with a batch size of %d", nrows, batch_count);
     TestUtils::StreamSparseDataset(
       dataset_handle,
       nrows,
@@ -213,7 +214,6 @@ void test_stream_sparse(
 
     dataset->FinishLoad();
 
-    Log::Info("Streaming sparse dataset, %d rows sparse data with a batch size of %d", nrows, batch_count);
     TestUtils::AssertMetadata(&dataset->metadata(),
                               labels,
                               weights,
--- a/tests/cpp_tests/test_stream.cpp
+++ b/tests/cpp_tests/test_stream.cpp
@@ -185,7 +185,7 @@ void test_stream_sparse(
         EXPECT_EQ(0, result) << "LGBM_DatasetCreateFromSampledColumn result code: " << result;
 
         dataset = static_cast<Dataset*>(dataset_handle);
-        dataset->InitStreaming(nrows, has_weights, has_init_scores, has_queries, nclasses, 1);
+        dataset->InitStreaming(nrows, has_weights, has_init_scores, has_queries, nclasses, 2);
         break;
       }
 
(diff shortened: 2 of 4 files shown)


svotaw:
This PR is 99% new CPP tests, which pass on the pipeline.  I'm not sure why there are so many other pipeline failures (first commit showed no such failures).  The 1-line change in production code seems to be innocuous for python tests, but feel free to tell me otherwise.

Although this helps with concurrency failures locally, we'll need to test this change further once checked in.  We have a concurrency failure that we only see in our CI Linux tests.

jameslamb (changes requested):
Thanks for your continued work on LightGBM, @svotaw !

> we'll need to test this change further once checked in. We have a concurrency failure that we only see in our CI Linux tests

I'm not sure who "we" / "our" refers to in this statement or what you mean by "once checked in", so I want to be sure to set the right expectation...we will not merge pull requests that do not pass all CI jobs into `master` in this repository.

This PR will not be merged until all of the existing CI jobs are passing.

> The 1-line change ... seems to be innocuous for python tests, but feel free to tell me otherwise

When working on LightGBM, it's important to understand that the Python package's tests are the most comprehensive tests this project has for its C/C++ code. Those tests cover more of the project's C/C++ code than the dedicated tests in https://github.com/microsoft/LightGBM/tree/master/tests/cpp_tests.

The R package tests also cover a significant portion of the C/C++ code, and as of #5312 at least one code path not covered by the Python tests (`LGBM_BoosterPredictForMatSingleRowFast()` and related code).

So when you push a change to C/C++ code in this project and see most of the R and Python tests failing, that is a strong indication that the change you've pushed is a breaking change to LightGBM.

Maybe you've uncovered a correctness bug while trying to fix a source of runtime exceptions (which would be great!) or maybe this change has a bigger impact on model training than just fixing a runtime exception related to concurrency issues. I'm not experienced enough with C++ or the implementation of `SparseBin` to say for sure. @guolinke or @shiyu1994 should be able to help.

Please let me know if you need help with how to run only the failing Python test cases locally (so you can work on this with a faster feedback cycle than waiting for all of this project's CI jobs to run).

src/io/sparse_bin.hpp
```diff
@@ -92,7 +92,9 @@ class SparseBin : public Bin {
   void Push(int tid, data_size_t idx, uint32_t value) override {
     auto cur_bin = static_cast<VAL_T>(value);
     if (cur_bin != 0) {
-      push_buffers_[tid].emplace_back(idx, cur_bin);
+      // Splitting this operation into separate calls avoids a concurrency exception. Do not simplify.
+      auto buffer = push_buffers_[tid];
```
guolinke, line 96: this will be a clone, rather than a reference?

svotaw:
> I'm not sure who "we" / "our" refers to in this statement or what you mean by "once checked in", so I want to be sure to set the right expectation...we will not merge pull requests that do not pass all CI jobs into `master` in this repository.

@jameslamb  By "we", I mean our SynapseML wrapper.  And yes, of course I would never expect anything to be checked in to LightGBM without passing all tests.  :)  I just thought I'd check and see if there were any current CI problems that might have caused those other things to fail.  They were passing on an earlier iteration of this that included the only production code change that is left.  But if there are no other issues, I will investigate further.  Probably this change then.

src/io/sparse_bin.hpp
svotaw: it should be the same either way (reference), but for some reason splitting it into 2 calls stopped the exceptions.  I will keep investigating to see if I can determine why.

The main point of this PR was to add in the multi-threading tests, since in the previous PR I promised I would add them in a subsequent PR.

guolinke: I believe most "=" operator in C++ is "copy assignment", refer to https://www.enseignement.polytechnique.fr/informatique/INF478/docs/Cpp/en/cpp/container/vector/operator%3D.html 
I also write a small test.
```
#include <iostream>
#include <vector>
int main() {
    std::vector<int> a = {1, 2, 3};
    auto b = a;
    printf("%d, %d\n", &a, &b);
    return 0;
}
```
And it shows the addresses of `a` and `b` is different. 

And you should use `auto& b = a` to get the reference.

guolinke:
@svotaw I am quite confused about this PR, what does "concurrency" mean?
The PushRow is already designed for multi-threading, and LightGBM also uses a multi-threading solution to construct Dataset.

svotaw:
> @svotaw I am quite confused about this PR, what does "concurrency" mean? The PushRow is already designed for multi-threading, and LightGBM also uses a multi-threading solution to construct Dataset.

You can also say multi-threading instead of concurrency.  I mean the same thing.

The multi-threaded tests in this PR were just added since someone asked for them in the previous PR. I said that I would add them, and this is the followup.

And by the way, the external LGBM_DatasetPushRowsByCSR is NOT thread safe, which is why I modified LGBM_DatasetPushRowsByCSRWithMetadata to be able to handle being called by multiple threads.  The old LGBM_DatasetPushRowsByCSR API will crash if you call it from multiple threads.  We found that out empirically.  Each external thread will generate the same set of OpenMP tid's, which means they will attempt to push to the same sparse buffers and eventually fail. LGBM_DatasetPushRowsByCSRWithMetadata is supposed to be completely thread safe (although see next).

As far as changing that one production code line in SparseBin.hpp, that is an experiment.  Even with the thread-safe design, we still see some crash failures in SparseBin.PushRow() in our own tests using the top of LightGBM main (we made a maven package of it for SynapseML).  They have been hard to debug, because they aren't reproducible locally and have only occurred in our CI pipeline runs (making it hard to get crash dumps).  We only know the method that fails.  I was using this PR to help find the issue, and making that change seemed to make the tests pass reliably. But admittedly, I'm not convinced yet that it is needed, so will experiment some more before asking to check this in.

The changes to tests (99% of this PR) should have no production impact, as long as they pass reliably and don't cause CI issues.  They are just to have the extra multi-threaded coverage.  Note that these tests only work for LGBM_DatasetPushRowsByCSRWithMetadata.  LGBM_DatasetPushRowsByCSR would fail if tested the same way.

Note that I manually made threads in the test since OpenMP does not seem to work for tests.  I tried, but settled on manual thread creation.

svotaw:
@guolinke here is the error we see with multi-threaded sparse data in our SynapseML tests:
# JRE version: OpenJDK Runtime Environment (8.0_345-b01) (build 1.8.0_345-b01)
# Java VM: OpenJDK 64-Bit Server VM (25.345-b01 mixed mode linux-amd64 compressed oops)
# Problematic frame:
# C  [lib_lightgbm.so+0x18a72b]  LightGBM::SparseBin<unsigned char>::Push(int, int, unsigned int)+0x3b

It might be something other than a threading issue but given the debugging I've done it seems highly likely.

jameslamb:
@shiyu1994 or @guolinke can you please re-review this, now that it's only testing changes and CI is passing?

------

> passing on an earlier iteration of this that included the only code change that is left

I don't think that's true.

The first commit in this PR, https://github.com/microsoft/LightGBM/pull/5437/commits/62a97c20715c3154dd2e544ccd94617b5f329898, which was passing all CI jobs except `lint`, contained only this change + testing changes:

image

Then https://github.com/microsoft/LightGBM/pull/5437/commits/cc3f4316c128a1a8f2a8586d4efbb1138673a4c2 moved that code to a different location **and introduced another change**.

image

From that commit until you removed the `SparseBin` changes in https://github.com/microsoft/LightGBM/pull/5437/commits/e796f5ec6948d3839afe9a7b6833b5e3ebb18964, most of the Python and R jobs were failing.

Just pointing it out in case it helps you in your investigation.

guolinke (approved):

src/io/sparse_bin.hpp
svotaw: yeah, I removed this since it was wrong and as you said, didn't do what I thought it was doing.

jameslamb (approved):
Thanks for the changes!

In the future, please don't push PRs with whitespace-only changes like this

image

unless those changes are directly related to the goal of the PR.

We rely on the git blame view as a source of history for changes to files, and whitespace-only changes that are not related to functionality add friction to the use of that view.

I'm approving this one since these tests are new and under such active development, and to avoid delaying you further.

svotaw:
@jameslamb  Ah, my apologies.  The teams I've been on have always had the policy to do simple style fixes whenever they saw them, but I will avoid doing that here from now on.
Do you just make separate PRs for this kind of change?

jameslamb:
> Do you just make separate PRs for this kind of change?

thanks very much! Yes, separate PRs are preferred for unrelated whitespace changes, typo/grammar fixes, etc. I personally have found that having all the changes in a PR's diff relate only to the stated purpose of the PR (in this example, "add more tests") makes PRs much easier to review.

And I personally use this project's git blame regularly to trace back to the relevant PRs that changed a specific section of the code.

tests/cpp_tests/test_stream.cpp
```diff
@@ -320,7 +320,7 @@ TEST(Stream, PushSparseRowsWithMetadata) {
   TestUtils::CreateRandomSparseData(nrows, ncols, nclasses, sparse_percent, &indptr, &indices, &vals, &labels, &weights, &init_scores, &groups);
 
   const std::vector<int32_t> batch_counts = { 1, nrows / 100, nrows / 10, nrows };
-  const std::vector<int8_t> creation_types = { 0, 1 };
+  const std::vector<int8_t> creation_types = { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 };
```
StrikerRUS, line 323: @svotaw 
Why there are only zeros here now and no any ones?
Previously `creation_types = { 0, 1 };` made sense to me, but now I'm very confused by such test parametrization.

svotaw: @StrikerRUS  oops, yes need to revert this.  It was a quick and dirty experiment just to force many repeats.  Some of the early testing failures were only sporadic, so I added this to give me better confidence on no failures.  I will fix.

I had planned to do a last pass over the PR to look for things like this, but it got checked in.  Not used to someone else being in control of the checkin. :)

StrikerRUS: No problem, thanks a lot for your awesome work on LightGBM!

Luckily, we need to revert just one line of the code. 🙂

PR #5413, opened 2022-08-10
Add default definition for GetColWiseData and GetColWiseData

shiyu1994:
This is to fix the compilation warnings mentioned in #5176.

PR #5403, opened 2022-08-04
[fix] change the destructor of ScoreUpdater to virtual (fixes #5400)

shiyu1994:
This is to fix #5400.

PR #5438, opened 2022-08-24
[ci][fix] Fix cuda_exp ci

shiyu1994:
Unfortunately, in a very long time (since the merge of #4630), the ci tests for `cuda_exp` has not actually run.
Since the task name for both `cuda` and `cuda_exp` is `cuda`
https://github.com/microsoft/LightGBM/blob/702db13b01bfb090c3f8f3455790a498f105b28f/.github/workflows/cuda.yml#L14
in `test.sh` the `cuda_exp` option cannot be detected, and it will always run for device `cuda`!
https://github.com/microsoft/LightGBM/blob/702db13b01bfb090c3f8f3455790a498f105b28f/.ci/test.sh#L198-L199

That's a serious mistake. And I think we need to fix it right now for our on-going development of `cuda_exp`.

### DIFF
--- a/.ci/test.sh
+++ b/.ci/test.sh
@@ -196,10 +196,10 @@ if [[ $TASK == "gpu" ]]; then
         cmake -DUSE_GPU=ON ..
     fi
 elif [[ $TASK == "cuda" ]]; then
-    if [[ $DEVICE == "cuda" ]]; then
+    if [[ $TASK == "cuda" ]]; then
         sed -i'.bak' 's/std::string device_type = "cpu";/std::string device_type = "cuda";/' $BUILD_DIRECTORY/include/LightGBM/config.h
         grep -q 'std::string device_type = "cuda"' $BUILD_DIRECTORY/include/LightGBM/config.h || exit -1  # make sure that changes were really done
-    elif [[ $DEVICE == "cuda_exp" ]]; then
+    elif [[ $TASK == "cuda_exp" ]]; then
         sed -i'.bak' 's/std::string device_type = "cpu";/std::string device_type = "cuda_exp";/' $BUILD_DIRECTORY/include/LightGBM/config.h
         grep -q 'std::string device_type = "cuda_exp"' $BUILD_DIRECTORY/include/LightGBM/config.h || exit -1  # make sure that changes were really done
         # by default ``gpu_use_dp=false`` for efficiency. change to ``true`` here for exact results in ci tests
@@ -210,9 +210,9 @@ elif [[ $TASK == "cuda" ]]; then
     fi
     if [[ $METHOD == "pip" ]]; then
         cd $BUILD_DIRECTORY/python-package && python setup.py sdist || exit -1
-        if [[ $DEVICE == "cuda" ]]; then
+        if [[ $TASK == "cuda" ]]; then
             pip install --user $BUILD_DIRECTORY/python-package/dist/lightgbm-$LGB_VER.tar.gz -v --install-option=--cuda || exit -1
-        elif [[ $DEVICE == "cuda_exp" ]]; then
+        elif [[ $TASK == "cuda_exp" ]]; then
             pip install --user $BUILD_DIRECTORY/python-package/dist/lightgbm-$LGB_VER.tar.gz -v --install-option=--cuda-exp || exit -1
         else
             exit -1
@@ -220,9 +220,9 @@ elif [[ $TASK == "cuda" ]]; then
         pytest $BUILD_DIRECTORY/tests/python_package_test || exit -1
         exit 0
     elif [[ $METHOD == "wheel" ]]; then
-        if [[ $DEVICE == "cuda" ]]; then
+        if [[ $TASK == "cuda" ]]; then
             cd $BUILD_DIRECTORY/python-package && python setup.py bdist_wheel --cuda || exit -1
-        elif [[ $DEVICE == "cuda_exp" ]]; then
+        elif [[ $TASK == "cuda_exp" ]]; then
             cd $BUILD_DIRECTORY/python-package && python setup.py bdist_wheel --cuda-exp || exit -1
         else
             exit -1
@@ -231,9 +231,9 @@ elif [[ $TASK == "cuda" ]]; then
         pytest $BUILD_DIRECTORY/tests || exit -1
         exit 0
     elif [[ $METHOD == "source" ]]; then
-        if [[ $DEVICE == "cuda" ]]; then
+        if [[ $TASK == "cuda" ]]; then
             cmake -DUSE_CUDA=ON ..
-        elif [[ $DEVICE == "cuda_exp" ]]; then
+        elif [[ $TASK == "cuda_exp" ]]; then
             cmake -DUSE_CUDA_EXP=ON ..
         else
             exit -1
--- a/.ci/test.sh
+++ b/.ci/test.sh
@@ -195,7 +195,7 @@ if [[ $TASK == "gpu" ]]; then
     elif [[ $METHOD == "source" ]]; then
         cmake -DUSE_GPU=ON ..
     fi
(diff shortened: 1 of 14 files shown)


shiyu1994:
See https://github.com/microsoft/LightGBM/pull/5425#issuecomment-1225352237

guolinke (approved):

.ci/test.sh
```diff
@@ -195,40 +195,48 @@ if [[ $TASK == "gpu" ]]; then
...
         grep -q 'std::string device_type = "cuda"' $BUILD_DIRECTORY/include/LightGBM/config.h || exit -1  # make sure that changes were really done
-    else
+    elif [[ $DEVICE == "cuda_exp" ]]; then
         sed -i'.bak' 's/std::string device_type = "cpu";/std::string device_type = "cuda_exp";/' $BUILD_DIRECTORY/include/LightGBM/config.h
         grep -q 'std::string device_type = "cuda_exp"' $BUILD_DIRECTORY/include/LightGBM/config.h || exit -1  # make sure that changes were really done
         # by default ``gpu_use_dp=false`` for efficiency. change to ``true`` here for exact results in ci tests
         sed -i'.bak' 's/gpu_use_dp = false;/gpu_use_dp = true;/' $BUILD_DIRECTORY/include/LightGBM/config.h
         grep -q 'gpu_use_dp = true' $BUILD_DIRECTORY/include/LightGBM/config.h || exit -1  # make sure that changes were really done
+    else
+        exit -1
```
jameslamb, line 209:
```suggestion
        echo "did not recognize DEVICE '${DEVICE}'. Should be either 'cuda' or 'cuda_exp'."
        exit -1
```

Can you please add a log line here explaining what happened? Otherwise, the job will just fail and it won't be clear why or that it happened because of this specific line.

jameslamb (changes requested):
Thank you very much for noticing and proposing a fix for this! But I think there's a much simpler fix that will achieve the same goal.

I think doing the following would prevent adding new complexity to `.ci/test.sh`:

1. replace `treelearner: ` in `.github/workflows/cuda.yml` with `task:`
2. remove this line: https://github.com/microsoft/LightGBM/blob/702db13b01bfb090c3f8f3455790a498f105b28f/.github/workflows/cuda.yml#L14
3. change the following line to `TASK="${{ matrix.task }}"`: https://github.com/microsoft/LightGBM/blob/702db13b01bfb090c3f8f3455790a498f105b28f/.github/workflows/cuda.yml#L89

I remember we had a discussion in #4630 about not adding more complexity to `test.sh`: https://github.com/microsoft/LightGBM/pull/4630#discussion_r820284099.

-----

In addition to that request...right now, both `cuda_exp` CI jobs are failing with the following error

> Fatal Python error: Aborted

([build link](https://github.com/microsoft/LightGBM/runs/7991788198?check_suite_focus=true))

We should never merge a PR that will result in CI being broken on `master`, so please do one of these:

* push a fix for that issue to this PR
* remove the `cuda_exp` CI jobs in this PR, document the work to add them back in a separate issue, merge this PR, and then push a follow-up PR fixing `cuda_exp` and adding back the CI jobs

I hope a fix can just be pushed here, but I don't know how complicated it will be to debug this. If it is very complicated, then we might as well eliminate those two `cuda_exp` jobs that are not actually testing `cuda_exp`, to save CI time.

And if the "make an issue and fix this later" approach is taken, then I think we should not merge *any* more `cuda_exp` PRs until those CI jobs have been re-enabled.

jameslamb:
I've added the label `maintenance` to this PR.

As I mentioned in https://github.com/microsoft/LightGBM/pull/5403#issuecomment-1205602981 and https://github.com/microsoft/LightGBM/pull/5413#issuecomment-1212065345, please add a labels when you create pull requests here, so your PR will be correctly categorized in the release notes.

shiyu1994:
> I think doing the following would prevent adding new complexity to .ci/test.sh:

Thanks for the suggestion. Already applied via [0206da8](https://github.com/microsoft/LightGBM/pull/5438/commits/0206da8f8e3f11b034ea508e2449e94bd3353f1a) and [eb34dd5](https://github.com/microsoft/LightGBM/pull/5438/commits/eb34dd5270bdf3304c3d2b0f52a42a1ee62affa0).

> We should never merge a PR that will result in CI being broken on master, so please do one of these:
> 1. push a fix for that issue to this PR
> 2. remove the cuda_exp CI jobs in this PR, document the work to add them back in a separate issue, merge this PR, and then push a follow-up PR fixing cuda_exp and adding back the CI jobs

I'd prefer the second choice, and already pushed the fixes via [ca7df38](https://github.com/microsoft/LightGBM/pull/5438/commits/ca7df38387c4beb3d005553a258d3022a394e10e).

>As I mentioned in https://github.com/microsoft/LightGBM/pull/5403#issuecomment-1205602981 and https://github.com/microsoft/LightGBM/pull/5413#issuecomment-1212065345, please add a labels when you create pull requests here, so your PR will be correctly categorized in the release notes.

Thanks again for the reminder.

.ci/test.sh
shiyu1994: Since now we don't introduce `device` and `DEVICE`, I've removed these changes in `.ci/test.sh`.

shiyu1994:
This is ready. @jameslamb @guolinke Could you please review this again. If it looks good to you, let's merge this so that we can work on related PRs. Thanks.

src/boosting/gbdt.cpp
```diff
@@ -440,23 +445,36 @@ bool GBDT::TrainOneIter(const score_t* gradients, const score_t* hessians) {
...
+    if (config_->boosting == std::string("goss")) {
+      // need to copy customized gradients when using GOSS
+      int64_t total_size = static_cast<int64_t>(num_data_) * num_tree_per_iteration_;
+      #pragma omp parallel for schedule(static)
+      for (int64_t i = 0; i < total_size; ++i) {
+        gradients_[i] = gradients[i];
+        hessians_[i] = hessians[i];
+      }
+      CHECK_EQ(gradients_pointer_, gradients_.data());
+      CHECK_EQ(hessians_pointer_, hessians_.data());
```
StrikerRUS, line 460: Aren't these checks slow? Maybe move them under `#ifdef DEBUG`?

tests/python_package_test/test_utilities.py
```diff
@@ -91,11 +91,17 @@ def dummy_metric(_, __):
         "INFO | [LightGBM] [Warning] CUDA currently requires double precision calculations.",
         "INFO | [LightGBM] [Info] LightGBM using CUDA trainer with DP float!!"
     ]
+    cuda_exp_lines = [
+        "INFO | [LightGBM] [Warning] Using sparse features with CUDA is currently not supported.",
```
StrikerRUS, line 95: Already in `gpu_lines`.
```suggestion
```

StrikerRUS (approved):
LGTM for `.github/workflows/cuda.yml` and `tests/python_package_test/test_utilities.py` except one minor simplification.

.ci/test.sh
jameslamb: agree, thanks

jameslamb (commented):
I don't have any additional comments.

Please address @StrikerRUS 's suggestions, and then I think @guolinke should re-review this (since you've added significant changes since his original review) prior to this being merged.

Thanks for fixing this!

src/boosting/gbdt.cpp
shiyu1994: Thanks. These checks are fast.

jameslamb:
@shiyu1994 Solaris support was removed in #5226. The `r-solaris` check no longer exists, and is no longer necessary.

But thank you for being so thorough!

shiyu1994:
@guolinke @jameslamb @StrikerRUS Thanks for reviewing this. I'll merge this since all tests have been passed and it is blocking many related PRs including #4827, #5425 and #4266.

PR #5446, opened 2022-08-29
[python-package] fix type hints on ctypes array converters

jameslamb:
Contributes to #3756.
Contributes to #3867.

`mypy` currently raises the following errors.

```text
python-package/lightgbm/basic.py:258: error: Function "ctypes.POINTER" is not valid as a type
python-package/lightgbm/basic.py:258: note: Perhaps you need "Callable[...]" or a callback protocol?
python-package/lightgbm/basic.py:266: error: Function "ctypes.POINTER" is not valid as a type
python-package/lightgbm/basic.py:266: note: Perhaps you need "Callable[...]" or a callback protocol?
python-package/lightgbm/basic.py:274: error: Function "ctypes.POINTER" is not valid as a type
python-package/lightgbm/basic.py:274: note: Perhaps you need "Callable[...]" or a callback protocol?
python-package/lightgbm/basic.py:282: error: Function "ctypes.POINTER" is not valid as a type
python-package/lightgbm/basic.py:282: note: Perhaps you need "Callable[...]" or a callback protocol?
```

These are correctly noting that `ctypes.POINTER` isn't a type...it's a function that *returns a type*. From https://docs.python.org/3/library/ctypes.html#ctypes.pointer

> [`ctypes.pointer()`] creates a new pointer instance, pointing to obj. The returned object is of the type `POINTER(type(obj))`

You can see this with the following example code.

```python
import ctypes

x = ctypes.c_float(1.1)
type(x)
# <class 'ctypes.c_float'>

ptr_type = ctypes.POINTER(type(x))
print(ptr_type)
# <class '__main__.LP_c_float'>

callable(ptr_type)
# True

isinstance(ctypes.pointer(x), ctypes.POINTER(ctypes.c_float))
# True
```

Since the functions `mypy` is complaining about all take in an object, check for a specific type with `isinstance()`, and raises an exception if the passed-in object doesn't have that type, I think it's appropriate to replace those `ctypes.POINTER` type hints with just `Any`.

### Notes for Reviewers

I tested this by running `mypy` as documented in #3867.

```shell
mypy \
    --ignore-missing-imports \
    python-package/
```

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -255,7 +255,7 @@ def _data_to_2d_numpy(data: Any, dtype: type = np.float32, name: str = 'list') -
                     "It should be list of lists, numpy 2-D array or pandas DataFrame")
 
 
-def cfloat32_array_to_numpy(cptr: ctypes.POINTER, length: int) -> np.ndarray:
+def cfloat32_array_to_numpy(cptr: Any, length: int) -> np.ndarray:
     """Convert a ctypes float pointer array to a numpy array."""
     if isinstance(cptr, ctypes.POINTER(ctypes.c_float)):
         return np.ctypeslib.as_array(cptr, shape=(length,)).copy()
@@ -263,7 +263,7 @@ def cfloat32_array_to_numpy(cptr: ctypes.POINTER, length: int) -> np.ndarray:
         raise RuntimeError('Expected float pointer')
 
 
-def cfloat64_array_to_numpy(cptr: ctypes.POINTER, length: int) -> np.ndarray:
+def cfloat64_array_to_numpy(cptr: Any, length: int) -> np.ndarray:
     """Convert a ctypes double pointer array to a numpy array."""
     if isinstance(cptr, ctypes.POINTER(ctypes.c_double)):
         return np.ctypeslib.as_array(cptr, shape=(length,)).copy()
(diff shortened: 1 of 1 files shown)


jmoralez:
I think we can instead use `ctypes.pointer[ctypes.c_int32]` for example. Reference: https://github.com/python/mypy/issues/7540#issuecomment-845741357

jameslamb:
@jmoralez hmmm, that doesn't look right to me.

```shell
docker run \
    --rm \
    --entrypoint="" \
    -it python:3.8-slim \
    python -c "import ctypes; print(ctypes.POINTER[ctypes.c_int32])"
```

>  File "<string>", line 1, in <module>
> TypeError: 'builtin_function_or_method' object is not subscriptable

I think that maybe the comment you found was someone getting lucky (or, I guess, unlucky) with the official type hints for `ctypes` being inaccurate.

See this from last month: https://github.com/python/typeshed/issues/8351#issuecomment-1191605215

jmoralez:
Ah I see, in the stub the lowercase pointer is a class but in ctypes it isn't haha. I think it's ok to go with Any.

jmoralez (approved):

guolinke (approved):

StrikerRUS:
Should we switch to `ctypes._Pointer` with a new mypy release in the future according to https://github.com/python/typeshed/pull/8446#issuecomment-1200426971?

jameslamb:
@StrikerRUS I'm nervous about relying on something preceded with a `_`, since I interpret that as being a private implementation detail that could change.

I want LightGBM's Python code to get the benefits of type-hinting, but I'm worried about the risk of user code being broken because of a change in `typeshed`.

StrikerRUS:
Just for reference:

> Especially given that `ctypes._Pointer` is actually documented -- it seems like it's not really *meant* to be a "private" class.
    https://github.com/python/typeshed/pull/8446#issuecomment-1214411604

However, I'm OK with `Any`.

StrikerRUS (approved):

jameslamb:
Thanks for pointing out that comment.

Even with that, I'm just nervous about the amount of change happening around this topic (indicated by the issues, you, @jmoralez , and I have linked to) and don't think the benefit of more accurate hints for these small utility functions is worth taking on the risk of incompatibilities of `lightgbm` with different versions of `ctypes`.

PR #5448, opened 2022-08-29
[ci] [R-package] ensure that MSVC jobs fail when tests fail (fixes #5439)

### LINKED ISSUE #5439: [R-package] [ci] unit test failures don't cause MSVC CI jobs to fail
## Description

The R-package CI jobs which compile the library with `CMake` + `MSVC` do not fail if the R package's unit tests fail.

* `r-package (windows-2019, MSVC, R 3.6, cmake)`
* `r-package (windows-2022, MSVC, R 4.1, cmake)`

## Reproducible example

I noticed this while looking at test failures on #5437 ([build link](https://github.com/microsoft/LightGBM/runs/8004155558?check_suite_focus=true)), but I suspect that this would happen for any change that causes the R unit tests to fail.

For example, I *suspect* (haven't confirmed) that adding something like that following anywhere in an R test case would reproduce this issue.

```r
stop("intentionally failing")
```

Screen Shot 2022-08-25 at 12 05 55 AM

But at the end of the tests, I see logs like the following.

```text
Error: Test failures
Execution halted

== Failed ======================================================================
-- 1. Failure (test_basic.R:93:3): train and predict binary classification
...
-- 10. Failure (test_basic.R:3190:5): lgb.train() only prints eval metrics when 

Maximum number of 10 failures reached, some test results may be missing.
```

## Additional Comments

I haven't investigated this much yet, just creating this issue so it isn't forgotten.

jameslamb:
Fixes #5439.


jameslamb:
In https://github.com/microsoft/LightGBM/pull/5448/commits/b9eebb391367784040a15cc89ad396472f7e2719, I tried intentionally causing a unit test failure, and this change:

```powershell
# before
Run-R-Code-Redirect-Stderr "source('testthat.R')" ; Check-Output $?

# after
Rscript.exe --vanilla "testthat.R" ; Check-Output $?
```

Saw that both CI jobs failed and reported test failures in logs ([build link](https://github.com/microsoft/LightGBM/runs/8063168637?check_suite_focus=true)).

Then, in https://github.com/microsoft/LightGBM/pull/5448/commits/7ce18e24f8445c23867ec6afdb035da26c9d012c, tried reverting the intentional test failure. Interestingly, the `windows-2022` job passed and the `windows-2019` job failed. Both *should have* passed.

* ❌ `windows-2019, MSVC, R 3.6, cmake` ([build link](https://github.com/microsoft/LightGBM/runs/8063315930?check_suite_focus=true))
* ✅ `windows-2022, MSVC, R 4.1, cmake` ([build link](https://github.com/microsoft/LightGBM/runs/8063315967?check_suite_focus=true))

I'll investigate this more tomorrow.

jameslamb:
iiiiinterresting!

commenting out all of the `dataset.R` tests results in the `windows-2019` job succeeding!

* commit: https://github.com/microsoft/LightGBM/pull/5448/commits/a075579f2851798a26b23d3480b6dbef6e3f1f68
* job: https://github.com/microsoft/LightGBM/runs/8219900714?check_suite_focus=true

jameslamb:
Alright, this is ready for review!

I believe the root cause of the original issue (#5439) is something in the intersection between:

* how the library compiled with MSVC interacts with stdout + stderr + R's output streams
* how `{testthat}` interacts with R's output streams
* the way that powershell sometimes treats output to stderr as a failure (see discussion from #3168)

This PR proposes two changes to ensure that MSVC R-package CI jobs actually fail when the tests fail:

* using `Rscript.exe testthat.R` directly to run the tests, instead of the `Run-R-Code-Redirect-Stderr` powershell function
* skipping one test in `test_dataset.R` that only fails on `windows-2019, MSVC, R 3.6, cmake` job

### Why I think it's ok to skip that one test only on that on job

* that test is passing on all other operating systems and versions of R tested in CI
    - including, especially, the combination of R 3.6 + Windows + cmake build (just with MinGW instead)
* the time spent investigating this one test on an old Visual Studio version + old powershell version + old R version could be better spent working on other items for the 4.0.0 release (#5153)
* re-enabling these two CI jobs will give LightGBM coverage of R + cmake + MSVC in CI again (which it currently doesn't really have, since all jobs silently pass)

### Evidence that the jobs will fail as of these changes

See the results from this commit which intentionally made one R unit test fail: https://github.com/microsoft/LightGBM/pull/5448/commits/aff9acf52db23224d9be635a8abf5a4afac0134f.

* ❌ `windows-2019, MSVC, R 3.6, cmake` ([build link](https://github.com/microsoft/LightGBM/runs/8240217966?check_suite_focus=true))
* ❌  windows-2022, MSVC, R 4.1, cmake ([build link](https://github.com/microsoft/LightGBM/runs/8240218037?check_suite_focus=true))

.ci/test_r_package_windows.ps1
```diff
@@ -283,7 +283,7 @@ if ($env:R_BUILD_TYPE -eq "cmake") {
 if ($env:COMPILER -eq "MSVC") {
   Write-Output "Running tests with testthat.R"
   cd R-package/tests
-  Run-R-Code-Redirect-Stderr "source('testthat.R')" ; Check-Output $?
+  Rscript.exe --vanilla "testthat.R" ; Check-Output $?
```
StrikerRUS, line 286: Please comment why you use `Rscript.exe` here to prevent accidental reversion of this PR in the future.

R-package/tests/testthat/test_dataset.R
```diff
@@ -218,6 +218,10 @@ test_that("lgb.Dataset: Dataset should be able to construct from matrix and retu
 })
 
 test_that("cpp errors should be raised as proper R errors", {
+  testthat::skip_if(
+    Sys.getenv("COMPILER", "") == "MSVC" && as.integer(R.Version()[["major"]]) < 4L
+    , message = "Skipping on R 3.x and Visual Studio 16 2019"
```
StrikerRUS, line 223:
```suggestion
    , message = "Skipping on R 3.x and Visual Studio"
```

StrikerRUS (approved):
Thanks for working on this!

LGTM except two minor suggestions below:

R-package/tests/testthat/test_dataset.R
jameslamb: ah good point, thanks. Updated in https://github.com/microsoft/LightGBM/pull/5448/commits/121e117edeb734e320cc85f1bd0a6660557823ef.

(I chose not to do this in the browser so that this change + the one for https://github.com/microsoft/LightGBM/pull/5448#discussion_r967875358 only generated one new commit, to save CI time).

.ci/test_r_package_windows.ps1
jameslamb: Added a comment in https://github.com/microsoft/LightGBM/pull/5448/commits/121e117edeb734e320cc85f1bd0a6660557823ef.

PR #5450, opened 2022-08-30
[python-package] fix mypy errors about callbacks

jameslamb:
Contributes to #3756.
Contributes to #3867.

`mypy` currently raises the following errors.

```text
-python-package/lightgbm/engine.py:688: error: Incompatible types in assignment (expression has type "Set[<nothing>]", variable has type "Optional[List[Callable[..., Any]]]")
-python-package/lightgbm/engine.py:692: error: Incompatible types in assignment (expression has type "Set[Callable[..., Any]]", variable has type "Optional[List[Callable[..., Any]]]")
-python-package/lightgbm/engine.py:695: error: Item "List[Callable[..., Any]]" of "Optional[List[Callable[..., Any]]]" has no attribute "add"
-python-package/lightgbm/engine.py:695: error: Item "None" of "Optional[List[Callable[..., Any]]]" has no attribute "add"
-python-package/lightgbm/engine.py:707: error: Item "None" of "Optional[List[Callable[..., Any]]]" has no attribute "__iter__" (not iterable)
-python-package/lightgbm/engine.py:708: error: Unsupported left operand type for - ("List[Callable[..., Any]]")
-python-package/lightgbm/engine.py:708: error: Unsupported left operand type for - ("None")
-python-package/lightgbm/engine.py:708: note: Left operand is of type "Optional[List[Callable[..., Any]]]"
-python-package/lightgbm/engine.py:709: error: Incompatible types in assignment (expression has type "List[Union[Callable[..., Any], Any]]", variable has type "Set[Union[Callable[..., Any], Any]]")
```

These errors come from variables' types changing. Specifically, the fact that in the code for `cv()`, the variable `callbacks` is initially assigned a value with type `Optional[List[Callable]]`, then is overwritten with a `set`, then the variable `callbacks_before_iter` is initialized as a `set` and overwritten with a list.

Similar to #5447, this PR proposes fixing those errors by using differently-named variables when types change.

### Notes for Reviewers

The same changes proposed in this PR were implemented in #4839 to fix the same code in `train()`.

I tested this by running `mypy` as documented in #3867.

```shell
mypy \
    --ignore-missing-imports \
    python-package/
```

### DIFF
--- a/python-package/lightgbm/engine.py
+++ b/python-package/lightgbm/engine.py
@@ -685,7 +685,7 @@ def cv(
 
     # setup callbacks
     if callbacks is None:
-        callbacks_set: Set[Callable] = set()
+        callbacks_set = set()
     else:
         for i, cb in enumerate(callbacks):
             cb.__dict__.setdefault('order', i - len(callbacks))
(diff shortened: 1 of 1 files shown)


StrikerRUS (approved):
Thanks for fixing this!

PR #5453, opened 2022-08-31
[CUDA] Add lambdarank objective for cuda_exp

shiyu1994:
Add lambdarank objective for `cuda_exp` (#5163). `rank_xendcg` will be added for `cuda_exp` after this PR is merged.


src/objective/objective_function.cpp
```diff
@@ -39,7 +40,7 @@ ObjectiveFunction* ObjectiveFunction::CreateObjectiveFunction(const std::string&
       return new CUDABinaryLogloss(config);
     } else if (type == std::string("lambdarank")) {
       Log::Warning("Objective lambdarank is not implemented in cuda_exp version. Fall back to boosting on CPU.");
```
guolinke, line 42: remove this warning?

src/objective/cuda/cuda_rank_objective.cu
```diff
@@ -0,0 +1,382 @@
...
+          }
+        }
+      }
+    }
+    __syncthreads();
+  }
+}
+
+template <data_size_t NUM_RANK_LABEL>
+__global__ void GetGradientsKernel_LambdarankNDCG_Sorted(
```
guolinke, line 189: differences compared with non sorted one?

src/objective/objective_function.cpp
shiyu1994: Done via [573bfd5](https://github.com/microsoft/LightGBM/pull/5453/commits/573bfd567cf3765a3bf5bd9e85f3cd32840b472c)

src/objective/cuda/cuda_rank_objective.cu
shiyu1994: `GetGradientsKernel_LambdarankNDCG_Sorted` does not sort the items. `GetGradientsKernel_LambdarankNDCG` sorts the items in shared memory. These two are for different numbers of items in a query. When a query contains too many items, the sorting cannot be performed in shared memory. Thus a sorting is done separately, and then `GetGradientsKernel_LambdarankNDCG_Sorted ` is used.

guolinke (approved):

src/objective/cuda/cuda_rank_objective.hpp
```diff
@@ -0,0 +1,56 @@
+/*!
+ * Copyright (c) 2021 Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See LICENSE file in the project root for
+ * license information.
+ */
+
+#ifndef LIGHTGBM_NEW_CUDA_RANKING_OBJECTIVE_HPP_
+#define LIGHTGBM_NEW_CUDA_RANKING_OBJECTIVE_HPP_
```
StrikerRUS, line 8:
```suggestion
#ifndef LIGHTGBM_OBJECTIVE_CUDA_CUDA_RANK_OBJECTIVE_HPP_
#define LIGHTGBM_OBJECTIVE_CUDA_CUDA_RANK_OBJECTIVE_HPP_
```

src/objective/cuda/cuda_rank_objective.hpp
```diff
@@ -0,0 +1,56 @@
...
+  const data_size_t* cuda_query_boundaries_;
+
+  // Host memory
+  int max_items_in_query_aligned_;
+};
+
+}  // namespace LightGBM
+
+#endif  // USE_CUDA_EXP
+#endif  // LIGHTGBM_NEW_CUDA_RANKING_OBJECTIVE_HPP_
```
StrikerRUS, line 56:
```suggestion
#endif  // LIGHTGBM_OBJECTIVE_CUDA_CUDA_RANK_OBJECTIVE_HPP_
```

StrikerRUS (commented):
Thanks for fixing `#ifndef` guards!
I don't have any other comments for this PR.

✔️

PR #5456, opened 2022-09-01
[python-package] simplify Dataset processing of label

jameslamb:
Contributes to #3756.
Contributes to #3867.

I'm beginning to work on the type hints and docstrings for the "array-like" inputs in `lightgbm` (https://github.com/microsoft/LightGBM/issues/3756#issuecomment-907622141), and decided to start with `label`.

While working on that, I noticed an opportunity to slightly simplify the `Dataset` code in `basic.py`.

This call in `Dataset._lazy_init()` is unnecessary

https://github.com/microsoft/LightGBM/blob/e02ddc445481b79b38de4fc44ff542af032c44cc/python-package/lightgbm/basic.py#L1504

because `label` is not referenced again until that method's call to `Dataset.set_label()`.

https://github.com/microsoft/LightGBM/blob/e02ddc445481b79b38de4fc44ff542af032c44cc/python-package/lightgbm/basic.py#L1577

which calls `_label_from_pandas()` again

https://github.com/microsoft/LightGBM/blob/e02ddc445481b79b38de4fc44ff542af032c44cc/python-package/lightgbm/basic.py#L2246

This PR proposes the following changes:

* eliminates that unnecessary call to `_label_from_pandas()`
* eliminates `_label_from_pandas()` entirely and moves its logic inside `Dataset.set_label()` (the only other place it's called)
* adds type hints for `label` in `Dataset`'s API


StrikerRUS:
I believe the `_label_from_pandas()` function itself can be removed now.
https://github.com/microsoft/LightGBM/blob/c8712a9e083bfa0922f6e860f49a8a374816f43e/python-package/lightgbm/basic.py#L608-L614

jameslamb:
ah! Totally agree @StrikerRUS . I had removed it originally, but lost that change accidentally while resolving a merge conflict.

Will push a commit here removing it.

jameslamb:
> `_label_from_pandas()` can be removed now

fixed in https://github.com/microsoft/LightGBM/pull/5456/commits/6ef5e7f030f04d1a404d1ee6c7dff934964c8348

PR #5457, opened 2022-09-01
[CUDA] Add L1 regression objective for cuda_exp

shiyu1994:
Add L1 regression objective for cuda_exp. This PR also includes implementations of basic CUDA algorithms like quantile calculation and more forms of bitonic sort.


src/cuda/cuda_algorithms.cu
```diff
@@ -139,6 +139,334 @@ void ShuffleReduceDotProdGlobal<label_t, double>(const label_t* values1, const l
...
+  SynchronizeCUDADevice(__FILE__, __LINE__);
+  GlobalInclusivePrefixSumReduceBlockKernel<REDUCE_T><<<1, GLOBAL_PREFIX_SUM_BLOCK_SIZE>>>(
+    block_buffer, num_blocks);
+  SynchronizeCUDADevice(__FILE__, __LINE__);
+  GlobalInclusivePrefixSumAddBlockBaseKernel<REDUCE_T><<<num_blocks, GLOBAL_PREFIX_SUM_BLOCK_SIZE>>>(
+    block_buffer, out_values, num_data);
+  SynchronizeCUDADevice(__FILE__, __LINE__);
+}
+
+template <>
```
guolinke, line 198: empty template?

src/objective/cuda/cuda_regression_objective.hpp
```diff
@@ -62,19 +59,48 @@ class CUDARegressionL2loss : public CUDAObjectiveInterface, public RegressionL2l
...
-  virtual void LaunchRenewTreeOutputCUDAKernel(
-    const double* /*score*/, const data_size_t* /*data_indices_in_leaf*/, const data_size_t* /*num_data_in_leaf*/,
-    const data_size_t* /*data_start_in_leaf*/, const int /*num_leaves*/, double* /*leaf_value*/) const {}
-
   const label_t* cuda_labels_;
   const label_t* cuda_weights_;
-  label_t* cuda_trans_label_;
-  double* cuda_block_buffer_;
+  CUDAVector<label_t> cuda_trans_label_;
+  CUDAVector<double> cuda_block_buffer_;
```
guolinke, line 65: cuda_labels_ and cuda_weights_ is `label_t*` type, why these two are `CUDAVector`?

guolinke (approved):
some minor comments

src/cuda/cuda_algorithms.cu
shiyu1994: This is for template instantiation.

src/objective/cuda/cuda_regression_objective.hpp
shiyu1994: cuda_labels_ and cuda_weights_ are just pointers to memory allocated by Metadata class, not allocated by the objective class. So we simply use pointers.

src/objective/cuda/cuda_regression_objective.cu
```diff
@@ -69,6 +69,126 @@ void CUDARegressionL2loss::LaunchGetGradientsKernel(const double* score, score_t
 }
 
 
+double CUDARegressionL1loss::LaunchCalcInitScoreKernel() const {
+  const double alpha = 0.9f;
```
StrikerRUS, line 73: @shiyu1994 Why did you use `0.9` here while if I'm not mistaken original CPU implementation uses `0.5`?
https://github.com/microsoft/LightGBM/blob/c8712a9e083bfa0922f6e860f49a8a374816f43e/src/objective/regression_objective.hpp#L237

PR #5396, opened 2022-07-31
[ci] migrate CI from macOS 10.15 to 11 (fixes #5391)

StrikerRUS:
Closed #5391.

Issue #5391, opened 2022-07-26
`macOS-10.15` Azure Pipelines environment is deprecated


jameslamb:
## Description

Noticed the following on a recent macOS build in Azure DevOps.

> ##[warning]The macOS-10.15 environment is deprecated, consider switching to macos-11(macos-latest), macos-12 instead. For more details see https://github.com/actions/virtual-environments/issues/5583

According to that issue

> **the image will be fully unsupported by 8/30/2022**

Between now and then, there will be several short brown-outs on GitHub Actions and Azure Devops where jobs will fail.
In addition, those services will start reducing capacity for the `macOS-10.15` environment, leading to increased queue times.

## Reproducible example

The `macOS regular` build in Azure DevOps have been showing this at least since July 26, 2022.
For example, see https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=13154&view=logs&j=31d31d6b-9897-5af8-16dd-840b6b6b4fac.

## Additional Comments

See #4797 for a discussion of why this is impactful (basically...we use this image to build wheels).

See #4849 for what had to be changed the last time this project made such an update.

jameslamb:
Noting that this has been extended, and https://github.com/actions/runner-images/issues/5583 now says

> will be fully unsupported by **12/1/22**

So we have a bit more time...but not much.

PR #5458, opened 2022-09-01
[python-package] add type hints on Dataset constructors

jameslamb:
Contributes to #3756.
Contributes to #3867.

Adds type hints on some methods in `Dataset` which don't currently have them.

### Notes for Reviewers

I'm beginning work on the "array-like" inputs in LightGBM's Python API, per https://github.com/microsoft/LightGBM/issues/3756#issuecomment-907622141.

This is a first step towards that.

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -1509,7 +1509,7 @@ class Dataset:
         feature_name='auto',
         categorical_feature='auto',
         params: Optional[Dict[str, Any]] = None
-    ):
+    ) -> "Dataset":
         if data is None:
             self.handle = None
             return self
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -1196,10 +1196,19 @@ class _InnerPredictor:
 class Dataset:
     """Dataset in LightGBM."""
 
-    def __init__(self, data, label=None, reference=None,
-                 weight=None, group=None, init_score=None,
-                 feature_name='auto', categorical_feature='auto', params=None,
-                 free_raw_data=True):
+    def __init__(
+        self,
+        data,
+        label=None,
+        reference: Optional["Dataset"] = None,
+        weight=None,
+        group=None,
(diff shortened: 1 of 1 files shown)


jmoralez:
WDYT about defining a type for the dataset handle like in https://github.com/microsoft/LightGBM/blob/0a5c5838eb679a43fc45443f71adf31ec506de22/include/LightGBM/c_api.h#L29

jameslamb:
> WDYT about defining a type for the dataset handle like in

Can you be more specific about what you're proposing for the Python side and what the benefit would be? I don't understand.

jmoralez:
Something like:
```python
DatasetHandle = ctypes.c_void_p

def __init_from_np2d(
    self,
    mat: np.ndarray,
    params_str: str,
    ref_dataset: Optional[DatasetHandle]
) -> "Dataset":
```

jameslamb:
ah ok, so just for documentation purposes? Yeah that works for me! I'll make that change here.

jameslamb:
updated in https://github.com/microsoft/LightGBM/pull/5458/commits/00b447789224c1c9aafcd4f393c3ce741e2ce6de

thanks for the suggestion!

jameslamb:
😬 

> This is a scheduled macOS-10.15 brownout. The macOS-10.15 environment is deprecated and will be removed on September 30th, 2022. For more details, see https://github.com/actions/virtual-environments/issues/5583

image

Azure DevOps build link: https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=13407&view=logs&j=382e14e0-e893-5ccd-73ac-68042457c8d8

Refer to #5391.

We have #5396 to fix this, but it's blocked by #5390, which is blocked by the need to switch to a new image for building wheels with a more recent version of glibc (https://github.com/microsoft/LightGBM/issues/5390#issuecomment-1221214158).

As that error message says, we have until September 30th to get this all done.

I'll prioritize changing that image and start working on it today.

PR #5460, opened 2022-09-01
[python-package] add more hints in sklearn.py

jameslamb:
Contributes to #3756.

Adds some more type hints to places in `sklearn.py` that are missing them.

### Notes for Reviewers

I intentionally avoided hints on anything "array-like", since those will require some extra research and should be handled separately (https://github.com/microsoft/LightGBM/issues/3756#issuecomment-907622141).


python-package/lightgbm/sklearn.py
```diff
@@ -686,16 +692,16 @@ def fit(
...
         eval_class_weight=None,
         eval_init_score=None,
         eval_group=None,
-        eval_metric=None,
+        eval_metric: Optional[_LGBM_ScikitEvalMetricType] = None,
         feature_name='auto',
         categorical_feature='auto',
         callbacks=None,
-        init_model=None
+        init_model: Optional[Union[str, Path, Booster, LGBMModel]] = None
```
jameslamb, line 704:
```suggestion
        init_model: Optional[Union[str, Path, Booster, "LGBMModel"]] = None
```

python-package/lightgbm/sklearn.py
```diff
@@ -43,6 +44,11 @@
         Union[_LGBM_EvalFunctionResultType, List[_LGBM_EvalFunctionResultType]]
     ],
 ]
+_LGBM_ScikitEvalMetricType = Union[
```
StrikerRUS, line 47: Please use new defined variable `_LGBM_ScikitEvalMetricType` in the `dask.py` file as well:
https://github.com/microsoft/LightGBM/blob/c8712a9e083bfa0922f6e860f49a8a374816f43e/python-package/lightgbm/dask.py#L408
https://github.com/microsoft/LightGBM/blob/c8712a9e083bfa0922f6e860f49a8a374816f43e/python-package/lightgbm/dask.py#L1040
https://github.com/microsoft/LightGBM/blob/c8712a9e083bfa0922f6e860f49a8a374816f43e/python-package/lightgbm/dask.py#L1166
https://github.com/microsoft/LightGBM/blob/c8712a9e083bfa0922f6e860f49a8a374816f43e/python-package/lightgbm/dask.py#L1337
https://github.com/microsoft/LightGBM/blob/c8712a9e083bfa0922f6e860f49a8a374816f43e/python-package/lightgbm/dask.py#L1492

jameslamb: agree, thanks for the suggestion!

just pushed https://github.com/microsoft/LightGBM/pull/5460/commits/347264d76c5c7cc3fc93f01a5fa42a4cbe9d7c15 with that change

python-package/lightgbm/sklearn.py
```diff
@@ -1187,16 +1193,16 @@ def fit(
...
         eval_set=None,
-        eval_names=None,
+        eval_names: Optional[List[str]] = None,
         eval_sample_weight=None,
         eval_init_score=None,
         eval_group=None,
-        eval_metric=None,
-        eval_at=(1, 2, 3, 4, 5),
+        eval_metric: Optional[_LGBM_ScikitEvalMetricType] = None,
+        eval_at: Iterable[int] = (1, 2, 3, 4, 5),
```
jmoralez, line 1201: I think this should be `Sequence[int]`, given that for example `range(5)` is an iterable but is not a valid type for this.

jameslamb: 🤩 Excellent point. I was just following the docstring and didn't consider this. Thanks very much for noting it!!!

I tried the following from the root of the repo, just to see what would happen:

<details><summary>sample code (click me)</summmary>

```python
from pathlib import Path
import numpy as np
import lightgbm as lgb
from sklearn.datasets import load_svmlight_file

rank_example_dir = Path('examples/lambdarank')
X_train, y_train = load_svmlight_file(str(rank_example_dir / 'rank.train'))
X_test, y_test = load_svmlight_file(str(rank_example_dir / 'rank.test'))
q_train = np.loadtxt(str(rank_example_dir / 'rank.train.query'))
q_test = np.loadtxt(str(rank_example_dir / 'rank.test.query'))
gbm = lgb.LGBMRanker(n_estimators=10)
gbm.fit(
    X_train,
    y_train,
    group=q_train,
    eval_set=[(X_test, y_test)],
    eval_group=[q_test],
    eval_at=range(3),
    callbacks=[
        lgb.early_stopping(10),
        lgb.reset_parameter(learning_rate=lambda x: max(0.01, 0.1 - 0.01 * x))
    ]
)
```

</details>

And you're right...passing a range for `eval_at` causes a failure when serializing the parameters to string to pass them through the C API functions.

```text
  File ".../site-packages/lightgbm/basic.py", line 326, in param_dict_to_str
    raise TypeError(f'Unknown type of parameter:{key}, got:{type(val).__name__}')
TypeError: Unknown type of parameter:eval_at, got:range
```

Given that, I think the hint here should be even stricter than `typing.Sequence`. Since this keyword argument is passed directly through to `params` and there's no other code in LightGBM manipulating its value, I think it can only accept values that are valid for `lightgbm.basic.param_dict_to_str()`.

For `eval_at`, I think that means only a list of ints or tuple of ints is valid. `param_dict_to_str()` supports `list`, `tuple`, and `set`, but `set` isn't appropriate for `eval_at` because sets aren't iterable (e.g. don't have any ordering).

https://github.com/microsoft/LightGBM/blob/3d4e08e1d8be9763f5d55180e3faf1e11eb20787/python-package/lightgbm/basic.py#L323

I just pushed https://github.com/microsoft/LightGBM/pull/5460/commits/81c234f27f35478ade88857d888190b8554ef47d which:

* sets the hint for `eval_at` to `Union[List[int], Tuple[int]]`
* replaces use of the word "iterable" in the relevant docstrings with "list or tuple of int"

python-package/lightgbm/dask.py
```diff
@@ -1490,7 +1490,7 @@ def fit(
         eval_init_score: Optional[List[_DaskVectorLike]] = None,
         eval_group: Optional[List[_DaskVectorLike]] = None,
         eval_metric: Optional[Union[_LGBM_ScikitCustomEvalFunction, str, List[Union[_LGBM_ScikitCustomEvalFunction, str]]]] = None,
```
StrikerRUS, line 1492:
```suggestion
        eval_metric: Optional[_LGBM_ScikitEvalMetricType] = None,
```

StrikerRUS (approved):
LGTM, thanks!

python-package/lightgbm/sklearn.py
jameslamb: @jmoralez I won't merge this until you have a chance to respond, since what I did here is slightly different than what you suggested.

jmoralez (approved):
Thanks!

PR #5461, opened 2022-09-01
Rename Metadata num_classes to be more clear

svotaw:
A recent completed PR (https://github.com/microsoft/LightGBM/pull/5299) created a new Metadata method num_classes(), but this can be confused with the actual num_classes in a Dataset.  It should definitely be the same number, but the new method is specifically returning the num_classes as configured and allocated in Metadata.  Reviewers requested to rename this method to be clearer.  New name is num_init_score_classes().


guolinke (approved):
Thank you!

PR #5467, opened 2022-09-04
fix: Revert leftover debugging in streaming test

svotaw:
Some debugging was leftover in some test code recently checked in.  This fixes it.


StrikerRUS (approved):
Thank you!

PR #5469, opened 2022-09-06
[CUDA] Add fair regression objective for cuda_exp

shiyu1994:
Add fair regression objective for `cuda_exp`. Fix the initial score calculation for `regression_l1` for `cuda_exp` as mentioned in comment https://github.com/microsoft/LightGBM/pull/5457#discussion_r962346600.


shiyu1994:
A slight difference in performance between `cuda_exp` and `cpu` is found with `fair` regression objective. In `test_regression` of this PR, with `objective=fair`, the `cpu` version produces a final `l2` metric of `16+`, while `cuda_exp` results in `11+`. The difference is recorded in issue #5470.

guolinke (approved):

Issue #5471, opened 2022-09-06
[CUDA] Random number generation on CUDA


shiyu1994:
## Summary
Support generation of random numbers on CUDA.

## Motivation
Sometimes random number generation is repeatedly called during the training process. For example, in the `rank_xendcg` objective. Supporting generation of random numbers on CUDA can accelerates these processes.

## References

See the code snippet below for random number generation in boosting with `rank_xendcg`
https://github.com/microsoft/LightGBM/blob/f0e3296232a98be70dc420977a3c5c5d731771ba/src/objective/rank_objective.hpp#L301-L328

shiyu1994:
Closed in favor of being in https://github.com/microsoft/LightGBM/issues/2302. We decided to keep all feature requests in one place.

Welcome to contribute this feature! Please re-open this issue (or post a comment if you are not a topic starter) if you are actively working on implementing this feature.

jameslamb:
Sorry, this was locked accidentally. Just unlocked it. We'd still love help with this feature!

PR #5472, opened 2022-09-06
[CUDA] Add rank_xendcg objective for cuda_exp

shiyu1994:
Add rank_xendcg objective for cuda_exp. The random number generation for boosting in `rank_xendcg` is still done in CPU.


shiyu1994:
Note that the efficiency of `rank_xendcg` can be further improved with the support of random number generation on CUDA, as is recorded in #5471.

guolinke (approved):

PR #5427, opened 2022-08-16
feature: Add serialization of reference dataset

svotaw:
## Summary
This is in reference to feature request: https://github.com/microsoft/LightGBM/issues/5426

This PR adds APIs for serializing/deserializing Datasets without their data to a byte array, effectively creating a "schema" or "reference" that can be used to create other Datasets.

## Implementation
The existing code for serializing Datasets to file was refactored to be able to go to any generic `BinaryWriter`, whether memory or file. The verbose serialization code was shared as much as possible, splitting methods into Header vs Data components.

Also, a generic `ByteBuffer` was created so that higher languages (e.g. Java) are removed from managing the byte memory of the serialized buffer.

## Test
New C++ tests were created to test both the serialization/deserialization and the new `ByteBuffer` functionality.

PR #5481, opened 2022-09-11
[tests] Fix cpp streaming data tests

StrikerRUS:
Based on https://github.com/microsoft/LightGBM/pull/5427#discussion_r956649785.


StrikerRUS:
cc @svotaw

jameslamb:
I've updated this with the newest changes on `master`.

@guolinke @shiyu1994 could you please review this? I think it should be a quick and non-controversial review.

If you have any requested changes, I can make them on this branch (to keep this PR moving forward, in case we don't hear from @StrikerRUS soon).

guolinke (approved):

PR #5484, opened 2022-09-14
[ci] test against R 4.2 for macOS and Linux CI jobs

jameslamb:
Contributes to #4881.

In #5274, I've been working on upgrading R to v 4.2.x in this project's CI jobs. Linux and macOS are working fine there, but adding Windows support has been challenging.

This PR proposes at least upgrading the Linux and macOS CI jobs, so we can start getting more test coverage of `{lightgbm}` on R 4.2.x. If this is merged, then I'll repurpose #5274 to focus only on adding the remaining Windows support.

### Notes for Reviewers

As of this writing, the latest version of R is v4.2.1 (https://cran.r-project.org/src/base/R-4/).

PR #5485, opened 2022-09-14
[R-package] improve safety of index selection in plotting

jameslamb:
I just observed the `lint` CI job failing on #5484 ([build link](https://github.com/microsoft/LightGBM/actions/runs/3054739687/jobs/4926997471)), with the following errors from R linting.

```text
[[1]]
/home/runner/work/LightGBM/LightGBM/R-package/R/lgb.plot.importance.R:82:12: warning: [seq] .N:1L is likely to be wrong in the empty edge case. Use seq_len(1L) instead.
  tree_imp[.N:1L,
           ^~~~~

[[2]]
/home/runner/work/LightGBM/LightGBM/R-package/R/lgb.plot.interpretation.R:149:23: warning: [seq] .N:1L is likely to be wrong in the empty edge case. Use seq_len(1L) instead.
  tree_interpretation[.N:1L,
                      ^~~~~
```

It looks like a new release of `{lintr}` (v3.0.1) just made it to CRAN (https://cran.r-project.org/web/packages/lintr/index.html), and they've added more cases caught by the `seq_linter()`.

This PR fixes those errors.

### DIFF
--- a/R-package/R/lgb.plot.importance.R
+++ b/R-package/R/lgb.plot.importance.R
@@ -79,7 +79,7 @@ lgb.plot.importance <- function(tree_imp,
     )
   )
 
-  tree_imp[.N:1L,
+  tree_imp[rev(seq_len(.N)),
            graphics::barplot(
                height = get(measure)
                , names.arg = Feature
--- a/R-package/R/lgb.plot.interpretation.R
+++ b/R-package/R/lgb.plot.interpretation.R
@@ -146,7 +146,7 @@ multiple.tree.plot.interpretation <- function(tree_interpretation,
   # create plot
   tree_interpretation[abs(Contribution) > 0.0, bar_color := "firebrick"]
   tree_interpretation[Contribution == 0.0, bar_color := "steelblue"]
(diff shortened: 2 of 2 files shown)


jmoralez (approved):
LGTM for fixing the linter errors. In the future we may want to check if the dataframes sent to these functions are empty because otherwise the errors are the same as when trying to plot empty vectors, which aren't very intuitive.
```
> barplot(height = numeric(0))
Error in plot.window(xlim, ylim, log = log, ...) : 
  need finite 'xlim' values
1: In min(w.l) : no non-missing arguments to min; returning Inf
2: In max(w.r) : no non-missing arguments to max; returning -Inf
3: In min(x) : no non-missing arguments to min; returning Inf
4: In max(x) : no non-missing arguments to max; returning -Inf
```

jameslamb:
closing and re-opening to re-run CI

jameslamb:
> In the future we may want to check if the dataframes sent to these functions are empty

sure, good idea! I'd support a PR like that if you want to open one.

PR #5488, opened 2022-09-15
[python-package] Print coefficients of linear leaves when plotting trees

AgFlore:
In #3299 we introduced linear trees, where the output of leaves are linear models of split-features (instead of constants). However the infomation about the linear model is missing in the outputs of `plot_tree` or `create_tree_digraph`, which is not desirable.


jameslamb (changes requested):
Thanks for this proposal.

Before we consider this, please include some images in the description showing what LightGBM produced prior to these changes and what you expect it to produce now. That will help reviewers to evaluate the usefulness of adding this information into plotting functions.

Please also add unit tests in https://github.com/microsoft/LightGBM/blob/master/tests/python_package_test/test_plotting.py testing each of the `if-else` branches in your proposed code change.

AgFlore:
Before these changes, linear leaf and non-linear leaf are printed the same, and the linear part of the linear leaf's output is missing.

After these changes, non-linear leaves will be unaffected, 

non-linear leaf

and linear leaves will have their linear part printed in the leaf.

linear leaf

Could you be more specific about what i shall implement and assert for these changes in the unit test?

jameslamb:
@AgFlore sorry for the VERY long delay in getting you a review on this!

if you're still interested in contributing this change, I'm ready and willing to work with you on it. To start, please merge the latest changes from `master` into your branch here.

> Could you be more specific about what i shall implement and assert for these changes in the unit test?

Look through the existing plotting tests in https://github.com/microsoft/LightGBM/blob/master/tests/python_package_test/test_plotting.py. Add new tests similar to those, but for a model using linear learners.

Add tests covering every combination of paths through this PR's proposed new `if` statements.

jameslamb:
It's been 1.5+ years since the last response here, so I'm assuming you're no longer interested in pursuing it. I'm closing this PR.

Thanks very much for your interest in LightGBM. We'd be happy to have you come back and contribute in the future if you have time to work with us.

PR #5491, opened 2022-09-15
[CUDA] Add multiclass_ova objective for cuda_exp

shiyu1994:
Add multiclass_ova objective for cuda_exp.


guolinke (approved):

shiyu1994:
@jameslamb @guolinke  Thanks for the help!

Issue #5495, opened 2022-09-22
Eigen/Dense”: No such file or directory


ArtificialZeng:
when I compile it. Eigen/Dense”: No such file or directory

## Environment info

LightGBM version or commit hash:

Command(s) you used to install LightGBM

```shell

```

ArtificialZeng:
> 

I‘m using Visual Studio. First I download the whole code, then I find out it's empty in the Eigen/Dense directory. It can't be compiled in the Visual Studio?

ArtificialZeng:
> Thanks for using LightGBM!
> 
> Since you didn't fill out the issue template (especially "commands you used to install LightGBM"), you're asking us to guess how you've installed LightGBM.
> 
> I suspect that maybe you are building the project from a cloned copy of this repository, and forgot to initialize the submodules.
> 
> If you've already cloned this repo with `git clone`, run the following prior to re-building.
> 
> ```shell
> git submodule init
> git submodule update --recursive
> ```
> 
> Then try rebuilding.
> 
> If this answer doesn't help you resolve the error, please fill out the description with all of the information that was requested.

Actually, What  I'm really asking is that: How to compile it from the source under Windows OS.

jameslamb:
LightGBM can be compiled with Visual Studio.

I'm not sure what specifically "download the code means", but if it is something other than `git clone --recursive`, then you need to run the `git` commands I shared to get Eigen and the other git submodules.

ArtificialZeng:
> LightGBM can be compiled with Visual Studio.
> 
> I'm not sure what specifically "download the code means", but if it is something other than `git clone --recursive`, then you need to run the `git` commands I shared to get Eigen and the other git submodules.

Or, the problem I mentioned is solved by using your method, but another bug appears: https://github.com/microsoft/LightGBM/issues/5499

jameslamb:
For those finding this issue from search engines... the "bug" mentioned here is that the source code zip file that GitHub generates for repos does not include the source of git submodules, and as a result LightGBM cannot be compiled from source code downloaded that way.

This is the same issue that was described and addressed in https://github.com/microsoft/LightGBM/issues/5038#issuecomment-1101503222.

I'm going to close this discussion. We can use #5498 to discuss how to update LightGBM's documentation to provide better guidance around this.

Issue #5038, opened 2022-02-28
eigen does not work


ysong2123:
## Description

I found out that:
Eigen in [https://github.com/microsoft/LightGBM/tree/fix/r-windows-ci/external_libs](https://nam11.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fmicrosoft%2FLightGBM%2Ftree%2Ffix%2Fr-windows-ci%2Fexternal_libs&data=04%7C01%7Cysong2%40paypal.com%7C3adb3c9d373a498d853708d9fa7dac8c%7Cfb00791460204374977e21bac5f3f4c8%7C0%7C0%7C637816240551678764%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C2000&sdata=SGdIN108SmHn0063E2Hhg4XgMrnnbg5Ln33mBtAUtfM%3D&reserved=0) is not reachable,   
and i can't  wget https://gitlab.com/libeigen/eigen either.

My goal:  wanted to  build lightgbm in my local, due to Mac Big Sur incompatible with latest lightgbm lib.
could you team help me on this?
## Environment info

LightGBM version or commit hash:

Command(s) you used to install LightGBM

```shell

```

jameslamb:
Thanks for using LightGBM and for this report. Since you said your goal is "to build lightgbm in my local", can you please fill in the request information "Command(s) you used to install LightGBM"? You haven't told us, even, whether your are trying to build the CLI, Python package, R package, etc.

You should not be trying to build the `fix/r-windows-ci` branch from #5037 and you shouldn't need to "wget" Eigen.

Please see:
* Installation Guide ([link](https://lightgbm.readthedocs.io/en/latest/Installation-Guide.html)): instructions for compiling LightGBM
* Python README ([link](https://github.com/microsoft/LightGBM/blob/master/python-package/README.rst)): instructions for installing the Python package
* R README ([link](https://github.com/microsoft/LightGBM/blob/master/R-package/README.md)): instructions for installing the R package

ysong2123:
my original plan is 'pip install lightgbm', assuming that everything will be ok afterwards.   but i got same issue when i tried to import lightgbm as lgb as below:
https://stackoverflow.com/questions/44937698/lightgbm-oserror-library-not-loaded
[...]

PR #368, opened 2017-03-27
[WIP]Initial GPU acceleration support for LightGBM

huanzhang12:
This set of patches adds initial GPU acceleration for LightGBM, by accelerating histogram construction on GPUs. The implementation is highly modular and does not affect existing features of LightGBM; GPU kernel code is mostly independent of other parts of LightGBM, so long-term maintenance is easier.

I add a new type of tree learner in gpu_tree_learner.h and gpu_tree_learner.cpp. All GPU interfacing code is inside these two files. Changes to other parts of LightGBM are kept minimal; I only make some small interface changes to make necessary data structure available for the GPU tree learner. GPU code (in folder src/treelearner/ocl/) are implemented in OpenCL and tested on both AMD and NVIDIA GPUs.

For building and testing procedure, and some initial performance comparison, please see:
https://github.com/huanzhang12/lightgbm-gpu
(We probably want to move these instructions to Wiki after merging)

PR #5498, opened 2022-09-23
[docs] fix GUI-only instructions for building with Visual Studio

ArtificialZeng:


ArtificialZeng:
or otherwise it will report a bug like this:https://github.com/microsoft/LightGBM/issues/5495

jameslamb (changes requested):
Thanks!

Yes, I can confirm that the zip archive linked to from this section doesn't contain the contents of `git` submodules, and this section's instructions therefore don't work.

These instructions were written AT LEAST 5 years ago (https://github.com/microsoft/LightGBM/blame/3eb68e85e99a221b04c0e19eb262e41e210f3da1/docs/Installation-Guide.rst#L59, #945), and maybe even longer back. I suspect that they were written prior to the first `git` submodule being introduced in April 2017 (#368), and have been broken since then. Thanks for catching it!

-----

However, I don't think the proposed fix is the best option.

For people who are already comfortable using `git`, I think the "From Command Line" option should be preferred, and that this "With GUI" option should be reserved for only using pointing and clicking.

I think we should alter the "With GUI" instructions so that the point users to the zip archive attached to each release, which does include all the submodules (as described in https://github.com/microsoft/LightGBM/issues/5038#issuecomment-1103394333)

That would mean that the "With GUI" instructions are insufficient for building versions of LightGBM other than released versions, but I think that's acceptable in exchange for minimizing the maintenance burden required to keep this documentation up to date. "You have to use `git` if you want to build an unreleased version" seems ok to me.

So instead of this, can you instead please change these instructions to something like the following:

```text
2. Navigate to https://github.com/microsoft/LightGBM/releases. Choose the release you'd like to build, and download the file with a name like ``LightGBM-complete_source_code_tar_gz.tar.gz``.
```

Would you like to make that change? If you'd prefer not to, I'm happy to open a pull request and `@` you in it for a review.

StrikerRUS (changes requested):
I absolutely agree with @jameslamb's suggestion above!

> and maybe even longer back. I suspect that they were written prior to the first `git` submodule being introduced in April 2017 (#368), and have been broken since then. 

Just small clarification: instructions are broken since introducing `fmt` and `fast_double_parser` dependencies (#3405, merged 8 Dec 2020), because `compute` dependency is needed only for GPU-version and it cannot be compiled via GUI.

> That would mean that the "With GUI" instructions are insufficient for building versions of LightGBM other than released versions

To overcome this, we can include those archives in our nightly builds and redirect users to them.

jameslamb:
It's been a few months without response here, so I'm going to close this in favor of #5649.

Thanks for your interest in LightGBMM, @ArtificialZeng ! We hope you'll consider coming back and contributing in the future when you have more time.

PR #5500, opened 2022-09-23
Support compiling with `std:c++latest` using MSVC  (fixes #5150)

### LINKED ISSUE #5150: [MSVC][std:c++latest] LightGBM failed to build with /std:c++latest on MSVC
## Description
Hi all,
Recently, we found some errors when building LightGBM with '/std:c++latest'. Could someone please take a first look? Thanks in advance.

## Reproducible example
1. git clone https://github.com/microsoft/LightGBM F:\gitP\microsoft\LightGBM
2. cd F:\gitP\microsoft\LightGBM & mkdir build_amd64
3. set CL= /std:c++latest
4. cd F:\gitP\microsoft\LightGBM\build_amd64
5. cmake -G "Visual Studio 16 2019" -A x64 -DCMAKE_SYSTEM_VERSION=10.0.18362.0 -DCMAKE_BUILD_TYPE=Release -DBUILD_CPP_TEST=ON ..
6. msbuild /m /p:Platform=x64 /p:Configuration=Release lightgbm.sln /t:Rebuild
## Environment info

LightGBM version or commit hash: 7e4780479da7a7845630eebc7e73fb6710fc7dc4
Visual Studio Version: Release 16.11.12
Platform: Windows

## Additional Comments
[build.log](https://github.com/microsoft/LightGBM/files/8479960/build.log)
Actual Behavior:
F:\gitP\microsoft\LightGBM\include\LightGBM/utils/common.h(1209,53): error C7595: 'fmt::v8::basic_format_string<char,const T &>::basic_format_string': call to immediate function is not a constant expression 
F:\gitP\microsoft\LightGBM\include\LightGBM/utils/common.h(1209,53): error C7595: 'fmt::v8::basic_format_string<char,const T &>::basic_format_string': call to immediate function is not a constant expression 
F:\gitP\microsoft\LightGBM\include\LightGBM/utils/common.h(1209,53): error C7595: 'fmt::v8::basic_format_string<char,const T &>::basic_format_string': call to immediate function is not a constant expression 
F:\gitP\microsoft\LightGBM\include\LightGBM/utils/common.h(1209,53): error C7595: 'fmt::v8::basic_format_string<char,const T &>::basic_format_string': call to immediate function is not a constant expression

QuellaZhang:
Fixes https://github.com/microsoft/LightGBM/issues/5150.


jameslamb (changes requested):
Sorry for the long delay, we had some CI issues and then needed to focus on a time-sensitive release.

Those things are resolved, and I'm ready to help move this forward. Could you please merge  the latest changes from `master` into this branch? Then we'll be able to provide more of a review.

Thanks very much for your patience, and sorry for the inconvenience.

QuellaZhang:
@jameslamb, I have updated the branch.

jameslamb (changes requested):
Thanks for this!

Based on successful CI jobs, this change looks ok to me...but the passing CI jobs don't prove that it fixes compilation with MSVC using `std:c++latest`. As I mentioned, LightGBM is using the C++11 standard everywhere: https://github.com/microsoft/LightGBM/issues/5150#issuecomment-1098667335.

@guolinke @shiyu1994 Are you comfortable with LightGBM committing to continue supporting `std:c++latest`, as new C++ standards come out and the specific spec referenced by `latest` changes? ([documentation](https://learn.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version?view=msvc-170)). If so, I think a new CI job using this option should be added as part of this PR, to ensure that future LightGBM doesn't break this support.

If not, and you'd instead prefer to instead just accept targeted patches like this based on reported issues from users, then it's probably ok to not add a new CI job at this time.

StrikerRUS:
cc @AlbertoEAF

StrikerRUS:
@jameslamb 
> I think a new CI job using this option should be added as part of this PR

WDYT about using `std:c++latest` flag **in a part** of already existing our CI jobs to save CI time? For example, that flag can be added in R-package CI job that runs against MSVC-2022
https://github.com/microsoft/LightGBM/blob/0c0eb2a6a1ba111f0896dd16a579342fa16f7629/.github/workflows/r_package.yml#L95-L101

jameslamb:
yeah @StrikerRUS , I'm ok with that!

I think that, based on https://learn.microsoft.com/en-us/cpp/build/reference/cl-environment-variables?view=msvc-170, this could be done by setting

```powershell
$env:CL = "/std:c++latest"
```

somewhere. Does that seem right to you?

jameslamb:
@QuellaZhang I'm ready to work with you on getting this PR merged.

Are you interested in trying to update it to use `std:c++latest` on the R package's MSVC job as @StrikerRUS suggested? If you don't have the time or interested (or if I don't hear back from you in the next few weeks), I'll open a separate PR with these changes + the testing changes.

PR #5503, opened 2022-09-24
[ci] [R-package] use R 4.2.1 in Windows CI jobs (fixes #4881)

### LINKED ISSUE #4881: [R-package] Support R 4.2

jameslamb:
Fixes #4881.
Replaces #5274.

### Changes in this PR

* upgrade from R 4.1.3 to R 4.2.1 in Windows CI jobs
* upgrade from Rtools40 to Rtools42 in Windows CI jobs
* explicitly pass version parsed from `VERSION.txt` into `test_r_package_windows.ps1` (instead of relying on wildcards like `lightgbm_*.tar.gz`
* extend "skip one test when compiling with MSVC" fix from #5448 to all MSVC jobs

### Why did the check in `configure.win` need to change?

Compiling the test program used to check for `inet_pton` with the compiler that ships with RTools42 (gcc 10.3), results in the following error.

```text
conftest.cpp: In function 'int main()':
conftest.cpp:3:13: error: invalid conversion from 'INT (*)(INT, LPCSTR, PVOID)' {aka 'int (*)(int, const char*, void*)'} to 'void*' [-fpermissive]
    3 |   void* p = inet_pton;
      |             ^~~~~~~~~
      |             |
      |             INT (*)(INT, LPCSTR, PVOID) {aka int (*)(int, const char*, void*)}
```

To avoid that casting, this PR proposes just assigning the address of `inet_pton` to a function pointer for a function with exactly the same signature as the one LightGBM's code expects.

### Why does the use of `"lightgbm_*.tar.gz"` need to be replaced with `"lightgbm_$env:LGB_VER.tar.gz"`?

I'm not sure...probably related to R or some tool shipping with RTools42 not working well with wildcards. While testing in #4881, I saw errors like the following:

> Error: no packages were specified

### Notes for Reviewers

#4881 refers to creating a comment-triggered workflow for the UCRT toolchain for Windows builds, but I don't think that's necessary any more now that R 4.2.0 has actually been released and Rtools42 contains the UCRT build tools.

I closed #4881 in favor of this because that PR was getting very busy with debugging commits and comments.

### References

There is some documentation about Rtools42 adding its own compilers to PATH, but I think that just means to R's PATH (referenced by `R CMD` commands).

e.g., see https://cran.r-project.org/bin/windows/base/howto-R-4.2.html

> To make the use of Rtools42 simpler, when R is installed via the binary installer it by default uses Rtools42 for the compilers and libraries. PATH will be set by R (inside front-ends like RGui and RTerm, but also R CMD) to include the build tools (e.g. make) and the compilers (e.g. gcc). In addition, R installed via the binary installer will automatically set R_TOOLS_SOFT (and LOCAL_SOFT for backwards compatibility) to the Rtools42 location for building R packages. This feature is only present in the installer builds of R, not when R is installed from source.

and https://github.com/r-lib/actions/issues/574#issue-1265095474

Some additional relevant information in the `r-lib/actions` project:

* https://github.com/r-lib/actions/commit/7d98418c6b5cfca592b100e2f348fbe90c18bcf3
* https://github.com/r-lib/actions/blob/8e8f7edba0e5bda9318f9377c0dc9c2f5d0fb786/setup-r/src/installer.ts#L431-L457

### How I tested this

Given the changes to `configure.win`, I also tested the project on R Hub and win-builder.

Generated the CRAN source distribution like this:

```shell
sh build-cran-package.sh
```

Then used that `.tar.gz` to test on win-builder and R Hub.

<details><summary>passing on win-builder (click me)</summary>

* ✅ `R-release` ([logs](https://win-builder.r-project.org/q36Xf4UFIU30/))
* ✅ `R-devel` ([logs](https://win-builder.r-project.org/MotfC88oY14R/))
* ✅ `R-oldrelease` ([logs](https://win-builder.r-project.org/kllIkud3FMk5/))

See https://win-builder.r-project.org/. This project is run by some of the CRAN maintainers, and very closely matches the Windows builds on CRAN.

There are functions in `{devtools}` to automate uploading to this service, e.g. `devtools::check_win_release()`. Unfortunately, those only take a *director* to be packaged with `R CMD build`. That doesn't work with the custom stuff `build-cran-package.sh` has to do to build vignettes:

https://github.com/microsoft/LightGBM/blob/dc4794b62f2c09103e558b7a8325bed66083a5c3/build-cran-package.sh#L206

So I manually changed the maintainer to myself in `DESCRIPTION`, built the package with `build-cran-package.sh`, then uploaded manually at https://win-builder.r-project.org/upload.aspx.

</details>

<details><summary>passing on R Hub (click me)</summary>

I tried checking the package built from this branch on all 4 platforms supported by R Hub.

* ✅  `windows-x86_64-devel` ([logs](https://builder.r-hub.io/status/lightgbm_3.3.2.99.tar.gz-230afcf8ccc9410aa046474f38d128df))
* ✅ `windows-x86_64-oldrel` ([logs](https://builder.r-hub.io/status/lightgbm_3.3.2.99.tar.gz-306552d2418447dcaa71ef8b0428f570))
* ✅ `windows-x86_64-patched` ([logs](https://builder.r-hub.io/status/lightgbm_3.3.2.99.tar.gz-1f2a23040e924bb4a6996cd8fa27ef95))
* ✅ `windows-x86_64-release` ([logs](https://builder.r-hub.io/status/lightgbm_3.3.2.99.tar.gz-1f077751637e40b1abfaf65156f8f27a))

```r
EMAIL <- "****" # my personal email
PACKAGE_TARBALL <- "lightgbm_3.3.2.99.tar.gz"

result <- rhub::check(
    path = PACKAGE_TARBALL
    , email = EMAIL
    , check_args = c(
        "--as-cran"
    )
    , platforms = c(
        "windows-x86_64-devel"
        , "windows-x86_64-oldrel"
        , "windows-x86_64-patched"
        , "windows-x86_64-release"
    )
    , env_vars = c(
        "R_COMPILE_AND_INSTALL_PACKAGES" = "always"
        , "_R_CHECK_FORCE_SUGGESTS_" = "true"
        , "_R_CHECK_CRAN_INCOMING_USE_ASPELL_" = "true"
    )
)
```

</details>

### DIFF
--- a/R-package/README.md
+++ b/R-package/README.md
@@ -97,12 +97,17 @@ After installing `Rtools` and `CMake`, be sure the following paths are added to 
     - If you have `Rtools` 4.0, example:
         - `C:\rtools40\mingw64\bin`
         - `C:\rtools40\usr\bin`
+    - If you have `Rtools` 4.2, example:
+        - `C:\rtools42\x86_64-w64-mingw32.static.posix\bin`
+        - `C:\rtools42\user\bin`
 * `CMake`
     - example: `C:\Program Files\CMake\bin`
 * `R`
     - example: `C:\Program Files\R\R-3.6.1\bin`
 
 NOTE: Two `Rtools` paths are required from `Rtools` 4.0 onwards because paths and the list of included software was changed in `Rtools` 4.0.
+
+NOTE: `Rtools42` takes a very different approach to the compiler toolchain than previous releases, and how you install it changes what is required to build packages. See ["Howto: Building R 4.2 and packages on Windows"](https://cran.r-project.org/bin/windows/base/howto-R-4.2.html).
 
 #### Windows Toolchain Options
 
--- a/R-package/README.md
+++ b/R-package/README.md
@@ -99,7 +99,7 @@ After installing `Rtools` and `CMake`, be sure the following paths are added to 
         - `C:\rtools40\usr\bin`
     - If you have `Rtools` 4.2, example:
         - `C:\rtools42\x86_64-w64-mingw32.static.posix\bin`
-        - `C:\rtools42\user\bin`
+        - `C:\rtools42\usr\bin`
 * `CMake`
     - example: `C:\Program Files\CMake\bin`
 * `R`
--- a/.ci/test_r_package_windows.ps1
+++ b/.ci/test_r_package_windows.ps1
@@ -54,11 +54,13 @@ Remove-From-Path ".*PostgreSQL.*"
 Remove-From-Path ".*\\R\\.*"
 Remove-From-Path ".*R Client.*"
 Remove-From-Path ".*rtools40.*"
+Remove-From-Path ".*rtools42.*"
 Remove-From-Path ".*shells.*"
 Remove-From-Path ".*Strawberry.*"
 Remove-From-Path ".*tools.*"
 
 Remove-Item C:\rtools40 -Force -Recurse -ErrorAction Ignore
+Remove-Item C:\rtools42 -Force -Recurse -ErrorAction Ignore
 
 # Get details needed for installing R components
 #
@@ -74,11 +76,11 @@ if ($env:R_MAJOR_VERSION -eq "3") {
   $env:RTOOLS_EXE_FILE = "rtools35-x86_64.exe"
   $env:R_WINDOWS_VERSION = "3.6.3"
 } elseif ($env:R_MAJOR_VERSION -eq "4") {
-  $RTOOLS_INSTALL_PATH = "C:\rtools40"
+  $RTOOLS_INSTALL_PATH = "C:\rtools42"
   $env:RTOOLS_BIN = "$RTOOLS_INSTALL_PATH\usr\bin"
-  $env:RTOOLS_MINGW_BIN = "$RTOOLS_INSTALL_PATH\mingw64\bin"
-  $env:RTOOLS_EXE_FILE = "rtools40v2-x86_64.exe"
-  $env:R_WINDOWS_VERSION = "4.1.3"
+  $env:RTOOLS_MINGW_BIN = "$RTOOLS_INSTALL_PATH\x86_64-w64-mingw32.static.posix\bin"
+  $env:RTOOLS_EXE_FILE = "rtools42-5253-5107.exe"
+  $env:R_WINDOWS_VERSION = "4.2.1"
 } else {
   Write-Output "[ERROR] Unrecognized R version: $env:R_VERSION"
   Check-Output $false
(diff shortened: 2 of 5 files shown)


R-package/README.md
```diff
@@ -97,13 +97,18 @@ After installing `Rtools` and `CMake`, be sure the following paths are added to
     - If you have `Rtools` 4.0, example:
         - `C:\rtools40\mingw64\bin`
         - `C:\rtools40\usr\bin`
+    - If you have `Rtools` 4.2, example:
+        - `C:\rtools42\x86_64-w64-mingw32.static.posix\bin`
+        - `C:\rtools42\user\bin`
```
jmoralez, line 102:
```suggestion
        - `C:\rtools42\usr\bin`
```

jmoralez (approved):
I'm approving to unblock but I'm not the best person to review this. Hopefully someone else can provide their review.

R-package/README.md
jameslamb: ha thank you for catching this! Now you know how I pronounce this path in my head 😂

jameslamb:
Thanks for much for that honesty, @jmoralez , much appreciated!

I agree, it would have been better to have @guolinke or @shiyu1994 review this. Normally I'd wait for one of them, but since #5525 is so time-sensitive, I'm going to merge this to unblock testing there.

These changes will make it into that PR so they'll have another chance to review there, and I think all of the extra testing with win-builder and R Hub there will be give us high confidence that this is working (or catch any issues with it 😬 ).

guolinke:
@jameslamb no problems, this PR looks good to me.

jameslamb:
ok thank you!

Issue #5507, opened 2022-09-25
[ci] Appveyor builds failing


jameslamb:
## Description

For the last 2 days, Appveyor CI jobs have been failing, blocking development on this project.
I've seen these changes on several, totally-unrelated PRs (e.g. #5503, #5504, #5505, #5506), so I'm fairly sure it's related to that job's configuration and not breaking changes on PRs.

## Reproducible example

See any recent appveyor jobs, for example this one from #5506: https://ci.appveyor.com/project/guolinke/lightgbm/builds/44872180/job/sxsvo19536lcd058.

## Additional Comments

LightGBM is compiling successfully, but with several errors in the Python unit tests.

```text
C:\Miniconda3-x64\envs\test-env\lib_pypy\_functools.py:80: in __call__
    return self._func(*(self._args + fargs), **fkeywords)
C:\Miniconda3-x64\envs\test-env\lib\site-packages\sklearn\utils\estimator_checks.py:3489: in check_fit_idempotent
    estimator.fit(X_train, y_train)
C:\Miniconda3-x64\envs\test-env\lib\site-packages\lightgbm\sklearn.py:1011: in fit
    init_model=init_model
C:\Miniconda3-x64\envs\test-env\lib\site-packages\lightgbm\sklearn.py:806: in fit
    callbacks=callbacks
C:\Miniconda3-x64\envs\test-env\lib\site-packages\lightgbm\engine.py:223: in train
    booster = Booster(params=params, train_set=train_set)
C:\Miniconda3-x64\envs\test-env\lib\site-packages\lightgbm\basic.py:2775: in __init__
    train_set.construct()
C:\Miniconda3-x64\envs\test-env\lib\site-packages\lightgbm\basic.py:1929: in construct
    self.feature_name = self.get_feature_name()
C:\Miniconda3-x64\envs\test-env\lib\site-packages\lightgbm\basic.py:2380: in get_feature_name
    ptr_string_buffers = (ctypes.c_char_p * num_feature)(*map(ctypes.addressof, string_buffers))
C:\Miniconda3-x64\envs\test-env\lib_pypy\_ctypes\array.py:203: in __init__
    self[i] = arg
C:\Miniconda3-x64\envs\test-env\lib_pypy\_ctypes\array.py:229: in __setitem__
    cobj = self._type_.from_param(value)
C:\Miniconda3-x64\envs\test-env\lib_pypy\_ctypes\primitive.py:382: in from_param
[...]

PR #5509, opened 2022-09-27
[ci] Install matplotlib and graphviz when creating conda env

shiyu1994:
This is trying to solve #5507.

### DIFF
--- a/.ci/test_windows.ps1
+++ b/.ci/test_windows.ps1
@@ -30,7 +30,8 @@ conda init powershell
 conda activate
 conda config --set always_yes yes --set changeps1 no
 conda update -q -y conda
-conda create -q -y -n $env:CONDA_ENV python=$env:PYTHON_VERSION ; Check-Output $?
+# matplotlib and python-graphviz have to be installed separately to prevent conda from downgrading to pypy
+conda create -q -y -n $env:CONDA_ENV python=$env:PYTHON_VERSION matplotlib python-graphviz ; Check-Output $?
 if ($env:TASK -ne "bdist") {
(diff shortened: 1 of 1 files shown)


jameslamb:
ha oh! Didn't know you were also working on this. I just opened #5510 with an alternative approach that I hope might be a bit more future-proof than re-ordering when we install things.

Let's see which one works 😬

shiyu1994:
@jameslamb Thanks. I think both solutions are okay as long as they work. I can close this one if you feel that #5510 is a better solution.

jameslamb (approved):
I do feel like #5510 might be a more permanent solution to this...but I'm approving this since it's building successfully for now. That way, you can merge this to unblock CI and work on other things (since it's late for me here in the U.S. and I'm about to go to sleep).

I'll probably continue experimenting with #5510 more this week.

Thanks for working on this!

shiyu1994:
Thanks. But it seems that Windows bdist build fails now with Azure DevOps. Not sure if it is related to this change. Will restart the tests to check.

shiyu1994:
https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=13642&view=logs&j=67ae2ff8-4b33-578b-0813-9c0440091667

jameslamb:
I just pushed a change over in https://github.com/microsoft/LightGBM/pull/5510 that I think could be a permanent fix. Learned today that it's possible to include build tags in the environment description for conda, so for example you can do

```shell
conda install 'python=3.7.12[build=*cpython]`
```

shiyu1994:
Close with a replacing solution in #5510.

PR #5505, opened 2022-09-24
[tests][dask] solve timeouts

jmoralez:
Sometimes the dask tests get stuck on the `_train_part` function with the following call stack:
```
File "/home/jose/mambaforge/envs/lgb-test/lib/python3.9/threading.py", line 937, in _bootstrap self._bootstrap_inner()
File "/home/jose/mambaforge/envs/lgb-test/lib/python3.9/threading.py", line 980, in _bootstrap_inner self.run()
File "/home/jose/mambaforge/envs/lgb-test/lib/python3.9/threading.py", line 917, in run self._target(*self._args, **self._kwargs)
File "/home/jose/mambaforge/envs/lgb-test/lib/python3.9/site-packages/distributed/threadpoolexecutor.py", line 57, in _worker task.run()
File "/home/jose/mambaforge/envs/lgb-test/lib/python3.9/site-packages/distributed/_concurrent_futures_thread.py", line 65, in run result = self.fn(*self.args, **self.kwargs)
File "/home/jose/mambaforge/envs/lgb-test/lib/python3.9/site-packages/distributed/worker.py", line 2882, in apply_function msg = apply_function_simple(function, args, kwargs, time_delay)
File "/home/jose/mambaforge/envs/lgb-test/lib/python3.9/site-packages/distributed/worker.py", line 2904, in apply_function_simple result = function(*args, **kwargs)
File "/hdd/github/LightGBM/python-package/lightgbm/dask.py", line 322, in _train_part model.fit(
File "/hdd/github/LightGBM/python-package/lightgbm/sklearn.py", line 1084, in fit super().fit(
File "/hdd/github/LightGBM/python-package/lightgbm/sklearn.py", line 797, in fit self._Booster = train(
File "/hdd/github/LightGBM/python-package/lightgbm/engine.py", line 223, in train booster = Booster(params=params, train_set=train_set)
File "/hdd/github/LightGBM/python-package/lightgbm/basic.py", line 2775, in __init__ train_set.construct()
File "/hdd/github/LightGBM/python-package/lightgbm/basic.py", line 1923, in construct self._lazy_init(self.data, label=self.label,
File "/hdd/github/LightGBM/python-package/lightgbm/basic.py", line 1578, in _lazy_init self.__init_from_np2d(data, params_str, ref_dataset)
File "/hdd/github/LightGBM/python-package/lightgbm/basic.py", line 1708, in __init_from_np2d _safe_call(_LIB.LGBM_DatasetCreateFromMat( 
```
[...]

Issue #5502, opened 2022-09-24
[R-package] CRAN error: r-devel-linux-x86_64-debian-clang


jameslamb:
## Description

CRAN checks for `{lightgbm}` are showing an ERROR for the `r-devel-linux-x86_64-debian-clang` check flavor.

<img width="526" alt="image" src="https://user-images.githubusercontent.com/7608904/192075976-0fdd0a00-09b7-47d7-a904-f57658110dda.png">

https://cran.r-project.org/web/checks/check_results_lightgbm.html

## Reproducible example

I haven't tried to reproduce this yet outside of the CRAN system.

According to CRAN's logs ([link](https://cran.r-project.org/web/checks/check_results_lightgbm.html)), exactly one test is failing.

```text
     == Failed ======================================================================
     -- 1. Failure (test_basic.R:1276:5): lgb.train() supports non-ASCII feature name
     dumped_model[["feature_names"]] not identical to iconv(feature_names, to = "UTF-8").
     4/4 mismatches
     x[1]: "F_é\u009b¶"
     y[1]: "F_<U+96F6>"
    
     x[2]: "F_äž\u0080"
     y[2]: "F_<U+4E00>"
    
     x[3]: "F_äº\u008c"
     y[3]: "F_<U+4E8C>"
    
     x[4]: "F_äž\u0089"
     y[4]: "F_<U+4E09>"
    
     == DONE ========================================================================
     Error: Test failures
     Execution halted
```

## Additional Comments

This project has a CI job (#4164) that is intended to exactly replicate CRAN's `r-devel-linux-x86_64-debian-clang` check. That job has been succeeding. I'm not sure yet what the difference is between that job and CRAN's setup.

jameslamb:
@shiyu1994 can you please check your email and let us know here if CRAN has contacted you?

The last time something similar happened (#4923), we were given just a few days to submit a patched version (#4930) before CRAN archived the package.

If we need a v3.3.3 release with a patch, it's important that we start that work as soon as possible, to avoid disruptions to users.

shiyu1994:
Sorry for ignoring the email. Will pay more attention to such emails later on.

jameslamb:
AH! @shiyu1994 thank you. This is a VERY serious problem.

It means we now only have 3 more days to upload a new version of the R package to CRAN, or `{lightgbm}` will be archived on CRAN, which would mean:

* users will not be able to install it with `install.packages()`
[...]

PR #5510, opened 2022-09-27
[ci] prefer CPython in Windows test environment and use safer approach for cleaning up network (fixes #5509)

jameslamb:
Fixes #5509.

`conda` sometimes "downgrades" Python from a CPython build to a PyPy build in our Windows CI jobs. This has historically been because of dependencies introduced by either the `python-graphviz` or `matplotlib` conda packages.

~This PR proposes trying to prevent that situation by explicitly passing flag `--no-update-deps`.~

~From the `conda` docs ([link](ttps://docs.conda.io/projects/conda/en/latest/commands/install.html))~

> ~`--freeze-installed`, `--no-update-deps`~
> ~Do not update or change already-installed dependencies.~

This PR proposes explicitly installing `python={version}[build=*cpython]` to prevent `conda` from environment solves that switch to pypy-based builds of Python.

I think this should be possible, based on https://docs.conda.io/projects/conda/en/latest/user-guide/concepts/pkg-specs.html#package-match-specifications.

### DIFF
--- a/.ci/test.sh
+++ b/.ci/test.sh
@@ -185,12 +185,12 @@ if [[ $TASK == "gpu" ]]; then
     if [[ $METHOD == "pip" ]]; then
         cd $BUILD_DIRECTORY/python-package && python setup.py sdist || exit -1
         pip install --user $BUILD_DIRECTORY/python-package/dist/lightgbm-$LGB_VER.tar.gz -v --install-option=--gpu || exit -1
-        pytest -s -vvv -Wall $BUILD_DIRECTORY/tests/python_package_test/test_dask.py || exit -1
+        pytest $BUILD_DIRECTORY/tests/python_package_test || exit -1
         exit 0
     elif [[ $METHOD == "wheel" ]]; then
         cd $BUILD_DIRECTORY/python-package && python setup.py bdist_wheel --gpu || exit -1
         pip install --user $BUILD_DIRECTORY/python-package/dist/lightgbm-$LGB_VER*.whl -v || exit -1
-        pytest -s -vvv -Wall $BUILD_DIRECTORY/tests/python_package_test/test_dask.py || exit -1
+        pytest $BUILD_DIRECTORY/tests || exit -1
         exit 0
     elif [[ $METHOD == "source" ]]; then
         cmake -DUSE_GPU=ON ..
--- a/.ci/test.sh
+++ b/.ci/test.sh
@@ -151,7 +151,7 @@ if [[ $TASK == "sdist" ]]; then
     if [[ $PRODUCES_ARTIFACTS == "true" ]]; then
         cp $BUILD_DIRECTORY/python-package/dist/lightgbm-$LGB_VER.tar.gz $BUILD_ARTIFACTSTAGINGDIRECTORY
     fi
-    pytest -s -vvv -Wall $BUILD_DIRECTORY/tests/python_package_test || exit -1
+    pytest -s -vvv -Wall $BUILD_DIRECTORY/tests/python_package_test/test_dask.py || exit -1
     exit 0
 elif [[ $TASK == "bdist" ]]; then
     if [[ $OS_NAME == "macos" ]]; then
@@ -173,7 +173,7 @@ elif [[ $TASK == "bdist" ]]; then
         fi
     fi
     pip install --user $BUILD_DIRECTORY/python-package/dist/*.whl || exit -1
-    pytest -s -vvv -Wall $BUILD_DIRECTORY/tests || exit -1
+    pytest -s -vvv -Wall $BUILD_DIRECTORY/tests/python_package_test/test_dask.py || exit -1
     exit 0
 fi
 
@@ -185,12 +185,12 @@ if [[ $TASK == "gpu" ]]; then
     if [[ $METHOD == "pip" ]]; then
         cd $BUILD_DIRECTORY/python-package && python setup.py sdist || exit -1
         pip install --user $BUILD_DIRECTORY/python-package/dist/lightgbm-$LGB_VER.tar.gz -v --install-option=--gpu || exit -1
-        pytest -s -vvv -Wall $BUILD_DIRECTORY/tests/python_package_test || exit -1
+        pytest -s -vvv -Wall $BUILD_DIRECTORY/tests/python_package_test/test_dask.py || exit -1
         exit 0
     elif [[ $METHOD == "wheel" ]]; then
         cd $BUILD_DIRECTORY/python-package && python setup.py bdist_wheel --gpu || exit -1
         pip install --user $BUILD_DIRECTORY/python-package/dist/lightgbm-$LGB_VER*.whl -v || exit -1
-        pytest -s -vvv -Wall $BUILD_DIRECTORY/tests || exit -1
+        pytest -s -vvv -Wall $BUILD_DIRECTORY/tests/python_package_test/test_dask.py || exit -1
         exit 0
     elif [[ $METHOD == "source" ]]; then
         cmake -DUSE_GPU=ON ..
@@ -213,7 +213,7 @@ elif [[ $TASK == "cuda" || $TASK == "cuda_exp" ]]; then
         else
             pip install --user $BUILD_DIRECTORY/python-package/dist/lightgbm-$LGB_VER.tar.gz -v --install-option=--cuda-exp || exit -1
         fi
-        pytest -s -vvv -Wall $BUILD_DIRECTORY/tests/python_package_test || exit -1
+        pytest -s -vvv -Wall $BUILD_DIRECTORY/tests/python_package_test/test_dask.py || exit -1
         exit 0
     elif [[ $METHOD == "wheel" ]]; then
         if [[ $TASK == "cuda" ]]; then
@@ -222,7 +222,7 @@ elif [[ $TASK == "cuda" || $TASK == "cuda_exp" ]]; then
             cd $BUILD_DIRECTORY/python-package && python setup.py bdist_wheel --cuda-exp || exit -1
         fi
         pip install --user $BUILD_DIRECTORY/python-package/dist/lightgbm-$LGB_VER*.whl -v || exit -1
-        pytest -s -vvv -Wall $BUILD_DIRECTORY/tests || exit -1
+        pytest -s -vvv -Wall $BUILD_DIRECTORY/tests/python_package_test/test_dask.py || exit -1
         exit 0
     elif [[ $METHOD == "source" ]]; then
         if [[ $TASK == "cuda" ]]; then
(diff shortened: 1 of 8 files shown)


jameslamb:
This worked! (https://ci.appveyor.com/project/guolinke/lightgbm/builds/44896165/job/xvf4hw3hv45lchyf)

Pushing another commit re-enabling all the other CI that I'd skipped just to save time, but then I think we should adopt this fix. @shiyu1994 @StrikerRUS

jmoralez:
It seems the linux jobs are getting stuck as well in this PR. I investigated a bit while working on #5505 but the environments seemed the same, it's strange that it started to happen now. Also they sometimes randomly pass.

jameslamb:
I've tried rebuilding the timed-out CUDA and Azure jobs a few times today, hoping to get lucky and have the builds not hit the Dask timeouts...so far, I haven't been successful.

I'll be traveling for the next few days, and I'm not sure how much I'll be able to work on LightGBM during that time.

@shiyu1994 @jmoralez @StrikerRUS if you're able to find a workaround for the Dask issues, it's ok with me if you want to push such fixes directly to this branch. So we can have one PR that resolves the CI issues.

shiyu1994:
Thanks. Good to see that the windows environment issue is resolved. Let's see whether we can fix the Dask issue together in this PR.

shiyu1994:
I can try to manually debug the Dask tests on our self-hosted CUDA CI agent. Hopefully we can find a solution soon.

jameslamb:
We could also try pinning all `dask` and `distributed` to a specific version, to try to isolate whether this new timeout issue is related to Dask itself or to other changes in the CI environment.

shiyu1994:
I think I've found the root of the timeout of Dask tests.

This is because `Network` class in our C++ code is defined as a `static` class, which is shared by all boosters of the same process.

However, because in our `test_dask.py`, a single `LocalCluster` is created for all the test cases, and the `LocalCluster` keeps the same subprocesses during the whole run of `test_dask.py`, without reinitializing them.

Then comes the issue. When a booster A is created in test case 1, it still exists in the memory space of the process. Then when we move to test case 2, a new booster B will be created. However, at this moment, somehow the garbage collection is triggered, and booster A is recycled. And the `__del__` of `Booster` is called.
https://github.com/microsoft/LightGBM/blob/dc4794b62f2c09103e558b7a8325bed66083a5c3/python-package/lightgbm/basic.py#L2826-L2836
which will deallocate all the network connections by calling `free_network`!
https://github.com/microsoft/LightGBM/blob/dc4794b62f2c09103e558b7a8325bed66083a5c3/python-package/lightgbm/basic.py#L2919-L2929
https://github.com/microsoft/LightGBM/blob/dc4794b62f2c09103e558b7a8325bed66083a5c3/src/c_api.cpp#L2511-L2515
Then booster A will train as a single process program since the `num_machines_` is set to `1`.
https://github.com/microsoft/LightGBM/blob/dc4794b62f2c09103e558b7a8325bed66083a5c3/src/network/network.cpp#L60-L66
However, since we are training in a distributed way, there's another process waiting for the response of the process running booster A. And that process gets stuck forever since there's no response from booster A's process.

I've pushed a quick workaround to this branch, which enforces using a new `LocalCluster` for each test case. However, this seems to make the testing much slower because initializing the processes for a cluster in Dask seems to be slow.

Could anybody familiar with Dask provide a better solution? Shortly speaking, we want to use new processes for distributed training in each test case.

jameslamb:
Wow, great investigation!!! I wonder if this is the cause of reports like #4771 or #4942.

I think creating a new Dask cluster every time is not a good solution...LightGBM's users wouldn't be happy with having to do that, since it would mean that, for example, the workflow of "initialize a Dask DataFrame, keep it in distributed memory with `.persist()`, then run multiple LightGBM training runs over it" would effectively not be supported.

Instead, could we have each distributed training process allocate its own `Network` and hold a pointer to it, and then have `free_network()` only free **that specific `Network`**? Or would that cause some problems?

jmoralez:
I agree with @jameslamb on this, users may want to run consecutive trainings in the same process when doing things like hyperparameter tuning. Also this has been the way we run the dask tests for more than a year (changed in #4159), @shiyu1994 can you think of a recent change that would cause this to start failing now?

jmoralez:
I tested on mac and in [one of the jobs](https://github.com/microsoft/LightGBM/actions/runs/3159873213/jobs/5143641670) there are many errors that we've previously seen for mac: `lightgbm.basic.LightGBMError: Socket recv error, Connection reset by peer (code: 54)`. I guess it has the same root cause, right? The network was freed while training so the connection was lost to one of the machines but mac somehow realizes this and raises the error instead?

jameslamb:
I'm going to try implementing a fix here where, at the end of training, the Python package explicitly calls `free_network()`.

I still think the ideal solution is the one described in https://github.com/microsoft/LightGBM/pull/5510#issuecomment-1263570805, because the Dask-only fix I'm proposing has the following drawbacks:

* doesn't help solve this issue for non-Dask settings (e.g. `SynapseML` or `lightgbm-ray`)
* means that only one LightGBM training run can be happening in the cluster at a time

If my proposed fix seems to work, I think we should adopt it to unblock CI and then I'll document this issue in a separate issue that could be worked on.

jameslamb:
I pushed a partial fix in https://github.com/microsoft/LightGBM/pull/5510/commits/da2b17393f3e0e1ef6f7f0592fd4a4fb2f19f797, but realize now that I misunderstood @shiyu1994 's explanation in https://github.com/microsoft/LightGBM/pull/5510#issuecomment-1263525112.

I thought that the issue was that the `Booster` from one training run might be garbage-cleaned by Python *while another training run was starting*, and tried to fix that by replacing a direct call to `_LIB.LGBM_NetworkFree()` with a call to `Booster.free_network()`, which will set `Booster.network = False`, avoiding another call of `LGBM_NetworkFree()` when the `Booster`'s `__del__()` method is called.

But I understand now that the problem occurs *during one training run*, while the process on rank 0 is waiting to finish training.

I think another option could be to simply never call `free_network()` from the Dask code, and just rely on the `Booster`'s `__del__()` method to do that.

Pushed a new commit, https://github.com/microsoft/LightGBM/pull/5510/commits/c37bb196a8969874ee868d0aaaad295df630760e, trying that.

jameslamb:
I just pushed https://github.com/microsoft/LightGBM/pull/5510/commits/d648a15f74643579279b6e1e151bea332054fb89 adding more logging, to try to see if we can get any other hints.

jameslamb:
Getting closer, and I definitely see some evidence of what @shiyu1994 is talking about.

On the current state of this branch, I've added the following changes to try to get through the tests:

* reduced `timeout` passed through LightGBM's params from `120` to `1`
* explicitly added a full garbage collection (`gc.collect()`) at the beginning of each `_train_part()` call, to force garbage collection of the old `Booster`s from previous tests prior to beginning new training
* switched from `_LIB.LGBM_NetworkFree()` to `model.booster_.free_network()` in `lightgbm.dask._train_part()`, so that when the `Booster` is eventually garbage cleaned, `_LIB.LGBM_NetworkFree()` won't actually run again (since it'll have property `self.network = False`

In that state, the jobs are still timing out, but 291 of the 368 Dask tests are able to complete in that time ([logs link](https://dev.azure.com/lightgbm-ci/8461a79b-5dce-4085-ad70-4410b7135276/_apis/build/builds/13667/logs/310).

I'll keep investigating this as I have time this week. It's my top priority in open source work right now, just trying to find the time to investigate things.

Now that I'm back from traveling, I am going to keep using this branch to debug, so @shiyu1994 @guolinke @jmoralez @StrikerRUS please open your own separate branch if you want to use LightGBM's CI to test your own theories.

jmoralez:
Nice investigation! Seems like it's now getting stuck at the test that calls `Client.restart`.

jameslamb:
Thanks! Yeah the timings in the logs suggested that only *certain* tests were timing out or taking a very very long time, so now the approach I'm taking is to try skipping those one-by-one until the CI jobs all succeed.

Given that, we might be able to learn something from the things that the skipped tests have in common. We could also then consider *temporarily* leaving those skips in, to unblock other work in the project while we investigate this specific issue.

jameslamb:
Ok, so! It seems like commenting out 2 of the `lightgbm.dask` tests + the other fixes I've pushed allows the jobs to succeed again!

([build link](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=13671&view=logs&j=fb919173-48b9-522d-5342-9a59a13eb10b&t=7fcbeee5-d4c1-571c-8072-1d5874591254))

* `test_training_succeeds_even_if_some_workers_do_not_have_any_data()`
* `test_machines_should_be_used_if_provided()`

Next, I'll try additional commits that remove some of the workarounds that I've introduced and see if that remains true (to narrow down the possible opportunities to fix this).

After that, I'll look more closely into what might be similar across those two tests that could result in this timeout problem.

----

By the way, this high-level documentation from `distributed` on an object's lifecycle in a Dask cluster seems highly relevant: https://distributed.dask.org/en/stable/journey.html#step-8-garbage-collection

jameslamb:
Ok, seems that **the changes related to `time_out` parameter had no effect on this issue** ([link to successful build](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=13673&view=results)).

So I'll leave all of those reverted.

Next, I'll try removing the explicit garbage collection and see if that matters.

jameslamb:
I know it is difficult to follow all the debugging comments and commits here, but CI is passing and this PR is ready for review 🎉 

After switching from `LIB.LGBM_NetworkFree()` to `Booster.free_network()` in the Dask module (to avoid a second call to `_LIB.LGBM_NetworkFree()` at an unpredictable time when the `Booster` is garbage cleaned), only one test is still causing timeout issues.

I think we should

* merge this PR (which adds a `pytest.skip()` on that test) to unblock development
* separately document the work to investigate and remove that `pytest.skip()`

@shiyu1994 @guolinke @jmoralez @StrikerRUS

jmoralez (approved):
Awesome job! And thanks @shiyu1994 for the insights.

jameslamb:
Thanks to both of you for your help! This was a really difficult one.

It would be interesting in the future to work on changing the strategy for how `LGBM_NetworkFree()` works, to try to enable multiple concurrent training runs on the same Dask cluster.

I'm going to merge this and start updating / merging some of the other approved PRs, starting with #5506.

shiyu1994:
Sorry for the late response. Just return from our 1-week national holiday.

jameslamb:
no problem, welcome back! Please look at my comment in https://github.com/microsoft/LightGBM/issues/5502#issuecomment-1256837200 as soon as possible and respond there...I'm nervous that the R package might be in danger of archiving on CRAN.

PR #5525, opened 2022-10-07
WIP: [DO NOT MERGE] Release v3.3.3

jameslamb:
Similar to #4930, this is an unusual release. It is `v3.3.3`, but doesn't include all of the changes made since `v3.3.2`.

Instead, this PR just contains the changes on top of `v3.3.2` needed to prevent the R package from being archived by CRAN.

**THIS PR SHOULD NEVER BE MERGED**

<hr>

### Setup

I created a new branch, `release/v3.3.2`, which contains only:

* LightGBM's source code as of tag `v3.3.2`
* CI configs and build scripts as of latest `master` (https://github.com/microsoft/LightGBM/commit/82b7d31c783a93cc96cf7fea24098b866182762f)

<details><summary>how I did that (click me)</summary>

```shell
# refresh local copy of tags
git fetch upstream --tags

# create a new branch with the v3.3.2 changes
git checkout v3.3.2
git checkout -b release/v3.3.2

# reset local state of submodules
git submodule update --recursive

# add the CI configs and build scripts from master onto that branch
git checkout master -- \
    .appveyor.yml \
    build-cran-package.sh \
    build_r.R \
    .ci/*.ps1 \
    .ci/*.py \
    .ci/lint_r_code.R \
    .ci/append_comment.sh \
    .ci/rerun_workflow.sh \
    .ci/set_commit_status.sh \
    .ci/setup.sh \
    .ci/test.sh \
    .ci/test_r_package.sh \
    .ci/test_r_package_valgrind.sh \
    .ci/trigger_dispatch_run.sh \
    .github/workflows/cuda.yml \
    .github/workflows/optional_checks.yml \
    .github/workflows/r_configure.yml \
    .github/workflows/r_package.yml \
    .github/workflows/r_valgrind.yml \
    .github/workflows/static_analysis.yml \
    .vsts-ci.yml

git commit -m "add CI configs and build scripts from master"
git push upstream release/v3.3.2
```

</details>

Tags and artifacts should be made directly from this `release_v3.3.3` branch instead of `master`.

### Release checklist:

Copied from #4930.

- [x] Update [VERSION.txt](https://github.com/microsoft/LightGBM/blob/master/VERSION.txt) number.
- [x] Update version in [Appveyor config file](https://github.com/microsoft/LightGBM/blob/master/.appveyor.yml).
- [x] Update version in [`configure` file of R-package](https://github.com/microsoft/LightGBM/blob/master/R-package/configure): `/gha run r-configure`.
[...]

PR #5527, opened 2022-10-08
[ci] run CI on pull requests targeting release/ branches

jameslamb:
Based on @guolinke 's suggestion in https://github.com/microsoft/LightGBM/pull/5525#issuecomment-1272332768, I'd like to try targeting PR #5525 at a branch, `release/v3.3.2`, which only contains:

* v3.3.2 of LightGBM
* CI scripts and configs from `master`

That would allow us to try to run all of LightGBM's CI over the potential `v3.3.3` release (we'll see if that works!).

This PR proposes modifying the configs for GitHub Actions and Azure Devops so that CI jobs will be triggered on all ~branches~ PRs targeting branches with a name like `release/*`.

Once this is merged, I'll update #5525.

### Notes for reviews

Documentation for how these configs work:

* GitHub Actions: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions
* Azure: https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/pr?view=azure-pipelines#pr-branchfilter-list

It isn't necessary to modify the Appveyor config....I believe it just builds for all branches automatically.

image

### DIFF
--- a/.github/workflows/cuda.yml
+++ b/.github/workflows/cuda.yml
@@ -7,6 +7,7 @@ on:
   pull_request:
     branches:
     - master
+    - release/*
 
 env:
   github_actions: 'true'
--- a/.github/workflows/optional_checks.yml
+++ b/.github/workflows/optional_checks.yml
@@ -4,6 +4,7 @@ on:
   pull_request:
     branches:
       - master
+      - release/*
 
 jobs:
   all-successful:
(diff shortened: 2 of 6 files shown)


jmoralez (approved):

jameslamb:
thanks so much for the quick review, @jmoralez !

Issue #5514, opened 2022-09-28
Use manylinux image to produce wheels


jameslamb:
## Summary

Created from https://github.com/microsoft/LightGBM/issues/5390#issuecomment-1204486025.

The following changes should be made in LightGBM:

* move from supporting `GLIBC >= 2.14` to `GLIBC >= 2.23`
* start producing Python wheels using one of the `manylinux` images (https://github.com/pypa/manylinux)
    - this implies ending official support for Ubuntu 14.04 in LightGBM

## Motivation

It has been really difficult to maintain support for older operating systems and versions of GLIBC.

References:

* https://github.com/microsoft/LightGBM/issues/5390#issuecomment-1203495465
* https://github.com/microsoft/LightGBM/pull/4893#issuecomment-1039204806
* https://github.com/microsoft/LightGBM/pull/4893#issuecomment-1039812782

Moving to newer base image and higher floor for GLIBC and supported Ubuntu version should help to alleviate some maintenance burden.

## Description

As described in https://github.com/microsoft/LightGBM/issues/5390#issuecomment-1204486025, LightGBM should upgrade to a `manylinux` base image supporting at least `GLIBC==2.23` (and not too much higher) for producing wheels.

The `manylinux` README shows the following compatibility for wheels produced on those images (https://github.com/pypa/manylinux).

<img width="872" alt="image" src="https://user-images.githubusercontent.com/7608904/192678826-17118fd3-ce7c-4e44-ab22-823555c11421.png">

https://ubuntu.com/about/release-cycle describes the release cycle for versions of Ubuntu

<img width="792" alt="image" src="https://user-images.githubusercontent.com/7608904/192678915-6b2dd392-fc74-47a7-8ae0-94f40fb4d643.png">

Using an approach like the following

```shell
docker run \
    --rm \
    --entrypoint="" \
    -it quay.io/pypa/manylinux1_x86_64 \
    ldd --version
```

I found that the GLIBC version for the published `manylinux` images is as follows.

| image                    | GLIBC |
|:-----------------:|:------:|
|manylinux1            | 2.5      |
|manylinux2010     | 2.12    |
|manylinux2014     | 2.17    |
|manylinux_2_28   | 2.28   |

## References

Other related conversations:

* discussion from 2020 about possibly using `manylinux_2010`: https://github.com/microsoft/LightGBM/pull/3421#issuecomment-711559332
[...]

PR #5544, opened 2022-10-16
[tests][dask] fix workers without data test (fixes #5537)

### LINKED ISSUE #5537: [ci] [dask] fix test on "only one worker has data" training
## Description

In #5510, we observed that the Dask tests were getting stuck and timing out after 60 minutes.

Narrowed that down to a single unit test, which is currently skipped.

https://github.com/microsoft/LightGBM/blob/0c0eb2a6a1ba111f0896dd16a579342fa16f7629/tests/python_package_test/test_dask.py#L1506-L1507

The testing work from #5510 should be continued.... that test should be re-enabled and its underlying cause should be identified and fixed.

## Reproducible example

Uncomment the test linked above and push a commit to a pull request in LightGBM.

## Environment info

Latest version of LightGBM (https://github.com/microsoft/LightGBM/commit/0c0eb2a6a1ba111f0896dd16a579342fa16f7629).

## Additional Comments

Especially see this comment describing a possible root cause of timeout issues: https://github.com/microsoft/LightGBM/pull/5510#issuecomment-1263525112.

jmoralez:
Changes `tests/python_package_test/test_dask.py::test_training_succeeds_even_if_some_workers_do_not_have_any_data` to have three workers and data in only two of them.

Fixes #5537.


jameslamb (changes requested):
Thanks for working on this!

Before I review, could you please explain, either in the description or in a comment here, what the root cause of the timeouts observed in #5537 were and *how* this fixes it?

The diff has a lot more changes in it than I'd expected from the description "have three workers and data in only two of them". For example, it's not obvious to me how that type of change would require bringing in new imports from `sklearn.metrics`.

I'm especially interested in the change where we're now passing an explicit list of `workers=` to `client.persist()`. Was the root problem just that "`collection_to_single_partition()` was resulting in, for example, `dX` being all on the first worker and `dy` being all on the second worker"? If so, couldn't the same effect be accomplished with the existing 2-worker cluster?

jmoralez:
> what the root cause of the timeouts observed in https://github.com/microsoft/LightGBM/issues/5537 were

I couldn't reproduce the timeouts locally.

> how this fixes it

The purpose of the test was to check that in a distributed training setting, having a worker without data wouldn't cause a problem, that's why I changed it to having three workers and only two with data, so distributed training is performed (because of the two workers having data), whereas previously it was probably falling back to just local training, and the definition of "succeeds" is that we achieve a good metric value (hence the `sklearn.metrics` import).

jameslamb (approved):
> previously it was probably falling back to just local training

OHHHHHHHHHHHH. Thank you so much for that explanation. Makes sense to me!

Let's merge this then 😁

jameslamb:
Will merge this whenever we get CI working again 😭 

Hopefully if/when #5514 is merged.

jameslamb:
thanks so much for fixing this @jmoralez !

PR #5555, opened 2022-10-23
[ci] prefer CPython in Linux and macOS test environment

StrikerRUS:
Similarly to #5510, explicitly ask conda to install CPython, not PyPy.


jameslamb (approved):
Excellent, thank you for doing this!

PR #5559, opened 2022-10-29
[python-package] remove unused imports

jameslamb:
Removes some unused imports.

Found with `flake8`:

```text
python-package/lightgbm/dask.py:20:1: F401 '.basic._safe_call' imported but unused
python-package/lightgbm/dask.py:24:1: F401 '.sklearn._LGBM_ScikitCustomEvalFunction' imported but unused
python-package/lightgbm/sklearn.py:6:1: F401 'typing.Iterable' imported but unused
```


jmoralez (approved):
Do you think we should add flake8 to the lint task?

jameslamb:
> Do you think we should add flake8 to the lint task?

I do! I've found it really valuable in the projects I've used it on. I opened #5566 to discuss it.

PR #5565, opened 2022-11-02
[docs] add postgresml to list of external repos

jameslamb:
Tonight I learned about `postgresml`, a project that allows you to train LightGBM models (and many other types of machine learning models) using SQL. `postgresml`  available as Postgres extension.

https://github.com/postgresml/postgresml

Here's the LightGBM binding: https://github.com/postgresml/postgresml/blob/e39d7e36ac285bb27496896bff33c87054e0d90b/pgml-extension/src/bindings/lightgbm.rs

cc `postgresml` maintainers: @montanalow @levkk


montanalow:
Nice! Let me know if there's anything we can do to help. I think the Rust bindings for LightGBM could use a little more love in general, but the MVP is looking good so far.

PR #5580, opened 2022-11-09
[ci] switch to manylinux_2_28 for Linux artifacts (fixes #5514, fixes #5589)

jameslamb:
Fixes #5514.
Fixes #5589.
Contributes to #5390.
Contributes to #5391 (via https://github.com/microsoft/LightGBM/pull/5396#issuecomment-1203168124).

Proposes switching from an Ubuntu 14.04 image to one based on `pypa/manylinux_2_28` for building artifacts on Linux including:

* Python source and built distributions (e.g. wheels)
* `lightgbmlib.jar` (Java interface to LightGBM)
* `lib_lightgbm.so` (C++ shared library)

And increases the floors on symbol versions for key linked libraries on Linux.

* `GLIBC`: `<= 2.14` ([October 2011](https://sourceware.org/glibc/wiki/Glibc%20Timeline)) to `<= 2.28` ([August 2018](https://sourceware.org/glibc/wiki/Glibc%20Timeline))
* `GLIBCXX`: `<= 3.4.19` to `<= 3.4.22`
* `GOMP`: `== 1.0` ([January 2006](https://github.com/gcc-mirror/gcc/blame/0cbb756fe9c8e13a743bdcf599481b31bd01ddbc/libgomp/libgomp.map#L243)) to `<= 4.5` ([October 2015](https://github.com/gcc-mirror/gcc/blame/0cbb756fe9c8e13a743bdcf599481b31bd01ddbc/libgomp/libgomp.map#L343))

### Notes for Reviewers

This PR uses images build from https://github.com/guolinke/lightgbm-ci-docker/pull/27. That `lightgbm-ci-docker` PR should be reviewed and merged first.

### Why all these `if-else` forks in the CI scripts?

`pypa/manylinux_2_28` is based on Alma Linux, so it used `dnf` and `yum` as package managers (where Ubuntu uses `apt`).

Issue #4998, opened 2022-02-08
version `GLIBCXX_3.4.18' not found


samoidev:
## Description
I'm trying to install and use lightgbm package in python, but it seems like there is a version conflict related to gcc.

After installing the package, using pip install lightgbm, when trying to import it, the following error appear:

```
>>>import lightgbm

OSError: /usr/company/pkgs/gcc/4.7.2/lib64/libstdc++.so.6: version `GLIBCXX_3.4.18' not found (required by /nfs/user/.venv_notebook/lib/python3.7/site-packages/lightgbm/lib_lightgbm.so
```

Although I have other versions of GCC installed, I'm not sure how to make it use the appropriate version it needs.

```
ls /usr/company/pkgs/gcc/

10.1.0  10.3.0  11.1.0  4.7.2  4.8.1  4.9.2  5.3.0  6.2.0  6.3.0b  6.4.0  6.5.0  7.2.0  8.2.0  8.3.0  9.1.0  9.2.0
```

## Environment info

LightGBM version or commit hash: 3.3.2

Command(s) you used to install LightGBM : 

```shell
pip install lightgbm
```

Linux machine (SUSE)

guolinke:
it looks like a environment probelm. Following link may help
https://stackoverflow.com/questions/19386651/how-to-fix-usr-lib-libstdc-so-6-version-glibcxx-3-4-15-not-found

PR #5582, opened 2022-11-10
[ci] add docs and specific error messages in check_dynamic_dependencies.py

jameslamb:
Contributes to #5514.

Each release of LightGBM includes a `lib_lightgbm.so`... the C++ shared library for LightGBM, intended for use on Linux systems.

Because that library is intended to be usable on a wide range of operating systems and compiler / linker versions, it's important to be careful that no too-new libraries were linked into that library at compile time. To avoid issues similar to #1945, #4998, and #5106.

LightGBM's CI contains a check to catch this situation for libraries that use [symbol versioning](https://gcc.gnu.org/wiki/SymbolVersioning)... OpenMP, libstdc++, and glibc.

https://github.com/microsoft/LightGBM/blob/f1d3181ced9fd01f4b2899054abd99be6773e939/.ci/test.sh#L262-L263

While working on a major update to the base image used to generate `lib_lightgbm.so` (#5580), I needed to update the code for that check, and saw a few things I wanted to change.

This PR proposes the following changes to `check_dynamic_dependencis.py`:

* expands docstring to explain why this script is important and how to find the release dates corresponding to symbol versions for the relevant libraries
* changes function name, `check_dependicies()` --> `check_dependencies()`
* adds more informative error messages to assertions

### Notes for Reviewers

This CI check was introduced in #1812.


helpers/check_dynamic_dependencies.py
```diff
@@ -2,13 +2,22 @@
 """Helper script for checking versions in the dynamic symbol table.
 
 This script checks that LightGBM library is linked to the appropriate symbol versions.
+
+Linking to newer symbol versions at compile time is problematic because it could result
+in built artifacts being unusable on older platforms that LightGBM users are on.
```
jameslamb, line 7:
```suggestion
in built artifacts being unusable on older platforms.
```

jmoralez (approved):

PR #5590, opened 2022-11-16
[ci] update versions of GitHub Actions

jameslamb:
Contributes to #3763.

Proposes updating all external Github Actions that LightGBM uses to their latest versions.

### Benefit of this change

Attempts to resolve the following deprecation warnings, before they become errors and result in LightGBM's CI failing.

> Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions/checkout@v2.4.0, r-lib/actions/setup-pandoc@v1

> r-lib/actions/setup-pandoc@v1 is deprecated. Please update your workflow to use the 'v2' version.

> The `save-state` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


jmoralez (approved):
Please let me know if you need help with the conflict (e.g. you don't have access to a computer)

jameslamb:
Appreciate the offer, thanks @jmoralez ! I can fix it right now, I haven't started traveling yet.

PR #5591, opened 2022-11-17
nrounds does not work as an alias for num_iterations

Kornel:
## Description
Setting the paremter `nrounds` for a `LGBMRegressor` does not seem to have any effect.

## Reproducible example
```python
import lightgbm

X, y = make_regression()
X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.2)

model = lightgbm.LGBMRegressor(max_depth=1, learning_rate=0.0001, nrounds=10)
model =model.fit(X_train, y_train, eval_set=[(X_test, y_test)])

model.booster_.num_trees()
```
## Environment info

LightGBM version: 3.3.3

Env setup as in this example https://github.com/Kornel/lightgbm-nrounds-is-not-supported

```shell
poetry install
```

### DIFF
--- a/docs/Parameters.rst
+++ b/docs/Parameters.rst
@@ -153,7 +153,7 @@ Core Parameters
 
    -  **Note**: can be used only in CLI version
 
--  ``num_iterations`` :raw-html:`<a id="num_iterations" title="Permalink to this parameter" href="#num_iterations">&#x1F517;&#xFE0E;</a>`, default = ``100``, type = int, aliases: ``num_iteration``, ``n_iter``, ``num_tree``, ``num_trees``, ``num_round``, ``num_rounds``, ``nrounds``, ``num_boost_round``, ``n_estimators``, ``max_iter``, constraints: ``num_iterations >= 0``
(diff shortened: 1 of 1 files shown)


Trackatbar:
hi，what is your sklearn version？

Kornel:
scikit learn is 1.1.3, you can see all dependencies in the `poetry.lock` file https://github.com/Kornel/lightgbm-nrounds-is-not-supported/blob/master/poetry.lock

jmoralez:
Thanks for raising this @Kornel, indeed it doesn't seem to work on previous versions, but it works on the current master:
```
lightgbm/engine.py:143: UserWarning: Found `nrounds` in params. Will use it instead of argument
  _log_warning(f"Found `{alias}` in params. Will use it instead of argument")
[LightGBM] [Warning] num_iterations is set=10, nrounds=10 will be ignored. Current value: num_iterations=10
```
so I don't think we need to remove it from the docs.

Trackatbar:
Okay！I get this! Thank you~




------------------&nbsp;原始邮件&nbsp;------------------
发件人:                                                                                                                        "microsoft/LightGBM"                                                                                    ***@***.***&gt;;
发送时间:&nbsp;2022年11月17日(星期四) 晚上11:36
***@***.***&gt;;
***@***.******@***.***&gt;;
主题:&nbsp;Re: [microsoft/LightGBM] nrounds does not work as an alias for num_iterations (PR #5591)





 
scikit learn is 1.1.3, you can see all dependencies in the poetry.lock file https://github.com/Kornel/lightgbm-nrounds-is-not-supported/blob/master/poetry.lock

PR #5605, opened 2022-11-25
[CUDA] Add quantile regression objective for new CUDA version

shiyu1994:
Add quantile regression objective for new CUDA version.


shiyu1994:
It is weird that 
```
nvlink error   : Entry function '_ZN8LightGBM33CUDAConstructHistogramDenseKernelIhfLm11120EEEvPKNS_20CUDALeafSplitsStructEPKfS5_PKT_PKjSA_PKii' uses too much shared data (0xcbfc bytes, 0xc000 max) (target: sm_60)
nvlink error   : Entry function '_ZN8LightGBM34CUDAConstructHistogramSparseKernelItmfLm11120EEEvPKNS_20CUDALeafSplitsStructEPKfS5_PKT_PKT0_SB_PKji' uses too much shared data (0xcbfc bytes, 0xc000 max) (target: sm_60)
```
should arise when no related code of `CUDAConstructHistogramSparseKernel` or `CUDAConstructHistogramDenseKernel` is changed in this PR for cuda 10.0.

Note that for cuda 10.0, we've already used a smaller shared memory size than the claimed `0c000` bytes (with which we should be able to use `DP_SHARED_HIST_SIZE = 6144`).
https://github.com/microsoft/LightGBM/blob/881106311557187a1fea2ec90c234f71771a32d8/include/LightGBM/cuda/cuda_row_data.hpp#L22-L27

It is unclear to me why adding code in this PR will have to make the `DP_SHARED_HIST_SIZE  = 5560` for cuda 10.0 smaller.

@guolinke Do you have any idea?

guolinke:
how many bytes were exceeded? It looks quite large. are there any functions that possibly inherit or call other functions with __shared__ memory?

shiyu1994:
> how many bytes were exceeded?

0xcbfc - 0xc000 = 0x0bfc = 3068 bytes

> are there any functions that possibly inherit or call other functions with shared memory?

No. The code compiles with cuda 11.0 but not with cuda 10.0.

shiyu1994:
@guolinke Let me adjust the `#define DP_SHARED_HIST_SIZE (5560)` to a lower value so that the compilation can succeed. And let's drop the support for CUDA 10.0 in another PR.

shiyu1994:
> It is unclear to me why adding code in this PR will have to make the DP_SHARED_HIST_SIZE  = 5560 for cuda 10.0 smaller.

Just use binary search to find out that the maximum allowed value is 5176.

shiyu1994:
@guolinke This is ready for review.

guolinke (approved):

PR #5611, opened 2022-11-30
[CUDA] Add rmse metric for new CUDA version

shiyu1994:
Add rmse metric for new CUDA version.


shiyu1994:
@guolinke This is ready.

PR #5581, opened 2022-11-10
[tests][python-package] remove remaining tests using load_boston (fixes #4793)

jmoralez:
Fixes #4793.

PR #5288, opened 2022-06-13
[ci] Use Ubuntu 22.04 as `ubuntu-latest` at CI (fixes #5186)

StrikerRUS:
Fixes #5186.

Depends on https://github.com/actions/virtual-environments/issues/5490 when `ubuntu-22.04` will become the default Ubuntu on GitHub Actions.

PR #5612, opened 2022-11-30
[python-package] replace .values usage with .to_numpy()

superlaut:
For both pandas.DataFrame and pandas.Series objects, pandas documentation recommends using `to_numpy() `instead of `.values`.

Especially for line 605 the change is impactful. Calling `to_numpy(target_dtype, copy=False)` instead of `astype(target_dtype, copy=False).values` can be more than two times as fast and uses half the memory. The latter is essential for the memory efficiency of the forced conversion of a pandas.DataFrame to a numpy array of dtype `float32` or `float64`. I ran into this problem when trying to train on very large datasets with limited memory.


superlaut:
Screenshot 2022-11-30 at 11 55 02

To demonstrate my point relating to time and memory.

`na_value` in `.to_numpy()` is an optional argument (and not included in this PR). I believe I do not follow your linked comment entirely, does na_value need to be specified? According to pandas documentation, the default value depends on dtype and the dtypes of the DataFrame columns. How is this handling things differently than .astype combined with .values?

jmoralez:
It needs to be specified because we now support the pandas nullable data types, for example `Float64`, and if you have that type calling `to_numpy` without the `na_value` argument raises an error as it's currently doing in the checks.
```python
import numpy as np
import pandas as pd

s = pd.Series([1, np.nan, 3], dtype='Float64')
s
# 0     1.0
# 1    <NA>
# 2     3.0
# dtype: Float64

# this is the current approach that turns the <NA> into np.nan, we then take the values
s.astype(np.float64)
# 0    1.0
# 1    NaN
# 2    3.0
# dtype: float64

# this fails because of the missing na_value argument
s.to_numpy(dtype=np.float64)
# ValueError: cannot convert to '<class 'numpy.float64'>'-dtype NumPy array with missing values. Specify an appropriate 'na_value' for this dtype.

# this works
s.to_numpy(dtype=np.float64, na_value=np.nan)
# array([ 1., nan,  3.])
```

superlaut:
Makes sense. Could we somehow add a control flow that runs `.to_numpy` with na_value specified in case pandas version is >= 1.1.0?

jameslamb:
> we could have like a try/except so that people with newer versions of pandas can benefit from it, WDYT @jameslamb?

Good idea @jmoralez , I support that!

superlaut:
@jmoralez I changed the code to work with try-except clauses as you proposed. There were three .values calls that I left untouched, as the combination of astype and .values only seems to impact memory and time.

jmoralez:
Can you change the exception type? It's a `ValueError`. Also I see some checks are failing because it will always make a copy when passing the na_value argument: https://github.com/pandas-dev/pandas/blob/e5fb655a74c828eb1bef026a54ba9be85de1f3e6/pandas/core/internals/managers.py#L1692 and we have a test https://github.com/microsoft/LightGBM/blob/90047b96686c4e18ee06934ad097c738d66e251f/tests/python_package_test/test_basic.py#L696-L704 to check that it won't. I think we can delete that test because indeed the memory usage is lower. Here's a test I performed:
<details>
<summary>Click to expand</summary>

```python
import numpy as np
import pandas as pd

N = 10000000
df = pd.DataFrame({
    'x1': np.random.rand(N),
    'x2': np.random.rand(N),
    'x3': np.random.rand(N),
    'x4': np.random.rand(N),
    'y': np.random.randint(N),
})
df['x1'] = df['x1'].astype('Float64')
df['y'] = df['y'].astype('Int32')
# current
df2 = df.astype(np.float64).values
# proposal
df2 = df.to_numpy(dtype=np.float64, na_value=np.nan)
```
</details>

with: `scalene --memory --cli mem_test.py` and I get:

Screenshot from 2022-12-01 20-48-01

superlaut:
Hi @jmoralez , I am afraid I cannot provide advice on whether to include these tests or not. I do not know the history of these tests and implications of removing them. I would say that is on your team. Indeed the memory efficiency holds for my tests too. I believe the contribution is important. 

Could you confirm on your local machine that it is a `ValueError` and not a `TypeError`?

I get

jmoralez:
The nullable data types that use `pd.NA` as their missing value ([Int64Dtype for example](https://pandas.pydata.org/pandas-docs/version/1.0/reference/api/pandas.Int64Dtype.html#pandas.Int64Dtype)) were introduced in pandas 1.0 along with the `na_value` argument of [pd.Series.to_numpy](https://pandas.pydata.org/pandas-docs/version/1.0/reference/api/pandas.Series.to_numpy.html), however for dataframe it got there in [1.1](https://pandas.pydata.org/pandas-docs/version/1.1/reference/api/pandas.DataFrame.to_numpy.html).

So I think we could have something like the following:

```python
try:
    # most common case (no nullable dtypes)
    data = data.to_numpy(dtype=target_dtype, copy=False)
except TypeError:
    # 1.0 <= pd version < 1.1 and nullable dtypes
    # raises error because array is casted to type(pd.NA) and there's no na_value argument
    # least common case
    data = data.astype(target_dtype).values
except ValueError:
    # data has nullable dtypes, but we can specify na_value argument and copy will be made
    data = data.to_numpy(dtype=target_dtype, na_value=np.nan)
```
This would probably be the most efficient, although we need to make sure we're considering all cases.

superlaut:
I will perform some experiments tonight and get back to you.

superlaut:
Hi Jose, I will not have any time until Monday to perform experiments. Will get back then!

superlaut:
I ran experiments and agree your function is the best way to go @jmoralez .
I have made a commit with your new code. I thought I found some other way to write the code for the label, but as the label can also be a pandas DataFrame (and not just a pandas Series) there was nothing to be achieved there.

superlaut:
@jmoralez all errors I see for the tests are ImportErrors... Anything I can do to help?

jameslamb:
@superlaut please just merge latest `master` into this branch.

If you search the error messages in this repo's issues / PRs, you'll find #5634 and #5581 and #4793.

`scikit-learn` 1.2, which was released a few days ago ([link](https://pypi.org/project/scikit-learn/#history)), removed `scikit_learn.datasets.load_boston()`.

#5581 removed all of LightGBM's use of that function in this project's tests.

jmoralez:
@superlaut can you restore the test? It should pass with the latest proposal.

superlaut:
@microsoft-github-policy-service agree [company="{Amazon Web Services}"]

superlaut:
@jmoralez only R-tests are failing now, probably not related to the scope of this code.

jmoralez (approved):
Thanks for the fix! We now go through the memory intensive path only when it's absolutely necessary. This also adds support for nullable data types in the label. I'd like @jameslamb to take a look at this as well, and we'll work on fixing the CI in the following days.

jameslamb:
I'll review this next week. In the interim, @superlaut please merge the latest `master` into this branch, to get the fixes from #5288 . Hopefully that will fix the CI here.

superlaut:
@jameslamb let me know if you need any input from my side :)!

jameslamb (approved):
I'm good with this! Thanks very much for the contribution.

-----

> pandas documentation recommends using to_numpy() instead of .values.

@superlaut in the future when you contribute here, please support claims like this with links. I think you meant this? https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.values.html

image

jameslamb:
I've updated this with the latest changes on `master`, and will merge it if/when it builds successfully.

Thanks again for the help @superlaut , we hope you'll come back and contribute more in the future!

And thanks for the thorough review @jmoralez . Really nice use of `scalene` 🤩

superlaut:
Thanks @jameslamb and @jmoralez , it was cool to learn how to approach a PR to such a large scale open source project, learned many lessons. And thanks for your guidance along the way :)

PR #5613, opened 2022-11-30
[python-package] prefix basic.convert_from_sliced_object with _

OmarManzoor:
Contributes to #5313


jameslamb (approved):
Thanks very much for the contribution!

Issue #5061, opened 2022-03-09
[RFC] [python] 'setup.py install' is deprecated


jameslamb:
## Description

Running `python setup.py install` for the Python package raises the following deprecation warning:

> /Users/jlamb/mambaforge/lib/python3.9/site-packages/setuptools/command/install.py:34: SetuptoolsDeprecationWarning: setup.py install is deprecated. Use build and pip and other standards-based tools.
  warnings.warn(

I found a very detailed write-up explaining this issue: ["Why you shouldn't invoke setup.py directly" (October 2021)](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html).

Here are some of the main things we need to know:

> ...as of the last few years **all direct invocations of setup.py are effectively deprecated** in favor of invocations via purpose-built and/or standards-based CLI tools like [pip](https://pip.pypa.io/en/stable/), [build](https://pypa-build.readthedocs.io/en/stable/) and [tox](https://tox.wiki/en/latest/).

> As mentioned in [this issue](https://github.com/pypa/setuptools/issues/2088), direct setup.py invocations have effectively been unmaintained for several years now. Whenever someone raises an issue with a reproducer that involves invoking setup.py, the maintainers ask for a reproducer that doesn't hit this code path, and if one can't be found the issue is closed. Put another way, direct invocations of setup.py that currently work for you do so essentially by chance — if something breaks, you are on your own.

## Reproducible example

I observed this warning building from source on macOS 12.2.1, with Python v3.9.7 and `setuptools==59.8.0`.

```shell
git clone --recursive https://github.com/microsoft/LightGBM.git
cd LightGBM/python-package
```

This warning is raised with `python setup.py install`.

```shell
python setup.py install
```

It is also raised with `pip`.

```shell
pip install -v .
```

You can also see this warning on LightGBM's CI builds that use `python setup.py {command}`. For example, the most recent `Linux regular` job running on Azure DevOps ([build link](https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=12350&view=logs&j=c28dceab-947a-5848-c21f-eef3695e5f11&t=fa158246-17e2-53d4-5936-86070edbaacf)).

<img width="1437" alt="Screen Shot 2022-03-08 at 10 59 33 PM" src="https://user-images.githubusercontent.com/7608904/157376033-3d395653-731d-484b-be7a-a57c115198e5.png">

## References

[...]

PR #5625, opened 2022-12-07
WIP: [python-package] [ci] use setup.cfg and pyproject.toml for package-building configuration

jameslamb:
Contributes to #5061 .

This PR proposes moving some of the configuration for how to build the Python package out of `setup.py` and into the two configuration files mentioned in [PEP 517](https://peps.python.org/pep-0517/) and [PEP 518](https://peps.python.org/pep-0518/).

* `setup.cfg`: package metadata like license file, requirements, trove classifiers, etc.
* `pyproject.toml`: description of which "build backend" to use and what is required to build sdists and wheels

This shouldn't break anything about LightGBM's existing Python package source distributions or wheels. But it gets the project a bit closer to supporting the package-building standards described in PEP 517 and PEP 518.

### Shouldn't `cmake` be a build dependency?

Probably, but I'm proposing deferring that to a later PR. Including `cmake` as a build dependency in `pyproject.toml` causes installation of the `cmake` from PyPI, regardless of whether or not you had separately installed `cmake` another way.

If you're on a system that PyPI doesn't have a `cmake` wheel for, that means building `lightgbm` from source will also require building `cmake` from source. See for example the failed CUDA CI jobs from an earlier state of this PR ([build link](https://github.com/microsoft/LightGBM/actions/runs/3640814330/jobs/6145981701)).

> Failed to build cmake
  ERROR: Could not build wheels for cmake, which is required to install pyproject.toml-based projects
  error: subprocess-exited-with-error

### References

* [PEP 517](https://peps.python.org/pep-0517/): "PEP 517 – A build-system independent format for source trees"
* [PEP 518](https://peps.python.org/pep-0518/): "Specifying Minimum Build System Requirements for Python Projects"
* "Dependencies management in setuptools" ([link to setuptools docs](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html))
* PyPA docs on `pyproject.toml` ([link](https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/))


jameslamb:
I've learned a lot since first opening this.I'm going to close it, lock it, and open a new PR with a different proposal in the next few days.

PR #5630, opened 2022-12-13
Support IBM i/PASE systems

zheddie:
This PR is for the support of IBM i/PASE env. The code changes are all covered by IBM i /PASE macros to avoid it affecting any other platforms. Please note, we also upgraded the fmt and fast_double_parser external packages to make it works on our platform. Please help to review it before merge. thanks.

### DIFF
--- a/python-package/setup.py
+++ b/python-package/setup.py
@@ -246,7 +246,11 @@ class CustomInstall(install):
         self.hdfs = False
         self.precompile = False
         self.time_costs = False
-        self.nomp = False
+        #Add for PASE on i
+        if system() == 'OS400':
+            self.nomp = True
+        else:
+            self.nomp = False
         self.bit32 = False
 
     def run(self) -> None:
@@ -292,7 +296,11 @@ class CustomBdistWheel(bdist_wheel):
         self.hdfs = False
         self.precompile = False
         self.time_costs = False
-        self.nomp = False
+        #Add for PASE on i
+        if system() == 'OS400':
+            self.nomp = True
+        else:
+            self.nomp = False
         self.bit32 = False
 
     def finalize_options(self) -> None:
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -289,6 +289,11 @@ if(USE_HDFS)
     set(HDFS_CXX_LIBRARIES ${HDFS_LIB} ${JAVA_JVM_LIBRARY})
 endif()
 
+# Add for IBM i PASE
+if (${CMAKE_SYSTEM_NAME} MATCHES "OS400")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-extern-tls-init")
+endif()
+
 include(CheckCXXSourceCompiles)
 check_cxx_source_compiles("
 #include <xmmintrin.h>
--- a/src/network/socket_wrapper.hpp
+++ b/src/network/socket_wrapper.hpp
@@ -35,7 +35,14 @@
 #include <sys/types.h>
 #include <unistd.h>
 
+#ifdef __PASE__
+#include <as400_protos.h>
+#define ifaddrs ifaddrs_pase
+#define getifaddrs Qp2getifaddrs
+#define freeifaddrs Qp2freeifaddrs
+#else
 #include <ifaddrs.h>
+#endif
 
 #endif  // defined(_WIN32)
 


jameslamb:
Thanks very much for your proposal.

Before we continue, please provide answers for the following.

* How would this patch improve LightGBM?
* What is "IBM i/PASE" and why is supporting it important for LightGBM?
* Do you work for IBM?
* How can we get access to an IBM i/PASE system to test these changes?
* Are you or some group you work with (company, research institution, etc.) committed to supporting these changes, e.g. responding in a timely manner if we receive bug reports about this support and `@` you?

Links to documentation would be greatly appreciated.

<hr>

> we also upgraded the fmt and fast_double_parser external packages

How did you do that, exactly? I see https://github.com/lemire/fast_double_parser/pull/70 but don't see a related change at https://github.com/fmtlib/fmt.

<hr>

In the future, I strongly recommend not using `master` on your fork when making pull requests to open source projects. That makes it very difficult to keep up with upstream changes during the period where your PR is open, and makes it difficult to work on multiple pull requests at the same time.

python-package/setup.py
```diff
@@ -246,7 +246,11 @@ def initialize_options(self) -> None:
         self.hdfs = False
         self.precompile = False
         self.time_costs = False
-        self.nomp = False
+        #Add for PASE on i
+        if system() == 'OS400':
```
jameslamb, line 250: Please do this in `CMakeLists.txt` so most builds of LightGBM will benefit from it. The Python package shouldn't need to be aware of this.

jameslamb (changes requested):
I left one small code suggestion based on a quick review. Otherwise, will wait for your answers to the questions I've asked about the motivations for this PR.

zheddie:
> @microsoft-github-policy-service agree [company="{your company}"]

@microsoft-github-policy-service agree [company="IBM"]

zheddie:
> * How would this patch improve LightGBM?
> 
> * What is "IBM i/PASE" and why is supporting it important for LightGBM?
> 
> * Do you work for IBM?
> 
> * How can we get access to an IBM i/PASE system to test these changes?
> 
> * Are you or some group you work with (company, research institution, etc.) committed to supporting these changes, e.g. responding in a timely manner if we receive bug reports about this support and `@` you?

Thanks for your comments @jameslamb . I am so sorry for the long time wait , per my first round of Covid-19. This may make it more difficult for us to keep track of the latest changes in *master branch.  Which branch would you suggest for this kind of PR? I may create a new one if you like. 

Let me try to answer the questions you raised:

    How would this patch improve LightGBM?  
    A: This patch is for IBM i platform. With the support of this platform, LightGBM could extend the scope of customers. This would give more profit and opportunity to improve its quality and performance.  
    What is "IBM i/PASE" and why is supporting it important for LightGBM? 
    A : IBM i/PASE platform is one of the IBM platform, we have millions of customers in financial over the world. With the support of this platform , LightGBM could affect more customers and business. 
    Do you work for IBM? 
    A: Yes.
    How can we get access to an IBM i/PASE system to test these changes?
    A: we would do the test on our system, and we would also make sure the changes would not break or affect lightGBM on any existing platforms. 
    Are you or some group you work with (company, research institution, etc.) committed to supporting these changes, e.g. responding in a timely manner if we receive bug reports about this support and @ you?
    A: For us, the ML/AI on our platform is also in the very beginning phase. We do not have any pressure to commit this kind of changes , although we wish to make it available ASAP, which would encourage our customers to use them.

zheddie:
@jameslamb as for this quesiton:
```
How did you do that, exactly? I see https://github.com/lemire/fast_double_parser/pull/70 but don't see a related change at https://github.com/fmtlib/fmt.
```
Yes, we did changes in fast_double_parse project to make it fit for our platform. For the fmt , we really have similar issue on current level of fmt integrated in LightGBM. But with the latest version of fmt(as I updated by this commit), the related code was changed, the original issue disappear, so, that's the reason why we do not have similar PR in fmt project. But just do an version upgrade .

jameslamb (changes requested):
Thanks for the answers!

Given that, I personally do not support this pull request.

LightGBM is an open-source project. We do not have a notion of "customers" or "profit", and I just don't believe the implication that this patch might bring LightGBM to "millions" of new users.

The fact that this requires updating the versions of `fast_double_parser` and `fmt` vendored into this project also directly contradicts the claim the claim that *"the changes would not ... affect LightGBM on any existing platforms"*.

The small team of maintainers here has very limited available time already, and I personally don't think the information you've provided justifies us spending any more of it to support a proprietary, for-profit platform.

I will leave this open for a few more days to give other maintainers a chance to comment, but I don't think this pull request should be merged and I don't support taking on support for IBM i/PASE at this time.

@guolinke @shiyu1994 @jmoralez @StrikerRUS what do you think?

jameslamb:
It's been 2+ months without a response from the author, so I'm going to close this.

PR #5645, opened 2022-12-23
[ci] make GitHub Actions branch protection stricter (fixes #5501)

### LINKED ISSUE #5501: [ci] Problematic `all-successful` GHA job implementation
_[a drive-by advice]_

Hi, I noticed that you have an `all-successful` job in the workflows. The idea is great but the way it's implemented may let broken changes slip through. There are corner cases when this job would be marked as `skipped` which branch protection would interpret as "not red, good to go". To solve this properly, it'd have to run with `if: always()` and enumerate+check statuses of each dependency-job.
This is a lot more work so I've generalized all necessary in an action @ https://github.com/marketplace/actions/alls-green#why.

Another problem I foresee is that there' multiple workflows reporting the same job name `all-successful` so if it happens that some workflows succeed but others fail (or would be slow) to post a status (hello, eventual consistency of GH), it'd also allow merging broken changes. I normally recommend having unique check/gate job names per workflow because of this.

jameslamb:
Fixes #5501.

Following advice in that issue and in https://github.com/orgs/community/discussions/26733#discussioncomment-3253155, this proposes two changes to the project's GitHub Actions jobs.

* use https://github.com/re-actors/alls-green to enforce multiple checks passing without needing to change branch protection settings every time checks are added or removed
* give each of those "all the checks passed" jobs a unique name

This will make use of GitHub auto-merging more reliable, and reduce the risk of accidentally merging PRs where some CI jobs are failing.

### Notes for Reviewers

@guolinke @shiyu1994 can you please go into the branch protection for `master` at https://github.com/microsoft/LightGBM/settings/branches and modify the section `Require status checks to pass before merging`?

The following should be checked as required:

* `all-cuda-jobs-successful` (GitHub Actions)
* `all-optional-checks-successful` (GitHub Actions)
* `all-python-package-jobs-successful` (GitHub Actions)
* `all-r-package-jobs-successful` (GitHub Actions)
* `all-static-analysis-jobs-successful` (GitHub Actions)

And the following should be unchecked:

* `all-successful` (GitHub Actions)

I think that only the two of you have permissions in this repo to modify that.

### DIFF
--- a/.github/workflows/cuda.yml
+++ b/.github/workflows/cuda.yml
@@ -107,10 +107,12 @@ jobs:
                 docker_img="${docker_img}-ubuntu$(lsb_release -rs)" 
             fi
             docker run --env-file docker.env -v "$GITHUB_WORKSPACE":"$ROOT_DOCKER_FOLDER" --rm --gpus all "$docker_img" /bin/bash $ROOT_DOCKER_FOLDER/docker-script.sh
-  all-successful:
-    # https://github.community/t/is-it-possible-to-require-all-github-actions-tasks-to-pass-without-enumerating-them/117957/4?u=graingert
+  all-cuda-jobs-successful:
+    if: always()
     runs-on: ubuntu-latest
     needs: [test]
     steps:
     - name: Note that all tests succeeded
-      run: echo "🎉"
+      uses: re-actors/alls-green@v1.2.2
+      with:
+        jobs: ${{ toJSON(needs) }}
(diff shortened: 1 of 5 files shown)


guolinke:
I think i don't have permission to do it, ping @shiyu1994

shiyu1994:
@jameslamb Done with that.

jameslamb:
thanks @shiyu1994 ! As soon as this is approved, let's merge it.

shiyu1994 (approved):
LGTM

jameslamb:
Thanks so much @shiyu1994 !

PR #5649, opened 2022-12-29
[docs] encourage use of releases for GUI-only installations

jameslamb:
Replaces #5498.

Implements the proposal in https://github.com/microsoft/LightGBM/pull/5498#pullrequestreview-1134818427.

Modifies the installation docs to encourage users who want to install LightGBM using only a GUI to use official releases. I hope this will help to prevent issues like #5495, where GUI-only users see broken builds because the GitHub source archive doesn't pull in git submodules.

### DIFF
--- a/docs/Installation-Guide.rst
+++ b/docs/Installation-Guide.rst
@@ -56,7 +56,7 @@ With GUI
 
 1. Install `Visual Studio`_ (2015 or newer).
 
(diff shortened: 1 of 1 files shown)


jameslamb:
@guolinke or @shiyu1994 can you help with a review on this? I think it should be pretty quick and non-controversial.

PR #5672, opened 2023-01-11
[python-package] fix mypy errors about missing annotations and incompatible types

IdoKendo:
`mypy` currently raises the following errors:
```
python-package\lightgbm\callback.py:249: error: Need type annotation for "best_score" (hint: "best_score: List[<type>] = ...")  [var-annotated]
python-package\lightgbm\callback.py:250: error: Need type annotation for "best_iter" (hint: "best_iter: List[<type>] = ...")  [var-annotated]
python-package\lightgbm\callback.py:251: error: Need type annotation for "best_score_list" (hint: "best_score_list: List[<type>] = ...")  [var-annotated]
python-package\lightgbm\callback.py:252: error: Need type annotation for "cmp_op" (hint: "cmp_op: List[<type>] = ...")  [var-annotated]
python-package\lightgbm\engine.py:169: error: Argument 1 to "set_feature_name" of "Dataset" has incompatible type "Union[List[str], str]"; expected "List[str]"  [arg-type]
python-package\lightgbm\engine.py:170: error: Argument 1 to "set_categorical_feature" of "Dataset" has incompatible type "Union[List[str], List[int], str]"; expected "Union[List[int], List[str]]"  [arg-type]
python-package\lightgbm\engine.py:262: error: Incompatible types in assignment (expression has type "Union[List[Tuple[str, str, float, bool]], List[Tuple[str, str, float, bool, float]]]", variable has type "List[Tuple[str, str, float, bool]]")  [assignment]
python-package\lightgbm\engine.py:677: error: Argument 1 to "set_feature_name" of "Dataset" has incompatible type "Union[str, List[str]]"; expected "List[str]"  [arg-type]
python-package\lightgbm\engine.py:678: error: Argument 1 to "set_categorical_feature" of "Dataset" has incompatible type "Union[str, List[str], List[int]]"; expected "Union[List[int], List[str]]"  [arg-type]
python-package\lightgbm\sklearn.py:897: error: Incompatible return value type (got "None", expected "int")  [return-value]
python-package\lightgbm\sklearn.py:911: error: Incompatible return value type (got "None", expected "int")  [return-value]
python-package\lightgbm\sklearn.py:1200: error: Incompatible default for argument "eval_at" (default has type "Tuple[int, int, int, int, int]", argument has type "Union[List[int], Tuple[int]]")  [assignment]
python-package\lightgbm\dask.py:1069: error: Argument "eval_at" to "_train" has incompatible type "Optional[Iterable[int]]"; expected "Union[List[int], Tuple[int], None]"  [arg-type]
```
These errors due to missing and mismatched annotations in the code. This PR proposes fixing these errors by aligning all the annotations in these areas of the code to share the same annotation.
## Notes for Reviewers
This was tested by running mypy as documented in https://github.com/microsoft/LightGBM/issues/3867.

```
mypy \
    --ignore-missing-imports \
    python-package/
```

### DIFF
--- a/python-package/lightgbm/callback.py
+++ b/python-package/lightgbm/callback.py
@@ -8,7 +8,8 @@ from .basic import _ConfigAliases, _LGBM_BoosterEvalMethodResultType, _log_info,
 
 _EvalResultTuple = Union[
     List[_LGBM_BoosterEvalMethodResultType],
-    List[Tuple[str, str, float, bool, float]]
+    List[Tuple[str, str, float, bool, float]],
+    None
 ]
 
 
@@ -43,6 +44,8 @@ CallbackEnv = collections.namedtuple(
 
 def _format_eval_result(value: _EvalResultTuple, show_stdv: bool = True) -> str:
     """Format metric string."""
+    if value is None:
+        raise ValueError("Wrong metric value")
     if len(value) == 4:
         return f"{value[0]}'s {value[1]}: {value[2]:g}"
     elif len(value) == 5:
@@ -246,10 +249,10 @@ class _EarlyStoppingCallback:
         self._reset_storages()
 
     def _reset_storages(self) -> None:
-        self.best_score = []
-        self.best_iter = []
-        self.best_score_list = []
-        self.cmp_op = []
+        self.best_score: List[float] = []
+        self.best_iter: List[int] = []
+        self.best_score_list: List[_EvalResultTuple] = []
+        self.cmp_op: List[Callable[[float, float], bool]] = []
         self.first_metric = ''
 
     def _gt_delta(self, curr_score: float, best_score: float, delta: float) -> bool:
--- a/python-package/lightgbm/dask.py
+++ b/python-package/lightgbm/dask.py
@@ -407,7 +407,7 @@ def _train(
     eval_init_score: Optional[List[_DaskCollection]] = None,
     eval_group: Optional[List[_DaskVectorLike]] = None,
     eval_metric: Optional[_LGBM_ScikitEvalMetricType] = None,
-    eval_at: Optional[Union[List[int], Tuple[int]]] = None,
+    eval_at: Union[List[int], Tuple[int, ...], None] = None,
     **kwargs: Any
 ) -> LGBMModel:
     """Inner train routine.
@@ -1039,7 +1039,7 @@ class _DaskLGBMModel:
         eval_init_score: Optional[List[_DaskCollection]] = None,
         eval_group: Optional[List[_DaskVectorLike]] = None,
         eval_metric: Optional[_LGBM_ScikitEvalMetricType] = None,
-        eval_at: Optional[Iterable[int]] = None,
+        eval_at: Union[List[int], Tuple[int, ...], None] = None,
         **kwargs: Any
     ) -> "_DaskLGBMModel":
         if not DASK_INSTALLED:
@@ -1168,7 +1168,7 @@ class DaskLGBMClassifier(LGBMClassifier, _DaskLGBMModel):
         eval_init_score: Optional[List[_DaskCollection]] = None,
         eval_metric: Optional[_LGBM_ScikitEvalMetricType] = None,
         **kwargs: Any
-    ) -> "DaskLGBMClassifier":
+    ) -> "_DaskLGBMModel":
         """Docstring is inherited from the lightgbm.LGBMClassifier.fit."""
         return self._lgb_dask_fit(
             model_factory=LGBMClassifier,
@@ -1215,13 +1215,29 @@ class DaskLGBMClassifier(LGBMClassifier, _DaskLGBMModel):
     {_lgbmmodel_doc_custom_eval_note}
         """
 
-    def predict(self, X: _DaskMatrixLike, **kwargs: Any) -> dask_Array:
+    def predict(
+        self,
+        X: _DaskMatrixLike,
+        raw_score: bool = False,
+        start_iteration: int = 0,
+        num_iteration: Optional[int] = None,
+        pred_leaf: bool = False,
+        pred_contrib: bool = False,
+        validate_features: bool = False,
+        **kwargs: Any
+    ) -> dask_Array:
         """Docstring is inherited from the lightgbm.LGBMClassifier.predict."""
         return _predict(
             model=self.to_local(),
             data=X,
             dtype=self.classes_.dtype,
             client=_get_dask_client(self.client),
+            raw_score=raw_score,
+            start_iteration=start_iteration,
+            num_iteration=num_iteration,
+            pred_leaf=pred_leaf,
+            pred_contrib=pred_contrib,
+            validate_features=validate_features,
             **kwargs
         )
 
@@ -1234,13 +1250,29 @@ class DaskLGBMClassifier(LGBMClassifier, _DaskLGBMModel):
         X_SHAP_values_shape="Dask Array of shape = [n_samples, n_features + 1] or shape = [n_samples, (n_features + 1) * n_classes] or (if multi-class and using sparse inputs) a list of ``n_classes`` Dask Arrays of shape = [n_samples, n_features + 1]"
     )
 
-    def predict_proba(self, X: _DaskMatrixLike, **kwargs: Any) -> dask_Array:
+    def predict_proba(
+        self,
+        X: _DaskMatrixLike,
+        raw_score: bool = False,
+        start_iteration: int = 0,
+        num_iteration: Optional[int] = None,
+        pred_leaf: bool = False,
+        pred_contrib: bool = False,
+        validate_features: bool = False,
+        **kwargs: Any
+    ) -> dask_Array:
         """Docstring is inherited from the lightgbm.LGBMClassifier.predict_proba."""
         return _predict(
             model=self.to_local(),
             data=X,
             pred_proba=True,
             client=_get_dask_client(self.client),
+            raw_score=raw_score,
+            start_iteration=start_iteration,
+            num_iteration=num_iteration,
+            pred_leaf=pred_leaf,
+            pred_contrib=pred_contrib,
+            validate_features=validate_features,
             **kwargs
         )
 
@@ -1339,7 +1371,7 @@ class DaskLGBMRegressor(LGBMRegressor, _DaskLGBMModel):
         eval_init_score: Optional[List[_DaskVectorLike]] = None,
         eval_metric: Optional[_LGBM_ScikitEvalMetricType] = None,
         **kwargs: Any
-    ) -> "DaskLGBMRegressor":
+    ) -> "_DaskLGBMModel":
         """Docstring is inherited from the lightgbm.LGBMRegressor.fit."""
         return self._lgb_dask_fit(
             model_factory=LGBMRegressor,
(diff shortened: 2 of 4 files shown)


IdoKendo:
@microsoft-github-policy-service agree

jameslamb (changes requested):
🎉 Thanks so much for working on this @IdoKendo !

Could you please move the two `basic.py` changes and the `eval_at` tuple fixes to another PR? I totally support and understand those, and we could merge a PR with them immediately.

The other changes touch more complicated parts of the code and will take me a bit longer to review.

IdoKendo:
@jameslamb Sure thing, I moved those changes to https://github.com/microsoft/LightGBM/pull/5674.
I'll continue working on the rest of the errors as well. LMK regarding the changes left in this PR.

jameslamb (changes requested):
Thanks for this!

I agree with the `predict()` changes for the Dask interface. Can you please pull those out into a separate PR? I'll approve and merge that quickly.

Sorry to ask you to do that again, but please understand that we're not only interested in resolving these warnings, but also in understanding why `mypy` is raising them. Since the root cause of some might be bugs or undesirable behavior in `lightgbm`, not just mistakes in the type hints. For example... why does `mypy` think there's a code path that results in `callback._format_eval_result()` being called on a value that's `None`? Should we just eliminate that code path? I'd like to investigate concerns like that before approving that change.

IdoKendo:
@jameslamb Sure thing, I moved those to https://github.com/microsoft/LightGBM/pull/5678.
I fully understand the investigation before approving!
I have some more fixes already lined up, should I add them to this PR then? Or would it be more comfortable for you to have them separately from the get-go?

jameslamb:
> I have some more fixes already lined up, should I add them to this PR then?

Awesome! I'd appreciate more, smaller PRs for this, each focused on one part of the codebase. That'll allow us to merge in the changes that require less investigation quickly.

jameslamb:
@IdoKendo Are you interested in pursuing this?

If so, please update this PR to latest `master` and I'll provide a review.

IdoKendo:
> @IdoKendo Are you interested in pursuing this?
> 
> If so, please update this PR to latest `master` and I'll provide a review.

Yeah for sure, updated!

python-package/lightgbm/sklearn.py
```diff
@@ -897,7 +897,7 @@ def n_features_(self) -> int:
         return self._n_features
 
     @property
-    def n_features_in_(self) -> int:
+    def n_features_in_(self) -> Optional[int]:
```
jameslamb, line 900: I think you're missing some important context around how these `n_features_in_` and `best_iteration_` attributes work.

As you can see in the methods, before returning a value the function `self.__sklearn_is_fitted__()` is called. That method returns `True` if `self.fitted_` exists and is `True` (see #4636).

https://github.com/microsoft/LightGBM/blob/29796eee6eabb79e161880976ec02fcd95dfcc4a/python-package/lightgbm/sklearn.py#L548-L549

If `self.__sklearn_is_fitted__()` returns `True`, the attributes referenced by `self.n_features_in_` and `self.best_iteration_` should not be `None`. Fitting a model should set all of those attributes.

But `mypy` can't possibly know about those specifics of how `lightgbm` is expected to work and be used, and it's rightly complaining that "hey, you initialized `self.best_iteration` to `None` and now you're returning it...therefore it's possible for that value to be `None`".

I don't want to change these type hints to `Optional[None]` because then any other code relying on these attributes (including code in projects downstream of `lightgbm` like `optuna`) will have to face this awkwardness and either use `# type: ignore` or add defensive code checking like `if model.best_iteration is not None: ...`.

I tried one approach to work around these back in 2021 (https://github.com/microsoft/LightGBM/pull/4837#discussion_r759757551) that we decided to not adopt, and looking back now I agree that adding assertions or other new complexity might not be the best way to handle this.

I think a better approach would be to initialize those attributes to values of the expected return type, e.g. to initialize `self.best_score_` to `-1.0`.

But I think that needs to be done by someone who's more familiar with LightGBM, like one of the maintainers here, because it's important to, for example, check for cases where those internal attributes are being compared to `None`.

Ok so! After all that.... please just remove these changes to `best_iteration_` and `n_features_in` from this PR. I will address them in a separate PR.

jameslamb (changes requested):
Thanks for this! Please see my next round of suggestions.

python-package/lightgbm/sklearn.py
jameslamb: I just put up #5718 with a proposed fix.

python-package/lightgbm/dask.py
```diff
@@ -1256,13 +1256,29 @@ def predict(
         X_SHAP_values_shape="Dask Array of shape = [n_samples, n_features + 1] or shape = [n_samples, (n_features + 1) * n_classes] or (if multi-class and using sparse inputs) a list of ``n_classes`` Dask Arrays of shape = [n_samples, n_features + 1]"
     )
 
-    def predict_proba(self, X: _DaskMatrixLike, **kwargs: Any) -> dask_Array:
+    def predict_proba(
```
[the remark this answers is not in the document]
IdoKendo: Sure! Moved to https://github.com/microsoft/LightGBM/pull/5728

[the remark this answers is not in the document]
jameslamb: thank you!

python-package/lightgbm/dask.py
```diff
@@ -1174,7 +1174,7 @@ def fit(  # type: ignore[override]
         eval_init_score: Optional[List[_DaskCollection]] = None,
         eval_metric: Optional[_LGBM_ScikitEvalMetricType] = None,
         **kwargs: Any
-    ) -> "DaskLGBMClassifier":
+    ) -> "_DaskLGBMModel":
```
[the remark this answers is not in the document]
jameslamb: I noticed you've removed these Dask changes from this PR and they aren't included in the other mypy-related PR you opened (#5731).

Are you interested in trying them out? If not, I'll fix these issues related to `lightgbm.dask` return types.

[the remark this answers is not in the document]
IdoKendo: Since we're doing `# type: ignore[override]` it seemed kind of redundant to me.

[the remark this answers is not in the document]
jameslamb: It is not. No worries, I'll attempt this particlar fix.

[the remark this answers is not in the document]
jameslamb: Created #5756.

To add more details on why the presence of `# type: ignore[override]` doesn't make this concern redundant...that check and these errors about return types are about different things.

The `[override]` check is saying "hey, you told me this was a child of `{ParentClass}`, but it has a method with a different signature than `{ParentClass}`, which means it can't be automatically used anywhere that `{ParentClass}` is used".

The errors about return types are saying "hey you told me that this `fit()` method returns an instance of `{ChildClass}` but it's returning an instance of `{ParentClass}`". That's a different thing, and important to fix for at least 2 reasons:

1. `{ChildClass}` might define additional attributes that `{ParentClass}` doesn't (e.g. extra configuration captured through the constructor)
2. People writing code that depends on `lightgbm` may wish to perform checks like `isinstance(model, lgb.DaskLGBMRegressor)`. Such checks would fail today if called on the result of `fit()`, since it's returning an internal class `lgb._DaskLGBMModel`.

python-package/lightgbm/callback.py
```diff
@@ -16,7 +16,8 @@
 _EvalResultDict = Dict[str, Dict[str, List[Any]]]
 _EvalResultTuple = Union[
     List[_LGBM_BoosterEvalMethodResultType],
-    List[Tuple[str, str, float, bool, float]]
+    List[Tuple[str, str, float, bool, float]],
+    None
```
jameslamb, line 20: Please don't include `None` in this type hint. That makes it difficult to use in other parts of the code where `None` isn't a valid value.

python-package/lightgbm/callback.py
```diff
@@ -254,10 +257,10 @@ def __init__(
         self._reset_storages()
 
     def _reset_storages(self) -> None:
-        self.best_score = []
-        self.best_iter = []
-        self.best_score_list = []
-        self.cmp_op = []
+        self.best_score: List[float] = []
+        self.best_iter: List[int] = []
+        self.best_score_list: List[_EvalResultTuple] = []
```
jameslamb, line 262:
```suggestion
        self.best_score_list: List[Union[_EvalResultTuple, None]] = []
```

Please include the `None` here in this hint, instead of as a part of `_EvalResultTuple`

jameslamb (changes requested):
Thanks for this! Especially for working through figuring out the signature of `cmp_op`.

Please make the one modification I've requested in this newest review, and update the list of `mypy` errors in the description.

python-package/lightgbm/callback.py
IdoKendo: Done :)

python-package/lightgbm/callback.py
```diff
@@ -363,12 +364,14 @@ def __call__(self, env: CallbackEnv) -> None:
...
                 continue  # train data for lgb.cv or sklearn wrapper (underlying lgb.train)
             elif env.iteration - self.best_iter[i] >= self.stopping_rounds:
-                if self.verbose:
-                    eval_result_str = '\t'.join([_format_eval_result(x, show_stdv=True) for x in self.best_score_list[i]])
-                    _log_info(f"Early stopping, best iteration is:\n[{self.best_iter[i] + 1}]\t{eval_result_str}")
-                    if self.first_metric_only:
-                        _log_info(f"Evaluated only: {eval_name_splitted[-1]}")
-                raise EarlyStopException(self.best_iter[i], self.best_score_list[i])
+                best_score_list = self.best_score_list[i]
+                if best_score_list is not None:
```
jameslamb, line 368: Why is this new refactoring that you pushed (pulling out this item of `self.best_score_list` and checking if it's `None`) necessary? If the purpose of the PR is to address 4 warnings that are just "need a type annotation"?

jameslamb (changes requested):

python-package/lightgbm/callback.py
IdoKendo: If we remove the None hint from `_EvalResultTuple` then this part introduces another error.

jameslamb: can you please tell me what the new error is that this refactoring helps to avoid?

Changing logic requires more review effort than just-changing-type-hints types of changes. I'm happy to provide that, but I want to understand first what feedback from `mypy` led to this proposed change.

IdoKendo: These errors:
```
python-package\lightgbm\callback.py:369: error: Item "None" of "Optional[Union[List[Tuple[str, str, float, bool]], List[Tuple[str, str, float, bool, float]]]]" has no attribute "__iter__" (not iterable)  [union-attr]
python-package\lightgbm\callback.py:373: error: Argument 2 to "EarlyStopException" has incompatible type "Optional[Union[List[Tuple[str, str, float, bool]], List[Tuple[str, str, float, bool, float]]]]"; expected "Union[List[Tuple[str, str, float, bool]], List[Tuple[str, str, float, bool, float]]]"  [arg-type]
```

jameslamb: hmmm I'm struggling to understand that. 

I don't understand why `mypy` thinks that the type at this point is

```text
Optional[Union[List[Tuple[str, str, float, bool]], List[Tuple[str, str, float, bool, float]]]]
```

I understand how `best_score_list` could *have a `None` as a list element*, but now how it could itself be `None`.

Anyway, I'll try to review this when I can. Now that the logic is changing, it'll take longer to review. Sorry about that.

jameslamb (changes requested):
I still need some time to investigate the proposed changes for `best_score_list`, sorry.

But I 100% support the changes for `best_score`, `best_iter`, and `cmp_op`. Could you please put those up in a separate PR? I'll approve and merge that quickly.

IdoKendo:
> I still need some time to investigate the proposed changes for `best_score_list`, sorry.
> 
> But I 100% support the changes for `best_score`, `best_iter`, and `cmp_op`. Could you please put those up in a separate PR? I'll approve and merge that quickly.

Sure, I think I'll do it the other way and just move the refactor to a different PR, keeping the types in this one.

jameslamb:
Ok that works, thank you!

jameslamb (approved):
thanks!

PR #5689, opened 2023-01-28
[ci] get clang-15 from bookworm repository (fixes #5688)

jameslamb:
Fixes #5688.

The `rhub/debian-*` container images are built from `debian:testing` ([code link](https://github.com/r-hub/rhub-linux-builders/blob/47ba102648e5db3037e100088452c9e8cd440367/debian/Dockerfile#LL3C6-L3C20)).

Seems like a few days ago, that image switched to Debian `bookworm` (the major release that follows `bullseye`).

```shell
docker run \
    --rm \
    debian:testing \
    /bin/bash -c "apt-get update -y && apt-get install -y lsb-release && lsb_release -a"
```

```text
Distributor ID:	Debian
Description:	Debian GNU/Linux bookworm/sid
Release:	n/a
Codename:	bookworm
```

This fixes the R CI job in this project using that image by switching to the package repository that LLVM developers publish bookworm packages to (see https://apt.llvm.org/).

PR #5690, opened 2023-01-28
[R-package] use C++17 in the CRAN package

jameslamb:
Proposes using the C++17 standard in the CRAN build of the R package.

### Motivation

CRAN recently started raising the following NOTE on r-devel.

> Result: NOTE
>     Specified C++11: please update to current default of C++17

(https://cran.r-project.org/web/checks/check_results_lightgbm.html)

That check just made it into R about 2 days ago ([commit link](https://github.com/wch/r-source/commit/eb2531413e22a939e051be97ed062f29306b12e9)), in preparation for C++17 becoming the default C++ standard in R's next release, v4.3.0.

So I expect that CRAN is going to ask us to make this change soon.

### Implications of this change

If this PR is accepted, 

* any C++ code included in the R package will need to conform to the C++17 standard
    - *that includes most of `src/`, `include/` and `external_libs/`*
    - *that means, for example, not using features removed in C++14/C++17 like `std::auto_ptr` and `std::random_shuffle` ([docs](https://en.cppreference.com/w/cpp/17))*
* the CRAN package will no longer be compilable with compilers that don't support C++17
    - `gcc >= 5` ([October 2017](https://gcc.gnu.org/releases.html), [supporting docs](https://gcc.gnu.org/projects/cxx-status.html#cxx17))
    - `clang >= 5` ([September 2017](https://en.wikipedia.org/wiki/Clang), [supporting docs](https://clang.llvm.org/cxx_status.html))
    - `MSVC >= 2019` ([April 2019](https://learn.microsoft.com/en-us/visualstudio/releases/2019/release-notes-v16.0), [supporting docs](https://learn.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance?view=msvc-170))
    - `nvcc` in CUDA 11+ (May 2020, [supporting docs](https://developer.nvidia.com/blog/cuda-11-features-revealed/))


.github/workflows/r_package.yml
```diff
@@ -308,7 +308,7 @@ jobs:
           #
           wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
           #
-          add-apt-repository "deb http://apt.llvm.org/bullseye/ llvm-toolchain-bullseye-15 main"
+          add-apt-repository "deb http://apt.llvm.org/unstable/ llvm-toolchain main"
```
jameslamb, line 311: This is from #5689, just to unblock CI on this PR. It'll drop out of the diff once that's merged.

jameslamb:
@guolinke @jmoralez could I please get a review on this proposal when you have time?

R-package/DESCRIPTION
```diff
@@ -62,5 +62,5 @@ Imports:
     parallel,
     utils
 SystemRequirements:
-    C++11
+    ~~CXXSTD~~
```
guolinke, line 65: The permission of this file also changed: "100755 → 100644", is that expected?

jameslamb: thanks for catching that!

I used GitHub's in-browser merge conflict resolution after merging #5341 . Maybe it changed the permissions.

Fixed in https://github.com/microsoft/LightGBM/pull/5690/commits/79ee233ba7a1605722ed41984bef21e35cb95970

guolinke (approved):

PR #5699, opened 2023-02-04
Continue training in CLI if one iteration produces a single-leaf tree

Samsagax:
As per discussion on #5051 and #5193 the python package does not stop training if a single leaf tree (stupm) is found and relies on early stopping methods to stop training. This commits removes the finish condition on training based on the result of `TrainOneIter()` and sets the `is_finished` flag on early stopping alone.

### DIFF
--- a/src/boosting/gbdt.cpp
+++ b/src/boosting/gbdt.cpp
@@ -231,10 +231,8 @@ void GBDT::Train(int snapshot_freq, const std::string& model_output_path) {
   bool is_finished = false;
   auto start_time = std::chrono::steady_clock::now();
   for (int iter = 0; iter < config_->num_iterations && !is_finished; ++iter) {
-    is_finished = TrainOneIter(nullptr, nullptr);
-    if (!is_finished) {
-      is_finished = EvalAndCheckEarlyStopping();
-    }
+    TrainOneIter(nullptr, nullptr);
+    is_finished = EvalAndCheckEarlyStopping();
     auto end_time = std::chrono::steady_clock::now();
     // output used time per iteration
     Log::Info("%f seconds elapsed, finished iteration %d", std::chrono::duration<double,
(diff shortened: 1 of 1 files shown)


jameslamb (commented):
Thanks very much for returning to this!

I've updated the PR title to more closely match the intent of this PR. Please let me know if I've misunderstood.

I think this PR matches the outcome of the discussion in #5051, but I'm not qualified to review this. Hopefully @shiyu1994 or @guolinke can help with a review.

jameslamb:
@shiyu1994 can you help with a review please? This has been open for several months now.

src/boosting/gbdt.cpp
```diff
@@ -421,6 +419,8 @@ bool GBDT::TrainOneIter(const score_t* gradients, const score_t* hessians) {
     models_.push_back(std::move(new_tree));
   }
 
+  ++iter_;
+
   if (!should_continue) {
```
guolinke, line 424: The one-leaf trees will be pop?

Samsagax: Tested the result and stopping or using it as this is works the same (assumming there is no other built tree down the line). Incrementing the `iter_` variable is needed for the boosting to take it as a finished tree (even if the tree is of no use). Otherwise it will continue forever.

guolinke: sorry for the late response. I think we should make the behavior the same as the python-package. I remember that the one-leaf tree will be kept in the python-package, but the current implementation will pop the one-leaf trees.

guolinke: @Samsagax what do you think about my comment?

jameslamb:
It's been almost a year since the last activity on this PR.

@guolinke what should we do with this?

guolinke:
"I think we should make the behavior the same as the python-package". 
if @Samsagax don't have time to continue working on it, we can close this PR for now, and create a feature request?

PR #5704, opened 2023-02-09
[python-package] remove some inner function definitions

jameslamb:
Contributes to #3756.

Some places in `python-package/basic.py` have functions which define additional internal functions every time they're called, like this:

```python
def register_logger(...):
    def _has_method(...) -> bool:
        return callable(getattr(logger, method_name, None))
    if not _has_method(...):
    # ...
```

This PR proposes pulling some of those functions up to standalone declarations.

Along the way, I'm also proposing:

* adding some types hints on the code touched by this PR
* removing unnecessary default values on internal functions that are only called once or twice, to reduce the risk of "forgot to pass through some argument" types of bugs

### Benefits of this change

* removes unnecessary computation from re-interpreting the same code each time a top-level function is called
* makes it easier to understand the flow of data through the code

### Notes for Reviewers

This PR's changes only alter internal details of the Python package.


python-package/lightgbm/basic.py
```diff
@@ -323,19 +325,22 @@ def _json_default_with_numpy(obj: Any) -> Any:
...
+
+
 def _param_dict_to_str(data: Optional[Dict[str, Any]]) -> str:
     """Convert Python dictionary to string, which is passed to C API."""
     if data is None or not data:
         return ""
     pairs = []
     for key, val in data.items():
         if isinstance(val, (list, tuple, set)) or _is_numpy_1d_array(val):
-            def to_string(x):
```
jameslamb, line 333: this is a good example of what this PR hopes to avoid...this `to_string()` method was being redefined for every element in `data` passed to every call to `_param_dict_to_str()`.

python-package/lightgbm/basic.py
```diff
@@ -564,15 +568,15 @@ def _c_int_array(data):
     return (ptr_data, type_data, data)  # return `data` to avoid the temporary copy is freed
 
 
-def _check_for_bad_pandas_dtypes(pandas_dtypes_series):
+def is_allowed_numpy_dtype(dtype) -> bool:
```
jmoralez, line 571: given that this is only used in a private function I think it should be private as well (I can't comment on the line that uses this function because it wasn't modified here but it'd need to change as well if you agree.
```suggestion
def _is_allowed_numpy_dtype(dtype) -> bool:
```

python-package/lightgbm/basic.py
```diff
@@ -1029,84 +1067,110 @@ def __create_sparse_native(self, cs, out_shape, out_ptr_indptr, out_ptr_indices,
...
-                ctypes.c_int(matrix_type),
-                out_shape.ctypes.data_as(ctypes.POINTER(ctypes.c_int64)),
-                ctypes.byref(out_ptr_indptr),
-                ctypes.byref(out_ptr_indices),
-                ctypes.byref(out_ptr_data)))
-            matrices = self.__create_sparse_native(csr, out_shape, out_ptr_indptr, out_ptr_indices, out_ptr_data,
-                                                   type_ptr_indptr, type_ptr_data, is_csr=True)
-            nrow = len(csr.indptr) - 1
-            return matrices, nrow
+        ptr_indptr, type_ptr_indptr, __ = _c_int_array(csr.indptr)
```
jmoralez, line 1086: I realize this was already like this, but this was probably meant to be a single underscore.
```suggestion
        ptr_indptr, type_ptr_indptr, _ = _c_int_array(csr.indptr)
```

jameslamb: yeah I think you're right, and I agree with the suggestion!

python-package/lightgbm/basic.py
jameslamb: agreed, thanks for catching that!

I did this in https://github.com/microsoft/LightGBM/pull/5704/commits/55abc4ab1d20843e20469b67b1bd3c6001dc86a8 while resolving merge conflicts, after merging #5694 .

jmoralez (approved):

PR #5710, opened 2023-02-14
[python-package] add more type hints in sklearn.py

jameslamb:
Contributes to #3756.
Contributes to #3867.

Adds a few additional type hints in `sklearn.py`, to improve static type checkers' ability to find bugs.

This PR also resolves the following `mypy` warnings:

```text
python-package/lightgbm/sklearn.py:1055: error: Incompatible types in assignment (expression has type "Dict[Any, Any]", variable has type "None")  [assignment]
python-package/lightgbm/sklearn.py:1057: error: Incompatible types in assignment (expression has type "Dict[Any, Any]", variable has type "None")  [assignment]
python-package/lightgbm/sklearn.py:1057: error: Value of type "None" is not indexable  [index]
python-package/lightgbm/sklearn.py:1060: error: Incompatible types in assignment (expression has type "int", variable has type "None")  [assignment]
python-package/lightgbm/sklearn.py:815: error: Incompatible types in assignment (expression has type "Dict[Any, Any]", variable has type "None")  [assignment]
python-package/lightgbm/sklearn.py:816: error: Incompatible types in assignment (expression has type "int", variable has type "None")  [assignment]
python-package/lightgbm/sklearn.py:817: error: Incompatible types in assignment (expression has type "Dict[Any, Any]", variable has type "None")  [assignment]
```

### DIFF
--- a/python-package/lightgbm/sklearn.py
+++ b/python-package/lightgbm/sklearn.py
@@ -7,9 +7,9 @@ from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
 import numpy as np
 
-from .basic import (Booster, Dataset, LightGBMError, _choose_param_value, _ConfigAliases, _LGBM_EvalFunctionResultType,
+from .basic import (Booster, Dataset, LightGBMError, _choose_param_value, _ConfigAliases, _LGBM_BoosterBestScoreType, _LGBM_EvalFunctionResultType,
                     _log_warning)
-from .callback import record_evaluation
+from .callback import _EvalResultDict, record_evaluation
 from .compat import (SKLEARN_INSTALLED, LGBMNotFittedError, _LGBMAssertAllFinite, _LGBMCheckArray,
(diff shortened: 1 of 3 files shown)


python-package/lightgbm/sklearn.py
```diff
@@ -519,18 +519,18 @@ def __init__(
         self.n_jobs = n_jobs
         self.importance_type = importance_type
         self._Booster: Optional[Booster] = None
-        self._evals_result = None
-        self._best_score = None
-        self._best_iteration = None
+        self._evals_result: _EvalResultDict = {}
+        self._best_score: _LGBM_BoosterBestScoreType = {}
```
jameslamb, line 523: When a variable doesn't have an explicit type hint, `mypy` will infer the type based on the first assignment of that variable.

Since these will eventually be dictionaries, initializing them to dictionaries convinces `mypy` that code which doesn't check that they're `None` is ok.

jmoralez (approved):

Issue #5682, opened 2023-01-24
`[LightGBM] [Fatal] Model format error, expect a tree here.` when using `lgb.Booster(model_str=...)`


jrvmalik:
## Description
This is a follow-up to [this issue I had](https://github.com/microsoft/LightGBM/issues/5609) where I wasn't able to use the .refit function to set the leaf scores of a booster.  I am going about it manually.  I have a new set of leaf scores that I would like to insert into an existing booster.  I think the only way to do this is to use the 
```
lgb.Booster.model_to_string
lgb.Booster(model_str=...) 
```
methods, where I can manually change the lines beginning with 
```
leaf_value=
```
in the string representation of the tree. However I get the following error when trying to do this. 
```
[LightGBM] [Fatal] Check failed: leaf_idx >= 0 && leaf_idx < models_[tree_idx]->num_leaves() at /home/conda/feedstock_root/build_artifacts/lightgbm_1641599973643/work/compile/src/boosting/gbdt.h, line 380 .

[LightGBM] [Fatal] Check failed: leaf_idx >= 0 && leaf_idx < models_[tree_idx]->num_leaves() at /home/conda/feedstock_root/build_artifacts/lightgbm_1641599973643/work/compile/src/boosting/gbdt.h, line 380 .

[LightGBM] [Fatal] Model format error, expect a tree here. met ree=1
terminate called after throwing an instance of 'std::runtime_error'
  what():  Model format error, expect a tree here. met ree=1
[LightGBM] [Warning] Model format error, expect a tree here. met ree=1

Process finished with exit code 134 (interrupted by signal 6: SIGABRT)
```
I looked around and found that some people have issues with the newline character, but wasn't able to use that information to fix the error. 

## Reproducible example
```
from collections import defaultdict
import copy

import lightgbm as lgb
import numpy as np

n = 30000  # number of data points
p = 200  # number of features
m = 2  # number of trees
lambda_l1 = 0.001
min_child_weight = 1
learning_rates = defaultdict(lambda: 1 / m)

X = np.random.randn(n, p)
y = np.random.randn(n)
y_new = np.random.randn(n)

base_train_data = lgb.Dataset(X, label=y, init_score=np.zeros_like(y))
base_validation_data = lgb.Dataset(X, label=y, init_score=np.zeros_like(y), params=base_train_data.params, reference=base_train_data)
train_data = lgb.Dataset(X, label=y_new, init_score=np.zeros_like(y_new))
[...]

PR #5712, opened 2023-02-14
[python-package] add Booster.set_leaf_output method

jmoralez:
Adds the `Booster.set_leaf_output` method to the python API.

Based on #5682.


jameslamb (approved):
Makes sense to me! I didn't even know about `LGBM_BoosterSetLeafValue()` until I saw your answer in #5682, thanks for that.

PR #5714, opened 2023-02-14
[python-package] fix mypy errors about _InnerPredictor

jameslamb:
Contributes to #3756.
Contributes to #3867.

Fixes the following errors from `mypy`

```text
python-package/lightgbm/basic.py:2117: error: Incompatible types in assignment (expression has type "List[int]", variable has type "None")  [assignment]
python-package/lightgbm/basic.py:2325: error: Item "None" of "Optional[_InnerPredictor]" has no attribute "current_iteration"  [union-attr]
python-package/lightgbm/basic.py:2328: error: Incompatible types in assignment (expression has type "_InnerPredictor", variable has type "None")  [assignment]
```

To make the flow of data through calls to internal function `_set_init_score_by_predictor()`  a bit easier to follow and less susceptible to "forgot to pass through an argument" issues, this PR also removes default values from that function's signature and changes all calls of it to use keyword arguments.

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -1375,9 +1375,9 @@ class Dataset:
         self.categorical_feature = categorical_feature
         self.params = deepcopy(params)
         self.free_raw_data = free_raw_data
-        self.used_indices = None
+        self.used_indices: Optional[List[int]] = None
         self.need_slice = True
-        self._predictor = None
+        self._predictor: Optional[_InnerPredictor] = None
         self.pandas_categorical = None
         self.params_back_up = None
         self.monotone_constraints = None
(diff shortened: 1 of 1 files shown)


python-package/lightgbm/basic.py
```diff
@@ -2323,16 +2336,27 @@ def _set_predictor(
         It is not recommended for user to call this function.
         Please use init_model argument in engine.train() or engine.cv() instead.
         """
-        if predictor is self._predictor and (predictor is None or predictor.current_iteration() == self._predictor.current_iteration()):
```
jameslamb, line 2326: This slight refactoring is to work around this `mypy` error:

> python-package/lightgbm/basic.py:2325: error: Item "None" of "Optional[_InnerPredictor]" has no attribute "current_iteration"  [union-attr]

I can't see a path through the code that could result in those `.current_iteration()` calls being run if either `predictor` or `self._predictor` was `None`. I suspect that maybe comparing values to `None` with `is` is treated specially in a way that results in `mypy` not being confident about that.

I think the proposed rewrite is easier for both humans and `mypy` to understand.

guolinke (approved):

PR #5720, opened 2023-02-16
[python-package] fix mypy error about log callback

jameslamb:
Contributes to #3867.

```text
python-package/lightgbm/basic.py:149: error: "CDLL" has no attribute "callback"  [attr-defined]
```

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -146,7 +146,7 @@ def _load_lib() -> ctypes.CDLL:
     lib = ctypes.cdll.LoadLibrary(lib_path[0])
     lib.LGBM_GetLastError.restype = ctypes.c_char_p
     callback = ctypes.CFUNCTYPE(None, ctypes.c_char_p)
-    lib.callback = callback(_log_callback)
+    lib.callback = callback(_log_callback)  # type: ignore[attr-defined]
     if lib.LGBM_RegisterLogCallback(lib.callback) != 0:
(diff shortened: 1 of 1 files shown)


python-package/lightgbm/basic.py
```diff
@@ -146,7 +146,7 @@ def _load_lib() -> ctypes.CDLL:
     lib = ctypes.cdll.LoadLibrary(lib_path[0])
     lib.LGBM_GetLastError.restype = ctypes.c_char_p
     callback = ctypes.CFUNCTYPE(None, ctypes.c_char_p)
-    lib.callback = callback(_log_callback)
+    lib.callback = callback(_log_callback)  # type: ignore[attr-defined]
```
jameslamb, line 149: Here's the corresponding C++ code using this:

https://github.com/microsoft/LightGBM/blob/29796eee6eabb79e161880976ec02fcd95dfcc4a/include/LightGBM/utils/log.h#L141-L154

Assigning to a new attribute on `lib` (which is a `CDLL` object holding a pointer to the loaded shared object, created with `dlopen`) effectively is like adding a new function in the shared library, I guess? 

I'm not 100% sure, but I do know that when I removed this assignment and just changed it to something like this:

```python
_lib_callback = callback(_log_callback)
if lib.LGBM_RegisterLogCallback(_lib_callback) != 0:
# ...
```

The unit tests produced segfaults immediately.

guolinke (approved):

PR #5723, opened 2023-02-17
[python-package] make some Booster and Dataset attributes private

jameslamb:
Contributes to #5313.

Proposes prefixing the following attributes with `_`, to indicate that user code isn't intended to rely on them.

* `Dataset.need_slice`
* `Dataset.params_back_up`
* `Booster.network`

And removing one unused attribute.

* `Dataset.monotone_constraints`

You can see that it's unused by running the following:

```shell
git grep -E '\.monotone_constraints'
```

### Notes for Reviewers

See https://github.com/microsoft/LightGBM/issues/5313#issuecomment-1368084618 for reasons supporting this. That comment mentions a few other properties, but I think making those private will be a bit more controversial and generate more discussion, so I'd like to defer them to a later PR.

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -1376,11 +1376,10 @@ class Dataset:
         self.params = deepcopy(params)
         self.free_raw_data = free_raw_data
         self.used_indices: Optional[List[int]] = None
-        self.need_slice = True
+        self._need_slice = True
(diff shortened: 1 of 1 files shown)


guolinke (approved):

jameslamb:
Thanks for the review, @guolinke !

I'm very happy we're simplifying the public interface with changes like this in the 4.0.0 release 😊

PR #5730, opened 2023-02-20
[python-package] remove default values in internal functions

jameslamb:
Proposes removing default argument values in the following internal functions:

* `Dataset.__init_from_seqs()`
* `Booster.__inner_eval()`
* `_format_eval_result()`
* `_make_n_folds()`
* `_check_not_tuple_of_2_elements()`
* `_float2str()`

Removing defaults helps to prevent "forgot to pass arguments through" types of bugs.

### Notes for Reviewers

This PR is just about changing reliance on default values, and isn't intended to actually change LightGBM's behavior. If you see any places with a user-facing change, please note them so I can revert them.

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -1784,7 +1784,7 @@ class Dataset:
     def __init_from_seqs(
         self,
         seqs: List[Sequence],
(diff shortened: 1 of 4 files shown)


guolinke (approved):
Thank you, looks good to me

PR #5738, opened 2023-02-23
Modifying brackets

agurimon:
) -> ]


jameslamb (changes requested):
Thanks for taking the time to contribute! But I don't think the brackets in this log message are incorrect.

For 3-class classification, for example, the 3 classes will have integer representations `0, 1, 2`. It's common in interval notation to use `)` to mean "excluding the final item in the interval".

So `[0, 3)` correctly refers to the set `0, 1, 2`.

See https://stackoverflow.com/a/4396303/3986677 and https://en.wikipedia.org/wiki/Bracket_(mathematics)#Intervals for more examples.

PR #5713, opened 2023-02-14
[ci] fix flaky dask tests on GPU

jameslamb:
Opening this to try to narrow down the flakiness in the GPU tests.

Reference https://github.com/microsoft/LightGBM/pull/5677#issuecomment-1409170835.

PR #5741, opened 2023-02-25
[ci] skip Dask tests on GPU builds

jameslamb:
Reverts the changes from #5292.

Since I merged that PR, the `test_dask.py` tests have been VERY flaky on GPU builds. My unscientific estimate is that 60% of commits since then have had at least one Azure DevOps GPU job fail.

This PR proposes skipping those tests again, to reduce the required manual effort of re-running jobs and to improve the contribution experience for outside contributors. 

I'm using #5713 to test fixes, but for now I think we should just skip these flaky tests again 😞


jmoralez (approved):

jameslamb:
thanks for the quick review @jmoralez 1

PR #5745, opened 2023-02-26
[python-package] add type hints on Dataset feature processing

jameslamb:
Contributes to #3756.

Adds type hints on `_data_from_pandas()` and `_dump_pandas_categorical()`, internal functions used during `Dataset` construction.

Fixes return type hint on `_load_pandas_categorical()`, which is used to recover information about `pandas` categorical columns from model text files.

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -33,6 +33,8 @@ _DatasetHandle = ctypes.c_void_p
 _LGBM_EvalFunctionResultType = Tuple[str, float, bool]
 _LGBM_BoosterBestScoreType = Dict[str, Dict[str, float]]
 _LGBM_BoosterEvalMethodResultType = Tuple[str, str, float, bool]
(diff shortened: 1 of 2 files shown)


python-package/lightgbm/basic.py
```diff
@@ -650,7 +660,7 @@ def _dump_pandas_categorical(pandas_categorical, file_name=None):
 def _load_pandas_categorical(
     file_name: Optional[Union[str, Path]] = None,
     model_str: Optional[str] = None
-) -> Optional[str]:
+) -> Optional[List[List]]:
```
jameslamb, line 663: this returns the output of `json.loads()` called on the `pandas: categorical` line of a model text file

https://github.com/microsoft/LightGBM/blob/0b937d1074fecd91243cad7a8bc5b99aaac20996/python-package/lightgbm/basic.py#L674

and that line will contain data that `json.loads()` reads in as a list of lists.

guolinke (approved):

PR #5746, opened 2023-02-26
WIP: [ci] include lightgbm's dependencies in linting job

jameslamb:
Contributes to https://github.com/microsoft/LightGBM/issues/3756

The `lint` CI job runs `mypy`, a static type-checker for Python.

The code in `python-package/compat.py` handles optional imports by creating placeholder empty classes and functions. For example, like this:

https://github.com/microsoft/LightGBM/blob/39ed8ea2980c0c58ef9d1512be4ab2c85840126a/python-package/lightgbm/compat.py#L25

Since many of `lightgbm`'s dependencies (like `pandas` and `matplotlib`) are not currently installed during the `lint` CI job, it's those placeholder definitions that `mypy` checks. And the presence of `# type: ignore` comments on many of them short-circuits the type-checking process.

As a result, the `lint` job is currently not able to detect issues like "pandas Series do not have attribute `.xyz`".

This PR proposes installing all of those dependencies in the `lint` CI job, to improve the ability of CI to catch such incompatibilities (at least with the latest version of all installed dependencies).


jameslamb:
This proved to be more difficult than I thought it would be. I'm going to close this to focus on other things. Sorry for the noise.

PR #5752, opened 2023-02-28
[python-package] add type hints on raw data passed to Dataset and Booster

jameslamb:
Contributes to #3756.

Adds type annotations on `data`, raw data passed to the `Dataset` constructor.

### Notes for Reviewers

I created this based on this docstring:

https://github.com/microsoft/LightGBM/blob/356a7806882c4aed4f54689d5efe29ae05d5bd6f/python-package/lightgbm/basic.py#L850

This added about 30 new `mypy` warnings. I'm proposing fixing those separately from this PR. There is just SIGNIFICANT complexity here given that `data` can be a path to a file, any of several different array libraries' data types, or a `Sequence` object. I think it'll be easier to work on and fix those new `mypy` warnings in smaller pieces after adding these hints.

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -35,6 +35,17 @@ _LGBM_BoosterBestScoreType = Dict[str, Dict[str, float]]
 _LGBM_BoosterEvalMethodResultType = Tuple[str, str, float, bool]
 _LGBM_CategoricalFeatureConfiguration = Union[List[str], List[int], str]
 _LGBM_FeatureNameConfiguration = Union[List[str], str]
+_LGBM_TrainDataType = Union[
+    str,
+    Path,
+    np.ndarray,
+    pd_DataFrame,
+    dt_DataTable,
+    scipy.sparse.spmatrix,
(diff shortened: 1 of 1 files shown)


python-package/lightgbm/basic.py
```diff
@@ -4034,7 +4045,7 @@ def refit(
 
         Parameters
         ----------
-        data : str, pathlib.Path, numpy array, pandas DataFrame, H2O DataTable's Frame or scipy.sparse
+        data : str, pathlib.Path, numpy array, pandas DataFrame, H2O DataTable's Frame, scipy.sparse, Sequence, list of Sequence or list of numpy array
```
jameslamb, line 4048: `data` passed through `Booster.refit()` is just directly passed through to the `Dataset` constructor.

https://github.com/microsoft/LightGBM/blob/356a7806882c4aed4f54689d5efe29ae05d5bd6f/python-package/lightgbm/basic.py#L4104-L4105

So it should have the same type hint and docstring as in `Dataset`.

guolinke (approved):

PR #5756, opened 2023-02-28
[python-package] [dask] fix mypy errors about Dask fit() return types

jameslamb:
Contributes to #3867.

Resolves the following `mypy` errors.

```text
python-package/lightgbm/dask.py:1178: error: Incompatible return value type (got "_DaskLGBMModel", expected "DaskLGBMClassifier")  [return-value]
python-package/lightgbm/dask.py:1381: error: Incompatible return value type (got "_DaskLGBMModel", expected "DaskLGBMRegressor")  [return-value]
python-package/lightgbm/dask.py:1553: error: Incompatible return value type (got "_DaskLGBMModel", expected "DaskLGBMRanker")  [return-value]
```

Switches the Dask estimators to returning `self`, so that e.g. `DaskLGBMRegressor.fit()` returns a `DaskLGBMRegressor`. This is useful for other user code and downstream libraries relying on LightGBM, to be able to do `isinstance()` checks and not need to rely on lightgbm-internal details like `_DaskLGBMModel`. It could also be relevant in the future is we add attributes to `DaskLGBMRegressor`, `DaskLGBMClassifier`, or `DaskLGBMRanker` that the other Dask estimators don't have.

It is, by the way, exactly the same approach the `sklearn` estimators use:

https://github.com/microsoft/LightGBM/blob/e666a4b412e00b1f9d8e3639e0896a271c0f0cd9/python-package/lightgbm/sklearn.py#L989-L1024

Related conversation: https://github.com/microsoft/LightGBM/pull/5672#discussion_r1107636706

### DIFF
--- a/python-package/lightgbm/dask.py
+++ b/python-package/lightgbm/dask.py
@@ -1175,7 +1175,7 @@ class DaskLGBMClassifier(LGBMClassifier, _DaskLGBMModel):
         **kwargs: Any
     ) -> "DaskLGBMClassifier":
         """Docstring is inherited from the lightgbm.LGBMClassifier.fit."""
-        return self._lgb_dask_fit(
+        self._lgb_dask_fit(
             model_factory=LGBMClassifier,
             X=X,
             y=y,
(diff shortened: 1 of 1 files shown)


guolinke (approved):

jameslamb:
@jmoralez what do you think about this one?

(no rush)

jmoralez (approved):

Issue #4788, opened 2021-11-09
LightGBM wheel on PyPI seems to have the wrong platform tag


tdoublep:
## Description

I'm having problems trying to use LightGBM as part of our a build process that runs inside a [manylinux2010](https://github.com/pypa/manylinux) container. In particular I am seeing the following error:

```
OSError: /lib64/libc.so.6: version `GLIBC_2.14' not found (required by /opt/python/cp37-cp37m/lib/python3.7/site-packages/lightgbm/lib_lightgbm.so)
```

Looking on PyPI, I see that LightGBM is publishing wheels with the `manylinux1` platform tag, which is what pip is pulling down in the above environment. Since `manylinux1` is based on older glibc than `manylinux2010`, and glibc is forward compatible, I was expecting this to work fine. Some deeper investigation (see below) seems to suggest that the wheel may have the wrong platform tag. 

This problem is not blocking us, since I can install using the `--no-binary` option and it works fine, but I thought it could be helpful to report this. 

Please let me know if any more information could help! Cheers! 

## Reproducible example

Some additional inspection of the LightGBM `manylinux1` wheel using the `auditwheel` tool (`pip install auditwheel`) shows:
```
$ auditwheel show lightgbm-3.3.1-py3-none-manylinux1_x86_64.whl 

lightgbm-3.3.1-py3-none-manylinux1_x86_64.whl is consistent with the
following platform tag: "linux_x86_64".

The wheel references external versioned symbols in these system-
provided shared libraries: libgcc_s.so.1 with versions {'GCC_3.0'},
libpthread.so.0 with versions {'GLIBC_2.3.4', 'GLIBC_2.2.5'},
libgomp.so.1 with versions {'OMP_1.0', 'GOMP_1.0'}, libc.so.6 with
versions {'GLIBC_2.3', 'GLIBC_2.6', 'GLIBC_2.17', 'GLIBC_2.14',
'GLIBC_2.2.5', 'GLIBC_2.3.4', 'GLIBC_2.4'}, libm.so.6 with versions
{'GLIBC_2.2.5'}, libstdc++.so.6 with versions {'CXXABI_1.3',
'GLIBCXX_3.4', 'GLIBCXX_3.4.19', 'GLIBCXX_3.4.9', 'GLIBCXX_3.4.14',
'CXXABI_1.3.5', 'GLIBCXX_3.4.11', 'GLIBCXX_3.4.18', 'CXXABI_1.3.7',
[...]

PR #5759, opened 2023-03-01
[python-package] [ci] switch to PEP 517 / 518 builds (remove `setup.py`) (fixes #5061)

### LINKED ISSUE #5061: [RFC] [python] 'setup.py install' is deprecated

jameslamb:
Fixes #5061.

This PR proposes removing the project's `setup.py`, and instead using `scikit-build-core` (https://github.com/scikit-build/scikit-build-core) as a build backend.

`scikit-build-core` has been developed with support from an NSF grant ([link](https://www.nsf.gov/awardsearch/showAward?AWD_ID=2209877&HistoricalAwards=false)) and is being developed by the same team developing `scikit-build`, which is used by other popular stats and machine learning projects, like:

* `cudf` (https://github.com/rapidsai/cudf/pull/10919)
* `cuml` (https://github.com/rapidsai/cuml/pull/4818)
* `opencv` ([code link](https://github.com/opencv/opencv-python/blob/474a1cc0ebf2086c596b60c050a9e1af658ff380/pyproject.toml#L4))
* `vaex` (https://github.com/vaexio/vaex/pull/1847)

*(I got that list from https://github.com/orgs/scikit-build/discussions/976)*

`scikit-build-core` is purpose-built for Python packages that use `CMake` to compile one or more other artifacts (in our case, the `lib_lightgbm` shared library). 

### What would happen if we did nothing?

Users would have to pin to older versions of `setuptools` and `pip` to be able to install `lightgbm` (which will make `lightgbm` harder to use alongside other packages). We've already started seeing evidence of that pain:

* #5897
* https://github.com/microsoft/LightGBM/issues/5061#issuecomment-1274724145

### Benefits of this approach

* users can use the newest versions of `pip`, `setuptools`, and `build` in environments where they install `lightgbm`
* no more need to maintain 400ish lines of Python code in `setup.py`
* makes it easier to compile wheels for more platforms in the future (e.g. MUSL-based wheels for Alpine Linux, `macosx-arm64` for the M1/M2 Macs)

### What other approaches could we take?

* could use `hatchling` like `xgboost` has recently done (see https://github.com/dmlc/xgboost/pull/9021#discussion_r1173171975)
    - *this has the advantage of supporting most of what `lightgbm` packaging currently supports, at the expense of needing to still maintain significant Python glue code*
* could use `scikit-build` (`scikit-build-core`'s predecessor)
    - *I don't think we should do this, as https://github.com/scikit-build/scikit-build says refers to `scikit-build-core` as "The next generation of scikit-build," and the author of both tools referred to `scikit-build` as "scikit-build classic" in https://github.com/orgs/scikit-build/discussions/815... implying that `scikit-build-core` is where most active development is going*
* could stop shipping source distributions entirely and use just a shell script to build wheels
    - *`catboost` was doing this until very recently*
* could switch build toolchains entirely, away from CMake to something like `meson`, `autotools`, `bazel`, etc. and the corresponding options for Python build backends
    - `numpy`, for example, is in the process of switching to `meson-python`: https://github.com/numpy/numpy/pull/23838

These are all possible routes to resolve #5061.

I'm proposing using `scikit-build-core` because it's powerful while also significantly simplifying the code that we have to manage in this project, and because it's maintained by members of PyPA who will keep it up to date with the latest changes in Python packaging (i.e. the latest accepted PEPs related to packaging).

### Existing features preserved with this approach

<details><summary>list of features (click me)</summary>

All of the follow commands are currently supported, and we should try to preserve as many of them as possible.

* `{command}` is `python setup.py` right now
* `{dist}` is either the name `lightgbm` (pull from PyPI) or a path to a `*.tar.gz` or `*.whl`

Installing from source tree.

```shell
{command} install
{command} install --precompile
{command} install --precompile --user
{command} install --mingw
{command} install --mpi
{command} install --nomp
{command} install --gpu
{command} install --gpu --opencl-include-dir=/usr/local/cuda/include/
{command} install --cuda
{command} install --hdfs
{command} install --bit32
```

Building an sdist.

```shell
{command} sdist
{command} sdist --formats gztar
```

Building wheels.

```shell
{command} bdist_wheel --plat-name=macosx --python-tag py3
{command} bdist_wheel --integrated-opencl --plat-name=$PLATFORM --python-tag py3
{command} bdist_wheel --integrated-opencl --plat-name=win-amd64 --python-tag py3
{command} bdist_wheel --gpu
{command} bdist_wheel --cuda
{command} bdist_wheel --mpi
```

Customized installation of an sdist (possibly from PyPI).

```shell
pip install {dist} --install-option=--nomp
pip install {dist} --install-option=--mpi
pip install {dist} --install-option=--gpu
pip install {dist} --install-option=--gpu --install-option="--opencl-include-dir=/usr/local/cuda/include/" --install-option="--opencl-library=/usr/local/cuda/lib64/libOpenCL.so"
pip install {dist} --install-option=--cuda
pip install {dist} --install-option=--hdfs
pip install {dist} --install-option=--mingw
pip install {dist} --install-option=--bit32
pip install {dist} --install-option=--time-costs
pip install {dist}[dask]
pip install --no-binary `:all:` {dist}
pip install --user {dist}
```

</details>

### Existing features lost with this approach

* automatically redirecting CMake logs to a file ([code link](https://github.com/microsoft/LightGBM/blob/f975d3fafcdbb3739dbe4eac40dc2b7e1e3244d7/python-package/setup.py#L95-L96))
    - *CMake logs will now just be written to stdout directly, and users will have to redirect it themselves if they want to save it to a file*
    - *honestly, I think this is an improvement... no more seeing a remote build like a CI job fail without knowing the underlying CMake error*
* using `MSBuild` to compile with `pip install {sdist}`, then multiple versions of MSVC ([link](https://github.com/microsoft/LightGBM/blob/f975d3fafcdbb3739dbe4eac40dc2b7e1e3244d7/python-package/setup.py#L168-L204))
    - *looks like `scikit-build` will check for multiple versions of MSVC:* https://github.com/scikit-build/scikit-build/blob/master/docs/generators.rst
    - *as of this PR, the only way to build the Python package with `MSBuild` will be to clone the repo and invoke `MSBuild.exe` directly*
* compiling the 32-bit version with `pip install {sdist}` ([link](https://github.com/microsoft/LightGBM/blob/f975d3fafcdbb3739dbe4eac40dc2b7e1e3244d7/python-package/setup.py#L250-L251))
    - *as of this PR, the only way to do that type of build will be to clone the repo and run `sh build-python.sh install --bit32`*


.ci/test_windows.ps1
```diff
@@ -96,8 +96,8 @@ elseif ($env:TASK -eq "bdist") {
   cd $env:BUILD_SOURCESDIRECTORY
   sh "build-python.sh" bdist_wheel --integrated-opencl ; Check-Output $?
   sh $env:BUILD_SOURCESDIRECTORY/.ci/check_python_dists.sh $env:BUILD_SOURCESDIRECTORY/dist ; Check-Output $?
-  cd dist; pip install --user @(Get-ChildItem *.whl) ; Check-Output $?
-  cp @(Get-ChildItem *.whl) $env:BUILD_ARTIFACTSTAGINGDIRECTORY
+  cd dist; pip install --user @(Get-ChildItem *py3-none-win_amd64.whl) ; Check-Output $?
+  cp @(Get-ChildItem *py3-none-win_amd64.whl) $env:BUILD_ARTIFACTSTAGINGDIRECTORY
```
jameslamb, line 100: This ensures we get a loud error if the wheels are not tagged in the way we'd expect.

Since as far as I can tell, you can't pass a `--plat-name` through `python -m build` the way you could with  `python setup.py bdist_wheel --plat-name=win-amd64`.

To be clear... `scikit-build-core` is doing exactly the right thing here and producing the expected tags. This is just a precaution to help us detect if that changes, before publishing mistagged wheels to PyPI (e.g. like #4788).

python-package/README.rst
```diff
@@ -84,7 +84,7 @@ Build GPU Version
 
 .. code:: sh
 
-    pip install lightgbm --install-option=--gpu
+    pip install lightgbm --config-setting=cmake.define.USE_GPU=ON
```
hcho3, line 87: Pip exposes `--config-settings` option: https://pip.pypa.io/en/stable/cli/pip_install/#cmdoption-C. Is `--config-setting` coming from somewhere else?

jameslamb: Oh wow, thank you! No you're absolutely right, this is a mistake on my part.

I was mixing two things... `pip` has `--config-settings` (with an `s`), and `build` has `--config-setting` (no `s`) ([code link](https://github.com/pypa/build/blob/e0c59cc4e17e63cfe9dbd4f366fae6d51b073282/src/build/__main__.py#L331-L339)).

I was following from https://github.com/scikit-build/scikit-build-core/blob/a145985892e4c904f4e4f089f987b4fef491543d/README.md?plain=1#L112-L113 and didn't read closely enough.

Just pushed https://github.com/microsoft/LightGBM/pull/5759/commits/41badff4121b6003c95e8dbc8e967348e1820e96 fixing that.

build-python.sh
```diff
@@ -94,75 +94,79 @@ while [ $# -gt 0 ]; do
             then shift;
         fi
         BOOST_DIR="${1#*=}"
-        BUILD_ARGS="${BUILD_ARGS} --boost-dir='${BOOST_DIR}'"
+        BUILD_ARGS="${BUILD_ARGS} --config-setting=cmake.define.Boost_DIR='${BOOST_DIR}'"
```
guolinke, line 97: ```--config-settings``` or ```--config-setting``` in this file?

jameslamb: `--config-settings` when passed to `pip`, `--config-setting` when passed to `build` 😅 

https://github.com/microsoft/LightGBM/pull/5759#discussion_r1224471802

guolinke (approved):

hcho3 (approved):
I find it interesting that you were able to use `scikit-learn-core` without adding any custom glue code. I suppose that `build-python.sh` does much of the heavy-lifting of packaging the right files in the right place?

For now, XGBoost adds a custom PEP 517 glue code to marshal C++ source files, but in the future I may consider writing a bash script instead, should we decide to transition to `scikit-build-core`. Thanks for the inspiration!

jameslamb:
Thanks so much for the review help @hcho3 ! I know you're busy enough with the other projects you work on, I always really appreciate your help here.

> build-python.sh does much of the heavy-lifting of packaging the right files in the right place

Yep, exactly! I described my perspective on this in #5837 if you want to see a bit more detail on how I'm thinking about it.

Essentially, I think any Python user of LightGBM whose needs aren't met by `pip install` (from PyPI) or `conda install` (from conda-forge) will already need to clone the repo and run some command(s). Making that "run a shell script" should be just as easy for those people as `python setup.py install` or `pip install .`, so it's worth switching to that in exchange for simplifying the build system for the artifacts we distribute via PyPI.

I'm also hoping that having a zero-glue-code setup with `scikit-build-core` will make it easier in the future to build wheels for more platforms, further reducing the need for people to build custom from the sources here in GitHub.

We've been used a similar approach with the R package in this repo for a few years, and I think it's worked well so far: https://github.com/microsoft/LightGBM/blob/master/build-cran-package.sh.

python-package/pyproject.toml
```diff
@@ -1,3 +1,63 @@
...
+    "Development Status :: 5 - Production/Stable",
+    "Intended Audience :: Science/Research",
+    "License :: OSI Approved :: MIT License",
+    "Natural Language :: English",
+    "Operating System :: MacOS",
+    "Operating System :: Microsoft :: Windows",
+    "Operating System :: POSIX",
+    "Operating System :: Unix",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.7",
```
jmoralez, line 13: Shouldn't Python 3.6 be here given that we have `requires-python = ">=3.6"`?

jameslamb: oh sure, I can add that.

It's not there in `setup.py` and I was just copying from that.

https://github.com/microsoft/LightGBM/blob/a323c7583ccdffb7aff49e7f9ca72042aca18395/python-package/setup.py#L341-L355

jmoralez (approved):

python-package/pyproject.toml
guolinke: refer to https://github.com/microsoft/LightGBM/pull/4891

jameslamb: Ah right, thanks @guolinke for reminding me ... in #4891 we dropped support, meaning:

* dropped that trove classifier `'Programming Language :: Python :: 3.6'` from the PyPI labels
* do not have any CI jobs testing Python 3.6

And we decided not to move up to `requires-python = ">=3.7"` so that people could *technically* try to build for Python 3.6 if they needed that a bit longer (as @jmoralez also referred to in https://github.com/microsoft/LightGBM/pull/5765#issuecomment-1454861675).

By the way, I tried tonight and it looks like `lightgbm` is still at least minimally compatible with Python 3.6 🎉 .

```shell
docker run \
    --rm \
    -it python:3.6 \
    bash

apt-get update
apt-get install \
    --no-install-recommends \
    -y \
        cmake

git clone --recursive \
    https://github.com/microsoft/LightGBM.git

cd LightGBM
sh ./build-python.sh install --nomp

pip install pandas
python ./examples/python-guide/simple_example.py
```

logs:

```text
Loading data...
Starting training...
[LightGBM] [Warning] Auto-choosing col-wise multi-threading, the overhead of testing was 0.002058 seconds.
Training until validation scores don't improve for 5 rounds
Did not meet early stopping. Best iteration is:
[20]	valid_0's l2: 0.198063	valid_0's l1: 0.431129
Saving model...
Starting predicting...
The RMSE of prediction is: 0.4450426449744025
```

I don't feel strongly either way about whether or not to have the Python 3.6 trove classifier. Since @StrikerRUS removed it in #4891, I think we should defer to his judgment and continue to omit it. Is that ok @jmoralez ? If, after reading this, you still feel strongly that it should be included since `lightgbm` at least *minimally* supports Python 3.6, I'll add it.

jmoralez: That's ok, I agree with not adding it back.

jameslamb: Alright, thanks for talking through it with us! In another PR, I might put up a minimal CI job that runs exactly that dockerized little script I just shared above. So we can rely on CI to catch issues of the form "this PR would be a hard compatibility break because it's not even syntactically valid Python 3.6".

I don't want to take on any more support than that for Python 3.6, but would be useful (I think) to get automated feedback at that level at least, and I'd expect that job shouldn't require too much maintenance burden.

jameslamb: For those finding this conversation from search... here's the PR I promised: #5936.

PR #5765, opened 2023-03-04
[python-package] drop Python 3.6 support, use dataclasses

jameslamb:
Contributes to #3756.
Contributes to #3867.
Inspired by discussions in #5672 (thanks @IdoKendo !).

This PR proposes the following:

* replacing two uses of `collections.namedtuple` in the Python package with `dataclasses.dataclass`
* dropping support for Python 3.6

### Why use `dataclasses`?

`dataclasses` are preferable to `namedtuple` in the following ways:

* easier to set type annotations + default values than named tuples
    - makes it easier for both human developers and and tools like `mypy` to understand the expected types of properties
* possible to attach other methods or computed properties with `@property` (just like normal classes)

### Why drop support for Python 3.6?

`dataclasses` has been a part of the Python standard library since Python 3.7 (https://docs.python.org/3/library/dataclasses.html).

Python 3.6 reached end-of-life (stopped receiving even security fixes) in December 2021 (https://devguide.python.org/versions/).

Most of `lightgbm`'s dependencies dropped Python 3.6 a long time ago, for example:

* `pandas` in August 2020: https://github.com/pandas-dev/pandas/pull/35214
* `dask` and `distributed` in December 2020: https://github.com/dask/community/issues/66
* `numpy` in December 2020: https://github.com/numpy/numpy/pull/17898

### DIFF
--- a/python-package/lightgbm/callback.py
+++ b/python-package/lightgbm/callback.py
@@ -3,9 +3,13 @@
 import collections
 from dataclasses import dataclass
 from functools import partial
-from typing import Any, Callable, Dict, List, Optional, Tuple, Union
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Tuple, Union
 
 from .basic import Booster, _ConfigAliases, _LGBM_BoosterEvalMethodResultType, _log_info, _log_warning
+
+if TYPE_CHECKING:
+    from .engine import CVBooster
+
 
 __all__ = [
     'early_stopping',
@@ -42,7 +46,7 @@ class EarlyStopException(Exception):
 # Callback environment used by callbacks
 @dataclass
 class CallbackEnv:
-    model: Booster
+    model: Union[Booster, "CVBooster"]
     params: Dict[str, Any]
     iteration: int
     begin_iteration: int
--- a/python-package/lightgbm/callback.py
+++ b/python-package/lightgbm/callback.py
@@ -1,10 +1,11 @@
 # coding: utf-8
 """Callbacks library."""
 import collections
+from dataclasses import dataclass
 from functools import partial
-from typing import Any, Callable, Dict, List, Tuple, Union
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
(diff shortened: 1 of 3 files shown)


python-package/lightgbm/callback.py
```diff
@@ -1,10 +1,15 @@
...
 import collections
+from dataclasses import dataclass
 from functools import partial
-from typing import Any, Callable, Dict, List, Tuple, Union
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Tuple, Union
+
+from .basic import Booster, _ConfigAliases, _LGBM_BoosterEvalMethodResultType, _log_info, _log_warning
+
+if TYPE_CHECKING:
+    from .engine import CVBooster
```
jameslamb, line 11: See https://mypy.readthedocs.io/en/stable/runtime_troubles.html#import-cycles for an explanation of this.

Adding this pattern to avoid cyclical imports... `engine.CVBooster` is needed for the type annotation on `CallbackEnv.model`, but `engine.py` also tries to import all of `callback`'s contents.

https://github.com/microsoft/LightGBM/blob/709ea4cad3ff4c8834bdf389eef8016345d9a1ac/python-package/lightgbm/engine.py#L12

jmoralez:
Even though some dependencies have dropped support for python 3.6, (I think) we don't rely on any new functionality from them, so if the drop is only for using dataclasses I don't think it's worth it, mainly because of the reasoning described in https://github.com/microsoft/LightGBM/pull/5006#issuecomment-1039464720, I know a couple of companies that are still stuck with py36 and since 4.0.0 will have many new features I don't think users should miss out on those just for the sake of type hints on a couple of data structures.

I'm not hardly against it though, if you feel it's time and think it's worth it I won't oppose.

jameslamb:
Thanks @jmoralez ! I think you're right, I shouldn't have coupled the "drop Python 3.6" decision together with this attempt to get slightly clearer type annotations.

I'm going to close this and open some new PRs proposing these `namedtuple` with regular-old Python classes.

PR #5766, opened 2023-03-06
[python-package] [dask] add type annotations on dask._HostWorkers

jameslamb:
Contributes to #3756.
Contributes to #3867.
Split from #5765.

Proposes replacing the use of `collections.namedtuple` for one internal class in the Dask module with a regular Python class, so that informative type annotations can be added.

This makes it a bit clearer, when reading the code, what the values of that class's attributes should be (unlike `namedtuple`, which is just a collection of attribute names).

I chose to use a regular Python class instead of `dataclasses.dataclass` to preserve support for Python 3.6 (https://github.com/microsoft/LightGBM/pull/5765#issuecomment-1454861675), and chose to avoid `typing.NamedTuple` (https://stackoverflow.com/a/50038614/3986677) to reduce the risk of incompatibilities caused by future updates to `typing`.

### DIFF
--- a/python-package/lightgbm/dask.py
+++ b/python-package/lightgbm/dask.py
@@ -44,6 +44,12 @@ class _HostWorkers:
         self.default = default
         self.all_workers = all_workers
 
+    def __eq__(self, other: "_HostWorkers") -> bool:
+        return (
+            self.default == other.default
+            and self.all_workers == other.all_workers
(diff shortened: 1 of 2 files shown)


jmoralez (approved):

jameslamb:
thanks @jmoralez !

PR #5773, opened 2023-03-08
[python-package] add type hints for functions accepting dtypes

jameslamb:
Contributes to #3756.
Contributes to #3867.

Fixes the following errors from `mypy`:

```text
python-package/lightgbm/basic.py:309: error: Argument 2 to "_cast_numpy_array_to_dtype" has incompatible type "type"; expected "dtype[Any]"  [arg-type]
python-package/lightgbm/basic.py:314: error: Argument 2 to "_cast_numpy_array_to_dtype" has incompatible type "type"; expected "dtype[Any]"  [arg-type]
```

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -240,7 +240,7 @@ def _is_numpy_column_array(data: Any) -> bool:
     return len(shape) == 2 and shape[1] == 1
 
 
-def _cast_numpy_array_to_dtype(array: np.ndarray, dtype: np.dtype) -> np.ndarray:
+def _cast_numpy_array_to_dtype(array: np.ndarray, dtype: "np.typing.DTypeLike") -> np.ndarray:
     """Cast numpy array to given dtype."""
     if array.dtype == dtype:
         return array
@@ -264,7 +264,7 @@ def _is_1d_collection(data: Any) -> bool:
 
 def _list_to_1d_numpy(
     data: Any,
-    dtype=np.float32,
+    dtype: "np.typing.DTypeLike" = np.float32,
     name: str = 'list'
 ) -> np.ndarray:
     """Convert data to numpy 1-D array."""
@@ -303,7 +303,11 @@ def _is_2d_collection(data: Any) -> bool:
     )
 
 
-def _data_to_2d_numpy(data: Any, dtype: type = np.float32, name: str = 'list') -> np.ndarray:
+def _data_to_2d_numpy(
+    data: Any,
+    dtype: "np.typing.DTypeLike" = np.float32,
+    name: str = 'list'
+) -> np.ndarray:
     """Convert data to numpy 2-D array."""
     if _is_numpy_2d_array(data):
         return _cast_numpy_array_to_dtype(data, dtype)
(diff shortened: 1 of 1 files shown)


python-package/lightgbm/basic.py
```diff
@@ -240,7 +240,7 @@ def _is_numpy_column_array(data: Any) -> bool:
     return len(shape) == 2 and shape[1] == 1
 
 
-def _cast_numpy_array_to_dtype(array: np.ndarray, dtype: np.dtype) -> np.ndarray:
+def _cast_numpy_array_to_dtype(array: np.ndarray, dtype: "np.typing.DTypeLike") -> np.ndarray:
```
jameslamb, line 243: these annotations come from https://numpy.org/devdocs/reference/typing.html.

`numpy.typing` is new as of `numpy` 1.20 ([January 2021](https://pypi.org/project/numpy/#history)).

Expressing these annotations as string literals means that they won't cause issues at runtime for users using `lightgbm` + older `numpy` versions. See https://mypy.readthedocs.io/en/stable/runtime_troubles.html#import-cycles.

python-package/lightgbm/basic.py
```diff
@@ -612,7 +616,7 @@ def _c_int_array(data):
     return (ptr_data, type_data, data)  # return `data` to avoid the temporary copy is freed
 
 
-def _is_allowed_numpy_dtype(dtype) -> bool:
+def _is_allowed_numpy_dtype(dtype: type) -> bool:
```
jameslamb, line 619: why is this one just `type`?

Setting it to `np.typing.DTypeLike` results in these `mypy` errors.

```text
python-package/lightgbm/basic.py:618: error: Argument 1 to "issubclass" has incompatible type "Union[dtype[Any], None, Type[Any], _SupportsDType[dtype[Any]], str, Union[Tuple[Any, int], Tuple[Any, Union[SupportsIndex, Sequence[SupportsIndex]]], List[Any], _DTypeDict, Tuple[Any, Any]]]"; expected "type"  [arg-type]
python-package/lightgbm/basic.py:619: error: Argument 1 to "issubclass" has incompatible type "Union[dtype[Any], None, Type[Any], _SupportsDType[dtype[Any]], str, Union[Tuple[Any, int], Tuple[Any, Union[SupportsIndex, Sequence[SupportsIndex]]], List[Any], _DTypeDict, Tuple[Any, Any]]]"; expected "type"  [arg-type]
```

And that's because it's not being passed a `dtype`... it's being passed a `type`.

https://github.com/microsoft/LightGBM/blob/1585ee1566631fca55cd3a391074f807b21bf9b5/python-package/lightgbm/basic.py#L280

https://github.com/microsoft/LightGBM/blob/1585ee1566631fca55cd3a391074f807b21bf9b5/python-package/lightgbm/basic.py#L624-L628

```python
import pandas as pd
s = pd.Series([1, 2, 3, 4, 5], dtype="Int64")
for column_name, pandas_dtype in s.to_frame().dtypes.items():
    print(pandas_dtype)
    print(type(pandas_dtype))

# Int64
# <class 'pandas.core.arrays.integer.Int64Dtype'>
```

jmoralez (approved):

PR #5774, opened 2023-03-08
[python-package] ignore some mypy errors in _DaskLGBMModel

jameslamb:
Contributes to #3867.

Resolves the following `mypy` errors.

```text
python-package/lightgbm/dask.py:1037: error: "_DaskLGBMModel" has no attribute "_other_params"  [attr-defined]
python-package/lightgbm/dask.py:1066: error: "_DaskLGBMModel" has no attribute "get_params"  [attr-defined]
python-package/lightgbm/dask.py:1089: error: "_DaskLGBMModel" has no attribute "set_params"  [attr-defined]
python-package/lightgbm/dask.py:1095: error: "_DaskLGBMModel" has no attribute "get_params"  [attr-defined]
python-package/lightgbm/dask.py:1104: error: Item "_DaskLGBMModel" of "Union[_DaskLGBMModel, LGBMModel]" has no attribute "get_params"  [union-attr]
```

Those errors all come from the fact that `_DaskLGBMModel` is used as a mixin alongside `lightgbm.sklearn.LGBMModel`...

https://github.com/microsoft/LightGBM/blob/1585ee1566631fca55cd3a391074f807b21bf9b5/python-package/lightgbm/dask.py#L1111

... and references methods and attributes that it expects to come from `lightgbm.sklearn.LGBMModel`

https://github.com/microsoft/LightGBM/blob/1585ee1566631fca55cd3a391074f807b21bf9b5/python-package/lightgbm/dask.py#L1098

There is in theory a way to handle this situation with `typing.Protocol` (https://mypy.readthedocs.io/en/stable/more_types.html#mixin-classes), but I couldn't figure it out tonight. So for now, given that these methods are all covered well by the many unit tests, this PR proposes just ignoring these errors from `mypy`.

### DIFF
--- a/python-package/lightgbm/dask.py
+++ b/python-package/lightgbm/dask.py
@@ -1034,7 +1034,7 @@ class _DaskLGBMModel:
     def _lgb_dask_getstate(self) -> Dict[Any, Any]:
         """Remove un-picklable attributes before serialization."""
         client = self.__dict__.pop("client", None)
-        self._other_params.pop("client", None)
+        self._other_params.pop("client", None)  # type: ignore[attr-defined]
         out = deepcopy(self.__dict__)
         out.update({"client": None})
         self.client = client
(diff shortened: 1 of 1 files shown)


jmoralez (approved):

jameslamb:
thanks for the reviews on all these @jmoralez ! I swear we're getting closer to having everything type-annotated and fully enabling `mypy` 😅

PR #5782, opened 2023-03-11
[python-package] fix mypy error about dask._HostWorkers.__eq__()

jameslamb:

```text
python-package/lightgbm/dask.py:47: error: Argument 1 of "__eq__" is incompatible with supertype "object"; supertype defines the argument type as "object"  [override]
python-package/lightgbm/dask.py:47: note: This violates the Liskov substitution principle
python-package/lightgbm/dask.py:47: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
python-package/lightgbm/dask.py:47: note: It is recommended for "__eq__" to work with arbitrary objects, for example:
python-package/lightgbm/dask.py:47: note:     def __eq__(self, other: object) -> bool:
python-package/lightgbm/dask.py:47: note:         if not isinstance(other, _HostWorkers):
python-package/lightgbm/dask.py:47: note:             return NotImplemented
python-package/lightgbm/dask.py:47: note:         return <logic to compare two _HostWorkers instances>
```

### DIFF
--- a/python-package/lightgbm/dask.py
+++ b/python-package/lightgbm/dask.py
@@ -44,9 +44,10 @@ class _HostWorkers:
         self.default = default
         self.all_workers = all_workers
 
-    def __eq__(self, other: "_HostWorkers") -> bool:
+    def __eq__(self, other: object) -> bool:
         return (
-            self.default == other.default
+            isinstance(other, type(self))
+            and self.default == other.default
             and self.all_workers == other.all_workers
         )
 


jmoralez:
I just learned about [typing.NamedTuple](https://docs.python.org/3.6/library/typing.html#typing.NamedTuple). We can have the `_HostWorkers` class inherit from that and we wouldn't need the `__eq__`, it works the same way as dataclass with some slight differences [stackoverflow reference](https://stackoverflow.com/questions/51671699/data-classes-vs-typing-namedtuple-primary-use-cases).

jameslamb:
I mentioned that in the description of #3756 

> I chose to use a regular Python class instead of dataclasses.dataclass to preserve support for Python 3.6 (https://github.com/microsoft/LightGBM/pull/5765#issuecomment-1454861675), and chose to avoid typing.NamedTuple (https://stackoverflow.com/a/50038614/3986677) to reduce the risk of incompatibilities caused by future updates to typing.

I'm nervous about introducing a dependency on `typing` for runtime (not just type-checking) behavior.

jameslamb:
sorry,I meant #5766

jmoralez:
Oh, sorry

jmoralez (approved):

jameslamb:
No problem at all! I know having to also have this `__eq__` feels like a lot just for slightly better type-checking on two constants. In a future where we drop Python 3.6 support, this definitely should be converted to a `dataclasses.dataclass` (but to be clear, I still agree with you that something bigger needs to happen to cause us to drop Python 3.6 support).

PR #5793, opened 2023-03-19
[python-package] fix type annotations for eval result tracking

jameslamb:
Contributes to #3756.
Contributes to #3867.
Follow-up to this discussion with @IdoKendo : https://github.com/microsoft/LightGBM/pull/5672#discussion_r1121869086

Fixes the following `mypy` errors.

```text
callback.py:341: error: Argument 1 to "_format_eval_result" has incompatible type "Union[Tuple[str, str, float, bool], Tuple[str, str, float, bool, float], Any]"; expected "Union[List[Tuple[str, str, float, bool]], List[Tuple[str, str, float, bool, float]]]"  [arg-type]
callback.py:341: error: Item "None" of "Optional[Union[List[Tuple[str, str, float, bool]], List[Tuple[str, str, float, bool, float]]]]" has no attribute "__iter__" (not iterable)  [union-attr]
callback.py:346: error: Argument 2 to "EarlyStopException" has incompatible type "Optional[Union[List[Tuple[str, str, float, bool]], List[Tuple[str, str, float, bool, float]]]]"; expected "Union[List[Tuple[str, str, float, bool]], List[Tuple[str, str, float, bool, float]]]"  [arg-type]
callback.py:367: error: Argument 1 to "_format_eval_result" has incompatible type "Union[Tuple[str, str, float, bool], Tuple[str, str, float, bool, float], Any]"; expected "Union[List[Tuple[str, str, float, bool]], List[Tuple[str, str, float, bool, float]]]"  [arg-type]
callback.py:367: error: Item "None" of "Optional[Union[List[Tuple[str, str, float, bool]], List[Tuple[str, str, float, bool, float]]]]" has no attribute "__iter__" (not iterable)  [union-attr]
callback.py:371: error: Argument 2 to "EarlyStopException" has incompatible type "Optional[Union[List[Tuple[str, str, float, bool]], List[Tuple[str, str, float, bool, float]]]]"; expected "Union[List[Tuple[str, str, float, bool]], List[Tuple[str, str, float, bool, float]]]"  [arg-type]
```


python-package/lightgbm/callback.py
```diff
@@ -55,7 +59,7 @@ def _format_eval_result(value: _EvalResultTuple, show_stdv: bool) -> str:
         return f"{value[0]}'s {value[1]}: {value[2]:g}"
     elif len(value) == 5:
         if show_stdv:
-            return f"{value[0]}'s {value[1]}: {value[2]:g} + {value[4]:g}"
+            return f"{value[0]}'s {value[1]}: {value[2]:g} + {value[4]:g}"  # type: ignore[misc]
```
jameslamb, line 62: These eval result tuples all have the same first four attributes:

```text
(eval_name, metric_name, eval_result, is_higher_better)
```

In `lgb.cv()`, that `eval_result` is the mean of the metric across all cross-validation folds. In that case, these tuples can optionally contain a fifth value with the standard deviation of that metric across all cross-validation folds.

https://github.com/microsoft/LightGBM/blob/2fe2bf0675ef127a29f9bc49511fc0099cf3e140/python-package/lightgbm/engine.py#L513

`mypy` can't tell automatically that being inside this `if len(value) == 5:` block means that the tuple has 5 elements, and raises the following error on this line:

> callback.py:62: error: Tuple index out of range  [misc]

🤷🏻

python-package/lightgbm/callback.py
```diff
@@ -350,12 +353,17 @@ def __call__(self, env: CallbackEnv) -> None:
...
             score = env.evaluation_result_list[i][2]
-            if self.best_score_list[i] is None or self.cmp_op[i](score, self.best_score[i]):
+            if first_time_updating_best_score_list or self.cmp_op[i](score, self.best_score[i]):
                 self.best_score[i] = score
                 self.best_iter[i] = env.iteration
-                self.best_score_list[i] = env.evaluation_result_list
+                if first_time_updating_best_score_list:
+                    self.best_score_list.append(env.evaluation_result_list)
+                else:
+                    self.best_score_list[i] = env.evaluation_result_list
```
jameslamb, line 366: On `master`, here's how `_EarlyStoppingCallback` currently populates `best_score_list`.

The first time the callback is called, run this `_EarlyStoppingCallback._init()` method.

https://github.com/microsoft/LightGBM/blob/2fe2bf0675ef127a29f9bc49511fc0099cf3e140/python-package/lightgbm/callback.py#L348-L350

In that `_init()` method, ensure that `self.best_score_list` is the same length as `CallbackEnv.evaluation_result_list` by padding it with `None`s.

https://github.com/microsoft/LightGBM/blob/2fe2bf0675ef127a29f9bc49511fc0099cf3e140/python-package/lightgbm/callback.py#L328-L330

And then whenever the callback is called, iterate over `CallbackEnv.evaluation_result_list` and put the corresponding item from it into `self.best_score_list` if:

* it's the very first time the callback was called (`self.best_score_list[i] is None`)
* the new evaluation result is better than the previous one by at least some margin `delta`, which defaults to 0.0 (`self.cmp_op[i](score, self.best_score[i])`

https://github.com/microsoft/LightGBM/blob/2fe2bf0675ef127a29f9bc49511fc0099cf3e140/python-package/lightgbm/callback.py#L355-LL358

So those `None` entries in `self.best_score_list` aren't expected to be there when the first call of `_EarlyStoppingCallback` completes. They're not visible to user code, and nothing else in `lightgbm` relies on them.

HOWEVER... their brief presence in that list tells `mypy` "well sometimes there can be `None` in this list". And honestly, it confused me too.

This PR proposes making `self.best_score_list` easier for both `mypy` and humans to understand by removing that `None`-padding, and instead changing the update logic to *"if it's the first time this list is being updated, build it up by `.append()`-ing every element of `env.evaluation_result_list` to it"*.

jameslamb:
@jmoralez @guolinke @shiyu1994 do you any of you have time to help out on some reviews with this and the other `mypy` PRs I have up this week?

I'm sorry to `@` you... I'm trying to do this in small, easy-to-review pieces but some of the type-hinting things are fairly interrelated so it gets hard when too many unmerged changes stack up.

jameslamb:
Thanks so much for the reviews @guolinke !

guolinke (approved):

PR #5794, opened 2023-03-19
[python-package] fix mypy error in Booster.get_split_value_histogram()

jameslamb:
Contributes to #3867.

```text
basic.py:4417: error: Need type annotation for "values" (hint: "values: List[<type>] = ...")  [var-annotated]
```


python-package/lightgbm/basic.py
```diff
@@ -4414,7 +4414,7 @@ def add(root: Dict[str, Any]) -> None:
         model = self.dump_model()
         feature_names = model.get('feature_names')
         tree_infos = model['tree_info']
-        values = []
+        values: List[float] = []
```
jameslamb, line 4417: This will only contain floats, not the categorical feature thresholds like `"A|B|C"`, because this method doesn't work for categorical features.

https://github.com/microsoft/LightGBM/blob/2fe2bf0675ef127a29f9bc49511fc0099cf3e140/python-package/lightgbm/basic.py#L4407-L4410

guolinke (approved):

Issue #5606, opened 2022-11-25
Add quantized training


shiyu1994:
## Summary

Add quantized training algorithm, which accelerates training on both CPU and GPU. See our latest NeurIPS 2022 paper (https://neurips.cc/virtual/2022/poster/53370).

jameslamb:
Interesting! Is there a link to the poster materials that doesn't require creating an account?

<img width="1424" alt="image" src="https://user-images.githubusercontent.com/7608904/204065122-3cb7faba-a042-441d-bf77-818e8e116b1d.png">

shiyu1994:
Maybe this link can be accessed without registration to the conference.
https://nips.cc/Conferences/2022/Schedule?showEvent=53370

jameslamb:
Yep that worked, thank you! Very exciting!

replacementAI:
Is there an update regarding the above?

AllenSun1024:
> Is there an update regarding the above?

It seems like that it's not updated.

jameslamb:
Given that quantized training is now supported for CPU and CUDA, I think this issue can be closed.

I'm closing this. @guolinke @shiyu1994 if I'm wrong about that please do re-open it and add a comment explaining what else still needs to be done.

Thanks for the quick translation from your research work into working, released code in this project that its users can get with just a `pip install`. Pretty exciting!!!

jameslamb:
Ah sorry, did not realize #5933 was not merged yet. I'll re-open this.

shiyu1994:
@jameslamb No worries. I will continue my work in #5933.

jameslamb:
I got too excited about closing old issues, sorry 😆

PR #5800, opened 2023-03-23
Add quantized training (CPU part)

shiyu1994:
This is to add the quantized training for CPU, which speed up the training. (#5606)


shiyu1994:
It is weird that the `test_monotone_penalty_max` fails in this commit [6a76fde], which doesn't seem to have any changes that will cause the failure. This should call for further attention to check if the test case fails randomly in some cases. Maybe we should simply make this test case more deterministic by setting the number of threads to be 1 or enable the deterministic flag.

(https://github.com/microsoft/LightGBM/pull/5800/commits/6a76fded10072b0ffcf43d2b0433c8f34a5e85e9)
https://dev.azure.com/lightgbm-ci/lightgbm-ci/_build/results?buildId=14615&view=logs&j=c28dceab-947a-5848-c21f-eef3695e5f11&t=fa158246-17e2-53d4-5936-86070edbaacf

shiyu1994:
> @guolinke This is ready for review.

Sorry I think one more test case for distributed quantized training should be added in `test_dask.py`. Please pending the review until that is ready.

include/LightGBM/bin.h
```diff
@@ -56,6 +59,28 @@ inline static void HistogramSumReducer(const char* src, char* dst, int type_size
   }
 }
 
+inline static void Int32HistogramSumReducer(const char* src, char* dst, int type_size, comm_size_t len) {
+  const int64_t* src_ptr = reinterpret_cast<const int64_t*>(src);
+  int64_t* dst_ptr = reinterpret_cast<int64_t*>(dst);
```
guolinke, line 64: is it okay to use int64_t pointer?

shiyu1994: I think the alignments of these buffer vectors is 32 bytes now. So, it should be safe to do this.
https://github.com/microsoft/LightGBM/blob/788e1aa6ae08a14b6bf6f2799b5c4d61426b2079/src/treelearner/parallel_tree_learner.h#L89-L91

shiyu1994:
@guolinke

guolinke (approved):

PR #5807, opened 2023-03-25
WIP: [ci] fix LaTeX tools installation in R jobs (fixes #5802)

jameslamb:

PR #5532, opened 2022-10-10
[python-package] prefix cast_numpy_array_to_dtype with _

Madnex:
Contributes to https://github.com/microsoft/LightGBM/issues/5313

PR #5808, opened 2023-03-26
[python-package] add scikit-learn-style API for early stopping

### LINKED ISSUE #3313: [Feature Request] Auto early stopping in Sklearn API
Is it possible to perform early stopping using cross-validation or automatically sampling data from the provided train set without explicitly specifying an eval set?

ClaudioSalvatoreArcidiacono:
Fixes [#3313](https://github.com/microsoft/LightGBM/issues/3313)

Implements Scikit-learn like interface for early stopping.


ClaudioSalvatoreArcidiacono:
> @ClaudioSalvatoreArcidiacono please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement
> ### Contribution License Agreement
> This Contribution License Agreement (**“Agreement”**) is agreed to by the party signing below (**“You”**), and conveys certain license rights to Microsoft Corporation and its affiliates (“Microsoft”) for Your contributions to Microsoft open source projects. This Agreement is effective as of the latest signature date below.
> 
> 1. **Definitions**.
>    **“Code”** means the computer software code, whether in human-readable or machine-executable form,
>    that is delivered by You to Microsoft under this Agreement.
>    **“Project”** means any of the projects owned or managed by Microsoft and offered under a license
>    approved by the Open Source Initiative ([www.opensource.org](http://www.opensource.org)).
>    **“Submit”** is the act of uploading, submitting, transmitting, or distributing code or other content to any
>    Project, including but not limited to communication on electronic mailing lists, source code control
>    systems, and issue tracking systems that are managed by, or on behalf of, the Project for the purpose of
>    discussing and improving that Project, but excluding communication that is conspicuously marked or
>    otherwise designated in writing by You as “Not a Submission.”
>    **“Submission”** means the Code and any other copyrightable material Submitted by You, including any
>    associated comments and documentation.
> 2. **Your Submission**. You must agree to the terms of this Agreement before making a Submission to any
>    Project. This Agreement covers any and all Submissions that You, now or in the future (except as
>    described in Section 4 below), Submit to any Project.
> 3. **Originality of Work**. You represent that each of Your Submissions is entirely Your original work.
>    Should You wish to Submit materials that are not Your original work, You may Submit them separately
>    to the Project if You (a) retain all copyright and license information that was in the materials as You
>    received them, (b) in the description accompanying Your Submission, include the phrase “Submission
>    containing materials of a third party:” followed by the names of the third party and any licenses or other
>    restrictions of which You are aware, and (c) follow any other instructions in the Project’s written
>    guidelines concerning Submissions.
> 4. **Your Employer**. References to “employer” in this Agreement include Your employer or anyone else
>    for whom You are acting in making Your Submission, e.g. as a contractor, vendor, or agent. If Your
>    Submission is made in the course of Your work for an employer or Your employer has intellectual
>    property rights in Your Submission by contract or applicable law, You must secure permission from Your
>    employer to make the Submission before signing this Agreement. In that case, the term “You” in this
>    Agreement will refer to You and the employer collectively. If You change employers in the future and
>    desire to Submit additional Submissions for the new employer, then You agree to sign a new Agreement
>    and secure permission from the new employer before Submitting those Submissions.
> 5. **Licenses**.
> 
> * **Copyright License**. You grant Microsoft, and those who receive the Submission directly or
>   indirectly from Microsoft, a perpetual, worldwide, non-exclusive, royalty-free, irrevocable license in the
>   Submission to reproduce, prepare derivative works of, publicly display, publicly perform, and distribute
>   the Submission and such derivative works, and to sublicense any or all of the foregoing rights to third
>   parties.
> * **Patent License**. You grant Microsoft, and those who receive the Submission directly or
>   indirectly from Microsoft, a perpetual, worldwide, non-exclusive, royalty-free, irrevocable license under
>   Your patent claims that are necessarily infringed by the Submission or the combination of the
>   Submission with the Project to which it was Submitted to make, have made, use, offer to sell, sell and
>   import or otherwise dispose of the Submission alone or with the Project.
> * **Other Rights Reserved**. Each party reserves all rights not expressly granted in this Agreement.
>   No additional licenses or rights whatsoever (including, without limitation, any implied licenses) are
>   granted by implication, exhaustion, estoppel or otherwise.
> 
> 6. **Representations and Warranties**. You represent that You are legally entitled to grant the above
>    licenses. You represent that each of Your Submissions is entirely Your original work (except as You may
>    have disclosed under Section 3). You represent that You have secured permission from Your employer to
>    make the Submission in cases where Your Submission is made in the course of Your work for Your
>    employer or Your employer has intellectual property rights in Your Submission by contract or applicable
>    law. If You are signing this Agreement on behalf of Your employer, You represent and warrant that You
>    have the necessary authority to bind the listed employer to the obligations contained in this Agreement.
>    You are not expected to provide support for Your Submission, unless You choose to do so. UNLESS
>    REQUIRED BY APPLICABLE LAW OR AGREED TO IN WRITING, AND EXCEPT FOR THE WARRANTIES
>    EXPRESSLY STATED IN SECTIONS 3, 4, AND 6, THE SUBMISSION PROVIDED UNDER THIS AGREEMENT IS
>    PROVIDED WITHOUT WARRANTY OF ANY KIND, INCLUDING, BUT NOT LIMITED TO, ANY WARRANTY OF
>    NONINFRINGEMENT, MERCHANTABILITY, OR FITNESS FOR A PARTICULAR PURPOSE.
> 7. **Notice to Microsoft**. You agree to notify Microsoft in writing of any facts or circumstances of which
>    You later become aware that would make Your representations in this Agreement inaccurate in any
>    respect.
> 8. **Information about Submissions**. You agree that contributions to Projects and information about
>    contributions may be maintained indefinitely and disclosed publicly, including Your name and other
>    information that You submit with Your Submission.
> 9. **Governing Law/Jurisdiction**. This Agreement is governed by the laws of the State of Washington, and
>    the parties consent to exclusive jurisdiction and venue in the federal courts sitting in King County,
>    Washington, unless no federal subject matter jurisdiction exists, in which case the parties consent to
>    exclusive jurisdiction and venue in the Superior Court of King County, Washington. The parties waive all
>    defenses of lack of personal jurisdiction and forum non-conveniens.
> 10. **Entire Agreement/Assignment**. This Agreement is the entire agreement between the parties, and
>     supersedes any and all prior agreements, understandings or communications, written or oral, between
>     the parties relating to the subject matter hereof. This Agreement may be assigned by Microsoft.

@microsoft-github-policy-service agree

jameslamb (changes requested):
Thanks for taking the time to contribute to LightGBM! I haven't reviewed this yet, but just leaving a blocking review to make it clear to other maintainers that I'd like the opportunity to review before this is merged. The discussion in #3313 talked about waiting for scikit-learn's interface for `HistGradientBoostingClassifier` to stabilize and mature and then adapting to the choices they made... I'd like the opportunity to look at that interface thoroughly prior to reviewing this.

jameslamb:
Until we have a chance to review, you can improve the chances of this being merged by addressing any CI failures you see. You can safely ignore failing `R-package` jobs, there are some known issues with those (fixed in #5807 ).

ClaudioSalvatoreArcidiacono:
Hey @jameslamb thanks for picking this up. I have started to take a look at the ci failures, I think I can solve most of them easily.

There is one check for which I need some input.

I see that one of the tests checks that the init parameters for the sklearn API and the Dask API have the same arguments (see [this one](https://github.com/microsoft/LightGBM/blob/master/tests/python_package_test/test_dask.py#L1682)). Early stopping is not available yet for the Dask API,  so I do not see how can we easily iron that out. Shall I add some more exceptions to that specific test?

jameslamb:
>  Shall I add some more exceptions to that specific test?

@ClaudioSalvatoreArcidiacono no, please do not do that. This test you've linked to is there exactly to catch such deviations.

If you absolutely need to add arguments to the constructors of the scikit-learn estimators in this project, add the same arguments in the same order to the Dask estimators, with default values of `None` or something, and raise `NotImplementedError` in the Dask interface when any non-`None` values are passed to those arguments.

For what it's worth, I am not totally convinced yet that we should take on the exact same interface as `scikit-learn`, (especially the large added complexity of this new `validation_set_split_strategy` argument).  #3313 was primarily about whether or not to enable early stopping by default... not explicitly about changing the signature of the `lightgbm.sklearn` estimators to match `HistGradientBoostingClassifier`.

If you are committed to getting the CI passing here I'm willing to consider it, but just want to set the right expectation that I expect you to also explain specifically the benefit of adding all this new complexity.

jameslamb:
By the way, it looks like you are not signing your commits with an email address tied to your GitHub account.

Screen Shot 2023-04-18 at 8 52 12 PM

See https://github.com/microsoft/LightGBM/pull/5532#issuecomment-1274032708 and the comments linked from it for an explanation of what I mean by that and an explanation of how to fix it.

jameslamb:
It has been about 6 weeks since I last provided a review on this PR, and there has not been any activity on it since then.

I'm closing this, assuming it's been abandoned. To make it clear to others onterested in this feature that they shouldn't be waiting on this PR.

@ClaudioSalvatoreArcidiacono if you have time in the future to work with maintainers here, we'd welcome future contributions.

ClaudioSalvatoreArcidiacono:
Hey @jameslamb, I did not have much time to take a look at this lately, I should be more available now and in the coming weeks.

If you also have some time to help me reviewing it I can pick this PR up again.

Regarding your previous comments, thanks for the heads up on signing commits, I will sign the next commits as you mentioned.

About the complexity of the proposed implementation, I am definitely open for feedback from the maintainers and I am willing to change the proposed implementation if necessary.

In the proposed implementation I tried to stick to what is written in the [FAQ](https://lightgbm.readthedocs.io/en/latest/FAQ.html#why-is-early-stopping-not-enabled-by-default-in-lightgbm):
> The appropriate splitting strategy depends on the task and domain of the data, information that a modeler has but which LightGBM as a general-purpose tool does not.

So, in the proposed implementation I tried to find a common ground between convenience of activating early stopping using only init params and customisability of the splitting strategy.

ClaudioSalvatoreArcidiacono:
Hey @jameslamb, no problem. Thanks for being transparent on your availability on this PR and for your feedback. I followed it and I made the PR easier to review now.

I have removed the functionality to use a custom splitter and I made the changes much smaller.

> * what does "Scikit-learn like interface" mean, precisely?
>   
>   * does it mean you've implemented exactly the same interface as `HistGradientBoostingClassifier` and `HistGradientBoostingRegressor` ? If so can you please link to code and docs showing that?

I have now implemented the same interface as `HistGradientBoostingClassifier` and `HistGradientBoostingRegressor`.

>   * and more broadly, why does this need to change ANYTHING about the public interface of `lightgbm`? And couldn't the existing mechanisms used inside `lightgbm.cv()` be used instead of adding all this new code for splitting? e.g. https://github.com/microsoft/LightGBM/blob/9f78cceee4911dd56f4635dfd36d4482363db5aa/python-package/lightgbm/engine.py#L562-L564

In this implementation I tried to reuse the splitting function used in `lightgbm.cv()`. Thanks for the tip.

> * what has to happen to make these changes consistent with how `lightgbm` currently works? For example...
>   
>   * what happens when `early_stopping_rounds` is passed to the estimator constructor via `**kwargs` and `n_iter_no_change` is set to a non-default value in the constructor... which value wins?
Good observation, I think it is indeed needed to rename the arguments so that they will be more consistent with LightGBM naming conventions.
>   * What happens if `early_stopping=True` is passed but `valid_sets` are also passed to `.fit()`? Does that disable the automatic splitting and just use the provided validation sets?
Correct.

python-package/lightgbm/sklearn.py
```diff
@@ -412,6 +412,9 @@ def __init__(
         random_state: Optional[Union[int, np.random.RandomState]] = None,
         n_jobs: Optional[int] = None,
         importance_type: str = 'split',
+        use_early_stopping: bool = False,
+        validation_fraction: Optional[float] = 0.1,
+        early_stopping_round: int = 10,
```
jameslamb, line 417: It shouldn't be necessary to add `n_iter_no_change` or `early_stopping_round` to the public interface of these classes.

All keyword arguments passed through `**kwargs` are appended to the `params` stored on the object and passed through to LightGBM's C++ code.

https://github.com/microsoft/LightGBM/blob/7a801f7b7b1ed8faf4cd1dbc09086c269fc51e1a/python-package/lightgbm/sklearn.py#L569

https://github.com/microsoft/LightGBM/blob/7a801f7b7b1ed8faf4cd1dbc09086c269fc51e1a/python-package/lightgbm/sklearn.py#L617-L621

`n_iter_no_change` is supported as a parameter alias for `early_stopping_rounds` in LightGBM.

https://github.com/microsoft/LightGBM/blob/7a801f7b7b1ed8faf4cd1dbc09086c269fc51e1a/src/io/config_auto.cpp#L83

And all aliases will be resolved in `lightgbm.train()`, which is called by the scikit-learn estimators' `fit()` methods.

https://github.com/microsoft/LightGBM/blob/7a801f7b7b1ed8faf4cd1dbc09086c269fc51e1a/python-package/lightgbm/sklearn.py#L842

https://github.com/microsoft/LightGBM/blob/7a801f7b7b1ed8faf4cd1dbc09086c269fc51e1a/python-package/lightgbm/engine.py#L175-L179

Can you please explain why you added this? If you just didn't realize LightGBM had this mechanism, please remove this keyword argument from all these classes (but keep passing it explicitly in the unit tests, if it's the preferred argument for `scikit-learn`).

python-package/lightgbm/sklearn.py
```diff
@@ -412,6 +412,9 @@ def __init__(
         random_state: Optional[Union[int, np.random.RandomState]] = None,
         n_jobs: Optional[int] = None,
         importance_type: str = 'split',
+        use_early_stopping: bool = False,
```
jameslamb, line 415: The Dask and `scikit-learn` interfaces should be identical.

Since `HistGradientBoosterClassifier` calls this `early_stopping` ([docs link](https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.HistGradientBoostingClassifier.html)), please also call it `early_stopping` here and in the Dask interface.

tests/python_package_test/test_dask.py
```diff
@@ -1659,6 +1659,19 @@ def test_dask_classes_and_sklearn_equivalents_have_identical_constructors_except
...
+@pytest.mark.parametrize(
+    "dask_class",
+    [
+        lgb.DaskLGBMClassifier,
+        lgb.DaskLGBMRegressor,
+        lgb.DaskLGBMRanker
+    ]
+)
+def test_dask_early_stopping_raises_not_implemented_error(dask_class):
+    with pytest.raises(NotImplementedError):
```
jameslamb, line 1671: Wherever you use `pytest.raises()` in this project, please use `match` to ensure the code is raising **exactly** the error message you're expecting. That improves the chances that other unrelated errors could be caught by this test.

In case you're unfamiliar with `pytest.raises()`, here are the docs:  https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-raises

And here's an example of the pattern I'm requesting:

https://github.com/microsoft/LightGBM/blob/7a801f7b7b1ed8faf4cd1dbc09086c269fc51e1a/tests/python_package_test/test_basic.py#L781-L788

python-package/lightgbm/sklearn.py
```diff
@@ -785,7 +808,24 @@ def fit(
 
         train_set = Dataset(data=_X, label=_y, weight=sample_weight, group=group,
                             init_score=init_score, categorical_feature=categorical_feature,
-                            params=params)
+                            params=params, free_raw_data=False)
```
jameslamb, line 811: Why is it necessary to pass `free_raw_data=False` here? That would be a significant breaking change, as it means that these estimators would default to now holding a copy of the raw training data in memory. That increases the likelihood of out-of-memory errors.

Learning that doing that was required to implement automatic early stopping would make me less confident that this feature should be enabled in LightGBM.

python-package/lightgbm/sklearn.py
jameslamb: Also why is this defaulting to `False`? This PR's title says "enable early stopping automatically", but if it was merged as-is early stopping would be OFF by default.

Since the goal of this PR is to make LightGBM's scikit-learn interface consistent with scikit-learn itself, please follow what `HistGradientBoostingClassifier` does. As seen in their docs: https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.HistGradientBoostingClassifier.html

> **early_stopping**: `‘auto’` or `bool`, `default=’auto’`
.
> If ‘auto’, early stopping is enabled if the sample size is larger than 10000. If True, early stopping is enabled, otherwise early stopping is disabled.

Please match that behavior.

jameslamb (changes requested):
@ClaudioSalvatoreArcidiacono now that LightGBM v4.0 is out, I'd be happy to return to reviewing this.

Can you please update it to the latest `master`? Please also see my next round of comments.

jameslamb:
No problem, thanks for letting us know! I'll be happy to continue working on this whenever you have time.

ClaudioSalvatoreArcidiacono:
Hey @jameslamb, Thanks again for your review comments! I was aware of the aliases mechanism but I did not fully understand how it worked. Your comment really helped me in understanding them.

I have worked on your feedback and I think now the PR is in good shape.

In this implementation I tried to stick to the scikit-learn interface of [HistGradientBoostingClassifier](https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.HistGradientBoostingClassifier.html), so the parameter `early_stopping` is `'auto'` by default.

Since we are not adding `early_stopping_round` to the public interface, the default value  for `early_stopping_round` is somewhere else in the code, I think it would be better to create a constant in the `sklearn.py` file where we set the default value, happy to hear alternative solutions from you.

Lastly, we should still mention somewhere in the documentation that early stopping is enabled by default in the scikit-learn interface of LightGBM and also we should mention how to forcefully disable it. Where would you suggest to add it?

python-package/lightgbm/sklearn.py
ClaudioSalvatoreArcidiacono: I have found a workaround for that, `free_raw_data` is not set to `False` anymore.

tests/python_package_test/test_dask.py
ClaudioSalvatoreArcidiacono: This test has been removed. In future usages of `pytest.raises` I will also set `match`.

python-package/lightgbm/sklearn.py
ClaudioSalvatoreArcidiacono: This parameter has been removed from the public interface.

python-package/lightgbm/sklearn.py
ClaudioSalvatoreArcidiacono: I did not fully understand how aliases worked in LightGBM, this comment was super helpful in unblocking me, thanks a lot for it.

I have removed `n_iter_no_change` from the public interface.

python-package/lightgbm/basic.py
```diff
@@ -2538,14 +2538,12 @@ def set_categorical_feature(
         self : Dataset
             Dataset with set categorical features.
         """
-        if self.categorical_feature == categorical_feature:
+        if self.categorical_feature == categorical_feature or categorical_feature == 'auto':
```
jameslamb, line 2541: I support moving this up here to reduce the complexity of the code but how is this related to this pull request? If it's not, can you please put up a separate PR with this change?

That would also have the added benefit of removing your status as a first-time contributor to LightGBM... so you'd no longer have to wait for a maintainer to click a button to run the CI jobs when you push a commit 😊

python-package/lightgbm/sklearn.py
```diff
@@ -566,6 +571,7 @@ def __init__(
         self._n_features_in: int = -1
         self._classes: Optional[np.ndarray] = None
         self._n_classes: int = -1
+        self.validation_fraction = validation_fraction
```
jameslamb, line 574: Please move this up to right after `self.importance_type = importance_type`.

If you look carefully, you'll see that all such assignments are ordered by the order of the keyword arguments in the construct... let's please preserve that ordering, to avoid mistakes and make the code easier to read.

python-package/lightgbm/sklearn.py
```diff
@@ -668,9 +674,24 @@ def _process_params(self, stage: str) -> Dict[str, Any]:
...
         if isinstance(params['random_state'], np.random.RandomState):
             params['random_state'] = params['random_state'].randint(np.iinfo(np.int32).max)
+
+        params = _choose_param_value(
+            main_param_name="early_stopping_round",
+            params=params,
+            default_value="auto",
+        )
+        if params["early_stopping_round"] == "auto":
+            params["early_stopping_round"] = 10 if hasattr(self, "n_rows_train") and self.n_rows_train > 10000 else None
```
jameslamb, line 688: Please break this into an `if-else` block instead of using such a long 1-line statement. This is very difficult to read (at least for me).

python-package/lightgbm/sklearn.py
```diff
@@ -789,44 +809,61 @@ def fit(
         train_set = Dataset(data=_X, label=_y, weight=sample_weight, group=group,
                             init_score=init_score, categorical_feature=categorical_feature,
                             params=params)
+        self._n_rows_train = _X.shape[0]
+        if params["early_stopping_round"] == "auto":
+            params["early_stopping_round"] = 10 if self.n_rows_train > 10000 else None
```
jameslamb, line 814: please make this an `if-else` block instead of a one-liner

tests/python_package_test/test_sklearn.py
```diff
@@ -257,6 +257,51 @@ def test_binary_classification_with_custom_objective():
...
+def test_binary_classification_with_auto_early_stopping(use_weight):
+
+    X, y = load_breast_cancer(return_X_y=True)
+    n_estimators = 1000
+    gbm = lgb.LGBMClassifier(
+        n_estimators=n_estimators, random_state=42, verbose=-1, early_stopping=True
+    )
+    weight = np.full_like(y, 2) if use_weight else None
+    gbm.fit(X, y, sample_weight=weight)
+    assert bool(gbm.best_iteration_)
```
jameslamb, line 270: Please make these tests stricter. Instead of just "`best_iteration_` is set", please add assertions checking that:

* `best_iteration_` is set *to the expected value* (e.g. that early stopping actually happened)
* the number of trees (`gbm.booster_.num_trees()`) is set to the expected value

For example, you're not testing here that `early_stopping_rounds` is set to `10` automatically if it's not provided.

If you need help with testing that early stopping actually happened, look around the other tests in the project or the one I've added in #6095.

tests/python_package_test/test_sklearn.py
```diff
@@ -257,6 +257,51 @@ def test_binary_classification_with_custom_objective():
     assert ret < 0.05
 
 
+@pytest.mark.parametrize('use_weight', [True, False])
+def test_binary_classification_with_auto_early_stopping(use_weight):
+
+    X, y = load_breast_cancer(return_X_y=True)
+    n_estimators = 1000
```
jameslamb, line 264:
```suggestion
    n_estimators = 15
```

Please try to keep these tests as small and fast as is necessary to test the correctness of the code. We run 500+ unit tests on every build of the the Python package... the smaller they are, the more testing cycles we can get through and the faster development on the project moves.

It would also be much appreciated if you'd set `num_leaves=5` or some similar smaller-than-default value.

tests/python_package_test/test_sklearn.py
jameslamb: (this applies to all 3 tests you've added)

tests/python_package_test/test_sklearn.py
```diff
@@ -609,21 +654,21 @@ def test_pandas_categorical():
     X[cat_cols_actual] = X[cat_cols_actual].astype('category')
     X_test[cat_cols_actual] = X_test[cat_cols_actual].astype('category')
     cat_values = [X[col].cat.categories.tolist() for col in cat_cols_to_store]
-    gbm0 = lgb.sklearn.LGBMClassifier(n_estimators=10).fit(X, y)
+    gbm0 = lgb.sklearn.LGBMClassifier(n_estimators=10, random_state=42).fit(X, y)
```
jameslamb, line 657: Why was this change necessary?

jameslamb (changes requested):
Thanks for your continued effort on this. I did a quick review this morning and left comments on a few things, but that set of suggestions isn't comprehensive. It will take some time to provide a more thorough review... I'll do that when I can.

In addition to the comments I left inline, please add tests covering each of the new `if-else` code branches you've introduced:

* early stopping is only enabled automatically if the training data has `> 10_000` rows
* passing `early_stopping_round=True` / `early_stopping_round=False` does the expected thing

python-package/lightgbm/sklearn.py
jameslamb: great, thank you

tests/python_package_test/test_dask.py
jameslamb: thank you

jameslamb:
> Lastly, we should still mention somewhere in the documentation that early stopping is enabled by default in the scikit-learn interface of LightGBM and also we should mention how to forcefully disable it. Where would you suggest to add it?

I think here would be appropriate:

https://github.com/microsoft/LightGBM/blob/921479b99fb5b691801e0e794f2196a94ea17d79/docs/Python-Intro.rst#L223

python-package/lightgbm/basic.py
ClaudioSalvatoreArcidiacono: This change is necessary in order to make early stopping work with setting categorical features during fit. Without this change the following error is raised

```
lightgbm.basic.LightGBMError: Cannot set categorical feature after freed raw data, set free_raw_data=False when construct Dataset to avoid this.
```

I will add a test to check that early_stopping=True works together with setting categorical features in fit.

tests/python_package_test/test_sklearn.py
ClaudioSalvatoreArcidiacono: Thanks for spotting this, it comes from a previous implementation where early stopping was true by default, now it is 'auto' by default. This test does not use early stopping indeed.

python-package/lightgbm/basic.py
jameslamb: oh! Right right, setting `categorical_feature` after the Dataset's constructed and when you don't have the raw data around any more is problematic.

Now I'm even more confused though... how does adding automatic early stopping interact with anything related to setting categorical features?

You are just enabling *automatically* something that is already possible in `lightgbm`, changes like this shouldn't be necessary.

tests/python_package_test/test_sklearn.py
jameslamb: Ok thank you. Please, as you continue on this, check the diff and be confident that every change you're proposing is necessary and can be justified. The more you can eliminate these "what is this? ... oh nevermind we don't need it any more" rounds of reviews, the faster this will get to a state that we feel confident merging.

python-package/lightgbm/basic.py
ClaudioSalvatoreArcidiacono: The thing is that in order to enable automatic early stopping, we also need to automatically create an evaluation set.

In the current implementation, the evaluation set is created by splitting the `train_set` ([python-package/lightgbm/sklearn.py#L818-831](https://github.com/ClaudioSalvatoreArcidiacono/LightGBM/blob/a5b0c603c34949f9ed7fd927d26089c485f1aaa4/python-package/lightgbm/sklearn.py#L818-831)) instead of splitting the raw data passed to fit (X, y). This has been done to reuse the existing code for creating evaluation sets which relies on the Dataset API.

I am not very fluent with the Dataset API of lightgbm, so if you could recommend better ways for splitting train and validation I would be open for changes.

ClaudioSalvatoreArcidiacono:
> > Lastly, we should still mention somewhere in the documentation that early stopping is enabled by default in the scikit-learn interface of LightGBM and also we should mention how to forcefully disable it. Where would you suggest to add it?
> 
> I think here would be appropriate:
> 
> https://github.com/microsoft/LightGBM/blob/921479b99fb5b691801e0e794f2196a94ea17d79/docs/Python-Intro.rst#L223

Thanks, I have added something there.
Shall we also mention something [here](https://github.com/microsoft/LightGBM/blob/master/docs/FAQ.rst#12-why-is-early-stopping-not-enabled-by-default-in-lightgbm)?

tests/python_package_test/test_sklearn.py
ClaudioSalvatoreArcidiacono: Sure, thanks for the suggestion.

ClaudioSalvatoreArcidiacono:
Hey @jameslamb, I think the PR is ready for a second look, please let me know if there are some Changes you would like me to do :)

jameslamb:
Thanks for returning to this.

@jmoralez could you take the next round of reviews on this?

jmoralez:
The original issue (#3313) requested the possibility of doing early stopping with the scikit-learn API by specifying arguments in the constructor, not automatically performing early stopping. I understand that's what HistGradientBoosting(Classifier|Regressor) do, but I think we could also consider doing what GradientBoosting(Classifier|Regressor) do, where the default is not to do it but having the arguments in the init signature to support it. Otherwise this would be a silently breaking change (I know we would list it as a breaking change but it could confuse people who don't read the release notes).

I would like to have this clear before reviewing, since we would be doing many things behind the user's back (automatically enabling early stopping for >10,000 rows, stratified if classification, setting the number of folds, shuffling). I would be more comfortable if it was an explicit decision (like setting `early_stopping_rounds>0` in the params for example).

ClaudioSalvatoreArcidiacono:
Hey @jmoralez, thanks for your comment. I agree with you, my preference would be for having early stopping off by default and it should be explicitly set by an init parameter. This was also my original implementation, but I have changed it due to [this comment](https://github.com/microsoft/LightGBM/pull/5808#discussion_r1297930419) from @jameslamb.

Could the two of you maybe agree on what would you like to see as the default behaviour :)?

jmoralez:
What are your thoughts on this (enabling early stopping by default) @borchero?

borchero:
> What are your thoughts on this (enabling early stopping by default) @borchero?

Sorry, I only saw this comment now 🫣 I wouldn't enable it by default as (1) it is not integral to use for boosted trees and (2) early stopping is not supported for all boosting strategies (I recently learnt that `dart` does not support it) which might cause confusion.

ClaudioSalvatoreArcidiacono:
Hey @jmoralez, @borchero I have set early stopping to be disabled by default and updated the tests and docs. Could you please review this PR when you have a moment ?

Do you maybe think I should mention the early stopping parameter in the docstring of the lightgbm scikit-learn classes?

python-package/lightgbm/sklearn.py
```diff
@@ -812,11 +818,29 @@ def _process_params(self, stage: str) -> Dict[str, Any]:
...
             params["random_state"] = int(params["random_state"].integers(np.iinfo(np.int32).max))
+
+        params = _choose_param_value(
+            main_param_name="early_stopping_round",
+            params=params,
+            default_value="auto",
+        )
+        if params["early_stopping_round"] == "auto":
+            if hasattr(self, "_n_rows_train") and self._n_rows_train > 10_000:
+                params["early_stopping_round"] = 10
```
jameslamb, line 835: In https://github.com/microsoft/LightGBM/pull/5808#issuecomment-2506680690, you said:

> *I have set early stopping to be disabled by default*

But this block of code directly contradicts that... it enables early stopping by default if there are more than 10K rows in the training data.

From all the prior discussion on this PR, it seems clear to me that @borchero and @jmoralez both requested you **not** do that, and instead limit the scope of this PR to just providing an API for early stopping that matches how `scikit-learn` estimators do it. I agree with them... it shouldn't be enabled by default as part of this PR.

Please, do the following:

* change the title of the PR (which will become a commit message and release notes item) to `[python-package] add scikit-learn-style API for early stopping`
* remove this logic about `early_stopping_round = "auto"` (and the corresponding changes further down where you moved parameter processing to after input data validation, so it could reference `self._n_rows_train`
* match whatever API `HistGradientBoostingClassifier` / `HistGradientBoostingRegressor` currently have for enabling early stopping

jameslamb (changes requested):
Thank you for picking this up again! We're happy to work with you but please see my comments... the code, your most recent comments, and the PR title still contradict each other about whether or not this PR is enabling early stopping automatically.

Confusion over this is one of the things that has delayed this PR so long. It's clear to me from @borchero and @jmoralez 's most recent comments, and I agree with them: **do not enable early stopping by default in this PR**.

I hope that just limiting this to "provide a scikit-learn-style for early stopping" will help reduce it's complexity and get it to a state where we can merge it faster.

python-package/lightgbm/sklearn.py
ClaudioSalvatoreArcidiacono: Hey @jameslamb, thanks a lot for your review, I highly appreciate that.

You are right with your comment, it was a mistake on my side, I have fixed it in my latest commits. Thank you for spotting it.

In the latest version i had:
- Renamed the PR as you requested.
- Removed the logic about early_stopping_round = "auto" and the corresponding changes where I switched parameters processing and input validation.
- I think the APIs of `HistGradientBoostingClassifier` / `HistGradientBoostingRegressor` and the current implementation are matching except that:
  - `HistGradientBoostingClassifier` / `HistGradientBoostingRegressor` support `early_stopping_round = "auto"` as this is their default parameter value, but as we discussed above we do not want this behaviour.
  - I have tried to support the parameters alias logic of LightGBM while working around the fact that `early_stopping` is an alias for `early_stopping_round(s)` and `n_iter_no_change`. In the current implementation setting `early_stopping_round(s)` to an integer will override the default value for `n_iter_no_change`, but setting `early_stopping` to an integer will not enable early stopping and will not override the default value for  `n_iter_no_change`. I have opted for this implementation to preserve backward compatibility as much as possible. Please let me know if you also agree with it or you have a different opinion, happy to change it.

ClaudioSalvatoreArcidiacono:
hey @jameslamb, any chance you could take a look at the latest changes ?

tests/python_package_test/test_sklearn.py
```diff
@@ -278,6 +278,112 @@ def test_binary_classification_with_custom_objective():
...
+    X_train, y_train = load_svmlight_file(str(rank_example_dir / "rank.train"))
+    q_train = np.loadtxt(str(rank_example_dir / "rank.train.query"))
+    n_estimators = 5
+    gbm = lgb.LGBMRanker(n_estimators=n_estimators, random_state=42, early_stopping=True, num_leaves=5)
+    gbm.fit(
+        X_train,
+        y_train,
+        group=q_train,
+        eval_at=[1, 3],
+        callbacks=[lgb.reset_parameter(learning_rate=lambda x: max(0.01, 0.1 - 0.01 * x))],
```
jameslamb, line 380:
```suggestion
```

Why is modifying the learning rate necessary for this test? If this is unnecessary and just copied from somwhere else, please remove it.

python-package/lightgbm/sklearn.py
```diff
@@ -509,7 +509,10 @@ def __init__(
         random_state: Optional[Union[int, np.random.RandomState, np.random.Generator]] = None,
         n_jobs: Optional[int] = None,
         importance_type: str = "split",
-        **kwargs: Any,
+        early_stopping: bool = False,
+        n_iter_no_change: int = 10,
+        validation_fraction: Optional[float] = 0.1,
+        **kwargs,
```
jameslamb, line 515:
```suggestion
        **kwargs: Any,
```

Why was this type hint removed? If it was just an accident, please put it back to reduce the size of the diff.

python-package/lightgbm/sklearn.py
jameslamb: > I have removed `n_iter_no_change` from the public interface.

I see `n_iter_no_change` in the signatures for all the estimators still. Please remove it.

This is not the first time in this PR when I've asked for something, you've said a change has been implemented, and I've spent time reviewing only to find that the code doesn't match what you've said. Please, look through the diff at https://github.com/microsoft/LightGBM/pull/5808/files carefully and compare it to the current state of the discussion threads before `@`-ing for another review.

jameslamb: Ah I see now that https://github.com/microsoft/LightGBM/pull/5808#discussion_r1863324478 explains this. Let's continue there and mark this thread resolved.

For the future... you could have saved us a little time by linking these two threads together.

python-package/lightgbm/sklearn.py
jameslamb: Thanks for that explanation.

> *Removed the logic about early_stopping_round = "auto" and the corresponding changes where I switched parameters processing and input validation.*

Agreed, thank you for removing it.

> *`early_stopping` is an alias for `early_stopping_round(s)` and `n_iter_no_change`

Thanks for being careful about backwards compatibility!

It's unfortunate that we have `early_stopping` as an alias for `early_stopping_rounds` and that `scikit-learn`'s estimators now assign a different meaning to that :/

https://github.com/microsoft/LightGBM/blob/1090a93b39e16f49621aa6824cd09d4390c3678a/include/LightGBM/config.h#L394-L397

https://github.com/scikit-learn/scikit-learn/blob/6cccd99aee3483eb0f7562afdd3179ccccab0b1d/sklearn/ensemble/_hist_gradient_boosting/gradient_boosting.py#L1572-L1575

Seems like that has been in `scikit-learn` for a long time (since v0.23):

https://github.com/scikit-learn/scikit-learn/blob/6cccd99aee3483eb0f7562afdd3179ccccab0b1d/sklearn/ensemble/_hist_gradient_boosting/gradient_boosting.py#L1577

I see your implementation of this comment in `LGBMModel._process_params()`:

```python
params = _choose_param_value("early_stopping_round", params, self.n_iter_no_change)
````

I think I see a path to resolving this. It'll be complex, but it would keep LightGBM in compliance with the rules from https://scikit-learn.org/1.5/developers/develop.html#instantiation.

* remove keyword argument `n_iter_no_change` (and property `self.n_iter_no_change`)... but ensure there are unit tests confirming that passing keyword argument `n_iter_no_change` through the scikit-learn estimators correctly sets the number of early stopping rounds
* make keyword argument `early_stopping` have type `Union[bool, int]`
* handle that difference in `LGMBModel._process_params()`, roughly like this:

rewrite this block:

https://github.com/microsoft/LightGBM/blob/1090a93b39e16f49621aa6824cd09d4390c3678a/python-package/lightgbm/sklearn.py#L850-L854

to something like this:

```python
        if stage == "fit":
            # use joblib conventions for negative n_jobs, just like scikit-learn
            # at predict time, this is handled later due to the order of parameter updates
            params = _choose_param_value("num_threads", params, self.n_jobs)
            params["num_threads"] = self._process_n_jobs(params["num_threads"])

            if not isinstance(self.early_stopping, bool) and isinstance(self.early_stopping, int):
                _log_warning(
                    f"Found 'early_stopping={self.early_stopping}' passed through keyword arguments.
                    "Future versions of 'lightgbm' will not allow this, as scikit-learn expects keyword argument "
                    "'early_stopping' to be a boolean indicating whether or not to perform early stopping with "
                    "a randomly-sampled validation set. To set the number of early stopping rounds, and suppress "
                    "this warning, pass early_stopping_rounds={self.early_stopping} instead."
                 )
                 params["early_stopping_round"] = _choose_param_value(
                     main_param_name="early_stopping_round",
                     params=params,
                     default_value=self.early_stopping
                 )
```

And then any other places in the code where you have `self.early_stopping is True` should use `isinstance(self.early_stopping, bool) and self.early_stopping is True`, to avoid this SyntaxWarning:

```python
1 is True
# <stdin>:1: SyntaxWarning: "is" with 'int' literal. Did you mean "=="?
```

I *think* that can work and provide backwards compatibility while allowing us to match scikit-learn's API. And then eventually (after several `lightgbm` releases), that warning could be converted to an error and the type of `early_stopping` could be narrowed to just `bool`.

What do you think?

python-package/lightgbm/basic.py
jameslamb: Looks like this is no longer in the diff, so I think you found another way to do this. Marking this thread resolved.

jameslamb (changes requested):
> hey @jameslamb, any chance you could take a look at the latest changes ?

I reviewed this again tonight. Please see my suggestions. We're getting closer, but there are still some significant unresolved issues. As you respond and change the code, please make every effort possible to keep your changes minimal and only focused on the narrow goal of this PR.

I also marked "resolved" any threads where it seemed that we'd reached a resolution that was reflected in the code. Let's please keep using threads on the "Files changed" tab in this way and resolving them as we move forward.

Thank you for continuing to work on this. I know it can be difficult to get so much critical feedback. That's partially a function of what you've taken on here... you've chosen a very complex problem on a highly-important piece of the library for your first code contribution to LightGBM. If you're willing to keep working with us, I'm willing to keep reviewing and moving this forward.

tests/python_package_test/test_sklearn.py
```diff
@@ -1168,6 +1274,7 @@ def fit_and_check(eval_set_names, metric_names, assumed_iteration, first_metric_
         "verbose": -1,
         "seed": 123,
         "early_stopping_rounds": 5,
+        "early_stopping": True,
```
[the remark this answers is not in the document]
ClaudioSalvatoreArcidiacono: Yes, this was a necessary change in order to make the test pass.

With the code in the current state, passing `early_stopping_rounds` while leaving `early_stopping` to its default value (`False`) does not enable early stopping. Before it did.

If we want to keep the previous behaviour and we want to enable auto early stopping when `early_stopping_rounds` is set to a value and `early_stopping` is left to its default value we need to set the default value of `early_stopping_rounds` to `None` and to change it in `process_parameters` to `10` when `early_stopping` is set to `True`, which will make the actual default value of `early_stopping_rounds` a bit hidden in the code.  Let me know what you prefer, both options have their own drawbacks.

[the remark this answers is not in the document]
jameslamb: Ok, that's a problem. And exactly why I have been so strict in asking that we only see **new** tests added in this PR.

I think that also answers my questions from https://github.com/microsoft/LightGBM/pull/5808#discussion_r1885538801

If the current state of this branch were merged, with this mix of behaviors:

* `early_stopping` keyword argument defaulting to `False`
* `early_stopping=False` meaning "do not perform **any** early stopping"

it would **turn off early stopping for all existing code using early stopping with `lightgbm.sklearn` estimators**. 

That's a very big user-facing breaking change, and one I do not support. Please, let's keep the scope of this PR limited to what I've written in the title... "add scikit-learn-style API for early stopping".

Specifically, this is the behavior I'm expecting:

* `early_stopping=True`:
  - *adds a new validation set to `eval_set` which is a random sample of `{validation_fraction} * num_rows` rows from the training data*
  - *enables early stopping, with `early_stopping_rounds` defaulting to 10 if not provided by any other mechanism*
  - *does not affect any other relevant early stopping behavior... e.g. if you also passed a dataset to `eval_set` in `fit(...)`, then that dataset is ALSO used for early stopping*
* `early_stopping=False`
  - *everything `lightgbm` does is identical to what it does today*

The docstring for the `early_stopping` keyword argument should also be updated to clearly explain this. For example:

```text
        early_stopping : bool, optional (default=False)
            Whether to enable scikit-learn-style early stopping. If set to ``True`, 
            a new validation set will be created by randomly sampling ``validation_fraction`` rows
            from the training data ``X`` passed to ``fit()``. Training will stop if the validation score
            does not improve for a specific number of rounds (controlled by ``n_iter_no_change``).
            This parameter is here for compatibility with ``scikit-learn``'s ``HistGradientBoosting``
            estimators... it does not affect other ``lightgbm``-specific early stopping mechanisms,
            like passing the ``lgb.early_stopping`` callback and validation sets to the ``eval_set``
            argument of `fit()`.
```

> *If we want to keep the previous behaviour and we want to enable auto early stopping when `early_stopping_rounds` is set to a value and `early_stopping` is left to its default value we need to set the default value of `early_stopping_rounds` to `None`*

I think this statement is also coming from you trying to tie this new `.early_stopping` property of the `scikit-learn` estimators to "whether or not *any* early stopping happens", and I guess you are referring to this:

https://github.com/microsoft/LightGBM/blob/4feee28a53b7b04e66ee09103d2aac260404b9cf/python-package/lightgbm/engine.py#L275-L288

I don't think we should do that. Let's limit the `early_stopping` keyword argument to the narrow purpose of "whether or not to add 1 randomly-sampled validation set and perform early stopping on it".

Then, the only thing that has to be figured out is how many rounds to to train without improvement before stopping. I gave a proposal for doing that here: https://github.com/microsoft/LightGBM/pull/5808#discussion_r1885546892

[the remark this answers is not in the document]
jameslamb: @ClaudioSalvatoreArcidiacono, I'd like to propose that we take a different approach to getting this PR done.

Could I take over this PR and push to it, to get it into a state that I'd feel comfortable merging?

Then:

1. *you* review the state and tell me if anything that's left concerns you
2. after you and I agree on the state, we ask other maintainers to review

This has been taking a really large amount of both your and my time and energy. I think this approach would get us to a resolution faster.

I would only push new commits and merge commits (no force-pushing), so you could see everything I changed and revert any of it.

Can we do that?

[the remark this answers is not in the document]
ClaudioSalvatoreArcidiacono: Hey @jameslamb, the statement

> If the current state of this branch were merged, with this mix of behaviours:
>
> - early_stopping keyword argument defaulting to False
> - early_stopping=False meaning "do not perform any early stopping"
>
> it would turn off early stopping for all existing code using early stopping with lightgbm.sklearn estimators.

Is not correct. If the current state of this branch were merged it would not enable early stopping **automatically**. Meaning that passing an `early_stopping` callback while leaving early_stopping keyword argument defaulting to False will still enable early stopping as it has always had. See for example [this other test](https://github.com/ClaudioSalvatoreArcidiacono/LightGBM/blob/3313-enable-auto-early-stopping/tests/python_package_test/test_sklearn.py#L172) which is left unchanged and it passes.

The only non backward compatible change introduced at this stage is passing early_stopping_rounds is not sufficient anymore to enable early stopping (if the early_stopping parameter is left to False). I am definetly open to change that.

Regarding your last suggestion, I am extremely thankful for the effort you have invested in this PR and I am sorry to hear that it is taking more effort than you have anticipated. I am putting my best intentions into this PR and I definetly do not want cause any harm to the library or waste any of your precious time unnecessarily.

I would like to give this PR one last attempt, addressing all of your comments and making sure everything is spot on. I will tag you once I think the PR is ready for a last review. If you think that after this last attempt the gap to be filled is still to big, I will step aside and I will happily let you take over and review your code. 

Are you fine with it :)?

python-package/lightgbm/sklearn.py
```diff
@@ -1006,6 +1030,27 @@ def fit(
 
                 valid_sets.append(valid_set)
 
+        elif self.early_stopping is True:
+            if self.validation_fraction is not None:
+                n_splits = max(int(np.ceil(1 / self.validation_fraction)), 2)
+                stratified = isinstance(self, LGBMClassifier)
```
ClaudioSalvatoreArcidiacono, line 1036: I am not a huge fan of how the validation set is created from the train set using the `_make_n_folds` function.

if 1/validation_fraction is not an integer, the result will be that the actual validation set size will not match the validation fraction specified by the user.

For example, if the validation fraction is 0.4 the number of splits calculated here will be 2, which will result in a fraction of 0.5, instead of 0.4.

Using something like [train_test_split](https://scikit-learn.org/1.5/modules/generated/sklearn.model_selection.train_test_split.html) from scikit-learn would solve the issue for the classification and the regression case, but for ranking tasks our best option is GroupShuffleSplit, which will inevitably suffer from the same issue expressed above. The options that I thought to solve this issue are:

1. Leave the code as-is and raise a warning when 1/validation_fraction is not an integer.
2. Use [train_test_split](https://scikit-learn.org/1.5/modules/generated/sklearn.model_selection.train_test_split.html) for creating the validation set in the classification and regression cases; Raise a warning when 1/validation_fraction is not an integer in the ranking case.

I would lean more towards option 2, but this will make the MR bigger.

@jameslamb I would like to hear your opinion on it, do you perhaps already have something else in mind?

python-package/lightgbm/sklearn.py
ClaudioSalvatoreArcidiacono: It's a great suggestion! thanks a lot for it, I have implemented it in the latest version.

tests/python_package_test/test_sklearn.py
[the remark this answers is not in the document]
ClaudioSalvatoreArcidiacono: @jameslamb, In your previous comment you mentioned

> - early_stopping=True:
>   - adds a new validation set to eval_set which is a random sample of {validation_fraction} * num_rows rows from the training data
>   - enables early stopping, with early_stopping_rounds defaulting to 10 if not provided by any other mechanism
>   - does not affect any other relevant early stopping behavior... e.g. if you also passed a dataset to eval_set in fit(...),  then that dataset is ALSO used for early stopping.

Do you think that we should create an extra validation set even if a validation set has been provided in the `fit()`?
I think that if a validation set has been provided in the `fit()` and the parameter `early_stopping=True` then we do not need to create a new validation set for early stopping. What do you think?

python-package/lightgbm/sklearn.py
```diff
@@ -589,6 +592,16 @@ def __init__(
             The type of feature importance to be filled into ``feature_importances_``.
             If 'split', result contains numbers of times the feature is used in a model.
             If 'gain', result contains total gains of splits which use the feature.
+        early_stopping : bool, optional (default=False)
+            Whether to enable early stopping. If set to True, training will stop if the validation score does not improve
+            for a specified number of rounds (controlled by `n_iter_no_change`).
+        n_iter_no_change : int, optional (default=10)
+            If early stopping is enabled, this parameter specifies the number of iterations with no
+            improvement after which training will be stopped.
+        validation_fraction : float or None, optional (default=0.1)
```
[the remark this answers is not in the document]
ClaudioSalvatoreArcidiacono: Thanks for your comment, I have added a couple of tests for that, I have used patching to achieve that, let me know if you think there is more that can be improved in those tests.

python-package/lightgbm/sklearn.py
```diff
@@ -507,6 +507,9 @@ def __init__(
         random_state: Optional[Union[int, np.random.RandomState, np.random.Generator]] = None,
         n_jobs: Optional[int] = None,
         importance_type: str = "split",
+        *,
+        early_stopping: Union[bool, int] = False,
+        validation_fraction: float = 0.1,
```
ClaudioSalvatoreArcidiacono, line 512: @jameslamb, I have removed the option of having `validation_fraction=None` and performing early stopping on the train set.

Early stopping in lightgbm is disabled when the validation set equals the train set and I did not want to alter that behavior in this PR, so I have decided to deviate a little from the scikit-learn implementation of auto early stopping and to not allow `validation_fraction=None`. 

Let me know your thoughts on it.

tests/python_package_test/test_sklearn.py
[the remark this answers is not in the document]
jameslamb: > *I think that if a validation set has been provided in the fit() and the parameter early_stopping=True then we do not need to create a new validation set for early stopping.*

Very good point. I agree with this...let's not create a new validation set if there's already at least one in `validation-sets`. That simplifies this even more, and it'd still allow for early stopping to be done in a way that's consistent with `scikit-learn`.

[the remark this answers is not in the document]
jameslamb: > *I think the PR is ready for a last review. If you think that after this last attempt the gap to be filled is still too big*

Thank you but yes, I still think there is much to do. I'd like to take this PR over to get it finished.

I will do that after the next release is done #6796. I'm sorry, I just am not willing to invest the time it'd take to finish this between now and when we want to put up that release.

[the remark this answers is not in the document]
ClaudioSalvatoreArcidiacono: Sure, then go ahead and take it over :)

Let me know when you want me to review it.

PR #5812, opened 2023-03-30
[python-package] add more type annotations in basic.py

jameslamb:
Contributes to #3756.

Adds more type annotations in `basic.py`.

### How I tested this

Checked with `mypy`, confirmed that no new errors are introduced relative to `master`.

```shell
mypy --ignore-missing-imports python-package/lightgbm
```

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -2331,7 +2331,7 @@ class Dataset:
     def set_field(
         self,
         field_name: str,
-        data: Optional[Union[List[float], List[int], np.ndarray, pd_Series, pd_DataFrame]]
+        data: Optional[Union[List[List[float]], List[List[int]], List[float], List[int], np.ndarray, pd_Series, pd_DataFrame]]
     ) -> "Dataset":
         """Set property into the Dataset.
 


python-package/lightgbm/basic.py
```diff
@@ -2330,7 +2331,7 @@ def _reverse_update_params(self) -> "Dataset":
     def set_field(
         self,
         field_name: str,
-        data
+        data: Optional[Union[List[float], List[int], np.ndarray, pd_Series, pd_DataFrame]]
```
jmoralez, line 2334: This can also be list of lists

jmoralez (changes requested):

python-package/lightgbm/basic.py
jameslamb:
```suggestion
        data: Optional[Union[List[List[float]], List[List[int]], List[float], List[int], np.ndarray, pd_Series, pd_DataFrame]]
```

Ohhhhhhhh for multi-class `init_score`, right! Missed that when I read through the docstring, thanks for catching it.

jmoralez (approved):

PR #5818, opened 2023-04-04
[python-package] adding max_category_values parameter to create_tree_digraph method (fixes #5687)

### LINKED ISSUE #5687: plot tree with high cardinality feature
When I try to plot a tree that splits on a feature with high cardinality the overall diagram gets missed up no matter how I tried to adjust the width and height
out
There must be a way to ignore writing all the values that are used on the split within the nodes

moziada:
fixes #5687

### DIFF
--- a/python-package/lightgbm/plotting.py
+++ b/python-package/lightgbm/plotting.py
@@ -657,7 +657,7 @@ def create_tree_digraph(
         Single row with the same structure as the training data.
         If not None, the plot will highlight the path that sample takes through the tree.
     max_category_values : int, optional (default=32)
-        The maximum number of category values to display in tree nodes.
+        The maximum number of category values to display in tree nodes, if the number of thresholds is greater than this value, thresholds will be collapsed and displayed on the label tooltip instead.
 
         .. warning::
 
--- a/python-package/lightgbm/plotting.py
+++ b/python-package/lightgbm/plotting.py
@@ -455,7 +455,7 @@ def _to_graphviz(
     orientation: str,
     constraints: Optional[List[int]],
     example_case: Optional[Union[np.ndarray, pd_DataFrame]],
-    max_category_values: Optional[int],
+    max_category_values: int,
     **kwargs: Any
 ) -> Any:
     """Convert specified tree to graphviz instance.
@@ -516,10 +516,11 @@ def _to_graphviz(
                         missing_type_str=root['missing_type'],
                         default_left=root['default_left']
                     )
-            if max_category_values is not None:
-                if root['decision_type'] == '==' and len(root['threshold'].split('||')) > max_category_values:
+            if root['decision_type'] == '==':
+                category_values = root['threshold'].split('||')
+                if len(category_values) > max_category_values:
                     tooltip = root['threshold']
-                    threshold = '...'
+                    threshold = '||'.join(category_values[:2]) + '||...||' + category_values[-1]
 
             label += f"<B>{_float2str(threshold, precision)}</B>"
             for info in ['split_gain', 'internal_value', 'internal_weight', "internal_count", "data_percentage"]:
@@ -611,7 +612,7 @@ def create_tree_digraph(
     precision: Optional[int] = 3,
     orientation: str = 'horizontal',
     example_case: Optional[Union[np.ndarray, pd_DataFrame]] = None,
-    max_category_values: Optional[int] = None,
+    max_category_values: int = 32,
     **kwargs: Any
 ) -> Any:
     """Create a digraph representation of specified tree.
@@ -655,7 +656,7 @@ def create_tree_digraph(
     example_case : numpy 2-D array, pandas DataFrame or None, optional (default=None)
         Single row with the same structure as the training data.
         If not None, the plot will highlight the path that sample takes through the tree.
-    max_category_values : int, optional (default=None)
+    max_category_values : int, optional (default=32)
         The maximum number of category values to display in tree nodes.
 
         .. warning::
--- a/python-package/lightgbm/plotting.py
+++ b/python-package/lightgbm/plotting.py
@@ -612,7 +612,7 @@ def create_tree_digraph(
     precision: Optional[int] = 3,
     orientation: str = 'horizontal',
     example_case: Optional[Union[np.ndarray, pd_DataFrame]] = None,
-    max_category_values: int = 32,
+    max_category_values: int = 10,
     **kwargs: Any
 ) -> Any:
     """Create a digraph representation of specified tree.
@@ -656,7 +656,7 @@ def create_tree_digraph(
     example_case : numpy 2-D array, pandas DataFrame or None, optional (default=None)
         Single row with the same structure as the training data.
         If not None, the plot will highlight the path that sample takes through the tree.
-    max_category_values : int, optional (default=32)
+    max_category_values : int, optional (default=10)
         The maximum number of category values to display in tree nodes, if the number of thresholds is greater than this value, thresholds will be collapsed and displayed on the label tooltip instead.
 
         .. warning::
--- a/python-package/lightgbm/plotting.py
+++ b/python-package/lightgbm/plotting.py
@@ -479,7 +479,6 @@ def _to_graphviz(
         fillcolor = 'white'
         style = ''
         tooltip = None
-        threshold = root['threshold']
         if highlight:
             color = 'blue'
             penwidth = '3'
@@ -490,6 +489,7 @@ def _to_graphviz(
             shape = "rectangle"
             l_dec = 'yes'
             r_dec = 'no'
+            threshold = root['threshold']
             if root['decision_type'] == '<=':
                 operator = "&#8804;"
             elif root['decision_type'] == '==':
--- a/python-package/lightgbm/plotting.py
+++ b/python-package/lightgbm/plotting.py
@@ -722,9 +722,9 @@ def create_tree_digraph(
         feature_names=feature_names,
         precision=precision,
         orientation=orientation,
-        max_category_values=max_category_values,
         constraints=monotone_constraints,
         example_case=example_case,
+        max_category_values=max_category_values,
         **kwargs
     )
 
--- a/python-package/lightgbm/plotting.py
+++ b/python-package/lightgbm/plotting.py
@@ -654,6 +654,20 @@ def create_tree_digraph(
         Can be 'horizontal' or 'vertical'.
     max_category_values : int, optional (default=None)
         The maximum number of category values to display in tree nodes.
+
+        .. warning::
+
+            Consider wrapping the SVG string of the tree graph with ``IPython.display.HTML`` when running on JupyterLab to get the `tooltip <https://graphviz.org/docs/attrs/tooltip>`_ working right.
+
+            Example:
+
+            .. code-block:: python
+
+                from IPython.display import HTML
+
+                graph = lgb.create_tree_digraph(clf, max_category_values=5)
+                HTML(graph._repr_image_svg_xml())
+
     example_case : numpy 2-D array, pandas DataFrame or None, optional (default=None)
         Single row with the same structure as the training data.
         If not None, the plot will highlight the path that sample takes through the tree.
--- a/python-package/lightgbm/plotting.py
+++ b/python-package/lightgbm/plotting.py
@@ -453,9 +453,9 @@ def _to_graphviz(
     feature_names: Union[List[str], None],
     precision: Optional[int],
     orientation: str,
-    max_category_values: Optional[int],
     constraints: Optional[List[int]],
     example_case: Optional[Union[np.ndarray, pd_DataFrame]],
+    max_category_values: Optional[int],
     **kwargs: Any
 ) -> Any:
     """Convert specified tree to graphviz instance.
@@ -479,6 +479,7 @@ def _to_graphviz(
         fillcolor = 'white'
         style = ''
         tooltip = None
+        threshold = root['threshold']
         if highlight:
             color = 'blue'
             penwidth = '3'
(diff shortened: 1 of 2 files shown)


python-package/lightgbm/plotting.py
```diff
@@ -453,6 +453,7 @@ def _to_graphviz(
     feature_names: Union[List[str], None],
     precision: Optional[int],
     orientation: str,
+    max_category_values: Optional[int],
```
jmoralez, line 456: Can you please move this to the end of the signature (before kwargs)?

python-package/lightgbm/plotting.py
```diff
@@ -513,7 +515,13 @@ def add(
                         missing_type_str=root['missing_type'],
                         default_left=root['default_left']
                     )
-            label += f"<B>{_float2str(root['threshold'], precision)}</B>"
+            if max_category_values and root['decision_type'] == '==' and len(root['threshold'].split('||')) > max_category_values:
```
jmoralez, line 518: Can you make this check specifically for `None`, i.e. `if max_category_values is not None`? and maybe move the other two conditions to that inner if? I think if you're going through the code and you're not interested in the path that `max_category_values` takes its easier to see that single if and go to the else or skip it entirely.

python-package/lightgbm/plotting.py
```diff
@@ -602,6 +610,7 @@ def create_tree_digraph(
     show_info: Optional[List[str]] = None,
     precision: Optional[int] = 3,
     orientation: str = 'horizontal',
+    max_category_values: Optional[int] = None,
```
jmoralez, line 613: Can you move this before kwargs as well?

python-package/lightgbm/plotting.py
```diff
@@ -697,6 +722,7 @@ def create_tree_digraph(
         feature_names=feature_names,
         precision=precision,
         orientation=orientation,
+        max_category_values=max_category_values,
```
jmoralez, line 725: Also move this one to the bottom please

jmoralez (changes requested):
Thanks for your contribution! I've left some minor comments

python-package/lightgbm/plotting.py
moziada: I have defaulted ```threshold = root['threshold'] ``` right after setting ```tooltip = None``` and now the if condition looks better
```diff
@@ -479,6 +479,7 @@ def _to_graphviz(
         fillcolor = 'white'
         style = ''
         tooltip = None
+        threshold = root['threshold']
         if highlight:
             color = 'blue'
             penwidth = '3'
@@ -515,11 +516,10 @@ def _to_graphviz(
                         missing_type_str=root['missing_type'],
                         default_left=root['default_left']
                     )
-            if max_category_values and root['decision_type'] == '==' and len(root['threshold'].split('||')) > max_category_values:
-                tooltip = root['threshold']
-                threshold = '...'
-            else:
-                threshold = root['threshold']
+            if max_category_values is not None:
+                if root['decision_type'] == '==' and len(root['threshold'].split('||')) > max_category_values:
+                    tooltip = root['threshold']
+                    threshold = '...'

```

python-package/lightgbm/plotting.py
moziada: done

python-package/lightgbm/plotting.py
moziada: I think this caused a problem with the automated test, the line
```diff
+        threshold = root['threshold']
```

should be under the condition
```python
if 'split_index' in root:  # non-leaf
```
I will test it and commit it

moziada:
I can's see why R-package checks are failing

jameslamb:
please merge laster `master` into this branch to get the changes from #5821

python-package/lightgbm/plotting.py
```diff
@@ -603,6 +611,7 @@ def create_tree_digraph(
     precision: Optional[int] = 3,
     orientation: str = 'horizontal',
     example_case: Optional[Union[np.ndarray, pd_DataFrame]] = None,
+    max_category_values: Optional[int] = None,
```
jameslamb, line 614: Instead of defaulting to `None` and needing to deal with the complexity of `if max_category_values is None` conditions throughout the code, in my opinion it would be preferable to just choose a high default value like `50` or `100` or whatever you think could reasonably be expected to display well by default.

@jmoralez what do you think?

jmoralez: I agree with that, @moziada can you please make that change? The type of max_category_values would be an int with the default value that you choose.

python-package/lightgbm/plotting.py
```diff
@@ -513,7 +516,12 @@ def add(
                         missing_type_str=root['missing_type'],
                         default_left=root['default_left']
                     )
-            label += f"<B>{_float2str(root['threshold'], precision)}</B>"
+            if max_category_values is not None:
+                if root['decision_type'] == '==' and len(root['threshold'].split('||')) > max_category_values:
+                    tooltip = root['threshold']
+                    threshold = '...'
```
jmoralez, line 522: I didn't really think this value through, maybe it would be a better user experience if it was something like "collapsed". What do you think?

python-package/lightgbm/plotting.py
```diff
@@ -646,6 +655,22 @@ def create_tree_digraph(
     example_case : numpy 2-D array, pandas DataFrame or None, optional (default=None)
         Single row with the same structure as the training data.
         If not None, the plot will highlight the path that sample takes through the tree.
+    max_category_values : int, optional (default=None)
+        The maximum number of category values to display in tree nodes.
```
jmoralez, line 659: We should probably also state here that if the number of thresholds is greater than this value it will be collapsed and shown in the tooltip instead.

python-package/lightgbm/plotting.py
moziada: I don't think that I am following what do you mean by "collapsed", do you mean setting "collapsed" to ```threshold``` instead of the dots?

jmoralez: Yes, I'm not sure if the dots are clear enough for users to try to hover over them. Maybe the text "collapsed" is clearer or something that would signal to the users that in order to see the actual thresholds they need to hover over the node. Could be enough with the docstring though. What do you think?

moziada: I thought of something like turning "1001||1029||2384||2383||3484" into "1001||1029..." or "1001||1029...||3484"

jmoralez: I think that's definitely better, maybe just have the dots between the ors? Like "1001||1029||...||3484"

jameslamb: +1 to having `...` in the middle, I like that!

python-package/lightgbm/plotting.py
moziada: I have defaulted ```max_category_values``` to ```32``` as ```max_cat_threshold``` [parameter](https://lightgbm.readthedocs.io/en/latest/Parameters.html#max_cat_threshold). What do you think?

python-package/lightgbm/plotting.py
moziada: great, I committed that change

python-package/lightgbm/plotting.py
moziada: done, please have a look at the modification

python-package/lightgbm/plotting.py
jmoralez: I think your original example in #5687 has around 32 values
image
the `max_cat_threshold` refers to bins I think, so for example bin 0 could be 1||3, bin 1 2||5, etc.

 Maybe around 10 would be better. WDYT?

moziada: yes on the example it had 32 values, I made a commit changing the default value to 10

jmoralez (approved):
LGTM. Thanks a lot for your contribution!

jmoralez:
@jameslamb do you want to review this as well?

jameslamb (changes requested):
Thanks very much for this!

Can you please just add unit tests covering this functionality? That would give us more confidence that this is working and prevent it from being broken in the future.

Please add two tests, both with datasets that have at least one categorical feature that is informative and used in splits.

* one where the `len(category_values) > max_category_values` condition you've added is True
* one where that condition is false

You could, for example, copy this test:

https://github.com/microsoft/LightGBM/blob/ffb298626e131b9d213b34b3655846be441fdc12/tests/python_package_test/test_plotting.py#L160

moziada:
After looking at sklearn datasets I did not find any classification dataset that contains categorical features, I am thinking of creating multiple base datasets with [sklearn.datasets.make_classification](https://scikit-learn.org/stable/modules/classes.html#module-sklearn.datasets) each one has different distribution then merging them into one dataset and use the label of each base dataset as a categorical feature. what do you think?

jmoralez:
I think that could work, you just need to make sure that the target depends on those features somehow (so that they're chosen for some splits), maybe something like what we do here https://github.com/microsoft/LightGBM/blob/5989405cd3489f9892b7abb4a1feeeec702ac088/tests/python_package_test/test_dask.py#L185-L205

moziada:
I have created a quantized version of  breast cancer dataset and used the features as a categorical features. I made only one test case where the condition is false (categorical values should not be compressed), but the problem is the way that lightgbm splits on categorical features is not deterministic, for example a feature could have 30 different categorical values but a random indexed tree may be only splits on one categorical value. what do you suggest and I need your review

tests/python_package_test/test_plotting.py
```diff
@@ -21,6 +22,20 @@ def breast_cancer_split():
                             test_size=0.1, random_state=1)
 
 
+@pytest.fixture(scope="module")
+def categorical_data():
```
jameslamb, line 26: What's the value of having a fixture return a function like this, that the tests then have to call? Unless there's something I'm not understanding, I think this is unnecessary complexity. Please just define a function `def _create_categorical_data` in this module instead (not a fixture), just as many of the other tests in this project are structured.

tests/python_package_test/test_plotting.py
```diff
@@ -21,6 +22,20 @@ def breast_cancer_split():
...
+    def _categorical_data(category_values_lower_bound, category_values_upper_bound):
+        X, y = load_breast_cancer(return_X_y=True)
+        X_df = pd.DataFrame()
+        rnd = np.random.RandomState(0)
+        n_cat_values = rnd.randint(category_values_lower_bound, category_values_upper_bound, size=X.shape[1])
+        for i in range(X.shape[1]):
+            bins = np.linspace(0, 1, num=n_cat_values[i]+1)
+            X_df[f"cat_col_{i}"] = pd.qcut(X[:, i], q=bins, labels=range(n_cat_values[i])).as_unordered()
+        return train_test_split(X_df, y, test_size=0.1, random_state=1)
```
jameslamb, line 35: If the only test using this function is going to just train on whatever data you pass it, performing `train_test_split()` is unnecessary computation and complexity. Please just return `X_df` and `y` directly, to simplify this function.

tests/python_package_test/test_plotting.py
```diff
@@ -188,6 +203,37 @@ def test_create_tree_digraph(breast_cancer_split):
     assert 'count' not in graph_body
 
 
+@pytest.mark.skipif(not GRAPHVIZ_INSTALLED, reason='graphviz is not installed')
+def test_tree_with_categories_below_max_category_values(categorical_data):
+    X_train, _, y_train, _ = categorical_data(2, 10)
+    gbm = lgb.LGBMClassifier(n_estimators=10, num_leaves=3, verbose=-1)
```
jameslamb, line 209: > but the problem is the way that lightgbm splits on categorical features is not deterministic

This is not true. It's possible to get `lightgbm` to deterministically produce the same model given the same input data... it's just not deterministic *by default*.

You can use the following code (pay special attention to the values of `params`) as an example.

```python
import lightgbm as lgb
import numpy as np
import pandas as pd

X_df = pd.DataFrame()
rnd = np.random.RandomState(0)
n_cat_values = rnd.randint(2, 10, size=X.shape[1])
for i in range(X.shape[1]):
    bins = np.linspace(0, 1, num=n_cat_values[i]+1)
    X_df[f"cat_col_{i}"] = pd.qcut(X[:, i], q=bins, labels=range(n_cat_values[i])).as_unordered()

params = {
    "n_estimators": 10,
    "num_leaves": 3,
    "min_data_in_bin": 1,
    "force_col_wise": True,
    "deterministic": True,
    "num_threads": 1,
    "seed": 708,
    "verbose": -1
}

gbm1 = lgb.LGBMClassifier(**params)
gbm1.fit(X_df, y)
gbm1_df = gbm1.booster_.trees_to_dataframe()

# prove that in 100 repeated training runs, the exact same model is produced
for i in range(100):
    print(f"iteration: {i}")
    gbm = lgb.LGBMClassifier(**params).fit(X_df, y)
    pd.testing.assert_frame_equal(gbm1_df, gbm.booster_.trees_to_dataframe())
```

Try modifying your tests to pass similar values of `params`, to produce the same model every time the tests are run.

jameslamb (changes requested):
Nice idea binning the `breast_cancer` dataset to create an informative dataset of categorical features! I think that's a great approach for this PR's tests.

Please see my suggestions on how to proceed, and please add at least one test where the number of categories in a feature is greater than the value of `max_category_values` as well.

tests/python_package_test/test_plotting.py
moziada: That is interesting, I did not know that ```deterministic``` parameter exists

jameslamb:
@moziada sorry for the failing R CI jobs...those are not related to your PR. We've recently fixed that issue in #5859 .

Can you please merge latest `master` into this branch? Once you do that, I'll review the testing changes you've pushed.

moziada:
Are there any updates? @jameslamb

jameslamb (approved):
Looks good to me, thanks for the contribution!

PR #5826, opened 2023-04-10
[python-package] fix mypy error in Dataset.__init_from_list_np2d()

jameslamb:
Contributes to #3867 .

```text
python-package/lightgbm/basic.py:2038: error: Argument 1 to "c_int" has incompatible type "Optional[int]"; expected "int"  [arg-type]
```

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -2025,7 +2025,7 @@ class Dataset:
                 mats[i] = np.array(mat.reshape(mat.size), dtype=np.float32)
 
             chunk_ptr_data, chunk_type_ptr_data, holder = _c_float_array(mats[i])
-            if type_ptr_data is not None and chunk_type_ptr_data != type_ptr_data:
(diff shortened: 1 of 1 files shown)


python-package/lightgbm/basic.py
```diff
@@ -2008,7 +2008,7 @@ def __init_from_list_np2d(
             ptr_data = (ctypes.POINTER(ctypes.c_float) * len(mats))()
 
         holders = []
-        type_ptr_data = None
+        type_ptr_data = -1
```
jameslamb, line 2011: This should always get overwritten in the `for` loop below, given that the only call to this function checks that `mats` has at least one element

https://github.com/microsoft/LightGBM/blob/638014d5c56fb92396bd55f344a47e3f651a3cd4/python-package/lightgbm/basic.py#L1851-L1853

So this initial value, whether it's `None` or `-1`, should never end up actually getting used.

guolinke (approved):

PR #5837, opened 2023-04-18
[ci] [python-package] replace 'python setup.py' with a shell script

jameslamb:
Contributes to #5061.

This PR proposes replacing all existing direct invocations of `python setup.py` with a shell script, `build-python-package.sh`.

### What will the use experience be like?

For people using `pip install` or `conda install`, **this PR does not change anything**.

Anyone currently building / installing the Python package from sources here on GitHub using commands like the following:

```shell
cd python-package
python setup.py {command} [OPTIONS]
```

Would now instead run

```shell
sh ./build-python.sh {command} [OPTIONS]
```

### Why a shell script?

Most Python build tools (e.g. `wheel`, `build`, `pip wheel`, `scikit-build`, `hatchling`, etc.) assume a directory structure with all source files at or below the level of where the tool is called from. This repo contains more than just a Python package, and so it can't be directly used that way.

This project currently solves that problem via Python code in a `setup.py` file which reaches up into the parent directory and copying files at build time.

https://github.com/microsoft/LightGBM/blob/f74875ed60e696ee7d223ddb409e66f51bddbb47/python-package/setup.py#L61-L68

As described #5061, the Python packaging community is moving away from supporting such `setup.py` invocations.

I'm proposing that we use an `sh`-compatible shell script to create an isolated source directory *prior to invoking any Python build tools* for the following reasons:

* that approach has been working well for the R package for 2+ years (since #2960)
* `sh` is highly portable
* allows maximum flexibility to change the strategy for how the package is built in the future without affecting users
    - e.g., enables work in a followup PR to switch from `setup.py` to something like `scikit-build-core` (which I'm working on in #5759) or `hatchling` (like XGBoost is trying in https://github.com/dmlc/xgboost/pull/9021)
    - really important, I think, given how quickly the Python packaging tools are evolving

### How I tested this

In addition to all off the checks run in CI, I did the following:

* ran many combinations of the commands and options for this new script locally on my Mac
* checked the sdist and wheel contents using `pydistcheck`, to confirm that they have the same contents as those built on `main`, and are the same size
    - reference: #5838

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -23,7 +23,7 @@ clone_depth: 5
 
 install:
   - git submodule update --init --recursive  # get `external_libs` folder
-  - set PATH=%PATH:C:\Program Files\Git\usr\bin;=%  # delete sh.exe from PATH (mingw32-make fix)
+  #- set PATH=%PATH:C:\Program Files\Git\usr\bin;=%  # delete sh.exe from PATH (mingw32-make fix)
   - set PATH=C:\mingw-w64\x86_64-8.1.0-posix-seh-rt_v6-rev0\mingw64\bin;%PATH%
   - set PYTHON_VERSION=%CONFIGURATION%
   - set CONDA_ENV="test-env"
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -14,8 +14,8 @@ branches:
 
 environment:
   matrix:
-    # - COMPILER: MSVC
-    #   TASK: python
+    - COMPILER: MSVC
+      TASK: python
     - COMPILER: MINGW
       TASK: python
 
@@ -23,7 +23,6 @@ clone_depth: 5
 
 install:
   - git submodule update --init --recursive  # get `external_libs` folder
-  #- set PATH=%PATH:C:\Program Files\Git\usr\bin;=%  # delete sh.exe from PATH (mingw32-make fix)
   - set PATH=C:\mingw-w64\x86_64-8.1.0-posix-seh-rt_v6-rev0\mingw64\bin;%PATH%
   - set PYTHON_VERSION=%CONFIGURATION%
   - set CONDA_ENV="test-env"
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -14,8 +14,8 @@ branches:
 
 environment:
   matrix:
-    - COMPILER: MSVC
-      TASK: python
+    # - COMPILER: MSVC
+    #   TASK: python
     - COMPILER: MINGW
       TASK: python
 
--- a/build-python.sh
+++ b/build-python.sh
@@ -245,7 +245,7 @@ fi
 
 if test "${BUILD_WHEEL}" = true; then
     echo "--- building wheel ---"
-    rm -f ../dist/*.whl
+    rm -f ../dist/*.whl || true
     python setup.py bdist_wheel \
         --dist-dir ../dist \
         ${BUILD_ARGS}
--- a/build-python.sh
+++ b/build-python.sh
@@ -2,6 +2,8 @@
 
 set -e -u
 
+echo "building lightgbm"
+
 # Default values of arguments
 INSTALL="false"
 BUILD_SDIST="false"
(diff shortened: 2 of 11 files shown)


.appveyor.yml
```diff
@@ -23,7 +23,6 @@ clone_depth: 5
 
 install:
   - git submodule update --init --recursive  # get `external_libs` folder
-  - set PATH=%PATH:C:\Program Files\Git\usr\bin;=%  # delete sh.exe from PATH (mingw32-make fix)
```
jameslamb, line 26: This was originally put in to work around the following error using `mingw32-make.exe` + CMake:

```text
CMake Error at C:/Program Files (x86)/CMake/share/cmake-3.16/Modules/CMakeMinGWFindMake.cmake:12 (message):
  sh.exe was found in your PATH, here:
  C:/Program Files/Git/usr/bin/sh.exe
  For MinGW make to work correctly sh.exe must NOT be in your path.
  Run cmake from a shell that does not have sh.exe in your PATH.
  If you want to use a UNIX shell, then use MSYS Makefiles.
  CMakeLists.txt:39 (project)
```

If we're going to encourage the use of a shell script to build a development version of the package, that's a problem.

Thankfully, I found that passing a flag through to `CMake` that says "don't worry, I'm confident having `sh.exe` on `PATH` won't be a problem" ([as suggested in this Stack Overflow post](https://stackoverflow.com/a/45104058/3986677)) works for LightGBM's Python build!

python-package/setup.py
```diff
@@ -46,41 +46,6 @@ def find_lib() -> List[str]:
     return LIB_PATH
 
 
-def copy_files(integrated_opencl: bool = False, use_gpu: bool = False) -> None:
```
jameslamb, line 49: All of this copying-files-around stuff is unnecessary now the `setup.py` will only be invoked from the root of a directory containing all of the C/C++ source files + `CMake` scripts in the expected places.

python-package/setup.py
```diff
@@ -160,7 +125,8 @@ def compile_cpp(
             if use_mpi:
                 raise Exception('MPI version cannot be compiled by MinGW due to the miss of MPI library in it')
             logger.info("Starting to compile with CMake and MinGW.")
-            silent_call(cmake_cmd + ["-G", "MinGW Makefiles"], raise_error=True,
+            # ref: https://stackoverflow.com/a/45104058/3986677
+            silent_call(cmake_cmd + ["-G", "MinGW Makefiles", "-DCMAKE_SH=CMAKE_SH-NOTFOUND"], raise_error=True,
```
jameslamb, line 129: see the successful Appveyor MinGW build for evidence that this works: https://ci.appveyor.com/project/guolinke/lightgbm/builds/46835644/job/ms18944oeam9bfb5

build-python.sh
```diff
@@ -0,0 +1,334 @@
...
+      BUILD_WHEEL="true"
+      ;;
+    ############################
+    # customized library paths #
+    ############################
+    --boost-include-dir|--boost-include-dir=*)
+        if [[ "$1" != *=* ]];
+            then shift;
+        fi
+        BOOST_INCLUDE_DIR="${1#*=}"
```
jameslamb, line 100: this pattern allows both of these invocations to work corrrectly:

```shell
sh ./build-python.sh bdist_wheel --boost-include-dir /usr/local/src/Boost/include
sh ./build-python.sh bdist_wheel --boost-include-dir=/usr/local/src/Boost/include
```

jameslamb:
Alright I'm excited to say that this is ready for review!

If other maintainers agree with this proposal, it'll be a big step towards finally resolving #5061, and I think it'll make the next PR proposing the use of `scikit-build-core` a bit easier to review.

Since this is a significant breaking change, I'd ideally like approvals from all 3 of @shiyu1994 @guolinke and @jmoralez whenever you have time. Thank you for considering it!

jmoralez (approved):

jameslamb:
@guolinke could you look through this and let me know what you think?

I know that removing the ability to run `python setup.py {command}` will be disruptive for some users but I think that that's going to be forced on us eventually by the Python packaging tools like `setuptools` and `wheel`, based on all the documentation in #5061, and that switching to a shell script means we'll never have to break "build the Python package from `git` sources" this way again.

guolinke:
Sorry for missing this PR, will check it today

guolinke (approved):

jameslamb:
Thanks for the help everyone! This is a painful but I think necessary step towards keeping `lightgbm` up to date with changes in Python packaging practices.

And I hope this'll also be a step towards unlocking some other features that newer build systems offer, like cross-compilation for platforms we don't currently publish wheels for.

I strongly suspect that this one is going to be found from search engines, so after merging I'm going to lock discussion on this PR.

If you are reading this comment right now and want to say or ask something about this change, please comment on #5061 or open a new issue.

PR #5848, opened 2023-04-24
fix percentile computation for regression objectives

### LINKED ISSUE #5847: Median wrongly computed
```
import numpy as np
import lightgbm as lgb
from sklearn.metrics import mean_absolute_error
import pandas as pd

data = {"X": [1, 2, 3, 4, 5],
        "Y": [1, 2, 3, 4, 100]}

df = pd.DataFrame(data)

X_train, y_train = df[["X"]], df["Y"]

# Create a LightGBM dataset from the training data

# Set the parameters for the LightGBM model
params = {
    'boosting_type': 'gbdt',
    'objective': 'regression_l1',  # Use MAE for the objective
    'metric': 'mae',
    'num_leaves': 2,
    'min_data_in_leaf': 10,  # Set the minimum number of samples per leaf to 10
    'learning_rate': 1,
    'feature_fraction':1.0,
    'num_iterations': 1,    # Only use one tree
}

# Train the LightGBM model
model = lgb.train(params, train_data)

# Make predictions on the test data
y_pred = model.predict(X_train)

# Output the predictions
print("Predictions:")
print(y_pred)
```

Given the above codes, the tree won't split because there is not enough data, so it should output the median (3).
However, the output is:

```
[LightGBM] [Warning] There are no meaningful features, as all feature values are constant.
[LightGBM] [Info] Total Bins 0
[LightGBM] [Info] Number of data points in the train set: 5, number of used features: 0
[LightGBM] [Info] Start training from score 3.500000
Predictions:
[3.5 3.5 3.5 3.5 3.5]
```

I find the issue. It's this line:

https://github.com/microsoft/LightGBM/blob/ac37bf8a1dcc981dedadd9943e338e10ec072c01/src/objective/regression_objective.hpp#L27

Given cnt_data = 5, and alpha = 0.5, the pos of median should be 2. but it gives 2.5.

The correct formula should be:
```
const double float_pos = static_cast<double>(cnt_data - 1) * alpha;
```

Is this intentionally or is it a bug?

zachary62:
Fixes #5847


jameslamb:
@shiyu1994  @guolinke can you review this please?

Gso-1:
> ¿Puedes revisar esto, por favor?

Si pero necesito que se comprometa a pagar una suscripción de 15000 CLP

guolinke:
sorry for missing this PR. I think this PR indeed fixes the bug. 
cc @shiyu1994 to confirm it.

shiyu1994 (approved):

jameslamb:
@zachary62 please merge latest `master` into this branch and push a new commit, so we can see the CI run again.

In the future, I recommend creating a new branch on your fork when contributing to open source projects here on GitHub, instead of committing to `master` / `main`, as that makes that operation a bit easier. Let us know if you need help.

guolinke:
It seems we need to fix some tests

jameslamb (dismissed):
I just renamed this. Pull request titles automatically become release note items in this project. Users of LightGBM reading the release notes won't care that the fix was "simple" or how many lines the change was.

I have a few questions for you @guolinke @shiyu1994 

1. Is my reworded title correct?
2. If it is... should this be labeled `breaking`? If I understand correctly, won't this change the results of any models using regression objectives? (as indicated by some of those breaking tests )
3. Are you sure this fix is correct? I understand that the initial score being chosen right now is not exactly the median and is supposed to be...but the failing tests are showing that as of this change `lightgbm` produces a *worse* in-sample fit. Isn't that a problem?

e.g.

https://github.com/microsoft/LightGBM/blob/5c9e61d19084295fe1c75375baeb9242678d92b1/tests/python_package_test/test_engine.py#L135-L146

> FAILED tests/python_package_test/test_engine.py::test_regression[regression_l1] - assert 368.9313884458552 < 338
> FAILED tests/python_package_test/test_engine.py::test_regression[quantile] - assert 1314.2383346017639 < 1311

([build link](https://github.com/microsoft/LightGBM/actions/runs/5837535814/job/15833208887?pr=5848))

shiyu1994:
I have pushed a fix for the positions in percentile computation. Also, I increase the threshold value for l1 objective in the `test_regression`. Note that what we test in the test cases are the metrics on test set, not training set. So, it is possible that after the fix the performance on test set would downgrade. Also, the metrics are MSE, instead of MAE. For the l1_regression, it is optimizing the MAE. So, fix a bug used by the l1_regression doesn't ensure that we will get a better MSE value.

The show that the fix makes sense, the on the training set of `test_regression` for `l1_regression` before the fix is `12.631088946445049`, and after the fix is `12.588430387696558`.

jameslamb:
Ok thanks @shiyu1994 . I've dismissed my review so it doesn't prevent merging.

For my own understanding, can you explain what was incorrect about my proposed title `fix initial score computation for regression objectives`? 

Is it not true that this affects the initial score that is set at the beginning of boosting? I thought that's what was happening based on this log line in #5847:

> [LightGBM] [Info] Start training from score 3.500000

shiyu1994:
> Is it not true that this affects the initial score that is set at the beginning of boosting?

Yes, it is true. But not only the initial score. In `l1_regression` and `quantile` regression. The leaf values are refit after the tree is built. And the refit is according to the percentile calculation.

https://github.com/microsoft/LightGBM/blob/5c9e61d19084295fe1c75375baeb9242678d92b1/src/objective/regression_objective.hpp#L253-L283

jameslamb:
ohhhh I see. Ok thanks for explaining that!

Issue #4805, opened 2021-11-16
add early stopping in dart mode


benwu232:
I checed #1893 and #1895. It is said that early stopping is disabled in dart mode. The question is I don't know when to stop training in dart mode. Is it possible to add early stopping in dart mode? or is there any way found best model in dart mode? 

Thanks a lot for the brilliant lgb!

jameslamb:
Thanks for using LightGBM and for your question!

Per https://github.com/microsoft/LightGBM/issues/1893#issuecomment-444803315

> I think early stopping and dart cannot be used together.
The reason is when using dart, the previous trees will be updated.

You can learn more about DART in the original DART paper ([link](http://proceedings.mlr.press/v38/korlakaivinayak15.pdf)), especially the section "Description of the DART Algorithm".

> any way found best model in dart mode

One way to do this is to use hyperparameter tuning over parameter `num_iterations` (number of trees to create), limiting the model complexity by setting conservative values of `num_leaves`, `max_depth`, and `min_data_in_leaf`.

Maybe others will be able to propose alternative ideas. If you are sure you want to use early stopping even knowing that it will produce unstable results because of how DART works, I think it's possible by providing a custom callback function in the Python API. But I don't have time at the moment to create an example of that and I'm not certain which examples / docs to point you to.

benwu232:
Thanks a lot for the suggestions! 
I'll try those ideas. 
Will feedback if get any useful thing.

cheongalc:
Hello @jameslamb and @benwu232,

With reference to this:
> If you are sure you want to use early stopping even knowing that it will produce unstable results because of how DART works, I think it's possible by providing a custom callback function in the Python API.

I have referenced [this kaggle notebook](https://www.kaggle.com/code/stautxie/lightgbm-dart-boosting-save-best-model/notebook) to write an early stopping and model checkpoint function for DART.
```python
def dart_early_stopping(save_path: str, stopping_rounds: int):
    best_score = None
    best_iteration = -1
    counter = 0
    
    def _callback(env):
        nonlocal best_score
        nonlocal best_iteration
        nonlocal counter
        if env.evaluation_result_list is not None:
            # Format of example entry in env.evaluation_result_list:
            # ('valid_0', 'binary_logloss', 0.2946547510575049, False)
            # Index 0 = Result group name, in this case valid_0
            # Index 1 = Metric name, in this case binary_logloss
            # Index 2 = Numeric score, in this case 0.2946547510575049
[...]

PR #5867, opened 2023-05-05
[python-package] Add `progress_bar` callback using tqdm

34j:
Use tqdm to display training progress and metrics and, if necessary, early_stopping results. tqdm does not need to be installed and is compatible with the existing code.

Example:
```python
X_train, X_test, y_train, y_test = train_test_split(...)
gbm = lgb.LGBMRegressor()
early_stopping_callback = lgb.early_stopping(stopping_rounds=50)
callback = lgb.progress_bar("rich", early_stopping_callback=early_stopping_callback)
gbm.fit(
    X_train,
    y_train,
    eval_set=[(X_test, y_test)],
    callbacks=[early_stopping_callback, callback],
)
```

image

```python
X_train, X_test, y_train, y_test = train_test_split(...)
gbm = lgb.LGBMRegressor()
early_stopping_callback = lgb.early_stopping(stopping_rounds=50)
callback = lgb.progress_bar(early_stopping_callback=early_stopping_callback)
gbm.fit(
    X_train,
    y_train,
    eval_set=[(X_test, y_test)],
    callbacks=[early_stopping_callback, callback],
)
```

image
I tried `sys.stderr.flush()` and so on, but the output was inevitably broken when using `tqdm.std.tqdm`.

### DIFF
--- a/.ci/test.sh
+++ b/.ci/test.sh
@@ -134,7 +134,8 @@ conda create -q -y -n $CONDA_ENV \
     ${CONDA_PYTHON_REQUIREMENT} \
     python-graphviz \
     scikit-learn \
-    scipy || exit -1
+    scipy \
+    tqdm || exit -1
 
 source activate $CONDA_ENV
 
--- a/.ci/test.sh
+++ b/.ci/test.sh
@@ -135,7 +135,8 @@ conda create -q -y -n $CONDA_ENV \
     python-graphviz \
     scikit-learn \
     scipy \
-    tqdm || exit -1
+    tqdm \
+    types-tqdm || exit -1
 
 source activate $CONDA_ENV
 
--- a/.ci/test_windows.ps1
+++ b/.ci/test_windows.ps1
@@ -56,7 +56,8 @@ conda create -q -y -n $env:CONDA_ENV `
   "python=$env:PYTHON_VERSION[build=*cpython]" `
   python-graphviz `
   scikit-learn `
-  scipy ; Check-Output $?
+  scipy `
+  tqdm ; Check-Output $?
 
 if ($env:TASK -ne "bdist") {
   conda activate $env:CONDA_ENV
--- a/.ci/test_windows.ps1
+++ b/.ci/test_windows.ps1
@@ -57,7 +57,8 @@ conda create -q -y -n $env:CONDA_ENV `
   python-graphviz `
   scikit-learn `
   scipy `
-  tqdm ; Check-Output $?
+  tqdm `
+  types-tqdm ; Check-Output $?
 
 if ($env:TASK -ne "bdist") {
   conda activate $env:CONDA_ENV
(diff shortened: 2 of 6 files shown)


jameslamb (changes requested):
Thanks for the suggestion! But I don't support adding this much complexity and maintenance burden to the project in exchange for a progress bar.

I'd welcome the addition of this callback as *documentation*, like in https://github.com/microsoft/LightGBM/tree/master/examples/python-guide, but not as an addition to the project's public API which needs to be maintained.

@jmoralez what do you think?

jameslamb:
> there are so many LightGBM extension packages in this world that users get confused

What is this a reference to? Can you share some links to what you mean by "LightGBM extension packages"?

jameslamb:
> the Black formatting just adds lines and does not change the existing code

`lightgbm` does not `black`. I'm not sure what this comment is referring to.

> I don't think adding documentation or merging would make any difference in the maintenance burden

This is not true. Here are some examples of things that could be avoided if this change were only made in `examples/`:

* it wouldn't be necessary to make the callback serializable with `pickle`
* it wouldn't be necessary to hardcode a list of specific `tqdm` classes
* `lightgbm` would not pick up a new optional dependency on `tqdm` and need to support a wide range of versions of that library
* installing `types-tqdm` in CI wouldn't be necessary

---

Nothing in `lightgbm` makes the use of a callback such as the one proposed impossible today.

This PR is proposing going further than "make it possible", instead proposing "please make it a new part of the public API of the package", which comes with the maintenance burden of continuing to guarantee that the code works in the wide range of operating systems and Python versions LightGBM supports.

I am not convinced that the inconvenience of not having such code available via `pip install lightgbm` is so great that this should be added to the library and maintained here.

If you choose to answer the request in https://github.com/microsoft/LightGBM/pull/5867#pullrequestreview-1414652160 for some specific examples of the claim "there are so many LightGBM extension packages in this world that users get confused", that might change my opinion.

If another maintainer like @jmoralez makes a convincing case for adding something like this to the library, that might change my opinion.

But as of now, I am -1 on including this code in LightGBM.

I'd *consider* adding a simpler version of it (for example, without the hard-coded list of `tqdm` classes) in `examples/`.

34j:
I have given up on PR and created a package that addresses #5867, #4805, #3313 / #5808 at once and would appreciate if you could add it to the examples.

jameslamb:
> I have given up on PR and created a package

Alright, I'll close this.

> would appreciate if you could add it to the examples. https://github.com/34j/lightgbm-callbacks

At this time, it doesn't look to me like that project is being used by many people (according to https://pypistats.org/packages/lightgbm-callbacks) or that it's doing something so significantly different from what's possible with LightGBM that it should be added to https://github.com/microsoft/LightGBM#external-unofficial-repositories.

If that changes in the future, I'd happily consider a PR to add a mention of it to https://github.com/microsoft/LightGBM#external-unofficial-repositories.

If you'd like to answer the question I posed at https://github.com/microsoft/LightGBM/pull/5867#issuecomment-1536241916 with several examples, I'd also consider a callbacks-for-the-Python-package-specific section of the project's documentation at https://github.com/microsoft/LightGBM/blob/master/python-package/README.rst.

Issue #4308, opened 2021-05-20
[RFC] Add pylint in CI "lint" task


jameslamb:
## Summary

I think that [`pylint`](https://www.pylint.org/) could be used in this project to provide higher confidence that certain types of bugs in Python code will fail builds and not make it into releases.

## Motivation

For example, in #4143 I accidentally introduced a typo in a variable reference used in an f-string. This wasn't caught by review or CI, and happened on a line that wasn't covered by unit tests.

As I mentioned in https://github.com/microsoft/LightGBM/pull/4293/files#r636513196, this particular type of error would have been caught by `pylint`.

Other classes of error that can be caught by `pylint`:

* unnecessary use of a list inside a comprehension (e.g. `all([x for x in things])` instead of `all(x for x in things)`
* no value for argument in function call
* expected something to be a method (based on usage), found it as a property
* unused arguments
* redefining a builtin, like  id 
* class attributes defined for the first time outside of __init__() 
* raising an error in a try-catch should use raise {error} from {something} to give more accurate stack trace
* cyclic imports

<details><summary>full list of possible warnings, from "pylint --list-msgs" as of pylint 2.7.0 (click me)</summary>

```text
:blacklisted-name (C0102): *Black listed name "%s"*
  Used when the name is listed in the black list (unauthorized names).
:invalid-name (C0103): *%s name "%s" doesn't conform to %s*
  Used when the name doesn't conform to naming rules associated to its type
  (constant, variable, class...).
:empty-docstring (C0112): *Empty %s docstring*
  Used when a module, function, class or method has an empty docstring (it
  would be too easy ;).
:unneeded-not (C0113): *Consider changing "%s" to "%s"*
  Used when a boolean expression contains an unneeded negation.
:missing-module-docstring (C0114): *Missing module docstring*
  Used when a module has no docstring.Empty modules do not require a docstring.
:missing-class-docstring (C0115): *Missing class docstring*
  Used when a class has no docstring.Even an empty class must have a docstring.
:missing-function-docstring (C0116): *Missing function or method docstring*
  Used when a function or method has no docstring.Some special methods like
  __init__ do not require a docstring.
:singleton-comparison (C0121): *Comparison %s should be %s*
  Used when an expression is compared to singleton values like True, False or
  None.
:misplaced-comparison-constant (C0122): *Comparison should be %s*
  Used when the constant is placed on the left side of a comparison. It is
  usually clearer in intent to place it in the right hand side of the
  comparison.
[...]

PR #5872, opened 2023-05-07
[ci] [python-package] move tool configuration to pyproject.toml

jameslamb:
Contributes to #3867.
Contributes to #4308.
Contributes to #5061.

Proposes moving configuration for all Python linting tools that support it into `pyproject.toml` (and for `flake8`, which doesn't support it per https://github.com/PyCQA/flake8/issues/234, into `setup.cfg`).

Also adds configuration to skip running linters over the `lightgbm-python/` directory (new local build directory introduced in #5837), to help with local development.

### Notes for Reviewers

Many Python tools have been moving in this direction, ever since PEP 518 was accepted back in 2016.

From the PEP ([link](https://peps.python.org/pep-0518/#tool-table)) ...

> *The `[tool]` table is where any tool related to your Python project, not just build tools, can have users specify configuration data as long as they use a sub-table within `[tool]`, e.g. the `flit` tool would store its configuration in `[tool.flit]`.*

And projects *using those tools* have followed. For example, https://github.com/dmlc/xgboost/pull/9112.

### Benefits of this Change

Sets up `pyproject.toml` and `setup.cfg` as sources of configuration for the Python project, reducing the size of the diff to be reviewed in #5759.

Puts tool configuration into a format used by many other Python projects, making it a bit easier (I hope) for new contributors to work on LightGBM.


UriyaHarpeness (approved):

.ci/lint-python.sh
```diff
@@ -1,41 +1,32 @@
 #!/bin/sh
 
-DIR_TO_CHECK=${1}
-
 echo "running flake8"
 flake8 \
+    --config=./python-package/setup.cfg \
     --ignore=E501,W503 \
     --exclude=./.nuget,./external_libs,./python-package/build,./python-package/compile \
```
jmoralez, line 7: Doesn't setting the config mean we can remove these two lines?

jameslamb: ah yes! Missed this one, thank you. I'll push a commit in a bit removing them.

jameslamb: fixed in https://github.com/microsoft/LightGBM/pull/5872/commits/03e624bea6c4bfba8b531b40d124ec0ffd56eb67, thanks for catching that!

jmoralez (approved):

jameslamb:
closing and re-opening to try to get rid of that `license/cla` check 🙃 

Screen Shot 2023-05-08 at 3 59 08 PM

PR #5880, opened 2023-05-10
add CMakeLists options to disable building CLI, installing headers

jameslamb:
Contributes to #5061.

Proposes adding 3 options to `CMakeLists.txt`:

* `__BUILD_FOR_PYTHON` = indicate that lib_lightgbm is being built for use in a Python wheel
* `BUILD_CLI` = do / do not compile the `lightgbm` CLI when running `make` or `make install`
* `INSTALL_HEADERS` = do/do not move LightGBM's headers into `CMAKE_INSTALL_PREFIX` (e.g. `/usr/local/include/LightGBM/*.h`)

### Benefits of this change

Prevents unnecessary compilation of the CLI when building the Python package with `scikit-build-core`.

Prevents unnecessary generation of Makefile code for targets that won't ever be used by the R and Python packages.

### More Details

LightGBM's `CMakeLists.txt` generates Makefile(s) where `make` / `make install` has 3 effects:

* compile `lib_lightgbm.{dll/so}` + install it
* compile `lightgbm` (the CLI) + install it
* install LightGBM's header files to the default system location for headers

When building for the R and Python packages, the steps about the CLI and installing headers aren't required. LightGBM currently avoids those in those contexts by running `make _lightgbm` instead of `make` or `make install`.

https://github.com/microsoft/LightGBM/blob/1c873af9c069ac09afae37a9787ee4f7f0f0ad68/python-package/setup.py#L174

https://github.com/microsoft/LightGBM/blob/1c873af9c069ac09afae37a9787ee4f7f0f0ad68/R-package/src/install.libs.R#L138

To address #5061, in #5759 I'm proposing that LightGBM use `scikit-build-core` (https://github.com/scikit-build/scikit-build-core) as its build backend.

As of this writing, that project unconditionally runs `make install` on the `CMake`-generated makefiles ([docs link](https://github.com/scikit-build/scikit-build-core/blob/4bc101dd5e7cb49ca33bb978fb955286899a45b1/README.md#L50)). With that tool, there's no way to say "just build the `_lightgbm` target".

This PR's changes make it possible to use LightGBM's `CMakeLists.txt` in a way that `make` and `make install` only build the shared library.

### DIFF
--- a/.ci/test.sh
+++ b/.ci/test.sh
@@ -199,7 +199,12 @@ if [[ $TASK == "gpu" ]]; then
     if [[ $METHOD == "pip" ]]; then
         cd $BUILD_DIRECTORY && sh ./build-python.sh sdist || exit -1
         sh $BUILD_DIRECTORY/.ci/check_python_dists.sh $BUILD_DIRECTORY/dist || exit -1
-        pip install --user $BUILD_DIRECTORY/dist/lightgbm-$LGB_VER.tar.gz -v --install-option=--gpu || exit -1
+        pip install \
+            --user \
+            -v \
+            --install-option=--gpu \
+            $BUILD_DIRECTORY/dist/lightgbm-$LGB_VER.tar.gz \
+        || exit -1
         pytest $BUILD_DIRECTORY/tests/python_package_test || exit -1
         exit 0
     elif [[ $METHOD == "wheel" ]]; then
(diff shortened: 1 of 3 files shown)


.ci/test.sh
```diff
@@ -199,7 +199,12 @@ if [[ $TASK == "gpu" ]]; then
     if [[ $METHOD == "pip" ]]; then
         cd $BUILD_DIRECTORY && sh ./build-python.sh sdist || exit -1
         sh $BUILD_DIRECTORY/.ci/check_python_dists.sh $BUILD_DIRECTORY/dist || exit -1
-        pip install --user $BUILD_DIRECTORY/dist/lightgbm-$LGB_VER.tar.gz -v --install-option=--gpu || exit -1
+        pip install \
+            --user \
+            -v \
+            --install-option=--gpu \
+            $BUILD_DIRECTORY/dist/lightgbm-$LGB_VER.tar.gz \
+        || exit -1
```
jameslamb, line 207: The changes here in `test.sh` aren't strictly related to the `CMakeLists.txt` changes.

This isn't actually event changing anything about this script. Just breaking these arguments onto individual lines, to make the size of the diff in #5759 a bit smaller (and therefore those changes a bit easier to review).

CMakeLists.txt
```diff
@@ -15,8 +15,11 @@ set(
   "Semicolon separated list of sanitizer names, e.g., 'address;leak'. \
 Supported sanitizers are address, leak, undefined and thread."
 )
+option(BUILD_CLI "Build the 'lightbgm' command-line interface in addition to lib_lightgbm" ON)
```
jameslamb, line 18: Note that the new options `BUILD_CLI` and `INSTALL_HEADERS` both default to `ON` so that this PR preserves the current default behavior on `master`.

jameslamb:
@guolinke @shiyu1994 can you please help woth revuews on this and #5884 ? So I can hopefully continue making progress on #5061.

guolinke (approved):

jameslamb:
Thank you!

PR #5883, opened 2023-05-10
[ci] [python-package] fix mypy error in Dataset.set_categorical_feature()

jameslamb:
Contributes to #3867.

Fixes the following error from `mypy`:

```text
python-package/lightgbm/basic.py:2480: error: Value of type variable "SupportsRichComparisonT" of "sorted" cannot be "object"  [type-var]
```

### DIFF
--- a/python-package/lightgbm/basic.py
+++ b/python-package/lightgbm/basic.py
@@ -2478,7 +2478,7 @@ class Dataset:
             else:
                 if self.categorical_feature != 'auto':
                     _log_warning('categorical_feature in Dataset is overridden.\n'
-                                 f'New categorical_feature is {sorted(list(categorical_feature))}')
+                                 f'New categorical_feature is {list(categorical_feature)}')
                 self.categorical_feature = categorical_feature
                 return self._free_handle()
         else:


python-package/lightgbm/basic.py
```diff
@@ -2478,7 +2478,7 @@ def set_categorical_feature(
             else:
                 if self.categorical_feature != 'auto':
                     _log_warning('categorical_feature in Dataset is overridden.\n'
-                                 f'New categorical_feature is {sorted(list(categorical_feature))}')
+                                 f'New categorical_feature is {list(categorical_feature)}')
```
jameslamb, line 2481: `mypy` is saying here "Hey I'm not sure what'll be in `categorical_feature` at this point, so I'm not sure it's possible to sort it".

Honestly, I think sorting in this error message is confusing and unnecessary anyway. Sorting here means that the order of values in `categorical_feature` displayed in this warning could be different than the order in the list users supplied.

So I think removing `sorted()` is beneficial to reduce that confusion, and it's just a nice side benefit that it also resolves a `mypy` error.

guolinke (approved):

Issue #5865, opened 2023-05-02
[Dask] Race condition in finding ports


adfea9c0:
## Description
Dask LightGBM will sometimes try to bind to ports that were previously free, but are now used by a different program.

Specifically, it seems that the python code in LGBMDaskRegressor [1] finds open ports, saves the port number and then immediately closes them. After that the C++ layer will try reopening the port [2]. This can go wrong when another program binds to the port between these two steps.

I would say most of my runs succeed but I've ran into the 'LightGBMError: Binding port blah failed' error a handful of times now, and I'm fairly confident the above race condition is the issue.

[1] https://github.com/microsoft/LightGBM/blob/master/python-package/lightgbm/dask.py#L86
[2] https://github.com/microsoft/LightGBM/blob/d0dfceec377d34cddd6722f870d073f7aa64ca2d/src/network/linkers_socket.cpp#L128

## Reproducible example
It's kind of hard to reproduce this reliably since it's effectively a race condition. I hope my description of the issue suffices, let me know if I can do more to help.

## Environment info
I'm using LightGBM 3.3.2 on Dask.

jmoralez:
Hi @adfea9c0, thanks for raising this. Indeed that can happen because of the reasons you describe. The optimal solution IMO would be allowing to pass 0 as port to the C++ side and have that find a random open port and acquire it immediately. I'll take a look soon to see if that'd be possible.

jameslamb:
+1 thanks for reporting this so we have a specific issue to track. Linking some other related things:

* #3775
* #4942
* #4960
* https://github.com/microsoft/LightGBM/pull/5510#issuecomment-1263525112

@jmoralez I have one idea to consider that might be a quick way to make these issues less likely (inspired by @adfea9c0 's comment that `_find_n_open_ports()` immediately frees the ports it finds).

Maybe we could change the contract for `_find_n_open_ports()` to something like `_claim_n_open_ports()`, where it opens but does not close a socket to claim the port, and then have `_train_part()` somehow free that port again right before calling `.fit()` here?

https://github.com/microsoft/LightGBM/blob/d0dfceec377d34cddd6722f870d073f7aa64ca2d/python-package/lightgbm/dask.py#L323

All the time spent between the beginning of `_train_part()` and that line collecting and concatenating all the local data is time for another process to come in and claim that port. If we could instead know that the port is claimed by LightGBM until right before `.fit()` is called, the window of time for such conflicts could be reduced significantly.

-----

[...]

PR #5890, opened 2023-05-17
[dask] hold ports until training

jmoralez:
Contributes to #5865 by holding on to the ports that LightGBM will use during training for as long as possible on the Python side, thus decreasing the chance of a race condition (but not eliminating it).


jmoralez:
Seems like the actor approach won't work here due to https://github.com/dask/distributed/issues/7842. Feel free to try the alternative here @jameslamb.

python-package/lightgbm/dask.py
```diff
@@ -134,22 +98,27 @@ def _assign_open_ports_to_workers(
...
-        for worker, port in zip(workers.all_workers, found_ports[hostname]):
-            worker_to_port[worker] = port
-    return worker_to_port
+
+    # schedule futures to retrieve each element of the tuple
+    worker_to_socket_future = {}
+    worker_to_port_future = {}
+    for worker, socket_future in worker_to_future.items():
+        worker_to_socket_future[worker] = client.submit(operator.itemgetter(0), socket_future)
+        worker_to_port_future[worker] = client.submit(operator.itemgetter(1), socket_future)
```
jameslamb, line 116: very cool use of `operator.itemgetter()`, I'd never seen that before!

python-package/lightgbm/dask.py
```diff
@@ -134,22 +98,27 @@ def _assign_open_ports_to_workers(
...
+
+    # retrieve ports
+    worker_to_port = client.gather(worker_to_port_future)
```
jameslamb, line 119: This is the most important line, and I don't fully understand how it'll work.

If each `_RemoteSocket` contains a `socket.socket` in its attributes, and `socket.socket` attributes aren't pickleable with `cloudpickle`... 

```python
import cloudpickle
import socket

s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
s.bind(('', 0))

port = s.getsockname()[1]
print(f"port: {port}")

with open("out.pkl", "wb") as f:
    cloudpickle.dump(s, f)

# TypeError: cannot pickle 'socket' object
```

... how can an object created by a run of `_acquire_port()` on a worker be collected back to the client?

And even if it were... how does this work in a non-`LocalCluster` cluster, where each worker exists in the process space of a different machine?

jameslamb (commented):
Very nice, thanks so much for working on this! This is a much better version of what I was kind of envisioning in https://github.com/microsoft/LightGBM/issues/5865#issuecomment-1535577335.

I'm unsure about exactly how it works, but mainly because of my own ignorance of Dask and TCP. If you tell me this is working for you on both `LocalCluster` and a multi-machine distributed Dask cluster, then I think it's worth the time for you to add the remaining type hints, tests, docs, etc. and I'll commit to testing it myself on both `LocalCluster` and some different configurations of multi-machine clusters from Coiled or with something like `dask-cloudprovider`.

python-package/lightgbm/dask.py
```diff
@@ -319,6 +289,9 @@ def _train_part(
         if eval_class_weight:
             kwargs['eval_class_weight'] = [eval_class_weight[i] for i in eval_component_idx]
 
+    if remote_socket is not None:
+        remote_socket.release()
+        time.sleep(0.1)
```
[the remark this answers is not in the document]
jmoralez: I'm not sure how to see the state but I think that's exactly what happened. I saw some failures due to LightGBM not being able to bind the port so I added that to wait a bit, but allowing the reuse seems way better, especially because we're not sending anything through them. I'll make the changes.

python-package/lightgbm/dask.py
jmoralez: That's why we return a tuple from the function and only collect the second element (the integer containing the port number). The first element (which contains the socket) stays on the worker, we just keep the future and then use it when calling the train function, so it only gets used in its local worker.

It's kind of like an actor, we're creating the object on the remote worker and only keeping a reference to it, we don't ever ask for it in the client, so it never has to be serialized.

jameslamb: oooooooo interesting! I guess it's the combination of `workers=[worker]` and `pure=False` that prevents Dask from ever trying to move that result (the tuple with the `_RemoteSocket` and `port`) over the network? That's very clever, @jmoralez 😁 

I support it!

jameslamb:
Based on https://github.com/microsoft/LightGBM/pull/5890#discussion_r1209590377, I'm feeling pretty optimistic about this approach! Awesome work.

Some time this week, I'll try to test this on a multi-machine cluster using Coiled or `dask-cloudprovider`.

jmoralez:
Seems like coiled doesn't provide free credits to use their infra anymore, so I haven't been able to test this on a real cluster. I'll try to find an alternative because I don't have an AWS account.

jameslamb:
I can test on my AWS account.

python-package/lightgbm/dask.py
```diff
@@ -319,6 +291,8 @@ def _train_part(
         if eval_class_weight:
             kwargs['eval_class_weight'] = [eval_class_weight[i] for i in eval_component_idx]
 
+    if remote_socket is not None:
+        remote_socket.release()
```
jameslamb, line 295: very very very minor comment... but would you consider moving this block down a line, to after `model = model_factory(**params)`?

The `Network` isn't initialized until `Booster.__init__()`

https://github.com/microsoft/LightGBM/blob/07e3cf47f6a01cc6f7692a70ff5288eee43ce12d/python-package/lightgbm/basic.py#L3085-L3090

which isn't called until `{estimator}.fiit(...)`

https://github.com/microsoft/LightGBM/blob/07e3cf47f6a01cc6f7692a70ff5288eee43ce12d/python-package/lightgbm/sklearn.py#L839-L839

Switching the order would further reduce the risk of some other process claiming that port between when the socket is released and when LightGBM claims it during `{estimator}.fit()`.

jameslamb (approved):
I tested this tonight on AWS Elastic Container Service (ECS) cluster, using `dask-cloudprovider`: https://github.com/jameslamb/lightgbm-dask-testing/blob/main/notebooks/demo-aws.ipynb.

Happy to say it worked well!

Screen Shot 2023-06-19 at 12 00 58 AM

I left one small comment, do what you want with it.

Great work, @jmoralez !!!

python-package/lightgbm/dask.py
jmoralez: Changed in 9c7cabb

jameslamb: thanks!

jameslamb:
Looks great to me!

You should be the one to click the button, @jmoralez 😁

adfea9c0:
Hey only saw this now, but thanks for looking into it !

PR #5900, opened 2023-05-30
[docs] Update Quick Start

jaceklaskowski:
Make `lightgbm` command executable on command line

### DIFF
--- a/docs/Quick-Start.rst
+++ b/docs/Quick-Start.rst
@@ -59,14 +59,14 @@ Run LightGBM
 
 ::
 
-    "./lightgbm" config=your_config_file other_args ...
+    lightgbm config=your_config_file other_args ...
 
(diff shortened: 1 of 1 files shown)


jaceklaskowski:
This term "executable" is used figuratively. Sorry about that. It's just that the Installation Guide allows to install `lightgbm` binary in any location and since I'm on macOS (and installed it using `brew` and on the `PATH`) I found this requirement to have it in the root directory of this project too rigid. Kinda surprising given the guide is for users not developers / maintainers of this project who could have the binary installed in the project root. 

Using double quotes made it even more wishing for a change.

All in all, I made the simple changes for two reasons:

1. It's the very first open-source project where quotes are used to execute a binary on command line (that was unacceptable for me 😉)
2. I wanted to check out the process and how it really works here

HTH ❤️

jameslamb:
I've updated this branch with the latest changes from `master`, and will merge it when it builds.

